PR #3605, opened 2015-04-20
vector-esri-edit.html uses non api method

bartvde:
It uses getMap on the select and draw interaction but this is not marked as ```@api``` so this will not work in the hosted example.


bartvde:
PR attached

marcjansen:
Good to go Bart.

PR #3609, opened 2015-04-21
Useful linestring getLength method added to ol.proj

bill-chadwick:
New method parameters enable the generation of useful LineString length results from
geometries in any projection. Calculation method can be specified, with
a suitable default being selected. Projection must always be specified (a breaking change).
New tests have been added for ol.Sphere (RhumbLineDistance) and ol.geom.LineString (getLength method). The Measure example is corrected.

I anticipate a new PR for fixing the getArea method of ol.geom.LinearRing and ol.geom.Polygon. This method also needs a mandatory projection parameter and should have an optional calculation method - Cartesian or Spherical.

When there is API access to the GreatCircle generator in ol.geom.flat.geodesic then a new example plotting the Rhumb Line and Great Circle between New York and Tokyo and showing their lengths, would make sense.

### DIFF
--- a/src/ol/sphere/sphere.js
+++ b/src/ol/sphere/sphere.js
@@ -6,7 +6,6 @@
  */
 
 // FIXME add intersection of two paths given start points and bearings
-// FIXME add rhumb lines
 
 goog.provide('ol.Sphere');
 
@@ -135,11 +134,14 @@ ol.Sphere.prototype.finalBearing = function(c1, c2) {
 
 
 /**
- * Returns the distance from c1 to c2 using the haversine formula.
- *
+ * Returns the distance from c1 to c2 using the Haversine formula.
+ * The Haversine formula computes geodesic distance on the sphere.
+ * A geodesic or great-circle, is the shortest distance between two points
+ * on the sphere. Great-circles are often used for navigation in the air.
+ * A great-circle (other than a meridian) is a curved line on an EPSG:3857 map.
  * @param {ol.Coordinate} c1 Coordinate 1.
  * @param {ol.Coordinate} c2 Coordinate 2.
- * @return {number} Haversine distance.
+ * @return {number} Haversine spherical geodesic distance.
  * @api
  */
 ol.Sphere.prototype.haversineDistance = function(c1, c2) {
@@ -264,4 +266,37 @@ ol.Sphere.prototype.offset = function(c1, distance, bearing) {
       Math.sin(bearing) * Math.sin(dByR) * Math.cos(lat1),
       Math.cos(dByR) - Math.sin(lat1) * Math.sin(lat));
   return [goog.math.toDegrees(lon), goog.math.toDegrees(lat)];
+};
+
+
+/**
+ * Returns the Rhumb line distance between c1 and c2,
+ * calculated on the sphere.
+ * A Rhumb line is a line of constant bearing.
+ * Such a line is often used for navigation at sea.
+ * A Rhumb line is a straight line on an EPSG:3857 map.
+ *
+ * @param {ol.Coordinate} c1 Coordinate 1.
+ * @param {ol.Coordinate} c2 Coordinate 2.
+ * @return {number} Rhumb Line Length.
+ * @api
+ */
+ol.Sphere.prototype.rhumbLineDistance = function(c1, c2) {
+
+  var dLon = goog.math.toRadians(c2[0] - c1[0]);
+  var lat1 = goog.math.toRadians(c1[1]);
+  var lat2 = goog.math.toRadians(c2[1]);
+  var dLat = lat2 - lat1;
+
+  var dPsi = Math.log(Math.tan(Math.PI / 4 + lat2 / 2) /
+      Math.tan(Math.PI / 4 + lat1 / 2));
+
+      // E-W course becomes ill-conditioned with 0/0
+      var q = Math.abs(dPsi) > 10e-12 ? dLat / dPsi : Math.cos(lat1);
+
+      // if dLon over 180 degrees take shorter rhumb across anti-meridian:
+      if (Math.abs(dLon) > Math.PI) {
+        dLon = dLon > 0 ? -(2 * Math.PI - dLon) : (2 * Math.PI + dLon);
+      }
+      return Math.sqrt(dLat * dLat + q * q * dLon * dLon) * this.radius;
 };
--- a/examples/measure.html
+++ b/examples/measure.html
@@ -1,10 +1,12 @@
 ---
 template: example.html
 title: Measure example
-shortdesc: Example of using the ol.interaction.Draw interaction for creating simple measuring application.
+shortdesc: Example of using the ol.interaction.Draw interaction for creating a simple measuring application.
 docs: >
-  <p><i>NOTE: If use geodesic measures is not checked, measure is done in simple way on projected plane. Earth curvature is not taken into account</i></p>
-tags: "draw, edit, measure, vector"
+  The Linestring coordinates and the view projection are passed to ol.proj.getLength to get a distance in metres.
+  The length calculation method is Haversine on a sphere with the WGS84 equatorial radius.
+  Area computation is also done on a sphere with the WGS84 equatorial radius.
+tags: "draw, edit, measure, vector, length, area"
 ---
 <div class="row-fluid">
   <div class="span12">
@@ -17,5 +19,4 @@ tags: "draw, edit, measure, vector"
       <option value="length">Length</option>
       <option value="area">Area</option>
     </select>
-    <label class="checkbox"><input type="checkbox" id="geodesic"/>use geodesic measures</label>
 </form>
--- a/examples/measure.js
+++ b/examples/measure.js
@@ -139,7 +139,6 @@ var map = new ol.Map({
 map.on('pointermove', pointerMoveHandler);
 
 var typeSelect = document.getElementById('type');
-var geodesicCheckbox = document.getElementById('geodesic');
 
 var draw; // global so we can remove it later
 function addInteraction() {
@@ -244,18 +243,12 @@ typeSelect.onchange = function(e) {
  */
 var formatLength = function(line) {
   var length;
-  if (geodesicCheckbox.checked) {
-    var coordinates = line.getCoordinates();
-    length = 0;
-    var sourceProj = map.getView().getProjection();
-    for (var i = 0, ii = coordinates.length - 1; i < ii; ++i) {
-      var c1 = ol.proj.transform(coordinates[i], sourceProj, 'EPSG:4326');
-      var c2 = ol.proj.transform(coordinates[i + 1], sourceProj, 'EPSG:4326');
-      length += wgs84Sphere.haversineDistance(c1, c2);
-    }
-  } else {
-    length = Math.round(line.getLength() * 100) / 100;
-  }
+
+  var sourceProj = map.getView().getProjection();
+
+  length = Math.round(
+      ol.proj.getLength(line.getCoordinates(), sourceProj) * 100) / 100;
+
   var output;
   if (length > 100) {
     output = (Math.round(length / 1000 * 100) / 100) +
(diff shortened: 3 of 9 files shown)


probins:
you need to rebase this on the latest master. 'examples_src' for example is now called 'examples'

bill-chadwick:
I will need step by step instructions on how to do such a rebase.

ahocevar:
`git pull --rebase upstream master` should do the trick, assuming that `upstream` points to your `git://github.com/openlayers/ol3.git` remote.

bill-chadwick:
Thanks.  I will give that a go tonight.

src/ol/sphere/sphere.js
```diff
@@ -58,19 +57,19 @@ ol.Sphere.prototype.cosineDistance = function(c1, c2) {
...
+* Robert. G. Chamberlain and William H. Duquette, "Some Algorithms for
+* Polygons on a Sphere", JPL Publication 07-03, Jet Propulsion
+* Laboratory, Pasadena, CA, June 2007
+*
+* @param {Array.<ol.Coordinate>} coordinates List of coordinates of a linear
+* ring. If the ring is oriented clockwise, the area will be positive,
+* otherwise it will be negative.
+* @return {number} Area.
+* @api
+*/
```
tschaub: Please don't reformat comment blocks like this.  It adds unnecessary cruft to the patch.  Comment blocks should consistently be formatted like this:
```
/**
 * The `*` in this line is aligned with the first `*` in the line above (not the `/`).
 * Same goes for this line.  And the line below.
 */
```

tschaub:
The [`line.getLength()` method](http://openlayers.org/en/master/apidoc/ol.geom.LineString.html#getLength) is marked as stable.  So this change would require a new major release.

The current `getLength` method returns the Cartesian 2D length of a linestring.  This is consistent with other libraries that deal with "unreferenced" geometries (e.g. [GEOS](http://geos.osgeo.org/doxygen/classgeos_1_1geom_1_1LineString.html#a72998f51df9ab2fdee425345ead6068f), [JTS](http://www.vividsolutions.com/jts/javadoc/com/vividsolutions/jts/geom/LineString.html#getLength()), [PostGIS's `ST_Length`](http://postgis.net/docs/ST_Length.html)).

I'd suggest a new method or optional arguments.  A new method would allow people to more easily build profiles of the library without the extra sphere and ellipsoid code.  An alternative to adding dependencies and more code to the geometry classes would be to add functions to `ol.sphere` or `ol.ellipsoid`.  I think some discussion is warranted before we decide what is the most appropriate.

src/ol/sphere/sphere.js
bill-chadwick: I used fixjsstyle on that file. Perhaps I had outdated rules for that that
will come good when I rebaseline.

bill-chadwick:
I share your concerns.

I had earlier proposed a new method that always returned an answer in
metres whatever the projection.

I am not sure that you want ol.sphere and ol.ellipsoid having a method with
an ol.geom.LineString parameter. Rather, methods therein would have to take
a simple array of coordinates.

bill-chadwick:
My motivation in all this is to avoid pitfalls for users such as that reported in

https://github.com/openlayers/ol3/issues/3533

The existing ol3 measure example at 

http://openlayers.org/en/v3.4.0/examples/measure.html 

is IMHO misleading. The significant change in results (displayed with the same units) when enabling its misnamed 'geodesic' option is due to the latitude of the EPSG:3857 points, not the change from a cartesian to great-circle calculation method.

Folk coming from say Google Maps will be used to using  

https://developers.google.com/maps/documentation/javascript/reference#spherical

with 4326 coords only.

There is no getLength method on a GMaps Polyline 

https://developers.google.com/maps/documentation/javascript/reference#Polyline

Perhaps the ol.geom.LineString.getLength method (and polygon/linearring getArea) should be deprecated for now and removed from the API at a future date once alternatives returning useful answers are available. I suggest that for now, the documentation of the existing getLength and getArea methods are updated with a 4326/3857 health warning and the user pointed to the more useful calculation methods in ol.Sphere and ol.Ellipsoid.

Some debate is required as to where such a getLength method as I have proposed would best fit in the existing structure of OL3 - ol.proj perhaps ? (perhaps without a Vincenty method which requires ol.ellipsoid).

Are there any plans for the geodesic rendering of geometries, as per the geodesic option for a polyline/polygon on GMaps? On 3857, the straight lines currently rendered for geometries are of course Rhumb lines. A geodesic rendering is virtually essential for plotting aircraft routes of significant length.

Please could you let me know which, if any, of the bits of the work in my PR should be persued now. I suggest:

1) The rhumbLineDistance method in ol.Sphere 

2) New 'health warnings' for the ol.geom getLength and getArea methods with a steer to ol.Sphere and ol.Ellipsoid

3) Fixing of the Measure Example.

src/ol/sphere/sphere.js
tschaub: Thanks for the additional detail.  I'd recommend against using `fixjsstyle.py`.  And please ignore that the current lint task asks you to do so.  I don't know if any other developers use this, but clearly it is not doing the right thing.

bill-chadwick:
I have reworked this change thus - I guess the PR really needs a revised title or else migrating to a new PR with a new title.

A getLength method is added to ol.proj for computing the length of a line string. 
The new method gives a result in metres for the global EPSG:4326 and 
EPSG:3857 projections, a result in projection units for non-global projections
and a result in pixels for projections with units of Pixels. 
Cartesian, Haversine and Rhumb Line length calculation methods are offered.
A projection must be specified (I don't think a default of EPSG:3857 is appropriate).

A new Rhumb Line distance method is added to ol.Sphere.

New tests for the new ol.proj.getLength method and ol.Sphere.rhumbLineDistance method.

Several comments enhanced. A warning comment is added to the ol.geom.LineString.getLength
documentation.

The Measure example is updated to use the new ol.proj.getLength method and so give a linestring length in metres at any latitude. The example's area calulation is now always done on the sphere, again to give the same results at all latitudes.

The only new coupling is that ol.proj now 'requires' ol.Sphere.WGS84.
ol.proj already used ol.Sphere.NORMAL but that is not suited for length calculations on
EPSG:4326 and EPSG:3857. It may be appropriate for ol.proj to drop the use of ol.Sphere.NORMAL
and use ol.Sphere.WGS84 throughout.

I am not quite sure why ol3 has ol.Sphere.NORMAL which uses the authalic (equal area) Earth Radius of the North American 1927 datum (Clarke 1866 ellipsoid).

bill-chadwick:
I have omitted Vincenty as a calculation method to save coupling ol.proj to ol.Ellipsoid. If folk know enough GIS to think they need Vincenty for the calculation of WGS84 ellipsoidal lengths then they should also be able to write code to use it. Also, ol.Ellipsoid is not in the API yet.

bill-chadwick:
Changing the radius in ol.Sphere.NORMAL to be the same as that in ol.Sphere.WGS84 causes just one ol.proj test to fail, that for meters per unit for EPSG:4326.

  1) ol.proj ol.proj.Projection.prototype.getMetersPerUnit() returns value in meters:

      Error: expected 111319.49079327358 to sort of equal 111194.87428468118

I actually think that the meters per unit for EPSG:4326 might better be defined in terms of the Earth radius derived from the international standard for a Nautical Mile - http://en.wikipedia.org/wiki/Nautical_mile

bill-chadwick:
A bit more feedback on this would be good. Is there a way of re-triggering Travis without doing a new commit ?

Is it worth noting, in epsg3857projection.js, as a comment for developers, that the 'units' of EPSG:3857 are only whole metres on the equator ?

elemoine:
> Is there a way of re-triggering Travis without doing a new commit ?

You can for example amend your last commit and force-push to the remote branch:

```
git commit --amend --no-edit
git push --force origin projection
```

ahocevar:
Good discussion here, but closing due to lack of activity. @bill-chadwick if you want to continue working on this, please create new tickets that suggest one improvement at a time to make this whole enchilada easier to digest.

bill-chadwick:
I am having another go at this in a simpler fashion though adding a
'constantScale' attribute to a projection. Ol.proj.getLength will use
phythagoras for constantScale and pixel projections, else Haversine.
ConstantScale is only nominal but would be used for UTM and BNG etc where
scale changes by less than 0.05% over the projection's extent (and a custom
getPointResolution function returning an unchanged resolution is used.)

It would make sense to me to do getArea at the same time - it would return
an answer in square metres for 3857 and 4326 projections using the
ol.sphere methods.

I would also like then to rework the measure example.

Would you prefer that as one PR or three?

Bill

bill-chadwick:
The constantScale attribute of a projection can replace the custom
getPointResolution method

ahocevar:
This sounds like a decent plan. It's always good when things can be simplified. I'd call the proposed `constantScale` option `constantPointResolution` instead. And I like the idea of removing the `getPointResolution` option from `olx.ProjectionOptions`. You can create a single pull request (or reuse this one) for the changes to getLength, getArea and the measure example.

bill-chadwick:
Thanks Andreas,

Your message and my new post to the ol3 google group crossed in the ether.

I will change constantScale to constantPointResolution

Observe that constantPointResolution will be automatically set true for
projections with units of Pixels, which are I think, not properly handled
by getPointResolution currently

Bill.

bill-chadwick:
I think that projections with units of degrees should always have
constantPointResolution. Rather than special case those in
getPointResolution the new getLength should make a special case for
projections with units  of degrees.

Bill

ahocevar:
> Rather than special case those in getPointResolution the new getLength should make a special case for projections with units  of degrees.

If projections with degrees as units have `constantPointResolution` set to true as well, then the special case is the same as for pixel units and can be handled in getLength. However, make sure that `getPointResolution` returns the correct value also for `constantPointResolution` projections in general.

bill-chadwick:
Andreas,

Out of curiosity, do you know where the value of earth radius in
ol.sphere.normal had come from? Its a bit different to the nornally used
value for the WGS84 authalic radius.

Bill

Rather than special case those in getPointResolution the new getLength
should make a special case for projections with units of degrees.

If projections with degrees as units have constantPointResolution set to
true as well, then the special case is the same as for pixel units and can
be handled in getLength. However, make sure that getPointResolution returns
the correct value also for constantPointResolution projections in general.

—
Reply to this email directly or view it on GitHub
<https://github.com/openlayers/ol3/pull/3609#issuecomment-150139433>.

ahocevar:
> Out of curiosity, do you know where the value of earth radius in ol.sphere.normal had come from? Its a bit different to the nornally used value for the WGS84 authalic radius.

It comes from the Normal Sphere as defined in the MetaCRS projects. I think the reason why that is used for the Haversine distance calculcations instead of the WGS84 Sphere is because the latter uses the radius at the equator, and we want a good approximation anywhere on the planet, instead of an accurate calculation for the equator only.

bill-chadwick:
Thats my point. I think we should be using the WGS84 authalic (not
equatorial) radius which is just 10 metres more than the value in
ol.sphere.normal.
Bill

bill-chadwick:
I will check the value used by some other web apis. It would be good to use
the same value.

bill-chadwick:
Leaflet looks to be using 6378137, which is the WGS84 equatorial radius,
for its Haversine length calculation.

Google do the same, defaulting to 6378137 see here
https://developers.google.com/maps/documentation/javascript/reference#spherical


I am not sure that 6378137 is the best choice.


https://en.wikipedia.org/wiki/Great-circle_distance suggests using a mean
WGS84 radius of
6371008. The WGS84 authalic radius is 6371007.

*Chris Veness at *http://www.movable-type.co.uk/scripts/latlong.html
defaults to 6371000

The current value in ol.sphere.NORMAL is 6370997

I have added an optional radius parameter to the new ol.proj.getLength and
getArea methods, defaulting to ol.sphere.NORMAL if a radius is not provided.

Bill

tschaub:
We now have `getLength()` in the `ol/sphere` module.  If there are improvements or fixes that need to be made, they should be made there.

Thanks.

PR #3611, opened 2015-04-22
Improve error handling in Esri JSON format

bartvde:
Currently the error message is simply passed to readFeaturesFromObject


bartvde:
Example only so merging without formal review.

PR #3601, opened 2015-04-20
Document how to use draw interaction

probins:
This wasn't clear to me, so after checking what it does I decided to write it up.

A related question: are there any plans to add methods to write drawn/modified features to a server, as with OL2's save strategy?

PR #3616, opened 2015-04-22
Add support for freehand drawing to the Draw interaction

### LINKED ISSUE #2826: Freehand Drawing with ol.interaction.Draw
New Feature request:

I was trying out the ol.interaction.Draw feature in Openlayers 3 and couldn't find a way to do freehand drawing like in OpenLayers 2.  Is this possible currently in OpenLayers 3? If not is there a plan to add? 

Here's an example of what I am trying to do implemented in OpenLayers 2

http://www.codechewing.com/demo/openlayers/freehand-mode-with-openlayers-draw-line.html

I started this question on google group and looks like , this feature does not exist yet. Just wanted to put in a feature request here.

Issue on google group:

https://groups.google.com/forum/#!topic/ol3-dev/Hxri0Qx0gPU

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/5246996-freehand-drawing-with-ol-interaction-draw?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ahocevar:
Linestrings and polygons can now be drawn in freehand mode, which is activated by holding down the Shift key (default) or any other condition configured by the user with the new `freehandCondition` constructor option.

Fixes #2826.

### DIFF
--- a/examples/draw-features.html
+++ b/examples/draw-features.html
@@ -3,8 +3,11 @@ template: example.html
 title: Draw features example
 shortdesc: Example of using the ol.interaction.Draw interaction.
 docs: >
-  Example of using the ol.interaction.Draw interaction.
-tags: "draw, edit, vector"
+  Example of using the Draw interaction. Select a geometry type from the
+  dropdown above to start drawing. To finish drawing, click the last
+  point. To activate freehand drawing for lines and polygons, hold the `Shift`
+  key.
+tags: "draw, edit, freehand, vector"
 ---
 <div class="row-fluid">
   <div class="span12">
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2360,7 +2360,8 @@ olx.interaction.DragZoomOptions.prototype.style;
  *     minPointsPerRing: (number|undefined),
  *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
  *     geometryName: (string|undefined),
- *     condition: (ol.events.ConditionType|undefined)}}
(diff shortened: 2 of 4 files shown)


tschaub:
Freehand mode is back!  This looks good to me @ahocevar.  I've lightly tested the drawing related examples and think the code change looks sensible.  It probably makes sense to let others test a bit, but I think this looks good to go.

tsauerwein:
I did a quick test and it seems to work pretty well!

Maybe we can add a short description to the draw-features examples and also to the class documentation of `ol.interaction.Draw` (see also #3601) on how to use the freehand drawing. Because I had to do a few tries to figure out that you have to press SHIFT when you start the dragging.

bartvde:
I agree on adding freehand as a keyword to one of the draw examples (draw-features.html?) at least and adding a descriptive text as well.

I had a bit of trouble finishing a line. I thought letting go of the shift key would do the trick, but it doesn't, only double click. But maybe this is just getting used to the functionality. Because also some of the time when doing the double click to finish I get into zoom box drawing mode.

But overall this is so great!

tsauerwein:
> I had a bit of trouble finishing a line. I thought letting go of the shift key would do the trick, but it doesn't, only double click.

You can also do a single-click at the last position.

ahocevar:
Thanks everyone for reviewing and testing!

PR #3617, opened 2015-04-22
Add ol.size.hasArea

tschaub:
This pulls the `hasArea` function from `map.js` and puts it in `size.js`.  The compiled output is the same with and without this, so it is mainly a stylistic change.  I'm working on removing cases where we shadow variables.  An alternative would be to rename the argument for `hasArea`, but it seems more appropriate to have `ol.size.hasArea`.


ahocevar:
Looks good to me. Please merge when Travis says so.

PR #3613, opened 2015-04-22
Add a one sentence summary for ol.interaction* exportable symbols

marcjansen:
This PR adds basic documentation to `@api` symbols of `ol.interaction`.

Please review.

See also

* #3488: first batch of docs by @tschaub)
* #3497: overview issue of missing docs
* #3583: docs for `ol.proj.*`, `ol.layer.*`
* #3580: docs for `ol.source.*`

PR #3583, opened 2015-04-18
Add a one sentence summary for ol.proj.* and ol.layer.* exportable symbols

marcjansen:
This PR adds basic documentation to `@api` symbols of `ol.proj.*` and `ol.layer.*`.

Please review.

See also #3497, #3488 and #3580.

PR #3580, opened 2015-04-17
Add a one sentence summary for ol.source.* exportable symbols

marcjansen:
This PR adds basic documentation to `@api` symbols, and is a follow-up of #3497 (which itself is a follow-up of #3488).

Please review.

Issue #3497, opened 2015-04-04
Add a one sentence summary for all exportable symbols.


tschaub:
We should have a one sentence (plain text) summary of all our exportable symbols (everything with an `@api` annotation).

The first batch of this work is handled in #3488.  Here's what remains:

 - [x] `ol.Sphere.*` (all `@api`-symbols are documented already)
 - [x] `ol.source.*` (#3580)
 - [ ] `ol.render.*`
 - [x] `ol.proj.*` (#3583)
 - [x] `ol.layer.*` (#3583)
 - [x] `ol.interaction.*` (#3613)
 - [x] `ol.geom.*` (#3619)
 - [ ] `ol.format.*` (#3624)
 - [ ] `ol.events.*` (#3624)
 - [x] `ol.dom.*` (nothing to document)
 - [ ] `ol.control.*` (#3624)
 - [x] `ol.color.*` (all `@api`-symbols are documented already)
 - [ ] `ol.MapBrowserPointerEvent.*` (#3624)
 - [x] `ol.renderer.*`  (all `@api`-symbols are documented already)

marcjansen:
This can be closed now.

PR #3619, opened 2015-04-22
Add a one sentence summary for ol.geom.* exportable symbols

marcjansen:
This PR adds basic documentation to `@api` symbols of `ol.geom*`. This never ends...

Please review.

See also

* #3488: first batch of docs by @tschaub
* #3497: overview issue of missing docs
* #3583: docs for `ol.proj.*`, ol.layer.*`
* #3580: docs for `ol.source.*`
* #3613: docs for `ol.interaction*`

### DIFF
--- a/src/ol/geom/circle.js
+++ b/src/ol/geom/circle.js
@@ -85,6 +85,7 @@ ol.geom.Circle.prototype.containsXY = function(x, y) {
 
 
 /**
+ * Return the center of the circle as {@link ol.Coordinate coordinate}.
  * @return {ol.Coordinate} Center.
  * @api
  */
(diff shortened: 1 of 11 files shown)


src/ol/geom/circle.js
```diff
@@ -85,6 +85,7 @@ ol.geom.Circle.prototype.containsXY = function(x, y) {
 
 
 /**
+ * Return the center of the circle as {@link ol.Coordinate coordinate}.
```
tschaub: This is something we'll probably have to deal with, but the builder client doesn't do any special parsing of JSDoc's custom `@lnk` style annotations.  It would reduce work in the near-term to just have plain text for the first sentence.

tschaub:
Nice @marcjansen.  Since my comment above applies to a number of commits and also reflects an issue with current docs, I don't think you have to deal with it here.  We can regex our way around the `@link` annotations in the build client.

marcjansen:
Cool, thanks, @tschaub.

PR #3623, opened 2015-04-23
ol3 overwrites WMS format_options instead of extending them.

bartvde:
I am working on an ol3 application that sends a WMS getMap request with the `FORMAT_OPTIONS` parameter set. If I run this application on a high-resolution display, the value of `FORMAT_OPTIONS` is overwritten by `FORMAT_OPTIONS=dpi%3A180`.

### DIFF
--- a/src/ol/source/imagewmssource.js
+++ b/src/ol/source/imagewmssource.js
@@ -295,7 +295,11 @@ ol.source.ImageWMS.prototype.getRequestUrl_ =
     switch (this.serverType_) {
       case ol.source.wms.ServerType.GEOSERVER:
         var dpi = (90 * pixelRatio + 0.5) | 0;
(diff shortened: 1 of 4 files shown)


bartvde:
PR attached TIA for any review.

ahocevar:
Doesn't this append the dpi string with every call so you'll end up with `,dpi:180,dpi:180,dpi:180`...?

bartvde:
good point, will fix

bartvde:
I'm actually unable to get into the situation you describe @ahocevar even after clearing the tile cache, but maybe I'm missing a possible code path here.

ahocevar:
You're right @bartvde. I just verified by adding another `source.getImage()` call to the tests. Thanks for fixing this. Please merge.

elemoine:
LGTM too. Thanks Bart.

PR #3625, opened 2015-04-23
Cluster Data Source, Resolves Panning Issue

cmiles74:
This patch resolves the issue where the cluster data source wouldn't refresh it's data when the map was panned (only when zoomed). More about this issue can be found [in this discussion thread](https://groups.google.com/forum/#!topic/ol3-dev/N7TGCRax_u0). I have also added a new example that demonstrates this working.

The other change alters the build script to use 'python2' when building if 'python2' is in the environment path. On some systems, like Arch Linux, there will be a 'python' with point to Python 3 and a 'python2' that points to Python 2.

I have submitted a OpenLayers Individual Contributor License Agreement. :-)

### DIFF
--- a/build.py
+++ b/build.py
@@ -15,6 +15,7 @@ from pake import Target
 from pake import ifind, main, output, rule, target, variables, virtual, which
 from Queue import Queue
 from threading import Thread
+from distutils import spawn
 
 def sigint_handler(signal, frame):
     print('Exiting')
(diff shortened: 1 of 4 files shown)


cmiles74:
I only changed a couple of lines in these files. Do you need me to bring these files into compliance with the linter in order to accept this pull request? If possible, I would prefer not to. `;-)`

alvinlindstam:
(I'm not a core contributor, so might be wrong)

Your pull request must pass Travis in order to be merged. You also need to clean up the commit history. Additionally, each PR should concern a single issue of functionality, so your build script change should be a separate PR.

Regarding your changes: Reclustering on each pan seems overkill, since many users might not use bbox feature loaders. The Cluster source listens to change events on the source, so simply calling loadFeatures on the base source should suffice.

Also, a test case might be nice.

How about keeping the resolution test, but moving the source.loadFeatures call out of it (before)? I have not tested it yet.

cmiles74:
I can cleanup the commit history and break this into two pull requests,
that will get done later today.

cmiles74:
Closing in order to break into three separate pull requests, even though this pull request has three commits. Note that the last three commits were only there to make the linter happy.

PR #3461, opened 2015-04-02
High level Modify interaction events

sweco-sebhar:
Related to the desired feature as discussed in #2946.

I think this PR needs test cases if the event is to be marked as stable. Also, events are currently triggered without checking for actual changes in the features(s) which probably should be avoided.

PR #3628, opened 2015-04-24
Add test case to verify default select toggling behaviour

sweco-sebhar:
While working on understanding/removing the undesired behaviour of Modify interaction as described in #3461 I had need of writing tests for the Modify interaction. To do this I need to simulate several consecutive single clicks and drags but I get unexpected results.

This PR introduces a test case (that unexpectedly fails) for the Select interaction that triggers two single clicks a non overlapping geometry.

From what I can gather Map#forEachFeatureAtPixel is called for each event which is expected but finds nothing on the second event.

I suspect the event/render simulation code but haven't been able to figure out what might be wrong with it.

### DIFF
--- a/test/spec/ol/interaction/selectinteraction.test.js
+++ b/test/spec/ol/interaction/selectinteraction.test.js
@@ -18,6 +18,9 @@ describe('ol.interaction.Select', function() {
     document.body.appendChild(target);
 
     var geometry = new ol.geom.Polygon([[[0, 0], [0, 40], [40, 40], [40, 0]]]);
+    var geometry2 = new ol.geom.Polygon([[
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Debugging further seem to indicate that the imageData inspected in ol.render.canvas.ReplayGroup#forEachFeatureAtCoordinate is 0 instead of 255 as expected on the second simulated single click.

I've been able to write the tests I wanted for Modify interaction in #3630 without the Select interaction so I'm giving up on this one for now.

ahocevar:
Closing due to lack of activity. @bjornharrtell please reopen if you continue working on this or want to discuss.

PR #3629, opened 2015-04-24
Problems with XYZ coordinates in snap interaction

tsauerwein:
As reported on the [mailing list](https://groups.google.com/forum/#!topic/ol3-dev/BtIOS4vd4_M), the snap interaction has issues with XYZ coordinates.

When the mouse pointer is snapped to a vertex with XYZ coordinates, the vertex coordinate with three values will be assigned to `event.coordinate` (see https://github.com/openlayers/ol3/blob/master/src/ol/interaction/snapinteraction.js#L586). Because `event.coordinate` is expected to have only two values, this will cause problems elsewhere.

I will work on a fix.


tsauerwein:
PR attached.

elemoine:
LGTM.

tsauerwein:
Thanks for the review, Eric.

PR #3630, opened 2015-04-26
Initial tests for Modify interaction vertex creation

bjornharrtell:

PR #3631, opened 2015-04-26
Restore Modify interaction constructor test

bjornharrtell:
I accidentally removed the existing constructor test that also tested some rbush internals when making additional tests in #3630, sorry. This PR should restore what I removed.


elemoine:
Thanks for fixing this @bjornharrtell.

PR #3632, opened 2015-04-26
Make host-examples target copy index.js

elemoine:
Since https://github.com/openlayers/ol3/pull/3527 the examples index page is empty. See http://openlayers.org/en/master/examples/. This is because the `host-examples` target of the `Makefile` does not add the `index.js` file to the "hosted" examples directory. This PR fixes the problem.


tschaub:
Looks good @elemoine.  Please merge.

PR #3635, opened 2015-04-27
Create vertex on boundary single click

bjornharrtell:
This PR changes the current behaviour of Modify interaction in that it will create a vertex when single clicking a boundary and raise revision by 1. 

Previously, single clicking a boundary would create a vertex but immediately delete it resulting in unmodified geometry but with revision raised by 2.

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -142,6 +142,14 @@ ol.interaction.Modify = function(options) {
   this.lastPixel_ = [0, 0];
 
   /**
+   * Keep track of the last inserted pixel location to avoid
+   * unintentional deletion.
+   * @type {ol.Pixel}
+   * @private
+   */
+  this.lastNewVertexPixel_ = [NaN, NaN];
+
+  /**
    * Segment RTree for each layer
    * @type {Object.<*, ol.structs.RBush>}
    * @private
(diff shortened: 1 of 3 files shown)


bjornharrtell:
Ready for review. Ping @tschaub.

bjornharrtell:
Rebased on 3.5.0 master to keep this ready for review.

src/ol/interaction/modifyinteraction.js
```diff
@@ -85,6 +85,14 @@ ol.interaction.Modify = function(options) {
   this.lastPixel_ = [0, 0];
 
   /**
+   * Keep track of the last inserted pixel location to avoid
+   * unintentional deletion.
+   * @type {ol.Pixel}
+   * @private
+   */
+  this.lastNewVertexPixel_ = [-1, -1];
```
tschaub: This could be initialized to `[NaN, NaN]`.

tschaub:
This looks like a nice solution @bjornharrtell.  What do you think about clearing the `lastNewVertexPixel_` on `pointermove`?  This would presumably allow someone to add a vertex and then delete it again (though I haven't confirmed that this doesn't work with your change).

src/ol/interaction/modifyinteraction.js
tsauerwein: Or maybe `this.lastNewVertexPixel_ = null`?

tschaub: Though I think that would break the checks below, right?

tsauerwein: I only noticed the checks below after I send my comment. You are right, `[NaN, NaN]` is cleaner because you don't have to do an additional check (`goog.isNull`).

bjornharrtell:
Changed to initialize with NaN. Clearing `lastNewVertexPixel_` in `handlePointerMove_` does not seem to affect behaviour or test case outcome. I think the behaviour is as @tschaub described already, if I'm not misunderstanding. If possible, run the modify-features on this branch to verify.

bjornharrtell:
With #3461 now merged I think this change is even more important so I've rebased this on master.

bartvde:
LGTM

probins:
I about to suggest this change for linestrings, but it looks like this PR also handles lines. I was misled by the title referring to 'boundaries'.

Turbo87:
@bjornharrtell could you maybe have a look into #3935? it seems to be related to this change.

bjornharrtell:
I haven't been able to understand the connection between this PR and #3935, admittedly without putting any real time into it, but I'll look at rebasing this when it's in.

bjornharrtell:
Ehm, seems I forgot that this PR is actually merged for quite some time :)

Turbo87:
@bjornharrtell my proposed fix for #3935 is in #3946. feel free to review.

PR #3636, opened 2015-04-28
added altclick option for select-features example

t27:
I added an alt+click option to the select-features example. This also shows how we can listen for 2 or more conditions in the Select Interaction for alt+click or shift+click.
This is my first contribution to OL3, and I hope to make more of these. Do review and let me know of any suggestions.


sweco-sebhar:
Looks like a resonable addition to the example, but you need to make Travis happy. :) You can check the code locally, see https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md.

examples/select-features.js
```diff
@@ -43,6 +43,16 @@ var selectPointerMove = new ol.interaction.Select({
   condition: ol.events.condition.pointerMove
 });
 
+var selectAltClick = new ol.interaction.Select({
+  condition: olConditionClubber(ol.events.condition.altKeyOnly,
+      ol.events.condition.click)
```
elemoine: I'd just do:

```js
  condition: function(mapBrowserEvent) {
    return ol.events.condition.click(mapBrowserEvent) &&
        ol.events.condition.altKeyOnly(mapBrowserEvent);
  }
```

t27: I had kept the function returning a function to show the possibility of adding more conditions. But yes having it as you've shown would work fine too. I'll incorporate and update the pull request.

t27:
travis has passed, I have also incorporated @elemoine suggestions on the older diff,

elemoine:
Could you please squash your commits into one commit?

t27:
I'll do that and make a fresh pull request, hope that'd be fine?

elemoine:
You could have used push --force to update your branch and this PR.

PR #3637, opened 2015-04-28
Implement ol.renderer.Layer#forEachFeatureAtCoordinate

fredj:
And remove identical implementation in children classes.


gberaudo:
Looks sensible to me.

tschaub:
Looks good.  Please merge.

elemoine:
For the record, that function was not on the base class for two reasons: (a) it should be a noop for tile layers, (b) it has a completely different implementation for vector layers.

Personally, I don't like adding a method to a base class that is completely overridden in child classes, because I find it confusing when reading the code, and because, if I don't use Image layers in my app, I don't want to pay for code I don't use.

fredj:
I have not seen that the tile layers inherits from that base class ... will revert

PR #3639, opened 2015-04-28
Add extent support to ol.tilegrid.TileGrid

ahocevar:
The work on this pull request started with the goal of giving tile grids a notion of extents, and ended up being a refactoring that even gets rid of `ol.tilegrid.XYZ`. `wrapX` handling was drastically simplified. Tile grids can now not only have an extent, but also widths and heights for each zoom level.


changelog/upgrade-notes.md
```diff
@@ -2,6 +2,13 @@
...
+
+* If you previously created an XYZ tilegrid, you could construct an `ol.tilegrid.XYZ`. This class has been replaced by a static `ol.tilegrid.createXYZ` function which takes the same options as the previous `ol.tilegrid.XYZ` constructor, but returns an `ol.tilegrid.TileGrid` instance. The internal tile coordinate scheme for XYZ tile grids has been changed. Previously, the `y` of tile coordinates was transformed to the coordinates used by sources by calculating `-y-1`. Now, it is transformed by calculating `height-y-1`, where height is the number of rows of the tile grid at the zoom level of the tile coordinate.
+* The `widths` constructor option of `ol.tilegrid.TileGrid` and subclasses is no longer available, and it is no longer necessary to get proper wrapping at the 180° meridian. However, for `ol.tilegrid.WMTS`, there is a new option `sizes`, where each entry is an `ol.Size` with the `width` ('TileMatrixWidth' in WMTS capabilities) as first and the `height` ('TileMatrixHeight') as second entry of the array. For other tile grids, users can
+now specify an `extent` instead of `widths´. These settings are used to restrict the range of tiles that sources will request.
```
elemoine: Incorrect backtick char after widths.

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,13 @@
...
+* For `ol.source.TileWMS`, the default value of `warpX` used to be `undefined`, meaning that WMS requests with out-of-extent tile BBOXes would be sent. Now `wrapX` can only be `true` or `false`, and the new default is `true`. No application code changes should be required, but the resulting WMS requests for out-of-extent tiles will be different.
```
elemoine: I think this deserves a bit more explanations. For example, how will the WMS requests for out-of-extent tiles will be different?

elemoine: BTW, it's a pleasure to start reviewing a patch by reading release notes.

examples/wms-custom-tilegrid-512x256.js
```diff
@@ -25,7 +25,6 @@ var layers = [
     source: new ol.source.MapQuest({layer: 'sat'})
   }),
   new ol.layer.Tile({
-    extent: [-13884991, 2870341, -7455066, 6338219],
```
elemoine: Any particular reason for removing this?

externs/olx.js
```diff
@@ -6014,19 +6013,31 @@ olx.tilegrid;
...
- *     widths: (Array.<number>|undefined)}}
+ *     tileSizes: (Array.<number|ol.Size>|undefined)}}
  * @api
  */
 olx.tilegrid.TileGridOptions;
 
 
 /**
+ * Extent for the tile grid. No tiles outside this extent will be requested by
+ * {@link ol.source.Source} sources. When no `origin` or `origins` are
```
elemoine: Do you want the link to be ol.source.Tile?

src/ol/tilegrid/tilegrid.js
```diff
@@ -162,6 +201,15 @@ ol.tilegrid.TileGrid.prototype.forEachTileCoordParentTileRange =
 
 
 /**
+ * Get the extent for this tile grid, if it was configured.
+ * @return {ol.Extent} Extent.
```
elemoine: Do we want to make this `@api`?

src/ol/tilecoord.js
```diff
@@ -149,25 +149,42 @@ ol.tilecoord.toString = function(tileCoord) {
...
 };
 
 
 /**
  * @param {ol.TileCoord} tileCoord Tile coordinate.
- * @param {ol.tilegrid.TileGrid} tileGrid Tile grid.
- * @param {ol.proj.Projection} projection Projection.
- * @return {ol.TileCoord} Tile coordinate.
+ * @param {!ol.tilegrid.TileGrid} tileGrid Tile grid.
+ * @return {ol.TileCoord} Tile coordinate
```
elemoine: Dot missing at the end of the line :-)

elemoine:
I am not finished reviewing but it looks like a lot of good work to me already! 

Quick question: I remember that the functions related child/parent tiles access in the WMS tile grid were optimized versions of their counterparts in the base tile grid class. Is this something you've taken into account?

examples/wms-custom-tilegrid-512x256.js
ahocevar: Yes, to show that no tiles outside the tile grid's extent are requested - the extent has not been removed, but moved to the tile grid configuration.

src/ol/tilegrid/tilegrid.js
ahocevar: I don't see a reason for now. Should I?

elemoine: I am pretty sure the need will come eventually. But you don't have to do it as part of this PR if you are not sure it is useful.

FWIW, I know I'll need `getSizes` on `ol.TileGrid.WMTS` (for creating MapFish Print spec objects).

ahocevar:
You mean in the XYZ tile grid? Yes, it's all simple math - no loops and arrays are reused (which they weren't when ol.tilegrid.TileGrid was initially created). So there should not be much difference.

examples/wms-custom-tilegrid-512x256.js
elemoine: I thought "extent" on the layer also prevented requesting tiles outside the extent. Do you know/understand the whole story? I'm not totally clear right now.

elemoine:
> You mean in the XYZ tile grid

Yes. Android keyboard replacing words for me...

> Yes, it's all simple math - no loops and arrays are reused (which they weren't when ol.tilegrid.TileGrid was initially created). So there should not be much difference.


Ok, just wanted to know if you knew about the optimizations and if you thought about it. 

Thanks.

examples/wms-custom-tilegrid-512x256.js
ahocevar: Yes I do. Both cause tiles that intersect the extent to be loaded. But setting the layer extent clips the data at the extent boundary. Especially for WMS, where extents advertised in GetCapabilities are usually not aligned to the tile grid, there is a problem: Label placement is optimized for map/tile boundaries, so you'll get clipped labels when you clip at the extent boundary. But when you use the full tiles that intersect set the tile grid extent, you get nicely placed labels. @bartvde can provide a screenshot that shows the problem.

elemoine: I understand. We used to have an extent option on ol.source.Tile to just prevent requesting tiles outside that extent. So your patch adds this feature back, except that the extent option is on the tile grid rather the source.

ahocevar: Exactly. See https://github.com/openlayers/ol3/issues/2091#issuecomment-43783362 for the reasoning. If this turns out useful, we could later add the extent as source option and pass it to the tile grid.

src/ol/source/wmtssource.js
```diff
@@ -428,6 +401,19 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
         /urn:ogc:def:crs:(\w+):(.*:)?(\w+)$/, '$1:$3'));
   }
 
+  var projectionExtent = projection.getExtent();
+  var extent;
+  if (!goog.isNull(projectionExtent)) {
```
probins: I don't understand the logic here. It should surely always use the caps BBox whether there's a projection extent or not.

ahocevar: Right, I will add a FIXME here. Currently our WMTS capabilities format parser only reads the WGS84BoundingBox of the layer and not the bounding box in different CRSs. The reason for this logic is that the WGS84BoundingBox cannot always be transformed to the data projection, and we need to safeguard against that. A cleaner solution will be to use the advertised BBOX in the data projection.

examples/wms-custom-tilegrid-512x256.js
probins: the layer extent is surely for a different purpose: client-side limitation rather than server-side; an app might allow a user to define a smaller extent for a layer than that provided by the server

src/ol/source/wmtssource.js
probins: but the spec doesn't require layers to have a bbox in the data projection (or a wgs84 one for that matter). http://www.ngi.be/cartoweb/1.0.0/WMTSCapabilities.xml is a simple example with layers in local projection and 3857. This is the Belgian mapping agency, so extent is limited to the area around Belgium.

ahocevar: Found a better solution @probins. Fixed with 6ee0d9f92702e86e97513975ba2f12a01217eeee.

src/ol/source/wmtssource.js
```diff
@@ -428,6 +394,27 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
...
+  if (goog.isDef(wgs84BoundingBox)) {
+    var wgs84ProjectionExtent = ol.proj.get('EPSG:4326').getExtent();
+    wrapX = (wgs84BoundingBox[0] == wgs84ProjectionExtent[0] &&
+        wgs84BoundingBox[2] == wgs84ProjectionExtent[2]);
+    extent = ol.proj.transformExtent(
+        wgs84BoundingBox, 'EPSG:4326', projection);
+    var projectionExtent = projection.getExtent();
+    if (!goog.isNull(projectionExtent)) {
+      // If possible, do a sanity check on the extent - it should never be
+      // bigger than the validty extent of the projection of a matrix set.
```
probins: validity

src/ol/source/wmtssource.js
probins: yes, I've not tested it, but that should do the trick

bartvde:
Confirmed that this fixes the MapStory use case when the tileGrid is configured with extent and resolutions from the WMS-C GetCapabilities document. See also: https://github.com/MapStory/story-tools/pull/208

ahocevar:
Thanks @bartvde for testing.

I addressed all your comments @elemoine. Anything else? Anyone else wants to take a look? Or is this good to merge?

elemoine:
I'll finish reviewing on Monday if ok by you.

ahocevar:
:thumbsup:

elemoine:
Reviewing again now…

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,13 @@
 
 ### v3.5.0
 
+#### `ol.tilegrid` changes
+
+* If you previously created an XYZ tilegrid, you could construct an `ol.tilegrid.XYZ`. This class has been replaced by a static `ol.tilegrid.createXYZ` function which takes the same options as the previous `ol.tilegrid.XYZ` constructor, but returns an `ol.tilegrid.TileGrid` instance. The internal tile coordinate scheme for XYZ tile grids has been changed. Previously, the `y` of tile coordinates was transformed to the coordinates used by sources by calculating `-y-1`. Now, it is transformed by calculating `height-y-1`, where height is the number of rows of the tile grid at the zoom level of the tile coordinate.
```
elemoine: I find the first sentence a bit confusing. Did you mean "If you previously created an XYZ source" (instead of tilegrid)?

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,13 @@
...
+* The `widths` constructor option of `ol.tilegrid.TileGrid` and subclasses is no longer available, and it is no longer necessary to get proper wrapping at the 180° meridian. However, for `ol.tilegrid.WMTS`, there is a new option `sizes`, where each entry is an `ol.Size` with the `width` ('TileMatrixWidth' in WMTS capabilities) as first and the `height` ('TileMatrixHeight') as second entry of the array. For other tile grids, users can
+now specify an `extent` instead of `widths`. These settings are used to restrict the range of tiles that sources will request.
+* For `ol.source.TileWMS`, the default value of `warpX` used to be `undefined`, meaning that WMS requests with out-of-extent tile BBOXes would be sent. Now `wrapX` can only be `true` or `false`, and the new default is `true`. No application code changes should be required, but the resulting WMS requests for out-of-extent tiles will uses BBOXes that are shifted to real world coordinates, and no more out-of-extent BBOXes.
```
elemoine: "will uses" -> "will use"

changelog/upgrade-notes.md
ahocevar: No, I really meant tilegrid - if you previously created an XYZ source, then nothing changes. But maybe I should create a new bullet point before "The internal tile coordinate scheme for XYZ tile grids has been changed.", because that refers to XYZ sources.

ahocevar: :+1:

externs/olx.js
```diff
@@ -6079,32 +6102,32 @@ olx.tilegrid.TileGridOptions.prototype.tileSizes;
...
 /**
- * Origin.
+ * Extent for the tile grid. No tiles outside this extent will be requested by
+ * {@link ol.source.Source} sources. When no `origin` or `origins` are
+ * configured, the `origin` will be calculated from the extent.
+ * When no `sizes` are configured, they will be calculated from the extent.
+ * @type {ol.Extent|undefined}
+ * @api
+ */
+olx.tilegrid.TileGridOptions.prototype.extent;
```
elemoine: It should be `olx.tilegride.WMTSOptions.prototype.extent` here.

externs/olx.js
```diff
@@ -6079,32 +6102,32 @@ olx.tilegrid.TileGridOptions.prototype.tileSizes;
...
+ *     tileSizes: (Array.<number|ol.Size>|undefined)}}
  * @api
  */
 olx.tilegrid.WMTSOptions;
 
 
 /**
- * Origin.
+ * Extent for the tile grid. No tiles outside this extent will be requested by
+ * {@link ol.source.Source} sources. When no `origin` or `origins` are
```
elemoine: You probably want to change the link to `ol.source.WMTS` here.

externs/olx.js
```diff
@@ -6140,6 +6164,18 @@ olx.tilegrid.WMTSOptions.prototype.matrixIds;
...
 /**
+ * Number of tile rows and columns of the grid for each zoom level. The values
+ * here are the `TileMatrixWidth` and `TileMatrixHeight` advertised in the
+ * GetCapabilities response of the WMTS, and define the grid's extent together
+ * with the `origin`. An `extent` can be configured in addition, and will
+ * further limit the extent for which tile requests are made by sources.
+ * @type {Array.<ol.Size>|undefined}
+ * @api
+ */
+olx.tilegrid.TileGridOptions.prototype.sizes;
```
elemoine: Again `olx.tilegride.WMTSOptions.prototype.sizes`.

src/ol/tilegrid/tilegrid.js
```diff
@@ -521,36 +534,47 @@ ol.tilegrid.getForProjection = function(projection) {
...
 
 
 /**
+ * Creates a tile grid with a standard XYZ tiling scheme.
+ * @param {olx.tilegrid.XYZOptions=} opt_options Tile grid options.
+ * @return {ol.tilegrid.TileGrid} Tile grid instance.
+ * @api
+ */
+ol.tilegrid.createXYZ = function(opt_options) {
+  var options = {};
```
elemoine: It would be good to type this.

src/ol/tilegrid/tilegrid.js
```diff
@@ -608,3 +632,36 @@ ol.tilegrid.extentFromProjection = function(projection) {
...
+          return null;
+        }
+        var z = tileCoord[0];
+        var fullTileRange = tileGrid.getFullTileRange(z);
+        var height;
+        if (goog.isNull(fullTileRange) || fullTileRange.minY < 0) {
+          height = 0;
+        } else {
+          height = fullTileRange.getHeight();
+        }
```
elemoine: You could use the ternary notation here:

```js
  var height = goog.isNull(fullTileRange) || fullTileRange.minY < 0 ? 0 : fullTileRange.getHeight();
```

elemoine:
I've just added a few other pretty minor comments. Otherwise looks good to me. +1 for merging when the remaining things are addressed. Thanks a lot for the work on this, I think this is great!

ahocevar:
Thanks a lot for the review @elemoine.

tschaub:
It looks like b05193fa454a963b33e6c5965181abde9b45b7fa breaks TileJSON sources (see #3721).

ahocevar:
I'm working on a fix now. Thanks for spotting this.

tschaub:
Thanks for the fix.  I'm trying not to think about what it means that it wasn't spotted for over two weeks :grimacing:

elemoine:
It may mean that people use ol3 releases rather than master. It may also mean that ol3 users do not use TileJSON too much. And/or it may mean that people are able to live without looking at the ol3 examples for two weeks. Did you have something else mind? :)

marcjansen:
On travis we actually check the examples, but only error out on compile problems, right? Perhaps we want to thoroughly test them, e.g. with rendering tests or with phantom. 

Perhaps that would have caught this error earlier.

PR #3645, opened 2015-04-29
Fix test-coverage.js script

### LINKED ISSUE #3643: `tasks/test-coverage.js` wrongly detects the outputfilename to store instrumented sources
`tasks/test-coverage.js` has: 

```javascript
var outfile = file.replace(/\/src\//, '/src-instrumented/');
```

which is from https://github.com/openlayers/ol3/blob/master/tasks/test-coverage.js#L142 

We noticed this actually leads to errors in #3640#issuecomment-97367282

elemoine:
Fix test-coverage by only replacing the last occurence of `/src/` when deriving the instrumented file path from the original file path.

Fixes #3643.

There may be a better way to do this, but that's the one I've been able to come up with.

### DIFF
--- a/tasks/test-coverage.js
+++ b/tasks/test-coverage.js
@@ -139,7 +139,15 @@ var foundAllJavaScriptSourceFiles = function(err, files) {
   files.forEach(function(file) {
     cnt++;
     var content = fs.readFileSync(file, 'utf-8');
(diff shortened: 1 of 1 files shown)


marcjansen:
Plus one and many thanjs for fixing this. There may be cornercases, but this script needs a rewrite in any case.

Please merge.

PR #3646, opened 2015-04-29
Use graceful-fs in place of fs

elemoine:
Change generate-requires.js to use the graceful-fs module instead of fs. This is to avoid EMFILE errors on OSX because the maximum number of file descriptors have been reached.

@bartvde, could you please test this change?


bartvde:
thanks @elemoine this does the trick

PR #3007, opened 2014-12-05
overview map uses projection parameter or base projection before defaults

relet:
closes #2998 as per #3001, but the ol.Map object is only initialized when setMap makes the base map for this control known.

PR #3647, opened 2015-04-29
fix projection and resolutions in overview map

jhwegener:
This PR tries to add stuff of #3007 and add resolutions as an second new parameter. Different from #3007, I'm not trying to get the projection of the main map, since it seams not working for me. But setting it in the parameters of overview map is enough I think.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1152,7 +1152,9 @@ olx.control.MousePositionOptions.prototype.undefinedHTML;
  *     collapsible: (boolean|undefined),
  *     label: (string|Node|undefined),
  *     layers: (Array.<ol.layer.Layer>|ol.Collection|undefined),
+ *     projection: (ol.proj.ProjectionLike|undefined),
  *     render: (function(ol.MapEvent)|undefined),
+ *     resolutions: (Array.<number>|undefined),
  *     target: (Element|undefined),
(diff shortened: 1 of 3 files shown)


jhwegener:
Noticed failed Travis CI build. Try to get dev environment running on my local machine.

Sverre-Iversen:
<h4>Here is a tricky workaround to fix projection in overview map - need a fix for resolutions</h4>

//Projection setting that does not work
var oProj = new ol.proj.Projection({...});
	ovControl = new ol.control.OverviewMap({
	className: 'ol-overviewmap',
	projection:  oProj, // <- does not work
	layers:[...],
});

//workaround to make it work
//Get map
//ovMap = ovControl.getMap(),// <-does not work. getMap() exists but return null
//Retrieve map object by traverseing ovControl & checking for instanceof ol.Map
var ovMap = ovControl[util.findKeyByClass(ovControl,ol.Map)],// <- custom function
//Get and set the projection in view
ovView = ovMap.getView(),
//No method for setting the projection
//Retrieve projection object by traverseing ovView & checking for instanceof ol.proj.Projection
ovProjKey = util.findKeyByClass(ovView,ol.proj.Projection);// <- custom function
//finally setting the projection in view
ovView[ovProjKey] = oProj;

PR #3655, opened 2015-05-04
Make mouse wheel interaction runtime configurable

Antreasgr:
Added maxdelta, timeout runtime configurable parameters in
ol.interaction.MouseWheelZoom class according to
http://stackoverflow.com/questions/28888047/mouse-wheel-scroll-limitation-in-openlayers-3


externs/olx.js
```diff
@@ -2581,6 +2581,21 @@ olx.interaction.MouseWheelZoomOptions.prototype.duration;
...
+ */
+olx.interaction.MouseWheelZoomOptions.prototype.maxdelta;
+
+/**
+ * Timeout duration. Default is `80`.
+ * @type {number}
+ * @api
+ */
+olx.interaction.MouseWheelZoomOptions.prototype.timeout;
+
```
bartvde: you need to add these new options here as well:

```
/**
 * @typedef {{duration: (number|undefined)}}
 * @api
 */
olx.interaction.MouseWheelZoomOptions;
```

ahocevar:
@bartvde Are you going to continue reviewing this, or should someone else take over?

bartvde:
This probably needs updating not use goo.match.clamp right? https://github.com/openlayers/ol3/pull/4175

waissbluth:
Is anybody on this? This feature would be really useful.

rixlabs:
+1

tschaub:
I like the idea of making this configurable.  But I think it is more important that we have better mouse wheel interaction by default.  Based on the [StackOverflow question](http://stackoverflow.com/questions/28888047/mouse-wheel-scroll-limitation-in-openlayers-3), I assume we'll have differing opinions of "better" though.  In my opinion, with a trackpad on a MacBook, the map zooms in/out way more than I want it to every time.  Based on experience watching people try to use OpenLayers for the first time, I think it is common to be surprised by the excessive zooming.

rixlabs:
I wish I could configure the interaction to avoid some eccessive call to my map services.
I noticed the problem cause I have poor arcgis test environment and with default config It happens that OL make several call for a mousewheel zoom that cause(my server is really under size) some timeout problems form time to time. To solve it I had to change the 80ms time to 250 going into the code (ol.js)

tschaub:

PR #3658, opened 2015-05-05
Release v3.5.0

tschaub:
This adds a changelog and bumps the package version.

### DIFF
--- a/changelog/v3.5.0.md
+++ b/changelog/v3.5.0.md
@@ -1,0 +1,289 @@
+# v3.5.0
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


probins:
what's the process for updating the cdnjs version? The hosted examples source display is now using https://cdnjs.cloudflare.com/ajax/libs/ol3/3.5.0/ol.js but this doesn't exist. So created fiddles don't work either.

And how often is http://openlayers.org/en/master/build/ol.js regenerated? This is currently using a commit from last week, before 3.5

ahocevar:
It is updated automatically by cdnjs, once a day I think. So examples should be working shortly.

ahocevar:
The master examples are updated manually, without a regular schedule.

probins:
cdnjs now updated - fiddles now work again, thanks

Issue #2827, opened 2014-10-13
Double Buffering enhancement for dynamic raster layers like WMS


shaunakv1:
Hello, 

I have a Tiled WMS layer on which I change the SLD on the fly using a updateParams button. The user interface control for doing this is a slider.  Currently when the slider is moved the there is a flicker while the refreshed WMS layer loads. Is there a way to avoid this and use double buffering? So that Openlayers wait to remove the old layer before the new one is ready and then swaps , so the flicker is avoided. 

I have asked this question on google groups 

https://groups.google.com/forum/#!topic/ol3-dev/j1GzQHgwzgI

But looks like no solution yet. 

Here's the example of the flicker I am talking about.  Please try the slider to change the WMS layer:

http://shaunakv1.github.io/ol3_wms_experiments/

PR #3660, opened 2015-05-05
Smooth rendering of parameter changes

thomasmoelhave:
Rationale:
---------------

There are many use-cases where one or more frequently changing parameter/dimension for a WMTS source are used. One such example is issue #2827 and in our SCALGO Live products we also have several different  types of WMTS layers where a continuous dimension made available in the UI through a slider. An example of this is the [sea-level flooding] (http://scalgo.com/live/global?center=-10546029.2%2C3387094.1&zoom=5&tool=zoom&lrs=basic%2Cglobal%2Fhydrosheds%3Adem%2Cglobal%2Fhydrosheds%3Asea-levels&flooding=10) layer, and there is a similar interaction for [flow accumulation](http://scalgo.com/live/global?center=1717825.6%2C5738245.9&zoom=5&tool=zoom&lrs=basic%2Cglobal%2Fhydrosheds%3Adem%2Cglobal%2Fhydrosheds%3Aedgeflow&edge_flow=3509720000). In the former, the sea-level rise is set via a dimension/parameter to the GetTile request, and when the user drags the slider around, the sea level changes (the WMTS server delivers tiles with the appropriate sea level). In stock openlayers, this is not well supported as no tiles are rendered until the new tiles are ready, resulting in heavy flickering. 

This non-trivial patch changes the image tile renderer to keep rendering old tiles while the new tiles are being loaded, this significantly improves the use cases where such rapid changes in a dimension are being performed. We have created an [example](http://scalgo.com/stuff/ol3/build/examples/smooth-params.html), supplementing the one in issue #2827, that clearly illustrates the use case. You can see the regular WMTS mode on the left, and the new on the right. Use the sea-level slider to change the sea level and notice how the regular WMTS layer flickers significantly, especially when you slowly drag the slider (which is a very common use case). This patch does not change the semantics of existing WMTS layers, the smooth mode has to be explicitly requested by designating which dimensions  should be treated this way.  We have already been running this patch in production for several months (and it's used in the links above) with no issues for both regular WMTS layers and WMTS layers with a smooth dimension.

Details:
---------------
First, the patch tells WTMS sources to ignore certain dimensions when building tile cache keys. That way, when these parameters change, the existing tiles are not immediately discarded (or ignored) from the cache. Instead, the old tiles are still rendered. Without any other changes to the code, just this change would cause the ignored dimensions to have no effect on rendering at all.

When the ignored dimensions do change, we added a new method that, when called, explicitly reloads all the currently-visible tiles, and purges all the not visible tiles.

Tile reloading for this method uses a mechanism we called "out-of-band" in the code. Essentially, we turned each tile in the cache into a linked list of tiles. The head of the list is the current tile, and the second node (if any) is a tile waiting to be loaded. If there's a third node in the linked list (which is less common), that means the second node tile is in a loading state, but yet another tile has been requested. If a linked list gets flooded with load requests (which often happens for e.g. slider inputs), the linked list builder drops tile nodes that it knows will never be rendered. This load shedding generally keeps the length of the linked list to at most three.

This tile loading mechanism uses the ol3 tile load queue, so it still takes advantage of all the nice work scheduling for image loading that the ol3 engine uses.

We also made a small change to the image tile renderer to check the state of the linked list for each tile to decide which one to render for that frame.

The patch passes the Travis CI test and is rebased on top of current ol3 master. We have split the patch into one commit for the base functionality (I am not sure it's necessary to further subdivide that commit) and a separate commit for the example. I have filled out the corporate contributor license agreement and can send that in when it becomes relevant.

We appreciate your comments.

### DIFF
--- a/examples/smooth-params.js
+++ b/examples/smooth-params.js
@@ -1,0 +1,133 @@
+goog.require('ol.Attribution');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.control');
+goog.require('ol.control.Attribution');
+goog.require('ol.control.Zoom');
+goog.require('ol.extent');
+goog.require('ol.layer.Tile');
+goog.require('ol.proj');
+goog.require('ol.source.OSM');
+goog.require('ol.source.WMTS');
+goog.require('ol.tilegrid.WMTS');
+
+
+// create the WMTS tile grid in the google projection
+var projection = ol.proj.get('EPSG:3857');
+var tileSizePixels = 256;
+var tileSizeMtrs = ol.extent.getWidth(projection.getExtent()) / tileSizePixels;
+var matrixIds = [];
+var resolutions = [];
+var attributions = [
+  new ol.Attribution({
+    html: '<a href="http://scalgo.com">SCALGO</a>'
+  }),
+  new ol.Attribution({
+    html: '<a href="http://www.cgiar-csi.org/data/' +
+        'srtm-90m-digital-elevation-database-v4-1">CGIAR-CSI SRTM</a>'
+  })
+];
+
+for (var i = 0; i <= 14; i++) {
+  matrixIds[i] = i;
+  resolutions[i] = tileSizeMtrs / Math.pow(2, i);
+}
+var tileGrid = new ol.tilegrid.WMTS({
+  origin: ol.extent.getTopLeft(projection.getExtent()),
+  resolutions: resolutions,
+  matrixIds: matrixIds
+});
+
+var scalgo_token = 'CC5BF28A7D96B320C7DFBFD1236B5BEB';
+
+// make the special WMTS layer
+var layerSmooth = new ol.layer.Tile({
+  opacity: 0.5,
+  source: new ol.source.WMTS({
+    url: 'http://ts2.scalgo.com/global/wmts?token=' + scalgo_token,
+    layer: 'hydrosheds:sea-levels',
+    format: 'image/png',
+    matrixSet: 'EPSG:3857',
+    attributions: attributions,
+    tileGrid: tileGrid,
+    style: 'default',
+    dimensions: {
+      threshold: 100
+    },
+    keyPrefixIgnoredDimensions: ['threshold']
+  })
+});
+
+var layer = new ol.layer.Tile({
+  opacity: 0.5,
+  source: new ol.source.WMTS({
+    url: 'http://ts2.scalgo.com/global/wmts?token=' + scalgo_token,
+    layer: 'hydrosheds:sea-levels',
+    format: 'image/png',
+    matrixSet: 'EPSG:3857',
+    attributions: attributions,
+    tileGrid: tileGrid,
+    style: 'default',
+    dimensions: {
+      threshold: 100
+    }
+  })
+});
+
+var mapSmooth = new ol.Map({
+  target: 'map_smooth',
+  view: new ol.View({
+    projection: projection,
+    center: [-3052589, 3541786],
+    zoom: 3
+  }),
+  controls: [
+    new ol.control.Zoom(),
+    new ol.control.Attribution()
+  ],
+  layers: [
+    new ol.layer.Tile({
+      source: new ol.source.OSM()
+    }),
+    layerSmooth
+  ]
+});
+
+var map = new ol.Map({
+  target: 'map',
+  view: mapSmooth.getView(),
+  controls: [
+    new ol.control.Zoom(),
+    new ol.control.Attribution()
+  ],
+  layers: [
+    new ol.layer.Tile({
+      source: new ol.source.OSM()
+    }),
+    layer
+  ]
+});
+
+var setLayerParam = function(map, layer, sliderVal, isSmooth) {
+  // update the layer dimensions
+  // normally this would result in e tile reload,
+  // but we're ignoring the 'threshold' dimension in this example
+  layer.getSource().updateDimensions({
+    threshold: sliderVal
+  });
+
+
+  if (isSmooth) {
+    // explicitly load the tiles for the new threshold
+    layer.getSource().reloadVisibleTilesOutOfBand(map);
+  }
+  document.getElementById('theinfo').innerHTML = sliderVal + ' meters.';
+};
+
+setLayerParam(map, layer, 10, false);
+setLayerParam(mapSmooth, layerSmooth, 10, true);
+
+document.getElementById('slider').addEventListener('input', function() {
+  setLayerParam(map, layer, this.value, false);
+  setLayerParam(mapSmooth, layerSmooth, this.value, true);
+});
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5193,7 +5193,8 @@ olx.source.VectorOptions.prototype.wrapX;
  *     tileClass: (function(new: ol.ImageTile, ol.TileCoord,
  *                          ol.TileState, string, ?string,
  *                          ol.TileLoadFunctionType)|undefined),
- *     wrapX: (boolean|undefined)}}
+ *     wrapX: (boolean|undefined),
+ *     keyPrefixIgnoredDimensions: (Array.<string>|undefined)}}
  * @api
  */
 olx.source.WMTSOptions;
@@ -5360,6 +5361,14 @@ olx.source.WMTSOptions.prototype.urls;
  * @api
  */
 olx.source.WMTSOptions.prototype.wrapX;
+
+
+/**
+ * Ignored dimensions for prefix keys
+ * @type {Array.<string>|undefined}
+ * @api
+ */
+olx.source.WMTSOptions.prototype.keyPrefixIgnoredDimensions;
 
 
 /**
--- a/src/ol/imagetile.js
+++ b/src/ol/imagetile.js
@@ -60,8 +60,108 @@ ol.ImageTile = function(tileCoord, state, src, crossOrigin, tileLoadFunction) {
    */
   this.tileLoadFunction_ = tileLoadFunction;
 
+  /**
+   * @private
+   * @type {number}
+   */
+  this.sequenceNumber_ = 0;
+
+
+  /**
+   * @private
+   * @type {ol.ImageTile}
+   */
+  this.outOfBandTile_ = null;
+
 };
 goog.inherits(ol.ImageTile, ol.Tile);
+
+
+/**
+ * set the url for the out-of-band tile
+ *
+ * @param {string} url The url to load out-of-band.
+ * @return {ol.ImageTile}
+ */
+ol.ImageTile.prototype.setOutOfBandUrl = function(url) {
+  if (this.outOfBandTile_ === null) {
+
+    // create a new tile
+    this.outOfBandTile_ = this.createOutOfBandTile_(url);
+    this.assertSequence_();
+    return this.outOfBandTile_;
+
+  } else if (this.outOfBandTile_.state == ol.TileState.IDLE) {
+    goog.asserts.assert(this.outOfBandTile_.outOfBandTile_ === null);
+
+    // replace the old tile, and cause it to be dropped from the tile queue
+    var newTile = this.createOutOfBandTile_(url);
+    // this sets the tile load priority to DROP
+    this.outOfBandTile_.tileCoord = [-1, 0, 0];
+    this.outOfBandTile_ = newTile;
+    this.assertSequence_();
+    return this.outOfBandTile_;
+
+  } else if (this.outOfBandTile_.state == ol.TileState.LOADING ||
+      this.outOfBandTile_.state == ol.TileState.LOADED ||
+      this.outOfBandTile_.state == ol.TileState.ERROR) {
+    return this.outOfBandTile_.setOutOfBandUrl(url);
+  } else {
+    throw "don't know what to do when outOfBandTile is in state" +
+        this.outOfBandTile_.state;
+  }
+};
+
+
+/**
+ * create the out-of-band tile
+ *
+ * @param {string} url The url to load out of band
+ * @return {ol.ImageTile}
+ * @private
+ */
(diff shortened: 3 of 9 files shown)


thomasmoelhave:
I am not entirely sure what the `expect.js called gets moody all too quickly` from Travis CI is about, it doesn't happen on my system. I am not sure it has anything to do with this patch?

tschaub:
This is a very well put together contribution and much appreciated (the code is generally very clean and your description is great).  We've just had a short discussion about the nature of this change on our weekly dev call.  We want to make sure the change is appropriately general (nothing tied to WMTS) and will handle multiple use cases.  I imagine we'll have some back and forth as people find time to review and potentially suggest some alternatives.

So, nothing more substantive from me right now.  But I wanted to at least express appreciation and say that the results look great.

examples/smooth-params.js
```diff
@@ -0,0 +1,133 @@
...
+  // normally this would result in e tile reload,
+  // but we're ignoring the 'threshold' dimension in this example
+  layer.getSource().updateDimensions({
+    threshold: sliderVal
+  });
+
+
+  if (isSmooth) {
+    // explicitly load the tiles for the new threshold
+    layer.getSource().reloadVisibleTilesOutOfBand(map);
```
tschaub: My preference would be to see something like a layer option that makes this work without an explicit call.  Since layers control other aspects about rendering, it feels like a layer option would be appropriate - but I haven't dug deep enough to see how that would work out.

externs/olx.js
```diff
@@ -5231,6 +5232,14 @@ olx.source.WMTSOptions.prototype.wrapX;
 
 
 /**
+ * Ignored dimensions for prefix keys
+ * @type {Array.<string>|undefined}
+ * @api
+ */
+olx.source.WMTSOptions.prototype.keyPrefixIgnoredDimensions;
```
tschaub: Are all of `key`, `prefix`, and `dimension` really meaningful in this name?  Is `ignoredDimensions` the wrong name (or too broad)?

I don't have a better suggestion, just saying that the name is potentially confusing/meaningless.

src/ol/imagetile.js
```diff
@@ -60,11 +60,111 @@ ol.ImageTile = function(tileCoord, state, src, crossOrigin, tileLoadFunction) {
...
+   * @type {ol.ImageTile}
+   */
+  this.outOfBandTile_ = null;
+
 };
 goog.inherits(ol.ImageTile, ol.Tile);
 
 
 /**
+ * set the url for the out-of-band tile
```
tschaub: Please use complete sentences (with capitalization and punctuation) for docs.  Same comment applies throughout.

src/ol/source/tileimagesource.js
```diff
@@ -191,3 +198,139 @@ ol.source.TileImage.prototype.useTile = function(z, x, y) {
...
+  // this lets multiple load requests in rapid succession overwrite each other
+  // it means we end up skipping some load requests entirely if they were
+  // immediately succeeded by another request for the same tile
+  // if this function is called by a UI slider (which can generate a huge
+  // number of requests in a short time), this delay can offer a significant
+  // speedup
+  var afterDelay = function() {
+    this.executeLoadRequests(map, z, tileRange, viewState.projection);
+  };
+  window.setTimeout(goog.bind(afterDelay, this), 100);
```
tschaub: Instead of simply delaying the call, do you instead want to throttle or debounce the calls?

src/ol/source/tileimagesource.js
```diff
@@ -191,3 +198,139 @@ ol.source.TileImage.prototype.useTile = function(z, x, y) {
...
+
+/**
+ * run the queued out-of-band load requests
+ *
+ * @param {ol.Map} map A map instance.
+ * @param {number} z The zoom.
+ * @param {ol.TileRange} tileRange A range of tiles.
+ * @param {ol.proj.Projection} projection The map view projection.
+ */
+ol.source.TileImage.prototype.executeLoadRequests =
```
tschaub: It looks like this could be `@private`.

src/ol/imagetile.js
```diff
@@ -60,11 +60,111 @@ ol.ImageTile = function(tileCoord, state, src, crossOrigin, tileLoadFunction) {
...
+    this.assertSequence_();
+    return this.outOfBandTile_;
+
+  } else if (this.outOfBandTile_.state == ol.TileState.LOADING ||
+      this.outOfBandTile_.state == ol.TileState.LOADED ||
+      this.outOfBandTile_.state == ol.TileState.ERROR) {
+    return this.outOfBandTile_.setOutOfBandUrl(url);
+  } else {
+    throw "don't know what to do when outOfBandTile is in state" +
+        this.outOfBandTile_.state;
```
tschaub: Please always throw `Error`s instead of strings.

Throughout the majority of the library, we use `goog.asserts.fail` instead of `throw` (exceptions are`wktformst.js` and `rbush.js`).  I don't have a strong opinion, but it would be more consistent to `goog.asserts.fail`.  The benefit is that error strings are not included in production code.  The drawback is that you get unexpected/silent failures in production.

thomasmoelhave:
@tschaub thanks for your review, I'm happy you appreciate the contribution. 

It certainly makes sense to get more eyeballs on the patch. The WMTS-specific changes are limited to simply managing the dimensions, and I am hopeful the underlying design will work for a larger set of sources. We have not worked much with WMS layers, someone else might want to confirm that the idea carries over.

Thank you for your line-by-line comments as well, I guess we should hold off with incorporating the detailed comments until there's consensus on the fundamental design?

elemoine:
I am interested in reviewing this, and eventually merge this work, possibly with changes. I'll spend some time on it tomorrow. Thanks for the contribution.

thomasmoelhave:
Excellent! I have a rebased it to master (as of a few days ago) in a branch in my own repo, I can update the pull request with that version, that might make things a little more convenient for you - I had to change a few things to adapt to the tilegrid coordinate scheme changes.

elemoine:
Yes please!

examples/smooth-params.js
elemoine: I agree that there should be an option somewhere (possibly on the tile source) and that the user should not need to call a `reloadVisibleTilesOutOfBand` function.

src/ol/source/tileimagesource.js
```diff
@@ -191,3 +198,139 @@ ol.source.TileImage.prototype.useTile = function(z, x, y) {
...
+            outOfBandTile.updateOutOfBandTile(cachedTile);
+            // replace the old tile
+            this.tileCache.replace(tileCoordKey, outOfBandTile);
+          }
+        } else {
+          // we've never seen this tile before, just add it to the cache
+          this.tileCache.set(tileCoordKey, outOfBandTile);
+        }
+        // make sure the map draws a frame after the tile is loaded
+        map.render();
```
elemoine: Tile sources should not know about the map. So I think this deserves some re-design.

thomasmoelhave:
I have rebased to current master and updated the pull request. Thanks for the review!.

src/ol/source/tileimagesource.js
tschaub: The source should be able to emit a `change` event instead.

elemoine:
Thanks again for your code and the details provided in the issue description.

I've been thinking about the whole problem, and I've come up the same design as you: organize tiles in a linked list.

Here are more comments:

The application code shouldn't need to call `reloadVisibleTilesOutOfBand` to load tiles. The loading of tiles should be done by the library.

Your patch adds a lot of code to `ol/imagetile.js` and `ol/source/tileimagesource.js`, and I've wondering if we could do simpler.

So I've been experimenting in my own branch ([dynamic-params](https://github.com/openlayers/ol3/compare/master...elemoine:dynamic-params)). My code uses a similar design to yours, but it includes fewer changes to both `ol/imagetile.js` and `ol/source/tileimagesource.js`. The new logic is basically in the `ol.source.TileImage#getTile` function.

But my code is not as complete as yours.

In particular, the length of tile linked lists is not handled. So linked lists may end up with lots of tiles, based on the number of possible dimension values. That is a problem, and, at this point, I don't quite understand how you addressed that issue.

Another problem is the disposal of tiles in the cache. Only the head of the linked list is a candidate for expiration. I am not sure your code handles that problem either. But in your case the problem is mitigated by the fact that you control the length of linked lists.

So, to conclude this comment, I'd like to come up with a patch that is simpler than yours but more complete that mine. It would be good if you could have a look at my patch, to tell me if it makes sense to you.

thomasmoelhave:
Excellent! Your new implementation is cleaner and smaller than ours (which we were also trying to make mostly additive for easier porting to new releases), and although our patch worked - I am sure we're better off with a patch by someone more familiar with the ol infrastructure. I am happy you got rid of the explicit `reloadVisibleTilesOutOfBand` call, that always seemed iffy.

I tested your patch out on my system and it appears to work reasonably well for the example, crucially, there is no flickering and tiles are simply updated with newer data as it becomes available. However, it does seem to have a significant responsiveness regression compared to our patch. When you drag the slider multiple times back and forth rather quickly, the map rendering lacks behind the current value of the slider, and it takes a while to "catch up". This is probably caused by the length of the linked list getting too long as you also point out, assuming the network and not rendering is the bottleneck here.

As for pruning the request, the idea is that if we know a request would never get loaded we can throw it away. I.e. if you have 2 requests queued up, and you're loading no 1, but not no 2, and then request no 3 comes in, you can just drop no 2 without ever loading it and skip straight to no 3.  This happens e.g.
https://github.com/openlayers/ol3/compare/master...thomasmoelhave:smoothparams_clean#diff-d4d79f1d362fadde99dd6ee2af0a76d8R97
and the hack in 
https://github.com/openlayers/ol3/compare/master...thomasmoelhave:smoothparams_clean#diff-c4373b774a597750cb202cf726c7303fR240
This has worked out well in practice in our non-trivial application, but a more sophisticated pruning algorithm might be desired down the line for use cases we maybe didn't consider - at the very least we probably should get rid of the timeout thing. We have also not tested this extensively if the dynamic parameter space has more than one dimension.

Good point about tile disposal from the cache, I'm honestly not entirely sure what this implication of this is in practice, but I don't think we explicitly handle that problem, as you also pointed out.
Causes many requests to get dropped, although doing it with a timeout is not ideal as @tschaub also pointed out.

In conclusion, I think your patch is architecturally more sound, but some form of load shedding probably needs to be put in place.

elemoine:
Thanks for your comments. Very useful. Starting Friday I am on vacation for three weeks. I'll see if I can spend more time on this work before my vacation. Otherwise it will have to wait until I return. I you want to help feel free to suggest changes/commits based on my branch.

thomasmoelhave:
I am glad you found the comments useful. I am also going on vacation starting Friday, so I am not sure much will happen until a few weeks from now :). 
Have a nice summer!

thomasmoelhave:
Hi @elemoine, I hope you had  a nice vacation.

I did some more performance testing of you branch, and looked on the length of the linked list here:
https://github.com/openlayers/ol3/compare/master...elemoine:dynamic-params#diff-c4373b774a597750cb202cf726c7303fR133

It does not seem to behave the way I imagined, in particular it appears the list is never cleared or reduced. After dragging the slider around a  little bit, I easily got a list of length > 70 for every tile in my viewport. Is that the intended behavior?

elemoine:
> It does not seem to behave the way I imagined, in particular it appears the list is never cleared or reduced. After dragging the slider around a little bit, I easily got a list of length > 70 for every tile in my viewport. Is that the intended behavior?

No. This is what I wrote in [my previous comment](https://github.com/openlayers/ol3/pull/3660#issuecomment-118355174):

> In particular, the length of tile linked lists is not handled. So linked lists may end up with lots of tiles, based on the number of possible dimension values. That is a problem, and, at this point, I don't quite understand how you addressed that issue.

thomasmoelhave:
Oh right, sorry about that. My memory was foggy after the vacation and I should have re-read your notes. I'll take a closer look at this later this week.

thomasmoelhave:
@elemoine, I have added a number of notes to the [commit](https://github.com/openlayers/ol3/commit/cbc8e7e57a037b8d836a240ed6f9a796058c21be). I hope they help you understand how we managed the length of the lists. I am not sure how to incorporate these ideas into your patch, but maybe you can take a look? If I didn't actually address your question, let me know and I'll try again :).

fredj:
@elemoine any news on this pull request?

ahocevar:
@elemoine Please take another look at this pull request. @thomasmoelhave deserves some feedback I think. If you don't have time, at least summarize your thoughts, so someone else can take over from here.

elemoine:
I am sure about the lack of activity on this. I do want to get back to this, but I haven't found the time up to now.

elemoine:
@thomasmoelhave, I've rebased my [dynamic-params branch](https://github.com/openlayers/ol3/compare/master...elemoine:dynamic-params) onto the current master. I've also made changes based on your comments.

With these changes the list length never exceeds 2. I hope this will make address the performance problems that you observed. I still need to make sure that my code is compatible with the "tile reprojection" code (I don't think it is at the moment).

Do you have time to test my branch again?

thomasmoelhave:
That is excellent! I am very interested in testing your code, but I might not be able to find the time for it until the week of the 16th. Does that work for you?

elemoine:
> That is excellent! I am very interested in testing your code, but I might not be able to find the time for it until the week of the 16th. Does that work for you?

I created a new PR based on my branch. I'll try to spend more time on it next week.

fredj:
Fixed with #4389

skalp:
Hello,
not sure if I ask in the right place, but do you plan to support this feature for TileWMS source on updateParams calls ?
I use TileWMS layer with time parameter, and it would be great if the smooth transition could be used in this case too.
Thanks

bartvde:
@skalp See: https://github.com/openlayers/ol3/pull/4527

skalp:
Great ! Thanks !

PR #3661, opened 2015-05-05
Provide the ability to set and get the layer in stamen sources

SDaron:
Add a setLayer and getLayer function to stamen sources.

### DIFF
--- a/src/ol/source/stamensource.js
+++ b/src/ol/source/stamensource.js
@@ -88,33 +88,75 @@ ol.source.StamenProviderConfig = {
  */
 ol.source.Stamen = function(options) {
 
-  var i = options.layer.indexOf('-');
-  var provider = i == -1 ? options.layer : options.layer.slice(0, i);
-  goog.asserts.assert(provider in ol.source.StamenProviderConfig,
-      'known provider configured');
-  var providerConfig = ol.source.StamenProviderConfig[provider];
+  /**
+   * url
+   * @type {string}
+   * @private
+   */
 
-  goog.asserts.assert(options.layer in ol.source.StamenLayerConfig,
-      'known layer configured');
-  var layerConfig = ol.source.StamenLayerConfig[options.layer];
+  this.rootUrl_ = goog.isDef(options.url) ? options.url :
+      'https://stamen-tiles-{a-d}.a.ssl.fastly.net/';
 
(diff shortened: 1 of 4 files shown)


src/ol/source/stamensource.js
```diff
@@ -88,36 +88,78 @@ ol.source.StamenProviderConfig = {
...
+   * url
+   * @type {string}
+   * @private
+   */
 
-  goog.asserts.assert(options.layer in ol.source.StamenLayerConfig,
-      'known layer configured');
-  var layerConfig = ol.source.StamenLayerConfig[options.layer];
+  this.rootUrl_ = goog.isDef(options.url) ? options.url :
+      'https://stamen-tiles-{a-d}.a.ssl.fastly.net/';
```
tschaub: There is a change here in the interpretation of the `url` option.  Previously, if `url` was provided, `layer` would be ignored (or at least not used to configure the URL).  I think it is reasonable to say this is a bug (since `layer` is required, it should not be ignored).

Your change makes it explicit that the `url` option is the "root URL," and I think this change should be documented.

Could you update the documentation in [`olx.js`](https://github.com/SDaron/ol3/blob/stamen-source/externs/olx.js#L4555-L4560)?  The `url` option should now *not* include `{x}`, `{y}` or `{-y}`, and `{z}` placeholders.

SDaron: Indeed, but maybe should we also rename the option as "rootUrl" to be more consistent? If url option is needed, the end user can still use the generic ol.source.XYZ.

SDaron: I'm not sure also if I should remove the "@api stable" notation in this case...

ahocevar: Once something is stable, we cannot make it unstable again. So please just proceed as suggested above, and update the documentation of the `url` option in [`olx.js`](https://github.com/SDaron/ol3/blob/stamen-source/externs/olx.js#L4555-L4560). Also, a mention of this change in the [upgrade notes](https://github.com/SDaron/ol3/blob/stamen-source/changelog/upgrade-notes.md) would be good.

SDaron:
Unless I've misunderstood you, I think 7c4675e3c4db3ae689a832bc0da329c006967aa3 update the documentation as suggested by @tschaub ?

tschaub:

PR #3663, opened 2015-05-05
Readme should not include the version number

tschaub:
We should not repeat the version number throughout the source.  Eventually, the download page will be http://openlayers.org/download/.


ahocevar:
Looks good to me.

PR #3673, opened 2015-05-08
More control over ol.interaction.Draw, to allow e.g. square drawing

ahocevar:
This pull request adds a `geometryFunction` config option to `ol.interaction.Draw`, which gives the application developer control over the geometry that is created from the draw sketch.

Circle draw mode is a good use case where such a hook makes sense. To be able to create a serializable geometry instead of an `ol.geom.Circle`, a convenience function is added to allow for creation of a regular `ol.geom.Polygon` geometry representing a circle or - with fewer sides - e.g. a square.

To show how this works, the `draw-features.html` example can now also draw squares.

### DIFF
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -129,15 +129,6 @@ ol.interaction.Draw = function(options) {
       options.snapTolerance : 12;
 
   /**
-   * The number of points that must be drawn before a polygon ring can be
-   * finished.  The default is 3.
-   * @type {number}
-   * @private
-   */
-  this.minPointsPerRing_ = goog.isDef(options.minPointsPerRing) ?
-      options.minPointsPerRing : 3;
-
-  /**
    * Geometry type.
    * @type {ol.geom.GeometryType}
    * @private
@@ -150,6 +141,26 @@ ol.interaction.Draw = function(options) {
    * @private
    */
   this.mode_ = ol.interaction.Draw.getMode_(this.type_);
+
+  /**
+   * The number of points that must be drawn before a polygon ring or line
+   * string can be finished.  The default is 3 for polygon rings and 2 for
+   * line strings.
+   * @type {number}
+   * @private
+   */
+  this.minPoints_ = goog.isDef(options.minPoints) ?
+      options.minPoints :
+      (this.mode_ === ol.interaction.DrawMode.POLYGON ? 3 : 2);
+
+  /**
+   * The number of points that can be drawn before a polygon ring or line string
+   * is finished. The default is no restriction.
+   * @type {number}
+   * @private
+   */
+  this.maxPoints_ = goog.isDef(options.maxPoints) ?
+      options.maxPoints : Infinity;
 
   var geometryFunction = options.geometryFunction;
   if (!goog.isDef(geometryFunction)) {
@@ -225,6 +236,13 @@ ol.interaction.Draw = function(options) {
   this.sketchPoint_ = null;
 
   /**
+   * Sketch coordinates. Used when drawing a line or polygon.
+   * @type {ol.Coordinate|Array.<ol.Coordinate>|Array.<Array.<ol.Coordinate>>}
+   * @private
+   */
+  this.sketchCoords_ = null;
+
+  /**
    * Sketch line. Used when drawing polygon.
    * @type {ol.Feature}
    * @private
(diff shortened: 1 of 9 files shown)


ahocevar:
Thanks for any review.

ahocevar:
Thanks for any review. @tschaub, I think you expressed interest in this.

bartvde:
in the example, during drawing, there is no outline shown on the rectangle. This is shown on the circle though.

screen shot 2015-05-18 at 11 30 01
screen shot 2015-05-18 at 11 29 52

src/ol/interaction/drawinteraction.js
```diff
@@ -661,6 +706,42 @@ ol.interaction.Draw.prototype.updateState_ = function() {
...
+ * @param {number=} opt_sides Number of sides of the regular polygon. Default is
+ *     32.
+ * @param {number=} opt_angle Angle of the first point in radians. 0 means East.
+ *     Default is 0.
+ * @return {ol.interaction.Draw.GeometryFunctionType} Function that draws a
+ *     polygon.
+ * @api
+ */
+ol.interaction.Draw.createRegularPolygon = function(opt_sides, opt_angle) {
+  var sides = goog.isDef(opt_sides) ? opt_sides : 32;
```
bartvde: is it needed to have the default defined here as well? The default is already in ol.geom.Polygon.fromCircle ?

src/ol/interaction/drawinteraction.js
```diff
@@ -661,6 +706,42 @@ ol.interaction.Draw.prototype.updateState_ = function() {
...
+  return (
+      /**
+       * @param {ol.Coordinate|Array.<ol.Coordinate>|Array.<Array.<ol.Coordinate>>} coordinates
+       * @param {ol.geom.SimpleGeometry=} opt_geometry
+       * @return {ol.geom.SimpleGeometry}
+       */
+      function(coordinates, opt_geometry) {
+        var center = coordinates[0];
+        var radius = Math.sqrt(
+            ol.coordinate.squaredDistance(coordinates[0], coordinates[1]));
```
bartvde: minor, but can use center instead of coordinates[0]

bartvde:
@ahocevar this looks good to me except for a few minor comments.
When using the example, it doesn't feel like the most intuitive way to draw a box, but maybe I just need to get used to it. Also I'd expect the square to have the size of the radius all the time, and not only when being at a corner, but this is minor.

screen shot 2015-05-18 at 12 03 19
screen shot 2015-05-18 at 12 03 12

ahocevar:
@bartvde The missing outline is because no outline is defined in the default style for the draw interaction. The radius of the square is half of the diagonal, so the square corners intersect the imaginary circle at each vertex.

Any suggestions for improvements? Do you think the example should use a custom style (adds a lot of code)? Or should I introduce a RegularPolygon geometry type and add a default style like for the Circle type? For the radius to look more natural, I could change the example so the start angle is the heading from the center to the end point of the sketch line. Other ideas?

bartvde:
Hard questions to answer @ahocevar 
I am trying to look at prior art wrt geometry types, such as ISO SQL MM.

https://msdn.microsoft.com/en-us/library/bb964711.aspx
http://doesen0.informatik.uni-leipzig.de/proceedings/paper/68.pdf

It seems our circle geometry type is called CircularString elsewhere?

At first sight it doesn't seem common to have a different geometry type for regular polygons?

ahocevar:
@bartvde I addressed all your concerns and added a box drawing mode to the example.

src/ol/interaction/drawinteraction.js
```diff
@@ -462,60 +527,51 @@ ol.interaction.Draw.prototype.startDrawing_ = function(event) {
...
-    }
+    last = this.sketchCoords_;
+  } else if (this.mode_ === ol.interaction.DrawMode.POLYGON) {
+    coordinates = this.sketchCoords_[0];
+    last = coordinates[coordinates.length - 1];
     if (this.atFinish_(event)) {
       // snap to finish
       coordinate = this.finishCoordinate_.slice();
     }
+    last = coordinates[coordinates.length - 1];
```
bartvde: this seems like a duplicate line?

bartvde:
Thanks for the rework @ahocevar looks good to me (I only placed one question in the latest diff).
Also the user experience is much better now!

PR #3677, opened 2015-05-08
Add metadata to examples

### LINKED ISSUE #3674: Name, tags and description of verbatim examples don't show up on example-index
undefined-examples

I noticed this for the following examples:

* `geolocation-orientation.html`
* `mobile-full-screen.html`

There may be more and I am not 100% sure that the reason is the `example-verbatim`-template, but it seems likely.

tschaub:
This adds missing metadata to the examples.

image

Fixes #3674.

### DIFF
--- a/examples/geolocation-orientation.html
+++ b/examples/geolocation-orientation.html
@@ -1,5 +1,8 @@
 ---
-template: "example-verbatim.html"
+template: example-verbatim.html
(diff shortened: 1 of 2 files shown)


bartvde:
LGTM

marcjansen:
Great, thanks.

PR #3688, opened 2015-05-12
Add two missing properties to extern of WebGLContextAttributes

### LINKED ISSUE #3687: ol.webgl.getContext: attributes renamed by the compiler
The attributes passed to `getContext` are renamed by the compiler: the type is [Object](https://github.com/openlayers/ol3/blob/master/src/ol/webgl/webgl.js#L29) and the keys [are](https://github.com/openlayers/ol3/blob/master/src/ol/has.js#L127-L129) [not](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/webgl/webglmaprenderer.js#L98-L104) quoted

fredj:
Fixes #3687

### DIFF
--- a/externs/closure-compiler.js
+++ b/externs/closure-compiler.js
@@ -30,3 +30,15 @@ Touch.prototype.webkitRadiusX;
 
 /** @type {number} */
 Touch.prototype.webkitRadiusY;
+
+
+
+/**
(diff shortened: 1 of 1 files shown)


elemoine:
Please merge.

PR #3689, opened 2015-05-12
Fix WMTS.optionsFromCapabilities if no OperationsMetadata section

probins:
This function is currently failing if there is no OperationsMetadata section in the document, but this is not required. I think this is the correct fix.

### DIFF
--- a/src/ol/source/wmtssource.js
+++ b/src/ol/source/wmtssource.js
@@ -458,7 +458,8 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
       'requestEncoding (%s) is one of "REST", "RESTful", "KVP" or ""',
       requestEncoding);
 
-  if (!wmtsCap['OperationsMetadata'].hasOwnProperty('GetTile') ||
+  if (!wmtsCap.hasOwnProperty('OperationsMetadata') ||
+      !wmtsCap['OperationsMetadata'].hasOwnProperty('GetTile') ||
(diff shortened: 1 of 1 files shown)


src/ol/source/wmtssource.js
```diff
@@ -458,7 +458,8 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
       'requestEncoding (%s) is one of "REST", "RESTful", "KVP" or ""',
       requestEncoding);
 
-  if (!wmtsCap['OperationsMetadata'].hasOwnProperty('GetTile') ||
+  if (!wmtsCap['OperationsMetadata'] || (wmtsCap['OperationsMetadata'] &&
+      !wmtsCap['OperationsMetadata'].hasOwnProperty('GetTile')) ||
```
elemoine: I'd use `hasOwnProperty` for the `OperationsMetadata` property as well:

```js
  if (!wmtsCap.hasOwnProperty('OperationsMetadata') ||
      !wmtsCap['OperationsMetadata'].hasOwnProperty('GetTile')) {

  }
```

probins:
ok. And you're right: the additional check for the property isn't necessary - if it isn't there the condition will be met, and the if clause will be executed.

elemoine:
> ok. And you're right: the additional check for the property isn't necessary - if it isn't there the condition will be met, and the if clause will be executed.

Yes, exactly.

PR #3693, opened 2015-05-13
Fix ol.extent.containsExtent documentation

tremby:
The `ol.extent.containsExtent` documentation had its arguments backwards (or the implementation did).

The documentation said "the first extent is contained by or on the edge of the second", but the function checked the opposite.

The wording was also a little strange, since from the name of the function alone `containsExtent` I'd guess that the first argument would be the (potential) container, and the second would be the (potentially) contained. But the documentation has the wording "check if one extent is *contained by* or on the edge of another", suggesting the first argument is the contained and the second the container.

This patch keeps the current functionality but clarifies the documentation.

### DIFF
--- a/src/ol/extent.js
+++ b/src/ol/extent.js
@@ -164,12 +164,15 @@ ol.extent.containsCoordinate = function(extent, coordinate) {
 
 
 /**
- * Check if one extent is contained by or on the edge of another.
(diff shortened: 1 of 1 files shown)


elemoine:
Thanks!

Issue #3379, opened 2015-03-23
Support blend modes for layer styles


tremby:
It would be very useful to specify a [blend mode](http://dev.w3.org/fxtf/compositing-1/#blendingseparable) along with a vector style's fill or stroke, or for a layer's tiles.

This would make it possible, for instance, to avoid obscuring labels on the underlying tiles when putting a brightly coloured vector overlay on top. For example if the background is white with dark labels, a vector overlay of any colour, full opacity and "multiply" blend mode will give the colour on any white areas but the labels will still be visible.

At present the normal blend mode is used, which mixes colours, so in the above example to keep labels visible we need to use a semi-transparent overlay. In this case we don't get the full brightness of the overlay and additionally the label below is washed out to some extent.

Any possibility of including this feature in future?

ahocevar:
This sounds like a good idea indeed.

lun471k:
This feature would be quite useful for the applications I develop using OL3 too.

kalessine:
Hi! could I request a bitwise OR blend... I would like to make a map with 3 vector layers, with each layer having a single channel (either red, green or blue) the final composite is dark at value of zero for all 3 vectors, white at max value, and coloured for all other in-between values..the desired effect is not achievable using alpha blending at the moment... if anyone has any hints how to do this with ol3 at the moment please message me =D
Thanks   
James

tremby:
That'd be possible with this feature, using the 'add' blend mode.

tremby:
I'd be willing to put a bounty on this feature. Is there a mechanism to do that?

ahocevar:
We currently do not have a mechanism to put bounties on a feature.

elmuSSo:
I would also like to support this feature request. I think blend modes (especially MULTIPLY) should be added to every map application software by default. I can't imagine merging two layers without blending modes. I was also lobbying for that feature in Locus and now I will lobby for it in MapProxy. It is a must-have feature.

lun471k:
I've been able to ahieve what I wanted (using multiply) by binding a custom function when a new layer/features is added to the map and edit the `globalCompositeOperation` property of the canvas' context. But it must be called each and every time I add an element. 

I still think this should be managed by OL3 itself.

tremby:
[...]

PR #3696, opened 2015-05-15
Add an example for various blend modes

marcjansen:
This PR suggests adding an example that shows how to use layer events to change the `globalCompositeOperation` of the canvas:

blend-difference

blend-xor

blend-source-out-no-green

See also #3379


tschaub:
Really nice looking example @marcjansen.  Please merge.

marcjansen:
Thanks for the review.

elmuSSo:
I presume this are the vector layers. Will it also work for raster Tiled layers?

elmuSSo:
I presume this are the vector layers. Will it also work for raster Tiled layers?
Maybe it would be beneficial to add a second example (or extend this one) with showing how to blend two (and/or more) Tile layers?

elmuSSo:
I presume this are the vector layers. Will it also work for raster Tiled layers?
- Maybe it would be beneficial to add a second example (or extend this one) with showing how to blend two (or more) Tile layers?
- Maybe it would be also nice to add a turn on/off checkbox for each of the layers (in this example blue/green/red)

elmuSSo:
I presume this are the vector layers. Will it also work for raster Tiled layers?
- Maybe it would be beneficial to add a second example (or extend this one) with showing how to blend two (or more) Tile layers?
- Maybe it would be also nice to add a turn on/off checkbox for each of the layers (in this example blue/green/red)

PS. There is a typo in the source of that example: "Unind the pre- and postcompose handlers to the passed layers."

PR #3699, opened 2015-05-18
Add support for scientific notation to WKT format

### LINKED ISSUE #3686: ol.format.WKT doesn't recognize scientific number notation
I'm having a problem where default C# / SQL Server implementation of geography to WKT methods return small numbers in scientific number format with character E in them. I get the following error
Uncaught Error: Unexpected `E` at position 123 in `MULTIPOLYGON(((-5.7156469384940491 50.082709053271955, -5.7059626929961649 ... **9.1993293867552632E-05** 53.535345770529375, ...))).
The most popular 3rd party library for spatial data in C# creates WKT in same format.

marcjansen:
This PR suggests that we add support for WKTs where the coordinates are in scientific notation.

Please review and share your thoughts.

Fixes #3686

### DIFF
--- a/src/ol/format/wktformat.js
+++ b/src/ol/format/wktformat.js
@@ -491,12 +491,23 @@ ol.format.WKT.Lexer.prototype.nextToken = function() {
 ol.format.WKT.Lexer.prototype.readNumber_ = function() {
   var c, index = this.index_;
   var decimal = false;
+  var scientificNotation = false;
   do {
     if (c == '.') {
       decimal = true;
+    } else if (goog.isDef(c) && c.toLowerCase() == 'e') {
+      scientificNotation = true;
     }
     c = this.nextChar_();
-  } while (this.isNumeric_(c, decimal));
+  } while (
+      this.isNumeric_(c, decimal) ||
+      // once c is defined we may have a scientific number indicated by 'e'
+      // but only if we haven't detected a scientific number before
+      !scientificNotation && goog.isDef(c) && c.toLowerCase() == 'e' ||
+      // once we know that we have a scientific number, both '-' and '+'
+      // are allowed
+      scientificNotation && (c == '-' || c == '+')
+  );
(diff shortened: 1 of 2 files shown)


src/ol/format/wktformat.js
```diff
@@ -491,12 +491,23 @@ ol.format.WKT.Lexer.prototype.nextToken = function() {
 ol.format.WKT.Lexer.prototype.readNumber_ = function() {
   var c, index = this.index_;
   var decimal = false;
+  var scientificNotation = false;
   do {
     if (c == '.') {
       decimal = true;
+    } else if (c !== undefined && c.toLowerCase() == 'e') {
```
elemoine: Any reason for not using `goog.isDef`, which we use anywhere else?

elemoine:
I just added a comment regarding the use of `goog.isDef`. Otherwise looks good to me. Thanks for the thorough tests.

src/ol/format/wktformat.js
gberaudo: What about simply testing `c == 'e' || c == 'E'`?

fredj: See https://github.com/openlayers/ol3/blob/master/src/ol/format/wktformat.js#L513 (`toUpperCase`): maybe the character is always `E` ? (haven't tried)

marcjansen:
Thanks for the review, Éric. Now using `goog.isDef`. I'll merge once travis is happy.

elemoine:
I love when people use `É` rather than `E` in my first name. Thanks for that :-)

src/ol/format/wktformat.js
fredj: > See https://github.com/openlayers/ol3/blob/master/src/ol/format/wktformat.js#L513 >  (toUpperCase): maybe the character is always E ? (haven't tried)

It's not the case

marcjansen:
I don't always think of it, but sometimes ...

PR #3709, opened 2015-05-19
vector-wfs example does not work in JSFiddle

bartvde:
This is because the callback is not global in that case.


bartvde:
PR attached

ahocevar:
Looks good. Please merge.

PR #3720, opened 2015-05-21
Add missing goog.provides in drawinteraction.js

elemoine:
This PR adds missing `goog.provides` to the `drawinteraction.js` file. The compilation of applications fails without these `goog.provides` if the draw interaction is not used.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -5,6 +5,7 @@
 #### `ol.interaction.Draw` changes
 
 * The `minPointsPerRing` config option has been renamed to `minPoints`. It is now also available for linestring drawing, not only for polygons.
+* The `ol.DrawEvent` and `ol.DrawEventType` types were renamed to `ol.interaction.DrawEvent` and `ol.interaction.DrawEventType`. This has an impact on your code only if your code is compiled together with ol3.
 
 #### `ol.tilegrid` changes
 
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2360,7 +2360,7 @@ olx.interaction.DragZoomOptions.prototype.style;
  *     maxPoints: (number|undefined),
  *     minPoints: (number|undefined),
  *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
- *     geometryFunction: (ol.interaction.Draw.GeometryFunctionType|undefined),
+ *     geometryFunction: (ol.interaction.DrawGeometryFunctionType|undefined),
  *     geometryName: (string|undefined),
  *     condition: (ol.events.ConditionType|undefined),
  *     freehandCondition: (ol.events.ConditionType|undefined)}}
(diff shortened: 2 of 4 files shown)


tschaub:
The [build failure](https://travis-ci.org/openlayers/ol3/builds/63491443) is a reminder (to me) that it will be nice to remove the need for `test_requires.js`.

elemoine:
Do you understand why the build fails? I don't right now.

elemoine:
I understood (with help from @tsauerwein) and fixed the problem with an additional commit.

bartvde:
LGTM

elemoine:
This deserves some comments before merging:

So https://github.com/elemoine/ol3/commit/879307da1bde22363c01548804cfdd6910e43e4b just added `goog.provide`'s for `ol.interaction.Draw.GeometryFunctionType` and `ol.interaction.DrawMode`. But with just that commit I get an error when attempting to run the tests:

```
[wrk12 ol3 ((879307d...) $%)]$ make test
node tasks/test.js
goog.require could not find: ol.interaction.Draw.GeometryFunctionType
Error: goog.require could not find: ol.interaction.Draw.GeometryFunctionType

  http://127.0.0.1:3001/test/loader.js/home/elemoine/src/ol3/node_modules/closure-util/.deps/library/b0b32b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/base.js:677
  http://127.0.0.1:3001/test/loader.js/home/elemoine/src/ol3/build/test_requires.js:173
make: *** [test] Error 1
```

Closure Library complains that the `goog.require` to `ol.interaction.Draw.GeometryFunctionType` in `test_requires.js` cannot be satisfied because it cannot find a corresponding `goog.provide`. Me: « What?! I just added that `goog.provide`, so Closure Library should be able to find it! ».

Actually doing the following is wrong:

```js
goog.provide('ol.interaction.Draw.GeometryFunctionType');

/**
 * @constructor
 */
ol.interaction.Draw = function() {
  // ...
};
```

Because the definition of the `ol.interaction.Draw` constructor "destroys" the `GeometryFunctionType` that the `goog.provide` created on the (initial) `ol.interaction.Draw` object! So a `goog.require` to `ol.interaction.Draw.GeometryFunctionType` later in the code will fail with the error reported when attempting to run the tests.

So I decided to change `ol.interaction.Draw.GeometryFunctionType` to `ol.interaction.DrawGeometryFunctionType`, which is consistent with the way we define `@typedef`s elsewhere.

And for consistency with [other files](https://github.com/openlayers/ol3/blob/master/src/ol/source/vectorsource.js#L6-L7) commit https://github.com/elemoine/ol3/commit/531b35d7c247159421a767bd9ce749e0ceb28554 also `ol.DrawEvent` and `ol.DrawEventType` to `ol.interaction.DrawEvent` and `ol.interaction.DrawEventType`.

I also added commit e766d00 that adds an upgrade note related to the renaming of the `DrawEvent` and `DrawEventType`  types.

Thanks @bartvde for the review.

fredj:
Thanks for the explanation

Issue #3724, opened 2015-05-22
Tap event sensitivity on tablet


pgiraud:
Using a Samsung Galaxy Tab S 10.5, the drawing tools don't behave as expecting. The user needs to touch the screen for a small amount of time before a point is added to the map. A short press doesn't have any effect. With a too long press, the system (android) wants to do copy/paste.

This video shows the problems: https://www.youtube.com/watch?v=dY4ng8BjE14

@petzlux can you please tell us which version of Chrome is installed on this tablet?

petzlux:
Its Chrome 42.0.2311.111 on Android 5.0.2

fredj:
I've the same kind on issues with a [Samsung Galaxy Tab 4 10.1](http://www.samsung.com/us/mobile/galaxy-tab/SM-T530NZWAXAR) and Chrome 42.0.2311.111.

http://openlayers.org/en/master/examples/measure.html

fredj:
The position between a down and a up event is quite big on this device and [the distance test](https://github.com/openlayers/ol3/blob/master/src/ol/interaction/drawinteraction.js#L390) often fails.

fredj:
@petzlux please test http://fredj.io/ol/click-tolerance/examples/measure.html
The click tolerance is 6px (2px in [master](https://github.com/openlayers/ol3/blob/master/src/ol/interaction/drawinteraction.js#L266))

petzlux:
@fredj  Super, works much much better !

elemoine:
Same issue here on my Android Phone with Chrome 43.0.2357.78.

elemoine:
> Same issue here on my Android Phone with Chrome 43.0.2357.78.

Actually it might not be as bad on my phone. I do get pointerdown/pointerup sequences that don't draw a point, but I still can draw a complete geometry without too much trouble.

But the user experience is definitely better when using a click tolerance of 6px (as in @fredj's click-tolerance branch) rather than 2px (master).

What I would like to do is to add a `clickTolerance` option to the Draw interaction. We can mark it as experimental for now, but we can play with it in applications to try to figure out what could be a good value.

PR #3726, opened 2015-05-22
[wip] Make the ol.interaction.Draw click tolerance configurable

fredj:
And change the click tolerance from 2 to 6 px.

See #3724


fredj:
Note: on a Nexus 5 (Chrome 42.0.2311.111) there is no position difference between a `pointerdown` and a `pointerup` event

elemoine:
Is it still WIP @fredj? It looks good to me. In my branch I used the current value (2 pixels) as the default value for clickTolerance, but we should probably use a larger value as using 2 pixels is not appropriate on some (sensitive?) touch devices.

fredj:
Your pull adds tests, we should use it instead of this one ...

PR #3732, opened 2015-05-26
Add a method to bind button bluring on mouseout/focusout

marcjansen:
This PR suggests adding a `protected` method `bindMouseOutFocusOutBlur` to `ol.control.Control` to have fewer duplicated code.

Please review.

### DIFF
--- a/src/ol/control/control.js
+++ b/src/ol/control/control.js
@@ -3,6 +3,7 @@ goog.provide('ol.control.Control');
 goog.require('goog.array');
 goog.require('goog.dom');
 goog.require('goog.events');
+goog.require('goog.events.EventType');
 goog.require('ol.MapEventType');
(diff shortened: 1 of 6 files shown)


fredj:
Very good idea ! The function namespace could be `ol.dom` instead (?)

marcjansen:
> The function namespace could be ol.dom instead (?)

Yeah probably, but since currently only controls use it, I was tempted to put it there.

fredj:
You're right. And since it's a non api function it could be charged later if needed

marcjansen:
Agreed. This is a LGTM, right :wink: ?

fredj:
> Agreed. This is a LGTM, right :wink: ?

it is

marcjansen:
Thanks!

PR #3735, opened 2015-05-28
Snap example: remove featureoverlay from tags

probins:
Minor point, but the snap example doesn't use a FeatureOverlay


fredj:
Thanks

PR #3736, opened 2015-05-28
Fix invalid example HTML markup

fredj:


bartvde:
LGTM

fredj:
Thanks for the review

PR #3738, opened 2015-05-28
Improve doucmentation for ol.TileUrlFunctionType

ahocevar:
I think our users deserve that we demystify the tileUrlFunction. This pull request is an attempt to explain how tile coordinates work.


ahocevar:
Invoking the "no code change" policy and merging.

bartvde:
Thanks for adding docs on this@ahocevar

PR #3740, opened 2015-05-29
Add @fires to select interaction

### LINKED ISSUE #3733: Select event in interaction.Select not in api docs
#3018 introduced the select event, but this is not listed in the api docs. IIRC it has to be added to externs/oli.js - is that correct?

Edit: looks like it already is in oli.js, so the problem must be elsewhere

probins:
I think this should fix #3733


marcjansen:
Yes, this fixes it, thanks @probins:

Before:
before

After:
after

PR #3745, opened 2015-05-29
ol.source.TileVector: expose and document forEachFeatureInExtentAtResolution

plepe:
There already was a forEachFeatureInExtentAtResolution function - exactly what I needed. I exposed it to the API and added a description.


plepe:
replaced by #3857

PR #3747, opened 2015-05-31
Make tileCoordTransform a member again

### LINKED ISSUE #3744: different Y-offset with new ol.tilegrid.createXYZ instead of ol.tilegrid.XYZ
I'm just testing the current development version, and I noticed that the XYZ tilegrid has changed -> it has been replaced by createXYZ, but that is not the problem. The problem is, that the Y-offset of the requested tiles is different.

I'm using an ol.source.TileVector with ol.tilegrid.XYZ(). I replaced the ol.tilegrid.XYZ by ol.tilegrid.createXYZ, e.g.:

```js
var source = new ol.source.TileVector({
  url: "something/{z}/{x}/{y}.json",
  tileGrid: new ol.tilegrid.createXYZ()
});
```

Now, the Y coordinate of the requested tiles is totally wrong, before it was e.g. 11361, now it is 21406. X and Z are correct.

### LINKED ISSUE #3750: tile-vector example is broken
#3747 fixes it.

ahocevar:
Instead of using the static createOriginTopLeftTileCoordTransform function, the correct transform is now a non-API config option of the tile grid. This allows e.g. `ol.tilegrid.createXYZ` to create a tile grid with the correct tile coord transform function built in.

Fixes #3744.
Fixes #3750.

### DIFF
--- a/src/ol/tilegrid/tilegrid.js
+++ b/src/ol/tilegrid/tilegrid.js
@@ -2,6 +2,7 @@ goog.provide('ol.tilegrid.TileGrid');
 
 goog.require('goog.array');
 goog.require('goog.asserts');
+goog.require('goog.functions');
 goog.require('goog.math');
 goog.require('goog.object');
 goog.require('ol');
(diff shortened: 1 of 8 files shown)


ahocevar:
This also fixes the currently broken tile-vector example.

marcjansen:
I just restarted the build.

tsauerwein:
Could we have a test so that this doesn't break again in the future? :)

ahocevar:
Adding a test now @tsauerwein.

ahocevar:
Tests added @tsauerwein.

src/ol/tilegrid/tilegrid.js
```diff
@@ -134,17 +149,9 @@ ol.tilegrid.TileGrid = function(options) {
...
-    var extentWidth = ol.extent.getWidth(extent);
-    var extentHeight = ol.extent.getHeight(extent);
-    var fullTileRanges = new Array(this.resolutions_.length);
-    var tileSize;
-    for (var z = 0, zz = fullTileRanges.length; z < zz; ++z) {
-      tileSize = ol.size.toSize(this.getTileSize(z), this.tmpSize_);
-      fullTileRanges[z] = new ol.TileRange(
-          0, Math.ceil(extentWidth / tileSize[0] / this.resolutions_[z]) - 1,
-          0, Math.ceil(extentHeight / tileSize[1] / this.resolutions_[z]) - 1);
+    if (!goog.isNull(extent)) {
```
bartvde: can we use a single if loop with goog.isDefAndNotNull ?

```
} else if (goog.isDefAndNotNull(extent)) {
```

bartvde:
LGTM @ahocevar I only put one small comment

tsauerwein:
Thanks for adding the tests! Much appreciated.

PR #3751, opened 2015-06-01
Do not rely on remote services for tests

ahocevar:
This avoids tests failing with slow or no internet connection.


bartvde:
LGTM

PR #1718, opened 2014-02-17
Add ol.format.TWKB (work in progress)

twpayne:
This builds on the support of `ol.format.Binary` in #1717 to add TWKB support, succeeding #1615.

This is a work in progress, but is mostly complete. It's missing support for GeometryCollections and needs better testing.

I'll add a comment to this PR when it is ready for final review.

PR #1615, opened 2014-01-28
Adding TWKB format

nicklasaven:
TWKB format is a compressed binary format. 

It is new and a moving target but is in the PostGIS trunk.

The format is described here:
https://github.com/nicklasaven/TWKB

and a working example can be found here:
http://sandbox.postgisonline.org/oltest/

It is working also in IE9 but I guess the rewrite of the XHRIO part needs to be discussed.

The code is passing all testes, but it is not very well tested yet with different geometry types and so on.

I have tried to follow the style of other code in the project but I am not a JavaScript guy so bare with some ugliness.

I will fix it when you point at it.

PR #3756, opened 2015-06-02
ol.format.TWKB

bjornharrtell:
With the TWKB format getting ready for production use with PostGIS 2.2 I think it would be nice to support it out of the box in a future OpenLayers release. This is a naive implementation using GeoJSON conversion. Future work could parse directly to OpenLayers features.


marcjansen:
We once already had a PR for TWKB

bjornharrtell:
Yes but that effort was abandoned and later removed for more or less good reasons. It was a fairly complex implementation done directly as a part of ol3 where as this PR suggest to use the twkb library as a dependency.

For reference the previous attempt was #1615, #1717, #1718.

sweco-sebhar:
I've come to the conclusion that custom formats is better plugged in as a custom loader for Vector sources. #3833 proposes to add a similar option for TileVector sources.

Implementing a format as a part of the OpenLayers code base does not makes good sense until it is mainstream and popular.

fredj:
> This PR can be closed.

Issue #3737, opened 2015-05-28
Add pixel/coordinate to ol.SelectEvent


probins:
If pixel and/or coordinate were available in SelectEvent, the code for creating popups/overlays for features such as in the vector-layer example could simply use the select interaction instead of creating its own FeatureOverlay. It looks pretty straightforward to change SelectEvent to pass these vars on from the MapBrowserEvent and I was in the middle of creating a PR when it struck me it might be better/simpler to change `ol.SelectEvent` so it inherits from `ol.MapBrowserEvent` instead of `goog.events.event`. Would there be any problems/issues with doing this?

bjornharrtell:
PR #3454 might be relevant.

probins:
thanks @bjornharrtell I managed to miss that one, which is also on my list. What do you think about adding pixel in that PR? Or would it be better in a separate one?

bjornharrtell:
Perhaps it's better to make it a separate one and I'd like to see the tests in #3453 in first.

bjornharrtell:
Was just hit by goog.events.event not being enough, will make a PR to change base to ol.MapBrowserEvent.

bjornharrtell:
@probins #3757 ready for review.

probins:
thanks for tackling this - we're obviously trying to do similar things in our apps :-)

ahocevar:
Fixed with #3757.

PR #3757, opened 2015-06-03
Add mapBrowserEvent as a member of ol.SelectEvent

bjornharrtell:
As discussed in #3737 the goog.event.Event is not enough sometimes, for instance you might want to know the shiftKey status or the pixel clicked.

### DIFF
--- a/externs/oli.js
+++ b/externs/oli.js
@@ -120,19 +120,25 @@ oli.MapEvent.prototype.frameState;
 /**
  * @interface
  */
- oli.SelectEvent = function() {};
+oli.SelectEvent = function() {};
 
 
- /**
-  * @type {Array.<ol.Feature>}
-  */
+/**
+ * @type {Array.<ol.Feature>}
+ */
 oli.SelectEvent.prototype.deselected;
 
 
- /**
-  * @type {Array.<ol.Feature>}
-  */
+/**
+ * @type {Array.<ol.Feature>}
+ */
 oli.SelectEvent.prototype.selected;
+
+
+/**
+ * @type {ol.MapBrowserEvent}
+ */
+oli.SelectEvent.prototype.mapBrowserEvent;
 
 
 
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -46,11 +46,13 @@ ol.interaction.SelectFilterFunction;
  * @param {string} type The event type.
  * @param {Array.<ol.Feature>} selected Selected features.
  * @param {Array.<ol.Feature>} deselected Deselected features.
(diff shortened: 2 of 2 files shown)


probins:
I think `oli.js` has to be changed too, no?

bjornharrtell:
I'm not sure. The current interface definition of ol.SelectEvent (and other interfaces) seem to leave out these details:

```js
/**
 * @interface
 */
 oli.SelectEvent = function() {};

bjornharrtell:
Or hmm, should the `oli.MapBrowserEvent` properties be repeated in the `oli.SelectEvent` interface? Not sure. :(

bjornharrtell:
But I am running a custom build (compiled as usual with closure in advanced mode) and having no issues accessing members from `oli.MapBrowserEvent` on the event.

ahocevar:
Not sure if this is the right approach. ol.MapBrowserEvent is a low level event fired by ol.Map. Other events like ol.SelectEvent are high level events that should not have to deal with pixels, coordinates and modifier keys.

I tried to understand the use case reading #3737, but I failed. So why exactly do you guys need this additional information on ol.SelectEvent?

probins:
take the vector-layer example. This is basically doing what the select interaction does: highlight a particular feature on click/hover. It displays the text in a fixed position, but if you want to display as a tooltip or overlay at a particular pixel/coordinate, you need that pixel/coordinate. The select interaction has this data, but it's not exported in the select event, so apps can't access it.

bjornharrtell:
I see your point @ahocevar.

My use case is that i want to defer some application logic that normally happens on select event based on if shift is pressed at that time.

Would you be more positive if MapBrowserEvent instead is a member of SelectEvent called originEvent?

ahocevar:
@bjornharrtell Adding a mapBrowserEvent property to ol.SelectEvent sounds like a better approach.

@probins In that example, there are several ways to achieve this. We wanted to show highlighting without a Select interaction.

probins:
> at a particular pixel/coordinate

what I should have said here was 'at the clicked/mouseover pixel/coordinate'. Adding the mapBrowserEvent property to ol.SelectEvent would be fine by me.

probins:
and in this case it would have to be added to `oli.js`, right?

bjornharrtell:
@probins, @ahocevar - I've reimplemented as suggested with added property mapBrowserEvent (yep, needed to be defined on the interface)

probins:
LGTM

probins:
Fixes #3737

ahocevar:
Thanks!

marcjansen:
Thanks from me as well. Great collaboration here.

PR #3758, opened 2015-06-03
Removal of ol.FeatureOverlay

### LINKED ISSUE #3704: Draw / Edit polygon over 180° or -180°
Drawing polygon over 180° or -180° (Example http://openlayers.org/en/v3.5.0/examples/draw-and-modify-features.html?q=draw&mode=raw), the polygon disapear when the mouse move outside [-180°,180°].
This is not exactly a disappearance because the polygon is translated (+360 ° or -360 °). The result is the same for high zoom levels where we can not observe these translations.

ahocevar:
With an option to disable the spatial index of `ol.source.Vector`, and a `map` option for `ol.layer.Layer` to render on top of other layers without adding it to the map's layers collection, and a way to manage features of an `ol.source.Vector` in an `ol.Collection`, we can get rid of `ol.FeatureOverlay`.

This simplifies code in many places, and it also improves date line handling, which was not possible with `ol.FeatureOverlay`.

Fixes #3704.
Fixes #2386.


externs/olx.js
```diff
@@ -3449,6 +3489,17 @@ olx.layer.TileOptions.prototype.source;
 
 
 /**
+ * Sets the layer as overlay on a map. The map will not manage this layer in its
+ * layers collection, and the layer will be rendered on top. This is useful for
+ * temporary layers. The standard way to add a layer to a map and have it
+ * managed by the map is use {@link ol.Map#addLayer}.
```
probins: to use ...

externs/olx.js
```diff
@@ -3544,6 +3596,17 @@ olx.layer.VectorOptions.prototype.hue;
...
```
probins: to use ...

probins:
a quick comment. As mentioned in #3737, a featureoverlay isn't needed in several of the examples if the select interaction is used instead; this also simplifies those examples. #3757 enables that, so it might be better to merge #3757 and change those examples first. Then they would no longer need to be changed in this PR. ISTM this would be better than changing them here, and then changing them again as and when #3757 is merged.

src/ol/source/vectorsource.js
```diff
@@ -290,6 +318,54 @@ ol.source.Vector.prototype.addFeaturesInternal = function(features) {
 
 
 /**
+ * @param {!ol.Collection.<ol.Feature>} collection Collection.
+ * @private
+ */
+ol.source.Vector.prototype.bindFeaturesCollection_ = function(collection) {
+  goog.asserts.assert(goog.isNull(this.featuresCollection_),
+      'bindFeatuesCollection can only be called once');
```
elemoine: Typo.

laball:
@ahocevar While I use ol3 and  dojo 1.8 to build my website,the dojo AMD loader always got an error "multipleDefine",but when I remove the script of ol3,the error never come out. So is there any conflict between dojo and ol3 ? I am a Chinese GIS Engineer and my English is poor,hope you can read this and help me.

bartvde:
@laball what is the relation with this PR?

laball:
@bartvde  no,there is no relation with this PR,I just want to ask that question,is there any solutions?

bartvde:
Commenting on a PR is not a way to ask questions

externs/olx.js
```diff
@@ -2471,6 +2472,14 @@ olx.interaction.DrawOptions.prototype.freehandCondition;
 
 
 /**
+ * Wrap the world horizontaly on the sketch overlay. Default is `false`.
+ * @type {boolean|undefined}
```
bartvde: horizontaly -> horizontally

externs/olx.js
```diff
@@ -2588,6 +2598,14 @@ olx.interaction.ModifyOptions.prototype.features;
 
 
 /**
+ * Wrap the world horizontaly on the sketch overlay. Default is `false`.
```
bartvde: horizontaly -> horizontally

externs/olx.js
```diff
@@ -2804,6 +2823,14 @@ olx.interaction.SelectOptions.prototype.filter;
 
 
 /**
+ * Wrap the world horizontaly on the selection overlay. Default is `true`.
```
bartvde: horizontaly -> horizontally

src/ol/interaction/drawinteraction.js
```diff
@@ -269,10 +269,14 @@ ol.interaction.Draw = function(options) {
...
   /**
    * Draw overlay where our sketch features are drawn.
-   * @type {ol.FeatureOverlay}
+   * @type {ol.layer.Vector}
    * @private
    */
-  this.overlay_ = new ol.FeatureOverlay({
+  this.overlay_ = new ol.layer.Vector({
+    source: new ol.source.Vector({
+      spatialIndex: false,
```
bartvde: shouldn't this be ```useSpatialIndex``` ?

src/ol/interaction/drawinteraction.js
```diff
@@ -674,7 +678,7 @@ ol.interaction.Draw.prototype.abortDrawing_ = function() {
     this.sketchFeature_ = null;
     this.sketchPoint_ = null;
     this.sketchLine_ = null;
-    this.overlay_.getFeatures().clear();
+    this.overlay_.getSource().clear();
```
bartvde: have you considered using true here for opt_fast ?

src/ol/interaction/drawinteraction.js
```diff
@@ -701,7 +705,9 @@ ol.interaction.Draw.prototype.updateSketchFeatures_ = function() {
   if (!goog.isNull(this.sketchPoint_)) {
     sketchFeatures.push(this.sketchPoint_);
   }
-  this.overlay_.setFeatures(new ol.Collection(sketchFeatures));
+  var overlaySource = this.overlay_.getSource();
+  overlaySource.clear();
```
bartvde: same question for opt_fast

src/ol/interaction/drawinteraction.js
ahocevar: I can try that. I'll also add a test to make sure that a features collection will stay in sync.

bartvde: Ah good point, it might no stay in sync in that case, didn't think about that.

bartvde:
I have 2 remaining questions which might be addressed simply by docs:

1) how does someone remove an unmanaged layer? setMap(null) ?
2) how is z-index managed if you have more than 1 layer unmanaged?

Since I'm not too familiar with the renderer code, it might be good to have someone more knowledgeable wrt the renderer code look specifically at the rendering changes as well. But I'll leave the call up to you @ahocevar

ahocevar:
@bartvde:

> 1) how does someone remove an unmanaged layer? setMap(null) ?

Exactly. I added this information to the API docs, and also added a unit test to verify that it works properly.

> 2) how is z-index managed if you have more than 1 layer unmanaged?

This is indeed the only breaking change after the removal of `ol.FeatureOverlay`. All feature overlays shared a single replay group, so `zIndex` was effectively stacked per map. All vector layers, even when unmanaged, are separate, so each layer stacks `zIndex` separately. I added this information to the upgrade notes.

externs/olx.js
```diff
@@ -5053,6 +5118,22 @@ olx.source.VectorOptions.prototype.url;
...
 
 /**
+ * By default, an RTree is used as spatial index. When features are removed and
+ * added frequently, and the total number of features is low, setting this to
+ * `false` may improve performance.
+ *
+ * When set to `false`, the features will be maintained in an
+ * {@link ol.Collection}, which can be retrieved through
+ * {@link ol.source.Vector#getFeaturesCollection}.
+ *
```
probins: I would add that the spatial queries for extent/coordinate only work if set to true

probins:
There's another change that I think should be mentioned in the release notes: if you were using a FO `getFeatures` returned a collection whereas in source.Vector it returns an array. Apps will have to take this into account. I have always found this a bit confusing. interaction.Select returns a collection too. IMO it would be better if this was consistent, though as these are marked as stable, perhaps it's too late to change.

externs/olx.js
ahocevar: Done.

ahocevar:
@probins:

> There's another change that I think should be mentioned in the release notes: if you were using a FO getFeatures returned a collection whereas in source.Vector it returns an array. Apps will have to take this into account. I have always found this a bit confusing. interaction.Select returns a collection too. IMO it would be better if this was consistent, though as these are marked as stable, perhaps it's too late to change.

I added that to the upgrade notes. You're right, it would be more consistent to use `getFeatures` wherever an array is returned, and `getFeaturesCollection` when a collection is returned. But since `getFeatures` is marked stable with different return types on `ol.interaction.Select` and `ol.source.Vector`, we have to live with this inconsistency for now.

ahocevar:
@bartvde:

> Since I'm not too familiar with the renderer code, it might be good to have someone more knowledgeable wrt the renderer code look specifically at the rendering changes as well. But I'll leave the call up to you

The only change to renderer code is the removal of the map replay group, which was only used by `ol.FeatureOverlay`. Do you feel comfortable reviewing this, just to verify that I did not remove more or less than necessary. I'm adding an upgrade note about the removed replay as well.

bartvde:
I did review it @ahocevar and it looked good to me.

ahocevar:
Thanks @bartvde. Is this good to merge?

bartvde:
LGTM

elemoine:
Shouldn't `unmanaged` be added to [`ol.layer.State`](https://github.com/openlayers/ol3/blob/master/src/ol/layer/layerbase.js#L29-L42)?

ahocevar:
Good call @elemoine. I'll create a pull request.

ahocevar:
See #3787 @elemoine

elemoine:
Quick note: proviously, all the feature overlays attached to a map were equal from the "feature render order" point of view. It's not the case with "umanaged" vector layers. "Unmanaged" vector layers are ordered: the features of the first "unmanaged" vector layer will be drawn below the features of the second "unmanaged" vector layer, etc.

ahocevar:
@elemoine True. I mentioned that in the upgrade notes.

bartvde:
@elemoine do you know of a use case where this will be problematic? We couldn't think of one.

elemoine:
> @elemoine do you know of a use case where this will be problematic? We couldn't think of one.

The problem is related to components that create vector layers internally (e.g. select interaction).

tschaub:
WOW

elemoine:
> The problem is related to components that create vector layers internally (e.g. select interaction).

In our application we have independent components, each using a `FeatureOverlay` with a specific style. The components could share an "unmanaged" vector layer, but I wonder how we can make the components use specific styles. We'll probably need to use per-feature styles, which will make our code messier.

ahocevar:
@elemoine As long as your styles don't use `zIndex`, there should not be a difference between unmanaged vector layers and feature overlays. If you rely on `zIndex`, you could also leverage an index for distinct feature groups. Assuming you had
```js
var overlay1 = new ol.FeatureOverlay({
  style: /* ... */
});
overlay1.addFeature(feature1);
var overlay2 = new ol.FeatureOverlay({
  style: /* ... */
});
overlay2.addFeature(feature2);
```
you could now do something like
```js
var layer = new ol.source.Vector({
  source: new ol.source.Vector(),
  style: /* ... */
});
featuresIndex = [{}, {}];
function addFeature(feature, index) {
  layer.getSource().addFeature(feature);
  featuresIndex[index][goog.getUid(feature)] = feature;
}
function removeFeature(feature, index) {
  layer.getSource().removeFeature(feature);
  featuresIndex[index][goog.getUi(feature)] = undefined;
}
```
In the styleFunction you just check for the feature whether it is in `featuresIndex[0]` or `featuresIndex[2]`, and style it appropriately.

elemoine:
As I said, currently, each component has its own `FeatureOverlay` and its own style. The components are independent and don't know about each other. It looks to me that, in the code you provided, the code that creates the shared vector layer knows the number of components, and it should know how each component wants to style its features. So, at this point, I honestly don't think this represents a good solution to my problem.

elemoine:
Maybe I can build an API/abstraction around your "index" idea. Any consumer of that API would first ask the API for an identifier or something. It will then use that identifier in subsequent operations like `setStyle`, `addFeature`, `removeFeature` and `clear`. That may work. Maybe that's what you had in mind.

ahocevar:
That sounds good @elemoine, much smarter than the app level solution I suggested above. Thanks!

probins:
There's a difference in display behaviour with this change. If you go to http://openlayers.org/en/master/examples/select-features.html and select the hover option, there's an annoying flickering as the cursor passes over the features. This doesn't occur with http://openlayers.org/en/v3.6.0/examples/select-features.html
Is it constantly adding/removing the feature from the selection layer?

ahocevar:
This is a bug. Thanks for reporting @probins. I'll create a separate ticket for that.

PR #3759, opened 2015-06-03
Mark tilegrid.createTileCoordTransform() @api

gberaudo:
This change is necessary for ol3-cesium since the removal of tilegridXYZ type.
See https://github.com/openlayers/ol3-cesium/pull/201

### DIFF
--- a/src/ol/tilegrid/tilegrid.js
+++ b/src/ol/tilegrid/tilegrid.js
@@ -114,11 +114,16 @@ ol.tilegrid.TileGrid = function(options) {
 
 
   /**
-   * TileCoord transform function for use with this tile grid. Returns unaltered
-   * tile coordinates by default.
+   * Creates a TileCoord transform function for use with this tile grid.
+   * Transforms the internal tile coordinates with bottom-left origin to
+   * the tile coordinates used by the {@link ol.TileUrlFunction}.
+   * The returned function expects an {@link ol.TileCoord} as first and an
+   * {@link ol.proj.Projection} as second argument and returns a transformed
+   * {@link ol.TileCoord}.
    * @param {{extent: (ol.Extent|undefined)}=} opt_options Options.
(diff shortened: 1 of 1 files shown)


ahocevar:
As discussed at the Sprint in Schladming, we now require at least one sentence of prose description for symbols we make exportable.

gberaudo:
Thank you @ahocevar for your quick feedback.
There is already a descriptive sentence so what should I do?

Another point, this method is a property of the `Tilegrid` class; do you think I should create a wrapping method?

gberaudo:
I checked the generated documentation. The descriptive sentence is present but not the parameters and return value of the function.

ahocevar:
It's ok to keep this function as property. Just change the descriptive sentence to something like

"Creates a TileCoord transform function for use with this tile grid. Transforms the internal tile coordinates with bottom-left origin to the tile coordinates used by the tileUrlFunction."

Since this is pretty internal, the user does not have to know the arguments and return type of the created function.

ahocevar:
It's ok to keep this function as property. Just change the descriptive sentence to something like

"Creates a TileCoord transform function for use with this tile grid. Transforms the internal tile coordinates with bottom-left origin to the tile coordinates used by the tileUrlFunction."

Since this is pretty internal, the user does not have to know the arguments and return type of the created function.

marcjansen:
I tend to disagree. When the docs are enhanced, why don't we document arguments and return type in prose? It takes 60 seconds and will help when one wants to use the library

ahocevar:
Something like 'The returned function expects an ol.TileCoord as argument, and returns a transformed ol.TileCoord.'?

marcjansen:
Great wording. That'll be enough. Thanks.

gberaudo:
Thank you @ahocevar and @marcjansen for your responsive feedback and guidance.
I have updated the PR, please merge if it is OK.

marcjansen:
Fine now, thanks.

PR #3761, opened 2015-06-04
fix ol.source.Vector#getFeaturesAtCoordinate

alvinlindstam:
As noted in https://github.com/openlayers/ol3/issues/3703, ol.source.Vector#getFeaturesAtCoordinate sometimes returns empty when it should return matching features.

Currently, it only returns features containing the coordinate, while the docs state that it should return all intersecting geometries. This PR changes this.

While doing this, I noted that ol.geom.Circle and ol.geom.LinearRing doesn't implements ol.geom.Geometry's abstract intersectsExtent method. This means that using this method will fail on any source containing those geometries. I'll open an issue about implementing intersectsExtent on those geometries. Perhaps that should be fixed, and these tests updated, before merging this PR?

### DIFF
--- a/src/ol/source/vectorsource.js
+++ b/src/ol/source/vectorsource.js
@@ -343,35 +343,6 @@ ol.source.Vector.prototype.forEachFeature = function(callback, opt_this) {
 
 
 /**
- * Iterate through all features whose geometries contain the provided
- * coordinate, calling the callback with each feature.  If the callback returns
- * a "truthy" value, iteration will stop and the function will return the same
- * value.
- *
- * @param {ol.Coordinate} coordinate Coordinate.
- * @param {function(this: T, ol.Feature): S} callback Called with each feature
- *     whose goemetry contains the provided coordinate.
- * @param {T=} opt_this The object to use as `this` in the callback.
- * @return {S|undefined} The return value from the last call to the callback.
- * @template T,S
- */
-ol.source.Vector.prototype.forEachFeatureAtCoordinateDirect =
-    function(coordinate, callback, opt_this) {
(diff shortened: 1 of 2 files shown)


alvinlindstam:
However, i just realized that omitting geometry.intersectsExtent from some features might be deliberate. 

I guess that LinearRings does not intersect an extent that is completely contained within it, which would match the expected behaviour of getFeaturesAtCoordinate.

But is ol.geom.Circle a round LineString or a round Polygon? Should circle.intersectsExtent return true or false for an extent completely contained by the circle? Should ol.source.Vector#getFeaturesAtCoordinate return true or false for a coordinate within the circle? The expected behaviour is probably depends on the application, and the style of the circle (filled->true, non-filled->false).

alvinlindstam:
I created a pull request for ol.geom.Circle#intersectsExtent here: https://github.com/openlayers/ol3/pull/3764

marcjansen:
I commented https://github.com/openlayers/ol3/issues/3703#issuecomment-109248047 and would wait from input of the @-mentioned people.

jonataswalker:
While this is not merged, ol.source.Cluster.getFeaturesAtCoordinate() is completely unusable. As mentioned, [ol.geom.Geometry.prototype.containsCoordinate](https://github.com/openlayers/ol3/blob/aa90c97ba3ffd59a0fba99669cce8702ed790d09/src/ol/geom/geometry.js#L134) always returns false.

ahocevar:
I'm not sure how to proceed with this, because the relevant information is scattered across too many places. I'm closing this due to lack of activity, and would suggest a new ticket that summarises the problems and potential solutions.

PR #3763, opened 2015-06-05
Standardise draw/modify descriptions

probins:
Modify at the moment might mislead, as it implies that it can also modify feature attributes. I was going to change select interaction as well ('Interaction for selecting vector features'), but that would conflict with #3758, so I'll leave that for the moment.


ahocevar:
Thanks @probins!

PR #3774, opened 2015-06-08
Measure tooltips touchdevice

### LINKED ISSUE #3370: Measure example not working well on mobile
The tooltip "Click tho start drawing" does not appear on the screen and when you zoom out it appears, but not at the good place.

Also, when you click on the screen, the measurement is not displayed.

Tested with a Nexus 5 via the following url:
http://openlayers.org/en/v3.3.0/examples/measure.html

pgiraud:
This pull requests makes it so the measure are displayed on touch devices as well. The measurement is displayed when the drawn geometry changes ie. when a new point is added on touch devices.

Also, the help tooltip is hidden when the cursor moves out of the map which prevents it from being displayed in a weird place sometimes.

Fixes #3370.
There's still a problem when user taps the zoomout button on a touch device.

Please review.

### DIFF
--- a/examples/measure.js
+++ b/examples/measure.js
@@ -103,27 +103,19 @@ var pointerMoveHandler = function(evt) {
   }
   /** @type {string} */
   var helpMsg = 'Click to start drawing';
-  /** @type {ol.Coordinate|undefined} */
-  var tooltipCoord = evt.coordinate;
 
   if (sketch) {
-    var output;
     var geom = (sketch.getGeometry());
     if (geom instanceof ol.geom.Polygon) {
-      output = formatArea(/** @type {ol.geom.Polygon} */ (geom));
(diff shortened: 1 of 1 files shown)


examples/measure.js
```diff
@@ -138,6 +132,10 @@ var map = new ol.Map({
 
 map.on('pointermove', pointerMoveHandler);
 
+map.on('mouseout', function() {
```
tschaub: Did you mean to use `jQuery.mouseleave()` here?  The map doesn't fire a `mouseout` event.

pgiraud: Indeed. I wanted to observe `mouseout` on `$(map.getViewport())` actually.
Good catch.

fredj:
LGTM

PR #3775, opened 2015-06-09
Add ol-touch but keep ol-viewport className.

pgiraud:
Currently when displaying a map on a touch device the `ol-touch` className is added to the viewport. Well this className actually replaces `ol-viewport`.

In my application I rely on `ol-viewport` to add styling behavior to element within the viewport and I don't want to add specific css selectors for touch devices.

With this pull request `ol-touch` is effectively added instead of replacing the existing one.

I've tested manually some examples and haven't seen any weird behavior.

Please review.


fredj:
LGTM

elemoine:
Looks good to me too.

PR #3780, opened 2015-06-10
Only expose transformed tile coordinates to the API

ahocevar:
Previously, the `tileUrlFunction` had to transform the internal bottom-left origin tile coordinates to the tile coordinates that the source uses in its tile URLs. The tile coordinates returned from `ol.tilegrid.TileGrid#getTileCoordForCoordAndResolution()` and `ol.tilegrid.TileGrid#getTileCoordForCoordAndZ()` had to be transformed by the application as well to match them with the ones used in tile URLs.

After this change, the `tileUrlFunction` is called with transformed tile coordinates. Accordingly, `ol.tilegrid.TileGrid#getTileCoordForCoordAndResolution()` and `ol.tilegrid.TileGrid#getTileCoordForCoordAndZ()` return transformed tile coordinates. `ol.source.TileDebug` displays transformed tile coordinates.

With these changes, it is no longer necessary for application developers to understand the internal tiling scheme. Custom tile url functions are much easier to write, as can be seen in the updated [xyz-esri-4326-512.js](https://github.com/openlayers/ol3/pull/3780/files#diff-03aee0cf3bbb96d26dcec4ea2a2b9665) example.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -34,6 +34,53 @@ With the removal of `ol.FeatureOverlay`, `zIndex` symbolizer properties of overl
 
 Note that `ol.FeatureOverlay#getFeatures()` returned an `{ol.Collection.<ol.Feature>}`, whereas `ol.source.Vector#getFeatures()` returns an `{Array.<ol.Feature>}`.
 
+#### `ol.TileCoord` changes
+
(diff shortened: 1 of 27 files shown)


ahocevar:
This is now ready to review. I split the pull request into several commits to make it easier for reviewers to understand the changes.

bartvde:
LGTM @ahocevar

PR #3782, opened 2015-06-10
Fix assert documentation typo

gberaudo:


ahocevar:
Good catch. Thanks @gberaudo.

PR #3783, opened 2015-06-10
Fix broken wmts-hidpi example

ahocevar:
The service switched from jpeg to png, so no tiles are shown any more.

The best way to fix this example is to use a WMTS GetCapabilities request, so changes won't break the example again.


fredj:
+1

PR #3784, opened 2015-06-10
Always write the GeoJSONFeature geometry property

fredj:
From http://geojson.org/geojson-spec.html#feature-objects:

*A feature object must have a member with the name "geometry". The value of the geometry member is a geometry object as defined above or a JSON null value.*

### DIFF
--- a/externs/geojson.js
+++ b/externs/geojson.js
@@ -123,7 +123,7 @@ var GeoJSONFeature = function() {};
 
 
 /**
- * @type {GeoJSONGeometry}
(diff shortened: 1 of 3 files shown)


marcjansen:
Can you add a test, @fredj?

fredj:
tests added

bartvde:
LGTM

fredj:
Thanks @marcjansen and @bartvde

marcjansen:
Thanks, @fredj :+1:

PR #3787, opened 2015-06-11
Add 'unmanaged' to ol.layer.LayerState

ahocevar:


src/ol/layer/layerbase.js
```diff
@@ -35,6 +35,7 @@ ol.layer.LayerProperty = {
  *            saturation: number,
  *            sourceState: ol.source.State,
  *            visible: boolean,
+ *            unmanaged: (boolean|undefined),
```
elemoine: We could also decide to always/systematically define it, but it's ok like this as well.

elemoine:
I added a comment, but please merge with or without taking this comment into account.

ahocevar:
@elemoine I changed the commit. Is this what you had in mind?

elemoine:
Yes!

marcjansen:
Would 'managed' be a better name? I've had bad experience most of the time when naming stuff with the negated variant

ahocevar:
I like the idea @marcjansen  - if you're willing to create a PR :smile:

marcjansen:
My bad, I had better kept my mouth shut :wink: I MAY have time tomorrow, for a PR

ahocevar:
@marcjansen How about I create a PR and you review - ETA 10 min.

marcjansen:
Ok, promised

ahocevar:
@marcjansen #3793. Thanks.

tschaub:
It would be great to have a bit of description here.  The diff clearly does what the title says, but some additional motivation or context would be useful for those reading a changelog.  Thanks!

tschaub:
I see https://github.com/openlayers/ol3/pull/3758#issuecomment-111031177 now.  Just catching up on some significant changes!  Still, it would be nice to avoid descriptionless tickets.

PR #3790, opened 2015-06-11
Remove unnecessary quotes around object keys

fredj:
The objects are typed, no need to quote the keys.


bartvde:
LGTM

fredj:
FYI, `build/ol.js` is the same before and after this change

elemoine:
LGTM too. Thanks.

PR #3791, opened 2015-06-11
Remove docs referring to removed feature overlay

marcjansen:
This is a follow-up PR of #3758 that removes now outdated and wrong docs.

/cc @ahocevar


ahocevar:
Thanks for catching this. Looks good.

PR #3792, opened 2015-06-11
Link to correct layer base class

marcjansen:
Two inline links were pointing at `ol.layer.Layer`, when actually `ol.layer.Base` is allowed.


ahocevar:
Good catch. Please merge.

PR #3793, opened 2015-06-11
Use 'managed' instead of 'unmanaged' in LayerState

ahocevar:


marcjansen:
Now looking into this.

marcjansen:
Very nice and exactly what I meant. Also exactly on time :wink:

Please merge once travis agrees with me.

PR #3794, opened 2015-06-11
VectorLayer and wrapX=true : Add support of points, polygons and lines over 180° (date line)

jcarbou:
First sorry for imprefection, this is my first pull request....

When wrapX option set to true for VectorLayer, add support of :
- polygons, lines, points over date 180° (date line)
- large polygons or lines

image

Large polygons or lines who intersect the left AND the right of the view extent

image


ahocevar:
Welcome, and thanks for your contribution @jcarbou.

Can you please resolve the conflicts with your pull request so Travis can run the CI tests on it?

The use case is not 100% clear to me yet. Can you please explain what currently does not work and what your pull request aims to improve?

jcarbou:
For the merge and CI tests, I will try to do that....

For Use cases 1 "Polygon over 180°" 

Below a example of the current behaviour with 3.6.0 version (Draw example with wrapX option set to true on VectorLayer) :
image

With my proposition the polygon is'nt crop (see screenshot of my first post)

For Use case 2 "Large polygon". This use case is strange but real. When you draw  a large polygon (width=160°), and the map view width is 300° for example, you can move the map to have one piece of the same polygon on the left of the map, and another piece on the right. The current behaviour draw one piece. With my proposition the two pieces are drawn :

image

ahocevar:
Thanks for the explanation @jcarbou, it is now clear to me. Once the CI tests are passed, I'll review and test.

marcjansen:
Plus one for getting this in! Thanks so far...

jcarbou:
I merge changes. I validate with other examples :

Vector Icon Example :
image

Modify Feature example :
image

I don't know "CI tests" but I will search....

jcarbou:
I find a bug with drawinteraction for polygon creation. 

Use case 1 :
- Draw a first polygon over 0° meridian => OK
- Draw a second polygon over 180° meridian => draw OK but after edition the polygon desapear
- You can draw a lot of polygons correctly if they are in [-180°,180°]
- All polygons over 180° desapear.

Use case 2 :
- Draw a first polygon over 180° meridian => OK
- Draw a second polygon over 0° meridian => draw OK but after edition the polygon desapear
- You can draw a lot of polygons correctly if they are in [0°,360°]
- All polygons over 0° desapear.

I think my pull request don't generate this bug but I haven't found time to validate that ! 

Sorry

ahocevar:
The reason why we recommend to use `wrapX: false` for editing across the date line is that you get coordinates that exceed the world bounds. This is the only way to tell most geospatial appliactions and servers that you want your edge drawn across the data line, and not across the world.

The behavior you are seeing in both your use cases is related to this. For your pull request to work, you would have to include geometries on both sides of the date line when you draw the other side, potentially resulting in duplicate geometries rendered. Also not that OpenLayers uses a coordinate range of -180 to 180, not 0 to 360.

jcarbou:
Hi ahocevar,

Thank you for explanation. I don't know all the best practices of ol3 and for example the respect the [-180,180]. It's a very good thing for code understanding and quality. 
My pull request does'nt respect that. For facility and  performance reasons,  my pull request draws geometries over 180° (or -180°) whitout splitting (one draw by geometry instead of two). To do that, for some situations, I disabled the canvascontext crop (crop to[ -180,180°]).
I can change this behaviour and it's simple for polygon and linestring. Indeed, the new code to support "large polygons" (see my first post) is very closed the code needed to "duplicate" the draw of geometries over date line. I have to implement this behaviour for point (image) and text.

Do you want I implements that in this pull request ?

ahocevar:
@jcarbou I think your pull request is fine the way it is. It's ok to not support all cases (wrapX with geometries that exceed the extent), and I do think that your pull request is an improvement even with the limitations you mentioned. So no need to implement splitting or geometry duplication.

I need to do some testing now with known corner cases. Please give me a few days to give your pull request the proper review that it deserves.

jcarbou:
Ok. Thank you !

ahocevar:
While reviewing this pull request, I kept wondering if there is a simpler solution. I may be missing something, but I came up with #3806. @jcarbou, can you please check if that use case also meets your requirements?

Jebjum:
I am having this same issue. I tried out #3806 and it does indeed allow for polys that cross the dateline to not be clipped. 

What I am wondering if it would also be possible to allow the draw interaction to cross the dateline as well. For example if I start drawing a polygon at -179 longitude and I cross over to + 179 longitude then the polygon line shoots all the way across the map to the right, rather than crossing the dateline to the left.

ahocevar:
> What I am wondering if it would also be possible to allow the draw interaction to cross the dateline as well. For example if I start drawing a polygon at -179 longitude and I cross over to + 179 longitude then the polygon line shoots all the way across the map to the right, rather than crossing the dateline to the left.

@Jebjum Did you manipulate the draw interaction's coordinates in some way? What you are describing should result in a polygon with x coordinates -179 to -181. This is the only way to tell the renderer that the geometry is meant to cross the date line, whereas x coordinates -179 to +179 actually mean to draw a segment across the whole world.

Jebjum:
@ahocevar  You were right the coordinates were being modified by my application prior to the polygon being rendered - thanks!

Now I just need to figure out how to take the polygon that crosses the dateline and turn it into a MultiPolygon whose longitude coordinates are all between -180 and 180 so that my server knows how to query within that region...

ahocevar:
Fixed with #3806.

PR #3538, opened 2015-04-11
multiline labels

mamayoleksandr:
Support for multiline labels using '\n'

PR #3797, opened 2015-06-12
Support for Multi-line Labels

Morgul:
This is a cleaned up version of #3538.

The changes in this are:
* Cleaned up the code formatting
* Added a small padding between lines for readability

I'm unsure if you want any unit tests or example changes for this. If so, I'm happy to add them.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -387,44 +387,23 @@ ol.render.canvas.Replay.prototype.replay_ = function(
                 goog.vec.Mat4.getElement(localTransform, 0, 3),
                 goog.vec.Mat4.getElement(localTransform, 1, 3));
           }
-          /*
-          if (stroke) {
-            context.strokeText(text, x, y);
-          }
-          if (fill) {
-            context.fillText(text, x, y);
-          }
-          if (scale != 1 || rotation !== 0) {
-            context.setTransform(1, 0, 0, 1, 0, 0);
-          }
-          */
 
           var fontSize = parseFloat(context.font);
-          var tabText = text.split('\n');
-          var cptText = tabText.length;
+          var lines = text.split('\n');
+          var numLines = lines.length;
+          var lineY = y - (((numLines - 1) / 2) * fontSize);
 
-          var newY = y;
-
-          newY = newY - (((cptText - 1) / 2) * fontSize);
-
-          for (var iTxt = 0; iTxt < cptText; iTxt++) {
-            var newText = tabText[iTxt];
+          for (var lineIndex = 0; lineIndex < numLines; lineIndex++) {
+            var line = lines[lineIndex];
             if (stroke) {
-              if (iTxt === 0) {
-                context.strokeText(newText, x, newY);
-              } else {
-                context.strokeText(newText, x, newY);
-              }
+              context.strokeText(line, x, lineY);
             }
             if (fill) {
-              if (iTxt === 0) {
-                context.fillText(newText, x, newY);
-              } else {
-                context.fillText(newText, x, newY);
-              }
+              context.fillText(line, x, lineY);
             }
+
             // Add a small padding when calculating the new Y size.
-            newY = newY + fontSize + 5;
+            lineY = lineY + fontSize + 5;
           }
         }
         ++i;
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -423,7 +423,8 @@ ol.render.canvas.Replay.prototype.replay_ = function(
                 context.fillText(newText, x, newY);
               }
             }
-            newY = newY + fontSize;
+            // Add a small padding when calculating the new Y size.
+            newY = newY + fontSize + 5;
           }
         }
         ++i;
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -387,6 +387,7 @@ ol.render.canvas.Replay.prototype.replay_ = function(
                 goog.vec.Mat4.getElement(localTransform, 0, 3),
                 goog.vec.Mat4.getElement(localTransform, 1, 3));
           }
+          /*
           if (stroke) {
             context.strokeText(text, x, y);
           }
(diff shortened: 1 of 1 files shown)


src/ol/render/canvas/canvasreplay.js
```diff
@@ -387,6 +387,7 @@ ol.render.canvas.Replay.prototype.replay_ = function(
                 goog.vec.Mat4.getElement(localTransform, 0, 3),
                 goog.vec.Mat4.getElement(localTransform, 1, 3));
           }
+          /*
```
tschaub: We shouldn't leave commented code blocks in the library.  They are sources of mystery to future readers.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -396,6 +397,35 @@ ol.render.canvas.Replay.prototype.replay_ = function(
...
           }
+          */
+
+          var fontSize = parseFloat(context.font);
+          var tabText = text.split('\n');
+          var cptText = tabText.length;
+
+          var newY = y;
+
+          newY = newY - (((cptText - 1) / 2) * fontSize);
```
tschaub: I think the above would be more legible with something like this:
```js
var fontSize = parseFloat(context.font);
var lines = text.split('\n');
var numLines = lines.length;
var lineY = y - (((numLines - 1) / 2) * fontSize);
```

And then I'd probably use `lineIndex` (instead of `iTxt`) and `line` (instead of `newText`) below.

The compiler makes names small for us.  So it's nice to make them legible in the source.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -396,6 +397,35 @@ ol.render.canvas.Replay.prototype.replay_ = function(
...
+
+          for (var iTxt = 0; iTxt < cptText; iTxt++) {
+            var newText = tabText[iTxt];
+            if (stroke) {
+              if (iTxt === 0) {
+                context.strokeText(newText, x, newY);
+              } else {
+                context.strokeText(newText, x, newY);
+              }
```
tschaub: The above two conditions run the same code, right?

src/ol/render/canvas/canvasreplay.js
```diff
@@ -396,6 +397,35 @@ ol.render.canvas.Replay.prototype.replay_ = function(
...
+            }
+            if (fill) {
+              if (iTxt === 0) {
+                context.fillText(newText, x, newY);
+              } else {
+                context.fillText(newText, x, newY);
+              }
```
tschaub: Same as above.

Morgul:
@tsauerwein So, I made an attempt at using `CanvasRenderingContext2D.measureText()`, however all properties other than `width` are not supported by most browsers. (see [here](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics#Browser_compatibility))

I can attempt to check for those properties and fall back to the font size, if that's what you want, but right now it will only work for chrome users who enable a flag.

tsauerwein:
@Morgul Thanks for investigating this. I wasn't aware of that there is no `height` property in the text metrics. There are [work-arounds](http://videlais.com/2014/03/16/the-many-and-varied-problems-with-measuring-font-height-for-html5-canvas/) to calculate the height, but does your current approximation give good results?

Morgul:
@tsauerwein It does. (We're currently using my patches internally, and so far everything's worked out very well.

Morgul:
@tsauerwein, @tschaub: Calculating the text height is a very painful proposition, it seems. The code from the blog post mentioned [here](https://gist.github.com/videlais/9589285#file-determinefontheightinpixels-js) seems like it could be adapted, but do we want this overhead? It's very complex, and has it's own set of caveats.

Alternatively, we could code in the simpler approximation of taking width of "M" as the height, and multiplying by 1.5 to get a 'rough' guess. (This would also get rid of the hard-coded 5px padding I add.) I can work up that code and see if it gives a good enough approximation.

jean-clovis:
Hi all,
any chance to see this fix in the next release soon ? 
thanks :)

ahocevar:
> Alternatively, I could code in the simpler approximation of taking width of "M" as the height, and multiplying by 1.5 to get a 'rough' guess.

@Morgul, I like this approach. Would be great if you could give it a try.

ahocevar:
> any chance to see this fix in the next release soon ?

@jean-clovis Probably too late for v3.10, but likely in v3.11 in a month.

Morgul:
@ahocevar I will attempt to get back around to this shortly. We're rolling out our new product at work, so the next week or two is a little crazy for me.

tschaub:
Does everybody feel ok about this going in as it is?  We can add enhancements in subsequent releases.

ahocevar:
Merging this as-is sounds good to me.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -387,14 +387,23 @@ ol.render.canvas.Replay.prototype.replay_ = function(
...
                 goog.vec.Mat4.getElement(localTransform, 1, 3));
           }
-          if (stroke) {
-            context.strokeText(text, x, y);
-          }
-          if (fill) {
-            context.fillText(text, x, y);
-          }
-          if (scale != 1 || rotation !== 0) {
-            context.setTransform(1, 0, 0, 1, 0, 0);
```
tsauerwein: Any reason that this statement was removed?

tsauerwein:
I added a question, but I agree that improvements like properly calculating the height can be addressed in separate PRs.

src/ol/render/canvas/canvasreplay.js
marcjansen: I think that was a mistake

tschaub:
It looks like this still needs some work.  Here is the vector labels example with and without this change:

labels

It would be great to get multi-line label support in.  If someone can fix it up, we should be able to have it in the next release.

tschaub:
Let's make sure to improve our rendering test coverage when getting this merged.  If the difference I'm seeing in the example is real, it would be nice to have a failing test.

tsauerwein:
Closing now that #4512 is merged. @Morgul, thanks for your initial work!

PR #3800, opened 2015-06-13
Remove further references to FeatureOverlay

probins:
while I'm at it, I've also added a link in the docs to the default style


marcjansen:
Nice. @ahocevar, further comments?

ahocevar:
Thanks @probins, looks good.

PR #3801, opened 2015-06-15
Respect the tile grid's extent in ol.source.TileVector

### LINKED ISSUE #3798: Tile grid extent settings causes creating lot of useless requests
In my 'app' http://drifted.in/other/ol/tiles_36_extent.html there is the 'extent' specified on vector tile grid to prevent loading vector tiles outside a valid area. But this settings causes creating thousands of requests. Like all my vector data for all tiles in all zoom levels are required.

While the script in Firefox can be interrupted, Chrome consumes almost all PC memory and must be killed.

ahocevar:
By adding a `getTileCoordForTileUrlFuction()` method like for `ol.source.Tile`, we can now properly handle extent and resolution restrictions, and reuse tiles on wrapped worlds. Also adds the missing `wrapX` option to `ol.source.TileVector`.

Fixes #3798.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -4045,7 +4045,8 @@ olx.source.TileImageOptions.prototype.wrapX;
  *     tileGrid: ol.tilegrid.TileGrid,
  *     tileUrlFunction: (ol.TileUrlFunctionType|undefined),
  *     url: (string|undefined),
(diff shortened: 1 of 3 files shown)


bartvde:
LGTM

PR #3804, opened 2015-06-15
Remove sentence that was only meant for WMTS tile grids

ahocevar:
This pull request fixes the documentation of the `extent` config option of `ol.tilegrid.TileGrid`. The reference to the `sizes` should not be there, because they are not part of the API. `sizes` is only exposed for `ol.tilegrid.WMTS`.


elemoine:
I don't see the relationship between the PR description and the commit content.

ahocevar:
Oh, thanks @elemoine. Committed the wrong change. Fixed now.

elemoine:
LGTM.

PR #3810, opened 2015-06-16
Improve TileGrid documentation and examples

ahocevar:
The term 'bottom-left corner' in the docs is misleading. This pull request aims to improve the documentation. It also simplifies the `xyz-esri-4326-512` example by actually using the `tileSize` variable.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -6149,8 +6149,8 @@ olx.tilegrid.TileGridOptions.prototype.minZoom;
 
 /**
  * The tile grid origin, i.e. where the `x` and `y` axes meet (`[z, 0, 0]`).
- * Tile coordinates increase from left to right and from bottom to top. Default
- * is null.
+ * Tile coordinates increase from left to right and from bottom to top. If not
+ * specified, `extent` or `origins` must be provided.
  * @type {ol.Coordinate|undefined}
  * @api stable
  */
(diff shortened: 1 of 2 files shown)


externs/olx.js
```diff
@@ -6148,7 +6148,9 @@ olx.tilegrid.TileGridOptions.prototype.minZoom;
 
 
 /**
- * Origin, i.e. the bottom-left corner of the grid. Default is null.
+ * Origin of the grid. The tile coordinate's `x` and `y` are `0` at the origin.
```
probins: how about 'The grid origin: where the `x` and `y` axes meet ([0,0]).' ?

probins:
I was wondering about 'bottom left' in #3738 - or is a urlfunction only used when coords are +ive?

ahocevar:
@probins The changes from #3738 have been removed again, they are no longer relevent. The TileUrlFunction is now always called with transformed tile coordinates, as defined by the internal transformTileCoord method of the tileGrid.

externs/olx.js
ahocevar: I didn't want to use [0,0], because tile coordinates are [z, x, y].

ahocevar: I think I found a good solution now @probins.

probins:
might also be worth adding a note that if 'origin' is not defined, the extent bottom-left is used - or just 'see extent'.

ahocevar:
Good suggestion @probins. Added.

marcjansen:
Very nice. Please merge.

elemoine:
Good change.

PR #3818, opened 2015-06-18
Make geometry.transform api stable again

probins:
As mentioned in https://github.com/openlayers/ol3/pull/3464#issuecomment-113155115 I'm assuming that the api stable tag on geom.Geometry.transform was removed by mistake. Only Circle implements this function, so if you try and generate an export for this method for other geometries you get an error. Putting this back should fix this, and mean the function appears in the api docs again as well.


tschaub:
Thanks @probins.

PR #3820, opened 2015-06-18
Make unmanaged vector layers behave more like ol.FeatureOverlay

### LINKED ISSUE #3819: Hover selection causes features to flicker
This was brought up in https://github.com/openlayers/ol3/pull/3758#issuecomment-113209937.
If you go to http://openlayers.org/en/master/examples/select-features.html and select the hover option, there's an annoying flickering as the cursor passes over the features. This doesn't occur with http://openlayers.org/en/v3.6.0/examples/select-features.html
Is it constantly adding/removing the feature from the selection layer?

ahocevar:
* Skipped features need to be hit-detected on unmanaged layers.
* updateWhileAnimating and updateWhileInteracting are recommended to achieve the same instant visual feedback that ol.FeatureOverlay had.

Fixes #3819.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -20,11 +20,13 @@ you will have to change this to:
 var collection = new ol.Collection();
 var featureOverlay = new ol.layer.Vector({
   map: map,
-  style: overlayStyle,
   source: new ol.source.Vector({
     features: collection,
     useSpatialIndex: false // optional, might improve performance
-  });
+  }),
+  style: overlayStyle,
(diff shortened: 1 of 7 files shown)


probins:
if these are new options, they need to be added to olx.js don't they?

ahocevar:
These options are not new. They are in 0LX.JS already.

probins:
on the layer. Your example in upgrade-notes has them on the source

ahocevar:
Thanks @probins, good catch again. I updated the pull request.

elemoine:
Please merge.

PR #3482, opened 2015-04-04
ol.View#fit()

malaretv:
This adds an ol.View#fit() that accepts either a geometry or an extent.

This combines two previously distinct functions -- ol.View#fitGeometry and ol.View#fitExtent
into a single more flexible call. Rotation support and options previously only available to fitGeometry can now be used when fitting extents.

Also ol.View#calculateExtent() now takes into account rotation state.

PR #3821, opened 2015-06-18
Allow ol.View.fitExtent to accept options for resolution constraining

patrickbr:
For the user, fitExtent and fitGeometry are very similar. At the moment,
fitExtent does not accept any options while fitGeometry() accepts
parameters like padding, minResolution or constrainResolution.
Argueably, a padding option is not really of any value here because
ol.Extent can be padded easily before calling fitExtent(). Also, a
minResolution/maxZoom option doesn't seem relevant for a function that
is specifically called to fit the map into the specified extent.

This commit however introduces 'constrainResolution' and 'nearest'
options for fitExtent. The resolution constraint is actually preventing
fitExtent() from fitting the map exactly into the provided extent. It
should be possible to turn this behavior off, because otherwise
ol.View.fitExtent  is unable to do what its name suggests.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -6389,6 +6389,30 @@ olx.tilegrid.ZoomifyOptions.prototype.resolutions;
  */
 olx.view;
 
+/**
+ * @typedef {{padding: !Array.<number>,
+ *     constrainResolution: (boolean|undefined),
+ *     nearest: (boolean|undefined)}}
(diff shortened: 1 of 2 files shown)


bartvde:
are you aware of #3482 ?

patrickbr:
I was indeed not. #3482 is the better solution.

PR #3822, opened 2015-06-19
Correct docs for updateWhileInteracting

probins:
at least, I assume this is what this should say :-)


ahocevar:
Good catch, thanks!

PR #3824, opened 2015-06-19
Improve docs for interaction.Select

probins:
Suggestion for expanding the current docs. I've left out the mention of the vector source, as this is purely internal and doesn't affect the api or how it works.


tschaub:
Thanks @probins.

PR #3826, opened 2015-06-19
Allow custom tileGrid in ol.source.XYZ

petrsloup:
It is required for certain dataset to be able to pass custom tileGrid to `ol.source.XYZ` constructor.
Mainly because the extent of the dataset is only a subset of the projection extent.

Creating viewer for such XYZ tiles was possible in OL2: https://7a21d5a95f2b008cad7a4e1c15217921ef62368c.googledrive.com/host/0BwUe1kqIRJWSfllBSnJTT3pZM1RHdHBzaVBnQ29wQUNZWFg2Uy1mWVRjd3NvWlU1TG9MUEE/openlayers.html, but it is very hard in OL3 now.

This PR adds `tileGrid` constructor option to `ol.source.XYZ` (the default behavior is unchanged).

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5358,6 +5358,7 @@ olx.source.WMTSOptions.prototype.wrapX;
  *     projection: ol.proj.ProjectionLike,
  *     maxZoom: (number|undefined),
  *     minZoom: (number|undefined),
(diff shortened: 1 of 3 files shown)


ahocevar:
Very nice, thanks!

PR #3829, opened 2015-06-20
incorrect assert message

kzr-pzr:
I stumbled across this assert message when playing with OL3. When I looked into OL3 source I noticed the message is in fact incorrect: assert checks for a string, message says a number. Probably a result of a copy-pasting.


elemoine:
Thanks

Issue #3754, opened 2015-06-01
Public format API to enable custom formats with external code


sweco-sebhar:
As far as I can tell a custom format implementation currently needs to be written in closure style and compiled together with ol3. I think it would be desirable to expose APIs to be able to create custom formats in external code.

When looking at the format internals I also wonder if there perhaps is a simplier API similar to the approach that was taken with the new Vector API. That is, make a format API that is more functional instead of the the current classical OOP design. An interface for a format could simply be a function signature with arbitary data as input and feature array as output.

sweco-sebhar:
Today I realised what I want is already exists for untiled Vector sources. Simply implement a custom ol.FeatureLoader and use it with a Vector source. For TileVector sources there is no loader option, I'll create a PR to introduce that. Closing this.

PR #3833, opened 2015-06-22
Enable use of custom XHR loader for TileVector sources

sweco-sebhar:
Currently it's possible to fully customize only the plain Vector source. The TileVector source forces the use of the built in XHR implementation which in turn is coupled to the Format API which is not extensible outside OpenLayers.

The use case is when one wants to use external custom data sources with custom format (see #3754 and #3756) without modifying and custom building OpenLayers.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -4022,10 +4022,11 @@ olx.source.TileImageOptions.prototype.wrapX;
 
 /**
  * @typedef {{attributions: (Array.<ol.Attribution>|undefined),
- *     format: ol.format.Feature,
+ *     format: (ol.format.Feature|undefined),
  *     logo: (string|olx.LogoOptions|undefined),
  *     tileGrid: ol.tilegrid.TileGrid,
  *     tileUrlFunction: (ol.TileUrlFunctionType|undefined),
+ *     tileLoadFunction: (ol.TileVectorLoadFunctionType|undefined),
  *     url: (string|undefined),
  *     urls: (Array.<string>|undefined),
  *     wrapX: (boolean|undefined)}}
@@ -4043,8 +4044,8 @@ olx.source.TileVectorOptions.prototype.attributions;
 
 
 /**
- * Format.
- * @type {ol.format.Feature}
+ * Format. Required unless tileLoadFunction is used.
+ * @type {ol.format.Feature|undefined}
  * @api
  */
 olx.source.TileVectorOptions.prototype.format;
@@ -4073,6 +4074,16 @@ olx.source.TileVectorOptions.prototype.tileGrid;
  * @api
  */
 olx.source.TileVectorOptions.prototype.tileUrlFunction;
+
+
+/**
+ * Optional function to override the default loading and format parsing behaviour.
+ * If this option is used format is ignored and the provided function will be
+ * responsible for data retrieval and transformation into features.
+ * @type {ol.TileVectorLoadFunctionType|undefined}
+ * @api
+ */
+olx.source.TileVectorOptions.prototype.tileLoadFunction;
 
 
 /**
--- a/src/ol/source/tilevectorsource.js
+++ b/src/ol/source/tilevectorsource.js
@@ -34,12 +34,9 @@ ol.source.TileVector = function(options) {
 
   /**
    * @private
-   * @type {ol.format.Feature}
+   * @type {ol.format.Feature|undefined}
    */
-  this.format_ = options.format;
-
-  goog.asserts.assert(goog.isDefAndNotNull(this.format_),
-      'ol.source.TileVector requires a format');
+  this.format_ = goog.isDef(options.format) ? options.format : null;
 
   /**
    * @private
@@ -52,6 +49,17 @@ ol.source.TileVector = function(options) {
    * @type {ol.TileUrlFunctionType}
    */
   this.tileUrlFunction_ = ol.TileUrlFunction.nullTileUrlFunction;
+
+  /**
+   * @private
+   * @type {?ol.TileVectorLoadFunctionType}
+   */
+  this.tileLoadFunction_ = goog.isDef(options.tileLoadFunction) ?
+      options.tileLoadFunction : null;
+
+  goog.asserts.assert(!goog.isNull(this.format_) ||
+      !goog.isNull(this.tileLoadFunction_),
+      'Either format or tileLoadFunction are required');
 
   /**
    * @private
@@ -299,9 +307,14 @@ ol.source.TileVector.prototype.loadFeatures =
             tileUrlFunction(urlTileCoord, 1, projection);
         if (goog.isDef(url)) {
           tiles[tileKey] = [];
-          var loader = ol.featureloader.loadFeaturesXhr(url, this.format_,
-              goog.partial(success, tileKey));
-          loader.call(this, extent, resolution, projection);
+          var tileSuccess = goog.partial(success, tileKey);
+          if (!goog.isNull(this.tileLoadFunction_)) {
+            this.tileLoadFunction_(url, tileSuccess);
+          } else {
+            var loader = ol.featureloader.loadFeaturesXhr(url,
+                /** @type {ol.format.Feature} */ (this.format_), tileSuccess);
+            loader.call(this, extent, resolution, projection);
+          }
         }
       }
     }
--- a/src/ol/tileloadfunction.js
+++ b/src/ol/tileloadfunction.js
@@ -1,4 +1,5 @@
 goog.provide('ol.TileLoadFunctionType');
+goog.provide('ol.TileVectorLoadFunctionType');
 
 
 /**
@@ -9,3 +10,13 @@ goog.provide('ol.TileLoadFunctionType');
  * @api
  */
 ol.TileLoadFunctionType;
+
+
+/**
+ * A function that is called with a tile url for the features to load and
+ * a callback that takes the loaded features as argument.
+ *
+ * @typedef {function(string, function(Array.<ol.Feature>))}
+ * @api
+ */
+ol.TileVectorLoadFunctionType;


ahocevar:
I have concerns about this from an API perspective. Wouldn't it be nicer to have a TileLoadFunction like for tile sources?

probins:
I have been meaning to raise a similar issue. The process for loading vector sources currently is:

1. fetch data
2. convert data to features
3. add features to source/layer

But I have several problems with how this currently works:
* no ability to use data that is not in a ol-supported format (same problem as in this ticket)
* not able to add features using the loader, common when reading from database: fetch `?type=restaurant` initially, then add `?type=hotel`
* not able to write back to the same server, as the url/format/projection are not saved

I can do this with a custom loader and by storing the url/format/projection in my app, but if I do this for some sources, I might as well do it for all, in which case the ol-supplied loader isn't used.

probins:
plus another use case is to add features loaded from a different source; again I can do this with a custom loader, but the basic issue is that ol.source assumes a one-time read-only source, which isn't always the case with vectors.

sweco-sebhar:
@ahocevar, I agree your proposal looks more clean and appropriate. I'll rewrite this PR to match when possible.

@probins, what other sources than Vector (already has custom loader) and TileVector do you think are relevant?

probins:
ISTM it would be better to pass a convert-data function to the loader, which could be format.readFeatures or could be any other conversion. It isn't the loading that needs to be customised; it's the conversion into features. These aren't necessarily 'custom formats'. You can easily use for example a MongoDB or Google Fusion Tables for storing features in the cloud. The geometries here are GeoJSON but they are not GeoJSON feature collections, and have to be rearranged to create OL features. I've been experimenting with Geobuf, which is similarly based on GeoJSON but not readable with format.GeoJSON.

sweco-sebhar:
Currently loading does need to be customised for binary formats (the current xhr-stuff in OpenLayers only supports text based xhr). Also, you might want to load from multiple sources (geometry from one source, attributes from another). So what I'm saying is that I think pure data conversion is too limited and would like to go with the proposal from @ahocevar.

probins:
the current xhr loader could be changed to handle non-text responses as well, no? I've not come across the need to load geometry and attributes from separate sources - agreed, you would need a custom loader for that :smiley:

As you say above, the loader is currently coupled with ol.format, and ISTM it would be better and more flexible if they were uncoupled. I should be able to pass any callback function to xhr to be run onload but before addFeatures.

ahocevar:
The nice thing about my proposal is that you don't even need xhr. You could also use JSONP or whatever. The app would be responsible for creating a features array (which is easy with the formats available, and possible with whatever your format is if you have application code for it), and once the features are available all the app needs to do is call a callback.

ahocevar:
@probins 
> I should be able to pass any callback function to xhr to be run onload but before addFeatures.
This is achieved with my proposal. Maybe you misread it?

probins:
@ahocevar Are you proposing something similar for source.Vector as well?

ahocevar:
@probins You can achieve the same with `ol.source.Vector`, by configuring a `loader` instead of a `url` and a `format`. Look at the `vector-osm` example to see how this is done.

The only difference is that you call `addFeatures` instead of a callback that is passed to a loader. So maybe both `ol.source.Vector` and `ol.source.TileVector` should work the same - either both with a callback, or both with manually calling `addFeatures`.

probins:
yes, that's what I was meaning - Vector and TileVector should work in the same (or very similar) way.

ahocevar:
Right. So we can either remove the tileUrlFunction option from ol.source.TileVector, or add a featureUrlFunction option to ol.source.Vector. I think the latter would be more convenient, and it would be more in line with what we have for image and tile sources.

sweco-sebhar:
So I've changed the implementation to add TileVectorLoadFunctionType like the proposal from @ahocevar. Can we keep this PR about that and add featureUrlFunction to ol.source.Vector in another?

ahocevar:
Sure, thanks @sweco-sebhar. I'll review the updated pull request tomorrow. ol.source.Vector can be changed separately of course.

externs/olx.js
```diff
@@ -4076,6 +4077,14 @@ olx.source.TileVectorOptions.prototype.tileUrlFunction;
 
 
 /**
+ * Optional function to customise loader and/or format behaviour.
+ * @type {ol.TileVectorLoadFunctionType|undefined}
+ * @api
+ */
+olx.source.TileVectorOptions.prototype.tileVectorLoadFunction;
```
ahocevar: The name of the option can be just `tileLoadFunction`.

src/ol/source/tilevectorsource.js
```diff
@@ -299,9 +305,14 @@ ol.source.TileVector.prototype.loadFeatures =
             tileUrlFunction(urlTileCoord, 1, projection);
         if (goog.isDef(url)) {
           tiles[tileKey] = [];
-          var loader = ol.featureloader.loadFeaturesXhr(url, this.format_,
-              goog.partial(success, tileKey));
-          loader.call(this, extent, resolution, projection);
+          var tileSuccess = goog.partial(success, tileKey);
+          if (goog.isDef(this.tileVectorLoadFunction_)) {
+            this.tileVectorLoadFunction_.call(this, url, tileSuccess);
```
ahocevar: Is there a reason to execute this with the source as `this` keyword? I don't think it is necessary.

sweco-sebhar: Thanks, removed.

sweco-sebhar:
Thanks for the review @ahocevar. Name changed to tileLoadFunctio and the constructor assertion changed to require format or tileLoadFunction.

src/ol/source/tilevectorsource.js
```diff
@@ -34,13 +34,10 @@ ol.source.TileVector = function(options) {
 
   /**
    * @private
-   * @type {ol.format.Feature}
+   * @type {ol.format.Feature|undefined}
    */
   this.format_ = options.format;
```
ahocevar: Should be
```js
this.format_ = goog.isDefAndNotNull(options.format) ? options.format : null;
```

src/ol/source/tilevectorsource.js
```diff
@@ -55,6 +52,16 @@ ol.source.TileVector = function(options) {
 
   /**
    * @private
+   * @type {ol.TileVectorLoadFunctionType|undefined}
+   */
+  this.tileLoadFunction_ = options.tileLoadFunction;
```
ahocevar: Should be
```js
this.tileLoadFunction_ = goog.isDef(options.tileLoadFunction) ? options.tileLoadFunction : null;
```

src/ol/source/tilevectorsource.js
```diff
@@ -55,6 +52,16 @@ ol.source.TileVector = function(options) {
...
+
+  goog.asserts.assert(goog.isDefAndNotNull(
+      this.format_ || this.tileLoadFunction_),
+      'ol.source.TileVector requires a format or a tileLoadFunction');
```
ahocevar: Should be something like
```js
goog.asserts.assert(!goog.isNull(this.format_) || !goog.isNull(this.tileLoadFunction),
    'Either format or tileLoadFunction are required`);
```

src/ol/source/tilevectorsource.js
```diff
@@ -299,9 +306,14 @@ ol.source.TileVector.prototype.loadFeatures =
             tileUrlFunction(urlTileCoord, 1, projection);
         if (goog.isDef(url)) {
           tiles[tileKey] = [];
-          var loader = ol.featureloader.loadFeaturesXhr(url, this.format_,
-              goog.partial(success, tileKey));
-          loader.call(this, extent, resolution, projection);
+          var tileSuccess = goog.partial(success, tileKey);
+          if (goog.isDef(this.tileLoadFunction_)) {
```
ahocevar: Should be
```js
if (!goog.isNull(this.tileLoadFunction_)) {
```

ahocevar:
Thanks for the updates @sweco-sebhar. I made a few more comments. When addressed, I'm going to merge this.

externs/olx.js
```diff
@@ -4076,6 +4077,14 @@ olx.source.TileVectorOptions.prototype.tileUrlFunction;
 
 
 /**
+ * Optional function to customise loader and/or format behaviour.
```
elemoine: I think it would be good to add a bit more context here.

sweco-sebhar:
Thanks for the input!

I've made changes from your comments @ahocevar. I had to introduce `ol.TileVectorLoadFunction.nullTileLoadFunction` to make closure compiler happy, similar to ol.TileUrlFunction.nullTileUrlFunction.

I've also tried to elaborate a bit more on the purpose of the `tileLoadFunction` option as pointed out by @elemoine.

ahocevar:
Your last change breaks the logic. Please consider 3bacbfb694c6539b7c938ab706da2351eb38a4a6.

sweco-sebhar:
Sorry I did not realize that it would break and I did not test. Picked your changes into a new sqashed commit.

sweco-sebhar:
Changed the externs to match the new type definition and have now tested locally by by modifying the tile-vector example.

sweco-sebhar:
`make ci` reveals further type issues which I do not understand yet but will try to fix.

ahocevar:
olx.js was correct already. There we want either undefined or the function. But for the member in ol.source.TileVector, we want null or the function. The type issues should go away when you change the type in olx.js back to `ol.TileVectorLoadFunctionType|undefined`.

A short unit test would be nice of course.

sweco-sebhar:
Yes came to the same conclusion and have reverted the changes. I'm lack the experience with closure to avoid the pitfalls. I'm not sure I can produce a unit test in a short time frame but will understand if you want it in place before merge.

ahocevar:
No it's ok. Your change is quite trivial in the end. Thanks for your work on it, I'm going to merge now.

ahocevar:
Quick note: I think it would be good if the TileVectorLoadFunction also received the projection as argument.

PR #3845, opened 2015-06-24
Prevent null array to be passed to an ol.Collection

fredj:
The compiler should be able to detect that:
```javascript
var collection = new ol.Collection(null);
```
or:
```javascript
collection.extend(null);
```
are invalid

### DIFF
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -221,8 +221,8 @@ ol.interaction.Select.handleEvent = function(mapBrowserEvent) {
   var set = !add && !remove && !toggle;
   var map = mapBrowserEvent.map;
   var features = this.featureOverlay_.getSource().getFeaturesCollection();
-  var /** @type {Array.<ol.Feature>} */ deselected = [];
-  var /** @type {Array.<ol.Feature>} */ selected = [];
+  var /** @type {!Array.<ol.Feature>} */ deselected = [];
+  var /** @type {!Array.<ol.Feature>} */ selected = [];
   var change = false;
   if (set) {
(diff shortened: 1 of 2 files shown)


src/ol/interaction/selectinteraction.js
```diff
@@ -221,8 +221,8 @@ ol.interaction.Select.handleEvent = function(mapBrowserEvent) {
   var set = !add && !remove && !toggle;
   var map = mapBrowserEvent.map;
   var features = this.featureOverlay_.getSource().getFeaturesCollection();
-  var /** @type {Array.<ol.Feature>} */ deselected = [];
-  var /** @type {Array.<ol.Feature>} */ selected = [];
+  var deselected = [];
+  var selected = [];
```
elemoine: Any specific reasons for removing the types here? Note that these were not typecasts.

elemoine:
LGTM. Just wondering about the removal of types in the select interaction.

fredj:
Without these types removal, the compiler considers that `{Array.<ol.Feature>}` may be null and returns an error:
```
ERR! compile found   : (Array<(null|ol.Feature)>|null)
ERR! compile required: Array<(null|ol.Feature)>
ERR! compile     features.extend(selected);
ERR! compile                     ^
ERR! compile 
```

fredj:
Instead of removing the type, I'll change it to `{!Array.<ol.Feature>}`

elemoine:
> Instead of removing the type, I'll change it to {!Array.<ol.Feature>}

Yes, I was going to suggest the same.

fredj:
Thanks for the review

PR #3849, opened 2015-06-25
Pad min. and sec. with leading zeros in DMS notation

### LINKED ISSUE #3776: DMS notation should use leading zeroes for minutes and seconds
http://openlayers.org/en/v3.6.0/apidoc/ol.coordinate.html#toStringHDMS

see: http://en.wikipedia.org/wiki/ISO_6709
```When minutes and seconds are less than ten, leading zeroes should be shown.```

pgiraud:
FIxes #3776


elemoine:
I restarted the build. Please merge when Travis is happy.

PR #3856, opened 2015-06-29
ol.source.TileVector(): bind success function of tileLoadFunction to source

### LINKED ISSUE #3851: Bug: using olsource.TileVector() with custom tileLoadFunction
Just noticed the tileLoadFunction which was added recently (#3833), exactly what I need to implement my loading indicator (#3746). I could implement it rather fast, but I found a small bug. The callback of the tileLoadingFunction is not bound to the tileSource, therefore instead of (copied the example from the pull request):
```js
function(url, callback) {
  var xhr = new XMLHttpRequest();
  xhr.open(url, true);
  xhr.onload = function() {
    callback(new ol.format.GeoJSON().readFeatures(xhr.responseText));
  }
  xhr.send('');
};
```

you need to do
```js
function(url, callback) {
  var xhr = new XMLHttpRequest();
  xhr.open(url, true);
  xhr.onload = function() {
    callback.call(my_tile_source, new ol.format.GeoJSON().readFeatures(xhr.responseText));
  }
  xhr.send('');
};
```

When using the function from above, I got an error "Uncaught TypeError: this.changed is not a function" (using ol-debug.js), because the success function was bound to Window.

Call `callback.call(my_tile_source, features)` is what the ol.featureLoader is doing.

plepe:
- fixes #3851

### DIFF
--- a/src/ol/source/tilevectorsource.js
+++ b/src/ol/source/tilevectorsource.js
@@ -309,7 +309,7 @@ ol.source.TileVector.prototype.loadFeatures =
           tiles[tileKey] = [];
           var tileSuccess = goog.partial(success, tileKey);
           if (!goog.isNull(this.tileLoadFunction_)) {
-            this.tileLoadFunction_(url, tileSuccess);
+            this.tileLoadFunction_(url, goog.bind(tileSuccess, this));
           } else {
             var loader = ol.featureloader.loadFeaturesXhr(url,
                 /** @type {ol.format.Feature} */ (this.format_), tileSuccess);


ahocevar:
Thanks for this fix. Can you please use `goog.bind(tileSuccess, this)` instead of `tileSuccess.bind(this)`? If you haven't already, please submit a [CLA](https://github.com/openlayers/CLA) so we can merge your pull request.

plepe:
I updated the commit accordingly and submitted the CLA.

ahocevar:
Thanks a lot for this contribution @plepe.

Issue #3830, opened 2015-06-21
ol.source.TileVector: getFeatures at resolution


plepe:
ol.source.TileVector has the following functions to query a list of features:
- getClosestFeatureToCoordinate()
- getFeatureById()
- getFeatures()
- getFeaturesAtCoordinate()
- getFeaturesAtCoordinateAndResolution()
- getFeaturesInExtent()

Why is there only getFeaturesAtCoordinateAndResolution() and not e.g. getFeaturesInExtentAndResolution()? I would think that it would make sense to add functions that filter for the resolution for the other types of functions too (at least I would need them). Or is there a function/property to query the resolution of a feature?

I think the following additional functions should be available:
- getClosestFeatureToCoordinateAtResolution()
- getFeatureByIdAtResolution()
- getFeaturesAtResolution()
- getFeaturesInExtentAtResolution()

I already created a pull request (#3745) to expose the forEachFeatureInExtentAtResolution() function, but there was no reaction?

Shall I try to implement the missing getFeatures...AtResolution() -functions (instead)?

PR #3857, opened 2015-06-29
ol.source.TileVector:  at resolution functions

plepe:
As I've written in #3830, I'm missing several ...AtResolution() functions:
- [ ] getClosestFeatureToCoordinateAtResolution()
- [ ] getFeatureByIdAtResolution()
- [x] getFeaturesAtResolution()
- [x] getFeaturesInExtentAtResolution()
also:
- [x] forEachFeatureAtResolution()
- [x] forEachFeatureInExtentAtResolution()

checked function names are already implemented by this pull request.

Now I'm a bit unsure how to proceed. I've seen that `getFeatureById()` and `getClosestFeatureToCoordinate()` use indexes. To also use these indexes for the ...AtResolution() functions, we should maybe add indexes per zoom level? Or shall I just iterate over all objects and ignore performance. The index could be added later on? The bounding box index could also be used by the forEachFeatureInExtentAtResolution() resp. getFeaturesInExtentAtResolution() functions.

If no one replies, I will just implement the iteration functions.

### DIFF
--- a/src/ol/source/tilevectorsource.js
+++ b/src/ol/source/tilevectorsource.js
@@ -149,6 +149,43 @@ ol.source.TileVector.prototype.forEachFeatureAtCoordinateAndResolution =
 
 
 /**
+ * Iterate through all features at the provided resolution, calling the
+ * callback with each feature.  If the callback returns a "truthy" value,
+ * iteration will stop and the function will return the same value.
+ *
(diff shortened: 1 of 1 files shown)


ahocevar:
Closing due to lack of activity. The proposed vector tile support (#4219) will make most of the suggested changes here obsolete.

PR #3806, opened 2015-06-15
Do not clip canvas for vector layers when wrapping the world

ahocevar:
Maybe an alternative to #3794, which does not have the [mentioned issues](https://github.com/openlayers/ol3/pull/3794#issuecomment-111882987).

cc @jcarbou

Fixes #3705.

PR #3858, opened 2015-06-30
Fix graticule on wrapped views

Jebjum:
Passing the vectorContext.extent_ instead of the frameState.extent to the createGraticule_ function allows the parallel and meridian linestrings to pass the ol.extent.intersects tests within addParallel_ and addMerdian_ when the graticule is being drawn on a wrapped view.

With this change the graticule will correctly draw on the last wrapped view that the user clicked.


Jebjum:
@ahocevar  - in the vein of #3867 and #3806 (i.e. other issues with wrapping the dateline) can this be considered for 3.7.0? The graticule is currently broken for wrapping otherwise.

ahocevar:
Milestone assigned, I hope I'll be able to review before the release is cut. It would be good if you could squash the commits into one.

ahocevar:
Sorry, but this change does not fix things on top of master. At least I cannot see any improvement when looking at the graticule example. Am I missing something?

bartvde:
I'm going to move this out of the 3.7 milestone since it's unlikely we'll hear back before I cut the release

Jebjum:
I will check again on latest master, but to reproduce the bug turn on graticule, click on a wrapped map, and zoom in until lines start disappearing - they start disappearing after about zoom level 5.

Jebjum:
Here is a fiddle showing how the graticule behaved in 3.6:

http://jsfiddle.net/cz4aaj2d/

Note that when you click on a wrapped map (e.g. -200 degrees longitude) and zoom in the grid lines (mostly the longitude lines) disappear. This is the problem that this pull request fixes. 

However...

Here is a fiddle showing how it behaves in master:

http://jsfiddle.net/pa71eeLo/

Now on master when you click on a wrapped map the graticule does not switch over to that map. This is a new problem and would explain why this pull request would not appear to have any affect on the current master. Is it intended that the graticule start behaving this way in 3.7?

ahocevar:
I'm going to merge this now. It improves things slightly, but a real fix would be to use a vector layer for the graticule.

ahocevar:
@Jebjum Thanks for the fix. Can you please sign a [CLA](https://github.com/openlayers/cla). Please add a note here when done.

Jebjum:
I have signed the CLA. Thanks.

PR #3860, opened 2015-07-01
Update example layout

### LINKED ISSUE #3676: Broken example layout
When the examples are resized, the header jumps way down the page.

image

We could take this opportunity to update Bootstrap (and/or simplify our use of it).

tschaub:
This uses the latest Bootstrap for examples.  This makes the header for the examples look more like the rest of [the website](http://openlayers.org/) and fixes the layout for smaller viewports.

image

Fixes #3676.


ahocevar:
Very nice, please merge.

elemoine:
Very nice. Thanks a lot for this change.

PR #3863, opened 2015-07-01
Handle CDATA in attribute parsing for GML format

nhambletCCRI:
Following [3827](https://github.com/openlayers/ol3/issues/3827),
handle CDATA XML nodes in attribute parsing of GML data. Currently
such data will be expected to be a geometry, and will fail to parse.

Treating the CDATA node as text is the easiest way to handle such
an attribute.


bartvde:
thanks, this looks good, have you signed a CLA already https://github.com/openlayers/cla ?

nhambletCCRI:
Yep, we sent one in at the end of last week.

bartvde:
thanks for your contribution!

PR #3864, opened 2015-07-02
Check projection.canWrapX() before wrapping tiles

petrsloup:
EDIT: see below for better solution

It probably does not make sense to have `wrapX:true` as default on `ol.source.XYZ` if the projection is not global (therefore `projection.canWrapX()` returns false).

This can lead to some strange behavior when the projection has `extent` defined (example: http://jsfiddle.net/30wjhq38/). It can be solved by manually specifying `wrapX: false` as the constructor option, but it should not be required if the projection can not wrap X.

### DIFF
--- a/src/ol/source/tilesource.js
+++ b/src/ol/source/tilesource.js
@@ -231,7 +231,7 @@ ol.source.Tile.prototype.getTileCoordForTileUrlFunction =
       opt_projection : this.getProjection();
   var tileGrid = this.getTileGridForProjection(projection);
   goog.asserts.assert(!goog.isNull(tileGrid), 'tile grid needed');
-  if (this.getWrapX()) {
+  if (this.getWrapX() && projection.canWrapX()) {
     tileCoord = ol.tilecoord.wrapX(tileCoord, tileGrid, projection);
   }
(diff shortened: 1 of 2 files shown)


ahocevar:
This change should not make a difference, because x-axis wrapping will only be applied when the projection's canWrapX() returns true anyway. Your fiddle shows no map for me, regardless of the wrapX setting.

ahocevar:
I just saw that we don't check the projection's canWrapX() for tile sources indeed.. So a better fix would be to change https://github.com/openlayers/ol3/blob/master/src/ol/source/tilesource.js#L234 to something like
```js
  if (this.getWrapX() && projection.canWrapX()) {
```

ahocevar:
Same thing is missing in ol.source.TileVector's getTileCoordForTileUrlFunction() method.

petrsloup:
@ahocevar I just fixed the fiddle, sorry about that!

Yes, that should also fix the problem -- although then there will be possible `wrapX == true` without any effect. However, it's indeed better solution than trying to cover all possible inputs in all sources.

I will change this PR to modify the `getTileCoordForTileUrlFunction` methods instead.

petrsloup:
Pushed the alternative (better) solution instead of the original one.

ahocevar:
Thanks for working on this @petrsloup.

PR #3865, opened 2015-07-02
ol.View#fit()

bartvde:
Implemented review comments on top of @malaretv's work so we can get this into the 3.7 release.

closes #3482


bartvde:
Only seeing now that not all commit messages start with a capital ....

PR #3867, opened 2015-07-02
Do not require projection extent for x-wrapping tile sources

ahocevar:
#3864 causes a regression, because a tile source now requires a projection with specified extent when `global` is set to true. This change fixes the regression.


ahocevar:
This should be considered for the 3.7 release. Should be a very easy review.

bartvde:
LGTM, a test case would be good to have but I won't block you on it (or you can do it after the release, btw I plan for late in day today).

ahocevar:
Tests added.

bartvde:
thanks @ahocevar please merge if Travis gives the okay

bartvde:
@ahocevar I'll wait for this one before cutting the release

Issue #3868, opened 2015-07-03
`@fires` for ol.Feature not correct


probins:
http://openlayers.org/en/master/apidoc/ol.Feature.html is not showing the change event correctly. Looks like the `@fires` should give the event type, which isn't the case with ol.Feature. I assume the intention here is to list what triggers a change. This needs changing but I'm not sure what to :-)

probins:
I think what needs to change here is:
* `@fires` line should be removed from https://github.com/openlayers/ol3/blob/master/src/ol/feature.js#L53
* if it's the case that 'id' and 'style' changes should trigger events, then `@observable` should be added to `getId`, `setId`, `getStyle`, `setStyle` (`getGeometry` and `setGeometry` are already observable)

If someone could confirm this is how this should be, I'll create a PR.

ahocevar:
Yes, that's correct. See https://github.com/openlayers/ol3/tree/master/config/jsdoc/api#events for the documentation on how to annotate events. Please go ahead with the changes as you suggested above.

PR #3871, opened 2015-07-04
Document change events properly

ahocevar:
With this change, the generic 'change' event is properly documented, as
all other events. It is no longer necessary to annotate `@fires change` for
every ol.Observable subclass.


ahocevar:
This is also a follow-up on #3868.

tschaub:
Nice improvement!

I notice that there is a link to `ol.Observable` next to the `change` event.

image

Not sure if this was intentional, but it is inconsistent with the other links to the event objects.  Also not a big deal if it is awkward to change.

Either way, please merge.

ahocevar:
Updated pull request to not show a link to ol.Observable.

PR #3876, opened 2015-07-06
Remove extent to the list of loaded extent in case of server error/restriction

geonux:
This method offers the possibility of removing the current extent of the list of already loaded extent . In this way, the current extent ( or smaller extents ) can be (re)loaded on server error.

The creation of this code has been motivated by the impossibility to load "children" tiles when a lower zoom tile was marked as treated , even if the server returned an error (especially due to the async. of http method in JS). In my case, the data are too dense to be displayed at certain zoom levels, but may be drawn on a higher zoom level. I could probably use the notion of minZoom , but in my case, the threshold zoom level can not be determined beforehand.

I'm not satisfied with the method name. Maybe can you have a better proposal ?

Usage (example with http helper from AngularJS) :

```
            var vectorSource = new ol.source.Vector({
              loader: function(extent, resolution, projection) {
                var _url = options.featuresUrl.replace("{id}", annotation.id) +
                  "?format=json&bbox=" + extent.join(',');

                $http.get(_url).success(function(data, status, headers, config) {
                  var features = jsonFormat.readFeatures(data);
                  vectorSource.addFeatures(features);
                }).error(function(data, status, headers, config) {
                  if (status === 500 && data.cause == "Too many results")
                    vectorSource.removeFromLoadedExtent(extent);
                });
              },
              strategy: ol.loadingstrategy.tile(ol.tilegrid.createXYZ({})),
              projection: 'EPSG:3857'
            });

```

Can also be a solution for :
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ol3-dev/vu9eyrGLIrY/reV2RXkR5Q0J

### DIFF
--- a/src/ol/source/vectorsource.js
+++ b/src/ol/source/vectorsource.js
@@ -793,6 +793,27 @@ ol.source.Vector.prototype.loadFeatures = function(
 
 
 /**
+ * Remove the current extent from the list of loaded extent.
+ * @param {ol.Extent} extent Extent.
+ * @api
+ */
+ol.source.Vector.prototype.removeFromLoadedExtent = function(
+    extent) {
+  var loadedExtentsRtree = this.loadedExtentsRtree_;
+  var obj;
+  loadedExtentsRtree.forEachInExtent(extent, function(object) {
+    if (ol.extent.equals(object.extent, extent)) {
+      obj = object;
+      return true;
+    }
+  });
+  if (obj) {
+    loadedExtentsRtree.remove(obj);
+  }
+};
+
+
+/**
  * Remove a single feature from the source.  If you want to remove all features
  * at once, use the {@link ol.source.Vector#clear source.clear()} method
  * instead.


alvinlindstam:
Unlisting a failed request seems reasonable.

As noted in https://github.com/openlayers/ol3/issues/3812, clearing the extent from the loaded extents will cause infinite reloads, even without panning or interaction with the map. Polling the server continuously might not always be desired (if the problem is persistent). Should it be up to the loader function to keep track of that and ignore repeated calls as it sees fit, or should that somehow be included in ol3?

How does ol3 handle failed tile requests, are they ever retried?

geonux:
Because of the asynchronous of ajax methods in javascript, the "custom" loader function can not return the status (success/failed) of a vector tile. This information can only be provided by a callback. Also, there is no check of the loader return value in OL3 loader code - not relevant because of async. 
Cf : ```ol.source.Vector.prototype.loadFeatures```:
```
    if (!alreadyLoaded) {
      this.loader_.call(this, extentToLoad, resolution, projection);
      loadedExtentsRtree.insert(extentToLoad, {extent: extentToLoad.slice()});
    }
```
In any case, the extent is added to the ```loadedExtentsRtree``` index.

The function I created is a kind of callback specialized in the management of failled extent.
With this solution, we can avoid infinite reloading because we monitor what is valid or not. In my case, I check in my loader when an extend must be marked as invalid or as to be retried (see example in the first post).
I dont have any problem of infinite reloading without panning or interraction (maybe because I use tile loading strategy - to be tested).

elemoine:
What if the loader was passed a callback that it would have to call to report completion. When success is reported the callback would mark the extent as loaded. Alternatively the source could expose methods to mark/unmark extents as loaded, and the loader would be responsible for calling these methods as appropriate.

geonux:
I think this is the same logic of this patch and that can be an alternative to my proposal.However, passing a new callback to the loader as you propose change the loader interface, which can cause significant migration cost.

ahocevar:
Closing due to lack of activity. @geonux, please reopen when you want to address the remaining concerns.

thewildman:
Could someone (@ahocevar ?)  explain what are the remaining concerns ? 
We are in the same situation as @geonux (i.e. handling of failed wfs requests), and simply exposing removeFromLoadedExtent to the user seems like a could solution to us.
Couldn't this patch be merged as-is ?

ahocevar:
I have no concerns except missing docs to actually make this useful. Can someone provide a custom loader code that uses this function?

ahocevar:
Sorry, just saw the usage example above. Will document and merge for v4.6.0.

thewildman:
Thanks a lot @ahocevar , it'll really help us !

ahocevar:
Replaced by #7538.

PR #3881, opened 2015-07-07
Add an example on how to create XYZ or ArcGISRest sources

bartvde:
from the capabilities JSON response

### DIFF
--- a/examples/arcgis-rest-capabilities.js
+++ b/examples/arcgis-rest-capabilities.js
@@ -12,112 +12,6 @@ goog.require('ol.tilegrid.TileGrid');
 
 var map;
 
-var getResolutionsFromConfig = function(config) {
-  var tileInfo = config['tileInfo'];
-  if (tileInfo) {
-    var resolutions = [];
-    for (var i = 0, ii = tileInfo.lods.length; i < ii; ++i) {
-      resolutions.push(tileInfo.lods[i].resolution);
-    }
-    return resolutions;
-  }
-};
-
-var getProjectionFromConfig = function(config) {
-  var epsg = 'EPSG:' + config.spatialReference.wkid;
-  var units = config.units === 'esriMeters' ? 'm' : 'degrees';
-  var projection = ol.proj.get(epsg) ? ol.proj.get(epsg) :
-      new ol.proj.Projection({code: epsg, units: units});
-  return projection;
-};
-
-var createXYZSource = function(config, serviceUrl, attributions, projection,
-    resolutions) {
-  var tileInfo = config['tileInfo'];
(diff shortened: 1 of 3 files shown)


bartvde:
Tested with the following services:

http://tiles.arcgis.com/tiles/nSZVuSZjHpEZZbRo/arcgis/rest/services/TMK1850/MapServer
http://services.arcgisonline.com/arcgis/rest/services/ESRI_Imagery_World_2D/MapServer
http://tiles.arcgis.com/tiles/nGt4QxSblgDfeJn9/arcgis/rest/services/Milan_Population_Density_6class/MapServer

examples/arcgis-rest-capabilities.js
```diff
@@ -0,0 +1,171 @@
...
+  var sourceInfo;
+  if (config.tileInfo) {
+    sourceInfo = createXYZSource(config, serviceUrl, attributions, projection,
+        resolutions);
+  } else {
+    sourceInfo = createArcGISRestSource(attributions, serviceUrl);
+  }
+  return new ol.layer.Tile({
+    source: sourceInfo.source,
+    extent: sourceInfo.extent
```
ahocevar: I don't think it's good to restrict the extent on the layer level. Should be done on the tilegrid level instead IMHO.

ahocevar:
Why not create an `ol.source.TileArcGISRest.optionsFromCapabilities` function with a proper tile grid configuration? Then the application code could be much simpler.

bartvde:
@ahocevar the code will create an XYZ or ArcGISRest source, or you don't think that is an issue here? Actually doing the config for the XYZ is 95% of this code.

ahocevar:
Ah, I see now. Still I think that's a lot of application code. Not sure where to fit it in the library though.

bartvde:
I agree, but I also couldn't find a good spot in the library.

bartvde:
Unless we create a new source (for example ol.source.ArcGISXYZ) that has a optionsFromCapabilities function.

ahocevar:
Ok, to bring this into the library we'll need an externs file and probably a format. So maybe just keep it like it is, but change the code so the extent is restricted through the tile grid, not the layer.

bartvde:
Okay it might make sense to create a separate library for ol3-esri integration, which will also contain the style conversion that I need to work on soon. We can then showcase these in an example will less code. So I'll hold off for this for now.

bartvde:
Started on the external lib idea here: https://github.com/bartvde/ol3-esri
Updated the branch to show usage.
WIP.

bartvde:
I'm closing this since the example will be in ```ol3-esri``` instead

PR #3883, opened 2015-07-07
Ignore layer filter for unmanaged layers

### LINKED ISSUE #3878: forEachFeatureAtPixel returns wrong layer if feature already selected
Following situation used to work, but no longer does since removal of FeatureOverlay:

* 2 select interactions, one on mousemove, one on click, so user gets title on hover, and can click for more details. So when clicks, the feature is already selected. Click should return the selected feature, but event.selected is empty.

This seems to be because map.forEachFeatureAtPixel is returning the wrong layer: the overlay layer not the original vector layer, so it's failing the layer filter. I think the problem is around https://github.com/openlayers/ol3/blob/master/src/ol/renderer/maprenderer.js#L173 which I think should also be applying the layer filter so it only returns features on the original layers, not the overlay layer.

ahocevar:
To make unmanaged vector layers work like the removed ol.FeatureOverlay, the layer filter for ol.Map#forEachFeatureAtPixel needs to ignore unmanaged layers.

Fixes #3878.

### DIFF
--- a/src/ol/renderer/maprenderer.js
+++ b/src/ol/renderer/maprenderer.js
@@ -167,8 +167,9 @@ ol.renderer.Map.prototype.forEachFeatureAtCoordinate =
   for (i = numLayers - 1; i >= 0; --i) {
     var layerState = layerStates[i];
     var layer = layerState.layer;
-    if (ol.layer.Layer.visibleAtResolution(layerState, viewResolution) &&
-        layerFilter.call(thisArg2, layer)) {
+    if (!layerState.managed ||
+        (ol.layer.Layer.visibleAtResolution(layerState, viewResolution) &&
+        layerFilter.call(thisArg2, layer))) {
       var layerRenderer = this.getLayerRenderer(layer);
       result = layerRenderer.forEachFeatureAtCoordinate(
           layer.getSource().getWrapX() ? translatedCoordinate : coordinate,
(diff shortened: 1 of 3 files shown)


bartvde:
LGTM

elemoine:
I wonder if this fixes https://github.com/openlayers/ol3/issues/3885 as well.

src/ol/renderer/maprenderer.js
```diff
@@ -167,8 +167,9 @@ ol.renderer.Map.prototype.forEachFeatureAtCoordinate =
   for (i = numLayers - 1; i >= 0; --i) {
     var layerState = layerStates[i];
     var layer = layerState.layer;
-    if (ol.layer.Layer.visibleAtResolution(layerState, viewResolution) &&
-        layerFilter.call(thisArg2, layer)) {
+    if (!layerState.managed ||
```
bartvde: shouldn't this be ```if (layerState.managed && ``` ?

bartvde: No I guess I took @elemoine's words in #4143 too literally here: ```PR #3883 made forEachFeatureAtPixel ignore unmanaged layers``` -> wrt layer filtering was meant I guess. I guess I'll need to filter it out myself, but if I'm reading things correctly, I can only do this after @elemoine's changes get in?

PR #3884, opened 2015-07-08
Provide a debug loader for the library

tschaub:
This adds a simple task for serving just the library in debug mode.

The server can be started with:

    npm run debug-server

Depends on openlayers/closure-util#66.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "async": "0.9.0",
-    "closure-util": "1.4.0",
+    "closure-util": "1.5.0",
     "fs-extra": "0.12.0",
     "glob": "5.0.3",
     "graceful-fs": "3.0.2",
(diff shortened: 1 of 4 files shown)


elemoine:
What's the difference between using `npm start` and using `npm run debug-server`?

Please merge when `closure-util` 1.5.0 is released.

tschaub:
> What's the difference between using npm start and using npm run debug-server?

The `tasks/serve.js` script configures a manager with the test scripts included in the lib.  So if no "main" identifier is found, the tests are loaded and run.  This is not what you want when you just want to load the library (and you aren't using `goog.require` for dependencies).

I'd like to change how the tests are run (ideally getting rid of `build/test_requires.js` and `build/test_rendering_requires.js`).  It may make sense to have a separate script for running the test server, and `tasks/serve.js` could be used as a debug loader for just the library.  Or maybe `tasks/serve.js` could be made to work as a debug loader for the library and also serve the tests.

But this has become an acute pain point (not being able to easily debug), and I wanted to have a solution.  The `tasks/serve-lib.js` script is just about serving the library (not building examples or running tests).

elemoine:
Cool. I think it's good to have your explanations written here. 

Please merge.

PR #3894, opened 2015-07-10
Install Python dependencies without sudo

### LINKED ISSUE #3893: Remove use of sudo on Travis
Our builds are running on Travis' "legacy" system because of our use of `sudo` to install Python packages.

It would be nice to be able to use the new container based infrastructure instead.  We should be able [to do this](http://docs.travis-ci.com/user/migrating-from-legacy/#How-can-I-use-container-based-infrastructure%3F) if we get rid of the `sudo pip install -r requirements.txt`.

tschaub:
This should allow us to use the [container based infrastructure](http://docs.travis-ci.com/user/migrating-from-legacy/) on Travis.

Fixes #3893.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
+language: python
+sudo: false
 env:
   - DISPLAY=:99.0
 
 before_install:
-  - "sudo pip install -r requirements.txt"
+  - "pip install -r requirements.txt"
   - "npm install -g npm && npm install"
 
 before_script:


elemoine:
Cool !

tschaub:
Hmm.  Thought we'd get a virtualenv's Python [by default](http://docs.travis-ci.com/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs), but it looks like this is still [trying to install things system-wide](https://travis-ci.org/openlayers/ol3/builds/70400939).

@elemoine any ideas?

tschaub:
Seeing if things work if we force a [Python VM image](http://docs.travis-ci.com/user/ci-environment/#Python-VM-images).

tschaub:
OK, it looks like we're successfully running on Travis CI's container infrastructure (compare [an old build](https://travis-ci.org/openlayers/ol3/builds/70305007) with [one from this branch](https://travis-ci.org/openlayers/ol3)).

I was hoping this would mean [faster build time](http://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F), but it doesn't look like a significant difference.  It should at least mean a faster startup time (but I haven't compared with an old build).

elemoine:
Yep, `language: python` need be used so `pip` installs in a virtual env.

elemoine:
LGTM. Thanks!

PR #3895, opened 2015-07-10
Rework the tile queue for multiple queues

aisaacs:
This fixes https://github.com/openlayers/ol3/issues/3889, which has a good explanation of the issue.

Re: Greediness. Before [this change](https://github.com/openlayers/ol3/commit/2142b538ac7b7aafdbae37d58613454f5f5baab5), the implementation of the queue would calculate the amount of tiles to load before dequeuing / loading. In the multiple queue case, after https://github.com/planetlabs/ol3/commit/b57cdb730c5b57450a2f9adcca44038994181165 this could cause a queue to not be fully utilized. For instance, if a queue were to be requested to load N new tiles, and had the capacity for it, if any of the dequeued tiles were in a state other than IDLE the total number of tiles that would start loading would be less than what had been requested. 
This change makes the queue always try to load tiles at its maximum capacity.

### DIFF
--- a/src/ol/tilequeue.js
+++ b/src/ol/tilequeue.js
@@ -94,9 +94,11 @@ ol.TileQueue.prototype.loadMoreTiles = function(maxTotalLoading, maxNewLoads) {
   var i, tile;
   for (i = 0; i < newLoads; ++i) {
     tile = /** @type {ol.Tile} */ (this.dequeue()[0]);
-    goog.events.listen(tile, goog.events.EventType.CHANGE,
-        this.handleTileChange, false, this);
-    tile.load();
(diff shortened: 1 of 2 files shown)


tschaub:
Very nice.

It should probably also be noted that this doesn't change anything for the single queue case (unless something else was calling `tile.load()` on tiles in the queue).  For all cases, this means that the `maxTotalLoading` and `maxNewLoads` arguments for `loadMoreTiles` are respected (per queue).

Thanks for tracking this down and for the nice fix @aisaacs.

ahocevar:
Thanks for this. Great work!

PR #3896, opened 2015-07-11
Fire WebGL precompose event in same sequence as other renderers

### LINKED ISSUE #3885: select feature v 3.7.0 problem, render webgl.
When I select a feature, using the interaction select, and using the render webgl ( I use webgl because I have a big amount of data), the style of the selection is not functionally, the feature is selected but it's transparent, with the default style I have the same problem.

This happens only in the version 3.7.0, with the 3.6.0. I have no problem.

ahocevar:
In other map renderers, the precompose event is fired before preparing and composing layer frames. In WebGL, it is fired in between. This change makes it so the sequence of events is the same for all renderers.

Because the WebGL renderer creates the list of layers to render before the precompose event, unmanaged layers are never rendered. This is also fixed by dispatching the precompose event earlier.

Fixes #3885.


tschaub:

PR #3900, opened 2015-07-13
Do not add Style as a feature property

### LINKED ISSUE #3832: ol.format.GeoJSON.writeFeatureObject adds a Style property which prevent JSON export
`ol.format.GeoJSON.writeFeatureObject` sometimes adds a Style property when trying to get it as a GEOJson object which prevents the export of the features to JSON because of cyclic references. The problem occurs when I import a KML file in the map. If I try to use `writeFeature`, I get a stackoverflow error instead.

Here is the code I use.
```javascript
var source = layer.getSource();
var features = source.getFeatures();
var geojsonFormat = new ol.format.GeoJSON();

for (var i = 0, ii = features.length; i < ii; ++i) {
    var feature = features[i];
    var geojsonFeature = geojsonFormat.writeFeatureObject(feature);

    // Remove style property to avoid problem when converting to JSON
    if (geojsonFeature.hasOwnProperty('properties') &&
            geojsonFeature.properties &&
            geojsonFeature.properties.hasOwnProperty('Style')) {
      delete geojsonFeature.properties.Style;
    }
}
```

I think it shouldn't behave this way and always allow the export to JSON. This is the behavior I expect after reading the documentation.

Tested with OpenLayers 3.6.0

Problem originally submitted to https://github.com/camptocamp/ngeo/pull/239

Jenselme:
This makes the KML format not add a Style property on features, as this may cause problems when serializing to another format such as JSON.

Close #3832

### DIFF
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -289,28 +289,6 @@ ol.format.KML.ICON_ANCHOR_UNITS_MAP_ = {
  */
 ol.format.KML.createFeatureStyleFunction_ = function(
     style, styleUrl, defaultStyle, sharedStyles) {
-  var findStyle = (
-      /**
-       * @param {Array.<ol.style.Style>|string|undefined} styleValue Style
-       *     value.
-       * @return {Array.<ol.style.Style>} Style.
-       */
-      function(styleValue) {
-        if (goog.isArray(styleValue)) {
-          return styleValue;
-        } else if (goog.isString(styleValue)) {
-          // KML files in the wild occasionally forget the leading `#` on
-          // styleUrls defined in the same document.  Add a leading `#` if it
-          // enables to find a style.
-          if (!(styleValue in sharedStyles) &&
-              ('#' + styleValue in sharedStyles)) {
-            styleValue = '#' + styleValue;
-          }
-          return findStyle(sharedStyles[styleValue]);
-        } else {
-          return defaultStyle;
-        }
-      });
   return (
       /**
        * @param {number} resolution Resolution.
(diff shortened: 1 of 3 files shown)


src/ol/format/kmlformat.js
```diff
@@ -323,6 +279,57 @@ ol.format.KML.ICON_ANCHOR_UNITS_MAP_ = {
...
+ * @param {string} styleUrl Style URL.
+ * @param {Array.<ol.style.Style>} defaultStyle Default style.
+ * @param {Object.<string, (Array.<ol.style.Style>|string)>} sharedStyles
+ * Shared styles.
+ * @return {ol.FeatureStyleFunction} Feature style function.
+ * @private
+ */
+ol.format.KML.createFeatureStyleFunction_ = function(
+    style, styleUrl, defaultStyle, sharedStyles) {
+  var findStyle = (
```
ahocevar: It should not be necessary to create a new `findStyle()` function for every call to `createFeatureStyleFunction_()`. Instead, you could create a static function, e.g. `ol.format.KML.findStyle(styleUrl, defaultStyle, sharedStyles)`.

src/ol/format/kmlformat.js
```diff
@@ -1668,13 +1675,27 @@ ol.format.KML.prototype.readPlacemark_ = function(node, objectStack) {
...
-    feature.setStyle(this.featureStyleFunction_);
+    var style = goog.object.get(object, 'Style');
+    var styleUrl = goog.object.get(object, 'styleUrl');
+    var styleFunction = ol.format.KML.createFeatureStyleFunction_(
+        style, styleUrl, this.defaultStyle_, this.sharedStyles_);
+    feature.setStyle(styleFunction);
   }
+  goog.object.remove(object, 'Style');
+  // we do not remove the styleUrl property from the object, so it
+  // gets stored on feature when setProperties is called
```
ahocevar: Since `styleUrl` is stored in the style function closure, there should not be any need to keep it here. Or am I missing something?

ahocevar:
Also see https://github.com/openlayers/ol3/issues/3832#issuecomment-121174757 for a an idea of a simpler solution.

src/ol/format/kmlformat.js
elemoine: I think you are right. I am going to provide a commit that addresses that.

src/ol/format/kmlformat.js
elemoine: We used to store both `Style` and `styleUrl` on the feature properties. With this PR we no longer store the `Style` on the feature properties, because `ol.style.Style` objects are not JSON-serializable. But we still store `styleUrl`, probably just because we do not have any reason not to store it anymore.

elemoine:

@Jenselme, please add https://github.com/elemoine/ol3/commit/f10fb8172ce8a16b4de5ba4a6127b41adfcb0200 to your branch. When done, I'll merge this PR. Thanks.

Jenselme:
PR updated.

cs09g:
`delete object['Style']` causes some kind of disadvantage. When you create a custom function to create feature which needs its position and icon url to display, there's no way to get icon url from KML data. because it only has style functions. style function is suitable when style has some condition but in this case there's no condition.

src/ol/format/kmlformat.js
cs09g: this causes some kind of disadvantage. When you have a function that creates feature with parameter (its position, icon url), there's no way to get icon url from KML data. It's better to use style as function, when style should be applied by some condition.

PR #3903, opened 2015-07-14
Ignore properties that cannot be serialized

### LINKED ISSUE #3832: ol.format.GeoJSON.writeFeatureObject adds a Style property which prevent JSON export

ahocevar:
Fixes #3832.

### DIFF
--- a/src/ol/format/geojsonformat.js
+++ b/src/ol/format/geojsonformat.js
@@ -544,8 +544,7 @@ ol.format.GeoJSON.prototype.writeFeatureObject = function(
   } else {
     object['geometry'] = null;
   }
-  var properties = feature.getProperties();
-  goog.object.remove(properties, feature.getGeometryName());
+  var properties = feature.getSerializableProperties();
   if (!goog.object.isEmpty(properties)) {
     object['properties'] = properties;
   } else {
--- a/src/ol/object.js
+++ b/src/ol/object.js
@@ -182,6 +182,23 @@ ol.Object.prototype.getProperties = function() {
 
 
 /**
+ * Get an object of all serializable property names and values.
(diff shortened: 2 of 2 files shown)


probins:
but a GeoJSON property can be an object, which would be ignored by this change
```
"properties": {
  "prop0": "value0",
  "prop1": {"this": "that"}
}
```
http://geojson.org/geojson-spec.html#examples

ahocevar:
Right @probins, we should be able to cover this. I'm working on it.

ahocevar:
Giving up on this. There is no efficient way to detect and avoid objects that are too big to serialize.

@probins: GeoJSON serialization is not affected by #3900.

PR #3906, opened 2015-07-15
Clear features properly when there is no spatial index

### LINKED ISSUE #3904: Error on ol.Vector.source.clear()
When you create a vector source with `useSpatialIndex: false`, calling `clear()` on it produces this error:

`Uncaught AssertionError: Assertion failed: featureChangeKeys is an empty object now`

Example Fiddle: https://jsfiddle.net/Burstaholic/dzyzw6t4/ (modified from the example in #3812 )

Workaround: calling `clear(true)` succeeds, but this may not be desirable in all cases.

ahocevar:
Previously `#clear()` only kept the features collection in sync whern there was also a features RTree.

Fixes #3904.


marcjansen:
Nice. LGTM please merge

PR #3919, opened 2015-07-23
Insert quotes for correct output

simon1389:
The quotes which have been removed in this commit:  https://github.com/openlayers/ol3/commit/0286564e8a4e19a67123fccae82ea7a18e524307

are necessary in order to get a correct output. Without the quotes, the names of the properties get minified.

### DIFF
--- a/src/ol/format/geojsonformat.js
+++ b/src/ol/format/geojsonformat.js
@@ -198,8 +198,8 @@ ol.format.GeoJSON.writeGeometry_ = function(geometry, opt_options) {
  */
 ol.format.GeoJSON.writeEmptyGeometryCollectionGeometry_ = function(geometry) {
   return /** @type {GeoJSONGeometryCollection} */ ({
-    type: 'GeometryCollection',
-    geometries: []
+    'type': 'GeometryCollection',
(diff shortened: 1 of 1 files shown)


fredj:
Do you use a custom build of the library?

It's working with v3.7.0 and master

simon1389:
Yes, i'm using a custom-build.

Does the  writeGeometry-method return non-minified properties with the 3.7.0-minified-version ?

probins:
you need to include externs/geojson.js

simon1389:
including externs/geojson.js also does the trick..
anyway reintroduce the quotes? i think the non-minified name of the properties is ALWAYS expected..

probins:
defining exports/externs outside the source is more flexible; people can include whatever they want/need at build time rather than having it hard-coded

simon1389:
Yes, you are right with that!
But i dont think, there is a usecase in which i want the property-names minified when i call the writeGeometry-function? So it's useless to minify the output of the function

bartvde:
Using externs in the custom build is really the solution for this, not reverting back to using quotes in the library, so closing.

PR #3921, opened 2015-07-25
Use Travis CI cache

bjornharrtell:
Travis CI cache for dependencies is now available for container builds which ol3 recently switched to so perhaps it can be used? npm install seem to typically take about 100s.


tschaub:
Thanks @bjornharrtell.

PR #3922, opened 2015-07-25
Document {?-?} pattern in expandUrl

probins:
This does not seem to be documented anywhere at the moment. If this wording is ok, I'll add it to all the other source options (there's quite a few) that call expandUrl().


tschaub:
Thanks @probins.

PR #3930, opened 2015-07-28
Use goog.functions.identity instead of goog.identityFunction

fredj:
`goog.identityFunction` was deprecated and has been removed from closure-library.


ahocevar:
Thanks, please merge.

PR #3542, opened 2015-04-12
Generate example index and rebuild examples on source changes

tschaub:
With this change, `tasks/build-examples.js` generates the `example/index.js` file in addition to generating the rest of the examples.

When `tasks/serve.js` is run (which is launched by `npm start`), the examples are built before the server is started.  Then, whenever an example source file changes, the examples are rebuilt (and any attached clients reload).

This is one step toward #3532.

I'd like to move the generated `examples` directory out of the root, as I think all generated files should go in `build`.  But this will involve more changes to `build.py`, and I'd rather handle it separately.

Fixes #3529.

PR #3931, opened 2015-07-28
Remove unused htmlparser2 package

fredj:
No longer used since #3542


ahocevar:
Great find, please merge!

PR #3932, opened 2015-07-28
SnapOptions: Fix typo in pixelTolerance JSDoc

Turbo87:
This was apparently copy-pasted from the `ModifyOptions` without adjustment


fredj:
thanks

PR #3933, opened 2015-07-28
Snap center to pixel to avoid floating point issues

ahocevar:
When working with degrees as map units, the viewState's center can cause floating point issues that cause the map to not be rendered. This can be seen [here](http://jsfiddle.net/azuxdpuu/1/), where the map image won't appear before the user zooms or pans the view.

By snapping the viewState's center to pixels, this issue can be avoided.

**Edit**: As the diff shows, this pull request also makes image layers look not blurry any more when using the WebGL renderer.


marcjansen:
The fiddle renders fine if FF btw.

ahocevar:
@marcjansen I updated the fiddle to use a fixed map size. You should be able to see the issue in Firefox as well now.

marcjansen:
Nice, thanks for the update and the fix. Please merge if travis gives green.

PR #3934, opened 2015-07-28
Fix `drawstart` and `drawend` events when drawing a point

### LINKED ISSUE #3928: ol.interaction.Draw point and drawstart events
http://jsfiddle.net/fredj/nxw73wyb/

As soon as the cursor enters the map, a `drawstart` event is fired (on `mousemove`).

The expected behaviors is to fire a `drawstart` on `mousedown` and a `drawend` on mouseup.

fredj:
Now a `drawstart` event is fired on `mouseup` immediately followed by a `drawend`.

The previous behavior was to fire a `drawstart` on the first `mousemove`.
Fixes #3928


tschaub:
Would it be possible to add to the tests so we don't get a regression when refactoring?

fredj:
I've updated the tests

elemoine:
This looks good to me.

Issue #3874, opened 2015-07-06
second argument to ol.source.WMTS.optionsFromCapabilities should be in olx.js


bartvde:
The second argument (config) to ol.source.WMTS.optionsFromCapabilities should be defined in olx.js

ahocevar:
Could also be improved to use and respect `olx.source.WMTSOptions`.

PR #3942, opened 2015-07-30
Fix WMTS TileMatrixSet lookup by SRS identifier

### LINKED ISSUE #3940: Adding projection in ol.source.WMTS.optionsFromCapabilities generates exception
If I add a projection attribute to WMTS options parsed form WMTS Capabilities (see http://jsfiddle.net/7hesnth5/7/) , I get an exception in line 115566 of ol-debug.js (version 3.7.0):

TypeError: elt.TileMatrixSet.SupportedCRS is undefined ol-debug.js:115566

Should it not be TileMatrixSet[0].SupportedCRS?

ahocevar:
TileMatrixSet lookup by SRS identifier is completely broken. In addition to this fix, the whole `ol.source.WMTS#optionsFromCapabilities()` function could use an overhaul (see #3874).

cc @bartvde 
 
Fixes #3940.

### DIFF
--- a/src/ol/source/wmtssource.js
+++ b/src/ol/source/wmtssource.js
@@ -351,19 +351,25 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
 
   goog.asserts.assert(l['TileMatrixSetLink'].length > 0,
       'layer has TileMatrixSetLink');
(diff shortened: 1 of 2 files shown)


marcjansen:
Looks good to me, before merging I'd favor a comment of @bartvde, though.

bartvde:
LGTM

PR #3945, opened 2015-07-31
Simplify icon example and show popup at clicked position

### LINKED ISSUE #3944: Popups always appear in originating worldview when data wraps dateline
http://openlayers.org/en/v3.7.0/examples/icon.html

Zoom out and click on an icon that has wrapped the dateline (not the original). The popup always centers over the original feature. Is that an issue with the library or is there a better way to write that code?

I'm sure the introduction of canvas and WebGL rendering has upped the complexity for determining feature interaction and clicks events, but unfortunately that complexity is very visible in the code (see icon example above) to implement a common task of creating a hover and click event for a feature in version 3 over version 2. Is there any plan to roll that into a higher level function to make popup creation easier moving forward?

ahocevar:
By using the event's coordinate instead of the feature's, we make sure that the popup is shown where the user clicked when on a wrapped world.

Fixes #3944.

### DIFF
--- a/examples/icon.css
+++ b/examples/icon.css
@@ -1,6 +1,3 @@
 #map {
   position: relative;
 }
-#popup {
-  padding-bottom: 45px;
-}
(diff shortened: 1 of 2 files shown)


marcjansen:
LGTM, please merge.

PR #3946, opened 2015-08-01
Fix vertex deletion for Modify interaction on mobile devices

### LINKED ISSUE #3935: ol.interaction.Modify.deleteCondition fails on Android
`ol.interaction.Modify` can be used to delete vertices of e.g. LineStrings, but this doesn't seem to work on Android and possibly other touchscreen devices.

From debugging it looks like the following condition is failing and due to that not removing the vertex:

````
    if (!(this.lastNewVertexPixel_[0] === this.lastPixel_[0] &&
        this.lastNewVertexPixel_[1] === this.lastPixel_[1])) {
````

I assume that is caused by `handlePointerMove_()` only being called when the feature is modified instead of "onMouseOver".

### PR DESCRIPTION
**Pull request by Turbo87**
This PR adds a few new tests for the `ol.interaction.Modify` class. Deleting vertices on mobile devices was previously not possible due to the missing `pointermove` events on that target (see #3935). A test for that case is added and the code itself is modified to support that use case.

This PR also adds a `ol.events.condition.doubleClick` predicate and is using it to test the `deleteCondition` option of the `Modify` class.

resolves #3935

### DIFF
--- a/test/spec/ol/interaction/modifyinteraction.test.js
+++ b/test/spec/ol/interaction/modifyinteraction.test.js
@@ -106,7 +106,6 @@ describe('ol.interaction.Modify', function() {
       expect(feature.getGeometry().getRevision()).to.equal(1);
       expect(feature.getGeometry().getCoordinates()[0]).to.have.length(5);
 
-      simulateEvent('pointermove', 10, -20, false, 0);
       simulateEvent('pointerdown', 10, -20, false, 0);
       simulateEvent('pointerup', 10, -20, false, 0);
       simulateEvent('click', 10, -20, false, 0);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tschaub** [comment]
Thanks @Turbo87.  Could you summarize the issue in the pull request description and title?  This makes it easier for those generating a changelog (easier than reading all commit messages).

**Comment by Turbo87** [comment]
@tschaub I hope the PR is now more descriptive. Please let me know if you need anything else.

File: test/spec/ol/interaction/modifyinteraction.test.js [hunk 1]
```diff
@@ -106,7 +106,6 @@ describe('ol.interaction.Modify', function() {
       expect(feature.getGeometry().getRevision()).to.equal(1);
       expect(feature.getGeometry().getCoordinates()[0]).to.have.length(5);
 
-      simulateEvent('pointermove', 10, -20, false, 0);
```
ahocevar: This makes me suspicious. These tests should still pass, without the need to modify the event sequence.

ahocevar: If you want to make sure that the tests also pass without the pointermove event, it would be better to add an additional test.

Turbo87: they do pass without the modification, I just wanted them to match reality. I can remove the commit if you don't want that.

tschaub:
Thanks @Turbo87.  I'm merging this to get it in to the 3.8 release.

PR #3947, opened 2015-08-02
Clarify documentation of Image source ratio option.

alvinlindstam:
Clarifies documentation for the *ratio* option of ImageWMS, ImageMapGuide, ImageVector and ImageCanvas. Should make it clear that the ratio should be at least *1*, and that the ratio is in relation to width/height (not area).

Fixes https://github.com/openlayers/ol3/issues/3926

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -4184,7 +4184,8 @@ olx.source.ImageMapGuideOptions.prototype.projection;
 
 /**
  * Ratio. `1` means image requests are the size of the map viewport, `2` means
- * twice the size of the map viewport, and so on. Default is `1`.
+ * twice the width and height of the map viewport, and so on. Must be `1` or 
+ * higher. Default is `1`.
  * @type {number|undefined}
  * @api stable
  */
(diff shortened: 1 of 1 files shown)


alvinlindstam:
All the Image sources with ratios use 1.5 as default, except MapGuideImage which uses 1. Is this intended?

tschaub:
Thanks for the doc improvement @alvinlindstam.  I think `1.5` makes a sensible default.

The MapGuide source [started out](https://github.com/openlayers/ol3/commit/60825f0f997f18242fd19f61add21ae51ca66eb4#diff-824b06584e6f27b4b8ed4ab2f7feb02bR67) with an unconfigurable ratio of `1`.  This was [later extracted](https://github.com/openlayers/ol3/commit/43e4292ec97415d0bcef62522f3abc3b951fa7d2) as an option with the same default value.  If there is a good reason to change it, it would make a pretty simple pull request.

alvinlindstam:
Ok. 1.5 seems like a reasonable default. I was thinking of changing to 1.5 for MapGuide as well, for consistency but backwards comparability is more important.

PR #3949, opened 2015-08-03
Remove count argument from `called` function

fredj:
This function takes no arguments, use the `callCount` property instead.


elemoine:
LGTM.

PR #3950, opened 2015-08-03
Remove reference to vbarray.js

### LINKED ISSUE #3702: Correct config/example.json
Since externs/vbarray.js has been deleted on 3.5.0, we should remove it from the config/example.json 

See https://github.com/openlayers/ol3/blob/master/config/example.json#L24

elemoine:
This file was removed with https://github.com/openlayers/ol3/pull/3516.

Fixes #3702.


bartvde:
LGTM

PR #3954, opened 2015-08-03
Remove ol.control.Control.bindMouseOutFocusOutBlur function

fredj:
It was a workaround added in #1761 to hide the button tooltip on
OSX / Chrome 32 and 33.

Because we're not using the tooltips anymore (see #2781) and this
version of chrome is deprecated this workaround can be removed.


elemoine:
+1

tschaub:
Merging this to get it in 3.8.

fredj:
thanks for the review

PR #3957, opened 2015-08-04
Properly handle vertex deletion with multiple features

### LINKED ISSUE #3956: Uncaught TypeError when deleting a shared vertex
If you select multiple features and delete a shared vertex, you'll see an error like:
```
Uncaught TypeError: Cannot read property 'closure_uid_45707381' of null
```

In the modify interaction, [we're calling `source.removeFeature(null)`](https://github.com/openlayers/ol3/blob/d3c8880b22c61b617989f4e3b5d9eaba1364385c/src/ol/interaction/modifyinteraction.js#L891).

I haven't done a `git blame`, but this works in the 3.6.0 release and not in 3.7.0.  The [modify features example](http://openlayers.org/en/v3.7.0/examples/modify-features.html) demonstrates the problem.

tschaub:
When deleting a vertex shared by multiple features, we iterate through drag segments and only need to remove the vertex feature once.  Currently, we throw on the second drag segment because we try to call `source.removeFeature(null)`.

Fixes #3956.

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -888,8 +888,10 @@ ol.interaction.Modify.prototype.removeVertex_ = function() {
             newSegmentData);
         this.updateSegmentIndices_(geometry, index, segmentData.depth, -1);
 
-        this.overlay_.getSource().removeFeature(this.vertexFeature_);
-        this.vertexFeature_ = null;
+        if (!goog.isNull(this.vertexFeature_)) {
+          this.overlay_.getSource().removeFeature(this.vertexFeature_);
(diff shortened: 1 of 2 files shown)


tschaub:
The problem can be seen in the [modify feature example](http://openlayers.org/en/v3.7.0/examples/modify-features.html) for the 3.7 release.  If you select three features and try to delete a vertex common to all three, the vertex is only removed from two of them (the unhandeled error comes after removing the vertex from the second feature).

ahocevar:
Looks good, please merge.

tschaub:
Thanks @ahocevar.  I was hoping I'd catch you awake!

I'll merge when the test passes.

PR #3960, opened 2015-08-04
Remove unused goog.require

fredj:
Found by the latest version of closure-compiler (v20150729)

### DIFF
--- a/src/ol/attribution.js
+++ b/src/ol/attribution.js
@@ -1,6 +1,5 @@
 goog.provide('ol.Attribution');
 
-goog.require('goog.asserts');
 goog.require('goog.math');
 goog.require('ol.TileRange');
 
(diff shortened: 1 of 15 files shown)


elemoine:
Excellent. Thanks @fredj.

tschaub:
Nice!  What's your process for identifying these @fredj?  Can something in the latest version of the compiler be used to replace `check-requires.py`?

fredj:
> Nice!  What's your process for identifying these @fredj?  Can something in the latest version of the compiler be used to replace `check-requires.py`?

With the new compiler version, wildcard are accepted for `jscomp` properties (that's what I used):
```
...
"jscomp_error": [
      "*"
],
...
```

fredj:
See https://github.com/openlayers/ol3/compare/master...fredj:update_compiler

PR #3961, opened 2015-08-04
Skip features without geometry during canvas replay

gberaudo:
Setting a feature geometry to null or undefined and triggering a hit
detection was throwing an exception.

This patch handles a batched feature without a geometry in the same
fashion as skipped features.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -252,7 +252,8 @@ ol.render.canvas.Replay.prototype.replay_ = function(
       case ol.render.canvas.Instruction.BEGIN_GEOMETRY:
         feature = /** @type {ol.Feature} */ (instruction[1]);
         var featureUid = goog.getUid(feature).toString();
-        if (goog.isDef(skippedFeaturesHash[featureUid])) {
+        if (goog.isDef(skippedFeaturesHash[featureUid]) ||
+            !goog.isDefAndNotNull(feature.getGeometry())) {
           i = /** @type {number} */ (instruction[2]);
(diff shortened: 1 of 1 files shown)


ahocevar:
What happens with the render replay? Are you saying the exception is only thrown upon hit detection? I'm not sure this is the right approach. It would be better to recreate the replay batch upon feature attribute change, because if attributes change, the style may change as well, which would result in different hit detection (and rendering).

gberaudo:
When my application code does `feature.setGeometry(null)`, what happens
is:
- the feature property is changed which triggers an event;
- an asynchronous rendering is requested in
  ol.Map#hadleLayerGroupMemberChanged_()

A few lines later, so before the rendering actually happens,
the hit detection is used (the user clicked on the map):
- the batch contains a feature without geometry
- it crashes on feature.getGeometry().getExtent().

Since we are already testing for skipped features it seems natural to
avoid this issue by also skipping features whose geometry is undefined
or null.

The other "solution" is to have the user force a synchronous rendering of
the map after the class to `feature.setGeometry(null)`. This way, the batch
will be updated and ready for hit detection. 

I prefer the first solution which avoid the race condition.
@ahocevar, is it more clear now?

ahocevar:
Yes, makes sense. Thanks for the fix.

gberaudo:
Thank you for the review @ahocevar.

elemoine:
@gberaudo, what about the WebGL replay? Should we add the same sort of fix?

gberaudo:
Indeed @elemoine, see https://github.com/openlayers/ol3/pull/3975

PR #3962, opened 2015-08-04
Upgrade JSDoc to 3.3.2 and remove obsoleted plugins

bjornharrtell:
On visual inspection the docs looks identical to the ones generated with 3.3.0-alpha9 and the now obsolete plugins.


tschaub:
Thanks @bjornharrtell!

tschaub:
It looks like the `inheritdoc` plugin is still actually needed.  This change makes it so methods with `@inheritDoc` are not properly documented.  See #4082.

PR #3964, opened 2015-08-04
Compute new image extent for comparison to existing image extent

### LINKED ISSUE #3334: ImageWMS infinite loop bug
I've been struggling with an issue where an ImageWMS layer will occasionally get stuck in the "loading" imageState, regardless of the fact that the actual image load has completed.  Once this has happened, a render loop continuously fires until the viewport is changed and the layer redrawn.  I've tried several things to detect and work around this issue, but none of them have worked.  Has anyone else ran into this problem?  It's occuring in Chrome for me, haven't tested other browsers.

### LINKED ISSUE #3963: ol.source.ImageWMS gets stuck loading image for certain extents
The `getImage` function checks if the source already has an image to respond with.  This includes an extent comparison.
https://github.com/openlayers/ol3/blob/f5cd9a3ebac748d248e04b30688db447435d23ab/src/ol/source/imagewmssource.js#L206

However, this extent comparison incorrectly compares the image.getExtent() with the raw extent requested.  Instead it should compare the image.getExtent() with an "image extent" computed from the raw extent.  "Image extents" are adjusted to allow for integer pixel heights/widths.

For certain extents, the extent comparison will never pass and renderring of the ImageWMS will end up in an infinite loop....repeatedly creating new ol.Image instances that never finish loading.

rattai:
This fixes #3963 and fixes #3334.

### DIFF
--- a/src/ol/source/imagewmssource.js
+++ b/src/ol/source/imagewmssource.js
@@ -94,6 +94,12 @@ ol.source.ImageWMS = function(opt_options) {
 
   /**
    * @private
+   * @type {ol.Extent}
+   */
+  this.imageExtent_ = null;
+
+  /**
+   * @private
    * @type {ol.Size}
    */
   this.imageSize_ = [0, 0];
@@ -198,6 +204,24 @@ ol.source.ImageWMS.prototype.getImage =
     pixelRatio = 1;
   }
 
+  var image = this.image_;
+  if (!goog.isNull(image) &&
+      this.renderedRevision_ == this.getRevision() &&
+      image.getResolution() == resolution &&
+      image.getPixelRatio() == pixelRatio &&
+      ol.extent.containsExtent(image.getExtent(), this.imageExtent_)) {
+    return image;
+  }
+
+  var params = {
+    'SERVICE': 'WMS',
+    'VERSION': ol.DEFAULT_WMS_VERSION,
+    'REQUEST': 'GetMap',
+    'FORMAT': 'image/png',
+    'TRANSPARENT': true
+  };
+  goog.object.extend(params, this.params_);
+
   extent = extent.slice();
   var centerX = (extent[0] + extent[2]) / 2;
   var centerY = (extent[1] + extent[3]) / 2;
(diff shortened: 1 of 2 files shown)


ahocevar:
Isn't this issue already fixed with #3933?

rattai:
@ahocevar, thanks.  I'll try your fix, but I think my change might still be valid.
Please, take a look at the change and the test that fails without it.

ahocevar:
Agreed @rattai. If you haven't already, please submit a [CLA](https://github.com/openlayers/cla) so we can merge this.

rattai:
My company previously listed me on a corporate CLA.  I also just submitted an individual one; I might end up contributing for work and on my own in the future.

ahocevar:
Thanks @rattai. To improve the pull request, you could also cache the image extent, e.g. as private `imageExtent_` member, instead of calculating it with every `getImage()` call.

rattai:
@ahocevar, how would I know that the cached extent is valid for the args passed to `getImage()`?  I guess I could cache the extent with an ID derived from the args.

If you're concerned about performance, I could try to squeeze what I can out of the extent calculation.  However, the only redundant computations I see are `imageResolution * width / 2` and `imageResolution * height / 2` on lines 220-223.

ahocevar:
In https://github.com/openlayers/ol3/pull/3964/files#diff-4e8b936e2e779b01c7ad31bef5e9aa6aR249, the image is assigned. There you could do the calculation. ol.source.ImageWMS does not have a real cache of images, it only stores one.

rattai:
I'm a little confused.  We're talking about a pretty small function; maybe you could clarify with some code of what you have in mind?

The whole point of my change is to calculate an *image extent* from the *passed extent* **before** the `containsExtent()` check: https://github.com/openlayers/ol3/pull/3964/files#diff-4e8b936e2e779b01c7ad31bef5e9aa6aR230

ahocevar:
@rattai 0b0ba6f822f1930d01f91381f8d4a1934598b30f is what I had in mind.

rattai:
Ah, now I get it.  Thank you, @ahocevar!
Looks good to me.  I've updated the PR with your exact suggestion.

elemoine:
The extent is already stored on `this.image_` so I fail to understand why we need to cache the current image extent on the source. Am I missing something?

ahocevar:
Thanks for double ckecking @elemoine. You are absolutely right. Please review #3971, which reverts the commit that adds this caching.

PR #3969, opened 2015-08-05
Add an option to use the mouse's location as an anchor when zooming

samuellapointe:
By default, zooming in or out zooms using the mouse's location as an anchor, which makes it possible to navigate the map using scrolling only, and makes zooming in more intuitive. 

This patch adds a new `useAnchor` option to the MouseWheelZoom interaction. It is enabled by default, which changes nothing to the current behavior of the interaction. When disabled, the interaction stops registering its lastAnchor_ value. This turns the zoom into a very basic one, that zooms to the center of the screen. 

The reason I am creating this patch is for a @mapgears project involving tracking points as they move on the map. We want to follow a point as it moves on the map, and stop the user from moving the map around as the point is followed. Leaving the zoom enabled would let the user move around using the mouse scroll, which could be confusing if they get lost and can't come back using panning.

A good example is [this demo](http://openlayers.org/en/master/examples/geolocation-orientation.html) on the openlayers website, where the same issue was solved by relocating to the point after zooming in or out.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2648,7 +2648,8 @@ olx.interaction.ModifyOptions.prototype.wrapX;
 
 
 /**
- * @typedef {{duration: (number|undefined)}}
+ * @typedef {{duration: (number|undefined),
+ *     useAnchor: (boolean|undefined)}}
  * @api
  */
 olx.interaction.MouseWheelZoomOptions;
@@ -2660,6 +2661,16 @@ olx.interaction.MouseWheelZoomOptions;
  * @api
  */
 olx.interaction.MouseWheelZoomOptions.prototype.duration;
+
+
+/**
+ * Enable zooming using the mouse's location as the anchor. Default is `true`.
+ * When set to false, zooming in and out will zoom to the center of the screen
+ * instead of zooming on the mouse's location.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.interaction.MouseWheelZoomOptions.prototype.useAnchor;
 
 
 /**
--- a/src/ol/interaction/mousewheelzoominteraction.js
+++ b/src/ol/interaction/mousewheelzoominteraction.js
@@ -41,6 +41,13 @@ ol.interaction.MouseWheelZoom = function(opt_options) {
 
   /**
    * @private
+   * @type {boolean}
+   */
+  this.useAnchor_ = goog.isDef(options.useAnchor) ?
+      options.useAnchor : true;
+
+  /**
+   * @private
    * @type {?ol.Coordinate}
    */
   this.lastAnchor_ = null;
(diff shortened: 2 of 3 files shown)


externs/olx.js
```diff
@@ -2647,6 +2647,14 @@ olx.interaction.MouseWheelZoomOptions.prototype.duration;
 
 
 /**
+ * Enable zooming using the mouse's location as the anchor. Default is `true`.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.interaction.MouseWheelZoomOptions.prototype.useAnchor;
```
adube: Add this new property in the options definition, above.

src/ol/interaction/mousewheelzoominteraction.js
```diff
@@ -119,3 +129,20 @@ ol.interaction.MouseWheelZoom.prototype.doZoom_ = function(map) {
...
+  this.useAnchor_ = false;
+};
+
+
+/**
+ * Enable using the mouse's location as an anchor when zooming
+ */
+ol.interaction.MouseWheelZoom.prototype.enableMouseAnchor = function() {
+  this.useAnchor_ = true;
+};
```
elemoine: What are `disableMouseAnchor` and `enableMouseAnchor` for? They are not exportable (no `@api` annotation), nor are they use internally. Can we do without them?

externs/olx.js
elemoine: It would be good to document the behavior when this option is set to `false`.

samuellapointe:
I added the `@api` annotation to `enableMouseAnchor` and `disableMouseAnchor`, I had simply forgotten them. I also added a line explaining what happens when `useAnchor` is set to false in olx.js

marcjansen:
I'd go with only one method: `setUseAnchor(boolean)`.

If you can also add some tests, and squash the commits into one single commit, this is good to go.

Thanks!

I am currently not able to check if we have a CLA from you.

adube:
Samuel works at Mapgears, who has a CLA dated of 2009/03/23.   Samuel was not present at that moment.  Is this valid ?  Do we need to update our CLA ?  Should he act as individual ?  Thanks.

marcjansen:
IANAL, but: he says he did this contribution for work, Mapgears, so this should be fine. A personal CLA never hurts, though.

samuellapointe:
I put all the commits into a single one and added three small tests to check the default value of useAnchor and to test the setUseAnchor function. 

For the CLA, as mentioned before by @adube, I wasn't on the original Mapgears CLA, but I'm fine joining it or going with a personal one, whichever works the best for you.

marcjansen:
A very nice first contribution, @samuellapointe. Thanks.

PR #3970, opened 2015-08-06
Pull builds from openlayers.org

tschaub:
This makes it so the JSFiddle and copyable examples use builds from openlayers.org.


elemoine:
Please merge.

PR #3971, opened 2015-08-06
Revert "Cache the image extent"

ahocevar:
This reverts commit eb2a30e46142394d3382ade61bfcedff62aef4f7.

### DIFF
--- a/src/ol/source/imagewmssource.js
+++ b/src/ol/source/imagewmssource.js
@@ -94,12 +94,6 @@ ol.source.ImageWMS = function(opt_options) {
 
   /**
    * @private
-   * @type {ol.Extent}
-   */
-  this.imageExtent_ = null;
-
-  /**
-   * @private
    * @type {ol.Size}
    */
   this.imageSize_ = [0, 0];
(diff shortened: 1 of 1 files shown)


src/ol/source/imagewmssource.js
```diff
@@ -246,7 +222,23 @@ ol.source.ImageWMS.prototype.getImage =
   extent[1] = centerY - imageResolution * height / 2;
   extent[3] = centerY + imageResolution * height / 2;
 
-  this.imageExtent_ = extent;
+  var image = this.image_;
+  if (!goog.isNull(image) &&
+      this.renderedRevision_ == this.getRevision() &&
+      image.getResolution() == resolution &&
+      image.getPixelRatio() == pixelRatio &&
+      ol.extent.containsExtent(image.getExtent(), extent)) {
```
elemoine: But isn't the original issue back with this? Shouldn't `this.image_.getExtent()` be used instead of `extent` here?

ahocevar: The extent is adjusted to snap to pixels before being used. See passing tests.

elemoine: Oh I see.

elemoine:
LGTM. Thanks for taking my comment into account. (/me trying to catch up with ol development)

PR #3975, opened 2015-08-07
Skip features without geometry during webgl replay

gberaudo:
Setting a feature geometry to null or undefined and triggering a hit
detection was throwing an exception.

This patch handles a batched feature without a geometry in the same
fashion as skipped features.

An equivalent patch has been applied to the canvas replay.


elemoine:
LGTM. Can be merged when Travis is happy.

gberaudo:
Build is passing please merge @elemoine.

PR #3977, opened 2015-08-07
Mark ol.style.Icon#load() @api

gberaudo:
We currently rely on the renderer to load the styles it intents to use.

In Ol3-Cesium we want to allow different styles for 3D and for 2D.
This commit fixes the issue by allowing application code to explicitely
initiate the loading process.


gberaudo:
Any review?

elemoine:
How about adding a line of documentation saying that the application developer normally doesn't need to load the icon image himself?

gberaudo:
@elemoine, I added some documentation. Is it clear enough?

gberaudo:
Thank you for the quick review @elemoine.

PR #3982, opened 2015-08-07
Rebuild examples if package.json changes

tschaub:
Currently, `make examples` will do nothing when we update the version in `package.json`.  The `tasks/build-examples.js` script uses the version in `package.json` to write out the URLs to build artifacts.  So `package.json` needs to be listed as a dependency of any built example target.


elemoine:
+1

PR #3983, opened 2015-08-07
Remove doc footer with timestamp

tschaub:
Including a footer with a timestamp adds a ton of noise to diffs (e.g. https://github.com/openlayers/openlayers.github.io/commit/ff9e7d0ed6e0e589f6646219d5172ced9ece7661).  I'd like to make it easier to see significant changes to the website.


bartvde:
:+1:

probins:
Having the date is useful on master, as there's nothing else to tell you which version of master it refers to.

tschaub:
@probins this gives you the same information: https://github.com/openlayers/openlayers.github.io/commits/master (timestamp of latest push).  However, it would be nice to make the manual push unnecessary and instead have the website build be triggered by a push to ol3 master.

probins:
that's true; in fact, looking at the commit history is better, as it has the additional advantage that I can search for 'version' to find which commit this is based on if I want.

PR #3984, opened 2015-08-07
Style source control links in examples

tschaub:
This removes the two blue buttons for copy & edit in favor of more subtle links.

compare

Regarding the use of Font Awesome, I think it makes sense in this context.  I *don't* think it makes sense to have the example content rely on things like Bootstrap and Font Awesome.  I'd like to update things so you don't need Bootstrap and all the `container-fluid`, `row-fluid`, etc. just to reproduce these examples.  But that can happen separately.

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -4,12 +4,13 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="chrome=1">
     <meta name="viewport" content="initial-scale=1.0, user-scalable=no, width=device-width">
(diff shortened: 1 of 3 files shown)


marcjansen:
I like the new look better, Tim. Please merge once the checks are green.

tschaub:
Hmm.  Firefox isn't letting Zero Clipboard do its thing with this `a` element instead of `button`.  I'll dig a bit.

tschaub:
> Hmm. Firefox isn't letting Zero Clipboard do its thing with this a element instead of button. I'll dig a bit.

Disregard that.  Same issue with `button`.  Could be zeroclipboard/zeroclipboard#268 or something else, but copy is not working for me on master w/ Firefox 38 (downloading update now).

PR #3986, opened 2015-08-09
Modify 'url' option of ol.source.Vector to accept a function

alvinlindstam:
Based on my observations/assumptions, many (most?) uses of custom loaders on Vector sources are all about dynamically creating the URL. The other parts of the loader (handling the ajax request, read the features through an ol.format and adding them to the source) are fairly standard and unnecessary in many cases.

This can be seen in two of our examples that uses a custom loader (the other two also did some error notifications).

This PR suggests accepting a function as the *url* option to ol.source.Vector, enabling users to provide custom urls (generally based on the extent) and still use ol.featureloader.xhr.

### DIFF
--- a/examples/vector-wfs.js
+++ b/examples/vector-wfs.js
@@ -10,32 +10,20 @@ goog.require('ol.style.Stroke');
 goog.require('ol.style.Style');
 
 
-// format used to parse WFS GetFeature responses
-var geojsonFormat = new ol.format.GeoJSON();
 
 var vectorSource = new ol.source.Vector({
-  loader: function(extent, resolution, projection) {
-    var url = 'http://demo.boundlessgeo.com/geoserver/wfs?service=WFS&' +
+  format: new ol.format.GeoJSON(),
+  url: function(extent, resolution, projection) {
+    return 'http://demo.boundlessgeo.com/geoserver/wfs?service=WFS&' +
         'version=1.1.0&request=GetFeature&typename=osm:water_areas&' +
-        'outputFormat=text/javascript&format_options=callback:loadFeatures' +
-        '&srsname=EPSG:3857&bbox=' + extent.join(',') + ',EPSG:3857';
-    // use jsonp: false to prevent jQuery from adding the "callback"
-    // parameter to the URL
-    $.ajax({url: url, dataType: 'jsonp', jsonp: false});
+        'outputFormat=application/json&srsname=EPSG:3857&' +
+        'bbox=' + extent.join(',') + ',EPSG:3857';
   },
   strategy: ol.loadingstrategy.tile(ol.tilegrid.createXYZ({
     maxZoom: 19
   }))
 });
 
-
-/**
- * JSONP WFS callback function.
- * @param {Object} response The response object.
- */
-window.loadFeatures = function(response) {
-  vectorSource.addFeatures(geojsonFormat.readFeatures(response));
-};
 
 var vector = new ol.layer.Vector({
   source: vectorSource,
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -1,4 +1,5 @@
 goog.provide('ol.FeatureLoader');
+goog.provide('ol.FeatureUrlFunction');
 goog.provide('ol.featureloader');
 
 goog.require('goog.asserts');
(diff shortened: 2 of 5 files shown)


probins:
this looks similar to the discussion for #3833

src/ol/featureloader.js
```diff
@@ -19,7 +20,14 @@ ol.FeatureLoader;
 
 
 /**
- * @param {string} url Feature URL service.
+ * @api
+ * @typedef {function(ol.Extent, number, ol.proj.Projection) : string}
```
tschaub: It would be good to have a bit more documentation here.  The argument types don't make it through to the API docs (e.g. see http://openlayers.org/en/master/apidoc/ol.html#.TileUrlFunctionType), so some additional prose is helpful.

tschaub:
This looks like a nice addition to me.  Interested to hear what others think as well.

ahocevar:
I like it.

examples/vector-wfs.js
```diff
@@ -10,18 +10,14 @@ goog.require('ol.style.Stroke');
...
+  url: function(extent, resolution, projection) {
+    return 'http://demo.boundlessgeo.com/geoserver/wfs?service=WFS&' +
         'version=1.1.0&request=GetFeature&typename=osm:water_areas&' +
-        'outputFormat=text/javascript&format_options=callback:loadFeatures' +
-        '&srsname=EPSG:3857&bbox=' + extent.join(',') + ',EPSG:3857';
-    // use jsonp: false to prevent jQuery from adding the "callback"
-    // parameter to the URL
-    $.ajax({url: url, dataType: 'jsonp', jsonp: false});
+        'outputFormat=application/json&srsname=EPSG:3857&' +
+        'bbox=' + extent.join(',') + ',EPSG:3857';
```
elemoine: JSONP was used previously. There are no cross-origin issues when using Ajax in this example?

elemoine:
Sounds good to me. +1 for merging this when the comments are addressed.

examples/vector-wfs.js
alvinlindstam: demo.boundless.com should send liberal CORS headers, right @ahocevar?

https://github.com/openlayers/ol3-cesium/issues/127

alvinlindstam:
Added documentation for ol.FeatureUrlFunction and ol.FeatureLoader.

examples/vector-wfs.js
ahocevar: Indeed.

alvinlindstam:
Rebased to current master, which fixed the Travis error (which seemed unrelated to this code)

elemoine:
@ahocevar, @tschaub, should we merge this? +1 from me.

src/ol/featureloader.js
```diff
@@ -11,6 +12,16 @@ goog.require('ol.xml');
 
 
 /**
+ * {@link ol.source.Vector} sources use a function of this type to load
+ * features.
+ *
+ * This function takes an {@link ol.Extent} representing the area to be loaded,
+ * a `{number}` representing the pixel ratio and an {@link ol.proj.Projection}
```
tschaub: Minor, but I think "pixel ratio" risks being confused with `window.devicePixelRatio`.  Instead, it should read "representing the resolution (map units per pixel)" or something similar.

src/ol/featureloader.js
```diff
@@ -19,7 +30,21 @@ ol.FeatureLoader;
 
 
 /**
- * @param {string} url Feature URL service.
+ * {@link ol.source.Vector} sources use a function of this type to get the url
+ * to load features from.
+ *
+ * This function takes an {@link ol.Extent} representing the area to be loaded,
+ * a `{number}` representing the pixel ratio and an {@link ol.proj.Projection}
```
tschaub: Same as above regarding "pixel ratio."

tschaub:
Minor comments above about language in the docs.  But otherwise I think this would be great to have in.  Thanks for the contribution @alvinlindstam.

ahocevar:
+1 from me as well.

alvinlindstam:
Updated with @tschaub's documentation suggestions.

elemoine:
Thanks again for the patch. Merging.

PR #3998, opened 2015-08-12
Simplify tileLoadFunction and crossOrigin handling

ahocevar:
It will be easier to work with tile classes other than `ol.ImageTile` when `crossOrigin` and `tileLoadFunction` are optional constructor parameters. In the case of the `tileLoadFunction`, the default should be provided by the tile class, not the tile source. And in the case of `crossOrigin`, providing `null` as value does not make sense, except for the OSM source, where uses must be able to unset the 'anonymous' default.

This change is a minor refactoring for introducing a new vector tile class, which I am currently working on.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,14 @@
 ## Upgrade notes
 
+#### Changed return type of `ol.source.TileImage#getTileLoadFunction()`
+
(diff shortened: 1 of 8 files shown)


ahocevar:
This is work in progress and not ready for review.

ahocevar:
Closing this for now, as it's not going to help really.

PR #2490, opened 2014-08-03
Export PDF's using jsPDF

gingerik:
As promised, here is an example on using the jsPDF library to creating PDF's from an `ol.Map`.

Each commit adds some additional feature to the export:

* The first only adds an example that exports a PDF, similar to the existing PNG example;
* The second commit gives the user the options to select an output size and resolution;
* As increasing the map size and/or resolution requires many more tiles, the third commit ensures these tiles have been loaded before exporting the map;
* The fourth & fifth commit simply restore the example so another export can be made.

PR #3999, opened 2015-08-12
Export PDF's using jsPDF

bartvde:
An example of using the jsPDF library to creating PDF's from an ol.Map
Based on #2490


bartvde:
Something got messed up in this branch, closing.

PR #4001, opened 2015-08-13
Export PDF's using jsPDF

bartvde:
This work is a follow-up to #2490 by @gingerik 
The example was updated to use the templates, and made some small modifications to the print button.

### DIFF
--- a/examples/export-pdf.js
+++ b/examples/export-pdf.js
@@ -1,0 +1,109 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.control');
+goog.require('ol.format.WKT');
+goog.require('ol.layer.Tile');
+goog.require('ol.layer.Vector');
+goog.require('ol.source.OSM');
+goog.require('ol.source.Vector');
+
+var raster = new ol.layer.Tile({
+  source: new ol.source.OSM()
+});
+
+var format = new ol.format.WKT();
+var feature = format.readFeature(
+    'POLYGON((10.689697265625 -25.0927734375, 34.595947265625 ' +
+        '-20.1708984375, 38.814697265625 -35.6396484375, 13.502197265625 ' +
+        '-39.1552734375, 10.689697265625 -25.0927734375))');
+feature.getGeometry().transform('EPSG:4326', 'EPSG:3857');
+
+var vector = new ol.layer.Vector({
+  source: new ol.source.Vector({
+    features: [feature]
+  })
+});
+
+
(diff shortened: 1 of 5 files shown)


marcjansen:
Back in #2490 we couldn't get an agreement on the added `@api` declarations. @elemoine wnated to look at it if I remember correctly.

bartvde:
yeah I was also wondering if we could have an event on the tile queue so we don't need to poll it but not sure

bartvde:
see https://github.com/bartvde/ol3/commit/a38721af36ece222a874bcd721180ca6fd057f54 for an alternative approach, this is using the generic change event (but this is only temporary to discuss this solution)

bartvde:
I do have situations in which I get black tiles in the PDF, but I thought I saw the same with the previous approach if I am not mistaken

screen shot 2015-08-13 at 11 22 33

ahocevar:
I may be missing something, but can't you use the 'loadend' event that we have available now?

bartvde:
@ahocevar you mean ```tileloadend``` ?

ahocevar:
@bartvde yes, 'tileloadend', sorry.

bartvde:
@ahocevar good point, I think that can work as well (will push a commit soon). It leads to a bit more application code. The original PR was probably done before we had the load events. I am still seeing some black spots in the PDF as well in this case.

gingerik:
Good to see this PR is being picked up again! I think the black tiles are caused by a delay between loading images and actually drawing them onto the canvas. Let me know when I can still help :-)

examples/export-pdf.js
```diff
@@ -0,0 +1,100 @@
...
+  map.once('postcompose', function(event) {
+    source.on('tileloadstart', function() {
+      ++loading;
+    });
+    source.on('tileloadend', function() {
+      ++loaded;
+      if (loading === loaded) {
+        loading = 0;
+        loaded = 0;
+      var canvas = event.context.canvas;
```
marcjansen: Why didn't `gjslint` belch here and in the following lines?

bartvde: Good point ;-) btw I still need to clean this up a bit.

bartvde:
sorry for the wait but this is now ready for review cc @marcjansen

marcjansen:
Sorry this slipped through, Bart. I do not have time for a big review, but it is 'just' an example… so: :+1: please merge.

bartvde:
thanks @marcjansen

PR #4002, opened 2015-08-13
Update closure-library to latest commit

fredj:
And 	remove `goog.style.getRelativePosition` workaround in ol.Map (thanks to @elemoine for his work on this)

Fixes https://github.com/openlayers/ol3/pull/3316

### DIFF
--- a/closure-util.json
+++ b/closure-util.json
@@ -1,3 +1,3 @@
 {
-  "library_url": "https://github.com/google/closure-library/archive/0011afd534469ba111786fe68300a634e08a4d80.zip"
+  "library_url": "https://github.com/google/closure-library/archive/a5f9b8a.zip"
 }
--- a/test/spec/ol/map.test.js
+++ b/test/spec/ol/map.test.js
@@ -271,12 +271,54 @@ describe('ol.Map', function() {
         });
       });
     });
+
+    describe('#getEventPixel', function() {
+
+      var target;
+
+      beforeEach(function() {
+        target = document.createElement('div');
+        target.style.position = 'absolute';
+        target.style.top = '10px';
+        target.style.left = '20px';
+        target.style.width = '800px';
+        target.style.height = '400px';
+
+        document.body.appendChild(target);
+      });
+      afterEach(function() {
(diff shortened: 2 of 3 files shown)


elemoine:
We need to wait a bit before merging this, as https://github.com/google/closure-library/pull/424 is still being discussed.

elemoine:
Our changes even got reverted: https://github.com/google/closure-library/commit/691aed7f1e6a9ddb15311abf2717447b8761e541.

elemoine:
The change has been rolled forward. See https://github.com/google/closure-library/commit/b2cc62aeec361ea1279db21b3ed70823375d8dad. So we're good. I just think we should add a test to prevent future regression (for example if this change is rolled back again). Could you do that @fredj? I can have a look if you can't.

closure-util.json
```diff
@@ -1,3 +1,3 @@
 {
-  "library_url": "https://github.com/google/closure-library/archive/0011afd534469ba111786fe68300a634e08a4d80.zip"
+  "library_url": "https://github.com/google/closure-library/archive/bbb7ae6.zip"
```
elemoine: We need to change the commit number again.

fredj:
@elemoine I've added a test. comments welcome.

test/spec/ol/map.test.js
```diff
@@ -271,12 +271,47 @@ describe('ol.Map', function() {
...
+            clientY: 200
+          }]
+        });
+        var position = map.getEventPixel(browserEvent.getBrowserEvent());
+        // 80 = clientX - target.style.left
+        expect(position[0]).to.eql(80);
+        // 190 = clientY - target.style.top
+        expect(position[1]).to.eql(190);
+
+        document.body.removeChild(target);
```
elemoine: Shouldn't you do that in an `afterEach` so it is always done, even if an exception occurs in the middle of the test? You'll also want to do the `appendChild` in `beforeEach` I guess.

elemoine:
Thanks @fredj. Please merge when you're done.

PR #4003, opened 2015-08-13
Allow multiple identical calls to layer.setMap function

### LINKED ISSUE #4000: Need to call setMap(map) twice to show the layer again
Steps to reproduce : 

1) Open [this js fiddle](http://jsfiddle.net/oterral/0yr2eyq0/1/)
2) Add points on the map
3) Click on hide button -> the layer disappears correctly
4) Click on show button -> nothing happens
5) Click twice on show button -> the layer appears

Expected:

on 4) the layer should appear

oterral:
Fix #4000

### DIFF
--- a/test/spec/ol/layer/layer.test.js
+++ b/test/spec/ol/layer/layer.test.js
@@ -602,6 +602,29 @@ describe('ol.layer.Layer', function() {
       map.dispatchEvent(new ol.render.Event('precompose', map, null,
           frameState, null, null));
       expect(frameState.layerStatesArray.length).to.be(0);
+
+      layer.setMap(null);
+      map.dispatchEvent(new ol.render.Event('precompose', map, null,
+          frameState, null, null));
+      expect(frameState.layerStatesArray.length).to.be(0);
+
+      layer.setMap(map);
+      map.dispatchEvent(new ol.render.Event('precompose', map, null,
+          frameState, null, null));
+
+      expect(frameState.layerStatesArray.length).to.be(1);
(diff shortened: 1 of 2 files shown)


elemoine:
Looking at this now…

elemoine:
Do you handle the case where the user does `layer.setMap(map1)` and then `layer.setMap(map2)`? I think it used to be handled.

elemoine:
How about https://github.com/openlayers/ol3/compare/master...elemoine:4003?

test/spec/ol/layer/layer.test.js
```diff
@@ -602,6 +602,29 @@ describe('ol.layer.Layer', function() {
...
+      expect(frameState.layerStates[goog.getUid(layer)]).to.equal(layerState);
+
+      layer.setMap(map);
+      map.dispatchEvent(new ol.render.Event('precompose', map, null,
+          frameState, null, null));
+
+      expect(frameState.layerStatesArray.length).to.be(2);
+      layerState = frameState.layerStatesArray[1];
+      expect(layerState.layer).to.equal(layer);
+      expect(frameState.layerStates[goog.getUid(layer)]).to.equal(layerState);
```
elemoine: I think we should separate out things and add new tests (i.e. new `it` functions) for the new use-cases under tests.

oterral:
> How about master...elemoine:4003?

Yes I've done that before but I was not sure if the layerStatesArray must be modify or not. With my PR when we do setMap(map) the 2nd time the layerStates array is not modified. But I don't know if it's relevant.

PR #4007, opened 2015-08-18
Do not use Function.prototype.bind in examples

elemoine:
PhantomJS 1.x does not have Function.prototype.bind, so do not use that function in examples.


bartvde:
LGTM

Issue #2946, opened 2014-11-13
ol.DrawEvent not implemented by ol.interaction.Modify


awahab07:
ol.DrawEvent is implemented by ol.interaction.Draw but ol.interaction.Modify is not implementing any event from ol.DrawEvent neither it has its own event type.

ol.interaction.Modify changes geometry of features and both features and geometry can be observed by listeners but it becomes crucial at certain points to hook up certain functionality when the ol.interaction selects, starts, ends and leaves the feature for modification.

It will be desired if ol.interaction.Modify maintains certain attributes in order to allow Undo / Redo functionality implementation compact.

ttsiodras:
I know this is sacrilege, but since there was no response all this time... I hacked it: 

    ol.interaction.Modify.prototype.handlePointerUp = function(evt) {
      inModifyMode = false;
      var segmentData;
      for (var i = this.dragSegments_.length - 1; i >= 0; --i) {
        segmentData = this.dragSegments_[i][0];
        this.rBush_.update(ol.extent.boundingExtent(segmentData.segment),
            segmentData);
      }
      // ttsiod ugly patch here
      this.map_.npMap_.modified();
      // ttsiod ugly patch ends
      return false;
    };

When I create the map, I monkey-patch a reference back to my wrapper class:

    this.map = new ol.Map({
    ...
    });
    this.map.npMap_ = this;

...and I now have callbacks into my class's `modified()` callback, emulating the missing 'modifyend' event.

awahab07:
I appreciate @ttsiodras's approach. But if you can't afford overriding ol.interaction.Modify.prototype.handlePointerUp, you can follow the approach implemented in http://erilem.net/ol3/drag-features/examples/drag-features.html

To mention specifically look how app.Drag.prototype.handleUpEvent is implemented to nullify a reference when drag is complete. I think this is the location that @barbalex needs.

carueda:
:+1: for this request, in particular for ability to listen to a specific event triggered when a modification (vertex has been moved, vertex has been added, or vertex has been deleted) has finished.

elemoine:
Closed with https://github.com/openlayers/ol3/pull/3461.

PR #4011, opened 2015-08-18
Make Modify interaction listen to feature changes

### LINKED ISSUE #3927: Make ol.interaction.Modify listen to geometry change events
Currently, the Modify interactions does not listen to change events on it's geometries. Altering the geometry after it's loaded by the Modify interaction will make the geometry out of sync with the interactions segment cache.

Is there a reason not to listen to the geometries? The primary source of changes to the geometries will be the interaction itself, so we would need to ignore those or unlisten during the modification.

Is adding a reloadFeature method to the interaction a better option, so that application devs that manually change the features are also responsible for triggering the reloads?

Related example: http://jsfiddle.net/rbha7f83/1/ (I used an intermediate ol.Collection in order to force the Modify interaction to reload the geometry after setting it's coordinates).

elemoine:
This commit makes the Modify interaction modify its segment data when the candidate features change.

Fixes #3927.

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -4,6 +4,7 @@ goog.require('goog.array');
 goog.require('goog.asserts');
 goog.require('goog.events');
 goog.require('goog.events.Event');
+goog.require('goog.events.EventType');
 goog.require('goog.functions');
 goog.require('ol.Collection');
 goog.require('ol.CollectionEventType');
@@ -169,6 +170,14 @@ ol.interaction.Modify = function(options) {
   this.snappedToVertex_ = false;
 
   /**
+   * Indicate whether the interaction is currently changing a feature's
+   * coordinates.
+   * @type {boolean}
+   * @private
+   */
+  this.changingFeature_ = false;
+
+  /**
    * @type {Array}
    * @private
    */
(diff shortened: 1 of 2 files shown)


ahocevar:
Looks good. I think you can safely merge this.

alvinlindstam:
I might be able to test it tomorrow. The question that prompted my issue was http://stackoverflow.com/questions/26878659/openlayers-3-how-to-register-feature-modified-event-as-featuremodified-in-open

alvinlindstam:
Perhaps relevant: https://github.com/openlayers/ol3/issues/2946

elemoine:
> I might be able to test it tomorrow.

Thanks. Here's a full build with my change: http://erilem.net/ol3/update-segment-data/build/ol.js.

elemoine:
> Perhaps relevant: #2946

I don't think so. #2946 was about making the Modify interaction trigger events. And this was addressed by https://github.com/openlayers/ol3/pull/3461.

elemoine:
> Looks good. I think you can safely merge this.

Thank you @ahocevar. I'll let @alvinlindstam test my changes before merging.

alvinlindstam:
Yes, it fixes the issues I had. I noticed https://github.com/openlayers/ol3/issues/4022 but it's unrelated to these changes.

elemoine:
Ok, merging this. Thanks everyone.

Issue #3886, opened 2015-07-08
ol.source.XYZ doesn't have a getUrls function


bartvde:

ahocevar:
With `ol.source.XYZ`, it is not guaranteed that a url is set. Users can configure a `tileUrlFunction` instead. In such cases, `#getUrls()` should simply return `null`.

On a side note: I don’t know your use case, but I would suggest calling the `tileUrlFunction` with a specific tile coordinate to get the url, instead of using `#getUrls()`.

bartvde:
I'm trying to remember my use case ;-) I should have put more detail in the ticket description

bartvde:
I was likely creating a unit test for https://github.com/boundlessgeo/ole/blob/master/src/LayerGenerator.js#L100:L107 similar to: https://github.com/boundlessgeo/ole/blob/3423e642a7071cface7d9217c3e30879858f651d/test/spec/LayerGenerator.test.js#L43

ahocevar:
Fixed with #4015.

PR #4015, opened 2015-08-19
Add a getUrls Method to ol.source.XYZ

weskamm:
added the `getUrls` method to get access to the `url` / `urls` property.
When using a `tileUrlFunction`, `getUrls` should return `null`.
This should fix issue #3886

### DIFF
--- a/src/ol/source/xyzsource.js
+++ b/src/ol/source/xyzsource.js
@@ -38,6 +38,12 @@ ol.source.XYZ = function(options) {
         tileSize: options.tileSize
       });
 
+  /**
+   * @private
+   * @type {!Array.<string>|null}
+   */
+  this.urls_ = null;
+
   goog.base(this, {
     attributions: options.attributions,
     crossOrigin: options.crossOrigin,
@@ -63,6 +69,18 @@ goog.inherits(ol.source.XYZ, ol.source.TileImage);
 
 
 /**
+ * Return the URLs used for this XYZ source.
+ * When a tileUrlFunction is used instead of url or urls,
+ * null will be returned.
+ * @return {!Array.<string>|null} URLs.
+ * @api
+ */
+ol.source.XYZ.prototype.getUrls = function() {
+  return this.urls_;
+};
+
+
+/**
  * Set the URL to use for requests.
  * @param {string} url URL.
  * @api stable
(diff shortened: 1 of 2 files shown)


src/ol/source/xyzsource.js
```diff
@@ -63,6 +69,16 @@ goog.inherits(ol.source.XYZ, ol.source.TileImage);
 
 
 /**
+ * Return the URLs used for this XYZ source.
+ * @return {!Array.<string>|null} URLs.
+ * @api
+ */
+ol.source.XYZ.prototype.getUrls = function() {
+  return goog.isDefAndNotNull(this.urls_) ? this.urls_ : null;
```
ahocevar: You can just return `this.urls_` here - it will be null already if it was not configured.

src/ol/source/xyzsource.js
```diff
@@ -38,6 +38,12 @@ ol.source.XYZ = function(options) {
         tileSize: options.tileSize
       });
 
+  /**
+   * @private
+   * @type {!Array.<string>|null}
+   */
+  this.urls_ = goog.isDefAndNotNull(options.urls) ? options.urls : null;
```
ahocevar: Just initialize with `null` - `#setUrls()` will populate it later.

src/ol/source/xyzsource.js
```diff
@@ -63,6 +69,16 @@ goog.inherits(ol.source.XYZ, ol.source.TileImage);
 
 
 /**
+ * Return the URLs used for this XYZ source.
```
ahocevar: Add a note that this will be null when a `tileUrlFunction` was configured instead of `url` or `urls`.

weskamm:
Thanks for feedback, i changed the lines you mentioned

ahocevar:
Thanks. Can you please squash your changes into a single commit, and set a commit message that follows our [guidelines](https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#use-clear-commit-messages)?

Also, do we have a [CLA](https://github.com/openlayers/cla) from you? If not, please submit one.

weskamm:
Commits merged and message adjusted. The CCLA of terrestris should cover this PR

ahocevar:
@weskamm, has this work been conducted on behalf of terrestris? If not, we need an ICLA from you.

weskamm:
@ahocevar: Yes, work has been done on behalf of terrestris

PR #4016, opened 2015-08-19
Make ol.has.* not depend on ol.ENABLE_*

elemoine:
With this PR `ol.has.WEBGL` does not depend on the value of the `ol.ENABLE_WEBGL` define variable. Same for `ol.has.CANVAS`, `ol.has.DOM` (which I completely removed) and `ol.has.TOUCH`.

`ol.has` is about the capabilities of the browser, so it doesn't make sense that variables set in `ol.has` depend on compilation settings (`@define`).

For example, currently, a custom build with `ol.ENABLE_WEBGL` set to `false` will have `ol.has.WEBGL` set to `false` as well, even if the browser does support WebGL. And if the user also generates a debug build (with no `compile` section in the json config) then `ol.ENABLE_WEBGL` will be `true` and `ol.has.WEBGL`  will be `false` only if the browser does not support WebGL. This PR fixes that inconsistency issue.

With `ol.ENABLE_WEBGL` set to `false` the build is slightly bigger with my changes (131660 bytes versus 131841 bytes compressed, see below). This is because, with my changes, the build always includes the code that tests whether the browser supports WebGL.

Please review.

----

Full build size with `ol.ENABLE_WEBGL` and `ol.ENABLE_DOM` set to `false`. First is `master` branch. Second is `has` branch.

```bash
[wrk12 ol3 (master *$% u+1196)]$ make build/ol.js
info ol Parsing dependencies
info ol Compiling 356 sources
uncompressed: 441277 bytes
  compressed: 131660 bytes
``` 

```bash
[wrk12 ol3 (has *$%)]$ make build/ol.js
info ol Parsing dependencies
info ol Compiling 356 sources
uncompressed: 441523 bytes
  compressed: 131841 bytes
```

### DIFF
--- a/src/ol/has.js
+++ b/src/ol/has.js
@@ -78,7 +78,7 @@ ol.has.GEOLOCATION = 'geolocation' in goog.global.navigator;
  * @type {boolean}
  * @api stable
  */
-ol.has.TOUCH = ol.ASSUME_TOUCH || 'ontouchstart' in goog.global;
+ol.has.TOUCH = 'ontouchstart' in goog.global;
 
 
 /**
--- a/src/ol/has.js
+++ b/src/ol/has.js
@@ -30,7 +30,7 @@ ol.has.CANVAS_LINE_DASH = false;
  * @type {boolean}
  * @api stable
  */
-ol.has.CANVAS = ol.ENABLE_CANVAS && (
+ol.has.CANVAS = (
     /**
      * @return {boolean} Canvas supported.
      */
--- a/src/ol/has.js
+++ b/src/ol/has.js
@@ -115,28 +115,26 @@ ol.has.WEBGL;
 
 
 (function() {
-  if (ol.ENABLE_WEBGL) {
-    var hasWebGL = false;
-    var textureSize;
-    var /** @type {Array.<string>} */ extensions = [];
+  var hasWebGL = false;
+  var textureSize;
+  var /** @type {Array.<string>} */ extensions = [];
 
-    if ('WebGLRenderingContext' in goog.global) {
-      try {
-        var canvas = /** @type {HTMLCanvasElement} */
-            (goog.dom.createElement(goog.dom.TagName.CANVAS));
-        var gl = ol.webgl.getContext(canvas, {
-          failIfMajorPerformanceCaveat: true
-        });
-        if (!goog.isNull(gl)) {
-          hasWebGL = true;
-          textureSize = /** @type {number} */
-              (gl.getParameter(gl.MAX_TEXTURE_SIZE));
-          extensions = gl.getSupportedExtensions();
-        }
-      } catch (e) {}
-    }
-    ol.has.WEBGL = hasWebGL;
-    ol.WEBGL_EXTENSIONS = extensions;
-    ol.WEBGL_MAX_TEXTURE_SIZE = textureSize;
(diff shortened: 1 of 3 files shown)


elemoine:
@tschaub, are you able to have a look at this?

elemoine:
Anyone else up for a review?

ahocevar:
I may be missing the use case. But I am not so sure about this. To me, `ol.has.WEBGL` means whether OpenLayers has WebGL, not whether the browser has it. So I don't see anything wrong with the way things are now. On the contrary, people who want to use OpenLayers in a non-browser environment might intentionally turn WebGL checking off by setting the ol.ENABLE_WEBGL define to false, so OpenLayers does not execute any checks that cause an exception in a non-browser environment.

elemoine:
To me `ol.has` is not about what ol has, it's about the browser capabilities. Look at the other properties in that namespace.

Also, think about custom builds: ol-custom.js and ol-custom-debug.js. If `ol.ENABLE_WEBGL` is `false` in the build config file then `ol.has.WEBGL` will be always `false` when using ol-custom.js, while it will be `true` when using ol-custom-debug.js (and if the browser supports WebGL obviously). So the app may behave differently when using the debug build vs the non-debug build. I qualify this as a bug.

ahocevar:
I won't object this pull request, but it would be good if others could weigh in too. To me, it makes no sense to include WebGL detection code when the build is not intended for using WebGL. The only thing I'd maybe fix would be the documentation:
```js
/**
 * True when OpenLayers was compiled with `ol.ENABLE_WEBGL`, if the
 * browser supports WebGL.
 * @const
 * @type {boolean}
 * @api stable
 */
ol.has.WEBGL;
```

>  So the app may behave differently when using the debug build vs the non-debug build. I qualify this as a bug.

That's a different issue. We could change our debug build process to include compiler defines. This might be a problem elsewhere too (e.g. ol.ENABLE_PROJ4JS).

elemoine:
@ahocevar, just for the record, my use-case is an OpenLayers 3/Cesium application, where a Canvas-only custom build is used for OpenLayers 3. And I wanted to use `ol.has.WEBGL` to test if the browser supports WebGL, to avoid rewriting the code we already have in OpenLayers 3. This works with the debug build of OpenLayers 3, but it doesn't work with the non-debug build as `ol.has.WEBGL` is always `false`.

ahocevar:
@elemoine There are several npm packages for testing whether WebGL is available, so you don't have to rewrite anything. See e.g. https://www.npmjs.com/package/webgl-enabled.

elemoine:
Sure. How would you address the inconsistencies between the debug build and the prod build though?

ahocevar:
When we concatenate the debug build, we could append the contents of the `"compile"/"define"` array from the `conf.json` used for the build.

elemoine:
> we could append the contents of the compile.define array

But wouldn't this code come to late in the build?

ahocevar:
In some cases indeed. To make these work, that code would have to be inserted after `ol` is defined.

elemoine:
In any way, I think you've convinced me to stick with `ol.has.WEBGL` set to `true` meaning that "OpenLayers can use WebGL in that browser", which implies that both OpenLayers and the browser support WebGL. So I am going to close this.

PR #4017, opened 2015-08-19
Do not overwrite projections in the registry simply by using `new ol.proj.Projection()`

ahocevar:
When proj4js is loaded, creating a projection with a SRS code that is already registered will cause the existing projection in the registry to be overwritten. This can become a problem e.g. when someone creates a projection object with a custom extent for use it in a specific context. That projection object with its custom extent would be set in the registry, causing it to be returned whenever `ol.proj.get()` is used with its SRS code.

### DIFF
--- a/src/ol/proj/proj.js
+++ b/src/ol/proj/proj.js
@@ -141,8 +141,10 @@ ol.proj.Projection = function(options) {
    */
   this.defaultTileGrid_ = null;
 
-  if (ol.ENABLE_PROJ4JS && typeof proj4 == 'function') {
-    var code = options.code;
+  var projections = ol.proj.projections_;
+  var code = options.code;
+  if (ol.ENABLE_PROJ4JS && typeof proj4 == 'function' &&
+      !goog.isDef(projections[code])) {
     var def = proj4.defs(code);
     if (goog.isDef(def)) {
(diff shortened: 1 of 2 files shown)


bartvde:
This would probably break my applications such as here: https://github.com/boundlessgeo/wps-gui/blob/master/src/wpsui.js#L23:L30

bartvde:
False alarm, I don't think I use such an override anymore, but it would be good to make a note in the release notes since this can potentially break people's apps?

ahocevar:
In the snippet you linked, you use `ol.proj.addProjection()`. In this case whatever exists in the registry for that SRS code will still be overwritten.

My pull request only prevents that registry entries get overwritten by just creating a projection with `new ol.proj.Projection()`. If someone has been relying on this, then I would say they were relying on a buggy behavior, so no need to mention this in the release notes.

bartvde:
Ah sorry then I misunderstood, I don't see a reason not to change the behaviour you point out.

ahocevar:
@bartvde I updated the ticket title.

bartvde:
LGTM @ahocevar

elemoine:
Out of curiosity, can you still overwrite a projection in the registry if you want to?

ahocevar:
Yes, using ol.proj.addProjection().

elemoine:
Thanks.

PR #4018, opened 2015-08-19
Fix rotate buttons in animation example

marcjansen:
The [animation-example](http://openlayers.org/en/v3.8.2/examples/animation.html) does not show icons for the rotate-buttons.

This PR suggests using two unicode characters for these buttons: [↻](http://www.fileformat.info/info/unicode/char/21bb/index.htm) and [↺](http://www.fileformat.info/info/unicode/char/21ba/index.htm).

I also added a `title`-attribute to the buttons.

* Before: 
before

* After:
after

### DIFF
--- a/examples/animation.html
+++ b/examples/animation.html
@@ -14,8 +14,8 @@ tags: "animation"
 </div>
 <div class="row-fluid">
   <div class="span12">
-    <button id="rotate-left"><i class="icon-arrow-left"></i></button>
-    <button id="rotate-right"><i class="icon-arrow-right"></i></button>
+    <button id="rotate-left" title="Rotate clockwise">↻</button>
+    <button id="rotate-right" title="Rotate counterclockwise">↺</button>
     <button id="rotate-around-rome">Rotate around Rome</button>
     <button id="pan-to-london">Pan to London</button>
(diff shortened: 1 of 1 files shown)


elemoine:
Thanks! Please merge.

marcjansen:
Cheers for the review, Éric.

fredj:
icons in http://openlayers.org/en/master/examples/brightness-contrast.html, http://openlayers.org/en/master/examples/export-map.html, http://openlayers.org/en/master/examples/hue-saturation.html are broken too.

font-awesome was updated to version 4.x but the css classes must be changed from `icon-xxx` to `fa fa-xxx`

marcjansen:
You want to do it @fredj?

tschaub:
I hope we can work toward not relying on Bootstrap, Font Awesome, etc. in the "body" of the examples.  Instead, I think it would be nice if the copy/paste part of the examples was vanilla JavaScript + OpenLayers with elements that have an acceptable default style .  In our hosted examples, we can add custom styling, but people should be able to reproduce them without loading external libraries.

marcjansen:
The changes in this PR are headed in that direction, or not?

PR #4024, opened 2015-08-20
Correct documentation for ol.FeatureStyleFunction

marcjansen:
Any `ol.FeatureStyleFunction` should return an array of styles, not a single style. The docs should be better here.

**Before:**

before

**After**

after


elemoine:
Please merge. Thanks.

marcjansen:
Thank you :tulip:

PR #4027, opened 2015-08-20
closure-compiler v20150729 compatibility

fredj:


tschaub:
This looks ok to me, but it's a bit unfortunate.

Can you update the title to better describe what this is about ("20150729 compatibility" is pretty obscure)?  It would also be nice to have the description explain why these `missingProvide` comments are here.

elemoine:
I am also interested to know why we need to add these `@suppress {missingProvide}` annotations.

fredj:
With the new compiler, every classes tagged with `@constructor` must by provided with `goog.provide`.

We can either add a `@suppress {missingProvide}` annotation or provide the class even if it is not used.

elemoine:
+1 for adding the `goog.provide`. We'll need them anyway when compiling together with ol3.

fredj:
`ol.SelectEvent` and `ol.ModifyEvent` should probably be renamed to `ol.interaction.SelectEvent` and `ol.interaction.ModifyEvent`

elemoine:
Very nice @fredj. LGTM.

fredj:
Thanks for the review @elemoine.

PR #4032, opened 2015-08-21
Improve ol.Overlay extensibility

gberaudo:
This PR:
- splits updatePixelPosition_() and make it protected;
- introduces a protected setVisible() function;
- makes autoPan_ protected.

This change is a simple refactoring and improve extensibity by using `@protected`.

### DIFF
--- a/src/ol/overlay.js
+++ b/src/ol/overlay.js
@@ -97,10 +97,10 @@ ol.Overlay = function(options) {
   this.element_.style.position = 'absolute';
 
   /**
-   * @private
+   * @protected
    * @type {boolean}
    */
-  this.autoPan_ = goog.isDef(options.autoPan) ? options.autoPan : false;
+  this.autoPan = goog.isDef(options.autoPan) ? options.autoPan : false;
 
   /**
    * @private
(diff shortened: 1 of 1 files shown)


src/ol/overlay.js
```diff
@@ -455,27 +455,48 @@ ol.Overlay.prototype.setPositioning = function(positioning) {
...
 
 /**
- * @private
+ * Modify the visibility of the element.
+ * @param {boolean} visible
+ * @protected
  */
-ol.Overlay.prototype.updatePixelPosition_ = function() {
+ol.Overlay.prototype.setVisibility = function(visible) {
+  if (this.rendered_.visible ^ visible) {
```
tschaub: Since we're not using bitwise operators for this type of comparison elsewhere, I think `!==` would be clearer here.

gberaudo: OK, commit updated.

src/ol/overlay.js
```diff
@@ -455,27 +455,48 @@ ol.Overlay.prototype.setPositioning = function(positioning) {
...
+ol.Overlay.prototype.setVisible = function(visible) {
+  if (this.rendered_.visible !== visible) {
+    goog.style.setElementShown(this.element_, visible);
+    this.rendered_.visible = visible;
+  }
+};
+
 
+/**
+ * Update pixel position
```
elemoine: Please add the final punctuation.

elemoine:
Looks good to me. I think this can be merged when the remaining comments are addressed. Thanks @gberaudo.

gberaudo:
Thank you for the review @tschaub and @elemoine.
I addressed all the comments and the PR is now ready for merging.

elemoine:
Thanks.

PR #4035, opened 2015-08-23
Remove unused sphere and ellipsoid methods

tschaub:
This removes `ol.Sphere` and `ol.Ellipsoid` methods that are not exportable and only used in the tests.


ahocevar:

PR #4036, opened 2015-08-23
Fix modify interaction event order

alvinlindstam:
`ol.interaction.Modify` emits `modifystart` and `modifyend` events. However, the timing of the events are not what would be expected.

Currently, the `modifystart` event is triggered at the `pointerdown` event (when on a feature). If the click is on a boundary (not a vertex), the down event also creates a new vertex. The `modifystart` event is triggered just after any new vertex is added. This PR changes this, so that the `modifystart` event is emitted before the geometries are modified. This event will be unchanged by this PR, except for the timing.

The `modifyend` event is currently triggered on the `pointerup` event, after the last drag event is handled. However, any potential vertex deletion is generally handled on the `click` events, after the `modifyend` has already been triggered. With this PR, the `modifyend` is not trigger until all geometry changes are done, including vertex deletions. A part from the timing, this PR will cause the `modifyend` to not always reference a `pointerup` event as it's `mapBrowserPointerEvent` attribute. If any vertex is deleted, the `modifyend` event will reference the event which triggered the deletion (generally a `click`, `singleclick` or `doubleclick`).

Fixes https://github.com/openlayers/ol3/issues/4022

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -10,6 +10,7 @@ goog.require('ol.Collection');
 goog.require('ol.CollectionEventType');
 goog.require('ol.Feature');
 goog.require('ol.MapBrowserEvent.EventType');
+goog.require('ol.MapBrowserPointerEvent');
 goog.require('ol.ViewHint');
 goog.require('ol.coordinate');
 goog.require('ol.events.condition');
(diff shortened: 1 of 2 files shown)


elemoine:
Looks solid to me.

elemoine:
The PR cannot be automatically merged. @alvinlindstam, can you please rebase your branch. I'll merge when done. Thanks.

ahocevar:
I rebased and merged, no further action required from you @alvinlindstam. Thanks for this improvement!

PR #4037, opened 2015-08-24
Don't install python packages in before_install

fredj:
Travis already installs the packages by default.
See http://docs.travis-ci.com/user/languages/python/#Travis-CI-uses-pip


elemoine:
+1

PR #4040, opened 2015-08-24
Better docs for ol.ENABLE_[WEBGL|CANVAS|DOM]

elemoine:
@ahocevar, this is my modest replacement for https://github.com/openlayers/ol3/pull/4016 :)

### DIFF
--- a/src/ol/has.js
+++ b/src/ol/has.js
@@ -25,7 +25,8 @@ ol.has.CANVAS_LINE_DASH = false;
 
 
 /**
- * True if browser supports Canvas.
+ * True if both the library and browser support Canvas.  Always `false`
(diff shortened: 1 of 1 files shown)


probins:
somewhat off-topic, but why is DOM not api?

elemoine:
> somewhat off-topic, but why is DOM not api?

Probably because it's currently useless for the user.

probins:
in which case, why not delete it?

ahocevar:
Nice improvement, thanks @elemoind. Please merge.

@probins `ol.has.DOM` is used internally (ok, only once, so could be replaced with `ol.ENABLE_DOM`. But that can be handled seperately if we decide to do so.

elemoine:
Thanks @ahocevar.

PR #4044, opened 2015-08-25
Add ol.interaction.Draw#continueDrawing

elemoine:
This commit adds a `continueDrawing` method to the draw interaction. This method makes it possible to continue the drawing a line string feature. It only works on line string features, where this functionality makes the most sense.

Here's an example on how to use `continueDrawing` to continue drawing a feature:

```js
var vectorSource = vectorLayer.getSource();
vectroSource.removeFeature(feature);
drawInteraction.continueDrawing(feature);
```

### DIFF
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -715,6 +715,33 @@ ol.interaction.Draw.prototype.abortDrawing_ = function() {
 
 
 /**
+ * Extend an existing geometry by adding additional points. This only works
+ * on features with `LineString` geometries, where the interaction will
+ * extend lines by adding points to the end of the coordinates array.
+ * @param {!ol.Feature} feature Feature to be extended.
+ * @api
+ */
+ol.interaction.Draw.prototype.extend = function(feature) {
+  var geometry = feature.getGeometry();
+  goog.asserts.assert(this.mode_ == ol.interaction.DrawMode.LINE_STRING,
+      'interaction mode must be "line"');
+  goog.asserts.assert(goog.isDefAndNotNull(geometry),
+      'feature must have a geometry');
+  goog.asserts.assert(geometry.getType() == ol.geom.GeometryType.LINE_STRING,
+      'feature geometry must be a line string');
+  var lineString = /** @type {ol.geom.LineString} */ (geometry);
+  this.sketchFeature_ = feature;
+  this.sketchCoords_ = lineString.getCoordinates();
+  var last = this.sketchCoords_[this.sketchCoords_.length - 1];
+  this.finishCoordinate_ = last.slice();
+  this.sketchCoords_.push(last.slice());
+  this.updateSketchFeatures_();
+  this.dispatchEvent(new ol.interaction.DrawEvent(
+      ol.interaction.DrawEventType.DRAWSTART, this.sketchFeature_));
+};
+
+
+/**
  * @inheritDoc
  */
 ol.interaction.Draw.prototype.shouldStopEvent = goog.functions.FALSE;
--- a/test/spec/ol/interaction/drawinteraction.test.js
+++ b/test/spec/ol/interaction/drawinteraction.test.js
@@ -781,12 +781,43 @@ describe('ol.interaction.Draw', function() {
       expect(coordinates[0][0][1]).to.roughlyEqual(20, 1e-9);
     });
   });
+
+  describe('extend an existing feature', function() {
+    var draw;
+    var feature;
+
+    beforeEach(function() {
+      draw = new ol.interaction.Draw({
+        source: source,
+        type: ol.geom.GeometryType.LINE_STRING
+      });
+      map.addInteraction(draw);
+      feature = new ol.Feature(
+          new ol.geom.LineString([[0, 0], [1, 1], [2, 0]]));
+    });
+
+    it('sets the initial state', function() {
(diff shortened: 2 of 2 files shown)


elemoine:
Build passed on Travis. This is now ready for review.

alvinlindstam:
This would serve the use case of being able to provide an existing feature to the draw interaction? Am I understanding things correctly?

I find the method name continueDrawing slightly confusing, as it indicates that you wish to continue something you're already doing. The use case might also be to add to existing features, not just continue a mistakenly finished drawing.

Also, why would this only be applicable to LineStrings?

elemoine:
> This would serve the use case of being able to provide an existing feature to the draw interaction? Am I understanding things correctly?

I think so.

The use-case is the following: you have a line drawn in a vector layer, and you want to continue drawing the line. You don't want to modify the line (by moving vertices), you simply want to add extra points past the current end point. 

> I find the method name continueDrawing slightly confusing, as it indicates that you wish to continue something you're already doing. The use case might also be to add to existing features, not just continue a mistakenly finished drawing.

Right. But I am not sure `continueDrawing` is a bad name. What name would you suggest? `draw` or `drawFeature`?

> Also, why would this only be applicable to LineStrings?

We could see this as a current limitation. But it doesn't make sense for points. There may be use-cases for polygons but it is less obvious to me.

probins:
`extendLineString`?

This is something I could use too. Lines I'm drawing are frequently wrongly closed in the middle of drawing because it closes if there are 2 clicks in roughly the same place rather than on a double click.

elemoine:
> extendLineString?

I find it too specific. Although extending polygons is probably not a common use-case, this is something we can consider in the future.

So far, `continueDrawing` sounds like the most natural name too me.

probins:
`extend` then. I agree with @alvinlindstam: this is not necessarily continuing something you have done, but just adding to any feature. Logically, it belongs in `interaction.modify`, though it would make no sense to duplicate the drawing code in `modify`.

elemoine:
I agree with using `extend`.

More comments? Anyone?

src/ol/interaction/drawinteraction.js
```diff
@@ -715,6 +715,29 @@ ol.interaction.Draw.prototype.abortDrawing_ = function() {
...
 
 /**
+ * Continue drawing a feature.  This only works on line features, that is
+ * features with a `LINE_STRING` geometry.
+ * @param {!ol.Feature} feature Feature to continue drawing.
+ * @api
+ */
+ol.interaction.Draw.prototype.extend = function(feature) {
+  var geometry = feature.getGeometry();
+  goog.asserts.assert(this.mode_ == ol.interaction.DrawMode.LINE_STRING);
```
fredj: Do we want to have a message for all assertion?

src/ol/interaction/drawinteraction.js
```diff
@@ -715,6 +715,29 @@ ol.interaction.Draw.prototype.abortDrawing_ = function() {
 
 
 /**
+ * Continue drawing a feature.  This only works on line features, that is
+ * features with a `LINE_STRING` geometry.
+ * @param {!ol.Feature} feature Feature to continue drawing.
```
fredj: Is the `!` really needed? It's the only function in this class to have this no-null annotation

elemoine: I don't want the `extend` function to assume that `null`may be passed by the user. So by using `!ol.Feature` the user is responsible for providing a feature that is not `null`.

src/ol/interaction/drawinteraction.js
elemoine: Right, we have messages everywhere else.

src/ol/interaction/drawinteraction.js
```diff
@@ -715,6 +715,32 @@ ol.interaction.Draw.prototype.abortDrawing_ = function() {
 
 
 /**
+ * Continue drawing a feature.  This only works on line features, that is
+ * features with a `LINE_STRING` geometry.
```
tschaub: To users, the draw mode (and geometry types) look like `'LineString'` (instead of `LINE_STRING`).  If you change this, I think this doc would be more informative with something like:

> Extend an existing geometry by drawing additional points.  This only works on features with `'LineString'` geometries, where the interaction will extend lines by adding points to the end of the coordinates array.

tschaub:
Nice simple change for some good extra functionality.  I like the name `extend` as well.  One doc comment above, but I think this looks good to merge.

src/ol/interaction/drawinteraction.js
elemoine: Very nice description. Thanks.

src/ol/interaction/drawinteraction.js
fredj: `!ol.Feature` is perfectly valid. I'm just wondering if the compiler can detect that passing `null` is wrong because `getGeometry()` is called in this function.

Side note: the new version of the compiler has a check that prevents the use of the implicit nullable convention. Example:
```
ERR! compile /home/fredj/public_html/ol3/src/ol/map.js:103: ERROR - Name Array in JSDoc is implicitly nullable.
ERR! compile Please add a ! to make it non-nullable, or a ? to make it explicitly nullable.
ERR! compile  * @type {Array.<ol.RendererType>}
```

fredj:
good to merge, thanks

PR #4045, opened 2015-08-25
Add parser for dimension property from WMTS Capabilities.xml

Jenselme:
This add the possibility to parse the Dimension section of a WMTS Capabilities.xml with OpenLayers. I have a small question regarding tests: currently none of your xml used for testing have such a section. Should I add one with for instance:

```xml
<Dimension>
    <ows:Identifier>Time</ows:Identifier>
    <Default>20110805</Default>
    <Value>20110805</Value>
    <Value>20081024</Value>
</Dimension>
```

Or are these xml real xml from a server that shouldn't be modified?

### DIFF
--- a/src/ol/format/wmtscapabilitiesformat.js
+++ b/src/ol/format/wmtscapabilitiesformat.js
@@ -160,6 +160,18 @@ ol.format.WMTSCapabilities.readTileMatrixSetLink_ = function(node,
  * @private
  * @param {Node} node Node.
  * @param {Array.<*>} objectStack Object stack.
+ * @return {Object|undefined} Dimension object.
+ */
+ol.format.WMTSCapabilities.readDimensions_ = function(node, objectStack) {
+  return ol.xml.pushParseAndPop({},
+      ol.format.WMTSCapabilities.DIMENSION_PARSERS_, node, objectStack);
+};
+
+
+/**
+ * @private
+ * @param {Node} node Node.
+ * @param {Array.<*>} objectStack Object stack.
  * @return {Object|undefined} Resource URL object.
  */
 ol.format.WMTSCapabilities.readResourceUrl_ = function(node, objectStack) {
@@ -303,6 +315,8 @@ ol.format.WMTSCapabilities.LAYER_PARSERS_ = ol.xml.makeStructureNS(
           ol.format.XSD.readString),
       'TileMatrixSetLink': ol.xml.makeObjectPropertyPusher(
           ol.format.WMTSCapabilities.readTileMatrixSetLink_),
+      'Dimension': ol.xml.makeObjectPropertyPusher(
+          ol.format.WMTSCapabilities.readDimensions_),
       'ResourceURL': ol.xml.makeObjectPropertyPusher(
           ol.format.WMTSCapabilities.readResourceUrl_)
     }, ol.xml.makeStructureNS(ol.format.WMTSCapabilities.OWS_NAMESPACE_URIS_, {
@@ -344,6 +358,23 @@ ol.format.WMTSCapabilities.TMS_LINKS_PARSERS_ = ol.xml.makeStructureNS(
       'TileMatrixSet': ol.xml.makeObjectPropertySetter(
           ol.format.XSD.readString)
     });
+
+
+/**
+ * @const
+ * @type {Object.<string, Object.<string, ol.xml.Parser>>}
+ * @private
+ */
+ol.format.WMTSCapabilities.DIMENSION_PARSERS_ = ol.xml.makeStructureNS(
+    ol.format.WMTSCapabilities.NAMESPACE_URIS_, {
+      'Default': ol.xml.makeObjectPropertySetter(
+          ol.format.XSD.readString),
+      'Value': ol.xml.makeObjectPropertyPusher(
+          ol.format.XSD.readString)
+    }, ol.xml.makeStructureNS(ol.format.WMTSCapabilities.OWS_NAMESPACE_URIS_, {
+      'Identifier': ol.xml.makeObjectPropertySetter(
+          ol.format.XSD.readString)
+    }));
 
 
 /**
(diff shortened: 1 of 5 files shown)


elemoine:
I think you can just add a `Dimension` section to the existing XML document.

src/ol/format/wmtscapabilitiesformat.js
```diff
@@ -353,6 +367,23 @@ ol.format.WMTSCapabilities.TMS_LINKS_PARSERS_ = ol.xml.makeParsersNS(
...
  * @private
  */
+ol.format.WMTSCapabilities.DIMENSION_PARSERS_ = ol.xml.makeParsersNS(
+    ol.format.WMTSCapabilities.NAMESPACE_URIS_, {
+      'Default': ol.xml.makeObjectPropertySetter(
+          ol.format.XSD.readString),
+      'Value': ol.xml.makeObjectPropertyPusher(
+          ol.format.XSD.readString)
+    }, ol.xml.makeParsersNS(ol.format.WMTSCapabilities.OWS_NAMESPACE_URIS_, {
+      'Identifier': ol.xml.makeObjectPropertySetter(
```
bartvde: doesn't this belong in ```ol.format.OWS``` instead?

Jenselme: I confess I have on idea. It looks like it is specific to WMTS but I have no objection to move it. Has anyone else an opinion on that?

bartvde: ```ows:Identifier``` is not specific to WMTS. OWS contains common things for multiple OGC specs, see: https://github.com/openlayers/ol3/blob/master/src/ol/format/owsformat.js and how this is used in the WMTS parser already

Jenselme:
Just updated the PR to add tests.

src/ol/format/wmtscapabilitiesformat.js
Jenselme: I see various places where Identifier is parsed in src/ol/format/wmtscapabilitiesformat.js but none of them uses something from owsformat.js.

I tried to add a ol.format.OWS.readIdentifier in owsformat but I always get undefined. I fail to see how to do that.
- in owsformat:
```js
/**
 * @param {Node} node Node.
 * @param {Array.<*>} objectStack Object stack.
 * @returns {Object|undefined}
 */
ol.format.OWS.readIdentifier = function(node, objectStack) {
  return ol.xml.pushParseAndPop({}, ol.format.OWS.IDENTIFIER_PARSERS_,
      node, objectStack);
};


/**
 * @const
 * @type {Object.<string, Object.<string, ol.xml.Parser>>}
 * @private
 */
ol.format.OWS.IDENTIFIER_PARSERS_ = ol.xml.makeParsersNS(
    ol.format.OWS.NAMESPACE_URIS_, {
      'Identifier': ol.xml.makeObjectPropertySetter(
          ol.format.XSD.readString)
    });
```
- in wmtscapabilitiesformat.js:
```js
ol.format.WMTSCapabilities.DIMENSION_PARSERS_ = ol.xml.makeParsersNS(
    ol.format.WMTSCapabilities.NAMESPACE_URIS_, {
      'Default': ol.xml.makeObjectPropertySetter(
          ol.format.XSD.readString),
      'Value': ol.xml.makeObjectPropertyPusher(
          ol.format.XSD.readString)
    }, ol.format.WMTSCapabilities.OWS_NAMESPACE_URIS_, {
      'Identifier':  ol.xml.makeObjectPropertySetter(
          ol.format.OWS.readIdentifier)
    });
```

bartvde:
okay I see now, didn't know there was stuff already like this in the WMTS parser, in that case I can't really object to adding another one outside of the OWS parser ;-)

Jenselme:
I am not opposed to add it to owsformat but I will need some help. I also guess that I will have to "correct" all the current usage.

bartvde:
Right, I think we can consider that out of scope for your PR. I'm happy to reconsider my comment, no problem.

Jenselme:
Any plans on continuing this review?

elemoine:
Let's set this to the 3.10.0 release.

@Jenselme, this at least needs a rebase.

Jenselme:
@elemoine sounds great. I just rebased the PR.

ahocevar:
@elemoine or @bartvde, I'm moving this to the 3.11 milestone. If you want to have it in 3.10, you shoiuld merge it in the next hour or so.

elemoine:
I am sorry, but this needs rebasing. I'll review and merge once rebased.

Jenselme:
@elemoine PR rebased

src/ol/source/wmtssource.js
```diff
@@ -401,12 +401,12 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
   if ('Dimension' in l) {
     l['Dimension'].forEach(function(elt, index, array) {
       var key = elt['Identifier'];
-      var value = elt['default'];
-      if (value !== undefined) {
-        goog.asserts.assert(ol.array.includes(elt['values'], value),
+      var value = elt['Default'];
+      if (goog.isDef(value)) {
```
elemoine: We don't use `goog.isDef` anymore. Please change it to `value !== undefined`.

elemoine:
Thanks. I'll merge when the `goog.isDef` is changed.

Jenselme:
@elemoine commit updated to remove `goog.isDef`.

src/ol/source/wmtssource.js
```diff
@@ -401,12 +401,12 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
   if ('Dimension' in l) {
     l['Dimension'].forEach(function(elt, index, array) {
       var key = elt['Identifier'];
-      var value = elt['default'];
+      var value = elt['Default'];
       if (value !== undefined) {
-        goog.asserts.assert(ol.array.includes(elt['values'], value),
+        goog.asserts.assert(ol.array.includes(elt['Value'], value),
```
elemoine: Regarding this change. Can you confirm that there was a bug previously?

Jenselme: Yes I can confirm that. This correction is required for the parser to work correctly according to my tests.

elemoine:
Merged. Thanks for the patch.

PR #4047, opened 2015-08-26
Remove pystache dependency

fredj:
And use the mustache node module instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,5 +323,5 @@ build/test_rendering_requires.js: $(SPEC_RENDERING_JS)
 	@mkdir -p $(@D)
 	@node tasks/generate-requires.js $^ > $@
 
-%shader.js: %.glsl src/ol/webgl/shader.mustache bin/pyglslunit.py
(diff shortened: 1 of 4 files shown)


elemoine:
Could you please explain the change to bin/pyglslunit.py?

fredj:
`bin/pyglslunit.py` now returns a json string used by [mustache](https://www.npmjs.com/package/mustache) to generate the final js file.

elemoine:
> bin/pyglslunit.py now returns a json string used by mustache to generate the final js file.

I see.

Can you also confirm that the `mustache` node module won't be downloaded when doing `npm install openlayers`? I guess having `mustache` be a dev dependency will prevent that.

fredj:
> Can you also confirm that the `mustache` node module won't be downloaded when doing `npm install openlayers`? I guess having `mustache` be a dev dependency will prevent that. 

`mustache` is not download with:
```
npm install git://github.com/fredj/ol3#rm_pystache

```

elemoine:
Ok, thanks for testing. Feel free to merge this.

fredj:
thanks for the review

PR #4050, opened 2015-08-26
Use view.getRotation and view.getResolution instead of view.getState

fredj:
view.getState calls 4 functions and creates an object and an array.
It's more efficient to use view.getRotation and view.getResolution.


bartvde:
Nice catch, LGTM

fredj:
Thanks for the review

PR #2257, opened 2014-06-26
Reprojected extent requests for WMS sources

geonux:
Add the capability to handle layers from WMS servers, event if theses layers are not available in the projection of the View.
This is an approximative raster reprojection using the WMS server capabilities, good enough at high zoom levels.

For exemple, with these modifications, it is possible to have a WMS layer only available in the EPSG:4326 projection inside of a Global Mercator view.

PR #4054, opened 2015-08-27
Reprojected extent requests for WMS sources

geonux:
My problem was to load layers from WMS servers that are not available in the projection of the map View. In my example, I wanted to load a WMS layer that is only available in the EPSG:4326 projection inside of a EPSG:3857 view.

After several searches, I get inspired by the solution used by Leaflet to address this common problem.
The solution is very simple, as it is only a transformation of the coordinate system of the extend to a projection known by the server, which will then be used in the BBOX parameter of the WMS query. The HEIGHT ans WIDTH parameters are themselves not modified.
The sever then do the transformation needed to approximate the reprojection.

So I created this little patch that uses this capability of WMS servers via the definition of the reprojected BBOX for the requested tile.

I insist the fact that it's an approximation, which has its limits on lowest zoom levels, but works fine from zoom level 5-6.

Unlike the currently developed solution (by Bill Chadwick and his team), the reprojection is done on the server side, freeing the client from this task. There is no slowdown induced by the use of this feature.

To use it, simply set the projection of a raster layer, witch can be different from the EPSG code of the map.

For example :
```javascript
var map = new ol.Map({
  target: 'map',
  view: new ol.View({
    projection: 'EPSG:3857', // <---- HERE IS THE VIEW PROJECTION
    center: [0, 0],
    zoom: 2
  }),
  layers: [
    new ol.layer.Tile({
      source: new ol.source.TileWMS({
        projection: 'EPSG:4326', // <---- HERE IS THE DATA SOURCE PROJECTION
        url: 'http://demo.boundlessgeo.com/geoserver/wms',
        params: {
          'LAYERS': 'ne:NE1_HR_LC_SR_W_DR'
        }
    })
  })]
});
```

This is a pull request reopening the pull request #2257, closed due to inactivity and reorganization of the branches on my OL3 fork. Now everything is back to normal.

This pull request can be a response to :
https://groups.google.com/d/topic/ol3-dev/YmBOj623jHo/discussion
https://groups.google.com/d/topic/ol3-dev/Q3Ah7CgIpRM/discussion
https://groups.google.com/d/topic/ol3-dev/uA3gL02GP_0/discussion
https://groups.google.com/d/topic/ol3-dev/zzJJXcJ53OQ/discussion
https://groups.google.com/d/topic/ol3-dev/dDgTw2rSaR8/discussion

### DIFF
--- a/src/ol/source/tilewmssource.js
+++ b/src/ol/source/tilewmssource.js
@@ -159,6 +159,12 @@ ol.source.TileWMS.prototype.getGetFeatureInfoUrl =
         tileResolution * gutter, tileExtent);
   }
 
+  if (!goog.isNull(this.getProjection()) &&
+      this.getProjection() !== projection) {
+    tileExtent = ol.proj.transformExtent(tileExtent, projection, this.getProjection());
+    projectionObj = ol.proj.get(this.getProjection());
+  }
+
   var baseParams = {
     'SERVICE': 'WMS',
     'VERSION': ol.DEFAULT_WMS_VERSION,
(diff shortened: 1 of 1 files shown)


ahocevar:
#4122 is a more generic solution for reprojecting WMS sources, with exactly the same API, so I'm closing this.

PR #4055, opened 2015-08-27
Improve graticule perf

fgravin:
This PR aims to improve graticule rendering performance.

The graticule performance is related to 2 issues:

1. All meridians and parallels contained in the projection maxExtent are created, but only the ones which intersect the map extent are drawn.

2. All meridians and parallels that are drawn are actually drawned for the whole projection maxExtent, they are not cliped to the map extent. This create most performance issues on low resolutions rendering.

This PR fix issue 2.
On graticule initialization, we compute projected extent for which we are sure all points can be transformed to `EPSG:4326` projection. On graticule drawing, we take the intersection of this valid extent and the current extent, we get `maxLat minLat maxLon minLon` that will be send to meridians and parallels creation method to bound the geometry and make it fit the current map extent.

### DIFF
--- a/examples/sphere-mollweide.js
+++ b/examples/sphere-mollweide.js
@@ -16,7 +16,7 @@ var sphereMollweideProjection = new ol.proj.Projection({
   code: 'ESRI:53009',
   extent: [-9009954.605703328, -9009954.605703328,
     9009954.605703328, 9009954.605703328],
-  worldExtent: [-179, -90, 179, 90]
+  worldExtent: [-179, -89.99, 179, 89.99]
 });
 
 var map = new ol.Map({
--- a/src/ol/graticule.js
+++ b/src/ol/graticule.js
@@ -62,6 +62,30 @@ ol.Graticule = function(opt_options) {
    * @type {number}
    * @private
    */
+  this.maxLatP_ = Infinity;
+
+  /**
+   * @type {number}
+   * @private
+   */
+  this.maxLonP_ = Infinity;
+
+  /**
+   * @type {number}
+   * @private
+   */
+  this.minLatP_ = -Infinity;
+
+  /**
+   * @type {number}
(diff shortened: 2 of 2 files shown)


elemoine:
This is very nice. I just checked how the performance is with you branch, and it's very fast, both in the `graticule` and `sphere-mollweide` examples. Thanks a lot. I'll have a look at the code tomorrow.

examples/sphere-mollweide.js
```diff
@@ -16,7 +16,7 @@ var sphereMollweideProjection = new ol.proj.Projection({
   code: 'ESRI:53009',
   extent: [-9009954.605703328, -9009954.605703328,
     9009954.605703328, 9009954.605703328],
-  worldExtent: [-179, -90, 179, 90]
+  worldExtent: [-179, -89.99, 179, 89.99]
```
elemoine: How did you pick `-89.99`? Would it make sense to use `-89` as for the longitude?

bill-chadwick:
I did a whole load of work with the graticule earlier in the year to try and generalise it for a graticule of one projection on a map of another see here http://www.bdcc.co.uk/ol3/examples/graticule.html , http://www.bdcc.co.uk/ol3/examples/os-graticule.html and here http://www.bdcc.co.uk/ol3/examples/sphere-mollweide.html which use the graticule code here http://www.bdcc.co.uk/ol3/ol/ol/graticule.js

In retrospect I think this got a bit over complicated. Its EPSG:3857 graticule is perhaps of academic interest only - if wanted/needed it should be implemented separately. Perhaps the original graticule should become just a 4326 graticule and the local (non-global) graticules like the BNG and Swiss grids be supported from a second graticule class.

elemoine:
This is a nice improvement. Merging. Thanks @fgravin.

PR #4059, opened 2015-08-28
Clarify drag interaction example to inform about ol.interaction.Translate

alvinlindstam:
The example http://openlayers.org/en/v3.8.2/examples/drag-features.html is more or less a duplicate of ol.interaction.Translate from https://github.com/openlayers/ol3/pull/4008

Ideally, the example should probably be replaced by an example of using a custom interaction that is not implemented in the library. This PR only modifies the description, to inform the reader of ol.interaction.Translate and to focus a little bit more about it being a custom interaction than about the actual dragging.

### DIFF
--- a/examples/drag-features.html
+++ b/examples/drag-features.html
@@ -1,10 +1,11 @@
 ---
 template: example.html
-title: Drag features example
-shortdesc: Example of a drag features interaction.
+title: Custom interaction example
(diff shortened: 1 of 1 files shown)


elemoine:
Thanks a lot!

marcjansen:
Post merge +1!

This sort of housekeeping is sooo important, thanks @alvinlindstam

PR #4060, opened 2015-08-31
Remove unnecessary cast

fredj:
Found by the latest version of closure-compiler (v20150729)

### DIFF
--- a/src/ol/feature.js
+++ b/src/ol/feature.js
@@ -302,13 +302,10 @@ ol.FeatureStyleFunction;
  * @return {ol.FeatureStyleFunction} A style function.
  */
 ol.Feature.createStyleFunction = function(obj) {
-  /**
-   * @type {ol.FeatureStyleFunction}
-   */
   var styleFunction;
 
   if (goog.isFunction(obj)) {
-    styleFunction = /** @type {ol.FeatureStyleFunction} */ (obj);
+    styleFunction = obj;
   } else {
     /**
      * @type {Array.<ol.style.Style>}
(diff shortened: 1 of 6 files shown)


tschaub:
Looks good.  Please merge.

src/ol/feature.js
```diff
@@ -302,13 +302,10 @@ ol.FeatureStyleFunction;
...
  */
 ol.Feature.createStyleFunction = function(obj) {
-  /**
-   * @type {ol.FeatureStyleFunction}
-   */
   var styleFunction;
 
   if (goog.isFunction(obj)) {
-    styleFunction = /** @type {ol.FeatureStyleFunction} */ (obj);
+    styleFunction = obj;
```
elemoine: The typing here and above makes sense to me. This is about making sure styleFunction is set to an ol.FeatureStyleFunction. Does the compiler complain about that when a specific flag is used?

elemoine:
I just added a minor comment. Feel free to merge.

src/ol/feature.js
fredj: With the `unnecessaryCasts` check enabled, the error is:
```
ERR! compile /home/fredj/public_html/ol3/src/ol/feature.js:311: ERROR - unnecessary cast
ERR! compile from: function (this:ol.Feature, number): (Array<(null|ol.style.Style)>|null)
ERR! compile to  : function (this:ol.Feature, number): (Array<(null|ol.style.Style)>|null)
ERR! compile     styleFunction = /** @type {ol.FeatureStyleFunction} */ (obj);
```

PR #4062, opened 2015-08-31
Remove ol.xml.makeParsersNS and use ol.xml.makeStructureNS instead

fredj:


elemoine:
I fail to understand why we have two functions currently. Anyway, please merge.

PR #4064, opened 2015-08-31
Use a private variable to cache the default style array

### LINKED ISSUE #4063: ol.style.defaultStyleFunction lazy create not working
See #1815.
Code https://github.com/openlayers/ol3/blob/master/src/ol/style/style.js#L267-L281

The replaced function is never called because `ol.layer.Vector` stores a reference to this function before it is replaced.

fredj:
See #1815
Fixes #4063

### DIFF
--- a/src/ol/style/style.js
+++ b/src/ol/style/style.js
@@ -240,47 +240,44 @@ ol.style.createStyleFunction = function(obj) {
 
 
 /**
+ * @type {Array.<ol.style.Style>}
+ * @private
+ */
+ol.style.defaultStyle_ = null;
+
+
+/**
  * @param {ol.Feature} feature Feature.
  * @param {number} resolution Resolution.
  * @return {Array.<ol.style.Style>} Style.
  */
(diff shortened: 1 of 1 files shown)


gberaudo:
+1

src/ol/style/style.js
```diff
@@ -240,47 +240,39 @@ ol.style.createStyleFunction = function(obj) {
...
-      stroke: stroke
-    })
-  ];
-
-  // Now that we've run it the first time, replace the function with
-  // a constant version. We don't use an immediately-invoked function
-  // and a closure so we don't get an error at script evaluation time in
-  // browsers that do not support Canvas. (ol.style.Circle does
-  // canvas.getContext('2d') at construction time, which will cause an.error
-  // in such browsers.)
```
elemoine: I just think that we should leave that comment, or a similar comment, to explain why we don't want to eagerly create the default styles.

elemoine:
I just added a comment. Please merge when addressed.

fredj:
comment addressed, thanks for the review

Issue #4053, opened 2015-08-27
Issues handling unmanaged ol.layer.Layer


gberaudo:
Unmanaged layers answer an important use case (having one or some layers on top of the others)
but their big drawbacks is that they are not discoverable and listenable.

In OL3-Cesium, a map is setup and then its layers discovered by a call to `map.getLayers()`.
Afterwards, addition and removal of layers is watched and triggers listeners.

This is not possible in the case of unmanaged layers since they are not part of the `ol.Map`
but hidden to the outside and only watching for the map precompose event.

I see several ways to fix or workaround the issue:
1- Modify `ol.Map` to export an ol.Collection object containing the `unmanaged` layers;
2- Add a `z-index` mechanism to layers in addition to current unmanaged mechanism;
3- Add a `z-index` mechanism to layers instead of the current unmanaged mechanism;
4- Push the workaround responsability to the user.

1- Modify `ol.Map` to export an ol.Collection object containing the `unmanaged` layers;
The most simple:
In `ol.layer.Layer#setMap()`, we add/remove the layer to some new `ol.Map#unmanagedLayers` object.
This object can be discovered easily and watched.

2- Add a `z-index` mechanism to layers in addition to current unmanaged mechanism;
We allow the user to set a `z-index` property in the layer which we will use in the renderer
to determine the rendering order. In this case, the user will use `normal` layers which will
show up in `ol.Map#getLayers()`. The rendering order will depend both on the optional z-index value and the
actual sequence of the layers. For example [a, b1, c, d1] will be rendered as [a, c, b1, d1].

3- Add a `z-index` mechanism to layers instead of the current unmanaged mechanism;
This is the same as 2 but also removes unmanaged layers since their use case
would be handled with z-index.

4- Push the workaround responsability to the user.
This is the worst solution, where each user has to find out, understand and implement
a complex workaround.

What do you think about this issue and proposed solutions?
Am I missing something and do you see other ways?

ahocevar:
Unmanaged layers are the successor of feature overlays, which were also hidden from applications unless created by the application itself. ol3-cesium did not support feature overlays.

Sounds to me like we only need to improve documentation to make it clear to users what unmanaged layers are for.

gberaudo:
@ahocevar, supporting layers on top is a real use case and an expectation from our users.
They want support both at the application level and also indirectly through the use in OL3 itself.

What is the value of OL3 interactions if they only work in 2D?
With that in mind, I think only the first three solutions are practical.

[...]

PR #4069, opened 2015-09-01
Add Z-index to layers

gberaudo:
This commit allows setting a Z-index property on the layers and uses
it in the canvas, WEBGL and DOM map renderers for ordering the layers
before rendering.

Default z-index is `0` for managed layers and `1000` for unmanaged ones.
It allows always on bottom, always on top and more complex layer layouts.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2927,6 +2927,7 @@ olx.layer;
  *     saturation: (number|undefined),
  *     visible: (boolean|undefined),
  *     extent: (ol.Extent|undefined),
+ *     zIndex: (number|undefined),
  *     minResolution: (number|undefined),
  *     maxResolution: (number|undefined)}}
  * @api
@@ -2992,6 +2993,15 @@ olx.layer.BaseOptions.prototype.extent;
 
 
 /**
+ * The z-index for layer rendering.  At rendering time, the layers will be
+ * ordered, first by Z-index and then by position. The default Z-index is 0.
+ * @type {number|undefined}
+ * @api
+ */
+olx.layer.BaseOptions.prototype.zIndex;
+
+
+/**
  * The minimum resolution (inclusive) at which this layer will be visible.
  * @type {number|undefined}
  * @api stable
@@ -3016,6 +3026,7 @@ olx.layer.BaseOptions.prototype.maxResolution;
  *     source: (ol.source.Source|undefined),
  *     visible: (boolean|undefined),
  *     extent: (ol.Extent|undefined),
+ *     zIndex: (number|undefined),
  *     minResolution: (number|undefined),
  *     maxResolution: (number|undefined)}}
  * @api
@@ -3091,6 +3102,15 @@ olx.layer.LayerOptions.prototype.extent;
 
 
 /**
+ * The z-index for layer rendering.  At rendering time, the layers will be
+ * ordered, first by Z-index and then by position. The default Z-index is 0.
+ * @type {number|undefined}
+ * @api
+ */
+olx.layer.LayerOptions.prototype.zIndex;
+
+
+/**
  * The minimum resolution (inclusive) at which this layer will be visible.
  * @type {number|undefined}
  * @api stable
@@ -3114,6 +3134,7 @@ olx.layer.LayerOptions.prototype.maxResolution;
  *     saturation: (number|undefined),
  *     visible: (boolean|undefined),
  *     extent: (ol.Extent|undefined),
+ *     zIndex: (number|undefined),
  *     minResolution: (number|undefined),
  *     maxResolution: (number|undefined),
  *     layers: (Array.<ol.layer.Base>|ol.Collection.<ol.layer.Base>|undefined)}}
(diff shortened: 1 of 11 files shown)


gberaudo:
Closes:  #4053

TODO:
- [ ] document the new ordering behaviour;
- [ ] document the special 0 and 1000 values;
- [ ] convert the example to a unit test (check the color of the central point:
   OK if it is red; KO if it is blue).

gberaudo:
The PR is now ready for review.

src/ol/layer/layer.js
```diff
@@ -152,7 +152,7 @@ ol.layer.Layer.prototype.handleSourcePropertyChange_ = function() {
  * its layers collection, layer filters in {@link ol.Map#forEachLayerAtPixel}
  * will not filter the layer, and it will be rendered on top. This is useful for
  * temporary layers. To remove an unmanaged layer from the map, use
- * `#setMap(null)`.
+ * `#setMap(null)`. Unmanaged layers behave as if their Z-index is 1000.
```
elemoine: Does this mean a user is able to place some of his managed layers above unmanaged layers?

gberaudo: Indeed, layers with Z-index > 1000 will be rendered above unmanaged layers.

src/ol/layer/layer.js
```diff
@@ -172,6 +172,7 @@ ol.layer.Layer.prototype.setMap = function(map) {
         map, ol.render.EventType.PRECOMPOSE, function(evt) {
           var layerState = this.getLayerState();
           layerState.managed = false;
+          layerState.zIndex = 1000;
```
tschaub: Maybe `Infinity` would be less arbitrary.

tschaub:
> The extended ordering behaviour is documented in ol.layer.Base#setZIndex(), ol.layer.Layer#setMap(). What is the appropriate place to write a detailed explanation?

I think `zIndex` would make sense as a layer option (a new `olx.layer.LayerOption` property), in which case it would be documented [in `olx.js`](https://github.com/openlayers/ol3/blob/f5df71ba7057a738f986903339cc8781b60580cd/externs/olx.js#L3010-L3031).

tschaub:
This looks nice @gberaudo.  I look forward to trying it out.

elemoine:
> I think zIndex would make sense as a layer option (a new olx.layer.LayerOption property), in which case it would be documented in olx.js.

Agree.

src/ol/layer/layer.js
elemoine: I've thought about that too. But using a non-infinity value makes things more flexible to the user, as it is still possible to draw layers above unmanaged layers.

Also, with @gberaudo's current patch, calling `setZIndex` on an unmanaged layer has no effect. I think this should at least me documented. But it may also be great to support calling `setZIndex` on an unmanaged layer. An unmanaged layer would then just be a layer that the map doesn't know about (so you can silently add and remove layers if you want). Maybe I'm going to far here. And if people think I am indeed going too far then I'll go with `Infinity` as suggested by Tim.

gberaudo: There are two arbitrary values, `0` and `1000`, which delimitate three main ranges for the user:
- below 0 for layers to place under all others;
- above 1000 for layers to place on top of all others, including unmanaged layers;
- between 0 and 1000 for placing between them.
The proposition of @elemoine to allow setting Z-index on user unmanaged layers fits well with this vision. It provides the maximum flexibility to the user.

elemoine: > The proposition of @elemoine to allow setting Z-index on user unmanaged layers fits well with this vision. It provides the maximum flexibility to the user.

The implementation is not obvious to me though.

gberaudo: What about the following diff from `setMap()`?
```
-          layerState.zIndex = 1000;
+          var zIndex = layerState.zIndex;
+          layerState.zIndex = zIndex !== 0 ? zIndex : 1000;
```

elemoine: I had assumed that `layer.setZIndex(0)` would work as expected :)

gberaudo:
I updated the PR using `Infinity` instead of `1000` for unmanaged layer Z-index and updated documentation in olx externs.
If it is now good for you, please merge.

externs/olx.js
```diff
@@ -3091,6 +3102,15 @@ olx.layer.LayerOptions.prototype.extent;
 
 
 /**
+ * The z-index for layer rendering.  At rendering time, the layers will be
+ * ordered, first by Z-index and then by position. The default Z-index is 0.
+ * @type {number|undefined}
+ * @api
+ */
+olx.layer.LayerOptions.prototype.zIndex;
```
elemoine: Shouldn't `zIndex` be an option on `olx.layer.BaseOptions`?

elemoine: @gberaudo just made me realize that `zIndex` in on `BaseOptions` already.

elemoine:
Thanks for the excellent work @gberaudo!

gberaudo:
Thank you for the review @tschaub and @elemoine.

PR #4070, opened 2015-09-02
Make the debug server work regardless of the current working directory

tschaub:
With this change, you can run `serve-lib.js` from any directory, and it will provide a debug loader for the OpenLayers source files.  In addition, it will provide a static server for files in the current working directory.  This combination provides a zero config debug server for developing with the library (and is required for the workshop changes I'm making).


elemoine:
Please merge!

tschaub:
Thanks @elemoine.

PR #4072, opened 2015-09-02
Update to closure-util 1.7.0

elemoine:
With this PR we change from closure-util 1.6.1 to 1.7.0. This is to use the latest version of Closure Compiler (v20150729). See https://github.com/openlayers/closure-util/pull/67.


fredj:
+1, thanks

elemoine:
Thanks for the review @fredj.

PR #4076, opened 2015-09-03
Add assertion for required option 'code' on ol.proj.Projection

weskamm:
When using proj4js together with ol3 and instantiation of
ol.proj.Projection is made, the `code` property is required.
An empty string is enough though.

The reason can be found in the class ol.proj.Projection, starting in line 145:
https://github.com/openlayers/ol3/blob/master/src/ol/proj/proj.js#L145-L147

```javascript
var code = options.code;
  if (ol.ENABLE_PROJ4JS && typeof proj4 == 'function' &&
      !goog.isDef(projections[code])) {
...
```

Maybe an additional check if `code` is defined would make sense?

Problem can be observed in 
http://openlayers.org/en/master/examples/wms-image-custom-proj.html
by entering 

```javascript
new ol.proj.Projection({
    units: 'pixels'
});
```
in the console.

### DIFF
--- a/src/ol/proj/proj.js
+++ b/src/ol/proj/proj.js
@@ -143,6 +143,8 @@ ol.proj.Projection = function(options) {
 
   var projections = ol.proj.projections_;
   var code = options.code;
+  goog.asserts.assert(goog.isDef(code),
+      'Option "code" is required for constructing instance');
   if (ol.ENABLE_PROJ4JS && typeof proj4 == 'function' &&
       !goog.isDef(projections[code])) {
     var def = proj4.defs(code);


ahocevar:
This note should not be necessary, because "code" is not optional. Otherwise the type would be noted as `{string|undefined}`.

marcjansen:
I disagree. We should have a note in the docs stating it is not optional.

For example on http://openlayers.org/en/v3.8.2/examples/attributions.html (no proj4js)

```javascript
new ol.proj.Projection({});
```

works fine, while on http://openlayers.org/en/v3.8.2/examples/wms-image-custom-proj.html (with proj4js)

```javascript
new ol.proj.Projection({});
```

yields an Error:

```
Uncaught TypeError: Cannot use 'in' operator to search for 'EPSG' in undefined
    at Function.d [as defs] (http://cdnjs.cloudflare.com/ajax/libs/proj4js/2.3.6/proj4.js:1:15574)
    at new ye (http://openlayers.org/en/v3.8.2/build/ol.js:64:350)
    at <anonymous>:2:1
    at Object.InjectedScript._evaluateOn (<anonymous>:905:140)
    at Object.InjectedScript._evaluateAndWrap (<anonymous>:838:34)
    at Object.InjectedScript.evaluate (<anonymous>:694:21)
```

because we don't check for undefined before passing it to `proj4.defs(code)`. 

We may change the wording, but a note should be there IMO.

ahocevar:
When a constructor option is optional, it has the "undefined" type. This is not the case for "code". This is consistent across all our type definition in the API docs.

marcjansen:
I know that. But when we pass `undefined` to `proj4.defs` the error arises.

I cannot see why adding this documentation line causes so much discussion.

bartvde:
should we not just protect against passing off undefined to proj4.defs?

ahocevar:
+1 on @bartvde's suggestion. I'm going to create a pull request.

marcjansen:
I'd go even further and add an assertion, if the code is not optional as the type definition states:

```javascript
goog.asserts.assert(code !== undefined,
    'Option "code" is required for constructing instance');
```

marcjansen:
mayxbe the OP can create the PR? So more people play this game?

ahocevar:
Happy with that. It's just about adding goog.asserts.assert... in 1 or 2 places.

ahocevar:
Note to @weskamm: If you consider @marcjansen's snippet above, it should be `goog.isDef(code)`, not `code !== undefined`.

weskamm:
I added the assertion as suggested.
@ahocevar: will you have another look?
Thanks everybody.

ahocevar:
Thanks @weskamm. I meant to add the assertion and to remove the change to olx.js. The reason is that I don't know where we would stop when we start adding "Required" to all options in olx.js that do not have "undefined" in their type.

marcjansen:
Agreed on the removal of the 'required' in the docs (but only for consistency).

I think (but this is out of the scope of this PR) we should have the required options be marked as required in prose everywhere.

Thanks to @weskamm for the multiple stabs he does at getting this better, and to @ahocevar for pushing  his idea of consistency.

ahocevar:
Thanks! Will merge as soon as Travis says ok.

weskamm:
I removed the comment from the doc. Although i think it would be more userfriednly having it in..

ahocevar:
We can change the apidoc template to mark all required options as required automatically. No need to do this in prose for every required option.

marcjansen:
Nice, @ahocevar. I was thinking the same. Will you do it or shall I try?

ahocevar:
If you try @marcjansen it would be great.

PR #4075, opened 2015-09-03
setOpacity of Image undefined

bartvde:
V: 3.8.2
In the debug version I can use setOpacity of Image (which sets "opacity_") but in the min version, it's undefined (I have to set "v")

```javascript
ol.style.Image.prototype.setOpacity = function(opacity) {
  this.opacity_ = opacity;
};
```


bartvde:
PR attached

ahocevar:
Thanks. Please merge when Travis says so.

PR #4078, opened 2015-09-03
Enable all the compiler checks

fredj:


gberaudo:
Nice simplification: +1.

config/example.json
```diff
@@ -29,41 +29,7 @@
...
-      "suspiciousCode",
-      "strictModuleDepCheck",
-      "typeInvalidation",
-      "undefinedNames",
-      "undefinedVars",
-      "uselessCode",
-      "visibility"
-    ],
-    "jscomp_off": [
-      "unknownDefines"
```
elemoine: I am not sure about this one. There was a good reason for turning off `unknownDefines`. I need to remember what it was.

fredj: And this config file is missing the others `jscomp_off` checks

fredj:
testing now to compile all the examples with:
```for f in examples/*.js; do make build/compiled-examples/${f:9:-3}.combined.js; done```

elemoine:
Thanks!

fredj:
`unknownDefines` is necessary for the examples with the `NOCOMPILE` header

elemoine:
> unknownDefines is necessary for the examples with the NOCOMPILE header

Not only I think. Without that flag set to off I guess the compiler will complain on any example that does not pull `goog.json` (because the `goog.json.USE_NATIVE_JSON` is unknown to the compiler if `goog/json/json.js` is not sent to the compiler).

fredj:
> Not only I think. Without that flag set to off I guess the compiler will complain on any example that does not pull `goog.json` (because the `goog.json.USE_NATIVE_JSON` is unknown to the compiler if `goog/json/json.js` is not sent to the compiler).

Correct, I will add `unknownDefines` to `jscomp_off`

fredj:
> `unknownDefines` is necessary for the examples with the `NOCOMPILE` header

`unknownDefines` != `undefinedVars` !

elemoine:
Thanks @fredj. Please merge when Travis is done.

Issue #4082, opened 2015-09-06
API docs are missing detail where @inheritDoc is used


tschaub:
It looks like with 6812de6a77ca93b2067e1ef12f683c8be38f3491 we lost the functionality that assigns detail to symbols with `@inheritDoc`.

See for example the `ol.geom.LineString#getType` documentation [in master](http://openlayers.org/en/master/apidoc/ol.geom.LineString.html#getType):

![image](https://cloud.githubusercontent.com/assets/41094/9704854/570a9b9c-5471-11e5-91df-84bd9458a237.png)

And [in 3.8.2](http://openlayers.org/en/v3.8.2/apidoc/ol.geom.LineString.html#getType):

![image](https://cloud.githubusercontent.com/assets/41094/9704878/c1470b4e-5471-11e5-80d0-a328be830d55.png)

I think this is a significant enough regression (in doc usability) that it should be addressed for 3.9.0.

@ahocevar - I wonder if this is as simple as bringing back the inheritdoc plugin (I haven't tried yet).

elemoine:
What was the reason for removing the plugin? Did we think that it was not necessary anymore?

bjornharrtell:
Yes I thought I was not needed anymore, @inheritDoc should be supported by JSDoc since some time (included in an alpha that was later than the previously used version in ol3) as noted in https://github.com/jsdoc3/jsdoc/issues/53. Sorry I did not notice it caused problems.

tschaub:
I agree @bjornharrtell, it looks like it should no longer be necessary.  It would be nice to determine why the doc properties are not getting copied over properly.  But if the quick fix is to bring back the plugin/hack, I think we should do that.

bjornharrtell:
Agreed, I'm looking into it now.

bjornharrtell:
Looks like bringing back the plugin fixes the problem. PR #4096 brings it back.

PR #4083, opened 2015-09-06
Add API descriptions

tschaub:
This adds descriptions to exportable symbols.  There are 164 exported symbols with no description currently.  Most of this is due to #4082, there are also symbols with no descriptions.


elemoine:
Please merge.

PR #4084, opened 2015-09-06
Fewer circle exports

tschaub:
This removes the experimental `getAnchor`, `getOrigin`, and `getSize` methods from `ol.style.Circle`.  It is not possible to modify a circle's anchor or origin, so it should not be necessary to get these properties.  In addition, these properties and the size were not set until the circle was rendered.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,19 @@
 ## Upgrade notes
 
+### v3.9.0
+
+#### `ol.style.Circle` changes
+
+The experimental `getAnchor`, `getOrigin`, and `getSize` methods have been removed.  The anchor and origin of a circle symbolizer are not modifiable, so these properties should not need to be accessed.  The radius and stroke width can be used to calculate the rendered size of a circle symbolizer if needed:
+
+```js
+// calculate rendered size of a circle symbolizer
+var width = 2 * circle.getRadius();
+if (circle.getStroke()) {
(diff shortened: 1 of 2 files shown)


bartvde:
what about things like SLD export? I know we don't have this now in the library but people could be doing this in application code?

tschaub:
Note that this doesn't take away any existing functionality.  It does remove some unpredictable methods though.

The `circle.getAnchor()` and `circle.getSize()` methods would return `null` if the symbolizer had not yet been rendered.  After rendering, they would return something different.

The `circle.getOrigin()` method would return `[0, 0]` before rendering and *might* return something different after (depending on whether or not an atlas manager is being used).

This is different than the behavior of the same methods on different symbolizers.  And I think it makes them unsafe to use.  If there is existing application code that is using them, I think it would be better off being updated.

Again, you cannot **set** origin or offset related properties for a circle symbolizer.  You can only set the radius.  So I think it is very clear to have the only getter be `circle.getRadius()`.

> what about things like SLD export?

I've included upgrade notes about how people can calculate the rendered size - but this would not be used in SLD export.  Can you give an example of an `ol.style.Circle` symbolizer that cannot be represented in SLD with this change?

bartvde:
I've checked my code in https://github.com/MapStory/story-tools and I'm not using these exports anywhere, so LGTM and thanks for the detailed explanation

tschaub:
Thanks for the review.

PR #4093, opened 2015-09-08
Fix drag zoom interaction: Choose proper anchor for zoom

alvinlindstam:
Fixes https://github.com/openlayers/ol3/issues/4068

dragzoom

Before this PR, `ol.interaction.DragZoom` chose the proper resolution to zoom to, but always zoomed with the selected extent's center as anchor. This works fine as long as you select areas in the middle of the screen, but will focus om the wrong area when using it closer to the edges.

In this PR, I calculate an anchor which correlates to the selected area's position in the view.

### DIFF
--- a/src/ol/interaction/dragzoominteraction.js
+++ b/src/ol/interaction/dragzoominteraction.js
@@ -1,8 +1,8 @@
 goog.provide('ol.interaction.DragZoom');
 
 goog.require('goog.asserts');
+goog.require('goog.math');
 goog.require('ol.events.condition');
-goog.require('ol.extent');
 goog.require('ol.interaction.DragBox');
 goog.require('ol.interaction.Interaction');
 goog.require('ol.style.Stroke');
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks for this very nice pull request @alvinlindstam.  However, I think there are still issues with this approach.  The calculated anchor is based on the requested resolution, which may be different than the constrained resolution applied by `ol.interaction.Interaction.zoom`.  So in cases where the two differ, the anchor is incorrect (as I think you can see in your gif above).

It would be pretty straightforward to use your method with the constrained resolution, but I think this is a lot of unnecessary calculation just to try to use the same `ol.interaction.Interaction.zoom` function.  What we really want is just to animate to the new center and resolution.  I think a more straightforward approach is #4095.

tschaub:
Thanks again for your work on this @alvinlindstam.  I hope you agree with the approach taken in #4095.

alvinlindstam:
Agreed.

PR #4094, opened 2015-09-09
Allow source.setAttributions() to be exported

tschaub:
This allows people to modify attributions after source construction.


fredj:
+1

PR #4095, opened 2015-09-09
Zoom to the extent of the drag box

### LINKED ISSUE #4068: Broken drag zoom interaction
The `ol.interaction.DragZoom` doesn't zoom to the expected extent.

broken

tschaub:
Currently, the drag zoom interaction uses `ol.interaction.Interaction.zoom`, which requires that an anchor be specified.  This is only really appropriate for the pinch zoom interaction, where the user expects the center of their pinch to remain anchored.  Instead of trying to force the drag zoom interaction to use the same code, it is more straightforward to determine the center of the zoom box, calculate the constrained resolution, and animate the transition.

drag-zoom

The result is that the center of the zoom box becomes the new center (as you would expect).

Fixes #4068.

### DIFF
--- a/src/ol/interaction/dragzoominteraction.js
+++ b/src/ol/interaction/dragzoominteraction.js
@@ -1,10 +1,11 @@
 goog.provide('ol.interaction.DragZoom');
 
 goog.require('goog.asserts');
+goog.require('ol.animation');
+goog.require('ol.easing');
 goog.require('ol.events.condition');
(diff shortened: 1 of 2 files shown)


marcjansen:
I like this approach and the outcome, @tschaub. Please merge.

tschaub:
Thanks for the review @marcjansen.  I'll confirm the new tests pass and then merge.

PR #4096, opened 2015-09-09
Reinstate inheritdoc to fix #4082

### LINKED ISSUE #4082: API docs are missing detail where @inheritDoc is used

bjornharrtell:


elemoine:
Thanks a lot.

PR #4102, opened 2015-09-10
Export ol.Observable#dispatchEvent function

### LINKED ISSUE #3994: ol.Object no longer expose dispatchEvent() method in release version since v3.8.X
ol.Object no longer exposes dispatchEvent() - something I relied until 3.7.0 even though it was undocumented. This makes extending ol.* classes even harder then it already was - without closure compiling app along with library.

ps It would be nice if ol.interaction.Interaction could expose it's getMap() method as well.

fredj:
fixes #3994

### DIFF
--- a/src/ol/observable.js
+++ b/src/ol/observable.js
@@ -62,6 +62,18 @@ ol.Observable.prototype.changed = function() {
 
 
 /**
+ * Dispatches an event and calls all listeners listening for events
+ * of this type. The event parameter can either be a string or an
+ * Object with a `type` property.
+ *
+ * @param {goog.events.EventLike} event Event object.
+ * @function
+ * @api
+ */
+ol.Observable.prototype.dispatchEvent;
+
+
+/**
  * Get the version number for this object.  Each time the object is modified,
  * its version number will be incremented.
  * @return {number} Revision.


src/ol/observable.js
```diff
@@ -62,6 +62,18 @@ ol.Observable.prototype.changed = function() {
...
 /**
+ * Dispatches an event (or event like object) and calls all listeners
+ * listening for events of this type. The type of the event is decided by the
+ * type property on the event object.
+ *
+ * @param {goog.events.EventLike} event Event object.
+ * @function
+ * @api
+ */
+ol.Observable.prototype.dispatchEvent;
```
elemoine: It won't work if it's `undefined`.

fredj: it's not `undefined`: `ol.Observable.prototype.dispatchEvent;` has no effect at all. It only tells the export system to expose the function.

The same system is used elsewhere in ol3, for example: https://github.com/openlayers/ol3/blob/master/src/ol/format/geojsonformat.js#L374-L384

elemoine: I think it should look like this:

```js
ol.Observable.prototype.dispatchEvent = goog.events.EventTarget.prototype.dispatchEvent;
```

And also, we don't want to use `@param {goog.events.EventLike}`, to avoid exposing `goog` types in the API doc. So I think we need to use `@param {string}`, although this is a bit limitating.

elemoine: > It only tells the export system to expose the function.

I'll create an alias to `undefined`.

elemoine: Ok maybe it works, I see that we have the same thing for `ol.format.GeoJSON.prototype.readFeature` [for example](https://github.com/openlayers/ol3/blob/master/src/ol/format/geojsonformat.js#L384).

fredj:
It works as expected in development and production mode. the api doc looks good.

fredj:
Todo: write a better function description

jonataswalker:
Thanks for this export. Make things easier on OL3 plugins for example.

fredj:
Forget about this one ...
@elemoine good to merge ?

elemoine:
Yep!

fredj:
Thanks again for the review

PR #4106, opened 2015-09-14
Focus to search field on page load

elemoine:
This commit changes the examples index page to focus on the search field on the first "keydown" event. So with this you can open the index page and start typing to search.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -170,13 +170,13 @@
         for (var i = 0; i < info.examples.length; ++i) {
           info.examples[i].link += window.location.search;
         }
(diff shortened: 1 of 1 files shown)


fredj:
And what about adding an [autofocus](http://caniuse.com/#feat=autofocus) attribute to the input?

elemoine:
Good suggestion. I changed the commit.

elemoine:
iOS doesn't support "autofocus" but I think it's ok.

fredj:
LGTM, thanks

fredj:
coveralls check is stuck but it's safe to merge anyway

PR #4107, opened 2015-09-14
Prevent page zoom on IE Edge

### LINKED ISSUE #4105: Panning with touch doesn't work in Microsoft Edge browser
This applies for every example I can find: ol3 won't pan using a touch-screen and the new browser in Windows 10. Works fine with normal click-and-drag, and also with other browsers.

Just as one example: http://openlayers.org/en/v3.9.0/examples/attributions.html

fredj:
The deprecated msTouchAction property was removed in IE Edge.

Fixes #4105


ahocevar:
Looks good, thanks for fixing this. Please merge after Travis says so.

fredj:
Thanks for the review

PR #4111, opened 2015-09-15
Use a blackish default for filling texts

marcjansen:
This suggests a blackish (`#333333`) default fill for `ol.style.Text` if they are constructed without actually specifying a fill. 

Please review.

### DIFF
--- a/src/ol/style/textstyle.js
+++ b/src/ol/style/textstyle.js
@@ -1,6 +1,9 @@
 goog.provide('ol.style.Text');
 
 
+goog.require('ol.style.Fill');
+
+
 
 /**
  * @classdesc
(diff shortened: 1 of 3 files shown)


fredj:
`#333333` can be *compressed* to `#333`

marcjansen:
Color updated.

src/ol/style/textstyle.js
```diff
@@ -77,6 +81,16 @@ ol.style.Text = function(opt_options) {
 
 
 /**
+ * The default fill color to use if no fill was set at construction time.
+ *
+ * @type {string} The color of the default fill, `#333`.
+ * @const
+ * @private
```
elemoine: Not directly related to your PR but I've been wondering if `@type` is necessary when `@const` is used. https://github.com/google/closure-compiler/wiki/A-word-about-the-type-Object is not super clear to me, but it's made me wonder. I'll ask the question on the Closure Compiler mailing list.

elemoine:
LGTM.

src/ol/style/textstyle.js
fredj: You're right, we should write `@const {string}` here.

marcjansen:
I'll annotate as @fredj suggested and will then merge. Thanks for your thoughts.

src/ol/style/textstyle.js
elemoine: Not what I was thinking of. What makes you think we should that?

PR #4116, opened 2015-09-16
Make ol.control.OverviewMap's view configurable

ahocevar:
This is useful for creating an OverviewMap in a different projection than EPSG:3857.

This should fix #3647, by configuring the OverviewMap control with a `view` with a custom projection or resolutions.

@Sverre-Iversen, can you please test this and report if it works for you?


Sverre-Iversen:
It works - both the projection and resolution parameters are possible to set.

I see however that the rectangle representing the area in the main map is changing in shape when the main map is rotated.  A map rotation should not result in a change in the rectangle's shape if the size of the viewport is fixed.

ahocevar:
Thanks for testing. I would say the rectangle problem is a different issue that you may want to ticket separately.

ahocevar:
Thanks for any review.

bartvde:
LGTM please merge

PR #4117, opened 2015-09-16
forEachFeatureAtPixel shouldn't fail if layer has no source

pgiraud:
Fixes https://github.com/openlayers/ol3/issues/3902.
Please review.


marcjansen:
For readability it'd be nice if you could add `expect(function(){ /* your call */ }).to.not.throwException()`.

Otherwise this looks good to me.

marcjansen:
Nice & thanks Pierre

PR #4120, opened 2015-09-17
Remove invalid link to featureOverlay in API doc

kalbermattenm:
This PR removes a link to the `ol.FeatureOverlay` API documentation. `ol.FeatureOverlay` was removed with release 3.7.0 (see https://github.com/openlayers/ol3/releases/tag/v3.7.0), thus we have an invalid link on the API doc page: http://openlayers.org/en/v3.9.0/apidoc/


fredj:
Thanks

Issue #3785, opened 2015-06-10
Raster reprojection with triangulation and Proj4js


klokan:
The modern web browsers provides technology allowing client-side raster reprojection of images on the fly - in JavaScript. OpenLayers v3 API was designed with the possibility to view raster geodata in a different map projection then the one served from a remote source (typically WMTS or WMS server, MapTiler tiles, etc).

This ticket presents a proposal (resulting from discussions at the Schladming code sprint) for an implementation of the on the fly transformation of raster data which should be directly integrated with the future version of the OpenLayers library.

The usage in API is going to be very simple:
```javascript
var map = new ol.Map({
  target: 'map',
  view: new ol.View({
    projection: 'EPSG:3857', // <---- HERE IS THE VIEW PROJECTION
    center: [0, 0],
    zoom: 2
  }),
  layers: [
    new ol.layer.Tile({
      source: new ol.source.TileWMS({
        projection: 'EPSG:4326', // <---- HERE IS THE DATA SOURCE PROJECTION
        url: 'http://demo.boundlessgeo.com/geoserver/wms',
        params: {
          'LAYERS': 'ne:NE1_HR_LC_SR_W_DR'
        }
    })
  })]
});
```

If a source (inheriting from `ol.source.TileImage` or `ol.source.Image`) has projection different from the current `ol.View`’s projection the reprojection happens automatically under the hood.

The overview for implementation of the reprojection:

- The reprojection process will be based on triangles -- the target raster will be divided into a limited number of triangles with vertices transformed using proj4js. The reprojection of pixels inside the triangle extent will be approximated with an affine transformation (with rendering hw-accelerated by the canvas 2d context):

![4d099f62-fe18-11e4-814d-da0c19a6f1b2](https://cloud.githubusercontent.com/assets/59284/8087649/81758cbc-0f9c-11e5-97ce-47706459678b.jpg)

- This way we can support wide range of projections from proj4js on almost any hardware and any web browser with canvas 2D support - without a need to have CORS headers on the remote server.
- The precision of the reprojection is limited by the number of triangles. To minimise the error a dynamic triangulation will be utilised:

![5a8c75ba-fe18-11e4-94a4-db173f6ff0ab](https://cloud.githubusercontent.com/assets/59284/8088179/bf394d42-0f9f-11e5-9bcb-dcfbc8694a56.png)

[...]

PR #4122, opened 2015-09-17
Raster reprojection

klokan:
Implementation of the raster reprojection with triangulation with Proj4js as described and discussed in #3785.

The raster reprojection is triggered automatically when `projection` in `ol.View` differs from `projection` of a raster source such as WMTS or WMS.

Examples:
- [Raster reprojection demo](http://klokantech.github.io/ol3/rasterreproj/final/examples/reprojection.html)
- [OpenStreetMap to WGS84 reprojection](http://klokantech.github.io/ol3/rasterreproj/final/examples/reprojection-wgs84.html)
- [Reprojection with EPSG.io database search](http://klokantech.github.io/ol3/rasterreproj/final/examples/reprojection-by-code.html)
- [Image reprojection](http://klokantech.github.io/ol3/rasterreproj/final/examples/reprojection-image.html)

Tutorial for use is being prepared at:
https://github.com/klokantech/ol3/wiki/Raster-Reprojection-Tutorial
and to be pushed to official tutorials.

We may add an example with UK's Ordnance Survey OpenSpace API tiles as well.

There are related tickets to some implementation issues:
- Stitch lines visible between triangles for semi-transparent sources #4123
- Source resolution #4124


bill-chadwick:
Great to see this PR.

It must represent a large amount of hard work.

src/ol/math.js
```diff
@@ -103,6 +103,69 @@ ol.math.squaredDistance = function(x1, y1, x2, y2) {
 
 
 /**
+ * Solves system of linear equations using Gaussian elimination method.
+ *
+ * @param {Array.<Array.<number>>} A Augmented matrix (n x n + 1 column)
```
marcjansen: Typo: A*n* *a*ugmented

src/ol/ol.js
```diff
@@ -179,6 +193,40 @@ ol.OVERVIEWMAP_MIN_RATIO = 0.1;
...
+/**
+ * @define {number} Maximum allowed size of triangle relative to world width.
+ *     When transforming corners of world extent between certain projections,
+ *     the resulting triangulation seems to have zero error and no subdivision
+ *     is performed.
+ *     If the triangle width is more than this (relative to world width; 0-1),
+ *     subdivison is forced (respecting `ol.RASTER_REPROJ_MAX_SUBDIVISION`).
+ *     Default is `0.25`.
+ */
+ol.RASTER_REPROJ_MAX_TRIANGLE_WIDTH = 0.25;
```
marcjansen: Good to see such good docs here! Thanks!

src/ol/reproj/reproj.js
```diff
@@ -0,0 +1,339 @@
...
+ * @private
+ */
+ol.reproj.browserAntialiasesClip_ = !goog.labs.userAgent.browser.isChrome() ||
+                                    goog.labs.userAgent.platform.isIos();
+
+
+/**
+ * Calculates ideal resolution to use from the source in order to achieve
+ * pixel mapping as close as possible to 1:1 during reprojection.
+ * The resolution is calculated regardless on what resolutions
```
marcjansen: typo: o**f**

src/ol/reproj/reproj.js
```diff
@@ -0,0 +1,339 @@
...
+
+    if (wrapXType == ol.reproj.WrapXRendering_.STITCH_EXTENDED) {
+      stitchContext.drawImage(src.image, wrapXShiftDistance + xPos, yPos,
+                              srcWidth, srcHeight);
+    }
+  });
+
+  var targetTL = ol.extent.getTopLeft(targetExtent);
+
+  goog.array.forEach(triangulation.getTriangles(), function(tri, i, arr) {
```
marcjansen: This is a very long anonymous function, can we possible put this in a helper?

marcjansen:
Awesome work, Petr! Thank you so much! I addes some minor comments, which you may want to address.

Other people can surely review this better than me, but I have the impression that we have never been closer to raster reprojection in OpenLayers since 2009 (you remember, right?). Excited when we will get this in.

bill-chadwick:
I think control of the Overlay Map opacity would add a bit extra to your first demo.

I am seeing the image, or parts of it, disappear when zooming in and panning around on the image reprojection demo using Firefox 40.0.3, operation on Chrome seems better (both Win 7).

src/ol/math.js
petrsloup: This is actually not a typo: `A` is the name of the parameter.

src/ol/reproj/reproj.js
petrsloup: It would be possible, but the function heavily relies on the precalculated values, so a lot of variables would have to be passed.

src/ol/reproj/reproj.js
petrsloup: Fixed, thanks.

src/ol/math.js
marcjansen: Sure, got it.

petrsloup:
@bill-chadwick The **image** reprojection does not sometimes work optimally especially on low-RAM and/or HiDPI devices, where creating large canvas can lead to strange behavior during rendering.
There is, however, no simple way around this at the moment (the image-based sources are always expected to return a single raster -- no tiling is possible).

I addressed the comments (thanks @marcjansen) and rebased on latest master (+ added some commits to follow the latest changes -- removed calls to `goog.math.clamp` and `goog.isDef`).

ahocevar:
Thanks again for this great work. I'm now starting to review.

It would be great if you could rebase the pull request one more time. Please also check the code for any occurrences of the goog.* functions that we stopped using in the meantime: goog.fx.easing, goog.string.remove, goog.string.startsWith, goog.string.trim(), goog.dom.appendChild, goog.isNull, goog.isDefAndNotNull, goog.object.get, goog.functions.constant, goog.object.getKeys, goog.nullFunction, goog.array.*, goog.now, goog.math.clamp, goog.isDef and goog.object.containsKey. If any of those are there, please replace them.

petrsloup:
@ahocevar Rebased and removed calls to all goog.* functions that you mentioned.

ahocevar:
Thanks @petrsloup, reviewing now.

ahocevar:
Sorry for the comments noise so far. I realised that it's better to review the final result instead of individual commits, because some of my concerns are addressed in later commits. Continuing with the final result now...

src/ol/ol.js
```diff
@@ -29,6 +29,13 @@ ol.DEFAULT_MIN_ZOOM = 0;
 
 
 /**
+ * @define {number} Default maximum allowed threshold  (in pixels) for
+ *     reprojection triangulation. Default is `0.5`.
+ */
+ol.DEFAULT_RASTER_REPROJ_ERROR_THRESHOLD = 0.5;
```
ahocevar: Couldn't hurt to be a bit more verbose and call this `ol.DEFAULT_RASTER_REPROJECTION_ERROR_THRESHOLD` .

src/ol/ol.js
```diff
@@ -160,6 +174,40 @@ ol.OVERVIEWMAP_MIN_RATIO = 0.1;
...
+ * @define {number} Maximum number of source tiles for raster reprojection of
+ *     a single tile.
+ *     If too many source tiles are determined to be loaded to create a single
+ *     reprojected tile the browser can become unresponsive or even crash.
+ *     This can happen if the developer defines projections improperly and/or
+ *     with unlimited extents.
+ *     If too many tiles are required, no tiles are loaded and
+ *     `ol.TileState.ERROR` state is set. Default is `100`.
+ */
+ol.RASTER_REPROJ_MAX_SOURCE_TILES = 100;
```
ahocevar: Same as above, and the same for the defines below: `RASTER_REPROJECTION_` instead of `RASTER_REPROJ_`.

src/ol/reproj/tile.js
```diff
@@ -0,0 +1,340 @@
...
+   */
+  this.sourcesListenerKeys_ = null;
+
+  /**
+   * @private
+   * @type {number}
+   */
+  this.srcZ_ = 0;
+
+  var targetExtent = targetTileGrid.getTileCoordExtent(this.getTileCoord());
```
ahocevar: You might be able to remove most (if not all) of the wrapX complexity in `ol.reproj.render` if you use `this.getTileCoordForTileUrlFunction()` here instead of `this.getTileCoord()`.

src/ol/reproj/tile.js
```diff
@@ -0,0 +1,340 @@
...
+  this.srcTiles_.forEach(function(tile, i, arr) {
+    if (tile && tile.getState() == ol.TileState.LOADED) {
+      sources.push({
+        extent: this.sourceTileGrid_.getTileCoordExtent(tile.tileCoord),
+        image: tile.getImage()
+      });
+    }
+  }, this);
+
+  var tileCoord = this.getTileCoord();
```
ahocevar: You might be able to remove most (if not all) of the wrapX complexity in `ol.reproj.render` if you use `this.getTileCoordForTileUrlFunction()` here instead of `this.getTileCoord()`.

src/ol/reproj/triangulation.js
```diff
@@ -0,0 +1,338 @@
...
+      ol.extent.getWidth(this.sourceProj_.getExtent()) : null;
+
+  /**
+   * @type {?number}
+   * @private
+   */
+  this.targetWorldWidth_ = this.targetProj_.getExtent() ?
+      ol.extent.getWidth(this.targetProj_.getExtent()) : null;
+
+  var tlDst = ol.extent.getTopLeft(targetExtent);
```
ahocevar: A general concern: verbose variable name improve readability. When minified, it does not matter how long a variable name is. So I'd prefer variable names like `destinationTopLeft` over `tlDst`.

ahocevar:
Very impressive work. I have added a few remarks (some of which may be obsolete, because I started the review on a per-commit basis).

The only major concern I have is wrapX handling. It adds a lot of complexity, which I think can be avoided when making proper use of tile ranges, and getTileCoordForTileUrlFunction instead of getTileCoord. It may make sense to set up a quick hangout where you walk me through the problem set, so I can better judge if I'm just missing something, or if there is really a huge potential for simplification here.

bill-chadwick:
I can't think of many use cases for raster reprojection that will involve
wrap X. I expect it to mostly be used to warp local, non-global projection
mapping to web mercator or lat/lon and vice versa.

Are there some use cases involving wrap X that the rest of you can think of?

bill-chadwick:
I guess there are some text book demos comparing different global
projections that would need wrap X.

ahocevar:
> Are there some use cases involving wrap X that the rest of you can think of?

There are some. Reproject from EPSG:3857 to EPSG:3338 for example.

src/ol/reproj/triangulation.js
```diff
@@ -0,0 +1,338 @@
...
+
+/**
+ * Calculates extent of the 'source' coordinates from all the triangles.
+ * The left bound of the returned extent can be higher than the right bound,
+ * if the triangulation wraps X in source (see
+ * {@link ol.reproj.Triangulation#getWrapsXInSource}).
+ *
+ * @return {ol.Extent} Calculated extent.
+ */
+ol.reproj.Triangulation.prototype.calculateSourceExtent = function() {
```
ahocevar: I think this is where we have potential for simplification: You could calculate the source tile range instead of the source extent, and use `source.getTileCoordForTileUrlFunction()` to get tile extents.

Or did you choose this approach because you wanted to support reprojecting global ImageStatic sources to local projections that cross the date line?

ahocevar: Even if you keep this method, what would happen if your left bound is always smaller than the right bound as it should be? I think it would save you some special case handling when working with the extent you get returned from this method.

src/ol/reproj/triangulation.js
elemoine: Agree 100%.

petrsloup:
The wrapX handling is now significantly simplified (thanks @ahocevar for the valuable input) -- the proj4js returns "wrapped" coordinates, so these are "unwrapped" and the rest of the reprojection more or less relies on  existing ol3 capability of working with extents and tile coordinates "outside" of the projection and tilegrid ranges (the wrapping is done when creating urls for the tiles).
This significantly simplifies the code (and also improves performance), but breaks wrapping when reprojecting un-tiled image sources -- I think this is not major problem at the moment. The unwrapped extent is passed to `ol.source.Image#getImageInternal`, but the implementations (e.g. `ol.source.ImageStatic#getImageInternal`) are not designed to handle this.

I also addressed the remarks from @ahocevar (most notably renamed defines, private properties and local variables to improve code readability) and again rebased on current master.

src/ol/reproj/triangulation.js
```diff
@@ -0,0 +1,340 @@
...
+  var transformInv = ol.proj.getTransform(this.targetProj_, this.sourceProj_);
+
+  /**
+   * @param {ol.Coordinate} c
+   * @return {ol.Coordinate}
+   * @private
+   */
+  this.transformInv_ = function(c) {
+    var key = c[0] + '/' + c[1];
+    return transformInvCache[key] || (transformInvCache[key] = transformInv(c));
```
tsauerwein: Minor but I am not a fan of doing an assignment in a return statement.

src/ol/reproj/triangulation.js
```diff
@@ -0,0 +1,340 @@
...
+
+  /**
+   * @type {boolean}
+   * @private
+   */
+  this.canWrapXInSource_ = this.sourceProj_.canWrapX() &&
+      !!maxSourceExtent &&
+      !!this.sourceProj_.getExtent() &&
+      (ol.extent.getWidth(maxSourceExtent) ==
+       ol.extent.getWidth(this.sourceProj_.getExtent()));
```
tsauerwein: The width could be a floating point value, right? Is it safe to compare with `==` in this case?

src/ol/reproj/triangulation.js
```diff
@@ -0,0 +1,340 @@
...
+
+  // when the quad is wrapped in the source projection
+  // it covers most of the projection extent, but not fully
+  var wrapsX = this.sourceProj_.canWrapX() &&
+               sourceCoverageX > 0.5 && sourceCoverageX < 1;
+
+  var needsSubdivision = false;
+
+  if (maxSubdivision > 0) {
+    if (this.targetProj_.isGlobal() && this.targetWorldWidth_) {
```
tsauerwein: `this.targetWorldWidth_` could be null (according to the type definition). Do you maybe need an assert?

src/ol/math.js
```diff
@@ -79,6 +79,69 @@ ol.math.squaredDistance = function(x1, y1, x2, y2) {
...
+ *                                   in row-major order.
+ * @return {Array.<number>} The resulting vector.
+ */
+ol.math.solveLinearSystem = function(A) {
```
tsauerwein: In know in maths matrices are written with a capital letter, but usually we start variable names in lower-case (also for matrices, for exampe [here](https://github.com/openlayers/ol3/blob/master/src/ol/vec/mat4.js#L15))

src/ol/reproj/reproj.js
```diff
@@ -0,0 +1,258 @@
...
+ * @param {number} width Width of the canvas.
+ * @param {number} height Height of the canvas.
+ * @param {number} pixelRatio Pixel ratio.
+ * @param {number} sourceResolution Source resolution.
+ * @param {ol.Extent} sourceExtent Extent of the data source.
+ * @param {number} targetResolution Target resolution.
+ * @param {ol.Extent} targetExtent Target extent.
+ * @param {ol.reproj.Triangulation} triangulation Calculated triangulation.
+ * @param {Array.<{extent: ol.Extent,
+ *                 image: (HTMLCanvasElement|Image)}>} sources Array of sources.
```
tsauerwein: The type of [ImageBase.getImage()](https://github.com/openlayers/ol3/blob/master/src/ol/imagebase.js#L94-L98) is `{HTMLCanvasElement|Image|HTMLVideoElement}`. So you could also get a `HTMLVideoElement` here.

src/ol/reproj/reproj.js
```diff
@@ -0,0 +1,258 @@
...
+  var targetTopLeft = ol.extent.getTopLeft(targetExtent);
+
+  triangulation.getTriangles().forEach(function(triangle, i, arr) {
+    context.save();
+
+    /* Calculate affine transform (src -> dst)
+     * Resulting matrix can be used to transform coordinate
+     * from `sourceProjection` to destination pixels.
+     *
+     * To optimize number of context calls and increase numerical stability,
```
tsauerwein: What do you mean with "numerical stability"? To avoid rounding errors?

src/ol/reproj/reproj.js
```diff
@@ -0,0 +1,258 @@
...
+
+    var augmentedMatrix = [
+      [x1, y1, 0, 0, u1 - u0],
+      [x2, y2, 0, 0, u2 - u0],
+      [0, 0, x1, y1, v1 - v0],
+      [0, 0, x2, y2, v2 - v0]
+    ];
+    var affineCoefs = ol.math.solveLinearSystem(augmentedMatrix);
+    if (!affineCoefs) {
+      return;
```
tsauerwein: Would you want to do `context.restore()` before returning?

src/ol/reproj/triangulation.js
elemoine: +1 with @tsauerwein. This is not something we do elsewhere in the code..

src/ol/reproj/tile.js
```diff
@@ -0,0 +1,325 @@
...
+ * @param {number} z Zoom level.
+ * @param {number} x X.
+ * @param {number} y Y.
+ * @param {number} pixelRatio Pixel ratio.
+ * @param {function(number, number, number, number) : ol.Tile} getTileFunction
+ *     Function returning source tiles (z, x, y, pixelRatio).
+ * @param {number=} opt_errorThreshold Acceptable reprojection error (in px).
+ * @param {boolean=} opt_renderEdges Render reprojection edges.
+ */
+ol.reproj.Tile = function(sourceProj, sourceTileGrid,
```
tsauerwein: Would it make sense to have `opt_errorThreshold` and `opt_renderEdges` also for `ol.reproj.Image`?

ahocevar:
I'm done with my review @petrsloup. Now please also answer @tsauerwein's questions and try to address his concerns. @tsauerwein, please also add a note when you're done with your review.

src/ol/ol.js
```diff
@@ -160,6 +174,40 @@ ol.OVERVIEWMAP_MIN_RATIO = 0.1;
...
+ * @define {number} Maximum number of source tiles for raster reprojection of
+ *     a single tile.
+ *     If too many source tiles are determined to be loaded to create a single
+ *     reprojected tile the browser can become unresponsive or even crash.
+ *     This can happen if the developer defines projections improperly and/or
+ *     with unlimited extents.
+ *     If too many tiles are required, no tiles are loaded and
+ *     `ol.TileState.ERROR` state is set. Default is `100`.
+ */
+ol.RASTER_REPROJECTION_MAX_SOURCE_TILES = 100;
```
tsauerwein: Do you have an example where you would need that many source tiles to reproject a single tile?

src/ol/ol.js
```diff
@@ -160,6 +174,40 @@ ol.OVERVIEWMAP_MIN_RATIO = 0.1;
...
+
+
+/**
+ * @define {number} Maximum number of subdivision steps during raster
+ *     reprojection triangulation. Prevents high memory usage and large
+ *     number of proj4 calls (for certain transformations and areas).
+ *     At most `2*(2^this)` triangles are created. Default is `10`.
```
tsauerwein: "At most `2*(2^this)` triangles are created ..." ... for each tile/image?

src/ol/reproj/image.js
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * @constructor
+ * @extends {ol.ImageBase}
+ * @param {ol.proj.Projection} sourceProj Source projection (of the data).
+ * @param {ol.proj.Projection} targetProj Target projection.
+ * @param {ol.Extent} targetExtent Target extent.
+ * @param {number} targetResolution Target resolution.
+ * @param {number} pixelRatio Pixel ratio.
+ * @param {function(ol.Extent, number, number):ol.ImageBase} getImageFunction
+ *     Function returning source images (extent, resolution, pixelRatio).
```
tsauerwein: You could also create a `@typedef` to document the signature of the function, for example like here:
https://github.com/openlayers/ol3/blob/master/src/ol/centerconstraint.js#L7-L10

src/ol/renderer/layerrenderer.js
```diff
@@ -191,7 +195,8 @@ ol.renderer.Layer.prototype.scheduleExpireCache =
              */
             function(tileSource, map, frameState) {
               var tileSourceKey = goog.getUid(tileSource).toString();
-              tileSource.expireCache(frameState.usedTiles[tileSourceKey]);
+              tileSource.expireCache(frameState.viewState.projection,
+                                     frameState.usedTiles[tileSourceKey]);
```
tsauerwein: Here you are expiring the reprojected tiles. Would it make sense also to expire the source tiles?

tsauerwein:
I am done with my review. I only added a couple of questions and minor suggestions. Thanks for this impressive work!

src/ol/reproj/triangulation.js
petrsloup: I updated the code to avoid the assignment in the returns statement.

src/ol/reproj/triangulation.js
petrsloup: Yes, but I think it should be safe to use `==` since there is no calculation involved (the extent is either identical or someone explicitly defines subset of the projection extent).

src/ol/reproj/triangulation.js
petrsloup: `this.targetWorldWidth_` is only used if truthy (this `if` statement ensures it), I don't think there's any need for an assert

src/ol/math.js
petrsloup: I changed the parameter name to `mat`.

src/ol/reproj/reproj.js
petrsloup: Added, thanks.

src/ol/reproj/reproj.js
petrsloup: Yes, without the shift, the coordinates of the triangle vertices would be almost identical.
This would decrease the numerical stability of the affine transformation (in the canvas itself) and would result in rendering artifacts on higher zoom levels.

src/ol/reproj/reproj.js
petrsloup: Thanks. I instead moved the `context.save()` after this check.

src/ol/ol.js
petrsloup: Yes, for each triangulated extent (tile/image) -- updated the docs to be more clear.

src/ol/renderer/layerrenderer.js
petrsloup: The `tileSource.expireCache` call actually expires tiles in all the projections (including the source tiles) if the respective TileCache capacity is over high water mark.
The `projection` parameter only specifies the projection of the `usedTiles` (and these are not removed from the cache even if it's over the high water mark).

ahocevar: > The projection parameter only specifies the projection of the usedTiles (and these are not removed from the cache even if it's over the high water mark).

Actually what your code does is: prune the cache for all projections that are not used (by passing `{}` as `usedTiles`), and expire the cache for the used projection (i.e. remove the least recently used tiles from the cache if it is over high water mark).

src/ol/ol.js
petrsloup: The resolution of the source data is determined in the center of the reprojected tile and is constant for the whole tile.
If the projection behaves significantly different in the center and at the edges of the tile, the number of required tiles can be very high.

Another case (probably easier to achieve):
This can also happen when the tile grid has more than 100 tiles on the least-detailed zoom level and the reprojection needs to render the whole extent. (It would have to download and process all the tiles, potentially crashing the browser.)

ahocevar:
Some metrics on build size:

* Current master:<br>
  uncompressed: 479853 bytes, compressed: 142132 bytes
* This branch without `ol.ENABLE_RASTER_REPROJECTION`:<br>
  uncompressed: 482146 bytes, compressed: 142573 bytes
* This branch with `ol.ENABLE_RASTER_REPROJECTION`:<br>
  uncompressed: 489926 bytes, compressed: 145345 bytes

src/ol/reproj/tile.js
petrsloup: Adding it only to `ol.reproj.Image` would not be very useful -- it would have to be added also to `ol.source.Image` and passed from all the subclasses, but I don't think it's worth the extra code:
`opt_renderEdges` is used for debugging only and it's probably enough to have this in tiled sources.
I also think that `opt_errorThreshold` is kind of "advanced" parameter for fine-tuning and nobody needs it when displaying image sources.
But if you think otherwise, I don't have any problem adding these (or just the `opt_errorThreshold`).

src/ol/renderer/layerrenderer.js
petrsloup: Are you sure about this? Looking at the code (https://github.com/openlayers/ol3/blob/master/src/ol/tilecache.js#L42) I don't think `ol.TileCache#expireCache` removes tiles only if the cache is over the high water mark (no matter what is passed as `usedTiles`). Am I missing something here?

petrsloup:
Most of the build size without `ol.ENABLE_RASTER_REPROJECTION` is caused by `reprojectionErrorThreshold` constructor options and `setRenderReprojectionEdges` methods (exported to the api) on the sources.

src/ol/reproj/tile.js
tsauerwein: No, because it is only for debugging I think it is not necessary.

tsauerwein:
Thanks for the clarifications, @petrsloup!

src/ol/renderer/layerrenderer.js
ahocevar: You are right, but @petrsloup checks `canExpireCache()` first, and that only returns true when the cache count is over high water mark.

marcjansen:
What a nice discussion here. Great collaboration, guys!

Can we ship it ?!? :shipit:

ahocevar:
> Can we ship it ?!?

Yes we can :-)

marcjansen:
Nice indeed! Thanks to everyone involved.

petrsloup:
Thanks for the comments and remarks, everyone! Great to see this merged!

bartvde:
w00t - great to see this in! Thanks everyone.

klokan:
Cool. Thanks!

ximex:
Thanks! This is a huge improvement!

PR #4126, opened 2015-09-17
Add CRS member to GeoJSON serialization

alvinlindstam:
Fixes https://github.com/openlayers/ol3/issues/2078

This PR changed `ol.format.GeoJSON#writeFeature` and `ol.format.GeoJSON#writeFeatures` to always include the CRS member. No API changes, but the output of API methods will change.

Things to consider:
- I set `crs: null` whenever the projection is unknown, such as when calling the write-methods without specifying a `featureProjection`. This is good for applications sending non-default projections without specifying it, since their GeoJSON will become valid. For applications sending EPSG:4326 data, some semantics will be lost unless they specify `featureProjection`.
- I always set the crs member. We could consider skipping it when it's `EPSG:4326` (default) or adding an option to include it or not.
- The output always uses the code of the output projection. It does not consider the _OGC CRS URNs such as "urn:ogc:def:crs:OGC:1.3:CRS84" shall be preferred over legacy identifiers such as "EPSG:4326"_ part of the spec.

### DIFF
--- a/test/spec/ol/format/geojsonformat.test.js
+++ b/test/spec/ol/format/geojsonformat.test.js
@@ -524,6 +524,92 @@ describe('ol.format.GeoJSON', function() {
           expect(geojson.features[0].properties.mygeom).to.eql(undefined);
         });
 
+    describe('writing CRS member', function() {
+
+      it('writes null CRS when unknown', function() {
+        var str = JSON.stringify(data),
+            array = format.readFeatures(str);
+        var geojson = format.writeFeatures(array);
+        var result = JSON.parse(geojson);
+        expect(result.crs).to.be(null);
+        for (var i = 0, ii = array.length; i < ii; ++i) {
+          expect(result.features[i].crs).to.be(undefined);
+        }
+      });
+
+      it('writes the correct CRS when using featureProjection', function() {
+        var str = JSON.stringify(data),
+            array = format.readFeatures(str);
+        var geojson = format.writeFeatures(array, {
+          featureProjection: 'EPSG:3857'
+        });
+        var result = JSON.parse(geojson);
+        expect(result.crs).to.eql({
+          type: 'name',
+          properties: {
+            name: 'EPSG:4326'
+          }
+        });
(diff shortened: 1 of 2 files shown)


elemoine:
LGTM. @tschaub, you may want to have a look before we merge this.

test/spec/ol/format/geojsonformat.test.js
```diff
@@ -524,6 +524,92 @@ describe('ol.format.GeoJSON', function() {
...
         });
 
+    describe('writing CRS member', function() {
+
+      it('writes null CRS when unknown', function() {
+        var str = JSON.stringify(data),
+            array = format.readFeatures(str);
+        var geojson = format.writeFeatures(array);
+        var result = JSON.parse(geojson);
+        expect(result.crs).to.be(null);
```
tschaub: I don't think this is what we want.  Instead, I think the `crs` member should be absent by default, implying the default CRS for GeoJSON.  The spec does not describe how `crs: null` should be interpreted, and some parsers may only assume the default CRS if the `crs` property is not present.

elemoine: I thought about the same, but was not totally sure.

tschaub:
This looks like the correct change, but we should at least add something to the release notes about it.  

The impact of this change is that we break applications that are using geographic data and strict parsers (unless they are supplying `featureProjection`) in favor of producing valid GeoJSON for applications where interoperability was presumably not a concern (since they were working with invalid GeoJSON previously).

I think it would be nice if we could omit the `crs` member when writing data in geographic/WGS84 - as I imagine there are parsers that fumble with the `crs` member when present.  That said, it is still probably the right change to make.

roemhildtg:
+1

bartvde:
@alvinlindstam any chance you have time to rebase this and incorporate the review comments? TIA

tschaub:
The [official GeoJSON spec](https://tools.ietf.org/html/rfc7946) doesn't describe a `crs` member.  All coordinates should be [geographic / WGS84 datum](https://tools.ietf.org/html/rfc7946#section-4).

Issue #4128, opened 2015-09-18
Remove dependency on the Closure Library


tschaub:
By moving away from the Closure Library, we free ourselves to write OpenLayers in a way that makes it more consumable for others.  Ideally, we could write ES6 modules.  We'll need to confirm that things will still work with the Compiler (if we need to transpile first, annotations will need to be preserved).

bjornharrtell:
Interesting!

FYI I'm trying to move to ES6 modules in JSTS, from the Java source via AST transformation . Still just experimental, but one issue (https://github.com/bjornharrtell/java2estree/issues/13) I've seem to hit is that ES6 modules does not seem to support cyclic dependencies in class hierarchies i.e a subclass cannot inherit from a superclass that imports the subclass directly or indirectly.

This has to do with the module spec and more information about it can be read here:
https://github.com/babel/babel/issues/2061
http://www.2ality.com/2014/09/es6-modules-final.html

Because of this I'm currently considering using CommonJS module format instead of ES6 modules. I have not investigated if/where but I'm guessing there might be some of these cyclic dependencies in OL3.

acanimal:
+1. 
Webpack can help with different module loading systems.

gberaudo:
To give a bit of context, we discussed limitations of the build procedure
last week at the FOSS4G Seoul.

I wanted to allow users to easily compile several projects depending on OL3
together, without requiring the user to use
the closure compiler in it own application. For example, to be able to
compile Ngeo, OL3-Cesium and OL3 in one step and use the result as a
dependency.

The path to reach this goal was believed to be in standardizing the build
process, notably by shrinking or eliminating our dependency on the goog
library (not the compiler).
In the long run, it seemed that most of our needs can be handled by ES6
(standard inheritance, modules, ...) and transpilation from ES6 to ES5
(closure compiler).

As a side effect of this standardization, it was believed that improving
the build experience with closure would benefit all users.

sweco-sebhar:
So is the project accepting PRs for any goog dependency removal from now on? Did some random checks and it seems goog.json (1 require occurence) and goog.string (12 require occurences) might be good candidates to work on. But perhaps it will be useful to create a script of some sorts that can make a top list of goog.* usage over the whole code base?

[...]

PR #4129, opened 2015-09-18
Remove use of goog.isDef

tschaub:
As part of a larger effort to remove our dependency on the Closure Library (see #4128), this branch removes our use of `goog.isDef()`.

I pushed this branch to the main OpenLayers repo so more people can contribute.  I'll happily cherry pick commits from other forks as well.

### DIFF
--- a/src/ol/animation.js
+++ b/src/ol/animation.js
@@ -17,8 +17,7 @@ goog.require('ol.easing');
 ol.animation.bounce = function(options) {
   var resolution = options.resolution;
   var start = options.start ? options.start : goog.now();
-  var duration = ol.isDef(options.duration) ?
-      /** @type {number} */ (options.duration) : 1000;
+  var duration = options.duration !== undefined ? options.duration : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.upAndDown;
   return (
@@ -56,8 +55,7 @@ ol.animation.pan = function(options) {
   var start = options.start ? options.start : goog.now();
   var sourceX = source[0];
   var sourceY = source[1];
-  var duration = ol.isDef(options.duration) ?
-      /** @type {number} */ (options.duration) : 1000;
+  var duration = options.duration !== undefined ? options.duration : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.inAndOut;
   return (
@@ -95,8 +93,7 @@ ol.animation.pan = function(options) {
 ol.animation.rotate = function(options) {
   var sourceRotation = options.rotation ? options.rotation : 0;
   var start = options.start ? options.start : goog.now();
-  var duration = ol.isDef(options.duration) ?
-      /** @type {number} */ (options.duration) : 1000;
+  var duration = options.duration !== undefined ? options.duration : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.inAndOut;
   var anchor = options.anchor ?
@@ -142,8 +139,7 @@ ol.animation.rotate = function(options) {
 ol.animation.zoom = function(options) {
   var sourceResolution = options.resolution;
   var start = options.start ? options.start : goog.now();
-  var duration = ol.isDef(options.duration) ?
-      /** @type {number} */ (options.duration) : 1000;
+  var duration = options.duration !== undefined ? options.duration : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.inAndOut;
   return (
--- a/src/ol/animation.js
+++ b/src/ol/animation.js
@@ -17,7 +17,8 @@ goog.require('ol.easing');
 ol.animation.bounce = function(options) {
   var resolution = options.resolution;
   var start = options.start ? options.start : goog.now();
-  var duration = ol.isDef(options.duration) ? options.duration : 1000;
+  var duration = ol.isDef(options.duration) ?
+      /** @type {number} */ (options.duration) : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.upAndDown;
   return (
@@ -55,7 +56,8 @@ ol.animation.pan = function(options) {
   var start = options.start ? options.start : goog.now();
   var sourceX = source[0];
   var sourceY = source[1];
-  var duration = ol.isDef(options.duration) ? options.duration : 1000;
+  var duration = ol.isDef(options.duration) ?
+      /** @type {number} */ (options.duration) : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.inAndOut;
   return (
@@ -93,7 +95,8 @@ ol.animation.pan = function(options) {
 ol.animation.rotate = function(options) {
   var sourceRotation = options.rotation ? options.rotation : 0;
   var start = options.start ? options.start : goog.now();
-  var duration = ol.isDef(options.duration) ? options.duration : 1000;
+  var duration = ol.isDef(options.duration) ?
+      /** @type {number} */ (options.duration) : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.inAndOut;
   var anchor = options.anchor ?
@@ -139,7 +142,8 @@ ol.animation.rotate = function(options) {
 ol.animation.zoom = function(options) {
   var sourceResolution = options.resolution;
   var start = options.start ? options.start : goog.now();
-  var duration = ol.isDef(options.duration) ? options.duration : 1000;
+  var duration = ol.isDef(options.duration) ?
+      /** @type {number} */ (options.duration) : 1000;
   var easing = options.easing ?
       options.easing : ol.easing.inAndOut;
   return (
--- a/src/ol/animation.js
+++ b/src/ol/animation.js
@@ -1,5 +1,6 @@
 goog.provide('ol.animation');
 
+goog.require('ol');
 goog.require('ol.PreRenderFunction');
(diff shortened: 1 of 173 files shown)


tschaub:
Removing `goog.isDef` from the rotate control (in 35ea29d99fe1df5486f924fade654c153f57a6e1) revealed a curious compiler issue.  Before this change, the compiler doesn't complain when we pass a `Node` to `goog.dom.classlist.add` (which expects `Element`).  After removing `goog.isDef`, the compiler does complain.  I can't imagine why this is, but I've changed the type of the label to be `string|Element`.

marcjansen:
I just push one other commit, which does not contain the byte savings in the message... You may want to rewrite this or do we don't care?

tschaub:
> I just push one other commit, which does not contain the byte savings in the message... You may want to rewrite this or do we don't care?

Thanks @marcjansen.  The size diff was tedious to calculate anyway.  I'll run a comparison at the end.

tschaub:
Another "bug" concealed by our use of `goog.isDef()` addressed in 241ffd5cd3d95eb6b7380097f7b46a1bdbb86003.  The `olx.format.WFSWriteGetFeatureOptions` object doesn't have `count` or `startIndex` properties defined in `olx.js`.  So I assume code like this gets removed during compilation:
```js
if (goog.isDef(options.count)) {
  node.setAttribute('count', options.count);
}
```

If this is right, it means the `count` and `startIndex` options actually don't work with compiled code.

When I remove the `goog.isDef` checks for these properties, the compiler complains about our use of them.  So I added them to `olx.js`.  But they are still lacking documentation there (cc @bartvde).

bartvde:
@tschaub I just pushed a commit that adds docs

marcjansen:
Hey everybody, when exactly do we want to use `ol.isDef`? Always when we are checking booleans? I'll add a commit for this.

marcjansen:
If I understand it correctly, it wouldn't hurt if we always used `ol.isDef`, right? Closure Compiler may probably inline the actual tests against `undefined`, right? How shall I proceed?

tschaub:
> If I understand it correctly, it wouldn't hurt if we always used ol.isDef, right? Closure Compiler may probably inline the actual tests against undefined, right? How shall I proceed?

I think `if (value) {}` and `if (key in object) {}` should be preferred.  Using `ol.isDef(value)` makes sense when you also want to handle `null`, `false`, `0`, or `''`.

The compiler derives extra knowledge from `if (value) {}` checks.  We have to add type assertions/casts when we use `ol.isDef(value)`.

src/ol/animation.js
```diff
@@ -15,9 +15,9 @@ goog.require('ol.easing');
  */
 ol.animation.bounce = function(options) {
   var resolution = options.resolution;
-  var start = goog.isDef(options.start) ? options.start : goog.now();
-  var duration = goog.isDef(options.duration) ? options.duration : 1000;
-  var easing = goog.isDef(options.easing) ?
+  var start = options.start ? options.start : goog.now();
+  var duration = options.duration ? options.duration : 1000;
```
fredj: `ol.isDef` should be used, options.duration may be equal to 0.

marcjansen: Addressed with bc28553, thanks.

marcjansen:
This branch currently couldn't be merged, as there are conflicts. Shall I resolve these with a manual merge? Or do we do it at the end?

marcjansen:
image … enough for one day.

elemoine:
I started reviewing this, but I am not clear on using `ol.isDef` versus `=== undefined`.

marcjansen:
I used `!== undefined` in assertions, because they get removed at compile time, right? I didn't want to add a `goog.require('ol')` if only an asserion needs `ol.isDef`.

In code I used `ol.isDef` where needed, but I may have missed some.

Is there any difference in using `'key' in object` vs `obj.key` (for standard objects)? The latter seems way more natural and readable to me.

I can adjust this any way we want, as long as we have consensus.

Thanks for the partial review, though, @elemoine

tschaub:
> Is there any difference in using 'key' in object vs obj.key (for standard objects)?
```js
var foo = {bar: false};
'bar' in foo; // true
foo.bar; // false
```

tschaub:
Thanks for chugging through this @marcjansen.

We could make things easier on ourselves and replace `goog.isDef` with `ol.isDef`.  This would still require assertions/casts, but it would be less error prone (and less tedious) than choosing the most suitable alternative form.  We can always replace `ol.isDef` with `in` or a truthy check later.

What do others think?

marcjansen:
Yes, that is a difference, but since the value is a boolean, I would have used `ol.isDef` in the first place.

marcjansen:
Do we want to do this for the remaining files or starting new from master with this mass-search-and-replace?

I've already spend some time on this, you know. I go with the majority vote, of course.

elemoine:
> I used !== undefined in assertions, because they get removed at compile time, right?

Do you mean that the assertion is **not** removed at compile time if you use `ol.isDef` rather than `=== undefined`?? I hope it's not the case.



> I didn't want to add a goog.require('ol') if only an asserion needs ol.isDef.

I don't think it's a problem. But I do think using `ol.isDef` here and `=== undefined` there is inconsistent and confusing.

elemoine:
> Is there any difference in using 'key' in object vs obj.key (for standard objects)? The latter seems way more natural and readable to me.

If what you want is just test the existence of a key in an object literal and ignore the value associated with that key then I think `key in obj` is the construct you want to use. Tim's [comment](https://github.com/openlayers/ol3/pull/4129#issuecomment-142814509) says the same.

elemoine:
> We could make things easier on ourselves and replace goog.isDef with ol.isDef. This would still require assertions/casts,

So I get from this sentence that

```js
goog.asserts.assert(goog.isDef(obj));
```

is not the same as

```js
goog.asserts.assert(ol.isDef(obj));
```

for the compiler. And that's the reason Marc uses:

```js
goog.asserts.assert(obj !== undefined);
```

When using `ol.isDef` in the assertion the assertion is removed at compile time (as when `goog.isDef` or `!== undefined` is here), but we may get a compile error later because for the compiler `undefined` is still a possible value. Am I correct? If this is the correct then I tend to think that we should remove `ol.isDef` and use `!== undefined` everywhere.

> but it would be less error prone (and less tedious) than choosing the most suitable alternative form. We can always replace ol.isDef with in or a truthy check later.
> 
> What do others think?

I think I'd do it the hard way, mainly because we started this way.

marcjansen:
I think this is an oversight.

I agree that my usage of isDef/undefined is confusing.

I think the compiler will remove both in assertions.

marcjansen:
> I think I'd do it the hard way, mainly because we started this way.

Fine for me.

marcjansen:
> When using ol.isDef in the assertion the assertion is removed at compile time (as when goog.isDef or !== undefined is here), but we may get a compile error later because for the compiler undefined is still a possible value. Am I correct?

It looks like exactly this is the case. And this is quite annoying, as it leads to some type casts that can be avoided when using `goog.isDef()`.

> If this is the correct then I tend to think that we should remove ol.isDef and use !== undefined everywhere.

Did we come to an agreement here, @elemoine and @tschaub?

elemoine:
> Did we come to an agreement here, @elemoine and @tschaub?

Not yet. Waiting for @tschaub to respond.

ahocevar:
I think for now you can leave `goog.isDef()` inside assertions untouched, because in a later step of the Closure library removal process we'll replace such assertions with type casts anyway.

elemoine:
> because in a later step of the Closure library removal process we'll replace such assertions with type casts anyway.

If we don't use Closure Library I think we should introduce and use `ol.assert`. Being able to add assertions is a very good thing in my opinion, and the compiler makes that possible.

elemoine:
So if our goal is to remove `goog.isDef` I suggest that we use `=== undefined` and `!== undefined` everywhere we currently use `goog.isDef` and `!goog.isDef`.

tschaub:
> So if our goal is to remove goog.isDef I suggest that we use === undefined and !== undefined everywhere we currently use goog.isDef and !goog.isDef.

Works for me.

tschaub:
Assuming that nobody is currently working on this.  I'm going to rebase to resolve conflicts.  If anybody has any work based on 426c56bd7c616918c04bd5fc7fcc269e49c36a21, we can cherry pick.

marcjansen:
Yes please rebase. I think it's only me and you, like in all thise songs :-)

tschaub:
Ok, now working on a commit that removes `ol.isDef()`.

tschaub:
This is now ready for review.  There are a lot of places where we could replace `if (foo !== undefined) {}` with `if (foo) {}`.  I'd be in favor of making those changes separately.

ahocevar:
Looks great. Not sure if ceb39b0d07acf534772eda00162016093a51fff1 was meant to be part of this pull request, but I think this is good to merge regardless. Great work gentlemen.

tschaub:

> Not sure if ceb39b0 was meant to be part of this pull request

Yeah, this is intentionally here.  See https://github.com/openlayers/ol3/pull/4129#issuecomment-141837686.  Removing the use of `goog.isDef()` revealed missing properties in `olx.js`.

marcjansen:
Nice to see this in so fast.

PR #4134, opened 2015-09-18
Patch for interactive inline SVG icons in IE11

### LINKED ISSUE #4090: SVG sprites don't work in IE11
OL 3.8.2
The sprite (inline SVG image) generates exception in File: ol-debug.js, Line: 59315, Column: 11 when called from IE11, no problem sin Chrome.

Details: 
The referenced line is "context.drawImage(image, originX, originY, width, height,..."
Looks like either sizes or origin aren't defined despite of the following lines in the code:

function getSVGIcon() {
   var DOMURL = window.URL || window.webkitURL || window;
   var img = new Image(20, 20);
   var svg = new Blob([ICON_RECT90()], { type: 'image/svg+xml;charset=utf-8' });
   img.src = DOMURL.createObjectURL(svg);
   return img;
}
...

var iconStyle = new ol.style.Style({
                    image: new ol.style.Icon({
                        anchor: [10, 10],
                        origin: [0, 0],
                        anchorXUnits: 'pixels',
                        anchorYUnits: 'pixels',
                        rotateWithView: true,
                        img: getSVGIcon(),
                        imgSize: [20, 20],
                    })
                });
iconFeature.setStyle(iconStyle);
MAP.getLayers().item(0).getSource().addFeatures([iconFeature]);

spatialillusions:
This will resolves #4090 the problem with inline SVG icons in IE11. (The image size must however be set to two integer values and can't handle fractional values, but that is in a compleatly other place in the code.)

Since inline SVG never is loaded determineTainting_() was never called for them, and there is what I have found no way to set that an image is tainting the canvas when you define your image. This code simply makes a tainting check for all images that are loaded when initialized so they get the right tainting value.

The other part of my fix is to move a line of code into the try statement since it otherwise will cause an error in IE11.


spatialillusions:
It is today possible to draw inline SVG icons to the canvas, I have a working demo up here:
http://spatialillusions.com/milsymbol/examples/openlayers3/imageicons.html

If you try to add drag interactions to them you will however get an security error in IE11, since IE11 can't tell de difference between SVG with forign objects and without and because of this will mark the canvas as tainted.

With this small patch you will get your inline SVG working with interactions as well and this would be useful in several usecases I have.

PR #4140, opened 2015-09-19
Add flight animation example

tsauerwein:
One more example demonstrating the use of `postcompose`.

flight-animation
http://tsauerwein.github.io/ol3/animation-flights/examples/flight-animation.html


elemoine:
The build fails:

```
info ol Parsing dependencies

info ol Compiling 358 sources

ERR! compile /home/travis/build/openlayers/ol3/build/compiled-examples/all.js:2207: ERROR - 'arc.GreatCircle' used but not goog.require'd

ERR! compile         var arcGenerator = new arc.GreatCircle(

ERR! compile                            ^

ERR! compile 

ERR! compile 

ERR! compile 1 error(s), 0 warning(s)

ERR! compile 

make: *** [build/compiled-examples/all.combined.js] Error 1
```

which makes sense as we don't have externs for the `arc` library.

Two options: (a) create an externs file for the `arc` library, (b) add `// NOCOMPILE` as the first line of the flight-animation.js file.

+1 for option (a).

tschaub:
@tsauerwein can you update this with `// NOCOMPILE`?  Having externs files for things only used in the examples seems like overkill to me.  But I know we already do this.

tsauerwein:
Thanks for the reminder, I rebased and added `NOCOMPILE`.

tsauerwein:
I will update the example once #4401 is merged.

tsauerwein:
PR updated, a review would be appreciated! :)

fredj:
LGTM

examples/flight-animation.js
```diff
@@ -0,0 +1,135 @@
...
+
+      var maxIndex = Math.min(elapsedPoints, coords.length);
+      var currentLine = new ol.geom.LineString(coords.slice(0, maxIndex));
+
+      // directly draw the line with the vector context
+      vectorContext.drawLineStringGeometry(currentLine, feature);
+    }
+  }
+  // tell OL3 to continue the postcompose animation
+  frameState.animate = true;
```
tschaub: Instead of having people mutate the frame state, I think we should encourage the use of `map.render()` (this is part of the documented API).

examples/flight-animation.js
```diff
@@ -0,0 +1,135 @@
...
+
+var flightsSource = new ol.source.Vector({
+  wrapX: false,
+  attributions: [new ol.Attribution({
+    html: 'Flight data by ' +
+        '<a href="http://openflights.org/data.html">OpenFlights</a>,'
+  })],
+  loader: function(extent, resolution, projection) {
+    var url = 'data/openflights/flights.json';
+    $.ajax({url: url, dataType: 'json', success: function(response) {
```
tschaub: After #4420 is complete, examples that use jQuery will need to have it in the `resources` list in the YAML front-matter.

I'd like to limit the use of jQuery to examples that demonstrate integration with a jQuery plugin (e.g. the tooltip/popover examples).  Instead of `$.ajax()`, I'm converting examples to use [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) (with a polyfill for IE/Safari).

No need to change this here yet, but just a heads up on what future examples will use.

tsauerwein:
Thanks for the feedback, merging.

Issue #2940, opened 2014-11-12
Select interaction ignores layer filter for feature overlays


ssaarela:
Map.forEachFeatureAtPixel is called for all features of a feature overlay regardless of layer filter. I have a couple of layers containing different kind of editable information and use Draw (polygons) for one and Select/Modify (points) for another. I try to target Select to the layer containing the relevant points with layers option, but as Draw uses feature overlay, Select grabs those polygons immediately. 

Would it be possible to use layerFilter also in the function(feature, layer) -argument of the map.forEachFeatureAtPixel? One can always supply a filter that accepts null-layer, but without it I see no way to skip feature overlays.

pgiraud:
I've written an example that may illustrate the issue.
http://jsfiddle.net/pgiraud/axmbruf5/

In this example, we have two layers that are "unmanaged". The select interaction is configured with a `layers` option which should limit the selection to only one layers. However, one is able to select any of the 2 points.

pgiraud:
In the tests (https://github.com/openlayers/ol3/blob/master/test/spec/ol/renderer/canvas/canvasmaprenderer.test.js#L49), it's clearly written that forEachFeatureAtPixel will always return features from the unmanaged layers.

1 - It should be documented in the API doc.
2 - Can someone explain this choice?

elemoine:
https://github.com/openlayers/ol3/pull/3883 and https://github.com/openlayers/ol3/issues/3878 are relevant.

And `ol.layer.Layer#setMap` includes docs about this already. See https://github.com/openlayers/ol3/blob/master/src/ol/layer/layer.js#L150-L161.

pgiraud:
OK. Thanks.

elemoine:
@pgiraud So I guess the problem is related to the Select interaction itself using an unmanaged layer for drawing selected features. If that unmanaged layer is filtered out then the selected features won't be returned by `forEachFeatureAtPixel`, making it impossible to unselect them. But there may be a better way to handle this. Instead of ignoring the layer filter for unmanaged layers we could make the Select interaction combine the user-provided layer filter with its own layer filter, in an OR operation.

elemoine:
See https://github.com/openlayers/ol3/pull/4143.

Issue #4114, opened 2015-09-15
Draw interaction does not work as expect when there is a select interaction


bartvde:
```
diff --git a/examples/draw-features.js b/examples/draw-features.js
index b7cb4ce..5f13bae 100644
--- a/examples/draw-features.js
+++ b/examples/draw-features.js
@@ -2,6 +2,7 @@ goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.geom.Polygon');
 goog.require('ol.interaction.Draw');
+goog.require('ol.interaction.Select');
 goog.require('ol.layer.Tile');
 goog.require('ol.layer.Vector');
 goog.require('ol.source.MapQuest');
@@ -46,6 +47,8 @@ var map = new ol.Map({
   })
 });
 
+map.addInteraction(new ol.interaction.Select());
+
 var typeSelect = document.getElementById('type');
 
 var draw; // global so we can remove it later
```

![screen shot 2015-09-15 at 19 26 23](https://cloud.githubusercontent.com/assets/319678/9884377/ab70a2ba-5bdf-11e5-8647-9e7be8a69ece.png)

Is there an obvious reason why this combination of interactions should be avoided?

bartvde:
Also, this does not happen if the layer is unmanaged. If the layer is managed, you can see the above effect.

bartvde:
This line is causing the observed behaviour: https://github.com/openlayers/ol3/blob/master/src/ol/interaction/selectinteraction.js#L257

bartvde:
In the end I made sure in my application I did not have these two interactions active at the same time. Workaround though.

ahocevar:
Can you reference the faulty logic me in selectinteraction.js for the revision in question instead of master? The line on master is a JSDoc comment.

PR #4143, opened 2015-09-21
Do not ignore layer filter for unmanaged layers

### LINKED ISSUE #3878: forEachFeatureAtPixel returns wrong layer if feature already selected

elemoine:
PR #3883 made `forEachFeatureAtPixel` ignore unmanaged layers. This commit reverts that change, and modifies the Select interaction to make sure that the unmanaged layer it uses internally is not filtered out when the user provides a layer filter.

So this PR fixes #3878 in a different way, and it addresses the limitation reported in #2940, where unmanaged layers cannot be filtered out.

@ahocevar, @probins, do you anticipate problems with that approach?

### DIFF
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -148,31 +148,6 @@ ol.interaction.Select = function(opt_options) {
   this.filter_ = goog.isDef(options.filter) ? options.filter :
       goog.functions.TRUE;
 
-  var layerFilter;
-  if (goog.isDef(options.layers)) {
-    if (goog.isFunction(options.layers)) {
-      layerFilter = options.layers;
-    } else {
-      var layers = options.layers;
-      layerFilter =
-          /**
-           * @param {ol.layer.Layer} layer Layer.
-           * @return {boolean} Include.
-           */
-          function(layer) {
-        return goog.array.contains(layers, layer);
-      };
-    }
-  } else {
-    layerFilter = goog.functions.TRUE;
-  }
-
-  /**
-   * @private
-   * @type {function(ol.layer.Layer): boolean}
-   */
-  this.layerFilter_ = layerFilter;
-
   /**
    * @private
    * @type {ol.layer.Vector}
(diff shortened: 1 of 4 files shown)


ahocevar:
Not sure actually. Since we create unmanaged layers for the Draw and Modify interactions as well, these two controls might also need modifications. Since this is a change in behavior, it should also be added to the release notes.

elemoine:
Thanks for your comments @ahocevar.

> Since we create unmanaged layers for the Draw and Modify interactions as well, these two controls might also need modifications.

Not sure.

With the current code, the layer filter is ignored for the Draw and Modify interactions' unmanaged layers. This means that features in these layers will always be detected by the Select interaction.

With my patch, the layer filter will no longer be ignored for the Draw and Modify interactions' unmanaged layers. It means that, if the Select interaction is configured with a layer filter, the Select interaction will not detect features drawn by the Draw and Modify interactions, which I think is fine.

My goal with this patch is to provide more flexibility, where the user can filter out unmanaged layers if that's what he wishes.




> Since this is a change in behavior, it should also be added to the release notes.

Sure. I can add that if you think my patch makes sense.

elemoine:
@ahocevar more opinions about this? As I said I'm happy to add release notes if you agree with the change.

ahocevar:
I'll leave the final judgement to someone else. I really don't understand the implications of this change. All I know is that the old feature overlays could not be filtered out as well, so I tend to lean towards thinking that this proposed change here is not a good one.

elemoine:
I am just looking at providing more flexibility and making unmanaged layers work similarly to managed layers. People use unmanaged layers because they're always on top, but they don't expect their layer filters to be ignored for unmanaged layers.

ahocevar:
I think we should discourage the use of unmanaged layers just for the sake of being on top. Especially since we can now set a z-index on a layer.

elemoine:
> I think we should discourage the use of unmanaged layers just for the sake of being on top. Especially since we can now set a z-index on a layer.

Also a good point.

probins:
> I think we should discourage the use of unmanaged layers just for the sake of being on top

I'd agree with that too. The use cases for unmanaged layers in apps (as opposed to internally in interactions for temporary feature display) have never been all that clear to me. The current examples (adapted from the old FeatureOverlay) were written before the select interaction existed. I think most of them could be changed (and simplified) to use a select interaction for visual highlighting of a particular feature instead.

Like Andreas, I'm not sure what implications this change would have.

bartvde:
I'm seeing something weird and I believe related in #4114 where unmanaged layers are not filtered out actually, i.e. the unmanaged layer from the draw interaction is taken into account when the select interaction uses ```map.forEachFeatureAtPixel```

bartvde:
So it seems my issue in #4114 could be handled with this PR and then:

```
map.addInteraction(new ol.interaction.Select({
  layers: function(layer) {
    if (layer instanceof ol.layer.Vector) {
      return layer.getSource().getFeaturesCollection() === null;
    } else {
      return true;
    }
  } 
}));
```

alexbrault:
> All I know is that the old feature overlays could not be filtered out as well, so I tend to lean towards thinking that this proposed change here is not a good one.

While the old feature overlays couldn't be filtered out, there was a way to effectively ignore them: the callback received a null layer. As it is, there's no simple way for application code to determine whether the layer was filter-approved  or if it's unmanaged.

I'm currently facing a problem where my application includes a Select interaction and a custom interaction using forEachFeatureAtPixel with a layer filter. The custom interaction should only interact with the features in the layers it manages, but unfortunately, it also finds the selected features. I can work around this issue by calling my layer filter myself instead of passing it to forEachFeatureAtPixel, but if this is the solution, it leaves me wondering what's the point of passing a layerFilter if it doesn't really filter out the layers I don't want

ahocevar:
> While the old feature overlays couldn't be filtered out, there was a way to effectively ignore them: the callback received a null layer.

Very good point, thanks for mentioning this. I'm preparing a pull request that brings this behaviour back.

ahocevar:
With #4450 merged, it finally makes sense to make the proposed change to `src/ol/renderer/maprenderer.js`. See #4471 for a remaining issue that's fixed with the proposed change.

ahocevar:
Updated to apply to current master as #4472.

PR #4144, opened 2015-09-21
Set context.fillStyle to a string

### LINKED ISSUE #4074: Feature Animation Example Causes Browser to Become Unresponsive
There is an issue with the Feature Animation example at the URL below:

http://openlayers.org/en/v3.8.2/examples/feature-animation.html 

When this example is loaded, Firefox and Chrome will both eventually become unresponsive.  A sample of the error is below.  The error is repeated in the developer console until the browser freezes.

12:25:36.500 Expected color but found '0'.       from DOM

The issue is repeatable.

elemoine:
The value we set `context.fillStyle` to should always be a string. This commit fixes a bug in circlestyle.js where `context.fillStyle` was set to an array of numbers instead of a string.

Fixes #4074. Thanks @wlerner for his help on tracking down this bug!

### DIFF
--- a/examples/feature-animation.js
+++ b/examples/feature-animation.js
@@ -72,8 +72,7 @@ function flash(feature) {
       snapToPixel: false,
       stroke: new ol.style.Stroke({
         color: 'rgba(255, 0, 0, ' + opacity + ')',
-        width: 1,
-        opacity: opacity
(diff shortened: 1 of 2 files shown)


fredj:
LGTM

PR #4145, opened 2015-09-21
API method for simplifying geometries

### LINKED ISSUE #4058: Using getSimplifiedGeometry() in project
Hi there,

we are using the undocumented function 'getSimplifiedGeometry()' in a project to simplify a LineString using Douglas-Peucker. This works perfectly fine as long as we are using the pre-build version of OL3. However, we would like to use our own custom build OL3. We digged through build/info.json but couldn't find any export that would provide as with getSimplifiedGeometry().

A related, more general question: what is usually the best practice for finding out the export strings for certain functions? For now, we just grep build/info.json for something that looks promising and test it.

Thanks,
patrick

tschaub:
This exposes a method for users to generate simplified geometries.  Internally, we continue to use the verbose/awkward `geometry.getSimplifiedGeometry(squaredTolerance)`.  Users get the more convenient `geometry.simplify(tolerance)` method.

Fixes #4058.

### DIFF
--- a/src/ol/geom/geometry.js
+++ b/src/ol/geom/geometry.js
@@ -170,9 +170,11 @@ ol.geom.Geometry.prototype.getExtent = function(opt_extent) {
 
 
 /**
- * Create a simplified version of this geometry using the {@link
+ * Create a simplified version of this geometry.  For linestrings, this uses
+ * the the {@link
  * http://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm
- * Douglas Peucker} algorithm.
+ * Douglas Peucker} algorithm.  For polygons, a quantization-based
+ * simplification is used to preserve topology.
  * @function
  * @param {number} tolerance The tolerance distance for simplification.
  * @return {ol.geom.Geometry} A new, simplified version of the original
--- a/src/ol/geom/geometry.js
+++ b/src/ol/geom/geometry.js
@@ -172,7 +172,7 @@ ol.geom.Geometry.prototype.getExtent = function(opt_extent) {
 /**
  * Create a simplified version of this geometry.  For linestrings, this uses
  * the the {@link
- * http://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm
+ * https://en.wikipedia.org/wiki/Ramer-Douglas-Peucker_algorithm
  * Douglas Peucker} algorithm.  For polygons, a quantization-based
  * simplification is used to preserve topology.
  * @function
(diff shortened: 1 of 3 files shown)


alvinlindstam:
Last time I checked, we used different algorithms for different types of geometries. Douglas-peucker for LineStrings and another, grid based for polygons. So I guess documentation should be updated for these differences.

alvinlindstam:
One option is to not commit to a certain algorithm, and document the desired output instead. That also has the advantage of allowing changes to getSimplifiedGeometry if needed further down. Something like "Returns a simplified copy of the geometry. The given `tolerance` determines the maximum difference between the original and simplified geometry."

bartvde:
Fixes #4058 as well

tschaub:
> So I guess documentation should be updated for these differences.

Thanks @alvinlindstam.  I updated the docs in 250e86e5f233175b50c58247e867789a230edd92.  In the future, we could add additional arguments that would be used to determine the method, but I'm primarily interested in exposing these methods for now.

> Fixes #4058 as well

@bartvde was that a thumbs up?

bartvde:
yep I like the approach here @tschaub

src/ol/geom/geometry.js
```diff
@@ -170,6 +170,22 @@ ol.geom.Geometry.prototype.getExtent = function(opt_extent) {
...
+ * Create a simplified version of this geometry.  For linestrings, this uses
+ * the the {@link
+ * http://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm
+ * Douglas Peucker} algorithm.  For polygons, a quantization-based
+ * simplification is used to preserve topology.
+ * @function
+ * @param {number} tolerance The tolerance distance for simplification.
+ * @return {ol.geom.Geometry} A new, simplified version of the original
+ *     geometry.
+ */
```
bartvde: doesn't this need @api ?

src/ol/geom/simplegeometry.js
```diff
@@ -144,6 +144,7 @@ ol.geom.SimpleGeometry.prototype.getLayout = function() {
 
 /**
  * @inheritDoc
+ * @api
```
bartvde: this is the internal one right, so it doesn't need the @api ?

src/ol/geom/geometry.js
tschaub: Yikes.  Thanks.  See 436561d9dd8fa58813e1e1276c05672224ae59c3.

src/ol/geom/simplegeometry.js
tschaub: Also fixed in 436561d9dd8fa58813e1e1276c05672224ae59c3.

bartvde:
LGTM now @tschaub

tschaub:

PR #4149, opened 2015-09-22
Remove use of goog.array.isEmpty

fredj:
A small step to remove our dependency on the Closure Library (see #4128)


bartvde:
LGTM

fredj:
Thanks for the review

PR #4150, opened 2015-09-22
Remove use of goog.object.remove

fredj:
If the return value is not used, `goog.object.remove(obj, key)` should be simplified to `delete obj[key]`


fredj:
see #4128

tschaub:
Looks good.  Please merge.

PR #4151, opened 2015-09-22
Remove bogus assertion

elemoine:
This commit removes an incorrect assertion from the modify interaction, which currently triggers when adding vertices to a line string.

The bug can be reproduced in http://jsfiddle.net/5m3897xu/1/. Click on the line string to edit it. Add a vertex on one of the two segments. This should trigger the assertion. The same example works as expected when ol.js is used instead of ol-debug.js.

Please review.

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -973,9 +973,8 @@ ol.interaction.Modify.prototype.setGeometryCoordinates_ =
 ol.interaction.Modify.prototype.updateSegmentIndices_ = function(
     geometry, index, depth, delta) {
   this.rBush_.forEachInExtent(geometry.getExtent(), function(segmentDataMatch) {
-    goog.asserts.assert(goog.isDef(segmentDataMatch.depth));
     if (segmentDataMatch.geometry === geometry &&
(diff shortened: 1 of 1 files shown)


fredj:
This assert was added just to avoid the closure compiler error.

```patch
diff --git a/src/ol/interaction/modifyinteraction.js b/src/ol/interaction/modifyinteraction.js
index 7914ff6..e70891b 100644
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -83,7 +83,7 @@ goog.inherits(ol.interaction.ModifyEvent, goog.events.Event);
 
 
 /**
- * @typedef {{depth: (Array.<number>|undefined),
+ * @typedef {{depth: Array.<number>,
  *            feature: ol.Feature,
  *            geometry: ol.geom.SimpleGeometry,
  *            index: (number|undefined),
```

fixes the issue but is a bit hackish

elemoine:
> fixes the issue but is a bit hackish

`depth` may be `undefined`.

I'll look at fixing the compiler error.

elemoine:
The build now passes on Travis.

fredj:
LGTM, thanks

PR #4153, opened 2015-09-22
Add a "features" option to ol.interaction.Select

elemoine:
This PR adds a `features` option to `ol.interaction.Select`. This option makes it possible to configure a `Select` interaction with a user-created collection.

```js
var featureCollection = new ol.Collection();
var selectInteraction = new ol.interaction.Select({
  features: featureCollection
});
```

We need this in our applications to be able to create independent components where the central point is the collection.

Please review.


tschaub:
Looks good.  Please merge.

PR #4156, opened 2015-09-23
Rename drag-features example to custom-interactions

elemoine:
We now have a `Translate` interaction for dragging features, and a [translate-features example](http://openlayers.org/en/master/examples/translate-features.html). So this PR suggests renaming the `drag-features` example to `custom-interactions`, since this example is really about showing how to create custom interactions.


fredj:
+1

PR #4163, opened 2015-09-23
Fixes for building with Node 4.x

### LINKED ISSUE #4127: Custom build fails with Node 4.0
Following the  [custom build tutorial]( http://openlayers.org/en/v3.9.0/doc/tutorials/custom-builds.html ) using NodeJS v4.0.0 (admittedly just released) fails with:

    $ node tasks/build.js build/ol-custom.json build/ol-custom.js
    ERR! Trouble generating info: Failed to parse output as JSON: . . .

Reverting to NodeJS v0.12.7 everything works as expected.  Only tested on OS 10.10.5 with Java 7

tschaub:
The `generate-info.js` task fails on Node 4.x because `process.stdout.write()` [is now async](https://github.com/nodejs/node/issues/2972), and [our exit handler](https://github.com/openlayers/ol3/blob/dd8422c8a07d36d7a97a69b722c43d39a36925f6/tasks/generate-info.js#L173-L186) is called before the child JSDoc task's publish method [finishes writing to stdout](https://github.com/openlayers/ol3/blob/dd8422c8a07d36d7a97a69b722c43d39a36925f6/config/jsdoc/info/publish.js#L171-L178). 

Because the `generate-info.js` task fails, the `build.js` task does as well.

It would be nice to use a published version of JSDoc.  I'll ask when the next version might be published.  I think this is a big enough issue that we should merge a fix before 3.10 is released - even if it means going with an unpublished version of JSDoc.  I also need to confirm that the API doc output is still good.

Fixes #4127.

### DIFF
--- a/config/jsdoc/api/conf.json
+++ b/config/jsdoc/api/conf.json
@@ -16,7 +16,7 @@
         ]
     },
     "plugins": [
-        "node_modules/jsdoc/plugins/markdown",
+        "node_modules/jsdoc-fork/plugins/markdown",
         "config/jsdoc/api/plugins/inheritdoc",
         "config/jsdoc/api/plugins/typedefs",
         "config/jsdoc/api/plugins/events",
--- a/config/jsdoc/api/plugins/api.js
+++ b/config/jsdoc/api/plugins/api.js
@@ -16,7 +16,7 @@ exports.defineTags = function(dictionary) {
         doclet.stability = level;
       } else {
         var errorText = util.format('Invalid stability level (%s) in %s line %s', tag.text, doclet.meta.filename, doclet.meta.lineno);
-        require('jsdoc/util/error').handle( new Error(errorText) );
+        require('jsdoc-fork/lib/jsdoc/util/error').handle( new Error(errorText) );
       }
     }
   });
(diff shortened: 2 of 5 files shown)


probins:
Promises are natively supported in Node 4. You probably don't need bluebird here.

ahocevar:
Looks great @tschaub, please merge.

@probins I think for now it makes sense to still support older versions of node, so using bluebird should be fine.

fredj:
Can we use jsdoc version 3.3.3?

tschaub:
> Can we use jsdoc version 3.3.3?

See jsdoc3/jsdoc#1070.  We need this commit: https://github.com/jsdoc3/jsdoc/commit/7d9fc7b2009350aaf1146875944832e8a03b30e3 (which is not in a published release yet).

The change we need is included in https://github.com/jsdoc3/jsdoc/commit/bd1cb9d1663bdccc68f1ba242a7882f020e78a03.  Which is understandably the source of a lot of confusion.

tschaub:
I've published a [forked version](https://www.npmjs.com/package/jsdoc-fork) of jsdoc.  I don't intend to maintain this fork, but I'd very much like to get our builds working with both Node 4 and npm 3.  The master branch of jsdoc has the changes we need for Node 4 (the changes in jsdoc3/jsdoc#1082 would make this even better).  If a new version is published with these changes, we can switch back.

tschaub:
Note that when/if there is a jsdoc version published that works with Node 4, we need to replace all `jsdoc-fork` in our source with `jsdoc`.

PR #4164, opened 2015-09-23
Fix for inline SVG icons in IE11 (messed up the old branch)

spatialillusions:
Sorry for my several pulls, but I had never rebased a branch and messed everything up... :-S At least now I know how to do it.

However, everything should work now, and you should be able to merge the pull without any problems. 

This pulls solve the problem that already loaded images never calls determineTainting_() and because of this might taint the canvas. Inline SVG images can in this case be considered as already loaded images. With the current code there is no problem to draw inline SVG icons to the map, but if you try to add interactions to them, like making it possible to move points, you will get a security error in IE11. By changing the lines in my pull everything will work as it should.

### DIFF
--- a/src/ol/style/iconstyle.js
+++ b/src/ol/style/iconstyle.js
@@ -426,7 +426,7 @@ ol.style.IconImage_ = function(image, src, size, crossOrigin, imageState) {
    * @type {boolean}
    */
   this.tainting_ = false;
-
+  if (this.imageState_ == ol.style.ImageState.LOADED)this.determineTainting_();
 };
 goog.inherits(ol.style.IconImage_, goog.events.EventTarget);
 
@@ -456,8 +456,8 @@ ol.style.IconImage_.get = function(image, src, size, crossOrigin, imageState) {
  */
 ol.style.IconImage_.prototype.determineTainting_ = function() {
   var context = ol.dom.createCanvasContext2D(1, 1);
-  context.drawImage(this.image_, 0, 0);
   try {
+    context.drawImage(this.image_, 0, 0);
     context.getImageData(0, 0, 1, 1);
   } catch (e) {
     this.tainting_ = true;


src/ol/style/iconstyle.js
```diff
@@ -426,7 +426,7 @@ ol.style.IconImage_ = function(image, src, size, crossOrigin, imageState) {
    * @type {boolean}
    */
   this.tainting_ = false;
-
+  if (this.imageState_ == ol.style.ImageState.LOADED)this.determineTainting_();
```
marcjansen: We usually have even one-line branches inside of curly braces.

marcjansen:
I have added a small comment, but am in no position to review this with the love it needs. Maybe @elemoine and @fredj can have a look?

@spatialillusions Do we have a CLA from you?

ahocevar:
We do have a CLA from @spatialillusions.

marcjansen:
Would this be an example, which one can use to test the changes? http://spatialillusions.com/milsymbol/examples/openlayers3/imageicons.html

@ahocevar: nice, thanks. How do find this out so fast?

ahocevar:
@marcjansen: https://github.com/openlayers/cla/blob/master/signed-agreements.md

marcjansen:
:tada: :+1:

spatialillusions:
Here is a jsfiddle for testing. It is when you add interactions to your vector features that things break in IE11 if you are using inline SVG icons or preloaded images.
http://jsfiddle.net/spatialillusions/5jx0dzpu/

Expected behavior: (Present i browsers other than IE11 or lower)
You are able to see a symbol representing an infantry platoon on the map and you are able to move the icon to a new location.

Behavior in IE11:
You can see an icon representing an infantry platoon, but if you try to move it everything will freeze in the browser and you will get a security error in the console.

src/ol/style/iconstyle.js
```diff
@@ -456,8 +456,8 @@ ol.style.IconImage_.get = function(image, src, size, crossOrigin, imageState) {
  */
 ol.style.IconImage_.prototype.determineTainting_ = function() {
   var context = ol.dom.createCanvasContext2D(1, 1);
-  context.drawImage(this.image_, 0, 0);
   try {
+    context.drawImage(this.image_, 0, 0);
```
ahocevar: Are you sure it it necessary to move the drawImage() call into the try/catch block?

spatialillusions: I did try to keep it outside first, but I still got problems in IE11 then. At the moment I had an idea of the reason for this, but it slipped my mind now... But to answer your question, yes I had to move the drawImage into the try/catch block.

ahocevar: The only explanation I have is that if the canvas was tainted already, it would fail. But with the code you added in the constructor, this should not be the case.

src/ol/style/iconstyle.js
ahocevar: @spatialillusions Any chance you could fix this coding style issue? If not let me know, then I can do it before merging.

ahocevar:
For the record: the fix is simply that when the canvas is tainted, the bounding rectangle of the icon will be used for hit detection instead of the icon itself. This is the case for all browsers, and was already before this change. All this change does is that it makes things work also for images that are already loaded when the icon style is created, and that the try/catch block will not only handle `getImageData`, but also `drawImage`. The performance impact of this change should be minor, since icon styles are usually reused and loading only happens once.

src/ol/style/iconstyle.js
spatialillusions: If you could fix it would be great! I need to practice my rebasing and updating of existing pulls elsewhere so I don't mess anything up. ;-)

PR #4165, opened 2015-09-24
Use ECMAScript 5.1 Array functions instead of goog.array

fredj:


fredj:
Thanks for any review

src/ol/interaction/selectinteraction.js
```diff
@@ -264,7 +264,7 @@ ol.interaction.Select.handleEvent = function(mapBrowserEvent) {
          * @param {ol.layer.Layer} layer Layer.
          */
         function(feature, layer) {
-          var index = goog.array.indexOf(features.getArray(), feature);
+          var index = features.getArray().indexOf(feature);
           if (index == -1) {
```
gberaudo: Could use `===`.

gberaudo:
LGTM.
Added a comment about a minor and unrelated improvement.

src/ol/interaction/selectinteraction.js
tschaub: Or if #4181 makes it in first, this could be shortened to `if (!ol.array.includes(features.getArray(), feature) {}`.

tschaub:
Looks good to me as well.

src/ol/interaction/selectinteraction.js
fredj: Now in https://github.com/fredj/ol3/commit/da611b9cb54ca6627b9b07a2a89a2447002a0531

PR #4166, opened 2015-09-24
Use version 20150920 of Closure Compiler

elemoine:
This PR updates OpenLayers to use the latest version of Closure Compiler (20150920). This requires using a recent version of Closure Library as well. It also requires a minor change to the `ol.inherits`'s signature.

### DIFF
--- a/closure-util.json
+++ b/closure-util.json
@@ -1,3 +1,3 @@
 {
-  "library_url": "https://github.com/google/closure-library/archive/a5f9b8a.zip"
+  "library_url": "https://github.com/google/closure-library/archive/5b25e65.zip"
 }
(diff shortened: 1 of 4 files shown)


elemoine:
I had to remove the `node_modules` cache to make Travis happy.

marcjansen:
You can delete the cache via travis webinterface, no need to change the yaml.

elemoine:
Thanks Marc. I have other issues now, related to Slimer and https://github.com/laurentj/slimerjs/commit/911fcb671a36049bd3ca5e7b8d5824a8d7f01dc7.

fredj:
+1 (we should add a note somewhere to not forget to use the `slimerjs` npm package instead of `slimerjs-edge` on the next release)

elemoine:
Thanks for the suggestion @fredj. I amended my last commit.

fredj:
please merge

Issue #4147, opened 2015-09-21
Tile Preload Example Documentation Description Error


wlerner:
This is a minor documentation issue.

There is an error of the description in the Tile Preload Example found at the URL below:

http://openlayers.org/en/v3.9.0/examples/preload.html

Specifically, the description states that the two visible maps are next to each other (viewed left to right) when in reality they are rendered on top of each other.  See the screen shot below and note the description.

![screenshot from 2015-09-21 17 43 41](https://cloud.githubusercontent.com/assets/14095650/10006010/8f1e87f0-6088-11e5-8415-95a5f75af5e3.png)

tschaub:
Thanks for catching this @wlerner.  I think it is actually a markup/css issue, but I would be in favor of simplifying the markup (not using Bootstrap to try to create a fluid layout with the maps in columns), and updating the text.

Changing the example text should make for a straightforward pull request.  The source for the text is in the `docs` metadata for the example: https://github.com/openlayers/ol3/blob/f65129850d43671f4835e7278ca4b14ed23a5e6e/examples/preload.html#L6

PR #4168, opened 2015-09-24
 Tile Preload Example Documentation Description Error #4147 Fix

wlerner:
Modified the preload.html example to reflect the correct position of the maps displayed in the example when viewed with a web browser.  This is related to #4147


tschaub:
Looks good.  Thanks @wlerner.

PR #4173, opened 2015-09-25
Only animate when resolution is about to be changed

### LINKED ISSUE #4115: Vector feature are moving after animation end
This odd behaviour seems to have appeared in v3.8.x., i.e. I can't reproduce this in v3.7.0: in a map that has vector features in it, after the zoom animation ends, we can see the features slightly move around.  This even occur when zooming out completely.  In fact, it's even more noticeable that way.

See http://openlayers.org/en/v3.8.2/examples/icon.html.  Zoom out using the mouse scroll.  Upon reaching the end, continue to scroll out.  You'll see the behaviour occur.

See in http://openlayers.org/en/v3.7.0/examples/icon.html, that didn't happen.

ahocevar:
Fixes #4115.

The reason for the observed behaviour is because for rendering, the center is snapped to pixels ((`view.getState().center`)). During the animation when fully zoomed in or out, floating point calculations cause the pixel snapped center to differ by one pixel at different steps in the animation.

### DIFF
--- a/src/ol/interaction/interaction.js
+++ b/src/ol/interaction/interaction.js
@@ -222,6 +222,7 @@ ol.interaction.Interaction.zoomWithoutConstraints =
     var currentResolution = view.getResolution();
     var currentCenter = view.getCenter();
     if (currentResolution !== undefined && currentCenter &&
+        resolution !== currentResolution &&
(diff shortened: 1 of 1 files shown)


tschaub:
@ahocevar looks good, but needs a rebase.

PR #4174, opened 2015-09-25
Remove hue, saturation, contrast, and brightness as layer properties

### LINKED ISSUE #4130: Remove brightness, hue, saturation, and contrast as layer properties
These properties are only supported by the WebGL renderer.  Instead of porting the functionality to the Canvas renderer, I think it makes sense to remove them entirely.

What do others think?

tschaub:
This removes the experimental hue, saturation, contrast, and brightness properties from layers.

See https://github.com/openlayers/ol3/issues/4130#issuecomment-142621941 for some justification.

Fixes #4130.


ahocevar:
+1, looks great to me.

PR #4175, opened 2015-09-27
Remove use of goog.math.clamp()

tschaub:
This replaces `goog.math.clamp()` with `ol.math.clamp()`.

See #4128 for context.


ahocevar:
Very nice! Please merge.

PR #4176, opened 2015-09-27
Remove use of goog.isDefAndNotNull()

tschaub:
This removes our use of `goog.isDefAndNotNull()`.  In most cases, simple truthy checks are preferable.

@bartvde another interesting issue revealed by this change.  When I replaced the `goog.isDefAndNotNull(esriJSONFeature.compressedGeometry)` call with a simple truthy check, the compiler discovered that the `compressedGeometry` property is not defined for `EsriJSONFeature`.  Since this was only an assertion and the `compressedGeometry` property was never accessed elsewhere, I removed this part of the assertion.

@fredj (or anyone else) - do we ever expect properties like `alpha` to be `null` for a `DeviceOrientationEvent` (or will these be non-null if defined).  The only remaining use of `goog.isDefAndNotNull()` is in the device orientation code, since I was uncertain if these values could really be `null`.

See #4128 for context.

### DIFF
--- a/src/ol/source/tileutfgridsource.js
+++ b/src/ol/source/tileutfgridsource.js
@@ -131,7 +131,7 @@ ol.source.TileUTFGrid.prototype.handleTileJSONResponse = function(tileJSON) {
   this.template_ = tileJSON.template;
 
   var grids = tileJSON.grids;
-  if (!goog.isDefAndNotNull(grids)) {
+  if (!grids) {
     this.setState(ol.source.State.ERROR);
     return;
   }
--- a/src/ol/source/tileutfgridsource.js
+++ b/src/ol/source/tileutfgridsource.js
@@ -268,8 +268,7 @@ ol.source.TileUTFGridTile_.prototype.getImage = function(opt_context) {
  * @return {Object}
  */
 ol.source.TileUTFGridTile_.prototype.getData = function(coordinate) {
-  if (goog.isNull(this.grid_) || goog.isNull(this.keys_) ||
-      !goog.isDefAndNotNull(this.data_)) {
+  if (goog.isNull(this.grid_) || goog.isNull(this.keys_) || !this.data_) {
     return null;
   }
   var xRelative = (coordinate[0] - this.extent_[0]) /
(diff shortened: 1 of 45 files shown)


fredj:
> do we ever expect properties like `alpha` to be `null` for a `DeviceOrientationEvent` (or will these be non-null if defined).

The angles may be `null`: "Implementations that are unable to provide all three angles must set the values of the unknown angles to null. "
Source: http://www.w3.org/TR/orientation-event/#deviceorientation

bartvde:
wrt compressedGeometry, this is very likely a left-over from something I removed in the end, so good to get rid of that assertion @tschaub

tschaub:
> The angles may be null: "Implementations that are unable to provide all three angles must set the values of the unknown angles to null. "
> Source: http://www.w3.org/TR/orientation-event/#deviceorientation

Thanks for the pointer @fredj.  So, do you think `event.alpha !== null` is the right check, or do we also need to check for `undefined`?

fredj:
> Thanks for the pointer @fredj.  So, do you think `event.alpha !== null` is the right check, or do we also need to check for `undefined`?

`event.alpha !== null` is sufficient

tschaub:
This is now ready for review.

tschaub:
@openlayers/ol3 - as [mentioned above](https://github.com/openlayers/ol3/pull/4176#issuecomment-144105089), this is ready for review.

src/ol/source/tileutfgridsource.js
```diff
@@ -292,9 +291,7 @@ ol.source.TileUTFGridTile_.prototype.getData = function(coordinate) {
   }
   code -= 32;
 
-  var key = this.keys_[code];
-
-  return goog.isDefAndNotNull(key) ? this.data_[key] : null;
+  return (code in this.keys_) ? this.data_[this.keys_[code]] : null;
```
marcjansen: I like this change!

marcjansen:
This looks good to me. Please merge and thanks.

tschaub:
Device orienteers (assuming you find your way back), please let us know if this breaks your applications.

PR #4178, opened 2015-09-28
Replace goog.isNull() with !== or === checks

marcjansen:
This PR includes commits that replace the use of `goog.isNull` with either `expression === null` or `expression !== null`.

I have split the commits in to multiple smaller groups, so reviewing is made easier.

Please review.

See #4128

### DIFF
--- a/src/ol/animation.js
+++ b/src/ol/animation.js
@@ -115,7 +115,7 @@ ol.animation.rotate = function(options) {
               (sourceRotation - frameState.viewState.rotation) * delta;
           frameState.animate = true;
           frameState.viewState.rotation += deltaRotation;
-          if (!goog.isNull(anchor)) {
+          if (anchor !== null) {
             var center = frameState.viewState.center;
             ol.coordinate.sub(center, anchor);
             ol.coordinate.rotate(center, deltaRotation);
--- a/src/ol/attribution.js
+++ b/src/ol/attribution.js
@@ -61,7 +61,7 @@ ol.Attribution.prototype.getHTML = function() {
  */
 ol.Attribution.prototype.intersectsAnyTileRange =
     function(tileRanges, tileGrid, projection) {
-  if (goog.isNull(this.tileRanges_)) {
+  if (this.tileRanges_ === null) {
     return true;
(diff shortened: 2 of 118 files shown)


tschaub:
Nice work @marcjansen.

I think there might still be a lack of consensus on this, but I'm in favor of using simple truthy checks where sensible.  I haven't looked through every line, but I imagine most here could be changed to be a truthy check.

I'm not suggesting that you make this change @marcjansen.  But I would like to hear what others think.  Is there a justification for `if (foo !== null) {}` rather than `if (foo) {}` for type `Object` and friends?

marcjansen:
I'd be interested as well, @tschaub. An update of this PR wouldn't be to hard, and I can do it once we have consensus.

Maybe @elemoine or @fredj know if the compiler benefits from either way, generally speaking.

marcjansen:
Here is an alternative branch that tries to use as few !== / === when checing against null as possible.

* Comparison to [this PR](https://github.com/marcjansen/ol3/compare/no-is-null...marcjansen:simpler-no-is-null?expand=1)
* Comparison to [this master](https://github.com/marcjansen/ol3/compare/openlayers:master...marcjansen:simpler-no-is-null?expand=1)

Build results

```
info ol Parsing dependencies
```

**`no-is-null`* (this branch)*
```
info ol Compiling 354 sources
uncompressed: 477982 bytes
  compressed: 141754 bytes
```

**`simpler-no-is-null`** (the alternative linked above)
```
info ol Compiling 354 sources
uncompressed: 474650 bytes
  compressed: 141269 bytes
```

**`master`**
```
info ol Compiling 357 sources
uncompressed: 484100 bytes
  compressed: 143067 bytes
```

tschaub:
`simpler-no-is-null` FTW!  Not just because of the byte savings - I think the `if (foo) {}` style is preferable.

marcjansen:
Ok. I'll create a replacement PR soon. Since most of the changes were done with regexes in the editor, I want to double check the outcome first.

marcjansen:
Closing in favor of #4192

PR #4179, opened 2015-09-28
Use Date.now() instead of goog.now()

marcjansen:
This PR suggests using `Date.now()` instead of `goog.now()`.

Date.now() is supported in all major browsers (IE from v9 onwards): http://kangax.github.io/compat-table/es5/#Date.now

The current implementation of `goog.now` is a little bit more complicated, but I feel we do not want to copy their implementation: https://github.com/google/closure-library/blob/master/closure/goog/base.js#L1884-L1888

The closure library (optionally) guards against non-trusted sites (in which some other code may have changed e.g. `Date.now`) and falls back to using `+new Date()` in those cases or IE < 9.

Please review.

See #4128

### DIFF
--- a/src/ol/animation.js
+++ b/src/ol/animation.js
@@ -16,7 +16,7 @@ goog.require('ol.easing');
  */
 ol.animation.bounce = function(options) {
   var resolution = options.resolution;
-  var start = options.start ? options.start : goog.now();
+  var start = options.start ? options.start : Date.now();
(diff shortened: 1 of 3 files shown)


fredj:
LGTM

marcjansen:
Cheers for the review.

elemoine:
Using `Date.now` makes great sense to me.

PR #4162, opened 2015-09-28
Remove ol.LEGACY_IE_SUPPORT

fredj:
Remove the optional IE 7-8 support and convert our code base to ECMAScript 5.

This will allow us to use the native `Array.prototype.forEach` instead of `goog.array.forEach` (same for `goog.array.map`, `goog.object.getKeys`, ...)

### DIFF
--- a/src/ol/dom/dom.js
+++ b/src/ol/dom/dom.js
@@ -30,15 +30,6 @@ ol.dom.createCanvasContext2D = function(opt_width, opt_height) {
 
 
 /**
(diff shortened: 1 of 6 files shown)


tschaub:
Thanks for tackling this @fredj.  Please merge.

elemoine:
@fredj, don't you think we should mention this change in the release notes?

fredj:
Yes, it must be mentioned

PR #4181, opened 2015-09-28
Remove use of goog.array.contains

fredj:
Use `arr.indexOf(item) >= 0` instead of `goog.array.contains(arr, item)`

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -38,6 +38,17 @@ ol.array.binaryFindNearest = function(arr, target) {
 
 
 /**
+ * Whether the array contains the given object.
+ * @param {Array.<*>} arr The array to test for the presence of the element.
+ * @param {*} obj The object for which to test.
+ * @return {boolean} The object is in the array.
+ */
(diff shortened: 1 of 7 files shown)


tschaub:
Doesn't need to be here, but if we ever want consistency with `Array.prototype.includes`, we'll need to handle `NaN` correctly (`indexOf` doesn't).

tschaub:
To be clear, I *don't* think we need to handle `NaN` differently than `indexOf` does right now (since `goog.array.contains` uses `indexOf` as well).

I think this looks good to merge.

src/ol/array.js
```diff
@@ -38,6 +38,17 @@ ol.array.binaryFindNearest = function(arr, target) {
 
 
 /**
+ * Whether the array contains the given object.
+ * @param {Array.<*>} arr The array to test for the presence of the element.
+ * @param {*} obj The object for which to test.
+ * @return {boolean} true if obj is present.
```
tschaub: This could read "The object is in the array."

PR #4186, opened 2015-09-28
Do not export ol.webgl.Context

elemoine:
We don't need to export the `ol.export.Context` constructor as users won't create `ol.webgl.Context` instances themselves.

I suppose we added an `@api` annotation on `ol.webgl.Context` for documentation purposes. This is a shame that we need to add `@api` annotations on things we don't want to export. We increase the build size just for the purpose of the documentation!

### DIFF
--- a/src/ol/webgl/context.js
+++ b/src/ol/webgl/context.js
@@ -26,7 +26,6 @@ ol.webgl.BufferCacheEntry;
  * @extends {goog.events.EventTarget}
  * @param {HTMLCanvasElement} canvas Canvas.
  * @param {WebGLRenderingContext} gl GL.
- * @api
  */
 ol.webgl.Context = function(canvas, gl) {
 


tschaub:
> This is a shame that we need to add @api annotations on things we don't want to export. We increase the build size just for the purpose of the documentation!

This is a bit of legacy from the Plovr days.  When we couldn't tell the Compiler about `extra_annotation_name`, we crammed everything into `@todo`.  The `@todo api` plugin was extended to handle too much.

I think we should add a `@doc` annotation or something else specific to the documentation plugins.  This will probably take a bit of digging/discussion to determine what we really need.

@ahocevar do you want to try to coordinate on this?

ahocevar:
It should be enough to just remove `@api` annotations from constructors that we don't want people to create instances from. The apidoc plugin will still create documentation if there are `@api` members, but will not list the constructor in the doc page.

elemoine:
Do you mean it already works this way?

ahocevar:
Yes, it should.

ahocevar:
I just confirmed that this works. Please merge.

elemoine:
Thanks.

PR #4187, opened 2015-09-28
Fix {-y} template calculation for custom (TMS) tile grids

ahocevar:
Previously, `{-y}` only worked for the standard web mercator tile grid. Now a tile grid with an extent is required (which we get from `ol.tilegrid.createXYZ()` anyway), and then the `y` calculation for TMS style
tile grids works as expected.

### DIFF
--- a/src/ol/source/tilejsonsource.js
+++ b/src/ol/source/tilejsonsource.js
@@ -75,7 +75,8 @@ ol.source.TileJSON.prototype.handleTileJSONResponse = function(tileJSON) {
   });
   this.tileGrid = tileGrid;
 
-  this.tileUrlFunction = ol.TileUrlFunction.createFromTemplates(tileJSON.tiles);
+  this.tileUrlFunction =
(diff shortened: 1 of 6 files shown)


ahocevar:
Thanks for any review.

tschaub:
Your changes look great.  I'm relying on the tests (sketchy, I know) instead of testing anything locally.

This may be unwarranted, but this is confirming a feeling I already have that storing the URL (template) and using an instance method would be more sensible than the `ol.TileUrlFunction.createFromTemplate` factory (since I've run into cases where I wanted to get the URL (template) to compare if it was different than what I was about to set, and the `ol.TileUrlFunction` was an annoyance).

tschaub:
By which I meant: please merge.

ahocevar:
Thanks for the review.

I have this feeling too. After my ongoing vector tile work, the whole tileUrlFunction handling will be in a single place, and then it will be easier to make such a change. So hold on.

PR #4188, opened 2015-09-29
Remove use of goog.object.getKeys

fredj:
Use `Object.keys` instead.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys


tschaub:
Looks good.  Please merge.

fredj:
Thanks for the review

PR #4192, opened 2015-09-29
Remove use of goog.isNull in favor of simple truthy checks

marcjansen:
This PR contains commits that use simple truthy/falsy checks instead of `goog.isNull` or `foo !== null` and friends. I have double checked every change herein, but would really like a thorough review.

One change has been made to the previous logic, which I consider a bug in the original code: https://github.com/openlayers/ol3/pull/4192/files#diff-f6fd0d8a60d56e0391a62ade939c9719L106 

Previousy the lines read

```javascript
// …
} else if (goog.isNull(this.lastFeature_)) {
  var geom = this.lastFeature_.getGeometry();
  // …
}
```

now these lines read

```javascript
// …
} else if (this.lastFeature_) {
  var geom = this.lastFeature_.getGeometry();
  // …
}
```

which makes a lot more sense.

Thanks for any review.

This PR supersedes #4178.

 See #4128

/cc @tschaub

### DIFF
--- a/src/ol/format/gpxformat.js
+++ b/src/ol/format/gpxformat.js
@@ -97,7 +97,7 @@ ol.format.GPX.parseLink_ = function(node, objectStack) {
   goog.asserts.assert(node.localName == 'link', 'localName should be link');
   var values = /** @type {Object} */ (objectStack[objectStack.length - 1]);
   var href = node.getAttribute('href');
-  if (!goog.isNull(href)) {
+  if (href !== null) {
     values['link'] = href;
   }
   ol.xml.parseNode(ol.format.GPX.LINK_PARSERS_, node, objectStack);
@@ -417,7 +417,7 @@ ol.format.GPX.WPT_PARSERS_ = ol.xml.makeStructureNS(
  * @private
  */
 ol.format.GPX.prototype.handleReadExtensions_ = function(features) {
-  if (goog.isNull(features)) {
+  if (!features) {
     features = [];
   }
   for (var i = 0, ii = features.length; i < ii; ++i) {
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -243,7 +243,7 @@ ol.interaction.Modify.prototype.addFeature_ = function(feature) {
     this.SEGMENT_WRITERS_[geometry.getType()].call(this, feature, geometry);
   }
   var map = this.getMap();
-  if (!goog.isNull(map)) {
+  if (map) {
     this.handlePointerAtPixel_(this.lastPixel_, map);
   }
   goog.events.listen(feature, goog.events.EventType.CHANGE,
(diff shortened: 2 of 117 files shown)


tschaub:
> One change has been made to the previous logic, which I consider a bug in the original code: https://github.com/openlayers/ol3/pull/4192/files#diff-f6fd0d8a60d56e0391a62ade939c9719L106

Yeah, I agree.  It would be nice to confirm that running this code is the right thing to do, since it would clearly throw as is.  If it is never supposed to run, then it should be removed altogether.  @Morgul or @elemoine any comments?

src/ol/format/gpxformat.js
```diff
@@ -97,7 +97,7 @@ ol.format.GPX.parseLink_ = function(node, objectStack) {
   goog.asserts.assert(node.localName == 'link', 'localName should be link');
   var values = /** @type {Object} */ (objectStack[objectStack.length - 1]);
   var href = node.getAttribute('href');
-  if (!goog.isNull(href)) {
+  if (href !== null) {
```
tschaub: Trying to imagine where someone would actually want `''` to make it through here.  Can't imagine it, but I think the conservative approach is ok too.

src/ol/map.js
```diff
@@ -1166,7 +1166,7 @@ ol.Map.prototype.isDef = function() {
  * @return {boolean} Is rendered.
  */
 ol.Map.prototype.isRendered = function() {
-  return !goog.isNull(this.frameState_);
+  return this.frameState_ !== null;
```
tschaub: Very minor byte savings with `return !!this.frameState_;`.

src/ol/interaction/modifyinteraction.js
```diff
@@ -584,7 +583,7 @@ ol.interaction.Modify.handleDownEvent_ = function(evt) {
         new ol.interaction.ModifyEvent(ol.ModifyEventType.MODIFYSTART,
         this.features_, evt));
   }
-  return !goog.isNull(this.vertexFeature_);
+  return this.vertexFeature_ !== null;
```
tschaub: Minor again, but this could be `return !!this.vertexFeature_;`.

src/ol/format/gpxformat.js
marcjansen: I thought the same. I cannot think of a good reason for `''`, but wanted this to behave as it did before

tschaub:
This is so good.  Very minor comments above - which you can take or leave.  It would be nice to get feedback on the translation interaction change, but that should not block.  I think this is good to merge.  Thanks for taking the time to rework this - I hope others appreciate the significant code-readability improvements (imagine any newcomer pondering all the presumably pointless `goog.isNull()` calls).

marcjansen:
Thanks for the review, Tim. I agree readability increased a lot with this and related changes. Parts of the library now actually look like JavaScript :smile:, just kidding.

I will update the PR with your suggestion as soon as I have access to a computer.

Morgul:
@tschaub As for the correct thing to do, it's been long enough since I worked on that, I'm not sure, but looking at it again, I think it's safe to remove. My confidence level's only about 80%, so we should probably wait for @elemoine to chime in before doing anything drastic.

marcjansen:
I will create a separate issue for this, if everybody agrees.

tschaub:
> I will create a separate issue for this, if everybody agrees.

Yes.  I think it makes sense to include the change you already made (not drastic, avoids throwing), and create a ticket to see if the whole block should be removed.

PR #4194, opened 2015-09-29
Remove usage of goog.color and closure named colors

bjornharrtell:
There are some mentions of goog.color in the comments, perhaps those should be removed too? In that case the private member ol.color.rgbColorRe_ will have no explanation, is that ok?

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+#### Removal of `ol.ENABLE_NAMED_COLORS`
+
+This option was previously needed to use named colors with the WebGL renderer but is no longer needed.
+
 ### v3.17.0
 
 #### `ol.source.MapQuest` removal
--- a/src/ol/color/color.js
+++ b/src/ol/color/color.js
@@ -1,13 +1,6 @@
-// We can't use goog.color or goog.color.alpha because they interally use a hex
-// string representation that encodes each channel in a single byte.  This
-// causes occasional loss of precision and rounding errors, especially in the
-// alpha channel.
-
 goog.provide('ol.color');
 
 goog.require('goog.asserts');
-goog.require('goog.color');
-goog.require('goog.color.names');
 goog.require('ol');
 goog.require('ol.math');
 
@@ -22,7 +15,7 @@ ol.color.hexColorRe_ = /^#(?:[0-9a-f]{3}){1,2}$/i;
 
 
 /**
- * @see goog.color.rgbColorRe_
+ * Regular expression for matching and capturing RGB style strings.
  * @const
  * @type {RegExp}
  * @private
@@ -32,13 +25,23 @@ ol.color.rgbColorRe_ =
 
 
 /**
- * @see goog.color.alpha.rgbaColorRe_
+ * Regular expression for matching and capturing RGBA style strings.
  * @const
  * @type {RegExp}
  * @private
  */
 ol.color.rgbaColorRe_ =
     /^(?:rgba)?\((0|[1-9]\d{0,2}),\s?(0|[1-9]\d{0,2}),\s?(0|[1-9]\d{0,2}),\s?(0|1|0\.\d{0,10})\)$/i;
+
+
+/**
+ * Regular expression for matching potential named color style strings.
+ * @const
+ * @type {RegExp}
+ * @private
+ */
+ol.color.namedColorRe_ =
+    /^([a-z]*)$/i;
 
 
 /**
@@ -71,6 +74,20 @@ ol.color.asString = function(color) {
     goog.asserts.assert(Array.isArray(color), 'Color should be an array');
     return ol.color.toString(color);
   }
+};
+
+/**
+ * Return named color as an rgba string.
+ * @param {string} color Named color.
+ * @return {string} Rgb string.
+ */
+ol.color.fromNamed = function(color) {
+  var el = document.createElement('div');
+  el.style.color = color;
+  document.body.appendChild(el);
+  var rgb = window.getComputedStyle(el).color;
+  document.body.removeChild(el);
+  return rgb;
 };
 
 
@@ -137,15 +154,13 @@ ol.color.fromString = (
  * @return {ol.Color} Color.
  */
 ol.color.fromStringInternal_ = function(s) {
+  var r, g, b, a, color, match;
 
-  var isHex = false;
-  if (ol.ENABLE_NAMED_COLORS && goog.color.names.hasOwnProperty(s)) {
-    s = goog.color.names[s];
-    isHex = true;
+  if (ol.color.namedColorRe_.exec(s)) {
+    s = ol.color.fromNamed(s);
   }
 
-  var r, g, b, a, color, match;
-  if (isHex || (match = ol.color.hexColorRe_.exec(s))) { // hex
+  if (ol.color.hexColorRe_.exec(s)) { // hex
     var n = s.length - 1; // number of hex digits
     goog.asserts.assert(n == 3 || n == 6,
         'Color string length should be 3 or 6');
--- a/src/ol/ol.js
+++ b/src/ol/ol.js
@@ -79,16 +79,6 @@ ol.ENABLE_IMAGE = true;
 
 
 /**
- * @define {boolean} Enable Closure named colors (`goog.color.names`).
- *     Enabling these colors adds about 3KB uncompressed / 1.5KB compressed to
- *     the final build size.  Default is `false`. This setting has no effect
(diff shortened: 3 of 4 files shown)


tschaub:
> There are some mentions of goog.color in the comments, perhaps those should be removed too

Yes, these comments should be updated/removed.

In addition, we'll need to add upgrade notes for people who were using `ol.ENABLE_NAMED_COLORS`.

sweco-sebhar:
Amended with updated comments and upgrade note.

src/ol/color/color.js
```diff
@@ -208,11 +201,6 @@ ol.color.fromString = (
 ol.color.fromStringInternal_ = function(s) {
 
   var isHex = false;
-  if (ol.ENABLE_NAMED_COLORS && goog.color.names.hasOwnProperty(s)) {
-    // goog.color.names does not have a type declaration, so add a typecast
-    s = /** @type {string} */ (goog.color.names[s]);
-    isHex = true;
-  }
```
tschaub: It looks like `isHex` should be removed entirely now.

changelog/upgrade-notes.md
```diff
@@ -6,6 +6,11 @@
 
 The experimental `setHue`, `setContrast`, `setBrightness`, `setSaturation`, and the corresponding getter methods have been removed.  These properties only worked with the WebGL renderer.  If are interested in applying color transforms, look for the `postcompose` event in the API docs.  In addition, the `ol.source.Raster` source provides a way to create new raster data based on arbitrary transforms run on any number of input sources.
 
+#### Removal of `ol.ENABLE_NAMED_COLORS`
+
+This option existed to support the use of Google Closure library named colors and was removed as a part of OpenLayers moving away from Google Closure library. Note that this do not affect the use of CSS named colors.
```
marcjansen: Typo: ... this do**es** not ...

marcjansen: Should we more explicit here? Sth. like: if you have used the canvas renderer (which is the default) for your vector layers, this change will not affect you. If you have been using the webgl renderer though, you cannot any longer use named colors but must provide them in <a suitable format.

tschaub: I agree it would be nice to be more explicit about what this change means.  I would remove all reference to the Google Closure Library, since I don't think that is relevant.  It would be more relevant for people to know if they need to change their application code.

sweco-sebhar: How about:

Note that this option default to `false` and that it has no effect on the canvas renderer (which is the default). If you have relied on this option to use named colors (created with `ol.color.fromString`) for use with the WebGL renderer you will need to translate them to RGB values.

bjornharrtell:
Thanks for the input, updated branch with isHex removed and new update notes based on feedback.

gberaudo:
This removal will affect our ability to handle named colors in the WEBGL renderer.
To avoid this, a straightforward solution is to copy the [goog color table](
https://closure-library.googlecode.com/git-history/docs/local_closure_goog_color_names.js.source.html#line28) into our code.

bjornharrtell:
Sure, could do that. I was actually initially going that way but decided/thought it wasn't a feature worth keeping. Opinions from previous reviewers @tschaub and @marcjansen?

marcjansen:
Fine for me if it isn't included if WEBGL isn't included.

sweco-sebhar:
I've made so that this PR only removes the use of goog.color. Cannot change the title of the PR to reflect that `ol.ENABLE_NAMED_COLORS` is left instact, but the commit message should be ok.

bjornharrtell:
I've made so that this PR only removes the use of goog.color. Cannot change the title of the PR to reflect that ol.ENABLE_NAMED_COLORS is left instact, but the commit message should be ok.

gberaudo:
Looks great to me. Thank you @bjornharrtell.

ahocevar:
I really think we don't need named colors in our library. If someone wants to use named colors with WegGL, this can be translated with something like
```js
function getRgbColor(color) {
  var el = document.createElement("div");
  el.style.color = "white";
  document.body.appendChild(el);
  var rgb = window.getComputedStyle(el).color;
  document.body.removeChild(el);
};
```
This could be done in the WebGL renderer, and of course `el` would not have to be created, added to the DOM and removed with every single call.

bjornharrtell:
I agree. I'll wait for consensus before having another go at this PR.

ahocevar:
Sorry @bjornharrtell for the back and forth on this.

gberaudo:
@ahocevar, using `window.getComputedStyle` sounds like a great idea.

Having the conversion in `ol.color.fromStringInternal_` would notably allow to share the cache; I think it would be better.

It would also make sense to have `ol.ENABLE_NAMED_COLORS` enabled by default for WEBGL users and allow to disable it as an optimization.

ahocevar:
> Having the conversion in ol.color.fromStringInternal_ would notably allow to share the cache; I think it would be better.

Fine with me.

> It would also make sense to have ol.ENABLE_NAMED_COLORS enabled by default for WEBGL users and allow to disable it as an optimization.

We can then get rid of ol.ENABLE_NAMED_COLORS, because we don't keep all the color names in the code any more.

gberaudo:
> We can then get rid of ol.ENABLE_NAMED_COLORS, because we don't keep all the color names in the code any more.

OK, it is fine with me if we remove it.

probins:
what's the status with this? As NAMED_COLORS is disabled by default, and AFAIK not documented anywhere, I doubt whether many people even know of its existence.

bjornharrtell:
I've rebased and made the changes so this PR simply removes `NAMED_COLORS` and any use of `goog.colors`.

I'm not sure how to complete the implementation suggestion using `window.getComputedStyle`, in particular to make it conditional in a way that does not slow or disrupt the Canvas renderer, so if it's still desirable to support named colors with the WebGL renderer that remains to be done.

fredj:
you can remove tests in `test/spec/ol/color.test.js`

gberaudo:
@bjornharrtell, if it is too complex to implement, then could you please update the `changelog/upgrade-notes.md` to state that using named color is no longer possible with the `WEBGL` renderer?

bjornharrtell:
Thanks @gberaudo and @probins. I've changed the upgrade notes to explain that named colors are no longer supported with WebGL. I've removed one test and adapted the other that tests the cache since it still seem relevant.

I took another look at the suggested method from @ahocevar and while it looks like a neat hack I still do not see how to properly integrate it.

ahocevar:
@bjornharrtell There were two typos in my [snippet](https://github.com/openlayers/ol3/pull/4194#issuecomment-146452741). I just updated it.

changelog/upgrade-notes.md
```diff
@@ -4,6 +4,10 @@
 
 This option is no longer needed, so it was removed from the API.
 
+#### Removal of `ol.ENABLE_NAMED_COLORS`
+
+This change only affects the WebGL renderer and allowed use of named colors for that renderer. Note that the default setting was false and that the Canvas and DOM renderers allows named colors regardless.
```
probins: ... renderers allow ... (just noticed this)

probins:
I'm a bit confused as to who should do what so this can be merged.

bjornharrtell:
As the WebGL renderer is progressing, I'd like to take a second try at implementing the proposal from @ahocevar.

bjornharrtell:
I've reimplemented so that named colors should be translated and cached. However I'm not sure how named colors are supported currently with `ol.ENABLE_NAMED_COLORS` at the default `false` and the canvas renderer - I suppose they don't need translation but I can't see the code path allowing that.

ahocevar:
Very nice, thanks @bjornharrtell. With such a small code snipped, it was definitely the right decision to get rid of `ol.ENABLE_NAMED_COLORS`.

probins:
@bjornharrtell see https://github.com/openlayers/ol3/pull/1985#issuecomment-50462922 `goog.color.names` are irrelevant for Canvas, which uses names like 'cyan' etc whether `ENABLE_NAMED_COLORS` is set or not

bjornharrtell:
Yeah I get that but I was worrying that my changes would mean that named colors now will be translated for the Canvas renderer regardless.

probins:
unnecessary added overhead, you mean. Also, is it necessary to create/add a new DOM element for every name? Would it not be better to have one named element and set the color?

ahocevar:
As far as I can tell, `ol.color.asArray` is only used by `ol.style.Icon` when the `color` option is set  and `ol.format.KML` when writing KML. I don't see any renderer related use.

ahocevar:
So the place to improve things is `ol.style.Icon`, where the color could be taken from the options as-is. Only the webgl renderer would have to call `ol.color.asArray`.

Any takers for a pull request with this improvement?

PR #4195, opened 2015-09-30
#4171: Prefer current layout as default on MultiLineString.setLineStrings()

### LINKED ISSUE #4171: Multipolygon.setPolygons() changes layout unintentionally
When calling Multipolygon.setPolygons() with an empty array (to clear it), the layout is set to XY, ignoring the current layout.
It is very likely a better approach to keep the current layout when calling with an empty array. Otherwise later calls to appendPolygon() are likely to fail.
BTW: The OL default layout used when parsing WFS features is XYZ, so switching to XY on clearing Multipolygon is especially bad in this cases.

awaterme:
Fixed #4171 for MultiLineStrings, too.


ahocevar:
Thanks!

PR #4198, opened 2015-09-30
Improve ol.control.Rotate.prototype.resetNorth_ performance

denilsonsa:
Changing. resetNorth_ function from O(n) to O(1).

To reproduce this issue, open the developer console and execute:

    map.getView().setRotation(5000000000);

Then click on the rotation button. Before the commit, it takes a few
seconds, after this commit it is instantaneous.

In addition, due to limited floating point precision, the previous code
calculates a different value from this one.

Note: I've written this pull request directly in the browser, but I tested it locally.


ahocevar:
Nice improvement, thanks!

PR #4202, opened 2015-10-01
Remove use of goog.dom.appendChild

fredj:
Use `parent.appendChild(child)` instead.

closure-library implementation: 
https://github.com/google/closure-library/blob/master/closure/goog/dom/dom.js#L1049-L1056


marcjansen:
Looks good to me, @fredj. Thanks and please merge.

PR #4212, opened 2015-10-02
Add a simple custom control example

yaugenka:
It is quite hard for users who are new to javascript to understand the <a href="http://openlayers.org/en/v3.9.0/examples/custom-controls.html">custom control example</a> which is based on class usage. So I suggest to add this simple one.

### DIFF
--- a/examples/simple-custom-control.css
+++ b/examples/simple-custom-control.css
@@ -1,0 +1,7 @@
+.rotate-north {
+  top: 65px;
+  left: .5em;
+}
+.ol-touch .rotate-north {
+  top: 80px;
+}


ahocevar:
I like the idea. Unfortunately the linter is not happy with your changes. If you check above why the Travis CI build failed, you'll see
```
Running gjslint...
----- FILE  :  /home/travis/build/openlayers/ol3/examples/simple-custom-control.js -----
Line 13, E:0220: No docs found for member 'button.innerHTML'
Line 16, E:0006: Wrong indentation: expected any of {2, 26} but got 4
Line 22, E:0220: No docs found for member 'element.className'
Line 26, E:0006: Wrong indentation: expected any of {2, 27} but got 4
Found 4 errors, including 2 new errors, in 1 files (486 files OK).
```
Are you able to address these problems? If not let me know, then I'll try to fix them before I merge it in. But that might take a while.

yaugenka:
@ahocevar please look into these issues because I am not experienced enough yet to address them myself.

tschaub:

PR #4216, opened 2015-10-04
Fix canvas replay which broke with 31a68e2

ahocevar:
See https://github.com/openlayers/ol3/pull/4192#commitcomment-13582218


marcjansen:
Nice catch, Andreas. Thanks for fixing this. Please merge.

tschaub:
For anybody else reading along at home and looking for something to do, we could use a test that renders labels (using the Canvas renderer).

marcjansen:
A visual test, or not?

tschaub:
> A visual test, or not?

Even basic unit tests would let us refactor with impunity.  I know there was discussion about this ad nauseum in Schladming, but the alternative is that we let our users discover regressions.

PR #4217, opened 2015-10-04
Replace ZeroClipboard with Clipboard.js to remove flash dependency

ThomasG77:
Due to the release of [Clipboard.js](http://zenorocha.github.io/clipboard.js/), I want to remove Flash dependency for clipping content from samples.


elemoine:
+1 with the idea. @ahocevar, what do you think?

ahocevar:
+1. Great to be able to move away from Flash.

ThomasG77:
Any opinion on my code in the PR as it seems to be OK on the principle?

ahocevar:
Code looks great, I'm going to merge this :-). Thanks @ThomasG77.

ahocevar:
Sorry, I have to revert this. It does not work. At least not on Safari on Mac. The demo on http://zenorocha.github.io/clipboard.js/ displays a "Press Cmd-C" tooltip, which isn't shown in the examples.

zenorocha:
Hey @ahocevar, the message you will display isn't included it on the library but you can easily copy our site's implementation https://github.com/zenorocha/clipboard.js/blob/gh-pages/assets/scripts/tooltips.js#L15-L31

PR #4219, opened 2015-10-05
Support tiled vector data and Mapbox vector tiles

ahocevar:
This pull request adds support for rendering tiled vector data. The key elements to achieve this are

1. `ol.layer.VectorTile`. A simple layer to hint renderers that a tiled renderer is to be used, instead of a viewport renderer like for `ol.layer.Vector`.
1. `ol.source.VectorTile` and `ol.VectorTile`. Features are handled by the tile, not the source. Like image tiles, vector tiles can be cached. The existing `ol.source.TileVector` is removed, and its functionality is integrated into `ol.source.VectorTile`. To improve the inheritance chain, `ol.source.UrlTile` was added. Both `ol.source.ImageTile` and `ol.source.VectorTile` inherit common functionality related to url loading from that new class.
1. `ol.renderer.canvas.VectorTileLayer`. This renderer gets tiles from `ol.source.VectorTile`, creates a replay group for each tile, and stores the replay group on the tile for rendering. Tiles are clipped at the tile boundary, so stroked polygons won't have a stroke at the tile boundary when a buffer is used on the server that creates the vector tiles.

Support for Mapbox vector tiles is provided by the new `ol.format.MVT` format, which relies on https://www.npmjs.com/package/pbf to read the protobuffers and https://www.npmjs.com/package/vector-tile to parse layers and features. To support pixel relative coordinates,`ol.proj.Units.TILE_PIXELS` was introduced. For efficient feature and geometry creation, a lightweight, read-only `ol.render.Feature` was added, which does not register any event listeners. Another performance improvement was achieved by skipping lineTo/moveTo calls for repeated coordinates when replaying vector batches at a lower resolution than it was created for.


examples/mapbox-vector-tiles-simple.js
```diff
@@ -0,0 +1,348 @@
...
+goog.require('ol.layer.VectorTile');
+goog.require('ol.source.VectorTile');
+goog.require('ol.style.Fill');
+goog.require('ol.style.Icon');
+goog.require('ol.style.Stroke');
+goog.require('ol.style.Style');
+goog.require('ol.style.Text');
+
+
+// Mapbox access token - request your own at http://mabobox.com
```
jonataswalker: A little typo here - http://mapbox.com

examples/mapbox-vector-tiles.js
```diff
@@ -0,0 +1,378 @@
...
+goog.require('ol.source.VectorTile');
+goog.require('ol.style.Fill');
+goog.require('ol.style.Icon');
+goog.require('ol.style.Stroke');
+goog.require('ol.style.Style');
+goog.require('ol.style.Text');
+goog.require('ol.tilegrid.TileGrid');
+
+
+// Mapbox access token - request your own at http://mabobox.com
```
jonataswalker: Same here - mapbox.com

src/ol/render/renderfeature.js
```diff
@@ -0,0 +1,156 @@
...
+goog.require('goog.asserts');
+goog.require('goog.functions');
+goog.require('ol.extent');
+goog.require('ol.geom.GeometryType');
+
+
+
+/**
+ * Lightweight, read-only, {@link ol.Feature} and {@link ol.geom.Geometry} like
+ * structure, optimized for rendering and styling. Geometry acces through the
```
jonataswalker: Little typo: "access"

marcjansen:
Awesome work. Very impressive!

package.json
```diff
@@ -37,9 +37,11 @@
     "metalsmith": "1.6.0",
     "metalsmith-templates": "0.7.0",
     "nomnom": "1.8.0",
+    "pbf": "1.3.4",
```
fredj: minor but version 1.3.5 was release yesterday

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -0,0 +1,447 @@
...
+      this.dirty_ = this.dirty_ || dirty;
+      replayState.dirty = replayState.dirty || dirty;
+    }
+  }
+
+  var features = tile.getFeatures();
+  if (renderOrder && renderOrder !== replayState.renderedRenderOrder) {
+    goog.array.sort(features, renderOrder);
+  }
+  goog.array.forEach(features, renderFeature, this);
```
fredj: `features.forEach(renderFeature, this)` could be used instead of `goog.array.forEach`

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -0,0 +1,447 @@
...
+        }
+      }
+
+    }
+  }
+
+  this.dirty_ = false;
+
+  /** @type {Array.<number>} */
+  var zs = goog.array.map(goog.object.getKeys(tilesToDrawByZ), Number);
```
fredj: `Object.keys(tilesToDrawByZ).map(Number)`

src/ol/interaction/selectinteraction.js
```diff
@@ -32,9 +32,11 @@ ol.interaction.SelectEventType = {
 
 
 /**
- * A function that takes an {@link ol.Feature} and an {@link ol.layer.Layer}
- * and returns `true` if the feature may be selected or `false` otherwise.
- * @typedef {function(ol.Feature, ol.layer.Layer): boolean}
+ * A function that takes an {@link ol.Feature} or {@link ol.Feature} and an
```
marcjansen: Typo:

```
…takes an {@link ol.Feature} or {@link ol.render.Feature}…
```

ahocevar:
I just moved the huge style function for the new examples into a gist, so this pull request is now smaller and should be easier to review.

ahocevar:
I have reworked the whole commit history to allow for sensible commit-by-commit reviewing. Thanks for any review.

topwood:
great! I just want to work on mapbox mvt tile on openlayers!

topwood:
when this can be merged to the master branch? thanks!

ahocevar:
@topwood This pull request needs to be reviewed by another core developer before it can be merged.

s093294:
Thumbs up on this. Nice work

topwood:
i have tried this branch, and when i have a mapbox mvt mbtiles file, I need a new style function for this file.right?

ahocevar:
@topwood Right. There is no support yet for the [Mapbox GL style spec](https://www.mapbox.com/mapbox-gl-style-spec/), so for now you have to translate the styles to an OpenLayers style function manually, like [I did](https://github.com/ahocevar/ol3/blob/vectortile/examples/resources/mapbox-streets-v6-style.js) for the examples.

topwood:
@ahocevar  thank you for your answering.

changelog/upgrade-notes.md
```diff
@@ -31,6 +31,10 @@ but with additional css:
 }
 ```
 
+#### Removal of `ol.source.TileVector`
+
+The experimental `ol.source.TileVector` class has been removed. Its functionality is now provided by `ol.source.VectorTile`. The API is the same, except for the `tileLoadFunction`, which is no longer an `ol.TileVectorLoadFunctionType`, but an `ol.TileLoadFunctionType`.
+
```
bartvde: Can you explain why, if the API is largely the same, the name was changed? What is against keeping the name as it was?

examples/mapbox-vector-tiles-simple.html
```diff
@@ -0,0 +1,17 @@
+---
+template: example.html
+title: Simple Mapbox vector tiles example
```
bartvde: This is very minor, but by reading this I don't understand off-hand why it's called simple. That would require me to look at the "normal" mapbox-vector-tiles example. I'd call this one simply mapbox-vector-tiles and rename the other one to mapbox-vector-tiles-advanced instead. But as I said, very minor and opinionated.

examples/mapbox-vector-tiles.html
```diff
@@ -0,0 +1,17 @@
+---
+template: example.html
+title: Mapbox vector tiles example
+shortdesc: Example of a Mapbox vector tiles map.
+docs: >
+  A vector tiles map which reuses the same tiles for subsequent zoom levels to save bandwith on mobile devices. **Note**: No map will be visible when the access token has expired.
+tags: "simple, mapbox, vector, tiles"
```
bartvde: simple needs to be removed here right as a tag?

changelog/upgrade-notes.md
ahocevar: ol.source.TileVector was a vector source with some tile functionality on top. ol.source.VectorTile is a tile source with some vector functionality on top, and works with ol.layer.VectorTile. The removal of the former came quite late in the development process, when I decided that only 2 methods needed to be added to make ol.source.VectorTile work for the existing examples.

Do you think it would make sense to change the name back? And maybe also rename ol.layer.VectorTile to ol.layer.TileVector?

examples/mapbox-vector-tiles-simple.html
ahocevar: Makes sense. I was wondering about the naming myself.

changelog/upgrade-notes.md
bartvde: I'd say if there is no obvious reason for changing the name, I'd keep it as is.

externs/olx.js
```diff
@@ -3547,7 +3598,7 @@ olx.layer.VectorOptions.prototype.renderBuffer;
 
 /**
  * Source.
- * @type {ol.source.Vector}
+ * @type {ol.source.Vector|ol.source.VectorTile}
```
bartvde: so there is a possible use case for combining ol.layer.Vector and ol.source.VectorTile?

ahocevar: I personally don't think there is one, but there are the existing TileVector examples. I don't know if anyone really uses that. I'd be more than happy to remove it, along with the existing TileVector examples (these could be changed to use ol.layer.Vector and a tile loading strategy). Removal would also make ol.source.VectorTile simpler, because the compatibility methods for ol.source.TileVector could be removed. See 1409cb20dbae9e182f1c8eaf77f9515f9c59dfb5.

changelog/upgrade-notes.md
s093294: One obvious reason could be that everywhere else (mapbox and related) has named it VectorTiles and I could see it make sense that a Tile of VectorTiles become a VectorTile. 

I am mostly thinking that new developers will google for VectorTile(s) and not TileVector when looking for related resources and I have no idea if google is smart enough to see that its the same thing. 

I also see why one would keep the original name if its been exposed to many of the current developers. I only speak as an outsider/consumer of ol3 here, so use the input if it's relevant.

externs/olx.js
bartvde: I think removal makes sense (and adapting the existing example to what you propose), it will only confuse users otherwise on what to use.

changelog/upgrade-notes.md
bartvde: Searching could be solved probably by putting vectortiles somewhere in the example tags? Or mentioning it in the API docs explicitly.

ahocevar: Since we just agreed on the removal of compatibility with legacy ol.source.TileVector, I'd say it makes sense to stick with what's in the pull request - ol.VectorTile, ol.source.VectorTile, ol.layer.VectorTile.

bartvde: That's true, fine with me.

externs/olx.js
```diff
@@ -3591,6 +3642,133 @@ olx.layer.VectorOptions.prototype.visible;
...
+ * @type {number|undefined}
+ * @api
+ */
+olx.layer.VectorTileOptions.prototype.renderBuffer;
+
+
+/**
+ * Render order. Function to be used when sorting features before rendering. By
+ * default features are drawn in the order that they are created. Use `null` to
+ * avoid the sort, but get an undefined draw order.
```
bartvde: This paragraph is a bit confusing to me. So when using undefined, a sort is applied (what type of sort)? Can you clarify the difference between using null and undefined more clearly here maybe?

externs/olx.js
```diff
@@ -3591,6 +3642,133 @@ olx.layer.VectorOptions.prototype.visible;
...
+olx.layer.VectorTileOptions.prototype.map;
+
+
+/**
+ * The bounding extent for layer rendering.  The layer will not be rendered
+ * outside of this extent.
+ * @type {ol.Extent|undefined}
+ * @api
+ */
+olx.layer.VectorTileOptions.prototype.extent;
```
bartvde: extent is not mentioned in the typedef itself

externs/olx.js
```diff
@@ -3591,6 +3642,133 @@ olx.layer.VectorOptions.prototype.visible;
...
+ * Opacity. 0-1. Default is `1`.
+ * @type {number|undefined}
+ * @api
+ */
+olx.layer.VectorTileOptions.prototype.opacity;
+
+
+/**
+ * Source.
+ * @type {ol.source.VectorTile}
```
bartvde: type in typedef includes ```|undefined```

externs/olx.js
```diff
@@ -3591,6 +3642,133 @@ olx.layer.VectorOptions.prototype.visible;
...
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.layer.VectorTileOptions.prototype.updateWhileInteracting;
+
+
+/**
+ * Visibility. Default is `true` (visible).
+ * @type {boolean|undefined}
+ * @api stable
```
bartvde: I guess this is marked stable since it's stable in ```olx.layer.BaseOptions``` ? But shouldn't the same apply to opacity, extent, minResolution and maxResolution then?

bartvde: On second thought, it's probably not possible to have stable properties inside of an unstable typedef, so this stable is likely a copy-paste left-over?

externs/olx.js
ahocevar: I just copied this from `olx.layer.VectorOptions.prototype.renderOrder`. Do you have a suggestion for rewording it (in both places)?

bartvde: Looking at the code, it seems there is no difference between undefined / null ?

```
  var renderOrder = layer.getRenderOrder();
  if (renderOrder === undefined) {
    renderOrder = null;
  }
```

bartvde: My confusion comes from the last 2 sentences, which seem to indicate there is a difference between using null and undefined, but looking at the code it seems there is no difference between the two at all.

externs/olx.js
```diff
@@ -3893,6 +4073,24 @@ olx.source.TileImageOptions.prototype.tileUrlFunction;
 
 
 /**
+ * URL template. Must include `{x}`, `{y}` or `{-y}`, and `{z}` placeholders.
+ * A `{?-?}` template pattern, for example `subdomain{a-f}.domain.com`, may be
+ * used instead of defining each one separately in the `urls` option.
+ * @type {string|undefined}
+ * @api stable
```
bartvde: Should this be stable or should we start without?

externs/olx.js
```diff
@@ -3905,33 +4103,41 @@ olx.source.TileImageOptions.prototype.wrapX;
...
-olx.source.TileVectorOptions.prototype.attributions;
+olx.source.VectorTileOptions.prototype.attributions;
 
 
 /**
- * Format. Required unless tileLoadFunction is used.
+ * Feature format for tiles. Used and required by the default
+ * `tileLoadFunction`.
  * @type {ol.format.Feature|undefined}
- * @api
```
bartvde: this still needs ```@api``` right?

externs/olx.js
```diff
@@ -3939,42 +4145,86 @@ olx.source.TileVectorOptions.prototype.format;
...
+olx.source.VectorTileOptions.prototype.tileUrlFunction;
 
 
 /**
  * URL template. Must include `{x}`, `{y}` or `{-y}`, and `{z}` placeholders.
+ * A `{?-?}` template pattern, for example `subdomain{a-f}.domain.com`, may be
+ * used instead of defining each one separately in the `urls` option.
  * @type {string|undefined}
- * @api
+ * @api stable
```
bartvde: was this changed to stable intentionally?

src/ol/format/format.js
```diff
@@ -5,6 +5,7 @@ goog.provide('ol.format.FormatType');
  * @enum {string}
  */
 ol.format.FormatType = {
+  ARRAY_BUFFER: 'arraybuffer',
```
bartvde: Is there no case to make for a base class? I.e. ```ol.format.ArrayBufferFeature``` ?

externs/olx.js
ahocevar: Thanks for spotting this. That is different from the way it is handled for ol.source.Vector, because there the render order depends on the way the RTree was populated. So I can remove the `null` part and simplify the documentation.

externs/olx.js
ahocevar: No you're right, the stable is missing for those.. We have the same in other layer type's typedefs.

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+
+  var options = opt_options ? opt_options : {};
+
+  /**
+   * @type {ol.proj.Projection}
+   */
+  this.defaultDataProjection = new ol.proj.Projection({
+    code: 'EPSG:3857',
+    units: ol.proj.Units.TILE_PIXELS
+  });
```
bartvde: This looks a bit weird to me at first sight, since it uses a common EPSG code with non-default units for it. Shouldn't this be more like the invented EPSG code that GeoServer uses for pixels?

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+   */
+  this.layerName_ = options.layerName ? options.layerName : 'layer';
+
+  /**
+   * @private
+   * @type {Array.<string>}
+   */
+  this.layers_ = options.layers ? options.layers : null;
+
+  this.rightHandedPolygons = true;
```
bartvde: in GeoJSON this is called ```rightHanded``` instead? Should we not have the same?

externs/olx.js
ahocevar: Yes. I moved the property from some subclasses that had it stable to the superclass. So I guess it should be stable for all subclasses.

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+  return ol.format.FormatType.ARRAY_BUFFER;
+};
+
+
+/**
+ * @private
+ * @param {Object} rawFeature Raw Mapbox feature.
+ * @param {string} layer Layer.
+ * @param {olx.format.ReadOptions=} opt_options Read options.
+ * @return {ol.Feature} Feature.
```
bartvde: can be ```ol.render.Feature``` as well for the return type?

src/ol/format/format.js
ahocevar: We had that and removed it. I think as long as ol.format.MVT is the only format that uses it, it makes no sense to have a base class. Just makes the code more complicated than it needs to be.

src/ol/format/mvtformat.js
bartvde: Ah no I see it can't since in that case this function will not get called.

src/ol/format/mvtformat.js
ahocevar: The tiles are projected to EPSG:3857, but use tile relative pixel coordinates. It's like a png image - pixels start in the top left corner at 0,0 and go to 256,256 for a 256x256 pixel tile.

I went back and forth on this, but ended up being quite happy with this solution. You cannot use such coordinates without knowing the tile coordinate. And in https://github.com/openlayers/ol3/pull/4219/files#diff-38eabaf9d6a74ee901d29683dac4163dR255 I assert that the units are not TILE_PIXELS, to make sure that such coordinates are not transformed.

src/ol/format/mvtformat.js
ahocevar: I think I had this at some point, but decided that extra verbosity cannot hurt. Do you want me to change it to `rightHanded`?

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+    geometryType = coords.length === 1 ?
+        ol.geom.GeometryType.POINT : ol.geom.GeometryType.MULTI_POINT;
+  } else if (type === 2) {
+    if (coords.length === 1) {
+      geometryType = ol.geom.GeometryType.LINE_STRING;
+    } else {
+      geometryType = ol.geom.GeometryType.MULTI_LINE_STRING;
+    }
+  } else {
+    geometryType = ol.geom.GeometryType.POLYGON;
```
bartvde: so there are no multi polygons in MVT?

bartvde: Should you handle type === 3 here as well (unknown)?

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  var properties = rawFeature.properties;
+  properties[this.layerName_] = layer;
```
bartvde: there is a slight chance here I guess that you'd overwrite an existing attribute with the same name?

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+
+  return new this.featureClass_(
+      geometryType, flatCoordinates, ends, rawFeature.properties);
```
bartvde: I'd use the local var properties here (I know it's the same object though).

src/ol/format/mvtformat.js
ahocevar: Yes, this is why it's configurable.

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+  var features = [];
+  var featureClass = this.featureClass_;
+  var layer, feature;
+  for (var name in tile.layers) {
+    if (layers && layers.indexOf(name) == -1) {
+      continue;
+    }
+    layer = tile.layers[name];
+
+    for (var i = 0, ii = layer.length; i < layer.length; ++i) {
```
bartvde: i < ii instead

src/ol/format/mvtformat.js
ahocevar: MultiPolygons are not in the spec. And good call, I'll explicitly check for `type === 3` and not just assume a polygon when it's not 1 or 2.

src/ol/format/mvtformat.js
ahocevar: Where exactly? Instead of `this.featureClass_`?

bartvde: No instead of ```rawFeature.properties``` very minor though.

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,262 @@
...
+    line = coords[i];
+    for (var j = 0, jj = line.length; j < jj; ++j) {
+      coord = line[j];
+      // Non-tilespace coords can be calculated here when a TileGrid and
+      // TileCoord are known.
+      flatCoordinates.push(coord.x, coord.y);
+    }
+    end += 2 * j;
+    ends.push(end);
+  }
```
bartvde: this code block is duplicated in this file

bartvde: Ah sorry I see now that ol.render.Feature does not take a geom instance. But still quite a bit of code should be reusable here.

ahocevar: It calculates flatCoordinates and ends, so I did not know what to return from a shared function. I'll see if I find a way to simplify this.

ahocevar: Maybe pass `flatCoordinates` as argument and return `ends` to a shared function? Weird signature, but then it can be reused.

src/ol/geom/geometry.js
```diff
@@ -250,6 +252,9 @@ ol.geom.Geometry.prototype.translate = goog.abstractMethod;
  * @api stable
  */
 ol.geom.Geometry.prototype.transform = function(source, destination) {
+  goog.asserts.assert(
+      ol.proj.get(source).getUnits() !== ol.proj.Units.TILE_PIXELS,
+      'cannot transform geometries with TILE_PIXELS units');
```
bartvde: I assume the opposite is also true, we cannot transform to a destination with TILE_PIXELS units?

src/ol/proj/proj.js
```diff
@@ -675,7 +677,8 @@ ol.proj.get = function(projectionLike) {
 ol.proj.equivalent = function(projection1, projection2) {
   if (projection1 === projection2) {
     return true;
-  } else if (projection1.getCode() === projection2.getCode()) {
+  } else if (projection1.getCode() === projection2.getCode() &&
+      projection1.getUnits() === projection2.getUnits()) {
     return true;
   } else if (projection1.getUnits() != projection2.getUnits()) {
```
bartvde: we now have this check twice, so we can reuse it in a local var:

```
var unitsEqual = (projection1.getUnits() === projection2.getUnits());
```

src/ol/render/canvas/canvasreplay.js
```diff
@@ -1158,6 +1180,8 @@ ol.render.canvas.PolygonReplay = function(tolerance, maxExtent, resolution) {
     miterLimit: undefined
   };
 
+  this.rightHanded_ = false;
```
bartvde: no type docs?

src/ol/render/canvas/canvasreplay.js
```diff
@@ -1376,6 +1402,16 @@ ol.render.canvas.PolygonReplay.prototype.getBufferedMaxExtent = function() {
 
 
 /**
+ * @param {boolean} rightHanded Whether polygons are assumed to follow
+ *     the right-hand rule.
+ */
+ol.render.canvas.PolygonReplay.prototype.setRightHanded =
```
bartvde: There is no way to keep this out of the renderer code? The overhead of "transforming" them at the feature / geom level is too much or some other reason? I.e. always have our geometries be non right-handed internally?

src/ol/render/renderfeature.js
```diff
@@ -0,0 +1,156 @@
...
+};
+
+
+/**
+ * Get the extent of this feature's geometry.
+ * @return {ol.Extent} Extent.
+ * @api
+ */
+ol.render.Feature.prototype.getExtent = function() {
+  if (!goog.isDef(this.extent_)) {
```
bartvde: could we take this from the MVT structure instead and set it upon construction? Since they have the extent? Or that doesn't make sense?

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -0,0 +1,446 @@
...
+  var rotation = viewState.rotation;
+  var source = this.getLayer().getSource();
+  goog.asserts.assertInstanceof(source, ol.source.VectorTile,
+      'Source is an ol.source.VectorTile');
+
+  var transform = this.getTransform(frameState, 0);
+
+  this.dispatchPreComposeEvent(context, frameState, transform);
+
+  var layer = this.getLayer();
```
bartvde: minor but we could move this up to line 89 so that:

```var source = layer.getSource();```

File: src/ol/source/urltilesource.js [hunk 34]
```diff
@@ -0,0 +1,209 @@
...
+ * @return {ol.TileUrlFunctionType} TileUrlFunction
+ * @api
+ */
+ol.source.UrlTile.prototype.getTileUrlFunction = function() {
+  return this.tileUrlFunction;
+};
+
+
+/**
+ * Return the URLs used for this XYZ source.
```
bartvde: not always XYZ right?

src/ol/vectortile.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param {ol.TileState} tileState Tile state.
+ */
+ol.VectorTile.prototype.setState = function(tileState) {
+  this.state = tileState;
+  this.changed();
+};
+
+
+/**
+ * Set the feeature loader for reading this tile's features.
```
bartvde: feeature -> feature

bartvde:
hey @ahocevar I'm done with my review, impressive work. Any chance of adding a rendering test case for this?

src/ol/format/format.js
bartvde: Okay, no objections from my side.

src/ol/format/mvtformat.js
bartvde: Thanks for the detailed explanation, I think it's fine to leave it as is then.

src/ol/format/mvtformat.js
bartvde: Let's leave it like it is.

src/ol/render/canvas/canvasreplay.js
ahocevar: Geometries can come from many places. We could transform feature geometries when we add features to the source. But we cannot enforce that when users modify geometries, use a replay in a postrender event, etc.

This has always been in the renderer code: `getOrientedFlatCoordinates` is used, and the oriented coordinates are cached.

src/ol/render/renderfeature.js
ahocevar: The `extent` in the MVT structure you are talking about is what we call `tileSize`. There is also a `bbox()` method, but it would be expensive to calculate that for every feature and it's not needed for rendering.

clint-tseng:
Not a code review, but I'm using this branch rebased against master right now and it's working beautifully with my non-MapBox VectorTile PBF server source. Thanks for the amazing work! :beers:

ahocevar:
@bartvde I have addressed all your comments and answered your questions. Thanks for this great review!

s093294:
Will coordiate transform work? 

I was loading geosjon vector tiles before with the following hack to support a map projection of epsg25832 but the vector tiles was generated using mapbox tools that always use 3857.

Would be nice If I could start using your implementation for the same usecase.

```
//VectorTiles Loader
                    var vt = new ol.source.TileVector({
                        projection: epsg25832, //This is not used for TileVector
                        tileGrid: tg,
                        tileUrlFunction: (tileCoord: number[], ratio, proj) => {
                            return tileCoord.join(",");
                        },
                        tileLoadFunction: (urlParamsStr: string, loaded) => {
                            //Consider if a mapbox rendere is better, currently we fetch tiles, convert coordinates and then visualize in canvas. 
                            //Data is optimized to be rended on a per tile case. 
                            var tileCoord = urlParamsStr.split(",").map((s) => parseInt(s));
                            var extent = getTileCoordExtent(tg, tileCoord, epsg25832, "EPSG:3857");
                            var resolution = tg.getResolution(tileCoord[0]);
                            var newZ = linearFindNearest(tilegrid.getResolutions(), resolution, 1);
                            var newTileRange = getTileRangeForExtentAndResolution(tilegrid, extent, tilegrid.getResolution(newZ));
                            var defers = 0;
                            var features = [];
                            var localCache = {};


                            var tileLoader = ((tz, tx, ty) => {
                                setTimeout(() => {
                                    var key = "" + tz + tx + ty;

                                    if (!tileCache[key]) {
                                        tileCache[key] = [];
                                        localCache[key] = true;
                                        client.getData(this.options.document.workSet, this.options.document.name + "/vt/" + tz + "/" + tx + "/" + ty + ".geojson")
                                            .done(result => {
                                                //If data is returned as string make sure its json.
                                                var data: any[] = typeof (result) === "string" ? JSON.parse(result) : result;

                                                for (var i = 0; i < data.length; i++) {
                                                    var obj = data[i].tags;
                                                    
                                                    obj.geometry = new ol.geom.LineString(data[i].geometry[0].map((coord: number[]) => {
                                                        //Convert from vector tiles local tile coordinates (mercator) 
                                                        var xpx = (coord[0] / 4096 + (tx)) / (1 << (tz));
                                                        var ypx = (coord[1] / 4096 + (ty)) / (1 << (tz));
                                                        var p = [((xpx - 0.5) * 2) * 6378137 * Math.PI, -((ypx - 0.5) * 2) * 6378137 * Math.PI];
                                                        //Convert to local coordinate system.
                                                        var p1 = ol.proj.transform(p, "EPSG:3857", epsg25832).concat(coord.splice(2));
                                                        return p1;
                                                    }));
                                                    var feature = new ol.Feature(obj);
                                                    feature.setId(obj.id); //Set the ID, not used yet, but need to look into if features with same id that are split across tiles can/should be combined.
                                                    feature.set("tile_sources", []); 
                                                    features.push(feature);
                                                    if (this.tracksVectorSource.getFeatureById(obj.id)) {
                                                        var oldFeature = this.tracksVectorSource.getFeatureById(obj.id);
                                                        oldFeature.get("tile_sources").push([tz, tx, ty]);
                                                        var geom = oldFeature.getGeometry<ol.geom.LineString>();
                                                        var newPoints = feature.getGeometry<ol.geom.LineString>().getCoordinates();
                                                        var oldPoints = geom.getCoordinates();
                                                       
                                                        var updatedlist = [];
                                                        var ai = oldPoints[0][2] === newPoints[0][2] ? 0:1; var bi = 0;
                                                                                                               
                                                        var aiend = newPoints.length - ( oldPoints[oldPoints.length - 1][2] === newPoints[newPoints.length - 1][2] ? 0:1);                                                   
                                                        while (ai < aiend || bi < oldPoints.length) {
                                                            if (ai === aiend) {
                                                                updatedlist.push(oldPoints[bi]); bi++;
                                                            } else if (bi === oldPoints.length) {
                                                                updatedlist.push(newPoints[ai]); ai++;
                                                            } else if (oldPoints[bi][2] === newPoints[ai][2]) {
                                                               
                                                                if (oldPoints[bi][0] !== newPoints[ai][0]) {
                                                                    console.log([bi,ai,oldPoints[bi], newPoints[ai]]);
                                                                }
                                                            
                                                                updatedlist.push(newPoints[ai]); bi++; ai++; 
                                                                
                                                            } else if (oldPoints[bi][2] < newPoints[ai][2]) {
                                                                updatedlist.push(oldPoints[bi]); bi++;
                                                            } else {
                                                                updatedlist.push(newPoints[ai]); ai++;
                                                            }
                                                        }
                                                    
                                                        geom.setCoordinates(updatedlist, 'XYM');
                                                    } else {
                                                        this.tracksVectorSource.addFeature(feature);
                                                    }// Have to add it to get support for getFeaturesInExtent
                                                    tileCache[key].push(feature);
                                                }

                                            }).always(() => {
                                                if (--defers === 0) {
                                                    loaded(features);
                                                }
                                            });
                                    } else {

                                        if (!localCache[key]) {
                                            localCache[key] = true;
                                            features.push.apply(features, tileCache[key]);
                                            if (--defers === 0) {
                                                loaded(features);
                                            }
                                        } else {
                                            if (--defers === 0) {
                                                loaded(features);
                                            }
                                        }
                                    }
                                });
                            });


                            for (var ix = newTileRange[0]; ix <= newTileRange[1]; ++ix) {
                                for (var iy = newTileRange[2]; iy <= newTileRange[3]; ++iy) {
                                    defers++;

                                    tileLoader(newZ, ix, -iy - 1);
                                }

                            }
                            
                        }
                    });
```

ahocevar:
@s093294:

> Will coordiate transform work?

It will be possible to make them work for ol.source.VectorTile, but it is not implemented yet. With raster reprojection in place now, a logic to find the correct tiles at the correct resolution, like you have in your hack, is already in the library. What's missing is the reprojection of features in ol.VectorTile and the tile boundary clip lines.

changelog/upgrade-notes.md
```diff
@@ -31,6 +31,12 @@ but with additional css:
 }
 ```
 
+#### Removal of `ol.source.TileVector`
+
+With the introduction of true vector tile support, `ol.source.TileVector` becomes obsolete. To continue rendering tiled vector data with a vector layer, use `ol.source.Vector` with an `ol.loadingstrategy.tile` strategy and `uniqueFeatureIds` set to `false` instead. You also have to clear the source when the map resolution changes. See the [tile-vector example](http://openlayers.org/en/v3.11.0/examples/tile-vector-html) for details.
+
```
bartvde: Still curious to hear your thoughts on when someone would actually do this, instead of using ol.layer.VectorTile and ol.source.VectorTile. Is this not so much of an edge case that it could be removed altogether? If I read these notes as an end user, I would not be able to make my choice.

s093294:
Cool. One more question that has been bothering me - not sure if its a good question though. But as in my hack I am merging features that has the same ID. Do this make sense - will your implementation support similar.

The reason why I did it was to push all the combined features to private tracksVectorSource = new ol.source.Vector({}); such i could query for the features.

examples/mapbox-vector-tiles-advanced.html
```diff
@@ -0,0 +1,17 @@
+---
+template: example.html
+title: Mapbox vector tiles example
```
bartvde: titles of both examples are the same, maybe add advanced here in the title?

examples/tile-vector.html
```diff
@@ -1,12 +1,16 @@
...
-title: Tile vector example
-shortdesc: Example of vector tiles from openstreetmap.us.
+title: Tiled vector data example
+shortdesc: Using tiled vector data from openstreetmap.us.
 docs: >
-  Example of vector tiles from openstreetmap.us.
-tags: "custom, control"
+  The vector data in this example is served from a static server. For this
+  purpose, vector features are cut into tiles. Feature geomertries are also
+  optiomized for each resolution. **Note: Usually you would use
```
bartvde: optiomized -> optimised

src/ol/render/canvas/canvasreplay.js
ahocevar: Hm. Now that I look at it again, the whole logic could actually be removed. Because ol.render.Feature will always assume righthanded anyway.

externs/olx.js
```diff
@@ -3591,6 +3642,133 @@ olx.layer.VectorOptions.prototype.visible;
...
+ * @type {number|undefined}
+ * @api stable
+ */
+olx.layer.VectorTileOptions.prototype.opacity;
+
+
+/**
+ * Source.
+ * @type {ol.source.VectorTile|undefined}
+ * @api
```
bartvde: can this be stable as well since it's stable in ```olx.layer.VectorOptions``` ?

externs/olx.js
```diff
@@ -3591,6 +3642,133 @@ olx.layer.VectorOptions.prototype.visible;
...
+ */
+olx.layer.VectorTileOptions.prototype.source;
+
+
+/**
+ * Layer style. See {@link ol.style} for default style which will be used if
+ * this is not defined.
+ * @type {ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined}
+ * @api
```
bartvde: same comments as above, stable in ```olx.layer.VectorOptions``` so could be stable here as well?

examples/mapbox-vector-tiles-advanced.js
```diff
@@ -0,0 +1,73 @@
...
+          html: '© <a href="https://www.mapbox.com/map-feedback/">Mapbox</a> ' +
+              '© <a href="http://www.openstreetmap.org/copyright">' +
+              'OpenStreetMap contributors</a>'
+        })],
+        format: new ol.format.MVT(),
+        tileGrid: new ol.tilegrid.TileGrid({
+          extent: ol.proj.get('EPSG:3857').getExtent(),
+          resolutions: resolutions
+        }),
+        tilePixelRatio: 16,
```
bartvde: wondering why 16 is not our default? Are there implementations out there that use something else than what MapBox does?

src/ol/featureloader.js
```diff
@@ -129,5 +171,5 @@ ol.featureloader.xhr = function(url, format) {
        */
       function(features) {
         this.addFeatures(features);
-      });
+      }, ol.nullFunction);
```
bartvde: should we add a FIXME here, since failure is not really handled in this case, but it also wasn't before this PR? But we did get rid of the FIXME in this PR.

src/ol/format/mvtformat.js
```diff
@@ -0,0 +1,259 @@
...
+ */
+ol.format.MVT.prototype.setLayers = function(layers) {
+  this.layers_ = layers;
+};
+
+
+/**
+ * @private
+ * @param {Object} coords Raw feature coordinates.
+ * @param {Array.<number>} flatCoordinates Flat coordiantes to be populated by
```
bartvde: coordinates -> coordinates

examples/mapbox-vector-tiles-advanced.js
ahocevar: If you use vector tiles from anything else than Mapbox, it's usually 1 (for GeoWebCache) or something else. So it does not make sense to use 16 as default.

ahocevar:
@s093294:

> in my hack I am merging features that has the same ID. Do this make sense - will your implementation support similar. The reason why I did it was to push all the combined features to private tracksVectorSource = new ol.source.Vector({}); such i could query for the features.

No. Vector tiles are for rendering, not for working querying features like you do with features you get e.g. from a WFS. If you want to do that, you should be using `ol.layer.Vector` and `ol.source.Vector`.

ahocevar:
@bartvde I think I have addressed all remaining comments. Please take another look to make sure.

bartvde:
LGTM @ahocevar please merge once Travis gives the final okay

ahocevar:
Thanks @bartvde for this thorough review - the code is much simpler, cleaner and smaller now.

ahocevar:
Some metrics regarding build size. Configuration used:
```json
  "exports": [
    "ol.Map", "ol.Map#*",
    "ol.View", "ol.View#*",
    "ol.layer.Vector", "ol.layer.Vector#*",
    "ol.source.Vector", "ol.source.Vector#*",
    "ol.format.GeoJSON",
    "ol.style.*"
  ],
  "compile": {
    "define": [
      "ol.ENABLE_VECTOR_TILE=false",
    ]
  }
}
```
So no vector tiles in this configuration.

Before:
```
uncompressed: 194232 bytes
  compressed: 63595 bytes
```
After:
```
uncompressed: 195050 bytes
  compressed: 63998 bytes
```
The increase of 818 bytes in code size comes mostly from performance optimisations in the canvas replay, and the support for array buffers in vector formats.

tonio:
Nice work!
Did you notice the gap between tiles in FF ?
Sounds like a rounding issue.

ahocevar:
@tonio I saw it only on Linux. Did you observe this on other platforms as well?

s093294:
How do you guys feel about the performance of http://openlayers.org/en/master/examples/mapbox-vector-tiles-advanced.html?q=vectortile using the canvas renderer? On my machine its suboptimale compared to the mapbox GL render. 

Dont get me wrong - its really amazing work.

bartvde:
Right I observed the same performance "issue", especially on pan.

bartvde:
@ahocevar @tonio I see it in FF on OSX as well, hadn't opened up the example in FF as yet to be honest

ahocevar:
@s093294 @bartvde In the process of working on this pull request, I spent a long time profiling and optimizing. The bottleneck is now in the renderer code itself. The majority of the time is spent applying strokes and fills. If the canvas renderer did a better job combining geometries and stroking/filling them at once, performance would be better. But that's a separate issue, as it affects all canvas rendering in OpenLayers. The goal is to reduce the number of stroke and fill instructions.

bartvde:
@ahocevar should we at least open up a new issue for this? So we don't loose your valuable information here?

ahocevar:
@bartvde Please do so.

tonio:
@ahocevar yep, I reproduce on win 7 & 10, too (browserstack).

ahocevar:
@tonio Please create a ticket for that.

tonio:
@ahocevar see #4329

ahocevar:
Thanks @tonio.

NathanW2:
@ahocevar is it possible to read tiles from a local folder and not from over the web e.g for offline use?

nicholas-l:
Hi @NathanW2 I don't think this is possible as there are security issues with accessing files using AJAX, see http://stackoverflow.com/a/18137280

ahocevar:
@NathanW2 You can define a custom `tileLoadFunction` on your `ol.source.VectorTile` if you want to use something else than AJAX to load your tile data.

kurundkar123:
Hello, Can anyone help me with how to create simple vector tile and how to add layers in it, how to add structures in it.

PR #4222, opened 2015-10-06
Remove unnecessary cast

fredj:
Remove unneeded cast found by the closure-compiler (`unnecessaryCasts` option).

### DIFF
--- a/src/ol/control/attributioncontrol.js
+++ b/src/ol/control/attributioncontrol.js
@@ -76,9 +76,9 @@ ol.control.Attribution = function(opt_options) {
    * @private
    * @type {Node}
    */
-  this.collapseLabel_ = /** @type {Node} */ (goog.isString(collapseLabel) ?
+  this.collapseLabel_ = goog.isString(collapseLabel) ?
       goog.dom.createDom(goog.dom.TagName.SPAN, {}, collapseLabel) :
-      collapseLabel);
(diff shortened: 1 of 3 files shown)


externs/olx.js
```diff
@@ -25,6 +25,14 @@ olx.AttributionOptions.prototype.html;
 
 
 /**
+ * Tile ranges.
+ * @type {Object.<string, Array.<ol.TileRange>>|undefined}
+ * @api
+ */
+olx.AttributionOptions.prototype.tileRanges;
```
ahocevar: No, we were discussing getting rid of this and changing the way attributions work. Should not be `@api`.

fredj: `@api` removed

ahocevar:
Please merge when Travis is happy, but add a ticket description to provide some context (like why these changes were made).

fredj:
`olx.AttributionOptions.prototype.tileRanges` was removed in #2509: it's probably not a good idea to add it again ...

ahocevar:
> olx.AttributionOptions.prototype.tileRanges was removed in #2509: it's probably not a good idea to add it again ...

Was going to suggest to not add it again, but didn't want to bother you too much. But now that you're at it already, please remove that commit from the pull request.

fredj:

PR #4223, opened 2015-10-06
Remove goog.fx.easing

fredj:


ahocevar:
Nice. Please merge when Travis says so.

bartvde:
@fredj we cannot simply copy code from an Apache 2.0 project and release it as BSD?

ahocevar:
@bartvde Have you looked at the change? It's trivial code.

bartvde:
@ahocevar yes, but it's still a copy action? Does the type of function you copy have any impact on the legal implications you think?

ahocevar:
I'm not a legal expert, but the change that @fredj made is a change from goog.fx function references to function with trivial math in it. He probably hasn't even copied it, but knew from math class that this is standard cubic easing, which can be found in scientific literature and e.g. [here](http://gizma.com/easing/). But it may make sense to change the commit message and mention that this is cubic easing.

fredj:
And I've found the same equations at http://joshondesign.com/2013/03/01/improvedEasingEquations (`easeInCubic` and `cubicOut`)

fredj:
Commit message changed

ahocevar:
> Has this already been discussed in general wrt getting rid of the Google closure library in ol3?

@bartvde, we won't copy code from Closure library directly. The biggest chunks are removal of goog.events, which we will replace with a custom simple events system (use browser events directly for MapBrowserEvent, and very simple events for anything else), and goog.Mat*, which we'll replace with a different 3rd party library that comes annotated for Closure compiler (like e.g. https://github.com/richardanaya/SylvesterClosure).

bartvde:
okay good to go then

fredj:
Thanks for the review

PR #4224, opened 2015-10-06
Don't transform the angle into degrees to rotate the element

fredj:
Use the radian value, see http://www.w3.org/TR/css3-values/#angle-value


elemoine:
+1 if it works in the browsers ol3 supports.

fredj:
it's working on IE9+, chrome and FF

elemoine:
Cool. Thanks for testing.

PR #4229, opened 2015-10-06
Avoid obtaining infinity with Math.tan function

rmourembles:
Check if coordinate value is in WORLD_EXTENT.
If yes, replace coordinate value to avoid obtaining the inifinite with Math.tan function.


ahocevar:
This is not a behaviour we want as default in the library, because it changes coordinates without notice. It is the responsibility of the app developer to avoid coordinate conversions outside the projection's validity extent.

If you want to help improve the behaviour of the library, you could add an assertion that warns the app developer when a coordinate conversion leads to Infinity results.

rmourembles:
Ok, thanks for your answer

marcjansen:
Plus 1 for the assertion. Will come in handy, I know.

tschaub:
I think it would be worth discussing this a bit more.  I think it would be pretty nice behavior to have `ol.proj.transform()` return finite coordinates where possible.  We could add documentation saying that transforms outside the validity extent cannot be considered reliable.

tschaub:

Issue #3990, opened 2015-08-11
If you use the requirejs. can't use proj4js. 


realrnjs:
proj4 is not defined

if (ol.ENABLE_PROJ4JS && typeof proj4 == 'function')

This condition is always false.

ahocevar:
Fixed in v3.8.

afernandezp:
I don't think it is fixed in v.3.8. Currently I'm working with OL3 V3.8.2 and I'm still getting that exception: "proj4 is not defined"

ahocevar:
You're right. The issue fixed with #3910 was for libraries that are baked into the library. Sorry for the confusion.

juristr:
Wouldn't it be enough to include it in the UMD at the top?? 

Something like

```javascript
(function(root, factory) {
  if (typeof exports === "object") {
    module.exports = factory(require('proj4');
  } else if (typeof define === "function" && define.amd) {
    define(['proj4'], factory);
  } else {
    root.ol = factory(root.proj4);
  }
}(this, function(proj4) {
   ...
});
```

ahocevar:
At the top of what?

juristr:
@ahocevar of the final build file that is being produced. Just download the latest release and check the `dist/ol.js` file.

ahocevar:
This should work, yes. Pull request welcome!

juristr:
@ahocevar Ok, I'm testing this out today on my local configuration and if everything works properly, I'll submit a PR.

fredj:
Fixed with #4609

PR #4230, opened 2015-10-06
fix UMD header to handle proj4

juristr:
The UMD header previously did not handle the inclusion of proj4 which had
to be exposed globally and thus didn't work in a UMD/AMD environment such as
when loading with RequireJS.

More in issue #3990 .

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -18,13 +18,13 @@ var root = path.join(__dirname, '..');
 
 var umdWrapper = '(function (root, factory) {\n' +
     '  if (typeof exports === "object") {\n' +
-    '    module.exports = factory();\n' +
+    '    module.exports = factory(require("proj4"));\n' +
     '  } else if (typeof define === "function" && define.amd) {\n' +
-    '    define([], factory);\n' +
+    '    define(["proj4"], factory);\n' +
     '  } else {\n' +
-    '    root.ol = factory();\n' +
+    '    root.ol = factory(root.proj4);\n' +
     '  }\n' +
-    '}(this, function () {\n' +
+    '}(this, function (proj4) {\n' +
     '  var OPENLAYERS = {};\n' +
     '  %output%\n' +
     '  return OPENLAYERS.ol;\n' +


juristr:
@ahocevar Here's the PR :smile:

ahocevar:
Thanks, this appears to work. Please submit a [CLA](https://github.com/openlayers/cla) so we can merge this.

juristr:
@ahocevar Done.

ahocevar:
Unfortunately this breaks the use of OpenLayers in an UMD/AMD environment WITHOUT proj4.js. So the fix needs more work.

juristr:
Hmm...I'm sorry about that. I'll give it a look tomorrow. Do you maybe already have a scenario that shows the break?

ahocevar:
@juristr Just add OpenLayers built from your branch to [this Hello World gist](https://gist.github.com/aj0strow/97f6ebb10e116713d825)'s `require-main.js`:
```js
require(
  [ 'helloworld', 'path/to/build/ol' ],

  function(helloworld, ol) {
    console.log(ol);
    document.write(helloworld);
  }
);
```
You'll get
 ```
[Error] Error: Script error for: proj4
http://requirejs.org/docs/errors.html#scripterror
```

**Comment by juristr** [comment]
@ahocevar I get the problem..hmm...

Now things get a lot more complex. What would have to be done is basically to optionally load the lib through requirejs (or any other AMD lib), silently ignoring any load failure. Solution approaches could be

- http://requirejs.org/docs/api.html#errbacks
- http://stackoverflow.com/a/19000137/50109

So the result would be something like

```javascript
(function(root, factory) {
  if (typeof exports === "object") {
    module.exports = factory(require('proj4'));
  } else if (typeof define === "function" && define.amd) {
    define([], function() {
      return require(['proj4'],
        function(proj4Module){
          factory(proj4Module);
        },
        function(err){
          var failedId = err.requireModules && err.requireModules[0];
          if (failedId === 'proj4') {
              requirejs.undef(failedId);

              factory();
          } else {
              //Some other error. Maybe show message to the user.
              console.error('Failed to load ' + failedId);
          }
        });
    });
  } else {
    root.ol = factory(root.proj4);
  }
}(this, function(proj4) {
   ...
});
```

The problem I'm having is that the result of the define now gets async...that's an issue. Do you have any ideas/hints on how to solve something like this?

ahocevar:
I think there is no good solution, other than using Browserify or Webpack instead of RequireJS. For those who really rely on a RequireJS, it will be best to include proj4.js with a separate script tag.

ahocevar:
Maybe it would make sense to add an `ol.proj.setProj4` function to tell OpenLayers where proj4.js is.

juristr:
@ahocevar Could you elaborate? You mean to set the location of proj4 and have it being loaded internally by ol.js?

ahocevar:
@juristr No, something like
```js
ol.proj.setProj4(require('proj4'));
```

juristr:
@ahocevar Hmm...ya. That could solve the problem of checking for the existence of the library. I'll try to verify this on my RequireJS setup and see whether that could help. Unfortunately that might take a bit longer as I'm off next week. I'll let u know.
Meanwhile, thx!

PR #4239, opened 2015-10-08
Changing highlighted to highlight

austinkeeley:
According to [the KML specifications](http://docs.opengeospatial.org/is/12-007r2/12-007r2.html#811), StyleMap key/value pairs should only have values `normal` and `highlight`. The test specs had some cases where it was looking for `highlighted`.


elemoine:
Looks good. Thanks.

Issue #2208, opened 2014-06-16
Support polygon fills


pagameba:
It might be nice to support polygon fills, which can be accomplished with the [canvas `createPattern`](http://www.html5canvastutorials.com/tutorials/html5-canvas-patterns-tutorial/) method.

MrPl0p:
Hello, 

What do you think of this ?

![image](https://cloud.githubusercontent.com/assets/11309450/6483923/dc8a8922-c276-11e4-859c-ea03e1a80fed.png)

Thank you

ahocevar:
Looks promising!

TheHeat:
Hi, I'd _love_ this feature. Looking good @MrPl0p

tremby:
Yes please.

MrPl0p:
Hello guys, 

Here's a white paper I wrote during my internship. 
http://www.quadratic.be/download/Advanced_styles_in_OpenLayers_3.pdf
I talk about OL3 styles research and development.
I hope it will be useful for the OL3 community, anyway I enjoyed to do it.

bartvde:
thanks for sharing @MrPl0p

BevolutionZ:
thanks for the document !

effetx:
Many many thanks @MrPl0p for the document. I read with attention what you wrote but I yet have some problems to "dress" the vector layes

  var image_repeat= new Image();
  image_repeat.src="graphicfill.png";

  var vectorlayer= new ol.layer.Vector({source: new ol.source.GeoJSON({projection: 'EPSG:3857',url: 'file.ge'}),opacity: 0.8,              
              style: (function() {          
              return function(feature, resolution) {
              var ruleStyle1 = [new ol.style.Style({fill: 
                new ol.style.Fill({
                  image:[image_repeat,0.4]
                }),
                stroke: new ol.style.Stroke({color:'#bbb',width:3})          
          })];
      
                        var attributo='singolo';
          if ((attributo === '')||(attributo == 'singolo')) {
            return ruleStyle1;}
            };
                            })()});
  map.addLayer(vectorlayer);
  
  
I hack the ol-debug.js file with the code proposed in your document (or I hope to having done this, but probability I make some mistakes) adding a row in ol.style.fill function and adding prototype.getImage function, in this way:

ol.style.Fill = function(opt_options) {
  var options = goog.isDef(opt_options) ? opt_options : {};
  this.color_ = goog.isDef(options.color) ? options.color : null;
[...]

PR #4241, opened 2015-10-09
Add generic Fill rendering as per #2208

RouquinBlanc:
Hi,

This adds the capability for an ol.style.Fill to render something else than a simple color.

Taking as a building block the MrPl0p and his nice https://github.com/MrPl0p/ol3-aswcr as discussed in  #2208, I extracted all the image processing and simply added a 'renderer' function, which can fill in anything.

For example, look at what it does if you want filling with stripes:
http://jsfiddle.net/dykko7mw/1/
screen shot 2015-10-09 at 09 37 41

The code would be:

```js
...
new ol.style.Fill({
    renderer: makeRedStripes
});

function makeRedStripes (context_) {
    var height = 100,
        width  = 56,
        x,
        ctx,
    $pattern = $('<canvas>')
        .attr('height', height)
        .attr('width', width)[0];

    ctx = $pattern.getContext('2d');
    ctx.strokeStyle = 'rgba(255, 0, 0, 0.15)';
    ctx.lineWidth = 3;

    x = 0;
    while(x < width * 2) {
        ctx.beginPath();
        ctx.moveTo(ctx.lineWidth + x, - 1);
        ctx.lineTo(ctx.lineWidth + x - 50, height + 1);
        ctx.stroke();
        x += 7;
    }
    
    return context_.createPattern($pattern, 'repeat');
}
...
```

Not sure if it can be integrated this way directly, but at least it gives you guys a base to work on.
Thanks for all the great work on OL3!

Cheers
Jonathan

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -435,12 +435,13 @@ ol.render.canvas.Replay.prototype.replay_ = function(
         ++i;
         break;
       case ol.render.canvas.Instruction.SET_FILL_STYLE:
-        goog.asserts.assert(goog.isString(instruction[1]) || goog.isFunction(instruction[1]),
+        goog.asserts.assert(goog.isString(instruction[1])
+            || goog.isFunction(instruction[1]),
             '2nd instruction should be a string or a function');
         if (goog.isString(instruction[1])) {
-            context.fillStyle = /** @type {string} */ (instruction[1]);
+          context.fillStyle = /** @type {string} */ (instruction[1]);
         } else {
-            context.fillStyle = instruction[1](context);
+          context.fillStyle = instruction[1](context);
         }
         ++i;
         break;
@@ -1392,9 +1393,9 @@ ol.render.canvas.PolygonReplay.prototype.setFillStrokeStyle =
     var fillStyleColor = fillStyle.getColor();
     var fillStyleRenderer = fillStyle.getRenderer();
     if (fillStyleRenderer) {
-        this.state_.fillStyle = fillStyleRenderer;
+      this.state_.fillStyle = fillStyleRenderer;
     } else {
-        state.fillStyle = ol.color.asString(fillStyleColor ?
+      state.fillStyle = ol.color.asString(fillStyleColor ?
             fillStyleColor : ol.render.canvas.defaultFillStyle);
     }
   } else {
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -1133,14 +1133,14 @@ ol.render.canvas.PolygonReplay = function(tolerance, maxExtent, resolution) {
 
   /**
    * @private
-   * @type {{currentFillStyle: (string|function(CanvasRenderingContext2D)|undefined),
+   * @type {{currentFillStyle: (string|function(CanvasRenderingContext2D):CanvasPattern|undefined),
    *         currentStrokeStyle: (string|undefined),
    *         currentLineCap: (string|undefined),
    *         currentLineDash: Array.<number>,
    *         currentLineJoin: (string|undefined),
    *         currentLineWidth: (number|undefined),
    *         currentMiterLimit: (number|undefined),
-   *         fillStyle: (string|function(CanvasRenderingContext2D)|undefined),
+   *         fillStyle: (string|function(CanvasRenderingContext2D):CanvasPattern|undefined),
    *         strokeStyle: (string|undefined),
    *         lineCap: (string|undefined),
    *         lineDash: Array.<number>,
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -1138,7 +1138,7 @@ ol.render.canvas.PolygonReplay = function(tolerance, maxExtent, resolution) {
    *         currentLineJoin: (string|undefined),
    *         currentLineWidth: (number|undefined),
    *         currentMiterLimit: (number|undefined),
-   *         fillStyle: (string|undefined),
+   *         fillStyle: (string|function(CanvasRenderingContext2D)|undefined),
    *         strokeStyle: (string|undefined),
    *         lineCap: (string|undefined),
    *         lineDash: Array.<number>,
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -1131,7 +1131,7 @@ ol.render.canvas.PolygonReplay = function(tolerance, maxExtent, resolution) {
 
   /**
    * @private
-   * @type {{currentFillStyle: (string|undefined),
+   * @type {{currentFillStyle: (string|function(CanvasRenderingContext2D)|undefined),
    *         currentStrokeStyle: (string|undefined),
    *         currentLineCap: (string|undefined),
    *         currentLineDash: Array.<number>,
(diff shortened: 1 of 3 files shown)


ahocevar:
This is very slick. I like it. Only one thing I'd change for performance reasons: Instead of reusing the SET_FILL_STYLE instruction, I'd introduce a new instruction type, e.g. SET_FILL_STYLE_FUNCTION.

What do others think?

To make this ready for prime time, the new `renderer` option also has to be added to `externs/olx.js`. And it would be awesome to have the above fiddle example added as example to the pull request.

elemoine:
I really like the general idea. But I wonder if we could make it more convenient, and more consistent with the rest of the style API. Instead of a function that returns a pattern the user could just provide a canvas (or any other image type), from which the renderer would create a pattern. So I'd just name the option `pattern`or `patternImage`.

elemoine:
And in the future the library could provide built-in pattern images, and/or built-in pattern image factories.

phelanHayes:
Does anyone have an update on this pull request?  It is a necessary feature for mapping apis and agreed, slick implementation. Would like to use ol3 in a project and this is a requirement.

cisso:
@phelanHayes: Until this is merged you can use the debug version of ol 3.10.0 and monkey-patch it by also including the code provided in the fiddle above (just remove the example code at the end).

cisso:
I also wrote a generic factory function to generate any stripe patterns, which could be included by default:

```javascript
function createStripesRenderer(colors, strokeWidth, angle) {
    var
        // Adjust for rounding errors
        pad     = .1,
        radian  = angle * Math.PI / 180,
        width   = strokeWidth * colors.length,
        height  = strokeWidth * colors.length,
        widthR  = width / Math.cos(radian),
        heightR = height / Math.sin(radian),
        canvas, ctx, i;

    canvas = document.createElement('canvas');
    canvas.width = widthR + pad;
    canvas.height = heightR;
    ctx = canvas.getContext('2d');
    ctx.lineWidth = strokeWidth;

    ctx.translate((widthR - width) / 2, 0);
    ctx.rotate(radian);
    for(i = -colors.length; i < (colors.length * 2); i++) {
        ctx.strokeStyle = colors[(i + colors.length) % colors.length];
        ctx.beginPath();
        ctx.moveTo(strokeWidth * i + strokeWidth * .5, -heightR);
        ctx.lineTo(strokeWidth * i + strokeWidth * .5, heightR);
        ctx.stroke();
    }

    return function(olContext) {
        return olContext.createPattern(canvas, 'repeat');
    }
}
````

The following is a simple example (colors can be any valid color declarations):
```javascript
    fill: new ol.style.Fill({
      renderer: createStripesRenderer(['blue', 'red', 'green', 'yellow', 'orange'], 30, 12)
    }),
```
producing the following output:
stripes-example

phelanHayes:
@cisso:  Thank you for your prompt reply!  And for the very useful createStripesRenderer function! I intend to put it to use in a proof of concept straight away.  Thanks to all for the excellent work on OL3.  I look forward to Generic Fill Rendering merge.

bjnsn:
@ahocevar Added pull-request (https://github.com/RouquinBlanc/ol3/pull/1) to @RouquinBlanc's repository which I believe addresses the concern about this pull-request you noted in <a href="#issuecomment-149038204">your comment</a>

ahocevar:
Thanks @bjnsn, this is exactly what I meant. Now the only remaining problem is that the linter is complaining about lines that are too long. See 'Details' in the above failed check, and please fix them if possible.

@bjnsn and @RouquinBlanc, can you please submit [CLAs](https://github.com/openlayers/cla) so we can merge this? Thanks.

bjnsn:
@ahocevar Pull-request for @RouquinBlanc  submitted that fixes the gjslint issues.

https://github.com/RouquinBlanc/ol3/pull/2

src/ol/style/fillstyle.js
```diff
@@ -26,6 +26,12 @@ ol.style.Fill = function(opt_options) {
 
   /**
    * @private
+   * @type {function(CanvasRenderingContext2D)}
+   */
+  this.renderer_ = options.renderer !== undefined ? options.renderer : null;
```
ahocevar: This option needs to be added to `olx.style.FillOptions` in `externs/olx.js`, otherwise this will only work in debug mode. And the option should be documented there.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -1386,8 +1393,13 @@ ol.render.canvas.PolygonReplay.prototype.setFillStrokeStyle =
   var state = this.state_;
   if (fillStyle) {
     var fillStyleColor = fillStyle.getColor();
-    state.fillStyle = ol.color.asString(fillStyleColor ?
-        fillStyleColor : ol.render.canvas.defaultFillStyle);
+    var fillStyleRenderer = fillStyle.getRenderer();
+    if (fillStyleRenderer) {
+      this.state_.fillStyle = fillStyleRenderer;
```
tschaub: This could be `state` instead of `this.state_`.

src/ol/style/fillstyle.js
tschaub: This function also needs to return `string|CanvasPattern|CanvasGradient`.

tschaub:
I'm also thinking the same as @elemoine above (https://github.com/openlayers/ol3/pull/4241#issuecomment-149642732).  Is there a strong use case for passing a function that gets called at render time?

One alternative would be to support something like this:

```js
var fill = new ol.style.Fill({
  img: myImg,
  imgRepetition: 'repeat' // this could be the default
});
```

This is pretty specific to `CanvasPattern`.  We could support `color`, `pattern`, or `gradient` options or overload `color` to accept `string|CanvasPattern|CanvasGradient` - since in the end this is just used to set [`context.fillStyle`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillStyle).

Curious what others think.

ahocevar:
> We could support color, pattern, or gradient options or overload color to accept string|CanvasPattern|CanvasGradient - since in the end this is just used to set context.fillStyle.

I was thinking along the same lines. It should not be too hard to make this change on top of what's been done already in this pull request. Maybe a challenging bonus task for our contributors here?

bjnsn:
Hmmm. The problem with overloading to accept a CanvasPattern or CanvasGradient is that a CanvasRenderingContext2D is needed to create them. I believe that this needs to be the context in which it will be used (due to the same-origin policy) although *I'm not absolutely sure about that*. However, we could certainly overload color with a factory that returns a CanvasPattern or CanvasGradient.

tschaub:
Here is an example creating a pattern on one context and using it on another.  I assume the information about the origin is preserved in the pattern (and then recorded on the target context).  It would be good to put together an example to confirm the origin information is not lost when using a different context.

There is a fairly significant difference in the current approach of accepting a function and the idea of accepting a image (or pattern or gradient).  The function is called with every render frame - so a new pattern is created for every frame.  If instead we accept an image/pattern/gradient, it only needs to be created once.

bjnsn:
@tschaub I'm not seeing the example you mention.

bjnsn:
@tschaub @ahocevar See https://github.com/openlayers/ol3/pull/4632 for an alternative.

tschaub:
> @tschaub I'm not seeing the example you mention.

Oops.  I updated the comment to include the link: https://jsfiddle.net/y1p73sct/

fredj:
fixed with #4632

phelanHayes:
Thank you Tim!


Sent from my Sprint Samsung Galaxy S® 6.
-------- Original message --------From: Tim Schaub <notifications@github.com> Date: 1/6/16  2:16 PM  (GMT-05:00) To: openlayers/ol3 <ol3@noreply.github.com> Cc: phelanHayes <matthewsmith413@yahoo.com> Subject: Re: [ol3] Add generic Fill rendering as per #2208 (#4241) 

@tschaub I'm not seeing the example you mention.


Oops.  I updated the comment to include the link: https://jsfiddle.net/y1p73sct/

PR #4242, opened 2015-10-09
Fix a drawText regression introduced with 31a68e2

### LINKED ISSUE #4240: Text fill style do not show unless they have stroke style as well.
Hello,

I upgraded to 3.10.0 this morning and when I run my app I have noticed that all labels disappeared.

After some research I have found the problem. If I have the fill style I also need the stroke style set for the ol.style.Text.
This is caused by this piece of code:

ol.render.canvas.TextReplay.prototype.drawText =
    function(flatCoordinates, offset, end, stride, geometry, feature) {
  if (this.text_ === '' ||
      !this.textState_ ||
      (this.textFillState_ &&
       !this.textStrokeState_)) {
    return;
  }
....

As you can see if fill style is set and stroke style is not - the method quits and does not render the text.
While the method:

ol.render.canvas.TextReplay.prototype.setTextStyle

sets the values for textFillState_ and textStrokeState_ independantly and does not require textStrokeState_ to exist if the fill is given.

As a temporal workaround I have added the stroke style to my text style
text.stroke = new ol.style.Stroke({color: 'rgba(0, 0, 0, 0)'})
just to make it work as expected.

But I am guessing it might be a bug :)

marcjansen:
This PR fixes a regression from 31a68e21a51ee378c3698354523d185eef52f543

Previously the relevant lines looked:

```javascript
  if (this.text_ === '' ||
      goog.isNull(this.textState_) ||
      (goog.isNull(this.textFillState_) &&
       goog.isNull(this.textStrokeState_))) {
    return;
  }
```

after 31a68e21a51ee378c3698354523d185eef52f543

They look like

```javascript
  if (this.text_ === '' ||
      !this.textState_ ||
      (this.textFillState_ &&
       !this.textStrokeState_)) {
    return;
  }
```

I think it should be 

```javascript
  if (this.text_ === '' || !this.textState_ ||
      (!this.textFillState_ && !this.textStrokeState_)) {
    return;
  }
```

Please review.

Thanks @jmekaelthas for noticing this in #4240.

Fixes #4240

/cc @tschaub, @ahocevar

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -1560,10 +1560,8 @@ goog.inherits(ol.render.canvas.TextReplay, ol.render.canvas.Replay);
  */
 ol.render.canvas.TextReplay.prototype.drawText =
     function(flatCoordinates, offset, end, stride, geometry, feature) {
-  if (this.text_ === '' ||
-      !this.textState_ ||
-      (this.textFillState_ &&
-       !this.textStrokeState_)) {
+  if (this.text_ === '' || !this.textState_ ||
+      (!this.textFillState_ && !this.textStrokeState_)) {
     return;
   }
   if (this.textFillState_) {


elemoine:
Looks good. Please merge.

I think this deserves a 3.10.1.

marcjansen:
I aggree, we should probably release 3.10.1. I cannot do it, sorry.

ahocevar:
Releasing 3.10.1 now.

ahocevar:
Done.

marcjansen:
Merci beaucoup! And sorry for introducing the bug

PR #4243, opened 2015-10-09
Add source to ol.source.VectorEvent

gberaudo:
This commit avoids creating one listener per source.
It also simplifyhow a listener is set: since the listener is reused,
setting the listener several time on the same source will be a no-operation.


elemoine:
Can't you just use `event.target`?

gberaudo:
Thank you @elemoine for pointing me this.

PR #4245, opened 2015-10-09
David hequet patch 1

DavidHequet:
Add a notify optionnal argument to object set and setProperties method

Compat : if not provided, Object.set and setProperties will notify event

Greatly improve performance if you update lot of features or frequently without need of beeing notified.


jonataswalker:
@DavidHequet did you read carefully this https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#running-the-check-target?

DavidHequet:
Yes, but got some restrictions at work downloading requiered applications.

I tried to make my pull request from editing sources through github, but copy/paste didn't work as expected..

i close the request until i get what's needed

PR #4248, opened 2015-10-09
Use Math.cosh of ES6/2015 if available

marcjansen:
This PR suggests two smaller refactorings with regard to `ol.math.cosh`:

* 449131a516f38e4816905c7db433e92287e20e08 refactors the implementation in a way which saves us one call to `Math.exp`. This variant performs better for me, and the code is taken from [the MDN reference implmentation](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Math/cosh)
* cd99d44704946519a82d22c89c08aac08e49263b then changes the code again, so that the native method `Math.cosh` (standardized in ES6 / 2015) is used if it is available (Which naturally gives us best prefromance).

Here is a jsperf compairing the implementations: http://jsperf.com/cosh-num-calls-math-exp

Please review.

### DIFF
--- a/src/ol/math.js
+++ b/src/ol/math.js
@@ -21,7 +21,8 @@ ol.math.clamp = function(value, min, max) {
  * @return {number} Hyperbolic cosine of x.
  */
 ol.math.cosh = function(x) {
-  return (Math.exp(x) + Math.exp(-x)) / 2;
+  var y = Math.exp(x);
+  return (y + 1 / y) / 2;
 };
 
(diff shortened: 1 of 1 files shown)


marcjansen:
Thanks for any input here.

marcjansen:
Thanks for any review.

ahocevar:

marcjansen:
Thanks for the review, Andreas.

PR #4249, opened 2015-10-11
Make whole example box clickable

### LINKED ISSUE #1300: example list: tap on the whole item instead of only the link in the top
I think it would be more friendly if people could tap on the whole item and not just the link in the example list, especially for mobile

http://ol3js.org/en/master/examples/

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/1449556-example-list-tap-on-the-whole-item-instead-of-only-the-link-in-the-top?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ahocevar:
Fixes #1300.

cc @bartvde.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -18,6 +18,7 @@
         color: #333;
       }
       .example {
+        display: block;
         padding: 10px;
         background-color: #F5F5F5;
         height: 140px;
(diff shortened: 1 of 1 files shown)


bartvde:
LGTM @ahocevar I remember opening up a ticket for this a long time ago but not sure where it was

bartvde:
ah right #1300 indeed I didn't see the reference
Nov 2013 ;-)

Thanks for fixing this.

PR #4252, opened 2015-10-12
optional silent mode for set and setProperties

DavidHequet:
Hi, 

I had to update 200+ features (geometry + 10~20 properties) each 2 seconds. The number of notify event's caused the map to hang on every updates. 

I modified the set and setProperties method with an optional parameter that allow the user to choose if he wan't to be notified of updated properties (in real usage: when you programatically update features, you know that properties ares updated, so being notified is useless)

Compatibility : if not provided, Object.set and setProperties will notify event, no adaptation is needed.

### DIFF
--- a/src/ol/object.js
+++ b/src/ol/object.js
@@ -202,12 +202,12 @@ ol.Object.prototype.notify = function(key, oldValue) {
  */
 ol.Object.prototype.set = function(key, value, notify) {
   if (goog.isDef(notify) && notify === false) {
-      this.values_[key] = value;
+    this.values_[key] = value;
   }
   else {
-      var oldValue = this.values_[key];
-      this.values_[key] = value;
-      this.notify(key, oldValue);
+    var oldValue = this.values_[key];
+    this.values_[key] = value;
+    this.notify(key, oldValue);
   }
 };
 
--- a/src/ol/object.js
+++ b/src/ol/object.js
@@ -197,7 +197,7 @@ ol.Object.prototype.notify = function(key, oldValue) {
  * Sets a value.
  * @param {string} key Key name.
  * @param {*} value Value.
- * @param {boolean=} notify update propertie silently
+ * @param {boolean=} opt_notify update propertie silently
  * @api stable
  */
 ol.Object.prototype.set = function(key, value, notify) {
@@ -216,7 +216,7 @@ ol.Object.prototype.set = function(key, value, notify) {
  * Sets a collection of key-value pairs.  Note that this changes any existing
  * properties and adds new ones (it does not remove any existing properties).
  * @param {Object.<string, *>} values Values.
- * @param {boolean=} notify update propertie silently
+ * @param {boolean=} opt_notify update propertie silently
  * @api stable
  */
 ol.Object.prototype.setProperties = function(values, notify) {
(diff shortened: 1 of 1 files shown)


DavidHequet:
I replacer opt_notify with opt_silent wich is more intuitive for an optionnal silent arguments :)

src/ol/object.js
```diff
@@ -197,12 +197,18 @@ ol.Object.prototype.notify = function(key, oldValue) {
  * Sets a value.
  * @param {string} key Key name.
  * @param {*} value Value.
+ * @param {boolean=} opt_silent update propertie silently
```
tschaub: This doc could read `Update property without triggering notification.`

src/ol/object.js
```diff
@@ -197,12 +197,18 @@ ol.Object.prototype.notify = function(key, oldValue) {
...
  * @api stable
  */
-ol.Object.prototype.set = function(key, value) {
-  var oldValue = this.values_[key];
-  this.values_[key] = value;
-  this.notify(key, oldValue);
+ol.Object.prototype.set = function(key, value, opt_silent) {
+  if (goog.isDef(opt_silent) && opt_silent === true) {
```
tschaub: There is no need for the `goog.isDef(opt_silent)` check here.  It should be enough to check `opt_silent === true`.

src/ol/object.js
```diff
@@ -197,12 +197,18 @@ ol.Object.prototype.notify = function(key, oldValue) {
...
+    this.values_[key] = value;
+  }
+  else {
```
tschaub: To conform with the style of the rest of the library, please write `else` on the same line as the preceding `}`.

src/ol/object.js
```diff
@@ -223,12 +230,15 @@ ol.Object.prototype.setProperties = function(values) {
...
+ * @param {boolean=} opt_silent update propertie silently
  * @api stable
  */
-ol.Object.prototype.unset = function(key) {
+ol.Object.prototype.unset = function(key, opt_silent) {
   if (key in this.values_) {
     var oldValue = this.values_[key];
     delete this.values_[key];
-    this.notify(key, oldValue);
+    if (!goog.isDef(opt_silent) || opt_silent !== true) {
```
tschaub: Same as above regarding the use of `goog.isDef()`.  It is enough to check `!opt_silent`.

tschaub:
> in real usage: when you programatically update features, you know that properties ares updated, so being notified is useless

Yes, you do not need to be notified.  But the library itself depends on this notification.  For example, features are re-rendered when properties change.

I think providing an option to disable notification is a nice enhancement though.

@DavidHequet if you are able to update your change to address the comments above, that would be great.  And to avoid the string of redundant commits, you could squash these commits into a single commit.

DavidHequet:
Hi, yes i undestand that the notify system is used deeply in OL3.
Maybe that a a warn in api docs could be usefull about this?

I think that some other enhancements can be done in the "setProperties" : the method notify about change, even if you set the same properties. 

Lots of lazy programer (including me before i discovered this :) ) can update all properties of features, even if just one propertie actually changed. If it cause re-render of feature at each notify, il could cause a drop in performance.

tschaub:
Thanks for the contribution @DavidHequet.  I've squashed your commits into one and opened #4394 to get this in.

PR #4258, opened 2015-10-12
Remove usage of goog.dom.TagName enum

marcjansen:
Instead of using e.g. `goog.dom.TagName.UL`, directly use an appropriate string; in this case `'UL'`.

The buildsizes are exactly equal before and after the suggested change.

Please review.


tschaub:
Looks good.  Please merge.

marcjansen:
Thanks for the review.

PR #4259, opened 2015-10-12
Remove use of toDegrees/toRadians util functions

marcjansen:
Instead of `goog.math.toRadians` and `goog.math.toDegrees`, we now use
our own implementations of these basic conversion functions.

Please review.


tschaub:
Looks good.  Please merge.

marcjansen:
Thanks!

Issue #4262, opened 2015-10-13
Map not moving at MS Edge on Win10 Mobile


goshi:
When click to the map and try to move it - nothing happens. Zoom works fine. Clicks on the overlays work fine too. 
Tested on the map on main page openlayers.org and projects for some clients

ahocevar:
Have you tried the latest version, v3.10.1? This issue should be fixed already. Try e.g. with http://openlayers.org/en/v3.10.1/examples/simple.html.

fredj:
Caused by #4167 (v3.10.1 if affected too).
I'm on it.

goshi:
Nope, not works. Edge version 20.10536.1004.0

fredj:
@goshi the issue is now fixed in the master branch. Thanks for the report.
You can tests it here: http://openlayers.org/en/master/examples/

goshi:
Thanks, @fredj. Now it is ok!

PR #4263, opened 2015-10-13
Add back CSSProperties.prototype.touchAction extern

fredj:
See #4262


ahocevar:
Looks good. Please merge when Travis is happy. And thanks for fixing this regression so quickly.

fredj:
FYI https://github.com/google/closure-compiler/pull/1206

PR #4261, opened 2015-10-13
ol.events.condition.mouseOnly may be wrong

fredj:
[The comment inside the function](https://github.com/openlayers/ol3/blob/master/src/ol/events/condition.js#L206-L208) says that if `pointerId` is `1` the event was triggered by a mouse. The [finalized version](http://www.w3.org/TR/pointerevents/#widl-PointerEventInit-pointerId) of the specification do not mention it.


fredj:
Thanks for any review

tsauerwein:
I guess you have tested on IE 10 and 11? Otherwise I have no objections with this change.

fredj:
Tested on IE edge, chrome and FF.
I will test on IE 9, 10 and 11

fredj:
ok on IE9, IE10 and IE11

tsauerwein:
Thanks for testing!

PR #4268, opened 2015-10-14
Add method for retrieving ol.Overlay by id

jonataswalker:
~~This PR follows the same guidelines of https://github.com/openlayers/ol3/pull/2087.~~

This PR adds:
- [x] ol.Overlay.id
- [x] ol.Overlay#getId
- [x] ol.Map#getOverlayById

The reason is pretty obvious: Get an overlay without a loop iteration.

### DIFF
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -343,6 +343,13 @@ ol.Map = function(options) {
   this.overlays_ = optionsInternal.overlays;
 
   /**
+   * A lookup of overlays by id.
+   * @private
+   * @type {Object.<string, ol.Overlay>}
+   */
+  this.overlayIdIndex_ = {};
+
+  /**
    * @type {ol.renderer.Map}
    * @private
    */
@@ -466,7 +473,7 @@ ol.Map = function(options) {
        * @this {ol.Map}
        */
       function(overlay) {
-        overlay.setMap(this);
+        this.addOverlayInternal(overlay);
       }, this);
 
   goog.events.listen(this.overlays_, ol.CollectionEventType.ADD,
@@ -474,7 +481,7 @@ ol.Map = function(options) {
        * @param {ol.CollectionEvent} event Collection event.
        */
       function(event) {
-        event.element.setMap(this);
+        this.addOverlayInternal(/** @type {ol.Overlay} */ (event.element));
       }, false, this);
 
   goog.events.listen(this.overlays_, ol.CollectionEventType.REMOVE,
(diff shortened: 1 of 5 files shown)


ahocevar:
This could be much simpler if you didn't make `id` an `ol.Object` property. Then you wouldn't have to handle id changes. An id shouldn't change anyway. Just make the id an instance property instead, and don't add a setter, but only make it configurable upon instantiation.

You'll also have to add the `id` to `olx.OverlayOptions` in `externs/olx.js`.

jonataswalker:
@ahocevar *"only make it configurable upon instantiation"*? This is quite different of `ol.Feature` getId/setId, is that you want?

```javascript
var overlay = new ol.Overlay({
    element: target,
    id: 'foo',
    position: [0, 0]
});
```

ahocevar:
@jonataswalker That's what I had in mind. Would that be sufficient for your use case?

jonataswalker:
Indeed, much simpler, bye bye lots of tests. Funny that I always wish to set `ol.Feature` id on constructor.

src/ol/map.js
```diff
@@ -466,7 +473,7 @@ ol.Map = function(options) {
        * @this {ol.Map}
        */
       function(overlay) {
-        overlay.setMap(this);
+        this.addOverlayInternal(overlay);
       }, this);
```
ahocevar: Can be replaced with
```js
this.overlays_.forEach(this.addOverlayInternal_, this);
```

src/ol/map.js
```diff
@@ -540,6 +547,20 @@ ol.Map.prototype.addOverlay = function(overlay) {
 
 
 /**
+ * This deals with map's overlay collection changes.
+ * @param {ol.Overlay} overlay Overlay.
+ * @protected
```
ahocevar: Should be `@private`.

src/ol/map.js
```diff
@@ -1246,6 +1281,20 @@ ol.Map.prototype.removeOverlay = function(overlay) {
...
+ */
+ol.Map.prototype.removeOverlayInternal = function(overlay) {
```
ahocevar: This method is only called once, so you can inline the code in the `ol.CollectionEventType.REMOVE` listener.

src/ol/overlay.js
```diff
@@ -188,6 +194,16 @@ ol.Overlay.prototype.getElement = function() {
 
 
 /**
+ * Get the feature identifier which is set on constructor.
```
ahocevar: This is not a *feature* identifier, but an *overlay* identifier. Also, does this need to be `@api`?

ahocevar:
Thanks @jonataswalker. I added a few more suggestions. When you're done addressing these, we can merge this.

jonataswalker:
@ahocevar thanks for review and improvement. All your notes are addressed.

ahocevar:
Thanks @jonataswalker, looks great now.

PR #4270, opened 2015-10-14
Fix Keyboard zoom interation options

adube:
Allow `0` values.


ahocevar:
Is `0` really a sensible value for duration and delta? It might be for duration, but I'm pretty sure it isn't for delta.

adube:
You're right.  I updated the PR accordingly.

ahocevar:
Nice, thanks for updating. I'll merge when Travis says ok.

PR #4271, opened 2015-10-14
Fix DragZoom interaction duration, allow `0` value

adube:


ahocevar:
Thanks for spotting this regression @adube. It would be nice if we had tests for this.

Issue #4264, opened 2015-10-13
Projection handling poorly documented for ol.source.Vector


Barryrowe:
I'm working through some simple examples, and I've found that for some reason valid GeoJSON data that can be read as features inline, or manually loaded into a layer via an AJAX call, fails to load and render properly when using 

    new ol.source.Vector({
        url: 'my-url/here.geojson',
        format: new ol.format.GeoJSON()
    })

It works fine when I use the example geoJson from the vector layer example: http://openlayers.org/en/master/examples/vector-layer.html

However, when I load in a custom geoJSON file through the url option, it never renders the features. I can debug and see the load featureloader.xhr succeeds and gets the text properly. I can even see it properly parse the features and validate they get added to the source. I also watched the projection at that point, and it is reflecting that set on the View. However, the features don't ever render.

I've created a repro here in this plnkr: http://plnkr.co/edit/Sw7kIad6NnxrUf6dfDdD

It creates 3 layers with the data in question:

1. does so inline, where the data is an inline literal javascript object
2. attempts to use the url option on a Vector source (should fill in yellow circles for the 2 points but doesn't)
3. manually makes an ajax call and adds calls readFeatures from the result (does so if you click the button)

I'm hoping this is something with my data, but it appears that the url option on ol.source.Vector is just partially broken

ahocevar:
It's your data, because the GeoJSON does not advertise the CRS of the coordinates it contains. But you can work around that by overriding the default data projection (which is EPSG:4326) of the format:
```
format: new ol.format.GeoJSON({
  defaultDataProjection: 'EPSG:3857'
}),
```

Barryrowe:
@ahocevar This does not solve the problem. I updated the plnkr to provide a defaultDataProjection on the source of the layer loading via url, and it still does not render in the features.

You can see in the plnkr that the inline version of the Layer also does not have a CRS property on the FeatureCollection and does not provide a defaultDataProjection on it's format, but loads without issue.

I can also manually load the data via ajax and create a layer like below and that works without setting defaultDataProjection.

     $.getJSON('two-points.geojson', function (data) {
              var remoteLayer = new ol.layer.Vector({
                  source: new ol.source.Vector({
                      features: (new ol.format.GeoJSON()).readFeatures(data)
                  }),
                  style: manualAjaxStyleFunction
              });
              map.addLayer(remoteLayer);
          });

[...]

PR #4273, opened 2015-10-14
Expand docs on ol.source url prop

Barryrowe:
Submitting documentation change for Issue #4264

Expanded documentation on the url property of the
ol.source.Vector property to include information about
how data loaded remotely will always be run through an
attempt to translate to the view's projection so the
format object must have the proper value for
defaultDataProjection when the remote data doesn't
include a CRS property.


externs/olx.js
```diff
@@ -5075,6 +5075,12 @@ olx.source.VectorOptions.prototype.strategy;
  * the given URL. Use a {@link ol.FeatureUrlFunction} to generate the url with
  * other loading strategies.
  * Requires `format` to be set as well.
+ * When a `string` is used and the default XHR feature loader is provided, the
+ * data loaded from the remote url will be read into features with using the
+ * provided `format` object, and will attempt to translate the features to the
+ * current view projection. If your remote data is not in the default 
+ * EPSG:4326 projection, make sure you provide the `defaultDataProjection` on
```
ahocevar: This also applies when a function is used. But 'default EPSG:4326' projection only applies to some formats. Others always advertise their projection. And the `defaultDataProjection` option is only available for GeoJSON and TopoJSON. So maybe something like:

When the default `loader` is used, features will be transformed from the data projection to the view projection during parsing. If your remote data source does not advertise its projection properly, this transformation will be incorrect. For some formats, the default projection (usually EPSG:4326) can be overridden by setting the `defaultDataProjection` constructor option on the format.

When pr

ahocevar:
Merged with 519e7976cd7ded2e83e2e2947dc2824df2fd5e2d.

PR #4275, opened 2015-10-15
Fix ol.events.condition.mouseOnly parameter type

fredj:
From `ol.MapBrowserEvent` to `ol.MapBrowserPointerEvent` and remove the
instanceof assert.


tsauerwein:
This makes sense, thanks!

fredj:
Thanks for the review Tobias

PR #4276, opened 2015-10-15
Better ol.events.condition.platformModifierKeyOnly documentation

fredj:
See https://github.com/google/closure-library/blob/master/closure/goog/events/browserevent.js#L175-L180


fredj:
The windows-key has nothing to do with it

elemoine:
Thanks. Please merge when build is finished.

PR #4277, opened 2015-10-15
Improve the box select example

fredj:
Changes the condition of the drag box from `shift` to `ctrl` to be able to still use the darg box zoom interaction.

### DIFF
--- a/examples/box-selection.html
+++ b/examples/box-selection.html
@@ -5,7 +5,7 @@ shortdesc: Using a DragBox interaction to select features.
 docs: >
   <p>This example shows how to use a <code>DragBox</code> interaction to select features. Selected features are added
   to the feature overlay of a select interaction (<code>ol.interaction.Select</code>) for highlighting.</p>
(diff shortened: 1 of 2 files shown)


examples/box-selection.html
```diff
@@ -5,7 +5,7 @@
 docs: >
   <p>This example shows how to use a <code>DragBox</code> interaction to select features. Selected features are added
   to the feature overlay of a select interaction (<code>ol.interaction.Select</code>) for highlighting.</p>
-  <p>Use <code>SHIFT+drag</code> to draw boxes.</p>
+  <p>Use <code>Ctrl+drag</code> to draw boxes.</p>
```
elemoine: Do you want to say "Meta on Macs"?

elemoine:
Minor comment, otherwise lgtm.

examples/box-selection.html
fredj: yes, fixed

PR #4278, opened 2015-10-15
Use DOM instead of map canvas for ol.render.Box

### LINKED ISSUE #1583: No zoom box when using WebGL or DOM renderers
can be seen here for instance:

http://ol3js.org/en/master/examples/bing-maps.html

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/1449541-no-zoom-box-when-using-webgl-or-dom-renderers?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ahocevar:
With this change, the drag box we use for `ol.interaction.DragZoom` and `ol.interaction.DragBox` is just a DOM element with CSS styling. This way it works for all renderers.

Fixes #1583.

### DIFF
--- a/css/ol.css
+++ b/css/ol.css
@@ -1,3 +1,8 @@
+.ol-box {
+  box-sizing: border-box;
+  border-radius: 2px;
+  border: 2px solid blue;
+}
 
 .ol-mouse-position {
   top: 8px;
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,35 @@
 
 ### v3.11.0
 
+#### `ol.interaction.DragBox` and `ol.interaction.DragZoom` changes
+
+Styling is no longer done with `ol.Style`, but with pure CSS. The `style` constructor option is no longer required, and no longer available. Instead, there is a `className` option for the CSS selector. The default for `ol.interaction.DragBox` is `ol-dragbox`, and `ol.interaction.DragZoom` uses `ol-dragzoom`. If you previously had
+```js
+new ol.interaction.DragZoom({
+  style: new ol.style.Style({
+    stroke: new ol.style.Stroke({
+      color: 'red',
+      width: 3
+    }),
+    fill: new ol.style.Fill({
+      color: [255, 255, 255, 0.4]
+    })
+  })
+});
+```
+you'll now just need
+```js
+new ol.interaction.DragZoom();
+```
+but with additional css:
+```css
+.ol-dragzoom {
+  border-color: red;
+  border-width: 3px;
+  background-color: rgba(255,255,255,0.4);
+}
+```
+
 ### v3.10.0
 
 #### `ol.layer.Layer` changes
--- a/examples/box-selection.css
+++ b/examples/box-selection.css
@@ -1,0 +1,4 @@
+.ol-dragbox {
+  background-color: rgba(255,255,255,0.4);
+  border-color: rgba(100,150,0,1);
+}
--- a/examples/box-selection.js
+++ b/examples/box-selection.js
@@ -8,9 +8,6 @@ goog.require('ol.layer.Tile');
 goog.require('ol.layer.Vector');
 goog.require('ol.source.OSM');
 goog.require('ol.source.Vector');
-goog.require('ol.style.Fill');
-goog.require('ol.style.Stroke');
-goog.require('ol.style.Style');
 
 
 var vectorSource = new ol.source.Vector({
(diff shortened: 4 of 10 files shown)


css/ol.css
```diff
@@ -1,3 +1,8 @@
+.ol-box {
+  box-sizing: border-box;
+  border-radius: 2px;
+  border: 2px solid rgba(0,0,255,1);
```
bartvde: why not simply blue?

bartvde:
Theoretically drag zoom and drag box could have different styles in the past, but not in your new approach. Should we have 2 css classes?

css/ol.css
ahocevar: Because the old style had `color: [0, 0, 255, 1]`. But you're right, `blue` would be simpler.

bartvde:
Btw, so happy to see this PR!

fredj:
> Theoretically drag zoom and drag box could have different styles in the past, but not in your new approach. Should we have 2 css classes?

A `className` could be added to `olx.interaction.DragBoxOptions`

ahocevar:
Good point. Instead of removing the `style` constructor property, I could replace it with a `className` property, using `ol-dragbox` and `ol-zoombox` respectively. Makes sense?

ahocevar:
@fredj You were faster typing what we both thought at the same time it seems :-)

bartvde:
Makes perfect sense. You and @fredj had the same thought.

fredj:
:-)

fredj:
And the box-selection example should be updated:
```diff
diff --git a/examples/box-selection.js b/examples/box-selection.js
index ded8174..69ed2a9 100644
--- a/examples/box-selection.js
+++ b/examples/box-selection.js
@@ -8,9 +8,6 @@ goog.require('ol.layer.Tile');
 goog.require('ol.layer.Vector');
 goog.require('ol.source.OSM');
 goog.require('ol.source.Vector');
-goog.require('ol.style.Fill');
-goog.require('ol.style.Stroke');
-goog.require('ol.style.Style');
 
 
 var vectorSource = new ol.source.Vector({
@@ -44,15 +41,7 @@ var selectedFeatures = select.getFeatures();
 
 // a DragBox interaction used to select features by drawing boxes
 var dragBox = new ol.interaction.DragBox({
-  condition: ol.events.condition.platformModifierKeyOnly,
-  style: new ol.style.Style({
-    fill: new ol.style.Fill({
-      color: [255, 255, 255, 0.4]
-    }),
-    stroke: new ol.style.Stroke({
-      color: [100, 150, 0, 1]
-    })
-  })
+  condition: ol.events.condition.platformModifierKeyOnly
 });
 
 map.addInteraction(dragBox);
```

ahocevar:
Good catch. Will do @fredj.

bartvde:
Other than that this LGTM @ahocevar

ahocevar:
Thanks for the great suggestions. I just force-pushed a new version of the pull request.

ahocevar:
@fredj, do you want to take another look? In addition to what you suggested, I also gave `olx.interaction.ZoomBoxOptions` a `className` property, and added css to the box-selection example to match the previous styles.

fredj:
LGTM thanks

1Map:
In ol 3.10 I had the following:

`
			ol.events.condition.ctrlKeyOnly = function(mapBrowserEvent) {
				var browserEvent = mapBrowserEvent.browserEvent;
				return (!browserEvent.shiftKey && browserEvent.ctrlKey);
			};

			this.circleInteraction = new ol.interaction.DragBox({
				disabled : true,
				condition : ol.events.condition.ctrlKeyOnly,
				style : new ol.style.Style({
					fill : new ol.style.Fill({
						color : [255, 0, 0, 0.3]
					}),
					stroke : new ol.style.Stroke({
						color : [255, 0, 0, 1],
						width : 1
					})
				})
			});
`

How would I implement the above with the latest version?

Before it was:

`
this.box_ = new ol.render.Box(style);
`

Now it is:

`
this.box_ = new ol.render.Box(options.className || 'ol-dragbox');
`

ahocevar:
The [v3.11 upgrade notes](https://github.com/ahocevar/ol3/blob/7a070dd8597332a28c044b63052998fd26da784e/changelog/upgrade-notes.md#olinteractiondragbox-and-olinteractiondragzoom-changes) should answer your question. Let us know if anything remains unclear.

PR #4286, opened 2015-10-17
Add a color manipulation example

tschaub:
This adds an example that lets the user adjust hue, chroma, and lightness before rendering a raster source.


ahocevar:
Very nice. Please merge.

PR #4287, opened 2015-10-18
Add Arrow style for vector features

tananaev:
All or most applications related to navigation or tracking require to indicate direction of objects on a map. Arrow style provides default customizable and stylable implementation of the arrow shape for vector features.

Please consider including this style into the main openlayers code base.

### DIFF
--- a/examples/arrow.html
+++ b/examples/arrow.html
@@ -1,0 +1,13 @@
+---
+template: example.html
+title: Arrow example
+shortdesc: Example of Arrow style.
+docs: >
+
+tags: "vector, symbol, arrow, style"
+---
+<div class="row-fluid">
+  <div class="span12">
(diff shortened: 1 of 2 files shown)


ahocevar:
I wonder if we shouldn't do something more generic than just adding fixed shapes. If an image style could be configured with an `ol.geom.Geometry` with pixel coordinates, users could use arbitrary shapes as image style.

tananaev:
I agree that it would be nice to have a more generic solution. I can try to implement it using `ol.geom.SimpleGeometry` class.

Any other comments?

tananaev:
To draw geometry objects I guess I need to use `ol.render.canvas.Immediate`. Any ideas on how to properly re-use methods from there in my style object?

tschaub:

PR #4289, opened 2015-10-19
Restrict maxZoom of the static-image example

### LINKED ISSUE #3712: ImageStatic with high zoom levels in Firefox
When I open the [static image example](http://openlayers.org/en/v3.5.0/examples/static-image.html) of the OL examples in Firefox 38 (Ubuntu 14.04) I get a strange error at high zoom levels. The image begins to wobble and flicker when switching from one zoom level to the next highest (in the example it is from `9` to `10`). The error persists at even higher zoom levels. In Chrome 42 everything works as expected.

Here are some screenshots with Firefox on the left and Chrome on the right:
ol-z9
**zoom level `9`**

ol-z10
**zoom level `10`**

I could reproduce this issue with a custom map using the `canvas` and `dom` renderers. With the `webgl` renderer it seems to work correctly.

### LINKED ISSUE #4274: ImageStatic at high zoom levels in Chrome
The [static image example](http://openlayers.org/en/v3.10.1/examples/static-image.html) has a problem with Chrome at high zoom levels: after a certain zoom level it just disappears.
In the example, it happens at zoom 23, but I had the same thing happen in another map at a different zoom level. E.g. in [this fiddle](http://jsfiddle.net/zwwn05m3/) using a red/yellow 4x4 image with dataUrl, it happens at zoom 24.

I tried with two different versions of Chrome (*46.0.2490.71 m* and *48.0.2536.0 canary*), with the same results (see screenshots below).
I have no problem zooming to max zoom with IE (11.0.9600.18015) or Firefox (41.0.1) or Firefox Developer Edition (43.0a2).

zoom22
*Zoom 22*

zoom23
*Zoom 23*

I know this is almost the opposite of [another issue](https://github.com/openlayers/ol3/issues/3712) ... strange, but this is what happens to me

ahocevar:
Upscaling the image to zoom levels > 8 does not make sense, and it may even cause canvas transformations to fail.

Fixes #4274.
Fixes #3712.

### DIFF
--- a/examples/static-image.js
+++ b/examples/static-image.js
@@ -36,6 +36,7 @@ var map = new ol.Map({
   view: new ol.View({
     projection: projection,
     center: ol.extent.getCenter(extent),
-    zoom: 2
+    zoom: 2,
+    maxZoom: 8
   })
 });


fredj:
LGTM

PR #4291, opened 2015-10-19
Add 'Move a feature along a line' example

jonataswalker:
This PR adds an example of how to use an encoded polyline and move a feature along it.

### DIFF
--- a/examples/feature-move-animation.js
+++ b/examples/feature-move-animation.js
@@ -1,0 +1,198 @@
+goog.require('ol.Feature');
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.format.Polyline');
+goog.require('ol.geom.Point');
+goog.require('ol.layer.Tile');
+goog.require('ol.layer.Vector');
+goog.require('ol.source.BingMaps');
+goog.require('ol.source.Vector');
+goog.require('ol.style.Circle');
+goog.require('ol.style.Fill');
+goog.require('ol.style.Icon');
+goog.require('ol.style.Stroke');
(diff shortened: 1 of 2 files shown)


examples/feature-move-animation.js
```diff
@@ -0,0 +1,205 @@
...
+      stopAnimation(true);
+      return;
+    }
+
+    var currentPoint = new ol.geom.Point(routeCoords[index]);
+    var feature = new ol.Feature(currentPoint);
+    vectorContext.drawFeature(feature, styles.geoMarker);
+  }
+  // tell OL3 to continue the postcompose animation
+  frameState.animate = true;
```
tschaub: I think we should encourage folks to use [`map.render()`](http://openlayers.org/en/master/apidoc/ol.Map.html#render) here instead.

tschaub:
Thanks for the nice example @jonataswalker.  See my one [comment above](https://github.com/openlayers/ol3/pull/4291#discussion_r43546878).  Since `frameState.animate` is not documented as part of the API, I think calling `map.render()` makes more sense here.

jonataswalker:
@tschaub For sure, changed.

tschaub:
This looks great @jonataswalker.  We typically ask that people squash redundant commits (force pushing to your branch to clean up history), but we can save that for next time.

Thanks for the contribution!

jonataswalker:
My pleasure to contribute! I'll remember that (force pushing) tip.

tschaub:
Published: http://openlayers.org/en/master/examples/feature-move-animation.html

PR #4298, opened 2015-10-20
Interaction zindex

Vic-thor:
I've to display the select overlay and one feature what frame vertex. The problem is this feature render under the select overlay.
The idea is to be able to set zIndex for interaction overlay (not necessary the select interaction), like that :

```
this.selectInteraction = new ol.interaction.Select({
    layers: [myLayer],
    zIndex: 10
});
```

Then we can create a layer with an higher zIndex to render it over. Of course this layer will be rendered over all layers added with addLayer map method.

Before :

image

Now :

image

Thanks


tschaub:

PR #4306, opened 2015-10-22
updates to kml to render labels pulled from names if the geometry is a point

tamarmot:
I tried to replicate Google Earth's behavior, which is to render labels based on names for placemarks which are points.  OpenLayers now does this, either using the found style, default style, or custom style.  My new code creates secondary text style which holds the unique text from the name for each placemark.

This only happens if the placemark has a point geometry and if the name is not empty.

Sample screenshot from OpenLayers:
screen shot 2015-10-22 at 9 38 51 am

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,17 @@
 
 ### v3.11.0
 
+#### `ol.format.KML` changes
+
+KML icons are scaled 50% so that the rendering better matches Google Earth rendering.
+
+If a KML placemark has a name and is a point, an `ol.style.Text` is created with the name displayed to the right of the icon (if there is an icon).
+This can be controlled with the showPointNames option which defaults to true.
+
+To disable rendering of the point names for placemarks, use the option:
+new ol.format.KML({ showPointNames: false });
+
+
 #### `ol.interaction.DragBox` and `ol.interaction.DragZoom` changes
 
 Styling is no longer done with `ol.Style`, but with pure CSS. The `style` constructor option is no longer required, and no longer available. Instead, there is a `className` option for the CSS selector. The default for `ol.interaction.DragBox` is `ol-dragbox`, and `ol.interaction.DragZoom` uses `ol-dragzoom`. If you previously had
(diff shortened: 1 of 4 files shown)


tamarmot:
This is for issue https://github.com/openlayers/ol3/issues/4305

tamarmot:
Apparently my local CI environment differs from the Travis one.  Apologies.

marcjansen:
You can still combine the 5 commits into one with e.g. git rebase -i HEAD~5. In the editor that will pop-up, just change `pick` to `fixup` for all but the first commit in the list. You'll need to force push afterwards. If you are doing this for the first time, a backup doesn't hurt. HTH

tamarmot:
I have done so; when I was doing that previously it was not triggering the
continuous integration build.  New single commit pushed.

marcjansen:
Please do not use goog.string.*  or goog.object.isEmpty (clone is fine).

Please try to add a testcase if you can.

The default color for texts elsewhere is #333, not #000.

Once this is done, I'll have a closer look. I am on a mobile now,  so this has got to wait a little.

A side note: I think it'll be best if we have an option controlling this auto labeling, probably defaulting to true. And just in case a paragraph on this changed behaviour would be great.

Sorry for being so picky, but I really like the generall direction of this PR. Thanks for your work on this so far!

marcjansen:
.... a paragraph *in the 3.11 release notes* ... on this changed behaviour

tamarmot:
Great, thanks for the feedback -- I'll make those changes and push when
they are done!
Tamar

tamarmot:
I've pushed those changes, can you tell me where to update the 3.11 release
notes?

marcjansen:
They don't exist yet. Just copy https://github.com/openlayers/ol3/blob/master/changelog/v3.10.1.md, rename it and update the contents.

Thanks so far!

elemoine:
No, release notes should be added to https://github.com/openlayers/ol3/blob/master/changelog/upgrade-notes.md. The v3.11.0.md file will be created when releasing 3.11.0.

marcjansen:
Thanks Éric.

tamarmot:
Release notes added, feel free to edit.  Thanks!

marcjansen:
Thanks again. I'll review tomorrow.

marcjansen:
I have created #4316 as replacement for this PR (no failing build, fixed color and better commit messages).

PR #4311, opened 2015-10-24
Add useful getLength and getArea methods to ol.proj

bill-chadwick:
Useful linestring getLength and polygon getArea methods are added to ol.proj and the measure example is updated to use them.

The new methods give good answers in metres, pixels or other projection units whatever the latitude, and whatever the projection. Results in metres and metres squared are returned for EPSG:4326 and EPSG:3857. An optional ol.Sphere can be specified, which might be used for example, with non terrestrial mapping.

The custom getPointResolution function attribute of ol.proj is replaced with a boolean constantPointResolution attribute, with a getter and setter in ol.proj. constantPointResolution should be set true for projections such as UTM or EPSG:27700, which have nominally constant scale throughout their extent, either at construction time, or afterwards. constantPointResolution will default to false unless the projection's units are pixels or degrees, in which case it is set true. ol.proj.getPointResolution will return an unchanged resolution for constantPointResolution projections. When constantPointResolution is true and projection units are not degrees, length and area calculations can use simple, fast and accurate planar cartesian methods rather than spherical methods.

Documentation warnings are added to ol.geom getLength and getArea methods which for EPSG:3857 and EPSG:4326 do not return results in metres or metres squared. Also, a warning is added to the documentation of ol.geom.circle about the radius not being in metres.

Test for the new ol.proj methods are included.

I have had a go and squashing my commits into one but have failed - sorry.


bill-chadwick:
A review of the earth radius used for the calculation of length and area on the earth represented as a sphere might be in order.

Leaflet and Google Maps V3 use 6378137 metres by default which is the WGS84 equatorial radius. 

ol3 currently uses 6370997 metres in ol.sphere.NORMAL, 6378137 in ol.sphere.WGS84 and 6378137 in ol.proj.EPSG3857.RADIUS. I have defaulted to ol.sphere.NORMAL in the getLength and getArea methods submitted with this PR.

6378137 must be used for the EPSG:3857 projection. The  choice of radius to use for length and area calculations on a WGS84 equivalent sphere is not so obvious.

The 6370997 in ol.sphere.NORMAL is close, but not equal to, the WGS84 authalic (equal area) and mean radius which are 6371007 and 6371009 respectively either of which is probably a good candidate for calculations on an equivalent sphere.

Ideally projections would reference a geoid which could be represented by a sphere or an ellipsoid.
Proj4 projections could get semi-major radius 'a' and semi-minor radius 'b' or reciprocal flattening 'rf' form definition data and compute a mean radius for a sphere see https://en.wikipedia.org/wiki/Earth_radius. Its because of this that I have used an optional ol.Sphere attribute for the new getLength and getArea methods - it might one day become ol.Geoid or some such, and so support ellipsoidal calculations (Vincenty for getLength)

There is a an interesting article about the shortcomings of Web Mercator here

http://www.hydrometronics.com/downloads/Web%20Mercator%20-%20Non-Conformal,%20Non-Mercator%20%28notes%29.pdf

bill-chadwick:
Companion to the proposed getLength and getArea methods might be getRectangle and getCircle methods. 
These would return a list of 2D coordinates. getRectangle would have parameters of centre, width and height in projection units or metres if the projection units are degrees. getCircle would have parameters of centre, number of steps and radius in projection units or metres if the projection units are degrees. Calculations would proceed as for the proposed getLength and getArea using simple maths for Pixel and constantPoint Resolution projections and the ol.Sphere.offset method for other projections.

src/ol/proj/proj.js
```diff
@@ -278,6 +282,32 @@ ol.proj.Projection.prototype.setGlobal = function(global) {
...
+ * The attribute can not be set for the EPSG:3857 projection.
+ * It should only be set for projections with nominally constant scale
+ * or those with units of Pixels or Degrees.
+ * @api
+ */
+ol.proj.Projection.prototype.setConstantPointResolution =
+    function() {
+  if (this.code_ != 'EPSG:3857') {
+    this.constantPointResolution_ = true;
+  }
```
tschaub: This is an awkward method.

src/ol/geom/circle.js
```diff
@@ -223,7 +227,9 @@ ol.geom.Circle.prototype.setFlatCoordinates =
 
 
 /**
- * Set the radius of the circle. The radius is in the units of the projection.
+ * Set the radius of the circle in projected units.
+ * For EPSG:3857 or EPSG:4326 circles, this method
+ * does not set the radius in meters.
```
tschaub: I think it adds confusion (and mystery) if the bulk of the documentation describes what *does not* happen.

Plus, I don't agree that the units for EPSG:3857 are not meters.

If you feel like documentation needs to be added that explains that circles are only circles in projected coordinates (which holds true to the definition of a circle as points on a plane equidistant from a center point), that could be done as part of a separate pull request.

tschaub:
I think the addition of `ol.proj.getLength()` and `ol.proj.getArea()` is worth considering.  But this would be easier to review without the extra documentation changes and extraneous history.

examples/measure.html
```diff
@@ -1,10 +1,12 @@
 ---
 template: example.html
 title: Measure example
-shortdesc: Example of using the ol.interaction.Draw interaction for creating simple measuring application.
+shortdesc: Example of using the ol.interaction.Draw interaction for creating a simple measuring application.
 docs: >
-  <p><i>NOTE: If use geodesic measures is not checked, measure is done in simple way on projected plane. Earth curvature is not taken into account</i></p>
-tags: "draw, edit, measure, vector"
+  <p><i>NOTE: Area and length calculations are done using ol.proj.getLength and ol.proj.getArea to get answers in meters or square meters at any latitude.
+  The ol.geom.Linestring.getLength method does not give a length in meters for EPSG:3857 or EPSG:4326 linestrings.
```
ahocevar: This is confusing. Meters should always be used. Same with square meters below. These methods should always return meters or square meters.

src/ol/proj/proj.js
```diff
@@ -370,22 +391,124 @@ ol.proj.Projection.prototype.getPointResolution_ = function(resolution, point) {
...
- * measuring its width and height on the normal sphere,
- * and taking the average of the width and height.
- * An alternative implementation may be given when constructing a
- * projection. For many local projections,
- * such a custom function will return the resolution unchanged.
- * @param {number} resolution Resolution in projection units.
- * @param {ol.Coordinate} point Point.
- * @return {number} Point resolution in projection units.
+ * Calculate the length of a line string in projection units except
+ * if the projection's units are DEGREES, when a length in meters is returned.
```
ahocevar: I was hoping you were going to remove this inconsistency. Why not always projection units? Or meters?

src/ol/proj/proj.js
```diff
@@ -370,22 +391,124 @@ ol.proj.Projection.prototype.getPointResolution_ = function(resolution, point) {
...
+ * You can use e.g. map.getView().getProjection() to get
+ * the projection to pass as the projection parameter.
+ * The calculation method is Cartesian for projections with constant point
+ * resolution or those with units of PIXELS and Haversine for all others.
+ * The Haversine method calculates great circle distances.
+ * @param {Array.<ol.Coordinate>} coordinates Coordinates of a line string.
+ * @param {ol.proj.ProjectionLike} projection of the coordinates.
+ * @param {ol.Sphere=} opt_sphere optional sphere, defaults to ol.sphere.Normal
+ * @return {number} Length sum of the length of the line segments.
```
ahocevar: Meters?

examples/measure.js
```diff
@@ -286,21 +274,16 @@ var formatLength = function(line) {
...
-        sourceProj, 'EPSG:4326'));
-    var coordinates = geom.getLinearRing(0).getCoordinates();
-    area = Math.abs(wgs84Sphere.geodesicArea(coordinates));
-  } else {
-    area = polygon.getArea();
-  }
+
+  var coordinates = polygon.getLinearRing(0).getCoordinates();
+  var proj = map.getView().getProjection();
+  var area = ol.proj.getArea(coordinates, proj);
```
ahocevar: If `getArea()` received an `{Array.<Array.<ol.Coordinate>>}`, it would also work for polygons with holes and you wouldn't have to do `getLinearRing(0)` above.

ahocevar:
I was hoping that this would simplify things more, and remove confusion. Looking at the documentation, it seems to me that it adds more confusion than it removes. See my inline suggestions.

examples/measure.js
ahocevar: Of course `ol.geom.Polygon` for `getArea()` and `ol.geom.LineString` for `getLength()` would make even more sense than coordinate arrays.

examples/measure.html
ahocevar: Looking at the current documentation, it would also simplify things if length and area were always returned in projection units. We have a stable `ol.proj.METERS_PER_UNIT` for the user to convert to whatever units they want.

src/ol/proj/proj.js
bill-chadwick: Is it the 3857 check that bothers you ?
Actually, as getPoint Resolution is overridden for 3857, the check for 3857 could be dropped.

src/ol/proj/proj.js
bill-chadwick: Pixel projections can't return answers converted to meters. For EPSG:4326, and other projections with units of degrees, an answer in degrees or degrees squared is not useful, hence the exception that these return lengths in metres. Imagine loading a track from an OCX file and wanting to know its length.

examples/measure.js
bill-chadwick: I was not sure that ol.proj should be requiring ol.geom (or vice versa). What do you think ?

examples/measure.html
bill-chadwick: I have taken trouble to make the new methods return answers in projection units whenever possible. It is not possible to do so for projections with units of degrees. You can't convert a length in degrees calculated using Pythagoras  to metres. 

I am happy to drop the comment about th ol.geom.getLength and getArea from the example. 
Though this whole issue arose out of a desire to stop non GIS experts making mistakes using those methods improperly. 

METERS_PER_UNIT is not meaningful for Pixel Projections.

bill-chadwick: I think you are suggesting that a linestring getLength method for projections with units of degrees should some up the angle between consecutive pairs of points on the sphere and the sphere's centre. That could be then converted with ol.proj.METERS_PER_UNIT as you suggest. I am not sure that that process extends to ellipsoidal very well.

ahocevar: I had misread your example documentation above, and thought you were talking about the new functions. Maybe just remove that note.

examples/measure.js
ahocevar: Good point. I'm convinced about ol.geom, but I still think that getArea should take the result of ol.geom.Polygon.getCoordinates() as argument.

src/ol/proj/proj.js
ahocevar: The method does not take an argument, so the caller cannot set anything with it.

src/ol/proj/proj.js
ahocevar: True about pixels. For everything else, I think the return value should be in meters.

examples/measure.html
bill-chadwick: Actually converting back from Haversine metres to degrees is going to be tricky for spheres other than the Earth. There is currently only one value for ol.proj.METERS_PER_UNIT[ol.proj.Units.DEGREES] which is for the Earth (ol.sphere.NORMAL.radius), and used by ol.proj.Projection.prototype.getMetersPerUnit

src/ol/proj/proj.js
bill-chadwick: so setHasConstantPointResolution would be a better name ?. 
There is no useful purpose in setting constantPointResolution_ false.

examples/measure.js
bill-chadwick: Agreed. It can probably take either an array of coordinates or an array of arrays of coordinates. Is the use of goog.isArray still OK ?.

tschaub: `Array.isArray()` is supported by our [target browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility).

`ol.proj` should *not* depend on `ol.geom.*`.

bill-chadwick:
Here is my plan for progress on this. I still think its worthwhile to help some folk along the way to getting meaningful answers for length and area calculations.

1) Omit the documentation updates to the ol.geom method getLength and getArea methods and ol.geom.Circle for now. 

2) Omit the comments in the updated measure example about the ol.geom methods not giving good answers for 3857 and 4326.

3) Change both the new getArea and getLength methods to accept an array of coordinates or an array of arrays of coordinates.

3) Leave the return units of the new methods as meters, for projections with units of degrees and projection units (including pixels) for all other projections. I believe this is the most useful compromise. I don't think that returning an answer in degrees, using the radius of ol.sphere.Normal or the optional ol.Sphere passed to the methods would be very useful. The alternative of always returning a length in meters does not fit well with Pixel projections.

4) Rename the setConstantPointResolution method to setHasConstantPointResolution and drop its check for 3857. Instead of the check, override the function getHasConstantPointResolution  in ol.proj.epsg3957_ to return false always.

5) Drop the comment added about the radius value to ol.sphere.NORMAL - though I would like to continue the discussion of the best value to use.

That should leave a change set of projection.js, epsg3857projection.js, olx.js, proj.jsdoc, proj.test.js plus the measure example.

I will have another go at squashing the commits (I may need some help with that) and resubmitting as above in a few days time.

Any further ideas/comments are still welcome.

ahocevar:
@bill-chadwick:

> 1) Omit the documentation updates to the ol.geom method getLength and getArea methods and ol.geom.Circle for now.

Sounds good. Only add a `{@see ol.proj.getLength}` or `{@see ol.proj.getArea}` respectively.

> 2) Omit the comments in the updated measure example about the ol.geom methods not giving good answers for 3857 and 4326.

Sounds good.

> 3) Change both the new getArea and getLength methods to accept an array of coordinates or an array of arrays of coordinates.

For `getArea()` yes. But since there is no natural way to handle inner polygon rings when calculating the circumference of a polygon, I'd not do it for `getLength()`.

> 3) Leave the return units of the new methods as meters, for projections with units of degrees and projection units (including pixels) for all other projections. I believe this is the most useful compromise. I don't think that returning an answer in degrees, using the radius of ol.sphere.Normal or the optional ol.Sphere passed to the methods would be very useful. The alternative of always returning a length in meters does not fit well with Pixel projections.

I'd be more in favour of always returning meters. And assert and document that `ol.proj.getLength()` and `ol.proj.getArea()` cannot be used with projections with pixels as units - also reference `{@see ol.geom.Polygon#getArea}` and `{@see ol.geom.LineString#getLength}`.

> 4) Rename the setConstantPointResolution method to setHasConstantPointResolution and drop its check for 3857. Instead of the check, override the function getHasConstantPointResolution in ol.proj.epsg3957_ to return false always.

Remove both the method and the config option. Just assume constant resolution when units are pixels or degrees. This will also help to untangle the documentation a bit. For projections such as UTM or EPSG:27700, `ol.geom.Polygon#getArea()` and `ol.geom.LineString#getLength()` will return the correct area and length anyway.

5) Drop the comment added about the radius value to ol.sphere.NORMAL - though I would like to continue the discussion of the best value to use.

Yes, drop it. If you want to discuss that value, you should do so on the [MetaCRS list](http://lists.osgeo.org/listinfo/metacrs), because that's what's used in most OpenSource geospatial projects.

src/ol/geom/circle.js
bill-chadwick: Its the change of scale with latitude on 3857 that is the problem. This is adjusted for by e.g. the scale line control. The existing measure example gives the same answers for both its methods zoomed in near the equator. Its the change of scale with latitude that make the two methods give different results not the difference between calculating on the sphere or locally projected plane. The same issue applies to e.g. setting the circle radius.

bill-chadwick:
I think we may need either the proposed has constantPointResolution or the
existing custom getPointResolution function to properly handle UTM BNG ...
for accurate graticule/grid drawing for example.

The new ol.proj  getLength and getArea methods could document that the
ol.geom methods are good for UTM, BNG etc. but not 3857 and 4326

probins:
I think the proposed changes remove the issue, but in general I don't think it's a good idea for code to refer to specific EPSG codes. Agreed, 3857 is very widely used, but there are other codes based on Mercator, which will suffer from the same issues. Docs would be better saying 'projections such as 3857' or similar. Same applies to coordinate systems other than 4326.

examples/measure.html
probins: I think the existing comment needs changing (the whole point of most projections is that they do take earth curvature into account), though that can be done in a separate PR.

bill-chadwick: With the proposed change, for UTM and BNG etc, the ol.geom methods and
proposed ol.proj methods will give identical results. If a user asks for
length between (0,0) and (1000,0) with such coordinates then an answer of
1000 is what is expected.

bill-chadwick:
The proposed ol.proj methods can be made to always return an answer in metres or metres*metres - the docs can note the only exception is that the answer will be in pixels for pixel projections.  Users can convert the returned metres to miles, yards, kilometers, nautical miles, degrees or whatever as they wish. 

I guess one issue with a constantPointResolution projection attribute is that folk need to know when to set it ! If we don't have such an attribute, then I don't think its possible to create a best accuracy + performance, 'one stop shop' for getLength and getArea. The alternative is that users need to know that the ol.geom methods will give better answers  than the ol.proj methods for the kind of projections typically used for topographic mapping, such as UTM and BNG . Conversely users need to know that  the ol.geom methods are no good for 3857, 4326 and similar projections when the proposed ol.proj methods should be used instead. That information can of course be written into the documentation for both sets of functions.

I think it would make sense to add a scale bar control to the measure example.

bill-chadwick:
I propose to delete this PR in favour of https://github.com/openlayers/ol3/pull/4372

PR #4313, opened 2015-10-25
Align logos nicely with attribution text

ahocevar:
Currently, when not using Bootstrap and a css like the one we use in the examples, logos in the attribution do not align nicely with the attribution text.

Before:
before

After:
after

### DIFF
--- a/css/ol.css
+++ b/css/ol.css
@@ -159,6 +159,7 @@
 .ol-attribution img {
   max-height: 2em;
   max-width: inherit;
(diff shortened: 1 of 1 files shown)


bartvde:
LGTM @ahocevar

marcjansen:
This has bugged me quite a while! Thanks @ahocevar !

tschaub:
Three years of work, and we have aligned the logo!!  Thanks @ahocevar.

PR #4316, opened 2015-10-25
Render name labels if the geometry is a point in KML format (@tamarmot)

marcjansen:
This is a follow-up PR for #4306. With a fixed build (`gjslint` failed) and a corrected color. I also changes the commit messages to be more in line with our recommendations.

Thanks for the continued work on this, @tamarmot

### DIFF
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -11,7 +11,6 @@ goog.require('goog.array');
 goog.require('goog.asserts');
 goog.require('goog.dom.NodeType');
 goog.require('goog.object');
-goog.require('goog.string');
 goog.require('ol');
 goog.require('ol.Feature');
 goog.require('ol.FeatureStyleFunction');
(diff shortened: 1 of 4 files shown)


src/ol/format/kmlformat.js
```diff
@@ -287,16 +306,66 @@ ol.format.KML.ICON_ANCHOR_UNITS_MAP_ = {
...
+      // Offset the label to be centered to the right of the icon, if there is
+      // one.
+      textOffset[0] = foundStyle.getImage().getScale() * imageSize[0] / 2;
+      textOffset[1] = -foundStyle.getImage().getScale() * imageSize[1] / 2;
+      textAlign = 'left';
+    }
+  }
+  if (!goog.object.isEmpty(foundStyle.getText())) {
+    textStyle = /** @type {ol.style.Text} */
+        (goog.object.clone(foundStyle.getText()));
```
tschaub: The use of `goog.object.isEmpty()` and `goog.object.clone()` should be avoided here.  See #4781.

PR #4321, opened 2015-10-26
Added wrapX override support for ClusterOptions

### LINKED ISSUE #4319: wrapX: false wont work on clustering source
I had a single map working with all my features (because of the wrapX option). As soon as I changed my source from Vector() to Cluster() it showed multiple features, as if there are more worlds to show. Strangely it only affects the source layer and not the map layer.
This have to be a bug, cause the docs say: Cluster() extends Vector() so wrapX has to be available in my opinion..

```javascript

... more code above here ...

var sourcePhotos = new ol.layer.Vector({
    source: new ol.source.Cluster({
        distance: 20,
            url: '/photo/markers', // works perfectly
        }),
        wrapX: false // <--- its there
    }),
    style: function (feature, resolution) {
        var size = feature.get('features').length;

        style = [new ol.style.Style({
                image: new ol.style.Icon(/** @type {olx.style.IconOptions} */ {
                    anchor: [8, 8],
                    anchorXUnits: 'pixels',
                    anchorYUnits: 'pixels',
                    src: '/img/camera.png'
                }),
                text: new ol.style.Text({
                    text: size.toString(),
                    fill: new ol.style.Fill({
                        color: '#000'
                    }),
                    offsetX: 18
                }),
            })];
        return style;
    }
});

.. more code below here ...
```
bug-1

FRizZL:
Fixes #4319

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3684,7 +3684,9 @@ olx.source.BingMapsOptions.prototype.wrapX;
  *     format: (ol.format.Feature|undefined),
  *     logo: (string|undefined),
  *     projection: ol.proj.ProjectionLike,
- *     source: ol.source.Vector}}
+ *     source: ol.source.Vector,
+ *     wrapX: (boolean|undefined)
+ }}
  * @api
  */
 olx.source.ClusterOptions;
@@ -3745,6 +3747,12 @@ olx.source.ClusterOptions.prototype.projection;
  */
 olx.source.ClusterOptions.prototype.source;
 
+/**
+ * WrapX. Default is true
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.source.ClusterOptions.prototype.wrapX;
 
 /**
  * @typedef {{preemptive: (boolean|undefined),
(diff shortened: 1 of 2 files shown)


ahocevar:
You'll also need to [pass the new option to the superclass](https://github.com/openlayers/ol3/blob/master/src/ol/source/clustersource.js#L27-L32).

FRizZL:
Thanks for letting me know!

ahocevar:
Looks good now, thanks! Before we can merge this, we need a [CLA](https://github.com/openlayers/cla) from you. If you use the online form, that should be a matter of 2 minutes.

ahocevar:
Oh, and your commit messages are not [clear](https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#use-clear-commit-messages), as demanded in our [pull request guidelines](https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#pull-request-guidelines). Ideally, you could squash the commits into one. If you're unable to, I can do it before merging, but that might take a while.

FRizZL:
I try to help you here with the commit messages. I already accepted the agreement :-)

PR #4322, opened 2015-10-26
Fixes: #4319 - added wrapX override support for ol.source.Cluster

### LINKED ISSUE #4319: wrapX: false wont work on clustering source

FRizZL:
It was not possible to override the wrapX option for the cluster source.
This commit adds that functionality.

Fixes #4319.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3684,7 +3684,8 @@ olx.source.BingMapsOptions.prototype.wrapX;
  *     format: (ol.format.Feature|undefined),
  *     logo: (string|undefined),
  *     projection: ol.proj.ProjectionLike,
- *     source: ol.source.Vector}}
+ *     source: ol.source.Vector,
+ *     wrapX: (boolean|undefined)}}
  * @api
  */
 olx.source.ClusterOptions;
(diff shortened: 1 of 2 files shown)


FRizZL:
Andreas, I had to remove my fork and create a new one. Here is the change in one full commit.

ahocevar:
Merged 05dfc226b788e5cad75264d3e1b317fcc8a1f87c.

ahocevar:
Thanks @FRizZL.

PR #4327, opened 2015-10-28
Allow GeoJSON features with id equal to 0

### LINKED ISSUE #4326: Zero value of feature ID is changed to Undefined
One of my rendered feature has intentionally ID set to 0. Since 3.10.1 when this feature is retrieved using vectorLayer.getSource().getFeatures(), that zero value ID is converted into 'Undefined'. Should I avoid such values or is it a newly introduced bug in OL?

fredj:
fixes #4326


bartvde:
LGTM @fredj please merge if Travis gives the ok

PR #4330, opened 2015-10-28
ol.format.TopoJSON is read only, update the jsdoc

fredj:


bartvde:
LGTM

fredj:
thanks for the review

PR #4332, opened 2015-10-28
Fix write out GeoJSON features with id equal to 0

fredj:


bartvde:
LGTM thanks for fixing all these

fredj:
Thanks for the review

PR #4334, opened 2015-10-28
Check view extent within image extent

### LINKED ISSUE #4325: Single tile WMS re-requesting layer after every pan
Since 3.8 if you pan the map, even a small amount, any single tile wms layers are re-requested.

This one is working fine:
http://openlayers.org/en/v3.8.2/examples/wms-tiled.html

This re-requests the wms on every pan:
http://openlayers.org/en/v3.10.1/examples/wms-image.html

Thanks

Stewart

ahocevar:
When checking whether a view extent is within the buffered WMS image extent, we must not use a buffered view extent.

Fixes #4325.


marcjansen:
Looks good to me. Nice fix. Please merge.

Issue #4288, opened 2015-10-18
Replace Resemble.js with Pixelmatch


mourner:
Thought you guys should know since you use Resemble.js for rendering tests — I just created a much much better image comparison library, https://github.com/mapbox/pixelmatch. Check it out! We're now using it for Mapbox GL rendering tests and it works very well.

cc @tsauerwein @tschaub @elemoine @marcjansen

tschaub:
Nice.  Thanks @mourner.

ahocevar:
Planning to work on a pull request for this. Who's in favour?

tsauerwein:
> Planning to work on a pull request for this. Who's in favour?

It's definitely worth a try!

bartvde:
+1

marcjansen:
BTW: Am I supposed to run the current rendering test suite with the following command (on `master`):

```
node tasks/test-rendering.js
```

? Because if I do that (and it may be just me) on my machine, I get:

![image](https://cloud.githubusercontent.com/assets/227934/10943466/c7fed0c6-8314-11e5-9e12-e0f223d210de.png)

And on the command line it says:

```
Gecko error: it seems /path/to/ol3/node_modules/slimerjs-edge/lib/slimer/xulrunner/xulrunner is not compatible with SlimerJS. See Gecko version compatibility.
```

Can anyone reproduce? If so I#ll create an issue.

tsauerwein:
Can you try with `make test-rendering`?

marcjansen:
Notat the moment, will do once I have access to a computer again.

marcjansen:
It now works for me regardless of I do `node tasks/test-rendering.js` or `make test-rendering`. Not sure what broke it.

marcjansen:
🎉🎉🎉 cool, we finally switched!

PR #4339, opened 2015-10-29
Make sure drawImage width and height are not too big

### LINKED ISSUE #4337: svg icon in image style not drawn in Safari if offset is used
This is similar to: http://stackoverflow.com/questions/29633355/canvas-context-drawimage-differences-across-browsers

Changing the draw instruction to:

```
         context.drawImage(image, originX, originY, width - originX, height - originY,
                x, y, width * pixelRatio, height * pixelRatio);
```

fixes this issue and the icon is drawn correctly.

bartvde:
in the case of an offset.

fixes #4337


ahocevar:
It would be nice to have a render test for this.

bartvde:
But the problem is only visible in Safari? So how would a render test help?

ahocevar:
Does this fix break things on other browsers?

In theory, if you create the render test reference image in a different browser, and run the test in Safari, it should fail. In practice, the resemble.js image comparison is often too picky. Maybe the render tests should use Pixelmatch instead. See #4288.

bartvde:
No it's still fine on other browsers (I checked Chrome and FF).

There is still no way to run the render tests on OSX right?

ahocevar:
It would at least be nice to have a passing render test before this change, to make sure that it doesn't break things on other browsers.

bartvde:
Agreed, okay I'll add one.

ahocevar:
> There is still no way to run the render tests on OSX right?

They run fine in the browser (http://localhost:3000/test_rendering/. In El Capitan with latest XCode, they also run with `make test-rendering` now.

bartvde:
Ah great, that's good to know.

bartvde:
@ahocevar and they currently pass all for you on OSX? I get 5 failures:

```
Barts-MacBook-Air:ol3_bartvde bartvandeneijnden$ make test-rendering
node tasks/test-rendering.js
Gecko: 33.1.1
ol.rendering.layer.VectorTile vector tile layer renders correctly with the canvas renderer
Error: Error: expected '2.12' to be below 1.5 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)

ol.rendering.reproj.Tile tile projection from EPSG:4326 to EPSG:3857
Error: Error: expected '7.42' to be below 1.5 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)

ol.rendering.reproj.Tile dateline wrapping wraps X when prime meridian is shifted
Error: Error: expected '6.20' to be below 1.5 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)

ol.rendering.reproj.Tile dateline wrapping displays north pole correctly (EPSG:3413)
Error: Error: expected '2.69' to be below 1.5 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)

ol.rendering.style.RegularShape #render tests the canvas renderer
Error: Error: expected '6.36' to be below 6 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)

5 test(s) failed.
```

bartvde:
Added a rendering test for an icon with an offset, let's see if Travis also thinks it's okay.

bartvde:
Hmm, but of course my change does influence the resulting image. So a rendering image created before this change is different.

bartvde:
So this needs more work.

bartvde:
Okay have a fix now that results in the same reference image before and after this change.

bartvde:
@ahocevar this is now really ready for review.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -343,8 +343,9 @@ ol.render.canvas.Replay.prototype.replay_ = function(
             context.globalAlpha = alpha * opacity;
           }
 
-          context.drawImage(image, originX, originY, width, height,
-              x, y, width * pixelRatio, height * pixelRatio);
+          context.drawImage(image, originX, originY, width - originX,
+              height - originY, x, y, (width - originX) * pixelRatio,
+              (height - originX) * pixelRatio);
```
marcjansen: Very very minor: For readability it'd be nice if thes eypressions would be variables. No must, of course

marcjansen:
I added in comment which you may want to take. Otherwise this looks good to me. I'd favor if @ahocevar looked at this as well.

Thanks, Bart.

src/ol/render/canvas/canvasreplay.js
bartvde: Yes makes sense, will do.

ahocevar:
Looking at this now.

ahocevar:
Looks great now, thanks @bartvde. Please merge.

PR #4344, opened 2015-10-30
Allow different resolutions in x and y direction for ol.source.ImageStatic

### LINKED ISSUE #2844: Incorrect rendered Extent of ImageStatic
If I add an ImageStatic to a Map, its rendered too wide (in horizontal direction). Its maxx (right extent) appears at the false position.

Example: http://jsfiddle.net/u7xyor0t/1/

The Image is intended to reach from 8° to 9° East. But its right border appears at about 9.5° E... (you can check this with the Mouse Position)

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/5296274-incorrect-rendered-extent-of-imagestatic?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bartvde:
fixes #2844 based on the work references in the e-mail to ol3-dev there


src/ol/imagebase.js
```diff
@@ -53,7 +53,7 @@ ol.ImageBase = function(extent, resolution, pixelRatio, state, attributions) {
 
   /**
    * @protected
-   * @type {number|undefined}
+   * @type {Array.<number>|number|undefined}
```
ahocevar: Looks like you could remove a lot of duplicated `if (goog.isArray(imageResolution))` code if you'd always have an `Array.<number>` here.

bartvde: Agreed.

src/ol/imagebase.js
```diff
@@ -24,7 +24,7 @@ ol.ImageState = {
  * @constructor
  * @extends {goog.events.EventTarget}
  * @param {ol.Extent} extent Extent.
- * @param {number|undefined} resolution Resolution.
+ * @param {Array.<number>|undefined} resolution Resolution.
```
ahocevar: Maybe add something like "First value is the x-resolution, second the y-resolution".

bartvde: Good point.

src/ol/source/imagestaticsource.js
```diff
@@ -26,10 +26,19 @@ ol.source.ImageStatic = function(options) {
...
+    xResolution = ol.extent.getWidth(imageExtent) / options.imageSize[0];
+    yResolution = ol.extent.getHeight(imageExtent) / options.imageSize[1];
+    if (xResolution !== yResolution) {
+      if (xResolution > yResolution) {
+        resolutions = [xResolution, yResolution];
+      } else {
+        resolutions = [yResolution, xResolution];
+      }
+    } else {
+      resolutions = [yResolution];
```
ahocevar: Shouldn't this be `[yResolution, yResolution]`?

src/ol/source/imagestaticsource.js
```diff
@@ -26,10 +26,19 @@ ol.source.ImageStatic = function(options) {
...
   if (options.imageSize !== undefined) {
-    resolution = ol.extent.getHeight(imageExtent) / options.imageSize[1];
-    resolutions = [resolution];
+    xResolution = ol.extent.getWidth(imageExtent) / options.imageSize[0];
+    yResolution = ol.extent.getHeight(imageExtent) / options.imageSize[1];
+    if (xResolution !== yResolution) {
+      if (xResolution > yResolution) {
+        resolutions = [xResolution, yResolution];
+      } else {
+        resolutions = [yResolution, xResolution];
```
ahocevar: Why would you want to do this?

src/ol/source/imagestaticsource.js
bartvde: Right, it should.

src/ol/source/imagestaticsource.js
bartvde: Otherwise I get an assertion error about the order of resolutions

src/ol/source/imagestaticsource.js
ahocevar: No, it should not. That's a resolutions array. I think this needs more work.

src/ol/source/imagestaticsource.js
ahocevar: Yes, because this is a resolutions array, not the two resolutions that make up the aspect ratio. I think you can either use `Math.min` or `Math.max` here.

src/ol/source/imagestaticsource.js
bartvde: I think I should separate out the resolutions array from the image resolution constructor param.

src/ol/source/imagestaticsource.js
bartvde: What about: https://github.com/bartvde/ol3/commit/c12fbe9ff37a5a133919a3faade09b22e25be860 ?

src/ol/imagebase.js
bartvde: Addressed with: https://github.com/bartvde/ol3/commit/0b51e263801bde72133fb0e2023e4eceb8a37b76

src/ol/source/imagestaticsource.js
```diff
@@ -26,10 +26,15 @@ ol.source.ImageStatic = function(options) {
...
 
-  var resolution, resolutions;
+  var xResolution, yResolution, resolutions;
   if (options.imageSize !== undefined) {
-    resolution = ol.extent.getHeight(imageExtent) / options.imageSize[1];
-    resolutions = [resolution];
+    xResolution = ol.extent.getWidth(imageExtent) / options.imageSize[0];
+    yResolution = ol.extent.getHeight(imageExtent) / options.imageSize[1];
+    if (xResolution !== yResolution) {
+      resolutions = [Math.min(xResolution, yResolution), Math.max(xResolution, yResolution)];
```
ahocevar: So with this, you get two native resolutions. Is this intentional?

bartvde: It was in the original patch and I didn't really think it through, let me see if this is actually needed, probably not indeed.

bartvde: This was not needed indeed.

bartvde:
@ahocevar I think this is now truly ready, thanks for your comments already

ahocevar:
Yep, looks right now. Thanks, and please merge!

PR #4347, opened 2015-10-30
Use require.resolve() to find jsdoc

### LINKED ISSUE #4221: Custom build fails with npm 3
Cause of the introduction of a flat dependency hierarchy the path of jsdoc has changed.

The problem occurs in `tasks/generate-info.js:17`.

```
var jsdoc = path.join(__dirname, '..', 'node_modules', '.bin', 'jsdoc');
```

needs to be changed to

```
var jsdoc = path.join(__dirname, '..', '..', '.bin', 'jsdoc');
```

to work correctly with npm3.

Error message:

```
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: spawn /Users/mlegenhausen/dev/werk85/dk-navigator-app/node_modules/openlayers/node_modules/.bin/jsdoc ENOENT
    at exports._errnoException (util.js:849:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:178:32)
    at onErrorNT (internal/child_process.js:344:16)
    at doNTCallback2 (node.js:430:9)
    at process._tickCallback (node.js:344:17)
```

tschaub:
When the `openlayers` package is installed as a dependency of another package, `jsdoc` might be installed as a sibling (with npm@3).  Instead of relying on `jsdoc` being in our `node_modules`, we can use the `require.resolve()` lookup to resolve the path to the `jsdoc` executable.

Fixes #4221.

### DIFF
--- a/tasks/generate-info.js
+++ b/tasks/generate-info.js
@@ -14,11 +14,13 @@ var externsPaths = [
   path.join(externsDir, 'geojson.js')
 ];
 var infoPath = path.join(__dirname, '..', 'build', 'info.json');
-var jsdoc = path.join(__dirname, '..', 'node_modules', '.bin', 'jsdoc');
+
+var jsdocResolved = require.resolve('jsdoc/jsdoc.js');
+var jsdoc = path.resolve(path.dirname(jsdocResolved), '../.bin/jsdoc');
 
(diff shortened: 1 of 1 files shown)


marcjansen:
Nice. Please merge.

PR #4353, opened 2015-10-31
Resolve path to jsdoc-fork

tschaub:
It looks like [the build](https://travis-ci.org/openlayers/ol3/builds/88557316) for 85ca4e18888f32ea853d0a47249612807317a0ea only passed because we are using a cached `node_modules` on Travis (where both the `jsdoc` and `jsdoc-fork` packages were found).  I've cleared our Travis cache, so this build will only pass if a clean install works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
 
 before_install:
   - "npm install -g npm && npm install"
+  - "npm prune"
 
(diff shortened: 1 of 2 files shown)


ahocevar:
Looks good, please merge when it passes on Travis.

PR #4355, opened 2015-11-01
Update PhantomJS

tschaub:
The phantomjs@1.9.10 package is hanging while installing PhantomJS with Node 4 and npm 3.  This looks to be fixed with phantomjs@1.9.18.

The mocha-phantomjs-core package is a subset of mocha-phantomjs with just the parts we need.


tschaub:
This also changes our Travis config to use `node_js` as the language.  This gives us control over the version of Node used to run tests.  To install the one remaining Python dependency (`gjslint`), `sudo` is used.  We can use the new [Trusty setup](http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/), which looks to run our tests faster than the container-based infrastructure - but it looks like we can't use a cached `node_modules`, so the additional install time makes the overall run time about the same as before.

ahocevar:
Very nice. Please merge.

elemoine:
Nice.

PR #4356, opened 2015-11-01
Update Metalsmith and layouts plugin

tschaub:
This upgrades Metalsmith and uses the `metalsmith-layouts` plugin instead of the deprecated `metalsmith-templates` plugin.

Metalsmith 2 doesn't run on Node 0.10.  Use [`nvm`](https://github.com/creationix/nvm) if you need to get a more recent version of Node.


ahocevar:
Thanks for fixing this. Please merge.

marcjansen:
Error message in the build examples script  still has 'template'. Otherwise good to go.

elemoine:
> Metalsmith 2 doesn't run on Node 0.10. Use nvm if you need to get a more recent version of Node.

I have Node 0.12.7 and I get this error:

```
[wrk12 ol3 (gpx $%)]$ make test
node tasks/test.js
/home/elemoine/src/ol3/node_modules/metalsmith/lib/index.js:190
Metalsmith.prototype.build = unyield(function*(){
                                             ^
SyntaxError: Unexpected token *
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/elemoine/src/ol3/node_modules/metalsmith/index.js:6:18)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
make: *** [test] Error 1
```

You mentioned Node 0.10 in your comment so I had assumed that 0.12 would work.

So what Node version should people use now?

tschaub:
Sorry for the hassle @elemoine.  On 0.12 you'd have to do this to build the examples:

    node --harmony tasks/build-examples.js

I know this is not convenient.  If you use nvm to install Node 4, things will work.

If there are other devs who can't upgrade for some reason, I'm not opposed to downgrading Metalsmith for 0.12 support.  Note that this doesn't limit where you can create a custom build, it only affects building the examples.

elemoine:
Thanks for the note. I finally installed Node 4, using nvm.

pgiraud:
I also have the problem running `make serve`.
I can confirm that running `node --harmony tasks/serve.js` works.

Would it be useful to add `--harmony` each time we're using the `node` command in the makefile?

tschaub:
@pgiraud this should be safe.  It would enable all ES features that the v8 team considers "in progress" (not yet stable).  Here's the list for v8 4.5.103.35 (which ships with Node 4.2.1):

```
$ node --v8-options | grep "in progress"
  --harmony_modules (enable "harmony modules" (in progress))
  --harmony_array_includes (enable "harmony Array.prototype.includes" (in progress))
  --harmony_regexps (enable "harmony regular expression extensions" (in progress))
  --harmony_proxies (enable "harmony proxies" (in progress))
  --harmony_sloppy (enable "harmony features in sloppy mode" (in progress))
  --harmony_unicode_regexps (enable "harmony unicode regexps" (in progress))
  --harmony_reflect (enable "harmony Reflect API" (in progress))
  --harmony_destructuring (enable "harmony destructuring" (in progress))
  --harmony_sharedarraybuffer (enable "harmony sharedarraybuffer" (in progress))
  --harmony_atomics (enable "harmony atomics" (in progress))
  --harmony_new_target (enable "harmony new.target" (in progress))
```

The potential risk in always using the `--harmony` flag is that we might make use of an unstable feature (maybe unintentionally?) that would break in a subsequent version.  I think this is a pretty low risk.

The alternative is to document that the development setup requires Node 4 (again, this is not a requirement for people creating custom builds, just building the examples).  For people that haven't clicked on the link yet, [nvm](https://github.com/creationix/nvm) makes Node version management extremely easy:

```
curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.29.0/install.sh | bash
nvm install 4
nvm alias default stable
```

This will install Node 4 and use it by default.  You can install any additional versions and switch between them as you choose.  Globally installed dependencies are installed per-version.  With locally installed (project) dependencies, you may need to re-install dependencies when you switch versions (in cases where you have native binaries installed).

PR #4360, opened 2015-11-02
Update getExtent return value jsdoc tag

fredj:
The extent is never `null`; if the source is empty, the extent is
`[Infinity, Infinity, -Infinity, -Infinity]`

rbush code: https://github.com/mourner/rbush/blob/v1.4.1/rbush.js#L132


elemoine:
> The extent is never undefined

`!` means not `null`.

Please merge.

fredj:
> `!` means not `null`.

I've corrected the PR description and the commit message.

elemoine:
Thanks. Please merge.

Issue #3229, opened 2015-02-12
Error when writing polygons to GPX format


jalik:
*Version: OpenLayers 3.1.1*

When serializing polygons using **ol.format.GPX.writeFeatures()**, I get the following error :

```TypeError: serializersNS[node.namespaceURI][node.localName] is undefined ol.js:83601```

Which refers to the code below :

```js
ol.xml.serialize = function(
    serializersNS, nodeFactory, values, objectStack, opt_keys, opt_this) {
  var length = (goog.isDef(opt_keys) ? opt_keys : values).length;
  var value, node;
  for (var i = 0; i < length; ++i) {
    value = values[i];
    if (goog.isDef(value)) {
      node = nodeFactory.call(opt_this, value, objectStack,
          goog.isDef(opt_keys) ? opt_keys[i] : undefined);
      if (goog.isDef(node)) {
        serializersNS[node.namespaceURI][node.localName]
            .call(opt_this, node, value, objectStack);
      }
    }
  }
};
```

And below the values from the line execution :

```js
serializersNS = Object { "null": Object, "http://www.topografix.com/GPX/1/0": Object, "http://www.topografix.com/GPX/1/1": Object }
node.namespaceURI = "http://www.topografix.com/GPX/1/1"
node.localName = undefined
```

So the **node.localName** initialization is missing somewhere... should be easy to fix, no ?

bartvde:
GPX does not support polygons

http://www.globalmapperforum.com/forums/suggestion-box/1761-support-polygon-gpx-export.html

jalik:
Thanks for your quick reply, I am sorry, I didn't know...

So what do you propose ?
- should it stay like this, throwing a meaningless error leaving the newbie thinking its a bug
- or maybe we can ignore polygons to avoid errors
- or display a more explicit error message (eg: polygons not supported)
- or anything else...

bartvde:
Those are good questions. At the very least we should handle this more gracefully I agree.

jalik:
Maybe the safer option would be to just display a custom error, because ignoring polygons could become a bug if GPX adds polygon support later.

bartvde:
yeah we normally use assertions that will show up in debug mode, would that work for you?

jalik:
Yes of course, it's better than the current behaviour.

elemoine:
Fixed witth https://github.com/openlayers/ol3/pull/4361.

PR #4361, opened 2015-11-02
Make GPX format not fail on unsupported geometries

elemoine:
Currently the GPX format's `writeFeatures` function raises a `TypeError` [*] if one of the features it is passed is a polygon feature (or any other type of geometries not supported by the format). With this patch non-supported geometries are just skipped.

[*] ```xml.js:723 Uncaught TypeError: Cannot read property 'call' of undefined```

### DIFF
--- a/src/ol/format/gpxformat.js
+++ b/src/ol/format/gpxformat.js
@@ -839,11 +839,13 @@ ol.format.GPX.GPX_NODE_FACTORY_ = function(value, objectStack, opt_nodeName) {
       'value should be an ol.Feature');
   var geometry = value.getGeometry();
   if (geometry) {
-    var parentNode = objectStack[objectStack.length - 1].node;
-    goog.asserts.assert(ol.xml.isNode(parentNode),
-        'parentNode should be an XML node');
-    return ol.xml.createElementNS(parentNode.namespaceURI,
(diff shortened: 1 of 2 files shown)


bartvde:
Fixed #3229

elemoine:
Thanks for pointing to the corresponding issue @bartvde. I took the "just skip unsupported geometries" approach, because I think it makes sense in most cases.

bartvde:
@elemoine would an assertion help people understand what's going on?

elemoine:
> @elemoine would an assertion help people understand what's going on?

Assertions are exceptions, they stop the code. I think it's too radical here. What we could possibly do is log a message when goog.DEBUG is set, but this is not something we do currently in the library. But I think this is too opinionated again. Another option would be to add a `failOnUnsupportedGeometries`, but I don't like it too much.

My use-case is the following: user draws points, lines, and polygons and then export them in a GPX or KML document. If the users drew polygons, we just want to skip them and still export a GPX for the lines and points he might have drawn as well. And if we really wants to warn the user when he attempts to export unsupported geometries we could do the test ourselves, at the application-level, before passing the features to the GPX format's `writeFeatures` function.

bartvde:
Okay I think you've persuaded me @elemoine

bartvde:
LGTM please merge

elemoine:
Thanks a lot @bartvde for the review and the good questions!

Issue #4348, opened 2015-10-30
Cannot override resetNorth on the ol.control.Rotate control


tamarmot:
I have a map using lunar polar stereographic projection for the south pole, and where my view is, North is not at rotation 0.  I want to use the rotation control to indicate north, and need to override the resetNorth method as well as the render method.  ResetNorth is not exposed as overridable.

Related to issue: https://github.com/openlayers/ol3/issues/4346

tschaub:
Fixed by #4362 (thanks @tamarmot).

PR #4362, opened 2015-11-02
Rotate control now takes optional resetNorth function

tamarmot:
patch for issue #4348, can now pass in a function for reset north for the rotate control.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1315,6 +1315,7 @@ olx.control.ScaleLineOptions.prototype.units;
  *     tipLabel: (string|undefined),
  *     target: (Element|undefined),
  *     render: (function(ol.MapEvent)|undefined),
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks for this @tamarmot.

If you have publicly accessible data, it would be great to see an example of your South Pole map!  We need more examples of interesting locations.

tamarmot:
Thanks Tim --

right now this particular implementation is not public, but for example we
are using this as our WMS base layer:
http://wms.lroc.asu.edu/lroc?projection=SP_STEREO&c_lon=-14.192&c_lat=-86.33&x=0&y=0&resolution=3474.8&layers=LUNA_WAC_GLOBAL

Their site is written using OpenLayers 2; ours uses ol3 ...

Tamar

PR #4368, opened 2015-11-03
Append KML placemark text style to existing styles

### LINKED ISSUE #4367: KML with style cannot be combined with showPointNames
If you try to visualize [this KML](http://rawgit.com/geoext/geoext3/master/examples/data/SanFranciscoPublicSchools-Points.kml) (e.g. in the [drag-and-drop-example](http://openlayers.org/en/master/examples/drag-and-drop.html)) you'll see an error: `Uncaught TypeError: (0 , d.g) is not a function`. 

This happens because the KML is parsed with defaults (e.g. `showPointNames` is true and the KML has a style). In [line 397](https://github.com/openlayers/ol3/blob/master/src/ol/format/kmlformat.js#L397 ) of kmlformat.js we return the stlyes then as a nested array, where the renderer only accepts an array of styles.

This is a regression from 15c1323ff95e058d7dfac930b6d42020449f2be7.

See als geoext/geoext3#104

I am working on a fix.

marcjansen:
This fixes the regression introduced with 15c1323ff95e058d7dfac930b6d42020449f2be7 which I reported in #4367.

Thanks for any review.

Fixes #4367

### DIFF
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -394,7 +394,7 @@ ol.format.KML.createFeatureStyleFunction_ = function(style, styleUrl,
           if (drawName) {
             nameStyle = ol.format.KML.createNameStyleFunction_(style[0],
                 name);
-            return [style, nameStyle];
+            return style.concat(nameStyle);
           }
           return style;
         }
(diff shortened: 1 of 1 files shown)


ahocevar:
Nice. Please merge when Travis is happy.

marcjansen:
I will, Andreas. Thanks for the review.

PR #4372, opened 2015-11-03
Useful getLength and getArea methods

bill-chadwick:
This is a new attempt at providing useful methods in ol.proj that give a LineString's length in metres and a Polygon's area in square metres.

The measure example is updated to use them. A scale bar is added to it.

Notes are added about when to use ol.proj.getArea, ol.proj.getLength, ol.geom.LineString.getLength and ol.geom.Polygon.getLength.

### DIFF
--- a/examples/measure.html
+++ b/examples/measure.html
@@ -1,10 +1,12 @@
 ---
 layout: example.html
 title: Measure example
-shortdesc: Example of using the ol.interaction.Draw interaction for creating simple measuring application.
+shortdesc: Example of using the ol.interaction.Draw interaction for creating a simple measuring application.
 docs: >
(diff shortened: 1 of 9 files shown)


bill-chadwick:
In retrospect ...

The new methods could be named getMetricArea and getMetricLength.

The getArea iterations might perform better as nested loops with .slice() used to copy coordinates for conversion to EPSG:4326.

bill-chadwick:
Would an approach more like OL2 be better  ?

It has a
getGeodesicLength(projection) method for LineString type geometries giving an answer in metres
and a
getGeodesicArea(projection) method for Polygon type geometries giving an answer in square metres

tschaub:
We have `getLength()` and `getArea()` functions in the `ol/sphere` module now.

PR #4375, opened 2015-11-04
Fixed documentation typo (ol.source.ImageWMS)

simonseyock:
the imageLoadFunction you can give the constructor of ol.source.ImageWMS is an ol.ImageLoadFunctionType not an ol.TileLoadFunctionType


ahocevar:
Thanks!

PR #3955, opened 2015-08-03
Serialize properly ol.style.Circle to KML

htulipe:
This PR fixes #3882.

While serializing `ol.geom.Point`, current KML format assumes to find an `IconStyle` which is not always the case,iIt could also be an `ol.style.Circle`

This PR allows to serialize an `ol.style.Circle`. It simply takes the `ol.style.Fill` color. It might not be the best serialization but at least it prevents errors and does not mess up the KML export.

PR #4376, opened 2015-11-04
Make KML format ignore image styles that aren't icons

### LINKED ISSUE #3882: Error creating KML when using Circle in image style
I am using a circle within an image element in a style for a Point when drawing and all works well until I try to render it out using KML.writeFeaturesNode(), and I get this error:

Uncaught TypeError: style.getSrc is not a function.

sty = new ol.style.Style({  
          image: new ol.style.Circle({  
                     radius: 12,                
                     fill: new ol.style.Fill({ color: "#990000" })
                 })
      }); 

e.feature.setStyle(sty);

        ...

kml = new ol.format.KML().writeFeatures(features);

I think this Is this a bug, with OL3, thinking it's an icon? It catches the error here:

ol.format.KML.writeIconStyle_ = function(node, style, objectStack) {
    var /** @type {ol.xml.NodeStackItem} */ context = {node: node};
    var properties = {};
--->var src = style.getSrc();<---
    var size = style.getSize();
    var iconImageSize = style.getImageSize();
    var iconProperties = {
        'href': src
    };

http://stackoverflow.com/questions/31212953/error-creating-kml-when-using-circle-in-image-style

elemoine:
As reported in https://github.com/openlayers/ol3/issues/3882, the KML format currently fails when attempting to serialize point features represented using image styles that are not icon styles.

This PR is an alternative to #3955. #3955 still creates an `<IconStyle>` when the point feature is represented as an image style that is not an icon.

I am more strict in this PR, not creating any `<IconStyle>` at all when the point feature is not represented with an icon style.

Fixes #3882.

### DIFF
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -2509,7 +2509,7 @@ ol.format.KML.writeStyle_ = function(node, style, objectStack) {
   var strokeStyle = style.getStroke();
   var imageStyle = style.getImage();
   var textStyle = style.getText();
-  if (imageStyle) {
+  if (imageStyle instanceof ol.style.Icon) {
     properties['IconStyle'] = imageStyle;
   }
   if (textStyle) {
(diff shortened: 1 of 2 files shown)


fredj:
LGTM

elemoine:
Thanks a lot for the review @fredj.

PR #4377, opened 2015-11-04
Fix rendering of mocha's special characters on travis

marcjansen:
This PR suggests to use a recent version of [`mocha-phantomjs`](https://www.npmjs.com/package/mocha-phantomjs) which fixes the output of the testsuite on travis.

The `tasks/test.js` had to be adjusted to directly use the `mocha-phantomjs` binary.

**Before:**

ol3-tests-on-travis-before

**After:**

ol3-tests-on-travis-after

Please review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "jquery": "2.1.4",
     "jshint": "2.8.0",
     "mocha": "2.3.3",
-    "mocha-phantomjs-core": "^1.2.1",
+    "mocha-phantomjs": "4.0.1",
     "mustache": "2.2.0",
     "phantomjs": "1.9.18",
     "proj4": "2.3.12",
--- a/tasks/test.js
+++ b/tasks/test.js
@@ -57,17 +57,25 @@ function runTests(includeCoverage, callback) {
       }
       var address = server.address();
       var url = 'http://' + address.address + ':' + address.port;
+
       var args = [
-        require.resolve('mocha-phantomjs-core'),
-        url + '/test/index.html'
+        '--ssl-protocol', 'any',
+        '--ignore-ssl-errors', 'true'
       ];
 
       if (includeCoverage) {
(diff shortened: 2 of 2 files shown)


marcjansen:
Coverage isn't gathered with this change. Investigating now.

marcjansen:
Coverage now also running.

marcjansen:
Thanks for any review.

ahocevar:
Wondering about why you switched back to mocha-phantomjs after 98c9b0eb3990ec2cdafe11a07ca63da2a5daa801. From the package description:

> Since 4.0, the phantomjs code now is in mocha-phantomjs-core. If you need full control over which phantomjs version to use and where to get it, or want to use it more programatically like a build system plugin, please use that package directly. This project is a node.js CLI around it.

So I guess there should be a better way than spawning a process, i.e. use mocha-phantomjs-core "more programmatically" somehow. Have you investigated this a bit maybe?

ahocevar:
The reason why I'm asking is because this looks like a character encoding issue, something like UTF-8 not being used by Travis.

marcjansen:
I didn't check the 'more programmatic' approach. That would probably be advisable, but I fear my node-foo (and time) are too limited to give this a shot right now. Maybe @tschaub can weigh in about how to proceed, if you don't mind.

marcjansen:
Looking at https://github.com/nathanboktae/mocha-phantomjs-core/blob/master/README.md it looks both interesting and doable. If you prefer a solution based on the core package, I wouldn't mind if we close this issue, @ahocevar. I am just not sure if I can come up with something quickly.

ahocevar:
Not sure if I find time, but maybe I can give it a crack this weekend.

ahocevar:
@marcjansen Please see #4390 for a more programmatic solution.

marcjansen:
It doesn't  get better than #4390.  Thanks.

PR #4378, opened 2015-11-04
Add a writeStyles option to KML format

elemoine:
This PR adds a `writeStyles` boolean option to `ol.format.KML`. If this option is set to `false` the format does not add styles to the generated KML document. This option complements the `extractStyles` we already have for *read*.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1796,7 +1796,8 @@ olx.format.IGCOptions.prototype.altitudeMode;
 /**
  * @typedef {{extractStyles: (boolean|undefined),
  *     defaultStyle: (Array.<ol.style.Style>|undefined),
- *     showPointNames: (boolean|undefined)}}
+ *     showPointNames: (boolean|undefined),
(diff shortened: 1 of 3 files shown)


externs/olx.js
```diff
@@ -1827,6 +1828,14 @@ olx.format.KMLOptions.prototype.defaultStyle;
 
 
 /**
+ * Write styles into KML. Default is `true`.
+ * @type {boolean|undefined}
+ * @api stable
```
fredj: `stable` because `extractStyles`is already `stable`?

elemoine: Yes, but I am happy to make this option experimental at this stage.

fredj: IMHO `stable` is fine

fredj:
LGTM

gberaudo:
Typo in the PR description: `If this option is set to true`: s/true/false/

elemoine:
Description updated. Thanks @gberaudo.

PR #4386, opened 2015-11-06
Include common.js in jsFiddle resources

tsauerwein:
Some examples use a function `common.getRendererFromQueryString()` which is defined in [common.js](https://github.com/openlayers/ol3/blob/master/examples/resources/common.js). When creating a jsFiddle for these examples, they do no work, because the common.js file is missing.

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -46,7 +46,7 @@
           <textarea class="hidden" name="css">{{ css.source }}</textarea>
           <textarea class="hidden" name="html">{{ contents }}</textarea>
           <input type="hidden" name="wrap" value="l">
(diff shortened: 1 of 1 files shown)


bartvde:
LGTM

ahocevar:
The examples should strip out lines that contain common.getRendererFromQueryString(). It was intentional that common.js is not included when creating a JSFiddle. Please don't merge now. Let's discuss why you want to support that in a JSFiddle.

tsauerwein:
Right now, examples using `common.getRendererFromQueryString()` do not work in jsFiddle and also the copy&paste version is not usable. I agree that it doesn't make much sense to use `getRendererFromQueryString()` on jsFiddle, but it does make sense for the copy&paste version.

We also could replace `common.getRendererFromQueryString()` with `'canvas'`, but then statements like the following would look confusing:

    var renderer = common.getRendererFromQueryString();
    if (!renderer) {
      renderer = 'webgl';
    }

https://github.com/openlayers/ol3/blob/master/examples/symbol-atlas-webgl.js#L110-L113

ahocevar:
Hm. Look at the simple.js example. The regex strips out the renderer config successfully there. In the symbol-atlas-webgl.js example it does not work because the regex does not pick it up the way it is written.

What about an alternative approach? See #4387.

ahocevar:
In addition to my comment above: If, for whatever reason, you want to include common.js in an individual example, you can add the following to its html handlebars template:
```
resources:
  - resources/common.js
```

tsauerwein:
Closing in favour of #4387

PR #4387, opened 2015-11-06
Add default argument to getRendererFromQueryString

ahocevar:
This is an alternative fix for #4386.


tsauerwein:
Thanks, I didn't know that this was already handled this way.  Your changes look good to me.

ahocevar:
Thanks for the review @tsauerwein. I'll see if I can come op with another pull request to fix the Travis problem before merging this one.

elemoine:
+1 for this change.

PR #4389, opened 2015-11-06
Smooth transitions on parameter changes

elemoine:
This PR is an alternative to https://github.com/openlayers/ol3/pull/3660 created by @thomasmoelhave. The code includes code and suggestions from @thomasmoelhave.

The code is not ready for review yet. I need to review it myself and add unit tests.

### DIFF
--- a/src/ol/renderer/canvas/canvastilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvastilelayerrenderer.js
@@ -316,19 +316,29 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame =
 
   var tmpExtent = ol.extent.createEmpty();
   var tmpTileRange = new ol.TileRange(0, 0, 0, 0);
-  var childTileRange, fullyLoaded, tile, tileState, x, y;
+  var childTileRange, fullyLoaded, tile, x, y;
+  var drawableTile = (
+      /**
+       * @param {!ol.Tile} tile Tile.
+       * @return {boolean} Tile is selected.
+       */
+      function(tile) {
+        var tileState = tile.getState();
+        return tileState == ol.TileState.LOADED ||
+            tileState == ol.TileState.EMPTY ||
+            tileState == ol.TileState.ERROR && !useInterimTilesOnError;
+      });
   for (x = tileRange.minX; x <= tileRange.maxX; ++x) {
     for (y = tileRange.minY; y <= tileRange.maxY; ++y) {
-
       tile = tileSource.getTile(z, x, y, pixelRatio, projection);
-      tileState = tile.getState();
-      if (tileState == ol.TileState.LOADED ||
-          tileState == ol.TileState.EMPTY ||
-          (tileState == ol.TileState.ERROR && !useInterimTilesOnError)) {
+      if (!drawableTile(tile) && tile.interimTile) {
+        tile = tile.interimTile;
+      }
+      goog.asserts.assert(tile);
+      if (drawableTile(tile)) {
         tilesToDrawByZ[z][ol.tilecoord.toString(tile.tileCoord)] = tile;
         continue;
       }
-
       fullyLoaded = tileGrid.forEachTileCoordParentTileRange(
           tile.tileCoord, findLoadedTiles, null, tmpTileRange, tmpExtent);
       if (!fullyLoaded) {
@@ -360,7 +370,7 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame =
   var origin = ol.extent.getTopLeft(tileGrid.getTileCoordExtent(
       [z, canvasTileRange.minX, canvasTileRange.maxY],
       tmpExtent));
-  var currentZ, index, scale, tileCoordKey, tileExtent, tilesToDraw;
+  var currentZ, index, scale, tileCoordKey, tileExtent, tileState, tilesToDraw;
   var ix, iy, interimTileRange, maxX, maxY;
   var height, width;
   for (i = 0, ii = zs.length; i < ii; ++i) {
--- a/src/ol/source/tileimagesource.js
+++ b/src/ol/source/tileimagesource.js
@@ -163,6 +163,36 @@ ol.source.TileImage.prototype.getTileCacheForProjection = function(projection) {
 
 
 /**
+ * @param {number} z Tile coordinate z.
+ * @param {number} x Tile coordinate x.
+ * @param {number} y Tile coordinate y.
+ * @param {number} pixelRatio Pixel ratio.
+ * @param {ol.proj.Projection} projection Projection.
+ * @param {string} key The key set on the tile.
+ * @return {ol.Tile} Tile.
+ * @private
+ */
+ol.source.TileImage.prototype.createTile_ =
+    function(z, x, y, pixelRatio, projection, key) {
+  var tileCoord = [z, x, y];
+  var urlTileCoord = this.getTileCoordForTileUrlFunction(
+      tileCoord, projection);
+  var tileUrl = urlTileCoord ?
+      this.tileUrlFunction(urlTileCoord, pixelRatio, projection) : undefined;
+  var tile = new this.tileClass(
+      tileCoord,
+      tileUrl !== undefined ? ol.TileState.IDLE : ol.TileState.EMPTY,
+      tileUrl !== undefined ? tileUrl : '',
+      this.crossOrigin,
+      this.tileLoadFunction);
+  tile.key = key;
+  goog.events.listen(tile, goog.events.EventType.CHANGE,
+      this.handleTileChange, false, this);
+  return tile;
+};
+
+
+/**
  * @inheritDoc
  */
 ol.source.TileImage.prototype.getTile =
@@ -176,7 +206,7 @@ ol.source.TileImage.prototype.getTile =
     var cache = this.getTileCacheForProjection(projection);
     var tileCoordKey = this.getKeyZXY(z, x, y);
     if (cache.containsKey(tileCoordKey)) {
-      return /** @type {!ol.Tile} */(cache.get(tileCoordKey));
+      return /** @type {!ol.Tile} */ (cache.get(tileCoordKey));
     } else {
       var sourceProjection = this.getProjection();
       var sourceTileGrid = this.getTileGridForProjection(sourceProjection);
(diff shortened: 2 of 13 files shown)


elemoine:
Example visible here: http://erilem.net/ol3/dynamic-params/examples/smooth-params.html

bartvde:
nice to see you working on this again @elemoine 
If you need a review next week or so, ping me.

elemoine:
@thomasmoelhave, is the WMTS service used in the `smooth-params` example reliable? And what about the token key used to access the service?

thomasmoelhave:
@elemoine the server is reliable, it is one of our production tileservers. The token is perpetual and was made for this specific purpose. It can be used in the examples going forward, also publicly.

elemoine:
Thanks a lot for your answer.

elemoine:
I worked on the code again today, to, among other things, add tests. One problem I have is I know that my code doesn't work when the tile layer is reprojected (because its projection is not the same as the view projection). Making my code work in that case too would require more work, that I am not sure I can right now.

@bartvde, @thomasmoelhave, feel free to take a look at my code, and tell me what you think about the "does not work with reprojected layers" limitation.

examples/smooth-params.html
```diff
@@ -0,0 +1,23 @@
...
+  <div class="span6">
+    <h4>Smooth</h4>
+    <div id="map_smooth" class="map"></div>
+  </div><br />
+  <div class="span6 offset3">
+    Sea-level:
+    <input id="slider" type="range" value="10" step="0.25" max="100" min="-1" />
+    <span id="theinfo"></span>
+  </div>
+</di>
```
bartvde: should be ```</div>```

examples/smooth-params.js
```diff
@@ -0,0 +1,126 @@
...
+    })
+  ]
+});
+
+var updateSourceDimension = function(source, sliderVal) {
+  source.updateDimensions({'threshold': sliderVal});
+  document.getElementById('theinfo').innerHTML = sliderVal + ' meters.';
+};
+
+updateSourceDimension(staticSource, 10);
```
bartvde: why not simply setting this value in the source dimensions?

bartvde: Ah I guess to have ```the info``` filled with the initial value?

examples/smooth-params.js
```diff
@@ -0,0 +1,126 @@
...
+  layer: 'hydrosheds:sea-levels',
+  format: 'image/png',
+  matrixSet: 'EPSG:3857',
+  attributions: attributions,
+  tileGrid: tileGrid,
+  style: 'default',
+  dimensions: {
+    'threshold': 100
+  },
+  dynamicDimensions: ['threshold']
```
bartvde: I'm not sure I like the term dynamicDimensions. Dynamic versus static does not really convey what this does IMHO?

src/ol/source/tileimagesource.js
```diff
@@ -163,6 +163,36 @@ ol.source.TileImage.prototype.getTileCacheForProjection = function(projection) {
 
 
 /**
+ * @param {number} z Tile coordinate z.
+ * @param {number} x Tile coordinate x.
+ * @param {number} y Tile coordinate y.
+ * @param {number} pixelRatio Pixel ratio.
+ * @param {ol.proj.Projection} projection Projection.
+ * @param {string} key Key.
```
bartvde: I think this parameter needs a bit more explaining, also why it needs to be set on the tile instance.

src/ol/source/tilesource.js
```diff
@@ -157,6 +157,14 @@ ol.source.Tile.prototype.getGutter = function() {
 
 
 /**
+ * @return {string} The dynamic parameters key.
```
bartvde: This needs more docs IMHO.

src/ol/structs/lrucache.js
```diff
@@ -230,6 +230,18 @@ ol.structs.LRUCache.prototype.pop = function() {
  * @param {string} key Key.
  * @param {T} value Value.
  */
+ol.structs.LRUCache.prototype.replace = function(key, value) {
+  var entry = this.entries_[key];
+  goog.asserts.assert(entry !== undefined, 'an entry exists for key %s', key);
+  this.get(key);
```
bartvde: get has this same assertion. Can you explain why we need to call ```get``` here?

src/ol/tile.js
```diff
@@ -44,6 +44,18 @@ ol.Tile = function(tileCoord, state) {
...
 
+  /**
+   * An "interim" tile for this tile. The interim tile may be used while this
+   * one is loading, for "smooth" transitions when changing dynamic params.
+   * @type {ol.Tile}
+   */
+  this.interimTile = null;
+
+  /**
+   * @type {string}
```
bartvde: could do with a bit more docs

bartvde:
@elemoine I don't think that not supporting tile reprojection needs to stand in the way of merging this PR. I've put a few small comments in the PR.

elemoine:
Thank you very much for your comments @bartvde! I didn't have time to look at them today, but I'll do it as soon as I can.

examples/smooth-params.js
elemoine: Right.

examples/smooth-params.js
elemoine: It may be clear already, but "dynamic dimensions" are dimensions that the user will change (with a time slider for example), and for which we will use interim tiles while new tiles are loading. But I agree with you when you say that the term doesn't really convey what this does. I just don't have a better name right now.

bartvde: Yes this was clear, but I also couldn't really think of a better name, though I did consider ```smoothDimensions``` briefly.

bartvde: Or maybe transitionDimensions?

bartvde: Normally @tschaub is good with names, so ping @tschaub ;-)

elemoine: We could also decide that dimensions are always "dynamic", and not introduce `dynamicDimensions` at all. But I decided against that when I initially worked on the patch, because I thought that there may be cases where displaying interim/old tiles is not at all acceptable.

bartvde: Agreed, I also thought this might make for a sensible default.

src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -317,18 +317,28 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame =
...
+      /**
+       * @param {!ol.Tile} tile Tile.
+       * @return {boolean} Tile is selected.
+       */
+      function(tile) {
+        var tileState = tile.getState();
+        return tileState == ol.TileState.LOADED ||
+            tileState == ol.TileState.EMPTY ||
+            tileState == ol.TileState.ERROR && !useInterimTilesOnError;
+      });
```
tschaub: Would it make sense to move this logic elsewhere?  For example, it might be nicer to be able to call `source.getTile()` and get the interim tile if the "main" one was not loaded (and yes, this would mean passing `useInterimTilesOnError` to `getTile`).

tschaub: I see below where this "drawable" check is needed twice.

src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -317,18 +317,28 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame =
...
       tile = tileSource.getTile(z, x, y, pixelRatio, projection);
-      tileState = tile.getState();
-      if (tileState == ol.TileState.LOADED ||
-          tileState == ol.TileState.EMPTY ||
-          (tileState == ol.TileState.ERROR && !useInterimTilesOnError)) {
+      if (!drawableTile(tile) && tile.interimTile) {
+        tile = tile.interimTile;
+      }
+      goog.asserts.assert(tile);
+      if (drawableTile(tile)) {
```
tschaub: Interesting that the DOM and WebGL separate the loaded/other case here.

src/ol/structs/lrucache.js
elemoine: `get` updates the LRU structure (`oldest`, `newest`, etc.).

Maybe this would be better:

```js
ol.structs.LRUCache.prototype.replace = function(key, value) {
  this.get(key);  // update `newest_`
  this.entries_[key] = value;
};
```

tschaub:
Great to see this work picked up again @elemoine.  I hope to make time for a proper review.  Before merging, I think it would be nice to update the example so it doesn't highlight the bad (regular) case.

src/ol/structs/lrucache.js
elemoine: I am also going to add a test for this new method.

src/ol/tile.js
elemoine: Yes, agree. Not easy to explain, but I'll try.

examples/smooth-params.js
elemoine: Do you mean that parameters should be dynamic by default? Or do you agree with adding a `dynamicDimensions` option (or one with a better name)?

bartvde: Parameters should be dynamic by default.

elemoine:
> Before merging, I think it would be nice to update the example so it doesn't highlight the bad (regular) case.

Indeed. I just changed the order. I am happy to try to make the example look nicer if there are suggestions. I have to admit that it's not the part I focused on.

src/ol/renderer/canvas/canvastilelayerrenderer.js
elemoine: I think it's related to the way the renderer draws "blank" tiles. I just reused the existing logic here.

examples/smooth-params.html
elemoine: Fixed.

examples/smooth-params.js
elemoine: So in that case we'd need introduce a `staticDimensions` variable.

src/ol/tile.js
elemoine: I've added more docs. I hope it clarifies a bit, but I am conscious it's not perfect.

elemoine:
@bartvde, I've taken your comments into account, but I am still not sure what to do with `dynamicDimensions` (or `staticDimensions`). I'll try to think more about it, but I am not sure I can come up with something better :)

marcjansen:
What about 'variableDimensions'?  Or 'changingDimensions'? Or 'movingDimensions'? Or 'flexibleDimensions'? 

From http://www.thesaurus.net/antonyms-for/Static & http://www.thesaurus.net/synonyms-for/Dynamic

marcjansen:
Volatile? Transient? Momentary? Impermanent? http://martinfowler.com/bliki/TwoHardThings.html

elemoine:
I gave this a bit more thinking. My new suggestion is to remove the `dynamicDimensions` option entirely, and make the new behavior (with interim tiles) non-configurable. If we want to make this configurable in the future we could add an `hardDimensions` option, with something like this for the documentation:

```js
/**
 * The list of dimensions that, when changed, invalidate the tiles in the cache. So when changing
 * one the "hard" dimensions the layer will disappear until new tiles are loaded.
 * @type {Array.<string>|undefined}
 */
olx.source.WMTSOptions.prototype.hardDimensions;
```

I'll modify my branch if you agree with making the new behaviour non-configurable.

bartvde:
I agree with making the new behaviour non configurable

elemoine:
Thanks for the quick response. I'll change my branch.

elemoine:
@bartvde, patch updated.

bartvde:
thanks @elemoine I'll have another look at it soon

src/ol/renderer/canvas/canvastilelayerrenderer.js
elemoine: > For example, it might be nicer to be able to call source.getTile() and get the interim tile if the "main" one was not loaded (and yes, this would mean passing useInterimTilesOnError to getTile).

I thought about that. But `manageTilePyramid` is another [user](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/layerrenderer.js#L320) of `getTile`.

examples/smooth-params.html
```diff
@@ -0,0 +1,18 @@
+---
+layout: example.html
+title: Smooth parameters example
```
bartvde: maybe we should now rename this example to ```wmts-dimensions.html``` instead?

elemoine: Yep, makes sense.

bartvde:
A put a small note on the example name, but LGTM @elemoine feel free to merge

elemoine:
I changed the name of the example. I also removed the `drawableTile` function and inlined the code in the DOM and WebGL renderers, where the `drawableTile` function was called only once. This is related to a comment @tschaub made.

thomasmoelhave:
It's great to see this idea get this much traction - thanks for the reworked patch @elemoine and for the reviews @bartvde and @tschaub! The new patch is significantly faster than your initial version.  If you are  interested in a little more thorough testing I will have time to look into updating our dev-version to this new version and test it out with our system next week. We have a large amount of layers with these types of parameters and if there is some hidden performance/feature regression from #3660 we are likely to find it.

elemoine:
Thanks @thomasmoelhave. More tests are always a good thing!

tschaub:
Nice work on this @elemoine.

thomasmoelhave:
Hi @elemoine et.al

Now that this has been merged I am not sure this pull request is the appropriate venue for this discussion, let me know if you want to move it elsewhere.

I tested  ol3 master and the dynamic WMTS dimension is significantly improved over the older unpatched OL3 version! However the performance is still significantly lower than our initial patch #3660 and this results in several rendering glitches. I adjusted our earlier example to use a window of the same size as the new dynamic WMTS dimension example to allow for an apples-to-apples comparison, so if you fire up `examples/wmts-dimensions.html` and compare it to http://scalgo.com/stuff/ol3/build/examples/smooth-params-big.html at the same window size you can compare them directly.

Zoom in on some area, e.g. the Netherlands and the eastern UK and drag the sea-level slider continuously, and at a decent speed, from one end of the slider range to the other. You should notice a  significant decrease in perceived performance. As a result, certain tiles "get stuck" for longer periods in the new version resulting in visible tile boundaries as shown in the attached figure. 
output
These glitches also appear in #3660, but they persist for much longer in this version. The combined result of this is a degradation of the user experience and the feeling that everything is slow. I have tested this in both Firefox and Chromium on Ubuntu 14.04.

At the moment I don't understand the structure of @elemoine's work to the extend that I can point to a particular thing that causes this issue, but it's probably worth looking into as the user experience could be better and this patch likely makes it much harder to apply the other patch at the same time. It might be something trivial, or something fundamental.

fredj:
> Now that this has been merged I am not sure this pull request is the appropriate venue for this discussion, let me know if you want to move it elsewhere.

You should open a new issue with the content of your comment. thanks

PR #4388, opened 2015-11-06
Quick-fix running tests with PhantomJS

ahocevar:
Fixes #4384.

I'll come up with a better (and more persistent) fix when I look into #4377, which is related.

PR #4390, opened 2015-11-06
Fix usage of mocha-phantomjs-core after update

ahocevar:
Fixes #4377.

At least that's my opinion. Let's see if Travis agrees.

If nothing else, it at least is a true fix that replaces the quick fix from #4388.


ahocevar:
Yep, Travis agrees.

marcjansen:
Nice. Please merge.

PR #4391, opened 2015-11-06
Pass null as forEachFeatureAtPixel layer arg for unmanaged layers

### LINKED ISSUE #4114: Draw interaction does not work as expect when there is a select interaction

ahocevar:
See https://github.com/openlayers/ol3/pull/4143#issuecomment-154518582.

Fixes #4114.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -46,6 +46,12 @@ but with additional css:
 
 With the introduction of true vector tile support, `ol.source.TileVector` becomes obsolete. Change your code to use `ol.layer.VectorTile` and `ol.source.VectorTile` instead of `ol.layer.Vector` and `ol.source.TileVector`.
 
+#### `ol.Map#forEachFeatureAtPixel` changes for unmanaged layers
+
+`ol.Map#forEachFeatureAtPixel` will still be called for unmanaged layers, but the 2nd argument to the callback function will be `null` instead of a reference to the unmanaged layer. This brings back the behavior of the abandoned `ol.FeatureOverlay` that was replaced by unmanaged layers.
+
+If you are affected by this change, please change your unmanaged layer to a regular layer by using e.g. `ol.Map#addLayer` instead of `ol.layer.Layer#setMap`.
+
 ### v3.10.0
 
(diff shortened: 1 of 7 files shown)


ahocevar:
@bartvde, you may want to have a look to see how this affects #4114.

bartvde:
#4114 is this still the same @ahocevar with this change
Assuming you mean if I still see the issue in your branch?

ahocevar:
@bartvde I meant you should look at my branch to see if it makes your life easier, so you don't need a code snippet like the one you posted in your application.

Your issue is unlikely to be fixed with this branch, but it might. Because now forEachFeatureAtPixel is only called once for features that are on multiple unmanaged layers.

ahocevar:
@bartvde I added a commit that fixes #4114. Please review.

bartvde:
thanks @ahocevar I can confirm that this now fixes my issue from #4114
I'll do a review of the code tomorrow morning if no one gets to it before then.

bartvde:
LGTM @ahocevar does this deserve a mention in the release notes?

ahocevar:
I can add that, yes.

ahocevar:
Done @bartvde.

bartvde:
LGTM @ahocevar please merge if Travis agrees again

PR #4393, opened 2015-11-10
Faster vector tiles

ahocevar:
With two very simple changes (one in example code only), we can significantly improve performance of the Mapbox vector tiles examples. Using integer stroke widths makes canvas strokes about 10 times faster, and not doing `goog.getUid()` on features for checking if they are in an empty `skippedFeaturesHash` saves 10% of the time spent in `renderFrame()`.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -246,6 +246,7 @@ ol.render.canvas.Replay.prototype.replay_ = function(
     goog.asserts.assert(pixelCoordinates === this.pixelCoordinates_,
         'pixelCoordinates should be the same as this.pixelCoordinates_');
   }
+  var skipFeatures = !goog.object.isEmpty(skippedFeaturesHash);
   var i = 0; // instruction index
   var ii = instructions.length; // end of instructions
   var d = 0; // data index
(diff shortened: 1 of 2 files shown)


ahocevar:
@bartvde, I think you'll like this.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -260,7 +261,8 @@ ol.render.canvas.Replay.prototype.replay_ = function(
       case ol.render.canvas.Instruction.BEGIN_GEOMETRY:
         feature = /** @type {ol.Feature|ol.render.Feature} */ (instruction[1]);
         var featureUid = goog.getUid(feature).toString();
-        if (skippedFeaturesHash[featureUid] !== undefined ||
+        if ((skipFeatures &&
+            skippedFeaturesHash[goog.getUid(feature).toString()]) ||
```
bartvde: why not use```featureUid``` instead?

bartvde: Ah I guess you wanted to get rid of the local var, Travis also complaining

bartvde:
hey @ahocevar if you get Travis to pass, this LGTM.
Thanks for improving the performance!

ahocevar:
Yep, forgot to remove the line that assigns `featureUid`. Not because I want to get rid of the local var, but because `goog.getUid()` is what takes so much time.

bartvde:
Makes complete sense. 

Today I was inspecting a feature in an application and noticed two internal google ids on the feature, but didn't look any further to see if this is unexpected (my first guess was that it was).

bartvde:
screen shot 2015-11-10 at 19 06 21

Here is a screenshot of a feature drawn by the Draw interaction.

bartvde:
@ahocevar just explained to me this is because of ol3-cesium

PR #4394, opened 2015-11-11
Allow ol.Object property update without notification

tschaub:
Add an optional notify argument to object set and setProperties methods.

If the argument is not provided, set and setProperties trigger an event.

If you update many objects or do so frequently and do not need notification, performance is improved by setting silent to true.

This is a squashed version of the commits in #4252.

### DIFF
--- a/src/ol/object.js
+++ b/src/ol/object.js
@@ -197,11 +197,11 @@ ol.Object.prototype.notify = function(key, oldValue) {
  * Sets a value.
  * @param {string} key Key name.
  * @param {*} value Value.
- * @param {boolean=} opt_silent Update property without triggering notification.
+ * @param {boolean=} opt_silent Update without triggering an event.
  * @api stable
  */
 ol.Object.prototype.set = function(key, value, opt_silent) {
-  if (opt_silent === true) {
+  if (opt_silent) {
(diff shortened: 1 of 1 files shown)


src/ol/object.js
```diff
@@ -210,12 +215,13 @@ ol.Object.prototype.set = function(key, value) {
  * Sets a collection of key-value pairs.  Note that this changes any existing
  * properties and adds new ones (it does not remove any existing properties).
  * @param {Object.<string, *>} values Values.
+ * @param {boolean=} opt_silent update propertie silently
```
elemoine: Capital U and final dot?

src/ol/object.js
```diff
@@ -223,12 +229,15 @@ ol.Object.prototype.setProperties = function(values) {
 /**
  * Unsets a property.
  * @param {string} key Key name.
+ * @param {boolean=} opt_silent update propertie silently
```
elemoine: propertie ->property

elemoine:
I just noted minor styling issues. Please merge when you think this is ready.

tschaub:
Thanks @elemoine.  Updated with 85f58b111a01cebbc332034cd0e87712ab2694cd.

PR #4401, opened 2015-11-12
Allow style function to return a style

### LINKED ISSUE #1838: when using style function on layer, it always needs to return an array
So if people just return an ol.style.Style object, things will not work. Is there a way to make it work with both an ol.style.Style object as well as an array of ol.style.Style objects?

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/1449510-when-using-style-function-on-layer-it-always-needs-to-return-an-array?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tschaub:
In addition to returning an array of styles, a style function should be able to return a single style.

Since things can't be optimized if at runtime a function does not return a consistent type, people are discouraged from sometimes returning an array and sometimes returning a single style.  But if you consistently return one or the other, this change doesn't make it any harder for runtime optimization.

This change also makes it easier for people to avoid extra array allocation when they only need to render with a single style.  In short, it makes things better for the common case, and doesn't change anything for the more complex case of an array of styles.

Fixes #1838.

### DIFF
--- a/examples/cluster.js
+++ b/examples/cluster.js
@@ -38,7 +38,7 @@ var clusters = new ol.layer.Vector({
     var size = feature.get('features').length;
     var style = styleCache[size];
     if (!style) {
-      style = [new ol.style.Style({
+      style = new ol.style.Style({
         image: new ol.style.Circle({
           radius: 10,
           stroke: new ol.style.Stroke({
@@ -54,7 +54,7 @@ var clusters = new ol.layer.Vector({
             color: '#fff'
           })
         })
-      })];
+      });
       styleCache[size] = style;
     }
     return style;
--- a/examples/drag-and-drop-image-vector.js
+++ b/examples/drag-and-drop-image-vector.js
@@ -19,7 +19,7 @@ goog.require('ol.style.Style');
 
 
 var defaultStyle = {
-  'Point': [new ol.style.Style({
+  'Point': new ol.style.Style({
     image: new ol.style.Circle({
       fill: new ol.style.Fill({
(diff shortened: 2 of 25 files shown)


tschaub:
This is ready for review.

If our example use is any indication, a style function returning an array of styles is not the most common use case.  Arrays of styles are nice, but we can do a better job supporting the common case.

cc @bartvde

bartvde:
happy to see this @tschaub ;-)

I'll review this today.

tsauerwein:
Good idea, I tripped on this several times. Your changes look good to me.

bartvde:
LGTM as well, I also wondered the same thing as @tsauerwein

fredj:
+1

tschaub:
@tsauerwein and @bartvde - regarding the extra array, there is really no need for it.  The underlying `ol.renderer.vector.renderFeature()` function takes a single `ol.style.Style` object.  The common use case is to use a single style object.  So there is no reason to have the extra array allocation (for every feature at every frame) just to normalize for the more advanced style case.

Despite the claims about support for a single style being bad for performance (see https://github.com/openlayers/ol3/issues/1838#issuecomment-37284332), we can actually handle the common and more advanced case efficiently.

PR #4404, opened 2015-11-12
Set correctly the opt_this parameter when writing a KML document

oterral:
This PR fixes a bug found since the `writeStyles` property has been introduced. The `this` wasn't set correctly in `writeDocument_` function so `this.writeStyles` was always `undefined` in ` writePlacemark_` function.

### DIFF
--- a/test/spec/ol/format/kmlformat.test.js
+++ b/test/spec/ol/format/kmlformat.test.js
@@ -1915,6 +1915,43 @@ describe('ol.format.KML', function() {
             '</kml>';
         expect(node).to.xmleql(ol.xml.parse(text));
       });
+
+      it('can write multiple features with Style', function() {
+        var style = new ol.style.Style({
+          fill: new ol.style.Fill({
+            color: 'rgba(12, 34, 223, 0.7)'
+          })
+        });
+        var feature = new ol.Feature();
+        feature.setStyle(style);
+        var feature2 = new ol.Feature();
+        feature2.setStyle(style);
+        var node = format.writeFeaturesNode([feature, feature2]);
+        var text =
+            '<kml xmlns="http://www.opengis.net/kml/2.2"' +
+            ' xmlns:gx="http://www.google.com/kml/ext/2.2"' +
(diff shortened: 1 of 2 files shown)


test/spec/ol/format/kmlformat.test.js
```diff
@@ -2443,6 +2480,8 @@ describe('ol.format.KML', function() {
             '  </Document>' +
             '</kml>';
         expect(node).to.xmleql(ol.xml.parse(text));
+        node = format.writeFeatures([feature1, feature2]);
```
bartvde: use writeFeaturesNode instead? So no ol.xml.parse is needed on node?

test/spec/ol/format/kmlformat.test.js
```diff
@@ -1915,6 +1915,43 @@ describe('ol.format.KML', function() {
...
       });
+
+      it('can write multiple features with Style', function() {
+        var style = new ol.style.Style({
+          fill: new ol.style.Fill({
+            color: 'rgba(12, 34, 223, 0.7)'
+          })
+        });
+        var feature = new ol.Feature();
+        feature.setStyle([style]);
```
bartvde: minor but no need for arrays here right?

test/spec/ol/format/kmlformat.test.js
```diff
@@ -1915,6 +1915,43 @@ describe('ol.format.KML', function() {
...
+        var feature2 = new ol.Feature();
+        feature2.setStyle([style]);
```
bartvde: same here

test/spec/ol/format/kmlformat.test.js
oterral: writeFeature calls writeFeaturesNode so I guess this test is not necessary.

bartvde: Agreed.

oterral:
Pr updated according to comments. Thx for the review @bartvde

bartvde:
thanks @oterral will merge when Travis gives the okay again

PR #4407, opened 2015-11-12
Use ratio when calculating ImageWMS width and height

### LINKED ISSUE #4403: WMS layer shift in >= 3.11.0
When adding WMS layer according to example at http://openlayers.org/en/v3.10.1/examples/wms-image.html with OpenLayer 3.11, the WMS is shifted. 

3.10. Works well

For testing, see 

http://jsfiddle.net/jachymc/cw30xa9z/

ahocevar:
Fixes #4403.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1315,6 +1315,7 @@ olx.control.ScaleLineOptions.prototype.units;
  *     tipLabel: (string|undefined),
  *     target: (Element|undefined),
  *     render: (function(ol.MapEvent)|undefined),
+ *     resetNorth: (function()|undefined),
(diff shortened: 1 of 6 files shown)


tschaub:
@ahocevar would you mind basing this on the [v3.11.x branch](https://github.com/openlayers/ol3/tree/v3.11.x) that I just pushed?

We can create the v3.11.1 release from that branch and then merge into master after releasing.

Thanks for digging into the regression.

ahocevar:
Ok tschaub, will do.

ahocevar:
See #4408.

Issue #4405, opened 2015-11-12
Vector rendering performance regression


ahocevar:
See https://groups.google.com/d/msgid/ol3-dev/56447d46.e8c9c20a.2003.32e8%40mx.google.com for a detailed report.

I'm looking into this now.

ahocevar:
This regression is caused by 979342091ac6dae4cbb8e5e8c187a6286ed10a28.

ahocevar:
Now looking into this. Unless @bartvde is looking already.

ahocevar:
There must be an equality check for the previous `/* {number} */ resolution` which always evaluates to true for a new `/* {Array} */` resolution somewhere. Because a new replay is generated for each frame.

bartvde:
I haven't looked as yet @ahocevar

ahocevar:
Ok @bartvde, then I'll continue digging.

bartvde:
I'm guessing http://openlayers.org/en/v3.11.0/examples/mapguide-untiled.html is broken as well due to the same change

tschaub:
Fixed with #4412.

PR #4408, opened 2015-11-12
Use ratio when calculating ImageWMS width and height

### LINKED ISSUE #4403: WMS layer shift in >= 3.11.0

ahocevar:
Fixes #4403.

### DIFF
--- a/src/ol/source/imagewmssource.js
+++ b/src/ol/source/imagewmssource.js
@@ -203,16 +203,8 @@ ol.source.ImageWMS.prototype.getImageInternal =
   var centerY = (extent[1] + extent[3]) / 2;
 
   var imageResolution = resolution / pixelRatio;
-
-  // Compute an integer width and height.
-  var width = Math.ceil(ol.extent.getWidth(extent) / imageResolution);
-  var height = Math.ceil(ol.extent.getHeight(extent) / imageResolution);
-
-  // Modify the extent to match the integer width and height.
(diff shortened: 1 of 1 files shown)


tschaub:
This looks like it corrects the single image WMS example (on a non-retina device at least).  It would be great to get in a regression test for this, but I think a patch release is more urgent.

Thanks for the fix.  Please merge.

ahocevar:
I was planning to add tests for this anyway, but decided to look into #4405, the other serious regression, first.

ahocevar:
BTW, fixed for Retina displays too.

tschaub:
Thanks @ahocevar.

The downside of using a branch as the base for patches is that the continuous-integration/travis-ci/pr job doesn't run.  I just pushed a change to our Travis config to make the continuous-integration/travis/push job run for branches named like `/^v.*\.x$/`.  Not as convenient since we don't get the commit status updated on a PR.  But we can check the build status periodically (e.g. https://travis-ci.org/openlayers/ol3/builds/90780614).

The alternative is to commit patches to master and then cherry-pick.  But that brings potential for conflicts.

Issue #4400, opened 2015-11-11
Overlay stopping event propagation conflicts with top level event handlers like Ember JS


bjornharrtell:
I've been using OpenLayers 3 with Ember JS recently and noticed that Ember JS cannot handle events if a component is used with `ol.Overlay`. I think the reason is because Ember JS expects events to bubble up to the top level.

The default is that `ol.Map` stops propagation of `ol.Overlay` events. The issue here is that events will never reach Ember JS and actions will not work.

If `ol.Overlay` is configured with `stopEvent: false` events will pass through to the top level and work with Ember JS but they will also be handled as map events by `ol.Map` which might not be wanted.

To be more friendly for top level event handlers perhaps `ol.Map` should filter out and ignore instead of stopping propagation.

bjornharrtell:
Another solution might be to be more restrictive in general what a map event is. Prototype PR #4411.

JulienPradet:
Hi,

Has this issue/PR any chance to be adressed in a near future ?
I've got this exact same usecase, except it's with React. And I believe that anyone actually using modern front-end libraries/framework will come up with the same problem.

Avoid using stopPropagation sounds like a good idea to me.
If it's not such a common use case, wouldn't it be possible to add a configuration like "stopMapEvent" instead of "stopEvent" that would trigger this behavior ?

bjornharrtell:
I have no short term plan to put more time into this issue.

In the particular case I had I came up with another design avoiding the use of an overlay.

PR #4411, opened 2015-11-12
Do not stop event propagation and ignore clicks except on direct viewport children

bjornharrtell:
This PR proposes to avoid stopping event propagation for reasons described in #4400. It's prototype code not ready for formal review but is rather up for discussion.


tschaub:

PR #4415, opened 2015-11-13
Scale StaticImage image to imageExtent

### LINKED ISSUE #2844: Incorrect rendered Extent of ImageStatic

ahocevar:
The imageSize property was only used to determine the scale. It's usage was never documented nor tested nor shown in an example, so I removed it because I considered its presence a bug.

Fixes #2844.


src/ol/source/imagestaticsource.js
```diff
@@ -42,16 +37,34 @@ ol.source.ImageStatic = function(options) {
...
-  this.image_ = new ol.Image(imageExtent, resolution, 1, attributions,
+  this.image_ = new ol.Image(imageExtent, undefined, 1, attributions,
       options.url, crossOrigin, imageLoadFunction);
+
+  goog.events.listen(this.image_, goog.events.EventType.CHANGE, function() {
+    if (this.image_.getState() == ol.ImageState.LOADED) {
+      var image = this.image_.getImage();
+      var resolution = ol.extent.getHeight(imageExtent) / image.height;
+      var pxWidth = Math.ceil(ol.extent.getWidth(imageExtent) / resolution);
+      var pxHeight = Math.ceil(ol.extent.getHeight(imageExtent) / resolution);
```
bartvde: minor but we're calling ```ol.extent.getHeight(imageExtent)``` twice here

bartvde:
Nice way to handle this @ahocevar much better.
LGTM.

src/ol/source/imagestaticsource.js
```diff
@@ -42,16 +37,34 @@ ol.source.ImageStatic = function(options) {
...
 
   /**
    * @private
    * @type {ol.Image}
    */
-  this.image_ = new ol.Image(imageExtent, resolution, 1, attributions,
+  this.image_ = new ol.Image(imageExtent, undefined, 1, attributions,
       options.url, crossOrigin, imageLoadFunction);
+
+  goog.events.listen(this.image_, goog.events.EventType.CHANGE, function() {
```
tschaub: There are now two listeners for `CHANGE`.  Is it important that this listener get called first?

What would you think about registering a single listener (`this.handleImageChange_`) and having that call the parent method (`this.handleImageChange`)?

externs/olx.js
```diff
@@ -4975,14 +4974,6 @@ olx.source.ImageStaticOptions.prototype.imageExtent;
 
 
 /**
- * Size of the image in pixels.
- * @type {ol.Size|undefined}
- * @api stable
- */
-olx.source.ImageStaticOptions.prototype.imageSize;
```
tschaub: Not sure if anybody was using it this way, but if anybody is using an SVG as an image source, the size needs to be set explicitly in IE.  Edge case perhaps, but it would be good to at least document in the upgrade notes that the image size can no longer be set.

ahocevar: Ok, I'll create a follow up pull request on Monday.

src/ol/source/imagestaticsource.js
ahocevar: Yes, it is important that this one has s called first.

externs/olx.js
ahocevar: Looks like it makes sense then to bring the imageSize property back. Will incorporate all of @tschaub's suggestions in the follow up pull request.

tschaub: @ahocevar your change looks like a definite improvement.  We should clarify exactly where the size property is needed.  I'm only guessing about the SVG/IE case because of a recent experience with an icon symbolizer.

ahocevar:
See #4426 for the follow-up pull request which addresses comments by @tschaub and @bartvde.

Issue #3567, opened 2015-04-15
Stop using gjslint.py


tschaub:
I'd like to move on from using `gjslint.py`.  [ESlint](http://eslint.org/) is a far more configurable and capable tool.  I'm going to start by replacing our use of JSHint with ESLint and then see if I can convince others that `gjslint.py` can be removed.

elemoine:
gjslint catches Closure-related things, such as a missing `@private` annotation for an underbar suffixed members. Do you think we just don't need a linter that catches Closure-related errors? Or do you think we can use ESLint plug-ins (if possible?) for Closure? Just asking...

probins:
Closure compiler's linter has just been separated out into its own jar. https://github.com/google/closure-compiler/commit/4782e1a9ea95bd25ae0dc10775be4f7f660fba7a Not sure how far this can be used as a substitute for gjslinter, though the comment that they'll be using it on the Closure lib itself sounds promising.

tschaub:
My editor is not configured to use `gjslint.py`, so I don't find out about it until `make check`, at which time I'd rather just let the Compiler catch any Compiler-specific issues.

I prefer the ESLint rules (and their configuration options) to what is enforced by `gjslint.py`, so I don't see a need for it in terms of coding conventions.  Rules are pretty straightforward to implement, so if we really feel there are things we want enforced, and there isn't a rule already, we could create one (and make it available to others who might share the same conventions).

The [`valid-jsdoc` rule](http://eslint.org/docs/rules/valid-jsdoc.html) provides some light documentation checking - which I think is sufficient for our needs.  We have very real problems with our documentation that are not being caught by `gjslint.py`, so I don't think our documentation quality would go down by getting rid of it.

In addition, I think ESLint brings a lot of rules that do catch real coding mistakes.  I can't think of a single example where `gjslint.py` has caught a real mistake, and it has not been uncommon for me to have ESLint correct errors before I even save a change.

> gjslint catches Closure-related things, such as a missing @private annotation for an underbar suffixed member

This is an interesting point.  If we were to ditch this convention of using underscore suffixed names for private members, we would not satisfy `gjslint.py`, but we could instead conform with the majority of the JavaScript development community - where underscore *prefixed* names are private.

bartvde:
I've also been using ESLint on a few projects and very happy with it

marcjansen:
We're also using it and have had good experience so far, especially the integration into editors is easy.

ahocevar:
[...]

PR #4420, opened 2015-11-14
Remove Bootstrap and jQuery from example snippets

tschaub:
People shouldn't have to load Bootstrap and jQuery to use OpenLayers.  This branch creates a new "strapless" template that doesn't include the extra libs in the example snippet (for copy/paste or edit on jsFiddle).

For examples that explicitly require Bootstrap and/or jQuery (e.g. `button-title.js`), these additional resources are listed in the YAML front-matter.

In addition, the new template indents content and uses newlines a bit more intentionally.  Here's an example of an updated snippet (I've taken the liberty of simplifying the simple example a bit):

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Simple example</title>
    <link rel="stylesheet" href="http://openlayers.org/en/v3.11.1/css/ol.css" type="text/css">
    <script src="http://openlayers.org/en/v3.11.1/build/ol.js"></script>
  </head>
  <body>
    <script>
      var map = new ol.Map({
        layers: [
          new ol.layer.Tile({
            source: new ol.source.OSM()
          })
        ],
        target: 'map',
        view: new ol.View({
          center: [0, 0],
          zoom: 2
        })
      });
    </script>
  </body>
</html>
```

I've pushed this to the to the `openlayers` org in case others want to help plow through the remaining examples.  After we've converted all, we can rename `strapless.html` back to `example.html`.


tschaub:
```
Running gjslint...
----- FILE  :  /home/travis/build/openlayers/ol3/examples/layer-spy.js -----
Line 27, E:0220: No docs found for member 'document.onkeydown'
Line 43, E:0220: No docs found for member 'viewport.onmousemove'
Line 48, E:0220: No docs found for member 'viewport.onmouseout'
Found 3 errors, including 0 new errors, in 1 files (515 files OK).
```

I'm so sick of `gjslint`.

bartvde:
One of the last hangouts it was suggested we can do PSC votes to get out of situations where there is no general agreement

marcjansen:
Is there not an agreememt on this change, @bartvde?

tschaub:
> Is there not an agreememt on this change, @bartvde?

We can discuss this on #3567 (I think we're pretty close to consensus - just need to make some time to switch to ESLint).

bartvde:
@marcjansen on this change yes, but not on the removal of eslint (I see @tschaub removed his original comment here, on which I reacted by e-mail ;-)). Sorry for the confusion.

tschaub:
Here was my original comment:

>  Running gjslint...
>  ----- FILE  :  /home/travis/build/openlayers/ol3/examples/layer-spy.js -----
>  Line 27, E:0220: No docs found for member 'document.onkeydown'
>  Line 43, E:0220: No docs found for member 'viewport.onmousemove'
>  Line 48, E:0220: No docs found for member 'viewport.onmouseout'
>  Found 3 errors, including 0 new errors, in 1 files (515 files OK).
>  I'm so sick of gjslint.

I decided this was only a distraction to leave on this pull request.

marcjansen:
Ok got it!

jonataswalker:
This is a great cleanup!

tschaub:
Ok @openlayers/ol3, this is ready for review.  It would be great to get this merged before other example changes break things.

For future example creators, there is no need to use jQuery unless you are demonstrating some jQuery specific integration - a jQuery plugin for example.  In addition, examples should use the minimum markup required and should not rely on Bootstrap's CSS (no more `pull-right alert alert success` please).

The new template pulls in [a polyfill](https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch) for [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) where needed.  So if you find yourself tempted to use `$.ajax()`, instead you can do this:
```js
fetch(url).then(function(response) {
  // typically
  return response.json();
}).then(function(json) {
  // deal with the data
}).catch(function(err) {
  // we always handle our errors
});
```

Thanks for any review.

tschaub:
To make for an easier review, I've (temporarily) published the examples here: http://tschaub.net/ol-strapless/examples/  (so you can compare http://tschaub.net/ol-strapless/examples/simple.html to http://openlayers.org/en/v3.11.1/examples/simple.html for example).

examples/accessible.html
```diff
@@ -1,6 +1,6 @@
 ---
 layout: example.html
-title: Accessibility example
+title: Acessible Map
```
marcjansen: Missing a `c`: A**c**cessible

examples/center.html
```diff
@@ -13,22 +13,16 @@
   <p>Use <code>Alt</code>+<code>Shift</code>+drag to rotate the map.</p>
```
marcjansen: Super minor: you have used different markup elsewhere. Nothing to worry about, but maybe we want to harmonize…

examples/simple.js
```diff
@@ -11,12 +10,6 @@ var map = new ol.Map({
       source: new ol.source.OSM()
     })
   ],
-  controls: ol.control.defaults({
-    attributionOptions: /** @type {olx.control.AttributionOptions} */ ({
-      collapsible: false
-    })
-  }),
```
marcjansen: I always thought that when we are using OSM we must have a non-collapsible attribution, right?

examples/translate-features.js
```diff
@@ -43,7 +30,7 @@ var translate = new ol.interaction.Translate({
 
 var map = new ol.Map({
   interactions: ol.interaction.defaults().extend([select, translate]),
-  layers: [raster, vector, vector2],
+  layers: [raster, vector],
```
marcjansen: :+1: on this example simplification

examples/vector-esri-edit.js
```diff
@@ -97,8 +95,12 @@ selected.on('remove', function(evt) {
       featureProjection: map.getView().getProjection()
     }) + ']';
     var url = serviceUrl + layer + '/updateFeatures';
-    $.post(url, { f: 'json', features: payload }).done(function(data) {
-      var result = JSON.parse(data);
+    var data = new FormData();
```
marcjansen: I think this wll not work in all target browser, right? We still support IE9, for example. https://developer.mozilla.org/docs/Web/API/FormData

examples/vector-esri-edit.js
```diff
@@ -117,8 +119,12 @@ draw.on('drawend', function(evt) {
     featureProjection: map.getView().getProjection()
   }) + ']';
   var url = serviceUrl + layer + '/addFeatures';
-  $.post(url, { f: 'json', features: payload }).done(function(data) {
-    var result = JSON.parse(data);
+  var data = new FormData();
```
marcjansen: See above

examples/wms-tiled-wrap-180.html
```diff
@@ -1,13 +1,9 @@
 ---
 layout: example.html
-title: Tiled WMS wrap 180° meridian example
+title: Tiled WMS Wrapping
 shortdesc: Example of a tiled WMS layer that wraps across the 180° meridian.
 docs: >
-  The `wrapX` option is set to `true` here so tiles will be wrapped around the x-axis.
+  By default, WMS tiles are reused across the 180° meridian.  The can be disabled by setting the `wrapX` option to `false`.
```
marcjansen: <strike>The</strike> <ins>This</ins> can be disabled

examples/xyz.js
```diff
@@ -1,41 +1,15 @@
...
 goog.require('ol.layer.Tile');
-goog.require('ol.source.OSM');
 goog.require('ol.source.XYZ');
 
 
-var attribution = new ol.Attribution({
-  html: 'Tiles &copy; <a href="http://maps.nls.uk/townplans/glasgow_1.html">' +
-      'National Library of Scotland</a>'
-});
-
```
marcjansen: I think we want to keep the attribution (and probably always show it)

marcjansen:
Very nice work, Tim. I have added some comments, which I hope you can adress. Most are minor but the attribution of OSM sources should IMO be tackled and I am unsure whether we can use `FormData` without a polyfill.

I tested several examples (mostly those were internal changes were made)  and al worked. I couldn't test all, though. I am very confdent, that they will work.

Thanks in any case for this huge amount of work!

examples/simple.js
tschaub: image

examples/xyz.js
tschaub: I also [removed the source](https://github.com/openlayers/ol3/pull/4420/files#r45166546) for which this attribution applies.

The problem with this example (and many others) is that people keep tacking additional functionality in without any good documentation.  This is our canonical XYZ example.  People shouldn't have to get puzzled by the weird `/** @type {olx.control.AttributionOptions} */` comment, the unexplainable `attributions` array (why an array with no apparent way to control which gets displayed when).

In addition, this example is ugly and generates a slew of 404s.

image

I'd like us to be more considerate of our users in examples (not just dump mysterious code without explanation).  This example isn't a particularly egregious violation of that, but we've got a long way to go to provide simple examples with good documentation, and I hope we don't have to debate every change headed in that direction.

examples/accessible.html
tschaub: Thanks for catching this @marcjansen.  I've addressed it in 19fe73e78e136fc28d38e534786e26f62804f595.

examples/wms-tiled-wrap-180.html
tschaub: Good catch.  Fixed in 19fe73e78e136fc28d38e534786e26f62804f595.

examples/center.html
tschaub: I think `<code>Alt+Shift+Drag</code>` is nicer than `<code>Alt</code>+<code>Shift</code>+drag` or some other variant.  I've updated things to be consistent in d3ff8d7fb0c59a8d12f94f208147b9dc2a3524fb.

examples/xyz.js
marcjansen: I didn't want to 'debate every change headed in this direction', not at all. In this particular case I think it was a misreading of the diff. I didn't realize that the source is gone as well.

I am +1 on example simplifications and that they concentrate one feature at a time.

examples/simple.js
marcjansen: From  http://www.openstreetmap.org/copyright/en : For a browsable electronic map, the credit should appear in the corner of the map.

So the collapsed version is sufficient? I simply don't know and faintly remember discussion in the ol3 repository that made us introduce this sort of awkward way to tell the attribution control to not be collapsible.

examples/vector-esri-edit.js
tschaub: Addressed in cbfd533d490738a36fcbc65a1bbd30f8e92a9b33.

examples/vector-esri-edit.js
tschaub: See also cbfd533d490738a36fcbc65a1bbd30f8e92a9b33.

examples/xyz.js
tschaub: Cool.  Sorry for implying that you were initiating a debate.  Your comments are great.  I was having flashbacks.

examples/simple.js
tschaub: I'm taking the very widespread use of collapsed attribution as a sign that it is acceptable.  If others have concern about this, I'm inclined to switch the base layer for this example to keep it simpler.

marcjansen: Agreed.

marcjansen: Just leave it collapsed to get this in.

marcjansen:
Please merge once the checks are green again.

Thanks again for rushing through all examples.

tschaub:
Thanks for the review and discussion @marcjansen.

examples/simple.js
tonio: When working on attributions, I wanted to have them collapsed by default, but there was some resistance… see https://github.com/openlayers/ol3/pull/2159

PR #4428, opened 2015-11-16
Add imageSize regression test for ol.source.ImageWMS

ahocevar:
Added test as follow-up on #4408.


marcjansen:
Gjslint now complaints that `image` is declared but not used.

ahocevar:
Ok, another attempt then...

ahocevar:
Now Travis is finally happy :smile:

marcjansen:

PR #4429, opened 2015-11-16
Dispatch a change event only if the value changes

### LINKED ISSUE #4427: Don't trigger the event "change:resolution" on max/min zoom
#### Settings
I have listening the event `change:resolution` on an `ol.View` and also set the limits of zoom (`minZoom and maxZoom`).

#### Problem
When i zoom-out or zoom-in on the map until reach the limit, the event is triggered even if the zoom isn't changed

I made this example (https://jsfiddle.net/he1m2t00/) to show you the problem.

---

_If you want more descriptive issue or i'm wrong just tell me_

fredj:
Fixes #4427

### DIFF
--- a/test/spec/ol/object.test.js
+++ b/test/spec/ol/object.test.js
@@ -168,6 +168,13 @@ describe('ol.Object', function() {
       expect(event.key).to.be('k');
     });
 
+    it('dispatches events only if the value is different', function() {
+      o.set('k', 1);
+      o.set('k', 1);
+      expect(listener1.calledOnce).to.be(true);
+      expect(listener2.calledOnce).to.be(true);
+    });
+
   });
 
   describe('setter', function() {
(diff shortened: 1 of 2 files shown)


ahocevar:
Thanks @fredj. Please merge when Travis is happy.

fredj:
A event will still be dispatched for different objects with the same value.
This is because the strict not equal operator is used:
```javascript
> [0, 1, 2] !== [0, 1, 2]
true
> {key: '42'} !== {key: '42'}
true
```

test/spec/ol/object.test.js
```diff
@@ -168,6 +168,13 @@ describe('ol.Object', function() {
       expect(event.key).to.be('k');
     });
 
+    it('dispatches events only if the value is not different', function() {
```
jonataswalker: Shouldn't be *only if the value **is** different*?

fredj: fixed, thanks

tschaub:
Looks good to merge.

PR #4433, opened 2015-11-17
Remove unused local variables

fredj:
Found with the `lintChecks` closure-compiler option


bartvde:
LGTM thanks for cleaning up ;-)

Issue #4438, opened 2015-11-17
crossOrigin valid type


fredj:
[`crossOrigin`](https://github.com/openlayers/ol3/blob/v3.11.1/externs/olx.js#L3975)  type may be `null`, `string` or `undefined`; what's the reason to accept `null`?

ahocevar:
As far as I remember, the reason was to allow to unset  crossOrigin for sources where the default is 'anonymous', e.g. ol.source.OSM.

bartvde:
@fredj is this still relevant after the merge of #4439 ?

fredj:
There's no difference between having `crossOrigin` `null` or `undefined`.
We can keep the externs as it.
(thanks for the reminder @bartvde)

PR #4439, opened 2015-11-17
Allow '' for crossOrigin (as Anonymous alias)

ahocevar:
This fixes a regression introduced by the goog.isDef removal.

From the HTML 5 spec: "The empty string is also a valid keyword, and maps to the Anonymous state"


fredj:
LGTM, thanks

ahocevar:
@fredj I needed to chance this pull request to check for `!== null` instead of `!== undefined`. Please have another look.

fredj:
You're right, it's `null` instead of `undefined`

fredj:
See also #4438

Issue #4441, opened 2015-11-17
Earthquake clusters example broken since v3.11.0


dtreiter:
The earthquake cluster example has an issue when hovering over a cluster.

## Expected behavior
Should display the features which are part of the cluster.

## Actual behavior
Rapidly blinks between showing the individual features and showing the cluster when the mouse is moving through a cluster's circle.

Verified in Chrome and Firefox.

## To reproduce
Go the earthquake cluster example and mouse over a cluster:
[Earthquake Example](http://openlayers.org/en/v3.11.1/examples/earthquake-clusters.html?q=cluster)

Note that this seems to be working prior to v3.11.0. For example, v3.10.1 is fine.
[v3.10.1 Example](http://openlayers.org/en/v3.10.1/examples/earthquake-clusters.html?q=cluster)

ahocevar:
Thanks for the great bug report. This looks like a regression, and i already have an idea what might be causing it. I'll investigate more after getting some sleep.

ahocevar:
I am working on this right now.

PR #4448, opened 2015-11-18
Check ol.source.UrlTile#urls property for null value

### LINKED ISSUE #4446: Cannot create ol.source.TileWMS with only tileLoadFunction in version 3.11.1
I just upgraded to version 3.11.1 from 3.10.1.
Unfortunately I cannot create ol.source.TileWMS without 'url' or 'urls' parameter in options.
The error comes from this method:
ol.source.TileWMS.prototype.resetCoordKeyPrefix_
It tries to iterate over 'this.urls' array which is actually null.
Previously this method iterated on 'this.urls_' and it worked fine.

fredj:
Fixes #4446

### DIFF
--- a/src/ol/source/tilearcgisrestsource.js
+++ b/src/ol/source/tilearcgisrestsource.js
@@ -91,7 +91,7 @@ ol.source.TileArcGISRest.prototype.getRequestUrl_ =
         pixelRatio, projection, params) {
 
   var urls = this.urls;
-  if (urls.length === 0) {
+  if (!urls) {
     return undefined;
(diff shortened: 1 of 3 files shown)


ahocevar:
Very nice fix, thanks @fredj. Please merge.

bartvde:
do we need a 3.11.2 release?

ahocevar:
If so, then we should wait until I have a fix for #4441.

tschaub:
I'll cherry pick these fixes for 3.11.2.

PR #4450, opened 2015-11-18
Fix select interaction regression caused by #4391

### LINKED ISSUE #4441: Earthquake clusters example broken since v3.11.0

ahocevar:
The select interaction determines selection candidate features by using `ol.Map#forEachFeatureAtPixel`. It uses an internal unmanaged layer that selected features are added to, and adds selected features to the skippedFeatures hash on the map. Before the change made with this pull request, a selected feature was immediately removed from the selection as soon as it was added to the unmanaged layer, because the unmanaged layer was ignored in forEachFeatureAtPixel (`null` as layer arg and `if (layer && filter())`). Now already selected features remain in the selection, because a `null` layer is accpted in the forEachFeatureAtPixel handler (`if (!layer || filter())`).

Fixes #4441.

### DIFF
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -283,7 +283,7 @@ ol.interaction.Select.handleEvent = function(mapBrowserEvent) {
          * @param {ol.layer.Layer} layer Layer.
          */
         function(feature, layer) {
-          if (layer && this.filter_(feature, layer)) {
+          if (!layer || this.filter_(feature, layer)) {
             selected.push(feature);
             this.addFeatureLayerAssociation_(feature, layer);
             return !this.multi_;
--- a/test/spec/ol/interaction/selectinteraction.test.js
+++ b/test/spec/ol/interaction/selectinteraction.test.js
@@ -179,7 +179,7 @@ describe('ol.interaction.Select', function() {
         unmanaged.setMap(map);
         map.renderSync();
         simulateEvent(ol.MapBrowserEvent.EventType.SINGLECLICK, 10, -20);
-        expect(spy.firstCall.args[0]).to.not.equal(feature);
+        expect(spy.getCalls().length).to.be(0);
         unmanaged.setMap(null);
       });
     });


tschaub:
@ahocevar can you provide a bit of a description and summary of the situation.  The diff alone is a bit mysterious, and the link to #4391 leads to another ticket without a description.  It would be great to get a bit of a knowledge dump on some of these tickets.

Thanks for the work getting to the bottom of this.

ahocevar:
@tschaub I updated the descriptions of both tickets. I hope it is clearer now - I'm a bit slow typing today, so I didn't get too verbose.

tschaub:
Thanks for the explanation @ahocevar.  This is very helpful.

Please merge.

tohu12:
@ahocevar, I use two "ol.interaction.Select" for two vector layers with different style for hover.
When you move with pointer to one Feature, show two different symbol from both interaction, because every interaction work with both unmanaged layer. 

Example http://jsfiddle.net/tohu/Lh76thmz/2/

What is right solution ? With v3.11.1 http://jsfiddle.net/tohu/Lh76thmz/1/ is alright, but now no.
Thanks

ahocevar:
The behavior you are seeing now is the same as in 3.10.1, but you're right that it should be considered a (separate) bug.

ahocevar:
@tohu12 I created #4471 to track this separate issue.

Issue #4431, opened 2015-11-16
Side-by-side example broken


tschaub:
On first load, not all tiles are rendered on all maps in the [side-by-side example](http://openlayers.org/en/master/examples/side-by-side.html)

![image](https://cloud.githubusercontent.com/assets/41094/11194628/573daaba-8c6a-11e5-9d8a-4cad1b993281.png)

fredj:
b57cdb730c5b57450a2f9adcca44038994181165 is the first bad commit

See #3895

fredj:
fixed with #4536

PR #4452, opened 2015-11-18
Call the change callback when a non idle tile is removed from the queue

fredj:
This PR fixes the side-by-side example (#4431), the regression was introduced in #3895

### DIFF
--- a/src/ol/tilequeue.js
+++ b/src/ol/tilequeue.js
@@ -100,6 +100,8 @@ ol.TileQueue.prototype.loadMoreTiles = function(maxTotalLoading, maxNewLoads) {
       tile.load();
       ++this.tilesLoading_;
       ++newLoads;
+    } else {
+      this.tileChangeCallback_();
     }
   }
 };


tschaub:
@fredj thanks for digging into this.  Do you understand how a tile in the queue becomes non-idle without the change callback being called?  It would be easier to understand if we only called the "change" callback in a change listener.

fredj:
probably because an other map loaded it

tschaub:
This suggests we should register a listener when a tile is added to the queue - not only when `loadMoreTiles` is called.

fredj:
Calling `this.tileChangeCallback_()` tells the other map that it must `this.Map#render` because a tile is loaded. I don't know if this change is enough

fredj:
This change fixes the side by side example but
b7e

marcjansen:
😆

bartvde:
LOL I feel like this very often ;-)

tschaub:
Let's make sure we get a solid fix in here.  Calling the change callback when there is no change feels like it will come back to haunt us.  I think the root of the issue is that we are not listening for change until too late.

fredj:
> This suggests we should register a listener when a tile is added to the queue - not only when `loadMoreTiles` is called.

Yes, good suggestion

fredj:
See https://github.com/openlayers/ol3/compare/master...fredj:tilequeue_listen_in_enqueue for a better solution

bartvde:
@fredj can you PR the better solution? And maybe close this one?

fredj:
@bartvde now in #4536

PR #4457, opened 2015-11-19
Add color option to ol.style.Icon

alexbrault:
This PR adds a color option to tint an ol.style.Icon.

See https://groups.google.com/forum/#!topic/ol3-dev/vPK4ygQHaqo

### DIFF
--- a/examples/icon.js
+++ b/examples/icon.js
@@ -2,7 +2,6 @@ goog.require('ol.Feature');
 goog.require('ol.Map');
 goog.require('ol.Overlay');
 goog.require('ol.View');
-goog.require('ol.color');
 goog.require('ol.geom.Point');
 goog.require('ol.layer.Tile');
 goog.require('ol.layer.Vector');
@@ -24,7 +23,7 @@ var iconStyle = new ol.style.Style({
     anchor: [0.5, 46],
     anchorXUnits: 'fraction',
     anchorYUnits: 'pixels',
-    color: ol.color.asArray('#8959A8'),
+    color: '#8959A8',
     opacity: 0.75,
     src: 'data/icon.png'
   }))
--- a/examples/icon.js
+++ b/examples/icon.js
@@ -2,6 +2,7 @@ goog.require('ol.Feature');
 goog.require('ol.Map');
 goog.require('ol.Overlay');
 goog.require('ol.View');
+goog.require('ol.color');
 goog.require('ol.geom.Point');
 goog.require('ol.layer.Tile');
 goog.require('ol.layer.Vector');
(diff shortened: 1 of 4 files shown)


alexbrault:
@ahocevar, I ended up multiplying pixel values instead of replacing white with the colour because the latter left undesired artifacts in the majority of cases (e.g. anything with antialiasing)

ahocevar:
Thanks for this pull request @alexbrault. Great enhancement! I'll review it within the next few days.

src/ol/style/iconstyle.js
```diff
@@ -492,6 +516,7 @@ ol.style.IconImage_.prototype.handleImageLoad_ = function() {
   this.imageState_ = ol.style.ImageState.LOADED;
   this.size_ = [this.image_.width, this.image_.height];
   this.unlistenImage_();
+  this.replaceColor_();
```
gberaudo: Applying the color will only work if the image is not tainted.
It should be moved after `determineTainting()`.

src/ol/style/iconstyle.js
```diff
@@ -576,6 +601,38 @@ ol.style.IconImage_.prototype.load = function() {
 
 
 /**
+ * @private
+ */
+ol.style.IconImage_.prototype.replaceColor_ = function() {
+  if (this.color_ === null) {
```
gberaudo: Should also return if `this.tainting_` is `true`.
See https://github.com/openlayers/ol3/pull/4457/files#r45316100

externs/olx.js
```diff
@@ -6004,6 +6004,7 @@ olx.style.FillOptions.prototype.color;
  *     anchorOrigin: (ol.style.IconOrigin|undefined),
  *     anchorXUnits: (ol.style.IconAnchorUnits|undefined),
  *     anchorYUnits: (ol.style.IconAnchorUnits|undefined),
+ *     color: (ol.Color|undefined),
```
ahocevar: Would be nice to also accept a `string` here.

alexbrault: I forget what my reasoning for omitting this was, but yes, it should

ahocevar:
This looks great @alexbrault. It looks like we do not have a [CLA](http://github.com/openlayers/cla) on file from you yet. Can you please submit one so we merge this?

examples/icon.js
```diff
@@ -23,6 +23,7 @@ var iconStyle = new ol.style.Style({
     anchor: [0.5, 46],
     anchorXUnits: 'fraction',
     anchorYUnits: 'pixels',
+    color: '#8959A8',
```
tschaub: I think it would be nicer to have a dedicated example that shows what this option should be used for.  We have a bad tendency to continue modifying examples that started out simple until they become overly complex.  And I think this particular example looked nicer before the `color` option was added (same goes for `opacity`).

image

alexbrault: Good point. I'll submit a PR to fork the example

PR #4472, opened 2015-11-22
Do not ignore layer filter for unmanaged layers

### LINKED ISSUE #2940: Select interaction ignores layer filter for feature overlays

### LINKED ISSUE #4471: Select interaction applies its select style to other select interactions
From https://github.com/openlayers/ol3/pull/4450#issuecomment-158753886:

I use two "ol.interaction.Select" for two vector layers with different style for hover.
When you move with pointer to one Feature, show two different symbol from both interaction, because every interaction work with both unmanaged layer.

Example http://jsfiddle.net/tohu/Lh76thmz/2/

What is right solution ? With v3.11.1 http://jsfiddle.net/tohu/Lh76thmz/1/ is alright, but now no.
Thanks

ahocevar:
PR #3883 made `forEachFeatureAtPixel` ignore unmanaged layers. This commit reverts that change.

This is an updated version of #4143.

Now that we return `null` as layer in the `ol.Map#forEachFeartureAtPixel` callback, we can pass unmanaged layers to the layer filter again, without the side effects that I feared when I reviewed #4143.

Fixes #4471.
Fixes #2940.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,18 @@
 ## Upgrade notes
 
+### v3.12.0
+
+#### `ol.Map#forEachFeatureAtPixel` changes
+
+The optional `layerFilter` function is now also called for unmanaged layers. To get the same behaviour as before, wrap your layer filter code in an if block like this:
+```js
+function layerFilter(layer) {
+  if (map.getLayers().getArray().indexOf(layer) !== -1) {
+    // existing layer filter code
+  }
+}
+```
+
 ### v3.11.0
 
 #### `ol.format.KML` changes
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -283,7 +283,7 @@ ol.interaction.Select.handleEvent = function(mapBrowserEvent) {
          * @param {ol.layer.Layer} layer Layer.
          */
         function(feature, layer) {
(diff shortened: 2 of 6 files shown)


ahocevar:
Sorry @tschaub, too many numbers again. But I feel like the select interaction and unmanaged layers mess is finally untangled now.

pgiraud:
+1

PR #4474, opened 2015-11-23
Remove use of goog.dom.createElement and goog.dom.createTextNode

fredj:
See [`goog.dom.createElement`](https://github.com/google/closure-library/blob/master/closure/goog/dom/dom.js#L804-L811) and [`goog.dom.createTextNode`](https://github.com/google/closure-library/blob/master/closure/goog/dom/dom.js#L814-L821) implementation.


marcjansen:
Looks good to me, @fredj. Please merge if the checks pass.

fredj:
Thanks for the review

PR #4475, opened 2015-11-23
Fix up failing tests in Internet Explorer

bartvde:
A lot of test cases currently fail in Internet Explorer.

This is a WIP on fixing them. Will update the description once I'm done.


ahocevar:
Are you also testing the examples manually to see if there is something in the library that needs to be fixed? Do you want me to help? I think we should fully support IE10+.

bartvde:
Not as yet @ahocevar but that might make sense to do as well after I finish up the unit tests.
Have we discarded IE9 as supported? I thought we still supported IE9?

bartvde:
do you remember what was the case with ```ol.source.Raster``` and IE? IIRC this is unsupported / can't be made to work in IE?

bartvde:
on second runs strange things happen, like pixelworks error for ol.source.TileJSON, maybe due to test cleanup?

screen shot 2015-11-23 at 11 04 44

bartvde:
if I take out the ```ol.source.Raster``` test spec, all is fine now on all runs, so something is going on there or with pixelworks ext

bartvde:
```ol.format.MVT``` uses ArrayBuffer so needs IE10+

**Comment by bartvde** [comment]
This seems to be the culprit test case:

```
    itNoPhantom('gets called before operations are run', function(done) {

      var count = 0;
      raster.setOperation(function(inputs) {
        ++count;
        return inputs[0];
      });

      raster.on('beforeoperations', function(event) {
        expect(count).to.equal(0);
        expect(!!event).to.be(true);
        expect(event.extent).to.be.an('array');
        expect(event.resolution).to.be.a('number');
        expect(event.data).to.be.an('object');
        done();
      });

      var view = map.getView();
      view.setCenter([0, 0]);
      view.setZoom(0);

    });
```

**Comment by ahocevar** [comment]
I'm looking into the raster issue now.

File: test/spec/ol/format/esrijsonformat.test.js [hunk 1]
```diff
@@ -304,15 +304,15 @@ describe('ol.format.EsriJSON', function() {
               -10579560.16462974, 4716567.373073828]))
             .to.be(true);
 
-        var last = result[305];
+        var last = result[8];
```
ahocevar: Now `last` is not a good name any more. `feature`?

bartvde: It's still the last feature, I just made the file smaller because it was very slow parsing in IE

ahocevar: Ah ok. You didn't push the data change.

bartvde: I did, it's here: test/spec/ol/format/esrijson/ksfields.json just on top of this

ahocevar: Oh sorry, you did. It's just collapsed in the diff.

bartvde:
The example template has a minor issue in IE11, the select combo option only shows up on hover, not by default. Seems to be caused by color: inherit from the bootstrap css.

screen shot 2015-11-23 at 17 34 48
screen shot 2015-11-23 at 17 34 43
screen shot 2015-11-23 at 17 34 32

bartvde:
Also the logo is stretched in IE, and I think we saw this before?

screen shot 2015-11-23 at 17 37 23

tschaub:
The raster source uses the `ImageData` constructor.  I'd suggest not running the raster tests in IE for now.  A workaround for the lack of an `ImageData` constructor in IE could go in the pixelworks lib.

bartvde:
@ahocevar your fix worked perfectly, and is in line with what @tschaub said. I applied the same fix for ol.format.MVT and IE9.

bartvde:
All tests now pass in IE9, IE10 and IE11

ahocevar:
Some commit messages do not comply with the rules, you may want to split them into title and description. Looks good otherwise.

PR #4189, opened 2015-09-29
Replace goog.nullFunction with ol.nullFunction

marcjansen:
Please review.

See #4128

PR #4477, opened 2015-11-23
Replace goog.nullFunction with ol.nullFunction

fredj:
See also #4189


marcjansen:
Looks good to me, @fredj. Please merge.

PR #4479, opened 2015-11-23
Export ol.interaction.Interaction#getMap function

### LINKED ISSUE #4169: ol.interaction.Interaction getMap()
It seems that the ol-debug.js version comes with the getMap() function available on ol.interaction.Interaction class, however that function is missing from the release version ol.js, and a call to that function will result in error.
  Is this a bug or am I missing something ?

fredj:
Fixes #4169


bartvde:
LGTM

PR #4480, opened 2015-11-23
[wip] Blur the zoom buttons after click

### LINKED ISSUE #3500: Zoom controls don't listen to touchend
On touch devices, the zoom buttons don't loose focus after being tapped.
Looking into the code, I figured it would be an easy fix, because the buttons only listen to mouseout and focusedout, but adding touchend still didn't result in a loss of focus.

Currently, the code in ol.control.Zoom is:

``` javascript
 goog.events.listen(inElement, [
   goog.events.EventType.MOUSEOUT,
    goog.events.EventType.FOCUSOUT
  ], function() {
    this.blur();
  }, false);
```

I verified: this.blur() gets called when goog.events.EventType.TOUCHEND has been added, but to no avail. You actually need to tap somewhere else for the focusout to kick in, after which the this.blur() does work. I'm not sure where to continue.

fredj:
Fixes #3500 

It will probably break the keyboard navigation
EDIT: yes, keyboard navigation is broken (as broken as Leaflet)

### DIFF
--- a/src/ol/control/zoomcontrol.js
+++ b/src/ol/control/zoomcontrol.js
@@ -85,6 +85,7 @@ goog.inherits(ol.control.Zoom, ol.control.Control);
 ol.control.Zoom.prototype.handleClick_ = function(delta, event) {
   event.preventDefault();
   this.zoomByDelta_(delta);
+  event.target.blur();
 };
 
 


fredj:
should be done at application level

PR #4484, opened 2015-11-24
Remove use of goog.dom.createDom

fredj:
Remove use of `goog.dom.createDom` when the function is only used to create an element and set its CSS class


ahocevar:
Thanks, looks good to me!

fredj:
More tests added

PR #4485, opened 2015-11-24
Use appendChild instead of goog.dom.append

fredj:


ahocevar:
+1, please merge.

fredj:
Thanks for the review

sbowler:
I just so happened to try to a new build and this change is causing issues with us creating a new overlay.

`Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'`

I'm guess the map may not have been added to the dom yet as its early on in the page load. goog.dom was probably doing some better error checking or handling. You may want to revert this.

fredj:
@sbowler I can not reproduce this issue; can you please give a short code snippet of your map and overlay creation?

I've tested with:
```javascript
var map = new ol.Map({
  layers: [
    new ol.layer.Tile({
    })
  ],
  view: new ol.View({
    center: [0, 0],
    zoom: 2
  })
});
var overlay = new ol.Overlay({});
overlay.setMap(map);
```

sbowler:
Sorry for the delay getting back. Please see https://github.com/openlayers/ol3/pull/4492 for the explanation of what happened.

PR #4487, opened 2015-11-24
Remove fallback css color

fredj:
All of out supported browsers supports CSS3 color: http://caniuse.com/#feat=css3-colors


bartvde:
LGTM

PR #4492, opened 2015-11-24
Revert "Use appendChild instead of goog.dom.append"

fredj:
Reverts openlayers/ol3#4485

Needs more work ...

### DIFF
--- a/src/ol/overlay.js
+++ b/src/ol/overlay.js
@@ -260,7 +260,7 @@ ol.Overlay.prototype.handleElementChanged = function() {
   goog.dom.removeChildren(this.element_);
   var element = this.getElement();
   if (element) {
-    this.element_.appendChild(element);
+    goog.dom.append(/** @type {!Node} */ (this.element_), element);
   }
 };
 
@@ -285,7 +285,7 @@ ol.Overlay.prototype.handleMapChanged = function() {
       goog.dom.insertChildAt(/** @type {!Element} */ (
           container), this.element_, 0);
     } else {
-      container.appendChild(this.element_);
+      goog.dom.append(/** @type {!Node} */ (container), this.element_);
     }
   }
 };


ahocevar:
Ok, what was the problem? Please revert if you feel you need to.

fredj:
Comment from @sbowler:
I just so happened to try to a new build and this change is causing issues with us creating a new overlay.

`Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'`

I'm guess the map may not have been added to the dom yet as its early on in the page load. goog.dom was probably doing some better error checking or handling. You may want to revert this.

alexbrault:
`goog.dom.appendChild` doesn't do any verification. It calls `append` on its first argument with the second as its parameter. If there is an issue with the appendChild's parameters, it's shouldn't come from calling it directly

See https://google.github.io/closure-library/api/source/closure/goog/dom/dom.js.src.html#l1052

fredj:
Thanks for the clarification @alexbrault

fredj:
Can't reproduce the issue reported by @sbowler, closing

oobrien:
This is causing problems with jQuery - regression from 3.11.2 to 3.12. 

var overlay = new ol.Overlay({
	  element: $('#popup')
	});

Uncaught TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.(…)

(since 3.12)

...but no error

var overlay = new ol.Overlay({
	  element: document.getElementById('overlay')
	});

oobrien:
OK - easy workaround for this one - get the underlying DOM element: .get(0) - I guess  goog.dom.append was doing this for me quietly and appendChild doesn't.

fredj:
`$('#popup')` returns an array of element, `ol.Overlay` expects a single element

sbowler:
Sorry for the delay. I figured out exactly what the problem is. Before you used to be able to pass in a JQuery element, but that is no longer the case since ```appendChild``` is not a valid function for a JQuery element (at least not for version 2.1.3 that I'm using)

My old code was 
```
new ol.Overlay({
     element: angular.element("<div class='circle-container'></div>"),
     stopEvent: false
 })
```

It kind of ugly to have to do the ```.get(0)``` to get the real HTML element, but I guess it's up to you guys if you want to put in some checks to handle a JQuery element and perhaps use the ```.append()``` function.

vlsg:
I am having the same issue as sbowler except that I have a lot of code relying on the element being passed in as a jQuery DOM element, not an HTML element, so if I try to replace with .get(0), none of my css-type properties or on-events will work.  Is there any chance this code could be updated to work with either a jQuery element ~or~ an HTML element so that it does not break existing, working implementations of ol?

probins:
the docs do state clearly that the element option should be an 'Element'. It was only by chance that your code worked with `goog.dom.append`.

vlsg:
Shux, I was hoping it could work either way, but I guess I just have some refactoring to do.  Thank you for your quick response.

alexbrault:
Ultimately, OL3 will need to use raw Elements internally. While it _could_ detect that you passed it a jQuery object, when you ask for it back, it would give you the raw Element (barring some extensive and quite unnecessary bookkeeping and possibly an additional dependency on jQuery).

Ultimately, the simple solution if you want to use jQuery elements in your own code is to use `.get(0)` when handing an element to OpenLayers (and only then), and to wrap the result of OpenLayers's `get*Element` in `$()` to convert them to jQuery

PR #4493, opened 2015-11-24
Remove use of goog.object

bjornharrtell:
This PR removes the remaining use of `goog.object` in source and tests by moving the original implementation of the functions that are still used to `ol.object`. I did some minor refactoring to some of the functions, i.e I make use of `Object.keys` where possible.

### DIFF
--- a/src/ol/control/attributioncontrol.js
+++ b/src/ol/control/attributioncontrol.js
@@ -7,12 +7,12 @@ goog.require('goog.dom');
 goog.require('goog.dom.classlist');
 goog.require('goog.events');
 goog.require('goog.events.EventType');
-goog.require('goog.object');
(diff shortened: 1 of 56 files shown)


bjornharrtell:
I'm aware some of these functions could be refactored away by using `Object.keys`. Let me know if that refactoring should be done in this PR.

bjornharrtell:
Ack.. not until the CI checks "goog.require classes must be alphabetized" was caught. Will fix.

bjornharrtell:
Missed some cases in the first attempt and had to alphabetize goog.require manually. Fixed, squashed and ready for review now.

bjornharrtell:
I think this work should be done in separate steps instead so I won't be continuing work on this PR. Closing.

Issue #4123, opened 2015-09-17
Raster reprojection stitch lines visible between triangles for semi-transparent sources


klokan:
Related to PR #4122

When rendering content of reprojection triangles, a clip path is set up on the canvas 2d context. However, in most browsers, the operations on the clip path edges are antialiased.

When rendering two neighboring triangles, the edge between them is slightly transparent (blending 50% opacity over 50% opacity does not produce 100% opacity):

![gap](https://cloud.githubusercontent.com/assets/707817/9254671/1429b59a-41e5-11e5-8a73-8ca3cc7be6cd.png)

**This does not happen in Chrome**, which does not antialias clip path edges, but it is currently reported as a bug and probably will be "fixed" in the future: http://crbug.com/424291

Although there are several possible `globalCompositeOperation` on canvas 2d context (see https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation for examples), none of them can be used to solve this issue -- the RGB values on the edge should be identical in both triangles, but we need to "sum" the alpha channel.

Currently, this is partially **solved by expanding the clip path triangles** by 1px so that the triangles slightly overlap and the gap is not visible.

This works well, **no gaps are visible usually**, unless the source data are semi-transparent - such as WMS serving partly transparent PNG tiles (the overlap results in noticeable line):

![semitrans](https://cloud.githubusercontent.com/assets/707817/9259881/b2646a6a-4204-11e5-9fd6-a9bc6a326747.png)

The special case of semi-transparent raster sources could be probably solved in future by implementing a manual (custom) compositing operation on the canvas in JavaScript, when alpha channel is treated non-standardly for this use-case.

bartvde:
These issues were created so we document the knowledge, but they are not active issues being worked on, so I'm going to close them.
Thanks for documenting them though @petrsloup they will be kept for posterity.

pksorensen:
Do anyone have an estimate on what it would cost to fix this? 

"Currently, this is partially solved by expanding the clip path triangles by 1px so that the triangles slightly overlap and the gap is not visible." How does one enable this?

https://github.com/openlayers/openlayers/issues/6342

ahocevar:
@pksorensen There is not really a way to fix this. But what you reported in #6342 can be solved. Let's continue the discussion there.

PR #4499, opened 2015-11-28
Fix gaps between vector tiles on Firefox

### LINKED ISSUE #4329: Gap between tiles in mapbox vector tile in Firefox

tsauerwein:
I was looking at this [nice demo](http://stvno.github.io/page/cbsexplorerol/) by @stvno and noticed that the tile borders were clearly visible on Firefox (or resizing the browser window made them appear).

The problem was that the coordinates used to clip the tiles were floating point numbers. Apparently this is no problem on Chrome but Firefox requires rounded integer values. My first attempt was to use `Math.round` for all coordinates of the clipping rectangle, but the gaps still showed up. So, now I am using `Math.floor` for the min. coordinates and `Math.ceil` for the max. coordinates.

@tonio Could you please test if this also fixes the problem on your system?

Closes #4329

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -1993,10 +1993,14 @@ ol.render.canvas.ReplayGroup.prototype.replay = function(
       flatClipCoords, 0, 8, 2, transform, flatClipCoords);
   context.save();
   context.beginPath();
-  context.moveTo(flatClipCoords[0], flatClipCoords[1]);
-  context.lineTo(flatClipCoords[2], flatClipCoords[3]);
-  context.lineTo(flatClipCoords[4], flatClipCoords[5]);
-  context.lineTo(flatClipCoords[6], flatClipCoords[7]);
+  context.moveTo(
+      Math.floor(flatClipCoords[0]), Math.floor(flatClipCoords[1]));
+  context.lineTo(
+      Math.floor(flatClipCoords[2]), Math.ceil(flatClipCoords[3]));
+  context.lineTo(
+      Math.ceil(flatClipCoords[4]), Math.ceil(flatClipCoords[5]));
+  context.lineTo(
+      Math.ceil(flatClipCoords[6]), Math.floor(flatClipCoords[7]));
   context.closePath();
   context.clip();
 


ahocevar:
Before this gets merged, please check with transparent polygon fills at tile borders whether this causes overlap artifacts in browsers other than Firefox.

tsauerwein:
@ahocevar I tried with a transparency of 80% and this is indeed a problem (both on Firefox and Chrome). I will have to take a closer look.

ol3-vt

ahocevar:
@tsauerwein The reason for this problem might be the same as the one described in #4123.

ahocevar:
I tested in all browsers I have available now, and Firefox is the only browser where I see this issue. So I tend to think that this is a Firefox bug, and not the same problem as the one described #4123.

ahocevar:
@tsauerwein Can you check if #4543 works for you? After a lot of experiments and research, I came up with that.

ahocevar:
@tsauerwein Since this is very specific to Firefox, I think the best thing we can do is create a minimal example (i.e. just draw to a canvas without OpenLayers) and report it as a Mozilla bug. Do you agree? Would you be able/willing to help me with that?

wlerner:
This issue is also visible in IE 11 (v. 11.0.9600.18124).

error pic 12

ahocevar:
Ok, with all the different canvas implementations, and antialiasing of clip lines being a reality, we need to fix it in a different way. I'd suggest we render vector tiles to a tile canvas and use drawImage to put the exact extent on the map canvas. This will also improve panning performance because we can cache the tile canvas, but things will get tricky for rotated maps. I'll spend some time this weekend trying out how this would work.

tsauerwein:
@ahocevar I came to think that it's exactly the same problem as #4123. The reason why your PR (#4543) did not work, is because the buffer was not big enough. The [pixelTileSize](https://github.com/openlayers/ol3/pull/4543/files#diff-91ca3be65c8f53d262a1b52f49309aefR186) is 4096, so 0.25 does not have an impact. I tried with the `tilePixelRatio` (16) instead, and it worked well.

We could use the same technique as in #4123 (for Chome use the default clipping, for all other browsers use a buffer). But there is still the problem with transparent fills...

> I'd suggest we render vector tiles to a tile canvas and use drawImage to put the exact extent on the map canvas. This will also improve panning performance because we can cache the tile canvas, but things will get tricky for rotated maps. I'll spend some time this weekend trying out how this would work.

I was also thinking about that. The idea is to draw (unrotated) to a canvas in the size of a tile. Everything that is outside the canvas bounds will be clipped automatically (without anti-aliasing!). And then this canvas is drawn onto the main canvas (like for normal raster tiles).
Like you say, the rotation will be tricky (for example labels would have to be drawn with the opposite angle). But I would say it is worth a try! :+1:

tsauerwein:
> Since this is very specific to Firefox, I think the best thing we can do is create a minimal example (i.e. just draw to a canvas without OpenLayers) and report it as a Mozilla bug. Do you agree?

@ahocevar Here is a small example. If you don't see the gap right away, try to resize the browser window.
http://jsfiddle.net/zarsm8fs/3/

But do you think it's a Firefox-only problem? In #4123 it sounds like it's a problem with all browsers except Chrome.

ahocevar:
> But do you think it's a Firefox-only problem? In #4123 it sounds like it's a problem with all browsers except Chrome.

Right. I think if we use drawImage instead of clipping, then we can later reuse the same transforms for image tile layers to avoid a per-layer composition canvas and render image tiles directly to the map canvas instead. So definitely worth trying :-).

ahocevar:
@tsauerwein, #4557 is the result of my extended weekend effort.

PR #4502, opened 2015-11-30
Improve ol.source.WMTS.optionsFromCapabilities jsdoc layout

fredj:


bartvde:
thanks for fixing, LGTM

PR #4505, opened 2015-11-30
Remove use of goog.object.getKeys

fredj:


fredj:
It was the last `goog.object.getKeys` in ol3, thanks for any review

marcjansen:
This looks good, @fredj. Do you think we must mention that it'll never be null from now on?

Please merge, with or without a note.

Thanks.

fredj:
`pointerMap` was already never `null` but the annotation was not precise.

fredj:
Thanks for the review Marc

PR #4510, opened 2015-12-02
Adding a check just in case the projection is not defined/supported

Delawen:
Adding a check just in case the projection is not defined/supported on current app. It will just ignore it, as if there was no containerSrs defined.

On the current implementation, an exception is thrown, making it difficult, for example, to get a featureinfo request right.

### DIFF
--- a/src/ol/format/gml/gml2format.js
+++ b/src/ol/format/gml/gml2format.js
@@ -64,7 +64,9 @@ ol.format.GML2.prototype.readFlatCoordinates_ = function(node, objectStack) {
   var axisOrientation = 'enu';
   if (containerSrs) {
     var proj = ol.proj.get(containerSrs);
-    axisOrientation = proj.getAxisOrientation();
+    if (proj) {
+      axisOrientation = proj.getAxisOrientation();
+    }
   }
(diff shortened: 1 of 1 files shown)


bartvde:
thanks @delawen do we have a (I)CLA from you?

Delawen:
I think I sent it a long time ago, but just to be sure I will fill it again.

bartvde:
Right seems to be missing from the list: https://github.com/openlayers/cla/blob/master/signed-agreements.md

Delawen:
Just sent it by the google form.

bartvde:
thanks!

bartvde:
@delawen for your next PR, you might consider splitting long commit messages into summary and paragraph as per: https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#use-clear-commit-messages

Delawen:
Will do it. Thanks for the quick response!

pvgenuchten:
some background info is that a featureinfo on data.rivm.nl/geoserver/wms fails due to geometry in featureinfo-gml is in local projection (epsg:28992) and the epsg:28992 projection definition is not available in the page.

PR #4515, opened 2015-12-04
Accept HTMLCanvasElement for an icon's img

### LINKED ISSUE #4513: Accept a HTMLCanvasElement as img for ol.style.Icon
Currently `ol.style.Icon` requires an image element with a `src` as `img` property . With a simple change to the way we cache icons, it could also accept a `HTMLCanvasElement`. This would allow users to programmatically create icons by drawing to a canvas.

ahocevar:
Fixes #4513.

### DIFF
--- a/examples/icon-negative.html
+++ b/examples/icon-negative.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: Icon Pixel Operations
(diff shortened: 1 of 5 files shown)


tsauerwein:
That's a nice improvement! Your changes look good to me.

PR #4518, opened 2015-12-04
Add support for OSM vector tiles

ahocevar:
Although OSM vector tiles are not ideal, they are used frequently. This pull request adds support for them, by implementing vector tile feature reprojection in a generic way.

Inspired by https://github.com/openlayers/ol3/issues/4516#issue-120399895.

### DIFF
--- a/examples/osm-vector-tiles.html
+++ b/examples/osm-vector-tiles.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: OSM Vector Tiles
+shortdesc: Using OpenStreetMap vector tiles.
+docs: >
+  A simple vector tiles map with OpenStreetMap vector tiles. **Note**: The tiles used in this example are not optimized for rendering - they clip tiles exactly at the tile boundary instead of adding a buffer, and use geographic coordinates instead of tile relative pixel coordinates in view projection.
+tags: "vector, tiles, osm"
+---
+<div id="map" class="map"></div>
--- a/examples/osm-vector-tiles.js
+++ b/examples/osm-vector-tiles.js
@@ -1,0 +1,142 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.format.TopoJSON');
+goog.require('ol.layer.VectorTile');
+goog.require('ol.proj');
+goog.require('ol.source.VectorTile');
+goog.require('ol.style.Fill');
+goog.require('ol.style.Stroke');
+goog.require('ol.style.Style');
+
+
+var format = new ol.format.TopoJSON();
+var tileGrid = ol.tilegrid.createXYZ({ maxZoom: 19 });
+var roadStyleCache = {};
+var roadColor = {
+  'major_road': '#776',
+  'minor_road': '#ccb',
+  'highway': '#f39'
+};
+var landuseStyleCache = {};
+var buildingStyle = [
+  new ol.style.Style({
+    fill: new ol.style.Fill({
+      color: '#666',
+      opacity: 0.4
+    }),
+    stroke: new ol.style.Stroke({
+      color: '#444',
+      width: 1
+    })
+  })
+];
+
+var map = new ol.Map({
+  layers: [
+    new ol.layer.VectorTile({
+      source: new ol.source.VectorTile({
+        format: format,
+        tileGrid: tileGrid,
+        url: 'http://{a-c}.tile.openstreetmap.us/' +
+            'vectiles-water-areas/{z}/{x}/{y}.topojson'
+      }),
+      style: new ol.style.Style({
+        fill: new ol.style.Fill({
+          color: '#9db9e8'
+        })
+      })
+    }),
+    new ol.layer.VectorTile({
+      source: new ol.source.VectorTile({
+        format: format,
+        tileGrid: tileGrid,
+        url: 'http://{a-c}.tile.openstreetmap.us/' +
+            'vectiles-highroad/{z}/{x}/{y}.topojson'
+      }),
+      style: function(feature, resolution) {
+        var kind = feature.get('kind');
+        var railway = feature.get('railway');
+        var sort_key = feature.get('sort_key');
+        var styleKey = kind + '/' + railway + '/' + sort_key;
+        var style = roadStyleCache[styleKey];
+        if (!style) {
+          var color, width;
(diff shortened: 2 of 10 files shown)


examples/osm-vector-tiles.html
```diff
@@ -0,0 +1,9 @@
+---
+layout: example.html
+title: OSM Vector Tiles
+shortdesc: Using OpenStreetMap vector tiles.
+docs: >
+  A simple vector tiles map with OpenStreetMap vector tiles. **Note**: The tiles used in this example are not optimized for rendering - they clip tiles exactly at the tile boundary instead of adding a buffer, and use geographic coordinates instead of tile relative pixel coordinates in view projection.
+tags: "simple, mapbox, vector, tiles"
```
bartvde: tags seem copy paste?

examples/osm-vector-tiles.js
```diff
@@ -0,0 +1,139 @@
...
+      }),
+      style: new ol.style.Style({
+        fill: new ol.style.Fill({
+          color: '#9db9e8'
+        })
+      })
+    }),
+    new ol.layer.VectorTile({
+      source: new ol.source.VectorTile({
+        format: new ol.format.TopoJSON(),
```
bartvde: minor but we could re-use the format?

examples/osm-vector-tiles.js
```diff
@@ -0,0 +1,139 @@
...
+          var color, width;
+          if (railway) {
+            color = '#7de';
+            width = 1;
+          } else {
+            color = {
+              'major_road': '#776',
+              'minor_road': '#ccb',
+              'highway': '#f39'
+            }[kind];
```
bartvde: minor again, should we have this color object outside of the style function?

examples/osm-vector-tiles.js
```diff
@@ -0,0 +1,139 @@
...
+            width = kind == 'highway' ? 1.5 : 1;
+          }
+          styleArray = [new ol.style.Style({
```
bartvde: no need for an array anymore given Tim's recent change to allow returning style objects?

examples/osm-vector-tiles.js
```diff
@@ -0,0 +1,139 @@
...
+    }),
+    new ol.layer.VectorTile({
+      source: new ol.source.VectorTile({
+        format: new ol.format.TopoJSON(),
+        tileGrid: ol.tilegrid.createXYZ({ maxZoom: 19 }),
+        url: 'http://{a-c}.tile.openstreetmap.us/' +
+            'vectiles-buildings/{z}/{x}/{y}.topojson'
+      }),
+      style: function(f, resolution) {
+        return (resolution < 10) ? buildingStyle : [];
```
bartvde: can be null instead of [] ?

src/ol/vectortile.js
```diff
@@ -72,6 +75,13 @@ ol.VectorTile = function(tileCoord, state, src, format, tileLoadFunction) {
   this.tileLoadFunction_ = tileLoadFunction;
 
   /**
+   * Feature coordinates are {@link ol.proj.Units.TILE_PIXELS}.
+   * @type {boolean}
+   * @private
+   */
+  this.pixelSpace_;
```
bartvde: can we not set it here automatically based on projection?

ahocevar: projection is feature projection, pixelSpace is from data projection. So no.

ahocevar:
I'll make the suggested example changes. Thanks for the review so far.

ahocevar:
I addressed all your comments @bartvde, and added more tests.

src/ol/proj/proj.js
```diff
@@ -675,14 +675,19 @@ ol.proj.get = function(projectionLike) {
...
   if (projection1 === projection2) {
     return true;
   } else if (projection1.getCode() === projection2.getCode()) {
-    return projection1.getUnits() === projection2.getUnits();
+    return equalUnits;
   } else {
     var transformFn = ol.proj.getTransformFromProjections(
         projection1, projection2);
-    return transformFn === ol.proj.cloneTransform;
+    return transformFn === ol.proj.cloneTransform && equalUnits;
```
marcjansen: Can this line ever return true?

ahocevar: Why not? `1 === 1 && true` returns `true` as well.

marcjansen: Good then. I was just wondering.

bartvde:
LGTM please merge

src/ol/proj/proj.js
```diff
@@ -675,14 +675,19 @@ ol.proj.get = function(projectionLike) {
  * @return {boolean} Equivalent.
  */
 ol.proj.equivalent = function(projection1, projection2) {
+  var units1 = projection1.getUnits();
+  var units2 = projection2.getUnits();
+  var tilePixels = ol.proj.Units.TILE_PIXELS;
+  var equalUnits = units1 === units2 ||
+      units1 === tilePixels || units2 === tilePixels;
```
tsauerwein: For me this is a special case that rather should be handled in `ol.featureloader.tile`. I find it kind of counter intuitive that two projections should be equivalent when only one of them has pixel units.

petrsloup: +1

ahocevar:
Done @tsauerwein.

ahocevar:
Did you want to take another look @tsauerwein?

tsauerwein:
Thanks, this looks good to me.

examples/osm-vector-tiles.js
```diff
@@ -0,0 +1,142 @@
...
+var format = new ol.format.TopoJSON();
+var tileGrid = ol.tilegrid.createXYZ({ maxZoom: 19 });
+var roadStyleCache = {};
+var roadColor = {
+  'major_road': '#776',
+  'minor_road': '#ccb',
+  'highway': '#f39'
+};
+var landuseStyleCache = {};
+var buildingStyle = [
```
tschaub: For future reference, this could also just be a `ol.style.Style` instance instead of an array (and you get a very slight performance gain for never using arrays).

ahocevar: Oh, one I forgot to change. Thanks, I created #4535.

PR #4521, opened 2015-12-06
Fix imagestyle docs

ahocevar:
The constructor docs of `ol.style.Image` are outdated. This pull request fixes that.


ahocevar:
Trivial doc only change, so merging.

PR #4523, opened 2015-12-06
Allow rendering geometries to an arbitrary canvas

### LINKED ISSUE #4514: Use ol.render.canvas.Immediate on an arbitrary canvas
Thinking #4513 further, we could make `ol.render.canvas.Immediate`  work to draw vector geometries on an arbitrary canvas. Users could then easily use arbitrary shapes as icon images, or use them to fill a polygon once #4241 is merged.

The API could look like this:
```js
canvas.width = 24;
canvas.height = 24;

/**
 * @type {ol.render.canvas.Immediate}
 */
var render = ol.render.toContext(canvas.getContext('2d'));

render.setFillStrokeStyle(new ol.style.Fill({color: 'blue'}));
render.drawPolygonGeometry(new ol.geom.Polygon([[[0, 0], [0, 23], [23, 0], [0, 0]]]);
```
The above would draw a blue triangle that fills the top left half of the 24x24 canvas.

ahocevar:
Fixes #4514.

### DIFF
--- a/src/ol/render.js
+++ b/src/ol/render.js
@@ -1,0 +1,43 @@
+goog.provide('ol.render');
+
+goog.require('goog.vec.Mat4');
+goog.require('ol.render.canvas.Immediate');
+goog.require('ol.vec.Mat4');
+
+
+/**
+ * Binds a Canvas Immediate API to a canvas context, to allow drawing geometries
+ * to the context's canvas.
+ * ```js
+ * var canvas = document.createElement('canvas');
+ * var render = ol.render.toContext(canvas.getContext('2d'),
+ *     { size: [100, 100] });
+ * render.setFillStrokeStyle(new ol.style.Fill({ color: blue }));
+ * render.drawPolygonGeometry(
+ *     new ol.geom.Polygon([[[0, 0], [100, 100], [100, 0], [0, 0]]]));
+ * ```
+ * Note that {@link ol.render.canvas.Immediate#drawAsync} and
+ * {@link ol.render.canvas.Immediate#drawFeature} cannot be used.
+ *
+ * @param {CanvasRenderingContext2D} context Canvas context.
+ * @param {olx.render.ToContextOptions=} opt_options Options.
+ * @return {ol.render.canvas.Immediate} Canvas Immediate.
+ * @api
+ */
+ol.render.toContext = function(context, opt_options) {
+  var canvas = context.canvas;
+  var options = opt_options ? opt_options : {};
+  var pixelRatio = options.pixelRatio || ol.has.DEVICE_PIXEL_RATIO;
+  var size = options.size;
+  if (size) {
+    canvas.width = size[0] * pixelRatio;
+    canvas.height = size[1] * pixelRatio;
+    canvas.style.width = size[0] + 'px';
+    canvas.style.height = size[1] + 'px';
+  }
+  var extent = [0, 0, canvas.width, canvas.height];
+  var transform = ol.vec.Mat4.makeTransform2D(goog.vec.Mat4.createNumber(),
+      0, 0, pixelRatio, pixelRatio, 0, 0, 0);
+  return new ol.render.canvas.Immediate(context, pixelRatio, extent, transform,
+      0);
+};
--- a/examples/render-geometry.html
+++ b/examples/render-geometry.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: Render geometries to a canvas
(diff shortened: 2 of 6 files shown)


examples/draw-geometry.html
```diff
@@ -0,0 +1,9 @@
+---
+layout: example.html
```
bartvde: as for the example name, maybe render-geometry might be a bit more self explaining? Draw geometry sounds more like people drawing a geometry with draw tools. Render geometry more captures the feeling of the GeoExt legend (which was called FeatureRenderer IIRC).

examples/draw-geometry.js
```diff
@@ -0,0 +1,29 @@
...
+goog.require('ol.geom.Polygon');
+goog.require('ol.render');
+goog.require('ol.style.Circle');
+goog.require('ol.style.Fill');
+goog.require('ol.style.Stroke');
+
+
+var canvas = document.getElementById('canvas');
+var pixelRatio = ol.has.DEVICE_PIXEL_RATIO;
+canvas['width'] = 100 * pixelRatio;
```
bartvde: so is the width and height in css still needed?

examples/draw-geometry.js
```diff
@@ -0,0 +1,29 @@
...
+var fill = new ol.style.Fill({ color: 'blue' });
+var stroke = new ol.style.Stroke({ color: 'black' });
+render.setFillStrokeStyle(fill, stroke);
+render.setImageStyle(new ol.style.Circle({
+  radius: 10,
+  fill: fill,
+  stroke: stroke
+}));
+
+render.drawLineStringGeometry(new ol.geom.LineString([[10, 10], [90, 90]]));
```
bartvde: maybe a silly question, but the coordinates do not need to be adjusted for the pixelRatio?

bartvde: Ah I guess this is handled automatically. Maybe it would be cleaner then if ol.render.toContext got a width and height parameter, so people do not have the apply the pixelRatio math there?

bartvde: Ah I see now the function only gets the canvas context, so that will not work.

bartvde:
Nice work @ahocevar this will greatly simplify things like GeoExt vector legends, not having to use a Map anymore.

I've put a few comments / questions in there.

examples/draw-geometry.js
ahocevar: Can you give me an API example (also in conjunction with css pixels) so I understand what you mean?

examples/draw-geometry.js
ahocevar: Yes, otherwise the canvas will be scaled to css pixels. But now that you mention it, that feels strange. Let me double check.

examples/draw-geometry.js
bartvde: My think was:

var size = [100, 100]
ol.render.toContext(canvas.getContext('2d'), pixelRatio , size);

where ol.render.toContext would set the size on the canvas instead.

ahocevar: Ah I see. Yes, that makes sense. Will change.

ahocevar:
@bartvde I have addressed your comments. The optional 2nd argument is now an object literal with options, including `size` as you suggested.

src/ol/render.js
```diff
@@ -0,0 +1,44 @@
...
+goog.require('ol.render.canvas.Immediate');
+goog.require('ol.vec.Mat4');
+
+
+/**
+ * Binds a Canvas Immediate API to a canvas context, to allow drawing geometries
+ * to the context's canvas.
+ * ```js
+ * var canvas = document.createElement('canvas');
+ * canvas.width = 100;
```
bartvde: I guess this needs to be updated as well to the new API?

src/ol/render.js
```diff
@@ -0,0 +1,44 @@
...
+ * canvas.height = 100;
+ * var render = ol.render.toContext(canvas.getContext('2d'));
```
bartvde: ```{size: [100, 100]}```

bartvde:
@ahocevar I put a small comment wrt the example code, but this LGTM
Please merge if Travis gives the okay

src/ol/render.js
ahocevar: Oops, right.

ahocevar: Done.

ahocevar:

tsauerwein:
Great idea!

PR #4525, opened 2015-12-06
Make render tests work on more devices

ahocevar:
By setting a device pixel ratio of 1.0 and increasing the tolerance for raster reprojection tests, the render tests now also run from the console on OSX devices with retina display.


tsauerwein:
Thanks, please merge.

PR #4527, opened 2015-12-07
Add support for smooth TileWMS dimensions

bartvde:
This continues on the work done in #4389 but now for tiled WMS

### DIFF
--- a/examples/wms-time.js
+++ b/examples/wms-time.js
@@ -1,0 +1,64 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.MapQuest');
+goog.require('ol.source.TileWMS');
+
+var startDate = new Date(Date.parse('2012-01-01T19:00:00Z'));
+var frameRate = 0.5; // frames per second
+var animationId = null;
+
+var counter = 0;
+
+var layers = [
+  new ol.layer.Tile({
+    source: new ol.source.MapQuest({layer: 'sat'})
+  }),
+  new ol.layer.Tile({
+    extent: [-13884991, 2870341, -7455066, 6338219],
+    source: new ol.source.TileWMS(/** @type {olx.source.TileWMSOptions} */ ({
+      url: 'http://oos.soest.hawaii.edu/thredds/wms/hioos/model/wav/ww3/' +
+          'WaveWatch_III_Global_Wave_Model_best.ncd?',
+      params: {'LAYERS': 'Thgt', 'TIME': startDate.toISOString()}
+    }))
+  })
+];
+var map = new ol.Map({
+  layers: layers,
+  target: 'map',
+  view: new ol.View({
+    center: [-10997148, 4569099],
+    zoom: 4
+  })
+});
+
+var setTime = function() {
+  startDate.setHours(startDate.getHours() + 1);
+  layers[1].getSource().updateParams({'TIME': startDate.toISOString()});
+  updateInfo();
+};
+
+var stop = function() {
+  if (animationId !== null) {
+    window.clearInterval(animationId);
+    animationId = null;
+  }
+};
+
+var play = function() {
+  stop();
+  animationId = window.setInterval(setTime, 1000 / frameRate);
+};
+
+var updateInfo = function() {
+  var el = document.getElementById('info');
+  el.innerHTML = startDate.toISOString();
+};
+
+var startButton = document.getElementById('play');
+startButton.addEventListener('click', play, false);
+
+var stopButton = document.getElementById('pause');
+stopButton.addEventListener('click', stop, false);
+
+updateInfo();
--- a/examples/wms-time.html
+++ b/examples/wms-time.html
@@ -1,0 +1,14 @@
+---
+layout: example.html
+title: WMS Time
+shortdesc: Example of smooth tile transitions when changing the time dimension of a tiled WMS layer.
+docs: >
+  Demonstrates smooth reloading of layers when changing the time dimension continously.
+tags: "wms, time, dimensions, transition"
+---
+<div id="map" class="map"></div>
+<div role="group" aria-label="Animation controls">
+  <button id="play" type="button">Play</button>
+  <button id="pause" type="button">Pause</button>
+  <span id="info"></span>
+</div>
--- a/src/ol/source/tilewmssource.js
+++ b/src/ol/source/tilewmssource.js
@@ -67,6 +67,13 @@ ol.source.TileWMS = function(opt_options) {
 
   /**
    * @private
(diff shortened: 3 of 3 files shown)


marcjansen:
What do you think of the following example diff:

```diff
diff --git a/examples/wms-time.html b/examples/wms-time.html
index a58362d..7478dfa 100644
--- a/examples/wms-time.html
+++ b/examples/wms-time.html
@@ -7,9 +7,8 @@ docs: >
 tags: "wms, time, dimensions, transition"
 ---
 <div id="map" class="map"></div>
-<div class="row-fluid">
-  <div class="span12">
-    <a id="start" href="#" class="btn">Start</a>
-    <a id="stop" href="#" class="btn">Stop</a>
-  </div>
+<div role="group" aria-label="Animation controls">
+  <button id="play" type="button">Play</button>
+  <button id="pause" type="button">Pause</button>
+  <span id="info"></span>
 </div>
diff --git a/examples/wms-time.js b/examples/wms-time.js
index 140aa51..8bae9f6 100644
--- a/examples/wms-time.js
+++ b/examples/wms-time.js
@@ -38,6 +38,7 @@ var map = new ol.Map({
 var setTime = function() {
   startDate.setHours(startDate.getHours() + 1);
   layers[1].getSource().updateDimensions({'TIME': startDate.toISOString()});
+  updateInfo();
 };
 
 var stop = function() {
@@ -52,12 +53,15 @@ var play = function() {
   animationId = window.setInterval(setTime, 1000 / frameRate);
 };
 
-var startButton = document.getElementById('start');
-startButton.addEventListener('click', function() {
-  play();
-}, false);
+var updateInfo = function(){
+  var el = document.getElementById('info');
+  el.innerHTML = startDate.toISOString();
+};
 
-var stopButton = document.getElementById('stop');
-stopButton.addEventListener('click', function() {
-  stop();
-}, false);
+var startButton = document.getElementById('play');
+startButton.addEventListener('click', play, false);
+
+var stopButton = document.getElementById('pause');
+stopButton.addEventListener('click', stop, false);
+
+updateInfo();
```

bartvde:
nice @marcjansen can I cherry pick from somewhere? Or should I just apply your diff?

bartvde:
@marcjansen I also just gave you push access right now to my ol3 repo, so that is also an option

marcjansen:
Just apply the diff. I was in a hurry... nothing pushed.

bartvde:
Sure thanks @marcjansen

tschaub:
Glad to see this improvement being applied elsewhere.  I'm confused why this has to be thought of as something that applies only to "dimensions" (whatever that means).  I think it would be nicer to see generic support for updating a tile URL without clearing previously rendered tiles.

So for WMS, `source.updateParams()` would look nice.  And for any other `ol.source.UrlTile` based source, `source.setUrl()` (and friends) would do the same.

I regret not pushing for this harder when work started on WMTS, but I think it would be unfortunate to go around adding additional options to all sources instead of having a more general solution.

externs/olx.js
```diff
@@ -5256,6 +5257,15 @@ olx.source.TileWMSOptions.prototype.crossOrigin;
 
 
 /**
+ * Additional "dimensions" for tile requests.  This is an object with properties
+ * named like the advertised dimensions (such as TIME).
+ * @type {Object|undefined}
+ * @api stable
```
tschaub: If this does make it through as a new option, I hope it can be kept experimental for at least one release.  That would allow us to make a more general solution later.

bartvde:
That also crossed my mind when doing this @tschaub I think I can change this, at least for TileWMS to work for all parameters. Will update this PR.

bartvde:
argh maybe it's not so easy after all, seems the logic depends on the dynamic / dimensions params to end up in ```paramsKey``` but not in ```tileCoordKey```

bartvde:
@tschaub please see if the updated PR is more what you had in mind

fredj:
@bartvde do we want this for the v3.13.0 release?

bartvde:
I guess it can wait @fredj 
This still needs review

fredj:
ok, I'll remove the milestone tag

tschaub:
I think this change looks great @bartvde.  I haven't tested it yet locally, but if you are confident that it works as expected, I think it would be good to get in the release.

bartvde:
thanks @tschaub I'll do a last test to verify it still works correctly and then merge

cc @fredj

tschaub:
I'm testing locally, and I can't tell if the WMS time example works as expected.  I'm on a slow connection, so this may be to blame, but after zooming in and out, I don't see tiles updating (though I'm watching the network traffic go by).

bartvde:
that works okay for me @tschaub

tschaub:
Maybe the issue I'm seeing is not restricted to this change.  I don't see tiles being updated until I pause.

bartvde:
There were some fixes in master after this that might be important.

tschaub:
Things improve as tiles are cached (I have your changes on top of master).  Again, this is likely my slow connection and probably is not specific to the WMS changes.  It looks like you can get in a situation where a new request is issued before the previous one returns and tiles are not updated.

bartvde:
Ah okay, thanks for the explanation. I guess my connection is too fast for that to happen, maybe if I increase the frameRate

tschaub:
I think I'm essentially seeing the same as @thomasmoelhave in https://github.com/openlayers/ol3/pull/4389#issuecomment-157745360.  It would be nice to revisit the underlying implementation to see if things can be improved.

bartvde:
Right I see what you mean if I change the frameRate to 2

examples/wms-time.js
```diff
@@ -0,0 +1,64 @@
...
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.MapQuest');
+goog.require('ol.source.TileWMS');
+
+var startDate = new Date(Date.parse('2012-01-01T19:00:00Z'));
+var frameRate = 0.5; // frames per second
+var animationId = null;
+
+var counter = 0;
```
tschaub: Defined but never used.  ESLint [caught this](https://travis-ci.org/openlayers/ol3/builds/103601244) after the merge.

examples/wms-time.js
```diff
@@ -0,0 +1,64 @@
...
+  view: new ol.View({
+    center: [-10997148, 4569099],
+    zoom: 4
+  })
+});
+
+var setTime = function() {
+  startDate.setHours(startDate.getHours() + 1);
+  layers[1].getSource().updateParams({'TIME': startDate.toISOString()});
+  updateInfo();
```
tschaub: ESLint will also want `updateInfo` to be defined before this.

PR #4530, opened 2015-12-08
Remove use of goog.isNull

fredj:


marcjansen:
I have just restarted the build.

marcjansen:
It looks as if 


```shell
node tasks/build.js config/examples-all.json build/compiled-examples/all.combined.js
```

doesn't work on travis …

ahocevar:
Looks like something is or was wrong with the image-load-events example. I restarted the build again. Let's see if it finishes now.

ahocevar:
Looks good, please merge.

fredj:
Thanks for the review

PR #4532, opened 2015-12-08
Mark olx.source.WMTSOptions#dimensions as not null

fredj:
FYI: in a project I have:
```js
var dimensions = source.getDimensions();
var dimensionKeys = Object.keys(dimensions);
```
But the compiler complains that the Object may be null:
```
ERR! compile: ERROR - actual parameter 1 of Object.keys does not match formal parameter
ERR! compile found   : (Object|null)
ERR! compile required: Object
ERR! compile   var dimensionKeys = Object.keys(dimensions);
ERR! compile                                   ^
ERR! compile 
ERR! compile 
ERR! compile 1 error(s), 0 warning(s), 
ERR! compile 97.4% typed
ERR! compile 
ERR! compile
```

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5537,7 +5537,7 @@ olx.source.VectorOptions.prototype.wrapX;
  *     version: (string|undefined),
  *     format: (string|undefined),
  *     matrixSet: string,
- *     dimensions: (Object|undefined),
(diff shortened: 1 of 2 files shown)


bartvde:
LGTM

PR #4535, opened 2015-12-09
Do not use an array for a single style

ahocevar:
Since arrays are used nowhere else in the example, it will be a bit faster to also use a plain style here. See https://github.com/openlayers/ol3/pull/4518/files#r47052600.


fredj:
LGTM

PR #4536, opened 2015-12-09
Register the change callback when the tile is enqueued

fredj:
See #4452 

The `change` callback is registered as soon as the tile is queued; this allows to a tile to be queued in more than one queue and be notified when the tile loads.

A new Object has been added to hold the ids of the loading tiles; this is needed to only decrements the counter if the queue instance has loaded the tile.

### DIFF
--- a/src/ol/structs/priorityqueue.js
+++ b/src/ol/structs/priorityqueue.js
@@ -122,6 +122,7 @@ ol.structs.PriorityQueue.prototype.dequeue = function() {
 /**
  * Enqueue an element. O(log N).
  * @param {T} element Element.
+ * @return {boolean} The element was added to the queue.
  */
 ol.structs.PriorityQueue.prototype.enqueue = function(element) {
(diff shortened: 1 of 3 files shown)


bartvde:
Let's see if we can get this fixed in 3.12. I'm going to move the release to mid next week.

bartvde:
closes #4431

fredj:
todo: `ol.TileQueue#getTilesLoading()` is wrong because `this.tilesLoading_` is decremented even if an other queue loaded the tile.

fredj:
Ready to be reviewed

bartvde:
@tschaub or @aisaacs any chance one of you could review this?

tschaub:
This looks good to me.  I didn't confirm that the side-by-side example is improved, but this looks like the right change.  Thanks for tackling it @fredj.  Please merge (assuming the examples work).

fredj:
The side by side example work

fredj:
Thanks for the review

PR #4537, opened 2015-12-09
Allow specifying non default DPI on ol.source.TileArcGISRest

### LINKED ISSUE #4312: Ability to set a custom DPI on ol.source.TileArcGISRest
It seems that ol.layer.Tile with source new ol.source.TileArcGISRest() ignores "projection" parameter. It uses EPSG:3857 instead. Am I wrong?

My code:
```javascript
new ol.layer.Tile({
	extent: [306520.146, 5973291.050, 680103.277, 6257813.452],
	source: new ol.source.TileArcGISRest({
		url: "http://www.domain.com/some_url/MapServer/",
		params: {
			FORMAT: "png8",
			DPI: 96
		},
		projection: ol.proj.get("EPSG:3346")
	})
})
```

By the way, I can't set custom DPI also! Documentation doesn't mention I can't do that!

bartvde:
closes #4312

### DIFF
--- a/src/ol/source/tilearcgisrestsource.js
+++ b/src/ol/source/tilearcgisrestsource.js
@@ -102,7 +102,9 @@ ol.source.TileArcGISRest.prototype.getRequestUrl_ =
   params['BBOX'] = tileExtent.join(',');
   params['BBOXSR'] = srid;
   params['IMAGESR'] = srid;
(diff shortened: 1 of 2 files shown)


ahocevar:
Looks good, please merge.

bartvde:
thanks for the review @ahocevar

PR #4541, opened 2015-12-10
Add new ol.geom.LineString#getCoordinateAt function

### LINKED ISSUE #3779: Exporting method for midpoint of a linestring
A method which returns a certain point of a linestring would be useful in many cases. There is already a function called "getFlatMidpoint" which returns the midpoint of the linestring.
I think this function should be exported and in addition it should provide a fraction as parameter in order to get any point on the linestring. For the midpoint i would call this method with "0.5" as argument.

fredj:
fixes #3779 

@openlayers/ol3 do you think it's a good addition to the API ?

### DIFF
--- a/src/ol/geom/linestring.js
+++ b/src/ol/geom/linestring.js
@@ -167,6 +167,23 @@ ol.geom.LineString.prototype.getCoordinates = function() {
 
 
 /**
+ * Return the coordinate at the provided fraction along the linestring.
+ * The `fraction` is a number between 0 and 1, where 0 is the start of the
+ * linestring and 1 is the end.
+ * @param {number} fraction Fraction.
+ * @param {ol.Coordinate=} opt_dest Optional coordinate whose values will
+ *     be modified. If not provided, a new coordinate will be returned.
+ * @return {ol.Coordinate} Coordinate of the interpolated point.
+ * @api
+ */
+ol.geom.LineString.prototype.getCoordinateAt = function(fraction, opt_dest) {
+  return ol.geom.flat.interpolate.lineString(
+      this.flatCoordinates, 0, this.flatCoordinates.length, this.stride,
+      fraction, opt_dest);
+};
+
+
+/**
  * Return the length of the linestring on projected plane.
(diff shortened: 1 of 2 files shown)


tschaub:
I think it is a useful addition (with minimal extra code).  We should talk about the name a bit.  Including "point" in the name might imply a `ol.geom.Point` is returned.  But `interpolatedCoordinate` is a mouthful.  We already have [`line.getCoordinateAtM()`](http://openlayers.org/en/master/apidoc/ol.geom.LineString.html#getCoordinateAtM).

Also wordy, but `line.getCoordinateAt(fraction)` *might* be a sensible name.

Or maybe we go for brevity with `line.interpolate(fraction)`.

Thoughts?

fredj:
Function renamed to `getCoordinateAt`

src/ol/geom/linestring.js
```diff
@@ -167,6 +167,21 @@ ol.geom.LineString.prototype.getCoordinates = function() {
 
 
 /**
+ * Return a point interpolated along the linestring, `fraction` is a number
```
tschaub: I think this should be broken into two sentences:

> Return the coordinate at the provided fraction along the linestring.  The `fraction` is
> a number between 0 and 1, where 0 is the start of the linestring and 1 is the end.

src/ol/geom/linestring.js
```diff
@@ -167,6 +167,21 @@ ol.geom.LineString.prototype.getCoordinates = function() {
...
+ * between 0 and 1 representing fraction of total length of linestring.
+ * @param {number} fraction Fraction.
+ * @param {ol.Coordinate=} opt_dest Destination.
```
tschaub: I think "destination" is a bit confusing.  How about "Optional coordinate whose values will be modified.  If not provided, a new coordinate will be returned."?

tschaub:
Thanks for updating this @fredj.  A couple minor doc comments.  This should make for pretty easy testing as well.

fredj:
Unit tests added

tschaub:
Looks great.  Please merge.

PR #4543, opened 2015-12-10
Use a subpixel overlap for render regions

### LINKED ISSUE #4329: Gap between tiles in mapbox vector tile in Firefox

ahocevar:
When rendering vector tiles, the clip lines may cause small gaps between neighbouring tiles. By introducing a 0.5 pixel overlap (0.25 pixels for each tile), this effect can be avoided.

This is an alternative to #4499, which also works for transparent layers/geometries.

Fixes #4329.

### DIFF
--- a/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
@@ -178,16 +178,18 @@ ol.renderer.canvas.VectorTileLayer.prototype.createReplayGroup = function(tile,
       'Source is an ol.source.VectorTile');
   var tileGrid = source.getTileGrid();
   var tileCoord = tile.getTileCoord();
+  var buffer = source.getOpaque() ? 0.25 * pixelRatio : 0;
+  var resolution = tileGrid.getResolution(tileCoord[0]);
   var pixelSpace = tile.getProjection().getUnits() == ol.proj.Units.TILE_PIXELS;
   var extent;
   if (pixelSpace) {
     var tilePixelSize = source.getTilePixelSize(tileCoord[0], pixelRatio,
         tile.getProjection());
-    extent = [0, 0, tilePixelSize[0], tilePixelSize[1]];
+    extent = [-0.25, -0.25, tilePixelSize[0] + 0.25, tilePixelSize[1] + 0.25];
   } else {
-    extent = tileGrid.getTileCoordExtent(tileCoord);
+    extent = ol.extent.buffer(tileGrid.getTileCoordExtent(tileCoord),
+        buffer * resolution);
   }
-  var resolution = tileGrid.getResolution(tileCoord[0]);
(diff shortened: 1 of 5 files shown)


tsauerwein:
This is the technique used in #4123, right?

On Firefox without transparent fill it looks fine (no artefacts when rotating, resizing the map or zooming in). But with transparent fills, I am still seeing the gaps:
ol3-ff-transparent

And even in Chrome I am now seeing gaps.

Without transparency:
ol3-chrome

With transparency:
ol3-chrome-transparent

Sorry, not there yet.. :)

ahocevar:
This is so platform specific! I tested on Windows with a device pixel ratio of 1 and on OSX with a device pixel ratio of 2, and it looked fine in all browsers. Maybe we should just leave things as they are?

ahocevar:
> This is the technique used in #4123, right?

The same idea at least, although there the overlap is bigger.

ahocevar:
Another option would be to respect the `opaque` option of the source, and use floor/ceil for the clip polygon edges like @tsauerwein did in #4499, but only when `opaque` is `true`.

ahocevar:
@tsauerwein I implemented the `opaque` approach now, and no rounding of clip polygon coordinates anymore. Can you please check how things are now on your platforms?

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -178,16 +178,18 @@ ol.renderer.canvas.VectorTileLayer.prototype.createReplayGroup = function(tile,
...
   var tileCoord = tile.getTileCoord();
+  var buffer = source.getOpaque() ? 0.25 * pixelRatio : 0;
+  var resolution = tileGrid.getResolution(tileCoord[0]);
   var pixelSpace = tile.getProjection().getUnits() == ol.proj.Units.TILE_PIXELS;
   var extent;
   if (pixelSpace) {
     var tilePixelSize = source.getTilePixelSize(tileCoord[0], pixelRatio,
         tile.getProjection());
-    extent = [0, 0, tilePixelSize[0], tilePixelSize[1]];
+    extent = [-0.25, -0.25, tilePixelSize[0] + 0.25, tilePixelSize[1] + 0.25];
```
tsauerwein: I guess you also wanted to use `buffer` here?

tsauerwein:
Chrome is ok both for transparent and non-transparent fill (when I replace `0.25` with `buffer`, see my comment).

On Firefox with a transparent fill and `opaque=false`, I see very thin gaps:
ol3-ff-gaps
And with a non-transparent fill and `opaque=true`, I also see gaps (more clear).

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
ahocevar: Yes indeed.

ahocevar:
So what should we do @tsauerwein? Leave things as they are? Use the opaque option? Or add a new layer property, e.g. `clipBuffer`?

PR #4548, opened 2015-12-11
Improve vector tile related docs

### LINKED ISSUE #4538: Layer Definition renderBuffer of Vector Tile
the renderBuffer of vector tile layer does not display label and point symbol properly.
the symbologies is splitted off outside the tiles..

When can be fixed ? It's critical for me hall...

ahocevar:
The documentation of the `renderBuffer` option for `ol.layer.VectorTile` was misleading, and vector tile related information was missing in the API doc landing page.

Fixes #4538.


marcjansen:
LGTM.

PR #4551, opened 2015-12-11
Handle tile coordinate wrapping when reprojecting raster tiles

### LINKED ISSUE #4447: raster reprojection wrapX
the raster reprojection don't extend the map when global or wrapX is defined.

petrsloup:
This PR adds handling of wrapX to the tile reprojection process.

This way, when reprojecting for example `EPSG:3857` to `EPSG:4326` the map is "repeated", because the default TileGrid for `EPSG:4326` (`ol.tilegrid.createForProjection('EPSG:4326')`) has wrapX enabled.

Internally, this uses `ol.source.Tile#getTileCoordForTileUrlFunction` to handle the wrapping.

Fixes #4447

@ahocevar I believe you did a lot of work on the wrapX feature. Can you please take a look if this is indeed the right way of handling this?


ahocevar:
This is exactly the right fix. Thanks @petrsloup, please merge.

petrsloup:
@ahocevar Thanks for the review.

PR #4552, opened 2015-12-11
Make sure compiled code calls ol.VectorTile#setProjection

ahocevar:
This pull request fixes an issue which can be observed when looking at http://openlayers.org/en/master/examples/mapbox-vector-tiles.html: the tiles are reprojected, because the compiler strips out the code that sets the tile projection with the correct `tile-pixels` units in `ol.featureloader.xhr` (for whatever reason).


ahocevar:
Thanks for any review.

bartvde:
LGTM

PR #4556, opened 2015-12-14
Addition of updateSketchCondition on Ol.interaction.draw as a way to fix #3907

### LINKED ISSUE #3907: Polygon draw interaction buggy on mobile devices
There seems to be a problem with ol.interaction.Draw when using it on mobile devices (primary on iOS).
The problem occurs when drawing a polygon and (as far as we've noticed) slightly move your finger when touching a point - the point will be drawn but when you add another point the previous one simply disappears!

We use Openlayers with default interactions (dragging, zooming etc) and dynamically add the draw interaction. I thought the problem would occur only in combination of these but it was also reproducable after removing any other interactions. And the problem also occurs on the demo page: http://openlayers.org/en/v3.5.0/examples/draw-features.html

It seems like there is some problem in the "event-detection" - meaning whether it's a simple touch event or drag on the user side. The point gets drawn on the map, but is never added to the polygon coordinates(?), and therefore disappears on the next click.
I could reproduce this problem on Android as well, but it's "harder" to do and unlikely because I have to actively move my finger on touching down (I think this is because Android and iOS handle touch/drag detection differently). The problem never appeared on desktop devices.

I've also uploaded a video where you can see the problem in action:
https://dl.dropboxusercontent.com/u/23708747/ShareX/2015-07/VIDEO0122.mp4

I hope this is enough information. I couldn't find anything likely on the web. And I would dig through the code myself but I'm not exactly a Javascript pro :)

daanoz:
By adding a updateSketchCondition we can fix issue #3907.

``` javascript
        var draw = new ol.interaction.Draw({
          features: featureCollection,
          type: 'Polygon',
          updateSketchCondition: function(event) {
            return event.originalEvent.type !== 'touchmove';
          }
        });
```

The concept of limiting the modifyDrawing_ function might also be used by other use cases like creating a boundingbox available to draw the features.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2603,6 +2603,16 @@ olx.interaction.DrawOptions.prototype.geometryName;
  */
 olx.interaction.DrawOptions.prototype.condition;
 
+/**
+ * A function that takes an {@link ol.MapBrowserEvent} and returns a boolean
+ * to indicate whether the current sketch being drawn should be updated. This
+ * can be used to either limit the scope of the feature being drawn, or to
+ * skip events caused panning/zooming on a touch device. By default
+ * {@link ol.events.condition.always}, i.e. it is always updated.
+ * @type {ol.events.ConditionType|undefined}
+ * @api
+ */
+olx.interaction.DrawOptions.prototype.updateSketchCondition;
 
 /**
  * Condition that activates freehand drawing for lines and polygons. This
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -297,6 +297,13 @@ ol.interaction.Draw = function(options) {
    * @private
    * @type {ol.events.ConditionType}
    */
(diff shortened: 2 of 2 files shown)


bjornharrtell:
Got hit by this issue today. Tried this fix and it seems to do the trick. In my case I needed and tested this with the measure example code.

@daanoz, thanks for the patch. Usually they need branch squashed and CLA from you to accept contributions.

bjornharrtell:
I've now also verified it helps resolve the general issues with `ol.interaction.Draw` as described in #3907.

daanoz:
Thanks @bjornharrtell, I'll see what I can to get the CLA approved.

bjornharrtell:
Great. I took the liberty of creating #4668 that squashed and fixes a minor documentation issue in case you where not available. You decide if you want to fix up this PR or close it in favor of mine (you are still be the main commit author in my PR).

tschaub:

PR #4557, opened 2015-12-14
Conditionally render tiles to a separate tile canvas

### LINKED ISSUE #4329: Gap between tiles in mapbox vector tile in Firefox

ahocevar:
Because clip geometries are anti-aliased in most browsers, there will be tiny
gaps between tiles. If tiles are rendered to a tile canvas which is then drawn
to the map canvas upon composition, these gaps can be avoided. For rotated
views, it is stil necessary to clip the tile, but in this case a 1-pixel
buffer is used.

This change also brings a huge performance improvement for panning, because
the fully rendered tiles can be reused.

Because of the added cost of using drawImage in addition to replaying the tile
replay group, we fall back to directly drawing to the map canvas when the tile
canvas would be too large, or during interaction/animation when resolution or
rotation change.

Fixes #4329.

### DIFF
--- a/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
@@ -88,6 +88,8 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
   var projection = viewState.projection;
   var resolution = viewState.resolution;
   var rotation = viewState.rotation;
+  var size = frameState.size;
+  var pixelScale = pixelRatio / resolution;
   var layer = this.getLayer();
   var source = layer.getSource();
   goog.asserts.assertInstanceof(source, ol.source.VectorTile,
@@ -116,9 +118,9 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
   var tileGrid = source.getTileGrid();
 
   var currentZ, height, i, ii, insertPoint, insertTransform, offsetX, offsetY;
-  var origin, pixelScale, pixelSpace, replayState, scale, size, tile;
-  var tileCenter, tileContext, tileExtent, tilePixelResolution, tilePixelSize;
-  var tileResolution, tileSize, tileTransform, width;
+  var origin, pixelSpace, replayState, scale, tile, tileCenter, tileContext;
+  var tileExtent, tilePixelResolution, tilePixelSize, tileResolution, tileSize;
+  var tileTransform, width;
   for (i = 0, ii = tilesToDraw.length; i < ii; ++i) {
     tile = tilesToDraw[i];
     replayState = tile.getReplayState();
@@ -127,10 +129,8 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
     currentZ = tile.getTileCoord()[0];
     tileSize = ol.size.toSize(tileGrid.getTileSize(currentZ), this.tmpSize_);
     pixelSpace = tile.getProjection().getUnits() == ol.proj.Units.TILE_PIXELS;
-    size = frameState.size;
     tileResolution = tileGrid.getResolution(currentZ);
     tilePixelResolution = tileResolution / source.getTilePixelRatio();
-    pixelScale = pixelRatio / resolution;
     scale = tileResolution / resolution;
     offsetX = Math.round(pixelRatio * size[0] / 2);
     offsetY = Math.round(pixelRatio * size[1] / 2);
--- a/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
@@ -11,6 +11,7 @@ goog.require('ol.ViewHint');
 goog.require('ol.array');
 goog.require('ol.dom');
 goog.require('ol.extent');
+goog.require('ol.geom.flat.transform');
 goog.require('ol.layer.VectorTile');
 goog.require('ol.proj.Units');
 goog.require('ol.render.EventType');
@@ -50,6 +51,18 @@ ol.renderer.canvas.VectorTileLayer = function(layer) {
    * @type {Array.<ol.VectorTile>}
    */
   this.renderedTiles_ = [];
+
+  /**
+   * @private
+   * @type {number}
+   */
+  this.resolution_ = NaN;
+
+  /**
+   * @private
+   * @type {number}
+   */
+  this.rotation_ = NaN;
 
   /**
    * @private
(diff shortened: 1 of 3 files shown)


tsauerwein:
Nice, Andreas!

I gave it a quick test, unfortunately I am seeing gaps in Chrome (with rotation) and also in Firefox (both rotated and unrotated):

Chrome (unrotated):
ol3-chrome

Firefox (unrotated):
ol3-ff

> For rotated views, it is stil necessary to clip the tile, but in this case a 1-pixel buffer is used.

Why is it necessary to use the buffer for rotated views?

Is there a difference on how to the tile canvas is drawn compared to how raster tile are drawn?

ahocevar:
@tsauerwein I made a few changes now, so at least without rotation things should be fine now. Can you please test again?

For rotated views, I did not find a way to make things work without clipping - maybe you want to give it a try or think it through? Right now the tile context holds a rotated image. Without clipping would only work if there is a way for the tile context to hold an unrotated image with rotated labels and symbols.

ahocevar:
@tsauerwein I have added another commit which changes the code to render unrotated tiles with rotated labels and use drawImage to render those to a rotated map canvas. This avoids the use of any clip geometries, but for some reason the labels do not work correctly when the pixel ratio is different than 1. I cannot find where the problem is. Can you take a look please? Maybe I am missing something obvious, or there is a bug elsewhere.

Rotated view with `ol.has.DEVICE_PIXEL_RATIO == 1`
pixelratio_1

Rotated view with `ol.has.DEVICE_PIXEL_RATIO == 2`
pixelratio_2

tsauerwein:
@ahocevar I am still seeing gaps on Firefox also without rotation. I have been wondering about the difference to the raster tile renderer. The raster tile renderer first renders a canvas on which each tile image is drawn un-transformed. Then this canvas is drawn onto the main canvas using the transformation matrix.

Right now, the position for each vector tile canvas is calculated using the transformation matrix. Do you think this could make a difference?

ahocevar:
The difference is that the raster tile renderer does not render to the map canvas, but to a layer canvas that is then composed with other layers by the map renderer. That layer canvas is bigger than the map canvas, because it covers the complete required tile range and not just the visible viewport. However, that is inefficient - try to load 5 raster layers and pan the map, which is almost unusable.

I'd like to move towards composing directly to the map canvas also for raster tile layers, so it'd be good to get this approach working. I consider it work in progress for now, and I'm sure the transforms and/or the way we put tiles on the map canvas can be improved.

The biggest show stopper at the moment is that I cannot figure out why the canvas replay does not rotate labels properly when a pixel ratio other than 1 is used. I was hoping to get another pair of eyes looking at the code, potentially figuring out what might be wrong.

ahocevar:
@tsauerwein I updated the transforms now to use integer pixels everywhere (when not rotated), and now the gaps should really be gone in Firefox (when not rotated). Improvements for rotated views should also be possible.

But I'd still appreciate if you could look at the code, to see if you find something wrong that causes the labels to be rendered incorrectly when the pixel ratio is not 1.

ahocevar:
I found the problem @tsauerwein - replay with text styles only works on contexts that are not transformed, because a transform is applied to the context, which resets the previous transform. I think I'll be able to fix this now.

tsauerwein:
> ... because a transform is applied to the context, which resets the previous transform. 

This?
https://github.com/openlayers/ol3/blob/master/src/ol/render/canvas/canvasreplay.js#L334-L340

I have been wondering about that, but I didn't understand why it worked for a pixel-ratio of 1. Curious about your fix! :)

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -113,35 +115,83 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
...
+    pixelScale = pixelRatio / resolution;
+    scale = tileResolution / resolution;
+    offsetX = Math.round(pixelRatio * size[0] / 2);
+    offsetY = Math.round(pixelRatio * size[1] / 2);
+    if (ol.extent.getWidth(tileExtent) > ol.extent.getWidth(extent)) {
+      if (pixelSpace) {
+        origin = ol.extent.getTopLeft(tileExtent);
+        tileTransform = ol.vec.Mat4.makeTransform2D(this.tmpTransform_,
+            offsetX, offsetY,
+            pixelScale * tilePixelResolution,
```
tsauerwein: The pixel-ratio is not used in this transform ([it used to be](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js#L131-L135)). Is that correct?

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -113,35 +115,83 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
...
+            -rotation,
+            -tileCenter[0], -tileCenter[1]);
+      }
+      tileContext = tile.getContext();
+      if (replayState.resolution !== resolution ||
+          replayState.rotation !== rotation) {
+        replayState.resolution = resolution;
+        replayState.rotation = rotation;
+        tileContext.canvas.width = width + 0.5;
+        tileContext.canvas.height = height + 0.5;
```
tsauerwein: Does this `0.5` have any effect? `width` and `height` are [integers](https://html.spec.whatwg.org/multipage/scripting.html#attr-canvas-width), so the decimal places are cut off.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -113,35 +115,83 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
...
+    tileExtent = tileGrid.getTileCoordExtent(
+        tile.getTileCoord(), this.tmpExtent_);
     currentZ = tile.getTileCoord()[0];
-    tileSize = tileGrid.getTileSize(currentZ);
-    tilePixelSize = source.getTilePixelSize(currentZ, pixelRatio, projection);
-    tilePixelRatio = tilePixelSize[0] /
-        ol.size.toSize(tileSize, this.tmpSize_)[0];
+    tileSize = ol.size.toSize(tileGrid.getTileSize(currentZ), this.tmpSize_);
+    pixelSpace = tile.getProjection().getUnits() == ol.proj.Units.TILE_PIXELS;
+    size = frameState.size;
```
tsauerwein: `size` could be set outside the loop.

ahocevar:
> This? https://github.com/openlayers/ol3/blob/master/src/ol/render/canvas/canvasreplay.js#L334-L340

Exactly. See ac64f7efd3937a9842699004a4714230df62bf33.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
ahocevar: Correct. Adding `0.5` has the same effect as rounding `width` and `height`. Example: For a calculated width of `511.9998` you would get a canvas width of `511`, but you want `512`.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
ahocevar: The pixel ratio is now in `pixelScale`.

ahocevar:
@tsauerwein Thanks for your help! I think this is now ready for review.

tsauerwein:
Wow, this is awesome Andreas! Thanks for your tenacity.

I tested with Firefox and Chrome, unrotated it looks great with transparent and non-transparent fills, and with a pixel-ratio of 1 and 2. Only with a rotation I see very thin lines (both on Firefox and Chrome).
Firefox:
ol3-ff
Chrome:
ol3-chrome

But that's a huge improvement! I will see that I find some time to review.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
tsauerwein: Ok, thanks for the explanation.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -179,16 +230,19 @@ ol.renderer.canvas.VectorTileLayer.prototype.createReplayGroup = function(tile,
...
   var tileCoord = tile.getTileCoord();
+  var buffer = 1;
+  var resolution = tileGrid.getResolution(tileCoord[0]);
   var pixelSpace = tile.getProjection().getUnits() == ol.proj.Units.TILE_PIXELS;
   var extent;
   if (pixelSpace) {
     var tilePixelSize = source.getTilePixelSize(tileCoord[0], pixelRatio,
         tile.getProjection());
-    extent = [0, 0, tilePixelSize[0], tilePixelSize[1]];
+    extent = [-buffer, -buffer,
```
tsauerwein: Is this buffer still needed?

ahocevar: Could indeed be removed now that direct rendering is a corner case. Will do so tomorrow morning.

tsauerwein:
I am +1 for merging. Thanks for the great work, Andreas!

ahocevar:
Thanks for the review and your help @tsauerwein.

syntax42:
Nice to have improvements in performance! Do i understand it correctly, that thin white lines can still appear?

ahocevar:
@syntax42 Yes, on rotation and during zoom animations. But these will go away soon in another round of improvements.

PR #4559, opened 2015-12-15
Fix up failing KML test in Firefox

### LINKED ISSUE #3439: Test ol.format.KML#readNetworkLinks fails on Firefox
Test: "returns an array of network links"

    expected 'http://localhost:3000/test/bar/bar.kml' to equal 'bar/bar.kml'

bartvde:
fixes #3439 

TIA for any quick review

### DIFF
--- a/test/spec/ol/format/kmlformat.test.js
+++ b/test/spec/ol/format/kmlformat.test.js
@@ -2790,7 +2790,7 @@ describe('ol.format.KML', function() {
       var nl = format.readNetworkLinks(text);
       expect(nl).to.have.length(2);
       expect(nl[0].name).to.be('bar');
-      expect(nl[0].href).to.be('bar/bar.kml');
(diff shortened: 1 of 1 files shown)


bartvde:
What we are seeing in Firefox is to be expected given https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI

ahocevar:
Try `.getAttribute('href')` instead of `.href`. Should be a nicer to read fix.

bartvde:
Not sure I understand @ahocevar this is the result object from the KML parser, not an XML structure?

ahocevar:
Ah I see. The test would be easier to read like this I think:
```js
expect(nl[0].href.replace(window.location.href, '').to.be('bar/bar.kml');
```

bartvde:
Agreed ;-) made the change

ahocevar:
Yep, looks good now. Please merge when Travis agrees.

PR #4561, opened 2015-12-15
Exclude feature loader MVT test from IE9

### LINKED ISSUE #4560: feature loader test case fails in IE9
screen shot 2015-12-15 at 09 57 23

bartvde:
closes #4560

### DIFF
--- a/test/spec/ol/featureloader.test.js
+++ b/test/spec/ol/featureloader.test.js
@@ -75,17 +75,18 @@ describe('ol.featureloader', function() {
       loader.call(tile, [], 1, ol.proj.get('EPSG:3857'));
     });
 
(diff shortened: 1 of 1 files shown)


ahocevar:
Instead of
```js
describe('foo', function() {
  beforeEach(function() {
    // setup
  });
  it('bar', function() {
    // the only test in this describe block
  })
});
```
it would be nicer to do
```js
it('foo bar', function() {
  // setup
  // the only test in this describe block
});
```
Would make the tests easier to read, and you can do `xit` instead of `xdescribe`.

bartvde:
agreed, will make the change

ahocevar:
Looks good now. Please merge when Travis has finished their review.

PR #4567, opened 2015-12-16
Always pass on a compare function to sort

### LINKED ISSUE #4566: Tiles disappearing on pan when zoom level change is big
Found this out in a screenshare with @ahocevar just now, to reproduce go to:

http://openlayers.org/en/v3.12.0/examples/xyz-esri.html

Zoom in with a very small bbox, and then pan the map. The tiles will blur completely sometimes. In 3.11.2 you will see a small section blurred only, where no tiles are present as yet.

bartvde:
fixes #4566

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -38,6 +38,18 @@ ol.array.binaryFindNearest = function(arr, target) {
 
 
 /**
+ * Default compare for array sort, will make sure number sort works okay.
+ * @param {*} a The first object to be compared.
+ * @param {*} b The second object to be compared.
+ * @return {number} A negative number, zero, or a positive number as the first
+ *     argument is less than, equal to, or greater than the second.
+ */
+ol.array.defaultCompare = function(a, b) {
+  return a > b ? 1 : a < b ? -1 : 0;
+};
+
+
+/**
  * Whether the array contains the given object.
  * @param {Array.<*>} arr The array to test for the presence of the element.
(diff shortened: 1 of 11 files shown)


bartvde:
seeing if I can add some test cases to prevent this from happening again

bartvde:
not easy to test unfortunately, if anybody has ideas for test cases let me know, I've added 2 unit tests for now that we're low hanging fruit.

bartvde:
ready for review, this probably warrants a 3.12.1 release

marcjansen:
This looks good to me, Bart. Plese merge.


 I think a patch release is appropriate. Maybe @tschaub, @ahocevar, @fredj or @tsauerwein have an opinion as well?

Thanks.

bartvde:
thanks @marcjansen I'll probably wait a few days to see if another regression pops up. If nothing pops up, and nobody is against it, I'll likely do a 3.12.1 at the end of this week.

ahocevar:
I think `defaultCompare` is a misleading name. What about `numberCompare`? And I agree that this warrants a patch release.

bartvde:
I can change it to ```numberCompare``` @ahocevar

ahocevar:
Or maybe `numberSafeCompare` would be even better.

bartvde:
okay we settled on ```numberSafeCompareFunction``` ;-)

PR #4570, opened 2015-12-16
Rename defaultSort to numberSafeCompareFunction on ol.array

bartvde:
see: https://github.com/openlayers/ol3/pull/4567#issuecomment-165092679

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -38,13 +38,13 @@ ol.array.binaryFindNearest = function(arr, target) {
 
 
 /**
- * Default compare for array sort, will make sure number sort works okay.
+ * Compare function for array sort that is safe for numbers.
  * @param {*} a The first object to be compared.
  * @param {*} b The second object to be compared.
  * @return {number} A negative number, zero, or a positive number as the first
  *     argument is less than, equal to, or greater than the second.
  */
-ol.array.defaultCompare = function(a, b) {
+ol.array.numberSafeCompareFunction = function(a, b) {
   return a > b ? 1 : a < b ? -1 : 0;
 };
 
--- a/test/spec/ol/array.test.js
+++ b/test/spec/ol/array.test.js
@@ -2,13 +2,12 @@ goog.provide('ol.test.array');
 
 describe('ol.array', function() {
 
-  describe('defaultCompare', function() {
+  describe('numberSafeCompareFunction', function() {
     it('sorts as expected', function() {
       var arr = [40, 200, 3000];
-      var expected = [40, 200, 3000];
       // default sort would yield [200, 3000, 40]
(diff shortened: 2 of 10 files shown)


bartvde:
cc @ahocevar

ahocevar:
Nice, please merge!

src/ol/array.js
```diff
@@ -38,13 +38,13 @@ ol.array.binaryFindNearest = function(arr, target) {
...
 /**
- * Default compare for array sort, will make sure number sort works okay.
+ * Compare function for array sort that is safe for numbers.
  * @param {*} a The first object to be compared.
  * @param {*} b The second object to be compared.
  * @return {number} A negative number, zero, or a positive number as the first
  *     argument is less than, equal to, or greater than the second.
  */
-ol.array.defaultCompare = function(a, b) {
+ol.array.numberSafeCompareFunction = function(a, b) {
```
tschaub: Not sure I get the motivation to rename this.  This works for strings or numbers.  As long as the name isn't exposed to users, I don't think it is a big deal, but I think `ol.array.defaultCompare` was sensible.

bartvde: cc @ahocevar

bartvde: We can always revert ;-)

ahocevar: I thought that something that isn't default should not be named default*. Honestly I wasn't aware that the original fix was merged already when I commented, it definitely won't hurt my feelings when this gets reverted :-).

marcjansen: Naming stuff is hard, and I think neither `defaultCompare` nor `numberSafeCompareFunction` are fully capturing what @tschaub correctly mentions. If we are to rename (again), I'd probably go with `safeCompareFunction`, `safeComparisonFunction` or the variants withouts the `…Function`-suffix.

test/spec/ol/array.test.js
```diff
@@ -2,13 +2,12 @@ goog.provide('ol.test.array');
...
-  describe('defaultCompare', function() {
+  describe('numberSafeCompareFunction', function() {
     it('sorts as expected', function() {
       var arr = [40, 200, 3000];
-      var expected = [40, 200, 3000];
       // default sort would yield [200, 3000, 40]
-      arr.sort(ol.array.defaultCompare);
-      expect(arr).to.eql(expected);
+      arr.sort(ol.array.numberSafeCompareFunction);
+      expect(arr).to.eql(arr);
```
tschaub, line 10: Oops.  This is always true.

src/ol/array.js
tschaub: I decided to propose a new name in #14311.

marcjansen: I like it.

PR #4571, opened 2015-12-16
Correct typedef of selectinteraction options obj

marcjansen:
The key `features` of the constructor options that can be passed to
`ol.interaction.Select` can also be `undefined`.

Rendered API docs before: before

… and after: after
 
Please review.


bartvde:
LGTM

marcjansen:
Cheers for the review, Bart.

PR #4578, opened 2015-12-17
Fix apidoc landing page markup

ahocevar:
This bug in the landing page markup causes the link to ol.layer.VectorTile not to be shown.


bartvde:
Please merge

PR #4584, opened 2015-12-21
Allow ol.source.Cluster to cluster by a physical distance

dtreiter:
Previously, ol.source.Cluster only created clusters using a distance in pixels on the screen. This is useful for decluttering features as a user changes zoom level.

This change adds a `ClusterUnits` option to ol.source.Cluster to allow clustering by a physical distance as well. This is useful when you want the clusters to be independent of zoom level. For example, consider a situation where you want to automatically pan/zoom between clusters of geographical features.

This new option defaults to the old behavior if it is undefined.


tschaub:

PR #4588, opened 2015-12-22
Performance management when rendering vector tiles

ahocevar:
To improve the user experience when zooming and rotating vector tile layers, this pull request suggests the introduction of a "fast lane" for rendering vector tiles that take longer than a certain threshold to render. When the fast lane is entered, existing pre-rendered tiles will be scaled to match the current view, without replaying the replay group.

When the layer is configured with `updateWhileAnimating` or `updateWhileInteracting`, the fast lane will be avoided. The threshold for the maximum rendering is defined in the new compiler define `ol.MAX_VECTOR_TILE_RENDER_TIME`.

### DIFF
--- a/src/ol/ol.js
+++ b/src/ol/ol.js
@@ -223,6 +223,16 @@ ol.SIMPLIFY_TOLERANCE = 0.5;
 
 
 /**
+ * @define {number} Maximum time in millsieconds spent for accurate rendering of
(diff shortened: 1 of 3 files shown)


ahocevar:
I don't really like how this complicates the code, and it also breaks tile positioning, re-introducing gaps.

PR #4590, opened 2015-12-23
Add getResolutionForExtent on ol.View to the API

### LINKED ISSUE #4573: view.getResolutionForExtent only works in OL_Debug Errors in ol.js
Enough said in the title

bartvde:
fixes #4573


ahocevar:
I don't think we need this. See my https://github.com/openlayers/ol3/issues/4573#issuecomment-167509927.

flipper44:
This ol.View#fit(Yields Different Results..
ol.View#fit() seems to not do a true fit, but must be taking tile Zoom level into acoount, whereas var res = view.getResolutionForExtent(initial_extent,size);
view.setResolution(res); sets the exact extents correctly for me

PR #4595, opened 2015-12-27
Pass 'opaque' option on to superclass

ahocevar:
The MapQuest sources do not benefit from the optimisations that are hinted by the `opaque` option. This pull request fixes that problem.


petrsloup:
Nice catch, LGTM

PR #4596, opened 2015-12-27
Use consistent tile coordinate keys

ahocevar:
Tile coordinate keys from ol.tilecoord.toString() are different than the ones
from Array.prototype.toString(). Both were used, potentially causing problems.
This change makes them consistent and removes the now unnecessary
ol.tileCoord.toString() function.


bartvde:
LGTM @ahocevar

bartvde:
The only visual change will be in the canvas tiles example where the key is displayed but this shouldn't matter

PR #4597, opened 2015-12-27
Render tiles directly to the map canvas

ahocevar:
This change removes the overhead of rendering tiles to an intermediate canvas.

There were two challenges with this approach:

1. **Avoid rendering of multiple resolutions of the same region for non-opaque layers.** To achieve this, we store a list of extents of all higher resolution tiles for each tile in the new `clipExtents` member. In `composeFrame`, we create a clip mask for each low resolution tile. The holes will be filled by higher resolution tiles.

2. **Avoid gaps between tiles.** ~~For non-rotated views, this is done by using the tile closest to the center as origin, and then adding tile widths and heights for other tiles. For rotated views, since we no longer can rotate the whole intermediate canvas, we render tiles with a 0.25 pixels overlap (0.5 pixels for opaque layers where tiny overlaps are not visible).~~ By rotating the map canvas after composing layers on an unrotated canvas, we can avoid gaps.


ahocevar:
@tschaub, you may want to have a look at this to make sure that my changes to `ol.source.Raster` make sense.

marcjansen:
I was looking forward to this! Thanks Andreas for this PR.

Do you see any chance for giving some sort of performance comparison before and after? Like fps comparisons e.g.?

Anyhow, I am really glad you tackle this.

ahocevar:
@marcjansen I did not do any accurate testing, just these using an updated version of http://openlayers.org/en/v3.1.1/resources/display-frame-rate.js: For very fast zooming through all zoom levels in the wms-tiled example: before this change, the frame rate drops to 30fps on my machine. After this change, it stays at 50-60fps. Very fast panning in the same example: 55fps before this change, and 60fps after this change. I think you'll be able to see more performance gains with your applications that have 10+ layers.

marcjansen:
@ahocevar: this sounds very good already! Thanks again. I am off duty so cannot look into it in detail, but I promise to test this early next year.

marcjansen:
This should not block this from being merged if someone else reviews it.

tonio:
With this change, it's not possible to listen to the `postcompose` event of a tile layer and have access to a canvas of just this layer (for *color* hit detection, for example), or I’m wrong?

ahocevar:
You are wrong @tonio. Like for vector and vector tile layers, a separate canvas is used when there are listeners for layer compose/render events.

tonio:
OK, good to know, thanks for the explanation.

tschaub:
Great to see you've taken this on @ahocevar.  I'll try to put some review time in early this week (I've been enjoying a chunk of computer-free vacation).

marcjansen:
As promised I tested this inside one of our applications. In that application ~25 single tile layers (sometimes even more) are rendered. The application is very laggy when panning and zooming (only ~7 fps). With this patch applied things do not change in a noticeable way, sadly. Nothing is broken, though.

ahocevar:
@marcjansen Do yo mean "single tile layers" as in ol.source.ImageWMS, or as in "distinct tile layers"? Because for the former, there are no changes with my patch - image layers are rendered to the map canvas directly already.

marcjansen:
@ahocevar I meant `ol.source.ImageWMS`… so it isn't surprising that no effect can be observed. Thanks for the clarification.

ahocevar:
@marcjansen Any chance you can change your app to use TileWMS sources and compare master with my patch?

marcjansen:
I will give feedback tomorrow. We have plenty of apps to change 😊.

marcjansen:
I just checked again with TileWMS sources (~25) and that particular application does not become noticeably faster, but I also couldn't find any detrimental effects. I was hopping to see a nice improvement, but the perceived laggyness didn't change in a noticeable way. Again, I don't consider this a blocker.

Is there any way we can have this measured reliably (with a script or sth.)?

ahocevar:
@tschaub I rebased this pull request, in case you still want to take a look.

ahocevar:
@tschaub I added another commit which implements what you had suggested in the hangout the other day - rotation of the map canvas after composition, instead of applying rotation for each layer. I'm pretty sure some transform can now be simplified, but I'd like to get at least a general ok before I spend even more time on this.

src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -105,86 +54,126 @@ goog.inherits(ol.renderer.canvas.TileLayer, ol.renderer.canvas.Layer);
...
+    this.context_.canvas.height = context.canvas.height;
+    renderContext = this.context_;
+  } else {
+    renderContext = context;
+  }
+  var offsetX = Math.round(pixelRatio * size[0] / 2);
+  var offsetY = Math.round(pixelRatio * size[1] / 2);
+
+  renderContext.translate(offsetX, offsetY);
+  //renderContext.rotate(rotation);
```
tschaub: This should be removed before merging.

src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -105,86 +54,126 @@ goog.inherits(ol.renderer.canvas.TileLayer, ol.renderer.canvas.Layer);
...
+        Math.round((insertPoint[0] - origin[0]) * pixelScale / tileWidth) *
+            tileWidth + tileOffsetX,
+        Math.round((origin[1] - insertPoint[1]) * pixelScale / tileHeight) *
+            tileHeight + tileOffsetY, tileWidth, tileHeight);
+    if (clipExtents) {
+      renderContext.restore();
+    }
+  }
 
+  //renderContext.rotate(-rotation);
```
tschaub: Another line to remove before merge.

src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -105,86 +54,126 @@ goog.inherits(ol.renderer.canvas.TileLayer, ol.renderer.canvas.Layer);
...
 };
 
 
 /**
  * @inheritDoc
  */
-ol.renderer.canvas.TileLayer.prototype.prepareFrame = function(frameState, layerState) {
-
-  //
-  // Warning! You're entering a dangerous zone!
```
tschaub: Happy to see this go :)

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -184,10 +184,10 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame = function(frameState,
       insertPoint = ol.geom.flat.transform.transform2D(
           ol.extent.getTopLeft(tileExtent), 0, 1, 2, insertTransform);
       replayContext.translate(offsetX, offsetY);
-      replayContext.rotate(rotation);
+      //replayContext.rotate(rotation);
```
tschaub: A couple more commented lines to remove here.

tschaub:
This looks great to me @ahocevar.  I haven't set up a comparison with many layers, but even if it isn't a big performance gain there, I think it is a good simplification.

ahocevar:
@tschaub Maybe you want to take another quick glance before I merge.

I was able to simplify transforms a bit. However, testing with all examples, I found that a transform needs to be applied now before dispatching layer pre-/postcompose events. I added that.

The layer-clipping example needed a few changes, because it made too many assumptions about the internal setup of the layer canvas. So I added a paragraph with code snippets to the upgrade notes.

Finally, I decided to optimize `ol.extent.getForViewAndSize` a bit to avoid garbage creation.

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,44 @@
 
 ### v3.14.0
 
+#### Layer pre-/postcompose event changes
+
+It is now the responsibility of the appliation to undo any canvas transform changes at the end of a layer 'precompose' or 'postcompose' handler. Previously, it was ok to set a null transform. The API now guarantees a device pixel coordinate system on the canvas with its origin in the top left corner of the map. However, applications should not rely on the canvas representing the visible viewport.
```
marcjansen: Typo: responsibility of the appli*c*ation

ahocevar:
Does anyone want to take another look at this?

marcjansen:
Great work, Andreas.

I will not have the time to honestly test this. From looking through the diff I really like the changes. Thanks for also including tests and upgrade notes. I will try to convince someone to throw this into our applications... but I cannot promise anything.

fredj:
LGTM

marcjansen:
Thanks Fred for jumping in! And thanks Andreas for providing the PR.

ahocevar:
Thanks for the reviews!

syntax42:
Are there a example page, where the improvement in rendering performance is illustrated? When i upgrade my project to using to 3.14.2, there is a real slowdown in perfomance. There more tilelayers the bigger difference in performance.

ahocevar:
@syntax42 Comparisons were made using the animation.html example, but that has only one tile layer. Reports from developers playing with many layers indicated that there is no significant performance gain, but also no slowdown. Are you able to share a map with many tile layers to compare?

Another thing that changed in 3.14.x is the way tiles are loaded. So the question is: are you seeing a slowdown in tile loading or in panning/zooming.

ahocevar:
> I also noticed a lot of artifacts when panning / zooming. And white lines between tiles where briefly visible.

This was an issue in 3.14.0 and 3.14.1 and should be fixed in 3.14.2. See #4960.

Another related issue still in 3.14.2, which only affects Firefox though, is #5024.

ptsagkis:
It seems that tile reprojection using chrome (on widnows OS) is not working as expected. The problem seems to exist since current fix (ol version 3.14 and above) . You may see the buggy attitude on  [this example](http://openlayers.org/en/v3.15.1/examples/reprojection-wgs84.html?q=reproject) using chrome on windows OS. Using Firefox no such problem exist. No errors in the console.

ahocevar:
@ptsagkis Did you do a `git bisect` to find out that this pull request causes #5236, or was this just a guess?

ptsagkis:
To be honest I am not very familiar with the technology behind.(I am not so sure what git bisect is) . It is a guess but it seems this change cause the problem as tile rendering has change since this fix and this where problems rised up. On version 3.13 there was no problem with the example above.

PR #4598, opened 2015-12-28
Make ol.source.Tile#getTilePixelRatio work correctly for all sources

### LINKED ISSUE #4583: Reprojection on TileArcGISRest Tiles ignores PixelRatio on Tablets and phones and loads at 1:pixelratio size in map
TileArcGISRest ignores PixelRatio on Tablets and phones and loads at 1:pixelratio size in map.
It seems that I have no issues with other tile Sources rendering correctly. 
When viewing on a Device with a pixel Ratio of 2 The tile comes in at 1/2 Height and 1/2 width (1/4 the area it should cover). It seems to be bringing in all the necessary data (image area) but just isn't being sized UP. This leaves a pattern of tiles spaced out on a map.

petrsloup:
This PR adds `pixelRatio` parameter to `ol.source.Tile#getTilePixelRatio` and it can be now overridden in subclasses (such as `ol.source.TileArcGISRest` and `ol.source.TileWMS`) instead of overriding `ol.source.Tile#getTilePixelSize`.

This makes the tile pixel ratio handling a little bit cleaner and fixes #4583.

### DIFF
--- a/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
@@ -94,7 +94,7 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame =
   var source = layer.getSource();
   goog.asserts.assertInstanceof(source, ol.source.VectorTile,
       'Source is an ol.source.VectorTile');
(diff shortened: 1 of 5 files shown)


ahocevar:
Very elegant solution. Please merge.

petrsloup:
Thanks.

PR #4602, opened 2016-01-02
Use ESLint

### LINKED ISSUE #3567: Stop using gjslint.py

tschaub:
This brings in ESLint as a replacement for gjslint.py and jshint.

We rely on the following changes that will be included in the ESLint 2 release:

 * https://github.com/eslint/eslint/commit/9ea6b366bdb78150ac4bd779a27841858c070b00
 * https://github.com/eslint/eslint/commit/5b7586bb2ea107d4097043d6b3362a24a0158483
 * https://github.com/eslint/eslint/commit/7c97201fdf10af4c0f06d3b9a7616da7927be123
 * https://github.com/eslint/eslint/commit/e5e06f817c5e4a50fbe0f1401e818ccd5d2fb701

In addition, it would be good to wait for (or work on) a fix for eslint/eslint#4900.

Fixes #3567.


tschaub:
Ok, so now ESLint is linting all that was run through gjslint.py.  As can be seen by the changes, ESLint is more strict about quite a lot with this rule configuration (params, args, and returns require descriptions, unused vars are not allowed, no implicit fall-through on switch/case, consistent brace style, etc.).  Still, I think it is easier to conform to than gjslint.py.

A few interesting "bugs" were revealed along the way.  For example, the `icon-sprite-webgl.js` example doesn't properly aggregate features (see if you can spot the [bug here](https://github.com/openlayers/ol3/blob/0f8373dd2d319ca4dd677d227ff3b0397d8d0556/examples/icon-sprite-webgl.js#L108-L128)).  Another odd one was `@return` [annotations](https://github.com/openlayers/ol3/blob/0f8373dd2d319ca4dd677d227ff3b0397d8d0556/src/ol/format/osmxmlformat.js#L131) on [functions](https://github.com/openlayers/ol3/blob/0f8373dd2d319ca4dd677d227ff3b0397d8d0556/src/ol/format/osmxmlformat.js#L146) without a `return` statement.

I've removed the 80 character line length limit (as discussed on our hangout).  This was forcing us to write some awkward multiline statements.  Instead, I think our convention should be "when a line feels like it is getting too long, wrap it."

Next I'm going to remove gjslint.py and jshint.

bartvde:
I'm +1 on this change, seems like a lot of work on your side @tschaub so thanks for tackling this

package.json
```diff
@@ -68,6 +70,22 @@
...
   },
+  "eslintConfig": {
+    "extends": "openlayers",
+    "globals": {
+      "ArrayBuffer": false,
+      "Uint16Array": false,
+      "Uint32Array": false,
+      "Uint8Array": false,
+      "Uint8ClampedArray": false,
+      "ol": false,
```
marcjansen: Shouldn't this be `"ol": true`? As we are writing to that object, or not?

marcjansen:
I am plus 1 as well. Thanks for the work on this, Tim.

tsauerwein:
:+1: Especially since gjslint is [discontinued](https://github.com/google/closure-linter/issues/81#issuecomment-170260192). Thanks for your effort, Tim! I like the idea of having the main linter configuration in a separate project, I am already making use of it in another project.

ahocevar:
+1, awesome! Thank you @tschaub.

package.json
tschaub: At this point I think we should only have implicit assignment with goog.provide, right?  When we get rid of that, ideally we'll be using es6 export (and perhaps have a build step that assigns to a global for a standalone build).  So I think we actually want to enforce that nobody tacks in arbitrary stuff to this right now.

marcjansen: Yeah, works for me. I was just wondering.

tschaub:
We can update the dependency to `eslint@2.0.0` when that is released.  I've submitted a proposed fix for eslint/eslint#4900, and the other fixes we need are already in the beta, so I'm going to merge this now.

Apologies to others will pull requests for which this will create conflicts.  Happy linting!

fredj:
Thanks Tim !

PR #4608, opened 2016-01-04
Fix type in vector source

gberaudo:
The closure compiler requires the type of the collection to be declared.
Otherwise this compilation error occurs:

```
ERR! compile
node_modules/openlayers/src/ol/source/vectorsource.js:174:
ERROR - actual parameter 1 of
ol.source.Vector.prototype.bindFeaturesCollection_ does not match formal
parameter
ERR! compile found   : (ol.Collection|undefined)
ERR! compile required: ol.Collection<(null|ol.Feature)>
ERR! compile     this.bindFeaturesCollection_(collection)
```

### DIFF
--- a/src/ol/source/vectorsource.js
+++ b/src/ol/source/vectorsource.js
@@ -157,7 +157,11 @@ ol.source.Vector = function(opt_options) {
    */
   this.featuresCollection_ = null;
 
-  var collection, features;
+  /**
+   * @type {!ol.Collection.<ol.Feature>}
+   */
+  var collection;
+  var features;
   if (options.features instanceof ol.Collection) {
     collection = options.features;
     features = collection.getArray();


gberaudo:
Fixed Travis.

tschaub:
Is the compiler still confused about the type if the `bindFeaturesCollection_` call looks like this:

```js
if (collection) {
  this.bindFeaturesCollection_(collection);
}
```
?

gberaudo:
Indeed @tschaub, it also fixes the compiler issue! Thanks.
The small drawback I see is that it introduces a bit of inconsistency with the coding style of the function.
But is it a problem? If no I can update the PR immediately.

gberaudo:
Updating the compiler fixed the issue.

PR #4612, opened 2016-01-04
Set oli.source.VectorEvent#feature to ol.Feature|undefined

fredj:
This commit remove incoherency between the class and the interface.

See https://github.com/openlayers/ol3/blob/master/src/ol/source/vectorsource.js#L865


tschaub:
Please merge.

fredj:
Thanks for the review

PR #4613, opened 2016-01-04
Include own layer in layerFilter and only select unselected features

### LINKED ISSUE #4607: PointerMove Interaction styles flicker when layers option is specified
Styles in select Interaction for pointerMove seems to flicker on/off when you keep moving the mouse over the same item. This only happens if the layers option is specified.

See fiddle: http://jsfiddle.net/6ysd6s2q/6/

ahocevar:
When selecting an already selected feature, it will be on the Select
interaction's featureOverlay_. So we need to include that featureOverlay_
in the layer filter, regardless of what the user set as layer filter.

When in toggle mode, we need to make sure that we only select features that
are not already included in the selection.

Fixes #4607


fredj:
LGTM, thanks

PR #4619, opened 2016-01-05
Add option to load TileJSON via XHR

petrsloup:
Some TileJSON sources (including Mapbox v4) do not allow to use JSONP (with `?callback=...`) to load the TileJSON.

This PR adds option to specify that XmlHttpRequest should be used instead.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5179,6 +5179,7 @@ olx.source.TileArcGISRestOptions.prototype.urls;
  *     reprojectionErrorThreshold: (number|undefined),
  *     tileLoadFunction: (ol.TileLoadFunctionType|undefined),
  *     url: string,
+ *     useXhr: (boolean|undefined),
  *     wrapX: (boolean|undefined)}}
  * @api
  */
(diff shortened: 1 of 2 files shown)


externs/olx.js
```diff
@@ -5179,6 +5179,7 @@ olx.source.TileArcGISRestOptions.prototype.urls;
  *     reprojectionErrorThreshold: (number|undefined),
  *     tileLoadFunction: (ol.TileLoadFunctionType|undefined),
  *     url: string,
+ *     useXhr: (string|undefined),
```
bartvde: boolean?

petrsloup: Oh, of course. Fixed, thanks.

bartvde:
LGTM @petrsloup I know this extends on our use of the goog lib, but the areas you are adding have not been refactored out as yet so I guess this is okay for now. So please merge.

petrsloup:
Yes, this uses `goog.net.XhrIo`, but it's already used in `ol.featureloader.loadFeaturesXhr` so it's not a new dependency.

Merged, thanks.

tschaub:
If the Mapbox v3 endpoint supports CORS, I think it would be reasonable to change the default to use XHR.  I think we risk frustrating more people by being broken in the future unless someone discovers the `useXhr` option.

What would others think about having a `jsonp` option, and having the default be `false` (treating the previous behavior as a bug)?

fredj:
jsonp only exists because of IE9?

+1 for using xhr by default

petrsloup:
I agree it makes sense to change the default now. See #4625.

PR #4625, opened 2016-01-06
Use XHR by default to load TileJSON in ol.source.TileJSON

petrsloup:
Based on comments in #4619 this PR makes XHR the default method of loading TileJSONs in `ol.source.TileJSON`. This ensures better future compatibility (especially with Mapbox API v4), although it might require changes to some existing applications when upgrading (see upgrade notes).

Some Mapbox v3 urls had to be changed from `.jsonp` (these actually does not seem to be mentioned anywhere in to documentation https://www.mapbox.com/developers/api/v3/#cors-and-jsonp) to `.json`.

This adds new closure dependency `goog.net.CorsXmlHttpFactory`, which internally uses `XDomainRequest` to ensure IE9 compatibility. This can be safely dropped in the future when we no longer support IE9 (possibly soon? https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support).

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5206,10 +5206,10 @@ olx.source.TileArcGISRestOptions.prototype.urls;
 /**
  * @typedef {{attributions: (Array.<ol.Attribution>|undefined),
  *     crossOrigin: (null|string|undefined),
+ *     jsonp: (boolean|undefined),
  *     reprojectionErrorThreshold: (number|undefined),
  *     tileLoadFunction: (ol.TileLoadFunctionType|undefined),
  *     url: string,
- *     useXhr: (boolean|undefined),
  *     wrapX: (boolean|undefined)}}
  * @api
  */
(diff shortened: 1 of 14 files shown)


externs/olx.js
```diff
@@ -5209,6 +5209,15 @@ olx.source.TileJSONOptions.prototype.crossOrigin;
 
 
 /**
+ * Use JSONP with callback to load the TileJSON. Useful when the server
+ * does not support CORS. * Default is `false`.
```
tschaub: It looks like an extraneous `*` here.

tschaub:
Looks good to me.  Thanks for the nice docs and upgrade notes on this.  One minor comment above, but otherwise I think we should merge.

fredj:
LGTM too, thanks @petrsloup

petrsloup:
@tschaub @fredj Thanks for the reviews!

PR #4626, opened 2016-01-06
Fixed documentation error

simonseyock:
The imageLoad function option of all ol.source.Image subclasses are not of type ol.TileLoadFunctionType but of type ol.ImageLoadFunctionType.


fredj:
Thanks!

PR #4630, opened 2016-01-06
Simplify meters per unit handling

### LINKED ISSUE #4627: Use correct degrees -> meters conversion
When calculating resolutions, `ol.proj.Projection#getMetersPerUnit` is used, which uses the `ol.proj.METERS_PER_UNIT` lookup table. For `degrees`, the meters per unit depend on the sphere used by the projection. OpenLayers uses the radius of the `ol.sphere.NORMAL` sphere, unless the projection is auto-configured from proj4js and has a `to_meter` property. In this case, a pseudo entry is added to the `ol.proj.METERS_PER_UNIT` lookup. The default for EPSG:4326 is an incorrect value, which is problematic when using `ol.tilegrid.WMTS.createFromCapabilitiesMatrixSet` to calculate the resolutions for a WMTS layer.

My suggestion is to add a `metersPerUnit` option to `ol.proj.Projection`. This option should be set when configuring a projection from proj4js, and `ol.proj.EPSG4326` should use the correct value (`111319.49079327358`) instead of deriving it from the radius of `ol.sphere.NORMAL` (`111194.87428468118`).

ahocevar:
Fixes #4627.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -429,6 +429,7 @@ olx.OverlayOptions.prototype.autoPanMargin;
  *     extent: (ol.Extent|undefined),
  *     axisOrientation: (string|undefined),
  *     global: (boolean|undefined),
+ *     metersPerUnit: (number|undefined),
  *     worldExtent: (ol.Extent|undefined),
  *     getPointResolution: (function(number, ol.Coordinate):number|undefined) }}
  * @api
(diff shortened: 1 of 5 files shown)


externs/olx.js
```diff
@@ -477,6 +478,15 @@ olx.ProjectionOptions.prototype.global;
 
 
 /**
+ * The meters per unit for the SRS. If not provided, the `units` are used to get
+ * the meters per unit from the {@link ol.proj.METERS_PER_UNIT} lookup table.
+ * @type {number}
```
fredj: should be `{number|undefined}`

fredj:
LGTM, thanks

PR #4632, opened 2016-01-06
Overloading fill color (polygon or text) with CanvasPattern and CanvasGradient

bjnsn:
An alternative to: https://github.com/openlayers/ol3/pull/4241 per https://github.com/openlayers/ol3/pull/4241#issuecomment-169143986

**Notes**
This creates a new 'ColorLike' typedef which encompasses types string, CanvasPattern and CanvasGradient. In theory this could also be used for stroke colors (canvas context strokeStyle), although that isn't necessary.

Also, I'm not sure of the purpose of fill style getCheckSum(), so that might need updates.

### DIFF
--- a/src/ol/colorlike.js
+++ b/src/ol/colorlike.js
@@ -1,0 +1,41 @@
+goog.provide('ol.ColorLike');
+goog.provide('ol.colorlike');
+
+goog.require('ol.color');
+
+
+/**
+ * A type accepted by CanvasRenderingContext2D.fillStyle.
+ * Represents a color, pattern, or gradient.
+ *
+ * @typedef {string|CanvasPattern|CanvasGradient}
+ * @api
+ */
+ol.ColorLike;
+
+
+/**
+ * @param {ol.Color|ol.ColorLike} color Color.
+ * @return {ol.ColorLike} The color as an ol.ColorLike
+ * @api
+ */
+ol.colorlike.asColorLike = function(color) {
+  if (ol.colorlike.isColorLike(color)) {
+    return /** @type {string|CanvasPattern|CanvasGradient} */ (color);
+  } else {
+    return ol.color.asString(/** @type {ol.Color} */ (color));
+  }
+};
+
+
+/**
+ * @param {?} color The value that is potentially an ol.ColorLike
+ * @return {boolean} Whether the color is an ol.ColorLike
+ */
+ol.colorlike.isColorLike = function(color) {
+  return (
+      goog.isString(color) ||
+      color instanceof CanvasPattern ||
+      color instanceof CanvasGradient
+  );
+};
--- a/src/ol/style/fillstyle.js
+++ b/src/ol/style/fillstyle.js
@@ -1,5 +1,6 @@
 goog.provide('ol.style.Fill');
 
+goog.require('ol.ColorLike');
 goog.require('ol.color');
 
 
@@ -17,7 +18,7 @@ ol.style.Fill = function(opt_options) {
 
   /**
    * @private
-   * @type {ol.Color|string}
+   * @type {ol.Color|ol.ColorLike}
    */
   this.color_ = options.color !== undefined ? options.color : null;
 
@@ -31,7 +32,7 @@ ol.style.Fill = function(opt_options) {
 
 /**
  * Get the fill color.
- * @return {ol.Color|string} Color.
+ * @return {ol.Color|ol.ColorLike} Color.
  * @api
  */
 ol.style.Fill.prototype.getColor = function() {
@@ -42,7 +43,7 @@ ol.style.Fill.prototype.getColor = function() {
 /**
  * Set the color.
  *
- * @param {ol.Color|string} color Color.
+ * @param {ol.Color|ol.ColorLike} color Color.
  * @api
  */
 ol.style.Fill.prototype.setColor = function(color) {
@@ -56,8 +57,15 @@ ol.style.Fill.prototype.setColor = function(color) {
  */
 ol.style.Fill.prototype.getChecksum = function() {
   if (this.checksum_ === undefined) {
-    this.checksum_ = 'f' + (this.color_ ?
-        ol.color.asString(this.color_) : '-');
+    if (
+        this.color_ instanceof CanvasPattern ||
+        this.color_ instanceof CanvasGradient
+    ) {
+      this.checksum_ = goog.getUid(this.color_).toString();
+    } else {
+      this.checksum_ = 'f' + (this.color_ ?
+          ol.color.asString(this.color_) : '-');
+    }
   }
 
   return this.checksum_;
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -6158,16 +6158,17 @@ olx.style.CircleOptions.prototype.atlasManager;
 
 
 /**
- * @typedef {{color: (ol.Color|string|undefined)}}
+ * @typedef {{color: (ol.Color|ol.ColorLike|undefined)}}
  * @api
  */
 olx.style.FillOptions;
 
 
 /**
- * Color. See {@link ol.color} for possible formats. Default null; if null,
- * the Canvas/renderer default black will be used.
- * @type {ol.Color|string|undefined}
+ * A color, gradient or pattern. See {@link ol.color} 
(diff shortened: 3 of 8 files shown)


src/ol/colorlike.js
```diff
@@ -0,0 +1,42 @@
...
+goog.provide('ol.colorlike');
+
+goog.require('ol.color');
+
+
+/**
+ * A type accepted by CanvasRenderingContext2D.fillStyle
+ *
+ * @typedef {string|CanvasPattern|CanvasGradient}
+ * @api
```
tschaub: Let's remove the `@api` annotation from here for now.

src/ol/colorlike.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+ol.ColorLike;
+
+
+/**
+ * Return the color as an ol.ColorLike
+ * @param {ol.Color|ol.ColorLike} color Color.
+ * @return {ol.ColorLike}
+ * @api
```
tschaub: Same here - this does not need to be exportable for now (no `@api` annotation).

src/ol/colorlike.js
tschaub: Actually, for doc purposes, you can leave the `@api` annotation here.

tschaub:
Thanks for the work on this @bjnsn.

Please also update the `olx.js` externs file to describe the new types for the `color` property of `olx.style.FillOptions`.  There are two places where you need to make changes.

First, [the `@typedef`](https://github.com/bjnsn/ol3/blob/fill-color-overload/externs/olx.js#L6061-L6065) for `olx.style.FillOptions` needs to be updated.  This `@typedef` is used by the compiler to ensure that fill options objects have properties of the correct type (if people are compiling an application together with the library) and that we respect the types of those properties where they are used in the library.

Next, the [type for the `color`](https://github.com/bjnsn/ol3/blob/fill-color-overload/externs/olx.js#L6068-L6074) property needs to be updated (and documented).  This is a bit redundant, but it serves two purposes.  The descriptions here show up in the API docs - we cannot use JSDoc to describe properties in a `@typedef`, so we need a separate place for this.  In addition, if an application is compiled together with the library, the compiler is able to rename properties in the calling code and where the properties are accessed in the library.

Since the `ol.ColorLike` type is marked as `@api`, it will appear in the API docs.  So the type for the `color` property in the `olx.js` file can be `ol.Color|ol.ColorLike`, and people will be able to click through to see what those types expand to from the [`ol.style.Fill` doc](http://openlayers.org/en/master/apidoc/ol.style.Fill.html).

src/ol/style/fillstyle.js
```diff
@@ -57,8 +58,14 @@ ol.style.Fill.prototype.setColor = function(color) {
  */
 ol.style.Fill.prototype.getChecksum = function() {
   if (this.checksum_ === undefined) {
-    this.checksum_ = 'f' + (this.color_ ?
-        ol.color.asString(this.color_) : '-');
+    if (this.color_ instanceof CanvasPattern) {
+      this.checksum_ = 'fCanvasPattern';
```
gberaudo: You should be able to use `goog.getUid(this.color_)` for the canvas pattern and gradient.

gberaudo:
I added a comment about checksum computation. Otherwise +1.

ahocevar:
@bjnsn Thanks for your efforts on this. Would you be able to rebase the pull request, so it does not conflict with master any more? I would like to give it a final review.

ahocevar:
Thanks @bjnsn for squashing the commits into one. I think something went wrong with the commit message though. Would be great if you could clean that up too. Should be easy to do with a `git commit --amend`.

ahocevar:
@bjnsn Travis got stuck on this. Can you please push the last commit again, and maybe fix the commit message? If not let me know, then I'll close this and create a new pull request with a cherry-pick of your change.

bjnsn:
@ahocevar Should be fixed now.

ahocevar:
Thanks @bjnsn for your patience!

phelanHayes:
Thanks for your work on this @bjnsn!  Would it possible to provide a sample of how to implement this functionality in ol 3.15.1 to create hashing for polygon fills?  I don't see a renderer attribute under the Fills Style. I am currently using the createStripesRender factory function with the hacked ol-debug from the previous posts.

bjnsn:
@phelanHayes http://jsfiddle.net/61b7szjn/

That is based on http://openlayers.org/en/master/examples/image-vector-layer.html - only important differences are the makePattern function & the pattern being used instead of a color string in the first ol.style.Fill instantiation.

bjnsn:
For completeness, here is the same example as the pattern one above, but using a canvas gradient to overload the fill color instead of a canvas pattern: http://jsfiddle.net/07366L44/

ca0v:
Am I correct in understanding this does not work for actual image data?  This ArcGIS symbol definition cannot be expresses with an openlayers style?

```
        {
            "type": "esriPFS",
            "url": "78ff6f6cb8ad6ca6d96eed86217d46ca",
            "imageData": "iVBORw0KGgoAAAANSUhEUgAAADgAAAA4CAYAAACohjseAAAAAXNSR0IB2cksfwAAAAlwSFlzAAAOxAAADsQBlSsOGwAAAyRJREFUaIG1WkGSAyEIJFX+JP9/Ut6SvaxVhkDTDYZTZkREBBqcrJfZ2/7pafYwQJu34rsxl+Wv9F/nSy/UP3cM8DJ7P80ep0wvJ1JSMSLiXYhR3VDEz8h8mj3OTfo1IoOxOi40qJ6o4lasMRWjb7mnHnCD6okid0dyog13KIxBvxBiVniRW53zz/hk5Veecr6HMYgUrDZQyVW8QZ170mLiJovBbMEbLqd4E1p3KYqwp8bAQKTcOc7Ec5VRX2bvFTEzJzDFKQVjWUNG4yti7gS96pJdF64SlueHMOEVysBY2XCmIOs9asIKNxgBZiWIXbBS8IZ3lEBfbagqnyp+ZozJ7tncDxxEGSqawDwr9aKSQJis6Q2fwgTjFiwuMnMYUjP7h4uiQI+K7qjEUoqGjCp3jvRA8ymYYF2SPfWuARhP+apkKoUYJZUkwcTWTYIwkT0zqZztRjrdBJJJnWCkrFp7KjDCyIyIkbtYS01qzypW1D40w+mw2FYBGI13ryymhTuiECYyQVXcsAlkgrEKb9pNqIsw8YlkMTcGSBbCRQkmJlUPcudOhcLqkbpopXingL4hO+PPZFI4yCqIbrYrBdmkox6E1NFPTgXJmmTRCnOhi7LVhoqf0yyK9NpFyf4Nrw0rnPtF/agU7hl9nKCSKDqB7+d1ugnmSjGb2+omvKI3jHGrhdry9ljqop1KhlFYqYD22KRIaN9so/eTyieSMWkCwq9LVSqe0sQALK7u39TXJcY1bt58T07M80AcREoqbslkxsk9KFWqVVZXwJnh72TdqpuI9GnXolNoUGV31klhgom5bt0ZvWNld5IdDRPd01GTgid0B7PHvZzy44ufzKblW8aqNlXJCrNoR1G0yOQbxE2i/+mk3qJN4gitqxYB1L8sGEW6c9SMiWrTEAd/ETdKtvuFYdMk0/keEPGrMckahOVLkwxSUF1M6f5Zt1WLj69KBgWvH+9mx5udCVMfUx1950Qnm7gZv2EWvfm568bGfXevyAizKJsIuteGjOxMhtpO0ZUM8+6XN2kMwX4QKTyNCRXcWWLkSBe/XjjLq8id8vhN/wHAM/YoQRgDtgAAAABJRU5ErkJggg==",
            "contentType": "image/png",
            "outline": {
                "type": "esriSLS",
                "style": "esriSLSSolid",
                "color": [255, 0, 0, 255],
                "width": 1
            },
            "width": 42,
            "height": 42
        }
```

ahocevar:
I see no reason why that would not work @ca0v. Try something like:
```js
var context = canvas.getContext('2d');
var img = new Image();
img.onload = function() {
  var pattern = context.createPattern(img, 'repeat');
  style.getFill().setColor(pattern);
}
img.src = 'data:' + contentType + ';base64,' + imageData;
```

PR #4638, opened 2016-01-08
Do not rely on projection extent

### LINKED ISSUE #4634: Bing Attribution Crashes when Map layers are reprojected.
Not sure Why this Happens. This does not happen With OSM or MapQuest (as far as I have tested).
This Happens in a very odd place in ol-debug as well (see attached image)
capture4

Can likely be tested in the new Indiana State plane East Sample that is to be merged later

ahocevar:
`ol.Attribution` should not require a projection with a configured extent.

Fixes #4634.


bartvde:

Issue #4542, opened 2015-12-10
Remove use of goog.bind


fredj:
Use `Function.prototype.bind` instead of `goog.bind`

fredj:
#4639

PR #4639, opened 2016-01-08
Remove use of goog.bind and use ES5 .bind

nicholas-l:
Changes all of the uses of goog.bind to the ES5 Func.bind. I created this as a quick pull request to reduce the use of closure compiler. Hope this is where the project is heading :+1:

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -11,7 +11,7 @@
     <link rel="stylesheet" href="./resources/layout.css" type="text/css">
     {{{ extraHead.local }}}
     {{{ css.tag }}}
-    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch"></script>
+    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Function.prototype.bind|gated"></script>
     <script src="./resources/zeroclipboard/ZeroClipboard.min.js"></script>
     <title>{{ title }}</title>
   </head>
--- a/examples/geolocation-orientation.html
+++ b/examples/geolocation-orientation.html
@@ -13,6 +13,7 @@ tags: "fullscreen, geolocation, orientation, mobile"
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" type="text/css">
     <link rel="stylesheet" href="../css/ol.css" type="text/css">
     <link rel="stylesheet" href="./resources/layout.css" type="text/css">
+    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Function.prototype.bind|gated"></script>
     <title>Mobile Geolocation Tracking with Orientation</title>
     <style type="text/css">
       html, body, .map {
--- a/examples/mobile-full-screen.html
+++ b/examples/mobile-full-screen.html
@@ -13,6 +13,7 @@ cloak:
     <meta http-equiv="X-UA-Compatible" content="chrome=1">
     <meta name="viewport" content="initial-scale=1.0, user-scalable=no, width=device-width">
     <title>Mobile full screen example</title>
+    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Function.prototype.bind|gated"></script>
     <style type="text/css">
       html, body, .map {
         margin: 0;
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -1763,8 +1763,8 @@ ol.format.KML.prototype.readDocumentOrFolder_ = function(node, objectStack) {
         'Document': ol.xml.makeArrayExtender(this.readDocumentOrFolder_, this),
         'Folder': ol.xml.makeArrayExtender(this.readDocumentOrFolder_, this),
         'Placemark': ol.xml.makeArrayPusher(this.readPlacemark_, this),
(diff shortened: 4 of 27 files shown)


tschaub:
Thanks for the work on this @nicholas-l.  This does help remove the dependency on the Closure Library (not the Compiler).

The change will involve a bit more work.  You can see that the [build fails](https://travis-ci.org/openlayers/ol3/builds/101032489) due to some indentation and documentation issues in `webglmaprenderer.js`.  If you're motivated to fix these, you can run the same checks locally with `make check` - this saves you from waiting for Travis to learn about linter errors.

In addition, we may need to include a shim in the tests (PhantomJS 1.x is missing `Function.prototype.bind`) - though I haven't confirmed that there is not one there already.

nicholas-l:
Oops meant the Closure Library. What is the roadmap for the removal of closure library and/or compiler?

Also yes I have hit the need for a shim of .bind. Is there a reason that PhantomJS 2.0 is not being used?

I had to add a return false to this.loadNexTileTexture_ to conform with the PostRenderFunction, is this the correct thing to do? Otherwise it would return undefined.

fredj:
Fixes #4542

fredj:
@nicholas-l can you please rebase your work on master ?

nicholas-l:
Had an old version (current stable) of eslint installed in node_modules so I was getting the wrong errors. Really great that Openlayers is moving to eslint :).

nicholas-l:
Also the above questions, mainly phantomjs? @tschaub is the a plan to move to phantomjs 2.0 or is there something blocking it?

nicholas-l:
Ok, might have found a problem with the latest eslint (though as you can imagine im not entirely confident with JSDoc), closure compiler wants 
```
(/**
JSDoc
*/ function (){
}).bind(this);
```
and eslint wants this:
```
(/**
JSDoc
*/ function (){
}.bind(this));
```
Note the braket placement.
Thoughts @tschaub ?

fredj:
Are the parentheses required ?
```
/**
 * JSDoc
 */ 
function() {

}.bind(this);
```

ahocevar:
> Also the above questions, mainly phantomjs? @tschaub is the a plan to move to phantomjs 2.0 or is there something blocking it?

We use the mocha-phantomjs-core package, which still relies on phantomjs 1.x in its latest version.

nicholas-l:
I think so as I get this error when I remove them:
```
ERR! compile /Users/nicholas/Development/openlayers-dev/src/ol/renderer/webgl/webglmaprenderer.js:136: ERROR - Misplaced function annotation. This JSDoc is not attached to a function node. Are you missing parentheses?
ERR! compile       function(element) {
ERR! compile       ^
ERR! compile 
ERR! compile 
ERR! compile 
ERR! compile 1
ERR! compile  error(s), 0 warning(s)
ERR! compile 
```

ahocevar:
```js
/**
 * JSDoc
 */ 
(function() {

}).bind(this);
```
satisfies both ESLint and the compiler.

nicholas-l:
Ah thanks @ahocevar also doesn't mention much about PhantomJS 2.0 on mocha-phantomjs-core expect at the bottom where they mention testing with both 1.9.7 and 2.0.0. Possibly works with it?

fredj:
@nicholas-l can you also squash all your commits (see https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit). thanks

gberaudo:
FWIW, mocha-phantomjs-core are actually doing Travis tests with PhantomJS 2.0, see
https://travis-ci.org/nathanboktae/mocha-phantomjs-core/builds/93123439
It was already the case for their latest release.

ahocevar:
> Ah thanks @ahocevar also doesn't mention much about PhantomJS 2.0 on mocha-phantomjs-core expect at the bottom where they mention testing with both 1.9.7 and 2.0.0. Possibly works with it?

I did not remember correctly what the problem was. I just tried to use the phantomjs2 package instead of phantomjs, and the reason for phantomjs2 not working for us is ariya/phantomjs#12900 and the phantomjs2 package not yet using 2.0.1. So some upstream fixes or pull requests would be necessary before we can use phantomjs2.

nicholas-l:
Is it ok to add a polyfill to the examples? I see there already is one there for the Fetch API. I will squash the commits down if it is ok :)

nicholas-l:
Ok @fredj all squashed and passing. Only major things with this pull request is:
 - Adding Function.prototype.bind to examples
 - Adding a return false to loadNextTileTexture_ in ol.renderer.webgl.Map
 - Removing goog.bind :+1:

config/examples/example.html
```diff
@@ -11,7 +11,7 @@
     <link rel="stylesheet" href="./resources/layout.css" type="text/css">
     {{{ extraHead.local }}}
     {{{ css.tag }}}
-    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch"></script>
+    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Function.prototype.bind|gated"></script>
```
tschaub: From http://cdn.polyfill.io/v2/docs/api:
> `gated` is no longer supported. Use `flags` instead.

tschaub:
This looks great.  Thanks for the contribution @nicholas-l.

PR #4641, opened 2016-01-11
Remove extra options from icon example

tschaub:
Before:

image

After:

image

I'd rather see us add dedicated examples that showcase specific options where they are useful.  In this case, I think `color` and `opacity` make for an uglier example.


marcjansen:
I agree. Looks good to me, please merge.

PR #3988, opened 2015-08-10
Remove bin/check-requires.py

fredj:
The `missingProvide` and `missingRequire` compiler checks are already checking
if the `goog.require` and `goog.provide` are correct.

See https://github.com/openlayers/ol3/pull/3960#issuecomment-127635513

Comments welcome

PR #4644, opened 2016-01-11
Remove check-requires script

fredj:
The `missingProvide` and `missingRequire` compiler checks are already checking
if the `goog.require` and `goog.provide` are correct.

Replace #3988


fredj:
There's something wrong with this; the compiler doesn't detect missing `goog.require`.
For example, if `goog.require('goog.array');`is removed from `src/ol/map.js` no error is raised

ahocevar:
> if goog.require('goog.array');is removed from src/ol/map.js no error is raised

I think this is because no code path that calls `ol.Map#removePreRenderFunction()` is ever entered, and the compiler detects that.

That said, I'm still in favour of merging this.

fredj:
`missingRequire` only checks the constructors

ahocevar:
Does this mean that you can actually get a broken build from missing requires?

fredj:
> Does this mean that you can actually get a broken build from missing requires?

I don't think so; the compiler will emit an error if something is missing.

Are you still ok to merge this?

probins:
Can this be merged? I'm running into problems with `typedefs.js`, as it thinks the sub-namespaces need to be `require`d. I could add a condition to ignore `typedefs.js`, but it would be simpler if check-requires were removed.

ahocevar:
+1 on merging this.

fredj:
Thanks for the review

PR #4647, opened 2016-01-11
Proper rounding for possibly negative numbers

### LINKED ISSUE #4645: Icon moves when it touches the left or top border
If you move the map so that an icon touches the left or top border, the icon will be repositioned. (I guess it's by one pixel.)

You can reproduce it with this example:
http://openlayers.org/en/v3.12.1/examples/icon.html

ahocevar:
Because image icons may be positioned with negative coordinates when they are only partially visible at the left or top edge of the map, we need to properly round them when snapping to pixels.

Fixes #4645.


ahocevar:
The performance impact of this change is negelctable: https://jsperf.com/math-round-vs-hack/93.

fredj:
LGTM, good catch

ahocevar:
Oh, the render test tested for an incorrect image. Let me fix that.

PR #4649, opened 2016-01-12
Missing parameter description for ol.source.Tile#getOpaque

petrsloup:
Minor jsdoc fix (error revealed by ESLint https://github.com/openlayers/ol3/pull/4602).


petrsloup:
Trivial jsdoc fix, so merging..

PR #4652, opened 2016-01-12
Use map.getTargetElement instead of map.getTarget

fredj:
Small simplification in `ol.control.FullScreen`


ahocevar:
Looks good. Please merge when lights are green.

fredj:
Thanks for the review

PR #4666, opened 2016-01-12
Add a 'boxdrag' event to ol.interaction.DragBox. Resolves #4563 .

### LINKED ISSUE #4563: Add a DragBoxEvent for dragging, not just start and end.
ol.interaction.DragBox does not fire an event for dragging, only start and end of a drag. I propose adding a new DragBoxEventType, 'boxdrag,' to handle this case. 

It is possible we may want to update a selection live, not just after the user has finished dragging. This allows for instant feedback and the perception of a more responsive interface.

pkovac:


fredj:
LGTM, thanks

ahocevar:
Looks good, thanks!

PR #4667, opened 2016-01-12
Fix rendering of transparent MapQuest layer type

### LINKED ISSUE #4660: Mapquest Image and Labels (Hybrid) has some sort of odd caching issue. See Example
http://openlayers.org/en/master/examples/mapquest.html

Switch Mode to Aerial with Labels and zoom in and out.

petrsloup:
This seems like an error exposed by https://github.com/openlayers/ol3/pull/4595 -- the `'hyb'` layer type is actually not opaque.

Fixes #4660.

### DIFF
--- a/src/ol/source/mapquestsource.js
+++ b/src/ol/source/mapquestsource.js
@@ -40,7 +40,7 @@ ol.source.MapQuest = function(opt_options) {
     logo: 'https://developer.mapquest.com/content/osm/mq_logo.png',
     maxZoom: layerConfig.maxZoom,
     reprojectionErrorThreshold: options.reprojectionErrorThreshold,
-    opaque: true,
+    opaque: layerConfig.opaque,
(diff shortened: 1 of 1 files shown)


ahocevar:
Good catch and elegant fix. Please merge.

ahocevar:
Wait, isn't the ' hyb'  layer a layer group of 2 layers? The first one in the group should be opaque.

ahocevar:
You can merge @petrsloup. The group is created in the example. 'hyb'  is just the overlay layer, which is non-opaque and correctly fixed here.

petrsloup:
@ahocevar Thanks for the review. I think it's quite unfortunate naming on MapQuest's side -- the layer type is named 'hyb' ~ hybrid, but it actually contains only the overlay data (https://otile3-s.mqcdn.com/tiles/1.0.0/hyb/10/300/385.jpg).

PR #4668, opened 2016-01-13
Added updateSketchCondition as a Draw Interaction option

bjornharrtell:
Enables the developer to specify on which browser event the sketch is updated.

This is a squashed and rebased version of #4556 which is a proposed fix for #3907.

One could debate if the current default behavior for the mobile case should be changed, but at least this makes it possible to change it.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2658,6 +2658,16 @@ olx.interaction.DrawOptions.prototype.geometryName;
  */
 olx.interaction.DrawOptions.prototype.condition;
 
+/**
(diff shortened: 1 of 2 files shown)


bjornharrtell:
Fixed missing "by" in documentation as noted by @tsauerwein, thanks.

tschaub:

bjornharrtell:
I've rebased the branch but have no rights to reopen this PR.

ahocevar:
@bjornharrtell Thanks for the branch update. Can you please create a new pull request from that branch? GitHub does not let us reopen this pull request.

PR #4581, opened 2015-12-18
Do not render in handleDownEvent

ahocevar:
Does anybody know why we call `map.render()` in `handleDownEvent_` in our interactions? To me, this does not make much sense, because nothing has happened yet at this point, so rendering should be pretty useless. Any other opinions? Or am I missing something? If not, we might consider merging this.

PR #4675, opened 2016-01-14
Force a map render when the view is not updated

### LINKED ISSUE #4674: Tiles loading issue
- http://openlayers.org/en/master/examples/simple.html
 - pan

-> the tiles are not loaded

The issue is not present in the latest release (v3.12.1)

fredj:
Fixes #4674

### DIFF
--- a/src/ol/interaction/dragpaninteraction.js
+++ b/src/ol/interaction/dragpaninteraction.js
@@ -117,6 +117,9 @@ ol.interaction.DragPan.handleUpEvent_ = function(mapBrowserEvent) {
       ]);
       dest = view.constrainCenter(dest);
       view.setCenter(dest);
+    } else {
+      // the view is not updated, force a render
+      map.render();
     }
     view.setHint(ol.ViewHint.INTERACTING, -1);
     return false;


fredj:
Thanks for any review

ahocevar:
It would be interesting to understand why this is needed: why is the animating hint reset properly, but not the interacting hint?

If it fixes the issue, then please merge. A nicer fix would be a better way to notify renderers of view hint changes.

ahocevar:
Do you think #4581 breaks other things too @fredj? If so, another solution would be to revert my changes from #4581 for now, and then implement something where renderers can get notified of hint changes without forcing a render.

fredj:
I've only found the issue with the pan without kinetic but I didn't test everything.

fredj:
ok, let's revert #4581 and find a better solution

tschaub:
Late to the part here, but I found this a pretty nice solution.  Did anybody find other issues with #4581?

ahocevar:
I tested mouse rotation and did not find any issues with #4581.

fredj:
https://github.com/openlayers/ol3/issues/4674#issuecomment-171569706 from @petrsloup :
"Yes, this seems to be happening for me only when zooming and dragging *simultaneously*."

ahocevar:
> "Yes, this seems to be happening for me only when zooming and dragging simultaneously."

How do I do that? I tested on my mobile now: dragging and pinch-zooming simultaneously works fine.

PR #4677, opened 2016-01-15
Make pixel projection handling work in compiled mode

### LINKED ISSUE #4676: Vector tiles rendering broken in compiled build when pixel units are used
See http://openlayers.org/en/master/examples/mapbox-vector-tiles.html.

ahocevar:
The compiler seems to have trouble with functions that can run with
different 'this' types. By moving the units handling to the callback, this
issue can be avoided.

Fixes #4676.

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -104,18 +104,9 @@ ol.featureloader.loadFeaturesXhr = function(url, format, success, failure) {
                   goog.asserts.fail('unexpected format type');
                 }
                 if (source) {
-                  if (ol.ENABLE_VECTOR_TILE && this instanceof ol.VectorTile) {
-                    var dataUnits = format.readProjection(source).getUnits();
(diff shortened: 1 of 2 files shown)


src/ol/featureloader.js
```diff
@@ -174,9 +174,10 @@ ol.featureloader.xhr = function(url, format) {
   return ol.featureloader.loadFeaturesXhr(url, format,
       /**
        * @param {Array.<ol.Feature>} features The loaded features.
+       * @param {ol.proj.Projection} dataProjectin Data projection.
```
tschaub: Should be `dataProjection`.  I've just confirmed that the `valid-jsdoc` rule catches this for a function declaration.  I'll dig a bit more to see why it isn't caught here.

tschaub: See eslint/eslint#4964 for the upstream issue.

tschaub:
This change looks good to me @ahocevar.  I didn't test manually, but assuming our tests cover it, this looks good to merge.

PR #4661, opened 2016-01-12
Fixes #4654 Developer Documentation Updated

maveric1977:
Metalsmith v2.0 and above uses generators which are not supported out of the box by node.js 0.12.x and below.

PR #4679, opened 2016-01-15
Add a source option for the full screen control

### LINKED ISSUE #4650: Allow definition of a target for FullScreen control
The ol.control.FullScreen gets always the map.target as it's FullScreen target.
I would suggest to extend the olx.control.FullScreenOptions with a new fullScreenTarget parameter and handle the FullScreen for this dom element when it is provided.

fullscreentarget

I have already developed this functionality for one of our intern projects as a custom ol control extending the default ol.control.FullScreen, but it may make more sense to include it in the default ol.control.FullScreen.

maveric1977:
Fixes #4650

### DIFF
--- a/examples/full-screen-source.js
+++ b/examples/full-screen-source.js
@@ -1,0 +1,28 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.control');
+goog.require('ol.control.FullScreen');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+
+
+var view = new ol.View({
+  center: [-9101767, 2822912],
+  zoom: 14
+});
+
+var map = new ol.Map({
+  controls: ol.control.defaults().extend([
+    new ol.control.FullScreen({
+      source: 'fullscreen'
+    })
+  ]),
+  layers: [
+    new ol.layer.Tile({
+      source: new ol.source.OSM()
+    })
+  ],
+  renderer: common.getRendererFromQueryString(),
+  target: 'map',
+  view: view
+});
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1054,7 +1054,8 @@ olx.control.DefaultsOptions.prototype.zoomOptions;
  *     labelActive: (string|Node|undefined),
  *     tipLabel: (string|undefined),
  *     keys: (boolean|undefined),
- *     target: (Element|undefined)}}
+ *     target: (Element|undefined),
+ *     source: (Element|string|undefined)}}
  * @api
  */
 olx.control.FullScreenOptions;
@@ -1110,6 +1111,12 @@ olx.control.FullScreenOptions.prototype.keys;
  */
 olx.control.FullScreenOptions.prototype.target;
 
+/**
+ * The element to be displayed fullscreen. When not provided, the element containing the map viewport will be displayed fullscreen.
+ * @type {Element|string|undefined}
+ * @api
+ */
+olx.control.FullScreenOptions.prototype.source;
 
 /**
  * @typedef {{className: (string|undefined),
@@ -4245,7 +4252,6 @@ olx.source.VectorTileOptions;
 
 
 /**
-/**
  * Attributions.
  * @type {Array.<ol.Attribution>|undefined}
  * @api
@@ -6317,10 +6323,10 @@ olx.style.RegularShapeOptions.prototype.radius;
 
 
 /**
-* Inner radius of a star.
-* @type {number|undefined}
-* @api
-*/
+ * Inner radius of a star.
+ * @type {number|undefined}
+ * @api
+ */
 olx.style.RegularShapeOptions.prototype.radius1;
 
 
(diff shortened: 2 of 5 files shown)


externs/olx.js
```diff
@@ -1110,6 +1111,12 @@ olx.control.FullScreenOptions.prototype.keys;
  */
 olx.control.FullScreenOptions.prototype.target;
 
+/**
+ * Target Element for FullScreen.
+ * @type {Element|string|undefined}
+ * @api
+ */
+olx.control.FullScreenOptions.prototype.fullScreenTarget;
```
tschaub: We usually use "target" for the destination.  For the control itself, the "target" is where the "controlling element" gets rendered (for example, a button to activate the control).

In this case, you want to provide an option for overriding the source element (not the destination).  I'd be tempted to call this option `source`.  In addition, it needs better documentation.  You should say that by default, the map is the source.  If the user would like a different element to be rendered in full-screen mode, they can provide a custom source.

maveric1977: @tschaub I see what you mean. The changes come right away.

examples/full-screen-source.js
```diff
@@ -0,0 +1,28 @@
...
+var view = new ol.View({
+  center: [-9101767, 2822912],
+  zoom: 14
+});
+
+var map = new ol.Map({
+  controls: ol.control.defaults().extend([
+    new ol.control.FullScreen({
+      fullScreenSource: 'fullscreen'
+    })
```
tschaub: Curious if anybody else has opinions on this.  I think `source` would make a sensible property name (instead of `fullScreenSource`).

bartvde: Agree

src/ol/control/fullscreencontrol.js
```diff
@@ -111,7 +121,8 @@ ol.control.FullScreen.prototype.handleFullScreen_ = function() {
   if (goog.dom.fullscreen.isFullScreen()) {
     goog.dom.fullscreen.exitFullScreen();
   } else {
-    var element = map.getTargetElement();
+    var element = this.fullScreenSource_ !== undefined ?
+      goog.dom.getElement(this.fullScreenSource_) : map.getTargetElement();
```
tschaub: A truthy check should suffice here.  And we typically do two indents for multiline ternary expressions (I'll see if I can configure the linter to enforce this).

So, this could look like:
```
var element = this.fullScreenSource_ ?
    goog.dom.getElement(this.fullScreenSource_) : map.getTargetElement();
```

externs/olx.js
```diff
@@ -1110,6 +1111,12 @@ olx.control.FullScreenOptions.prototype.keys;
  */
 olx.control.FullScreenOptions.prototype.target;
 
+/**
+ * Source element for FullScreen. When not provided the `map.target` DOM element will be used as FullScreen element.
```
tschaub: The spec spells this `Fullscreen`.  The document property is spelled `fullscreenElement`.  In any case, I don't think this should be "element for FullScreen."  How about something like this:

> The element to be displayed fullscreen.  When not provided, the element containing the map viewport will be displayed fullscreen.

examples/full-screen-source.js
maveric1977: Will be changed as proposed to `source`.

src/ol/control/fullscreencontrol.js
maveric1977: OK

externs/olx.js
maveric1977: OK

tschaub:
The change looks good.  Thanks for working through this @maveric1977.  Since the merges and intermediate changes are extraneous, we'd prefer to merge in just a single commit with your relevant change.  Can you please squash these commits into one and rebase on `master`?

In the future, you can maintain your feature branches by rebasing periodically and force pushing.

```
git push --force your-remote your-feature-branch
```

The force push is acceptable in this case, since you are working alone on this feature branch.  If it were a branch with more collaborators, you'd want to coordinate before any force pushes.

Let us know if you could use assistance [squashing](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) these commits.

maveric1977:
@tschaub it's a pleasure to be able to contribute :)
all changes are squashed in one commit now.

maveric1977:
Dear @tschaub , could you please merge this pull request?

maveric1977:
The build failed due to timeout

marcjansen:
I have just restarted the build.

maveric1977:
@marcjansen Thank you very much :)

marcjansen:
Hey @maveric-1977, sorry to bother again: do we have a CLA from you? I have a shaky internet connection and can't check myself.

If I know we have a CLA, I'll merge this.

maveric1977:
Yes, I filled it a month ago prior to my Pull Request #4661

marcjansen:
Thanks a lot for the continued work on this feature!

maveric1977:
Thank you @marcjansen .

PR #4688, opened 2016-01-19
Give map a focus-outline in accessible-example

marcjansen:
This makes it easier to see if the map-element currently has the focus:

outline


fredj:
+1

marcjansen:
Thanks for the review.

PR #4693, opened 2016-01-20
Release v3.13.0

fredj:
This updates the changelog and package version for 3.13.0.

Please tell me if something missing in the changelog

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,10 +1,12 @@
 ## Upgrade notes
 
+### v3.14.0
+
 ### v3.13.0
 
 #### `proj4js` integration
 
-Before this release, OpenLayers depended on the global proj4 namespace. When using for instance browserify you might not want to depend on the global `proj4` namespace. You can use the `ol.proj.setProj4` function to set the proj4 function object. For example in a browserify ES6 environment:
+Before this release, OpenLayers depended on the global proj4 namespace. When using a module loader like Browserify, you might not want to depend on the global `proj4` namespace. You can use the `ol.proj.setProj4` function to set the proj4 function object. For example in a browserify ES6 environment:
 
 ```js
 import ol from 'openlayers';
@@ -14,7 +16,7 @@ ol.proj.setProj4(proj4);
 
 #### `ol.source.TileJSON` changes
 
-The `ol.source.TileJSON` now uses `XmlHttpRequest` to load the TileJSON instead of JSONP with callback.
+The `ol.source.TileJSON` now uses `XMLHttpRequest` to load the TileJSON instead of JSONP with callback.
 When using server without proper CORS support, `jsonp: true` option can be passed to the constructor to get the same behavior as before:
 ```js
 new ol.source.TileJSON({
@@ -22,7 +24,7 @@ new ol.source.TileJSON({
   jsonp: true
 })
 ```
-Also for Mapbox v3, make sure you use urls ending with `.json` (which are able to handle both `XmlHttpRequest` and JSONP) instead of `.jsonp`.
+Also for Mapbox v3, make sure you use urls ending with `.json` (which are able to handle both `XMLHttpRequest` and JSONP) instead of `.jsonp`.
 
 ### v3.12.0
 
--- a/changelog/v3.13.0.md
+++ b/changelog/v3.13.0.md
@@ -1,0 +1,109 @@
+# v3.13.0
+
+## Summary
+
+The v3.13.0 release includes features and fixes from 68 pull requests since the v3.12.1 release.  New features and improvements include:
+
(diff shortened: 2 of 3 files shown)


changelog/upgrade-notes.md
```diff
@@ -1,5 +1,7 @@
 ## Upgrade notes
 
+### v3.14.0
+
```
tschaub: I think the `v3.14.0` heading should be removed.

changelog/v3.13.0.md
```diff
@@ -0,0 +1,108 @@
+# v3.13.0
+
+## Summary
+
+The v3.13.0 release includes features and fixes from 67 pull requests since the v3.12.1 release.  New features and improvements include:
+
+* Improved tiles rendering for the canvas renderer
+* Improved MapQuest rendering
+* Add color option to ol.style.Icon
+* Load TileJSON sources via XmlHttpRequest by default
```
tschaub: Super minor, but XMLHttpRequest is the spelling.

changelog/v3.13.0.md
```diff
@@ -0,0 +1,108 @@
...
+* Add new ol.geom.LineString#getCoordinateAt function
+* Simplify meters per unit handling
+* Use ESLint as a replacement for gjslint.py and jshint
+
+## Upgrade notes
+
+#### `proj4js` integration
+
+Before this release, OpenLayers depended on the global proj4 namespace. When using for instance browserify you might not want to depend on the global `proj4` namespace. You can use the `ol.proj.setProj4` function to set the proj4 function object. For example in a browserify ES6 environment:
```
tschaub: The second sentence could say "When using a module loader like Browserify, you might ..."

changelog/v3.13.0.md
```diff
@@ -0,0 +1,108 @@
...
+
+```js
+import ol from 'openlayers';
+import proj4 from 'proj4';
+ol.proj.setProj4(proj4);
+```
+
+#### `ol.source.TileJSON` changes
+
+The `ol.source.TileJSON` now uses `XmlHttpRequest` to load the TileJSON instead of JSONP with callback.
```
tschaub: Again, `XMLHttpRequest`.

changelog/v3.13.0.md
```diff
@@ -0,0 +1,108 @@
...
+When using server without proper CORS support, `jsonp: true` option can be passed to the constructor to get the same behavior as before:
+```js
+new ol.source.TileJSON({
+  url: 'http://serverwithoutcors.com/tilejson.json',
+  jsonp: true
+})
+```
+Also for Mapbox v3, make sure you use urls ending with `.json` (which are able to handle both `XmlHttpRequest` and JSONP) instead of `.jsonp`.
```
tschaub: One more `XMLHttpRequest`.

changelog/upgrade-notes.md
fredj: I've followed the [Release Procedure](https://github.com/openlayers/ol3/wiki/Release-Procedure) but I agree that it's not necessary

fredj:
if everyone agrees, I will merge this PR and continue the release procedure

bartvde:
+1

changelog/upgrade-notes.md
tschaub: Maybe it is not clear enough, but I think [that documentation](https://github.com/openlayers/ol3/wiki/Release-Procedure#create-a-changelog) is about creating a section for the current release.

> Now edit the changelog/upgrade-notes.md to add a new section at head of the file, ### v3.1.0 for example.

Since that doc is about creating a `v3.1.0` release, that just means to create a heading in the upgrade notes for the current release.

bartvde: Right, but the last few releases I think that header was already there when doing the release. At least when I did 3.12 it was already there.

tschaub:
Thanks for the release work @fredj!

changelog/upgrade-notes.md
tschaub: Ok.  The important part is that the upgrade notes have a header for the current release - so future changes get added above instead of mixed into the previous notes.  I changed the [release procedure docs](https://github.com/openlayers/ol3/wiki/Release-Procedure/_compare/01d8f2998bdc39af9e1e1dbc43855ca1985a602e) to try to clarify this.

If someone adds a header between releases and it looks like `### Next Release`, then you need to edit it to read `### v3.1.0` (assuming you are releasing `v3.1.0`).

If it already reads `### v3.1.0`, you don't need to do anything.  However, guessing the next release version number is a mistake since a patch release may come before the next minor release.  So people should *not* try to add a header between releases that includes the next release number.

Hope this makes sense.

bartvde: Makes sense @tschaub thanks for clarifying

PR #4694, opened 2016-01-20
Fix eslint errors in wms-time example

fredj:


bartvde:
LGTM sorry about this @fredj I didn't think about the eslint changes

fredj:
@bartvde no problem !

PR #4697, opened 2016-01-20
Bind tileUrlFunction to the source

gberaudo:
In order to have the function returned by ol.source.UrlTile#getTileUrlFunction()
to return a url even when executed outside the context of the source, then it
should be bound to the source.

This closes https://github.com/openlayers/ol3/issues/4695.

### DIFF
--- a/src/ol/source/urltilesource.js
+++ b/src/ol/source/urltilesource.js
@@ -63,8 +63,9 @@ ol.source.UrlTile = function(options) {
    * @protected
    * @type {ol.TileUrlFunctionType}
    */
-  this.tileUrlFunction =
-      this.fixedTileUrlFunction || ol.TileUrlFunction.nullTileUrlFunction;
+  this.tileUrlFunction = this.fixedTileUrlFunction ?
+      this.fixedTileUrlFunction.bind(this) :
+      ol.TileUrlFunction.nullTileUrlFunction;
 
   /**
    * @protected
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks @gberaudo. I'll merge when Travis is happy.

ahocevar:
Note @gberaudo: We usually do not reference tickets in commit comments.

gberaudo:
Thanks for the review and the tip @ahocevar.

tschaub:
> We usually do not reference tickets in commit comments.

And the justification for this is that it may take many commits to "fix" an issue.  Referencing an issue in a merge commit could make sense (especially if we wanted traceability without GitHub).  But for now we just reference tickets in pull request descriptions (where merging can take an action on the referenced ticket).

tschaub:
I just restarted [the build](https://travis-ci.org/openlayers/ol3/builds/103632283) to see if the previous failure with the [image filter example](http://openlayers.org/en/master/examples/image-filter.html) was ephemeral.  If it persists, we may have a new PhantomJS issue to deal with.

ahocevar:
We have a new PhantomJS issue it seems. I don't have time to investigate right now. If anyone has, please add a note here. Otherwise I'll look into it later this week.

gberaudo:
Thanks @tschaub for the explanations.
I have removed the reference to github in the commit message and forced a Travis rebuild.

PR #4703, opened 2016-01-21
Source options fixes

gberaudo:

PR #4705, opened 2016-01-21
Remove remaining unnecessary ol.source.State casts

gberaudo:
Forgotten in #4703.


fredj:
Thanks

gberaudo:
Thanks for the quick review.

PR #4680, opened 2016-01-17
Remove goog array

nicholas-l:
Remove the use of goog.array functions for ES5/ES5.1

PR #4709, opened 2016-01-22
[WIP] Tests for ol.array.* methods

marcjansen:
Work in progress, depends upon #4680.

Notes: Based on the tests for [`goog.array.*`](https://github.com/google/closure-library/blob/master/closure/goog/array/array_test.js)

Issues so far:

* `binarySearch` fails the following test:

```js
    describe('finding first index when multiple candidates', function() {
      it('should find the index of the first 0', function() {
        // TODO here our implementation is different from the goog-one
        //      we need to ensure this is intended.
        expect(ol.array.binarySearch([0, 0, 1], 0)).to.be(0);
      });
    });
```
* `extend` fails a lot of tests as we curently do not support mutliple arguments. Do we want to?
* `equals` fails one test, because `ol.array.equals([1], [1, 2])` evaluates to `true` :fearful: 
* `equals` fails a lot of tests as we currently don't support a custom comparison function

the failing tests are included (travis is very likely to fail).


nicholas-l:
> binarySearch fails the following test:

Implementation assumes that each element is unique due to line 36 in array.js:
```
/* Key found. */
    else
      return mid;
```
I am not sure how we would change it, maybe do a check on either side?
> extend fails a lot of tests as we curently do not support mutliple arguments. Do we want to?

I assumed not since we do not have any uses for it at the moment and we don't need to use ```arguments```. 
> equals fails one test, because ol.array.equals([1], [1, 2]) evaluates to true :fearful:

Oops realised the problem, needs a length check:
```
if (arr1.length !== arr2.length) {
  return false;
}
```
> equals fails a lot of tests as we currently don't support a custom comparison function

I think I removed it in the review by @ahocevar when he pointed out we don't use it. We can add it back in if we need it.

ahocevar:
`ol.array.binarySearch` may be called with an array of none unique elements in `ol.geom.flat.interpolate.lineString`. So it will probably be good to add a test there that performs several interpolations on a line which has two points with the same coordinate. If that works, then it should be documented that our implementation may return either of non-unique values in a non deterministic manner. If that test fails, then the implementation should support non-unique values properly.

`ol.array.equals` should definitely `expect(ol.array.equals([1], [1, 2])).to.be(false)`. But it does not need to support custom comparison functions for now, so these tests can be removed.

`ol.array.extend` does not need to support multiple arguments.

nicholas-l:
for binary search we could add a found var and change it to: 
```
...
/* Key found. */
    else {
      found = true;
      high = mid;
    }
  }

  return found ? mid : ~low;
```

ahocevar:
I think that won't work. What if you hit the 2nd occurrence when found?

As I said, it might be no problem when the binary search has this behaviour. We just need to test if line interpolation is still stable, even when a line has multiple consecutive vertices on the same coordinate.

nicholas-l:
I was modelling it off of the goog binarysearch but they don't use the second (cmp < 0) check instead use !compareResult. Yes @ahocevar  you were right, doesn't work, however the below does but it is very similar to the goog version.

```
ol.array.binarySearch = function(haystack, needle, opt_comparator) {
  var cmp;
  var comparator = opt_comparator || ol.array.numberSafeCompareFunction;
  var low = 0;
  var high = haystack.length;
  var found = false;
  var mid = -1;

  while (low < high) {
    /* Note that "(low + high) >>> 1" may overflow, and results in a typecast
     * to double (which gives the wrong results). */
    mid = low + (high - low >> 1);
    cmp = +comparator(haystack[mid], needle);

    /* Too low. */
    if (cmp < 0.0)
      low  = mid + 1;

    /* Key found. */
    else {
      found = !cmp;
      high = mid;
    }
  }

  /* Return key if found otherwise return insert location. */
  return found ? low : ~low;
}
```

marcjansen:
@ahocevar & @nicholas-l: I will not have time to continue working on this.

I think the easiest would be to cherry pick the following commits:

* Sort tested methods alphabetically 4a5c20c
* Check length of arrays when checking equality c421ca4
* Add tests for ol.array.binarySearch bf078bb
* Add tests for ol.array.equals 2f5a4af
* Add tests for ol.array.extend 279db39
* Add tests stubs 1b54850

I have changed the tests @ahocevar has suggested, and only kept the one failing test which @nicholas-l  may want to address.

nicholas-l:
Cherry picked the above commits into #4680 so this pull request can be closed.

PR #4711, opened 2016-01-24
Removal of goog.events.*

ahocevar:
The goal of this pull request is to completely remove the dependency on `goog.events.*`. Tasks required to achieve this goal:

- [x] Removal of all library use of `goog.events.listen`, `goog.events.unlisten`, `goog.events.Event` and `goog.events.EventType`.
- [x] Removal of `goog.events.KeyHandler`, `goog.events.MouseWheelEvent` and `goog.events.MouseWheelHandler`.
- [x] Removal of `goog.fs.FileReader` and `goog.events.FileDropHandler`.
- [x] Removal of `goog.fx.Dragger`.
- [x] Removal of `goog.dom.ViewportSizeMonitor`.
- [x] Removal of `goog.net.XhrIo`, `goog.net.Jsonp` and `goog.async.AnimationDelay`.
- [x] Additional tests and documentation for the new components that replace removed ones.


examples/tilejson.js
```diff
@@ -8,7 +8,8 @@ var map = new ol.Map({
   layers: [
     new ol.layer.Tile({
       source: new ol.source.TileJSON({
-        url: 'http://api.tiles.mapbox.com/v3/mapbox.geography-class.json',
+        jsonp: true,
+        url: 'http://api.tiles.mapbox.com/v3/mapbox.geography-class.jsonp',
```
tschaub: Do we have to revert this change from #4625?

tschaub:
This is so cool.  Thanks for the huge effort on this @ahocevar.  I look forward to digging in more later.

examples/tilejson.js
ahocevar: Oops, thanks for spotting this, it's an uniintentional change. I had just changed it for testing.

ahocevar:
This is ready for review. Or as my brain would say right now:
```js
expect(this).to.be.done();
```
:smile: 

For those interested in metrics:

This branch:
```
info ol Compiling 326 sources
uncompressed: 488202 bytes
  compressed: 143497 bytes
```

master:
```
info ol Compiling 365 sources
uncompressed: 515066 bytes
  compressed: 154113 bytes
```

src/ol/events.js
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * @param {ol.events.EventTargetLike} target Target.
+ * @param {ol.events.EventType|string} type Type.
+ * @return {Array.<ol.events.ListenerObjType>|undefined} Listeners.
+ */
+ol.events.getListeners = function(target, type) {
+  var listenerMap = target[ol.events.LISTENER_MAP_PROP_];
+  if (listenerMap) {
+    return target[ol.events.LISTENER_MAP_PROP_][type];
```
marcjansen: You can reuse `listenerMap` here, right?

src/ol/control/control.js
```diff
@@ -108,7 +108,7 @@ ol.control.Control.prototype.setMap = function(map) {
     goog.dom.removeNode(this.element);
   }
   if (this.listenerKeys.length > 0) {
-    this.listenerKeys.forEach(goog.events.unlistenByKey);
+    this.listenerKeys.forEach(/** @type {Function} */ (ol.events.unlistenByKey));
```
marcjansen: I am surprised that we need the cast here.

ahocevar: I was surprised as well.

src/ol/events.js
```diff
@@ -0,0 +1,347 @@
...
+ol.events.findListener_ = function(
+    listenerArray, listener, useCapture, opt_this, opt_remove) {
+  var listenerObj;
+  for (var i = 0, ii = listenerArray.length; i < ii; ++i) {
+    listenerObj = listenerArray[i];
+    if (listenerObj.listener === listener &&
+        listenerObj.useCapture == useCapture &&
+        listenerObj.bindTo === opt_this) {
+      if (opt_remove) {
+        listenerArray.splice(i, 1);
```
marcjansen: Maybe add a `TODO` that we may want to use `ol.array.remove` here, once we have it?

src/ol/events.js
```diff
@@ -0,0 +1,347 @@
...
+ *     {@link ol.events.EventTarget}, `true` simply means that the listener will
+ *     be called before already registered listeners. Default is false.
+ * @param {Object=} opt_this Object referenced by the `this` keyword in the
+ *     listener. Default is the `target`.
+ * @param {boolean=} opt_once If true, add the listener as one-off listener.
+ * @return {ol.events.Key} Unique key for the listener.
+ */
+ol.events.listen = function(
+    target, type, listener, opt_useCapture, opt_this, opt_once) {
+  if (Array.isArray(type)) {
```
marcjansen: Is this supported everywhere? IE 9?

src/ol/events.js
```diff
@@ -0,0 +1,347 @@
...
+    var keys = [];
+    type.forEach(function(t) {
```
marcjansen: Same question here. I am currently not sure if `Array.prototype.forEach` can be assumed...

marcjansen: Seems to be ok to use: http://caniuse.com/#search=forEach

src/ol/events/eventtarget.js
```diff
@@ -0,0 +1,137 @@
...
+ * @return {boolean|undefined} `false` if anyone called preventDefault on the
+ *     event object or if any of the listeners returned false.
+ */
+ol.events.EventTarget.prototype.dispatchEvent = function(event) {
+  var evt = goog.isString(event) ? new ol.events.Event(event) : event;
+  var type = evt.type;
+  evt.target = this;
+  var listeners = this.listeners_[type];
+  if (listeners) {
+    for (var i = listeners.length - 1; i >= 0; --i) {
```
marcjansen: Do I understand this correct, that if I add two listeners, the latter one will be called first?

src/ol/events.js
ahocevar: That would be unefficient, because we have the index here already, which `ol.array.remove` would calculate in another loop.

src/ol/events/condition.js
```diff
@@ -75,6 +75,22 @@ ol.events.condition.click = function(mapBrowserEvent) {
...
+ * By definition, this includes left-click on windows/linux, and left-click
+ * without the ctrl key on Macs.
+ *
+ * @param {ol.MapBrowserEvent} mapBrowserEvent Map browser event.
+ * @return {boolean} The result.
+ */
+ol.events.condition.mouseActionButton = function(mapBrowserEvent) {
+  var browserEvent = mapBrowserEvent.originalEvent;
+  return browserEvent.button == 0 &&
+      !(goog.userAgent.WEBKIT && goog.userAgent.MAC && browserEvent.ctrlKey);
```
marcjansen: Can wemake use of `ol.has.XXX`

src/ol/events.js
ahocevar: It's ECMAScript 5, which is not fully supported by IE9, but there are polyfills. Same story as with `Function.prototype.bind()`, which we use already.

src/ol/events.js
marcjansen: True, but there is also a performance test hinting that splice may be slow. This is not super important though

src/ol/events/eventtarget.js
ahocevar: No, because listeners are added with `unshift()`, not `push()`. Looping in reverse order allows listeners to remove themselves or other listeners without disturbing the loop.

src/ol/events.js
ahocevar: Definitely, thanks for spotting this!

test/index.html
```diff
@@ -32,22 +32,6 @@
...
-     * In addition, calling goog.events.listen on the global object (as done
-     * in deviceorientation.js) creates a second leak by setting
-     * goog.events.LISTENER_MAP_PROP_ on the global object.
-     *
-     * We preemptively set both of these properties so Mocha can compare the
-     * global before and after tests.  The call to goog.events.listen also
-     * calls goog.getUid.
-     */
-    goog.events.listen(this, 'test', function() {});
-
```
marcjansen: Very happy to see this gone!

src/ol/events/eventtarget.js
marcjansen: Makes sense, and other api docs hinted that to me. Thanks for the clarification.

src/ol/events/condition.js
ahocevar: For MAC yes, for WEBKIT no (I think).

test/spec/ol/control/zoomslidercontrol.test.js
```diff
@@ -85,12 +86,104 @@ describe('ol.control.ZoomSlider', function() {
...
+      event.clientY = 4 * control.limits_.height / 8;
+      dragger.dispatchEvent(event);
+      event.type = ol.pointer.EventType.POINTERUP;
+      dragger.dispatchEvent(event);
+      expect(control.currentResolution_).to.be(1);
+      expect(control.dragListenerKeys_).to.be(null);
+      expect(control.dragging_).to.be(false);
+    });
+  });
+
```
marcjansen: Very nice tests here! 👏

src/ol/events.js
ahocevar: True, but `ol.array.remove` uses `Array.prototype.splice` as well: https://github.com/openlayers/ol3/pull/4680/files#diff-610d4c9e0938c945058e31c6e5df3a7dR208

src/ol/events.js
ahocevar: Actually it is supported in IE9: http://kangax.github.io/compat-table/es5/#Array.isArray

src/ol/events.js
marcjansen: Yeah, just saw as well. Just ignore the minor comment

src/ol/events.js
marcjansen: Even better ☺

marcjansen:
Awesome work, Andreas. Up until now I have only read through the diff (that probably explains some stupid comments)  but still I am very impressed. I will have another look for sure, but this really is amazing.

Short note: the `ol.net.jsonp` method isn't really related, is it? Are there tests for the method?

ahocevar:
@marcjansen You are right, `ol.net.jsonp` tuned out to not use anything of `goog.events.*` in the end, but it was a simple replacement and easy to test: https://github.com/ahocevar/ol3/blob/6461590332a63cfa5bf5f3cd0d9ac334ba450f40/test/spec/ol/net.test.js.

externs/closure-compiler.js
```diff
@@ -15,3 +15,13 @@
  * @see http://www.w3.org/TR/pointerevents/#the-touch-action-css-property
  */
 CSSProperties.prototype.touchAction;
+
+/**
+ * @type {number}
+ */
+WheelEvent.DOM_DELTA_LINE;
```
fredj: FYI I've added these to the closure-compiler externs (see https://github.com/google/closure-compiler/pull/1478)

ahocevar: Awesome @fredj, thanks! I was surprised to see that the compiler knows about it.

ahocevar: Ah wait, you mean I can remove this? Will do.

fredj: we need to wait for the next closure-compiler release

marcjansen:
I have tested this locally and I still very happy with this changes! Great work, @ahocevar. Some further notes:

* tests all pass for me when run through phantomjs, Chrome (48, Linux) and Firefox (43, Linux)
* In IE 11
  * F12, Document mode: 'Edge (default)', the following tests fail:
    * `ol.control.ZoomSlider` => `Pointer event handling` (2 tests fail, the `horizontal` ones: *expected 0.9533213309466342 to equal 1*)
      * The zoomslider example is working fine, IMO
    * `ol.source.TileImage` => `#getTile` (4 tests fail, but with a timeout, also when run via grep)
  * F12, Document mode: '10', the following test fails:
    * `net.test.js` throws an exception "Object doesn't support property or method 'olc_5628'", test is `accepts a custom callback param`
    * `ol.Map` => `#getEventPixel` (2 tests fail, "expected 3306 to sort of equal 190")
  * F12, Document mode: '9', the following test fails:
    * `mocha.js` (!) throws an exception (around the graticule tests)
    * The testsuite then stops with 72 successes and 1 failure (3%)

I don't know which of the above are introduced by this PR, but I doubt that all errors from above stem from this PR. I also don't know how reliable the results with different modes of IE 11 are.

All the examples that I have tested worked fine in the aforementioned browsers (only the edge-mode of IE11 tested)

Can you say something wrt the failures, @ahovevar?

src/ol/control/control.js
```diff
@@ -60,7 +60,7 @@ ol.control.Control = function(options) {
 
   /**
    * @protected
-   * @type {!Array.<?number>}
+   * @type {!Array.<ol.events.ListenerObjType>}
```
fredj: Should be `{!Array.<ol.events.Key>}` (?)

ahocevar:
@marcjansen The first is a test-only css issue. For the others, can you please try if you also get them with master before I investigate? Thanks.

Makefile
```diff
@@ -181,7 +181,7 @@ build/timestamps/check-%-timestamp: $(BUILD_HOSTED)/examples/%.html \
                                     $(BUILD_HOSTED)/build/ol.js \
                                     $(BUILD_HOSTED)/css/ol.css
 	@mkdir -p $(@D)
-	./node_modules/.bin/phantomjs --ssl-protocol=any --ignore-ssl-errors=true bin/check-example.js $<
+	./node_modules/.bin/phantomjs --web-security=no --ssl-protocol=any --ignore-ssl-errors=true bin/check-example.js $<
```
fredj: `--local-to-remote-url-access=true` could be used instead of `--web-security=no`

marcjansen:
On `master`:

* In IE 11 (Edge-mode)
  * `ol.source.TileImage` => `#getTile` (4 tests fail, but with a timeout, also when run via grep)
* In IE 11 (document-mode  '10')
  * `ol.source.TileImage` => `#getTile` (4 tests fail, but with a timeout, also when run via grep)
  * `ol.Map` => `#getEventPixel` (2 tests fail, "expected 3306 to sort of equal 190")
* In IE 11 (document-mode  '9')
  * `ol.source.TileImage` => `#getTile` (4 tests fail, but with a timeout, also when run via grep)
  * `ol.Map` => `#getEventPixel` (2 tests fail, "expected 3306 to sort of equal 190")
  * The suite stops at 88% with the same mocha.js exception.

This leaves the failures introduced here as follows:
* `ol.control.ZoomSlider` => Pointer event handling (2 tests fail, the horizontal ones: *expected 0.9533213309466342 to equal 1*)
* net.test.js throws an exception `Object doesn't support property or method 'olc_5628'`

ahocevar:
Very useful information, thanks @marcjansen. I can easily fix the two remaining ones.

Makefile
ahocevar: Thanks for this hint, I was not aware of that option.

ahocevar:
@fredj @marcjansen I addressed all your comments so far.

src/ol/events.js
```diff
@@ -0,0 +1,344 @@
...
+ * @param {boolean=} opt_useCapture Use capture. For listeners on an
+ *     {@link ol.events.EventTarget}, `true` simply means that the listener will
+ *     be called before already registered listeners. Default is false.
+ * @param {Object=} opt_this Object referenced by the `this` keyword in the
+ *     listener. Default is the `target`.
+ * @param {boolean=} opt_once If true, add the listener as one-off listener.
+ * @return {ol.events.Key} Unique key for the listener.
+ */
+ol.events.listen = function(
+    target, type, listener, opt_useCapture, opt_this, opt_once) {
```
tschaub: In all 100+ call sites for `ol.events.listen()`, the `opt_useCapture` argument is `false` or `undefined`.  In all but a handful of these sites, we call with `opt_this`.  It would make for slightly nicer code if `opt_this` were the 4th argument and `opt_useCapture` were last.  And since we never use it, I think we'd be fine removing `opt_useCapture` altogether.

I've pushed two commits to my [`no-use-capture` branch](https://github.com/ahocevar/ol3/compare/no-goog-events...tschaub:no-use-capture).  If you agree about reordering the arguments so `opt_useCapture` is last, you can cherry pick the first.  If you agree that we could get rid of `opt_useCapture` altogether, you can cherry pick both.

ahocevar: Great work @tschaub, thanks! I agree with both changes - very much in fact - and am going to cherry-pick both.

marcjansen:
Let's wait where @tschaub is headed, I'll then have another short look.

src/ol/mapbrowserevent.js
```diff
@@ -37,25 +36,19 @@ ol.MapBrowserEvent = function(type, map, browserEvent, opt_dragging,
   goog.base(this, type, map, opt_frameState);
 
   /**
-   * @const
-   * @type {goog.events.BrowserEvent}
-   */
-  this.browserEvent = browserEvent;
```
tschaub: Glad to see this gone.

For consistency, I think it would be nice if `ol.pointer.PointerEvent` also had an `originalEvent` property instead of `browserEvent`.  I know this goes beyond the changes you are making here, but if you agree, you can cherry-pick from my [`original-events` branch](https://github.com/ahocevar/ol3/compare/no-goog-events...tschaub:original-event) (this is mostly argument/variable renaming for the sake of readability).

src/ol/net.js
```diff
@@ -0,0 +1,33 @@
...
+    delete goog.global[key];
+    if (opt_errback) {
+      opt_errback();
+    }
+  }, 10000);
+  goog.global[key] = function(data) {
+    goog.global.clearTimeout(timer);
+    delete goog.global[key];
+    callback(data);
+  };
```
tschaub: Minor, but both of the above functions could call a `cleanup` function.
```js
function cleanup() {
  delete goog.global[key];
  var parent = script.parentNode;
  if (parent) {
    parent.removeChild(script);
  }
}
```

src/ol/source/tilejsonsource.js
```diff
@@ -44,21 +40,20 @@ ol.source.TileJSON = function(options) {
...
         this.handleTileJSONError.bind(this));
   } else {
-    var xhr = new goog.net.XhrIo(new goog.net.CorsXmlHttpFactory());
-    goog.events.listen(xhr, goog.net.EventType.COMPLETE, function(e) {
-      if (xhr.isSuccess()) {
-        var response = /** @type {TileJSON} */(xhr.getResponseJson());
+    var xhr = new XMLHttpRequest();
+    xhr.open('GET', options.url, true);
+    xhr.onload = function(e) {
+      if (xhr.status < 400) {
```
tschaub: For network issues or CORS problems, `status` will be `0`, and we'd want the error handler to be called.  I'd change this to `xhr.status >= 200 && xhr.status < 300` (we aren't going to have the expected response text if a redirect isn't followed).

src/ol/deviceorientation.js
```diff
@@ -83,13 +83,13 @@ ol.DeviceOrientation = function(opt_options) {
 
   /**
    * @private
-   * @type {goog.events.Key}
+   * @type {ol.events.Key|null}
```
tschaub: I'm guessing `|null` isn't needed here (but haven't tested).

tschaub:
This is really great work @ahocevar.  Thanks for maintaining the energy required to get it done.  I've got a proposal for one more change that I hope sounds ok.

Instead of having `ol.events.Key` be `ol.events.ListenerObjType|Array<ol.events.ListenerObjType>`, I think things would be simpler (and less error prone) if it were just `ol.events.ListenerObjType`.

In addition (or really part of the same issue), I think `ol.events.listen()` (and friends) should only accept a single `type`.  As it is currently, the `type` argument can be `ol.events.EventType|string|Array.<(ol.events.EventType|string)>`.  Ideally this would be just a single enum value.

To give users a more convenient API, we can have `observable.on()` accept an array of strings, but I think internally we should use the stronger typed `ol.events.listen()`.

I've pushed up a [`key-type` branch](https://github.com/ahocevar/ol3/compare/no-goog-events...tschaub:key-type) with these changes.  Please cherry pick if you agree.

ahocevar:
This proposal sounds great @tschaub. Cherry-picking and pushing right now.

tschaub:
One "final" [change](https://github.com/tschaub/ol3/commit/206576f5a15efb1bef57945e11e1ea80393afba6) in the same branch is to replace `string|ol.events.EventType` with `string`.  The two are equivalent since `ol.events.EventType` is a string enum.  We could get explicit and list all of the possible types (e.g. `ol.events.EventType|ol.pointer.EventType|ol.source.VectorEvent|ol.render.EventType|etc.`) but that would get pretty verbose given all the places we use event types.

ahocevar:
Awesome, thanks @tschaub! Just cherry-picked it.

I also made one final [commit](https://github.com/openlayers/ol3/commit/6f8f19e4c487306b4ce83b6ac49e505f3aa601f9), which - I think - improves readability a bit. You might want to take a look at that too.

tschaub:
This is looking great @ahocevar.  I think it is merge time.

It sounds like there may be some outstanding issues with tests on IE (https://github.com/openlayers/ol3/pull/4711#issuecomment-178549279).  @marcjansen can you ticket whatever remains?

ahocevar:
Thanks @tschaub, @marcjansen and @fredj for the valuable help with this!

marcjansen:
> It sounds like there may be some outstanding issues with tests on IE (#4711 (comment)). @marcjansen can you ticket whatever remains?

I did already: #4768. Or were new ones introduced?

tschaub:
> (The above is ticketed as #4768)
> 
> This leaves the failures introduced here as follows:
> 
> ol.control.ZoomSlider => Pointer event handling (2 tests fail, the horizontal ones: expected 0.9533213309466342 to equal 1)
> net.test.js throws an exception Object doesn't support property or method 'olc_5628'

@marcjansen from your [comment above](https://github.com/openlayers/ol3/pull/4711#issuecomment-178549279) it sounded like there were two failures you thought were introduced with the changes in this branch (and not ticketed in #4768).

ahocevar:
@tschaub Those two are fixed. I rebased the fixes into existing commits, so you do not see them.

tschaub:
Thanks @ahocevar.  I was unsure if that was what you meant in https://github.com/openlayers/ol3/pull/4711#issuecomment-178599262.

PR #4718, opened 2016-01-26
Improve raster reprojection behavior when tiles fail to load

### LINKED ISSUE #4682: Raster reprojection do not handle tile load error
It seems that raster reprojection do not try to reuse cached tiles in case of tile load error. Without reprojection zooming in more than tileserver last resolution will reuse cached tiles.
With raster reprojection this will not happen:
jsfiddle: https://jsfiddle.net/1pwndhwe/

petrsloup:
This PR slightly improves behavior of reprojection of tiled raster sources:
When all of the required source tiles fail to load, no reprojection is performed and the tile state is set to `ol.TileState.ERROR` (this hints the renderer to use the previous zoom level and scale it up rather than displaying empty tile).
(If only *some* of the tiles fail to load, the behavior is unchanged and reprojected tile is created from the available tiles.)

Closes #4682.

### DIFF
--- a/src/ol/reproj/tile.js
+++ b/src/ol/reproj/tile.js
@@ -254,21 +254,25 @@ ol.reproj.Tile.prototype.reproject_ = function() {
   }, this);
   this.sourceTiles_.length = 0;
 
-  var z = this.wrappedTileCoord_[0];
-  var size = this.targetTileGrid_.getTileSize(z);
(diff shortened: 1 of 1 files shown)


bartvde:
LGTM please merge

petrsloup:
Thanks @bartvde !

PR #4720, opened 2016-01-26
Improve tileLoadFunction docs

ahocevar:
I found that I could not use a custom `tileLoadFunction` without looking in the sources. I think this pull request makes things more clear.

I also found that we still have a TileVectorLoadFunction type in the docs, which we do not use any more. So I removed it.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3943,7 +3943,12 @@ olx.source.BingMapsOptions.prototype.reprojectionErrorThreshold;
 
 
 /**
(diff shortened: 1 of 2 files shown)


fredj:
`goog.provide('ol.TileVectorLoadFunctionType');` in `src/ol/tileloadfunction.js`can be removed

fredj:
LGTM, thanks

ahocevar:
Ok, done. Waiting for Travis before merging, just to be sure.

PR #4721, opened 2016-01-26
Fix reprojection of raster sources with gutter

### LINKED ISSUE #4681: Raster reprojection with TileWMS source
ol-3.12.1.
I tried to use raster reprojection from EPSG:3301 to EPSG:3857. At first I notice horisontal caps between tilerows at larger zoomlevels, then I tried to use gutter and result was gutter-sized caps over tiles.

jsfiddle: https://jsfiddle.net/aymvra7g/

note: not sure does this wms server accessible outside of Estonia. Screen:
http://sookoll.ee/failid/ol_reproject_bug.png

petrsloup:
This is a possible solution to reprojecting raster sources with gutter (currently only `ol.source.TileWMS`).

The `ol.source.Tile#getGutter` gets additional `projection` parameter which is used to distinguish whether to return 0 (when reprojecting) or the actual gutter value of the source tiles (when not reprojecting). And the gutter handling is added to the reprojection process so only the proper parts of the tiles are used.

I'm not 100 % sure this is the best way to solve this, so any ideas are welcome.

Demo of the problem: https://jsfiddle.net/aymvra7g/ (from #4681)

Closes #4681.

### DIFF
--- a/src/ol/renderer/canvas/canvastilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvastilelayerrenderer.js
@@ -67,7 +67,7 @@ ol.renderer.canvas.TileLayer.prototype.composeFrame = function(
   var source = layer.getSource();
   goog.asserts.assertInstanceof(source, ol.source.Tile,
       'source is an ol.source.Tile');
-  var tileGutter = source.getGutter();
+  var tileGutter = source.getGutter(projection);
   var opaque = source.getOpaque(projection);
 
   var transform = this.getTransform(frameState, 0);
--- a/src/ol/renderer/dom/domtilelayerrenderer.js
+++ b/src/ol/renderer/dom/domtilelayerrenderer.js
@@ -94,7 +94,7 @@ ol.renderer.dom.TileLayer.prototype.prepareFrame = function(frameState, layerSta
       'layer is an instance of ol.layer.Tile');
   var tileSource = tileLayer.getSource();
   var tileGrid = tileSource.getTileGridForProjection(projection);
-  var tileGutter = tileSource.getGutter();
+  var tileGutter = tileSource.getGutter(projection);
   var z = tileGrid.getZForResolution(viewState.resolution);
   var tileResolution = tileGrid.getResolution(z);
   var center = viewState.center;
--- a/src/ol/renderer/webgl/webgltilelayerrenderer.js
+++ b/src/ol/renderer/webgl/webgltilelayerrenderer.js
@@ -169,7 +169,7 @@ ol.renderer.webgl.TileLayer.prototype.prepareFrame = function(frameState, layerS
   var pixelRatio = tilePixelSize[0] /
       ol.size.toSize(tileGrid.getTileSize(z), this.tmpSize_)[0];
   var tilePixelResolution = tileResolution / pixelRatio;
-  var tileGutter = tileSource.getGutter();
+  var tileGutter = tileSource.getGutter(projection);
 
   var center = viewState.center;
   var extent;
--- a/src/ol/reproj/image.js
+++ b/src/ol/reproj/image.js
@@ -163,7 +163,7 @@ ol.reproj.Image.prototype.reproject_ = function() {
         this.targetResolution_, this.targetExtent_, this.triangulation_, [{
           extent: this.sourceImage_.getExtent(),
           image: this.sourceImage_.getImage()
-        }]);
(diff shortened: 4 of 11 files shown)


petrsloup:
Rebased

ahocevar:
I am not convinced that using a gutter would solve the issue reported in #4681. I updated the fiddle using Boundless's demo GeoServer, and I am seeing the same problem: http://jsfiddle.net/aymvra7g/3/. It looks like wrong scaling in a transform (see how the border is interrupted).

ahocevar:
The same problem was also reported here: http://stackoverflow.com/questions/35335067/openlayers-3-white-gaps-between-osm-map-tiles

petrsloup:
Fix for precision issue with reprojection in separate PR: https://github.com/openlayers/ol3/pull/4869.

This fix for missing gutter support is still required (rebase will be needed).

ahocevar:
Nice! Avoiding these canvas transforms should also improve performance. Please merge.

ahocevar:
Sorry, I commented on the wrong pull request. I meant to comment on #4869.

ahocevar:
It doesn't look like this will give the desired results, because the gutter overlaps are not really used in the reprojected output.I think at this point, it would be better to just document that gutter does not work on reprojected views.

petrsloup:
@ahocevar There are two gutter-related issues why the http://jsfiddle.net/aymvra7g/ does not work:
- The `ol.reproj.render` uses whole source images (ignores the gutter value), so the gutter pixels are present twice in the reprojected tiles.
- The `#getGutter` method on the source always returns the gutter value for the source tiles although the reprojected tiles are always created without gutter.

This PR changes both of these: `ol.reproj.render` *ignores* the additional gutter pixels + `#getGutter` takes additional `projection` argument and returns different value for the data projection (actual gutter value) or 0 for other projections (no gutter when reprojecting).

With changes in this PR, the code in http://jsfiddle.net/aymvra7g/ seems to be working as expected for arbitrary gutter value.

ahocevar:
Doesn't #4869 also fix http://jsfiddle.net/aymvra7g/?

My point was that it makes no sense to request images with a gutter from the original source when the gutter is discarded:
```
stitchContext.drawImage(
  src.image,
  gutter, gutter, // discards left/top gutter from the source image
  src.image.width - 2 * gutter, src.image.height - 2 * gutter, // discards right/bottom gutter
  xPos * stitchScale, yPos * stitchScale,
  srcWidth * stitchScale, srcHeight * stitchScale);
});
```
This is why I think it would make more sense to document that reprojecting sources with a gutter is not supported. You could assert that gutter is 0 before reprojecting tiles.

petrsloup:
Unfortunately, #4869 only fixes the thin white line visible in http://sookoll.ee/failid/ol_reproject_bug.png. The rest is specific to the gutter issue.

I see your point, but the gutter is discarded in the renderers as well:
```
renderContext.drawImage(tile.getImage(), tileGutter, tileGutter,
        tilePixelSize[0], tilePixelSize[1],
        ...
```
(from `ol.renderer.canvas.TileLayer#composeFrame`), so the behavior is similar here.

Anyway, the `#getGutter` in currently only implemented in `ol.source.TileWMS` where the `gutter` parameter documentation says:

> The size in pixels of the gutter around image tiles to ignore. By setting this property to a non-zero value, images will be requested that are wider and taller than the tile size by a value of 2 x gutter. Defaults to zero. Using a non-zero value allows artifacts of rendering at tile edges to be ignored. ...

(http://openlayers.org/en/v3.13.1/apidoc/ol.source.TileWMS.html).
So it's ok in my opinion to discard the additional pixels.

I agree that not using gutter is actually more elegant and correct solution :), but since OL3 has support for sources with gutter, we should probably support this during the raster reprojection as well.
(Besides, turning the gutter off is possible for TileWMS, but it might not be possible for some static sources with prerendered tiles.)

ahocevar:
Ah sorry, I had misunderstood how gutter works. Thanks for the explanation.

I'm not a big fan of passing the projection to getGutter, but since we do this already for other methods, it is probably ok.

Please merge.

petrsloup:
Thanks for having a look at this!

I agree passing `projection` to `#getGutter` is not very clear, but I can't think of any better solution.
(It's similar to `#getOpaque`, which also gets `projection` parameter -- see https://github.com/openlayers/ol3/pull/4628.)

Merged.

PR #4725, opened 2016-01-26
Untangle vector tile feature reprojection

ahocevar:
When using an `ol.format.MVT` with `featureClass` set to `ol.Feature`, OpenLayers thinks it has to reproject features, even though they are drawn in the same projection ('EPSG:3857').

This change makes it so reprojection happens later in the renderer, where more information is available than in the feature format. In the future, this change here will also make it easier to transform features to a different projection, even if the data projection is in tile-pixels units.

### DIFF
--- a/src/ol/format/mvtformat.js
+++ b/src/ol/format/mvtformat.js
@@ -41,7 +41,7 @@ ol.format.MVT = function(opt_options) {
    * @type {ol.proj.Projection}
    */
   this.defaultDataProjection = new ol.proj.Projection({
-    code: 'EPSG:3857',
+    code: '',
     units: ol.proj.Units.TILE_PIXELS
   });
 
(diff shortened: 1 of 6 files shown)


src/ol/format/mvtformat.js
```diff
@@ -41,7 +41,7 @@ ol.format.MVT = function(opt_options) {
    * @type {ol.proj.Projection}
    */
   this.defaultDataProjection = new ol.proj.Projection({
-    code: 'EPSG:3857',
+    code: '',
```
bartvde: can you give some background on this change? Does it mean pixel space projections will always have an empty string as the code?

bartvde:
LGTM @ahocevar I put a small question in the diff

src/ol/format/mvtformat.js
ahocevar: Not ultimately, but for now I think it makes more sense to not determine the code. Neither is it possible to change the `defaultDataProjection`, nor can pixel coordinates be reprojected. If we add support for reprojecting pixel coordinates, we can make `defaultDataProjection` configurable, and use 'EPSG:3857' as default.

bartvde: Ok.

PR #4727, opened 2016-01-27
Fix phantomjs

nicholas-l:
Fixes running phantomjs. Couldn't find usage/need of mode=advanced in the examples as it seems to only check for raw, is this an issue?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ build/timestamps/check-%-timestamp: $(BUILD_HOSTED)/examples/%.html \
                                     $(BUILD_HOSTED)/build/ol.js \
                                     $(BUILD_HOSTED)/css/ol.css
 	@mkdir -p $(@D)
-	./node_modules/.bin/phantomjs --ssl-protocol=any --ignore-ssl-errors=true bin/check-example.js $(addsuffix ?mode=advanced, $<)
+	./node_modules/.bin/phantomjs --ssl-protocol=any --ignore-ssl-errors=true bin/check-example.js $<
 	@touch $@
 
 build/timestamps/check-requires-timestamp: $(SRC_JS) $(EXAMPLES_JS) \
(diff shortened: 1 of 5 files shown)


nicholas-l:
Once this is done we can remove Function.bind polyfill :)

package.json
```diff
@@ -44,6 +44,7 @@
     "metalsmith-layouts": "1.4.2",
     "nomnom": "1.8.1",
     "pbf": "1.3.5",
+    "phantomjs": "^2.1.2",
```
ahocevar: This is already in `devDependencies`.

Makefile
```diff
@@ -181,7 +181,7 @@ build/timestamps/check-%-timestamp: $(BUILD_HOSTED)/examples/%.html \
                                     $(BUILD_HOSTED)/build/ol.js \
                                     $(BUILD_HOSTED)/css/ol.css
 	@mkdir -p $(@D)
-	./node_modules/.bin/phantomjs --ssl-protocol=any --ignore-ssl-errors=true bin/check-example.js $(addsuffix ?mode=advanced, $<)
+	./node_modules/.bin/phantomjs --ssl-protocol=any --ignore-ssl-errors=true bin/check-example.js $<
```
ahocevar: Good catch, thanks!

package.json
nicholas-l: thats why it rebased so easily, didn;t look much into it :(

ahocevar:
Also, since you based your branch on openlayers:greenkeeper-phantomjs-2.1.2 instead of master, we do not see if Travis runs through with this.

nicholas-l:
Yes, as it says in the greenkeeper pull request that we can work on that for any fixes? I can make a pull request to master but that will fail because this has a dependency of PhantomJS2

ahocevar:
Ok, so please remove the addition of phantomjs to `dependencies`, and then we merge this and see what Travis says on #4717.

nicholas-l:
Already done :)

ahocevar:
Thanks!

PR #4732, opened 2016-01-27
Fix '@see' link in src/ol/deviceorientation.js

fredj:


ahocevar:
Looks good, please merge.

PR #4733, opened 2016-01-27
Avoid rendering too big and too small images for vector tiles

ahocevar:
We fall back from image to replay rendering in the vector tile layer renderer when the tile is at a resolution lower than its coordinate resolution, which we determine by looking at the tilePixelRatio. But in this calculation, we may allow huge tile sizes when the tile set uses a high tilePixelRatio.

The current behaviour is also buggy, because it does not take the pixelRatio into account. This results in replay rendering on retina displays at native tile resolutions, as can be seen in the osm-vector-tiles examples.

### DIFF
--- a/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
@@ -92,7 +92,6 @@ ol.renderer.canvas.VectorTileLayer.prototype.composeFrame = function(frameState,
   goog.asserts.assertInstanceof(source, ol.source.VectorTile,
       'Source is an ol.source.VectorTile');
   var tilePixelRatio = source.getTilePixelRatio(pixelRatio);
-  var maxScale = tilePixelRatio / pixelRatio;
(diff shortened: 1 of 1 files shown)


tschaub:
> But in this calculation, we may allow huge tile sizes when the tile size uses a high tilePixelRatio.

Does this result in a rendering "bug"?  It would make it easier to understand the proposed changes with a screen shot.

ahocevar:
It does not result in a rendering bug, it only makes the browser very unresponsive. You can see that when you open the http://openlayers.org/en/v3.13.0/examples/mapbox-vector-tiles.html example and scroll-zoom all the way in to a much higher zoom level. This will result in canvas sizes of > 8000 pixels for lower resolution tiles. On a Mac, I get a spinning wheel even when I do that.

tschaub:
Thanks for the additional description @ahocevar.  I haven't been able to reproduce what you're talking about, but I may not be doing the right thing.  Rendering still looks good to me with this change and the code looks sensible.  Please merge.

PR #4741, opened 2016-01-27
Report on installed versions in Travis

tschaub:
Checking to see how our cached `node_modules` might affect things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ before_install:
 before_script:
   - "rm src/ol/renderer/webgl/*shader.js"
   - "sh -e /etc/init.d/xvfb start"
+  - "npm ls"
 
 script: "make ci"
 


fredj:
What about to always display the version with `npm ls --depth=0`?

tschaub:
That could be useful @fredj.

I was just suspicious here because I didn't see the updated Phantom getting installed in the log for [build 105115206](https://travis-ci.org/openlayers/ol3/builds/105115206).  So I was wondering if `npm install` wasn't doing what we want with the cached `node_modules`.  But on digging deeper I see that there was a [previous build](https://travis-ci.org/openlayers/ol3/builds/104777530) for the same branch where the correct version of Phantom was installed, and the cached `node_modules` for this branch was used in the subsequent build.

tschaub:
@fredj I repurposed this to report on all versions based on your suggestion.  This serves as a sanity check at least.

fredj:
+1

PR #4742, opened 2016-01-27
Upgrade to eslint@2.0.0-beta.2

tschaub:
The changes in #4602 worked around two issues in ESLint that are now addressed.

Prior to https://github.com/eslint/eslint/commit/0f469b4, object properties would incorrectly be assigned comment blocks for parent statements.  And prior to https://github.com/eslint/eslint/commit/5cd5429, comment blocks for function expressions in call expressions would not be linted.


fredj:
LGTM

PR #4753, opened 2016-01-28
Export ol.geom.Geometry#containsCoordinate function

fredj:


src/ol/geom/geometry.js
```diff
@@ -150,6 +151,7 @@ ol.geom.Geometry.prototype.computeExtent = goog.abstractMethod;
  * @param {number} x X.
  * @param {number} y Y.
  * @return {boolean} Contains (x, y).
+ * @api
```
tschaub: Are you working with an application where it is a performance hit to use `containsCoordinate`?

src/ol/geom/geometry.js
```diff
@@ -132,6 +132,7 @@ ol.geom.Geometry.prototype.getClosestPoint = function(point, opt_closestPoint) {
 /**
  * @param {ol.Coordinate} coordinate Coordinate.
  * @return {boolean} Contains coordinate.
+ * @api
```
tschaub: I'm in favor of exporting `containsCoordinate`.  But we need to have better docs when we export.  This needs at least a short, one sentence description.  With more detail if necessary.

It would be good to describe our interpretation of "contains" here.  A strict topological interpretation might say that ["polygons do not contain their boundary"](http://www.vividsolutions.com/jts/javadoc/com/vividsolutions/jts/geom/Geometry.html#contains(com.vividsolutions.jts.geom.Geometry)).  So a point that "touches" a polygon but is not in the interior would return false for `poly.containsCoordinate(point)`.

fredj: > I'm in favor of exporting `containsCoordinate`.  But we need to have better docs when we export.  This needs at least a short, one sentence description.  With more detail if necessary.

@gberaudo told me the same thing 5 minutes ago :-)

src/ol/geom/geometry.js
fredj: My application only use `containsCoordinate`

src/ol/geom/geometry.js
fredj: Note: for `ol.geom.Polygon`, this algorithm is used: https://www.ecse.rpi.edu/Homepages/wrf/Research/Short_Notes/pnpoly.html

fredj:
I've updated the function description, don't hesitate to correct bad English ...

src/ol/geom/geometry.js
```diff
@@ -130,8 +130,11 @@ ol.geom.Geometry.prototype.getClosestPoint = function(point, opt_closestPoint) {
 
 
 /**
+ * Returns true if this geometry contains the specified coordinate. If the
+ * coordinate is on the boundary of the geometry, returns false.
```
tschaub: I'm guessing that our implementation doesn't really follow the [traditional definition](http://www.vividsolutions.com/jts/javadoc/com/vividsolutions/jts/geom/Geometry.html#contains(com.vividsolutions.jts.geom.Geometry)) of contains for polygons.  Looking at the code, I'd guess that a point on the boundary of an exterior ring would be considered "contained", while a point on the boundary of an interior ring would not be contained.  A more strict implementation would say that a polygon only contains a point if it is in the interior but not touching a ring.

tschaub:
I brought up the definition of "contains" because I think we want to decide if the current implementation is a bug or intended behavior for `polygon.containsCoordinate(point)`.  A strict contains method would only return true if the point was in the interior of the polygon.

src/ol/geom/geometry.js
tschaub: Right, I'm only questioning the use of the word "contains".  It may be more appropriate to call this "intersects".

fredj:
Can we merge this pull request and revisit the implementation later?

src/ol/geom/geometry.js
```diff
@@ -130,8 +130,11 @@ ol.geom.Geometry.prototype.getClosestPoint = function(point, opt_closestPoint) {
 
 
 /**
+ * Returns true if this geometry includes the specified coordinate. If the
+ * coordinate is on the boundary of the geometry, returns false.
```
tschaub: This would be the common description of "contains".  However, it is not how this method behaves:
```js
var poly = new ol.geom.Polygon([[[0, 0], [1, 0], [1, 1], [0, 1], [0, 0]]]);
poly.containsCoordinate([0, 0]) // true
```

tschaub:
@fredj I'm not opposed to making `containsCoordinate` exportable.  I also sort of like that the docs describe how it should behave - then we could consider the current behavior a bug.  But I think it is a little misleading to say that part about returning `false` for coordinates that only touch a geometry since we know that isn't how this method works.

tschaub:
The quick fix would be to rename this to `ol.geom.Geometry.prototype.intersectsCoordinate`.

bartvde:
@fredj does the quick fix that @tschaub suggests also make sense to you?

fredj:
@bartvde code updated

bartvde:
LGTM please merge

fredj:
Thanks for the review

PR #4762, opened 2016-02-01
chore(jspm): add support of JSPM (Package Manager based on SystemJS)

davinkevin:
[JSPM Website](http://jspm.io/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,5 +101,16 @@
       "name": "vectortile",
       "browserify": true
     }
-  ]
+  ],
+  "jspm" : {
+    "main": "ol",
+    "dependencies": {
+      "css": "jspm:css@*"
+    },
+    "shim": {
+      "ol": {
+        "deps": ["./ol.css!"]
+      }
+    }
+  }
 }


tschaub:
@davinkevin thanks for the proposed change.  I'm in support of this type of change.  I imagine we'll get questions related to JSPM use, and I'm uncertain how many of the current contributors are JSPM users.  It would be great if in addition to proposing that the library support JSPM, you would be willing to help others interested in using JSPM and OpenLayers.

I imagine that if we merge this change, the [override in the JSPM registry](https://github.com/jspm/registry/commit/ff7f8720cb592c9bd03ad0b109029e3ed0d51d85) could be removed.  Is there anything else needed after this change to use OpenLayers with JSPM?

probins:
You can use an ol build with jspm at the moment without any problem - I've been doing so for a couple of years now. The registry is just a convenience.

davinkevin:
@tschaub Sorry for the late anwser, I didn't see your response.
What kind of help do you need for JSPM users ? I'll be glad to help if you want (and if my knowledge are sufficient enough).

@probins, Yes, we can use a custom build, but having the JSPM parameters in the package.json allow to define the dependency needed (like the css file for example).

tschaub:
Let's revisit this after we transform sources to ES modules for the 5.0 release.

PR #4765, opened 2016-02-01
Expand urls before setting this.urls

### LINKED ISSUE #4764: Tiled WMS source url expansion broken
An url like `https://{0-4}domain` used to be expanded to an array of subdomains.
The feature is broken in master/3.13.0 for Tiled WMS sources and maybe others.
It may be due to commit 4784b7f2 from PR #4219.

I edited the Tiled WMS example URLs to show a live example: http://jsfiddle.net/soveayvc/.
Note that **these URLs are just random subdomains and are not expected to actually resolve or serve tiles**.

In chrome, the console shows several lines like: 
```
GET http://wms%7B0-4%7D.geo.admin.ch/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&…=&BBOX=573671.3208%2C124338.29039999997%2C661473.0688%2C212140.03839999996 net::ERR_NAME_NOT_RESOLVED
```
Notice the escaped format and the presence of the `0-4` in the URL.
I would expect normal URLs instead, like `http://wms0.geo.admin.ch/...`.

An easy workaround is to use `ol.TileUrlFunction.expandUrl` in application code.
But is this situation expected and what can be done?

ahocevar:
Fixes #4764.

@gberaudo, we should really have tests for this. Feel free to create a new pull request with tests and cherry-pick my change here.

### DIFF
--- a/src/ol/source/urltilesource.js
+++ b/src/ol/source/urltilesource.js
@@ -183,8 +183,7 @@ ol.source.UrlTile.prototype.setTileUrlFunction = function(tileUrlFunction) {
  * @api stable
  */
 ol.source.UrlTile.prototype.setUrl = function(url) {
-  this.urls = [url];
-  var urls = ol.TileUrlFunction.expandUrl(url);
+  var urls = this.urls = ol.TileUrlFunction.expandUrl(url);
   this.setTileUrlFunction(this.fixedTileUrlFunction ?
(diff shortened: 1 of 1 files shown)


gberaudo:
Thanks @ahocevar.

ahocevar:
Superseded by #4777.

PR #4771, opened 2016-02-02
Use innerHTML instead of innerText to populate the status element

fredj:
Removes the `&nbsp;` string in the status element


bartvde:
LGTM thanks for fixing

tschaub:
Or alternatively replace `'&nbsp;'` with `' '` :)

PR #4772, opened 2016-02-02
Use node.setAttribute to set namespaceURI of a node

adube:
I faced an issue today while using the `ol.format.WMSGetFeatureInfo` format in an environment where ol3 is compiled along with the lib that uses it.  I have no idea why, but the following line failed under the compiled mode:

    node.namespaceURI = "some text"

I got an error like this:

> TypeError: Cannot set property namespaceURI of #<Element> which has only a getter

After reading what I found about this in the following [article on StackOverflow](http://stackoverflow.com/questions/32468352/uncaught-typeerror-cannot-set-property-style-of-htmlelement-which-has-only-a), I tried using the `Node#setAttribute` method instead and it worked.

This PR is the result of that.


adube:
Ready for review.

bartvde:
LGTM @adube thanks for the fix

PR #4778, opened 2016-02-03
Remove use of goog.object

tschaub:
Work in progress toward the removal of `goog.object.*`.

### DIFF
--- a/src/ol/object.js
+++ b/src/ol/object.js
@@ -1,54 +1,11 @@
 goog.provide('ol.Object');
 goog.provide('ol.ObjectEvent');
 goog.provide('ol.ObjectEventType');
-goog.provide('ol.object');
 
+goog.require('ol.Observable');
 goog.require('ol.events');
 goog.require('ol.events.Event');
-goog.require('ol.Observable');
-
-
-/**
- * Polyfill for Object.assign().  Assigns enumerable and own properties from
- * one or more source objects to a target object.
- *
- * @see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
- * @param {!Object} target The target object.
- * @param {...Object} var_sources The source object(s).
- * @return {!Object} The modified target object.
- */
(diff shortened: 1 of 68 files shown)


src/ol/object.js
```diff
@@ -8,6 +9,33 @@ goog.require('ol.Observable');
 
 
 /**
+ * Polyfill for Object.assign().  Assigns enumerable and own properties from
+ * one or more source objects to a target object.
```
marcjansen: This is from https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill, right? Maybe we should link there?

tschaub: Link added in https://github.com/tschaub/ol3/commit/34c705f150d8997aea9b976a76df7212ed2d4d54.

tschaub:
This is ready for review now.  The one remaining use of `goog.object` is in the KML format, but that should be going away (see #4781).

gberaudo:
Great work @tschaub. I added some suggestions. Feel free to do how you think best.

ahocevar:
This looks great @tschaub, thanks. Feel free to merge with or without addressing the comment about the `wantedTiles` count in `src/ol/map.js`.

tschaub:
Thanks for the reviews.  I'll merge when Travis is green.

One comment on PR reviews.  Comments left [on the diff](https://github.com/openlayers/ol3/pull/4778/files) persist through rebases (though they will be rendered collapsed like https://github.com/openlayers/ol3/pull/4778#discussion-diff-51721955).  Comments left directly on the commits are harder to find in the future (unless you remember the sha for the orphaned commit).

I try to respond to all comments before rebasing, but I think the end result is nicer if the comments are left on the diff instead of the commits.  Everyone has their own review style, but it works well for me to keep the diff open while reviewing individual commits.  Going to the diff to leave comments also means that you don't leave an outdated comment on something that was addressed by a subsequent commit.  Easy to slip up though.  And comments on individual commits have merit too - as they persist forever on that commit.

PR #4789, opened 2016-02-04
Use ol.array instead of goog.array

fredj:
And remove the `goog.array.ASSUME_NATIVE_FUNCTIONS` define in config files


bartvde:
LGTM

PR #4792, opened 2016-02-04
Use ol.events.listen instead of ol.Observable#on

fredj:
`ol.Observable#on` may return an array of `ol.events.Key`, `ol.events.listen` always
return a single `ol.events.Key`.


ahocevar:
Good catch, please merge.

PR #4793, opened 2016-02-04
[WIP] New type inference

gberaudo:
Switch to the new type inference version of the compiler.
See https://github.com/google/closure-compiler/wiki/Using-NTI-%28new-type-inference%29

There are only 2 errors left:
```
ERR! compile /home/gberaudo/dev/ol3/src/ol/observable.js:22: ERROR - ol.Observable cannot extend this type; structs can only extend structs
ERR! compile ol.Observable = function() {
ERR! compile                 ^
ERR! compile 
ERR! compile 
ERR! compile /home/gberaudo/dev/ol3/src/ol/renderer/maprenderer.js:38: ERROR - ol.renderer.Map cannot extend this type; structs can only extend structs
ERR! compile ol.renderer.Map = function(container, map) {
ERR! compile                   ^
ERR! compile 
ERR! compile 
ERR! compile 2 error(s), 0
ERR! compile  warning(s)
ERR! compile 
```

TODO:
- [x] fix `structs can only extend structs` error;
- [ ] fix publish.js

### DIFF
--- a/src/ol/events/event.js
+++ b/src/ol/events/event.js
@@ -38,7 +38,11 @@ ol.events.Event = function(type, opt_target) {
 /**
  * Stop event propagation
  */
-ol.events.Event.prototype.preventDefault =
+ol.events.Event.prototype.preventDefault = function() {
+  this.stopPropagation();
+};
+
+
 ol.events.Event.prototype.stopPropagation = function() {
   this.propagationStopped = true;
 };
--- a/src/ol/format/gml/gml3format.js
+++ b/src/ol/format/gml/gml3format.js
@@ -1332,27 +1332,3 @@ ol.format.GML3.prototype.writeFeaturesNode = function(features, opt_options) {
  * @api stable
  */
 ol.format.GML = ol.format.GML3;
-
-
-/**
- * Encode an array of features in GML 3.1.1 Simple Features.
- *
- * @function
- * @param {Array.<ol.Feature>} features Features.
- * @param {olx.format.WriteOptions=} opt_options Options.
- * @return {string} Result.
- * @api stable
- */
-ol.format.GML.prototype.writeFeatures;
-
-
-/**
- * Encode an array of features in the GML 3.1.1 format as an XML node.
- *
- * @function
- * @param {Array.<ol.Feature>} features Features.
- * @param {olx.format.WriteOptions=} opt_options Options.
- * @return {Node} Node.
- * @api
- */
-ol.format.GML.prototype.writeFeaturesNode;
--- a/config/jsdoc/info/publish.js
+++ b/config/jsdoc/info/publish.js
@@ -69,12 +69,13 @@ exports.publish = function(data, opts) {
           types: ['{}']
         });
       } else {
-        var typedef = typedefs[typedefs.length - 1];
-        var type = typedef.types[0];
-        typedef.types[0] = type
-            .replace(/\}$/, ', ' + doc.longname.split('#')[1] +
-                ': (' + getTypes(doc.type.names).join('|') + ')}')
-            .replace('{, ', '{');
+// Nothing to do!
+//        var typedef = typedefs[typedefs.length - 1];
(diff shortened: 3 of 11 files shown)


gberaudo:
@fredj, if you want to have a look...

src/ol/events/event.js
```diff
@@ -47,7 +47,7 @@ ol.events.Event.prototype.stopPropagation = function() {
 /**
  * @param {Event|ol.events.Event} evt Event
  */
-ol.events.Event.stopPropagation = function(evt) {
+ol.events.Event.prototype.stopPropagation = function(evt) {
```
tschaub: This looks like a mistake to me.  There are still a number of places in the library that use the `ol.events.Event.stopPropagation` function.  With this change, that will be `undefined`.  I hope this causes at least a few test failures.

Not sure why you'd want it to be a prototype member.  A name like `ol.events.stopPropagation` could make sense, but it doesn't need to be accessible from `ol.events.Event` instances.

src/ol/events/event.js
```diff
@@ -55,6 +55,6 @@ ol.events.Event.stopPropagation = function(evt) {
 /**
  * @param {Event|ol.events.Event} evt Event
  */
-ol.events.Event.preventDefault = function(evt) {
+ol.events.Event.prototype.preventDefault = function(evt) {
```
tschaub: Same as above.  This is only used [one place](https://github.com/openlayers/ol3/blob/34d8d77344ee0b653770f065c593d4ab7b5d102b/src/ol/renderer/dom/dommaprenderer.js#L70-L72) in the library, but it cannot be renamed here without changing it there.  Again, if we are going to rename it, `ol.events.preventDefault` makes sense.  But again, it doesn't need to be accessible from every `ol.events.Event` instance.

src/ol/format/gml/gml3format.js
```diff
@@ -1332,27 +1332,3 @@ ol.format.GML3.prototype.writeFeaturesNode = function(features, opt_options) {
...
-/**
- * Encode an array of features in GML 3.1.1 Simple Features.
- *
- * @function
- * @param {Array.<ol.Feature>} features Features.
- * @param {olx.format.WriteOptions=} opt_options Options.
- * @return {string} Result.
- * @api stable
- */
-ol.format.GML.prototype.writeFeatures;
```
tschaub: Are we going to need another method of [documenting this method](http://openlayers.org/en/v3.14.1/apidoc/ol.format.GML3.html#writeFeatures)?

src/ol/format/gml/gml3format.js
```diff
@@ -1332,27 +1332,3 @@ ol.format.GML3.prototype.writeFeaturesNode = function(features, opt_options) {
...
-/**
- * Encode an array of features in the GML 3.1.1 format as an XML node.
- *
- * @function
- * @param {Array.<ol.Feature>} features Features.
- * @param {olx.format.WriteOptions=} opt_options Options.
- * @return {Node} Node.
- * @api
- */
-ol.format.GML.prototype.writeFeaturesNode;
```
tschaub: Same question here regarding [documentation](http://openlayers.org/en/v3.14.1/apidoc/ol.format.GML3.html#writeFeaturesNode).

gberaudo:
The `structs can only extend structs` error seems to be gone after @tschaub's PR.
However there are lots of other errors, notably about nullable types.

probins:
@gberaudo I notice a couple of typedef-related changes in here. First, LRUCache looks like it should be changed to an interface, so I won't include this in my typedef PR. Second, in `raster/operation.js` you're adding `goog.require('ol.raster.Pixel')` even though that is only used in comments/typedefs. Were you getting some sort of error here that prompted you to add the `goog.require()`?

gberaudo:
@probins, thank you for caring about this PR.
I think it is better to ignore the changes here since some of them may be due to goog dependencies or the way the code is organized. Let's move the typedefs as discussed.

PR #4794, opened 2016-02-05
Make tile loading count no longer depend on source count

### LINKED ISSUE #4790: Consider keeping max tile loads constant
Currently, the number of concurrent tile loads allowed [scales with the number of tile sources](https://github.com/openlayers/ol3/blob/master/src/ol/map.js#L1032-L1033).  This means degraded application performance with more tile sources.  We should try keeping this constant and test performance in applications with many tile sources.

See https://github.com/tschaub/ol3/commit/c2831c7e8517609d5e5792e5db2d7f2348cf682a#commitcomment-15879564 for the original discussion.

ahocevar:
This pull request changes the tile queue behaviour so the number of loading tiles is no longer influenced by the number of sources that request tiles. This should give better user experience in applications with many tile layers.

Fixes #4790.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -7008,7 +7008,6 @@ olx.view.FitOptions.prototype.maxZoom;
  *     size: ol.Size,
  *     skippedFeatureUids: Object.<string, boolean>,
  *     tileQueue: ol.TileQueue,
- *     tileSourceCount: number,
  *     time: number,
  *     usedTiles: Object.<string, Object.<string, ol.TileRange>>,
  *     viewState: olx.ViewState,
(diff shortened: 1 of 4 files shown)


ahocevar:
@marcjansen, you may want to test this in one of your applications with many tile layers.

marcjansen:
I am mostly offline, maybe @weskamm may give this a try

fredj:
ping @marcjansen

marcjansen:
Sorry I won't have time to review.

weskamm:
i also dont have time to test this, sorry

ahocevar:
This is a change that only affects applications with multiple tile layers. For two layers, like in the `wms-tiled.html` example, I notice no negative impact on the user experience.

Anyone else up for a review?

fredj:
LGTM

ahocevar:
Thanks @fredj.

PR #4796, opened 2016-02-05
Remove use of goog.isString()

marcjansen:
This PR replaces all usages of `goog.isString()` with inlined `typeof xxx === 'string'` checks.

I am not sure if we want this, or rather a dedicated method `ol.isString()`.

### DIFF
--- a/src/ol/color/color.js
+++ b/src/ol/color/color.js
@@ -106,7 +106,7 @@ ol.color.asArray = function(color) {
   if (goog.isArray(color)) {
     return color;
   } else {
-    goog.asserts.assert(goog.isString(color), 'Color should be a string');
+    goog.asserts.assert(typeof color === 'string', 'Color should be a string');
     return ol.color.fromString(color);
   }
 };
@@ -119,7 +119,7 @@ ol.color.asArray = function(color) {
  * @api
  */
 ol.color.asString = function(color) {
-  if (goog.isString(color)) {
+  if (typeof color === 'string') {
     return color;
   } else {
     goog.asserts.assert(goog.isArray(color), 'Color should be an array');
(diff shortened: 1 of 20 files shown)


tschaub:
I think we can get away without having `ol.isString()`.  The `goog.assert` functions will go away as well, and there aren't that many remaining potential call sites.  I'm curious if the compiler would inline this anyway.

marcjansen:
Some statistics:

**master**

```
info ol Parsing dependencies
info ol Compiling 327 sources
uncompressed: 488864 bytes
  compressed: 144023 bytes
node tasks/build.js config/ol-debug.json build/ol-debug.js
info ol Parsing dependencies
info ol No compile options found.  Concatenating 327 sources
uncompressed: 3499330 bytes
  compressed: 639805 bytes
```

**this branch**

```
info ol Parsing dependencies
info ol Compiling 327 sources
uncompressed: 489336 bytes
  compressed: 144046 bytes
node tasks/build.js config/ol-debug.json build/ol-debug.js
info ol Parsing dependencies
info ol No compile options found.  Concatenating 327 sources
uncompressed: 3499544 bytes
  compressed: 639939 bytes
```

This branch is always bigger:

```
compiled uncompressed: +472 bytes
compiled compressed: +23 bytes
concatenated uncompressed: +214 bytes
concatenated compressed: +134 bytes
```

marcjansen:
If we can live with this increased sizes, I can rename the branch and merge.

ahocevar:
I think this looks good. Build size will shrink when all goog.* components that pull in goog.isString are gone.

marcjansen:
Cheerio for the reviews, @tschaub & @ahocevar.

marcjansen:
I need to rebase, then I'll merge

PR #4804, opened 2016-02-08
Use Array.isArray instead of goog.isArray

fredj:
Depends on https://github.com/google/closure-compiler/pull/1451


tschaub:
It would be nice if the compiler understood `Array.isArray()` in the way it understands `goog.isArray()`.

tschaub:
> It would be nice if the compiler understood Array.isArray() in the way it understands goog.isArray().

Left that comment before reading the associated issues.

fredj:
Ready to be reviewed

marcjansen:
I like this. +1 on merging. @tschaub?

PR #4805, opened 2016-02-08
Add new cacheSize option to ol.source

### LINKED ISSUE #4702: Change value DEFAULT_TILE_CACHE_HIGH_WATER_MARK in code
Exist the other option, how can i set value DEFAULT_TILE_CACHE_HIGH_WATER_MARK without own build library OL3? My application can be compilated on more computers and our problem is hold everywhere the same version, moreover we use the package manager NPM. I need set this value before loading map.

Thanks.

fredj:
Fixes #4702 

Option added to:
 - olx.source.BingMapsOptions
 - olx.source.MapQuestOptions
 - olx.source.OSMOptions
 - olx.source.StamenOptions
 - olx.source.TileArcGISRestOptions
 - olx.source.TileImageOptions
 - olx.source.TileJSONOptions
 - olx.source.TileWMSOptions
 - olx.source.XYZOptions
 - olx.source.ZoomifyOptions

All sources except vector, static image and single image

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,19 @@
 ## Upgrade notes
 
+### v3.15.0
+
+#### Removal of `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK`
+
+The `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK` define has been removed. The size of the cache can now be defined on every tile based `ol.source`:
+```js
+new ol.layer.Tile({
+  source: new ol.source.OSM({
+    cacheSize: 128
+  })
+})
+```
+The default cache size is `2048`.
+
 ### v3.14.0
 
 #### Internet Explorer 9 support
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3904,7 +3904,8 @@ olx.source;
 
 
 /**
- * @typedef {{culture: (string|undefined),
+ * @typedef {{cacheSize: (number|undefined),
+ *     culture: (string|undefined),
  *     key: string,
  *     imagerySet: string,
  *     maxZoom: (number|undefined),
@@ -3914,6 +3915,14 @@ olx.source;
  * @api
  */
 olx.source.BingMapsOptions;
+
+
+/**
+ * Cache size. Default is `2048`.
+ * @type {number|undefined}
+ * @api
+ */
+olx.source.BingMapsOptions.prototype.cacheSize;
 
 
 /**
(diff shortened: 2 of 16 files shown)


ahocevar:
I'm wondering if the tile cache should be per-map instead of per-source. Opinions?

lol2x:
Is there any plan to merge it to dist version? Actually usage of RAM in firefox is unaceptable (especially on older PC's) or there is any "newer" way to solve caching size (without rebuilding whole ol)

ahocevar:
After doing  bit more research, I think it is reasonable to continue with the approach you proposed here @fredj.

An improvement for another pull request would be to completely disable the tile cache or set its size to 0 for sources that are known to send sensible cache headers, because the browser cache properly handles them. I did some testing with OSM, and I didn't notice any visual performance penalties with a disabled tile cache.

fredj:
ok, I will finish this PR

fredj:
thanks for any review

lol2x:
@fredj Is it added to ol.source.WMTS as well? Thanks for that commit, will make my work easier

fredj:
Yes, it has been added to `ol.source.WMTS`

ahocevar:
Thanks @fredj. Two minor comments:

* The docs are not 100% correct (default is the value of `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK`, which has a default of `2048`). Maybe you can change that?
* There is also ol.source.VectorTile which has a tile cache, but with a smaller default size. Maybe you want to make that configurable as well?

Both comments can also be addressed in separate pull requests, but if you're still at it, you might want to address them here.

It's up to you to merge now or after making these changes :smile:.

fredj:
> The docs are not 100% correct (default is the value of `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK`, which has a default of `2048`). Maybe you can change that?

It's already `2048` in externs/olx.js or do I miss something ?

> There is also ol.source.VectorTile which has a tile cache, but with a smaller default size. Maybe you want to make that configurable as well?

Yes, I will add the option to ` ol.source.VectorTile`

ahocevar:
> It's already 2048 in externs/olx.js or do I miss something?

I was just wondering if `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK` should be mentioned, because it can be changed as a compiler define.

fredj:
> I was just wondering if `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK` should be mentioned, because it can be changed as a compiler define.

ok, got it.

Now that we have a simple way to change this value what about removing `ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK` ?

ahocevar:
> Now that we have a simple way to change this value what about removing ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK ?

:+1: - with an upgrade note.

ahocevar:
Great work, thanks @fredj.

PR #4808, opened 2016-02-10
Type cleanup

fredj:
All these error have been found while I was working on trying to simplify `externs/olx.js` by using `@record` instead of `@typedef` (see https://github.com/google/closure-compiler/wiki/Structural-Interfaces-in-Closure-Compiler)

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1350,7 +1350,7 @@ olx.control.RotateOptions.prototype.className;
 /**
  * Text label to use for the rotate button. Default is `⇧`.
  * Instead of text, also a Node (e.g. a `span` element) can be used.
- * @type {string|Node|undefined}
+ * @type {string|Element|undefined}
  * @api stable
  */
 olx.control.RotateOptions.prototype.label;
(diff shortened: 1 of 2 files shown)


fredj:
Thanks for any review.

src/ol/tilegrid/tilegrid.js
```diff
@@ -541,9 +541,7 @@ ol.tilegrid.createForExtent = function(extent, opt_maxZoom, opt_tileSize, opt_co
  * @api
  */
 ol.tilegrid.createXYZ = function(opt_options) {
-  var options = /** @type {olx.tilegrid.TileGridOptions} */ ({});
-  ol.object.assign(options, opt_options !== undefined ?
-      opt_options : /** @type {olx.tilegrid.XYZOptions} */ ({}));
+  var options = ol.object.assign({}, opt_options ? opt_options : {});
```
gberaudo: Shouldn't `options` be casted to `olx.tilegrid.TileGridOptions` at this point?

fredj: `options` can't be a `olx.tilegrid.TileGridOptions` because it has the `maxZoom` property

fredj:
Any other review?

petrsloup:
I think all these changes make sense. LGTM

fredj:
Thanks for the review

PR #4809, opened 2016-02-10
fix multipolygon clone

fperucic:


tschaub:
Thanks for the proposed fix @fperucic.

Can you put together a test that demonstrates the correct behavior (and prevents regressions)?

fperucic:
I created a simple test, please see if it is enough

tschaub:
Thanks @fperucic.  Looks good.

tschaub:
After some deep debugging, I've just run into what looks like a regression related to cloning MultiPolygon.  I'm still trying to figure out how it was introduced, but in 3.11.1 things were working and in 3.14.0 we're seeing an issue where you get into an infinite loop trying to call `multiPolygon.clone().getCoordinates()`.

I'll put together a test to demonstrate.

tschaub:
If anybody is interested in hanging their browser, try running this snippet: https://gist.github.com/tschaub/4a1ce0869a4dd2e64cc7

Hangs if run here: http://openlayers.org/en/v3.14.1/examples/geojson.html
Works if run here: http://openlayers.org/en/v3.13.0/examples/geojson.html

Basically, it comes down to this:

```js
var offset = [178];
var end = [568];
var stride = 2;
for (var j = offset; j < end; j += stride) {
  // this is taking forever!
}
```

Things that are interesting about this:

 * the compiler isn't saving us from passing `Array<number>` where we have `@type {number}`
 * there still aren't enough tests to catch this
 * comparing arrays as numbers is fun

Handy tip: `pkill Chrome` (to kill Chrome when it hangs above).

Thanks @aisaacs for the debugging fun.

@ahocevar this is the regression I was talking about this morning.  The change from @fperucic here fix the problem (cloning MultiPolygon was very broken).  I didn't do a bisect to figure out when the problem was introduced, but in 3.13 calling `getCoordinates()` on a cloned MultiPolygon didn't result in an endless loop.  Even though cloning MultiPolygon was broken then as well.

tschaub:
Mea culpa: d67f33bcd66a8dd9152e6be91043548a0c661d81

marcjansen:
Not solely your bad. No tests!

ahocevar:
Thanks @tschaub and @aisaacs for the investigation, and thanks @fperucic for the fix. I'm going to include this in the 3.14.2 release.

PR #4811, opened 2016-02-10
Make ol.style.Style a @struct

fredj:
And take into account that `styles` may be a style or an array of style.

### DIFF
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -2390,10 +2390,10 @@ ol.format.KML.writePlacemark_ = function(node, feature, objectStack) {
     // FIXME the styles returned by the style function are supposed to be
     // resolution-independent here
     var styles = styleFunction.call(feature, 0);
-    if (styles && styles.length > 0) {
-      var style = styles[0];
+    if (styles) {
+      var style = goog.isArray(styles) ? styles[0] : styles;
       if (this.writeStyles_) {
-        properties['Style'] = styles[0];
+        properties['Style'] = style;
       }
       var textStyle = style.getText();
       if (textStyle) {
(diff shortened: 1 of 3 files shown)


src/ol/format/kmlformat.js
```diff
@@ -2390,10 +2390,10 @@ ol.format.KML.writePlacemark_ = function(node, feature, objectStack) {
     // FIXME the styles returned by the style function are supposed to be
     // resolution-independent here
     var styles = styleFunction.call(feature, 0);
-    if (styles && styles.length > 0) {
-      var style = styles[0];
+    if (styles) {
+      var style = goog.isArray(styles) ? styles[0] : styles;
```
ahocevar: Would be nice to use `Array.isArray()` instead.

src/ol/source/imagevectorsource.js
```diff
@@ -264,6 +264,9 @@ ol.source.ImageVector.prototype.renderFeature_ = function(feature, resolution, p
     return false;
   }
   var i, ii, loading = false;
+  if (!goog.isArray(styles)) {
```
ahocevar: idem re `isArray`

src/ol/style/style.js
```diff
@@ -19,6 +19,7 @@ goog.require('ol.style.Stroke');
  * feature or layer that uses the style is re-rendered.
  *
  * @constructor
+ * @struct
```
ahocevar: What exactly are the implications of marking this `@struct`?

src/ol/format/kmlformat.js
fredj: We can't use Array.isArray because of https://github.com/google/closure-compiler/pull/1451

src/ol/style/style.js
fredj: To prevent the use of properties not declared in the constructor. We use on every class except here and in ol.format.Format

ahocevar:
Thanks for the explanations, please merge @fredj.

fredj:
Thanks for the review

PR #4815, opened 2016-02-11
wrapX false in synthetic examples 

fredj:


ahocevar:
Makes sense, please merge.

PR #4816, opened 2016-02-11
Improve canvas rendering performances

fredj:
Improve the BEGIN_GEOMETRY instruction by computing the feature's uid and extent
in ol.render.canvas.Replay#beginGeometry instead of ol.render.canvas.Replay#replay_

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -209,11 +209,13 @@ ol.render.canvas.Replay.prototype.appendFlatCoordinates = function(flatCoordinat
  * @param {ol.Feature|ol.render.Feature} feature Feature.
  */
 ol.render.canvas.Replay.prototype.beginGeometry = function(geometry, feature) {
+  var uid = goog.getUid(feature).toString();
+  var extent = geometry ? geometry.getExtent() : null;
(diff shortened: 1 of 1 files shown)


ahocevar:
While I agree that this could improve performance for the first hit detection or feature selection, it probably has a performance impact on rendering. The uid is only needed for the skipped features lookup, and the extent is only needed for hit detection. Both are irrelevant when we just want to render.

I assume that the compiler inlines calls to goog.getUid() and geometry.getExtent(), so with the current code there should only be a performance hit when uid and extent are needed for the first time.

My other concern is that storing the uid in the replay instruction uses more memory than storing a reference to the feature. This might be a problem with vector tile rendering, where we cache replays.

Do you have any profiling results that teach us otherwise?

ahocevar:
I've been thinking about this more. The replay knows about the feature. So what if we got rid of the `skippedFeatureUids` hash on the map, and instead add a `renderHint` property to `ol.Feature`. This would be an enum with currently one value, `SKIP`.

Later, we could add another value, like `OVERLAY`, where the replay could skip the feature and render it on top of everything else in a postrender function. Then interactions like Select would not need to maintain a separate layer for overlaid features any more, and the problems we have with features duplicated on the interaction's internal layer would go away.

PR #4818, opened 2016-02-11
beforedrawstart event triggered before draw start

riverstream:
Create beforedrawstart event triggered before draw start

### DIFF
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -368,7 +368,7 @@ ol.interaction.Draw.handleDownEvent_ = function(event) {
   var continueDrawing = this.dispatchEvent(new ol.interaction.DrawEvent(
       ol.interaction.DrawEventType.BEFOREDRAWSTART, event));
   if (continueDrawing === false) {
-      return false;
+    return false;
   }
   if (this.condition_(event)) {
     this.downPx_ = event.pixel;
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -35,6 +35,12 @@ goog.require('ol.source.Vector');
  */
 ol.interaction.DrawEventType = {
   /**
+   * Triggered before feature draw start
+   * @event ol.interaction.DrawEvent#beforedrawstart
+   * @api stable
+   */
+  BEFOREDRAWSTART: 'beforedrawstart',
+  /**
    * Triggered upon feature draw start
    * @event ol.interaction.DrawEvent#drawstart
    * @api stable
@@ -359,6 +365,11 @@ ol.interaction.Draw.handleEvent = function(mapBrowserEvent) {
  * @private
  */
 ol.interaction.Draw.handleDownEvent_ = function(event) {
+  var continueDrawing = this.dispatchEvent(new ol.interaction.DrawEvent(
+      ol.interaction.DrawEventType.BEFOREDRAWSTART, event));
+  if (continueDrawing === false) {
+      return false;
+  }
   if (this.condition_(event)) {
     this.downPx_ = event.pixel;
     return true;


riverstream:
Sample Usage:
draw.on('beforedrawstart',
                function (evt) {
                    // Validate before drawing starts
                    if (!validateFunction())
                    {
                        return false;
                    }
                }, this);

ahocevar:
For interactions, we use conditions elsewhere, not events that allow listeners to stop execution. What exactly are you trying to achieve?

riverstream:
Need to allow user to select tool to create a point (or other shape) but not allow to actually create the feature if certain validation does not pass.  This was done with beforefeatureadded event in OpenLayers 2 and am porting code to OL3.

ahocevar:
How about configuring the Draw interaction with a collection of features instead of a source?
```js
var source = new ol.source.Vector();
var features = new ol.Collection();
var draw = new ol.interaction.Draw({
  features: features
});
features.on('add', function(e) {
  var feature = e.feature;
  // check condition here
  if (pass) {
    source.addFeature(feature);
  }
  features.pop();
}
```

riverstream:
Andreas, nice work around.  Tested your suggestion.  But by adding and removing the feature from the screen causes a "flicker" effect.  Since beforedrawstart is an optional event and not required to be created, there would me no noticiable perfermance hit on the client side.

ahocevar:
See http://jsfiddle.net/f1rcsjnp/. No flicker, no add and remove.

riverstream:
Andreas, please check [http://jsfiddle.net/f1rcsjnp/7/](http://jsfiddle.net/f1rcsjnp/7/) Still flickering in IE9 and IE10 on feature add/remove.  Did not test in IE11.

ahocevar:
@riverstream I don't know what you mean by feature add/remove - the `features` collection is not rendered anywhere. What's rendered is only the feature you're currently drawing, and then all features that are added to the source.

riverstream:
The feature currently drawn is the one that flickers.  When I mean feature add/remove, I was referring to the collection variable "features".

ahocevar:
@riverstream It would flicker the same way with your pull request, because the rendering workflow is exactly the same. So I closed this pull request.

riverstream:
Andreas, to get back on track and the reason of opening this pull request is to stop the feature from being created at all.  Same behavior as in OL2 as **beforefeatureadded** is what I was trying to reproduce:  [http://dev.openlayers.org/apidocs/files/OpenLayers/Layer/Vector-js.html](http://dev.openlayers.org/apidocs/files/OpenLayers/Layer/Vector-js.html)

Here is the description:

**beforefeatureadded**  = Triggered before a feature is added.  Listeners will receive an object with a feature property referencing the feature to be added.  **To stop the feature from being added, a listener should return false**

riverstream:
No, it would not flicker with the pull request I created because you never actually add the feature to the canvas.  Please test.  Thank you.

ahocevar:
> I created because you never actually add the feature to the canvas

The same is the case with my code. I'm unable to see a flicker with my code, so there is no point in comparing.

riverstream:
There is flicker with your code because you are actually adding the feature to the canvas and then deleting it.  With the pull request I created, the feature is never actually added and therefore does not cause flickering.

ahocevar:
Ah, your pull request looks different now than when I first looked at it. The way it is now, it isn't adding anything new. You can simply set a draw condition. So instead of
```js
draw.on('beforedrawstart', function (evt) {
  // Validate before drawing starts
  return validateFunction(evt);
}, this);
```
you can do
```js
new ol.interaction.Draw({
  condition: function(evt) {
    return ol.condition.noModifierKeys(evt) && validateFunction(evt)
  }
});

**Comment by riverstream** [comment]
Thanks that worked.  I just set it to:

```
condition: function (evt) {
                    return validateFunction(evt);
                }
```

Because  ol.condition.noModifierKeys(evt) gave an error

**Comment by riverstream** [comment]
Is there a way to prevent or determine if the user is dragging and not creating a feature?  It is trying to execute the vaildateFunction while dragging also.

**Comment by ahocevar** [comment]
Yes, `ol.MapBrowserEvent` has a [dragging](http://openlayers.org/en/v3.13.1/apidoc/ol.MapBrowserEvent.html#dragging) property.

**Comment by riverstream** [comment]
Andreas, tried this:

```
function validateFunction(evt)
        {

            // Do not validate while dragging
            if (evt.dragging)
            {
                return true;
            }
...
```

But always is false even when dragging.

**Comment by ahocevar** [comment]
Probably because you have a DragPan interaction which catches and stops dragging events.

**Comment by riverstream** [comment]
Check this example never executes dragging logic: [http://jsfiddle.net/f1rcsjnp/9/](http://jsfiddle.net/f1rcsjnp/9/)

**Comment by ahocevar** [comment]
The behaviour you are seeing is correct. The handled event is the pointerdown event. At this point there is no dragging yet.

**Comment by riverstream** [comment]
My apologies.   When I said "dragging" I actually meant panning in the GIS world.  It is executing the vaildateFunction while panning.

**Comment by riverstream** [comment]
To give you an idea of how it is suppose to work, I created an example in OL2 which only blocks point creation but still allows panning: [http://jsfiddle.net/qSK87/6/](http://jsfiddle.net/qSK87/6/)

**Comment by ahocevar** [comment]
The draw workflow is different in OpenLayers 3. The condition determines whether you allow to start a drawing. When you do, a sketch feature is created. All this happens on pointer down. When done drawing (for points on pointer up), the drawn feature is added to a features collection or a source. Code snippets on this ticket show you how to hook in either phase of this workflow. I tend to think that what you want is what I suggested first.

**Comment by ahocevar** [comment]
Also note that no feature will be added when you pan between pointer down and up, because the DragPan interaction stops the up event then.

**Comment by riverstream** [comment]
Ok here is an update, I tried using the collection method you originally suggested and that would not work correctly since sometimes it creates one feature but other times (after double clicking the canvas to complete the feature), it adds two identical features to the collection instead of just one feature.  It went back to try to using the second suggestion with the source layer for the ol.interaction.Draw.  This adds features correctly but again the problem is the **condition:** is executing not only when a feature is added but also while panning.  My question is, inside my validateFunction, how can I differentiate between a feature complete drawn or a panning of the canvas.

For example:

```
function validateFunction(evt)
        {
// If panning do not validate
if (evt.panning === true)
{
return true; //exit validate if this is a panning event
}
// continue on to validation here before adding a new feature.
...
}
```

ahocevar:
> My question is, inside my validateFunction, how can I differentiate between a feature complete drawn or a panning of the canvas.

You cannot. With or without your pull request, the validateFunction gets the pointerdown event - this is before you know whether it's going to be a drag or a doubleclick.

riverstream:
Andreas thank you, you are correct. I have tested and agree.  Also, it does not work with the collections work-around since a single click produces multiple features (tested in IE 9 and IE 10).  How can I accomplish the following in OL3:

[http://jsfiddle.net/qSK87/6/](http://jsfiddle.net/qSK87/6/)

Thank You

ahocevar:
I feel like we are spinning in circles here. Maybe this fiddle helps, which does exactly the same as your OpenLayers 2 fiddle:

http://jsfiddle.net/6LdLs07u/

Issue #4822, opened 2016-02-12
Having 2 select interactions active at the same time


probins:
I reported in https://github.com/openlayers/ol3/issues/3878#issuecomment-182061870 that the problem I had when 2 select interactions are active at the same time was reoccurring. I've been investigating this further, and it looks like the problem only occurs when there is a 'layers' option. I tried reproducing the problem on the select-features example, and failed, but then noticed that there is no 'layers' option there. I am using the layers option in my code simply to restrict the layers to the vector layers, which should make it marginally faster, but is not essential. If I remove this option, the code works again.

So, strictly speaking, this is a regression: code which was working before 3.12 no longer works, but it's not necessarily an important one. However, if there is a limitation on when the 'layers' option can be used, then this should be documented. There was a release note on forEachFeatureAtPixel, but it's not at all obvious that this also applies to the 'layers' option on the select interaction.

Also, I notice that the docs for the layers option currently say this is 'Required'; this is surely not correct.

ahocevar:
This issue was probably fixed with #4736. Are you able to test that?

For the documentation issue, I created #4832.

probins:
My previous code does not work with master (at least, when I tried a couple of days ago), so, no, #4736 doesn't fix the issue.

ahocevar:
Can you provide a JSFiddle with your test setup?

probins:
http://jsfiddle.net/4exzxbr3/ adapted from select-features example.

Hover correctly selects, click select is triggered but `e.selected` is 0.
Comment out line 30, and click `e.selected` is correct: 1
Remove the layers option on both interactions, and `e.selected` is correct.

ahocevar:
@probins I got a fix - see #4854.

probins:
I'm off on my travels again on Friday, and have a lot to do before then. I'll try and find time to test this with my code. However, if my fiddle above works with the new patch, then it's likely my code will too, so I'd be fine with merging your change.

probins:
@ahocevar just tested this on my code. Can confirm that my code both with and without layers option seems to work now. Thanks a lot. :-)

PR #4823, opened 2016-02-12
Fix type for layers option on ol.interaction.Select

ahocevar:
Parsing of type definitions is fragile for function with a return type. Elsewhere we put other types first to avoid this issue. By doing the same for ol.interaction.Select, we avoid that the `layers` option is marked 'Required' (see #4822).


marcjansen:
Thanks for this. Please merge once everything is green.

PR #4824, opened 2016-02-12
Don't use goog.string.newlines.*

marcjansen:
…instead simply use `String.prototype.split` with a regular expression.

I was also thinking of inlining the regular expression, so if everybody likes that better, I can update this PR.

Please review.


tschaub:
Looks good.  I think it is nice to create once and reuse the regex.  Please merge.

marcjansen:
Thanks again, Tim.

marcjansen:
I had to rebase. Will merge once everything is green again.

PR #4825, opened 2016-02-12
Don't use goog.string.isEmpty

marcjansen:
…instead test directly against an appropriate regular expression.

In order to give some comments about the regular expression I chose to declare it as private constant.

Please review.


tschaub:

marcjansen:
Thanks Tim.

PR #4826, opened 2016-02-12
Don't use goog.string.endsWith

marcjansen:
And instead use our own utility method which in turn uses `String.prototype.endsWith` if available.

Please review.


marcjansen:
Thanks for any review.

ahocevar:
I think that only occurrence of `goog.string.endsWith` can be solved in a simpler way with regular expressions. I created #4826.

marcjansen:
I also prefer #4838. Closing in favor of that PR.

PR #4827, opened 2016-02-12
JSDoc fixes

jumpinjackie:
This pull request contains various fixes to JSDoc comments to not only improve the API docs of function types, but also to make the ol3 source more amenable to be able to generate a TypeScript definition from (which I am currently working on through my JSDoc plugin: https://github.com/jumpinjackie/jsdoc-typescript-plugin)

 * Use `@callback` with `@param` and `@return` (where appropriate) to document function types instead of `@typedef`
 * `olx.control.OverviewMapOptions#layers` should have `ol.Collection<ol.layer.Layer>` instead of `ol.Collection`
 * Remove an unused `goog.require` from `src/ol/tilequeue.js` as a result of the wholesale change to `@callback`

The change to `@callback` results in documentation that currently looks like this (v3.13.1)

ol3_typedefs_before

To this

ol3_typedefs_after

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1184,7 +1184,7 @@ olx.control.MousePositionOptions.prototype.undefinedHTML;
  *     collapseLabel: (string|Node|undefined),
  *     collapsible: (boolean|undefined),
  *     label: (string|Node|undefined),
- *     layers: (Array.<ol.layer.Layer>|ol.Collection|undefined),
+ *     layers: (Array.<ol.layer.Layer>|ol.Collection<ol.layer.Layer>|undefined),
  *     render: (function(ol.MapEvent)|undefined),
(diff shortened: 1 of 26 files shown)


jumpinjackie:
Closing this. What I want to do (using `@callback`) is incompatible with the closure compiler

tschaub:
That's a pretty big improvement in the docs.  The major drawback of `@typedef` is that we cannot document the individual properties.  It would be nice to find a way to achieve this while still providing annotations that the compiler can make sense of.

fredj:
> `olx.control.OverviewMapOptions#layers` should have `ol.Collection<ol.layer.Layer>` instead of `ol.Collection`

This one can be fixed, see #4839

PR #4830, opened 2016-02-13
Make sure window.proj4 is always restored in tests

ahocevar:
When the OpenLayers test suite is used with a custom build of proj4.js that does not include the `somerc` projection, it fails with the very misleading error `Uncaught ReferenceError: proj4 is not defined`. This is because the 'can use an alternative namespace for proj4' test fails and does not restore `window.proj4`. See proj4js/proj4js#170 for a more background.

By restoring `window.proj4` in an `afterEach` function, we can avoid this misleading error.


bartvde:
good catch, LGTM

fredj:
+1

PR #4836, opened 2016-02-14
Use lowercase for all user agent checks

ahocevar:
The box-selection example is currently broken, because the `ol.has.MAC` check is incorrect. This regression is fixed with the change here.


tschaub:
Looks good.  Please merge.

PR #4838, opened 2016-02-14
Use regular expression instead of String#endsWith() check

ahocevar:
Alternative to #4826.


marcjansen:
I loke this one better than #4826. Please merge once the checks are green.

PR #4843, opened 2016-02-15
Remove use of goog.dom.getParentElement

fredj:


ahocevar:
Looks good, please merge.

PR #4845, opened 2016-02-15
Fix geolocation error cast

gberaudo:
Despite the dispatchEvent method accepting a hash with a `target` property of `undefined`, closure may still require the `target` property to be explicit.

### DIFF
--- a/src/ol/geolocation.js
+++ b/src/ol/geolocation.js
@@ -191,7 +191,7 @@ ol.Geolocation.prototype.positionChange_ = function(position) {
 ol.Geolocation.prototype.positionError_ = function(error) {
   error.type = ol.events.EventType.ERROR;
   this.setTracking(false);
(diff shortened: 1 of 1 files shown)


gberaudo:
Avoid the following error:
```
ERR! compile /home/gberaudo/dev/ol3-cesium/ol3/src/ol/geolocation.js:194: ERROR - actual parameter 1 of ol.Observable.prototype.dispatchEvent does not match formal parameter
ERR! compile found   : {type: string}
ERR! compile required: (null|ol.events.Event|string|{target: (EventTarget|null|ol.events.EventTarget|undefined), type: string})
ERR! compile   this.dispatchEvent(/** @type {{type: string}} */ (error));
```

Uncovered by https://github.com/openlayers/ol3-cesium/pull/317.

fredj:
LGTM

PR #3453, opened 2015-04-01
Consider multi in add/remove/toggle select logic

sweco-sebhar:
Currently the add/remove/toggle logic will select all overlapping features regardless of the multi option. This means, for example, that with multi set to false all overlapping features will be selected with a single click if the shift key is pressed. With this change the multi option is considered similar to the "set" logic.

The background of this change has been discussed in #3402.

PR #4854, opened 2016-02-16
Always report skipped feature hits for the original layer

### LINKED ISSUE #4822: Having 2 select interactions active at the same time

ahocevar:
In the process of getting to the bottom of #4822, it occurred to me why so many iterations of patching the select interaction fixed one issue, only to create another: hit detection of features should return the same features and layers, no matter be features selected or not.

I think this fix makes the affected code easier to read and understand, and turns the select interaction into one that could almost be loved.

Fixes #4822.

### DIFF
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -177,7 +177,7 @@ ol.interaction.Select = function(opt_options) {
        */
       layerFilter = function(layer) {
         goog.asserts.assertFunction(options.layers);
-        return layer === featureOverlay || options.layers(layer);
+        return options.layers(layer);
       };
     } else {
       var layers = options.layers;
@@ -186,7 +186,7 @@ ol.interaction.Select = function(opt_options) {
        * @return {boolean} Include.
        */
       layerFilter = function(layer) {
-        return layer === featureOverlay || ol.array.includes(layers, layer);
+        return ol.array.includes(layers, layer);
       };
     }
   } else {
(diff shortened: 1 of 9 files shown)


marcjansen:
Nice branch name 😉

tschaub:
> Nice branch name 😉

Agreed.

Other things I like:

 * Net 7 lines removed!
 * And that is with something like 15 test lines added!

Things that make me scared:

 * This code always breaks.
 * This changes a test in a way that makes it seem like applications would break, but I'm so scared of the select interaction that I don't know if this is expected behavior or not.

In summary:

 * Andreas is awesome for his perseverance with the select interaction.
 * I hope that someone else who uses the select interaction can review this.
 * It would be really nice to have a clear understanding of what was broken before and what is fixed now.

ahocevar:
@tschaub I wrote a blog post about the whole story: http://ahocevar.net/openlayers/2015/02/17/trilogy-of-error.html. I hope it explains the whole evolution of these problems.

ahocevar:
@tschaub:
> This changes a test in a way that makes it seem like applications would break, but I'm so scared of the select interaction that I don't know if this is expected behavior or not.

There is no real need to make this change, so I changed it back to the way it was.

The key change of this pull request is that skipped features are now hit-detected the same way as if they had never been chandged on their original layer. This means that the also won't be hit-detected again when duplicated on an unmanaged layer. Which is what the Select and Modify interactions do.

ahocevar:
> It would be really nice to have a clear understanding of what was broken before and what is fixed now.

When a feature is selected, the Select interaction calls `ol.Map#skipFeature()` with the selected feature, and copies the feature to its internal unmanaged layer. Before this change, an already selected feature was reported by `ol.Map#forEachFeatureAtPixel` to come from a `null` layer. After this change, it is still reported to come from the layer that it was selected from.

For user created unmanaged layers, nothing changes.

ischas:
>  I wrote a blog post about the whole story: http://ahocevar.net/openlayers/2015/02/17/trilogy-of-error.html.
@ahocevar So if you knew that all one year before (see publishing date), why didn't you fix it earlier? ;-)

ahocevar:
Thanks @ischas - I shouldn't be writing blog posts at 2am.

ischas:
@ahocevar 404

ahocevar:
@ischas Took a few seconds to update on the server. Now it's live.

jonataswalker:
> Andreas is awesome for his perseverance with the select interaction.

No doubt. All of us are benefited at his cost.

bjornharrtell:
Nice work @ahocevar, did some testing with an application of mine that makes good use of the Select interaction here and there and it works as expected.

This also makes #3453 easier to fix so I've rebased that on this PR.

ahocevar:
Anyone brave enough to review this? @fredj or @tsauerwein maybe?

fredj:
LGTM, thanks Andreas

PR #4855, opened 2016-02-17
Use console.error instead of goog.log.error

fredj:
`goog.log.error` was only used in `src/ol/webgl/context.js` and only in debug mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
       "proj4": false
     },
     "rules": {
+      "no-console": 0,
       "no-constant-condition": 0
     }
   },
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -7,12 +7,9 @@ goog.provide('ol.MapProperty');
 
 goog.require('goog.asserts');
 goog.require('goog.async.nextTick');
-goog.require('goog.debug.Console');
 goog.require('goog.dom');
 goog.require('goog.dom.classlist');
 goog.require('goog.functions');
-goog.require('goog.log');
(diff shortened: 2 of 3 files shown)


ahocevar:
Looking at the changed lines, it seems that `goog.asserts` could be used as well. Any reason why that was not used in the first place?

Does the compiler strip out `console.*` calls in compiled mode?

fredj:
> Looking at the changed lines, it seems that `goog.asserts` could be used as well. Any reason why that was not used in the first place?

you mean `console.assert`? It was probably not used in the first place because the console functions were not available for all the supported browsers. It's [not the case anymore](http://caniuse.com/#search=console)

> Does the compiler strip out `console.*` calls in compiled mode?

I don't think so but it needs to be tested

ahocevar:
> you mean `console.assert`?

No, I actually mean `goog.asserts.assert`, like we use everywhere else.

fredj:
> > Does the compiler strip out `console.*` calls in compiled mode?

> I don't think so but it needs to be tested

The `console.*` are not striped out

ahocevar:
`console.assert` would be a future replacement option for `goog.asserts.assert`, but only if stripped out in the compiled library.

fredj:
The closure compiler is able to [strip code](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/StripCode.java) but this feature is not available in the CLI

fredj:
#5082

PR #4863, opened 2016-02-18
Remove use of goog.dom.createElement

fredj:
leftover from #4474


marcjansen:
Looks good to me, Fred. Please merge and thanks.

PR #4864, opened 2016-02-18
Use querySelectorAll instead of goog.dom.getElementsByClass

fredj:


marcjansen:
:+1: Thanks, Fred.

Please merge.

fredj:
Thanks for the review

PR #4866, opened 2016-02-18
Use requestAnimationFrame polyfill (for IE9)

### LINKED ISSUE #4865: Examples failed to load on IE 9
Just open http://openlayers.org/en/master/examples/animation.html on IE9

Js Error: 
```
SCRIPT438: L’objet ne gère pas la propriété ou la méthode « requestAnimationFrame »
```

It seems the problem comes from [here](https://github.com/openlayers/ol3/blob/master/src/ol/map.js#L1233)

ahocevar:
IE9 does not have native support for requestAnimationFrame.

Fixes #4865.

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -11,7 +11,7 @@
     <link rel="stylesheet" href="./resources/layout.css" type="text/css">
     {{{ extraHead.local }}}
     {{{ css.tag }}}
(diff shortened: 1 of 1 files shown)


fredj:
+1

marcjansen:
:+1:

PR #4869, opened 2016-02-18
Improve precision of ol.reproj.render

petrsloup:
There are sometimes visible gaps inside reprojected raster tiles (see https://github.com/openlayers/ol3/issues/4681#issuecomment-185331288).

It turns out to be a problem with precision of operations inside canvas 2d context and this issue can be fixed by doing the scaling multiplication in javascript instead of the canvas.


ahocevar:
Nice! Avoiding these canvas transforms should also improve performance. Please merge.

PR #4874, opened 2016-02-19
Assert we have a feature id

bartvde:
GeoServer does not handle an Update WFS-T without a filter well (it will try and update all features), so we should assert that our features have ids.


ahocevar:
Looks good, please merge when Travis is happy.

PR #4876, opened 2016-02-19
Add `target` property to Drag&Drop interaction

tsauerwein:
This PR proposes to add a `target` property to the drag&drop interaction. Currently the drop-zone of the interaction is always the viewport element, which makes it a bit difficult to style the drop-zone The proposed `target` property would make the interaction more flexible. E.g. the drop-zone could be outside the map, the whole document, ...

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2372,7 +2372,8 @@ olx.interaction.DoubleClickZoomOptions.prototype.delta;
 
 /**
  * @typedef {{formatConstructors: (Array.<function(new: ol.format.Feature)>|undefined),
- *     projection: ol.proj.ProjectionLike}}
+ *     projection: ol.proj.ProjectionLike,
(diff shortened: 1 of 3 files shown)


ahocevar:
I like it, and it looks good. Please merge.

externs/olx.js
```diff
@@ -2372,7 +2372,7 @@ olx.interaction.DoubleClickZoomOptions.prototype.delta;
 
 /**
  * @typedef {{formatConstructors: (Array.<function(new: ol.format.Feature)>|undefined),
- *     projection: ol.proj.ProjectionLike}}
+ *     projection: ol.proj.ProjectionLike, target: (Element|undefined)}}
```
fredj: should be in a new line

fredj:
+1

tsauerwein:
Thanks for the reviews!

PR #4884, opened 2016-02-23
Remove use of goog.math.isFiniteNumber()

fredj:
This replaces `goog.math.isFiniteNumber()` with `isFinite()`.

### DIFF
--- a/src/ol/reproj/reproj.js
+++ b/src/ol/reproj/reproj.js
@@ -2,7 +2,6 @@ goog.provide('ol.reproj');
 
 goog.require('goog.labs.userAgent.browser');
 goog.require('goog.labs.userAgent.platform');
-goog.require('goog.math');
 goog.require('ol.dom');
 goog.require('ol.extent');
 goog.require('ol.math');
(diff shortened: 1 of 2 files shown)


marcjansen:
The input is in both case guarateed a number and never a string;  this is good to go IMO.

Otherwise `Number.isFinite` would have probably been my choice. But this would need a polyfill...

Long story short: LGTM please merge.

fredj:
You're right `Number.isFinite` would have been a better choice ...

Thanks for the review

petrsloup:
I'm not sure about this -- I think I've seen some cases where `ol.proj#getPointResolution(resolution, point)` returns `NaN` (possibly for point outside of projection extent?) so the `compensationFactor` and `sourceResolution` can be `NaN`.
The function can be defined by user (`ol.proj.#setGetPointResolution`) so we cannot be sure it returns valid number (but if returning `NaN` is considered developer error, it should probably be documented there, right?)

fredj:
`isFinite(NaN)` returns `false` so it should be correct no ?

petrsloup:
Oh, you are right.. in that case disregard my previous comment.
I fail to see the importance of `isNaN` in `goog.math.isFiniteNumber` implementation then:
```javascript
/**
 * Returns whether the supplied number is finite and not NaN.
 * @param {number} num The number to test.
 * @return {boolean} Whether {@code num} is a finite number.
 */
goog.math.isFiniteNumber = function(num) {
  return isFinite(num) && !isNaN(num);
};
```
Probably some ancient browsers behaved differently.

However, LGTM.

fredj:
Yes, it's probably a workaround for old browsers

fredj:
See https://github.com/google/closure-library/issues/661

PR #4885, opened 2016-02-23
Mark overlayContainer and overlayContainerStopEvent as non-nullable

fredj:
And remove an unnecessary cast


ahocevar:
Looks good, please merge.

PR #4887, opened 2016-02-23
Use less aggressive DOM function overrides

### LINKED ISSUE #4886: ol.net.Jsonp tests fail when run in browsers
Looks like the `document.getElementsByTagName` mockup makes Mocha fail.

ahocevar:
Fixes #4886.


fredj:
LGTM

PR #4889, opened 2016-02-23
Use requestAnimation polyfill for examples and update release notes

ahocevar:
For the tests to run in IE9, they need a polyfill for `requestAnimationFrame`.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,10 @@
 
 ### v3.14.0
 
+#### Internet Explorer 9 support
+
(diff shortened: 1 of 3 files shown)


fredj:
+1
i'm wondering how long we will support IE9 ...

ahocevar:
> i'm wondering how long we will support IE9 ...

Until @bartvde, @marcjansen and @oterral say we can stop :smile:

marcjansen:
🎵🎵Don't stop me now!🎶🎶🎤

We still need it, if posible

bartvde:
I agree it's useful to have still.

ischas:
I'd suggest to keep OL IE9-compatible til the end of the extended support of Windows Vista, since IE9 is the highest available version for this OS.

PR #4891, opened 2016-02-23
Fix tests so they all pass in IE9

ahocevar:
With this change, all tests pass in IE9. All it took was to test for less items in a big array, and an inline image that also works in IE9.


bartvde:
LGTM thanks for fixing

ahocevar:
untitled

PR #4892, opened 2016-02-23
Stricter check for ImageData constructor

ahocevar:
`ol.source.Raster` tests fail in Edge - Edge reports `typeof ImageData` to be a `function`, but throws an exception when it is used as constructor.


bartvde:

PR #4894, opened 2016-02-23
Only run raster operations after image sources have loaded

### LINKED ISSUE #4893: ol.source.Raster tests are failing
`git bisect` output:

98b823c5fcd2a9f9971b26490226f8f1da53ed2e is the first bad commit
```
commit 98b823c5fcd2a9f9971b26490226f8f1da53ed2e
Author: Andreas Hocevar <andreas.hocevar@gmail.com>
Date:   Fri Dec 25 22:12:01 2015 +0100

    Render tiles directly to the map canvas
    
    This change removes the overhead of rendering tiles to an intermediate canvas.

:040000 040000 6b33db809340fb908090ce34f8d18ae2711503f5 acc6ce3be70a8512d293f0733714dc4bb8d16941 M	src
```

tschaub:
The raster tests were not properly cleaning up after themselves.  Listeners for `afteroperations` were getting called a second time after we had called `done()` in the tests.  This made it look like the wrong tests were failing.

This cleans up the raster tests by calling `obj.dispose()` on reused test objects and only listening for events once where we call `done()` in the listener.

There may still be an issue introduced in 98b823c5fcd2a9f9971b26490226f8f1da53ed2e with operations being run more often than expected.  But we should create a specific test that demonstrates the issue if one can be found.  The existing test failures were the result of sloppy test running.

**update** - this should now fix #4893.

### DIFF
--- a/src/ol/renderer/canvas/canvasimagelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvasimagelayerrenderer.js
@@ -202,5 +202,5 @@ ol.renderer.canvas.ImageLayer.prototype.prepareFrame = function(frameState, laye
     this.updateLogos(frameState, imageSource);
   }
 
-  return true;
+  return !!this.image_;
 };
--- a/src/ol/source/rastersource.js
+++ b/src/ol/source/rastersource.js
@@ -343,7 +343,9 @@ ol.source.Raster.prototype.onWorkerComplete_ = function(frameState, callback, er
  * @private
  */
 ol.source.Raster.getImageData_ = function(renderer, frameState, layerState) {
-  renderer.prepareFrame(frameState, layerState);
+  if (!renderer.prepareFrame(frameState, layerState)) {
+    return null;
+  }
   var width = frameState.size[0];
   var height = frameState.size[1];
   if (!ol.source.Raster.context_) {
--- a/test/spec/ol/source/rastersource.test.js
+++ b/test/spec/ol/source/rastersource.test.js
@@ -112,7 +112,7 @@ maybeDescribe('ol.source.Raster', function() {
         }
       });
 
-      source.on('afteroperations', function() {
+      source.once('afteroperations', function() {
         expect(log.length).to.equal(4);
         var inputs = log[0];
         var pixel = inputs[0];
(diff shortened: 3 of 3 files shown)


tschaub:
In this branch, tests fail before 24b7b278368c80cb197c361512731d1800f493ef and pass after it.  So there was really only one change needed.  But the other changes made the problem clearer.

ahocevar:
This looks much clearer now. Thanks @tschaub, and please merge when Travis is happy.

tschaub:
Well, Chrome is still not happy.  I lied about 24b7b278368c80cb197c361512731d1800f493ef making all tests pass.  The first `setOperation()` test is still failing.  This looks legit and could be the root of the problem.

ahocevar:
@tschaub, 9983586e7a5d7311321027ec090c6893f8b52825 fixes the issue. I haven't tested yet whether this has unwanted side effects, but the problem is that previously the renderer's `getImage()` method was called, which returned null if the source was not ready. Now, `composeFrame()` is called, and we do that without checking the return value from `prepareFrame()` to see whether there is something to render at all.

tschaub:
Thanks for the underlying fix @ahocevar.  To summarize for others:

 * prior to 98b823c5fcd2a9f9971b26490226f8f1da53ed2e, the raster source would not run pixel operations before any data had been rendered (see https://github.com/openlayers/ol3/blob/557372203f25c1502a16d52b7e1000eb7e5b5265/src/ol/source/rastersource.js#L347-L355).
 * the changes in 98b823c5fcd2a9f9971b26490226f8f1da53ed2e improved things by not rendering to an intermediate canvas, but there was no way for image layers to indicate that no work needed to be done before an image had loaded
 * the failing tests noted in #4893 were misleading because the tests were not cleaning up after themselves, but the underlying issue is that now raster operations were run once before an image loaded and once after
 * the first few commits here cleaned up the tests and made it clearer which one was a legitimate failure
 * f2d5d572abae7c59f9e4e8450ef4f71592c9f67d makes it so the renderer knows there is no more work to be done if an image is not loaded (fixing the real issue)

So, it is nice that we have a test that actually covers the problem.  But it is a bummer that these tests are not run as part of our CI setup.  Travis/Sauce integration can help - though it is a bit involved to get it working for 3rd party pull requests.

PR #4896, opened 2016-02-23
Ignore XML sequence when comparing GML

### LINKED ISSUE #4848: Units tests failure
A test is failing but only when loaded in a browser: http://localhost:3000/test/

ol.format.GML3 / when parsing TOPP states GML / writes back features as GML 
```
TypeError: Cannot read property 'replace' of null
    at assertElementNodesEqual (test-extensions.js:156:32)
    at assertElementNodesEqual (test-extensions.js:267:11)
    at assertElementNodesEqual (test-extensions.js:267:11)
    at assertElementNodesEqual (test-extensions.js:267:11)
    at Assertion.expect.Assertion.xmleql (test-extensions.js:289:5)
    at Context.<anonymous> (loader.js/home/fredj/public_html/ol3/test/spec/ol/format/gmlformat.test.js:1037:29)


```

ahocevar:
It may not be a good idea to do this - see https://github.com/openlayers/ol3/issues/4848#issuecomment-187940197. However, I think there is no practical use in enforcing the property sequence of a FeatureType. @bartvde, please correct me if I'm wrong, but I do not remember a single case where a WFS-T implementation complained on a WFS Insert because of a violated property sequence.

Fixes #4848.

### DIFF
--- a/test/spec/ol/format/gmlformat.test.js
+++ b/test/spec/ol/format/gmlformat.test.js
@@ -1034,7 +1034,7 @@ describe('ol.format.GML3', function() {
 
     it('writes back features as GML', function() {
       var serialized = gmlFormat.writeFeaturesNode(features);
-      expect(serialized).to.xmleql(ol.xml.parse(text));
+      expect(serialized).to.xmleql(ol.xml.parse(text), {ignoreElementOrder: true});
     });
 
   });
--- a/test/test-extensions.js
+++ b/test/test-extensions.js
@@ -125,6 +125,11 @@
           }
         }
       }
+      if (options && options.ignoreElementOrder) {
+        nodes.sort(function(a, b) {
+          return a.nodeName > b.nodeName ? 1 : a.nodeName < b.nodeName ? -1 : 0;
+        });
+      }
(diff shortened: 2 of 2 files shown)


fredj:
+1, thanks

bartvde:
@ahocevar are you talking about the sequence of ```<wfs:Property>``` in transactions? That should not matter at all.

bartvde:
ah maybe on insert, but I don't think know of a WFS out there enforcing the sequence

ahocevar:
I just threw an Insert at GeoServer, with completely arbitrary order of properties, and it worked just fine. so you agree that it should be safe to merge this @bartvde?

bartvde:
+1 @ahocevar

drnextgis:
@ahocevar, [TinyOWS](http://www.mapserver.org/tinyows/) checks property sequence. I've faced with this during working from QGIS. Here is my [report](https://hub.qgis.org/issues/14420).

ahocevar:
@drnextgis, it would be more convenient for everyone if TinyOWS accepted properties in any order.

PR #4899, opened 2016-02-24
Use ol.format.GMLBase#srsName instead of undeclared srsName_

fredj:
The variable was changed from private to protected in 8182396


bartvde:
Nice catch LGTM

PR #4900, opened 2016-02-24
Make ol.format.XML a @struct

fredj:
And replace the undeclared `this.version` by a local variable.


ahocevar:
Looks good, please merge.

PR #4905, opened 2016-02-24
Remove use of goog.style.setPosition

fredj:
Use element.style.left and element.style.top instead


ahocevar:
+1, please merge.

fredj:
Thanks for the review

PR #4908, opened 2016-02-24
Remove use of goog.math.Rect

### LINKED ISSUE #4906: Use of goog.math.Rect in ol.control.ZoomSlider
[`ol.control.ZoomSlider#limits_`](https://github.com/openlayers/ol3/blob/v3.14.0/src/ol/control/zoomslidercontrol.js#L65-L68) is a `goog.math.Rect` but [only the width and height](https://github.com/openlayers/ol3/blob/v3.14.0/src/ol/control/zoomslidercontrol.js#L180-L187) are initialized; left and top are always 0.

I think that `goog.math.Rect` was needed by [`goog.fx.Dragger`](https://github.com/openlayers/ol3/commit/7e27ba1d420fe7515c4bddfcd6698189edd01291) but it should be simplified

fredj:
fixes #4906

### DIFF
--- a/src/ol/control/zoomslidercontrol.js
+++ b/src/ol/control/zoomslidercontrol.js
@@ -4,7 +4,6 @@ goog.provide('ol.control.ZoomSlider');
 
 goog.require('goog.asserts');
 goog.require('goog.dom');
-goog.require('goog.math.Rect');
 goog.require('goog.style');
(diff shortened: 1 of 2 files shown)


tschaub:
Looks good.  Please merge if the control still works.

PR #4909, opened 2016-02-24
Add ol.renderer.dom.VectorLayer#clearFrame

### LINKED ISSUE #4907: Layer groups with vector layers can not be hidden in dom renderer
Hey all,
I noticed that when you set the map renderer to 'dom', layer groups with vector layers cannot be hidden. We have switched from canvas to dom renderer, as there were problems with the canvas renderer as well (this is another problem). Additionally we are also using the ol3-layerswitcher plugin. Unselecting the layer group holding the vector layers in the layer switcher seems to disable the recalculation of the position for that layer, but does not hide it from the map.

Here's an example you can play around with. Code is taken from the openlayers website as well as the ol3-layerswitcher github page's basic example.
http://codepen.io/chrgl86/pen/BjXKrK

fredj:
Fixes #4907

### DIFF
--- a/src/ol/renderer/dom/domvectorlayerrenderer.js
+++ b/src/ol/renderer/dom/domvectorlayerrenderer.js
@@ -93,6 +93,16 @@ goog.inherits(ol.renderer.dom.VectorLayer, ol.renderer.dom.Layer);
 /**
  * @inheritDoc
  */
+ol.renderer.dom.VectorLayer.prototype.clearFrame = function() {
+  this.context_.canvas.width = 0;
+  this.context_.canvas.height = 0;
+  this.renderedRevision_ = 0;
+};
+
+
+/**
+ * @inheritDoc
+ */
 ol.renderer.dom.VectorLayer.prototype.composeFrame = function(frameState, layerState) {
 
   var vectorLayer = /** @type {ol.layer.Vector} */ (this.getLayer());


fredj:
Thanks for any review

ahocevar:
Looks great, please merge.

ahocevar:
Oh wait, I think it is not allowed to set the with and height of a canvas to 0.

ahocevar:
Please disregard my previous comment. Everything looks fine. I just tried
```js
canvas.width = 0;
canvas.height = 0;
var context = canvas.getContext('2d');
```
All without error. So I think this is good to merge.

fredj:
Thanks for the review

ahocevar:
From the [spec](https://www.w3.org/TR/html5/scripting-1.html#attr-canvas-width):
> The canvas element has two attributes to control the size of the element's bitmap: width and height. These attributes, when specified, must have values that are valid non-negative integers.

src/ol/renderer/dom/domvectorlayerrenderer.js
```diff
@@ -93,6 +93,16 @@ goog.inherits(ol.renderer.dom.VectorLayer, ol.renderer.dom.Layer);
 /**
  * @inheritDoc
  */
+ol.renderer.dom.VectorLayer.prototype.clearFrame = function() {
+  this.context_.canvas.width = 0;
+  this.context_.canvas.height = 0;
```
tschaub: It used to be that the `width` (or `height`) setting trick was consistently faster.  It looks like this is not consistent any more (http://jsperf.com/canvasclear/30).

I don't think it is worth changing, but I think the following would be more straightforward:
```js
var canvas = this.context_.canvas;
this.context_.clearRect(0, 0, canvas.width, canvas.height);
```

This would avoid any (potential future?) issues with setting height/width to 0.  If we do want to use the `width` (or `height`) trick, it would be enough to do this:

```js
var canvas = this.context_.canvas;
canvas.width = canvas.width;
```

fredj: You're right; we always do
```javascript
var canvas = this.context_.canvas;
canvas.width = canvas.width;
```
to clear the canvas, let's do the same here.

Issue #4913, opened 2016-02-25
Export ol.proj.equivalent


sebasbaumh:
I would like to use the `ol.proj.equivalent` function to compare two projections.
The projections could be either created by constructing them directly or using `ol.proj.get`.
The `ol.proj.equivalent` function is perfectly suited for comparing both projections and is also used internally for that. In contrast to e.g. `ol.extent.equals` it is not exposed publicly.

So I will create a small pull request to just expose this function after adding this issue.
There are already tests for the function itself, e.g. in https://github.com/openlayers/ol3/blob/master/test/spec/ol/proj/proj.test.js#L16.

Related issues:
@twpayne is talking about exporting this method in #935

PR #4914, opened 2016-02-25
Export ol.proj.equivalent

sebasbaumh:
Export `ol.proj.equivalent`

The function `ol.proj.equivalent` is used for comparing two projections,
but not exposed publicly. It should be accessible using a public API to
be able to compare projections in the same way as OpenLayers does.

A test for the `ol.proj.equivalent` function itself already exists in
https://github.com/openlayers/ol3/blob/master/test/spec/ol/proj/proj.test.js

github issue #4913


fredj:
Fixes #4913

fredj:
LGTM

PR #4915, opened 2016-02-25
Use {} instead of Object

fredj:
We probably have these to trick the compiler but it's not needed anymore


bartvde:
LGTM

fredj:
Thanks for the review

PR #4916, opened 2016-02-25
Fix ol.format.GeoJSON#writeFeatureObject return type

fredj:
`GeoJSONFeature` instead of `GeoJSONObject`


bartvde:

fredj:
Thanks for the review

PR #4917, opened 2016-02-25
Make ol.source.Cluster more flexible by adding a geometryFunction option

### LINKED ISSUE #3167: Clustering fails when non-point geometries are included in layer
Suggest modifying clustering algorithm to simply ignore non-point features instead of producing an error. This would allow a mixed layer to be viewed with clustered points while lines and polygons would render normally.

### LINKED ISSUE #4814: ol.source.Cluster worked with ol.geom.MultiPoint
Feasible or not?

ahocevar:
Currently, `ol.source.Cluster` only works with sources that contain point features. With this change, applications get full flexibility of what and how to cluster. If you wanted to cluster points and polygons (using an interior point), but exclude linestrings, you could use the following configuration:
```js
  var source = new ol.source.Cluster({
    geometryFunction: function(feature) {
      var geom = feature.getGeometry();
      if (geom.getType() == 'Point') {
        return geom;
      } else if (geom.getType() == 'Polygon') {
        return geom.getInteriorPoint();
      }
      return null;
    },
    source: new ol.source.Vector({/*...*/})
  });
```

Fixes #3167.
Fixes #4814.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3984,6 +3984,7 @@ olx.source.BingMapsOptions.prototype.wrapX;
  *     distance: (number|undefined),
  *     extent: (ol.Extent|undefined),
  *     format: (ol.format.Feature|undefined),
+ *     geometryFunction: (undefined|function(ol.Feature):ol.geom.Point),
  *     logo: (string|undefined),
  *     projection: ol.proj.ProjectionLike,
  *     source: ol.source.Vector,
@@ -4015,6 +4016,25 @@ olx.source.ClusterOptions.prototype.distance;
  * @api
  */
 olx.source.ClusterOptions.prototype.extent;
+
+
+/**
+ * Function that takes an {@link ol.Feature} as argument and returns an
+ * {@link ol.geom.Point} as cluster calculation point for the feature. When a
+ * feature should not be considered for clustering, the function should return
+ * `null`. The default, which works when the underyling source contains point
+ * features only, is
+ * ```js
+ * function(feature) {
+ *   return feature.getGeometry();
+ * }
+ * ```
+ * See {@link ol.geom.Polygon#getInteriorPoint} for a way to get a cluster
+ * calculation point for polygons.
+ * @type {undefined|function(ol.Feature):ol.geom.Point}
+ * @api
+ */
+olx.source.ClusterOptions.prototype.geometryFunction;
 
 
 /**
@@ -6174,8 +6194,8 @@ olx.style.FillOptions;
 
 
 /**
- * A color, gradient or pattern. See {@link ol.color} 
- * and {@link ol.colorlike} for possible formats. Default null; 
+ * A color, gradient or pattern. See {@link ol.color}
+ * and {@link ol.colorlike} for possible formats. Default null;
  * if null, the Canvas/renderer default black will be used.
  * @type {ol.Color|ol.ColorLike|undefined}
  * @api
--- a/src/ol/source/clustersource.js
+++ b/src/ol/source/clustersource.js
@@ -14,7 +14,9 @@ goog.require('ol.source.Vector');
 
 /**
  * @classdesc
- * Layer source to cluster vector data.
+ * Layer source to cluster vector data. Works out of the box with point
+ * geometries. For other geometry types, or if not all geometries should be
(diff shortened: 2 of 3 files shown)


fredj:
Fixes #4814 ?

ahocevar:
Yes, thanks @fredj.

fredj:
+1

syntax42:
Nice feature! Do i understand it correctly, that when the geometryFunction returns null, the feature should not be considered for clustering, but still be drawn on the layer? I ask because because in my implementation, the feature is not displayed, when the function returns null.

ahocevar:
> Do i understand it correctly, that when the geometryFunction returns null, the feature should not be considered for clustering, but still be drawn on the layer?

No, features won't be drawn when the `geometryFunction` returns `null`. If you want that, you need to create a 2nd layer (`ol.layer.Vector`) with the same `ol.source.Vector` as the raster source and a style function that contains logic like
```js
styleFunction: function(feature, resolution) {
  if (!geometryFunction(feature)) {
    return myStyle;
  }
}
```

syntax42:
OK! That was my fall back solution as well. Thanks.

icemagno:
geom.getInteriorPoint() is returning `geom.getInteriorPoint is not a function` for `MultiPolygon`

ahocevar:
It is `getInteriorPoints()` for MultiPolygons.

evandrododo:
Trying to cluster LineString with `geom.getLastCoordinate()` don't return an error, but ignores LineStrings.

My debug code was:
```
geometryFunction: function(feature) {
                var geom = feature.getGeometry();
                if (geom.getType() == 'Point') {
                  console.log('Point',geom);
                  return geom;
                } else if (geom.getType() == 'Polygon') {
                  console.log('Polygon', geom.getInteriorPoint());
                  return geom.getInteriorPoint();
                } else if (geom.getType() == 'LineString') {
                  console.log('LineString', geom.getLastCoordinate());
                  return null;
                }
                return null;
              }
```
but they seems to be different objects:
captura de tela de 2017-04-27 17 25 12
Is this right?

evandrododo:
Noticed that the return from `geom.getInteriorPoint()` {ol.geom.Point} is different from `geom.getLastCoordinate` {ol.Coordinate}
Solved with
```javascript
...
    } else if (geom.getType() == 'LineString') {
        let linePoint = new ol.geom.Point(geom.getLastCoordinate());
        console.log('LineString', linePoint);
        return linePoint;
    }
```

it works like a charm

Skuriles:
Hm.. 
maybe I did something wrong, but with 

 ```
 var clusterSource = new ol.source.Cluster({
                        geometryFunction: function (feature) {
                            return new ol.geom.Point(ol.extent.getCenter(feature.getGeometry().getExtent()));
                        },
                        distance: 1,
                        source: source
                    });
```
and source:
```
var feature = new ol.Feature({
                        geometry: new ol.geom.MultiPoint(data)
                    });

                    var source = new ol.source.Vector({
                        features: [feature]
                    });
```
while data is an array of coords (around 15.000)
My problem: All points are clustered to only one point
How can I achieve that clustering notes distance properties like the default clustering with point geometries?
As it is now, this is not usable. 
Problem is: If I use one feature per point instead of multiPoint clustering works well, but memory usage explodes (factor 10). Using multiple layers with single Point features (20.000 points)  with around kills the browser, especially on mobile devices

Any ideas to solve this problem?

Gokulraj23:
Hi,
Trying to cluster LineString nothing is happening. it's not throwing any error also.

vm.clusterArc = new ol.source.Cluster({
                source: vm.arcSource,
                distance: parseInt(40, 10),
                geometryFunction: function(feature) {
                    var geom = feature.getGeometry();
                    if(geom.getType() == 'LineString')
                    {
                    var linePoint = new ol.geom.Point(geom.getLastCoordinate());   
                    console.log('LineString', linePoint);
                    return linePoint;
                    }
                }
            }); 
            
            var flightsLayer = new ol.layer.Vector({
                source: vm.clusterArc,
                style: function(feature) {
                     if (feature.get('features')[0].get('finished')) {
                        return [defaultStyle];
                        } else {
                        return [];
                    }
                }
            });

Gokulraj23:
Hi,
**There is any possibility to cluster flight animation using line string**.

PR #4918, opened 2016-02-25
Don't change the canvas size to clear it

fredj:
See https://github.com/openlayers/ol3/pull/4909#discussion_r54098230

### DIFF
--- a/src/ol/renderer/dom/domvectorlayerrenderer.js
+++ b/src/ol/renderer/dom/domvectorlayerrenderer.js
@@ -94,8 +94,9 @@ goog.inherits(ol.renderer.dom.VectorLayer, ol.renderer.dom.Layer);
  * @inheritDoc
  */
 ol.renderer.dom.VectorLayer.prototype.clearFrame = function() {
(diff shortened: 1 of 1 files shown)


src/ol/renderer/dom/domvectorlayerrenderer.js
```diff
@@ -94,8 +94,9 @@ goog.inherits(ol.renderer.dom.VectorLayer, ol.renderer.dom.Layer);
  * @inheritDoc
  */
 ol.renderer.dom.VectorLayer.prototype.clearFrame = function() {
-  this.context_.canvas.width = 0;
-  this.context_.canvas.height = 0;
+  var canvas = this.context_.canvas;
+  canvas.width = canvas.width;
+  canvas.height = canvas.height;
```
tschaub: Setting only `width` should be enough to do the trick, right?

fredj: Yes, fixed

ahocevar:
Looks good to me, please merge.

PR #4921, opened 2016-02-25
Unlisten before calling listener and bind to target by default

### LINKED ISSUE #4919: Layer on change event issue when migrating from 3.10.1 to 3.14.0
I was using the following code to access features on a layer once they are loaded. It no longer works, and I get the following error: `Uncaught TypeError: Cannot read property 'getSource' of undefined`. Here is the code that worked under 3.10.1:

    // Wait until all features are loaded, then process features
    myLayer.once( "change", function ( evt ) {
        this.getSource().forEachFeature( function ( feature ) {
            // set feature id's etc...
        })
    })

`this` is undefined. Is there another/better way to do this now?

ahocevar:
The problem reported in #4919 is a regression caused by the `goog.events.*` removal. Listeners were always called synchronously, but the difference with the new implementation is that a `once` listener is unregistered *after* calling the listener, whereas it was unregistered *before* previously. The second problem is that when `opt_this` is not provided, we no longer bind to the event target by default.

Fixes #4919.

### DIFF
--- a/src/ol/events.js
+++ b/src/ol/events.js
@@ -107,11 +107,12 @@ ol.events.ListenerObjType;
  */
 ol.events.bindListener_ = function(listenerObj) {
   var boundListener = function(evt) {
-    var rv = listenerObj.listener.call(listenerObj.bindTo, evt);
+    var listener = listenerObj.listener;
+    var bindTo = listenerObj.bindTo || listenerObj.target;
     if (listenerObj.callOnce) {
       ol.events.unlistenByKey(listenerObj);
     }
-    return rv;
+    return listener.call(bindTo, evt);
   }
   listenerObj.boundListener = boundListener;
   return boundListener;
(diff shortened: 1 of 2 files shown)


examples/vector-layer.js
```diff
@@ -41,6 +41,12 @@ var vectorLayer = new ol.layer.Vector({
     return style;
   }
 });
+var id = 0;
+vectorLayer.once('change', function() {
+  this.getSource().forEachFeature(function(feature) {
+    feature.setId(id++);
+  });
+});
```
tschaub: Was this meant to be a test instead? :)

ahocevar: Yes, I'll be pushing an update shortly.

ahocevar:
Tests added.

tschaub:
Nice fix @ahocevar.  Please merge.

ahocevar:
Thanks for the quick review @tschaub.

PR #4923, opened 2016-02-26
Update custom-builds.md

xlhomme:
not sure we should add 
"ol.layer.Heatmap#setRadius",
"ol.layer.Heatmap#setBlur"
but it's work.


fredj:
Thanks

PR #4926, opened 2016-02-26
Add CartoDB tile source.

aisaacs:
This is a work in progress. It adds a carto db source that supports both named and anonymous maps. 

* http://docs.cartodb.com/cartodb-platform/maps-api.html#anonymous-maps
* http://docs.cartodb.com/cartodb-platform/maps-api.html#named-maps

I've been sitting on this for a while, which is why it still uses `goog.net`. I plan on removing that and replacing it with straight `XMLHttpRequest`.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -6209,6 +6209,104 @@ olx.source.XYZOptions.prototype.urls;
  */
 olx.source.XYZOptions.prototype.wrapX;
 
+/**
+ * @typedef {{attributions: (Array.<ol.Attribution>|undefined),
+ *     crossOrigin: (null|string|undefined),
+ *     logo: (string|olx.LogoOptions|undefined),
+ *     projection: ol.proj.ProjectionLike,
+ *     maxZoom: (number|undefined),
+ *     minZoom: (number|undefined),
+ *     wrapX: (boolean|undefined),
+ *     config: (Object|undefined),
+ *     map: string,
+ *     account: (string|undefined)}}
+ * @api
+ */
+olx.source.CartoDBOptions;
+
+
+/**
+ * Attributions.
+ * @type {Array.<ol.Attribution>|undefined}
+ * @api stable
+ */
+olx.source.CartoDBOptions.prototype.attributions;
+
+
+/**
+ * The `crossOrigin` attribute for loaded images.  Note that you must provide a
+ * `crossOrigin` value if you are using the WebGL renderer or if you want to
+ * access pixel data with the Canvas renderer.  See
+ * {@link https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image}
+ * for more detail.
+ * @type {null|string|undefined}
+ * @api stable
+ */
+olx.source.CartoDBOptions.prototype.crossOrigin;
+
+
+/**
+ * Logo.
+ * @type {string|olx.LogoOptions|undefined}
+ * @api stable
+ */
+olx.source.CartoDBOptions.prototype.logo;
+
+
+/**
+ * Projection. Default is `EPSG:3857`.
+ * @type {ol.proj.ProjectionLike}
+ * @api
+ */
+olx.source.CartoDBOptions.prototype.projection;
+
+
+/**
+ * Optional max zoom level. Default is `18`.
+ * @type {number|undefined}
+ * @api
+ */
+olx.source.CartoDBOptions.prototype.maxZoom;
+
+
+/**
+ * Whether to wrap the world horizontally. Default is `true`.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.source.CartoDBOptions.prototype.wrapX;
+
+
+/**
+ * If using anonymous maps, the CartoDB config to use. See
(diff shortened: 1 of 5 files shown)


externs/olx.js
```diff
@@ -6025,6 +6025,104 @@ olx.source.XYZOptions.prototype.urls;
...
+/**
+ * @typedef {{attributions: (Array.<ol.Attribution>|undefined),
+ *     crossOrigin: (null|string|undefined),
+ *     logo: (string|olx.LogoOptions|undefined),
+ *     projection: ol.proj.ProjectionLike,
+ *     maxZoom: (number|undefined),
+ *     minZoom: (number|undefined),
+ *     wrapX: (boolean|undefined),
+ *     config: (Object|undefined),
+ *     map: string,
```
tschaub: It [looks like](https://travis-ci.org/openlayers/ol3/builds/112149102) maybe this should be `(string|undefined)` based on your example use.

externs/olx.js
```diff
@@ -6025,6 +6025,104 @@ olx.source.XYZOptions.prototype.urls;
...
+ * @api
+ */
+olx.source.CartoDBOptions.prototype.config;
+
+
+/**
+ * If using named maps, this will be the name of the template to load.
+ * See {@link http://docs.cartodb.com/cartodb-platform/maps-api.html#named-maps}
+ * for more detail.
+ * @type {boolean|undefined}
```
gberaudo: Shouldn't it be a `string` instead of a `boolean`?

src/ol/source/cartodb.js
```diff
@@ -0,0 +1,127 @@
...
+ * Issue a request to initialize the CartoDB map.
+ * @private
+ */
+ol.source.CartoDB.prototype.initializeMap_ = function() {
+  var paramHash = JSON.stringify(this.config_);
+  if (this.templateCache_[paramHash]) {
+    this.applyTemplate_(this.templateCache_[paramHash]);
+    return;
+  }
+  var protocol = window.location.protocol;
```
gberaudo: What about always retrieving the tiles securely through `https`?

src/ol/source/cartodb.js
```diff
@@ -0,0 +1,127 @@
...
+  if (xhrIo.isSuccess()) {
+    this.applyTemplate_(data);
+  }
+  this.templateCache_[paramHash] = data;
+};
+
+
+/**
+ * Apply the new tile urls returned by carto db
+ * @param {Object} data Result of carto db call.
```
gberaudo: It would be cleaner to declare a typedef in an extern file and here require the type instead of `Object`.
In the function body, accesses to the hash content would use the dot notation.

src/ol/source/cartodb.js
ahocevar: It's good practice to use the same protocol as the application if possible. The code here is fine as it is now.

bartvde: @ahocevar but since https://github.com/openlayers/ol3/pull/3579 our practice is to use https:// where available, always?

ahocevar: Ah, thanks @bartvde. I had missed that. So yes, please change to always using `https://` like @gberaudo suggested.

src/ol/source/cartodb.js
tschaub: The problem with adding an extern file is that it requires everybody to add it to their build config in order for this to work properly.  So we break full build configs when we release new library features that require an additional externs files.  The `@typedef` could be added to our `olx.js` file, but since we suggest people who compile applications together with the library use that as a source file, this would allow these property names to be minified.

In the case of receiving data from an external service, the property names can never be renamed.  I think the clearest way to achieve this is to quote them.  Externs provide a convenient way to avoid renaming properties/functions for 3rd party libraries (and it allows us to avoid quoting the same name many times), but I think they are more of a hassle than a benefit for something like this.

gberaudo: Skipping the extern declaration would prevent closure from detecting typos and incorrect type usages.
For me, type checking is the best selling point of closure.

What about adding a unique extern file dedicated to all such cases?

tschaub: We can either specify the type right here in this file or in an externs file.  The benefit of doing it right here in this file is that nobody has to change their build config to keep it working.  I don't see any benefit of doing it in an externs file.

ahocevar:
Hey @aisaacs - I just wanted to say I like this pull request, and I'm hoping you find time to finish it.

src/ol/source/cartodb.js
```diff
@@ -0,0 +1,133 @@
+goog.provide('ol.source.CartoDB');
+
+goog.require('goog.asserts');
+goog.require('ol.source.State')
```
tschaub: Looks like a missing semicolon.  (We could make the linter warn here so we're not left with [confusing build errors](https://travis-ci.org/openlayers/ol3/builds/119613653).)

tschaub:
Looks great @aisaacs.  Thanks for sticking with it.  It will be nice to see this get some additional use.

PR #4927, opened 2016-02-27
Upgrade JSTS in example

bjornharrtell:


tschaub:
Thanks @bjornharrtell.

tschaub:
@bjornharrtell it looks like [the build is now failing](https://travis-ci.org/openlayers/ol3/builds/112260082) with this:

```
JavaScript ERROR: Unhandled promise rejection TypeError: Network request failed
TRACE:
 -> https://cdn.rawgit.com/bjornharrtell/jsts/gh-pages/1.0.1/jsts.min.js: 24
  phantomjs://code/check-example.js:20 in onError
```

Not sure why it would be failing on master but not for your branch here.  I've restarted the build, but get the same issue.

Though this error makes it look like a network failure, I'm guessing this is an unhandeled rejected promise that could occur of there are any unhandeled errors in the feature parsing, buffering, or writing.  And the trace makes it look like the error is coming from jsts.min.js (line 24).

Any ideas?

bjornharrtell:
I get the same error running `make ci` locally on master, but have no idea what is going on. The example runs fine when run in a real browser. Will try to debug further asap.

bjornharrtell:
Debugged further and it seems this is caused by ES6 Promise polyfill by core-js and the actual failure that is output seems to come from https://github.com/zloirock/core-js/blob/v2.1.1/modules/es6.promise.js#L113.

Funny that bringing in Polyfills to make the code run in older browsers make it fail in Phantom.

Core-js does seem to be overly complex to say the least and why the Promise is even involved I'm not sure. I'll try to find something that is simplier and works with Phantom.

tschaub:
The example uses `fetch` which implicitly depends on `Promise`.  Neither is available in Phantom 1.x - but we should confirm that the polyfill is required for Phantom 2.x.

In any case, I'm assuming the error does not come from the polyfill.  Instead, it is likely here: https://github.com/bjornharrtell/ol3/blob/94debcfc3ae164d35861bd9a9c417407340260bd/examples/jsts.js#L18-L35

bjornharrtell:
Well, I think the Promise polyfill that was in JSTS 1.0.1 interfered with the fetch. I've released JSTS 1.0.2 that only have the needed polyfills for IE 10 compatibility. Added bonus, 100 Kb smaller build. :)

PR #4930, opened 2016-02-29
Use ADVANCED instead of ADVANCED_OPTIMIZATIONS

fredj:


ahocevar:
Good catch, please merge.

PR #4936, opened 2016-02-29
Remove jquery from getfeatureinfo-layers example

fredj:


ahocevar:
Looks good, please merge.

PR #4937, opened 2016-02-29
Get rid of goog.json

fredj:
Use JSON.parse and JSON.stringify instead.


ahocevar:
Looks good, please merge.

fredj:
Thanks for the review

PR #4938, opened 2016-02-29
Get rid of goog.dom.getFirstElementChild

fredj:
Use [firstElementChild](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/firstElementChild) instead.


tschaub:
Code looks good.  Assuming examples work, please merge.

fredj:
Thanks for the review

PR #4939, opened 2016-02-29
Fix rendering of opaque layers with opacity != 1

### LINKED ISSUE #4934: Tile layer opacity issue in v3.14
In v3.14, the layer opacity behavior has changed. It appears that the opacity is no longer directly applied when the tile is loaded, causing flicker when dragging the map.

Refer to the following fiddles and drag the map around to load new tiles:
v3.13.1: http://jsfiddle.net/norric/g52zuf2x/
v3.14.0: http://jsfiddle.net/norric/h74hnrhu/

Related to https://github.com/openlayers/ol3/pull/4597 ?

ahocevar:
When a layer has an opacity set, clip rectangles have to be calculated even when the source has the opaque flag set. Otherwise lower resolution tiles will change the visual opacity of rendered tiles.

To avoid 1-pixel lines at clip extent boundaries, the clip rectangle coordinates are now calculated the same way as the tile insert points.

Fixes #4934.

### DIFF
--- a/src/ol/renderer/canvas/canvastilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvastilelayerrenderer.js
@@ -26,9 +26,9 @@ ol.renderer.canvas.TileLayer = function(tileLayer) {
 
   /**
    * @private
-   * @type {!Object.<string, Array.<ol.Extent>>}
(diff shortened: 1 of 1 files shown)


fredj:
LGTM

PR #4947, opened 2016-03-01
IE support for ol.source.Raster

tschaub:
This brings in `pixelworks@1.1.0` which has a workaround for the lack of an `ImageData` constructor in IE (add your votes here to support its addition: https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/8804656-support-for-the-imagedata-constructor).

For `image` type operations, workers can return `ImageData`-like objects (with `data`, `width`, and `height` properties).  On IE, they will be called with the same.  No changes should be required for `pixel` type operations.

### DIFF
--- a/examples/region-growing.js
+++ b/examples/region-growing.js
@@ -59,7 +59,7 @@ function growRegion(inputs, data) {
     }
     edge = newedge;
   }
-  return new ImageData(outputData, width, height);
+  return {data: outputData, width: width, height: height};
 }
 
 function next4Edges(edge) {
--- a/examples/shaded-relief.js
+++ b/examples/shaded-relief.js
@@ -99,7 +99,7 @@ function shade(inputs, data) {
     }
   }
 
-  return new ImageData(shadeData, width, height);
+  return {data: shadeData, width: width, height: height};
 }
 
 var elevation = new ol.source.XYZ({
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "metalsmith-layouts": "1.6.4",
     "nomnom": "1.8.1",
     "pbf": "1.3.5",
-    "pixelworks": "1.0.0",
+    "pixelworks": "1.1.0",
     "rbush": "1.4.2",
     "temp": "0.8.3",
     "vector-tile": "1.2.0",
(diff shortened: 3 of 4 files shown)


ahocevar:
This is awesome, please merge.

tschaub:
Thanks for the review @ahocevar.  The raster examples now work on IE 11 (I didn't test elsewhere).  And I'm happier that the raster source tests are being run in Phantom now (the [pixelworks tests](https://github.com/tschaub/pixelworks/blob/master/test/karma.conf.js) run in Chrome, Firefox, and IE 11).

fredj:
The tests are failing for IE 9 and IE 10

tschaub:
> The tests are failing for IE 9 and IE 10

Workers and typed arrays are absent for IE 9.  I don't think there's a good reason to try for a workaround there.  IE 10 is missing `Uint8ClampedArray` but could be made to work if someone was interested.  I'll put together a pull request that skips tests if `Uint8ClampedArray` is missing.

fredj:
>  I'll put together a pull request that skips tests if `Uint8ClampedArray` is missing.

perfect, thanks

tschaub:
@fredj see #4962.  I only have access to a machine with IE 11 right now.  If anyone with 10 and 9 could help test there, that would be great.

Aside: are we really saying we support IE 9?

tschaub:
> Aside: are we really saying we support IE 9?

Maybe we could say we ditch it [at 5%](https://www.netmarketshare.com/report.aspx?qprid=3&qpaf=&qpcustom=Microsoft+Internet+Explorer+9.0&qpcustomb=0&qpsp=182&qpnp=24&qptimeframe=M&qpdt=1)?

image

marcjansen:
Sad enough; we still have certain customers needing that support. If we ditched IE 9, we would be locked to that last ol3 version still supporting it... which would basically add pressure to upgrade. That said, I am not blocking to drop support. It'd be nice to know it one point release before we actually drop it.

tschaub:
Good to know @marcjansen.  Doesn't seem like we're doing a ton of extra work for IE 9.  But it also seems like it is going away fast.

bartvde:
not so fast in government agencies ;-)

fredj:
> not so fast in government agencies ;-)

same here ...

PR #4950, opened 2016-03-02
Remove use of goog.Disposable()

tschaub:
This replaces `goog.Disposable()` with a simpler `ol.Disposable()`.

### DIFF
--- a/src/ol/disposable.js
+++ b/src/ol/disposable.js
@@ -1,5 +1,7 @@
 goog.provide('ol.Disposable');
 
+goog.require('ol');
+
 /**
  * Objects that need to clean up after themselves.
  * @constructor
@@ -27,4 +29,4 @@ ol.Disposable.prototype.dispose = function() {
  * Extension point for disposable objects.
  * @protected
  */
-ol.Disposable.prototype.disposeInternal = function() {};
+ol.Disposable.prototype.disposeInternal = ol.nullFunction;
--- a/src/ol/disposable.js
+++ b/src/ol/disposable.js
@@ -1,0 +1,30 @@
+goog.provide('ol.Disposable');
+
+/**
+ * Objects that need to clean up after themselves.
+ * @constructor
+ */
+ol.Disposable = function() {};
+
+/**
+ * The object has already been disposed.
+ * @type {boolean}
+ * @private
+ */
+ol.Disposable.prototype.disposed_ = false;
+
+/**
+ * Clean up.
+ */
+ol.Disposable.prototype.dispose = function() {
+  if (!this.disposed_) {
+    this.disposed_ = true;
+    this.disposeInternal();
+  }
+};
+
+/**
+ * Extension point for disposable objects.
+ * @protected
+ */
+ol.Disposable.prototype.disposeInternal = function() {};
--- a/src/ol/control/zoomslidercontrol.js
+++ b/src/ol/control/zoomslidercontrol.js
@@ -118,14 +118,17 @@ ol.control.ZoomSlider = function(opt_options) {
       [className, ol.css.CLASS_UNSELECTABLE, ol.css.CLASS_CONTROL],
       thumbElement);
 
-  var dragger = new ol.pointer.PointerEventHandler(containerElement);
-  this.registerDisposable(dragger);
+  /**
+   * @type {ol.pointer.PointerEventHandler}
+   * @private
+   */
+  this.dragger_ = new ol.pointer.PointerEventHandler(containerElement);
 
-  ol.events.listen(dragger, ol.pointer.EventType.POINTERDOWN,
+  ol.events.listen(this.dragger_, ol.pointer.EventType.POINTERDOWN,
       this.handleDraggerStart_, this);
-  ol.events.listen(dragger, ol.pointer.EventType.POINTERMOVE,
+  ol.events.listen(this.dragger_, ol.pointer.EventType.POINTERMOVE,
       this.handleDraggerDrag_, this);
-  ol.events.listen(dragger, ol.pointer.EventType.POINTERUP,
+  ol.events.listen(this.dragger_, ol.pointer.EventType.POINTERUP,
       this.handleDraggerEnd_, this);
 
   ol.events.listen(containerElement, ol.events.EventType.CLICK,
@@ -141,6 +144,15 @@ ol.control.ZoomSlider = function(opt_options) {
   });
 };
 goog.inherits(ol.control.ZoomSlider, ol.control.Control);
+
+
+/**
+ * @inheritDoc
+ */
+ol.control.ZoomSlider.prototype.disposeInternal = function() {
+  this.dragger_.dispose();
+  goog.base(this, 'disposeInternal');
+};
 
 
 /**
--- a/src/ol/events/eventtarget.js
+++ b/src/ol/events/eventtarget.js
@@ -1,6 +1,6 @@
 goog.provide('ol.events.EventTarget');
 
-goog.require('goog.Disposable');
+goog.require('ol.Disposable');
 goog.require('ol.events');
 goog.require('ol.events.Event');
 
@@ -20,7 +20,7 @@ goog.require('ol.events.Event');
  *    returns false.
  *
  * @constructor
- * @extends {goog.Disposable}
+ * @extends {ol.Disposable}
  */
 ol.events.EventTarget = function() {
 
@@ -33,7 +33,7 @@ ol.events.EventTarget = function() {
   this.listeners_ = {};
 
 };
-goog.inherits(ol.events.EventTarget, goog.Disposable);
+goog.inherits(ol.events.EventTarget, ol.Disposable);
 
 
 /**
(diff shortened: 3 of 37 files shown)


tschaub:
Replacing the use of `goog.dispose(obj)` with `obj.dispose()` revealed that we were passing replay groups to `goog.dispose()` though they are not disposable.  In addition, I notice that there is more cleaning up that the controls should do (unregistering listeners on DOM elements) in their `disposeInternal` methods - but that can be a separate effort.

src/ol/disposable.js
```diff
@@ -0,0 +1,30 @@
+goog.provide('ol.Disposable');
+
+/**
+ * Objects that need to clean up after themselves.
+ * @constructor
+ */
+ol.Disposable = function() {};
```
gberaudo: Could it be annotated with `@struct` like `ol.Observable`?

src/ol/disposable.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * The object has already been disposed.
+ * @type {boolean}
+ * @private
+ */
+ol.Disposable.prototype.disposed_ = false;
```
gberaudo: Have you considered setting the property in the constructor?

src/ol/disposable.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Clean up.
+ */
+ol.Disposable.prototype.dispose = function() {
```
gberaudo: Should the method be marked `@api`?

ahocevar:
This looks great @tschaub. I created a heap snapshot after running all tests from master and from your branch. And your branch results in a smaller heap with less DOM entries. So I'm in favour of merging this.

Regarding @gberaudo's comments, I'd not make `ol.Disposable.prototype.dispose` exportable. No `dispose` method is exportable now either. And I think it makes sense to keep `ol.Disposable.prototype.disposed_` on the prototype, because it saves a bit of memory if the value is not set on every instance.

src/ol/disposable.js
tschaub: Done with 9b58672967fe0030304d7a93e6e08c5da525a78d.

src/ol/disposable.js
tschaub: I think this could make sense.  But we don't currently expose a `dispose` method, so I didn't do that here.

src/ol/disposable.js
tschaub: I couldn't come up with any justification for that.  But would be open to a demonstration of what it would make possible or improve.

src/ol/disposable.js
tschaub: Actually, marking as `@struct` is creating compiler issues with `EventTarget` (where it uses bracket notation).  See https://travis-ci.org/openlayers/ol3/builds/113174215 for the failures.

tschaub:
I'm not liking [this failure](https://travis-ci.org/openlayers/ol3/builds/113177618):

```
node tasks/test-rendering.js
Gecko: 40.0.3 
Script Error: [JavaScript Error: "TypeError: str is undefined" {file: "http://127.0.0.1:3001/node_modules/mocha/mocha.js" line: 5936}]
       Stack:
         -> http://127.0.0.1:3001/node_modules/mocha/mocha.js: 5936
1456939202517	addons.repository	WARN	cacheEnabled: Couldn't get pref: extensions.getAddons.cache.enabled
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

src/ol/disposable.js
gberaudo: Struct is needed for https://github.com/openlayers/ol3/pull/4793 ; I can make the change after your PR is merged and I rebase on top of it.

src/ol/disposable.js
gberaudo: I thought it would confuse the compiler and prevent using `@struct` but I just tried and it compiled.
I see no major drawback in using this optimization.

src/ol/disposable.js
gberaudo: Not sure to understand whether calling `dispose()` is exclusively the responsability of the library or if the user should do it sometimes.
For example, the `ol.Map` class inherits from `ol.Disposable`. Is the user supposed to call the `dispose()` method when he wants to get ride of a map instance?

tschaub:
Ok, the [rendering test failure](https://travis-ci.org/openlayers/ol3/builds/113177618) mentioned above was a real issue.  But pretty badly buried.

The first issue is that when our `afterEach` test functions throw, we get an obscure error:
```
Uncaught TypeError: Cannot read property 'replace' of undefined
    exports.clean @ mocha.js:5937
    (anonymous function) @ mocha.js:2518
    EventEmitter.emit @ mocha.js:200
    (anonymous function) @ mocha.js:2451
    EventEmitter.emit @ mocha.js:200
    Runner.fail @ mocha.js:4564
    Runner.failHook @ mocha.js:4593
    (anonymous function) @ mocha.js:4638
    done @ mocha.js:4251
    Runnable.run @ mocha.js:4286
    next @ mocha.js:4629
    (anonymous function) @ mocha.js:4651
    timeslice @ mocha.js:12619
```

At the top of the stack, [`clean`](https://github.com/mochajs/mocha/blob/v2.4.5/mocha.js#L5929-L5948) is getting called with `undefined`.  Digging down, [it looks like](https://github.com/mochajs/mocha/blob/v2.4.5/mocha.js#L2518) `test.body` should be defined, but it isn't.  I didn't put together a test case to see if this is limited to our rendering tests or has something to do with the test runner, but it would be good to track that down.  It means `afterEach` (and probably `beforeEach`) is a liability.

The real issue is that in the WebGL map renderer's [`disposeInternal` method](https://github.com/openlayers/ol3/blob/34d8d77344ee0b653770f065c593d4ab7b5d102b/src/ol/renderer/webgl/webglmaprenderer.js#L297-L316), it calls `goog.dispose(this.context_)`.  This is a no-op currently, because this function checks to see if the provided arg has a `dispose` method and does nothing if it doesn't.

But this convenient `goog.dispose()` function is masking the problem.  The `ol.webgl.Context` [*thinks* it is an event target](https://github.com/openlayers/ol3/blob/34d8d77344ee0b653770f065c593d4ab7b5d102b/src/ol/webgl/context.js#L20-L29), which should have a `dispose` method.  And it [implements `disposeInternal`(https://github.com/openlayers/ol3/blob/34d8d77344ee0b653770f065c593d4ab7b5d102b/src/ol/webgl/context.js#L160-L181), so someone thought this was going to be called.  But nope, the `ol.webgl.Context` [is missing](https://github.com/openlayers/ol3/blob/34d8d77344ee0b653770f065c593d4ab7b5d102b/src/ol/webgl/context.js#L104) the magic `goog.inherits()` call.

So `ol.webgl.Context` is not actually an `ol.events.EventEmitter`, and it doesn't have a `dispose` method.  And when `goog.dispose(obj)` is replaced with `obj.dispose()`, we get a failure.  Which unfortunately is happening in a `afterEach` function that isn't handling errors as expected.

Anyway, it is not clear to me why `ol.webgl.Context` needs to be an event target.  Maybe it should just be a disposable?  Would be nice to have something that demonstrated (or tested) the expected functionality.

tschaub:
It looks like `ol.webgl.Context` has been an event-target-that-isn't-an-event-target [from the start](https://github.com/openlayers/ol3/commit/10c41cd064976ba045f1971396311a1b5d5dbbe1#diff-56c9bfd95972e67312b8e0332a946eea).  Anybody care to speculate about whether it should inherit from `ol.events.EventTarget` or `ol.Disposable`?

src/ol/disposable.js
tschaub: @gberaudo I agree this is a good discussion to have.  Previously, we said the user should call `map.setTarget(null)` to do a bit of cleanup, but we didn't expose `map.dispose()`.  I don't have time to find the links right now.  But because this pull request doesn't change anything about exports, I think it is best to have that discussion on a fresh ticket.

ahocevar:
>  Anybody care to speculate about whether it should inherit from `ol.events.EventTarget` or `ol.Disposable`?

This is really just a speculation, but since it doesn't dispatch any events, I'd say it should inherit from `ol.Disposable`

src/ol/disposable.js
gberaudo: Thank you @tschaub for the explanation; it is fine with me to eventually continue this discussion in a fresh ticket. You were probably referring to https://github.com/openlayers/ol3/pull/3420.

fredj:
+1 for `ol.Disposable` and a `ol.events.unlistenAll(this.canvas_)` must be added in `ol.webgl.Context.prototype.disposeInternal`

tschaub:
> This is really just a speculation, but since it doesn't dispatch any events, I'd say it should inherit from ol.Disposable

I've changed this in e1f0e8d75b640f09085c6c44e45139f46a40da4a.

> a ol.events.unlistenAll(this.canvas_) must be added in ol.webgl.Context.prototype.disposeInternal

I've added this in 2f9d4e68ebe4fa1e2645473bfc01976b61d4ae2d.

Thanks for the reviews @gberaudo, @ahocevar, and @fredj.  Does this look in good shape to everyone now?

gberaudo:
LGTM, thanks.

ahocevar:
Looks great. Thanks for the hard work @tschaub. Please merge.

fredj:
+1

PR #4960, opened 2016-03-02
Simplify clipping of lower resolution regions

ahocevar:
This fixes the remaining issue with lines at tile borders described in #4934. In addition, it fixes another issue that can be seen in http://jsfiddle.net/norric/mopu8d7x/ when zooming in and out fast and for a long time: the clip geometry was created in a way that inner rings could have overlaps, which resulted in no clipping.

In addition to fixing several issues, this pull request simplifies the code and should make higher resolution tile clipping easier to read and understand.


ahocevar:
https://github.com/openlayers/ol3/issues/4934#issuecomment-191669241 confirms that the regression is finally fixed with this change.

Thanks for any review.

fredj:
LGTM

PR #4962, opened 2016-03-02
Skip tests where features are not available

tschaub:
This adds a global `where` function that can be used to selectively ignore tests where features are not available.

Ideally, this makes tests pass in IE < 11 (where `Uint8ClampedArray` and `ol.source.Raster` do not work).

### DIFF
--- a/test/spec/.eslintrc
+++ b/test/spec/.eslintrc
@@ -6,6 +6,7 @@
     "afterLoadText": false,
     "expect": false,
     "proj4": false,
-    "sinon": false
+    "sinon": false,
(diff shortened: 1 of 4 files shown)


marcjansen:
Nice and readable solution, Tim. I'll test in old IEs tomorrow. If you are eager getting this in, you can merge before my feedback as this looks like such a nice improvement.

ahocevar:
Thanks for addressing this @tschaub. I just tested in IE10 and IE9, and all tests pass. I think this is good to merge.

PR #4963, opened 2016-03-02
Handle more XHR errors in the TileJSON source

### LINKED ISSUE #4948: XmlHttpRequest inside of TileJSON constructor provides no error mechanism
When the TileJSON constructor initializes the xhr, callbacks for success and fail are defined. However there is no callback for onerror() defined. Currently, if an application hits a valid (but non-OpenLayer) URL and an error occurs there will be no response (handleTileJSONError is never called) and the application has no way of notifying the user.

tschaub:
The TileJSON source currently fires a `change` event and reports an error state on the `load` event for a XMLHttpRequest if the status is not 2xx.

This branch makes it so the same thing happens on the XHR `error` event (this happens on network failures, missing CORS headers, etc.) and when JSON parsing fails.

Fixes #4948.

### DIFF
--- a/src/ol/source/tilejsonsource.js
+++ b/src/ol/source/tilejsonsource.js
@@ -44,21 +44,45 @@ ol.source.TileJSON = function(options) {
     ol.net.jsonp(options.url, this.handleTileJSONResponse.bind(this),
         this.handleTileJSONError.bind(this));
   } else {
-    var xhr = new XMLHttpRequest();
-    xhr.open('GET', options.url, true);
-    xhr.onload = function(e) {
-      if (xhr.status >= 200 && xhr.status < 300) {
-        var response = /** @type {TileJSON} */(JSON.parse(xhr.responseText));
-        this.handleTileJSONResponse(response);
-      } else {
-        this.handleTileJSONError();
-      }
-    }.bind(this);
-    xhr.send();
+    var client = new XMLHttpRequest();
+    client.addEventListener('load', this.onXHRLoad_.bind(this));
+    client.addEventListener('error', this.onXHRError_.bind(this));
+    client.open('GET', options.url);
+    client.send();
   }
 
 };
 goog.inherits(ol.source.TileJSON, ol.source.TileImage);
+
+
+/**
+ * @private
+ * @param {Event} event The load event.
+ */
+ol.source.TileJSON.prototype.onXHRLoad_ = function(event) {
+  var client = /** @type {XMLHttpRequest} */ (event.target);
+  if (client.status >= 200 && client.status < 300) {
+    var response;
+    try {
+      response = /** @type {TileJSON} */(JSON.parse(client.responseText));
+    } catch (err) {
+      this.handleTileJSONError();
+      return;
+    }
+    this.handleTileJSONResponse(response);
+  } else {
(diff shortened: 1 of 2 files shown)


tschaub:
Hrm.  Would be nice if the [compiler knew of](https://travis-ci.org/openlayers/ol3/builds/113235845) `XMLHttpRequestProgressEvent`.  Pull request anyone?

tschaub:
I added regression tests that cover more errors than we were checking previously.  Prior to f35d0106b82f0157728ac261a14cb05bdf8a13f5, CORS and other network errors would have updated the source state to `error` and triggered the `change` event.  JSON parsing errors were not handled previously.

So I think it is fair to call this a regression (`goog.net.EventType.COMPLETE` is dispatched on XHR `error` events, and we were calling `source.handleTileJSONError()` for unexpected status codes).

src/ol/source/tilejsonsource.js
```diff
@@ -62,6 +56,36 @@ goog.inherits(ol.source.TileJSON, ol.source.TileImage);
...
+};
+
+
+/**
+ * @private
+ * @param {Event} event The error event.
+ */
+ol.source.TileJSON.prototype.onXHRError_ = function(event) {
+  this.handleTileJSONError();
+};
```
marcjansen: Technically we don't need this method as we could directly register `this.handleTileJSONError`, right?

marcjansen:
One minor comment which you may want to take. Otherwise looks good to me. Please merge at will and thanks.

src/ol/source/tilejsonsource.js
tschaub: Well, this gets called with a `XMLHttpRequestProgressEvent`, which the `handleTileJSONError` method doesn't expect.  We currently ignore this event, but it is the source of all the useful information, and it could make sense to make it available somehow.  But we also have things like parsing errors that result in `handleTileJSONError`, so it doesn't make sense that this method should always expect a `XMLHttpRequestProgressEvent`.

tschaub:

ahocevar:
Thanks for addressing this @tschaub.

fredj:
> Hrm.  Would be nice if the [compiler knew of](https://travis-ci.org/openlayers/ol3/builds/113235845) `XMLHttpRequestProgressEvent`.  Pull request anyone?

What about `ProgressEvent`: https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L2402-L2406

fredj:
>> Hrm.  Would be nice if the [compiler knew of](https://travis-ci.org/openlayers/ol3/builds/113235845) `XMLHttpRequestProgressEvent`.  Pull request anyone?

>What about `ProgressEvent`: https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js#L2402-L2406

No, the externs are incorrect:
```
ERR! compile ol3/src/ol/source/tilejsonsource.js:48: ERROR - actual parameter 2 of XMLHttpRequest.prototype.addEventListener does not match formal parameter
ERR! compile found   : function ((ProgressEvent|null)): undefined
ERR! compile required: (EventListener|function (Event): (boolean|undefined)|null)
ERR! compile     client.addEventListener('load', this.onXHRLoad_.bind(this));
```

PR #4964, opened 2016-03-03
Handle aborted tiles in ol.TileQueue

### LINKED ISSUE #2646: TileCache corruption and unhandled exception
I was playing around with lower ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK values because the default 2048 can become very big (multiple gigabytes) with many tiled layers.

With ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK set to a rather low value like 32 after a while zooming and panning no new tiles are loaded and cached tiles will dissapear gradually when zooming panning leaving a blank map. Some time after this behaviour I noticed errors in the console log and looking into them I find that usedCache can become undefined here:
https://github.com/openlayers/ol3/blob/master/src/ol/tilecache.js#L48 which will be logged as an console message similar to "Uncaught TypeError: Cannot use 'in' operator to search for '9' in undefined." in Chrome.

I've tried to reproduce directly in ol3 examples (Stamen and WMS tile) by pausing and setting ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK to 32 I get the same unwanted behaviour just after a short while panning and zooming but as of yet I haven't been able to reproduce the unhandled exception in those examples.

I've reproduced this error multiple times on Chrome 36 in Linux but have heard reports of the same in Chrome OS X and Internet Explorer on Windows.

I have not been able to reproduce anything with the default ol.DEFAULT_TILE_CACHE_HIGH_WATER_MARK but perhaps it's just a matter of time panning and zooming a heavy tiled map.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/4086072-tilecache-corruption-and-unhandled-exception?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #4961: Tile fetching issue when maximum cache size is reached
#4805 reveals an issue with tile fetching. It can be easily reproduced by e.g. using the simple.html example and changing the code to:
```js
var map = new ol.Map({
  layers: [
    new ol.layer.Tile({
      source: new ol.source.OSM({cacheSize: 16})
    })
  ],
  target: 'map',
  view: new ol.View({
    center: [0, 0],
    zoom: 2
  })
});
```
After zooming in and out using wheel zoom a few times, no more tiles will be loaded and only a single tile will remain.

fredj:
Fixes #4961
Fixes #2646

### DIFF
--- a/src/ol/dom/dom.js
+++ b/src/ol/dom/dom.js
@@ -36,31 +36,29 @@ ol.dom.canUseCssTransform = (function() {
     if (canUseCssTransform === undefined) {
       goog.asserts.assert(document.body,
           'document.body should not be null');
-      if (!goog.global.getComputedStyle) {
-        // this browser is ancient
-        canUseCssTransform = false;
-      } else {
-        var el = document.createElement('P'),
-            has2d,
-            transforms = {
-              'webkitTransform': '-webkit-transform',
-              'OTransform': '-o-transform',
-              'msTransform': '-ms-transform',
-              'MozTransform': '-moz-transform',
-              'transform': 'transform'
-            };
-        document.body.appendChild(el);
-        for (var t in transforms) {
-          if (t in el.style) {
-            el.style[t] = 'translate(1px,1px)';
-            has2d = goog.global.getComputedStyle(el).getPropertyValue(
-                transforms[t]);
-          }
-        }
-        document.body.removeChild(el);
+      goog.asserts.assert(goog.global.getComputedStyle,
+          'getComputedStyle is required (unsupported browser?)');
 
-        canUseCssTransform = (has2d && has2d !== 'none');
+      var el = document.createElement('P'),
+          has2d,
+          transforms = {
+            'webkitTransform': '-webkit-transform',
+            'OTransform': '-o-transform',
+            'msTransform': '-ms-transform',
+            'MozTransform': '-moz-transform',
+            'transform': 'transform'
+          };
+      document.body.appendChild(el);
+      for (var t in transforms) {
+        if (t in el.style) {
+          el.style[t] = 'translate(1px,1px)';
+          has2d = goog.global.getComputedStyle(el).getPropertyValue(
+              transforms[t]);
+        }
       }
+      document.body.removeChild(el);
+
+      canUseCssTransform = (has2d && has2d !== 'none');
     }
     return canUseCssTransform;
   };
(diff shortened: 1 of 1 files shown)


ahocevar:
I did some testing, and unfortunately this only fixes a part of the issue. It is still easy to reach a point where no more tiles are loaded. The improvement is that all previously loaded tiles are still available, not just a single tile.

fredj:
I've tested the bing example with a 0 `cacheSize` and it seems to work

ahocevar:
I have tested the simple example. Tried cache sizes from 0 to 20. Frequent and fast wheel-zooming in and out triggers the problem.

ahocevar:
bing-maps example with this branch, `cacheSize: 20`, after aggressively wheel-zooming in and out for about 20 seconds:
untitled

fredj:
ok, I was not fast enough

fredj:
The `ol.TileState.ABORT` state was only set if the tile was loading; it is now always set

tschaub:
The tile queue tests are still pretty weak.  It would be nice to add tests for this change.

fredj:
tests added

ahocevar:
Unfortunately I still run into the issue, despite your changes @fredj.

ahocevar:
The issue I'm seeing now is probably unrelated, because once the map is resized, tiles will continue to load, which wasn't the case before your change @fredj.

ahocevar:
I think this is a great improvement, so please merge. I'll create a separate ticket for the other issue.

ahocevar:
Actually I don't know what to write in a new ticket - the description is the same. The improvement is that tiles will continue to load when `map.render()` is called (programmatically or by zooming/panning again). @fredj, do you have any idea what might be causing this?

fredj:
I'm not sure but maybe `ol.Map#renderFrame_` must be called as long as there is tiles to load

ahocevar:
> I'm not sure but maybe ol.Map#renderFrame_ must be called as long as there is tiles to load

I was thinking along the same lines. But then I tried with the standard cache size of 2048, and even when the cache was full, I did not run into the issue. So there might be something else wrong that we have not found yet.

fredj:
> I'm not sure but maybe `ol.Map#renderFrame_` must be called as long as there is tiles to load

I've just found that every time that the state of a tile change; `render` is called (`ol.Map#handleTileChange_` callback)

fredj:
pull request automatically closed by github because I pushed master to my branch, continues in #5051

fredj:
Thanks a lot for the tests and review @ahocevar

PR #4967, opened 2016-03-03
Add missing goog.inherits for ol.render.canvas.Immediate

### LINKED ISSUE #4966: Extends annotation but no goog.inherits
* `ol.webgl.Context` `@extends {ol.events.EventTarget}` without `goog.inherits`
 * `ol.render.canvas.Immediate` `@extends {ol.render.VectorContext}` without `goog.inherits`

fredj:
Fixes #4966


tschaub:
While our smoke tests for constructors seem a bit silly, they do serve to reinforce our intentions and would catch errors like this missing `inherits`.  Here the, the `ol.renderer.canvas.Immediate` tests could be given a couple assertions about `instanceof`.

tschaub:
Looks good @fredj.  Please merge when green.

PR #4977, opened 2016-03-06
Add basic tests for ol.geom.flat.length methods

marcjansen:
This PR suggests to add basic tests for the previously untested `ol.geom.flat.length.lineString()` and `ol.geom.flat.length.linearRing()`.

Please review.


marcjansen:
Hmmm. No coverage here, strange.

marcjansen:
Coverage was calculated, but not reported https://coveralls.io/builds/5322099

fredj:
+1, thanks !

marcjansen:
Thanks Fred!

PR #4983, opened 2016-03-07
Update Bing Maps Portal url

ahocevar:
It looks like http://bingmapsportal.com/ is no longer working. So I changed the URL to the working http://www.bingmapsportal.com/.


fredj:
+1

PR #4984, opened 2016-03-07
Add new ol.geom.Geometry#rotate function

fredj:


fredj:
Thanks for any review

tsauerwein:
That looks good to me! Maybe a test for `ol.geom.Geometry.prototype.rotate` would be nice, but as you already have solid tests for `ol.geom.flat.transform.rotate` I wouldn't say it's required.

ca0v:
How about doing the same with scale?  See #5684.

src/ol/geom/geometry.js
```diff
@@ -170,6 +170,16 @@ ol.geom.Geometry.prototype.getExtent = function(opt_extent) {
 
 
 /**
+ * Rotate the geometry around a given coordinate. This modifies the geometry
+ * coordinates in place.
+ * @param {number} angle Rotation angle in radians.
```
tschaub: Always nice to specify the direction (e.g. "Rotation angle in radians (counter-clockwise)").

PR #4990, opened 2016-03-08
Add new ol.source.TileJSON#getTileJSON function

### LINKED ISSUE #4987: ol.source.TileJSON contructor does not return full JSON payload
When a http response comes back, handleTileJSONResponse strips certain attributes from the full JSON payload to put into the returned source object. For instance if an application wants to know the display-friendly name of a map (vice the id), it has no way of knowing since the source object/handleTileJSONResponse never sets a 'this.name'. The TileJSON constructor should either redefine the source object to support the full payload info, or have the full JSON payload be returned as part of an event parameter as discussed in issue #4948.

fredj:
Fixes #4987

### DIFF
--- a/src/ol/source/tilejsonsource.js
+++ b/src/ol/source/tilejsonsource.js
@@ -29,6 +29,12 @@ goog.require('ol.source.TileImage');
  */
 ol.source.TileJSON = function(options) {
 
+  /**
+   * @type {TileJSON}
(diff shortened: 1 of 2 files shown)


marcjansen:
Nice. I like it. Please merge.

fredj:
Thanks for the review

PR #4991, opened 2016-03-08
Do not remove listeners while dispatching event

### LINKED ISSUE #4986: ol.events.EventTarget#dispatchEvent fails if a handler causes listeners to be removed
If an event handler causes other handlers for the same event type to be removed, `
ol.events.EventTarget#dispatchEvent` may call handlers multiple times. If multiple handlers are removed, dispatchEvent may throw an error.

This is dependent on the order of listeners, but see my fiddle here with a very basic worst-case (error) example:
http://jsfiddle.net/9cgmw6r8/2/

This is a regression from the Closure event stack, which copies the listener array and checks if each listener has been removed before calling it (see [goog.events.EventTarget#fireListeners](https://github.com/google/closure-library/blob/master/closure/goog/events/eventtarget.js#L261)).

ahocevar:
When listeners are removed in listeners for the same event type, unexpected behaviour can be observed (exceptions thrown, listeners called more than once). To fix this, we can defer listener removal until all listeners have been called.

Fixes #4986.

### DIFF
--- a/src/ol/events/eventtarget.js
+++ b/src/ol/events/eventtarget.js
@@ -70,11 +70,13 @@ ol.events.EventTarget.prototype.dispatchEvent = function(event) {
   var type = evt.type;
   evt.target = this;
   var listeners = this.listeners_[type];
+  var propagate;
   if (listeners) {
     this.pendingRemovals_[type] = 0;
     for (var i = 0, ii = listeners.length; i < ii; ++i) {
       if (listeners[i].call(this, evt) === false || evt.propagationStopped) {
-        return false;
+        propagate = false;
+        break;
       }
     }
     var pendingRemovals = this.pendingRemovals_[type];
@@ -82,6 +84,7 @@ ol.events.EventTarget.prototype.dispatchEvent = function(event) {
     while (pendingRemovals--) {
       this.removeEventListener(type, ol.nullFunction);
     }
+    return propagate;
   }
 };
 
--- a/src/ol/events/eventtarget.js
+++ b/src/ol/events/eventtarget.js
@@ -72,7 +72,9 @@ ol.events.EventTarget.prototype.dispatchEvent = function(event) {
   var listeners = this.listeners_[type];
   var propagate;
   if (listeners) {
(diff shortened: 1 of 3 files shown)


ahocevar:
This is now ready for review.

ahocevar:
[coveralls.io result](https://coveralls.io/builds/5344617)

src/ol/events/eventtarget.js
```diff
@@ -64,12 +71,17 @@ ol.events.EventTarget.prototype.dispatchEvent = function(event) {
   evt.target = this;
   var listeners = this.listeners_[type];
   if (listeners) {
-    for (var i = listeners.length - 1; i >= 0; --i) {
-      if (listeners[i].call(this, evt) === false ||
-          evt.propagationStopped) {
+    this.pendingRemovals_[type] = 0;
+    for (var i = 0, ii = listeners.length; i < ii; ++i) {
+      if (listeners[i].call(this, evt) === false || evt.propagationStopped) {
```
tschaub: It would be nice to clean up the pending removals here too.  Maybe set a flag and `break`.  Then `return false` after cleaning up if the flag is set.

src/ol/events/eventtarget.js
```diff
@@ -64,12 +71,17 @@ ol.events.EventTarget.prototype.dispatchEvent = function(event) {
   evt.target = this;
   var listeners = this.listeners_[type];
   if (listeners) {
-    for (var i = listeners.length - 1; i >= 0; --i) {
-      if (listeners[i].call(this, evt) === false ||
-          evt.propagationStopped) {
+    this.pendingRemovals_[type] = 0;
```
tschaub: This is a narrow case that we may decide is not worth supporting, but if a listener dispatches another of the same event, this will incorrectly be set to `0` (if previous listeners have been removed).

ahocevar:
I addressed both of your comments @tschaub, and added additional tests for the cases you mentioned.

tschaub:
Looks great @ahocevar.  Slick solution.  Please merge.

Issue #4993, opened 2016-03-08
Alpha value in ol.Color


jonataswalker:
When `ol.Color` is used as Array, `alpha` value could be `1` by default instead of [`undefined`](https://github.com/openlayers/ol3/blob/master/src/ol/color/color.js#L296), so we can use like `[51, 51, 51]` instead of `[51, 51, 51, 1]`.

tschaub:
I agree this would be nice.

fredj:
Fixes with #4994

PR #4994, opened 2016-03-08
Add alpha default value to ol.Color array form

jonataswalker:
See #4993.

If `ol.Color` is informed like `[51, 51, 51]`, alpha value is default `1`.

### DIFF
--- a/src/ol/color/color.js
+++ b/src/ol/color/color.js
@@ -17,7 +17,8 @@ goog.require('ol.math');
 /**
  * A color represented as a short array [red, green, blue, alpha].
  * red, green, and blue should be integers in the range 0..255 inclusive.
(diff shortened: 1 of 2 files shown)


fredj:
can you also add a test in `test/spec/ol/color.test.js`?

jonataswalker:
@fredj test added, let me know if it's what you have in mind.

fredj:
Yes, thanks

tschaub:
It feels like a bit of a risk to have `ol.color.isValid()` return `false` for the 3 item form.  In addition, it would be nice to have `ol.color.normalize()` work for the same.  But it looks like neither of these is accessible from places where you might provide a 3 item array.

I've proposed the removal of other unused code that doesn't work with the 3 item form in #4995.

PR #5003, opened 2016-03-09
Add 'function' jsdoc tag to ol.geom.Geometry#rotate

fredj:
Otherwise the function is in the 'Members' section of the API documentation.


ahocevar:
Interesting, but please merge.

PR #5006, opened 2016-03-09
Use the maintained slimerjs package

### LINKED ISSUE #5005: Custom build
I've a got a lot errors on a custom building like:

```
ERR! compile /home/www/ol3/node_modules/closure-util/.deps/library/c00d1243c49bb8dad16f671f09a3facded9f3cd9/closure/goog/array/array.js:55: ERROR - Name Array in JSDoc is implicitly nullable.
ERR! compile Please add a ! to make it non-nullable, or a ? to make it explicitly nullable.
ERR! compile  * @typedef {Array|NodeList|Arguments|{length: number}}
ERR! compile              ^
```

ahocevar:
Fixes #5005.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "proj4": "2.3.12",
     "resemblejs": "2.2.0",
     "sinon": "1.17.3",
-    "slimerjs-edge": "0.10.0-pre-3",
+    "slimerjs": "0.906.1",
     "wrench": "1.5.8"
   },
   "eslintConfig": {
--- a/tasks/test-rendering.js
+++ b/tasks/test-rendering.js
@@ -6,7 +6,7 @@
 var path = require('path');
 var spawn = require('child_process').spawn;
 
-var slimerjs = require('slimerjs-edge');
+var slimerjs = require('slimerjs');
 
 var serve = require('./serve');
 var listen = require('./test').listen;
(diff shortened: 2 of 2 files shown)


ahocevar:
Yay! Works for me locally on OSX, and Travis is happy too. Bye-bye slimerjs-edge.

fredj:
LGTM

tsauerwein:
:+1:

marcjansen:
I am not so sure.

I changed an expected image so the rendering test fails (in a browser), but the error is only reported when run on the commandline:

```bash
$ make test-rendering 
node tasks/test-rendering.js
Gecko: 38.0.5

Script Error: [JavaScript Error: "Error: expected '48.60' to be below 1.5" {file: "http://127.0.0.1:3001/node_modules/expect.js/index.js" line: 96}]
       Stack:
         -> http://127.0.0.1:3001/node_modules/expect.js/index.js: 96

ol.rendering.Map #pan() tests the canvas renderer
Error: Error: expected '48.60' to be below 1.5 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)

1 test(s) failed.
```

image

**The return code of `make test-rendering` is `0` i.e. `success`, though.**

I think if we switch back to plain `slimerjs` we need to include the grepping of `stdout` against `test(s) failed`.

marcjansen:
Something like

```diff
diff --git a/tasks/test-rendering.js b/tasks/test-rendering.js
index 04095e0..1332129 100644
--- a/tasks/test-rendering.js
+++ b/tasks/test-rendering.js
@@ -38,8 +38,17 @@ serve.createServer(function(err, server) {
       url + '/test_rendering/index.html'
     ];
 
-    var child = spawn(slimerjs.path, args, {stdio: 'inherit'});
+    var child = spawn(slimerjs.path, args, {stdio: 'pipe'});
+    var output = '';
+    child.stdout.on('data', function(data) {
+      output += data.toString();
+    });
     child.on('exit', function(code) {
+      var failedTestsRegExp = /[1-9][\d]* test\(s\) failed\./g;
+      if (failedTestsRegExp.test(output)) {
+        process.stderr.write('Test failures detected.\n');
+        code = 1;
+      }
       process.exit(code);
     });
   });
```

works for me.

ahocevar:
Thanks @marcjansen, I updated the pull request with a modified version of your suggestion. Can you take another look please?

marcjansen:
Works great now. Thanks for updating, @ahocevar.

Please merge if the build also passes on travis.

tsauerwein:
The return code always being 0 was the original reason we used `slimerjs-edge`. But the pragmatic work-around by Mark is a good idea. Maybe we should match for "All tests passed", because there might be other reasons that running the tests fails, e.g. https://github.com/ahocevar/ol3/blob/fa17196cb3f4d08c7340e50b68a4b9466d0002b2/test_rendering/test.js#L28

marcjansen:
> Maybe we should match for "All tests passed", …

I'd be OK with that as well.

ahocevar:
Updated pull request to test for "All tests passed."

marcjansen:
Also easier to read an reason about. Plus one for merging

tsauerwein:
Thanks Andreas!

marcjansen:
Yeah, thanks!

PR #5007, opened 2016-03-10
Accept simpler forms of specifying attribution(s) for sources.

### LINKED ISSUE #4100: Simplify attributions
I think the most straightforward way to support source attribution would be to allow either of the following:

```js
var source = new ol.source.XYZ({
  attribution: 'My <b>favorite</b> tile source'
})
```

or

```js
var source = new ol.source.XYZ({
  attribution: function(view) {
    // or frameState if we want awkwardness
    return 'Maybe this depends on what is rendered';
  }
});
```

Instead we have an `attributions` option for sources where each `ol.Attribution` accepts an `html` option and an undocumented `tileRanges` option (of type `Object.<string, Array.<ol.TileRange>>`).  Since the `tileRanges` option is undocumented and unnecessarily tied to one particular implementation (not all sources are tile sources), users do not have real control over which of the attributions is rendered, and there is no purpose to supplying an array.

Since `ol.Attribution` is marked as stable, one option would be to give it a `render` option (or similar) that is type `function(ol.View):boolean` and determines when the attribution is rendered.

Another option would be to remove the experimental `attributions` options from tile sources and add a single `attribution` option that was type `string|function(ol.View):boolean`.  This would leave `ol.Attribution` orphaned and deprecated.

Any opinions?

marcjansen:
The way we can specify attributions for `ol.source.Source` and subclasses is quite verbose for the most common use case: on simple string attribution.

This PR suggests to accept a 

* `string`
* `Array<string>`
* `ol.Attribution`
* `Array<ol.Attribution>`
* `undefined`

as configuration value for the `attributions` key in the constructor options of `ol.source.Source`.

This API is fully backwards compatible. 

I also updated our examples to make use of the simpler way of specifying attributions.

Please review.

/cc @tschaub, I know you wanted less awkwardness here, so I'd value your feedback here. Other may of course also chime in.

Fixes #4100

### DIFF
--- a/src/ol/source/source.js
+++ b/src/ol/source/source.js
@@ -21,7 +21,7 @@ ol.source.State = {
 
 
 /**
- * @typedef {{attributions: (Array.<ol.Attribution>|undefined),
+ * @typedef {{attributions: (olx.source.AttributionOption|undefined),
  *            logo: (string|olx.LogoOptions|undefined),
  *            projection: ol.proj.ProjectionLike,
  *            state: (ol.source.State|undefined),
@@ -57,8 +57,7 @@ ol.source.Source = function(options) {
    * @private
    * @type {Array.<ol.Attribution>}
    */
-  this.attributions_ = options.attributions !== undefined ?
-      options.attributions : null;
+  this.attributions_ = ol.source.Source.toAttributionsArray_(options.attributions);
 
   /**
    * @private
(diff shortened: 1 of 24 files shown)


probins:
I'm in favour of this change (once the errors have been resolved)

bartvde:
I haven't looked at the code but I like the idea

marcjansen:
Travis passes now. Please review.

fredj:
You also need to update the `*.prototype.attributions;` types in olx.js, otherwise the API doc will be incorrect

src/ol/source/source.js
```diff
@@ -82,6 +81,37 @@ ol.source.Source = function(options) {
...
+ *     `ol.Attribution`, array of `ol.Attribution` or undefined.
+ * @return {Array.<ol.Attribution>} The array of `ol.Attribution` or null if
+ *     `undefined` was given.
+ */
+ol.source.Source.toAttributionsArray_ = function(attributionLike) {
+  if (typeof attributionLike === 'string') {
+    return [new ol.Attribution({html: attributionLike})];
+  } else if (attributionLike instanceof ol.Attribution) {
+    return [attributionLike];
+  } else if (goog.isArray(attributionLike)) {
```
tschaub: This should be `Array.isArray()`, right?

marcjansen:
> You also need to update the *.prototype.attributions; types in olx.js

The API docs still don't look perfect :disappointed: 

`ol.source.Source#setAttributions`
image

Constructor `ol.source.Vector`
image

Type definition:
image

If there is anythoing I can do to enhance this, I am open to suggestions.

src/ol/source/source.js
marcjansen: Will change.

tschaub:
This looks great @marcjansen.  Thanks for taking this on, I think it is a good way to provide convenience and maintain compatibility (although `ol.Attribution` is completely useless).

After a string, the second most useful thing to provide would be a function that accepted a resolution and extent and returned a string.  But I don't think this needs to be added now.

fredj:
Fixes #4100

marcjansen:
I need to rebase, so I'll have to postpone this one. Thanks everybody for the comments.

PR #5009, opened 2016-03-10
Update tutorials

gberaudo:


gberaudo:
Related to https://github.com/openlayers/ol3/issues/5005

gberaudo:
// cc @ahocevar

ahocevar:
Looks good, thanks @gberaudo.

gberaudo:
Thanks for the quick review and merge @ahocevar.

probins:
what effect does removing "goog.dom.ASSUME_STANDARDS_MODE=true" have?

fredj:
See https://github.com/google/closure-library/blob/master/closure/goog/dom/dom.js#L59-L63

probins:
yes, but that doesn't explain what effect removing has. @gberaudo calls it 'disturbing' in the commit msg; if it's disturbing, why's it in the config/*json files?

gberaudo:
It is disturbing for a tutorial, where we should focus on the most important points.
This define is a micro optimization, which is still usind in the last example of the closure.md tutorial.

PR #5012, opened 2016-03-10
Mark some ol.events.* functions as @api

adube:
This PR marks the following methods as `@api`:

```
ol.events.listen
ol.events.listenOnce
ol.events.unlisten
ol.events.unlistenByKey
ol.events.unlistenAll
```

I require them in an library where OpenLayers isn't built along it.

### DIFF
--- a/src/ol/events.js
+++ b/src/ol/events.js
@@ -211,6 +211,7 @@ ol.events.removeListeners_ = function(target, type) {
  *     listener. Default is the `target`.
  * @param {boolean=} opt_once If true, add the listener as one-off listener.
  * @return {ol.events.Key} Unique key for the listener.
+ * @api
  */
 ol.events.listen = function(target, type, listener, opt_this, opt_once) {
   var listenerMap = ol.events.getListenerMap_(target);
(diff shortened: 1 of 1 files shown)


adube:
This isn't much, but I guess it's ready for review.

ahocevar:
You should be using `ol.Observable#on`, `#un` and `#unByKey`. We won't make these ol.events.* functions exportable.

adube:
That's unfortunate.  These are easier to manage because they always return the key.  The `on` method returns an array of keys, making the un-listening slightly less user-friendly to manage.

Is there any way we could reconsider this requirement ?

ahocevar:
Not sure I understand. With the API, you can do
```js
var key = myEventTarget.on('foo', listener);
myEventTarget.unByKey(key);
key = myEventTarget.on(['foo', 'bar'], listener);
myEventTarget.unByKey(key);
```
Isn't that user friendly?

adube:
I just noticed that the `ol.Observable.unByKey` accepts an array or a key.  So yeah, it's very user-friendy!  I agree with using `on`.  Thanks for you time.

PR #5024, opened 2016-03-12
Resize the map canvas less aggressively

### LINKED ISSUE #5022: Jerky panning in Firefox
I see a very noticeable stutter every second or so when panning. Doesn't happen in Chrome or Edge.

ff

ahocevar:
It looks like Firefox performance suffers when the width and height of a canvas change too frequently. This pull request changes things so we check if the size is still appropriate and only do a clearRect in this case.

Fixes #5022.


fredj:
LGTM

PR #5027, opened 2016-03-13
Clarify default value of featureProjection on geometry write

probins:
I noticed that this is not currently specified.

Docs only, no logic changes


bartvde:
thanks @probins

PR #5031, opened 2016-03-14
support zooming out for ol.interaction.DragZoom

drnextgis:


ahocevar:
What is the intended use? As a second DragZoom interaction with a different condition? Will two DragZoom interactions on the same map work? Maybe an example would clarify things a bit.

src/ol/interaction/dragzoominteraction.js
```diff
@@ -66,6 +80,9 @@ ol.interaction.DragZoom.prototype.onBoxEnd = function() {
   var currentCenter = view.getCenter();
   goog.asserts.assert(currentCenter !== undefined, 'center should be defined');
 
+  var resolution = view.constrainResolution(
```
ahocevar: Was there a reason for moving this block down?

drnextgis: Because `extent` values is changed within `if (this.out_)` block.

ahocevar: I meant why you moved it down to below the assertions.

drnextgis: This is not intentionally, I'll move it back.

drnextgis:
Yes, it is intended to use as second DragZoom interaction with a different condition. Here is example: http://jsfiddle.net/_DR_/gxee94o6/

ahocevar:
Ok cool. Can you please add a unit test and submit a [CLA](https://github.com/openlayers/cla). Thanks!

drnextgis:
I've already submitted [CLA](https://github.com/openlayers/cla/blob/master/signed-agreements.md) some years ago: `Rykov, Denis Rykov (@dr)`

ahocevar:
Thanks @drnextgis. Are you still using the @dr handle? If not, I'll update the CLA list.

drnextgis:
Yes update it please. I've added unit test and perform PR rebasing.

ahocevar:
Thanks @drnextgis.

PR #5033, opened 2016-03-14
Document GPX geometry output types

probins:
Outputting LineStrings as Routes and MultiLineStrings as Tracks seems a bit dubious to me, but at any rate it should be documented.

Docs only, no changes to logic.


ahocevar:
Thanks @probins.

Issue #3220, opened 2015-02-09
SVG Icon in a Vector Layer causes IndexSizeError in IE11


jnpritchard:
`var layer = new ol.layer.Vector({`
...
`style: new ol.style.Style({`
`                    image: new ol.style.Icon({`
`                        src: 'images/myicon.svg'`
`                    })`
`                })`
`            });`

Adding an icon as a style attribute to a Vector layer results in errors at the case ol.render.canvas.Instruction.DRAW_IMAGE: in ol.render.canvas.Replay.prototype.replay_

Found in IE11, it appears that the width and height of the image are zero, which causes the error to occur.

DanielSundberg:
I've got this as well. 

The only way for me to get rid of this error was to use a PNG image instead.

fredj:
See https://connect.microsoft.com/IE/feedbackdetail/view/925655/svg-image-has-0x0-size-in-ie11

DanielSundberg:
@fredj, doesn't help us since we have to support IE11.

ahocevar:
Can't you just set the `imgSize` of your SVG icon? Like
```js
new ol.style.Style({
  image: new ol.style.Icon({
    src: 'images/myicon.svg',
    imgSize: [24, 24]
  })
})
```

DanielSundberg:
For some reason that did not work for me.

ahocevar:
Sorry, my snippet above is invalid. We currently assert that `imgSize` is not set when `src` is set.

ahocevar:
Fixed with #5036.

tomchadwin:
I'm still getting this issue. Is #5036 in 3.16.0?

ahocevar:
Yes it is @tomchadwin. Can you provide a JSFiddle that shows the error with the latest release?

tomchadwin:
I'll try. It occurs in output automatically generated by qgis2web, so I'll have to see what happens as I strip the output down to functional minimum. Thanks.

tomchadwin:
Perhaps my error (from inherited code). So `imgSize` is required for IE11 SVGs, *even if they use `src`, and not `img*`?

ahocevar:
We are aware of the fact that the `size` property is very misleading. It is mostly for image sprites, if you only want to use a subset of the image. If you want to resize client-side, you'll need to configure `scale`. So if you want half the size, configure `scale: 0.5`.

And yes, for IE11, you need to set `imgSize` even when you configure `src` instead of `img`.

[...]

PR #5035, opened 2016-03-15
Check SVG icon size returns 0 to show SVG icon on IE 11

GeoDodo:
I can reproduce the problem using an SVG icon on Internet Explorer 11 as described in #3220.  

The solution from @ahocevar (add 'size' to the icon configuration) works until the calculation for the safari was added (936cd6a). On the IE 11 image.height and image.width returns 0  based on the bug mentioned in the ticket.

With the patch on the IE 11 the svg-icon use the size from the configuration.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -348,8 +348,8 @@ ol.render.canvas.Replay.prototype.replay_ = function(
             context.globalAlpha = alpha * opacity;
           }
 
-          var w = (width + originX > image.width) ? image.width - originX : width;
-          var h = (height + originY > image.height) ? image.height - originY : height;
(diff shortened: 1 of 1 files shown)


ahocevar:
Not sure if this is the right fix - can you please test #5036 if it also solves the problem?

GeoDodo:
#5036 do not work because the image var (https://github.com/openlayers/ol3/blob/master/src/ol/render/canvas/canvasreplay.js#L308) is the HTMLImageElement `` and not the ol.style.IconImage.

ahocevar:
Updated #5036. Can you test again please @GeoDodo?

marcjansen:
Though I am happy to see you provide a fixe here, @GeoDodo, I personally like the approach in #5036 better.

ahocevar:
Superseded by #5036.

PR #2944, opened 2014-11-13
Add parsing of some WMS 1.1.1 specific nodes

oterral:
This PR add the parsing of `<SRS>` and `<LatLonBoundingBox>` nodes (WMS 1.1.1)

Adds also tests for parsing of a WMS  Getcapabilities 1.1.1 file

PR #5040, opened 2016-03-15
WMS_MS_Capabilities 1.1.1 supports SRS instead of CRS

AnWeber:
WMS_MS_Capatiblities prior 1.3 only supports SRS instead of attribute CRS. In the documentation of WMSCapabilities is no version restriction available. Please include the SRS in the parsed WMSCapabilities.

http://schemas.opengis.net/wms/1.1.1/WMS_MS_Capabilities.dtd


bartvde:
see #2944 the format isn't currently meant for anything other than WMS 1.3.0

There is also JSONIX which can do the parsing for 1.1.1

AnWeber:
Thanks for your help.

PR #5042, opened 2016-03-15
WMSCapabilities1_1.js

AnWeber:
see #2944

I created a WMSCapabilities1_1.js according to this DTD: 
http://schemas.opengis.net/wms/1.1.1/WMS_MS_Capabilities.dtd

I don't like to change the existing WMSCapabilities to WMSCapabilities1_3 because this would be a breaking change.

### DIFF
--- a/src/ol/format/wmscapabilitiesformat1_1.js
+++ b/src/ol/format/wmscapabilitiesformat1_1.js
@@ -19,12 +19,12 @@ goog.require('ol.xml');
  */
 ol.format.WMSCapabilities1_1 = function() {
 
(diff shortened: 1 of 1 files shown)


bartvde:
I'm honestly not sure if we would want another non-feature format in ol3. Last time this was discussed with some of the developers we said use JSONIX for non feature formats. It would be more likely to get rid of WMS GetCapabilities altogether IMHO than adding support for another version if it, but that's my personal opinion on this.

AnWeber:
Thansk for the information. If this feature format is not wanted, I will delete the brancht

PR #5044, opened 2016-03-15
Snap interaction now handle feature without geometry properly

fblackburn:
The snap interaction was raising an error when trying to add a feature without geometry. This pull request now just pass over a feature without geometry.

### DIFF
--- a/src/ol/interaction/snapinteraction.js
+++ b/src/ol/interaction/snapinteraction.js
@@ -172,25 +172,30 @@ goog.inherits(ol.interaction.Snap, ol.interaction.Pointer);
  */
 ol.interaction.Snap.prototype.addFeature = function(feature, opt_listen) {
   var listen = opt_listen !== undefined ? opt_listen : true;
+  var feature_uid = goog.getUid(feature);
   var geometry = feature.getGeometry();
-  var segmentWriter = this.SEGMENT_WRITERS_[geometry.getType()];
-  if (segmentWriter) {
-    var feature_uid = goog.getUid(feature);
-    this.indexedFeaturesExtents_[feature_uid] = geometry.getExtent(
-        ol.extent.createEmpty());
-    segmentWriter.call(this, feature, geometry);
+  if (geometry) {
+    var segmentWriter = this.SEGMENT_WRITERS_[geometry.getType()];
+    if (segmentWriter) {
+      this.indexedFeaturesExtents_[feature_uid] = geometry.getExtent(
+          ol.extent.createEmpty());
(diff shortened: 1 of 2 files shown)


src/ol/interaction/snapinteraction.js
```diff
@@ -160,23 +160,25 @@ goog.inherits(ol.interaction.Snap, ol.interaction.Pointer);
...
+          ol.extent.createEmpty());
+      segmentWriter.call(this, feature, geometry);
+
+      if (listen) {
+        this.geometryModifyListenerKeys_[feature_uid] = ol.events.listen(
+            geometry,
+            ol.events.EventType.CHANGE,
+            this.handleGeometryModify_.bind(this, feature),
+            this);
+        this.geometryChangeListenerKeys_[feature_uid] = ol.events.listen(
```
ahocevar: This listener should also be registered when the feature has no geometry (yet).

fblackburn:
Now listen to the geometry change event on feature without geometry.

ahocevar:
Looks good now. Do you think you can add a test, verifying that the correct listeners are registered?

fblackburn:
I would like to add some test but I have some problems implementing it. I think the test case should be:

1. Create a snap interaction containing a feature without geometry.
2. Implement some way to proof that the feature is not considered by the interaction since it has no geometry
3. Modify the feature to add a geometry
4. Show the feature is now properly handled by sending an event to the snap interaction and checking the result.

My main problem is with the second point. Do you have any idea of a satisfying way to show that a feature is not considered by the interaction since it has no geometry?

ahocevar:
Thanks! No need to test for 2., because if there is no geometry, there is no geometry change listener. The important test is that the geometry gets picked up when it is set.

fblackburn:
I added a test that should properly show that a feature without geometry is correctly handled.

ahocevar:
This looks great now, thanks @fblackburn. Can you please submit a [CLA](https://github.com/openlayers/cla) so we can merge this? Thanks.

adube:
@ahocevar, fblackburn works at Mapgears.  Does he still need to submit a CLA ?

ahocevar:
@adube Thanks for the info. The Mapgears CCLA covers @fblackburn then. But please submit a pull request to update https://github.com/openlayers/cla/blob/master/signed-agreements.md. Also add github user names for existing Mapgears developers please. Thanks!

adube:
Will do.

adube:
Done.  Thanks, @ahocevar 

https://github.com/openlayers/cla/pull/5

PR #5045, opened 2016-03-15
Add autoPanIncludeMargin to the overlay options

adube:
This PR introduces the `autoPanIncludeMargin` options for the `ol.Overlay` object.  When set, the `autoPanMargin` value is taken into account when calculating whether the map should be automatically panned or not, which allows overlays to close to an edge (within the buffer created by the margin) to make the map automatically pan.

That option is set to `false` by default to keep the existing behaviour "as-is".

Opened for discussions.

### DIFF
--- a/examples/popup.js
+++ b/examples/popup.js
@@ -21,6 +21,7 @@ var closer = document.getElementById('popup-closer');
 var overlay = new ol.Overlay(/** @type {olx.OverlayOptions} */ ({
   element: container,
   autoPan: true,
+  autoPanIncludeMargin: true,
   autoPanAnimation: {
     duration: 250
   }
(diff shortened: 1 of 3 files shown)


ahocevar:
Couldn't you achieve the same by setting a css margin on the overlay element instead of using the autoPanMargin option?

adube:
I tried this and it didn't work for me.  Look at this snippet.  In it, I define a margin of `40px` and the `autoPanMargin` is set to 0.  I don't get the behaviour I want:

snippet

adube:
See a JSFiddle demonstrating this: http://jsfiddle.net/v4yd5epq/

adube:
It also messes the actual position of the popup as well (the pin doesn't point where we click).

tschaub:
I think you are saying you want your popup to be in an element with some padding.  So you could just create a wrapper element and provide that to the `ol.Overlay`.

E.g.

http://jsfiddle.net/bnq0aeh3/1/

image

adube:
That's a good idea.  I'll try to implement this in my app and come back and close this if it works.  Thanks for your support.

adube:
Yep, I've been able to accomplish what I need with an additional wrapper around the div.  That's clever.

Thanks to both of you!

Issue #3953, opened 2015-08-03
Option to specify precision of coordinate values on format write


probins:
A common floating-point arithmetic issue, when reading in 4326 coordinates to a projected view and then writing them back out in 4326:

value in: 45.460979
value out: 45.46097900000001

or

value in: 45.453625
value out: 45.45362499999999

On large geometries, this significantly increases the size of the output for no added value. How to improve this? Have a 'round to decimal places' option for transforms? In this case, for example, specify 6 places on output/write.

tschaub:
I think you mean that you would like to specify a precision when serializing geometries.  This would be an option to the write methods of a format.  This isn't strictly about transform methods.

probins:
yes, I've changed the issue title, as I've just come across a related issue when drawing. Drawing on OSM, for example, coordinate stored as [66041.59243839228, 5230127.223384898], which is a ridiculous level of precision for metre projections, but probably doesn't matter internally. Gets output in 4326 as [0.59326171875, 42.45994035221656], which is also spuriously precise, and in this case does matter as it bloats the size of the file unnecessarily. 6 decimal places should be ample for most apps.

probins:
Another example: incoming and outgoing geometries and view in 4326. Move a coordinate; new position is [11.120000000000001, 44.77]; presumably this is coming from the pixel->coordinate calculation and has nothing to do with transform.

If the incoming and outgoing geometries and the view are in the same projection, there is no transform and unmodified geometries are output as input. If not, then even if the only change is to an attribute the outgoing coordinates will not be the same as the input. IIRC OL2 had an option to store the incoming data, so if the only edit is to attributes, you could output unchanged geometries from there. To do this in OL3, you would have to fetch and store the file in the app.

Looking around the web for a solution, `parseFloat(45.45362499999999.toPrecision(8))` or `parseFloat(11.120000000001.toPrecision(8))` seems to do the trick. '8' is the maximum precision wanted, and `parseFloat` removes trailing zeroes.

probins:
where would be the best place to add this? After looking at the code, I'm thinking that it would be best to include this in `ol.format.Feature.transformWithOptions`: if the precision option is set, `parseFloat` on each transformed coordinate before returning the geometry.

probins:
I've just noticed there's a similar issue with other output fields. KML has a scale output, and this can be output for example as '0.6400000000000001'. Not sure what others there might be; anything stored as a float, I suppose.

ahocevar:
[...]

PR #5046, opened 2016-03-15
Fix cruft on kml scale output

probins:
I mentioned this in #3953. As this is easier to fix than the output coordinates, here's a PR to fix it; I'll look into the coordinates tomorrow.

Problem is that the default scale for text/labels is 0.8. `writeScaleTextNode_()` does 0.8*0.8, which produces not 0.64 as you might imagine but 0.6400000000000001. The change in this PR gets rid of this unwanted junk at the end.

Not sure what the best value for the no of decimal places is. I've picked 6, but this can easily be changed - or made a constant at the beginning of the file?

### DIFF
--- a/src/ol/format/kmlformat.js
+++ b/src/ol/format/kmlformat.js
@@ -2484,7 +2484,9 @@ ol.format.KML.writePolyStyle_ = function(node, style, objectStack) {
  * @private
  */
 ol.format.KML.writeScaleTextNode_ = function(node, scale) {
-  ol.format.XSD.writeDecimalTextNode(node, scale * scale);
+  // the Math is to remove any excess decimals created by float arithmetic
+  ol.format.XSD.writeDecimalTextNode(node,
+      Math.round(scale * scale * 1e6) / 1e6);
 };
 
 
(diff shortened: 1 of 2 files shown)


probins:
PS for a demo of the issues, I went into examples/kml.html and did something like:
```
kml=new ol.format.KML();
f=vector.getSource().getFeatures()[0];
kml.writeFeatures([f],{featureProjection:'EPSG:3857'});
```
This shows the issue with label scale, and most of the coordinates have unwanted junk at the end too.

ahocevar:
I think instead of using toFixed, you could do something like `Math.round(scale * scale * 1e6) / 1e6`

probins:
yes, that would have the same effect.  I can change this if you like. Probably best to add a comment to explain what's going on too. :-)

probins:
though if this formula is also being used for coords based on an option, then ISTM using `toFixed(option)` is simpler than converting option into an exponential.

ahocevar:
> though if this formula is also being used for coords based on an option, then ISTM using toFixed(option) is simpler than converting option into an exponential.

toFixed creates a string from the number, and always adds trailing zeros. Rounding leaves the number a number. Not sure what you mean by "converting option into an exponential".

probins:
converting '6' into '1e6'

ahocevar:
As easy as `Math.pow(10, 6)`

probins:
what about if option = 0, i.e. no decimal places (likely to be the case with meter projections)?

probins:
ok, I suppose that works too with power

ahocevar:
`Math.round(3000.1234 * Math.pow(10, 0)) / Math.pow(10, 0); // == 3000`

probins:
ok, you've convinced me that it's better to do it with Math and keep as a number.

I managed to screw up the PR so this is replace by #5055

PR #5048, opened 2016-03-16
Add tests for ol.dom methods

marcjansen:
This PR adds tests for the methods in `ol.dom`.

To test whether we correctly detect if CSS transforms (2D and 3D) are supported, a custom build of [Modernizr](https://modernizr.com/) is included.

Additionally an unneded and IMO wrong `goog.provide` (for `ol.dom.BrowserFeature`) is removed.

Please review.

### DIFF
--- a/src/ol/dom/dom.js
+++ b/src/ol/dom/dom.js
@@ -1,5 +1,3 @@
-// FIXME add tests for browser features (Modernizr?)
-
 goog.provide('ol.dom');
 goog.provide('ol.dom.BrowserFeature');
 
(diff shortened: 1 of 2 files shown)


ahocevar:
It would have been nice to get rid of `goog.userAgent` in the process of adding these tests. But still, this is a great improvement in test coverage. So please merge when green.

marcjansen:
Cheers for the review, Andreas. The removal of `goog.userAgent` should be tackled as a follow-up

PR #5049, opened 2016-03-16
Supported browsers all have getComputedStyle

marcjansen:
This simplifies the code in `ol.dom` a tiny bit by assuming we always have `window.getComputedStyle`. This method should be there in all supported browsers: http://caniuse.com/#feat=getcomputedstyle

Please review.

### DIFF
--- a/src/ol/dom/dom.js
+++ b/src/ol/dom/dom.js
@@ -36,31 +36,29 @@ ol.dom.canUseCssTransform = (function() {
     if (canUseCssTransform === undefined) {
       goog.asserts.assert(document.body,
           'document.body should not be null');
-      if (!goog.global.getComputedStyle) {
-        // this browser is ancient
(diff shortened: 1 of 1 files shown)


ahocevar:
Nice simplification, but don't you want to continue using `goog.global` instead of `window`?

marcjansen:
> …don't you want to continue using goog.global instead of window?

Good question. I saw that `ol.dom.outerHeight` and `ol.dom.outerWidth` were already using `window` instead of `goog.global`. I can harmonize this the other way around (using `goog.global` everywhere), if you prefer.

ahocevar:
>  I can harmonize this the other way around (using goog.global everywhere), if you prefer.

Yes please, that would be great. And please merge when done and green.

marcjansen:
Thanks for the review.

PR #5050, opened 2016-03-16
Add rotateWithView option to ol.style.Text

### LINKED ISSUE #3005: Text "rotateWithView" Request
Is it possible to add rotateWithView to the text style parameters?
Also a scaleWithView would be very useful for Text as well as Icons.

Currently trying to use vector data to label things like streets is not possible to do well.
I know you might say just use a WMS or OSM tiles but on my project that's not really an option.

fredj:
fixes #3005


fredj:
note to self: try with `map.getRenderer().context_.imageSmoothingEnabled = false`

bartvde:
nice work LGTM please merge

PR #5055, opened 2016-03-16
Fix cruft on kml scale output

probins:
Replacement for #5046

I also added a comment in case anyone in future wonders what this is all about


ahocevar:
Nice, thanks @probins.

PR #5056, opened 2016-03-16
Note in docs that WebGL can render points

probins:
The current docs are a bit out of date, so this update may help answer an FAQ.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -281,7 +281,8 @@ olx.MapOptions.prototype.overlays;
  * Renderer. By default, Canvas, DOM and WebGL renderers are tested for support
  * in that order, and the first supported used. Specify a
  * {@link ol.RendererType} here to use a specific renderer.
- * Note that at present only the Canvas renderer supports vector data.
+ * Note that at present only the Canvas renderer fully supports vector data,
+ * though WebGL can render Point geometries.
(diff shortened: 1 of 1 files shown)


fredj:
thanks

ahocevar:
Post-merge comment: also the DOM renderer has full vector support.

probins:
ah, does it - I must have missed that. Further PR coming up then

probins:
Presumably, canvas is still the recommended renderer, though?

probins:
is there an example using DOM for vectors?

tsauerwein:
> is there an example using DOM for vectors?

E.g. http://openlayers.org/en/master/examples/icon.html?mode=raw&renderer=dom

ahocevar:
> Presumably, canvas is still the recommended renderer, though?

Yes.

probins:
> E.g. http://openlayers.org/en/master/examples/icon.html?mode=raw&renderer=dom

well, yes, but afaik the rendererfromquerystring thing is only used by some examples and isn't documented anywhere, so any newbie is not going to know about it.

In any case, that doesn't help someone who goes to the examples page and searches for 'dom vectors' or even just 'dom'

tsauerwein:
Agree, it would make sense to have an example that directly uses the DOM renderer.

tschaub:
Alternatively, we could remove the DOM renderer.

PR #5058, opened 2016-03-16
Note in docs that DOM renderer can also render vectors

probins:
as discussed in #5056


ahocevar:
Very nice, thanks @probins.

PR #5060, opened 2016-03-17
(WIP) Fire pre/postcompose events on layer groups

tschaub:
Opening this now to get comments.  Not ready for review.

We're relying heavily on `precompose` and `postcompose` layer events for manipulating the canvas context.  For layer groups, it is awkward to achieve the same.  It is possible to register for `precompose` on the first layer in the group and `postcompose` in the last layer, but this has problems.  Layers may not be rendered in group order, so first and last in the group may not be meaningful.  But the bigger problem is that layers with limited extent may go in and out of the rendering extent, meaning that you cannot always rely on `precompose` being fired on the first and `postcompose` on the last (and at times you may only get `precompose` but not `postcompose` or vice versa).

So it would be very convenient if the library dispatched `precompose` before composing any layers in a group and `postcompose` after composing all layers in a group.  Unfortunately, this is awkward to achieve with the way layer groups and layer rendering are handled.  The approach I think I'll take is to add a `parent` property to `ol.layer.LayerState` that is set when `group.getLayerStatesArray()` is called (and it will be a reference to the group for a layer).  Then when dispatching compose events in the layer renderer, we can check for `layerState.parent`.  To determine whether or not to dispatch the event on the parent, we'll need to store additional information on the frame state (so that events aren't dispatched on the parent for every layer in a group).

Any comments on this or alternative implementations are welcome.

### DIFF
--- a/src/ol/layer/layerbase.js
+++ b/src/ol/layer/layerbase.js
@@ -25,6 +25,7 @@ ol.layer.LayerProperty = {
 
 /**
  * @typedef {{layer: ol.layer.Layer,
+ *            parent: (ol.layer.Base|undefined),
  *            opacity: number,
  *            sourceState: ol.source.State,
  *            visible: boolean,
--- a/src/ol/layer/layergroup.js
+++ b/src/ol/layer/layergroup.js
@@ -212,6 +212,7 @@ ol.layer.Group.prototype.getLayerStatesArray = function(opt_states) {
   var i, ii, layerState;
   for (i = pos, ii = states.length; i < ii; i++) {
     layerState = states[i];
+    layerState.parent = this;
     layerState.opacity *= ownLayerState.opacity;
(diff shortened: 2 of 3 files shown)


ahocevar:
I like the proposed approach. And thanks for starting out with tests. Makes it easy to understand.

marcjansen:
Is this discontinued, @tschaub?

tschaub:
It continues to be a pain point for us (registering listeners on all layers in a group, unregistering the old layers and registering the new layers when the group changes, etc.).  So I'd like to put time into making it easier from a user perspective.  Unfortunately, things started to get messy.  I think it had to do with knowing when to fire `postcompose`.  Storing an additional `lastInGroup` property in the layer state feels hackish.

I'll close this for now, but may pick it up again later.

Benjaki2:
👍

PR #5063, opened 2016-03-17
Remove inappropriate information from API docs

drnextgis:
The `base` method is not available.

fixes #5063


ahocevar:
Thanks a lot @drnextgis.

PR #5071, opened 2016-03-18
correct way to determine freehand mode is enabled

drnextgis:
Fix issue https://github.com/openlayers/ol3/issues/5070


ahocevar:
Very solid work. Thanks @drnextgis.

Issue #5067, opened 2016-03-17
Rendering issues with the immediate API


tschaub:
In 3.13.0, it was possible to use `precompose` and `postcompose` along with the immediate rendering API to clip layer rendering to a geometry.  See this example clipping a layer to a polygon: http://jsfiddle.net/yqwyp8md/

![v3 13 0](https://cloud.githubusercontent.com/assets/41094/13852393/6e862298-ec27-11e5-8e26-774840f7e957.gif)

In 3.14.2, the same example shows nothing rendered for the layer: http://jsfiddle.net/s4jrdqdc/

![v3 14 2](https://cloud.githubusercontent.com/assets/41094/13852404/7f4352c2-ec27-11e5-8300-194899fbf6d6.gif)

In master (with the current hosted build of master), the example sort of works.  It renders initially, but then flashes during rotation, and will disappear completely when the map is resized: http://jsfiddle.net/r87meu6c/

![master](https://cloud.githubusercontent.com/assets/41094/13852411/85e4bcce-ec27-11e5-8bf7-591cdc251457.gif)

tschaub:
So 600c139a617cdb7ad24194eb6b40677ad865c631 is the cause for the difference between [3.14.2](http://jsfiddle.net/s4jrdqdc/) and [master](http://jsfiddle.net/r87meu6c/).

tschaub:
This turns out to be because the `precompose` event handler code runs to completion before the "immediate" rendering actually happens (the path isn't drawn until `vectorContext.flush()` is called).  So in the example above, `context.clip()` is getting called before the geometry is drawn.

I don't think we need to make this work.  Instead, I think we could have convenient (and predictable) geometry rendering with `setStyle()` and `drawGeometry()` methods (see #5072).

tschaub:
I'd also be up for revisiting #1473 and making all "immediate" rendering calls synchronous.  This would mean we would say that if you want `zIndex` support, use a vector layer.  Curious what others think.

ahocevar:
Making "immediate" calls synchronous would also mean that users no longer need to listen for both `precompose` and `postcompose` for certain operations. The loss of `zIndex` support should not be a big problem, because users have full control over what gets rendered when as soon as the whole API is synchronous. Or am I missing something?

tschaub:
I agree about `zIndex` support not being a problem.  If you care about `zIndex` you can either order your calls to the immediate API or use a vector layer.

[...]

PR #5072, opened 2016-03-18
Add setStyle() and drawGeometry() to the immediate rendering API

tschaub:
The `vectorContext.drawFeature()` method is a bit of a hazard to use since it doesn't result in an "immediate" rendering.  So, for example, you cannot `context.clip()` after drawing a feature (see #5067).

The other methods are awkward when you are dealing with an arbitrary geometry and style.

Instead of these 11 methods:

 * `drawPointGeometry`
 * `drawLineStringGeometry`
 * `drawPolygonGeometry`
 * `drawMultiPointGeometry`
 * `drawMultiLineStringGeometry`
 * `drawMultiPolygonGeometry`
 * `drawGeometryCollectionGeometry`
 * `drawCircleGeometry`
 * `setFillStrokeStyle`
 * `setImageStyle`
 * `setTextStyle`

I think we could do with two:

 * `setStyle`
 * `drawGeometry`

This branch adds the above two methods.  I'm tempted to remove `@api` from the others in hopes of making a more easily understood API.  But will wait for comments from others.

### DIFF
--- a/src/ol/graticule.js
+++ b/src/ol/graticule.js
@@ -442,11 +442,11 @@ ol.Graticule.prototype.handlePostCompose_ = function(e) {
   var i, l, line;
   for (i = 0, l = this.meridians_.length; i < l; ++i) {
     line = this.meridians_[i];
-    vectorContext.drawLineStringGeometry(line, null);
+    vectorContext.drawLineString(line, null);
   }
   for (i = 0, l = this.parallels_.length; i < l; ++i) {
     line = this.parallels_[i];
-    vectorContext.drawLineStringGeometry(line, null);
+    vectorContext.drawLineString(line, null);
   }
 };
 
--- a/src/ol/render/webgl/webglimmediate.js
+++ b/src/ol/render/webgl/webglimmediate.js
@@ -1,6 +1,8 @@
 goog.provide('ol.render.webgl.Immediate');
+goog.require('goog.asserts');
 goog.require('ol.array');
 goog.require('ol.extent');
+goog.require('ol.geom.GeometryType');
 goog.require('ol.render.VectorContext');
 goog.require('ol.render.webgl.ImageReplay');
 goog.require('ol.render.webgl.ReplayGroup');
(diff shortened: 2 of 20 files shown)


ahocevar:
Now this looks very slick! I'd also be in favour of removing the other 11 methods, since it's an easy upgrade change for users to make if documented well in the release notes.

marcjansen:
The API is obviously cleaner than before, nice. The methods that will be removed were marked as `experimental`, right? If so, I am `+1` on removing them if we add some migration notes.

I first thought to just deprecate them for one month… but I don't think this is necessary. 

For the record, does this also fix the behavior you reported in #5067?

tschaub:
> For the record, does this also fix the behavior you reported in #5067?

The issue reported in #5067 could be fixed by something like  https://github.com/openlayers/ol3/compare/master...tschaub:more-immediate

The changes in this branch just make it convenient to immediately set a style and render a geometry (I was previously calling `vectorContext.drawFeature(feature, style)` for this - but it is not really immediate).

marcjansen:
>> For the record, does this also fix the behavior you reported in #5067?
> The issue reported in #5067 could be fixed by something like master...tschaub:more-immediate

Ahh, OK, I see. Thanks for the clarification.

tschaub:
Ok, this is ready for review.  The changes here make it so the immediate API for vector contexts is:

 * `vectorContext.drawGeometry(geometry)`
 * `vectorContext.setStyle(style)`
 * `vectorContext.drawFeature(feature, style)`

The `drawFeature()` method is not well behaved (because it is not immediate).  I'll address that in https://github.com/openlayers/ol3/compare/master...tschaub:more-immediate after getting this in.

src/ol/graticule.js
```diff
@@ -442,11 +442,11 @@ ol.Graticule.prototype.handlePostCompose_ = function(e) {
   var i, l, line;
   for (i = 0, l = this.meridians_.length; i < l; ++i) {
     line = this.meridians_[i];
-    vectorContext.drawLineStringGeometry(line, null);
+    vectorContext.drawLineString(line, null);
```
marcjansen: Or `drawGeometry`?

src/ol/graticule.js
```diff
@@ -442,11 +442,11 @@ ol.Graticule.prototype.handlePostCompose_ = function(e) {
...
   }
   for (i = 0, l = this.parallels_.length; i < l; ++i) {
     line = this.parallels_[i];
-    vectorContext.drawLineStringGeometry(line, null);
+    vectorContext.drawLineString(line, null);
```
marcjansen: See my question above.

src/ol/render/webgl/webglimmediate.js
```diff
@@ -132,52 +163,35 @@ ol.render.webgl.Immediate.prototype.drawFeature = function(feature, style) {
...
-    render.setTextStyle(style.getText());
-    var type = geometry.getType();
-    var renderGeometry = ol.render.webgl.Immediate.GEOMETRY_RENDERERS_[type];
-    // Do not assert since all kinds of geometries are not handled yet.
-    // In spite, render what we support.
-    if (renderGeometry) {
-      renderGeometry.call(render, geometry, null);
-    }
+    render.setStyle(style);
+    goog.asserts.assert(geometry);
```
marcjansen: We have messages elsewhere when asserting.

marcjansen:
I have added minor questions and one comment, but I have only been looking at the diff, so this was no full review.

I really like this, though.

src/ol/graticule.js
tschaub: We can still use the unexported methods internally.  This is akin to us exporting `observable.on()` and using `ol.events.listen()` internally.  This gives us stricter type checking internally.

src/ol/render/webgl/webglimmediate.js
tschaub: I'll add one.  I find the line numbers more useful.  But I know the message can be comforting.

src/ol/graticule.js
tschaub: See also https://github.com/openlayers/ol3/pull/5072#discussion_r56730389.

ahocevar:
Looks great to me as well. Please merge.

tschaub:
Thanks for the reviews.

PR #5074, opened 2016-03-18
Do not rotate map canvas after composition

### LINKED ISSUE #5004: updated 3.12.1 => 3.14.2 and got blurry view
Before:

After:

As you can see both tile layer and vector labels layer are blurry. What can cause such problem?
Now I'm trying to find out which version is causing the problem.

### LINKED ISSUE #5067: Rendering issues with the immediate API

ahocevar:
This requires to draw tile layers to an intermediate canvas again, but only when the view is rotated.

Fixes #5004.
Fixes #5067.


ahocevar:
@tschaub, you might want to take a look at this, because it fixes #5067 and brings back the rendering quality from v3.13.1, while keeping things as simple as previously in master.

tschaub:
Checkin it.

tschaub:
Rendering looks really nice.

tschaub:
This looks great.  The rendering tests are helpful.  The examples I tested work well.  Please merge.

PR #4689, opened 2016-01-19
Revert "Do not render in handleDownEvent"

fredj:
Reverts openlayers/ol3#4581

See #4675

Issue #5053, opened 2016-03-16
Map is not always updated when tiles are loaded


ahocevar:
This is the remaining issue mentioned in https://github.com/openlayers/ol3/pull/4964#issuecomment-197228158.

To reproduce: Set `cacheSize` to 20 in the "simple" example. Open the example, and heavily zoom in and out with wheel zoom. When you stop zooming, only lower resolution tiles are displayed.

I did a git bisect, and the first bad commit is fc2c9176a6eb5f6d0eb7f1dada459c3537309464.

ahocevar:
ping @fredj.

ahocevar:
@fredj, any chance you could take another look at this? Thanks.

PR #5078, opened 2016-03-18
(WIP) Remove unnecessary map.render() calls

tschaub:
We're forcing extra map renders that are masking other problems.  For example, #5067 is only solved because we have extra render calls at the end of interactions.

I think we'd be better off seeing what real problems are revealed when we get rid of the unnecessary `map.render()` calls.

Prior art:

 * #4581
 * #4689
 * #4675

### DIFF
--- a/src/ol/interaction/dragpaninteraction.js
+++ b/src/ol/interaction/dragpaninteraction.js
@@ -87,7 +87,6 @@ ol.interaction.DragPan.handleDragEvent_ = function(mapBrowserEvent) {
     ol.coordinate.rotate(center, viewState.rotation);
     ol.coordinate.add(center, viewState.center);
     center = view.constrainCenter(center);
-    map.render();
     view.setCenter(center);
   }
   this.lastCentroid = centroid;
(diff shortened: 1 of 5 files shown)


tschaub:
With the `map.render()` calls removed from the drag pan and drag rotate interactions, I cannot reproduce #4674.  I'd be curious if @fredj or others could.

fredj:
> With the `map.render()` calls removed from the drag pan and drag rotate interactions, I cannot reproduce #4674.  I'd be curious if @fredj or others could.

I can reproduce the issue with the pan interaction but only if the kinetic effect does not occur.

fredj:
> I can reproduce the issue with the pan interaction but only if the kinetic effect does not occur.

See https://github.com/fredj/ol3/commit/8c99fd596d06d67fb5fcccf1337931efe4768320 for a fix

ahocevar:
This might also be related to #5053. Curious if there is a way to make things work without code like https://github.com/fredj/ol3/commit/8c99fd596d06d67fb5fcccf1337931efe4768320.

gjn:
Can we expect a merge soon?

bartvde:
Just talked to @ahocevar and decided there is never a good time for this type of change, so we will just go ahead and put this change in. I'll resolve conflicts in a new branch based on this one, since I assume @tschaub is not in a position to do now.

bartvde:
closed with #5764 thanks @tschaub

PR #5080, opened 2016-03-19
Fix geojson write GeometryCollection

probins:
I tested my changes mentioned in #3953 on a GeometryCollection, and found they didn't work. However, when I tried doing the same with the current master, I found that didn't work either. examples/geojson has a GeometryCollection so if you do something like
```
format=new ol.format.GeoJSON();
format.writeFeature(vectorSource.getFeatures()[6], {featureProjection:'EPSG:3857'});
```
you can see that the coordinates are all screwy. The reason for this is that `writeGeometry_` is getting run for the Collection as a whole and then for each individual geometry. `writeGeometry_` runs `ol.format.Feature#transformWithOptions()`, so the coordinates for each geometry are transformed twice. This PR calls write for the individual geoms without the projection options, so the 2nd transform doesn't occur.

From the fact that no-one's noticed this before, I conclude that GeometryCollections aren't used very much. :-) I think tests should be added for this case, but would prefer to include that in my fixes for #3953 so the rounding can be tested as well.

### DIFF
--- a/src/ol/format/geojsonformat.js
+++ b/src/ol/format/geojsonformat.js
@@ -211,7 +211,9 @@ ol.format.GeoJSON.writeGeometryCollectionGeometry_ = function(
   goog.asserts.assertInstanceof(geometry, ol.geom.GeometryCollection,
       'geometry should be an ol.geom.GeometryCollection');
   var geometries = geometry.getGeometriesArray().map(function(geometry) {
-    return ol.format.GeoJSON.writeGeometry_(geometry, opt_options);
+    var options = ol.object.assign({}, opt_options);
+    delete options.featureProjection;
+    return ol.format.GeoJSON.writeGeometry_(geometry, options);
   });
   return /** @type {GeoJSONGeometryCollection} */ ({
     type: 'GeometryCollection',
--- a/test/spec/ol/format/geojsonformat.test.js
+++ b/test/spec/ol/format/geojsonformat.test.js
@@ -725,6 +725,28 @@ describe('ol.format.GeoJSON', function() {
           .to.be.lessThan(0.0000001);
     });
 
+    it('transforms and encodes geometry collection', function() {
+      var collection = new ol.geom.GeometryCollection([
+        new ol.geom.Point([2, 3]),
+        new ol.geom.LineString([[3, 2], [2, 1]])
+      ]);
+      var geojson = format.writeGeometry(collection, {
+        featureProjection: 'EPSG:3857'
+      });
+      var got = format.readGeometry(geojson, {
+        featureProjection: 'EPSG:3857'
+      });
+      var gotGeometries = got.getGeometries();
+      var geometries = collection.getGeometries();
+      expect(geometries[0].getCoordinates()[0]).to.roughlyEqual(
+          gotGeometries[0].getCoordinates()[0], 1e-8);
+      expect(geometries[0].getCoordinates()[1]).to.roughlyEqual(
+          gotGeometries[0].getCoordinates()[1], 1e-8);
+      expect(geometries[1].getCoordinates()[0][0]).to.roughlyEqual(
+          gotGeometries[1].getCoordinates()[0][0], 1e-8);
+      expect(geometries[1].getCoordinates()[0][1]).to.roughlyEqual(
+          gotGeometries[1].getCoordinates()[0][1], 1e-8);
+    });
   });
 
 });


src/ol/format/geojsonformat.js
```diff
@@ -211,7 +211,10 @@ ol.format.GeoJSON.writeGeometryCollectionGeometry_ = function(
   goog.asserts.assertInstanceof(geometry, ol.geom.GeometryCollection,
       'geometry should be an ol.geom.GeometryCollection');
   var geometries = geometry.getGeometriesArray().map(function(geometry) {
-    return ol.format.GeoJSON.writeGeometry_(geometry, opt_options);
+    var options = (opt_options && opt_options.rightHanded) ? {
+      rightHanded: opt_options.rightHanded
+    } : undefined;
+    return ol.format.GeoJSON.writeGeometry_(geometry, options);
```
ahocevar: Probably safer to do something like
```js
var options = ol.object.assign({}, opt_options);
delete options.featureProjection;
return ol.format.GeoJSON.writeGeometry_(geometry, options);
```

probins: shorter too :-) ok, will change this

ahocevar:
> I think tests should be added for this case, but would prefer to include that in my fixes for #3953 so the rounding can be tested as well.

I'd feel better if you added a test for this pull request too. You can use `exepct(1.1).to.roughlyEqual(1.10001, 0.0001);`.

marcjansen:
What exactly would that test?

tschaub:
> What exactly would that test?

I think @ahocevar is just saying that you can use the `roughlyEqual()` matcher in testing (e.g. `expect(point.getCoordinates()[0]).to.roughlyEqual(10, 1e-4)`).

Thanks for catching this and providing a fix @probins.  I agree we should have tests before merging.

src/ol/format/geojsonformat.js
```diff
@@ -211,7 +211,9 @@ ol.format.GeoJSON.writeGeometryCollectionGeometry_ = function(
   goog.asserts.assertInstanceof(geometry, ol.geom.GeometryCollection,
       'geometry should be an ol.geom.GeometryCollection');
   var geometries = geometry.getGeometriesArray().map(function(geometry) {
-    return ol.format.GeoJSON.writeGeometry_(geometry, opt_options);
+    var options = ol.object.assign({}, opt_options);
+    delete options.featureProjection;
```
tschaub: Would it make sense to delete `options.dataProjection` as well?

probins:
> I think @ahocevar is just saying that you can use the roughlyEqual() matcher in testing (e.g. expect(point.getCoordinates()[0]).to.roughlyEqual(10, 1e-4)).

where is `roughlyEqual` defined? I couldn't find any docs

src/ol/format/geojsonformat.js
probins: that's not necessary. If `featureProjection` isn't set, no transform is performed.

marcjansen:
https://github.com/openlayers/ol3/blob/master/test/test-extensions.js

probins:
ok. Basic test added, based on the Point test above.

test/spec/ol/format/geojsonformat.test.js
```diff
@@ -725,6 +725,32 @@ describe('ol.format.GeoJSON', function() {
...
+      });
+      var got = format.readGeometry(geojson, {
+        featureProjection: 'EPSG:3857'
+      });
+      var gotGeometries = got.getGeometries();
+      var geometries = collection.getGeometries();
+      expect(geometries[0].getCoordinates()[0]).to.roughlyEqual(
+          gotGeometries[0].getCoordinates()[0], 1e-8);
+      expect(
+          Math.abs(geometries[0].getCoordinates()[1] -
```
ahocevar: You can also use roughlyEquals here.

test/spec/ol/format/geojsonformat.test.js
```diff
@@ -725,6 +725,32 @@ describe('ol.format.GeoJSON', function() {
...
+          gotGeometries[0].getCoordinates()[1]))
+          .to.be.lessThan(0.0000001);
+      expect(geometries[1].getCoordinates()[0][1]).to.roughlyEqual(
+          gotGeometries[1].getCoordinates()[0][1], 1e-8);
+      expect(
+          Math.abs(geometries[1].getCoordinates()[0][0] -
```
ahocevar: And here.

test/spec/ol/format/geojsonformat.test.js
probins: yeah, I was wondering why the Point test wasn't using roughlyEquals for both coordinates

ahocevar: roughlyEqual is just for numbers, so you have to test x and y of a coordinate separately or in a forEach function.

probins: the 'transforms and encodes a point' test is using roughlyEqual for the x coord and this Math.abs formula for the y. I'm wondering why.

probins:
test changed as suggested, so this tests a Point and the first coordinate of a LineString within a GeometryCollection, which should be sufficient.

I'll submit another PR to change the Point test to use roughlyEqual for both coords - might as well make these things consistent.

ahocevar:
> I'll submit another PR to change the Point test to use roughlyEqual for both coords - might as well make these things consistent.

Thanks. I think a `roughlyEqualCoordinate` method could also make sense.

Issue #3850, opened 2015-06-25
Unable to snap only on vertices or edges.


dufop:
In OL2 there was options to activate the snapping only on vertices or edges and it would be great to have those options in OL3:
vertex {Boolean} Snap to vertices in target layer. Default is true.
edge {Boolean} Snap to edges in target layer. Default is true.

AndreasTheMonkey:
I Agree!  Please add this feature.

ahocevar:
Pull request anyone?

fredj:
#5081

Issue #4177, opened 2015-09-28
OL3 (ol.interation.modify)


MAHESHGONDA:
i am using modify interaction , but i am not able to move only vertices(in Line String feature) while modifying the feature.

please tell me how to move only vertices using ol.interation.modify or any other API in OL3

thanks a lot in advance.

PR #5081, opened 2016-03-19
Snap to vertices or edges

drnextgis:
Fix for issue #3850

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3098,7 +3098,9 @@ olx.interaction.SelectOptions.prototype.wrapX;
  * @typedef {{
  *     features: (ol.Collection.<ol.Feature>|undefined),
  *     pixelTolerance: (number|undefined),
- *     source: (ol.source.Vector|undefined)
+ *     source: (ol.source.Vector|undefined),
+ *     edge: (boolean|undefined),
+ *     vertex: (boolean|undefined)
  * }}
  * @api
  */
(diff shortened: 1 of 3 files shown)


drnextgis:
Demo: http://jsfiddle.net/_DR_/vjxzyqx5/

ahocevar:
Thanks for this, great work! Can you please either add upgrade notes that snapping to edges is now active, or set edge snapping to false by default?

drnextgis:
Maybe I misunderstood you but if I set edge snapping to false by default it will change current `ol.interaction.Snap` default behaviour.

ahocevar:
@drnextgis Sorry, I understood you! I was under the impression that we currently have only vertex snapping, which is not true of course.

drnextgis:
In API docs about `ol.interaction.Snap`:

> Handles snapping of vector features while modifying or drawing them.

But I've found that `ol.interaction.Modify` has own snapping part and `ol.interaction.Snap` doesn't affect it. Any thoughts how we can fix it? Related issue https://github.com/openlayers/ol3/issues/4177

ahocevar:
I don't think #4177 is related. If I understand that issue correctly, it is about restricting the Modify interaction to disallow the addition of new vertices.

Regarding this issue, the Modify interaction has no snapping. What feels like snapping is only the positioning of the edit vertex and the decision whether to place it on an existing vertex or on an edge. Once you move that vertex, there is no snapping whatsoever. You can influence that behaviour with the `pixelTolerance` option.

PR #5082, opened 2016-03-19
Get rid of goog.log

bjornharrtell:
This PR suggests to remove `goog.log` usage in the quest for closure library independence.

From what I can find `goog.log` was only used in `src/ol/webgl/context.js` and in those instances duplicated with almost identical assertions.

### DIFF
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -7,12 +7,9 @@ goog.provide('ol.MapProperty');
 
 goog.require('goog.asserts');
 goog.require('goog.async.nextTick');
-goog.require('goog.debug.Console');
 goog.require('goog.dom');
 goog.require('goog.dom.classlist');
 goog.require('goog.functions');
-goog.require('goog.log');
-goog.require('goog.log.Level');
 goog.require('goog.style');
 goog.require('goog.vec.Mat4');
 goog.require('ol.Collection');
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks @bjornharrtell.

gberaudo:
It is an improvement to remove the `goog.log` mechanism since it is hardly used in the code.
However, why remove the WebGL error detection code?

bartvde:
@gberaudo are they not covered by the assertions?

gberaudo:
@bartvde, with this change the application will fail in debug mode but we will no longer know why.

bjornharrtell:
The assertions seem to do the same detection @gberaudo? I guess you mean that info from `getShaderInfoLog` and `getProgramInfoLog` is lost, something I didn't think through when proposing the changes. I'll try to find a solution (debug conditional + console.log?).

gberaudo:
@bjornharrtell, that would be great.

tschaub:
@bjornharrtell I don't think we need conditional `console.log()` calls.  Instead, the `assert()` messages could be made more informative.  E.g.

```js
    goog.asserts.assert(
        gl.getShaderParameter(shader, goog.webgl.COMPILE_STATUS) ||
        gl.isContextLost(),
        gl.getShaderInfoLog(shader) || 'illegal state, shader not compiled or context lost');
```

I added the guard there since I do not know if `gl.getShaderInfoLog(shader)` always returns a message in the above two cases.

bjornharrtell:
Thanks for the detailed suggestion @tschaub, makes sense. I created PR #5098 to reintroduce the log messages in the manner you propose.

PR #5083, opened 2016-03-19
Remove unnecessary lineTo calls in immediate polygon rendering

tschaub:
This removes unnecessary `lineTo()` calls in ring drawing in the canvas immediate API.

As the [current tests](https://github.com/openlayers/ol3/blob/309843b53d9317c5f9cec2ba1d40fb54bf7a5b01/test/spec/ol/render/canvasimmediate.test.js#L216-L219) demonstrate, we do [two unnecessary](https://github.com/openlayers/ol3/blob/309843b53d9317c5f9cec2ba1d40fb54bf7a5b01/test/spec/ol/render/canvasimmediate.test.js#L230-L233) `lineTo()` calls to the last point of the ring before calling `closePath()`.  Instead, we can omit the `lineTo()` call to the last point altogether (since it should be the same as the first point) and just call `closePath()`.

The rendering tests demonstrate that things work as expected.  And the unit tests are updated to avoid making assertions about the unnecessary `lineTo()` calls.

This means that we are not tolerant of open rings in the immediate API.  We could always do a `lineTo()` to the last point in a ring and then call `closePath()` if we want to tolerate open rings.


ahocevar:
This is great. Couldn't the canvas replay get the same optimisation?

Please merge.

PR #5085, opened 2016-03-20
DEVELOPING.md: add note on running Travis on development branches

probins:
just ran into this little issue :-)

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -17,7 +17,10 @@ Then every time you push to your fork, the test suite will be run. This means
 errors can be caught before creating a pull request. For those making
 small or occasional contributions, this may be enough to check that your contributions
 are ok; in this case, you do not need to install the build tools on your local environment
-as described below.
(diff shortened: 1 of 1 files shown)


ahocevar:
You can either run `make ci` to run the same tests that Travis runs, or you can create a pull request. Once you have a pull request, Travis will check your branch.

That said, I do not think it is a good idea to advise devs to change `.travis.yml`, especially when they are newcomers and not so familiar with git to perform tasks like squash or amend.

probins:
yes, but the point is that this file is out of date. Travis will no longer run on your fork (unless you always push to the master branch). Either DEVELOPER.md needs to be changed as I suggest, or this section about running Travis on your fork should be removed. The purpose of this section is that you can use Travis to run the ci without installing anything locally.

ahocevar:
Ah I see. Maybe just say that Travis will check the branch as soon as the pull request is submitted, or really remove the section.

probins:
replaced by #5088

PR #5086, opened 2016-03-20
Get rid of goog.functions

bjornharrtell:
This PR suggests to remove `goog.functions` usage in the quest for closure library independence.

The reusable functions returning `true` or `false` have been ported to `ol.functions`. Other use of `goog.functions` had only a single call site and has been replaced by local code.

### DIFF
--- a/src/ol/functions.js
+++ b/src/ol/functions.js
@@ -1,0 +1,17 @@
+goog.provide('ol.functions');
+
+/**
+ * Always returns true.
+ * @returns {boolean} true.
+ */
+ol.functions.TRUE = function() {
+  return true;
+};
+
+/**
+ * Always returns false.
+ * @returns {boolean} false.
+ */
+ol.functions.FALSE = function() {
+  return false;
+};
--- a/src/ol/style/atlasmanager.js
+++ b/src/ol/style/atlasmanager.js
@@ -2,9 +2,7 @@ goog.provide('ol.style.Atlas');
 goog.provide('ol.style.AtlasManager');
 
 goog.require('goog.asserts');
-goog.require('goog.functions');
 goog.require('ol');
-
 
 /**
  * Provides information for an image inside an atlas manager.
@@ -186,7 +184,7 @@ ol.style.AtlasManager.prototype.add = function(id, width, height,
   // the hit-detection atlas, to make sure that the offset is the same for
   // the original image and the hit-detection image.
   var renderHitCallback = opt_renderHitCallback !== undefined ?
-      opt_renderHitCallback : goog.functions.NULL;
+      opt_renderHitCallback : ol.nullFunction
 
   /** @type {?ol.style.AtlasInfo} */
   var hitInfo = this.add_(true,
--- a/src/ol/events/condition.js
+++ b/src/ol/events/condition.js
@@ -2,7 +2,7 @@ goog.provide('ol.events.ConditionType');
 goog.provide('ol.events.condition');
 
 goog.require('goog.asserts');
-goog.require('goog.functions');
+goog.require('ol.functions');
 goog.require('ol.MapBrowserEvent.EventType');
 goog.require('ol.MapBrowserPointerEvent');
 
@@ -59,7 +59,7 @@ ol.events.condition.altShiftKeysOnly = function(mapBrowserEvent) {
  * @function
  * @api stable
  */
-ol.events.condition.always = goog.functions.TRUE;
+ol.events.condition.always = ol.functions.TRUE;
 
 
 /**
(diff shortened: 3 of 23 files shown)


nicholas-l:
Haha awkward... #5084 should I close my one?

bjornharrtell:
Hah, yep. :) But there is actually value in this duplicated effort. I spot at least three things:

1. Good idea to have test cases, though they should be finished.
2. You are missing type annotations for some of the functions in `ol.functions`.
3. Ported code differs (at least in annotation). Should make sure code is ported from the correct closure library version (closure-util indicates version dated 20160208).

nicholas-l:
Yes, I am a bit new to type annotations.
One thing I did note was arguments usage, I am a bit unsure as to if this gets de-optimized. However  it is noted in [V8 Optimization Killers](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#what-is-safe-arguments-usage) that it can be if the function is bound and apply is used. Thats why I added the copying of arguments. Thoughts?

src/ol/functions.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+/**
+ * Creates a function that calls the functions passed in in sequence, and
+ * returns the value of the last function. For example,
+ * (ol.functions.sequence(f, g))(x) is equivalent to f(x),g(x).
+ * @param {...Function} var_args A list of functions.
+ * @return {!Function} A function that calls all inputs in sequence.
+ */
+ol.functions.sequence = function(var_args) {
```
tschaub: The one place this is used, we want to provide an array of functions.  I'd be tempted to use a `for` loop at the call site rather than define this and call it with `apply`.

src/ol/functions.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Creates a function that returns true if each of its components evaluates
+ * to true. The components are evaluated in order, and the evaluation will be
+ * short-circuited as soon as a function returns false.
+ * For example, (ol.functions.and(f, g))(x) is equivalent to f(x) && g(x).
+ * @param {...Function} var_args A list of functions.
+ * @return {function(...?):boolean} A function that ANDs its component
+ *      functions.
+ */
+ol.functions.and = function(var_args) {
```
tschaub: Also only one call site for this.  Again tempting just to return an anonymous function where we use it.

sweco-sebhar:
@nicholas-l, I think we got good input (thanks @tschaub) to make a final effort on removing goog.functions usage. Do you want to do the honors or is it up to me? :)

nicholas-l:
Probably going to run out of time, grab anything from my pull request if you like

bjornharrtell:
I've made a new effort on this now and updated the PR description to the current status. I've replaced single code site usage with local code as recommended by @tschaub.

nicholas-l:
Looks great and now it doesn't look like it needs any tests added

bjornharrtell:
Rebased and ready for review. Ping @tschaub.

src/ol/style/atlasmanager.js
```diff
@@ -186,7 +184,9 @@ ol.style.AtlasManager.prototype.add = function(id, width, height,
   // the hit-detection atlas, to make sure that the offset is the same for
   // the original image and the hit-detection image.
   var renderHitCallback = opt_renderHitCallback !== undefined ?
-      opt_renderHitCallback : goog.functions.NULL;
+      opt_renderHitCallback : function() {
+        return null;
+      };
```
marcjansen: Can we use `ol.nullFunction` here?

marcjansen:
I really like this. I added one minor comment / question.

Plus one for merging.

A follow-up PR might move [`ol.nullFunction`](https://github.com/openlayers/ol3/blob/master/src/ol/ol.js#L274) into the `ol.function` namespace.

Thanks for the work on this,  @bjornharrtell and @nicholas-l

bjornharrtell:
Thanks @marcjansen, wasn't aware of `ol.nullFunction`, updated branch to use it as suggested.

marcjansen:
Thanks. Can you rebase again, so I can merge this, @bjornharrtell?

bjornharrtell:
Rebased. Is it just me or was GitHub more lenient before, not requiring rebasing of out of date branches if they did not have merge conflicts?

marcjansen:
> Is it just me or was GitHub more lenient before, not requiring rebasing of out of date branches if they did not have merge conflicts?

It's not you alone… I think the same. It isn't that bad for everyone, though.

nicholas-l:
Could be to do with [this](https://help.github.com/articles/about-required-status-checks/)

bjornharrtell:
Ah nice find @nicholas-l. Should perhaps be discussed by the core devs. Problem is I don't know which forum, if any, to use to reach the core devs except here. @marcjansen any thoughts?

marcjansen:
Mhhm… maybe the dev mailing list or a dedicated issue here with a good rationale and some suggestions.

In this ticket here, the context is obviously wrong :wink:

probins:
Github have just changed the way this works when you have status checks https://github.com/blog/2137-protected-branches-improvements

PR #5087, opened 2016-03-21
assertion to check that there's a DOM node for target element

drnextgis:
Fix for issue https://github.com/openlayers/ol3/issues/2974


ahocevar:
I was confused when I read your change first. Wouldn't something like
```js
var targetElement;
if (this.getTarget()) {
  targetElement = this.getTargetElement();
  goog.asserts.assert(targetElement !== null,
      'expects a non-null value for targetElement');
}
```
be easier to read?

drnextgis:
Yes, pretty easier. I've updated PR.

PR #3959, opened 2015-08-04
Minor updates to release procedure

tschaub:
This makes sure the latest dependencies are installed before building release artifacts.  In addition, the `travis/push` job is configured to only build pushes to `master` (while this may look like a confusing change to `.travis.yml`, it doesn't affect the `travis/pr` job).

PR #5088, opened 2016-03-21
DEVELOPING.md: remove section on running Travis on fork

probins:
no longer works since #3959. Replacement for #5085


ahocevar:
Thanks @probins.

PR #5089, opened 2016-03-21
Test for both transform coords should use roughlyEqual

probins:
as mentioned in https://github.com/openlayers/ol3/pull/5080#issuecomment-199249328


ahocevar:
Thanks!

PR #5090, opened 2016-03-21
Update custom-builds.md

### LINKED ISSUE #4351: ol.ext install should be documented
It's a while since I created a new build, and I've just run into a problem trying to do so. I got a dependency error that it couldn't find ol.ext/pbf. AFAICS this was because I'd upgraded to latest node packages using `npm update`; `build/ol.ext` is however created with an install script which isn't run when you do `npm update`.

AFAICS, `ol.ext` isn't documented anywhere, so I think as a minimum a section on upgrading to the latest package versions should be added to the custom build tutorial. If my comments above are right, then this should specify that people should always use `npm install` to update, not `npm update`.

probins:
Fixes #4351. While I was at it, I also added a note on minimum versions for Node/Java

### DIFF
--- a/doc/tutorials/custom-builds.md
+++ b/doc/tutorials/custom-builds.md
@@ -13,7 +13,7 @@ This particular tutorial explains how to create custom builds of OpenLayers 3.
 
 ## Requirements
 
-OpenLayers 3's build tools use Node and Java, so you need to have Node and Java installed on your machine. You can run `node --version` and `java -version` to test that Node and Java are installed, respectively.
+OpenLayers 3's build tools use Node and Java, so you need to have Node and Java installed on your machine. You can run `node --version` and `java -version` to test that Node and Java are installed, respectively. See [developing guide](https://github.com/openlayers/ol3/blob/master/DEVELOPING.md) for minimum version numbers required.
 
 ## Download OpenLayers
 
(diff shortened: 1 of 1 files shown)


doc/tutorials/custom-builds.md
```diff
@@ -223,6 +223,10 @@ The exports are given here in the order in which they occur in the `heatmaps-ear
 
 As this example uses a vector layer it is necessary to remove `"ol.ENABLE_VECTOR=false"` in the `define` section of the configuration.
 
+## Maintaining the code
+
+If you installed OpenLayers from the Node package, you can use `npm` to upgrade to the latest version. If you cloned the Github repo, simply pulling in the latest code may not be enough, as some of the packages used, for example, the compiler, may need upgrading too. Do this by using `npm install` rather than `npm update`.
```
ahocevar: This document only talks about installing OpenLayers from the npm package, so the information about the cloned repo is misleading. All that'd be needed to get the latest would be to run `npm install openlayers`.

probins: the download section at the top says you can clone, download a release archive or install node package

probins: I think the fact that you have to run `npm install` however you download needs documenting, and I don't see where else you can document it other than this doc

ahocevar: Agreed. My suggestion was to add a very small section:
```
## Maintaining the code

To update to the latest version of OpenLayers, simply run `npm install openlayers` again.
```

probins: If you cloned the repo, you don't want to install openlayers, as you already have it. You probably want to go into your ol dir and then you probably want `npm install --production`.

ahocevar: > If you cloned the repo [...]

then you'd already have read `DEVELOPING.md`.

ahocevar:
Oh wait, the main `README.md` also mentions cloning the repo. Reconsidering.

PR #5094, opened 2016-03-22
FAQ: answer question 4

probins:
Just read through the FAQ, and found that the answer to question 4 doesn't actually answer it. :-) So here's an attempt to do so succinctly.


ahocevar:
Very good explanation. Thanks @probins.

PR #5098, opened 2016-03-22
Reintroduce log messages for relevant WebGL assertions

bjornharrtell:
Was removed as part of #5082 but they have potential value in debugging efforts so this PR reintroduces them.


gberaudo:
Thanks @bjornharrtell, it looks good to me.

ahocevar:
Thanks @bjornharrtell, and thanks @tschaub for suggesting this approach.

PR #5100, opened 2016-03-22
assert whether radius is a number in setRadius

3XE:
If a number-like string is accidentally passed to this function strange things can happen because of JavaScript's type coercion.  I.e. string concatenation is performed instead of addition at the assignment of this.flatCoordinates[this.stride].

-Peter


ahocevar:
We decided to not add assertions for types that are clearly type-documented arguments in JSDoc.

3XE:
Fair enough.  Is there any other PR which can make this better because currently it does not error but just produces a ~10m radius?  I could imagine a lot of people getting caught out by this.

PR #5106, opened 2016-03-23
Get rid of goog.math.Vec2

fredj:


ahocevar:
Very nice. Please merge.

fredj:
Thanks for the review

PR #5107, opened 2016-03-23
Fix Region growing example

drnextgis:
Fix issue https://github.com/openlayers/ol3/issues/5075

### DIFF
--- a/examples/region-growing.js
+++ b/examples/region-growing.js
@@ -10,16 +10,34 @@ goog.require('ol.source.Raster');
 function growRegion(inputs, data) {
   var image = inputs[0];
   var seed = data.pixel;
+  var rotation = data.rotation;
+  var center = data.center;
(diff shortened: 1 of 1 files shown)


fredj:
@tschaub can you please take a look at this? thanks

tschaub:
Thanks for the effort here.  I'm concerned that there is a bug lurking though.  I haven't dug yet, but wouldn't you expect `map.getPixelFromCoordinate(coordinate)` to provide a useful pixel coordinate even when rotated?

drnextgis:
Yes, fetching useful pixel coordinate even when rotated is expected behavior of `getPixelFromCoordinate`. Current behavior is not obvious and user have to recalculate pixels manually.

simonseyock:
I am looking into this now. A proposed solution is a `getRenderedPixelFromCoordinate` at the layer.

PR #5111, opened 2016-03-23
Remove CSS filter in examples

fredj:
Was used to set the opacity for IE <= 8


ahocevar:
Nice. Please merge when green.

fredj:
Thanks for the review

Issue #3200, opened 2015-02-03
Error on ol.format.KML().writeFeatures() for geometryCollection


dennisstam:
Following the WKT example from the latest OL3.1.1 release, I was trying to batch convert a large number of WKT's to KML. However I noticed that the GEOMETRYCOLLECTION's are failing for some reason. Could it be that the writeFeatures() was never implemented for that subset of features?

Example:
```javascript
[...]

PR #5112, opened 2016-03-23
Add decimals option to format write

### LINKED ISSUE #3953: Option to specify precision of coordinate values on format write

probins:
Fixes #3953.

I have tested output in GeoJSON, KML and GPX for Point, LineString, Polygon and GeoJSON GeometryCollection (KML output for GC currently fails (#3200) and GPX ignores Polygons and GC). I've added a couple of simple tests for GeoJSON, though, as stated in 3953, I'm not sure what makes sense here. As only featureformat is being changed, all format types should work in the same way.

### DIFF
--- a/src/ol/format/featureformat.js
+++ b/src/ol/format/featureformat.js
@@ -71,6 +71,9 @@ ol.format.Feature.prototype.adaptOptions = function(options) {
           options.dataProjection : this.defaultDataProjection,
       rightHanded: options.rightHanded
     };
+    if (options.decimals) {
+      updatedOptions.decimals = options.decimals;
+    }
   }
   return updatedOptions;
 };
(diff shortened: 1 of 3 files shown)


src/ol/format/featureformat.js
```diff
@@ -165,21 +169,47 @@ ol.format.Feature.transformWithOptions = function(
...
           write ? dataProjection : featureProjection);
     }
   } else {
-    return geometry;
+    transformed = geometry;
+  }
+  if (write && opt_options && opt_options.decimals) {
+    var power = Math.pow(10, opt_options.decimals);
+    // if decimals option on write, round each coordinate appropriately
+    if (Array.isArray(transformed)) {
```
ahocevar: You could simplify this significantly by using `ol.geom.Geometry#applyTransform`:
```js
var power = Math.pow(10, opt_options.decimals);
function transform(coordinates) {
  for (var i = 0, ii = coordinates.length; i < ii; ++i) {
    coordinates[i] = Math.round(coordinates[i] * power / power);
};
if (Array.isArray(transformed)) {
  transform(transformed);
} else {
  transformed.applyTransform(transform);
}

Comment (probins) [inline src/ol/format/featureformat.js]: yes, that would be simpler, but I'm having trouble with the typedefs. If I just use your code, I get
```
actual parameter 1 of transformed.applyTransform does not match formal parameter
ERR! compile found   : function (?): undefined
ERR! compile required: function ((Array<number>|null), (Array<number>|null)=, number=): (Array<number>|null)
```
Presumably I need to add a `@param` to `transform()` but can't work out what form that should take to match the `TransformFunction` type.

Comment (ahocevar) [inline src/ol/format/featureformat.js]: Probably enough to annotate the function with
```js
/**
 * @param {Array.<number>} coordinates Coordinates.
 * @return {Array.<number>} Transformed coordinates.
 */
function transform(coordinates) {
  for (var i = 0, ii = coordinates.length; i < ii; ++i) {
    coordinates[i] = Math.round(coordinates[i] * power / power);
  }
  return coordinates;
};
```

probins: that compiles if I cast `transformed` to `ol.Extent` before calling `transform()`. Thanks, I'll test some more and repush.

probins:
ok, changed as suggested with amendments so it passes (a) linter, (b) compiler (and works properly)

src/ol/format/featureformat.js
```diff
@@ -165,21 +168,46 @@ ol.format.Feature.transformWithOptions = function(
...
+     */
+    var transform = function(coordinates) {
+      for (var i = 0, ii = coordinates.length; i < ii; ++i) {
+        coordinates[i] = Math.round(coordinates[i] * power) / power;
+      }
+      return coordinates;
+    };
+    if (Array.isArray(transformed)) {
+      transformed = /** @type {ol.Extent} */ (transformed);
+      transform(transformed);
```
ahocevar: Super minor, but the above two lines could be simplified to
```js
transform(/** @type {ol.Extent} */ (transformed));
```

ahocevar:
Thanks for your continued effort on this @probins. I made one more minor comment. Let me know if you don't have time to address it. Then I'll merge as-is and create a follow-up pull request myself.

probins:
also changed; Travis seems to like it.

ahocevar:
Thanks @probins .

PR #5113, opened 2016-03-24
Memory leak in ol.control.FullScreen

### LINKED ISSUE #5110: Memory leak in ol.control.FullScreen
The full-screen control attaches an event listener to the goog.global.document. This event handler is not cleaned up when map is removed with `map.setTarget(null); map = null;`.

AvatharDG:
Fixes #5110.

Register and unregister listener in setMap method.


ahocevar:
Very elegant fix. Thanks @AvatharDG. Can you please submit a [CLA](https://github.com/openlayers/cla) so we can merge this?

AvatharDG:
No problem @ahocevar. CLA submitted.

PR #5116, opened 2016-03-24
Improve examples pages on mobile devices a bit

ahocevar:
Only show navigation when there is enough space.

This does not make the example page look perfect on narrow screens, but at least it makes the example navigation and the examples themselves usable again.


ahocevar:
Before:
before

After:
after

tschaub:
Looks like an improvement.  Please merge.

PR #5118, opened 2016-03-24
Remove goog.dom.classlist usage

fredj:
Remove some `goog.dom.classlist` calls where simple `className` updates can be made.

PR #5124, opened 2016-03-25
Removed goog.dom.classlist

nicholas-l:
Removal of goog.dom.classlist means that IE9 requires a polyfill for classList.

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -11,7 +11,7 @@
     <link rel="stylesheet" href="./resources/layout.css" type="text/css">
     {{{ extraHead.local }}}
     {{{ css.tag }}}
-    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,requestAnimationFrame"></script>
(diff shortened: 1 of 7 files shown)


ahocevar:
Any reason why you closed this @nicholas-l? There is nothing wrong about requiring a polyfill for IE9.

nicholas-l:
Thought that @fredj was working on it as well, can reopen it if no overlap

nicholas-l:
See #5118 for the other part of classlist removal.

fredj:
#5118 is now merged

nicholas-l:
Rebased

ahocevar:
Thanks @nicholas-l. Can you please also change `config/examples/example.html` to add the polyfill for IE9? A bonus would be if you could create a follow-up pull request to do the same in https://github.com/openlayers/openlayers.github.io/blob/build/src/pages/index.hbs.

ahocevar:
Great work @nicholas-l, thanks a lot!

PR #5130, opened 2016-03-26
Remove goog.labs.userAgent

nicholas-l:
Not sure if there is a cleaner way of doing this so comments welcomed :)

### DIFF
--- a/src/ol/reproj/reproj.js
+++ b/src/ol/reproj/reproj.js
@@ -5,10 +5,6 @@ goog.require('ol.extent');
 goog.require('ol.math');
 goog.require('ol.proj');
 
-// Adapted from http://stackoverflow.com/questions/4565112/javascript-how-to-find-out-if-the-user-browser-is-chrome
-var winNav = window.navigator;
-var isOpera = winNav.userAgent.indexOf('OPR') > -1;
-var isIEedge = winNav.userAgent.indexOf('Edge') > -1;
 
 /**
  * We need to employ more sophisticated solution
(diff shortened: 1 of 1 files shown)


src/ol/reproj/reproj.js
```diff
@@ -18,8 +20,9 @@ goog.require('ol.proj');
  * @type {boolean}
  * @private
  */
-ol.reproj.browserAntialiasesClip_ = !goog.labs.userAgent.browser.isChrome() ||
-                                    goog.labs.userAgent.platform.isIos();
+ol.reproj.browserAntialiasesClip_ = !(!winNav.userAgent.match('CriOS') && window.chrome !== null &&
+    window.chrome !== undefined && winNav.vendor === 'Google Inc.' &&
+    isOpera == false && isIEedge == false);
```
tschaub: You could use an immediately invoked function expression here.  So, something like:

```js
ol.reproj. browserAntialiasesClip_ = (function() {
  var winNav = window.navigator;
  var isOpera = winNav.userAgent.indexOf('OPR') > -1;
  var isIEedge = winNav.userAgent.indexOf('Edge') > -1;
  return // ...
}
```

nicholas-l:
@tschaub Changed to a function

nicholas-l:
Anyone able to review?

src/ol/reproj/reproj.js
```diff
@@ -18,8 +16,18 @@ goog.require('ol.proj');
...
+  var isOpera = winNav.userAgent.indexOf('OPR') > -1;
+  var isIEedge = winNav.userAgent.indexOf('Edge') > -1;
+  return !(
+    !winNav.userAgent.match('CriOS') &&  // Not Chrome on iOS
+    winChrome !== null && winChrome !== undefined && // Has chrome in window
+    winNav.vendor === 'Google Inc.' && // Vendor is Google.
+    isOpera == false && // Not Opera
+    isIEedge == false // Not Edge
+  );
+})(window.navigator, window.chrome)
```
ahocevar: I'd use `goog.global` instead of `window` here, so this does not fail in non-browser environments.

ahocevar:
I added one more comment @nicholas-l. When you're able to address it, we can merge this. Thanks!

src/ol/reproj/reproj.js
nicholas-l: What about ```this```? goog.global = this in http://google.github.io/closure-library/api/source/closure/goog/base.js.src.html#l48

nicholas-l:
@ahocevar I have changed the global to ```this``` as I didn't want to add another goog :).

ahocevar:
Are you sure that `this` will also work in a browserify context? We use `goog.global` for now everywhere else.

nicholas-l:
I believe it would work as in this is what goog.global is doing, maybe we should have an ol.global? http://google.github.io/closure-library/api/source/closure/goog/base.js.src.html#l48

ahocevar:
`ol.global` would be a next step. This is why I'd prefer to use `goog.global` for now, because it will be easier and less error prone to change to `ol.global` later.

nicholas-l:
ok sounds good

ahocevar:
Thanks @nicholas-l.

nicholas-l:
Have created #5178 to tackle global

PR #5132, opened 2016-03-26
Remove goog.string

nicholas-l:
Added a utils.js file which has one function in it at the moment (padNumber) or would this be better suited in a string.js file?

### DIFF
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -2,8 +2,8 @@ goog.provide('ol.Coordinate');
 goog.provide('ol.CoordinateFormatType');
 goog.provide('ol.coordinate');
 
-goog.require('goog.string');
 goog.require('ol.math');
+goog.require('ol.string');
 
 
 /**
@@ -133,8 +133,8 @@ ol.coordinate.degreesToStringHDMS_ = function(degrees, hemispheres, opt_fraction
   var x = Math.abs(3600 * normalizedDegrees);
   var dflPrecision = opt_fractionDigits || 0;
   return Math.floor(x / 3600) + '\u00b0 ' +
-      goog.string.padNumber(Math.floor((x / 60) % 60), 2) + '\u2032 ' +
-      goog.string.padNumber((x % 60), 2, dflPrecision) + '\u2033 ' +
+      ol.string.padNumber(Math.floor((x / 60) % 60), 2) + '\u2032 ' +
+      ol.string.padNumber((x % 60), 2, dflPrecision) + '\u2033 ' +
       hemispheres.charAt(normalizedDegrees < 0 ? 1 : 0);
 };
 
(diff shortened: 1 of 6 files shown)


nicholas-l:
Also not sure how to tackle compareVersions, could we have a very simplify case and only deal with numeric types? i.e. not have to deal with "1.2b"

ahocevar:
Looks good. +1 on only handling numeric versions. Just one thing: could you rename utils to string? Also the file name.

nicholas-l:
Done :) not sure about the numeric versions. I put tests in for it and they seem to pass :) So ready for review

test/spec/ol/utils.test.js
```diff
@@ -0,0 +1,76 @@
+goog.provide('ol.test.utils');
```
tschaub: This should be `ol.test.string` and the file should be named `string.test.js`.

test/spec/ol/utils.test.js
```diff
@@ -0,0 +1,76 @@
...
+      expect(ol.string.padNumber(9.9, 2, 0)).to.be('10');
+      expect(ol.string.padNumber(5, 0, 0)).to.be('5');
+      expect(ol.string.padNumber(5, 1, 1)).to.be('5.0');
+      expect(ol.string.padNumber(5, 2, 1)).to.be('05.0');
+      expect(ol.string.padNumber(5, 2, 0)).to.be('05');
+    });
+
+  });
+
+  describe('ol.string.padNumber', function() {
```
tschaub: This should be `ol.string.compareVersions()`.

test/spec/ol/utils.test.js
```diff
@@ -0,0 +1,76 @@
...
+    var f = ol.string.compareVersions;
+    expect(f(1, 1)).to.be(0);
```
tschaub: These assertions need to be in an `it()` call for failures to be properly caught and reported.

test/spec/ol/utils.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+    expect(f('1', '1.1')).to.be.below(0);
+    expect(f('2.2', '2')).to.be.above(0);
+
+    // expect(f('1.1', '1.1b')).to.be.above(0);
+    // expect(f('1.1b', '1.1')).to.be.below(0);
+    // expect(f('1.1b', '1.1b')).to.be(0);
+
+    // expect(f('1.1b', '1.1a')).to.be.above(0);
+    // expect(f('1.1a', '1.1b')).to.be.below(0);
```
tschaub: Please remove the commented out code.

tschaub:
This looks good @nicholas-l.  Thanks for the work on it.  A few comments above about the tests.  Any failures in `ol.string.compareVersions()` assertions aren't currently being caught.

nicholas-l:
@tschaub changed according to comments

fredj:
@nicholas-l can you please squash all your commits into one? (there a good explanation on how to do that here: https://github.com/openlayers/ol3/pull/4868#issuecomment-187785482)

nicholas-l:
@fredj squashed

fredj:
Thanks @nicholas-l

PR #5134, opened 2016-03-26
Remove goog.dom

nicholas-l:
Not sure how far I will be able to go with removing goog.dom but this is a start.

- [X] createDom
- [X] removeNode
- [X] replaceNode
- [X] removeChildren
- [X] insertChildAt
- [X] insertSiblingAfter
- [X] getElement
- [x] fullscreen.*
- [x] nodeType

### DIFF
--- a/src/ol/control/attributioncontrol.js
+++ b/src/ol/control/attributioncontrol.js
@@ -3,8 +3,8 @@
 goog.provide('ol.control.Attribution');
 
 goog.require('goog.asserts');
-goog.require('goog.dom');
 goog.require('ol');
+goog.require('ol.dom');
 goog.require('ol.Attribution');
 goog.require('ol.control.Control');
 goog.require('ol.css');
@@ -68,30 +68,37 @@ ol.control.Attribution = function(opt_options) {
 
   var collapseLabel = options.collapseLabel !== undefined ? options.collapseLabel : '\u00BB';
 
-  /**
-   * @private
-   * @type {Node}
-   */
-  this.collapseLabel_ = typeof collapseLabel === 'string' ?
-      goog.dom.createDom('SPAN', {}, collapseLabel) :
-      collapseLabel;
+  if (typeof collapseLabel === 'string') {
+    /**
+     * @private
+     * @type {Node}
+     */
+    this.collapseLabel_ = document.createElement('span');
+    this.collapseLabel_.textContent = collapseLabel;
+  } else {
+    this.collapseLabel_ = collapseLabel;
+  }
 
   var label = options.label !== undefined ? options.label : 'i';
 
-  /**
-   * @private
-   * @type {Node}
-   */
-  this.label_ = typeof label === 'string' ?
-      goog.dom.createDom('SPAN', {}, label) :
-      label;
+  if (typeof label === 'string') {
+    /**
+     * @private
+     * @type {Node}
+     */
+    this.label_ = document.createElement('span');
+    this.label_.textContent = label;
+  } else {
+    this.label_ = label;
+  }
+
 
   var activeLabel = (this.collapsible_ && !this.collapsed_) ?
       this.collapseLabel_ : this.label_;
-  var button = goog.dom.createDom('BUTTON', {
-    'type': 'button',
-    'title': tipLabel
-  }, activeLabel);
+  var button = document.createElement('button');
+  button.setAttribute('type', 'button');
+  button.title = tipLabel;
+  button.appendChild(activeLabel);
 
   ol.events.listen(button, ol.events.EventType.CLICK, this.handleClick_, this);
 
@@ -99,8 +106,10 @@ ol.control.Attribution = function(opt_options) {
       ol.css.CLASS_CONTROL +
       (this.collapsed_ && this.collapsible_ ? ' ol-collapsed' : '') +
       (this.collapsible_ ? '' : ' ol-uncollapsible');
-  var element = goog.dom.createDom('DIV',
-      cssClasses, this.ulElement_, button);
+  var element = document.createElement('div');
+  element.className = cssClasses;
+  element.appendChild(this.ulElement_);
+  element.appendChild(button);
 
   var render = options.render ? options.render : ol.control.Attribution.render;
 
@@ -240,7 +249,7 @@ ol.control.Attribution.prototype.updateElement_ = function(frameState) {
       }
       delete hiddenAttributions[attributionKey];
     } else {
-      goog.dom.removeNode(this.attributionElements_[attributionKey]);
+      ol.dom.removeNode(this.attributionElements_[attributionKey]);
       delete this.attributionElements_[attributionKey];
       delete this.attributionElementRenderedVisible_[attributionKey];
     }
@@ -293,7 +302,7 @@ ol.control.Attribution.prototype.insertLogos_ = function(frameState) {
 
   for (logo in logoElements) {
     if (!(logo in logos)) {
-      goog.dom.removeNode(logoElements[logo]);
+      ol.dom.removeNode(logoElements[logo]);
       delete logoElements[logo];
     }
   }
@@ -311,9 +320,8 @@ ol.control.Attribution.prototype.insertLogos_ = function(frameState) {
       if (logoValue === '') {
         logoElement = image;
       } else {
-        logoElement = goog.dom.createDom('A', {
-          'href': logoValue
-        });
+        logoElement = document.createElement('a');
+        logoElement.href = logoValue;
         logoElement.appendChild(image);
       }
       this.logoLi_.appendChild(logoElement);
@@ -342,9 +350,9 @@ ol.control.Attribution.prototype.handleClick_ = function(event) {
 ol.control.Attribution.prototype.handleToggle_ = function() {
   this.element.classList.toggle('ol-collapsed');
   if (this.collapsed_) {
-    goog.dom.replaceNode(this.collapseLabel_, this.label_);
+    ol.dom.replaceNode(this.collapseLabel_, this.label_);
   } else {
-    goog.dom.replaceNode(this.label_, this.collapseLabel_);
+    ol.dom.replaceNode(this.label_, this.collapseLabel_);
   }
   this.collapsed_ = !this.collapsed_;
 };
--- a/src/ol/control/fullscreencontrol.js
+++ b/src/ol/control/fullscreencontrol.js
@@ -1,13 +1,11 @@
 goog.provide('ol.control.FullScreen');
 
 goog.require('goog.asserts');
-goog.require('goog.dom');
-goog.require('goog.dom.fullscreen');
-goog.require('goog.dom.fullscreen.EventType');
 goog.require('ol.events');
 goog.require('ol.events.EventType');
 goog.require('ol');
 goog.require('ol.control.Control');
+goog.require('ol.dom');
 goog.require('ol.css');
 
 
@@ -58,19 +56,21 @@ ol.control.FullScreen = function(opt_options) {
       document.createTextNode(labelActive) : labelActive;
 
   var tipLabel = options.tipLabel ? options.tipLabel : 'Toggle full-screen';
-  var button = goog.dom.createDom('BUTTON', {
-    'class': this.cssClassName_ + '-' + goog.dom.fullscreen.isFullScreen(),
-    'type': 'button',
-    'title': tipLabel
-  }, this.labelNode_);
+  var button = document.createElement('button');
+  button.className = this.cssClassName_ + '-' + ol.control.FullScreen.isFullScreen();
+  button.setAttribute('type', 'button');
+  button.title = tipLabel;
+  button.appendChild(this.labelNode_);
 
   ol.events.listen(button, ol.events.EventType.CLICK,
       this.handleClick_, this);
 
   var cssClasses = this.cssClassName_ + ' ' + ol.css.CLASS_UNSELECTABLE +
       ' ' + ol.css.CLASS_CONTROL + ' ' +
-      (!goog.dom.fullscreen.isSupported() ? ol.css.CLASS_UNSUPPORTED : '');
-  var element = goog.dom.createDom('DIV', cssClasses, button);
+      (!ol.control.FullScreen.isFullScreenSupported() ? ol.css.CLASS_UNSUPPORTED : '');
+  var element = document.createElement('div');
+  element.className = cssClasses;
+  element.appendChild(button);
 
   ol.control.Control.call(this, {
     element: element,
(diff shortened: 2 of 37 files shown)


nicholas-l:
Ready for review. Not sure about testing (coveralls)

src/ol/control/attributioncontrol.js
```diff
@@ -70,30 +70,37 @@ ol.control.Attribution = function(opt_options) {
...
-  this.collapseLabel_ = typeof collapseLabel === 'string' ?
-      goog.dom.createDom('SPAN', {}, collapseLabel) :
-      collapseLabel;
+  if (typeof collapseLabel === 'string') {
+    /**
+     * @private
+     * @type {Node}
+     */
+    this.collapseLabel_ = document.createElement('span');
+    this.collapseLabel_.appendChild(document.createTextNode(collapseLabel));
```
ahocevar: `this.collapseLabel_.textContent = collapseLabel;` would be shorter.

src/ol/control/attributioncontrol.js
```diff
@@ -70,30 +70,37 @@ ol.control.Attribution = function(opt_options) {
...
-  this.label_ = typeof label === 'string' ?
-      goog.dom.createDom('SPAN', {}, label) :
-      label;
+  if (typeof label === 'string') {
+    /**
+     * @private
+     * @type {Node}
+     */
+    this.label_ = document.createElement('span');
+    this.label_.appendChild(document.createTextNode(label));
```
ahocevar: Could also set `textContent` instead.

src/ol/control/attributioncontrol.js
```diff
@@ -70,30 +70,37 @@ ol.control.Attribution = function(opt_options) {
...
 
   var activeLabel = (this.collapsible_ && !this.collapsed_) ?
       this.collapseLabel_ : this.label_;
-  var button = goog.dom.createDom('BUTTON', {
-    'type': 'button',
-    'title': tipLabel
-  }, activeLabel);
+  var button = document.createElement('button');
+  button.setAttribute('type', 'button');
+  button['title'] = tipLabel;
```
ahocevar: Would be more consistent to also use `setAttribute` for 'title'.

src/ol/control/attributioncontrol.js
```diff
@@ -244,7 +253,9 @@ ol.control.Attribution.prototype.updateElement_ = function(frameState) {
       }
       delete hiddenAttributions[attributionKey];
     } else {
-      goog.dom.removeNode(this.attributionElements_[attributionKey]);
+      if (this.attributionElements_[attributionKey] && this.attributionElements_[attributionKey].parentNode) {
```
ahocevar: Maybe assign `this.attributionElements_[attributionKey]` to avoid repeated dereferencing.

src/ol/control/attributioncontrol.js
```diff
@@ -297,7 +308,9 @@ ol.control.Attribution.prototype.insertLogos_ = function(frameState) {
 
   for (logo in logoElements) {
     if (!(logo in logos)) {
-      goog.dom.removeNode(logoElements[logo]);
+      if (logoElements[logo] && logoElements[logo].parentNode) {
+        logoElements[logo].parentNode.removeChild(logoElements[logo]);
```
ahocevar: Same here re dereferencing

src/ol/control/attributioncontrol.js
```diff
@@ -311,9 +324,8 @@ ol.control.Attribution.prototype.insertLogos_ = function(frameState) {
       if (logoValue === '') {
         logoElement = image;
       } else {
-        logoElement = goog.dom.createDom('A', {
-          'href': logoValue
-        });
+        logoElement = document.createElement('a');
+        logoElement.href = logoValue;
```
ahocevar: Maybe use `setAttribute('href', logoValue)` instead?

marcjansen:
The decreased coverage is rational, because more code now is untested as we do the DOM manipulation ourselves. I think it'd be great if we had some basic tests for the following classes:

* `src/ol/control/fullscreencontrol.js`
* `src/ol/control/overviewmapcontrol.js`
* `src/ol/control/attributioncontrol.js`
* `src/ol/format/xmlfeatureformat.js`
* `src/ol/overlay.js`

Do you think you can these? If not someone else may be willing to jump in.

src/ol/control/fullscreencontrol.js
```diff
@@ -59,19 +57,21 @@ ol.control.FullScreen = function(opt_options) {
...
   var tipLabel = options.tipLabel ? options.tipLabel : 'Toggle full-screen';
-  var button = goog.dom.createDom('BUTTON', {
-    'class': this.cssClassName_ + '-' + goog.dom.fullscreen.isFullScreen(),
-    'type': 'button',
-    'title': tipLabel
-  }, this.labelNode_);
+  var button = document.createElement('button');
+  button.className = this.cssClassName_ + '-' + ol.control.FullScreen.isFullScreen();
+  button.setAttribute('type', 'button');
+  button.title = tipLabel;
```
ahocevar: Maybe `setAttribute`?

src/ol/control/fullscreencontrol.js
```diff
@@ -108,23 +108,33 @@ ol.control.FullScreen.prototype.handleClick_ = function(event) {
  * @private
  */
 ol.control.FullScreen.prototype.handleFullScreen_ = function() {
-  if (!goog.dom.fullscreen.isSupported()) {
+  if (!(document.body.webkitRequestFullscreen ||
```
ahocevar: Use your new `ol.control.FullScreen.isFullScreenSupported` instead.

src/ol/control/attributioncontrol.js
nicholas-l: I was going on how the closure library operated, tho I changed to .title later. My impression is it is better to use dot notation but default back to setAttribute when non standard. But I really dont mind, if you prefer I can change them to setAttribute?

src/ol/control/overviewmapcontrol.js
```diff
@@ -56,30 +56,37 @@ ol.control.OverviewMap = function(opt_options) {
...
-  this.collapseLabel_ = typeof collapseLabel === 'string' ?
-      goog.dom.createDom('SPAN', {}, collapseLabel) :
-      collapseLabel;
+  if (typeof collapseLabel === 'string') {
+    /**
+     * @private
+     * @type {Node}
+     */
+    this.collapseLabel_ = document.createElement('span');
+    this.collapseLabel_.appendChild(document.createTextNode(collapseLabel))
```
ahocevar: Using `textContent` would be shorter.

src/ol/control/overviewmapcontrol.js
```diff
@@ -56,30 +56,37 @@ ol.control.OverviewMap = function(opt_options) {
...
 
   var activeLabel = (this.collapsible_ && !this.collapsed_) ?
       this.collapseLabel_ : this.label_;
-  var button = goog.dom.createDom('BUTTON', {
-    'type': 'button',
-    'title': tipLabel
-  }, activeLabel);
+  var button = document.createElement('button');
+  button.setAttribute('type', 'button');
+  button.title = tipLabel;
```
ahocevar: Use `appendChild` for 'title'?

src/ol/control/rotatecontrol.js
```diff
@@ -46,18 +47,20 @@ ol.control.Rotate = function(opt_options) {
...
 
-  var button = goog.dom.createDom('BUTTON', {
-    'class': className + '-reset',
-    'type' : 'button',
-    'title': tipLabel
-  }, this.label_);
+  var button = document.createElement('button');
+  button.className = className + '-reset';
+  button.setAttribute('type', 'button');
+  button['title'] = tipLabel;
```
ahocevar: `setAttribute`?

src/ol/control/zoomcontrol.js
```diff
@@ -36,21 +35,25 @@ ol.control.Zoom = function(opt_options) {
...
 
-  var inElement = goog.dom.createDom('BUTTON', {
-    'class': className + '-in',
-    'type' : 'button',
-    'title': zoomInTipLabel
-  }, zoomInLabel);
+  var inElement = document.createElement('button');
+  inElement.className = className + '-in';
+  inElement.setAttribute('type', 'button');
+  inElement['title'] = zoomInTipLabel;
```
ahocevar: `setAttribute`?

src/ol/control/zoomcontrol.js
```diff
@@ -36,21 +35,25 @@ ol.control.Zoom = function(opt_options) {
...
 
-  var outElement = goog.dom.createDom('BUTTON', {
-    'class': className + '-out',
-    'type' : 'button',
-    'title': zoomOutTipLabel
-  }, zoomOutLabel);
+  var outElement = document.createElement('button');
+  outElement.className = className + '-out';
+  outElement.setAttribute('type', 'button');
+  outElement['title'] = zoomOutTipLabel;
```
ahocevar: `setAttribute`?

src/ol/control/zoomtoextentcontrol.js
```diff
@@ -33,17 +32,21 @@ ol.control.ZoomToExtent = function(opt_options) {
   var label = options.label !== undefined ? options.label : 'E';
   var tipLabel = options.tipLabel !== undefined ?
       options.tipLabel : 'Fit to extent';
-  var button = goog.dom.createDom('BUTTON', {
-    'type': 'button',
-    'title': tipLabel
-  }, label);
+  var button = document.createElement('button');
+  button.setAttribute('type', 'button');
+  button['title'] = tipLabel;
```
ahocevar: `setAttribute`?

ahocevar:
Great work @nicholas-l. I think it would make sense to add and use `ol.dom.removeNode` and `ol.dom.replaceNode`, given how many call sites we have. And we have `ol.dom` already. If you add tests for those, you might even get into the threshold for passing coveralls.

Not sure what the best way for setting the title of an element is. If the compiler does not rename `element.title = 'myTitle'`, then that would be the way to go. Otherwise `element.setAttribute('title', 'myTitle')`.

src/ol/control/attributioncontrol.js
ahocevar: I agree that dot notation is nicer. I hope the compiler does not rename it.

nicholas-l:
@ahocevar True, I had forgot about the compiler renaming attributes. It seems MDN recommend dot notation [MDN](https://developer.mozilla.org/en/docs/Web/API/Element/setAttribute#Notes).

Will do about replaceNode/removeNode

fredj:
The compiler will not rename the attributes

ahocevar:
@nicholas-l 

> It seems MDN recommend dot notation MDN

Cool, and given @fredj's comment above, please proceed with dot notation.

probins:
what's the status with this?

ahocevar:
Looks like all comments were addressed. @nicholas-l, would you be able to rebase this pull request and squash the commits?

nicholas-l:
Rebased but need to eslint, will squash after that

src/ol/renderer/dom/domtilelayerrenderer.js
```diff
@@ -65,7 +64,9 @@ ol.inherits(ol.renderer.dom.TileLayer, ol.renderer.dom.Layer);
  * @inheritDoc
  */
 ol.renderer.dom.TileLayer.prototype.clearFrame = function() {
-  goog.dom.removeChildren(this.target);
+  while (this.target.lastChild) {
+    this.target.removeChild(this.target.lastChild);
+  }
```
ahocevar: I'm wondering if it wouldn't make sense to create an `ol.dom.removeChildren` function, because this sequence here occurs multiple times.

ahocevar:
I added one more remark @nicholas-l. If you get a chance, it would be great if you could address it. And I think you need to rebase the pull request one more time so it merges without conflicts. Thanks for your effort on this!

nicholas-l:
Looks like I missed a goog.dom.contains in pointereventhandler.js, not sure how to tackle that as goog looks like they have some IE specific code

nicholas-l:
I got the error ```ERROR - unknown @define variable goog.dom.ASSUME_STANDARDS_MODE``` from the compiler so have remove the goog.dom.ASSUME_STANDARDS_MODE defines in the configs. Hope this is the correct thing to do :)

ahocevar:
Yes, this is correct @nicholas-l. Since we don't use `goog.dom.*` any more, there is also no `goog.dom.ASSUME_STANDARDS_MODE` any more.

src/ol/control/fullscreencontrol.js
```diff
@@ -156,8 +163,73 @@ ol.control.FullScreen.prototype.setMap = function(map) {
...
   if (map) {
     this.listenerKeys.push(
-        ol.events.listen(ol.global.document, goog.dom.fullscreen.EventType.CHANGE,
+        ol.events.listen(ol.global.document, ol.control.FullScreen.CHANGETYPE,
           this.handleFullScreenChange_, this)
     );
   }
 };
+
+ol.control.FullScreen.isFullScreenSupported = function() {
```
ahocevar: Would you mind annotating this function and the ones below? Thanks!

ahocevar:
Thanks for this huge effort @nicholas-l. I only added one more minor remark about annotating the new functions in `src/ol/control/fullscreencontrol.js`. Would be great if you could address that. This is definitely good to merge!

src/ol/control/fullscreencontrol.js
nicholas-l: Along the lines of the below?
```
/**
 * @return {boolean} Returns if fullscreen is supported by the current platform.
 */
```

ahocevar: Yes. Can even be simpler:
```js
/**
 * @return {boolean} Fullscreen is supported by the current platform.
 */
```

nicholas-l: Does the ol.control.FullScreen.CHANGETYPE need annotated?

ahocevar: Yes, can be as simple as
```js
/**
 * @type {string}
 */
```

src/ol/control/fullscreencontrol.js
```diff
@@ -156,8 +163,73 @@ ol.control.FullScreen.prototype.setMap = function(map) {
...
+  if (body.webkitRequestFullscreen) {
+    return 'webkitfullscreenchange';
+  } else if (body.mozRequestFullScreen) {
+    return 'mozfullscreenchange';
+  } else if (body.msRequestFullscreen) {
+    return 'MSFullscreenChange';
+  } else if (body.requestFullscreen) {
+    return 'fullscreenchange';
+  }
+  return null;
```
ahocevar: Since this is a string, it's better to return `undefined` instead of `null`.

ahocevar:
Thanks for the excellent work @nicholas-l!

probins:
reduces build size too 👍

bartvde:
We have a report on Twitter and I wonder if it might be related to this PR see: https://twitter.com/Programster_UK/status/749144247534845952

bartvde:
```
ol.js:116 Uncaught TypeError: Cannot read property 'webkitRequestFullscreen' of null
```

bartvde:
Since the script tag is in the head for our examples, I think we're assuming here that the DOM is ready when it is not?

probins:
related to #5548?

ahocevar:
Yes, related to and fixed with #5548. @bartvde, this may warrant a bugfix release.

PR #5138, opened 2016-03-29
Remove goog vec

nicholas-l:
Branch to try to change to glMatrix, currently very much a WIP

- [x] identity
- [x] translate
- [x] scale
- [x] createNumber
- [x] getElement
- [x] createNumberIdentity
- [x] invert
- [x] setFromArray
- [x] rotateZ
- [x] multMat
- [x] makeTranslate/makeRotateZ/makeScale
- [x] multVec4

### DIFF
--- a/externs/gl-matrix.js
+++ b/externs/gl-matrix.js
@@ -51,3 +51,13 @@ var quat;
  * @typedef {Array}
  */
 var quat4;
+
+/**
+ * @typedef {*}
+ */
+var global;
+
+/**
+ * @typedef {*}
+ */
+var require;
--- a/externs/gl-matrix.js
+++ b/externs/gl-matrix.js
@@ -1,0 +1,53 @@
+var SIMD = {};
+
+
+/**
+ * @typedef {Float32Array|Array}
+ */
+var Type;
+
+
+/**
+ * @typedef {Array} mat2
+ */
+var mat2;
+
+/**
+ * @typedef {Array} mat2
+ */
+var mat2d;
+
+/**
+ * @typedef {Array}
+ */
+var mat3;
+
+/**
+ * @typedef {Array}
+ */
+var mat4;
+
+/**
+ * @typedef {Array}
+ */
+var vec2;
+
+/**
+ * @typedef {Array}
+ */
+var vec3;
+
+/**
+ * @typedef {Array}
+ */
+var vec4;
+
+/**
+ * @typedef {Array}
+ */
+var quat;
+
+/**
+ * @typedef {Array}
+ */
+var quat4;
--- a/config/examples-all.json
+++ b/config/examples-all.json
@@ -22,7 +22,8 @@
       "externs/jquery-1.9.js",
       "externs/proj4js.js",
       "externs/tilejson.js",
-      "externs/topojson.js"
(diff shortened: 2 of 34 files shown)


nicholas-l:
All tests are going to fail because there is two ```@fileoverview```s in the file meaning the ```@suppress``` gets ignored.

nicholas-l:
Ready for review. Larger build size but maybe with ES6ify of glMatrix that @bjornharrtell is d[oing](https://github.com/toji/gl-matrix/pull/209), this could reduce.

ahocevar:
Thanks for this huge effort @nicholas-l. Out of interest: I see that npmjs.com has many sub-packages of gl-matrix. Wouldn't it make sense to use only those needed, instead of including the whole thing? Also, is there a reason for not using the module from npmjs.com, but your branch instead?

nicholas-l:
The reason for using my branch is that it fails when there is two fileoverviews (one from the package and one from the headers the externs script adds). I changed to my own branch in the start but can see if it works with the npm package. Also I think the separate npm packages are out of date (could be wrong :)). Eventually I see using the ES6 glmatrix when it comes out and it won't matter as we can import the modules we need.

bjornharrtell:
I was motivated to do https://github.com/toji/gl-matrix/pull/209 for just that reason, to avoid having to split up a library into several packages. I'm a bit allergic to the package-as-module that the NPM registry has become. But it needs (alot?) more work both on glMatrix side and OpenLayers to be useful.

externs/gl-matrix.js
```diff
@@ -0,0 +1,63 @@
+var SIMD = {};
```
fredj: not used?

nicholas-l:
I get an error with the current package on NPM but I have already submitted a fix (and a subsequent one) to toji, and it has already been merged. After these fixes, I still get errors (473) like below:
```
ERR! compile /home/nicholas/Development/ol3/build/ol.ext/glmatrix.js:6490: ERROR - inconsistent return type
ERR! compile found   : boolean
ERR! compile required: (Boolean|null)
ERR! compile     return (Math.abs(a0 - b0) <= glMatrix.EPSILON*Math.max(1.0, Math.abs(a0), Math.abs(b0)) &&
```
which I think it is due to two ```@fileoverview```s and the suppresses not being applied.

externs/gl-matrix.js
nicholas-l: Not used but the compiler complains about undefined global variable.

tschaub:
I haven't dug into it yet, but I was hoping to first do a bit of an audit into our vec/mat use.  I'm curious, for example, to see if we are doing a lot of mat4 operations where we could be doing mat3 operations.  Are there places where we need mat4 currently?

nicholas-l:
Not sure but I assumed that the Mat4 operations were due to eventual view/3D/tilting.

probins:
@nicholas-l now that #5335 has been merged, could you rename the typedef `ol.vec.Mat4.Number` to say `ol.Mat4Number` and move it to `typedefs.js`. This will bring it into line with the other typedefs. Thanks.

nicholas-l:
@probins sorry about the radio silence! Yes, will rebase and change typedef

PR #5141, opened 2016-03-29
Reuse dragListenerKeys_ variable in ol.control.ZoomSlider

fredj:
Memory optimization: avoids to create a new array.


ahocevar:
Looks good, please merge.

PR #5149, opened 2016-03-30
Batch polygon fill and stroke instructions

### LINKED ISSUE #4232: forEachFeatureAtPixel freezes/super slow performance on ff and ie
I have 2000+ features on the map and I want to display popup on hover to the features. I have used the forEachFeatureAtPixel  and on chrome it works as expected but with firefox and internet explorer it's completely useless. Whole map freezes and stops working.

I have tried with interactions (I think it uses forEachFeatureAtCoordinate and forEachFeatureAtPixel) and with map.on('pointermove'...). Works ok with Chrome but freezes on FF/IE.

Is there any workarounds, hacks or anything that would/could/should work?

ahocevar:
I did some profiling while I investigated #4232, and found that the canvas vector renderer spends most of its time in `context.fill()` instructions. In a first iteration, I changed the renderer code to create a single batch of `context.fill()` and `context.stroke()` instructions with the same `fillStyle` and `strokeStyle`. It turned out that this makes rendering slow in Safari. In a second iteration, I limited the batch size to 200 polygons, which results in fast rendering in all tested browsers.

Fixes #4232.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -1196,7 +1196,9 @@ ol.render.canvas.PolygonReplay = function(tolerance, maxExtent, resolution) {
    *         lineDash: Array.<number>,
    *         lineJoin: (string|undefined),
    *         lineWidth: (number|undefined),
-   *         miterLimit: (number|undefined)}|null}
+   *         miterLimit: (number|undefined),
+   *         pendingFill: number,
+   *         pendingStroke: number}|null}
    */
   this.state_ = {
     currentFillStyle: undefined,
@@ -1212,7 +1214,9 @@ ol.render.canvas.PolygonReplay = function(tolerance, maxExtent, resolution) {
     lineDash: null,
     lineJoin: undefined,
     lineWidth: undefined,
-    miterLimit: undefined
+    miterLimit: undefined,
+    pendingFill: 0,
+    pendingStroke: 0
   };
 
 };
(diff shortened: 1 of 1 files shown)


src/ol/render/canvas/canvasreplay.js
```diff
@@ -1246,19 +1255,15 @@ ol.render.canvas.PolygonReplay.prototype.drawFlatCoordinatess_ = function(flatCo
...
+    state.pendingFill++;
   }
   if (state.strokeStyle !== undefined) {
     goog.asserts.assert(state.lineWidth !== undefined,
         'state.lineWidth should be defined');
-    var strokeInstruction = [ol.render.canvas.Instruction.STROKE];
-    this.instructions.push(strokeInstruction);
-    this.hitDetectionInstructions.push(strokeInstruction);
+    state.pendingStroke++;
+    this.hitDetectionInstructions.push([ol.render.canvas.Instruction.STROKE]);
```
tsauerwein: Would it maybe make sense to also batch hit-detections instructions?

ahocevar: Time in `context.fill()` is almost only spent for rendering, not for hit detection. So I think that filling off-screen geometries is already optimised away by the Canvas 2D engine.

tsauerwein: Ok, that sounds good. Btw. thanks for your research on this! :+1:

fredj:
LGTM, thanks for your work on this

PR #5150, opened 2016-03-31
Use fs-extra instead of wrench

fredj:


tasks/test-coverage.js
```diff
@@ -87,9 +73,9 @@ var setupBackupAndInstrumentationDir = function() {
   }
 
   log('• copy src files to backup folder');
-  wrench.copyDirSyncRecursive(dir, backupDir, copyOpts);
+  fse.copySync(dir, backupDir);
```
nicholas-l: This wont overwrite existing files does that matter? Also preserving timestamps? ```{clobber: true, preserveTimestamps: true}```

fredj: The backupDir should be empty at this time, but I will add the options to have the same as we had with wrench.

fredj: done

ahocevar:
Looks good, please merge.

tasks/test-coverage.js
```diff
@@ -8,8 +8,8 @@
  */
 
 var fs = require('fs');
+var fse = require('fs-extra');
```
marcjansen: You can just do

```javascript
var fs = require('fs-extra');
```

and remove the `fse` variable

tasks/test-coverage.js
```diff
@@ -87,9 +80,9 @@ var setupBackupAndInstrumentationDir = function() {
   }
 
   log('• copy src files to backup folder');
-  wrench.copyDirSyncRecursive(dir, backupDir, copyOpts);
+  fse.copySync(dir, backupDir, copyOpts);
```
marcjansen: that'd be than `fs.copySync` also at the other places

marcjansen:
The function `deleteFolderRecursive` can also be replaced with `fs.remove`.

If you could take my suggestions, @fredj, that'd be great. Otherwise looks good to me as well.

fredj:
> The function `deleteFolderRecursive` can also be replaced with `fs.remove`.

done, thanks

tasks/test-coverage.js
fredj: you're right but we have the `fs` and `fse` vars in all the others scripts in the tasks folder.
we can change that in an other PR

marcjansen: Agreed. :+1:

marcjansen:
Thanks for the additional commit @fredj. Please merge when green.

fredj:
thanks for the review

PR #5152, opened 2016-04-01
Cartodb cleanup

fredj:


fredj:
Thanks for any review

gberaudo:
LGTM

tschaub:
Not sure if this is related, but the http://openlayers.org/en/v3.15.1/examples/cartodb.html example doesn't work.

Also, was there some discussion about the justification for adding an externs file?

tschaub:
> Not sure if this is related, but the http://openlayers.org/en/v3.15.1/examples/cartodb.html example doesn't work.

Unrelated.  See #5217

> Also, was there some discussion about the justification for adding an externs file?

Here is some related discussion on adding an externs file: https://github.com/openlayers/ol3/pull/4926#discussion-diff-54385613

My opinion hasn't changed since then (I don't think there is value in adding an externs file and I think it is an additional burden on users - existing full build configs won't work for example).

PR #5157, opened 2016-04-01
Improve docs for measure example

probins:
The current description in the measure example is not very accurate, so this is an attempt to improve it.

While I'm at it, I've also corrected a minor error in the js file.

Docs only, no change in logic.


ahocevar:
Thanks @probins.

PR #5162, opened 2016-04-02
Improve docs/examples for loadingstrategy

probins:
I was looking for docs on the bbox loadingstrategy. The typedef currently has no description, so I've added a minimal one. I searched the examples for 'bbox', and it finds 3, none of which actually uses a bbox strategy :confused: So I've removed the bbox description from the ESRI examples, and changed the WFS example so that does use `ol.loadingstrategy.bbox` - AFAICS this continues to work with no problem, so I'm not sure why it currently uses a Tile strategy, which seems like an unnecessary complication. No doubt someone will let me know if I'm missing something ;-)


probins:
review anyone?

bartvde:
thanks @probins this makes sense to me.

PR #5163, opened 2016-04-04
Release v3.15.0

fredj:


nicholas-l:
Might have forgot one change, math.modulo and lerp in full list :(

fredj:
@nicholas-l added, thanks

fredj:
I've also added a note about IE 9 and the `goog.dom.classlist` removal (#5124)

PR #5165, opened 2016-04-04
Remove goog.addSingletonGetter function

fredj:
See #4128

### DIFF
--- a/src/ol/style/icon.js
+++ b/src/ol/style/icon.js
@@ -1,7 +1,7 @@
 goog.provide('ol.style.Icon');
 goog.provide('ol.style.IconAnchorUnits');
-goog.provide('ol.style.IconImageCache');
 goog.provide('ol.style.IconOrigin');
+goog.provide('ol.style.iconImageCache');
 
 goog.require('ol.events');
 goog.require('ol.events.EventTarget');
@@ -458,7 +458,7 @@ ol.inherits(ol.style.IconImage_, ol.events.EventTarget);
  */
 ol.style.IconImage_.get = function(image, src, size, crossOrigin, imageState,
                                    color) {
-  var iconImageCache = ol.style.IconImageCache.getInstance();
+  var iconImageCache = ol.style.iconImageCache;
   var iconImage = iconImageCache.get(src, crossOrigin, color);
   if (!iconImage) {
     iconImage = new ol.style.IconImage_(
@@ -638,6 +638,7 @@ ol.style.IconImage_.prototype.unlistenImage_ = function() {
 
 /**
  * @constructor
+ * @private
  */
 ol.style.IconImageCache = function() {
 
@@ -660,7 +661,6 @@ ol.style.IconImageCache = function() {
    */
   this.maxCacheSize_ = 32;
 };
-goog.addSingletonGetter(ol.style.IconImageCache);
 
 
 /**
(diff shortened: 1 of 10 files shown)


src/ol/objectutil.js
```diff
@@ -68,3 +68,21 @@ ol.object.isEmpty = function(object) {
   }
   return !property;
 };
+
+
+/**
+ * Adds a {@code getInstance} static method that always returns the same
```
tschaub: Elsewhere we use markdown (backtick) for code formatting.

fredj: good catch, fixed

tschaub:
In all cases where `ol.foo.Boo.getInstance()` is called, I think it would be clearer to instead access `ol.foo.bar`.  And in these cases, we never explicitly call `new ol.foo.Bar()`.

In other words, rather than copying the Closure Library style here, I think it would be nicer to get rid of this convention altogether.  So instead of

```js
// where it is defined
goog.addSingletonGetter(ol.style.IconImageCache);

// where it is used
ol.style.IconImageCache.getInstance();
```

We would have this:

```js
// where it is defined
ol.style.iconImageCache = new ol.style.IconImageCache();

// where it is used
ol.style.iconImageCache
```

src/ol/objectutil.js
```diff
@@ -68,3 +68,21 @@ ol.object.isEmpty = function(object) {
...
+
+/**
+ * Adds a `getInstance` static method that always returns the same
+ * instance object.
+ * @param {!Function} object The constructor for the class to add the static
+ *     method to.
+ */
+ol.object.addSingletonGetter = function(object) {
+
+  object.getInstance = function() {
```
tschaub: I'm curious if the Compiler knows anything about the type returned by this function.

probins:
status?

ahocevar:
I don't like the `*.instance` naming convention. I like better what @tschaub had suggested, so e.g. 
```js
ol.style.iconImageCache = new ol.style.IconImageCache()
```
instead of
```js
ol.style.IconImageCache.instance = new ol.style.IconImageCache()
```
When addressed, I'd say this is good to merge.

@fredj, are you still active on this, or should we create a new pull request?

fredj:
After changing ```ol.style.IconImageCache.getInstance()``` to ```ol.style.IconImageCache``` the compiler complains that ```goog.require('ol.style.IconImageCache')``` is unnecessary.

What's the right way to fix that?

ahocevar:
I think you'll have to `goog.provide('ol.style.iconImageCache')` instead of `goog.provide('ol.style.IconImageCache' )`. Same goes for `goog.require`.

fredj:
done. I had to make `ol.style.IconImageCache` `@private` to be able to remove the `goog.provide('ol.style.IconImageCache');`

src/ol/style/icon.js
```diff
@@ -638,6 +638,7 @@ ol.style.IconImage_.prototype.unlistenImage_ = function() {
 
 /**
  * @constructor
+ * @private
  */
 ol.style.IconImageCache = function() {
```
ahocevar: Now you could rename this to `ol.style.IconImageCache_`

ahocevar:
Looks good now, thanks! Please merge when lights are green.

fredj:
Thanks for the review

PR #5167, opened 2016-04-04
Remove FastClick from mobile-full-screen example

fredj:
See https://bugs.webkit.org/show_bug.cgi?id=149968

FastClick is not needed anymore to prevent the tap delay for Safari; if the viewport is unscalable (user-scalable=no) the delay is removed (just like firefox and chrome).


bartvde:
+1

fredj:
Thanks for the review

PR #5171, opened 2016-04-05
Move olx.source.AttributionOption out of externs/olx.js

### LINKED ISSUE #5168: Build error when using ol-externs.js
In an application I build using OpenLayers' ol-externs.js file, after upgrading to v3.15.0 I get the following error.

```
ERR! compile node_modules/openlayers/build/ol-externs.js:18672:
    ERROR - Bad type annotation. type not recognized due to syntax error
ERR! compile  * @typedef {{}}
ERR! compile               ^
ERR! compile 
ERR! compile 
ERR! compile 1 error(s), 0 warning(s)
ERR! compile 
```

I would be willing to work on a fix for this, but I'm not sure how this should be addressed.  If someone could give me a hint, I'd work on a PR for this.

Also, this is not the first time I get errors when using the ol-externs.js file after upgrading.  It would make sense if the Continuous Integration could test that file also for tickets causing those errors be responsible of making sure that the proper fix is made immediately.

Thoughts about this would be welcome.

Thanks,

Alexandre

fredj:
should fix #5168 

can be tested with `node tasks/generate-externs.js ol-externs.js`

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3947,13 +3947,6 @@ olx.source;
 
 
 /**
- * @typedef {string|Array.<string>|ol.Attribution|Array.<ol.Attribution>}
- * @api
- */
-olx.source.AttributionOption;
-
-
-/**
  * @typedef {{cacheSize: (number|undefined),
  *     culture: (string|undefined),
  *     key: string,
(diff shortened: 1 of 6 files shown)


marcjansen:
Nice! Thanks for the quick fix.

Is there any chance we can detect this automatically? Because `make check` and `make ci` didn't catch this.

Anyhow. Please merge once green, @fredj

fredj:
Thanks for the review @marcjansen

adube:
Thanks to the both of you for this quick fix and review!

PR #5177, opened 2016-04-05
Add render options for vector tile layers

### LINKED ISSUE #4897: Slow Zoom In
Unfortunately, I don't have much time to look into this now, but I thought I'd bring this to your guys attention and check with you on it.

I set up a Nodejs server to serve mbtile files using mbtiles and tile-live modules. I set up your latest OL3 version to handle them using Vector tiles. For the most part I think everything was working fine. 

The biggest issue seemed to be that there was a large lag every time I zoomed in. I think Firefox may have been the worst, but If I remember correctly all browsers had this problem. What I thought was odd is that panning and zooming out was very fast. The issue was only when zooming in.

 I was only using 10m state data from Natural Earth converted to mbtile using Mapbox Studio Classic. Any thoughts or ideas?

ahocevar:
The goal of this pull request is to simplify code and improve performance and rendering quality:

* Change `ol.renderer.canvas.VectorTileLayer` so it inherits from `ol.renderer.canvas.TileLayer`. With this change, the new improved tile rendering with clipped lower resolution regions is used, and no gaps between vector tiles are visible any more - not even on rotated views.
* Instead of deciding whether to render vector tile replays to a tile image or directly to the output canvas based on the tile pixel size, we can now do both tile image and output canvas rendering at the same time, even with different subsets of the replay group. By default, polygon and line elements are rendered to a tile image, and images and texts to the output canvas.
* The new `renderMode` option on `ol.layer.VectorTile` allows users to control the rendering behaviour. The default is `'hybrid'`, which means polygons and lines are no longer replayed with every render frame, but rendered to a tile image. Images and texts are replayed, so they stay upright during rotation and maintain their scale during zoom animations.

Fixes #4897.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3798,6 +3798,7 @@ olx.layer.VectorOptions.prototype.visible;
  *     maxResolution: (number|undefined),
  *     opacity: (number|undefined),
  *     renderBuffer: (number|undefined),
+ *     renderMode: (ol.layer.VectorTileRenderType|string|undefined),
  *     renderOrder: (function(ol.Feature, ol.Feature):number|undefined),
  *     source: (ol.source.VectorTile|undefined),
  *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
@@ -3821,6 +3822,22 @@ olx.layer.VectorTileOptions;
  */
 olx.layer.VectorTileOptions.prototype.renderBuffer;
 
+
+/**
+ * Render mode for vector tiles:
+ *  * `'image'`: Vector tiles are rendered as images. Great performance, but
+ *    point symbols and texts are always rotated with the view and pixels are
+ *    scaled during zoom animations.
+ *  * `'hybrid'`: Polygon and line elements are rendered as images, so pixels
+ *    are scaled during zoom animations. Point symbols and texts are accurately
+ *    rendered as vectors and can stay upright on rotated views.
+ *  * `'vector'`: Vector tiles are rendered as vectors. Most accurate rendering
+ *    even during animations, but slower performance than the other options.
+ *  The default is `'hybrid'`.
+ * @type {ol.layer.VectorTileRenderType|string|undefined}
+ * @api
+ */
+olx.layer.VectorTileOptions.prototype.renderMode;
 
 /**
  * Render order. Function to be used when sorting features before rendering. By
@@ -4361,7 +4378,6 @@ olx.source.TileImageOptions.prototype.wrapX;
  *            cacheSize: (number|undefined),
  *            format: (ol.format.Feature|undefined),
  *            logo: (string|olx.LogoOptions|undefined),
- *            opaque: (boolean|undefined),
  *            projection: ol.proj.ProjectionLike,
  *            state: (ol.source.State|undefined),
  *            tileClass: (function(new: ol.VectorTile, ol.TileCoord,
@@ -4410,14 +4426,6 @@ olx.source.VectorTileOptions.prototype.format;
  * @api
  */
 olx.source.VectorTileOptions.prototype.logo;
-
-
-/**
- * Whether the layer is opaque.
- * @type {boolean|undefined}
- * @api
- */
-olx.source.VectorTileOptions.prototype.opaque;
 
 
 /**
--- a/src/ol/renderer/canvas/canvastilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvastilelayerrenderer.js
@@ -317,11 +317,8 @@ ol.renderer.canvas.TileLayer.prototype.renderTileImages = function(context, fram
       pixelExtents.push(pixelExtent);
     }
     var tilePixelSize = source.getTilePixelSize(currentZ, pixelRatio, projection);
-    var image = tile.getImage();
-    if (image) {
-      renderContext.drawImage(image, tileGutter, tileGutter,
-          tilePixelSize[0], tilePixelSize[1], left, top, w, h);
-    }
+    renderContext.drawImage(tile.getImage(), tileGutter, tileGutter,
+        tilePixelSize[0], tilePixelSize[1], left, top, w, h);
     if (!opaque) {
       renderContext.restore();
     }
--- a/src/ol/renderer/canvas/canvastilelayerrenderer.js
+++ b/src/ol/renderer/canvas/canvastilelayerrenderer.js
@@ -9,9 +9,9 @@ goog.require('ol.TileState');
 goog.require('ol.array');
 goog.require('ol.dom');
 goog.require('ol.extent');
-goog.require('ol.layer.Tile');
 goog.require('ol.render.EventType');
 goog.require('ol.renderer.canvas.Layer');
+goog.require('ol.size');
 goog.require('ol.source.Tile');
 goog.require('ol.vec.Mat4');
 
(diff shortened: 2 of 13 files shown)


ahocevar:
I managed to make the compiler happy, so this is now ready for review.

probins:
> Change ol.renderer.canvas.VectorTileLayer so it inherits from ol.renderer.canvas.VectorTileLayer

... canvas.TileLayer

ahocevar:
Thanks @probins, description updated.

ahocevar:
Tests added.

bjornharrtell:
Checked this out and played around with the three different settings for `renderMode` and works as expected hugely improving the use of vector tiles in the default `hybrid` mode which also looks great. Awesome work @ahocevar!

ahocevar:
Any additional review would be much appreciated. Thanks.

ahocevar:
@tsauerwein, would you be able to review this? You have reviewed other vector tile related pull requests before, so it's probably easier for you than for others. Thanks!

tsauerwein:
@ahocevar I will have a look later today.

externs/olx.js
```diff
@@ -4361,7 +4378,6 @@ olx.source.TileImageOptions.prototype.wrapX;
  *            cacheSize: (number|undefined),
  *            format: (ol.format.Feature|undefined),
  *            logo: (string|olx.LogoOptions|undefined),
- *            opaque: (boolean|undefined),
```
tsauerwein: Do you think it would make sense to mention the removal of `opaque` in the upgrade notes?

src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -323,3 +211,140 @@ ol.renderer.canvas.TileLayer.prototype.forEachLayerAtPixel = function(
...
+          renderContext.lineTo(clipExtent[0], clipExtent[3]);
+          renderContext.closePath();
+          renderContext.clip();
+        }
+      }
+      pixelExtents.push(pixelExtent);
+    }
+    var tilePixelSize = source.getTilePixelSize(currentZ, pixelRatio, projection);
+    var image = tile.getImage();
+    if (image) {
```
tsauerwein: I saw you added this check. At this point, can there be tiles that do not have an image?

ahocevar: Thanks for spotting this. The check can indeed be removed.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -508,3 +380,65 @@ ol.renderer.canvas.VectorTileLayer.prototype.renderFeature = function(feature, s
...
+ * @param {olx.FrameState} frameState Frame state.
+ * @param {ol.layer.LayerState} layerState Layer state.
+ * @param {Array.<string>} skippedFeatures Skipped features.
+ */
+ol.renderer.canvas.VectorTileLayer.prototype.renderTileImage = function(
+    tile, frameState, layerState, skippedFeatures) {
+  var layer = this.getLayer();
+  var replays = ol.renderer.canvas.IMAGE_REPLAYS[layer.getRenderMode()];
+  if (!replays) {
+    return;
```
tsauerwein: maybe add a comment like "only continue when renderMode is not 'vector'"?

externs/olx.js
ahocevar: Good question. The option will not hurt if someone has it in their code. And it did not work properly previously. Happy to add it if you think. Do you?

src/ol/map.js
```diff
@@ -1490,7 +1490,7 @@ ol.Map.createOptionsInternal = function(options) {
     var logo = options.logo;
     if (typeof logo === 'string') {
       logos[logo] = '';
-    } else if (goog.isObject(logo)) {
+    } else if (goog.isObject(logo) && !(logo instanceof HTMLElement)) {
```
tsauerwein: I think you committed this one accidentally.

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -508,3 +380,66 @@ ol.renderer.canvas.VectorTileLayer.prototype.renderFeature = function(feature, s
...
+    var currentZ = tile.getTileCoord()[0];
+    var resolution = tileGrid.getResolution(currentZ);
+    var tileSize = ol.size.toSize(tileGrid.getTileSize(currentZ));
+    var tileResolution = tileGrid.getResolution(currentZ);
+    var resolutionRatio = tileResolution / resolution;
+    var width = tileSize[0] * pixelRatio * resolutionRatio;
+    var height = tileSize[1] * pixelRatio * resolutionRatio;
+    tileContext.canvas.width = width / resolutionRatio + 0.5;
+    tileContext.canvas.height = height / resolutionRatio + 0.5;
+    tileContext.scale(1 / resolutionRatio, 1 / resolutionRatio)
```
tsauerwein: missing semicolon

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -508,3 +380,66 @@ ol.renderer.canvas.VectorTileLayer.prototype.renderFeature = function(feature, s
...
 };
+
+
+/**
+ * @param {ol.VectorTile} tile Tile.
+ * @param {olx.FrameState} frameState Frame state.
+ * @param {ol.layer.LayerState} layerState Layer state.
+ * @param {Array.<string>} skippedFeatures Skipped features.
+ */
+ol.renderer.canvas.VectorTileLayer.prototype.renderTileImage = function(
```
tsauerwein: This method could be private.

externs/olx.js
tsauerwein: True, it will not break applications. But I think it would be good if we document when options are removed from the API.

tsauerwein:
I am not sure if I understand everything, but the idea of having these three render-modes sounds very good to me, and your changes and refactorings look fine!

And I really don't see any gaps any more! :+1:

ahocevar:
Thanks @tsauerwein for the thorough review. I know it wasn't an easy one. Do you feel comfortable saying that this is good to merge, or would you like to have someone else take another look? Or do you want any questions about the whole concept answered, so it's easier for you to understand?

fredj:
The changes LGTM too, +1 for merging

tsauerwein:
@ahocevar I didn't try to understand every line of the rendering logic, because it was a bit hard to see which parts of the code are new, and which parts were only moved to a different method. But the concept is clear and I agree with it. So, +1 for merging.

ahocevar:
Thanks again for the hard review work!

tsauerwein:
Thanks to you for doing the actual work! :)

I-Maps:
I upgraded my maps to OpenLayers 17.1 and noticed that my Mapzen vector tile roads where broken on my cell phone map versions.  I tracked this problem down to this change that was made in the vector tile rendering in version 16.   On the cell phone maps, the roads are broken or replicated as scattered broken road segments.   I tried adding the renderMode options, but none helped.   The 'vector' option even broke the raster tile rendering on the cell phone map.   

The roads are not broken in the desktop browser versions, but I noticed that the maps rendered a lot slower with this new vector rendering.  Also, after using my function to zoom to point locations , when I zoom back out, the vector lines for my red interstates will make huge black rendered lines before the new red interstate lines are made for the new zoom level.  

Everything was working fine before, so I was wondering if the default rendering could be set back to the way it was before version 16.   Otherwise, what should I do to fix all of these new problems?

ahocevar:
This bug report is very vague. Can you maybe point to a JSFiddle that shows the issue with 3.17.1 and another that shows the expected behaviour with 3.16?

I-Maps:
I'll try to stick to one issue at a time.   After more investigation, I found the issue with the broken roads is isolated to my Windows cell phone.   I did not see it on other Android phones.   I saw that Microsoft finally made an upgrade to Windows 10 mobile for my phone.   I installed it this morning and the problem went away.   On the jpeg link below, you can see what I was seeing on the Windows 8.1 system.  The image shows a side by side comparison of my map with OL 15 and OL 16.  Windows 8 is going away, so I wouldn't worry about it.   http://geoserverweb.usipp.net/OL_v15-v16_Roads.jpg

PR #5178, opened 2016-04-06
Remove goog.global

nicholas-l:
Not sure on the best way to do this, thoughts?

### DIFF
--- a/externs/closure-compiler.js
+++ b/externs/closure-compiler.js
@@ -15,9 +15,3 @@
  * @see http://www.w3.org/TR/pointerevents/#the-touch-action-css-property
  */
 CSSProperties.prototype.touchAction;
-
-
-/**
- * @type {Object}
- */
-var global;
--- a/externs/closure-compiler.js
+++ b/externs/closure-compiler.js
@@ -15,3 +15,9 @@
  * @see http://www.w3.org/TR/pointerevents/#the-touch-action-css-property
  */
 CSSProperties.prototype.touchAction;
+
+
+/**
+ * @type {Object}
+ */
+var global;
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -7441,3 +7441,9 @@ olx.style.AtlasManagerOptions.prototype.maxSize;
  * @api
  */
 olx.style.AtlasManagerOptions.prototype.space;
+
+
+/**
+ * @type {Object}
+ */
+var global;
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -7441,9 +7441,3 @@ olx.style.AtlasManagerOptions.prototype.maxSize;
  * @api
  */
 olx.style.AtlasManagerOptions.prototype.space;
-
-
-/**
- * @type {Object}
- */
-var global;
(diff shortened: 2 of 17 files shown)


nicholas-l:
I imagine there will be a bit of discussion about this as a lot of code relies on the ability to create document elements etc.

externs/olx.js
```diff
@@ -7441,3 +7441,9 @@ olx.style.AtlasManagerOptions.prototype.maxSize;
  * @api
  */
 olx.style.AtlasManagerOptions.prototype.space;
+
+
+/**
+ * @type {Object}
+ */
+var global;
```
nicholas-l: Not sure how this should be done as I got an error about ```global``` being undefined.

ahocevar:
I think this is the right approach. I'm only surprised about the change to `externs/olx.js`. Was that just to make the compiler happy?

nicholas-l:
Yea otherwise it errors out:
```
info ol Compiling 315 sources
ERR! compile /home/nicholas/Development/ol3/src/ol/ol.js:278: ERROR - variable global is undeclared
ERR! compile   if (typeof global !== 'undefined') {
ERR! compile              ^
ERR! compile 
ERR! compile 
ERR! compile 1 error(s), 
ERR! compile 0 warning(s)
ERR! compile 
ERR! Process exited with non-zero status, see log for more detail: 1 ```

**Comment by ahocevar** [comment]
I think this is a bug in the closure compiler. It should know about `global`. If I am correct, then `global` should be defined in `externs/closure-compiler.js`, not `externs/olx.js`. cc @fredj - what does our Closure expert say?

**Comment by fredj** [comment]
This can be simplified to `ol.global = this;`
If we are in a nodejs environment, `this` will be `global` and `window` if we are in a browser.

That's how it's done in closure-library

**Comment by nicholas-l** [comment]
Initially I thought that ```ol.global = this``` this but in a module environment this is meant to be ```undefined``` according to http://exploringjs.com/es6/ch_modules.html#sec_modules-vs-scripts

**Comment by ahocevar** [comment]
@fredj If you search for 'global' in the compiled code, you'll see that the compiler does something along the lines of what is done in this pull request.

**Comment by marcjansen** [comment]
would

```js
ol.global = (function() {
  return this;
}());
```

work everywhere?

fredj:
In es5 strict mode, `this` is also `undefined`; I'm wondering how the compiler handles this

marcjansen:
Obscure, but

```
ol.global = Function('return this')()
```

looks OK for me everyhwere. (Idea from http://stackoverflow.com/a/3277192)

externs/closure-compiler.js
```diff
@@ -15,3 +15,9 @@
  * @see http://www.w3.org/TR/pointerevents/#the-touch-action-css-property
  */
 CSSProperties.prototype.touchAction;
+
+
+/**
+ * @type {Object}
+ */
+var global;
```
ahocevar: This should no longer be needed now.

ahocevar:
I think this looks great now. Thanks @nicholas-l, and thanks @marcjansen for the research.

PR #5179, opened 2016-04-06
Use correct beginPath - fill - stroke order when skipping features

### LINKED ISSUE #5175: Strange behavior of feature setStyle on Selection interaction
As of version 3.15.0, I've noticed that the setting feature style on the 'select' event for the Selection interaction has behaved very strangely. The style I'm attempting to set on the selected feature will either be set on seemingly random nearby features as well, or the entire background will have the style applied... I'm not sure what's going on here.

I've made an attempt at creating a minimal jsfiddle of the problem [here.](http://jsfiddle.net/enfumpz5/2/)

At first I thought it was a problem with using setStyle on features in general, but I couldn't reproduce the problem until using the selection interaction.

### LINKED ISSUE #5186: Circle rendering regression
In 3.15, when adding two or more circle geometries to a vector layer, a line is now being drawn between them. This was not the case in 3.14.2 

* 3.14.2: http://jsfiddle.net/e0jh9bob/

* 3.15.0: http://jsfiddle.net/4gfr1d6L/

ahocevar:
Fixes #5175.
Fixes #5186.

I don't have time to add tests right now, but I think we should have tests for that. Anyone volunteering to provide a cherry pick?


ahocevar:
Also, there might be a nicer way to do this.

ahocevar:
There are remaining issues which can be seen when multi-selecting in the select-features example. That results in multiple fills (with both master and this fix).

ahocevar:
Better solution implemented, tests added. This is now ready for review.

ahocevar:
Now also fixes #5186.

ahocevar:
I'm going to add more tests.

marcjansen:
Can we use the example images from #5186 as visual tests?

marcjansen:
Also thanks a lot for tackling this @ahocevar.

ahocevar:
> Can we use the example images from #5186 as visual tests?

On it already.

ahocevar:
There is more work to do. The rendering test is still failing.

ahocevar:
Ok, the problem with grouping batches is that overlapping fills won't create a higher visual opacity:

Without batching:
not-batched

With batching:
batched

Should we revert the whole batching optimisation? I'd love to hear opinions.

fredj:
Do you think it's faisable to have higher visual opacity with batching?

What about reverting, create a patch release and work on this later?

fredj:
And it would be nice to commit the rendering tests anyway

ahocevar:
I don't know if it is possible to batch fills and still get increased opacity for overlapping geometries.

@fredj I'll create a pull request with just the rendering tests, and one that reverts the batch optimisation so you can create a patch release.

ahocevar:
Will create a new pull request for batching fills and strokes.

PR #5187, opened 2016-04-07
Remove goog.functions.and from ol.interaction.Modify

fredj:


src/ol/interaction/modifyinteraction.js
```diff
@@ -113,14 +112,21 @@ ol.interaction.Modify = function(options) {
   });
 
   /**
+   * @private
+   * @param {ol.MapBrowserEvent} mapBrowserEvent Browser event.
+   * @return {boolean} Combined condition result.
```
marcjansen: Return type `ol.events.ConditionType`?

marcjansen:
One minor question, otherwise this looks good to me. Please merge at will with or without the question adressed (assuming checks turn green)

src/ol/interaction/modifyinteraction.js
fredj: It's `{boolean}`; the function returns the results from the functions invocation.

marcjansen: Yeah, I was missreading the (now thankfully gone `@type`). Thanks

fredj:
Thanks for the review

PR #5190, opened 2016-04-07
Revert "Batch polygon fill and stroke instructions"

### LINKED ISSUE #5175: Strange behavior of feature setStyle on Selection interaction

### LINKED ISSUE #5186: Circle rendering regression

ahocevar:
Reverts openlayers/ol3#5149

Fixes #5175.
Fixes #5186.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -153,10 +153,11 @@ goog.inherits(ol.render.canvas.Replay, ol.render.VectorContext);
  * @param {number} offset Offset.
  * @param {number} end End.
  * @param {number} stride Stride.
+ * @param {boolean} close Close.
  * @protected
  * @return {number} My end.
  */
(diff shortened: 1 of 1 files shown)


fredj:
Thanks, please merge

PR #5193, opened 2016-04-07
Add target option to attribution logo

samuellapointe:
The original problem I had can be found [here](https://groups.google.com/forum/#!topic/ol3-dev/waTmFWpSK7o). In short, I wanted to create an attribution logo that opened in a new tab when clicked.

This commit adds a target option to the LogoOptions. This makes it possible to left-click on an attribution logo and make it open in a new tab. For example, when creating a map:

```
var map = new ol.Map({
        target: 'map',
        layers: [
          new ol.layer.Tile({
          })
        ],
        view: new ol.View({
          center: ol.proj.fromLonLat([37.41, 8.82]),
          zoom: 4
        }),
        logo: {
          'href': "http://www.google.ca",
          'src': "logo.jpg",
          'target': "_blank"
        }
      });
```

The logo will appear as logo.jpg, and when clicked, it will open a new tab to google.ca.
When this option is not set, the target for the logo's link is not set, and it acts like before.

To implement this, I had to change the logos dict format from `{src: href}` to `{src: [href, target]}`.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -98,6 +98,15 @@ olx.LogoOptions.prototype.src;
 
 
 /**
+ * Link target for the logo. If undefined, none will be set and clicking on
+ * the logo will open the link in the current tab.
+ * @type {string|undefined}
+ * @api
+ */
+olx.LogoOptions.prototype.target;
+
+
+/**
  * @typedef {{map: (ol.Map|undefined),
  *     maxLines: (number|undefined),
  *     strokeStyle: (ol.style.Stroke|undefined),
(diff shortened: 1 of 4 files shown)


adube:
This seems to be ready for review.  Is there something more required here for this to be reviewed ?

marcjansen:
My personal view here: I am not a friend of `target` let alone `target="_blank"` and feel it's an antipattern to decide for users, what happens if they clock a link. There may be exceptions, of course.

Could we come up with something more generic? To set the `title` or other attributes?

I wint block this and appreciate the effort, I am just not convinced this is a real added value.

adube:
Thanks for your feedback, @marcjansen.  Just an FYI, this is something that has been asked to us twice so far.  I agree with you that it's an antipattern to decide for users the behaviour of `target`, but people still want to be able to control this nevertheless.

I agree that it would make more sense to have some sort of `key:value` object as options instead of an array.  This would allow any attribute to be set. This would also require a change in the migration guide, though.

What do you think ?

ahocevar:
What about accepting {boolean|string|olx.LogoOptions|HTMLElement} as logo and not modify `olx.LogoOptions`? Then you could provide your custom markup.

adube:
That's an excellent idea.  +1

samuellapointe:
Thanks for the feedback, I will create a new pull request using @ahocevar's method today.

PR #5194, opened 2016-04-07
Get rid of goog.inherits and goog.base

fredj:
See https://developer.mozilla.org/en/docs/Web/JavaScript/Inheritance_and_the_prototype_chain

### DIFF
--- a/src/ol/collection.js
+++ b/src/ol/collection.js
@@ -44,7 +44,7 @@ ol.CollectionEventType = {
  */
 ol.CollectionEvent = function(type, opt_element, opt_target) {
 
-  goog.base(this, type, opt_target);
+  ol.events.Event.call(this, type, opt_target);
 
   /**
    * The element that is added to or removed from the collection.
@@ -54,7 +54,7 @@ ol.CollectionEvent = function(type, opt_element, opt_target) {
   this.element = opt_element;
 
 };
-goog.inherits(ol.CollectionEvent, ol.events.Event);
+ol.inherits(ol.CollectionEvent, ol.events.Event);
 
 
 /**
@@ -82,7 +82,7 @@ ol.CollectionProperty = {
  */
 ol.Collection = function(opt_array) {
 
-  goog.base(this);
+  ol.Object.call(this);
 
   /**
    * @private
@@ -93,7 +93,7 @@ ol.Collection = function(opt_array) {
   this.updateLength_();
 
 };
-goog.inherits(ol.Collection, ol.Object);
+ol.inherits(ol.Collection, ol.Object);
 
 
 /**
--- a/src/ol/format/ogc/filter.js
+++ b/src/ol/format/ogc/filter.js
@@ -247,9 +247,9 @@ ol.format.ogc.filter.Filter.prototype.getTagName = function() {
  * @extends {ol.format.ogc.filter.Filter}
  */
 ol.format.ogc.filter.Logical = function(tagName) {
-  goog.base(this, tagName);
+  ol.format.ogc.filter.Filter.call(this, tagName);
 };
-goog.inherits(ol.format.ogc.filter.Logical, ol.format.ogc.filter.Filter);
+ol.inherits(ol.format.ogc.filter.Logical, ol.format.ogc.filter.Filter);
 
 
 /**
@@ -265,7 +265,7 @@ goog.inherits(ol.format.ogc.filter.Logical, ol.format.ogc.filter.Filter);
  */
 ol.format.ogc.filter.LogicalBinary = function(tagName, conditionA, conditionB) {
 
-  goog.base(this, tagName);
+  ol.format.ogc.filter.Logical.call(this, tagName);
 
   /**
    * @public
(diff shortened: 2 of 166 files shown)


nicholas-l:
Nice!

probins:
@fredj I've been thinking about this in the context of changing to modules. I'm a bit surprised that the compiler doesn't complain that there's no `goog.require('ol')` for `ol.inherits`, and I think this will be necessary with modules. If you have to write `var ol = goog.require('ol');` (or `import { inherits } from 'ol'`) in every file, then there's not only the extra overhead of importing another file for the 2-line `ol.inherits` function, but it's also not much shorter than writing the function out in every file, for example, for `ol.Overlay`:
```
ol.Overlay.prototype = Object.create(ol.Object.prototype);
ol.Overlay.prototype.constructor = ol.Overlay;
```
So I'm wondering if it wouldn't be better not to use `ol.inherits`; the function is currently exported, so can be kept in `ol.js` for bc.

Btw, you link to MDN, but there's an better example of this on the [Object.create() page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create), lines 15-21 of the classical inheritance section.

probins:
> I'm wondering if it wouldn't be better not to use ol.inherits

I'll work on a PR for this for further discussion

gberaudo:
Hi @probins. I think @fredj was planning to work on this early next week.

My opinion is that it is preferable to abstract the inheritance mechanism in only one place instead of duplicating the concept everywhere. It avoid bugs and ease evolutions.

I found this (2013) article about automatic inlining by the v8 engine. It may be that modern javascript engines automatically optimize this function call.

Would be happy to hear what others think about this case?

tschaub:
With ES2015, this would be written as:

```js
class Foo extends Bar {
  // ...
}
```

As an intermediate step, replacing `goog.inherits()` with `ol.inherits()` seems like the simplest change to me.

fredj:
this is now ready for review

fredj:
todo: add a note in the release note

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
```diff
@@ -47,7 +47,7 @@ ol.renderer.canvas.VECTOR_REPLAYS = {
  */
 ol.renderer.canvas.VectorTileLayer = function(layer) {
 
-  goog.base(this, layer);
+  ol.renderer.canvas.Layer.call(this, layer);
```
gberaudo: Should be `ol.renderer.canvas.TileLayer`.

gberaudo:
Thank you @fredj for this huge and meticulous work.
I checked every line and could only found one single issue.

+1 for merging when this is addressed and a note is added about how to migrate.

src/ol/collection.js
```diff
@@ -82,7 +82,7 @@ ol.CollectionProperty = {
  */
 ol.Collection = function(opt_array) {
 
-  goog.base(this);
+  ol.Object.call(this);
```
ahocevar: Wouldn't `goog.base(this)` (single argument) equivalent to `ol.Object.apply(this, arguments)` rather than `ol.Object.call(this)`?

src/ol/renderer/canvas/canvasvectortilelayerrenderer.js
fredj: thanks, fixed

src/ol/collection.js
probins: better surely to use call() throughout

ahocevar: So `ol.Object.call(this, opt_array);`.

gberaudo: `goog.base` is used both in constructors and to call super class methods, which is quite confusing.

@ahocevar, are you refering to the `arguments` received by the `goog.base` function as used in https://github.com/google/closure-library/blob/master/closure/goog/base.js#L2460-L2465?

That one should be different from the `arguments` received by the `ol.Collection` constructor.

ahocevar: Ok, please disregard my `goog.base` comments above. I was under the impression that `goog.base(this)` (single argument) has a special meaning, which it apparently does not have.

src/ol/format/ogc/filter.js
```diff
@@ -219,15 +219,15 @@ ol.format.ogc.filter.like = function(propertyName, pattern,
...
+  ol.Object.call(this);
 
   /**
    * @private
    * @type {!string}
    */
   this.tagName_ = tagName;
 };
-goog.inherits(ol.format.ogc.filter.Filter, ol.Object);
+ol.inherits(ol.format.ogc.filter.Filter, ol.Object);
```
tschaub: Not introduced here, but this stands out as odd to me (see https://github.com/openlayers/ol3/pull/5252/files#r65902356).

fredj: ping @tsauerwein

tschaub:
Thanks for the work getting this done @fredj.  I think this looks good to merge.

fredj:
Thanks for the review

PR #5195, opened 2016-04-07
Remove unneeded ol.VectorTile#disposeInternal function

fredj:
This function only calls it's parent class; it can be removed


ahocevar:
This is also done in #5177.

ahocevar:
Feel free to merge this regardless.

If you feel like spending some time on a review, I'd be grateful if you could take a look at #5177.

PR #5196, opened 2016-04-07
Batch polygon and circle fills and strokes

### LINKED ISSUE #4232: forEachFeatureAtPixel freezes/super slow performance on ff and ie

ahocevar:
Fixes #4232.

Better solution than the reverted one from #5149.


ahocevar:
Opened for discussion. See description above.

ahocevar:
How about this: when creating the replay, we check if there is any opacity. If so, we do not batch. Otherwise we do. I can update the pull request easily for a proof of concept.

ahocevar:
How about this: when creating the replay, we check if there is any opacity. If so, we do not batch. Otherwise we do. I can update the pull request easily for a proof of concept.

ahocevar:
I think this is a nice approach. Ready for review.

bjornharrtell:
While I agree this approach is a nice default that works, I would find an option to force the batching always on useful in cases where I don't care about the opacity-correctness.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -1440,13 +1484,20 @@ ol.render.canvas.PolygonReplay.prototype.setFillStrokeStyle = function(fillStyle
     var fillStyleColor = fillStyle.getColor();
     state.fillStyle = ol.colorlike.asColorLike(fillStyleColor ?
         fillStyleColor : ol.render.canvas.defaultFillStyle);
+    if (!this.transparency && ol.color.isRgba(state.fillStyle)) {
```
tschaub: Since `ol.color.isRgba()` is only used to test if alpha is not `1`, it would be simpler just to have `ol.color.isOpaque()` or something (rather than first testing if it is RGBA and then creating an array and then testing the alpha value).

src/ol/render/canvas/canvasreplay.js
```diff
@@ -1440,13 +1484,20 @@ ol.render.canvas.PolygonReplay.prototype.setFillStrokeStyle = function(fillStyle
...
+      this.transparency = ol.color.asArray(
+          /** @type {ol.Color|string} */ (state.fillStyle))[0] != 1;
```
tschaub: I think you mean to test the alpha value here (not the red value).

tschaub: Is `state.fillStyle` ever `{CanvasPattern|CanvasGradient}` here?

(It would be nicer if we knew we always had a string at this point.)

tschaub: Ok, I see that we won't get here with a pattern or gradient.  But I think things would be more efficient to have a `ol.color.isOpaque()` function.  And I think it is really only needed here to check strings.

tschaub:
I don't want to rain on the parade, but I'm dubious about the effort to batch fills and strokes in this way.  Here is a picture with batched fills and strokes:

image

I don't think this is what people want.

I'm also curious about how this affects what is reported in #4232, since I think that with the changes here, things are not being batched (because `this.transparency` is only `true` if the red value is `1`, see https://github.com/openlayers/ol3/pull/5196#discussion_r59698640).

tschaub:
I can imagine we could gain performance when there is no fill or no stroke.  In those cases, we could defer stroking or filling (respectively) until the end.  We could also delay stroking or filling until a fill or stroke is encountered (no need to stroke until you are going to fill and no need to fill until you are going to stroke).  But these performance gains would be limited to pretty narrow cases.

ahocevar:
Thanks @tschaub and @bjornharrtell for the valuable feedback. Based on that, I introduced a new `overlaps` option for `ol.source.Vector` and `ol.source.VectorTile`, which allows users to decide whether the fill and stroke optimisations should be applied. Does this look better to you now?

bjornharrtell:
Looks good to me. I like the note on which kind of data this works on, another argument for my mission to convince people to work with non-overlapping topological data.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -258,6 +265,10 @@ ol.render.canvas.Replay.prototype.replay_ = function(
   var localTransform = this.tmpLocalTransform_;
   var localTransformInv = this.tmpLocalTransformInv_;
   var prevX, prevY, roundX, roundY;
+  var pendingFill = 0;
+  var pendingStroke = 0;
+  var batchSize =
+      this.instructions != instructions || this.overlaps ? 0 : 200;
```
tschaub: Curious about the `200` here.  Did you find that performance is degraded if the batch size gets too big?  It might be nice to document the justification for this default on its own line.

tschaub:
This sounds like a nice solution @ahocevar.  It looks like your tests may cover it, but it would be even clearer to see rendering tests for `overlaps: false` with and without changing stroke and fill styles (to make sure we always properly stroke/fill when the styles change).

ahocevar:
Thanks for the review @tschaub. I have addressed your comments. Can you take another look please?

ahocevar:
Is this good to merge now @tschaub?

ahocevar:
Any chance to get a final review on this? Thanks in advance.

ahocevar:
Updated pull request so it can be merged into master.

Is this good to merge, or does anyone want to give this another review?

ahocevar:
To compare performance with and without `overlaps: false`, hover over the USA in the following fiddles:

* with `overlaps: false`: http://jsfiddle.net/4cxeh29o/
* without `overlaps: false`: http://jsfiddle.net/4cxeh29o/1/

That said, I'd still appreciate a final review on this. @tschaub maybe?

jbrueggmann:
We've been running a build of this in our dev- and demo environments for about a month now with very positive result.

ahocevar:
Rebased again, still hoping for a final review.

ahocevar:
Rebased one more time.

externs/olx.js
```diff
@@ -4416,6 +4417,17 @@ olx.source.VectorTileOptions.prototype.logo;
 
 
 /**
+ * This source may have overlapping geometries. Default is `true`. Setting this
+ * to `false` (e.g. for sources with polygons that represent adminstrative
```
bartvde: administrative

externs/olx.js
```diff
@@ -5850,6 +5863,17 @@ olx.source.VectorOptions.prototype.logo;
...
```
bartvde: administrative

src/ol/source/vectortile.js
```diff
@@ -53,6 +53,12 @@ ol.source.VectorTile = function(options) {
   this.format_ = options.format ? options.format : null;
 
   /**
+   * @private
+   * @type {boolean}
+   */
+  this.overlaps_ = options.overlaps || true;
```
bartvde: minor but in vector source this is determined slightly different (```options.overlaps == undefined ? true : options.overlaps```)

bartvde:
LGTM @ahocevar

ahocevar:
Thanks for the reviews @tschaub and @bartvde.

probins:
sorry for being late, but shouldn't this addition to the API be documented in the release notes?

ahocevar:
> sorry for being late, but shouldn't this addition to the API be documented in the release notes?

Changes like this can be added to the release notes by the release manager when the release is created.

jeroenvheel:
the performance boost with this change is amazing !!

mstrop:
This says the issue #4232 is closed by this issue, I but still have terrible performance issue using IE (11 or Edge) - FF is a bit better, but not much. I use OL 3.19 version. Is there any progress on this?

ahocevar:
@mstrop Have you configured your source with `overlaps: false`?

mstrop:
Yep, both - with and without. The difference is minimal. In addition to this, having the source defined as overlaps: false is causing display bugs for certain zooms.

image

ahocevar:
@mstrop Looks like your source does not have topological data, so you cannot use the `overlaps: false` option anyway. I think you'll have to apply different optimisations here, because your use case is not one that this fix was made for.

mstrop:
@ahocevar ups, it's not a good message. Is there a chance this issue will be solved in future?

ahocevar:
@mstrop The only two places where I see additional potential for library side performance optimization is text and regular shape (including circle) rendering.

In your case, however, I suspect that you could optimize the style functions for your vector layers on the application level.

mstrop:
@ahocevar thanks a lot.

mstrop:
@ahocevar sry, one more question - I was thinking about your comment "I suspect that you could optimize the style functions for your vector layers on the application level". What is relation between a styling function and forEachFeatureAtPixel method? I use forEachFeatureAtPixel inside pointer move event as I need to change cursor in case it hovers above a feature. So the styling function is not called as the map is not being redrawn. Or am I missing anything?

ahocevar:
Even if you are only using `forEachFeatureAtPixel` for changing the cursor, without highlighting any features on the map, the style function will be called. Because for hit detection, features near the cursor will be rendered to a hidden small canvas to detect which features are rendered at the cursor location.

ahocevar:
My mistake @mstrop. The style function is called when creating the replay, not when replaying. And during mousemoves, it is only replayed.

However, without seeing your code, it is hard to guess what could be the culprit. Does your vector source use a spatial index? Have you tried with a lower `hitTolerance` as option on your `forEachFeatureAtPixel` calls?

PR #5201, opened 2016-04-08
Small clarification for vector source loader docs

probins:
It's not currently correct for the case when both loader and url are provided.


ahocevar:
Thanks @probins.

PR #5208, opened 2016-04-09
Stop GPX writing rtept properties

### LINKED ISSUE #5034: GPX write for linestring adds feature properties to every point
If I have a LineString with for example name and desc properties, and write it in GPX format, I get for example:
```
<gpx xmlns="http://www.topografix.com/GPX/1/1">
  <rte><name>nn</name><desc>desc</desc>
    <rtept lat="12.554564" lon="-18.28125"><name>nn</name><desc>desc</desc></rtept>
    <rtept lat="48.922499" lon="29.53125"><name>nn</name><desc>desc</desc></rtept>
    <rtept lat="-1.4061088" lon="29.53125"><name>nn</name><desc>desc</desc></rtept>
  </rte>
</gpx>
```
so it outputs the feature properties in every single point. This makes output files much larger than they should be.

ISTM the feature properties should only be output on `<wpt>` nodes, not `<rtept>`.

I took a quick look at the code, and it looks like this is because writeWptType_() is used by points, routes and tracks, and I'm thinking `context['properties']` should only be used in the case of points.

probins:
Fixes #5034 as proposed [there](https://github.com/openlayers/ol3/issues/5034#issuecomment-207815064). I think this fits in with the way the existing code works.

I've also added some docs for read rte and trk.

### DIFF
--- a/src/ol/format/gpxformat.js
+++ b/src/ol/format/gpxformat.js
@@ -432,6 +432,8 @@ ol.format.GPX.prototype.handleReadExtensions_ = function(features) {
 
 /**
  * Read the first feature from a GPX source.
+ * Routes (`<rte>`) are converted into LineString geometries, and tracks (`<trk>`)
+ * into MultiLineString. Any properties on route and track waypoints are ignored.
  *
  * @function
  * @param {Document|Node|Object|string} source Source.
@@ -466,6 +468,8 @@ ol.format.GPX.prototype.readFeatureFromNode = function(node, opt_options) {
 
 /**
  * Read all features from a GPX source.
+ * Routes (`<rte>`) are converted into LineString geometries, and tracks (`<trk>`)
+ * into MultiLineString. Any properties on route and track waypoints are ignored.
  *
  * @function
  * @param {Document|Node|Object|string} source Source.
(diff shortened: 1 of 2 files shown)


probins:
review anyone?

src/ol/format/gpxformat.js
```diff
@@ -569,7 +573,17 @@ ol.format.GPX.writeWptType_ = function(node, coordinate, objectStack) {
...
       // pass
   }
-  var orderedKeys = ol.format.GPX.WPT_TYPE_SEQUENCE_[namespaceURI];
+  var orderedKeys;
+  switch (node.nodeName) {
+    case 'rtept':
+      orderedKeys = ol.format.GPX.RTEPT_TYPE_SEQUENCE_[namespaceURI];
+      break;
+    case 'trkpt':
+      orderedKeys = ol.format.GPX.TRKPT_TYPE_SEQUENCE_[namespaceURI];
```
bartvde: since the keys are the same for RTEPT and TRKPT, is there really a need for these to be separate? Or can they be combined into 1 sequence instead?

bartvde:
Also any chance of adding a simple unit test @probins ?

src/ol/format/gpxformat.js
probins: yeah, we can have `RTEPT_TRKPT_TYPE_SEQUENCE` or something instead

probins:
ok, I'll look into adding a test for this

probins:
adding a couple of tests proved to be a useful exercise, as I found there is not currently a problem with trkpt as I assumed (note to self: never assume), but only with rtept.

So, code simplified, and I added a test for trk anyway, and also for rte.

bartvde:
thanks @probins

Issue #4945, opened 2016-03-01
Make builds work with manage_closure_dependencies


tschaub:
Creating builds with `manage_closure_dependencies: true` can result in smaller build sizes.  While this doesn't seem like it should be the case, there is unused code that is unconditionally included in the builds (see #3447 for example).  Using `manage_closure_dependencies` improves the situation by limiting the files that are included for compilation.  However, because our `olx.js` file relies on types defined throughout the library, the `olx.js` file cannot safely be passed to the compiler when `manage_closure_dependencies` is set to `true`.

This can be demonstrated by creating a build config that only exports a single source (e.g. `ol.source.Raster`).  This generates compiler errors like the following:

```
externs/olx.js:2409: ERROR - Bad type annotation. Unknown type ol.interaction.DragBoxEndConditionType
 *     boxEndCondition: (ol.interaction.DragBoxEndConditionType|undefined)}}
                         ^

externs/olx.js:2446: ERROR - Bad type annotation. Unknown type ol.interaction.DragBoxEndConditionType
 * @type {ol.interaction.DragBoxEndConditionType|undefined}
          ^

2 error(s), 0 warning(s)
```

One alternative could be to distribute the externs files around the `src` directory (e.g. `src/ol/interaction/dragboxinteraction.externs.js`) and have a build tool concatenate what is needed for a custom build - but this is piling more custom logic and complexity into the build tool.

It would be nice if there were an annotation we could used to describe our externs, rather than needing to include them in separate files.  A custom annotation and build step to generate externs is another option (similar to the one above).

Anybody else have any ideas?

ahocevar:
I don't have any alternative that would not include custom build steps. Maybe @fredj knows something?

probins:
I thought that moving the typedefs into 1 file would fix this problem, but it doesn't. It fixes the issues for typedefs, such as `ol.interaction.DragBoxEndConditionType` mentioned, but the compile still fails with `Unknown type ol.layer.VectorTileRenderType`, which is an enum not a typedef. These could be moved into one file as well, but that still wouldn't fix the problem, as there are still constructor types referred to in olx, and these have to remain in their own module/file.

I encountered the opposite problem when I first started creating custom builds. I thought it would be really clever to only include those olx externs that I actually needed, but this too fails, as the code refers to other olx types that I don't use. I suppose the basic issue is that the compiler has to know about _all_ the type definitions, whether they're needed for a particular build or not.

PR #5210, opened 2016-04-10
Remove unrequired requires

probins:
Whilst thinking some more about conversion to ES6 modules, I noticed that there's some inconsistency in the `goog.require`s. In map, for example, `ol.Coordinate`, `ol.Pixel`, `ol.PreRenderFunction` and `ol.PostRenderFunction` are not used in the code, but only for defining a type in the docs comments, so in module terms, there is no import. `ol.Coordinate` does not have a `goog.require`, whereas the other 3 do. As the compiler doesn't, er, require a require for these, ISTM it would simplify the conversion if these unneeded requires were removed. There are probably others around, and I'll submit other PRs if I spot them. The first baby step towards ES6 modules :smile:


tschaub:
We have an unsolved problem for people who might want to use `manage_closure_dependencies`.  If the types are not in the dependency tree (i.e. not explicitly `goog.require()`'d), the Compiler will fail (see #4945).  So this is potentially exacerbating that problem.

Even in the ES6 modules world, we'll have to somehow depend on the appropriate `@typedef`s to pass code to the Compiler.  We could have a single `types.js` file to handle this (this would be prepended to a bundle before passing to the compiler).  But it may be nicer from a development standpoint to have types defined throughout the library and imported/required only where they are used.

Anybody else have thoughts on the subject?

Either way, I think this could be merged (though it increases the lurking `manage_closure_dependencies` issue).

probins:
I don't know if it affects anything, but `manage_closure_dependencies` has now been deprecated, as of February release.

fredj:
`manage_closure_dependencies` was renamed to `dependency_mode=LOOSE`.
see https://github.com/google/closure-compiler/blob/3ad2498/src/com/google/javascript/jscomp/CommandLineRunner.java#L449-L454

gberaudo:
There are tricky cases with typedefs.

In https://github.com/gberaudo/ol3/commit/2ff8a1b7d8dd0246edf9113a22da8585a0c97779 the require was missing from `ol/source/source.js` but the type error occured in `ol/source/bingmapssource.js`.

With one file per type it is easy for a user to find a type, be it constructor, enum or typedef since the location of the file directly depend on the type name. With a single file the user needs to know about the typedefs file and then browse a huge file.

Would a single typedef file work with custom builds (restricted list of exports)? Do we need goog.requires?

I would name an eventual typedef file `typedefs` since constructors and enums are also types.

probins:
> We could have a single types.js file to handle this (this would be prepended to a bundle before passing to the compiler)

if we had that, and it's documented that those creating custom builds should always include this file, then #4945 goes away, right?

probins:
these unneeded requires are hard to spot, but it occurs to me that if we change to `goog.module` and the `var x = goog.require('x')` pattern, then these will be flagged as unused vars.

gberaudo:
I experimented with using `goog.module` for typedefs but it was producing errors.
It seems more reasonable to keep `goog.provide` for typedefs since they bear no code, but are only instructions for the closure compiler.

probins:
I'll close this, and include it in the PR for #5299

PR #5215, opened 2016-04-12
Add an optional feature filter to getClosestFeatureToCoordinate

### LINKED ISSUE #5205: Allow filtering features in getClosestFeatureToCoordinate
I want to able to skip some features when using `getClosestFeatureToCoordinate`. Something like this:

``` js
function isFoo(feature) {
    return feature.get("class") === "foo";
}
var feature = source.getClosestFeatureToCoordinate(pixel, isFoo);
```

I'm thinking this has some overlap with #3830

fredj:
fixes #5205


bartvde:
LGTM
a test case would be nice though, but I won't block you on it

fredj:
simple test added

fredj:
Thanks for the review

PR #5216, opened 2016-04-12
Don't draw a sketch line string when drawing a polygon

### LINKED ISSUE #3235: weird flickr effect when drawing w/ dashed style
When using a dashed line style for the draw interaction, the dashes may move along the line which is visually bad and can be disrupting.
It's easy to reproduce by drawing a polygon using a counter clock wise motion.
http://bl.ocks.org/pgiraud/raw/dc1ff40c0821b3f73120/

ol3_draw_dash_
ol3_draw_dash

fredj:
Add a stroke to the polygon sketch style and remove the sketch line when drawing a polygon.

Fixes #3235 

The default sketch style is changed from:
draw_before

to:
draw_after

work in progress

### DIFF
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -243,14 +243,14 @@ ol.interaction.Draw = function(options) {
   this.sketchCoords_ = null;
 
   /**
-   * Sketch line. Used when drawing polygon.
+   * Sketch line. Used when drawing a circle.
    * @type {ol.Feature}
    * @private
    */
   this.sketchLine_ = null;
 
   /**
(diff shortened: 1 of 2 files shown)


tschaub:
Interested to hear what others think on this.  I think the existing style helps suggest that to finish drawing, the ring can be closed.  I wonder if more people would be confused about how to finish drawing with the new style.

ahocevar:
I agree with @tschaub here.

giohappy:
I agree with @tschaub too. The new style seems to suggest that the drawing can be finished without closing the ring

fredj:
OK, thanks for the comments.

After thinking about it, I also think that the style is better as it is

PR #5223, opened 2016-04-13
Create tmpSize_ before calling calculateTileRanges_

fredj:
otherwise the variable is not used in the function and generates garbage


tschaub:
Good catch.  Please merge.

PR #5225, opened 2016-04-13
Remove unused ol.extent.transform2D function

fredj:
The function is not used since https://github.com/openlayers/ol3/pull/1754#issuecomment-36792294

And that's one `goog.vec.Mat4` less to migrate ...


tschaub:

PR #5228, opened 2016-04-13
Add 'layers' option to Translate interaction

adube:
This PR introduces the `layers` option in the `ol.interaction.Translate` interaction to allow features only for the given layers to be managed.

It works exactly like what is done in the `ol.interaction.Select`.


adube:
Ready for review.

fredj:
LGTM, thanks @adube

adube:
Thank you @fredj!

PR #5230, opened 2016-04-14
Reuse frame state arrays

fredj:
Reuse the extent and viewHints arrays from the previous frameState to reduce the number of temporary array

### DIFF
--- a/src/ol/view.js
+++ b/src/ol/view.js
@@ -256,10 +256,17 @@ ol.View.prototype.getCenter = function() {
 
 
 /**
+ * @param {Array.<number>=} opt_hints Destination array.
  * @return {Array.<number>} Hint.
  */
-ol.View.prototype.getHints = function() {
-  return this.hints_.slice();
+ol.View.prototype.getHints = function(opt_hints) {
+  if (opt_hints !== undefined) {
+    opt_hints[0] = this.hints_[0];
+    opt_hints[1] = this.hints_[1];
+    return opt_hints;
+  } else {
+    return this.hints_.slice();
+  }
 };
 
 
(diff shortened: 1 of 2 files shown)


src/ol/view.js
```diff
@@ -256,10 +256,17 @@ ol.View.prototype.getCenter = function() {
 
 
 /**
+ * @param {Array.<number>=} opt_hints Destination array.
  * @return {Array.<number>} Hint.
  */
-ol.View.prototype.getHints = function() {
-  return this.hints_.slice();
+ol.View.prototype.getHints = function(opt_hints) {
```
tschaub: I'm not opposed to this.  But do you see any problem with just returning `this.hints_`?

tschaub:
This looks like a good idea to me.  I didn't check where else `view.getHints()` is called, but I wonder if we could just pass around a single array instead of adding the `opt_hints` arg.

I'm also curious to learn if you are profiling to help identify places where garbage collection can be reduced.  Or if you are just identifying optimization opportunities like this by looking through the code.

fredj:
> I'm also curious to learn if you are profiling to help identify places where garbage collection can be reduced. Or if you are just identifying optimization opportunities like this by looking through the code.

It was very manual; I've added breakpoints in `size.js`, `extent.js`, ... where the optional value is not provided. For example with :
```js
ol.size.buffer = function(size, buffer, opt_size) {
 ...
};
```
I've added a conditional breakpoint if `opt_size == undefined`.

src/ol/view.js
fredj: Not sure about this; `frameState.viewHints` is updated in (at least) the [animation functions](https://github.com/openlayers/ol3/blob/v3.15.1/src/ol/animation.js#L32)

tschaub: Ok, I agree we can't reuse the same hints array everywhere.  Though I think it is a pity that view hints are specific to a frame instead of a view.

If I ever get around to changing how animation works (so view properties are properly animated and we could finally have a "side-by-side" example that look nice), then I think we won't need the extra code here.

tschaub:
This looks good to merge.

fredj:
Thanks for the review

src/ol/view.js
fredj: Thanks for the reminder, #5234

PR #5238, opened 2016-04-15
Add 'jsonp' option for TileUTFGrid sources.

aberenyi:
As Mapbox API v4 dropped JSONP support, XHR support is needed
for TileUTFGrids. 'jsonp' is an optional parameter with the default
 value of false.

See https://github.com/openlayers/ol3/issues/5211


tschaub:
Thanks for this work @aberenyi.  We should have test coverage for the new behavior.  Are you able to create new tests for the XHR path by adapting some of the existing tests?

aberenyi:
@tschaub my pleasure. I'll try to write some tests as soon as I can, however, I'm afraid I cannot give you an estimated time yet.

fredj:
fixed with #5329

Issue #3253, opened 2015-02-19
Touch event default behaviours disabled in control preventing scrolling


kaij:
I added a custom control to a map containing a scrollable list. This list is not scrollable on touch platforms (but works normal on desktop/mouse interaction).

Reason: ol.pointer.TouchSource.prototype.touchmove calls preventDefault on the event, therefore disabling normal touch scrolling behaviour on the target control. Comment out the preventDefault call and everything works.

Why is preventDefault called? If it is really needed, how could I cleanly implement touch scrolling?

kaij:
I found a nice solution by looking at the hierarchy of event propagation. OL handles the touchmove event as last handler. So I catch touchstart,touchmove,touchend for my control and call stopPropagation to prevent the event from going through to OL. Scrolling works now nicely.

tsauerwein:
Nice that you found a work-around! I am not 100% sure anymore why this `preventDefault` was necessary. Maybe to avoid that panning the map scrolls the page, but maybe this also handled else where.

I think you could also show the list not as an ol control, but simply as a container with position absolute and a high z-index. By doing that you should get the event before the listener on the map.

davidpinheiro:
I had the same problem.
Commented preventDefault() and now everything is working fine.
Thanks kaij!
I hope that someone fixes this.

PR #5243, opened 2016-04-18
Overlay: No touch scrolling in ol.Overlay

AnWeber:
Touch scrolling in an overlay is prevented by ol.pointer.TouchSource.prototype.touchmove. If the overlay has set the style properties max-height and overflow: auto, the overlay is not scrollable on touch devices (iOS, Android, Chrome Dev Tools)

Fiddle: http://jsfiddle.net/ttgo4wr0/1/

The workaround in issue #3253 doesn't apply because the position of the overlay should get moved with the map.

A fix for this issue would be to add ol.events.EventType.TOUCHMOVE in the overlayContainerStopEvents array in map.js: c5384a4

### DIFF
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -282,6 +282,7 @@ ol.Map = function(options) {
     ol.events.EventType.DBLCLICK,
     ol.events.EventType.MOUSEDOWN,
     ol.events.EventType.TOUCHSTART,
+    ol.events.EventType.TOUCHMOVE,
     ol.events.EventType.MSPOINTERDOWN,
     ol.MapBrowserEvent.EventType.POINTERDOWN,
     ol.events.EventType.MOUSEWHEEL,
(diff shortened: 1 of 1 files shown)


ahocevar:
This has an unpleasant side effect: when dragging the map and reaching a control with the pointer/finger, dragging will stop. I think this is undesired.

AnWeber:
This side effect is really unpleasant. Another way to fix this issue would be to only stopPropagation of the touchmove Event, when the touchstart event was triggered inside the overlay. Maybe this should be only enabled by options on the overlay. I will check the source for a better implementation.

AnWeber:
The touchmove Event is only attached, if the touchstart event is inside the overlay. Please review the changes.

ahocevar:
I'm not sure what the implications of this are, especially if a touchmove sequence is aborted. Are we guaranteed to always get a touchend after a touchstart? I think that the root of the problem might be somewhere else, and might still somehow be related to #3253. @tsauerwein, could you take a look please? You're more familiar with the pointerevent stack.

Manjulajntuk:
How can i pass user defined external wmsurl and layer to the base layer ? Can you suggest any sample code to work on with reference links for basic understanding sir.

bartvde:
@Manjulajntuk can you stop commenting on unrelated github issues? Please ask on stack overflow instead.

PR #5252, opened 2016-04-19
Add filter option to ol.format.WFS#writeGetFeature

tsauerwein:
[ol.format.WFS#writeGetFeature](http://openlayers.org/en/master/apidoc/ol.format.WFS.html#writeGetFeature) currently only allows to set a bbox filter, this PR proposes to add a `filter` option which supports a sub-set of the OGC filters.

Example:
```javascript
var request = new ol.format.WFS().writeGetFeature({
  ...
  filter: ol.format.wfs.filter.equalTo('name', 'New York')
});
```
Or:
```javascript
var f = ol.format.wfs.filter;
var request = new ol.format.WFS().writeGetFeature({
  ...
  filter: f.or(f.equalTo('name', 'New York'), f.equalTo('area', 1234))
});
```

The following filters are currently supported:

* Not/And/Or
* PropertyIsEqualTo/PropertyIsNotEqualTo
* PropertyIsLessThan/PropertyIsLessThanOrEqualTo
* PropertyIsGreaterThan/PropertyIsGreaterThanOrEqualTo
* PropertyIsLike
* PropertyIsNull
* PropertyIsBetween
* BBOX

### DIFF
--- a/src/ol/format/ogc/filter.js
+++ b/src/ol/format/ogc/filter.js
@@ -1,0 +1,637 @@
+goog.provide('ol.format.ogc.filter');
+goog.provide('ol.format.ogc.filter.Filter');
+goog.provide('ol.format.ogc.filter.Logical');
+goog.provide('ol.format.ogc.filter.LogicalBinary');
+goog.provide('ol.format.ogc.filter.And');
+goog.provide('ol.format.ogc.filter.Or');
+goog.provide('ol.format.ogc.filter.Not');
+goog.provide('ol.format.ogc.filter.Bbox');
+goog.provide('ol.format.ogc.filter.Comparison');
+goog.provide('ol.format.ogc.filter.ComparisonBinary');
+goog.provide('ol.format.ogc.filter.EqualTo');
+goog.provide('ol.format.ogc.filter.NotEqualTo');
+goog.provide('ol.format.ogc.filter.LessThan');
+goog.provide('ol.format.ogc.filter.LessThanOrEqualTo');
+goog.provide('ol.format.ogc.filter.GreaterThan');
+goog.provide('ol.format.ogc.filter.GreaterThanOrEqualTo');
+goog.provide('ol.format.ogc.filter.IsNull');
+goog.provide('ol.format.ogc.filter.IsBetween');
+goog.provide('ol.format.ogc.filter.IsLike');
+
+goog.require('ol.Object');
+
+
+/**
+ * Create a logical `<And>` operator between two filter conditions.
+ *
+ * @param {!ol.format.ogc.filter.Filter} conditionA First filter condition.
+ * @param {!ol.format.ogc.filter.Filter} conditionB Second filter condition.
+ * @returns {!ol.format.ogc.filter.And} `<And>` operator.
+ * @api
+ */
+ol.format.ogc.filter.and = function(conditionA, conditionB) {
+  return new ol.format.ogc.filter.And(conditionA, conditionB);
+};
+
+
+/**
+ * Create a logical `<Or>` operator between two filter conditions.
+ *
+ * @param {!ol.format.ogc.filter.Filter} conditionA First filter condition.
+ * @param {!ol.format.ogc.filter.Filter} conditionB Second filter condition.
+ * @returns {!ol.format.ogc.filter.Or} `<Or>` operator.
+ * @api
+ */
+ol.format.ogc.filter.or = function(conditionA, conditionB) {
+  return new ol.format.ogc.filter.Or(conditionA, conditionB);
+};
+
+
+/**
+ * Represents a logical `<Not>` operator for a filter condition.
+ *
+ * @param {!ol.format.ogc.filter.Filter} condition Filter condition.
+ * @returns {!ol.format.ogc.filter.Not} `<Not>` operator.
+ * @api
+ */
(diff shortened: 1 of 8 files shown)


tsauerwein:
Anyone up for a review? @bartvde maybe, our expert for OGC topics? :)

bartvde:
@tsauerwein I can have a look at this but won't have time until later in this week (mid/end).

tsauerwein:
Thanks, Bart! @fredj will take a look, but a 2nd opinion is welcome!

src/ol/format/wfs/filter.js
```diff
@@ -0,0 +1,617 @@
...
+ * @classdesc
+ * Represents a logical `<And>` operator between two filter conditions.
+ *
+ * @constructor
+ * @param {!ol.format.wfs.filter.Filter} conditionA First filter condition.
+ * @param {!ol.format.wfs.filter.Filter} conditionB Second filter condition.
+ * @extends {ol.format.wfs.filter.LogicalBinary}
+ * @api
+ */
+ol.format.wfs.filter.And = function(conditionA, conditionB) {
```
fredj: Is it really needed to implement this function if it only calls the parent? if it for jsdoc ?

fredj: forget the last comment, slow back from vacation ...

fredj:
LGTM but I would like to have a review from Bart

bartvde:
I'll have a quick look now.

src/ol/format/wfs/filter.js
```diff
@@ -0,0 +1,617 @@
+goog.provide('ol.format.wfs.filter');
```
bartvde: shouldn't we put this in ```ol.format.ogc.filter``` instead? Since filter can be used outside of wfs and is part of the ogc namespace not the wfs one.

src/ol/format/wfs/filter.js
```diff
@@ -0,0 +1,617 @@
...
+ * @returns {!ol.format.wfs.filter.IsNull} `<PropertyIsNull>` operator.
+ * @api
+ */
+ol.format.wfs.filter.isNull = function(propertyName) {
+  return new ol.format.wfs.filter.IsNull(propertyName);
+};
+
+
+/**
+ * Creates a `<PropertyIsBetween>` comparison operator to  test whether an expression
```
bartvde: two spaces after to

src/ol/format/wfsformat.js
```diff
@@ -551,43 +559,179 @@ ol.format.WFS.writeQuery_ = function(node, featureType, objectStack) {
...
+ * @param {Array.<*>} objectStack Node stack.
+ * @private
+ */
+ol.format.WFS.writeIsLikeFilter_ = function(node, filter, objectStack) {
+  goog.asserts.assertInstanceof(filter, ol.format.wfs.filter.IsLike,
+    'must be IsLike comparison filter');
+  node.setAttribute('wildCard', filter.wildCard);
+  node.setAttribute('singleChar', filter.singleChar);
+  node.setAttribute('escapeChar', filter.escapeChar);
+  ol.format.WFS.writeOgcPropertyName_(node, filter.propertyName);
```
bartvde: minor but missing support for matchCase here

bartvde:
Nice work @tsauerwein I only put a few minor remarks in there. Otherwise LGTM.

src/ol/format/wfs/filter.js
tsauerwein: Yes, that makes sense.

src/ol/format/wfsformat.js
tsauerwein: If I get the OGC specification right, `matchCase` is not supposed to be used with `PropertyIsLike`, only with `PropertyIsEqualTo` and `PropertyIsNotEqualTo`. Correct?

bartvde: I've always thought that was a mistake in the schema, and I even reported it as early as 2006 I think: http://www.ogcnetwork.net/node/962
https://trac.osgeo.org/openlayers/ticket/3247
http://osgeo-org.1560.x6.nabble.com/ogc-filter-1-1-doesn-t-support-matchCase-in-PropertyIsLike-td5079577.html

tsauerwein: Ok, thanks for the info. I will also add it to `PropertyIsLike`.

tsauerwein:
Thanks a lot for the reviews! I renamed `ol.format.wfs.filter` to `ol.format.ogc.filter` and added support for the `matchCase` attribute. I will merge when Travis agrees.

bartvde:
thanks @tsauerwein !

src/ol/format/ogc/filter.js
```diff
@@ -0,0 +1,637 @@
...
+
+  goog.base(this);
+
+  /**
+   * @private
+   * @type {!string}
+   */
+  this.tagName_ = tagName;
+};
+goog.inherits(ol.format.ogc.filter.Filter, ol.Object);
```
tschaub: I'm curious why this inherits from `ol.Object`.  Is this used anywhere?

tsauerwein: Good point, there should be no need to inherit from `ol.Object`.

fredj: I'm creating a pull request to fix it

PR #5255, opened 2016-04-20
No special treatment for unmanaged layers in forEachFeatureAtPixel

### LINKED ISSUE #5219: Translate Interaction does not work with unmanaged layers (since 3.14.0)
Looks like https://github.com/openlayers/ol3/pull/4854 created an issue with Translate - if the feature selected to be translated was on an unmanaged layer, it is ignored by `map.forEachFeatureAtPixel` and cannot be translated.

The pull request notes that:

> For user created unmanaged layers, nothing changes, because a user cannot call `ol.Map#skipFeature()`.

but of course, Select still can, which means anything that depends on knowing if a feature has been selected no longer works with an unmanaged layer (e.g. a FeatureOverlay work-alike).

ahocevar:
This pull request simplifies things, but changes the behaviour of `ol.Map#forEachFeatureAtPixel`. It also fixes #5219.

Since this is another one of [those changes](http://ahocevar.net/openlayers/2016/02/17/trilogy-of-error.html), I won't blame anyone if this doesn't pass the review phase.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+#### Map forEachFeatureAtPixel callback changes
+
+Previously, unmanaged layers (i.e. layers that were not configured as `layers` array/collection on the map or added with `ol.Map#addLayer()`), were reported as `null` in the `ol.Map#forEachFeatureAtPixel()` callback. Now, they are reported like regular layers.
+
 ### v3.15.0
 
 #### Internet Explorer 9 support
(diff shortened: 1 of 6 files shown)


ahocevar:
Ok, I went through all issues mentioned in my [blog post](http://ahocevar.net/openlayers/2016/02/17/trilogy-of-error.html) again and tested all mentioned Fiddles against this pull request. Everything worked fine.

Note that the bug fixed with 96676f2 is not related, but was revealed by the additional tests I created to make sure that `ol.interaction.Select#getLayer()` always reports the correct layer.

I can now say with good confidence that this is ready for review.

Burstaholic:
I read your blog post while researching our issue, and I hated to be the bearer of more bad news for you. I'm glad the fix was relatively easy :smile_cat:

ahocevar:
Anyone else willing to review/test this?

fredj:
LGTM, we should probably add a note in the upgrade notes

ahocevar:
Closing this - it makes #5271 even worse. I'm not going to spend more time on the Select interaction, but my guts feeling says that with our skipped feature handling, we buy a lot of trouble. It would be better to just change the style of a selected feature. At least with the canvas renderer, this isn't really slower than skipping features in a replay.

PR #5257, opened 2016-04-21
Add readFeatures and readProjection as MVT experimental API

bjornharrtell:
My motivation to have these exposed are that they are needed if you want to implement a custom tileLoadFunction.


ahocevar:
Looks good, thanks @bjornharrtell. Will merge when green.

PR #5259, opened 2016-04-21
Include squash commit merges in changelog

### LINKED ISSUE #5180: tasks/changelog.sh should take the 'squashed commits' into account
The script doesn't take the [squashed commits](https://github.com/blog/2141-squash-your-commits) into account.

example: https://github.com/openlayers/ol3/pull/5130

ahocevar:
Fixes #5180.

It is not perfect, because we cannot extract the Github user name from squash commit merges. So we use the git author instead, and create a link with a Github user search. Better than nothing I'd say.

### DIFF
--- a/tasks/changelog.sh
+++ b/tasks/changelog.sh
@@ -11,6 +11,12 @@ set -o errexit
 #
 MERGE_RE=Merge\ pull\ request\ #\([0-9]+\)\ from\ \([^/]+\)\/[^\ ]+\ \(.*\)
 
+#
+# Regex to match the squash commit message. Creates capture groups for git
+# author, commit subject and pull request number.
+#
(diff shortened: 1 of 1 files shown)


tsauerwein:
Or maybe a link like this: https://github.com/openlayers/ol3/commits/master?author=ahocevar

Thanks for fixing this!

ahocevar:
Thanks for the review @tsauerwein. A link like you suggested won't work because I don't get the 'ahocevar' from the git log - only 'Andreas Hocevar' or the e-mail address.

tsauerwein:
> The link like you suggested won't work...

Sorry, I totally missed the point with that one... :hushed:

ahocevar:
No problem 😄 - is this good to merge @tsauerwein?

tsauerwein:
:+1:

PR #5261, opened 2016-04-22
add finishCondition to ol.interaction.Draw

### LINKED ISSUE #5260: Add finishFunction to ol.interaction.Draw
My proposal is to add finishFunction  to ol.interaction.Draw options to let the user have the last word before finishDrawing is called. AFAICS there's no other mean to stop finishDrawing after _atFinish has been called, i.e. when a potential coordinate is spot during pointer up handling.

finishFunction would be called only is _atFinish returns true, and it will be passed the pointer event. If it returns true the finishDrawing will be called otherwise the dragging continues.

I have already submitted a PR (#5258) for this. I'm sorry for having missed the contibution guidelines before...

giohappy:
Fixes #5260

### DIFF
--- a/src/ol/interaction/drawinteraction.js
+++ b/src/ol/interaction/drawinteraction.js
@@ -160,6 +160,13 @@ ol.interaction.Draw = function(options) {
    */
   this.maxPoints_ = options.maxPoints ? options.maxPoints : Infinity;
 
+  /**
+   * A function to decide if a potential finish coordinate is permissable
+   * @private
+   * @type {ol.events.ConditionType}
+   */
+  this.finishCondition_ = options.finishCondition ? options.finishCondition : goog.functions.TRUE;
+
   var geometryFunction = options.geometryFunction;
   if (!geometryFunction) {
     if (this.type_ === ol.geom.GeometryType.CIRCLE) {
@@ -404,7 +411,9 @@ ol.interaction.Draw.handleUpEvent_ = function(event) {
     } else if (this.mode_ === ol.interaction.DrawMode.CIRCLE) {
       this.finishDrawing();
     } else if (this.atFinish_(event)) {
-      this.finishDrawing();
+      if (this.finishCondition_(event)) {
+        this.finishDrawing();
+      }
     } else {
       this.addToDrawing_(event);
     }
--- a/test/spec/ol/interaction/drawinteraction.test.js
+++ b/test/spec/ol/interaction/drawinteraction.test.js
@@ -332,6 +332,54 @@ describe('ol.interaction.Draw', function() {
     });
 
   });
+  
+  describe('drawing with a finishCondition', function() {
+    beforeEach(function() {
+      var draw = new ol.interaction.Draw({
+        source: source,
+        type: ol.geom.GeometryType.LINE_STRING,
+        finishFunction: function(event){
+          if (ol.array.equals(event.pixel,[30,20])){
+            return true;
+          }
+          return false;
+        }
+      });
+      map.addInteraction(draw);
+    });
+
+    it('draws a linestring failing to finish it first, the finishes it', function() {
+      var features;
+      // first point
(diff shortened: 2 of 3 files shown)


ahocevar:
You'll need to add the new option to `externs/olx.js`, otherwise it will not work in a minified build and not show up in the documentation.

Also, it would be much desirable if you could add a test for the new feature. If you'd even be able to add an example, it would be a huge bonus.

src/ol/interaction/drawinteraction.js
```diff
@@ -160,6 +160,13 @@ ol.interaction.Draw = function(options) {
    */
   this.maxPoints_ = options.maxPoints ? options.maxPoints : Infinity;
 
+  /**
+   * A function to decide if a potential finish coordinate is permissable
+   * @private
+   * @type {ol.events.ConditionType}
+   */
+  this.finishCondition_ = options.finishCondition ? options.finishCondition : goog.functions.TRUE;
```
ahocevar: Please use `ol.functions.TRUE` instead of `goog.functions.TRUE`.

giohappy: ok

giohappy:
Ok. The other conditions miss both tests and examples too. I will add one for my commit.

ahocevar:
> Ok. The other conditions miss both tests and examples too. I will add one for my commit.

I know, sorry for that. We're slowly catching up on test coverage, so it helps us a lot when new pull requests come with tests added already.

giohappy:
I've committed a test. I've done it for LineString only, but I think it will suffice, dont'you?
I've also added the option to olx.js

ahocevar:
You'll also need to add the new option to the DrawOptions typedef in external/olx.js.

giohappy:
I apologize @ahocevar , it's my first PR on OL3. Next ones will be cleaner ;)

test/spec/ol/interaction/drawinteraction.test.js
```diff
@@ -333,6 +333,54 @@ describe('ol.interaction.Draw', function() {
...
+      simulateEvent('pointermove', 30, 20);
+      simulateEvent('pointerdown', 30, 20);
+      simulateEvent('pointerup', 30, 20);
+
+      // cannot finish on this point
+      simulateEvent('pointerdown', 30, 20);
+      simulateEvent('pointerup', 30, 20);
+
+      features = source.getFeatures();
+      expect(features).to.have.length(1);
```
ahocevar: Shouldn't the length be 0 here?

ahocevar: No worries, thanks for your effort!

giohappy: No, this coordinate is ok. Any other coordinate will be refused.

ahocevar: Ah, misread the condition. Thanks for clarifying.

src/ol/interaction/drawinteraction.js
ahocevar: This one is still missing. Looks good otherwise.

test/spec/ol/interaction/drawinteraction.test.js
giohappy: I wrote the condition with my case in mind :)

giohappy:
Here it is https://github.com/openlayers/cla/pull/4

giohappy:
Mmm, it seems Travis is gone down badly...

marcjansen:
Build on travis restarted.

giohappy:
The test I've pushed is wrong. It was a previous version, with finishFunction. I'll submit a new one.

ahocevar:
Merging now. Will create a follow-up pull request to get rid of `goog.functions.TRUE`. Thanks @giohappy.

PR #5264, opened 2016-04-25
Fix for close draw polygon error

### LINKED ISSUE #5145: Uncaught TypeError when modifying shape right after drawing it.
Hi,

I'd like to modify a polygon after it has been drawn. For this I've created a listener for 'drawend' and another for 'addfeature' events. In the first handler the draw interaction is removed from the map and the feature is stored in a temporary list. With the help of this list I can determine when the feature drawn will be added to the layer's source. When it happens I add a modify interaction to the map.
However when I close the polygon with a single click it produces an error that comes from modify interaction.

jsFiddle: https://jsfiddle.net/tomi85/4qhpsz67/2/
Platform: Windows 8.1 / Chrome 49 (in FireFox it's ok)

How to reproduce the error:
Please draw an arbitrary shape, but when you are ready to close it to finish your polygon, move your mouse exactly right above to the closing point then click (single click) and don't move until you get an error: "Uncaught TypeError: Cannot read property 'length' of null". This error comes from ol.interaction.Modify.prototype.removeVertex_ function, where this.dragSegments_ value is null. This error or bug can be fixed by setting the line (row 191) from "this.dragSegments_ = null;" to "this.dragSegments_ = [];" in modifyinteraction.js. But you should further check this.

Many thanks,
Tomi

matjos:
Fixes #5145

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -194,7 +194,7 @@ ol.interaction.Modify = function(options) {
    * @type {Array}
    * @private
    */
-  this.dragSegments_ = null;
+  this.dragSegments_ = [];
 
   /**
    * Draw overlay where sketch features are drawn.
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks @matjos. Can you submit a [CLA](https://github.com/openlayers/cla) please so we can merge this?

fredj:
Please also change line 550 to `this.dragSegments_.length = 0;`
(see https://github.com/openlayers/ol3/issues/5145#issuecomment-203406222)

matjos:
@ahocevar CLA is mailed.
@fredj changed line 550 in the pullrequest also.

PR #5270, opened 2016-04-26
Load erroneous image in ol.Image#load and ol.ImageTile#load

### LINKED ISSUE #4338: Retry loading failed tiles
Is there any way I can force OL to retry to load a tile if it wasn't found it the first place? 

I saw that I can get the tile in the tileloaderror event, but I didn't managed yet to force a retry only for that particular tile. If a tile is not found OpenLayers doesn't try to get him a second time. I am using a Zoomify tile source with a canvas map renderer.

I saw that was achievable in ol2 like in this [post](http://stackoverflow.com/questions/21516765/is-it-possible-to-define-a-layer-which-will-retry-to-load-e-g-with-exponential). Can this still be done in OL3 ?

In my use case I have control over the server where the tiles are generated, so I want to display to the user something as soon as possible, event if not all the tiles are ready yet, but I need a mechanism to automatically retry loading them until they are eventually found.

fredj:
Allow tiles in `ol.ImageState.ERROR` state to be loaded.

To be able to reload erroneous tiles.

Fixes #4338


src/ol/image.js
```diff
@@ -122,10 +122,11 @@ ol.Image.prototype.handleImageLoad_ = function() {
 
 
 /**
- * Load not yet loaded URI.
+ * Load not yet loaded or erroneous image.
```
tschaub: I think `Load the image or retry if loading previously failed.` would make for clearer docs.  Loading an "erroneous" image makes it sound like the intent is to load the *wrong* image.

fredj: Thanks for the suggestion; I will change

fredj:
Ready to be reviewed, thanks

gjn:
@fredj Will this load ad-infinitum when e.g. connection to internet breaks?

fredj:
@gjn no, this only expose the load function and allow error tiles/images to be loaded again

ahocevar:
I may be missing something, but to me this looks indeed like error tiles will be retried over and over.

fredj:
> I may be missing something, but to me this looks indeed like error tiles will be retried over and over.

This needs to be checked but ol3 doesn't retry to load tiles on error.

ahocevar:
You are right @fredj: https://github.com/openlayers/ol3/blob/v3.16.0/src/ol/tilequeue.js#L115-120 and https://github.com/openlayers/ol3/blob/v3.16.0/src/ol/reproj/tile.js#L321-L323 only call `ol.Tile#load()` when the tile state is IDLE.

That said, this is good to merge.

src/ol/image.js
```diff
@@ -122,10 +122,11 @@ ol.Image.prototype.handleImageLoad_ = function() {
 
 
 /**
- * Load not yet loaded URI.
+ * Load the image or retry if loading previously failed.
```
ahocevar: Maybe add a note that loading is taken care of by the tile queue, and calling this method is only needed for preloading or for reloading in case of an error.

fredj: note added

fredj:
Thanks for the review

PR #5274, opened 2016-04-26
Allow using elements in logo attribution options

samuellapointe:
Follow-up to https://github.com/openlayers/ol3/pull/5193

This commit lets the user use an element object instead of a {src: href} object for an attribution logo when creating a map. This opens a lot of possibilities for that logo, for example setting the target to force the logo to open in a new tab when clicked.

However, there is a problem that I couldn't resolve. When using this method, the chrome debugger console shows a single failed GET request to [current url]/logo. I wasn't able to pinpoint where that request was being made. The example I created has that problem.

### DIFF
--- a/examples/custom-icon.js
+++ b/examples/custom-icon.js
@@ -1,0 +1,27 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+
+var logoElement = document.createElement('a');
+logoElement.href = 'http://www.osgeo.org/';
+logoElement.target = '_blank';
+
+var logoImage = new Image();
+logoImage.src = 'http://www.osgeo.org/sites/all/themes/osgeo/logo.png';
+
+logoElement.appendChild(logoImage);
+
+var map = new ol.Map({
+  layers: [
+    new ol.layer.Tile({
+      source: new ol.source.OSM()
+    })
+  ],
+  target: 'map',
+  view: new ol.View({
+    center: [0, 0],
+    zoom: 2
+  }),
+  logo: logoElement
+});
--- a/examples/custom-icon.js
+++ b/examples/custom-icon.js
@@ -7,7 +7,7 @@ var logoElement = document.createElement('a');
 logoElement.href = 'http://www.osgeo.org/';
 logoElement.target = '_blank';
 
-var logoImage = new Image();
+var logoImage = document.createElement('img');
 logoImage.src = 'http://www.osgeo.org/sites/all/themes/osgeo/logo.png';
 
 logoElement.appendChild(logoImage);
--- a/src/ol/control/attributioncontrol.js
+++ b/src/ol/control/attributioncontrol.js
@@ -303,15 +303,16 @@ ol.control.Attribution.prototype.insertLogos_ = function(frameState) {
 
   var image, logoElement, logoKey;
   for (logoKey in logos) {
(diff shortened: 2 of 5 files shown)


examples/custom-icon.js
```diff
@@ -0,0 +1,27 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+
+var logoElement = document.createElement('a');
+logoElement.href = 'http://www.osgeo.org/';
+logoElement.target = '_blank';
+
+var logoImage = new Image();
```
ahocevar: Without having tried, but maybe changing this line to `var logoImage = document.createElement('img');` fixes the problem you described.

src/ol/control/attributioncontrol.js
```diff
@@ -309,6 +309,9 @@ ol.control.Attribution.prototype.insertLogos_ = function(frameState) {
       var logoValue = logos[logoKey];
       if (logoValue === '') {
         logoElement = image;
+      } else if (goog.dom.isElement(logoValue)) {
```
ahocevar: Please use `logoValue instanceof HTMLElement` instead.

src/ol/map.js
```diff
@@ -1490,6 +1490,8 @@ ol.Map.createOptionsInternal = function(options) {
     var logo = options.logo;
     if (typeof logo === 'string') {
       logos[logo] = '';
+    } else if (goog.dom.isElement(logo)) {
```
ahocevar: Please use `logo instanceof HTMLElement` instead.

ahocevar:
Please disregard all but my first comments. The problem is with keying the logo. You should not add Element logos to the logos cache (which is only for not creating an element over and over). So instead of adding your element to the cache as `logos['logo'] = ...`, don't add it at all.

src/ol/map.js
```diff
@@ -1490,6 +1490,8 @@ ol.Map.createOptionsInternal = function(options) {
...
+      logos['logo'] = logo;
```
ahocevar: Use `logos[goog.getUid(logo).toString()] = logo` instead.

src/ol/control/attributioncontrol.js
ahocevar: This whole block won't work. Instead, add the following before line 306:
```js
if (logos[logoKey] instanceof HTMLElement) {
  logoElements[logoKey] = logos[logoKey];
}

ahocevar:
I have added more comments inline now, which should help you fix this.

examples/custom-icon.js
samuellapointe: Thanks for your suggestion. Unfortunately, doing this didn't fix the problem.

src/ol/map.js
samuellapointe: I tried doing it but I got an error while building, it told me the check was always going to be false since logo was always going to be a string. I probably missed something, I'll continue checking tomorrow.

samuellapointe:
Thanks for your help. I applied your suggestions and everything seems to work correctly now.

samuellapointe:
I forgot to mention, this PR is ready for review.

adube:
Is there something more required for this to be reviewed ?  Thanks.

ahocevar:
This looks good now, thanks @samuellapointe.

adube:
Thank you, @ahocevar.

PR #5275, opened 2016-04-27
Fix feature-layer association on replaced selections

ahocevar:
This pull request fixes the results returned by `ol.interaction.Select#getLayer`. The additional test added by the pull request fails on master.


fredj:
+1

ahocevar:
Thanks for the review @fredj.

PR #5278, opened 2016-04-27
Fix invalid generics type instantiation

fredj:


ahocevar:
Looks good, please merge.

PR #5279, opened 2016-04-27
Using a collection for ol.interaction.Select instead of an unmanaged layer

simonseyock:
Refering to:
https://groups.google.com/forum/#!topic/ol3-dev/Yh7eMfcUFBI

This is just a first pull request to start a discussion about replacing the unmanaged layer of ol.interaction.Select with an ol.Collection. 

There are some uneccessary whitespace changes i will address later. The obvious tests are failing at the moment, all which have to do with the unmanaged layer.

I am still wondering about the addFeatureLayerAssociation_ and  removeFeatureLayerAssociation_ method calls inside of the handleEvent method. Wouldn't it be easier to call these in the addFeature_ and removeFeature_ method? Or am I missing something?

This would also make it possible to call getLayer for features added programmtically to the collection.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3010,8 +3010,7 @@ olx.interaction.PointerOptions.prototype.handleUpEvent;
  * @typedef {{addCondition: (ol.events.ConditionType|undefined),
  *     condition: (ol.events.ConditionType|undefined),
  *     layers: (undefined|Array.<ol.layer.Layer>|function(ol.layer.Layer): boolean),
- *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
- *     dontStyle: (boolean|undefined)
+ *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|null|undefined),
  *     removeCondition: (ol.events.ConditionType|undefined),
  *     toggleCondition: (ol.events.ConditionType|undefined),
  *     multi: (boolean|undefined),
@@ -3063,7 +3062,8 @@ olx.interaction.SelectOptions.prototype.layers;
 
 /**
  * Style for the selected features. By default the default edit style is used
- * (see {@link ol.style}).
+ * (see {@link ol.style}). If set to null the selected features won't be styled
+ * by the interaction.
  * @type {ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined}
  * @api
  */
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3011,6 +3011,7 @@ olx.interaction.PointerOptions.prototype.handleUpEvent;
  *     condition: (ol.events.ConditionType|undefined),
  *     layers: (undefined|Array.<ol.layer.Layer>|function(ol.layer.Layer): boolean),
  *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *     dontStyle: (boolean|undefined)
  *     removeCondition: (ol.events.ConditionType|undefined),
  *     toggleCondition: (ol.events.ConditionType|undefined),
  *     multi: (boolean|undefined),
--- a/src/ol/interaction/selectinteraction.js
+++ b/src/ol/interaction/selectinteraction.js
@@ -149,12 +149,23 @@ ol.interaction.Select = function(opt_options) {
   this.filter_ = options.filter ? options.filter :
       ol.functions.TRUE;
 
+  var style = options.style;
+
+  if (style !== undefined) {
+    if (goog.isFunction(style)) {
+      style = function(resolution) {
+        return /** @type {ol.style.StyleFunction} */ (style)(this, resolution);
+      };
+    }
+  } else {
+    style = ol.interaction.Select.getDefaultStyleFunction();
+  }
+
   /**
    * @private
-   * @type {ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|null}
+   * @type {ol.style.Style|Array.<ol.style.Style>|ol.FeatureStyleFunction|null}
    */
-  this.style_ = (options.style !== undefined) ? options.style :
-      ol.interaction.Select.getDefaultStyleFunction();
+  this.style_ = style;
 
   /**
    * An association between selected feature (key)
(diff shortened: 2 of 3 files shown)


externs/olx.js
```diff
@@ -3011,6 +3011,7 @@ olx.interaction.PointerOptions.prototype.handleUpEvent;
  *     condition: (ol.events.ConditionType|undefined),
  *     layers: (undefined|Array.<ol.layer.Layer>|function(ol.layer.Layer): boolean),
  *     style: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *     dontStyle: (boolean|undefined)
```
ahocevar: You could get rid of this - `style` could be set to `null` to indicate "don't style".

ahocevar:
> I am still wondering about the addFeatureLayerAssociation_ and removeFeatureLayerAssociation_ method calls inside of the handleEvent method. Wouldn't it be easier to call these in the addFeature_ and removeFeature_ method? Or am I missing something?

Would be great if that worked, but it seems to me we don't know a feature's layer in `addFeature_`/`removeFeature_`. Or am I missing something?

simonseyock:
Yeah, true. Didn't thought about that.

I will remove the tests checking for the unmanaged layer. Some new ones might be good.

And btw. for anyone who has not read the topic in the dev group:
This PR should fix:
https://github.com/openlayers/ol3/issues/4882
https://github.com/openlayers/ol3/issues/5232
https://github.com/openlayers/ol3/issues/4729
as the feature is not moved to another layer anymore

simonseyock:
I looked into removing the skipFeature functionality. This seems to me quite big:
1. remove skipFeature and unskipFeature from ol.Map. Those methods are not used anymore.
2. remove the skippedFeatureUids from the frameState object
3. remove the parameter skippedFeaturesHash from ol.render.canvas.Replay.prototype.replay_, ol.render.canvas.Replay.prototype.replay, ol.render.canvas.Replay.prototype.replayHitDetection, ol.render.canvas.ReplayGroup.prototype.forEachFeatureAtCoordinate, ol.render.canvas.ReplayGroup.prototype.replay and ol.render.canvas.ReplayGroup.prototype.replayHitDetection_
4. the same for the other renderers
5. remove the parameter skippedFeatureUids from ol.source.Source.prototype.forEachFeatureAtCoordinate  & ol.source.ImageVector.prototype.forEachFeatureAtCoordinate

I can do this if this PR is wanted.

ahocevar:
I'd like to hear opinions from others before @simonseyock starts digging too deep into this. cc @fredj, @tsauerwein, @bartvde, @tschaub.

bartvde:
I'm in favour of getting rid of all the skipped feature logic since this will simplify the code base a lot. Have we already verified the performance gains of this approach are negligible for canvas?

tsauerwein:
https://github.com/openlayers/ol3/pull/5255#issuecomment-214836638

>  I'm not going to spend more time on the Select interaction, but my guts feeling says that with our skipped feature handling, we buy a lot of trouble. It would be better to just change the style of a selected feature. At least with the canvas renderer, this isn't really slower than skipping features in a replay.

Is that the idea? Instead of marking a feature as skipped, you would change the style so that a feature is not visible? This means the replay group would have to be re-generated, right?

At least for the WebGL renderer this will make a big difference. Currently when skipping a feature, the WebGL renderer will simply do 2 `drawElements` calls instead of a single one. If a new replay group is created, all data has to be re-uploaded to the GPU. I am worried that this might not work that well when changing the style on mouse-hover.

ahocevar:
@tsauerwein:

> Is that the idea? Instead of marking a feature as skipped, you would change the style so that a feature is not visible?

The idea was to be able to render the feature with the select style, and that would be the default after the changes suggested in this pull request.

An alternative is to set the style of the Select interaction to `null`, which means the Select interaction will not change any style on the original layer. Users could then create an (unmanaged) layer with the Select interaction's features collection to render the selected features with a different style on top of the original features.

> At least for the WebGL renderer this will make a big difference. Currently when skipping a feature, the WebGL renderer will simply do 2 drawElements calls instead of a single one.

I know, and it turned out that this is a problem only for the WebGL renderer. Actually, this was the reason why the whole skipFeature logic was introduced a long time ago. I think in most cases, rendering the selected feature a 2nd time on top of the original one with a different style will give good visual hover feedback, while not having to upload a new replay group to the GPU for the original layer.

src/ol/interaction/selectinteraction.js
```diff
@@ -149,12 +149,23 @@ ol.interaction.Select = function(opt_options) {
   this.filter_ = options.filter ? options.filter :
       ol.functions.TRUE;
 
+  var style = options.style;
```
simonseyock: i think it would quite inconsistent with the other interactions to accept an ol.FeatureStyleFunction instead of an ol.style.StyleFunction, even if that would that would make the following lines unnecessary.

ahocevar: I can see pros and cons of both approaches. But I tend to agree.

simonseyock:
The visual feedback of rendering the feature a 2nd time on top of the original one is a good point, this would not be done automatically anymore. But that could be done by the user himself, as ahocevar already said. In my opinion the benefits of being able to have more then one select interactions (see https://groups.google.com/forum/#!topic/ol3-dev/Yh7eMfcUFBI) and maybe needing to organise a visual feedback by myself would outweigh the benefits of not being worried about the visual feedback, but not being able to have more than one select interaction (This actually restricts me in what i can do, the other not). 

And well, there would be a very easy solution to have the feature shown on top of the others: Just give it a style with a higher zIndex.

I can't say anything about the webGL problem, i never used the webgl renderer.

simonseyock:
In my opinion it would be really helpful to be able to add multiple select interactions. Especially because i would not want to write my own select mechanism for tooltips, popups, etc. But even apart from this trivial use cases it is thinkable that slightly different select interactions might be active at the same time. What do you think about the possibility to add multiple select interactions at once?

Another idea i had is to somehow check if a feature already is in an unmanaged layer and then don't remove it from there, but add it to the other layer, too. I don't know if it is possible to check only if a layer is unmanaged or if it would neccessary to check if the layer is used by a select interaction, too. 
(It is not possible to just check getMap as the layers in group layers have this property set to null, too - or do these count as unmanaged layers?)

PR #5285, opened 2016-04-28
Add more tests for the scaleline control

marcjansen:
Title says it all.

Please review, I wasn't 100% sure if all tests actually make sense besides raising the coverage.


ahocevar:
Looks good to me, please merge. Even if not every single test makes sense in practice, it makes it easier to catch regressions.

marcjansen:

PR #5286, opened 2016-04-28
Skip unnecessary render functions

ahocevar:
This fixes an issue introduced by c5f81a57067f4c180f0821c3dcb7f948d4a1f11c in #5177. @tsauerwein, I knew there was a reason for the if condition, but couldn't remember it. I think with the way I have implemented it now, it is much clearer.


ahocevar:
Maybe a bit more explanation: After c5f81a5, 'vector' render mode raises an error, because no image is available for rendering.

tsauerwein:
Ok, that looks good! :+1:

PR #5287, opened 2016-04-28
Add and fix vectortile tests

ahocevar:
The tests added by #5286 were incorrect, because the spy was set on the wrong renderer. This is now fixed, and additional tests are added.


fredj:
LGTM, thanks

PR #5288, opened 2016-04-28
Expose removeVertex_ as removePoint

giohappy:


ahocevar:
This looks great, thanks @giohappy.

PR #5290, opened 2016-04-28
Add extent interaction

tbarsballe:
As described in https://github.com/openlayers/ol3/issues/5283

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2706,6 +2706,47 @@ olx.interaction.DrawOptions.prototype.wrapX;
 
 
 /**
+ * @typedef {{extent: (ol.Extent|undefined),
+ *     boxStyle: (ol.style.Style|Array.<ol.style.Style>|ol.StyleFunction|undefined),
+ *     pointerStyle: (ol.style.Style|Array.<ol.style.Style>|ol.StyleFunction|undefined),
+ *     wrapX: (boolean|undefined)}}
+ * @api
+ */
+olx.interaction.ExtentOptions;
+
+/**
+ * Initial extent. Defaults to no inital extent
+ * @type {ol.Extent|undefined}
+ * @api
+ */
+olx.interaction.ExtentOptions.prototype.extent;
+
+/**
+ * Style for the drawn extent box. 
+ * Defaults to ol.style.Style.createDefaultEditing()[ol.geom.GeometryType.POLYGON]
+ * @type {ol.style.Style|Array.<ol.style.Style>|ol.StyleFunction|undefined}
+ * @api
+ */
+olx.interaction.ExtentOptions.prototype.boxStyle;
+
+/**
+ * Style for the cursor used to draw the extent. 
+ * Defaults to ol.style.Style.createDefaultEditing()[ol.geom.GeometryType.POINT]
+ * @type {ol.style.Style|Array.<ol.style.Style>|ol.StyleFunction|undefined}
+ * @api
+ */
+olx.interaction.ExtentOptions.prototype.pointerStyle;
+
+/**
+ * Wrap the drawn extent across multiple maps in the X direction? 
+ * Only affects visuals, not functionality. Defaults to false.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.interaction.ExtentOptions.prototype.wrapX;
+
+
+/**
  * @typedef {{
  *     features: (ol.Collection.<ol.Feature>|undefined),
  *     layers: (undefined|Array.<ol.layer.Layer>|function(ol.layer.Layer): boolean)
--- a/src/ol/interaction/extent.js
+++ b/src/ol/interaction/extent.js
@@ -1,0 +1,506 @@
+goog.provide('ol.interaction.Extent');
+goog.provide('ol.interaction.ExtentEvent');
+goog.provide('ol.interaction.ExtentEventType');
+
+goog.require('ol');
+goog.require('ol.Feature');
+goog.require('ol.MapBrowserEvent.EventType');
+goog.require('ol.MapBrowserPointerEvent');
+goog.require('ol.coordinate');
+goog.require('ol.events.Event');
+goog.require('ol.extent');
+goog.require('ol.geom.GeometryType');
+goog.require('ol.geom.Point');
+goog.require('ol.geom.Polygon');
+goog.require('ol.interaction.Pointer');
+goog.require('ol.layer.Vector');
+goog.require('ol.source.Vector');
+goog.require('ol.style.Style');
+
+
+/**
+ * @enum {string}
+ */
+ol.interaction.ExtentEventType = {
+  /**
+   * Triggered after the extent is changed
+   * @event ol.interaction.ExtentEvent
+   * @api
+   */
+  EXTENTCHANGED: 'extentchanged'
+};
+
+/**
+ * @classdesc
+ * Events emitted by {@link ol.interaction.Extent} instances are instances of
+ * this type.
+ *
+ * @constructor
+ * @param {ol.Extent} extent the new extent
+ * @extends {ol.events.Event}
+ */
+ol.interaction.ExtentEvent = function(extent) {
+  ol.events.Event.call(this, ol.interaction.ExtentEventType.EXTENTCHANGED);
+
+  /**
+   * The current extent.
+   * @type {ol.Extent}
+   * @api
+   */
+  this.extent_ = extent;
+};
+
+ol.inherits(ol.interaction.ExtentEvent, ol.events.Event);
+
+/**
+ * @classdesc
+ * Allows the user to draw a vector box by clicking and dragging on the map.
+ * Once drawn, the vector box can be modified by dragging its vertices or edges.
+ * This interaction is only supported for mouse devices.
+ *
+ * @constructor
+ * @extends {ol.interaction.Pointer}
+ * @fires ol.interaction.ExtentEvent
+ * @param {olx.interaction.ExtentOptions} opt_options Options.
+ * @api
+ */
+ol.interaction.Extent = function(opt_options) {
+
+  /**
+   * Extent of the drawn box
+   * @type {ol.Extent}
+   * @private
+   */
+  this.extent_ = null;
+
+  /**
+   * Handler for pointer move events
+   * @type {function (ol.Coordinate): ol.Extent|null}
+   * @private
+   */
+  this.pointerHandler_ = null;
+
+  /**
+   * Pixel threshold to snap to extent
+   * @type {number}
+   * @private
+   */
+  this.pixelTolerance_ = 10;
+
+  /**
+   * Last known pixel coordinate of the pointer
+   * @type {ol.Pixel}
+   * @private
+   */
+  this.lastPixel_ = null;
+
+  /**
+   * Is the pointer snapped to an extent vertex
+   * @type {boolean}
+   * @private
+   */
+  this.snappedToVertex_ = false;
+
+  /**
+   * Feature for displaying the visible extent
+   * @type {ol.Feature}
+   * @private
+   */
+  this.extentFeature_ = null;
+
+  /**
+   * Feature for displaying the visible pointer
+   * @type {ol.Feature}
+   * @private
+   */
+  this.vertexFeature_ = null;
+
+  if (!opt_options) {
+    opt_options = {};
+  }
+
+  if (opt_options.extent) {
+    this.setExtent(opt_options.extent);
+  }
+
+  /* Inherit ol.interaction.Pointer */
+  ol.interaction.Pointer.call(this, {
+    handleDownEvent: ol.interaction.Extent.handleDownEvent_,
+    handleDragEvent: ol.interaction.Extent.handleDragEvent_,
+    handleEvent: ol.interaction.Extent.handleEvent_,
+    handleUpEvent: ol.interaction.Extent.handleUpEvent_
+  });
+
+  /**
+   * Layer for the extentFeature
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.extentOverlay_ = new ol.layer.Vector({
+    source: new ol.source.Vector({
+      useSpatialIndex: false,
+      wrapX: !!opt_options.wrapX
+    }),
+    style: opt_options.boxStyle ? opt_options.boxStyle : ol.interaction.Extent.getDefaultExtentStyleFunction_(),
+    updateWhileAnimating: true,
+    updateWhileInteracting: true
+  });
+
+  /**
+   * Layer for the vertexFeature
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.vertexOverlay_ = new ol.layer.Vector({
+    source: new ol.source.Vector({
+      useSpatialIndex: false,
+      wrapX: !!opt_options.wrapX
+    }),
+    style: opt_options.pointerStyle ? opt_options.pointerStyle : ol.interaction.Extent.getDefaultPointerStyleFunction_(),
+    updateWhileAnimating: true,
+    updateWhileInteracting: true
+  });
+};
+
+ol.inherits(ol.interaction.Extent, ol.interaction.Pointer);
+
+/**
+ * @param {ol.MapBrowserEvent} mapBrowserEvent Event.
+ * @return {boolean} Propagate event?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.handleEvent_ = function(mapBrowserEvent) {
+  if (!(mapBrowserEvent instanceof ol.MapBrowserPointerEvent)) {
+    return true;
+  }
+  //display pointer (if not dragging)
+  if (mapBrowserEvent.type == ol.MapBrowserEvent.EventType.POINTERMOVE && !this.handlingDownUpSequence) {
+    this.handlePointerMove_(mapBrowserEvent);
+  }
+  //call pointer to determine up/down/drag
+  ol.interaction.Pointer.handleEvent.call(this, mapBrowserEvent);
+  //return false to stop propagation
+  return false;
+};
+
+/**
+ * @param {ol.MapBrowserPointerEvent} mapBrowserEvent Event.
+ * @return {boolean} Event handled?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.handleDownEvent_ = function(mapBrowserEvent) {
+  var pixel = mapBrowserEvent.pixel;
+  var map = mapBrowserEvent.map;
+
+  var extent = this.getExtent();
+  var vertex = this.snapToVertex_(pixel, map);
+
+  //find the extent corner opposite the passed corner
+  var getOpposingPoint = function(point) {
+    var x_ = null;
+    var y_ = null;
+    if (point[0] == extent[0]) {
+      x_ = extent[2];
+    } else if (point[0] == extent[2]) {
+      x_ = extent[0];
+    }
+    if (point[1] == extent[1]) {
+      y_ = extent[3];
+    } else if (point[1] == extent[3]) {
+      y_ = extent[1];
+    }
+    if (x_ !== null && y_ !== null) {
+      return [x_, y_];
+    }
+    return null;
+  };
+  if (vertex && extent) {
+    var x = (vertex[0] == extent[0] || vertex[0] == extent[2]) ? vertex[0] : null;
+    var y = (vertex[1] == extent[1] || vertex[1] == extent[3]) ? vertex[1] : null;
+
+    //snap to point
+    if (x !== null && y !== null) {
+      this.pointerHandler_ = ol.interaction.Extent.getPointHandler_(getOpposingPoint(vertex));
+    //snap to edge
+    } else if (x !== null) {
+      this.pointerHandler_ = ol.interaction.Extent.getEdgeHandler_(
+        getOpposingPoint([x, extent[1]]),
+        getOpposingPoint([x, extent[3]])
+      );
+    } else if (y !== null) {
+      this.pointerHandler_ = ol.interaction.Extent.getEdgeHandler_(
+        getOpposingPoint([extent[0], y]),
+        getOpposingPoint([extent[2], y])
+      );
+    }
+  //no snap - new bbox
+  } else {
+    vertex = map.getCoordinateFromPixel(pixel);
+    this.setExtent([vertex[0], vertex[1], vertex[0], vertex[1]]);
+    this.pointerHandler_ = ol.interaction.Extent.getPointHandler_(vertex);
+  }
+  return true; //event handled; start downup sequence
+};
+
+/**
+ * @param {ol.MapBrowserPointerEvent} mapBrowserEvent Event.
+ * @return {boolean} Event handled?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.handleDragEvent_ = function(mapBrowserEvent) {
+  this.lastPixel_ = mapBrowserEvent.pixel;
+  if (this.pointerHandler_) {
+    var pixelCoordinate = mapBrowserEvent.coordinate;
+    this.setExtent(this.pointerHandler_(pixelCoordinate));
+    this.createOrUpdatePointerFeature_(pixelCoordinate);
+  }
+  return true;
+};
+
+/**
+ * @param {ol.MapBrowserPointerEvent} mapBrowserEvent Event.
+ * @return {boolean} Stop drag sequence?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.handleUpEvent_ = function(mapBrowserEvent) {
+  this.pointerHandler_ = null;
+  //If bbox is zero area, set to null;
+  var extent = this.getExtent();
+  if (!extent || ol.extent.getArea(extent) === 0) {
+    this.setExtent(null);
+  }
+  return false; //Stop handling downup sequence
+};
+
+/**
+ * Returns the default style for the drawn bbox
+ *
+ * @return {ol.StyleFunction} Default Extent style
+ * @private
+ */
+ol.interaction.Extent.getDefaultExtentStyleFunction_ = function() {
+  var style = ol.style.Style.createDefaultEditing();
+  return function(feature, resolution) {
+    return style[ol.geom.GeometryType.POLYGON];
+  };
+};
+
+/**
+ * Returns the default style for the pointer
+ *
+ * @return {ol.StyleFunction} Default pointer style
+ * @private
+ */
+ol.interaction.Extent.getDefaultPointerStyleFunction_ = function() {
+  var style = ol.style.Style.createDefaultEditing();
+  return function(feature, resolution) {
+    return style[ol.geom.GeometryType.POINT];
+  };
+};
+
+/**
+ * @param {ol.Coordinate} fixedPoint corner that will be unchanged in the new extent
+ * @returns {function (ol.Coordinate): ol.Extent} event handler
+ * @private
+ */
+ol.interaction.Extent.getPointHandler_ = function(fixedPoint) {
+  return function(point) {
+    return ol.extent.boundingExtent([fixedPoint, point]);
+  };
+};
+
+/**
+ * @param {ol.Coordinate} fixedP1 first corner that will be unchanged in the new extent
+ * @param {ol.Coordinate} fixedP2 second corner that will be unchanged in the new extent
+ * @returns {function (ol.Coordinate): ol.Extent|null} event handler
+ * @private
+ */
+ol.interaction.Extent.getEdgeHandler_ = function(fixedP1, fixedP2) {
+  if (fixedP1[0] == fixedP2[0]) {
+    return function(point) {
+      return ol.extent.boundingExtent([fixedP1, [point[0], fixedP2[1]]]);
+    };
+  } else if (fixedP1[1] == fixedP2[1]) {
+    return function(point) {
+      return ol.extent.boundingExtent([fixedP1, [fixedP2[0], point[1]]]);
+    };
+  } else {
+    return null;
+  }
+};
+
+/**
+ * @param {ol.Extent} extent extent
+ * @returns {Array<Array<ol.Coordinate>>} extent line segments
+ * @private
+ */
+ol.interaction.Extent.getSegments_ = function(extent) {
+  return [
+    [[extent[0], extent[1]], [extent[0], extent[3]]],
+    [[extent[0], extent[3]], [extent[2], extent[3]]],
+    [[extent[2], extent[3]], [extent[2], extent[1]]],
+    [[extent[2], extent[1]], [extent[0], extent[1]]]
+  ];
+};
+
+/**
+ * @param {ol.Pixel} pixel cursor location
+ * @param {ol.Map} map map
+ * @returns {ol.Coordinate|null} snapped vertex on extent
+ * @private
+ */
+ol.interaction.Extent.prototype.snapToVertex_ = function(pixel, map) {
+  var pixelCoordinate = map.getCoordinateFromPixel(pixel);
+  var sortByDistance = function(a, b) {
+    return ol.coordinate.squaredDistanceToSegment(pixelCoordinate, a) -
+        ol.coordinate.squaredDistanceToSegment(pixelCoordinate, b);
+  };
+  var extent = this.getExtent();
+  if (extent) {
+    //convert extents to line segments and find the segment closest to pixelCoordinate
+    var segments = ol.interaction.Extent.getSegments_(extent);
+    segments.sort(sortByDistance);
+    var closestSegment = segments[0];
+
+    var vertex = (ol.coordinate.closestOnSegment(pixelCoordinate,
+        closestSegment));
+    var vertexPixel = map.getPixelFromCoordinate(vertex);
+
+    //if the distance is within tolerance, snap to the segment
+    if (Math.sqrt(ol.coordinate.squaredDistance(pixel, vertexPixel)) <=
+        this.pixelTolerance_) {
+
+      //test if we should further snap to a vertex
+      var pixel1 = map.getPixelFromCoordinate(closestSegment[0]);
+      var pixel2 = map.getPixelFromCoordinate(closestSegment[1]);
+      var squaredDist1 = ol.coordinate.squaredDistance(vertexPixel, pixel1);
+      var squaredDist2 = ol.coordinate.squaredDistance(vertexPixel, pixel2);
+      var dist = Math.sqrt(Math.min(squaredDist1, squaredDist2));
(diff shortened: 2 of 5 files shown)


src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+  }
+  var self = this;
+
+  /**
+   * @param {ol.MapBrowserEvent} mapBrowserEvent Event.
+   * @return {boolean} Propagate event?
+   * @this {ol.interaction.Extent}
+   * @private
+   */
+  this.handleEvent = function(mapBrowserEvent) {
```
ahocevar: To match the coding style we use for other components, can you please define all member methods outside the constructor, i.e. `ol.interaction.Extent.prototype.handleEvent = function(mapBrowserEvent) ...`.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+      useSpatialIndex: false,
+      wrapX: !!opts.wrapX
+    }),
+    style: opts.pointerStyle ? opts.pointerStyle : ol.interaction.Extent.getDefaultPointerStyleFunction(),
+    updateWhileAnimating: true,
+    updateWhileInteracting: true
+  });
+
+  /* Other Callbacks */
+  var calculateExtent = function(point1, point2) {
```
ahocevar: You can use `ol.extent.boundingExtent([point1, point2])` instead.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+goog.inherits(ol.interaction.Extent, ol.interaction.Pointer);
+
+/* Public Methods */
+/**
+ * @inheritDoc
+ */
+ol.interaction.Extent.prototype.setMap = function(map) {
+  this.bboxOverlay_.setMap(map);
+  this.vertexOverlay_.setMap(map);
+  ol.interaction.Interaction.prototype.setMap.call(this, map);
```
ahocevar: Use `goog.base(this, 'setMap', map);` instead.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+};
+/**
+ * Returns the current drawn extent in the requested projection
+ *
+ * @param {ol.proj.Projection=} projection Projection to retun the extent in
+ * @return {ol.Extent} Drawn extent in the passed projection, or in the native projection if undefined.
+ */
+ol.interaction.Extent.prototype.getExtent = function(projection) {
```
ahocevar: Looks like all methods from here on could be `@private`.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+  };
+
+  /* Internal Methods */
+  this.snapToVertex = function(pixel, map) {
+    var pixelCoordinate = map.getCoordinateFromPixel(pixel);
+    var sortByDistance = function(a, b) {
+      return ol.coordinate.squaredDistanceToSegment(pixelCoordinate, a) -
+          ol.coordinate.squaredDistanceToSegment(pixelCoordinate, b);
+    };
+    var getSegments = function(extent) {
```
ahocevar: You can use `ol.geom.Polygon.fromExtent` here.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+/**
+ * @classdesc
+ * Events emitted by {@link ol.interaction.ExtentInteraction} instances are instances of
+ * this type.
+ *
+ * @constructor
+ * @extends {ol.events.Event}
+ */
+ol.interaction.ExtentEvent = function() {
+  goog.base(this, 'bbox');
```
ahocevar: For convenience, it would make sense to set the extent on the event object.

And for consistency with other events, it would make sense to define `'bbox'` as ExtentEventType.

externs/olx.js
```diff
@@ -2734,6 +2734,17 @@ olx.interaction.DrawOptions.prototype.wrapX;
 
 
 /**
+ * @typedef {{extent: (ol.Extent|undefined),
+ *     extentProjection: (ol.proj.ProjectionLike|undefined),
```
ahocevar: Name could just be `projection`.

externs/olx.js
```diff
@@ -2734,6 +2734,17 @@ olx.interaction.DrawOptions.prototype.wrapX;
...
+ *     bboxStyle: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
```
ahocevar: Name could be `boxStyle` or just `style`.

externs/olx.js
```diff
@@ -2734,6 +2734,17 @@ olx.interaction.DrawOptions.prototype.wrapX;
...
+ *     pointerStyle: (ol.style.Style|Array.<ol.style.Style>|ol.style.StyleFunction|undefined),
+ *     wrapX: (boolean|undefined)}}
+ * @api
+ */
+olx.interaction.ExtentOptions;
```
ahocevar: In addition to this typedef, you'll need to define each option as `@type` below (see other options in `externs/olx.js`).

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+  this.extent = extent ? extent : null;
+  this.createOrUpdateExtentFeature(extent);
+  this.dispatchEvent(new ol.interaction.ExtentEvent());
+};
+/**
+ * Add a listener for Extent change events
+ *
+ * @param {function (?): ?} cb Event callback
+ */
+ol.interaction.Extent.prototype.addExtentListener = function(cb) {
```
ahocevar: Probably not needed - users can use `ol.interaction.Extent#on()`.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+  ol.events.listen(this, 'bbox', cb);
+};
+/**
+ * Remove a listener for Extent change events
+ *
+ * @param {function (?): ?} cb Event callback
+ */
+ol.interaction.Extent.prototype.removeExtentListener = function(cb) {
```
ahocevar: Probably not needed - users can use `ol.interaction.Extent#un()`.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,429 @@
...
+  }
+  return extent;
+};
+/**
+ * Manually sets the drawn extent, in the requested projection. Uses the native projection if undefined.
+ *
+ * @param {ol.Extent} extent Extent
+ * @param {ol.proj.Projection|string=} projection Projection to return the extent in
+ */
+ol.interaction.Extent.prototype.setExtent = function(extent, projection) {
```
ahocevar: If you decide to make this `@api`, I'd go without the `projection` argument here and the `projection` constructor option, and use the view projection. Otherwise it will be confusing for users, because we do not do such thing elsewhere.

ahocevar:
Thanks @tbarsballe for this pull request. I added a few remarks. Let me know if you have questions.

Would you be able to add a few unit tests? As you can see, our automated tests currently fail when the test coverage drops by more than 0.5%.

marcjansen:
Very nice work! And exceptional nice review, @ahocevar 

👌👏👏

src/ol/interaction/extentinteraction.js
tbarsballe: setExtent is also intended to be public.

Sample use case is if you have an extent that you can set by text field and this interaction, you want each to be able to update the other.

src/ol/interaction/extentinteraction.js
tbarsballe: > For convenience, it would make sense to set the extent on the event object.

Not sure what you mean by this? If you mean that the event should be responsible for calling setExtent, that makes having setExtent as a public mehtod problematic (see above)

ahocevar: I meant you could give the event an `extent` property, so you could do something like
```js
extentInteraction.on('bbox', function(e) {
  console.log(e.extent.toString();
});
```
This also reminds me that `'bbox'` is probably not a good event name. Maybe `'modify'`?

src/ol/interaction/extentinteraction.js
tbarsballe: Fair enough. The projection stuff is all carry-over from the origional use case, which had the view in EPSG:3857 but displayed the extent values in lat/lon. I will remove the projection stuff from this.

src/ol/interaction/extentinteraction.js
tbarsballe: Ah, that makes sense. Will do.

src/ol/interaction/extentinteraction.js
tbarsballe: No, I can't. This returns a list of line segments.

tbarsballe:
I made the changes you suggested @ahocevar (except where otherwise noted in the discussion above) and added test cases.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+ * @param {ol.Extent} extent extent
+ * @returns {ol.Feature} extent as featrue
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.prototype.createOrUpdateExtentFeature = function(extent) {
+  var bboxFeature = this.bboxFeature_;
+
+  if (!extent) {
+    extent = [0,0,0,0];
```
ahocevar: You should be able to use `ol.extent.createEmpty();` here.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+};
+/**
+ * @param {ol.Pixel} pixel cursor location
+ * @param {ol.Map} map map
+ * @returns {ol.Coordinate|null} snapped vertex on extent
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.prototype.snapToVertex = function(pixel, map) {
+  var pixelCoordinate = map.getCoordinateFromPixel(pixel);
```
ahocevar: I'm pretty sure you could simplify this method to
```js
return this.extent_ ?
    this.bboxFeature_.getGeometry().closestPointXY(map.getCoordinateFromPixel(pixel)) :
    null;`
```

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+ * @private
+ */
+ol.interaction.Extent.prototype.createOrUpdateExtentFeature = function(extent) {
+  var bboxFeature = this.bboxFeature_;
+
+  if (!extent) {
+    extent = [0,0,0,0];
+  }
+  if (!bboxFeature) {
+    bboxFeature = new ol.Feature(new ol.geom.Polygon([[
```
ahocevar: What was the reason for not using `ol.geom.Polygon.fromExtent(extent)` here?

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+      [extent[2], extent[3]],
+      [extent[0], extent[3]],
+      [extent[0], extent[1]]
+    ]], ol.geom.GeometryLayout.XY));
+
+    this.bboxFeature_ = bboxFeature;
+    this.bboxOverlay_.getSource().addFeature(bboxFeature);
+  } else {
+    var geometry = (bboxFeature.getGeometry());
+    geometry.setCoordinates([[
```
ahocevar: You may want to use the opposite winding order here if you make the other changes I suggested.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+};
+goog.inherits(ol.interaction.Extent, ol.interaction.Pointer);
+
+
+ /* Internal Methods */
+
+/**
+ * @param {ol.MapBrowserEvent} mapBrowserEvent Event.
+ * @return {boolean} Propagate event?
+ * @this {ol.interaction.Extent}
```
ahocevar: No `@this` annotations should be necessary when you define methods like this.

ahocevar: Ah, scratch that! These handlers are not supposed to be members, but static functions. So `ol.interaction.Extent.handleEvent` instead of `ol.interaction.Extent.prototype.handleEvent` below. Same for the other handlers you pass to the superclass constructor.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+  return false;
+};
+
+/**
+ * @param {ol.MapBrowserPointerEvent} mapBrowserEvent Event.
+ * @return {boolean} Event handled?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.prototype.handleDownEvent = function(mapBrowserEvent) {
```
ahocevar: For `@private` functions, we append `_` to the function name elsewhere.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+ * @classdesc
+ * Events emitted by {@link ol.interaction.ExtentInteraction} instances are instances of
+ * this type.
+ *
+ * @constructor
+ * @param {ol.Extent} extent the new extent
+ * @extends {ol.events.Event}
+ */
+ol.interaction.ExtentEvent = function(extent) {
+  goog.base(this, ol.interaction.ExtentEventType.EXTENTCHANGED);
```
ahocevar: You'll need to set the extent on the class, e.g.
```js
  /**
   * The current extent.
   * @type {ol.Extent}
   * @api
   */
  this.extent = extent;
```

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+  this.extent = null;
+  this.pointerHandler = null;
+
+  this.pixelTolerance_ = 10;
+  this.lastPixel_ = null;
+  this.snapToVertex_ = null;
+  this.snappedToVertex_ = null;
+  this.bboxFeature_ = null;
+  this.vertexFeature_ = null;
+  this.bboxOverlay_ = null;
```
ahocevar: No need to initialize anything here that is initialized elsewhere in the constructor.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+  /**
+   * Extent Extent
+   * @type {ol.Extent}
+   * @private
+   */
+  this.extent = null;
+  this.pointerHandler = null;
+
+  this.pixelTolerance_ = 10;
+  this.lastPixel_ = null;
```
ahocevar: You should add `@type` (and `@private`, if appropriate) annotations for any member properties.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,460 @@
...
+ */
+ol.interaction.Extent.prototype.removePointerFeature = function() {
+  var vertexFeature = this.vertexFeature_;
+  if (vertexFeature) {
+    this.vertexOverlay_.getSource().removeFeature(vertexFeature);
+    this.vertexFeature_ = null;
+  }
+};
+
+/* Public Methods */
```
ahocevar: We do not add comments like this elsewhere.

src/ol/interaction/extentinteraction.js
ahocevar: Oh, I see that you only snap within a threshold. But still, you can combine the above with `ol.extent.closestSquaredDistanceXY` to determine whether you're within the threshold.

ahocevar:
Thanks for your continued effort on this @tbarsballe. This is in good shape already. I added a few more comments, most about coding style deviations from what we use elsewhere, some suggestions for simplifications, and one about fixing the event being emitted.

probins:
this will have to be added to #5194

src/ol/interaction/extentinteraction.js
tbarsballe: This one I adapted straight from modifyinteraction.js. I only snap within a threshold. Additionally, if I am close to a corner, I snap to the vertex rather than the edge.

src/ol/interaction/extentinteraction.js
tbarsballe: None of these are inititalized anywhere else in the constructor - I was getting linter warnings for all of them. These only get set in reaction to events, while the extent is active.

ahocevar: `bboxOverlay_` and `vertexOverlay_` are defined a bit further down in the constructor. Note that all should get a `@type` annotation.

src/ol/interaction/extentinteraction.js
tbarsballe: I was unaware of that function. I will update.

src/ol/interaction/extentinteraction.js
tbarsballe: That makes more sense. I thought I was missing something

src/ol/interaction/extentinteraction.js
tbarsballe: Ah, yes, you are right. I guess I was a bit overzealous with those two.

tbarsballe:
Next batch of fixes is added.
Changed event handlers to static
Added more docstrings to initializations
Added some function calls where applicable
Some code style fixup

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,478 @@
...
+ */
+ol.interaction.ExtentEventType = {
+  /**
+   * Triggered after the extent is changed
+   * @event ol.interaction.ExtentEvent
+   * @api
+   */
+  EXTENTCHANGED: 'extentchanged'
+};
+/**
```
probins: for consistency, there should be blank lines here

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,478 @@
...
+ * Returns the current drawn extent in the requested projection
+ *
+ * @return {ol.Extent} Drawn extent in the passed projection, or in the native projection if undefined.
+ * @api
+ */
+ol.interaction.Extent.prototype.getExtent = function() {
+  return this.extent;
+};
+/**
+ * Manually sets the drawn extent, in the requested projection. Uses the native projection if undefined.
```
ahocevar: Docs need to be updated now that there is no `projection` argument any more.

src/ol/interaction/extentinteraction.js
```diff
@@ -0,0 +1,478 @@
...
+/**
+ * @param {ol.Extent} extent extent
+ * @returns {ol.Feature} extent as featrue
+ * @private
+ */
+ol.interaction.Extent.prototype.createOrUpdateExtentFeature_ = function(extent) {
+  var extentFeature = this.extentFeature_;
+
+  if (!extent) {
+    extent = [0,0,0,0];
```
ahocevar: I'm not sure about this. What about creating a feature without geometry if `extent` is `null`?

tbarsballe: That seemds like a better approach. Will update.

tbarsballe:
Next batch of fixes in.
* Added proper newlines
* Updated docs to remove mention of projection
* Use a feature without geometry if extent is null

Also, sorry for the extended delay getting these changes in, I suddenly ran out of time to work on this mid-May.

tbarsballe:
Rebased against master and updated to conform to changes introduced by #5194.

probins:
pls replace `goog.inherits` with `ol.inherits` as well.

tbarsballe:
Fixed

bartvde:
Having another look at this one now.

bartvde:
@tbarsballe any chance you can add an example to show off this new interaction? TIA

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * Triggered after the extent is changed
+   * @event ol.interaction.ExtentEvent
+   * @api
+   */
+  EXTENTCHANGED: 'extentchanged'
+};
+
+/**
+ * @classdesc
+ * Events emitted by {@link ol.interaction.ExtentInteraction} instances are instances of
```
bartvde: Should be ```{@link ol.interaction.Extent}``` now

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   */
+  this.extent = extent;
+};
+
+ol.inherits(ol.interaction.ExtentEvent, ol.events.Event);
+
+/**
+ * @classdesc
+ * Allows the user to draw a vector box by clicking and dragging on the map.
+ * Once drawn, the vector box can be modified by draging its vertices or edges.
```
bartvde: dragging

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * This interaction is only supported for mouse devices.
+ *
+ * @constructor
+ * @extends {ol.interaction.Pointer}
+ * @fires ol.ExtentEvent
```
bartvde: ```ol.interaction.ExtentEvent```

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * @param {olx.interaction.ExtentOptions} opts Options.
+ * @api
+ */
+ol.interaction.Extent = function(opts) {
```
bartvde: very minor but we use options everywhere else

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @private
+   */
+  this.extent = null;
+
+  /**
+   * pointerHandler_ Handler for mointer move events
+   * @type {function (ol.Coordinate): ol.Extent|null}
+   * @private
+   */
+  this.pointerHandler = null;
```
bartvde: private should be with _ at the end

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * @api
+ */
+ol.interaction.Extent = function(opts) {
+
+  /**
+   * extent Extent
+   * @type {ol.Extent}
+   * @private
+   */
+  this.extent = null;
```
bartvde: private should be with _ at the end

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * @constructor
+ * @extends {ol.interaction.Pointer}
+ * @fires ol.ExtentEvent
+ * @param {olx.interaction.ExtentOptions} opts Options.
+ * @api
+ */
+ol.interaction.Extent = function(opts) {
+
+  /**
+   * extent Extent
```
bartvde: I'd take out the first line since it doesn't add much here, or change it to read one times extent only

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {ol.Extent}
+   * @private
+   */
+  this.extent = null;
+
+  /**
+   * pointerHandler_ Handler for mointer move events
```
bartvde: just Handler for mointer move events, no need to repeat the name of the variable at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {function (ol.Coordinate): ol.Extent|null}
+   * @private
+   */
+  this.pointerHandler = null;
+
+  /**
+   * pixelTolerance_ pixel threshold to snap to extent
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {number}
+   * @private
+   */
+  this.pixelTolerance_ = 10;
+
+  /**
+   * lastPixel_ Last known pixel coordinate of the pointer
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {ol.Pixel}
+   * @private
+   */
+  this.lastPixel_ = null;
+
+  /**
+   * snappedToVertex_ Is the pointer snapped to an extent vertex
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {boolean}
+   * @private
+   */
+  this.snappedToVertex_ = false;
+
+  /**
+   * extentFeature_ Feature for displaying the visible extent
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {ol.Feature}
+   * @private
+   */
+  this.extentFeature_ = null;
+
+  /**
+   * vertexFeature_ Feature for displaying the visible pointer
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {ol.Feature}
+   * @private
+   */
+  this.vertexFeature_ = null;
+
+  if (!opts) {
```
bartvde: since it's optional, it needs to be renamed with the opt_ prefix

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+  /* Inherit ol.interaction.Pointer */
+  ol.interaction.Pointer.call(this, {
+    handleDownEvent: ol.interaction.Extent.handleDownEvent_,
+    handleDragEvent: ol.interaction.Extent.handleDragEvent_,
+    handleEvent: ol.interaction.Extent.handleEvent_,
+    handleUpEvent: ol.interaction.Extent.handleUpEvent_
+  });
+
+  /**
+   * extentOverlay_ Layer for the extentFeature
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+   * @type {ol.layer.Vector}
+   * @private
+   */
+  this.extentOverlay = new ol.layer.Vector({
```
bartvde: missing _

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+      useSpatialIndex: false,
+      wrapX: !!opts.wrapX
+    }),
+    style: opts.boxStyle ? opts.boxStyle : ol.interaction.Extent.getDefaultExtentStyleFunction_(),
+    updateWhileAnimating: true,
+    updateWhileInteracting: true
+  });
+
+  /**
+   * vertexOverlay_ Layer for the vertexFeature
```
bartvde: No need to have the variable name at the front

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+/**
+ * @param {ol.MapBrowserPointerEvent} mapBrowserEvent Event.
+ * @return {boolean} Event handled?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.handleDragEvent_ = function(mapBrowserEvent) {
+  this.lastPixel_ = mapBrowserEvent.pixel;
+  if (this.pointerHandler) {
+    var pixelCoordinate = mapBrowserEvent.map.getCoordinateFromPixel(mapBrowserEvent.pixel);
```
bartvde: why not use ```mapBrowserEvent.coordinate``` here?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * @param {ol.MapBrowserPointerEvent} mapBrowserEvent Event.
+ * @return {boolean} Stop drag sequence?
+ * @this {ol.interaction.Extent}
+ * @private
+ */
+ol.interaction.Extent.handleUpEvent_ = function(mapBrowserEvent) {
+  this.pointerHandler = null;
+  //If bbox is zero area, set to null;
+  var extent = this.getExtent();
+  if (!extent || extent[0] == extent[2] || extent[1] == extent[3]) {
```
bartvde: ```!extent && ol.extent.getArea(extent) > 0``` ?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * Returns the default style for the drawn bbox
+ *
+ * @return {ol.StyleFunction} Default Extent style
+ * @private
+ */
+ol.interaction.Extent.getDefaultExtentStyleFunction_ = function() {
+  var style = ol.style.Style.createDefaultEditing();
+  return function(feature, resolution) {
+    return style[ol.geom.GeometryType.POLYGON];
+  };
```
bartvde: why do we need to create a style function here? Can't we just use the style object directly?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+
+/**
+ * Returns the default style for the pointer
+ *
+ * @return {ol.StyleFunction} Default pointer style
+ * @private
+ */
+ol.interaction.Extent.getDefaultPointerStyleFunction_ = function() {
+  var style = ol.style.Style.createDefaultEditing();
+  return function(feature, resolution) {
```
bartvde: why do we need to create a style function here? Can't we just use the style object directly?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+    return style[ol.geom.GeometryType.POINT];
+  };
+};
+
+/**
+ * @param {ol.Coordinate} fixedPoint corner that will be unchanged in the new extent
+ * @returns {function (ol.Coordinate): ol.Extent} event handler
+ * @private
+ */
+ol.interaction.Extent.prototype.getPointHandler_ = function(fixedPoint) {
```
bartvde: can this be static instead?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+  };
+};
+
+/**
+ * @param {ol.Coordinate} fixedP1 first corner that will be unchanged in the new extent
+ * @param {ol.Coordinate} fixedP2 second corner that will be unchanged in the new extent
+ * @returns {function (ol.Coordinate): ol.Extent|null} event handler
+ * @private
+ */
+ol.interaction.Extent.prototype.getEdgeHandler_ = function(fixedP1, fixedP2) {
```
bartvde: can this be static instead?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+  }
+};
+
+/**
+ * @param {ol.Extent} extent extent
+ * @returns {Array<Array<ol.Coordinate>>} extent line segments
+ * @private
+ */
+ol.interaction.Extent.getSegments_ = function(extent) {
+  return [
```
bartvde: could we use ```ol.geom.Polygon.fromExtent(extent).getCoordinates()``` instead?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+    if (!extent) {
+      extentFeature = new ol.Feature({});
+    } else {
+      extentFeature = new ol.Feature(ol.geom.Polygon.fromExtent(extent));
+    }
+    this.extentFeature_ = extentFeature;
+    this.extentOverlay.getSource().addFeature(extentFeature);
+  } else {
+    if (!extent) {
+      extentFeature.setGeometry(null);
```
bartvde: I wonder if this should this be undefined instead?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ * @inheritDoc
+ */
+ol.interaction.Extent.prototype.setMap = function(map) {
+  this.extentOverlay.setMap(map);
+  this.vertexOverlay_.setMap(map);
+  ol.interaction.Pointer.prototype.setMap.call(this, map);
+};
+
+/**
+ * Returns the current drawn extent in the native projection
```
bartvde: with native do you mean the view projection?

src/ol/interaction/extent.js
```diff
@@ -0,0 +1,506 @@
...
+ *
+ * @return {ol.Extent} Drawn extent in the native projection.
+ * @api
+ */
+ol.interaction.Extent.prototype.getExtent = function() {
+  return this.extent;
+};
+
+/**
+ * Manually sets the drawn extent, useing the native projection.
```
bartvde: using

test/spec/ol/interaction/extent.test.js
```diff
@@ -0,0 +1,157 @@
...
+      expect(interaction.snapToVertex_([230,90], map)).to.eql([50,0]);
+      expect(interaction.snapToVertex_([230,89], map)).to.eql([50,1]);
+      expect(interaction.snapToVertex_([231,90], map)).to.eql([50,0]);
+    });
+  });
+
+  describe('draw extent', function() {
+
+    it('drawing extent works', function() {
+      var interaction = new ol.interaction.Extent();
```
bartvde: these two lines could be in beforeEach?

bartvde:
thanks @tbarsballe I'm done with my review. If you have some time to adapt the PR we can get this in quickly. TIA.

src/ol/interaction/extent.js
tbarsballe: Didn't read the doc / didn't know it was there. Updating.

src/ol/interaction/extent.js
tbarsballe: That is clearer, although the current one is computationally more efficient (though minor enough that it doesn't really matter).

src/ol/interaction/extent.js
tbarsballe: If something changes the default style mid application, I would assume we want everything to update. (Is changing the default style function something that is plausible to occur?)

src/ol/interaction/extent.js
tbarsballe: Note that this is three levels of nested arrays (line secments rather than coordinates)

src/ol/interaction/extent.js
tbarsballe: No issue with making this static. Given that it is a private method, shouldn't really matter either way. I can update (I think - need to double chech how to do static JS/OL3 methods)

src/ol/interaction/extent.js
tbarsballe: Undefined may be marginally more correct. Functionally, it will not make a difference.

src/ol/interaction/extent.js
tbarsballe: Original implementation allowed for providing extents in arbitrary projections, so this is a bit of a holdover from that. But yes, native means the view projection.

tbarsballe:
Updated based on feedback. Example coming shortly.

tbarsballe:
Added extent interaction example.

src/ol/interaction/extent.js
bartvde: No that is not possible since it's not in the API

bartvde:
Thanks @tbarsballe LGTM

PR #5291, opened 2016-04-28
Allow once listeners to dispatch events of same type

### LINKED ISSUE #5282: Once event being removed early causing an error
ol.events.unlistenByKey function is not taking into account pending removals created by key.target.removeEventListener. As a result when it does a splice to remove an event, the parameters of the for loop in the function ol.events.EventTarget.prototype.dispatchEvent are changed. This creates an error on the last iteration where ii is greater than the actual number of listeners. As a result an attempt is made to call a undefined listener at the end of the loop. The removal of the event should wait till the for loop is completed as would seem to be the purpose of the pendingRemovals variable. From my quick glance of the code, the root of the problem would seem to be in the ol.events.unlistenByKey function not taking into account pending removals properly.

ahocevar:
Fixes #5282.

### DIFF
--- a/src/ol/events/eventtarget.js
+++ b/src/ol/events/eventtarget.js
@@ -35,6 +35,12 @@ ol.events.EventTarget = function() {
 
   /**
    * @private
+   * @type {!Object.<string, number>}
+   */
+  this.dispatching_ = {};
+
+  /**
+   * @private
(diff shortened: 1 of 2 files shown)


fredj:
LGTM

ahocevar:
Added a test, will merge when green. Thanks @fredj for the review.

PR #5293, opened 2016-04-29
Block removePoint while dragging

giohappy:
This is to avoid removePoint being called while dragging. In that case the interaction internal data structures get corrupted.

### DIFF
--- a/src/ol/interaction/modifyinteraction.js
+++ b/src/ol/interaction/modifyinteraction.js
@@ -889,7 +889,7 @@ ol.interaction.Modify.prototype.insertVertex_ = function(segmentData, vertex) {
  */
 ol.interaction.Modify.prototype.removePoint = function() {
   var handled = false;
-  if (this.lastPointerEvent_) {
+  if (this.lastPointerEvent_ && this.lastPointerEvent_.type != ol.MapBrowserEvent.EventType.POINTERDRAG) {
     var evt = this.lastPointerEvent_;
     this.willModifyFeatures_(evt);
     handled = this.removeVertex_();


ahocevar:
Would you be able to provide a bit more context, especially on how the internal data structures get corrupted and the issues caused by that? Or even add a test case that fails now and passes with your fix?

giohappy:
I will try to set up a test, anyway I will try to explain it by words. When removePoint is called programmatically while dragging, the vertex is correctly removed and everything continues to work fine until the user release the mouse. At this point the ol.interaction.Modify.handleUpEvent_ calls ol.structs.RBush.update passing the segmentData which, because of vertex removal, is stale. When ol.structs.RBush.update tries to obtain the uid of the segmentData we obtain an assertion error because that uid doesn't exist anymore.
In other words, after modifyend we cannot have a pointerup event.
With this fix I simply avoid this chain of events disabling removePoint while dragging.

ahocevar:
Ok got it, so this is an issue only since removePoint can be called programmatically now.

ahocevar:
Thanks @giohappy. If you can add a test in a follow-up pull request, it would be great. If not, it's not a problem either.

bruceceng:
Does this fix [Issue 5127](https://github.com/openlayers/ol3/issues/5127#issuecomment-213599707)?
I wasn't able to consistently reproduce the behavior, but when I tried to delete densely spaced points of polylines, this (or similar) corruption would often occur following the mouseup event. 

I'm guessing that somehow it was triggering a delete event on one vertex and a drag event on another one at the same time due to their proximity.

Would it help to block drag events if the delete condition is true?

giohappy:
Maybe yes. I haven't replicated it though. Can you give it a try with the master version?

PR #5212, opened 2016-04-11
KML: fix GeometryCollection write

probins:
should fix #3200. I've added a simple test, and done a few tests with my own code, and all seems to be working ok. But it would be better if someone who knows the xml handling better than I do checks my changes.

PR #5296, opened 2016-04-29
Document handling of KML MultiGeometries

probins:
Meant to do this as part of #5212 but forgot.

Docs only; no logic change.


ahocevar:
Nice, thanks!

PR #5302, opened 2016-05-02
Remove use of goog.partial

fredj:
Use Function.prototype.bind instead, see:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Partially_applied_functions_(currying)


gberaudo:
+1

tschaub:
Looks good to me too.

fredj:
Thanks for the review

PR #5310, opened 2016-05-03
Get raster sources working again

tschaub:
This reverts 8e1ffbf91e4430e1ad15d1f035f70b447a549c7d

We should take a bit more time with this, but this is required to get raster sources working again.


fredj:
please merge

tschaub:
I'd like to fully understand why 8e1ffbf91e4430e1ad15d1f035f70b447a549c7d made things not work with the way raster sources create a frame state.  In the meantime, I think it makes sense to revert this so we can deploy working examples.

PR #5313, opened 2016-05-04
Rastersource framestate extent

ahocevar:
Brings back the change from #5230 in a different way, and makes the raster source use a snapshot of the frameState extent.

### DIFF
--- a/src/ol/source/rastersource.js
+++ b/src/ol/source/rastersource.js
@@ -219,15 +219,16 @@ ol.source.Raster.prototype.getImage = function(extent, resolution, pixelRatio, p
     return null;
   }
 
-  if (!this.isDirty_(extent, resolution)) {
+  var currentExtent = extent.slice();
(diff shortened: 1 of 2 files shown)


src/ol/source/rastersource.js
```diff
@@ -219,32 +219,33 @@ ol.source.Raster.prototype.getImage = function(extent, resolution, pixelRatio, p
     return null;
   }
 
-  if (!this.isDirty_(extent, resolution)) {
+  var currentExtent = ol.object.assign({}, extent);
```
tschaub: How about `var currentExtent = extent.slice()` here instead?  This way we end up with the type we expect.

ahocevar: Definitely. Don't know what I was thinking.

tschaub:
Assuming the examples work as expected, please merge.

PR #5315, opened 2016-05-07
Remove unused ol.TileCoordTransformType

probins:
I noticed whilst investigating typedefs that this one is not api and isn't used anywhere else in the library. Any reason for keeping it?


ahocevar:
Thanks for catching this. It's a leftover from a long time ago.

PR #5320, opened 2016-05-09
Add a 'condition' option to ol.interaction.Modify

fredj:


fredj:
FYI, I need this new option to not add a vertex to a linestring when the right click is used:
```js
new ol.interaction.Modify({
  condition: ol.events.condition.mouseActionButton,
  ...
});
```

Thanks for any review

ahocevar:
Wouldn't ol.events.condition.mouseActionButton be a better default than what we currently have? Also, the docs could be a bit more verbose, e.g. 'Condition that determines whether the event will be considered to add a vertex to the sketch.'.

fredj:
> Wouldn't ol.events.condition.mouseActionButton be a better default than what we currently have?

I suspect that `ol.events.condition.mouseActionButton` will break the edition on touch devices (I will check)

> Also, the docs could be a bit more verbose, e.g. 'Condition that determines whether the event will be considered to add a vertex to the sketch.'.

Yes, good suggestion

fredj:
That it: `ol.events.condition.mouseActionButton` always returns false on touch devices

ahocevar:
Ok, thanks for investigating. So let's keep the default as it is.

fredj:
I've updated the code and added a `ol.events.condition.primaryAction` condition that works with all the devices.

I want to do more tests on touch device

fredj:
ready for review, thanks

ahocevar:
Very nice. Please merge.

fredj:
Thanks for the review

PR #3429, opened 2015-03-31
Adding missing provide for SelectFilterFunction

pgiraud:
Please review.

PR #5324, opened 2016-05-09
Remove unneeded goog.provides/requires for typedefs

probins:
This is the first part of my work on typedefs. Independent of any change in where typedefs are stored, there are currently a lot of `goog.provide` and `goog.require` which aren't necessary. Getting rid of them simplifies the code and removes things which look like dependencies but aren't.

In principle, provides/requires aren't necessary for any typedefs, but the compiler throws errors in some cases (I've not been able to work out in which circumstances), so I had to add a few back to my original commit to keep it happy. I'm keeping the 2 commits separate for future reference when merging back into my other branch, but this PR can be squash-merged.


ahocevar:
I didn't track down the others, but at least for `ol.Pixel`, there was still a `goog.provide` after your first commit. Which explains why you needed to add back `goog.require`s in the 2nd commit.

probins:
problem is the `goog.provide` can't be removed from files like `pixel.js` (`raster/pixel.js` is the same), as then there would be no provides or requires, and the file would be discarded by closure-util (unless you specifically add them as compiler options).

ahocevar:
You may want to consider cherry-picking a096c3d to work around the problem of standalone typedefs in separate files.

probins:
this PR is only the first part of the more general project of moving the typedefs out of the code and into one or more separate files. It's not the finished article, but can be merged now independently of the move. The 2nd part will remove the standalone typedefs like `pixel.js`.

ahocevar:
Alright, got it. Thanks a lot for your efforts so far.

fredj:
This change breaks my custom build:
```
node buildtools/build.js .build/examples-all.json .build/examples/all.min.js
info ol Parsing dependencies
info ol Compiling 436 sources
ERR! compile node_modules/openlayers/externs/olx.js:3040: ERROR - Bad type annotation. Unknown type ol.interaction.SelectFilterFunction
ERR! compile  *     filter: (ol.interaction.SelectFilterFunction|undefined),
ERR! compile                 ^
ERR! compile
ERR! compile
ERR! compile node_modules/openlayers/externs/olx.js:3142: ERROR - Bad type annotation. Unknown type ol.interaction.SelectFilterFunction
ERR! compile  * @type {ol.interaction.SelectFilterFunction|undefined}
ERR! compile           ^
ERR! compile
ERR! compile 2 error(s), 0 warning(s)
ERR! compile
Makefile:294: recipe for target '.build/examples/all.min.js' failed
```

probins:
@fredj could you give some more details of the build? This is presumably the same issue as #3429. We can put that one back in again, but I would like to try and understand why this is failing specifically on your build and not on others.

fredj:
I will prepare a minimal example

fredj:
It was an issue with `manage_closure_dependencies` in my build file, sorry for the noise

ahocevar:
Failing builds because of not included typedefs did already happen before this was merged. I'd really like to see all typedefs in a single, always included file rather sooner than later - that would definitely fix the problem.

PR #5328, opened 2016-05-10
Use Node 6 on Travis

tschaub:
If we continue using Node 4 on Travis, we should upgrade to `npm@3`.  I wanted to check first if all would pass with Node 6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: node_js
 
 node_js:
-  - "4"
+  - "6"
 
 cache:
   directories:


ahocevar:
Looks like this works, please merge.

I'll do the same on my local machine.

tschaub:
For anybody who wants to do the same upgrade, you'll want to clear any cached `node_modules` (local, not global).  The change from `npm@2` to `npm@3` is smoother if you start with a fresh install.

I'll clear the cached `node_modules` on Travis.

Issue #4507, opened 2015-12-01
redirect to the latest version of the api doc


fredj:
It would be nice if when browsing an old version of the api doc, a message is displayed to use the latest version.
Something like: "This version of the documentation is outdated! Click here for the latest released version"

ahocevar:
Not sure about this. The API docs have to match the version of the library to be useful, and the reader may not be using the latest version of the library. An alternative solution might be to mention the version in a prominent place on each API docs page.

tonio:
The problem is that Goggle refers to various version of the api when
searching with it.
You never know which version you browse, except with the URL.

Being able to clearly see current version & go to «latest stable» &
«master» should be doable without manually changing url.

On 1 December 2015 at 13:04, Andreas Hocevar <notifications@github.com>
wrote:

> Not sure about this. The API docs have to match the version of the library
> to be useful, and the reader may not be using the latest version of the
> library. An alternative solution might be to mention the version in a
> prominent place on each API docs page.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openlayers/ol3/issues/4507#issuecomment-160947664>.
>



-- 
Antoine Abt

Camptocamp France SAS
http://www.camptocamp.com

ahocevar:
Yes, a link to the latest release API makes sense, and also clearly displaying the release the reader is currently viewing.

ahocevar:
Fixed with #5330.

ahocevar:
Fixed with openlayers/openlayers.github.io#70

PR #5330, opened 2016-05-10
Fix source links, display version and warn about outdated docs

ahocevar:
With the latest changes that put the most recent documentation on http://openlayers/en/master/latest/apidoc, the source links did not work any more. Also, we can now warn when viewing outdated docs, and redirect to the latest. The docs now also show the current version:

version


probins:
so this fixes #4507

fredj:
+1, thanks

probins:
so #4507 is now fixed, right?

ahocevar:
> so #4507 is now fixed, right?

For future releases yes. I'm currently working on an openlayers/openlayers.gihub.io pull request to fix it for old releases too.

PR #5332, opened 2016-05-11
Warn when viewing an outdated example, and offer a redirect to latest

### LINKED ISSUE #5235: Add notice to old release examples about a newer release
There should be a notice that pops up on old release examples (and docs) that points people to the same for the latest release.

ahocevar:
This fixes #5235 for future releases.

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -36,6 +36,11 @@
 
     <div class="container-fluid">
 
+      <div id="latest-check" class="alert alert-warning alert-dismissible" role="alert" style="display:none">
+        <button id="latest-dismiss" type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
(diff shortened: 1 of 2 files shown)


ahocevar:
Still work in progress. Should be a div at the top of the page rather than a confim/alert box.

ahocevar:
This is now ready for review. 

The link to latest will be the same example for the latest version if it exists, and to the examples index otherwise. Once the dialog is dismissed, it will not show up again until the latest version was updated. The result will look like this:

warn-latest

marcjansen:
This will also appear if one os viewing the latest, right? If doable, it'd only show the warning in case a newer version exists. Of you can find the time to change it like this, that'd be much appreciated.

ahocevar:
@marcjansen See https://github.com/ahocevar/ol3/blob/19692ed60a6bf8388a3e7b8350d37bd9ed240902/config/examples/example.html#L104 - the regex test excludes any paths that do not include a version number. So the warning won't be shown for the latest. And it also won't be shown if the latest is v3.15.1 and you're viewing v3.15.1, because then `'{{ olVersion }}'` will not equal `latestVersion`.

marcjansen:
Even better! Convinced, please merge. Thanks.

Issue #5299, opened 2016-04-30
Make typedefs compatible with modular structure


probins:
As discussed in #5226 and briefly in https://github.com/openlayers/ol3/issues/4128#issuecomment-201858324, currently typedefs are defined as part of the `ol` namespace in source files, some of them with a `goog.provide`. Some of them are also `goog.require()`d in source files that refer to them. Typedefs are however only used by the compiler with type checking (and also for reference in the docs), so are not really part of the source code. They are also never assigned a value, and so are not namespaces and are not part of the dependency structure. So they do not need a `goog.provide`, nor is a `goog.require` needed when the types they define are referred to in the code.

So the proposal is to change this as part of the move away from global namespaces towards standard modules. I see 2 distinct phases to this:

- [x] move the typedefs out of the source files, and into a separate `typedefs.js` file, removing any `goog.provide`, and also any `goog.require` referring to them.
- [x] deal with `ol.vec.Mat4.Number` - see #5138

There is a third task, which is to remove the `ol` namespace structure from their names, for example, `ol.AttributionLike` would become just `AttributionLike`. This would not affect people using the supplied builds or those creating a custom build, as apps don't refer to the types. It might however affect someone compiling code together with the library, as their code might refer to these types. So this step can be left until conversion to ES6 modules.

I'll draw up a PR to tackle the first task, and report back if I find any further issues. ~~As a side effect, I believe this will also fix #4945, as the typedefs will no longer rely 'on types defined throughout the library', but be collected in one file that is always included in builds. I will also test this.~~  (Crossed out, as doesn't fix problem.)

gberaudo:
Thanks @probins for going forward on this issue.

A small precision: typedefs are not only in the `ol` namespace but also in subnamespaces, like for example `ol.interaction.DrawGeometryFunctionType`.

I think it is better to keep the typedef symbols unchanged (no phase 2). It would avoid:
- conflicts with javascript;
- conflicts between symbols from different subnamespaces;
- API breakage: all uses of the typedefs would have to be changed in client code.

Regarding phase 1, it will greatly simplify dependency management and I think it is a good move.

probins:
[...]

PR #4931, opened 2016-02-29
Replace typedef by record in externs/olx.js

fredj:
work in progress

TODO:
- [ ] typedefs jsdoc plugin

PR #5335, opened 2016-05-12
Move typedefs to separate file

probins:
As discussed at length in #5299. I've split the changes into separate commits so the changes should be easier to follow.

After this PR, there are no `goog.require`s for typedefs apart from the 2 problem cases. I checked the api docs and it all looks ok.

The WFS typedefs are the only api ones in a sub-sub-namespace. When I put these in typedefs.js, not only were there compiler warnings/errors, but test.js decided to join in the fun and failed with "TypeError: undefined is not an object (evaluating 'ol.format.WFS.FeatureCollectionMetadata')
   in global code at http://127.0.0.1:3001/test/loader.js/home/travis/build/probins/ol3/src/ol/typedefs.js:389". Rather than spend time working out what this error was due to (presumably to do with the namespace), I moved these 2 typedefs to their own file, which resolves all problems with them. IMHO, it would not be a big deal renaming these to `ol.format.WFSFeatureCollectionMetadata` so they could go in `typedefs.js` with all the others. Strictly speaking, it would be an api-breaking change, but is not likely to affect many people and IMHO would not justify a major version number. However, I'll leave that to someone else to decide.

`ol.vec.Mat4.Number` is still outstanding, so #5299 should remain open until #5138 is resolved.

If there are any other outstanding PRs  which change typedefs, they will probably conflict with this one. And future PRs which add typedefs should add them to the new file not to the module code.


ahocevar:
@probins:
> IMHO, it would not be a big deal renaming these to ol.format.WFSFeatureCollectionMetadata so they could go in typedefs.js with all the others. Strictly speaking, it would be an api-breaking change.

I do not think we consider types part of the API, because they are not exposed by the compiled library. Such a change would only affect the few users who compile the library together with their application. If mentioned in the upgrade notes, this should not be a hard change to make for these users.

ahocevar:
I need to look at this a bit closer to see what these circular dependencies are about. But wouldn't these problems all go away if typedefs had just flat names, either directly in the `ol` namespace, or without namespace at all?

probins:
@ahocevar yes, the issues are caused by the namespace, not the actual typedef itself. The namespace is created by `goog.provide`, so the 'owner', as the compiler calls it, is the main file for the namespace, which has a `goog.provide` so will create the namespace if not already present. This leads to 2 issues if there is no `goog.provide` for the namespace in the typedefs file:
* if the owner refers to the typedef, it 'depends' on it so the typedef has to come first but if the typedef file has no `goog.provide` for the namespace the compiler throws a 'namespace defined before its owner' warning
* application code compiled with the library may not use the namespace, so the 'owner' isn't present and the namespace never created, which leads to a compiler error. This case occurs when compiling the examples, as 2 of them refer to `ol.raster.Pixel` but the `ol.raster` namespace is not otherwise used, and never created.

Is that clear?

Several of the sub-namespaces have only non-api namespaces, so these could be renamed and put in the `ol` namespace. For the api ones, I doubt whether many people are using the WFS ones (the number may even be 0), but it's more likely that people are using, say, `ProjectionLike`, and it's a nuisance to have to change these references if the type name is changed - not a show-stopper, but a nuisance.

ahocevar:
Thanks for the good explanation @probins. Like I said, I'm all in favour of renaming typedefs so they do not cause problems.

To move this forward, are you going to make the renaming here, or should that be done in a follow-up pull request?

probins:
It's probably simpler if we do the renaming in a separate PR, then we can have discussion about api-breaking there and not hold up the basic project of moving the typedefs out of the code.

ahocevar:
Alright, thanks @probins.

gberaudo:
Sorry I was at Foss4g-fr conference this week, with a very bad internet connection and I could not follow as much as I would have wanted.

For me the situation is still quite confusing and it would have been better to investigate the full solution before merging.

Are we going to move the typedefs symbols out of any namespace? That would be insane as making the symbols globals they would conflict with any other library or system symbol.
Should everything be in ol? We may still have internal conflicts.
Should it be one more level? Why not two or three or more?

If the issue is only namespace ownership then moving typedefs to a dedicated `olt` hierarchy could have been a solution (it is what we do for `olx` and `oli`). It would have broken all typedefs but we could have provided a migration script for the library itself and for all users of the library.

probins:
not sure I understand the problem

> making the symbols globals they would conflict with any other library or system symbol

typedefs are not exported anywhere, so they aren't symbols, and they aren't globals. 'api' in this case only means that they should appear in the docs.

IMO we can keep them in `ol`. Even when we switch to ES6, they can still stop there; with a `goog.provide`, GCC can continue to use them.

`olx` is different. The names of the options are exported and should not be renamed by the compiler. In any case, #4931 proposes changing these to `@record`.

gberaudo:
They are symbols for the compiler and these symbols could conflict with other symbols (typedefs, enum, constructor...) coming from other libraries. All toplevel symbols are global.

When you are compiling several libraries together, it will fail if there are symbol conflicts.

probins:
so we leave them in the `ol` namespace

tschaub:
Having a dot in the name is arbitrary.  `ol.foo.Bar` is no less likely to conflict than `ol_foo_Bar`.  Both would be problems only when compiling together with a library that happens to pick the same name for a typedef, enum, constructor, etc.  The difference between the two is that the second one is uglier.

I imagine we'll revisit this with ES6 modules.  Having dots in the names will be awkward there.  TypeScript handles this with `export interface`.

probins:
what's the procedure with change logs? Should I have put the changes in `upgrade-notes.md` rather than creating a new 3.16 file?

ahocevar:
> Should I have put the changes in upgrade-notes.md rather than creating a new 3.16 file

Yes indeed. I had missed that during the review, because I did not look at the file name. Thanks for asking.

probins:
ok, another PR coming up ...

PR #5339, opened 2016-05-12
Use a div instead of alert for version warning

### LINKED ISSUE #5338: Scrollbar disappears in API documentation
Screenshots:

workspace 1_014

Some more scrolling:

workspace 1_015

ahocevar:
This is consistent with the suggested examples change in #5332:

warn-latest

With this change, it is also no longer necessary to display the version in the navigation bar. The accidently removed doctype, which is likely the cause for #5338, is also brought back.

Fixes #5338.


marcjansen:
I like this a lot. Thanks and please merge.

tschaub:
Looking good @ahocevar!

Issue #3447, opened 2015-04-01
Format code unconditionally included in builds


tschaub:
When we don't set `manage_closure_dependencies` true, the compiler cannot determine that some of the format code is not used.

For example, the GPX format [calls `ol.xml.makeChildAppender(ol.format.XSD.writeDateTimeTextNode)`](https://github.com/openlayers/ol3/blob/62ce9b44c88fd359d353ec215ed10a8e67adbe44/src/ol/format/gpxformat.js#L795).  The `ol.format.XSD.writeDateTimeTextNode` function [calls `goog.string.padNumber`](https://github.com/openlayers/ol3/blob/62ce9b44c88fd359d353ec215ed10a8e67adbe44/src/ol/format/xsdformat.js#L138-L151).  When I build a profile of the library that doesn't export anything that depends on `ol.format.GPX`, the build includes the `goog.string.padNumber` function even though none of the exported code calls it.  I assume this is because the compiler can not reliably trace code paths into objects like [the `ol.format.GPX.RTE_PARSERS_`](https://github.com/openlayers/ol3/blob/62ce9b44c88fd359d353ec215ed10a8e67adbe44/src/ol/format/gpxformat.js#L310-L327).  I'm guessing that if these parsers were prototype properties instead of functions assigned at run time, the compiler would have an easier time knowing that the exported code does not make them accessible.

In my case, setting `manage_closure_dependencies` true removes 30 KB from the build.  Unfortunately, this option cannot always be used because the `olx.js` externs reference types that are not always defined when only a limited set of the library is compiled.

tschaub:
For anybody interested in digging into this, here are two build configs that demonstrate a significant build size difference with `manage_closure_dependencies`: https://gist.github.com/tschaub/5e26e51ece8bc13167b5

(The difference in the above builds may not strictly be due to the way our formats are authored, but I wanted to provide an example config to start with.)

PR #5341, opened 2016-05-12
Test individual example compilation

gberaudo:
Related to https://github.com/openlayers/ol3/issues/5340.

This PR adds a new task `test-compile-each-example.sh` which allows to compile each example individually and thus detect more errors than the fast combined examples we run with `make ci`.

This task is very time consuming so should be run only on occasions, for examples to prepare a release or after a significant commit.

The task returns `1` on error.

### DIFF
--- a/config/example.json
+++ b/config/example.json
@@ -42,6 +42,6 @@
     "warning_level": "VERBOSE",
     "output_wrapper": "(function(){%output%})();",
     "use_types_for_optimization": true,
-    "manage_closure_dependencies": true
+    "manage_closure_dependencies": false
   }
 }
--- a/examples/earthquake-custom-symbol.js
+++ b/examples/earthquake-custom-symbol.js
@@ -13,7 +13,7 @@ goog.require('ol.style.Stroke');
 goog.require('ol.style.Style');
 
 
-var symbol = [[0, 0], [4, 2], [6, 0], [10, 5], [6, 3], [4, 5], [0, 0]];
+var symbols = [[0, 0], [4, 2], [6, 0], [10, 5], [6, 3], [4, 5], [0, 0]];
 var scale;
 var scaleFunction = function(coordinate) {
   return [coordinate[0] * scale, coordinate[1] * scale];
@@ -39,7 +39,7 @@ var styleFunction = function(feature) {
       fill: new ol.style.Fill({color: 'rgba(255, 153, 0, 0.4)'}),
       stroke: new ol.style.Stroke({color: 'rgba(255, 204, 0, 0.2)', width: 2})
     }));
-    vectorContext.drawGeometry(new ol.geom.Polygon([symbol.map(scaleFunction)]));
+    vectorContext.drawGeometry(new ol.geom.Polygon([symbols.map(scaleFunction)]));
     style = new ol.style.Style({
       image: new ol.style.Icon({
         img: canvas,
--- a/examples/layer-z-index.js
+++ b/examples/layer-z-index.js
@@ -83,6 +83,7 @@ layer0.setMap(map);
 
 function bindInputs(id, layer) {
   var idxInput = document.getElementById('idx' + id);
+  /** @this {Element} */
   idxInput.onchange = function() {
     layer.setZIndex(parseInt(this.value, 10) || 0);
   };
--- a/examples/mapbox-vector-tiles-advanced.js
+++ b/examples/mapbox-vector-tiles-advanced.js
@@ -5,11 +5,6 @@ goog.require('ol.format.MVT');
 goog.require('ol.layer.VectorTile');
 goog.require('ol.proj');
 goog.require('ol.source.VectorTile');
-goog.require('ol.style.Fill');
-goog.require('ol.style.Icon');
-goog.require('ol.style.Stroke');
-goog.require('ol.style.Style');
-goog.require('ol.style.Text');
 goog.require('ol.tilegrid.TileGrid');
 
 
--- a/examples/mapbox-vector-tiles.js
+++ b/examples/mapbox-vector-tiles.js
@@ -5,11 +5,6 @@ goog.require('ol.View');
 goog.require('ol.format.MVT');
 goog.require('ol.layer.VectorTile');
 goog.require('ol.source.VectorTile');
-goog.require('ol.style.Fill');
-goog.require('ol.style.Icon');
-goog.require('ol.style.Stroke');
-goog.require('ol.style.Style');
-goog.require('ol.style.Text');
 goog.require('ol.tilegrid');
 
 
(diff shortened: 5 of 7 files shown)


tschaub:
The reason we went with compiling the examples together is that the build times for individual compilation made waiting for approval on a pull request very painful.

I haven't checked here yet.  What does this mean for the time it takes to run `make-ci`?

marcjansen:
We could run this check only if the pr number is evenly dividable by 10, so most PRs will get fast feedback.

I like it as long as it uncovers probable issues, even though we may get broken builds that stem from a previous unrelated change.

probins:
it doesn't have to be in CI though. It can be run as a separate task.

gberaudo:
The total compilation time is very long, at least 30min on my fast computer, which makes it unreasonable to run by Travis every time. As a result it is not triggered by `make ci` or other.

The task being avaiable, anyone can run it manually as desired.
But I think it could be great to run it before each release. It would then be up to the release manager to fix the issues or release any way.

bartvde:
Right let's add it to the release procedure on the Wiki so the release manager cannot forget about the step.

marcjansen:
Agreed!

gberaudo:
[logs.txt](https://github.com/openlayers/ol3/files/263033/logs.txt)

See attached the logs. Currently, 5 tests have errors and 135 are passing.

gberaudo:
I fixed many errors and one warning about a global use of `this` which was confusing closure.

What is left is:
- one example with many warnings about a global use of `this`;
- some examples with errors in `olx`.

These errors look like (see logs for details):
```
1 getfeatureinfo-layers examples/getfeatureinfo-layers.js
info ol Parsing dependencies
info ol Compiling 318 sources
ERR! compile externs/olx.js:2435: ERROR - Bad type annotation. Unknown type ol.interaction.DragBoxEndConditionType
ERR! compile  *     boxEndCondition: (ol.interaction.DragBoxEndConditionType|undefined)}}
ERR! compile                          ^
ERR! compile 
ERR! compile 
ERR! compile externs/olx.js:2472: ERROR - Bad type annotation. Unknown type ol.interaction.DragBoxEndConditionType
ERR! compile  * @type {ol.interaction.DragBoxEndConditionType|undefined}
ERR! compile           ^
ERR! compile 
ERR! compile externs/olx.js:3818: ERROR - Bad type annotation. Unknown type ol.layer.VectorTileRenderType
ERR! compile  *     renderMode: (ol.layer.VectorTileRenderType|string|undefined),
ERR! compile                     ^
ERR! compile 
ERR! compile externs/olx.js:3854: ERROR - Bad type annotation. Unknown type ol.layer.VectorTileRenderType
ERR! compile  * @type {ol.layer.VectorTileRenderType|string|undefined}
ERR! compile           ^
ERR! compile 
```

Do someone has a clue about the cause of these errors?

probins:
see #4945

gberaudo:
Thanks @probins.

Does that mean we should stop using `manage_closure_dependencies: true` for example builds? In https://github.com/openlayers/ol3/blob/master/doc/tutorials/custom-builds.md is stated the option should always be activated; should we change the wording there?

probins:
only for the sort of minimal examples that create these sort of errors. Those that are causing errors are not maps but just capabilities and other minimal things (I think you can skip index.js :-) ) which don't include the types referred to. Alternatively suppress/ignore that error for those examples.

probins:
btw, if you could use the latest master whenever you next run this, I'd be interested to know if putting the typedefs in one file fixes the typedef errors.

probins:
> the sort of minimal examples

what I mean is that these cases aren't realistic, as you wouldn't compile the library with the code to for example parse a capabilities doc

tschaub:
> Does that mean we should stop using manage_closure_dependencies: true for example builds? In https://github.com/openlayers/ol3/blob/master/doc/tutorials/custom-builds.md is stated the option should always be activated; should we change the wording there?

I think we need to stop using `manage_closure_dependencies` until the library is structured in a way that it is safe to use.  And it may be that we never recommend its use.  This is unfortunate because the compiler is not able to remove as much unused code as it should when we throw the whole library at it (see #3447).

gberaudo:
+1 to stop using `manage_closure_dependencies` for now. I updated the configuration file and now all examples except one are compiling properly.

The remaining example complains about a dangerous use of the global this object:
```
ERR! compile /home/gberaudo/dev/ol3/build/examples/snap.js:51: WARNING - dangerous use of the global this object
ERR! compile     this.select = new ol.interaction.Select();
ERR! compile     ^
ERR! compile 
ERR! compile 
ERR! compile /home/gberaudo/dev/ol3/build/examples/snap.js:51: WARNING - dangerous use of this in static method Modify.init
ERR! compile     this.select = new ol.interaction.Select();
ERR! compile     ^
ERR! compile 
ERR! compile /home/gberaudo/dev/ol3/build/examples/snap.js:52: WARNING - dangerous use of the global this object
ERR! compile     map.addInteraction(this.select);
ERR! compile                        ^
ERR! compile 
```

Should I add annotations to the example to fix the warnings?
Then there is the PR on closure-util: https://github.com/openlayers/closure-util/pull/150.

gberaudo:
@probins you asked for the result of a run on top of your work about moving the typedefs. I rebased on to p of master and tested (with manage_closure_dependencies=false) and it works (only one warning about the global this).

probins:
it should work with manage_closure_dependencies=true too, for the typedefs. You'll still have problems with `ol.layer.VectorTileRenderType` as that's an enum.

gberaudo:
This now returns 1 on errors (warnings are ignored). I updated `tasks/readme.md` and rebased on top of master.

The task compiles all the examples without errors so I think it is now ready for merging.

@probins, I tried restoring managed dependencies but I got these errors:
```
ERR! compile externs/olx.js:1263: ERROR - Bad type annotation. Unknown type ol.control.ScaleLine.Units
ERR! compile  *     units: (ol.control.ScaleLine.Units|string|undefined)}}
ERR! compile                ^
ERR! compile 
ERR! compile 
ERR! compile externs/olx.js:1303: ERROR - Bad type annotation. Unknown type ol.control.ScaleLine.Units
ERR! compile  * @type {ol.control.ScaleLine.Units|string|undefined}
ERR! compile           ^
ERR! compile 
ERR! compile 
ERR! compile 2 error(s),
```

gberaudo:
Updated the PR description.

probins:
yup, this is the same issue as #4945. You could fix it for the enums by doing the same as what we did for typedefs: move them all into one file which is always included in the build script and put them all in `ol` namespace so there are no circular dependency issues.

ahocevar:
Not sure we want enums I. A separate file. With one provide per file and enums tugged on the class they belong to, this issue should be gone.

tschaub:
Let's revisit this after transforming sources to ES modules for the 5.0 release.

PR #5343, opened 2016-05-13
Rename non-api sub-namespace typedefs

probins:
As discussed in #5335, first part of clean-up of typedefs.js. The typedefs for sub-namespaces which only have non-api typedefs (most of them) can be renamed so they're in the `ol` namespace without problem.


ahocevar:
Thanks @probins!

ahocevar:
One thing I noticed after merging that there are now several typedefs that do not follow or camel casing rule. So e.g. instead of `ol.xmlParser`, it should be `ol.XmlParser`. Any chance to get a follow-up pull request to fix those?

probins:
ok, I can do that, though as you'll have no doubt gathered, I simply removed the dots.

PR #5344, opened 2016-05-13
Remove raster enum from test

probins:
and replace with string, which is what would be used in apps. Another unneeded require bites the dust.


ahocevar:
Awesome, thanks @probins.

Issue #5346, opened 2016-05-14
Clarifying 'api' for typedefs/enums/options


probins:
As a follow-up to my suggestion in #5335 that the two `ol.format.WFS.xxx` typedefs be renamed despite their currently being labelled 'api', I think it would be good if the documentation explicitly defined what the supported api for all the object types consists of, in other words what the 'contract' is between the library and the application developer. AFAICS the only documentation for this at the moment is in the [Introduction tutorial](http://openlayers.org/en/latest/doc/tutorials/introduction.html), which mentions the 'exportable properties and methods'. The constructor and static function names should be added to this.

With the constructor options, it is the property names that are important, not the name of the options object itself. The `olx` definitions are included in the docs, and are linked to when part of another options object, for example, http://openlayers.org/en/master/apidoc/ol.source.Vector.html links to `olx.LogoOptions`, so these names are part of the api docs, even though they are never used in application code and not exported.

Similarly, the typedef and enum names are referred to in the docs, but are not used by application code or exported. With enums it is the enumerated values that are used in apps and so are part of the api; for example, the possible values of `ol.OverlayPositioning` are the important thing, not the name of the enum itself. In the case of typedefs, I think the only effect of marking one as 'api' is that it appears in the docs.

As another example, I would suggest getting rid of `ol.raster`. This only consists of 2 typedefs and 1 enum, and in module terms it makes no sense to import a module just for an enum. The enum is only used in `source.raster`, so could be moved there - provided that renaming it and the 2 typedefs doesn't break the api.

Although changing the names of typedefs and enums doesn't affect the majority who use built versions of the library, it might affect people compiling app code together with the library if they refer to the types in their code. So maybe part of the 'contract' is defining under what circumstances they may be changed, and documenting changes in release notes

So, I think the definition of 'api' should be clarified. Is the introduction tutorial the best place for this? Perhaps have a section in the api docs landing page?

ahocevar:
The `@api` *annotation* on typedefs is mostly there to simply include the documentation of the type in the API documentation.

The *API* is what you see in the API docs and what you work with in your applications. The typedefs in the API docs are just links to a section that shows the actual type, e.g. `ol.Pixel` -> `Array.<number>`.

The *names* of these typedefs, as you said correctly, only matter to users who use Closure compiler to compile applications together with the library.

[...]

PR #5353, opened 2016-05-16
Add API definition to docs landing page

probins:
As discussed in #5346, some suggested wording for defining what is and is not 'api'. This isn't very satisfactory, as ISTM illogical to say that the API is defined by the `@api` annotation, but then go on to say that some things, like typedefs, are labelled `@api` but are not part of the API. I think it would be better if there were a different annotation tag for them, such as `@document`, meaning 'should be documented but is not part of api'. However, I don't really have time to look into that atm.

### DIFF
--- a/config/jsdoc/api/index.md
+++ b/config/jsdoc/api/index.md
@@ -39,3 +39,26 @@ Interactions for [vector features](ol.Feature.html)
 [ol.Geolocation](ol.Geolocation.html)<br>
 [ol.Overlay](ol.Overlay.html)<br></td>
 </tr></table>
+
+&nbsp;
+
+#### API change policy
+
+The OpenLayers 3.x API consists of
+* names of classes, class methods and properties
+* names of static functions and constants
+* order and types of function arguments
+* types of function return values
+
+API elements marked as `experimental` provide stable and functioning code, but may change.
+Any changes will be documented in upgrade notes so application code can be changed appropriately
+before using the new version of the library. All other API elements will remain compatible throughout the 3.x releases so that no changes to existing application code are necessary when upgrading to a later version.
+
+*Note*: The API change policy does not cover CSS class names that are used to theme the
+OpenLayers UI.
+
+*Note for Closure Compiler users compiling their application code together with OpenLayers*:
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for addressing this @probins. I think the wording in your suggested change would be suitable for a section in `DEVELOPING.md`. On the API docs landing page, we need something simpler that does not mention `@api` annotations, because they are not visible in the docs.

Maybe something like this could work:
```
#### API change policy

The OpenLayers 3.x API consists of
* names of classes, class methods and properties
* names of static functions and constants
* order and types of function arguments
* types of function return values

API elements marked as `experimental` provide stable and functioning code, but are subject to
change in a way that following upgrade nodes between releases allows for easy updating of
application code. All other API elements will remain compatible throughout the 3.x releases in a
way that no changes to existing application code are necessary when upgrading to a later version.

*Note*: The API change policy does not cover CSS class names that are used to theme the
OpenLayers UI.

*Note for users Closure Compiler users compiling their application code together with OpenLayers*:
The names of type definitions provided by OpenLayers (e.g. `ol.Pixel`) are subject to change. It
is therefore recommended to either use the resolved type as listed in the API docs (e.g.
`Array.<number>` instead of `ol.Pixel`), or pay attention to the upgrade notes, which will lis
name changes of type definitions.

probins:
the docs do include links to the source, but in general I think your text is better than mine, so I've used it with some small changes, mainly for the last bit, as it's not just about the typedefs; the most common `@type` used in the examples is `olx.control.AttributionOptions`.

probins:
one other point: should name changes to non-api typedefs be in the release notes too?

ahocevar:
> should name changes to non-api typedefs be in the release notes too

No - changes to those also should not affect Closure Compiler users (or am I missing something?).

ahocevar:
Thanks @probins, I really like it the way it's in the pull request now. Merging.

probins:
> should name changes to non-api typedefs be in the release notes too
> No - changes to those also should not affect Closure Compiler users

so there shouldn't be any problem merging #5343 :-)

PR #5358, opened 2016-05-18
Handle multiple featureMember elements for different types (WFS format)

tsauerwein:
Parsing a GML document with multiple `featureMember` elements for different types was not handled correctly. `parseFeatures` for example returned `[ol.Feature, ol.Feature, ol.Feature, Array[0], Array[0]]`. The proposed fix makes sure that the returned array only contains valid features.

Closes https://github.com/openlayers/ol3/issues/5357

### DIFF
--- a/test/spec/ol/format/wfsformat.test.js
+++ b/test/spec/ol/format/wfsformat.test.js
@@ -791,6 +791,34 @@ describe('ol.format.WFS', function() {
 
   });
 
+  describe('when parsing multiple feature types separately', function() {
+
+    var lineFeatures, polygonFeatures;
+    before(function(done) {
+      afterLoadText('spec/ol/format/gml/multiple-typenames.xml', function(xml) {
+        try {
+          lineFeatures = new ol.format.WFS({
+            featureNS: 'http://localhost:8080/official',
+            featureType: ['planet_osm_line']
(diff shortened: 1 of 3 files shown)


bartvde:
thanks @tsauerwein this looks good to me, please merge if Travis agrees

test/spec/ol/format/wfsformat.test.js
```diff
@@ -811,6 +839,57 @@ describe('ol.format.WFS', function() {
...
+          features = new ol.format.WFS().readFeatures(xml);
+        } catch (e) {
+          done(e);
+        }
+        done();
+      });
+    });
+
+    it('reads all features', function() {
+      expect(features.length).to.be(3);
```
tsauerwein: I just realized that this test is wrong. If you don't give a featureType you should get the features for all types (5 in this example). That is another issue.

bartvde: Good point, indeed 3+2 should be the result.

tsauerwein: Ok, I updated the branch, @bartvde, would you mind doing a new review?

bartvde:
LGTM

tsauerwein:
Thanks for the review!

PR #5359, opened 2016-05-18
Move upgrade notes to correct place

probins:
correct error in #5335


ahocevar:
Thanks a lot for this fix!

PR #5363, opened 2016-05-18
Convert all typedefs to CamelCase

probins:
as requested in https://github.com/openlayers/ol3/pull/5343#issuecomment-220066248


ahocevar:
Thanks a lot for this effort @probins.

Issue #4455, opened 2015-11-18
Performance of WMTS Dimension changes


thomasmoelhave:
#4389 was recently merged and provides improved handling of WMTS layers when the dimensions change.  #4389 is a reworked version of #3660 that (amont other things) much more closely aligns with ol3 coding and documentation conventions. However, much effort went into tuning the perceived performance of #3660 and the performance of the currently merged solution based on #4389 does not achieve the same performance.

I adjusted the #3660 example to use a window of the same size as the new dynamic WMTS dimension example in ol3 master to allow for an apples-to-apples comparison, so if you fire up `examples/wmts-dimensions.html` from ol3 master and compare it to http://scalgo.com/stuff/ol3/build/examples/smooth-params-big.html at the same window size you can compare them directly.

Zoom in on some area, e.g. the Netherlands and the eastern UK and drag the sea-level slider continuously, and at a decent speed, from one end of the slider range to the other. You should notice a significant decrease in perceived performance. As a result, certain tiles "get stuck" for longer periods in the new version resulting in visible tile boundaries as shown in the attached figure. 
![d3e5f280-8e0d-11e5-8f37-ed7b44d5a70a](https://cloud.githubusercontent.com/assets/179143/11251196/84117298-8e30-11e5-83ba-3da6011a4571.png)
These glitches also appear in #3660, but they persist for much longer in this version. The combined result of this is a degradation of the user experience and the feeling that everything is slow, which is unfortunate as this goes against the fundamental appeal of dynamic WMTS dimensions. I have tested this in both Firefox and Chromium on Ubuntu 14.04.

At the moment I don't understand the structure of @elemoine's work to the extend that I can point to a particular thing that causes this issue, but it's probably worth looking into as the user experience could be better and this patch likely makes it much harder to apply the other patch at the same time. It might be something trivial, or something fundamental.

@elemoine are you already aware of any low-hanging fruit, or of ideas from #3660 that you were unable to port over that I could have a look at?

thomasmoelhave:
We are still stuck on v3.7 due to this issue (since we can no longer cleanly apply our own version due to conflicts with the current solution), so I spent a little more time investigating the perfomance regression.

Specifically I looked at the the network tab in Chromium on Linux while draggin the sea-level slider in the version at http://scalgo.com/stuff/ol3/build/examples/smooth-params-big.html that implements our original submission in #3660 with the version from the latest release at http://openlayers.org/en/latest/examples/wmts-dimensions.html.

[...]

PR #5364, opened 2016-05-19
Make it so things don't flash so much when you change the tile source URL

tschaub:
This generalizes the work in #4389 (originally #3660) and #4527 so that all tile sources can enjoy flash-free rendering when things change.  Instead of a WMTS or WMS specific property, all tile sources have a `key` that is assigned to tiles created by the source (as the `tile.key` property).  The WMTS source uses its `dimensions` property to generate the key (if `dimensions` change, the `key` changes).  The WMS source uses its `params` property to generate the key (if `params` change, the `key` changes).  The URL sources (e.g. `ol.source.XYZ`) use the `url` property to generate the key.

As a result, when you call `source.setUrl()`, the currently rendered tiles *are not cleared*.  Instead, new tiles are loaded and rendered in place of existing tiles after the load.

This means that `source.setUrl()` can (and should) be used for "smooth rendering" on changes (as was the original intent with #3660.  In cases where people want to completely clear currently rendered tiles, `source.refresh()` can be called.  Alternatively, `layer.setSource()` can be used to erase currently rendered tiles and render new tiles as they load.

I'm still interested to see if the perceived performance of interim tile loading/rendering can be improved (see #4455).  But this is a separate issue.  This change only makes it so all `ol.source.Tile` can use the same mechanism.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+#### Rendering change for tile sources
+
+Previously, if you called `source.setUrl()` on a tile source, all currently rendered tiles would be cleared before new tiles were loaded and rendered.  This clearing of the map is undesirable if you are trying to smoothly update the tiles used by a source.  This behavior has now changed, and calling `source.setUrl()` (or `source.setUrls()`) will *not* clear currently rendered tiles before loading and rendering new tiles.  Instead, previously rendered tiles remain rendered until new tiles have loaded and can replace them.  If you want to achieve the old behavior (render a blank map before loading new tiles), you can call `source.refresh()` or you can replace the old source with a new one (using `layer.setSource()`).
+
 #### Move of typedefs out of code and into separate file
 
 This change should not affect the great majority of application developers, but it's possible there are edge cases when compiling application code together with the library which cause compiler errors or warnings. In this case, please raise a GitHub issue. `goog.require`s for typedefs should not be necessary.
--- a/examples/reusable-source.html
+++ b/examples/reusable-source.html
@@ -1,0 +1,12 @@
+---
+layout: example.html
+title: Reusable Source
+shortdesc: Updating a tile source by changing the URL.
+docs: >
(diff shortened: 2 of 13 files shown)


bjornharrtell:
This got me wondering if `source.refresh()` could/should not have the same (optional) feature if possible. I've used vector tiles recently and modify style of the layer dynamically. To redraw the vector tiles with the new style I use `source.refresh()` and in that case I think it would be nice to avoid the clear.

bartvde:
Nice to see this improvement @tschaub !

bartvde:
@tschaub should setUrl be used here as well then? https://github.com/openlayers/ol3/blob/master/examples/wms-time.js#L40

ahocevar:
@bjornharrtell:

> I've used vector tiles recently and modify style of the layer dynamically

When you call `setStyle()` on an `ol.layer.VectorTile`, it will redraw. I think (but haven't tried) that you could achieve the same when you call `layer.changed()`. In both cases, the source won't be cleared.

ahocevar:
Very nice generalisation @tschaub. To me, this looks great. Please merge.

@bartvde: For WMS sources, `updateParams` results in a url change, but is a nicer API because you do not have to build the URL from the params manually. So yes, you are taking @tschaub's words too strictly, because he meant the generic case of `ol.source.Tile`, where there is no params handling.

bjornharrtell:
Thank you @ahocevar for clarifying and sorry for the noise.

tschaub:
Thanks for the reviews.  I've added tests and upgrade notes.  I'll merge when green.

@bartvde - I meant that `source.setUrl()` should be used to achieve smooth transitions rather than some other method (we have a few applications that do a ton of extra work maintaining a "loader" source, registering tile load listeners, trying to guess when loading is done, and finally changing the URL of another source - extra code, unnecessary overhead, and sometimes buggy results).  I agree with @ahocevar that `source.setParams()` should be used for WMS sources to achieve the same.

bartvde:
cool thanks for the clarification and thanks for the PR @tschaub

tschaub:
In case the effect of this is not clear, here's [the example](http://openlayers.org/en/master/examples/reusable-source.html) in action:

less-flashy

And the old behavior:

more-flashy

Issue #5365, opened 2016-05-19
Infinite JS Heap growth when using ol.source.VectorTile


sclose:
When added to the map, instances of ol.source.VectorTile contribute a (if not THE) most major growing allocation of JS Heap.  While viewing vectiles with many features, on a single change:resolution or moveend event, the JS Heap can increase by as much as 100 to 200mb.  This seems to happen regardless of tileLoadFunction, styling features, or other parameters that may or may not be set on the source.VectorTile or layer.VectorTile itself.

```
var layer = new ol.layer.VectorTile({
  source: new ol.source.VectorTile({
       format: new ol.format.TopoJSON(),
       projection: 'EPSG:3857',
       tileGrid: ol.tilegrid.createXYZ({
            maxZoom: 19
       }),
  url: someurlhere
  }),
  opacity: 0.7,
  preload: preload,
  style: somestylehere
  });
```

Fortunately, application of ...

`map.getView().on('change:resolution', function() {
	layer.getSource().refresh()});`

...or similar work OK as an invoked garbage collector, but perhaps a pretty sizable memory leak may be occurring when using this source and/or layer type.  

Other source types temporarily increase JS Heap, but OL seems to reset these within a few seconds after map resize/move....  as soon as new tiles are finished loading, but the same does not occur when using source.VectorTile.  [using OL version 3.14.  I have not tested v3.15, and I don't recall the issue occurring with our previous library- v3.8.1- however 3.8.1 used the deprecated ol.source.TileVector].

ahocevar:
This is because `ol.source.VectorTile` caches way more than just the tile image. It also caches the replay group and the features for each tile. What you can do to reduce heap usage is configure a smaller `cacheSize` on your `ol.source.VectorTile`. The default is 128 tiles.

sclose:
Tested and confirmed, thanks AH

PR #5367, opened 2016-05-19
Remove temporary array creation in ol.source.VectorTile#getTile

fredj:
See #5365


ahocevar:
I think this will break things when the tile grid's tile size is stored as array, because it might modify the tile size on the tile grid then.

fredj:
You're right, I forgot this case ...

PR #5368, opened 2016-05-19
Move WFS typedefs to typedefs.js

probins:
A further step in cleaning up typedefs. As these are marked 'api', I've added a note to the release notes. If people agree with the principle of renaming these things so their names no longer include sub-namespaces, I'll change the others. For the new names, I've followed the principle of leaving the sub-namespace in the name where there's ambiguity, for example, `ol.xml.Parser` becomes `ol.XmlParser` not `ol.Parser`, but removing it where there is no ambiguity, as with these two WFS names where there's no need to preserve 'format'.


gberaudo:
If these typedefs make sense in the context of a format, I would keep `format` in the name. I think it is not related to ambiguity.

ahocevar:
I like the names in this pull request. It's consistent with what we do with names in `externs/olx.js`.

gberaudo:
In the case of `ol.WFSFeatureCollectionMetadata` and `ol.WFSTransactionResponse` I think there are good arguments to do the proposed change: the typedefs could be reused outside of the formaters. I don't know if it is a good systematic change though, hence my remark.

That being said, if it tastes better this way, the change is fine with me.

probins:
it's what I did for the WKT and KML changes. ISTM it's obvious that they're to do with formats. Also `layer.LayerState` - do you really gain anything by making that `LayerLayerState`? OTOH creating `ol.Parser` doesn't tell you much - what Parser? So `ol.XmlParser` seems better in this case.

tschaub:
Can't recall if I left this comment elsewhere, but I think we should have a dedicated annotation for documenting symbols instead of overloading the `@api` annotation.  The `@api` annotation should only be used for things that are exportable - and this defines our API.  In my opinion, we should not expand the concept of the API to include annotations specific to the Closure Compiler.  Instead, the API is JavaScript only.

probins:
@tschaub yeah, that was my suggestion in https://github.com/openlayers/ol3/pull/5353#issue-154974267

PR #5369, opened 2016-05-19
format/igc: Remove duplicate regex exec() call

Turbo87:
HFDTE_RECORD_RE_.exec(line) has already been called above


fredj:
Thanks @Turbo87

PR #5375, opened 2016-05-23
Ability to define ol.style.Circle without a fill or stroke color

fredj:
If a stroke style is defined by not the color; use the default `ol.render.canvas.defaultStrokeStyle` value.
Same for the fill style.

### DIFF
--- a/src/ol/style/circlestyle.js
+++ b/src/ol/style/circlestyle.js
@@ -238,7 +238,12 @@ ol.style.Circle.prototype.render_ = function(atlasManager) {
   var strokeWidth = 0;
 
   if (this.stroke_) {
-    strokeStyle = ol.color.asString(this.stroke_.getColor());
+    strokeStyle = this.stroke_.getColor();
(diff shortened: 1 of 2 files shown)


ahocevar:
Wouldn't it be more intuitive to allow an ol.style.Circle constructed without `fill` and `stroke`?

tschaub:
> Wouldn't it be more intuitive to allow an ol.style.Circle constructed without fill and/or stroke?

Then people would need to explicitly pass `null` to avoid having one or the other.  E.g.

```js
// default radius, stroke, and fill
var circle = new ol.style.Circle();

// default radius and fill, no stroke
var circle = new ol.style.Circle({
  stroke: null
});

// default radius and stroke, no fill
var circle = new ol.style.Circle({
  fill: null
});
```

I think this would make for a nice API.  If we want to go this way, we should be consistent across all symbolizers.  Right now I think we have a mix - some sensible defaults, other places where you have to be explicit or you get nothing.

Issue #5376, opened 2016-05-23
jspdf ol3 conflicts / pdf example broken due to redirections of jspdf.js


dermoench42:
I started playing around with openlayers and run into an issue regarding the proposed jspdf use to export the map.
I tried the current example and it didn't work in firefox 46 (gentoo-linux). I tried with my own map and find out:
When loading jspdf.debug.js after ol-debug.js the browsers reloads the map.html file at a randomly place while debugging without any hint.
When loading jspdf.debug.js after ol-debug.js the map won't appear with an debugging message "ol is not defined".

Im using local copies from ol and jspdf

Even if I comment out all pdf-creating codelines the map reloads or won't appear. So it is an issue at load time / initialisation.

Unfortunately I feel overwhelmed because I'm not an experienced javascript developer yet, and I have no Idea where to look further or how to debug this issue.

My map code: [editor.shtml.txt](https://github.com/openlayers/ol3/files/277094/editor.shtml.txt)

BTW in the example http://openlayers.org/en/latest/examples/export-pdf.html?q=pdf the script reference http://mrrio.github.io/jsPDF/dist/jspdf.min.js now redirects to https://parall.ax/products/jspdf/dist/jspdf.min.js which ist a html document and therefore would not work either.

ervin (dermoench42)

dermoench42:
I found out: jspdf uses require.js.
If it is loaded before ol3 then this piece of code:

(function (root, factory) {
  if (typeof exports === "object") {
    module.exports = factory();
  } else if (typeof define === "function" && define.amd) {
    define([], factory);
  } else {
    root.ol = factory();
  }
}

if the define function from require is available than no ol object is generated, and an error is thrown.

Loading jspdf after ol3 produces an event around the pdf save logic ehich causes a reload.

How to get both together?
ervin

bartvde:
Issue was resolved, question is better suited for stack overflow

PR #5377, opened 2016-05-23
Fix link to jspdf.min.js in export-pdf example

fredj:
See #5376


ahocevar:
Thanks, please merge.

fredj:
Thanks for the review

PR #5382, opened 2016-05-23
Handle UTC midnight wrap around, fixes #5371

### LINKED ISSUE #5371: Midnight wraparound in IGC Format
The IGC format reader is parsing the start date from the `HFDTE` record and is using it to calculate fix times. At the moment the reader isn't accounting for the UTC midnight wraparound though which leads to wrong timestamps if a flight log starts e.g. at 23:00:00 UTC and ends at 01:00:00 UTC on the next UTC day.

/cc @twpayne

twpayne:


fredj:
LGTM, thanks Tom !

PR #5390, opened 2016-05-24
Fix the requestEncoding automatic selection

### LINKED ISSUE #5380: Invalid requestEncoding in ol.source.WMTS.optionsFromCapabilities
To reproduce the error, edit `examples/wmts-hidpi.js` and remove the `requestEncoding` option in `ol.source.WMTS.optionsFromCapabilities`.

The `KVP` encoding is selected but the server only support `REST`

fredj:
If the `requestEncoding` parameter was not provided, the function used `KVP`
but without checking if it was supported.

fixes #5380

### DIFF
--- a/test/spec/ol/source/wmtssource.test.js
+++ b/test/spec/ol/source/wmtssource.test.js
@@ -162,6 +162,7 @@ describe('ol.source.WMTS', function() {
           var options = ol.source.WMTS.optionsFromCapabilities(
               capabilities, {
                 layer: 'Demographics_USA_Population_Density',
+                requestEncoding: 'KVP',
                 matrixSet: 'default028mm'
               });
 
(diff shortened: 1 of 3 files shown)


fredj:
The test had to be changed from KVP to REST because REST is the first allowed value.
See https://github.com/openlayers/ol3/blob/v3.16.0/src/ol/source/wmtssource.js#L301-L302

test/spec/ol/source/wmtssource.test.js
```diff
@@ -157,7 +157,7 @@ describe('ol.source.WMTS', function() {
       });
     });
 
-    it('can create KVP options from spec/ol/format/wmts/arcgis.xml',
+    it('can create REST options from spec/ol/format/wmts/arcgis.xml',
```
ahocevar: While you're at it, you could split this test into two - one for KVP (with `requestEncoding` set in the read options) and one for REST.

fredj: done

ahocevar:
Nice! Is this still work in progress? If not, please merge.

fredj:
> Nice! Is this still work in progress? If not, please merge.

I was removing the [wip] from the title :-) thanks for the review

PR #5395, opened 2016-05-25
Allow definition of an offset for an ol.source.Zoomify

jacmendt:
The pull-request just adds an offset parameter to the `options` of the zoomify source. Equal to `imageExtent` for `ol.source.ImageStatic` this option allows to me to display multiple zoomify layers next to each other in one map.


tschaub:

PR #5399, opened 2016-05-26
Remove ol.raster namespace

probins:
The next and penultimate part of typedef move.

The namespace is only used currently for 2 typedefs and 1 enum, so it serves little purpose.
So the 2 typedefs are renamed in the ol namespace, and the enum moved to rastersource.js

@ahocevar if this looks familiar, it's much as you proposed in https://github.com/openlayers/ol3/commit/a096c3dae6b187e2226e1a8fffbd794540ff0b20. The main difference is that you renamed the enum to `ol.source.RasterOperationType`. I prefer `ol.RasterOperationType` so it's consistent with `ol.RasterOperation`. And in any case, the sub-namespaces will be disappearing as part of conversion to modules.

I generated apidocs, and the refs all look ok.


probins:
now rebased to account for #5427. As `ol.raster.Pixel` has now been removed, the `ol.raster` namespace now makes even less sense.

ahocevar:
Thanks @probins.

Issue #5416, opened 2016-06-02
Replace `@api` annotation with `@doc` for types


probins:
As mentioned in https://github.com/openlayers/ol3/pull/5368#issuecomment-220344323 and https://github.com/openlayers/ol3/pull/5353#issue-154974267, `@api` should be limited to what is exported and usable in user apps. Types, such as typedefs, enums, and options objects, are not exported, but are currently labelled `@api` so they appear in the docs. This should be changed to a new `@doc` annotation for those types which should appear in the docs but not be exported.

I can probably find time for this this month, but it would help if someone could point me in the right direction. Do we need a new `config/jsdoc/api/plugins/` similar to `api.js`? Anything else needed?

gberaudo:
Is this complexity due to the `custom builds` feature?
Do we have stats about the usage of this feature?

ahocevar:
> Is this complexity due to the custom builds feature?

No. `@api` was meant to mark exportable symbols (i.e. those that the compiler does not mangle). It does not matter whether you use all of them (default full build) or a subset (custom build).

ahocevar:
> This should be changed to a new @doc annotation for those types which should appear in the docs but not be exported.

Maybe the doc generator could be made smart enough to automatically include all typedefs, enums and options objects that are used by exportable functions/properties. Something similar is done already with superclasses whose constructors are not exportable.

probins:
so you mean remove the `@api` for these types, and make `jsdoc` output them without any annotation?

ahocevar:
> so you mean remove the @api for these types, and make jsdoc output them without any annotation?

Yes - but it's been a while since I've written the jsdoc plugins for our doc generation, so I cannot provide much guidance at this point unfortunately.

gberaudo:
>> Is this complexity due to the custom builds feature?

> No. @api was meant to mark exportable symbols (i.e. those that the compiler does not mangle). It does not matter whether you use all of them (default full build) or a subset (custom build).

The compiler [now understands](https://github.com/google/closure-compiler/wiki/Releases#june-9-2015-v20150609) the `@export` annotation for properties. With this annotation, exports are automatically generated by the compiler.Could drastically simplify the build process.

probins:
`@export`s are always exported though, which isn't wanted for custom builds

probins:
> but it's been a while since I've written the jsdoc plugins for our doc generation, so I cannot provide much guidance at this point unfortunately.

[...]

PR #5425, opened 2016-06-02
Change typedefs from @api to @doc

probins:
As discussed in #5416.

Simply copying the current api plugin to another one called doc seems to work just fine. In fact, it's so simple, I'm sure there must be something wrong :-) I regenerated the docs, and all seems to be exactly as before.

If people agree with this, I'll change the olx typedefs as well. And then look into enums. Anything else that should not be `@api`?

### DIFF
--- a/config/jsdoc/api/plugins/doc.js
+++ b/config/jsdoc/api/plugins/doc.js
@@ -1,0 +1,130 @@
+/**
+ * Define an @doc tag
+ */
+var conf = env.conf.stability;
+var defaultLevels = ["deprecated","experimental","unstable","stable","frozen","locked"];
+var levels = conf.levels || defaultLevels;
+var util = require('util');
+exports.defineTags = function(dictionary) {
+  dictionary.defineTag('doc', {
+    mustHaveValue: false,
+    canHaveType: false,
+    canHaveName: false,
+    onTagged: function(doclet, tag) {
+      var level = tag.text || "experimental";
+      if (levels.indexOf(level) >= 0) {
+        doclet.stability = level;
+      } else {
+        var errorText = util.format('Invalid stability level (%s) in %s line %s', tag.text, doclet.meta.filename, doclet.meta.lineno);
+        require('jsdoc/lib/jsdoc/util/error').handle( new Error(errorText) );
+      }
+    }
+  });
+};
+
+
+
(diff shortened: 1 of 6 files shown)


gberaudo:
Is copying the recommended way to achieve this?

config/jsdoc/api/plugins/doc.js
```diff
@@ -0,0 +1,130 @@
+/**
+ * Define an @doc tag
+ */
+var conf = env.conf.stability;
+var defaultLevels = ["deprecated","experimental","unstable","stable","frozen","locked"];
```
ahocevar: I think we don't need to handle levels here.

probins: there is the stable/unstable distinction though. Is that not what this is handling?

ahocevar:
I feel a bit uncomfortable with this. The `doc` plugin repeats a lot of work that is done by the `api` plugin, and not related to the new `doc` annotation. I'll try to wrap my head around this and see if we can change things so referenced enums, object types and typedefs are always documented.

config/jsdoc/api/plugins/doc.js
ahocevar: It is, but I don't think we need to handle that for things that are not part of the API.

probins: I suppose there is the question of what stable/unstable means in the case of typedefs

probins: in which case, maybe we should just remove `stable` from typedefs (and other non-api types)

ahocevar: yes, and they should not show up as "experimental" in the docs.

probins:
ok, so I'll leave that up to you for the moment. If you like, I can remove the `stable` and remove the level logic from the `@doc` plugin

ahocevar:
> ok, so I'll leave that up to you for the moment.

Thanks for your work so far! I'll see if I can find a quick solution. If not, I'd appreciate if you could finish this. But I'll let you know, no later than tomorrow.

ahocevar:
@probins, I was able to implement what I had in mind. See #5427.

probins:
closed in favour of #5427.

PR #5426, opened 2016-06-02
Fix icon example to anchor popup to feature location

frankrowe:
In the  [icon example](http://openlayers.org/en/latest/examples/icon.html), the icon popup is anchored to the coordinates of the mouse click. This meant that if you zoomed in after clicking, the popup would not be in the correct location (demonstrated below).

screen shot 2016-06-02 at 3 25 14 pm

This PR fixes the example by anchoring to the feature location instead of the click location, and adding an offset to the overlay.

### DIFF
--- a/examples/icon.js
+++ b/examples/icon.js
@@ -59,7 +59,8 @@ var element = document.getElementById('popup');
 var popup = new ol.Overlay({
   element: element,
   positioning: 'bottom-center',
-  stopEvent: false
+  stopEvent: false,
+  offset: [0, -50]
 });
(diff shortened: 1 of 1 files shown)


marcjansen:
I can reproduce the issue. Also looks like a lean fix to me. Do we have a CLA from you?

Also:  many thanks for noticing this, bringing it to our attention and fixing it 👌

frankrowe:
No problem! :smiley:

I just filled out the CLA, so you should have it now.

marcjansen:
Ok, thanks!

PR #5427, opened 2016-06-03
Document typedefs and enums used by API symbols

ahocevar:
With this change, it is no longer necessary to add `@api` annotations to typedefs and enums that are used by API symbols.

This is an alternative to #5425.

### DIFF
--- a/src/ol/typedefs.js
+++ b/src/ol/typedefs.js
@@ -32,7 +32,6 @@ goog.provide('ol.style.AtlasBlock');
 
 /**
  * @typedef {string|Array.<string>|ol.Attribution|Array.<ol.Attribution>}
- * @api
  */
 ol.AttributionLike;
 
@@ -54,7 +53,6 @@ ol.CanvasFillState;
  *
  * @typedef {function(this:ol.source.ImageCanvas, ol.Extent, number,
  *     number, ol.Size, ol.proj.Projection): HTMLCanvasElement}
- * @api
  */
 ol.CanvasFunctionType;
 
@@ -90,7 +88,6 @@ ol.CenterConstraintType;
  * alpha should be a float in the range 0..1 inclusive. If no alpha value is
  * given then `1` will be used.
  * @typedef {Array.<number>}
- * @api
  */
 ol.Color;
 
@@ -100,7 +97,6 @@ ol.Color;
  * Represents a color, pattern, or gradient.
  *
  * @typedef {string|CanvasPattern|CanvasGradient}
- * @api
  */
 ol.ColorLike;
 
@@ -108,7 +104,6 @@ ol.ColorLike;
 /**
  * An array of numbers representing an xy coordinate. Example: `[16, 48]`.
  * @typedef {Array.<number>} ol.Coordinate
- * @api stable
  */
 ol.Coordinate;
 
@@ -118,7 +113,6 @@ ol.Coordinate;
  * `{string}`.
  *
  * @typedef {function((ol.Coordinate|undefined)): string}
- * @api stable
  */
 ol.CoordinateFormatType;
 
@@ -126,7 +120,6 @@ ol.CoordinateFormatType;
 /**
  * An array of numbers representing an extent: `[minx, miny, maxx, maxy]`.
  * @typedef {Array.<number>}
- * @api stable
  */
 ol.Extent;
 
@@ -142,7 +135,6 @@ ol.Extent;
  *
  * The function is responsible for loading the features and adding them to the
  * source.
- * @api
  * @typedef {function(this:ol.source.Vector, ol.Extent, number,
  *                    ol.proj.Projection)}
  */
@@ -156,7 +148,6 @@ ol.FeatureLoader;
  *
  * @typedef {function(this: ol.Feature, number):
  *     (ol.style.Style|Array.<ol.style.Style>)}
- * @api stable
  */
 ol.FeatureStyleFunction;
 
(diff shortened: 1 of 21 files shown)


probins:
looks good - a considerable simplification. I'll pull in your branch and see what the generated docs look like.

probins:
what about the typedefs in olx? These options objects are also not exported, so are not part of the api, so should also not have the `@api` annotation.

ahocevar:
> what about the typedefs in olx? These options objects are also not exported, so are not part of the api, so should also not have the @api annotation.

Done in b28af3b.

gberaudo:
Nice work @ahocevar!

probins:
I've generated the docs and it all looks very good - haven't spotted anything missing, anyway :-)

Should there be something in the release notes about this? Shouldn't affect anybody's code, but it is a change to what was previously annotated as 'api'.

ahocevar:
If anyone still uses the generate-externs task (cc @adube), then there is one difference: the generated externs file does not contain the typedefs any more. I don't know if this could become a problem, but on the other hand I don't know if it would be acceptable to remove the generate-externs task entirely.

probins:
I notice one small difference: `ol.raster.Pixel` is missing. This is presumably because it isn't used anywhere in the code, though it is used in a couple of examples.

ahocevar:
So I need to address this [`TODO`](https://github.com/openlayers/ol3/pull/5427/files#diff-f3a2c4319211773b9b02ed80089d14bbR82).

ahocevar:
The case of `ol.raster.Pixel` was a bit more complicated, because it is 1) only referenced by another typedef and 2) only used in a `{function(argType):returnType}` typedef, which is not parsed at all by JSDoc. So I updated the documentation to not `@link` to `ol.raster.Pixel`, and I think we should be good now.

probins:
so is `ol.raster.Pixel` still needed at all?

This will conflict with #5399, so I'll rebase that when this is merged.

I'm also thinking it would be good to add a note to `doc/tutorials/closure.md` stating that developers can use any of the types defined in the library, but should be aware that those not marked `@api` (which I think is now just the constructor functions) may be renamed or otherwise changed.

src/ol/typedefs.js
```diff
@@ -632,34 +605,31 @@ ol.interaction.SnapSegmentDataType;
...
- * will be called with an array of {@link ol.raster.Pixel} data and should
- * return an array of the same.  For `'image'` type operations, functions will
- * be called with an array of {@link ImageData
- * https://developer.mozilla.org/en-US/docs/Web/API/ImageData} and should return
- * an array of the same.  The operations are called with a second "data"
- * argument, which can be used for storage.  The data object is accessible
- * from raster events, where it can be initialized in "beforeoperations" and
- * accessed again in "afteroperations".
+ * will be called with an array of numbers representing pixel data (e.g.
+ * [R, G, B], [R, G, B, A] or [H, C, L]) and should return an array of the same.
```
tschaub: For `pixel` type operations, the function will be called with an array of pixels, where each pixel is an array of four numbers (`[r, g, b, a]`) in the range of 0 - 255.

If you don't want to use the word "pixel" in the docs, then you'd have to say the raster operation will be called with an array of arrays.  But it is not correct to say that it will be called with an array of numbers (and the length of the pixel arrays will always be 4 instead of sometimes 3 and sometimes 4).

ahocevar: Thanks @tschaub, I fixed the docs. It is fine to use the word "pixel", I was just (obviously) confused.

ahocevar:
@probins:

> so is ol.raster.Pixel still needed at all?

No, I just removed it.

> I'm also thinking it would be good to add a note to doc/tutorials/closure.md stating that developers can use any of the types defined in the library, but should be aware that those not marked @api (which I think is now just the constructor functions) may be renamed or otherwise changed.

I think we should not encourage anyone to use anything that we do not consider API, especially since our few Closure compiler users seem to use more than what's `@api` at their own risk anyway. We have covered the API change policy in `config/api/index.md`.

src/ol/typedefs.js
```diff
@@ -632,39 +605,30 @@ ol.interaction.SnapSegmentDataType;
...
+ * and should return an array of the same.  The operations are called with a
+ * second "data" argument, which can be used for storage.  The data object is
+ * accessible from raster events, where it can be initialized in
+ * "beforeoperations" and accessed again in "afteroperations".
  *
- * @typedef {function((Array.<ol.raster.Pixel>|Array.<ImageData>), Object):
- *     (Array.<ol.raster.Pixel>|Array.<ImageData>)}
- * @api
+ * @typedef {function((Array.<Array.<number>>|Array.<ImageData>), Object):
+ *     (Array.<Array.<number>>|Array.<ImageData>)}
```
tschaub: It's clear that the compiler isn't helping us here.  This should be:
```
function((Array<Array<number>>|Array<ImageData>), Object): (Array<number>|ImageData)
```

That is, a raster operation takes either an array of pixels or an array of `ImageData` and returns either a single pixel or a single `ImageData` object.  For a `pixel` type operation, the function takes an array of pixels and returns a single pixel.  For an `image` type operation, the function takes an array of `ImageData` and returns a single `ImageData` object.

ahocevar: Does this mean that both documentation and typedef are wrong, even before this change?

ahocevar: @tschaub I have changed this, and updated the prose docs accordingly.

ahocevar:
Can this be merged?

tschaub:
Looks good to me (thought I didn't generate the docs locally).  Thanks for the work on this @ahocevar.

This leaves with a simpler definition of the `@api` annotation.  We add `@api` to functions, methods, and properties that are part of the (JavaScript) API.  Where the value is `stable`, we only break backwards compatibility with a new major release.  The options for a constructor, for example, are stable if the constructor is annotated with `@api stable`.

PR #5429, opened 2016-06-03
Remove enums from tests

### LINKED ISSUE #5418: Remove enums from tests
As discussed in https://github.com/openlayers/ol3/commit/a096c3dae6b187e2226e1a8fffbd794540ff0b20#commitcomment-17469103 tests should be using the string literals as applications would, not the enums. This only works ATM because the test framework uses the debug build. There are quite a lot of these, all with `goog.require`, and removing them would considerably simplify the dependency structure and change to modules.

I'll work on a PR for this.

probins:
Fixes #5418. Think I've got them all. :-)

### DIFF
--- a/test/spec/ol/control/zoomslidercontrol.test.js
+++ b/test/spec/ol/control/zoomslidercontrol.test.js
@@ -63,7 +63,7 @@ describe('ol.control.ZoomSlider', function() {
       control.setMap(map);
       control.initSlider_();
 
-      var horizontal = ol.control.ZoomSlider.direction.HORIZONTAL;
+      var horizontal = 1;
       expect(control.direction_).to.be(horizontal);
 
       control.dispose();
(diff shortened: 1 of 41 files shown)


probins:
review anyone?

ahocevar:
Looks good, thanks @probins.

ahocevar:
Oh, I had missed https://github.com/openlayers/ol3/issues/5463#issuecomment-226516869. And I have to admit I see the benefit of having enums. @probins, I don't see anything negative about having these enums in the tests, and they help us test the enums indirectly. That said, I think I'm going to revert this.

probins:
but surely the tests are there to test how apps would use the library. Apps don't use the enums

probins:
I'm not saying there shouldn't be enums, just that they don't belong in the tests

ahocevar:
> but surely the tests are there to test how apps would use the library. Apps don't use the enums

In theory (and of course in computer science school) yes, but many thing we test use internal/unexported variables/functions, so we'll have to run our tests with a debug build for the time being.

ahocevar:
Looking through this changeset again, I think the biggest loss is the link to `ol.TileState` and `ol.ImageState`. If we decide to use string values (e.g. `'loading'`) for those like we do everywhere else, there will be more effort to update the tests.

PR #5432, opened 2016-06-03
Remove use of goog.uri

### LINKED ISSUE #5430: WMS param STYLES is by default sent without =
This is a regression in 3.16.0. 

Problem can be seen in the example:
http://openlayers.org/en/v3.16.0/examples/wms-image.html?q=wms

WMS request like the following can be observed:
`http://demo.boundlessgeo.com/geoserver/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=topp%3Astates&CRS=EPSG%3A3857&STYLES&WIDTH=986&HEIGHT=532&BBOX=-15492472.25%2C2003371.5%2C-5847584.75%2C7205188.5`

Issue existed before and was fixed #242 (pre 3.0) but probably had another cause back then.

bjornharrtell:
Resolves #5430.

### DIFF
--- a/src/ol/source/imagewmssource.js
+++ b/src/ol/source/imagewmssource.js
@@ -3,7 +3,6 @@
 goog.provide('ol.source.ImageWMS');
 
 goog.require('goog.asserts');
-goog.require('goog.uri.utils');
 goog.require('ol');
 goog.require('ol.Image');
 goog.require('ol.events');
@@ -15,6 +14,7 @@ goog.require('ol.source.Image');
 goog.require('ol.source.wms');
 goog.require('ol.source.wms.ServerType');
 goog.require('ol.string');
+goog.require('ol.uri');
 
 
 /**
@@ -312,7 +312,7 @@ ol.source.ImageWMS.prototype.getRequestUrl_ = function(extent, size, pixelRatio,
   }
   params['BBOX'] = bbox.join(',');
 
-  return goog.uri.utils.appendParamsFromMap(this.url_, params);
+  return ol.uri.appendParams(this.url_, params);
 };
 
 
--- a/src/ol/uri.js
+++ b/src/ol/uri.js
@@ -1,0 +1,21 @@
+goog.provide('ol.uri');
+
+
+/**
+ * Appends query parameters to a URI.
+ *
+ * @param {string} uri The original URI, which may already have query data.
+ * @param {!Object} params An object where keys are URI-encoded parameter keys,
+ *     and the values are arbitrary types or arrays.
+ * @return {string} The new URI.
+ */
+ol.uri.appendParams = function(uri, params) {
+  var qs = Object.keys(params).map(function(k) {
+    return k + '=' + encodeURIComponent(params[k]);
+  }).join('&');
+  // remove any trailing ? or &
+  uri = uri.replace(/[?&]$/, '')
+  // append ? or & depending on whether uri has existing parameters
+  uri = uri.indexOf('?') === -1 ? uri + '?' : uri + '&'
+  return uri + qs;
+};
--- a/src/ol/source/imagearcgisrestsource.js
+++ b/src/ol/source/imagearcgisrestsource.js
@@ -1,7 +1,6 @@
 goog.provide('ol.source.ImageArcGISRest');
 
 goog.require('goog.asserts');
-goog.require('goog.uri.utils');
 goog.require('ol');
 goog.require('ol.Image');
 goog.require('ol.events');
@@ -10,6 +9,7 @@ goog.require('ol.extent');
 goog.require('ol.object');
 goog.require('ol.proj');
 goog.require('ol.source.Image');
+goog.require('ol.uri');
 
 
 /**
@@ -215,7 +215,7 @@ ol.source.ImageArcGISRest.prototype.getRequestUrl_ = function(extent, size, pixe
   if (modifiedUrl == url) {
     goog.asserts.fail('Unknown Rest Service', url);
   }
-  return goog.uri.utils.appendParamsFromMap(modifiedUrl, params);
+  return ol.uri.appendParams(modifiedUrl, params);
 };
 
 
--- a/src/ol/source/imagemapguidesource.js
+++ b/src/ol/source/imagemapguidesource.js
@@ -2,11 +2,11 @@ goog.provide('ol.source.ImageMapGuide');
 
 goog.require('ol.events');
 goog.require('ol.events.EventType');
-goog.require('goog.uri.utils');
 goog.require('ol.Image');
 goog.require('ol.extent');
 goog.require('ol.object');
 goog.require('ol.source.Image');
+goog.require('ol.uri');
 
 
 /**
@@ -222,7 +222,7 @@ ol.source.ImageMapGuide.prototype.getUrl = function(baseUrl, params, extent, siz
     'SETVIEWCENTERY': center[1]
   };
   ol.object.assign(baseParams, params);
-  return goog.uri.utils.appendParamsFromMap(baseUrl, baseParams);
+  return ol.uri.appendParams(baseUrl, baseParams);
 };
 
 
--- a/src/ol/source/tilearcgisrestsource.js
+++ b/src/ol/source/tilearcgisrestsource.js
@@ -1,7 +1,6 @@
 goog.provide('ol.source.TileArcGISRest');
 
 goog.require('goog.asserts');
-goog.require('goog.uri.utils');
 goog.require('ol');
 goog.require('ol.extent');
 goog.require('ol.object');
@@ -10,6 +9,7 @@ goog.require('ol.proj');
 goog.require('ol.size');
 goog.require('ol.source.TileImage');
 goog.require('ol.tilecoord');
+goog.require('ol.uri');
 
 
 /**
@@ -114,7 +114,7 @@ ol.source.TileArcGISRest.prototype.getRequestUrl_ = function(tileCoord, tileSize
   if (modifiedUrl == url) {
     goog.asserts.fail('Unknown Rest Service', url);
   }
-  return goog.uri.utils.appendParamsFromMap(modifiedUrl, params);
+  return ol.uri.appendParams(modifiedUrl, params);
 };
 
 
--- a/src/ol/source/tilewmssource.js
+++ b/src/ol/source/tilewmssource.js
@@ -5,7 +5,6 @@
 goog.provide('ol.source.TileWMS');
 
 goog.require('goog.asserts');
-goog.require('goog.uri.utils');
 goog.require('ol');
 goog.require('ol.extent');
 goog.require('ol.object');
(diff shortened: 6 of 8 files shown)


src/ol/source/imagewmssource.js
```diff
@@ -312,7 +312,7 @@ ol.source.ImageWMS.prototype.getRequestUrl_ = function(extent, size, pixelRatio,
   }
   params['BBOX'] = bbox.join(',');
 
-  return goog.uri.utils.appendParamsFromMap(this.url_, params);
+  return this.url_ + '?' + ol.ext.querystring.stringify(params);
```
bartvde: what if people pass in a & or ? already in the url?

bjornharrtell: Mmm yes, good point. Perhaps it would be better to incorporate goog.uri.utils.appendParamsFromMap into ol with the quirk for empty strings removed?

bjornharrtell: goog.uri.utils.appendParamsFromMap is pretty hairy... I will see if I can find any suitable replacement.

bjornharrtell: No luck. :( Tried out using a full URI parser (https://github.com/petkaantonov/urlparser) but on second thought it seems too heavy.

probins: can you not just do `(url.indexOf('?') == -1) ? '?' : '&'` or similar?

probins:
why introduce query-string here? Converting a params object to a `&key=value` string is surely not that complicated.
```
var qs = [];
for (var p in params) {
  if (params.hasOwnProperty(p)) {
    qs.push(p + "=" + encodeURIComponent(params[p]));
  }
}
return this.url_ + qs.join("&");
```
or similar

bjornharrtell:
Agreed, I'll try to make a replacement for `appendParamsFromMap` without external dependencies.

bjornharrtell:
Reimplemented based on suggestions from @probins and @bartvde, thanks!

src/ol/uri.js
```diff
@@ -0,0 +1,17 @@
+goog.provide('ol.uri');
+
+
+/**
+ * Appends query parameters from a map to an URI.
```
tschaub: I would avoid the use of the term "map" here.  In the case of our library, this is too easily confused with an instance of `ol.Map`.  In terms of the language, this is too easily confused with an instance of `Map`.  Instead, this could read:

> Appends query parameters to a URL.

src/ol/uri.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param {string} uri The original URI, which may already have query data.
```
tschaub: Do our uses of this always expect params to be appended rather than merged?

src/ol/uri.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param {!Object} map An object where keys are URI-encoded parameter keys,
+ *     and the values are arbitrary types or arrays.
+ * @return {string} The new URI.
+ */
+ol.uri.appendParamsFromMap = function(uri, map) {
```
tschaub: I'd be tempted to just call this `ol.uri.appendParams(uri, params)`.

src/ol/uri.js
```diff
@@ -0,0 +1,17 @@
...
+  var qs = Object.keys(map).map(function(k) {
+    return k + '=' + encodeURIComponent(map[k]);
```
tschaub: Don't we also want to `encodURIComponent(k)` here?

tschaub:
Thanks for the work on this @bjornharrtell.  It should be straightforward to write tests for this.

src/ol/uri.js
bjornharrtell: From what I can tell the original `goog.uri.utils.appendParamsFromMap` only does append no merge.

bjornharrtell:
Thanks @tschaub for additional input. Makes sense so I've made adjustments accordingly and have added two test cases.

src/ol/uri.js
```diff
@@ -0,0 +1,17 @@
+goog.provide('ol.uri');
+
+
+/**
+ * Appends query parameters to a URI.
+ *
+ * @param {string} uri The original URI, which may already have query data.
+ * @param {!Object} params An object where keys are URI-encoded parameter keys,
```
marcjansen: After the recent changes, this doc should be changed, right?

src/ol/uri.js
```diff
@@ -0,0 +1,17 @@
...
+ *     and the values are arbitrary types or arrays.
+ * @return {string} The new URI.
+ */
+ol.uri.appendParams = function(uri, params) {
+  var qs = Object.keys(params).map(function(k) {
+    return encodeURIComponent(k) + '=' + encodeURIComponent(params[k]);
+  });
+  return uri + (uri.indexOf('?') === -1 ? '?' : '&') + qs.join('&');
```
probins: I think this should be a bit more flexible to allow for people including '?' in the uri, as there may already be one or more parameters in uri. If uri ends with '?' or '&' just add the params without any prefix.

tschaub: I think you'll have to leave examples about what you think should be supported to make it clearer.  These can then be added to the tests below.

probins: for example, http://www.icc.cat/eng/Home-ICC/Digital-geoinformation/Online-services-Geoservices/Quick-WMS-WMTS-of-raster-cartography includes the '?' in the base url. http://www.ign.es/wms-inspire/mapa-raster?request=GetCapabilities&service=WMS has a parameter in the base url and finishes with '&'. I've found this sort of thing to be quite common with WMS. Of course, you could get around this by specifying in the docs that any trailing '?' or '&' should be left out, but ISTM better if the library can handle this. Having said that, it's probably also the case that browsers are clever enough to ignore duplicates like this.

probins: just tried some tests on these examples. In neither case does it seem to matter if there is a duplicate '??' or '&&'. However, the current `goog.uri` function removes any duplicates and if there are browsers/servers where this would create an error, it would be a regression if this new function doesn't do this. It should be pretty easy to remove the duplicate by not adding the '?' or '&' if `uri.slice(-1)` is '?' or '&' (or alternatively `uri = uri.slice(0,-1)`) or whatever.

marcjansen: In my experiemce the double questionmark in URLs is an error

ahocevar: To bring this to an end: @bjornharrtell, can you please change the line to
```js
return uri.replace(/\?&?/, '') + (uri.indexOf('?') === -1 ? '?' : '&') + qs.join('&');
```
Thanks.

probins: yup, that should also do it. I'd agree with @tschaub that it would be good to have tests for these cases

tschaub: Or just:
```js
uri.replace(/[?&]$/, '')
```

And test cases should be added for these:

```
http://example.com/foo
http://example.com/foo?
http://example.com/foo?bar=bam
http://example.com/foo?bar=bam&
http://example.com/foo?bar=bam&baz=bat
http://example.com/foo?bar=bam&baz=bat&bop=
// etc
```

bjornharrtell:
Thanks for the thorough review and regexp expertise. I hope to be able to rework this PR shortly, as in today or next few days at most.

probins:
it might also be a good idea to have a test for WMS that checks that `STYLES=` is output - which is what started this whole discussion :-) Could go in a separate PR though.

bjornharrtell:
I'm a bit a loss on the intent of the regexp. The following:

```js
var uri = 'http://example.com/foo?'
uri.replace(/[?&]$/, '') + (uri.indexOf('?') === -1 ? '?' : '&')
```

Results in `http://example.com/foo&` which doesn't seem right.

bjornharrtell:
Trying out `goog.uri.utils.appendParamsFromMap` it does not seem to handle duplicate ?? or &&. @probins you claimed otherwise?

I've tried this out in this fiddle: https://jsfiddle.net/s4sp7ne8/.

bjornharrtell:
Implemented handling URLs ending with ? or & using `uri.slice(-1)` as suggested by @probins and added additional test cases.

probins:
> Trying out goog.uri.utils.appendParamsFromMap it does not seem to handle duplicate ?? or &&. @probins you claimed otherwise?

what I meant was that if you define '?' or '&' in the uri string (which I'm currently doing to use the 2 example WMS servers I mentioned), it correctly does not append a second '?' or '&'.

probins:
> Results in http://example.com/foo& which doesn't seem right.

yeah, because the 2nd part isn't operating on the 1st part. :-) Something like this should work:
```
var uri = 'http://example.com/foo?';
uri = uri.replace(/[?&]$/, '');
uri = uri + (uri.indexOf('?') === -1 ? '?' : '&');
```

probins:
> var uri = 'http://example.com/foo?';
> uri = uri.replace(/[?&]$/, '');

or combine these 2 lines:
`var uri = 'http://example.com/foo?'.replace(/[?&]$/, '');`

sweco-sebhar:
Thanks @probins, now I get it. :)

Note that I reverted the URL encoding of keys, since `goog.uri.utils.appendParamsFromMap` does not.

bjornharrtell:
Thanks @probins, now I get it. :)

Note that I reverted the URL encoding of keys, since goog.uri.utils.appendParamsFromMap does not.

src/ol/uri.js
bjornharrtell: Changed back into leaving keys untouched as this is the existing behaviour with `goog.uri.utils.appendParamsFromMap`.

src/ol/uri.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {string} uri The original URI, which may already have query data.
+ * @param {!Object} params An object where keys are URI-encoded parameter keys,
+ *     and the values are arbitrary types or arrays.
+ * @return {string} The new URI.
+ */
+ol.uri.appendParams = function(uri, params) {
+  var qs = Object.keys(params).map(function(k) {
+    return k + '=' + encodeURIComponent(params[k]);
+  }).join('&');
+  // remove extraneous ? or &
```
probins: I think I'd change this to 'remove any trailing ? or &'

src/ol/uri.js
```diff
@@ -0,0 +1,21 @@
...
+  uri = uri.replace(/[?&]$/, '')
+  // append ? or & if uri has existing parameters
```
probins: I think this would be better replacing 'if' by 'depending on whether'

probins:
a couple of small comments on comments, but otherwise this LGTM. Thanks.

sweco-sebhar:
Amended comments and squashed. Thanks to all involved for the great review work.

bjornharrtell:
Amended comments and squashed. Thanks to all involved for the great review work.

ahocevar:
Thanks for this effort @bjornharrtell. I'm merging this now despite missing semicolons at the end of two lines, but I'll create a follow-up pull request with updated eslint rules so we can enforce those.

tschaub:
> Note that I reverted the URL encoding of keys, since goog.uri.utils.appendParamsFromMap does not.

I think we should encode keys and values.  Not sure what the justification is for not encoding keys (except that `goog` doesn't).

bjornharrtell:
I made the assumption that encoding keys could potentially cause regressions.

PR #5434, opened 2016-06-04
Fix externs in closure compiler tutorial

tsauerwein:
See also https://github.com/openlayers/openlayers-closure-application/pull/4


fredj:
thanks @tsauerwein

PR #5438, opened 2016-06-06
Don't make ol.format.ogc.filter.Filter inherit from ol.Object

fredj:
See https://github.com/openlayers/ol3/pull/5252#discussion_r65902356


tsauerwein:
Thanks, that looks good.

fredj:
Thanks for the review

PR #5444, opened 2016-06-07
Fix feature without geometry error in renderers

fblackburn:
This pull request make sure that we don't try to get a style object for a feature without geometry in the renderer. Trying to fetch a style via the style function on an empty geometry cause an error.


tschaub:
It would be nice to see a regression test here.  If there is currently an unhandled error when we try to render a feature without a geometry, it should be easy to add a test that demonstrates this.

It would also be good to make sure we aren't changing behavior for people who generate a geometry in a style - perhaps this is an edge case, but it would be possible now to render a feature without a geometry by dynamically creating a geometry in a style.

fblackburn:
To prevent regression, would it be better to make sure the default style option don't crash on feature without geometry? I'm not sure what the side effect would be farther down the rendering chain.

PR #5445, opened 2016-06-08
Select interaction default styling function can now handle geometry less features

fblackburn:
This fix make sure the feature passed to the default styling function of the select interaction don't crash when trying to return the style of a feature without geometry.

### DIFF
--- a/src/ol/interaction/select.js
+++ b/src/ol/interaction/select.js
@@ -301,6 +301,9 @@ ol.interaction.Select.getDefaultStyleFunction = function() {
       styles[ol.geom.GeometryType.LINE_STRING]);
 
   return function(feature, resolution) {
+    if (!feature.getGeometry()) {
+      return null;
+    }
     return styles[feature.getGeometry().getType()];
   };
 };


bartvde:
When would a feature without a geometry get selected? Are you adding it manually to the collection of the select interaction? Can't you protect it there?

fblackburn:
In our application, a map is shown next to a table containing a list of feature. In this list some have geometry and some don't. In order to keep the map and table in sync, we push selected element in table to the select interaction. We could filter the geometry less feature at this point, but since all our modification tool assume that the modified feature are all in the select interaction, it break a lot of thing if they're not pushed in the interaction.

I know it's a pretty specific use case, but since it's possible to add a geometry less feature to the select interaction I feel like it's a good thing to be able to handle it cleanly.

marcjansen:
I agree with the reasoning here. Do we have a CLA @fblackburn? This is good to be merged once we have one.

ahocevar:
CLA confirmed.

marcjansen:
Thanks @ahocevar for the confirmation and to @fblackburn for the fix.

marcjansen:
@fblackburn, this needs to be rebased, I fear. Can you do it? If not just tell us and we'll sort it out together.

fblackburn:
I'll work on it later today. Thanks!

fblackburn:
The branch has been rebased

marcjansen:
Cool thanks again!

PR #5446, opened 2016-06-08
Add a note in upgrade-notes.md about the goog.inherits removal

fredj:
See #5194 

And please don't hesitate to correct my English ...

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -6,6 +6,24 @@ Users compiling their code with the library and using types in the `ol.raster` n
 * `ol.raster.Operation` to `ol.RasterOperation`
 * `ol.raster.OperationType` to `ol.RasterOperationType`
 
+#### Removal of goog.inherits
+
+goog.inherits is not used anymore in the library. Users extending OpenLayers classes and compiling their code with the library must now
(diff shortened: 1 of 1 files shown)


changelog/upgrade-notes.md
```diff
@@ -1,5 +1,24 @@
 ## Upgrade notes
 
+#### Removal of goog.inherits
+
+goog.inherits is not used anymore in the library. Users extending OpenLayers classes and compiling their code with the library must now
+use ol.inherits instead. This also means that goog.base is no more available; the base class must now be called. If you previously had:
```
probins: /no more/no longer/

ahocevar:
Do we need this? Neither `goog.base` nor `goog.inherits` are part of the OpenLayers API. And if people compile their applications together with the library, they should still be able to use both if they also use the Closure library. Or am I missing something?

changelog/upgrade-notes.md
fredj: Thanks, fixed

fredj:
> [...] they should still be able to use both if they also use the Closure library

Yes, it needs to be checked

tschaub:
I agree this should not be needed as the `goog.inherits()` removal was an internal implementation detail only.

fredj:
ok, closing

PR #5449, opened 2016-06-09
Make example data/ paths absolute for source view and JSFiddle

### LINKED ISSUE #5447: LineString Arrows example code does not work
The supplied code of the LineString example does not work. Example can be found here: http://openlayers.org/en/latest/examples/line-arrows.html

The demo on that page works. But the code which you can copy below it does not.

ahocevar:
Fixes #5447.

### DIFF
--- a/tasks/build-examples.js
+++ b/tasks/build-examples.js
@@ -82,7 +82,9 @@ function augmentExamples(files, metalsmith, done) {
       if (!(jsFilename in files)) {
         throw new Error('No .js file found for ' + filename);
       }
-      var jsSource = files[jsFilename].contents.toString();
+      var jsSource = files[jsFilename].contents.toString()
+          // Change data paths to absolute urls
+          .replace(/'data\//g, '\'http://openlayers.org/en/v' + pkg.version + '/examples/data/');
       if (file.cloak) {
         for (var key in file.cloak) {
           jsSource = jsSource.replace(new RegExp(key, 'g'), file.cloak[key]);


tasks/build-examples.js
```diff
@@ -82,7 +82,9 @@ function augmentExamples(files, metalsmith, done) {
       if (!(jsFilename in files)) {
         throw new Error('No .js file found for ' + filename);
       }
-      var jsSource = files[jsFilename].contents.toString();
+      var jsSource = files[jsFilename].contents.toString()
+          // Change data paths to absolute urls
+          .replace(/'data\//g, '\'http://openlayers.org/en/v' + pkg.version + '/examples/data/');
```
tsauerwein: Should we maybe do the same for `/resources`?
https://github.com/openlayers/ol3/tree/master/examples/resources

marcjansen:
OK, @ahocevar I started with something like this:

```diff
diff --git a/examples/custom-interactions.html b/examples/custom-interactions.html
index d73a85a..b60f13b 100644
--- a/examples/custom-interactions.html
+++ b/examples/custom-interactions.html
@@ -6,5 +6,7 @@ docs: >
   This example demonstrates creating a custom interaction by subclassing `ol.interaction.Pointer`.
   Note that the built in interaction `ol.interaction.Translate` might be a better option for moving features.
 tags: "drag, feature, vector, editing, custom, interaction"
+cloak:
+  data/arrow.png: http://openlayers.org/en/v_VERSION_/examples/data/arrow.png
 ---
 <div id="map" class="map"></div>
diff --git a/examples/line-arrows.html b/examples/line-arrows.html
index 915eac0..8a2fa2f 100644
--- a/examples/line-arrows.html
+++ b/examples/line-arrows.html
@@ -5,5 +5,7 @@ shortdesc: Example of drawing arrows for each line string segment.
 docs: >
   Example of drawing arrows for each line string segment.
 tags: "draw, vector, arrow"
+cloak:
+  data/arrow.png: http://openlayers.org/en/v_VERSION_/examples/data/arrow.png
 ---
 <div id="map" class="map"></div>
diff --git a/tasks/build-examples.js b/tasks/build-examples.js
index 25e8eb7..0ff725d 100644
--- a/tasks/build-examples.js
+++ b/tasks/build-examples.js
@@ -34,6 +34,18 @@ function getRequires(src) {
 }
 
 /**
+ * Called with a cloaked value, this function will return its input but with
+ * placeholders (like `_VERSION_`) being replaced with their respective value.
+ *
+ * @param {string} cloakValue The cloak value in which placeholders will be
+ *   replaced.
+ * @return {string} The input with placeholders replaced.
+ */
+function replaceCloakVariables(cloakValue) {
+  return cloakValue.replace(/_VERSION_/g, pkg.version);
+}
+
+/**
  * Takes an array of the names of required OpenLayers symbols and returns an
  * HTML-snippet with an unordered list to the API-docs for the particular
  * classes.
@@ -85,7 +97,8 @@ function augmentExamples(files, metalsmith, done) {
       var jsSource = files[jsFilename].contents.toString();
       if (file.cloak) {
         for (var key in file.cloak) {
-          jsSource = jsSource.replace(new RegExp(key, 'g'), file.cloak[key]);
+          var cloaked = replaceCloakVariables(file.cloak[key]);
+          jsSource = jsSource.replace(new RegExp(key, 'g'), cloaked);
```

Which approach do you like more?

marcjansen:
I think I like yours better.

tasks/build-examples.js
ahocevar: That's done already @tsauerwein. See https://github.com/openlayers/ol3/blob/v3.16.0/tasks/build-examples.js#L115-L117.

ahocevar:
I have to humbly admit that I like mine better as well.

tasks/build-examples.js
marcjansen: I don't thibnk we need this. The advamnced mapbox vector tiles example e.g. uses `resources`, and in the generated HTML correct URIs are already referenced. Do I miss something?

marcjansen:
This looks good to me, @ahocevar. Please merge.

PR #5450, opened 2016-06-09
Remove sub-namespaces from all remaining typedefs

probins:
My final work on typedefs, which removes the need for all `goog.provide`s in `typedefs.js` except for the 2 strange problem cases. As before, I've kept the sub-namespace in the new name where I thought there was ambiguity, but otherwise removed it. This highlighted one unneeded require in `geom/circle` which I've removed. As there is no longer an api distinction for typedefs, I've listed all the renamed ones in the release notes file, even though I doubt whether most of them are used.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -6,6 +6,29 @@ Users compiling their code with the library and using types in the `ol.raster` n
 * `ol.raster.Operation` to `ol.RasterOperation`
 * `ol.raster.OperationType` to `ol.RasterOperationType`
 
+#### All typedefs now in ol namespace
+
+Users compiling their code with the library should note that the following typedefs have been renamed; your code may need changing if you use these:
+* ol.events.ConditionType to ol.EventsConditionType
+* ol.events.EventTargetLike to ol.EventTargetLike
+* ol.events.Key to ol.EventsKey
+* ol.events.ListenerFunctionType to ol.EventsListenerFunctionType
+* ol.interaction.DragBoxEndConditionType to ol.DragBoxEndConditionType
+* ol.interaction.DrawGeometryFunctionType to ol.DrawGeometryFunctionType
+* ol.interaction.SegmentDataType to ol.ModifySegmentDataType
+* ol.interaction.SelectFilterFunction to ol.SelectFilterFunction
+* ol.interaction.SnapResultType to ol.SnapResultType
+* ol.interaction.SnapSegmentDataType to ol.SnapSegmentDataType
+* ol.proj.ProjectionLike to ol.ProjectionLike
+* ol.style.AtlasBlock to ol.AtlasBlock
+* ol.style.AtlasInfo to ol.AtlasInfo
+* ol.style.AtlasManagerInfo to ol.AtlasManagerInfo
+* ol.style.CircleRenderOptions to ol.CircleRenderOptions
+* ol.style.ImageOptions to ol.StyleImageOptions
+* ol.style.GeometryFunction to ol.StyleGeometryFunction
+* ol.style.RegularShapeRenderOptions to ol.RegularShapeRenderOptions
+* ol.style.StyleFunction to ol.StyleFunction
+
 ### v3.16.0
 
 #### Rendering change for tile sources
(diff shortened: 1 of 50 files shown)


gberaudo:
Thanks @probins for documenting the mapping between the old symbols and the new ones. Ideally there should be a script to migrate code but this is minor and we can write one when we update OL3.

ahocevar:
@probins Can you remind me what the "strange problem cases" were?

probins:
see https://github.com/openlayers/ol3/issues/5299#issuecomment-216207123
These compile if there is a `goog.provide` in `typedefs.js` and a `goog.require` in `format/ogc/filter.js` and `events.js` respectively.

I have no idea why these errors are occurring, so if you can shed any light ...

ahocevar:
I found several problems related to this pull request. I'll dig deeper tomorrow and report a summary of my findings here.

ahocevar:
Sorry, this is taking longer than expected. I need to experiment and research a bit more.

ahocevar:
@probins, feel free to cherry-pick 920a38ea27d087660bfa1a3a513e610b23d327c4.

The issues I encountered were the following:
* Without a `goog.provide` or `goog.require`, the `src/ol/typedefs.js` file would no longer be added to the list of files that get sent to the compiler. So it is now added manually in `tasks/build.js`.
* With `--manage_closure_dependencies` set to true, the `src/ol/typedefs.js` file would be excluded from compilation even if sent to the compiler. So we configure the compiler now with `--dependency_mode=LOOSE` instead.
* `ol.Extent` did not work as expected in `src/ol/format/ogc/filter.js` because of google/closure-compiler#1052.
* `ol.EventTargetLike` was simplified to work correctly. Looks like the compiler had a problem with the original complex typedef.
* The recursive type definition of `ol.LRUCacheEntry` only worked by chance. See google/closure-compiler#533. I simplified the type definition and had to add a few type casts instead.

probins:
> With --manage_closure_dependencies set to true, the src/ol/typedefs.js file would be excluded from compilation even if sent to the compiler. So we configure the compiler now with --dependency_mode=LOOSE instead

don't think that's necessary; the trick is to add typedefs.js to the list of js files, not to the src files. I'll check that out.

Why have you removed `ol.extent` from all those files? That is surely still needed.

ahocevar:
@probins:

> don't think that's necessary; the trick is to add typedefs.js to the list of js files, not to the src files. I'll check that out.

Would be nice if you could figure that out. Thanks.

> Why have you removed ol.extent from all those files? That is surely still needed.

Oops, search/replace error. New commit: ef7f3487d68a6cf0bc4e62a49519f77670bc1092

probins:
ok, thanks for sorting out the 2 cases. I'll get cherry-picking ...

probins:
ok, @ahocevar, now included your commit, adapted to continue using 'manage_closure_dependencies'.

Hey, does this mean we're now finished with typedefs? 😄

ahocevar:
Thanks @probins! You mentioned that the trick is to add typedefs.js to the list of js files, not to the src files. What would be required to do that, instead of manually adding `src/ol/typedefs.js` to the list of src files?

probins:
typedefs.js is now passed directly to the compiler as a compile option. The problem with adding it to the src files, as you were doing, is that it then runs through closure-util's `getDependencies()`, which ignores files which have no `goog.provide/require`, so it never gets to the compiler.

probins:
though, having said that, I suppose you could add it as a compile option in `config/ol.json`, though it would then also have to be added to custom build config files too.

ahocevar:
I was just thinking the same. Not sure which option I like better.

probins:
I don't normally like hard-coding filenames into scripts, but in this case I think I prefer having it this way, so people don't have to add it to their config files.

ahocevar:
Ok, agreed. Thanks for your ongoing effort on this. Merging now 😄 .

probins:
just remembered that we're not quite finished with typedefs: `ol.vec.Mat4.Number` is still outstanding https://github.com/openlayers/ol3/pull/5138#issuecomment-218984068

PR #5453, opened 2016-06-09
Reference ol.MapBrowserEvent instead of ol.MapBrowserPointerEvent

### LINKED ISSUE #5414: MapBrowserPointerEvent not in api docs
For example, http://openlayers.org/en/master/apidoc/ol.interaction.ModifyEvent.html points to http://openlayers.org/en/master/apidoc/ol.MapBrowserPointerEvent.html, but this page does not exist.

Am I right in thinking that `oli.MapBrowserPointerEvent` needs to be added to `externs/oli.js` and that `ol.MapBrowserPointerEvent` needs to implement it?

ahocevar:
There is nothing exportable in ol.MapBrowserEvent in addition to what is exportable in ol.MapBrowserEvent.

Fixes #5414.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+#### `ol.interaction.ModifyEvent` changes
+
+The event object previously had a `mapBrowserPointerEvent` property, which has been renamed to `mapBrowserEvent`.
+
 #### Removal of ol.raster namespace
 
 Users compiling their code with the library and using types in the `ol.raster` namespace should note that this has now been removed. `ol.raster.Pixel` has been deleted, and the other types have been renamed as follows, and your code may need changing if you use these:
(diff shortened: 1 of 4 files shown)


changelog/upgrade-notes.md
```diff
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+#### `ol.interaction.ModifyEvent` changes
+
+The event object previously had a `mapBrowserPointerEvent` property, which has bewen renamed to `mapBrowserEvent`.
```
probins: `map` should read `Map`. 'been renamed'.

probins:
MapBrowserPointerEvent is also referred to in http://openlayers.org/en/master/apidoc/ol.events.condition.html (mouseOnly). That seems to be the only reference in the apidocs apart from interaction.Modify

changelog/upgrade-notes.md
ahocevar: The former isn't true, and I fixed the latter.

ahocevar:
I removed that too now. No more `ol.MapBrowserPointerEvent` in the API.

fredj:
LGTM

PR #3476, opened 2015-04-02
Webgl polygons and lines (experiment)

gberaudo:
Basic support for rendering polygons with WEBGL.
Status:
[x] PolygonGeometry
[x] MultPolygonGeometry
[x] Fill with color
[x] Stroke with line width and color
[ ] picking / hit detection
[ ] feature skipping
[ ] Render to the eye?

PR #5462, opened 2016-06-13
WebGL vector support

GaborFarkas:
This is an effort to write a proper WebGL vector renderer for OL3. There are a lot of things to do, it will take some time to reach a stable status. The basis of this work is #3476.

TODO list:
- [x] LineString triangulation with RTE (Relative To Eye)
- [x] Handle line joins
- [x] Handle line caps
- [x] Properly handle edge cases
- [x] Polygon triangulation with RTE
- [x] Implement hit detection
- [x] Remove `ol.ext.earcut`
- [x] Add circle renderer
- [x] Clean up and make the code compilable
- [x] Add tests

There is much more to do in order to make a robust WebGL rendering engine. Supporting line dashes, and adding text support are subjects for another work.


src/ol/render/webgl/webglreplay.js
```diff
@@ -114,6 +128,23 @@ ol.render.webgl.ImageReplay = function(tolerance, maxExtent) {
...
+  SQUARE_CAP: 2,
+  BEVEL_FIRST: 3,
+  BEVEL_SECOND: 4,
+  MITER_BOTTOM: 5,
+  MITER_TOP: 6,
+  ROUND_JOIN: 7,
+  ROUND_CAP: 8,
+  ROUND_BOTH: 9
+};
+
```
GaborFarkas: I think this might go to webgl.js.

src/ol/render/webgl/webglreplay.js
```diff
@@ -1011,27 +1028,180 @@ goog.inherits(ol.render.webgl.LineStringReplay, ol.render.webgl.Replay);
...
+ * @private
+ */
+ol.render.webgl.LineStringReplay.prototype.isClosed_ = function(flatCoordinates, offset, end, stride) {
+  var lastCoord = end - stride;
+  if (flatCoordinates[offset] === flatCoordinates[lastCoord] &&
+      flatCoordinates[offset + 1] === flatCoordinates[lastCoord + 1]) {
+    return true;
   }
+  return false;
 };
```
GaborFarkas: This is a helper function used only by the LineString WebGL renderer. On the other hand, it is a general function working on flat geometries. Should it go to the ol.geom.flat namespace?

src/ol/render/webgl/webglreplay.js
```diff
@@ -1240,6 +1243,38 @@ ol.render.webgl.LineStringReplay.prototype.isClosed_ = function(flatCoordinates,
...
+ * Check if the linestring can be drawn (i. e. valid).
+ * @param {Array.<number>} flatCoordinates Flat coordinates.
+ * @param {number} offset Offset.
+ * @param {number} end End.
+ * @param {number} stride Stride.
+ * @param {boolean} closed The linestring is a boundary.
+ * @return {boolean} The linestring can be drawn.
+ * @private
+ */
+ol.render.webgl.LineStringReplay.prototype.isValid_ = function(flatCoordinates, offset, end, stride,
```
GaborFarkas: Same thoughts as on `isClosed_`. The function is operating on flat LineStrings, and thus it might go in the `ol.geom.flat` namespace.

bjornharrtell:
Nice to see more work on a WebGL renderer! I'm curious, what is Relative To Eye that seemingly is to replace earcut?

GaborFarkas:
Thank you! RTE is a method used by Cesium for eliminating [jitter](http://blogs.agi.com/insight3d/index.php/2008/09/03/precisions-precisions/) caused by rounding long numbers in the GPU. We handle every coordinate related to the eye of the camera, and transform them with an offset matrix to clipping space. Luckily, we only have to deal with 2D coordinates, thus the eye of the camera means the center of the scene (in map units). Furthermore, as the offset matrix is already implemented in the image renderer, we just have to use it in the other classes.

I'm planning to replace earcut with an algorithm which is more precise, but not much slower. I'll dig into the related literature to find a suitable one.

gberaudo:
Great @GaborFarkas.

Cesium [is using earcut](https://github.com/AnalyticalGraphicsInc/cesium/commit/c1c4efce5a32bcf8fa8382f5cd9af2e1d23c7a8c). Have you tried an up-to-date version?

Anyway, I am curious about what you can find.

GaborFarkas:
Wow, cool, I didn't know about that @gberaudo.

My main issue with earcut is noted on its [GitHub page](https://github.com/mapbox/earcut#why-another-triangulation-library).

> If you want to get correct triangulation even on very bad data with lots of self-intersections and earcut is not precise enough, take a look at libtess.js.

According to @mourner's test, libtess.js is much slower, than earcut. I will try to find a good compromise, although if I can't find one (or can't implement it), than I guess, we have to choose between speed and precision.

GaborFarkas:
I just fixed a problem which I couldn't resolve for some time. This might be a bug in WebGL, or a feature, but we can't link the first indices of a closed linestring to the last ones directly to close the line. I assume, the used vertices are popped from the stack after some time, and when it comes to the final linking, the old indices point to vertices which are not there anymore.

There is one more degeneration to handle (sharp angles), and the linestring renderer can be considered stable.

marcjansen:
Very glad to see such nice progress here! 👍👏👏

bjornharrtell:
I would say it's not too much to ask that the geometry source is somewhat clean so that it can work with earcut. To me it would even make sense to restrict the webgl renderer to pretessellated data, optimally a pretessellated variant of mapbox vector tile, or at least making on the fly tessellation optional.

GaborFarkas:
Than you @marcjansen! @bjornharrtell, I agree, that we could ask for sanitized geometries, however I don't think we should. One of the biggest strengths of OL3 is, it's a general purpose library. It can be used for a wide variety of applications from small scale web mapping to creating a basic GIS on the web. If we cannot render degenerate geometries, we blow away that use case, in which OL3 has quite a monopole status in the FOSS world. I'm eager to learn your, and other developers' opinion on this perspective.

gberaudo:
The renderer produces simplified polygons we will have to tesselate anyway.

bjornharrtell:
I agree that the ability to handle degenerate geometries (within some reason?) is a strength of OL3 and do not mean to suggest it should dropped as a goal. At the same time, performance is probably why you would be interested in WebGL in the first place and you can always fall back on the Canvas renderer.

I don't understand your statement @gberaudo, could you clarify?

GaborFarkas:
@bjornharrtell The geometry simplification is hardcoded into the rendering pipeline.

There are some very unusual cases, when the linestring renderer can fail (create artifacts). However, creating those situations are more like bugging out the renderer, than having bad geometries. I'm not planning to patch it up, however it should be noted, it could be further improved.

GaborFarkas:
As a second thought, neither cap rounding, nor eliminating overdraw works on Firefox for me. I'm working on a Debian Jessie, Firefox ESR 45.2.0 is producing this unusual behaviour.

gberaudo:
@bjornharrtell, for performance reasons, what is rendered by OL3 is not the plain geometry but simplifications of it according to zoom level / bbox (using the Ramer-Douglas-Peucker. algorithm).

bjornharrtell:
@gberaudo do you mean this https://boundlessgeo.com/2014/03/openlayers-vector-rendering?

For preoptimized data sources like mapbox vector tiles I think it might make sense to make that optional.

gberaudo:
@bjornharrtell,see https://github.com/openlayers/ol3/blob/cc9d4d87948e1eed283f1bbb0b7e942cb360f424/src/ol/render/vector.js#L111-L122.

ahocevar:
@bjornharrtell:

> For preoptimized data sources like mapbox vector tiles I think it might make sense to make that optional.

That is bypassed already for sources that provide `ol.RenderFeature` instead of `ol.Feature` features. Currently only `ol.source.VectorTile` does that.

mourner:
> I'm planning to replace earcut with an algorithm which is more precise, but not much slower. I'll dig into the related literature to find a suitable one.

Looking forward to that! It's a very difficult challenge.

GaborFarkas:
Difficult indeed, that's why I'm so excited about it. I might fail, there is quite a good chance; but not before I did my best to solve it.

GaborFarkas:
I'm kinda lost in understanding this whole hit detection concept. Could somebody explain what is it, and why do we need it?

gberaudo:
@GaborFarkas, when you click on a pixel of the map, you would like to know which features are underneath.

GaborFarkas:
So as far as I get it, there is one main canvas for the map, and one internal canvas for every layer. I don't have to do any special, just render every feature on the internal canvas again which is provided by the renderer to the `drawHitDetectionReplay` function.

gberaudo:
Yes, just check how it is done with the ImageReplay. There are several hitdetection methods there.

tsauerwein:
For WebGL the hit-detection uses a [1x1px frame-buffer](https://github.com/openlayers/ol3/blob/2117f2e/src/ol/webgl/context.js#L285-L310). For [forEachFeatureAtPixel](http://openlayers.org/en/master/apidoc/ol.Map.html#forEachFeatureAtPixel) the features are [drawn one-by-one](https://github.com/openlayers/ol3/blob/2117f2e3b117da88af5b255c860e80b7f99c5ba1/src/ol/render/webgl/webglreplay.js#L782) into this framebuffer. For [hasFeatureAtPixel](http://openlayers.org/en/master/apidoc/ol.Map.html#hasFeatureAtPixel) all features are [drawn at "once"](https://github.com/openlayers/ol3/blob/2117f2e3b117da88af5b255c860e80b7f99c5ba1/src/ol/render/webgl/webglreplay.js#L756) (by texture group).

Then you [check](https://github.com/openlayers/ol3/blob/2117f2e3b117da88af5b255c860e80b7f99c5ba1/src/ol/render/webgl/webglreplay.js#L1110-L1114) if there is a color in the framebuffer, which means that there is a feature at the position.

GaborFarkas:
Thank you, @gberaudo and @tsauerwein, you saved me quite a lot of time!

src/ol/render/webgl/webglreplay.js
```diff
@@ -1433,46 +1428,21 @@ ol.render.webgl.LineStringReplay.prototype.drawReplay_ = function(gl, context, s
...
-  if (typeof tmpDepthFunc === 'number') {
-    gl.depthFunc(tmpDepthFunc);
+  if (!hitDetection) {
+    gl.clear(gl.DEPTH_BUFFER_BIT);
+    gl.disable(gl.DEPTH_TEST);
+    //Restore GL parameters.
+    //NOTE: Type checks will be obsolete when Closure Compiler recognizes the return type of specific
+    //gl.getParameter() calls.
+    if (typeof tmpDepthMask === 'boolean') {
+      gl.depthMask(tmpDepthMask);
```
tsauerwein: You can use `var tmpDepthMask = /** @type {boolean} */ (gl.getParameter(gl.DEPTH_FUNC));` to tell the compiler the type.

GaborFarkas: Thank you! I tried that in the first place, but without the extra parentheses.

GaborFarkas:
Were there some changes in the build system I'm not aware of? Since the last rebase, if I build the debug version, I don't have a `layer` namespace, while if I build the compiled version, I don't even have the `ol` object. BTW both running `tasks/build.js` and `npm build` produce the same symptoms.

tsauerwein:
Can you try a `make clean` and then build again?

GaborFarkas:
Yes, and even dumped the `node_modules` and `build` folders to have a clean install. No change.

GaborFarkas:
Okay, this might be a regression introduced with [this](https://github.com/openlayers/ol3/commit/6b465902cdac75eab642753aa0dce92a56c6ca69#diff-1dc700f9e7dea94aca6a9190ab6ed6b3R245) line. My theory is if we put OL3 in the `head` element, the `body` element isn't created when `CHANGETYPE` is enumerated.

GaborFarkas:
How about increasing the line width by 1 pixel for hit detection?

GaborFarkas:
> node tasks/test-rendering.js
Gecko: 38.0.5
Script Error: [JavaScript Error: "Error: expected '14.52' to be below 3" {file: "http://127.0.0.1:3001/node_modules/expect.js/index.js" line: 96}]
       Stack:
         -> http://127.0.0.1:3001/node_modules/expect.js/index.js: 96
ol.rendering.style.LineString different strokes tests the WebGL renderer
Error: Error: expected '14.52' to be below 3 (http://127.0.0.1:3001/node_modules/expect.js/index.js:96)
1 test(s) failed.
make: *** [test-rendering] Error 1

I don't really understand what's happening, both `make test-rendering` and `make ci` run on my machine without a problem. Would setting the threshold to 14.52 or 14.6 be problematic?

marcjansen:
Maybe @tsauerwein can help?

GaborFarkas:
I think we should create a rendering test similar to the Turing test for AIs or the moose test for vehicles. It should be called the Natural Earth test. If a rendering engine can flawlessly render the small scale vector datasets, it is a damn solid one.

No offence though, I love open data, it's just another use case.

GaborFarkas:
FYI @mourner, earcut almost passed the NE test. It cannot mangle South Africa though. A country with a hole in it, and another one on its border.

sa

GaborFarkas:
Sorry, my bad, earcut **passed** the NE test.

bjornharrtell:
Great to see your progress @GaborFarkas. Already works in some limited tests for me.

Perhaps two things to add to the TODO list:
* Text support
* VectorTileLayer renderer implementation

GaborFarkas:
Thank you, @bjornharrtell. Those features are definitely needed for a viable WebGL renderer, but they will be subjects for other works with their respective PRs. The goal of this PR is creating a solid line and polygon renderer.

GaborFarkas:
There will be less visible progress on this work in the near future. There are two main reasons:
- I'm currently working on the aforementioned ear clipping algorithm, which needs to reach a more or less stable status before pushing.
- One of my articles got declined, as it needs serious amount of work before it can be published according to reviews, and that is my number one priority right now.

mourner:
> I'm planning to replace earcut with an algorithm which is more precise, but not much slower. I'll dig into the related literature to find a suitable one.

Have you found a suitable algorithm? I'm slightly confused that you're replacing earcut with another ear-clipping implementation, which won't address your initial earcut concern of handling self-intersecting input.

GaborFarkas:
No, I did not found any other suitable approach, thus I stack with ear clipping. The main difference between this algorithm and earcut or FIST is that it stores _segments_ in the linked list, and uses RBush to index them in an R-Tree. While the polygon is simple, it follows the same rules as FIST CE2. If it detects a self-intersecting or self-overdrawing poly, however, it checks for intersections at every convex point. It tries to proceed with the simple ruleset extended with the intersection check (CE1), until it runs out of ears. If the intersections cannot be resolved locally, that will invoke the original ear clipping methodology (partition the poly into a set of simple ones, etc.), and it will be the next thing to write.

mourner:
@GaborFarkas nice! Have you done any benchmarks that compare the two approaches?

GaborFarkas:
Thank you, @mourner! I'm planning to run some tests tomorrow.

GaborFarkas:
Measured on the NE dataset, earcut is 40x faster on average. There are some huge outliers, which could be good candidates for optimization though.

GaborFarkas:
Some raw, statistically not representative numbers on the NE dataset (286 polygons):

- Processing time: SUM: 146.13 ms, AVG: 0.510 ms.
- Bridging time: SUM: 0.745 ms, AVG: 0.745 ms (there is only one hole in South Africa)
- Triangulating time: SUM: 374 ms, AVG: 1.308 ms

As these numbers do not add up to my previous statement, I will prepare a new, more precise benchmark soon.

GaborFarkas:
The results of the benchmark: our approach is not significantly slower than earcut. The measured dataset was NE again, benchmarked with benchmark.js. The flat coordinates were prepared, then our `drawCoordinates_` function was measured against this one:

``` javascript
ol.render.webgl.PolygonReplay.prototype.drawEarcut_ = function(
    flatCoordinates, holeFlatCoordinates, stride) {
  var coordArr = flatCoordinates;
  var holeArr = holeFlatCoordinates.length ? [] : null;
  var i, ii;
  if (holeFlatCoordinates.length) {
    for (i = 0, ii = holeFlatCoordinates.length; i < ii; ++i) {
      holeArr.push(coordArr.length / stride);
      ol.array.extend(coordArr, holeFlatCoordinates[i]);
    }
  }
  var indices = ol.ext.earcut(coordArr, holeArr, stride);
  ol.array.extend(this.vertices_, coordArr);
  ol.array.extend(this.indices_, indices);
};
```

This way, not the plain Earcut lib's speed, but its implementation into OL3 was measured against the native ear clipping algorithm, which was tailored to OL3.

Results after 5 consecutive measurements:

- Earcut: 236 - 572 opts/sec
- OL3: 186 - 401 opts/sec

GaborFarkas:
I was wondering if it was OK if I tamper with the main rendering engine, and extend it with a dedicated Circle renderer. It would point back to PolygonReplay in the canvas renderer, but for the WebGL engine, there would be a different class for it with distance checks instead of triangulating. I think it would be faster to run (as it is done in the GPU) and more convenient to code. What do you think about this change?

gberaudo:
@GaborFarkas having a dedicated WebGL renderer for circles makes sense to me.
Is anyone seeing an issue with the change in the canvas renderer?

tschaub:
I agree it makes sense to have WebGL specific circle rendering.  I don't immediately see anything that would need to change with the canvas renderer.  I would think it would be up to `ol.renderer.webgl.ReplayGroup` to provide a `ol.render.webgl.PolygonReplay` with a `drawCircle` method.

Not strictly related, but see also #5775.  It would be nice to move the AtlasManager functionality to the WebGL renderer.

GaborFarkas:
The problem is with the WebGL replays. One replay uses one specific GLSL program, and while triangulating circles is possible, it would be faster and more convenient to use a distance check based program for circles. That requires an `ol.render.webgl.CircleReplay`. However, if I write a circle replay, that would require a circle renderer, which needs to point at `ol.render.canvas.PolygonReplay` in the canvas engine.

Thank you for showing the issue with AtlasManager, I will keep that in mind. TBH, I was thinking about  redirecting points with circle icons to the circle renderer, as the current texture based point renderer produces blurry points. Points with other icons could still use the current point renderer.

GaborFarkas:
On the other hand, I cannot see a way we could do this redirection without creating two separate point categories in the main renderer.

GaborFarkas:
I've just seen the structural changes in the renderers' source files. Thank you @tschaub, it's much more clear now. I've got only one thing to add/consider: wouldn't it be nicer if the replays in the webgl renderer are named after their respective contents instead of `index.js`? There will be four replays at least, and I think that seeing four `index.js` tabs in the editor during development isn't too convenient.

GaborFarkas:
There's something weird about the select interaction. If I select a feature, the underlying feature still gets rendered, while some other seemingly random one disappears. This issue wasn't present until the rebase. Were there any changes in the interactions/renderers which could cause this phenomenon?

ahocevar:
> There's something weird about the select interaction.

The select interaction is weird anyway. I don't know of any renderer related changes, but the interaction itself had a few changes.

tschaub:
> wouldn't it be nicer if the replays in the webgl renderer are named after their respective contents instead of index.js

I'm open to this.  To be consistent, this would mean:

 * `src/ol.js` instead of `src/ol/index.js`
 * `src/ol/style.js` instead of `src/ol/style/index.js`
 * etc.

EladIsraeli:
Can i ask something that maybe doe not right to ask here, but this is just expertimental ? Or i can really use it on my project ? 

Because i want to use the webgl to draws points on my maps but i cant do it in the latest ol3 version.

tsauerwein:
> Because i want to use the webgl to draws points on my maps but i cant do it in the latest ol3 version.

Drawing points in WebGL already works with the current version, see: https://jsfiddle.net/uyckp2Lh/

EladIsraeli:
@tsauerwein does it have the same performance as here in this example : 

http://dev.camptocamp.com/files/ol3/webgl/examples/webgl-points.html

tsauerwein:
@EladIsraeli I answered in https://github.com/openlayers/ol3/issues/5942, because the discussion is not related to this PR.

GaborFarkas:
As you can see, introducing a circle renderer altered the rendering order in the canvas engine making some rendering tests fail. The quick fix (which I applied) changes some expected images and threshold values, however I do not consider it elegant. Refactoring the replay types would be also messy. What do you think we should do about this issue?

GaborFarkas:
My proposal: as we have a very limited set of replays, if a replay group cannot find a specific replay type, it should iterate through the existing ones (per Z index), and if it finds a matching one (with instanceof checks), it simply points that replay to the specified type. This would keep the current behaviour of the canvas renderer while it would also allow the webgl renderer to use a dedicated circle renderer. Furthermore, with this approach, there would be plenty of space to separate the point type to circles and other icons, and use the webgl circle renderer to render circular points.

GaborFarkas:
Apart from the aforementioned rendering order issue, I consider my work done and ready for review.

marcjansen:
So happy to read this!!! 👏👏

I will look into it but am by no means capable of giving it the review this work deserves.

bjornharrtell:
Impressive work @GaborFarkas!

GaborFarkas:
Thank you for the compliments. I hope the code reflects my pleasure in writing it.

marcjansen:
I have just had a quick look at this by changing some vector examples map renderer to use `'webgl'`. Not too many worked out of the box, as colors weren't as expected in the examples (assertions failed with code number 14), unsupported styles (`Assertion failed: Text constructor missing from ol.render.webgl.ReplayGroup.BATCH_CONSTRUCTORS`) or some cross domain issues (`vector-esri.html`).

Others worked, even though the perceived performance wasn't so great (`wkt.html`); this might also be because my computer isn't the fastest when it comes to WebGL. 

I repeat what I have stated earlier: My impression is this is awesome code which deserves a detailed review, which I sadly cannot provide. Anyone else from @openlayers/ol3?

marcjansen:
> The coloring problem bothers me a little bit though, did you experience that issue with examples using canvas patterns and gradients?

No for me e.g. `vector-osm.html` would give that error:

image

> …there isn't a text renderer for WebGL

I knew that and was expecting the error, to be clear.

GaborFarkas:
I can see. In that example, the color codes are provided in a canvas syntax instead of a color array. I will harden the color check ASAP. Thank you for your observation.

gberaudo:
@GaborFarkas, I am impressed by your work on this PR. I will find time this week for a review.

GaborFarkas:
Thank you @gberaudo. One sidenote to the triangulation: don't expect a huge performance impact. We will need some sort of caching mechanism for storing (per resolution?) triangulation information. That would give a great boost, until then only the first time rendering performance is improved (I guess the `CanvasRenderingContext2D` uses some sort of internal caching).

gberaudo:
Hi @GaborFarkas, I switched the default renderer to WebGL in the examples and opened some related to vector layers.

Here is how I make the examples which does not force a specific renderer use the WebGL one:
```diff
diff --git a/src/ol/map.js b/src/ol/map.js
index 78c989b..a4f9dbf 100644
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -77,7 +77,6 @@ ol.OL3_LOGO_URL = 'data:image/png;base64,' +
  * @const
  */
 ol.DEFAULT_RENDERER_TYPES = [
-  ol.renderer.Type.CANVAS,
   ol.renderer.Type.WEBGL
 ];
 
@@ -1448,7 +1447,7 @@ ol.Map.createOptionsInternal = function(options) {
   /**
    * @type {function(new: ol.renderer.Map, Element, ol.Map)}
    */
-  var rendererConstructor = ol.renderer.Map;
+  var rendererConstructor = ol.renderer.webgl.Map;
 
   /**
    * @type {Array.<ol.renderer.Type>}
```


Below are the examples I ran and the results I got, with chromium and Firefox on Linux. The examples were run in `raw` mode (unminified), which can be selected by passing the `mode=raw` parameter in the example URL. The examples are compiled in to the hosted directory with `make host-examples`.

1. http://localhost:3000/build/hosted/webgl_vector/examples/geojson.html?q=vector&mode=raw

a. In chromium (WebGL crash displaying the famous "rats! WebGL hit a snag")
```
replay.js:339  WebGL: INVALID_ENUM: stencilFunc: invalid function 
geojson.html:1 GL_INVALID_OPERATION : glGenSyncTokenCHROMIUM: fence sync must be flushed before generating sync token
geojson.html:1 WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost
```

(something in WebGL broke, I had to restart chromium to begin again in a clean state).

b. In Firefox (no crash but nothing is drawn)

Here are the logs
```
Error: WebGL: texImage2D: Chosen format/type incured an expensive reformat: 0x1908/0x1401context.js:382:3
Error: WebGL: uniform1i: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:169:5
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:172:3
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:174:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:176:3
Error: WebGL: drawElements: Invalid `type`: 0x1405replay.js:364:3
attempted to delete uncached buffercontext.js:141:15

Error: WebGL: uniform1i: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:169:5
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:172:3
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:174:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argumentlayer.js:176:3
Error: WebGL: drawElements: Invalid `type`: 0x1405replay.js:364:3
Error: WebGL: No further warnings will be reported for this WebGL context. (already reported 32 warnings)replay.js:364:3
GET 
https://c.tile.openstreetmap.org/0/0/0.png [HTTP/2.0 200 OK 188 ms]
TypeError: bufferCacheEntry is undefined
[En savoir plus]context.js:145:5
attempted to delete uncached buffercontext.js:141:15
```

2. http://localhost:3000/build/hosted/webgl_vector/examples/draw-features.html?q=vector&mode=raw
Then I tried the draw feature example. This example allows to select a feature type and draw it on the map. It is convenient since it allows drawing one type at a time.

First I tried drawing a point: it was OK.
Then I tried a line: I got error in the console and WebGL crashed (chromium):
```
draw-features.html:1 GL_INVALID_OPERATION : glGenSyncTokenCHROMIUM: fence sync must be flushed before generating sync token
draw-features.html:1 WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost
```

@marcjansen and @GaborFarkas, could you please try the examples above? Do they work as expected?

The crashes may be specific to my machine. A look at the output of dmesg shows issues in the gfx stack:
```dmesg
[ 1963.747460] [drm] stuck on render ring
[ 1963.748383] [drm] GPU HANG: ecode 7:0:0x85dffff8, in chromium-browse [4363], reason: Ring hung, action: reset
[ 1963.748385] [drm] GPU hangs can indicate a bug anywhere in the entire gfx stack, including userspace.
[ 1963.748386] [drm] Please file a _new_ bug report on bugs.freedesktop.org against DRI -> DRM/Intel
[ 1963.748387] [drm] drm/i915 developers can then reassign to the right component if it's not a kernel issue.
[ 1963.748387] [drm] The gpu crash dump is required to analyze gpu hangs, so please always attach it.
[ 1963.748388] [drm] GPU crash dump saved to /sys/class/drm/card0/error
```

Regarding the topics already evoked by @marcjansen:
- text renderer: IMO text styles should not stop rendering anymore (what about addind a no-op text render for now?);
- handling of colors: what is underlying issue with `ol.color.fromStringInternal_`? Is it a developer issue to call it with floating point value? Is there an error message? What can we do to fix that?

GaborFarkas:
Applying the prescribed modifications, compiling the examples with `make host-examples` and running the server with `make serve` still uses the canvas renderer for me. Additionally, appending `mode=raw` to the URL breaks the examples.

In `ol.color.fromStringInternal_` the RegExs should be altered in a way they can process floating point values (at least for alpha).

marcjansen:
I tend to agree with the solution to fix the regex of ol.color.

gberaudo:
With my modifications the renderer returned by map.getRenderer() is the WebGL renderer.
Nevertheless, I tried reverting my change and forcing the WebGL renderer in the example but I get the same WebGL crash. I rebooted the computer, tried adjusting chromium WebGL flags (about:flags) but still get no improvements yet.

```diff
diff --git a/examples/draw-features.js b/examples/draw-features.js
index a7c77ae..8dd91a6 100644
--- a/examples/draw-features.js
+++ b/examples/draw-features.js
@@ -18,6 +18,7 @@ var vector = new ol.layer.Vector({
 
 var map = new ol.Map({
   layers: [raster, vector],
+  renderer: 'webgl',
   target: 'map',
   view: new ol.View({
     center: [-11000000, 4600000],
```

GaborFarkas:
Thank you for reminding me about about:flags in Chromium. I also tried to tinker with WebGL related flags, but to no avail. It looks like Chromium is having a hard time with WebGL. Have you tried with the latest version of Firefox (i.e. not Iceweasel as it is deprecated)? BTW I'm using Debian Jessie with Google Chrome from its repo (http://dl.google.com/linux/chrome/deb/ stable main) and Firefox which is provided by the Debian repos.

gberaudo:
@GaborFarkas, I am using Ubuntu Xenial 16.04.1 LTS with:
- chromium 53.0.2785.143-0ubuntu0.16.04.1.1254 from the Ubuntu repos;
- firefox 49.0.2+build2-0ubuntu0.16.04.2 from the Ubuntu repos.

gberaudo:
@GaborFarkas, I was already able to test the draw-feature example on one other computer, it worked great on both chromium and Firefox. It is very cool to see this running!

ThomasG77:
I've done some tests. Adding in the "Vector Layer", the renderer `renderer: /** @type {ol.renderer.Type} */ ('webgl'),` http://localhost:3000/build/examples/vector-layer.html, I got errors related to Text support like below

web-gl-issue-text

I don't know if text support was intended in this PR.

marcjansen:
> I don't know if text support was intended in this PR.

Initially it wasn't AFAICT. But the current suggestion from @gberaudo was IMO to have a placeholder implementation which doesn't draw texts but also doesn't throw an error.

GaborFarkas:
Thank you for your review, @gberaudo , I will implement your suggestions later today.

@ThomasG77 text renderer was indeed not intended in this PR. The same goes for dashed lines and vector tiles.

How about adding a no-op text renderer for now? Everyone agrees on this one?

marcjansen:
> How about adding a no-op text renderer for now? Everyone agrees on this one?

I am :+1: on this. Makes it easier to get this in (less errors). As long as we clearly communicate this.

gberaudo:
I tested this morning on two more machines and it worked great. I am the only unlucky one?

GaborFarkas:
I guess so. Both your OS and Firefox are much more recent than mine. Maybe some hardware related problems? A missing third party driver for an exotic video card perhaps?

gberaudo:
It is a plain integrated Intel card, which works fine pretty fine with Cesium.

GaborFarkas:
Very weird. Maybe we should ask for some advice from the Cesium team? If my info is correct, @pjcozzi is an OpenGL and WebGL expert who could support us with some very valuable advices/guidelines.

ThomasG77:
@gberaudo I have similar issues

## <http://localhost:3000/build/hosted/webgl_vector/examples/geojson.html?q=vector&mode=raw>

In **Chrome**, I got a crash too
```
Assertion failed: illegal state, shader not compiled or context lost		context.js:221
Assertion failed: illegal state, shader not linked or context lost		context.js:251
WebGL: INVALID_OPERATION: getUniformLocation: program not linked	defaultshader.js:94
 WebGL: INVALID_OPERATION: getUniformLocation: program not linked	defaultshader.js:100
 WebGL: INVALID_OPERATION: getUniformLocation: program not linked	defaultshader.js:106
 WebGL: INVALID_OPERATION: getUniformLocation: program not linked	defaultshader.js:112
 WebGL: INVALID_OPERATION: getUniformLocation: program not linked	defaultshader.js:118
 WebGL: INVALID_OPERATION: getAttribLocation: program not linked	defaultshader.js:124
 WebGL: INVALID_OPERATION: getAttribLocation: program not linked	defaultshader.js:130
 WebGL: INVALID_OPERATION: getAttribLocation: program not linked	defaultshader.js:136
 WebGL: INVALID_OPERATION: getAttribLocation: program not linked	defaultshader.js:142
 WebGL: INVALID_OPERATION: getAttribLocation: program not linked	defaultshader.js:148
 WebGL: INVALID_OPERATION: useProgram: program not valid	context.js:321
 GL_INVALID_OPERATION : glGenSyncTokenCHROMIUM: fence sync must be flushed before generating sync token	geojson.html:1
 WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost	geojson.html:1
```

In **Firefox** it does not display and return a different error

```
Error: WebGL: texImage2D: Chosen format/type incured an expensive reformat: 0x1908/0x1401
```
and after some time,

```
Error: WebGL: uniform1i: object from different WebGL context (or older generation of this one) passed as argument	layer.js:169:5
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argument	layer.js:172:3
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argument	layer.js:174:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	layer.js:176:3
Error: WebGL: uniform2fv: object from different WebGL context (or older generation of this one) passed as argument	index.js:469:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:470:3
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentreplay.js:313:3
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentreplay.js:315:3
Error: WebGL: uniformMatrix4fv: object from different WebGL context (or older generation of this one) passed as argumentreplay.js:317:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argumentreplay.js:319:3
Error: WebGL: uniform4fv: object from different WebGL context (or older generation of this one) passed as argument	index.js:627:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:628:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:629:3
Error: WebGL: drawElements: Invalid `type`: 0x1405replay.js:364:3
Error: WebGL: uniform4fv: object from different WebGL context (or older generation of this one) passed as argument	index.js:627:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:628:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:629:3
Error: WebGL: drawElements: Invalid `type`: 0x1405replay.js:364:3
Error: WebGL: uniform4fv: object from different WebGL context (or older generation of this one) passed as argument	index.js:627:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:628:3
Error: WebGL: uniform1f: object from different WebGL context (or older generation of this one) passed as argument	index.js:629:3
Error: WebGL: drawElements: Invalid `type`: 0x1405
```

## <http://localhost:3000/build/hosted/webgl_vector/examples/draw-features.html?q=vector&mode=raw>

I got the same issues as @gberaudo too for Chrome and Firefox.

IMO, the issues seem related to i915 drivers support on Ubuntu with kernels 4.x series... (default when using Ubuntu 16.04).
I get also similar crash/reset logs when looking at dmesg

An additional point to sort the issues could be the "Problems detected" in chrome://gpu/
pb-detected-gpu

GaborFarkas:
Thanks for clarifying the problem's nature, @ThomasG77. We should see what we can do about this in the future. Cesium running on your machines indicates these problems can be fixed, we just need to find out how.

GaborFarkas:
There was a pesky little bug in the circle renderer. If we zoom in the circle, and its center gets far enough from the view's center, the stroke gets pixelated and the fill starts to jitter. As a hotfix, I collapsed the rendering (only for the given circle), if the radius of the circle gets more, than 10 000 screens wide. What do you think about this approach?

gberaudo:
Isn't there somehting similar in the canvas renderer?

GaborFarkas:
I didn't study the canvas renderer before. Thank you for the tip, I fiddled with the same circle using the canvas renderer, and the problem is the same there. Before the circle disappeared, I could see the same visual artifacts and jittering. I'm currently trying to make the webgl circle renderer a little more stable, but if nothing works, I'll stick with the current solution.

GaborFarkas:
If you don't have more suggestions, this PR reached its final stage. Future work includes:

- A functional text replay
- Line dash support
- Clipping lines to rotated clip space borders
- Caching triangulation information
- Vector tile support
- Micro-optimisations such as going through the ear clipping algorithm again, using mat3s instead of mat4s etc.

asdawn:
Very good, thanks, now more than 7000 grids can be shown in one second!
But the display region sometimes is smaller than it should be, I think it might be caused by the viewport settings. 
bug

tchandelle:
Could you take a look at #6123. My solution seems too easy to be true, but you are probably more familiar to it than me.

gberaudo:
@tchandelle, indeed it is a bug. I have added a comment to your commit. Is this commit part of a PR?

GaborFarkas:
@asdawn I'm glad you can benefit from this feature. It's working as expected on Chrome, Opera, Firefox, supports HDPI, but I don't use any commercial software, thus I don't know about IE or Edge. I will look at the problem when I'll have access to a Windows machine.

@tchandelle Please check your solution with `rotateWithView` set to `true`. If it works as expected, it's just a bug.

tchandelle:
@gberaudo I've replaced my commit with your comments : 1ce2cd605ae8ae39a1c1a78a113e5372e5050eff

@GaborFarkas I've tested both solutions, they both work.

GaborFarkas:
@tchandelle  Then it's both easy and true. :) Good catch!

schmidtk:
I'm seeing serious rendering artifacts when enabling the WebGL renderer with the `polygon-styles` example. Here's a screen cap:

Polygon Styles

This is in Chrome 54 on a Mac mini (Late 2014), Intel Iris graphics card.

GaborFarkas:
@schmidtk The example is rendered perfectly for me in Chrome, Firefox and Opera. Chrome 52 here.

schmidtk:
@GaborFarkas I'm seeing the same behavior in Chrome 54, Firefox 48, and Safari 10. I'm guessing it's related to the graphics card, but I use Cesium frequently and have never experienced this so it's unlikely to be a general WebGL issue. I'll test on other machines in our office.

GaborFarkas:
Thank you! It would be nice to track down the problematic code part. However, without a problematic machine, I won't be able to do so, thus any help is appreciated.

GaborFarkas:
The weird thing in this error is I don't remember any test cases I used the `RegularShape` class. I would risk to say, I did not use it at all. Any hints on a solution? If I comment out line 25 in regularshape.js, everything is fine.

asdawn:
@GaborFarkas 
I think I found the reason, it is related with zoom level (web browser's under Linux and Windows, system's global setting's under Windows ) . Later I'll try to run the tests. Maybe you can reproduce similar things just set your browser's zoom level to somehow larger than 100%.
The clues are:
Under Windows 10,
1. 100% global zooming + ff/edge 100% (original size),  okay
2. 150% global zooming + ff , the problem occurs.
3. 150% global zooming + edge 100%,  okay.
Under Linux Mint 18,
1. Linux Mint 100% ,120% font zooming + ff 100%, okay.
2. Linux  Mint 100% font zooming + ff 120%,, the problem occurs.
Further more, everything using the original settings and open the page, its's okay, even zoom it; but if open it with zoomed settings, the problem occurs, even you rescale to 100% it still exists.

GaborFarkas:
@asdawn Thank you for the clarification. That problem should be fixed by now, I did not multiply the screen dimensions with the pixel ratio when using `gl.scissor`.

The problem I can't get my head around is the current build error. I'm seeking help from devs, who are familiar with the build tools OL3 uses. It produces a short and hardly traceable error message when running `make test`:
``` javascript
must provide either "radius" or "radius1"
   in RegularShape at http://127.0.0.1:3001/test/loader.js/home/travis/build/openlayers/ol3/src/ol/style/regularshape.js:25
make: *** [test] Error 1
```
As my master branch runs correctly, and my PR branch built successfully prior to the rebase, I suspect some kind of regression, however, I'm not sure about that.

tchandelle:
I recently replaced ol.style.Circle by RegularShape in #6146. Is this related ?

GaborFarkas:
@tchandelle Related indeed. Thank you, prior to your refactoring, I could instantiated an empty `ol.style.Circle` constructor, which is now impossible in debug mode. The tests I created for `ol.render.webgl.Immediate` for greater test coverage contained such an instantiation.

tchandelle:
I think you shouldn't instantiate styles and geometries in `describe` but in a `beforeEach` instead. Asserts in `describe` fail all the tests, but asserts in hooks and *it*s don't.

GaborFarkas:
Awesome tip! Thank you!

schmidtk:
@asdawn Good tip! On the retina Macbook Pro, adjusting the browser zoom to 50% rendered to the full canvas. Sounds like @GaborFarkas has it resolved in the renderer. Thanks!

As for my [first issue](https://github.com/openlayers/ol3/pull/5462#issuecomment-261990020) (rendering artifacts), it seems to be a bug in older (2014) Intel graphics chipsets. It affects a 2014 Intel Iris card, but not a 2015 Intel Iris Pro.

I found [an old bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=414280) from the Cesium team related to Intel graphics on Macs and that issue seems to be fixed now. It's possible they may have some insight on Intel-related problems.

GaborFarkas:
Sounds like refactoring the linestring GLSL is imminent.

gberaudo:
@schmidtk I remember well that Cesium issue. It was actually also a problem on my Linux machine.
The issue was coming from a variable not initialized in the shader.

GaborFarkas:
I refactored the LineString GLSL, now the register permissions are a little bit stricter. Also, the refactoring made the code more transparent, so it should be easier to track down the error. I do not assume, it is now working on the problematic Intel chipsets, although could you give it a try, @schmidtk?

schmidtk:
@GaborFarkas The cropping issue I reported on the retina MBP is resolved, thank you!

I will be out of the office the rest of this week so I can't test the rendering artifact issue on the Mac mini until Monday. I will check it out ASAP when I get back in. I did test on a late 2013 MBP (older than the mini) with an Intel graphics chipset and it works great there.

Thanks again for the quick support!

marcjansen:
I ran into another issue with regard to colors somehow (too many decimals for the opacity) and created #6151 (probably replaced by #6164). This is only marginally related to this but I wanted to mention for good measure.

Otherwise this still looks awesome to me and I think it is ready for primetime. As I stated earlier, I am not the right guy for a detailed and final review, so if e.g. @gberaudo, @tschaub or @ahocevar could give green lights, that would be cool.

gberaudo:
I tested again on my machine with both Firefox and Chromium.
Great news, it now works perfectly!  :sunglasses:

@GaborFarkas thanks again for this great contribution and the followup fixing issues.
@marcjansen, to me the PR is now looking very good. +1 for merging it now.

GaborFarkas:
WOW! Frankly, I did not do anything besides
- Giving permissions to registers with `inout` modifiers and functions
- Checking floating points converted from integers (e.g. `1.0`) for equivalence with EPSILON value (a total waste of resources in my eyes, but who knows)
- Assigning full floating points instead of half ones (e.g. `1.0` instead of `1.`)

I can't imagine how these modifications could solve the rendering issue on your machine, but I'm leaving this "changelog" here if anyone needs it in the future.

asdawn:
The latest version is very good. Everything goes well now.
Data source: 
 point: 20000 POIs  from WFS
 polygon: about 7000 grids (0.005 degree)  from WFS
 raster: base map tiles, using the URL directly
Platforms:
 1. Windows 10, Intel/Nvidia graphic cards, Chrome X64 52/54, Edge, FF 43, Opera 43, IE 11
 2. Linux Mint 18, Intel graphic card, FF 49
 3. Android 5.0 phone, Qualcomm 618, default browser
All these can get correct content. Page can be opened in 2 second, zooming and moving are quite f
The most happy thing is even my phone can open the page in 3 seconds, and the lag is <0.5s on PC and 1s on phone. Using canvas there will be a two second lag after dragging the map even using very good hardware.
Thanks again.

marcjansen:
@gberaudo 
>
> @marcjansen, to me the PR is now looking very good. +1 for merging it now.
>

Cool. I am also :+1:

@GaborFarkas In order for this to be merged we need to have a [CLA](https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#contributor-license-agreement) from you. Did you sign it?

GaborFarkas:
Thank you @gberaudo, @marcjansen! I already signed it when I made my first contribution.

marcjansen:
Thanks to all the 13 participants that worked on this PR by providing, code, comments, tests or reviews! Great collaboration.

A special thanks to @gberaudo and of course @GaborFarkas for the major share of this work. :clap: :clap: :clap:

gberaudo:
@GaborFarkas, I am very pleased to see your PR merged in. Thanks also to you @marcjansen for your enthousiasm related to WebGL and your contiunous effort for getting work merged in.

ThomasG77:
Thanks for the great job! Happy to see this PR merged.
I just wonder if adding in new PR an example using WebGL with lines or polygons could be useful to illustrate this evolution? I didn't see any associated new example for this evolution in the PR.

GaborFarkas:
I agree with @ThomasG77, especially in increasing feedback of occasional rendering bugs on some video cards.

gberaudo:
@ThomasG77 there is already http://openlayers.org/en/latest/examples/draw-features.html but it is canvas2D only.

What about offering the possibility to select the `renderer` in all examples? Something similar to the `mode` ( production and development).

The select would have two values: `canvas` and `WebGL`.

If it makes sense to you I can work on a PR.

ahocevar:
We've had a way to select the renderer already, and we removed it only recently. I'd be more in favor of a single example (e.g. vector-layer-webgl) that shows the use of the webgl renderer.

schmidtk:
@GaborFarkas Update on the Intel graphics artifacts - cleaning up the line shader made rendering _much_ better, though there is still an occasional glitch. It's possible updating to macOS Sierra would fix the problem via graphics driver update, but we don't have approval to do so yet. I don't think it's worth pursuing further unless others are experiencing the same problem.

gberaudo:
@ahocevar I guess you are referring to https://github.com/openlayers/ol3/issues/5726 and the associated PR.
Bad timing indeed! Fine for me to do like you said. It can be revisited when we get a complete WebGL renderer (texts).

dunko:
Hi @GaborFarkas , @gberaudo  Sorry for the late addition to the post.  Something caught my eye which was the description from Gabor which I am experiencing now after a recent upgrade from OL v3.16.0 to OL v4.6.4.

"There's something weird about the select interaction. If I select a feature, the underlying feature still gets rendered, while some other seemingly random one disappears."

This is happening for us now, points and polygons, however not entirely random, the last feature added disappears.

Thanks in advance for any help you can offer.

PR #5465, opened 2016-06-13
Fix custom icon positions and rotation

ahocevar:
The http://openlayers.org/en/latest/examples/earthquake-custom-symbol.html shows positions that are completely wrong. The reason is incorrect scaling of the symbol using pixelRatio instead of scaling the coordinates.


fredj:
LGTM

PR #5471, opened 2016-06-14
Fix GPX write of xsi ns

### LINKED ISSUE #4699: GPX export not compatible with several programs e.g. Garmin Basecamp
The GPX files exported with ol2 where using tracks (trk tags with trkpt)
The GPX files exported with ol3 are using routes (rte with rtept)

Several programs like Garmin basecamp or the iPhone app "Scout" don't manage to read these files.
They can manage to read the files produced by ol2 though.

[gpx_bug.zip](https://github.com/openlayers/ol3/files/97687/gpx_bug.zip)

probins:
Fixes #4699

As this changes the `<gpx>` element itself, I had to change all the tests that compare output with input so that the input matches the new output - a bit of a pain, but it does at least test that the new attributes are being output correctly :-)

### DIFF
--- a/src/ol/format/gpxformat.js
+++ b/src/ol/format/gpxformat.js
@@ -57,6 +57,15 @@ ol.format.GPX.NAMESPACE_URIS_ = [
 
 
 /**
+ * @const
+ * @type {string}
+ * @private
+ */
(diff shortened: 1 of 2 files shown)


probins:
PS this should bring OL3 in line with OL2

probins:
review anyone?

ahocevar:
Looks good, thanks @probins.

Issue #5463, opened 2016-06-13
enums in modular context


probins:
Following on from my work on typedefs, I'm now looking at enums in the context of changing to modules. We've had some discussion of this in #5226, but here's some analysis of how they're currently used.

enums are like typedefs in that they are not part of the API and not used in user applications, but are types which may be used by those compiling their code together with the library. Unlike typedefs, they are code (lists of constants), which can be used in other parts of the library. This usage though is purely an optimization thing, enabling constants to be defined once in the library and reused elsewhere: remove them and the code would continue to work as before, just not as optimally compressed.

There are currently 65 enums in 52 files. Assuming #5429 is merged, of these, 25 are only used internally (in 1 file), 3 are used only in olx (as a type), 10 are used in 1 other file, and 5 in olx and 1 other file. So roughly 2/3 are not much used outside. 3 are used as typecasts in examples.

15 are object Property enums; these are mainly internal only, but 5 are used in 1 other file.

EventTypes (of which there are 18), ImageState, TileState, ViewHint, format.FormatType, geom.GeometryType, geom.GeometryLayout, proj.Units, source.State, style.ImageState are used in many external files.

There is no current consistency with `goog.provide`. 4 of the internal-only ones have one, and 2 of the 3 used only in olx have one. As I doubt whether the compiler needs these, I'll create a PR to remove them.

The benefit of those only used in 1 other file seems marginal to me, so I would suggest changing these to use the string literal instead. For example, `ol.OverlayPositioning` is only used in `overviewmapcontrol.js`, and if `ol.OverlayPositioning.BOTTOM_LEFT` is changed to `'bottom-left'`, this dependency, along with `goog.provide`, can be removed.

However, I'm not sure it makes sense to change those that are used in many files, so I'm thinking it might be better to keep these as they are. This does mean though that, despite not being API, they will have to be exported, so other files can use them.

gberaudo:
Aren't enums used to catch typos?
If one types `ol.OverlayPositioning.BOTTON_LEFT` instead of `ol.OverlayPositioning.BOTTOM_LEFT` the compiler will catch it. If one types `botton-left` instead of `bottom-left` the compiler will not catch it as any `string` would be OK.

probins:
yes, that's also true.

tschaub:
I don't think we should change our use of enums.  There are ample uses of enums in ES2015 modules.  And we can have modules that export objects (or "enums") *without* having those be part of the public API.

PR #5475, opened 2016-06-15
Remove goog.provide from enums only used internally or in olx.js

probins:
Another minor tidy of `goog.provide`. As mentioned in #5463, these are not needed.


fredj:
LGTM, @ahocevar any opinion ?

gberaudo:
LGTM too.

ahocevar:
Looks good to me too.

PR #5482, opened 2016-06-15
Fix for regression in tile layer extent

### LINKED ISSUE #5489: View/Tile loading does not respect extent defined (XYZ source).
Hi,

I have developed a large image viewer on top of OpenLayers3.16.

To do that, I am using a XYZ source and a custom extent.

In this example, my image size is (width, height) = (33864,22561) px. Therefore there is 9 level of zoom available.

Here is the projection I am using:
```
new ol.proj.Projection({
	code: 'pixels',
	units: 'pixels',
	extent: [0, 0, currentData.view.width, currentData.view.height]
});
```

Here is the view I am using:
```
new ol.View({
	center: [currentData.view.width / 2, currentData.view.height / 2],
	zoom: currentData.view.zoom.default
	maxZoom: currentData.view.zoom.max,
	projection: projection
});
```

Despite setting this view, OpenLayers seems to use a larger view and try to unload unexisting tiles (see image):

Here is the tiles OpenLayers tried to load:

- (x, y) = (0, 0): correct
- (x, y) = (1, 0): correct
- (x, y) = (0, 1): not-correct
- (x, y) = (1, 1): not-correct

I do not know what type info should I add to this issue to help solving.
Is this a bug, or do I use something incorrectly ?

Thanks for your help.

ischneider:
Demonstrates regression in tile layer with extent set.

The modified example (1st commit) allows for verifying the regression - use the 'draw' button and draw a rectangle extent. The intersecting tiles are rendered but not clipped to the extent. The rotation tool was also added (see below).

~~The 2nd commit removes what appears to be some (dead) remnants of #4597 left behind by subsequent changes.~~

The ~~3rd~~ 2nd commit is a fix ~~while the tiles are properly clipped to the layer's extent, when rotated, something is going wrong. I poked at this for a while but missed something... I initially attempted to extract out and reuse the clipping in `canvaslayerrenderer` but an addition offset was required from the larger canvas used when rendering rotated. When I attempted to use `rotateAtOffset` in the inlined extent clipping (as used in `canvaslayerrenderer`), this resulted in a worse form of bad.~~

Fixes #5489.


ahocevar:
Thanks @ischneider! I have to dedicate some time for a proper review. Just a quick question and a note.

*Question*: What is the regression that this pull request fixes? I get that it has to do with layer extents not being respected in a way they used to, but I'm failing to understand the details.

*Note*: Your 2nd commit removes code that is not dead. It is used for layers that do not have the 'opaque' flag set. You can see it in action e.g. in the http://openlayers.org/en/latest/examples/wms-tiled.html. Without that code, transparency in tiles would not work properly when lower resolution tiles overlap with current resolution tiles. Just zoom in and out in the wms-tiled example with your branch to see what I mean.

ischneider:
Hey @ahocevar

_Answer_:
The expected behavior is that when a tiled layer has an extent set, tiles will be clipped to that extent. The current behavior and the regression is that only the tiles that *intersect* the extent are rendered. In the modified example, you can use the 'draw' button to draw an arbitrary rectangle and witness the regression.

_Note_:
You are correct about the not dead code - I misread the scope of the pixelExtents early this morning! Resurrecting :)

ischneider:
So strange, `make check` worked when I ran it here... Looking at the example.

ahocevar:
> So strange, make check worked when I ran it here...

Try to rebase to latest master - this is probably because of a recent problem with slimerjs, which was fixed today.

ischneider:
rebased and (hopefully) fixed:

* example was failing due to missing type hint (`make ci` runs more examples checks than `make check`)
* fixed extent clipping when rotated ... found that missing bit :)

ahocevar:
Thanks for digging - the code looks good now @ischneider. Do you think you could create a render test in `test_rendering` instead of using an existing example to test? We try to keep examples simple so users can use them as learning resource. (And yes, I know we have several examples that do not meet this goal, but we're trying to not worsen the situation.)

If that sounds like too much effort, maybe you could make a simpler change to the example to clip to 4 different countries instead of hemispheres.

ischneider:
> Thanks for digging - the code looks good now @ischneider. Do you think you could create a render test in test_rendering instead of using an existing example to test?

Happy to dig! So far in my ol3 use, not much need to dig/fix :+100:

Added tests based off of existing. Tried WebGL but they failed - easy to set up based on existing tests, but implementation is beyond my current abilities/time :smile: 

> We try to keep examples simple so users can use them as learning resource. (And yes, I know we have several examples that do not meet this goal, but we're trying to not worsen the situation.)

Understood. I left the rotation control in...

> If that sounds like too much effort, maybe you could make a simpler change to the example to clip to 4 different countries instead of hemispheres.

Great idea! Done.

Also modified example description slightly - hope that works?

tschaub:
Any idea what caused the regression?  `git bisect` can be painful with the need to `npm install` at each step, but it would be easier to review if we could point to a commit that caused the regression.

tschaub:
So far I've found this example looks good in `v3.13.0` and bad in `v3.14.0` (and you have to remove `slimerjs-edge` from `package.json` to get `v3.13.0` to `npm install`).

tschaub:
According to `git bisect`, 98b823c5fcd2a9f9971b26490226f8f1da53ed2e is the first bad commit.

tschaub:
This commit makes it slightly easier to create reference images for rendering tests: https://github.com/tschaub/ol3/commit/d14726008bee3aba04204e3f4ab18ae237749cce

Still finding it painful to add a new rendering test for this ...

tschaub:
Awesome work on this @ischneider.  This looks like the right fix to me.  I'm curious how you generated the reference images without something like https://github.com/tschaub/ol3/commit/d14726008bee3aba04204e3f4ab18ae237749cce (the canvas is 100% width/height without that for me).  Feel free to cherry-pick if you agree with this addition.

I'll leave this up in case anybody else wants to review, but I think this is good to merge.

examples/layer-extent.js
```diff
@@ -3,17 +3,17 @@ goog.require('ol.View');
 goog.require('ol.layer.Tile');
 goog.require('ol.proj');
 goog.require('ol.source.TileJSON');
+goog.require('ol.interaction.DragRotateAndZoom');
```
tschaub: Note that `alt-shift-drag` rotates the map without this addition.

examples/layer-extent.js
```diff
@@ -43,6 +43,8 @@ var map = new ol.Map({
   })
 });
 
+map.addInteraction(new ol.interaction.DragRotateAndZoom());
```
ahocevar: As @tschaub mentioned, alt-shift-drag always allows you to rotate. So this could be removed.

ahocevar:
Looks great @ischneider. I left one more small note which you might or might not want to address.

ahocevar:
@ischneider, can you also please submit a [CLA](https://github.com/openlayers/cla) so we can merge this? Thanks!

tschaub:
> @ischneider, can you also please submit a CLA so we can merge this? Thanks!

We've put in a request for a CCLA from @planetlabs.  Will report back when there is an update ...

ischneider:
> I'm curious how you generated the reference images without something like tschaub@d147260 (the canvas is 100% width/height without that for me)

Maybe a platform specific thing? I just right-clicked and downloaded...

I'll follow up w/ the minor change and clean up the commits when the CCLA is ready.

tschaub:
@ischneider CCLA received.  Do you still want to make additional changes, or should we merge this as is?  It would be great to let the 3.17 release proceed now that we have the CCLA.

tschaub:
So that the 3.17 release can proceed, I'll merge this as is.  We can follow up with a second PR to address any remaining (non-critical) comments.

Thanks @ischneider for the great fix.

ischneider:
Thanks!

PR #5495, opened 2016-06-20
Get rid of goog.isFunction

fredj:
or should we add this in a function (like `ol.isFunction`)?


tschaub:
Looks good to me as well.

fredj:
Thanks for the review

PR #5506, opened 2016-06-22
Export getTileRangeForExtentAndZ and getZForResolution from ol.tilegrid.TileGrid

### LINKED ISSUE #5502: Please expose several public methods of ol.tilegrid.TileGrid in public api
Please expose the following methods of ol.tilegrid.TileGrid.prototype in your api:

1. ol.tilegrid.TileGrid.prototype.getZForResolution
2. ol.tilegrid.TileGrid.prototype.getTileRangeForExtentAndZ

I have implemented CartoDb Torque Layer for OpenLayers. I would like to contribute this implementation back to CartoDb and it is imposible without these methods.   Thanks.

fredj:
fixes #5502

### DIFF
--- a/src/ol/tilegrid/tilegrid.js
+++ b/src/ol/tilegrid/tilegrid.js
@@ -475,6 +475,7 @@ ol.tilegrid.TileGrid.prototype.getFullTileRange = function(z) {
  *     If 1, the nearest lower resolution will be used. If -1, the nearest
  *     higher resolution will be used. Default is 0.
  * @return {number} Z.
+ * @api
  */
 ol.tilegrid.TileGrid.prototype.getZForResolution = function(
     resolution, opt_direction) {
--- a/src/ol/tilegrid/tilegrid.js
+++ b/src/ol/tilegrid/tilegrid.js
@@ -305,6 +305,7 @@ ol.tilegrid.TileGrid.prototype.getTileRangeForExtentAndResolution = function(ext
  * @param {number} z Z.
  * @param {ol.TileRange=} opt_tileRange Temporary tile range object.
  * @return {ol.TileRange} Tile range.
+ * @api
  */
 ol.tilegrid.TileGrid.prototype.getTileRangeForExtentAndZ = function(extent, z, opt_tileRange) {
   var resolution = this.getResolution(z);


ahocevar:
Not sure about `#getTileRangeForExtentAndZ` - we have kept tile ranges internal until now, and I think we should keep it like this.

ashchurova:
Maybe I ask why you prefer keeping #getTileRangeForExtentAndZ internal? It returns range of tiles based on extent and zoom values. Are there complications from making it available to developers? Because to me the method is harmless if you know what I mean.

I can tell you my reasons for this request - in my implementation of Ol Torque's Layer I need to get tthe range of tiles that should me loaded based on the current extent and pass them to tourge's engine to load them. In Ol the only way I have found so far is by calling the method above. I don't think there is a need to reinvent the wheel if the functionality is already there.  If there is an alternative way to get the tile range based on the extent  I would love to know. Thanks.

ahocevar:
@ashchurova It would be easier to discuss this if we could see your code. Because other tile layer sources defer to the map's tile queue for loading tiles.

The reason why I'm hesitant to make `#getTileRangeForExtentAndZ` exportable is that the whole concept of tile ranges is currently being kept internal. If we start exposing `ol.TileRange`, we have to commit to it. The method itself may be harmless, but `ol.TileRange` and its API is quite awkward and not something our users should be confronted with.

ahocevar:
@ashchurova How about providing an `ol.tilegrid.TileGrid#forEachTileCoordInExtent` method, which you would call like this:
```js
myTileGrid.forEachTileCoordInExtent(function(tileCoord) {
  // do something with tileCoord
});
```

ashchurova:
@ahocevar  Fair enough. I am waiting for legal to approve my request today. They have already forked Torque under https://github.com/bloomberg/torque and I will commit my changes to that repository.

ol.tilegrid.TileGrid#forEachTileCoordInExtent would also work. That would be great.

Basically this is the short snippet of code but you will get the idea:

`   var zoom =  this._tileGrid.getZForResolution(this._view.getResolution());
    var extent = this._view.calculateExtent(this._map.getSize());
    var tileRange = this._tileGrid.getTileRangeForExtentAndZ(extent, zoom);
    
 this._addTilesFromCenterOut(tileRange, zoom);
`

ashchurova:
@ahocevar though I need to be able to pass the extent as well so the signature should be like: 
 
```javascript
myTileGrid.forEachTileCoordInExtent(extent, function(tileCoord) {
  // do something with tileCoord
});
```

ahocevar:
See #5509 for an implementation of the method I suggested above.

ashchurova:
@ahocevar looks great. Thank you.

fredj:
continues in #5509

PR #5507, opened 2016-06-22
Remove use of goog.vec.*

ahocevar:
This is an alternative to #5138.

After reviewing our use of 4x4 matrices, it turned out that we do not do any calculations with them. So I went ahead and implemented the few 3x3 matrix functions we need in a way that they can easily be applied to a canvas context, and added a function to set the 2d transformation on a 3d identity matrix for the places where we need it in webgl and dom code.


ahocevar:
Work in progress because I need to add tests for the functions in `ol.matrix`. But other than that, everything appears to be fully functional with this pull request applied.

tschaub:
Nice @ahocevar!  Haven't taken time for a proper review, but glad you are tackling this.

tschaub:
I know this is largely internal, but I think it would be good to discuss names a bit.  This is all about affine transformations.  The current `ol.matrix.create()` function doesn't return a 3 x 3 identity matrix, but rather parameters for an affine identity transform.  How about these names:

```
var transform = ol.transform.create();
ol.transform.rotate(transform, angle);
ol.transform.scale(transform, x, y);
ol.transform.translate(transform, dx, dy);
ol.transform.invert(transform);
ol.transform.reset(transform);
ol.transform.set(transform, a, b, c, d, e, f);
```

ahocevar:
I like these names, and I can easily change them. Thanks for the suggestions @tschaub.

ahocevar:
@tschaub Do you also have a name suggestion for the `multiply` function?

tschaub:
I assume `frameState.coordinateToPixelMatrix` could become `frameState.coordinateToPixelTransform` and `frameState.pixelToCoordinateMatrix` could become `frameState.pixelToCoordinateTransform`.  And then our current use of `multVec2` could look like this:

```js
var transform = frameState.coordinateToPixelTransform;
var pixel = ol.transform.apply(transform, coordinate);
``` 

And hopefully everybody would better understand what that means.

And for `multiply`, I guess I'd go with `ol.transform.multiply()`.

ahocevar:
@tschaub Done as you suggested.

src/ol/vec/mat4.js
```diff
@@ -1,90 +1,23 @@
...
-  return resultVec;
+ol.vec.Mat4.fromMatrix = function(mat) {
+  var mat4 = ol.vec.Mat4.tmpMatrix_;
+  mat4[0] = mat[0];
+  mat4[1] = mat[1];
+  mat4[4] = mat[2];
+  mat4[5] = mat[3];
+  mat4[12] = mat[4];
+  mat4[13] = mat[5];
+  return mat4;
```
tschaub: Two calls to this will result in the first one's return being modified, right?

src/ol/vec/mat4.js
```diff
@@ -1,90 +1,23 @@
...
-  var m10 = goog.vec.Mat4.getElement(mat, 1, 0);
-  var m01 = goog.vec.Mat4.getElement(mat, 0, 1);
-  var m11 = goog.vec.Mat4.getElement(mat, 1, 1);
-  var m03 = goog.vec.Mat4.getElement(mat, 0, 3);
-  var m13 = goog.vec.Mat4.getElement(mat, 1, 3);
-  var x = vec[0], y = vec[1];
-  resultVec[0] = m00 * x + m01 * y + m03;
-  resultVec[1] = m10 * x + m11 * y + m13;
-  return resultVec;
+ol.vec.Mat4.fromMatrix = function(mat) {
```
tschaub: This could now be `ol.vec.Mat4.fromTransform(transform)`.

ahocevar:
Tests added, all comments addressed. Ready for a final review.

test/spec/ol/transform.test.js
```diff
@@ -0,0 +1,86 @@
...
+      var transform2 = [1, 2, 3, 4, 5, 6];
+      expect(ol.transform.setFromArray(transform1, transform2)).to.eql(transform2);
+      expect(transform1).to.eql(transform2);
+    });
+  });
+
+  describe('ol.transform.translate()', function() {
+    var transform = ol.transform.create();
+    expect(ol.transform.translate(transform, 3, 4)).to.eql([1, 0, 0, 1, 3, 4]);
+    expect(transform).to.eql([1, 0, 0, 1, 3, 4]);
```
tschaub: So that mocha can handle failure, these assertions (and the ones below) need to be in `it` functions.

See https://github.com/tschaub/ol3/commit/9a3e150acf2f0db74672674a9f6470e663deaae2 for a change that does this.

tschaub:
This is a really awesome change @ahocevar.  I think this makes the code far more readable.  In addition to avoiding the extra allocation (for unnecessary 3x3 and 4x4 matrices), this removes 6 KB from the full build.

As [mentioned above](https://github.com/openlayers/ol3/pull/5507#discussion_r68426135), the new assertions need to be moved into `it` functions.

Not sure what you think about this, but the composition of T'SRT is so common, that I think it makes sense to optimize the implementation a bit.  So, where we previously did this:

```js
var transform = ol.transform.create();
ol.transform.translate(transform, dx1, dy1);
ol.transform.scale(transform, sx, sy);
ol.transform.rotate(transform, angle);
ol.transform.translate(transform, dx2, dy2);
```

I added a convenience function so we can now do this:
```js
var transform = ol.transform.compose(ol.transform.create(), dx1, dy1, sx, sy, angle, dx2, dy2);
```

While this sacrifices some legibility, I think it is nice to avoid the unnecessary matrix multiplication per step.  Because the matrix multiplication for these transforms is not commutative, the `ol.transform.compose()` function only works for this order of transforms.  This is similar to the old `ol.vec.Mat4.makeTransform2D()` function except that it doesn't do all the unnecessary matrix multiplication and carry around the extra 10 unnecessary terms.

If you agree with these changes, you can cherry pick them here: https://github.com/ahocevar/ol3/compare/remove-goog-vec...tschaub:remove-goog-vec.  At a minimum, the test changes above need to be made.

I'll leave it to you to decide whether to merge this in before or after 3.17.0.  It might be nice to let any bugs shake out first.

ahocevar:
Thanks a lot @tschaub for your help with this and the additional improvements. I cherry-picked all your commits.

I'm fine waiting with merging this until after the 3.17.0 release.

probins:
so presumably #5138 can be closed

probins:
one thing I notice with this is that `goog.require('ol.transform')` is not in alpha sequence

ahocevar:
We've removed that restriction, since the compiler does not care.

PR #5508, opened 2016-06-22
Fix problems with ol.format.WFS in compiled mode

tsauerwein:
In `ol.format.WFS` there are several places where properties are accessed via the dot-syntax and via  string references at the same time which causes problems with the compilation (for example [here](https://github.com/openlayers/ol3/blob/e289bfb/src/ol/format/wfsformat.js#L523) and [here](https://github.com/openlayers/ol3/blob/e289bfb/src/ol/format/wfsformat.js#L816-L824)).

I considered creating a `@typedef`, but I wasn't sure how to create a typedef that *inherits* from [ol.XmlNodeStackItem](https://github.com/openlayers/ol3/blob/1d20215/src/ol/typedefs.js#L669-L676), which is the base `context` structure used for XML parsing/reading.

Instead I did it like in the other XML readers/parsers (for example like in [ol.format.GPX](https://github.com/openlayers/ol3/blob/1d20215bfa/src/ol/format/gpxformat.js#L590)), and used the dot-syntax for the property defined in `ol.XmlNodeStackItem` (`node`) and used string-references for all properties specific to WFS.

### DIFF
--- a/src/ol/format/wfsformat.js
+++ b/src/ol/format/wfsformat.js
@@ -106,10 +106,10 @@ ol.format.WFS.prototype.readFeatures;
  * @inheritDoc
  */
 ol.format.WFS.prototype.readFeaturesFromNode = function(node, opt_options) {
-  var context = {
+  var context = /** @type {ol.XmlNodeStackItem} */ ({
     'featureType': this.featureType_,
     'featureNS': this.featureNS_
-  };
+  });
   ol.object.assign(context, this.getReadOptions(node,
       opt_options ? opt_options : {}));
   var objectStack = [context];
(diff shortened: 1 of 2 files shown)


src/ol/format/wfsformat.js
```diff
@@ -416,7 +416,7 @@ ol.format.WFS.writeDelete_ = function(node, feature, objectStack) {
 
 
 /**
- * @param {Node} node Node.
+ * @param {Node} node Node.q
```
tschaub: Extra `q`.

tsauerwein: Thanks!

fredj:
LGTM

tsauerwein:
Thanks for the review.

PR #5509, opened 2016-06-22
Make it easier to work with tile ranges in custom source implementations

### LINKED ISSUE #5502: Please expose several public methods of ol.tilegrid.TileGrid in public api

ahocevar:
This is an alternative to #5506. Instead of exposing tile ranges, this pull request makes a tile grid's tile coordinates available by extent and zoom level.

I do not like the name of the new function though. Any suggestions anyone? @tschaub maybe, since you're in a naming mood today it seems?

Fixes #5502.

### DIFF
--- a/src/ol/tilegrid/tilegrid.js
+++ b/src/ol/tilegrid/tilegrid.js
@@ -151,6 +151,24 @@ ol.tilegrid.TileGrid.tmpTileCoord_ = [0, 0, 0];
 
 
 /**
+ * Call a function with each tile coordinate for a given extent and zoom level.
+ *
+ * @param {ol.Extent} extent Extent.
+ * @param {number} zoom Zoom level.
+ * @param {function(ol.TileCoord)} callback Function called with each tile coordinate.
+ * @api
+ */
+ol.tilegrid.TileGrid.prototype.forEachTileCoordInExtentAndZ = function(extent, zoom, callback) {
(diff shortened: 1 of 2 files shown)


ashchurova:
Do you have an eta? Thanks.

ahocevar:
This needs to be reviewed by another trusted developer. Since this is not paid work, it might take a while. But it shouldn't be too hard to review this one.

ahocevar:
@fredj, I cherry-picked your other commit from #5506, so this pull request replaces that one. A review would be much appreciated.

fredj:
LGTM, thanks Andreas

tschaub:
For what it's worth, I don't think method names need to repeat the arguments.  So, instead of this:

```js
grid.forEachTileCoordInExtentAndZ(extent, z, callback);
```

my preference would be this:

```js
grid.forEachTileCoord(extent, z, callback);
```

ahocevar:
Thanks @tschaub, this was the feedback I was hoping for :-). Please review #5512.

PR #5511, opened 2016-06-23
Do not optimize lines away entirely

ahocevar:
Currently, if a linestring has only two points, and if these two points are equal, they might not get rendered at all.

Fixes the issue described in https://groups.google.com/d/msgid/ol3-dev/30136824-347b-464c-a246-7d25e4b9da8f%40googlegroups.com.


tsauerwein:
Rendering test? :) Or is the issue not reproducible all the time?

ahocevar:
@tsauerwein Rendering test added.

tsauerwein:
Perfect, thanks! Your change looks good to me.

PR #5512, opened 2016-06-23
Use forEachTileCoord name instead of forEachTileCoordInExtentAndZ

ahocevar:
Name change, as suggested in https://github.com/openlayers/ol3/pull/5509#issuecomment-228033504.


tschaub:
Looks good to me :)

Please merge.

tschaub:
@bartvde - it would be nice to get this change in before cutting 3.17.

PR #5513, opened 2016-06-23
Add getters for min and max resolution of the view

bartvde:
In order to support external sliders, it would make sense to have getters for min and max resolution of the view.


ahocevar:
Sliders are so retro! But making these available can be useful regardless. Please merge.

PR #5516, opened 2016-06-24
Remove goog.Uri from KML format

### LINKED ISSUE #5486: Replacing goog.Uri
This is only used in KML format. It would be possible to reproduce the current code, but essentially this is already natively available as the URL() constructor in all browsers except IE https://developer.mozilla.org/en-US/docs/Web/API/URL/URL My suggestion would be to use URL() and add a comment to the docs and the release notes that those wanting to support IE should use a polyfill, for example, from cdn.polyfill.io

probins:
I think this is ready for review now. As discussed in #5486. Fixes #5486.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -4,6 +4,10 @@
 
 This option was previously needed to use named colors with the WebGL renderer but is no longer needed.
 
+#### KML format now uses URL()
+
+The URL constructor is supported by all modern browsers, but not by older ones, such as IE. To use the KML format in such older browsers, a URL polyfill will have to be loaded before use.
+
 ### v3.17.0
 
 #### `ol.source.MapQuest` removal
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 ## Supported Browsers
 
-OpenLayers runs on all modern browsers that support [HTML5](https://html.spec.whatwg.org/multipage/) and [ECMAScript 5](http://www.ecma-international.org/ecma-262/5.1/). This includes Chrome, Firefox, Safari and Edge. For older browsers and platforms like Internet Explorer (down to version 9) and Android 4.x, [polyfills](http://polyfill.io) for `requestAnimationFrame` and `Element.prototype.classList` are required.
+OpenLayers runs on all modern browsers that support [HTML5](https://html.spec.whatwg.org/multipage/) and [ECMAScript 5](http://www.ecma-international.org/ecma-262/5.1/). This includes Chrome, Firefox, Safari and Edge. For older browsers and platforms like Internet Explorer (down to version 9) and Android 4.x, [polyfills](http://polyfill.io) for `requestAnimationFrame` and `Element.prototype.classList` are required, and using the KML format requires a polyfill for `URL`.
 
 ## Documentation
 
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -11,7 +11,7 @@
     <link rel="stylesheet" href="./resources/layout.css" type="text/css">
     {{{ extraHead.local }}}
     {{{ css.tag }}}
-    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,requestAnimationFrame,Element.prototype.classList"></script>
+    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,requestAnimationFrame,Element.prototype.classList,URL"></script>
     <script src="./resources/zeroclipboard/ZeroClipboard.min.js"></script>
     <title>{{ title }}</title>
   </head>
@@ -74,7 +74,7 @@
     &lt;title&gt;{{ title }}&lt;/title&gt;
     &lt;link rel="stylesheet" href="http://openlayers.org/en/v{{ olVersion }}/css/ol.css" type="text/css"&gt;
     &lt;!-- The line below is only needed for old environments like Internet Explorer and Android 4.x --&gt;
-    &lt;script src="http://cdn.polyfill.io/v2/polyfill.min.js?features=requestAnimationFrame,Element.prototype.classList"&gt;&lt;/script&gt;
+    &lt;script src="http://cdn.polyfill.io/v2/polyfill.min.js?features=requestAnimationFrame,Element.prototype.classList,URL"&gt;&lt;/script&gt;
     &lt;script src="http://openlayers.org/en/v{{ olVersion }}/build/ol.js"&gt;&lt;/script&gt;{{#if extraHead.remote}}
 {{ indent extraHead.remote spaces=4 }}{{/if}}{{#if css.source}}
     &lt;style&gt;
--- a/doc/tutorials/introduction.md
+++ b/doc/tutorials/introduction.md
@@ -26,7 +26,7 @@ Unlike in, say, Node, where a module's exports are fixed in the source, with Clo
 ## Renderers and Browser Support
 The library currently includes three renderers: Canvas, DOM, and WebGL. All three support both raster data from tile/image servers, and vector data; WebGL however only supports Point vectors and does not support labels. Clearly only those browsers that [support Canvas](http://caniuse.com/canvas) can use the Canvas renderer. Equally, the WebGL renderer can only be used on those devices and [browsers](http://caniuse.com/webgl) that support WebGL.
 
-OpenLayers runs on all modern browsers that support [HTML5](https://html.spec.whatwg.org/multipage/) and [ECMAScript 5](http://www.ecma-international.org/ecma-262/5.1/). This includes Chrome, Firefox, Safari and Edge. For older browsers and platforms like Internet Explorer (down to version 9) and Android 4.x, [polyfills](http://polyfill.io) for `requestAnimationFrame` and `Element.prototype.classList` are required.
+OpenLayers runs on all modern browsers that support [HTML5](https://html.spec.whatwg.org/multipage/) and [ECMAScript 5](http://www.ecma-international.org/ecma-262/5.1/). This includes Chrome, Firefox, Safari and Edge. For older browsers and platforms like Internet Explorer (down to version 9) and Android 4.x, [polyfills](http://polyfill.io) for `requestAnimationFrame` and `Element.prototype.classList` are required, and using the KML format requires a polyfill for `URL`.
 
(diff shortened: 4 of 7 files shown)


changelog/upgrade-notes.md
```diff
@@ -33,6 +33,10 @@ Users compiling their code with the library should note that the following typed
 * ol.style.RegularShapeRenderOptions to ol.RegularShapeRenderOptions
 * ol.style.StyleFunction to ol.StyleFunction
 
+#### KML format now uses URL()
+
+The URL constructor is supported by all modern browsers, but not by older ones, such as IE. To use the KML format in such older browsers, a URL polyfill will have to be loaded before use.
```
marcjansen: Which IE versions will need the polyfill?

probins: all

marcjansen: Ouch. Thanks.

bartvde: I wonder if this will lead to lots of questions / inconvenience? I still think we should consider using https://www.npmjs.com/package/url-parse It should simply work on the latest IE out of the box IMHO?

marcjansen: Exactly my thought.

probins: if you don't use URL(), then I would prefer putting the code in the KML format. It's not a big deal, as it's only appending a string to a base url.

bartvde: Yeah that approach would also work for me.

bartvde: But isn't it also translating a relative url into an absolute url in the case of the KML format?

probins: I'll check that out. The tests are a better fit for URL(), as they're testing all the pieces of the uri, so we can keep the polyfill there.

probins: > But isn't it also translating a relative url into an absolute url in the case of the KML format?

the only logic for which `goog.Uri` is used (in 4 places in KML format, mainly for styles/stylemaps, but also for other links) is:

if string is absolute url (starts with http or https), use that, else append string to node.baseURI

This could be added, to the KML format or perhaps better to `ol.uri`. Unlike the tests, which parse a url into its constituent parts, the KML format doesn't need any real parsing of the url other than determining if it's absolute or relative.

I don't think using an npm package is a good idea, as it penalises the majority using modern browsers (extra code overhead) just to support the minority using old browsers.

ahocevar:
Ok, to move this forward: I agree with @probins. It makes sense to use the `URL()` constructor and a polyfill in the tests, while adding custom code to a `ol.uri` namespace to support the one use case in `ol.format.KML`.

Sorry for the back and forth on this. @probins, would you still be willing to update this pull request as outlined above, so everyone is happy?

probins:
yes, ok. probably tomorrow

probins:
ok. Changed so URL() no longer needed in src but still used in tests. I've simplified the code a bit and made it a bit more consistent. `node.baseURI` is always absolute so there's no need for any check on that. So I don't see the need for any function in `ol.uri`.

bartvde:
@probins have you checked Firefox? See: https://github.com/openlayers/ol3/issues/3439

probins:
don't understand https://github.com/openlayers/ol3/issues/3439#issuecomment-89537262 - `node.baseURI` is not an empty string in Chrome (just double-checked using http://openlayers.org/en/master/examples/kml.html?mode=raw). In fact, I'm not sure the check for `node.baseURI` is actually needed; according to https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI it should be always set to the absolute uri.

bartvde:
hmm maybe this was changed in Chrome then

probins:
https://dom.spec.whatwg.org/#dom-node-baseuri - "The baseURI attribute’s getter must return node document’s document base URL, serialized"

bartvde:
Code LGTM @probins have you verified that KML unit tests pass in all major browsers that we support?

probins:
no, I don't have access to Windows/Mac. If that's the objective, perhaps getting Travis to run tests on Sauce Labs would be a good idea? (Though the advantage of standards-based browsers _should_ be that they should all Just Work. :-))

bartvde:
@probins I have 2 test failures in Firefox on your branch:

screen shot 2016-06-30 at 13 05 32

probins:
yes, well, there is rather a basic error in the code; I have no idea how this passed the tests on Travis. Carrying out more extensive tests before pushing a correction.

probins:
@bartvde could you try again?

bartvde:
One failure left @probins 

screen shot 2016-06-30 at 15 21 53

probins:
ok, I think I now understand your previous comment on `node.baseURI` being null on Chrome. According to the KML spec, "Relative URLs can be used in this tag and are evaluated relative to the enclosing KML file." So if for example one of the files used in OL's KML examples, say http://openlayers.org/en/master/examples/data/kml/2012-02-10.kml, had a NetworkLink `node.BaseURI` would = that filename. However, in the case of this test, the KML file is created from a string, so there is no 'enclosing KML file'. In this case in Chrome `node.baseURI` is indeed null. It looks like Firefox is using `document.baseURI` instead.

Not sure this test is meaningful. If you want to test relative URLs, ISTM it would be better to lead the test from a file rather than from a string.

bartvde:
Would it make more sense then to change to test to be more meaningful? If you think that is the way forward, I agree.

probins:
> If you want to test relative URLs

it's not just the relative ones: if `node.baseURI` is null, the code skips the relative/absolute check and just outputs the incoming value. This would also explain why Travis passed my previous code even though it had an error in it: it was just testing that a==a.

If we change the test, I think that should be in a separate PR.

probins:
I've been checking this URI subject in more detail. To summarise:
* KML format is the only place in the library where `goog.Uri` is used
* there are 2 types of URI supported in the KML format:
  * `<styleUrl>` elements refer to `<Style>` or `<StyleMap>` elements. These are prefixed with '#', and can be in the same file, or in an external file where the reference can be an absolute or relative Url: examples from KML spec `<styleUrl>#myIconStyleID</styleUrl>`, `<styleUrl>http://someserver.com/somestylefile.xml#restaurant</styleUrl>`, `<styleUrl>eateries.kml#my-lunch-spot</styleUrl>`
  * `<href>` elements occur in Icon and Link elements. These are standard links to an external file with an absolute or relative Url
* these are handled by `readStyleUrl_`, `readSharedStyle_`, `readSharedStyleMap_`, `readURI_`
* there should be tests for all these cases, but only StyleUrl in the same file, absolute Icon, and absolute and relative Link hrefs are currently being tested
* the code uses `node.baseURI`; the normal case is that the KML is in a file, and in this case `node.baseURI` will be the absolute Url of that file. However, the KML could also come from a string, for example, when created programatically, and in this case browsers seem to vary; in Chrome it's null, but Firefox seems to set it to `document.baseURI`. If it's null, the code does not use `goog.Uri` to create an absolute url, but simply returns what was passed in
* the tests create KML from a string; PhantomJS uses QtWebkit, which like Chrome sets `node.baseURI` to null. So although it looks like the tests are currently passing, in fact they're simply testing that what was passed in gets returned, which is always the case; this isn't very useful

I don't think this PR is currently handling all the above cases correctly, so I'll review this and resubmit.

There is though the question of what baseURI should be used if the KML comes from a string. Should it follow Firefox and always use `document.baseURI`?

Although this doesn't have anything to do with replacing `goog.Uri`, tests should be added to fetch styles in external files. As the common case is that the KML is in a file, they should test this. There should also be tests for external links in KML read from a file. Should these replace the current KML from strings, or be in addition?

probins:
what I'll do is separate the 2 parts of this PR:

1. this PR can be for replacing `goog.Uri` in kml; although adding new tests isn't strictly for that, we need proper tests to be sure the new code works properly
2. I'll create a new PR for replacing `goog.Uri` in the non-kml tests

probins:
looks like there's another issue in kml url handling: `readStyleUrl_()` resolves the url relative to `node.baseURI`, but this is only executed when the `styleUrl` element is a child of a `StyleMap`, not when it is a child of a `Placemark`.

ahocevar:
I still agree with this solution. Thanks @probins, and sorry for the back and forth. @marcjansen, would you agree too if we had a prominent place in the documentation that lists supported browsers and required polyfills?

probins:
bear in mind too that use of these old browsers is slowly but steadily decreasing. According to http://gs.statcounter.com/#all-browser-ww-monthly-201506-201606 IE is now below 10%.

I'm currently looking into adding tests. This can either be added to this PR or put in a separate one.

ahocevar:
@probins, please proceed with the tests as you prefer. I guess the only thing really required to finalise this pull request is to update the polyfills list in `README.md` and `docs/quickstart.hbs`.

probins:
docs updated. It will probably be next week before I can get around to the tests. In the meantime, there is #5558 to review.

probins:
ready for review again. I added a couple of tests, but it looks like `node.baseURI` is null when reading in files too, so I'm not sure the NetworkLink test I've added is worth the bother; this may well be just a limitation of the test environment. I also found that, although `node.baseURI` is set when using `XMLHttpRequest()`, it is null when you `fetch()` a file, at least in Chrome - this could well be a little bug in Chrome's implementation of `fetch()`. I have tested in a proper web environment and apart from this it all seems to work correctly.

ahocevar:
@probins Not sure I entirely understand the implications of your comment above. To me, it appears that everything works as expected now. Are you saying that in Chrome, resolving relative network links will fail when the KML file was retrieved with `fetch()`? Also, what limitation of the test environment are you suspecting?

probins:
> Are you saying that in Chrome, resolving relative network links will fail

if `node.baseURI` is null, then the `URL()` (or `goog.Uri`) logic is skipped, which means that relative URLs (for styles and icons as well as network links) are resolved relative to the document rather than the kml file, which may not be the same. I thought at first that this was only in the case where the kml is created from a string, but there seems to be a wider problem. I've not tested in other browsers to see what they do. In practice, this may not matter too much, as I suspect that few links in kml files will use a relative url anyway.

ahocevar:
Thanks a lot @probins. The code looks great, and I'm just going to merge it.

In case issues come up, they can be handled in separate pull requests, and the respective fixes should include tests.

PR #5529, opened 2016-06-29
Optimize no exports generation path

gberaudo:
Immediately return an empty string if no export must be generated.

### DIFF
--- a/tasks/generate-exports.js
+++ b/tasks/generate-exports.js
@@ -187,6 +187,11 @@ function generateExports(symbols, namespace) {
  *     error generating it.
  */
 function main(config, callback) {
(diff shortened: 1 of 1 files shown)


tschaub:
@gberaudo - I'm curious where `main` is getting called with `exports: []`.

gberaudo:
@tschaub, see https://github.com/camptocamp/ngeo/blob/master/buildtools/mako_build.json#L19-L20 for example. It is done this way in all our combined builds (application + library).

marcjansen:
This doesn't hurt, does it? Can we merge this?

tschaub:
We'll be reworking the build tooling for the 5.0 release.  So I'm closing this for now.

PR #5533, opened 2016-06-30
Make sure reprojected tiles are recreated on updateParams

ahocevar:
Because we are very smart when it comes to re-fetching tiles after calling `updateParams`, tile caches in other projections need some smart tile invalidation handling too.


bartvde:
@ahocevar is this the correct changeset for this PR? I can't really match your description with the diff?

ahocevar:
@bartvde You're right, I messed things up. Updated pull request just now.

ahocevar:
Updated pull request with a much smarter solution - no flickering, no clearing the reprojected cache.

anzeznidarsic:
Perfect fix, works for our case where we're reprojecting some of the layers.

bartvde:
LGTM @ahocevar it might make sense to update the first paragraph (top section) of this PR accordingly.

PR #5534, opened 2016-06-30
Fix ScaleLine control documentation

### LINKED ISSUE #5503: ScaleLine size is calculated incorrectly with EPSG4326
The ScaleLine is supposed to adjust its size when you pan vertically, as you can see in the OL Example: http://jsfiddle.net/g6t89mxb/. When you pan the map up and down, the scale line gets wider and smaller.

**But**, when you change the View projection to `EPSG:4326`, the scale line does not adjust its size anymore: http://jsfiddle.net/jtbtotq2/

I think it was caused by this commit: https://github.com/openlayers/ol3/commit/6b8da1237c4009974c351e0c43de283ad5f9310f#diff-a431a0bf935fd3aa3295a310b6a39a42R196

EPSG:4326 uses degrees, and it appears to implement the `.getPointResolution` method incorrectly:
https://github.com/openlayers/ol3/blob/29c8af755930c2ad1633377ffceb61fa8eab0a84/src/ol/proj/epsg4326projection.js#L40
```
ol.proj.EPSG4326_.prototype.getPointResolution = function(resolution, point) {
  return resolution;
};
```

Because the point resolution should change as the latitude changes, right? One pixel at the north pole is a fraction of the distance of one at the equator. So the implementation of... wait.. EPSG4326 uses degrees, so the degree distance would be the same at every latitude. So this method is correct.

**So actually**, it would seem that the `ol.control.ScaleLine.prototype.updateElement_()` method is the thing that isn't working well with ESPG4326 (or any projection that uses degrees as its units, I think).

ahocevar:
The scale line shown is derived from y-axis distances, not x-axis.

Fixes #5503.

### DIFF
--- a/src/ol/control/scalelinecontrol.js
+++ b/src/ol/control/scalelinecontrol.js
@@ -36,10 +36,11 @@ ol.control.ScaleLineUnits = {
 
 /**
  * @classdesc
- * A control displaying rough x-axis distances, calculated for the center of the
- * viewport.
- * No scale line will be shown when the x-axis distance cannot be calculated in
- * the view projection (e.g. at or beyond the poles in EPSG:4326).
+ * A control displaying rough y-axis distances, calculated for the center of the
+ * viewport. For conformal projections (e.g. EPSG:3857, the default view
+ * projection in OpenLayers), the scale is valid for all directions.
+ * No scale line will be shown when the y-axis distance of a pixel at the
+ * viewport center cannot be calculated in the view projection.
  * By default the scale line will show in the bottom left portion of the map,
  * but this can be changed by using the css selector `.ol-scale-line`.
  *


bill-chadwick:
It might be worth the docs mentioning that for many 'common' projections, the y-axis and x-axis scales at the center are the same and that 4326 is an 'uncommon' case.

ahocevar:
@bill-chadwick Do you have any suggestion for exact wording? Then I could quickly update the pull request with that.

bill-chadwick:
how about 

"For 'Conformal' projections, the x-axis scale at the center of the view will be the same as the y-axis scale."

ahocevar:
Thanks @bill-chadwick, I like it. Pull request updated.

ahocevar:
Anyone want to give this a quick review? It's a doc only change.

bill-chadwick:
Andreas,

I know it was my sentence. Conformal does mean that scale is the same in
all directions about a point.

See  http://www.progonos.com/furuti/MapProj/Dither/ProjConf/projConf.html

Bill

probins:
not sure how many developers will know what a conformal projection is, though. I would imagine that most developers using OL are not cartographers/geographers.

ahocevar:
I updated the pull request again, hopefully with better wording.

marcjansen:

PR #5536, opened 2016-06-30
Update to layer-extent example

tschaub:
This follows up on the remaining comments in #5482 by removing the extra interaction from the layer extent example.


bartvde:
thanks @tschaub

PR #5541, opened 2016-07-01
Replace goog.isNumber

probins:
`goog.isNumber` just does `== 'number'` but I use `===` to be consistent with string


tschaub:
Looks good.

PR #5543, opened 2016-07-01
Update rbush to 2.0.1

tsauerwein:
Closes https://github.com/openlayers/ol3/pull/5531

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "nomnom": "1.8.1",
     "pbf": "2.0.1",
     "pixelworks": "1.1.0",
-    "rbush": "1.4.3",
+    "rbush": "2.0.1",
     "temp": "0.8.3",
     "vector-tile": "1.2.1",
     "walk": "2.3.9"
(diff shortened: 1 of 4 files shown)


ahocevar:
Looks good. Please merge.

tsauerwein:
Thanks for the review

bartvde:
this leads to 48 failing unit tests in Chrome and FF

bartvde:
@tsauerwein I'm going to revert this so I can do the 3.17 release today. We can do more testing on this after that I guess?

tsauerwein:
@bartvde Are you sure your local instance is updated correctly? The file `build/ol.ext/rbush.js` has changed. For me the tests pass both in FF and Chrome.

bartvde:
ah you're right, I forgot to do npm i after this step, stupid, false alarm :)

bartvde:
I did run make serve, but apparently it did not pick up the change

bartvde:
confirmed all is fine, sorry about the noise. After a fresh npm i I do see that the % in the test page does not update anymore, is the same happening for anyone else?

screen shot 2016-07-01 at 14 09 00

tsauerwein:
> After a fresh npm i I do see that the % in the test page does not update anymore, is the same happening for anyone else?

Same for me.

ahocevar:
Same for me too. It's been like this for a while.

PR #5545, opened 2016-07-01
Do not set evt.target attribute, if it's not defined yet

jachym:
ol.events.EventTarget.dispatchEvent(event) sets `event.target` attribute in the headless way - the target can be already set in ol.events.Event https://github.com/openlayers/ol3/blob/master/src/ol/events/event.js#L38 and here it is overwritten without checking - it does not make sense IMHO.

This PR tries to solve the issue and make openlayers to behave more predictable.

### DIFF
--- a/src/ol/events/eventtarget.js
+++ b/src/ol/events/eventtarget.js
@@ -75,7 +75,7 @@ ol.events.EventTarget.prototype.addEventListener = function(type, listener) {
 ol.events.EventTarget.prototype.dispatchEvent = function(event) {
   var evt = typeof event === 'string' ? new ol.events.Event(event) : event;
   var type = evt.type;
-  evt.target = this;
+  evt.target = evt.target || this;
   var listeners = this.listeners_[type];
   var propagate;
   if (listeners) {
--- a/test/spec/ol/events/eventtarget.test.js
+++ b/test/spec/ol/events/eventtarget.test.js
@@ -106,6 +106,14 @@ describe('ol.events.EventTarget', function() {
       expect(events[0].type).to.be('foo');
       expect(events[0].target).to.equal(eventTarget);
     });
+    it('passes a ol.events.Event object but with defined target', function() {
+      eventTarget.addEventListener('foo', spy1);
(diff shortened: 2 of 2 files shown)


ahocevar:
Very good point, thanks @jachym. Do you think you'd be able to add a unit test that verifies the new, more predictable behaviour?

jachym:
@ahocevar the second commit should do that

ahocevar:
Thanks @jachym. Looking at the new test, I started wondering why `ol.events.Event` has an `opt_target` at all. It is not used anywhere except for a test. And since an `ol.events.EventTarget` dispatches an `ol.Event`, I find it quite unnatural that an event can override the target.

Out of curiosity, what made you stumble across this?

jachym:
In our code, we use targets a lot and we are now upgrading from ol3 < 3.14 to 3.14 (and further) - the event model has changed, so we had to fix all our event-related code (about 400+ bugs) to openlayers event model. I'm in the middle of the process right now.

goog.events.Event preserved target attribute - ol.events.Event seemed to do the same .. at first sight.

I use the opt_target for any attribute, I need to pass to target event handler.

Not sure, whether I'm answering you question in a way, you are able to understand.

ahocevar:
Thanks for the clarification @jachym. So you were clearly abusing the `opt_target` argument, because that was encouraged by `goog.events.*`. I'd be inclined to remove that argument entirely, but I'm unsure what implications this would have on your code. One option that comes to mind is that we add `{type: string}` as allowed type for `ol.events.EventTarget#dispatchEvent`, so you could pass an arbitrary object with a `type` property.

jachym:
Well, it has zero implications to my code right now - need to fix it anyway and so, I'm going with custom classes with parent from ol.events.Event with custom attributes. Do as you need to do.

jachym:
Only implication it has for me: it does not work as it was supposed to work ;-)

ahocevar:
Ok, sorry for the trouble @jachym. I'll submit an alternative to this pull request then, removing the `opt_target` argument from the `ol.events.Event` constructor.

PR #5546, opened 2016-07-01
Remove useless target argument from ol.Event and subclass constructors

ahocevar:
This is an alternative to #5545, and simplifies the code base quite a bit.

Closes #5545.


fredj:
LGTM

PR #5548, opened 2016-07-01
Fix CHANGETYPE in ol.control.FullScreen

### LINKED ISSUE #5554: "null is not an object" if ol.js is loaded before the body tag
With version 3.17.0 I get the following error if ol.js is loaded before the body-tag in my document:
TypeError: null is not an object (evaluating 'a.webkitRequestFullscreen')
ol-debug.js:25495

I get similar errors in Safari, Chrome and Firefox.

How to reproduce:
- Go to http://openlayers.org/en/v3.17.0/examples/animation.html
- Copy the code and create a new html file just containing the code
- Open the new html file.

GaborFarkas:
Fixes error from early enumeration of a constant relying on the existence and proper instantiation of the `body` DOM element.

Closes #5553.
Fixes #5554.

### DIFF
--- a/src/ol/control/fullscreencontrol.js
+++ b/src/ol/control/fullscreencontrol.js
@@ -89,6 +89,24 @@ ol.control.FullScreen = function(opt_options) {
    */
   this.source_ = options.source;
 
+  /**
+   * @private
+   * @type {string}
+   */
+  this.changeType_ = (function() {
+    var body = document.body;
+    if (body.webkitRequestFullscreen) {
+      return 'webkitfullscreenchange';
+    } else if (body.mozRequestFullScreen) {
+      return 'mozfullscreenchange';
+    } else if (body.msRequestFullscreen) {
+      return 'MSFullscreenChange';
+    } else if (body.requestFullscreen) {
+      return 'fullscreenchange';
+    }
+    return undefined;
+  })();
+
 };
 ol.inherits(ol.control.FullScreen, ol.control.Control);
 
@@ -163,7 +181,7 @@ ol.control.FullScreen.prototype.setMap = function(map) {
   ol.control.Control.prototype.setMap.call(this, map);
   if (map) {
     this.listenerKeys.push(
-        ol.events.listen(ol.global.document, ol.control.FullScreen.CHANGETYPE,
+        ol.events.listen(ol.global.document, this.changeType_,
           this.handleFullScreenChange_, this)
     );
   }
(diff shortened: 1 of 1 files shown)


ahocevar:
Better, but still not 100% safe. When the `ol.control.FullScreen` instance is created before the document's body is available, this will still fail. Better to determine the type in `#setMap()`.

GaborFarkas:
Right. Now it is set only once (presumably), however I think this micro-optimization could be removed, if that would produce a cleaner code.

ahocevar:
The micro-optimisation is fine, but it would look nicer in its own `ol.control.FullScreen.getChangeType_` function:
```js
ol.control.FullScreen.getChangeType_ = function() {
  var changeType;
  return function() {
    if (!changeType) {
      var body = document.body;
      if (body.webkitRequestFullscreen) {
        changeType = 'webkitfullscreenchange';
      } else if (body.mozRequestFullScreen) {
        changeType = 'mozfullscreenchange';
      } else if (body.msRequestFullscreen) {
        changeType = 'MSFullscreenChange';
      } else if (body.requestFullscreen) {
        changeType = 'fullscreenchange';
      }
    }
    return changeType;
  }
};
```
That would also store the value across instances.

GaborFarkas:
Please clarify about storing the value across instances. Should we store the return value in a constant on first call?

nicholas-l:
What if we use document.exitFullscreen to get the correct event type? Could this work? it is making assumptions about prefixing though.
i.e.
```
CHANGETYPE = (function() {
  if (document.exitFullscreen) {
    return 'fullscreenchange';
  } else if (document.msExitFullscreen) {
    return 'MSFullscreenChange';
  } else if (document.mozCancelFullScreen) {
    return 'mozfullscreenchange';
  } else if (document.webkitExitFullscreen) {
    return 'webkitfullscreenchange';
  }
  return undefined;
})();
```

ahocevar:
> Please clarify about storing the value across instances. Should we store the return value in a constant on first call?

The snippet I posted above does that already. The constant is `changeType`.

GaborFarkas:
@nicholas-l, that would also work. The error is weird, `make ci` runs without an error on my side.

ahocevar:
@GaborFarkas Thanks for updating this pull request. Reading the diff, I realized that my snippet returns a function that returns a string instead of a string. Would you mind cherry-picking 96e7af9bd6d57f1432d8408dd683904a6798dba0? Then everything should be fine. Thanks

GaborFarkas:
Done, @ahocevar. Thank you for the fix.

ahocevar:
@bartvde This fixes a regression. Not sure if it warrants a bugfix release, but I imagine many users include the OpenLayers script in the`<head>` of the document, and this is broken in v3.17.0.

ahocevar:
Thanks @GaborFarkas. One minor thing: you missed the `@private` annotation for the `ol.control.FullScreen.getChangeType_` function.

GaborFarkas:
Now it should be good. Thanks for the review.

bartvde:
Right, I'll see if I can do 3.17.1 with this fix today @ahocevar thanks for handling this over the weekend guys.

PR #5550, opened 2016-07-01
Remove use of ol.vec.Mat4 calls remaining after #5482

ahocevar:
This is a leftover after merging #5507. I think I'll just merge this, because it does not really change code, just names that need to be adjusted to match the new internal API.


tschaub:
I meant to mention this on #5507 after merging #5482.  Thanks for fixing it up.

PR #5552, opened 2016-07-02
Replace goog.webgl

probins:
[`goog.webgl`](https://github.com/google/closure-library/blob/master/closure/goog/webgl/webgl.js) consists entirely of constants, mainly hex numbers. There are a large number of these, most of which OL does not use, so I have copied all those currently used into `webgl/webgl.js` and renamed to the `ol.webgl` namespace. If any others are needed in the future, they can be similarly added.


tschaub:
Nice @probins.

PR #5553, opened 2016-07-02
Moved <script> tag for ol.js into body

nikolajbaer:
In Firefox 47.0 using this example doesn't work, as the ol.js script references the document.body, and this element doesn't necessarily exist when parsing the <head> tag. Moving it to the end of the body (as is typical for scripts these days) resolves the issue and makes this example work when copy/pasted.

Also tested in Google Chrome 51


ahocevar:
#5548 will probably also fix this issue. We should make sure that including the OpenLayers script in the head works.

PR #5560, opened 2016-07-04
Get rid of goog.isArrayLike

fredj:
With this change, it's no longer possible to extend an array with
an "array like" object (an object with a 'length' property).
We were not using this feature in the library.


ahocevar:
Very nice, please merge!

PR #5563, opened 2016-07-04
Return if a vertex was removed in ol.interaction.Modify#removeVertex_

### LINKED ISSUE #5146: Suspicious return value from ol.interaction.Modify#removeVertex_
The function [returns true if the vertex was removed](https://github.com/openlayers/ol3/blob/v3.14.2/src/ol/interaction/modifyinteraction.js#L879) but [`true`](https://github.com/openlayers/ol3/blob/v3.14.2/src/ol/interaction/modifyinteraction.js#L974) is always returned.

fredj:
Fixes #5146


bartvde:
LGTM

fredj:
Thanks for the review

PR #5566, opened 2016-07-05
Fix ol.style.Image#getAnchor documentation

### LINKED ISSUE #5253: Documentation ol.style.Icon, getAnchor() wrong?
It seems to me, that the documentation for the method [getAnchor()](http://openlayers.org/en/master/apidoc/ol.style.Icon.html#getAnchor) in `ol.style.Icon` is wrong as it says:
> (...) Its units are determined by anchorXUnits and anchorYUnits.

But the [implementation](https://github.com/openlayers/ol3/blob/a81150508dd1f15ab4dd99ee1ef6f748a33a37fd/src/ol/style/iconstyle.js#L206) always returns the normalized anchor (in pixel units).

I'd be willing to provide a PR fixing this, if you agree with the above.

fredj:
fixes #5253

### DIFF
--- a/src/ol/style/imagestyle.js
+++ b/src/ol/style/imagestyle.js
@@ -109,8 +109,8 @@ ol.style.Image.prototype.getSnapToPixel = function() {
 
 
 /**
- * Get the anchor point.  The anchor determines the center point for the
- * symbolizer.  Its units are determined by `anchorXUnits` and `anchorYUnits`.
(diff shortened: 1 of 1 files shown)


ahocevar:
Looks good. Please merge.

Side note, maybe worth a separate pull request: we have no explanation of what the anchor means in the docs.

PR #5567, opened 2016-07-05
Add new olx.source.ZoomifyOptions.imageExtension option

### LINKED ISSUE #3805: Zoomify: allow file extension as a property
A lot of projects use Zoomify layers for pixel-based views of large images. Normally Zoomify tiles are JPEGs, but using PNGs would enable transparency. Right now, the .jpg extension is hard-coded into the tileUrlFunction. It would be great if this could be set as a property for ol.Source.Zoomify.

https://github.com/openlayers/ol3/blob/v3.6.0/src/ol/source/zoomifysource.js

fredj:
fixes #3805

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -6389,6 +6389,7 @@ olx.source.CartoDBOptions.prototype.account;
  *     logo: (string|olx.LogoOptions|undefined),
  *     reprojectionErrorThreshold: (number|undefined),
  *     url: !string,
+ *     imageExtension: (string|undefined),
  *     tierSizeCalculation: (string|undefined),
  *     size: ol.Size}}
  */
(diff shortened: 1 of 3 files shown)


ahocevar:
Should this really be marked as stable? As far as I know, the Zoomify spec always uses png, and a more generic solution to misusing Zoomify would be to let users provide a url template with `{TileGroup}`, `{z}`, `{x}` and `{y}` instead of an image extension.

fredj:
No, it shouldn't be marked as stable (copy/paste issue ...) and yes; an url template would be a better solution

marcjansen:
@fredj, will you continue where you left here?

fredj:
I don't have time now to continue

marcjansen:
OK

PR #5573, opened 2016-07-07
Determine ol.global in a way that works in more environments

### LINKED ISSUE #5572: Content Security Policy error (since v3.16.0, included)
In my project we have Content Security Policy setup (https://w3c.github.io/webappsec/specs/content-security-policy, https://developer.chrome.com/extensions/contentSecurityPolicy).

While trying to upgrade the OpenLayers I am getting the following error since v3.16.0 (included)

 ``Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "default-src 'self'". ``

in the following line:

 `` ol.global = Function('return this')();  ``

Note that using unsafe-eval is strongly recommended against (https://developer.chrome.com/extensions/contentSecurityPolicy#relaxing-eval)

ahocevar:
Fixes #5572.

### DIFF
--- a/externs/closure-compiler.js
+++ b/externs/closure-compiler.js
@@ -15,3 +15,5 @@
  * @see http://www.w3.org/TR/pointerevents/#the-touch-action-css-property
  */
 CSSProperties.prototype.touchAction;
+
+var global;
--- a/src/ol/ol.js
+++ b/src/ol/ol.js
@@ -275,4 +275,13 @@ ol.inherits = function(childCtor, parentCtor) {
 ol.nullFunction = function() {};
 
 
-ol.global = Function('return this')();
+/**
+ * @see https://github.com/tc39/proposal-global
+ */
+if (typeof window !== 'undefined') {
+  ol.global = window;
+} else if (typeof global !== 'undefined') {
+  ol.global = global;
+} else if (typeof self !== 'undefined') {
+  ol.global = self;
+}


marcjansen:
Is there any chance we can actually test this? I would be very happy then...

I read the SO answer, but would prefer the explanation inline.

ahocevar:
> Is there any chance we can actually test this? I would be very happy then...

Not a practicable one I guess. A pragmatic alternative approach would be to always assume a browser environment first, because the way things stand now, OpenLayers instantiation will fail anyway if `ol.global` is something different than `window`:
```js
ol.global = typeof window == 'undefined' ? Function('return this')() : window;
```
@spatialillusions, does the above work in your environment?

probins:
I think we should also be thinking ahead to ES6 modules, where `this` is undefined. Not sure how `goog.module` handles `this`.

tschaub:
The "common" way to do this would be:
```js
if (typeof window !== 'undefined') {
  ol.global = window;
} else if (typeof global !== 'undefined') {
  ol.global = global;
} else if (typeof self !== 'undefined') {
  ol.global = self;
}
``

Yes, OpenLayers currently relies on many `window` properties, but these can be provided by something like jsdom (I'm currently doing this with a universal React app where OpenLayers is loaded in Node).

probins:
see also https://github.com/tc39/proposal-global

marcjansen:
Great link, @probins. I would favor sth. as shown there for as long as needed.

ahocevar:
At first I was surprised that @tschaub did not submit an alternative pull request, but just commented with working code. Now I know why: copying and pasting @tschaub's code: 20 seconds, entertaining the Closure compiler: 10 minutes.

Pull request changed to @tschaub's solution.

marcjansen:
Good to go, @ahocevar. Please merge. And many thanks for your long breath here.

probins:
> Not sure how goog.module handles this.

To answer my own question, this has just recently been changed to be 'undefined' as with ES6 modules: https://github.com/google/closure-library/commit/ee292bc0b2216d4bf3fc819df2b2184618e975d9

So I think the merged proposal is the way forward here

PR #5575, opened 2016-07-07
Add a bit more info about the library to the README

ahocevar:
For those who come to OpenLayers through GtiHub or npmjs.com, it cannot hurt to tell a bit more about what OpenLayers is for.


fredj:
+1, thanks

PR #5576, opened 2016-07-07
 Use the word "documentation"

ahocevar:
To get better search engine rankings, we should use our new /latest/ resources wherever we can.


probins:
but it's already pointing to latest 😕

ahocevar:
Oh, I had accidently sneaked parts of this change into #5575.

ahocevar:
Updated the topic of this pull request - the change remaining is now only to use "the API documentation" instead of "API docs".

fredj:
+1

PR #5577, opened 2016-07-07
Fix olx.source.OSMOptions#url default value documentation

fredj:
See https://github.com/openlayers/ol3/blob/v3.17.1/src/ol/source/osmsource.js#L31


marcjansen:
Looks good to me, thanks and please merge.

fredj:
Thanks for the review

PR #5583, opened 2016-07-10
Fix dead link for change event in API docs

### LINKED ISSUE #5505: broken link on API document
image

at Fires - change
http://openlayers.org/en/latest/apidoc/global.html#event:change

** another 'change' button links are broken as well because their URLs are the same.

ahocevar:
Fixes #5505.


fredj:
+1, thanks

PR #5584, opened 2016-07-10
add new geometry curve line

### LINKED ISSUE #5582: curve line geometry
Hi,
I am building an indoor navigation application and I want the line string geometry corners be curved (by given radius).
I have searched for a build in solution all over the project and the internet and after I did not find one I have created new geometry "curve line".
can I send  a pull request of this new geometry?
Thanks.

itayod:
new geometry "CurveLine"
draw a regular lineString with rounded corner base on radius
example:
{
   "type": "Feature",
   "geometry": {"type": "CurveLine", "coordinates": [[3,3], [5, 15], [7, 15], [10, 8],[15, 5],[13, 2]],"radius":1}
}

notice: I was having errors when I used the function "goog.vec.Mat4.getElement" first parameter documentation to accept null: 
@param {goog.vec.Mat4.AnyType | "null"}  mat

The error occurred in /src/ol/render/canvas/canvasreplay.js line: 485,486.
when I am calculating the new radius with the new transform.

please tell me if you have better way to solve that without changing the goog.vec.Mat4.getElement first parameter.

Fixes #5582.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -42,7 +42,8 @@ ol.render.canvas.Instruction = {
   SET_FILL_STYLE: 9,
   SET_STROKE_STYLE: 10,
   SET_TEXT_STYLE: 11,
-  STROKE: 12
+  STROKE: 12,
+  MOVE_TO_LINE_TO_QUADRATIC_CURVE_TO:13
 };
 
 
@@ -468,6 +469,65 @@ ol.render.canvas.Replay.prototype.replay_ = function(
             prevY = roundY;
           }
         }
+        ++i;
+        break;
+      case ol.render.canvas.Instruction.MOVE_TO_LINE_TO_QUADRATIC_CURVE_TO:
+        goog.asserts.assert(typeof instruction[1] === 'number',
+          '2nd instruction should be a number');
+        d = /** @type {number} */ (instruction[1]);
+        goog.asserts.assert(typeof instruction[1] === 'number',
+          '3rd instruction should be a number');
+        dd = /** @type {number} */ (instruction[2]);
+        goog.asserts.assert(typeof instruction[1] === 'number',
+          '4rd instruction should be a number');
+
+        var rx = transform[0];
+        var ry = transform[1];
+
+        var radius = instruction[3] * Math.sqrt(rx * rx + ry * ry);
+
+        var roundedCornerCoords = [];
+        var p1,p2,p3;
+        var dPrev,dNext;
+        var nextPt, prevPt;
+        for (; d < dd; d += 2) {
+          dPrev = d - 2;
+          dNext = d + 2;
+          if (dPrev < 0) {
+            dPrev = dd - 2;
+          }
+          if (dNext == dd) {
+            dNext = 0;
+          }
+          p1 = [pixelCoordinates[dPrev],pixelCoordinates[dPrev + 1]];
+          p2 = [pixelCoordinates[d],pixelCoordinates[d + 1]];
+          p3 = [pixelCoordinates[dNext],pixelCoordinates[dNext + 1]];
+
+          var distPrev = Math.sqrt(Math.pow(p2[0] - p1[0], 2) + Math.pow(p2[1] - p1[1], 2));
+          var ratioPrev =   Math.max(0.5, (distPrev - radius) / distPrev);
+          prevPt = [p1[0] + (ratioPrev * (p2[0] - p1[0])), p1[1] + (ratioPrev * (p2[1] - p1[1]))];
+
+          var distNext = Math.sqrt(Math.pow(p3[0] - p2[0], 2) + Math.pow(p3[1] - p2[1], 2));
+          var ratioNext = Math.min(0.5, radius / distNext);
+          nextPt = [p2[0] + (ratioNext * (p3[0] - p2[0])), p2[1] + (ratioNext * (p3[1] - p2[1]))];
+
+
+          roundedCornerCoords.push([prevPt[0], prevPt[1], p2[0], p2[1], nextPt[0], nextPt[1]]);
+        }
+
+        var pt, len = roundedCornerCoords.length;
+        for (var j = 0; j < len; j++) {
+          pt = roundedCornerCoords[j];
+          if (j == 0) {
+            context.moveTo(pt[0], pt[1]);
+          } else {
+            context.lineTo(pt[0], pt[1]);
+          }
+          if (radius > 0 && j < len - 1 && j > 0) {
+            context.quadraticCurveTo(pt[2], pt[3], pt[4], pt[5]);
+          }
+        }
+
         ++i;
         break;
       case ol.render.canvas.Instruction.SET_FILL_STYLE:
@@ -934,7 +994,8 @@ ol.render.canvas.LineStringReplay = function(tolerance, maxExtent, resolution) {
    *         lineDash: Array.<number>,
    *         lineJoin: (string|undefined),
    *         lineWidth: (number|undefined),
-   *         miterLimit: (number|undefined)}|null}
+   *         miterLimit: (number|undefined),
+   *         cornerRadius: (number|undefined)}|null}
    */
   this.state_ = {
     currentStrokeStyle: undefined,
@@ -1054,8 +1115,14 @@ ol.render.canvas.LineStringReplay.prototype.drawLineString = function(lineString
       [ol.render.canvas.Instruction.BEGIN_PATH]);
   var flatCoordinates = lineStringGeometry.getFlatCoordinates();
   var stride = lineStringGeometry.getStride();
-  this.drawFlatCoordinates_(
-      flatCoordinates, 0, flatCoordinates.length, stride);
+  var radius = state.cornerRadius;
+  if (radius > 0) {
+    this.drawCurve_(
+      flatCoordinates, 0, flatCoordinates.length, stride,radius);
+  } else {
+    this.drawFlatCoordinates_(
+        flatCoordinates, 0, flatCoordinates.length, stride);
+  }
   this.hitDetectionInstructions.push([ol.render.canvas.Instruction.STROKE]);
   this.endGeometry(lineStringGeometry, feature);
 };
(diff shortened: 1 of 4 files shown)


ahocevar:
Without looking at the code in detail, a few quick remarks:

* In the geospatial realm, a more common name for such a geometry type is "Curve", not "CurveLine".
* The GeoJSON spec does not have a Curve geometry type. So if someone wanted to use a standards compliant transport to load curve geometries, GML/WFS would be the format of choice. Note that `ol.format.GML3` also understands the `Curve` type, but currently it is translated into a `LineString`. Same for `Surface`, which is a `Polygon` made of linear rings that are `Curve`s instead of `LineString`s.
* We have removed `goog.vec.Mat4` from the library. Use for `ol.transform` instead.

probins:
`goog.isNumber` has also been removed #5541

ahocevar:
Another remark: What you are trying to achieve is usually done with Bezier curves, and `CanvasContext2D` is capable of drawing Bezier curves.

itayod:
I have used other function calls quadraticCurveTo, I have tested in when zooming and rotating and it seems to work perfectly.
the Bezier curves did not work as expected while zooming from some reason.

itayod:
I have 2 errors in the pull request,
in the file "/src/ol/render/canvas/canvasreplay"
inside the big function ol.render.canvas.Replay.prototype.replay_
I am using the following code in order to calculate the radius every view change with the "trasnform" param of the function:

` var rx = goog.vec.Mat4.getElement(transform, 0, 0);
  var ry = goog.vec.Mat4.getElement(transform, 1, 0);`

the error I am getting is:
ERROR - actual parameter 1 of goog.vec.Mat4.getElement does not match formal parameter
ERR! compile found   : (Array<number>|null)
ERR! compile required: (Array<number>|Float32Array|Float64Array)
ERR! compile         var rx = goog.vec.Mat4.getElement(transform, 0, 0);

can you please help me with a solution for this problem?

Thanks.

ahocevar:
Again, you should not be using `goog.vec.Mat4.*`. Since the removal of `goog.vec.Mat4.*`, OpenLayers only uses the transform matrix part of a 3x3 matrix, which is a plain array of 6 elements that you can access directly. See the souce code of `ol.transform.*` in `src/ol/transform.js`.

It looks like what you are trying to implement is only the `CircleByCenterPoint` segment point, for which it might make sense to use a quadratic curve. I think that for some other types, like `LineStringSegment`, bezier curves are the proper representation. But I'm only guessing here. I haven't dug into how curves work so I cannot provide real guidance here.

itayod:
I see,
I have updated the pull request and used transform array directly as you suggested and also changed the geometry name to curve.
please let me know if there is anything else you want me to do.

Thanks.

src/ol/format/geojsonformat.js
```diff
@@ -159,6 +160,16 @@ ol.format.GeoJSON.readMultiPolygonGeometry_ = function(object) {
...
+/**
+ * @param {GeoJSONGeometry} object Object.
+ * @private
+ * @return {ol.geom.Curve} Curve.
+ */
+ol.format.GeoJSON.readGeometryCurve_ = function(object) {
+  goog.asserts.assert(object.type == 'Curve',
+    'object.type should be Curve');
+  return new ol.geom.Curve(object.coordinates,object.radius);
+};
```
tschaub: From the [GeoJSON spec](https://datatracker.ietf.org/doc/draft-ietf-geojson/?include_text=1):
> 7\.  GeoJSON Types are not Extensible
>
>   Implementations MUST NOT extend the fixed set of GeoJSON types:
>   FeatureCollection, Feature, Point, LineString, MultiPoint, Polygon,
>   MultiLineString, MultiPolygon, and GeometryCollection.

We need to decide what to do when *writing* a `ol.geom.Curve` with the GeoJSON format (e.g. throw, omit it, write a LineString?).  But we should not add a reader since GeoJSON cannot have a "Curve" type geometry.

src/ol/geom/curve.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+/**
+ * @classdesc
+ * Curve geometry.
+ *
+ * @constructor
+ * @extends {ol.geom.LineString}
+ * @param {Array.<ol.Coordinate>} coordinates Coordinates.
+ * @param {number} radius Radius.
```
tschaub: Where else do you see curves being constructed with a fixed radius per segment?

src/ol/geom/curve.js
```diff
@@ -0,0 +1,51 @@
...
+ * @param {ol.geom.GeometryLayout=} opt_layout Layout.
+ * @api
+ */
+ol.geom.Curve = function(coordinates, radius,opt_layout) {
+
+  ol.geom.LineString.call(this,coordinates,opt_layout);
+  this.radius = radius;
```
tschaub: This would be `@private` since you have a getter and setter below.

src/ol/geom/curve.js
```diff
@@ -0,0 +1,51 @@
...
+ * Curve geometry.
+ *
+ * @constructor
+ * @extends {ol.geom.LineString}
+ * @param {Array.<ol.Coordinate>} coordinates Coordinates.
+ * @param {number} radius Radius.
+ * @param {ol.geom.GeometryLayout=} opt_layout Layout.
+ * @api
+ */
+ol.geom.Curve = function(coordinates, radius,opt_layout) {
```
tschaub: Please use a single space after each comma (same below).

src/ol/geom/curve.js
```diff
@@ -0,0 +1,51 @@
...
+};
+goog.inherits(ol.geom.Curve, ol.geom.LineString);
+
+
+/**
+ * @inheritDoc
+ */
+ol.geom.Curve.prototype.getType = function() {
+
+
```
tschaub: No need for the extra whitespace above.

ahocevar:
For reference: These are the different segment types supported by the GML spec: http://www.datypic.com/sc/niem21/t-gml32_CurveSegmentArrayPropertyType.html

At first I thought your implementation here is for `CircleByCenterPoint`, but it isn't. I don't see what else it could be.

itayod:
I can send you a picture of it...
basically it's a line string geometry with curved edges.

if you don't want another geometry maybe we can add it as a style property
(radius),

itayod:
Andreas is it ok if i will consult with you about a performance problem
that I am having with my appliction?

ahocevar:
> if you don't want another geometry maybe we can add it as a style property (radius)

Sounds like a good idea to me. `cornerRadius` on `ol.style.Stroke` could make sense.

itayod:
great,
so in that case what do think should be the right design?
right now when the user define a new Curve geometry I am changing the
instructions so instead of MOVE_TO_LINE_TO I enter a new action calls
MOVE_TO_LINE_TO_QUADRATIC_CURVE_TO that basically does all the calculation
and draw the line string with rounded corners, so the drawing process is a
little different.

you can see it in the pull request in canvasreplay.js line 475.

ahocevar:
@itayod By the time you create the replay instructions for a geometry, you also know the style. So the drawing process should be pretty much the same. Or am I missing something?

itayod:
the problem is that I need to change the instructions if I have radius
style property (instead of MOVE_TO_LINE_TO use
my MOVE_TO_LINE_TO_QUADRATIC_CURVE_TO) because the move_to line_to are
calculated differently in order to create the curve.

itayod:
I have made  another pull request and added new property to the
ol.style.Stroke object instead of new geometry,
it is not completed yet, can you please review my code and let me know if
this is the way you want me to implement it so I will complete the feature.

Thanks.

ahocevar:
Thanks for your continued effort on this @itayod. I'm currently in the middle of a big pull request, but I'll review your work first thing when I'm done. Might take a few more days though.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -470,6 +471,65 @@ ol.render.canvas.Replay.prototype.replay_ = function(
         }
         ++i;
         break;
+      case ol.render.canvas.Instruction.MOVE_TO_LINE_TO_QUADRATIC_CURVE_TO:
```
ahocevar: You are doing all the math at replay time, i.e. it is repeated with every render frame. It would be better to do the math only once, upon replay creation, instead.

ahocevar:
I like this approach @itayod. I only added one comment for now: it would be better to do the math in `drawCurve` and store the results in the instruction, instead of calculating it over and over at replay time.

itayod:
Ok, great I will keep working on it next week.
בתאריך 21 ביול 2016 17:03, "Andreas Hocevar" <notifications@github.com> כתב:

marcodelpercio:
Hi,

Is this pull is almost 1 year old, is it still happening? 
I need to render a maritime route done with waypoints and an optional turn-radius for each waypoint, this change by itayod is very close to what I'm looking for with the exception that in his case the radius is assumed to be constant for every point, in my case the turn-radius can be 0 (no curve) or can be different from waypoint to waypoint.

I tried running his changes and as I wanted to carry on from there however it doesn't seem to be working as expected or maybe I'm missing something. Here's what I found:

1. I have to enter quite a large value for the turn radius, What unit is it intended? Meters for EPSG:3857?
2. The larger is the radius and the more the start of the line shifts from the actual starting point coordinates (see picture attached).
3. The line doesn't reach the final point for some reason (see picture attached).

This is the test code-snippet I'm trying to run:
`
var pointStyle = new ol.style.Style({
    image: new ol.style.Circle({
      radius: 5,
      fill: new ol.style.Fill({color: 'black'}),
      stroke: new ol.style.Stroke({
        color: 'white', width: 2
      })
    })
});

var curveStyle = new ol.style.Style({
    stroke: new ol.style.Stroke({
          color: [201, 95, 90, 1],
          width: 2,
          cornerRadius: 10400
    })
});

var featuresArray = [];

function addPoint(coords) {
    var featPoint = new ol.Feature({
        type: 'waypoint',
        geometry: (new ol.geom.Point(coords))
    });
    featPoint.setStyle(pointStyle);
    featuresArray.push(featPoint);
}
var wpA = [-5639523.95, -3501274.52];
var wpB = [-3209523.95, -1001274.52];
var wpC = [-1100523.95, -3481274.52];
var lineString = [];
lineString.push(wpA);
lineString.push(wpB);
lineString.push(wpC);
addPoint(wpA);
addPoint(wpB);
addPoint(wpC);
var featLine = new ol.Feature({
        type: 'curve',
        geometry: new ol.geom.LineString(lineString)
});
featLine.setStyle(curveStyle);
featuresArray.push(featLine);
var vectorLayer = new ol.layer.Vector({
    features: featuresArray
  })
});
var center = [-5639523.95, -3501274.52];
var map = new ol.Map({
  target: document.getElementById('map'),
  loadTilesWhileAnimating: true,
  view: new ol.View({
    center: center,
    zoom: 10,
    minZoom: 2,
    maxZoom: 19
  }),
  layers: [
    new ol.layer.Tile({
    source: new ol.source.OSM({
    	crossOrigin: null,
    })
}),
    vectorLayer
  ]
});  
`

finish-point
zommed-out
star-point
curve-waypoint

marcodelpercio:
I just wanted to add that in my modest opinion it was making more sense to have a different geometry for this. I would think that a curve is a special case of a line (i.e. curve extends line) and the radius is a geometric property, not really a style property.
It would be nice if Openlayers would allow to extend existing geometries and define how they render.

tschaub:

PR #5587, opened 2016-07-12
Lazily create KML style defaults

ahocevar:
Creating styles is something that does not need to be done when the library code is parsed first. This pull request suggests creating KML style defaults lazily when they are first needed.


tschaub:

PR #5594, opened 2016-07-13
Add renderBuffer option for ol.source.ImageVector

ahocevar:
Currently, when doing hit detection on an `ol.layer.Image` with an `ol.source.ImageVector`, the hit detection replay group will replay all features, instead of only those in proximity to the pixel of interest. `ol.layer.Vector` uses a `renderBuffer` option to limit the extent of interest. This pull request introduces the same for `ol.source.ImageVector`.

This pull request brings a huge improvement for hit detection performance when using `ol.source.ImageVector`. See #4232.

### DIFF
--- a/src/ol/source/imagevectorsource.js
+++ b/src/ol/source/imagevectorsource.js
@@ -58,6 +58,12 @@ ol.source.ImageVector = function(options) {
 
   /**
    * @private
+   * @type {number}
+   */
+  this.renderBuffer_ = options.renderBuffer == undefined ? 100 : options.renderBuffer;
+
+  /**
+   * @private
    * @type {ol.render.canvas.ReplayGroup}
    */
   this.replayGroup_ = null;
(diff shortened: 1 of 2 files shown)


bartvde:
Nice find @ahocevar LGTM please merge if Travis agrees

src/ol/source/imagevectorsource.js
```diff
@@ -58,6 +58,12 @@ ol.source.ImageVector = function(options) {
 
   /**
    * @private
+   * @type {number}
+   */
+  this.renderBuffer_ = options.renderBuffer || 100;
```
fredj: I don't know if it's a realistic use case but this prevent to have a `0` renderBuffer

fredj: `this.renderBuffer_ = options.renderBuffer !== undefined ? options.renderBuffer : 100;`

ahocevar: Thanks @fredj - a `renderBuffer` of 0 definitely makes sense for polygon rendering. Updated.

fredj:
thanks

PR #5595, opened 2016-07-13
Use tile keys instead of coord keys

### LINKED ISSUE #4455: Performance of WMTS Dimension changes

ahocevar:
This change ensures that tiles with old params or dimensions are discarded.

Fixes #4455.

### DIFF
--- a/src/ol/tile.js
+++ b/src/ol/tile.js
@@ -84,7 +84,7 @@ ol.Tile.prototype.getImage = goog.abstractMethod;
  * @return {string} Key.
  */
 ol.Tile.prototype.getKey = function() {
-  return goog.getUid(this).toString();
+  return this.key + '/' + this.tileCoord;
 };
 
 
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -937,8 +937,7 @@ ol.Map.prototype.getTilePriority = function(tile, tileSourceKey, tileCenter, til
   if (!frameState || !(tileSourceKey in frameState.wantedTiles)) {
     return ol.structs.PriorityQueue.DROP;
   }
-  var coordKey = tile.tileCoord.toString();
-  if (!frameState.wantedTiles[tileSourceKey][coordKey]) {
+  if (!frameState.wantedTiles[tileSourceKey][tile.getKey()]) {
     return ol.structs.PriorityQueue.DROP;
   }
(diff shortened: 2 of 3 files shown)


src/ol/tile.js
```diff
@@ -84,7 +84,7 @@ ol.Tile.prototype.getImage = goog.abstractMethod;
  * @return {string} Key.
  */
 ol.Tile.prototype.getKey = function() {
-  return goog.getUid(this).toString();
+  return [this.key, this.tileCoord].toString();
```
fredj: Could be:
```
return this.key + '/' + this.tileCoord;
```
to avoid to create a temporary array (untested)

fredj:
LGTM, thanks

PR #5597, opened 2016-07-14
Avoid unnecessary getImageData calls in hit detection

### LINKED ISSUE #5527: Skipped feature hit detection
At [canvasvectorlayerrenderer.js:170](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/canvas/canvasvectorlayerrenderer.js#L170), the `forEachFeatureAtPixel` method deliberately doesn't use the `frameState.skippedFeatureUids` property. Is there a particular reason why an empty object is passed in place of the skipped features for the purposes of hit detection? 

It seems like this would add hit detection overhead. It's also causing a problem where hidden features are being hit detected and highlighted on hover in our application. This can be worked around within the app, but it seems like they shouldn't even be hit detected in the first place if they are hidden.

schmidtk:
If a feature is skipped or outside the hit extent, don't execute the end geometry instruction that calls getImageData. Partially resolves #5527.

### DIFF
--- a/src/ol/render/canvas/canvasreplay.js
+++ b/src/ol/render/canvas/canvasreplay.js
@@ -270,7 +270,7 @@ ol.render.canvas.Replay.prototype.replay_ = function(
           i = /** @type {number} */ (instruction[2]);
         } else if (opt_hitExtent !== undefined && !ol.extent.intersects(
             opt_hitExtent, feature.getGeometry().getExtent())) {
-          i = /** @type {number} */ (instruction[2]);
+          i = /** @type {number} */ (instruction[2]) + 1;
         } else {
           ++i;
         }


schmidtk:
@ahocevar Please review when you have a chance. Testing in our application at the office was resulting in a dramatic performance improvement with a large (25k+) number of points.

src/ol/render/canvas/canvasreplay.js
```diff
@@ -267,10 +267,10 @@ ol.render.canvas.Replay.prototype.replay_ = function(
         if ((skipFeatures &&
             skippedFeaturesHash[goog.getUid(feature).toString()]) ||
             !feature.getGeometry()) {
-          i = /** @type {number} */ (instruction[2]);
+          i = /** @type {number} */ (instruction[2]) + 1;
```
ahocevar: I think this change might break things with the Select interaction. But the one below is fine.

ahocevar:
Thanks @schmidtk. I added one comment. I think the end instruction should be executed for skipped features, but not for features outside the hit extent. Can you confirm that?

schmidtk:
Ahh that's possible. We manage our feature selection a little differently so it won't manifest itself as an issue in our application. I can remove the change for skipped features and only change it for the hit extent case, since that's the only route impacting performance here.

schmidtk:
Squashed the commits to only change the hit extent line.

ahocevar:
Thanks a lot for this @schmidtk!

Issue #5542, opened 2016-07-01
Replacing goog.isObject


probins:
does:
```
var type = typeof val;
return type == 'object' && val != null || type == 'function';
```
Reproduce this (or something similar) as `ol.isObject`?

jonataswalker:
Maybe you want to exclude arrays:

```javascript
return (typeof val === 'object' || typeof val === 'function') && !Array.isArray(val) && val !== null;
```

probins:
this is mainly used in asserts, otherwise only used in 3 places. Create `ol.isObject` or just reproduce inline?

ahocevar:
I think the asserts are just there to satisfy the compiler. In the remaining cases, I'm wondering if it wouldn't be enough to just check for `type == 'object'`. @probins, were you planning to create a pull request?

probins:
yes. If it's just checking for `type == 'object'`, that's very straightforward. There is though the bigger question of what is planned for `goog.asserts`.

ahocevar:
For `goog.asserts`, any solution wrapped in an `ol.asserts` will do (e.g. use `console.assert`, throw an error, etc.). If we remove `goog.asserts`, many assertions will have to be replaced with type casts, because we only have them in the code to satisfy the compiler. All other assertions will work fine with any replacement.

The important thing is that assertions need to be stripped out in a build step, because we do not want to execute assertions in compiled code. Is this something that tools like browserify or whatever we might be using in the future can do?

probins:
Don't think so. These tools will remove unused code - if you only use one import from a module that has five exports, Rollup for example will leave out the others - but in this case the code isn't unused. ISTM it would be better to replace assertions with type casts where possible, as type casts are just comments which any minifier will remove.

For the moment, I'll put together a PR to replace `goog.isObject`. We can deal with asserts separately.

ahocevar:
Thinking through this again, I got reminded but `typeof null == 'object'` is true. So it is probably really better to introduce an `ol.isObject` function in `src/ol/objectutil.js`. `goog.isObject` also gives a positive for `typeof obj == 'function'`, but I don't think we need this.

ahocevar:
> For the moment, I'll put together a PR to replace goog.isObject. We can deal with asserts separately.

Sounds good @probins. I hope my https://github.com/openlayers/ol3/issues/5542#issuecomment-232892097 above helps.

probins:
[...]

PR #5605, opened 2016-07-16
Remove goog.isObject not in assertions

probins:
as discussed in #5542. Not sure the null check is needed.

### DIFF
--- a/src/ol/format/jsonfeatureformat.js
+++ b/src/ol/format/jsonfeatureformat.js
@@ -26,11 +26,11 @@ ol.inherits(ol.format.JSONFeature, ol.format.Feature);
  * @return {Object} Object.
  */
 ol.format.JSONFeature.prototype.getObject_ = function(source) {
-  if (goog.isObject(source)) {
-    return source;
-  } else if (typeof source === 'string') {
+  if (typeof source === 'string') {
     var object = JSON.parse(source);
     return object ? /** @type {Object} */ (object) : null;
+  } else if (source !== null) {
+    return source;
   } else {
     goog.asserts.fail();
     return null;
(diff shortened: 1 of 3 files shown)


src/ol/format/jsonfeatureformat.js
```diff
@@ -26,7 +26,7 @@ ol.inherits(ol.format.JSONFeature, ol.format.Feature);
  * @return {Object} Object.
  */
 ol.format.JSONFeature.prototype.getObject_ = function(source) {
-  if (goog.isObject(source)) {
+  if (source !== null && typeof source === 'object') {
```
ahocevar: It's safer here to just check `if (source !== null)`, because if it's a Node or Document, `typeof source` would probably be `'function'` and not `'object'`.

src/ol/renderer/layerrenderer.js
```diff
@@ -224,7 +224,7 @@ ol.renderer.Layer.prototype.updateLogos = function(frameState, source) {
   if (logo !== undefined) {
     if (typeof logo === 'string') {
       frameState.logos[logo] = '';
-    } else if (goog.isObject(logo)) {
+    } else if (logo !== null && typeof logo === 'object') {
```
ahocevar: Also here, a truthy check for logo should be enough.

src/ol/map.js
```diff
@@ -1486,7 +1486,7 @@ ol.Map.createOptionsInternal = function(options) {
       logos[logo] = '';
     } else if (logo instanceof HTMLElement) {
       logos[goog.getUid(logo).toString()] = logo;
-    } else if (goog.isObject(logo)) {
+    } else if (logo !== null && typeof logo === 'object') {
```
ahocevar: A truthy check for logo should suffice here as well. Then you'll also probably have to add a type cast to logo to satisfy the compiler.

src/ol/format/jsonfeatureformat.js
probins: in fact, move it down so it does the string check first

probins:
ok, changes made as suggested; looks like a type cast isn't needed. Also changed commit/PR title to clarify only replacing those not in asserts

ahocevar:
Very nice, thanks @probins.

Issue #5366, opened 2016-05-19
Feature id for MVT layer features is undefined


drnextgis:
I'm trying add select capability to MVT layer as proposed by @ahocevar [here](https://github.com/openlayers/ol3/issues/5248#issuecomment-211878312), but as I can see `feature.getId()` always returns `undefined` because `feature._id` is `undefined`. Am I missing something?

ahocevar:
The suggestion you cited above was for a VectorTile layer that used a format different than MVT (e.g. GeoJSON). We cannot parse ids in `ol.format.MVT`. The vector-tile library we use does not provide an id (see https://github.com/mapbox/vector-tile-js), except when using `VectorTileFeature#toGeoJSON()`, which we don't use for performance reasons.

If you want to have access to ids from MVT features, the first step would be to create a pull request on https://github.com/mapbox/vector-tile-js to make the currently private `_id` on `VectorTileFeature` a public `id` property.

drnextgis:
Make VectorTileFeature#id a public property pull request has landed to vector-tile-js repository.

drnextgis:
Please review PR.

drnextgis:
Fixed via https://github.com/openlayers/ol3/commit/8e31b65a87a00ae3b1bfcda12e9ab1b232e6ee4e

PR #5613, opened 2016-07-19
Parse MVT id property

drnextgis:
Fix issue #5366. It requires `"vector-tile": "1.3.0"` (PR #5612).


ahocevar:
Very slick. Thanks @drnextgis!

PR #5614, opened 2016-07-19
Replace goog.abstractMethod

probins:
With very convenient timing, the [June compiler release](https://github.com/google/closure-compiler/wiki/Releases) "Added support for `@abstract` on methods as an ES6-compatible alternative to goog.abstractMethod" - see also https://developers.google.com/closure/compiler/docs/js-for-compiler#tags

So I've replaced all `goog.abstractMethod` with empty functions and an `@abstract` annotation (plus a couple of small docs changes to keep the linter happy).

### DIFF
--- a/src/ol/format/featureformat.js
+++ b/src/ol/format/featureformat.js
@@ -28,9 +28,10 @@ ol.format.Feature = function() {
 
 
 /**
+ * @abstract
  * @return {Array.<string>} Extensions.
  */
-ol.format.Feature.prototype.getExtensions = goog.abstractMethod;
+ol.format.Feature.prototype.getExtensions = function() {};
 
 
 /**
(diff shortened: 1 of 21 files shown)


ahocevar:
I'm wondering if `ol.nullFunction` would work instead of defining an empty function every time. Maybe the type of `ol.nullFunction` would have to be changed to something like `/** @type {function(...*): *?} */`.

probins:
the compiler objects if the params in the abstract method don't match those in the implementing method, so I'm not sure that would work. What would you see as the advantage?

ahocevar:
I just played with this, and I could not find a way to make it work with an alternative approach. But I could verify that the compiler strips these definitions out completely - the build size is the same with and without this change.

ahocevar:
Thanks @probins!

Issue #5600, opened 2016-07-15
Replace goog.asserts.*


ahocevar:
See #4128.

There are several options to replace `goog.asserts.*`. After doing some research, I would like to propose the following:

* Introduce a global `DEBUG` `@define` as replacement for `goog.DEBUG`.
* Use `DEBUG && console.assert()` instead of `goog.asserts.*`.
* Replace assertions that are only there to satisfy the compiler with type casts. These can be easily identified, because the complier will issue errors once they are replaced with `DEBUG && console.assert`.

This approach will work fine in the future with all minifiers that support global defines and can remove dead code. The reason why I suggest to use `DEBUG` and not `ol.DEBUG` is because minifiers like UglifyJS2 do not support namespaces in global defines.

marcjansen:
Can we put this line

```js
DEBUG && console.assert
```

In a method `ol.assert`? Will this be removable as well?

ahocevar:
No, this won't work as nicely @marcjansen. Most minifiers will leave an empty function there. For UglifyJS2, this could be avoided by calling it with `--compress pure_funcs=[\'ol.assert\']`, but that's not a nice workaround I would say.

probins:
Do you have an idea what proportion can be replaced with typecasts?

ahocevar:
@probins I'm not sure, but definitely more than 50%.

tschaub:
Do people feel like we are getting any value from the asserts that are not about type casting?

I can see value in an assert function that was left in for production builds.  Maybe something like:

```js
ol.assert(foo, 123);
```

And if `foo` was not truthy, we would throw with a message like "Expected `false` to be truthy, see http://openlayers.org/en/v4.5.6/errors/#123 for detail.  This would require that we maintain and deploy a page with error messages.

This would make for a much nicer user experience.

ahocevar:
I'm open to suggestions like the above, as long as we do not sacrifice performance. Currently we have some assertions on the hot path, like canvas replays. Those should only remain in minified builds if they are super inexpensive to calculate.

tschaub:
I agree about not having assertions in hot paths.

ahocevar:
Ok cool, I'm working on a pull request now.

probins:
[...]

PR #5616, opened 2016-07-19
Fix event type and documentation of ol.events.condition.mouseOnly

ahocevar:
Now that we are removing `goog.asserts.*` (#5600), it makes sense to fix the documentation of `ol.events.condition.mouseOnly` in a better way than the hack from #5453.

### DIFF
--- a/doc/errors/index.md
+++ b/doc/errors/index.md
@@ -224,3 +224,7 @@ Hex color should have 3 or 6 digits.
 ### 55
 
 The `{-y}` placeholder requires a tile grid with extent.
+
+### 56
+
+`mapBrowserEvent` must originate from a pointer event.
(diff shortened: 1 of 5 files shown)


tschaub:
Does this mean `ol.MapBrowserPointerEvent` shows up in the docs?

ahocevar:
Yes, it does. And also `ol.pointer.PointerEvent`.

probins:
can this one be merged? Get rid of the final `goog.asserts`.

ahocevar:
Changed so `ol.MapBrowserPointerEvent` and `ol.pointer.PointerEvent` do not show up in the docs. @tschaub, you'll like this.

bartvde:
LGTM

PR #5619, opened 2016-07-19
Remove goog.asserts.*

ahocevar:
This pull requests replaces type check hint assertions with type casts, library sanity check assertions with conditional console.assert statements in debug mode, and runtime sanity checks with assertions that throw an ol.AssertionError with an error code for lookup outside the library.

See #5600.

There is one remaining `goog.asserts.assertInstanceof`, for which there is a separate pull request (#5616).

Fixes #5619.


src/ol/ol.js
```diff
@@ -275,3 +287,46 @@ if (typeof window !== 'undefined') {
...
+  /**
+   * @type {string}
+   */
+  this.message = 'Assertion failed. See ' +
+      (ol.VERSION ? 'http://openlayers.org/en/' + ol.VERSION.split('-')[0] : '') +
+      '/doc/errors.html#' + code + ' for details.';
+
+  /**
+   * Error code. The meaning of the code can be found on
+   * {@link http://openlayers.org/en/lates/errors.html} (replace `latest` with
```
probins: /lates/latest/

src/ol/feature.js
```diff
@@ -304,8 +301,7 @@ ol.Feature.createStyleFunction = function(obj) {
     if (Array.isArray(obj)) {
       styles = obj;
     } else {
-      goog.asserts.assertInstanceof(obj, ol.style.Style,
-          'obj should be an ol.style.Style');
+      ol.assert(obj instanceof ol.style.Style, 41);
```
probins: you might want to add the error as a comment here too

probins:
ISTM there's 2 types of runtime errors, which need to be handled differently. Most of them catch application errors; in this case, the developer should see the error when testing the app, and correct the problem. But some of them catch data errors, like problems with vector files. These may be user-supplied and beyond the developer's control, so the developer has to catch these, and inform the user. It might be worth having a different error type or process for these, but in any case developers need to consider how to handle them. I think this has to be documented and mentioned in the release notes.

ahocevar:
@probins The lines between the two are blurry, so I decided to not introduce two error types. Think of someone reading in WMC context files or some other user provided map configurations. In this case, configuration errors can occur in a context that is beyond the app developer's influence.

But I agree that this change in general needs to be mentioned in the release notes. I'm going to add that.

probins:
ok. I'd also document the console.asserts. Something like 'There are additional console assertion checks when in debug mode (when the new `ol.DEBUG` flag is `true`. As this is `true` by default, it's recommended that those creating custom builds set this to `false` so these assertions are stripped.' I'd also document this in `ol.js` where `ol.DEBUG` is defined.

ahocevar:
Good suggestion. Done.

src/ol/ol.js
```diff
@@ -297,3 +311,48 @@ if (typeof window !== 'undefined') {
...
+ * @param {number} code Error code.
+ */
+ol.AssertionError = function(code) {
+
+  /**
+   * @type {string}
+   */
+  this.message = 'Assertion failed. See ' +
+      (ol.VERSION ? 'http://openlayers.org/en/' + ol.VERSION.split('-')[0] : '') +
+      '/doc/errors.html#' + code + ' for details.';
```
tschaub: URL snob comment: how about `openlayers.org/en/latest/errors/#123` instead (adding `doc/errors/index.md` instead of `doc/errors.md`?

ahocevar: Sounds reasonable. I'll make that change. Thanks for the suggestion :smile:.

changelog/upgrade-notes.md
```diff
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+#### Changes in the way assertions are handled
+
+Previously, minified builds of the library did not have any assertions. This caused applications to fail silently or with cryptic stack traces. Starting with this release, developers get notified of many runtime errors through the new `ol.AssertionError`. This error has a `code` property. The meaning of the code can be found on http://openlayers.org/en/latest/doc/errors/. There are additional console assertion checks when in debug mode when the new `ol.DEBUG` compiler flag (previously `goog.DEBUG`) is `true`. As this is `true` by default, it is recommended that those creating custom builds set this to `false` so these assertions are stripped.'
```
tschaub: There is still a lot of `goog.DEBUG` in the library (and our build configs).  Would it make sense to switch to `ol.DEBUG` when the rest of `goog` is gone?  With this branch, I assume a build config would have to set both to `false` to remove debug code.

src/ol/geom/circle.js
```diff
@@ -223,7 +222,7 @@ ol.geom.Circle.prototype.setFlatCoordinates = function(layout, flatCoordinates)
  * @api
  */
 ol.geom.Circle.prototype.setRadius = function(radius) {
-  goog.asserts.assert(this.flatCoordinates,
+  ol.DEBUG && console.assert(this.flatCoordinates,
```
tschaub: Some of these asserts are a bit gratuitous.  If somehow we deleted `this.flatCoordinates`, calling `setRadius()` would result in this:
```
circle.js:225 Assertion failed: truthy this.flatCoordinates expectedol.geom.Circle.setRadius @ circle.js:225(anonymous function) @ VM1746:1
circle.js:227 Uncaught TypeError: Cannot read property '0' of undefined(…)
```

Here the statement immediately after the `console.assert()` throws under the same condition that the assert arg is truthy.  So, without the assert, we would get an unhandled error with a message and stack trace identifying the problem.  In this case, I don't think there is a lot of value in the assert.

In other cases, there is value.  For example, if someone calls `setCenter()` above with a center that has a different length than the original center, things break later in mysterious ways (there may be no error thrown at all).

You've done a ton of work here already, but maybe later we could go back and remove this type of assert (where an error would be thrown in the next statement (or very nearby) in cases where the assert arg is truthy).

test/spec/ol/tilerange.test.js
```diff
@@ -58,9 +58,11 @@ describe('ol.TileRange', function() {
     });
 
     describe('with mixed z', function() {
-      expect(function() {
-        return new ol.TileRange.boundingTileRange([3, 1, 3], [4, 2, 0]);
-      }).to.throwException();
+      it('returns the expected TileRange', function() {
```
tschaub: Nice catch.  This one has been around for a while (589c2a85054cffa59f45cabd7c45abf565c2c69e)!

tschaub:
Wow.  Thanks for taking this on @ahocevar.

What do you think about leaving `goog.DEBUG` until the rest is gone?

ahocevar:
@tschaub I have renamed `ol.DEBUG` to `goog.DEBUG`, and removed gratuitous asserts like the one you identified.

tschaub:
Very nice.  Merge it!

ahocevar:
Thanks for the review and the good suggestions @tschaub.

probins:
👏  removing the 'gratuitous' asserts is a considerable simplification. That's the biggest outstanding part of the Closure library. Just remains a few bits and pieces ...

probins:
👏  removing the 'gratuitous' asserts is a considerable simplification.

That's the biggest outstanding part of the Closure library. Just a few bits and pieces remain ...

probins:
this creates a whole lot of extra `goog.DEBUG`s. I'm not convinced that anybody actually uses the existing `goog.DEBUG`s, but wouldn't want to let that get in the way of removing `goog`. So the obvious way forward is to add `ol.DEBUG` as suggested here at first, change the release notes, and then simply find/replace `goog.DEBUG` with `ol.DEBUG`. I should be able to find time for this either today or over the weekend, or is this something you want to do @ahocevar. Anything else that needs to be considered?

ahocevar:
@probins Thanks for willing to take this on! Before renaming `goog.DEBUG` to `ol.DEBUG`, we should remove the few remaining Closure library bits (`goog.dom` and a few other minor ones I think).

PR #5620, opened 2016-07-20
Replace goog.getUid with ol.getUid

### LINKED ISSUE #5615: Replacing goog.getUid
There are currently 75 matches for this in the library. It simply maintains a counter for the library as a whole and then adds one for each call, storing the result in a custom property on the object concerned. It should be straightforward to reproduce this as `ol.uidCounter` and `ol.getUid`. Anyone got any better solutions, or should I put together a PR to do this?

probins:
As discussed in #5615. Fixes #5615.

A simplified version of the goog function, as I don't think there's any need to cater for 2 OLs in one page. So the uid property is simply called 'ol_uid'.

### DIFF
--- a/src/ol/control/attributioncontrol.js
+++ b/src/ol/control/attributioncontrol.js
@@ -167,14 +167,14 @@ ol.control.Attribution.prototype.getSourceAttributions = function(frameState) {
     if (!source) {
       continue;
     }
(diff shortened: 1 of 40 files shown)


ahocevar:
Very nice. Thanks @probins.

PR #5622, opened 2016-07-21
Update TileWMS.coordKeyPrefix_ on setUrls

### LINKED ISSUE #5617: UpdateParams makes the TileWMS layer flicker when called for the first time
OL 3.15
1. Create a TileWMS source without providing an URL in the options.
2. Pass the URLs with the `setUrls` method.
3. Call updateParams after the layer is displayed.
When called for the first time, `updateParams` makes the layer flicker. Probably because `coordKeyPrefix_` is set to an empty string initially. Shouldn't `setUrls` call `resetCoordKeyPrefix_`?

rrutkows:
Outdated coordKeyPrefix_ causes the TileWMS layer to flicker
on the first updateParams call.

Fixes #5617.

### DIFF
--- a/src/ol/source/tilewmssource.js
+++ b/src/ol/source/tilewmssource.js
@@ -355,6 +355,14 @@ ol.source.TileWMS.prototype.fixedTileUrlFunction = function(tileCoord, pixelRati
       pixelRatio, projection, baseParams);
 };
 
+/**
+ * @inheritDoc
+ */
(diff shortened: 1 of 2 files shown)


ahocevar:
Very nice and clean. Thanks @rrutkows. Can you please sign a [CLA](https://github.com/openlayers/cla) so we can merge this pull request?

rrutkows:
Signed.

PR #5625, opened 2016-07-21
Add interface for ol.MapBrowserPointerEvent

oterral:
All in the title. Without this we can't export the `pointerEvent` property in a custom build.


probins:
see #5616

oterral:
Thx @probins !!! I wil wait the old one to be merged.

PR #5630, opened 2016-07-22
chore(build): guard against unterminated scripts

tlvince:
If ol3 is included as part of a concatenated build and the script before it does
not end in a semi-colon, the IIFE is interpreted incorrectly (often resulting in
`TypeError: (intermediate value)`).


tschaub:
Thanks @tlvince.

PR #5632, opened 2016-07-22
Improve docs for the ol.LoadingStrategy type

ahocevar:
The docs for `ol.LoadingStrategy` are misleading, and omit the fact that a user provided function will also work.


fredj:
+1

PR #5640, opened 2016-07-25
Add example for usage of HERE Map Tile API

chrismayer:
This PR adds an example showing how to use the [HERE Map Tile API](https://developer.here.com/rest-apis/documentation/enterprise-map-tile) with an ``ol.source.XYZ``.

Some background informations:

At the FOSSGIS 2016 conference @ahocevar and me had some talks and agreed that it would be good to have HERE as another source for map tiles in OpenLayers. Therefore I contacted the German HERE partner [geomer](http://www.geomer.de/) in Heidelberg, who sponsored us an evaluation access for the HERE Map Tile API, to build an official OL example showing how to use HERE maps with OL. The key is valid until November and in the meantime geomer will contact HERE directly to convince them to give OpenLayers a long-term access to their Map Tile API for the OL example section. If this happens we could think of implementing an official ``ol.source.HERE`` as well.

TIA for any review and sharing your thoughts.

### DIFF
--- a/examples/here-maps.html
+++ b/examples/here-maps.html
@@ -1,0 +1,21 @@
+---
+layout: example.html
+title: HERE Map Tile API
+shortdesc: Example of a map with map tiles from HERE.
+docs: >
+  <p><a href="https://developer.here.com/rest-apis/documentation/enterprise-map-tile">HERE Map Tile API</a> used with <code>ol.source.XYZ</code>.</p>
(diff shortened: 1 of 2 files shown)


ahocevar:
Very nice, thanks for the effort @chrismayer. Please keep us posted about the long-term access key.

chrismayer:
FYI: On the initiative of geomer HERE extended the validity of the key used in the OL example until the end of November 2017.

PR #5641, opened 2016-07-25
De-randomize the listener map property

ahocevar:
Since it makes no sense to have two OpenLayers scripts loaded on one page, we do not need to randomize the listener map property.


fredj:
LGTM

Issue #5586, opened 2016-07-11
Add imageData of pixel to 'forEachLayerAtPixel' method callback in tile Layers.


Benjaki2:
It would be beneficial to add the imageData obj to the [callback](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/canvas/canvastilelayerrenderer.js#L208).

`return callback.call(thisArg, this.getLayer(), imageData);`

fredj:
#5670

PR #5650, opened 2016-07-26
returned imageData value in forEachlayerAtPixel tile.layer method

Benjaki2:
The dataValue is already being retrieved and could be useful in the 'forEachlayerAtPixel' callback.
Issue #5586

### DIFF
--- a/src/ol/proj/proj.js
+++ b/src/ol/proj/proj.js
@@ -678,10 +678,8 @@ ol.proj.get = function(projectionLike) {
         ol.proj.addProjection(projection);
       }
     }
-  } else {
-    projection = null;
   }
-  return projection;
+  return projection || null;
 };
 
 


src/ol/renderer/canvas/canvastilelayerrenderer.js
```diff
@@ -191,7 +191,12 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame = function(
 
 
 /**
- * @inheritDoc
+ * @param {ol.Pixel} pixel Pixel.
+ * @param {olx.FrameState} frameState Frame state.
+ * @param {function(this: S, ol.layer.Layer, Object): T} callback Layer callback.
```
tsauerwein: You could use [ol.Color](https://github.com/openlayers/ol3/blob/master/src/ol/typedefs.js#L102-L109) as type for the pixel value.

tsauerwein:
I like the idea! Could you please also update [all other definitions](https://github.com/openlayers/ol3/search?q=forEachLayerAtPixel&utf8=%E2%9C%93) of `forEachLayerAtPixel`, most importantly the one in [ol.Map](https://github.com/openlayers/ol3/blob/6d67ba539cbf6048dd2900818e350fa2da1aa4b6/src/ol/map.js#L629-L662) which is the one visible in the API docs?

Benjaki2:
@tsauerwein  These two `forEachLayerAtPixel` definitions : [renderer.webgl.vectorlayer] (https://github.com/openlayers/ol3/blob/master/src/ol/renderer/webgl/webglvectorlayerrenderer.js#L159) and [renderer.layer](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/layerrenderer.js#L63) use alternative methods to determine if the layer is present at the pixel point. Should I just return the ol.Color as undefined in these cases?

tsauerwein:
Currently `forEachLayerAtPixel` works for all layer types and renderers. It might be confusing if the color is only set for some. Though I must admit that it seems to be a bit involved to make it work for all.

ahocevar:
I think if the documentation clearly states which layer types and renderers can provide the pixel value, then a partial implementation should be fine.

fredj:
The `Array.from` function is not available in IE; see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility

PR #5652, opened 2016-07-28
Make XHR work for file:// urls

### LINKED ISSUE #5647: Some samples don't work normally in Chrome by file protocol since 3.14.2
Hi,

I have a question that since 3.14.0, some samples like earthquake-clusters.html, vector-labels.html etc don't work normally by file:// in Chrome (Version 52.0.2743.82 m in my case), even chrome opened with option  --allow-file-access-from-files. However it works by http:// in Chorme and In Firefox, these samples work by both file and http protocols.

I have to use these in nwjs (Chrome as the core) which use file protocol. Anybody have clue for solving the problem?

ahocevar:
The response status will be 0 instead of a HTTP status code for file urls, so we need to consider a status of 0 as success too.

Fixes #5647.

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -41,7 +41,8 @@ ol.featureloader.loadFeaturesXhr = function(url, format, success, failure) {
          * @private
          */
         xhr.onload = function(event) {
-          if (xhr.status >= 200 && xhr.status < 300) {
+          // status will be 0 for file:// urls
+          if (!xhr.status || xhr.status >= 200 && xhr.status < 300) {
             var type = format.getType();
             /** @type {Document|Node|Object|string|undefined} */
             var source;
--- a/src/ol/source/cartodb.js
+++ b/src/ol/source/cartodb.js
@@ -125,7 +125,8 @@ ol.source.CartoDB.prototype.initializeMap_ = function() {
  */
 ol.source.CartoDB.prototype.handleInitResponse_ = function(paramHash, event) {
   var client = /** @type {XMLHttpRequest} */ (event.target);
-  if (client.status >= 200 && client.status < 300) {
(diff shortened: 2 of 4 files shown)


marcjansen:
I'd love to have a comment explaining the negationof `statusCode` in the code, because I am sure I won't understand what is going on in 2 weeks time.

Do you agree?

ahocevar:
@marcjansen You could always do a git blame and see the associated commit comment. But I just added comments anyway because it doesn't hurt :-).

marcjansen:
I know, @ahovevar 😎

https://www.w3.org/TR/XMLHttpRequest/#the-status-attribute tells me that there are other situations when this may be `0`. Should we really have this additional check or do we then possibly silently ignore errors?

ahocevar:
Users are no longer able to fetch `file://` urls, which is a regression introduced with the removal of  `goog.xhrio`. See #5647. To fix this, we need to consider a status of `0` as success, which we also did in OpenLayers 2 and which as far as I can tell is a sane thing to do:

Quotes from https://www.w3.org/TR/XMLHttpRequest/#the-status-attribute:

> If the state is UNSENT or OPENED, return 0.

`readyState` will be >= `4` (i.e. not `0`-UNSENT or `1`-OPENED) in the `load` handler where we check the `status`.

> If the error flag is set, return 0.

As far as I can tell, the error flag (indicates some type of network error or fetch termination) won't be set, since we have reached the `onload` handler. 

> Return the HTTP status code.

Since a `file://` url is not fetched using HTTP, there will be no HTTP status code. Since the `status` is initially `0`, it will still be `0` when a `file://` resource was fetched.

marcjansen:
Great explanation! Thanks and please merge, @ahocevar.

PR #5653, opened 2016-07-28
Fix namespace in filter docs

ahocevar:
This documentation problem was mentioned in http://stackoverflow.com/questions/38606595/openlayers-3-geometric-filters-in-method-wfs-getfeature.


tsauerwein:
Thanks! :+1:

PR #5667, opened 2016-08-02
Do not return undefined from ol.proj.get

### LINKED ISSUE #5666: ol.proj.get return undefined or null?
I've just been looking at `ol.proj.get` and there seems to be an inconsistency in what it returns if the code is not in the list. Docs say 'null' if code is not in the list, but if the code is not in `projections_` and/or there is no proj4js def for that code, it will return 'undefined' not 'null'. It only returns 'null' if code is neither a projection object nor a string. ISTM it would be better if it were to return 'null' or 'undefined' in all cases.

ahocevar:
Fixes #5666.

### DIFF
--- a/src/ol/proj/proj.js
+++ b/src/ol/proj/proj.js
@@ -678,10 +678,8 @@ ol.proj.get = function(projectionLike) {
         ol.proj.addProjection(projection);
       }
     }
-  } else {
-    projection = null;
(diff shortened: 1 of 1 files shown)


fredj:
LGTM

probins:
a neat way of doing that :-)

PR #5669, opened 2016-08-02
Fix Github links to specific lines on jsdoc pages

dtreiter:
It seems that Github line links only work with a capital `L`. I _think_ this is all that has to change.


ahocevar:
Thanks @dtreiter, this looks like exactly the right fix.

ahocevar:
One more thing @dtreiter: Can you please sign a [CLA](https://github.com/openlayers/cla) and report back here when done? Thanks.

dtreiter:
Thanks @ahocevar - I think my employer, FlightAware LLC, has already signed one covering me (Daniel Reiter). Should I fill out an individual one as well?

ahocevar:
Thanks @dtreiter - I added your Github handle to the FlightAware LCC CCLA. So you're all set, no need to submit another CLA.

PR #5670, opened 2016-08-02
Added image data value to 'forEachLayerAtPixel' method callback

Benjaki2:
The update has been added to the docs for all 'forEachLayerAtPixel' prototypes
and currently returns null for non-supporting definitions. #5586 

@tsauerwein @ahocevar @fredj closed [Pull request #5650](https://github.com/openlayers/ol3/pull/5650#issuecomment-236831740) when trying to reference the issue.

@fredj `Array.from()` has been replaced by  `Array.prototype.slice.call()`

### DIFF
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -619,10 +619,11 @@ ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this, opt
  * execute a callback with each matching layer. Layers included in the
  * detection can be configured through `opt_layerFilter`.
  * @param {ol.Pixel} pixel Pixel.
- * @param {function(this: S, ol.layer.Layer): T} callback Layer
- *     callback. Will receive one argument, the {@link ol.layer.Layer layer}
- *     that contains the color pixel. To stop detection, callback functions can
- *     return a truthy value.
+ * @param {function(this: S, ol.layer.Layer, ol.Color): T} callback Layer
+ *     callback. This callback will recieve two arguments: first is the
+ *     {@link ol.layer.Layer layer}, second argument is {@link ol.Color}
+ *     and will be null for layer types that do not currently support this
+ *     argument. To stop detection callback functions can return a truthy value.
  * @param {S=} opt_this Value to use as `this` when executing `callback`.
  * @param {(function(this: U, ol.layer.Layer): boolean)=} opt_layerFilter Layer
  *     filter function. The filter function will receive one argument, the
--- a/src/ol/typedefs.js
+++ b/src/ol/typedefs.js
@@ -104,7 +104,7 @@ ol.CircleRenderOptions;
  * red, green, and blue should be integers in the range 0..255 inclusive.
  * alpha should be a float in the range 0..1 inclusive. If no alpha value is
  * given then `1` will be used.
- * @typedef {Array.<number>}
+ * @typedef {Array.<number>|Uint8Array|Uint8ClampedArray}
  */
 ol.Color;
 
--- a/src/ol/renderer/canvas/imagelayer.js
+++ b/src/ol/renderer/canvas/imagelayer.js
@@ -69,7 +69,13 @@ ol.renderer.canvas.ImageLayer.prototype.forEachFeatureAtCoordinate = function(co
 
 
 /**
- * @inheritDoc
+ * @param {ol.Pixel} pixel Pixel.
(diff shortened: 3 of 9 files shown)


fredj:
LGTM, @Benjaki2 thanks for changing the implementation.

@ahocevar , @tsauerwein any other comment?

src/ol/map.js
```diff
@@ -631,10 +631,11 @@ ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this, opt
...
  * detection can be configured through `opt_layerFilter`.
  * @param {ol.Pixel} pixel Pixel.
- * @param {function(this: S, ol.layer.Layer): T} callback Layer
- *     callback. Will receive one argument, the {@link ol.layer.Layer layer}
+ * @param {function(this: S, ol.layer.Layer, ol.Color): T} callback Layer
+ *     callback. Will receive two arguments, the {@link ol.layer.Layer layer}
  *     that contains the color pixel. To stop detection, callback functions can
- *     return a truthy value.
+ *     return a truthy value. The second argument is {@link ol.Color} and will
+ *     be null for layer types that do not currently support this argument.
```
tsauerwein: Minor, could you maybe change the sentence order?

```
 * @param {function(this: S, ol.layer.Layer, ol.Color): T} callback Layer
 *     callback. Will receive two arguments, the {@link ol.layer.Layer layer}
 *     that contains the color pixel. And the second argument is {@link ol.Color}
 *     which will be `null` for layer types that do not currently support this argument.
 *     To stop detection, callback functions can return a truthy value. 
```

tsauerwein:
That looks good from my side (just a small remark about wording), thanks for updating @Benjaki2!

ahocevar:
Question: what would be wrong about passing `imageData` directly, instead of `Array.prototype.slice.call(imageData)`? Since this event is dispatched on a hot path, it would be better to not create a new array every time.

Benjaki2:
@ahocevar good point. We would just need to edit the typeDef of ol.Color to be an `Array|Uint8ClampedArray`.

Benjaki2:
Alright. Those adjustments are in. Any other concerns?

ahocevar:
@Benjaki2 I wanted to take another look, but the branch has conflicts now due to the restructuring we've done in the meantime. Can you resolve the conflicts please?

Benjaki2:
np @ahocevar

ahocevar:
Thanks @Benjaki2. The documentation could be improved (i.e. explain that it is the color of the detected pixel, and list the supported layer types), but that can be done in a follow-up pull request.

Benjaki2:
My pleasure @ahocevar. I will work on making the documentation more explicit here. Thanks!

src/ol/typedefs.js
```diff
@@ -104,7 +104,7 @@ ol.CircleRenderOptions;
  * red, green, and blue should be integers in the range 0..255 inclusive.
  * alpha should be a float in the range 0..1 inclusive. If no alpha value is
  * given then `1` will be used.
- * @typedef {Array.<number>}
+ * @typedef {Array.<number>|Uint8Array|Uint8ClampedArray}
```
tschaub: The `ol.color.asArray()` method accepts an `ol.Color` argument and throws when called with a `Uint8Array` or a `Uint8ClampedArray`.  The type cast in that function keeps the compiler from complaining about it.  I'll add some failing tests to demonstrate the issue.

tschaub:
I think things might be getting a bit mixed up here with `ol.Color`.

```js
ol.color.isValid(new Uint8ClampedArray([255, 255, 255, 255])) // -> false
ol.color.asArray(new Uint8ClampedArray([255, 255, 255, 255])) // throws ol.AssertionError
```

Both of those functions accept `ol.Color` as an argument.  I think both results are unexpected.  If we're using Uint*Array for colors, we need to reconcile this with the current array use where the alpha value ranges from 0..1 (not 0..255).

src/ol/typedefs.js
tschaub: e8682493dfbf62a8d84bf54555debf129173223c adds a failing test.  Ticketed as #5871.

tschaub:
I don't see real benefit to overloading `ol.Color` for this functionality.  Wouldn't it be more straightforward to have the type of the new argument be `Uint8ClampedArray`?

tschaub:
I've simplified the `ol.Color` type in #5872.

PR #5675, opened 2016-08-03
Remove unused goog.provide

fredj:
ping @probins @ahocevar


ahocevar:
Should be safe to remove. Looks good to me.

fredj:
Thanks for the review

PR #5681, opened 2016-08-04
Remove extra goog.require

fredj:


ahocevar:
Good catch, please merge.

PR #5689, opened 2016-08-05
Fix cursor change in translate interaction

### LINKED ISSUE #5457: ol.interaction.Translate.handleMoveEvent_ should use layerFilter_
Affects version: 3.16.0

I had some performance issues in IE when using _ol.interaction.Translate_.

When I inspected the code I found out that _ol.interaction.Translate.handleMoveEvent__ should use layerfilter that is provided to the interaction.

I added the layerfilter to this code in _ol.interaction.Translate.handleMoveEvent__ and the performance increased significantly

Before:
```
  var intersectingFeature = event.map.forEachFeatureAtPixel(event.pixel,
      function(feature) {
        return feature;
      });
```
After:
```
  var intersectingFeature = event.map.forEachFeatureAtPixel(event.pixel,
      function(feature) {
        return feature;
      }, **this, this.layerFilter_**);
```

pgiraud:
This pull request aims to fix the following issues:

 - in webkit and non webkit browsers, the cursor change behaves differently. The cursor is "grab" when not dragging and "grabbing" when dragging on Chrome, which is correct. It's the opposite on FF.
The test on `this.lastCoordinate_` is different between the two cases.
https://github.com/openlayers/ol3/blob/master/src/ol/interaction/translateinteraction.js#L235 vs. https://github.com/openlayers/ol3/blob/master/src/ol/interaction/translateinteraction.js#L240

 - it's in my opinion not translate interaction's job to change the cursor to "pointer" for other features. It should only manage its own features and change the cursor from `grab` to `grabbing`. In other words if the translate interaction is combined with a select interaction, only the selected features should be taken into account for the cursor change.

The code gets simplified by using `featuresAtPixel_` and fixes #5457 at the same time.

Please review.

### DIFF
--- a/src/ol/interaction/translate.js
+++ b/src/ol/interaction/translate.js
@@ -215,30 +215,18 @@ ol.interaction.Translate.handleDragEvent_ = function(event) {
  */
 ol.interaction.Translate.handleMoveEvent_ = function(event) {
   var elem = event.map.getTargetElement();
-  var intersectingFeature = event.map.forEachFeatureAtPixel(event.pixel,
-      function(feature) {
-        return feature;
-      });
 
-  if (intersectingFeature) {
-    var isSelected = false;
-
-    if (this.features_ &&
-        ol.array.includes(this.features_.getArray(), intersectingFeature)) {
-      isSelected = true;
-    }
-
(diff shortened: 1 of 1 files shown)


fredj:
LGTM, thanks

Issue #5679, opened 2016-08-03
One provide per file


tschaub:
If the library were organized so that there was one `goog.provide()` per file, and the name of the provide was mapped directly to the path, we could release with a CommonJS-friendly set of modules and still support the current builds.

For example, if `src/ol/foo/bar.js` included the following:

``` js
goog.provide('ol.foo.Bar');
goog.require('ol.bam');

ol.foo.Bar = function() {
  this.baz = ol.bam.baz();
};
```

We could write a transform (e.g. for use with [`jscodeshift`](https://github.com/facebook/jscodeshift)) that rewrote this to something like `package/foo/Bar.js`:

``` js
var $ol$foo$Bar = {};
var $ol$bam = require('../bam');

$ol$foo$Bar = function() {
  this.baz = $ol$bam.baz();
};

module.exports = $ol$foo$Bar;
```

And then we could publish the content in `package` to npm, so people could do this in their applications:

``` js
var Bar = require('ol/foo/Bar'); // assuming the package 'ol' was available, which it isn't
```

This would be a lot easier without the dependency on the Closure Library (see #4128).  With that gone, a valid question is "why would we still use `goog.provide()`?".  I'm not convinced we need to keep using it, but there are build tools and other things that need to change before we remove it.  My point here is that we could have a release that had _both_ CommonJS friendly artifacts (perhaps published as a separate package) and the existing sources/tools for custom builds (and the existing gargantuan full build).

marcjansen:
Plus one for one provide per file.

probins:
the conclusion from https://github.com/openlayers/ol3/issues/5226#issuecomment-216159657 was that putting enums in their own file would lead to circular dependencies

probins:
what's the advantage of this proposal over using `goog.module`? If we're changing to standard modules, whether that's CJS or ES6, the bulk of the work is changing the scope: replacing the current global refs `ol.xxx` with imported vars. With this proposal, that would still need to be done at some point. With `goog.module` that would be done now but, once it's done, a transform from `var x = goog.require(x)` to `var x = require(x)` is a much simpler transform than what's proposed here.

gberaudo:
There are several kinds of enums, see https://github.com/openlayers/ol3/issues/5226#issuecomment-215126637. It is hard to tell how many enums will be problematic.

tschaub:
Here is a branch with a test that can get us started: https://github.com/openlayers/ol3/compare/one-provide

[...]

PR #5693, opened 2016-08-06
Reorganize files so they are named like their default export

tschaub:
This organizes files in the library so they are named like their (primary) provide.  This serves as the foundation for having one provide per file (see #5679).

This also adds a new set of tests that are run in Node.  Currently, the only new tests make assertions about their being only one provide per file and having the file named like the provide.  These tests fail, so they are not run as part of `npm test`.  You can run the tests manually with `npm run test-node`.  When these tests are passing, we can include them as part of the existing test target.


tschaub:
I've added a few more commits with patterns that I think we can follow.

In the case of `ol/color/color.js`, there is no reason for the `color` directory, so I've renamed this to `ol/color.js`.

In the case of `ol/control/controldefaults.js`, this provides `ol.control`, so the file should be named either `ol/control.js` or `ol/control/index.js`.  Since we have other provides under `ol.control.*`, I've moved this to `ol/control/index.js` (and people will be able to `require('ol/control').defaults`).

Note that this new pattern for naming files will result in multiple files with the same name (but with different paths).  E.g. there may be many `index.js` files in the library.  This requires that developers are comfortable quickly opening files based on path, not just the filename alone.

tschaub:
stahp

tschaub:
Ok, this is ready for review.

There are multiple parts of addressing #5679.  This pull request is mainly about the first part: naming files like their primary provide.  In most cases this involved renaming files.  In some cases this involved renaming provides/types.  I've documented the latter in the upgrade notes (and this will not be relevant to people who only use the `@api` symbols).

There are two tests in the new `one-provide.js` file.  With the changes here, the ["has a path that maps to the provide" test](https://github.com/openlayers/ol3/blob/3a151b8eae36c1f72521ffce2703d82dda9fbedb/test/node/one-provide.js#L50-L66) passes.

The next part of the work will be to rework files with multiple provides.  There are two alternatives for addressing this.  The first is to create new files for the provides (breaking `ol/webgl/shader.js` into three files in bdc998af5517af6b02ef37c097d5473405f16673 is an example of this).  The second is to add properties to things already provided (7d3d675c5d11c2e2468768fb0d67b4d2865975da is an example of this type of change).

When this next part of work is done, both tests in `one-provide.js` will pass and we can run it as part of the normal test target.  I'll leave that next part of work to be done elsewhere.

marcjansen:
Hats off for this awesome work, Tim.

marcjansen:
Yes, please merge. I think this is huge step in the right direction. Thanks again
!

tschaub:
Ok, rebased.  Will merge when green.

tschaub:
To make sure that other changes conform with the same, I've added the 'has a path that maps to the provide' test to our `ci` target.

So, `make test-node` runs this one test, asserting that files are named like their "default export" (in future terms).  After we get the other part working (one provide per file), we can run all of the `one-provide` tests.  All Node tests can be run with `npm run test-node`.

probins:
👍 I much prefer the new names, simpler and IMHO cleaner than the old `ol/control/attributioncontrol.js` pattern

gberaudo:
> In the case of ol/control/controldefaults.js, this provides ol.control, so the file should be named either ol/control.js or ol/control/index.js. Since we have other provides under ol.control.*, I've moved this to ol/control/index.js (and people will be able to require('ol/control').defaults).

@tschaub are you sure using `index.js` is standard? According to [this thread](http://stackoverflow.com/questions/37159533/es6-ecmascript-2015-modules-import-index-js) it is specific to node and some transpilers.

The closure [documentation](https://github.com/google/closure-compiler/wiki/JS-Modules), states that `import foo from './folder/source'` will be interpreted as "load the folder/source.js" (not "folder/source/index.js`).

Based on the two links above, it looks like we should avoid using `index.js` for now.

probins:
and I was just about to comment that this can be a bit confusing. I was just looking at `ol/has.js` which requires `ol.dom` and `ol.webgl`; `ol.dom` is in `ol/dom.js` whereas `ol.webgl` is in `ol/webgl/index.js`. There's no way of telling from the require which pattern this follows.

Also, with this pattern, shouldn't `source.jsdoc` be in `source/index.jsdoc` and `style.jsdoc` in `style/index.jsdoc`?

probins:
in ES6, the whole question of module specifiers is still to be decided. Initial browser implementations (using `<script type="module">`) follow script tag rules: absolute and relative urls only

PR #5695, opened 2016-08-07
Remove unneeded requires in renderers

probins:
a couple more requires that are only used in type comments and so not required

### DIFF
--- a/src/ol/renderer/maprenderer.js
+++ b/src/ol/renderer/maprenderer.js
@@ -9,7 +9,6 @@ goog.require('ol.events.EventType');
 goog.require('ol.extent');
 goog.require('ol.functions');
 goog.require('ol.layer.Layer');
-goog.require('ol.renderer.Layer');
 goog.require('ol.style.IconImageCache');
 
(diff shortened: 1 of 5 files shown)


src/ol/renderer/maprenderer.js
```diff
@@ -9,7 +9,6 @@ goog.require('ol.events.EventType');
 goog.require('ol.extent');
 goog.require('ol.functions');
 goog.require('ol.layer.Layer');
-goog.require('ol.renderer.Layer');
```
tschaub: We'll need to leave this one in since`ol.layer.Layer.visibleAtResolution()` is used [in this file](https://github.com/probins/ol3/blob/52afa50eec58a3a0c86442687725ab9343619afb/src/ol/renderer/maprenderer.js#L206).

tschaub: Excuse me, I was looking at the wrong `require`.

tschaub:
Looks good to me.  Thanks @probins.

PR #5696, opened 2016-08-07
Remove unneeded requires from featureloader

probins:
more requires only used in JSDoc comments. There seem to be quite a few of these; if I find time, I'll go through the library more systematically.


tschaub:
I've removed these as part of #5698.  With what is there now, you can do this to batch remove unused `goog.require()` calls:

```
eslint src --fix
```

I don't anticipate we'll need to use the `--fix` functionality going forward, but it is less tedious than going through the sources one by one :).

PR #5698, opened 2016-08-07
Custom rule to identify unused goog.require() calls

tschaub:
This sets us up for running custom ESLint rules on our sources.  The `rules` directory includes the rule sources.  This directory is installed as a local package (see our main `package.json`) so that it can be used as a plugin for ESLint.  The ESLint config in our `package.json` configures the plugin and extra rules.

The only rule I've added here is to check for unused `goog.require()` calls.  This may still need some work if it generates false positives.

**Update**: more rules added, [see below](https://github.com/openlayers/ol3/pull/5698#issuecomment-238145215).


tschaub:
Here's a sample from an editor with ESLint integration:

image

tschaub:
So I think the rule is behaving well.  There may be some cases where it causes trouble (false positives) or misses something down the line.

I've made the rule "fixable" - to make it easier to batch remove unused requires.  E.g. `eslint examples --fix` will remove all unused requires from the examples.

However, there are two cases that are causing trouble:

1. The Mapbox Vector tile examples `goog.require()` a bunch of stuff that isn't used in the examples, but is an implicit dependency of the `resources/mapbox-streets-v6-style.js` script.  It would be nice if this could be reworked.  As it is now, the rule is marking these examples as errors.

2. All our tests have the `goog.require()` calls at the end of the file.  I cannot remember what the reason for this was, but they should be moved to the top (where they would behave more like `require()` calls).  The rule is identifying these as unused because it expects `goog.require()` calls to be an ancestor of the expressions that use the required thing.

tschaub:
I've refactored the new rules.  We now have:

 * `no-duplicate-requires` - (fixable) disallow duplicate `goog.require()` calls
 * `no-unused-requires` - (fixable) disallow unused `goog.require()` calls
 * `requires-first` - require that all `goog.require()` precede other statements (except `goog.provide()`)
 * `valid-requires` - require that all `goog.require()` have a valid arg and appear at the top level

I'll look into reworking the MVT examples now - these are the only files not passing the linter.

tschaub:
Interfaces don't serve any purpose outside of the compiler.  We want to only use `goog.require()` where we have explicit dependencies.  These should map directly to `require()` calls or `import` statements in the future.

After getting rid of all the unused requires, the Compiler complained about a missing require for `ol.render.IReplayGroup`.  This is the only case in the library where we use an interface instead of a base class for type checking.  So I converted `ol.render.IReplayGroup` into a `ol.render.ReplayGroup` base class so it would be more like the rest.  And now we can `goog.require()` it explicitly where it is needed.

tschaub:
Ok, this is ready for review now.

The new linter rules should make it so we don't add any more unnecessary requires.  We can create another rule to ensure we aren't missing any requires.  I'm also inclined to convert the `test/node/one-provide.js` tests into linter rules.  But this is enough for now.

probins:
👍 I was thinking it would be good to have a programatic way to identify unused requires, but hadn't got around to working out how to do this. This looks like a good solution. Another considerable simplification of the dependency structure.

ahocevar:
Great work! Please merge.

marcjansen:
Nice work indeed! Are these rules probably helpful to other people? E.g. is it worth publishing them as standalone rules?

marcjansen:
Is it just me or do your changes only work when on node v6?

**v4**

```
$ npm run test-node
 openlayers@3.17.1 test-node /home/ol3
> mocha test/node

/home/ol3/test/node/one-provide.js:19
      let count = files.length;
      ^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at /home/ol3/node_modules/mocha/lib/mocha.js:220:27
    at Array.forEach (native)
    at Mocha.loadFiles (/home/ol3/node_modules/mocha/lib/mocha.js:217:14)
    at Mocha.run (/home/ol3/node_modules/mocha/lib/mocha.js:485:10)
    at Object.<anonymous> (/home/ol3/node_modules/mocha/bin/_mocha:405:18)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:139:18)
    at node.js:968:3
```

**v6**

```
09:56 $ npm run test-node

> openlayers@3.17.1 test-node /home/ol3
> mocha test/node



  each file
    1) has a single provide
    ✓ has a path that maps to the provide


  1 passing (91ms)
  1 failing

  1) each file has a single provide:

      AssertionError: expected a single provide: ol/control/scaleline.js
      + expected - actual

      -2
      +1
      
      at Context.it (test/node/one-provide.js:46:14)
```

… an error but an expected one.

fredj:
Thanks @tschaub !

marcjansen:
Examples don't work for me when on this branch, E.g. http://localhost:3000/build/examples/animation.html 

```
xyz.js:34 Uncaught TypeError: Cannot read property 'createXYZ' of undefined
```

marcjansen:
@tschaub: see [this possible fix](https://github.com/tschaub/ol3/compare/no-unused-requires...marcjansen:tschaub-no-unused-requires?expand=1) => 367a1b60a12967f0a35f90d2b4a53e85745810f9

probins:
re v4 vs v6, might be better to "use strict" for the moment, then it will run in both

tschaub:
@marcjansen - the `test-node` script was added as part of #5693.  This branch just adds the new linter rules.  Explicit 'strict' mode is a good suggestion though.  I'll fix that here.

tschaub:
Ok, back to "work in progress."  The dependency structure here still needs untangling.  We've got some circular dependencies and sloppy "exports" (e.g. people tacking things on to `ol.tilegrid` when the module provides `ol.tilegrid.TileGrid` - rule of thumb is that you can only add properties to what you provide, not a parent of what you provide).

tschaub:
This is once again ready for review.  Here is a synopsis of what was broken before (see https://github.com/openlayers/ol3/pull/5698#issuecomment-238175675):

 * In b05193fa454a963b33e6c5965181abde9b45b7fa, `ol.tilegrid.createXYZ()` was added to `src/ol/tilegrid/tilegrid.js` without a corresponding `goog.provide('ol.tilegrid')`.  So though our examples used it and it was exportable, there was no way to explicitly require `ol.tilegrid.createXYZ()` - but most things would get it because they had an implicit (and perhaps unnecessary) dependency on `ol.tilegrid.TileGrid`.
 * In 29990abc7eebb248de08645acd56231202bce990, unused requires were removed from examples and had been removed from tests.  At this point, the only things in the library that actually use `ol.tilegrid.TileGrid` are `ol.source.Zoomify` and `ol.tilegrid.WMTS`.  However, many things in the library depend on the `ol.tilegrid.createXYZ()` function (but have no way to require it).
 * In 789aeb1c64692eb09b482e1510751938f7bb1e3a, `ol.tilegrid.createXYZ()` was moved to `src/ol/tilegrid/index.js`, but there was a circular dependency: `src/ol/tilegrid/index.js` depends on `src/ol/tilegrid/tilegrid.js` which depends on `src/ol/tilecoord.js` which depends on `src/ol/tilegrid/index.js` (`ol.tilecoord.wrapX()` uses `ol.tilegrid.extentFromProjection()`).
 * To break the circular dependency, `ol.tilecoord.wrapX()` is renamed to `ol.tilegrid.wrapX()` in 568f0c81bb4aaedd9538a819232f03368018748c.

The examples I've tested are working now.  But there may be other situations where we were relying on a similarly fragile dependency graph.  I'm confident that this is the right way forward (given the way we have named things, it makes sense that someone would require `ol/tilegrid` to get the `createXYZ()` instead of requiring `ol/tilegrid/tilegrid`), but could use more help confirming that there aren't additional broken examples.

It also feels pretty important to create a similar linter rule to assert that we aren't missing requires.  All this is unnecessarily complicated given how sloppy `goog.provide()` and `goog.require()` allow us to be.

ahocevar:
This looks great and is definitely an improvement! Thanks for your continued effort here @tschaub. I'd say this is good to merge, and we can check for and fix additional broken examples separately.

tschaub:
Thanks for the reviews.

tsauerwein:
What about types that are only referenced in type definitions? For example in [featureloader.js](https://github.com/openlayers/ol3/blob/master/src/ol/featureloader.js#L86) `ol.VectorTile` is only used in a type definition. When doing a custom build, where you do not explicitly reference `ol.VectorTile`, it will fail because `ol.VectorTile` is not known. See also https://github.com/openlayers/workshop/issues/56#issuecomment-240144065

PR #5702, opened 2016-08-09
Add rule to identify missing requires

tschaub:
Before we can convert from `goog.require()` to `require()` (or `import`), we need to clean up our dependencies (and it turns out this is closely related to having one provide per file as in #5679).


tschaub:
650 missing requires left.  This one may do me in.

marcjansen:
You'll make it, Tim! Go for it!

image

bartvde:
@tschaub amazing work lately! Thanks a lot.

tschaub:
Ok, some assorted rules to follow (we've arrived at the same several times over the history of the project, but have not stuck with them):

 * give classes their own files (e.g. `ol.foo.Bar` goes in `src/ol/foo/bar.js`) unless you want to tack them on to an existing class (this is ugly, but `ol.Foo.Bar` could be provided by `src/ol/foo.js`)
 * functions, objects, and constants should be provided by a file named like the object of which they are a property (e.g. `ol.foo.bar()`, `ol.foo.someInstance`, and `ol.foo.SOME_CONSTANT` all go in `src/ol/foo.js`)
 * following the two rule above, don't tack things on to objects higher up the member expression tree (e.g. `src/ol/foo/bar/bam.js` could provide `ol.foo.bar.Bam` and should not tack on any properties to `ol`, `ol.foo`, or `ol.foo.bar`)

As an example, 28e801d1149393c65ae5cadb09fea818f37f81c4 reworks the changes from #5165.

 * `ol.style.IconImageCache` goes in `src/ol/style/iconimagecache.js`
 * as a convenience, `ol.style.iconImageCache` is provided by `src/ol/style/index.js`

Note that `src/ol/style/index.js` becomes a file that aggregates other things from the `src/ol/style/` directory.  Following this pattern, other things in `src/ol/style/` **should not** depend on `src/ol/style/index.js` - the dependency goes the other way around.  Things that are common dependencies of `src/ol/style/*` modules should go in more specific modules (e.g. `src/ol/style/util.js` or whatever).  Again, `src/ol/style/index.js` is a place to aggregate exports, not to export things that are required lower down that same directory tree.

tschaub:
`ol.global` is going to be troublesome.  We use things like `ol.global.document.foo`.  With otherwise dumb static analysis, we would think this would be a property exported by a `src/ol/global/document.js` or `src/ol/global/document/index.js` file.

Some alternatives:

 1. where we use it, do `var global = ol.global;` before more deeply accessing properties
 2. make the `no-missing-requires` rule know that the special `ol.global` comes from `src/ol/index.js`
 3. make the `no-missing-requires` rule not report whenever we `deeply.access.a.property` if we require *anything* up the chain (`deeply.js`, `deeply/access.js`, or `deeply/access/a.js` could make that available)

I think it is going to make it harder to switch to CommonJS / ES2015 modules if we go with the third alternative (we might not catch that someone failed to require `src/ol/foo/bar/bam.js` because their require for `src/ol/foo/index.js` covers everything else they might access under that).

I'm inclined to move things toward the first alternative before going with the second (or any other).  Other opinions welcome.

ahocevar:
Alternative 1 sounds good to me. Thanks for the hard work on this @tschaub.

tschaub:
5893d733f5fdaddf4c9e4ef5d405917cbbb7ec9d is an example that has some Alternative 1 in it.

tschaub:
The [build failure](https://travis-ci.org/openlayers/ol3/builds/151834823) here should be addressed by #5715.  I'll rebase when that is merged and add a summary of the changes in this branch.

In the meantime, feel free to review.

ahocevar:
All I can say right now is: wow! Very impressive work. Thank you so much for this effort. The new rule simply rules ;-). Please merge.

marcjansen:
Awesome work, and very impressive. Thanks for tackling this tedious work. Please merge.

tschaub:
Thanks for the reviews.  I still need to write up some detail on the rules. The no-missing-requires one has some complications.

For now, if anybody wants to work on restructuring so we have one provide and no missing requires, run ESLint directly (not via make).  The new rules are set to warn.  We can change to error when they pass.

PR #5708, opened 2016-08-10
Remove unnecessary closePath() instructions

ahocevar:
We currently use CanvasContext2D `closePath()` instructions before applying a fill or stroke to a polygon or circle, and before calling `clip()`. Since all our polygons repeat the first point as last point and use the correct winding order for outer and inner rings, and `clip()` implicitly closes an open path, we do not need those.

This change can significantly improve vector rendering performance, especially when there are many small polygons to render.


ahocevar:
Looks like there is a problem with polyfill.cdn.io which makes Travis fail currently. It appears to be unrelated to this pull request.

gberaudo:
Link to the specification: https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-clip

tschaub:
I think we'd be better off not calling `lineTo()` on the last point and only calling `closePath()`.  This makes a difference when stroking with `lineJoin`.  See https://jsfiddle.net/8c455xf6/ for example.

image

(Closed on the left, unclosed on the right.)

tschaub:
> I think we'd be better off not calling lineTo() on the last point and only calling closePath().

Deja vu: #5083.  Maybe I'm not reading your change right.  Too many late nights.

fredj:
> Looks like there is a problem with polyfill.cdn.io's fetch polyfill in PhantomJS which makes Travis fail currently. It appears to be unrelated to this pull request.

(maybe) related to https://github.com/Financial-Times/polyfill-service/issues/827

ahocevar:
Thanks for the review @tschaub. Obviously too many late nights for me as well. I have reworked the pull request to omit `closePath()` only when we have no stroke. With this, I get the same performance improvements without the negative side effects you encountered.

ahocevar:
@fredj:

> (maybe) related to Financial-Times/polyfill-service#827

Seems unrelated. The issue here must be something with `fetch()`, but I cannot reproduce it anywhere except in PhantomJS.

ahocevar:
Ticket for the fetch polyfill issue created: #5712.

src/ol/render/canvas/replay.js
```diff
@@ -1205,18 +1205,26 @@ ol.render.canvas.PolygonReplay.prototype.drawFlatCoordinatess_ = function(flatCo
...
   var i, ii;
   for (i = 0, ii = ends.length; i < ii; ++i) {
     var end = ends[i];
     var myBegin = this.coordinates.length;
-    var myEnd = this.appendFlatCoordinates(
-        flatCoordinates, offset, end, stride, true);
+    var myEnd = this.appendFlatCoordinates(flatCoordinates, offset, end, stride,
+        // Performance optimization: only close the ring when we do not have a
+        // stroke. Otherwise closePath() will take care of that.
+        !stroke);
```
tschaub: Surprised to see something specific to rings here.  This is called for linestrings as well, right?  I would think this would mean we wouldn't draw the last point in a linestring and instead would instead close it as a ring (below).

tschaub: Oh.  I guess it wouldn't make a rendering difference if we aren't stroking.  But in that case, shouldn't we bail out entirely?

ahocevar: This should only be called for polygons, because we are in a PolygonReplay's method. Or am I missing something?

ahocevar: Good suggestion about bailing out. I've added code to bail out when there is neither a fill nor stroke.

src/ol/render/canvas/replay.js
```diff
@@ -1202,31 +1202,43 @@ ol.inherits(ol.render.canvas.PolygonReplay, ol.render.canvas.Replay);
  */
 ol.render.canvas.PolygonReplay.prototype.drawFlatCoordinatess_ = function(flatCoordinates, offset, ends, stride) {
   var state = this.state_;
+  var fill = state.fillStyle !== undefined;
+  var stroke = state.strokeStyle != undefined;
+  if (!fill && !stroke) {
+    return offset;
```
marcjansen: Can you add a test for this? I fear we might need to return something else, right.

ahocevar: Thanks @marcjansen. Test added.

marcjansen:
Looking great! Please merge.

PR #5713, opened 2016-08-11
Check examples with http protocol instead of file

ahocevar:
Fixes #5712.

PR #5715, opened 2016-08-12
Create a check-example task for smarter example checking with PhantomJS

ahocevar:
This is a follow-up pull request to #5713, which avoids the problems mentioned there.

### DIFF
--- a/tasks/check-example.js
+++ b/tasks/check-example.js
@@ -1,0 +1,65 @@
+/*eslint-env es6*/
+
+const http       = require('http');
+const path       = require('path');
+const serveFiles = require('serve-files');
+const spawn      = require('child_process').spawn;
+
+if (!process.argv[2]) {
+  process.stdout.write(`USAGE: node ${path.basename(module.filename)} [example_path]\n`);
+  process.exit(0);
+}
+
+const root = process.cwd();
+const port = 8000;
+const host = null;
+const examplePath = process.argv[2];
+const phantomPath = require('phantomjs-prebuilt').path;
+
+const server = http.createServer(serveFiles.createFileResponseHandler({
+  documentRoot       : root,
+  followSymbolicLinks: false,
+  cacheTimeInSeconds : 3600
+}));
+
+server.listen(port, host, null, function() {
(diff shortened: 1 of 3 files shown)


tasks/check-example.js
```diff
@@ -0,0 +1,65 @@
+/*eslint-env es6*/
+
+const http       = require('http');
+const path       = require('path');
+const serveFiles = require('serve-files');
+const spawn      = require('child_process').spawn;
```
tschaub: Watch out, we might add a linter rule that prohibits this sort of ASCII art in the future :).

tasks/check-example.js
```diff
@@ -0,0 +1,65 @@
+/*eslint-env es6*/
```
tschaub: We could add `tasks/.eslintrc` with `{"env": {"es6": true}}` to get the same for all tasks.

tasks/check-example.js
```diff
@@ -0,0 +1,65 @@
...
+['SIGHUP', 'SIGINT', 'SIGQUIT', 'SIGILL', 'SIGTRAP', 'SIGABRT',
+ 'SIGBUS', 'SIGFPE', 'SIGUSR1', 'SIGSEGV', 'SIGUSR2', 'SIGTERM'].forEach(signal => {
+   process.once(signal, () => {
+     process.stdout.write(`Got ${signal}, stopping...\n`),
+     server.close(() => {
+       process.stdout.write('Stopped.\n');
+       process.exit(0);
+     });
+
+     Object.keys(connections).forEach(cid => connections[cid].destroy());
```
tschaub: Surprised to see that this needs to be done.  What is the result if you don't destroy these?

tschaub:
Didn't test it out locally, but this looks good to me.  Nice work tightening this up.

tschaub:
@ahocevar I'm inclined to merge this since this is the only branch with passing tests.  I'd like to do some rebasing after this is in.  Hope that is ok by you.

ahocevar:
Thanks @tschaub, that's perfectly fine. I'll create a follow-up that addresses your comments.

PR #5724, opened 2016-08-12
Fix clipping of the tile layer extent on HiDPI devices

### LINKED ISSUE #5683: performance issue in the new version
I upgraded the ol version in my application and i have noticed that it caused me huge performance problems in mobile.

I have made a small video that demonstrate the problem:
in the video we are using a small application that basically upload a very simple map,
using the old ol version (3.13) and then the new version (3.17).

the link to the video:
https://www.dropbox.com/s/lq78vz6j5vou7gm/ol_new_version_issues.mp4?dl=0

### LINKED ISSUE #5716: Rotation of the view failed on HiDPI screens
Following the issue #5489, when rotating the view of my map the result depend of the screen type I used.

Taking the example given by @ahocevar http://jsfiddle.net/otbvz824/, here is the result I got.

On my external monitor:
capture d ecran 2016-08-12 a 12 09 57

On my internal monitor (HiDPI, MacBookPro retina 13", 'more space' configuration):
capture d ecran 2016-08-12 a 12 10 12

Hope it will help.

ahocevar:
As part of the fix for layer clipping on HiDPI devices, this pull request also fixes the drawScale calculation. The drawScale has caused us to create ridiculously large canvases when rotating on HiDPI devices, so this pull request also fixes the performance problem reported in #5683.

Fixes #5716.
Fixes #5683.


fredj:
Will this fixes the ipad issues related to the maximum canvas size?

ahocevar:
@fredj It will at least improve the situation there, because the canvas size during rotation will be much smaller after this change.

bartvde:
code LGTM

ahocevar:
Thanks for the review. I'm adding some tests now to make sure we don't get more regressions like this later on.

bartvde:
LGTM thanks for adding all the tests

Issue #5601, opened 2016-07-15
'forEachLayerAtPixel' tile layer method does not account for pixelRatio of device.


Benjaki2:
The method only works if the pixelRatio of the device is 1.

The canvas [height and width ](https://github.com/openlayers/ol3/blob/master/src/ol/renderer/canvas/canvastilelayerrenderer.js#L200-L201) both need to be multiplied by the pixelRatio.

ahocevar:
@Benjaki2, would you be able to provide a pull request with a fix?

Benjaki2:
@ahocevar, sure thing.

anzeznidarsic:
Was a fix for this worked on?

ahocevar:
Fixed with #5731.

ahocevar:
Actually not fixed with #5731,

PR #5731, opened 2016-08-15
Added a pixelRatio adjust to the canvasTileLayer 'forEachLayerAtPixel…

Benjaki2:
…' method

This method was composing a frame that was only accurate for devices
with a pixelRatio of 1. Now the canvas widths and heights are
multiplied beforehand by the pixelRatio. #5601


ahocevar:
Thanks for the fix @Benjaki2!

PR #5736, opened 2016-08-16
Show `resultType` attribute in docs

tsauerwein:
`ol.format.WFS` [supports](https://github.com/openlayers/ol3/blob/d02cf1e/src/ol/format/wfs.js#L794-L796) the `resultType` option to only get the number of features when making a WFS request. But it was not shown in the docs because it was missing in the type definition.


fredj:
+1

PR #5739, opened 2016-08-17
Add task to test custom builds

tsauerwein:
Replacement for https://github.com/openlayers/ol3/pull/5738

This PR adds a task that tests custom build configurations provided in the directory `test_custom_builds`.

To make the test pass, I had to add a `goog.require` for a type that is only used in a JSDoc type definition. These were removed in https://github.com/openlayers/ol3/pull/5698, but otherwise the compiler will fail.

### DIFF
--- a/test_custom_builds/workshop.json
+++ b/test_custom_builds/workshop.json
@@ -1,0 +1,51 @@
+ {
+   "exports": [
+       "ol.Map",
+       "ol.View",
+       "ol.format.KML",
+       "ol.layer.Tile",
+       "ol.layer.Vector",
+       "ol.proj.fromLonLat",
+       "ol.source.OSM",
+       "ol.source.Vector",
+       "ol.style.Fill",
+       "ol.style.Stroke",
+       "ol.style.Style",
+       "ol.style.Text"
+   ],
+   "jvm": [],
+   "umd": true,
+   "compile": {
+     "externs": [
+       "externs/bingmaps.js",
+       "externs/closure-compiler.js",
+       "externs/esrijson.js",
+       "externs/geojson.js",
+       "externs/oli.js",
+       "externs/olx.js",
+       "externs/proj4js.js",
+       "externs/tilejson.js",
+       "externs/topojson.js"
+     ],
+     "define": [
+       "ol.ENABLE_DOM=false",
+       "ol.ENABLE_WEBGL=false",
+       "ol.ENABLE_PROJ4JS=false",
+       "ol.ENABLE_IMAGE=false",
+       "goog.DEBUG=false"
+     ],
+     "jscomp_error": [
+       "*"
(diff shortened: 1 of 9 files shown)


marcjansen:
I like it a lot, @tsauerwein. I know this again makes the CI target a little bit slower, but I guess it is worth the effort.

Please merge, and thanks again.

tsauerwein:
Thanks, Marc. I will wait with the merge because I would like to get a feedback from @tschaub about the `goog.requires/provides` that I added again.

marcjansen:
OK

probins:
this is doing something similar to #5341 - combine?

test_custom_builds/workshop.json
```diff
@@ -0,0 +1,54 @@
...
+       "externs/esrijson.js",
+       "externs/geojson.js",
+       "externs/oli.js",
+       "externs/olx.js",
+       "externs/proj4js.js",
+       "externs/tilejson.js",
+       "externs/topojson.js"
+     ],
+     "define": [
+       "goog.array.ASSUME_NATIVE_FUNCTIONS=true",
```
probins: this is no longer needed

test_custom_builds/workshop.json
```diff
@@ -0,0 +1,54 @@
...
+       "ol.ENABLE_PROJ4JS=false",
+       "ol.ENABLE_IMAGE=false",
+       "goog.DEBUG=false"
+     ],
+     "jscomp_error": [
+       "*"
+     ],
+     "jscomp_off": [
+       "useOfGoogBase",
+       "unnecessaryCasts",
```
probins: ditto

test_custom_builds/workshop.json
```diff
@@ -0,0 +1,54 @@
...
+       "ol.ENABLE_WEBGL=false",
+       "ol.ENABLE_PROJ4JS=false",
+       "ol.ENABLE_IMAGE=false",
+       "goog.DEBUG=false"
+     ],
+     "jscomp_error": [
+       "*"
+     ],
+     "jscomp_off": [
+       "useOfGoogBase",
```
probins: ditto

tsauerwein:
> this is doing something similar to #5341 - combine?

The motivation, to check that there are no regressions with custom builds, is the same. Though I would like to do it only with a small set of example configurations, so that the check can be run in CI.

marcjansen:
Thanks for comments @probins and thanks for taking them @tsauerwein :clap:

tschaub:
Hmm.  It feels like we may just want to always throw the whole library at the Compiler and never use the "manage_closure_dependencies" option.

tschaub:
Let's revisit this after transforming sources to ES modules for the 5.0 release.

PR #5741, opened 2016-08-17
Point to the English workshop

bartvde:
We should probably point directly to the English workshop on the main page


fredj:
+1

bartvde:
thanks @fredj

PR #5745, opened 2016-08-17
removed dependencies on goog.object.clone and fixed placemark rendering

tamarmot:
Per discussion here:
https://github.com/openlayers/ol3/issues/4781

I have removed dependencies on goog.object.clone.

I have also reviewed and repaired rendering of icons and text for placemarks coming from KML, focusing on providing good default values.

### DIFF
--- a/src/ol/format/kml.js
+++ b/src/ol/format/kml.js
@@ -222,6 +222,13 @@ ol.format.KML.createStyleDefaults_ = function() {
 
   /**
    * @const
+   * @type {string}
+   * @private
+   */
+  ol.format.KML.DEFAULT_NO_IMAGE_STYLE_ = 'NO_IMAGE';
+
+  /**
+   * @const
    * @type {ol.style.Stroke}
    * @private
    */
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks for picking this up @tamarmot.  I'm not able to review it now, but hope someone can help get this in soon.

src/ol/format/kml.js
```diff
@@ -1078,6 +1108,9 @@ ol.format.KML.readStyle_ = function(node, objectStack) {
   var imageStyle = /** @type {ol.style.Image} */
       ('imageStyle' in styleObject ?
           styleObject['imageStyle'] : ol.format.KML.DEFAULT_IMAGE_STYLE_);
+  if (imageStyle == ol.format.KML.DEFAULT_NO_IMAGE_STYLE_) {
```
ahocevar: Couldn't you initialize the `ol.format.ol.KML.DEFAULT_NO_IMAGE_STYLE_` with `undefined` instead? Then you wouldn't need this if-block.

tamarmot: There are acutally 3 states: One is undefined, one is explicit where the
Icon block is included but no icon is defined because the user wants no
icon to be drawn; and the third is defined.  So this DEFAULT_NO_IMAGE_STYLE
flag is to indicate explicitly no image.  We *could* have null mean one
thing and undefined mean another, but I thought that calling out NO_IMAGE
would be more clear, and it's only when the KML is being loaded that this
flag is used.

ahocevar: Thanks for the explanation. In this case I'd opt for `null` - smaller build size and the name of the constant has all information that's needed when reading the code.

tamarmot: Very good; I'll make that modification next week.

ahocevar:
Merging this now, because I'm working on removing the final Closure Library bits. Thanks @tamarmot for the excellent work.

probins:
this one should have been squashed

ahocevar:
True, but too late now.

PR #5757, opened 2016-08-22
Add ol.source.Cluster#setDistance function

### LINKED ISSUE #5521: Method to set cluster distance
It would be useful with a method to set distance for a cluster source. This could be used to set different cluster distances depending on resolution, One use case would be setting the distance to 0 at certain resolutions to prevent clustering without having to resort to create a separate layer.

fredj:
fixes #5521


ahocevar:

fredj:
Thanks for the review

PR #5763, opened 2016-08-23
Fix assert namespace

ahocevar:
This is a follow-up pull request to #5616. `ol.assert` has moved to `ol.asserts.assert`.


bartvde:
LGTM thanks

PR #5764, opened 2016-08-23
Remove unnecessary map.render() calls

bartvde:
Supersedes #5078


fredj:
Doesn't work when the kinetic effect is not active.

fredj:
Please read the comments in #5078

bartvde:
@fredj is https://github.com/fredj/ol3/commit/8c99fd596d06d67fb5fcccf1337931efe4768320 still relevant? I forgot about that one indeed.

fredj:
To reproduce the issue:
open examples/simple.html and slowly pan so that the kinetic effect doesn't occur -> the tiles are not loaded

bartvde:
thanks, confirmed it is still an issue indeed

bartvde:
@fredj can you open up a PR for your fix? If we cannot think of a better solution, we should go ahead and merge your fix I think. What do you think? Ping @ahocevar as well.

PR #5765, opened 2016-08-23
Force a map render when the view is not updated

fredj:
See https://github.com/openlayers/ol3/pull/5764#issuecomment-241665841


bartvde:
I think this change makes sense, forcing a render in a very specific case.

@ahocevar any objections from your side given your previous comments?

ahocevar:
I cannot come up with a better solution for now, so please merge.

PR #5768, opened 2016-08-23
Enable extent clipping for vector layers

### LINKED ISSUE #5761: ol.layer.Vector extent property not clipping
I asked this question on [StackOverflow](http://stackoverflow.com/questions/39021766/openlayers-vector-layer-is-not-clipped-by-extent-parameter?noredirect=1#comment65473105_39021766) and it was suggested that I open an issue.

The problem is with the behavior of the `extent` property on the ol.layer.Vector object. I created a simple example which you can see here: [codepen](http://codepen.io/mikeskaug/pen/XKOGLN) or clone from this [gist](https://gist.github.com/mikeskaug/c4cc9a97755c0945051e9d0fee5e3680).
Here is a snippet of the code:
```
      var base = new ol.layer.Vector({
          format: new ol.format.TopoJSON(),
          url: 'https://gist.githubusercontent.com/quizzicol/7068241/raw/4f6ced3a2412a48141843bee07f6ae2034eea21b/world110.json'
        }),
        extent: ol.proj.transformExtent([0, -45, 90, 45], 'EPSG:4326', 'EPSG:3857')
      });

      var map = new ol.Map({
        layers: [base],
        target: 'map',
        view: new ol.View({
          center: [0, 0],
          zoom: 1
        })
      });
```
I expected the `extent` option to limit rendering of the vector layer to the specified bounding box. However, the entire layer is still rendered. Changing the values of the extent has no affect on what is rendered in the vector layer.

ahocevar:
Fixes #5761.

### DIFF
--- a/src/ol/renderer/canvas/vectorlayer.js
+++ b/src/ol/renderer/canvas/vectorlayer.js
@@ -89,7 +89,7 @@ ol.renderer.canvas.VectorLayer.prototype.composeFrame = function(frameState, lay
   var clipExtent = layerState.extent;
   var clipped = clipExtent !== undefined;
   if (clipped) {
-    this.clip(context, frameState, clipExtent);
+    this.clip(context, frameState,  /** @type {ol.Extent} */ (clipExtent));
   }
   var replayGroup = this.replayGroup_;
   if (replayGroup && !replayGroup.isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/ol/renderer/canvas/vectorlayer.js
```diff
@@ -85,6 +85,12 @@ ol.renderer.canvas.VectorLayer.prototype.composeFrame = function(frameState, lay
 
   this.dispatchPreComposeEvent(context, frameState, transform);
 
+  // clipped rendering if layer extent is set
+  var clipExtent = layerState.extent;
```
bartvde: minor but this could be defined within the if statement

bartvde:
LGTM

PR #5769, opened 2016-08-23
Only call addToDrawing if sketchFeature

### LINKED ISSUE #5759: Finish drawing a geometry while freehand is active cause error on next pan
1. http://openlayers.org/en/latest/examples/draw-and-modify-features.html
2. Start drawing a polygon
3. Hold SHIFT to enable freehand mode
4. Double-click to finish the geometry
5. Try to pan, it throws an exception : `Cannot read property 'getGeometry' of null`

bartvde:
closes #5759


ahocevar:
Looks good, please merge.

PR #5778, opened 2016-08-25
Use Mapzen instead of openstreetmap.us tiles

### LINKED ISSUE #5473: Vector Tiles data source will soon disappear
Hi,

In approximately two months, @iandees and I will be turning off the experimental Vector Tiles service hosted at https://openstreetmap.us/~migurski/vector-datasource/. Going through our referer logs, we’ve found these openlayers.org pages using the service:

* http://openlayers.org/en/latest/examples/osm-vector-tiles.html?q=vectortile
* http://openlayers.org/en/v3.15.1/examples/osm-vector-tiles.html?q=vector
* http://openlayers.org/en/v3.0.0/examples/tile-vector.html
* http://openlayers.org/en/v3.1.0/examples/tile-vector.html
* http://openlayers.org/en/v3.5.0/examples/tile-vector.html
* http://openlayers.org/en/master/examples/osm-vector-tiles.html
* http://openlayers.org/en/v3.2.1/examples/tile-vector.html
* http://openlayers.org/en/v3.15.1/examples/osm-vector-tiles.html
* http://openlayers.org/en/latest/examples/osm-vector-tiles.html

My recommendation is to switch these pages to using Mapzen’s equivalent, supported service: https://mapzen.com/projects/vector-tiles/. We may provide HTTP 301 redirects for individual tiles after August, which may work seamlessly.

ahocevar:
The openstreetmap.us tile service will be shut down before the end of the month.

Fixes #5473.

### DIFF
--- a/examples/osm-vector-tiles.js
+++ b/examples/osm-vector-tiles.js
@@ -10,6 +10,11 @@ goog.require('ol.style.Style');
 goog.require('ol.tilegrid');
 
 
+var key = 'vector-tiles-5eJz6JX';
+
+var attribution = [new ol.Attribution({
+  html: '&copy; OpenStreetMap contributors, Who’s On First, Natural Earth, and openstreetmapdata.com'
+})];
 var format = new ol.format.TopoJSON();
 var tileGrid = ol.tilegrid.createXYZ({maxZoom: 19});
 var roadStyleCache = {};
@@ -18,7 +23,6 @@ var roadColor = {
   'minor_road': '#ccb',
   'highway': '#f39'
 };
-var landuseStyleCache = {};
 var buildingStyle = new ol.style.Style({
   fill: new ol.style.Fill({
     color: '#666',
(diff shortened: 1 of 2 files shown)


examples/osm-vector-tiles.js
```diff
@@ -10,6 +10,7 @@ goog.require('ol.style.Style');
 goog.require('ol.tilegrid');
 
 
+var key = 'vector-tiles-5eJz6JX';
```
marcjansen: Can we cloak this? Or is it overkill?

marcjansen:
Ignore that question from above. Please merge

examples/osm-vector-tiles.js
ahocevar: This will be cloaked in the example source code shown below the example, and when used in a JSFiddle.

marcjansen:
... can you please add appropriate attribution?

ahocevar:
Thanks for the review. Attribution added.

probins:
might be worth telling Mapzen about this for inclusion in their [examples/community section](https://mapzen.com/projects/vector-tiles/)

PR #5782, opened 2016-08-26
Compare URLs in an IE compatible way

ahocevar:
By using the URL constructor and regular expressions in tests, we can make sure that URL comparisons work on all browsers.


ahocevar:
Merging as trivial change that does not make changes to the library.

PR #5790, opened 2016-08-27
Fix vector tile rotation on HiDPI devices

### LINKED ISSUE #5791: Regression: Rotation of vector tile layers broken
This regression only affects HiDPI devices. Go to http://openlayers.org/en/v3.18.0/examples/mapbox-vector-tiles.html and rotate the map. Everything that is rendered to an image (hybrid modes, so lines and polygons) will be shifted to somewhere outside the map viewport. The same works fine in http://openlayers.org/en/v3.18.0/examples/mapbox-vector-tiles.html.

ahocevar:
This is a regression from 3.17.1, introduced by #5724: the tile images generated in hybrid and image mode disappear when rotating the map on retina displays, as can be seen in the mapbox-vector-tiles example.

Fixes #5791.


bartvde:
@ahocevar the issue is not only on retina displays, I have it on my non-retina screen as well

ahocevar:
Good to know @bartvde. Does this pull request fix things there as well?

bartvde:
Will look at it soon and report back here.

bartvde:
Yes fixes it as well on non-retina

bartvde:
LGTM thanks for adding more rendering tests!

PR #5793, opened 2016-08-28
Closure bye bye

### LINKED ISSUE #4128: Remove dependency on the Closure Library

ahocevar:
This pull request removes the remaining bits of Closure Library, and renames `goog.DEBUG` to `ol.DEBUG`.

Fixes #4128.

### DIFF
--- a/src/ol/style/stroke.js
+++ b/src/ol/style/stroke.js
@@ -1,7 +1,5 @@
 goog.provide('ol.style.Stroke');
 
-goog.require('goog.crypt');
-goog.require('goog.crypt.Md5');
 goog.require('ol.color');
 
 
@@ -207,7 +205,7 @@ ol.style.Stroke.prototype.setWidth = function(width) {
  */
 ol.style.Stroke.prototype.getChecksum = function() {
   if (this.checksum_ === undefined) {
-    var raw = 's' +
+    this.checksum_ = 's' +
         (this.color_ ?
             ol.color.asString(this.color_) : '-') + ',' +
         (this.lineCap_ !== undefined ?
@@ -220,10 +218,6 @@ ol.style.Stroke.prototype.getChecksum = function() {
             this.miterLimit_.toString() : '-') + ',' +
         (this.width_ !== undefined ?
             this.width_.toString() : '-');
-
-    var md5 = new goog.crypt.Md5();
-    md5.update(raw);
-    this.checksum_ = goog.crypt.byteArrayToString(md5.digest());
   }
 
   return this.checksum_;
--- a/test/spec/ol/dom/dom.test.js
+++ b/test/spec/ol/dom/dom.test.js
@@ -1,7 +1,6 @@
 /*global Modernizr*/
 goog.provide('ol.test.dom');
 
-goog.require('goog.userAgent');
 goog.require('ol.transform');
 goog.require('ol.dom');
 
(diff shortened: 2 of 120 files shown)


probins:
👏

probins:
there's still `goog.dom` in kml test

ahocevar:
Oh, I only looked in `src/` to get the build right. Would you be willing to provide a cherry-pick commit @probins?

probins:
yes, it's only a `has.IE` afaics

probins:
@ahocevar take a look at https://github.com/openlayers/ol3/blob/master/test/spec/ol/dom/dom.test.js#L117. Replace this with one test for `ol.has.IE`?

ahocevar:
Thanks for the pointer @probins, that was easy. Done with baaeda0.

probins:
👍

probins:
`goog.DEBUG` is also used in `tasks/readme.md`, `config/examples-all.json`, `doc/tutorials/closure.md`, `doc/tutorials/custom-builds.md` - I think these should be added to https://github.com/openlayers/ol3/pull/5793/commits/76a6117078e938193b61fe2d02839a3f3f585e3a

ahocevar:
> goog.DEBUG is also used in tasks/readme.md, config/examples-all.json, doc/tutorials/closure.md, doc/tutorials/custom-builds.md - I think these should be added to 76a6117

Done.

test/spec/ol/dom/dom.test.js
```diff
@@ -116,34 +115,29 @@ describe('ol.dom', function() {
 
     it('sets transform origin for IE 9', function() {
       // save old user agent information
-      var originalIsIE = goog.userAgent.IE;
-      var originalIsVersionOrHigher = goog.userAgent.isVersionOrHigher;
+      var originalIsIE = ol.has.IE;
```
tschaub: No need to change it here since this is the way things were, but it would be more robust to cache the original value in a `beforeEach` and restore it in `afterEach`.  The way things are here, if there is an unhandled error, the value isn't restored.

src/ol/style/stroke.js
```diff
@@ -220,10 +218,6 @@ ol.style.Stroke.prototype.getChecksum = function() {
             this.miterLimit_.toString() : '-') + ',' +
         (this.width_ !== undefined ?
             this.width_.toString() : '-');
-
-    var md5 = new goog.crypt.Md5();
-    md5.update(raw);
-    this.checksum_ = goog.crypt.byteArrayToString(md5.digest());
```
tschaub: Crazy.  Would be interesting to look at the build size before and after this commit alone.

fredj: Added in 509fbaee1c7cee9dc04dcd576085d7458be38981 (ping @tsauerwein )

tschaub: 6216 Bytes removed in this commit.  So about 1/20th of Leaflet.

tsauerwein: :+1:

tschaub:
Nice work @ahocevar.  Please merge.

ahocevar:
Thanks for the reviews. I addressed @tschaub's comment about `ol.dom` tests. Will merge when green.

tschaub:
Looking green!

tschaub:
Hope it's ok by you if I merge this @ahocevar.  There are a couple other branches I'd like to base on this.

ahocevar:
Cool, thanks for merging @tschaub.

Issue #5532, opened 2016-06-30
Not stopping pending tile requests


ardean80:
I'm using ol.layer.Tile with ol.source.XYZ to load tiles from a specific tile server. 
Panning and zooming the map I noticed that the old pending tile requests (for example, the ones made to load a previous zoom level) are not automatically aborted, they keep going until they get a response.

ahocevar:
This is correct, and there is currently no easy way to change this. The reason why we didn't consider this case was because there is no good way to abort a pending requests for an Image element's `src`. You would have to load all images using `fetch` or `XMLHttpRequest`.

Do you have an application or example where the current behaviour is problematic? As long as you have not set `loadTilesWhileAnimating` and `loadTilesWhileInteracting` to `true` on the map, there should not really be many useless tile requests.

ardean80:
When a user quickly (but not continuously, so that loadTilesWhileAnimating and loadTilesWhileInteracting wound not have effect) switch from a zoom level to another one, keeping alive old requests could slow down some (already slow) tile servers. I know that some important work should be done to improve server performances, but may also be useful to limit useless requests from the client.

ahocevar:
We'd very gladly accept pull requests to improve this, if someone has an idea for a fix that does not complicate the code too much.

fredj:
According to http://stackoverflow.com/a/5278475, setting `src` to an empty string abort the request

ahocevar:
Actually that is not correct. Setting `src`to an empty string means to set `window.location.href` as `src`:
```
> var img = document.createElement('IMG');
< undefined
> img.src=""
< "" = $1
> img.src
< "http://stackoverflow.com/questions/5278304/how-to-cancel-an-image-from-loading/5278475" = $2
```

tsauerwein:
> Actually that is not correct. Setting `src` to an empty string means to set window.location.href as src:

But there will be no new request made to `window.location.href`! This seemed to be the case, but no longer:  https://www.nczonline.net/blog/2009/11/30/empty-image-src-can-destroy-your-site/

fredj:
I've tried with:
```
var img = document.createElement("IMG");
img.src = "http://imgsrc.hubblesite.org/hu/db/images/hs-2015-02-a-hires_jpg.jpg";
// wait
img.src = "";
```
(source https://jsfiddle.net/nw34gLgt/)
And the image loading was stopped (but yes, `img.src` was wrong after that (equal to `window.location.href`))

[...]

PR #5795, opened 2016-08-28
Render older loaded tiles while waiting for new tiles

thomasmoelhave:
This patch improves the perceived performance of TileImage (e.g. WMTS) tile transitions. This patch is based on the discussion in #4455 and uses an idea from #3660 that did not make the cut for #4389. The problem it solves can be illustrated by the following timeline:

1. Layer initially has dimension `x` set to 0, and everything is rendered.
2. User sets `x=10` and a bunch of tile requests are issued (e.g. a number of tiles are in the LOADING state)
3. User sets `x=20`, and a new bunch of tile requests are issued. The requests for `x=10` are *not* aborted by ol.
4. The tiles for `x=10` start arriving from the WMTS server, however since the current key is now `x=20`, these tiles are discarded.
5. User sets `x=30`, and a new bunch of tile requests are issued. Any pending requests for `x={10,20}`  are *not* aborted by ol.
6. The tiles for `x=20` and straddlers from `x=10` start arriving from the WMTS server, however since the current key is now `x=30`, these tiles are also discarded.
7. The user stops updating `x`, and once the `x=30` tiles arrive from the server they are rendered on the screen. Until that point, the map still showed the `x=0` tiles and this gives a bad user experience.

Any rendering of the map occuring in steps 2 through 6 would use `x=0` tiles through the current interim tile mechanism. This ensures that the map does not flicker - but instead the map remains static. At step 7, we're back to an up-to-date view. If 3 and 5 are triggered by something like a slider (e.g. as in the[ WMTS transition example](http://openlayers.org/en/latest/examples/wmts-dimensions.html)), it can take several seconds for step 7 to occur if the user repeatedly updates the dimension beyond `x=30`.  Although multiple newer tiles have been requested and received from the server, those are not used.

This patch changes this behavior by displaying most recently loaded tile for a given tile coordinate. It should not change anything about what tile requests are made and how they are scheduled, it simply tries to use as many of the received tiles as possible. 

To see the effect of this patch, visit the [WMTS transition example]((http://openlayers.org/en/latest/examples/wmts-dimensions.html)), and drag the slider at a relatively quick pace (if you have a very low-latency collection to the SCALGO tile server, the effect is less noticable). 

### Technical Details

The patch works by making interimTiles into a linked list of tiles, ordered by update time. When rendering a tile position where the tile corresponding to the current dimension value is not in the LOADED state, the list is traversed and the most recent LOADED tile is used instead. The list is repeatedly garbage collected such that no superfloues tiles are kept around. In practice the list has length < 3. Under heavy parameter changing, the state of the list is usually:

1. List head: The currently wanted tile, state IDLE until scheduled. 
2. head+1: tiles in the LOADING state for this position (mostly only one of these is present).
3. head+2: The most recent loaded tile for this position. [anything after this position is no longer relevant and are garbage collected]

Since tiles can arrive out of order from the tile, the time a tile has been loaded is not the deciding factor for determining the most recent loaded tile. Instead, we introduce a global sequence counter that is incremented when a new tile is created. When comparing to loaded tiles for the same tile coordinate, the tile with the highest sequence number is the most recent. 

The code is ready for review. In particular I am unsure about where to put the global sequence counter - and I am sure you will have other suggestions as well. The bulk of the code is in two new functions that manages the linked list of interim tiles, plus a sprinkling of changes that integrates the list into the renderer and introduces the sequence numbers into the tile class.


src/ol/tile.js
```diff
@@ -4,6 +4,7 @@ goog.require('ol');
 goog.require('ol.events.EventTarget');
 goog.require('ol.events.EventType');
 
+var lastSequenceNumber = 0;
```
thomasmoelhave: This should probably not go here,  probably a static member is more appropriate.

ahocevar: I think it's ok to do this like you did.

ahocevar: Oh wait, will this become a global variable in the concatenated debug build?

ahocevar:
Thanks for the effort you put into this @thomasmoelhave. Looking through the code, I started to wonder what would be the most efficient way of dealing with sources that frequently change would be.

Your pull request optimises for tile sources that provide equally valid dimensions for the user to choose from. In such a case, it makes sense to not abort pending tile requests but finish them and add them to the cache in case the user wants to see these tiles later on.

Other sources may provide realtime data, so older version tiles won't be needed any more and pending requests should be immediately aborted when a new version is requested.

If we found a good way to abort pending requests so we could optimise for the realtime case, I wonder if the dimensions case would also benefit from such a change. It would probably change the user experience in a way that the dimension the user wants to see is faster available, because the tile queue is not saturated with requests for dimensions the user does not want to see right now.

That said, I think we should keep this open for now and see if @fredj or @tsauerwein can come up with a good solution to abort tile requests (#5532). Or maybe you want to get involved in that as well @thomasmoelhave? #5805 is maybe also related, but that can be addressed later. Once we have a way to abort tile request, let's revisit the perceived user experience of the WMTS dimensions example.

src/ol/tile.js
thomasmoelhave: Yes, I think it will. That's why it should probably be a static (private) variable - I simply forgot to change that before submitting the pull request.

thomasmoelhave:
Hi @ahocevar,

Thank you for your comments.  I think there's two relatively independent considerations:

1. Does this pr improve on the current situation? I think the behavior implemented in this pr is superior to the current behavior in general. Right now,  ol still renders outdated tiles (to avoid flickering, that was the point of the interim tiles introduced in #4389), this patch simply makes sure that we use the most recent outdated tiles. In the current behavior we're displaying the most recent tile that happend to finish loading for a particular tilecoord before the dimension was changed again. This will still correspond to a different dimension value for different tilecoords on the screen in many cases. Thus,  I can not think of a use case where the current behavior is better (but maybe I am missing something?).

2. Is aborting always desirable in this case?  I think you are right in your observation that for certain use cases, you would want to abort tiles. However for the use case illustrated by the WMTS transitions example, I think the behavior exhibited by this patch is superior:
  1. First of all, ol is good at not having multiple tiles for the same tilecoord in the LOADING state    simultaneously, thus we're only really waiting for one tile to finish before the next is loaded.
  2. Responsiveness is very important to a good user experience, for an optimized WMTS server, the gain from aborting a request that already in-flight is fairly low. In my case, the average request time for a single WMTS tile from the sea-level server in the example is only on the order of 30-40ms. Thus, if we assume that we only have one "outdated" LOADING tile for a given tilecoord, then we're only delaying the arrival of the final tiles by that amount once the user has stopped changing the slider. I think having a portion of the map respond to intermediate settings is preferable to having a static map for the duration of the dimension update interaction

  However, as you said, this reasoning may be less valid for slower servers (be it WMS/WMTS or something else) or for other types of data (real-time/categorized/??) - in which case aborting might make sense.  However, the behavior enabled by this patch is no worse than the current behavior in that respect.

Due to (1), I think this pr (or a different implementation of the same type of feature) warrants inclusion, in particular because it would not be immediately outdated if ol also starts to support aborting tile request (as argued in (2)).  I guess it's up to you guys to decide if you want to go forward with this :).

To address your  you asked in #5532, the the list maintained in this pr should make it fairly simply to find the tiles that are LOADING and are also outdated due to a dimension change. Currently these are allowed to continue (if they are more recent than the most recent LOADED tile) or are simply forgotten/gc'ed (if they're older than the most recent LOADED tile). If we had a way to abort a request (e.g. one of the methods discussed in #5532), then it would be relatively simple to invoke that method on the correct tiles here. Thus there may be a simple way forward based on the work in this pr where we can choose to abort those tile requests that are outdated because of a change in dimension. It would not address tile request that are outdated b/c the user changed the view settings (zoom/location), that would need to be handled some other way I think - but I could instrument this patch to abort updated tiles if provided with a method to do so.

ahocevar:
My major concern was the addition of complexity, but a first glance at the code reveals some things that can be simplified. So let's move on with this. I'll be adding a couple of review comments now.

Review comments aside, we'll need unit tests for the added functionality.

src/ol/tile.js
ahocevar: You can get rid of the sequence number and use `ol.getUid()` instead. Maybe add a comment there that the generated uids are in ascending order. In the `ol.Tile` constructor, simply call `ol.getUid(this);`.

src/ol/source/tile.js
```diff
@@ -204,6 +204,19 @@ ol.source.Tile.prototype.getTile = function(z, x, y, pixelRatio, projection) {};
 
 
 /**
+ * Get the interim tile most suitable for rendering.
+ * @param {!ol.Tile} tile Most recent tile (head of the interim tile list)
+ * @return {!ol.Tile} Best tile for rendering at this location.
+ *
+ */
+ol.source.Tile.prototype.getInterimTile = function(tile) {
```
ahocevar: This should not be necessary. Instead, the tiles consumed by `ol.source.TileImage` should be typed `ol.ImageTile` instead of `ol.Tile`.

ahocevar: Even better, the method could be added to `ol.ImageTile`.

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
 };
 
+/**
+ * Get the interim tile most suitable for rendering using the chain of interim tiles.
+ * This corresponds to the loaded tile with the highest sequence, if no such
+ * tile exists, the original tile is returned.
+ * @param {ol.Tile} tile Most recent tile (head of the interim tile list)
+ * @return {!ol.Tile} Best tile for rendering at this location.
+ */
+ol.source.TileImage.prototype.getInterimTile = function(tile) {
```
ahocevar: Can be moved to `ol.ImageTile`

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
+};
+
+/**
+ * Goes through the chain of interim tiles starting at the provided tile and discards
+ * sections of the chain that are no longer relevant.
+ * @param {ol.Tile} tile Head of chain of interim tiles.
+ * @return {void}
+ * @private
+ */
+ol.source.TileImage.prototype.refreshInterimChain = function(tile) {
```
ahocevar: Can be moved to `ol.ImageTile`

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
+  }
+
+  tile = head.interimTile;
+  var prev = head;
+
+  var discard = function(prev, tile) {
+    prev.interimTile = tile.interimTile;
+    return prev;
+  };
+  var step = function(prev,tile) {
```
ahocevar: Not needed, see below.

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
+      //we have a loaded tile, we can discard the rest of the list
+      //TODO: in an ideal world we would could abort any LOADING tile request
+      //with a sequence number lower that of this tile (e.g. any loading
+      //tile following this entry in the chain)
+      tile.interimTile = null;
+      break;
+    } else if (tile.getState() == ol.Tile.State.LOADING) {
+      //keep this LOADING tile any loaded tiles later in the chain are
+      //older than this tile (lower seq num), so we're still interested in the request
+      prev = step(prev,tile);
```
ahocevar: Replace with `prev = tile;`

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
+    } else if (tile.getState() == ol.Tile.State.IDLE) {
+      //the head of the list is the most current tile, we don't need
+      //to start any other requests for this chain
+      prev = discard(prev,tile);
+    } else {
+      prev = step(prev,tile);
```
ahocevar: Replace with `prev = tile;`.

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
+  var /** @type {ol.Tile} */ head = tile;
+
+  if (!head.interimTile) {
+    return;
+  }
+
+  tile = head.interimTile;
+  var prev = head;
+
+  var discard = function(prev, tile) {
```
ahocevar: Not needed. See below.

src/ol/source/tileimage.js
```diff
@@ -278,6 +278,84 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
...
+      tile.interimTile = null;
+      break;
+    } else if (tile.getState() == ol.Tile.State.LOADING) {
+      //keep this LOADING tile any loaded tiles later in the chain are
+      //older than this tile (lower seq num), so we're still interested in the request
+      prev = step(prev,tile);
+    } else if (tile.getState() == ol.Tile.State.IDLE) {
+      //the head of the list is the most current tile, we don't need
+      //to start any other requests for this chain
+      prev = discard(prev,tile);
```
ahocevar: Replace with `prev.interimTile = tile.interimTile;`.

ahocevar:
I'm done with my review. Good work, only a few minor comments to simplify things and shuffle stuff around a bit.

src/ol/source/tile.js
thomasmoelhave: This was done for the benefit of the renderer that deals with a ol.source.Tile, not an ol.source.TileImage as far as I understand. It makes sense to move the getInterimTile to the ol.TileImage, but we would then also need a method on ol.Tile matching the method you commented on. I'll try it out and you can see what you think when I update the pull request, perhaps I am misunderstanding something.

src/ol/source/tileimage.js
thomasmoelhave: Done

src/ol/source/tileimage.js
thomasmoelhave: Done

src/ol/source/tileimage.js
thomasmoelhave: I like your update to the linked list traversal stuff, it certainly saves some code.

src/ol/tile.js
thomasmoelhave: ok, wasn't sure you wanted to impose an order on the id's (comparing two id's is usually nonsensical) but if the uids happen to have these properties - and you happy about making that assumption part of the api, then I'll happily use that.

ahocevar: Yes, that's fine. We're doing it [elsewhere](https://github.com/openlayers/ol3/blob/v3.18.1/src/ol/object.js#L105-L109) already, and adding that comment to `ol.getUid()` is overdue.

thomasmoelhave: I suppose there should be a unit test for this behavior then, is that already the case?

thomasmoelhave: Actually, I just realized in this latest version of the patch I had simplified it enough that the sequence number was only ever explicitely referenced in the assert statement checking that the order of the tiles in the list was correct. Since this will now be tested in a unit test, we can get rid of the sequence number/uid stuff altogether, further simplifying things.

thomasmoelhave:
Thanks a lot for your review @ahocevar. I have rebased on top of master and addressed most of your concerns. I think they made the code simpler and easier to understand :+1:.

1. got rid of the sequence number stuff, I used it during development but realized that it was only used for assertions in the final version, so I decided just to drop it completely for simplicity. I figure the unit test will test this aspect instead.

2. moved the methods from `ol.source.TileImage`  into `ol.ImageTile`. I also need `getInterimTile` in `ol.Tile` (like I needed it in `ol.source.Tile` since the renderers only work with objects of the latter kind. If the object is not an `ol.ImageTile` the current behavior is preserved (return the interim tile if it is not null).

3. I do not have time to do unit tests today, but I will get to them asap and update this pull request when they are done,.

thomasmoelhave:
I added a unit test, I think we're ready for a second review.

thomasmoelhave:
I have rebased to newest master resolving a conflict caused by the removal of the dom renderer due to #5815.

ahocevar:
Reviewing now.

ahocevar:
Nice! This is almost good to go. Please consider cherry-picking b156e13f197d002fab4e7c22662d3db34da554d4, which moves the interim tile related methods to `ol.Tile`, saving a few more bytes of unnecessary code.

I'll merge when you have cherry-picked.

thomasmoelhave:
I cherry-picked your commit and added another one on top of that which moved the test away from the ImageTile tests, I guess that makes the most sense.

ahocevar:
Very nice. Thanks for your effort on this @thomasmoelhave.

tschaub:
This looks like really nice work.  I haven't taken time to dig into it, but I appreciate your very clear descriptions @thomasmoelhave.  Thanks for the great contributions!

thomasmoelhave:
Thanks for the review and discussion @ahocevar. Thanks @tschaub  for the encouragement :).

PR #5798, opened 2016-08-29
Move to 90° along the circle, not 0°

### LINKED ISSUE #5797: Circle rotation view problem
A circle drawn with ol.geom.Circle is not properly rendered in latest version. When the map is rotated, an extra line comes making a chord in the circle. 

To reproduce the problem:
* Go to http://openlayers.org/en/latest/examples/draw-features.html.
* Draw a circle
* Rotate the map
* You can see extra line in the circle

Please fix.

ahocevar:
Fixes #5797.


fredj:
+1, thanks

PR #5794, opened 2016-08-28
[WIP] Fix more missing requires

ahocevar:
This is a follow-up on #5702 to resolve all dependency problems by applying the rules outlined in https://github.com/openlayers/ol3/pull/5702#issuecomment-238585089.

This is a branch of the main repository so all core committers can contribute easily. Other contributors, please mention commits that can be cherry-picked in comments on this pull request.

Thanks in advance to anyone helping to get this done.

PR #5801, opened 2016-08-29
Remove ol.style.ImageState and use ol.ImageState instead

### LINKED ISSUE #5799: ol.ImageState vs ol.style.ImageState
We have two enums for the loading state of an image:
 * `ol.ImageState`: https://github.com/openlayers/ol3/blob/v3.18.0/src/ol/imagebase.js#L9-L17
 * `ol.style.ImageState`: https://github.com/openlayers/ol3/blob/v3.18.0/src/ol/style/image.js#L5-L13

We should only keep one

fredj:
fixes #5799

### DIFF
--- a/src/ol/image.js
+++ b/src/ol/image.js
@@ -1,8 +1,8 @@
 goog.provide('ol.Image');
 
 goog.require('ol');
+goog.require('ol.Image');
 goog.require('ol.ImageBase');
-goog.require('ol.ImageState');
 goog.require('ol.events');
 goog.require('ol.events.EventType');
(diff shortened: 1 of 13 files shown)


ahocevar:
Good to merge! I assume you were planning to rename ol.ImageState to ol.Image.State as part of #5794.

fredj:
`ol.ImageState` renamed to `ol.Image.State`

ahocevar:
Nice, please merge.

fredj:
Thanks for the review

PR #5809, opened 2016-08-30
Reuse ol.LayerState object

fredj:
Reduce the number of temporary objects by reusing the layer state object.


tschaub:
Looks reasonable.

PR #5811, opened 2016-08-31
Use ol.global.getComputedStyle instead of window.getComputedStyle

fredj:


probins:
good catch. I just did a search and it looks like this is the only place where `window` is used.

fredj:
Thanks for the review @probins

PR #5815, opened 2016-08-31
Remove the DOM renderer

### LINKED ISSUE #5813: Remove the DOM renderer
I anticipate that "build flags" will not be used frequently and that even when used, may not always result in complete dead code removal for unused renderers.  To reduce build sizes and reduce the library size overall, I suggest we remove the DOM renderer.

In order to avoid breaking applications that use it, we could fall back to the Canvas renderer if "dom" is supplied.

I'm interested to hear if others think there is value in keeping it around.

tschaub:
This removes the DOM renderer.  For backwards compatibility, maps constructed with `renderer: 'dom'` will still work (and will get the Canvas renderer).

Fixes #5813.


tschaub:
This removes ~12KB from builds (about 0.1 Leaflet) and increases our test coverage by 1.5%.

examples/side-by-side.html
```diff
@@ -1,10 +1,10 @@
 ---
 layout: example.html
 title: Shared Views
-shortdesc: The three maps, one WebGL, one Canvas, one DOM, share the same center, resolution, rotation and layers.
+shortdesc: Two maps width different renderers share view properties
```
ischas: width --> with

ahocevar:
If we do this (which I'm in favor of), we can remove at least 3 functions from ol.dom: `ol.dom.transformElement2D`, `ol.dom.canUseCssTransform3D`and `ol.dom.setTransform`.

examples/side-by-side.html
tschaub: Good catch.  Addressed in 8316b0b.

tschaub:
Interesting [failure](https://travis-ci.org/openlayers/ol3/builds/156670398):
```
JavaScript ERROR: TypeError: undefined is not a function (evaluating 'ArrayBuffer.b(a)')
TRACE:
 -> http://localhost:8000/build/hosted/HEAD/build/ol.js: 492 (in function "f")
 -> http://localhost:8000/build/hosted/HEAD/build/ol.js: 503 (in function "Fa")
 -> http://localhost:8000/build/hosted/HEAD/build/ol.js: 360
 -> : 0
  phantomjs://code/check-example.js:20 in onError
make: *** [build/timestamps/check-mapbox-vector-tiles-timestamp] Error 1
```

tschaub:
This is now ready for review.

tschaub:
Rebasing to resolve conflicts.

tschaub:
> If we do this (which I'm in favor of), we can remove at least 3 functions from ol.dom: ol.dom.transformElement2D, ol.dom.canUseCssTransform3Dand ol.dom.setTransform.

Good catch @ahocevar .  This ended up being a nice simplification.

Thanks for any reviews.

fredj:
LGTM

ahocevar:
Nice, please merge.

PR #5816, opened 2016-08-31
(WIP) Add a method to abort tile loading

tschaub:
Not tested yet, just pushing this up for discussion.

### DIFF
--- a/src/ol/imagetile.js
+++ b/src/ol/imagetile.js
@@ -62,9 +62,18 @@ ol.inherits(ol.ImageTile, ol.Tile);
 /**
  * @inheritDoc
  */
+ol.ImageTile.prototype.abort = function() {
+  this.disposeInternal();
+};
+
+
+/**
+ * @inheritDoc
+ */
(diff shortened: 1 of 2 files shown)


ahocevar:
This is definitely the safest way to abort a request, at least to a point that the browser does not have to deal with it any more. Thanks for getting this started @tschaub.

The interesting part will be how to hook this up to the current tile queue and tile load handling.

tsauerwein:
Good idea! I gave it a try: https://jsfiddle.net/nw34gLgt/11/

If you change the `src` attribute, the download of the previous image is stopped. In the developer tools the request is shown with a status code `200 OK`. But if you request the image again, it is properly download again (so you won't see a broken, partly downloaded image).

I tested with a local server (`python -mSimpleHTTPServer`) and the connection socket was simply closed (`error: [Errno 32] Broken pipe`). No HTTP abort was sent.

tschaub:
Maybe replaced by #6939.  Closing due to lack of activity.

PR #5826, opened 2016-09-01
Revert "Do not close the ring when we use closePath()"

### LINKED ISSUE #5822: vector render problem
ol version: 3.18.1
[http://jsfiddle.net/cakbfwff/6/](http://jsfiddle.net/cakbfwff/6/)
if you extend the map area in this fiddle you'll see that the right square will lose a vertex

ahocevar:
This reverts commit 370c0c084673259d1c90c211b34217be0a2a0ba1. That optimisation left us with a missing end point when the last vertex of a polygon was skipped.

Fixes #5822.

### DIFF
--- a/src/ol/render/canvas/replay.js
+++ b/src/ol/render/canvas/replay.js
@@ -1268,10 +1268,8 @@ ol.render.canvas.PolygonReplay.prototype.drawFlatCoordinatess_ = function(flatCo
   for (var i = 0; i < numEnds; ++i) {
     var end = ends[i];
     var myBegin = this.coordinates.length;
-    var myEnd = this.appendFlatCoordinates(flatCoordinates, offset, end, stride,
-        // Performance optimization: only close the ring when we do not have a
-        // stroke. Otherwise closePath() will take care of that.
-        !stroke);
+    var myEnd = this.appendFlatCoordinates(
+        flatCoordinates, offset, end, stride, true);
     var moveToLineToInstruction =
         [ol.render.canvas.Instruction.MOVE_TO_LINE_TO, myBegin, myEnd];
     this.instructions.push(moveToLineToInstruction);


tschaub:
Not looking at the test case, but is the problem that we are omitting the final vertex when simplifying?  If so, that could be fixed.

ahocevar:
Yes, that's the problem @tschaub.

tschaub:
If this fixes the rendering issue, then I think it makes sense to merge and get in a patch release.  It would be good to revisit this, add tests, and fix the underlying problem - it feels like there are two optimizations working at odds here (adding the last point is redundant if we are going to close the path, but we need to close the path if we are not adding the last point).

src/ol/render/canvas/replay.js
```diff
@@ -1268,10 +1268,8 @@ ol.render.canvas.PolygonReplay.prototype.drawFlatCoordinatess_ = function(flatCo
   for (var i = 0; i < numEnds; ++i) {
     var end = ends[i];
     var myBegin = this.coordinates.length;
-    var myEnd = this.appendFlatCoordinates(flatCoordinates, offset, end, stride,
-        // Performance optimization: only close the ring when we do not have a
-        // stroke. Otherwise closePath() will take care of that.
-        !stroke);
+    var myEnd = this.appendFlatCoordinates(
+        flatCoordinates, offset, end, stride, true);
```
tschaub: Rather than rebasing, it might be easier to reset to `openlayers/master` and manually apply this change to `src/ol/render/canvas/polygonreplay.js`.

ahocevar:
Sounds good. Thanks for the review.

ahocevar:
Trying to reopen with a new branch.

ahocevar:
Not working.

PR #5828, opened 2016-09-01
Always close polygon rings

### LINKED ISSUE #5822: vector render problem

ahocevar:
This ensures that we do not optimize away the last vertex of a polygon.

Replaces #5826.

Fixes #5822.


ahocevar:
See review and comments in #5826.

cs09g:
related to #6063?

PR #5830, opened 2016-09-01
Add clone methods to styles

simonseyock:
See https://github.com/openlayers/ol3/issues/5823


simonseyock:
Was using an old version accidentaly rebasing and then opening again.

Issue #5775, opened 2016-08-25
ol.style.Image classes should defer rendering to the layer renderer


ahocevar:
Currently `ol.style.Image` subclasses render an image to an AtlasManager. Instead, layer renderers should decide what to do with the information stored in an Image style. For example in the case of `ol.style.Circle`, the Canvas renderer could more efficiently just render a circle with the provided radius, fill and stroke, instead of the style having to add an image to the AtlasManager for every radius-fill-stroke combination.

ahocevar:
Also see #5748.

tschaub:
Agreed.  It would be nice if the AtlasManager became a WebGL specific thing.

ahocevar:
@wglas85, if you want to work on this, here is a brief outline of what I have in mind:

#### 1. Create a new `ol.style.Shape` class
A `ol.style.Shape` would be configured with a `ol.Size` and a `ol.geom.Geometry` (can also be a `ol.geom.GeometryCollection`. The geometry defines the shape in pixel space, bounded by a rectangle with extent `[0, 0, size[0], size[1]]`.

#### 2. Make `ol.style.RegularShape` inherit from `ol.style.Shape`
All rendering code can be removed, and the shape will be described as `ol.geom.Geometry` instead.

A related refactoring would be to merge `ol.style.Icon` and `ol.style.Image` into a single class.

#### 3. Update point renderers
This change mostly involves `ol.renderer.vector.renderPointGeometry_`,  `ol.render.canvas.Immediate.prototype.drawPoint` and `ol.render.canvas.Immediate.prototype.drawPoint` (and WebGL counterparts, if any). Now that `ol.style.Style#getImage()` can return an `ol.stlye.Icon` or an `ol.style.Shape`, we do not always get an already rendered image any more. These functions are now responsible for creating and caching images (and making an AtlasManger handle them for WebGL only) from the geometry and size defined in an `ol.style.Shape`.

With these changes in place, `ol.style.*` classes do not have any logic any more, and are just containers for properties. And AtlasManager will no longer be an implicit dependency of `ol.style.RegularShape`, reducing the size of builds that do not need a WebGL renderer.

wglas85:
@ahocevar could you please tell me how to temporarily activate the immediate canvas renderer, so I can start by implementing this bare-bone render without having to tackle all the replay complexities?

ahocevar:
[...]

PR #5832, opened 2016-09-01
Add clone methods to styles

simonseyock:
This pull request adds clone methods to most style objects. The clone methods are supposed to clone all properties which support getting cloned.


src/ol/style/regularshape.js
```diff
@@ -113,7 +113,13 @@ ol.style.RegularShape = function(options) {
    */
   this.hitDetectionImageSize_ = null;
 
-  this.render_(options.atlasManager);
+  /**
+   * @private
+   * @type {ol.style.AtlasManager|undefined}
+   */
+  this.atlasManager_ = options.atlasManager;
```
simonseyock: I am saving the atlasManger to apply it to the cloned style, too. Is this desired?

test/spec/ol/style/circle.test.js
```diff
@@ -76,6 +76,40 @@ describe('ol.style.Circle', function() {
...
+      original.setScale(1.5);
+      var clone = original.clone();
+      expect(original.getFill()).to.not.be(clone.getFill());
+      expect(original.getFill().getColor()).to.be(clone.getFill().getColor());
+      expect(original.getOpacity()).to.be(clone.getOpacity());
+      expect(original.getRadius()).to.be(clone.getRadius());
+      expect(original.getScale()).to.be(clone.getScale());
+      expect(original.getSnapToPixel()).to.be(clone.getSnapToPixel());
+      expect(original.getStroke()).to.not.be(clone.getStroke());
+      expect(original.getStroke().getColor()).to.be(clone.getStroke().getColor());
```
simonseyock: These kind of test can only check the values which have getters. Private properties cant be tested, this is not a problem for the circle, but that would be a problem for the icon. Think of the color to taint it. Is there a way to check if the style is really rendered the same?

src/ol/style/style.js
```diff
@@ -72,6 +72,22 @@ ol.style.Style = function(opt_options) {
 
 
 /**
+ * Clones the style.
+ * @return {ol.style.Style} The cloned style.
+ * @api
+ */
+ol.style.Style.prototype.clone = function() {
+  return new ol.style.Style({
+    geometry: this.getGeometry(),
```
simonseyock: should the geometry get cloned?

marcjansen:
I would expect a deep clone. This would mean to clone everything that is cloneable (e.g. geometries) changing sth. In th clone should usually not affect the original and vice versa. There are certain properties where this will be hard/undoable (geometryfunction, e.g.)

simonseyock:
Sounds reasonable.

What i did not look into yet is a clone method for the AtlasManager. I dont think it is needed. What do you think?

src/ol/style/regularshape.js
simonseyock: Instead there could be an optional atlasManager parameter for clone for shape styles.

marcjansen:
TBH I know next to nothing about the AtlasManager. @ahocevar hinted he might rework that anyway.

Arrays, like the colors, might also need special love.

ahocevar:
See #5775 for what I had in mind. It would mean to take anything renderer specific out of `ol.style.*` (mostly affects `ol.style.Image` and subclasses). The atlas manager would then become a WebGL renderer option.

ahocevar:
But what I said above should not affect this pull request. I don't think people would expect the atlas manager to be cloned, but @tsauerwein may have better advice here.

tsauerwein:
> What I said above should not affect this pull request. I don't think people would expect the atlas manager to be cloned, but @tsauerwein may have better advice here.

Yes, it makes sense not to clone the atlas manager. It is meant to be shared among styles.

test/spec/ol/style/fill.test.js
```diff
@@ -0,0 +1,24 @@
...
+      expect(clone instanceof ol.style.Fill).to.eql(true);
+      expect(clone).to.not.be(original);
+    });
+
+    it('clones all values', function() {
+      var original = new ol.style.Fill({
+        color: '#319FD3'
+      });
+      var clone = original.clone();
+      expect(original.getColor()).to.be(clone.getColor());
```
simonseyock: this will be changed to: to.eql

simonseyock:
Hey, i added a clone functionality for the image property of the icon style. What do you think about it?

src/ol/style/icon.js
```diff
@@ -179,17 +179,29 @@ ol.inherits(ol.style.Icon, ol.style.Image);
...
+    if (oldImage.tagName.toUpperCase() === 'IMG') {
+      newImage = /** @type {Image} */ (this.getImage(1).cloneNode(true));
+    } else {
+      newImage = /** @type {HTMLCanvasElement} */ (document.createElement('canvas'));
+      var context = newImage.getContext('2d');
+      newImage.width = oldImage.width;
+      newImage.height = oldImage.height;
+      context.drawImage(oldImage, 0, 0);
+    }
+  }
```
simonseyock: image and canvas cloning

marcjansen:
I will have a look tomorrow, OK?

simonseyock:
Sure :) Thanks in advance

src/ol/style/fill.js
```diff
@@ -31,6 +31,18 @@ ol.style.Fill = function(opt_options) {
 
 
 /**
+ * Clones the style.
+ * @return {ol.style.Fill} The cloned style.
+ * @api
+ */
+ol.style.Fill.prototype.clone = function() {
+  return new ol.style.Fill({
+    color: (this.getColor() instanceof Array) ? this.getColor().slice(0) : this.getColor()
```
marcjansen: What about a utility method in the `ol.color` namespace?

```js
ol.color.clone(color) {
  var clone = color;
  if (Array.isArray(color)) {
    clone = Array.from(color);
  } else if (color instanceof Uint8Array) {
    clone = Uint8Array.from(color);
  } else if (color instanceof Uint8ClampedArray) {
    clone = Uint8ClampedArray.from(color);
  }
  // String, CanvasPattern or CanvasGradient
  // are left untouched and not cloned
  return clone;
}
```

Untested.

src/ol/style/icon.js
```diff
@@ -172,6 +174,47 @@ ol.inherits(ol.style.Icon, ol.style.Image);
...
+ * Clones the style.
+ * @return {ol.style.Icon} The cloned style.
+ * @api
+ */
+ol.style.Icon.prototype.clone = function() {
+  var oldImage = this.getImage(1);
+  var newImage;
+  if (this.iconImage_.getImageState() === ol.Image.State.LOADED) {
+    if (oldImage.tagName.toUpperCase() === 'IMG') {
+      newImage = /** @type {Image} */ (this.getImage(1).cloneNode(true));
```
marcjansen: Reuse `oldImage`?

src/ol/style/icon.js
```diff
@@ -172,6 +174,47 @@ ol.inherits(ol.style.Icon, ol.style.Image);
...
+    } else {
+      newImage = /** @type {HTMLCanvasElement} */ (document.createElement('canvas'));
+      var context = newImage.getContext('2d');
+      newImage.width = oldImage.width;
+      newImage.height = oldImage.height;
+      context.drawImage(oldImage, 0, 0);
+    }
+  }
+  return new ol.style.Icon({
+    anchor: this.anchor_.slice(0),
```
marcjansen: Better `Array.from` and hint that a polyfill might be needed? Unsure here. Wouldn't block a `slice(0)`.

src/ol/style/icon.js
```diff
@@ -172,6 +174,47 @@ ol.inherits(ol.style.Icon, ol.style.Image);
...
+    anchorOrigin: this.anchorOrigin_,
+    anchorXUnits: this.anchorXUnits_,
+    anchorYUnits: this.anchorYUnits_,
+    crossOrigin: this.crossOrigin_,
+    color: this.color_ !== null ? this.color_.slice(0) : undefined,
```
marcjansen: Reuse the clone method suggested above

src/ol/style/regularshape.js
```diff
@@ -140,6 +146,31 @@ ol.inherits(ol.style.RegularShape, ol.style.Image);
...
+ * Clones the style.
+ * @return {ol.style.RegularShape} The cloned style.
+ * @api
+ */
+ol.style.RegularShape.prototype.clone = function() {
+  var style = new ol.style.RegularShape({
+    fill: this.getFill() ? this.getFill().clone() : undefined,
+    points: this.getPoints(),
+    radius: this.getRadius(),
+    radius2: this.getRadius2(),
```
marcjansen: According to the API docs there is also a `radius1` configuration for `RegularShape`. I cannot find it in the OpenLayers code right now, though… :astonished:

src/ol/style/stroke.js
```diff
@@ -63,6 +63,23 @@ ol.style.Stroke = function(opt_options) {
 
 
 /**
+ * Clones the style.
+ * @return {ol.style.Stroke} The cloned style.
+ * @api
+ */
+ol.style.Stroke.prototype.clone = function() {
+  return new ol.style.Stroke({
+    color: (this.getColor() instanceof Array) ? this.getColor().slice(0) : this.getColor(),
```
marcjansen: Reuse the `ol.color.clone` method suggested above

src/ol/style/regularshape.js
bartvde: https://github.com/openlayers/ol3/blob/00905e813123036d5ed85899d931b787647759e8/src/ol/style/regularshape.js#L70:L71 ?

bartvde: But radius1 is just an alias for radius

src/ol/style/style.js
```diff
@@ -72,6 +72,26 @@ ol.style.Style = function(opt_options) {
...
+ * @api
+ */
+ol.style.Style.prototype.clone = function() {
+  var geometry = this.getGeometry();
+  if (geometry && geometry.clone) {
+    geometry = geometry.clone();
+  }
+  return new ol.style.Style({
+    geometry: geometry,
+    fill: this.getFill() ? this.getFill().clone() : undefined,
```
marcjansen: What about `image`? Did you leave it out intentionally

src/ol/style/text.js
```diff
@@ -96,6 +96,28 @@ ol.style.Text.DEFAULT_FILL_COLOR_ = '#333';
...
 /**
+ * Clones the style.
+ * @return {ol.style.Text} The cloned style.
+ * @api
+ */
+ol.style.Text.prototype.clone = function() {
+  return new ol.style.Text({
+    font: this.getFont(),
+    rotation: this.getRotation(),
+    rotateWithView: this.getRotateWithView(),
```
marcjansen: The above two lines work but aren't mentoined on the API docs… hmmm.

marcjansen:
Very nice work so far, @simonseyock! I have added a couple of inline comments and suggestions, but this is in good shape already.

I would be pleased if we had a special test where on changes properties in the clone and the original is unaffected and vice-versa. I know you have this when you e.g. say 

```js
expect(original.getFill()).to.not.be(clone.getFill());
// or somewhere else
expect(original.color_).to.not.be(clone.color_);
```

…but I would appreciate the extra verbosity in a dedicated test.

Also the description of this PR should be a little bit more descriptive and up-to-date.

Once you have updated the pull request, just ping me and I will gladly have a final look. Thanks so far!

A question for @openlayers/ol3 : Can we use `Array.from` and friends? IE 11 would need a polyfill, Edge should be OK.

src/ol/style/regularshape.js
marcjansen: Ahhh! Thanks Bart.

src/ol/style/icon.js
simonseyock: Yes, that is an option, i was not sure if an image would be altered after giving it to a style. It seems a bit unlikely to me. You agree on cloning the canvas, though?

src/ol/style/style.js
simonseyock: Uhh, that slipped, thanks! Later i will add a test for this, too.

src/ol/style/text.js
simonseyock: That was added recently, are you sure you checked the most up to date documentation?

src/ol/style/fill.js
simonseyock: Sounds good!

src/ol/style/icon.js
marcjansen: With reusing I meant replacing `this.getImage(1)` with the already existing `oldImage` so that the line reads:

```js
newImage = /** @type {Image} */ (newImage.cloneNode(true));
```

src/ol/style/icon.js
```diff
@@ -172,6 +174,47 @@ ol.inherits(ol.style.Icon, ol.style.Image);
...
+  var oldImage = this.getImage(1);
+  var newImage;
+  if (this.iconImage_.getImageState() === ol.Image.State.LOADED) {
+    if (oldImage.tagName.toUpperCase() === 'IMG') {
+      newImage = /** @type {Image} */ (this.getImage(1).cloneNode(true));
+    } else {
+      newImage = /** @type {HTMLCanvasElement} */ (document.createElement('canvas'));
+      var context = newImage.getContext('2d');
+      newImage.width = oldImage.width;
+      newImage.height = oldImage.height;
```
marcjansen: Can be replaced with `ol.dom.createCanvasContext2D(oldImage.width, oldImage.height)`

Otherwise I like it.

src/ol/style/icon.js
marcjansen: Looks OK to me, yes.

src/ol/style/text.js
marcjansen: Not sure right now… might have been looking at sth. outdated :see_no_evil:.

marcjansen:
API docs of the clone methods should also state exactly what isn't cloned, IMHO.

src/ol/style/icon.js
simonseyock: Ah well, sure :)

src/ol/style/icon.js
simonseyock: yes, i need the HTMLCanvasElement

src/ol/style/icon.js
tschaub: Do we not expect `this.anchor_` to be an array?

I think `array.slice()` is a fine way to shallow clone an array.

marcjansen: Fine for me.

src/ol/style/fill.js
simonseyock: if i would use slice anyways to shallow copy an array, i could use this simplified variant:

```
ol.color.clone = function(color) {
  if (color.slice) {
    return color.slice();
  } else {
    return color;
  }
};
```

tschaub: I'm slightly curious why @marcjansen keeps suggesting `Array.from()` (and friends), but think @simonseyock understands we don't need it (`foo.slice()` creates a `Uint8ClampedArray` if `foo` is a `Uint8ClampedArray`).

tschaub:
> A question for @openlayers/ol3 : Can we use Array.from and friends? IE 11 would need a polyfill, Edge should be OK.

I think we should avoid making people use a polyfill when there is no clear reason for us to use the functionality it provides.  In the case of `Array.from()`, I don't think we have any need for it (we aren't dealing with `Map`, `Set`, or other array-like objects.

src/ol/style/fill.js
simonseyock: @marcjansen already agreed on using `.slice()`. i just wanted to ask about skipping all the instanceof checks.

tschaub: I think checking for `slice` is fine (and quite a bit nicer than `if (typeof color === 'string' || Array.isArray(color) || color instanceof Uint8Array || color instanceof Uint8ClampedArray)`).

As a side note, it looks like `ol.color.asArray()` wouldn't work when called with a `Uint8Array` or `Uint8ClampedArray`.  I'll create a ticket for that.

tschaub: I've proposed simplifying the `ol.Color` type in #5872 (to make the `ol.color` functions work with `ol.Color`).

marcjansen:
As stated earlier, fine for me. Go ahead with slice.

src/ol/style/fill.js
```diff
@@ -31,13 +31,14 @@ ol.style.Fill = function(opt_options) {
...
- * Clones the style.
+ * Clones the style. The color is not cloned if it is an {@link ol.ColorLike}.
  * @return {ol.style.Fill} The cloned style.
  * @api
  */
 ol.style.Fill.prototype.clone = function() {
+  var color = this.getColor();
   return new ol.style.Fill({
-    color: (this.getColor() instanceof Array) ? this.getColor().slice(0) : this.getColor()
+    color: (color && color.slice) ? color.slice() : color || undefined
```
simonseyock: I backed down from using an ol.color.clone method because sometimes the type to clone is {ol.Color|string} and sometimes it {ol.Color|ol.ColorLike} and it was easier this way. I could use ol.colorlike.isColorLike, though.

src/ol/style/fill.js
```diff
@@ -31,6 +31,19 @@ ol.style.Fill = function(opt_options) {
...
 
 /**
+ * Clones the style. The color is not cloned if it is an {@link ol.ColorLike}.
+ * @return {ol.style.Fill} The cloned style.
+ * @api
+ */
+ol.style.Fill.prototype.clone = function() {
+  var color = this.getColor();
+  return new ol.style.Fill({
+    color: (color && color.slice) ? color.slice() : color || undefined
```
simonseyock: I backed down from using an ol.color.clone method because sometimes the type to clone is {ol.Color|string} and sometimes it {ol.Color|ol.ColorLike} and it was easier this way. I could use ol.colorlike.isColorLike, though.

src/ol/style/regularshape.js
```diff
@@ -401,11 +431,12 @@ ol.style.RegularShape.prototype.draw_ = function(renderOptions, context, x, y) {
...
   context.beginPath();
+  var points = this.points_;
   if (this.radius2_ !== this.radius_) {
-    this.points_ = 2 * this.points_;
+    points = 2 * points;
   }
-  for (i = 0; i <= this.points_; i++) {
-    angle0 = i * 2 * Math.PI / this.points_ - Math.PI / 2 + this.angle_;
+  for (i = 0; i <= points; i++) {
+    angle0 = i * 2 * Math.PI / points - Math.PI / 2 + this.angle_;
```
simonseyock: i introduced a new variable here to store the doubled points, else the value would be doubled after each clone (and each call of draw_ and drawHitDetectionCanvas_).

test/spec/ol/style.test.js
```diff
@@ -1,97 +0,0 @@
...
-    expect(styleFunction).to.be(original);
-  });
-
-  it('throws on (some) unexpected input', function() {
-    expect(function() {
-      ol.style.Style.createFunction({bogus: 'input'});
-    }).to.throwException();
-  });
-
-});
```
simonseyock: I moved the style.test.js file to the style/ folder. I don't know if this fits into this PR, i could open an other PR for this.

simonseyock:
Hey, i added tests for the remaining clone methods and fixed some issues related to the clone methods. I think this PR is ready for review. Would be nice if you could look into it again @marcjansen.

src/ol/style/regularshape.js
```diff
@@ -401,11 +431,12 @@ ol.style.RegularShape.prototype.draw_ = function(renderOptions, context, x, y) {
   context.translate(x, y);
 
   context.beginPath();
+  var points = this.points_;
   if (this.radius2_ !== this.radius_) {
-    this.points_ = 2 * this.points_;
+    points = 2 * points;
```
simonseyock: i introduced a new variable here to store the doubled points, else the value would be doubled after each clone (and each call of draw_ and drawHitDetectionCanvas_).

tschaub:
@simonseyock - can you please rebase this on master?  Depending on the name of your remote, this could look something like this:

```js
git push --force cloneStyle
```

(It looks like you are not set up with the same email everywhere you push commits.  You could solve this by running `git config --global user.email "your@email.com"` everywhere you run git.)

simonseyock:
Sure.

simonseyock:
I accidentally comitted a .orig file of kdiff, so i had to force push again.

src/ol/style/regularshape.js
simonseyock: This will cause a change of behaviour of the getPoints method:
It will now always return the initially given number of points, regardless of radius2 was given to the constructor or not.

marcjansen: If I understand that correct you could have done the following alternatively (in `clone`), right:

```js
var style = new ol.style.RegularShape({
  // …
  points: this.radius2_ !== this.radius_ ? this.getPoints() / 2 : this.getPoints()
  // …
});
```

This way we wouldn't need the above changes, right?

marcjansen:
This is really looking good from my point of view, @simonseyock. I have added one comment with regard to regular shapes and `getPoints()`. If my understanding is correct we come away without changing the meaning of the returned value from `getPoints`.

I am :+1: for merging (once the `getPoints` thing is settled).

@tschaub, do you want to weigh in?

src/ol/style/regularshape.js
simonseyock: Yes, that is true, but for me it felt like a bug that getPoints would return the double amount of points i had given to the constructor. This isn't documented, too.

marcjansen: Still, I would like to keep the amount of both
* code in this PR, and
* the number of changes for users
minimal.

If I construct a regular polygon with two circles each having 5 points, the total number of points is 10, right? So the current behaviour makes some sense (to me at least).

For this PR I suggest to not change the semantics of that method. Feel free to open an issue where to discuss this. I would not make that change in this PR as it makes it more complicated than it has to be.

It is basically the same you did with the `Style` test file. I would have left it where it is, to keep the changes to review for this PR minimal. (It is OK that you did move it, no worries to change that back again.). An issue where that is discussed would have been enough, I guess.

Don't get me wrong, I **really** appreciate the time and effort you have put into this already. I honestly want to get this in rather sooner than later, but the less changes you introduce, the less controversial it is.

simonseyock: Ok, i agree with you, it is better to seperate the concerns. I open another issue for this. I thought that about the moving of the file, too. Thanks for the review!

simonseyock:
I restored the behaviour of ol.style.RegularShape#getPoints. (and it catched my eye that i was using eql and be a little inconsistent, but this does not change anything)

marcjansen:
Thanks for the work on this and the patience to get it in, @simonseyock. Great work.

If someone has additional comments on this, we can always do it in follow-up PRs.

PR #5834, opened 2016-09-01
Rename goog.DEBUG to ol.DEBUG

thomasmoelhave:
617dca34f53 introduced a stray goog.DEBUG.


tschaub:
Thanks @thomasmoelhave.  We should add a new custom rule for the linter that makes sure the only member expressions that start with `goog` are `goog.require` and `goog.provide` (until those are gone).

PR #5835, opened 2016-09-01
document and test ol.getUid sequence behavior

thomasmoelhave:
@ahocevar mentioned (in the #5795 review) that we should document the fact that ol.getUid returns an increasing sequence of IDs. If that's still the policy, maybe this would be helpful.

### DIFF
--- a/test/spec/ol/index.test.js
+++ b/test/spec/ol/index.test.js
@@ -1,0 +1,23 @@
+goog.provide('ol.test');
+goog.require('ol');
+
+describe('getUid()', function() {
+  it('is constant once generated', function() {
+    var a = {};
+    expect(ol.getUid(a)).to.be(ol.getUid(a));
+  });
+
+  it('generates a strictly increasing sequence', function() {
+    var a = {} , b = {}, c = {};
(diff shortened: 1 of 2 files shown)


test/spec/ol/index.test.js
```diff
@@ -0,0 +1,23 @@
+goog.provide('ol.test.index');
+goog.require('ol');
```
thomasmoelhave: Not entirely sure what to `goog.provide` for `index.js`, should it be just `ol.test`?

ahocevar:
Very nice @thomasmoelhave. Thanks for taking this on.

test/spec/ol/index.test.js
ahocevar: `ol.test` sounds good.

thomasmoelhave: Fixed this in the newest update.

ahocevar:
@thomasmoelhave, can you please sign a [CLA](https://github.com/openlayers/cla) so we can merge this? Thanks.

thomasmoelhave:
Oh right, done.

PR #5837, opened 2016-09-02
Remove ol.global

### LINKED ISSUE #5814: Remove ol.global
I'm not sure we get any real benefit from using `ol.global` rather than relying on regular scoping behavior.  We are already inconsistent with our use of globals.

E.g. we check for a global `navigator` [in some places](https://github.com/openlayers/ol3/blob/3b6f1f99bc9b867845fcab48420aebce7ab8a8cf/src/ol/has.js#L8-L9) and `ol.global.navigator` [in others](https://github.com/openlayers/ol3/blob/3b6f1f99bc9b867845fcab48420aebce7ab8a8cf/src/ol/geolocation.js#L136-L141).  We also rely on a global `Image` constructor rather than using `ol.global.Image`.

We'll need to change how we override globals in tests, but I don't think this will make any difference for users of the library.  You already have to provide global shims if you run OpenLayers outside a browser.

tschaub:
The use of `ol.global` was very inconsistent (sometimes `setTimeout`, sometimes, `ol.global.setTimeout` - same for `navigator`, `document`, etc.).  This change removes `ol.global` in favor of relying on regular scoping behavior and using `window` where needed.  To run OpenLayers outside of a browser, `window` and browser globals must be provided.

Fixes #5814.

### DIFF
--- a/src/ol/color.js
+++ b/src/ol/color.js
@@ -1,6 +1,5 @@
 goog.provide('ol.color');
 
-goog.require('ol');
 goog.require('ol.asserts');
 goog.require('ol.math');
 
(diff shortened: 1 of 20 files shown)


ahocevar:
Beautiful. Please merge.

PR #5838, opened 2016-09-02
Require ol where ol.DEBUG is being used

marcjansen:
This is a follow up in the work done in #5794.

It removes the eslint warnings

```
warning  Missing goog.require('ol.DEBUG') or goog.require('ol')  openlayers-internal/no-missing-requires
```

by adding

```js
goog.require('ol');
```

to the appropriate files.

I hope this is the right way to fix this.

Please review.

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -1,5 +1,7 @@
 goog.provide('ol.array');
 
+goog.require('ol');
+
 
 /**
  * Performs a binary search on the provided sorted list and returns the index of the item if found. If it can't be found it'll return -1.
--- a/src/ol/extent.js
+++ b/src/ol/extent.js
@@ -2,6 +2,7 @@ goog.provide('ol.extent');
 goog.provide('ol.extent.Corner');
 goog.provide('ol.extent.Relationship');
 
+goog.require('ol');
 goog.require('ol.asserts');
 
 
--- a/src/ol/geom/flat/closest.js
+++ b/src/ol/geom/flat/closest.js
@@ -1,5 +1,6 @@
 goog.provide('ol.geom.flat.closest');
 
+goog.require('ol');
 goog.require('ol.math');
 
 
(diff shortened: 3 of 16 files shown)


marcjansen:
Before:

```
✖ 227 problems (0 errors, 227 warnings)
```

After:
```
✖ 176 problems (0 errors, 176 warnings)
```

ahocevar:
Not sure if this is the right fix. Thinking about the proposed codemods to get CommonJS modules, there will probably be one to replace `ol.DEBUG`with `DEBUG`, so a `uglifyjs` build flag can be used. Same with other `@define`s.

An alternative approach could be to put all `@define`s in a new `ol.define` namespace, and `goog.require('ol.define');` in those cases. Any other ideas? @tschaub maybe?

marcjansen:
> Not sure if this is the right fix.

I was unsure myself. Would appreciate input by @tschaub (or everyone else of course).

probins:
May be a silly question, but are the debug msgs useful? I never use them myself, as I rarely use a debug build.

probins:
for UglifyJS, you can set `drop_console` as a compressor option, which should have the same effect for these debug cases.

marcjansen:
@tschaub, any chance you might share your thiughts on this?

marcjansen:
@tschaub, master of CommonJS: is this the right approach? Any chance you might take a look at this? TIA

tschaub:
Yes, this makes sense.

`ol.DEBUG` is assigned in `ol/index.js` (which has `provide('ol')`).  So `goog.require('ol')` is the way to get `ol.DEBUG`.

This would be similar to the following CommonJS code:

`ol/index.js` ~
```js
exports.DEBUG = true;
```

and elsewhere:
```js
var ol = require('ol');
ol.DEBUG ? yes() : no();
```

We can change how we support minifiers that support stripping out debug code later.  This is the right change for now.

marcjansen:
I assume this is a 'please merge', then. Thanks for you clarification, Tim.

PR #5839, opened 2016-09-02
Rename source.VectorEvent(*) to source.Vector.Event(*)

marcjansen:
In order to have only one provide per file, this PR suggests renaming

* `ol.source.VectorEvent` => `ol.source.Vector.Event`
* `ol.source.VectorEventType` => `ol.source.Vector.EventType`

It is basically the same as we did for `ol.source.TileEvent` in 74047e747d8c016644780a3dc0193ee9eddafa17 (part of #5794)

Please review.


ahocevar:
Looks good. Please merge when green.

marcjansen:
Thanks for the review. Will do as suggested.

PR #5842, opened 2016-09-02
Remove extra provide causing a circular dependency

ahocevar:
Addresses https://github.com/openlayers/ol3/commit/bed95e3b5b3fbefaf4cb2e8b2bd9ffaad95f261f#commitcomment-18885169.

### DIFF
--- a/src/ol/image.js
+++ b/src/ol/image.js
@@ -1,7 +1,6 @@
 goog.provide('ol.Image');
 
 goog.require('ol');
-goog.require('ol.Image');
 goog.require('ol.ImageBase');
 goog.require('ol.events');
 goog.require('ol.events.EventType');


ahocevar:
I think this does not require a formal review. I'll merge when green.

bartvde:
+1

ahocevar:
Feels even better with a revíew. Thanks @bartvde.

marcjansen:
You'll get one nonetheless. Looks good to me, please merge when green.

Why didn't ci catch this?

ahocevar:
Thanks, so many reviews :-)

> Why didn't ci catch this?

I think the way we build the dependency tree it does not matter, and we have no eslint rule checking if we require something that we provide in the same file.

Issue #5642, opened 2016-07-25
build errors


jachym:
Hi,

in last version of OpenLayers (3.17.1), we are not able to get it build using closure-builder from google

```
node_modules/openlayers/src/ol/array.js:173: ERROR - actual parameter 1 of ol.array.flatten does not match formal parameter
found   : *
required: (Array<*>|null)
      return flattened.concat(ol.array.flatten(value));
                                               ^

node_modules/openlayers/src/ol/color/color.js:53: ERROR - inconsistent return type
found   : (Array<number>|null|string)
required: (Array<number>|null)
    return color;
           ^

node_modules/openlayers/src/ol/feature.js:305: ERROR - assignment
found   : (Array<(null|ol.style.Style)>|ol.style.Style)
required: (Array<(null|ol.style.Style)>|null)
      styles = obj;
      ^

node_modules/openlayers/src/ol/format/featureformat.js:206: ERROR - actual parameter 1 of transform does not match formal parameter
found   : (Array<number>|null|ol.geom.Geometry)
required: (Array<number>|null)
      transform(transformed);
                ^

node_modules/openlayers/src/ol/interaction/mousewheelzoominteraction.js:93: ERROR - Property DOM_DELTA_PIXEL never defined on ol.global.WheelEvent
          wheelEvent.deltaMode === ol.global.WheelEvent.DOM_DELTA_PIXEL) {
                                                        ^

node_modules/openlayers/src/ol/interaction/mousewheelzoominteraction.js:96: ERROR - Property DOM_DELTA_LINE never defined on ol.global.WheelEvent
      if (wheelEvent.deltaMode === ol.global.WheelEvent.DOM_DELTA_LINE) {
                                                        ^

node_modules/openlayers/src/ol/map.js:1519: ERROR - assignment
found   : (Array<(ol.RendererType|string)>|null|string)
required: (Array<ol.RendererType<string>>|null)
      rendererTypes = options.renderer;
      ^

node_modules/openlayers/src/ol/renderer/canvas/canvasvectorlayerrenderer.js:334: ERROR - Cannot do '[]' access on a struct
          replayGroup, feature, styles[i],
                                       ^

node_modules/openlayers/src/ol/renderer/canvas/canvasvectorlayerrenderer.js:340: ERROR - actual parameter 3 of ol.renderer.vector.renderFeature does not match formal parameter
found   : (Array<(null|ol.style.Style)>|ol.style.Style)
required: (null|ol.style.Style)
[...]

PR #5843, opened 2016-09-02
Cast to ol.VectorTile so getReplayState is valid

ahocevar:
Addresses https://github.com/openlayers/ol3/commit/c1b16217f28f50d4585a2ffda603b83042cbd2cc#commitcomment-18885235.

### DIFF
--- a/src/ol/renderer/canvas/vectortilelayer.js
+++ b/src/ol/renderer/canvas/vectortilelayer.js
@@ -143,7 +143,7 @@ ol.renderer.canvas.VectorTileLayer.prototype.renderTileReplays_ = function(
   var currentZ, i, ii, offsetX, offsetY, origin, pixelSpace, replayState;
   var tile, tileExtent, tilePixelResolution, tileResolution, tileTransform;
   for (i = 0, ii = tilesToDraw.length; i < ii; ++i) {
-    tile = tilesToDraw[i];
+    tile = /** @type {ol.VectorTile} */ (tilesToDraw[i]);
     replayState = tile.getReplayState();
     tileExtent = tileGrid.getTileCoordExtent(
         tile.getTileCoord(), this.tmpExtent);


ahocevar:
@schmidtk I'm wondering how your setup differs from ours, because as far as I know, we have all type checking enabled for the Closure Compiler, and did not run into an issue here.

ahocevar:
Maybe related to #5642.

ahocevar:
@gberaudo, didn't you change some compiler settings a while ago that made it smarter detecting types?

ahocevar:
I'll be away for a week now, so if this needs to be merged or closed, please go ahead @openlayers/core.

gberaudo:
I experimented switching to the [new type inference](https://github.com/google/closure-compiler/wiki/Using-NTI-(new-type-inference)) but I got lots of errors and decided to suspend the experiment until this matures in the compiler. So nothing was merged.

@ahocevar, if a cast is needed, it is better to do it early on the `tilesToDraw` variable (as an `Array<ol.VectorTile>`).

schmidtk:
@ahocevar I'm out of the office for about a week as well so I don't have access to our compiler options, but I can provide them when I get back if you'd like.

I also had issues with the new typedefs file. Our build was providing the file to the compiler, but I got a pile of unknown type errors for references to those types. I think syncing back up with the latest OL3 is on tap in our next sprint, so I'll have more information then.

pgiraud:
@ahocevar It needs to be done in https://github.com/ahocevar/ol3/blob/c4ee4e2359a55330b7c0871141782008124277dc/src/ol/renderer/canvas/vectortilelayer.js#L293 as well.

ahocevar:
I'd still like to hear about circumstances when the compiler complains about this, because it does not when building OpenLayers.

pgiraud:
Here's a way to reproduce the issue.
Follow the "[Compile application](http://openlayers.org/en/latest/doc/tutorials/closure.html)" tutorial. Fix the issue with bad annotation types by adding "node_modules/openlayers/src/ol/typedefs.js" in the "js" section.
Compiling the application + OL3 code seem to work well.
However, adding `manage_closure_dependencies` and setting it to true in the compiler configuration seem to break the compilation.

schmidtk:
We're compiling with `--dependency_mode=strict --entry_point=<root namespace>` which will build the dependency list from provides/requires. Since typedefs.js doesn't have a provide, it doesn't get included and types are reported as unknown.

ahocevar:
I don't think this pull request is relevant any more.

fredj:
> I don't think this pull request is relevant any more.
@ahocevar and what about #5927

PR #5845, opened 2016-09-03
Report when file requires what it is providing

marcjansen:
This catches circular dependencies where one writes…

```js
goog.provide('ol.foo.Humpty');
```

…and then later…

```js
goog.require('ol.foo.Humpty');
```

See also ae41b83a78235fdb15f64e0d8492d61230119386 and #5842.
Please review.

### DIFF
--- a/rules/valid-requires.js
+++ b/rules/valid-requires.js
@@ -10,8 +10,13 @@ exports.rule = {
   },
 
   create: function(context) {
+    let provides;
     return {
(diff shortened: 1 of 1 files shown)


marcjansen:
To test this you have to replace the correct file in your node_modules directory

marcjansen:
Opinions?

marcjansen:
Anybody want to have a look at this?

Issue #5742, opened 2016-08-17
Map crashes when clustering with a large amount of features


cmortazavi:
If you use the openlayers example for clustering 
http://openlayers.org/en/latest/examples/cluster.html

and you change the number of features to 200,000 then the map will crash once you zoom in too far.  This should be fixed since clustering will need to be used with large amounts of data.

ChouGary:
hello ,i has same problem as you,the best solution is Supercluster.
https://github.com/mapbox/supercluster/
https://www.mapbox.com/blog/supercluster/

cmortazavi:
This is a leaflet plugin?  Is there an openlayers solution?

ahocevar:
Supercluster is not a Leaflet plugin. It can be used with OpenLayers, preferably an `ol.layer.VectorTile`. If anyone wants to create an official example, pull requests are welcome.

We could also improve `ol.source.Cluster` by not clustering the whole layer extent, but only the visible extent with a buffer. Clusters would have to be recreated when panning then. Also for this improvement, pull requests are welcome.

ThomasG77:
I've created [a supercluster demo](https://webgeodatavore.github.io/supercluster/demo/index-ol.html) based on this thread  and the code is also available on [this Github repo](https://github.com/webgeodatavore/supercluster/commits/ol3-version).

If I submit a PR, adding some worker code (for perfs reasons) is ok although less ol3 related?

ahocevar:
You mean a pull request to improve `ol.source.Cluster` performance by introducing workers? I was more thinking about a pull request adding a supercluster example, and getting rid of `ol.source.Cluster`.

ahocevar:
But then again, removal of `ol.source.Cluster` may be too much. What did you really have in mind @ThomasG77?

ThomasG77:
@ahocevar To push a supercluster demo. Didn't plan to touch `ol.source.Cluster` and no opinion on removing/keeping it.

ahocevar:
Ok cool, please go ahead @ThomasG77. Keep it as simple as possible though.

cmortazavi:
Is there any chance of getting a fix for the ol.source.Cluster.  I think the api itself should handle the worker rather than us having to handle it.  The worker makes it difficult for our application to work since we have many clustered layers which can be turned on and off at run time and our feature definition is dynamic.  Instead of a plugin, I think the original api should be fixed.  What is the problem with the ol.source.Cluster?  is it a memory leak?

[...]

PR #5848, opened 2016-09-03
Add demo based on 3rd party supercluster library

ThomasG77:
We added the library in the resources as we didn't found it on a CDN or the official github repository.

We named the sample 'supercluster.html' and used the associated 'supercluster.js'.
The particularity of this example is the use of web workers with 'supercluster-worker.js'.
We added the GeoJSON version of the KML earthquakes data as 'supercluster' needs GeoJSON for input.

We may need to correct eslint comment '// eslint-disable-next-line' as it was
the fastest way to make eslint happy with importScripts and supercluster
declarations in the worker file.

Related to #5742 discussion.


ThomasG77:
Do you need a rebase before reviewing?

marcjansen:
No, not necessarilly. Makes things easier most of the time, though. Also the commit history is then more cleaned up.

I will have look tomorrow, whether this is rebased or not.

marcjansen:
I have had a look at the example. Thanks for it.

Right now I don't want to merge it, because I want an external resource (URL) of `supercluster`, and not included in the `resource/` directory.

If at one point an online URL exists, I would also want a bigger dataset that gets clustered, if possible.

Thanks any how. This PR can be left open, so you can update it when things change.

ThomasG77:
Ongoing PR on supercluster repo to publish library to a CDN. Will update PR when done

ThomasG77:
> If at one point an online URL exists, I would also want a bigger dataset that gets clustered, if possible.

Didn't find one at the moment. Except this, all requirements for PR review & merge OK IMO.

ThomasG77:
> If at one point an online URL exists, I would also want a bigger dataset that gets clustered, if possible.

Except this, ready for review & merge IMO

tschaub:
Please reopen if this can be completed before the 4.6 release.

PR #5851, opened 2016-09-05
Remove unused imageByContext_ in ol.ImageTile

fredj:
Leftover from #5815


bartvde:
Good catch, please merge when green

fredj:
Thanks for the review

PR #5853, opened 2016-09-05
Allow more than one digit range in ol.TileUrlFunction.expandUrl

fredj:
To be able to have number range like `{9-13}`


ahocevar:
The code would be easier to read if you would not assign `match` inside the `if` statement.

fredj:
I've moved the assignment outside of the `if`

fredj:
Thanks for the review

PR #5861, opened 2016-09-07
More renaming to have one provide per file

fredj:


bartvde:
LGTM thanks for the continued work on this

fredj:
Thanks for the review

PR #5872, opened 2016-09-10
Keep ol.Color as Array<number>

### LINKED ISSUE #5871: Typed arrays don't work with `ol.color` functions
In #5670, the `ol.Color` type was changed from `Array.<number>` to `Array.<number>|Uint8Array|Uint8ClampedArray`.  This means that all functions that accept `ol.Color` need to be able to handle the new type.

e8682493dfbf62a8d84bf54555debf129173223c demonstrates that `ol.color.asArray()` doesn't work with typed arrays.  In addition, `ol.color.isValid()` gives unexpected results with something like `new Uint8Array([255, 255, 255, 255])`.

tschaub:
The changes in #5670 made it so `ol.Color` could be a typed array, but typed arrays don't work with the `ol.color` functions that accept an `ol.Color` (see #5871).

The change here makes it so the callback passed to `map.forEachLayerAtPixel()` gets a `Uint8ClampedArray` (or `Uint8Array` in old IE) instead of an `ol.Color`.  This lets us keep `ol.Color` as `Array<number>` (which makes the `ol.color` functions work again).

Fixes #5871.

### DIFF
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -603,11 +603,12 @@ ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this, opt
  * execute a callback with each matching layer. Layers included in the
  * detection can be configured through `opt_layerFilter`.
  * @param {ol.Pixel} pixel Pixel.
- * @param {function(this: S, ol.layer.Layer, ol.Color): T} callback Layer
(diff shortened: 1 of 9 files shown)


ahocevar:
Looks good. Please merge.

PR #5876, opened 2016-09-12
Use clone method in kml format

simonseyock:
Making use of the clone method in the kml format.

See https://github.com/openlayers/ol3/issues/4781#issuecomment-180082908. This is a follow up to #5832.


simonseyock:
@tschaub, @tamarmot: I changed createNameStyleFunction_ to use the new clone functionality

marcjansen:
This is looking good to me, thanks again, @simonseyock.

PR #5878, opened 2016-09-13
Smarter handling of closed rings and closePath() instructions

ahocevar:
This is a follow-up on #5826, making two optimisations when appending flat coordinates to a polygon replay:

 1. Do not repeat the last coordinate for polygon rings, because the flat coordinates already have the first coordinate repeated as last.
 1. Optionally skip the first coordinate when we're going to call closePath(), because that will take care of closing the ring.


ahocevar:
In addition to the above description, c944ce2 removes redundant checks for the existence of stroke and fill styles, and adds a debug assertion instead.

tschaub:
These sound like good changes.  Since our regressions around this have been exposed only when rendering, what do you think about adding some rendering tests?  In particular, a regression test against #5822 would be nice.

ahocevar:
Rendering tests added.

marcjansen:
Nice work, @ahocevar. Please merge.

PR #5882, opened 2016-09-14
ol3-echarts example

FreeGIS:
Example of using ol3 and echarts together,demo:http://freegis.github.io/examples/ol3-echarts.html


tschaub:

PR #5883, opened 2016-09-14
Fixes setUrl() for WMTS sources (ol.source.WMTS)

### LINKED ISSUE #5881: setUrls() on ol.source.WMTS causes Template URLs to change template placeholders
Hi,
A WMTS Source takes template URLs in the form `http://my-mapservice/1.0.0/{TileMatrix}/{TileRow}/{TileCol}` during constuction time:

```
var source = new ol.source.WMTS({
  /// .....
  urls: ['http://my-mapservice/1.0.0/{TileMatrix}/{TileRow}/{TileCol}']
});
```

During initialization, ol.source.WMTS uses the private `createFromWMTSTemplate()` function to parse and replace the URL template.

However, if you later re-set the URLs using the `setUrl` function, the wrong URL template function is set and you have to use `{x}`, `{y}`, `{z}` instead:

```
source.setUrls(['http://my-mapservice/1.0.0/{z}/{y}/{x}']);
```

It seems that this is caused during `setUrls` in `ol.source.UrlTile`, about here: https://github.com/openlayers/ol3/blob/v3.12.1/src/ol/source/urltilesource.js#L182

I think `setUrl()` needs to be overridden in ol.source.WMTS to fix this issue.

bylexus:
fixes #5881: setUrl was inherited from ol.source UrlTile, which creates a wrong tileUrlFunction: The official WMTS URL Template variables were no longer replaced.

This fix corrects this behaviour, hopefully not breaking anything. I created a minimal test set to cover this case.

### DIFF
--- a/src/ol/source/wmts.js
+++ b/src/ol/source/wmts.js
@@ -147,7 +147,7 @@ ol.source.WMTS = function(options) {
           return url;
         }
       });
-  }
+  };
 
   var tileUrlFunction = (urls && urls.length > 0) ?
     ol.TileUrlFunction.createFromTileUrlFunctions(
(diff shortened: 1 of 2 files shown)


marcjansen:
@bartvde, do you think you can have a look at this?

tschaub:
Can @bartvde or @fredj or anybody who uses WMTS give this a review?

fredj:
I've made some minor update in 90d3b6f, otherwise looks good to me

PR #5885, opened 2016-09-15
Remove unused pruneTileRange function

### LINKED ISSUE #5524: ol.TileCache.prototype.pruneTileRange function can be optimized
After removing item from cache "i" counter variable can be decreased by 1. Currently system doing unnecessary iteration for every removed item.

Proposed code:
```
ol.TileCache.prototype.pruneTileRange = function(tileRange) {
   var i = this.getCount(),
       key;
   while (i-- > 0) {
     key = this.peekLastKey();
     if (tileRange.contains(ol.tilecoord.createFromString(key))) {
       this.pop().dispose();
       i--;
     } else {
       this.get(key);
     }
   }
};
```

bartvde:
Closes #5524 thanks to @gzhebrunov for pointing this out.

### DIFF
--- a/src/ol/tilecache.js
+++ b/src/ol/tilecache.js
@@ -2,7 +2,6 @@ goog.provide('ol.TileCache');
 
 goog.require('ol');
 goog.require('ol.structs.LRUCache');
-goog.require('ol.tilecoord');
 
 
 /**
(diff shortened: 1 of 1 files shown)


bartvde:
I'll add a unit test.

bartvde:
@ahocevar seems this function is not used anymore in the library?

ahocevar:
Oh. Would be good to know since when it is unused. It seems to me that this function is important.

bartvde:
Unless I'm misreading things, it was not even used in 3.0.0, which means it was never used?

bartvde:
Can't find it used in any of the 3.x.0 releases

ahocevar:
Ok cool, so let's get rid of it.

fredj:
the function was added in #850 and removed in #1551

bartvde:
thanks @fredj

PR #5886, opened 2016-09-15
Remove unused functions

fredj:
Remove some unused and non api functions

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -38,38 +38,6 @@ ol.array.binarySearch = function(haystack, needle, opt_comparator) {
   return found ? low : ~low;
 };
 
-/**
- * @param {Array.<number>} arr Array.
- * @param {number} target Target.
- * @return {number} Index.
- */
-ol.array.binaryFindNearest = function(arr, target) {
-  var index = ol.array.binarySearch(arr, target,
-      /**
-       * @param {number} a A.
-       * @param {number} b B.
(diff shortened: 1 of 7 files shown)


ahocevar:
Out of curiousity, how do you identify these?

bartvde:
Right was wondering the same, is there a compiler option or something?

ahocevar:
Also, do we have unused API functions? Just asking because you mentioned "non api functions".

fredj:
> Out of curiousity, how do you identify these?

I just looked at `ol.array`, `ol.coordinate`, `ol.size` and `ol.TileRange`and `git grep`ed the (non api) functions name.

fredj:
> Also, do we have unused API functions? Just asking because you mentioned "non api functions".

Yes probably but I didn't check

fredj:
@ahocevar, @bartvde is it good to merge?

ahocevar:
Thanks @fredj, great work!

fredj:
Thanks for the review

ischas:
> Also, do we have unused API functions? Just asking because you mentioned "non api functions".

Out of curiosity, how will you identify these since you don't know which API functions are used by developers using OL3?

bartvde:
Indeed @ischas , we can only check which API functions are not used by the library itself, but still they can be used by applications. So no need to look for those IMHO.

PR #5890, opened 2016-09-15
No gaps between tiles of different resolutions

### LINKED ISSUE #5690: White lines for Zoomify source starting from 3.14.0
I noticed that in the last version 3.17.1 there is white lines are blinking when you fast zoom in or zoom out (in Chrome, FF, IE). And looks like it happens only for Zoomify sources. Such behavior has started from 3.14. 

To show you I uploaded some examples (custom code is the same, but different OL):
1) With white lines (using 3.17.1): http://demo.smartinmedia.com/openlayers/examples/issue17.html
2) Without (using 3.13.1): http://demo.smartinmedia.com/openlayers/examples/issue13.html

For now we decided to fallback to 3.13.1. But if you will have time - just have a look.

ahocevar:
When tiles of different resolutions are combined for a layer (which is the case before all tiles of the best resolution have been loaded), lower resolution tiles may not align with higher resolution tiles. Because OpenLayers supports arbitrary tile grids (not just grids where a tile is divided into 4 when increasing the zoom level by 1), we cannot solve the alignment problem with powers of 2 pyramids.

In the sketch below, we are looking at a map between zoom 11 and 12. The view is scaled by 0.69 during the zoom animation. When rounding to integers, a 256 pixel tile from zoom 12 will be 177 pixels. But a tile from zoom 11 will be 353 pixels, which is 1 less than 2*177. At each resolution, we stitch tiles with integer widths and heights together from a common origin so they align without gaps. In the sketch below, the origin for zoom 11 and 12 is the same and marked with `O` (but it could be different across resolutions in the current implementation). And because of rounding after scaling, tiles from zoom 12 will be 1 pixel short with each tile further away from the origin:
```
+---------+---------+                   +------
|         |         |----------------+  |
|         |         |                |  |
|         |         |    no higher   |  |
+---------+---------+    resolution  |  +------
|         |         |    available   |  |
|         |         |                |  |
|         |         |                |  |
+---------+---------O-------------------+------
|         |         |         |         |
|         |         |         |         |
```

To fix this issue, I took the following approach:

 1. Try to use the same origin across all resolutions. By using `ol.tilegrid.TileGrid#getTileCoordForCoordAndZ()` for the tile that contains the center at the lowest resolution, we get a coordinate that will also be a tile corner at higher resolutions for powers of 2 tile grids.
 1.  Calulate the tile size in pixels for the highest resolution tiles. By multiplying this size with the resolution factor (tile resolution divided by highest resolution) and taking into account possibly different tile sizes at different resolutions, we do not need to round tile sizes any more to get integer values for powers of 2 tile grids.

This approach won't change anything for tile grids that have no powers of 2 layout, but it will result in pixel perfect alignment for scaled tile sizes even across resolutions for powers of 2 tile grids.

Fixes #5690.

### DIFF
--- a/src/ol/renderer/canvas/tilelayer.js
+++ b/src/ol/renderer/canvas/tilelayer.js
@@ -12,6 +12,7 @@ goog.require('ol.extent');
 goog.require('ol.render.canvas');
 goog.require('ol.render.EventType');
 goog.require('ol.renderer.canvas.Layer');
+goog.require('ol.size');
 
 
 /**
@@ -214,6 +215,11 @@ ol.renderer.canvas.TileLayer.prototype.forEachLayerAtPixel = function(
  * @protected
  */
 ol.renderer.canvas.TileLayer.prototype.renderTileImages = function(context, frameState, layerState) {
+  var tilesToDraw = this.renderedTiles;
+  if (tilesToDraw.length == 0) {
+    return;
+  }
+
   var pixelRatio = frameState.pixelRatio;
   var viewState = frameState.viewState;
   var center = viewState.center;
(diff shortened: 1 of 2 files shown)


ahocevar:
The effect of this change can be seen best when zooming in and out fast and repeatedly in the bing-maps example. Before the change, there are gaps between tiles during zooming. After this change, there are none.

Thanks for any review.

ahocevar:
I'd really appreciate a review on this. I'll even buy the reviewer a beer at the next sprint or conference 😄

src/ol/renderer/canvas/tilelayer.js
```diff
@@ -222,6 +223,11 @@ ol.renderer.canvas.TileLayer.prototype.forEachLayerAtPixel = function(
  * @protected
  */
 ol.renderer.canvas.TileLayer.prototype.renderTileImages = function(context, frameState, layerState) {
+  var tilesToDraw = this.renderedTiles;
+  if (tilesToDraw.length == 0) {
+    return;
+  }
+
```
marcjansen: Unquestionable.

src/ol/renderer/canvas/tilelayer.js
```diff
@@ -266,7 +272,15 @@ ol.renderer.canvas.TileLayer.prototype.renderTileImages = function(context, fram
   var alpha = renderContext.globalAlpha;
   renderContext.globalAlpha = layerState.opacity;
 
-  var tilesToDraw = this.renderedTiles;
+  // Origin of the lowest resolution tile that contains the map center. We will
+  // try to use the same origin for all resolutions for pixel-perfect tile
+  // alignment across resolutions.
+  var minZOrigin = ol.extent.getBottomLeft(tileGrid.getTileCoordExtent(
+      tileGrid.getTileCoordForCoordAndZ(center,
+          tilesToDraw[0].getTileCoord()[0], this.tmpTileCoord_), this.tmpExtent));
```
marcjansen: I'd favor separate variable declarations for readability.

PR #5894, opened 2016-09-16
Reuse arrays for rendered tiles and extent

ahocevar:
This pull request suggests a few improvements to the tile layer renderer, by reusing arrays instead of recreating them with every frame.


tschaub:
Looks good to me too.

PR #5895, opened 2016-09-16
Always use the frameState's extent

ahocevar:
Since the tile renderer aligns tiles to pixels anyway, we do not need to
modify the extent to make its center align with a pixel.


fredj:
Is it possible to do the same with the [webgl renderer](https://github.com/openlayers/ol3/blob/v3.18.2/src/ol/renderer/webgl/tilelayer.js#L168-L174) ?

ahocevar:
Renderer tests fail when the same change is made to the WebGL renderer.

ahocevar:
@fredj I simplified things for the WebGL renderer too, and got rid of `ol.renderer.Layer#snapCenterToPixel`.

tschaub:
These look like good changes to me.

fredj:
Thanks @ahocevar

PR #5903, opened 2016-09-20
Rename ol.format.ogc.filter namespace to ol.format.filter

fredj:
See https://github.com/openlayers/ol3/issues/5877#issuecomment-246913962


fredj:
Thanks for the review Bart

PR #5904, opened 2016-09-20
Fix changelog/upgrade-notes.md formating

fredj:
See https://github.com/openlayers/ol3/blob/master/changelog/upgrade-notes.md#googdebug-define-was-renamed-to-oldebug


fredj:
trivial, merging

PR #5914, opened 2016-09-26
Allow clipping by MultiPolygon

tschaub:
Currently, the canvas immediate renderer calls `context.beginPath()` once for each composite part of a MultiPolygon.  This makes it so a MultiPolygon cannot be used to generate a clip path.  For example, here is a geometry representing two diamonds used as a clip path:

wrong

Only the last composite part is used in clipping.  Instead, all parts should be used.

This branch makes it so we only call `context.beginPath()` once per MultiPolygon (instead of once per composite part).  In addition, we only fill and stroke after drawing all parts - this means MultiPolygon parts should not be overlapping.

The tests show the corrected behavior when clipping by a MultiPolygon.


ahocevar:
Nice! Please merge when green.

PR #5917, opened 2016-09-27
Accept feature projection in GeoJSON constructor

### LINKED ISSUE #5687: Support featureProjection as a format option
It would be convenient if it were possible to construct a feature format with a `featureProjection` option.  So something like this would work:

``` js
var format = new ol.format.GeoJSON({featureProjection: 'EPSG:3857'});
var feature = format.readFeature(obj);
```

Options passed to the read/write methods would take precedence.

This would apply to:
- [ ] `ol.format.EsriJSON`
- [ ] `ol.format.GML`
- [ ] `ol.format.GPX`
- [x] `ol.format.GeoJSON` (see #5917)
- [ ] `ol.format.KML`
- [ ] `ol.format.MVT`
- [ ] `ol.format.Polyline`
- [ ] `ol.format.TopoJSON`
- [ ] `ol.format.WKT`

tschaub:
This adds support for a `featureProjection` option to the GeoJSON constructor.

See #5687.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -1666,7 +1666,8 @@ olx.format.WriteOptions.prototype.decimals;
 
 /**
  * @typedef {{defaultDataProjection: ol.ProjectionLike,
- *     geometryName: (string|undefined)}}
+ *     geometryName: (string|undefined),
+ *     featureProjection: ol.ProjectionLike}}
  */
 olx.format.GeoJSONOptions;
 
@@ -1677,6 +1678,15 @@ olx.format.GeoJSONOptions;
  * @api stable
  */
 olx.format.GeoJSONOptions.prototype.defaultDataProjection;
+
+
+/**
+ * Projection for features read or written by the format.  Options passed to
+ * read or write methods will take precedence.
+ * @type {ol.ProjectionLike}
+ * @api stable
+ */
+olx.format.GeoJSONOptions.prototype.featureProjection;
 
 
 /**
--- a/src/ol/format/feature.js
+++ b/src/ol/format/feature.js
@@ -1,6 +1,7 @@
 goog.provide('ol.format.Feature');
 
 goog.require('ol.geom.Geometry');
+goog.require('ol.obj');
 goog.require('ol.proj');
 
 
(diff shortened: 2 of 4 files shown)


src/ol/format/feature.js
```diff
@@ -64,19 +71,10 @@ ol.format.Feature.prototype.getReadOptions = function(source, opt_options) {
...
-    };
-    if (options.decimals) {
-      updatedOptions.decimals = options.decimals;
-    }
-  }
-  return updatedOptions;
+  return ol.obj.assign({
+    dataProjection: this.defaultDataProjection,
+    featureProjection: this.defaultFeatureProjection
+  }, options);
```
marcjansen: Where did `rightHanded` and `decimals` go?

externs/olx.js
```diff
@@ -1680,6 +1681,15 @@ olx.format.GeoJSONOptions.prototype.defaultDataProjection;
 
 
 /**
+ * Projection for features read or written by the format.  Options passed to
+ * read or write methods will take precedence.
+ * @type {ol.ProjectionLike}
+ * @api stable
```
fredj: do we want to mark this option stable?

src/ol/format/feature.js
tschaub: If `rightHanded` or `decimals` are provided in `option`, these will be applied to the return.

externs/olx.js
tschaub: I think it makes sense to call this stable.  I'll merge as is, and if there is disagreement, we can change it before the release.

tschaub:

src/ol/format/feature.js
marcjansen: Ahh yes, I somehow didn't see the `options` being passed in. Thanks for the clarification, @tschaub.

bartvde:
@tschaub locally tests fail with make check for me:

```
  1) ol.format.GeoJSON #writeGeometry accepts featureProjection:

      expected [ 10, 19.999999999999986 ] to sort of equal [ 10, 20 ]
      + expected - actual

       [
         10
      -  19.999999999999986
      +  20
       ]
      
```

```
  2) ol.format.GeoJSON #writeGeometry respects featureProjection passed to constructor:

      expected [ 10, 19.999999999999986 ] to sort of equal [ 10, 20 ]
      + expected - actual

       [
         10
      -  19.999999999999986
      +  20
       ]

```

thhomas:
I get the same error in make test.

PR #5924, opened 2016-09-30
Add missing goog.provide for several components

adube:
This PR adds a missing provide for:
- ol.control.ScaleLines.Units
- ol.Overlay.Positioning
- ol.format.IGC.Z
- ol.source.WMTS.RequestEncoding
- ol.style.Icon.AnchorUnits
- ol.style.Icon.Origin
- ol.layer.VectorTile.RenderType
- ol.Tile.State

With or without this, the `build` works just fine.  However, when trying to use Closure-util unless we define those missing provide we get errors like these:

```
ERR! compile node_modules/openlayers/externs/olx.js:1263: ERROR - Bad type annotation. Unknown type ol.control.ScaleLine.Units
ERR! compile  *     units: (ol.control.ScaleLine.Units|string|undefined)}}
ERR! compile                ^
ERR! compile 
ERR! compile 
ERR! compile node_modules/openlayers/externs/olx.js:1303: ERROR - Bad type annotation. Unknown type ol.control.ScaleLine.Units
ERR! compile  * @type {ol.control.ScaleLine.Units|string|undefined}
ERR! compile           ^
ERR! compile 
ERR! compile 
ERR! compile 2 error(s), 0
ERR! compile  warning(s)
```

Ready for review.

### DIFF
--- a/src/ol/control/scaleline.js
+++ b/src/ol/control/scaleline.js
@@ -1,4 +1,5 @@
 goog.provide('ol.control.ScaleLine');
+goog.provide('ol.control.ScaleLine.Units');
 
 goog.require('ol');
 goog.require('ol.Object');
--- a/src/ol/format/igc.js
+++ b/src/ol/format/igc.js
@@ -1,4 +1,5 @@
 goog.provide('ol.format.IGC');
+goog.provide('ol.format.IGC.Z');
 
 goog.require('ol');
 goog.require('ol.Feature');
--- a/src/ol/layer/vectortile.js
+++ b/src/ol/layer/vectortile.js
@@ -1,4 +1,5 @@
 goog.provide('ol.layer.VectorTile');
+goog.provide('ol.layer.VectorTile.RenderType');
 
 goog.require('ol');
 goog.require('ol.asserts');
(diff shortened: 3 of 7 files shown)


adube:
There are more errors related to the same kind of issue, i.e. missing require.  I'll update the PR.

ahocevar:
Only one provide per file please. See https://github.com/openlayers/ol3/pull/5702#issuecomment-238585089.

ahocevar:
But before you update the pull request, let's decide how to deal with enums in general. The problem is obviously the `externs/olx.js` file which depends on enums and typedefs. typedefs are now in a separate file, but the difference with enums is that they are actual code.

To be honest, I don't have an idea, other than always throwing all files at the compiler and set `manage_closure_dependencies` to true. I think this was suggested elsewhere already.

tschaub:
Since enums are real code, we could put them in their own files.  But people might get tired of this.

I think throwing all files at the compiler and always setting `manage_closure_dependencies` to **`false`** is the way to go (if `true`, then types that aren't explicitly required don't get used by the compiler and it breaks).

ahocevar:
Thanks for the explanation @tschaub. What changes would be required (and where?) to throw all files at the compiler?

fredj:
@tschaub fixed with #6207 ?

tschaub:
Let's call this fixed with #6207.  @adube please file a new issue if you are still getting errors (with our build script).

We now enforce one provide per file.

adube:
Everything seems fine on my side.  Thank you very much for your work on this.

PR #5925, opened 2016-09-30
Fix closure tutorial, missing typedefs.js to include

adube:
This PR fixes the "Closure" tutorial.  With the addition of `typedefs.js` in OpenLayers, the proposed configuration no longer worked.  You need to add `typedefs.js` in the list of JavaScript files to include.  The reason why is that this file doesn't have any `goog.provide`, thus isn't theorically required to do the build but it is.  You need to tell closure that.  In ol3, it's added in the build.js script.  We need to to the same for the tutorial.

Ready for review.

### DIFF
--- a/doc/tutorials/closure.md
+++ b/doc/tutorials/closure.md
@@ -173,6 +173,7 @@ The minimum config file looks like this:
       "ol.ENABLE_WEBGL=false"
     ],
     "js": [
+      "node_modules/openlayers/src/ol/typedefs.js",
       "node_modules/openlayers/externs/olx.js",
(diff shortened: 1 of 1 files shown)


adube:
Question related to this (but out of scope of this PR) - is there a reason why this file has no provide?

probins:
because they're not needed :-) But you're right that this tutorial should add the file manually, as it does not use `build.js` which includes it automatically.

adube:
Thanks.

I think this is safe for merge, then.  I don't have merge access.

fredj:
fixed with #6704

PR #5927, opened 2016-09-30
Ensure ol.VectorTile type

adube:
This PR makes sure that the type of tile object within the `ol.renderer.canvas.VectorTileLayer` are properly cast to `ol.VectorTile`.  OpenLayers' build script works just fine with or without these cast, but when using Closure-util we get errors like:

```
ERR! compile node_modules/openlayers/src/ol/renderer/canvas/vectortilelayer.js:312:
  ERROR - Property getReplayState never defined on ol.Tile
ERR! compile     replayGroup = tile.getReplayState().replayGroup;
```

Ready for review (and if you followed the couple last PRs I just made you'll see that I'm confused as why the OL3 build doesn't complain about those where the build in Closure-util does).


ahocevar:
I have raised the same concern earlier too. And strange enough, when I create a custom build, I only get warnings about those, and not errors.

ahocevar:
Is this still relevant? At least when #6779 is merged, it shouldn't.

fredj:
closing, please reopen if the issue persist

PR #5935, opened 2016-10-05
Keep transformed coordinates within valid y range

tschaub:
This clamps y-values in forward EPSG:3857 transforms.  Without this change, if people don't modify their data before rendering it, they face hard to debug problems.


fredj:
See also https://github.com/openlayers/ol3/issues/4715#issuecomment-184210214 (ping @bartvde)

tschaub:
By not clamping y-values in our forward transform (the one everybody uses), we end up with confusingly bad results.  For example, here is [a world-spanning polygon with a star shaped hole](http://bl.ocks.org/anonymous/raw/d27b206e3b4d2f2664365fb340645af2/):

image

And when rendered in OpenLayers:

image

(Here, the interior ring is incorrectly filled.)

marcjansen:
Thanks for this, Tim. Please merge.

bartvde:
Post merge thank you! Great to see this handled @tschaub

PR #5944, opened 2016-10-10
Allow readFeature and readFeatures to be called with GeoJSON geometries

tschaub:
We allow `format.readFeatures()` to be called with a GeoJSON Feature or FeatureCollection.  It would be convenient to also allow it to be called with geometries.  For example, when writing an application with drag-and-drop support, it would be nice not to inspect the dropped data, switch on the type, and call different methods.

### DIFF
--- a/test/spec/ol/format/geojson.test.js
+++ b/test/spec/ol/format/geojson.test.js
@@ -154,6 +154,14 @@ describe('ol.format.GeoJSON', function() {
       expect(feature.get('prop0')).to.be('value0');
     });
 
+    it('can read a single point geometry as a feature feature', function() {
+      var feature = format.readFeature(pointGeoJSON.geometry);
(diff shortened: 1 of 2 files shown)


test/spec/ol/format/geojson.test.js
```diff
@@ -154,6 +154,14 @@ describe('ol.format.GeoJSON', function() {
       expect(feature.get('prop0')).to.be('value0');
     });
 
+    it('can read a single point geometry as a feature feature', function() {
```
marcjansen: Double `feature`?

tschaub: Good catch.  Addressed in 0d686fe88b99d9d4baa017020ecbce4c0c8d5c11.

marcjansen: :+1:

PR #5948, opened 2016-10-11
Fix API docs of ol.Feature#getStyle

marcjansen:
The `ol.Feature#getStyle` prose doc was:

> Get the feature's style. This return for this method depends on what was provided …

This PR suggests to reword this as follows:

> Get the feature's style. Will return what was provided …

Please review.


marcjansen:
Thanks for the reviews.

PR #5950, opened 2016-10-11
Move gradient and pattern fills with the view

ahocevar:
Opened for discussion.

Currently, when using a gradient or pattern to fill polygons, gradient and pattern won't change their position when panning/rotating the map. See https://openlayers.org/en/master/examples/canvas-gradient-pattern.html.

To change this, I have added a `moveWithView` option to `ol.style.Fill`. When set to true, fills will move and rotate with the view. This is especially useful for pattern fills, as can be seen here:

`moveWithView: false`:
oct-11-2016 18-09-52

`moveWithView: true`:
oct-11-2016 18-12-29

Any opinions on whether this is the way to go, or if we should solve this in a different way?

### DIFF
--- a/examples/canvas-gradient-pattern.js
+++ b/examples/canvas-gradient-pattern.js
@@ -1,6 +1,7 @@
 goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.format.GeoJSON');
+goog.require('ol.has');
 goog.require('ol.layer.Vector');
 goog.require('ol.proj');
 goog.require('ol.source.Vector');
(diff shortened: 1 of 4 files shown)


marcjansen:
I love the effect! And the name makes sense as does the place where you suggest to set it. I like it!

marcjansen:
And can we somehow get the same with the gradient or am I missing sth.?

ahocevar:
> And can we somehow get the same with the gradient or am I missing sth.?

That would require more changes to the example (different start and end coordinates for the gradient), but should also work with the new option. I did not apply it to the gradient because I don't see much use in filling polygons with a gradient.

bartvde:
I wonder if the current behaviour is needed at all, or if we should just make this the default?

marcjansen:
Generally I agree, Bart. Why not add the option and default it to true? That would be a behavioural change but easily revertable by our users. Though I doubt many actually use this feature.

tschaub:
I'd be tempted to call it a bug fix (instead of exposing an option).

ahocevar:
I updated the pull request according to the suggestions and made the new behavior the default, without a new option to change to the previous behavior. I also changed the example to use a per-feature gradient instead of a global one, which probably makes more sense:

oct-12-2016 18-16-38

marcjansen:
Very nice, especially the pixelRatio usage in the example. The explanation there could probably have one sentence about that, but I'll leave it up to you.

examples/canvas-gradient-pattern.js
```diff
@@ -24,67 +28,50 @@ var gradient = (function() {
...
 var pattern = (function() {
-  var canvas = document.createElement('canvas');
-  var context = canvas.getContext('2d');
-  canvas.width = 11;
-  canvas.height = 11;
+  canvas.width = 11 * pixelRatio;
+  canvas.height = 11 * pixelRatio;
+  // white background
+  context.fillStyle = 'white';
+  context.fillRect(0, 0, 11 * pixelRatio, 11 * pixelRatio);
```
bartvde: minor but could use canvas.width and canvas.height here

externs/olx.js
```diff
@@ -6645,6 +6645,14 @@ olx.style.FillOptions.prototype.color;
 
 
 /**
+ * Move `CanvasGradient` or `CanvasPattern` polygon fills with the view when
+ * panning or rotating. Default is `false`.
+ * @type {boolean}
+ */
+olx.style.FillOptions.prototype.moveWithView;
```
bartvde: not needed anymore right?

src/ol/renderer/canvas/tilelayer.js
```diff
@@ -208,15 +208,18 @@ ol.renderer.canvas.TileLayer.prototype.forEachLayerAtPixel = function(
 };
 
 
+ol.renderer.canvas.TileLayer.prototype.renderTileImage = function(replayContext, frameState, tile, tileGrid, tilePixelRatio, skippedFeatureUids) {
+
+};
```
bartvde: why is this empty function needed?

externs/olx.js
ahocevar: Removed.

examples/canvas-gradient-pattern.js
ahocevar: Changed.

ahocevar:
Thanks for the comments and reviews!

PR #5954, opened 2016-10-12
Produce sharp tiles with the WebGL renderer

GaborFarkas:
Commit 3026fda successfully removes ol.renderer.Layer#snapCenterToPixel and pixel rounds the center in every case. However, as far as I can currently understand, one step is missing from the rounding formula. This commit attempts to fix this, and produce sharp tiles, if the view is not rotated. These are the minimal changes to render sharp tiles by default, and therefore I do not know, if it has any ill-effects in other circumstances.

Before:
blurry

After:
sharp


tschaub:

GaborFarkas:
Fine by me. I'll look into it after 4.6. Thank you for the notice.

PR #5955, opened 2016-10-12
Add getters to BingMaps source for api key and imagerySet

bartvde:
These options can be passed into the source constructor, but cannot be retrieved after construction. This is useful for serialising state.


bartvde:
thanks @fredj

PR #5959, opened 2016-10-13
Remove unnecessary and misleading assertion

### LINKED ISSUE #5958: ol.interaction.Select#getLayer() should not assert that the passed feature is an ol.Feature
The `getLayer()` method should work with `ol.render.Feature` just as fine. So the assertion should be removed.

ahocevar:
Fixes #5958.

### DIFF
--- a/src/ol/interaction/select.js
+++ b/src/ol/interaction/select.js
@@ -1,10 +1,8 @@
 goog.provide('ol.interaction.Select');
 
 goog.require('ol');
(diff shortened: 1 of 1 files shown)


probins:
I think this entry can be removed from `doc/errors/index.md` as well, right?

ahocevar:
@probins I saw your comment too late, but you are absolutely right.

PR #5962, opened 2016-10-13
Use same proj4js version in examples and tests

ahocevar:
The version was updated in package.json, but not in the examples. This pull request makes it so the same version is used everywhere.


fredj:
+1, thanks

PR #5963, opened 2016-10-14
Remove the vertex on ol.interaction.Modify deactivation

### LINKED ISSUE #5938: ol3 - adding kml feature via string causes exception
a) kml webservice when provided as url directly to ol.source.Vector and rendered works.
var vectorKMLLayer = new ol.layer.Vector({
	  url: 'http://localhost:8080/TrackServiceBTS/webresources/track/getTrackData/BTS02/BTSDIR01',
	  format: new ol.format.KML()
      })
  });

b) when the data from the service is retrieved into string for filtering, and feature added to the layer individually exception is returned:
                var kmlSource = (new ol.format.KML()).readFeatures(data["value"]);
                var vectorKMLSource = (new ol.source.Vector()).addFeatures(kmlSource);
                var vectorKMLLayer = new ol.layer.Vector().setSource(vectorKMLSource);
                -- below results in exception:
                workerView.options.mapView.addLayer(vectorKMLLayer);

ol-debug.js:3026 Uncaught TypeError: Cannot read property 'ol_uid' of undefinedol.getUid @ 
ol-debug.js:3026
ol.layer.Group.handleLayersAdd_ @ ol-debug.js:28743
boundListener @ ol-debug.js:3462
ol.events.EventTarget.dispatchEvent @ ol-debug.js:3880
ol.Collection.insertAt @ ol-debug.js:12318
ol.Collection.push @ ol-debug.js:12342
ol.Map.addLayer @ ol-debug.js:45278
(anonymous function) @ workerView.js:270
m.each.m.forEach @ underscore-min.js:5
onWorkerNotification @ workerView.js:233

Kml is valid since the url method renders it fine, looks like the interactive method causes event to be fired - could not find in API how to turn-off the event notification when a feature is added.

### LINKED ISSUE #5953: ol.interaction.Modify.setActive(active)
I have a feature with LineString as a geometry. I defined the modify interaction on this feature. I turn this interaction on and off according to zoom level of the map view. For this purpose I use ol.View.on("change:resolution", handler).  Everything works fine until the moment when mouse cursor is over the linestring and interaction is active. The interaction style  is visibly applied, and then, without moving the cursor i only zoom out and disable the interaction by setActive(false). What happens is that the interaction is actually disabled as expected, but the interaction style remains visible and there is no way how to get rid off it other than to zoom again to set the interaction active. The similar but opposite problem is when the interaction is inactive, the mouse cursor is over the LineString and you zoom in so that the interaction becomes active. The interaction style is not visible until you move the cursor.

http://jsfiddle.net/xjpmcsd4/

fredj:
fixes #5953

### DIFF
--- a/src/ol/interaction/modify.js
+++ b/src/ol/interaction/modify.js
@@ -261,6 +261,18 @@ ol.interaction.Modify.prototype.removeFeatureSegmentData_ = function(feature) {
 /**
  * @inheritDoc
  */
+ol.interaction.Modify.prototype.setActive = function(active) {
+  if (this.vertexFeature_ && !active) {
+    this.overlay_.getSource().removeFeature(this.vertexFeature_);
+    this.vertexFeature_ = null;
+  }
+  ol.interaction.Pointer.prototype.setActive.call(this, active);
+};
+
+
+/**
+ * @inheritDoc
+ */
 ol.interaction.Modify.prototype.setMap = function(map) {
   this.overlay_.setMap(map);
   ol.interaction.Pointer.prototype.setMap.call(this, map);
--- a/test/spec/ol/interaction/modify.test.js
+++ b/test/spec/ol/interaction/modify.test.js
@@ -534,4 +534,20 @@ describe('ol.interaction.Modify', function() {
     });
   });
 
+  describe('#setActive', function() {
+    it('removes the vertexFeature of deactivation', function() {
+      var modify = new ol.interaction.Modify({
+        features: new ol.Collection(features)
+      });
+      map.addInteraction(modify);
+      expect(modify.vertexFeature_).to.be(null);
+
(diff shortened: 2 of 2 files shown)


fredj:
tests added, thanks for any review

src/ol/interaction/modify.js
```diff
@@ -184,6 +184,10 @@ ol.interaction.Modify = function(options) {
    */
   this.lastPointerEvent_ = null;
 
+  ol.events.listen(this,
```
ahocevar: Wouldn't it be better to override the `setActive()` method, instead of registering a listener here?

fredj: you're right

fredj:
Thanks for the review

PR #5964, opened 2016-10-14
Set one-off transform instead of transforming back and forth

### LINKED ISSUE #5940: Error with map.forEachFeatureAtPixel() on Safari
Hi

I experience an error when trying to detect which features I hover over (or click) in a vector layer. The error only happens in Safari, not e.g. Chrome, Firefox or Edge, and only if some of the features are displayed using scaled icons (no error if icons are not scaled).

Basically, after a while the call to map.forEachFeatureAtPixel() will start returning too many features, and once OpenLayers has entered this bogus state, it will continue reporting false positives.

For verification, please use the following JSFiddle test with Safari:
https://jsfiddle.net/peder280370/s2myue71/3/
Initially, when you hover over the three features in turn, the correct one will be listed in the white panel, bottom-left. However, if you keep mousing over the three features, then at some stage OpenLayers will start wrongly reporting two matches at a time, and keep on doing that.

Versions: OpenLayers 3.18.2, Safari 10.0 on macOS Sierra.

Thanks,
- Peder

ahocevar:
Fixes #5940.

### DIFF
--- a/src/ol/render/canvas/replay.js
+++ b/src/ol/render/canvas/replay.js
@@ -116,7 +116,7 @@ ol.render.canvas.Replay = function(tolerance, maxExtent, resolution, overlaps) {
    * @private
    * @type {ol.Transform}
    */
-  this.tmpLocalTransformInv_ = ol.transform.create();
+  this.resetTransform_ = ol.transform.create();
 };
 ol.inherits(ol.render.canvas.Replay, ol.render.VectorContext);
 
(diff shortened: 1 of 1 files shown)


ahocevar:
Not sure if this is the best fix though, because it has a significant performance impact.

ahocevar:
Updated pull request to use `setTransform` instead. This fixes the issue, and should even improve performance a bit.

tschaub:
Code looks good (hoping we've got rendering test coverage here).

PR #5334, opened 2016-05-12
Load the examples resources with HTTPS

fredj:
If available, load the examples resources with HTTPS instead of HTTP.

Some domains doesn't provide HTTPS:
 - demo.boundlessgeo.com
 - tile.opencyclemap.org
 - tile.openstreetmap.us
 - tiles.openseamap.org

Issue #5276, opened 2016-04-27
geolocation and device orientation examples broken


fredj:
The examples are broken on Chrome >= 50 because they are running on insecure origin.
To fix this issue, http://openlayers.org/ must run on HTTPS

ahocevar:
See isaacs/github#156.

dddesign:
I updated my android to 6.0.1 and geolocation is not working since then (chrome 50.0.2661.89, I don't know the version before). it's not working for the latest geolocation example on openlayers.org and also not on openlayer v3.11.1 which I use for my webapp rraaddaarr.com/firepit
do I have to run all on https (so I need to download all js of openlayers etc. to my server)? or is there another way or a bug fix in ol js?

ischas:
This issue is independent from OL. Chrome doesn't give geolocation information to non-HTTPS-applications anymore. There is no fix but switching to HTTPS.

dddesign:
I see, thanks.
So I switched all to HTTPS (with let's encrypt cert).
It would be great if openlayers.org could be accessibly by HTTPS to get js and css.

ischas:
How about this CDN: https://cdnjs.com/libraries/ol3

probins:
I don't think this issue should be closed. I'm not entirely convinced by 'https everywhere', but Google and others are, and are gradually introducing restrictions like this. So, if Github Pages can't/won't support https fully, then it would be better either to move openlayers.org elsewhere (I'm currently experimenting with Gitlab Pages), or to use a front-end like Cloudflare.

ahocevar:
@probins Sorry, I thought this was a duplicate, but realised just now that it isn't. Reopening.

bartvde:
https://github.com/blog/2186-https-for-github-pages

probins:
I don't think that helps @bartvde because (AFAIK) it only covers the subdomains of github.io, not custom domains

ahocevar:
Correct. See https://help.github.com/articles/securing-your-github-pages-site-with-https/: "HTTPS is not supported for GitHub Pages using custom domains."

Screenshot from our site's settings page:
<img width="634" alt="untitled" src="https://cloud.githubusercontent.com/assets/211514/15907397/5584bbe2-2dbd-11e6-9d0c-618e3cdc4a2b.png">

bartvde:
ah hadn't seen that

Issue #3768, opened 2015-06-07
OpenLayers 3.6.0 jsfiddle and copy examples codes are not working


cholmes:
Just got confused by this for a bit. Root reason seems to be that they give links to https://cdnjs.cloudflare.com/ajax/libs/ol3/3.6.0/ol.js and https://cdnjs.cloudflare.com/ajax/libs/ol3/3.6.0/ol.css But cloudflare doesn't have 3.6.0 (yet?). The 3.5.0 examples work fine.

ahocevar:
Unfortunately we cannot trigger a cdnjs build, and it takes up to 24 hours for the new version to be available on cdnjs. This is why I did not announce the release on Twitter yet.

I think next time we'll wait with updating the website until the new version is available on cdnjs.

tschaub:
An alternative would be to just use the openlayers.org hosted builds.
```
$ curl -w @curl-format.txt -o /dev/null -s http://openlayers.org/en/v3.5.0/build/ol.js
    time_namelookup:  2.537
       time_connect:  2.542
    time_appconnect:  0.000
   time_pretransfer:  2.542
      time_redirect:  0.000
 time_starttransfer:  9.049
                    ----------
         time_total:  9.049
```

Compare to the cloudflare time:
```
$ curl -w @curl-format.txt -o /dev/null -s https://cdnjs.cloudflare.com/ajax/libs/ol3/3.5.0/ol.js
    time_namelookup:  2.541
       time_connect:  4.569
    time_appconnect:  7.986
   time_pretransfer:  7.986
      time_redirect:  0.000
 time_starttransfer:  10.240
                    ----------
         time_total:  11.308
```
(These with `curl-format.txt` from [this post](https://josephscott.org/archives/2011/10/timing-details-with-curl/).)

(And you can now see why I don't work from home.)

probins:
this is now working. I'd agree with @ahocevar that it would be better to hold off updating the website until the new cdnjs version is available.

bartvde:
if so, we should add this to the release procedure

probins:
I noticed that this new procedure was followed with 3.7. I think this issue can be closed

bartvde:
Agreed. I just added a note to the release procedure.

PR #5966, opened 2016-10-14
Change JSFiddle OL examples files to 'https'

### LINKED ISSUE #5250: JSFiddle examples broken
If you go to [one of the hosted examples](http://openlayers.org/en/v3.15.1/examples/simple.html) and click "edit" you get taken to JSFiddle but the example doesn't work.

In the console are two errors related to http://jsfiddle.net/css//css/light.css?update_19_04_2016_1 not being found.  And one uncaught `ReferenceError: ol is not defined`.

jonataswalker:
This PR is to avoid "Mixed Content" since #5334 loads JSFiddle with HTTPS.

I guess this fixes #5250.

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -66,7 +66,7 @@
           <textarea class="hidden" name="css">{{ css.source }}</textarea>
           <textarea class="hidden" name="html">{{ contents }}</textarea>
           <input type="hidden" name="wrap" value="l">
-          <input type="hidden" name="resources" value="http://openlayers.org/en/v{{ olVersion }}/css/ol.css,http://openlayers.org/en/v{{ olVersion }}/build/ol.js{{ extraResources }}">
+          <input type="hidden" name="resources" value="https://cdnjs.cloudflare.com/ajax/libs/ol3/{{ olVersion }}/ol.css,https://cdnjs.cloudflare.com/ajax/libs/ol3/{{ olVersion }}/ol.js{{ extraResources }}">
         </form>
         <pre><code id="example-source" class="language-markup">&lt;!DOCTYPE html&gt;
 &lt;html&gt;


fredj:
openlayers.org can't be loaded in https.

See also #5276 and https://github.com/openlayers/openlayers.github.io/issues/68

jonataswalker:
I (somehow) had the impression that would be possible. Let's keep this PR opened until it's possible?

probins:
one way around this would be to load the OL resources in JSFiddle from cdnjs which should work with https. That doesn't help with those examples which don't work on openlayers.org, but it should fix the issue with JSFiddle

jonataswalker:
Not sure which examples shouldn't work with cdnjs files, but if you approve I will change to cdnjs.

@probins @fredj

probins:
all the examples should work on JSFiddle if line 69 loads ol.css and ol.js from cdnjs https instead of from openlayers.org. Try it and see :-)

jonataswalker:
There you are. Changed to https://cdnjs.com/libraries/ol3

ThomasG77:
Another alternative that could be useful is unpkg.com.
IMO, it's not better than cdnjs but just avoid to make a PR on the cdnjs repo.

unpkg.com helps you serving any published npm package with https.

For OpenLayers, it could be https://unpkg.com/openlayers@3.18.2/dist/ol.js and https://unpkg.com/openlayers@3.18.2/dist/ol.css

jonataswalker:
Good to know @ThomasG77 but as far as I know it's not necessary a PR on cdnjs.

tschaub:
The problem we had with this approach in the past was that examples failed before the full build was available on cdnjs.  See #3768.

tschaub:
I'll look into https for openlayers.org.

jonataswalker:
Then I'm afraid this is a dead end after #5334.

Unless:
> I'll look into https for openlayers.org

probins:
there is no need for a PR for cdnjs, but, as noted in #3768, their auto-update only runs once a day or so. That would be an advantage of using unpkg.com: those files should be available as soon as published on npm (minus a small delay the first time it's accessed). Being able to use https on openlayers.org would be the better all-round solution, but unpkg.com would solve the specific issue of having usable files on JSFiddle.

tschaub:
I've moved our DNS records to CloudFlare.  So we can get SSL for openlayers.org.  Just waiting to get access to the registrar to update nameservers.

tschaub:
Waiting for name server changes to propagate.  After confirming that all the records work as expected, I'll set up a rule to redirect http to https.  Then this pull request can be changed to make the examples use https://openlayers.org instead of http://openlayers.org.  And we can do a global search/replace for the same in the openlayers/openlayers.github.io repo.

tschaub:
Things are looking good from where I sit.  I've set up the redirect - so http://openlayers.org bumps you to https://openlayers.org.  I'll replace all http://openlayers.org with https://openlayers.org in the openlayers/openlayers.github.io repo now.

ahocevar:
Thanks for your effort on this @tschaub.

tschaub:
I've temporarily turned off the rule to redirect from http to https.  We've got a number of examples to fix up first.  #5976 addresses most of these, and the same changes will need to be made in the openlayers.github.io repo.

tschaub:
Thanks for the motivation to get this fixed @jonataswalker.  The changes in #5976 should make the examples work with https on JSFiddle (except for the ones that still have mixed content).

PR #5970, opened 2016-10-14
Allow circles, squares, and boxes to be drawn in freehand mode

tschaub:
This makes it so the draw interaction can be used to draw circles, squares, and boxes in "freehand" mode.  For example, to draw a circle you press down, drag, and release.  This conforms with the drawing behavior people encounter in other tools.

### DIFF
--- a/src/ol/interaction/draw.js
+++ b/src/ol/interaction/draw.js
@@ -292,7 +292,8 @@ ol.interaction.Draw.prototype.setMap = function(map) {
  */
 ol.interaction.Draw.handleEvent = function(mapBrowserEvent) {
   if ((this.mode_ === ol.interaction.Draw.Mode.LINE_STRING ||
-    this.mode_ === ol.interaction.Draw.Mode.POLYGON) &&
+    this.mode_ === ol.interaction.Draw.Mode.POLYGON ||
+    this.mode_ === ol.interaction.Draw.Mode.CIRCLE) &&
     this.freehandCondition_(mapBrowserEvent)) {
     this.freehand_ = true;
   }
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks for the review @ahocevar.  I also wanted to make it easy to enable freehand mode without a modifier key.  The change in 139c88820d91684f79a1bf8f0d71c2218c5bea52 makes this possible.

Previously, using `ol.events.condition.always` would not work because the check for freehand mode in the `pointerdown` handler [would never be reached](https://github.com/openlayers/ol3/blob/a80d073da9999a4e74c4e9f45f81ab01e2fbd60f/src/ol/interaction/draw.js#L320-L333) if there was no modifier key (we first checked `ol.events.condition.noModifierKeys`).  So, the new behavior is to first check if we're in freehand mode and if not only handle down if the main condition is met (no modifier by default).

As a convenience, I added a `freehand` option to the interaction.  This makes it so people don't have to know about and include `ol.events.condition.always` in their builds.

tschaub:
I'm tempted to make it so `pointerup` ends drawing if in freehand mode (for all shape types).  This is how circle drawing now works (as well as box and square in the draw features example).  Line and polygon drawing still require clicking the last point to finish.

tschaub:
I've removed 139c88820d91684f79a1bf8f0d71c2218c5bea52 from this branch.  I'll open up a separate PR with that change.  I'll also separately address the issue about finishing in freehand mode.

ahocevar:
> I'm tempted to make it so pointerup ends drawing if in freehand mode (for all shape types).

I'd be in favor of this @tschaub

tschaub:
> I'd be in favor of this @tschaub

I'll create a PR for it.

tschaub:
The change in #5975 makes it so drawing starts on `pointerdown` and ends on `pointerup` while in freehand mode.

PR #5971, opened 2016-10-14
Add a reference to the French translation of the workshop

ThomasG77:
The link has been changed with https://github.com/openlayers/ol3/commit/ede6fc6e013a3ebf74ad31034dff9161b945be79 as we assumed most readers are speaking English but it completely hides the French version.

This PR adds a sentence for at least mention the availability of the French version of the workshop.

### DIFF
--- a/doc/index.hbs
+++ b/doc/index.hbs
@@ -9,7 +9,7 @@ If you're eager to get your first OpenLayers 3 map on a page, dive into the [qui
 
 For a more in-depth overview of OpenLayers 3 core concepts, check out the [tutorials](tutorials/).
 
-Make sure to also check out the [OpenLayers 3 workshop](/workshop/en/).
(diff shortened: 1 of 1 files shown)


bartvde:
Should we link to the French version here as well in your sentence @ThomasG77 ?

Alternatively I'm okay with reverting my change that you reference.

ThomasG77:
This PR is about making both translations visible to end users IMO.
Speaking about French users to English speaking people is not the best solution IMO.
It was just a way to adapt to your change.
Reverting would be better if OK for the others contributors.

ahocevar:
This pull request does not make much sense the way things are structured now. Reverting is probably the better option for now.

ThomasG77:
Waiting for  the revert to close the PR.
Thanks @bartvde & @ahocevar for your feedback

ThomasG77:
Thanks

PR #5976, opened 2016-10-16
Use https where available

tschaub:
This makes use of https where available.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -77,7 +77,7 @@ Then, just point your browser <http://localhost:3000/build/examples> in your bro
 Run examples against the `ol.js` standalone build:
 
 The examples can also be run against the `ol.js` standalone build, just like
-the examples [hosted](http://openlayers.org/en/master/examples/) on GitHub.
+the examples [hosted](https://openlayers.org/en/master/examples/) on GitHub.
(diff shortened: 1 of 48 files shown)


tschaub:
As mentioned in https://github.com/openlayers/ol3/pull/5966#issuecomment-254031540, I've turned off the redirect from http to https until we get the mixed content examples fixed up.  Most of the remaining ones are just warnings, but there are a few that fail (e.g. https://openlayers.org/en/master/examples/vector-wfs-getfeature.html).

tschaub:
As we find more URLs to update, we need to make the same changes to the `master` branch of [openlayers.github.io](https://github.com/openlayers/openlayers.github.io).

Here is a (potentially dangerous) script that I have been using:
```bash
DOMAIN=${1}

find en -type f -path '*/examples/*.js' -exec sed -i '' "s/http:\/\/${DOMAIN}/https:\/\/${DOMAIN}/" {} + && \
    git add en && \
    git commit -m "Use https for ${DOMAIN}" && \
    git push origin master
```

With that, you can update the website examples with something like `./use-https epsg.io`.

ahocevar:
@tschaub It should be safe to turn on https redirection now. As far as I can tell, all examples are fixed.

PR #5977, opened 2016-10-16
Convenience function for drawing boxes

tschaub:
This adds a `geometryFunction` that can be used with the `ol.interaction.Draw` to draw box-shaped polygons.

In addition, I split draw features example into two examples: one showing drawing standard geometry types and one using the `geometryFunction` option to draw custom shapes.  And a third example demonstrates the draw interaction in freehand mode.

### DIFF
--- a/examples/draw-features.html
+++ b/examples/draw-features.html
@@ -5,11 +5,8 @@ shortdesc: Example of using the ol.interaction.Draw interaction.
 docs: >
   Example of using the Draw interaction. Select a geometry type from the
   dropdown above to start drawing. To finish drawing, click the last
-  point. To activate freehand drawing for lines and polygons, hold the `Shift`
-  key. Square drawing is achieved by using Circle mode with a `geometryFunction`
-  that creates a 4-sided regular polygon instead of a circle. Box drawing uses a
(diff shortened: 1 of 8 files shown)


tschaub:
@ahocevar I was initially planning on supporting `'Box'` as a type, but that is `type: ol.geom.GeometryType`.

So then I started thinking about going ahead and adding `ol.geom.Box`.  But then I thought about the weird places where it probably wouldn't make sense, the confusion with `ol.Extent`, and all the test writing.

So then I thought we could change the type of the option to be `ol.interaction.Draw.Types` or something (note that "mode" and "type" are two separate things).  In the end, I think this makes the most sense - we could add more types as a convenience for determining the "mode" and the geometry function.

But if we do this second option, then we'll still want the static geometry function.

So the question is really whether we want to expose the "box" geometry function as part of the API, or make it so the `type` option for the draw interaction doesn't need to be `ol.geom.GeometryType`?

tschaub:
I decided to merge this as is and to take up the discussion about extending the `type` option vs. creating new geometry types in #5983.

ahocevar:
Cool, thanks for creating that ticket.

PR #5982, opened 2016-10-17
Use a https enabled GeoServer

ahocevar:
Until https://demo.boundlessgeo.com/geoserver is available, I stood up a GeoServer instance that serves everything we use from that server. We can either keep this one, or switch back to demo.boundlessgeo.com when SSL is enabled there.

### DIFF
--- a/examples/epsg-4326.js
+++ b/examples/epsg-4326.js
@@ -9,7 +9,7 @@ goog.require('ol.source.TileWMS');
 var layers = [
   new ol.layer.Tile({
     source: new ol.source.TileWMS({
-      url: 'http://demo.boundlessgeo.com/geoserver/wms',
+      url: 'https://ahocevar.com/geoserver/wms',
       params: {
         'LAYERS': 'ne:NE1_HR_LC_SR_W_DR'
       }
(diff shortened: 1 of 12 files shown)


tschaub:
Thanks @ahocevar.  What do you think about examples for previous releases?  I think our alternatives are to update them all to use your personal GeoServer (and potentially update them back later) or to just let the WFS examples for older releases fail when we move everything to https.

ahocevar:
I was thinking about another pull request that changes the urls on master for the website. If it's a single commit, it should be easy to revert later.

ahocevar:
Well, or change to demo.boundlessgeo.com instead of reverting of course.

ahocevar:
I think there would be another option: only redirect to https for http://openlayers.org/index.html and http://openlayers.org/en/latest/

tschaub:
It is straightforward to make the changes for previous releases (see e.g. https://github.com/openlayers/ol3/pull/5976#issuecomment-254060965) - and I've been pushing commits without review there for these URL changes.

Guess I was asking more in terms of the reliability (and additional load on your personal GeoServer).  But perhaps the additional traffic from old releases will be small.  So if it is going to change for the latest release, I imagine the same GeoServer could handle traffic from old releases as well.

I think it would be great to get this change in.  I'm interested to redirect all http to https as soon as we can.

ahocevar:
Cool, then I will make the same change on master as soon as this is merged and the website for master is updated.  My GeoServer should handle all these requests fine - I got a hardware upgrade recently.

PR #5985, opened 2016-10-17
Update Mapzen urls to new scheme

ahocevar:
Mapzen have changed their tile urls, and it seems the old urls were phased out earlier than planned. Essentially, this broke the osm-vector-tiles example, which is fixed by this pull request.


tschaub:
Looks good.  Please merge.

PR #5991, opened 2016-10-18
Only test ol.format.MVT if ArrayBuffer.isView is supported

### LINKED ISSUE #5990: Failing tests in IE 10
related to #5807: `Array.isView` is not supported on IE 10

### LINKED ISSUE #6003: Failing tests in IE 10 because pbf calls the undefined method ArrayBuffer.isView
On IE 10 (Windows 7) the following tests fail:

* `ol.featureloader`
  * `ol.featureloader.tile`
    * `sets features on the tile and updates proj units`

* `ol.format.MVT`
  * `#readFeatures`
    * `uses ol.render.Feature as feature class by default`
    * `parses only specified layers`
    * `parses geometries correctly`
    * `parses id property`

All with a `TypeError`: `Object doesn't support property or method 'isView'.

This comes from `ol3/build/ol.ext/pbf.js`, Line 106, which reads:

```js
this.buf = ArrayBuffer.isView(buf) ? buf : new Uint8Array(buf || 0);
```

Even though IE 10 has `ArrayBuffer` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)), it lacks the static method `ArrayBuffer.isView(arg)` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView))

`polyfill.io` doesn't have a polyfill for this method, but one can implement one for our test suite. PR to come.

fredj:
fixes #5990 
fixes #6003

### DIFF
--- a/test/spec/ol/format/mvt.test.js
+++ b/test/spec/ol/format/mvt.test.js
@@ -6,7 +6,7 @@ goog.require('ol.ext.vectortile');
 goog.require('ol.format.MVT');
 goog.require('ol.render.Feature');
 
-where('ArrayBuffer').describe('ol.format.MVT', function() {
+where('ArrayBuffer.isView').describe('ol.format.MVT', function() {
 
   var data;
   beforeEach(function(done) {
--- a/test/spec/ol/featureloader.test.js
+++ b/test/spec/ol/featureloader.test.js
@@ -2,9 +2,10 @@ goog.provide('ol.test.featureloader');
 
 goog.require('ol.events');
 goog.require('ol.VectorTile');
+goog.require('ol.Feature');
 goog.require('ol.featureloader');
 goog.require('ol.format.GeoJSON');
-goog.require('ol.format.MVT');
+goog.require('ol.format.TextFeature');
 goog.require('ol.proj');
 goog.require('ol.source.Vector');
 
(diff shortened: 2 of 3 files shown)


test/spec/ol/format/mvt.test.js
```diff
@@ -82,4 +85,21 @@ where('ArrayBuffer').describe('ol.format.MVT', function() {
 
   });
 
+  describe('ol.featureloader.tile', function() {
```
ahocevar: This is actually a test for the featureloader working with tile-pixels as units. It could be done with a mock format as well. So it makes sense to keep it in featureloader.test.js. But instead of an ol.format.MVT, you could use anything else that uses tile-pixels as units.

fredj: ok, I will change it

marcjansen:
@ahocevar, is this now good to go from your side? Alternatively (with more overhead) we could have our own polyfill as suggested in https://github.com/openlayers/ol3/issues/6003#issuecomment-255064819 You decide.

marcjansen:
@fredj Did you test whether this passes in IE 10?

fredj:
@marcjansen re-checking now

fredj:
Thanks for the review

Issue #5862, opened 2016-09-07
Selecting on mobile devices is almost impossible


giohappy:
Selecting points and lines with touch is a metter of luck. It becomes more reliable from 10px up, but 10px wide lines (and points) are generally too big.
It's related to #2105.
Anybody willing to work on this? We could fund it (at least partially).

giohappy:
Probably the solution implemented in ol.interaction.Modify could be ported to selection too.

cvenzin:
You can implement something by your self on top of the ol3 api. I've chosen an algorithm from here [Looping in a spiral](http://stackoverflow.com/questions/398299/looping-in-a-spiral), with the idea to call forEachFeatureAtPixel on all neighboring pixels in a limited range. After the first feature is hit the algorithm stops and I progress in my workflow. 
I've also seen other solutions using getClosestFeatureToCoordinate on the vector layers. Not sure which way is more performant. I'm fine with my solution.
```
map.on('click', (evt) => {
                this.calcHitMatrix(evt.pixel);
        });
private calcHitMatrix(pixel: number[]) {
        let X = 11; // -5 to + 5 in x
        let Y = 11; // -5 to + 5 in y
        let x = 0;
        let y = 0;
        let dx = 0;
        let dy = -1;
        let t = Math.max(X, Y);
        let maxI = t * t;
        let pixelX = pixel[0];
        let pixelY = pixel[1];
        for (let i = 0; i < maxI; i++) {
            if ((-X / 2 <= x) && (x <= X / 2) && (-Y / 2 <= y) && (y <= Y / 2)) {
                if (this.displayFeatureInfo([pixelX + x, pixelY + y])) {
                    return;
                }
            }
            if ((x === y) || ((x < 0) && (x === -y)) || ((x > 0) && (x === 1 - y))) {
                t = dx;
                dx = -dy;
                dy = t;
            }
            x += dx;
            y += dy;
        }
    }

private displayFeatureInfo(pixel) {
        return this.mapService.map.forEachFeatureAtPixel(pixel, (feature, layer) => {
            return this.cbForEachFeatureAtPixel(feature, layer, pixel);
        });
    }

private cbForEachFeatureAtPixel(feature: any, layer: any, pixel: number[]) {
        if (layer && layer.get('hxgnReport')) {
            this.mapService.overlayLayer.getSource().addFeature(feature);
            let report = layer.get('hxgnReport');
[...]

PR #5995, opened 2016-10-19
Add hit tolerance parameter to ol.Map#forEachFeatureAtPixel

simonseyock:
TODO:
- [X] Write a test
- [X] Change the signature of map.forEachFeatureAtCoordinate and map.forEachFeatureAtPixel in the source code, the examples and the tests
- [X] Add a hitTolerance parameter to all internal forEachFeatureAtCoordinate methods of the renderer
- [X] Make ol.render.canvas.ReplayGroup.prototype.forEachFeatureAtCoordinate work with a hitContext of variable size
- [x] Implement the midpoint circle algorithm and checking only pixels inside the circle
- [x] Update documentation
- [X] hitTolerance should also be added to ol.Map#hasFeatureAtPixel.
- [X] example
- [X] caching
- [x] Test with retina screens (devicePixelRatio > 1)
- [X] Handle VectorTiles.
~Write a test for VectorTiles.~
- [x] Update changelog.
- [x] Created test for ol.render.canvas.ReplayGroup#getCircleArray_

See discussion: #5862

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -15,16 +15,16 @@ Old syntax:
 
 New syntax:
 ```
-    map.forEachFeatureAtPixel(pixel, callback, callbackThis, {
-        layerFilter: layerFilterFn,
-        layerFilterThis: layerFilterThis
+    map.forEachFeatureAtPixel(pixel, callback, {
+        layerFilter: layerFilterFn
     });
     
     map.hasFeatureAtPixel(pixel, {
-        layerFilter: layerFilterFn,
-        layerFilterThis: layerFilterThis
+        layerFilter: layerFilterFn
     });
 ```
+
+To bind a function to a this, please use the bind method of the function (See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind).
 
 This change is due to the introduction of the `hitTolerance` parameter which can be passed in via this `ol.AtPixelOptions` object, too.
 
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,32 @@
 
 ### Next release
 
+#### `ol.Map#forEachFeatureAtPixel` and `ol.Map#hasFeatureAtPixel` parameters have changed
+ 
+If you are using the layer filter of one of these methods, please note that you now have to pass in the layer filter via an `ol.AtPixelOptions` object. If you are not using the layer filter the usage has not changed.
+
+Old syntax:
+```
+    map.forEachFeatureAtPixel(pixel, callback, callback_this, layerFilter_function, layerFilter_this);
+    
+    map.hasFeatureAtPixel(pixel, layerFilter_function, layerFilter_this);
+```
+
+New syntax:
+```
+    map.forEachFeatureAtPixel(pixel, callback, callback_this, {
+        layerFilter: layerFilter_function,
+        layerFilterThis: layerFilter_this
+    });
+    
+    map.hasFeatureAtPixel(pixel, {
+        layerFilter: layerFilter_function,
+        layerFilterThis: layerFilter_this
+    });
+```
+
+This change is due to the introduction of the `hitTolerance` parameter which can be passed in via this `ol.AtPixelOptions` object, too.
+
 #### Use `view.animate()` instead of `map.beforeRender()` and `ol.animation` functions
 
 The `map.beforeRender()` and `ol.animation` functions have been deprecated in favor of a new `view.animate()` function.  Use of the deprecated functions will result in a warning during development.  These functions are subject to removal in an upcoming release.
@@ -73,20 +99,6 @@ As last step in the removal of the dependency on Google Closure Library, the `go
 
 `ol.format.ogc.filter` was simplified to `ol.format.filter`; to upgrade your code, simply remove the `ogc` string from the name.
 For example: `ol.format.ogc.filter.and` to `ol.format.filter.and`.
-
-#### `ol.Map#forEachFeatureAtPixel` and `ol.Map#hasFeatureAtPixel` parameters have changed
- 
-If you are using the layer filter of one of these methods, please note that you now have to pass in the layer filter via an `ol.AtPixelOptions` object. If you are not using the layer filter the usage has not changed.
- 
-New syntax:
-```
-    map.forEachFeatureAtPixel(pixel, callback, callback_this, {
-        layerFilter: layerFilterFunction,
-        layerFilterThis: someObject
-    })
-```
-
-This change is due to the introduction of the `hitTolerance` parameter which can be passed in via this `ol.AtPixelOptions` object, too.
 
 #### Changes only relevant to those who compile their applications together with the Closure Compiler
 
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -74,6 +74,20 @@ As last step in the removal of the dependency on Google Closure Library, the `go
 `ol.format.ogc.filter` was simplified to `ol.format.filter`; to upgrade your code, simply remove the `ogc` string from the name.
 For example: `ol.format.ogc.filter.and` to `ol.format.filter.and`.
 
+#### `ol.Map#forEachFeatureAtPixel` and `ol.Map#hasFeatureAtPixel` parameters have changed
+ 
+If you are using the layer filter of one of these methods, please note that you now have to pass in the layer filter via an `ol.AtPixelOptions` object. If you are not using the layer filter the usage has not changed.
+ 
+New syntax:
+```
+    map.forEachFeatureAtPixel(pixel, callback, callback_this, {
+        layerFilter: layerFilterFunction,
+        layerFilterThis: someObject
+    })
+```
+
+This change is due to the introduction of the `hitTolerance` parameter which can be passed in via this `ol.AtPixelOptions` object, too.
+
 #### Changes only relevant to those who compile their applications together with the Closure Compiler
 
 A number of internal types have been renamed.  This will not affect those who use the API provided by the library, but if you are compiling your application together with OpenLayers and using type names, you'll need to do the following:
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -8,21 +8,21 @@ If you are using the layer filter of one of these methods, please note that you 
 
 Old syntax:
 ```
-    map.forEachFeatureAtPixel(pixel, callback, callback_this, layerFilter_function, layerFilter_this);
+    map.forEachFeatureAtPixel(pixel, callback, callbackThis, layerFilterFn, layerFilterThis);
     
-    map.hasFeatureAtPixel(pixel, layerFilter_function, layerFilter_this);
+    map.hasFeatureAtPixel(pixel, layerFilterFn, layerFilterThis);
 ```
 
 New syntax:
 ```
-    map.forEachFeatureAtPixel(pixel, callback, callback_this, {
-        layerFilter: layerFilter_function,
-        layerFilterThis: layerFilter_this
+    map.forEachFeatureAtPixel(pixel, callback, callbackThis, {
+        layerFilter: layerFilterFn,
+        layerFilterThis: layerFilterThis
     });
     
     map.hasFeatureAtPixel(pixel, {
-        layerFilter: layerFilter_function,
-        layerFilterThis: layerFilter_this
+        layerFilter: layerFilterFn,
+        layerFilterThis: layerFilterThis
     });
 ```
 
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2910,7 +2910,8 @@ olx.interaction.ExtentOptions.prototype.wrapX;
 /**
  * @typedef {{
  *     features: (ol.Collection.<ol.Feature>|undefined),
- *     layers: (undefined|Array.<ol.layer.Layer>|function(ol.layer.Layer): boolean)
+ *     layers: (undefined|Array.<ol.layer.Layer>|function(ol.layer.Layer): boolean),
+ *     hitTolerance: (number|undefined)
  * }}
  */
 olx.interaction.TranslateOptions;
@@ -2935,6 +2936,16 @@ olx.interaction.TranslateOptions.prototype.features;
  * @api
  */
 olx.interaction.TranslateOptions.prototype.layers;
+
+
+/**
+ * Hit-detection tolerance. Pixels inside the radius around the given position
+ * will be checked for features. This only works for the canvas renderer and
+ * not for WebGL.
+ * @type {number|undefined}
+ * @api
+ */
+olx.interaction.TranslateOptions.prototype.hitTolerance;
 
 
 /**
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -301,6 +301,49 @@ olx.MapOptions.prototype.target;
  */
 olx.MapOptions.prototype.view;
 
+/**
+ * The filter function will receive one argument, the
+ * {@link ol.layer.Layer layer-candidate} and it should return a boolean
+ * value.
+ * @typedef {(function(ol.layer.Layer): boolean)}
+ */
+olx.LayerFilterFunction;
+
+/**
+ * Object literal with options for the forEachFeatureAtCoordinate methods.
+ * @typedef {{layerFilter: (olx.LayerFilterFunction|undefined),
+ *    layerFilterThis: (Object|undefined),
+ *    hitTolerance: (number|undefined)}}
+ */
+olx.ForEachFeatureOptions;
+
+
+/**
+ * Layer filter function. Only layers which are visible and for which this function returns
+ * `true` will be tested for features. By default, all visible layers will
+ * be tested.
+ * @type {olx.LayerFilterFunction|undefined}
+ * @api stable
+ */
+olx.ForEachFeatureOptions.prototype.layerFilter;
+
+
+/**
+ * Value to use as `this` when executing `layerFilter`.
+ * @type {Object}
+ * @api stable
+ */
+olx.ForEachFeatureOptions.prototype.layerFilterThis;
+
+
+/**
+ * Value of a radius in whichs area around the given coordinate features are
+ *    called.
+ * @type {number}
+ * @api stable
+ */
+olx.ForEachFeatureOptions.prototype.hitTolerance;
+
 
 /**
  * Object literal with config options for the overlay.
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -309,7 +309,7 @@ olx.MapOptions.prototype.view;
  *    layerFilterThis: (Object|undefined),
  *    hitTolerance: (number|undefined)}}
  */
-olx.FeatureAtPixelOptions;
+olx.AtPixelOptions;
 
 
 /**
@@ -320,7 +320,7 @@ olx.FeatureAtPixelOptions;
  * @type {((function(ol.layer.Layer): boolean)|undefined)}
  * @api stable
  */
-olx.FeatureAtPixelOptions.prototype.layerFilter;
+olx.AtPixelOptions.prototype.layerFilter;
 
 
 /**
@@ -328,7 +328,7 @@ olx.FeatureAtPixelOptions.prototype.layerFilter;
  * @type {Object|undefined}
  * @api stable
  */
-olx.FeatureAtPixelOptions.prototype.layerFilterThis;
+olx.AtPixelOptions.prototype.layerFilterThis;
 
 
 /**
@@ -338,7 +338,7 @@ olx.FeatureAtPixelOptions.prototype.layerFilterThis;
  * @type {number|undefined}
  * @api
  */
-olx.FeatureAtPixelOptions.prototype.hitTolerance;
+olx.AtPixelOptions.prototype.hitTolerance;
 
 
 /**
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3212,7 +3212,8 @@ olx.interaction.PointerOptions.prototype.handleUpEvent;
  *     multi: (boolean|undefined),
  *     features: (ol.Collection.<ol.Feature>|undefined),
  *     filter: (ol.SelectFilterFunction|undefined),
- *     wrapX: (boolean|undefined)}}
+ *     wrapX: (boolean|undefined),
+ *     hitTolerance: (number|undefined)}}
  */
 olx.interaction.SelectOptions;
 
@@ -3324,6 +3325,16 @@ olx.interaction.SelectOptions.prototype.filter;
  * @api
  */
 olx.interaction.SelectOptions.prototype.wrapX;
+
+
+/**
+ * Hit-detection tolerance. Pixels inside the radius around the given position
+ * will be checked for features. This only works for the canvas renderer and
+ * not for WebGL.
+ * @type {number|undefined}
+ * @api
+ */
+olx.interaction.SelectOptions.prototype.hitTolerance;
 
 
 /**
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -303,39 +303,42 @@ olx.MapOptions.prototype.view;
 
 
 /**
- * Object literal with options for the forEachFeatureAtCoordinate methods.
+ * Object literal with options for the forEachFeatureAtPixel and
+ * hasFeatureAtPixel methods.
  * @typedef {{layerFilter: ((function(ol.layer.Layer): boolean)|undefined),
  *    layerFilterThis: (Object|undefined),
  *    hitTolerance: (number|undefined)}}
  */
-olx.ForEachFeatureOptions;
+olx.FeatureAtPixelOptions;
 
 
 /**
- * Layer filter function. Only layers which are visible and for which this function returns
- * `true` will be tested for features. By default, all visible layers will
- * be tested.
+ * Layer filter function. The filter function will receive one argument, the
+ * {@link ol.layer.Layer layer-candidate} and it should return a boolean value.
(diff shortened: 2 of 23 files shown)


src/ol/renderer/canvas/vectortilelayer.js
```diff
@@ -276,7 +276,7 @@ ol.renderer.canvas.VectorTileLayer.prototype.createReplayGroup = function(tile,
 /**
  * @inheritDoc
  */
-ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = function(coordinate, frameState, callback, thisArg) {
+ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = function(coordinate, frameState, hitTolerance, callback, thisArg) {
   var resolution = frameState.viewState.resolution;
   var rotation = frameState.viewState.rotation;
   var layer = this.getLayer();
```
simonseyock: While updating this method i recognized that here is some code which probably needs to be updated, too. Is it sufficient to buffer up the extent by hitTolerance*resolution?

src/ol/map.js
```diff
@@ -585,27 +570,25 @@ ol.Map.prototype.disposeInternal = function() {
  *     the {@link ol.layer.Layer layer} of the feature and will be null for
  *     unmanaged layers. To stop detection, callback functions can return a
  *     truthy value.
- * @param {S=} opt_this
+ * @param {S=} opt_this Value to use as this when executing callback.
+ * @param {olx.ForEachFeatureOptions=} opt_options Optional options.
```
simonseyock: The compilers demands all optional parameters to appear at the end of the signature. To be concise with the other uses of opt_this i left it right behind the callback and inserted the opt_options behind that.

externs/olx.js
```diff
@@ -301,6 +301,49 @@ olx.MapOptions.prototype.target;
  */
 olx.MapOptions.prototype.view;
 
+/**
+ * The filter function will receive one argument, the
+ * {@link ol.layer.Layer layer-candidate} and it should return a boolean
+ * value.
+ * @typedef {(function(ol.layer.Layer): boolean)}
+ */
+olx.LayerFilterFunction;
```
simonseyock: I introduced this typedef because i was not able to annotate it properly otherwise. I will see if i can remove this.

src/ol/map.js
```diff
@@ -570,30 +570,25 @@ ol.Map.prototype.disposeInternal = function() {
...
  * @return {T|undefined} Callback result, i.e. the return value of last
  * callback execution, or the first truthy callback return value.
- * @template S,T,U
+ * @template S,T
  * @api stable
  */
-ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this, opt_layerFilter, opt_this2) {
+ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this, opt_options) {
```
simonseyock: The compilers demands all optional parameters to appear at the end of the signature. To be concise with the other uses of opt_this i left it right behind the callback and inserted the opt_options behind that.

simonseyock:
This development has been funded by KLAUS BENNDORF and GIS3W. The review has been funded by CAMPTOCAMP.

ischas:
The message above had to be added for @giohappy's company as proof of expenses for their customers.

simonseyock:
signature of ol.Map.forEachFeatureAtPixel is now:
`forEachFeatureAtPixel(pixel, callback, opt_this, opt_options)` (reasons see above)
For the internal methods i added the hitTolerance after the frameState.

tsauerwein:
I will reserve some time next week to do the review!

But already one comment: Changing the signature of `ol.Map#forEachFeatureAtPixel` breaks backward-compatibility. This would mean that the change would have to be described in the [upgrade notes](https://github.com/openlayers/ol3/blob/master/changelog/upgrade-notes.md) and everyone using `forEachFeatureAtPixel` with a layer filter would have to adapt the method call.

A method with 6 parameters is not *optimal*, but breaking backward-compatibility is something that we usually try to avoid (especially for a widely used functionality). What do others think?

ischas:
The optional options-object was proposed by @tschaub here: https://github.com/openlayers/ol3/issues/5862#issuecomment-248637081

src/ol/renderer/canvas/vectortilelayer.js
simonseyock: After looking further into this, i realized there is more to do.

simonseyock: As i don't know how to setup a test for this, i can't really work on this problem.
How could i set up a vectortilelayer containing only one feature which is very near to a corner of tile?
i tried something like this, but it does not seem to work: 

      // 1 x 1 pixel black icon
      var img = document.createElement('img');
      img.onload = function() {
        var target = document.createElement('div');
        target.style.width = '256px';
        target.style.height = '256px';
        document.body.appendChild(target);

        var tileGrid = ol.tilegrid.createXYZ();
        var zoom = 10;
        //choose coordinate at an edge of the tileGrid
        var coordinate = ol.extent.getTopLeft(tileGrid.getTileCoordExtent([zoom, 0, 0]));

        var map = new ol.Map({
          view: new ol.View({
            center: coordinate,
            zoom: zoom
          }),
          target: target
        });

        var feature = new ol.Feature(new ol.geom.Point(coordinate));
        feature.setStyle([new ol.style.Style({
          image: new ol.style.Icon({
            img : img,
            imgSize: [1, 1]
          })
        })]);

        var TileClass = function() {
          ol.VectorTile.apply(this, arguments);
          if (this.tileCoord[0] === zoom && this.tileCoord[1] === 0 && this.tileCoord[2] === 0) {
            this.setFeatures([feature]);
          }
          this.setState('loaded');
          this.setProjection(ol.proj.get('EPSG:3857'));
        };
        ol.inherits(TileClass, ol.VectorTile);

        var source = new ol.source.VectorTile({
          format: new ol.format.MVT(),
          tileClass: TileClass,
          tileGrid: tileGrid
        });
        var layer = new ol.layer.VectorTile({
          source: source
        });
        map.addLayer(layer);
        map.renderSync();

        var cb0 = sinon.spy();

        var pixel = map.getPixelFromCoordinate(coordinate);

        map.forEachFeatureAtPixel(pixel, cb0);
        expect(cb0.callCount).to.be(1);  // this already fails
        expect(cb0.firstCall.args[1]).to.be(layer);

Another question about the vectortilelayer for me is: how are features stored which are in multiple tiles? Could a simple buffering up the extent result in multiple calls of the callback for the same feature?

tsauerwein: > Another question about the vectortilelayer for me is: how are features stored which are in multiple tiles? Could a simple buffering up the extent result in multiple calls of the callback for the same feature?

Yes, that could happen. We probably want to check that the callback is only called once for features with the same id?

tsauerwein: Maybe @ahocevar can help you with your other questions.

src/ol/render/canvas/replaygroup.js
```diff
@@ -65,22 +65,64 @@ ol.render.canvas.ReplayGroup = function(
...
+ * @param {number} radius Radius.
+ * @returns {Array.<Array.<Boolean>>} an array with marked circel points.
+ * @private
+ */
+ol.render.canvas.ReplayGroup.createPixelCircle_ = function(radius) {
+  var arraySize = radius * 2 + 1;
+  var arr = new Array(arraySize);
+  for (var i = 0; i < arraySize; i++) {
+    arr[i] = new Array(arraySize);
+    for (var j = 0; j < arraySize; j++) {
```
simonseyock: good idea, that should work.

src/ol/render/canvas/replaygroup.js
```diff
@@ -65,22 +65,64 @@ ol.render.canvas.ReplayGroup = function(
...
-  this.hitDetectionContext_ = ol.dom.createCanvasContext2D(1, 1);
+/**
+ * This methods creates a circle inside a fitting array. Points inside the
+ * circle are marked by true, points on the outside are marked with false.
+ * It uses the midpoint circle algorithm.
+ * @param {number} radius Radius.
+ * @returns {Array.<Array.<Boolean>>} an array with marked circel points.
+ * @private
+ */
+ol.render.canvas.ReplayGroup.createPixelCircle_ = function(radius) {
```
simonseyock: I was thinking about it, but it is an almost 1:1 implementation of the standard algorithm. A test would be possible, though.

externs/olx.js
```diff
@@ -303,6 +303,42 @@ olx.MapOptions.prototype.view;
...
+ * @api stable
+ */
+olx.ForEachFeatureOptions.prototype.layerFilterThis;
+
+
+/**
+ * Value of a radius in whichs area around the given coordinate features are
+ *    called.
+ * @type {number}
+ * @api stable
```
simonseyock: sounds reasonable ;)

simonseyock:
The replaygroup now caches the arrays. Both methods and the cache are static so the arrays can be used by all replaygroups.
I added a hitTolerance parameter to the select interaction and integrated the hitTolerance into the select example.
Still TODO:
[ ] - Test with retina screens (devicePixelRatio > 1). <- I can only test this via the google chrome emulator at the moment.
[ ] - Handle VectorTiles.
[ ] - Update changelog.

simonseyock:
I forgot to mention i added hitTolerance to hasFeatureAtPixel. What about forEachLayerAtPixel?

tsauerwein:
> I forgot to mention i added hitTolerance to hasFeatureAtPixel. What about forEachLayerAtPixel?

`forEachFeatureAtPixel` is usually used in combination with `hasFeatureAtPixel`. That's why it was important for me that `hasFeatureAtPixel` has support for the tolerance.

`forEachLayerAtPixel` is something different, but if you also want to work on it as well, you are welcome! Maybe in a separate PR to not bloat this one.

simonseyock:
Ok, i will leave it out for now.

changelog/upgrade-notes.md
```diff
@@ -9,6 +9,20 @@ As last step in the removal of the dependency on Google Closure Library, the `go
...
+New Syntax:
+```
+    map.forEachFeatureAtPixel(pixel, callback, callback_this, {
+        layerFilter: layerFilterFunction,
+        layerFilterThis: someObject
+    })
+```
+
+This change is due to the introduction of the new `hitTolerance` parameter which can be passed in via the `ol.AtPixelOptions` object. 
+
```
simonseyock: Is this the right place? (i will move it up after the rebase)

changelog/upgrade-notes.md
```diff
@@ -9,6 +9,20 @@ As last step in the removal of the dependency on Google Closure Library, the `go
...
+New syntax:
+```
+    map.forEachFeatureAtPixel(pixel, callback, callback_this, {
+        layerFilter: layerFilterFunction,
+        layerFilterThis: someObject
+    })
+```
+
+This change is due to the introduction of the `hitTolerance` parameter which can be passed in via this `ol.AtPixelOptions` object, too.
+
```
simonseyock: Is this the right place? (i will move it up after the rebase)

simonseyock:
I recognized that i could test the pixelRatio with a unit-test.
What is the expected behaviour? Should a click on a device with a higher pixel ratio hit the same features as on a device with a lower pixel ratio or should it hit more features?

simonseyock:
Or to say it with other words: should the hitTolerance depend on the physical distance on the screen or should that problem be left up to the developer to decide? (He could just use `hitTolerance * devicePixelRatio` if he would want that). I think that normally you would want that it stays the same physical distance.

src/ol/render/canvas/replaygroup.js
simonseyock: I wrote a test for this (`ol.test.render.canvas.ReplayGroup#getCircle_`)

simonseyock:
I buffer the tile extent by `hitTolerance * pointResolution` in `ol.renderer.canvas.VectorTileLayer#forEachFeatureAtCoordinate`. I don't know exactly if this works correctly.
As there is already a check in there that ensures that the callback is called only once for each feature i don't need to do that.

simonseyock:
@tsauerwein: can you clarify how the pixelRatio should be handled? can you review this whole thing again (maybe after i implemented everything regarding the pixelRatio)?

giohappy:
@tsauerwein could you review it again? we're funding this improvement and it is critical to us having it merged. it seems that pixelRatio handling (for retina displays) is the only remaining part to implement, isn'it?

tsauerwein:
> @tsauerwein could you review it again? we're funding this improvement and it is critical to us having it merged.

Well, you are not funding my company. This week is quite busy for me, I will plan some time for next week.

giohappy:
@tsauerwein I didn't meant to be controversial. We just needed a feedback ;)
Thanks for planning some time for it.

giohappy:
great work @simonseyock !
I've just tested your branch with our webgis editing tools taking advantage of the new hit tolerance, and it works great from (Huawei) smartphone and (Samsung) tablet. I will do more tests with different devices. While pixelRatio won't be managed automatically, we will tweek the hit tolerance value by ourselves.

externs/olx.js
```diff
@@ -303,6 +303,45 @@ olx.MapOptions.prototype.view;
 
 
 /**
+ * Object literal with options for the forEachFeatureAtPixel and
+ * hasFeatureAtPixel methods.
```
tsauerwein: You could use: `{@link ol.Map#forEachFeatureAtPixel}` and `{@link ol.Map#hasFeatureAtPixel}`

src/ol/map.js
```diff
@@ -573,30 +573,28 @@ ol.Map.prototype.disposeInternal = function() {
...
+ * @param {S=} opt_this Value to use as this when executing callback.
+ * @param {olx.AtPixelOptions=} opt_options Optional options.
  * @return {T|undefined} Callback result, i.e. the return value of last
  * callback execution, or the first truthy callback return value.
- * @template S,T,U
+ * @template S,T
  * @api stable
  */
-ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this, opt_layerFilter, opt_this2) {
+ol.Map.prototype.forEachFeatureAtPixel = function(pixel, callback, opt_this,opt_options) {
```
tsauerwein: Doesn't the linter complain about the missing space before `opt_options`?

src/ol/render/canvas/replaygroup.js
```diff
@@ -71,18 +71,88 @@ ol.render.canvas.ReplayGroup = function(
...
-   * @private
-   * @type {ol.Transform}
-   */
-  this.hitDetectionTransform_ = ol.transform.create();
 
+/**
+ * @type {Object.<number, Array.<Array.<(boolean|undefined)>>>}
+ * @private
+ */
+ol.render.canvas.ReplayGroup.circleArrayCache_ = {
```
tsauerwein: Maybe you could leave a short comment what this cache is used for.

src/ol/render/canvas/replaygroup.js
```diff
@@ -130,20 +204,30 @@ ol.render.canvas.ReplayGroup.prototype.forEachFeatureAtCoordinate = function(
     ol.extent.buffer(hitExtent, resolution * this.renderBuffer_, hitExtent);
```
tsauerwein: I am wondering if you would also have to take the `hitTolerance` into account when calculating the `hitExtent`.

For example: If you have a point at `0,0` (in pixel coordinates) which has a symbol of 300x300 pixels, you would set the `renderBuffer` to 150px, so that the point can be selected properly. Even when clicking at `150,0` the point could be selected.
But now if you set the `hitTolerance` to 10px, you would expect to get the point if you click at `155,0`. But the `hitExtent` would be `5,-150,305,150`, so the point would not be considered for the hit detection.

Does that make sense? What do you think?

tsauerwein:
@simonseyock I agree with you that it would make sense to adapt the hit-tolerance depending on the pixel-ratio. And this also seems to be what @giohappy is proposing.

src/ol/map.js
simonseyock: Apparently not ... i added the space.

src/ol/render/canvas/replaygroup.js
simonseyock: Yes that seems to be true, as all draw instructions not intersecting the hitContext are skipped in the replay. Normally this would not happen as the the renderBuffer defaults to 100px and therefore is much greater than the hitTolerance.

I add this.

Seeing here that the hitBuffer is only multiplied with the normal resolution i will apply the same method to the extent of the vectortiles.

simonseyock:
@tsauerwein: thanks for looking into this again

simonseyock:
The hitTolerance is multiplied by the pixelRatio now.

As @giohappy mentioned they want to use the hitTolerance with the TranslateInteraction, i added a hitTolerance parameter to it.

For my part I have done everything I can think of for now.

tsauerwein:
Thanks for your continuous effort and good work on this!

Before we merge, I would like to get a *thumbs-up* from the rest of the core-team about the proposed API change. As documented in the [changelogs](https://github.com/KlausBenndorf/ol3/blob/86eee7fa7543957a829e476c83bddeeeb172bf90/changelog/upgrade-notes.md#olmapforeachfeatureatpixel-and-olmaphasfeatureatpixel-parameters-have-changed), all users that previously used a layer filter in [ol.Map#forEachFeatureAtPixel](https://github.com/KlausBenndorf/ol3/blob/86eee7fa7543957a829e476c83bddeeeb172bf90/src/ol/map.js#L563-L583) or [ol.Map#hasFeatureAtPixel](https://github.com/KlausBenndorf/ol3/blob/86eee7fa7543957a829e476c83bddeeeb172bf90/src/ol/map.js#L640-L649) will have to adapt their code.

The proposed syntax for these methods is:
```
    map.forEachFeatureAtPixel(pixel, callback, callbackThis, {
        layerFilter: layerFilterFn,
        layerFilterThis: layerFilterThis,
        hitTolerance: 10
    });

    map.hasFeatureAtPixel(pixel, {
        layerFilter: layerFilterFn,
        layerFilterThis: layerFilterThis,
        hitTolerance: 10
    });
```

@tschaub, you always have good ideas about API design. Does this look ok for you?

ahocevar:
My suggestion would be a slightly simpler API:
```js
    map.forEachFeatureAtPixel(pixel, callback, {
        layerFilter: layerFilterFn,
        hitTolerance: 10
    });

    map.hasFeatureAtPixel(pixel, {
        layerFilter: layerFilterFn,
        hitTolerance: 10
    });
```
In other words: let's get rid of `callbackThis` and `layerFilterThis`. With `Function.bind()` being standard JavaScript now, it does not make sense to carry around `this` configurations.

simonseyock:
@ahocevar: i agree that the this are not needed nowadays.

But this would be inconsistent with the other code in the library (just search for opt_this). So we might need to do a follow up PR that gets rid of opt_this in the whole source ... but this is quite much work.

ahocevar:
@simonseyock In this case we introduce an object literal with options, and we do not have `this` options in any other options object literal.

ischas:
Any chance to get this into v3.20.0?

ahocevar:
> Any chance to get this into v3.20.0?

@ischas If the remaining review comments get addressed today, then there is a chance.

ischas:
No free ressources here to make it today. Looking forward to get it into v3.21.0 (and using our fork until then).

ahocevar:
I'm adding this ticket to the v3.20 milestone regardless. When the comments are addressed by the time the release is cut, it will be in, otherwise it will be moved to the v3.21 milestone.

simonseyock:
I simplified the api and left a short note in the upgrade-notes about function.bind with a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind

externs/olx.js
```diff
@@ -324,14 +324,6 @@ olx.AtPixelOptions.prototype.layerFilter;
 
 
 /**
- * Value to use as `this` when executing `layerFilter`.
- * @type {Object|undefined}
- * @api stable
- */
-olx.AtPixelOptions.prototype.layerFilterThis;
```
fredj: please also remove `layerFilterThis` in the typedef (line 309)

simonseyock: done

ischas:
Once again, conflicts generated by another merge. @simonseyock: Please resolve the conflicts and remove the "(WIP)" since the only missing point is the VectorTiles-test which we are not able to provide (so maybe you should strike this out in the todo list). This can be a task for a follow-up pull request of someone who is familar with VectorTiles. At last, please ping the involved core team members after resolving the conflicts to get this in before this becomes a never ending story (and before you have to change your user name to Sisyphus ;-) ).

src/ol/renderer/canvas/vectortilelayer.js
```diff
@@ -212,7 +213,7 @@ ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = functi
     }
     replayGroup = tile.getReplayState().replayGroup;
     found = found || replayGroup.forEachFeatureAtCoordinate(
-        tileSpaceCoordinate, resolution, rotation, {},
+        tileSpaceCoordinate, resolution, rotation, hitTolerance, {},
```
ahocevar: I do not think this will yield the expected result. Should probably be `hitTolerance * resolution`.

src/ol/render/canvas/replaygroup.js
```diff
@@ -101,6 +174,7 @@ ol.render.canvas.ReplayGroup.prototype.finish = function() {
  * @param {ol.Coordinate} coordinate Coordinate.
  * @param {number} resolution Resolution.
  * @param {number} rotation Rotation.
+ * @param {number} hitTolerance Hit tolerance.
```
ahocevar: What are the units of `hitTolerance`? Pixels or map units?

marcjansen:
TBH I left this PR unmerged because @tsauerwein had reviewed it. He should also merge it IMO.

Once the conflicts are resolved, I will merge it nonetheless, given how much traction it has.

deal?

marcjansen:
@ahocevar has the ball right now.

src/ol/render/canvas/replaygroup.js
```diff
@@ -109,16 +183,19 @@ ol.render.canvas.ReplayGroup.prototype.finish = function() {
...
  */
 ol.render.canvas.ReplayGroup.prototype.forEachFeatureAtCoordinate = function(
-    coordinate, resolution, rotation, skippedFeaturesHash, callback) {
+    coordinate, resolution, rotation, hitTolerance, skippedFeaturesHash, callback) {
 
-  var transform = ol.transform.compose(this.hitDetectionTransform_,
-      0.5, 0.5,
+  hitTolerance = Math.round(hitTolerance);
+  var contextSize = hitTolerance * 2 + 1;
+  var transform = ol.transform.compose(ol.transform.create(),
```
ahocevar: To produce less garbage, it would be better here to reuse an existing transform, like `this.hitDetectionTransform_` in the current code.

src/ol/render/canvas/replaygroup.js
```diff
@@ -109,16 +183,19 @@ ol.render.canvas.ReplayGroup.prototype.finish = function() {
...
+      hitTolerance + 0.5, hitTolerance + 0.5,
       1 / resolution, -1 / resolution,
       -rotation,
       -coordinate[0], -coordinate[1]);
-
   var context = this.hitDetectionContext_;
-  context.clearRect(0, 0, 1, 1);
+  context.canvas.width = contextSize;
+  context.canvas.height = contextSize;
+  context.clearRect(0, 0, contextSize, contextSize);
```
ahocevar: For performance reasons, ou should either set  `width` and `height` on the canvas, or call `clearRect()`. Not both. My recommendation for best performance would be to only set `width` and `height` when `contextSize` is different than the current `width` and `height`, and call `clearRect()` otherwise. This is also what we do elsewhere.

externs/olx.js
```diff
@@ -303,6 +303,36 @@ olx.MapOptions.prototype.view;
...
+
+
+/**
+ * Hit-detection tolerance. Pixels inside the radius around the given position
+ * will be checked for features. This only works for the canvas renderer and
+ * not for WebGL.
+ * @type {number|undefined}
+ * @api
+ */
+olx.AtPixelOptions.prototype.hitTolerance;
```
ahocevar: You should document the units of `hitTolerance`. Pixels? Map units?

src/ol/renderer/canvas/vectortilelayer.js
simonseyock: hitTolerance is in pixels.

src/ol/render/canvas/replaygroup.js
simonseyock: Well, yes. Thanks for noticing this. I don't know why I removed it.

externs/olx.js
simonseyock: Done.

simonseyock:
Ok. I did the requested changes. Thanks for finding these @ahocevar

tsauerwein:
Thanks for your additional comments, @ahocevar! I will let you do the merge, when you are satisfied with the changes.

ahocevar:
Thanks to everyone involved in this pull request. I'm going to merge this now.

Personally I'd be more in favour of a separate example that shows the effect of the new `hitTolerance` parameter, instead of squeezing it in the already bloated `select-features.html` example. But that can be addressed in a follow-up pull request.

ischas:
Thank you all for the continuous effort to get this in.

giohappy:
Great work. Thanks everyone!

ahocevar:
It looks like some of the tests added with this pull request never passed in browsers. I created #6465 to track that issue.

Issue #5996, opened 2016-10-19
Failing tests in MS Edge


marcjansen:
We currently have 45 failing tests in Microsoft Edge 38.14393.0.0.

44 of them can be fixed by always using a [URL](https://developer.mozilla.org/en/docs/Web/API/URL)-polyfill, not only when the user-agent signals it to `polyfill.io`. PR to come.

The remaining error will be examined and tackled separately.

marcjansen:
The remaining error is an IndexSizeError and occurs in `src/ol/renderer/canvas/tilelayer.js` (Line 364), when `renderContext.drawImage` is called.

From MSDN

> The numeric arguments are not valid (for example, the destination is a 0x0 rectangle). Versions earlier than Internet Explorer 10 use INDEX_SIZE_ERR.

The arguments for 

```js
renderContext.drawImage(tile.getImage(), tileGutter, tileGutter,
        tilePixelSize[0], tilePixelSize[1], left, top, w, h);
```

are:

```
tileGutter             (sx):    0
tileGutter             (sy):    0
tileGutilePixelSize[0] (sw):    256
tileGutilePixelSize[1] (sh):    256
left                   (dx):    -20037499
top                    (dy):    -60112487
w                      (dw):    40075017
h                      (dh):    40075017
```

The same values also come up in Chrome, but there the test passes.

Can anybody share an idea what is wrong there? @ahocevar possibly?

(Above values com from a variant of the current code, where the `img` from `getImage` is an image created with `new Image(10, 10)`, not sure if this has any impact.)

Are the remarks from [this MSDN-page on `canvas element | canvas object`](https://msdn.microsoft.com/en-us/library/ff975062(v=vs.85).aspx) somehow relevant?

> Note  The maximum size of the rendered area on a canvas is from 0,0 to 8192 x 8192 pixels, regardless of the size of the canvas. For example, a canvas is created with a width and height of 8292 pixels. A rectangular fill is then applied as "ctx.fillRect (0,0, canvas.width, canvas.height)".Only the area within the coordinates (0, 0, 8192, 8192) can be rendered, leaving a 100 pixel border on the right and bottom of the canvas.

ahocevar:
Checking this now. Looks like we render to a map with an incorrect size.

ahocevar:
@marcjansen #6000 fixes the test.

PR #5997, opened 2016-10-19
Always load a gated URL polyfill when testing

marcjansen:
Otherwise the tests fail to run in MS Edge, whose user agent doesn't trigger the inclusion of the polyfill without `always` on `polyfill.io`.

With this in we are down from 45 failing tests in MS Edge to 1 remaining failure.

See also #5996

Please review.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -15,7 +15,9 @@
   <script type="text/javascript" src="../node_modules/mocha/mocha.js"></script>
   <script type="text/javascript" src="../node_modules/proj4/dist/proj4.js"></script>
   <script type="text/javascript" src="test-extensions.js"></script>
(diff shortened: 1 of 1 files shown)


marcjansen:
Travis restarted…

tschaub:
Looks good @marcjansen.

Where did you find information on having to always load this for Edge?  I wonder if there are plans to address this upstream (in polyfill.io).

marcjansen:
@tschaub I got it straight from [their API](https://polyfill.io/v2/docs/api), doesn't look like they will remove it in the `v2`-series.

fredj:
I've asked the question upstream: https://github.com/Financial-Times/polyfill-service/issues/941

Issue #5936, opened 2016-10-06
Internet Exlporer 11 + WMSTile Layer with TileGrid and map renderer 'dom' -> IndexSizeError


hanny2:
Hello,

i am currently using some different WMS Layer on the same map, but for some reason the WMS Layer with a tilegrid size of 512x512 is causing an error in IE 11 and Edge (chrome and firefox working).
Then I changed the default map renderer to 'dom' and it also worked in IE and Edge. 

The Error with 'Canvas' renderer is: "IndexSizeError" in ol-debug.js (38069,5). The Error is inside of:
ol.renderer.canvas.TileLayer.prototype.renderTileImages, in line: 

```
renderContext.drawImage(tile.getImage(), tileGutter, tileGutter,
        tilePixelSize[0], tilePixelSize[1], left, top, w, h);
```

My Openlayers Version is 3.18.2. I also testet 3.16.0 and 3.17.0 with the same result.

Here is also some example code of initializing the map and its layers: 

```

var tileGrids = {};
var mapMinZoom = 5;
var mapMaxZoom = 18;
var wmsTileCallid = 0;

var mapProjection = null;
var mapView = null;
var map = null;

var wmsBaseLayer = null;
var wmsLayer0 = null;
var wmsLayer1 = null;
var wmsLayer2 = null;
var wmsLayer3 = null;

var mapProjectionName = 'EPSG:900913';
var wgs84ProjectionName = 'EPSG:4326';
var mapExtent = [ -20037508.34, -20037508.34, 20037508.34, 20037508.34 ];
var mapGWCExtent = [ 270402.7634448332, 5933032.807406616, 1775732.018771907, 7437567.703351199 ];
var mapResolutions = [ 4891.9698095703125, 2445.9849047851562,
                       1222.9924523925781, 611.4962261962891, 305.74811309814453,
                       152.87405654907226, 76.43702827453613, 38.218514137268066,
                       19.109257068634033, 9.554628534317017, 4.777314267158508,
                       2.388657133579254, 1.194328566789627, 0.5971642833948135 ];

function initMap() {
    proj4.defs(mapProjectionName, "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext  +no_defs ");
    mapProjection = ol.proj.get(mapProjectionName);
    mapProjection.setExtent(mapExtent);

[...]

PR #5999, opened 2016-10-19
Use tile image dimensions in rendering

thomasmoelhave:
Some servers sometimes return tiles that do not match the tile size
specification (e.g. it can return a 1x1 blank tile where there's no
data). This can cause an out-of-bound error in drawImage - at least in ie 10+11.

See the discussion in #5936, this patch stems from the comment by @tschaub.

### DIFF
--- a/src/ol/renderer/canvas/tilelayer.js
+++ b/src/ol/renderer/canvas/tilelayer.js
@@ -360,9 +360,8 @@ ol.renderer.canvas.TileLayer.prototype.renderTileImages = function(context, fram
       }
       pixelExtents.push(pixelExtent);
     }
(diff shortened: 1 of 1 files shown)


thomasmoelhave:
I am not sure why the test fails - it does happen on my box and on Travis so probably something to look into.

ahocevar:
You need to account for gutter and different device pixel ratio and source pixel ratio combinations.

hanny2:
Since the dom renderer is removed in 3.19.0 (which isn't a bad thing, because canvas fells better performing and also the rendering when zooming seems a lot better), it would be great to have this one fixed for IE 10 + 11 and Edge.

thomasmoelhave:
Deprecated by #6082.

PR #6000, opened 2016-10-19
Fix mockup of tile layer to keep drawImage arguments in range

### LINKED ISSUE #5996: Failing tests in MS Edge

ahocevar:
Fixes #5996.


marcjansen:
Checking this tomorrow, thanks Andreas

PR #6002, opened 2016-10-20
Use tile pixel ratio for gutter calculation

### LINKED ISSUE #6001: gutter and pixelRatio
Since https://github.com/openlayers/ol3/pull/5273 our tiled WMS are not well aligned on devices with pixelRatio different higher than 1

There is an offset on top-left.

Here is the bug : https://jsfiddle.net/t8ycm06f/2/

Here is how th map should be: https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&X=187963.00&Y=673434.25&zoom=1&layers=ch.swisstopo.fixpunkte-lfp2

In the 2nd link the map is defined with the property pixelRatio to 1. 

We also have a problem with this PR https://github.com/openlayers/ol3/pull/5731 . Detection have also an offset on top-left

@fredj do you have an idea why?

ahocevar:
Fixes #6001.


fredj:
much better, thanks @ahocevar 

I can add the webgl support and fix the rendering test

ahocevar:
Please go ahead @fredj.

fredj:
please squash this commit:
https://github.com/fredj/ol3/commit/3e928dd8381835e821c0663c5b35160bcfd9c46f

ahocevar:
@fredj Cherry-picked your commit without squashing, so you get the credit you deserve 😄

fredj:
Looks good, thanks

PR #5979, opened 2016-10-17
Use FileSaver.js in export-map example

fredj:
Instead of relying on the download attribute.

fixes #2968

PR #6007, opened 2016-10-21
Add note to example that canvas.toBlob() is required

marcjansen:
Follow-up of #5979.


marcjansen:
Example change only, merging without formal review.

PR #6011, opened 2016-10-21
Remove unused function

fredj:
Remove some unused and non api functions


fredj:
thanks for the review

PR #6012, opened 2016-10-22
Combine standard coverage and rendering coverage

marcjansen:
This is open for discussion.

Wouldn't it be nice if we knew the code coverage of our rendering tests? And wouldn't it even be better if we knew the combined coverage of standard-tests and coverage ones?

This PR contains a commit that does that, it is open for your input. Just share what you think.

(PS: with rendering tests participating in the total coverage percentage, we're at ~86%, plus ~4.8%)

### DIFF
--- a/tasks/test-coverage.js
+++ b/tasks/test-coverage.js
@@ -20,6 +20,8 @@ var backupDir = path.join(__dirname, '../src-backup');
 var instrumentedDir = path.join(__dirname, '../src-instrumented');
 var coverageDir = path.join(__dirname, '../coverage');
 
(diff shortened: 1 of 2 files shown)


marcjansen:
this needs more work... it'll take some time for me as i am offline now.

marcjansen:
I think this is ready to be reviewed.

marcjansen:
Anyone up for a review, @openlayers/ol3?

bartvde:
looking at it now @marcjansen

marcjansen:
Thanks for the review, Bart.

PR #6020, opened 2016-10-24
Test that ol.obj.assign throws on undefined/null

marcjansen:
…we did not test that part previously.

Please review.


marcjansen:
Thanks again, Bart.

PR #6027, opened 2016-10-25
Close polygon sketch at all times

### LINKED ISSUE #6023: draw polygon
auxiliary shading is drawn from the second point

issue_openlayer

bjornharrtell:
Fixes #6023.


marcjansen:
Thanks for tackling this. Would you be able to add a test for this regression?

bjornharrtell:
Yes I think so (tomorrow or so) if it's ok to write tests that inspects private members?

marcjansen:
Yes,  we do that elsewhere as well.

bjornharrtell:
Added validation of sketch polygon to existing test.

ahocevar:
Awesome, thanks for your effort @bjornharrtell!

PR #6034, opened 2016-10-27
Use geometry extent's top left corner as pattern/gradient origin

### LINKED ISSUE #6010: Regression: Fill style with CanvasPattern looks worse
Since f655f6740e262f14c376ca55964daacd69707664, Canvas Pattern looks worse at high zoom (> 20).

You can try here : https://openlayers.org/en/latest/examples/canvas-gradient-pattern.html, zoom in 17 times on France for example.

Zoom 19 :
image
Zoom 20 :
image
Zoom 21 :
image

Tested on Chrome 53.0.2785.143 m and Firefox 49.0.1 on Windows 7.

ahocevar:
This change improves the situation with fill patterns at high resolutions. Unless when zooming in on a huge polygon (e.g. country polygon of Canada on a world map) to a very high resolution, patterns will be rendered correctly even at the highest resolution.

This change also impacts how applications should define patterns and gradients, so I added a paragraph to the upgrade notes.

Fixes #6010


fredj:
Thanks @ahocevar

tchandelle:
Tested it in our application, it works great. Thank you.

marcjansen:
Very nice. Should I add this into the upcoming patch release?

ahocevar:
> Very nice. Should I add this into the upcoming patch release?

If you create one anyway, it cannot hurt I'd say.

marcjansen:
I'll include it.

src/ol/render/canvas/polygonreplay.js
```diff
@@ -342,9 +343,13 @@ ol.render.canvas.PolygonReplay.prototype.setFillStrokeStyles_ = function() {
   var lineJoin = state.lineJoin;
   var lineWidth = state.lineWidth;
   var miterLimit = state.miterLimit;
-  if (fillStyle !== undefined && state.currentFillStyle != fillStyle) {
-    this.instructions.push(
-        [ol.render.canvas.Instruction.SET_FILL_STYLE, fillStyle, typeof fillStyle != 'string']);
+  if (typeof fillStyle !== 'string' || fillStyle !== undefined && state.currentFillStyle != fillStyle) {
```
tchandelle: I think there is a problem here. `typeof fillStyle !== 'string' || fillStyle !== undefined` is always true, even when undefined.

PR #6036, opened 2016-10-28
Add new timeout option to ol.interaction.MouseWheelZoom

fredj:
And remove the `ol.MOUSEWHEELZOOM_TIMEOUT_DURATION` define.


fredj:
Thanks for the review

PR #6039, opened 2016-10-28
Do not fill when there is no fillStyle

### LINKED ISSUE #6038: Error when selecting features
When selecting features with Select interaction, there is a failed assert : 

    Assertion failed: 2nd instruction should be a string, CanvasPattern, or CanvasGradient

I've added a comment there, I think this is the cause :  https://github.com/openlayers/ol3/pull/6034#pullrequestreview-6202214

ahocevar:
This is a follow-up on #6034, which fixes the problem mentioned in https://github.com/openlayers/ol3/pull/6034#pullrequestreview-6202214.

Fixes #6038.


marcjansen:
Build restarted.

marcjansen:
I had to restart the build again as this time the test failed while calculating coverage…

marcjansen:
All green, @ahocevar 😉

PR #6074, opened 2016-11-04
Fix typo on style name on Modify Features Test example

tchandelle:


bartvde:
thanks again @tchandelle

PR #6075, opened 2016-11-04
Keep the other dim from the original segment when modifying vertex

tchandelle:
Previously, when modifying a 3d geometry, the third dimension was lost when you move a vertex, replaced by zero. This commit fixes it.

However, when adding a new vertex in a segment, the third dimension will still be zero. Should it be the average between the previous and next vertex?


marcjansen:
This looks very solid. I hope to have a deeper look the next days.

tchandelle:
Is there a chance to have this in next version, we would really appreciate this correction.
@marcjansen ?

PR #6076, opened 2016-11-04
New rotate features interaction

tchandelle:
A new interaction that allow user to rotate features.

I would like feedback about a few things :
- Are the `RotateOptions` correctly named?
- What informations should the `RotateEvent` contains?
- Is there a better way to determine the default anchor than the center of the extent?
  - the problem with the current method is that if I rotate a feature at angle *α*, then back at angle *-α*, it may not return at its original position because its extent may have changed between the two operations
- Should the position of the handle configurable? Which corners? With a margin?


tchandelle:
For your convenience, you can test it here : https://run.plnkr.co/plunks/KK7XDhi1SRh2nTJkXLim/examples/rotate-features.html

rashadkk:
is ol.interaction.Rotate available in openlayers 4 ?
i got the following error

TypeError: ol.interaction.Rotate is not a constructor

PR #6078, opened 2016-11-05
Use winding number algorithm for linearRingContainsXY

bjornharrtell:
Found out about this algorithm in another project and it seems nice so I propose to replace the old one.

### DIFF
--- a/src/ol/geom/flat/contains.js
+++ b/src/ol/geom/flat/contains.js
@@ -36,22 +36,30 @@ ol.geom.flat.contains.linearRingContainsExtent = function(flatCoordinates, offse
  * @return {boolean} Contains (x, y).
  */
 ol.geom.flat.contains.linearRingContainsXY = function(flatCoordinates, offset, end, stride, x, y) {
-  // http://www.ecse.rpi.edu/Homepages/wrf/Research/Short_Notes/pnpoly.html
-  var contains = false;
+  // http://geomalgorithms.com/a03-_inclusion.html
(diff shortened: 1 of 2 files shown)


bartvanandel:
Why? Is it faster?

bjornharrtell:
Is is slightly faster in some cases and as fast in all cases according to the author. It can also handle non simple polygons.

Sources:
https://en.wikipedia.org/wiki/Point_in_polygon
http://geomalgorithms.com/a03-_inclusion.html

bartvanandel:
Interesting! I think your latter point is even more important: for polygons with self-intersections the currently used Crossing Number algorithm will give wrong answers in the parts where the polygon overlaps with itself an even number of times. The Winding Number algorithm does not have this problem.

marcjansen:
hey @bjornharrtell,  any chance of adding tests for   non simple polygons?

bjornharrtell:
You got it @marcjansen. :)

marcjansen:
Thanks for adding the tests, @bjornharrtell. Is it just me or are these tests passing with or without changing the algorithm?

I still think this looks good, but just want to confirm.

bjornharrtell:
The test `returns true for point inside an overlap of a non-simple polygon` should fail with the existing algorithm. I just cherry picked d34409846fc47d9499a1c52a98c28a4968c011d1 on master and it fails as expected. Though I noticed that I had forgot to change the description of the test cases, I've fixed that now.

marcjansen:
Thanks for this @bjornharrtell.

PR #6079, opened 2016-11-05
Add view animation

### LINKED ISSUE #5234: rework ol.animations
from #2588:

I'm planning on reworking how animation is done.  Instead of setting the final view state and then adding functions that tween the frame state, the view will be animated.  This will make synchronized maps work as expected during animated transitions.

Instead of a `beforeRender` function on the map, the view may have an `animate` method.

tschaub:
This adds a `view.animate()` method that allows running animations in series or parallel.

For example, to animate both center and zoom:
```js
view.animate({
  center: ol.proj.fromLonLat([-110, 45]),
  zoom: 10
});
```

To run two animations in series:
```js
view.animate({
  center: ol.proj.fromLonLat([-110, 45]),
  zoom: 10
}, {
  rotation: view.getRotation() + 2 * Math.PI
});
```

A callback can be provided that gets called at the end of the animation series:
```js
view.animate({
  zoom: 10
}, function(complete) {
  // complete is false if a user interaction interrupted the animation
});
```

Animation options can include a `duration` and `easing` function.  Two or more animations can be run in parallel by calling `view.animate()` more than once.  All animations are cancelled when the user interacts with the map or when anything calls `view.setCenter()`, `view.setResolution()`, or `view.setRotation()`.

This removes the need for all the `ol.animation` functions and the `map.beforeRender()` method.  And, since we now actually animate the view, we can now get the expected behavior when two maps share a view (they both animate together).

shared-views

A snapshot of the docs:

image

Fixes #5234.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,40 @@
 ## Upgrade notes
 
+### Next release
+
+#### Use `view.animate()` instead of `map.beforeRender()` and `ol.animation` functions
+
+The `map.beforeRender()` and `ol.animation` functions have been deprecated in favor of a new `view.animate()` function.  Use of the deprecated functions will result in a warning during development.  These functions are subject to removal in an upcoming release.
+
+For details on the `view.animate()` method, see the API docs and the view animation example.  Upgrading should be relatively straightforward.  For example, if you wanted to have an animated pan, zoom, and rotation previously, you might have done this:
+
+```js
+var zoom = ol.animation.zoom({
+  resolution: view.getResolution()
+});
+var pan = ol.animation.pan({
+  source: view.getCenter()
+});
+var rotate = ol.animation.rotate({
+  rotation: view.getRotation()
+});
+
+map.beforeRender(zoom, pan, rotate);
+
+map.setZoom(1);
+map.setCenter([0, 0]);
+map.setRotation(Math.PI);
+```
+
+Now, the same can be accomplished with this:
+```js
+view.animate({
+  zoom: 1,
+  center: [0, 0],
+  rotation: Math.PI
+});
+```
+
 ### v3.19.1
 
 #### `ol.style.Fill` with `CanvasGradient` or `CanvasPattern`
--- a/doc/tutorials/introduction.md
+++ b/doc/tutorials/introduction.md
@@ -34,7 +34,7 @@ The library is intended for use on both desktop/laptop and mobile devices.
 OL3 uses a similar object hierarchy to the Closure library. There is a top-level `ol` namespace (basically, `var ol = {};`). Subdivisions of this are:
 
 * further namespaces, such as `ol.layer`; these have a lower-case initial
-* simple objects containing static properties and methods, such as `ol.animation`; these also have a lower-case initial
+* simple objects containing static properties and methods, such as `ol.easing`; these also have a lower-case initial
 * types, which have an upper-case initial. These are mainly 'classes', which here means a constructor function with prototypal inheritance, such as `ol.Map` or `ol.layer.Vector` (the Vector class within the layer namespace). There are however other, simpler, types, such as `ol.Extent`, which is an array.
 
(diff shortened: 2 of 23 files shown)


tschaub:
With the changes here, we no longer use `map.beforeRender()` or any of the `ol.animation` functions in the library.  What do others think about logging a deprecation warning in `map.beforeRender()` and removing it and the `ol.animation` functions in a subsequent release?

tschaub:
At a minimum, I'll add something to the release notes that encourages people to use `view.animate()`.  We could also log a deprecation warning - I imagine that more people would read that rather than the release notes.  If we want to remove `map.beforeRender()` and the `ol.animation` functions altogether, it would look like 3e5024c3a5d43f065e211c024ffefd1c13d4fcfa (and improve our test coverage).

tschaub:
Ok, I opted for deprecation warnings instead of code removal.  This should be ready for review now.

marcjansen:
I really like this change! Code is looking good to me. I am happy you took the deprecation way.

Please merge and many thabks for this nice API improvement.

marcjansen:
Two questions:

* Would it make sense to expose `cancelAnimations_` in the public API? Might make sense to call this when a user wants to assign a new `view`
* Do we leave it up to the user if animations can be cancelled at all by user input? Or does it make sense to configure an animation with e.g. `cancelable`?

Otherwise this is still looking very good to me.

tschaub:

Good questions & suggestions.  Here are my thoughts:

> Would it make sense to expose cancelAnimations_ in the public API? Might make sense to call this when a user wants to assign a new view

If a user does `map.setView(newView)`, any running animations on the old view will have no effect (nobody should be listening for view changes any more).  Another way to cancel an animation is to do something like `view.setCenter(view.getCenter())`.  I grant that this is not as nice as `view.cancelAnimations()`.  But I'm inclined to get this in with a minimal API and see if there are solid use cases for expanding on it.

> Do we leave it up to the user if animations can be cancelled at all by user input? Or does it make sense to configure an animation with e.g. cancelable?

I like that the new animations are consistently cancelable.  This is not the case with the `map.beforeRender()` and `ol.animation` functions (for example, you get weird results if you start dragging the map on the current animation example).  With the new animations, it would be possible to add a `cancelable` property.  This would mean that calls to `view.setCenter()` et al. would be *ignored* - which is a bit weird.  As above, rather than adding a property to the API with uncertain use, I'd rather wait to see if there is a solid use case that we want to support.

ahocevar:
@tschaub This change introduces a problem. Maybe another issue with the tile queue. It can be reproduced with all tile layer examples (e.g. simple.html): Press the "Zoom out" button, and press it immediately again (without waiting for the animation to finish). No new tiles will be loaded, you have to pan the map a bit to get them loaded.

tschaub:
> @tschaub This change introduces a problem.

I've opened #6102 to track this.  I assume it is because the animation hint is still set in the final render.  This is likely the same issue we have been dealing with when trying to remove the seemingly unnecessary render calls in interactions.  I'll confirm this now and make it so the final frame of an animation does not have the animation hint set if this turns out to be the problem.

ahocevar:
Unfortunately e447df39f7a8929026185d4f6c4d7bee24b2685d causes a regression, which I ticketed as #6118.

Status420:
I've updated to 4.0.1 and noticed the removal of beforeRender, but I don't  think you have taken in to consideration where this might have been used for running a function before the rendering takes place, in my case I need to tell the map to do something if someone tries to zoom the map so I do this.
` map.getView().on('propertychange', function(evt) {
                // Detect zoom change.
                if (evt.key === 'resolution'){
                    view.animate
                    map.beforeRender(function(_map, frameState) {
                        _onMapZoom(_map, frameState)
                    })     
                }
            });,` 
in the _onMapZoom( function is dispatch events to tell the map to switch something off, so how am I supposed to handle this now?
I need to trigger this event when the mqp zooming starts.

ahocevar:
@Status420 What about something like
```js
map.getView().on('change:resolution', function(evt) {
  map.on('precompose', function(map, frameState) {
     _onMapZoom(map, frameState);
  });
});
```

Status420:
@ahocevar thank you, will give that a try.

PR #6082, opened 2016-11-06
Unified canvas rendering

### LINKED ISSUE #5601: 'forEachLayerAtPixel' tile layer method does not account for pixelRatio of device.

### LINKED ISSUE #5690: White lines for Zoomify source starting from 3.14.0

### LINKED ISSUE #5821: forEachLayerAtPixel with ol.layer.Image broken for rotated views and device pixel ratio <> 1
I have implemented a tooltip function with popups with forEachLayerAtPixel function. It works fine on desktop PC's, but most of mobile devices not. The forEachLayerAtPixel works fine on icon and polygon but on images is founded the image layer not in correct place. If i click a little left-buttom near the image, then the layer is founded, so the place (founded layers on a pixel) is calculated wrong in case of images. (I think the difference, that icon and polygon are vector layers, but the image not.) After some analyse we have found, that it caused, when the css ratio is not 1 (window.devicePixelRatio <> 1). 
- at rotation (alt+shift+mouse) it works faulty also on desktop! The calculated place remaining somewhere on the original place, it's not moving with the picture.

I've attached a sample program:
[popup-1.zip](https://github.com/openlayers/ol3/files/449112/popup-1.zip)

### LINKED ISSUE #5936: Internet Exlporer 11 + WMSTile Layer with TileGrid and map renderer 'dom' -> IndexSizeError

ahocevar:
This pull request is the result of a cleanup and refactoring of the canvas renderer. In particular, its goals are:

* Share as much code as possible between the layer renderers.
* Use transforms instead of pixel calculations wherever possible. It turned out that pixel calculations for tile positioning were hard to understand and error prone.
* No more direct composition of tiles to the target canvas - the performance gain was minimal, at the cost of code that was hard to read and understand, and poor rendering quality during zoom animations. The new image composition code is more efficient than the one we had before the direct composition refactoring. It creates a canvas with all tiles that intersect the view extent and reuses it as long as that extent fully contains the view extent.
* Avoid overlapping rendering of vector tiles of different resolutions during loading. This is done by rendering replays for high resolution tiles first, and cut extents of already rendered high resolution tiles out of lower resolution tiles that are rendered later.

This refactoring also fixes several bugs:

Fixes #5690.
Fixes #5936 (alternative to #5999).
Fixes #5821.
Fixes #5601.


ahocevar:
This is now ready for review.

tschaub:
Impressive work @ahocevar.  I haven't found time for a proper review yet.  I'd also like to bring a build of this branch into an application where we often render 30-40 tile layers at once.  I was guessing that we were benefitting from not creating an intermediate canvas for each.  I should be able to benchmark and profile the two approaches.

tschaub:
Here are some rendering results with built versions of the library before and after this change.

First, two animated zoom loops with a full tile cache.

Zooming with the single canvas (old) approach:

old

Zooming with the multiple canvas (new) approach:

new

tschaub:
Here are two timelines during zoom animations.  I've zoomed in on a single animated transition in both cases.  The second section (with horizontal pink bars) shows the frame time.

Zooming with the single canvas (old) approach:

old

Zooming with the multiple canvas (new) approach:

new

In the second timeline, there are more "long" frames (e.g. the 110ms frames or ~9fps).  The third section of the charts highlights the JS heap over time.

tschaub:
Not sure if there is too much to surmise from the graphics above.  These are perhaps extreme examples, with 30 layers in each case.  Ordinarily, I think that number of layers would be hard to justify - we've been taking advantage of the fact that there is not much additional cost per layer with the current (or old) approach.  We can rework this particular example to use server-side tile composition, resulting in far fewer client-side layers being visible at once.

Clearly there is value in addressing the rendering issues that motivated this change.  So perhaps that's enough to outweigh any drop in rendering performance (this drop might not be noticeable in most cases).

ahocevar:
Tough decision. A drop in performance is not a good thing, but 30 layers is indeed a lot. I tried a map with 4 layers, and didn't *feel* much difference. Maybe @marcjansen can also weigh in here, because I know he's been struggling with performance on maps with many layers as well.

In addition to what @tschaub said, This change also makes the tile renderer code much easier to maintain. Maybe also something to consider.

tschaub:
I've reviewed the code changes and think they look great.

I'll admit that I still don't get why composition to an intermediate canvas is required. I wonder if there is a simple way to demonstrate what is possible by rendering to two intermediate canvases and then composing to a third that would not be possible by rendering directly to a single canvas.  I can imagine cases involving opacity changes on the intermediate canvases (where we might have overlapping images rendered) - but I don't *think* that is the kind of thing we're trying to solve here.

@ahocevar is it possible to illustrate the kind of thing that is not possible with direct composition in a simple static example?  I understand if this would take too much effort to produce, but it would be great to be able to illustrate the problem very directly.  Are we dealing with bugs/limitations in the Canvas implementations in certain browsers?  Or is this about something that is really not possible with the Canvas API?

My comments shouldn't be taken as blocking this change.  The code looks really nice.  I'm just not the right one to push for it to get in since I have not been hit by any of the existing bugs that this addresses and am hit by the drop in performance.  If this improves things for most people, then I'm in favor.  I would just be a bit more satisfied if I understood that the existing problems cannot be solved without a performance hit.

ahocevar:
@tschaub Simply put, the problem with rendering directly to the map canvas is that as soon as a scale or rotate transform is applied to the map canva, you get gaps between tiles.

tschaub:
Curious if we could scale and rotate the map canvas in CSS.  This would mean generating a larger map canvas than the viewport - but that is also required by the intermediate canvas alternative.

ahocevar:
I thought the CSS scaling and rotation approach through. Rotation would work, but scaling wouldn't. This is because we could have tile layers with different nominal tile resolutions (depending on their tile grids). So CSS scaling could only be applied when rendering at fractional *view* zoom levels, but we'd still have a problem with fractional *tile source* zoom levels.

ahocevar:
@tschaub I pushed 7fc5379 with a minimal change to render tiles to the map canvas directly. Here is the result with the bing-maps example:
nov-13-2016 09-56-03
If you're able and willing to spend some time on this, we could try to find a solution together. Maybe I'm just doing something wrong and the gaps when rendering to a transformed canvas can be avoided? I think the renderer code in this pull request is simple enough to make it easy to experiment with it.

twpayne:
The reasons for rendering tile layers to an intermediate canvas were:

1. Visual quality: To avoid tile seams. I don't know of any other way to avoid this. When the map is scaled fractionally or rotated, then pixel values will be interpolated from pixels in the tiles. Pixels near tile edges need to be interpolated from pixels from two tiles (at the edges) or four tiles (at the corners). If the tiles are rendered one-by-one, the data required for interpolation at tile edges and corners is simply never available.

2. Visual quality: To simply the rendering of interim tiles (the use of a low resolution tile while waiting for high resolution tiles to load.

3. Performance: during interactions, the intermediate canvas can reused without being redrawn (it's a single draw call at the graphics layer). This makes zooming and rotation very fast, even on low power devices (e.g. mobile), and reduces garbage generation and so avoids jank.

marcjansen:
I hope to test the performance impact in one of our apps with many layers and report back on monday

ahocevar:
@marcjansen When you test, please make sure you test with fdb51a1 (and not 4814538). Thanks!

marcjansen:
I am sorry to say that I couldn't find time to properly test this yesterday. I am very unsure if I will have time the next days.

ahocevar:
@tschaub 9c3cf13ff75090f8b7ae281fff6e5decc3406ade is a commit on top of this pull request which still does everything with transforms, but with direct composition (and resulting in gaps when rendering rotated or scaled views). Everything works with that commit, so it should be easy to experiment with it.

bartvanandel:
Wouldn't relying on CSS for rotation (and scaling) cause issues with text labels that don't scale and rotate with the map? Or is this just for the tiles? Note that I'm not working on this issue, I just stumbled upon it.

ahocevar:
@marcjansen Any chance you could try again and set aside some time to test this? What I want to know is whether the performance is acceptable or not. I know that the current approach is faster than what I'm proposing here, which is a change to simplify the code base and make it more readable and maintainable.

marcjansen:
Hey @ahocevar,

I have had a look at the effects of the changes in this PR. I took the `wms-tiled ` and issued the following lines in the console:

```js
var numLayers = 30;
var opacity = 1 / numLayers;
while(--numLayers > 0) {
  var s = new ol.source.TileWMS({
    url: 'http://ows.terrestris.de/osm/service',
    params: {LAYERS: 'OSM-WMS'}
  });
  var l = new ol.layer.Tile({
    source: s,
    opacity: opacity
  });
  map.addLayer(l);
}
```

then I zoomed in and out, while panning around. 

This is not a real setup (30 times the same layer isn't the usual case, right?), but I saw what you have already stated: the code-execution in this branch feels less snappy compared to master. I cannot put this into something measurable, only my feelings while playing with the example. Map interaction feels slower.

I cannot qualify how much more cleaned up the code is (wrt to future changes, improvements and maintenance). 

I hope this findings help at least a little bit, Andreas.

ahocevar:
Thanks @marcjansen, this helped indeed. I now made a significant improvement: when a tile layer renderer has an image with tiles already, the image will now be used if too much time was spent already in a renderer frame. This was, image composition to the intermediate canvas no longer need to happen in the same render frame for all layers, which significantly improves user experience and reduces CPU/GPU load.

@tschaub, you may want to give this a final test with your setup as well.

tschaub:
The latest changes do look like a significant improvement compared to the last commit I tested.  I'm still seeing more long frames in this branch compared to master, but fewer than before.  Testing with something like @marcjansen's example would make for more straightforward profiling.

Reviewing the code now.

src/ol/renderer/canvas/layer.js
```diff
@@ -214,12 +215,21 @@ ol.renderer.canvas.Layer.prototype.prepareFrame = function(frameState, layerStat
...
+ * @param {ol.Coordinate} coordinate Coordinate.
+ * @param {olx.FrameState} frameState Frame state.
+ * @param {function(this: S, ol.layer.Layer, (Uint8ClampedArray|Uint8Array)): T} callback Layer callback.
+ * @param {S} thisArg Value to use as `this` when executing `callback`.
+ * @return {T|undefined} Callback result.
+ * @template S,T
  */
-ol.renderer.canvas.Layer.prototype.getPixelOnCanvas = function(pixelOnMap, imageTransformInv) {
-  return ol.transform.apply(imageTransformInv, pixelOnMap.slice());
+ol.renderer.Layer.prototype.forEachLayerAtCoordinate = function(coordinate, frameState, callback, thisArg) {
```
tschaub: Should this be `ol.renderer.canvas.Layer`?

test/spec/ol/renderer/canvas/tilelayer.test.js
```diff
@@ -50,7 +53,7 @@ describe('ol.renderer.canvas.TileLayer', function() {
       var canvas = document.createElement('canvas');
       canvas.width = 200;
       canvas.height = 100;
-      var context = {
+      context = {
```
tschaub: Is this intentionally global?

tschaub:
@ahocevar - thanks for your persistence on this.  I left a few minor comments that I think are worth addressing, but otherwise I think this looks good to go.

ahocevar:
Thanks for testing @tschaub and @marcjansen, and thanks for the thorough review @tschaub.

Benjaki2:
@ahocevar I appears that the `map.forEachLayerAtPixel` method is broken by the merge. At least for canvas.tilelayer renderers. The context [here](https://github.com/openlayers/ol3/blob/621589b318aff3cb704dd63e018e9810c1e75af0/src/ol/renderer/canvas/tilelayer.js#L283) is dimensionless which leads me to believe it isn't being precomposed. Any Ideas?

ahocevar:
Do you have a JSFiddle to help me reproduce this issue? As soon as you see the image rendered, the context will have dimensions and contain the image you see. You can also see it working in the `getfeatureinfo-tile.html` example.

Are you trying to call `map.forEachLayerAtPixel()` before anything is rendered?

Benjaki2:
@ahocevar After further testing it looks what I was seeing is related to the zoom. If you look at `getfeatureinfo-tile.html` after zooming around you should see that you are getting 'hits' when you shouldn't.

ahocevar:
@Benjaki2 I created #6211 for this. Feel free to create a pull request with a fix if you're willing/able to.

Benjaki2:
@ahocevar I will dig into it in the next day or two.

ahocevar:
The remaining issue mentioned by @Benjaki2 has been fixed with #6222.

PR #6091, opened 2016-11-08
Apply pixelRatio to line dash

### LINKED ISSUE #6090: DevicePixelRatio is not considered at dashed lines
The dash line calculation of open layers depends on line width parameter when painting line, but not increasing the space and line length (parameter "lineDash:")accordingly. 
On desktop browsers, where the cssratio=1, the dashed lines are remaining always dashed lines. 
On mobile devices, where the css ratio can be higher (e.g LG Nexus -> cssratio=3), then the dashed line width is corrected with css ratio, but the line/space length not. It means, we can set such configuration, where the line is dashed on desktop, but on some mobile is seems as a continuous line (configuration type: relative dick line with short dash parameters).

I made some screeshots about the problem. The line style definition seems so in the code:
	  _var blueStyle = new ol.style.Style({
          stroke: new ol.style.Stroke({
            color: 'blue',
            width: 3,
		lineDash: [2,6]
          }),_

Parameters: width=3, lineDash=2,6. Desktop (cssRatio=1)
screenshot-chrome-desktop-3-2-6_2

Parameters: width=3, lineDash=2,6. Android on Chrome browser (cssRatio=3)
screenshot-chrome-android-3-2-6_2

Parameters: width=6, lineDash=30,30. Desktop (cssRatio=1)
screenshot-chrome-desktop-6-30-30_2

Parameters: width=6, lineDash=30,30. Android on Chrome browser (cssRatio=3)
screenshot-chrome-android-6-30-30_2

tchandelle:
Fixes #6090


src/ol/render/canvas/replay.js
```diff
@@ -525,7 +525,10 @@ ol.render.canvas.Replay.prototype.replay_ = function(
         context.lineJoin = /** @type {string} */ (instruction[4]);
         context.miterLimit = /** @type {number} */ (instruction[5]);
         if (ol.has.CANVAS_LINE_DASH) {
-          context.setLineDash(/** @type {Array.<number>} */ (instruction[6]));
+          var lineDash = instruction[6].map(function(dash) {
```
ahocevar: Since this operation is on the hot path, it would be better to reuse the modified lineDash, instead of creating a new array every time. Also, there is no need to modify the lineDash array when the pixel ratio is 1.

tchandelle:
@ahocevar I don't see how I should a reuse the previous line dash array.

ahocevar:
@tchandelle The easiest would be to add the pixel ratio as another instruction with SET_STROKE_STYLE. It would initially be 1. When replaying at a different pixel ratio, you'd store the modified lineDash in the instruction set, and update the pixel ratio. So when creating the replay:
```js
      this.instructions.push([
        ol.render.canvas.Instruction.SET_STROKE_STYLE,
        strokeStyle, lineWidth, lineCap, lineJoin, miterLimit, lineDash, true, 1
      ]);
```
And when replaying:
```js
var lineDash = /** @type {Array.<number>} */ (instruction[6]);
var renderedPixelRatio = instruction[8];
if (usePixelRatio && pixelRatio !== renderedPixelRatio) {
  lineDash = lineDash.map(function(dash) {
    return dash * pixelRatio / renderedPixelRatio;
  });
  instruction[6] = lineDash;
  instruction[8] = pixelRatio;
}

tchandelle:
@ahocevar : I've added both tests. I'm not sure if I've implemented the unit test the correct way.

fredj:
@ahocevar do you want to take a final look?

PR #6098, opened 2016-11-09
Fix some type errors

sbrunner:
Kind of error I have without that:
https://travis-ci.org/camptocamp/ngeo/jobs/174469862#L614-L631

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -2,6 +2,7 @@ goog.provide('ol.featureloader');
 
 goog.require('ol');
 goog.require('ol.Tile');
+goog.require('ol.VectorTile');
 goog.require('ol.format.FormatType');
 goog.require('ol.xml');
 
--- a/externs/oli.js
+++ b/externs/oli.js
@@ -48,6 +48,9 @@ oli.events.Event.prototype.preventDefault = function() {};
  */
 oli.events.Event.prototype.stopPropagation = function() {};
 
+/**
+ */
+oli.Collection = {};
 
 /**
  * @interface
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
       "no-constant-condition": 0,
       "openlayers-internal/no-duplicate-requires": 2,
       "openlayers-internal/no-missing-requires": 1,
-      "openlayers-internal/no-unused-requires": 2,
       "openlayers-internal/one-provide": 1,
       "openlayers-internal/requires-first": 2,
       "openlayers-internal/valid-provide": 2,
--- a/src/ol/renderer/canvas/vectortilelayer.js
+++ b/src/ol/renderer/canvas/vectortilelayer.js
@@ -137,6 +137,7 @@ ol.renderer.canvas.VectorTileLayer.prototype.renderTileReplays_ = function(
   var alpha = replayContext.globalAlpha;
   replayContext.globalAlpha = layerState.opacity;
 
+  /** @type {Array.<ol.VectorTile>} */
   var tilesToDraw = this.renderedTiles;
   var tileGrid = source.getTileGrid();
 
@@ -283,6 +284,7 @@ ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = functi
   /** @type {Object.<string, boolean>} */
   var features = {};
 
+  /** @type {Array.<ol.VectorTile>} */
   var replayables = this.renderedTiles;
   var source = /** @type {ol.source.VectorTile} */ (layer.getSource());
   var tileGrid = source.getTileGrid();


src/ol/featureloader.js
```diff
@@ -87,8 +87,10 @@ ol.featureloader.tile = function(url, format) {
        * @this {ol.VectorTile}
        */
       function(features, dataProjection) {
-        this.setProjection(dataProjection);
-        this.setFeatures(features);
+        /** @type {ol.VectorTile} */
+        var me = this;
```
sbrunner: I suspect That he takes this definition: https://github.com/sbrunner/ol3/blob/95d30c1cbf592c49b67ff4247dda8b5fae43c076/src/ol/featureloader.js#L12

The `setState` from the function bellow, is also available in the `Source`...

sbrunner: What do you mean with grouping?

I just also try this but it's not working :-(
```
  /** 
   * @this {ol.VectorTile}
   * @param {Array.<ol.Feature>} features The loaded features.
   * @param {ol.proj.Projection} dataProjection Data projection.
   */
  var success = function(features, dataProjection) {
    this.setProjection(dataProjection);
    this.setFeatures(features);
  };  
  /** 
   * @this {ol.VectorTile}
   */
  var failure = function() {
    this.setState(ol.Tile.State.ERROR);
  };  
  return ol.featureloader.loadFeaturesXhr(url, format, success, failure);
```
The error message I have is:
```

ERR! compile /home/external/workspace/ngeo/node_modules/openlayers/src/ol/featureloader.js:89: ERROR - Property setProjection never defined on this
ERR! compile     this.setProjection(dataProjection);
ERR! compile          ^
ERR! compile 
ERR! compile 
ERR! compile /home/external/workspace/ngeo/node_modules/openlayers/src/ol/featureloader.js:90: ERROR - Property setFeatures never defined on this
ERR! compile     this.setFeatures(features);
ERR! compile          ^
ERR! compile 
ERR! compile 
ERR! compile 2
ERR! compile  error(s), 
ERR! compile 0
ERR! compile  warning(s), 
ERR! compile 96,4
ERR! compile %
ERR! compile  typed
ERR! compile 
ERR! compile 
```

ahocevar: I'd try the following:
```js
  /** 
   * @param {Array.<ol.Feature>} features The loaded features.
   * @param {ol.proj.Projection} dataProjection Data projection.
   * @this {ol.VectorTile}
   */
  function success(features, dataProjection) {
    this.setProjection(dataProjection);
    this.setFeatures(features);
  };  
  /** 
   * @this {ol.VectorTile}
   */
  function failure() {
    this.setState(ol.Tile.State.ERROR);
  };
```
The same (function without `var`, `@this` as last annotation) works everywhere else.

```

Comment (gberaudo) [inline src/ol/featureloader.js]: Thanks @ahocevar for the suggestion. Unfortunately I just tried it but I got the same error.

Comment (gberaudo) [inline src/ol/featureloader.js]: OK, I got it working by add a dependency:
```
goog.require('ol.VectorTile');
```.
It is a bit of a pity though since it forces every one using the feature loarder to embed the code for vector tile. I think the solution from @sbrunner is preferable in this case.

Comment (tschaub) [inline src/ol/featureloader.js]: I don't see any reason why `ol.featureloader.tile` needs to be in this module.  This could be moved to the only place where it is used (in `src/ol/source/vectortile.js`).  I'd make it a method of `ol.source.VectorTile`.

I also don't see any place where `ol.featureloader.loadFeaturesXhr` is used.  Looks like this whole file could be removed.

Please don't add workarounds for behavior we don't understand about the compiler.

Comment (sbrunner) [inline src/ol/featureloader.js]: I try the solution from @gberaudo (and update this pull request),
I also move back the `oli.js`, and `olx.js` back to `--js` in the build and now it's working whell, but I have an other question: the files `oli.js`, and `olx.js` are in the `external`folder but they are not external files in the GCC meaning..., When why they are not named as (with folder) respectively `src/ol/interfaces.js` and `src/ol/export.js`, like the `src/ol/typedef.js`.

And finally we should add this to make everything working:
https://github.com/camptocamp/ngeo/pull/2009/commits/0f2554c3c1e06d0a8a05f0c0c76d447d1c606932#diff-b835803948af99299e646c466f3127f5R6
```
// Required by olx
/** @suppress {extraRequire} */
goog.require('ol.format.IGC');
/** @suppress {extraRequire} */
goog.require('ol.Overlay');
/** @suppress {extraRequire} */
goog.require('ol.control.ScaleLine');
/** @suppress {extraRequire} */
oog.require('ol.source.WMTS');
/** @suppress {extraRequire} */
goog.require('ol.style.Icon');
/** @suppress {extraRequire} */
goog.require('ol.layer.VectorTile');
```
Shouldn't be this also added in the `olx.js` file?

sbrunner: After a cleanall I have the same error as @gberaudo :-(

tschaub:
See #6105 for a proposal to move the vector tile loader function into the vector tile module.

tschaub:
See #6106 for a proposal to add type annotations to address some of the build failures in https://travis-ci.org/camptocamp/ngeo/jobs/174469862.

sbrunner:
Looks good, thanks @tschaub :-)

PR #6099, opened 2016-11-09
Modify interaction tests : check for change events

tchandelle:
Most of the time, there is only 3 events. This loop is supposed to check the middle one, but never does.

### DIFF
--- a/test/spec/ol/interaction/modify.test.js
+++ b/test/spec/ol/interaction/modify.test.js
@@ -140,7 +140,7 @@ describe('ol.interaction.Modify', function() {
     // make sure we get change events to events array
     expect(events.length > 2).to.be(true);
     // middle events should be feature modification events
-    for (var i = 1; i < events.length - 2; i++) {
+    for (var i = 1; i < events.length - 1; i++) {
       expect(events[i]).to.equal('change');
     }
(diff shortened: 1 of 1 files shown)


test/spec/ol/interaction/modify.test.js
```diff
@@ -140,7 +140,7 @@ describe('ol.interaction.Modify', function() {
     // make sure we get change events to events array
     expect(events.length > 2).to.be(true);
     // middle events should be feature modification events
-    for (var i = 1; i < events.length - 2; i++) {
+    for (var i = 1; i < events.length - 1; i++) {
```
ahocevar: I don't think this is correct. Look a few lines above, at the `endevent` check, which checks `events[events.length - 1]`. If tests still pass after this change, we're probably doing something very wrong.

tchandelle: If `events.length = 3`, `events = [modifystart, change, modifyend]`, the loop is only on 1, which is `change`. I don't think there is anything wrong

ahocevar: Oh, you're right - `<` and not `<=`. Thanks for catching this.

Issue #3803, opened 2015-06-15
Enhancement: Make ol.style.Icon scalable in both X and Y.


pconnell99:
Currently the scale parameter of an **ol.style.Icon** has a scale parameter that takes a single float value between 1 and 0 the defines the scale of the icon.  To make the Icons more useful to model real-world items (cars, planes, ships etc.) they would need to be scalable in both X and Y axis.

Suggested that this is done to allow this.  What would be involved?  I have been looking in **ol.render.canvas.Replay.prototype.replay_** and 'fiddling' with values in there seems to perform scaling and rotation.

pconnell99:
Feeding in custom x/y coords (debug) into replay_ (case **ol.render.canvas.Instruction.DRAW_IMAGE**) and moving the rotation IF clause in **ol.vec.Mat4.makeTransform2D** above the scale works - no skewing.

I'm manipulating the scale, scale clause in 
`ol.vec.Mat4.makeTransform2D(
                localTransform, centerX, centerY, scale, scale,
                2, -centerX, -centerY);`

pconnell99:
Nobody?

tschaub:
It would be nice to allow `scale` to be either a `number` or `Array.<number>` (instead of adding `scaleX` and `scaleY`).  Not sure how others feel about that.

pconnell99:
That is what I did, it passed build and works, but making the document/api updates is a bit beyond my knowledge of the codebase.

Is anyone willing to 'sponsor' (i.e. do) thus update for me?


On 25 June 2015 16:18:20 BST, Tim Schaub <notifications@github.com> wrote:
>It would be nice to allow `scale` to be either a `number` or
>`Array.<number>` (instead of adding `scaleX` and `scaleY`).  Not sure
>how others feel about that.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/openlayers/ol3/issues/3803#issuecomment-115289772

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

pconnell99:
Looking at the codebase I'll need to rewire the webgl render a bit as well to use scale[0] and scale[1] throughout.  As to the scale parameter - as it's still experimental, should we just switch to an Array<number> ongoing, defaulted to [1,1] if not set?

[...]

Issue #2720, opened 2014-09-17
Size properties on ol.style.Icon


lenoil:
Hi,

the modification of the  size  parameter has no effect.

In this example http://openlayers.org/en/v3.0.0/examples/icon.html
I try to change the size of the image but it doesn't work ... why ?

``` js
var iconStyle = new ol.style.Style({
  image: new ol.style.Icon(/** @type {olx.style.IconOptions} */ ({
    /*anchor: [0.5, 46],
    anchorXUnits: 'fraction',
    anchorYUnits: 'pixels',*/
    opacity: 1,
    size : [100,50],
    rotation: 40 * Math.PI / 180,
    src: 'data/icon.png'
  }))
});
```

Thanks

greggian:
I am having a similar problem with ol.stlye.Icon.size in OL version 3.1.1.

In my situation my src image is slightly larger, say 25 x 25px. If I set a size on my ol.style.Icon of 20 x 20px then my image is truncated rather than scaled (as graphicHeight/graphicWidth did in OL2). 

```javascript
var iconStyle = new ol.style.Style({
  image: new ol.style.Icon(/** @type {olx.style.IconOptions} */ ({
    size : [20,20],
    src: 'data/icon.png' //25 x 25px image
  }))
});
```
![screenshot from 2015-01-09 15 46 45](https://cloud.githubusercontent.com/assets/107471/5687075/c9281c58-9816-11e4-9b4e-54f855721434.png)

elemoine:
Le 9 janv. 2015 21:48, "Greg Gianforcaro" <notifications@github.com> a
écrit :
>
> I am having a similar problem with ol.stlye.Icon.size in OL version 3.1.1.
>
> In my situation my src image is slightly larger, say 25 x 25px. If I set
a size on my ol.style.Icon of 20 x 20px then my image is truncated rather
than scaled (as graphicHeight/graphicWidth did in OL2).
>
> var iconStyle = new ol.style.Style
> ({
>   image: new ol.style.Icon(/** @type {olx.style.IconOptions} */ ({
>     size : [20,20],
>     src: 'data/icon.png' //25 x 25px image
>   }))
> });

Have you tried using the "scale" option? In your case, scale: 0.8.

greggian:
Yeah, I am aware of the scale option but it is desirable to set an absolute height/width on an Icon. In some cases, icons could come from a variety of sources and we may not know their original size so simply using scale isn't reliable.

[...]

PR #6100, opened 2016-11-10
Make ol.style.Icon scalable in both X and Y

tchandelle:
Implements #2720 and #3803. Add scaleX, scaleY, width and height to ol.style.Image.

Still a few problems: 

- [x] Scale and rotation doesn't work together
- [x] Rendering problem on Firefox


tchandelle:
That's weird, rendering test on canvas fail on Firefox, generated image does not scale correctly.
But changing scaleX and/or scaleY on an example (icon-color) works.

tschaub:
The change here makes it so we have:

 * size
 * width
 * height
 * scale
 * scaleX
 * scaleY
 * imgSize

We'll need very clear documentation on which does what and where one takes precedence over others.  If there is not a need for all (and I hope there isn't), we should avoid duplication.

@tchandelle - can you provide detail on how each works and why all are needed?

tchandelle:
- `size` and `imgSize` : I didn't change them. And I still don't know the difference. But it's about the size of the source image.
- `width` and `height` : Change the destination width and/or height of the image. If only one of them is declared, the other one is the source size
- `scaleX` and `scaleY` : Apply a scale factor horizontally and/or vertically to the destination image size, can be paired with `width` and/or `height`
- `scale` : It's not used as-is anymore. If defined, it will set both `scaleX` and `scaleY` to the same value. If scaleX and scaleY are different, getScale() will return 1, I'm not sure what else to return.

@tschaub, is this more clear, any suggestions ?

tchandelle:
After some thoughts, here is my new suggestion :
 - (not changed) `imgSize` : original size of the image, if it's manually loaded
 - (not changed) `size` : size of the image to be cropped
 - `destinationSize` : pair of number, to resize the image, 
    - If one of the two values is `null`, it will be calculated by the other, keeping the image ratio.  
An image of `size = [10, 20]` with `destinationSize = [20, null]` will be resized to `[20, 40]`
 - `scale` : either a pair of number (`[scaleX, scaleY]`), or a number

tchandelle:
Rebased, solved conflicts and changed options names as mentioned in my previous comment.

tchandelle:
Rebased with the recent changes made to the webgl renderer.

tchandelle:
What's the status for this PR, it has been opened for quite some time now ? Should I make any changes ?

ahocevar:
I think this looks good now. @tschaub, do you want to take another look?

tschaub:
At a glance, I'm confused.  Perhaps it is just a documentation issue.  Are `scale` and `destinationSize` supposed to be used together?  If these properties are redundant, let's remove one.

tchandelle:
Both serve the same purpose, but one represent a relative percentage, and the other an absolute size in pixels. They can be used together (as illustrated in my last example here-below), although I personally doesn't see the point.

Considering an image of size `[16, 32]` : 
 - `scale: 2` will result to an image of `[32, 64]`
 - `scale: [4, 2]`, an image of size `[64, 64]`
 - `destinationSize: [32, 64]`, an image of size `[32, 64]`
 - `destinationSize: [32, null]` or `[null, 64]`, an image of size `[32, 64]`, keeping the original size ratio
 - `destinationSize: [32, 64], scale: 2`, an image of size `[64, 128]`

omarkljung:
I've tried it out and it works well for my use case when I know the size I want it to have but not the original size so scale is not possible to provide without preloading the image separately.

It's a bit confusing with two different size properties though.
Wouldn't it be possible to just expand the meaning of `size` and `scale` using something similar to what `destinationSize` is here?
What you need is to provide a way to auto scale and not crop. So perhaps if `scale` could take `null` and treat it as unknown scale and therefore automatically scale the whole image (or just one axis?) it would be sufficient?
Giving a proper `size` and `scale` would still crop and scale the cropped image as is the case now.
A `size` and `scale: null` would auto scale.
Combining `size` and `scale` as in your last example would no longer be possible.

Another possibility could perhaps be to introduce a new property to control the cropping i.e `crop: true`.
That would then tell how the `size` property would be applied. `scale` can then be applied afterwards in both cases. For `crop: false` you'd then get auto scaling.

PR #6101, opened 2016-11-10
Avoid starting new zoom animation while already animating

### LINKED ISSUE #6096: Weird zoom animation
When zooming in or out while moving cursor, the animation looks weird, I don't know how to describe it :

zoom

It looks like it's caused by #6079, especially commit 3acfe79, @tschaub ?

tschaub:
This is a stop-gap improvement for mousewheel zooming.  It makes it so we don't start a new zoom animation if there is an existing one underway.

Fixes #6096.

### DIFF
--- a/src/ol/interaction/mousewheelzoom.js
+++ b/src/ol/interaction/mousewheelzoom.js
@@ -133,13 +133,14 @@ ol.interaction.MouseWheelZoom.handleEvent = function(mapBrowserEvent) {
  * @param {ol.Map} map Map.
  */
 ol.interaction.MouseWheelZoom.prototype.doZoom_ = function(map) {
-  var maxDelta = ol.MOUSEWHEELZOOM_MAXDELTA;
-  var delta = ol.math.clamp(this.delta_, -maxDelta, maxDelta);
-
   var view = map.getView();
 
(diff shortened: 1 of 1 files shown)


tchandelle:
The effect is gone, but unfortunately, it's slower. On simple map example, it takes me 6 seconds to go from highest zoom to lowest. With this patch, it takes me 12 seconds.

ahocevar:
This also fixes the problem I described in https://github.com/openlayers/ol3/pull/6079#issuecomment-259140978.

I agree with @tchandelle that wheel-zooming across several zoom levels is very slow now. Would it be possible to modify a running zoom animation so there are no stop gaps at all?

tschaub:
The right solution is to separate zooming with trackpad scrolling and mouse wheeling.  I've started on this, but it is not complete.

ahocevar:
I observed the same as @tchandelle, using a trackpad.

With this change here, the zooming experience is very similar to Leaflet. So definitely an improvement. The nicer experience in Google Maps and mapbox-gl-js does not have any stop gaps.

tschaub:
Agreed.

tschaub:
> ~~This also fixes the problem I described in #6079 (comment).~~ Edit: That issue persists.

Ticketed as #6102.

PR #6105, opened 2016-11-11
Move vector tile loader functions into vector tile module

tschaub:
This moves the feature loader function that is specific to vector tiles into `src/ol/vectortile.js`.  It also load success and failure handlers to the vector tile prototype.


gberaudo:
Very nice! I think it is the good approach.

PR #6106, opened 2016-11-11
Add type annotations for vector tiles

tschaub:
Since `getProjection` and `getReplayState` are not defined for `ol.Tile`, the compiler needs to know we are talking about `ol.VectorTile` here.

See https://travis-ci.org/camptocamp/ngeo/jobs/174469862#L614-L631 for a related build failure.  It would be nice to know why the compiler doesn't warn of the same in our own build jobs.


tschaub:
Merging as an annotation only change.

PR #6107, opened 2016-11-11
Reset hint first and notify of change when cancelling animations

tschaub:
With this change, when `view.cancelAnimations()` is called, we first set the animating hint to false (so `view.getAnimating()` returns `false`), then call any animation callbacks, and then fire the `change` event.


tschaub:
Thanks for the review.

One lurking issue here is that if animation callbacks start new animations even when they are called with `false`, things get in a bad state.  For now, I think we could just say this is user error - animation callbacks should not start new animations if they get called with `false` (indicating that something has cancelled animations).  We could guard against this, but I don't think this is critical to address.

PR #6113, opened 2016-11-12
Smooth trackpad zooming

### LINKED ISSUE #6030: Weird behavior on mouse wheel zooming with magic mouse
1. Put two fingers on magic mouse
2. Drag map -- dragging work is fine
3. One of the fingers up and start scroll ( one is on the mouse, the other one is scrolling the map)
4. Drag map again -- it works with changing resolutions
5. All fingers up and back to the step 1, 2 -- fine again.

I don't know why the wheel sensitivity became so high after step 3. 
When dragging the map, my fingers may move a little ahead and back on the mouse but it is the same when step 1, 2.

tschaub:
This reworks the mousewheel zoom interaction for nicer trackpad zooming.

After trying a number of different strategies, by far the nicest interaction is to allow the view to settle at fractional zoom levels.  We could add an option to always end on an integer zoom level, but this makes for jumpier interaction, and I would prefer that it not be the default.

Fixes #6030.

### DIFF
--- a/src/ol/interaction/mousewheelzoom.js
+++ b/src/ol/interaction/mousewheelzoom.js
@@ -66,6 +66,12 @@ ol.interaction.MouseWheelZoom = function(opt_options) {
    */
   this.timeoutId_ = undefined;
 
+  /**
+   * @private
+   * @type {ol.interaction.MouseWheelZoom.Mode|undefined}
+   */
+  this.mode_ = undefined;
+
 };
 ol.inherits(ol.interaction.MouseWheelZoom, ol.interaction.Interaction);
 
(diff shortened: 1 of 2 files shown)


ahocevar:
Great work @tschaub. I just tried this branch with a trackpad, and I love the behavior! Google Maps and Mapbox-GL-JS also do not restrict to integer zoom levels on trackpad/wheel zoom, so I'm convinced our users will love this as default too.

tschaub:
I've tested on Chrome, Firefox, and Safari with a trackpad and mouse with wheel.  I'm happy with the result, so this is ready for review now.

To make it easier for others to see the result, I've put up http://tschaub.net/ol3/scroll-zoom/examples/simple.html (temporarily).

tschaub:
Thanks for the review.  I think more folks will get a chance to test it in master.  So, I'll merge now and we can make any adjustments later.

cs09g:
Great work ever! Thank you @tschaub

PR #6124, opened 2016-11-14
Allow up to 16 decimals for alpha channel when parsing rgba strings to colors

marcjansen:
This changes the `ol.color.RGBA_COLOR_RE_` regular expression in a way that the following string is now also parseable:

`'rgba(255, 255, 0, 0.3333333333333333)'`

Effectively we allow now up to 16 decimal digits for the alpha channel.

To me this seems the correct solution, backed by this quote from the wikipedia article [IEEE Standard for Floating-Point Arithmetic (IEEE 754)](https://en.wikipedia.org/wiki/IEEE_floating_point#Character_representation):

> …
> When using a decimal floating point format the decimal representation will be preserved using:
> 
> * 7 decimal digits for decimal32
> * 16 decimal digits for decimal64
> * 34 decimal digits for decimal128
> …

See also https://github.com/openlayers/ol3/pull/5462#issuecomment-258830006

Please review.

/cc @GaborFarkas @gberaudo


GaborFarkas:
It works great for me. I like how it does not affect RGB values, only alpha.

gberaudo:
Thanks @marcjansen, looks great!

marcjansen:
I consider this a 'Please merge' :smiley:

gberaudo:
Will say it next time  :stuck_out_tongue_winking_eye:!

PR #6127, opened 2016-11-14
Format WKT Z, M and ZM

omarkljung:
Adds support for reading and writing WKT with the XYZ layout.
Also adds a respect to the optional layout parameter in the implementations of SimpleGeometry.setCoordinates when no coordinates are given.
This was needed to get empty geometries to be parsed and written (ie POINT Z EMPTY).

### DIFF
--- a/src/ol/format/wkt.js
+++ b/src/ol/format/wkt.js
@@ -6,12 +6,14 @@ goog.require('ol.format.Feature');
 goog.require('ol.format.TextFeature');
 goog.require('ol.geom.GeometryCollection');
 goog.require('ol.geom.GeometryType');
+goog.require('ol.geom.GeometryLayout');
 goog.require('ol.geom.LineString');
 goog.require('ol.geom.MultiLineString');
 goog.require('ol.geom.MultiPoint');
 goog.require('ol.geom.MultiPolygon');
 goog.require('ol.geom.Point');
 goog.require('ol.geom.Polygon');
+goog.require('ol.geom.SimpleGeometry');
 
 
 /**
@@ -50,6 +52,27 @@ ol.format.WKT.EMPTY = 'EMPTY';
 
 
 /**
+ * @const
+ * @type {string}
+ */
+ol.format.WKT.Z = 'Z';
+
+
+/**
+ * @const
+ * @type {string}
+ */
+ol.format.WKT.M = 'M';
+
+
+/**
+ * @const
+ * @type {string}
+ */
+ol.format.WKT.ZM = 'ZM';
+
+
+/**
  * @param {ol.geom.Point} geom Point geometry.
  * @return {string} Coordinates part of Point as WKT.
  * @private
@@ -59,7 +82,7 @@ ol.format.WKT.encodePointGeometry_ = function(geom) {
   if (coordinates.length === 0) {
     return '';
   }
-  return coordinates[0] + ' ' + coordinates[1];
+  return coordinates.join(' ');
 };
 
 
@@ -102,7 +125,7 @@ ol.format.WKT.encodeLineStringGeometry_ = function(geom) {
   var coordinates = geom.getCoordinates();
   var array = [];
   for (var i = 0, ii = coordinates.length; i < ii; ++i) {
-    array.push(coordinates[i][0] + ' ' + coordinates[i][1]);
+    array.push(coordinates[i].join(' '));
   }
   return array.join(',');
 };
@@ -155,6 +178,23 @@ ol.format.WKT.encodeMultiPolygonGeometry_ = function(geom) {
   return array.join(',');
 };
 
+/**
+ * @param {ol.geom.SimpleGeometry} geom SimpleGeometry geometry.
+ * @return {string} Potential dimensional information for WKT type.
+ * @private
+ */
+ol.format.WKT.encodeGeometryLayout_ = function(geom) {
+  var layout = geom.getLayout();
+  var dimInfo = '';
+  if (layout === ol.geom.GeometryLayout.XYZ || layout === ol.geom.GeometryLayout.XYZM) {
+    dimInfo += ol.format.WKT.Z;
+  }
+  if (layout === ol.geom.GeometryLayout.XYM || layout === ol.geom.GeometryLayout.XYZM) {
+    dimInfo += ol.format.WKT.M;
+  }
+  return dimInfo;
+};
+
 
 /**
  * Encode a geometry as WKT.
@@ -168,6 +208,12 @@ ol.format.WKT.encode_ = function(geom) {
   ol.DEBUG && console.assert(geometryEncoder, 'geometryEncoder should be defined');
   var enc = geometryEncoder(geom);
   type = type.toUpperCase();
+  if (geom instanceof ol.geom.SimpleGeometry) {
+    var dimInfo = ol.format.WKT.encodeGeometryLayout_(geom);
+    if (dimInfo.length > 0) {
+      type += ' ' + dimInfo;
+    }
+  }
   if (enc.length === 0) {
     return type + ' ' + ol.format.WKT.EMPTY;
   }
@@ -194,7 +240,7 @@ ol.format.WKT.GeometryEncoder_ = {
 /**
  * Parse a WKT string.
  * @param {string} wkt WKT string.
- * @return {ol.geom.Geometry|ol.geom.GeometryCollection|undefined}
+ * @return {ol.geom.Geometry|undefined}
  *     The geometry created.
  * @private
  */
@@ -535,10 +581,10 @@ ol.format.WKT.Parser = function(lexer) {
   this.token_;
 
   /**
-   * @type {number}
+   * @type {ol.geom.GeometryLayout}
    * @private
    */
-  this.dimension_ = 2;
+  this.layout_ = ol.geom.GeometryLayout.XY;
 };
 
 
@@ -550,6 +596,16 @@ ol.format.WKT.Parser.prototype.consume_ = function() {
   this.token_ = this.lexer_.nextToken();
 };
 
+/**
+ * Tests if the given type matches the type of the current token.
+ * @param {ol.format.WKT.TokenType} type Token type.
+ * @return {boolean} Whether the token matches the given type.
+ */
+ol.format.WKT.Parser.prototype.isTokenType = function(type) {
+  var isMatch = this.token_.type == type;
+  return isMatch;
+};
+
 
 /**
  * If the given type matches the current token, consume it.
@@ -557,7 +613,7 @@ ol.format.WKT.Parser.prototype.consume_ = function() {
  * @return {boolean} Whether the token matches the given type.
  */
 ol.format.WKT.Parser.prototype.match = function(type) {
-  var isMatch = this.token_.type == type;
+  var isMatch = this.isTokenType(type);
   if (isMatch) {
     this.consume_();
   }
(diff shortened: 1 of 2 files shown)


omarkljung:
Fixed eslint errors and squashed commits to one.

src/ol/format/wkt.js
```diff
@@ -168,16 +168,21 @@ ol.format.WKT.encode_ = function(geom) {
   ol.DEBUG && console.assert(geometryEncoder, 'geometryEncoder should be defined');
   var enc = geometryEncoder(geom);
   type = type.toUpperCase();
+  if (geom.getLayout) {
+    var layout = geom.getLayout();
+    if (layout === ol.geom.GeometryLayout.XYZ || layout === ol.geom.GeometryLayout.XYZM) {
+      type += ' Z';
```
tchandelle: You could also add `'M'` if layout is `XYM` or `XYZM`

src/ol/format/wkt.js
```diff
@@ -194,13 +199,15 @@ ol.format.WKT.GeometryEncoder_ = {
 /**
  * Parse a WKT string.
  * @param {string} wkt WKT string.
- * @return {ol.geom.Geometry|ol.geom.GeometryCollection|undefined}
+ * @return {ol.geom.SimpleGeometry|ol.geom.GeometryCollection|undefined}
  *     The geometry created.
  * @private
  */
 ol.format.WKT.prototype.parse_ = function(wkt) {
+  var dimensions = wkt.indexOf('Z') === -1 ? 2 : 3;
```
tchandelle: Also check if there is an `'M'`.

src/ol/format/wkt.js
```diff
@@ -538,7 +546,7 @@ ol.format.WKT.Parser = function(lexer) {
    * @type {number}
    * @private
    */
-  this.dimension_ = 2;
+  this.dimensions_ = dimensions || 2;
```
tchandelle: Wouldn't it be more simple to replace dimension by GeometryLayout ?

omarkljung: Yes, it crossed my mind but then more code then necessary needs to be changed.

src/ol/format/wkt.js
omarkljung: Yeah, but there is no code that handles it. So I thought it was more clear if there wasn't semi-support.

src/ol/format/wkt.js
omarkljung: Yes, but there is not support elsewhere for `'M'`.

omarkljung:
I'll look into what it would take to also support `'M'`

src/ol/format/wkt.js
tchandelle: It shouldn't be more complicated. `XYM` is like `XYZ`, but with an M. And `XYZM` is the same thing with 4 dimensions.

omarkljung:
Updated and rewritten to also support M and ZM.

marcjansen:
Code looks good first at first view, thanks @omarkljung.

Thanks for this contribution.

In order for this to be merged we need to have a [CLA](https://github.com/openlayers/ol3/blob/master/CONTRIBUTING.md#contributor-license-agreement) from you. Did you sign it?

Hey @bartvde, any chance you can have a look at this PR? You are the standards guy after all.

omarkljung:
OK. I've now signed the individual CLA.

bartvde:
I'll have a look at it today @marcjansen

src/ol/format/wkt.js
```diff
@@ -168,16 +169,28 @@ ol.format.WKT.encode_ = function(geom) {
...
     return type + ' ' + ol.format.WKT.EMPTY;
   }
   return type + '(' + enc + ')';
 };
 
-
 /**
  * @const
- * @type {Object.<string, function(ol.geom.Geometry): string>}
+ * @type {Object.<string, function((ol.geom.SimpleGeometry|ol.geom.GeometryCollection)): string>}
```
bartvde: why was this type of change needed? I took those changes out and the compiler doesn't complain?

omarkljung: It was needed at some point. Perhaps before the rewrite to use GeometryLayout.
I believe it's more correct since it describes the limitations better. But I can change back to Geometry now.

src/ol/geom/circle.js
```diff
@@ -186,7 +186,7 @@ ol.geom.Circle.prototype.setCenter = function(center) {
  */
 ol.geom.Circle.prototype.setCenterAndRadius = function(center, radius, opt_layout) {
   if (!center) {
-    this.setFlatCoordinates(ol.geom.GeometryLayout.XY, null);
+    this.setFlatCoordinates(opt_layout || ol.geom.GeometryLayout.XY, null);
```
omarkljung: For Circle it's not needed since CIRCLE isn't part of WKT. Just thought it should be consistent and respect the sent parameters, if present.

bartvde: But if there are no coordinates, layout should not matter right? I took them all out and all tests passed. setLayout is called when a true coordinates argument is provided or am I missing something?

src/ol/geom/point.js
```diff
@@ -104,7 +104,7 @@ ol.geom.Point.prototype.intersectsExtent = function(extent) {
  */
 ol.geom.Point.prototype.setCoordinates = function(coordinates, opt_layout) {
   if (!coordinates) {
-    this.setFlatCoordinates(ol.geom.GeometryLayout.XY, null);
+    this.setFlatCoordinates(opt_layout || ol.geom.GeometryLayout.XY, null);
```
omarkljung: I think it's more correct to respect the passed parameter, if present.
It is also needed to keep dimension information for empty (i.e. POINT ZM EMPTY).
Se failing tests if you remove it.

src/ol/geom/circle.js
omarkljung: Not for Circle no.

omarkljung:
Ok, so the respect of `opt_layout` in `ol.geom.SimpleGeometry.setCoordinates` is not needed in other places than in `ol.geom.Point`. Otherwise the tests for WKT EMPTY does not pass since the dimension information is lost.
Can I leave it in `ol.geom.Point.setCoordinates` or should I remove the three WKT EMPTY tests?

bartvde:
I'd be in favour of the latter

bartvde:
Do we have tests for the simple forms, i.e. ```POINT EMPTY```, ```MULTIPOLYGON EMPTY``` etc. ? I think that would be sufficient. Have you seen examples of EMPTY geometries with dimensions in the wild?

omarkljung:
Yes. I took that test and added M, Z and ZM versions to cover the additions.
No, I've never seen any WKT EMPTY example in the wild.
Let's say if it is EMPTY the you don't get to say in what dimensions it is empty.
I'll remove it and update the PR.

omarkljung:
Removed all unnecessary changes and relaxed the compiler directives.
Removed the EMPTY tests for M, Z and ZM.

src/ol/format/wkt.js
```diff
@@ -168,13 +169,25 @@ ol.format.WKT.encode_ = function(geom) {
...
+      type += ' ' + dimInfo;
+    }
+  }
   if (enc.length === 0) {
     return type + ' ' + ol.format.WKT.EMPTY;
   }
   return type + '(' + enc + ')';
 };
 
-
```
bartvde: I'm not sure if we still enforce this automatically, but there need to be 2 lines here

src/ol/geom/linestring.js
```diff
@@ -246,7 +246,7 @@ ol.geom.LineString.prototype.intersectsExtent = function(extent) {
  */
 ol.geom.LineString.prototype.setCoordinates = function(coordinates, opt_layout) {
   if (!coordinates) {
-    this.setFlatCoordinates(ol.geom.GeometryLayout.XY, null);
+    this.setFlatCoordinates(ol.geom.GeometryLayout.XY, []);
```
bartvde: is this an intentional change?

src/ol/geom/multipoint.js
```diff
@@ -173,7 +173,7 @@ ol.geom.MultiPoint.prototype.intersectsExtent = function(extent) {
  */
 ol.geom.MultiPoint.prototype.setCoordinates = function(coordinates, opt_layout) {
   if (!coordinates) {
-    this.setFlatCoordinates(ol.geom.GeometryLayout.XY, null);
+    this.setFlatCoordinates(ol.geom.GeometryLayout.XY, []);
```
bartvde: is this an intentional change?

bartvde:
thanks @omarkljung almost good to go I think, see my last comments / questions, thanks for your work

src/ol/geom/linestring.js
omarkljung: Yes. The EMPTY tests does not pass otherwise.
Could be handled in `getCoordinates` instead. That code does not handle that this.flatCoordinates is null and throws.

src/ol/geom/multipoint.js
omarkljung: Yes. The EMPTY tests does not pass otherwise.
Could be handled in `getCoordinates` instead. That code does not handle that this.flatCoordinates is null and throws.

src/ol/geom/linestring.js
omarkljung: I've already started to investigate. Checked with master and flatCoordinates is set to null but when getCoordinates is called it is an empty array. Looking into it.

bartvde:
I wonder if you can change it here instead:

```
var coordinates = isEmpty ? [] : parser.call(this);
```

so [] instead of null?

src/ol/format/wkt.js
```diff
@@ -577,26 +589,40 @@ ol.format.WKT.Parser.prototype.parse = function() {
...
+    var isEmpty = false;
+    if (this.match(ol.format.WKT.TokenType.TEXT)) {
+      var dimInfo = dimToken.value;
+      if (dimInfo === 'Z') {
+        this.layout_ = ol.geom.GeometryLayout.XYZ;
+      } else if (dimInfo === 'M') {
+        this.layout_ = ol.geom.GeometryLayout.XYM;
+      } else if (dimInfo === 'ZM') {
+        this.layout_ = ol.geom.GeometryLayout.XYZM;
+      } else if (dimInfo === 'EMPTY') {
```
bartvde: we should use ```ol.format.WKT.EMPTY``` here instead?

omarkljung:
Yes. For all others but Point it works.
It's handled in the parsers but if dimInfo is found the text is already consumed.
I could pass the isEmpty to the parsers to be handled correctly. Better?

src/ol/format/wkt.js
omarkljung: Should I add constanst for `M`, `Z` and `ZM` as well?

bartvde: Yes makes sense to me. Btw, can't we use isEmptyGeometry_ here as well? Or alternatively, can we leave the empty handling to the more specific functions? As you can see I'm also not so familiar with this part of the code base ;-)

bartvde: Ah I see your comment now (It's handled in the parsers but if dimInfo is found the text is already consumed.) which explains why this isn't possible

bartvde:
I think passing isEmpty to the parsers would be better indeed

omarkljung:
Extending each parser didn't seem right. So I created `isTokenType`. It's like `match` only it doesn't consume. For the geometry layout parsing we only wish to consume if there was any layout text present. If we hit other text (`ol.format.WKT.EMPTY`) we don't what to consume but instead move on and let each parser handle the case with emptiness.
I also extracted much of the added code into the functions `parseGeometryLayout_` and `encodeGeometryLayout_`.
I hope you like the improvements :)

bartvde:
I like the changes @omarkljung thank you very much, will merge

PR #6132, opened 2016-11-15
Add getter for color property to ol.style.Icon

### LINKED ISSUE #6131: Addition getter for Color property on ol.style.Icon
I'd like to add to [ol.style.Icon](http://openlayers.org/en/latest/apidoc/ol.style.Icon.html) getter for property "color". Please add getColor() method to ol.style.Icon similar like in [ol.style.Fill](http://openlayers.org/en/latest/apidoc/ol.style.Fill.html) and [Stroke](http://openlayers.org/en/latest/apidoc/ol.style.Stroke.html) ;)
Thanks Tom

tohu12:
Added getColor() method to class ol.style.Icon similar like in ol.style.Fill and ol.style.Stroke

Fixes #6131.

### DIFF
--- a/src/ol/style/icon.js
+++ b/src/ol/style/icon.js
@@ -260,6 +260,16 @@ ol.style.Icon.prototype.getAnchor = function() {
 
 
 /**
+ * Get the icon color.
+ * @return {ol.Color} Color.
+ * @api
+ */
+ol.style.Icon.prototype.getColor = function() {
+  return this.color_;
+};
+
+
+/**
  * Get the image icon.
  * @param {number} pixelRatio Pixel ratio.
  * @return {Image|HTMLCanvasElement} Image or Canvas element.
--- a/test/spec/ol/style/icon.test.js
+++ b/test/spec/ol/style/icon.test.js
@@ -75,7 +75,7 @@ describe('ol.style.Icon', function() {
       expect(original.anchorXUnits_).to.eql(clone.anchorXUnits_);
       expect(original.anchorYUnits_).to.eql(clone.anchorYUnits_);
       expect(original.crossOrigin_).to.eql(clone.crossOrigin_);
-      expect(original.color_).to.eql(clone.color_);
+      expect(original.getColor()).to.eql(clone.getColor());
       expect(original.getImage(1).src).to.eql(clone.getImage(1).src);
       expect(original.getImage(1).toDataURL()).to.eql(original.getImage(1).toDataURL());
       expect(original.offset_).to.eql(clone.offset_);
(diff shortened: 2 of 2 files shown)


src/ol/style/icon.js
```diff
@@ -260,6 +260,16 @@ ol.style.Icon.prototype.getAnchor = function() {
 
 
 /**
+ * Get the icon color.
+ * @return {ol.Color|string} Color.
```
tchandelle: `color_` is parsed as ol.Color. So it shouldn't be a `string`

src/ol/style/icon.js
```diff
@@ -260,6 +260,16 @@ ol.style.Icon.prototype.getAnchor = function() {
...
+ * @api
+ */
+ol.style.Icon.prototype.getColor = function() {
+  return this.color_;
```
tchandelle: and since `ol.Color` is an array, you should probably make a copy of it

tohu12: but is it same as in[ fill.js](/openlayers/ol3/blob/v3.19.1/src/ol/style/fill.js#L51) and [stroke.js](/openlayers/ol3/blob/v3.19.1/src/ol/style/stroke.js#L88) ?

tchandelle: Apparently, it's not initialised the same way in the constructor : 
[ol/style/fill.js#L19-L23](https://github.com/openlayers/ol3/blob/48c1d09a95a11e974d05582e5d4c15c5492c0254/src/ol/style/fill.js#L19-L23)
[ol/style/icon.js#L98-L103](https://github.com/openlayers/ol3/blob/48c1d09a95a11e974d05582e5d4c15c5492c0254/src/ol/style/icon.js#L98-L103)

tohu12: OK, thanks

src/ol/style/icon.js
```diff
@@ -260,6 +260,16 @@ ol.style.Icon.prototype.getAnchor = function() {
 
 
 /**
+ * Get the icon color.
+ * @return {ol.Color} Color.
+ * @api
+ */
+ol.style.Icon.prototype.getColor = function() {
+  return this.color_.slice();
```
tchandelle: `this.color_` could be null

src/ol/style/icon.js
```diff
@@ -260,6 +260,16 @@ ol.style.Icon.prototype.getAnchor = function() {
 
 
 /**
+ * Get the icon color.
+ * @return {ol.Color} Color.
+ * @api
+ */
+ol.style.Icon.prototype.getColor = function() {
+  return (this.color_ && this.color_.slice) ? this.color_.slice() : this.color_;
```
ahocevar: Usually our simple getters just return what was provided. So in this case, you would just always return `this.color_`.

ahocevar:
@tohu12 Can you please sign a [CLA](https://github.com/openlayers/cla) so we can merge this? Thanks.

tohu12:
I made "squash" over SourceTree my desktop aplication, I never do this before. I hope that now is right?

ahocevar:
Sorry, that did not look right. I took the liberty to fix it for you.

tohu12:
Thanks, and now ?

tohu12:
@ahocevar I submited a digital Individual CLA

PR #6133, opened 2016-11-15
Doc clarification ol.MapBrowserEvent#pixel, ol.MapBrowserEvent#coordinate, ol.Map#getEventCoordinate

simonseyock:
Clarified the description of ol.MapBrowserEvent#pixel, ol.MapBrowserEvent#coordinate and ol.Map#getEventCoordinate.
The current description is somewhat misleading:
`pixel {ol.Pixel} The pixel of the original browser event.` could be understood as the pixel attribute having the value [e.clientX, e.clientY] of the original browser event.
Changed it to `The map pixel relative to the viewport corresponding to the original browser event.`

The corresponding change was made for the coordinate property and the description of getEventCoordinate was updated to the same text.

### DIFF
--- a/src/ol/mapbrowserevent.js
+++ b/src/ol/mapbrowserevent.js
@@ -40,14 +40,14 @@ ol.MapBrowserEvent = function(type, map, browserEvent, opt_dragging,
   this.originalEvent = browserEvent;
 
   /**
-   * The pixel of the original browser event.
+   * The map pixel relative to the viewport corresponding to the original browser event.
    * @type {ol.Pixel}
    * @api stable
    */
   this.pixel = map.getEventPixel(browserEvent);
 
   /**
-   * The coordinate of the original browser event.
+   * The geographical coordinate corresponding the original browser event.
    * @type {ol.Coordinate}
    * @api stable
    */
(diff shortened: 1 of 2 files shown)


src/ol/mapbrowserevent.js
```diff
@@ -40,14 +40,14 @@ ol.MapBrowserEvent = function(type, map, browserEvent, opt_dragging,
...
-   * The pixel of the original browser event.
+   * The map pixel relative to the viewport corresponding to the original browser event.
    * @type {ol.Pixel}
    * @api stable
    */
   this.pixel = map.getEventPixel(browserEvent);
 
   /**
-   * The coordinate of the original browser event.
+   * The geographical coordinate corresponding the original browser event.
```
ahocevar: corresponding *to*? Also, it's not a "geographical coordinate", but the "coordinate in view projection".

simonseyock: I borrowed the text from here: http://openlayers.org/en/master/apidoc/ol.Map.html#getEventCoordinate

simonseyock: I will change it there, too

PR #6139, opened 2016-11-16
Match equivalent projections from WMTS caps

bartvde:
Currently, if there is a matrixSet with EPSG:900913, it will not get matched when asking for projection EPSG:3857, we should match equivalent projections when possible and fall-back to the current behaviour if the projections are not known.

### DIFF
--- a/src/ol/source/wmts.js
+++ b/src/ol/source/wmts.js
@@ -317,9 +317,14 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
             var tileMatrixSet = ol.array.find(tileMatrixSets, function(el) {
               return el['Identifier'] == elt['TileMatrixSet'];
             });
-            return tileMatrixSet['SupportedCRS'].replace(
-                /urn:ogc:def:crs:(\w+):(.*:)?(\w+)$/, '$1:$3'
-                   ) == config['projection'];
+            var supportedCRS = tileMatrixSet['SupportedCRS'].replace(/urn:ogc:def:crs:(\w+):(.*:)?(\w+)$/, '$1:$3');
+            var proj1 = ol.proj.get(supportedCRS);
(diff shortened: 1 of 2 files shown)


src/ol/source/wmts.js
```diff
@@ -317,9 +317,14 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
             var tileMatrixSet = ol.array.find(tileMatrixSets, function(el) {
               return el['Identifier'] == elt['TileMatrixSet'];
             });
-            return tileMatrixSet['SupportedCRS'].replace(
-                /urn:ogc:def:crs:(\w+):(.*:)?(\w+)$/, '$1:$3'
-                   ) == config['projection'];
+            var supportedCRS = tileMatrixSet['SupportedCRS'].replace(/urn:ogc:def:crs:(\w+):(.*:)?(\w+)$/, '$1:$3');
```
tchandelle: Not directly related to this PR, but could not we do this regex replace in `ol.proj.get` instead? I see it's done at three different positions in the code.

Should I make an issue and another PR?

bartvde: Might make sense, @ahocevar do you remember if there was a specific reason we are not doing this in ```ol.proj.get```?

bartvde: We currently use a list of equivalent projections instead of the regex see e.g. https://github.com/openlayers/ol3/blob/master/src/ol/proj/epsg3857.js#L75:L83 but I cannot remember the rationale for this

ahocevar: We decided not to do any guessing or search/replace to match given CRS code notations to available CRS codes. Applications can easily provide aliases for different notations.

Looks like some convenience translation has sneaked into the WMTS format regardless.

bartvde:
thanks @ahocevar

Issue #6049, opened 2016-10-30
Point geometry features don't consider ol.Map.pixelRatio


Gussy:
#### Issue

When drawing `ol.geom.Point` features on a map with a `ol.style.Circle` style, the rendered `ol.style.Image` scale is hardcoded to `1`, which gives the features a blurry look on HiDPI displays. 

See https://github.com/openlayers/ol3/blob/master/src/ol/style/circle.js#L110

Patching the `scale` property on the line above to equal `1/ol.has.DEVICE_PIXEL_RATIO` and multiplying the circe `radius` and stroke `width` properties by `ol.has.DEVICE_PIXEL_RATIO` (doubling the feature size and halving the image scale) fixes the issue but probably isn't the correct way of implementing a fix. _(If it is, I will PR!)_
#### To reproduce

On a HiDPI display, use http://openlayers.org/en/latest/examples/draw-features.html to draw some `Point` features and some `Circle` features of roughly the same size. Notice the `Point` features are blurry but the `Circle` features are crisp.

Issue #2192, opened 2014-06-12
Image style ignores pixelRatio in immediate API


ahocevar:
When rendering points on a FeatureOverlay, they appear smaller than on a Vector layer on HiDPI devices. This is because the pixel ratio is [hard-coded to 1](https://github.com/openlayers/ol3/blob/master/src/ol/render/canvas/canvasimmediate.js#L887-L888).

PR #6144, opened 2016-11-18
ol.style.Circle high DPI support

### LINKED ISSUE #1574: ol.style.Circle high DPI support
<bountysource-plugin>

---

Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/1449543-ol-style-circle-high-dpi-support?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Gussy:
Fixes #1574, relates to #6049 and #2192

Adds support to `ol.style.Circle` for high DPI rendering. Most of this PR is plumbing to get the `pixelRatio` into the `ol.style.Circle#getImage()` function calls, then the pixel ratio is applied to the circle radius and stroke.

Tests have been updated/added to check for correct functionality. I'm open to suggestions for more tests if they would be useful.

### DIFF
--- a/src/ol/render/canvas/imagereplay.js
+++ b/src/ol/render/canvas/imagereplay.js
@@ -242,17 +242,18 @@ ol.render.canvas.ImageReplay.prototype.finish = function() {
 /**
  * @inheritDoc
  */
-ol.render.canvas.ImageReplay.prototype.setImageStyle = function(imageStyle) {
+ol.render.canvas.ImageReplay.prototype.setImageStyle = function(imageStyle, pixelRatio) {
   ol.DEBUG && console.assert(imageStyle, 'imageStyle should not be null');
+  // Call getImage to set the correct pixelRatio, before calling getAnchor, getSize etc.
+  var image = imageStyle.getImage(pixelRatio);
+  ol.DEBUG && console.assert(image, 'image should not be null');
   var anchor = imageStyle.getAnchor();
   ol.DEBUG && console.assert(anchor, 'anchor should not be null');
   var size = imageStyle.getSize();
   ol.DEBUG && console.assert(size, 'size should not be null');
-  var hitDetectionImage = imageStyle.getHitDetectionImage(1);
+  var hitDetectionImage = imageStyle.getHitDetectionImage(pixelRatio);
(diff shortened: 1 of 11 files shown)


Gussy:
ICLA signed.

src/ol/style/circle.js
```diff
@@ -107,7 +113,7 @@ ol.style.Circle = function(opt_options) {
     opacity: 1,
     rotateWithView: false,
     rotation: 0,
-    scale: 1,
+    scale: (1 / this.pixelRatio_),
```
tchandelle: `this.pixelRatio_` will always be `1`

tchandelle:
Can you please add rendering tests. 
I'm trying to see if there is any conflicts with #6100.

src/ol/style/circle.js
ahocevar: It won't. It may be changed in `#getImage()`.

tchandelle: I don't see how it could be changed between line 102 and 112. `render_()` is called in between and it does not seems to change it neither.

ahocevar:
Not your fault @Gussy, but I think a better, but also more invovled fix for this would be to move all rendering out of `ol.style.Image` and leave the decision on how to render to the respective renderer. See #5775.

src/ol/style/circle.js
ahocevar: The `#render()` call above should not be necessary, and I also see now that at this call site `this.pixelRatio_` will always be `1`.

tchandelle:
If the work is done for `ol.style.Circle`, could it not be done also for `ol.style.RegularShape` ? In fact we could even get rid of `Circle`, I'll send a PR about that.

Gussy:
No worries @ahocevar! I'm interested in helping get this High DPI support in whichever way is best, even if this isn't the ideal way.

Gussy:
Rebased on master due to changes in #6146

daspecster:
👍

PR #6146, opened 2016-11-18
Merge ol.style.Circle and RegularShape together

tchandelle:
As I suggested in https://github.com/openlayers/ol3/pull/6144#issuecomment-261542842, 95% is common. Both classes could be merged.

### DIFF
--- a/src/ol/style/circle.js
+++ b/src/ol/style/circle.js
@@ -1,13 +1,7 @@
 goog.provide('ol.style.Circle');
 
 goog.require('ol');
-goog.require('ol.color');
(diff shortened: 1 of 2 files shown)


tschaub:
I like all the red.

I still think it makes sense for the renderers to render circles.  It also makes sense to move the responsibility for regular shape rendering.  In addition, as @tchandelle mentions in https://github.com/openlayers/ol3/pull/6144#issuecomment-261542842, both circle and regular shape should get the same high DPI treatment.  So I think this change makes sense before either of these other efforts proceed.

PR #6151, opened 2016-11-21
Accept more valid rgb/rgba color strings when parsing colors

marcjansen:
This changes the regular expressions used in `ol.color.fromStringInternal_` to now parse less strict and accept e.g. the following strings:

* `rgb( \t 0  ,   0 \n , 255  )` (more whitespace characters alllowed, at more places)
* `rgba( \t 0  ,   0 \n , 255 , 0.4567   )` (more whitespace characters alllowed, at more places)
* `rgba(1, 1, 1, 0.0)` (decimal digits for 0)
* `rgba(1, 1, 1, 0.000000)` (up to 16 decimal digits for 0)
* `rgba(1, 1, 1, 1.0)` (decimal digits for 1)
* `rgba(1, 1, 1, 1.000000)` (up to 16 decimal digits for 1)

Please review.


tchandelle:
Why are decimal limited to 16 digits?

marcjansen:
>Why are decimal limited to 16 digits?

That was already the case before this PR. Here is a bit background: #6124

Does that answer the question?

tchandelle:
I was just thinking there is no reason to limit at all. If the user want to put 17 digits or more in his alpha value, there is no reason to stop him, `a = Number(match[4]);` won't fail.

marcjansen:
I wouldn't be opposed to

```diff
diff --git a/src/ol/color.js b/src/ol/color.js
index 9d2928f..ea4c02a 100644
--- a/src/ol/color.js
+++ b/src/ol/color.js
@@ -30,7 +30,7 @@ ol.color.RGB_COLOR_RE_ =
  * @private
  */
 ol.color.RGBA_COLOR_RE_ =
-    /^(?:rgba)?\(\s*(0|[1-9]\d{0,2})\s*,\s*(0|[1-9]\d{0,2})\s*,\s*(0|[1-9]\d{0,2})\s*,\s*(0|1|0\.\d{1,16}|1\.0{1,16})\s*\)$/i;
+    /^(?:rgba)?\(\s*(0|[1-9]\d{0,2})\s*,\s*(0|[1-9]\d{0,2})\s*,\s*(0|[1-9]\d{0,2})\s*,\s*(0|1|0\.\d+|1\.0+)\s*\)$/i;
```

(replaces `\d{1,16}` with `\d+` at two places)

Is that what you suggest? All tests herein also pass with that modification.

tchandelle:
Yes thanks, that's exactly what I meant.

tschaub:
```js
s.slice(5, -1).split(',').map(Number);
```

marcjansen:
@tschaub 

> ```js
> s.slice(5, -1).split(',').map(Number);
> ```

I took your comment as a hint that the parsing of colors is too complicated as is, right?

I have created #6164. In case everybody likes that better, I'd close this PR.

tchandelle:
I prefer with the full regex. 
With the current solution, if the color start with `rgb(` or `rgba(` followed by incorrect characters, it would emit error 14 :

> Invalid color. Valid colors are all CSS colors.
(which is, by the way, not true. Not all CSS colors are validated, like HSL, or RGB with pourcentage)

With #6164, it would be parsed as #000000

schmidtk:
The full regex is great to reduce false positives, but it should allow sufficient number of decimal places. We encountered this issue parsing colors from KML due to this line in `ol.format.KML.readColor_`:

```javascript
parseInt(hexColor.substr(0, 2), 16) / 255
```

The parsed alpha had varying precision depending on the browser, I believe up to 18 decimal places in our testing. This caused problems when translating the color to an RGBA string because it would not be matched by the regular expression.

In short, an arbitrary number of decimals would be preferable but 18 was the most we encountered.

marcjansen:
@schmidtk 

> The full regex is great to reduce false positives, but it should allow sufficient number of decimal places.

This should be the case with the changes in this PR

marcjansen:
@openlayers/ol3 : any opinions on this PR or #6164? TIA for any review.

marcjansen:
Closing in favor of #6164, as suggested by @ahocevar.

PR #6155, opened 2016-11-22
Set GeometryLayout correctly when reading GPX

### LINKED ISSUE #6148: Set GeometryLayout correctly when reading GPX
The reader of `ol.format.GPX` currently always creates geometry with layout `ol.geom.GeometryLayout.XYZM`.

I am not sure if Z/M are ever used in the GPX parser, but I think when reading a GPX file that only has lat/lon trackpoints, the layout of the imported geometry should be `ol.geom.GeometryLayout.XY`.

romanzoller:
Fixes #6148

### DIFF
--- a/src/ol/format/gpx.js
+++ b/src/ol/format/gpx.js
@@ -66,12 +66,13 @@ ol.format.GPX.SCHEMA_LOCATION_ = 'http://www.topografix.com/GPX/1/1 ' +
 
 /**
  * @param {Array.<number>} flatCoordinates Flat coordinates.
+ * @param {ol.LayoutOptions} layoutOptions Layout options.
  * @param {Node} node Node.
  * @param {Object} values Values.
  * @private
  * @return {Array.<number>} Flat coordinates.
  */
-ol.format.GPX.appendCoordinate_ = function(flatCoordinates, node, values) {
+ol.format.GPX.appendCoordinate_ = function(flatCoordinates, layoutOptions, node, values) {
   ol.DEBUG && console.assert(node.nodeType == Node.ELEMENT_NODE,
       'node.nodeType should be ELEMENT');
   flatCoordinates.push(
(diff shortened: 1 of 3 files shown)


src/ol/format/gpx.js
```diff
@@ -80,12 +81,14 @@ ol.format.GPX.appendCoordinate_ = function(flatCoordinates, node, values) {
   if ('ele' in values) {
     flatCoordinates.push(/** @type {number} */ (values['ele']));
     delete values['ele'];
+    layoutOptions['hasZ'] = true;
```
romanzoller: Good point, fixed. Thanks for the quick review.

bartvde:
2 more questions / remarks:

* I guess you looked into generating the correct flatCoordinates instead of modifying them afterwards?
* a test case for XYZM would be good

TIA

src/ol/format/gpx.js
```diff
@@ -94,6 +97,51 @@ ol.format.GPX.appendCoordinate_ = function(flatCoordinates, node, values) {
...
+  } else if (layoutOptions.hasZ) {
+    layout = ol.geom.GeometryLayout.XYZ;
+    stride = 3;
+  } else if (layoutOptions.hasM) {
+    layout = ol.geom.GeometryLayout.XYM;
+    stride = 3;
+  }
+  if (stride !== 4) {
+    var i;
+    for (i = 0; i < flatCoordinates.length / 4; i++) {
```
tchandelle: Or `for (i = 0, ii = flatCoordinates.length; i < ii; i+= stride)` and removing `* stride` below ? Isn't it better, @bartvde ?

romanzoller:
- Yes, basically the problem is that in the beginning, the final geometry layout / stride is not known, so another approach would be to traverse the xml tree twice, once to determine the layout, and once to extract coordinates. But I think the approach I chose is better in terms of code complexity, and comparable in terms of performance (although of course the only way to know that for sure is to implement both and measure).
- Yes, good point, I didn't realize that there was no XYZM case left after my modifications.

src/ol/format/gpx.js
romanzoller: fixed

bartvde:
Ah cool, I had missed those test cases since I only looked at the changeset.

This all looks good to me, any objections @openlayers/core to merging this?

bartvde:
thank you @romanzoller

romanzoller:
Thanks for the review!

PR #6173, opened 2016-11-26
Topolis example

bjornharrtell:
I've made a library for topology management, which is a dear old subject of mine. The library is called topolis and implements the same API as in PostGIS aka. ISO/IEC 13249-3 (SQL/MM). It's an early version and some work remains but the basic operations are in place.

This PR demonstrates the use of topolis integrated with OpenLayers to allow drawing and removal of edges, splitting existing edges and faces as required. Automatically created nodes and faces are visualized with sequential ids.

Screencast of the example in action:
topolis

The example is published here:
http://bjornharrtell.github.io/ol3/build/hosted/topolis-example/examples/topolis.html

And finally the GitHub project page for topolis:
https://github.com/bjornharrtell/topolis

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -7,6 +7,8 @@
     "d3": false,
     "jsPDF": false,
     "jsts": false,
+    "topolis": false,
+    "toastr": false,
     "saveAs": false,
     "topojson": false,
     "turf": false
--- a/examples/topolis.html
+++ b/examples/topolis.html
@@ -1,0 +1,15 @@
+---
+layout: example.html
+title: topolis integration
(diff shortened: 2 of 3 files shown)


tschaub:
This is awesome @bjornharrtell.  I hope to find an excuse to dig into topolis at some point soon.

Minor issue, but I notice that if you end up with an error (a self-intersecting curve), you are left with orphaned nodes.

bjornharrtell:
Thanks @tschaub. As you noticed there are some issues, both in the example and in topolis, so I also hope I will find an opportunity to put more time into this.

PR #6175, opened 2016-11-28
Generate projection transform function only when requested

### LINKED ISSUE #6174: Performance issue with Projections
There is a performance issue when using a lot of projections. 

For each new projections, transform functions are created from the new projection to every other projection and vice versa. `ol.proj.transforms_` contains *n²* functions, where *n* is the number of initialized projections. But we rarely need to convert coordinates from *n* projections to *n* projections. Most common use seems to be from *1* to *n*, and *n* to *1*. 

These transform functions should be generated only when requested.

tchandelle:
Fix #6174

### DIFF
--- a/src/ol/proj/index.js
+++ b/src/ol/proj/index.js
@@ -152,20 +152,6 @@ ol.proj.Projection = function(options) {
         if (options.units === undefined) {
           this.units_ = def.units;
         }
-        var currentCode, currentDef, currentProj, proj4Transform;
-        for (currentCode in projections) {
-          currentDef = proj4js.defs(currentCode);
-          if (currentDef !== undefined) {
-            currentProj = ol.proj.get(currentCode);
-            if (currentDef === def) {
-              ol.proj.addEquivalentProjections([currentProj, this]);
-            } else {
-              proj4Transform = proj4js(currentCode, code);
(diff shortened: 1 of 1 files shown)


src/ol/proj/index.js
```diff
@@ -739,6 +725,27 @@ ol.proj.getTransformFromProjections = function(sourceProjection, destinationProj
...
+        } else {
+          var proj4Transform = proj4js(destinationCode, sourceCode);
+          ol.proj.addCoordinateTransforms(destinationProjection, sourceProjection,
+              proj4Transform.forward, proj4Transform.inverse);
+        }
+        if (sourceCode in transforms && destinationCode in transforms[sourceCode]) {
+          transform = transforms[sourceCode][destinationCode];
+        }
+      } else {
+        transform = undefined;
```
tchandelle: If I don't do that here, the builder complains, but I can't figure out why. I'd be interested if someone has an logical explanation.

ahocevar:
@tchandelle I took the liberty to simplify the code a bit, which also made the compiler happy. I don't have a logical explanation though, other than that I don't like the compiler, but have learned how to make it behave by trial and error.

tchandelle:
Thanks @ahocevar, it looks better that way.

PR #6176, opened 2016-11-28
Doc: easing option is optional

tchandelle:
Easing options was marked as *Required*.

jsdoc was probably interpreting `function(number):number|undefined` as a function that returns a number or undefined.


tchandelle:
@ahocevar : ok, i've pushed these changes instead

fredj:
@ahocevar please merge if you're agree with the changes

ahocevar:
Thanks @tchandelle.

PR #6177, opened 2016-11-28
Add duration and easing options to view.fit() for animations

### LINKED ISSUE #6170: No non-deprecated alternative for view.fit() with animation
Since `map.beforeRender()` will be deprecated in 3.20, it's not possible to correctly animate `view.fit()` without using `beforeRender`, like in this example : https://jsfiddle.net/fnsyLj34/

tchandelle:
Fix #6170


tchandelle:
See also #6186 for an alternative fix.

tschaub:
This looks great @tchandelle.

PR #6181, opened 2016-11-29
Replace expired Bing key with a new one

ahocevar:
It happened again.


bartvde:
Oh no not again ;-)

Thanks for tackling

PR #6182, opened 2016-11-29
Unified canvas rendering with direct composition

ahocevar:
This is essentially the same as #6082, but with direct tile composition instead of rendering to an intermediate canvas.

Other than in current master, there are no hacks applied to make gaps between tiles go away on rotated views, but there are no gaps in all other situations.


ahocevar:
@tschaub I did 7b101b0 while lying in bed sick today, and I do not have enough energy left to do the rotation with a css transform instead of a canvas transform in the map renderer. It would be great if you could give that a try.

ahocevar:
@tschaub I made some progress with css rotation, but there are still a few issues with gradient/pattern fill rendering and postcompose transforms. Please let me know if you are able to help with any of those.

ahocevar:
Closing in favor of #6082 - it just gets unnecessarily complicated when we try to address rendering issues (gaps between tiles) with direct composition.

Issue #5297, opened 2016-04-29
Ignore duplicated attributions + fire change event once the UI element has been updated


aberenyi:
Hi,

Currently duplicated attributions cannot be filtered out easily.

Additionally the attribution control's `change` event is not fired once the control's UI element has been update.

Cheers

tschaub:
Your change in eb4f4327055b7fec6bd2ce94d7f517cdae93bf97 looks sensible to me @aberenyi (though I didn't review in detail).

tschaub:
(If the linter rules don't catch the inconsistent brace style in eb4f4327055b7fec6bd2ce94d7f517cdae93bf97, then we need to update our config.)

fredj:
> (If the linter rules don't catch the inconsistent brace style in eb4f432, then we need to update our config.)

The brace style is correctly catch by eslint:
```
/home/fredj/code/ol3/src/ol/control/attribution.js
  190:9   error  Opening curly brace does not appear on the same line as controlling statement  brace-style
  192:11  error  Opening curly brace does not appear on the same line as controlling statement  brace-style
  194:13  error  Opening curly brace does not appear on the same line as controlling statement  brace-style
  200:9   error  Opening curly brace does not appear on the same line as controlling statement  brace-style
```

ahocevar:
@aberenyi Would you be able to fix the brace style and create a pull request? Thanks.

aberenyi:
@tschaub @fredj @ahocevar thanks guys for the quick response, I'll submit a PR later today

aberenyi:
Gents, I've got 2 failing tests, however, I'm quite sure that they are not related to this.

```
  1) ol.format.GeoJSON #writeGeometry accepts featureProjection:

      expected [ 10, 19.999999999999986 ] to sort of equal [ 10, 20 ]
      + expected - actual

       [
         10
      -  19.999999999999986
      +  20
       ]
```

```
  2) ol.format.GeoJSON #writeGeometry respects featureProjection passed to constructor:

      expected [ 10, 19.999999999999986 ] to sort of equal [ 10, 20 ]
      + expected - actual

       [
         10
      -  19.999999999999986
      +  20
       ]
```

TL;DR; Full build log [here](https://gist.github.com/d44926018ff438563ef3b23782dce7bf)

bartvde:
Same for me see: https://github.com/openlayers/ol3/pull/5917#issuecomment-253225880

tschaub:
[...]

PR #6185, opened 2016-11-30
Ignore duplicated attributions

aberenyi:
Ignore duplicated attributions - see [#5297](https://github.com/openlayers/ol3/issues/5297).

### DIFF
--- a/src/ol/control/attribution.js
+++ b/src/ol/control/attribution.js
@@ -182,11 +182,21 @@ ol.control.Attribution.prototype.getSourceAttributions = function(frameState) {
       } else {
         intersectsTileRange = false;
       }
+      var attributionAlreadyAdded = false;
       if (intersectsTileRange) {
         if (sourceAttributionKey in hiddenAttributions) {
           delete hiddenAttributions[sourceAttributionKey];
         }
-        attributions[sourceAttributionKey] = sourceAttribution;
+        for (var existingKey in attributions) {
+          if (attributions.hasOwnProperty(existingKey)) {
+            if (attributions[existingKey].getHTML() === sourceAttribution.getHTML()) {
+              attributionAlreadyAdded = true;
+            }
+          }
+        }
+        if (!attributionAlreadyAdded) {
+          attributions[sourceAttributionKey] = sourceAttribution;
+        }
       } else {
         hiddenAttributions[sourceAttributionKey] = sourceAttribution;
       }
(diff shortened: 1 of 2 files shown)


probins:
pls squash all these commits

aberenyi:
done

tschaub:
I'm pulling in changes for the upcoming 3.20 release.  I don't feel comfortable bringing in changes without tests that clearly demonstrate the bug and fix.  There are zero attribution control tests currently, so I recognize that this is a fair bit of work for what might feel like a minor change.  But I'd like to see a test that fails in `master` and passes with this change before merging.  Perhaps someone else could lend a hand - especially if this is a real problem that people are experiencing.

ahocevar:
Working on a test right now.

aberenyi:
Many thanks

ahocevar:
Test added.

src/ol/control/attribution.js
```diff
@@ -182,11 +182,21 @@ ol.control.Attribution.prototype.getSourceAttributions = function(frameState) {
...
         intersectsTileRange = false;
       }
+      var attributionAlreadyAdded = false;
       if (intersectsTileRange) {
         if (sourceAttributionKey in hiddenAttributions) {
           delete hiddenAttributions[sourceAttributionKey];
         }
-        attributions[sourceAttributionKey] = sourceAttribution;
+        for (var existingKey in attributions) {
+          if (attributions.hasOwnProperty(existingKey)) {
```
tschaub: We create the `attributions` object and don't guard against people adding enumerable properties to `Object.prototype` elsewhere, so this `hasOwnProperty()` call is not necessary.

tschaub:
@aberenyi or @ahocevar - can you rebase to remove the 6 extra merge commits?

tschaub:
I rebased these commits and added one more in #6230.  @aberenyi and @ahocevar, let me know if you agree with those changes.

tschaub:
Thanks for the contribution @aberenyi and for the tests @ahocevar.  I've merged #6230.

probins:
any reason why this PR only applies to tile sources? If I add 10 vector sources with the same attribution, it's displayed 10 times.

ahocevar:
The whole attribution handling is unnecessarily complicated. We wouldn't need tile handling at all. All we'd need would be an `attribution` config option on `ol.source.Source`, which accepts a `string` or a `function(extent, resolution)`.

probins:
> The whole attribution handling is unnecessarily complicated

I'd certainly agree with that. :-) A related issue is that if you have several visible sources with attributions, there's nothing to tell you which attribution belongs to which source - you have to specify that in the attribution text. If you have for example an OSM layer with vectors, as in the vector-layer example, the OSM attribution implies that the copyright also applies to the vectors.

PR #6198, opened 2016-12-02
Fix sourceResolution value in view.animate

### LINKED ISSUE #6197: flyTo with resolution param is broken
Here a [jsfiddle](https://jsfiddle.net/w8gpLsb0/2/).

Click on `fly to Bern`button . The pan re-initialize the view 's resolution when the 2nd animation on resolution begins.

oterral:
Fix #6197


tschaub:
Thanks for the fix @oterral.

PR #6213, opened 2016-12-07
Add preload option to olx.layer.VectorTileOptions

drnextgis:
`preload` option can be passed to `ol.layer.VectorTile` but is not mentioned in docs


fredj:
coveralls check is stuck, anyone knows how to restart it?

ahocevar:
I think you can merge this now @fredj.

PR #6218, opened 2016-12-07
Make modify interaction more robust

gberaudo:
When the map has not been rendered yet, adding a feature should not
trigger coordinate related functions as they make use of the framestate.


gberaudo:
Thank you for the review @fredj.

PR #6220, opened 2016-12-07
Fix ol.Collection#push return value

### LINKED ISSUE #6219: Collection.push() returns wrong length
Collection.push() should return the new length of the collection after pushing the element but it returns the old length.

Current code:

```
ol.Collection.prototype.push = function(elem) {
  var n = this.array_.length;
  this.insertAt(n, elem);
  return n;
};
```

Possible fix:

```
ol.Collection.prototype.push = function(elem) {
  var n = this.array_.length;
  this.insertAt(n, elem);
  return this.array_.length;
};
```

fredj:
fixes #6219

### DIFF
--- a/src/ol/collection.js
+++ b/src/ol/collection.js
@@ -145,13 +145,13 @@ ol.Collection.prototype.pop = function() {
 /**
  * Insert the provided element at the end of the collection.
  * @param {T} elem Element.
- * @return {number} Length.
+ * @return {number} New length of the collection.
  * @api stable
  */
 ol.Collection.prototype.push = function(elem) {
-  var n = this.array_.length;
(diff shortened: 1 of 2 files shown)


tchandelle:
As I mentioned in my comment https://github.com/openlayers/ol3/issues/6219#issuecomment-265415127, would this be a case we should consider ?

````js
var collection = new ol.Collection();

collection.on('add', function(event) {
  if (event.element === 'foo') {
    collection.remove(event.element);
  }
});

collection.push('foo'); // length is still 0
collection.push('bar'); // length is 1
````

fredj:
@tchandelle code updated. I've added a test for this case

Issue #6223, opened 2016-12-08
PinchZoom: allow the view to stay at fractional zoom levels


aAXEe:
Currently the `ol.interaction.PinchZoom ` zooms to the next whole-number zoom after the pinch completes.  
This is of course good for classic image tiles but not for vector maps. A concrete user request for this: http://stackoverflow.com/questions/40882212/openlayers3-fractional-zoom-on-touch-device-using-pinch-gesture

I suggest to add a parameter like `keepFractionalZoomLevel` to the `ol.interaction.PinchZoom` constructor so that the user can choose whether to keep the fractional zoom or not.

I already tested this idea: https://github.com/aAXEe/ol3/pull/1
A fiddle with a working example: https://jsfiddle.net/wm78prro/7/

ahocevar:
Since wheel zooming now also does not restrict to integer zoom levels any more, I think such a change would make sense. Would you be able and willing to create a pull request?

aAXEe:
As discussed in #6224 keeping the fractional zoom is now default.

To get the old behavior set the new `constrainResolution` parameter.

PR #6224, opened 2016-12-08
Pinch zoom: allow fractional zoom

aAXEe:
This adda a parameter `keepFractionalZoomLevel` to the `ol.interaction.PinchZoom` constructor so that the user can choose whether to keep the fractional zoom or not.

See #6223

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3104,6 +3104,13 @@ olx.interaction.PinchZoomOptions;
  */
 olx.interaction.PinchZoomOptions.prototype.duration;
 
+/**
+ * Zoom to the closest integer zoom level after the pinch gesture ends. Default is `false`.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.interaction.PinchZoomOptions.prototype.constrainResolution;
+
 
 /**
  * @typedef {{handleDownEvent: (function(ol.MapBrowserPointerEvent):boolean|undefined),
@@ -7613,7 +7620,7 @@ olx.view.FitOptions.prototype.maxZoom;
 
 
 /**
- * The duration of the animation in milliseconds. By default, there is no 
+ * The duration of the animation in milliseconds. By default, there is no
  * animations.
  * @type {number|undefined}
  * @api
--- a/src/ol/interaction/pinchzoom.js
+++ b/src/ol/interaction/pinchzoom.js
@@ -29,6 +29,12 @@ ol.interaction.PinchZoom = function(opt_options) {
 
   /**
    * @private
+   * @type {boolean}
+   */
+  this.constrainResolution_ = options.constrainResolution || false;
+
+  /**
+   * @private
    * @type {ol.Coordinate}
    */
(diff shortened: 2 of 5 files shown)


ahocevar:
Nice! In the spirit of #6113 I'd even say that unconstrained zooming should become the default, and the option to constrain resolutions should be named `constrainResolution`.

tschaub:
I agree with @ahocevar on the name and the default behavior.

marcjansen:
And maybe add a note in the upgrade notes about the changed behaviour and how to reestablish the old one?

aAXEe:
I changed the parameter to `constrainResolution` and the default behavior to keep fractional zooms.

The example now shows how to establish the old behavior with a constrained zoom.

Can I already add a note to the upgrade notes?

ahocevar:
Thanks @aAXEe. Yes, you can add a note to `changelog/upgrade-notes.md`, under the `### Next release` heading. The release manager will take these notes and copy them to their own file when the release is being cut.

externs/olx.js
```diff
@@ -3104,6 +3104,15 @@ olx.interaction.PinchZoomOptions;
  */
 olx.interaction.PinchZoomOptions.prototype.duration;
 
+/**
+ * `true` to zoom to next whole-number zoom after the pinch completes;
+ * `false` to keep the zoom at an fractional number.
+ * Default is `false`.
```
tschaub: When documenting booleans, just describe the behavior when true (and then identify the default).  For example, this could be:

> Zoom to the closest integer zoom level after the pinch gesture ends.  Default is `false`.

src/ol/interaction/pinchzoom.js
```diff
@@ -29,6 +29,13 @@ ol.interaction.PinchZoom = function(opt_options) {
 
   /**
    * @private
+   * @type {boolean}
+   */
+  this.constrainResolution_ = options.constrainResolution ?
+options.constrainResolution : false;
```
tschaub: I'll see about updating our linter config to catch this.  There should be two indents in a multi-line conditional expression.  E.g.
```js
var foo = bar ?
    bam : baz;
```

tschaub: Alternatively, to avoid repeating the variable as the conditional and consequent, this could be:
```js
this.constrainResolution_ = options.constrainResolution || false;
```

aAXEe: The second one is not only simpler and easier to read .. it also solves the indention problem ;)
Thanks for the notes!

tschaub:
Thanks for the great contribution @aAXEe!

I-Maps:
This really made a mess out of all my zoom level dependent functions and layer changes.   I had to add the new PinchZoom code with constrainResolution set to true to fix it.   My comment is that it would probably be better to leave old functionality the as the default on new additions like this.  It would be better to let everyone choose to add new code to get new functionality rather than requiring everyone to add code to fix things that get broken if they don't add the new code.

PR #6226, opened 2016-12-09
Improve the headers styles in the api doc

### LINKED ISSUE #5093: Cannot select words within first sentence of API docs
Open for example [this](http://openlayers.org/en/v3.14.2/apidoc/ol.source.Raster.html) page and try to select any word in first sentence:

> A source that transforms data from any number of input sources using an array of ...

This problem caused by CSS style from [jaguar.css](https://github.com/openlayers/ol3/blob/master/config/jsdoc/api/template/static/styles/jaguar.css):

``` css
.nameContainer .name, .prettyprint.source .pln {
  padding-top: 50px;
  margin-top: -50px;
}
```

What is the goal of this style?

tchandelle:
Fix #5093

To clarify the problem, it's impossible to select the last line (in the issue, the first line happens to be the last) of any members or methods (except the very last one) because the next header use a big padding-top that cover that last line.

To correct the problem, I moved the anchor (id=".name-of-prop") to an empty div with no width.

### DIFF
--- a/config/jsdoc/api/template/static/styles/jaguar.css
+++ b/config/jsdoc/api/template/static/styles/jaguar.css
@@ -45,9 +45,11 @@
 body {
   padding-top: 50px;
 }
-.nameContainer .name, .prettyprint.source .pln {
+.nameContainer .anchor {
   padding-top: 50px;
   margin-top: -50px;
+  width: 0px;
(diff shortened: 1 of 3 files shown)


tschaub:
Thanks @tchandelle!

PR #6230, opened 2016-12-10
Ignore duplicated attributions

tschaub:
This is a rebased/squashed version of @aberenyi's work in #6185.  I added one additional commit to avoid looping through all attributions for each source.


aberenyi:
@tschaub looks good to me

ahocevar:
Post-merge +1. Thanks!

PR #6243, opened 2016-12-14
Consistent spacing between keys and values in object literal

fredj:
Related to https://github.com/openlayers/eslint-config-openlayers/pull/9


fredj:
Thanks for the review

PR #6246, opened 2016-12-14
Re-render vector tiles when layer has changed

### LINKED ISSUE #6241: VectorTile.changed() regression in 3.20.0
Changing some properties affecting how a VectorTile layer is rendered and then calling `ol.layer.VectorTile.changed()` to redraw the layer works in 3.19.1 and earlier but has no effect in 3.20.0. If the layer is redrawn for another reason (panned, etc.), the changes show so it's clearly the `changed()` method not doing anything anymore.

Is this a regression or intended behavior? Note that I do not want it to reload data from the source, just redraw itself.

ahocevar:
When a vector tile layer changes (e.g. because its style changes), we should not skip rendering of the tiles.

Fixes #6241.

### DIFF
--- a/src/ol/renderer/canvas/tilelayer.js
+++ b/src/ol/renderer/canvas/tilelayer.js
@@ -34,10 +34,10 @@ ol.renderer.canvas.TileLayer = function(tileLayer) {
   this.renderedExtent_ = null;
 
   /**
-   * @private
+   * @protected
    * @type {number}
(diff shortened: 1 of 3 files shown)


marcjansen:
Thanks @ahocevar, very clean & the tests make the change easily understandable.

Please merge.

ahocevar:

PR #6255, opened 2016-12-15
Don't reset the css cursor if it's not needed

### LINKED ISSUE #6252: Translate Interaction Overrides Cursor Style
## Issue

I have a listener for the map "pointermove" event such that anytime the user "mouses" over a feature, the cursor will change from the default to the "pointer" cursor. Inside the listener, I essentially use the "forEachFeatureAtPixel" fn to return back whether the mouse was over a feature or not.

However, when also utilizing the Translate interaction (configured for **only** 1 feature on init, not all) then the Translate interaction appears to override the cursor (for all features on the map) meaning no "pointer" cursor is displayed and it resorts to the default cursor.

When searching in the Translate interaction source code, I came across the featuresAtPixel_ fn (https://github.com/openlayers/ol3/blob/v3.20.0/src/ol/interaction/translate.js#L199-L210) and the handleMoveEvent_ fn (https://github.com/openlayers/ol3/blob/v3.20.0/src/ol/interaction/translate.js#L168-L187) although I doubt those are the culprits.

## Versions Tested

- 3.18.2 (issue exists)
- 3.20.0 (issue exists)

## Use Case

I have several different layers with different types of features in each layer. Some layers contain features that will be selectable (hence why the cursor needs to be changed to a pointer) and some layers contain features that will **not** be selectable but instead simply be moveable (hence the Translate interaction).

So I utilized the "pointermove" event to display a "pointer" cursor on any of the features that are to be selectable. The Translate interaction's feature list was also set to only the features which are to be moveable.

## Steps to Reproduce

To reproduce the issue, I created a simple fiddle at the following link...https://jsfiddle.net/d3gpcdeL/.

In the simplified fiddle, I created a layer containing a line which is intended to be the layer containing features which are only moveable. There is also another layer containing a polygon which is intended to be the layer with features that should be selectable (but not moveable) which should help illustrate the Use Case I mentioned above.

- If you open that fiddle and hover over the polygon, the cursor will not change to a pointer despite the listener that changes the cursor to the pointer.
- If you remove lines 12 - 14 and line 39 (basically take out the Translate interaction) and hover over the polygon, the cursor will be a pointer which is the intended behavior.

## Intended Behavior

The Translate interaction should not override any cursor styles that it's not managing in it's internal features list.

fredj:
fixes #6252

### DIFF
--- a/src/ol/interaction/translate.js
+++ b/src/ol/interaction/translate.js
@@ -171,7 +171,8 @@ ol.interaction.Translate.handleMoveEvent_ = function(event) {
   // Change the cursor to grab/grabbing if hovering any of the features managed
   // by the interaction
   if (this.featuresAtPixel_(event.pixel, event.map)) {
-    this.previousCursor_ = elem.style.cursor;
+    this.previousCursor_ = this.previousCursor_ !== undefined ?
+        this.previousCursor_ : elem.style.cursor;
     // WebKit browsers don't support the grab icons without a prefix
     elem.style.cursor = this.lastCoordinate_ ?
         '-webkit-grabbing' : '-webkit-grab';
@@ -179,9 +180,8 @@ ol.interaction.Translate.handleMoveEvent_ = function(event) {
     // Thankfully, attempting to set the standard ones will silently fail,
     // keeping the prefixed icons
     elem.style.cursor = this.lastCoordinate_ ?  'grabbing' : 'grab';
-  } else {
-    elem.style.cursor = this.previousCursor_ !== undefined ?
-        this.previousCursor_ : '';
+  } else if (this.previousCursor_ !== undefined) {
+    elem.style.cursor = this.previousCursor_;
     this.previousCursor_ = undefined;
   }
 };
--- a/test/spec/ol/interaction/translate.test.js
+++ b/test/spec/ol/interaction/translate.test.js
@@ -56,13 +56,13 @@ describe('ol.interaction.Translate', function() {
   });
 
   /**
-     * Simulates a browser event on the map viewport.  The client x/y location
-     * will be adjusted as if the map were centered at 0,0.
-     * @param {string} type Event type.
-     * @param {number} x Horizontal offset from map center.
-     * @param {number} y Vertical offset from map center.
-     * @param {boolean=} opt_shiftKey Shift key is pressed.
-     */
+   * Simulates a browser event on the map viewport.  The client x/y location
+   * will be adjusted as if the map were centered at 0,0.
+   * @param {string} type Event type.
+   * @param {number} x Horizontal offset from map center.
+   * @param {number} y Vertical offset from map center.
+   * @param {boolean=} opt_shiftKey Shift key is pressed.
+   */
   function simulateEvent(type, x, y, opt_shiftKey) {
     var viewport = map.getViewport();
     // calculated in case body has top < 0 (test runner with small window)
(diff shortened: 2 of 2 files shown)


fredj:
@mxrugg can you please check if this fixes you issue?

mxrugg:
Thanks for the quick update @fredj! :+1: 

I have a few meetings today but after that I will **immediately** try out the fixes.

mxrugg:
Hey @fredj, I took a look at the code and tested it out and works great. Thanks a ton!

I was able to have a layer devoted to "selectable" only features and another layer devoted to "moveable" only features and the cursor was **not** being overriden when hovering over a "selectable" feature (e.g. it remained the "pointer" cursor).

:question:- Would it be possible to pull up the nested "if" statement and combine it with the "else" as indicated below? I know however there must've been a reason to keep it in the body of the "else" which I probably overlooked, maybe for future expansion.

```js
ol.interaction.Translate.handleMoveEvent_ = function(event) {
    var elem = event.map.getTargetElement();

    // Change the cursor to grab/grabbing if hovering any of the features managed
    // by the interaction
    if (this.featuresAtPixel_(event.pixel, event.map)) {
        ...
    } else if (this.previousCursor_ !== undefined) {
        elem.style.cursor = this.previousCursor_;
        this.previousCursor_ = undefined;
    }
};
```

I ran through the logic a few times and it seems to function like so (if I'm correct)...If the cursor is not over a feature that is managed by the interaction and the previousCursor_ was set (meaning we were just over a managed feature) then set the cursor back to the original cursor (aka previousCursor_) that was displayed prior to moving over a managed feature by the interaction. So essentially, if you never moved over a managed feature of the interaction, previousCursor_ would never be set and there would be no change at all to the cursor unlike the previous version!

With that however, I did a preliminary test but did not check it against also having a Select interaction or Modify interaction, just the Translate so yea. I will _eventually_ need to be able to modify (drag/resize) the features that are "moveable" so I'm not sure if/how the interactions may/could compete against each other (race condition) if the **forEachFeatureAtPixel** fn is maybe used in every interaction (containing feature management). But a whole different topic here. :smile:

fredj:

fredj:
Thanks for the review

mxrugg:
@fredj after a bit more of testing I uncovered what **might** be another potential issue?

Let's say you have 3 features stacked right next to each other side-by-side (e.g. [][][]).

The middle feature is the only one configured with the translate interaction and the first and third features are defined to have different cursors (the first with a pointer cursor and the third with a crosshair cursor - for the sake of a better example) set via the **forEachFeatureAtPixel** fn.

If you move the mouse from the first feature (with the pointer cursor) to the middle feature (configured with the translate interaction) the cursor will change to the grab/grabbing cursor which is the intended behavior (and the pointer cursor is then set as the last known cursor before the translate interaction overrode it). However, if you then move the mouse from the middle feature to the third feature (with the crosshair cursor) then I'm pretty sure a race condition may occur where the cursor over the third feature will be set to the pointer cursor (last known cursor from the translate interaction) and not the intended crosshair cursor set via the **forEachFeatureAtPixel** fn..

I would consider this a **rare** issue as further moving the mouse over the third feature should restore the correct crosshair cursor back.

I would imagine a potential fix may be to also store the id of the feature which could be used to check against when reverting the cursor back but this may not only overcomplicate the logic, but not all features may have an id, and it may also seem wrong for the translate interaction to store a feature id that may not be associated with it's internal feature list.

PR #6258, opened 2016-12-15
Set geometry name properly

### LINKED ISSUE #6253: ol.format.MVT not working with custom geometryName when a 'geometry' property is present
When there's attribute named 'geometry' and 'layer' in the raw protobuff, it was removed and replaced with other values (ol.geom object and layer name).

ahocevar:
Fixes #6253.


ahocevar:

PR #6260, opened 2016-12-15
Enhance documentation for ol.View#fit

marcjansen:
I stumbled upon this today, it took me some time to realize I can also pass in a `ol.Extent` directly. This small change should make it easier to see what you can pass to fit a view to a geometry or an extent.

Before

fit-docs-before

After:

fit-docs-after

Please review.


gberaudo:
Make sense to me. +1.

marcjansen:
Thanks for the review.

PR #6264, opened 2016-12-16
Remove the requirement to provide a size to ol.View#fit()

ahocevar:
Let's make `ol.View#fit()` great - in most cases we do not need to specify a size.

Note that I have marked `ol.View#fit()` stable as part of this pull request. Is this ok?

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,28 @@
 ## Upgrade notes
 
+### Next release
+
+#### Simplified `ol.View#fit()` API
+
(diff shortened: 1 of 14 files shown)


tschaub:
Great idea.  This is one of the most awkward parts of the API.  Please merge.

If we want to change the implementation, I wouldn't be opposed to having the view store a reference to the most recent map for which `map.setView(view)` is called.  I think this would be a pragmatic approach.

tschaub:
A compliment to this change would be to make `view.calculateExtent()` work without a `size`.

PR #6270, opened 2016-12-17
Add a magnify example

tschaub:
Another example demonstrating `postcompose`.

magnify


tschaub:
Merging as an example only change.

ahocevar:
Nice!

PR #6277, opened 2016-12-18
Name modules more like their provide

tschaub:
This makes it so we can unambiguously resolve the path to a file from its `goog.provide()` arg.  For example, `goog.provide('ol.foo.bar')` will be in `src/ol/foo/bar.js` (and not `src/ol/foo/bar/index.js`).

In making this change, I've pulled the `eslint-plugin-openlayers-internal` plugin out into [a dedicated repository](https://github.com/openlayers/eslint-plugin-openlayers-internal).  The new behavior for the `valid-provide` rule came with https://github.com/openlayers/eslint-plugin-openlayers-internal/commit/a0894a76c41f0dc78be4104cf18bd94794b600dc.


tschaub:

PR #6280, opened 2016-12-19
Continue loading tiles when image is not ready yet

### LINKED ISSUE #6279: ol.source.Raster examples broken
The `ol.source.Raster` examples are not working anymore (for example http://openlayers.org/en/master/examples/sea-level.html)

```
ee7b89435041bf518a7d0b29fb1464a6f2f8b71c is the first bad commit
commit ee7b89435041bf518a7d0b29fb1464a6f2f8b71c
Author: Andreas Hocevar <andreas.hocevar@gmail.com>
Date:   Mon Nov 7 09:37:09 2016 +0100

    Refactor VectorTile renderer and add lower resolution clipping

:040000 040000 3f8bc5a5e5e0d9fe103dd23b71881b9f7a16c94f 36cf5f68fca77104345314e10d4a8c51e9f68ece M      src
:040000 040000 6de3592a0145182778c9e5de4aebe1b66cc96cc5 a8c779a606fdebec3470f0f4ebaf1654c066a1c8 M      test
:040000 040000 1da8282d27b626bb0b17f9dd03c83d7e9936664a 75a312f074f36c0c815a35f5626fd2231fbb4fdb M      test_rendering
```

ahocevar:
When the image of a raster source's source is not available, one of the reasons can be that no tiles are loaded yet. So we need to instruct the tile queue to load more tiles also in this case.

(This problem is another reason why we should rethink or get rid of our tile queue)

Fixes #6279.

### DIFF
--- a/src/ol/source/raster.js
+++ b/src/ol/source/raster.js
@@ -287,16 +287,19 @@ ol.source.Raster.prototype.composeFrame_ = function(frameState, callback) {
       imageDatas[i] = imageData;
     } else {
       // image not yet ready
-      return;
+      imageDatas = null;
+      break;
     }
   }
 
-  var data = {};
-  this.dispatchEvent(new ol.source.Raster.Event(
-      ol.source.Raster.EventType.BEFOREOPERATIONS, frameState, data));
+  if (imageDatas) {
+    var data = {};
+    this.dispatchEvent(new ol.source.Raster.Event(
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks for identifying the issue @fredj and for fixing it @ahocevar.

Please use the "regression" label for any other similar fixes.  I'll cut a patch release with the regression fixes.

PR #6283, opened 2016-12-19
Avoid modifying coordinate in forEachLayerAtCoordinate

### LINKED ISSUE #6282: Canvas rendered forEachLayerAtPixel does not work with multiple layers
The [Coordinate value](https://github.com/ahocevar/ol3/blob/3e8c8791d08d9a5ffce82453564b4969684dc530/src/ol/renderer/canvas/map.js#L217-L218) is being manipulated inside the [layer loop](https://github.com/ahocevar/ol3/blob/3e8c8791d08d9a5ffce82453564b4969684dc530/src/ol/renderer/canvas/map.js#L221-L232) so that the method only works for the first layer.

tschaub:
This lets people call `map.forEachLayerAtPixel()` with more than one layer.

Fixes #6282.


tschaub:
Looking into tests.

marcjansen:
Thanks for fixing this!

PR #6285, opened 2016-12-20
Make size argument of ol.View#calculateExtent() optional

ahocevar:
Follow-up to #6264, as suggested by @tschaub (https://github.com/openlayers/ol3/pull/6264#issuecomment-268036472).


ahocevar:
The `[100, 100]` default is completely arbitrary. It's just so we don't bail out and have to throw an exception if no map viewport is found on the page.

PR #6294, opened 2016-12-22
Update code for closure-compiler v20170124

fredj:


src/ol/format/feature.js
```diff
@@ -15,6 +15,7 @@ goog.require('ol.proj');
  * file formats.  See the documentation for each format for more details.
  *
  * @constructor
+ * @abstract
```
tschaub: Curious what makes this `@abstract`.  And if abstract, should the constructor do anything?

src/ol/format/gmlbase.js
```diff
@@ -31,6 +31,7 @@ goog.require('ol.xml');
  * ol.format.GML3.
  *
  * @constructor
+ * @abstract
```
tschaub: Also surprised to see `@abstract` on a function that has code that looks like it should be executed.

tschaub: Ok, I see now that this is about generating warnings if this is instantiated directly.

gberaudo:
Great to see the Travis light turn green.There is a `wip` commit to reword, otherwise it looks good to me and good for merging. Thank you @fred for your efforts on this PR.

probins:
closure-util needs updating and package.json changing accordingly. (Then the next step once this is merged is to replace closure-util with compiler package, at least for building.)

There's some background on the new checks for abstracts which are the reason for most of these changes at https://github.com/google/closure-compiler/wiki/@abstract-classes-and-methods

gberaudo:
Thanks for the reminder @probins. @fredj just opened https://github.com/openlayers/closure-util/pull/274.

probins:
I've run my custom build against this branch, and notice 2 differences from before. One is the addition of polyfills as mentioned in https://github.com/openlayers/openlayers/issues/6338#issuecomment-273205365; if you look at the build, you can see these towards the front of the file. These duplicate the code already present in ol, so either remove the code in ol or turn off the `rewrite_polyfills` flag. The other is that my gulp process using the compiler package, which throws all the src/ files at the compiler, now raises 'Unknown Type' warnings for the externs I'm not using; easily resolved by adding all the extern files whether I use them or not.

src/ol/source/tileutfgrid.js
```diff
@@ -479,7 +480,7 @@ ol.source.TileUTFGrid.Tile_.prototype.onXHRError_ = function(event) {
 
 
 /**
- * Load not yet loaded URI.
+ * @inheritDoc
```
tsauerwein: You couldn't use `@override` here? In the [docs](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#override) it says:

> If no other annotations are included, the method or property automatically inherits annotations from its superclass.

src/ol/render/vectorcontext.js
```diff
@@ -16,7 +16,6 @@ ol.render.VectorContext = function() {
 /**
  * Render a geometry.
  *
- * @abstract
```
tsauerwein: By removing `@abstract` from the methods, the compiler would no longer complain if you would create a child class which did not implement all of the methods. Is that what you want?

src/ol/format/igc.js
```diff
@@ -202,3 +202,24 @@ ol.format.IGC.prototype.readFeaturesFromText = function(text, opt_options) {
  * @api
  */
 ol.format.IGC.prototype.readProjection;
+
+
+/**
+ * Not implemented.
+ * @inheritDoc
+ */
+ol.format.IGC.prototype.writeFeatureText = function(feature, opt_options) {};
```
tsauerwein: Any reason that one time

```
/**
 * Not implemented.
 * @inheritDoc
 */
ol.format.IGC.prototype.writeFeatureText = function(feature, opt_options) {};
```
... is used, and another time
```
/**
 * @override
 */
ol.format.IGC.prototype.readGeometryFromText = function() {
  throw new Error('Not implemented');
};
```

What's the difference?

src/ol/format/xmlfeature.js
```diff
@@ -7,6 +7,9 @@ goog.require('ol.format.FormatType');
 goog.require('ol.xml');
 
 
+/* eslint-disable valid-jsdoc */
```
tsauerwein: Why is this required? Maybe leave a comment explaining the problem.

src/ol/format/igc.js
```diff
@@ -226,8 +226,9 @@ ol.format.IGC.prototype.writeGeometryText = function(geometry, opt_options) {};
 
 
 /**
- * @override
+ * @inheritDoc
  */
-ol.format.IGC.prototype.readGeometryFromText = function() {
-  throw new Error('Not implemented');
+ol.format.IGC.prototype.readGeometryFromText = function(text, opt_options) {
```
tsauerwein: Maybe you could add a (basic) test for this new function?

fredj:
@probins `rewrite_polyfills` is now `false`. Thanks

probins:
@fredj I see there was another release last week. Have you tried compiling with that? If there aren't any major issues, ISTM it would be better to move on to that release.

gberaudo:
@probins, yes that version compiles fine with OL3 with just a small change to ol-ext. That said it would require updating closure-util once more and this PR is already long overdue.
So I prefer we do it in another PR, which I will propose as soon as this one is merged.

gberaudo:
@probins https://github.com/openlayers/closure-util/pull/280 was merged incredibily fast so I pushed a new commit to switch to v20170124.

probins:
looks like the same change needs making to the examples compilation :-)

@ahocevar mentioned a new ol release. I think this PR should be part of that. As you say, it's long overdue.

gberaudo:
@probins, @fredj the build is now green.

src/ol/format/igc.js
gberaudo: Right, fixed.

fredj:
Thanks for your work on this @gberaudo

gberaudo:
Your welcome, thanks for initiating it @fredj.

PR #6300, opened 2016-12-23
Add missing hidpi option for ol.source.ImageArcGISRest

### LINKED ISSUE #6272: Missing hidpi option for ol.source.ImageArcGISRest
Hi,

I'm working on a multi layer project that ArcGIS Server is one of the providers and I'm having problems with ol.source.ImageArcGISRest and ol.layer.image. As I checked, the map extent is not suitable with the extent in the URL? 

I tried different things to fix it, but I can't. But I found a issue if the map width is larger than 1000px, than image is not aligned to map. 

Also I realized that even I'm not using the retina screen, my image size sent to server is doubled due to pixelRatio. 

Any ideas to fix the problem?

**Image that has 800px width :**
width-800px

**Image that has 1200px width :**
width-1200px

fredj:
fixes #6272

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5064,6 +5064,7 @@ olx.source.OSMOptions.prototype.wrapX;
 /**
  * @typedef {{attributions: (Array.<ol.Attribution>|undefined),
  *     crossOrigin: (null|string|undefined),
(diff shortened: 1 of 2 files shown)


fredj:
Thanks for the review

Issue #5226, opened 2016-04-13
Proposition:  use goog.module


gberaudo:
I have been experimenting with goog.module. The results look promising; before investing more time I would be pleased to get your feedback.

In relation to #4128.

Rationnal: gradual and non-breaking step towards ES6

Details:

OL3 dependencies are handled using goog.provide / goog.require.
The symbols are evaluated in the global scope and made available in a hierarchy attached to the global object.

On the other hand, ES6 and AMD/common.js evalutate the code in an isolated scope.
It is not attached to the global object.

goog.module goals is to allow both styles, at the same time, making it possible to transition gradually towards ES6 support, without breaking backward compatibility.

A good explanation about goog.modules is here:
https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide

With @fredj, we created a branch in closure-util:
https://github.com/gberaudo/closure-util/tree/goog_module
It makes goog.module understood by closure-util and do the wrapping to isolate the module.

Here are experiments in OL3 code:
https://github.com/gberaudo/ol3/tree/goog_modules
It shows how to migrate namespaces, constructors, typedefs and enums.

Notes:
A file using goog.module can have only one goog.module and no goog.provide.
Typedefs must be moved to their own files but keep goog.provide.
There are some issues with jsdoc parsing for generating the exports (see commits).

probins:
ISTM there's 2 types of task here:

1. those, like moving the typedefs, which don't affect the actual code and can be done anyway; I'd suggest putting these in a separate directory, like the externs. Also in this category are removing the `goog.require`s which aren't needed anyway (as in #5210); and renaming files to match module names, such as `layer/vectorlayer.js` to `layer/Vector.js`
2. changing the actual code, as with changing the current global vars to the `exports` var. The problem with fundamental changes like this is that they're likely to quickly lead to merge conflicts, so if this can be done on a file by file basis, that would be better.

I was going to offer to put some time into the first, if people agree that makes sense, so happy to work with you on this if that can be arranged.

I'd agree that using `goog.module` could be a good transitional step. It doesn't remove dependency on GCC, but it does move in that direction by changing to something that is module-ish.

tschaub:
Looks promising @gberaudo.  I agree with @probins about pulling out the changes that aren't specific to `goog.module` - assuming these are changes we can make without upgrading `closure-util`.

[...]

PR #6302, opened 2016-12-23
Publish a package for use with ES module bundlers

tschaub:
The objective of this branch is to make it so we can publish a package that is more useful for application developers.  We currently publish an [`openlayers` package](https://www.npmjs.com/package/openlayers) that is really only usable if people develop applications with the Closure Library and Closure build tools or use our custom build tool.  This branch proposes that we publish an [`ol` package](https://www.npmjs.com/package/ol) that can be used by mainstream package managers and module bundlers (e.g. webpack, Rollup, JSPM, or Browserify).

The branch includes a codemod (for use with [jscodeshift](https://github.com/facebook/jscodeshift)) that transforms our sources into ES6 modules.  Running `make package` generates modules in the `build/package` directory.  The `ol` package can then be published from this directory.

There are still things to work out, but here is how to use it.

Install the package.  For example with `npm`:
```
npm install ol@beta
```

Import just what you need:
```js
import Map from 'ol/map';
import View from 'ol/view';
import TileLayer from 'ol/layer/tile';
import XYZ from 'ol/source/xyz';

new Map({
  target: 'map',
  layers: [
    new TileLayer({
      source: new XYZ({
        url: 'https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png'
      })
    })
  ],
  view: new View({
    center: [0, 0],
    zoom: 2
  })
});
```

A few things to note:
 * module paths are like the the `ol.foo.Bar` names, except they have `/` instead of `.` and are lowercase
 * ~~boolean defines are global, use `_` instead of `.`, and are uppercase (e.g. `OL_DEBUG` instead of `ol.DEBUG`)~~ (update: punting on defines for now)
 * builds are going to be huge because of how much code we unconditionally require in the library (e.g. +100 kB of unused WebGL code - see #6303 for an alternative) (update: [see below](https://github.com/openlayers/ol3/pull/6302#issuecomment-269412335) for detail on advanced compilation with the Closure Compiler) (update 2: with #6312 and #6314, far less unused code is included in bundles and Uglify minification is getting reasonable)

More complete examples using the `ol@beta` package:

 * for use [with webpack & Uglify](https://gist.github.com/tschaub/79025aef325cd2837364400a105405b8)
 * for use [with Rollup & Uglify](https://gist.github.com/tschaub/8beb328ea72b36446fc2198d008287de)
 * for use [with Rollup & Closure Compiler](https://gist.github.com/tschaub/32a5692bedac5254da24fa3b12072f35)
 * for use [with Browserify & Uglify](https://gist.github.com/tschaub/4bfb209a8f809823f1495b2e4436018e)

TODO:
 - [x] ~~use build flags and~~ compress with Uglify
 - [x] demonstrate use with Browserify
 - [x] demonstrate advanced compilation with the Closure Compiler


probins:
I think this is a great piece of work. I've not come across jscodeshift before. It takes a while to work out, but once you get used to it (and once you discover tools like http://astexplorer.net/) it's a great way to make bulk changes to code: writing programs to change programs.

But ...

The problem I have with publishing this as a package is that it creates a different api which is not compatible with existing apps - a major version change - without any documentation or examples showing how to use it. If people look in the source, they find comments which also relate to the ol3 api, not this new one. I think this is likely to confuse people. I think the readme should acknowledge this, and include instructions for use: include your 'few things to note' and tell people that this is mainly intended for adapting existing ol3 apps to mainstream build/bundle tools: "take your existing app and change all the global `ol` references to use imported variable names".

It's also not clear to me how this fits in with the aim of converting to ES6 modules. The codemod can easily be adapted to change to ES6, and I can see how the output from that could be used as the basis for a new improved ol4. If the comments/types were changed as well, you could use advanced-mode Closure on that to create a build similar to the present one, but with much more flexibility, as it would be based on standard ES6 modules rather than Closure's esoteric globals. JSDoc similarly. Node tools are slowly moving to ES6 - Webpack 2 for instance - a trend which I think will accelerate once Node implements V8's ES6 module implementation. Apps can then be modified to use import/export instead of the current globals. But if you go with CJS, at some point this will have to be changed to ES6, and people will have to change their apps twice. Not only that, but CJS is only usable in Node, whereas ES6 is designed to be usable in both browser and server. You comment on build sizes, but this should improve with tools like Rollup and Webpack 2 which can do 'tree-shaking' with ES6 modules. Webpack 2 can also split the code, a major aim for me, though I've not tried this yet to see how it might work in practice. All in all, going with ES6 seems a much better option to me than CJS.

tschaub:
@probins - I've updated the codemod to generate ES6 modules.  Clearly ES6 modules are the way forward.  I'm primarily interested in making OpenLayers easier to use - I don't think there is as much value in a goal like "use ES6 modules."  I've put together [one example](https://gist.github.com/tschaub/8beb328ea72b36446fc2198d008287de) of how the `ol@beta` package can be used (right now, not in the future) with Rollup.  We can do the same with Webpack etc. (unfortunately, it is more involved to consume ES6 dependencies with Browserify).  It is also worth considering that our project has dependencies on CommonJS modules (pbf, pixelworks, rbush, vector-tile).  And those dependencies bring in transitive dependencies that are also CommonJS modules.  So again, while we agree that ES6 modules are the future, my goal is to make the project more usable right now.  After fleshing out some additional examples, we'll see if that is successful.

I'll also take a look at generating Closure Compiler builds from the Rollup output.  Though everybody likes to repeat "tree shaking!", we'll need a major restructuring of the library to really benefit from selective named imports (this will be far easier to accomplish with a major release).

tschaub:
I added instructions for using the `ol@beta` package with `webpack@beta` here: https://gist.github.com/tschaub/79025aef325cd2837364400a105405b8

tschaub:
If anybody is interested in experimenting with rollup->closure, here is a `rollup.config.js` that works:
```js
const compile = require('google-closure-compiler-js').compile;

module.exports = {
  entry: 'main.js',
  targets: [
    {dest: 'bundle.js', format: 'iife'}
  ],
  plugins: [
    require('rollup-plugin-node-resolve')(),
    require('rollup-plugin-commonjs')(),
    {
      name: 'closure-compiler-js',
      transformBundle: function(code) {
        const flags = {
          createSourceMap: true,
          jsCode = [{
              src: code
          }]
        };
        const output = compile(flags);
        return {code: output.compiledCode, map: output.sourceMap};
      }
    }
  ]
};
```

This assumes you have:
```
npm install --save-dev rollup rollup-plugin-commonjs rollup-plugin-node-resolve google-closure-compiler-js
npm install --save ol@beta
```

This could be improved with logging as in `rollup-plugin-closure-compiler-js` - though we can't use `processCommonJsModules` as in that plugin.  And this is just simple compilation.  Still huge bundle size.

probins:
there is a known issue with defines in modules https://github.com/google/closure-compiler/issues/1601 - not sure if this is the same problem. Are they being renamed/mangled? Did you try window.FLAG? Or exporting the symbols?

Closure compiler needs typedefs.js, so would suggest leaving it in src/ and changing codemod to ignore it.

> If the comments/types were changed as well, you could use advanced-mode Closure on that to create a build similar to the present one

actually, types shouldn't need changing for the compiler, as types can be any string. Might need changing for JSDoc though.

I totally agree on reducing dependencies. This is same complaint as in https://github.com/openlayers/ol3/issues/2381#issue-37865709

probins:
#### Use in browser

I would suggest changing the codemod to add '.js' to all the import specifiers. This should work the same in Node, and would also work in the browser. I have made this change, added a test directory, and pushed the result to https://github.com/probins/oles6

Then non-map loads like https://rawgit.com/probins/oles6/master/test/testextent.html should work in Edge with experimental JS flag enabled (the only browser which has so far released a usable module loader).

maps do not work, because ol.ext also need to be converted to ES6 modules. A simple map with no vector layers does not need ol.ext, but these are always loaded because of excessive dependencies. If I hack the code to (a) define flags, and (b) remove unwanted dependencies, as in https://github.com/probins/oles6/commit/3a66183656bb5b477fd904ea1c17a6ded93b5215, then a simple map as in https://github.com/probins/oles6/blob/master/test/testes6.js should also work in Edge.

For other browsers, I have converted to System.register
```
babel --plugins transform-es2015-modules-systemjs es6 -d reg
```
and pushed to https://github.com/probins/olreg. Then - ta-da - the simple map in https://rawgit.com/probins/olreg/master/test/testreg.html, which uses the SystemRegisterLoader, will work. This is effectively a debug load, so not recommended on a flaky mobile connection, but response times are acceptable on my fibre connection. Run the files through uglify (or Closure simple, or babili, which is looking promising), create a release which can then be cached and loaded via cdn.rawgit.com, and preload dependencies to eliminate round-trips, and this should be usable in production. I still have to test this out, but having independently loadable modules as in my extent example is a major step forward, as it removes the need to load a large one-size-fits-all build upfront.

So, my priorities are:

* sort out ol.ext (for example, for rbush, could use build on unpkg.com, keeping this as a global, though it's probably better to remove the UMD wrapper and replace with an ES6 one)
* remove unwanted dependencies, making the define flags unnecessary

tschaub:
I've ticketed google/closure-compiler#2217 to reflect the remaining issue with `@define`.

It is even clearer to me now that we need to stop using `@define` to configure conditional dependencies - specifically the alternative map renderers and layer-specific renderers.

In addition, I think it would be worth reconsidering our use of `ol.DEBUG`.  I'm not convinced that we are getting a lot of benefit from it, and since it is awkward to work unless working with the Closure Compiler and the global `ol` type builds, it would be unfortunate to start including more code behind `ol.DEBUG` conditions.

tschaub:
Ok, at this point it looks to me like the Compiler cannot successfully compile a Rollup bundle in advanced mode.  I'm seeing an issue where the Compiler is optimizing away an enum but not recognizing the use of all of its properties.  Specifically, `_ol_layer_Base_.Property` is compiled to a variable declaration that is missing a declarator for the `SOURCE` property.  All other properties are used by `_ol_layer_Base_` methods.  The `SOURCE` property is used in `_ol_layer_Layer_` methods.  Rollup adds a `_ol_layer_Base_$1` alias for `_ol_layer_Base_`, `_ol_layer_Layer_` methods use this alias.  The Compiler isn't recognizing that `_ol_layer_Base_$1.Property.SOURCE` is accessing `_ol_layer_Base_.Property.SOURCE`.

I've confirmed that the Rollup bundle runs as expected.  I've tried producing a minimal example that demonstrates the Compiler issue, but the optimization strategies are different with smaller, contrived examples.  Though looking at the uncompiled and compiled code for the [map example above](https://github.com/openlayers/ol3/pull/6302#issuecomment-269098457) (minus the `defines`), it looks clearly like a Compiler issue to me.

I'm guessing that this problem would be resolved by moving these enums to dedicated modules instead of tacking them on to existing constructors.

tschaub:
Ok, a bit of a breakthrough.  This now includes the code from #6308.  After running into rollup/rollup#1189 (which may be considered expected behavior), I've reworked the codemod here so our modules make it through Rollup a bit more intact.  And boom, `ADVANCED` compilation achieved.

Here is a complete, end-to-end example of how you can build an application with the `ol@beta` package, Rollup, and the Closure Compiler (in `ADVANCED` mode).  And with camelaissani/rollup-plugin-closure-compiler-js#4 in, [the Rollup config](https://gist.github.com/tschaub/32a5692bedac5254da24fa3b12072f35#file-rollup-config-js) got even simpler.

As noted in [the readme](https://gist.github.com/tschaub/32a5692bedac5254da24fa3b12072f35#file-_readme_-md), the Compiler is crazy noisy.  Unfortunately, the `extraAnnotationNames` flag is not supported for the JS version of the compiler.  So we get warnings for our `@api` annotations.  I'm sure this can be addressed too.

This is starting to feel promising to me.  I'd still like to demonstrate an end-to-end example with Browserify and webpack.  And, it would be great to deal with our defines problem.  In the meantime, I'm going to make this package opinionated about the map renderer (though we still get unnecessary layer renderer code).

Goodbye `closure-util`.

probins:
thanks for all your work on this, @tschaub 

The noisy compiles should be reduced by #6294, right? Though I notice that the rollup plugin is not using the latest compiler version, so there may be some additional issues there. The annotation tag warning should be suppressable https://github.com/google/closure-compiler/wiki/Warnings - or maybe better would be to raise an issue for the js version to add this.

I think you should change `var Map =` in your example. Like `var Object/Math/JSON/... =`, it will not actually produce an error, but will screw up any following code that tries to use the global that you've just redefined. This is less of a problem in a module, but `var OLMap =` or something would be better practice IMO.

tschaub:
Thanks for the feedback @probins.  The noise from the Compiler is largely about our use of unrecognized annotations (`@api` mostly).  These are just warnings and can be suppressed.  It would be nice if the JS version of the compiler supported `extraAnnotationNames` (ticketed as google/closure-compiler-js#48).

Regarding the use of `Map` in the examples, please keep in mind that these are just examples.  The library doesn't use the identifier `Map`.

bjornharrtell:
Nice to see work towards this modernization, and excellent information on the usefulness of codemod. I adapted a smallish project to use `3.21.0-beta.12` from npm and I haven't found any issues yet. Though I agree on the point that major restructuring will be needed to enable named imports etc. and that it might require a new major release.

tschaub:
I added a basic [Browserify example](https://gist.github.com/tschaub/4bfb209a8f809823f1495b2e4436018e).

Still digging to find ways to make Uglify builds more compact.  I'm getting 30K gzipped builds with the Closure Compiler and 60K gzipped with Uglify.  There is still a lot of code that is unconditionally imported/executed (e.g. the map requires the WebGL renderer which has a transitive dependency on RBush) and Uglify cannot effectively determine that it is all side-effect free.

@ahocevar and I have talked this through and think that the way to get smaller builds for all is to have `WebGL*` versions of the map, tile layer, image layer, and vector layer.  I'll keep digging to see if there are other easy wins.

tschaub:
In case anybody is interested, here is an [interactive bundle explorer](https://bl.ocks.org/tschaub/raw/83ba19ce84a1c20540e72946d623cb1d/) that shows the relative contribution to the bundle size:

[image](https://bl.ocks.org/tschaub/raw/83ba19ce84a1c20540e72946d623cb1d/)

I'm using the same `main.js` from [the examples above](https://github.com/openlayers/openlayers/pull/6302#issue-197436435) in this example.  In the graphic above, I'm hovering over `ol.render.webgl`, which contributes 20% to the bundle size.  We're still working to make this and other unused code easier to exclude.

tschaub:
I'll stop dropping big graphics here, but this is good news.

image

After #6314, simple dead code elimination brings the `ol.render` code down to 6% of the bundle size (where before `ol.render.webgl` made up 20%).

tschaub:
Here are the latest gzipped bundle sizes with `ol@3.21.0-beta.14` and the `main.js` above:

| Method | Size |
|---------|-----|
| Rollup + Closure Compiler | 31K |
| Rollup + Uglify | 55K |
| Webpack + Uglify | 59K |
| Browserify + Uglify | 64K |

Results with [Babili](https://www.npmjs.com/package/babili) are similar to (slightly worse than) the Uglify results.  It would be nice to get Uglify builds down closer to the Closure Compiler builds - but the difference should be less pronounced for applications that use more of the library.

bjornharrtell:
Is it not hard/impossible to reach closure levels of dead code elimination without named exports (vs. single object default export)?

tschaub:
> Is it not hard/impossible to reach closure levels of dead code elimination without named exports (vs. single object default export)?

@bjornharrtell - this is not really about named exports vs. single default export.  It is more about functional vs. object oriented design.  For example, if you use `ol.geom.Polygon`, the Compiler figures out that you don't need all of `ol.geom.flat.simplify` because you don't call `polygon.simplify()`.  If instead we organized the library so you needed to explicitly import a `simplify` function and call `simplify(polygon)`, then you wouldn't be importing dead code in the first place.

ahocevar:
@tschaub I think what you just said only is a good cure for things we expose through the API. Does it also help for things we use internally? E.g. since we use `polygon.simplify` internally (the vector renderer does), we would still explicitly import `simplify`, only somewhere else in the library.

bjornharrtell:
I agree that it's about functional vs. object oriented design but there are already some degree of functional design in ol3. For example the extent.js module a collection of pure functions. But the current codemod exports applies them to single object which is then default exported, so you can only import the module as a whole. So, to be able to explicitly import a single function from the extent module would it not be a requirement to export each function as a named export?

tschaub:
@bjornharrtell - yes, good example.  I was picking on `Polygon` since it pulls in a bunch of code that is unused in the example above.  The `ol/geom` code accounts for 12% of the bundle and is entirely unused (only required because the drag-box interaction has a `getGeometry()` method).  In that same bundle, `ol/extent` accounts for 3% of the build.  So it might be possible to shave a bit off there - though I haven't confirmed how much of those functions are already used in the bundle.

So, when we convert the library to use ES6 modules, we definitely can restructure things to use more named imports and exports.  It may be possible to make the codemod use named exports and imports where we assign to (non-prototype) properties of the `goog.provide()` arg.  I'll try to dig into this - though I'm dubious it will result in significantly smaller bundles (with simple tree-shaking and dead code removal) given that [the big offenders](https://bl.ocks.org/tschaub/83ba19ce84a1c20540e72946d623cb1d) in the example bundle are all `Constructors` (plus prototype properties) and not straight `functions`

bjornharrtell:
Agreed, with the current design perhaps there are only small savings to be had regarding named experts. But is not a functional redesign pretty far off? Makes me wonder what short term gains there is to be had with es6 modules and alternative bundling for ol, even though i very much prefer to write new code in that way.

probins:
> The ol/geom code accounts for 12% of the bundle and is entirely unused

not quite true: view.fit() requires SimpleGeometry and Polygon

tschaub:
> Makes me wonder what short term gains there are to be had with ES6 modules and alternative bundling for ol

@bjornharrtell - I agree that ES6 modules don't bring us a ton of immediate wins (over CommonJS modules).  As I've said elsewhere, my goal is to provide a more usable package for consumers of the library.  While it will continue to change quickly, my guess is that ES6 modules involve *more* work for people to consume than CommonJS modules.  But I know that either ES6 modules or CommonJS modules are accessible to more people than the current files with `goog.*` dependencies.

tschaub:
> not quite true: view.fit() requires SimpleGeometry and Polygon

@probins - good point.  So "raster-only" apps are stuck with a lot of unused geom code (as long as we have things like `polygon.simplify()`).

probins:
> "raster-only" apps are stuck with a lot of unused geom code

though both cases can quite easily be removed. dragzoom (and so dragbox) can be removed from default interactions (or, better, change map so defaults are no longer 'required' #6311). `view.fit()` only requires geom to convert an extent to a polygon; change that so that only a polygon is handled and the requires can be removed; developers that want to fit to an extent can easily use `polygon.fromExtent()` in their apps. However, as this is 'api stable', this would be an api-breaking change (even if only a minor one).

gberaudo:
It worries me to see PRs like https://github.com/openlayers/openlayers/pull/6313 merged without much discussion, during the Christmas break.

@tschaub, I propose we organize a hangout this week where you can present your proposition and see what are the implications in OL3 and for the projects using OL3. We can also discuss my proposition about `goog.module` and explain the Camptocamp vision for the OpenLayers library and what we are doing in our projects related to using ES6 / `goog.module` and getting ride of `closure-util`.

gberaudo:
Hi all interested in modernizing OpenLayers and projects depending on OpenLayers.
Are you available for a discussion this Thursday (usual time (17:30 Swiss zone, CET) and hangout URL (http://bit.ly/ol-hangout)) ?

Discussion topics:
- using ES6 (code, modules) in OpenLayers;
- future of closure-util, feedbacks of using other loaders for development/debug;
- paths to ES6 (projects compiling with closure compiler, goog.module, ...)
- ...

Ping @openlayers.
Ping @tschaub, @probins, @ahocevar, @sbrunner, @fredj, @tsauerwein.

marcjansen:
I'll likely have no time, sorry

bjornharrtell:
Is eavesdropping ok? :)

gberaudo:
@bjornharrtell, sure it is fine with me.

ahocevar:
I'd be available for a call tomorrow.

tschaub:
I'll be there as well.

probins:
I won't be able to participate, so here's a few thoughts.

I think this PR makes it clear that there's more to 'modernizing OpenLayers' than just changing syntax. ES module loading in the browser has been a long time coming, but initial implementations are either released or on the way (landed in Safari TP last week, for example). The ES modules produced by this codemod can be used in these implementations, but are not very satisfactory because the amount of code that has to be loaded is way too big. The last time I looked, the most important class, map, had 154 dependencies. As another example, OL formats could in theory be used standalone for converting data from one format to another, but geojson has 70 dependencies and kml 80. Removing WebGL from map, separating out vector classes and making the default controls/interactions/projections optional have been mentioned. The main problem with the formats is that they require all geometry types whether they're needed by the data in question or not; they all require ol.Feature, which requires style, even though most formats don't support style; it would probably also be better to separate read and write. You can probably resolve these issues whilst keeping the same api by adding new classes or something similar, but is this worth it? Is it not better to stop/wind down development of v3, and start on a new, though similar, api based on small independently loadable ES modules?

There are programs like lebab which can help converting current code to ES6. Using standard ES instead of custom Closure code should make it easier for more people to contribute, either directly or by creating alternative/custom modules. But there are other simplifications I'd like to be considered:
* OL implements its own version of Object; is this justified?
* ES6 introduced Set/Map for collections; is ol.Collection still needed?
* can ES6 class get/set replace current custom accessors?
* are there simpler ways of implementing observables and the current custom event handling? (there is a TC39 Observable proposal, but it's not progressing very quickly)
* I'd like to explore replacing the current DOM handling with custom elements (`ol-map`, `ol-overlay`, `ol-scaleline` ...), using shadow DOM for CSS scoping. Simple maps could then be created by people who know HTML but are not JS programmers.

tschaub:
@probins - please make these suggestions in a separate thread.  The purpose of this pull request is to publish a package that is more usable.  That is, we want to publish a package that is consumable by mainstream module bundlers.  And, it is my belief (shared with other core developers) that this can be achieved in the 3.x series.

Clearly there is a lot that can be done in a major revision of the library.  But I'd like to keep the discussion on this pull request more focussed.

tschaub:
> Great but I don't understand why create a script to convert the sources files instance of having a way to support ES6 modules natively as it's purpose here: #5226

@sbrunner - #5226 is not about supporting ES6 modules natively.  That pull request is about introducing another dependency on the Closure Library (for the definition of `goog.module()`).  We have worked very hard to remove the dependency on the Closure Library, and have no interest in brining it back.  For anybody interested in talking about `goog.module()` (or any other conventions from the Closure Library), please do so on separate threads.

This pull request is about publishing a more usable package.

gberaudo:
Thank you all for your participation. To summarize for the ones that could not come, we discussed many topics, notably:
- there is a common vision that the OpenLayers code will be written in ES6 with modules, compilable with the Google Closure Compiler;
- transforms will be applied to this source code to generate other module formats, eventually transpiling code to ES5;
- before this happens, we will generate experimental betas, without any guarantee (see this PR #6302);
- https://github.com/openlayers/closure-util/pull/267 is intended for projects wanting to use goog.module, merging would avoid un unnecessary fork;
- Camptocamp is switching project code to goog.module. There is no requirement to OpenLayers to switch;
- the transforms to convert goog.provide code to goog.module will be contributed in ./transforms/; (after #6302);
- https://github.com/openlayers/closure-util/pull/274 and https://github.com/openlayers/openlayers/pull/6294 port the OpenLayers code base to latest GCC release and is ready for review;
- GCC is implementing the nodejs and browser module resolution algorithms (see https://github.com/google/closure-compiler/pull/2130);
- there are issues with some OpenLayers enums when compiling project code with GCC, OpenLayers code should take / or accept strings for enums.

Please leave a comment if I misunderstood or forget something.
Happy hacking!

sbrunner:
ŧhanks for the resume @gberaudo :-)

ahocevar:
@tschaub, what needs to be done to finish this pull request? Anything I can help with? I would like to create a release soon, and make updating the `ol` package part of the release process.

probins:
> It would be nice if the JS version of the compiler supported extraAnnotationNames

this should be fixed with the newest compiler release, so if #6294 is merged before a new ol release, then this package should compile with latest JS compiler.

probins:
one other point: I think the readme should document that webgl is disabled

tschaub:
I've updated the readme and associated gists.  I think this is ready to go.  We can shuffle things around to accommodate additional transforms for other packages if there is reason.  That and additional changes to this transform can be handled after this is merged.

PR #6305, opened 2016-12-27
Moved hit Tolerance parameter to own example.

simonseyock:
This PR separates the select interaction and hit tolerance examples.

### DIFF
--- a/examples/hit-tolerance.html
+++ b/examples/hit-tolerance.html
@@ -1,0 +1,21 @@
+---
+layout: example.html
+title: Hit Tolerance
+shortdesc: Example of using the hitTolerance parameter.
+docs: >
+  Choose the hit tolerance parameter in pixels and try to click on or near the line feature. If the feature gets hit, it is colored green.
+tags: "hitTolerance"
+---
+<div id="map" class="map"></div>
+<form class="form-inline">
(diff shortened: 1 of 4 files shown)


examples/hit-tolerance.html
```diff
@@ -0,0 +1,21 @@
+---
+layout: example.html
+title: Hit Tolerance
+shortdesc: Example of using the hitTolerance parameter.
+docs: >
+  Choose the hit tolerance parameter in pixels and try to click on or near the line feature. If the feature gets hit, it is colored green.
```
tschaub: Instead of leading with instructions on how to use the example, I think it makes for better documentation if you first describe what feature the example is illustrating.  For example, you could say something like:

> By default, the `map.forEachFeatureAtPixel()` function only considers features that are directly under the provided pixel.  This can make it difficult to interact with features on touch devices.  To consider features within some distance of the provided pixel, use the `hitTolerance` option.  For example, `map.forEachFeatureAtPixel(pixel, callback, {hitTolerance: 3})` will call the callback with all features that are within three pixels of the provided pixel.

simonseyock: Thanks for the feedback! I used your text, i hope thats fine.

tschaub:
Thanks @simonseyok.

PR #6307, opened 2016-12-27
Dedicated modules for enums

tschaub:
This moves shared enums to dedicated modules (rather than having them tacked on to a constructor).


tschaub:
Thanks @ahocevar.  There are more enums to be moved around.  Some are only used in the file where they are defined.  I'd suggest we mark these explicitly as `@private` (and rename with `_`).  And then all shared enum should be in dedicated modules.

tschaub:
See #6307 for the rest.

PR #6308, opened 2016-12-27
Dedicated modules for shared enums

tschaub:
This brings consistency to our enums:

 - if an enum is shared between two or more files, it needs to be in a dedicated file
 - otherwise, the enum must be private

A linter rule now enforces this convention.

### DIFF
--- a/config/jsdoc/api/readme.md
+++ b/config/jsdoc/api/readme.md
@@ -34,7 +34,7 @@ Events are documented using `@fires` and `@event` annotations:
  * Constants for event names.
  * @enum {string}
  */
-ol.MapBrowserEvent.EventType = {
+ol.MapBrowserEventType = {
   /**
    * A true single click with no dragging and no double click. Note that this
    * event is delayed by 250 ms to ensure that it is not a double click.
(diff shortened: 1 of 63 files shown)


tschaub:
Ok, a bit of a breakthrough.  This now includes the code from #6308.  After running into rollup/rollup#1189 (which may be considered expected behavior), I've reworked the codemod here so our modules make it through Rollup a bit more intact.  And boom, `ADVANCED` compilation achieved.

Here is a complete, end-to-end example of how you can build an application with the `ol@beta` package, Rollup, and the Closure Compiler (in `ADVANCED` mode).  And with rollup-plugin-closure-compiler-js#4 in, the Rollup config got even simpler.

As noted in the readme, the Compiler is crazy noisy.  Unfortunately, the `extraAnnotationNames` flag is not supported for the JS version of the compiler.  So we get warnings for our `@api` annotations.  I'm sure this can be addressed too.

This is starting to feel promising to me.  I'd still like to demonstrate an end-to-end example with Browserify and webpack.  And, it would be great to deal with our defines problem.  In the meantime, I'm going to make this package opinionated about the map renderer (but we still get unnecessary layer renderer code).

PR #6309, opened 2016-12-28
Show API on examples and more partial searching

zspitzer:
Display the API being used on the examples page
Tweak searching to allow partial matching when less than 3 matches 

for example:
arc (before 1, after 8 results)
position (before 1 result, after 4 results)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -29,6 +29,11 @@
       .example:hover {
         background-color: #ddd;
       }
+      
+      .api {
+        font-size: smaller;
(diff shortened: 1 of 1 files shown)


ahocevar:
I just tried browsing examples with this branch, and I am not so sure it improves the situation. Personally I think it would be better to re-tag some of our example (e.g. not tag every example that uses an OSM base map with `osm`). What do others think?

tschaub:

Issue #6303, opened 2016-12-24
Explicitly require the renderer


tschaub:
We currently offer runtime renderer configuration with `renderer: 'canvas'` or `renderer: 'webgl'` as map options.  This was a nice idea, but it has the major drawback of not being able to explicitly depend on just one renderer.  We assumed everybody would be using Closure Compiler build flags like `ENABLE_WEBGL=false`.  Instead people use the full build - which includes far more code than anybody needs - since it is far more convenient than using the Closure Compiler.  While the WebGL renderer has been gaining functionality, it is still incomplete, and I don't think there has been a good reason to push it on all of our users for 2.5 years.

So, I think it would make sense to specify more explicitly which renderer an application depends on.  I see two alternatives for doing this.

### Alternative 1

Make it so the `renderer` option accepts a renderer constructor instead of a string.  Keep the Canvas renderer the default.

People who want to use the Canvas renderer would not change any application code.  People who want to use the WebGL renderer would do this:
```
var map = new ol.Map({
  renderer: ol.renderer.webgl.Map,
  // ...
});
```
To get a smaller build with the Closure Compiler, people using the WebGL renderer would set `ENABLE_CANVAS=false` (as they have to now).  The downside of this alternative is that all layer renderer types are still required for each layer type. 

### Alternative 2

Create a new map constructor that pulls in the WebGL renderer.  Keep the `ol.Map` constructor using the Canvas renderer.  Create new layer renderers that pull in WebGL layer rendering code.  Keep the existing layer constructors using Canvas rendering code.

People who want to use the Canvas renderer would not change any application code.  People who want to use the WebGL renderer would do this:
```
var map = new ol.WebGLMap({
  layers: [
    new ol.layer.WebGLTile({
      // ...
    })
  ]
});
```

There would be no renderer-related build flags in this alternative.  Everybody could get smaller builds just by using the correct type of map and layer.

ahocevar:
I have been thinking through this toda as well, and I think **Alternative 2** is the way to go.

probins:
I would suggest a third alternative: split map.js in two. Move the map class definition to a new file, called, say, BaseMap. Have map.js inherit from basemap, and add the defaults in createOptionsInternal(). Then Map would remain as is for bc, and people who did not want 2 renderers and all the other defaults could use BaseMap and add what they did want in their apps. BaseMap would be an addition to the api, and could have any new functions it needed. This would also enable specifying the renderer at run time:
```js
var map = new BaseMap();
[...]

PR #6312, opened 2016-12-28
(WIP) Make layers responsible for creating layer renderers

tschaub:
Instead of having the map renderers depend on all layer types and all layer renderer types, the layers can take responsibility for creating layer renderers.

This will make it easier to create small builds without having to rely on `@define`s to indicate dependencies.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,10 @@
 
 ### Next release
 
+#### Removed build flags (`@define`)
+
+The `ol.ENABLE_TILE`, `ol.ENABLE_IMAGE`, `ol.ENABLE_VECTOR`, and `ol.ENABLE_VECTOR_TILE` build flags are no longer necessary and have been removed.  If you were using these in a `define` array for a custom build, you can remove them.
+
 #### Simplified `ol.View#fit()` API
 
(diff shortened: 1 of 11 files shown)


tschaub:
There is a circular dependency here that is resolved by #6308.  The vector tile renderer depends on an enum that is a property of the vector tile layer.  And the vector tile layer depends on the vector rile renderer.  After #6308, they both can depend on the render type enum without creating a circular dependency.

tschaub:
Rebasing now that #6308 is in.

probins:
a good step forward, but it makes #6303 more difficult, doesn't it?

tschaub:
@probins - we have two significant problems.  One is that each of the map renderers depend on layer renderers for all types of layers.  The other is that the map depends on both types of map renderer.  The suggestion in #6303 would get rid of the `ol.ENABLE_WEBGL` flag, but we'd be left with the problem of depending on layer renderer types for all layer types.  That is, we'd be left with `ol.ENABLE_TILE`, `ol.ENABLE_IMAGE`, `ol.ENABLE_VECTOR`, and `ol.ENABLE_VECTOR_TILE`.

I opted to get rid of four build flags first.  That leaves us with one problem.  We can pick up the discussion on #6303 - I'm not sure what the solution will be.

PR #6313, opened 2016-12-29
Remove ol.DEBUG

tschaub:
This proposes that we get rid of the `ol.DEBUG` build flag.

Justification:
 * These build flags are hard to use outside of the Closure Compiler.  Even with the Compiler, they only work in a "global" context, so will not work with ES6 modules (or anything that wraps code in a closure before passing to the Compiler).  This means that people include a lot of unnecessary code in their applications.
 * Most of the uses are logical expressions that guard a `console.assert()` call.  This causes trouble where `console.assert` is not defined.  It could make sense to `throw` in these situations instead - the `console.assert()` call is a bit odd because it allows execution to continue so things still break in unexpected ways after the call.
 * We need to stop piling unnecessary code into the library and relying on the Compiler to remove it all for us.
 * I think we can do a lot more to improve the experience for people developing applications with OpenLayers.  This feature doesn't feel like it is providing much.

I'm open to hearing alternative perspectives.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -4,7 +4,7 @@
 
 #### Removed build flags (`@define`)
 
-The `ol.ENABLE_TILE`, `ol.ENABLE_IMAGE`, `ol.ENABLE_VECTOR`, and `ol.ENABLE_VECTOR_TILE` build flags are no longer necessary and have been removed.  If you were using these in a `define` array for a custom build, you can remove them.
+The `ol.DEBUG`, `ol.ENABLE_TILE`, `ol.ENABLE_IMAGE`, `ol.ENABLE_VECTOR`, and `ol.ENABLE_VECTOR_TILE` build flags are no longer necessary and have been removed.  If you were using these in a `define` array for a custom build, you can remove them.
 
 #### Simplified `ol.View#fit()` API
 
--- a/config/example.json
+++ b/config/example.json
@@ -23,9 +23,6 @@
       "externs/tilejson.js",
       "externs/topojson.js"
     ],
-    "define": [
-      "ol.DEBUG=false"
-    ],
     "jscomp_error": [
       "*"
     ],
--- a/config/examples-all.json
+++ b/config/examples-all.json
@@ -23,9 +23,6 @@
       "externs/tilejson.js",
       "externs/topojson.js"
     ],
-    "define": [
-      "ol.DEBUG=false"
-    ],
     "jscomp_error": [
       "*"
     ],
--- a/config/ol.json
+++ b/config/ol.json
@@ -14,9 +14,6 @@
       "externs/tilejson.js",
       "externs/topojson.js"
     ],
-    "define": [
-      "ol.DEBUG=false"
-    ],
     "jscomp_error": [
       "*"
     ],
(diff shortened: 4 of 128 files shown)


bartvde:
I'm not opposed to this change, looking forward to hear what others think as well

tschaub:
Looks like the codemod stripped some critical comments.  I'll rerun later.

ahocevar:
I also agree that we can remove the `ol.DEBUG` assertions.

tschaub:
WebGL rendering is broken in this branch.

gberaudo:
This PR should not have been merged.

Notably, it removes the `WebGL` code for detecting errors and displaying them and the code for debugging shaders. How are we going to debug `WebGL` code?

The asserts this PR removes allowed to detect many developer errors and it was the first think to try when getting strange behaviour from an application. From my experience, it was an easy help for catching some errors that will now be harder to find out, especially for people not fluent with the OL3 library.
I am also thinking about all the work done by @marcjansen to have this feature and it feels bad to me to see it removed without more discussion.

marcjansen:
> I am also thinking about all the work done by @marcjansen to have this feature and it feels bad to me to see it removed without more discussion.

Which work exactly do you mean? We still have functioning WebGL rendering (except text) even with this merged, right?

gberaudo:
@marcjansen I mean the assert documentation you worked on during[ the code sprint in Schladming in 2015](https://github.com/openlayers/openlayers/wiki/OpenLayers-3-Austria-Code-Sprint-Spring-2015).

```
Ideas from Bart:
    Provide messages for all goog.asserts.assert statements
```

marcjansen:
Ahh, OK, now I see what you mean. Yeah, these could also be the text of exceptions we could throw.

tschaub:
> This PR should not have been merged.

I think there is a stronger argument to be made that the experimental WebGL rendering code should not be something we foist on users by default.

webgl

> Notably, it removes the WebGL code for detecting errors and displaying them and the code for debugging shaders. How are we going to debug WebGL code?

I sympathize with the need to debug the WebGL rendering code.  See #6373 for a proposal to add a flag specific to debugging WebGL shader sources.

That said, we can't keep deprioritizing user experience for the sake of having multiple renderers.

gberaudo:
Thank you @tschaub.

There are also the errors in the WebGL context, which are especially important. See https://github.com/openlayers/openlayers/pull/6313/files#diff-56c9bfd95972e67312b8e0332a946eea. It is a great benefit to catch errors.

More generally it would be helpful to restore all possible error catching in the WebGL renderer: introducing incorrect data in the GPU can create hard to spot issues. It is much easier to catch the errors early, in the JS code.

Ideally, I think the whole PR should be reverted in favor of a jscodeshift transform removing all code like:
```
ol.DEBUG && ...
```
and 
```
if (ol.DEBUG) {
   ...
}
```
I can implement such transform if you want.

ahocevar:
@gberaudo, instead of reverting, wouldn't it be nicer to just add those WebGL context debug assertions back with  `if (ol.DEBUG_WEBGL) {/*...*/}` when #6373 is merged?

gberaudo:
@ahocevar, assertions where present in many other places throughout the library.
If they are all restored, it may be easier using some sort of revert than to manually add them back one after the other.

It is fine to me to have a specific `ol.DEBUG_WEBGL` define for WebGL code. However, if we introduce a jscodeshift transform, will it be useful to have this specific define?

marcjansen:
Do we foist webgl at all?

PR #6314, opened 2016-12-29
WebGL conditional compilation

ahocevar:
This pull request makes it so mainstream minifiers can strip out code based on the `ol.ENABLE_WEBGL` define. This should move us much closer to the goals described in #6303.

### DIFF
--- a/src/ol/render/webgl.js
+++ b/src/ol/render/webgl.js
@@ -1,68 +1,72 @@
 goog.provide('ol.render.webgl');
 
-/**
- * @const
- * @type {ol.Color}
- */
-ol.render.webgl.defaultFillStyle = [0.0, 0.0, 0.0, 1.0];
+if (ol.ENABLE_WEBGL) {
 
-/**
- * @const
- * @type {string}
- */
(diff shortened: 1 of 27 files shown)


probins:
would you do the same for canvas? I'm not a fan of this approach, as it adds code which isn't really necessary. ISTM better to get rid of the define switches, which don't fit into ES6's static dependency structure. Compile-time switches also don't help run-time decisions such as 'if device supports webgl, use it else use canvas' - you do not know this at compile/build-time. So ISTM these decisions should be made in apps not in the library, and each renderer should be independently loadable.

tschaub:
@ahocevar and I have discussed this as a pragmatic, stop-gap step toward making smaller builds of the library easy to generate.  We agree we don't want to keep relying on `@define` - but for now, a single replacement with `ol.ENABLE_WEBGL=false` can be used to remove a significant chunk of unused code.

#6303 has proposals for making renderer dependencies explicit.

probins:
> a pragmatic, stop-gap measure

ok, that makes sense to me too

tschaub:
Looks good @ahocevar.  Please merge.

PR #6319, opened 2017-01-02
Remove the '3' from OpenLayers

ahocevar:
This pull request removes the "3" from OpenLayers, so we can transition to semantic versioning if we want to.


marcjansen:
+1 for semver.

probins:
which also means anyone with a clone needs to update their remote

ahocevar:
> which also means anyone with a clone needs to update their remote

GitHub has a redirect, so `ol3` will continue to work.

tschaub:
Looks good please merge.

Issue #6315, opened 2016-12-30
Min zoom ignored for pinchZoom


mstrop:
I have my view defined as

new ol.View(
{
        center: ol.proj.transform(MapAPI.defCenter, 'EPSG:4326', 'EPSG:3857'),
        zoom: MapAPI.defZoom,
        minZoom: 12,
        maxZoom: 18,
        extent: MapAPI.extent
});

Using zoom slider or zoom +/- controls works fine (map is displayed only between the defined zoom boundaries), but when I use my fingers to change the zoom, I can go beyond the defined min zoom - handle bar of zoom slider is moved outside the slider and map is zoomed out. See the attached pictures for the details. I use OL3 version 3.20.1.

![screenshot_20161230-102709](https://cloud.githubusercontent.com/assets/12169104/21562683/152ee112-ce7b-11e6-9bf8-164e3544ed43.png)
![screenshot_20161230-102725](https://cloud.githubusercontent.com/assets/12169104/21562684/15341b28-ce7b-11e6-8cf6-a7782a0f6eea.png)

bartvde:
I can see in the code that it does this on purpose apparently, but I don't know the rationale behind it to be honest:

https://github.com/openlayers/ol3/blob/master/src/ol/interaction/pinchzoom.js#L101:L105

bartvde:
See also this change: https://github.com/openlayers/ol3/commit/8d24311bc83adee9f0be098c9013135eb141f0b9

ping @fredj

bartvde:
Looks like I misread the code, that is only for intermediate pinches, not the final one

tschaub:
I assume this came with #6224.  I've proposed #6324 as a fix.

tschaub:
Partial fix applied.  See #6330 for the rest.

fredj:
fixed with #6324 and #6330

PR #6324, opened 2017-01-04
Ensure resolution stays in range when pinching

tschaub:
This makes it so the view returns to the configured resolution range after pinching.

Partial fix for #6315.


tschaub:
There are still cases where the map "sticks" beyond the min/max zoom.  See #6330 for detail.

PR #6326, opened 2017-01-04
Add lineDashOffset to stroke style

### LINKED ISSUE #5866: Is it possible to add lineDashOffset  to the line style?

tchandelle:
Fix #5866


test/spec/ol/renderer/canvas/replay.test.js
```diff
@@ -140,21 +141,33 @@ describe('ol.render.canvas.ReplayGroup', function() {
...
+    it('applies the pixelRatio to the linedash array and offset', function() {
+      var lineDash, lineDashCount = 0,
+          lineDashOffset, lineDashOffsetCount = 0;
 
       context.setLineDash = function(lineDash_) {
         lineDashCount++;
         lineDash = lineDash_.slice();
       };
 
+      Object.defineProperty(context, 'lineDashOffset', {
```
ahocevar: Please do not use `Object.defineProperty()` in tests, because we need to run the tests in such old browsers as IE9.

test_rendering/spec/ol/style/linestring.test.js
```diff
@@ -91,6 +91,23 @@ describe('ol.rendering.style.LineString', function() {
         })
       }));
       vectorSource.addFeature(feature);
+
+      feature = new ol.Feature({
```
ahocevar: I was going to say that this new feature should be tested in a new set of tests, to avoid unnoticed rendering regressions with existing features. But then I saw that this whole set of tests is written in a way that all style options are shown in a single reference image. So it's not your fault that things are as they are, and your changes here are ok.

test/spec/ol/renderer/canvas/replay.test.js
tchandelle: `Object.defineProperty()` works on IE9, but with some limitations that are not relevant here.

Do you suggest another way to test that ?

By the way, there are several other tests that fail on IE9, while this one works.

ahocevar: Ok, thanks for checking. However, for consistency with  `setLineDash` above, you could have done the exact same thing. Makes it harder to read if the same thing is achieved in two different ways.

tchandelle: But `CanvasRenderingContext2D.setLineDash` is a method and `CanvasRenderingContext2D.lineDashOffset` is a property. So I can't really use the same way for both.

Do I just remove it ?

ahocevar: Ah, thanks for the explanation. I had not thought about that! In this case your approach makes perfect sense.

PR #6328, opened 2017-01-04
ol.geom.polygon documentation update on the coordinates format

quentin-ol:
Should we refer to the GeoJSON specifications (https://tools.ietf.org/html/rfc7946#section-3.1.2) ?
I'm tempted to do so but I'm unsure about some requirements of the specifications, _i.e._:
```
- A linear ring MUST follow the right-hand rule with respect to the
      area it bounds, i.e., exterior rings are counterclockwise, and
      holes are clockwise.
- For Polygons with more than one of these [linear rings], the first MUST be
      the exterior ring, and any others MUST be interior rings.  The
      exterior ring bounds the surface, and the interior rings (if
      present) bound holes within the surface.
```

### DIFF
--- a/src/ol/geom/polygon.js
+++ b/src/ol/geom/polygon.js
@@ -26,7 +26,8 @@ goog.require('ol.math');
  *
  * @constructor
  * @extends {ol.geom.SimpleGeometry}
- * @param {Array.<Array.<ol.Coordinate>>} coordinates Coordinates.
+ * @param {Array.<Array.<ol.Coordinate>>} coordinates Coordinates of the
+ *     vertices of the polygon where the first coordinate and the last are equivalent.
  * @param {ol.geom.GeometryLayout=} opt_layout Layout.
  * @api stable
  */
--- a/src/ol/geom/polygon.js
+++ b/src/ol/geom/polygon.js
@@ -26,8 +26,12 @@ goog.require('ol.math');
  *
  * @constructor
  * @extends {ol.geom.SimpleGeometry}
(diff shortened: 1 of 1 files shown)


ahocevar:
The right-hand rule is currently not a requirement for our rings. The outer-ring-first rule is.

quentin-ol:
@ahocevar Ok, I will update the documentation accordingly.
One last question: would it make sense that the documentation states all of the GeoJSON requirements, even though Openlayers does not enforce them all?

quentin-ol:
Follow-up / comment on the last question:
the `getCoordinates` documentation states that it returns a GeoJSON coordinate array.
It also states that `coordinate orientation will depend on how the geometry was constructed.` which could be in contradiction with the GeoJSON specifications.

ahocevar:
I'd say the pull request is good as-is, now that the requirements are listed. Linking the GeoJSON spec might add confusion.

ahocevar:
Thanks @quentin-ol for taking the time to improve the documentation!

quentin-ol:
Do you want me to add a brief about the outer-ring-first rule ?

ahocevar:
> Do you want me to add a brief about the outer-ring-first rule ?

Yes, please!

quentin-ol:
@ahocevar @bartvde 
Documentation update: attempt to explain the linear rings and their rules. It's a bit verbose but this is my best attempt (yet) at explaining without referring to the GeoJSON specifications.

ahocevar:
Thanks @quentin-ol, looks great!

PR #6329, opened 2017-01-04
Fix WMTS theme location for sea-levels layer

### LINKED ISSUE #6245: wmts-dimensions example broken
http://openlayers.org/en/latest/examples/wmts-dimensions.html

The wmts server at scalgo.com respond with a "Unknown theme." message.

thomasmoelhave:
Fixes #6245.
The layer was accessed through an incorrect theme which is no longer accepted by the SCALGO tile servers.


bartvde:
thanks @thomasmoelhave

thomasmoelhave:
No problem, sorry about breaking the example.

PR #6332, opened 2017-01-06
Constrain the center in zoomByDelta

### LINKED ISSUE #5824: Using zoom interactions should constrain center
Create a view with `extent: [0, 0, 0, 0]`.

Use the zoom interaction (mouse wheel or double click). 

The center is now outside the extent.

tchandelle:
Fix #5824


ahocevar:
I'm not sure this fixes the problem. Here is what I tried
* Modified the `wms-tiled` example by setting the layer's extent also as view extent
* Zoomed out so I could see Central Europe
* Wheel-zoomed in
* Observed the center to be outside the view extent, like it is in master

tchandelle:
@ahocevar are you using a trackpad?

ahocevar:
Yes I am.

tchandelle:
Can you test with this last commit? I can't test it myself.

ahocevar:
Thanks @tchandelle, it works now. Great work as always!

PR #6339, opened 2017-01-10
Code cleanup

fredj:


ahocevar:
Can you explain what the `analyzerChecks` option does and why it makes sense to remove it?

fredj:
`analyzerChecks` checks that private properties and functions are not unused and checks the return type of functions (see https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DiagnosticGroups.java#L609-L625)

ahocevar:
Thanks for the explanation @fredj.

Issue #6343, opened 2017-01-10
WMS GetMap request width/height rounding errors (with Math.ceil)


giohappy:
Recently QGIS Server has introdeced a strict compliancy to OGC WMS 1.3 about WIDTH/HEIGHT interpretation with GetMap requests. If sizes are not consistent with BBOX the image is distorted according to the WIDTH/HEIGHT ratio.

I noticed this because OL [introduces roundings](https://github.com/openlayers/openlayers/blob/master/src/ol/source/imagewms.js#L227) when calculating image width and height. I know that some rounding is necessary but I wonder why Math.ceil() is being used instead of Math.round(). The latter would introduce less rounding errors, doesn't it?

As an example, I had some case where imageWidth was 406.0000009. Math.ceil will round it to 407 causing the higher distortion in the requested image.

I suggest to change the rounding to Math.round(). What's your opinion

giohappy:
fixed in #d0e97df34

PR #6345, opened 2017-01-10
Mitigate rounding errors in GetMap width/height calculation

giohappy:
This patch wants to mitigate widht/height rounding errors that affect GetMap requests as described in issue #6343

### DIFF
--- a/src/ol/source/imagewms.js
+++ b/src/ol/source/imagewms.js
@@ -194,8 +194,16 @@ ol.source.ImageWMS.prototype.getImageInternal = function(extent, resolution, pix
   var centerY = (extent[1] + extent[3]) / 2;
 
   var imageResolution = resolution / pixelRatio;
-  var imageWidth = ol.extent.getWidth(extent) / imageResolution;
-  var imageHeight = ol.extent.getHeight(extent) / imageResolution;
+  var imageWidth = Math.ceil(ol.extent.getWidth(extent) / imageResolution);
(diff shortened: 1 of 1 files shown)


giohappy:
Right @ahocevar I've pushed a new patch where Math.ceil is still used and the BBOX is calculated on the base of the round sizes

ahocevar:
Looks like you'll need to update some tests to reflect the bbox changes that this modification results in.

giohappy:
Yes I see...

ahocevar:
Hm. Unfortunately there was a reason for this test. See #4334. The changes suggested here break this indeed.

I'm tempted to accept the first commit of this pull request (the one that uses `Math.round`). It's a pragmatic solution with only a minor flaw. Can you please update your pull request to remove the 2nd commit? I want to play around with it and see if the 1 pixel gap with `ratio: 1` is a problem at all.

giohappy:
@ahocevar honestly I don't like very much the first solution because it works fine for small roundings but in the worst case, e.g., 100.5 being rounded to 101, a relevant distortion happens even with Math.round if the WMS server doesn't accomodate the different BBOX and H/W ratios (and it shouldn't do it according to WMS 1.3 spec).
Probably we should find a solution that fits both issues...

giohappy:
this new commit accomodate both needs. Maybe it could be more elegant but it does the work.

bjornharrtell:
Thanks for this @giohappy, I think I've seen issues that was caused by this before but did not connect the dots as to what was going on.

giohappy:
@ahocevar squash done

PR #6346, opened 2017-01-10
Ensure view.getZoom() always returns an integer

probins:
Hard to reproduce these fp arithmetic issues, but if you go to for example http://openlayers.org/en/master/examples/simple.html and keep zooming in and out, `map.getView().getZoom()` will sooner or later return something like `9.000000000000002`. I'm using zoom level as an array index, which fails when this happens.


tchandelle:
Why don't you do `Math.round(map.getView().getZoom())` in your code instead ?

probins:
ok, I was forgetting about partial zooms. Will close and continue to round in app.

PR #6356, opened 2017-01-11
Bring custom builds tutorial up to date

### LINKED ISSUE #6301: Chrome mobile, custom build, pan zoom broken
Google chrome released a new mobile version. In it, my custom build did not work, and it worked in FF. Pan and zoom were broken.

The solution to my problem was to add `externs/closure-compiler.js` to the `compile.externs` tag in the custom json file. I don't know if this is expected behaviour, but I'm submiting an issue because in the previous versions it worked without `externs/closure-compiler.js`.

Regards

probins:
I realised when looking at the tutorial rgt https://github.com/openlayers/openlayers/issues/6301#issuecomment-271850640 that it's outdated. This PR brings it up to date (hope I've got everything), and also includes `externs/closure-compiler.js` as a requirement in config files.

Fixes #6301

### DIFF
--- a/doc/tutorials/custom-builds.md
+++ b/doc/tutorials/custom-builds.md
@@ -147,14 +147,17 @@ As a test, you can use the following HTML file to verify that your custom build 
 
 ### `define`'s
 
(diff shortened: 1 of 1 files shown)


fredj:
Thanks @probins

PR #6360, opened 2017-01-12
Remove ol.array.flatten function

fredj:
`ol.array.flatten` was only used in `ol.format.EsriJSON`, the function is replaced by `ol.geom.flat.deflate.coordinates`.


fredj:
Thanks for the review

PR #6361, opened 2017-01-12
Remove "margin" at the bottom of the canvas

### LINKED ISSUE #1963: Map canvas exceed its height when its container is resized
I have a application that dislpays two div :one header of fixed height (50 px) and another div that take all remainning height and display a map.
When the window browser is resized, the map which has height of 100% of its parent does not resize correctly. Its height become greter than its parent and a scrollbar appears.

This fiddle describes the problem better :
  http://jsfiddle.net/LrrxM/

At the initial display, the height is already not good : the background color (yellow) of the map div is visible but it should not. The bottom of map should stop to 5px of bottom of window.

<bountysource-plugin>

---

Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/1532679-map-canvas-exceed-its-height-when-its-container-is-resized?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tchandelle:
Fix #1963

### DIFF
--- a/src/ol/renderer/canvas/map.js
+++ b/src/ol/renderer/canvas/map.js
@@ -41,6 +41,7 @@ ol.renderer.canvas.Map = function(container, map) {
 
   this.canvas_.style.width = '100%';
   this.canvas_.style.height = '100%';
+  this.canvas_.style.display = 'block';
   this.canvas_.className = ol.css.CLASS_UNSELECTABLE;
   container.insertBefore(this.canvas_, container.childNodes[0] || null);
 
--- a/src/ol/renderer/webgl/map.js
+++ b/src/ol/renderer/webgl/map.js
@@ -40,6 +40,7 @@ if (ol.ENABLE_WEBGL) {
         (document.createElement('CANVAS'));
     this.canvas_.style.width = '100%';
     this.canvas_.style.height = '100%';
+    this.canvas_.style.display = 'block';
     this.canvas_.className = ol.css.CLASS_UNSELECTABLE;
     container.insertBefore(this.canvas_, container.childNodes[0] || null);
 
--- a/test/spec/ol/map.test.js
+++ b/test/spec/ol/map.test.js
@@ -355,38 +355,6 @@ describe('ol.Map', function() {
 
   });
 
-  describe('#updateSize', function() {
-    var map, target;
-
-    beforeEach(function(done) {
-      target = document.createElement('div');
-      document.body.appendChild(target);
-      map = new ol.Map({
-        controls: [],
-        target: target,
-        view: new ol.View({
-          center: [0, 0],
-          zoom: 2
-        })
-      });
-      map.once('postrender', function() {
-        done();
-      });
-    });
-
-    afterEach(function() {
-      map.setTarget(null);
-      document.body.removeChild(target);
-    });
-
-    it('should always generate the same size', function() {
-      var initialSize = map.getSize();
-      map.updateSize();
-      expect(map.getSize()).to.eql(initialSize);
-    });
-
-  });
-
   describe('create interactions', function() {
 
     var options;
--- a/test/spec/ol/map.test.js
+++ b/test/spec/ol/map.test.js
@@ -355,6 +355,38 @@ describe('ol.Map', function() {
 
   });
 
+  describe('#updateSize', function() {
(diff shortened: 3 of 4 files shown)


tchandelle:
I was not able to make a unit test that work with `make test`.
So please find below a browser-only test that works with my PR (and fail without) :

```patch
 test/index.html          |  1 +
 test/spec/ol/map.test.js | 31 +++++++++++++++++++++++++++++++
 2 files changed, 32 insertions(+)

diff --git a/test/index.html b/test/index.html
index 9e79a0b..6bd8f9e 100644
--- a/test/index.html
+++ b/test/index.html
@@ -4,6 +4,7 @@
   <title>OL Spec Runner</title>
   <meta charset="utf-8">
   <link rel="stylesheet" type="text/css" href="../node_modules/mocha/mocha.css">
+  <link rel="stylesheet" type="text/css" href="../css/ol.css">
 </head>
 
 <body>
diff --git a/test/spec/ol/map.test.js b/test/spec/ol/map.test.js
index a1b2a79..e80a57a 100644
--- a/test/spec/ol/map.test.js
+++ b/test/spec/ol/map.test.js
@@ -355,6 +355,37 @@ describe('ol.Map', function() {
 
   });
 
+  describe('#updateSize', function() {
+    var map, target;
+
+    beforeEach(function(done) {
+      target = document.createElement('div');
+      document.body.appendChild(target);
+      map = new ol.Map({
+        target: target,
+        view: new ol.View({
+          center: [0, 0],
+          zoom: 2
+        })
+      });
+      map.once('postrender', function() {
+        done();
+      });
+    });
+
+    afterEach(function() {
+      map.setTarget(null);
+      document.body.removeChild(target);
+    });
+
+    it('should always generate the same size', function() {
+      var initialSize = map.getSize();
+      map.updateSize();
+      expect(map.getSize()).to.eql(initialSize);
+    });
+
+  });
+
   describe('create interactions', function() {
 
     var options;

```

bjornharrtell:
Great stuff, @ahocevar any chance of getting this into the soon-to-be-released release?

ahocevar:
I'm not sure if this is the right way to fix it. Wouldn't it make more sense to change `src/ol/renderer/canvas/map.js` around line 42 with a hard-coded `this.canvas_.style.margin = '0'`? It feels wrong to me to have such an important css property in an external css file. With my suggestion implemented, the test above should also work in mocha.

tchandelle:
@ahocevar, it doesn't work with the css margin. As you can see [here](https://jsfiddle.net/5npstu7t/), there is a space between the bottom of the div and the bottom of the canvas. And no matter where you set a margin or a padding, it doesn't move. The only way I found is by setting a vertical alignment. 

Anyway, I declared it in the JS like you suggested.

ahocevar:
@tchandelle Sorry, I meant `vertical-align`, not margin. I'm wondering why both `'top'` (now) and `'bottom'` (previously) did the trick. Any explanation for that?

fredj:
canvas element are inline elements and div are block elements.

an other solution is to set `display: block;`to canvas see: https://jsfiddle.net/5npstu7t/1/

tchandelle:
Thanks for the suggestion, `display: block` seems to work better. `vertical-align` sometime leaves a small 0.5px gap that doesn't seem to cause any problems.

fredj:
@bjornharrtell can you test if this new version (using `display = 'block'`) works for you?

bjornharrtell:
I'll be able to test it tonight.

bjornharrtell:
I can verify that this new version works as expected, in my limited tests.

fredj:
I've moved the tests in the rendering test and it works (test pass with the changes and fail without):
https://github.com/fredj/openlayers/commit/6b7cf174cbf6568d2efd7a5e9dbdba198341df9e

ahocevar:
Makes sense @fredj. @tchandelle, can you cherry-pick fredj@6b7cf17?

fredj:
@ahocevar the commit is cherry-picked. If you agree, can you "approuve" the pull request?

PR #6365, opened 2017-01-13
Make enums for draw and modify interactions public

gberaudo:
Project code compiled together with OpenLayers code by the closure compiler relies on these enums.

### DIFF
--- a/src/ol/geolocation.js
+++ b/src/ol/geolocation.js
@@ -4,6 +4,7 @@ goog.provide('ol.Geolocation');
 
 goog.require('ol');
 goog.require('ol.Object');
+goog.require('ol.GeolocationProperty');
 goog.require('ol.events');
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -97,7 +97,6 @@
         }
       ],
       "no-constant-condition": 0,
-      "openlayers-internal/enum": 2,
```
tschaub: Instead of disabling this rule, you can follow it by giving public enums their own file.

gberaudo:
OK. @tschaub, PR updated, please review.

tschaub:
Looks like the rule still needs to be re-enabled.

gberaudo:
Sorry I overlooked it. @tsauerwein, the rule is now restored.

fredj:
@tschaub good to merge ?

PR #6370, opened 2017-01-16
Feature apidoc, fix events and observable properties

tchandelle:
`id` and `style` are not observable, and so, don't trigger `change:id` and `change:style` events, but just a `change` event


ahocevar:
Interesting. It looks like `geometry` is also not observable.

tchandelle:
`geometry` is observable, but only `geometryName`  is `'geometry'`

ahocevar:
Ah, I see now that [`setGeometryName()`](https://github.com/tchandelle/ol3/blob/deacb0aae3357157d83242661c9337c9494e7ab2/src/ol/feature.js#L270) updates the object listener for the geometry name. So the geometry is observable.

PR #6371, opened 2017-01-16
Adjust copyright to include 2017

kolosov-sergey:


bartvde:
Shouldn't we simply do ```-present``` like e.g. Facebook does in the React license?

ahocevar:
@bartvde Yes, I think that would make sense. Would be great if you could create a pull request.

bartvde:
sure will do ;-)

PR #6373, opened 2017-01-17
Use the ol.DEBUG_WEBGL flag to debug shader sources

tschaub:
This makes it so shader sources are debugabble by default.

All users should always either
1.  set `ol.ENABLE_WEBGL` to `false` to avoid including unnecessary renderer code, or
2. set `ol.DEBUG_WEBGL` to `false` to avoid including debuggable shader code if you are experimenting with the WebGL renderer in production.


gberaudo:
Thank you @tschaub, please merge.

Issue #6167, opened 2016-11-24
LogicalBinary filter should accept 2 or more conditions


bartvanandel:
Currently `ol.format.filter.LogicalBinary` (and related functions and classes such as `ol.format.filter.{And,Or,and,or}`) accept exactly 2 conditions as parameters. However, according to the OGC filter schema (http://schemas.opengis.net/filter/1.1.0/filter.xsd), it should accept 2 *or more* conditions. Excerpt from that file:

```xml
...
<xsd:complexType name="FilterType">
  ...
  <xsd:element name="And" type="ogc:BinaryLogicOpType" substitutionGroup="ogc:logicOps"/>
  <xsd:element name="Or" type="ogc:BinaryLogicOpType" substitutionGroup="ogc:logicOps"/>
  ...
</xsd:complexType>
...
<xsd:complexType name="BinaryLogicOpType">
  <xsd:complexContent>
    <xsd:extension base="ogc:LogicOpsType">
      <xsd:choice minOccurs="2" maxOccurs="unbounded">
        <xsd:element ref="ogc:comparisonOps"/>
        <xsd:element ref="ogc:spatialOps"/>
        <xsd:element ref="ogc:logicOps"/>
        <xsd:element ref="ogc:Function"/>
      </xsd:choice>
    </xsd:extension>
  </xsd:complexContent>
</xsd:complexType>
...
```

The current workaround for this problem is to combine multiple `And` (or `Or`) filters, or write the filter XML by hand, which is of course less preferred.

PR #6376, opened 2017-01-17
Make ol.format.filter.or/and accept n filter conditions

tsauerwein:
This PR makes it possible to call `ol.format.filter.or/and` with more than two filter conditions.

Dealing with variable arguments is relatively ugly with ES5, I am open for suggestions! :)

Closes https://github.com/openlayers/openlayers/issues/6167


src/ol/format/filter/logicalnary.js
```diff
@@ -0,0 +1,28 @@
+goog.provide('ol.format.filter.LogicalNary');
```
tschaub: This leaves `ol.format.filter.Logical` as a pretty useless "abstract" class - with only one implementation and nothing using the type.  What do you think about removing that?

tsauerwein: `ol.format.filter.Not` and `ol.format.filter.LogicalNary` both provide an implementation for `ol.format.filter.Logical`. But I don't have a strong opinion on this, I removed the class.

tsauerwein:
Thanks for the reviews!

Issue #6377, opened 2017-01-17
Overview map causes memory leak


vsa:
Removing a map (with an overview map control) from the document leaks memory. Here is a fiddle for reproducing the issue: http://jsfiddle.net/pefn7L5n/2/.

This looks similar to #3420 (`handleResize_` appears in the memory profiler), but is related to the overview map control (when no such control is added to the map the leak no longer occurs).

Is there some other way to explicitly destroy the map (other then call `map.setTarget(null)`)?

ahocevar:
In your fiddle, you loop through all controls and remove them from the map. Does the situation change if you don't do that?

`map.setTarget(null)` should be enough to make the map a candidate for garbage collection. Maybe you took a memory profiler before it was garbage collected? Most browsers' developer consoles have an option to manually trigger garbage collection.

vsa:
No, throwing out everything except `map.setTarget(null)` doesn't help, see http://jsfiddle.net/pefn7L5n/3/. I only added the rest because I was hoping that I overlooked something else that needs to be cleaned up (thank you for the clarification).

I'm aware of the garbage collection, but that's not the issue at hand. Do you need a more detailed explanation on how I profiled this?

ahocevar:
Ok, so the issue here is probably that the overview map's target is never set to `null`. You could manually do that with `overviewMapControl.getOverviewMap().setTarget('null')`, but we'd also accept a pull request that does this in `ol.control.OverviewMap#setMap`.

vsa:
That helps, thank you for the suggestion.

Unfortunately I won't be able to provide a pull request anytime soon because my current environment does not support `make`.

fredj:
fixed with #6379

PR #6379, opened 2017-01-18
Set the overview map target in ol.control.OverviewMap.setMap

fredj:
See #6377


fredj:
Thanks for the review

PR #6381, opened 2017-01-19
geojson-vt integration example

### LINKED ISSUE #6273: geojson-vt integration
For large datasets, Openlayer's GeoJSON performance is poor compared to geojson-vt. There should be a way take geojson-vt output an feed it into Openlayer easily. Too bad there are no working example for geojson-vt and ol3 integration. Are there any efforts on to improve big data capabilities of ol3 to handle large GeoJSONs using a librray similar to geojson-vt ?

drnextgis:
Fixes #6273.

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -5,14 +5,14 @@
     "common": false,
     "createMapboxStreetsV6Style": false,
     "d3": false,
+    "geojsonvt": false,
     "jsPDF": false,
     "jsts": false,
-    "topolis": false,
-    "toastr": false,
     "saveAs": false,
+    "toastr": false,
     "topojson": false,
-    "turf": false,
-    "geojsonvt": false
+    "topolis": false,
+    "turf": false
   },
   "rules": {
     "no-unused-vars": [2, {"varsIgnorePattern": "^map"}]
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -11,7 +11,8 @@
     "toastr": false,
     "saveAs": false,
     "topojson": false,
-    "turf": false
+    "turf": false,
+    "geojsonvt": false
   },
   "rules": {
     "no-unused-vars": [2, {"varsIgnorePattern": "^map"}]
--- a/examples/geojson-vt.js
+++ b/examples/geojson-vt.js
@@ -1,0 +1,92 @@
+// NOCOMPILE
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.format.GeoJSON');
+goog.require('ol.source.OSM');
+goog.require('ol.source.VectorTile');
+goog.require('ol.layer.Tile');
+goog.require('ol.layer.VectorTile');
+goog.require('ol.tilegrid');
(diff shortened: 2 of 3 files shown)


examples/.eslintrc
```diff
@@ -11,7 +11,8 @@
     "toastr": false,
     "saveAs": false,
     "topojson": false,
-    "turf": false
+    "turf": false,
+    "geojsonvt": false
```
tchandelle: You should probably order it alphabetically.

examples/geojson-vt.js
```diff
@@ -0,0 +1,92 @@
...
+          type: 'FeatureCollection',
+          features: data.features
+        }, replacer));
+      tile.setLoader(function() {
+        tile.setFeatures(features);
+        tile.setProjection(tilePixels);
+      });
+    },
+    tileUrlFunction: function(tileCoord) {
+      return [tileCoord[0], tileCoord[1], -tileCoord[2] - 1];
```
tchandelle: `tileUrlFunction` is supposed to return an URL in string. Why don't you do that in `tileLoadFunction` instead ?

ahocevar: At some point, we have to do a hack here anyway. The way it's done here is one. Another would be to use an arbitrary `url` instead of a `tileUrlFunction`, and get use `tile.getTileCoord()` in the `tileLoadFunction`. Yet another would be to use `{z},{x},{y}` as url and use `url.split(',').map(Number)` to get the tile coordinate in the `tileLoadFunction`. What's the least ugly hack in your opinion @tchandelle?

examples/geojson-vt.js
```diff
@@ -0,0 +1,92 @@
...
+    tileGrid: ol.tilegrid.createXYZ(),
+    tilePixelRatio: 16,
+    tileLoadFunction: function(tile, tileCoord) {
+      var format = tile.getFormat();
+      var data = tileIndex.getTile.apply(tileIndex, tileCoord);
+
+      var features = format.readFeatures(
+        JSON.stringify({
+          type: 'FeatureCollection',
+          features: data.features
```
tchandelle: `data` can be `null`.

examples/geojson-vt.js
tchandelle: > The way it's done here is one. Another would be to use an arbitrary url instead of a tileUrlFunction, and get use tile.getTileCoord() in the tileLoadFunction

It's the way I would have done it. With a comment that explains that the URL is arbitrary.

Because the way it's currently done bring some confusion. `tileUrlFunction` not returning an URL, And `tileLoadFunction` not respecting the definition of a `ol.TileLoadFunctionType`

examples/.eslintrc
drnextgis: Done.

drnextgis:
@ahocevar thanks for cooperation! Looks like CI is happy. Should I make any work on PR or it ready to be merged?

ahocevar:
@drnextgis Thanks for your effort, I'm going to merge this now. Also thanks @tchandelle for the valuable review comments.

PR #6387, opened 2017-01-20
Add projection option to ol.source.Zoomify

### LINKED ISSUE #6384: ol.source.Zoomify scaling problem
There is some problem with height and width of a Zoomify source. I've made a jsfiddle (http://jsfiddle.net/daps01h0/) to show the problem. You can see a Zoomify image surrounded by a red VectorLayer rectangle. If you change the variable "scale" (line 1) to 1/8, 1/4, 1/2, 1, 2, 4, 8, 16, ... it works perfect. With other numbers the VectorLayer is not around the Image.

ahocevar:
Fixes #6384.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -6695,6 +6695,7 @@ olx.source.CartoDBOptions.prototype.account;
  *     cacheSize: (number|undefined),
  *     crossOrigin: (null|string|undefined),
  *     logo: (string|olx.LogoOptions|undefined),
+ *     projection: (ol.ProjectionLike|undefined),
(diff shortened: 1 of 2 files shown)


externs/olx.js
```diff
@@ -6740,6 +6741,14 @@ olx.source.ZoomifyOptions.prototype.logo;
 
 
 /**
+ * Projection.
+ * @type {ol.ProjectionLike}
```
fredj: type should be `ol.ProjectionLike|undefined`

PR #6390, opened 2017-01-20
Tag deprecated functions with @deprecated

fredj:
The closure-compiler (and maybe other tools) warns if the functions are used.
The functions are also marked as deprecated in the api doc.

### DIFF
--- a/src/ol/animation.js
+++ b/src/ol/animation.js
@@ -7,11 +7,11 @@ goog.require('ol.easing');
 
 
 /**
- * Deprecated (use {@link ol.View#animate} instead).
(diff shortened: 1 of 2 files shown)


tchandelle:
Why not remove the text at the beginning of the description (*Deprecated (use ... instead)*) and move it next to the @deprecated tag. This information will be displayed in the error message by the compiler, and displayed right below the deprecated tag in the apidoc.

fredj:
apidoc output:
deprecated_jsdoc

gberaudo:
Great, LGTM.

PR #6398, opened 2017-01-23
Be more tolerant of map and position value

fredj:
The `map` property type is `ol.Map|undefined` but `null` is set when
removing the overlay from the map:
https://github.com/openlayers/openlayers/blob/v3.20.1/src/ol/map.js#L453

This raise an exception with:
```js
map.removeOverlay(overlay);
overlay.setPosition(undefined);
```

Because `map === undefined` is used in `ol.Overlay.setPosition`.

This PR suggest to allow both `null` and `undefined`


fredj:
Thanks for the review

PR #6400, opened 2017-01-24
Gml 3.2 support

wglas85:
We have tried to access the WFS features from the federal of tyrol from

https://gis.tirol.gv.at/arcgis/services/Service_Public/OGD/MapServer/WFSServer?request=GetCapabilities&service=WFS

and recognized, that the WFS layers are provided as GML version 3.2.

After dissecting the GML3 parsers of ol3, we came up with a patch for adding GML-3.2 read support in a way, that it does not seem to heavily interfere with existing code.

Please review our changes and give us feedback. This is our first contribution to openlayers, so be graceful, whatever regrets may be, we have done our best ;-)

  Best Regards, Wolfgang Glas

### DIFF
--- a/src/ol/format/gml3.js
+++ b/src/ol/format/gml3.js
@@ -120,18 +120,17 @@ ol.format.GML3.prototype.readMultiSurface_ = function(node, objectStack) {
   }
 };
 
-/**
- * @param {Object} obj An object to add to a map for all known nemspace URIs.
- * @private
- * @return {Object.<string, Object>} A map with known gml namespace URIs as keys.
- */
-ol.format.GML3.makeNamespaceMap_ = function(obj) {
 
-  var ret = {};
-  ret[ol.format.GMLBase.GMLNS] = obj;
-  ret[ol.format.GMLBase.GMLNS_3_2] = obj;
-  return ret;
-};
+/**
+ * @const
+ * @type {Array.<string>}
+ * @private
+ */
+ol.format.GML3.NAMESPACE_URIS_ = [
+  ol.format.GMLBase.GMLNS,
+  ol.format.GMLBase.GMLNS_3_2
+];
+
 
 /**
  * @param {Node} node Node.
@@ -396,11 +395,11 @@ ol.format.GML3.prototype.readFlatPosList_ = function(node, objectStack) {
  * @type {Object.<string, Object.<string, ol.XmlParser>>}
  * @private
  */
-ol.format.GML3.prototype.GEOMETRY_FLAT_COORDINATES_PARSERS_ =
-  ol.format.GML3.makeNamespaceMap_({
-    'pos': ol.xml.makeReplacer(ol.format.GML3.prototype.readFlatPos_),
-    'posList': ol.xml.makeReplacer(ol.format.GML3.prototype.readFlatPosList_)
-  });
+ol.format.GML3.prototype.GEOMETRY_FLAT_COORDINATES_PARSERS_ = ol.xml.makeStructureNS(
+    ol.format.GML3.NAMESPACE_URIS_, {
+      'pos': ol.xml.makeReplacer(ol.format.GML3.prototype.readFlatPos_),
+      'posList': ol.xml.makeReplacer(ol.format.GML3.prototype.readFlatPosList_)
+    });
 
 
 /**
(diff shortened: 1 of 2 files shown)


wglas85:
Sorry, that we bothered travis for the first part, hopefully I've fixed these issues now.

ahocevar:
If you do not have a requirement to use WFS 2.0, you can use WFS 1.1.0 and get GML 3.1, which is fully supported by OpenLayers: https://gis.tirol.gv.at/arcgis/services/Service_Public/OGD/MapServer/WFSServer?request=GetCapabilities&service=WFS&version=1.1.0

Since XML based feature formats are not really efficient for transporting vector features to a client, you might also want to look into using GeoJSON as transport format.

wglas85:
Hello Andreas,

Thanks for your advice,  I was really blind concerning the downgrade capability of out tyrolean service implementation.

Unfortunately, the guys there do not offer application/json as an output format, that's why we started with alls the GML hazzle :-/

BTW, we've improved and extended a django project for delivering WFS layers as XML and JSON, see https://github.com/wglas85/django-wfs ;-)

However, I think that my patch (apart from the newbie travis problems, sorry again) is quite consistent and cleans up other issues in the GML3 code such as using the constant ol.format.GMLBase.GMLNS instead of eternally replicating the string constant 'http://www.opengis.net/gml'.

So please consider to merge it, maybe we should state 'basic support for reading GML-3.2 features is provided' in the docs.

Best Regards, Wolfgang

ahocevar:
@bartvde is the OCG format expert, so I think he should decide.

bartvde:
I'm not sure about this whole ol.format.GML3.makeNamespaceMap_ stuff, why is it needed? KML has multiple namespaces as well for instance https://github.com/openlayers/openlayers/blob/master/src/ol/format/kml.js#L117:L122

wglas85:
Hello Bart,

My makeNamespaceMap_ function is an exact duplicate of ol.xml.makeStructureNS, of which I was unaware.

The reason for assembling the maps from namespace to handler function in this way is, that you may want to reuse constants like ol.format.GMLBase.GMLNS, which is impossible, if you directly instantiate a javascript object by the {key: value,....} syntax, because - according to my experience - there are only string constants allowed as keys, not complex expressions like ol.format.GMLBase.GMLNS.

I will refactor the patch to use ol.xml.makeStructureNS, so the GML code is in line with the coding practice of the KML module.

  Regards, Wolfgang

wglas85:
OK, I brought my gml-3.2 patch in line with the KML practices,

Regards, Wolfgang

bartvde:
this looks better @wglas85 only wondering now if there is a way to keep the GML3.2 specific stuff out of the base class?

wglas85:
Hello Bart,

  I think, that we could make the GML base class unaware of the detailed GML namespaces, if we refactor the method ol.format.GMLBase.prototype.readFeaturesInternal, currently parses FeatureCollection, featureMembers, featureMember(GML) and member(GML-3.2) tags.

  When debugging, I recognized that this method occurs twice in the call stack, because the FeatureCollection parser calls  ol.xml.pushParseAndPop, which in turn comes back to ol.format.GMLBase.prototype.readFeaturesInternal when hitting member tags.

  I've no clue, whether this call structure is needed for other variants of GML and a refactoring of this method seems to be unrelated to my GML_3_2, but maybe you find a better solution rightaway.

  Maybe the refactoring of ol.format.GMLBase.prototype.readFeaturesInternal  could be discussed in the aftermath?

  What's your opinion?

    Regards, Wolfgang

Hanmac:
how is the status of this? i might need to load WFS2.0.0/GML3.2 data with openlayers
and might not be able to fall back to WFS1.1.0

wglas85:
GML-3.2 support was deemed to be not so important, so the maintainers refused to merge my pull request, maybe you might convince them.

Hanmac:
@wglas85 got a similar problem with gml3 Data flipped coordinate  (see my pull request)

They ignored my issue Ticket first,
But my change break on the test cases

ahocevar:
@bartvde What do you think about this in its current state? To me it looks good, and it does not add complexity. But I'm lacking the in-depth insight into our XML parsers.

Issue #6402, opened 2017-01-25
ol.Map.forEachLayerAtPixel() not executing callback consistently when renderer = 'canvas'


Patchez:
`ol.Map.forEachLayerAtPixel()` is not executing callback consistently when `renderer = 'canvas'` even when the layerFilter returns true.  Changing the `renderer` to `'webgl'` resolves this issue.

Looking through the source code, we found some odd differences between the code for the canvas renderer and the webgl renderer when it came to `forEachLayerAtPixel`.

marcjansen:
Which differences?

Patchez:
We're currently working on a pull request with changes that we think resolve this.  Should have it submitted shortly.

glenselle:
Just to add some more info on this one. When clicking on the map we perform an `ol.Map.forEachLayerAtPixel()` call with the pixel location at the click event. In the filter callback, we return a boolean based on a condition we're checking. Execution of this callback is intermittent.

To reproduce the issue, simply return `true` in the `ol.Map.forEachLayerAtPixel()`filter callback every time. When the canvas renderer is used, the callback to handle a match will be intermittently called. As soon as the renderer is changed to `webgl`, the callback is called every time as would be expected since the filter is hard-coded to return `true` every time. 

Below is the code that is intermittent:

```
let raster = new ol.layer.Tile({
  title: 'OSM Street',
  source: new ol.source.OSM({
    layer: 'osm'
  })
})
let map = window.olMap = new ol.Map({
  layers: [raster],
  target: 'map',
  controls: [],
  renderer: 'canvas'
})
map.on('click', function (e) {
  console.log('CLICK')

  map.forEachLayerAtPixel(e.pixel, function (layer) {
    console.warn('CALLBACK')
  }, this, function (layer) {
    console.log('FILTER')
    return true
  })
}
```

<img width="472" alt="screen shot 2017-01-26 at 12 20 00 pm" src="https://cloud.githubusercontent.com/assets/1056932/22344602/54957fb2-e3c2-11e6-9690-e8c86419af1d.png">

glenselle:
It appears a JSfiddle doing this isn't able to repro the issue. We'll continue to delve further into what we might be doing with our map that creates the issue.

http://jsfiddle.net/8o5954k7/

Patchez:
[...]

PR #6403, opened 2017-01-25
Canvas renderer forEachLayerAtCoordinate should proxy to itself

glenselle:
The canvas renderer was calling `forEachFeatureAtCoordinate` inside the `forEachLayerAtCoordinate` method. This seems incorrect, but our knowledge of OL internals is limited. When we started to dig into the canvas render's version of `forEachLayerAtCoordinate` we noticed it appeared to be working with features and not layers. If anything, this may add clarity to why the inconsistent behavior is being seen in #6402 

BTW, from some initial testing, this change appears to resolve the inconsistent behavior seen in #6402

### DIFF
--- a/src/ol/renderer/canvas/layer.js
+++ b/src/ol/renderer/canvas/layer.js
@@ -2,7 +2,6 @@ goog.provide('ol.renderer.canvas.Layer');
 
 goog.require('ol');
 goog.require('ol.extent');
-goog.require('ol.functions');
 goog.require('ol.render.Event');
 goog.require('ol.render.EventType');
 goog.require('ol.render.canvas');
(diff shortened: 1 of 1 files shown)


ahocevar:
Can you please add a unit test that fails without and passes with your fix? Also note that your fix will break hit detection on vector layers with polygons that have no fill.

glenselle:
Once we're more clear on what exactly causes the issue, we can certainly add unit tests. My JSfiddle is not reproducing the issue yet. Does hit detection need to be added? It's interesting you mention hit detection b/c we've had issues with  detection and transparent image layers in the past. We decided to avoid transparent layers b/c it wasn't working like we expected. 

What confuses me, why is it trying to find features at a coordinate? When dealing with image/tile layers, what kind of feature is it expecting to find?

ahocevar:
For image and tile layers, the `#forEachLayerAtCoordinate()` method is overridden by `ol.renderer.canvas.IntermediataCanvas`. So no dealing with features there.

ahocevar:
Closing, see #6402.

PR #6405, opened 2017-01-25
Add an arrow shape an a test.

wglas85:
We are using an arrow shape for quite while now in order to visualize the direction of road segments.

The shape is based on the code for ol.style.RegularShape and I finally managed to set up a clean pull request for this feature including an example.

The approach of RegularShape and ArrowShape, which are base don ol.style.Image is working, but in the long run we should drop rendering to an ephemere canvas, which in turn is rendered to the map canvas as an image when drawing the feature to which the shape is associated.

I would prefer to directly draw the regular shape or arrow to the map canvas without rendering to an intermediate canvas, which would for sure incur a rewrite of parts of the ol.style package and maybe the feature rendering loop. 

TIA for looking at our contribution,

  Best Regards, Wolfgang

### DIFF
--- a/examples/arrow.html
+++ b/examples/arrow.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: Arrow Style
+shortdesc: Example of GeoJSON LineStrings with an arrow style.
+docs: >
+  Example for styling LineStrings with an arrow.
+tags: "arrow, geojson, vector, openstreetmap"
+---
+<div id="map" class="map"></div>
--- a/examples/arrow.js
+++ b/examples/arrow.js
@@ -1,0 +1,142 @@
+goog.require('ol.Map');
+goog.require('ol.View');
+goog.require('ol.control');
+goog.require('ol.format.GeoJSON');
+goog.require('ol.layer.Tile');
+goog.require('ol.layer.Vector');
+goog.require('ol.source.OSM');
+goog.require('ol.source.Vector');
+goog.require('ol.style.Stroke');
+goog.require('ol.style.ArrowShape');
+goog.require('ol.style.Style');
+
+
+var lineStyle = new ol.style.Style({
+  stroke: new ol.style.Stroke({
+    color: 'red',
+    width: 1
+  })
+});
+
+var extractRotation = function extractRotation(lineString) {
+
+  var n0,n1,nm,p0,p1;
+  var coordinates = lineString.getCoordinates();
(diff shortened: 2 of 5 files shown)


wglas85:
Please have mercy on me, I have no clue how to overcome the coverage drop, maybe someone could give me some advice?

ahocevar:
Have you seen http://openlayers.org/en/latest/examples/earthquake-custom-symbol.html? With that technique, you would be able to create arrow (and any arbitrary) shapes for easy point symbolisation in style functions.

In general, before submitting a pull request, please create a ticket for discussion, and only submit pull requests when the ticket has a "pull request accepted" tag. This procedure is documented in https://github.com/openlayers/openlayers/blob/master/CONTRIBUTING.md.

Since we're planning to transition vector shapes from `ol.style.Image` to direct canvas rendering, we're currently hesitant to accept pull requests that extend `ol.style.Image`.

However, if you're willing to submit an example pull request that uses vector context functions to render custom shapes, we'd be happy to accept it.

wglas85:
Hello Andreas,

  Thanks for your feedback, I really wanted to provide our shape, because it helped us for the past months and I agree with you, that we need a generalization of 'vector' shape styles with direct canvas rendering.

  I'd love to participate in the implementation of this feature, maybe you should set up a discussion at the right place and gather other interested contributors there?

  Regards, Wolfgang

ahocevar:
Please take a look at http://openlayers.org/en/latest/examples/earthquake-custom-symbol.html. The generalization you are talking about is already available, and you might want to use it in an example pull request. That's what I meant.

ahocevar:
The above said, thanks for offering help to transition shape rendering away from ol.style.Image to direct canvas rendering. The ticket for that has already been created: #5775.

wglas85:
AFAICS, the http://openlayers.org/en/latest/examples/earthquake-custom-symbol.html example is also using ephemeral canvases, which are in turn used as images for ol.style.Icon. All this temporary canvas instantiation stuff is slow, but I will take my time and work out an example an will join the discussion #5775.

Thanks so far for your feedback.

ahocevar:
@wglas85 You are right - it probably also does not make much sense to create an example pull request for arrows now. With whatever comes out of #5775, I'd envision a vector context for a point symbolizer to be handled by the renderer, instead of having to create an image for an `ol.style.Icon` on the application level.

PR #6412, opened 2017-01-28
Create Map OSM-Fr with Layer Switcher and button call KML

Rocca2017:
Hi all

I would create a example MAP with OSM-Fr in the web site OL3.  [http://openlayers.org/en/latest/examples/](url)

But i don't understand the manipulation.

I have .css code and .js code

Can help me please ?


ahocevar:
You need to configure the OSM source with `crossOrigin: null`, because that French OSM server does not support CORS. See https://jsfiddle.net/zfrw1cuw/.

PR #6417, opened 2017-01-30
Fix getMap width/height calculation regression

giohappy:
HEIGHT / WIDTH were not correctly rounded to integer values

### DIFF
--- a/src/ol/source/imagewms.js
+++ b/src/ol/source/imagewms.js
@@ -189,39 +189,23 @@ ol.source.ImageWMS.prototype.getImageInternal = function(extent, resolution, pix
     pixelRatio = 1;
   }
 
-  extent = extent.slice();
-  var centerX = (extent[0] + extent[2]) / 2;
-  var centerY = (extent[1] + extent[3]) / 2;
-
   var imageResolution = resolution / pixelRatio;
-  var imageWidth = Math.ceil(ol.extent.getWidth(extent) / imageResolution);
-  var imageHeight = Math.ceil(ol.extent.getHeight(extent) / imageResolution);
 
-  var halfWidth = imageWidth * imageResolution / 2;
-  var halfHeight = imageHeight * imageResolution / 2;
-
-  extent[0] = centerX - halfWidth;
-  extent[1] = centerY - halfHeight;
-  extent[2] = centerX + halfWidth;
-  extent[3] = centerY + halfHeight;
+  var center = ol.extent.getCenter(extent);
+  var viewWidth = Math.ceil(ol.extent.getWidth(extent) / imageResolution);
+  var viewHeight = Math.ceil(ol.extent.getHeight(extent) / imageResolution);
+  var viewExtent = ol.extent.getForViewAndSize(center, resolution, 0,
+      [viewWidth, viewHeight]);
+  var requestExtent = ol.extent.getForViewAndSize(center, resolution, 0,
+      [viewWidth * this.ratio_, viewHeight * this.ratio_]);
 
   var image = this.image_;
   if (image &&
       this.renderedRevision_ == this.getRevision() &&
       image.getResolution() == resolution &&
       image.getPixelRatio() == pixelRatio &&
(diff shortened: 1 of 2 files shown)


ahocevar:
Can you please describe what the problem is, and point to the change that caused the "regression"?

giohappy:
The extent and the image sizes where coherent but [this product](https://github.com/openlayers/openlayers/blob/master/src/ol/source/imagewms.js#L236) introduces decimals in width/height values

giohappy:
@ahocevar The ''returns same image for consecutive calls with same args'' test fails because [this](https://github.com/giohappy/ol3/blob/9eeedd96d595592ff9c0872ed17330b62659ee89/src/ol/source/imagewms.js#L213) introduces a spurious rounding error (10.01 -> 10.010000000000002) so the consecutive calls don't return exactly the same image.
How should it be solved?

giohappy:
@ahocevar I think we need to circumnvent the rounding error, arising with the test, but the patch needs to be applied otherwise  WMS servers will (correctly) complain for not integer WIDTH/HEIGHT values

bartvde:
@giohappy I think you  can use roughlyEqual for that e.g.:

```
 expect(destinationExtent[3]).to.roughlyEqual(8399737.889818361, 1e-8);
```

I agree that we need to enforce WIDTH and HEIGHT to be integers

giohappy:
@bartvde I also was thinking to change the test and use roughlyEqual. If it's ok for you I will proceed fixing the test.

ahocevar:
I'd like to know which commit introduced this regression. Because I seem to remember a change around here in the code, which was made for a reason.

giohappy:
@ahocevar I made the previous patch to fix the problem of having WIDTH/HEIGHT not coherent with extent requested by getMap. This caused problems because WMS 1.3 requires them to have the same ratio otherwise distortion in the image output is applied.
We had to also if susequent calls were already included in the (buffered) image to avoid making new, unuseful requests.
Both the requirements were met by the previous patch but I overlooked the fact that [this lines](https://github.com/openlayers/openlayers/blob/master/src/ol/source/imagewms.js#L236) were introducing decimal values if ratio is not integer.
The new commit fix this while keeping the previous requirements, but [this test](https://github.com/openlayers/openlayers/blob/master/test/spec/ol/source/imagewms.test.js#L161) fails becasue of a Javascript rounding error, explained above.

@bartvde your solution doesn't solve the specific test, because the test should return the same image if we hadn't rounding errors. In case the roughlyEqual should be used in code but I'm not sure it's the right approach...

giohappy:
@ahocevar commit #0bf94a35 was the previous one.

ahocevar:
The test does not fail because of a JavaScript rounding error, but because you changed the logic. I created #6427 with a different approach that does not break tests.

ahocevar:
@giohappy Can you please add tests for #6427 that ensure width and height are integers, so we don't get regressions here again? Thanks.

giohappy:
Are you sure? I tested it carefully and the problem (at elat here) is the following:
The first image is requested with extent [10.01, 20, 30.01, 40]. 
The rounding makes the final image extent [10.010000000000002, 20, 30.01, 40];
When the second image is requested (with the same original extent) the test against the already available image (this._image) fails becasue [10.01, 20, 30.01, 40] is not inside [10.010000000000002, 20, 30.01, 40]. So another image is created and consequently expect(image1).to.equal(image2) will fail.

It's not a logic problem. I've double tested the new logic against Geoserver, Mapserver and QGIS Server (WMS 1.1.1 and 1.3) and it works correctly.

ahocevar:
I did not mean the logic of calculating the request extent - that is perfectly fine. What I meant is that before this patch, the extent was adjusted for pixel resolution before the `containsExtent` check. Now it is not any more. And this inconsistency between pixel width/height and view extent causes the test to fail.

fredj:
@giohappy can we close this pull request ?

giohappy:
Fixed with PR https://github.com/openlayers/openlayers/pull/6427

PR #6427, opened 2017-01-31
Ensure WMS width and height are always integers

ahocevar:
Alternative to #6417.

### DIFF
--- a/src/ol/source/imagewms.js
+++ b/src/ol/source/imagewms.js
@@ -189,39 +189,25 @@ ol.source.ImageWMS.prototype.getImageInternal = function(extent, resolution, pix
     pixelRatio = 1;
   }
 
-  extent = extent.slice();
-  var centerX = (extent[0] + extent[2]) / 2;
-  var centerY = (extent[1] + extent[3]) / 2;
-
   var imageResolution = resolution / pixelRatio;
-  var imageWidth = Math.ceil(ol.extent.getWidth(extent) / imageResolution);
-  var imageHeight = Math.ceil(ol.extent.getHeight(extent) / imageResolution);
 
-  var halfWidth = imageWidth * imageResolution / 2;
-  var halfHeight = imageHeight * imageResolution / 2;
-
-  extent[0] = centerX - halfWidth;
-  extent[1] = centerY - halfHeight;
-  extent[2] = centerX + halfWidth;
-  extent[3] = centerY + halfHeight;
+  var center = ol.extent.getCenter(extent);
+  var viewWidth = Math.ceil(ol.extent.getWidth(extent) / imageResolution);
+  var viewHeight = Math.ceil(ol.extent.getHeight(extent) / imageResolution);
+  var viewExtent = ol.extent.getForViewAndSize(center, resolution, 0,
+      [viewWidth, viewHeight]);
+  var requestWidth = Math.ceil(this.ratio_ * ol.extent.getWidth(extent) / imageResolution);
+  var requestHeight = Math.ceil(this.ratio_ * ol.extent.getHeight(extent) / imageResolution);
+  var requestExtent = ol.extent.getForViewAndSize(center, resolution, 0,
+      [requestWidth, requestHeight]);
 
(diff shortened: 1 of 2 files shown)


giohappy:
@ahocevar your patch is a regression. After rounding the sizes aren't coherent with extent anymore. That's why I did the first patch :)

ahocevar:
@giohappy That's not correct. The way the whole calculation works, the rounding will just ensure the sizes are not something like 1200.0000000001 or 1199.9999999999.

giohappy:
Are you sure that, after rounding the height/width ratio will be the same as the extent ratio?

ahocevar:
Yes, because it is calculated from the extent.

To end this guessing game, it would be good to get better test coverage for the bbox and width/height calculation.

giohappy:
Ok, I will try to setup some tests, but I'm not sure that rounding will ensure the same original ratio: a value can go the higher integer and one to the lower one.
To be sure the bbox extent has the same ratio it must be calculated again after rounding. That was the reason I wrote the first patch.

ahocevar:
I refactored the thing a bit for readability, by using `ol.extent.getForViewAndSize()`.

giohappy:
Tomorrow I will test your patch with QGIS (which implements a strict WMS 1.3 compliancy regarding w/h and bbox ratios). 
I bet I will face the same issues that were solved by my commit, because after rounding the image height/width their ratio will be slightly different from the requestedExtent bbox ratio. 

I'll let you know...

giohappy:
@ahocevar as I suspected your patch reintroduces the same problem I solved: the rounding to the height/width sizes indouces a distorion in the outpu image.

Here it is a correct image:
correct

And here when the rounding is applied:
error

We can improve the readability but we should follow the approach of recalculating the extent AFTER rounding width and height.
My PR is not wrong (even if it could be more elegant). As I already said the test fails not for a logical problem but for a rounding error. You can verify it by yourself too...

ahocevar:
@giohappy Thanks for investigating! Can you please turn your findings into unit tests so we can ensure we won't get further regressions? I'm envisioning a test that compares bbox and width/height, ensuring the ratio is roughly the same. I'll be glad to cherry-pick into this pull request.

Regarding your snippet above: there is no need to recalculate `requestExtent` - `getForViewAndSize` ensures that all values are adjusted for the pixel resolution in the first place.

ahocevar:
Tests added. @giohappy, no need for you to add more. The tests now ensure both integer width/height and correct aspect ratio.

Anyone up for a review?

giohappy:
@ahocevar ok, I did it already in my updated patch  
https://github.com/openlayers/openlayers/pull/6417 .
Anyway it's ok, thanks

giohappy:
@ahocevar I think this patch is worth a point release, because the current version can't be used with some QGIS Server versions (>= 2.14.10 < master, where WMS compliancy regarding aspect ratio was a bit relaxed)

ahocevar:
We only do bugfix releases for regressions from the previous release. But I have already started the procedure to get the next release out of the door anyway. Will take a few more days though, because we have some things to decide on before the next release.

giohappy:
ok, thanks,

ahocevar:
@giohappy I just published ca7540fa3cce376c29741d16e813b5739ff88fbd as `v3.21.0-beta.1`.

giohappy:
Thx @ahocevar. I'll tell my collegues on the QGIS ml.

giohappy:
@ahocevar just for curiosity why didn't you do a v.3.21.0-beta.1 tag?

ahocevar:
@giohappy Because we don't do official betas. The npm package was a favor to you.

ahocevar:
@giohappy Update: I just saw that there were other beta tags already, so I added `v3.21.0-beta.1` for for you too.

giohappy:
@ahocevar thanks, that's why I asked about the tag ;)

ahocevar:
Actually I made a mistake publishing to npm. Will publish `v3.21.0-beta.2`.

PR #6434, opened 2017-02-01
Stop handling down/up sequence only when all fingers have been removed

### LINKED ISSUE #6423: pinchzoom issue
- open http://openlayers.org/en/master/examples/simple.html on a mobile device
 - pinch zoom
 - on pinch release, the intermediate tiles are displayed

tchandelle:
Fix #6423

And also fix another issue I discovered :
1. Start translating a feature with one finger. 
2. While still holding that finger, just tap with another finger.
3. On v3.20.1, Translate Interaction stop working. On master, it's more random.

### DIFF
--- a/src/ol/interaction/pointer.js
+++ b/src/ol/interaction/pointer.js
@@ -187,8 +187,8 @@ ol.interaction.Pointer.handleEvent = function(mapBrowserEvent) {
     if (mapBrowserEvent.type == ol.MapBrowserEventType.POINTERDRAG) {
       this.handleDragEvent_(mapBrowserEvent);
     } else if (mapBrowserEvent.type == ol.MapBrowserEventType.POINTERUP) {
-      this.handleUpEvent_(mapBrowserEvent);
-      this.handlingDownUpSequence = false;
+      this.handlingDownUpSequence = this.handleUpEvent_(mapBrowserEvent) &&
+          this.targetPointers.length > 0;
     }
   } else {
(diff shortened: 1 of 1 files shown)


src/ol/interaction/pointer.js
```diff
@@ -187,8 +187,8 @@ ol.interaction.Pointer.handleEvent = function(mapBrowserEvent) {
     if (mapBrowserEvent.type == ol.MapBrowserEventType.POINTERDRAG) {
       this.handleDragEvent_(mapBrowserEvent);
     } else if (mapBrowserEvent.type == ol.MapBrowserEventType.POINTERUP) {
-      this.handleUpEvent_(mapBrowserEvent);
-      this.handlingDownUpSequence = false;
+      this.handlingDownUpSequence = this.handleUpEvent_(mapBrowserEvent) &&
+          this.targetPointers.length > 0;
```
tschaub: @tchandelle can you rework this to avoid having side effects in the logical expression?  E.g.

```
var unhandledUp = this.handleUpEvent_(mapBrowserEvent);
this.handlingDownUpSequence = unhandledUp && this.targetPointers.length > 0;
```

tschaub:
This looks to me like it introduces a new pinch zoom bug.  Unless I very carefully remove both fingers at the same time, there is an unwanted pan at the end of a zoom.

tschaub:
Thanks for the suggested fix.  I incorporated this into #6445.  There may be more to do, but pinch zooming is working again now.

PR #6439, opened 2017-02-02
Improve tile render performance

### LINKED ISSUE #6415: Version 3.20.0 broke something with minZoom
Please have a look at this jsfiddle which runs with version 3.20.1 : https://jsfiddle.net/202a443s/1/

In this sample I have set "minZoom" to less than zero on ol.View :
- with version 3.20.0 and upper :  if you zoom out and if zoom goes under 0, panning the map will become slow and sluggish. The more you zoom under 0, the more the map will be slow when panning.
- with version 3.19.1 : everything is smooth for all zooms.

Please also note that once the map goes under zoom 0, it will also become sluggish when panning when zoom comes back over 0, but I could not demonstrate this with this sample. This is however the behavior wee see in our app which use a custom tile source (I can't do a Jsfiddle with that). Somehow our app cannot run smoothly at all with versions > 3.19.1 because of this.

ahocevar:
This pull request avoids oversampling when the view resolution is lower than the lowest available tile source resolution. It also avoids redraws when the view extent is still covered by the intermediate canvas's image extent.

Fixes #6415.

### DIFF
--- a/src/ol/renderer/canvas/tilelayer.js
+++ b/src/ol/renderer/canvas/tilelayer.js
@@ -30,6 +30,12 @@ ol.renderer.canvas.TileLayer = function(tileLayer) {
 
   /**
    * @private
+   * @type {number}
+   */
+  this.oversampling_;
+
+  /**
+   * @private
    * @type {ol.Extent}
    */
   this.renderedExtent_ = null;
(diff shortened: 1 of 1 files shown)


ogo-adp:
Seems like you patch does the job perfectly @ahocevar.

When negative zooming, i see oversampling going up to 4 and canvas size reduced instead of a big increase, and chrome likes that.

Thanks for the fix :)

ogo-adp:
Just a remark line 35 :

Should `this.oversampling_; ` be `this.oversampling_=null; ` ?

ahocevar:
@ogo-adp:

> Should `this.oversampling_;` be `this.oversampling_ = null;` ?

No, we do not initialise numbers, strings and booleans.

PR #6443, opened 2017-02-03
goog.require cleanup

fredj:
I've found the unused require by running eslint in the `build/package/` directory from #6302


tschaub:
Thanks @fredj.  Please merge.

It would be great to rework the shaders so we can configure the missing requires rule at the error level.

PR #6445, opened 2017-02-03
Fix pinch zooming

### LINKED ISSUE #6423: pinchzoom issue

tschaub:
This incorporates @tchandelle's fix in #6434 and adds to it to fix pinch zooming.

There is some fragile logic in here since both the pinch zoom and drag pan interactions are handling both drag and up events.  I think things could be made simpler if pointer interactions could stop propagation for these events (they currently only stop down events).  But that might be a mess, so I'm not taking it on.

Fixes #6423.

### DIFF
--- a/src/ol/interaction/dragpan.js
+++ b/src/ol/interaction/dragpan.js
@@ -112,6 +112,11 @@ ol.interaction.DragPan.handleUpEvent_ = function(mapBrowserEvent) {
     view.setHint(ol.ViewHint.INTERACTING, -1);
     return false;
   } else {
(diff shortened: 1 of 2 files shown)


wglas85:
Thanks for this fix, it mitigated a problem we had, when we put other controls like sliders  taking touch events beneath the map.

In our case the map was stuck with the INTERACTING hint residing at a value larger than 0, which inhibited further loading of all layer data.

PR #6454, opened 2017-02-06
[WIP] More rendering control over tile layers

### LINKED ISSUE #5580: Option to re-render vector tiles at zoom levels not in source in hybrid render mode
As far as I know with `renderMode` set to `hybrid` tiles will only be rendered at the zoom levels existing in the source. When zooming to a level not existing in the source rendered tiles will be rescaled.

While this can make sense as vector tiles can/should be generalized to fit the zoom level, I also think it would be useful to allow tiles from the closest zoom level to be re-rendered at zoom levels not available in the source.

### LINKED ISSUE #6453: Pixelated vector tiles at view resolutions not provided by the tile source
When zooming in on a vector layer with a limited set of available resolutions, lines and polygons get pixelated in `hybrid` and `image` render modes. When switching to `vector` render mode, the problem goes away, but performance may suffer for complex vector tile sets.

It would be nice if a source could be configured to provide native scale tiles at more resolutions than the ones that tiles are loaded for. Something like:
```js
new ol.source.VectorTile({
  tileGrid: ol.tilegrid.createXYZ({
    maxZoom: 17,
    loadZooms: [1, 3, 5, 7, 9, 11, 13, 15]
  }),
  // ...
});
```
This configuration would mean only tiles from odd zoom levels between 1 and 15 are loaded. But in `hybrid` rendering mode, tiles would be pre-rendered for all zoom levels from 0 to 17. When going beyond zoom level 18, rendering mode would switch to `vector` internally, to avoid pixelation.

### LINKED ISSUE #6510: OL 4 with Tileserver GL
I've built my own server with using Tileserver GL and OSM data. Tileserver gives me a REST address which I can use in OL configuration as a data source, but resolution on the level of buildings is not efficient. You can see on this example:
https://openmaptiles.org/docs/website/openlayers3/

In the future do you expect any support for Tileserver GL?

ahocevar:
This pull request adds a new `loadZooms` option for tile grids. With that option, it is possible to limit the zoom levels that are fetched from the source. For vector tile layers in `hybrid` or `image` rendering mode, images are pre-rendered for all zoom levels of the source. When going beyond minZoom or maxZoom, vector tiles are always rendered in `vector` mode, to avoid up- and downscaling.

Fixes #6453.
Fixes #5580.
Fixes #6510.

### DIFF
--- a/examples/mapbox-vector-tiles-advanced.js
+++ b/examples/mapbox-vector-tiles-advanced.js
@@ -18,6 +18,7 @@ var key = 'pk.eyJ1IjoiYWhvY2V2YXIiLCJhIjoiRk1kMWZaSSJ9.E5BkluenyWQMsBLsuByrmg';
 var map = new ol.Map({
   layers: [
     new ol.layer.VectorTile({
+      renderMode: 'vector',
       source: new ol.source.VectorTile({
         attributions: '© <a href="https://www.mapbox.com/map-feedback/">Mapbox</a> ' +
           '© <a href="https://www.openstreetmap.org/copyright">' +
--- a/examples/mapbox-vector-tiles-advanced.js
+++ b/examples/mapbox-vector-tiles-advanced.js
@@ -25,8 +25,7 @@ var map = new ol.Map({
           'OpenStreetMap contributors</a>',
         format: new ol.format.MVT(),
         tileGrid: ol.tilegrid.createXYZ({
-          maxZoom: 19,
-          loadZooms: [0, 4, 8, 12, 16]
+          levels: [0, 3, 6, 9, 12, 15]
         }),
         url: 'https://{a-d}.tiles.mapbox.com/v4/mapbox.mapbox-streets-v6/' +
               '{z}/{x}/{y}.vector.pbf?access_token=' + key,
--- a/examples/mapbox-vector-tiles-advanced.js
+++ b/examples/mapbox-vector-tiles-advanced.js
@@ -10,52 +10,26 @@ goog.require('ol.style.Icon');
 goog.require('ol.style.Stroke');
 goog.require('ol.style.Style');
 goog.require('ol.style.Text');
-goog.require('ol.tilegrid.TileGrid');
+goog.require('ol.tilegrid');
 
 
 var key = 'pk.eyJ1IjoiYWhvY2V2YXIiLCJhIjoiRk1kMWZaSSJ9.E5BkluenyWQMsBLsuByrmg';
 
-// For how many zoom levels do we want to use the same vector tiles?
-// 1 means "use tiles from all zoom levels". 2 means "use the same tiles for 2
-// subsequent zoom levels".
-var reuseZoomLevels = 2;
-
-// Offset of loaded tiles from web mercator zoom level 0.
-// 0 means "At map zoom level 0, use tiles from zoom level 0". 1 means "At map
-// zoom level 0, use tiles from zoom level 1".
-var zoomOffset = 1;
-
-// Calculation of tile urls
-var resolutions = [];
-for (var z = zoomOffset / reuseZoomLevels; z <= 22 / reuseZoomLevels; ++z) {
-  resolutions.push(156543.03392804097 / Math.pow(2, z * reuseZoomLevels));
-}
-function tileUrlFunction(tileCoord) {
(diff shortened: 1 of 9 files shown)


examples/mapbox-vector-tiles-advanced.js
```diff
@@ -18,48 +18,49 @@ var key = 'pk.eyJ1IjoiYWhvY2V2YXIiLCJhIjoiRk1kMWZaSSJ9.E5BkluenyWQMsBLsuByrmg';
...
-        tileGrid: new ol.tilegrid.TileGrid({
-          extent: ol.proj.get('EPSG:3857').getExtent(),
-          resolutions: resolutions
-        }),
-        tilePixelRatio: 16,
-        tileUrlFunction: tileUrlFunction
+  layers: [new ol.layer.VectorTile({
+    // crisp rendering by not using rasterized lines and polygons
+    renderMode: 'vector',
+    // no downscaling by always requestin tiles for the lower resolution
```
bartvde: requesting

src/ol/layer/vectortile.js
```diff
@@ -35,17 +35,20 @@ ol.layer.VectorTile = function(opt_options) {
...
+    this.setRenderMode(options.renderMode);
+  }
 
   /**
    * @private
-   * @type {ol.layer.VectorTileRenderType|string}
+   * @type {number}
    */
-  this.renderMode_ = options.renderMode || ol.layer.VectorTileRenderType.HYBRID;
+  this.zDirection_ = options.zDirection || 0;
```
bartvde: I guess this won't happen, but if we ever change the default not to be 0, this would have issues right, i.e. not being able to set 0? Would: ```this.zDirection_ = options.zDirection !== undefined ? options.zDirection : 0;``` be safer here?

src/ol/layer/vectortile.js
```diff
@@ -91,7 +94,15 @@ ol.layer.VectorTile.prototype.getRenderMode = function() {
  */
 ol.layer.VectorTile.prototype.getUseInterimTilesOnError = function() {
   return /** @type {boolean} */ (
-      this.get(ol.layer.VectorTile.Property_.USE_INTERIM_TILES_ON_ERROR));
+      this.get(ol.layer.TileProperty.USE_INTERIM_TILES_ON_ERROR));
+};
+
+
+/**
+ * @return {number} Z direciton.
```
bartvde: direction

ahocevar:
I need to think about this a bit more. Playing with the new settings and the `mapbox-vector-tiles-advanced.html` example, I still think that the API should be simpler.

Stay tuned, more to come here in the next couple of days.

ahocevar:
I have updated this pull request and the corresponding ticket. I think the API change suggested here gives more flexibility than a configurable `zDirection` and a `renderMode` that can be set after layer creation.

externs/olx.js
```diff
@@ -7443,6 +7444,16 @@ olx.tilegrid.TileGridOptions.prototype.extent;
 
 
 /**
+ * Zoom levels to load from a tile source. This array defines a subset of the
+ * available zoom levels. When configured, only tiles for the specified zoom
+ * levels will be requested from a tile source.
+ * @type {Array.<number>|undefined}
+ * @api
+ */
+olx.tilegrid.TileGridOptions.prototype.loadZooms;
```
tschaub: I'm curious how this interacts with the `resolutions` options.

tschaub: Also, do you think `levels` would be too generic?

ahocevar: I'm adding tests right now, then it should be clear. But a quick explanation: If you pass [8, 4, 2, 1, 0.5] as `resolutions`, then there will be 5 zoom levels. So a `loadZooms` configuration of [0, 3] would result in tiles for resolutions 8 and 1 to be loaded.

ahocevar: Not sure. I tend to like `levels` better indeed, but it might make users think it's a replacement for `resolutions`.

ahocevar: Maybe `loadLevels`?

ahocevar: @tschaub, would you like an API like outlined in https://github.com/openlayers/openlayers/issues/6453#issuecomment-280723770 better? I think I would.

ahocevar:
Replaced by #6779.

PR #6455, opened 2017-02-06
Make all @api annotations imply stability

tschaub:
This branch suggests removing the stability level from `@api` annotations - treating all `@api` annotated symbols as stable.


bartvde:
I like this simplification

PR #6460, opened 2017-02-07
Remove ol.Observable#unByKey

fredj:


changelog/upgrade-notes.md
```diff
@@ -6,6 +6,8 @@
 
 The deprecated `ol.animation` functions and `map.beforeRender()` method have been removed.  Use `view.animate()` instead.
 
+The deprecated `ol.Observable#unByKey` function has been removed.  Use `ol.Observable.unByKey` instead.
```
bartvde: This might seem a bit cryptic to some? Can we add some wording that the prototype / instance function has been removed here and that the static function should be used instead? It might not be obvious for everyone what the difference between ```ol.Observable#unByKey``` and ```ol.Observable.unByKey``` is?

fredj: You're right, changed. (and don't hesitate to correct my English ...)

changelog/upgrade-notes.md
```diff
@@ -6,6 +6,17 @@
 
 The deprecated `ol.animation` functions and `map.beforeRender()` method have been removed.  Use `view.animate()` instead.
 
+The deprecated `unByKey` function of the `ol.Observable` class has been removed.  Use the `ol.Observable.unByKey` static function instead:
```
tschaub: I think there is still room for confusion here.  The difference between a "function of a class" and a "static function" is too subtle.  I'd suggest this:

> The `unByKey()` method has been removed from `ol.Observable` instances.  Use the `ol.Observable.unByKey()` static function instead.

fredj: changed, thanks

fredj:
thanks for the review

PR #6461, opened 2017-02-07
Add button to edit examples in CodePen

tchandelle:
It's more of a personal preference, I prefer CodePen over JSFiddle. And I think it would be a nice addition to the examples to be able to choose between both, similar to the examples in the MDN API.

This is what it looks like. Fortunately, the icon was already there in the css.
image

And this is what our examples looks like in CodePen : [Simple Map](https://codepen.io/anon/pen/mRjQWP?&editors=1010), [Advanced View Positioning](https://codepen.io/anon/pen/apjQWM?&editors=1110)

### DIFF
--- a/config/examples/example.html
+++ b/config/examples/example.html
@@ -60,6 +60,7 @@
         <div id="source-controls">
           <a id="copy-button"><i class="fa fa-clipboard"></i> Copy</a>
           <a id="jsfiddle-button"><i class="fa fa-jsfiddle"></i> Edit</a>
+          <a id="codepen-button"><i class="fa fa-codepen"></i> Edit</a>
         </div>
         <form method="POST" id="jsfiddle-form" target="_blank" action="https://jsfiddle.net/api/post/library/pure/">
           <textarea class="hidden" name="js">{{ js.source }}</textarea>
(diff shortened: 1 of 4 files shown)


tschaub:
Let's pick one.  I don't think there is justification for having two options for editing examples.  Can anyone list benefits and drawbacks of using CodePen instead of JSFiddle?

tchandelle:
I prefer CodePen because I think it looks better and cleaner. The preview zone is bigger and is the right size for an OL map.
Also, it auto-refresh as you type.
And we can automatically hide the CSS tab if it's empty, which is 75% of the examples.

tschaub:
I'd be in favor of switching to CodePen.

tchandelle:
I've added a commit about the removal of JSFiddle. Let me know if we include it in this PR or not

bartvde:
I have no strong preference, so I won't block a switch to codepen if others think this is better

fredj:
No preference

ahocevar:
I like CodePen, because it allows to upload assets in pro mode. Might make sense for users who want to try more advanced features that require additional resources.

I'd like to include this pull request in the next release. Objections?

PR #6466, opened 2017-02-08
Small XML error in KML test

tchandelle:


fredj:
nice catch, thanks

tschaub:
How was this passing before?

PR #6468, opened 2017-02-08
Remove console.assert in ol.structs.PriorityQueue tests

fredj:
All `console.assert` have been removed

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,15 +89,6 @@
       "openlayers-internal"
     ],
     "rules": {
-      "no-console": [
-        2,
-        {
-          "allow": [
-            "assert",
-            "warn"
-          ]
-        }
-      ],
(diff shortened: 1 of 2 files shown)


fredj:
Added a new commit that removes the `no-console` eslint rule

ahocevar:
@fredj does this mean that eslint now does not complain about the use of `console` anymore? If so, the rule should be added back, but without the exceptions.

fredj:
@ahocevar we are extending eslint-config-openlayers and the use of `console` is not allowed:
https://github.com/openlayers/eslint-config-openlayers/blob/v7.0.0/index.js#L20

(I've just checked and an error is correctly raised)

ahocevar:
Thanks for the clarification @fredj, all is great then :smile:.

PR #6475, opened 2017-02-09
Use an url template for ol.source.Zoomify

### LINKED ISSUE #3805: Zoomify: allow file extension as a property

fredj:
replace #5567 
fix #3805

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,26 +2,9 @@
 
 ### Next release
 
-#### `ol.source.Zoomify` `url` is now a template
+#### Simpler `ol.source.Zoomify` `url` configuration
 
-The `url` for the `ol.source.Zoomify` source is now a template.  The `{x}`, `{y}`, `{z}` and `{TileGroup}` placeholders
-must be included in the `url`.
-
-If you had:
-```js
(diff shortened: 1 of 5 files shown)


fredj:
This change breaks the v3 api, it would be nice to include this in the v4 release

ahocevar:
Do we really need to make this a breaking change? I think we could keep this backwards compatible with something like this in the constructor:
```js
var url = options.url;
if (options.url.indexOf('{tileGroup}') == -1) {
  url += '{TileGroup}/{z}-{x}-{y}.jpg';
}
var urls = ol.TileUrlFunction.expandUrl(options.url)

ahocevar:
@fredj, if you agree with 25fd857830267f8b052e4a3971e7f953524b993a, please cherry-pick it and merge.

fredj:
@ahocevar thanks !

Issue #6473, opened 2017-02-09
Rendering issue for tileSize: [192, 256]


kollhof:
I probably am missing something, but it appears that there is a rendering issue when the `tileSize` for a source is taller than wide.

To reproduce the issue I have used the following:
https://jsfiddle.net/f1zvwfqo/1/

```javascript
var source = new ol.source.XYZ({
	tileSize: [192, 256]
});

var map = new ol.Map({
  layers: [
    new ol.layer.Tile({
      source: new ol.source.TileDebug({
        projection: 'EPSG:3857',
        tileGrid: source.getTileGrid()
      })
    })
  ],
  target: 'map',
  view: new ol.View({
    center:[0, 0],
    zoom: 1
  })
});
```
When the map renders, it looks as if an extra set of tiles (bottom row) is not being rendered.
The issue does not appear when the `tileSize` is wider than tall, e.g `[256, 192]`.

I have started to step through the code to better understand what is going on.
My assumption so far is that the `tileRange` is not being calculated correctly:
https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/canvas/tilelayer.js#L125

ahocevar:
Issue confirmed. Thanks for the investigation. Do you think you'd be able to create a pull request with a fix?

kollhof:
I am going to give a try.

kollhof:
After some more debugging, I believe the tile range is actually being calculated correctly. 
I have narrowed down the problem to the calculation of the canvas dimensions in: https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/canvas/tilelayer.js#L183

It is incorrectly using`tilePixelSize[0]` rather than `tilePixelSize[1]` to calculate the height.

The webgl renderer does not appear to have this problem and is using the correct `tilePixelSize[1]`:
https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/webgl/tilelayer.js#L176

PR #6483, opened 2017-02-13
Fix renderer.canvas.TileLayer to calculate correct canvas height for drawing.

kollhof:
Fixes https://github.com/openlayers/openlayers/issues/6473


tschaub:
Looks like the right fix.  It would be great to have a rendering test for this.

ahocevar:
I took the liberty to add a rendering test.

ahocevar:
@kollhof, can you please submit a [CLA](https://github.com/openlayers/cla/) so we can merge this? Thanks!

kollhof:
@ahocevar Thank you for the adding the tests, I shall read through them.
I have emailed a signed CLA.

waxenegger:
Sorry to sound a bit pushy but it would be very much appreciated if this was part of the next release. Using the TileLayer and building with the sources from the release I'm going to have to patch it in my build  for now since I have a few cut-off images in the y direction.

kollhof:
@waxenegger I am using the npm package ol@4.0.1-beta.2, it has the fix and works like a charm.

waxenegger:
@kollhof thank you, that's definitely an option.

why it didn't occur to me that there are packages out there other than the 'official'/stable release, God only knows...

PR #6486, opened 2017-02-13
Do not set center when touches count has changed

### LINKED ISSUE #6485: 4.0.0 touch zoom 'jump'
Hi,

Congrats on the 4.0.0 release! Having recently updated to 4.0.0 i've noticed that when zooming on a touch device (in this instance an iPhone 7) if you place one finger on the screen before the other and zoom, it can cause a 'jumping' effect. This is not present on v3.20.1. 

Maybe it has something to do with #6444 or #6445 as these worked on zoom logic? ~~I've also noticed that it occurs more often if you start the zoom with your fingers further apart.~~

In this screen recording, I'm trying to zoom in on "Clapham Junction" and "Wimbledon". You can see how I sometimes can't keep them centered (at one point I lose "Clapham Junction"), as it jumps out of view.

https://www.youtube.com/watch?v=s_78ksSR8ds

I'll try and put together a demo asap. Any ideas as to what might be causing this?

Thanks,
Ross

ahocevar:
Currently, the DragPan interaction always sets a center when dragging. On touch devices, when a touch is added or removed, it can drastically change the centroid of all touches. So it is better to not set a center when the number of touches differs from the one in the previous drag.

Fixes #6485.

### DIFF
--- a/src/ol/interaction/dragpan.js
+++ b/src/ol/interaction/dragpan.js
@@ -70,21 +70,23 @@ ol.interaction.DragPan.handleDragEvent_ = function(mapBrowserEvent) {
   var targetPointers = this.targetPointers;
   var centroid =
       ol.interaction.Pointer.centroid(targetPointers);
-  if (this.kinetic_) {
-    this.kinetic_.update(centroid[0], centroid[1]);
-  }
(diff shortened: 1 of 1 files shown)


tschaub:
This looks right to me.  Thanks for addressing this @ahocevar.

PR #6498, opened 2017-02-15
Travis failing on master - use Node v6.1.x

ahocevar:
Travis is currently failing. ~~It seems dependency versions are not parsed correctly by npm~~ . This pull request is mostly there to debug the issue. If it makes Travis pass, then I think it is a valid change and should be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: node_js
 
 node_js:
-  - "6"
+  - "6.1"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
 before_script:
   - "rm src/ol/renderer/webgl/*shader.js"
   - "sh -e /etc/init.d/xvfb start"
-  - "npm ls"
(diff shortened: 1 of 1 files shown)


ahocevar:
Still the same problem unfortunately:
```
npm ERR! missing: esprima@^3.1.1, required by js-yaml@3.8.1
```
This does not happen for me locally, so I am a bit lost and don't really know where to go from here. For a previous pull request, removing `npm ls` made Travis pass again, but I think `npm ls` was added for a reason and is a good sanity check.

bartvde:
This is a long shot but have you tried deleting the Travis cache?

ahocevar:
Yes, I deleted the cache. It seems npm/npm/#8241 is the issue we are having.

ahocevar:
1f2f8f0 makes Travis pass again. That change makes it so `npm ls` is still run so its output is available for debugging, but the build does not fail when it errors.

Opinions?

bartvde:
I think I had the same issue in sdk and I got rid of the npm ls call, I like your solution better
npm is just too buggy unfortunately

PR #6499, opened 2017-02-15
Be more tolerant when comparing vertices when modifying a circle

### LINKED ISSUE #6495: Circle modifying interaction doesn't work constantly
If you try this [CodePen](https://codepen.io/tchandelle/pen/rjPqBq), changing the radius multiple times, you'll notice that sometime it fails.

And here is a test that fails: tchandelle@0764be77c4544c9f181215865a3baf23cdd2201a

tchandelle:
Fix #6495

### DIFF
--- a/src/ol/interaction/modify.js
+++ b/src/ol/interaction/modify.js
@@ -550,6 +550,7 @@ ol.interaction.Modify.handleDownEvent_ = function(evt) {
     return false;
   }
   this.handlePointerAtPixel_(evt.pixel, evt.map);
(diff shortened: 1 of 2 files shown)


ahocevar:
I'm not sure this is the right fix. An `1e-9` tolerance means something completely different with different projection units. If doing such a check, it would have to be done in pixel space.

tchandelle:
@ahocevar, I wasn't confident about my correction neither.
I pushed another correction that seems easier.

PR #6504, opened 2017-02-16
Determine if we should handle the drawing while the pointer is moving

### LINKED ISSUE #6501: Problem with drawing in freehand mode when last point is close to the first point
Go to http://openlayers.org/en/latest/examples/draw-freehand.html

Draw a LineString or Polygon like this gif :

freehand

tchandelle:
Fix #6501


tschaub:
Nice solution @tchandelle.

PR #6518, opened 2017-02-20
Fix olx.style.IconOptions.prototype.src type

fredj:
The `olx.style.IconOptions` `@typedef` was already correct.


fredj:
Thanks for the review

PR #6524, opened 2017-02-21
Don't append feature prefix twice in WFS requests

Jenselme:
Some WFS servers like tinyOWS require the feature prefix to be in the feature type for the feature to be found (eg `prefix:layer`).

The problem was, the prefix was always added to the feature type which gave us a type name like `prefix:prefix:feature` or `feature:prefix:feature`. The requests were then rejected by the WFS
server.

We now check if the feature type starts with the prefix. If it does, we don't append it again. If it doesn't we do.

### DIFF
--- a/test/spec/ol/format/wfs.test.js
+++ b/test/spec/ol/format/wfs.test.js
@@ -839,6 +839,47 @@ describe('ol.format.WFS', function() {
     });
   });
 
+  describe('when writing out a Transaction request', function() {
+    var text;
+    before(function(done) {
+      afterLoadText('spec/ol/format/wfs/TransactionMulti.xml', function(xml) {
+        text = xml;
+        done();
+      });
+    });
+
+    it('do not add feature prefx twice', function() {
+      var format = new ol.format.WFS();
+      var insertFeature = new ol.Feature({
+        the_geom: new ol.geom.MultiPoint([[1, 2]]),
+        foo: 'bar',
+        nul: null
+      });
+      insertFeature.setGeometryName('the_geom');
+      var inserts = [insertFeature];
+      var updateFeature = new ol.Feature({
+        the_geom: new ol.geom.MultiPoint([[1, 2]]),
+        foo: 'bar',
+        // null value gets Property element with no Value
+        nul: null,
+        // undefined value means don't create a Property element
+        unwritten: undefined
+      });
+      updateFeature.setId('fid.42');
+      var updates = [updateFeature];
+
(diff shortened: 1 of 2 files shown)


Jenselme:
Any thoughts on this?

Jenselme:
Thanks for the review @bartvde Test added and commits squashed.

Jenselme:
Just resolved merged conflict by rebasing on top of master.

test/spec/ol/format/wfs.test.js
```diff
@@ -839,6 +839,47 @@ describe('ol.format.WFS', function() {
...
+  describe('when writing out a Transaction request', function() {
+    var text;
+    before(function(done) {
+      afterLoadText('spec/ol/format/wfs/TransactionMulti.xml', function(xml) {
+        text = xml;
+        done();
+      });
+    });
+
+    it('do not add feature prefx twice', function() {
```
cs09g: prefx -> prefix

fredj: see #6740

bartvde:
I'm seeing a difference in behaviour between 4.0.1 and 4.2.0 where featurePrefix is not specified, and wonder if it's related to this PR

```
<Transaction xmlns="http://www.opengis.net/wfs" service="WFS" version="1.1.0" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><Update typeName="feature:bar" xmlns:undefined="http://foo"><Property><Name>foo</Name><Value>baz</Value></Property><Filter xmlns="http://www.opengis.net/ogc"><FeatureId fid="foo.1"/></Filter></Update></Transaction>
```

versus: 
 ```<Transaction xmlns="http://www.opengis.net/wfs" service="WFS" version="1.1.0" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><Update typeName="feature:bar" xmlns:feature="http://foo"><Property><Name>foo</Name><Value>baz</Value></Property><Filter xmlns="http://www.opengis.net/ogc"><FeatureId fid="foo.1"/></Filter></Update></Transaction>```

bartvde:
Okay see: https://github.com/openlayers/openlayers/pull/6931 for a proposed fix

PR #6526, opened 2017-02-21
Draw interaction: add abortDrawing method and drawabort event

tchandelle:


jeremy-smith-maco:
@fredj Can you re-open this and review it please?

Issue #6521, opened 2017-02-20
ol.source.WMTS.optionsFromCapabilities looses crossOrigin and urls parameters in config parameter


SanderH:
When providing crossOrigin as an argument in the config parameter, it doesn't get returned back into the returned options.
The same applies to the urls parameter which I need to override to prevent Mixed content warnings in the browser. The GetCapabilities seem to always return http url wether calling it through http or https.

```
var parser = new ol.format.WMTSCapabilities();
fetch('https://my.server.test/wmts?&request=GetCapabilities&service=WMS').then(function(response) {
	return response.text();
}).then(function(text) {
var result = parser.read(text);
var options = ol.source.WMTS.optionsFromCapabilities(result, {
	layer: 'my_layer', matrixSet: 'EPSG:28992', format: 'image/png', crossOrigin: 'Anonymous', urls: ['https://my.server.test/wmts?']});

// workaround: crossOrigin gets lost in options, so need to set it explicitly
options.crossOrigin = 'Anonymous';
// workaround: url gets overwritten in options, so need to set it explicitly
options.urls = {'https://my.server.test/wmts?'};
```
Although the workarounds work, I think it's cleaner when passing it along with the optionsFromCapabilities call.

ahocevar:
The urls are actually set from the Capabilities response. It is the responsibility of the server to return a sensible url. Which should be a https url when the capabilities are requested through https.

Regarding the `crossOrigin` option, I agree that `optionsFromCapabilities` should accept that. Do you want to create a pull request?

SanderH:
I'll try to create a pull request for the crossOrigin shortly.

I fully agree for your comment about the urls and have already filed a request to the supplier, but who knows when they'll fix it...
For my own Geoserver instance I had the same issue that it always returned either http or https depending on a fixed configuration and decided it was better to always return https than http, but it shows that apparently not all WMS/WMTS/WFS software supports being flexible regarding to the protocol requested.

SanderH:
I seem unable to satisfy Travis...

PR #6538, opened 2017-02-25
Adding crossOrigin to optionsFromCapabilities

SanderH:
This pull request should resolve: https://github.com/openlayers/openlayers/issues/6521

Now with var for crossOrigin and single quotes

### DIFF
--- a/src/ol/source/wmts.js
+++ b/src/ol/source/wmts.js
@@ -292,6 +292,7 @@ ol.source.WMTS.prototype.updateDimensions = function(dimensions) {
  *  - style - {string} The name of the style
  *  - format - {string} Image format for the layer. Default is the first
  *       format returned in the GetCapabilities response.
+ *  - crossOrigin - {string} Cross origin.
  * @return {olx.source.WMTSOptions} WMTS source options object.
  * @api
  */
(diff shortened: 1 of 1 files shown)


src/ol/source/wmts.js
```diff
@@ -444,6 +449,7 @@ ol.source.WMTS.optionsFromCapabilities = function(wmtsCap, config) {
     tileGrid: tileGrid,
     style: style,
     dimensions: dimensions,
-    wrapX: wrapX
+    wrapX: wrapX,
+    crossOrigin: crossOrigin
```
fredj: The value should be `undefined` by default, same as the `crossOrigin` option for the WMTS layer.
The code lines 437-440 can be removed and this line should be:
```
crossOrigin: config['crossOrigin']
```

src/ol/source/wmts.js
```diff
@@ -292,6 +292,7 @@ ol.source.WMTS.prototype.updateDimensions = function(dimensions) {
  *  - style - {string} The name of the style
  *  - format - {string} Image format for the layer. Default is the first
  *       format returned in the GetCapabilities response.
+ *  - crossOrigin - {string} Cross origin.
```
fredj: The type should be the same as the `crossOrigin` option for the WMTS layer: `{string|null|undefined}`

fredj:
fixed with #6577

PR #6540, opened 2017-02-27
More precise ol.Geolocation#getAccuracyGeometry return type

fredj:


fredj:
thanks for the review

PR #6545, opened 2017-03-01
chore(package): update proj4 to version 2.4.1

fredj:
Closes #6543

https://greenkeeper.io/


fredj:
tests (probably) fails because of https://github.com/proj4js/proj4js/issues/240

PR #6547, opened 2017-03-02
Clean up left-over code from animation changes

bartvde:
It seems some code was not cleaned up when we did the changes for view animation in here


bartvde:
thanks for the review @fredj

PR #6551, opened 2017-03-02
Use the default fill and stroke color

### LINKED ISSUE #6502: RegularShape without stroke or fill color
According to the API doc, stroke `color` options has black as a default value. But it doesn't seems to work with RegularShape.
Same problems with fill color.

````js
new ol.style.Style({
  image: new ol.style.RegularShape({
    points: 4,
    radius: 20,
    stroke: new ol.style.Stroke({
      width: 2
    })
  })
});
````
throws
````
Uncaught TypeError: Cannot read property '0' of null
    at Object.ol.color.toString (ol-debug.js:12801)
    at Object.ol.color.asString (ol-debug.js:12665)
    at Object.ol.colorlike.asColorLike (ol-debug.js:12831)
    at ol.style.RegularShape.render_ (ol-debug.js:23151)
    at new ol.style.RegularShape (ol-debug.js:22941)
````

fredj:
fixes #6502


fredj:
Rendering tests added, thanks for any review

fredj:
Thanks for the review

PR #6556, opened 2017-03-04
Reading kml xunits/yunits insetPixels

### LINKED ISSUE #6550: ol.format.KML does not support insetPixel for icon hotspot
ol.format.KML does not support xunit/yunit insetPixel for changing the anchor-origin for x and/or y pixel values. See https://developers.google.com/kml/documentation/kmlreference#elements-specific-to-iconstyle

After looking into the source i think it would be fairly easy to include this into the class:
* `ol.format.KML.ICON_ANCHOR_UNITS_MAP_ ` should include 'insetPixel'
https://github.com/openlayers/openlayers/blob/master/src/ol/format/kml.js#L286
* if xunit or yunit is 'insetPixel' the icon-anchor should be adjusted accordingly
https://github.com/openlayers/openlayers/blob/master/src/ol/format/kml.js#L567

Am I missing something?

simonseyock:
Fixes #6550

The test might be a little bulky though.

### DIFF
--- a/src/ol/format/kml.js
+++ b/src/ol/format/kml.js
@@ -285,7 +285,8 @@ ol.format.KML.createStyleDefaults_ = function() {
  */
 ol.format.KML.ICON_ANCHOR_UNITS_MAP_ = {
   'fraction': ol.style.IconAnchorUnits.FRACTION,
-  'pixels': ol.style.IconAnchorUnits.PIXELS
+  'pixels': ol.style.IconAnchorUnits.PIXELS,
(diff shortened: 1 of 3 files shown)


simonseyock:
@ahocevar: Is this interesting for openlayers? It improves the support of the kml standard.

ahocevar:
I'll defer to @fredj or @bartvde, they are more familiar with the KML and XML code.

bartvde:
thanks @simonseyock

PR #6559, opened 2017-03-06
Remove unneeded type cast

fredj:


fredj:
Thanks for the review

PR #6560, opened 2017-03-06
Don’t crash if feature loader xhr response returns 500

### LINKED ISSUE #6561: ol.featureloader.loadFeaturesXhr doesn't call failure callback on 500
ol.featureloader.loadFeaturesXhr doesn't call failure callback on 500 beacuse it is not catched by onload.

it is necessary to add onerror function which calls failure callback

AlesJiranek:
xhr.onload doesn't catch 500 responses and therefore failure callback is not called if server crashes.

This PR fixes #6561

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -64,6 +64,12 @@ ol.featureloader.loadFeaturesXhr = function(url, format, success, failure) {
             failure.call(this);
           }
         }.bind(this);
+        /**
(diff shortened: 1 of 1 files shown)


bartvde:
thank you @AlesJiranek have you submitted a (I)CLA so we can merge this? https://github.com/openlayers/cla

AlesJiranek:
yes, I did

Issue #5459, opened 2016-06-10
Overlay not correctly positioned


pgiraud:
In the following example, I first add an overlay to the map (green icon). After a click on this icon, I create a new overlay that I add to the map. The second overlay doesn't take its positioning configuration into account correctly. In order to have the correct positioning I need to either click on the map after the overlay is added or call `map.renderSync` in the javascript.

http://run.plnkr.co/9CzGlBUACCrL6UDe/
http://plnkr.co/edit/3R8yGo186e3JLnAocTRs

kolosov-sergey:
Ran into the same problem. After adding overlay on the map it's positioned in 'top-left', though I provided 'center-center' in the params. After any click on the map overlay jumps to its expected position.

See the demo https://jsfiddle.net/996d97gy/1/

nagytech:
My PR resolves your issue.  Please try it out.

fredj:
should be fixed with #6567

b310L:
replace direction:'ltr' instead direction:'rtl'

PR #6567, opened 2017-03-08
Overlay visible before updating render position

### LINKED ISSUE #6566: Overlay position determined before overlay is visible in the DOM.
Under the condition where the overlay container is not initially visible (ie. the style on the `.ol-overlay-container` is `display: none`) the rendered location is incorrect.  (See also #5459)

A call to `setPosition` on an overlay initiates a rendering sequence which should determine the position of the overlay within the map frame.  However, since the overlay position is determined before the container is visible the container's height and width cannot be taken into account when evaluating the overlay's `positioning` configuration.

Accordingly, the problem originates at [this point](https://github.com/openlayers/openlayers/blob/ac84e560bc4a2b7e0e37cf306decf0e676a9cf05/src/ol/overlay.js#L486) and [this point](https://github.com/openlayers/openlayers/blob/ac84e560bc4a2b7e0e37cf306decf0e676a9cf05/src/ol/overlay.js#L510) in `updateRenderedPosition` where the `offsetHeight` and `offsetWidth` properties of the container are `0`.

nagytech:
Make the overlay visible before updating the render position
so that the `offsetWidth` and `offsetHeight` can be 
considered in evaluating the offset and positioning of the 
element

Fixes #6566

### DIFF
--- a/src/ol/overlay.js
+++ b/src/ol/overlay.js
@@ -465,7 +465,7 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
   var positioning = this.getPositioning();
 
   this.setVisible(true);
-  
+
   var offsetX = offset[0];
   var offsetY = offset[1];
   if (positioning == ol.OverlayPositioning.BOTTOM_RIGHT ||
--- a/src/ol/overlay.js
+++ b/src/ol/overlay.js
@@ -464,6 +464,8 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
 
   var positioning = this.getPositioning();
 
+  this.setVisible(true);
+  
   var offsetX = offset[0];
   var offsetY = offset[1];
   if (positioning == ol.OverlayPositioning.BOTTOM_RIGHT ||
@@ -514,8 +516,6 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
       this.rendered_.top_ = style.top = top;
     }
   }
-
-  this.setVisible(true);
 };
 
 


pgiraud:
This unfortunately doesn't seem to solve #5459

nagytech:
Looking at #5459, you're using an older version which references the height and width differently from the current build (https://github.com/openlayers/openlayers/commit/96a14e05155260127e6d0bd18b184769a231ee6e#diff-8124dda9f0608aec0133ccd3108bf18c).

So, if you're just applying the `this.setVisible` change I've made directly to the build you're using, it won't work.

Here's your plnkr from #5459 using `src/ol/overlays.js` before and after my change.  Works fine for me.

### Before my change
http://plnkr.co/edit/IVn9M7n8c83uMcVYYtt4?p=preview

### After my change
http://plnkr.co/edit/cRHNMnpOIq6VYMac2iYZ?p=preview

nagytech:
Done, Thanks.

ChibiOokami86:
Not sure if I should open up a new conversation for this. My apologies if I should have.

I have looked into the Plunkers provided by @nagytech and I see the change in behavior, but I am still experiencing situations when the overlay is not positioned correctly, initially.

As noted in [#6566](https://github.com/openlayers/openlayers/issues/6566), I believe that the `offsetHeight` and `offsetWidth` properties are 0 when the overlay is initially added to the map. I start getting a proper value for these properties once the map is clicked/moved.

Using a modified demo provided in [#5459](https://github.com/openlayers/openlayers/issues/5459) so show the behavior in both OL ver. 5.1.3, which I am currently using, and ver. 5.3.0.
JSFiddle - OL 5.1.3 [Link](https://jsfiddle.net/gw9dhors/)
JSFiddle - OL 5.3.0 [Link](https://jsfiddle.net/e9abypdq/)

The overlay appears to update upon clicking or moving the map in OL 5.1.3 but the map must be moved in OL 5.3.0. map.renderSync() could be used but on slower machines the jump can still be seen.

Please let me know if this was not the intended fix that was merged above. Thank You,

ahocevar:
This happens because of the non-standard positioning, which requires OpenLayers to calculate the dimensions of the overlay's element. A quick workaround would be to call `setPosition()` after the overlay is added to the map, instead of configuring it with a `pos`:
```js
  var marker = new ol.Overlay({
    positioning: 'center-center',
    element: elem,
    stopEvent: false
  });
  map.addOverlay(marker);
  marker.setPosition(pos);
```
If that does not work for you, please create a new issue.

ChibiOokami86:
@ahocevar 

Thank you for your response.
The workaround provided works and I no longer need to apply map.renderSync(), which was causing the overlay to appear to jump on slower machines.

PR #6569, opened 2017-03-08
Return null if the layer was not found in the WMTS capabilities

fredj:


fredj:
thanks for the review

Issue #6165, opened 2016-11-24
Handle IIP protocol 


thhomas:
As OL3 supports [Zoomify](https://openlayers.org/en/latest/examples/zoomify.html) protocol, it would be interesting to supports [IIP protocol](http://iipimage.sourceforge.net/documentation/protocol/) too as it does not require a lot of changes.
I already made a branch on my own, I am ready to make a PR.

fredj:
see #7221

PR #6582, opened 2017-03-10
add ol.source.IIP, tests and example

thhomas:
Related to #6165


ahocevar:
If this is so similar to Zoomify, wouldn't it make sense to make Zoomify more flexible to also support IIP, and maybe call it DeepZoom instead?

thhomas:
As you can read on [the IIPImage server page](http://iipimage.sourceforge.net/documentation/protocol/)  IIP, IIIFZoomify, DeepZoom seems to be different protocols (even if quite similar). 
Basically, we would just need to change the [createFromTemplate](https://github.com/thhomas/openlayers/blob/iip-source/src/ol/source/zoomify.js#L96) function to adapt to IIP url with template `&JTL={z},{tileIndex}`

ahocevar:
Cool. We'd accept a pull request that adds a `{tileIndex}` placeholder to `ol.source.Zoomify`. Then you could use a configuration like this to get images from your server:
```js
new ol.source.Zoomify({
  url: ''http://vips.vtech.fr/cgi-bin/iipsrv.fcgi?FIF=/mnt/MD1/AD00/plan_CHU-4HD-01/FOND.TIF&WID=256&JTL={z},{tileIndex}',
  // ...
});
```
If you create such a pull request, please make sure to add appropriate documentation to mention the new supported protocol and how to use it.

thhomas:
moved to #7221

Issue #6310, opened 2016-12-28
error when modifying polygon


hyperknot:
I just run into the following error while playing around, editing a simple polygon.

Unfortunately I have no idea what triggered the bug, I was just (double) clicking around for testing.

The polygon was created simply by a draw interaction.

3.20.1 / Chrome / OS X 10.11.6

```
Assertion failed: uid (%s) of value (%s) does not exist 137 Objectdepth: Array[1]feature: ol.Featuregeometry: ol.geom.Polygonindex: 4ol_uid: 137segment: Array[2]__proto__: Object
ol.structs.RBush.update @ ol-debug.js:34819
ol.interaction.Modify.handleUpEvent_ @ ol-debug.js:69494
ol.interaction.Pointer.handleEvent @ ol-debug.js:18627
ol.interaction.Modify.handleEvent @ ol-debug.js:69539
ol.Map.handleMapBrowserEvent @ ol-debug.js:41784
boundListener @ ol-debug.js:3441
ol.events.EventTarget.dispatchEvent @ ol-debug.js:3859
ol.MapBrowserEventHandler.handlePointerUp_ @ ol-debug.js:17027
boundListener @ ol-debug.js:3441
ol.events.EventTarget.dispatchEvent @ ol-debug.js:3859
ol.pointer.PointerEventHandler.fireNativeEvent @ ol-debug.js:16802
ol.pointer.NativeSource.pointerUp @ ol-debug.js:15645
ol.pointer.PointerEventHandler.eventHandler_ @ ol-debug.js:16591
boundListener @ ol-debug.js:3441

Uncaught TypeError: Cannot read property 'minX' of undefined
    at ol.structs.RBush.update (ol-debug.js:34823)
    at ol.interaction.Modify.handleUpEvent_ (ol-debug.js:69494)
    at ol.interaction.Modify.ol.interaction.Pointer.handleEvent (ol-debug.js:18627)
    at ol.interaction.Modify.handleEvent (ol-debug.js:69539)
    at ol.Map.handleMapBrowserEvent (ol-debug.js:41784)
    at ol.MapBrowserEventHandler.boundListener (ol-debug.js:3441)
    at ol.MapBrowserEventHandler.ol.events.EventTarget.dispatchEvent (ol-debug.js:3859)
    at ol.MapBrowserEventHandler.handlePointerUp_ (ol-debug.js:17027)
    at ol.pointer.PointerEventHandler.boundListener (ol-debug.js:3441)
[...]

PR #6605, opened 2017-03-20
Throws Error when modifying a geometry. Caused by not finding the item when trying to update the extent of the item when it is modified.

elimau:
There is bug occurs when modifying geometry.
It throws Error when modifying a geometry. Caused by not finding the item when trying to update the extent of the item when it is modified.
As per issue described here: https://github.com/openlayers/openlayers/issues/6310
The issue shows the problem occurring in the openlayers demo and includes screenshots which includes the stack trace of the Error.

The problem is that: 
when the modify interaction tries to update the extent of the geometry by calling the ol.structs.RBush.prototype.update() function with the 'value', there is no 'item' in 'this.items_' with that value. Hence the Error.

The Work around:
I have worked around the problem by catching the Error. The Error was converted to a  console.error, so we can still see the error message bug but the openlayers map does not crash completely.

However this is just a work around as it does not address the root cause of the problem.
It would be greatly appreciated if there is anyone with more indepth knowledge of openlayers could help identify the root cause of the issue.

Thank you.

### DIFF
--- a/openlayers.fork.iml
+++ b/openlayers.fork.iml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="WEB_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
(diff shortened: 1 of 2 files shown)


ahocevar:
#6606 provides a fix instead of this workaround.

elimau:
Thanks very much @ahocevar! Any idea when a release 4.0.2 will be available on npm?

elimau:
I've retested this problem using the master branch. The problem still occurs. The problem occurs without deleting any vertices. So I think there is a different fix still needed. @ahocevar .
Thank you.

ahocevar:
Please provide a detailed description so I can reproduce the remaining issue.

PR #6607, opened 2017-03-20
Allow an array of conditions in And/Or filter

adube:
This PR allows giving an array of conditions to the `ol.format.filter.And` and `ol.format.filter.Or` filters. Giving several separate conditions as arguments still work.

This is useful when the number of conditions is indeterminated.

### DIFF
--- a/src/ol/format/filter/and.js
+++ b/src/ol/format/filter/and.js
@@ -8,12 +8,17 @@ goog.require('ol.format.filter.LogicalNary');
  * Represents a logical `<And>` operator between two or more filter conditions.
  *
  * @constructor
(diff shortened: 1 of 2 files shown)


ahocevar:
You could achieve the same on the application level:
```js
ol.format.filter.And.apply(null, conditions);
```

adube:
Hello @ahocevar.  That's a neat trick.  The above doesn't work, thought.  It returns: `undefined`.

If, instead, the static method is used instead, then it works:

```
ol.format.filter.and.apply(null, conditions);
```

There's no need for a PR, then.  I'm closing this.  Thank you very much for the tip!

PR #6610, opened 2017-03-20
Fix event type case of MSPointerDown

### LINKED ISSUE #5478: Events not stopped when using IE10
I stumbled upon an issue using internet explorer 10 and openlayers 3.14

There seems to be a regression between ol 3.13 and 3.14 regarding event propagation in internet explorer 10.

The issues I have can be reproduced right in the openlayers examples, like these two:

http://openlayers.org/en/v3.14.0/examples/measure.html
http://openlayers.org/en/v3.14.0/examples/popup.html

In the popup example, in ie10, if you click on the popup, be it in the body or the "x" button, the event goes through and another popup will open up. A popup will also open every time you click the "+" or "-" button to zoom in or out.

The problem is also visible with the measure tool, you may start drawing a line by clicking the +, -, or the attribution buttons.

All of these issues are not occurring if you visit the[ 3.13 version](http://openlayers.org/en/v3.13.0/examples/popup.html) of the examples. The issue is still present with the [master](http://openlayers.org/en/master/examples/popup.html) hosted version of the examples.

ahocevar:
This one was hard to find, but finally I managed to fix the overlay container event handling for IE10. Thanks @jlap for the bug report and the investigation.

Fixes #5478.

### DIFF
--- a/src/ol/events/eventtype.js
+++ b/src/ol/events/eventtype.js
@@ -26,7 +26,7 @@ ol.events.EventType = {
   MOUSEOUT: 'mouseout',
   MOUSEUP: 'mouseup',
   MOUSEWHEEL: 'mousewheel',
-  MSPOINTERDOWN: 'mspointerdown',
+  MSPOINTERDOWN: 'MSPointerDown',
   RESIZE: 'resize',
   TOUCHSTART: 'touchstart',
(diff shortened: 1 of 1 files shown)


jlap:
Oh my, really wouldn't have expected that.

Thanks a lot !

PR #6523, opened 2017-02-21
Add Gml2 serializers to support WFS-T 1.0.0

Jenselme:
This PR adds `writeFeatureElement` and related serializers to `ol.format.GML2`. This will allow support for WFS-T version 1.0.0 (currently only 1.1.0 is supported witch GML3) I tested the serializers against a WFS 1.0.0 server (GeoServer) for these geometries:
- Lines
- Polygons
- Points
- MultiLines
- MultiPoints
- MultiPolygons

For these tests, in addition to these patches, I also applied a patch to use `ol.format.GML2.prototype.writeFeatureElement` in `ol.format.WFS.prototype.writeTransaction`. For those interested, the patch is [here](https://github.com/Jenselme/ol3/commit/857c6994043110c99392dc3258d1b45665c0dd22). I intend to open a PR for this as soon as this one is merged.

PR #6612, opened 2017-03-21
Add support for WFS 1.0.0 to ol.format.WFS#writeTransaction

Jenselme:
- Add a version property to olx.format.WFSWriteTransactionOptions to set
the WFS protocol version to use
- Use the specified version to use the correct schema and version in the
GML request
- Use the version to select the proper GML serializer

This is a follow up to #6523

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2285,7 +2285,8 @@ olx.format.WFSWriteGetFeatureOptions.prototype.resultType;
  *     srsName: (string|undefined),
  *     handle: (string|undefined),
  *     nativeElements: Array.<Object>,
- *     gmlOptions: (olx.format.GMLOptions|undefined)}}
+ *     gmlOptions: (olx.format.GMLOptions|undefined),
+ *     version: (string|undefined)}}
  */
 olx.format.WFSWriteTransactionOptions;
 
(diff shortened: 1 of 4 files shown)


externs/olx.js
```diff
@@ -2348,6 +2348,15 @@ olx.format.WFSWriteTransactionOptions.prototype.gmlOptions;
 
 
 /**
+ * WFS version to use for the transaction. Can be either `1.0.0` or `1.1.0`.
+ * Default is `1.1.0`.
+ * @type {string|undefined}
+ * @api
+ */
+olx.format.WFSWriteTransactionOptions.prototype.version;
```
bartvde: you need to add it to the @typedef itself as well

src/ol/format/wfs.js
```diff
@@ -851,8 +877,10 @@ ol.format.WFS.prototype.writeTransaction = function(inserts, updates, deletes,
     options) {
   var objectStack = [];
   var node = ol.xml.createElementNS(ol.format.WFS.WFSNS, 'Transaction');
+  var version = options.version ? options.version : '1.1.0';
```
bartvde: can we use ```ol.format.WFS.DEFAULT_VERSION``` instead?

Jenselme:
PR updated to take your comments into account.

bartvde:
thank you again

Jenselme:
@bartvde Thanks for the quick merge.

PR #6620, opened 2017-03-21
Fix the parsing of flat coordinates in GML2 for 3D geometies

Jenselme:
The code used to rely on the dimension. However, the dimension can only
be read from a GML3 document. This caused, for 3D geometries:
- An assertion error to be raised since the flatCoordinates list
contains more than 3 elements: "Assertion failed: flatCoordinates should
have a length of 3" The value of flatCoordinates for a 3D point was
something like `[2586394, 1232407, 0, 731, NaN, 0]`
- The value of the Z coordinates to be incorrect and always set to 0

This patch simplifies and correct the parsing of the coordinates:
- Don't parse groups of coords and the coordonates at the same time.
- Detect the dimension for the coordinates.
- If the Z coordinate exists, its value is used, otherwise, we use 0.
- Correct the presentation of test data to make it work with the new
parser.
- Add a test for a 3D point.

### DIFF
--- a/src/ol/format/gml2.js
+++ b/src/ol/format/gml2.js
@@ -59,7 +59,6 @@ ol.format.GML2.prototype.readFlatCoordinates_ = function(node, objectStack) {
   var s = ol.xml.getAllTextContent(node, false).replace(/^\s*|\s*$/g, '');
   var context = /** @type {ol.XmlNodeStackItem} */ (objectStack[0]);
   var containerSrs = context['srsName'];
-  var containerDimension = node.parentNode.getAttribute('srsDimension');
   var axisOrientation = 'enu';
   if (containerSrs) {
     var proj = ol.proj.get(containerSrs);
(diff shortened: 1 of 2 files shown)


src/ol/format/gml2.js
```diff
@@ -59,32 +59,21 @@ ol.format.GML2.prototype.readFlatCoordinates_ = function(node, objectStack) {
...
-  }
+  var coordsGroups = s.trim().split(/\s+/);
   var x, y, z;
   var flatCoordinates = [];
-  for (var i = 0, ii = coords.length; i < ii; i += dim) {
-    x = parseFloat(coords[i]);
-    y = parseFloat(coords[i + 1]);
-    z = (dim === 3) ? parseFloat(coords[i + 2]) : 0;
+  for (var i = 0, ii = coordsGroups.length; i < ii; i++) {
+    var coords = coordsGroups[i].split(/,+/);
```
tschaub: I think these regexes aren't going to cut it.  Don't we expect things like this:

```xml
<gml:coordinates>1, 2 3, 4</gml:coordinates>
```

Jenselme: @tschaub Good point. Working on it.

Jenselme:
I just updated to PR to support things like:
```xml
<gml:coordinates>1, 2 3, 4</gml:coordinates>
```

To do that, just after the regexp, I check if the "group" coords ends by a comma. If it does, I group it with the next one.

tschaub:
@bartvde - any thoughts on this?  Feels to me like the tests could cover a lot more of the variety of GML.

bartvde:
@tschaub to be honest, I've not really come across GML coordinates that are not exactly formatted according to the spec, i.e. with extra whitespace like in your example. If space is the tuple separator (like the default), it should not be used outside of separating tuples IMHO.

I think we should support alternative decimal, cs and ts attributes though, not sure if this is the case right now though.

Jenselme:
> I've not really come across GML coordinates that are not exactly formatted according to the spec, i.e. with extra whitespace like in your example. If space is the tuple separator (like the default), it should not be used outside of separating tuples IMHO.

If we don't have to take into account cases like `<gml:coordinates>1, 2 3, 4</gml:coordinates>`, this would simplify the parsing of the coordinates.

> I think we should support alternative decimal, cs and ts attributes though, not sure if this is the case right now though.

From what I read, alternative decimal don't seem supported. IMHO, this support should be part of another PR. But I can add it in this one if needed.

bartvde:
Right I agree it doesn't have to be part of this PR @Jenselme

bartvde:
@tschaub what do you think? I'm inclined to assume clean formatting in the parser? Or is this naive?

Jenselme:
@bartvde Now that I am thinking about it: the problem is that OL is supporting these "invalid" formatting. So if we remove it, we may introduce a regression for some users.

tschaub:
Think I got my example from the spec.  Perhaps Wikipedia.  I recall the spec says there is no way to validate.

bartvde:
Yes the XML schema will not help, it will still be valid if it's "malformed"

bartvde:
looks like Wikipedia has it

bartvde:
in the spec it looks to be all clean

tschaub:
Ok by me to make the parser stricter (not tolerating spaces around coordinate value separators).

Jenselme:
> Ok by me to make the parser stricter (not tolerating spaces around coordinate value separators).

Cool. Then I can simplify the parser and revert to my first solution. I should be able to do this by the end of the week.

Jenselme:
PR updated with the simplified parser.

bartvde:
thanks @Jenselme

Jenselme:
Thanks for the review and the merge.

PR #6622, opened 2017-03-22
Fix proj4 type

gberaudo:
Previous typing was confusing the compiler. It could not figure out the difference between the proj4 property and the constructor, which lead to errors such as:

- ol.proj.setProj4(proj4)
ERR! compile /home/gberaudo/dev/ngeo/src/proj/epsg21781.js:8: ERROR - actual parameter 1 of ol.proj.setProj4 does not match formal parameter
ERR! compile found   : function (new:proj4): ?
ERR! compile required: (null|proj4)
ERR! compile   ol.proj.setProj4(proj4);
ERR! compile                    ^^^^^

- ol.proj.proj4.get().defs('EPSG:21781', epsg21781def)
ERR! compile /home/gberaudo/dev/ngeo/src/proj/epsg21781.js:26: ERROR - Property defs never defined on proj4
ERR! compile   ol.proj.proj4.get().defs('EPSG:21781', epsg21781def);
ERR! compile                       ^^^^

### DIFF
--- a/externs/proj4js.js
+++ b/externs/proj4js.js
@@ -7,25 +7,31 @@
 /**
  * @constructor
  */
-var proj4 = function() {};
+var Proj4 = function() {};
 
 
 /**
- * @type {function(Array.<number>): Array.<number>}
+ * @param {ol.Coordinate} coordinates
+ * @return {ol.Coordinate}
  */
-proj4.prototype.forward;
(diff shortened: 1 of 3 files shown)


gberaudo:
Thanks @ahocevar for the quick review.

PR #6627, opened 2017-03-22
Change css selector for user-select none

### LINKED ISSUE #6618: Zoom with box also selects text in page in IE11
selection_144

bartvde:
closes #6618


bartvde:
coveralls seems down, so merging with only travis build

PR #6642, opened 2017-03-27
Put image data to correct canvas only

### LINKED ISSUE #6564: Raster calculations cause unwanted transition effects
`ol.layer.Image` with raster source shows strange transition effects when panning around and zooming in/out, see example below. Raster operations performed on pixels are very simple and basically the same as in [Sea Level example](https://openlayers.org/en/latest/examples/sea-level.html) - assigning a colour based on index value.

transitions2

I would appreciate any thoughts on what might cause these unwanted transitions. Please see [code example](https://codepen.io/anon/pen/PpbYXM) for details on implementation.

ahocevar:
Because the canvas is shared across `ol.ImageCanvas` instances, we need to make sure that the worker's result is only written to the canvas when the `ol.ImageCanvas` instance is the right one. Otherwise resolution and extent will be incorrect.

Fixes #6564.

### DIFF
--- a/src/ol/source/raster.js
+++ b/src/ol/source/raster.js
@@ -298,8 +298,9 @@ ol.source.Raster.prototype.composeFrame_ = function(frameState, callback) {
     this.dispatchEvent(new ol.source.Raster.Event(
         ol.source.Raster.EventType_.BEFOREOPERATIONS, frameState, data));
 
+    var imageCanvasUid = ol.getUid(this.renderedImageCanvas_);
     this.worker_.process(imageDatas, data,
-        this.onWorkerComplete_.bind(this, frameState, callback));
+        this.onWorkerComplete_.bind(this, frameState, imageCanvasUid, callback));
   }
 
   frameState.tileQueue.loadMoreTiles(16, 16);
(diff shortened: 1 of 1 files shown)


ahocevar:
Only partially fixes the issue. Needs more research and work.

ahocevar:
This is now ready for review.

tschaub:
Thanks for digging into this @ahocevar.

Retaining the old Canvas definitely has a performance benefit - I haven't taken the time to quantify this, but the sea level example demonstrates that we aren't able to render as frequently with this change.

Do we have a reliable way to reproduce the bug that this is addressing?  It would be unfortunate to have a significant performance drop for what may be a minor issue (I don't have enough info to say whether the performance drop is significant or the issue is minor, just saying I'm wary of the tradeoff).

tschaub:
I see https://codepen.io/anon/pen/PpbYXM and see that this is easy to reproduce on the first zoom in.  Feels like we may be able to address it by invalidating the old canvas in the right place.

ahocevar:
> Feels like we may be able to address it by invalidating the old canvas in the right place.

I've experimented with that, and could not come up with a solution that solved the problem entirely. But feel free to play with it more.

ahocevar:
@tschaub The problem here is that the `frameState` used in the `isDirty()` check is a snapshot from the time the canvas was requested. We'd need to use the current `frameState` at the time of composition, which is not available to the source.

ahocevar:
@tschaub, I updated the pull request so the canvas is now always reused.

tschaub:
Nice @ahocevar!

Looks great from the phone.  I look forward to trying it out on a full size device shortly.

tschaub:
Ok.  Looking at this more closely now.  It is intentional that when a worker finishes, we render the returned data even if there is another worker pending.  With the change in d613900838e17bcb98bcb02874b3f7809a45edf8, this is no longer the case.

As [you mention](https://github.com/openlayers/openlayers/pull/6642#issuecomment-289917774)
> The problem here is that the frameState's resolution and extent used in the isDirty() check are a snapshot from the time the canvas was requested. We'd need to use resolution and extent from the current frameState at the time of composition, which is not available to the source.

We can solve this by keeping track of the most recently requested frame state and using that in place of the dirty check (e.g. https://github.com/tschaub/openlayers/commit/32995f6f4543ef2fe845ff055196d99c9771f520).  However, this turns out not to be the problem.  The problem still exists with that commit.

I can confirm that your change fixes the issue.  But I'd like to spend a bit more time on it - I think we can maintain the existing behavior (using results of a completed worker even if more work is underway) while addressing this bug.

ahocevar:
@tschaub, feel free to commit to this pull request's branch when you come up with something. Thanks!

ahocevar:
#6658 solves the issue in a better way. Closing.

PR #6655, opened 2017-03-29
Dhivehi Maps source added

dhivehi:
Dhivehi Maps is a free map to all & would like to include in this library.


tschaub:
Thanks for the suggestion @dhivehi.

Benefits of adding things like this:
 * makes it easy for people to get the attribution and URL right

Drawbacks:
 * XYZ source adds an option, people using this source can't apply it
 * URL changes, apps need to wait for a new release instead of updating their own code, meanwhile, library is broken

What do others think of adding "convenience" sources like these?  My preference would be to see an example added instead of a new source type.

bartvde:
I'm in favor of just an example

Issue #6654, opened 2017-03-29
Documentation for zIndex in regards to ol.layer.Group needs clarification


PatchesMaps:
It is unclear in the documentation as to how `zIndex` is handled when rendering `ol.layer.Group` objects with sub layers.

After digging into the source code a little, it appears to me that [`zIndex` set on the parent layer (`ol.layer.Group`) will not be inherited by the child layer](https://github.com/openlayers/openlayers/blob/ac84e560bc4a2b7e0e37cf306decf0e676a9cf05/src/ol/layer/group.js#L208).  So if no `zIndex` is set on the child layers, their `zIndex` will default to zero (even if the parent layer has a `zIndex` set to 100) and any non-group layer that is added to the map with a non-zero `zIndex` will be rendered in front of them.  Is `zIndex` on `ol.layer.Group` ever respected in the rendering context?  If so can we clarify how it interacts with the child layers and other layers on the map?

https://jsfiddle.net/wajxb9eb/1/

ahocevar:
I think the current behavior is unexpected. Layers should inherit `zIndex` from their parent, unless a `zIndex` is already set on the layer.

Are you willing to create a pull request?

Patchez:
Sure, I can see what I can do.

Patchez:
@ahocevar Do you have any input on this PR?

bryantevans00:
Could this be reopened as the issue is still present?

@ahocevar:
> I think the current behavior is unexpected. Layers should inherit zIndex from their parent, unless a zIndex is already set on the layer.

Another way to think about nested z-indices would be to consider layer groups with z-index set to act like [DOM stacking contexts][1]: if a layer group has a set z-index, then the z-indices of its children would only have an effect to the relative ordering of the group layers. That way layer groups could act as an useful mechanism to enforce a specific z-index for multiple layers at once while still having the ability to control the stacking of the group contents. What do you think?

[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

PR #6661, opened 2017-03-31
Inherit zIndex from Group layer if present

Patchez:
Addresses [Issue-6654](https://github.com/openlayers/openlayers/issues/6654)

### DIFF
--- a/src/ol/layer/group.js
+++ b/src/ol/layer/group.js
@@ -221,8 +221,9 @@ ol.layer.Group.prototype.getLayerStatesArray = function(opt_states) {
         layerState.extent = ownLayerState.extent;
       }
     }
-    if (layerState.zIndex === 0 && ownLayerState.zIndex !== 0) {
+    if (layerState.layer.getZIndex() === undefined) {
       layerState.zIndex = ownLayerState.zIndex;
+      layerState.layer.setZIndex(ownLayerState.zIndex);
     }
   }
 
--- a/src/ol/layer/group.js
+++ b/src/ol/layer/group.js
@@ -223,7 +223,6 @@ ol.layer.Group.prototype.getLayerStatesArray = function(opt_states) {
     }
     if (layerState.layer.getZIndex() === undefined) {
       layerState.zIndex = ownLayerState.zIndex;
-      layerState.layer.setZIndex(ownLayerState.zIndex);
     }
   }
 
--- a/src/ol/layer/group.js
+++ b/src/ol/layer/group.js
@@ -221,6 +221,9 @@ ol.layer.Group.prototype.getLayerStatesArray = function(opt_states) {
         layerState.extent = ownLayerState.extent;
       }
     }
+    if (layerState.zIndex === 0 && ownLayerState.zIndex !== 0) {
+      layerState.zIndex = ownLayerState.zIndex;
+    }
   }
 
   return states;
--- a/test_rendering/spec/ol/map.test.js
+++ b/test_rendering/spec/ol/map.test.js
@@ -16,6 +16,7 @@ describe('ol.rendering.Map', function() {
     target = createMapDiv(50, 50);
 
     var vectorLayer = new ol.layer.Vector({
+      zIndex: 0,
       source: new ol.source.Vector({
         features: [new ol.Feature({
           geometry: new ol.geom.Point([0, 0])
--- a/test_rendering/spec/ol/map.test.js
+++ b/test_rendering/spec/ol/map.test.js
@@ -16,7 +16,6 @@ describe('ol.rendering.Map', function() {
     target = createMapDiv(50, 50);
 
     var vectorLayer = new ol.layer.Vector({
-      zIndex: 0,
       source: new ol.source.Vector({
         features: [new ol.Feature({
           geometry: new ol.geom.Point([0, 0])
--- a/src/ol/layer/base.js
+++ b/src/ol/layer/base.js
@@ -47,6 +47,7 @@ ol.layer.Base = function(options) {
    */
   this.state_ = /** @type {ol.LayerState} */ ({
     layer: /** @type {ol.layer.Layer} */ (this),
+    zIndex: 0,
     managed: true
   });
 
--- a/src/ol/layer/base.js
+++ b/src/ol/layer/base.js
@@ -33,8 +33,7 @@ ol.layer.Base = function(options) {
       options.opacity !== undefined ? options.opacity : 1;
   properties[ol.layer.Property.VISIBLE] =
       options.visible !== undefined ? options.visible : true;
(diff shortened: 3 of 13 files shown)


src/ol/layer/group.js
```diff
@@ -221,6 +221,9 @@ ol.layer.Group.prototype.getLayerStatesArray = function(opt_states) {
         layerState.extent = ownLayerState.extent;
       }
     }
+    if (layerState.zIndex === 0 && ownLayerState.zIndex !== 0) {
```
gberaudo: @Patchez, the user may set a `zIndex` of 0 so it cannot be used to mean `no value`. Instead, the zIndex should be tested against `undefined`.

You will have to change the `zIndex` initialization at https://github.com/openlayers/openlayers/blob/master/src/ol/layer/base.js#L36 and change places where the zIndex is used to fallback to 0 if `undefined`.

Patchez: Thanks, didn't think of that.

Patchez: I've changed the `zIndex` initialization to allow `undefined` and changed the conditional:
```javascript
if (layerState.zIndex === undefined) {
      if (ownLayerState.zIndex !== undefined) {
        layerState.zIndex = ownLayerState.zIndex;
        layerState.layer.setZIndex(ownLayerState.zIndex);
      } else {
        layerState.zIndex = 0;
        ownLayerState.zIndex = 0;
        layerState.layer.setZIndex(0);
        ownLayerState.layer.setZIndex(0);
      }
    }
```
I also made changes to [`ol.renderer.Map.sortByZIndex`](https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/map.js#L333) to order `zIndex: undefined` the same as `zIndex: 0`.  However, many existing tests expect a layer to initialize with a default `zIndex` of `0`.   Should I go ahead and change these tests to expect `zIndex: 0`?

gberaudo: Yes I think it makes sense.

Patchez:
I think this now functions as expected and should be ready for review

test_rendering/spec/ol/map.test.js
```diff
@@ -16,6 +16,7 @@ describe('ol.rendering.Map', function() {
     target = createMapDiv(50, 50);
 
     var vectorLayer = new ol.layer.Vector({
+      zIndex: 0,
```
gberaudo: Is this change necessary?

src/ol/layer/group.js
```diff
@@ -221,6 +221,10 @@ ol.layer.Group.prototype.getLayerStatesArray = function(opt_states) {
         layerState.extent = ownLayerState.extent;
       }
     }
+    if (layerState.layer.getZIndex() === undefined) {
+      layerState.zIndex = ownLayerState.zIndex;
+      layerState.layer.setZIndex(ownLayerState.zIndex);
```
gberaudo: Do you need to set the zIndex here?
I would expect that the internal mechanism of computing default zIndex would not interfere with the values set by the user.

Patchez: I was debating this myself but figured since we are inheriting the `ol.layer.Group`'s zIndex we would want to carry that over to the properties.  I'll remove it if you don't think it makes sense.

test_rendering/spec/ol/map.test.js
Patchez: I was having some weird behavior with the rendering tests, the generated image would visually match the expected image but get `Error: done() called multiple times`.  Adding `zIndex: 0` resolved these issues.

src/ol/layer/group.js
gberaudo: The problem I see is that if the parent group zIndex change, the child zIndex will no longer be computed again, so yes I think it is necessary to change the above code.
I will be in vacations next week, so I will not be able to continue the review. But there are others in the community to continue the review. Maybe @fredj, or @ahocevar will be interested to help having your fix merged.

test_rendering/spec/ol/map.test.js
Patchez: I dug into this issue a bit more and found the underlying cause.  I'll include the fix in the next commit.

Patchez: I addressed this in [Commit-b843d7d](https://github.com/openlayers/openlayers/pull/6661/commits/b843d7de52e4ea89746c05dbdef2b8f31ea57293)

src/ol/layer/group.js
Patchez: I addressed this in [Commit-b843d7d](https://github.com/openlayers/openlayers/pull/6661/commits/b843d7de52e4ea89746c05dbdef2b8f31ea57293)

Patchez:
@gberaudo How is this looking?  If if there are any more changes that are needed I'd be more than happy to make them

Patchez:
@ahocevar I took a look at that build failure and while all of the tests passed, it still failed with this message:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```
Any ideas on what could be causing this?

ahocevar:
@Patchez I restarted the build. The error you saw was probably because of a timeout of a service that was not available. However, your most recent commit does not address my review comments. Let me know if you have any questions.

Patchez:
@ahocevar Sorry about that, could you clarify what you'd like to see?  I didn't fully understand what you were asking for.

ahocevar:
@Patchez I'd prefer to see the zIndex **always** initialised with `undefined` instead of `0`. To make that work consistently, you can more or less go back to your original version of this pull request where it accepted `undefined`. You'll only need to modify `ol.renderer.Map.sortByZIndex` so it returns `0` when one or both compared `zIndex`s are `undefined`.

Patchez:
Okay, thank you.  I'll get on that.

Patchez:
Sorry about the delay, I've been very busy recently and haven't had time to work on this.  I'll hopefully be able work on this again soon.

ahocevar:
@Patchez Are you planning to finish this?

Patchez:
Sorry, @ahocevar I've been under some crazy deadlines extraneous to this.  I think I'll have a spare moment this week to take a look at wrapping this up but if someone wants to help out with this it would definitely be appreciated.

Patchez:
@ahocevar I'm having trouble understanding what you mean by
> To make that work consistently, you can more or less go back to your original version of this pull request where it accepted undefined

Do you mean the first commit (`0ab0d659a8de613af8ac3609a5f6917d9e4eff55`)?  Because I don't think that one really changed how z-index is initialized.

Sorry again for the delays, I honestly got a little burned out there for a while but I am really trying to get back into this.

PR #6662, opened 2017-04-02
dhivehi-maps.html

dhivehi:
As suggested, i have made an example instead of a source


ahocevar:
Just an XYZ source. I think we have enough examples already showing how to work with these.

PR #6666, opened 2017-04-04
Use the optional extent in ol.structs.RBush#getExtent

fredj:


fredj:
> It looks like this `getExtent` method is not used anywhere. Could it just be removed instead?

I thought the same but this method is used here: https://github.com/openlayers/openlayers/blob/v4.0.1/src/ol/source/vector.js#L617

ahocevar:
Ok, but it is used there without the optional extent. So I think the handling of the optional extent should either be removed from `ol.structs.RBush#getExtent()`, or `ol.source.Vector` should call the method with an extent.

fredj:
I've added the optional extent in `ol.source.Vector#getExtent`

src/ol/source/vector.js
```diff
@@ -610,11 +610,12 @@ ol.source.Vector.prototype.getClosestFeatureToCoordinate = function(coordinate,
  *
  * This method is not available when the source is configured with
  * `useSpatialIndex` set to `false`.
+ * @param {ol.Extent=} opt_extent Destination extent.
```
ahocevar: Hm. Do we expose reusable arrays elsewhere in the API? As a user, I'd probably never pass that `opt_extent`. But maybe that's just me.

If we decide to expose that to the API, the option should at least be better documented, so users know what "Destination extent" means exactly.

fredj: We expose this `opt_extent` in almost every `ol.extent.*` function but it's also very poorly documented there.

fredj:
Much better, thanks Andreas

PR #6675, opened 2017-04-06
Reset cursor when translate interaction is removed or deactivated

### LINKED ISSUE #6673: Translate grab cursor is still visible when interaction is deactivated while mouse is still on the feature.
See this, follow the instructions : https://codepen.io/tchandelle/pen/oZOdVj

(edit: oops, wrong account, was supposed to be from @tchandelle)

tchandelle:
fixes #6673

### DIFF
--- a/src/ol/interaction/translate.js
+++ b/src/ol/interaction/translate.js
@@ -2,6 +2,7 @@ goog.provide('ol.interaction.Translate');
 
 goog.require('ol');
 goog.require('ol.Collection');
+goog.require('ol.events');
 goog.require('ol.events.Event');
 goog.require('ol.functions');
 goog.require('ol.array');
@@ -82,6 +83,11 @@ ol.interaction.Translate = function(opt_options) {
    * @private
    */
   this.lastFeature_ = null;
+
+  ol.events.listen(this,
+      ol.Object.getChangeEventType(ol.interaction.Property.ACTIVE),
+      this.handleActiveChanged_, this);
+
 };
 ol.inherits(ol.interaction.Translate, ol.interaction.Pointer);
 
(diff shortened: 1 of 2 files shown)


src/ol/interaction/translate.js
```diff
@@ -234,6 +240,35 @@ ol.interaction.Translate.prototype.setHitTolerance = function(hitTolerance) {
...
+  ol.interaction.Pointer.prototype.setMap.call(this, map);
+  this.updateState_(oldMap);
+};
+
+
+/**
+ * @param {ol.Map} oldMap Old map.
+ * @private
+ */
+ol.interaction.Translate.prototype.updateState_ = function(oldMap) {
```
tschaub: It looks like `oldMap` will be an `ol.Object.Event` when the `active` property is changed.

tschaub:
This doesn't need to be handled here, but it would be nice if interactions accepted a `cursor` option.  Then all the cursor handling logic could go in `ol.interaction.Interaction`.

src/ol/interaction/translate.js
tchandelle: @tschaub : When the `active` property is changed to `false`, `getMap()` should returns something, so `oldMap` would not be used. But if nonetheless `getMap()` returns `null`, I don't see any cases where `previousCursor_` would be defined.

tschaub: My point is that you are registering `updateState_` as a [listener for object change events](https://github.com/tchandelle/openlayers/blob/d9976b990ef4a84766da166d87307919196befa8/src/ol/interaction/translate.js#L87-L89).  The first argument to event listeners is an event, in this case `ol.Obect.Event`.  But your function expects `ol.Map`.

The compiler isn't catching the type mismatch.  And the current code will not break, but if someone else modifies this function and expects `oldMap` to always be an `ol.Map`, they will be surprised when they receive `ol.Object.Event` instead.

tschaub: As an alternative, you could override `interaction.setActive()` and this could call `interaction.updateState_()` with the appropriate arguments (just as you have done with `interaction.setMap()`).

Or, if you want to continue registering a listener for change events, add a listener that expects an `ol.Object.Event` and then call `interaction.updateState_()` with the appropriate arguments.

Overriding `setActive` feels more straightforward to me.

tchandelle: Ok, I did that. But I didn't override `setActive` because someone could change the active status with `.set('active', false);`

tschaub:
Thanks for the work on this @tchandelle.

PR #6677, opened 2017-04-06
Add an option to writeTransaction to support 3D geometries

### LINKED ISSUE #6630: Support for 3D WFS-T
Currently the GML serializers will never use the Z coordinates when serialiazing a point. See:
- In GML2: [getCoords_](https://github.com/openlayers/openlayers/blob/master/src/ol/format/gml2.js#L473)
- In GML3: [writePos_](https://github.com/openlayers/openlayers/blob/master/src/ol/format/gml3.js#L566) and [getCoords_](https://github.com/openlayers/openlayers/blob/master/src/ol/format/gml3.js#L591)

This means that edition through WFS-T of 3D layers cannot work. What I expect is: edition works and OpenLayers uses the Z provided by the server in the `GetFeature` request.

I looked into it yesterday and even if it is easy to add the Z coordinates to the request, it is not easy in these methods to know if we are working with 3D geometries or not since OpenLayers will always add Z even if it is not present in the response from the server. We could add Z to all requests, even for 2D geometries, but, in this case, the WFS-T server is likely to reject the transaction.

The solution I see is to add a boolean `3D` option to `olx.format.WFSWriteTransactionOptions`. We could then use it to add or not the Z coordinate. Before doing this and proposing a PR, I would like feedback on this solution and eventually a better solution.

Jenselme:
Close #6630

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2299,6 +2299,7 @@ olx.format.WFSWriteGetFeatureOptions.prototype.resultType;
  *     featureType: string,
  *     srsName: (string|undefined),
  *     handle: (string|undefined),
+ *     is3D: (boolean|undefined),
  *     nativeElements: Array.<Object>,
  *     gmlOptions: (olx.format.GMLOptions|undefined),
  *     version: (string|undefined)}}
@@ -2345,6 +2346,15 @@ olx.format.WFSWriteTransactionOptions.prototype.srsName;
  * @api
  */
 olx.format.WFSWriteTransactionOptions.prototype.handle;
+
+
+/**
+ * Must be set to true if the transaction is for a 3D layer. This will allow
+ * the Z coordinate to be included in the transaction.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.format.WFSWriteTransactionOptions.prototype.is3D;
 
 
 /**
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2299,7 +2299,7 @@ olx.format.WFSWriteGetFeatureOptions.prototype.resultType;
  *     featureType: string,
  *     srsName: (string|undefined),
  *     handle: (string|undefined),
- *     is3D: (boolean|undefined),
+ *     hasZ: (boolean|undefined),
  *     nativeElements: Array.<Object>,
  *     gmlOptions: (olx.format.GMLOptions|undefined),
  *     version: (string|undefined)}}
@@ -2354,7 +2354,7 @@ olx.format.WFSWriteTransactionOptions.prototype.handle;
  * @type {boolean|undefined}
  * @api
  */
-olx.format.WFSWriteTransactionOptions.prototype.is3D;
+olx.format.WFSWriteTransactionOptions.prototype.hasZ;
 
 
 /**
--- a/src/ol/format/gml2.js
+++ b/src/ol/format/gml2.js
@@ -471,7 +471,9 @@ ol.format.GML2.prototype.getCoords_ = function(point, opt_srsName, opt_is3D) {
       point[0] + ',' + point[1] :
       point[1] + ',' + point[0]);
   if (opt_is3D) {
-    coords += ',' + point[2];
+    // For newly created points, Z can be undefined.
+    var z = point[2] || 0;
+    coords += ',' + z;
   }
 
   return coords;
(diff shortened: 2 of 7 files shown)


Jenselme:
Any thoughts on this ?

bartvde:
Reviewing now

externs/olx.js
```diff
@@ -2333,6 +2333,15 @@ olx.format.WFSWriteTransactionOptions.prototype.handle;
 
 
 /**
+ * Must be set to true if the transaction is for a 3D layer. This will allow
+ * the Z coordinate to be include in the transaction.
+ * @type {boolean|undefined}
+ * @api
+ */
+olx.format.WFSWriteTransactionOptions.prototype.is3D;
```
bartvde: you need to add it to the @typedef as well

externs/olx.js
```diff
@@ -2333,6 +2333,15 @@ olx.format.WFSWriteTransactionOptions.prototype.handle;
 
 
 /**
+ * Must be set to true if the transaction is for a 3D layer. This will allow
+ * the Z coordinate to be include in the transaction.
```
bartvde: included

externs/olx.js
bartvde: I wonder if ```hasZ``` is a better name since we are using that already in the library?

src/ol/format/gml2.js
```diff
@@ -456,17 +458,25 @@ ol.format.GML2.prototype.writeRing_ = function(node, ring, objectStack) {
 /**
  * @param {Array.<number>} point Point geometry.
  * @param {string=} opt_srsName Optional srsName
+ * @param {boolean=} opt_is3D wether the geometry is 3D or not.
```
bartvde: whether

test/spec/ol/format/wfs.test.js
```diff
@@ -839,6 +839,86 @@ describe('ol.format.WFS', function() {
     });
   });
 
+  describe('when writting out a transaction request', function() {
```
bartvde: writing

externs/olx.js
Jenselme: > I wonder if hasZ is a better name since we are using that already in the library?

Good question. I have no strong opinion on this. If it's already used, I say I should change is3D to hasZ.

bartvde: Makes sense to me.

Jenselme:
PR updated to take into accounts your remarks. I can squash the commit about typo with e3db4f0 ( Add an option to writeTransaction to support 3D geometries ) if you want (I didn't yet to ease the review of the corrections).

bartvde:
if you can squash, great, TIA

Jenselme:
Thanks for the review @bartvde Commit about typo squashed. I don't know why travis failed on the previous build.

bartvde:
let's see what Travis reports now

bartvde:
it seems rendering tests fail on all PRs right now, not related to this

bartvde:
@Jenselme can you resolve the conflict? It's because #6524 got merged just now

Jenselme:
@bartvde conflicts solved by rebasing on top of master.

bartvde:
I'm going to merge as an admin, since the rendering test failure is unrelated IMHO

src/ol/format/gml3.js
```diff
@@ -1168,7 +1187,7 @@ ol.format.GML3.prototype.GEOMETRY_NODE_FACTORY_ = function(value, objectStack, o
 ol.format.GML3.prototype.writeGeometryNode = function(geometry, opt_options) {
   opt_options = this.adaptOptions(opt_options);
   var geom = ol.xml.createElementNS('http://www.opengis.net/gml', 'geom');
-  var context = {node: geom, srsName: this.srsName,
+  var context = {node: geom, hasZ: this.hasZ, srsName: this.srsName,
```
fredj: `this.hasZ` is used but this variable is not initialized (in the constructor)

fredj: ping @Jenselme

Jenselme: Indeed. From what I see, this is also the case for `this.featureNS` and `this.srsName`. Should I create a PR to initialize these variables in the constructor like the others?

fredj: `this.featureNS` and `this.srsName` are initialized in the parent constructor (`ol.format.GMLBase`).

Jenselme: Indeed. I'll just initialize `hasZ` in `GML3` then.

PR #6687, opened 2017-04-10
Clip image only if the layer extent intersects the view extent

### LINKED ISSUE #6685: Performance issue on Firefox with layer extent
We have a lot of layers to display (more than 200). On Chrome and IE, there is no problems. But on Firefox, big performance issue, moving on the map is very slow.

But if I remove the `extent` on my 200 `ol.layer.Image`, it's nearly as fast as the other browser.

tchandelle:
Fixes #6685

### DIFF
--- a/src/ol/renderer/canvas/intermediatecanvas.js
+++ b/src/ol/renderer/canvas/intermediatecanvas.js
@@ -3,6 +3,7 @@ goog.provide('ol.renderer.canvas.IntermediateCanvas');
 goog.require('ol');
 goog.require('ol.coordinate');
 goog.require('ol.dom');
+goog.require('ol.extent');
 goog.require('ol.renderer.canvas.Layer');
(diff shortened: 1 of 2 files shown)


src/ol/renderer/canvas/intermediatecanvas.js
```diff
@@ -45,7 +45,8 @@ ol.renderer.canvas.IntermediateCanvas.prototype.composeFrame = function(frameSta
 
     // clipped rendering if layer extent is set
     var extent = layerState.extent;
-    var clipped = extent !== undefined;
+    var clipped = extent !== undefined &&
+        !ol.extent.containsExtent(extent, frameState.extent);
```
ahocevar: We need to clip when the layer extent intersects but does not contain the frame state's extent. This check only does the 2nd part of the logic.

PR #6681, opened 2017-04-10
Add constrainResolution in olx.interaction.PinchZoomOptions typedef

fredj:

PR #6689, opened 2017-04-11
Add new constrainResolution option to olx.interaction.DefaultsOptions

fredj:
If `constrainResolution` is needed for both the pinch and mouse wheel interaction, the interactions initialization is something like:
```js
ol.interaction.defaults({
  pinchZoom: false,
  mouseWheelZoom: false
}).extend([
  new ol.interaction.PinchZoom({
    constrainResolution: true
  }),
  new ol.interaction.MouseWheelZoom({
    constrainResolution: true
  })
]);
```
This PR suggest to add a new option to `ol.interaction.defaults` to be able to write:
```js
ol.interaction.defaults({
  constrainResolution: true
});
```

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,16 @@
 
 ### Next release
 
+#### Simpler `constrainResolution` configuration
+
+The `constrainResolution` configuration for `ol.interaction.PinchZoom` and `ol.interaction.MouseWheelZoom`
+can now be set directly with an option in `ol.interaction.defaults`:
+```js
+ol.interaction.defaults({
+  constrainResolution: true
+});
+```
+
 ### v4.0.0
 
 #### Simpler `ol.source.Zoomify` `url` configuration
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2398,7 +2398,9 @@ olx.interaction;
 
 /**
  * Interactions for the map. Default is `true` for all options.
- * @typedef {{altShiftDragRotate: (boolean|undefined),
+ * @typedef {{
+ *     altShiftDragRotate: (boolean|undefined),
+ *     constrainResolution: (boolean|undefined),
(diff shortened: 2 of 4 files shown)


fredj:
And I will also add a note for the new `constrainResolution` option for `ol.interaction.MouseWheelZoom`

externs/olx.js
```diff
@@ -2399,6 +2399,7 @@ olx.interaction;
 /**
  * Interactions for the map. Default is `true` for all options.
  * @typedef {{altShiftDragRotate: (boolean|undefined),
```
cs09g: Doesn't it need a new line? like [#6681](https://github.com/openlayers/openlayers/pull/6681/commits/639f40cb94adaae3ef18bc9dbecfba38ff404d40)

fredj: you're right; it's not necessary but I personally find:
```
@typedef {{
    foo: ...,
    bar: ...
}}
```
nicer than:
```
@typedef {{foo: ...,
    bar: ...}}
```

fredj:
@ahocevar can you please review the changes in `changelog/upgrade-notes.md`? thanks

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,26 @@
...
+
+The `constrainResolution` configuration for `ol.interaction.PinchZoom` and `ol.interaction.MouseWheelZoom`
+can now be set directly with an option in `ol.interaction.defaults`:
+```js
+ol.interaction.defaults({
+  constrainResolution: true
+});
+```
+
+#### `ol.interaction.MouseWheelZoom` no longer zooms to a whole-number zoom level
```
ahocevar: This change was introduced with v3.20.0 already, so it should not need a mention here.

fredj: removed

ahocevar:
I just created #6692, which is somewhat related (doc improvements).

PR #6695, opened 2017-04-12
Unique layers

### LINKED ISSUE #6694: Adding a layer to a map multiple times breaks things
We currently don't complain if someone does this:
```js
map.addLayer(layer);
map.addLayer(layer);
```

But then we fail in mysterious ways when someone does this:
```js
map.getLayers().clear();
```

In addition to layer groups using `ol.getUid(layer)` to keep track of layer listeners, the map renderer keeps track of layer renderers based on the same.  So removing a layer after adding it twice gets things in a bad place.

Instead of waiting until someone tries to remove a layer that was added twice, I think we should throw when a user tries to add a layer twice.  If we didn't have the extra (unnecessary) abstraction of `ol.Collection`, this could be handled in the map.  But the way things are, it feels like supporting a `unique` option for collections is the way to go.

tschaub:
This adds support for unique collections and makes it so map layers must be unique.

image

image

It is already a requirement that map layers be unique.  This just makes it so we fail faster and more clearly.

Fixes #6694.

### DIFF
--- a/src/ol/collection.js
+++ b/src/ol/collection.js
@@ -23,18 +23,33 @@ goog.require('ol.events.Event');
  * @extends {ol.Object}
  * @fires ol.Collection.Event
  * @param {!Array.<T>=} opt_array Array.
+ * @param {olx.CollectionOptions=} opt_options Collection options.
  * @template T
  * @api
  */
-ol.Collection = function(opt_array) {
+ol.Collection = function(opt_array, opt_options) {
 
   ol.Object.call(this);
+
+  var options = opt_options || {};
+
+  /**
+   * @private
+   * @type {boolean}
+   */
+  this.unique_ = !!options.unique;
 
   /**
    * @private
    * @type {!Array.<T>}
    */
   this.array_ = opt_array ? opt_array : [];
+
+  if (this.unique_) {
+    for (var i = 0, ii = this.array_.length; i < ii; ++i) {
+      this.assertUnique_(this.array_[i], i);
+    }
+  }
 
   this.updateLength_();
 
@@ -125,6 +140,9 @@ ol.Collection.prototype.getLength = function() {
  * @api
  */
 ol.Collection.prototype.insertAt = function(index, elem) {
+  if (this.unique_) {
+    this.assertUnique_(elem);
+  }
   this.array_.splice(index, 0, elem);
   this.updateLength_();
   this.dispatchEvent(
@@ -150,6 +168,9 @@ ol.Collection.prototype.pop = function() {
  * @api
  */
 ol.Collection.prototype.push = function(elem) {
+  if (this.unique_) {
+    this.assertUnique_(elem);
+  }
   var n = this.getLength();
   this.insertAt(n, elem);
   return this.getLength();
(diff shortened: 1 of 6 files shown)


test/spec/ol/collection.test.js
```diff
@@ -298,4 +298,67 @@ describe('ol.collection', function() {
     });
   });
 
+  describe('unique collection', function() {
+    it('allows unique items in the constructor', function() {
+      new ol.Collection([{}, {}, {}], {unique: true});
+    });
+
+    it('throws if duplicate items are passed to the contructor', function() {
```
marcjansen: typo: con**s**tructor

test/spec/ol/collection.test.js
```diff
@@ -298,4 +298,67 @@ describe('ol.collection', function() {
...
+
+    it('throws if duplicate items are added via setAt', function() {
+      var unique = new ol.Collection(undefined, {unique: true});
+      var item = {};
+      unique.setAt(0, item);
+      var call = function() {
+        unique.setAt(1, item);
+      };
+      expect(call).to.throwException();
+    });
```
marcjansen: You might want to add: 

```js
it('doesn\'t throw when same item is set via setAt', function() {
  var item = {};
  var unique = new ol.Collection([item], {unique: true});
  var call = function() {
    unique.setAt(0, item);
  };
  expect(call).to.not.throwException();
});
```

This should pass if I read the code correctly.

src/ol/collection.js
```diff
@@ -225,6 +249,20 @@ ol.Collection.prototype.updateLength_ = function() {
...
 
 /**
+ * @private
+ * @param {T} elem Element.
+ * @param {number=} opt_except Optional index to ignore.
+ */
+ol.Collection.prototype.assertUnique_ = function(elem, opt_except) {
+  for (var i = 0, ii = this.array_.length; i < ii; ++i) {
+    if (this.array_[i] === elem && i !== opt_except) {
+      throw new Error('Duplicate item added to a unique collection');
```
ahocevar: I'm wondering if this should use `ol.asserts.assert` with an error code instead. Would save a few bytes in a build probably.

tschaub: Yeah, was thinking the same.  There is a related issue with our version detection that I'd like to look into (links in error messages from our custom builds are wrong - this may no longer be a problem with the ol package).  This issue has plagued us in production for a long time.  At first it was awkward to properly identify due to a useless stack trace and odd failure point.  With the ol package we have a good stack trace but puzzling failure point.  I'm wanting this to fail at the right point and with a clear message to save others the same pain.

I'm compelled by the byte saving argument.  But if we are really committed to that, we'd get rid of ol.Collection entirely.

ahocevar: > I'm compelled by the byte saving argument. But if we are really committed to that, we'd get rid of ol.Collection entirely.

Also true. So I think you can merge this, and removal of `ol.Collection` is something we can save for later.

test/spec/ol/collection.test.js
tschaub: Good catch.  Addressed in 9c8a3ba688fc4f7f79cfd252a3db737ffe94faad.

test/spec/ol/collection.test.js
tschaub: Added in 9c8a3ba688fc4f7f79cfd252a3db737ffe94faad.

src/ol/collection.js
tschaub: Looks like `ol.asserts.assert()` isn't helpful when using the `ol` package since we don't provide `ol.VERSION`.  I'll leave this as something we can address separately.

tschaub: See #6707 for setting `VERSION` in the `ol` package.

PR #6701, opened 2017-04-12
Modify ignored evt.coordinate, which did not play well with Snap.

polar:
The Modify interaction used map.getCoordinateFromPixel in handleDownEvent and handlePointerAtPixel instead of using evt.coordinate. evt.coordinate is set by the Snap interaction. Depending on the resolution, map.getCoordinateFromPixel will not return the same result as evt.coordinate leading to not modifying the intented feature.


ahocevar:
Can you elaborate a bit on cases where `evt.coordinate` and `map.getCoordinateFromPixel` yield different results? When does that happen in the context of the Modify interaction? Ideally you could also come up with a test that fails before and passes after your change.

ahocevar:
Closing due to lack of response.

polar:
Sorry, I didn't get any notification on this comment. It's merely different due to the mathematics. Pixel granularity is not decimal coordinate granularity. Translating from pixel to coordinate gives a value but maybe only the center of the radius of coordinate points, which may not exactly match the coordinate of the vertex in the Modify interaction. Therefore, the Snap interaction will not match the intended vertex.

ahocevar:
Thanks for your response. Can you add a test that fails before and passes after your change?

polar:
Sorry, I'm just getting back to this. I will create a test to show the problem.

polar:
I have a couple of JS Fiddles that illustrate the problem:
First is the problem,
https://jsfiddle.net/polar/kh20ytmr/
The above illustrates that when you grab the dots(point), which are snapped, most of the time, the underlying line vertexes are modified and not the point that was supposed to be snapped. This problem
has to do with line segments that are close to the snapped feature.

The following fiddle is the fix based on the pull request
https://jsfiddle.net/polar/env5nkdt/
This fiddle is using an ol-debug.js with the patch and is residing on my S3 bucket. The Snap grabs the vertex that is snapped and puts it in the event. However, it calculates the pixel (and rounds it to an integer). This causes conversion problems when modify is using map.getCoordinateFromPixel. It just doesn't match the vertex. The pull request merely carries the event.coordinate through the handle calls. 

I've developed a test that will fail in the current version v4.6.4, and will succeed with the patch.

[modify.snap.test.js.txt](https://github.com/openlayers/openlayers/files/1698178/modify.snap.test.js.txt)

PR #6704, opened 2017-04-13
Include typedefs.js in closure compiler tutorial build configurations

### LINKED ISSUE #6698: Compiling v. 4.0.1 with Closure Compiler throws warnings
When trying to follow [this](https://openlayers.org/en/latest/doc/tutorials/closure.html) tutorial, I'm flooded with Closure Compiler warnings (more than 1700) "bad type annotation" (with the types like ol.Coordinate, ol.Size etc.).
I can't find the definisions of these types in any file of the current code (4.0.1), as I could find them in older version (namely 3.14.2). Is it an issue with current version, or am I missing something?

marcjansen:
This changes the configurations used in https://openlayers.org/en/latest/doc/tutorials/closure.html to include `typedefs.js`.

Fixes #6698 

Please review.

### DIFF
--- a/doc/tutorials/closure.md
+++ b/doc/tutorials/closure.md
@@ -172,6 +172,7 @@ The minimum config file looks like this:
       "ol.ENABLE_WEBGL=false"
     ],
     "js": [
+      "node_modules/openlayers/src/ol/typedefs.js",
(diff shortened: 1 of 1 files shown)


marcjansen:
Thanks for the review, Frédéric.

PR #6707, opened 2017-04-13
Set version when building ol package

tschaub:
This makes `VERSION` available in the `ol` package.

```js
import ol from 'ol';

console.log(ol.VERSION);
```

In addition, if `ol.VERSION` is not set, assertion errors report the "latest" URL (e.g. https://openlayers.org/en/latest/doc/errors/#43).

The final change here replaces a normal `Error` with an `AssertionError` and moves the message to the website.

### DIFF
--- a/doc/errors/index.md
+++ b/doc/errors/index.md
@@ -220,3 +220,7 @@ The `{-y}` placeholder requires a tile grid with extent.
 ### 57
 
 At least 2 conditions are required.
+
+### 58
+
+Duplicate item added to a unique collection.  For example, it may be that you tried to add the same layer to a map twice.  Check for calls to `map.addLayer()` or other places where the map's layer collection is modified.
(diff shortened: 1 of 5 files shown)


marcjansen:
Thanks for this :+1:

nufosmatic:
> This makes `VERSION` available in the `ol` package.
> 
> ```js
> import ol from 'ol';
> 
> console.log(ol.VERSION);
> ```

Is this coding supposed to work on recent releases of OL (like v6.3.0)?

I have examples from the website that work (icon.html) and when I attempt to build and run them on my own server they do not. Can't complain because I have no idea what version is actually running in the two cases.

ahocevar:
@nufosmatic See https://codesandbox.io/s/simple-u6pvi.

The relevant code:
```js
import {VERSION} from `ol`;

console.log(VERSION);
```

Issue #6696, opened 2017-04-12
ol.interaction.Translate does not work with imageVector 


ChouGary:
vector is ok like example
http://openlayers.org/en/master/examples/translate-features.html
imageVector can't translate feature , this is demo
https://codepen.io/anon/pen/LyPqWM?editors=0010

ChouGary:
reference #6696 
https://codepen.io/anon/pen/QvwpLV?editors=0010

ChouGary:
verify 4.2.0 not solve
imageVector  can't translate feature , this is demo
https://codepen.io/anon/pen/NggdEL?editors=0010

PR #6719, opened 2017-04-17
Update imagevector.js

ChouGary:
fix map.hasFeatureAtPixel does not work when imageVector feature is selected
#6696


ahocevar:
Replaced by #7501.

PR #6724, opened 2017-04-18
Add wrapX option to Stamen source

### LINKED ISSUE #6723: ol.source.Stamen property wrapX not passed to ol.source.XYZ
When instantiating a new ol.source.Stamen and passing wrapX in the options object the constructor don't pass the wrapX to the parent ol.source.XYZ.

```
ol.source.Stamen = function(options) {
  var i = options.layer.indexOf('-');
  var provider = i == -1 ? options.layer : options.layer.slice(0, i);
  var providerConfig = ol.source.Stamen.ProviderConfig[provider];

  var layerConfig = ol.source.Stamen.LayerConfig[options.layer];

  var url = options.url !== undefined ? options.url :
      'https://stamen-tiles-{a-d}.a.ssl.fastly.net/' + options.layer +
      '/{z}/{x}/{y}.' + layerConfig.extension;

  ol.source.XYZ.call(this, {
    attributions: ol.source.Stamen.ATTRIBUTIONS,
    cacheSize: options.cacheSize,
    crossOrigin: 'anonymous',
    maxZoom: options.maxZoom != undefined ? options.maxZoom : providerConfig.maxZoom,
    minZoom: options.minZoom != undefined ? options.minZoom : providerConfig.minZoom,
    opaque: layerConfig.opaque,
    reprojectionErrorThreshold: options.reprojectionErrorThreshold,
    tileLoadFunction: options.tileLoadFunction,
    wrapX: (typeof(options.wrapX) !== 'undefined') ? options.wrapX : true, //MISSING something like this
    url: url
  });
};
```

fredj:
fixes #6723

### DIFF
--- a/src/ol/source/stamen.js
+++ b/src/ol/source/stamen.js
@@ -35,7 +35,8 @@ ol.source.Stamen = function(options) {
     opaque: layerConfig.opaque,
     reprojectionErrorThreshold: options.reprojectionErrorThreshold,
     tileLoadFunction: options.tileLoadFunction,
-    url: url
+    url: url,
+    wrapX: options.wrapX
   });
 };
 ol.inherits(ol.source.Stamen, ol.source.XYZ);
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -5558,14 +5558,17 @@ olx.source.ImageWMSOptions.prototype.url;
 
 
 /**
- * @typedef {{cacheSize: (number|undefined),
+ * @typedef {{
+ *     cacheSize: (number|undefined),
  *     layer: string,
  *     minZoom: (number|undefined),
  *     maxZoom: (number|undefined),
  *     opaque: (boolean|undefined),
  *     reprojectionErrorThreshold: (number|undefined),
  *     tileLoadFunction: (ol.TileLoadFunctionType|undefined),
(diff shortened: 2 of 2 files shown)


src/ol/source/stamen.js
```diff
@@ -35,7 +35,8 @@ ol.source.Stamen = function(options) {
     opaque: layerConfig.opaque,
     reprojectionErrorThreshold: options.reprojectionErrorThreshold,
     tileLoadFunction: options.tileLoadFunction,
-    url: url
+    url: url,
+    wrapX: options.wrapX
```
Menecats: The default for wrapX in `ol.surce.XYZ` is `true`, in this case if omitted in 'options' the default will be `false`, this could be a better solution:
```
wrapX: typeof(options.wrapX) !== 'undefined' ? options.wrapX : true
```

fredj: See https://github.com/openlayers/openlayers/blob/master/src/ol/source/xyz.js#L57 (parent class)

If the option is omitted (`undefined`) the value will be `true`

Menecats: Sorry, I missed it.

fredj:

fredj:
Thanks for the review

PR #4369, opened 2015-11-03
Add graticule labels feature

Brictarus:
Hello, this is a pull request with an implementation of graticule labels. Please review ! :). 
Fixes #3210
Fixes #4307 (issues are bound)

PR #6737, opened 2017-04-23
Graticule labels

ahocevar:
This pull request replaces #4369. It provides the same API. I implemented things quite differently to improve performance, and tried to use sensible defaults so graticule labels work fine out of the box.

### DIFF
--- a/src/ol/graticule.js
+++ b/src/ol/graticule.js
@@ -1,13 +1,17 @@
 goog.provide('ol.Graticule');
 
+goog.require('ol.coordinate');
 goog.require('ol.extent');
 goog.require('ol.geom.GeometryLayout');
 goog.require('ol.geom.LineString');
+goog.require('ol.geom.Point');
 goog.require('ol.geom.flat.geodesic');
 goog.require('ol.math');
 goog.require('ol.proj');
 goog.require('ol.render.EventType');
+goog.require('ol.style.Fill');
 goog.require('ol.style.Stroke');
+goog.require('ol.style.Text');
 
 
 /**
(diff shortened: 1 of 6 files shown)


src/ol/graticule.js
```diff
@@ -104,31 +108,118 @@ ol.Graticule = function(opt_options) {
...
+  this.parallelsLabels_ = null;
+
+  if (options.showLabels == true) {
+    var degreesToString = ol.coordinate.degreesToStringHDMS;
+
+    /**
+     * @type {null|function(number):string}
+     * @private
+     */
+    this.lonLabelFormatter_ = options.latLabelFormatter == undefined ?
```
bartvde: should be ```options.lonLabelFormatter``` instead

src/ol/graticule.js
```diff
@@ -104,31 +108,118 @@ ol.Graticule = function(opt_options) {
...
+        degreesToString.bind(this, 'EW') :
+        options.latLabelFormatter;
```
bartvde: same here

fredj:
nice improvement, thanks!

src/ol/graticule.js
ahocevar: Thanks for the review. I fixed this and also modified the tests to catch it.

bartvde: Great, I was thinking the same when reading the tests since they had the same function bodies for the formatter functions

cmdelatorre:
Hello, I have a question: is this feature available in OL 4.1.1 ?

ahocevar:
No, but it will be in v4.2.0, which will be released tomorrow.

cmdelatorre:
Awesome! Great news! Thank you very much for the great work!

PR #6741, opened 2017-04-24
Allow user selection in overlay container

### LINKED ISSUE #6720: Allow text selection in overlays
from https://github.com/openlayers/openlayers/issues/6618#issuecomment-294102180:

It is impossible to select text in overlay-based popups after this change. Compare: [4.0.1](http://openlayers.org/en/v4.0.1/examples/popup.html) and [4.1.0](http://openlayers.org/en/latest/examples/popup.html). What about adding `user-select` option to `.ol-overlay-container`?

fredj:
* remove `-ktml` css prefix
 * add new `ol.css.CLASS_SELECTABLE` variable

fixes #6720
replace #6729

### DIFF
--- a/css/ol.css
+++ b/css/ol.css
@@ -37,13 +37,18 @@
 .ol-viewport, .ol-unselectable {
   -webkit-touch-callout: none;
   -webkit-user-select: none;
-  -khtml-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
   -webkit-tap-highlight-color: rgba(0,0,0,0);
 }
-
+.ol-selectable {
+  -webkit-touch-callout: default;
+  -webkit-user-select: auto;
+  -moz-user-select: auto;
+  -ms-user-select: auto;
+  user-select: auto;
+}
 .ol-control {
   position: absolute;
   background-color: rgba(255,255,255,0.4);
--- a/src/ol/css.js
+++ b/src/ol/css.js
@@ -11,6 +11,14 @@ ol.css.CLASS_HIDDEN = 'ol-hidden';
 
 
 /**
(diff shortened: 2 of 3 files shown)


css/ol.css
```diff
@@ -37,13 +37,18 @@
...
   -webkit-tap-highlight-color: rgba(0,0,0,0);
 }
-
+.ol-selectable {
+  -webkit-touch-callout: default;
+  -webkit-user-select: auto;
+  -moz-user-select: auto;
+  -ms-user-select: auto;
+  user-select: auto;
+}
```
tonio: Are all those prefix really useful? http://caniuse.com/#search=user-select

fredj: Yes, only Chrome uses the unprefixed version

css/ol.css
```diff
@@ -37,13 +37,18 @@
...
   user-select: none;
   -webkit-tap-highlight-color: rgba(0,0,0,0);
 }
-
+.ol-selectable {
+  -webkit-touch-callout: default;
+  -webkit-user-select: auto;
+  -moz-user-select: auto;
+  -ms-user-select: auto;
+  user-select: auto;
```
walkermatt: @fredj from my testing while looking into [this question on SO](https://stackoverflow.com/questions/49450503/copy-content-from-popup/49487767#49487767) I think we probably need a value of `text` as `auto` will resolve to `none` if a parent has `user-select: none` as is the case in OpenLayers due to `user-select: none` being set on `.ol-viewport`.

Further details of `auto` in spec: https://drafts.csswg.org/css-ui-4/#valdef-user-select-auto

fredj: @walkermatt feel free to open a pull request to change from `auto` to `text`
Note: the original PR #6729 had `text` instead of `auto`; I don't remember why it was changed.

PR #6759, opened 2017-04-29
Regular shape points

### LINKED ISSUE #5875: ol.style.RegularShape#getPoints may return double amount of points given to the constructor
The method getPoints of ol.style.RegularShape returns double the amount of points given to the constructor if a radius2 was provided.

I think this follows the logic that the points given to the constructor are applied to each circle, thus resulting in the double amount.
But this means that the constructor option `points` and the points returned from getPoints are different from each other, which is not comprehensible by distinguished names or the documentation.

I can think of the following solutions:
1. Match the meanings by having to specify the total amount of points for both circles in the constructor.
2. Match the meanings by letting getPoints only return the number of points on one circle.
3. Distinguish the meanings by renaming the methods name to getTotalPoints or something similar.
4. Distinguish the meanings by documentation only.

I can create a PR for one of these solutions.

And btw i think it would be nicer to have this doubling of points in the constructor, where it can be found easily and not in both the methods draw_ and drawHitCanvas_

simonseyock:
Fixes #5875

In the first commit I changed the draw methods to only double points temporarily and not doubling them every time they get called. This results in getPoints() returning the same number as initially provided to the constructor.

In the second commit I set the radius2 to undefined if it wasn't provided. This results in getRadius2() returning the same value that was provided. This makes the tests that radius1 is used instead of radius2 if radius2 was not provided impossible so i removed them. I don't know if this second commit is really desired, i can remove it if not.

### DIFF
--- a/src/ol/style/regularshape.js
+++ b/src/ol/style/regularshape.js
@@ -66,10 +66,9 @@ ol.style.RegularShape = function(options) {
 
   /**
    * @private
-   * @type {number}
+   * @type {number|undefined}
    */
-  this.radius2_ =
-      options.radius2 !== undefined ? options.radius2 : this.radius_;
+  this.radius2_ = options.radius2;
 
   /**
    * @private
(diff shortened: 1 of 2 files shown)


simonseyock:
Because of the changed return values of getPoints and getRadius2, should there be a release note?

ahocevar:
> Because of the changed return values of getPoints and getRadius2, should there be a release note?
(Just thought about it as I was reading this again)

Could indeed make sense. Would you be willing to create a follow-up pull request? Thanks!

simonseyock:
See #6901

Issue #6762, opened 2017-04-30
view.getHints undefined in ol-debug.js


mblinsitu:
In my application I use `view.getHints()[ol.ViewHint.INTERACTING]` to figure out whether the view change events are due to local interaction.

When I use the CDN `ol.js` (version 4.1.0), `ol.ViewHint` and `view.getHints()` are undefined.
When I use the CDN `ol-debug.js`, all is fine.

Is this intended, or is it a bad side effect of the minimization of the `ol.js` library?

Thanks,
michel

ahocevar:
It's intended, because the `getHints()` method is not part of the official API.

mblinsitu:
Is there an official way then to find out if an interaction or an animation is ongoing?
I need this to ignore requests coming from a server while the user is interacting with the map.
I would rather not have to redefine the interactions to do exactly what `getHints()` is doing...

Cheers,
michel

ahocevar:
Yes, there is. [`ol.View#getAnimating()`](http://openlayers.org/en/latest/apidoc/ol.View.html#getAnimating)

mblinsitu:
Indeed, but there is no `ol.View#getInteracting()` :-(

ahocevar:
We'd accept a pull request to add `ol.View#getInteracting()`.

mblinsitu:
The pull request is here: https://github.com/openlayers/openlayers/pull/6764

PR #6764, opened 2017-05-01
Added View#getInteracting() to the api

mblinsitu:
Pull request for issue https://github.com/openlayers/openlayers/issues/6762

`View#getInteracting()` returns `true` when a user interaction is ongoing, e.g. during panning.
It is similar to `View#getAnimating()` for animations.

Note that a number interactions are continued by an animation, e.g. the view continues to pan after dragging it. In this case `View#getInteracting()` becomes `false` when the user releases the mouse button, but `View#getAnimating()` is `true` while the animation is running.

### DIFF
--- a/test/spec/ol/view.test.js
+++ b/test/spec/ol/view.test.js
@@ -310,11 +310,11 @@ describe('ol.View', function() {
       });
 
       expect(view.getHints()).to.eql([0, 0]);
-      expect(view.getInteracting()).to.eql(0);
+      expect(view.getInteracting()).to.eql(false);
 
       view.setHint(ol.ViewHint.INTERACTING, 1);
       expect(view.getHints()).to.eql([0, 1]);
-      expect(view.getInteracting()).to.eql(1);
+      expect(view.getInteracting()).to.eql(true);
     });
 
     it('triggers the change event', function(done) {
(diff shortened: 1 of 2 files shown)


test/spec/ol/view.test.js
```diff
@@ -310,9 +310,11 @@ describe('ol.View', function() {
       });
 
       expect(view.getHints()).to.eql([0, 0]);
+      expect(view.getInteracting()).to.eql(0);
```
fredj: please use `to.eql(false);` instead of `to.eql(0);`

test/spec/ol/view.test.js
```diff
@@ -310,9 +310,11 @@ describe('ol.View', function() {
...
 
       view.setHint(ol.ViewHint.INTERACTING, 1);
       expect(view.getHints()).to.eql([0, 1]);
+      expect(view.getInteracting()).to.eql(1);
```
fredj: `to.eql(true);`

test/spec/ol/view.test.js
```diff
@@ -323,6 +325,7 @@ describe('ol.View', function() {
 
       view.on('change', function() {
         expect(view.getHints()).to.eql([0, 1]);
+        expect(view.getInteracting()).to.eql(1);
```
fredj: `to.eql(true);`

fredj:
Thanks for this contribution @mblinsitu. It looks good! Can you please submit a [CLA](https://github.com/openlayers/cla)  so we can merge this? Thanks!

mblinsitu:
Thanks for the feedback.
I've submitted the CLA and fixed the test.

ahocevar:
ICLA received.

PR #6765, opened 2017-05-01
Add new callback function to view.FitOptions

### LINKED ISSUE #6763: Feature request: "completed" callback for view#fit with animation
The `view#animate` function takes an optional function which is called once the animation completes.

`view#fit` can be set up to use animation when fitting the view to the given extent, but it is not currently possible to pass a similar callback to this function.

Feature request: Expose a similar callback function for the `view#fit` method.
This could easily be an additional property on the `options` object.

Motivation: When the user selects a feature either from search or from a list, I want to zoom to the selected feature and animate an effect to draw attention to it. The "attention" animation should not start until the feature is in view.

fredj:
fixes #6763

### DIFF
--- a/src/ol/view.js
+++ b/src/ol/view.js
@@ -872,6 +872,7 @@ ol.View.prototype.fit = function(geometryOrExtent, opt_options) {
   var centerX = centerRotX * cosAngle - centerRotY * sinAngle;
   var centerY = centerRotY * cosAngle + centerRotX * sinAngle;
   var center = [centerX, centerY];
+  var callback = options.callback ? options.callback : ol.nullFunction;
 
   if (options.duration !== undefined) {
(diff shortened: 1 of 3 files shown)


src/ol/view.js
```diff
@@ -872,17 +872,19 @@ ol.View.prototype.fit = function(geometryOrExtent, opt_options) {
...
       resolution: resolution,
       center: center,
       duration: options.duration,
       easing: options.easing
-    });
+    }, callback);
   } else {
     this.setResolution(resolution);
     this.setCenter(center);
+    callback(true);
```
ahocevar: It would be safer to call this asynchronously with `setTimeout`.

fredj:
Thanks for the review

Issue #6753, opened 2017-04-28
ES2015 ol package: ol.interaction.Translate constructor fails


AnWeber:
With the new Release v4.0.1 the constructor call of ol.interaction.Translate throws an error. Version 4.0.1-beta2 is working.

###  Error
> angular.js:14525 TypeError: Cannot read property 'getChangeEventType' of undefined
    at new _ol_interaction_Translate_ (...)

### Source of Error
Source of Error was added with 5c23123. Code is working using [openlayers](https://www.npmjs.com/package/openlayers), but not with es2015 [ol](https://www.npmjs.com/package/ol/).

```
 _ol_events_.listen(this,
      _ol_.Object.getChangeEventType(_ol_.interaction.Property.ACTIVE),
      this.handleActiveChanged_, this);
 ```

#### Code

```
import olInteractionTranslate from 'ol/interaction/translate';
new olInteractionTranslate({
  features: ...
});
```

AnWeber:
Thanks for the fast fix. I really appreciate the ES2015 packages. Great work.

tschaub:
Thanks for the report @AnWeber.  The [`ol@4.1.1` release](https://www.npmjs.com/package/ol) includes the fix in #6755.  The same problem shouldn't crop up in future releases after #6770.  (And when we get remaining `closure-util` users to adapt, we can author in ES modules and avoid this problem altogether.)

PR #6770, opened 2017-05-02
Enforce the "no missing requires" rule

tschaub:
This configures the `openlayers-internal/no-missing-requires` rule at "error" level.  This wasn't done previously because:

 * we use lowercase constructors in `ol.ext` (e.g. `new ol.ext.pbf()`) - this is addressed with 584d4f67bda150da4bb24ec9946542bf62e8c4a6
 * we deeply require `ol.ext.foo` things (e.g. `ol.ext.vectortile.VectorTile`) - this is addressed with 3f973e6fa3988473cdf1f3fc6875c1c2912a7a20
 * the generated shader code creates single modules with multiple constructors - this is side stepped with 479d98212b16c3196d80f6fae35c8e7fb5b54798

With the changes above, we can fail if the rule is not satisfied (49b5bbc964555d2cb8d430d44db82db150ed9473).

In addition, to avoid things like #6753 in the future, we ensure the package builds and run the linter on the built package (ignoring `**/*webgl*`) as part of the Travis job.


tschaub:
image

I don't have the patience to deal with the Compiler right now.  I'll get back to this later.

tschaub:
I put together a new task that uses Rollup to wrap dependencies for the Compiler (and the full build is now 3K smaller).

This is ready for review now.  The main purpose of this branch is to make it so we don't accept changes that break the `ol` package.

tschaub:
Thanks for the reviews.

> What about removing the --quiet option to eslint in the Makefile?

Removed in 3c6239653393f9e88ca3280a19dca8e074fd38ba.

PR #6771, opened 2017-05-03
Use ol.obj.clear to reduce garbage generation

fredj:


fredj:
I tested using the cluster example. In this case `this.nullGeometryFeatures_` is always empty so `ol.obj.clear`take no time.

More tests to come

fredj:
I did more tests (with the chrome devtools) and the gain is not not significant enough to justify this change.

PR #6778, opened 2017-05-08
Add a small tolerance when testing pointer event positions

### LINKED ISSUE #6776: Tapping on map on Android doesn't always work after a Chrome fix
After we fixed [an old bug](https://crbug.com/632067) with PointerEvents in Chrome 58, an issue in OpenLayer's tap handling got exposed: a tap on Android doesn't always work.  More precisely, until Chrome 57 we used to (wrongly) suppress <code>pointermove</code> events in a small ("slop-suppression") region around the <code>pointerdown</code> point, which we fixed in Chrome 58.  After the fix, in OpenLayers tap doesn't work unless it's a very quick tap---the pointermoves within the old slop-region is perhaps triggering a pan followed by a "move end" event.

This is evident from the following examples:
- [box-selection.html](https://openlayers.org/en/latest/examples/box-selection.html)
- [moveend.html](https://openlayers.org/en/latest/examples/moveend.html)

I confirmed that it's not a Chrome bug through adding a hack <code>pointermove</code> listener through devtools that drops the move events close to the <code>pointerdown</code>.  I will close the [new Chrome bug](https://crbug.com/718540).

fredj:
fixes #6776 

demo: https://codepen.io/anon/pen/QvOPzp

### DIFF
--- a/src/ol/mapbrowsereventhandler.js
+++ b/src/ol/mapbrowsereventhandler.js
@@ -1,6 +1,7 @@
 goog.provide('ol.MapBrowserEventHandler');
 
 goog.require('ol');
+goog.require('ol.has');
 goog.require('ol.MapBrowserEventType');
 goog.require('ol.MapBrowserPointerEvent');
 goog.require('ol.events');
@@ -42,6 +43,12 @@ ol.MapBrowserEventHandler = function(map) {
    * @private
    */
   this.dragListenerKeys_ = [];
+
+  /**
+   * @type {number}
+   * @private
+   */
+  this.moveTolerance_ = ol.has.DEVICE_PIXEL_RATIO;
 
   /**
    * The most recent "down" type event (or null if none have occurred).
@@ -241,11 +248,9 @@ ol.MapBrowserEventHandler.prototype.handlePointerDown_ = function(pointerEvent) 
  * @private
  */
 ol.MapBrowserEventHandler.prototype.handlePointerMove_ = function(pointerEvent) {
-  // Fix IE10 on windows Surface : When you tap the tablet, it triggers
-  // multiple pointermove events between pointerdown and pointerup with
-  // the exact same coordinates of the pointerdown event. To avoid a
-  // 'false' touchmove event to be dispatched , we test if the pointer
-  // effectively moved.
+  // Between pointerdown and pointerup, pointermove events are triggered.
+  // To avoid a 'false' touchmove event to be dispatched, we test if the pointer
+  // moved a significant distance.
   if (this.isMoving_(pointerEvent)) {
     this.dragging_ = true;
     var newEvent = new ol.MapBrowserPointerEvent(
@@ -281,8 +286,8 @@ ol.MapBrowserEventHandler.prototype.relayEvent_ = function(pointerEvent) {
  * @private
  */
 ol.MapBrowserEventHandler.prototype.isMoving_ = function(pointerEvent) {
-  return pointerEvent.clientX != this.down_.clientX ||
-      pointerEvent.clientY != this.down_.clientY;
+  return Math.abs(pointerEvent.clientX - this.down_.clientX) > this.moveTolerance_ ||
+      Math.abs(pointerEvent.clientY - this.down_.clientY) > this.moveTolerance_;
 };
 
 


fredj:
@ischas can you test https://codepen.io/anon/pen/QvOPzp ?

ischas:
@simonseyock ping

simonseyock:
I am sorry, i can't test this. The problem does not occur on my phone.

src/ol/mapbrowsereventhandler.js
```diff
@@ -281,8 +279,8 @@ ol.MapBrowserEventHandler.prototype.relayEvent_ = function(pointerEvent) {
  * @private
  */
 ol.MapBrowserEventHandler.prototype.isMoving_ = function(pointerEvent) {
-  return pointerEvent.clientX != this.down_.clientX ||
-      pointerEvent.clientY != this.down_.clientY;
+  return Math.abs(pointerEvent.clientX - this.down_.clientX) > 1 ||
+      Math.abs(pointerEvent.clientY != this.down_.clientY) > 1;
```
mustaqahmed: Typo: the second condition should be `pointerEvent.clientY - this.down_.clientY`.

One question: Any chance the threshold `1` may be too large? The real offset that are causing the bug seems well below `1`. May be `0.5` or even `0.9` is a better alternative?

gastonmz: I'm using 1 and works fine

ischas:
For me it's nearly impossible to select a country in this demo:
http://openlayers.org/en/latest/examples/select-features.html

The linked CodePen works for me (but I guess @mustaqahmed is right with its review comment).

src/ol/mapbrowsereventhandler.js
gastonmz: I have problems handling the two fingers in order to make a pinch zoom...

fredj: oups, thanks @mustaqahmed

fredj: CodePen updated

fredj:
I changed the threshold to 0.75 as suggested and updated the CodePen.

ischas:
Now selecting is better than with master, but less easy as with the first commit.

fredj:
@ischas back to 1px, CodePen updated

ischas:
@simonseyock and I tested on my Samsung S5 with Android 6.0.1 and Chrome 58.

While with the select feature demo only round about every fifth to tenth touch is registered as touch, with the CodePen maybe every second touch is registered as touch. That's much better, but not good.

Please consider that the first commit behaved "better" in context of selection because of the bug and `isMoving_` always returning false.

fredj:
I tested on a Samsung S6 and it's not working. I've added the `devicePixelRatio` into the mix

ischas:
CodePen now works fine for me.

fredj:
@ischas and are the pan, pinch zoom and rotate working as well?

fredj:
@simonseyock can you also test on the phone that doesn't have the issue if everything continues to work as expected?

ischas:
I Tested tap, double tap, pan, pinch zoom, rotate, everything's fine. Did I miss anything else?

@simonseyock is out of office until Friday, but maybe he'll find some time for testing in its spare time. ;-)

fredj:
I'm tested with a Nexus 5x (doesn't have the issue) and it works as expected

nagytech:
Tested Nexus 4, Android 5.1.1, Chrome 58.0.3029.83 - previously had the issue now working fine.

fredj:

src/ol/mapbrowsereventhandler.js
```diff
@@ -44,6 +45,12 @@ ol.MapBrowserEventHandler = function(map) {
   this.dragListenerKeys_ = [];
 
   /**
+   * @type {number}
+   * @private
+   */
+  this.moveTolerance_ = 1 * ol.has.DEVICE_PIXEL_RATIO;
```
ahocevar: The `1 * ` is unnecessary.

ischas:
Any chance for a patch release?

mbj4668:
I have tested the CodePen with a Sony Z3 Compact, Android 6.0.1, Chrome 58.0.3029.83, and I still have problems selecting countries.  It works slightly better than http://openlayers.org/en/latest/examples/select-features.html, but not good.

I have also tested with a Sony Z4 Tablet, Android 7.0, Chrome 58.0.3029.83, and there the CodePen works fine, and the select-features example does not work.

fredj:
What's the device pixel ratio of these devices ? see http://devicepixelratio.com/

mbj4668:
On the Z3 compact it reports "2".

On the Z4 tablet it also reports "2".


/martin

2017-05-12 8:44 GMT+02:00 Frédéric Junod <notifications@github.com>:

joaodiasafonso:
Hello.
I am facing this problem since Android webview was updated. If I uninstall latest version and stick to default one, it starts working.
Anyway, none of the workarounds/samples in this conversation worked for me. My device's pixel ratio is 4.
Thanks in advance. Regards,

joaodiasafonso:
Need to make a correction because I was only checking in codepen and ol samples: after properly make @fredj update in library js, it began working as supposed. So, this has worked for me.

jhwegener:
Nice PR, thanks @fredj !
Is there any chance to get a patch release for this?

ahocevar:
We'll be releasing v4.2.0 in the next couple of days anyway.

PR #6779, opened 2017-05-08
Decouple source and rendered tile grid of vector tile sources

### LINKED ISSUE #5580: Option to re-render vector tiles at zoom levels not in source in hybrid render mode

### LINKED ISSUE #6453: Pixelated vector tiles at view resolutions not provided by the tile source

### LINKED ISSUE #6510: OL 4 with Tileserver GL

ahocevar:
This pull request improves rendering quality and performance of vector tile layers for sources that have a limited zoom level range available.

In `hybrid` (default) and `image` rendering mode, images are now pre-rendered for each standard tile grid resolution of the view projection. This means that ESRI vector tile sets which usually have a zoom level range of 1-15 will now render beautifully even at e.g. zoom level 21. This can be seen in the updated `mapbox-vector-tiles-advanced.html` example.

Replaces #6454.
Fixes #6453.
Fixes #5580.
Fixes #6510.

### DIFF
--- a/src/ol/renderer/canvas/vectortilelayer.js
+++ b/src/ol/renderer/canvas/vectortilelayer.js
@@ -116,9 +116,8 @@ ol.renderer.canvas.VectorTileLayer.prototype.createReplayGroup_ = function(
     return;
   }
 
-  var sourceTiles = tile.getSourceTiles();
-  for (var t = 0, tt = sourceTiles.length; t < tt; ++t) {
-    var sourceTile = sourceTiles[t];
+  for (var t = 0, tt = tile.tileKeys.length; t < tt; ++t) {
+    var sourceTile = tile.getTile(tile.tileKeys[t]);
     sourceTile.replayGroup = null;
     replayState.dirty = false;
 
@@ -244,9 +243,8 @@ ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = functi
     if (!ol.extent.containsCoordinate(bufferedExtent, coordinate)) {
       continue;
     }
-    var sourceTiles = tile.getSourceTiles();
-    for (var t = 0, tt = sourceTiles.length; t < tt; ++t) {
-      var sourceTile = sourceTiles[t];
+    for (var t = 0, tt = tile.tileKeys.length; t < tt; ++t) {
+      var sourceTile = tile.getTile(tile.tileKeys[t]);
       if (sourceTile.getProjection().getUnits() === ol.proj.Units.TILE_PIXELS) {
         var sourceTileCoord = sourceTile.tileCoord;
         var sourceTileExtent = sourceTileGrid.getTileCoordExtent(sourceTileCoord, this.tmpExtent);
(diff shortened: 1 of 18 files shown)


ahocevar:
This is now ready for review.

src/ol/renderer/canvas/vectortilelayer.js
```diff
@@ -213,54 +227,62 @@ ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = functi
...
   var tile, tileCoord, tileExtent, tilePixelRatio, tileResolution;
-  for (i = 0, ii = replayables.length; i < ii; ++i) {
-    tile = replayables[i];
+  for (i = 0, ii = renderedTiles.length; i < ii; ++i) {
+    tile = renderedTiles[i];
     tileCoord = tile.tileCoord;
-    tileExtent = source.getTileGrid().getTileCoordExtent(tileCoord, this.tmpExtent);
-    if (!ol.extent.containsCoordinate(ol.extent.buffer(tileExtent, hitTolerance * resolution), coordinate)) {
+    tileExtent = tileGrid.getTileCoordExtent(tileCoord, this.tmpExtent);
+    bufferedExtent = ol.extent.buffer(tileExtent, hitTolerance * resolution);
```
fredj: minor: `bufferedExtent` can be passed as a third argument to `ol.extent.buffer` to avoid creating a new array.

fredj:
In the mapbox-vector-tiles-advanced.html, at zoom level 12 the following error is raised:
```
vectorimagetile.js:204 Uncaught TypeError: Cannot read property '2,0,-3' of null
    at ol.VectorImageTile.getTile (vectorimagetile.js:204)
    at ol.VectorImageTile.disposeInternal (vectorimagetile.js:130)
    at ol.VectorImageTile.ol.Disposable.dispose (disposable.js:24)
    at ol.TileCache.expireCache (tilecache.js:45)
    at ol.source.VectorTile.ol.source.Tile.expireCache (tile.js:93)
    at layer.js:155
    at ol.Map.handlePostRender (map.js:980)
```

fredj:
Less features (buildings, pois, ...) are displayed mapbox-vector-tiles-advanced example (compared to http://openlayers.org/en/master/examples/mapbox-vector-tiles-advanced.html). Is that expected?

ahocevar:
> Less features (buildings, pois, ...) are displayed mapbox-vector-tiles-advanced example (compared to http://openlayers.org/en/master/examples/mapbox-vector-tiles-advanced.html). Is that expected?

Yes, that's expected. Previously, tiles were used from and styled for a higher zoom level than the viewed one.

ahocevar:
> In the mapbox-vector-tiles-advanced.html example, at zoom level 12 the following error is raised:
   ...
Thanks for catching that. It came from the `disposeInternal` method of `ol.VectorImageTile`. I fixed it. Adding some tests for that method now to make sure we've got everything right.

fredj:
> Yes, that's expected. Previously, tiles were used from and styled for a higher zoom level than the viewed one.

ok, thanks for the clarification

ahocevar:
Thanks for the thorough review @fredj.

PR #6786, opened 2017-05-09
Add ol.format.filter.during

fredj:
WFS During temporal operator.
Tested with mapserver

### DIFF
--- a/src/ol/format/filter.js
+++ b/src/ol/format/filter.js
@@ -2,6 +2,7 @@ goog.provide('ol.format.filter');
 
 goog.require('ol.format.filter.And');
 goog.require('ol.format.filter.Bbox');
+goog.require('ol.format.filter.During');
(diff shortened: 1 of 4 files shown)


bartvde:
@fredj is this from Filter 2.0?

fredj:
from http://docs.opengeospatial.org/is/09-026r2/09-026r2.html

bartvde:
but we only support WFS 1.0 and 1.1 right now which are tied to Filter 1.0 and Filter 1.1?

bartvde:
has MapServer backported this filter to WFS 1.0 or 1.1 or are you using WFS 2.0 somehow?

fredj:
I guess WFS 2.0 because the during filter comes from this PR: https://github.com/mapserver/mapserver/pull/4815

fredj:
Thanks for the review Bart

PR #6792, opened 2017-05-10
Create intermediate canvas when resolutions have changed

### LINKED ISSUE #6791: Blurred tiles when layers change during zoom animation
When a map is configured with two layers with `minResolution` and `maxResolution` so a different layer is shown at different zoom levels, the tiles can become blurred.

See https://codepen.io/anon/pen/eWVedo. Steps to reproduce:

* Wait until tiles are loaded.
* Zoom in.
* Wait again until tiles are loaded.
* Zoom out.
* Zoom in.

The tiles will be blurred now. If you do not see the issue immediately, reload the page and repeat the steps.

ahocevar:
Fixes #6791.


ahocevar:
Thanks for the review @fredj.

PR #6793, opened 2017-05-10
Webgl text

GaborFarkas:
This is the bare minimum a WebGL text renderer in OL3 can work with (faulty of course). Before fixing it, we should consider refactoring the image and text replays. As they share the same code (mostly), we could have a `TextureReplay` constructor from both the `TextReplay` and `ImageReplay` constructors could inherit from. I wouldn't go into such changes without asking your opinions about this, so what do you think?


gberaudo:
Hi @GaborFarkas, I will take time to review your code tomorrow.

gberaudo:
Hi @GaborFarkas, yes fine to me to introduce a `TextureReplay` base class.

GaborFarkas:
Thank you, @gberaudo! I will start working on it as soon as I can.

GaborFarkas:
I'm closing and reopening to trigger another Travis build, as I'm very confused about the `ol.View#animate` failed test.

GaborFarkas:
Closing an reopening again, now it's with `ol.Overlay#setVisible`. Nevertheless, if the rendering tests pass, this PR is ready for review.

GaborFarkas:
An `ol.Overlay#setVisible` fail again. I ran out of ideas, it runs perfectly on my node with the same version (6.1.0).

GaborFarkas:
According to some basic tests, it's working fine on HIDPI devices. Of course, it would be even better if we would handle pixelRatio in the replay. For now, we should leave it as is, as fixing it would require some fundamental changes in the WebGL replay system.

Thank you for your review @gberaudo!

PR #6802, opened 2017-05-11
Remove unused setFeatures and getFormat methods

ahocevar:
The new `ol.VectorImageTile` class has a `setFeatures` API method, which is neither working nor used. This pull request suggests to remove it, along with the unused `getFormat` method.


ahocevar:
I think the `getFormat` method and the privat `format_` member can also be removed. Working on another commit.

ahocevar:
Commit for removal of `getFormat()` pushed.

PR #6817, opened 2017-05-16
Multiple layers in TopoJSON vector tiles

ahocevar:
This change introduces `layerName` and `layers` options on `ol.format.TopJSON`. This is useful when using TopoJSON vector tiles. It adds a grouping capability equal to `ol.format.MVT`.

This new feature is shown in an updated `osm-vector-tiles.html` example, and in a simplification of the `topojson.html` example. For `osm-vector-tiles.html`, now that Mapzen introduces quite low tile request limits, it is better to request only one tile set instead of three.

### DIFF
--- a/examples/osm-vector-tiles.html
+++ b/examples/osm-vector-tiles.html
@@ -3,7 +3,7 @@ layout: example.html
 title: OSM Vector Tiles
 shortdesc: Using OpenStreetMap vector tiles.
 docs: >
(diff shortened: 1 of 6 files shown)


ahocevar:
Updated the pull request to use the same API as in `ol.format.MVT`.

ahocevar:
What wording were you envisioning @fredj? Existing apps continue to work without change.

ahocevar:
@fredj: Or do you just want me to squash the commits - `nameAsProperty` was introduced with this PR's first commit, and then changed to `layerName`

ahocevar:
Commits squashed.

PR #6819, opened 2017-05-16
Adjust/fix API for working with vector tile features

ahocevar:
This pull request suggests three changes:

1. Fix tile load events for `ol.source.VectorTile`: Before #6779, when registering for tile load events on `ol.source.VectorTile`, load events were fired for `ol.VectorTile` instances. Now they are fired with `ol.VectorImageTile` instances, which is problematic because these tiles neither match the `tileGrid` that the source is configured with, nor the tiles that are passed to the `tileLoadFunction` of the source.
2. Make `ol.VectorTile#getFeatures` and `#getProjection` exportable.
3. Remove `ol.VectorImageTile` from the API.

**Background:**
For applications that need to work with vector tile features/geometries, it is currently impossible to do so. By monitoring tile loads, and having access to the tile grid and each tile's features, it should be possible to transform feature geometries from tile pixels to view coordinates. A possible workflow would be the following:
```js
var tiles = {};
vectorTileSource.on('tileloadend', function(e) {
  tiles[e.tile.getTileCoord()] = e.tile;
});
// Now we know the loaded tiles, let's get the geometries inside the viewport
vectorTileSource.getTileGrid().forEachTileCoord(view.calculateExtent(), Math.round(view.getZoom()), function(tileCoord) {
  var tile = tiles[tileCoord];
  var features = tile.getFeatures();
  var projection = tile.getProjection();
  if (projection.getUnits() == 'tile-pixels') {
    // Transform feature geometry coordinates
    // Do something with geometries
  }
});
```

### DIFF
--- a/src/ol/renderer/canvas/vectortilelayer.js
+++ b/src/ol/renderer/canvas/vectortilelayer.js
@@ -108,7 +108,8 @@ ol.renderer.canvas.VectorTileLayer.prototype.createReplayGroup_ = function(
   var pixelRatio = frameState.pixelRatio;
   var projection = frameState.viewState.projection;
   var revision = layer.getRevision();
-  var renderOrder = layer.getRenderOrder() || null;
+  var renderOrder = /** @type {ol.RenderOrderFunction} */
+      (layer.getRenderOrder()) || null;
 
   var replayState = tile.getReplayState();
   if (!replayState.dirty && replayState.renderedRevision == revision &&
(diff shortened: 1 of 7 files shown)


ahocevar:
Updated description to better explain what this pull request does.

shunshizhechang:
I want to get features from vectortile and get the features's coordinate。I try this way,but failed.
Can you give me some example?

ahocevar:
Sorry, I don't have a complete example. If you share your attempt in a codepen, I can help you make it work though.

PR #6832, opened 2017-05-19
Force state of error tiles when usInterimTilesOnError is false

### LINKED ISSUE #5269: Interim tiles shown when we switch time dimension
Here an [example](http://codepen.io/oterral/pen/BKOwLZ), Click on play button.

The problem is the tiles of 2013 are shown when the layer switch between year 1944 to 1947.

I think the interim tiles are displayed when there is no tiles requested at the same place, even when we use `useInterimTilesOnError: false`

Any way to avoid this flicker?

oterral:
Attempt to fix #5269 .

[Bug on codepen](http://codepen.io/oterral/pen/BKOwLZ)
[Bug fixed on codepen](http://codepen.io/oterral/pen/EmdOZW)

Just push the play button.

### DIFF
--- a/src/ol/imagetile.js
+++ b/src/ol/imagetile.js
@@ -94,6 +94,7 @@ ol.ImageTile.prototype.getKey = function() {
  */
 ol.ImageTile.prototype.handleImageError_ = function() {
   this.state = ol.TileState.ERROR;
+  this.image_ = ol.ImageTile.blankImage;
   this.unlistenImage_();
   this.changed();
 };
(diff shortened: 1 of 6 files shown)


src/ol/imagetile.js
```diff
@@ -94,6 +95,10 @@ ol.ImageTile.prototype.getKey = function() {
  */
 ol.ImageTile.prototype.handleImageError_ = function() {
   this.state = ol.TileState.ERROR;
+  var ctx = ol.dom.createCanvasContext2D(1, 1);
+  ctx.fillStyle = 'rgba(0,0,0,0)';
+  ctx.fillRect(0, 0, 1, 1);
+  this.image_ = ctx.canvas;
```
tschaub: We could also use a single static transparent image.  Maybe something like:

```js
ol.ImageTile.BlankImage = new Image();
ol.ImageTile.BlankImage.src = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
```

And then use that image in the error handler.

oterral: Yes good idea, thx

oterral:
PR updated according to  @tschaub comment

gjn:
@tschaub @fredj Is this good to merge. It would help us fix some bugx in map.geo.admin.ch

PR #6834, opened 2017-05-20
Move tolerance option

### LINKED ISSUE #6831: Option for adjusting move distance tolerance
If a users pan the map when clicking the click will be parsed as a move instead of a click. A group of our mobile users have problem with clicking on the map without triggering a move. Our suggested solution is to make the `moveTolerance` variable in `ol.MapBrowserEventHandler` configurable via a `moveTolerance` option in `ol.Map`.

Proposed implementation:
https://github.com/notnotse/openlayers/tree/move-tolerance-option

itjope:
Make the `moveTolerance` variable in `ol.MapBrowserEventHandler` configurable via a `moveTolerance` option in `ol.Map`

Fix #6831

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -292,6 +292,7 @@ olx.interaction.InteractionOptions.prototype.handleEvent;
  *     loadTilesWhileAnimating: (boolean|undefined),
  *     loadTilesWhileInteracting: (boolean|undefined),
  *     logo: (boolean|string|olx.LogoOptions|Element|undefined),
+ *     moveTolerance: (number|undefined),
  *     overlays: (ol.Collection.<ol.Overlay>|Array.<ol.Overlay>|undefined),
  *     renderer: (ol.renderer.Type|Array.<ol.renderer.Type>|undefined),
  *     target: (Element|string|undefined),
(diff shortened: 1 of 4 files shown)


bjornharrtell:
Nice work @itjope, I think would be useful to have this in.

ahocevar:
Thanks for this contribution @itjope. Can you please submit a [CLA](https://github.com/openlayers/cla/) so we can merge this?

externs/olx.js
```diff
@@ -386,6 +387,17 @@ olx.MapOptions.prototype.logo;
 
 
 /**
+ * The minimum distance in pixels the cursor must move to be detected
+ * as a map move event instead of a click. Increasing this value can make it
+ * easier to click on the map.
+ * Default is `1`.
```
fredj: Default is `window.devicePixelRatio`.

ahocevar: > Default is window.devicePixelRatio.
Depends on how we define 'pixel'. Elsewhere, e.g. for `ol.Map#getSize`, we use JS pixels. So I'd prefer to leave this as `1`.

ahocevar: Actually the default is `1`, because the value given here is multiplied with the device pixel ratio.

fredj: you're right, the default is `1`, sorry for the noise

itjope:
@ahocevar  I have now sent a scanned signed copy of the CCLA.

TheNaderio:
Is this tolerance available in OpenLayers.Map()? (2.13)

PR #6842, opened 2017-05-23
add condition to check if active

### LINKED ISSUE #6840: broken setActive on ol.interaction.DragAndDrop
"setActive" on most of ol.interaction works fine but on ol.interaction.DragAndDrop, it seems it doesn't work. 
You can keep drag and drop your file after deactivate DragAndDrop interaction.
added deactive at the bottom of the code.
[codepen](https://codepen.io/anon/pen/wdRwpr?editors=0010)

cs09g:
#6840
"setMap" registers listeners to dropArea, but it doesn't check whether "active" is true or false. Simply add condition to check "active".


cs09g:
Thank you, I submitted CLA.

PR #6849, opened 2017-05-24
Create URL object only when we can

### LINKED ISSUE #6848: KML parsing failed when node.baseURI = 'about:blank'
On latest Safari (Mac, iOS) parsing of `styleUrl` tag trigger a typeError  [here](https://github.com/openlayers/openlayers/blob/master/src/ol/format/kml.js#L492) because `node.baseURI = 'about:blank'` instead of null.

Generally speaking `new URL(a, b)` will trigger an error if the two parameters are not real url.

oterral:
Fix #6848

### DIFF
--- a/src/ol/format/kml.js
+++ b/src/ol/format/kml.js
@@ -488,7 +488,7 @@ ol.format.KML.readFlatCoordinates_ = function(node) {
  */
 ol.format.KML.readURI_ = function(node) {
   var s = ol.xml.getAllTextContent(node, false).trim();
-  if (node.baseURI) {
+  if (node.baseURI && node.baseURI !== 'about:blank') {
     var url = new URL(s, node.baseURI);
     return url.href;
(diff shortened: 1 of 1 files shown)


src/ol/format/kml.js
```diff
@@ -488,7 +488,7 @@ ol.format.KML.readFlatCoordinates_ = function(node) {
  */
 ol.format.KML.readURI_ = function(node) {
   var s = ol.xml.getAllTextContent(node, false).trim();
-  if (node.baseURI) {
+  if (node.baseURI && node.baseURI !== 'about:blank') {
```
ahocevar: Strange that Safari now returns `'about:blank'` as baseURI, but I agree that this is the most pragmatic way to work around the issue.

oterral:
thx for the quick megre @ahocevar

Issue #6311, opened 2016-12-28
Map defaults should be optional not required


probins:
I've no problem with sensible defaults, but atm map.js requires the defaults, so in ES6 terms they are static dependencies, always imported whether they're used or not. But the map is basically just a container, and should have few if any dependencies. The renderer is currently the biggest issue here, as it involves by far the largest amount of code. This is the subject of #6303, but this still leaves the default controls, interactions, and projections. A renderer is always needed and it's unlikely that people would want to write their own, but people should be able to use their own UI widgets if they wish, and they should be able to, for example, exclude support for keyboard and mousewheel for devices which don't have such things. proj.common seems particularly illogical to me, as proj4js is enabled by default and includes support for 4326<->3857; adding proj.common just duplicates this code. In any case, if I don't have any 3857 sources, why am I required to include the code to handle/transform it?

My suggestion for removing these requires while keeping bc would be to split map.js in two as described in https://github.com/openlayers/ol3/issues/6303#issuecomment-269210698, but if anyone has any better ideas?

PR #6855, opened 2017-05-28
Pluggable renderers

### LINKED ISSUE #6303: Explicitly require the renderer

tschaub:
### The Problem

We have a lot of unnecessary dependencies within the library.

For example, `ol.layer.Tile` [depends on both](https://github.com/openlayers/openlayers/blob/29a6473c3dfd7f97ecf84c77dcc9391033217dfd/src/ol/layer/tile.js#L8-L9) `ol.renderer.canvas.TileLayer` and `ol.renderer.webgl.TileLayer`.  Same situation with the rest of the layers.

And `ol.Map` [depends on both](https://github.com/openlayers/openlayers/blob/29a6473c3dfd7f97ecf84c77dcc9391033217dfd/src/ol/layer/tile.js#L8-L9) `ol.renderer.canvas.Map` and `ol.renderer.webgl.Map`.

The map also [depends on](https://github.com/openlayers/openlayers/blob/29a6473c3dfd7f97ecf84c77dcc9391033217dfd/src/ol/control.js#L4-L6) `ol.control.Attribution`, `ol.control.Zoom`, and `ol.control.Rotate`.  And a [whole lot of interactions](https://github.com/openlayers/openlayers/blob/29a6473c3dfd7f97ecf84c77dcc9391033217dfd/src/ol/interaction.js#L4-L13).

The Compiler and build flags can ameliorate some of this, but we can improve the situation.

### Proposed Solution

We can treat map and layer renderers as plugins.  To maintain backwards compatibility (and for convenience), `ol.Map` can register both map renderers and all the layer renderers.

In addition, `ol.Map` can continue to use the `defaults()` functions for controls and interactions (and these will retain their dependencies).

Then, for people who don't need all the extra stuff (and don't mind a bit of inconvenience), we can create a `ol.PluggableMap` (and `ol.Map` will extend this).  People using `ol.PluggableMap` will register just the map renderer and layer renderers that they need.  And they can provide controls and interactions directly (by default, `ol.PluggableMap` comes with no controls or interactions).  In this way, all dependencies will be explicit.  No build flags will be necessary and other minifiers can be used.

Fixes #6303.


tschaub:
This works, but I need to create some `olx` types to keep the Compiler from renaming the static methods.

bartvde:
interesting @tschaub thanks for taking this forward

probins:
Presumably, this replaces #6303? (And fixes much of #6311.)

A dramatic reduction in the number of unwanted dependencies, and way more flexible than the current setup. Once this passes Travis, I'll pull in the branch and try some tests, and also look at the new dependency tree.

I'm not sure about the `PluggableMap` name, which doesn't seem very descriptive to me, as the only 'pluggables' are the renderers. It's essentially a map without any defaults, so perhaps `BaseMap` (or 'raw' or 'bare' or something) would be better?

Something that's missing here is documentation, particularly for those wanting to use `PluggableMap`. The custom build tutorial has always stated that you list the classes/methods you use in your app in your build config, but the renderers aren't referred to in apps, and do not appear in the docs anywhere, so people have to be told what to include. I notice that `api` (stable?) has been added to the renderer constructors, but they have no docs text. ISTM there should be some examples of using `PluggableMap` too, though that could be in a separate PR.

probins:
also, it would be good if `View` could be included in this, as the `fit` method is currently 'requiring' a lot of vector files, which are only needed if you use a geometry with `fit()`, probably only used by a small %age of maps. `view.js` and `map.js` can remain as is for bc. View is only included in `map.js` to create a default view if none is provided in the constructor options, but I can use `setView()` at some later time, so is this actually necessary? This view creation can be omitted from `PluggableMap`, and if a View _is_ needed at construction time, then those using `PluggableMap` can be told they have to supply one in the options.

If we then create `viewwithoutgeoms.js`, and put `fit()` (or at least the geom bit) in a separate file, then `PluggableMap` users can use the no-geoms view, and add `fit`/geoms separately if they wish. This would then enable people to have maps without vectors if they want.

tschaub:
For people who want to explicitly depend on just what they use, there is now a `PluggableMap` constructor.  To use it, you first need to register the map and layer renderers that you use.  It also doesn't come with any interactions or controls, so you add those explicitly.  For example:
```js
import Map from 'ol/pluggablemap';
import MapRenderer from 'ol/renderer/canvas/map';
import PluginType from 'ol/plugintype';
import TileLayer from 'ol/layer/tile';
import TileLayerRenderer from 'ol/renderer/canvas/tilelayer';
import View from 'ol/view';
import XYZSource from 'ol/source/xyz';
import control from 'ol/control';
import interaction from 'ol/interaction';
import plugins from 'ol/plugins';

plugins.register(PluginType.MAP_RENDERER, MapRenderer);
plugins.register(PluginType.LAYER_RENDERER, TileLayerRenderer);

new Map({
  target: 'map-container',
  interactions: interaction.defaults(),
  controls: control.defaults(),
  renderer: MapRenderer,
  layers: [
    new TileLayer({
      source: new XYZSource({
        url: 'http://tile.stamen.com/terrain/{z}/{x}/{y}.jpg'
      })
    })
  ],
  view: new View({
    center: [0, 0],
    zoom: 2
  })
});
```

For a bit of convenience, there is now a `CanvasMap` constructor.  So to get just the canvas renderer (this still depends on all canvas layer renderer types and you get the default controls and interactions), you can do this:
```js
import Map from 'ol/canvasmap';
import View from 'ol/view';
import Tile from 'ol/layer/tile';
import XYZ from 'ol/source/xyz';

new Map({
  target: 'map-container',
  layers: [
    new Tile({
      source: new XYZ({
        url: 'http://tile.stamen.com/terrain/{z}/{x}/{y}.jpg'
      })
    })
  ],
  view: new View({
    center: [0, 0],
    zoom: 2
  })
});
```

Both examples above will be possible with the next release.

nboisteault:
@tschaub Hi, I don't find `ol/plugins` nor `ol/PluginType` in OL6. Can you update your code above for OL6 please ? Thanks.

MoonE:
Have a look at the `ol/Map` code, there is now a `createRenderer` method instead.
https://github.com/openlayers/openlayers/blob/main/src/ol/Map.js

nboisteault:
@MoonE Thanks for the hint but if I load `ol/Map` instead of `ol/PluggableMap`, how do I avoid `module:ol/interaction~defaults` in final bundle?

ahocevar:
@nboisteault You can continue using `ol/PluggableMap`. Just make sure you implement the `createRenderer` method, e.g. like in https://github.com/openlayers/openlayers/blob/2cac6c4/src/ol/Map.js#L75-L77

nboisteault:
Thanks @ahocevar. What I did if it can help someone : 

```javascript
import PluggableMap from 'ol/PluggableMap';
import CompositeMapRenderer from 'ol/renderer/Composite';
import { assign } from 'ol/obj';

class olMap extends PluggableMap {
    /**
     * @param {import("./PluggableMap.js").MapOptions} options Map options.
     */
    constructor(options) {
        options = assign({}, options);

        super(options);
    }

    createRenderer() {
        return new CompositeMapRenderer(this);
    }
}

const myMap = new olMap({...});
```

PR #6858, opened 2017-05-29
Webgl vector improvements

### LINKED ISSUE #6823: Webgl polygon rendering
I have a simple map with one feature, if I switch renderer to canvas, everything is ok, but if I switch renderer to webgl, feature is hidden and sometimes browser freeze. Is it a problem with bad feature or with rendering?
[renderer.zip](https://github.com/openlayers/openlayers/files/1007630/renderer.zip)

GaborFarkas:
Some minor improvements in rendering vectors with the WebGL renderer.

- Fixes #6823's rendering problems.
- Makes WebGL vector replays compatible with `ol.render.Feature`.
- Maps flat coordinates to numbers (see also #6823).

### DIFF
--- a/src/ol/render/webgl/linestringreplay.js
+++ b/src/ol/render/webgl/linestringreplay.js
@@ -324,16 +324,19 @@ if (ol.ENABLE_WEBGL) {
    */
   ol.render.webgl.LineStringReplay.prototype.drawMultiLineString = function(multiLineStringGeometry, feature) {
     var indexCount = this.indices.length;
-    var lineStringGeometries = multiLineStringGeometry.getLineStrings();
+    var ends = multiLineStringGeometry.getEnds();
+    ends.unshift(0);
+    var flatCoordinates = multiLineStringGeometry.getFlatCoordinates();
+    var stride = multiLineStringGeometry.getStride();
     var i, ii;
-    for (i = 0, ii = lineStringGeometries.length; i < ii; ++i) {
-      var flatCoordinates = lineStringGeometries[i].getFlatCoordinates();
-      var stride = lineStringGeometries[i].getStride();
-      if (this.isValid_(flatCoordinates, 0, flatCoordinates.length, stride)) {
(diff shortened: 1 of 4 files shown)


tschaub:
How did the string values get it the flat coordinates?

GaborFarkas:
I tried to trace it back, but failed to catch anything suspicious. Thus, I have absolutely no idea.

GaborFarkas:
How should we deal with the rendering tests? There are different threshold errors on my local CI test. Should we raise the threshold? Then the WebGL rendering tests will be practically useless (not that they are so useful in their current form).

gberaudo:
@GaborFarkas, regarding the failing test, there is actually a noticeable difference in the rendered image:
image
instead of the expected:
image

GaborFarkas:
Than you @gberaudo, I must have tested another branch.

GaborFarkas:
This PR is ready for review.

fredj:
@GaborFarkas about `.map(Number)`, in what condition you had non number values ? (example, unit test, ...)

GaborFarkas:
@fredj with the example data @ondrejlaga uploaded in issue #6823. When you zoom around that polygon, it disappears on higher zoom levels. If you log the flat coordinates in debug mode, you will notice, when the polygon disappears, the coordinates are strings.

GaborFarkas:
Thank you @gberaudo for your review!

fredj:
@GaborFarkas that's because the data in #6823 is not valid GeoJSON; the coordinates are encoded as string instead of numbers (`["-585733.80000000", "-1138268.90000000"]` instead of `[-585733.80000000, -1138268.90000000]`)

I've uploaded a fixed version:
  https://gist.github.com/fredj/ea28e945e49bc93e26339fdafb99c300

Please test with this file and remove ea4d451 if it works

GaborFarkas:
Thank you @fredj, it works as expected with your GeoJSON. Before merging I would like to fix another edge case I just noticed. I will remove ea4d451 in the final rebase.

GaborFarkas:
Sorry for the mess. Now the PR is ready for review again.

Issue #6862, opened 2017-05-30
Can not flyTo animation with View.fit


cs09g:
as mentioned #6170, View.fit only takes `duration` and `easing` for animation effect. in this case, we can panTo but flyTo because there's no `zoom`.

tschaub:
This is true.  The way to put together a "fly to" type animation is with `view.animate()`.  As mentioned in https://github.com/openlayers/openlayers/issues/6170#issuecomment-304855656, the view doesn't expose all the methods necessary to get the target resolution for a geometry (or extent).

If you'd like to see `view.getResolutionForExtent()` exposed, please create a ticket for that.

Turbo87:
@cs09g it looks like you were successful in animating `view.fit()` with a "fly to" animation. do you have a code example on how you did it?

cs09g:
@Turbo87 Get the resolution for the extent, then you can use `animate` with middle point of the path. something like this. but I decided to use linear animation but fly animation in order to flyTo should be considered about low to high and high to low case.
![image](https://user-images.githubusercontent.com/7428017/57054171-ce7e7400-6ccd-11e9-8995-563d1308cefc.png)

To resolve this, there should be complex calculation like https://www.win.tue.nl/~vanwijk/zoompan.pdf which is used in mapboxgl.
```javascript
// this looks good for the same resolution(zoom) moving
resolution = map.getView().getResolutionForExtent(extent);
centerFrom = <currentCenter>;
centerTo = <centerOfExtent>;
duration = <durationForAnimation>;

map.getView().animate({
  resolution: resolution * 4, // jump effect
  center: [
    (centerFrom[0] + centerTo[0]) / 2,
    (centerFrom[1] + centerTo[1]) / 2
  ],
  duration: duration / 2
}, {
  resolution: resolution, // back again
  center: centerTo,
  duration: duration / 2
}
```

PR #6863, opened 2017-05-31
change size to optional

cs09g:
#6170 #6862
To use flyTo effect for extent, `View.fit` is not suitable. 
`getResolutionForExtent` with `constrainResolution` makes flyTo possible.

### DIFF
--- a/src/ol/view.js
+++ b/src/ol/view.js
@@ -662,11 +662,13 @@ ol.View.prototype.getResolutions = function() {
 /**
  * Get the resolution for a provided extent (in map units) and size (in pixels).
  * @param {ol.Extent} extent Extent.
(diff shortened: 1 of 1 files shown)


src/ol/view.js
```diff
@@ -662,11 +662,13 @@ ol.View.prototype.getResolutions = function() {
 /**
  * Get the resolution for a provided extent (in map units) and size (in pixels).
  * @param {ol.Extent} extent Extent.
- * @param {ol.Size} size Box pixel size.
+ * @param {ol.Size} opt_size Box pixel size.
```
fredj: please change the type to `{ol.Size=}`

cs09g: Thanks for the review,

Issue #6854, opened 2017-05-26
ol.extent.getArea is not in the Api


polar:
The functions ol.extent.getWidth and ol.extent.getHeight and ol.extent.isEmpty are in the api, but why not ol.extent.getArea?

Cheers,
-Polar

cs09g:
Can I create a PR?

fredj:
@cs09g yes, please create a PR

polar:
Thanks!

Cheers,
-Polar

PR #6865, opened 2017-05-31
add getArea to api

cs09g:
#6854


cs09g:
Opps. mistakes on git command. will do new PR.

PR #3690, opened 2015-10-13
Remove the mouse only limitation for ol.interaction.DragBox

fredj:
If possible, remove the hard coded mouse only limitation for `ol.interaction.DragBox`, `ol.interaction.DragRotateAndZoom` and `ol.interaction.DragRotate`.

See https://groups.google.com/d/msg/ol3-dev/9ojeuOXXREQ/iceX5ODQC4oJ

PR #6868, opened 2017-06-01
Remove the mouse only limitation for ol.interaction.DragBox

tchandelle:
Following up on #3690, and add the ability for `DragBox` to handle 2 pointers.

However, I didn't think it was useful to remove the mouse-only limitation on `DragRotate` and `DragRotateAndZoom`, since `PinchZoom` and `PinchRotate` work.


ahocevar:
@bartvde @fredj Could either of you please review this, since you were both involved in #3690? Thanks.

PR #6874, opened 2017-06-05
Remove unused var from ol.proj

probins:
A minor tidy-up I noticed when looking at proj. `addProjections()` used to return the added projections, but looks like it stopped doing so in #1228, so the variable is no longer used.


probins:
you're right. Changed

fredj:
Thanks

PR #6876, opened 2017-06-06
Only test ol.interaction.DragAndDrop if FileReader is supported

fredj:


ahocevar:
I'm wondering about the decreased test coverage. Could it be possible that the feature detection returns a false negative for FileReader? Because it looks like Coveralls was previously able to run the tests, and does not run them any more.

fredj:
In phantomjs, `typeof FileReader` is `"object"`, not `"function"`.
Same goes for `typeof ArrayBuffer`.

fredj:
I will change the test to `'FileReader' in window`

fredj:
Thanks for the review

PR #6884, opened 2017-06-08
Added the srs property to the js object that is the result of the par…

BasieP:
…sing of the xml capabilities.

see: http://docs.geoserver.org/stable/en/user/services/wms/reference.html

### DIFF
--- a/src/ol/format/wmscapabilities.js
+++ b/src/ol/format/wmscapabilities.js
@@ -96,6 +96,7 @@ ol.format.WMSCapabilities.readBoundingBox_ = function(node, objectStack) {
 
   return {
     'crs': node.getAttribute('CRS'),
+    'srs': node.getAttribute('SRS'),
     'extent': extent,
     'res': resolutions
   };


BasieP:
oke i've been working for more then 20  hours now to get the tests running on my windows (cygwin) machine but it simply won't work..
so sorry i can't make a pull request that fixes the build. if you require it of me, please write a decent guide, cause so far the 10 lines written here: https://github.com/openlayers/openlayers/blob/master/CONTRIBUTING.md#building-on-windows are simply not working (i get dependency errors)

ontopic:
I want to use for example this WMS server:
https://geodata.nationaalgeoregister.nl/ahn2/wms?request=GetCapabilities&version=1.1.1
it doesn't return an xml with CRS fields, but with SRS fields.
It changed in version 1.3 but in older versions of the WMS protocol it still says SRS.
If openlayers is providing a parser (which it does) it beter support both versions.

I love to fix the problem, but if you guys don't want it fixed i don't start on it, cause it will be a waste of my time...

bartvde:
WMS Capabilities only supports 1.3.0

bartvde:
Your service supports WMS 1.3.0 so why can't you use that version?

bartvde:
Also see https://github.com/openlayers/openlayers/pull/2944 which got stalled

BasieP:
yeah that is certain...

the point is that i can fix that.......

ahocevar:
Instead of this pull request, maybe a doc improvement pull request that states the supported version(s) of our WMS Capabilities parser would be the right thing to do.

BasieP:
or throw an exception when the version is not high enough... cause right now it parses perfectly.. just with empty fields..

PR #6886, opened 2017-06-08
Make ol.source.Cluster extensible

gberaudo:
Change fields and methods from private to protected to allow extension. The refresh method now calls the super method.


gberaudo:
Fixed the tests.

gberaudo:
Thanks for the review @fredj. Can you please merge it?

PR #6887, opened 2017-06-08
Disable rotation for views with enableRotation: false

ahocevar:
This pull request makes configuration of maps without rotation capability easier. Previously, interactions had to be configured manually. Now, all it takes is to set `enableRotation: false` on the view.

This pull request also leverages a typedef for storing the constraints on the view, and gets rid of `ol.Constraints`.


fredj:
`enableRotation` is never read in `ol.View` and `ol.RotationConstraint.disable` is not defined

ahocevar:
@fredj:

> `enableRotation` is never read in ol.View

It is read and used [here](https://github.com/openlayers/openlayers/blob/5475841ca022b7ab923566e4008e33652b38f2fb/src/ol/view.js#L1124-L1137).

> `ol.RotationConstraint.disable` is not defined

It is defined [here](https://github.com/openlayers/openlayers/blob/535e8f5e35e4f935cfd297acd9bd6da5e5d29558/src/ol/rotationconstraint.js#L11-L17).

fredj:
oups, I didn't remember that `enableRotation` is already here ...

ahocevar:
Thanks for the review @fredj.

PR #6890, opened 2017-06-08
Add a movestart event

### LINKED ISSUE #3985: Add 'movestart' event on map.
Just like there is a 'moveend' event available on map, it would be very convenient to have a `movestart` event.

**Use Case**
In case when triggering a `ol.animation.pan` I may want to toggle once before the animation starts and toggle it back once after animation ends. 
We have to remember here that animation might not actually happen based on where we are panning to. So just doing something before triggering pan is not enough, because toggle back might not happen if it doesn't actually pan (`moveend') doesn't get called in this event. 

**fix 1**
This can be fixed , if `moveend` is called even if the actual move doesn't happen, if pan was triggered. 

**fix2**
`movestart` event should be provided which just like `moveend` _doesnt_ get called if there is no movement happening in the map. 

Currently it is possible to achieve this using `map.beforeRender` but, it just seems too much ugly coding to achieve something very simple. 

*\* current  solution**

``` javascript
           var pan = ol.animation.pan({
                duration: 700,
                source: map.getView().getCenter()
            });

            map.beforeRender(function(map, frameState) {
              var stillPanning = pan(map, frameState); // returns false panning is done

              if (stillPanning){
                 // do movestart stuff here
                 if(!everDone){
                       doSomething();
                       everDone = true;
                  }
              }
              else{
                 // do move end stuff here
                 undoSomething();
                 everDone = false; 
              }              
              return stillPanning;
            });

            map.getView().setCenter(geom);    
```

This could all have been done in two simple  `map.on("movestart")` and  `map.on("moveend")` callbacks like it was supported in OpenLayers2. 

Here's a [jsFiddle](http://jsfiddle.net/j2cP4/114/) that demonstrates what I need to do with the non-ideal solution I posted.  Try pressing the pan-map second time and see how `moveend` does not fire, but animation end does. 
## update

I have updated my original question and answer on stackoverflow to summarize the discussion here: 

[http://stackoverflow.com/questions/31896300/openlayers-3-movestart-event-on-map/31897512](http://stackoverflow.com/questions/31896300/openlayers-3-movestart-event-on-map/31897512).
Thanks!

ahocevar:
Fixes #3985.

### DIFF
--- a/src/ol/map.js
+++ b/src/ol/map.js
@@ -210,7 +210,7 @@ ol.Map = function(options) {
    * @private
    * @type {ol.Extent}
    */
-  this.previousExtent_ = ol.extent.createEmpty();
+  this.previousExtent_ = null;
 
   /**
    * @private
@@ -1200,6 +1200,7 @@ ol.Map.prototype.renderFrame_ = function(time) {
   var size = this.getSize();
   var view = this.getView();
   var extent = ol.extent.createEmpty();
+  var previousFrameState = this.frameState_;
   /** @type {?olx.FrameState} */
   var frameState = null;
   if (size !== undefined && ol.size.hasArea(size) && view && view.isDef()) {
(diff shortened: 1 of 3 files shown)


ahocevar:
Thanks for any review.

src/ol/map.js
```diff
@@ -1201,6 +1201,8 @@ ol.Map.prototype.renderFrame_ = function(time) {
   var view = this.getView();
   var extent = ol.extent.createEmpty();
   /** @type {?olx.FrameState} */
```
fredj: Minor but I'm wondering if this type cast is needed (for `var previousFrameState = this.frameState_;` )

Issue #6746, opened 2017-04-25
setMinZoom, setMaxZoom doesn't work when view has resolution property


cs09g:
If you give resolutions property when you create view for map, setMinZoom and setMaxZoom doesn't work. 

example's [here](https://codepen.io/anon/pen/ybaWXv?editors=0011)

cs09g:
What about to have setMinResolution/setMaxResolution or setResolutions likewise getResolutions

cs09g:
any comment please?

ahocevar:
If your goal is to fix `setMinZoom()` and `setMaxZoom()` for the case where the view was constructed with `resolutions`, then please submit a pull request.

cs09g:
Right, setting minimum/maximum zoom is the goal. not restriction of resolutions. I'll have a look. Thanks.

cs09g:
fixed with #7521

PR #6898, opened 2017-06-09
fix setMinZoom/setMaxZoom

cs09g:
#6746


cs09g:
finally ready for review

Issue #6885, opened 2017-06-08
Suggestions for small improvements to ol.proj


probins:
1. simplify handling of sphere radii. `ol.proj.Units` and `ol.proj.EPSG4326` are currently using `ol.sphere` objects to get a radius. This means importing a file which creates a sphere object for a particular radius and then using the radius of that object. This seems an over-complicated way of defining a constant to me, and I would suggest just defining the radius in `ol.proj.Units` and `ol.proj.EPSG4326`, removing the dependency on `ol.sphere`. If these constants were widely used in the library, you could store them in one place, perhaps in a `SPHERERADIUS` enum, but they're not.
2. consistent use of projection code/object in params:
    * docs say apps don't need to use `Projection` objects, as `ProjectionLike` can be used in all options instead. But this is not true of `getPointResolution`, so I'd suggest changing this. Strictly speaking, I suppose this is an api change, but it shouldn't break anything.
    * `proj.transforms` is keyed on projection code, which is also used by `get()`, but `add()` and `remove()` use a Projection object. This is not api, so won't affect any apps, but it seems unnecessarily confusing to me, and I would suggest changing `add`/`remove` to use code as well.

ahocevar:
+1 on everything except the last item (use projection code for adding/removing transforms). Are you able to provide a pull request?

fredj:
`ol.control.MousePosition` and `ol.Geolocation` could also accept `ol.ProjectionLike`, see #6892

probins:
@ahocevar what don't you like about the last item? (not a major issue, just interested)

ahocevar:
@probins: Not that it adds much overhead, but do not use `ol.proj.get` much internally.

probins:
I'll close this, as most of the issues now fixed

PR #6899, opened 2017-06-09
Use number literal for sphere radius

probins:
See #6885. I could create a const for this, but as this is purely internal I think a literal with a comment should be enough.

### DIFF
--- a/src/ol/proj/epsg3857.js
+++ b/src/ol/proj/epsg3857.js
@@ -31,6 +31,8 @@ ol.inherits(ol.proj.EPSG3857.Projection_, ol.proj.Projection);
 
 
 /**
+ * Radius of WGS84 sphere
+ *
  * @const
  * @type {number}
  */
--- a/src/ol/proj/epsg4326.js
+++ b/src/ol/proj/epsg4326.js
@@ -3,7 +3,6 @@ goog.provide('ol.proj.EPSG4326');
 goog.require('ol');
 goog.require('ol.proj.Projection');
 goog.require('ol.proj.Units');
-goog.require('ol.sphere.WGS84');
 
 
 /**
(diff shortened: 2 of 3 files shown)


ahocevar:
I'm wondering if it wouldn't make sense to get rid of `ol.sphere.WGS84` and `ol.sphere.NORMAL` entirely. Instead, `ol.Sphere#haversineDistance`, `ol.Sphere#offset` and `ol.Sphere#geodesicArea` could become static functions with a `radius` argument in the `ol.sphere` namespace. `ol.Sphere` could be marked `@deprecated` and removed in 5.0.

What do others think?

fredj:
+1 to have static functions. That will be in line with what we already doing in `ol.extent.*`, `ol.size.*`, ...

probins:
+1

I used to use `ol.sphere.WGS84` in my own apps, but changed to static functions quite some time ago. Two examples use it, but they can easily be changed.

Once ES modules are more widely used, there will probably be freely available modules containing this sort of utility function; then OL won't need to provide its own version.

probins:
just noticed that `proj.EPSG3857`, which also uses this radius, defines it as a constant. It would be better for both this and 4326 to do the same thing.

tschaub:
> What do others think?

@ahocevar - I agree it would be nice to provide a convenient way to do spherical calculations on ol geometries.  Probably best for another ticket, but I like the idea of functions with convenient defaults.  E.g.

```
ol.sphere.getArea(geometry, options);
ol.sphere.getLength(geometry, options);
ol.sphere.getDistance(geometry1, geometry2, options);

// options.projection defaults to 'EPSG:3857'
// options.radius defaults to the WGS:84 radius
```

probins:
amended so that 4326 uses a constant as 3857 does (and added a comment to 3857)

@ahocevar I too would prefer to get rid of `ol.Sphere`, but I think that's a separate issue/PR. The aim of this PR is to remove a couple of unnecessary dependencies.

PR #6902, opened 2017-06-10
Require minimum duration for kinetic animation

### LINKED ISSUE #6861: Map freezing when pinch zooming on ipad
We are getting this error on IOS 10.3.2 on safari, trying to migrate an app from 3.19.1 to 4.1.1, and when making fast pinch zoom in and out. All the map disappear (only the overlays stay there) and the map freeze completely. 

We also get this unhandled exception from webkit : DOM Exception 11 : The object is in an invalid state

From the stack trace (of the minified code) the offending line is the drawImage() call from this function :
 `function qv(a) { zt.call(this, a); this.u = yh(); this.j = null } u(qv, zt); qv.prototype.O = function (a, b, c) {
    Bt(this, "precompose", c, a, void 0);
    var d = this.Y();
    if (d) {
        var e = b.extent, f = void 0 !== e && !Ua(e, a.extent) && nb(e, a.extent);
        f && At(c, a, e);
        var e = this.A(), g = c.globalAlpha;
        c.globalAlpha = b.opacity;
        c.drawImage(d, 0, 0, +d.width, +d.height, Math.round(e[4]), Math.round(e[5]), Math.round(d.width * e[0]), Math.round(d.height * e[3]));
        c.globalAlpha = g;
        f && c.restore()
    } this.cf(c, a, b)
};`

For now i'm unable to repro this in a jsfiddle.

ahocevar:
Fixes #6861. Thanks @ogo-adp for the research and the suggested fix.

### DIFF
--- a/src/ol/kinetic.js
+++ b/src/ol/kinetic.js
@@ -94,7 +94,15 @@ ol.Kinetic.prototype.end = function() {
   while (firstIndex > 0 && this.points_[firstIndex + 2] > delay) {
     firstIndex -= 3;
   }
+
   var duration = this.points_[lastIndex + 2] - this.points_[firstIndex + 2];
+  // we don't want a duration of 0 (divide by zero)
+  // we also make sure the user panned for a duration of at least one frame
+  // (1/60s) to compute sane displacement values
+  if (duration < 1000 / 60) {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


tschaub:
Looks sensible.

PR #6908, opened 2017-06-12
Change overlay positioning to use percentage of mapSize

fmcd:
Change overlay positioning to use percentage of mapSize rather than absolute pixel values.

This resolves overlay position appearing differently in print preview
or when printing from the browser. I believe the problem is the rasterization of
the image for printing causes a different effective pixel size for the map,
but no javascript events fire during printing or print preview preparation that
allow map to be reloaded or overlays repositioned.

By positioning overlay using percentage of the parent, it avoids this problem and
the print preview in the browser shows the correct overlay position.

### DIFF
--- a/src/ol/overlay.js
+++ b/src/ol/overlay.js
@@ -475,7 +475,7 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
     if (this.rendered_.left_ !== '') {
       this.rendered_.left_ = style.left = '';
     }
-    var right = Math.round(mapSize[0] - pixel[0] - offsetX) + 'px';
+    var right = (Math.round(mapSize[0] - pixel[0] - offsetX) * 100 / mapSize[0]) + '%';
     if (this.rendered_.right_ != right) {
       this.rendered_.right_ = style.right = right;
     }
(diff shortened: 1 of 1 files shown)


ahocevar:
As an alternative, you could try the following in your application:
```
if ('matchMedia' in window) {
  window.matchMedia('print').addListener(function (media) {
    map.updateSize(); map.renderSync();
  });
} else {
  window.onbeforeprint = function () {
    map.updateSize(); map.renderSync();
  }
}
```

ahocevar:
@fmcd, have you tried the approach I suggested above?

ahocevar:

sapetoku17:
I do have the same problem when printing. The alternative did not work for me.
If this fix solves the problem I would be happy if it would be included.

fredj:
The alternative provided by @ahocevar works for me:
 - https://openlayers.org/en/latest/examples/measure.html
 - open the browser developer console
 - paste the code
 - print

The tooltip is correctly positioned in the print preview

fmcd:
Apologies for late reply. Yes.. I had tried the alternative above and it did not work. the onbeforeprint event and the print image rendering processes run asynchronously, so the rendering process completes before the event code is run.

This issue also depends on the browser. It is more noticeable in Firefox on Linux, and some versions of IE on Windows. The patch I submitted was the only method I found that successfully corrected the issue in all three browsers.

PR #6930, opened 2017-06-14
Handle setActive(false) on an interaction without map

fredj:


fredj:
I've restarted the build and coverage is back to normal.

fredj:
@ahocevar what "Should just be a comment" ?

ahocevar:
@fredj I had revoked my disapproval, and GitHub asked me why: I just wanted to comment, and not disapprove.

So please merge.

fredj:
Thanks for the review

PR #6931, opened 2017-06-15
Make sure we use the default featurePrefix

bartvde:
This seems to be a change between 4.0.1 and 4.2.0, previously, when not specifying featurePrefix in a transaction, you would get ```feature``` but now you get undefined, this PR restores the old behaviour


bartvde:
@Jenselme any chance you could have a look at this one as a reviewer?

bartvde:
ah you're right @Jenselme I will wait for another review then, thanks for taking the time to look at it though

bartvde:
thanks @ahocevar for the review

PR #6933, opened 2017-06-17
Improve proj.get() logic

probins:
The majority case is that there is already a projection in `projections`, in which case there's no need to access proj4 whether it's enabled or not.


probins:
changed to be similar to [`proj.getTransformFromProjections`](https://github.com/openlayers/openlayers/blob/master/src/ol/proj.js#L363)

tschaub:
Looks like a nice improvement.

PR #6939, opened 2017-06-19
Abort loading when tile is disposed

### LINKED ISSUE #5532: Not stopping pending tile requests

ahocevar:
This is a follow-up on #5816.

Fixes #5532.

### DIFF
--- a/src/ol/imagetile.js
+++ b/src/ol/imagetile.js
@@ -31,7 +31,7 @@ ol.ImageTile = function(tileCoord, state, src, crossOrigin, tileLoadFunction) {
 
   /**
    * @private
-   * @type {Image}
+   * @type {Image|HTMLCanvasElement}
    */
   this.image_ = new Image();
   if (crossOrigin !== null) {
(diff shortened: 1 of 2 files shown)


fredj:
Note: `'image/gif'` is not supported on (at least) Chrome but the browser fallback to `'image/png'`

ahocevar:
Updated to use `image/png`.

thomasmoelhave:
Interesting stuff! Does it increase perceived performance when panning/zooming around on the map?

ahocevar:
@thomasmoelhave It does when you configure your tile source with a small or zero `cacheSize`.

thomasmoelhave:
Interesting, I guess it could have a big impact of the performance of the tile transition use-case as well (i.e. this stuff https://openlayers.org/en/latest/examples/wmts-dimensions.html )

PR #6945, opened 2017-06-22
Use Prettier

tschaub:
This makes use of Prettier via our ESLint config (see openlayers/eslint-config-openlayers#10 for the config changes).

If others are interested in making this change, it will create conflicts with all other PRs.  I'm happy to continue to rebase the commits in this branch as we get any other work in and can help others manage conflicts after this is in (running `npm run lint -- --fix` formats everything without manual intervention).


examples/accessible.js
```diff
@@ -13,9 +12,9 @@ var map = new ol.Map({
   ],
   target: 'map',
   controls: ol.control.defaults({
-    attributionOptions: /** @type {olx.control.AttributionOptions} */ ({
+    attributionOptions /** @type {olx.control.AttributionOptions} */: {
```
tschaub: This will break things for the Compiler.  I'll open a Prettier issue about this.

tschaub: Already reported as prettier/prettier#1445

tschaub:
I'll revisit this if/when prettier/prettier#1445 is addressed (if anybody wants to contribute a fix, I'm sure they'd welcome it).

PR #6953, opened 2017-06-25
Run tests in real browsers with Karma

tschaub:
Goals of this branch:

 * get all of our tests running together (no separate rendering and non-rendering tests)
 * get our tests running in multiple browsers

There is no need to have a separate process for running rendering tests.  In cases where we want to test functionality that only exists in limited browsers, we can use the existing `where()` function in our test extensions.

I've set up an account for the project on Sauce Labs, and there we can test in all browser/os combos we want to support.  I've included an [initial configuration](https://github.com/tschaub/openlayers/blob/33c3aaf69dcd1d1945f56e659d44427e1535770c/test/karma.config.js#L49-L75) that will run tests on Chrome, Firefox, Safari, IE 11, and Edge when the tests are run on Travis.  During development, I suggest we run tests in Chrome - although this can be configured.

There is more work to do, but if anybody wants to try it out, to run tests once, do this:
```
npm run karma -- --single-run
```

And to keep the tests running continuously:
```
npm run karma
```

For the continuous run case, more work is needed to update things as the dependency tree changes.


bjornharrtell:
Out of curiosity will this aim to get the tests, at least when running locally for development, in a headless Chrome if supported? (has been as of Chrome 59 AFAIK)

marcjansen:
I am happy you have started this, Tim. Looking good so far. Can you outline the remaining steps? I might lend a hand for certain tasks.

tschaub:
> Out of curiosity will this aim to get the tests, at least when running locally for development, in a headless Chrome if supported? (has been as of Chrome 59 AFAIK)

@bjornharrtell - it is straightforward to use Karma with Chrome in headless mode, but I'm not sure there is a good reason to do so.  Here is the required `karma.config.js` change:
```diff
diff --git a/test/karma.config.js b/test/karma.config.js
index dbd1943..0f2798b 100644
--- a/test/karma.config.js
+++ b/test/karma.config.js
@@ -93,7 +93,17 @@ module.exports = function(karma) {
     });
   } else {
     karma.set({
-      browsers: ['Chrome']
+      browsers: ['ChromeHeadless'],
+      customLaunchers: {
+        ChromeHeadless: {
+          base: 'Chrome',
+          flags: [
+            '--no-sandbox',
+            '--headless',
+            '--remote-debugging-port=9222'
+          ]
+        }
+      }
     });
   }
 };
```

The nice thing about running the tests in Chrome (with a head) is that you can quickly open the debug page and start debugging failing tests.  With headless Chrome, you can do the same, but you need to first open another instance of Chrome (with a head) and attach it to the test server.  It may be that the headless version has a faster startup time (I haven't tried), but I think the common use case during development is to be able to add and quickly debug failing tests.

tschaub:
> Can you outline the remaining steps?

@marcjansen here is what remains:

 - [ ] make things work when updating dependencies (and avoid redundant file watchers)
 - [ ] pull the rendering tests into the same directly structure as the rest of the tests, adding `where('WebGL')` calls for WebGL-specific tests
 - [ ] get rid of the old test runners

I'm starting at the top right now.

bjornharrtell:
I agree @tschaub, not sure why I focused on local development. But I think the remaining interesting case is replacing PhantomJS in CI.

tschaub:
> But I think the remaining interesting case is replacing PhantomJS in CI.

Right.  Two good alternatives on Travis: run all tests in Firefox (we could do this right now) or run all tests in multiple browsers using Sauce connect (what this branch will enable).

tschaub:
I'm uncertain about how to proceed with the WebGL renderer.  The renderer doesn't work for me (in Chrome or Firefox), so the rendering tests fail as well.  I'd like to make a condition (e.g. with `where('WebGL')`) that only runs the tests where we expect the renderer to work - but I don't know how to determine that.

Here is an example of the WebGL renderer in Chrome 59 for me:

image

So it should be no surprise that rendering tests fail.

tschaub:
To move forward with this, we need a decision on what to do with the WebGL rendering tests.  The WebGL renderer doesn't work on my os/browser combos (and the tests do not pass).  So for now, `where('WebGL')` prefixed tests are *always* skipped.  If someone can write an expression that evaluates to `true` where the tests pass, we can conditionally run them.

GaborFarkas:
May I have the layer producing those artifacts, @tschaub? I'm currently trying to come up with a straightforward fix, as with very detailed polygons I also experience some problems on smaller scales. It would be nice to have more problematic layers to play with.

tschaub:
> May I have the layer producing those artifacts, @tschaub?

@GaborFarkas - the screenshot in https://github.com/openlayers/openlayers/pull/6953#issuecomment-312724697 is from our `examples/vector-layer.js` using the WebGL renderer.

tschaub:
There is a bit of work left to get tests passing in Safari and Edge.  The rendering tests are failing in both.  In addition, both are [missing `new URL(url).searchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams#Browser_compatibility) so we'll need to rework the tests that use that.  I'll leave that for another PR.

tschaub:
I've left the coverage stuff out for now.  We can bring this back in after making a few other tooling changes.

.travis.yml
```diff
@@ -1,33 +1,18 @@
-sudo: false
-
+sudo: required
```
marcjansen: Is this required?

test/test-extensions.js
```diff
@@ -481,8 +489,17 @@
       throw new Error('where() called with unknown key: ' + key);
     }
     return {
-      describe: features[key] ? global.describe : global.xdescribe
+      describe: features[key] ? global.describe : global.xdescribe,
+      it: features[key] ? global.id : global.xit
     };
   };
 
+  // throw if anybody appends a div to the body and doesn't remove it
```
marcjansen: I love this

PR #6956, opened 2017-06-26
Image / Tile WMS proxy

gpcboekema:
imagewms and tilewms accept an optional proxy, and ol.uri.appendParams adds it to the url or inserts it via a placeholder.


ahocevar:
Have you considered configuring a `tileLoadFunction` on the source to apply your proxy setting? Something like
```js
  tileLoadFunction: function(image, src) {
    // either without auth
    image.getImage().src = 'http://example.com/proxy/' + src;
    // or with auth
    image.getImage().src = 'http://example.com/proxy/?Uri=###&auth=true'.replace('###', src);
  };
```

PR #6968, opened 2017-06-28
Shortest arc rotation animation improvements and upgrade notes

ahocevar:
After #6965, spin effect animations like "Spin to Rome" or "Rotate around Rome", as shown in the `animation.html` example, do not animate any more. This pull request updates the example and adds upgrade notes to restore the previous behavior.

This pull request also improves the situation after rotating beyond 360 degrees, by normalizing the rotation to an angle between -180 and 180 degrees at the end of an animation. The result of this can be seen in the `animation` example after rotating. Previously, the button of the `Rotation` control did not disappear. Now it does.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,26 @@
 
 ### Next release
 
+#### `ol.animate` now takes the shortest arc for rotation animation
+
+Usually rotation animations should animate along the shortest arc. There are rare occasions where a spinning animation effect is desired. So if you previously had something like
+```js
+map.getView().animate({
+  rotation: 2 * Math.PI,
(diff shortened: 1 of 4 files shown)


ahocevar:
Updated, now ready for review.

TimelyToga:
How long will it be until this patch lands in a release? Trying to decide if we need to serve this patch from our server or if it will be downloadable via the CDN soon enough.

ahocevar:
I just published v4.3.0-beta.2, both for the [ol](https://npmjs.com/package/ol) and [openlayers](https://npmjs.com/package/openlayers) packages. It should also be available on cdnjs.com within the next view hours.

Issue #6962, opened 2017-06-27
Error when using many tileLayers and adding/removing quickly


cmortazavi:
I have a map with many tile layers (about 15).  All tile layers are using an ol.source.XYZ with a tileCache of 0.  The user has a legend which allows him to add/remove these on the fly.  It seems like if he switches these on/off (add/removes)  at a quick pace he gets an error (see attached).

Seems like the usedTiles object in the expireCache function is null.  If I do this, then it fixes the issue.  Is this the correct way to fix this problem?

```
/**
 * @param {Object.<string, ol.TileRange>} usedTiles Used tiles.
 */
ol.TileCache.prototype.expireCache = function(usedTiles) {
    var tile, zKey;
    while (this.canExpireCache()) {
        tile = this.peekLast();
        zKey = tile.tileCoord[0].toString();
        if (usedTiles == null) {
            this.pop().dispose();
        }
        else if (zKey in usedTiles && usedTiles[zKey].contains(tile.tileCoord)) {
            break;
        } else {
            this.pop().dispose();
        }
    }
};
```

![error](https://user-images.githubusercontent.com/21088421/27604463-2dae7816-5b2e-11e7-9963-fb3779b933ed.jpg)

ahocevar:
The reason is probably a source whose tiles have never been used before. I tend to think that the issue should be fixed [https://github.com/openlayers/openlayers/blob/7d6929c710fd6205a69de49cf1f0538dbe406e8f/src/ol/renderer/layer.js#L153-L157](here) by not calling expireCache when there are no `usedTiles` for the respective `tileSourceKey`.

Are you willing to create a pull request?

cmortazavi:
How can I do this?  I don't have a full environment on my end.

ahocevar:
You can do a blind attempt by editing the respective source file directly in GitHub. Travis CI will run all checks and give green if your changes are formally correct and do not break tests.

cmortazavi:
Ok I think I did this correctly but please let me know if something seems off.

Thanks!
Cameron

fredj:
fixed with #6970

PR #6970, opened 2017-06-28
Fix Bug when adding/removing layer with no cache

cmortazavi:
https://github.com/openlayers/openlayers/issues/6962

### DIFF
--- a/src/ol/renderer/layer.js
+++ b/src/ol/renderer/layer.js
@@ -152,8 +152,10 @@ ol.renderer.Layer.prototype.scheduleExpireCache = function(frameState, tileSourc
      */
     var postRenderFunction = function(tileSource, map, frameState) {
       var tileSourceKey = ol.getUid(tileSource).toString();
-      tileSource.expireCache(frameState.viewState.projection,
-          frameState.usedTiles[tileSourceKey]);
+      if (tileSourceKey in frameState.usedTiles) {
+        tileSource.expireCache(frameState.viewState.projection,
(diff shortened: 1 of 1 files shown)


src/ol/renderer/layer.js
```diff
@@ -152,8 +152,10 @@ ol.renderer.Layer.prototype.scheduleExpireCache = function(frameState, tileSourc
      */
     var postRenderFunction = function(tileSource, map, frameState) {
       var tileSourceKey = ol.getUid(tileSource).toString();
-      tileSource.expireCache(frameState.viewState.projection,
-          frameState.usedTiles[tileSourceKey]);
+      if (frameState.usedTiles !== null && frameState.usedTiles[tileSourceKey] !== null) {
```
ahocevar: `frameState.usedTiles` should never be null. And I think that `frameState.usedTiles[tileSourceKey]` would rather be `undefined` and not `null`. Does it still work if you change the `if` condition to `if (tileSourceKey in frameState.usedTiles)`?

cmortazavi:
Changes made

ahocevar:
Thanks @cmortazavi. We'll need a [CLA](https://github.com/openlayers/cla) from you so we can merge this.

cmortazavi:
What is a CLA and how can I give you a CLA?

cmortazavi:
I think I figured out how to do this.  Does everything look correct?

ahocevar:
Thanks @cmortazavi, everything looks correct.

Issue #6974, opened 2017-06-30
DeviceOrientation not absolute in Chrome 50+


nagytech:
Since Chrome 50, the `alpha` attribute on `DeviceOrientationEvent` as raised from the `deviceorientation` DOM event has been supplied as a value relative to that of the initial `alpha` value.   This is indicated by the `absolute` property containing a value of `false`.  As a result, using  `ol.DeviceOrientation` to rotate the map view leads to inconsistent results.

This behaviour is similar to that of Safari on iOS, which OpenLayers has already accounted for through normalizing `alpha` into the `getHeading` method.  However, while the solution on iOS is to utilize the `webkitCompassHeading` propery of the event,  Chrome provides no such interface. 
 
Since Chrome 50, the `deviceorientationabsolute` event has been implemented (_see, DeviceOrientation Event Specification sec 4.2_) in order to assure the caller that the `alpha` property will contain an absolute value.

Note that while the W3C spec is still in draft, and MDN [has issued a caution](https://developer.mozilla.org/en-US/docs/Web/API/Window/ondeviceorientationabsolute) that the feature is non-standard, there is currently no other convenient way to obtain the device's absolute `alpha`.  Additionally, it is possible that different devices will report different values according to their orientation (ie. portrait vs landscape).  I have attempted to resolve possible inaccuracies across devices by applying an offset using `gamma` constrained to intervals of PI/2.  So far, this works for me on my S8 with Chrome 59 but will require further testing.  As a result, I have provided configuration values for utilizing the `deviceorientationabsolute` event, as well as that for calculating the offset via `gamma`.

I will issue a pull request shortly to address this issue. If anyone can assist with testing on other devices, it would be appreciated.

fredj:
We will deprecate `ol.DeviceOrientation`, see #7189

PR #6975, opened 2017-06-30
Utilize `deviceorientationabsolute` event.

nagytech:
See #6974.

I have further changes to address the gamma offset, but will commit them tomorrow, possibly under separate pull request to facilitate easier testing.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -39,7 +39,8 @@ olx.AttributionOptions.prototype.html;
 
 
 /**
- * @typedef {{tracking: (boolean|undefined)}}
+ * @typedef {{tracking: (boolean|undefined),
+ *    absolute: (boolean|undefined)}}
  */
 olx.DeviceOrientationOptions;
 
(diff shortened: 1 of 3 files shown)


fredj:
I'm not in favor of adding a non-standard feature in the API. Can you listen to this event in you application code ?

Side note: I don't think that `ol.DeviceOrientation` (as well as `ol.control.FullScreen`) has it's place in OpenLayers: it's only some glue around DOM APIs and is not related to maps

bartvde:
@fredj do you think we should consider removing those before we do OL5 ?

ahocevar:
> Side note: I don't think that ol.DeviceOrientation (as well as ol.control.FullScreen) has it's place in OpenLayers: it's only some glue around DOM APIs and is not related to maps

Agreed. I'd be in favor of removing it.

fredj:
@bartvde yes, I'm willing to remove it but keep the example and use the DOM API

fredj:
See #7189

PR #6976, opened 2017-07-02
Example - Earthquake Clusters - Change evt.type of interaction

ehanoj:
OriginalEvent 'mousemove' break default OL interaction move of map canvas. Using OL event 'pointermove' have same function and allow move of map canvas.


ahocevar:
Thanks for this improvement @ehanoj. Can you please submit a [CLA](https://github.com/openlayers/cla) so we can merge this?

ehanoj:
> Thanks for this improvement @ehanoj. Can you please submit a CLA so we
can merge this?
*** yes, I send "Agree" with CLA now.

d.

PR #6977, opened 2017-07-03
Fixed modify feature test

simonseyock:
The example was not working.
See: http://openlayers.org/en/master/examples/modify-test.html
and: https://codepen.io/anon/pen/MoGWNj

### DIFF
--- a/examples/modify-test.js
+++ b/examples/modify-test.js
@@ -225,7 +225,7 @@ var modify = new ol.interaction.Modify({
   style: overlayStyle,
   insertVertexCondition: function() {
     // prevent new vertices to be added to the polygons
-    return !this.features_.getArray().every(function(feature) {
+    return !select.getFeatures().getArray().every(function(feature) {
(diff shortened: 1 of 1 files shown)


fredj:
What was not working is the modification of the inner triangle right ?

simonseyock:
Yes, this was caused by the use of the undefined property features_ of the interaction inside of insertVertexCondition.
After changing this to select.getFeatures() the example is working again ...

But now I am asking myself: why was insertVertexCondition called in the first place?

simonseyock:
The call is caused by the if-clause in line https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/modify.js#L610

If i would rearrange the conditions so insertVertexCondition comes last it would only be called if the other conditions for inserting a vertex are met. 
This is probably not very important. Should I do that?

fredj:
@simonseyock thanks for this fix, and sorry for the long delay

PR #6981, opened 2017-07-04
Render transparent vector layers to an intermediate canvas

gberaudo:
Until now, the features of transparent vector layers were rendered using the layer opacity. This caused colors to mix together and the opacities to stack up to an higher value than the expected layer opacity.

With this commit, the features are rendered at 100% opacity to an intermediate canvas which ensures colors do not mix up even in the case of features using an array of styles. The intermediate canvas is then composed to the map canvas using the layer opacity.

Transparent layers are automatically detected, non-transparent layers are not affected by the change.

Before (the orange mixes together with the white):
ol3_original_rendering

Now (no more mixing, exact layer opacity):
ol3_intermediate_canvas_rendering


gberaudo:
I need this fix in some project. Could you please review?

gberaudo:
@fredj, I added a rendering test.

icemagno:
Will Vector layers have `setOpacity()` working as Tile layers ?

ahocevar:
Like tile layers, yes.

Issue #6987, opened 2017-07-05
Memory leak with WMS time source with reprojection


ch08532:
It appears there is a memory leak in Chrome when using a WMS source with a different projection ('EPSG:4326' reprojected to 'EPSG:3857').   

I've created a code pen using the WMS time example and just swapped it with our WMS feed.  

https://codepen.io/ch08532/pen/NgBWdM?editors=0010#0

Play the example and watch the memory jump and eventually crash Chrome.  I've tested in Firefox and it leaks as well.

ch08532:
Ran a heap profiler and compared a few snapshots and did notice a lot of detached DOM tree elements for Image and HTMLCanvasElement constructors.

Each Image constructor references image_ in ol.ImageTile and sourceTiles_ in ol.reproj.Tile



![snapshot](https://user-images.githubusercontent.com/16633974/27893731-9c9070a6-61d5-11e7-94fe-1affd769f691.png)

ahocevar:
I'd be interested if the situation improves when you configure a smaller `cacheSize` (e.g. `1024`) for the WMS source.

ch08532:
cacheSize did not make a difference.  Lowest cacheSize I could go was 64.  

This issue looks similar to issue [#6940](https://github.com/openlayers/openlayers/issues/6940) which you closed.  

Our application is served in two different projections and we reproject WMS/WMTS layers.  It appears that reprojected tiles leak memory when panning/zooming (which causes new tiles to be fetched).  Memory is in the GB size after 10 minutes of normal use.

ahocevar:
@petrsloup - do you have any idea?

ch08532:
@ahocevar, I was playing with the code in src/ol/source/tileimage.js where it instantiates a new reproj.Tile and determined that if I commented out the line
      
271	        newTile.interimTile = tile;

The memory leak stopped and did not see any detached objects on my heap snapshot. I did loose the "smoothness" interimTile provides (it basically acted like I wipped the layer clean and displayed the new set of tiles).

petrsloup:
@ch08532 Yes, the issue is probably that the new tile always starts with a reference to the interimTile and this reference is never released.

I think we should be able to safely clear the `interimTile` reference after the new tile is ready and reprojected, right?

ahocevar:
Sounds like the right fix @petrsloup. Are you or @ch08532 able to provide a pull request?

ch08532:
For fun, I added the following code after line 271:
[...]

PR #6996, opened 2017-07-07
6987: Memory leak with WMS time source with reprojection

ch08532:
As discussed, the collection of interimTile for the reproj.Tile was never being cleaned up and thus the references would still hang around. The fix is to perform:

tileImage.js:

newTile.refreshInterimChain(); 

after the line:
newTile.interimTile = tile 

The refreshInterimChain is only done inside the creation of a reproj.Tile (called within getTileInternal) for an ol.ImageTile. It wasn't being done on the created reproj.Tile.

### DIFF
--- a/src/ol/source/tileimage.js
+++ b/src/ol/source/tileimage.js
@@ -269,6 +269,7 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
 
       if (tile) {
         newTile.interimTile = tile;
+        newTile.refreshInterimChain();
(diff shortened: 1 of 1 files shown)


ch08532:
Issue #6987

ahocevar:
Thanks @ch08532 for the investigation and fix! Can you please submit a [CLA](https://github.com/openlayers/cla) so we can merge this?

ch08532:
@ahocevar I've just submitted the CLA.

PR #6999, opened 2017-07-09
Make VectorTile source work with multiple layers

### LINKED ISSUE #6994: re-using vectortilesource does not currently work
Only one layer will get rendered, see this test project https://github.com/theduckylittle/ol-multilayer-vectortile-bug @theduckylittle

There are 2 code paths that currently block the rendering:

https://github.com/openlayers/openlayers/blob/4a0f97ac6ad4e11df16604913d7e770b91a406b9/src/ol/renderer/canvas/vectortilelayer.js#L115
https://github.com/openlayers/openlayers/blob/4a0f97ac6ad4e11df16604913d7e770b91a406b9/src/ol/renderer/canvas/vectortilelayer.js#L437

ahocevar:
Fixes #6994.


theduckylittle:
Thanks a ton @ahocevar !!!

PR #7004, opened 2017-07-10
Use https for bing and stamen attributions

### LINKED ISSUE #7002: Attribution for BingMaps still served on HTTP, making our HTTPS site unsecure
Our website uses OL3. We recently enabled HTTPS on our website, but maps on Bing still serve the attribution on HTTP, making our map pages unsecure. I found that the links to OpenLayers and the Bing logo is served on HTTP (see screenshot). Can you please fix this immediately so we avoid an audit issue?

Thank you.
attribution

fredj:
fixes #7002

### DIFF
--- a/src/ol/source/bingmaps.js
+++ b/src/ol/source/bingmaps.js
@@ -85,7 +85,7 @@ ol.inherits(ol.source.BingMaps, ol.source.TileImage);
  */
 ol.source.BingMaps.TOS_ATTRIBUTION = new ol.Attribution({
   html: '<a class="ol-attribution-bing-tos" ' +
-      'href="http://www.microsoft.com/maps/product/terms.html">' +
+      'href="https://www.microsoft.com/maps/product/terms.html">' +
       'Terms of Use</a>'
 });
 
--- a/src/ol/source/stamen.js
+++ b/src/ol/source/stamen.js
@@ -48,8 +48,8 @@ ol.inherits(ol.source.Stamen, ol.source.XYZ);
  */
 ol.source.Stamen.ATTRIBUTIONS = [
   new ol.Attribution({
-    html: 'Map tiles by <a href="http://stamen.com/">Stamen Design</a>, ' +
-        'under <a href="http://creativecommons.org/licenses/by/3.0/">CC BY' +
+    html: 'Map tiles by <a href="https://stamen.com/">Stamen Design</a>, ' +
+        'under <a href="https://creativecommons.org/licenses/by/3.0/">CC BY' +
         ' 3.0</a>.'
   }),
   ol.source.OSM.ATTRIBUTION


joelaldor:
Hi there. We're still encountering the same issue. Still serving http://fonts.googleapis.com and http://dev.virtualearth.net. However http://fonts.googleapis.com is no longer accessible.

Please see screenshot. 
attribution

fredj:
http://fonts.googleapis.com comes from https://www.philippineheritagemap.org/static/css/app.css

fredj:
And the resource from http://dev.virtualearth.net is the bing logo but it should be converted to https:
https://github.com/openlayers/openlayers/blob/master/src/ol/source/bingmaps.js#L128-L131

joelaldor:
Oh thanks for correcting. When will this take effect? Right now the image is still being served over HTTP.

fredj:
What version of OpenLayers are you using?

fredj:
Looks like you're using v3.1.1 : https://www.philippineheritagemap.org/static/plugins/openlayers/ol.js

That's very old, and the http -> https conversion was not included in the version; you should update

joelaldor:
Hi there,

Much as we would want to update, it'll take some time before we can upgrade to the next version of our software that uses the latest version of OpenLayers, for various reasons. Is it possible the Bing logo image can be hosted on an HTTPS site so that browsers won't see our map pages are insecure? I believe that's the only issue remaining here. Appreciate all your help so far. Thank you.

fredj:
Maybe by providing your own `getLogo` function, something like:
```js
var bing = new ol.source.BingMaps({
    key: '<api_key>',
    imagerySet: 'Road'
});

bing.getLogo = function() {
    return "https://dev.virtualearth.net/Branding/logo_powered_by.png"
};

layers.push(new ol.layer.Tile({
  source: bing
}));
```

joelaldor:
Where should I put this code?

fredj:
Where `ol.source.BingMaps` is created

fredj:
Probably in https://www.philippineheritagemap.org/static/js/map/base-layers.js

mradamcox:
Hi @fredj, I was trying this out like so

    _.each(bingLayers, function(layer) {
 
        var bing = new ol.source.BingMaps({
            key: arches.bingKey,
            imagerySet: layer.id
        });

        bing.getLogo = function() {
            return "https://dev.virtualearth.net/Branding/logo_powered_by.png"
        };
        
        layer.layer = new ol.layer.Tile({
            visible: false,
            preload: Infinity,
            source: bing
        });
    });

to no avail. getLogo() didn't seem to be called to find the logo. Any thoughts? Is there a way to make sure ol uses that function to acquire the logo?

joelaldor:
Thanks for helping out @mradamcox !

joelaldor:
@fredj any updates on this one?

PR #7013, opened 2017-07-11
Remove ol.sphere.WGS84 and ol.sphere.NORMAL

tschaub:
These two `ol.Sphere` instances are only used by `ol.Geolocation` and `ol.proj.getPointResolution()`.  Instead of dedicated modules providing them, they are created where needed.

The next change that would be nice would be to export handy calculations for working with `ol.geom.Geometry` instances on a sphere.

For example, `ol.sphere.getLength(line)` would return the great circle length of the line (with sensible defaults for geometry projection and sphere radius).  It would be nice to have a breaking change (removing the `ol.Sphere` constructor) for this.


probins:
are the package.json changes supposed to be in there? - looks unrelated

tschaub:
The `package.json` changes come from `npm` auto-indenting.  I'll include them in this change since it will come up for anybody `npm install`ing with newer npm.

PR #7022, opened 2017-07-14
Allow styles to configure a custom renderer

ahocevar:
This pull request adds a `renderer` option to `ol.style.Style`. The `renderer` is simply a function that receives (among other arguments) the simplified pixel coordinates for the resolution that the replay was created for, and is called for every geometry with every render frame.

Two new examples show how custom renderers can be used to render text along paths, and to declutter labels using 3rd party libraries:

vetor-label-decluttering.html
vetor-label-decluttering.html

street-labels.html
street-labels.html


src/ol/render/canvas/replay.js
```diff
@@ -303,6 +385,21 @@ ol.render.canvas.Replay.prototype.replay_ = function(
...
         break;
+      case ol.render.canvas.Instruction.CUSTOM:
+        d = /** @type {number} */ (instruction[1]);
+        dd = instruction[2];
+        var geometry = /** @type {ol.geom.SimpleGeometry} */ (instruction[3]);
+        var renderer = instruction[4];
+        var coords;
+        if (instruction.length == 6) {
+          var fn = instruction[5];
+          coords = fn(pixelCoordinates, d, dd, 2, geometry.getRenderCoordinates());
```
bartvde: answering my own previous question, here the render coordinates will get set by ```fn```

src/ol/style/style.js
```diff
@@ -93,6 +99,28 @@ ol.style.Style.prototype.clone = function() {
...
+ * @api
+ */
+ol.style.Style.prototype.getRenderer = function() {
+  return this.renderer_;
+};
+
+
+/**
+ * Sets a custom renderer function for this style. When set, `fill`, `stroke`
+ * and `image` optins of the style will be ignored.
```
bartvde: options

tasks/build-examples.js
```diff
@@ -10,7 +10,7 @@ var markupRegEx = /([^\/^\.]*)\.html$/;
 var cleanupJSRegEx = /.*(\/\/ NOCOMPILE|goog\.require\(.*\);)[\r\n]*/g;
 var requiresRegEx = /.*goog\.require\('(ol\.\S*)'\);/g;
 var isCssRegEx = /\.css$/;
-var isJsRegEx = /\.js$/;
+var isJsRegEx = /\.js(\?.*)?$/;
```
bartvde: can you explain why this change was needed?

marcjansen:
Very nice examples!

chrismayer:
Wow, the examples are very impressive. This functionality offers great possibilities.

examples/street-labels.js
```diff
@@ -0,0 +1,177 @@
...
+
+function dist2D(p1, p2) {
+  var dx = p2[0] - p1[0];
+  var dy = p2[1] - p1[1];
+  return Math.sqrt(dx * dx + dy * dy);
+}
+
+// Modified from https://github.com/Viglino/ol3-ext/blob/7d17eef5720970fd36798ebc889ea44d9f04b059/style/settextpathstyle.js
+function textPath(ctx, text, path, fid, pixelRatio) {
+  var di, dpos = 0;
```
ahocevar: I am moving most of the code from this example into npm packages, so it will be much easier to read.

tasks/build-examples.js
ahocevar: See `examples/street-labels.html`. It requires a polyfill for `Set`, so it includes `https://cdn.polyfill.io/v2/polyfill.min.js?features=Set` as resource - that would not work with the previous regex.

examples/vector-label-decluttering.js
```diff
@@ -0,0 +1,130 @@
...
+function setStyle(context) {
+  context.font = '12px Calibri,sans-serif';
+  context.fillStyle = '#000';
+  context.strokeStyle = '#fff';
+  context.lineWidth = 3;
+  context.textBaseline = 'hanging';
+  context.textAlign = 'start';
+}
+
+// A separate canvas context for measuring label width and height.
```
ahocevar: This example caches labels to separate canvas elements, so there is no need to save/restore the context. This makes things a bit more efficient.

ahocevar:
@bartvde Thanks for the review! I have addressed all comments and answered your questions.

examples/vector-label-decluttering.js
ahocevar: Regarding the changelog: I think this new feature should be described in the release notes upon release. In the changelog, we usually only mention breaking changes.

bartvde: You're right

klokan:
@ahocevar this is great! Wow.

PR #7028, opened 2017-07-17
Fix Graticule use of incorrect min/maxLon values

### LINKED ISSUE #7020: Poor Graticule Performance in Chrome
I noticed when using the Graticule feature in Chrome, some intermediate zoom levels render very slowly relative to other zoom levels. Sometimes a zoom action will lock up for a number of seconds.

Steps to reproduce:
- Open http://openlayers.org/en/latest/examples/graticule.html in Chrome
- Zoom in about 5 zoom levels
- Notice that the animation gets slower as you continue to zoom in.
- If you continue to zoom in, it begins to render quicker again.

I was able to reproduce this in Chrome Version 59.0.3071.115 (Official Build) (64-bit) on Windows 7 and Ubuntu 16.04

greggian:
Appears to have just been a typeo, using the minLon_, maxLon_ member variables which represent the world extent rather than the function parameters provided which are the viewport extent.

This was creating very wide lines of parallel which were sometimes very slow in Chrome.

Fixes #7020

### DIFF
--- a/src/ol/graticule.js
+++ b/src/ol/graticule.js
@@ -519,7 +519,7 @@ ol.Graticule.prototype.getMeridians = function() {
 ol.Graticule.prototype.getParallel_ = function(lat, minLon, maxLon,
     squaredTolerance, index) {
   var flatCoordinates = ol.geom.flat.geodesic.parallel(lat,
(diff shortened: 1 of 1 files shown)


greggian:
Thanks for the quick merge and thanks to all the committers for the great project.

PR #7033, opened 2017-07-19
fixes dependencies.

BasieP:
(moved all dependencies to dev-dependencies since the npm package is a bundle which includes all the dependencies)


tschaub:
Please use the ol package instead.  It is more appropriate for application development.

Issue #7040, opened 2017-07-20
lineDashOffset has no effect


gkresic:
Tested on `ol.style.Stroke`. LineDash works correctly, but offset is ignored.

See live example: https://codepen.io/gkresic/pen/gRVXgL

To test browser support (works on mine Chrome 60): https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineDashOffset

OpenLayers 4.2.0

ahocevar:
`lineDashOffset` is not supported by the immediate renderer, which is used for rendering regular shapes. That functionality would have to be added to `ol.render.canvas.Immediate.prototype.setContextStrokeState_` in `src/ol/render/canvas/immediate.js`. We'd accept a pull request for that.

fredj:
fixed with #7042

PR #7042, opened 2017-07-24
Line dash offset

gkresic:
Solves #7040.

Modified examples show that offset works, but I didn't submit those examples because they are not worthy. Tested "running ants" feature in my personal GWT project and it works as expected.

**Note**: My JavaScript-foo is pretty weak these days (I'm a Java bot), so I'm not sure that I've tested everything right. I did make an extra effort to check that offset is set/cleared/propagated everywhere but it would be best if someone more experienced with the project tests this thoroughly.


ahocevar:
Thanks for the contribution! Please submit a [CLA](https://github.com/openlayers/cla) so we can merge this. Also, if possible, please add a rendering test to `test_rendering/spec/ol/style/regularshape.js` and `test_rendering/spec/ol/render.test.js`. See the [README](https://github.com/openlayers/openlayers/tree/master/test_rendering) for instructions on how to write rendering tests.

gkresic:
I did already sign Individual Contributor License Agreement. Can you confirm that you can see my signature?

Regarding tests, I couldn't find any that tests line dash, so I didn't add any for line dash offset either. Main reason is that I could see that line dash feature is conditionally enabled in OL by browser capabilities, and adding those tests to JavaScript is beyond my knowledge.

ahocevar:
CLA received, rendering tests added.

PR #7051, opened 2017-07-26
Changing the EPSG3857.PROJECTION array assignment and adding urn:ogc:…

wnordmann:
adding projection urn:ocg:crs:EPSG::3857 
while removing a duplicate EPSG:102113
Changed epsg3857.projection to match spsg4326.projection

### DIFF
--- a/src/ol/proj/epsg3857.js
+++ b/src/ol/proj/epsg3857.js
@@ -73,6 +73,7 @@ ol.proj.EPSG3857.PROJECTIONS = [
   new ol.proj.EPSG3857.Projection_('EPSG:3857'),
   new ol.proj.EPSG3857.Projection_('EPSG:102100'),
   new ol.proj.EPSG3857.Projection_('EPSG:102113'),
+  new ol.proj.EPSG3857.Projection_('EPSG:900913'),
   new ol.proj.EPSG3857.Projection_('urn:ogc:def:crs:EPSG:6.18:3:3857'),
   new ol.proj.EPSG3857.Projection_('urn:ogc:def:crs:EPSG:3857'),
   new ol.proj.EPSG3857.Projection_('urn:ogc:def:crs:EPSG::3857'),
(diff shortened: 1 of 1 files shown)


src/ol/proj/epsg3857.js
```diff
@@ -64,30 +64,20 @@ ol.proj.EPSG3857.WORLD_EXTENT = [-180, -85, 180, 85];
...
  * @const
  * @type {Array.<ol.proj.Projection>}
  */
-ol.proj.EPSG3857.PROJECTIONS = ol.proj.EPSG3857.CODES.map(function(code) {
-  return new ol.proj.EPSG3857.Projection_(code);
-});
+ol.proj.EPSG3857.PROJECTIONS = [
+  new ol.proj.EPSG3857.Projection_('EPSG:3857'),
+  new ol.proj.EPSG3857.Projection_('EPSG:102100'),
+  new ol.proj.EPSG3857.Projection_('EPSG:102113'),
```
ahocevar: You removed the most important alternative code, 'EPSG:900913'. Please add it back.

wnordmann: Copy paste error, fixed

bartvde:
Which data contained ```urn:ogc:def:crs:EPSG:3857```

Isn't that an invalid URN?

http://docs.opengeospatial.org/is/12-063r5/12-063r5.html

wnordmann:
Agreed, but the original code was using the urn:ogc:def:crs:EPSG:3857 I
didn't want to break other code

bartvde:
Am I reading the diff incorrectly then? I only see ```urn:ogc:def:crs:EPSG:3857``` added

bartvde:
```urn:ogc:def:crs:EPSG::3857``` which was there is a valid URN

wnordmann:
You are right, urn:ogc:def:crs:EPSG::3857 is valid and
urn:ogc:def:crs:EPSG:3857 is invalid, somewhere yesterday I got something
mixed up.  Sorry for the confusion.

ahocevar:
@wnordmann: Can you please submit a follow-up pull request that removes the invalid URN?

Issue #7056, opened 2017-07-28
feature request: more interaction.Extent events


marcosox:
Currently `ol.interaction.Extent` fires only "`extentchanged`" events (by the way, [the docs](https://openlayers.org/en/latest/apidoc/ol.interaction.Extent.html#.event:Event) specify a wrong "Event.event" name).

It would be useful to have separate events also for draw start and end, (with different names obviously), to notify about drag start and drag ends and to have a finer control over this interaction.

ahocevar:
I have created pull request #7066 to fix the event name.

Regarding the feature request, can you describe a use case where simply knowing when the drawn extent changes would not be sufficient?

marcosox:
I would use for example the "end" event to call a function that performs a serie of actions based on the new extent. Calling that function at every pixel change would be too intensive. I could also decide to deactivate the interaction when certain conditions are met, and it would be confusing for the user to have the interaction disabled in the middle of a dragging action. Another example could be changing the mouse cursor during the interaction.

rasa2k:
I also find this very useful and currently need the functionality.

Chris1234567899:
Any news on this for 'start' and 'end' events?

PR #7066, opened 2017-07-30
Fix ol.interaction.Extent event type and documentation

ahocevar:
Fixes the wrong event name mentioned in #7056.


bartvde:
the Travis fail seems unrelated? Or is it because of Travis's dist change?

ahocevar:
Thankjs for the review. I restarted the Travis build, and now it completes without errors. Merging.

PR #7070, opened 2017-08-02
Get tilePixelRatio from MVT tiles

ahocevar:
When working with MVT tiles, we can determine the `tilePixelRatio` from metadata in the tile. This pull request changes things so users no longer need to configure the source with a `tilePixelRatio` when `ol.format.MVT` is used to parse tiles.

### DIFF
--- a/src/ol/source/tile.js
+++ b/src/ol/source/tile.js
@@ -244,12 +244,11 @@ ol.source.Tile.prototype.getTileCacheForProjection = function(projection) {
 /**
  * Get the tile pixel ratio for this source. Subclasses may override this
  * method, which is meant to return a supported pixel ratio that matches the
- * provided `opt_pixelRatio` as close as possible. When no `opt_pixelRatio` is
- * provided, it is meant to return `this.tilePixelRatio_`.
- * @param {number=} opt_pixelRatio Pixel ratio.
+ * provided `pixelRatio` as close as possible.
+ * @param {number} pixelRatio Pixel ratio.
  * @return {number} Tile pixel ratio.
  */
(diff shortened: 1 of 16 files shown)


ahocevar:
Tests added.

src/ol/source/tile.js
```diff
@@ -244,12 +244,11 @@ ol.source.Tile.prototype.getTileCacheForProjection = function(projection) {
...
  * method, which is meant to return a supported pixel ratio that matches the
- * provided `opt_pixelRatio` as close as possible. When no `opt_pixelRatio` is
- * provided, it is meant to return `this.tilePixelRatio_`.
- * @param {number=} opt_pixelRatio Pixel ratio.
+ * provided `pixelRatio` as close as possible.
+ * @param {number} pixelRatio Pixel ratio.
  * @return {number} Tile pixel ratio.
  */
-ol.source.Tile.prototype.getTilePixelRatio = function(opt_pixelRatio) {
+ol.source.Tile.prototype.getTilePixelRatio = function(pixelRatio) {
```
tschaub: Is this change required?

ahocevar: It is not strictly required, but makes things consistent again. The vector tile renderer introduced a 2nd mode of `getTilePixelRatio()`: when called without a pixel ratio, it was meant to return the source's `tilePixelRatio`, which was the `8` used in the examples.

Originally, when called with a pixel ratio, the function was meant to return the `tilePixelRatio` that best matches the provided `pixelRatio`. This is usually the provided `pixelRatio` for vector tiles and WMS servers that have a dpi param, or `1` for image tile sources that have no HiDPI support.

Now we're back to that 2nd, original meaning.

PR #7073, opened 2017-08-03
Make ol.layer.Group change handling consistent

gberaudo:
Changes to the layer were only propagated in some cases like addition/removal of a layer to the group.

To make it consistent both with the rest of the group.js file and with the layer.js file, the changes are now always notified.


gberaudo:
Thanks for the review @fredj.

PR #7094, opened 2017-08-09
Add missing zIndex options

### LINKED ISSUE #7091: Layer constructor options missing zIndex
- [x] If reporting a bug, I have tried with the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/)

Layer classes that inherit from `ol.layer.Layer` don't have `zIndex` documented even though it seems like they all inherit from `olx.layer.LayerOptions`.

icholy:
Fixes #7091

- [x] This pull request addresses an issue that has been marked with the 'Pull request accepted' label.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3851,7 +3851,8 @@ olx.layer.GroupOptions.prototype.layers;
  *     maxResolution: (number|undefined),
  *     opacity: (number|undefined),
  *     source: (ol.source.Vector|undefined),
- *     visible: (boolean|undefined)}}
+ *     visible: (boolean|undefined),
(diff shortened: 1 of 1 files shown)


fredj:
The `zIndex` option must also be set the the object prototype, for example:
```
/**
 * The z-index for layer rendering.  At rendering time, the layers will be
 * ordered, first by Z-index and then by position. The default Z-index is 0.
 * @type {number|undefined}
 * @api
 */
olx.layer.HeatmapOptions.prototype.zIndex;
```

icholy:
@fredj added

tschaub:
Thanks for the contribution @icholy.

Issue #4577, opened 2015-12-17
Feature text with background color


omnik:
I don't found how to create background box for ol.Feature text style.
Is it possible? It would be great if ol3 has such feature.

ahocevar:
You can set a  `stroke` on your `ol.style.Text` to create a halo. A rectangular box is not possible.

sanderd17:
The easiest to add a rectangular background is using an image. Creating a data-uri for an svg image is quite easy. Just do `data:image/svg+xml;utf8," + encodeURIComponent("<svg>...</svg>")` with a rectangle in the svg definition.

szel:
I found this issue in 2017 and would like to elaborate. Why it is not possible? Seems like quite basic feature..

ahocevar:
With the new text renderer, an image is created for the text anyway. That image could be given a stroke and fill. For `ol.style.Text`, there could be new `backgroundFill`, `backgroundStroke` and ´padding´ properties. Anyone interested in creating a pull request?

marcjansen:
sounds like a nice improvement, @ahocevar.

pavle-lekic-htec:
I have a problem, I have set the backgroundFill but for some reason it doesn't work. I'm using openlayers 4.4.2
Here is the code:

```
return new ol.style.Style({
        image: new ol.style.Icon({
            anchor: [0.5, 0.5],
            opacity: 1,
            scale: 1,
            src: 'public-images/map/stena_port_unselected.png'
        }),
        text: new ol.style.Text({
            offsetY: 20,
            text: portName,
            placement: 'point',
            textAlign: 'center',
            fill: new ol.style.Fill({color: 'rgb(255, 169, 163)'}),
            backgroundFill: new ol.style.Fill({color: 'black'}),
            padding: [3, 5, 3, 5]
        })
    });
```

Can you please help me? Am I doing something wrong?

bartvde:
this code is not in 4.4.2, please try the latest release 4.6.2

aaron1uk:
is there anyway yet implemented to have a radius setting to curve the background box?

ahocevar:
No, there is not.

PR #7098, opened 2017-08-09
Use fractional coordinates for CSS positioning

ahocevar:
Especially on HiDPI devices, `ol.Overlay` overlays wiggle on top of the map when panning. This can be avoided by not rounding the css coordinates.

### DIFF
--- a/src/ol/overlay.js
+++ b/src/ol/overlay.js
@@ -475,7 +475,7 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
     if (this.rendered_.left_ !== '') {
       this.rendered_.left_ = style.left = '';
     }
-    var right = Math.round(mapSize[0] - pixel[0] - offsetX) + 'px';
+    var right = (mapSize[0] - pixel[0] - offsetX) + 'px';
     if (this.rendered_.right_ != right) {
       this.rendered_.right_ = style.right = right;
     }
@@ -488,7 +488,7 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
         positioning == ol.OverlayPositioning.TOP_CENTER) {
       offsetX -= this.element_.offsetWidth / 2;
     }
-    var left = Math.round(pixel[0] + offsetX) + 'px';
+    var left = (pixel[0] + offsetX) + 'px';
     if (this.rendered_.left_ != left) {
       this.rendered_.left_ = style.left = left;
     }
@@ -499,7 +499,7 @@ ol.Overlay.prototype.updateRenderedPosition = function(pixel, mapSize) {
     if (this.rendered_.top_ !== '') {
       this.rendered_.top_ = style.top = '';
     }
-    var bottom = Math.round(mapSize[1] - pixel[1] - offsetY) + 'px';
+    var bottom = (mapSize[1] - pixel[1] - offsetY) + 'px';
     if (this.rendered_.bottom_ != bottom) {
       this.rendered_.bottom_ = style.bottom = bottom;
     }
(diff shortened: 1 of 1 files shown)


wayofthefuture:
Thanks Guys!

hccampos:
Nice stuff! Even better would be to use `translate(x, y, 0)` to improve the performance. Tricky to keep compatibility with older browsers though.

wayofthefuture:
I've been testing this for a while now and it did not fix the overlay wiggle. It can be most seen on an Android mobile device running Chrome browser. Does anyone have any other ideas of how to solve this issue? We get considerable wiggle on our label overlays.

[Video Located Here](https://vimeo.com/233140974)

 Thanks

ahocevar:
@wayofthefuture It looks like the wiggle in your video is caused by something else. It's more than a pixel. Maybe you're using some custom placement options? Or maybe you apply css transform in additon to the ol.Overlay positioning?

wayofthefuture:
Yes we are definitely using css transforms for sure. The element we are passing into the ol.Overlay constructor options has a transform (rotate) and transform origin applied. So maybe it has something to do with that.

ogo-adp:
Hi,

We have a performance problem trying to migrate from version 4.2.0 to version 4.3.0, and I think this commit is the culprit.

Our app can display a lot of overlays on screen, and I see a few problems with this commit :
- Safari on IOS gets a performance hit just because of this change
- If overlays are used to display text on a low dpi screen (like a standard PC) then it will not be able to get "pixel perfect" alignment of text, and it will become fuzzy. However it seems like chrome does some rounding internally anyway making this change pointless. However I didn't measure if chrome gets a performance hit with this change, nor did a test with Firefox
- Not rounding css positions is not consistent with the behaviour or geometries in OL. Many OL styles have a propery called "snapToPixel" which defaults to "true".

So I suggest to revert back to the old behaviour to :
1) Restore performance to what it was before
2) Be consistent with what is done in styles

An I also suggest to add a "snapToPixel" (defaulting to "true") to ol.Overlay() constructor for people that would like to have non rounded css positions.

ahocevar:
Sounds reasonable @ogo-adp. Pull request welcome.

hccampos:
@ogo-adp @ahocevar You should be able to get rid of your performance problems by sub-classing `ol.Overlay`. Here is an example in Typescript:

```
import olOverlay from 'ol/overlay';

export class Overlay extends olOverlay {
  private element_: HTMLElement;

  updateRenderedPosition(pixel: [number, number], _: any) {
    const x = Math.round(pixel[0]);
    const y = Math.round(pixel[1]);
    this.element_.style.transform = `translate(${x}px, ${y}px)`;

    (this as any).setVisible(true);
  }
}
```

ogo-adp:
@hccampos I'm not sure it is feasbile to subclass anything when using the npm package.
@ahocevar Currently i'm not able to prepare a pull request (i'd love to ...). Would you accept to make the modifications ?

ahocevar:
@ogo-adp Not a priority for me. I don't recommend to use `ol.Overlay` for anything other than tooltips or info popups, with only on visible at a time, and for that use case there is no performance issue. For anything else, you should be using vector layers and text styles. #4577 might be of interest in this context.

ogo-adp:
@ahocevar Well we have a true use case for overlays (stickers for moving objects, and will need HTML for that, and even #4577 would not be enough for us). Could it be possible to at least revert #7098 in next release because it make performance worse for no actual benefit ?

ahocevar:
Maybe someone else interested in this can submit a pull request?

ahocevar:
@ogo-adp you can also send me an e-mail, maybe we can figure out a way.

tschaub:
If this change didn't fix an issue that anybody was running into, and it is causing a new issue, it seems like reverting is the right thing to do.  If that sounds ok to you @ahocevar, I can cut another patch release with the change.

ahocevar:
Looks like @wayofthefuture's issue was not fixed with this change anyway. I'll revert.

ahocevar:
#7258 created. @tschaub, I will merge as soon as Travis is green. Then you can create a patch release. Thanks!

ogo-adp:
Many thanks @tschaub and @ahocevar :)

tschaub:
Published `ol@4.3.3` (website and full-build artifacts up soon).

PR #7108, opened 2017-08-11
IE11: Avoid text selection when using Shift + Click in Map

NisStrom:
When using Shift + Click on the map inside IE 11 for zooming to an area or free hand drawing a linesegement all Text around the map get selected by the browser.

Make sure these boxes are checked before submitting your pull request. Thank you!

- [ ] This pull request addresses an issue that has been marked with the 'Pull request accepted' label.
- [x ] It contains one or more small, incremental, logically separate commits, with no merge commits.
- [x ] I have used clear commit messages.

### DIFF
--- a/src/ol/mapbrowsereventhandler.js
+++ b/src/ol/mapbrowsereventhandler.js
@@ -202,7 +202,7 @@ ol.MapBrowserEventHandler.prototype.isMouseActionButton_ = function(pointerEvent
  * @private
  */
 ol.MapBrowserEventHandler.prototype.handlePointerDown_ = function(pointerEvent) {
-  var isIE11 = navigator.userAgent.match(/Trident\/7.0; rv 11.0/);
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks for the contribution @NisStrom.  I think preventing the default on all `pointerdown` events has potential for negative side effects (like maybe `click` events not firing).  It feels like we should be able to handle this with `user-select: none` in the right place.

Does this problem occur with [the hosted examples](http://openlayers.org/en/master/examples/)?  If not, can you post a simple example (on https://codepen.io/) that demonstrates the problem?

NisStrom:
The problem occurs with the hosted examples and as I already mentioned is only an issue with IE 11, it does happen with Edge, Firefox or Chrome. One of the example with Shift + Drag for zooming would be https://openlayers.org/en/latest/examples/min-zoom.html.

PR #7122, opened 2017-08-14
Immediately complete no-op animations

tschaub:
Currently, if you call `view.animate()` with properties that imply no view change, the view stays "animating" for the duration of the animation.  This means that no tiles load for the default case.  Instead, this no-op animation should immediately complete (and any others in the series should start).

### DIFF
--- a/src/ol/view.js
+++ b/src/ol/view.js
@@ -302,7 +302,14 @@ ol.View.prototype.animate = function(var_args) {
     }
 
     animation.callback = callback;
-    start += animation.duration;
+
+    // check if animation is a no-op
+    if (ol.View.isNoopAnimation(animation)) {
+      animation.complete = true;
+      // we still push it onto the series for callback handling
+    } else {
+      start += animation.duration;
+    }
     series.push(animation);
   }
   this.animations_.push(series);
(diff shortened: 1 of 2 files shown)


src/ol/view.js
```diff
@@ -1158,3 +1165,27 @@ ol.View.createRotationConstraint_ = function(options) {
...
+
+
+/**
+ * Determine if an animation involves no view change.
+ * @param {ol.ViewAnimation} animation The animation.
+ * @return {boolean} The animation involves no view change.
+ */
+ol.View.isNoopAnimation = function(animation) {
+  if (animation.sourceCenter) {
+    if (animation.sourceCenter[0] !== animation.targetCenter[0]) {
```
fredj: use `ol.coordinate.equals` ?

tschaub: Added in 7f0c8fe89961fe75d60ed4abd536ffb178c23e7c.

test/spec/ol/view.test.js
```diff
@@ -613,7 +642,7 @@ describe('ol.View', function() {
       expect(view.getHints()[ol.ViewHint.ANIMATING]).to.be(2);
 
       view.animate({
-        rotate: Math.PI,
+        rotation: Math.PI,
```
tschaub: This (and same below) was a typo in the test.  We were creating no-op animations (`rotate`) is not an animation property and asserting that the view was actually animating.  With the new no-op checking, these tests failed.  With the proper spelling of `rotation`, the tests pass again.

PR #7125, opened 2017-08-14
Do not try to render error tiles from VectorTile source

ahocevar:
Currently, vector tiles will render even when some of their source tiles failed to load. This allows tile loaders to re-request these tiles, and the will render fine the next time they need to be rendered. But the attempt to render them will fail, because error tiles will e.g. return null on '#getProjection'. This change makes it so these error tiles are just skipped.


ahocevar:
I'll spend a bit more time on this to make sure that source tiles that were loaded again after an error trigger a re-render.

ahocevar:
This is now ready for review.

PR #7131, opened 2017-08-15
Print ES6 import hint on each doc page

ahocevar:
This pull request improves the docs for `ol` package users:

es6

### DIFF
--- a/config/jsdoc/api/template/tmpl/container.tmpl
+++ b/config/jsdoc/api/template/tmpl/container.tmpl
@@ -10,40 +10,53 @@
 <?js } else { ?>
 
 <section>
-    
+
 <header>
     <h2><?js if (doc.ancestors && doc.ancestors.length) { ?>
         <span class="ancestors"><?js= doc.ancestors.join('') ?></span><?js } ?><?js= doc.name ?>
     <?js if (doc.variation) { ?>
         <sup class="variation"><?js= doc.variation ?></sup>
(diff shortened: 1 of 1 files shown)


config/jsdoc/api/template/tmpl/container.tmpl
```diff
@@ -10,40 +10,53 @@
...
+      var ancestors = doc.ancestors.map(a => a.replace(/>\./g, '>').replace(/\.</g, '<')).join('/');
+      var name = doc.name.toLowerCase();
+      var parts = [];
+      if (ancestors) {
+        parts.push(ancestors);
+      }
+      parts.push(name);
+      var importPath = parts.join('/');
+    ?>
+    <pre class="prettyprint source"><code>import <?js= doc.name ?> from '<?js= importPath ?>';</code></pre>
```
tschaub: I'm puzzled by the replacement of `>.` with `>` and `.<` with `<` in the ancestors (not clear to me when we have or want angle brackets in import paths).

ahocevar: The `ancestors` contain links and namespace dots. I could have taken `longname` instead and just replace `'.'` width `'/'`, but then we would lose the links.

tschaub: Makes sense.  Thanks for the explanation.

PR #7103, opened 2017-08-10
Set blank image data when images fail to load

mattlovett:
This PR is a fix for https://github.com/openlayers/openlayers/issues/7090

- [x] This pull request addresses an issue that has been marked with the 'Pull request accepted' label.
- [x] It contains one or more small, incremental, logically separate commits, with no merge commits.
- [x] I have used clear commit messages.

PR #7136, opened 2017-08-16
Use data URI instead of whole empty image

### LINKED ISSUE #7090: ol.ImageTile doesn't load() properly after errors
We use a tile source that often fails, so our code includes a listener that calls load() on the tile to retry after an error.

We recently noticed that when we load a large map, if there are errors, then the same map image gets displayed in several places on the map. I think this is due to one of the recent changes in ol.ImageTile.

Here's what happens when a tile load error happens:
- ImageTime.handleImageError_() sets the image_ to point to a blank tile.
- Our code gets a callback telling us about the error, and we call load()
- The default loading function does "imageTile.getImage().src = src;"

This has just set a real source onto the 'blank' tile. Each time we get a load error we do the same thing, and all of the failing tiles end up sharing the same 'blank' image. I think a simple fix would be so set up a new image each time we attempt the load... but there may be something even easier.

ahocevar:
Replaces #7103.

Fixes #7090.

### DIFF
--- a/src/ol/imagetile.js
+++ b/src/ol/imagetile.js
@@ -60,7 +60,7 @@ ol.inherits(ol.ImageTile, ol.Tile);
 ol.ImageTile.prototype.disposeInternal = function() {
   if (this.state == ol.TileState.LOADING) {
     this.unlistenImage_();
-    this.image_.src = ol.ImageTile.blankImage.toDataURL('image/png');
+    this.image_.src = ol.ImageTile.blankImageUrl;
   }
   if (this.interimTile) {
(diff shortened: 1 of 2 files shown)


ahocevar:
Accidentally committed more than I wanted. Will fix that.

src/ol/imagetile.js
```diff
@@ -152,9 +152,9 @@ ol.ImageTile.prototype.unlistenImage_ = function() {
  * A blank image.
  * @type {HTMLCanvasElement}
```
fredj: should be `@type {string}`

PR #7137, opened 2017-08-17
Developer documentation updates

tschaub:
This adds instructions on running tests in other browsers and instructs people to use Node 8.  The `package-lock.json` means we get the same version of dependencies everywhere.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -17,7 +17,7 @@ The minimum requirements are:
 
 * GNU Make
 * Git
-* [Node.js](http://nodejs.org/) (higher than 0.12.x)
+* [Node.js](http://nodejs.org/) (version 8 and above)
 * Python 2.6 or 2.7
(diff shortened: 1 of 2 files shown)


marcjansen:
I think that if we add package-lock.json, we also should enable [greenkeeper-lockfile](https://github.com/greenkeeperio/greenkeeper-lockfile) otherwise gk will only update the package.json

ahocevar:
Good call. Please create a pull request.

tschaub:
> I think that if we add package-lock.json, we also should enable greenkeeper-lockfile otherwise gk will only update the package.json

We're running into the same issue described in greenkeeperio/greenkeeper-lockfile#44 (see https://travis-ci.org/openlayers/openlayers/builds/265595019 for example).  I tried to troubleshoot but think it is not worth the effort at this point.

I'll remove the `package-lock.json` until more people are using an updated `npm` and things are more stable.

PR #7138, opened 2017-08-17
Improvements to the new test setup

### LINKED ISSUE #4288: Replace Resemble.js with Pixelmatch

ahocevar:
This change replaces resemble.js with pixelmaps so renderer tests ignor antialiasing difference between browsers, and adds a `pixelRatio` option to all maps so generated test images have the same size regardless of the browser's device pixel ratio.

Fixes #4288

### DIFF
--- a/test/module-global.js
+++ b/test/module-global.js
@@ -1,0 +1,1 @@
+window.module = {};
--- a/test/test-extensions.js
+++ b/test/test-extensions.js
@@ -386,35 +386,47 @@ goog.require('ol.renderer.webgl.Map');
   };
 
   function resembleCanvas(canvas, referenceImage, tolerance, done) {
-    if (showMap) {
-      var wrapper = document.createElement('div');
-      wrapper.style.width = canvas.width + 'px';
-      wrapper.style.height = canvas.height + 'px';
-      wrapper.appendChild(canvas);
-      document.body.appendChild(wrapper);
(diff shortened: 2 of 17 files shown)


test/test-extensions.js
```diff
@@ -386,35 +386,47 @@ goog.require('ol.renderer.webgl.Map');
...
+        diffCanvas.width = width;
+        diffCanvas.height = height;
+        diffCanvas.getContext('2d').putImageData(output, 0, 0);
+        document.body.appendChild(diffCanvas);
+      }
+      expect(mismatchPct).to.be.below(tolerance);
+      done();
+    });
+    image.addEventListener('error', function() {
+      expect().fail('Reference image could not be lowded');
```
marcjansen: typo: loaded, not lowded

test/module-global.js
```diff
@@ -0,0 +1 @@
+window.module = {};
```
tschaub: Was this left over from debugging?

ahocevar: No, it provides `module` so pixelmatch works without bundling. It has `module.eports = pixelmatch` in it.

PR #7147, opened 2017-08-17
Remove unnecessary import in events.test.js

tschaub:
The `no-missing-requires` rule (correctly) assumes that our use of `ol.events` means we need to import `ol`.  In this case, we are actually using the export from `ol/events`.  Modifying exports like this (in the tests) won't always work, so we will likely need to find another way to test stuff like this eventually.  But for now it should work.


tschaub:
@marcjansen - did your phone break?

tschaub:
@marcjansen?

PR #7159, opened 2017-08-18
Remove bundling magic for the Mapbox styles script

ahocevar:
By passing the style constructors as arguments to the `resources/mapbox-streets-v6-style.js`, we no longer need to require files that are not used.

### DIFF
--- a/examples/resources/mapbox-streets-v6-style.js
+++ b/examples/resources/mapbox-streets-v6-style.js
@@ -1,20 +1,28 @@
 // Styles for the mapbox-streets-v6 vector tile data set. Loosely based on
 // http://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v6.json
 
-function createMapboxStreetsV6Style() {
-  var fill = new ol.style.Fill({color: ''});
-  var stroke = new ol.style.Stroke({color: '', width: 1});
-  var polygon = new ol.style.Style({fill: fill});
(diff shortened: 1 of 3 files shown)


examples/resources/mapbox-streets-v6-style.js
```diff
@@ -1,20 +1,28 @@
...
-  var line = new ol.style.Style({stroke: stroke});
-  var text = new ol.style.Style({text: new ol.style.Text({
+/**
+ * @param {function(olx.StyleOptions):ol.style.Style} Style constructor.
+ * @param {function(olx.FillOptions):ol.style.fill} Fill constructor.
+ * @param {function(olx.StrokeOptions):ol.style.Stroke} Stroke constructor.
+ * @param {function(olx.IconOptions):ol.style.Icon} Icon constructor.
+ * @param {function(olx.TextOptions):ol.style.Text} Text constructor.
+ * @return {function((ol.Feature|ol.render.Feature), number): (ol.style.Style|Array.<ol.style.Style>|undefined)} Style function.
+ */
```
tschaub: If you remove this comment block altogether, does the compiler stop complaining?

ahocevar: Unfortunately it doesn't.

ahocevar:
Giving up on this. Probably not worth the effort. What's going on is that we'd need externs for the `createMapboxStreetsV6Style` function.

ahocevar:
Resurrecting this - found a solution.

PR #7163, opened 2017-08-20
Remove bundling magic for Mapbox styles script

ahocevar:
Another attempt on #7159.

### DIFF
--- a/externs/example.js
+++ b/externs/example.js
@@ -13,7 +13,12 @@ common.getRendererFromQueryString = function(opt_default) {};
 
 
 /**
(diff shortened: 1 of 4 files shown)


ahocevar:
Now finally the examples build with this change.

externs/example.js
```diff
@@ -13,7 +13,12 @@ common.getRendererFromQueryString = function(opt_default) {};
 
 
 /**
+ * @param {function(new:ol.style.Style, olx.style.StyleOptions=)} Style Style.
```
marcjansen: This textual description is a bit misleading, the argument is not a Style, but a Style constructor. here and for the other params... and totally minor

PR #7165, opened 2017-08-20
More assorted test fixes

tschaub:
More changes to make the tests better.

### DIFF
--- a/test/spec/ol/assertionerror.test.js
+++ b/test/spec/ol/assertionerror.test.js
@@ -4,9 +4,9 @@ goog.require('ol');
 goog.require('ol.AssertionError');
 
 describe('ol.AssertionError', function() {
-  it('generates a message', function() {
+  it('generates an error', function() {
     var error = new ol.AssertionError(42);
-    expect(error.message).to.be('Assertion failed. See https://openlayers.org/en/latest/doc/errors/#42 for details.');
+    expect(error).to.be.an(Error);
   });
 
   it('generates a message with a versioned url', function() {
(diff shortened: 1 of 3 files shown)


test/spec/ol/assertionerror.test.js
```diff
@@ -4,9 +4,9 @@ goog.require('ol');
 goog.require('ol.AssertionError');
 
 describe('ol.AssertionError', function() {
-  it('generates a message', function() {
+  it('generates an error', function() {
     var error = new ol.AssertionError(42);
-    expect(error.message).to.be('Assertion failed. See https://openlayers.org/en/latest/doc/errors/#42 for details.');
+    expect(error).to.be.an(Error);
   });
```
tschaub: This test was redundant with the one below, but it fails whenever there is an `ol.VERSION` (so the URL no longer includes "latest").

test/spec/ol/style/iconimagecache.test.js
```diff
@@ -39,22 +39,22 @@ describe('ol.style.IconImageCache', function() {
       expect(cache.cacheSize_).to.eql(4);
 
       src = '4';
-      iconImage = new ol.style.IconImage(src, null);
+      iconImage = new ol.style.IconImage(null, src);
```
tschaub: In https://github.com/openlayers/openlayers/commit/4981cde729d8f749cdb87fe5312ea0d1a780b23a#diff-7004065dfacb69406f951aa64d6e2c55R290, the `IconImage` constructor (then `IconImage_`) was given a new first argument (`img`).  But these tests were not updated.  Since then the tests have been creating icons with a string as the `image_` member.  I haven't dug into why they have been passing since then, but they start failing when we run the tests after using the ES modules transform.  Correcting the arguments to the constructor makes these tests pass again.

ahocevar: Thanks for investigating this! I'm glad things are no longer a mystery here.

PR #7171, opened 2017-08-21
Use the `input` option for rollup

tschaub:
This is a followup to #7166.  The `entry` option has been renamed to `input` (see rollup/rollup#1479).


tschaub:
Closing for now.  We should be able to stop re-bundling dependencies soon...

PR #7174, opened 2017-08-22
[WIP] Reenable testing WebGL

marcjansen:
This PR suggests to use `ol.has.WEBGL` to reenable WebGL tests where they are supported.

On Sauce Labs that is currently not the case, see e.g. here: https://saucelabs.ideas.aha.io/ideas/SLIDEA-I-54 (Comment from Chiarng Lin of Sauce Labs from April 8, 2017, 02:28).

Funny enough for @ahocevar s Mac the tests would execute successfully, for @tschaub s not.

This PR should probably not be merged until we know what is going on.


tschaub:

PR #7187, opened 2017-08-28
Simpler tile pixel handling and faster parsing for ol.format.MVT vector tiles

ahocevar:
This change removes the code paths for `tile-pixels` units handling in the vector tile renderer. Instead, coordinates are now transformed to view projection only once when the features are parsed. This simplifies the vector tile renderer.

To remove some parsing overhead, this pull request also gets rid of the vector-tile dependency. Instead, the pbf is parsed directly in `ol.format.MVT`. This also reduces the build size.

### DIFF
--- a/src/ol/format/mvt.js
+++ b/src/ol/format/mvt.js
@@ -3,8 +3,8 @@
 goog.provide('ol.format.MVT');
 
 goog.require('ol');
+goog.require('ol.asserts');
 goog.require('ol.ext.PBF');
-goog.require('ol.ext.vectortile.VectorTile');
 goog.require('ol.format.Feature');
 goog.require('ol.format.FormatType');
 goog.require('ol.geom.GeometryLayout');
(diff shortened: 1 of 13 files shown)


src/ol/format/mvt.js
```diff
@@ -80,81 +82,270 @@ ol.inherits(ol.format.MVT, ol.format.Feature);
...
+    if (tag === 3) {
+      var layer = {
+        keys: [],
+        values: [],
+        features: []
+      };
+      var end = pbf.readVarint() + pbf.pos;
+      pbf.readFields(ol.format.MVT.pbfReaders_.layer, layer, end);
+      layer.length = layer.features.length;
+      //var layer = new VectorTileLayer(pbf, pbf.readVarint() + pbf.pos);
```
tschaub: Looks like this could be removed.

src/ol/geom/geometry.js
```diff
@@ -244,6 +253,22 @@ ol.geom.Geometry.prototype.translate = function(deltaX, deltaY) {};
...
+      var pixelExtent = source.getExtent();
+      var projectedExtent = source.getWorldExtent();
+      var scale = ol.extent.getHeight(projectedExtent) / ol.extent.getHeight(pixelExtent);
+      ol.transform.compose(tmpTransform,
+          projectedExtent[0], projectedExtent[3],
+          scale, -scale, 0,
+          0, 0);
+      ol.geom.flat.transform.transform2D(inCoordinates, 0, inCoordinates.length, stride,
+          tmpTransform, outCoordinates);
+      return ol.proj.getTransform(source, destination)(inCoordinates, outCoordinates, stride);
```
tschaub, line 269: Looks like there might be a bug here. This works when `inCoordinates` and `outCoordinates` refer to the same array. But I think we should be calling `getTransform(source, destination)(outCoordinates, outCoordinates, stride)` instead.

PR #7190, opened 2017-08-29
Use jsts version 1.4.0 in example

fredj:
ping @bjornharrtell


bjornharrtell:
Thanks @fredj, should be risk free. :)

PR #7191, opened 2017-08-29
Fix provide/require for autogenerated shader files

ahocevar:
This pull request changes the provide/require structure of the autogenerated shader files to comply with the "one provide per module" rule. In addition to that, it reduces the amount of generated classes related to shaders, and transitions the glsl compression utility from Python to JavaScript.


tschaub:
Awesome!  Thanks for tackling this @ahocevar.

PR #7193, opened 2017-08-30
Fix KML links for documents created locally in Safari

ahocevar:
This pull request fixes relative links in KML documents that were locally created in Safari, where `document.baseURI` would be `about:blank` instead of the URL of the document that created the KML.

To make sure this is fixed, I also enabled tests in Safari for Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ branches:
   only:
   - master
 addons:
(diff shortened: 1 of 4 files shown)


marcjansen:
Checks fail due to sauce labs:

image

https://status.saucelabs.com/

PR #7196, opened 2017-08-30
renderSync() to make sure overlay is initially visible

### LINKED ISSUE #7169: Periodically failing overlay test
We have an overlay test that [periodically fails](https://travis-ci.org/openlayers/openlayers/builds/266820014) (perhaps only on Firefox):
```
Firefox 54.0.0 (Windows 7 0.0.0) ol.Overlay #setVisible() changes the CSS display value FAILED
	expected '' to equal 'none'
	Assertion.prototype.assert@/home/travis/build/openlayers/openlayers/node_modules/expect.js/index.js:96:13
	Assertion.prototype.equal@/home/travis/build/openlayers/openlayers/node_modules/expect.js/index.js:216:5
	Assertion/this[name]@/home/travis/build/openlayers/openlayers/node_modules/expect.js/index.js:69:20
	@spec/ol/overlay.test.js:94:7
```

It would be good to figure out what is up and make this more reliable.

ahocevar:
Fixes #7169.

Depending on whether the map was already rendered or not, the first test for `display: 'none'` passed or failed. Now we call `renderSync()` on the map before testing, so we know that the overlay is visible.

### DIFF
--- a/test/spec/ol/overlay.test.js
+++ b/test/spec/ol/overlay.test.js
@@ -91,8 +91,7 @@ describe('ol.Overlay', function() {
         position: [0, 0]
       });
       map.addOverlay(overlay);
-      expect(overlay.element_.style.display).to.be('none');
-      overlay.setVisible(true);
+      map.renderSync();
       expect(overlay.element_.style.display).not.to.be('none');
       overlay.setVisible(false);
       expect(overlay.element_.style.display).to.be('none');


test/spec/ol/overlay.test.js
```diff
@@ -91,8 +91,7 @@ describe('ol.Overlay', function() {
         position: [0, 0]
       });
       map.addOverlay(overlay);
-      expect(overlay.element_.style.display).to.be('none');
```
marcjansen: Unsure why you removed this test

ahocevar: Previously, we expected the initial `display` to be `'none'` because we expected the map not yet to be rendered. But sometimes it was already rendered. So now, instead, we ensure the map is rendered, and we expect the `display` initially to be not `'none'`.

marcjansen: makes sense

PR #7198, opened 2017-08-31
Use geometry name in WFS updates

### LINKED ISSUE #7197: ol.format.WFS write update does not use geometry name
As noticed by @theduckylittle ```ol.format.WFS``` does not always correctly handle the geometry name

bartvde:
closes #7197


bartvde:
thanks for the review @ahocevar

PR #7217, opened 2017-09-04
Mark ol.format.filter.Spatial as abstract class

TDesjardins:
Add missing abstract class description for class ol.format.filter.Spatial like in all other abstract classes.


fredj:
@TDesjardins see #7220, I've took your change and added the missing `@abstract` tag to the others abstract classes. Thanks !

ahocevar:
Replaced by #7220.

PR #7219, opened 2017-09-05
Change cartodb domain from cartodb.com to carto.com

### LINKED ISSUE #7216: Fix carto endpoint
Since July 2016 all CARTO accounts work with `carto.com` domain and only the accounts created before that date work with `cartodb.com` so the source in OL needs to be updated.

It looks like it's just about updating [this line](https://github.com/openlayers/openlayers/blob/master/src/ol/source/cartodb.js#L105) but maybe @aisaacs  (as apparently the main author of the code) can confirm.

Let me know if I can help on testing or whatever.

### LINKED ISSUE #7218: CartoDb Source example  Response for preflight is invalid (redirect) error
Hello, 
I saw CartoDb source example on openlayers web site and create a map on carto. It's seem completely what I want on carto map. But when I  tried on my c# application with openlayers, I'm getting this console error:

    XMLHttpRequest cannot load https://seymacengiz.cartodb.com/api/v1/map. Response for preflight is invalid (redirect)

This map include only one row, a point data. Also at first, it was working. So,What's changed?Is this problem because of codes or map settings?

       var mapConfig = {
            'layers': [{
                'type': 'cartodb',
                'options': {
                    'cartocss_version': '2.1.1',
                    'cartocss': '#layer { marker-width: 7; marker-fill: #FFB927; marker-fill-opacity: 0.9; marker-line-color: #FFF; marker-line-width: 1; marker-line-opacity: 1; marker-placement: point; marker-type: ellipse; marker-allow-overlap: true; }',
                    'sql': 'select * from isim'
                }
            }]
        };

        var cartoDBSource = new ol.source.CartoDB({
            account: 'seymacengiz' ,
            config: mapConfig
        });

        var map = new ol.Map({
            layers: [
              new ol.layer.Tile({
                  source: new ol.source.OSM()
              }),
              new ol.layer.Tile({
                  source: cartoDBSource
              })
            ],
            target: 'map',
            view: new ol.View({
                center: [0, 0],
                zoom: 2
            })
        });

fredj:
fixes #7216

fixes #7218

### DIFF
--- a/src/ol/source/cartodb.js
+++ b/src/ol/source/cartodb.js
@@ -8,7 +8,7 @@ goog.require('ol.source.XYZ');
 
 /**
  * @classdesc
- * Layer source for the CartoDB tiles.
+ * Layer source for the CartoDB Maps API.
  *
  * @constructor
  * @extends {ol.source.XYZ}
@@ -102,7 +102,7 @@ ol.source.CartoDB.prototype.initializeMap_ = function() {
     this.applyTemplate_(this.templateCache_[paramHash]);
     return;
   }
-  var mapUrl = 'https://' + this.account_ + '.cartodb.com/api/v1/map';
+  var mapUrl = 'https://' + this.account_ + '.carto.com/api/v1/map';
 
   if (this.mapId_) {
     mapUrl += '/named/' + this.mapId_;


fredj:
Thanks for the review

PR #7220, opened 2017-09-05
Mark ol.format.filter.Spatial as abstract class

fredj:
fixes #7217 and add missing `@abstract` tag


fredj:
Yes, it make sense to remove the `@api` annotation but will it require a new major release (5.0.0) ?

ahocevar:
Hm. Well, you cannot add a `@deprecated` annotation, because you will get a compiler warning. But you could mention that those are deprecated in their docs. And we can remove the `@api` annotation later, i.e. before we release 5.0.

fredj:
@ahocevar I've added the deprecation warning in the doc

fredj:
Thanks for the review

PR #7225, opened 2017-09-06
Fix hit detection for image layers

### LINKED ISSUE #7222: map.forEachLayerAtPixel misses layers
Based on the numerous issues, you guys are probably sick of this, but I'm also still having issues with the map.forEachLayerAtPixel not including an ol.layer.Image layer with an ImageStatic source on a canvas map. If I zoom in enough so the image is quite large it seems to work better than when I'm more zoomed out. I believe this broke like many others shortly after 3.18.2. 

I'm going to try to use a custom layerFilter function and see if that will help get the layer, but I doubt it based on your docs. I'll also try to get you guys a fiddle or something showing the issue if I can. Thanks for all your guys great work on this.

After a little further investigation it does appear this works perfectly with webGL. However, webGL is not wrapping the geoJSON vector map and isn't always filling in the polygons (countries) for whatever reason at this point so.

The images are PNG images that tend to have a fair bit of transparency, i.e. aren't usually very square.

ahocevar:
Fixes #7222.


tschaub:
Sounds like this addresses the issue that @sbowler raised in #7222.

ahocevar:
Regression test added, waiting for Travis to finish.

sbowler:
Just to add to this here, I have manually tested the fix and it seems good. I did not notice any other issues with any of our mapping.

marcjansen:
Here are your bonus points, @ahocevar ⚪⚫🔴, in case you need them in the future. Nice fix!

PR #7230, opened 2017-09-06
Remove unused context handling for ol.Image

ahocevar:
This pull request removes a leftover from when we had a DOM renderer.


tschaub:
Coveralls failed to respond.  This should be good to merge though.

PR #7234, opened 2017-09-07
Fix abort handling of tileload events

ahocevar:
Tile loading events, like used in the `tile-load-events.html` example, get out of sync between loading and loaded when tiles are aborted, e.g. because of a small and full tile cache.

By keeping track of which tiles are being loaded, and also dispatching a tileloadend event for aborted tiles, this issue can be fixed.

### DIFF
--- a/src/ol/source/urltile.js
+++ b/src/ol/source/urltile.js
@@ -61,6 +61,12 @@ ol.source.UrlTile = function(options) {
     this.setTileUrlFunction(options.tileUrlFunction);
   }
 
+  /**
+   * @private
+   * @type {Object.<number, boolean>}
+   */
+  this.tileLoadingKeys_ = {};
+
 };
 ol.inherits(ol.source.UrlTile, ol.source.Tile);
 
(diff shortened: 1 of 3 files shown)


src/ol/source/urltile.js
```diff
@@ -110,21 +116,20 @@ ol.source.UrlTile.prototype.getUrls = function() {
...
-    default:
-      // pass
+  var uid = ol.getUid(tile);
+  var tileState = tile.getState();
+  var type;
+  if (tileState == ol.TileState.LOADING) {
+    this.tileLoadingKeys_[uid] = true;
+    type = ol.source.TileEventType.TILELOADSTART;
+  } else if (uid in this.tileLoadingKeys_) {
+    delete this.tileLoadingKeys_[ol.getUid(tile)];
```
tschaub: You could reuse `uid` here.

test/spec/ol/source/tileimage.test.js
```diff
@@ -172,4 +173,67 @@ describe('ol.source.TileImage', function() {
...
+      source.on('tileloaderror', function(e) {
+        setTimeout(function() {
+          e.tile.setState(ol.TileState.LOADING);
+          e.tile.setState(ol.TileState.LOADED);
+        }, 0);
+        errorSpy();
+      });
+      source.on('tileloadend', function() {
+        expect(startSpy.callCount).to.be(2);
+        expect(errorSpy.callCount).to.be(1);
```
tschaub: This takes a while to understand.  And I'm still not sure I get it.  It looks like the tile state is set to `LOADED` twice in this test (once at 204 and once at 212).  But `tileloadend` only gets called once?  Or will it get called a second time but we are already `done()`?

I wonder if there is a simpler test to confirm that `ERROR` state results in both `tileloaderror` and `tileloadend` being fired (if that is the correct behavior).

tschaub: Does this pass?

```js
source.setTileLoadFunction(function(tile) {
  tile.setState(ol.TileState.ERROR);
});

var errorSpy = sinon.spy();
source.on('tileloaderror', errorSpy);
source.on('tileloadend', function() {
  expect(errorSpy.callCount).to.be(1);
  done();
});
```

tschaub: Actually, maybe that isn't what this is about.

ahocevar: The life cycle of the tile is `IDLE` - `LOADING` - `ERROR` - `LOADING` - `LOADED`.

I think this is reasonable, because users might respond to `ERROR` by loading the tile again, finally resulting in a `LOADED` state. L211-L212 mimics what a user would do to reload the tile. In a real application, the `LOADING` state would be set again by calling [`ol.ImageTile#load()`](https://github.com/openlayers/openlayers/blob/2a117cce8f08f9f31306fa4d5ae5d645b1508aa3/src/ol/imagetile.js#L125-L137), which sets up `load` and `error` listeners for an `Image`.

tschaub: Got it.  I started with the test description (it "dispatches tileerror events") and was just surprised/confused by how involved the sequence was to assert that.

ahocevar:
Thanks for the review. I updated the test description to make it clearer what it tests.

PR #7236, opened 2017-09-08
reusing images in ol.style.Icon#clone

### LINKED ISSUE #6583: Clone style doesn't work for ol.style.Icon using Internet Explorer 11
I have my style defined as  (simplified):
```js
var style = new ol.style.Style({
                    image: new ol.style.Icon({
                        src: 'images/preview_nomove_point.png',
                        anchor: [0.5, 0.5],
                        anchorXUnits: "fraction",
                        anchorYUnits: "fraction",
                        anchorOrigin: "top-left",
                        rotation: 0,
                        scale: 0.5
                    })
                });

return style;
```

In case I return it from my style function like this, it works fine. In case I return

style.clone();

the icon is not visible on the map.

simonseyock:
Fixes #6583 

Cloned ol.style.Icon now reuse the underlying image.

### DIFF
--- a/src/ol/style/icon.js
+++ b/src/ol/style/icon.js
@@ -176,24 +176,11 @@ ol.inherits(ol.style.Icon, ol.style.Image);
 
 
 /**
- * Clones the style.
+ * Clones the style. The underlying Image/HTMLCanvasElement is not cloned.
  * @return {ol.style.Icon} The cloned style.
  * @api
  */
 ol.style.Icon.prototype.clone = function() {
-  var oldImage = this.getImage(1);
-  var newImage;
-  if (this.iconImage_.getImageState() === ol.ImageState.LOADED) {
-    if (oldImage.tagName.toUpperCase() === 'IMG') {
-      newImage = /** @type {Image} */ (oldImage.cloneNode(true));
-    } else {
-      newImage = /** @type {HTMLCanvasElement} */ (document.createElement('canvas'));
-      var context = newImage.getContext('2d');
-      newImage.width = oldImage.width;
-      newImage.height = oldImage.height;
-      context.drawImage(oldImage, 0, 0);
-    }
-  }
   return new ol.style.Icon({
     anchor: this.anchor_.slice(),
     anchorOrigin: this.anchorOrigin_,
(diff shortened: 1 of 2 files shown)


src/ol/style/icon.js
```diff
@@ -176,34 +176,19 @@ ol.inherits(ol.style.Icon, ol.style.Image);
...
     anchor: this.anchor_.slice(),
     anchorOrigin: this.anchorOrigin_,
     anchorXUnits: this.anchorXUnits_,
     anchorYUnits: this.anchorYUnits_,
     crossOrigin: this.crossOrigin_,
     color: (this.color_ && this.color_.slice) ? this.color_.slice() : this.color_ || undefined,
-    img: newImage ? newImage : undefined,
-    imgSize: newImage ? this.iconImage_.getSize().slice() : undefined,
-    src: newImage ? undefined : this.getSrc(),
+    src: this.getSrc(),
```
simonseyock: I am making use of the ol.style.iconCache here that will provide the clone with the same image via the src.

test/spec/ol/style/icon.test.js
```diff
@@ -91,33 +92,29 @@ describe('ol.style.Icon', function() {
         src: src
       });
       var clone2 = original2.clone();
-      expect(original2.getSrc()).to.be(clone2.getSrc());
+      expect(clone2.iconImage_.getImageState() === ol.ImageState.LOADING);
```
ahocevar: Interesting. I would expect the state to be the same as on the original. Is the original also still loading?

simonseyock: Yes. It is loaded as soon as the browser gets some computation time.
But maybe it would be better to give a real http url here?

simonseyock: or maybe better use `expect(clone2.iconImage_.getImageState()).to.eql(original2.iconImage_.getImageState());`

simonseyock:
I think this tests are better because they just check if the style is correctly cloned and not something about the internal state of the images. If you agree I can rebase to a single commit.

ahocevar:
Yes, that looks good. Please rebase.

simonseyock:
done

PR #7242, opened 2017-09-11
Use EMPTY and LOADED state properly on ol.VectorImageTile

ahocevar:
Currently, vector image tiles with ERROR source tiles are set to EMPTY state. But the EMPTY state means that the renderer has nothing to do for the extent of that tile. Instead, we want the `loadInterimTilesOnError` logic to kick in when there are error tiles, so tiles from lower resolutions can fill the spots with error tiles. For this to work, we have to set the state to ERROR instead.

With #7206 fixed, we also no longer need to set a vector image tile to LOADED when only some of its source tiles are loaded, because source tiles will only be created if they are available on the source. By setting an ERROR state when just one source tile has an ERROR state, we ensure that the `loadInterimTilesOnError` logic works here too.

### DIFF
--- a/src/ol/vectorimagetile.js
+++ b/src/ol/vectorimagetile.js
@@ -273,7 +273,7 @@ ol.VectorImageTile.prototype.finishLoading_ = function() {
     this.loadListenerKeys_.forEach(ol.events.unlistenByKey);
     this.loadListenerKeys_.length = 0;
   }
-  this.setState(loaded > 0 ? ol.TileState.LOADED : ol.TileState.EMPTY);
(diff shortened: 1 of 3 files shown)


ahocevar:
To see the issue that this pull request fixes, open the https://openlayers.org/en/master/examples/mapbox-vector-tiles.html example, and enter
```js
map.getLayers().item(0).getSource().setUrl(
    'https://{a-e}.tiles.mapbox.com/v4/mapbox.mapbox-streets-v6/{z}/{x}/{y}.vector.pbf?' + 
    'access_token=' + key);
```
in the console. This adds a non-existent server (`e.tiles.mapbox.com`) to cause loading errors. Then zoom in to a region that you're interested in, and you will see blank tiles.

PR #7181, opened 2017-08-27
(WIP) Updates to the dev environment for working with modules

tschaub:
The purpose of this branch is to migrate our development tooling so that we can work with ES modules directly.

Ignore the first commit ("Transformed").  The changes in this commit are generated by running `npm run transform`.  I'll be rebasing this branch on `master` frequently and reworking this first commit based on changes to the transform script.  The focus of this branch is the commits on top of that first commit.

Things that need to be working:
 - [x] a script to build/serve examples in development mode (`npm run serve-examples`)
 - [x] a script to build examples for production (`npm run build-examples`)
 - [x] a script that runs the tests (`npm test`)
 - [ ] a script that builds the docs
 - [ ] a script that builds with the Closure Compiler
 - [ ] an equivalent of `make ci` for Travis

Assorted blocking TODOs:
 - [x] fix `test-extensions.js` to avoid `instanceof` checks (see #7182);
 - [x] fix `bin/pyglslunit.py` and shader templates so they work with the "one provide per module" rule (see #7191)
 - [x] write a script that renames files so they are cased like the default export (e.g. `ol/Map.js`) (see #7249)
 - [x] update the module transform to use `.js` (see #7223)
 - [x] update the module transform to add [`@module` annotations](http://usejsdoc.org/howto-es2015-modules.html) (see #7223)
 - [ ] update the module transform to rewrite type annotations [using relative names](https://github.com/google/closure-compiler/wiki/JS-Modules)

For people interested in experimenting with ES Modules and the Closure Compiler, the key is setting `module_resolution: 'NODE'`.  Here is an example build script that you can start with: https://github.com/openlayers/closure-es-modules/blob/master/build.js

PR #7249, opened 2017-09-12
Script to rename files so the case matches the module name

ahocevar:
This pull request modifies the `npm run transform` script so the module file names match the case of the default export.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "changecase-src": "node tasks/filename-case-from-module.js",
     "transform-examples": "jscodeshift --transform transforms/module.js examples",
     "transform-test": "jscodeshift --transform transforms/module.js test",
-    "transform": "npm run transform-src && npm run changecase-src && npm run transform-examples && npm run transform-test && npm run lint -- --fix"
+    "transform": "npm run changecase-src && npm run transform-src && npm run transform-examples && npm run transform-test && npm run lint -- --fix"
   },
   "main": "dist/ol.js",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -19,9 +19,10 @@
     "debug-server": "node tasks/serve-lib.js",
     "karma": "node tasks/test.js start test/karma.config.js",
     "transform-src": "jscodeshift --transform transforms/module.js src",
+    "changecase-src": "node tasks/filename-case-from-module.js",
     "transform-examples": "jscodeshift --transform transforms/module.js examples",
     "transform-test": "jscodeshift --transform transforms/module.js test",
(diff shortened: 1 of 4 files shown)


gberaudo:
@ahocevar, great to see this change.
Do you need the script part? I would expect the names to be correct after your change in the transform.

marcjansen:
Will this mean that users of the ol package have to change their imports? If so we should tell people this, right?

ahocevar:
@marcjansen I need to change this pull request. It was meant to accompany #7181, and not make changes to the ol package for now. For changes people need to make once what we've outlined in #7181 is in place, we are going to create a roadmap page with instructions on the wiki.

ahocevar:
Pull request updated.

package.json
```diff
@@ -19,9 +19,10 @@
     "debug-server": "node tasks/serve-lib.js",
     "karma": "node tasks/test.js start test/karma.config.js",
     "transform-src": "jscodeshift --transform transforms/module.js src",
+    "changecase-src": "node tasks/filename-case-from-module.js",
     "transform-examples": "jscodeshift --transform transforms/module.js examples",
     "transform-test": "jscodeshift --transform transforms/module.js test",
-    "transform": "npm run transform-src && npm run transform-examples && npm run transform-test && npm run lint -- --fix"
+    "transform": "npm run transform-src && npm run changecase-src && npm run transform-examples && npm run transform-test && npm run lint -- --fix"
```
tschaub: I think we should be changing the case of file names *before* running the module transform on them.  In the end, we want `src/ol/Map.js` to look like:
```js
/**
 * @module ol/Map
 */
import _ol_View_ from './View.js';
// etc.
```

ahocevar: Good point. Changing.

ahocevar: Done, commit added.

ahocevar:
Strange Travis issue when building api docs. I'm sure it's unrelated to this pull request, but am unable to investigate this now.

tschaub:
See https://github.com/jsdoc3/jsdoc/issues/1438 for the build failure.

ahocevar:
Thanks @tschaub for investigating the Travis issue. I've set the node version to 8.4 in `.travis.yml` for now.

tschaub:
> Will this mean that users of the ol package have to change their imports? If so we should tell people this, right?

@marcjansen - you were right.  We should have left out the changes to `transforms/module.js` until 5.0.  I published `ol@4.4.0` with this change before realizing that it is breaking (on case sensitive filesystems).  I put up #7313 to fix and will get out a patch release when that is in.

tschaub:
In the meantime, `ol@4.4.1-beta.1` is out (I can't wait until the whole release process is as easy as publishing the `ol` package is now).

PR #7252, opened 2017-09-13
fix osmxml to read ways before the definition of nodes

### LINKED ISSUE #7250: osmxml cannot read ways before the definition of nodes
While using OSMXML to read osm file, if the ways is defined before nodes, it will reads the nodes in the ways as undefined:

```
    it('can read ways before nodes', function() {
      var text =
          '<?xml version="1.0" encoding="UTF-8"?>' +
          '<osm version="0.6" generator="my hand">' +
          '  <way id="3">' +
          '    <tag k="name" v="3"/>' +
          '    <nd ref="1" />' +
          '    <nd ref="2" />' +
          '  </way>' +
          '  <node id="1" lat="1" lon="2">' +
          '    <tag k="name" v="1"/>' +
          '  </node>' +
          '  <node id="2" lat="3" lon="4">' +
          '    <tag k="name" v="2"/>' +
          '  </node>' +
          '</osm>';
      var fs = format.readFeatures(text);
      expect(fs).to.have.length(3);
      var line = fs[0];
      expect(line).to.be.an(ol.Feature);
      var g = line.getGeometry();
      expect(g).to.be.an(ol.geom.LineString);
      expect(g.getCoordinates()).to.eql([[2, 1], [4, 3]]);
    });
```

It will get error: 

```
Error: expected [ [ undefined, undefined ] ] to sort of equal [ [ 2, 1 ], [ 4, 3 ] ] 
```

I'm not very familiar with the OSM XML format. But the file generated by Merkaartor may define ways before their nodes. So if you confirm this as a bug, I think I can help to fix this.

wb14123:
This fixes #7250

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-- '8'
+- '8.4'
 cache:
   directories:
   - node_modules
--- a/src/ol/format/osmxml.js
+++ b/src/ol/format/osmxml.js
@@ -71,34 +71,14 @@ ol.format.OSMXML.readNode_ = function(node, objectStack) {
  * @private
  */
 ol.format.OSMXML.readWay_ = function(node, objectStack) {
-  var options = /** @type {olx.format.ReadOptions} */ (objectStack[0]);
   var id = node.getAttribute('id');
   var values = ol.xml.pushParseAndPop({
+    id: id,
     ndrefs: [],
     tags: {}
   }, ol.format.OSMXML.WAY_PARSERS_, node, objectStack);
   var state = /** @type {Object} */ (objectStack[objectStack.length - 1]);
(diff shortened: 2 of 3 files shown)


wb14123:
Hi, I've run `make check` on my computer and it has passed. I'm not sure if the CI build failed is related with this PR.

ahocevar:
@wb14123 The failing CI build is caused by the update from Node 8.4 to 8.5. It happened in other pull requests too. In one of my pull requests, I temporarily fixed it with 1d84bc52c6740fe98e7d17fce86fe1612f9040b3. You can cherry-pick that if you want the build to pass.

ahocevar:
Thank you @wb14123 for this excellent contribution. Can you please submit a [CLA](https://github.com/openlayers/cla) so we can merge this?

wb14123:
I've signed it.

PR #7267, opened 2017-09-16
Render tiles with an opacity transition

### LINKED ISSUE #594: Tile fading
<bountysource-plugin>

---

Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/1449635-tile-fading?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tschaub:
This makes it so tiles render with an opacity transition.

fade

Tasks:
 - [x] make the transition configurable (or at least disable-able) from the source (feels like a layer option, but code-wise it is closer to the source)
 - [x] fix up failing render tests (should involve disabling the transition)
 - [x] make sure everything else works

Fixes #594.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -472,7 +472,7 @@ olx.TileOptions;
 
 /**
  * A duration for tile opacity transitions.  By default, tiles will render with
- * a opacity transition that lasts 275 ms.  To change the duration, pass a
+ * an opacity transition that lasts 275 ms.  To change the duration, pass a
  * number in milliseconds.  A duration of 0 disables the opacity transition.
  * @type {number|undefined}
  * @api
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -472,7 +472,7 @@ olx.TileOptions;
 
 /**
  * A duration for tile opacity transitions.  By default, tiles will render with
- * an opacity transition that lasts 275 ms.  To change the duration, pass a
+ * an opacity transition that lasts 250 ms.  To change the duration, pass a
  * number in milliseconds.  A duration of 0 disables the opacity transition.
  * @type {number|undefined}
  * @api
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -464,6 +464,23 @@ olx.SphereMetricOptions.prototype.radius;
 
 
 /**
+ * Options for tile constructors.
+ * @typedef {{transition: (number|undefined)}}
+ */
+olx.TileOptions;
+
+
+/**
+ * A duration for tile opacity transitions.  By default, tiles will render with
+ * a opacity transition that lasts 250 ms.  To change the duration, pass a
+ * number in milliseconds.  A duration of 0 disables the opacity transition.
+ * @type {number|undefined}
+ * @api
+ */
+olx.TileOptions.prototype.transition;
+
+
+/**
  * Object literal with options for the {@link ol.Map#forEachFeatureAtPixel} and
  * {@link ol.Map#hasFeatureAtPixel} methods.
  * @typedef {{layerFilter: ((function(ol.layer.Layer): boolean)|undefined),
@@ -4597,7 +4614,8 @@ olx.source;
  *     maxZoom: (number|undefined),
  *     reprojectionErrorThreshold: (number|undefined),
  *     tileLoadFunction: (ol.TileLoadFunctionType|undefined),
- *     wrapX: (boolean|undefined)}}
+ *     wrapX: (boolean|undefined),
+ *     transition: (number|undefined)}}
  */
 olx.source.BingMapsOptions;
 
@@ -4679,6 +4697,15 @@ olx.source.BingMapsOptions.prototype.tileLoadFunction;
  * @api
  */
 olx.source.BingMapsOptions.prototype.wrapX;
+
+
+/**
+ * Duration of the opacity transition for rendering.  To disable the opacity
+ * transition, pass `transition: 0`.
+ * @type {number|undefined}
+ * @api
+ */
+olx.source.BingMapsOptions.prototype.transition;
 
 
 /**
@@ -4845,7 +4872,8 @@ olx.source.TileUTFGridOptions.prototype.url;
  *            tileUrlFunction: (ol.TileUrlFunctionType|undefined),
  *            url: (string|undefined),
  *            urls: (Array.<string>|undefined),
- *            wrapX: (boolean|undefined)}}
+ *            wrapX: (boolean|undefined),
+ *            transition: (number|undefined)}}
  */
 olx.source.TileImageOptions;
 
@@ -4999,6 +5027,15 @@ olx.source.TileImageOptions.prototype.wrapX;
 
 
 /**
+ * Duration of the opacity transition for rendering.  To disable the opacity
+ * transition, pass `transition: 0`.
+ * @type {number|undefined}
+ * @api
+ */
+olx.source.TileImageOptions.prototype.transition;
+
+
+/**
  * @typedef {{attributions: (ol.AttributionLike|undefined),
  *            cacheSize: (number|undefined),
  *            format: (ol.format.Feature|undefined),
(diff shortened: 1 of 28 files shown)


tschaub:
Under the "make sure everything else works" category, the canvas still needs to be cleared in some cases.  I've removed some unnecessary canvas clearing, but we still need to clear when the rendered extent changes.  And I am seeing some occasional unwanted flashes during rendering that would be good to track down.

src/ol/renderer/canvas/tilelayer.js
```diff
@@ -183,7 +183,8 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame = function(frameState, layer
       if (this.isDrawableTile_(tile)) {
         if (tile.getState() == ol.TileState.LOADED) {
           tilesToDrawByZ[z][tile.tileCoord.toString()] = tile;
-          if (!newTiles && this.renderedTiles.indexOf(tile) == -1) {
+          var inTransition = tile.getAlpha && tile.getAlpha(ol.getUid(this), frameState.time) !== 1;
+          if (!newTiles && (this.renderedTiles.indexOf(tile) === -1 || inTransition)) {
```
tschaub: The second condition can be flipped to avoid the unnecessary `indexOf` lookup when in transition.

tschaub:
I've reworked the strategy for finding alt tiles to render when a tile is not loaded or still in transition.  Previously, we would favor lower zoom level tiles, only rendering higher zoom level tiles if the lower level ones did not cover the target tile.

This leads to ugly rendering when zooming out after having previously loaded low zoom level tiles.  For example, this is an animation of the [vector tile example](http://openlayers.org/en/master/examples/mapbox-vector-tiles.html) on master:

before-zooming-out

Here I'm slowly zooming out to emphasize the problem.  You can see that the low zoom level tiles cover up the previously rendered high zoom level tiles as you zoom out.  This is not very nice.

In this branch, we only render low zoom level tiles if there aren't tiles from a higher zoom level that cover the target tile.  Here is an animation of the same example from this branch:

after-zooming-out

tschaub:
Ok, I've got more confidence in this now and would appreciate some additional eyes on it.  The transition is pretty subtle.  I've played around with longer durations and think that they make tile loading feel sluggish.  So I've settled on something that I think makes tile rendering less jarring while still feeling fast.  Also, the opacity transition only applies for the first render of a tile - when it happens depends a bit on how we manage cached tiles.

Thanks for any reviews.

tschaub:
Thanks for taking a look @ahocevar.  I knew this problem was lurking, but hadn't seen it yet and forgot to address it.  If I'm right about what you were seeing, 96659c424c9c9696ec06b275bbe6e55fee9258a8 should fix it.

externs/olx.js
```diff
@@ -464,6 +464,23 @@ olx.SphereMetricOptions.prototype.radius;
...
 /**
+ * Options for tile constructors.
+ * @typedef {{transition: (number|undefined)}}
+ */
+olx.TileOptions;
+
+
+/**
+ * A duration for tile opacity transitions.  By default, tiles will render with
+ * a opacity transition that lasts 275 ms.  To change the duration, pass a
```
marcjansen: typo: a**n** opacity transition.

src/ol/renderer/canvas/tilelayer.js
```diff
@@ -1,13 +1,10 @@
-// FIXME find correct globalCompositeOperation
```
marcjansen: Is the above fixed by this PR, or just no longer relevant?

src/ol/renderer/canvas/tilelayer.js
```diff
@@ -225,15 +227,25 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame = function(frameState, layer
...
     /** @type {Array.<number>} */
     var zs = Object.keys(tilesToDrawByZ).map(Number);
-    zs.sort(ol.array.numberSafeCompareFunction);
+    zs.sort(function(a, b) {
+      if (a === z) {
+        return 1;
+      } else if (b === z) {
+        return -1;
+      } else {
+        return a > b ? 1 : a < b ? -1 : 0;
```
marcjansen: Should we use `ol.array.numberSafeCompareFunction` here?

test/spec/ol/tile.test.js
```diff
@@ -1,10 +1,72 @@
...
+      expect(startAlpha < 1).to.be(true);
+
+      time += tile.transition_ / 2;
+      var midAlpha = tile.getAlpha(id, time);
+      expect(midAlpha > startAlpha).to.be(true);
+      expect(midAlpha < 1).to.be(true);
+
+      time += tile.transition_ / 2;
+      var endAlpha = tile.getAlpha(id, time);
+      expect(endAlpha).to.be(1);
```
marcjansen: Does this only pass because of the default of 275ms? It'd be nice if the default could particiate in the test, IMO. But I might also just be misreading it.

ahocevar:
@tschaub: 96659c4 does fix the issue for image tiles (including pre-rendered images of vector tiles), but the same (or a similar) issue persists for vector tiles (i.e. symbols and labels):

labels

src/ol/renderer/canvas/tilelayer.js
tschaub: We're using the correct `globalCompositeOperation` already (`'source-over'`).  There are situations where `'copy'` is useful, but not in this code where we are only rendering to a portion of the canvas at a time.

src/ol/renderer/canvas/tilelayer.js
tschaub: My personal opinion is that it is better to have some copied code in cases like this.  I would use `ol.array.numberSafeCompareFunction()` if 1) it was always going to be inlined (this is probably only going to be the case with Closure Compiler), or 2) it involved something more than this trivial expression.

test/spec/ol/tile.test.js
tschaub: This is true because we are calling`tile.getAlpha(id, time)` with a time delta that is equal to `tile.transition_`.  The assertion would pass if this value were something different than 275. We're not relying on any delta in clock time, just a delta in the `time` between the first call to `getAlpha` and this one.

tschaub:
@ahocevar let me know if you can reproduce the issue after 872e6a0f5797cf11eca990acae43cd65b0ee0b01.

ahocevar:
@tschaub The issue persists after 872e6a0. I think something else must be causing this, because the labels and symbols are drawn to the map canvas after the intermediate canvas with the tile images, in `ol.renderer.canvas.VectorTileLayer.prototype.postCompose`.

src/ol/renderer/canvas/tilelayer.js
marcjansen: OK

test/spec/ol/tile.test.js
marcjansen: Got it.

tschaub:
Thanks for the update.  I've got an understanding of the issue and will rework the "in transition" handling to address it.

tschaub:
@ahocevar - I wasn't able to reproduce the issue you are seeing, but I can imagine why it was happening: we were including alt tiles for rendering even at the end of a transition.  So the last tiles to reach alpha 1 would also have higher z tiles rendered.  This wasn't a problem for raster tiles since we clear the canvas before rendering alpha 1 tiles.  But it was a problem for vector tile elements getting rendered post-compose.

So we now consider a tile "in transition" if it hasn't yet been rendered at alpha 1.  When a tile is in transition, we can't bail out of pre-compose early (so we set `newTiles` - this could be renamed to `moreToRender` or something).  When a tile is in transition and alpha < 1, we look for alt tiles.  When a tile is in transition and at alpha 1, we skip the alt tile lookup.  After rendering the tile at alpha 1, we end the transition.

I've fiddled a bit with using tile alpha in the vector tile layer renderer's post-compose.  But there is some unwanted clearing before rendering new tiles that makes this not so nice.  I think we can improve on the transition for vector tiles, but if there aren't any more issues with this now, it could make sense to get it in and keep iterating on it.

PR #7274, opened 2017-09-20
Add ability to change the loader of a vector source

bartvde:
Currently there is no option to change the loader of a vector source. This PR adds that ability.


bartvde:
thanks for the review @tschaub

PR #7285, opened 2017-09-23
Convert pointerId to string for object lookups

### LINKED ISSUE #7202: Map freezing with iOS 11 on ipad mini
I was testing an app on iOS11 beta and noticed some performance issues with rotation. Then I tried some websites that are using opnelayers and most of the time the maps are freezing as soon as there is an interaction (zoom in, rotate).

Here is an example: https://wow.metoffice.gov.uk/

So I'm just wondering if there is an issue with the new version of iOS.

Thanks,
Alessandro

tschaub:
We're using the expression [`property in object`](https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/pointer.js#L131) where `property` is a number, and apparently Safari on IOS11 doesn't like that when the number is negative.  Although `event.pointerId` [is a number](https://www.w3.org/TR/pointerevents/#widl-PointerEvent-pointerId) it is legitimate to convert to a string when we use it as an object property name.  So even if this is a Safari bug, I think it is a legitimate change to make.

Thanks @ogo-adp for debugging and identifying the fix.

Fixes #7202.

### DIFF
--- a/src/ol/interaction/pointer.js
+++ b/src/ol/interaction/pointer.js
@@ -69,7 +69,7 @@ ol.interaction.Pointer = function(opt_options) {
   this.handlingDownUpSequence = false;
 
   /**
-   * @type {Object.<number, ol.pointer.PointerEvent>}
+   * @type {Object.<string, ol.pointer.PointerEvent>}
    * @private
    */
   this.trackedPointers_ = {};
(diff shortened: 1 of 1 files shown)


wayofthefuture:
You guys are a godsend if this works!

Issue #7284, opened 2017-09-23
unnecessary opt_this / thisArg parameters


simonseyock:
In some places in the openlayers source code there are unnecessary opt_this / thisArg paramters. All of these could be replaced by use of the javascript builtin .bind method of functions. Additionally es6 and es6 transpilers offer the locally bound arrow functions which even lessen the need for this parameters.

Smaller function signatures are an advantage, also the code gets a little more coherent.

Is a PR wanted?

https://github.com/openlayers/openlayers/search?q=opt_this
https://github.com/openlayers/openlayers/search?q=thisArg

marcjansen:
I would love to habe less this param in the methods. I'd be plus 1, but I think others should weigh in as well. This would also be an API breaking change, and would force us to release a 5.0.0

simonseyock:
I could prepare it and rebase it when enough PRs are gathered to justify a version 5.0.0, if this change is not enough.

tschaub:
@simonseyock - this would be appreciated.  We're working toward a 5.0 release now (details on the wiki next week) and plan to do one more 4.x release with recent changes.  So this shouldn't have to wait for too long.

simonseyock:
This will also make the bindTo property of the listenerObj unneeded.

simonseyock:
I just started to work on it a little bit, so you can see what it looks like.

simonseyock:
I started to work on it a little bit so you can see how it would look like.

simonseyock:
We would need to decide how to deal with eventlisteners that are class methods. I see two possibilities:
1. bind the listener at the point where `ol.events.listen` is called. This causes a new function to be created each time this part is called. but as these functions are really small I don't think it is much of a problem. On the other hand it is not easy to use `ol.events.unlisten` anymore without storing the key and use `ol.events.unlistenByKey` (what is better anyhow in my opinion)
2. bind the listener in the constructor. It could look like this: `this.boundHandleSomething_ = this.handleSomething.bind(this)` or even `this.handleSomething_ = this.handleSomething_.bind(this)` as the method is useless without `this` anyways. This produces new code noise, but not in the function parameters what i would prefer.

simonseyock:
Like this for example:

```
ol.control.MousePosition = function(opt_options) {

  var options = opt_options ? opt_options : {};

  ...

  ol.control.Control.call(this, {
    element: element,
    render: render,
    target: options.target
  });
  
[...]

PR #7286, opened 2017-09-23
(WIP) remove thisArgs / opt_this

simonseyock:
See: #7284

### DIFF
--- a/src/ol/renderer/canvas/vectortilelayer.js
+++ b/src/ol/renderer/canvas/vectortilelayer.js
@@ -237,7 +237,8 @@ ol.renderer.canvas.VectorTileLayer.prototype.drawTileImage = function(
 /**
  * @inheritDoc
  */
-ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = function(coordinate, frameState, hitTolerance, callback, thisArg) {
+ol.renderer.canvas.VectorTileLayer.prototype.forEachFeatureAtCoordinate = function(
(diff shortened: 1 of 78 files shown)


src/ol/renderer/canvas/vectortilelayer.js
```diff
@@ -413,12 +412,12 @@ ol.renderer.canvas.VectorTileLayer.prototype.renderFeature = function(feature, s
...
       loading = ol.renderer.vector.renderFeature(
           replayGroup, feature, styles[i], squaredTolerance,
-          this.handleStyleImageChange_, this) || loading;
+          this.handleStyleImageChange_.bind(this)) || loading;
     }
   } else {
     loading = ol.renderer.vector.renderFeature(
         replayGroup, feature, styles, squaredTolerance,
-        this.handleStyleImageChange_, this) || loading;
+        this.handleStyleImageChange_.bind(this)) || loading;
```
simonseyock: This has the problem that a closure is created every time it is called. Alternatively it is possible to store a bound version of the method and use that one.

tschaub:
@simonseyock - I think we should spilt this work into two parts.  The first would remove the use of this arg from API functions.  The second would remove any unnecessary use of this arg internally.  We will soon be preparing for the next major release, transforming all sources to ES modules.  We can start again on the work you have done here after making the change to modules.

So I'll close this for now, but appreciate your work on it and welcome your help getting it going again after the modules transform is applied.

simonseyock:
Sounds good, I will continue my work on this then.

PR #7305, opened 2017-10-03
Fix #7304: Re-calculate the resolution when the WMS source is reprojected

### LINKED ISSUE #7304: getFeatureInfoUrl doesn't work for reprojected raster
When we use WMS with the on-the-fly reprojection the getFeatureInfoUrl needs the equivalent resolution and the coordinates requested in  the original projection, not those from the current view.

Since `ol.reproj.calculateSourceResolution`is not exportable, I see 2 solutions:

 1) Make `ol.reproj.calculateSourceResolution` exportable
or
 2) Automatically recalculate the resolution and the coordinate requested in the `getFeatureInfoUrl` function

What do you think?
The 2nd solution seems more clean.

oterral:
Fix #7304

### DIFF
--- a/src/ol/source/imagewms.js
+++ b/src/ol/source/imagewms.js
@@ -10,6 +10,7 @@ goog.require('ol.events.EventType');
 goog.require('ol.extent');
 goog.require('ol.obj');
 goog.require('ol.proj');
+goog.require('ol.reproj');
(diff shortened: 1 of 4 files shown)


oterral:
Thx for the review.

PR #7315, opened 2017-10-06
Add new ol.format.filter.Contains spatial operator

fredj:


fredj:
Thanks for the review @bartvde

PR #7323, opened 2017-10-06
Only clear the canvas when needed

### LINKED ISSUE #7322: Tile rendering regression
The 4.4 release introduces a rendering regression (most likely due to #7267) when calling `source.setUrl()` on a tile source.

Compare the reusable source example in the [4.4.0](http://openlayers.org/en/v4.4.0/examples/reusable-source.html) release to the behavior in [4.3.4](http://openlayers.org/en/v4.3.4/examples/reusable-source.html).

The [current rendering](http://openlayers.org/en/v4.4.0/examples/reusable-source.html) includes an unwanted flash of no tiles before the new tiles are rendered:

bad

The [old rendering](http://openlayers.org/en/v4.3.4/examples/reusable-source.html) showed no flash of white before rendering new tiles:

good

tschaub:
Instead of always clearing the canvas before rendering new tiles, we should be able to only clear when the extent has changed.  After tile transitions have completed, we already [clear before rendering](https://github.com/tschaub/openlayers/blob/ad4a258c87574f76397ed719cf4ac4a529ce7632/src/ol/renderer/canvas/tilelayer.js#L318-L320) at alpha 1, so this should be enough for non-opaque layers.

Fixes #7322.

### DIFF
--- a/src/ol/renderer/canvas/tilelayer.js
+++ b/src/ol/renderer/canvas/tilelayer.js
@@ -228,7 +228,9 @@ ol.renderer.canvas.TileLayer.prototype.prepareFrame = function(frameState, layer
         canvas.width = width;
         canvas.height = height;
       } else {
-        context.clearRect(0, 0, width, height);
+        if (this.renderedExtent_ && !ol.extent.equals(imageExtent, this.renderedExtent_)) {
+          context.clearRect(0, 0, width, height);
+        }
         oversampling = this.oversampling_;
       }
     }


tschaub:
Thanks @bartvde.

waxenegger:
@tschaub being one of the rare cases who happens to want the cache and canvas cleared at times/on demand (on a baselayer level) is there still an "elegant" programmatic way left to do so. by "elegant" I guess I mean outside of alpha changing and getting the html element of the canvas and clearing manually. Either cache.clear() or source.setTileUrlFunction() used to do the job in the past.

tschaub:
@waxenegger - calling `source.refresh()` clears the cache (for tile sources) and triggers re-rendering.  If that doesn't cover your use case, please open an issue with more detail about what you are trying to accomplish.

waxenegger:
thanks for pointing out that method. totally missed that there was already a handy function like that.

zforler:
Why can't I download it,the download page cannot be accessed

PR #7328, opened 2017-10-08
Declutter text and images

ahocevar:
This pull request adds a new `declutter` option for `ol.layer.Vector` and `ol.layer.VectorTile`, which will result in text and images no longer overlapping. Labels and images with a higher z-index are given precedence over those with a lower z-index.

nola


src/ol/layer/vectortile.js
```diff
@@ -34,17 +34,23 @@ ol.layer.VectorTile = function(opt_options) {
...
+  var renderMode = options.renderMode;
+
+  ol.asserts.assert(renderMode == undefined ||
+      renderMode == ol.layer.VectorTileRenderType.IMAGE ||
+      renderMode == ol.layer.VectorTileRenderType.HYBRID ||
+      renderMode == ol.layer.VectorTileRenderType.VECTOR,
   28); // `renderMode` must be `'image'`, `'hybrid'` or `'vector'`
 
+  if (options.declutter && renderMode == ol.layer.VectorTileRenderType.IMAGE) {
+    renderMode = ol.layer.VectorTileRenderType.HYBRID;
```
marcjansen: Can we somehow notify users that we change the rendering mode?

externs/olx.js
```diff
@@ -4310,6 +4311,14 @@ olx.layer.VectorOptions.prototype.source;
 
 
 /**
+ * Declutter images and text. Default is `false`.
```
marcjansen: I am leaning towards `true` as default. Why would one want a cluttered map by default?

src/ol/render/canvas/replay.js
```diff
@@ -360,7 +377,30 @@ ol.render.canvas.Replay.prototype.fill_ = function(context, rotation) {
...
+      if (!this.declutterTree.collides(this.declutterTree.toBBox(declutterGroup))) {
+        this.declutterTree.insert(declutterGroup.slice(0, 4));
+        var drawImage = ol.render.canvas.drawImage;
+        for (var j = 5, jj = declutterGroup.length; j < jj; ++j) {
+          if (declutterGroup[j]) {
+            drawImage.apply(undefined, /** @type {Array} */ (declutterGroup[j]));
+          }
+        }
+      }
+      declutterGroup.length = 5;
```
marcjansen: A lot of magic numbers, but I guess that is just how renderers / replays are right now.

src/ol/render/canvas/replaygroup.js
```diff
@@ -326,33 +426,40 @@ ol.render.canvas.ReplayGroup.prototype.isEmpty = function() {
...
   // visible outside the current extent when panning
-  var flatClipCoords = this.getClipCoords(transform);
   context.save();
-  context.beginPath();
-  context.moveTo(flatClipCoords[0], flatClipCoords[1]);
-  context.lineTo(flatClipCoords[2], flatClipCoords[3]);
-  context.lineTo(flatClipCoords[4], flatClipCoords[5]);
-  context.lineTo(flatClipCoords[6], flatClipCoords[7]);
-  context.clip();
+  this.clip(context, transform);
```
marcjansen: Nice change.

src/ol/layer/vectortile.js
ahocevar: It's documented [here](https://github.com/openlayers/openlayers/pull/7328/files#diff-044fc1d937cdd730e52ddad85740de2fR4412). Were you thinking of a more prominent place? I don't think anyone has ever used the`'image'` mode.

externs/olx.js
ahocevar: Not sure. Note that decluttering not only affects labels, but also point symbols. What do others think?

src/ol/render/canvas/replay.js
ahocevar: The structure is documented [here](https://github.com/openlayers/openlayers/pull/7328/files#diff-d6310f7e908b5b923bb2bc0abf9d7db2R43). I agree that these flat arrays are a bit hard to debug, but some believe they perform better than other structures.

src/ol/layer/vectortile.js
marcjansen: I was thinking about logging to the console, but as we don't do that anywhere else the documentation you linked to is enough.

externs/olx.js
marcjansen: Ah OK, I didn't get that first. Considering that also point symbols are affected, I think `false` is a sane default.

ahocevar: I'll try to change the default to `true` and check all the examples. If they all still look as expected (but uncluttered), then `true` it shall be 😄

marcjansen: But then we also need a big note in the readme, don't you think? As I said, I changed my mind because of the point symbols that are affected… Would be nice to hear what other think.

weskamm:
I really like this PR, very useful feature!
But i realized that somehow using the compiled ol-debug.js performs way better than the ol.js after doing a `make build` of this PR.
Any clues where this may come from?

ahocevar:
> But i realized that somehow using the compiled ol-debug.js performs way better than the ol.js after doing a make build of this PR.

Interesting. Maybe Closure Compiler tries to make an optimization that actually is none? I'd recommend using the https://npmjs.com/package/ol package anyway.

ahocevar:
@weskamm Can you provide instructions for how to see that difference between ol.js and ol-debug.js? I'd like to do some profiling to see where that difference might come from.

ahocevar:
This is now ready. Thanks for any review.

examples/vector-label-decluttering.html
```diff
@@ -7,7 +7,7 @@
   - https://unpkg.com/rbush@2.0.1/rbush.min.js
   - https://unpkg.com/labelgun@0.1.1/lib/labelgun.min.js
```
fredj: this can be removed (`labelgun.min.js`) and the tags updated

weskamm:
@ahocevar I prepared two examples to see the difference between ol and ol-debug.
They use the same code except the openlayers source.
Performance drops heavily with ol.js when zooming one step in

http://rawgit.com/weskamm/ol3vectortiles/ol4/indexlabels-debug.html
http://rawgit.com/weskamm/ol3vectortiles/ol4/indexlabels.html

ahocevar:
Thanks for the examples @weskamm. Looks like no decluttering is taking place with `ol.js`. Investigating now.

weskamm:
Not really sure but i think non functional decluttering in `ol.js` appeared after the latest commits

ahocevar:
@weskamm If you could lend a hand finding the culprit, it would be much appreciated. Even if it's just a `git bisect` result.

weskamm:
well, seems i was wrong, decluttering seems to be non functional in any of the commits in this PR for `ol.js`

ahocevar:
@weskamm I found the culprit - Closure compiler kills RBush because minX, maxX, minY and maxY get renamed. I'll see if I can make it work without relying on the ol.structs.RBush wrapper, which adds unnecessary overhead.

ahocevar:
@weskamm I added a commit which fixed the missing decluttering in ol.js. Can you please test if performance is now restored?

weskamm:
@ahocevar Issue seems to be fixed, declutter now works and performance is back again in ol.js.
I have updated the examples to use the latest build if you want to investigate anything further :+1:

ahocevar:
There are some failing tests after the last change. Need to investigate, back to WIP.

ahocevar:
All fixed, and ready for review.

examples/vector-label-decluttering.js
```diff
@@ -38,30 +17,35 @@ var map = new ol.Map({
...
+          widest = width;
+          geometry = polygon;
+        }
+      }
+    }
+    return geometry;
+  },
+  text: new ol.style.Text({
+    font: '12px Calibri,sans-serif',
+    exceedLength: true,
```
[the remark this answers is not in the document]
bartvde: ah I see this option is not added by this PR but was already there

src/ol/render/canvas/replay.js
```diff
@@ -24,12 +25,24 @@ goog.require('ol.transform');
  * @param {number} resolution Resolution.
  * @param {number} pixelRatio Pixel ratio.
  * @param {boolean} overlaps The replay can have overlapping geometries.
+ * @param {?} declutterTree Declutter tree.
```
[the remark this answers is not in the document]
ahocevar: It's because `ol.ext.RBush` is a component that the compiler does not know about.

src/ol/render/canvas/linestringreplay.js
```diff
@@ -17,11 +17,13 @@ goog.require('ol.render.canvas.Replay');
  * @param {number} resolution Resolution.
  * @param {number} pixelRatio Pixel ratio.
  * @param {boolean} overlaps The replay can have overlapping geometries.
+ * @param {?} declutter Declutter tree.
```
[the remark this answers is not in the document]
ahocevar: Good point. Will change.

examples/vector-label-decluttering.js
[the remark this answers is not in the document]
ahocevar: Yeah it was already added for text along path rendering. But if you have a better name suggestion, we could change it in 5.0.

src/ol/render/feature.js
```diff
@@ -102,6 +118,66 @@ ol.render.Feature.prototype.getExtent = function() {
...
+  }
+  return this.flatInteriorPoints_;
+};
+
+
+/**
+ * @return {Array.<number>} Flat interior points.
+ */
+ol.render.Feature.prototype.getFlatInteriorPoints = function() {
+  if (!this.flatInteriorPoints_) {
```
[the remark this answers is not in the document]
ahocevar: `ol.render.Feature` does not really distinguish between geometry types. It has the renderer relevant parts of the API of `ol.Feature` and all geometries, but only works with `flatCoordinates` and `ends` internally. Depending on whether the underlying geometry is a Polygon or a MultiPolygon, the renderer will only call `getFlatInteriorPoint` or `getFlatInteriorPoints`.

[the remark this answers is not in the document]
bartvde: Ah ok, thanks for explaining, then it makes sense

examples/vector-label-decluttering.js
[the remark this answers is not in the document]
bartvde: Ah reading up on it, it's not so bad. ```allowExceedLength``` would be more descriptive IMHO

ahocevar:

examples/vector-label-decluttering.js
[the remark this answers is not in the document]
tschaub: Or `overflow`.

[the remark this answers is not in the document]
bartvde: Even better. Leave it up to @tschaub to think of good names ;-)

PR #7329, opened 2017-10-08
Reworked attribution handling

### LINKED ISSUE #6237: ol.control.Attribution -> intersection test and uniqueAttributions only considering sources of type ol.source.TileImage
Why are only sources of the type ol.source.TileImage considered in these tests?
In my opinion, it doesn't make sense if I have a TileWMS with attributions "Swisstopo" and an ImageWMS with attributions "Swisstopo", that the ol.control.Attribution shows the attribution "Swisstopo" twice.
In addition, if I have layers with sources of type ol.source.Vector or ol.source.ImageWMS only covering Zurich, I do not expect these attributions showing up when the current map extent is only covering Bern for example.

tschaub:
This adds support for attribution functions and removes the need for `ol.Attribution`.

The `attributions` option for sources can now be a function that returns a single string attribution or an array of strings.
```js
var source = new ol.source.XYZ({
  attributions: function(frameState) {
    // inspect the frame state and return attributions
    return 'some attribution'; // or ['multiple', 'attributions'] or null
  }
});
```

As before, the `attributions` option can also just be a string or an array of strings.  Providing a function supports conditionally showing attribution based on extent, resolution, etc.  And for backwards compatibility, the option can still be a `ol.Attribution` or an array of the same.  The attribution control deduplicates visible attributions and renders them in layer order.

Since `source.getAttributions()` was marked with `@api`, that method still returns an array of `ol.Attribution`.  That method returns `null` if a source is constructed with the new function type attribution.

In the next major release, we can replace the old `source.getAttributions()` method with the new `source.getAttributions2()` method.  In the meantime, `source.getAttributions2()` is not documented as part of the API.

This reworks the internal handling of attributions so that the renderers no longer deal with them.  Instead, the attribution control is in charge of rendering visible attributions.  So if you don't have an attribution control, no unnecessary work is done.  (Next up, logos.)

Fixes #6237.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,36 @@
 
 ### Next Release
 
+#### Deprecation of `ol.Attribution`
+
+`ol.Attribution` is deprecated and will be removed in the next major version.  Instead, you can construct a source with a string attribution or an array of strings.  For dynamic attributions, you can provide a function that gets called with the current frame state.
+
+Before:
+```js
+var source = new ol.source.XYZ({
+  attributions: [
+    new ol.Attribution({html: 'some attribution'})
+  ]
+});
+```
+
+After:
+```js
+var source = new ol.source.XYZ({
+  attributions: 'some attribution'
+});
+```
+
+In addition to passing a string or an array of strings for the `attributions` option, you can also pass a function that will get called with the current frame state.
+```js
+var source = new ol.source.XYZ({
+  attributions: function(frameState) {
+    // inspect the frame state and return attributions
+    return 'some attribution'; // or ['multiple', 'attributions'] or null
+  }
+});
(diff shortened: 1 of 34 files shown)


cs09g:
when source doesn't have an attribution and tries to add custom logo using:
```
  ol.Map({
    ...
    controls: [ol.control.Attribution()],
    logo: {
      href: "link",
      src: "image url"
    }
```
then logo doesn't appear at all. I'm not sure I was using it wrong from the first because `attributions` is not required parameter for `source`. should `attributions` be needed for source? not for map?
note: 
https://github.com/openlayers/openlayers/blob/master/src/ol/control/attribution.js#L176
in above line of code, with 4.2.0, attributions = [{}, {}] but with 4.5.0, attributions = null

PR #7337, opened 2017-10-09
Always create a new blank image to avoid CSP violations

### LINKED ISSUE #7336: 4.4.0 Content Security Policy issue with generated inline png content
4.4.0 (and 4.4.1) cause the browser to report a Content Security Policy violation for directive "img-src 'self'" (when CSP is enforced)

Likely introduced by PR https://github.com/openlayers/openlayers/pull/7136
and similar to an already solved issue https://github.com/openlayers/openlayers/issues/6934

ahocevar:
Fixes #7336.

### DIFF
--- a/src/ol/imagetile.js
+++ b/src/ol/imagetile.js
@@ -61,7 +61,7 @@ ol.inherits(ol.ImageTile, ol.Tile);
 ol.ImageTile.prototype.disposeInternal = function() {
   if (this.state == ol.TileState.LOADING) {
     this.unlistenImage_();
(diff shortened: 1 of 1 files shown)


mattlovett:
I think this change has caused a regression... if we load tiles from a source that fails, and use an error handler to call load() to retry the tile, then we end up with blank tiles on the map instead of the real data.

Here's a codepen with a blank map... https://codepen.io/anon/pen/aVLjPo?editors=1111

If you alter the code to use an earlier OL release then you see the complete map.

ahocevar:
@mattlovett Thans for reporting this. See #7480 for a fix.

PR #7339, opened 2017-10-09
Use correct text stroke on HiDPI devices

ahocevar:
This fixes a regression with text strokes. In the current release, stroke widths are rendered too small on HiDPI devices.


ahocevar:
Looks like this is not consistent across browsers. Safari does not scale the lineWidth, but other browsers to. Will investigate further and see if this needs to be reverted.

ahocevar:
Actually this is not a regression. This difference between Safari and other browsers was also present in previous releases. See #7345 for a follow-up.

Issue #7334, opened 2017-10-09
Text rendering problem with scale in ol >= 4.4.0


ogo-adp:
We are trying to upgrade our app from 4.3.4 to 4.4.1 and we encounter a rendering problem.

We use features with point geometries and a text style to draw moving objects on a map. We can scale them according to some rules (map zoom or other). By using special fonts (like fontawesome) we can draw complex shapes of any size and get pixel perfect and fast rendering at any scale (like a vector rendering).

With ol >= 4.4.0 we do not get pixel perfect rendering anymore but some sort of pixelated rendering. 

Please look at this repro : https://codepen.io/anon/pen/KXoQmv?editors=0010

On this repro we use scale, and this does not render properly on 4.4.1 but render perfectly on 4.3.4.

I did not find a way to get back to previous behavior.

ahocevar:
This is because text is now pre-rendered to images, for better performance. Have you tried changing the font size and stroke width instead of the scale? Or use a font size for the highest possible scale, and scale down instead of up?

ogo-adp:
Yes, I've tried this, but it does not give good results. Shape is blurry, and for a reason I don't understand yet, sometimes ol does not render the shape properly (bad font used) in 4.4 (will dig further, this is probably a second problem).

About the performance choice, are you sure that drawImage() on canvas is any faster than fillText() especially when doing scaling of image ?

ahocevar:
The issue with the wrong font used is maybe that you are using a web font which is not loaded yet when the images are pre-rendered. This was the same previously, but you probably never noticed because another fillText() call would use the font as soon as it's loaded.

Regarding performance: We did a lot of profiling, and rendering text with drawImage is definitely much faster than calling fillText() and strokeText() over and over. That's for the common use case of rendering text for labels of course. What you're doing would probably be better solved with an ol.style.Icon that uses an SVG image with the symbol you want to use.

ahocevar:
Looking at the new code in `ol/render/canvas/textreplay`, I see that it should be doable to apply the scale before pre-rendering instead of afterwards. If you are willing to create a pull request, that would be welcome.

ogo-adp:
[...]

PR #7340, opened 2017-10-09
Clear label cache when fonts become available

ahocevar:
With this pull request, web fonts will be used when they become available, just like in OpenLayers < v4.4.0 when fillText/strokeText was used. This fixes the problem mentioned in https://github.com/openlayers/openlayers/issues/7334#issuecomment-335154208.

### DIFF
--- a/examples/vector-labels.js
+++ b/examples/vector-labels.js
@@ -11,6 +11,7 @@ goog.require('ol.style.Stroke');
 goog.require('ol.style.Style');
 goog.require('ol.style.Text');
 
+var openSansAdded = false;
 
 var myDom = {
   points: {
@@ -96,6 +97,13 @@ var createTextStyle = function(feature, resolution, dom) {
   var maxAngle = dom.maxangle ? parseFloat(dom.maxangle.value) : undefined;
   var exceedLength = dom.exceedlength ? (dom.exceedlength.value == 'true') : undefined;
   var rotation = parseFloat(dom.rotation.value);
+  if (dom.font.value == 'Open Sans' && !openSansAdded) {
+    var openSans = document.createElement('link');
+    openSans.href = 'https://fonts.googleapis.com/css?family=Open+Sans';
+    openSans.rel = 'stylesheet';
+    document.getElementsByTagName('head')[0].appendChild(openSans);
+    openSansAdded = true;
+  }
   var font = weight + ' ' + size + ' ' + dom.font.value;
   var fillColor = dom.color.value;
   var outlineColor = dom.outline.value;
--- a/examples/vector-labels.js
+++ b/examples/vector-labels.js
@@ -97,7 +97,7 @@ var createTextStyle = function(feature, resolution, dom) {
   var maxAngle = dom.maxangle ? parseFloat(dom.maxangle.value) : undefined;
   var exceedLength = dom.exceedlength ? (dom.exceedlength.value == 'true') : undefined;
   var rotation = parseFloat(dom.rotation.value);
-  if (dom.font.value == 'Open Sans' && !openSansAdded) {
+  if (dom.font.value == '\'Open Sans\'' && !openSansAdded) {
     var openSans = document.createElement('link');
     openSans.href = 'https://fonts.googleapis.com/css?family=Open+Sans';
     openSans.rel = 'stylesheet';
--- a/src/ol/css.js
+++ b/src/ol/css.js
@@ -49,7 +49,7 @@ ol.css.CLASS_CONTROL = 'ol-control';
  * Get the list of font families from a font spec.  Note that this doesn't work
  * for font families that have commas in them.
  * @param {string} The CSS font property.
- * @return {Array.<string>} The font families (or null if the input spec is invalid).
+ * @return {Object.<string>} The font families (or null if the input spec is invalid).
  */
 ol.css.getFontFamilies = (function() {
   var style;
--- a/src/ol/css.js
+++ b/src/ol/css.js
@@ -43,3 +43,30 @@ ol.css.CLASS_UNSUPPORTED = 'ol-unsupported';
  * @type {string}
  */
 ol.css.CLASS_CONTROL = 'ol-control';
+
+
+/**
+ * Get the list of font families from a font spec.  Note that this doesn't work
+ * for font families that have commas in them.
+ * @param {string} The CSS font property.
+ * @return {Array.<string>} The font families (or null if the input spec is invalid).
+ */
+ol.css.getFontFamilies = (function() {
+  var style;
+  var cache = {};
+  return function(font) {
+    if (!style) {
+      style = document.createElement('div').style;
+    }
+    if (!(font in cache)) {
+      style.font = font;
+      var family = style.fontFamily;
+      style.font = '';
+      if (!family) {
+        return null;
+      }
+      cache[font] = family.split(/,\s?/);
+    }
+    return cache[font];
+  };
+})();
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -108,7 +108,7 @@ ol.render.canvas.checkFont = (function() {
 
   function isAvailable(fontFamily) {
     if (!context) {
-      context = ol.dom.createCanvasContext2D();
+      context = ol.dom.createCanvasContext2D(1, 1);
       context.font = '32px monospace';
       referenceWidth = context.measureText(text).width;
     }
@@ -120,6 +120,10 @@ ol.render.canvas.checkFont = (function() {
       // fallback was used instead of the font we wanted, so the font is not
       // available.
       available = width != referenceWidth;
+      // Setting the font back to a different one works around an issue in
+      // Safari where subsequent `context.font` assignments with the same font
+      // will not re-attempt to use a font that is currently loading.
+      context.font = '32px monospace';
     }
     return available;
   }
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -1,6 +1,9 @@
 goog.provide('ol.render.canvas');
 
 
+goog.require('ol.css');
+goog.require('ol.dom');
+goog.require('ol.structs.LRUCache');
 goog.require('ol.transform');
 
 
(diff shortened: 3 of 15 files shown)


src/ol/render/canvas/textreplay.js
```diff
@@ -304,7 +304,13 @@ ol.render.canvas.TextReplay.prototype.getImage_ = function(text, fill, stroke) {
...
   var lines = text.split('\n');
   var numLines = lines.length;
-  if (!ol.render.canvas.TextReplay.labelCache_.containsKey(key)) {
+  var hasLabel = ol.render.canvas.TextReplay.labelCache_.containsKey(key);
+  var checkLabel = hasLabel && !this.labelsChecked_;
+  if (!hasLabel || checkLabel) {
+    if (checkLabel) {
+      label = ol.render.canvas.TextReplay.labelCache_.get(key);
+      this.labelsChecked_ = true;
+    }
```
tschaub: This looks like we only check once.  I'm confused how this works if the font loads after this check.

ahocevar: We check when a new replay is created, which is when a new tile is loaded or a new resolution for an existing tile is used.

Do you think we should run the check once per render frame instead?

tschaub: Curious if it would be possible to use document.fonts.check() to avoid the check and ready() to render after load.  https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API

ahocevar: Good hint. I'll check that.

ahocevar: Unfortunately that `document.fonts` API is unavailable in Internet Explorer and Edge, and gives false positives in Safari. But I rewrote the whole pull request to use a simple custom check for available fonts, and to re-render as soon as a font is loaded. This should now even work better than in previous versions, where at least an interaction or animation was required to update the view when fonts became available.

ahocevar:
New version of this pull request pushed, and ready for review.

src/ol/render/canvas/textreplay.js
```diff
@@ -122,6 +122,61 @@ ol.render.canvas.TextReplay.labelCache_ = new ol.structs.LRUCache();
...
+  var text = 'wmytzilWMYTZIL@#/&?$%10';
+  var context;
+
+  function isLoading(fontSpec) {
+    if (!context) {
+      context = ol.dom.createCanvasContext2D();
+    }
+    var loading = false;
+    //FIXME fontFamily detection fails for a font name with '[0-9]px ' in it.
+    var fontFamily = fontSpec.split(/[0-9]px /).pop();
```
tschaub: Curious if we can find a more robust way to get the name.  The `font-size` units could also be `em` or `%`.  And could also be a relative (`smaller`) or absolute (`small`) size.

tschaub: We could use the browser's css parser.  E.g.:
```js
var element = document.createElement('div');
element.style.font = fontSpec;
var fontFamily = element.style.fontFamily;
```

tschaub: Maybe something like this:
```js
/**
 * Get the font family from a font spec.
 * @param {string} The CSS font property.
 * @return {string} The font family.
 */
ol.css.getFontFamily = (function() {
  var style = document.createElement('div').style;
  var cache = {};
  return function(fontSpec) {
    if (!(fontSpec in cache)) {
      style.font = fontSpec;
      cache[fontSpec] = style.fontFamily;
    }
    return cache[fontSpec];
  };
})();
```

examples/vector-labels.js
```diff
@@ -96,6 +97,13 @@ var createTextStyle = function(feature, resolution, dom) {
...
   var exceedLength = dom.exceedlength ? (dom.exceedlength.value == 'true') : undefined;
   var rotation = parseFloat(dom.rotation.value);
+  if (dom.font.value == 'Open Sans' && !openSansAdded) {
+    var openSans = document.createElement('link');
+    openSans.href = 'https://fonts.googleapis.com/css?family=Open+Sans';
+    openSans.rel = 'stylesheet';
+    document.getElementsByTagName('head')[0].appendChild(openSans);
+    openSansAdded = true;
+  }
   var font = weight + ' ' + size + ' ' + dom.font.value;
```
tschaub: You didn't introduce this problem, but the font family should be quoted here.  As this is written, we end up with things like `bold 10px Courier New` where it should be `bold 12px "Courier New"`.

tschaub:
In the vector labels example, I'm seeing `isLoading()` return true all the time.  Not clear to me that it is working as intended yet.

examples/vector-labels.js
tschaub: Feels like the `option` values should be quoted (so if someone added `serif` as an option, we wouldn't accidentally quote it when concatenating here).

So something like this should be enough:
```diff
diff --git a/examples/vector-labels.html b/examples/vector-labels.html
index 67b00a402..69c6acf50 100644
--- a/examples/vector-labels.html
+++ b/examples/vector-labels.html
@@ -70,8 +70,8 @@ tags: "geojson, vector, openstreetmap, label"
     <label>Font: </label>
     <select id="points-font">
       <option value="Arial" selected="selected">Arial</option>
-      <option value="Courier New">Courier New</option>
-      <option value="Open Sans">Open Sans</option>
+      <option value="'Courier New'">Courier New</option>
+      <option value="'Open Sans'">Open Sans</option>
       <option value="Verdana">Verdana</option>
     </select>
     <br />
@@ -179,8 +179,8 @@ tags: "geojson, vector, openstreetmap, label"
     <label>Font: </label>
     <select id="lines-font">
       <option value="Arial">Arial</option>
-      <option value="Courier New" selected="selected">Courier New</option>
-      <option value="Open Sans">Open Sans</option>
+      <option value="'Courier New'" selected="selected">Courier New</option>
+      <option value="'Open Sans'">Open Sans</option>
       <option value="Verdana">Verdana</option>
     </select>
     <br />
@@ -288,8 +288,8 @@ tags: "geojson, vector, openstreetmap, label"
     <label>Font: </label>
     <select id="polygons-font">
       <option value="Arial">Arial</option>
-      <option value="Courier New">Courier New</option>
-      <option value="Open Sans">Open Sans</option>
+      <option value="'Courier New'">Courier New</option>
+      <option value="'Open Sans'">Open Sans</option>
       <option value="Verdana" selected="selected">Verdana</option>
     </select>
     <br />
```

tschaub: I added c2460dd8d8db2ac1907949459cf93591fc920d28 to address this.

src/ol/render/canvas/textreplay.js
tschaub: I added `ol.css.getFontFamilies()` in e54f8c9dbe5100de62e033e4a9203b41c7498312.

tschaub:
I'm uncertain about the return on investment here.  There are a number of cases to consider.  

Below, "a label cannot be properly rendered with a font" means that none of the listed font families are available or are missing a needed glyph for one of the characters in the label.  When this happens, specifying another fallback font can change the rendering.

Consider rendering a single label.

  1. User specifies `font: 12px FontA;`
      i. The label can be properly rendered.
      ii. The label cannot be properly rendered.
          a. FontA is not available and will never be loaded.
          b. FontA is loading.
              a. When FontA loads, the label can be properly rendered.
              b. When FontA loads, the label still cannot be properly rendered.
          c. FontA will be loaded at some point in the future.  Same two sub-points as above apply here.
  2. User specifies `font: 12px FontA, FontB;`
      i. The label can be rendered using FontB as a fallback.
          a. FontA is not available and will never be loaded.
          b. FontA is loading.
              a. When FontA loads, the label can be rendered without falling back to FontB.
              b. When FontA loads, FontB is still used as a fallback.
          c. FontA will be loaded at some point in the future.  Same two sub-points above apply here.
      ii. The label cannot be properly rendered with FontA or FontB.
          a. FontA and FontB are not available and will never be loaded.
          b. FontA or FontB are loading.
              a. When fonts load, the label can be properly rendered.
              b. When fonts load, the label still cannot be properly rendered.
          c. FontA or FontB will be loaded in the future.  Same two sub-points above apply here.

The trick of adding a fallback and checking if rendering width changes doesn't work for `2.i` above.  And I think this would be pretty common.  If a user is loading fonts with `@font-face`, I imagine they will specify a fallback to be used before the preferred font loads.  If we decide whether or not we can cache rendered labels by appending another fallback, we will incorrectly cache a labels.

There are other cases above where the "cache only if adding a fallback doesn't change rendering" strategy is not so great.  In `1.ii.a`, for example, we do a lot of extra work.

It feels like the only reliable way to deal with cached labels is to invalidate the cache any time a font loads.  So, where available we could:

```js
document.fonts.addEventListener('loadingdone', invalidateLabelCache);
```

Of course this [doesn't work for IE or Edge](https://caniuse.com/#feat=font-loading).  One option would be to only cache if `document.fonts` is available.

ahocevar:
@tschaub I reworked this pull request again, and took advantage of the font family parser that you added. The pull request now does what you suggested: it invalidates the label cache when a previously unavailable font family becomes available, but does not rely on `document.fonts` for checking. In addition to that, it makes it so layer renderers listen to `clear` events on the label cache, and redraw to reflect changes immediately.

src/ol/css.js
```diff
@@ -43,3 +43,30 @@ ol.css.CLASS_UNSUPPORTED = 'ol-unsupported';
  * @type {string}
  */
 ol.css.CLASS_CONTROL = 'ol-control';
+
+
+/**
+ * Get the list of font families from a font spec.  Note that this doesn't work
+ * for font families that have commas in them.
+ * @param {string} The CSS font property.
+ * @return {Array.<string>} The font families (or null if the input spec is invalid).
```
bartvde: isn't it ```Object.<string>``` ?

src/ol/render/canvas.js
```diff
@@ -79,6 +84,87 @@ ol.render.canvas.defaultLineWidth = 1;
...
+      context.font = '32px ' + fontFamily + ',monospace';
+      var width = context.measureText(text).width;
+      // If width and referenceWidth are the same, then the 'monospace'
+      // fallback was used instead of the font we wanted, so the font is not
+      // available.
+      available = width != referenceWidth;
+      // Setting the font back to a different one works around an issue in
+      // Safari where subsequent `context.font` assignments with the same font
+      // will not re-attempt to use a font that is currently loading.
+      context.font = '32px monospace';
```
bartvde: minor, but we can put this into a common variable since it's used more than once

src/ol/render/canvas.js
```diff
@@ -79,6 +84,87 @@ ol.render.canvas.defaultLineWidth = 1;
...
+          checked[font] = 60;
+          labelCache.clear();
+        } else {
+          ++checked[font];
+          done = false;
+        }
+      }
+    }
+    if (!done) {
+      window.setTimeout(check, 25);
```
bartvde: should we clear the timer somewhere?

test/spec/ol/render/canvas/index.test.js
```diff
@@ -1,10 +1,72 @@
...
-
+goog.require('ol.events');
+goog.require('ol.obj');
 goog.require('ol.render.canvas');
 
 
 describe('ol.render.canvas', function() {
 
+  var font = document.createElement('link');
+  font.href = 'https://fonts.googleapis.com/css?family=Inconsolata';
```
bartvde: this will make it impossible to run our tests offline but I assume that is already the case (haven't checked)?

ahocevar: Yes, our tests already require an internet connection.

PR #7346, opened 2017-10-11
Pre-render text images for configured scale

### LINKED ISSUE #7334: Text rendering problem with scale in ol >= 4.4.0

ahocevar:
Fixes #7334.


src/ol/render/canvas/textreplay.js
```diff
@@ -310,6 +310,7 @@ ol.render.canvas.TextReplay.prototype.getImage_ = function(text, fill, stroke) {
     var fillState = this.textFillState_;
     var textState = this.textState_;
     var pixelRatio = this.pixelRatio;
+    var scale = this.textScale_;
```
fredj: minor: a `pixelRatio * scale` variable could be created to avoid to compute the value multiple times

Issue #7054, opened 2017-07-28
feature request: interaction.extent pixel tolerance


marcosox:
It would be nice to have the pixel tolerance for the `ol.interaction.Extent` interaction configurable with an option passed to the constructor, in the same way as the `pixelTolerance` option parameter works for `ol.interaction.Modify`.

marcosox:
PR #7355 adds this feature

marcosox:
The pull request has been accepted and merged, so this can be closed.

PR #7355, opened 2017-10-13
Pass pixel tolerance as a parameter to constructor of ol.interaction.Extent

marcosox:
This adds the pixelTolerance option parameter to the constructor of ol.interaction.Extent, as requested in issue #7054 
In this way the user can override the current default value of 10.

### DIFF
--- a/src/ol/interaction/extent.js
+++ b/src/ol/interaction/extent.js
@@ -31,7 +31,7 @@ goog.require('ol.style.Style');
  */
 ol.interaction.Extent = function(opt_options) {
 
-  var options = opt_options ? opt_options : {};
+  var options = opt_options || {};
 
   /**
    * Extent of the drawn box
--- a/src/ol/interaction/extent.js
+++ b/src/ol/interaction/extent.js
@@ -31,6 +31,8 @@ goog.require('ol.style.Style');
  */
 ol.interaction.Extent = function(opt_options) {
 
+  var options = opt_options ? opt_options : {};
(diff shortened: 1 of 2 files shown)


src/ol/interaction/extent.js
```diff
@@ -31,6 +31,8 @@ goog.require('ol.style.Style');
  */
 ol.interaction.Extent = function(opt_options) {
 
+  var options = opt_options ? opt_options : {};
```
lodenrogue: You can alternatively do:

`var options = opt_options || {};`

Looks cleaner in my opinion.

src/ol/interaction/extent.js
```diff
@@ -50,7 +52,8 @@ ol.interaction.Extent = function(opt_options) {
    * @type {number}
    * @private
    */
-  this.pixelTolerance_ = 10;
+  this.pixelTolerance_ = options.pixelTolerance !== undefined ?
+    options.pixelTolerance : 10;
```
lodenrogue: Same thing here.

this.pixelTolerance_ = options.pixelTolerance || 10;

marcosox:
@tschaub: I made the requested changes, hope everything is correct now.

@lodenrogue: thanks for the tip, I modified that line as well.

lodenrogue:
@marcosox You're welcome! Glad to help!

tschaub:
Thanks for the contribution @marcosox.

PR #7356, opened 2017-10-13
Fix documentation for target option in the controls

EduardoNogueira:
The ol.control.Control subclasses accept a string as target option, and was missing some documentation for this option.


tschaub:
Thanks for the fix @EduardoNogueira.

Issue #7303, opened 2017-10-03
ol.format.GeoJSON (readFeatures) reading geometry_name 


EduardoNogueira:
We are working with wfs layers that we don't know the geometry name beforehand and we cannot use the geometryName parameter from the ol.format.GeoJSON. The wfs layers are hosted on a geoserver that returns the geometry_name field in the geojson output format. We are wondering if that could be an option like `readGeometryNameFromFeature` that in the case of the geometryName option being unset looks for a geometry_name field on the geojson.
We are available to do a PR, if you interest in include this feature on openlayers.

Geoserver output format docs:
http://docs.geoserver.org/stable/en/user/services/wfs/outputformats.html#json-and-jsonp-output

ahocevar:
A GeoJSON document will always have the geometry as "geometry" property. This can also be seen in the snippet on the doc page you linked.

Why exactly do you want to use a different geometry name in the features that are read from the GeoJSON?

EduardoNogueira:
When we send wfs transactions to geoserver as the default geometry name is "geometry", if it doesn't correspond to the one declared in the feature model it will throw an error and because geometry_name isn't in the properties we can't call `setGeometryName` programmatically. By reading the geometry_name from the geojson this problem won't happen.

PR #7362, opened 2017-10-16
Added option to the ol.format.GeoJSON to allow the reading of the geometry_name from the geojson

Alexandre27:
This PR provides a solution to the issue #7303. In spite of the geometry_name field not being in the geojson specification, some mapping servers like geoserver add the geometry_name field to it. It can be useful that the geojson reader reads that field from the geojson, because if the loaded features are used to write WFS transactions to update that feature after some modifications and in that case if the geometry name doesn't match the one in the server it throws an error.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2029,7 +2029,8 @@ olx.format.GeoJSONOptions.prototype.geometryName;
  * Certain GeoJSON providers include the geometry_name field in the feature
  * geoJSON. If set to `true` the geoJSON reader will look for that field to
  * set the geometry name. If both this field is set to `true` and a
- * `geometryName` is provided the `geometryName` will take precedence.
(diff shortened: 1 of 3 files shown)


externs/olx.js
```diff
@@ -2025,6 +2026,17 @@ olx.format.GeoJSONOptions.prototype.geometryName;
 
 
 /**
+ * Certain GeoJSON providers include the geometry_name field in the feature
```
ahocevar: You should also document that the default is `false`.

Alexandre27: Thanks for the review. The documentation was updated.

ahocevar:
Thanks, this looks good now @Alexandre27. Can you please submit a [CLA](https://github.com/openlayers/cla) so we can merge this?

Alexandre27:
Thanks @ahocevar, CLA submitted.

PR #7365, opened 2017-10-17
Enable z-index for select interactions

bvx89:
When creating a select interaction, the layer that's created is not managed by the layercollection of the map.

This commit allows you to set a Z-index when creating the layer, so that the layers can be ordered.

This PR is based on the same issue as discussed in #4298

----

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -2941,7 +2941,8 @@ olx.interaction.PointerOptions.prototype.handleUpEvent;
  *     features: (ol.Collection.<ol.Feature>|undefined),
  *     filter: (ol.SelectFilterFunction|undefined),
  *     wrapX: (boolean|undefined),
- *     hitTolerance: (number|undefined)}}
+ *     hitTolerance: (number|undefined),
+ *     zIndex: (number|undefined)}}
  */
 olx.interaction.SelectOptions;
 
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -3067,6 +3067,14 @@ olx.interaction.SelectOptions.prototype.hitTolerance;
 
 
 /**
+ * The z-index for layer rendering. The default Z-index is 0.
+ * @type {number|undefined}
+ * @api
+ */
+olx.interaction.SelectOptions.prototype.zIndex;
+
+
+/**
  * Options for snap
  * @typedef {{
  *     features: (ol.Collection.<ol.Feature>|undefined),
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -96,7 +96,8 @@ var Select = function(opt_options) {
     style: options.style ? options.style :
       Select.getDefaultStyleFunction(),
     updateWhileAnimating: true,
-    updateWhileInteracting: true
+    updateWhileInteracting: true,
+    zIndex: goog.isDef(options.zIndex) ? options.zIndex : Infinity
   });
 
   /**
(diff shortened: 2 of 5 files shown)


src/ol/interaction/select.js
```diff
@@ -97,7 +97,8 @@ ol.interaction.Select = function(opt_options) {
     style: options.style ? options.style :
       ol.interaction.Select.getDefaultStyleFunction(),
     updateWhileAnimating: true,
-    updateWhileInteracting: true
+    updateWhileInteracting: true,
+    zIndex: goog.isDef(options.zIndex) ? options.zIndex : Infinity
```
tschaub: Please use `options.zIndex !== undefined` instead of `goog.isDef(options.zIndex)`.

tschaub: Please use `options.zIndex !== undefined` instead of `goog.isDef(options.zIndex)`.

src/ol/layer/layer.js
```diff
@@ -180,8 +180,9 @@ ol.layer.Layer.prototype.setMap = function(map) {
     this.mapPrecomposeKey_ = ol.events.listen(
         map, ol.render.EventType.PRECOMPOSE, function(evt) {
           var layerState = this.getLayerState();
+          var zIndex = this.getZIndex();
           layerState.managed = false;
-          layerState.zIndex = Infinity;
+          layerState.zIndex = (goog.isDef(zIndex) && zIndex !== 0) ? zIndex : Infinity;
```
tschaub: Curious why `0` wouldn't be a valid z-index here.

As above, please do not use `goog.isDef()`.

externs/olx.js
```diff
@@ -3538,7 +3538,8 @@ olx.interaction.PointerOptions.prototype.handleUpEvent;
  *     features: (ol.Collection.<ol.Feature>|undefined),
  *     filter: (ol.SelectFilterFunction|undefined),
  *     wrapX: (boolean|undefined),
- *     hitTolerance: (number|undefined)}}
+ *     hitTolerance: (number|undefined),
+ *     zIndex: (number|undefined)}}
  */
```
tschaub: You'll also need to add a block to document the `zIndex` option here.  See the [`hitTolerance` option](https://github.com/bvx89/openlayers/blob/c4fee4551a838f176fbd671c1c22d6fcb2f64f6d/externs/olx.js#L3656-L3663) for an example.

src/ol/layer/layer.js
bvx89: I based my code on the pull request #4298, and thought it had to be that way. Should I just allow 0 as an index then?

bvx89:
@tschaub Does the pull request look okay now?

aalbericio:
I believe this has something to do with the issue I'm having right now:

If I have different layers with different internals (Layer.Image, etc), as soon as I select a feature, it is brought on front.

But that "selection layer" does not work as the original layer containing the feature (Layer.Image in this case) but as a "default" Layer.Vector.

Selected features are displayed in a different way (specially when zooming) and they don't dissappear whenever the source Layer (Layer.Image) hides its features due to some min/max zoom/resolutions configuration.

ahocevar:
@aalbericio Your issue is not related to this pull request. I think for your use case you'd better not use the Select interaction, but a click handler which just changes the style of the clicked feature.

aalbericio:
@ahocevar Yes, that's what I have finally done but I get the feeling I'm reinventing the wheel...

Thanks anyway.

Stockholmsnovis:
Until this feature is ready, a possible solution could be to use a second layer in conjunction with the select interaction. The extra layer can use the same style and feature source as the select interaction. When the select interaction is disabled, the hidden layer can become visible. Since the hidden layer is a "normal" layer, z-index can be controlled.

In a way I think it kind of makes sense that select interactions should live outside (always on top) of all other layers, because otherwise one could easily get into situations where stuff doesn't seem to be selected because it is hidden underneath some other layer. However, it is nice to be able to control z-index of selected items if necessary.

Jinkwon:
In my case, resolve issue with this way.
```
const select: Select = new Select({
  condition: pointerMove,
  style: style,
  layers: [layer]
});
const l = select.getOverlay();
if (l) {
  l.setZIndex(99);
}
```

PR #7368, opened 2017-10-18
WFS: add srsName handling into WFS, and don't let GMLbase overwrite it

### LINKED ISSUE #6688: WFS GML3 does not have correct srsProjection
i have wfs/gml3 in the following form:
```xml
<wfs:FeatureCollection xmlns:gml="http://www.opengis.net/gml" xmlns:wfs="http://www.opengis.net/wfs" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd http://www.opengis.net/gml http://schemas.opengis.net/gml/3.1.1/base/gml.xsd">
    <gml:boundedBy>
        <gml:Envelope srsName="EPSG:4326">
            <gml:lowerCorner>47.141223013472548 5.5612518515690752</gml:lowerCorner>
            <gml:upperCorner>55.102916878371701 15.578555265395364</gml:upperCorner>
        </gml:Envelope>
    </gml:boundedBy>
    <gml:featureMember>
    </gml:featureMember>
</wfs:FeatureCollection>
```
with the srsName inside the Envolope in the head of wfs instead of the gml extra.

that might be fixed with extending ``ol.format.WFS.prototype.readProjectionFromNode`` with this:
```javascript
if (node.hasAttribute('srsName')) {
    return node.getAttribute('srsName');
}
```

==
currently inside a loader i do `` format.readFeatures(response);``

```javascript
new ol.source.Vector({
    format: new ol.format.WFS({
        gmlFormat: new ol.format.GML3()
    }),
    loader: function(extent, resolution, projection) {
        var src = this;
        //local file
	$.ajax('vg.gml').then(function(response) {
             // THIS function does not return the correct value
	     srcProjection = src.getFormat().readProjection(response);
             data = src.getFormat().readFeatures(response);

             
             // need to switch X and Y because for unknown reason the data is wrong
            for (var i=0;i<data.length;i++) {
                data[i].getGeometry().applyTransform(function (coords, coords2, stride) {
                    for (var j=0;j<coords.length;j+=stride) {
                         var y = coords[j];
                         var x = coords[j+1];
                         coords[j] = x;
                         coords[j+1] = y;
                     }
                });
                // needed to get from srcProjection to wanted projection
                data[i].getGeometry().transform(srcProjection, projection);
            }
		
            src.addFeatures(data);
        });
    }
})
```

it works correct for now but it does not when the features / the file is in a different projection then the map view projection.

i needed the swap of  the X and Y values, for some reason. (the service url does return gml2 and gml3 data, but i use gml3 one) openlayers3 has no option for

also i am not sure why the ``getGeometry().transform`` is needed.

I currently have v3.20.1 but the source of v4 shows it might not be fixed there.

might be related to #6235

Hanmac:
it fixed the srsName for gml inside wfs, for example `gml:Envelope`
that fixed that the objects is read with the correct srs and specially the axes rotation works

Fixed #6688

---

### DIFF
--- a/src/ol/format/gmlbase.js
+++ b/src/ol/format/gmlbase.js
@@ -197,14 +197,19 @@ ol.format.GMLBase.prototype.readFeaturesInternal = function(node, objectStack) {
  */
 ol.format.GMLBase.prototype.readGeometryElement = function(node, objectStack) {
   var context = /** @type {Object} */ (objectStack[0]);
-  context['srsName'] = node.firstElementChild.getAttribute('srsName');
-  context['srsDimension'] = node.firstElementChild.getAttribute('srsDimension');
+  var newContext = ol.obj.assign({}, context);
+  if (node.firstElementChild.hasAttribute('srsName')) {
+    newContext['srsName'] = node.firstElementChild.getAttribute('srsName');
+  }
+  if (node.firstElementChild.hasAttribute('srsDimension')) {
+    newContext['srsDimension'] = node.firstElementChild.getAttribute('srsDimension');
+  }
   /** @type {ol.geom.Geometry} */
   var geometry = ol.xml.pushParseAndPop(null,
       this.GEOMETRY_PARSERS_, node, objectStack, this);
   if (geometry) {
     return /** @type {ol.geom.Geometry} */ (
-      ol.format.Feature.transformWithOptions(geometry, false, context));
+      ol.format.Feature.transformWithOptions(geometry, false, newContext));
   } else {
     return undefined;
   }
--- a/src/ol/format/wfs.js
+++ b/src/ol/format/wfs.js
@@ -134,7 +134,8 @@ ol.format.WFS.prototype.readFeatures;
 ol.format.WFS.prototype.readFeaturesFromNode = function(node, opt_options) {
   var context = /** @type {ol.XmlNodeStackItem} */ ({
     'featureType': this.featureType_,
-    'featureNS': this.featureNS_
+    'featureNS': this.featureNS_,
+    'srsName': this.readProjectionFromNode(node)
   });
   ol.obj.assign(context, this.getReadOptions(node,
       opt_options ? opt_options : {}));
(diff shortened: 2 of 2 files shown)


Hanmac:
the travis test currently failed,

but for my `'EPSG:3857'` data,
the data is shown in the correct projection after the change.

probably the test data might be incorrect?

Hanmac:
@fredj: can someone check why the travis tests are failing?

my test data with 'EPSG:3857' did worked with that changes (they didn't worked before)

maybe the test data might be incorrect?

jessyst:
Unfortunately this did not solve my problem.   
In the function :
ol.format.GML3.prototype.readFlatPosList_ = function(node, objectStack)

When I try to parse a 3D-polygon (again Lines and Points work fine) it tries to get the dimension from te node provide:
var containerDimension = node.parentNode.getAttribute('srsDimension');


The node which is entered is:
```
 <gml:posList>
586411.04415116 4513276.93414081 -13.99350134 586398.36072733 4513266.49453056 -13.93448107 586415.54812812 4513256.95443897 -14.31830783 586411.04415116 4513276.93414081 -13.99350134
</gml:posList>
```


While the srsName and srsDimension id on the feature node: 
```
<gml:Polygon srsName="http://www.opengis.net/gml/srs/epsg.xml#26918" srsDimension="3">
<gml:exterior>
<gml:LinearRing>
<gml:posList>
586395.71101396 4513249.97706339 -14.09935104 586371.84299445 4513150.80882541 -11.96109535 586368.68838707 4513252.48427715 -5.76905809 586395.71101396 4513249.97706339 -14.09935104
</gml:posList>
</gml:LinearRing>
</gml:exterior>
</gml:Polygon>
```

As you can see  node.parentNode will get you gml:LinearRing and not gml:Polygon

Hanmac:
@jessyst i think the srsDimension should be get from the context,
and that should be set from the parent node(s) too
i did't test with your data but i need to try it

jessyst:
I agree, it should get it from the context, but with polygons the geometry is nested deeper

PR #7371, opened 2017-10-19
Add @api annotation to ol.VectorTile.getExtent

### LINKED ISSUE #7370: VectorTile.getExtent not available in minified version of Openlayers
According to the upgrade notes for version 4.4.0 when using a custom tile loader for tiles with MVT format the tiles extent should be passed to the reader:

```
tile.setLoader(function() {
  var data = // ... fetch data
  var format = tile.getFormat();
  tile.setFeatures(format.readFeatures(data, {
    featureProjection: map.getView().getProjection(),
    extent: tile.getExtent()
  }));
);
```

In the minified version of openlayers getExtent is not defined on the tile (because of the minification).

bobobobo:
This resolves #7370

### DIFF
--- a/src/ol/vectortile.js
+++ b/src/ol/vectortile.js
@@ -92,6 +92,7 @@ ol.VectorTile.prototype.disposeInternal = function() {
 /**
  * Gets the extent of the vector tile.
  * @return {ol.Extent} The extent.
+ * @api
  */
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for the contribution @bobobobo!

PR #7375, opened 2017-10-19
Read 'Abstract', 'AccessConstraints' and 'Fees' in ol.format.OWS

### LINKED ISSUE #7373: Read fees / access constraints (ol.format.WMTSCapabilities)
In our openlayers application the user has to be informed about fees / access constraints (e.g. print output) dynamically.
For WMS everthing works fine. For WMTS only _ServiceType_, _ServiceTypeVersion_ and _Title_ are received.

Any plans to extend wmts parser?

openlayers version: 4.4.2
demo url: https://jsfiddle.net/va83172c/
service url: https://www.wmts.nrw.de/geobasis/wmts_nw_dop20/1.0.0/WMTSCapabilities.xml

> We are aware that the current service capability isn't valid against ogc wmts schema (wrong attribute KeywordList). We are working on that issue.
> With valid capability document (on local webserver) the parser can't read fees / access constraints too.

fredj:
fixes #7373

### DIFF
--- a/src/ol/format/ows.js
+++ b/src/ol/format/ows.js
@@ -403,6 +403,9 @@ ol.format.OWS.SERVICE_CONTACT_PARSERS_ =
 ol.format.OWS.SERVICE_IDENTIFICATION_PARSERS_ =
     ol.xml.makeStructureNS(
         ol.format.OWS.NAMESPACE_URIS_, {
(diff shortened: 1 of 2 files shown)


fredj:
Thanks for the review

Issue #7241, opened 2017-09-11
No option to create unselectable overlays


localjo:
Currently, [all overlays get the class `ol-selectable`](https://github.com/openlayers/openlayers/blob/master/src/ol/overlay.js#L62), regardless of their purpose. In our app, none of the overlays we're using should be selectable by the user, and I imagine that this is a common use case. Writing CSS to override the existing styles and make `.ol-selectable` unselectable is both counter intuitive and makes it difficult to have some overlays which are selectable, while others are not. I propose adding an option to give overlays the `ol-unselectable` class instead. Example;
```
new ol.Overlay({
  element: overlayEl,
  id: overlayId,
  selectable: false
});
```

Happy to open a PR if someone can confirm that my understanding is correct and that I'm not simply misunderstanding how to control an overlay's selectability.

fredj:
Note: the selectable css class was added in #6741

fredj:
Instead of having a `selectable` option, what about adding a `className` like all the ol.controls?
The default value will be `ol-overlay-container ol-selectable`.

localjo:
Yes, I think being consistent with other ol.controls is a good idea. A class name option should work just as well.

fredj:
A PR will be greatly appreciated!

bobobobo:
PR #7377 implements support for className on ol.Overlay

fredj:
fixed with #7377

PR #7377, opened 2017-10-19
Add support to specify CSS class name when creating ol.Overlay

bobobobo:
This adds a method to set CSS class name for ol.Overlay as requester in #7241

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -521,7 +521,8 @@ olx.AtPixelOptions.prototype.hitTolerance;
  *     insertFirst: (boolean|undefined),
  *     autoPan: (boolean|undefined),
  *     autoPanAnimation: (olx.OverlayPanOptions|undefined),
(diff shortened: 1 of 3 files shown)


externs/olx.js
```diff
@@ -623,6 +623,12 @@ olx.OverlayOptions.prototype.autoPanAnimation;
  */
 olx.OverlayOptions.prototype.autoPanMargin;
 
+/**
+ * CSS class name. Default is `ol-overlay-container ol-selectable`.
+ * @type {string|undefined}
+ * @api
+ */
+olx.OverlayOptions.prototype.className;
```
fredj: please also add this in the `olx.OverlayOptions` `@typedef` above

bobobobo: Fixed!

fredj:
Thanks @bobobobo !

Issue #5121, opened 2016-03-24
Incorrect docs about ol.source.Raster 


drnextgis:
[This](http://openlayers.org/en/v3.14.2/apidoc/ol.source.Raster.html) page says:

> A source that transforms data from any number of input sources using an array of ol.raster.Operation functions to transform input pixel values into output pixel values.

But `operation` option is not an array.

itjope:
PR #7378 fixes this.

fredj:
fixed with #7378

PR #7378, opened 2017-10-19
Fix incorrect docs about ol.source.Raster

itjope:
As requested in #5121.


fredj:
The job never started on travis. Merging anyway

PR #7383, opened 2017-10-21
Handle null tile coordinates correctly

ahocevar:
This is a follow-up on #7349, which introduced a potential bug by comparing `urlTileCoord` with `undefined` instead of `null`.

### DIFF
--- a/src/ol/source/vectortile.js
+++ b/src/ol/source/vectortile.js
@@ -120,7 +120,7 @@ ol.source.VectorTile.prototype.getTile = function(z, x, y, pixelRatio, projectio
         tileCoord, projection);
     var tile = new ol.VectorImageTile(
         tileCoord,
-        urlTileCoord !== undefined ? ol.TileState.IDLE : ol.TileState.EMPTY,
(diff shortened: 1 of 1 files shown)


src/ol/source/vectortile.js
```diff
@@ -120,7 +120,7 @@ ol.source.VectorTile.prototype.getTile = function(z, x, y, pixelRatio, projectio
         tileCoord, projection);
     var tile = new ol.VectorImageTile(
         tileCoord,
-        urlTileCoord !== undefined ? ol.TileState.IDLE : ol.TileState.EMPTY,
+        urlTileCoord !== null ? ol.TileState.IDLE : ol.TileState.EMPTY,
```
tschaub: Do all falsey tile coords mean `EMPTY` or just `null`?

ahocevar: `#getTileCoordForTileUrlFunction()` returns either an `ol.TileCoord` of `null`, but never `undefined` or other falsey values.

Issue #7386, opened 2017-10-23
ol.proj.equivalent function


flexf1t:
Hello,
I've noticed that ol.proj.equivalent function does not compare some projection options/params, for example axisOrientation and extent. 
So it leads to this behavior:
`var proj = new ol.proj.Projection({ code: "EPSG:3857" });`
`var proj_neu = new ol.proj.Projection({ code: "EPSG:3857", axisOrientation: 'neu' });`
`ol.proj.equivalent(proj , proj_neu) // -> true`
This behavior is not what i was expecting and it breaks my application.
I think it is not good. May be i just miss something?
So, in my opinion it has two solutions:
1. add some new comparsions to the existing function ol.proj.equivalent (it works fine for me):
`var equalAxis = projection1.getAxisOrientation() === projection2.getAxisOrientation();`
`var equalExtent = projection1.getExtent() === projection2.getExtent();`
`var equal = equalUnits && equalAxis && equalExtent;`
2. add some new comparsions to the new function (for example ol.proj.equivalentHarder, etc) 

Thanks

fredj:
yes, it make sense to also compare the extent and the axis orientation

fredj:
see #7388

PR #7388, opened 2017-10-24
Add the axis orientation and the extent compare to ol.proj.equivalent function

flexf1t:
#7386 
Please note that CRC:84 and EPSG:4326 are not equivalent because of the axis orientation. 
The standard EPSG:4326 is lat-long axis. CRC:84 changing the axis from lat-long to long-lat. 

### DIFF
--- a/src/ol/proj.js
+++ b/src/ol/proj.js
@@ -331,12 +331,15 @@ ol.proj.equivalent = function(projection1, projection2) {
     return true;
   }
   var equalUnits = projection1.getUnits() === projection2.getUnits();
+  var equalAxis = projection1.getAxisOrientation() === projection2.getAxisOrientation();
+  var equalExtent = projection1.getExtent() === projection2.getExtent();
+  var equal = equalUnits && equalAxis && equalExtent;
   if (projection1.getCode() === projection2.getCode()) {
-    return equalUnits;
+    return equal;
   } else {
     var transformFn = ol.proj.getTransformFromProjections(
         projection1, projection2);
-    return transformFn === ol.proj.cloneTransform && equalUnits;
+    return transformFn === ol.proj.cloneTransform && equal;
   }
 };
 
--- a/test/spec/ol/proj/index.test.js
+++ b/test/spec/ol/proj/index.test.js
@@ -54,27 +54,25 @@ describe('ol.proj', function() {
     });
 
     it('gives that urn:ogc:def:crs:EPSG:6.6:4326, EPSG:4326 are ' +
-        'equivalent',
-    function() {
+        'equivalent', function() {
       _testAllEquivalent([
         'urn:ogc:def:crs:EPSG:6.6:4326',
         'EPSG:4326'
       ]);
     });
 
-    it('gives that CRS:84, EPSG:4326 are ' +
-       'not equivalent',
-    function() {
-      var proj1 = new ol.proj.Projection({
-        code: 'CRS:84'
-      });
-      var proj2 = new ol.proj.Projection({
-        code: 'EPSG:4326'
-      });
-      expect(ol.proj.equivalent(proj1, proj2)).to.not.be.ok();
-    });
+    it('gives that CRS:84, EPSG:4326 are not equivalent',
+        function() {
+          var proj1 = new ol.proj.Projection({
+            code: 'CRS:84'
+          });
+          var proj2 = new ol.proj.Projection({
+            code: 'EPSG:4326'
+          });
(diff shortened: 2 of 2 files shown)


flexf1t:
For example i have ol.View with proj "EPSG: 28405" (it has an 'enu' axis orientation) and i also have some WMS sources also with "EPSG: 28405" but 'neu' axis orientation. 
If i just set to source proj axisOrientation: 'neu' it will make no changes because ol.proj.equivalent returns true in this case.
If i also set 'neu' axis orientation to the view proj it will work well, but it will not work with 'enu' axis oritations sources. This fixes makes everything work perfect for me.
Wrong:
default
Right:
default
Why "EPSG: 28405" has 'enu' but WMS sources - 'neu'? That is a difficult question. By the way Gauss–Krüger coordinate system originally has 'neu' axis orientation.

ahocevar:
Can you share a code example that shows your setup?

flexf1t:
Actually its very simple code (but EPSG:28405 only works with this fix: [proj4js #258](https://github.com/proj4js/proj4js/pull/258/commits/f99791853e924a63fc1d227f3c3bb761fb2b98e5)):
```
var map = new Map();
function Map() {
    this._wms = null;
    this._map = null;
    this._proj = new ol.proj.Projection({
        code: "EPSG:28405"
    });
    this._view = new ol.View({
        center: ol.proj.transform([39.71, 43.59], 'EPSG:4326', this._proj),
        projection: this._proj,
        zoom: 4
    });
    this._layers = [
        new ol.layer.Tile({
            title: 'OSM',
            source: new ol.source.OSM()
        }),
        new ol.layer.Tile({
            title: 'WMS',
            source: new ol.source.TileWMS({
                projection:
                    new ol.proj.Projection({
                    code: 'EPSG:28405',
                    axisOrientation: 'neu',
                }),
                url: 'http://somelocalwms/with/epsg/28405',
                params: { 'LAYERS': "atlas", 'TILED': true },
            })
        })
    ];
    this._map = new ol.Map({
        target: 'map',
        layers: this._layers,
        view: this._view
    });
};
```

ahocevar:
Ok, like I suspected. This works with your pull request, but it goes through unnecessary and expensive raster reprojection for the WMS layer. It's better to do something like this instead:
```js
proj4.defs('urn:x-ogc:def:crs:EPSG:28405', proj4.defs('EPSG:28405'));
ol.proj.get('EPSG:28405').setAxisOrientation('neu');
```

Then use `projection: 'urn:x-ogc:def:crs:EPSG:28405'` for the view and `projection: 'EPSG:28405'` for the WMS source.

flexf1t:
Of cource it works when you change proj4 definition, but is it right way?

ahocevar:
Yes, that's exactly the way it should be. According to the EPSG database, EPSG:2845 has an axis order of 'northing, easting' (an information you do not get from proj4.js, so you modify the `ol.Projection` object. And the alternative code 'urn:x-ogc:def:crs:EPSG:28405' is simply an alias that we use here with the standard axis order of 'easting, northing'.

ahocevar:
@flexf1t Were you able to verify that my suggested setup works for you?

flexf1t:
@ahocevar It works this way: 
```
proj4.defs('urn:x-ogc:def:crs:EPSG:28405', proj4.defs('EPSG:28405'));
//look at '+axis=neu' here:
proj4.defs('EPSG:28405', "+proj=tmerc +lat_0=0 +lon_0=27 +k=1 +x_0=5500000 +y_0=0 +ellps=krass +towgs84=23.92,-141.27,-80.9,-0,0.35,0.82,-0.12 +axis=neu +units=m +no_defs"); 
```
But
`ol.proj.get('EPSG:28405').setAxisOrientation('neu');` 
There is no 'setAxisOrientation' function in api :(

Hanmac:
i have the similar problem with my projections from user data
so i need to load the info from proj4,  http://epsg.io
but i don't always know if the axis are correct (i don't know if i need '+axis=neu')

i tried to fix it with a similar problem that the read data might have the wrong axes #7368

for WFS it might be depend on how the projections are formated inside the WFS file?
http://docs.geoserver.org/stable/en/user/services/wfs/webadmin.html

so EPSG:XXXX need a flip compared urn:x-ogc:def:crs:EPSG:XXXX and to urn:ogc:def:crs:EPSG::XXXX ?

ahocevar:
@fredj @flexf1t OpenLayers does pick up the axis orientation from the proj4 definition, so I'd say adding `+axis=neu` to that custom definition is the way to go.

PR #7410, opened 2017-11-01
Add getFeatureType and setFeatureType functions to ol.format.WFS

fredj:
To be able to change the `featureType` of an exiting `ol.format.WFS` instance


fredj:
Thanks for the review

PR #7411, opened 2017-11-01
Add getLayers and setLayers functions to ol.format.WMSGetFeatureInfo

fredj:
Same idea as #7410 but for `ol.format.WMSGetFeatureInfo`

### DIFF
--- a/src/ol/format/wmsgetfeatureinfo.js
+++ b/src/ol/format/wmsgetfeatureinfo.js
@@ -64,6 +64,22 @@ ol.format.WMSGetFeatureInfo.layerIdentifier_ = '_layer';
 
 
 /**
+ * @return {Array.<string>} layers
+ */
+ol.format.WMSGetFeatureInfo.prototype.getLayers = function() {
+  return this.layers_;
+};
+
+
+/**
+ * @param {Array.<string>} layers Layers to parse.
+ */
+ol.format.WMSGetFeatureInfo.prototype.setLayers = function(layers) {
+  this.layers_ = layers;
+};
+
+
+/**
  * @param {Node} node Node.
  * @param {Array.<*>} objectStack Object stack.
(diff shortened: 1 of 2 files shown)


src/ol/format/wmsgetfeatureinfo.js
```diff
@@ -64,6 +64,22 @@ ol.format.WMSGetFeatureInfo.layerIdentifier_ = '_layer';
 
 
 /**
+ * @return {Array.<string>} layers
+ */
+ol.format.WMSGetFeatureInfo.prototype.getlayers = function() {
```
ahocevar: Shouldn't this be camel case (`getLayers`)?

src/ol/format/wmsgetfeatureinfo.js
```diff
@@ -64,6 +64,22 @@ ol.format.WMSGetFeatureInfo.layerIdentifier_ = '_layer';
...
+  return this.layers_;
+};
+
+
+/**
+ * @param {Array.<string>} layers Layers to parse.
+ */
+ol.format.WMSGetFeatureInfo.prototype.setlayers = function(layers) {
```
ahocevar: Shouldn't this be camel case (`setLayers`)?

src/ol/format/wmsgetfeatureinfo.js
fredj: ouch. yes, changed

fredj:
I don't need the `@api` annotations because OpenLayers is compiled with my application but I can add them.

ahocevar:
> I don't need the @api annotations because OpenLayers is compiled with my application but I can add them.

I'll leave it up to you. Do you want to have this documented and maintain it? Then add the annotation. If you don't want anyone to know about it and use it, then leave it as-is.

fredj:

PR #7414, opened 2017-11-02
Only split text at line angle changes

ahocevar:
We can do better than we currently do when it comes to rendering text along lines. Currently, we render character by character. But since we don't use glyphs, but fonts, we lose ligatures by doing so.

This pull request suggests changing the rendering of text along lines to only split text at angle changes. This will result in ligatures being preserved, unless they happen to cross an angle change.

Performance wise, I did a lot of profiling and optimized the new code so it performs equally well as the current one.


ahocevar:
@marcjansen, you might be interested in this since I know you're a friend of ligatures (afi) :smile:.

marcjansen:
I won't be able to look into this before next week, unfortunately

PR #7421, opened 2017-11-05
Keep longitude between -180 and 180

tschaub:
This makes it so the `ol.proj.toLonLat()` function returns longitude values between -180 and 180.

### DIFF
--- a/src/ol/proj.js
+++ b/src/ol/proj.js
@@ -3,6 +3,7 @@ goog.provide('ol.proj');
 goog.require('ol');
 goog.require('ol.Sphere');
 goog.require('ol.extent');
+goog.require('ol.math');
 goog.require('ol.proj.EPSG3857');
 goog.require('ol.proj.EPSG4326');
 goog.require('ol.proj.Projection');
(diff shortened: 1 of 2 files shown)


src/ol/proj.js
```diff
@@ -282,8 +283,13 @@ ol.proj.fromLonLat = function(coordinate, opt_projection) {
  * @api
  */
 ol.proj.toLonLat = function(coordinate, opt_projection) {
-  return ol.proj.transform(coordinate,
+  const lonLat = ol.proj.transform(coordinate,
```
fredj: do we want to start using `const` from es6?

tschaub: Thanks for catching this @fredj.  I'll change this to var for consistency and then we can discuss > es5 separately.

marcjansen: Nice catch.

tschaub: When we do introduce features beyond ES5 to the source, we should use Babel to handle transforms to our target browsers when publishing (instead of introducing a dependency on the Closure Compiler).  That said, if we are not targeting IE 10, `const` (and `let`) would make it through untransformed.

PR #7431, opened 2017-11-08
Avoid returning undefined zoom

tschaub:
There is still a bit of (user hostile) legacy in our handling of zoom levels.  This is based on the misperception that people would stop caring about zoom and would find true meaning in resolution.  Part of the remaining legacy is that `view.getZoom()` sometimes returns `undefined` instead of a number.

The goal of this change is to make it so that `view.getZoom()` always returns a number if the view resolution is defined.  Specifically:
 * When resolution is `undefined`, the method returns `undefined`.
 * When the resolution is higher than the max resolution, it returns a number that is lower than the min zoom.
 * When the resolution is lower than the min resolution, it returns a number that is higher than the max zoom.

I hope the parallels are clear.

### DIFF
--- a/src/ol/view.js
+++ b/src/ol/view.js
@@ -788,8 +788,9 @@ ol.View.prototype.getState = function() {
 
 
 /**
- * Get the current zoom level. Return undefined if the current
- * resolution is undefined or not within the "resolution constraints".
+ * Get the current zoom level.  If you configured your view with a resolutions
+ * array (this is rare), this method may return non-integer zoom levels (so
+ * the zoom level is not safe to use as an index into a resolutions array).
  * @return {number|undefined} Zoom.
  * @api
  */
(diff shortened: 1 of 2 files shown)


src/ol/view.js
```diff
@@ -788,8 +788,9 @@ ol.View.prototype.getState = function() {
 
 
 /**
- * Get the current zoom level. Return undefined if the current
- * resolution is undefined or not within the "resolution constraints".
+ * Get the current zoom level.  If you configured your view with a resolutions
+ * array (this is rare), this method may return non-integer zoom levels (so
+ * the zoom level is not safe to use as an index into a resolutions array).
```
gberaudo: Returning the closest zoom level looks desirable to me.

However I always assumed the zoom level was an the index in the resolutions array, an integer.
It looks strange to me to return a non-integer value. Why not return the closest integer?

I am probably missing something.
@fredj, what do you think about this change?

gberaudo: I now understand the rationale. With a non-integer value it becomes possible to design a richer zoom level selector, where the user can visually see where he stands between 2 integer / pixel perfect zoom levels. Great.

bartvde: this is what we called fractionalZoom in openlayers 2 http://dev.openlayers.org/examples/fractional-zoom.html

gberaudo: Thanks @bartvde for the pointer.

tschaub: Also note that we return non integer values for zoom *before* this change.  The difference here is the behavior when the resolution is higher than the max or lower than the min.  See the tests for detail.

gberaudo: I see. Thanks @tschaub for the explanation.

Issue #7434, opened 2017-11-08
Tiled WMS source does not use defined projection code to fetch tiles


pjeweb:
CodePen: https://codepen.io/anon/pen/dZOZmw?editors=0010

As my WMS server is not accessible for the public, the CodePen won't show the layer anyways, but the http requests made (see developer tools) show that the code `EPSG:3857` is used for fetching when I expect the defined code`EPSG:900913` to be used.

    http://domain.tld/mapserver?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=MyLayer&WIDTH=256&HEIGHT=256&CRS=EPSG%3A3857&STYLES=&BBOX=993069.8714810088%2C6990624.85884908%2C997961.8412912601%2C6995516.828659331

instead of

    http://domain.tld/mapserver?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=MyLayer&WIDTH=256&HEIGHT=256&CRS=EPSG%3A900913&STYLES=&BBOX=993069.8714810088%2C6990624.85884908%2C997961.8412912601%2C6995516.828659331

<hr>

- [x] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [x] I have searched GitHub to see if a similar bug or feature request has already been reported.
- [x] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [x] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

pjeweb:
Thanks for the quick merge and cleanup work. Btw. is there a time schedule for future releases? I could not find one.

PR #7435, opened 2017-11-08
Always use source projection loading image tiles

pjeweb:
Related issue: https://github.com/openlayers/openlayers/issues/7434

Use the defined source projection despite it being equivalent to the rendering projection to enable using the specified alias code name instead of the code provided by the renderer's projection object. For example: The server only knows the name "EPSG:900913", which is an alias for the equivalent "EPSG:3857", that the renderer uses (by default).

<hr>


pjeweb:
Closing due to failing checks and wrong committer (business user instead of CLA-signed person.

PR #7436, opened 2017-11-08
Always use source projection loading image tiles

pjeweb:
Related issue: https://github.com/openlayers/openlayers/issues/7434

Use the defined source projection despite it being equivalent to the rendering projection to enable using the specified alias code name instead of the code provided by the renderer's projection object. For example: The server only knows the name "EPSG:900913", which is an alias for the equivalent "EPSG:3857", that the renderer uses (by default).

<hr>

### DIFF
--- a/src/ol/source/tileimage.js
+++ b/src/ol/source/tileimage.js
@@ -241,7 +241,7 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
       !this.getProjection() ||
       !projection ||
       ol.proj.equivalent(this.getProjection(), projection)) {
-    return this.getTileInternal(z, x, y, pixelRatio, /** @type {!ol.proj.Projection} */ (projection));
+    return this.getTileInternal(z, x, y, pixelRatio, /** @type {!ol.proj.Projection} */ (this.getProjection() || projection));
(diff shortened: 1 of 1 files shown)


src/ol/source/tileimage.js
```diff
@@ -241,7 +241,7 @@ ol.source.TileImage.prototype.getTile = function(z, x, y, pixelRatio, projection
       !this.getProjection() ||
       !projection ||
       ol.proj.equivalent(this.getProjection(), projection)) {
-    return this.getTileInternal(z, x, y, pixelRatio, /** @type {!ol.proj.Projection} */ (projection));
+    return this.getTileInternal(z, x, y, pixelRatio, /** @type {!ol.proj.Projection} */ (this.getProjection() || projection));
```
tschaub: This function now has four `this.getProjection()` calls.  Could you please move the `sourceProjection` declaration [from below](https://github.com/pjeweb/openlayers-1/blob/b9313e76f4fed26b2fbfc953e6c3e3de6976616d/src/ol/source/tileimage.js#L257) to the top of the function and use it in the other three places?

ahocevar: I created #7438 to address this.

PR #7438, opened 2017-11-08
Call getProjection() only once

ahocevar:
Follow-up on #7436

Addresses https://github.com/openlayers/openlayers/pull/7436#discussion_r149706028.


ahocevar:
That was an oversight. Fixed. I also successfully removed the 2nd type cast.

Issue #7437, opened 2017-11-08
Determine what to do about failing tests


tschaub:
I'm seeing frequent test failures that eventually resolve themselves after restarting the Travis job.  Given how long it takes to run a CI job, it is getting pretty painful to prepare a pull request.  I'm going to use this ticket to list the failed tests as they come up:

 - [x] `Edge 15.15063.0 (Windows 10 0.0.0) ol.render.canvas ol.render.canvas.checkFont() clears the label cache for fonts that become available FAILED` (#7440)
 - [ ] `Edge 15.15063.0 (Windows 10 0.0.0) ol.View #animate() clears the ANIMATING hint when animations are cancelled FAILED
	Error: expected 1 to equal 2`
 - [ ] `Edge 15.15063.0 (Windows 10 0.0.0) ol.View #animate() completes multiple staggered animations run in parallel FAILED
	Error: expected false to equal true`
 - [ ] `07 11 2017 21:56:26.180:ERROR [launcher.sauce]: Can not start safari (macOS 10.12)
  Failed to start Sauce Connect:
  7 Nov 21:56:26 - Sauce Connect could not establish a connection.` ([build failure](https://travis-ci.org/openlayers/openlayers/builds/298793515), #7456)

We should either:
 * decide these are bugs in the tests and fix them
 * decide these are bugs in the code and fix them
 * throw in the towel and change the tests to fail less often

bartvde:
yes saw this one as well in https://github.com/openlayers/openlayers/pull/7420#issuecomment-342116814

ahocevar:
Found two more:

 * `Edge 15.15063.0 (Windows 10 0.0.0) ol.View #animate() clears the ANIMATING hint when animations are cancelled FAILED
	Error: expected 1 to equal 2`
 * `Edge 15.15063.0 (Windows 10 0.0.0) ol.View #animate() completes multiple staggered animations run in parallel FAILED
	Error: expected false to equal true`

tschaub:
I've seen those as well.  I copied those up to the list at the top.  Feel free to edit and add more.

I also updated the title to reflect that our goal should really be to determine why these are failing.

mike-000:
Another one?  Adding a checkbox to an example in #9931 doesn't seem related to ol.TileQueue

```
Chrome 77.0.3844 (Linux 0.0.0) ol.TileQueue #loadMoreTiles() works when tile queues share tiles FAILED
[...]

PR #7440, opened 2017-11-09
Attempt to make font loading tests more stable

ahocevar:
Also changes things to use setInterval() instead of setTimeout().

This should fix the checkFont() test that sometimes fails in Edge (see https://github.com/openlayers/openlayers/issues/7437#issue-272294985).

### DIFF
--- a/test/spec/ol/render/canvas/index.test.js
+++ b/test/spec/ol/render/canvas/index.test.js
@@ -6,7 +6,7 @@ goog.require('ol.render.canvas');
 describe('ol.render.canvas', function() {
 
   var font = document.createElement('link');
-  font.href = 'https://fonts.googleapis.com/css?family=Inconsolata';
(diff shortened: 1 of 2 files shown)


ahocevar:
I ran the tests 5 times now, and the test in question did not fail a single time in Edge.

test/spec/ol/render/canvas/index.test.js
```diff
@@ -6,7 +6,7 @@ goog.require('ol.render.canvas');
 describe('ol.render.canvas', function() {
 
   var font = document.createElement('link');
-  font.href = 'https://fonts.googleapis.com/css?family=Inconsolata';
+  font.href = 'https://fonts.googleapis.com/css?family=Abel&text=wmytzilWMYTZIL@#/&?$%10';
```
marcjansen: Ithink the value of the text parameter should be url encoded.

ahocevar: Good catch. Updated.

PR #7445, opened 2017-11-10
Load css rules from full-screen.css in examples

### LINKED ISSUE #7443: Full page view in ie
In one of your examples, there is a missing clause, and the example doesn't go full page view in internet explorer.

https://openlayers.org/en/latest/examples/full-screen-drag-rotate-and-zoom.html?q=

Because it only uses:
map:-moz-full-screen
map:-webkit-full-screen 
map:fullscreen

However, in another example, you have one more clause, and it works for ie.

https://openlayers.org/en/latest/examples/full-screen.html

This one includes:
map:-moz-full-screen
map:-webkit-full-screen 
map:fullscreen
map:-ms-fullscreen

fredj:
fixes #7443

### DIFF
--- a/examples/full-screen-drag-rotate-and-zoom.css
+++ b/examples/full-screen-drag-rotate-and-zoom.css
@@ -4,6 +4,9 @@
 .map:-webkit-full-screen {
   height: 100%;
 }
+.map:-ms-fullscreen {
+  height: 100%;
+}
 .map:fullscreen {
   height: 100%;
 }


fredj:
> Doesn't this make the source code view and editing in codepen unusable for this example?

yes; it will break the codepen

fredj:
@ahocevar can you take a second look please?

fredj:
Thanks for the review

PR #7454, opened 2017-11-10
Add text background rendering and text padding

### LINKED ISSUE #4577: Feature text with background color

ahocevar:
This pull request adds the ability to render background boxes around text. The following new `ol.style.Text` options are added:

 * `backgroundFill` An `{ol.style.Fill}` for the background box.
 * `backgroundStroke` An `{ol.style.Stroke}` for the background box outline.
 * `padding` An `{Array<number>}` with top, right, bottom and left padding values in pixels. The padding does not only apply to text, but is also used as padding for decluttering.

Background boxes are only supported for text styles with `placement: 'point'`.

fuchsia-olive

Fixes #4577.


ahocevar:
This is now ready for review.

ahocevar:
Thanks for the revier @marcjansen.

PR #7462, opened 2017-11-13
Respect metersPerUnit in ScaleLine control

### LINKED ISSUE #7309: ScaleLine control does not respect metersPerUnit of the projection
- [x] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue. -> snippet here

I've been using the following projection and scalebar code to create a metric scalebar for an image viewer, using zoomify raster tiles.

```
var getPointResolution = function(a) {
  return a
}

var projOptions = {
  code: 'customcode',
  units: 'pixels',
  extent: [0, 0, width, height],
  metersPerUnit: 1 / data.pixelsPerMeter,
  getPointResolution: getPointResolution,
}

var olProj = new ol.proj.Projection(projOptions)

var olView = new ol.View({
  projection: viewer.olProj,
  minZoom: 1,
  maxZoom: viewer.maxNativeZoom,
})

var controlScale = new ol.control.ScaleLine({
  units: 'metric',
  target: ...,
})

map.addControl(controlScale)
```

This worked perfectly in all 3.x and 4.x versions up till 4.2.0. It displays totally different values starting from 4.3.0, including the latest 4.3.4 version. The displayed value is 1000 m instead of 5 mm for example.

Since this was not in the release notes, I believe it is a bug.

ahocevar:
Fixes #7309.

### DIFF
--- a/src/ol/control/scaleline.js
+++ b/src/ol/control/scaleline.js
@@ -174,6 +174,9 @@ ol.control.ScaleLine.prototype.updateElement_ = function() {
     ol.proj.Units.METERS;
   var pointResolution =
       ol.proj.getPointResolution(projection, viewState.resolution, center, pointResolutionUnits);
+  if (units != ol.control.ScaleLineUnits.DEGREES) {
+    pointResolution *= projection.getMetersPerUnit();
+  }
 
   var nominalCount = this.minWidth_ * pointResolution;
   var suffix = '';
--- a/test/spec/ol/control/scaleline.test.js
+++ b/test/spec/ol/control/scaleline.test.js
@@ -1,9 +1,9 @@
-
-
+goog.require('ol');
 goog.require('ol.Map');
 goog.require('ol.View');
 goog.require('ol.control.ScaleLine');
 goog.require('ol.proj');
+goog.require('ol.proj.Projection');
 
 describe('ol.control.ScaleLine', function() {
   var map;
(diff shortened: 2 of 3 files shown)


src/ol/control/scaleline.js
```diff
@@ -174,6 +174,9 @@ ol.control.ScaleLine.prototype.updateElement_ = function() {
     ol.proj.Units.METERS;
   var pointResolution =
       ol.proj.getPointResolution(projection, viewState.resolution, center, pointResolutionUnits);
+  if (units != ol.control.ScaleLineUnits.DEGREES) {
+    pointResolution *= projection.getMetersPerUnit();
+  }
```
bartvde: can we do an if else here to avoid calling ```ol.proj.getPointResolution``` when it's not needed?

ahocevar: `getPointResolution` is always needed. We're just multiplying it with the meters per unit, because these are not taken into account by `getPointResolution`.

bartvde: ah I missed the ```*=``` part, looks good then

test/spec/ol/control/scaleline.test.js
```diff
@@ -265,6 +265,40 @@ describe('ol.control.ScaleLine', function() {
       var innerHtml4326 = ctrl.element_.innerHTML;
       expect(innerHtml4326).to.not.be(innerHtml3857);
     });
+
+    it.only('Projection\'s metersPerUnit affect scale for non-degree units', function() {
```
oterral: `only` should be removed

bartvde: good call :)

ahocevar: Good that our coverage check fails when an `.only` is accidentally left in the test code.

PR #7477, opened 2017-11-15
Make text states available for replay time

ahocevar:
This fixes a regression introduced by #7414, causing text along lines to be rendered with incorrect styles and text measurements: At replay time, we cannot use the text replay's fillState, strokeState and textState - these are only kept in sync with the style being rendered at replay creation time. Instead, we now map these states to the cache keys so we can access them.


marcjansen:
Code looks good to me. Please merge at will.

PR #7480, opened 2017-11-16
Create a new image when loading tile after an error

ahocevar:
Fixes a regression described in https://github.com/openlayers/openlayers/pull/7337#issuecomment-344891623.


mattlovett:
Hello Andreas,

Thank you for your quick response!

Yours, Matt

PR #7481, opened 2017-11-16
Make zoom to extent control extensible

gberaudo:
With this change it becomes possible to enhance the base control. For example, with 3D capabilities.

### DIFF
--- a/src/ol/control/zoomtoextent.js
+++ b/src/ol/control/zoomtoextent.js
@@ -22,9 +22,9 @@ ol.control.ZoomToExtent = function(opt_options) {
 
   /**
    * @type {ol.Extent}
-   * @private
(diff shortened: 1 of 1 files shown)


src/ol/control/zoomtoextent.js
```diff
@@ -67,11 +67,11 @@ ol.control.ZoomToExtent.prototype.handleClick_ = function(event) {
 
 
 /**
- * @private
+ * @protected
  */
 ol.control.ZoomToExtent.prototype.handleZoomToExtent_ = function() {
```
fredj: remove the `_` from the function name

gberaudo: Indeed, now fixed.

PR #7489, opened 2017-11-21
Allow string to be passed as ol.interaction.Draw type

fredj:


fredj:
Thanks for the review

PR #7499, opened 2017-11-23
Remove attributions from olx.FrameState instances

fredj:
`attributions` was removed from `olx.FrameState` in #7329


tschaub:
Looks good @fredj.

PR #7516, opened 2017-11-27
Deprecate exceedLength and replace with overflow

ahocevar:
Addresses a suggestion made in https://github.com/openlayers/openlayers/pull/7328#discussion_r148940858.

### DIFF
--- a/externs/olx.js
+++ b/externs/olx.js
@@ -7853,6 +7853,7 @@ olx.style.StrokeOptions.prototype.width;
  *     maxAngle: (number|undefined),
  *     offsetX: (number|undefined),
  *     offsetY: (number|undefined),
+ *     overflow: (boolean|undefined),
  *     placement: (ol.style.TextPlacement|string|undefined),
  *     scale: (number|undefined),
  *     rotateWithView: (boolean|undefined),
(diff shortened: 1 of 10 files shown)


externs/olx.js
```diff
@@ -7917,6 +7921,16 @@ olx.style.TextOptions.prototype.offsetX;
 olx.style.TextOptions.prototype.offsetY;
 
 
+/**
+ * For polygon labels or when `placement` is set to `'line'`, allow text to
+ * exceed the width of the polygon at the the label position or the length of
```
bartvde: double the

src/ol/render/canvas/textreplay.js
```diff
@@ -510,7 +510,7 @@ ol.render.canvas.TextReplay.prototype.setTextStyle = function(textStyle, declutt
     var font = textStyle.getFont() || ol.render.canvas.defaultFont;
     ol.render.canvas.checkFont(font);
     var textScale = textStyle.getScale();
-    textState.exceedLength = textStyle.getExceedLength();
+    textState.overflow = textStyle.get();
```
bartvde: should be getOverflow()

PR #7521, opened 2017-11-28
fix setMinZoom/setMaxZoom

cs09g:
fix setMinZoom/setMaxZoom method when view has resolutions property. 

#6746

### DIFF
--- a/src/ol/interaction/interaction.js
+++ b/src/ol/interaction/interaction.js
@@ -181,12 +181,14 @@ ol.interaction.Interaction.zoom = function(view, resolution, opt_anchor, opt_dur
 ol.interaction.Interaction.zoomByDelta = function(view, delta, opt_anchor, opt_duration) {
   var currentResolution = view.getResolution();
   var resolution = view.constrainResolution(currentResolution, delta, 0);
-  var resolutions = view.getResolutions();
 
-  resolution = ol.math.clamp(
-      resolution,
-      view.getMinResolution() || resolutions[resolutions.length - 1],
-      view.getMaxResolution() || resolutions[0]);
+  if (resolution !== undefined) {
+    var resolutions = view.getResolutions();
+    resolution = ol.math.clamp(
+        resolution,
+        view.getMinResolution() || resolutions[resolutions.length - 1],
+        view.getMaxResolution() || resolutions[0]);
+  }
 
   // If we have a constraint on center, we need to change the anchor so that the
   // new center is within the extent. We first calculate the new center, apply
--- a/src/ol/interaction/interaction.js
+++ b/src/ol/interaction/interaction.js
@@ -6,6 +6,7 @@ goog.require('ol');
 goog.require('ol.Object');
 goog.require('ol.easing');
 goog.require('ol.interaction.Property');
+goog.require('ol.math');
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/ol/interaction/interaction.js
```diff
@@ -180,6 +181,12 @@ ol.interaction.Interaction.zoom = function(view, resolution, opt_anchor, opt_dur
 ol.interaction.Interaction.zoomByDelta = function(view, delta, opt_anchor, opt_duration) {
   var currentResolution = view.getResolution();
   var resolution = view.constrainResolution(currentResolution, delta, 0);
+  var resolutions = view.getResolutions();
+
+  resolution = ol.math.clamp(
+      resolution,
```
cs09g: it's very strict type check. should change `resolution || 0` for `resolution: undefined`?

ahocevar: No, only do the clamp when `resolution !== undefined`.

src/ol/view.js
```diff
@@ -815,7 +821,7 @@ ol.View.prototype.getZoomForResolution = function(resolution) {
   var max, zoomFactor;
   if (this.resolutions_) {
     var nearest = ol.array.linearFindNearest(this.resolutions_, resolution, 1);
-    offset += nearest;
+    offset = nearest;
```
ahocevar: The `offset` variable is no longer used.

cs09g: `offset` is used when it returns.

ahocevar: Right :smile: - had missed that in the diff view.

cs09g:
Thanks for the review!!

src/ol/interaction/interaction.js
```diff
@@ -181,6 +182,14 @@ ol.interaction.Interaction.zoomByDelta = function(view, delta, opt_anchor, opt_d
   var currentResolution = view.getResolution();
   var resolution = view.constrainResolution(currentResolution, delta, 0);
 
+  if (resolution !== undefined) {
+    var resolutions = view.getResolutions();
+    resolution = ol.math.clamp(
+        resolution,
+        view.getMinResolution() || resolutions[resolutions.length - 1],
+        view.getMaxResolution() || resolutions[0]);
+  }
```
tschaub: Previously, we would let people temporarily zoom outside of the min/max range.  This was intentional.  When using a trackpad for scrolling, for example, it feels nice to have the map temporarily zoom in/out beyond the constraints and then settle back to the constrained value.  This change breaks that behavior.

tschaub: That might not be true...

tschaub: Looks like this isn't the case.  Ignore the noise.

cs09g: Thanks for the mention. I've got to check on mac device but without that condition, it doesn't work as expected for setMin/MaxZoom.

waxenegger:
@cs09g and @tschaub 

I'd just like to point at one side-effect of this PR in the case when one wants to solely work with resolutions and hands no zoom bounds to the View. 

https://github.com/openlayers/openlayers/blob/v4.6.4/src/ol/view.js#L1113

In that case both, min/max resolution are forced to be the zoom max/min defined further up. What's more, if one happens to have lots of resolution levels the default value for min resolution will be clamped at index 28 of the resolutions (the default max for zoom levels)

cs09g:
@waxenegger you're right about the second one. and when the zoom changed by `setZoom`, it can be moved to beyond min/max zoom. should be fixed.

PR #7524, opened 2017-11-29
Snap view center to pixel

### LINKED ISSUE #7500: Jittery text and icons when panning around
I am drawing text from GeoJSON, going from `ol.source.Vector` to `ol.layer.Vector` via a styling function that  sets `ol.style.Text`.

When dragging around slowly with the mouse, but also visible during Kinetic animations, the text is jittery, as it moves around the various pixel boundaries. This is visible also in the [example](http://openlayers.org/en/latest/examples/vector-labels.html), if you pan around slowly. I've tested this with Safari, Chrome and Firefox on macOS.

A similar thing happens with icons that are not using `snapToPixel: false`. 

A few possible fixes that would fix this jittery movement while keeping everything crisp would be:

1. Ensure the Kinetic Animations only do whole pixel movements. 
2. During Kinetic Animation, disable pixel snapping globally. Re-enable it when the animation finishes and re-render to the pixel grid.

Thanks!

- [X] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

ahocevar:
This change also revealed two issues with label rendering:

 * 9307c35 improves vertical text placement in Firefox. It seems that Firefox measures font heights differently. By using a 'middle' textBaseline on the canvas, we get much closer to the placement in other browsers.
 * b468f1c adds the pixel ratio to the cache key. Otherwise, labels rendered with different pixel ratio get mixed up
 * 74bf846 fixes the tests to do geometry transforms on clones instead of the original. Otherwise, the result is different depending on which tests have run already.

Fixes #7500.


ahocevar:
Thanks for the review.

PR #7543, opened 2017-12-06
Donut polygon labels do not get a chance to get rendered

ahocevar:
Since polygon labels are by default only rendered when they do not overflow the polygon, donut polygons with the label point inside an inner ring do not get labels rendered any more. This is because intersections are only calculated for the outer ring, and not for all rings of a polygon.


ahocevar:

PR #7566, opened 2017-12-11
Proper projection construction

tschaub:
The tests were missing a `new` when constructing a projection (since ffb7d72c905ccbb542974bbf4597197eb3f24993).  This causes a test failure when we transform the sources to modules (`this` is undefined in the projection function).


ahocevar:
Actually the projection `proj` is not needed for this test - not for the source neither for the view. So it could as well be removed. The test is really only about different tile grids for source and view.

PR #7575, opened 2017-12-12
Add named exports for extent

ahocevar:


marcjansen:
I am so happy to see us finally moving in this direction. Thanks Andreas and Tim!

I hope to cut some time tomorrow for looking into this, if you review each other before that, that is fine as well.

gberaudo:
@ahocevar, I remember to have seen a discussion about using named exports and their benefit in some comments of an issue, but I could not find it again.
Do you have information or a pointer? Is it an improvement for tree shaking?

ahocevar:
@gberaudo I don't find the ticket comment either. But we discussed in a meeting a while ago. Anyway, this is indeed an improvement for tree shaking.

ahocevar:
This is now ready for review.

fredj:
(disclaimer: I've no es6 module experience)

Not specific to this pull request but I don't understand why we are using named and default export:
```js
// foo.js
export function bar() { ... }
export function baz() { ... }

export default {
  bar: bar,
  baz: baz
};
```
Is it to be able to support the following use `import` ?
```js
import foo from 'foo.js'

foo.bar();
foo.baz();
```
And:
```js
import {bar, baz} from 'foo.js'
bar();
baz();
```

ahocevar:
Exactly, and mostly for compatibility with the existing ol package. We will provide a codeshift transform to update the paths (CamelCase classes instead of lowercase), and running that transform is all that ol v4.x users will need to do to upgrade.

tschaub:
Regarding the default export, it might be worth revisiting this before release.  In the transform, we could rewrite things like this:
```js
// using ol@4
import extent from 'ol/extent';
```
```js
// using ol@5
import * as extent from 'ol/extent';
```

This would be relatively straightforward to implement if we consistently provide named exports from lowercase modules (and provide a default export from uppercase modules).

I don't think we should make any more changes in this branch though.  We can remove the default exports after preparing the transform.

fredj:
I've also [read](https://medium.com/@timoxley/named-exports-as-the-default-export-api-670b1b554f65) that the default export can be created by importing and exporting itself:
```js
// foo.js
import * as myself from './foo.js';
export default myself

export function bar() { ... }
export function baz() { ... }
```

tschaub:
I've opened #7580 to get rid of the default export from `ol/index.js`.  I think we can follow the rules mentioned there and still provide a decent upgrade path for ol@4 users.

PR #7576, opened 2017-12-12
Named exports from ol/index.js

tschaub:
This adds named exports to `ol/index.js`.  For backwards compatibility, the default export is an object with all of the old API properties.  Internally, we only use the named exports.


marcjansen:
I am so happy to see us finally moving in this direction. Thanks Andreas and Tim!

I hope to cut some time tomorrow for looking into this, if you review each other before that, that is fine as well.

PR #7580, opened 2017-12-13
No default export from lowercase modules

tschaub:
We had originally thought we would provide a default export in addition to named exports from modules like `ol/proj` and `ol/extent`.  The reason for this was to make upgrading from the ol@4 package easier - we would provide a code transform that updated the case of import specifiers for things like `ol/Map` and leave `ol/proj` type imports untouched.

Instead, we can follow these rules:
 * if a module name is uppercase (e.g. `ol/Map`), it has a default export (in the future, it may also have named exports, but this is not currently the case)
 * if a module name is lowercase (e.g. `ol/proj`), it has named exports and no default export.

So our code transform for the 4 to 5 upgrade will do this:
```js
// before, using ol@4
import extent from 'ol/extent';
```
```js
// after, using ol@5
import * as extent from 'ol/extent';
```

### DIFF
--- a/src/ol/index.js
+++ b/src/ol/index.js
@@ -119,15 +119,3 @@ var uidCounter_ = 0;
 export function getUid(obj) {
   return obj.ol_uid || (obj.ol_uid = ++uidCounter_);
 }
-
-
-export default {
-  getUid: getUid,
-  nullFunction: nullFunction,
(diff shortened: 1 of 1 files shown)


ahocevar:
Works for me, and seems very consistent. I'll update the wiki.

PR #7582, opened 2017-12-13
Remove default export from ol/extent

ahocevar:
Follow-up on #7580.


tschaub:
I reworked some of these same examples in #7581 to use named imports from `ol/extent`.  I'll merge this so I can rebase that work.

PR #7584, opened 2017-12-14
Named exports from ol/easing

fredj:
adds named exports from the `ol/easing` module.


fredj:
Thanks for the review, I've created #7587 to not forget about the version number

PR #7597, opened 2017-12-15
More es6 named export

fredj:


fredj:
renamed, thanks for the review

PR #7598, opened 2017-12-15
Modular proj4 integration

### LINKED ISSUE #5628: Modular handling of proj4js
As part of the conversion to modules, I would like to import proj4js in a standard way as a module instead of creating a global. There is currently the `setProj4` function, but this is not really standards-compliant. To support both those who want to use Proj4 and those who don't, I think you would have to have 2 separate classes, one of which imports Proj4 and another which doesn't, but I'm not sure how you would implement that without changing the API.

So what I've done is disable proj4js in OL, and implement the functionality in my app. Code like the following supports the most common case of transforming in/out of 4326, assuming `ol` and `proj4` are imported:

``` javascript
function addProjection(projCode, projDef) {
  // add the projection to Proj4 defs
  proj4.defs(projCode, projDef);
  // add the projection object to ol.proj
  ol.proj.addProjection(new ol.proj.Projection({
    code: projCode,
    units: 'm' // could be taken from proj4.defs
  }));
  // get transform functions from Proj4 ...
  var proj4Transform = proj4('EPSG:4326', projCode);
  // ... and add to ol.proj
  ol.proj.addCoordinateTransforms('EPSG:4326', projCode,
      proj4Transform.forward, proj4Transform.inverse);
}
```

This is similar to what is done in `examples/wms-custom-proj.js`. However, the current `ol.proj` code supports transforms in/out of all the added projections `ol.proj.projections_`. ATM this can't be done in the app, because no function to get the list of current projections exists. For the moment, I am maintaining my own duplicate list, but it would be better to add this export.

Any comments, or other/better ways of doing this?

ahocevar:
This pull request changes the API for proj4 integration a bit, so proj4 related code is only added to the build when the users requests so. The braking change is that the `ol.proj.setProj4()` function is gone, and instead there is a `ol/proj/proj4` module with a `register()` function. This function should be called whenever changes are made to the proj4 registry, e.g.
```js
import proj4 from `proj4`;
import {register} from 'ol/proj/proj4';

proj4.defs('EPSG:4326', '+proj=longlat +datum=WGS84 +no_defs');
register(proj4);
```
From here on, everything is the same as previously.

This pul request also changes exports to named exports for `ol/proj/transforms`.

Fixes #5628.

### DIFF
--- a/examples/reprojection-by-code.js
+++ b/examples/reprojection-by-code.js
@@ -6,6 +6,7 @@ import {get as getProjection, getTransform} from '../src/ol/proj.js';
 import {register} from '../src/ol/proj/proj4.js';
 import _ol_source_OSM_ from '../src/ol/source/OSM.js';
 import _ol_source_TileImage_ from '../src/ol/source/TileImage.js';
+import proj4 from 'proj4';
 
 
 var map = new _ol_Map_({
--- a/examples/reprojection-by-code.js
+++ b/examples/reprojection-by-code.js
@@ -3,6 +3,7 @@ import _ol_View_ from '../src/ol/View.js';
 import * as _ol_extent_ from '../src/ol/extent.js';
 import _ol_layer_Tile_ from '../src/ol/layer/Tile.js';
 import {get as getProjection, getTransform} from '../src/ol/proj.js';
+import {register} from '../src/ol/proj/proj4.js';
 import _ol_source_OSM_ from '../src/ol/source/OSM.js';
 import _ol_source_TileImage_ from '../src/ol/source/TileImage.js';
 
@@ -42,6 +43,7 @@ function setProjection(code, name, proj4def, bbox) {
 
   var newProjCode = 'EPSG:' + code;
   proj4.defs(newProjCode, proj4def);
+  register(proj4);
   var newProj = getProjection(newProjCode);
   var fromLonLat = getTransform('EPSG:4326', newProj);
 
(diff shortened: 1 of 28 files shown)


ahocevar:
This is now ready for review.

examples/reprojection-by-code.js
```diff
@@ -3,6 +3,7 @@ import _ol_View_ from '../src/ol/View.js';
 import * as _ol_extent_ from '../src/ol/extent.js';
 import _ol_layer_Tile_ from '../src/ol/layer/Tile.js';
 import {get as getProjection, getTransform} from '../src/ol/proj.js';
+import {register} from '../src/ol/proj/proj4.js';
```
tschaub: Should this example also `import proj4 from 'proj4'` instead of relying on the global?

examples/reprojection.js
```diff
@@ -4,6 +4,7 @@ import * as _ol_extent_ from '../src/ol/extent.js';
 import _ol_format_WMTSCapabilities_ from '../src/ol/format/WMTSCapabilities.js';
 import _ol_layer_Tile_ from '../src/ol/layer/Tile.js';
 import {get as getProjection} from '../src/ol/proj.js';
+import {register} from '../src/ol/proj/proj4.js';
```
tschaub: Same question about importing proj4.  Would be nice to have all our examples explicitly import dependencies eventually.

tschaub:
Looks great @ahocevar.  Seeing your upgrade notes above made me realize we should be trying to add some detail to the `upgrade-notes.md` as we go (I should have done this for the deprecation removal PR).  If not here, we should starting adding detail there soon to make things easier on ourselves when we release.

PR #7601, opened 2017-12-15
Rename interaction variables

marcjansen:
There are more renames, but this is hard to look at already.

Please review.

### DIFF
--- a/examples/snap.js
+++ b/examples/snap.js
@@ -1,6 +1,6 @@
 import _ol_Map_ from '../src/ol/Map.js';
 import _ol_View_ from '../src/ol/View.js';
-import _ol_interaction_Draw_ from '../src/ol/interaction/Draw.js';
+import Draw from '../src/ol/interaction/Draw.js';
 import _ol_interaction_Modify_ from '../src/ol/interaction/Modify.js';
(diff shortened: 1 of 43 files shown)


examples/snap.js
```diff
@@ -73,7 +73,7 @@ Modify.init();
 
 var optionsForm = document.getElementById('options-form');
 
-var Draw = {
+var ExampleDraw = {
```
tschaub: Curious if the linter caught this after rename, if you searched for 'var Draw' before, or if something else.

fredj:
rebased

tschaub:
Hrm.  That was a failed experiment using GitHub's web-based conflict editor.  I'll fix this up...

tschaub:
@fredj - I'm curious to hear how you did the rebase.  GitHub is letting me modify this branch through the web editor, but I can't push directly (and I failed at conflict resolution with the web editor).

I opened #7607 with a rebased version of @marcjansen's commits.  I'll close this PR in favor of that one.

examples/snap.js
marcjansen: The linter warns, as the variable was already used.

fredj:
@tschaub I've rebased locally on the pushed force to Marc's repository.

PR #7608, opened 2017-12-15
Facilitate the removal of externs/olx.js

tschaub:
This distributes things in the `externs/olx.js` file to places where they belong.

The `@typedef` annotations for the compiler will get added to the relevant module.  For example, a `FooOptions` type will go in the `ol/Foo` module.  The other `@type` annotations will get distributed as `@param` annotations in the modules where they belong.  For example, `FooOptions.prototype.bar` will become `options.bar` in for the constructor in `ol/Foo`.

Note that for 5.x, we assume applications being built together with the library, so no externs like the ones in `externs/olx.js` are needed any more. Since the full build will not be built with Closure Compiler any more, we do not need the externs there either (previously they helped us avoid renaming of properties in object literals passed from the application to the library).

### DIFF
--- a/tasks/t.js
+++ b/tasks/t.js
@@ -1,0 +1,191 @@
+const fs = require('fs');
+const util = require('util');
+const path = require('path');
+
+const readFile = util.promisify(fs.readFile);
+const writeFile = util.promisify(fs.writeFile);
+const externs = path.resolve(__dirname, '../externs');
+
+function parse(source) {
+  const length = source.length;
+
+  let index = 0;
+  const IN_CODE = 'in code';
+  const IN_COMMENT = 'in comment';
+
+  function next() {
+    const char = source.charAt(index);
+    ++index;
+    return char;
+  }
+
+  function scanLine() {
+    let char = next();
+    let line = '';
+    while (index < length && char !== '\n') {
+      line += char;
+      char = next();
+    }
+    return line;
+  }
+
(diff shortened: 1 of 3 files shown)


tasks/t.js
```diff
@@ -0,0 +1,191 @@
...
+    const params = data[name].params;
+    if (!params.length) {
+      throw new Error(`No params for ${name}`);
+    }
+
+    source += '/**\n';
+    source += ` * @param {${name}} options TODO: repace this\n *\n`;
+    params.forEach(param => {
+      const description = param.description.split('\n').join('\n * ');
+      source += ` * @param ${param.type} options.${param.name} ${description}\n`;
```
fredj: it should be `${param.name}` instead of `options.${param.name}`

tschaub:
@fredj and @ahocevar - I think we should merge this to facilitate the work outlined in #7947.

If you agree, the instructions on that pull request can be updated to be something like this:
 * distribute comments from `externs/xol.js` to other parts of the library
 * remove the corresponding parts from `externs/olx.js`
 * run `node tasks/t.js`
 * create a pull request

This will keep us from having to update this branch when working on removing things from `externs/olx.js`.  It will also let others contribute to the effort even if they cannot edit this branch.  When we are done we can remove `tasks/t.js`.

tschaub:
Merging to make it easier to work on olx redistribution.

PR #7613, opened 2017-12-16
forEachLayerAtPixel signature, hitTolerance

### LINKED ISSUE #7283: hitTolerance for ol.Map#forEachLayerAtPixel
The forEachLayerAtPixel is missing the hitTolerance parameter that forEachFeatureAtPixel has.

I would create a PR for the next major version release of openlayers as it induces an API change.

simonseyock:
This pull request is a follow up to #5995, fixes #7283 and also adresses #7284.

What it contains:
* unified signature for forEachLayerAtPixel and forEachFeatureAtPixel. This is an api change.
* hitTolerance parameter for forEachLayerAtPixel.
* removes some uneccessary thisArgs

There is one thing missing, and I don't really know how to handle this at the moment. it is about renderer/canvas/IntermediateCanvas#forEachLayerAtPixel. There the same mechanism regarding the hitTolerance is needed that is normally used in the render/canvas/ReplayGroup.
I see 2 solutions for this right now: 1. move the hitTolerance mechanism to another class that is accessible for both classes or 2. defer the hitDetection functionality in IntermediateCanvas to some render class that has access.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -45,6 +45,28 @@ The following methods did get the optional this (i.e. opt_this) arguments remove
 * source/TileUTFGrid#forDataAtCoordinateAndResolution
 * source/Vector#forEachFeature, #forEachFeatureInExtent, #forEachFeatureIntersectingExtent
 
+
+#### `Map#forEachLayerAtPixel` parameters have changed
+ 
+If you are using the layer filter, please note that you now have to pass in the layer filter via an `AtPixelOptions` object. If you are not using the layer filter the usage has not changed.
+
+Old syntax:
+```
+    map.forEachLayerAtPixel(pixel, callback, callbackThis, layerFilterFn, layerFilterThis);
+```
+
+New syntax:
+
+```
+    map.forEachLayerAtPixel(pixel, callback, {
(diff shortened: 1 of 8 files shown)


simonseyock:
I don't exactly understand under which circumstances the code in this else branch is called:
https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/canvas/IntermediateCanvas.js#L130

If I compare it with the code in forEachFeatureAtCoordinate (right above) and the circumstance under which these methods are called, I would assume that this else branch is never called **or** the code in forEachFeatureAtCoordinate is buggy and could lead to an exception if there is no forEachFeatureAtCoordinate method on the source.

forEachFeatureAtCoordinate is called here:
https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/Map.js#L148
forEachLayerAtCoordinate is called here:
https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/canvas/Map.js#L229

Both methods are only called when visibleAtResolution and layerFilter return true. I can't see more restrictions. Do I miss something here?

Or am I mistaken and there is something more different for these calls?

tschaub:
I reworked these commits and merged them with #8314.  Thanks @simonseyock.

PR #7615, opened 2017-12-16
Remove logos

tschaub:
This removes the `logo` option from the map and sources.  As mentioned in the upgrade notes, to add a logo to the map, the desired markup can be added directly to the viewport container.  And source attributions can be used to add source-specific markup (like a logo).


fredj:
Thanks Tim !

PR #7621, opened 2017-12-17
Replace ZeroClipboard with Clipboard.js

fredj:
See #4217

clipboard.js now supports Safari 10.0+


fredj:
Thanks for the review

PR #7622, opened 2017-12-17
Named export

fredj:


src/ol/TileRange.js
```diff
@@ -45,15 +45,15 @@ var _ol_TileRange_ = function(minX, maxX, minY, maxY) {
  * @param {ol.TileRange|undefined} tileRange TileRange.
  * @return {ol.TileRange} Tile range.
  */
-_ol_TileRange_.createOrUpdate = function(minX, maxX, minY, maxY, tileRange) {
+TileRange.createOrUpdate = function(minX, maxX, minY, maxY, tileRange) {
```
tschaub: Instead of tacking this property on to the `TileRange` function, we can have `createOrUpdate` be a named export.  This is probably better handled as part of a separate effort.

fredj: good point

PR #7624, opened 2017-12-17
Remove eslint-plugin-openlayers-internal rules

fredj:
No needed anymore.

The only one that is not `goog` related is the [enum](https://github.com/openlayers/eslint-plugin-openlayers-internal/blob/master/enum.js) check; do we want/need to keep it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "coveralls": "3.0.0",
     "eslint": "4.13.1",
     "eslint-config-openlayers": "7.0.0",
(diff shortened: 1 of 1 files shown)


tschaub:
I think we can remove the enum check as well @fredj.  It isn't a bad convention, but it was only really needed to facilitate the modules transform.  Either way, please merge.

fredj:
eslint-plugin-openlayers-internal is now completely removed

PR #7626, opened 2017-12-17
Named export from ol/asserts

tschaub:
This exports `assert` from `ol/asserts`.


tschaub:
I missed #7622 when creating this.  Closing in favor of that.

PR #7632, opened 2017-12-19
Named exports from ol/css module

marcjansen:
Title says it all.

Please review.

### DIFF
--- a/src/ol/css.js
+++ b/src/ol/css.js
@@ -1,7 +1,6 @@
 /**
  * @module ol/css
  */
-var _ol_css_ = {};
 
 
 /**
@@ -10,7 +9,7 @@ var _ol_css_ = {};
  * @const
  * @type {string}
  */
-_ol_css_.CLASS_HIDDEN = 'ol-hidden';
+export var CLASS_HIDDEN = 'ol-hidden';
 
 
 /**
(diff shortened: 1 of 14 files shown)


marcjansen:
Makes sense, will see how and when I can update this.

marcjansen:
@ahocevar updated as requested. This is ready for review again.

I was thinking of using `const`, but we do not use it elsewhere, right? E.g.:

```js
// instead of …
export var CLASS_CONTROL = 'ol-control';
// … we could use const
export const CLASS_CONTROL = 'ol-control';
```

 If you want, I can update accordingly

src/ol/css.js
```diff
@@ -54,7 +54,7 @@ _ol_css_.CLASS_CONTROL = 'ol-control';
  * @param {string} The CSS font property.
  * @return {Object.<string>} The font families (or null if the input spec is invalid).
  */
-_ol_css_.getFontFamilies = (function() {
+export var getFontFamilies = (function() {
```
fredj: `export function` instead of `export var` ?

marcjansen: I stumbled over this as well, here we are using an IIFE, so we can cache certain things. I wasn't sure how to handle this. simpy changing as you suggested raises an 'Expression expected' error. What do you suggest?

IMO `var` isn't technically false, is it?

marcjansen: Since things are passing with this, I am tempted to merge this.

fredj: yes, please merge

marcjansen:
Thanks for the reviews, @ahocevar & @fredj

PR #7633, opened 2017-12-19
Import assert function directly

marcjansen:
This changes one remaining occurence from

```js
import asserts from './asserts.js';
```

to

```js
import {assert} from './asserts.js';
```

and gets rid of the following warning:

```
WARNING in ./src/ol/extent.js
491:4-11 "export 'default' (imported as 'asserts') was not found in './asserts.js'
 @ ./src/ol/extent.js
 @ ./test/spec/ol/extent.test.js
```

Please review.

### DIFF
--- a/src/ol/extent.js
+++ b/src/ol/extent.js
@@ -1,7 +1,7 @@
 /**
  * @module ol/extent
  */
-import asserts from './asserts.js';
+import {assert} from './asserts.js';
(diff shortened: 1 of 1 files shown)


marcjansen:
Thanks for the review, @ahocevar

fredj:
Thanks Marc

PR #7637, opened 2017-12-19
Named exports from the ol/color module

marcjansen:
Title says it all, please review.

### DIFF
--- a/src/ol/color.js
+++ b/src/ol/color.js
@@ -3,7 +3,6 @@
  */
 import {assert} from './asserts.js';
 import _ol_math_ from './math.js';
-var _ol_color_ = {};
 
 
 /**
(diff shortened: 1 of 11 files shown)


src/ol/color.js
```diff
@@ -167,38 +164,39 @@ _ol_color_.fromStringInternal_ = function(s) {
...
     assert(false, 14); // Invalid color
   }
   return /** @type {ol.Color} */ (color);
-};
+}
 
 
 /**
+ * TODO this function is only used in the test, we probably shouldn't export it
  * @param {ol.Color} color Color.
```
tschaub: This could be moved to the test module.

tschaub: Scratch that.  I see it used above.

marcjansen:
Thanks for the review.

PR #7639, opened 2017-12-19
Don't store private function into GeoJSON

fredj:


fredj:
> We can also remove the trailing underscore and @Private annotation.

done

PR #7643, opened 2017-12-20
Replace ol/Sphere with ol/sphere

### LINKED ISSUE #7628: Remove ol.Sphere class
See https://github.com/openlayers/openlayers/issues/4954#issuecomment-191164846:
Instead, we should have an `ol.sphere` namespace that provides static functions to be called with a radius, and constants for common radii.

tschaub:
This gets rid of the `Sphere`, renames `ol/Sphere.js` to `ol/sphere.js`, and creates named exports for the previously static `getArea` and `getLength` functions.  In addition, the `getDistance` function is exported to replace `sphere#haversineDistance`.

The `getArea` and `getLength` functions did not work for non-geographic geometry collections with more than one component.

Since this touched `ol/geom/Polygon`, I also included a commit to use named exports rather than static functions there.

Fixes #7628.


src/ol/geom/Polygon.js
```diff
@@ -366,32 +367,35 @@ Polygon.prototype.setFlatCoordinates = function(layout, flatCoordinates, ends) {
...
  *     the polygon vertices.
  * @param {number=} opt_n Optional number of vertices for the resulting
  *     polygon. Default is `32`.
+ * @param {number=} opt_sphereRadius Optional radius for the sphere (defaults to
+ *     the Earth's mean radius using the WGS84 ellipsoid).
  * @return {ol.geom.Polygon} The "circular" polygon.
  * @api
  */
-Polygon.circular = function(sphere, center, radius, opt_n) {
+export function circular(center, radius, opt_n, opt_sphereRadius) {
```
marcjansen: I think we need upgrade notes here as well because of the changed arguments. Hopefully not to much work, and probably helpful.

test/spec/ol/sphere.test.js
```diff
@@ -1,97 +1,173 @@
...
+  describe('getDistance()', function() {
+
+    var expected = [{
+      c1: [0, 0],
+      c2: [0, 0],
+      distance: 0
+    }, {
+      c1: [0, 0],
+      c2: [45, 45],
+      distance: 6671704.814011975
```
marcjansen: For the record: the values differ because of a different radius we used in the tests, right? Looking over the diff this feels odd at first.

tschaub: The more significant difference is that the old `haversineDistance` returned values in kilometers.  The `getDistance` function returns values in meters.  But yes, the numbers are also different because of the different radius.  The old `haversineDistance` used the WGS84 semi-major axis (not really appropriate for a spherical calculation).  The new `getDistance` uses the mean radius of the same ellipsoid (more appropriate for a spherical calculation).

src/ol/geom/Polygon.js
tschaub: Good point, I'll add a commit that updates the upgrade doc.

PR #7648, opened 2017-12-21
Overriding ol.source.Vector#refresh() to trigger an actual refresh, fix #7044

### LINKED ISSUE #7044: ol.source.Vector.refresh() does not update features removed from remote source
Hi all, I dug through closed and open issues but couldn't find an answer, so I'm opening this.

Documentation says about `ol.source.Vector.refresh()`:
> Refreshes the source and finally dispatches a 'change' event.

Imagine having a remote source, for example an API: 			
 ```
 source : new ol.source.Vector({
    url: "/api/tasks"
}
```

After deleting or adding a feature in the database/source, I expected calling `refresh()` to update the state of my map, including removal of deleted features or adding of new ones. However, this does not even trigger a GET request on the source in many cases. On the other side `ol.source.Vector.clear()` works surprisingly well. I expected it to just clear all features from the source but it also fetched up-to-date data from the source. Can anyone reproduce this? Did I misunderstand documentation here? Maybe this also is a chance to improve documentation. `refresh()` was introduced with #4868 which gives some insight, but still does not really explain why this is not working as expected for me.

By the way, thank you very much, all of you, for your work and time spent on this project. 👍

sedot42:
This pull request is supposed to fix #7044

ol.source.refresh() only dispatches a 'change' event. While this
triggers a reload, the #loadFeatures() method checks if the extents of
already loaded features match the extents of features to be loaded.
Features where only the content changed (eg metadata) won't be refreshed
using this approach.
Hence this commits overrides ol.source.Vector#refresh() to clear all
features. The triggered reload will properly refresh all features.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -885,6 +885,14 @@ VectorSource.prototype.loadFeatures = function(extent, resolution, projection) {
   }
 };
 
+/**
+ * Refreshes the source by clearing all features. clear() dispatches a
+ * 'change' event that triggers a reload of all features.
+ * @api
+ */
+VectorSource.prototype.refresh = function() {
+  this.clear();
+};
 
 /**
  * Remove an extent from the list of loaded extents.


RalphLAtGitHub:
From my point of view a similar issue arises at **ol.sourceXYZ.refresh()**. The function does not update the tiles from the server. Replacing the layer by a new one as suggested here for OL3 (http://jsfiddle.net/ahocevar/tLt5wdpz/) doesn't solve the problem. A solution would be great. Thanks a lot in advance!

sedot42:
I updated the source, but still am unsure whether this is the right way to solve the problem.

PR #7669, opened 2018-01-01
Use declutter tree only for text and image replays

### LINKED ISSUE #7664: Declutter selection error
Hi, I found a bug:
with the declutter option in a vector layer when I try to select features with a text style, selection is possible only for features with the text label visible in the viewport.

In the example if you click on the linestring at the top (Feature 1) this is not selecteable until the label is visible.

https://codepen.io/anon/pen/YYNgeP?editors=0010

ahocevar:
Fixes #7664.

### DIFF
--- a/src/ol/render/canvas/ReplayGroup.js
+++ b/src/ol/render/canvas/ReplayGroup.js
@@ -35,11 +35,6 @@ var _ol_render_canvas_ReplayGroup_ = function(
   _ol_render_ReplayGroup_.call(this);
 
   /**
-   * @type {ol.render.ReplayType}
-   */
-  this.currentReplayType_;
-
-  /**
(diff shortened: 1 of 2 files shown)


ahocevar:
We could `call()` the `hitDetectionCallback`, then we could avoid the `bind()`. I'll make that change before merging.

sbrandwoo:
I believe there are two slightly different issues with the interaction between `declutter` and `map.forEachFeatureAtPixel`.

* Features without text cannot be clicked on
* Features with text that has been hidden because of decluttering cannot be clicked on

Does this PR fix both of these cases?

Thanks

ahocevar:
This pull request fixed the former, the latter was fixed with #7519.

PR #7683, opened 2018-01-08
Move IGCZ enum to src/ol/format/IGC.js

fredj:


externs/olx.js
```diff
@@ -1538,15 +1538,15 @@ olx.format.TopoJSONOptions.prototype.layers;
...
+ * @typedef {{altitudeMode: (ol.format.IGCZ|string|undefined)}}
  */
 olx.format.IGCOptions;
 
 
 /**
  * Altitude mode. Possible values are `barometric`, `gps`, and `none`. Default
  * is `none`.
- * @type {ol.format.IGCZ|undefined}
+ * @type {ol.format.IGCZ|string|undefined}
```
tschaub: We'll have to remember to update this when we finish the `olx.js` transition work (see #7608).  When `IGCOptions` moves to `ol/format/IGC.js`, this type will become `{IGCZ|undefined}`.

tschaub: Actually, rather than trying to remember this later, can you change this type now (to `{IGCZ|undefined}`)?

PR #7686, opened 2018-01-08
More import renaming

tschaub:
More renaming work.


marcjansen:
Thanks for this. I am also `+1`.

PR #7692, opened 2018-01-11
Don't export private variables

fredj:


tschaub:
I agree on exporting enums.  Though they could be named exports.  E.g.

    import DragBox, {EventType} from 'ol/interaction/DragBox';

tschaub:
We need to work on transforming types in annotations to use relative paths.  This will be easier to automate (at least in part) with the way we currently  have enums structured.  Before moving enums, I think it would make sense to complete the type annotations change.

PR #7693, opened 2018-01-11
More import renaming

fredj:


tschaub:
Ack.  Wish I had checked the PRs before starting on #7695.  Fortunately, we made all the same renaming decisions :)

Thanks for tackling this @fredj.  I'll rebase after this is in and finish up the rest (with default exports only) in #7695.

tschaub:
I'll merge this now so I can rebase #7695.

PR #7697, opened 2018-01-12
Use const and let

tschaub:
This change brings blocked scoped variables and we enforce the ESLint [no-var](https://eslint.org/docs/rules/no-var) and [prefer-const](https://eslint.org/docs/rules/prefer-const) rules as errors.  This keeps us from inadvertently reassigning variables and discourages declaring them in a larger scope than where they are used.

If there is still interest in supporting IE 10 before we release 5.0 (it looks like IE 10 is about 0.8% of the market), we can transform when publishing.  But I'd be in favor of dropping IE 10 altogether (not sure if we even say we support it now).

Less significantly, this also drops our custom indent configuration.  While personal preferences may differ, I don't think there is strong justification for maintaining our own custom configuration.

Fixes #7696.


tschaub:
@marcjansen - I'm curious if reviewing this will break your phone.

marcjansen:
I habe various devices, so let's see...

tschaub:
Looking into the test failures now...

tschaub:
Ok, that was a fun one to chase down.  We currently declare `feature` **within** [a `while` loop](https://github.com/openlayers/openlayers/blob/0bf2b04deefe5bac7d381bee59e17799bfa4392a/src/ol/render/canvas/Replay.js#L586) where we [later reassign](https://github.com/openlayers/openlayers/blob/0bf2b04deefe5bac7d381bee59e17799bfa4392a/src/ol/render/canvas/Replay.js#L758) and then expect the value to be used [on subsequent iterations](https://github.com/openlayers/openlayers/blob/0bf2b04deefe5bac7d381bee59e17799bfa4392a/src/ol/render/canvas/Replay.js#L640) of the loop.  This worked only because the declaration had function scope - so although we were declaring it down in the while loop, it was assigned on for the enclosing function and was available on subsequent iterations of the loop.

This doesn't work with blocked scoped variables.
```js
function tryThis() {
  let foo;
  while (true) {
    let bar;
    if (!foo) {
      foo = true;
      bar = 42;
    } else {
      console.log(bar);
      break;
    }
  }
}
```

The version with `let` logs `undefined` while the same with `var` would log `42`.

Fortunately, our tests caught the issue and 50890c93e25c15cd4aaf4f7a7bd5bdb703b3832e addresses it.

We frequently declare variables in a larger scope than where they are needed (based on old habits of making big comma delimited `var` lists at the top of a function (this is still the case with things like `x` and `y` in the commit above).  We should work on narrowing the scope of these variables now that they are block scoped.  I'm hoping there aren't other cases where we declared a var too narrowly (within a block where it is needed outside the block or on subsequent iterations of `for` or `while`), as these will be new bugs to chase down now that they are block scoped.

tschaub:
I know this is a pain to review (the GitHub changes view breaks Chrome for me), but given that it is going to conflict with every other change, I'd like to not let it sit around for too long.  Thanks in advance for any reviews.

tschaub:
Thanks @ahocevar.  I too imagine that there could be problems lurking, but I don't really think it is all that common to declare a variable within a loop, conditionally set its value, and then rely on the value in a subsequent iteration (that is the pattern that breaks going from `var` to `let`).  But, we'll see!

PR #7698, opened 2018-01-12
Fix for loop in color test

fredj:


marcjansen:
Thanks please merge and nice catch.

PR #7703, opened 2018-01-14
Improved drawing experience on touch devices

ahocevar:
With this change, the Draw interaction allows accurate drawing when using finger/stylus. Geometries are not being updated while panning the map, and the vertex being drawn can be moved before lifting the finger/stylus when in a long press gesture.

On mouse devices, long press works now too, but nothing else changes.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,13 @@
 
 ### Next release
 
+#### Changed behavior of the `Draw` interaction
+
+For better drawing experience, especially on mobile devices, two changes were made to the behavior of the Draw interaction:
+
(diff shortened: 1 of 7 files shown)


ahocevar:
I have now made the new behavior configurable (it can also be turned off), and added tests.

tschaub:
@ahocevar - are the videos above labeled correctly?  In the "after" one, it looks to me like vertices are getting added on drag.  While in "before", you can drag the map without adding a vertex.

ahocevar:
They are labeled correctly @tschaub. The "after" video also shows how a vertex that is already set can be dragged on long press. It's probably easier to see live with the branch checked out.

ahocevar:
Thanks for the review @tschaub.

PR #7713, opened 2018-01-17
Better variables scoping

fredj:
related to https://github.com/openlayers/openlayers/pull/7697#issuecomment-357170584

just scratching the surface ...


fredj:
An other advantage of moving the `let` into the `for` loop is that the following error is detected by eslint:
```js
for (let i = 0, ii < invalidColors.length; i < ii; ++i) {
   // ...
}
```
```
/home/fredj/code/openlayers/test/spec/ol/color.test.js
  125:26  error  Parsing error: Unexpected token <
```

PR #7716, opened 2018-01-17
Better variables scoping

fredj:


fredj:
Thanks for the review

PR #7740, opened 2018-01-25
WIP: Upgrade to webpack 4

tschaub:
Hoping to see some build time improvements...

Current blockers:
 - [x] webpack-contrib/copy-webpack-plugin#211
 - [x] a release that includes webpack-contrib/uglifyjs-webpack-plugin#215
 - [ ] figure out why `splitChunks` is not working like I thought it should
 - [ ] figure out why `devtool: 'source-map'` no longer works

### DIFF
--- a/examples/webpack/config.js
+++ b/examples/webpack/config.js
@@ -1,10 +1,7 @@
-const MinifyPlugin = require('babel-minify-webpack-plugin');
 const CopyPlugin = require('copy-webpack-plugin');
 const ExampleBuilder = require('./example-builder');
 const fs = require('fs');
-const merge = require('webpack-merge');
 const path = require('path');
-const webpack = require('webpack');
 
 const src = path.join(__dirname, '..');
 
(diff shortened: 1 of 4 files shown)


tschaub:
I've opened webpack/webpack#6389 in hopes of getting suggestions on resolving production build issues.

schmidtk:
@tschaub I was trying to run the `build` task on master and ran into the `devtool: 'source-map'` issue. I poked around a bit and found these related issues:

webpack-contrib/babel-minify-webpack-plugin/issues/68
webpack/webpack-sources/issues/28

They unfortunately don't offer any solutions yet, but may be a good place to start. I'm using `eval-source-map` as a temporary workaround, but that results in `ol.js` being 6.4MB so also not a solution.

tschaub:
Thanks for the info @schmidtk.  We've been working on an alternative approach for building the examples (and the rest of the website) that seems like it will work.  This uses Rollup for bundling examples - and so far it is fast and sourcemaps work.

Building an application works well with webpack 4, and we'll make some examples of this available before the 5.0 release.  I'll close this for now since I don't think we'll have to make webpack 4 work building all the examples at once.

PR #7741, opened 2018-01-25
Type checking and full build creation

ahocevar:
This branch brings type checking capabilities and a script to create a full build of OpenLayers.
 * [x] Generate a task to create an entry point for a full build of the library.
 * [x] Add `build` script to `package.json` to create a full build.
 * [x] Find a way to perform type checks from our existing (or slightly modified) JSDoc annotations.
 * [x] Use [babel-plugin-jsdoc-closure](https://github.com/openlayers/babel-plugin-jsdoc-closure) to generate a transpiled source for type checking
 * [x] Add `typecheck` script to `package.json` to run type checks.


ahocevar:
At this point we have a full build again:

    npm run build

creates `build/ol/js` with source map, so no need for an `ol-debug.js` any more.

ahocevar:
I just created [https://github.com/openlayers/webpack-jsdoc-closure-loader](webpack-jsdoc-closure-loader), which will be used to create a bundle with webpack that Closure Compiler can consume for type checking and building.

ahocevar:
This is now ready for review.

Note that the type checker currently uses warnings instead of errors. We need to update all types before we can switch to errors. See #7792.

marcjansen:
Great work Andreas! I am the wrong person to give this a proper review but the code looks nice.

We usually do not use carets in packacke.json, though.

Looking forward to 5!

config/webpack.js
```diff
@@ -0,0 +1,14 @@
...
+  entry: './src/index.js',
+  output: {
+    filename: 'build/ol.js'
+  },
+  devtool: 'source-map',
+  plugins: [
+    new webpack.optimize.ModuleConcatenationPlugin(),
+    new MinifyPlugin()
+  ]
+};
```
tschaub: If not in this branch, we should rename this to be something more specific.  Like `webpack-full-build.js` or something.

tasks/generate-index.js
```diff
@@ -0,0 +1,157 @@
...
+      imports[defaultImport] = true;
+    } else if (namedExport.length > 1) {
+      const from = namedExport[0].replace(/^module\:/, './');
+      const importName = from.replace(/[.\/]+/g, '_');
+      const namedImport = `import * as ${importName} from '${getPath(from)}';`;
+      imports[namedImport] = true;
+    }
+  });
+
+  callback(null, symbols, Object.keys(imports).sort());
```
tschaub: Looks like this could just return instead of taking a callback.

tschaub:
Looks great @ahocevar.  Thanks for the work on this.

The comments above don't need to be addressed here.  Separately we can rework our tasks to remove the use of the async module (since we now have async/await).

ahocevar:
Thanks for the reviews @marcjansen and @tschaub. I'm merging this as-is, and we can make things nicer as we progress.

PR #7749, opened 2018-01-29
Check if the sketch feature is defined in finishDrawing

### LINKED ISSUE #7035: Function FinishDrawing cause error if zero point in geometry
https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/draw.js#L640

Calling ol.interaction.Draw.FinishDrawing doesn't check if there is a geometry and cause a javascript error.

Step to reproduce:

start an ol.interaction.draw
call finishDrawing() before adding any point
it cause an error

fredj:
Fixes #7035

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -714,6 +714,9 @@ Draw.prototype.removeLastPoint = function() {
  */
 Draw.prototype.finishDrawing = function() {
   const sketchFeature = this.abortDrawing_();
+  if (!sketchFeature) {
+    return;
+  }
   let coordinates = this.sketchCoords_;
   const geometry = /** @type {ol.geom.SimpleGeometry} */ (sketchFeature.getGeometry());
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Draw.js
```diff
@@ -714,37 +714,39 @@ Draw.prototype.removeLastPoint = function() {
...
-    // remove the redundant last point
-    coordinates.pop();
-    this.geometryFunction_(coordinates, geometry);
-  } else if (this.mode_ === Draw.Mode_.POLYGON) {
-    // remove the redundant last point in ring
-    coordinates[0].pop();
-    this.geometryFunction_(coordinates, geometry);
-    coordinates = geometry.getCoordinates();
-  }
+  if (sketchFeature) {
```
tschaub: Might be more readable & less prone to future error to shorten this block.  E.g.
```js
if (!sketchFeature) {
  return;
}
// ...
```

fredj: yes, changed

Issue #3050, opened 2014-12-16
Add hit and render properties to `ol.style.Style`


tschaub:
I'm interested in getting feedback on the idea of adding a `hit` and `render` properties to `ol.style.Style`.  This would allow developers to determine what style was used for hit detection vs. normal rendering.

By default, both `hit` and `render` would be true, and the behavior would be what you get today (all styles are rendered to the display context and the hit context).  By adding a style with `render` set to `false`, a developer could control the hit region for a feature.  I think this would be the most useful new configuration, and would allow people to avoid things like fills with 0.1% opacity.

Setting `hit` to `false` (and leaving `render` `true`) would support things like rendering a big shadow on an icon that wasn't "clickable" for example.

Setting both `hit` and `render` to false would support pointless power consumption.  The alternative to two booleans would be one property with three acceptable values, but this is more typo prone (for external developers not using enums and the compiler).

PR #7750, opened 2018-01-29
Don't assume fill for canvas polygon hit detection.

### LINKED ISSUE #4846: Unfilled polygon hit detection
The canvas replay renderer is currently inserting a default fill for the hit detection instructions for any polygon, regardless of whether it is filled or not:

https://github.com/openlayers/ol3/blob/master/src/ol/render/canvas/canvasreplay.js#L1334

My team would prefer to only hit detect polygons that are actually filled when hovering over the interior and not hit detect unfilled ones. The fillStyle from the actual polygon is right there on the `state` in that function, so why not just use that? There isn't a good way to work around this short of overriding the whole function (that I can see). Would this give you any heartburn to change? Or to be made configurable?

schmidtk:
The canvas renderer will hit detect an unfilled polygon when hovering inside of the outline. This is undesirable and also inconsistent with the WebGL polygon hit detection, which only detects the polygon if the mouse is over the outline. Filled polygons will be hit detected as they were before, either by hovering the fill or outline.

Fixes #4846.

### DIFF
--- a/src/ol/render/canvas/PolygonReplay.js
+++ b/src/ol/render/canvas/PolygonReplay.js
@@ -61,9 +61,9 @@ CanvasPolygonReplay.prototype.drawFlatCoordinatess_ = function(flatCoordinates, 
     }
     offset = end;
   }
-  this.hitDetectionInstructions.push(fillInstruction);
   if (fill) {
     this.instructions.push(fillInstruction);
+    this.hitDetectionInstructions.push(fillInstruction);
   }
   if (stroke) {
     this.instructions.push(strokeInstruction);
@@ -85,11 +85,12 @@ CanvasPolygonReplay.prototype.drawCircle = function(circleGeometry, feature) {
   }
   this.setFillStrokeStyles_(circleGeometry);
   this.beginGeometry(circleGeometry, feature);
-  // always fill the circle for hit detection
-  this.hitDetectionInstructions.push([
-    CanvasInstruction.SET_FILL_STYLE,
-    asString(defaultFillStyle)
-  ]);
+  if (state.fillStyle !== undefined) {
+    this.hitDetectionInstructions.push([
+      CanvasInstruction.SET_FILL_STYLE,
+      asString(defaultFillStyle)
+    ]);
+  }
   if (state.strokeStyle !== undefined) {
     this.hitDetectionInstructions.push([
       CanvasInstruction.SET_STROKE_STYLE,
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks @schmidtk.  I'm curious to hear other opinions on this.  We intentionally hit detect unfilled polygons now.  We could add an option if opinions are split.

schmidtk:
@tschaub Do you know why the Openlayers WebGL renderer chose not to hit detect unfilled polygons? That behavior is consistent with other libraries like Leaflet and Cesium, but diverges from the behavior of the canvas renderer.

ahocevar:
I have personally never used hit detection on unfilled polygons, but since quite some effort was put into making their whole covered area detectable, I'd assume that there are users relying on it. If we agree to change this behavior, then some code cleanup in addition to what is done in this pull request would be in order.

That said, I have no strong opinion on this.

The WebGL renderer is incomplete, so that's probably why it doesn't do the same.

tschaub:
I haven't dug through the history yet, but my recollection is that the hit behavior with fills was intended to be consistent with the DOM - where pointer events are relayed regardless of fill opacity.

Here is an [example with SVG](https://codepen.io/tschaub/pen/MQYEeE) for comparison.  The `hover` pseudo-class and click listener apply equally to the rect with and without a fill.

fill-events

Having an option to control this value would be nice, but perhaps awkward to implement.

fredj:
I also don't remember having used hit detection on unfilled polygons. But if I had to do it without this support in OpenLayers I would add an "invisible" fill color of something like `rgba(0, 0, 0, 0.0001)`.

And I guess removing (or making it optional) this feature would improve the performances.

schmidtk:
@tschaub Fill with no opacity is different than no fill in SVG. See [this example](https://codepen.io/anon/pen/YePrMy?editors=1111) where `fill="none"` is used instead of `fill-opacity="0"`. Perhaps the solution is to require a fill to hit detect inside a polygon, even if the opacity is 0?

tschaub:
Good clarification @schmidtk.  Oversight on my part.

Originally we did not hit detect polygons with no fill (see 3ef0b059491221ec3571c3e94b382f9a49d7c070).  And people started adding fills with very small opacities (I don't have a reference, but recall this).  Then we decided to always hit regardless of fill (see https://github.com/openlayers/openlayers/pull/1355/files and I think the prior implementation probably did the same).

At one point we thought about having a dedicated `hit` property on styles (see #3050).  But this went nowhere.  We could add an option to always hit detect polygon interiors regardless of the fill style (this would achieve the current behavior).  And by default we could be consistent with other libraries.  If we really want consistency with the DOM, as @schmidtk points out, we would hit if there is a fill regardless of the opacity.

tschaub:
It might be most straightforward to implement the DOM-consistent behavior:

 * if fill and non-zero opacity, use that for hit detection
 * if fill and zero opacity, use default fill for hit detection
 * if no fill, don't hit detect polygon interiors

And as an optimization, we could avoid rendering fills with zero opacity.  This would avoid passing around a very specific option.

schmidtk:
I agree with that. Is the first bullet necessary? I was thinking, "if fill, use default fill for hit detection". I wouldn't expect the color/opacity to matter for hit detection but let me know if that's incorrect.

tschaub:
> Is the first bullet necessary?

@schmidtk - I was imagining fills with a pattern that had some 0 opacity areas.  But perhaps simpler just to say if you have any fill, the whole interior is a hit.

schmidtk:
I refactored my changes to:
- Hit detect any polygon with a defined fill style, disregarding opacity.
- Skip rendering a fill if opacity is 0.
- Skip rendering and hit detection if fill style is not defined.

If anyone has feedback on a smarter way to check if a `ol.ColorLike` has 0 opacity, please let me know. I tried using `asArray` first, but that fails if passed `CanvasPattern|CanvasGradient`. `asString` works due to its default values, but that feels like a bit of a hack. It works, but I'm not fond of it.

schmidtk:
The Travis CI build failures appear to be caused by intermittent timeouts with the tests, unrelated to the PR changes. I see these locally (macOS / Chrome 64) from time to time as well, but they seem more common on Travis.

tschaub:
Thanks @schmidtk.  Apologies for the delay in getting it in.

If anyone wants to contribute rendering tests for this, they would be welcome.

ahocevar:
Getting hit detection on polygons that are not filled can be now be achieved by adding a fill with zero opacity:
```js
fill: new Fill({color: 'rgba(0, 0, 0, 0)'})
```

mike-000:
Should this logic also be applied to CircleStyle without fill?  Currently, for example, if a large box is draw over a small one the small box can be selected if fill is not defined, but if a large circle is drawn over a small one the small circle cannot be selected even if the CircleStyle has no fill.

PR #7765, opened 2018-02-05
Named exports from events condition

marcjansen:
Named exports from events condition. Please review.


tschaub:
@marcjansen - it looks like we are trying to use the default export [in `MouseWheelZoom`](https://github.com/openlayers/openlayers/blob/bdcb4bfbaeb0909234409bc9a8fe84e8746501b4/src/ol/interaction/MouseWheelZoom.js#L6).

marcjansen:
Thanks for the pointer!

marcjansen:
@tschaub now green. Thanks again for the hint.

marcjansen:
Thanks Tim.

PR #7767, opened 2018-02-05
Transformation functions: Fix spelling and use the symbol ×

marcjansen:
Title says it all. Please review.


marcjansen:
Thanks for the review.

PR #7771, opened 2018-02-06
Named exports from loadingstrategy

marcjansen:
Please review.


marcjansen:
Updated as suggested in #7772.

marcjansen:
Thanks for the review.

PR #7772, opened 2018-02-06
Named exports from filter module

marcjansen:
Please review.

### DIFF
--- a/src/ol/format/filter.js
+++ b/src/ol/format/filter.js
@@ -27,10 +27,10 @@ import Within from '../format/filter/Within.js';
  * @returns {!ol.format.filter.And} `<And>` operator.
  * @api
  */
-export const and = function(conditions) {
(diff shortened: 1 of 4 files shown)


src/ol/format/filter.js
```diff
@@ -28,9 +27,9 @@ const _ol_format_filter_ = {};
  * @returns {!ol.format.filter.And} `<And>` operator.
  * @api
  */
-_ol_format_filter_.and = function(conditions) {
+export const and = function(conditions) {
```
fredj: could you change it to `export function and(conditions) {`, like it's done elsewhere?

marcjansen: Of course. Will do later.

marcjansen:
Updated as suggested, @fredj

marcjansen:
Thanks for the reviews.

PR #7774, opened 2018-02-06
Pass destination extent to avoid garbage generation

fredj:
An other small memory optimization


tschaub:
The [test failures](https://travis-ci.org/openlayers/openlayers/builds/338068608) look related.

fredj:
Yes, the test failure is related.
When there's no intersection between the extents, the function returns an empty entent: `[Infinity, Infinity, -Infinity, -Infinity]`. But if an `opt_extent`is provided, this value is returned instead of an empty extent

fredj:
I've added a commit to fix this `getIntersection` issue

fredj:
Thanks for the review

Issue #7779, opened 2018-02-07
Use named exports from utility modules


tschaub:
This is not a formal categorization, but we have two broad categories of modules in the library:
 1. modules that export a constructor or an enum
 2. modules that export a collection of functions or constants

The first category of modules have UpperCamelCase names.  For example, [`ol/layer/VectorTile.js`](https://github.com/openlayers/openlayers/blob/3266ffc58ff50bef012af6bfd93ec6e3c5a7fe69/src/ol/layer/VectorTile.js) exports the `VectorTile` layer constructor.  To use the constructor, people can import the default export.  E.g.
```js
import VectorTile from 'ol/layer/VectorTile';
```

Enums like [`GeometryType`](https://github.com/openlayers/openlayers/blob/3266ffc58ff50bef012af6bfd93ec6e3c5a7fe69/src/ol/geom/GeometryType.js) fit into the same category.  E.g.
```js
import GeometryType from 'ol/geom/GeometryType';
// now use something like GeometryType.POINT
```

The second category of modules export just functions or constants.  For consistency, we want these to use named exports and *not* have a default export.  The `ol/proj` module is an example of this category.  When people use it, they import just what they need of the named exports.  E.g.
```js
import {toLonLat} from 'ol/proj';
```

Before the 5.0 release, we need to finish the work of adding named exports and removing the default export from these lowercase, utility modules.  As of this writing, it looks like there are 31 modules that still need reworking (based on `grep -E 'export default _([a-z]|_)+\b' -r src`):

- [x] `src/ol/net.js` (see #7780)
- [x] `src/ol/tilegrid.js` (see #7800)
- [x] `src/ol/renderer/webgl/defaultmapshader.js` (see #7788)
- [x] `src/ol/renderer/webgl/tilelayershader.js` (see #7788)
- [x] `src/ol/renderer/vector.js` (see #7800)
- [x] `src/ol/size.js` (see #7780)
- [x] `src/ol/reproj.js` (see #7780)
- [x] `src/ol/coordinate.js` (see #7800)
- [x] `src/ol/xml.js` (see #7783)
- [x] `src/ol/render/webgl/texturereplay/defaultshader.js` (see #7788)
- [x] `src/ol/render/webgl/linestringreplay/defaultshader.js` (see #7788)
- [x] `src/ol/render/webgl/circlereplay/defaultshader.js` (see #7788)
- [x] `src/ol/render/webgl/polygonreplay/defaultshader.js` (see #7788)
[...]

PR #7783, opened 2018-02-07
Named exports from ol/xml

raiyni:
Some long import lists. 

Per https://github.com/openlayers/openlayers/issues/7779


marcjansen:
This work is much appreciated 👏

Issue #2914, opened 2014-11-05
mouseposition control: add option not to clear out when mouse leaves map viewport


dasakigr:
I am new in github so sorry for my mistakes.
I started to learn ol3 recently. I noticed that when the mouse is removed from the map erases the contents of mouseposition something that does not happen in other programs GIS.
I managed this by changing the function undefinedHTML but there was a problem. There isn't setUndefinedHTML method. This I managed by changing the mousePositionControl.O variable when the mouse moves equal to the mouseposition text each time.
I try to set mousePositionControl.O variable on event mousepositionout but didn't work.
So the scope is to fix setUndefinedHTML and mousepositionout event

pedros007:
Maintainers (@tschaub @bartvde and @ahocevar?) What do you think of one of the following ways to implement this? I could try and work on a pull request.

1. Add an ol.control.MousePosition option to override [handleMouseOut](https://github.com/openlayers/openlayers/blob/c2f7baf940d0cbf5466d05076c5fee006f37379c/src/ol/control/MousePosition.js#L159-L162), similar to the way a user can override the `render` method
2. Add an option to  to not bind  [EventType.MOUSEOUT](https://github.com/openlayers/openlayers/blob/c2f7baf940d0cbf5466d05076c5fee006f37379c/src/ol/control/MousePosition.js#L176-L177)

marcjansen:
I would suggest to add an option `autoClear`, `clearOnMouseOut` (or a similar name) with `true` being the default. A PR would be welcomed!

fredj:
And what if instead of adding a new option, setting `undefinedHTML` to `undefined` keep the value?
It will break the API but it will be logical to me: `undefined` doesn't do anything and an empty string (or any string) replaces the coordinates with it

pedros007:
@marcjansen Sounds good, thanks.  I have it coded, but I am running into trouble writing a test for the feature.

[Here's a CodePen of my attempt](https://codepen.io/pschmitt/pen/WMGpaZ). I set a breakpoint in `MousePosition.prototype.handleMouseMove`, but it never seems to be triggered.  I expected this code to trigger [handleMouseMove](https://github.com/openlayers/openlayers/blob/master/src/ol/control/MousePosition.js#L148-L152):
```
pointerEvent = new ol.pointer.PointerEvent('mousemove', {clientX: 200, clientY: 300})
event = ol.MapBrowserPointerEvent('mousemove', map, pointerEvent);
[...]

PR #7784, opened 2018-02-07
Adds option to retain MousePosition when pointer leaves viewport

pedros007:
Implements #2914. 

Note I added a new option `clearOnMouseOut` rather than override [the default value](https://github.com/openlayers/openlayers/blob/master/src/ol/control/MousePosition.js#L56) for `undefinedHTML` as suggested here:  https://github.com/openlayers/openlayers/issues/2914#issuecomment-363461367

Let me know if you'd rather I change the default `undefinedHTML` behavior, which would break the API as fredj noted.

### DIFF
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -31,7 +31,9 @@ const COORDINATE_FORMAT = 'coordinateFormat';
  * callback.
  * @property {Element|string} [target] Specify a target if you want the
  * control to be rendered outside of the map's viewport.
- * @property {string} [undefinedHTML=''] Markup for undefined coordinates.
+ * @property {string} [undefinedHTML=''] Markup for undefined coordinates.  If
+ * `undefined`, then the last pointer position is retained when the pointer
+ * moves outside the viewport.
  */
 
 
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -72,6 +72,8 @@ const MousePosition = function(opt_options) {
     this.setProjection(options.projection);
   }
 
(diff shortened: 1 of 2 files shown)


tschaub:
I like @fredj's suggestion of avoiding another option.  I think that would be a pretty minor change to your work here @pedros007.  Something like:
```js
this.clearOnMouseOut_ = options.undefinedHTML !== undefined;
```

And the property name could be `renderOnMouseOut_` rather than `clearOnMouseOut_` in this case.

pedros007:
Thanks for the feedback!  What should `undefinedHTML` default to?  If it the default is not `undefined`, how do I set it?  Currently the default is set like so:

```javascript
  this.undefinedHTML_ = options.undefinedHTML !== undefined ? options.undefinedHTML : '';
```

which will not work, as `options.undefinedHTML=undefined` would carry meaning as a valid option, so the following returns `true`:

```javascript
({}.foo === undefined) === ({foo: undefined}.foo === undefined)
```

pedros007:
Nevermind, I could do something like
```javascript
this.undefinedHTML_ = 'undefinedHTML_' in options ?  options.undefinedHTML : '';
```
I'll work on an update tomorrow.

pedros007:
@tschaub I implemented your suggestions in de2a448

pedros007:
Do you need anything from me to get this in the next release?   I know there's a lot of work related to preparations of the 5.0 release.

The API change is that now `undefinedHTML=undefined` now has special meaning: do not clear the `ol-mouse-position` div when mouse exits the viewport. Priror to this change, `undefinedHTML=undefined` would set the `ol-mouse-position` container to `''` (empty string) when the mouse exits the viewport.

marcjansen:
I am sorry, I have next to zero time for a review... Probably next week I can spare some time.

Issue #7616, opened 2017-12-16
Rename all _ol_* default exports


tschaub:
The transform to ES modules generated names for imports and exports that start with `_ol_`.  These should be changed to more natural names.  For example, `_ol_View_` can be renamed to `View`.  Some care is needed in doing the renaming to ensure that existing variable and argument names do not shadow the renamed imports or new export variable names.  In general, we should favor short names.  Where we need to disambiguate, we can make names longer.  For example, `_ol_Tile_` can become `Tile` and `_ol_source_Tile_` can become `TileSource`.  Module names *should not* be renamed (this is only about changing identifiers within the modules).

See #7610, #7607, and #7591 for inspiration.

See #7779 for a related ticket about using named exports from utility modules.  This issue is just about renaming the remaining default exports modules that just export a constructor or an enum.  At the time of this writing, there are 17 more of these to be reworked:

- [x] `src/ol/proj/EPSG4326.js` (see #7782)
- [x] `src/ol/proj/EPSG3857.js` (see #7782)
- [x] `src/ol/renderer/webgl/tilelayershader/Locations.js` (see #7788)
- [x] `src/ol/renderer/webgl/defaultmapshader/Locations.js` (see #7788)
- [x] `src/ol/ImageBase.js`  (see #7793)
- [x] `src/ol/webgl/Vertex.js`(see #7797)
- [x] `src/ol/webgl/Shader.js` (see #7787)
- [x] `src/ol/webgl/Fragment.js` (see #7797)
- [x] `src/ol/webgl/Context.js` (see #7787)
- [x] `src/ol/webgl/Buffer.js` (see #7787)
- [x] `src/ol/render/webgl/Immediate.js` (see #7797)
- [x] `src/ol/render/webgl/texturereplay/defaultshader/Locations.js` (see #7788)
- [x] `src/ol/render/webgl/linestringreplay/defaultshader/Locations.js` (see #7788)
- [x] `src/ol/render/webgl/circlereplay/defaultshader/Locations.js` (see #7788)
- [x] `src/ol/render/webgl/polygonreplay/defaultshader/Locations.js` (see #7788)
- [x] `src/ol/Image.js` (see #7793)
- [x] `src/ol/format/GML.js` (see #7793)

Please leave a comment here if you are going to take on some of this work so we can avoid redundant efforts.

tschaub:
I'll take on `src/ol/proj/EPSG4326.js` and `src/ol/proj/EPSG3857.js`.

[...]

PR #7787, opened 2018-02-08
More renaming

fredj:
see #7616

### DIFF
--- a/src/ol/webgl/Buffer.js
+++ b/src/ol/webgl/Buffer.js
@@ -4,12 +4,21 @@
 import _ol_webgl_ from '../webgl.js';
 
 /**
+ * @enum {number}
+ */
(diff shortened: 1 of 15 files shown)


src/ol/webgl/Buffer.js
```diff
@@ -21,35 +30,24 @@ const _ol_webgl_Buffer_ = function(opt_arr, opt_usage) {
    * @private
    * @type {number}
    */
-  this.usage_ = opt_usage !== undefined ?
-    opt_usage : _ol_webgl_Buffer_.Usage_.STATIC_DRAW;
+  this.usage_ = opt_usage !== undefined ? opt_usage : BufferUsage;
```
fredj: must be `BufferUsage.STATIC_DRAW`

PR #7788, opened 2018-02-08
Fix shaders generation

fredj:
Fix shaders generation and add named exports from WebGL shaders


src/ol/render/webgl/PolygonReplay.js
```diff
@@ -873,14 +873,14 @@ WebGLPolygonReplay.prototype.getDeleteResourcesFunction = function(context) {
  */
 WebGLPolygonReplay.prototype.setUpProgram = function(gl, context, size, pixelRatio) {
   // get the program
-  const fragmentShader = _ol_render_webgl_polygonreplay_defaultshader_.fragment;
-  const vertexShader = _ol_render_webgl_polygonreplay_defaultshader_.vertex;
+  const fragmentShader = fragment;
+  const vertexShader = vertex;
   const program = context.getProgram(fragmentShader, vertexShader);
```
tschaub: It looks like the two declarations above could be removed and this could become:
```js
const program = context.getProgram(fragment, vertex);
```

Same in other modules.

src/ol/webgl/shader.mustache
```diff
@@ -1,17 +1,17 @@
...
 // This file is automatically generated, do not edit.
 // Run `make shaders` to generate, and commit the result.
 
-import {DEBUG_WEBGL} from '../../index.js';
-import _ol_webgl_Fragment_ from '../../webgl/Fragment.js';
-import _ol_webgl_Vertex_ from '../../webgl/Vertex.js';
-var {{namespace}} = {};
+import {DEBUG_WEBGL} from '../../../index.js';
+import _ol_webgl_Fragment_ from '../../../webgl/Fragment.js';
+import _ol_webgl_Vertex_ from '../../../webgl/Vertex.js';
```
tschaub: Might be worth renaming these while you're here.

tschaub: I've renamed the default exports in #7797.  But renaming these imports would be easier done in your branch.

src/ol/render/webgl/PolygonReplay.js
fredj: done in 2b6a3aa

src/ol/webgl/shader.mustache
fredj: I've added the renaming to this branch

PR #7789, opened 2018-02-08
Use STATIC_DRAW as default WebGL buffer usage

fredj:
Fixup for  #7787 

See: https://github.com/openlayers/openlayers/pull/7787/commits/62b29003483e969183eeb0249b77b645c5ff5456#r166941451


fredj:
trivial, merging

PR #7791, opened 2018-02-08
Named exports from ol/tilecoord

marcjansen:
See #7779.

Please note that I changed the test `createFromTemplates` in `tileurlfunction.test.js` which used to override the `hash` method with sinon. I have removed most of the test and cannot say whether this is still testing anything useful.

Please review.


marcjansen:
Thanks again, Tim.

PR #7797, opened 2018-02-09
Better default export names for WebGL modules

tschaub:
This renames the remaining default exports from WebGL modules.

The names used when importing for WebGLVertex and WebGLFragment can be changed in the templates - see https://github.com/openlayers/openlayers/pull/7788#discussion_r167105120.

See #7616.


fredj:
WebGLFragment and WebGLVertex renaming are done in #7788

tschaub:
@fredj - rebased.  This now just renames WebGLImmediateRenderer.

PR #7800, opened 2018-02-09
More named exports

tschaub:
This adds more named exports.  See #7779.


tschaub:
Looks like a few tests to fix up...

tschaub:
I see a bad import in the overview map.  Will fix it up tomorrow.

ahocevar:
I'm taking on the overview map import.

PR #7826, opened 2018-02-14
Better code indentation and modules renaming 

fredj:


src/ol/layer/Tile.js
```diff
@@ -53,9 +53,7 @@ inherits(TileLayer, Layer);
  * @api
  */
 TileLayer.prototype.getPreload = function() {
-  return (
-  /** @type {number} */ this.get(_ol_layer_TileProperty_.PRELOAD)
-  );
+  return (/** @type {number} */ this.get(TileProperty.PRELOAD));
```
tschaub: These parens might have gotten messed up by the initial code transform.  Shouldn't this be like:
```js
return /** @type {number} */ (this.get(TileProperty.PRELOAD));
```

(I'll be happy when we get rid of these typecasts altogether.)

src/ol/layer/Tile.js
```diff
@@ -86,9 +84,7 @@ TileLayer.prototype.setPreload = function(preload) {
  * @api
  */
 TileLayer.prototype.getUseInterimTilesOnError = function() {
-  return (
-  /** @type {boolean} */ this.get(_ol_layer_TileProperty_.USE_INTERIM_TILES_ON_ERROR)
-  );
+  return (/** @type {boolean} */ this.get(TileProperty.USE_INTERIM_TILES_ON_ERROR));
```
tschaub: Same here regarding the grouping.

src/ol/layer/VectorTile.js
```diff
@@ -77,9 +75,7 @@ VectorTileLayer.prototype.getPreload = function() {
  * @api
  */
 VectorTileLayer.prototype.getUseInterimTilesOnError = function() {
-  return (
-  /** @type {boolean} */ this.get(_ol_layer_TileProperty_.USE_INTERIM_TILES_ON_ERROR)
-  );
+  return (/** @type {boolean} */ this.get(TileProperty.USE_INTERIM_TILES_ON_ERROR));
```
tschaub: Grouping.

src/ol/layer/Tile.js
fredj: in a8f4348

PR #7834, opened 2018-02-15
Named exports from ol/transform

marcjansen:
See #7779.


marcjansen:
Thanks for the review.

PR #7835, opened 2018-02-15
Add and use a constant for the collapsed CSS class

marcjansen:
This is a small refactoring which introduces `CLASS_COLLAPSED` which is used by two controls.

Please review.


marcjansen:
Thanks for the review.

PR #7863, opened 2018-02-20
Add note about null geom in VectorSource.forEachFeature

### LINKED ISSUE #7862: ol.source.Vector#forEachFeature skips features without geometry
When calling forEachFeature(callback, opt_this), the callback is not called with features that have no geometry.

Code that shows the issue:
https://codepen.io/anon/pen/LQdrVa
`
var source = new ol.source.Vector();
source.addFeature(new ol.Feature());

console.log(source.getFeatures().length); // 1

source.forEachFeature(function (f){
  console.log("forEachFeature"); // won't show up
}); 
`

If it is by design then the docs don''t state it clearly.
Excerpt from the api docs:
https://openlayers.org/en/latest/apidoc/ol.source.Vector.html#forEachFeature
"Iterate through all features on the source, ...."

Tested with v4.6.4

fredj:
This function was not considering features with null geometry.

fixes #7862

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -487,6 +487,7 @@ VectorSource.prototype.clear = function(opt_fast) {
  * Iterate through all features on the source, calling the provided callback
  * with each one.  If the callback returns any "truthy" value, iteration will
  * stop and the function will return the same value.
+ * Note: this function only iterate through the feature that have a defined geometry.
  *
  * @param {function(module:ol/Feature~Feature): T} callback Called with each feature
  *     on the source.  Return a truthy value to stop iteration.


fredj:
@openlayers/core do you have any opinion on this?

ahocevar:
When someone is interested in efficiently iterating over features with geometry, then `forEachFeature` is more efficient because it used the `forEach` function from `rbush`. So it might be better to just document that it's only for iterating over features with geometry, instead of removing it.

fredj:
@ahocevar yes, good point. I'll update the documentation

fredj:
trivial, merging.

PR #7864, opened 2018-02-20
Fix device-orientation example

oterral:
device-orientation example is broken  in production mode because ol.math.toRadians is not exported.


tschaub:
This was a problem with the 4.6.4 release.  See https://github.com/openlayers/openlayers/blob/v4.6.4/src/ol/math.js#L166-L174 and http://openlayers.org/en/v4.6.4/examples/device-orientation.html.  The example was broken with 34428cfea618bb3991d41ec0c7499df9dcfc9f5c.  The master branch is quite far ahead of 4.6.4 at this point, and `toRadians` is correctly exported from the `ol/math` module.

I don't think it is worth issuing a 4.6 patch release to fix this example.  We'll be publishing an updated version of the hosted examples when we're closer to the 5.0 release.

oterral:
ok

PR #7872, opened 2018-02-21
More named export

fredj:


src/ol/style/IconImage.js
```diff
@@ -101,16 +100,14 @@ inherits(IconImage, EventTarget);
  * @param {ol.Color} color Color.
  * @return {ol.style.IconImage} Icon image.
  */
-IconImage.get = function(image, src, size, crossOrigin, imageState,
-  color) {
+export function get(image, src, size, crossOrigin, imageState, color) {
```
tschaub: It looks like the tests [still use `IconImage.get()`](https://github.com/fredj/openlayers/blob/8cf1b88caa9017879c1a42a39a5dfcf54d848cc4/test/spec/ol/style/icon.test.js#L20-L22)

PR #7881, opened 2018-02-23
Better variables scoping

fredj:
And some minor indentation fixes,

see https://github.com/openlayers/openlayers/pull/7697#issuecomment-357170584


tschaub:
I'll merge this now so we can base other work on it.

PR #7904, opened 2018-02-27
Add focus condition, e.g. for wheel zoom

ahocevar:
On pages with a lot of scrollable content and an embedded map, it may be desirable to only have the map prevent the default wheel/trackpad behavior when it is focused.

To make such configuration easier, this pull request introduces a `focus` condition, and shows how to use it in the `mousewheel-zoom` example.


tschaub:
Nice solution @ahocevar.

PR #7908, opened 2018-02-28
ScaleLine fix for non-degree based units

### LINKED ISSUE #7907: Wrong ScaleLine calculation for non-degree based maps with Imperial-like units
When the map is non-degree based (such as a Zoomify scanned document) the pointResolution calculation is wrong in ScaleLine.js, when using Imperial-like units.

The problem is with this lines:
https://github.com/openlayers/openlayers/blob/85d4642450a1f9bebb0e0280fd20dcf78cfca5cb/src/ol/control/ScaleLine.js#L182-L184

The correct behaviour should be to always run
```
pointResolution *= projection.getMetersPerUnit();
```
when `projection.getUnits() != Units.DEGREES`.

That is, the correct fix is to remove 
```
&& units == ScaleLineUnits.METRIC
``` 

from the condition.

This is only happening on non-degree based maps when using Imperial-like units, which must be quite a niche.

If needed, I can make an example code, but it should be logical to see why this is the case. I can also submit a PR.

hyperknot:
I'm thinking of possibly using `[FEET, METERS, USFEET]` instead for `projection.getUnits()` test, it's not yet clear which is the right solution. 

Fixes #7907.

### DIFF
--- a/src/ol/control/ScaleLine.js
+++ b/src/ol/control/ScaleLine.js
@@ -191,7 +191,8 @@ ScaleLine.prototype.updateElement_ = function() {
     Units.METERS;
   let pointResolution =
       getPointResolution(projection, viewState.resolution, center, pointResolutionUnits);
-  if (projection.getUnits() != Units.DEGREES && units == ScaleLineUnits.METRIC) {
+  if (projection.getUnits() != Units.DEGREES && projection.getMetersPerUnit()
+    && pointResolutionUnits == Units.METERS) {
     pointResolution *= projection.getMetersPerUnit();
   }
 
--- a/test/spec/ol/control/scaleline.test.js
+++ b/test/spec/ol/control/scaleline.test.js
@@ -289,7 +289,7 @@ describe('ol.control.ScaleLine', function() {
       ctrl.setUnits('nautical');
       expect(ctrl.element_.innerText).to.be('0.05 nm');
 
-      ctrl.setUnits('imperial');
+      ctrl.setUnits('us');
       expect(ctrl.element_.innerText).to.be('500 ft');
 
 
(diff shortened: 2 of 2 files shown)


src/ol/control/ScaleLine.js
```diff
@@ -179,7 +179,7 @@ ScaleLine.prototype.updateElement_ = function() {
     Units.METERS;
   let pointResolution =
       getPointResolution(projection, viewState.resolution, center, pointResolutionUnits);
-  if (projection.getUnits() != Units.DEGREES && units == ScaleLineUnits.METRIC) {
+  if (projection.getUnits() != Units.DEGREES) {
```
ahocevar: You'll probably need
```js
if (projection.getUnits() != Units.DEGREES && units != ScaleLineUnits.DEGREES) {
```

hyperknot: I think the `projection.getUnits()` 
part needs to be 
```
includes([Units.FEET, Units.METERS, Units.USFEET], projection.getUnits())
```

and `pointResolutionUnits == Units.METERS` or what you wrote (maybe this one is better documenting that we are converting from meters).

ahocevar: Please add a test case that shows what you're trying to achieve. It seems to me you're confusing projection units and scale line units. Also please reference the issue that this pull request is supposed to fix, for future reference.

hyperknot:
OK, final solution:
1. Instead of hard-coded array, let's use `projection.getMetersPerUnit()` (!= undefined), as that's exactly the case when we can multiply by it.
2. Except for DEGREES, like it was originally.
3. Use `pointResolutionUnits == METERS` which is exactly the same condition as the one you suggested.

hyperknot:
I'll add a test case, but need to familiarise myself with the code.

hyperknot:
@ahocevar I've added the tests for all units, and also included an imperial unit test for the zoom levels loop (unrelated to this).

PR #7915, opened 2018-03-01
Untie feature loader and XML code

gberaudo:
Remove the dependency on `xml.js` from the feature loader.
This change makes it possible to use the featureloader in a web worker, to some extent.


tschaub:
Thanks @gberaudo.

PR #7918, opened 2018-03-01
Move the nullFunction to ol/functions

fredj:
And rename it to `UNDEFINED`.

### DIFF
--- a/src/ol/Disposable.js
+++ b/src/ol/Disposable.js
@@ -1,7 +1,7 @@
 /**
  * @module ol/Disposable
  */
-import {nullFunction} from './index.js';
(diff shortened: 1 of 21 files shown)


fredj:
Thanks for the review

marcjansen:
I find the new name as confusing as the old one. Can we call it `NO_OP`... or is this not better?

fredj:
@gberaudo suggested the same to me 10 minutes ago :-)
Why not call it `NOOP` ? (jQuery, angular and lodash all have a `noop` function)

marcjansen:
noop works as well.

ahocevar:
VOID?

PR #7614, opened 2017-12-16
remove opt_this from api

simonseyock:
This PR removes most optional this arguments from api methods (See #7284).
Only the thisArg in forEachLayerAtPixel is not removed in this PR but in #7613.

PR #7930, opened 2018-03-05
Use listen function instead of on

fredj:
In #7614, the `opt_this` argument was removed from `Observable#on`, `#once` and `#un` but internal code was not adapted.


fredj:
I think the `typecheck` npm script should have detected this; once we have path types everywhere we will need to make the compiler checks more strict.

PR #7944, opened 2018-03-07
Make wrap work with points in the vector-labels example

### LINKED ISSUE #7941: Label issue with point feature on version 4.6.4
After upgrading my web app to 4.6.4, only some of my point features were being labeled. In researching the problem, I revisited the [Vector Labels example page](https://openlayers.org/en/latest/examples/vector-labels.html?mode=advanced). When you change the text setting to 'wrap' for the points, the points and the labels disappears.  Chrome reports the following error on it's console tab:

vector-labels.js:68 Uncaught TypeError: Cannot read property 'value' of undefined
    at getText (vector-labels.js:68)
    at createTextStyle (vector-labels.js:103)
    at pointStyleFunction (vector-labels.js:167)
    at hk.<anonymous> (ol.js:274)
    at hk.k.$c (ol.js:274)
    at kj.bh (ol.js:205)
    at K.k.pq (ol.js:135)
    at K.<anonymous> (ol.js:120)

I use the stringDivider function that is in this example (I copied and pasted it into my web app).  What's really weird: all of my labels fail to draw if I do NOT use stringDivider on my string.  I tried truncating my string using` text.substring(0,8)` but none of them appear.  

I don't display any points, I only need the labels.  Here's an example of the geoJson that's returned by for these points: http://rainpursuit.org/huc_points.php?bbox=-9499479.219739275,4198609.526676447,-9382759.877547814,4265721.237510831

ahocevar:
Fixes #7941.

### DIFF
--- a/examples/vector-labels.js
+++ b/examples/vector-labels.js
@@ -78,7 +78,7 @@ const getText = function(feature, resolution, dom) {
     text = '';
   } else if (type == 'shorten') {
     text = text.trunc(12);
-  } else if (type == 'wrap' && dom.placement.value != 'line') {
+  } else if (type == 'wrap' && (!dom.placement || dom.placement.value != 'line')) {
     text = stringDivider(text, 16, '\n');
   }
 


venugopal-rao:
arrange text to the orientation of the line help me

PR #7948, opened 2018-03-07
Better type definition

fredj:
Fixes the `Invalid type(s) for operator IN.` warnings from the closure compiler.

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -126,15 +126,13 @@ export function reverseSubArray(arr, begin, end) {
 
 /**
  * @param {Array.<VALUE>} arr The array to modify.
- * @param {Array.<VALUE>|VALUE} data The elements or arrays of elements
- *     to add to arr.
+ * @param {!Array.<VALUE>|VALUE} data The elements or arrays of elements to add to arr.
  * @template VALUE
  */
 export function extend(arr, data) {
-  let i;
   const extension = Array.isArray(data) ? data : [data];
   const length = extension.length;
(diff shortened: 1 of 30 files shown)


src/ol/array.js
```diff
@@ -126,15 +126,13 @@ export function reverseSubArray(arr, begin, end) {
...
  * @param {Array.<VALUE>} arr The array to modify.
- * @param {Array.<VALUE>|VALUE} data The elements or arrays of elements
- *     to add to arr.
+ * @param {Array.<VALUE>|VALUE} data The elements or arrays of elements to add to arr.
  * @template VALUE
  */
 export function extend(arr, data) {
-  let i;
-  const extension = Array.isArray(data) ? data : [data];
+  const extension = /** @ type {Array.<VALUE> */ (Array.isArray(data) ? data : [data]);
```
ahocevar: Is this needed? And if so, shouldn't it be `@type` instead of `@ type`?

fredj: Yes, that's an error. Instead of a type cast, the warning can be fixed by changing the `data` type to `{!Array.<VALUE>|VALUE}`

The compiler warning is:
```
./build/src-closure/ol/array.js:136: WARNING - The right side in the assignment is not a subtype of the left side.
Expected : VALUE
Found    : (VALUE|null)
More details:
The found type is a union that includes an unexpected type: null
    arr[arr.length] = extension[i];
```

fredj:
Thanks for the review

PR #7960, opened 2018-03-09
Fix memory leak in CanvasImageRenderer

### LINKED ISSUE #7954: Memory leak when rendering a Vector Layer with renderMode: 'image'
When rendering a VectorLayer, CanvasImageRenderer creates a CanvasVectorRenderer.
CanvasVectorRenderers add an event listener to ol/render/canvas/labelCache.
CanvasImageRenderers do not dispose of their CanvasVectorRenderer, leaving that listener, keeping the whole layer in memory indefinitely.
In addition, due to duplicate plugins being added to layerRendererPlugins, CanvasImageRenderer actually creates two VectorImageRenderers, leaving another hanging listener.

CodePen: https://codepen.io/erinc-petrode/pen/PRobvV

- [x] I am submitting a bug or feature request, not a usage question.
- [x] I have verified that the issue is present in the latest version of OpenLayers.
- [x] If reporting a bug, I have created a [CodePen] that shows the issue.

erinc-petrode:
...when rendering a VectorLayer with renderMode: 'image'.
Added tests to CanvasImageLayerRenderer for this issue.

Fixes #7954

### DIFF
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -80,10 +80,22 @@ CanvasImageLayerRenderer['create'] = function(mapRenderer, layer) {
       const candidate = /** @type {Object.<string, Function>} */ (candidates[i]);
       if (candidate !== CanvasImageLayerRenderer && candidate['handles'](RendererType.CANVAS, layer)) {
         renderer.setVectorRenderer(candidate['create'](mapRenderer, layer));
+        break;
       }
     }
   }
   return renderer;
+};
+
+
+/**
+ * @inheritDoc
+ */
+CanvasImageLayerRenderer.prototype.disposeInternal = function() {
+  if (this.vectorRenderer_) {
+    this.vectorRenderer_.dispose();
+  }
+  IntermediateCanvasRenderer.prototype.disposeInternal.call(this);
 };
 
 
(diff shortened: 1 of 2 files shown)


src/ol/renderer/canvas/ImageLayer.js
```diff
@@ -80,13 +80,25 @@ CanvasImageLayerRenderer['create'] = function(mapRenderer, layer) {
       const candidate = /** @type {Object.<string, Function>} */ (candidates[i]);
       if (candidate !== CanvasImageLayerRenderer && candidate['handles'](RendererType.CANVAS, layer)) {
         renderer.setVectorRenderer(candidate['create'](mapRenderer, layer));
+        break;
```
tschaub: This looks like a good fix, but I'm curious when you are seeing more than one appropriate renderer being registered here.  I'm wondering if there are more bugs.

erinc-petrode: The CanvasRenderer plugins are added in both Map.js and CanvasMap.js.

https://github.com/openlayers/openlayers/blob/ba17527d29a765460a295c8f0709e5115c31951a/src/ol/Map.js#L22-L28

https://github.com/openlayers/openlayers/blob/ba17527d29a765460a295c8f0709e5115c31951a/src/ol/CanvasMap.js#L18-L24

To my knowledge, the only other location where that could cause an issue was already handling it with a break.

https://github.com/openlayers/openlayers/blob/ba17527d29a765460a295c8f0709e5115c31951a/src/ol/renderer/Map.js#L210-L216

tschaub: Makes sense (though an application should not be importing both `Map` and `CanvasMap`).

tschaub:
Thanks for the fixes @erinc-petrode.

src/ol/renderer/canvas/ImageLayer.js
erinc-petrode: Ah, fair. That is an issue in the most recent released build, and I hadn't considered that it would be fixed with the move to es6 import/export.

Issue #7947, opened 2018-03-07
Get rid of externs/olx.js and src/ol/typedefs.js


ahocevar:
We have everything in place now for moving type definitions from `externs/olx.js` to other places. The new typedefs for copying/pasting into other places (class modules for classes, `src/ol/types.js` for things like FrameState) can be found in the temporary `externs/xol.js` file.

An example pull request for part of the work is #7946, which is based on #7609. Additional pull requests should remove types from `externs/olx.js` as they are moved to other places.

The desired result is outlined in #7838, which also involves getting rid of `src/ol/typedefs.js`, whose types are redistributed to the place where they are used if there is only one and its internal only, or to `src/ol/types.js`.

This issue keeps track of progress. Here are the namespaces to tick away:

- [x] #7953 `ol`
- [x] #7965 `ol/control`
- [x] #7953 `ol/events`
- [x] #7983 `ol/extent`
- [x] #7968 `ol/format`
- [x] #7971 `ol/geom`
- [x] #7989 `ol/interaction`
- [x] #8016 #8025 `ol/layer`
- [x] `ol/pointer`
- [x] #7971 `ol/proj`
- [x] `ol/render`
- [x] `ol/renderer`
- [x] #8084 `ol/reproj`
- [x] #8032 #8041 `ol/source`
- [x] `ol/structs`
- [x] #8076 `ol/style`
- [x] #8075 `ol/tilegrid`
- [x] `ol/vec`
- [x] `ol/webgl`

ahocevar:
I'm taking on `ol`.

ahocevar:
I'm also taking on `ol/events`.

tschaub:
I'll do `ol/control`.

tschaub:
I'll take `ol/format` next.

ahocevar:
Working on `ol/extent` now.

ahocevar:
Now taking on `ol/interaction`.

romanzoller:
I will have a look at `ol/layer`.

Thanks @ahocevar for the intro to the topic and greetings from the [OSGeo Code Sprint](https://wiki.osgeo.org/wiki/OSGeo_Code_Sprint_2018)

![basecamp](https://user-images.githubusercontent.com/23193956/37781838-d6473f1c-2df1-11e8-8164-1886bdae53f1.jpeg)

romanzoller:
[...]

PR #7965, opened 2018-03-11
Redistribute types from externs/olx.js

tschaub:
See #7947.


tschaub:
@ahocevar - the `publish.js` script is failing here https://github.com/openlayers/openlayers/blob/e3409b1f7805129d0f328c28ea2240972817be15/config/jsdoc/info/publish.js#L70-L77 when parsing a member expression (e.g. `olx.control`) if there are no `typedef`s already parsed.  This starts happening at 57d135e9f66977a9f91c93400bc7c2b3569dc9b5 after removing all the `typedef`s above `olx.control`.  I'll add block that throws an error with more detail in this case and remove the `olx.control` member expression, but I think the block above might need more reworking (it isn't clear to me why it only reworks the first type of the last typedef, so I won't try to change the logic).

ahocevar:
Not sure about the logic in `publish.js`, I do not remember that code. Will the current code there be a problem when we‘re done with the olx removal?

tschaub:
I think we'll be able to simplify `publish.js` when we are no longer documenting stuff in `externs` (that block above came with f0242a76c45fe483900f46f06f20d5343923b62b).

PR #7975, opened 2018-03-12
Mark optional property with square brackets

fredj:


tschaub:
Will this also require work on the tools that create the typedefs for the Compiler?

fredj:
I think it's only for the api doc because the type are already optional (with the `undefined` type) but that's only a supposition.

ahocevar:
> Will this also require work on the tools that create the typedefs for the Compiler?

The babel-plugin-jsdoc-closure supports this already.

src/ol/View.js
```diff
@@ -146,9 +146,8 @@ import Units from './proj/Units.js';
  * the animation.
  * @property {ol.Coordinate|undefined} anchor Optional anchor to remained fixed
  * during a rotation or resolution animation.
- * @property {number|undefined} duration The duration of the animation in milliseconds
- * (defaults to `1000`).
- * @property {undefined|function(number):number} easing The easing function used
+ * @property {number|undefined} [duration=1000] The duration of the animation in milliseconds.
+ * @property {undefined|function(number):number} [easing] The easing function used
```
ahocevar: Oh sorry for the inaccurate review. You should have removed the `undefined` in all places where you set the type in square parentheses. The babel-plugin-closure-jsdoc will resolve that to the correct type.

ahocevar:
@fredj Can you please create a follow-up pull request to address https://github.com/openlayers/openlayers/pull/7975#discussion_r174076795? Thanks.

src/ol/View.js
fredj: see #7979

Issue #7422, opened 2017-11-06
True raster support


GaborFarkas:
Hi,

Recently, I've been working on a raster module for OL. I created a basis, which for now only has a raster storage mechanism, and a simple raster source. I need some feedback on whether you can imagine this work as part of the core lib. As your decision will affect how I will code the rendering part, I would like to get some feedback ASAP.

You can reach the aforementioned branch [here](https://github.com/GaborFarkas/ol3/tree/raster_base).

GaborFarkas:
Remaining work (for now):
 - WCS support
 - GeoTIFF support
 - A dedicated renderer for raster layers

Integrating this module in the core lib has of course some pros and cons.

Pros:
 - More functionality, therefore potential
 - By disabling raster support, you can have the old, smaller lib
 - Hookless and hackless raster renderer, which is definitely faster than written as an external module
 - We can close every WCS related issues once and for all

Cons:
 - Noticeably larger precompiled lib (if both raster and geotiff support is enabled by default)
 - Plus one dependency to manage (geotiff.js)

GaborFarkas:
Currently it seems like, integrating geotiff.js as a module (`ol.ext`) is not possible, as it has an unnecessary escape, which cannot be suppressed with the `@suppress` tag. I'm starting to get convinced, this work should be written as a separate module (like the layer switcher), however the renderer part bothers me. What do you think about the following scenarios?
 - Integrating core raster support with a renderer in OL. All of the raster classes, plus a basic API raster source, from which users can create their own classes.
 - Integrating the current work, but for the GeoTIFF source requiring geotiff.js to be loaded separately, just like proj4js.
 - Modifying the code in a way, users can register their own renderers for their own layer types.

GaborFarkas:
Closing as there is a PR for this.

PR #7978, opened 2018-03-13
[WIP] Rendering raw coverage data

GaborFarkas:

- [ ] Existing tests pass for me locally & I have added or updated tests for new or changed functionality.

Hi,

As I did not get any feedback for my idea about this feature, and OL became modular, therefore now it can bear more features, I created the proposed coverage renderer from #7422. It can render ArcInfo ASCII Grids and GeoTIFF coverages as regular, texture-based rasters, and as vector-based coverages. Coverages are of course slower, but they have a lot of advantages. For further reading about this concept, there is an [article](https://gaborfarkas.github.io/papers/alternative_raster_display_model.pdf) on the topic. There are also some demos showing the concept of the coverage model [here](https://gaborfarkas.github.io/coverage_demo/1), [here](https://gaborfarkas.github.io/coverage_demo/2), and [here](https://gaborfarkas.github.io/coverage_demo/3).

If you consider this PR as a merge candidate, I will create proper unit tests.


GaborFarkas:
@openlayers/core I know you are in a serious work towards 5.0, but could you give a preliminary feedback on considering merging this work later?

tschaub:
Hello @GaborFarkas.  Thanks for your work on potential rendering improvements.

Our current focus toward 5.0 is in making the library more developer friendly.  So we're working on making the development environment simpler, the documentation better, the library less awkward to use, etc.

Part of that work is making it so people do not bundle together a lot of unnecessary rendering code together with their application (e.g. all the WebGL rendering code if they are only actually using the Canvas renderer).  We have decided to make renderers be explicit dependencies rather than supporting the `renderers` option or continuing with the idea of pluggable renderers.

So, in the 5.0 release, the map will depend on a single renderer.  Since the WebGL renderer is not a suitable renderer for most applications, that single renderer will be the Canvas renderer.  If the WebGL renderer gets to the point where it surpasses the functionality of the Canvas renderer, we can make the decision to switch the single map renderer.  Alternatively, we may decide to take a different approach where layers depend on a renderer, and we can have a mixed rendering approach.  But the dependency on the renderer(s) will still be explicit (for example, it could be that in the future that one layer type depends on a WebGL-based renderer and another depends on a Canvas-based renderer).

For these reasons, we don't want to take on additional experimental WebGL rendering code at this point.  I really like the idea of a high performance raster layer renderer, but I think this would make more sense after we have a better story for mixing renderer types based on the layer.  In addition, we would want to have some discussion about the functionality we want to expose to the user (e.g. in deciding how those rasters are rendered).  If we don't go with a per-layer rendering approach, then the WebGL renderer needs significant improvement before it could be a replacement for the current map renderer.  So rather than broadening the scope of the WebGL renderer, I think it would make more sense to focus on improving the current capabilities.

While it is painful to close a pull request that represents a lot of hard work on the part of the author, I think that is the most honest thing to do at this point.  I hope what I've said above makes some sense.

gberaudo:
Hi @GaborFarkas, I had a very quick look to your PR and I think it is pretty cool.

There was a talk at FOSS4G 2016 about rendering geotiffs in both OpenLayers and Cesium, did you know about it?
https://eox.at/2016/02/visualizing-raster-data/
https://ftp.gwdg.de/pub/misc/openstreetmap/FOSS4G-2016/foss4g-2016-1219-geotiff_js_and_plotty_js_-_visualizing_scientific_raster_data_in_the_browser-hd.webm

The OpenLayers team is focused on the ES6 modules migration and the major OL5 release.
In addition, there are some reflections about moving non-core code out of the library. It is not clear yet, but whatever the decisions, I believe there will be / should be a place to contribute nice code as yours.

Regarding the renderers, it was discussed last week that:
- we want to be able to mix renderers inside a Map, some layers could use Canvas while others could use WebGL, or other libs;
- there will be no fundamental change in the renderers in OL5;
- there is no obligation of feature parity between the renderers.

For information, we also discussed WebGL vector tiles rendering.

I think OpenLayers would greatly benefit from a plugin ecosystem. [Leaflet has one](https://github.com/Leaflet/Leaflet/blob/master/PLUGIN-GUIDE.md), and it is a big part of its success. @tschaub, until we have one and code is moved out of the lib, couldn't we review / merge PRs like this one?

GaborFarkas:
While I don't necessarily agree on these 5.0 rendering goals (but as far as I can understand from @gberaudo's comment this already changed somewhat), I understand your point @tschaub. And of course, if this is the decision of the core dev team, I accept it. Thanks for considering the PR. Just as a sidenote, as it does not really weights much, the whole thing has a finely working canvas part. It's just simply slow at the moment, as the canvas engine cannot handle as many polygons properly.

I'm glad you like the new features, @gberaudo. Thanks for the update about the team's view on the future of OL renderers. A plugin ecosystem would be great, it proved itself so many times in the FOSS world (just think about GRASS or QGIS if Leaflet is not enough). If feature parity won't be a requirement, I wouldn't force the canvas coverage renderer, although using pyramids might give acceptable frame rates. About those presentations, I pretty well know those work, and they are the basis for some of this code. In the GeoTIFF source, that particular GeoTIFF library is assumed/used.

I'm leaving this branch here for now, as I can feel, you did not fully agree on its fate yet. However, whether it gets merged or not, allowing users to create their own renderers with their own layer types would be a great plus.

If the final decision will be no, then I will create another PR with some of the more general commits (e.g. decoupling the triangulation process from the WebGL polygon replay).

Issue #7587, opened 2017-12-14
Remove chrome and firefox version in saucelab config


fredj:
Latest Chrome and Firefox currently do not work in saucelabs, so version 62.0 (chrome) and 58 (firefox) were set in `test/karma.config.js`.

Remove these version settings when latest versions work again.

PR #7983, opened 2018-03-14
Use path type instead of ol.Extent

ahocevar:
With this pull request, `ol.Extent` is replaced with its path type counterpart throughout the library. See #7947.


ahocevar:
With firefox fixed to version 58, tests pass again. See #7587.

PR #7985, opened 2018-03-15
Use path type instead of ol.tilegrid

fredj:
See #7947.

### DIFF
--- a/src/ol/tilegrid.js
+++ b/src/ol/tilegrid.js
@@ -53,15 +53,13 @@ export function wrapX(tileGrid, tileCoord, projection) {
  *     DEFAULT_MAX_ZOOM).
  * @param {number|module:ol/size~Size=} opt_tileSize Tile size (default uses
  *     DEFAULT_TILE_SIZE).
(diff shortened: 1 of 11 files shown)


fredj:
changed. I've also added a commit that fixes the name for extent Corner Relationship

fredj:
Thanks for the review

src/ol/tilegrid.js
```diff
@@ -129,14 +127,12 @@ function resolutionsFromExtent(extent, opt_maxZoom, opt_tileSize) {
  *     DEFAULT_MAX_ZOOM).
  * @param {number|module:ol/size~Size=} opt_tileSize Tile size (default uses
  *     DEFAULT_TILE_SIZE).
- * @param {module:ol/extent~Extent.Corner=} opt_corner Extent corner (default is
- *     module:ol/extent~Extent.Corner.BOTTOM_LEFT).
+ * @param {module:ol/extent/Corner~Corner=} opt_corner Extent corner (default is `'top-left'`).
```
fredj: note: the comment was wrong, the default value is really `'top-left'` (see `createForExtent` function)

PR #7994, opened 2018-03-18
Explicit renderer dependencies

tschaub:
In 4.x releases, `ol/Map` depended on both the Canvas and WebGL renderers.  The result is that applications include a ton of unused code.  In the 5.x releases, `ol/Map` will use the Canvas renderer.  To experiment with the WebGL renderer, people can `import Map from 'ol/WebGLMap'`.

There are two potential futures for renderers in the library:
 1. We continue with the single map renderer.  The Canvas renderer could be replaced by the WebGL renderer if that ever becomes a more capable renderer for all layer types.
 2. Instead of a single map renderer, layers depend on a renderer, and the map can compose different rendering approaches (e.g. composing a Canvas based layer with a WebGL based layer).

Whichever alternative we choose, the map and layers should have explicit dependencies on the renderer.  This means removing the `renderers` option from the API.


tschaub:
This is now ready for review.

A map now depends on a map renderer.  The next logical step would be for a layer to depend on a layer renderer (then all dependencies would be explicit).  For now, the `Map` and `WebGLMap` constructors register all possible layer renderers with their map renderer.

If someone wants a build with a single layer renderer, they would have to create a custom map constructor.
```js
import PluggableMap from 'ol/PluggableMap';
import CanvasMapRenderer from 'ol/renderer/canvas/Map';
import CanvasTileLayerRenderer from 'ol/renderer/canvas/TileLayer';
import {inherits} from 'ol';

function CustomMap(options) {
  // set default controls and interactions here
  PluggableMap.call(this, options);
}
inherits(CustomMap, PluggableMap);

CustomMap.prototype.createRenderer = function() {
  const renderer = new CanvasMapRenderer(this);
  renderer.registerLayerRenderers([CanvasTileLayerRenderer]);
  return renderer;
};
```

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,39 @@
 
 ### Next release
 
+#### Removal of the renderer option for maps
+
+The `renderer` option has been removed from the `Map` constructor.  The purpose of this change is to avoid bundling code in your application that you do not need.  Previously, code for both the Canvas and WebGL renderers was included in all applications - even though most people only use one renderer.  The `Map` constructor now gives you a Canvas (2D) based renderer now.  If you want to try the WebGL renderer, you can import the constructor from `ol/WebGLMap`.
```
marcjansen: duplicated word 'now' in 

'The `Map` constructor now gives you a Canvas (2D) based renderer now'

tschaub: Thanks for the catch @marcjansen.  Fixed in c7a10fc7b68c85f9ce26c696dbbed3ea0bd880a5.

PR #8003, opened 2018-03-20
Add an example showing Chaikins smoothing algorithm

marcjansen:
This example uses the npm package [`chaikin-smooth`](https://www.npmjs.com/package/chaikin-smooth) which implements [Chaikins algorithm](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html)
  to smooth drawn lines.

  Start by drawing on the map. Once you finish a drawing, the feature's geometry will be smoothed
  as configured via the form elements.

image

/cc @chrismayer & @ahocevar & @bartvde

### DIFF
--- a/examples/chaikin.html
+++ b/examples/chaikin.html
@@ -3,7 +3,7 @@ layout: example.html
 title: Smoothing lines using Chaikins algorithm
 shortdesc: This uses Chaikins algorithm to smooth drawn lines.
 docs: >
-  This example uses the npm package [`chaikin-smooth`](https://www.npmjs.com/package/chaikin-smooth) which 
+  This example uses the npm package [`chaikin-smooth`](https://www.npmjs.com/package/chaikin-smooth) which
   implements [Chaikins algorithm](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html)
   to smooth drawn lines.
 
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -0,0 +1,9654 @@
+{
```
sbrunner: Shouldn't we ignore this file?

marcjansen: Correct, thanks. PR updated.

sbrunner: :-)

examples/chaikin.html
```diff
@@ -0,0 +1,26 @@
...
+
+  Start by drawing on the map. Once you finish a drawing, the feature's geometry will be smoothed
+  as configured via the form elements.
+tags: "smooth, smoothing, chaikin"
+---
+<div id="map" class="map"></div>
+<form class="form-inline">
+  <label for="type">Geometry type &nbsp;</label>
+  <select id="type">
+    <option value="LineString">LineString</option>
```
fredj: if only LineString is available, the select should be removed (or hidden)

examples/chaikin.js
```diff
@@ -0,0 +1,74 @@
...
+});
+
+const typeSelect = document.getElementById('type');
+const shallSmoothen = document.getElementById('shall-smoothen');
+const numIterations = document.getElementById('iterations');
+
+let draw; // global so we can remove it later
+function addInteraction() {
+  const value = typeSelect.value;
+  if (value !== 'None') {
```
fredj: The only possible value is `LineString`, never `None`

examples/chaikin.html
marcjansen: True. There is an open PR to get support for Polygons. But you are probably right. Will change.

examples/chaikin.js
marcjansen: Will change.

marcjansen:
@fredj udpated and rebased as suggested. TIA for another look.

marcjansen:
Thanks for the review.

Issue #2907, opened 2014-11-03
layer ids


bartvde:
So in OpenLayers 2 we had layer.id which made it easy for e.g. a GeoExt LayerStore to filter duplicates. In ol3 we have ids by use of `goog.getUid` but we keep the ids internal.

Just wanted to re-assess if there is no way to get a public id, and we don't want to make any changes that facilitate this kind of thing.

Thoughts?

fredj:
Closing due to lack of activity.

PR #8020, opened 2018-03-23
Update Property.js

zhangfytech:
add a property named 'ID' to identify layer,so it can be found easily from ol.Map object

- [ ] It contains one or more small, incremental, logically separate commits, with no merge commits.
- [ ] I have used clear commit messages.
- [ ] Existing tests pass for me locally & I have added or updated tests for new or changed functionality.

### DIFF
--- a/src/ol/layer/Property.js
+++ b/src/ol/layer/Property.js
@@ -12,5 +12,6 @@ export default {
   Z_INDEX: 'zIndex',
   MAX_RESOLUTION: 'maxResolution',
   MIN_RESOLUTION: 'minResolution',
-  SOURCE: 'source'
+  SOURCE: 'source',
+  ID:'id'
 };


fredj:
See also #2907

fredj:
You can already do that; any property set in the options is set as a property of the layer.
Example:
```js
const map = new Map({
  layers: [
    new TileLayer({
      // set the layer id:
      id: 'foo',
      source: new OSM()
    })
  ],
  target: 'map',
  view: new View({
    center: [0, 0],
    zoom: 2
  })
});

// find the layer with id === 'foo'
map.getLayers().getArray().find(function(layer) {
  return layer.get('id') === 'foo'
});
```
No need to use `id` as key; only value is possible (except the ones defined in ol/layer/Property)

PR #8021, opened 2018-03-23
Move olx.ViewState to ol/View

fredj:


fredj:
yes, `State` is nicer

Issue #7838, opened 2018-02-15
Get rid of the olx namespace and simplify typedefs


ahocevar:
By using the latest version of [babel-plugin-jsdoc-closure](https://npmjs.com/babel-plugin-jsdoc-closure), we can now introduce JSDoc path types with [module identifiers](http://usejsdoc.org/howto-commonjs-modules.html#module-identifiers) and JSDoc [typedefs](http://usejsdoc.org/tags-typedef.html).

The work we have ahead of us here will wrap up #7608, and simplifies the approach started in #7609.

### Option object literals

We no longer need to duplicate option object literals for constructors and methods. With JSDoc typedefs, we can define and document all options, and can inline them later into the constructor documentation with a simple [JSDoc plugin](https://github.com/openlayers/closure-es-modules/blob/master/jsdoc-inline-options.js).

For options that are only used for a constructor (and its subclasses, like e.g. for `PluggableMap` and `Map`) or a method (and its subclasses, like e.g. `FeatureFormat#readFeatures`), the typedef can be placed in the file that implements the topmost class in the hierarchy. The format is straightforward:

Snippet from `src/ol/PluggableMap.js`:
```js
/**
 * @module ol/PluggableMap
 */

/**
 * @typedef {Object} MapOptions
 * @property {ol.Collection.<ol.control.Control>|Array.<ol.control.Control>|undefined} controls
 * Controls initially added to the map. If not specified,
 * {@link ol.control.defaults ol.control.defaults()} is used.
 * @property {number} [pixelRatio=window.devicePixelRatio] The ratio between physical
 * pixels and device-independent pixels (dips) on the device.
 * (...)
 */

/**
 * @constructor
 * @param {module:ol/PluggableMap~MapOptions} options Map options.
 */
const PluggableMap = function(options) {
  // ...
}
```

Also note that JSDoc and the plugin can handle optional properties and defaults. See the `pixelRatio` property above. Its type will be transformed to `(undefined|number)` by the plugin, and JSDoc will list the default value `window.devicePixelRatio`.

Other typedefs, like `olx.FrameState` or those from `src/ol/typedefs.js`, can go to appropriate places (e.g. `src/ol/PluggableMap.js` for `olx.FrameState`), using the same syntax.

### Referencing types in other files

When using types in different files, they are referenced using module types.

Snippet from `src/ol/Map.js`:
```js
/**
 * @constructor
 * @extends {module:ol/PluggableMap}
[...]

PR #8025, opened 2018-03-23
Move olx.layer.* to ol/layer/*

romanzoller:
Continuation of #8016, move the remaining `ol.layer.*` externs to `ol/layer/*`, see #7947 for an overview of this ongoing work.

Last commit: Fix setting of the `useInterimTilesOnError` in `ol/layer/VectorTile`.

### Notes

All default values were moved from documentation text to the JSDoc syntax `[opacity=1]`, which will be supported for documentation generation according to https://github.com/openlayers/openlayers/issues/7838.

Sometimes this would lead to minor changes in the documentation text, e.g.

> Preload. Load low-resolution tiles up to `preload` levels. By default `preload` is `0`, which means no preloading.

was transformed to

> Preload. Load low-resolution tiles up to `preload` levels. `0` means no preloading.

### Questions

Other things I stumbled over while working on this, not necessarily to be addressed in this PR, but I would by happy to get some feedback from someone more familiar with ol:

- The `options` Object is optional everywhere except for `ol/layer/Base` and `ol/layer/Layer`, but neither of them have any mandatory options-attributes. Should the `options` Object be optional everywhere?
- Is there a preferred order for attributes in the `Options` typedefs? I was currently following more or less the existing order and generally putting all inherited options first for easy copy-pasting.
- Minor: Is there a preference for putting one or two spaces after periods in documentation text?


romanzoller:
Closing and re-opening to trigger new CI build, tests were passing locally...

ahocevar:
@romanzoller, answering your other questions:

> * The options Object is optional everywhere except for ol/layer/Base and ol/layer/Layer, but neither of them have any mandatory options-attributes. Should the options Object be optional everywhere?

Good question. Since both `layer/Base` and `layer/Layer` are base classes that are not used for anything but extending them, it does not matter. For any class that can be constructed in an application, `options` should be optional when an instance can be created without any option.

> * Is there a preferred order for attributes in the Options typedefs? I was currently following more or less the existing order and generally putting all inherited options first for easy copy-pasting.

We should sort options by importance, i.e. those that are mandatory or recommended first, and those for special configuration cases last. This does not need to happen as part of this current effort, but should be done eventually.

ahocevar:
Thanks fo working on this, and for the good summary of the changes you made. I'm reviewing in detail now.

romanzoller:
Thanks for the quick review and for the clarifications @marcjansen and @ahocevar!

PR #8026, opened 2018-03-25
[doc] dispatchChangeEvent -> changed

nyurik:
I think this is the only place in code that documents dispatchChangeEvent, but I think it means "changed" instead.  I also created  https://github.com/openlayers/openlayers/compare/master...nyurik:patch-1#diff-946befd18169a7196da938115c099af4  but I'm not sure if that's an autogenerated file or not.

P.S. It would be great to make it into a doc link instead of simply text.  Not sure what the doc semantics would be.


fredj:
Thanks for the pull request; you can create a link by using `@link`:
`{@link module:ol/Observable~Observable#changed}`

nyurik:
@fredj thx, fixed.

fredj:
thanks @nyurik

PR #8031, opened 2018-03-26
Export all interaction events

fredj:
Some of the events emitted by the interaction were not exported.


fredj:
I'm using `ol.interaction.Draw.Event` in an application but thinking about this again, I'm now changing it to use [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) instead.

PR #8052, opened 2018-04-04
More Jsdoc module fixes

fredj:


src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -365,7 +365,7 @@ CanvasVectorTileLayerRenderer.prototype.handleStyleImageChange_ = function(event
 CanvasVectorTileLayerRenderer.prototype.postCompose = function(context, frameState, layerState) {
   const layer = this.getLayer();
   const declutterReplays = layer.getDeclutter() ? {} : null;
-  const source = /** @type {ol.source.VectorTile} */ (layer.getSource());
+  const source = /** @type {module:ol/source/VectorTile~VectorTileSource} */ (layer.getSource());
```
ahocevar: This is not consistent with the rest of the library. It's probably better to change the `VectorTile` import name in `src/ol/source/VectorTile` to `Tile`, and name the constructor `VectorTile`.

fredj: > and name the constructor `VectorTile`.

which one? in `ol/renderer/canvas/VectorTileLayer` (this file) ?

ahocevar: No, sorry for being unspecific. In `src/ol/source/VectorTile.js`.

PR #8075, opened 2018-04-17
Move olx.tilegrid.* to ol/tilegrid/*

pfirpfel:
Changes according to #7838, subtask of #7947


pfirpfel:
Tests are successful locally...

PR #8083, opened 2018-04-18
Move structs typedefs to ol/structs/*

pfirpfel:
Changes according to #7838, subtask of #7947


src/ol/structs/LRUCache.js
```diff
@@ -6,6 +6,16 @@ import {assert} from '../asserts.js';
 import EventTarget from '../events/EventTarget.js';
 import EventType from '../events/EventType.js';
 
+
+/**
+ * @typedef {Object} Entry
+ * @property {string} key_
+ * @property {Object} newer
+ * @property {Object} older
+ * @property {value_} value_
```
tschaub: The type here should be `*` (instead of `value_`).

PR #8087, opened 2018-04-19
Move various typedefs out of typdefs.js

pfirpfel:
Subtask of #7947

ol.Attribution and ol.AttributionLike are used all over ol/source. Although not really related to the base class ol/source/Source, it did seem as the best place to put the typedefs there.

Together with my other pull requests, this moves all typdefs out of typdefs.js but `ol.LoadingStrategy`. I don't know where to put that.


fredj:
rebased into master

PR #8103, opened 2018-04-20
Fix some jsdoc path types

ahocevar:
The latest version of `babel-plugin-jsdoc-closure` (#8102) reveals a few path type conversion mistakes, which are fixed by this pull request.

### DIFF
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -25,7 +25,7 @@ const Property = {
  * @constructor
  * @extends {module:ol/events/Event~Event}
  * @implements {oli.CollectionEvent}
- * @param {module:ol/CollectionEventType~CollectionEventType} type Type.
+ * @param {module:ol/CollectionEventType} type Type.
  * @param {*=} opt_element Element.
  */
 export const CollectionEvent = function(type, opt_element) {
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -59,7 +59,7 @@ inherits(CollectionEvent, Event);
  * Collection as a whole.
  *
  * @constructor
- * @extends {module:ol/Object~Object}
+ * @extends {module:ol/Object~BaseObject}
  * @fires module:ol/Collection~CollectionEvent
  * @param {Array.<T>=} opt_array Array.
  * @param {module:ol/Collection~Options=} opt_options Collection options.
--- a/src/ol/ImageTile.js
+++ b/src/ol/ImageTile.js
@@ -9,10 +9,16 @@ import {listenOnce, unlistenByKey} from './events.js';
 import EventType from './events/EventType.js';
 
 /**
+ * @typedef {function(new: module:ol/ImageTile~ImageTile, module:ol/tilecoord~TileCoord,
+ * module:ol/TileState, string, ?string, ol.TileLoadFunctionType)} TileClass
+ * @api
+ */
+
+/**
  * @constructor
  * @extends {module:ol/Tile~Tile}
  * @param {module:ol/tilecoord~TileCoord} tileCoord Tile coordinate.
- * @param {module:ol/TileState~TileState} state State.
+ * @param {module:ol/TileState} state State.
  * @param {string} src Image source URI.
  * @param {?string} crossOrigin Cross origin.
  * @param {module:ol/Tile~LoadFunction} tileLoadFunction Tile load function.
--- a/src/ol/layer/Base.js
+++ b/src/ol/layer/Base.js
@@ -28,7 +28,7 @@ import {assign} from '../obj.js';
  * Abstract base class; normally only used for creating subclasses and not
  * instantiated in apps.
  * Note that with `module:ol/layer/Base~BaseLayer` and all its subclasses, any property set in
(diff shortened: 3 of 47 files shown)


src/ol/layer/Base.js
```diff
@@ -28,7 +28,7 @@ import {assign} from '../obj.js';
  * Abstract base class; normally only used for creating subclasses and not
  * instantiated in apps.
  * Note that with `module:ol/layer/Base~BaseLayer` and all its subclasses, any property set in
- * the options is set as a {@link module:ol/Object~BaseObject} property on the layer object, so
+ * the options is set as a {@link } property on the layer object, so
```
fredj: the link is empty

src/ol/source/TileWMS.js
```diff
@@ -43,9 +43,7 @@ import {appendParams} from '../uri.js';
  * @property {module:ol/proj~ProjectionLike} projection Projection.
  * @property {boolean} [reprojectionErrorThreshold=0.5] Maximum allowed reprojection error (in pixels).
  * Higher values can increase reprojection performance, but decrease precision.
- * @property {function(new: module:ol/ImageTile~ImageTile, module:ol/tilecoord~TileCoord,
- *                 module:ol/TileState~TileState, string, ?string,
- *                 ol.TileLoadFunctionType)} [tileClass] Class used to instantiate image tiles.
+ * @property {module:ol/ImageTile~TileClasee} [tileClass] Class used to instantiate image tiles.
```
fredj: should be `module:ol/ImageTile~TileClass`

ahocevar:
@fredj Fixed, and added a few more fixes.

src/ol/Collection.js
```diff
@@ -25,7 +25,7 @@ const Property = {
  * @constructor
  * @extends {module:ol/events/Event~Event}
  * @implements {oli.CollectionEvent}
- * @param {module:ol/CollectionEventType~CollectionEventType} type Type.
+ * @param {module:ol/CollectionEventType} type Type.
```
tschaub: I'm curious about where we should/shouldn't add the ~.

src/ol/ImageTile.js
```diff
@@ -8,11 +8,17 @@ import {createCanvasContext2D} from './dom.js';
...
+ * module:ol/TileState, string, ?string, ol.TileLoadFunctionType)} TileClass
+ * @api
+ */
+
 /**
  * @constructor
  * @extends {module:ol/Tile~Tile}
  * @param {module:ol/tilecoord~TileCoord} tileCoord Tile coordinate.
- * @param {module:ol/TileState~TileState} state State.
+ * @param {module:ol/TileState} state State.
```
tschaub: Here too.

src/ol/style/index.jsdoc
```diff
@@ -28,16 +28,16 @@
  *  var white = [255, 255, 255, 1];
  *  var blue = [0, 153, 255, 1];
  *  var width = 3;
- *  styles[module:ol/geom/GeometryType~GeometryType.POLYGON] = [
+ *  styles[module:ol/geom/GeometryType.POLYGON] = [
```
tschaub: Seems like we should write valid JS here.  Either 'Polygon' or GeometryType.POLYGON, right?

ahocevar:
I agree it would be nice and also easier to transform if the tilde would be required only for named exports. However, JSDoc does not handle them that way. We could handle those in a JSDoc plugin though.

ahocevar:
@tschaub Expanding path types so they point to the default export in the docs is tested in https://github.com/openlayers/closure-es-modules/commit/d94d144ecd353847b5d9085f52f96a3a6582b208. In a follow-up pull request to this one, I could apply a babel transform derived from `babel-plugin-jsdoc-closure` to the code base to remove all `~DefaultExportFromVariable` parts from name paths.

tschaub:
Thanks for the work on this @ahocevar.

I understand that there may be JSDoc limitations involved here, but in terms of type annotations, I think it is important that we have flexibility in how we author our modules.

For example, these two modules export the same thing:

```js
const foo = 'bar';
export default foo;
```

```js
export default 'bar';
```

So our type annotations should be the same for both styles (e.g. `module:path/to/foo` or more explicitly `module:path/to/foo~default`).

And ideally, `module:path/to/foo~foo` would raise an error somewhere, since the module doesn't have an export namewd `foo`.

src/ol/layer/Vector.js
```diff
@@ -52,6 +52,22 @@ import {createDefaultStyle, toFunction as toStyleFunction} from '../style/Style.
...
+ *    scaled during zoom animations.
+ *  * `'vector'`: Vector layers are rendered as vectors. Most accurate rendering
+ *    even during animations, but slower performance.
+ * @api
+ */
+export const RenderType = {
+  IMAGE: 'image',
+  VECTOR: 'vector'
+};
+
```
tschaub: I'm wondering if adding this was an accident.  We have this enum in `VectorRenderType` already.

tschaub: See #8845.

Issue #8110, opened 2018-04-23
Re-export constructors


tschaub:
We can make it more convenient for users to import constructors by re-exporting them from common modules.  For example, all of the control constructors can be re-exported from `ol/control`.  Internally, we will continue to import from the origin, but we can update our examples to import from these common modules.  The end result should be fewer `import` statements for users.

E.g.

Before:
```js
import Style from 'ol/style/Style';
import Stroke from 'ol/style/Stroke';
import Fill from 'ol/style/Fill';
```

After:
```js
import {Style, Stroke, Fill} from 'ol/style';
```

Some of this work may involve shuffling things around to avoid cyclic dependencies (e.g. `iconImageCache` should be moved to `ol/style/IconImageCache~shared` or similar).

Here is a list of the work to be done:

- [x] export controls from `ol/control.js` (#8111)
- [x] export formats from `ol/format.js` (#8112)
- [x] export geometries from `ol/geom.js` (#8113)
- [x] export interactions from `ol/interaction.js` (#8114)
- [x] export layers from `ol/layer.js` (#8115)
- [x] export sources from `ol/source.js` (#8116)
- [x] export styles from `ol/style.js` (#8117)

PR #8117, opened 2018-04-23
Re-export style constructors from ol/style

tschaub:
This makes style constructors available from `ol/style`.

See #8110.

### DIFF
--- a/test/spec/ol/style/icon.test.js
+++ b/test/spec/ol/style/icon.test.js
@@ -1,5 +1,5 @@
 import {getUid} from '../../../../src/ol/index.js';
-import {iconImageCache} from '../../../../src/ol/style.js';
+import {shared as iconImageCache} from '../../../../src/ol/style/IconImageCache.js';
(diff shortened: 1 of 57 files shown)


test/spec/ol/style/icon.test.js
```diff
@@ -1,5 +1,5 @@
 import {getUid} from '../../../../src/ol/index.js';
-import {iconImageCache} from '../../../../src/ol/style.js';
+import {shared as iconImageCache} from '../../../../src/ol/style/IconImageCache';
```
ahocevar: We import with the `.js` Extension elsewhere, sobI think it should be done the same way here.

ahocevar: I added one remark. Please merge when addressed. Thanks for the effort on this!

tschaub: Good catch.  Addressed in e28305186c139545d366fe99bf44e1c7fa5d047f.

Issue #7970, opened 2018-03-12
Update type annotations


tschaub:
We need to update all of the legacy type annotations from the old `ol.foo.Bar` style to the new `module:ol/foo/Bar~Bar` style.

tschaub:
@ahocevar - I'm curious about default exports.  Do we always repeat the module name?  For example:
```js
/**
 * @module ol/geom/GeometryType
 */

/**
 * @enum
 */
export default {
  POINT: 'Point' // ...
};
```

Will `module:ol/geom/GeometryType~GeometryType` be right here?  Or is `module:ol/geom/GeometryType` right?

ahocevar:
For babel-plugin-jsdoc-closure and type checking it does not matter. For JSDoc there is a difference: the shorter will link to the module page (which also contains typedefs and named exports of that module, if any). The longer will link to the enum directly.

I also went a bit back and forth on this, and in the end decided to use the long version because we do the same for all other default exports, e.g. classes.

ger-benjamin:
A `git grep 'ol[.]render' src/` (for instance) still give me results like: 
 - https://github.com/openlayers/openlayers/blob/master/src/ol/render/canvas.js#L67 (link)
 - https://github.com/openlayers/openlayers/blob/master/src/ol/render/webgl/Immediate.js#L359 (type)
 - https://github.com/openlayers/openlayers/blob/master/src/ol/render/webgl/Immediate.js#L91 (param)
- and other ref.

I'll improve what I can.
Should we care about things outside of the "src/" ? because there is a lot of these notations in examples and the doc folders.

It mainly concerns:
 - [ ] ol.format
 - [ ] ol.pointer
 - [ ] ol.render
 - [ ] ol.DeclutterGroup
 - [ ] ol.Canvas* (mutiples)
 - [ ] ol.structs
 - [ ] ol.source
 - [ ] ol.ImageLoadFunctionType

ahocevar:
@ger-benjamin Thanks for any help you can provide. Examples and docs should also no longer use the `ol.*` notation.

ger-benjamin:
But what do you want for the documentation ? Like here: https://github.com/openlayers/openlayers/blob/master/doc/faq.md, for instance, for a `new ol.Map({...})`. Do you want that, in the code section, we add a line `import Map from ol.Map`, then use `new Map({...})` ?

ahocevar:
[...]

PR #8123, opened 2018-04-25
More Jsdoc module fixes

fredj:
see #7970


src/ol/render/canvas.js
```diff
@@ -155,8 +155,8 @@ export const defaultLineWidth = 1;
 
 /**
  * The label cache for text rendering. To change the default cache size of 2048
- * entries, use {@link ol.structs.LRUCache#setSize}.
- * @type {ol.structs.LRUCache.<HTMLCanvasElement>}
+ * entries, use {@link module:ol/structs/LRUCache~LRUCache#setSize}.
```
ahocevar: Deep `@link`s like this are the only case where we want to continue using the long module path.

PR #8125, opened 2018-04-25
Shorter module paths for default exports

ahocevar:
This implements the change suggested in https://github.com/openlayers/openlayers/pull/8103#issuecomment-383634301. To still get a link to the named export in JSDoc, we can use a simple plugin like the one introduced in https://github.com/openlayers/closure-es-modules/commit/d94d144ecd353847b5d9085f52f96a3a6582b208#diff-28a8c6de8bfb580271ebed73a8a9ae25.

### DIFF
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -23,7 +23,7 @@ const Property = {
  * type.
  *
  * @constructor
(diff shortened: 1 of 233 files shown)


ahocevar:
Thanks for the review @fredj-

PR #8129, opened 2018-04-26
Fix some types

ahocevar:
This pull request fixes a few type paths.


src/ol/Feature.js
```diff
@@ -252,7 +252,7 @@ Feature.prototype.setStyle = function(style) {
  * Set the feature id.  The feature id is considered stable and may be used when
  * requesting features or comparing identifiers returned from a remote source.
  * The feature id can be used with the
- * {@link module:ol/source/Vector~Vector#getFeatureById} method.
+ * {@link module:ol/source/Vector#getFeatureById} method.
```
tschaub: In https://github.com/openlayers/openlayers/pull/8123#discussion_r184369564 you mention that we need to use the long form for deep links like this.  Curious which will work.

src/ol/source/Vector.js
```diff
@@ -545,11 +545,11 @@ VectorSource.prototype.forEachFeatureAtCoordinateDirect = function(coordinate, c
  * value, iteration will stop and the function will return the same value.
  *
  * If you are interested in features whose geometry intersects an extent, call
- * the {@link module:ol/source/Vector~VectorSource#forEachFeatureIntersectingExtent
+ * the {@link module:ol/source/VectorSource#forEachFeatureIntersectingExtent
```
tschaub: I think these links need reworking (`ol/source/VectorSource` is not a module).

src/ol/style/Style.js
```diff
@@ -7,6 +7,18 @@ import CircleStyle from '../style/Circle.js';
...
 
+
+/**
+ * A function that takes an {@link module:ol/Feature} and a `{number}`
+ * representing the view's resolution. The function should return a
+ * {@link module:ol/style/Style} or an array of them. This way e.g. a
+ * vector layer can be styled.
+ *
+ * @typedef {function((module:ol/Feature|module:ol/render/Feature), number):
+ *     (module:ol/style/Style|Array.<module:ol/style/Style>)} Function
```
tschaub: Does the Compiler care if we have a typedef named `Function` even though `Function` is [already a type](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System).

ahocevar: It works because typedefs need to be imported in the transformed code, so `Function` ends up being part of a variable name. But you're right, it looks dangerous so I'm going to rename it to `StyleFunction`. BTW, the reason why I moved this from `ol/style` to `ol/style/Style` is because the compiler did not find it there, for some reason I can not make sense of. I agree with what you said a few days ago: TypeScript is looking good right now.

ahocevar:
Review comments addressed and rebased.

PR #8170, opened 2018-05-09
Basic apidocs

ahocevar:
This pull request brings back basic API documentation. It just touches the JSDoc config, not the template. We'll probably be using a tool that generates the final docs from a JSON, so we'll use the JSON template instead.

One thing that will probably need some additional attention is how we document events, so we can get rid of the events plugin. I'll handle that in a separate pull request.


config/jsdoc/api/conf.json
```diff
@@ -1,36 +1,34 @@
 {
-    "opts": {
-        "recurse": true,
-        "template": "config/jsdoc/api/template"
```
fredj: does it mean that the `config/jsdoc/api/template` directory can be removed or will it be used in a second step?

ahocevar: In the medium term we‘ll probably remove it. Leaving it here for now for testing/comparison reasons.

PR #8171, opened 2018-05-09
Remove wrong override tags

fredj:


fredj:
The override is for the closure compiler: 
https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#override

The problem here is that non of the superclasses implements the function: for example Control inherits from BaseObject which has no `setMap` function. The warning from the compiler is:
```
./build/src-closure/ol/control/Control.js:120: WARNING - property setMap not defined on any supertype of Control$$module$build$src_closure$ol$control$Control
Control.prototype.setMap = function(map) {

```

ahocevar:
Thanks for the explanation!

PR #8180, opened 2018-05-10
Fix some JSDoc issues

ahocevar:
Also makes a few minor changes to the template, so it can be used until
we have a new doc generation process. JSDoc output is written to
build/apidoc.

### DIFF
--- a/config/jsdoc/api/conf.json
+++ b/config/jsdoc/api/conf.json
@@ -1,6 +1,7 @@
 {
   "opts": {
-    "recurse": true
+    "recurse": true,
+    "template": "config/jsdoc/api/template"
   },
   "tags": {
     "allowUnknownTags": true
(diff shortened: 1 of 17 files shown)


src/ol/interaction.js
```diff
@@ -155,3 +134,26 @@ export function defaults(opt_options) {
...
+export {default as KeyboardZoom} from './interaction/KeyboardZoom.js';
+export {default as Modify} from './interaction/Modify.js';
+export {default as MouseWheelZoom} from './interaction/MouseWheelZoom.js';
+export {default as PinchRotate} from './interaction/PinchRotate.js';
+export {default as PinchZoom} from './interaction/PinchZoom.js';
+export {default as Pointer} from './interaction/Pointer.js';
+export {default as Select} from './interaction/Select.js';
+export {default as Snap} from './interaction/Snap.js';
+export {default as Translate} from './interaction/Translate.js';
+export {defaults};
```
tschaub: I'm curious if something about the documentation process requires that export statements be grouped at the bottom.  Feels like a fragile thing to have to maintain.

tschaub:
Thanks for making this work and for removing the JSDoc-specific workarounds from the source.

PR #8187, opened 2018-05-12
Hide attribution control if there are no attributions

tschaub:
Currently, if no sources have any attribution to show, we still display the attribution control.  With this change, we hide the control if there are no attributions.

Before:
image

After:
image


marcjansen:
A very nice change!

PR #8197, opened 2018-05-16
Fix #8195: Make child layer inherits from parent values

### LINKED ISSUE #8195: Layer doesn't inherit CRS from parent in WMS GetCap
If a WMS Get Cap fits this configuration : 
```
<Layer id="1">
    <CRS>EPSG:2056</CRS>
    <Layer id="2">
        <Layer id="3"
```
The layer 3 doesn't have any CRS in its list but it should inherit the list of CRS from layer 1

It's happening because the parser goes to the end of layer tree first before inheriting parent properties, see [here](https://github.com/openlayers/openlayers/blob/f284b95cb897b3ce1b49b985b58ebe1ca76be1fe/src/ol/format/WMSCapabilities.js#L451)

oterral:
Fix #8195

### DIFF
--- a/src/ol/format/WMSCapabilities.js
+++ b/src/ol/format/WMSCapabilities.js
@@ -447,12 +447,9 @@ function readCapabilityLayer(node, objectStack) {
  */
 function readLayer(node, objectStack) {
   const parentLayerObject = /**  @type {!Object.<string,*>} */ (objectStack[objectStack.length - 1]);
+  let layerObject = {};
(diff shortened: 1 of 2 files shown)


oterral:
any chance for a quick review @fredj @bartvde pls?

fredj:
@oterral The code looks good to me, can you please rebase into master?

davidoesch:
Ping @oterral

PR #8210, opened 2018-05-22
Align patterns and gradients to a grid

ahocevar:
This changes improves alignment of canvas patterns and gradients, especially when working with vector tiles.

#### Before
before

#### After
after

This is achieved by aligning pattern origins to a grid of 512 css pixels, which is anchored to map coordinate `[0, 0]`.


marcjansen:
very nice!

PR #8217, opened 2018-05-23
Re-enable type checking

ahocevar:
Looks like the recent Closure Compiler update changed things so type checking the way we want it now requires `compilation_level: 'ADVANCED'`.

With this change, we're back at >2000 warnings.

@fredj, maybe you want to take a look at those? I've pushed this branch to openlayers if you want to add commits.

### DIFF
--- a/tasks/typecheck.js
+++ b/tasks/typecheck.js
@@ -13,6 +13,7 @@ const compiler = new Compiler({
   dependency_mode: 'STRICT',
   checks_only: true,
   jscomp_error: ['newCheckTypes'],
+  compilation_level: 'ADVANCED',
   // Options to make dependencies work
   process_common_js_modules: true,
   hide_warnings_for: 'node_modules'


ahocevar:
Looks like path types for enums and structural interfaces are broken again in the current version of Closure Compiler. Given the fact that this seems to change with every release, I don't feel right now like the decision to continue using Closure Compiler for type checking was right.

fredj:
c1c9c77 deactivates extra checks made the NTI. With this the error count goes from 1238 to 387

ahocevar:
@fredj Will you be able to continue working on this?

fredj:
@ahocevar I'm not sure it's worthwhile to continue to enable the New Type Inference checks. What about changing the Closure Compiler configuration to have the same checks as the 4.x branch?

ahocevar:
This is not a blocker for the 5.0.0 release. If it's not finished by the time we cut the release, we'll publish 5.0.0 without type checking.

After the release, we can experiment with typescript JSDoc annotations (see Microsoft/TypeScript/#5158 and e.g. webpack/webpack#7031) to replace the JSDoc/Closure type annotations we currently use.

If that turns out to work, we'll be able to use better tooling for the library, and Closure compilers users should still have a way to automatically transform these annotations to ones that Closure compiler understands -- similar to what [https://npmjs.com/package/babel-transform-jsdoc-closure](babel-transform-jsdoc-closure) currently does.

fredj:
ok, removing the 5.0.0 milestone

PR #8223, opened 2018-05-24
Better vector tile performance and user experience

ahocevar:
This change brings interim tiles to vector image tiles. So when e.g. zooming in or panning, vector tile data from lower resolutions can be used and rendered. Users no langer have to stare at blurry lines and polygons while waiting for the tiles to load. The upscaled images are now only used when the frame time budget is exceed, so they are either not shown at all, or only for a very short time.

Overall performance also improves with this change, because the frame time budget is exceeded less often, so animations are more likely to finish within the scheduled time.

**Before**
may-24-2018 17-04-28

**After**
may-24-2018 17-04-17

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,10 @@
 
 ### Next release
 
+#### `transition` option of `ol/source/VectorTile` is ignored
+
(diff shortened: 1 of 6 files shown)


ahocevar:
Closing because Travis does not like thie PR.

ahocevar:
Resubmitted as #8227.

PR #8226, opened 2018-05-25
Accept renderers as sources for the Raster source

### LINKED ISSUE #8142: Consider keeping ol.source.ImageVector

To keep this project manageable for maintainers, we ask you to please check all boxes below before submitting an issue.

- [x] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [x] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [x] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

I've made a case to manage vector like a raster layer using `ol.source.imageVector` and styling it.
I combined this output to process raster with it using `ol.source.Raster`. From my understanding, as OpenLayers will remove the `ol.source.ImageVector` due to deprecation, it will not be possible to process it within `ol.source.Raster`.
You can see the demo at http://bl.ocks.org/ThomasG77/5bf03d9c2acaa4562fecb94aadfb04e4 to demonstrate my current way of doing. In this demo, it's a simple case but it could be more complex operations between vector and raster inputs.

ahocevar:
This pull request makes the raster source more flexible. In addition to accepting `ol/source/Source` subclasses as sources, it now supports `ol/render/canvas/Layer` subclasses as sources.

This can be useful for using a vector layer as input for the raster source. The modified tests show how to do that.

Fixes #8142.

### DIFF
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -14,9 +14,12 @@ import IntermediateCanvasRenderer from '../canvas/IntermediateCanvas.js';
 import {create as createTransform, compose as composeTransform} from '../../transform.js';
 
 /**
+ * Renderer for {@link module:ol/layer/Image} layers. When a vector renderer is
+ * set with the {@link module:ol/renderer/canvas/ImageLayer#setVectorRenderer}
+ * method, it can also render vector layers to an image.
  * @constructor
  * @extends {module:ol/renderer/canvas/IntermediateCanvas}
- * @param {module:ol/layer/Image} imageLayer Single image layer.
+ * @param {module:ol/layer/Image|module:ol/layer/Vector} imageLayer Image or vector layer.
  * @api
  */
 const CanvasImageLayerRenderer = function(imageLayer) {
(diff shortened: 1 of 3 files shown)


ahocevar:
Something appears to be broken in Sauce Labs. Tests pass locally.

tschaub:
Only reading on the phone, but it looks like there is a check for a source that is a TileLayer in here.  If we are going to update the raster source to accept layers as sources, we could accept a vector layer instead of requiring that it be wrapped in s renderer (which we might want to keep out of tha API).

ahocevar:
@tschaub With the current renderer plugin system, this would only work if the Raster source gets a dependency on the vector renderer. This (and the fact that the renderer constructors are already part of the `@api`) is why I chose the renderer approach in the end.

ahocevar:
@tschaub I found a way to support layers instead of renderers. See #8229.

Issue #8192, opened 2018-05-14
jsts example not working in version 5.x 


fredj:
http://openlayers.org/en/beta/examples/jsts.html

The error message is:
```
OL3Parser.js:28 Uncaught (in promise) TypeError: Cannot read property 'Point' of undefined
    at zi.read (OL3Parser.js:28)
    at jsts.js:1
```

fredj:
ping @bjornharrtell

bjornharrtell:
Oh.. :) Will have a look at this ASAP.

fredj:
thanks @bjornharrtell but there's no hurry !

PR #8231, opened 2018-05-28
Fix JSTS usage with OL classes injected

bjornharrtell:
This makes the JSTS example run again and also updates ref to JSTS to the latest stable version.

ref https://github.com/openlayers/openlayers/issues/8192.

### DIFF
--- a/examples/jsts.html
+++ b/examples/jsts.html
@@ -7,6 +7,6 @@ docs: >
   with OpenLayers.
 tags: "vector, jsts, buffer"
 resources:
-  - https://cdn.rawgit.com/bjornharrtell/jsts/gh-pages/1.4.0/jsts.min.js
+  - https://cdn.rawgit.com/bjornharrtell/jsts/gh-pages/1.6.0/jsts.min.js
 ---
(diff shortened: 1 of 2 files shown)


bjornharrtell:
No I agree that would be nice @ahocevar, but I was hesitant if example deps is ok to add to devDependencies which I guess is what is needed?

bjornharrtell:
Oh.. and actually I'm not sure to handle it on JSTS side - OpenLayers is a runtime dependency.

ahocevar:
I think it‘s ok to add example dependencies as devDependencies. We‘re doing it for proj4 already.

ahocevar:
> OpenLayers is a runtime dependency.

So the inject call is necessary regardless?

bjornharrtell:
@ahocevar, afraid so I think.. until dynamic import() "just works" I guess. :( Injection could probably done in a nicer way though.

ahocevar:
So should we call it good and merge as-is, or do you want to play with it a bit more?

bjornharrtell:
Good and merge as is is my vote.

PR #8233, opened 2018-05-29
Use rollup instead of webpack for the full build

ahocevar:
This results in a smaller build size and prepares us for the new website and example generation tools (see the `site` branch).


ahocevar:
> I’m curious what features require Bublé

The uglify plugin throws errors without it.

PR #8250, opened 2018-06-06
Fix various jsdoc types 

fredj:
found by the closure compiler


fredj:
Thanks for the review

PR #8258, opened 2018-06-08
Use Webpack 4 for tests (and examples)

ahocevar:
This pull request concludes @tschaub's work in #7740, and focuses on improvements to our test runner. So in addition to making the examples work with Webpack 4, it also changes our test configuration to use Webpack 4, run tests faster by creating a single bundle, and adding an inline source map which makes debugging tests in the browser a joyful experience.


package.json
```diff
@@ -41,15 +41,15 @@
   "dependencies": {
     "pbf": "3.1.0",
     "pixelworks": "1.1.0",
-    "rbush": "2.0.2"
+    "rbush": "2.0.2",
+    "uglifyjs-webpack-plugin": "^1.2.5"
```
fredj: shouldn't it be moved to `devDependencies`?

PR #8299, opened 2018-06-21
Make classes in ol/ available as named exports from ol

ahocevar:
For consistency with the rest of the API, this pull request makes classes in the ol folder available as named exports of `ol`, so users can e.g.
```js
import {Map, View} from 'ol';


bartvde:
@ahocevar will this work in combination with tree shaking?

ahocevar:
@bartvde Like for `ol/control` or `ol/source`, where we have such re-exports already, treeshaking should not be affected. For webpack, we have `sideEffects: false` in `package.json`.

gberaudo:
Special +1 for having cleaned out the `index.js`.

PR #8301, opened 2018-06-22
Do not render vectors in image mode

ahocevar:
This pull request improves performance for `"image"` render mode with vector tiles, by not trying to render vectors at all.


fredj:
note: easier to view the diff with the `w=1` param:
https://github.com/openlayers/openlayers/pull/8301/files?w=1

PR #8319, opened 2018-06-27
Remove reference to setProj4 function in documentation

fredj:
The function was removed in #7598


fredj:
I've added a link to the register function

fredj:
Thanks for the review

PR #8322, opened 2018-06-27
Revert "Make wrap work with points in the vector-labels example"

venugopal-rao:
Reverts openlayers/openlayers#7944


ahocevar:
What is the intention of this pull request?

tschaub:
I'm going to assume that this pull request was opened by accident.  Please open a ticket with the issue you are hoping to address if this is not the case.

PR #8341, opened 2018-07-04
Set sideEffects: false flag

### LINKED ISSUE #8337: Bundle size is too big
Hi, I've been using OL with custom build with version 4.5. The build I've customized with build.json is about 240kb. but with 5.0 served from npm, its bundle size is 1.7mb. I just followed [bundle tutorial](https://github.com/openlayers/openlayers/blob/master/doc/tutorials/bundle.md) to load basic OSM map though. How can I include only things that are really required?

ahocevar:
Bundlers start looking for a `"sideEffects": "false"` property in `package.json` for more aggressive tree shaking. By setting this flag, we help users get smaller bundle sizes.

Example: with the [Webpack example project](https://gist.github.com/tschaub/79025aef325cd2837364400a105405b8), this change reduces the bundle size from ~440kB to ~225kB.

Fixes #8337.

### DIFF
--- a/src/ol/package.json
+++ b/src/ol/package.json
@@ -9,5 +9,6 @@
     "pbf": "3.1.0",
     "pixelworks": "1.1.0",
     "rbush": "2.0.2"
-  }
+  },
+  "sideEffects": "false"
 }


bartvde:
nice catch!

bartvde:
@ahocevar does this require webpack 4+ ?

ahocevar:
@bartvde I don't know if it requires webpack 4+, but it definitely makes a difference with webpack 4+.

PR #8345, opened 2018-07-04
TypeScript (1/n)

tschaub:
This is an initial, work-in-progress branch that replaces Closure Compiler type checking with TypeScript type checking.  The first phase of this work will be to replace our custom type annotations with those understood by `tsc`.

This branch will transform the `module:ol/foo/Bar~Options` types in an automated way.  For example, say we currently have a typedef in `ol/foo/Bar.js`:

```js
// in ol/foo/Bar.js

/**
 * @typedef {Object} Options.
 * @property {string} bam A string.
 */
```

Currently, we use that typedef in another module like this:

```js
// before in ol/foo/Baz.js

/**
 * @param {module:ol/foo/Bar~Options} options Some options.
 */
// ...
```

After this change, we will "import" the type and use it like this:
```js
// after in ol/foo/Baz.js

/** @typedef {import("./Bar.js").Options} _ol_foo_Bar_Options_ */

/**
 * @param {_ol_foo_Bar_Options_} options Some options.
 */
// ...
```

We can follow up this change with hand edited type names (e.g. `BarOptions` instead of `_ol_foo_Bar_Options_`).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,60 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
+    "target": "ES2017",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "lib": ["es2017", "dom"],                        /* Specify library files to be included in the compilation. */
+    "allowJs": true,                          /* Allow javascript files to be compiled. */
+    "checkJs": true,                          /* Report errors in .js files. */
+    // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+    // "declaration": true,                   /* Generates corresponding '.d.ts' file. */
+    // "sourceMap": true,                     /* Generates corresponding '.map' file. */
+    // "outFile": "./",                       /* Concatenate and emit output to single file. */
+    // "outDir": "./",                        /* Redirect output structure to the directory. */
+    // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
+    // "removeComments": true,                /* Do not emit comments to output. */
+    "noEmit": true,                           /* Do not emit outputs. */
+    // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
+    // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+    // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+
+    /* Strict Type-Checking Options */
+    "strict": false,                          /* Enable all strict type-checking options. */
+    // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+    // "strictNullChecks": true,              /* Enable strict null checks. */
+    // "strictFunctionTypes": true,           /* Enable strict checking of function types. */
+    // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+    // "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
+    // "alwaysStrict": true,                  /* Parse in strict mode and emit "use strict" for each source file. */
+
+    /* Additional Checks */
+    // "noUnusedLocals": true,                /* Report errors on unused locals. */
+    // "noUnusedParameters": true,            /* Report errors on unused parameters. */
+    // "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
+    // "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
+
+    /* Module Resolution Options */
+    // "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    // "baseUrl": "./",                       /* Base directory to resolve non-absolute module names. */
+    // "paths": {},                           /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
+    // "rootDirs": [],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
+    // "typeRoots": [],                       /* List of folders to include type definitions from. */
+    // "types": ["node"],                        /* Type declaration files to be included in compilation. */
+    // "allowSyntheticDefaultImports": true,  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
+    "esModuleInterop": true                   /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
+    // "preserveSymlinks": true,              /* Do not resolve the real path of symlinks. */
+
+    /* Source Map Options */
+    // "sourceRoot": "./",                    /* Specify the location where debugger should locate TypeScript files instead of source locations. */
+    // "mapRoot": "./",                       /* Specify the location where debugger should locate map files instead of generated locations. */
+    // "inlineSourceMap": true,               /* Emit a single file with source maps instead of having a separate file. */
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+  },
+  "include": [
+    "src/ol/**/*.js"
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "rbush": "2.0.2"
   },
   "devDependencies": {
+    "@openlayers/eslint-plugin": "^4.0.0-beta.1",
     "buble": "^0.19.3",
     "buble-loader": "^0.5.1",
     "chaikin-smooth": "^1.0.4",
@@ -84,7 +85,19 @@
     "webpack-dev-server": "^3.1.4"
   },
   "eslintConfig": {
-    "extends": "openlayers"
+    "extends": "openlayers",
+    "plugins": [
+      "@openlayers"
+    ],
+    "rules": {
+      "valid-jsdoc": "off",
+      "@openlayers/valid-tsdoc": [
+        "error",
+        {
+          "requireReturn": false
+        }
+      ]
+    }
   },
   "sideEffects": [
     "ol.css"
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "build-legacy": "rm -rf build && npm run build-index && rollup --config config/rollup.js && cleancss --source-map src/ol/ol.css -o build/ol.css",
     "copy-css": "cp src/ol/ol.css build/ol/ol.css",
     "transpile": "rm -rf build/ol && mkdir -p build && buble --input src/ol --output build/ol --no modules --sourcemap",
+    "typecheck": "tsc --pretty",
     "apidoc": "jsdoc config/jsdoc/api/index.md -c config/jsdoc/api/conf.json -P package.json -d build/apidoc"
   },
   "main": "index.js",
@@ -74,6 +75,7 @@
     "rollup-plugin-sourcemaps": "0.4.2",
     "rollup-plugin-uglify": "5.0.2",
     "sinon": "^6.0.0",
+    "typescript": "^3.1.0-dev.20180905",
     "uglifyjs-webpack-plugin": "^1.2.5",
     "url-polyfill": "^1.0.13",
     "walk": "^2.3.9",
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -24,7 +24,7 @@ const Property = {
 export class CollectionEvent extends Event {
 
   /**
-   * @param {module:ol/CollectionEventType} type Type.
+   * @param {import("./CollectionEventType.js").default} type Type.
    * @param {*=} opt_element Element.
    */
   constructor(type, opt_element) {
@@ -63,7 +63,7 @@ class Collection extends BaseObject {
 
   /**
    * @param {Array<T>=} opt_array Array.
-   * @param {module:ol/Collection~Options=} opt_options Collection options.
+   * @param {Options=} opt_options Collection options.
    */
   constructor(opt_array, opt_options) {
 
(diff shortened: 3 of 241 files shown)


tschaub:
Looks like we will need to differentiate between imported typedef and imported classes.

tsconfig.json
```diff
@@ -0,0 +1,60 @@
...
+    // "mapRoot": "./",                       /* Specify the location where debugger should locate map files instead of generated locations. */
+    // "inlineSourceMap": true,               /* Emit a single file with source maps instead of having a separate file. */
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+  },
+  "include": [
+    "src/ol/style/Text.js"
```
Turbo87: this should probably be `src/**/*.js`, right?

tschaub:
@Turbo87 - it would be great to have your help.  We need to make the change to `class` syntax first (see #8385).

fgravin:
> Looks like we will need to differentiate between imported typedef and imported classes.

I think you don't need to declare a @typedef for classes, typescript handles the class type checking with the classic import.

To me
```js
/** @typedef {import("./Collection.js")} _ol_Collection_ */
```

is already handled by
```js
import Collection from "./Collection"
```
then you can use `{Collection}` for type checking.

fgravin:
> /** @typedef {import("./Bar.js").Options _ol_foo_Bar_Options_ */

I think 

```js
import {Options} from "./Bar.js"
/**
 * @param {Options} options Some options.
 */
```
works as well

tschaub:
We still need the comment style import where we only use the type in a comment (and only code style import where we use the import in code) if we keep the same linter rules.

The comment style import works for classes when using class syntax in the other module.  There may still be issues to work out, but the transform here and the work in #8385 is going in the right direction.

tschaub:
Ok, I've found an import comment syntax that will work now that we have proper `class`es.  The key is to be explicit about `default` exports.  So, when working with a type from another module, these are two options:

Options 1 (local typedef):
```js
/** @typedef {import("./Foo.js").default} Foo */

/**
 * @param {Foo} foo The foo.
 * @returns {string} A string.
 */
export function getBar(foo) {
  return foo.getBar();
}
```

Option 2 (inline import):
```js
/**
 * @param {import("./Foo.js").default} foo The foo.
 * @returns {string} A string.
 */
export function getBar(foo) {
  return foo.getBar();
}
```

I'm partial to Option 2 since I think it will make documentation slightly less hard.  I'll update [the codemod](https://gist.github.com/tschaub/1ea498c9d1e5268cf36d212b3949be4e) to rework our comments to match this style now.

The other semi-blocker I've run into is `@link` parsing.  TypeScript 3 has a fix for `@link` parsing (see Microsoft/TypeScript#25206), but this is not yet in a stable release (Microsoft/TypeScript#16498 is the broader issue).  I've updated our dependency to `typescript@3.0.0-rc`, so our `npm run typecheck` script can work.  But editors won't yet use this version, so it would be nice to have a workaround.  I'm inclined to replace our `@link`s with markdown links wherever they occur in a `@typedef`.  I'll check now to see how many of these are links to other types - these would be more fragile to change (since we may change the doc anchor names).

tschaub:
Another thing to ticket: it looks like our grouping with parens doesn't work in function typedefs.

For example, instead of
```js
/**
 * @typedef {function((string|undefined)): (string|undefined)} SomeType
 */
```

We need to do
```
/**
 * @typedef {function(string|undefined): (string|undefined)} SomeType
 */
```

I've opened Microsoft/TypeScript#25779, but I think it would be good to start getting rid of the unnecessary `(` `)`.

fredj:
@tschaub I've fixed some errors detected by typescript, see https://github.com/tschaub/openlayers/pull/2

tschaub:
Thanks @fredj.  It looks like those changes could all be made against master.  To the extent possible, I’d like to fix the issues we encounter here as separate PRs.  And I’ll keep rebasing this.

fredj:
@tschaub I've created a pull request against master, see #8402

tschaub:
Ok, the next blocking(ish) issue is that `valid-jsdoc` treat's TypeScript's imported type syntax as invalid.  I've proposed eslint/doctrine#216 as a fix.  If something like this doesn't make its way into ESLint, we can create a custom `valid-tsdoc` rule with the change.

sandersn:
@tschaub Our nightlies haven't been publishing because npm reset our tokens. I'll update this thread when we get that fixed.

sandersn:
@weswigham pointed out in my discussion with him that Typescript also checks jsdoc for validity, so `valid-jsdoc` may be redundant. I'm not sure how much overlap there is between the two checks. `valid-jsdoc` might still provide some errors that Typescript doesn't.

tschaub:
> @weswigham pointed out in my discussion with him that Typescript also checks jsdoc for validity, so valid-jsdoc may be redundant. I'm not sure how much overlap there is between the two checks. valid-jsdoc might still provide some errors that Typescript doesn't.

Thanks for the updates @sandersn - and good pointer about checking doc validity with tsc.  When we get this first round of Typescript work in, we can adjust our linting rules.  I'll check to see if there are checks we still want to keep from `valid-jsdoc`.  It would be ideal to use a single tool for type checking and doc extraction.

tschaub:
Looks like our next blocker is lack of @enum support (Microsoft/TypeScript#18161).

OSHistory:
Where can i find typescript related tickets, chores that need to 
be done moving the publication of ts-index files from within ol along? 
Is there a central issue to keep track of the ts-developments in ol?

ahocevar:
@OSHistory Thanks for willing to help out. I just created #8448 to track the effort. Any comments and contributions are welcome.

sandersn:
@tschaub What does the `@enum` annotation do to the annotated object literal? As far as I can tell it just makes sure that the properties always get their literal type. That is,

```js
var e = {
   True = ‘True’,
   False = ‘False’
}
```

Has the incorrect type `{ True: string, False: string }`, but would have type `{ True: “True”, False: “False” }` if annotated with @enum. Is that correct?

tschaub:
@sandersn the `@enum` tag serves two purposes with the Closure Compiler.

 1. When used in member expressions, the property must exist in the object.
 2. Values are type checked where used.

For example:
```js
/**
 * The geometry type.
 * @enum {string}
 */
const GeometryType = {
  POINT: 'Point',
  LINE_STRING: 'LineString',
  POLYGON: 'Polygon'
};


/**
 * @return {string}
 */
function validUse() {
  return GeometryType.POINT;
}

/**
 * Inexistent enum element.
 * @return {string}
 */
function invalidUse1() {
  return GeometryType.FOO;
}

/**
 * Inconsistent return type.
 * @return {boolean}
 */
function invalidUse2() {
  return GeometryType.POINT;
}
```

It isn't clear to me yet how this maps to TypeScript's `enum`.

sandersn:
@tschaub I don't think it does map to Typescript's enum. From typescript's perspective, all `@enum` adds is making the object literal "closed" -- typescript allows all JS object literals to have any property, but only allows reference to declared properties in TS.

sandersn:
@tschaub Do you have a script or something that converts the `module:/path/etc` references to `import("../path/etc")` references? I want to play around with my addition of `@enum` to the Typescript compiler but its swamped by parse errors right now.

tschaub:
@sandersn I neglected to update the branch after rebasing.  The latest commit in my [`ts` branch](https://github.com/openlayers/openlayers/pull/8345/commits/6797f560497fc4237e0556629de1d9395c3ac81a) has the TypeScript `import` types.  Those changes are generated by running this on `master`:

```
# ts.js from https://gist.github.com/tschaub/1ea498c9d1e5268cf36d212b3949be4e
jscodeshift --transform ts.js src
```

(This is overkill for `jscodeshift`, but I was initially thinking it would involve more ast transforms.)

Let me know if there is anything I can help test out.  And thanks a ton for your ideas & help with this!

ahocevar:
@tschaub Did you find out anything about simple typedefs like
```js
/**
 * @typedef {Array<number>} Coordinate
 */
```
I'm not able to use or import those. It will probably be easiest to get rid of those and just use the underlying type instead.

ahocevar:
> I'm not able to use or import those. It will probably be easiest to get rid of those and just use the underlying type instead.

Looks like I did something else wrong. These work fine now.

ahocevar:
Microsoft/TypeScript#26067 has been merged and published under the `@latest` tag,, so we should be able to start fixing types in our code base, based on this branch's output from `tsc --pretty`.

tschaub:
With the latest `typescript@next` (3.1.0-dev.20180830), we still get a lot of spurious errors reported from `npm run typecheck`.  But some of them are things we can address here.  Others depend on upstream TypeScript fixes.  Here is a breakdown of a few of the error types:

1. Incorrect types.  The XML parsers use `Node` all over the place.  In many places, the functions really only work with `Element` or `HTMLElement`.  Here is an example error of this type:

    ```
    src/ol/xml.js(408,17): error TS2339: Property 'firstElementChild' does not exist on type 'Node'.
    ```

2. Multiline types.  TypeScript does not properly parse JSDoc type expressions that span more than one line.  See Microsoft/TypeScript#26528 for a proposed fix.  Until this addressed, most of the errors we get from `npm run typecheck` have to be ignored.

    Here is an example of a multiline type expression (from [`ol/Tile.js`](https://github.com/openlayers/openlayers/blob/01d8b9be87fa94d7596c113e79af2c0210a58889/src/ol/Tile.js#L57-L58)):

    ```js
    /**
     * @typedef {function(import("./tilecoord.js").TileCoord, number,
     *           import("./proj/Projection.js").default): (string|undefined)} UrlFunction
     */
    ```

    And here is an example of the resulting error (when calling a `UrlFunction` with three args):

    ```
    src/ol/tileurlfunction.js(85,16): error TS2554: Expected 4 arguments, but got 3.
    ```


3. Typedefs whose names are not on the same line as the tag.  This is part of the broader problem of skipping newline + asterisk in TypeScript, but this is an issue that we could choose to fix in the library (by moving names onto the same line as the tag and type).  We could also propose a fix upstream that is similar to the fix for multiline `@param` parsing.

    Here is an example of a typedef whose name is not on the same line as the tag (from [`ol/style/Style.js`](https://github.com/openlayers/openlayers/blob/01d8b9be87fa94d7596c113e79af2c0210a58889/src/ol/style/Style.js#L122-L123)):

    ```js
    /**
     * @typedef {function((import("../coordinate.js").Coordinate|Array<import("../coordinate.js").Coordinate>|Array<Array<import("../coordinate.js").Coordinate>>),import("../render.js").State)}
     * RenderFunction
     */
    ```

    And here is an example of the resulting error:

    ```
    src/ol/style/Style.js(186,15): error TS2304: Cannot find name 'RenderFunction'.
    ```

It would be great if people wanted to start fixing the errors of the first type (improper use of `Node` in the XML parsers).

I'll look into a fix for the third type of error (typedef names on a different line than the tag).  I don't think it is necessarily worth changing these in our library, since the broader multiline JSDoc parsing issue needs to be addressed if we are going to be able to use TypeScript for type checking.

If anybody wants to weigh in on the TypeScript multiline JSDoc type parsing issue, please see Microsoft/TypeScript#23667 and the proposed partial solution in Microsoft/TypeScript#26528.

tschaub:
Here is a proposed fix for the "typedefs whose names are not on the same line..." issue: Microsoft/TypeScript#26775.

tschaub:
I ticketed the issue with XML formats in #8560.

tschaub:
Getting closer :)
```
$ npx tsc | wc -l
    1665
```

Most of the remaining issues are due to multiline types.

There are also a number of errors due to our use of `@abstract`.  It looks like TypeScript is [not parsing](https://github.com/Microsoft/TypeScript/blob/45101491c0b077c509b25830ef0ee5f85b293754/src/compiler/parser.ts#L6543-L6581) these.  For example, we get this error:
```
src/ol/webgl/Shader.js:26:15 - error TS2355: A function whose declared type is neither 'void' nor 'any' must return a value.

26    * @return {number} Type.
                 ~~~~~~
```

For this code in [`ol/webgl/Shader.js`](https://github.com/openlayers/openlayers/blob/a35b0465ab67347795910db2b0eedec85e1c033a/src/ol/webgl/Shader.js#L24-L28):
```js
  /**
   * @abstract
   * @return {number} Type.
   */
  getType() {}
```

I'll ticket this upstream.

tschaub:
> Most of the remaining issues are due to multiline types.

This might not have been true.

I've updated Microsoft/TypeScript#26528 so it now works for multiline types (including multiline function signatures).  After these changes, `npx tsc | wc -l` is at 1509.  The good news, there is plenty of work to go around!

OSHistory:
I ran a bash line to count the different typescript errors: 

`npx tsc | grep 'error TS[0-9]*' -o | cut -d" " -f2 | sort | uniq -c | sort -n`

The result is: 

~~~
      1 TS2363
      1 TS2555
      1 TS8024
      2 TS2425
      2 TS2447
      2 TS2557
      2 TS2684
      2 TS2719
      3 TS1345
      3 TS2362
      3 TS2694
      4 TS1005
      4 TS2540
      5 TS2351
      8 TS2349
      8 TS2367
     10 TS2416
     11 TS2551
     16 TS2352
     28 TS2554
     35 TS2306
     56 TS2355
    103 TS2304
    129 TS2345
    197 TS2322
    308 TS2339
~~~

The four most frequent are
TS2339: Property 'x' does not exist on type 'y'
TS2322: Type 'x' is not assignable to type 'y'
TS2345:  Argument of type 'x' is not assignable to parameter of type 'y'
TS2304: Cannot find name 'x'

tschaub:
I've updated things so we now depend on typescript@3.1.0-dev.20180905 which includes the fix for multiline type expressions (Microsoft/TypeScript#26528).

There are still some issues parsing (e.g. [`@abstract` parsing](Microsoft/TypeScript/issues/17227)), but I think the majority of the issues reported by `npm run typecheck` would be easier to address if this were merged.

So, this is ready for review.

idevtech:
What's the status of getting typescript definitions for v5x (?5.3.2) This is holding up deployment of an app because of the need for some events new to v5.x.

hanreev:
I've generated [typescript declarations for v5.3.2](https://github.com/hanreev/openlayers-docset/tree/master/typings/ol) using JSDoc.

This is modified [JSDoc template](https://github.com/hanreev/openlayers-docset/tree/master/jsdoc) for generating Dash docset and typescript declarations
based on [openlayers JSDoc template](https://github.com/openlayers/openlayers/tree/master/config/jsdoc/api).

OSHistory:
@hanreev Sounds great, thanks! I suppose, you moved it to a separate repo (https://github.com/hanreev/types-ol)? I checked it out, first impression seems good. Have you checked, if your approach works with the 6.0 developments? Might be an idea to advertise your efforts in some other issues related to TS-Integration.

hanreev:
@OSHistory I've released beta version for openlayers 6.0.0-beta.11
[https://github.com/hanreev/types-ol/tree/beta](https://github.com/hanreev/types-ol/tree/beta)

PR #8353, opened 2018-07-05
Avoid error when calling getExtent on null geometry

tonio:
If `flatCoordinates` were null, an error was thrown.

### DIFF
--- a/src/ol/geom/SimpleGeometry.js
+++ b/src/ol/geom/SimpleGeometry.js
@@ -3,7 +3,7 @@
  */
 import {inherits} from '../util.js';
 import {FALSE} from '../functions.js';
-import {createOrUpdateFromFlatCoordinates, getCenter} from '../extent.js';
+import {createOrUpdateFromFlatCoordinates, getCenter, createOrUpdateEmpty} from '../extent.js';
 import Geometry from '../geom/Geometry.js';
 import GeometryLayout from '../geom/GeometryLayout.js';
 import {rotate, scale, translate, transform2D} from '../geom/flat/transform.js';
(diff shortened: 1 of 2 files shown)


ahocevar:
I'm not sure this is the right fix. It would be better to change the type of the constructor's coordinates argument so it does not accept `null` coordinates.

tonio:
Not a big fan of the new version because you have to fake length as well, but ready for another review.

tonio:
Looks like I misread your comment. There is a test to check null geom is allowed. So looks like its on purpose ?

ahocevar:
Hm. Accepting null as coordinates in a geometry constructor was introduced with 1a9d19a2fb8ff978d82f43c5cb8135a0660c490c. In addition to being used in `draw-shapes` example, geometries with null coordinates are used in many feature formats.

So maybe for now it makes sense to go back to a modification of your initial pull request, i.e. return `null` for `getExtent()` on geometries without coordinates.

tonio:
To have coherent typing, empty extent looks better to me than `null`. Or we'll have to deal with «maybe null »`getExtent` elsewhere in the code. We'll still have the same issue with `getFirstCoordinate` & `getLastCoordinate`, not sure what's best for them (as there's no «empty» coordinate).

ahocevar:
Ideally, we‘d disallow null coordinates in the constructor, and add a SimpleGeometry#fromFlatCoordinates instead.

tonio:
Sounds good. I'll look at it.

ahocevar:
@tonio Probably even better, we could make the geometry constructors smarter by adding an additional `ends` argument (for polygons etc.), and accepting flat coordinates in addition to coordinates. I'll start a branch to show what I mean, and you can add commits to that.

ahocevar:
I have created #8362, which replaces this pull request and fully implements what I suggested in https://github.com/openlayers/openlayers/pull/8353#issuecomment-402956736.

PR #8362, opened 2018-07-07
Smart flat coordinates

ahocevar:
This pull request simplifies geometry construction and improves performance by avoiding a change event upon construction.

Replaces #8353.


ahocevar:
Thanks for the review @tonio. I'll check what happened to your commit rights. Looks like @sbrunner's are also gone.

tonio:
thanks!

PR #8363, opened 2018-07-08
Autogenerate src/ol/package.json

ahocevar:
This pull request removes `src/ol/package.json` and replaces the `set-version` task with a `prepare-package` task, which autogenerates `src/ol/package.json`.


ahocevar:
Any chance to get a review on this? I'd like to use the generated package.json for the 5.0.3 release, because it will give us a link to Github and the website.

Thanks in advance!

PR #8364, opened 2018-07-08
Make examples work in IE11

### LINKED ISSUE #8356: IE11 not supported anymore
Openlayers does not work anymore with Internet Explorer 11:
See for example https://openlayers.org/en/latest/examples/zoomify.html

Error: Const must be initialized

IE11 does not support something like:
for(const a of [1, 2, 3]){
}

ahocevar:
The full build works already in IE 11 (thanks to [buble](https://github.com/openlayers/openlayers/blob/04e825e49aae22ad8c71f73b33f82c54b790c6a2/config/rollup.js#L17)), so we only need to add babel transpilation to the examples.

Fixes #8356.

### DIFF
--- a/examples/webpack/config.js
+++ b/examples/webpack/config.js
@@ -18,6 +18,18 @@ module.exports = {
   context: src,
   target: 'web',
   entry: entry,
+  module: {
+    rules: [{
+      use: {
+        loader: 'buble-loader'
+      },
+      test: /\.js$/,
+      include: [
+        path.join(__dirname, '..', '..', 'src'),
+        path.join(__dirname, '..')
+      ]
+    }]
+  },
   optimization: {
     runtimeChunk: {
       name: 'common'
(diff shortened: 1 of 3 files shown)


ahocevar:
I'd like to include this in the v5.0.3 patch release. Thanks for any review!

ahocevar:
The examples now also work in IE10 again.

TrevorOchmonek:
Most of the examples don't work for me in IE11. Take the [very first](https://openlayers.org/en/latest/examples/accessible.html) one for example.
Looks like the ... / spread operator in index.es.js line 7944 is the problem:
`function extend (output, ...inputs) {`

mike-000:
> Most of the examples don't work for me in IE11. Take the [very first](https://openlayers.org/en/latest/examples/accessible.html) one for example.
> Looks like the ... / spread operator in index.es.js line 7944 is the problem:
> `function extend (output, ...inputs) {`

See #11866  There are working examples using the fixed dev version available via https://openlayers.org/en/main/examples/

PR #8376, opened 2018-07-12
copy ZM values to polygon in makeRegular

bkuster:
When creating a polygon using `Polygon.fromCircle` the `Z` or `ZM` values in `XYZ` or `XYZM` geometry layouts are copied to the new geometries coordinates, instead of assuming `0`.


bkuster:
Good catch, didn't think of that. It just seemed like less changes.

PR #8385, opened 2018-07-17
Use class syntax

### LINKED ISSUE #8328: Use ES6 classes and inheritance
The OpenLayers5 code is using constructor functions and inheritance using util.inherits().
ES6 introduced the concept of class which:
- is well understood by IDEs;
- allow using `super` in constructor and methods;
- is a prerequisite to using typescript as JSDoc or direct typescript authoring as described in #8120.

I have started experimenting with [lebab](https://github.com/lebab/lebab) for transforming our 250 constructors to classes.
Before I invest more time:
- is a PR welcome?
- have anyone already experimented with such transform?

### LINKED ISSUE #8384: IE 11 support (without transpilation)
When converting to `const` and `let`, it was an oversight on our part that IE 11 does not support `const` or `let` in `for..in` loops.  Unfortunately, this means that all consumers have to set up special rules (e.g. in webpack) to transform the `ol` dependency.  It would be nice for consumers to be able to retain IE 11 support with minimal effort.

Options:
 1. use `for (var foo in bar)` instead of `for (const foo in bar)` in the source
 2. transform sources before publishing
 3. make everybody do more work to get IE 11 support

I like that we don't currently transform before publishing because there are clear limits to the syntax we use.  If we do transform, we will need to decide how far ahead of our target browsers to go - and we risk writing code with crappy performance when transpiled.  But, we could probably control things with our linter rules.

tschaub:
This branch converts our sources to use `class` syntax.  The purpose of this change is to get TypeScript type checking (see #8345).

The initial commit is an automated transform, but there will be a lot of manual work to complete this effort, so I think it makes sense to pause other work (hold off on merging other PRs) until this is complete.  Contributions are welcome - either pushed directly to this branch or pushed to other branches and cherry picked here.

Remaining work:

 - [x] Replace use of `inherits` with `extends` and call `super()` instead of `Parent.call(this)`
 - [x] Rework things to avoid use of `this` before `super()`
 - [x] Remove all `@extends` and `@constructor` annotations
 - [x] Add a transform to the test config (and fix any failing tests)
 - [x] Add a transform step to build the package before publishing

Fixes #8328.
Fixes #8384.


Turbo87:
@fredj are you working on this right now? wouldn't want to do duplicate work 😉

fredj:
@Turbo87 yes I am moving the jsdoc comments from the class to the constructor function.

Turbo87:
hmm, that created conflicts with the `extends` changes locally already. I'll wait until you're done...

fredj:
@Turbo87 I've push a new commit but I will not have more time today to work on this. You can go ahead with your changes !

ahocevar:
@Turbo87 What are you working on right now? So we don't duplicate efforts.

ahocevar:
@Turbo87 Ok. For files with names starting with A to M, I'm adding `extends` while removing `inherits()`, and replacing prototype `call`s with `super()` now.

gberaudo:
For information, I have some patches for lebab to:
- detect our `inherits` calls;
- move constructor comments to the constructor method.

See https://github.com/gberaudo/lebab/tree/openlayers_improvements.

ahocevar:
Done with A-M, now making the aforementioned changes for N-Z.

ahocevar:
Done with classes starting with N-Z, now touching classes in `ol/control`.

tschaub:
🎉

fredj:
I'm starting to work on classes in `ol/geom/`

ahocevar:
I'm taking on classes in `ol/events`.

ahocevar:
Now working on classes in `ol/format/`.

src/ol/control/Attribution.js
```diff
@@ -41,11 +41,13 @@ import {visibleAtResolution} from '../layer/Layer.js';
...
  *
- * @constructor
- * @param {module:ol/control/Attribution~Options=} opt_options Attribution options.
  * @api
  */
 class Attribution extends Control {
+
+  /**
+   * @param {module:ol/control/Attribution~Options=} opt_options Attribution options.
+   */
```
ahocevar: @fredj I think for organizing the documentation, it will be easier when exported constructors also also get an `@api` annotation.

ahocevar:
Done with `ol/format/`, moving on to `ol/format/filter/`

tschaub:
I'll start on the transforms for running tests and publishing the package.

ahocevar:
Now moving on to `ol/interaction/`.

ahocevar:
Moving on to `ol/layer/*`

tschaub:
New/updated scripts:
```
npm run build-package # generates build/ol for publishing the ol package
npm run build-legacy # generates build/ol.js et al. for the full build
```

tschaub:
I'll work through `ol/pointer` now.

tschaub:
And `ol/renderer/canvas`.

tschaub:
I wrote [a codemod](https://gist.github.com/tschaub/66f1ad6ef6e130da7d3bb317d7e05a74) to handle the replacement of `inherits` with `extends`, replacement of `Parent.call()` with `super()`, and movement of the constructor comment.  So I'll do the remaining cleanup now.

tschaub:
A few things about the codemod:

 * We lost some "type casts" in super calls (e.g. `Parent.call(this, /** @type {foo} */ (options))` became `super(options)`).  I'm 100% fine with this since we will be reworking things more for TypeScript.
 * We are left with some use of `this` before `super()`.  These will require manual cleanup.
 * We are left with unused `inherits` imports.  These require manual cleanup.

The linter catches the last two remaining chunks of work.  I've updated the TODOs at the top of this PR to reflect what work remains.

tschaub:
I'll remove unused `inherits` imports now.

tschaub:
I'm going to sleep now.

fredj:
I'm removing the unneeded `@struct` and `@constructor` tags

ahocevar:
I'm resolving the `this` before `super()` issues now.

ahocevar:
Working on the tests now.

ahocevar:
Tests fixed, now modifying build config to make them work in IE11 too.

ahocevar:
Looks like this is complete now 🎉 .

Issue #8394, opened 2018-07-18
ol/interaction/Draw/createRegularPolygon is not support odd number or Bug?


iamplex:
I'm use createRegularPolygon with odd number param
[http://jsfiddle.net/zy_plex/kzac216p/](http://jsfiddle.net/zy_plex/kzac216p/)

![draw](https://user-images.githubusercontent.com/11267832/42866803-2e0dfbae-8aa0-11e8-8760-ab8eb909c2db.gif)
you can see the mouse point sometimes would leave the sketch

can fix that?

ahocevar:
Pull request welcome. The mouse point is on the circle that all regular polygon vertices are on. A fix would change the start angle of the regular polygon to match the sketch.

iamplex:
I'd gladly to fix this issues. It's done, now start angle always match the sketch

PR #8396, opened 2018-07-18
Fix ol.interaction.Draw~createRegularPolygon

iamplex:
fix [#8394](https://github.com/openlayers/openlayers/issues/8394)

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -956,19 +956,12 @@ export function createRegularPolygon(opt_sides, opt_angle) {
       squaredCoordinateDistance(center, end));
     const geometry = opt_geometry ? /** @type {module:ol/geom/Polygon} */ (opt_geometry) :
       fromCircle(new Circle(center), opt_sides);
-    let angle = opt_angle ? opt_angle :
-      Math.atan((end[1] - center[1]) / (end[0] - center[0]));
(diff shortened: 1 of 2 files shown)


ahocevar:
Can you add tests for this please? Also to prove that it works with all cases of 2*n+1 edges?

iamplex:
For prove that it works with all cases of 2*n+1 edges, just make sure the sketch of first coordinate(or last coordinate) always equals the mouse point coordinate, but i don't know how to get the mouse position in test case.

Could you provide an idea for this tests? I have no experience in unit test, but i still happy and learn mocha for that

ahocevar:
Thanks for looking into tests. To get the pointer coordinate, you can modify the existing `simulateEvent` function to return the simulated event. Then you can do something like
```js
const event = simulateEvent('pointermove', 80, 80);
const pointerCoordinate = map.getEventCoordinate(event);
```

iamplex:
Sorry, i try to add test case but always fail. Other problem, in local when i run `npm run test` or `npm run karma` it always build success, so i can't debug, that's why there are so many commit.

I think i need to take some time to learn unit tests. Thank you.

ahocevar:
I took the liberty to add tests and simplify the calculation you introduced.

iamplex:
Nice simplify!

PR #8409, opened 2018-07-20
Add abstract translate to ol/geom/Geometry

pjeweb:
This seems to be missing even though both subclasses (SimpleGeometry and GeometryCollection) implement it and SimpleGeometry even tries to @inheritDoc.

### DIFF
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -153,6 +153,15 @@ class Geometry extends BaseObject {
   scale(sx, opt_sy, opt_anchor) {}
 
   /**
+   * Translate the geometry. This modifies the geometry coordinates in place.
+   * @abstract
+   * @param {number} deltaX Delta X.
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks @pjeweb.

fredj:
The function was already present in Geometry, we currently have two `translate` functions:
* https://github.com/openlayers/openlayers/blob/c0f44f6d9d49710396443ac4b795c8c1f0218424/src/ol/geom/Geometry.js#L155-L162
* https://github.com/openlayers/openlayers/blob/c0f44f6d9d49710396443ac4b795c8c1f0218424/src/ol/geom/Geometry.js#L215-L222

pjeweb:
I' sorry for that. Might have been a bit too hot in my office … I was sure my IDE could not find it.

Can you revert my commit?

ahocevar:
@pjeweb Reverted in #8452.

fredj:
@pjeweb no problem

PR #8422, opened 2018-07-22
Improve JSDoc such that `ng build --prod` with angular/cli 6.0.8 succeeds again

### LINKED ISSUE #8421: JSDoc error when compiling with Angular CLI 6.0.8
When trying to compile a project using `ng build --prod` with Angular CLI 6.0.8 and OpenLayers 5.0.0 and the project includes for example Feature.js using `import Feature from 'ol/Feature';`, the following error occurs:

`Module build failed: TypeError: Cannot read property 'kind' of undefined`

I've debugged the error back to the line:

` * @param {(module:ol/format/Feature~WriteOptions|module:ol/format/Feature~ReadOptions)=} opt_options`

When changing this line to

` * @param {module:ol/format/Feature~WriteOptions|module:ol/format/Feature~ReadOptions=} opt_options`

the error doesn't occur.

jkoelewijn:
Fixes #8421.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -61,7 +61,7 @@ import {create as createTransform, apply as applyTransform} from './transform.js
 
 /**
  * @typedef {Object} AtPixelOptions
- * @property {((function(module:ol/layer/Layer): boolean)|undefined)} layerFilter Layer filter
+ * @property {function(module:ol/layer/Layer): boolean|undefined} layerFilter Layer filter
  * function. The filter function will receive one argument, the
  * {@link module:ol/layer/Layer layer-candidate} and it should return a boolean value.
  * Only layers which are visible and for which this function returns `true`
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -61,7 +61,7 @@ import {create as createTransform, apply as applyTransform} from './transform.js
 
 /**
  * @typedef {Object} AtPixelOptions
- * @property {function(module:ol/layer/Layer): boolean|undefined} layerFilter Layer filter
+ * @property {undefined|function(module:ol/layer/Layer): boolean} layerFilter Layer filter
  * function. The filter function will receive one argument, the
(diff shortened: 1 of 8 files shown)


src/ol/format/Feature.js
```diff
@@ -200,7 +200,7 @@ export default FeatureFormat;
 /**
  * @param {module:ol/geom/Geometry|module:ol/extent~Extent} geometry Geometry.
  * @param {boolean} write Set to true for writing, false for reading.
- * @param {(module:ol/format/Feature~WriteOptions|module:ol/format/Feature~ReadOptions)=} opt_options
+ * @param {module:ol/format/Feature~WriteOptions|module:ol/format/Feature~ReadOptions=} opt_options
```
ahocevar: This one should probably become
```js
 * @param {(module:ol/format/Feature~WriteOptions=|module:ol/format/Feature~ReadOptions=} opt_options
```

src/ol/PluggableMap.js
```diff
@@ -61,7 +61,7 @@ import {create as createTransform, apply as applyTransform} from './transform.js
 
 /**
  * @typedef {Object} AtPixelOptions
- * @property {((function(module:ol/layer/Layer): boolean)|undefined)} layerFilter Layer filter
+ * @property {function(module:ol/layer/Layer): boolean|undefined} layerFilter Layer filter
```
tschaub: This makes the type ambiguous.  The type could be a function or undefined.  As it is written here, the function could return a boolean or undefined.

ahocevar: Right. In this case you‘d have to put `undefined` first, and then the function.

jkoelewijn:
Thanks for your comments. I've added some commits to improve the pull request.

tschaub:
Thanks @jkoelewijn.  We're working toward using TypeScript for type checking (see #8345).  I'm hoping that is what the Angular CLI uses too.  Otherwise, I think it will be hard to maintain a type syntax that doesn't break the `ng` tool.

Issue #8425, opened 2018-07-23
Cancel old image source requests


josepgomes:
Hello,
I noticed that in the recent ol versions wms requests are managed differently. In the 2.10 version, if i zoom in and out, the previous wms requests are canceled.

![image](https://user-images.githubusercontent.com/3487504/43063731-909c545c-8e54-11e8-85e1-9286a6306287.png)

Nowadays, every single request is processed.

![image](https://user-images.githubusercontent.com/3487504/43063775-c0e3a1e2-8e54-11e8-8963-48fec34c8391.png)

This leads to a very low performance.

Is it possible to cancel a wms request when we already ask for another one?

Many Thanks!

ahocevar:
OpenLayers now uses a TileQueue, which continues to load tiles until its tile cache is saturated. You can enforce that images are dropped when you configure the `source/TileWMS` with a small `cacheSize`, e.g.
```js
new TileWMSSource({
  cacheSize: 16,
  // ...
});
```

josepgomes:
This is true for ImageWMS, too?

ahocevar:
No, that's only the case for tiles. For ImageWMS, you could achieve that with a custom `imageLoadFunction` on the source, e.g.
```js
const blankSrc = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=';
let oldImg;
new ImageWMSSource({
  imageLoadFunction: function(image, src) {
    if (oldImg) {
      oldImg.src = blankSrc;
    }
    const img = image.getImage();
    img.addEventListener('load', function() {
      oldImg = undefined;
    });
    img.src = src;
  }
});
```

josepgomes:
@ahocevar , i tested that solution, but the performance is still very bad, taking some times more than 5 seconds to completely load all the requests. This is because the browser will process all the "ok" status requests. Is there no way to abort requests like in ol2? Thanks for your time!

josepgomes:
I tested the abort state, but for the browser is still a 200 status. It seems that there isn't anything that really cancels the request, right?

josepgomes:
Can you tell me where is the request made? Did not found any xhr doing the request. Thanks.

ahocevar:
It‘s just setting the src on the Image, no XHR. You‘d have to combine setting the ABORT state with the snippet I posted above.

[...]

PR #8430, opened 2018-07-24
Cancel old image source requests

josepgomes:
Header line:  Cancel old image source requests

When we navigate using images, instead of tiles, every request that we make is processed, even when we already made a new request. Old requests are not being aborted. With this changes, old requests are canceled.


josepgomes:
https://github.com/openlayers/openlayers/issues/8425

ahocevar:
This does not look like the right fix. Can you please test #8432 to see if that also works for you?

PR #8432, opened 2018-07-25
Abort image loading for image layers

### LINKED ISSUE #8425: Cancel old image source requests

ahocevar:
This is an alternative solution to #8430.

Fixes #8425.

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -36,8 +36,9 @@ class ImageWrapper extends ImageBase {
    * @param {string} src Image source URI.
    * @param {?string} crossOrigin Cross origin.
    * @param {module:ol/Image~LoadFunction} imageLoadFunction Image load function.
+   * @param {HTMLImageElement=} opt_image Image element
    */
-  constructor(extent, resolution, pixelRatio, src, crossOrigin, imageLoadFunction) {
+  constructor(extent, resolution, pixelRatio, src, crossOrigin, imageLoadFunction, opt_image) {
 
     super(extent, resolution, pixelRatio, ImageState.IDLE);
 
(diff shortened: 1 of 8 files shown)


josepgomes:
It seems is not working as expected. Did it worked in your tests?

ahocevar:
@josepgomes Can you provide a test case that you expect to pass?

josepgomes:
Hi,
I can't, but i can explain the problem and my current solution (for the product that i'm working for).
The problem is simple, when zooming, using images not tiles, the browser makes a wms request for each new zoom, because the image.SRC was updated.

The problem is that when we zoom multiple times, SRC is updated multiple times, and the browser will process every single request, even when we just need the last one.

My current solution is set a custom imageLoadFunction:

```typescript
private previousImageElement: HTMLImageElement;

const source = this.getSource() as ol.source.ImageWMS;
source.setImageLoadFunction(this.imageLoadFunction.bind(this));

private imageLoadFunction(image, src) {
    this.previousImageElement.src = src;
    this.previousImageElement.onload = () => {
        image.getImage().src = this.previousImageElement.src;
    };
}
```

With this solution loading times are fine. Hope you can understand and implement this ideia.

ahocevar:
Tell me if I'm missing something. To me it looks like your solution debounces image loading, but in a weird way (i.e. by loading the new url in the previous image element, and then when loaded loading it again in the current image element). In the end, your solution also sets an image src for every zoom level, but it does so on the previous image element.

josepgomes:
I don't fully understand why this works, but i tested your changes and they don't abort any request. I will try to figure this out.

ahocevar:
I don't think the extra code justifies the outcome. Closing.

PR #8437, opened 2018-07-26
Avoid shadowing EventTarget

tschaub:
The `npm run typecheck` script in #8345 revealed that our `EventTarget` class shadows the DOM `EventTarget` type.  This branch changes our class to `Target` to differentiate.

### DIFF
--- a/src/ol/ImageBase.js
+++ b/src/ol/ImageBase.js
@@ -1,7 +1,7 @@
 /**
  * @module ol/ImageBase
  */
-import EventTarget from './events/EventTarget.js';
+import EventTarget from './events/Target.js';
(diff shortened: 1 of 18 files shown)


Rolandocepeda:
this currently errors out when running build with --prod. I think the import statements should have been
`import Target from './events/Target.js';`

ahocevar:
This should not be the case, because `import EventTarget from './events/Target.js'` imports a default export, which you can assign to any local name. In which context are you seeing this issue?

tschaub:
The problem was in the EventTarget.js module.  The EventTargetLike type was EventTarget (the DOM one) or EventTarget (our one).  And ours shadowed the real one.

PR #8452, opened 2018-07-27
Remove extra translate function in Geometry, add missing api tag

fredj:
see https://github.com/openlayers/openlayers/pull/8409#issuecomment-408389061


bartvde:
Should we add https://eslint.org/docs/rules/no-dupe-class-members to our eslint configuration?

fredj:
> Should we add https://eslint.org/docs/rules/no-dupe-class-members to our eslint configuration?

See https://github.com/openlayers/eslint-config-openlayers/pull/14

PR #8459, opened 2018-07-28
Update has.js

josephzhao:
&& typeof navigator.userAgent !== 'undefined'


tschaub:
Also curious to hear where this comes up as an issue.  No guarantee that `userAgent` is a string either (meaning if we were being pedantic we would check for a `toLowerCase` method as well before calling it).

fredj:

PR #8463, opened 2018-07-30
Handle mapBrowserEvent without pointerEvent in 'primaryAction' condition

### LINKED ISSUE #8011: Setting ol.events.condition.primaryAction for a select interaction breaks with mousewheel events.

To keep this project manageable for maintainers, we ask you to please check all boxes below before submitting an issue.

- [X] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

Here's my fiddle http://jsfiddle.net/6frLm7vk/

I've added the select interaction (even though there's nothing to be selected). However whenever I use the mouse wheel it raises a JS error.

I've tested this on Chrome not sure about other browsers.

Error raised in ol-debug.js line 20838:
```
ol.events.condition.primaryAction = function(mapBrowserEvent) {
  var pointerEvent = mapBrowserEvent.pointerEvent; 
  return pointerEvent.isPrimary && pointerEvent.button === 0;
};
```
It seems like wheel events don't have a pointerEvent object in Chrome . Not sure if this is only chrome or in general. It makes sense that they don't have the pointerEvent because the mouse wheel is not really covered in the pointer event spec so probably adding an undef check would help here?

fredj:
fixes #8011

### DIFF
--- a/src/ol/events/condition.js
+++ b/src/ol/events/condition.js
@@ -245,5 +245,5 @@ export const mouseOnly = function(mapBrowserEvent) {
  */
 export const primaryAction = function(mapBrowserEvent) {
   const pointerEvent = mapBrowserEvent.pointerEvent;
-  return pointerEvent.isPrimary && pointerEvent.button === 0;
+  return pointerEvent !== undefined && pointerEvent.isPrimary && pointerEvent.button === 0;
 };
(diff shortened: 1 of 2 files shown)


src/ol/events/condition.js
```diff
@@ -245,5 +245,5 @@ export const mouseOnly = function(mapBrowserEvent) {
  */
 export const primaryAction = function(mapBrowserEvent) {
   const pointerEvent = mapBrowserEvent.pointerEvent;
-  return pointerEvent.isPrimary && pointerEvent.button === 0;
+  return pointerEvent !== undefined && pointerEvent.isPrimary && pointerEvent.button === 0;
```
tschaub: I wonder if we'll want `instanceof MapBrowserPointerEvent` here instead for type checking.  Also curious about the performance compared to this duck typing.

PR #8464, opened 2018-07-30
[Needs review] Implemented ol.layer.Vector.setRenderMode() (fixes #8436)

### LINKED ISSUE #8436: [FeatureRequest]: add ol.layer.Vector.setRenderMode('image' || 'vector')
`ol.layer.Vector `allows to define the `renderMode` in the constructor.
However, the `renderMode` cannot be changed (or I overlooked something) after constructing a vector.

For performance reason, I would like to change the `renderMode` to `image` as sometimes (depending on the amount of features) the performance might be insufficient while in general `vector` works.
So, I would like to let the user change the `renderMode` depending on his needs _after_ loading the features.

_Working solutions:_
At the moment, the solution be to recreate the `vector` (including coping all other settings) and replace it in the layergroup.

_Desired solution:_
Adding `ol.layer.Vector.setRenderMode('image' || 'vector')`

_My workaround:_ creating new `ol.layer.Vector`, copying all (relevant) attributes, and set renderMode accordingly.

```
    var collectionLayers = this.getMap().getLayerGroup().getLayers();

    for(let i = 0; i < collectionLayers.getLength(); i++) {
        let layer = collectionLayers.item(i);

        if (layer instanceof ol.layer.Vector) {
            let layerNew = new ol.layer.Vector({
                source: layer.getSource(),
                style: layer.getStyle(),
                renderMode: this._shouldBeFast ? "image" : "vector" //TODO should use ol.layer.VectorRenderType
            });

            collectionLayers.setAt(i, layerNew);
        }
    }
```

dennisguse:
Added `ol.layer.Vector.setRenderModer()` that allows changing the `renderMode` after a `VectorLayer` was created.
In addition, `ol.renderer.Map` was extended to listen for changes of `ol.layer.Layer`.
On change the renderer for this layer is re-created.

__ATTENTION:__
`ol.renderer.Map` listens to `EventType.CHANGE`.
Therefore, _any_ change of `ol.layer.Layer` will trigger re-creating the renderer.
As this also includes adding new features, this patch might have _serious_ implications on performance.

__Solution:__
Listen to a specific event (e.g., Property.RENDER_MODE) and thus avoid unnecessary re-creating the renderer.

### DIFF
--- a/src/ol/layer/Vector.js
+++ b/src/ol/layer/Vector.js
@@ -73,7 +73,8 @@ export const RenderType = {
  * @private
  */
 const Property = {
-  RENDER_ORDER: 'renderOrder'
+  RENDER_ORDER: 'renderOrder',
+  RENDER_MODE: 'renderMode'
 };
 
 
@@ -145,11 +146,7 @@ class VectorLayer extends Layer {
     this.updateWhileInteracting_ = options.updateWhileInteracting !== undefined ?
       options.updateWhileInteracting : false;
 
-    /**
-    * @private
-    * @type {module:ol/layer/VectorTileRenderType|string}
-    */
-    this.renderMode_ = options.renderMode || VectorRenderType.VECTOR;
+    this.setRenderMode(options.renderMode || VectorRenderType.VECTOR);
 
     /**
     * The layer type.
@@ -256,20 +253,19 @@ class VectorLayer extends Layer {
   * @return {module:ol/layer/VectorRenderType|string} The render mode.
   */
   getRenderMode() {
-    return this.renderMode_;
+    return (
+    /** @type {module:ol/render~OrderFunction|null|undefined} */ (this.get(Property.RENDER_MODE))
+    );
   }
 
   /**
    * @param {module:ol/layer/VectorRenderType|string} renderMode The render mode.
    */
+
   setRenderMode(renderMode) {
-    if (renderMode === this.renderMode_) {
-      return;
+    if (renderMode === VectorRenderType.VECTOR || renderMode === VectorRenderType.Image) {
+      this.set(Property.RENDER_MODE, renderMode);
     }
-
-    this.renderMode_ = renderMode;
-
-    this.changed();
   }
 }
 
--- a/src/ol/layer/Vector.js
+++ b/src/ol/layer/Vector.js
@@ -258,6 +258,19 @@ class VectorLayer extends Layer {
   getRenderMode() {
     return this.renderMode_;
   }
+
+  /**
+   * @param {module:ol/layer/VectorRenderType|string} renderMode The render mode.
+   */
+  setRenderMode(renderMode) {
+    if (renderMode === this.renderMode_) {
+      return;
+    }
+
+    this.renderMode_ = renderMode;
+
+    this.changed();
+  }
 }
 
 
--- a/src/ol/renderer/Map.js
+++ b/src/ol/renderer/Map.js
@@ -28,6 +28,12 @@ class MapRenderer extends Disposable {
 
     /**
      * @private
+     * @type {!Object.<string, module:ol/layer/Layer>}
+     */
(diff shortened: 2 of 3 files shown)


dennisguse:
_NOTE:_ Testing fails, because the renderers are recreated.
This would be fixed automatically by listening to a specific event.

ahocevar:
Thanks for your effort on this @dennisguse. I took the liberty to add a commit that makes things work in a simpler way, without requiring additional event listeners, and without breaking tests. Would you be able to add tests for the new functionality?

dennisguse:
I love the approach: cleaner and less code.
But there is a slight issue that does not occur using the event-based solution.
The map is not immediately updated and if zoomed/panned/rotated, the marker is removed and re-appears.
I attached two videos (as ZIP): _changedEvent.webm_ and _property.webm_ showing the difference between the two solutions.

(Without any debugging; it's too late):
The issues with the _property_ feels like a race condition.
- the layer is _not_ immediately flagged as dirty and thus not re-rendered
- renderer is removed while panning and thus the maker removed
- as `getUpdateWhileAnimating() == false`, the update takes place _after_ zooming/panning/rotating

Test case:
1. Load KML (via DnD) using `defaultStyle: true`
```
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://www.opengis.net/kml/2.2">
  <Placemark>
    <name>Simple placemark</name>
    <description>Attached to the ground. Intelligently places itself 
       at the height of the underlying terrain.</description>
    <Point>
      <coordinates>-122.0822035425683,37.42228990140251,0</coordinates>
    </Point>
  </Placemark>
</kml>
```

2. Rotate via `ol.interaction.DragRotateAndZoom()`
__RESULT:__ Switching between 'image' and 'vector' (and vice versa): the rotation of the text is either fixed (to the map) or rotating (and thus readable).

ahocevar:
The behavior you're observing with my updated code is intentional - changing renderers is better done lazily, because the map image won't change anyway (except for rotated texts). If you need an immediate update, you can call `map.render()` or `map.renderSync()` after calling `layer.setRenderMode()`.

dennisguse:
Somehow the upload did not work - here is a GIF of the property solution.
It shows the "flickering" of the marker, as the renderer is removed (and thus the complete visualization) and the new renderer is added after the panning stop (`getUpdateWhileAnimating() == false`).
And from an API perspective - I would find it confusing to additionally call `map.render()` as I would think it would be implied...
Can we somehow avoid removing the rendered visualization, but replace the renderer?

output

ahocevar:
> Can we somehow avoid removing the rendered visualization, but replace the renderer?

You could do that on the application level, by creating two layers from the same source and switching their `visible` property.

ahocevar:
As I said already, I'm not convinced that setRenderer should immediately change the rendering, and I pointed out how you can achieve that on the application level if needed (by calling `changed()`). I don't think we should continue working in the direction this pull request is going now. In addition to the failing tests, this solution might introduce a memory leak when a layer is removed from the map.

dennisguse:
Null problemo.
Then I will stick with my solution to remove, recreate, and re-add the layers. (see https://github.com/openlayers/openlayers/issues/8436)
This solution also immediately changes the renderer and thus does not incur any flickering.

@ahocevar Thanks for your time.
And you are right: this pull request might incur memory leak or even a race condition.

PS: Their is actually only one test failing - the others fail because the tests are not independent of each other (spy.restore() is not called if a test fails).
This is the failing test:

 ```
     layer.renderMode_ = 'image';
      const spy = sinon.spy(CanvasVectorTileLayerRenderer.prototype,
        'getTransform');
      map.renderSync();
      expect(spy.callCount).to.be(0);
      spy.restore();
```
What does this test actually do?
With my pull request, `getTransform()` is called always (I checked by calling `map.renderSync()`multiple times.

Issue #8404, opened 2018-07-19
Tiles stop loading


scroach:
We've got an App which uses Openlayers in a Cordova Environment no iOS.
Our current problem is that tiles stop loading "after some time". What I've found after hours of testing: it seems to be related to tile loading errors. After 16(?) errors the map just completely stops trying to load tiles from any layer.

So I found these two issues/changes: #6969 #6972
The seem to match my problem, but it says they are fixed?
I've tried different versions of Openlayers from 4.3.4 to 4.6.5 a the most recent 5.0.3

The error is present in all versions and I have no idea what to do anymore.
Any help would be greatly appreciated.

ahocevar:
Are you using image or vector tiles. If the former: what kind of loading errors are you experiencing? Everything that triggers an `error` event on the `Image` element of the tile will cause the tile to be unqueued with an `ERROR` state. So there must be something else going on.

scroach:
@ahocevar thank for the quick reply.
The app is using image tiles which are stored locally on the device. I just know that I can see the errors in the safari console when connected to the ipad.

How could I check what happens to the queue or why those tiles are not unqueued? Any hints on setting good breakpoints?

scroach:
@ahocevar alright... I've got more infos. I think I just found a (hacky?) solution to our problem.
Because you said the tile should be unqueued on errors so I thought maybe Openlayers doens't get the info that there has been an error.

I've looked into it and my collegue has implemented a custom `tileLoaderFunction` which loads and transforms the url to a blob url.
````
tileLoadFunction: function(imageTile, src) {

                            var xhr = new XMLHttpRequest();
                            xhr.responseType = "blob";
                            console.log(imageTile);
                            xhr.addEventListener("loadend", function (evt) {
                                var data = this.response;
                                if (data !== undefined) {
                                    var url = window.URL.createObjectURL(data);
                                    imageTile.getImage().src = url;
                                }
                            });
                            xhr.addEventListener('error', function () {
                                totalErrors++;
                                console.error(totalErrors + ' - error loading xhr tile: ' + src.substr(30));

                                // set error state for tile
                                imageTile.setState(3);

                            });
                            xhr.open("GET", src);
                            xhr.send();
                        }
````

I've just added the error handler, there never was any. And I've set the Tile state to 3=ERROR - this seems to work... BUT: is it the right way, is it hacky or is there any better solution? :D

ahocevar:
[...]

PR #8470, opened 2018-08-01
Update Tile loading API docs

scroach:
Updated API docs according to #8404


ahocevar:
Thanks for this doc improvement! I took the liberty to refine it a bit, and I think it looks great now.

PR #8471, opened 2018-08-01
Add getUrl and getImageExtent to ImageStatic API

samuel-girard:
Hello,

I added 2 new methods to ImageStatic API:
- getUrl
- getImageExtent

The reason why is that we need these for ol-cesium to synchronize ImageStatic layers (https://github.com/openlayers/ol-cesium/issues/458)

### DIFF
--- a/src/ol/source/ImageStatic.js
+++ b/src/ol/source/ImageStatic.js
@@ -38,8 +38,6 @@ class Static extends ImageSource {
    * @param {module:ol/source/ImageStatic~Options=} options ImageStatic options.
    */
   constructor(options) {
-    const imageExtent = options.imageExtent;
-
     const crossOrigin = options.crossOrigin !== undefined ?
       options.crossOrigin : null;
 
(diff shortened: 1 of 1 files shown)


src/ol/source/ImageStatic.js
```diff
@@ -52,11 +50,23 @@ class Static extends ImageSource {
...
+     * @type {module:ol/extent~Extent}
+     */
+    this.imageExtent_ = options.imageExtent;
+
     /**
      * @private
      * @type {module:ol/Image}
      */
-    this.image_ = new ImageWrapper(imageExtent, undefined, 1, options.url, crossOrigin, imageLoadFunction);
+    this.image_ = new ImageWrapper(this.imageExtent_, undefined, 1, options.url, crossOrigin, imageLoadFunction);
```
fredj: minor: use `this.url_` instead of `options.url` here

samuel-girard: Thanks for your feedback

PR #8481, opened 2018-08-03
Expose some internal functions

gberaudo:
In openlayers/ol-cesium/pull/622 I am creating an example of legacy usage of the `OL-Cesium` library.
Instead of bundling `OL-Cesium` and `OpenLayers` together with webpack, the `OpenLayers` library is used standalone.

It does not work out-of-the-box though since `OL-Cesium` is using some internal` OpenLayers` APIs and the `tasks/generate-index.js` script is limiting the exported symbols to the ones marked `@api`.

In this PR 2 functions are exposed: `getUid` and `stableSort` which I think are better exposed in `OpenLayers` than reimplemented in `OL-Cesium`. With these functions exposed `OL-Cesium` is able to handle a basic map.

My goal is to provide a basic example working out-of-the-box so that people interested in this usage of `OL-Cesium` and `OpenLayers` can have an official starting point.

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -194,8 +194,10 @@ export function equals(arr1, arr2) {
 
 
 /**
+ * Sort the passed array such that the relative order of equal elements is preverved.
+ * See https://en.wikipedia.org/wiki/Sorting_algorithm#Stability for details.
  * @param {Array<*>} arr The array to sort (modifies original).
- * @param {Function} compareFnc Comparison function.
+ * @param {!function(*, *): number} compareFnc Comparison function.
  * @api
  */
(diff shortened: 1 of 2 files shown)


src/ol/array.js
```diff
@@ -196,6 +196,7 @@ export function equals(arr1, arr2) {
 /**
  * @param {Array<*>} arr The array to sort (modifies original).
  * @param {Function} compareFnc Comparison function.
+ * @api
```
tschaub: This should have some documentation if it is going to be part of the API.

gberaudo: OK, I added a description and specified the signature of the comparison function.

PR #8487, opened 2018-08-06
Add new getExtent option for Graticule

### LINKED ISSUE #8427: Add 'extent' to Graticule options
**Description**
It appears that the `ol/Graticule` inherits its extent from the map's projection extent. I am proposing to add an `extent` option (and `setExtent`/ `getExtent` methods) with the default continuing to be the projection extent.

**Reason for request**
I would like to apply layer-swipe like functionality to graticules.
https://openlayers.org/en/latest/examples/layer-swipe.html

fredj:
Adds a new `getExtent` option to the Graticule. This function takes the view extent as argument and can return a new extent. The graticule is rendered into this extent.

fixes #8427 

comments welcome

### DIFF
--- a/examples/graticule.js
+++ b/examples/graticule.js
@@ -5,6 +5,7 @@ import TileLayer from '../src/ol/layer/Tile.js';
 import {fromLonLat} from '../src/ol/proj.js';
 import OSM from '../src/ol/source/OSM.js';
 import Stroke from '../src/ol/style/Stroke.js';
+import {getWidth} from '../src/ol/extent.js';
 
 
 const map = new Map({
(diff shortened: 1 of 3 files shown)


src/ol/Graticule.js
```diff
@@ -71,6 +71,9 @@ const INTERVALS = [
  * @property {number} [latLabelPosition=1] Latitude label position in fractions
  * (0..1) of view extent. 0 means at the left of the viewport, 1 means at the
  * right.
+ * @property {function(module:ol/extent~Extent):module:ol/extent~Extent} [getExtent] An optional function
+ * that returns the extent where the graticule will be displayed.
+ * The function receives the view extent as argument.
```
ahocevar: It doesn't look like it does, and I think it's not needed.

src/ol/Graticule.js
```diff
@@ -71,6 +71,9 @@ const INTERVALS = [
  * @property {number} [latLabelPosition=1] Latitude label position in fractions
  * (0..1) of view extent. 0 means at the left of the viewport, 1 means at the
  * right.
+ * @property {function(module:ol/extent~Extent):module:ol/extent~Extent} [getExtent] An optional function
```
ahocevar: Better to name the option `extent`, and accept a fixed extent in addition to a function.

examples/graticule.js
```diff
@@ -22,8 +23,21 @@ const map = new Map({
   })
 });
 
+const swipe = document.getElementById('swipe');
```
ahocevar: I think it is better to keep the example simple instead of adding this feature to it. Better document the option with a sensible example snippet.

src/ol/Graticule.js
ahocevar: If users modify the passed extent in place, things will break wherever the library relies on `frameState.extent`.

fredj:
fixed with  #8944

PR #8489, opened 2018-08-06
Use .prototype. only where necessary

ahocevar:
This removes unnecessary occurrences of `.prototype.`, which we missed in #8385.

### DIFF
--- a/src/ol/Disposable.js
+++ b/src/ol/Disposable.js
@@ -1,13 +1,22 @@
 /**
  * @module ol/Disposable
  */
-import {VOID} from './functions.js';
 
 /**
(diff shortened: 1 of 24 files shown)


src/ol/Disposable.js
```diff
@@ -17,19 +27,12 @@ class Disposable {
...
-/**
- * Extension point for disposable objects.
- * @protected
- */
-Disposable.prototype.disposeInternal = VOID;
+  /**
+   * Extension point for disposable objects.
+   * @protected
+   */
+  disposeInternal() {}
```
fredj: `VOID` import can be removed

PR #8490, opened 2018-08-07
WMTS getCapabilities readCoodinates more than one whitespace delimiter

MarquesDeAzevedo:
Some WMTS getCapabilities return <MatrixSet><TopLeftCorner> with more than one whitespace as delimiter between the coordinates. This leads to the layer not being displayed, because the origin_ property of ol.tilegrid.TileGrid is not set.
Error in ol.tilegrid.TileGrid.prototype.getTileCoordForXYAndZ_: JavaScript runtime error: Unable to get property '0' of undefined or null reference

### DIFF
--- a/src/ol/format/WMTSCapabilities.js
+++ b/src/ol/format/WMTSCapabilities.js
@@ -340,7 +340,7 @@ function readLegendUrl(node, objectStack) {
  * @return {Object|undefined} Coordinates object.
  */
 function readCoordinates(node, objectStack) {
-  const coordinates = readString(node).replace(/\s\s+/g, ' ').split(' ');
+  const coordinates = readString(node).split(/\s+/g);
   if (!coordinates || coordinates.length != 2) {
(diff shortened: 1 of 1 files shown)


MarquesDeAzevedo:
Thanks for the good advice. I updated the code.

src/ol/format/WMTSCapabilities.js
```diff
@@ -340,7 +340,7 @@ function readLegendUrl(node, objectStack) {
  * @return {Object|undefined} Coordinates object.
  */
 function readCoordinates(node, objectStack) {
-  const coordinates = readString(node).split(' ');
+  const coordinates = readString(node).split(/\s+/g);
```
ahocevar: The regular expression `/\s+/` (without the `g` flag) should be sufficient.

MarquesDeAzevedo:
Thanks again. I updated the code.

PR #8499, opened 2018-08-08
Round center in viewState to pixels

### LINKED ISSUE #8498: ol5: render changes features positions
Hi,
We changed openlayers version from 4.6.5 to 5.1.3.
We noticed that when first time map is rendered (with 5.1.3 version) the view looks exactly same than with old version (4.6.5) of openlayers. But when map is rendered again some features moves little bit on the screen. I made some debugging and look source code and I found the reason:
In PluggableMap.renderFrame function in version 4.6.5 there is:
```
    viewState = view.getState();
    var center = viewState.center;
    var pixelResolution = viewState.resolution / this.pixelRatio_;
    center[0] = Math.round(center[0] / pixelResolution) * pixelResolution;
    center[1] = Math.round(center[1] / pixelResolution) * pixelResolution;
```
and in 5.1.3
```
      viewState = view.getState();
      let focus = this.focus_;
      if (!focus) {
        focus = viewState.center;
        const pixelResolution = viewState.resolution / this.pixelRatio_;
        focus[0] = Math.round(focus[0] / pixelResolution) * pixelResolution;
        focus[1] = Math.round(focus[1] / pixelResolution) * pixelResolution;
      }
```
So in 4.6.5 version, rounding of 'center' is done always but in 5.1.3 version only on first time.

ahocevar:
Fixes #8498.

In addition to fixing the rounding regression, I propose to change the `ol/style/Image`'s `snapToPixel` option default from `true` to `false`, to avoid jitter during kinetic pan. I'll also introduce a `snapToPixel` option for `ol/style/Text`.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,15 @@
 ## Upgrade notes
 
+### Next version
+
+#### Removal of the `snapToPixel` option for `ol/style/Image` subclasses
+
+The `snapToPixel` option has been removed, and the `getSnapToPixel` and `setSnapToPixel` methods are deprecated.
+
+The renderer now snaps to integer pixels when no interaction or animation is running to get crisp rendering. During interaction or animation, it does not snap to integer pixels to avoid jitter.
+
+When rendering with the Immediate API, symbols will no longer be snapped to integer pixels. To get crisp images, set `context.imageSmoothingEnabled = false` before rendering with the Immediate API, and `context.imageSmoothingEnabled = true` afterwards.
+
 ### v5.1.0
 
 #### Geometry constructor and `setCoordinates` no longer accept `null` coordinates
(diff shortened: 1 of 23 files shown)


tschaub:
It could be nice to snap to pixel only when not animating or interacting.  This would keep people from having to make the choice between crisp and jittery or blurry and smooth.

ahocevar:
The `snapToPixel` option has now been removed, as suggested by @tschaub in 
https://github.com/openlayers/openlayers/pull/8499#issuecomment-411619512. To make this look nice, I also modified the DragPan interaction to not trigger a re-render with `interacting` flag set on down/up without dragging.

ahocevar:
Thanks in advance for any review.

ahocevar:
The fix is in v5.2.0.

PR #8503, opened 2018-08-09
Avoid block scope issues in transpiled code

### LINKED ISSUE #8500: view.animate callback function is buggy (read: broken)
Hello it's me again (sorry)

I have some trouble with the view.animate function in combination with a callback.

The specs say that:
> If you provide a function as the last argument to the animate method, it will get called at the end of an animation series.

but when i have this example, i get a console error.
https://jsfiddle.net/29ps8xwa/14/

This is because the callback is used from within a 'setTimeout' function in the View.js, like so:
```javascript
const callback = series[0].callback;
        if (callback) {
          setTimeout(function() {
            callback(true);
          }, 0);
        }
```

edit:
> there can be only one!

the bug is more visible in this example:
https://jsfiddle.net/29ps8xwa/16/ where i gave both the parallel animations a callback argument.

You can see that both animations trigger.... but both trigger the FIRST callback.. so the second animation triggers the first animations's callback.

In the previous example the first animation didn't have a callback.. the second animation created the timeout function (called setTimeout) but was going to trigger the first animation's callback... which was undefined.. hence the console error..

ahocevar:
Fixes #8500, and makes callback invocation more consistent (i.e. also execute in a new cycle when animations are cancelled.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -428,9 +428,7 @@ class View extends BaseObject {
         this.setRotation(state.rotation);
       }
       if (callback) {
-        setTimeout(function() {
-          callback(true);
-        }, 0);
+        animationCallback(callback, true);
       }
       return;
     }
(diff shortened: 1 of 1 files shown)


ahocevar:
The fix is in v5.2.0.

PR #8504, opened 2018-08-09
Link to Browserify example project

### LINKED ISSUE #8497: using openlayers with browserify
i always used the pre-compiled version of openlayers due to the fact that the entire closure compiler was way to complicated a way to build a simple app.

Now openlayers 5 is here, so i am rebuilding my 'simple app' (a npm package) to 'the new way of life'

I removed all the local files, did an npm install ol and started with this simple piece of code:

```javascript
import Map from 'ol/Map.js';
import TileLayer from 'ol/layer/Tile.js';
import OSM from 'ol/source/OSM.js';

module.exports = function () {
	var map = new Map({
		layers: [
			new TileLayer({
				visible: false,
				preload: Infinity,
				source: new OSM()
			})
		],
		target: 'map',
		view: new View({
			center: [-6655.5402445057125, 6709968.258934638],
			zoom: 13
		})
	});
};
```

now as you can see i use module.exports because we use browserify. Since the new openlayers uses the ES6 style import and export functions, we need to use babel to pre-compile our javascript (including dependencies like openlayers) to Ecmascript 5.

I use this buildscript in my package.json
```
"build": "\"browserify src/js/test.js -t [ babelify --presets [ \"@babel/preset-env\" ] ]",
``` 

and i have the necessary dependencies for babel installed.
Still i get this error: 

d:\mySimpleProject\node_modules\ol\View.js:4
import {DEFAULT_TILE_SIZE} from './tilegrid/common.js';
^
ParseError: 'import' and 'export' may appear only with 'sourceType: module'

I'm sure i'm doing something wrong but i can't get my finger on it. If i goolgle, i only find bugreports on git for other projects... not really helping.

ahocevar:
Fixes #8497.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ See the following examples for more detail on bundling OpenLayers with your appl
  * Using [Rollup](https://github.com/openlayers/ol-rollup)
  * Using [Webpack](https://github.com/openlayers/ol-webpack)
  * Using [Parcel](https://github.com/openlayers/ol-parcel)
+ * Using [Browserify](https://github.com/openlayers/ol-browserify)
 
 ## Supported Browsers
 


bartvde:
We should probably also change this gist by @tschaub https://gist.github.com/tschaub/4bfb209a8f809823f1495b2e4436018e

which reads "Please choose one of the other options for bundling ES Modules." and is found by google search

ahocevar:
No code change, merging for the v5.2.0 release.

PR #8510, opened 2018-08-13
Fix WMTS URLs with dimensions

gberaudo:
Fix a regression introduced in https://github.com/openlayers/openlayers/commit/8991dc9718cfceb51dc531d6d8eda5038715a145.

When the function is evaluated, `this` was not the WMTS source so the `this.dimensions_` property was not found which resulted in an `undefined` dimension value used in the URL string.

It is now fixed and I added a test for the URLs of WMTS sources with dimensions.

### DIFF
--- a/src/ol/source/WMTS.js
+++ b/src/ol/source/WMTS.js
@@ -493,6 +493,7 @@ function createFromWMTSTemplate(template) {
     });
 
   const tileGrid = this.tileGrid;
+  const dimensions = this.dimensions_;
 
   return (
     /**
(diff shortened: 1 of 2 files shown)


fredj:
@gberaudo instead of an arrow function; I've added a local `dimensions` variable (the same is already done for `tileGrid`)

fredj:
@openlayers/core can you please take a look at this one?

fredj:
Thanks for the review

ahocevar:
The fix is in v5.2.0.

PR #8511, opened 2018-08-13
Update IGN API key

### LINKED ISSUE #8509: IGN WMTS example is broken
https://openlayers.org/en/latest/examples/wmts-ign.html

The server returns HTTP 403 error because the API key has expired:
https://github.com/openlayers/openlayers/blob/a94dff2c0635cce72f2ca6de32fe6805d26c0444/examples/wmts-ign.js#L40-L42

pgiraud:
Fixes #8509
Needs to be tested.

### DIFF
--- a/examples/wmts-ign.js
+++ b/examples/wmts-ign.js
@@ -37,12 +37,11 @@ const tileGrid = new WMTSTileGrid({
   matrixIds: matrixIds
 });
 
-// API key valid for 'openlayers.org' and 'localhost'.
-// Expiration date is 06/29/2018.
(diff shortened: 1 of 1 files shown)


fredj:
It doesn't seems to work; I still have HTTP 403 errors on localhost

pgiraud:
@fredj I found this https://geoservices.ign.fr/blog/2017/06/28/geoportail_sans_compte.html

pgiraud:
I forgot to mention that the PR has been updated as well.

PR #8524, opened 2018-08-18
Fix compatiblity with XHTML content type

NeoRaider:
HTML5 can be delivered with Content-type application/xhtml+xml. Only a few small fixes are necessary to make OpenLayers compatible with the additional constraints of XHTML: HTML tag names are case-sensitive, and most HTML entities like \&nbsp; are not available.


fredj:
thanks @NeoRaider

PR #8532, opened 2018-08-21
Add 'rendercomplete' event

### LINKED ISSUE #6447: Provide way to determine if all tile sources have finished loading
At the moment it is not possible to determine if all tiles of map have finished loading. In #3248 a workaround was proposed by counting `tileload` events (present in two examples: "Tile load events" and "Export PDF"). Unfortunately this technique will fail if number of tiles is larger than 16 - after loading first 16 tiles, there is a moment when number of loading tiles is 0 (as per `tileload` event counting or `map.tileQueue_.isEmpty()`). Thus a different approach is needed to determine if all tiles have been loaded. Similar functionality was asked in #2832 (to be on a par with OL 2).

I have come up with a solution that satisfies me (marisn/openlayers@8541ebcb879a12ab8e1af853ccf5c8004a171cbf) by monitoring `map.frameState_.wantedTiles` as it will contain also tiles that have not been requested yet (and thus no `tileloadstart` event has fired!). In #2832 an event was asked for, still I found it to be not so practical, as event should not fire if all tiles already have been loaded (i.e. thanks to caching). A variable to check after each rendering is better as it allows to know in `postrender` function if the current rendering is complete. The downside of monitoring loading events is already spotted with PDF generation example as #5780. 
Unfortunately my hack does not extend to non-tiled sources (i.e. `ol.source.Image`), although it could be useful to know if all layers have finished loading.

### LINKED ISSUE #7969: Printing the map from the canvas is incomplete even after tiles loaded and postrender fired
I am trying to print the map that I create on the fly. However the image that I am getting from the map canvas sometimes looks half baked:

[https://i.stack.imgur.com/RSnbT.png](https://i.stack.imgur.com/RSnbT.png)

In the screen shot, the top portion is the openlayers map that eventually got rendered and the bottom portion is what I captured from the canvas after postrender.

I am starting from the official Export PDF example, but I am working with a map that just got created. I keep a track of the tiles loading and the tiles loaded through the tileloadstart and tileloadend map events. I also tried incorporating the postrender event as per this answer on stackoverflow. When all the tiles have been loaded I wait for the postrender event to capture the map image from the canvas.

The problem:

I think the problem has to do with the postrender event firing off two early and I don't know how much time to wait after the event before capturing the image from the canvas. I started with a delay of 100 milliseconds and then moved to 1 second and it seemed like it was working fine. But if I add more layers I had to take the delay to 3 seconds for it to start behaving correctly. So I don't know what the delay should be for maps with a lot of layers.

I also get the problem that sometimes the tileloadstart and tileloadend event keeps on firing even after postrender because some tiles had yet to be loaded and even though the tile load count had equalized.

I have created a codepen to demonstrate the issue: Print map using tile loading and postrender events
You would need to enter a Bing Maps Key at the top of the javascript file. A Bing Map Key can be generated from: https://www.bingmapsportal.com/. With the BingMap layer the problem is reproducible more consistently. In the code pen I am also capturing the problem when tileloadstart is called after the postrender and writing an error to the top of the page. Also please note that the behavior is inconsistent and you might have to reload several times for it to reproduce.

ahocevar:
This pull request adds a `rendercomplete` event which is triggered when the final map image has been rendered for the visible viewport.

Fixes #7969
Fixes #6447

### DIFF
--- a/examples/export-map.js
+++ b/examples/export-map.js
@@ -30,7 +30,7 @@ const map = new Map({
 });
 
 document.getElementById('export-png').addEventListener('click', function() {
-  map.once('postcompose', function(event) {
+  map.once('rendercomplete', function(event) {
     const canvas = event.context.canvas;
     if (navigator.msSaveBlob) {
       navigator.msSaveBlob(canvas.msToBlob(), 'map.png');
--- a/examples/export-pdf.js
+++ b/examples/export-pdf.js
@@ -3,7 +3,6 @@ import View from '../src/ol/View.js';
 import {defaults as defaultControls} from '../src/ol/control.js';
 import WKT from '../src/ol/format/WKT.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
-import {unByKey} from '../src/ol/Observable.js';
 import {OSM, Vector as VectorSource} from '../src/ol/source.js';
 
 const raster = new TileLayer({
@@ -48,9 +47,6 @@ const dims = {
   a5: [210, 148]
 };
 
-let loading = 0;
-let loaded = 0;
-
 const exportButton = document.getElementById('export-pdf');
 
 exportButton.addEventListener('click', function() {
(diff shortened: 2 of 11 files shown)


ThomasG77:
I don't see `rendercomplete` appears within the generated documentation of `ol.Map` e.g http://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html whereas I see it at `ol.PluggableMap` http://openlayers.org/en/latest/apidoc/module-ol_PluggableMap-PluggableMap.html
Don't think it's normal as I see it's also available on `ol.Map` as http://openlayers.org/en/latest/examples/export-map.html
Can you confirm?

ahocevar:
I think events are not properly shown in the api docs right now. I‘m currently working on making TypeDoc work for us instead of JSDoc, and there all events will be properly documented when ready.

ahocevar:
The event is definitely available on `ol/Map`.

PR #8567, opened 2018-08-31
Use 'Element' type instead of 'Node'

### LINKED ISSUE #8560: Incorrect types in XML based parsers
The XMLish parsers use `Node` as the type for many function params.  Many of those functions only work with `Element` or `HTMLElement` type nodes.

Here is an example from the [KML format](https://github.com/openlayers/openlayers/blob/d8efa569d9d6722c4fa653b9e1891799a7c58245/src/ol/format/KML.js#L670-L695):
```js
  /**
   * @param {Node} node Node.
   * @return {string|undefined} Name.
   */
  readNameFromNode(node) {
    for (let n = node.firstElementChild; n; n = n.nextElementSibling) {
      if (includes(NAMESPACE_URIS, n.namespaceURI) &&
          n.localName == 'name') {
        return readString(n);
      }
    }
    for (let n = node.firstElementChild; n; n = n.nextElementSibling) {
      const localName = n.localName;
      if (includes(NAMESPACE_URIS, n.namespaceURI) &&
          (localName == 'Document' ||
           localName == 'Folder' ||
           localName == 'Placemark' ||
           localName == 'kml')) {
        const name = this.readNameFromNode(n);
        if (name) {
          return name;
        }
      }
    }
    return undefined;
  }
```

This function takes a `Node` but then accesses `node.firstElementChild`.  This property is only implemented by `Document`, `DocumentFragment`, or `Element` type nodes (not the base `Node`).

To fix this, instead of adding typecasts wherever `firstElementChild` is accessed, I think it would make more sense to update the signature of the functions and work up the call stack fixing types.

fredj:
There is still some `Node` typing used but `npx tsc | grep "^src.*type 'Node'"` and `npx tsc | grep "^src.*type 'Element'"` are now empty.

Fixes #8560

### DIFF
--- a/src/ol/format/GML2.js
+++ b/src/ol/format/GML2.js
@@ -95,7 +95,7 @@ class GML2 extends GMLBase {
   }
 
   /**
-   * @param {Node} node Node.
+   * @param {Element} node Node.
    * @param {Array<*>} objectStack Object stack.
    * @private
    * @return {module:ol/extent~Extent|undefined} Envelope.
@@ -110,7 +110,7 @@ class GML2 extends GMLBase {
   }
 
   /**
-   * @param {Node} node Node.
+   * @param {Element} node Node.
    * @param {Array<*>} objectStack Object stack.
    * @private
    */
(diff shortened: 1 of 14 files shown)


tschaub:
Thank you @fredj!

Issue #8566, opened 2018-08-31
Problem with ScaleLine and Shared View from iframe


tohu12:
Hi,
when creating two maps with [shared view](http://openlayers.org/en/latest/examples/side-by-side.html) and every map is in another iframe and both maps have control [ScaleLine](http://openlayers.org/en/latest/apidoc/module-ol_control_ScaleLine-ScaleLine.html), then OL returns exception error "Cannot read property 'getPointResolutionFunc' of null" at getPointResolution ([proj.js:193](https://github.com/openlayers/openlayers/blob/v5.2.0/src/ol/proj.js#L193))
(Yes, I need used iframe for synchronized maps ;))

Reason:
ScaleLine is called [getPointResolution()](https://github.com/openlayers/openlayers/blob/v5.2.0/src/ol/control/ScaleLine.js#L180) from proj.js with first parametr [projection object](http://openlayers.org/en/latest/apidoc/module-ol_proj_Projection-Projection.html) from shared view, and getPointResolution() calls [get(projection)](https://github.com/openlayers/openlayers/blob/v5.2.0/src/ol/proj.js#L191) and function get() uses condition [`if (projectionLike instanceof Projection) {` ](https://github.com/openlayers/openlayers/blob/v5.2.0/src/ol/proj.js#L160). But `instanceof` not recognizes projection from shared view. instanceof ol/proj/Projection is failed because projection is from [another iframe](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_context_(e.g._frames_or_windows))!

Solution:
1.) from ScaleLine call [getPointResolution()](https://github.com/openlayers/openlayers/blob/v5.2.0/src/ol/control/ScaleLine.js#L180) only with projection string code e.g. 'EPSG:4326'
`let pointResolution =
        getPointResolution(projection.getCode(), viewState.resolution, center, pointResolutionUnits);`

2.) in [proj.js function get()](https://github.com/openlayers/openlayers/blob/v5.2.0/src/ol/proj.js#L158) do not use condition only with instanceof, but any like this:
`if (projectionLike instanceof Projection || typeof projectionLike === 'object') {`
or
`if (projectionLike instanceof Projection || (projectionLike && projectionLike.getCode)) {`
or
do not return null, but original function input

Or any other solution how resolve this problem ?

T

tschaub:
[...]

PR #8568, opened 2018-08-31
ScaleLine call getPointResolution with SRS identifier instead of Proj…

tohu12:
ScaleLine call getPointResolution with SRS identifier instead of Projection object. Is better for shared view between more maps from iframe.

fix: #8566


ahocevar:
Replaced by #8569.

PR #8579, opened 2018-09-03
Better type annotation

marcjansen:
See #8345.

With this 4 (!) warnings (`error TS2367: This condition will always return 'true' since the types 'number' and 'string' have no overlap.`) will be gone.


marcjansen:
Thanks for the review.

PR #8580, opened 2018-09-03
Cast POINTER_ID before comparison

marcjansen:
-1 one tsc error; `error TS2367: This condition will always return 'true' since the types 'string' and 'number' have no overlap.`

See #8345.


marcjansen:
Thx for the hint. Updated as suggested.

marcjansen:
This is embarassing, sorry. Updated again.

PR #8595, opened 2018-09-06
Optional arguments and typedef properties

tschaub:
This marks a few optional arguments and typedef properties as optional (not the same as `type|undefined`).

Before:
```
$ npx tsc | wc -l
    1374
```

After:
```
$ npx tsc | wc -l
    1329
```

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -136,15 +136,15 @@ import Units from './proj/Units.js';
 
 /**
  * @typedef {Object} AnimationOptions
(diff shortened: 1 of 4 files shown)


tschaub:
Hmm, I see that we use another syntax in other places.  `@property {type} [name]` instead of `@property {type=} name`.  I kind of like the latter better but I think we use the former in more places (and it does work better with default values), so I'll update this.

Issue #8614, opened 2018-09-10
Rework transformWithOptions


tschaub:
The `ol/Feature` module exports a function called `transformWithOptions`.  This expects either a geometry or an extent and returns a geometry or an extent.  The extent handling appears only to be used by the GML format.  Instead, we should have two functions: `transformGeometryWithOptions` and `transformExtentWithOptions`.  Each will expect a single type and return the same.  The GML format can call the one with extent.  All the others call the one with geometry.  This will fix type issues where we are assigning the result of the current function to a geometry.

PR #8615, opened 2018-09-10
Use GeoJSON types from @types/geojson

### LINKED ISSUE #8598: Add types for GeoJSON
```
src/ol/format/GeoJSON.js(84,15): error TS2304: Cannot find name 'GeoJSONFeature'.
src/ol/format/GeoJSON.js(92,30): error TS2304: Cannot find name 'GeoJSONGeometry'.
src/ol/format/GeoJSON.js(92,46): error TS2304: Cannot find name 'GeoJSONGeometryCollection'.
src/ol/format/GeoJSON.js(117,38): error TS2304: Cannot find name 'GeoJSONObject'.
```

We got rid of these types in #8273. Similar to #8596, we need to add in `@typedef`s for those that we use. In addition, I think we'll want to quote property access on these objects - but this should probably be confirmed.

tschaub:
This adds `@types/geojson` as a development dependency and imports the type definitions in the GeoJSON format.  I've added it as a development dependency since our exported methods do not rely on the types.  If this turns out to be a problem for people who depend on the `ol` package, we can move it to be a regular dependency.

Before:
```
$ npx tsc | wc -l
    1280
```

After:
```
$ npx tsc | wc -l
    1238
```

The two remaining type errors in the GeoJSON format are due to an issue with the `transformWithOptions` function (see #8614).

Fixes #8598.

### DIFF
--- a/src/ol/format/GeoJSON.js
+++ b/src/ol/format/GeoJSON.js
@@ -17,6 +17,21 @@ import Point from '../geom/Point.js';
 import Polygon from '../geom/Polygon.js';
 import {assign, isEmpty} from '../obj.js';
 import {get as getProjection} from '../proj.js';
+import GeometryType from '../geom/GeometryType.js';
+
+/**
+ * @typedef {import("geojson").GeoJSON} GeoJSONObject
+ * @typedef {import("geojson").Feature} GeoJSONFeature
+ * @typedef {import("geojson").FeatureCollection} GeoJSONFeatureCollection
+ * @typedef {import("geojson").Geometry} GeoJSONGeometry
+ * @typedef {import("geojson").Point} GeoJSONPoint
+ * @typedef {import("geojson").LineString} GeoJSONLineString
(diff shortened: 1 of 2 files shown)


src/ol/format/GeoJSON.js
```diff
@@ -17,6 +17,21 @@ import Point from '../geom/Point.js';
 import Polygon from '../geom/Polygon.js';
 import {assign, isEmpty} from '../obj.js';
 import {get as getProjection} from '../proj.js';
+import GeometryType from '../geom/GeometryType.js';
+
+/**
+ * @typedef {import("geojson").GeoJSON} GeoJSONObject
```
tschaub: I was surprised to find that `import("@types/geojson")` does not work and that this is the way to import packages from the `@types` scope.  Unfortunate that this makes it so people cannot import from arbitrary packages.

ahocevar: That's interesting indeed.

PR #8618, opened 2018-09-11
Use TopoJSON types from @types/topojson

### LINKED ISSUE #8599: Add types for TopoJSON*
```
src/ol/format/TopoJSON.js(83,43): error TS2304: Cannot find name 'TopoJSONTopology'.
src/ol/format/TopoJSON.js(104,32): error TS2304: Cannot find name 'TopoJSONGeometryCollection'.
src/ol/format/TopoJSON.js(262,12): error TS2304: Cannot find name 'TopoJSONGeometry'.
src/ol/format/TopoJSON.js(285,12): error TS2304: Cannot find name 'TopoJSONGeometryCollection'.
```

We got rid of these types in #8273. Similar to #8596, we need to add in `@typedef`s for those that we use. In addition, I think we'll want to quote property access on these objects - but this should probably be confirmed.

fredj:
fixes #8599

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "devDependencies": {
     "@openlayers/eslint-plugin": "^4.0.0-beta.1",
     "@types/geojson": "^7946.0.4",
+    "@types/topojson-specification": "^1.0.0",
     "buble": "^0.19.3",
     "buble-loader": "^0.5.1",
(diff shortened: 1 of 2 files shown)


fredj:
I've added the brackets

ahocevar:
Thanks!

PR #8626, opened 2018-09-11
Better type annotations / type casts

marcjansen:
This removes three errors `error TS2345: Argument of type 'Foo' is not assignable to parameter of type 'Bar'.`

Please review.


marcjansen:
Thanks for the review.

Issue #8067, opened 2018-04-11
Opacity transition with preloaded tile causes some unexpected behavior.


yjwoo14:
To reproduce the behavior, go to https://openlayers.org/en/latest/examples/wmts-dimensions.html and do the following.

1. Set the sea-level slider to 100M
2. Zoom one level in.
3. Changing the slider causes flickering. (See attachment)

![flickering](https://user-images.githubusercontent.com/2172213/38613449-4390549e-3d8a-11e8-8188-92b6bf0a6a33.gif)

As far as I understood, this is because opacity transition can be made from preloaded tiles in a higher zoom level (#7267 may be relevant) and getting preloaded tiles does not check the dimension (which contains this slider value).

In my experiments, I could get around this issue by using only preloaded tiles whose key matches to the current key of tile. 

* Add condition `this.getKey() == tile.key` on line 143
https://github.com/openlayers/openlayers/blob/568b3d3559b4388c732dbe7b0e35d53f5b9d2f49/src/ol/source/Tile.js#L140-L143

Any opinion  would be appreciated.

- [x] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [x] I have searched GitHub to see if a similar bug or feature request has already been reported.
- [x] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [ ] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

ahocevar:
It would be interesting to see how the change you suggest affects fade-in user experience for maps with just one dimension. If that does not case a decrease in quality, we'd accept a pull request.

yjwoo14:
I went through sources to find out the ones setting 'key' and only `UrlTile` (including its child classes) calls `setKey()`. `TileImage` and `VectorTileSource` inherits directly from `UrlTile`, and  `TileImage` source sets `key`  when creating tiles. However, `VectorTileSource` does not set `key` at all, so my fix prevents from using preloaded vector tiles in the opacity transitions, so I would also suggest the following:

* Add `this.key = this.getKey();` on line 173
https://github.com/openlayers/openlayers/blob/004f91a020bd8f3d69d5c758e8a7537937059b56/src/ol/source/VectorTile.js#L164-L174

dimin:
[...]

PR #8642, opened 2018-09-14
fixes for optional key passing, issue #8067 for tile sources

dimin:
This fixes the same problem as described in #8067. While setting different tile urls setUrl() on a source the animation would always jump to the very first source url that was set. The solution was to set distinguishing optional keys for each source which would select proper tiles form cache for animation. Minor fixes were needed to make sure optional keys can be passes through tile sources.


src/ol/source/TileImage.js
```diff
@@ -83,7 +83,8 @@ class TileImage extends UrlTile {
       url: options.url,
       urls: options.urls,
       wrapX: options.wrapX,
-      transition: options.transition
+      transition: options.transition,
+      opt_key: options.opt_key
```
tschaub: I think this property name should just be `key` (we only use the `opt_` prefix for function params).  In addition, the property should be added to the [`@typedef` above](https://github.com/openlayers/openlayers/blob/7d77d05a575ab98391ae9de4038585bd2bfa1bee/src/ol/source/TileImage.js#L17-L55) with a bit of documentation describing what it does.

src/ol/source/UrlTile.js
```diff
@@ -64,6 +64,7 @@ class UrlTile extends TileSource {
      */
     this.tileUrlFunction = this.fixedTileUrlFunction ?
       this.fixedTileUrlFunction.bind(this) : nullTileUrlFunction;
+    this.key_ = options.opt_key || this.key_;
```
tschaub: Since this is a new property for the `UrlTile` constructor options, it needs to be added to the [`@typedef` here](https://github.com/openlayers/openlayers/blob/7d77d05a575ab98391ae9de4038585bd2bfa1bee/src/ol/source/UrlTile.js#L11-L27) as well.  As above, I would change the property name to just `key`.

tschaub:
Thanks for tackling this @dimin.  I hope the comments above make sense.  Let me know if not.

fredj:
@dimin do you plan to work on this pull request?

dimin:
Sorry guys, was super busy with some other work. I'll apply all the required fixes soon!

tschaub:
Thanks @dimin.  Any chance you can provide a like to an example that demonstrates the problem and the fix?  I’ll take a look at the example in #8067 when I get a chance.

dimin:
I'm using something similar to this example:
https://openlayers.org/en/latest/examples/reusable-source.html

I start with one url and change them for different z positions. Although simply setting the source url multiple times does work, smooth transitions would always start with the same tile, i.e. the tile produced from the very first URL I've given my source. The problem was with the key not being set and thus used when finding tiles in the cache.

In order to solve that, I needed to make sure the first URL had a key along with it and I needed to pass the new key along with the new url to the tiles being created and also ensure that the old key is used to fetch the tiles to initialize the transition. I would pass the initial key to the source being created first.

In my case I have my own source derived from TileImage and I've changed the setUrl like this:
  setUrl(url, key) {
    const urls = expandUrl(url);
    const tileUrlFunction = createFromTileUrlFunctions(urls.map(this.myCreateFromTemplate));
    this.setTileUrlFunction(tileUrlFunction, key || this.key_);
  }

I'm guessing this step is not quite needed as long as the first url has a key associated with it since UrlTile is setting the key from the url itself. Some of that wasn't working for me and I've had to set the keys explicitly.

tschaub:
Thanks @dimin.

I think we may need to revisit this – at least to add documentation and potentially to make it easier to get the behavior people want.

dimin:
Sure @tschaub, I'll help with what I can and really thank you for creating and maintaining such an amazing and extensible viewer!

Issue #8448, opened 2018-07-27
Type checking and API docs with TypeScript JSDoc annotations


ahocevar:
We are currently transitioning the code base from type annotations understood by JSDoc to TypeScript JSDoc annotations. This ticket is an overview on progress and remaining tasks:

- [x] #8345 Rewrite JSDoc annotations to those [understood by TypeScript](https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript)
- [x] #8449 New tool chain to generate API docs from the source code and JSDoc annotations
- [ ] Investigate TypeScript integration: will `--allowJS` be enough for users, or do we need to publish `.d.ts` files?

Since TypeScript knowledge is limited among the OpenLayers dev community, we appreciate any help on this. Also, please comment if we are missing something in the list above.

OSHistory:
@ahocevar 
Thanks! I'll take the liberty to ping some users that showed interest in the whole 
ts-issue, so we can pool some manpower.

OSHistory:
I pinged you because your activity on the openlayers project indicates you might 
be interested in the current plan to publish typescript definitions within the openlayers 
ol-package (as opposed to use the outdated DefinitlyTyped version, which used to 
worked well for the 4.x series BTW thanks @yairtawil  

If you are in fact not interested in the issue (anymore), I am sorry to have 'spammed' you. 

@8byr0
@achimha
@agxs 
@bjornharrtell
@BobbyT
@BrkCoder
@ca0v
@craigsnyders
@cvenzin
@DarthPinguin
@fgravin
@gberaudo
@h3imdall
@hei-pa
@icholy
@idevtech
@josepgomes
@jumpinjackie
@martin-reiterer
@mrfx
@niftydevelopment
@quentinol
@vrmerlin
@wulfsberg
@yairtawil

wulfsberg:
I have no experience with the JSDoc-annotation style of type definitions for TypeScript, so I'll be of limited use there
I am primarily coming at it as a consumer of OpenLayers in Angular applications, and have never been deep into the build stack of OpenLayers itself, so I'll probably be most useful as a guinea pig for the integration.

icholy:
Once the transition is complete, the typescript compiler can automatically generate a .d.ts file from the sources. Shipping with the definition file makes it easier to consume.

marcjansen:
I wanna also mention

@chrismayer
@kaivolland
@dnlkoch
@jansule
@weskamm
@atd-schubert
[...]

PR #8643, opened 2018-09-14
Fix style TypeScript errors

schmidtk:
I'm researching moving our own project to ES6/TypeScript, so I hopped on some low-hanging fruit for TypeScript errors (#8448). This fixes `tsc` errors for some files in `src/ol/style`.


tschaub:
Thanks @schmidtk!

PR #8657, opened 2018-09-18
Typescript fixes

fgravin:
Try to work along with typescript typechecking migration.
See https://github.com/openlayers/openlayers/issues/8448

### DIFF
--- a/src/ol/Geolocation.js
+++ b/src/ol/Geolocation.js
@@ -174,9 +174,12 @@ class Geolocation extends BaseObject {
    * @param {PositionError} error error object.
    */
   positionError_(error) {
-    error.type = EventType.ERROR;
+    const event = {
+      type: EventType.ERROR,
+      target: undefined
+    };
     this.setTracking(false);
-    this.dispatchEvent(/** @type {{type: string, target: undefined}} */ (error));
+    this.dispatchEvent(event);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


fgravin:
@fredj or @tschaub please review cause people open new PRs to fix again what i've fixed here, overlaping.
Thanks !

test/spec/ol/feature.test.js
```diff
@@ -102,7 +102,7 @@ describe('ol.Feature', function() {
 
     it('returns null for null geometry (constructor)', function() {
```
tschaub: It would be nice to update this description to `it('returns undefined ...')`, but probably not worth it given the flakiness of the tests in Travis right now.

src/ol/Geolocation.js
```diff
@@ -174,9 +174,12 @@ class Geolocation extends BaseObject {
    * @param {PositionError} error error object.
    */
   positionError_(error) {
-    error.type = EventType.ERROR;
```
fredj: We this change, we lost the detail of the error (error code, message) as defined in https://developer.mozilla.org/en-US/docs/Web/API/PositionError

fgravin: First the `PositionError` type does not contain a `type` property.
Then, you were returning the error object as the event, that was an error IMO.
Nothing is done with the error object in the end.

Where would you like to store `error.code` ?

PR #8663, opened 2018-09-18
Add vendor-specific TS declarations for the Fullscreen API.

### LINKED ISSUE #8660: Use of vendor-specific Fullscreen API causes TypeScript errors
TypeScript 3.1 [removed several vendor-specific types](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#some-vendor-specific-types-are-removed-from-libdts), including the vendor prefixed fullscreen API (`webkitIsFullScreen`, `webkitRequestFullscreen`, etc). This is causing a number of TypeScript errors because those are no longer defined.

The recommended fix is to define them locally, as needed to meet minimum support browsers.

schmidtk:
Fixes #8660.

I'm not fond of the `Element.ALLOW_KEYBOARD_INPUT` solution, but TypeScript does not support declaration merging for `declare var`. This prevents OpenLayers from adding vendor properties to their [Element declaration](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L4814), so a `@typedef` and cast was used as a workaround. If anyone knows a better way to do this, please let me know.

### DIFF
--- a/types/dom.d.ts
+++ b/types/dom.d.ts
@@ -1,0 +1,29 @@
+/**
+ * Type declarations extending TypeScript's lib/lib.dom.d.ts.
+ * https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts
+ */
+
+interface Document {
+  readonly mozFullScreen: boolean;
+  readonly webkitIsFullScreen: boolean;
+
+  readonly fullscreenElement: Element;
+  readonly mozFullScreenElement: Element;
+  readonly msFullscreenElement: Element;
+  readonly webkitFullscreenElement: Element;
+
+  readonly mozFullScreenEnabled: boolean;
+  readonly msFullscreenEnabled: boolean;
+  readonly webkitFullscreenEnabled: boolean;
+
+  mozCancelFullScreen(): void;
+  msExitFullscreen(): void;
+  webkitExitFullscreen(): void;
+}
+
+interface Element {
+  mozRequestFullScreen(): Promise<void>;
+  mozRequestFullScreenWithKeys(): Promise<void>;
+  msRequestFullscreen(): Promise<void>;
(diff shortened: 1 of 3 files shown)


fredj:
@schmidtk thanks for the pull request! 
Because `Element.ALLOW_KEYBOARD_INPUT` is only present in Chrome, I think it's not worth using it OpenLayers.

fredj:
@schmidtk I've pushed a commit that removes the `ALLOW_KEYBOARD_INPUT` parameter.

schmidtk:
Great, I was hoping it could be removed. Thanks @fredj.

fredj:
Thanks for your work on this @schmidtk

types/dom.d.ts
```diff
@@ -0,0 +1,29 @@
...
+  mozCancelFullScreen(): void;
+  msExitFullscreen(): void;
+  webkitExitFullscreen(): void;
+}
+
+interface Element {
+  mozRequestFullScreen(): Promise<void>;
+  mozRequestFullScreenWithKeys(): Promise<void>;
+  msRequestFullscreen(): Promise<void>;
+  webkitRequestFullscreen(allowKeyboardInput?: number): Promise<void>;
```
tschaub: I'm curious what makes these extend rather than override the existing types.  What would it look like if we wanted to override a type?

schmidtk: These extend interfaces defined in the default library files loaded by `tsc` via TypeScript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). To replace declarations, you would have to use the `--lib` option to `tsc` to override the default libs and omit those you want to replace. Docs on that [here](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

tschaub: Good to know.  Thanks @schmidtk.

PR #8665, opened 2018-09-18
Fix type check errors in RBush

wallw-bits:
This was previously taking advantage of Javascript's subtle type coercions for arithmetic operations. Looping over object keys via `for...in` or the result of `Object.keys()` always results in a string key, even if the key was set via `obj[3] = "thing"`. Therefore, do the conversion explicitly (since it is happening under the hood anyway) in order to satisfy `tsc`.

Update:
Since the bitwise operations are apparently superfluous, simply removing those will satisfy `tsc` without any further additions.

### DIFF
--- a/src/ol/structs/RBush.js
+++ b/src/ol/structs/RBush.js
@@ -233,7 +233,7 @@ class RBush {
   concat(rbush) {
     this.rbush_.load(rbush.rbush_.all());
     for (const i in rbush.items_) {
-      this.items_[i | 0] = rbush.items_[i | 0];
+      this.items_[i] = rbush.items_[i];
     }
   }
 


wallw-bits:
Is the test failure in that build something that is likely to be intermittent (and we should just restart the build) or is it coming from master? I'm not getting any failures from `npm run test` when running locally.

marcjansen:
I have restarted the build for now, as th errors seem unrelated.

```
...............
Edge 17.17134.0 (Windows 10 0.0.0) ol.View #animate() completes complex animation using resolution FAILED
	Error: expected false to equal true (/home/travis/build/openlayers/openlayers/node_modules/expect.js/index.js:102)
....
Edge 17.17134.0 (Windows 10 0.0.0) ol.View #getZoom returns correct zoom levels (with resolutions array) FAILED
	TypeError: Unable to get property 'setResolution' of undefined or null reference``

File: src/ol/structs/RBush.js [hunk 1]
```diff
@@ -233,7 +233,8 @@ class RBush {
   concat(rbush) {
     this.rbush_.load(rbush.rbush_.all());
     for (const i in rbush.items_) {
-      this.items_[i | 0] = rbush.items_[i | 0];
+      const x = Number(i);
+      this.items_[x | 0] = rbush.items_[x | 0];
```
tschaub: We shouldn't need the bitwise or here anymore (we aren't intentionally truncating to integer, and the value is already a number).

tschaub:
We should probably decide how we want to handle this more generally.  Our options are roughly:

1. Always use `string` as the type for object keys (e.g. `Object<string, Entry>`) and explicitly cast values in JS (e.g. `items[String(getUid(value))]`) before property access.
2. Always use `string` as the type for object keys (e.g. `Object<string, Entry>`) and add typecasts to tell the checker that we understand coercion (e.g. `items[/** @type {string} */ (getUid(value))]`) before property access.
3. Pretend object properties are not strings (e.g. `Object<number, Entry>`) and explicitly cast values in JS (e.g. `items[Number(str)]`).
3. Pretend object properties are not strings (e.g. `Object<number, Entry>`) and add typecasts (e.g. `items[/** @type {number} */ (str)]`).

The change here is alternative 3.  I think my preference would be to do 2 everywhere (no unnecessary JS at runtime).  Curious to hear what others think.

wallw-bits:
Options 1 and 2 actually execute more unnecessary JS than 3 or 4 in this case. `items[str | 0]` is really `items[String(Number(str) | 0)]` under the hood. For simpler cases, `items[num]` is really `items[String(num)]` under the hood. Those conversions should be implemented natively and should be less statements overall than calling `getUid()`.

A better solution might be to take advantage of `Map` instead, but that might depend on what kind of browser/node support you're targeting.

src/ol/structs/RBush.js
wallw-bits: If we remove those, `tsc` doesn't complain at all and no casts are necessary.

tschaub:
I wasn't suggesting `items[str | 0]` in 1 and 2.

wallw-bits:
I meant that the addition of `String(getUid(value))` is unnecessary because either `String(value)` or `value.toString()` is already occurring under the hood for non-string types.

tschaub:
Right.  This is what I was suggesting in alternative 2.  No explicit JS conversion (no explicit `String()` call).  Just add a comment (typecast) if needed for `tsc`.

tschaub:
I think this is the simplest change that removes the typechecking errors in `RBush.js`:
```diff
diff --git a/src/ol/structs/RBush.js b/src/ol/structs/RBush.js
index ab7746db1..3a5dfc9fc 100644
--- a/src/ol/structs/RBush.js
+++ b/src/ol/structs/RBush.js
@@ -233,7 +233,7 @@ class RBush {
   concat(rbush) {
     this.rbush_.load(rbush.rbush_.all());
     for (const i in rbush.items_) {
-      this.items_[i | 0] = rbush.items_[i | 0];
+      this.items_[i] = rbush.items_[i];
     }
   }
 
```

No extra JS, no extra JSDoc.

wallw-bits:
Yup. That's what I've got now as of the latest push.

tschaub:
And this looks like exactly what you've done in 7ff974be2c2a979ebd2c1e5a6bd5bfa948f787cc.  I was still looking at the original change with the explicit `Number(i)` call.

tschaub:
Thanks @wallw-bits.

PR #8670, opened 2018-09-19
Define UrlTile#fixedTileUrlFunction as an abstract function

fredj:
Fixes the TypeScript error:
```
Class 'TileImage' defines instance member property 'fixedTileUrlFunction', but extended class 'TileWMS' defines it as instance member function.
```


fredj:
My change is incorrect; `fixedTileUrlFunction` must be a property, not an abstract function.

fredj:
fixed with #8765

PR #8675, opened 2018-09-19
Get width as number from number|Size

wallw-bits:
This satisfies `tsc` for `ol/source/Zoomify.js` in addition to being a bit more defensive from custom tile grids, even if using them is very uncommon for this source.


tschaub:
Nice to see that this effort has resulted in at least one bug fix (although maybe very unlikely to hit).  Thanks @wallw-bits.

PR #8677, opened 2018-09-19
Type check fixes for VectorSource

wallw-bits:
- Added parameter types to event listeners.
- Fixed return type for feature collection branch.
- Removed duplicate template definition.

This fixes all the outstanding issues from `tsc` except those caused by `@abstract` method overrides.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -420,6 +420,9 @@ class VectorSource extends Source {
   bindFeaturesCollection_(collection) {
     let modifyingCollection = false;
     listen(this, VectorEventType.ADDFEATURE,
+      /**
+       * @param {VectorSourceEvent} evt The vector source event
+       */
       function(evt) {
         if (!modifyingCollection) {
           modifyingCollection = true;
@@ -428,6 +431,9 @@ class VectorSource extends Source {
         }
       });
     listen(this, VectorEventType.REMOVEFEATURE,
+      /**
+       * @param {VectorSourceEvent} evt The vector source event
+       */
       function(evt) {
         if (!modifyingCollection) {
           modifyingCollection = true;
@@ -436,6 +442,9 @@ class VectorSource extends Source {
         }
       });
     listen(collection, CollectionEventType.ADD,
+      /**
+       * @param {import("../Collection.js").CollectionEvent} evt The collection event
+       */
       function(evt) {
         if (!modifyingCollection) {
           modifyingCollection = true;
@@ -444,6 +453,9 @@ class VectorSource extends Source {
         }
       }, this);
     listen(collection, CollectionEventType.REMOVE,
+      /**
+       * @param {import("../Collection.js").CollectionEvent} evt The collection event
+       */
       function(evt) {
         if (!modifyingCollection) {
           modifyingCollection = true;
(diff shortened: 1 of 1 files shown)


src/ol/source/Vector.js
```diff
@@ -511,7 +523,7 @@ class VectorSource extends Source {
     if (this.featuresRtree_) {
       return this.featuresRtree_.forEach(callback);
     } else if (this.featuresCollection_) {
-      return this.featuresCollection_.forEach(callback);
+      this.featuresCollection_.forEach(callback);
```
tschaub: I'm surprised we don't need a specific `return` (or `return undefined`) here.  I would think the return type would have to include `void` without the return.

wallw-bits: `Collection.forEach` does not have a return (equivalent to `return undefined` in Javascript). However, I think `tsc` differentiates between no return (`void`) and `return undefined`. This function allows returning `T|undefined`, so simply letting it fall through to the end of the function avoids returning `void` (which doesn't match) from `Collection.forEach`.

tschaub: Well, sometimes.

```js
/**
 * A function that does not have a return statement.
 * @return {undefined} The return.
 */
function noReturnBad() {
  // nothing to return
}

/**
 * A function that does not have a return statement.
 * @return {void} The return.
 */
function noReturnGood() {
  // nothing to return
}

/**
 * A function that has a branch with no return.
 * @param {boolean} maybe Should return a value.
 * @return {boolean|undefined} The return.
 */
function mayReturnUnexpected(maybe) {
  if (maybe) {
    return maybe;
  }
}

/**
 * A function that has a branch with no return.
 * @param {boolean} maybe Should return a value.
 * @return {boolean|void} The return.
 */
function mayReturnExpected(maybe) {
  if (maybe) {
    return maybe;
  }
}
```

The `noReturnBad` function results in `A function whose declared type is neither 'void' nor 'any' must return a value.`.  This is what I expected.  No `return` means return type is `void` (or `any`).

The `noReturnGood` function is correctly typed.

There are no errors in the `mayReturnUnexpected` function.  This is not what I expected.  One branch of the function does not have a return statement.  The other returns a `boolean`.  For consistency with the rule above, I assumed the return type would need to be `boolean|void`.

The `mayReturnExpected` function is also correct according to `tsc`.  This is what I would expect to have to do for a function that might return a boolean or might not have an explicit return.

(And yes, I know that the value of the return is `undefined` without an explicit `return`.  It is the `TS2355` error above that led me to believe that `tsc` would require `void` in the type if there is a branch without a `return` statement.)

I'm not saying that anything needs to change here, just saying I was surprised by the apparent inconsistency in `void` handling.

wallw-bits: That is weird, and I'm a bit disappointed that we will now have to deal with `null`, `undefined`, _and_ `void` in types.

wallw-bits: Would you prefer that we future-proof this against newer versions of `typescript` and explicitly return `undefined`?

tschaub: No, that's ok.  Let's change it when/if it becomes a problem.

Type annotations aside, I think the explicit return is nice.  If someone comes along and adds code to this function without looking above (unlikely), they may introduce a bug.  If the `else` block had an explicit `return`, this would not happen.

I don't think we need to change anything here, just saying that I think an explicit return in a branch is nice.

PR #8678, opened 2018-09-19
Ensure image is Image or Video before settings src

wallw-bits:
Because canvas elements do not contain the `src` property.


wallw-bits:
I verified that the tests are running successfully on my local machine for this branch.

tschaub:
Thanks @wallw-bits.

PR #8680, opened 2018-09-19
TypeScript fixes for ol/events

schmidtk:
This fixes most remaining errors in `ol/events`. #8666 handles errors in `ol/events/condition`.

The `ol_lm` errors were fixed in a similar manner to `ol_uid`, by restricting references to that map to the `getListenerMap` call and typing the target as an Object in that call. This required adding a param to enable map creation when it didn't exist, but the function isn't exported so the change only impacts `events.js`.

### DIFF
--- a/src/ol/events.js
+++ b/src/ol/events.js
@@ -78,24 +78,33 @@ export function findListener(listeners, listener, opt_this, opt_setDeleteIndex) 
  * @return {Array<EventsKey>|undefined} Listeners.
  */
 export function getListeners(target, type) {
-  const listenerMap = target.ol_lm;
+  const listenerMap = getListenerMap(target);
   return listenerMap ? listenerMap[type] : undefined;
 }
 
 
 /**
- * Get the lookup of listeners.  If one does not exist on the target, it is
- * created.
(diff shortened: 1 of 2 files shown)


schmidtk:
The three remaining errors are caused by a TypeScript issue handling [call signatures of union types](https://github.com/Microsoft/TypeScript/issues/7294#issuecomment-190335544). Example:

```
src/ol/events.js:122:7 - error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '((type: string, callback: EventListenerOrEventListenerObject, options?: boolean | EventListenerOptions) => void) | ((type: string, listener: (arg0: Event | Event) => boolean | void) => void)' has no compatible call signatures.

122       target.removeEventListener(type, listeners[i].boundListener);
```

Both `EventTarget` (native) and `Target` (OL) are compatible with this call, but their signatures are not identical so the union of `EventTarget|Target` does not have a call signature. I can resolve the errors with an explicit cast of `target` to one of the two types, but that's a workaround not a solution. Any ideas on a fix for those three errors?

tschaub:
Thanks @schmidtk.

PR #8682, opened 2018-09-19
Add TypeScript type declaration for Esri JSON

### LINKED ISSUE #8597: Add types for EsriJSON*

```
src/ol/format/EsriJSON.js(304,12): error TS2304: Cannot find name 'EsriJSONGeometry'.
src/ol/format/EsriJSON.js(404,27): error TS2304: Cannot find name 'EsriJSONPoint'.
src/ol/format/EsriJSON.js(452,13): error TS2304: Cannot find name 'EsriJSONPolyline'.
src/ol/format/EsriJSON.js(506,13): error TS2304: Cannot find name 'EsriJSONMultipoint'.
src/ol/format/EsriJSON.js(523,13): error TS2304: Cannot find name 'EsriJSONPolygon'.
```

We got rid of these types in #8273.  Similar to #8596, we need to add in `@typedef`s for those that we use.  In addition, I think we'll want to quote property access on these objects - but this should probably be confirmed.

schmidtk:
This adds a TypeScript declaration file for Esri JSON, based on the old Closure externs file.

As a general rule of thumb, I used bracket notation (or `key in obj`) to determine the Esri JSON type from an `Object` and dot notation when the type was known.

Resolves #8597

### DIFF
--- a/src/ol/format/EsriJSON.js
+++ b/src/ol/format/EsriJSON.js
@@ -104,9 +104,8 @@ class EsriJSON extends JSONFeature {
    * @inheritDoc
    */
   readFeaturesFromObject(object, opt_options) {
-    const esriJSONObject = /** @type {EsriJSONObject} */ (object);
     const options = opt_options ? opt_options : {};
-    if (esriJSONObject.features) {
+    if ('features' in object) {
       const esriJSONFeatureCollection = /** @type {EsriJSONFeatureCollection} */ (object);
       /** @type {Array<import("../Feature.js").default>} */
       const features = [];
@@ -171,7 +170,7 @@ class EsriJSON extends JSONFeature {
       object['geometry'] = writeGeometry(geometry, opt_options);
       if (opt_options && opt_options.featureProjection) {
         object['geometry']['spatialReference'] = /** @type {EsriJSONCRS} */({
-          wkid: getProjection(opt_options.featureProjection).getCode().split(':').pop()
+          wkid: Number(getProjection(opt_options.featureProjection).getCode().split(':').pop())
         });
       }
     }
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks @schmidtk.

We were envisioning using only JS and JSDoc for this type checking effort.  Since we already have `@typedef` elsewhere, it is nice to think that people wouldn't have to learn (or wonder about) two different ways to define types.  I didn't get these comments in before #8660, so the change here is consistent with that.  But I'm curious what others think about trying to stick with JS and JSDoc vs. mixing JS, JSDoc, and TypeScript (cc @ahocevar and @fredj).

We definitely shouldn't make it harder on ourselves - so if using TypeScript is far easier then that is a benefit.  But it does add some overhead to have to think about types defined in `*.js` sources in addition to these `*.d.ts` files.

tschaub:
For comparison, here is a similar change that adds types with JSDoc: https://github.com/openlayers/openlayers/pull/8596/files

ahocevar:
Maybe it would make sense to publish the ESRI JSON types as separate package as part of Definitely Typed.

I agree with @tschaub that we should use JSDoc types only in this repo.

schmidtk:
That’s fine, I can implement with typedef’s instead. Do you foresee any difference between TS type declarations vs JSDoc typedef’s in compilation/minification? I know Closure is no longer used, but I wasn’t sure if webpack would avoid minification of libraries defined in TS declarations, similar to externs with Closure.

I’m out of the office for a bit, but I’ll reimplement this next week. Thanks for the feedback!

schmidtk:
Closing this PR, since the implementation with Definitely Typed will be very different. I'll start working that.

PR #8693, opened 2018-09-21
Fix type check errors for Image source

wallw-bits:
The parent class does not accept an `extent` property on its options. Also check that the image is an `img` or `video` tag before setting the source as `Canvas` does not have the `src` property.


tschaub:
Thanks @wallw-bits.

Issue #7485, opened 2017-11-18
I hope the Zoom Slider Control behavior of OpenLayers version 4 return to as in version 3.


astroace:
Since the time Openlayers change from version 3 to version 4, the zoom slider control became difficult to use.

The behavior of the zoom slider of version 4.0.0 or later is as follows.

(Start reproduction)
First, drag the toggle of the zoom slider, and while holding down the mouse button, and move the mouse pointer outside of the Zoom Slider Control range. (The toggle remains within the range of the Zoom Slider Control)
Next, release the mouse button in area of outside of the Zoom Slider Control range.
Next, with the mouse button released, move the mouse pointer into the range of the Zoom Slider Control.
And then, the zoom slider toggle suddenly starts to move. At the same time, the zoom level is changed without permission.
(Reproduction end)

I am suffering from this Zoom Slider Control behavior. So please.
I want the behavior of Zoom Slider Control of OpenLayers version 4 to return as in version 3.
To that end, I would like to ask to restore the "dragListenerKeys_" and the related codes (in zoomslider.js) which existed up to version 3.20.1.

that's all, thank you.

Thank you for your interest in making OpenLayers better!

To keep this project manageable for maintainers, we ask you to please check all boxes below before submitting an issue.

- [ o] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [ o] I have searched GitHub to see if a similar bug or feature request has already been reported.
- [ o] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [ o] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

ahocevar:
Please submit a pull request with the suggested fix.

marcjansen:
If you need any help providing the PR, just drop a line here.

astroace:
Thank you for your replies.

Specifically, it is as follows.

In the following, (+) means row addition.

(Excerpt by current codes in the "zoomslider.js(version 4.5.0)")
     /**
      * @type {boolean}
      * @private
      */
     this.dragging_;

     /**
      * @type {number}
      * @private
      */
     this.heightLimit_ = 0;

(My proposal referring codes "zoomslider.js (version 3.20.1)".)
     /**
      * @type {boolean}
      * @private
      */
[...]

PR #6422, opened 2017-01-31
Don't listen twice to move and end events in ol.control.ZoomSlider

fredj:
tested on chrome 55 desktop and chrome 55 mobile

fixes #5526

PR #8698, opened 2018-09-24
Fix zoom slider events

tr00st:
Fix as discussed in #7485 - reverting the changes from #6422 and replacing them with a fix that leaves the intended events back in. Makes sure that if the user moves the mouse outside of the ZoomSlider, the expected events still fire correctly.


tr00st:
Seen the issues raised from Travis CI - will get those sorted.

PR #8707, opened 2018-09-25
Fix TypeScript errors in ol/format/Feature

schmidtk:
Remaining errors are caused by Microsoft/TypeScript/issues/17227.


schmidtk:
Test failures appear unrelated and do not happen locally.

PR #8519, opened 2018-08-15
GML Format Improvements #8516 #8517 #8518

NielsCharlier:
This PR is for three issues unfortunately, because the patches are interdependent. There are however three separate commits.

faetureMember tag support #8516 
basic GML 3.2 format and namespace support #8517  
complex feature support #8518

all patches have test cases included.

PR #8711, opened 2018-09-25
Fix TypeScript errors in ol/format/GML

schmidtk:
Fixes all errors in `ol/format/GML*.js`.

### DIFF
--- a/src/ol/format/GML2.js
+++ b/src/ol/format/GML2.js
@@ -179,7 +179,7 @@ class GML2 extends GMLBase {
   writeFeatureElement(node, feature, objectStack) {
     const fid = feature.getId();
     if (fid) {
-      node.setAttribute('fid', fid);
+      node.setAttribute('fid', /** @type {string} */ (fid));
     }
     const context = /** @type {Object} */ (objectStack[objectStack.length - 1]);
     const featureNS = context['featureNS'];
@@ -285,7 +285,7 @@ class GML2 extends GMLBase {
   writeGeometryElement(node, geometry, objectStack) {
     const context = /** @type {import("./Feature.js").WriteOptions} */ (objectStack[objectStack.length - 1]);
     const item = assign({}, context);
-    item.node = node;
+    item['node'] = node;
     let value;
     if (Array.isArray(geometry)) {
       if (context.dataProjection) {
@@ -588,9 +588,9 @@ class GML2 extends GMLBase {
 /**
  * @const
  * @type {Object<string, Object<string, import("../xml.js").Parser>>}
- * @private
+ * @protected
  */
-GML2.prototype.GEOMETRY_FLAT_COORDINATES_PARSERS_ = {
+GML2.prototype.GEOMETRY_FLAT_COORDINATES_PARSERS = {
   'http://www.opengis.net/gml': {
     'coordinates': makeReplacer(GML2.prototype.readFlatCoordinates_)
   }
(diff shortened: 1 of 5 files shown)


schmidtk:
This PR will break #8519, due to the parser maps that were changed to `@protected`. The new `GML32` class will need to be updated.

src/ol/format/GML2.js
```diff
@@ -590,7 +590,7 @@ class GML2 extends GMLBase {
  * @type {Object<string, Object<string, import("../xml.js").Parser>>}
  * @private
  */
-GML2.prototype.GEOMETRY_FLAT_COORDINATES_PARSERS_ = {
+GML2.prototype.GEOMETRY_FLAT_COORDINATES_PARSERS = {
```
fgravin: Why does typescript complain ?
Shouldn't we remove things from prototype ?

schmidtk: The TypeScript error is caused by using `this. GEOMETRY_FLAT_COORDINATES_PARSERS_` in `GMLBase`, which is why this needs to be protected. Creating it here creates one instance on the prototype, versus putting it in the constructor and recreating this for each instance. This is the same pattern used for `GMLBase.prototype.RING_PARSERS`.

src/ol/format/GMLBase.js
```diff
@@ -469,6 +471,37 @@ class GMLBase extends XMLFeature {
...
+ */
+GMLBase.prototype.FLAT_LINEAR_RINGS_PARSERS = {
+  'http://www.opengis.net/gml': {}
+};
+
+
+/**
+ * @const
+ * @type {Object<string, Object<string, import("../xml.js").Parser>>}
+ * @private
```
fredj: `@protected` instead of `@private`

src/ol/format/GML2.js
```diff
@@ -590,7 +590,7 @@ class GML2 extends GMLBase {
  * @type {Object<string, Object<string, import("../xml.js").Parser>>}
  * @private
```
fredj: `@protected` instead of `@private`

src/ol/geom/MultiPolygon.js
```diff
@@ -28,7 +28,7 @@ import {quantizeMultiArray} from '../geom/flat/simplify.js';
 class MultiPolygon extends SimpleGeometry {
 
   /**
-   * @param {Array<Array<Array<import("../coordinate.js").Coordinate>>>|Array<number>} coordinates Coordinates.
+   * @param {Array<Array<Array<import("../coordinate.js").Coordinate>>>|Array<import("../geom.js").Polygon>|Array<number>} coordinates Coordinates.
```
fredj: `Polygon` can be used instead of `import("../geom.js").Polygon` (es6 import)

src/ol/geom/MultiLineString.js
```diff
@@ -23,7 +23,7 @@ import {douglasPeuckerArray} from '../geom/flat/simplify.js';
 class MultiLineString extends SimpleGeometry {
 
   /**
-   * @param {Array<Array<import("../coordinate.js").Coordinate>|import("../geom.js").MultiLineString>|Array<number>} coordinates
+   * @param {Array<Array<import("../coordinate.js").Coordinate>|import("../geom.js").LineString>|Array<number>} coordinates
```
fredj: `LineString` can be used instead of `import("../geom.js").LineString` (es6 import)

schmidtk:
@fredj I believe I have addressed all of your feedback. Thank you for the review!

schmidtk:
Rebased and removed the `instanceof` check (see #8771).

PR #8725, opened 2018-09-27
Export SelectEvent class

### LINKED ISSUE #8699: SelectEvent is unknown type in OL 5.2
How can I call dispatchEvent method and pass an instance of SelectEvent object when SelectEvent is unknown type?

fredj:
Fixes #8699

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -102,7 +102,7 @@ const SelectEventType = {
  * Events emitted by {@link module:ol/interaction/Select~Select} instances are instances of
  * this type.
  */
-class SelectEvent extends Event {
+export class SelectEvent extends Event {
   /**
    * @param {SelectEventType} type The event type.
    * @param {Array<import("../Feature.js").default>} selected Selected features.


tschaub:
Is it clear why an application would need to dispatch this event?

fredj:
ping @zgeorgiev

zgeorgiev:
@fredj Thanks a lot.
@tschaub I have a feature depending on another one and when I click on the first feature I would like to dispatch select event to the second feature. Both features are on different places on the map.

ahocevar:
I'm still not convinced we need this change. @zgeorgiev Can you share your code where you are using this exported event?

zgeorgiev:
@ahocevar unfortunately I cannot share the code but will try to explain my issue. I have a feature for car on the map and feature for the trail of this car. I would like to select the car feature even if I click on the trail feature. My code with older version of OL was in the select event handler of the trail feature I was triggering dispatchEvent select car feature. Now this is not possible, because SelectEvent object is not accessible. I will appreciate if you have a different solution for this case.

PR #8730, opened 2018-09-27
Fix TypeScript errors in ol/format/KML

schmidtk:
Fixes several errors in `ol/format/KML.js`. The remaining errors are caused by Microsoft/TypeScript/issues/27215.


schmidtk:
The build for this one completed successfully, but the issue is still reporting it as in progress. This has happened a few times now (I re-triggered the build with a rebase to master).

PR #8732, opened 2018-09-27
Fix type check in ol/source/Tile

wallw-bits:
Remove `extent` because the parent Source class does not contain `extent` on its `Options` type or use `options.extent` in the constructor.


fredj:
Can you also remove the extent in the typedef above? thanks

wallw-bits:
Done.

PR #8735, opened 2018-09-27
Fix TypeScript errors in ol/format/WKT

schmidtk:
Fixes all but one error in `ol/format/WKT.js`. The remaining error is caused by Microsoft/TypeScript/issues/27047.


schmidtk:
Test error is unrelated to the changes.

PR #8744, opened 2018-09-28
Fix type checks in VectorTileSource

wallw-bits:
Use typedef for TileClass instead of redefining the type. Also fixed the typedef since it no longer matched the constructor.

### DIFF
--- a/src/ol/source/VectorTile.js
+++ b/src/ol/source/VectorTile.js
@@ -22,7 +22,7 @@ import {createXYZ, extentFromProjection, createForProjection} from '../tilegrid.
  * stroke operations.
  * @property {import("../proj.js").ProjectionLike} projection Projection.
  * @property {import("./State.js").default} [state] Source state.
- * @property {import("../VectorTile.js").TileClass} [tileClass] Class used to instantiate image tiles.
+ * @property {typeof import("../VectorTile.js").default} [tileClass] Class used to instantiate image tiles.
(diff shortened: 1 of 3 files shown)


src/ol/source/VectorTile.js
```diff
@@ -123,10 +123,9 @@ class VectorTile extends UrlTile {
     this.overlaps_ = options.overlaps == undefined ? true : options.overlaps;
 
     /**
-       * @protected
-       * @type {function(new: Tile, import("../tilecoord.js").TileCoord, TileState, string,
-       *        import("../format/Feature.js").default, import("../Tile.js").LoadFunction)}
-       */
+     * @protected
+     * @type {import("../VectorTile.js").TileClass}
```
ahocevar: Couldn't this just be `@type {typeof import("../VectorTile.js").default}`? Then the `TileClass` typedef in `ol/VectorTile` could be removed.

wallw-bits: Updated per your suggestion.

ahocevar:
Hm, looks like our validate-jsdoc eslint rule does not like `{typeof import("...").default}`. Need to take a closer look at this.

ahocevar:
I have published `@openlayers/eslint-rules@4.0.0-beta.2`, which references `@openlayers/doctrine@2.2.0`, which contains tschaub/doctrine#1. With this update, `typeof` is accepted in `@param` annotations.

PR #8749, opened 2018-09-29
Fix typecheck for TileWMS source

wallw-bits:
Fixed `tsc` error about member function rather than member property.

### DIFF
--- a/src/ol/source/TileWMS.js
+++ b/src/ol/source/TileWMS.js
@@ -18,6 +18,56 @@ import {compareVersions} from '../string.js';
 import {appendParams} from '../uri.js';
 
 /**
+ * @param {import("../tilecoord.js").TileCoord} tileCoord The tile coordinate
+ * @param {number} pixelRatio The pixel ratio
+ * @param {import("../proj/Projection.js").default} projection The projection
+ * @return {string|undefined} The tile URL
+ * @this {TileWMS}
+ */
+const fixedTileUrlFunction = function(tileCoord, pixelRatio, projection) {
+
+  let tileGrid = this.getTileGrid();
+  if (!tileGrid) {
+    tileGrid = this.getTileGridForProjection(projection);
+  }
+
+  if (tileGrid.getResolutions().length <= tileCoord[0]) {
+    return undefined;
+  }
+
+  if (pixelRatio != 1 && (!this.hidpi_ || this.serverType_ === undefined)) {
+    pixelRatio = 1;
+  }
+
+  const tileResolution = tileGrid.getResolution(tileCoord[0]);
+  let tileExtent = tileGrid.getTileCoordExtent(tileCoord, this.tmpExtent_);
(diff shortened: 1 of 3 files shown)


src/ol/source/TileWMS.js
```diff
@@ -104,6 +154,8 @@ class TileWMS extends TileImage {
       transition: options.transition
     });
 
+    this.fixedTileUrlFunction = fixedTileUrlFunction;
```
ahocevar: Needs to be
```js
this.fixedTileUrlFunction = fixedTileUrlFunction.bind(this);
```

wallw-bits: Correct. And it also needs to be run before the `super()` call. What's the reasoning behind the lint rule that prevents `this` before `super`?

ahocevar: > What's the reasoning behind the lint rule that prevents this before super?

It's a language requirement, not just a linter rule.

wallw-bits: In that case, we can't bind any function to `this` before `super`. Therefore, should we check for the existence of `options.tileUrlFunction` later and set it explicitly via `this.setTileUrlFunction()`? The current implementation wants to call `this.fixedTileUrlFunction` when `this.setUrl()` or `this.setUrls()` is called from the super class constructor.

I can also leave it the way it was an add `// @ts-ignore` just before the function.

ahocevar: > check for the existence of `options.tileUrlFunction` later and set it explicitly via `this.setTileUrlFunction()`?

That sounds reasonable, yes. Thanks!

wallw-bits: If we go that route, should we remove the `fixedTileUrlFunction` logic altogether since it isn't usable?

ahocevar: Yes, please remove that logic.

wallw-bits:
I don't think the errors in that build are related. Could we please restart it to see if those resolve?

src/ol/source/UrlTile.js
```diff
@@ -188,9 +178,7 @@ class UrlTile extends TileSource {
   setUrls(urls) {
     this.urls = urls;
     const key = urls.join('\n');
-    this.setTileUrlFunction(this.fixedTileUrlFunction ?
-      this.fixedTileUrlFunction.bind(this) :
-      createFromTemplates(urls, this.tileGrid), key);
+    this.setTileUrlFunction(createFromTemplates(urls, this.tileGrid), key);
   }
```
wallw-bits: This is what I meant by "remove that logic". However, I'm not sure if we were on the same page since this is technically a breaking change.

ahocevar: This is not exactly what I had in mind. I've created #8765, which replaces this pull request and #8750. Please let me know what you think.

ahocevar:
Fixed with #8765.

PR #8750, opened 2018-09-29
Fix type check for TileArcGISRest source

wallw-bits:
Fix `tsc` error about member function rather than member property.

### DIFF
--- a/src/ol/source/TileArcGISRest.js
+++ b/src/ol/source/TileArcGISRest.js
@@ -10,6 +10,46 @@ import TileImage from '../source/TileImage.js';
 import {hash as tileCoordHash} from '../tilecoord.js';
 import {appendParams} from '../uri.js';
 
+
+/**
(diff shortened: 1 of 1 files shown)


src/ol/source/TileArcGISRest.js
```diff
@@ -82,6 +122,8 @@ class TileArcGISRest extends TileImage {
       transition: options.transition
     });
 
+    this.fixedTileUrlFunction = fixedTileUrlFunction;
```
ahocevar: Needs to be
```js
this.fixedTileUrlFunction = fixedTileUrlFunction.bind(this);
```

wallw-bits:
Removing the bit of logic specified in https://github.com/openlayers/openlayers/pull/8749#discussion_r221683421 causes this fix to be highly related. Therefore, I'm going to close this and move the change into the other PR.

PR #8753, opened 2018-09-29
Fix type check in ol/PluggableMap.js

wallw-bits:
- Added JSDoc to `getRenderer()`
- Cast default options to types
- Cast event to `MouseEvent` to satisfy `tsc`. Non mouse/touch events will simply produce `[NaN, NaN]`
- Check if layer is of type `Layer` before calling `getSource()`, as it does not exist on `Base`

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -464,6 +464,10 @@ class PluggableMap extends BaseObject {
 
   }
 
+  /**
+   * @abstract
+   * @return {import("./renderer/Map.js").default} The map renderer
+   */
   createRenderer() {
     throw new Error('Use a map type that has a createRenderer method');
   }
@@ -565,7 +569,8 @@ class PluggableMap extends BaseObject {
       return;
     }
     const coordinate = this.getCoordinateFromPixel(pixel);
-    opt_options = opt_options !== undefined ? opt_options : {};
+    opt_options = opt_options !== undefined ? opt_options :
+      /** @type {AtPixelOptions} */ ({});
     const hitTolerance = opt_options.hitTolerance !== undefined ?
       opt_options.hitTolerance * this.frameState_.pixelRatio : 0;
     const layerFilter = opt_options.layerFilter !== undefined ?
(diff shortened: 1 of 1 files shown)


fredj:
Can you also replace all `import("./layer/Layer.js").default` with `Layer`? Thanks

src/ol/PluggableMap.js
```diff
@@ -663,7 +670,9 @@ class PluggableMap extends BaseObject {
    */
   getEventPixel(event) {
     const viewportPosition = this.viewport_.getBoundingClientRect();
-    const eventPosition = event.changedTouches ? event.changedTouches[0] : event;
+    const eventPosition = event instanceof TouchEvent ? event.changedTouches[0] :
```
ahocevar: `TouchEvent` is not known on all browsers. I'd prefer a cast of `event` to `TouchEvent` here:
```js
 const eventPosition = 'changedTouches' in event ? /** @type {TouchEvent} */ (event).changedTouches[0] :`
```

src/ol/PluggableMap.js
```diff
@@ -464,6 +465,10 @@ class PluggableMap extends BaseObject {
 
   }
 
+  /**
+   * @abstract
```
ahocevar: I think TypeScript does not support the `@abstract` annotation, so it can be removed.

src/ol/PluggableMap.js
```diff
@@ -1424,9 +1433,12 @@ function getLoading(layers) {
     if (layer instanceof LayerGroup) {
       return getLoading(layer.getLayers().getArray());
     }
-    const source = layers[i].getSource();
-    if (source && source.loading) {
-      return true;
+
+    if (layer instanceof Layer) {
```
ahocevar: Can simply be an `else` block, if necessary with a type cast to `Layer`. We should try to avoid `instanceof` checks wherever possible.

wallw-bits:
Can we please restart the build for this? The test error appears to be unrelated, and the tests are passing locally for me.

PR #8757, opened 2018-10-01
Use typeof to simplify JSDoc class types

schmidtk:
Replaces all but one use of `function(new: ...` with `typeof` to simplify JSDoc class types. The remaining case in `WKT.js` throws `tsc` errors like `Type 'typeof Point' is not assignable to type 'typeof Geometry'.`. With `tsc` this is usually caused by the constructor signatures not matching.

### DIFF
--- a/src/ol/ImageTile.js
+++ b/src/ol/ImageTile.js
@@ -8,8 +8,7 @@ import {listenOnce, unlistenByKey} from './events.js';
 import EventType from './events/EventType.js';
 
 /**
- * @typedef {function(new: ImageTile, import("./tilecoord.js").TileCoord,
- * TileState, string, ?string, import("./Tile.js").LoadFunction)} TileClass
+ * @typedef {typeof ImageTile} TileClass
  * @api
  */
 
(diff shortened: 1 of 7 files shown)


src/ol/ImageTile.js
```diff
@@ -8,8 +8,7 @@ import {listenOnce, unlistenByKey} from './events.js';
 import EventType from './events/EventType.js';
 
 /**
- * @typedef {function(new: ImageTile, import("./tilecoord.js").TileCoord,
- * TileState, string, ?string, import("./Tile.js").LoadFunction)} TileClass
+ * @typedef {typeof ImageTile} TileClass
```
ahocevar: Would be even nicer to use `{typeof ImageTile}` directly where `{TileClass}` is currently used, and get rid of the `TileClass` typedef.

schmidtk: I agree, updated to replace `TileClass`.

schmidtk:
The build passed, but status didn't update here.

PR #8762, opened 2018-10-01
Fix type errors from interaction event handlers

schmidtk:
There are ~20 TypeScript errors caused by interaction event handler setup. Example:

```
src/ol/interaction/DragBox.js:143:7 - error TS2322: Type 'typeof handleDownEvent' is not assignable to type '(arg0: MapBrowserPointerEvent) => boolean'.
  Type 'handleDownEvent' is not assignable to type 'boolean'.

143       handleDownEvent: handleDownEvent,
          ~~~~~~~~~~~~~~~

  src/ol/interaction/Pointer.js:43:4
    43  * @property {function(MapBrowserPointerEvent):boolean} [handleDownEvent]
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    44  * Function handling "down" events. If the function returns `true` then a drag
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    45  * sequence is started.
       ~~~~~~~~~~~~~~~~~~~~~~~
    46  * @property {function(MapBrowserPointerEvent)} [handleDragEvent]
       ~~~
    The expected type comes from property 'handleDownEvent' which is declared here on type 'Options'
```

This seems to be caused by the `handleDownEvent` function being defined in the module scope, and not being bound to an interaction instance when referenced. Binding the function would cause a `'this' is not allowed before 'super()'` error.

I would like to resolve this by defining those handlers as methods on the interaction classes. This will allow extending classes to override the functions normally, instead of passing the new function through options.

I started going down this path and it successfully fixes the typing problems with the bonus of being able to replace those functions easily in extending classes. It does require moving a lot of code within each interaction, so I would like to get feedback on this approach before moving forward. If this approach seems sound I'll continue with the refactor, but please let me know if anyone has a better/preferred solution.

### DIFF
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -115,11 +115,7 @@ class DragBox extends PointerInteraction {
    */
   constructor(opt_options) {
 
-    super({
-      handleDownEvent: handleDownEvent,
-      handleDragEvent: handleDragEvent,
-      handleUpEvent: handleUpEvent
-    });
+    super();
 
     const options = opt_options ? opt_options : {};
 
@@ -159,7 +155,22 @@ class DragBox extends PointerInteraction {
     * @type {EndCondition}
     */
     this.boxEndCondition_ = options.boxEndCondition ?
-      options.boxEndCondition : defaultBoxEndCondition;
+      options.boxEndCondition : this.defaultBoxEndCondition;
+  }
+
+  /**
+   * The default condition for determining whether the boxend event
+   * should fire.
+   * @param {import("../MapBrowserEvent.js").default} mapBrowserEvent The originating MapBrowserEvent
+   *     leading to the box end.
+   * @param {import("../pixel.js").Pixel} startPixel The starting pixel of the box.
+   * @param {import("../pixel.js").Pixel} endPixel The end pixel of the box.
+   * @return {boolean} Whether or not the boxend condition should be fired.
+   */
+  defaultBoxEndCondition(mapBrowserEvent, startPixel, endPixel) {
+    const width = endPixel[0] - startPixel[0];
+    const height = endPixel[1] - startPixel[1];
+    return width * width + height * height >= this.minArea_;
   }
 
   /**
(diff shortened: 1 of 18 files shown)


schmidtk:
@tschaub / @ahocevar: Let me know what your thoughts are on this approach. It would fewer code changes to add setters for each handler (like `this.setHandleDownEvent(handleDownEvent.bind(this))`), but defining those functions as methods on the class seems like a better approach from an extensibility point of view.

ahocevar:
I agree with this approach @schmidtk, thanks for the effort! I added one commit (b7a72a0) that gets rid of the private handler function members in `ol/interaction/Pointer.js`, which improves readability in my opinion.

Please continue with the refactor.

schmidtk:
Finished the refactor and added tests for overriding event handlers via options or class extension. This fixes nearly all TS errors in `ol/interaction`.

The remaining errors are in `ol/interaction/Snap.js` and are caused by how `tsc` determines compatible call signatures with union types.

schmidtk:
@ahocevar While working this I noticed there are a lot of imports with a longer path than necessary. [Here is one example](https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/DoubleClickZoom.js#L5), where the `../interaction/` portion of the path could just be `./`. Is this done intentionally, or a side effect from automating the transition to `import`?

ahocevar:
Thanks for your work on this @schmidtk. The import paths are a result of the automated transition from OpenLayers 4.6 to JSDoc path types to TypeScript import types. At some point (low priority) these should be shortened.

schmidtk:
@ahocevar & @tschaub: With respect to the automated transition from OpenLayers 4.6, is there a repo with scripts used to migrate away from Closure? I'm starting to write jscodeshift scripts to transition [OpenSphere](https://github.com/ngageoint/opensphere) away from Closure so those would be very helpful.

ahocevar:
@schmidtk This is @tschaub's script that was used: https://gist.github.com/tschaub/1ea498c9d1e5268cf36d212b3949be4e.

If you need a starting point for other JSDoc comment transformations, you may also find https://github.com/openlayers/babel-plugin-jsdoc-closure useful as starting point - it does different transformation, but provides a good infrastructure for modifying JSDoc comments in an AST.

schmidtk:
@ahocevar Thanks I'll keep track of those for when we get closer to switching to ES6. We're currently working on replacing use of the Closure Library. Was any of that automated for OpenLayers? I know a lot of folks worked that over time, just curious if any of it was scripted/reusable. Appreciate your help!

PR #8764, opened 2018-10-01
Fix TypeScript errors in ol/interaction/Draw

schmidtk:
This fixes most TypeScript errors in `ol/interaction/Draw.js`, and the remaining errors will be fixed in #8762.

This PR also uses `window.setTimeout` instead of `setTimeout` when the return value is used, because TypeScript with the `es2017` library uses [Node's `setTimeout` signature](https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_args), which returns a Timer object. See Microsoft/TypeScript/issues/842 for more details.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -82,11 +82,35 @@ import {createEditingStyle} from '../style/Style.js';
 
 
 /**
+ * Coordinate type when drawing points.
+ * @typedef {import("../coordinate.js").Coordinate} PointCoordType
+ */
+
+
+/**
+ * Coordinate type when drawing lines.
+ * @typedef {Array<import("../coordinate.js").Coordinate>} LineCoordType
+ */
+
+
+/**
+ * Coordinate type when drawing polygons.
+ * @typedef {Array<Array<import("../coordinate.js").Coordinate>>} PolyCoordType
+ */
+
+
+/**
+ * Types used for drawing coordinates.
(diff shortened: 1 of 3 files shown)


src/ol/interaction/Draw.js
```diff
@@ -888,9 +912,9 @@ function handleDownEvent(event) {
     return true;
   } else if (this.condition_(event)) {
     this.lastDragTime_ = Date.now();
-    this.downTimeout_ = setTimeout(function() {
+    this.downTimeout_ = window.setTimeout(function() {
       this.handlePointerMove_(new MapBrowserPointerEvent(
-        MapBrowserEventType.POINTERMOVE, event.map, event.pointerEvent, event.frameState));
+        MapBrowserEventType.POINTERMOVE, event.map, event.pointerEvent, false, event.frameState));
```
ahocevar: Wow, good catch!

schmidtk:
Thanks @ahocevar. The `setTimeout` solution was based off of [this commit](https://github.com/openlayers/openlayers/commit/f3fb88905ec5249db89df855da19fe57173bc717). Should that one be changed as well, or is it acceptable in that file because it's for browser events and shouldn't be used with Node?

ahocevar:
Thanks for pointing that out @schmidtk. [That commit](
https://github.com/openlayers/openlayers/commit/f3fb88905ec5249db89df855da19fe57173bc717) should also be changed back to `setTimeout`.

PR #8765, opened 2018-10-02
Simplify logic for fixed tile url functions

ahocevar:
Replaces #8749 and #8750.


fredj:
`fixedTileUrlFunction` is still used in ol/source/WMTS, would it be possible to remove it as well?

ahocevar:
@wallw-bits Yes, but `fixedTileUrlFunction` was never part of the API. Your PR had a more severe breaking change, i.e. when calling `setUrl` or `setUrls` on a WMS or ArcGisREST source, it would reconfigure the source with a template url function instead of the fixed one.

ahocevar:
@fredj `fixedTileUrlFunction` logic from `ol/source/WMTS` removed.

PR #8776, opened 2018-10-03
Use setTimeout without the window namespace

ahocevar:
See https://github.com/openlayers/openlayers/pull/8764#issuecomment-426596239.


fredj:
ping @ahocevar

ahocevar:
Thanks for the review and the reminder @fredj!

PR #8781, opened 2018-10-03
Fix type check errors

schmidtk:
Fixes a few more type check errors.

After this, #8711, and I believe one more incoming PR from @wallw-bits, the remaining type check errors are caused by issues with `tsc`. The most significant two are:

- Microsoft/TypeScript/issues/27215: 133 errors (`TS2322` and `TS2345`)
- Microsoft/TypeScript/issues/17227: 56+ errors (`TS2355`, possibly `TS2554` as well)


marcjansen:
Very nice! Thanks for the effort so far.

PR #8787, opened 2018-10-05
Replace instanceof checks with other logic

### LINKED ISSUE #8771: Should we get rid of instanceof checks?
Reviewing pull requests that fix TypeScript type errors, I saw that `instanceof` checks were introduced in many places. Although it's a clean and readable way to check whether a value we get is what we expect, I'm a bit concerned about this for two reasons:

 1. **Performance**: some of our `instanceof` checks (e.g. event conditions for move events in `ol/events/condition.js`) are executed very often, and according to https://jsperf.com/duck-vs-instanceof they are 3 times slower than duck type checks.

 1. **Environments where the  [ol](https://npmjs.com/package/ol) package is used in multiple versions**: In these cases, `instanceof` checks will not return true when the constructors are from different versions of `ol`. We have seen this problem often in the past e.g. with `ol/Style` instances created by third party libraries (e.g. the [ol-mapbox-style](https://npmjs.com/package/ol-mapbox-style) package). 

What do others think? Should we try to get rid of `instanceof` checks?

ahocevar:
Fixes #8771.

### DIFF
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -5,8 +5,6 @@ import {assert} from './asserts.js';
 import {listen, unlisten, unlistenByKey} from './events.js';
 import EventType from './events/EventType.js';
 import BaseObject, {getChangeEventType} from './Object.js';
-import Geometry from './geom/Geometry.js';
-import Style from './style/Style.js';
 
 /**
  * @typedef {typeof Feature|typeof import("./render/Feature.js").default} FeatureClass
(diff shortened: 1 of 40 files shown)


src/ol/Feature.js
```diff
@@ -106,10 +104,9 @@ class Feature extends BaseObject {
       this, getChangeEventType(this.geometryName_),
       this.handleGeometryChanged_, this);
 
-    if (opt_geometryOrProperties !== undefined) {
-      if (opt_geometryOrProperties instanceof Geometry) {
-
-        const geometry = /** @type {Geometry} */ (opt_geometryOrProperties);
+    if (opt_geometryOrProperties) {
+      if (typeof /** @type {?} */ (opt_geometryOrProperties).getSimplifiedGeometry === 'function') {
```
fredj: can you use the `isGeometry` function here?

ahocevar:
Thanks for the review and the additional commit @fredj. Actually I got rid of using an `isGeometry` function, I just forgot to remove its definition. The reason why I decided against it was that the name might be misleading - if there's anything else that has a `getSimplifiedGeometry` function, it would also be treated as geometry. So I found it better to use a check that's as close as possible to the context it is used in.

Would it be ok for you if I remove the `isGeometry` and `isCollection` functions?

fredj:
> Would it be ok for you if I remove the isGeometry and isCollection functions?

yes, remove  them

PR #8798, opened 2018-10-10
Use unpkg.com instead of rawgit.com

fredj:
rawgit.com will be shutdown soon, see https://rawgit.com/, this pull request changes the urls in the examples from rawgit.com to unpkg.com

The other commits fix some bad indentations and move the adaptAttributions function out of the Source class


fredj:
Thanks for the review

PR #8804, opened 2018-10-11
Add possibility to disable collapsible attributions from Source

### LINKED ISSUE #8621: Contradiction between collapsible docs and OSM source
the OSM() source has a default behavior of hidding the attribution behind a button in the bottom right corner even on large screens (see it here https://openlayers.org/en/latest/examples/simple.html )

On https://openlayers.org/en/latest/apidoc/module-ol_control_Attribution-Attribution.html, it reads

> If you use an OSM source, [collapsible] should be set to false — see href="http://www.openstreetmap.org/copyright">www.openstreetmap.org/copyright OSM Copyright —

Shouldn't the OSM source create an Attribution object with collapsible set to false on its own, instead of providing only a copyright string ?

bobobobo:
This will fix #8621

### DIFF
--- a/src/ol/source/OSM.js
+++ b/src/ol/source/OSM.js
@@ -76,7 +76,8 @@ class OSM extends XYZ {
       reprojectionErrorThreshold: options.reprojectionErrorThreshold,
       tileLoadFunction: options.tileLoadFunction,
       url: url,
-      wrapX: options.wrapX
+      wrapX: options.wrapX,
+      attributionsCollapsible: false
     });
 
   }
--- a/src/ol/source/Source.js
+++ b/src/ol/source/Source.js
@@ -31,6 +31,7 @@ import SourceState from './State.js';
 /**
  * @typedef {Object} Options
  * @property {AttributionLike} [attributions]
+ * @property {boolean} [attributionsCollapsible] Whether attributions should be collapsible.
  * @property {import("../proj.js").ProjectionLike} projection
  * @property {SourceState} [state]
  * @property {boolean} [wrapX]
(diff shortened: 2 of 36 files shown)


src/ol/source/OSM.js
```diff
@@ -76,7 +76,8 @@ class OSM extends XYZ {
       reprojectionErrorThreshold: options.reprojectionErrorThreshold,
       tileLoadFunction: options.tileLoadFunction,
       url: url,
-      wrapX: options.wrapX
+      wrapX: options.wrapX,
+      attributionsCollapsible: true
```
ahocevar: Did you mean to initialize this with `false`?

src/ol/source/Source.js
```diff
@@ -66,6 +67,12 @@ class Source extends BaseObject {
      */
     this.attributions_ = this.adaptAttributions_(options.attributions);
 
+    /**
+     * @private
+     * @type {boolean}
+     */
+    this.attributionsCollapsible_ = options.attributionsCollapsible;
```
ahocevar: You could default this to `true` here, then other subclasses do not need to be changed.

src/ol/source/Source.js
```diff
@@ -120,6 +127,10 @@ class Source extends BaseObject {
     return this.attributions_;
   }
 
+  getAttributionsCollapsible() {
```
ahocevar: JSDoc comment with `@return` missing.

src/ol/source/OSM.js
bobobobo: Yes, this was a misstake, it is not initialized to false

src/ol/source/Source.js
bobobobo: JSDoc comment added

bobobobo:
I will try to add a unit test for the `Attribution` control

PR #8827, opened 2018-10-16
Removed unused forEachFeatureAtCoordinate from ol/source/Source

### LINKED ISSUE #8826: TypeScript errors from assigning a function with params to VOID
The following TypeScript errors are caused by `ol/source/Source~forEachFeatureAtCoordinate` being assigned to `ol/functions~VOID`. I would expect `tsc` to handle this better, but it seems like something that could also be fixed in OpenLayers.

```
src/ol/renderer/canvas/IntermediateCanvas.js:101:12 - error TS2554: Expected 0 arguments, but got 6.

101     return source.forEachFeatureAtCoordinate(
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
102       coordinate, resolution, rotation, hitTolerance, skippedFeatureUids,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
108         return callback.call(thisArg, feature, layer);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
109       });
    ~~~~~~~~

src/ol/renderer/webgl/ImageLayer.js:87:12 - error TS2554: Expected 0 arguments, but got 6.

 87     return source.forEachFeatureAtCoordinate(
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 88       coordinate, resolution, rotation, hitTolerance, skippedFeatureUids,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 95         return callback.call(thisArg, feature, layer);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 96       });
    ~~~~~~~~
```

It seems the default value of `ol/source/Source~forEachFeatureAtCoordinate` could be changed to an empty function (instead of `VOID`), but that would break the following related conditions:
https://github.com/openlayers/openlayers/blob/43ca7642938c9e8756e789b68d3aa2256ff51cd4/src/ol/renderer/canvas/IntermediateCanvas.js#L120-L123

https://github.com/openlayers/openlayers/blob/43ca7642938c9e8756e789b68d3aa2256ff51cd4/src/ol/renderer/webgl/ImageLayer.js#L233-L243

Is there a recommended approach to refactoring those?

schmidtk:
It seems `ol/source/Source~forEachFeatureAtCoordinate` is a legacy function and no longer used/implemented by any classes in OpenLayers. This removes that function, and the related renderer code.

Fixes #8826.

### DIFF
--- a/src/ol/renderer/canvas/IntermediateCanvas.js
+++ b/src/ol/renderer/canvas/IntermediateCanvas.js
@@ -4,7 +4,6 @@
 import {scale as scaleCoordinate} from '../../coordinate.js';
 import {createCanvasContext2D} from '../../dom.js';
 import {containsExtent, intersects} from '../../extent.js';
-import {VOID} from '../../functions.js';
 import CanvasLayerRenderer from '../canvas/Layer.js';
 import {create as createTransform, apply as applyTransform} from '../../transform.js';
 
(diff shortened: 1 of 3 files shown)


fredj:
Rebased into master (I should have read #8825 more carefully)

src/ol/renderer/canvas/IntermediateCanvas.js
```diff
@@ -93,20 +92,7 @@ class IntermediateCanvasRenderer extends CanvasLayerRenderer {
...
-    return source.forEachFeatureAtCoordinate(
-      coordinate, resolution, rotation, hitTolerance, skippedFeatureUids,
-      /**
-       * @param {import("../../Feature.js").FeatureLike} feature Feature.
-       * @return {?} Callback result.
-       */
-      function(feature) {
-        return callback.call(thisArg, feature, layer);
-      });
+    return undefined;
```
ahocevar: Do `@abstract` methods need to be implemented by every subclass? If not, this function could be removed.

schmidtk: You're right, this can be removed. It's functionally no different than the abstract class and is not needed here.

PR #8828, opened 2018-10-16
Re-export VectorTile from ol/source

elemoine:
It looks like VectorTile is not re-exported from ol/source. This PR fixes that.


bartvde:
welcome back @elemoine

PR #8829, opened 2018-10-17
Change getUid return type from number to string

### LINKED ISSUE #8780: Change return type of getUid to string?
There are currently many places where we use `String(getUid(obj))` or `getUid(obj).toString()`. Wouldn't it be better to have `getUid` create and return a string instead, so we can just use `getUid(obj)`?

fredj:
fixes #8780


fredj:
Thanks for the review

PR #8832, opened 2018-10-17
Remove source foreachfeatureatcoordinate from ol/renderer/webgl/ImageLayer

fredj:
#8827 follow up

@schmidtk Can you please take a look?

### DIFF
--- a/src/ol/renderer/webgl/ImageLayer.js
+++ b/src/ol/renderer/webgl/ImageLayer.js
@@ -78,27 +78,6 @@ class WebGLImageLayerRenderer extends WebGLLayerRenderer {
   /**
    * @inheritDoc
    */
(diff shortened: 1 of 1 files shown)


schmidtk:
There may be some more cleanup here. `ol/renderer/webgl/ImageLayer~hasFeatureAtCoordinate` seems it would always return false. Should that be removed in favor of the default implementation in `ol/renderer/Layer`?

fredj:
> There may be some more cleanup here. `ol/renderer/webgl/ImageLayer~hasFeatureAtCoordinate` seems it would always return false, given the removal of `forEachFeatureAtCoordinate`. Should that be removed in favor of the default implementation in `ol/renderer/Layer`?

yes, good catch

PR #8844, opened 2018-10-23
Set crossOrigin to anonymous in mapbox-streets-v6-style

elemoine:
This PR suggests setting `crossOrigin` to `anonymous` in the options passed to the `Icon` constructor in mapbox-streets-v6-style. This is to be able to use mapbox-streets-v6-style with WebGL. I have checked that unpkg.com has `Access-Control-Allow-Origin: *` and it does.


elemoine:
Thanks Andreas!

PR #8847, opened 2018-10-23
Fix API docs

### LINKED ISSUE #8449: New API docs
API doc generation no longer works after #8345. We need to find a new tool chain to generate API docs.

### Possible options:

 1. A while ago, @tschaub and I started an effort in the [site](https://github.com/openlayers/openlayers/tree/site) branch to use the output from `tasks/generate-info.js` as starting point for a [Gatsby](https://www.gatsbyjs.org) generated API documentation. This turns out to be a huge effort, almost equivalent to writing an API doc generator from scratch.

 1. Investigate tools like [ESDoc](https://esdoc.org), [TypeDoc](http://typedoc.org) or not-yet-available tools that may come out of Microsoft's [TSDoc](https://github.com/Microsoft/tsdoc) effort.

 1. Generate `*.d.ts` files and see if there are tools to generate documentation from those.

Comments and thoughts on these options, or ideas for other options, are much appreciated.

ahocevar:
This pull request fixes the API docs and makes them work with TypeScript's `import("path/to/module").export` types.

The quality of the API docs is now equal to or better than in v4.6.5.

Fixes #8449.


tschaub:
Thanks for all the work on this @ahocevar!

config/jsdoc/api/template/static/scripts/main.js
```diff
@@ -108,4 +113,9 @@ $(function () {
 
     $(window).on('hashchange', _onHashChange);
     _onHashChange();
+
+    // Make links that JSDoc forgot clickable
+    $('p').each(function(i, p) {
+      p.innerHTML = anchorme(p.innerHTML);
+    });
```
tschaub: Is this a JSDoc bug or something we can undo later?

sbrunner:
The goal is that OpenLayer continue to use jsdoc at medium term?

ahocevar:
@sbrunner As long as there are no better options available yes.

config/jsdoc/api/template/static/scripts/main.js
ahocevar: Actually it's a bug, not a feature:
https://github.com/jsdoc3/jsdoc/blob/2c47d4b/lib/jsdoc/util/markdown.js#L59-L67

However, I'll create a follow-up pull request to handle this at doc creation time instead.

sbrunner:
OK, thanks :-)

config/jsdoc/api/template/static/scripts/main.js
ahocevar: Follow-up pull request #8852.

Issue #8601, opened 2018-09-07
Legacy build is broken


gberaudo:
Following the merging of the typescript branch the `build-legacy` target is broken:

$ `npm run build-legacy`
```
Controls.": Invalid type expression "import("../Collection.js").default<import("./Control.js").default>": Expected "!", "=", "?", "[]", "|" or end of input but "(" found.
ERROR: Unable to parse a tag's type expression for source file /home/gberaudo/dev/ol3/src/ol/control/util.js in line 42 with tag title "return" and text "{import("../Collection.js").default<import("./Control.js").default>}
Controls.": Invalid type expression "import("../Collection.js").default<import("./Control.js").default>": Expected "!", "=", "?", "[]", "|" or end of input but "(" found.
/home/gberaudo/dev/ol3/config/jsdoc/info/publish.js:98
          paramInfo.types = getTypes(param.type.names);
                                                ^

TypeError: Cannot read property 'names' of undefined
    at /home/gberaudo/dev/ol3/config/jsdoc/info/publish.js:98:49
    at Array.forEach (<anonymous>)
    at /home/gberaudo/dev/ol3/config/jsdoc/info/publish.js:93:20
```

Proposed fix:

I am currently working on a generic tool to generate a legacy library from the `default exports` and `named exports` of an ES6 modules library. That tool could be used by OpenLayers.
In a second step I plan to add `custom build` support, similar to what we had with our GCC custom builds.

Implementation details:

The code would be a simplification of current code, see draft at https://github.com/openlayers/openlayers/compare/v5.2.0...gberaudo:generate_all_exports_library.

Please let me know if there is interest / new ideas / or if someone is already working on this.

tschaub:
I know this isn't going to be popular, but I think we should stop generating the legacy build altogether.

The reality is that a single file full build of OpenLayers hasn't made sense since 2.x.  The position taken during 3.x development was essentially "the compiler will get rid of unused stuff, so its okay to add very seldom used things to the library."  This means that it became a worse and worse idea to use a single file build as the 3.x releases progressed (full build users were burdened with the addition of a half functioning renderer in addition to a working renderer, format parsers and more functionality than they were ever going to use).  The problem was that the tools to create custom builds were too much of a hassle for most people to use.

[...]

PR #8862, opened 2018-10-25
Legacy build and apidoc improvements

ahocevar:
This pull request fixes the generate-index task, which makes the build-legacy target work again. This is possible by using the new `jsdoc-plugin-typescript` that I wrote to make JSDoc support TypeScript types.

This pull request also simplifies the whole process and removes the rollup dependency, because it now uses webpack, which we have available already.

In addition to fixing the legacy build, this pull request adds a few more apidoc fixes to make them really neat again.


ahocevar:
When this is merged, the `openlayers.github.io` repo also needs to be updated. I have already prepared openlayers/openlayers.github.io#97 for that.

gberaudo:
LGTM @ahocevar.
Also fixes: #8601.

Some side notes:
- I have some warnings about `@abstract` tag, it can goes away by moving the tag to the bottom of the jsdoc comment;
- import snippets links are broken (see the Map constructor, `import Map from 'ol/Map';`).

gberaudo:
Yes you are right @ahocevar. Thanks for the explanation.

PR #8879, opened 2018-11-05
Build with CircleCI

tschaub:
Potential replacement for Travis.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,36 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:10-browsers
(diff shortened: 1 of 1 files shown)


tschaub:
We don't need to throw away our Travis config yet, but it looks like CircleCI might provide a nicer setup.  For example, I've set it up so examples are built by the CircleCI job and posted as artifacts.  Haven't read the docs to see how long these stay up yet, but this is pretty nice: https://6-4723248-gh.circle-artifacts.com/0/Examples/index.html (you need to explicitly add `index.html` to get to the index page).  This will make it easier to review changes that might affect examples.

PR #8883, opened 2018-11-05
New test runner

tschaub:
This updates the CircleCI config to run tests on ~Firefox and~ Chrome.  With this change, we no longer run tests on IE.  This is something we can revisit if it becomes a problem.

If others agree with this change, we can update the branch settings to require the CircleCI job to pass (and remove the Travis requirement).


tschaub:
Looks like some rendering tests are failing regularly on CircleCI's version of Firefox.  See https://circleci.com/gh/openlayers/openlayers/23 for example.

The rendering tests are going to need a significant overhaul with the upcoming renderer changes (we will no longer have a single target canvas element for rendering).  My plan was to separate out and simplify the rendering tests and run them with puppeteer.  It would be nice to have cross-browser rendering and unit tests, but we could try simplifying and making the tests more reliable first and then rebuild cross-browser support if/when we feel like we need it.

ahocevar:
For more reliable rendering tests, we'd also need to rework them to use larger reference images. For now, I'd say it's ok only test in Chrome.

PR #8895, opened 2018-11-08
New rendering tests

tschaub:
Before #8848, we need a new setup for rendering tests.  This branch uses Puppeteer to render simple examples, comparing actual screenshots to expected screenshots.

To run the rendering tests:
```
npm run test-rendering
```

To accept all screenshots as expected (replacing any `actual.png` with the rendered output):
```
npm run test-rendering -- --fix
```

The rendering cases should look like regular examples, with one exception.  The examples have a `render` function injected in the global scope.  Screenshots are created after this function is called (and the network is idle).

Here is an example (minus the imports):
```js
new Map({
  layers: [
    new TileLayer({
      source: new OSM()
    })
  ],
  target: 'map',
  view: new View({
    center: [0, 0],
    zoom: 0
  })
});

render();
```

Each case must include a `index.html` and `main.js`.  The screenshots are 256x256.  If we need this (or any other aspect of the test case) to be configurable, we can add support for an optional `config.json` in the same directory as the other resources.


fredj:
Thanks Tim; it is much more easy to create new rendering tests with this system !

I've tried to add a [new test case](https://github.com/fredj/openlayers/commit/7feb79ff57a136bac1fe30936c86fce40f747a93) but I have the folling error in the console:
```
$ npm run test-rendering

> ol@5.3.0 test-rendering /home/fredj/code/ol3
> node rendering/test.js

Failed to add page binding with name render: window['render'] already exists!
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! ol@5.3.0 test-rendering: `node rendering/test.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the ol@5.3.0 test-rendering script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/fredj/.npm/_logs/2018-11-08T08_40_15_827Z-debug.log
```

tschaub:
Thanks for trying this out @fredj.  I'll fix up the `render` injection now.  I also need to update the webpack config to use a common bundle that can be shared between many rendering tests.  There are a number of other things I have in mind that I think could make development of rendering tests really nice.

For example, `npm run test-rendering -- --interactive` could start running the tests in interactive mode.  In interactive mode, the runner would stop at the first failure and prompt you on how to proceed.  Options would include `skip`, `fix`, and `fix-all` – where "fix" means copy actual to expected.  After running through tests, the runner would pause.  During interactive mode, you open a browser to view/debug any of the test cases.  After modifying or adding tests, the runner would start again.  In addition, by default the runner will only run test cases where mtime of library sources or test case is newer than mtime of actual screenshot (and a `--force` flag could force all test cases to run).

tschaub:
I think we'll want to continue improving on this, but I also think it makes sense to get it in and start porting over old rendering tests.

You can see the available options by running `node rendering/test.js --help`.  The `interactive` option can be significantly improved, but for now it just leaves the test server running.  For now, it can be useful to debug a failing test.  After running `node rendering/test.js --interactive --log-level info`, you can load the available test cases at `http://localhost:3000`.

By default, only outdated tests run.  So on the second run, `node rendering/test.js` should be fast.

PR #8637, opened 2018-09-14
Image draw buffer

benVigie:
As discussed in #8473, we have a clipping effect when panning on image `VectorLayer`. It's because we draw images with viewport size by default.
    
To remove this clipping effect, I created a new parameter for `VectorLayer` to specify a buffer to extend the viewport size drawing.
    
Please note that this parameter will need to increase `renderBuffer` (if smaller) to retrieve buffer area features. Obviously, it has an impact on performances as the new drawing zone is bigger than the default one. So the default parameter is no buffer, as today.

PR #8897, opened 2018-11-08
Add imageRatio option for vector layer image mode

ahocevar:
Alternative to #8637. The buffer is specified as a ratio here, which is more in line with the rest of the library, and we do not add to the `frameState` object.

To avoid nasty conflicts there, this should not be merged before #8848. This will need to be reworked after #8848 is merged, and tests should also be added.


fredj:
@ahocevar I've adapted the code for master: https://github.com/openlayers/openlayers/compare/master...fredj:image-vector-ratio
If it's the right change I will add tests

PR #8901, opened 2018-11-09
Fixes issue openlayers#8863

timkeane:
replace ol/geom/Geometry#containsXY with an implementation suitable for all point and linear geometry types

PR #8910, opened 2018-11-12
Implements containsXY for all geometries

### LINKED ISSUE #8863: ol/source/Vector#getFeaturesAtCoordinate and implementations of ol/geom/Geometry#containsXY
While trying to use ```ol/source/Vector#getFeaturesAtCoordinate``` to retrieve an array of coincident point features at a given coordinate I found that the lack of an override implementation of ```ol/geom/Geometry#containsXY``` always results in an empty array even when using the exact coordinate of a given feature's point geometry.  What is the reasoning behind not implementing ```containsXY``` for ```Point```, ```MultiPoint```, ```LineString``` and ```MultiLineString``` geometries?

I can achieve my goals by providing my own override as follows, but I am wondering why such implementations are excluded, and if I should be taking a different approach to getting my desired result:

```
import Point from 'ol/geom/Point'

Point.prototype.containsXY = function(x, y) {
  const coord = this.getCoordinates()
  return coord[0] === x && coord[1] === y
}
```

fredj:
replaces #8901
fixes #8863

All code from @timkeane

### DIFF
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -92,7 +92,8 @@ class Geometry extends BaseObject {
    * @return {boolean} Contains (x, y).
    */
   containsXY(x, y) {
-    return false;
+    const coord = this.getClosestPoint([x, y]);
+    return coord[0] === x && coord[1] === y;
   }
 
   /**
--- a/src/ol/geom/LineString.js
+++ b/src/ol/geom/LineString.js
@@ -90,16 +90,6 @@ class LineString extends SimpleGeometry {
 
   /**
    * @inheritDoc
-   * @override
-   * @api
-   */
-  containsXY(x, y) {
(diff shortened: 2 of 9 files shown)


src/ol/geom/Geometry.js
```diff
@@ -92,7 +92,8 @@ class Geometry extends BaseObject {
    * @return {boolean} Contains (x, y).
    */
   containsXY(x, y) {
-    return false;
+    const coord = this.getClosestPoint([x, y]);
```
tschaub: We can do significantly less work than this function to determine intersection.  But this looks like a good start.

timkeane:
originally i had planned on doing a test on Point and  MultiPoint for x===x
&& y===y, and doing a y=mx+b test on Line and PolyLine but this was
quicker.  is the previous preferred?

PR #8911, opened 2018-11-12
Use a temporary array in stableSort

fredj:
Note: I've haven't checked if this change really reduce the garbage size. (but this function is called a lot)


fredj:
sorting could be removed instead: #8915

PR #8912, opened 2018-11-12
Rotation support in the composite renderer

tschaub:
This adds support for rotated views to the composite renderer.

Currently, labels are rotated with the view (and they are double rotated if `rotateWithView: true` in a text symbolizer).  It will be straightforward to support unrotated labels, but this will break the current canvas map renderer.  We could add another argument to the replay method to support both the old and the new rendering types, but I think we should instead get a few other layer renderers working and then decide if we want to continue supporting multiple map rendering strategies.


src/ol/renderer/canvas/VectorLayer.js
```diff
@@ -275,12 +278,7 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
     const viewHints = frameState.viewHints;
     const snapToPixel = !(viewHints[ViewHint.ANIMATING] || viewHints[ViewHint.INTERACTING]);
 
-    // TODO: deal with rotation (this should not be necessary)
-    if (rotation) {
-      rotateAtOffset(context, -rotation, width / 2, height / 2);
-    }
-
-    let transform = this.getTransform(frameState, 0);
+    let transform = this.getRenderTransform(frameState, [width, height], 0);
```
fredj: instead of creating a array we could pass width and height to avoid creating temporary arrays

src/ol/renderer/Composite.js
```diff
@@ -30,9 +30,13 @@ class CompositeMapRenderer extends MapRenderer {
      * @type {HTMLDivElement}
      */
     this.element_ = document.createElement('div');
+    const style = this.element_.style;
+    style.display = 'flex';
```
fredj: we will lose ie < 10 support (do we still care ?)

tschaub: As discussed, if someone comes along with funds for IE 9 support, we can implement this in a different way.

src/ol/renderer/canvas/VectorLayer.js
tschaub: Reworked in afc64258afeb1e2c2a7fed25aa742316b2e0ff23.

PR #8913, opened 2018-11-12
Implement renderFrame function for intermediate canvas renderer

fgravin:
Follow up of https://github.com/openlayers/openlayers/pull/8848

Implement `renderFrame()` function for composite map renderer in Intermediate canvas renderer and all subclasses.

For `TileLayer` we should consider removing inheritance to IntermediateCanvas.

### DIFF
--- a/src/ol/renderer/Composite.js
+++ b/src/ol/renderer/Composite.js
@@ -37,6 +37,7 @@ class CompositeMapRenderer extends MapRenderer {
     style.width = '100%';
     style.height = '100%';
 
+    this.element_.style.position = 'relative';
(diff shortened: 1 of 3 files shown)


src/ol/renderer/Composite.js
```diff
@@ -37,6 +37,7 @@ class CompositeMapRenderer extends MapRenderer {
     style.width = '100%';
     style.height = '100%';
 
+    this.element_.style.position = 'relative';
```
tonio:
```suggestion
    style.position = 'relative';
```

PR #8914, opened 2018-11-12
[wip] Add new className property to ol/layer/Base

fredj:
for be able to set CSS properties to layer elements(for example `filter`)


jingyebuchen:
I modeled your example but it didn't work. I just want the loaded map basemap layer to have a filter effect. Can you help me?

jingyebuchen:
> for be able to set CSS properties to layer elements(for example `filter`)

Or can you give me a complete example for my reference? I will thank you very much.

fredj:
This feature is only present in master, not yet in a release.

jingyebuchen:
I solved the problem based on your help, thank you very much.

PR #8915, opened 2018-11-12
Use CSS z-index to order layers

fredj:


src/ol/renderer/Composite.js
```diff
@@ -98,6 +95,10 @@ class CompositeMapRenderer extends MapRenderer {
       const layerRenderer = this.getLayerRenderer(layer);
       if (layerRenderer.prepareFrame(frameState, layerState)) {
         const element = layerRenderer.renderFrame(frameState, layerState);
+        const zIndex = layerState.zIndex;
+        if (zIndex !== element.style.zIndex) {
```
tonio: Is this check really necessary? I don't think set an identical `zIndex` will have an effect…

fredj: My idea is that changing the `zIndex` even if it's the same value will trigger a reflow or something similar (but I haven't tried if it's really the case)

tonio: I just test to confirm, looks like it isn't the case.

tschaub:
@fredj it should be pretty easy to fix up these conflicts if we want to get this in.

fredj:
@tschaub done

PR #8919, opened 2018-11-13
Re-use style local variable

fredj:
thanks @tonio


fredj:
Because of the flex positioning, the `position = 'relative';` is not needed anymore

PR #8923, opened 2018-11-13
Use the composite renderer

### LINKED ISSUE #8514: Efficiently rendering hundreds of unique vector map features
I'm working on an application that needs to render hundreds of unique vector map features on ol.layer.Vector layers with renderMode set to "image" on a canvas-based map. Each feature has three things that make it unique:

1. Icon
2. Associated polygonal geometry
3. Textual label

Of the three things, the icon that's used tends to be reused between features. So, although I'm putting hundreds of features on the map there may only be a dozen unique icons in use.

The second and third things, though, will be unique for each feature that gets drawn.

With my current implementation I'm creating a new ol.style.Style object for each feature (because each one has a unique appearance), but my performance is quite bad (in the single digits for FPS). I've commented that logic out from my application and instead rendered everything using a single layer-defined ol.style.Style and my performance is perfect (~60 FPS), but unfortunately I can't move forward with that approach because the features are no longer drawn uniquely.

Am I missing something that would alleviate this issue and give me good performance with so many unique features?

tschaub:
This makes `ol/Map` use the composite renderer.

 - [x] make min/max zoom work for sources (TS)
 - [x] make vector tile layer renderer return two canvases (TS)
 - [x] make ~~unmanaged~~ empty vector layers work (AH) (also fixes #8514)
 - [x] rotation for image layers (TS)
 - [x] make vector image layers work
 - [x] implement `forEachLayerAtPixel` for all canvas layer renderers
 - [x] make rotation work for text
 - [x] remove intermediate canvas layer renderer
 - [x] make graticule a vector layer (OG)
 - [x] fix the overview map (OG)
 - [x] add function to create vector context from render event (AH)
 - [x] rework examples that rely on pre/post compose
 - [x] get all tests passing
 - [x] document breaking changes


examples/dynamic-data.js
```diff
@@ -4,14 +4,14 @@ import {MultiPoint, Point} from '../src/ol/geom.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import OSM from '../src/ol/source/OSM.js';
 import {Circle as CircleStyle, Fill, Stroke, Style} from '../src/ol/style.js';
+import {getVectorContext} from '../src/ol/render';
```
fredj: We should be consistant and always have `.js` in our import.
Please cherry-pick:  https://github.com/openlayers/openlayers/commit/0eaccbb0eb97fe59258377d74b4f29155913e6ff

tschaub: Done in 2eede7f46591937d4bdcec16b54cbd23da15283e.

tschaub:
Thanks for the collective effort!

marcjansen:
Very nice work everyone! Thanks!

PR #8924, opened 2018-11-13
Reuse vector tile replays for interim tiles

ahocevar:
To improve rendering quality, we currently create replay groups for an interim tile's target resolution from source tiles with a different resolution.

This pull request changes things so we reuse an existing replay in this case, which improves performance after zooming and when panning on a map that has lower resolution tiles already in the current viewport.

I also changed the `getValues` function in `ol/obj` to use JS `Object.values` where available.


fredj:
For the failing tests I think we should do the same as `Object.values` and not accepting `null` or `undefined`

ahocevar:
Updated the pull request to not touch `ol/obj#getValues`, and be smarter about which low res replay group to choose.

PR #8928, opened 2018-11-13
Add vector tile example using ol-mapbox-style

ahocevar:
This simply adds an example that shows how to use ol-mapbox style and how a vector tile map from tilehosting.com would look.


examples/mapbox-style.html
```diff
@@ -0,0 +1,32 @@
+---
+layout: example-verbatim.html
+title: Vector tiles creatde from a Mapbox Style object
```
fredj:
```suggestion
title: Vector tiles created from a Mapbox Style object
```

examples/mapbox-style.html
```diff
@@ -0,0 +1,32 @@
...
+shortdesc: Example of using ol-mapbox-style with tiles from tilehosting.com.
+tags: "fullscreen, geolocation, mobile"
```
fredj:
```suggestion
tags: "vector tiles"
```

PR #8929, opened 2018-11-13
Use nodejs tasks for build instead of `cp`, `rm` and `mkdir`

FilipJirsak:
On Windows `npm run build-package` fails because it depends on unix utilities `cp`, `rm` and `mkdir`. These utilities can be installed into Windows for example with Cygwin, but they are not installed in common.
This patch replaces this with `nodejs` scripts which do the same thing but they are platform independent.


fredj:
@FilipJirsak I've commit a fix that used `shx` instead of adding new scripts. See #8950

PR #8931, opened 2018-11-13
Add Mapbox-gl-js example

fgravin:
Based on composite map renderer (https://github.com/openlayers/openlayers/pull/8922), add an example on how to integrate a mapbox-gl-js layer.

A new layer class is created and implement the `render()` method that returns the canvas of the mapbox-gl-js map.
Synchronisation between the 2 maps is done by css to avoid latency.

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -10,6 +10,7 @@
     "GyroNorm": false,
     "jsPDF": false,
     "jsts": false,
+    "mapboxgl": false,
     "saveAs": false,
     "toastr": false,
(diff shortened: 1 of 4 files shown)


tschaub:
Looks great: https://170-4723248-gh.circle-artifacts.com/0/examples/mapbox-layer.html

(We should look into having CircleCI post the [example index page](https://170-4723248-gh.circle-artifacts.com/0/examples/) back to GitHub - maybe as a check, maybe as a comment).

PR #8938, opened 2018-11-14
Port render-toContext rendering test

fredj:


tschaub:
Looks like we might need to add some tolerance when asserting that screenshots match.

PR #8940, opened 2018-11-14
Make layers mandatory in the OverviewMap control

jahow:
Layers must now always be specified for the control.
Base example was amended to reflect that.

The breaking change was added in the update notes.

Note: this was done to be compatible with the composite layer renderer, see #8923

### DIFF
--- a/examples/overviewmap.js
+++ b/examples/overviewmap.js
@@ -4,13 +4,22 @@ import {defaults as defaultControls, OverviewMap} from '../src/ol/control.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import OSM from '../src/ol/source/OSM.js';
 
+const source = new OSM();
+const overviewMapControl = new OverviewMap({
+  layers: [
+    new TileLayer({
+      source: source
+    })
+  ]
+});
+
 const map = new Map({
   controls: defaultControls().extend([
-    new OverviewMap()
+    overviewMapControl
   ]),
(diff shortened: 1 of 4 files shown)


examples/overviewmap.js
```diff
@@ -4,13 +4,22 @@ import {defaults as defaultControls, OverviewMap} from '../src/ol/control.js';
...
+
 const map = new Map({
   controls: defaultControls().extend([
-    new OverviewMap()
+    overviewMapControl
   ]),
   layers: [
     new TileLayer({
-      source: new OSM()
+      source
```
fredj:
```suggestion
      source: source
```
To stick with es5

src/ol/control/OverviewMap.js
```diff
@@ -45,8 +45,7 @@ const MIN_RATIO = 0.1;
  * @property {string|HTMLElement} [label='»'] Text label to use for the collapsed
  * overviewmap button. Instead of text, also an element (e.g. a `span` element) can be used.
  * @property {Array<import("../layer/Layer.js").default>|import("../Collection.js").default<import("../layer/Layer.js").default>} [layers]
```
fredj:
```suggestion
  * @property {Array<import("../layer/Layer.js").default>|import("../Collection.js").default<import("../layer/Layer.js").default>} layers
```

To mark the parameter as mandatory

PR #8942, opened 2018-11-14
[wip] Investigation on webworker bundling

fgravin:
This is not supposed to be merged but propose a way to bundle the webworker within the library.

The idea is 

- To build separatly the worker during build phase (2 options with webpack or rollup).
- Then, the built worker files are stringified and exported to ol code as a string.
- The worker is then instanciated from this string content using blob format.

```
npm run build-worker
```

In this PR, the worker is used to manage the fetch and parsing of vector tiles.

Beside this PR, i tested a rollup application exemple using vector tiles, the webworker is actually used.


package.json
```diff
@@ -36,6 +38,7 @@
     "url": "https://github.com/openlayers/openlayers/issues"
   },
   "dependencies": {
+    "mapbox-gl": "^0.51.0",
```
gberaudo: You should remove this extra dependency.

src/ol/worker/loader.js
```diff
@@ -0,0 +1,43 @@
+import MVT from "../format/MVT";
+import RenderFeature from "../render/Feature";
+import OLWorker from './export';
+
+var blob;
+try {
+  blob = new Blob([OLWorker], {type: 'application/javascript'});
+} catch (e) { // Backwards-compatibility
+  blob = new MSBlobBuilder();
```
gberaudo: The catch should not be necessary, IE10 seems to support standard Blob. See https://developer.mozilla.org/fr/docs/Web/API/Blob.

PR #8943, opened 2018-11-15
Port text style rendering tests

fredj:
Only points for now


rendering/cases/text-style/main.js
```diff
@@ -0,0 +1,124 @@
+import Map from '../../../src/ol/CompositeMap.js';
```
tschaub: If we port these tests using the regular `ol/Map`, then we can more easily rebase #8923.

```suggestion
import Map from '../../../src/ol/Map.js';
```

PR #8944, opened 2018-11-15
Use a layer for the graticule instead of a control

jahow:
The graticule functionality has been migrated to use a `VectorLayer` instead of a control relying on `postcompose`.
This was mainly done to be compatible with the new composite layer rendering (see #8923).

The `Graticule` layer derives from `VectorLayer` and implements a custom loading function that will populate the feature source with lines & labels according to the current extent.

The two examples have been updated to reflect this, and the `Graticule` control was removed.

Two new features have been added to the graticule along the way:
* if an extent is defined for the layer, the graticule will only be rendered inside it
* a `wrapX` option have been added, default is true (as in vector sources)

As such this shoud close https://github.com/openlayers/openlayers/issues/8427 as well.


fredj:
please add an entry in changelog/upgrade-notes.md

jahow:
> please add an entry in changelog/upgrade-notes.md

Just added a commit with documentation about the change.

tschaub:
Thanks @jahow.

PR #8946, opened 2018-11-15
Make the map renderer tolerant of layers without renderers

tschaub:
People who extend our `ol/layer/Layer` may not implement `createRenderer`.  In this case, `getRenderer` returns `null`.  The map renderer should be tolerant to this type of layer.


src/ol/renderer/Map.js
```diff
@@ -196,33 +196,24 @@ class MapRenderer extends Disposable {
...
     this.layerRendererListeners_[layerKey] = listen(renderer, EventType.CHANGE, this.handleLayerRendererChange_, this);
     return renderer;
   }
 
-  /**
-   * @param {string} layerKey Layer key.
-   * @protected
-   * @return {import("./Layer.js").default} Layer renderer.
-   */
-  getLayerRendererByKey(layerKey) {
```
fredj: good catch

PR #8950, opened 2018-11-15
Use shx to run unix commands in package.json

fredj:
this is an alternative to #8929


marcjansen:
Every day I learn sth. new. Thanks :+1:

fredj:
Whoa, so much approval!

PR #8951, opened 2018-11-15
Remove legacy WebGL code & implement a lightweight API for point rendering

jahow:
This PR aims at simplifying the codebase by removing the bulk of the WebGL-related classes:
* Layer renderers
* Replays & replay groups
* Utilities
* Atlas management
* ...

What is kept is a `WebGLHelper` class that acts as a wrapper around the `gl` API, offering the following possibilities:
* Creating and binding data buffers
* Compiling and binding shaders
* Creating and binding textures
* Rendering primitives

A new layer renderer was also added: `WebGLPointsRenderer`, which only renders point geometries without using the traditional styles. Instead, the user can define custom shaders that will give full control over the shape and color of the points.

The user can also define a post processing shader. This was used to rewrite the `Heatmap` layer to get better performance overall. See the example [here](https://320-4723248-gh.circle-artifacts.com/0/examples/heatmap-earthquakes.html).

To do:
- [x] Offer an entry point for binding textures either for point rendering or post processing
- [x] Complete the `Heatmap` rewrite by passing the gradient texture to the points renderer
- [x] General improvement of the `WebGLHelper` API
- [x] Actually remove the legacy code & tests
- [x] Document
- [x] Add unit ~~& rendering~~ tests


jahow:
@tschaub @ahocevar this *should* be good to go... No rendering test for now as the webgl rendering does not work in puppeteer. Also the heatmap example fails in firefox, for browser issues most likely (see [search](https://www.google.ch/search?q=firefox+GL_ARB_gpu_shader5+not+supported&oq=firefox+GL_ARB_gpu_shader5+not+supported&aqs=chrome..69i57j33.2243j0j7&client=ubuntu&sourceid=chrome&ie=UTF-8)). 

There is still a lot of work to do on this topic of course.

examples/heatmap-earthquakes.js
```diff
@@ -1,4 +1,4 @@
-import Map from '../src/ol/Map.js';
+import Map from '../src/ol/CompositeMap.js';
```
tschaub:
```suggestion
import Map from '../src/ol/Map.js';
```

tschaub:
Thanks for the huge effort here @jahow!

tschaub:
@jahow - it looks like there are some issues to work out with the [earthquakes heatmap example](https://343-4723248-gh.circle-artifacts.com/0/examples/heatmap-earthquakes.html).

jahow:
@tschaub I could not reproduce this issue either with the provided link or on master. Could you maybe describe which browser you are using, and if the issue is present on all browsers? Thanks

PR #8958, opened 2018-11-16
Remove extra key in restore_cache

fredj:
Only `v1-dependencies-{{ checksum "package.json" }}` key is used in save_cache. Or do I miss something?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,9 +10,7 @@ jobs:
       - checkout
 
       - restore_cache:
-          keys:
-          - v1-dependencies-{{ checksum "package.json" }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -10,9 +10,7 @@ jobs:
       - checkout
 
       - restore_cache:
-          keys:
-          - v1-dependencies-{{ checksum "package.json" }}
-          - v1-dependencies-
+          key: v1-dependencies-{{ checksum "package.json" }}
```
tschaub: The second one is a fallback [if the checksum fails](https://circleci.com/docs/2.0/caching/#example-caching-configuration).

Issue #4251, opened 2015-10-12
ZoomSlider Control Fractional Zoom


GeorgeNarcis:
It seems that currently ZoomSlider Control doesn't support fractional zoom. Looking at the source code  I saw that the resolution is constrained using view.constrainResolution. 
For my use case I really needed the fractional zoom functionality so I manually overwrote ol functions.
 I was wondering if this should't be a configurable option, to enable or not the fractional zoom for the slider ?

Note: I am using Zoomify source for an image viewer

ahocevar:
This is something you can easily configure on the view:
```js
map.getView().contrainResolution = function(resolution) {
  return resolution;
};
```
If you are already doing this, you're doing it the right way.

GeorgeNarcis:
If I use this function, I lose the ability to change the resolution, the image can t be zoomed any more using the mouse wheel, the resolution remaining constant.

ahocevar:
In this case, you'll have to conditionally reuse the original `ol.View#constrainResolution()` function, e.g.
```js
var view = map.getView();
var constrainResolution = view.constrainResolution;
view.constrainResolution = function(resolution, delta, direction) {
  return delta ?
    constrainResolution.call(view, resolution, delta, direction) :
    resolution; 
}
```

ahocevar:
It may make sense to add a `constrainCenter` boolean property. If set to false, you would get fractional zoom.

raiyni:
Would it make sense to add an option `contrainResolution` and if that's true, constrainResolution

here https://github.com/openlayers/openlayers/blob/master/src/ol/control/ZoomSlider.js#L284

and 

here https://github.com/openlayers/openlayers/blob/master/src/ol/control/ZoomSlider.js#L222

otherwise, animate to exact resolution

PR #8959, opened 2018-11-16
Zoomslider: add target and constrain resolution options

raiyni:
Closes https://github.com/openlayers/openlayers/issues/4251

### DIFF
--- a/src/ol/control/ZoomSlider.js
+++ b/src/ol/control/ZoomSlider.js
@@ -26,10 +26,13 @@ const Direction = {
 
 /**
  * @typedef {Object} Options
+ * @property {boolean} [constrainResolution=true] Constrain zoom level to an integer.
  * @property {string} [className='ol-zoomslider'] CSS class name.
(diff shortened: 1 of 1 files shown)


src/ol/control/ZoomSlider.js
```diff
@@ -26,10 +26,12 @@ const Direction = {
 
 /**
  * @typedef {Object} Options
+ * @property {boolean} [constrainResolution=true] Constrain zoom level to an integer.
  * @property {string} [className='ol-zoomslider'] CSS class name.
  * @property {number} [duration=200] Animation duration in milliseconds.
  * @property {function(import("../MapEvent.js").default)} [render] Function called when the control
  * should be re-rendered. This is called in a `requestAnimationFrame` callback.
+ * @property {HTMLElement|string} [target] Specify a target if you want the control to be
```
tschaub: It looks like there should be some more words in this description.

raiyni: You are correct.

PR #8926, opened 2018-11-13
Uncouple replay creation and rendering

gberaudo:
The goal of this PR is to split the replays in 2 parts:
- classes for building instructions (typically used in web workers);
- classes for executing the instructions (typically used on the main thread).

Plan is:
- [x] have a replay group instance for building instructions and one for executing them;
- [x] be able to extract and replace instructions;
- [x] physically split the builders and executor in separate classes;
- [ ] only pass information in instructions and coordinate array;

Postponed to a later phase:
- [ ] make instructions serializable (remove image creation from the styles, create and cache images in the instruction executor);
- [ ] binary transferable representation of the instructions (will be a lot of work)

PR #8960, opened 2018-11-16
Avoid creating a new function for measuring text

gberaudo:
Follow up optimization. See #8926.


gberaudo:
@ahocevar, as promised, this PR get rid of the function creation.

To keep tests easy to understand I have introduced a mockable `measure` function. In addition I moved the code out of the flat geometry directory since it is now explicitly depending on canvas.

Does it looks good to you?

tschaub:
Thanks for the work on this @gberaudo.  We collaborated on an alternative solution and decided to merge #8988.

gberaudo:
Thanks @ahocevar and @tschaub, your solution is better.
I will try to find some time this week to continue working on moving image loading from style creation to the renderer.

PR #8968, opened 2018-11-17
Use Object.values if available

fredj:
As `Object.values`, the arguments now can't be `null` or `undefined`.


test/spec/ol/objectutil.test.js
```diff
@@ -58,7 +58,7 @@ describe('ol.obj.getValues()', function() {
 
   it('gets a list of property values from an object', function() {
     expect(getValues({foo: 'bar', num: 42}).sort()).to.eql([42, 'bar']);
-    expect(getValues(null)).to.eql([]);
+    expect(getValues([])).to.eql([]);
```
tschaub: I think you mean {} here, but both work apparently.

fredj: Yes, I meant {}  !

PR #8970, opened 2018-11-17
Fix zIndex positioning

ahocevar:
This pull request fixes layer positioning so layers with a z-index do not overlay the controls container.

### DIFF
--- a/src/ol/renderer/Composite.js
+++ b/src/ol/renderer/Composite.js
@@ -31,6 +31,8 @@ class CompositeMapRenderer extends MapRenderer {
     const style = this.element_.style;
     style.width = '100%';
     style.height = '100%';
+    style.position = 'relative';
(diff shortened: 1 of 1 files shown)


ahocevar:
Tests fixed and ready for review.

tschaub:
Definitely fixes the problem. Really, all children of the viewport should have absolute position.  And then we wouldn’t need z-index 0 here.  We need to rework our “stop event” container for this to work though.

Ok by me to merge and continue reworking later.  Hard to explain why this would have relative position though.

ahocevar:
Thanks for the review @tschaub. I tried your suggestions, but couldn't get them to work. Maybe I misunderstood. But the one thing that did work was changing the position from `'relative'` to `'absolute'`, so I did that.

PR #8976, opened 2018-11-18
Scale tiles instead of canvas

### LINKED ISSUE #8791: How to change the marker icon direction on map?
Dear experts,
I have integrated openlayers in one of my project where I am adding marker icons like pushpin inside of a polygon. Pushpin or marker icons adding automatically calculating the space by rows and columns inside the polygon. Is it possible to change the direction of pushpin or marker icon according to the shape of polygon? 

Has any feature in openlayers code base to make it possible?

### LINKED ISSUE #8971: drawImage slow in tile layer renderer when rendering lower z tiles or when rotating
The current implementation of the tile layer renderer creates a canvas 1.5 times the size of the viewport when using lower z tiles, and even bigger when rotating.

This causes a significant slowdown in rendering performance for HiDPI tiles, i.e. for vector tiles and for raster tiles when the server provides them (e.g. the `wmts-hidpi` example). 

The slowdown is caused by too much time spent in drawImage operations, which occur with every render frame.

To reproduce the issue, just open the `mapbox-style` example, which initially is set to z 11.6. Panning is very slow at the initial zoom (which triggers a 1.5 scale), but when zooming to an integer zoom level (e.g. with +/-), which results in a canvas with the size of the viewport, panning is smooth again. The same can be observed when rotating, and the result is worst when at the initial zoom and rotated.

The previous implementation did have a large canvas too (the intermediate canvas), but that was only recreated when its tile range did not contain the full viewport, and not with every render frame.

@tschaub, do you have an idea for a fix of the current implementation, or do we need to go back to using an intermediate canvas?

ahocevar:
After this change, the CSS scale for the canvas tile layer only accounts for the tile pixel ratio. The scale that results from the ratio between view resolution and tile resolution is applied by using scaled offsets and sizes in `drawImage`. To avoid sub-pixel gaps between tiles, the `dx`, `dy`, `dw` and `dh` arguments of the `drawImage` call are integers and aligned with the next tile in row/column.

Panning - before this change:
before

Panning - after this change:
after

Fixes #8971.

### DIFF
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -6,7 +6,7 @@ import TileRange from '../../TileRange.js';
 import TileState from '../../TileState.js';
 import {createEmpty, getIntersection, getTopLeft} from '../../extent.js';
 import CanvasLayerRenderer from './Layer.js';
-import {compose as composeTransform, makeInverse, toString as transformToString} from '../../transform.js';
+import {apply as applyTransform, compose as composeTransform, makeInverse, toString as transformToString} from '../../transform.js';
 
 /**
  * @classdesc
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for the review!

PR #8981, opened 2018-11-19
Do not create a new measure function with every call

ahocevar:
Alternative to #8960.

### DIFF
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -325,6 +325,22 @@ export function measureTextWidth(font, text) {
 
 
 /**
+ * Measure text width using a cache.
+ * @param {string} text The text to measure.
+ * @param {string} font The font.
+ * @param {Object<string, number>} cache A lookup of cached widths by text.
+ * @returns {number} The text width.
+ */
+export function measureAndCacheTextWidth(text, font, cache) {
+  if (text in cache) {
+    return cache[text];
+  }
+  const width = cache[text] = measureTextWidth(font, text);
+  return width;
+}
+
+
+/**
  * @param {string} font Font to use for measuring.
  * @param {Array<string>} lines Lines to measure.
  * @param {Array<number>} widths Array will be populated with the widths of
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -22,7 +22,7 @@ import {
 
 
 import {createCanvasContext2D} from '../../dom.js';
-import {labelCache, defaultTextAlign, measureTextHeight, measureTextWidth, measureAndCacheTextWidth, measureTextWidths} from '../canvas.js';
+import {labelCache, defaultTextAlign, measureTextHeight, measureAndCacheTextWidth, measureTextWidths} from '../canvas.js';
 
 
 /**
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -22,7 +22,7 @@ import {
 
 
 import {createCanvasContext2D} from '../../dom.js';
-import {labelCache, defaultTextAlign, measureTextHeight, measureTextWidth, measureTextWidths} from '../canvas.js';
+import {labelCache, defaultTextAlign, measureTextHeight, measureTextWidth, measureAndCacheTextWidth, measureTextWidths} from '../canvas.js';
 
 
 /**
@@ -736,29 +736,22 @@ class CanvasExecutor {
 
           const textState = this.textStates[textKey];
           const font = textState.font;
-          const textScale = textState.scale;
+          const textScale = textState.scale * measurePixelRatio;
 
-          let widths = this.widths_[font];
-          if (!widths) {
-            this.widths_[font] = widths = {};
+          let cachedWidths;
+          if (font in this.widths_) {
+            cachedWidths = this.widths_[font];
+          } else {
+            cachedWidths = this.widths_[font] = {};
           }
 
-          //FIXME Do not create this function on every call
-          const measure = function(text) {
-            let width = widths[text];
-            if (!width) {
-              width = widths[text] = measureTextWidth(font, text);
-            }
-            return width * textScale * measurePixelRatio;
-          };
-
           const pathLength = lineStringLength(pixelCoordinates, begin, end, 2);
-          const textLength = measure(text);
+          const textLength = textScale * measureAndCacheTextWidth(font, text, cachedWidths);
           if (overflow || textLength <= pathLength) {
             const textAlign = this.textStates[textKey].textAlign;
             const startM = (pathLength - textLength) * TEXT_ALIGN[textAlign];
             const parts = drawTextOnPath(
-              pixelCoordinates, begin, end, 2, text, measure, startM, maxAngle);
+              pixelCoordinates, begin, end, 2, text, startM, maxAngle, measureAndCacheTextWidth, font, textScale, cachedWidths);
             if (parts) {
               let c, cc, chars, label, part;
               if (strokeKey) {
--- a/src/ol/geom/flat/textpath.js
+++ b/src/ol/geom/flat/textpath.js
@@ -3,22 +3,23 @@
  */
 import {lerp} from '../../math.js';
 
-
 /**
  * @param {Array<number>} flatCoordinates Path to put text on.
  * @param {number} offset Start offset of the `flatCoordinates`.
  * @param {number} end End offset of the `flatCoordinates`.
  * @param {number} stride Stride.
  * @param {string} text Text to place on the path.
- * @param {function(string):number} measure Measure function returning the
- * width of the character passed as 1st argument.
  * @param {number} startM m along the path where the text starts.
  * @param {number} maxAngle Max angle between adjacent chars in radians.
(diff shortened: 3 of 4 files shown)


src/ol/render/canvas/Executor.js
```diff
@@ -47,6 +47,20 @@ const tmpExtent = createEmpty();
...
+let measureFont, measureScale, measureWidths;
+/**
+ * @param {string} text Text.
+ * @return {number} Measured width, using `measureFont`, `measureScale` and `measureWidths`.
+ */
+function measure(text) {
+  let width = measureWidths[text];
+  if (!width) {
+    width = measureWidths[text] = measureTextWidth(measureFont, text);
+  }
```
tschaub: Do we ever get `0` from `measureTextWidth`?

Might be nicer to
```js
if (!(text in measuredWidths)) {
    measuredWidths[text] = measureTextWidth(measureFont, text);
}
return measuredWidths[text] * measureScale;
```

Or with `let width` to avoid the second lookup.

src/ol/render/canvas/Executor.js
```diff
@@ -735,25 +749,14 @@ class CanvasExecutor {
...
 
           const textState = this.textStates[textKey];
-          const font = textState.font;
-          const textScale = textState.scale;
-
-          let widths = this.widths_[font];
-          if (!widths) {
-            this.widths_[font] = widths = {};
+          measureFont = textState.font;
+          measureWidths = this.widths_[measureFont];
```
tschaub: I think `currentFont` and `measuredWidths` make sense for these names.

tschaub:
If we want to avoid having state in the Executor module, we could have a function that measured and cached the width.  And we would call it with all the arguments it needs.
```js
/**
 * Measure text width using a cache.
 * @param {string} text The text to measure.
 * @param {string} font The font.
 * @param {Object<string, number>} cache A lookup of cached widths by text.
 * @returns {number} The text width.
 */
export function measureAndCacheTextWidth(text, font, cache) {
  if (text in cache) {
    return cache[text];
  }
  const width = cache[text] = measureTextWidth(font, text);
  return width;
}
```

This would mean that `drawTextOnPath` would need to be called with the cache and the scale instead of the measure function.

ahocevar:
> This would mean that drawTextOnPath would need to be called with the cache and the scale instead of the measure function.

This would not work, because drawTextOnPath needs to measure text chunks that have not been measured before. Or am I missing something?

tschaub:
So we'd have to pass a bit more to `drawTextOnPath` to keep it from being (canvas) implementation specific.  Basically, instead of passing it a `measure` function that relies on state elsewhere, we could pass it a `measureAndCacheTextWidth` function along with all the args (minus `text`) that the function needs.

Something like this:

```diff
diff --git a/src/ol/geom/flat/textpath.js b/src/ol/geom/flat/textpath.js
index f358d8791..8ffa2c28a 100644
--- a/src/ol/geom/flat/textpath.js
+++ b/src/ol/geom/flat/textpath.js
@@ -3,22 +3,23 @@
  */
 import {lerp} from '../../math.js';
 
-
 /**
  * @param {Array<number>} flatCoordinates Path to put text on.
  * @param {number} offset Start offset of the `flatCoordinates`.
  * @param {number} end End offset of the `flatCoordinates`.
  * @param {number} stride Stride.
  * @param {string} text Text to place on the path.
- * @param {function(string):number} measure Measure function returning the
- * width of the character passed as 1st argument.
  * @param {number} startM m along the path where the text starts.
  * @param {number} maxAngle Max angle between adjacent chars in radians.
+ * @param {function(string, string, Object<string, number>):number} measureAndCacheTextWidth Measure and cache text width.
+ * @param {string} font The font.
+ * @param {number} scale The product of the text scale and the device pixel ratio.
+ * @param {Object<string, number>} cache A cache of measured widths.
  * @return {Array<Array<*>>} The result array of null if `maxAngle` was
  * exceeded. Entries of the array are x, y, anchorX, angle, chunk.
  */
 export function drawTextOnPath(
-  flatCoordinates, offset, end, stride, text, measure, startM, maxAngle) {
+  flatCoordinates, offset, end, stride, text, startM, maxAngle, measureAndCacheTextWidth, font, scale, cache) {
   const result = [];
 
   // Keep text upright
@@ -41,7 +42,7 @@ export function drawTextOnPath(
     index = reverse ? numChars - i - 1 : i;
     const char = text.charAt(index);
     chunk = reverse ? char + chunk : chunk + char;
-    const charLength = measure(chunk) - chunkLength;
+    const charLength = scale * measureAndCacheTextWidth(font, chunk, cache) - chunkLength;
     chunkLength += charLength;
     const charM = startM + charLength / 2;
     while (offset < end - stride && segmentM + segmentLength < charM) {
diff --git a/src/ol/render/canvas.js b/src/ol/render/canvas.js
index 43bdc72fc..40089cb88 100644
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -324,6 +324,22 @@ export function measureTextWidth(font, text) {
 }
 
 
+/**
+ * Measure text width using a cache.
+ * @param {string} text The text to measure.
+ * @param {string} font The font.
+ * @param {Object<string, number>} cache A lookup of cached widths by text.
+ * @returns {number} The text width.
+ */
+export function measureAndCacheTextWidth(text, font, cache) {
+  if (text in cache) {
+    return cache[text];
+  }
+  const width = cache[text] = measureTextWidth(font, text);
+  return width;
+}
+
+
 /**
  * @param {string} font Font to use for measuring.
  * @param {Array<string>} lines Lines to measure.
diff --git a/src/ol/render/canvas/Executor.js b/src/ol/render/canvas/Executor.js
index cb018d1e4..3f73f6167 100644
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -22,7 +22,7 @@ import {
 
 
 import {createCanvasContext2D} from '../../dom.js';
-import {labelCache, defaultTextAlign, measureTextHeight, measureTextWidth, measureTextWidths} from '../canvas.js';
+import {labelCache, defaultTextAlign, measureTextHeight, measureTextWidth, measureAndCacheTextWidth, measureTextWidths} from '../canvas.js';
 
 
 /**
@@ -736,29 +736,22 @@ class CanvasExecutor {
 
           const textState = this.textStates[textKey];
           const font = textState.font;
-          const textScale = textState.scale;
+          const textScale = textState.scale * measurePixelRatio;
 
-          let widths = this.widths_[font];
-          if (!widths) {
-            this.widths_[font] = widths = {};
+          let cachedWidths;
+          if (font in this.widths_) {
+            cachedWidths = this.widths_[font];
+          } else {
+            cachedWidths = this.widths_[font] = {};
           }
 
-          //FIXME Do not create this function on every call
-          const measure = function(text) {
-            let width = widths[text];
-            if (!width) {
-              width = widths[text] = measureTextWidth(font, text);
-            }
-            return width * textScale * measurePixelRatio;
-          };
-
           const pathLength = lineStringLength(pixelCoordinates, begin, end, 2);
-          const textLength = measure(text);
+          const textLength = textScale * measureAndCacheTextWidth(font, text, cachedWidths);
           if (overflow || textLength <= pathLength) {
             const textAlign = this.textStates[textKey].textAlign;
             const startM = (pathLength - textLength) * TEXT_ALIGN[textAlign];
             const parts = drawTextOnPath(
-              pixelCoordinates, begin, end, 2, text, measure, startM, maxAngle);
+              pixelCoordinates, begin, end, 2, text, startM, maxAngle, measureAndCacheTextWidth, font, textScale, cachedWidths);
             if (parts) {
               let c, cc, chars, label, part;
               if (strokeKey) {
```

ahocevar:
This is still work in progress - the rendered result is not correct somehow.

tschaub:
@ahocevar curious what you are checking to test the result.  #8988 is what I had in mind.

image

A sample result from the mapbox-style example.

ahocevar:
Thanks for the suggestions @tschaub. I replaced my commit mit your suggestion, but the rendered commit results in incorrect line labels along lines (easy to see when zooming in on the `mapbox-style` example).

Feel free to try to fix and merge, or close. I have to exchange laptop and screen with bed and pillow now ;-). We still have #8960 if you think that is better.

tschaub:
Sounds like a good time for a break!  If you get the chance tomorrow to drop a screenshot of bad example, that would be great.

src/ol/render/canvas.js
```diff
@@ -324,6 +324,22 @@ export function measureTextWidth(font, text) {
...
 
 
+/**
+ * Measure text width using a cache.
+ * @param {string} text The text to measure.
+ * @param {string} font The font.
+ * @param {Object<string, number>} cache A lookup of cached widths by text.
+ * @returns {number} The text width.
+ */
+export function measureAndCacheTextWidth(text, font, cache) {
```
tschaub: While adding and editing comments on your PR, I decided to juggle the order of these args to better match the `measureTextWidth` signature.  So the reason things aren't working here is that this function is called with `font, text, cache`.

As mentioned above, I pushed a version of the commit I had in mind to #8988.  Tests are passing there.

ahocevar:
Closing in favor of #8988.

ahocevar:
This is what I got with this pull request:

wrong

PR #8982, opened 2018-11-19
Stacking context for layers, overlays, and controls

tschaub:
This makes it so all immediate viewport children have absolute positioning and a z-index of 0.  This gives layers, overlays, and controls predictable stacking contexts.  Layers will always appear below overlays.  Controls will always appear over overlays.  Within each group, z-index determines rendering order.

Since the parent element for the controls now has (absolute) positioning, we no longer register a listener on it to stop all events.  Instead, we only propagate events to other listeners if the target is not a child of the "stop event" container.  This should also improve interoperability with other frameworks that do their own event handling (since we no longer stop propagation of all control events).

The new rendering test shows that z-index for layers and controls is independent.

image

The red layer is above the green layer since it has a higher z-index (200 vs -200).  The blue control is above the orange control since it has a higher z-index (1 vs -1).  All controls appear above all layers since they share a stacking context which is after the layer stacking context in the document.

~~The expected output for all rendering tests has changed since we now render controls above layers (whether or not the default `ol.css` is included).  The examples now include the default CSS, so the default controls are rendered in the tests.~~


tschaub:
Rendering test failures look like they are due to font rendering for the controls:

https://409-4723248-gh.circle-artifacts.com/0/rendering/layer-vectortile-rotate/expected.png
https://409-4723248-gh.circle-artifacts.com/0/rendering/layer-vectortile-rotate/actual.png

I don't think we want to bump up the tolerance for all tests to cover this difference.  I'll poke around for alternative solutions.

tschaub:
Instead of trying to get the same fonts everywhere (CI + all dev machines), I decided to hide the controls by default in the rendering tests.  The stacking test still demonstrates that controls are properly rendered over layers.

tschaub:
It looks like the tolerance for the `text-style` case needs to be higher than others (font rendering again).  So I bumped that up to 2% so [expected](https://410-4723248-gh.circle-artifacts.com/0/rendering/text-style/expected.png) is in range of [actual](https://410-4723248-gh.circle-artifacts.com/0/rendering/text-style/actual.png).

tschaub:
The only significant change in the expected results of rendering tests is for the `layer-vectortile-simple` case (choose the rich diff display option to compare):
layer-vectortile-simple-diff

The [expected image](https://github.com/openlayers/openlayers/blob/5d14666376d88c532ae69095ebc16c4ae0e48402/rendering/cases/layer-vectortile-simple/expected.png) in master is missing points.  You can see the points in the [expected image](https://github.com/openlayers/openlayers/blob/5d14666376d88c532ae69095ebc16c4ae0e48402/rendering/cases/layer-vectortile-rotate/expected.png) for the rotated case.  The tolerance for this case was 2%.  I dropped it back down to 1% in hopes that we wouldn't miss this type of issue again (though I think it would be more obvious if we had tests that applied a non-default style).

PR #8992, opened 2018-11-20
Use standard tile coords

tschaub:
I think it is safe to say that the majority of tile coordinate systems in widespread use have a column and row order that follows standard image coordinate systems – rows increase downward and columns increase to the right, with the origin at the top left.  This branch proposes that our tile coordinates do the same.

From the upgrade notes:

### New internal tile coordinates

Previously, the internal tile coordinates used in the library had an unusual row order – the origin of the tile coordinate system was at the top left as expected, but the rows increased upwards.  This meant that all tile coordinates within a tile grid's extent had negative `y` values.

Now, the internal tile coordinates used in the library have the same row order as standard (e.g. XYZ) tile coordinates.  The origin is at the top left (as before), and rows or `y` values increase downward.  So the top left tile of a tile grid is now `0, 0`, whereas it was `0, -1` before.

```
x, y values for tile coordinates

origin 
  *__________________________
  |        |        |        |
  |  0, 0  |  1, 0  |  2, 0  |
  |________|________|________|
  |        |        |        |
  |  0, 1  |  1, 1  |  2, 1  |
  |________|________|________|
  |        |        |        |
  |  0, 2  |  1, 2  |  2, 2  |
  |________|________|________|
```

This change should only affect you if you were using a custom `tileLoadFunction` or `tileUrlFunction`.  For example, if you used to have a `tileUrlFunction` that looked like this:

```js
// before
function tileUrlFunction(tileCoord) {
  const z = tileCoord[0];
  const x = tileCoord[1];
  const y = -tileCoord[2] - 1;
  // do something with z, x, y
}
```

You would now do something like this:
```js
// after
function tileUrlFunction(tileCoord) {
  const z = tileCoord[0];
  const x = tileCoord[1];
  const y = tileCoord[2];
  // do something with z, x, y
}
```

In addition (this should be exceedingly rare), if you previously created a `ol/tilegrid/WMTS` by hand and you were providing an array of `sizes`, you no longer have to provide a negative height if your tile origin is the top-left corner (the common case).  On the other hand, if you are providing a custom array of `sizes` and your origin is the bottom of the grid (this is uncommon), your height values must now be negative.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -4,6 +4,53 @@
 
 #### Backwards incompatible changes
 
+##### New internal tile coordinates
+
+Previously, the internal tile coordinates used in the library had an unusual row order – the origin of the tile coordinate system was at the top left as expected, but the rows increased upwards.  This meant that all tile coordinates within a tile grid's extent had negative `y` values.
+
+Now, the internal tile coordinates used in the library have the same row order as standard (e.g. XYZ) tile coordinates.  The origin is at the top left (as before), and rows or `y` values increase downward.  So the top left tile of a tile grid is now `0, 0`, whereas it was `0, -1` before.
+
+```
+x, y values for tile coordinates
+
+origin
+  *__________________________
+  |        |        |        |
+  |  0, 0  |  1, 0  |  2, 0  |
+  |________|________|________|
+  |        |        |        |
+  |  0, 1  |  1, 1  |  2, 1  |
+  |________|________|________|
+  |        |        |        |
+  |  0, 2  |  1, 2  |  2, 2  |
+  |________|________|________|
+```
+
+This change should only affect you if you were using a custom `tileLoadFunction` or `tileUrlFunction`.  For example, if you used to have a `tileUrlFunction` that looked like this:
+
+```js
+// before
+function tileUrlFunction(tileCoord) {
+  const z = tileCoord[0];
+  const x = tileCoord[1];
+  const y = -tileCoord[2] - 1;
+  // do something with z, x, y
+}
+```
+
+You would now do something like this:
+```js
+// after
+function tileUrlFunction(tileCoord) {
+  const z = tileCoord[0];
+  const x = tileCoord[1];
+  const y = tileCoord[2];
+  // do something with z, x, y
+}
+```
+
+In addition (this should be exceedingly rare), if you previously created a `ol/tilegrid/WMTS` by hand and you were providing an array of `sizes`, you no longer have to provide a negative height if your tile origin is the top-left corner (the common case).  On the other hand, if you are providing a custom array of `sizes` and your origin is the bottom of the grid (this is uncommon), your height values must now be negative.
+
 ##### Removal of the "vector" render mode for vector tile layers
 
 If you were previously using `VectorTile` layers with `renderMode: 'vector'`, you have to remove this configuration option. That mode was removed. `'hybrid'` (default) and `'image'` are still available.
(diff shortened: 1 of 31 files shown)


tschaub:
A few key examples:

 * [Debug Tiles](https://457-4723248-gh.circle-artifacts.com/0/examples/canvas-tiles.html)
 * [WMTS Layer from Capabilities](https://457-4723248-gh.circle-artifacts.com/0/examples/wmts-layer-from-capabilities.html)
 * [Vector Tiles w/ Custom Grid](https://457-4723248-gh.circle-artifacts.com/0/examples/mapbox-vector-tiles-advanced.html)
 * [ArcGIS Rest w/ 512 Tiles](https://457-4723248-gh.circle-artifacts.com/0/examples/xyz-esri-4326-512.html)

philip-firstorder:
I think there is a bug in the calculation of tileGrid.fullTileRanges_

Check this sandbox:
https://codesandbox.io/s/mapbox-vector-tiles-advanced-2niff

Result (OpenLayers 6):
[{"minX":0,"maxX":0,"minY":-1,"maxY":-1},{"minX":0,"maxX":0,"minY":-2,"maxY":-1},{"minX":0,"maxX":1,"minY":-4,"maxY":-1},{"minX":0,"maxX":3,"minY":-8,"maxY":-1},{"minX":0,"maxX":7,"minY":-16,"maxY":-1},{"minX":0,"maxX":15,"minY":-32,"maxY":-1}]

Expected (OpenLayers 5):
"[{"minX":0,"maxX":0,"minY":0,"maxY":0},{"minX":0,"maxX":0,"minY":0,"maxY":1},{"minX":0,"maxX":1,"minY":0,"maxY":3},{"minX":0,"maxX":3,"minY":0,"maxY":7},{"minX":0,"maxX":7,"minY":0,"maxY":15},{"minX":0,"maxX":15,"minY":0,"maxY":31}]"

As you can see in the result above the minY and maxY values are all negative after your merge in OpenLayers 6.

In OpenLayers 5 when creating a new TileGrid the result was the one below, which I also expected now.

Can you please check the sandbox sample and confirm if there is a bug?

tschaub:
It definitely looks like a problem.  The tests for `tileGrid.getTileRangeForExtentAndZ()` look wrong:
https://github.com/openlayers/openlayers/blob/c851f6c6bfed9f62fada321149c68e287113d915/test/spec/ol/tilegrid/tilegrid.test.js#L833-L867

philip-firstorder:
I wonder why nobody reported it for over a year. I got this error immediately after updating from 5 to 6 and my map stopped working. I also use the getTileRangeForExentAndZ function in my production code exactly where the negative values appear. But then I traced it even further to the fullTileRanges_ values.

Will you provide a fix?

tschaub:
@philip-firstorder - after taking a closer look, the tests above are actually doing the right thing.  My confusion with those tests was based on the origin being set to `[0, 0]`.  See https://github.com/openlayers/openlayers/blob/c851f6c6bfed9f62fada321149c68e287113d915/test/spec/ol/tilegrid/tilegrid.test.js#L16-L20

A tile grid's origin is the top left corner, and the `origin` property is in map coordinates.  The y-value for tile coordinates increase downward in a tile grid.  For the tests above, tile coordinates are being requested for map extents with positive y-values, and those are above the origin - so negative in terms of the y-value for tile coordinates.

Your example can be fixed in two ways.  If you need to set the origin of a tile grid, the y-value of the origin should correspond to the maximum y-value of the map extent - so `origin: [extent[0], extent[3]]` instead of `origin: [extent[0], extent[1]]` in your example.  Alternatively, you can remove the `origin` option from the tile grid constructor altogether.  This lets the constructor pick the origin from the extent, and it will use the maximum y-value of the extent you provide.

Here is an updated version of your example that shows the correct tile range for your custom tile grid: https://codesandbox.io/s/custom-tile-grid-hihbt

As an aside, we should update the tile grid tests to use an origin that makes more sense - then we would not have negative y-values for tile coordinates.

philip-firstorder:
This fixed it, thank you for your explanation.

PR #8996, opened 2018-11-22
Add imageRatio option for VectorImage layers

fredj:
replaces #8897 
fixes #8637
fixes #8897

### DIFF
--- a/src/ol/layer/VectorImage.js
+++ b/src/ol/layer/VectorImage.js
@@ -2,10 +2,13 @@
  * @module ol/layer/VectorImage
  */
 import BaseVectorLayer from './BaseVector.js';
+import {assign} from '../obj.js';
 import CanvasVectorImageLayerRenderer from '../renderer/canvas/VectorImageLayer.js';
 
 /**
  * @typedef {import("./BaseVector.js").Options} Options
+ * @property {number} [imageRatio=1] Ratio by which the rendered extent should be larger than the
+ * viewport extent A larger ratio avoids cut images during panning, but will cause a decrease in performance.
  */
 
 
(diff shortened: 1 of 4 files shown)


src/ol/layer/VectorImage.js
```diff
@@ -23,7 +26,25 @@ class VectorImageLayer extends BaseVectorLayer {
    * @param {Options=} opt_options Options.
    */
   constructor(opt_options) {
+    const options = opt_options ? opt_options : /** @type {Options} */ ({});
+
+    const baseOptions = assign({}, options);
+    delete baseOptions.imageRatio;
     super(opt_options);
```
ahocevar: I think you meant

```suggestion
     super(baseOptions);
```

src/ol/layer/VectorImage.js
```diff
@@ -2,10 +2,13 @@
  * @module ol/layer/VectorImage
  */
 import BaseVectorLayer from './BaseVector.js';
+import {assign} from '../obj.js';
 import CanvasVectorImageLayerRenderer from '../renderer/canvas/VectorImageLayer.js';
 
 /**
  * @typedef {import("./BaseVector.js").Options} Options
+ * @property {number} [imageRatio=1] Ratio by which the rendered extent should be larger than the
+ * viewport extent A larger ratio avoids cut images during panning, but will cause a decrease in performance.
```
tschaub:
```suggestion
 * viewport extent. A larger ratio avoids cut images during panning, but will cause a decrease in performance.
```

src/ol/layer/VectorImage.js
```diff
@@ -2,10 +2,13 @@
  * @module ol/layer/VectorImage
  */
 import BaseVectorLayer from './BaseVector.js';
+import {assign} from '../obj.js';
 import CanvasVectorImageLayerRenderer from '../renderer/canvas/VectorImageLayer.js';
 
 /**
  * @typedef {import("./BaseVector.js").Options} Options
+ * @property {number} [imageRatio=1] Ratio by which the rendered extent should be larger than the
+ * viewport extent. A larger ratio avoids cut images during panning, but will cause a decrease in performance.
```
tschaub: It looks like adding a property to a type with this syntax is not supported by TypeScript.  See #9263.

PR #9002, opened 2018-11-25
Fix module path

### LINKED ISSUE #9000: Typo in Composite.js
In this file:
https://github.com/openlayers/openlayers/blob/master/src/ol/renderer/Composite.js

The 2nd line, it should be:
@module ol/renderer/Composite

ahocevar:
Fixes #9000.


ahocevar:
Trivial change, merging.

PR #9005, opened 2018-11-26
Simplify import path in examples

fredj:
To have the same path (starting with `ol/`, without `.js`) as in the documentation.
The support was added in the webpack config in #8928

The generated examples: https://477-4723248-gh.circle-artifacts.com/0/examples/index.html

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -18,6 +18,7 @@
     "turf": false
   },
   "rules": {
+    "import/no-unresolved": 0,
     "no-unused-vars": [2, {"varsIgnorePattern": "^map"}]
   }
 }
--- a/examples/accessible.js
+++ b/examples/accessible.js
@@ -1,7 +1,7 @@
-import Map from '../src/ol/Map.js';
-import View from '../src/ol/View.js';
-import TileLayer from '../src/ol/layer/Tile.js';
-import OSM from '../src/ol/source/OSM.js';
+import Map from 'ol/Map';
+import View from 'ol/View';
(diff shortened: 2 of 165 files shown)


ahocevar:
I think you'll have to change `examples/.eslintrc` to disable the `import/no-unresolved` rule. And you can remove the code in `examples/example-builder.js` that changes the import paths in the `jsSource`.

fredj:
Instead of disabling `import/no-unresolved`, I'm now using https://www.npmjs.com/package/eslint-import-resolver-webpack (but having other error).

And thanks for the `examples/example-builder.js` code removal, I will do that

fredj:

tschaub:
I’ve seen clear benefits of the eslint import rules.  I also appreciate being able to follow relative import paths in my editor.  

It isn’t clear to me that there are benefits of the shorter/absolute paths in our sources (having them in the published examples makes sense because they are intended for use with published packages).

I’m curious if there is a benefit of this change that justifies removing the import checking and ability to navigate relative paths during development.

tschaub:
For example, this was the development experience before (pretty good):

before

And this is the development experience after (not as good):

after

fredj:
Thinking of this again, I agree that it's a now inconvenient to edit the examples. I will fix that

ahocevar:
Let's not forget the goal to make the examples easier to build (without example builder plugin) and use (e.g. directly with codesandbox.io). We started that effort a while ago in the [site](https://github.com/openlayers/openlayers/tree/site) branch.

fredj:
see #9009

PR #9008, opened 2018-11-27
Decouple render instruction creation from rendering

ahocevar:
This pull request changes vector tile rendering so instruction creation is no longer done in an animation frame.

Instruction creation is not supposed to perform any canvas operations any more, so it makes sense to do it as soon as all source tiles have finished loading. In the future, this could be done in a worker.

To avoid blocking interactions and animations during rendering, we now only render new tile images when no interaction or animation is going on, or when we have a lower resolution interim tile available. Even when rendering a high fidelity map, we spend no more than 100 ms in a frame, to avoid blocking interactions that the user might start when the map image from the previous render has settled.

The result of these changes is an improved user experience, where the map is much more responsive to user interactions.

### DIFF
--- a/src/ol/Disposable.js
+++ b/src/ol/Disposable.js
@@ -14,15 +14,15 @@ class Disposable {
      * @type {boolean}
      * @private
      */
-    this.disposed = false;
+    this.disposed_ = false;
   }
 
   /**
    * Clean up.
    */
   dispose() {
-    if (!this.disposed) {
-      this.disposed = true;
+    if (!this.disposed_) {
+      this.disposed_ = true;
       this.disposeInternal();
     }
   }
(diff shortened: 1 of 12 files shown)


src/ol/Disposable.js
```diff
@@ -14,15 +14,15 @@ class Disposable {
      * @type {boolean}
      * @private
```
fredj: `@private` could be removed

ahocevar:
This is now ready for review.

ahocevar:
https://21-4974939-gh.circle-artifacts.com/0/examples/mapbox-style.html is a good example to test the improvements in perceived performance.

ahocevar:
I have now cleaned up the commit history. Thanks for any review.

fredj:
ping @gberaudo

ahocevar:
Thanks for the review @fredj.

PR #9013, opened 2018-11-28
Added scalebar option to ol.control.scaleline

weskamm:
This adds an option to show scalebars instead of a scaleline in the `ScaleLine` control.

scalebar

Example and API doc have been adjusted.

### DIFF
--- a/examples/scale-line.js
+++ b/examples/scale-line.js
@@ -4,12 +4,27 @@ import {defaults as defaultControls, ScaleLine} from '../src/ol/control.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import OSM from '../src/ol/source/OSM.js';
 
+let scaleType = 'scaleline';
+let scaleBarSteps = 4;
+let scaleBarText = true;
+let control;
 
-const scaleLineControl = new ScaleLine();
-
+function scaleControl() {
+  if (scaleType === 'scaleline') {
+    control = new ScaleLine();
+    return control;
+  }
+  control = new ScaleLine({
+    scaleBar: true,
+    scaleBarSteps: scaleBarSteps,
+    scaleBarText: scaleBarText,
+    minWidth: 200
+  });
+  return control;
+}
 const map = new Map({
   controls: defaultControls().extend([
-    scaleLineControl
+    scaleControl()
   ]),
   layers: [
     new TileLayer({
@@ -23,10 +38,40 @@ const map = new Map({
   })
 });
 
-
 const unitsSelect = document.getElementById('units');
+const typeSelect = document.getElementById('type');
+const stepsSelect = document.getElementById('steps');
+const scaleTextCheckbox = document.getElementById('showScaleText');
+const showScaleTextDiv = document.getElementById('showScaleTextDiv');
 function onChange() {
-  scaleLineControl.setUnits(unitsSelect.value);
+  control.setUnits(unitsSelect.value);
+}
+function onChangeType() {
+  scaleType = typeSelect.value;
+  if (typeSelect.value === 'scalebar') {
+    stepsSelect.style.display = 'inline';
+    showScaleTextDiv.style.display = 'inline';
+    map.removeControl(control);
+    map.addControl(scaleControl());
+  } else {
+    stepsSelect.style.display = 'none';
+    showScaleTextDiv.style.display = 'none';
+    map.removeControl(control);
+    map.addControl(scaleControl());
+  }
+}
+function onChangeSteps() {
+  scaleBarSteps = parseInt(stepsSelect.value, 10);
+  map.removeControl(control);
+  map.addControl(scaleControl());
+}
+function onChangeScaleText() {
+  scaleBarText = scaleTextCheckbox.checked;
+  map.removeControl(control);
+  map.addControl(scaleControl());
 }
 unitsSelect.addEventListener('change', onChange);
+typeSelect.addEventListener('change', onChangeType);
+stepsSelect.addEventListener('change', onChangeSteps);
+scaleTextCheckbox.addEventListener('change', onChangeScaleText);
 onChange();
--- a/examples/scale-line.js
+++ b/examples/scale-line.js
@@ -4,6 +4,12 @@ import {defaults as defaultControls, ScaleLine} from '../src/ol/control.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import OSM from '../src/ol/source/OSM.js';
 
+const unitsSelect = document.getElementById('units');
+const typeSelect = document.getElementById('type');
+const stepsSelect = document.getElementById('steps');
+const scaleTextCheckbox = document.getElementById('showScaleText');
+const showScaleTextDiv = document.getElementById('showScaleTextDiv');
+
 let scaleType = 'scaleline';
 let scaleBarSteps = 4;
 let scaleBarText = true;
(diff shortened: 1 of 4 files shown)


src/ol/control/ScaleLine.js
```diff
@@ -45,6 +45,11 @@ const LEADING_DIGITS = [1, 2, 5];
  * @property {HTMLElement|string} [target] Specify a target if you want the control
  * to be rendered outside of the map's viewport.
  * @property {Units|string} [units='metric'] Units.
+ * @property {boolean} [scaleBar=false] Render scalebars instead of a line.
```
tschaub:
```suggestion
 * @property {boolean} [bar=false] Render scale bars instead of a line.
```

src/ol/control/ScaleLine.js
```diff
@@ -45,6 +45,11 @@ const LEADING_DIGITS = [1, 2, 5];
...
+ * @property {number} [scaleBarSteps=4] Number of steps the scalebar should use. Use even numbers
```
tschaub:
```suggestion
 * @property {number} [steps=4] Number of steps the scalebar should use. Use even numbers
```

src/ol/control/ScaleLine.js
```diff
@@ -45,6 +45,11 @@ const LEADING_DIGITS = [1, 2, 5];
...
+ * for best results. Only useful when `scaleBar` is `true`.
```
tschaub:
```suggestion
 * for best results. Only applies when `bar` is `true`.
```

src/ol/control/ScaleLine.js
```diff
@@ -45,6 +45,11 @@ const LEADING_DIGITS = [1, 2, 5];
...
+ * @property {boolean} [scaleBarText=false] Render a scale ontop of the scalebar. Only useful
```
tschaub:
```suggestion
 * @property {boolean} [text=false] Render the text scale above the scalebar. Only applies
```

src/ol/control/ScaleLine.js
```diff
@@ -45,6 +45,11 @@ const LEADING_DIGITS = [1, 2, 5];
...
+ * when `scaleBar` is `true`.
```
tschaub:
```suggestion
 * when `bar` is `true`.
```

src/ol/control/ScaleLine.js
```diff
@@ -57,6 +62,8 @@ const LEADING_DIGITS = [1, 2, 5];
  * viewport center cannot be calculated in the view projection.
  * By default the scale line will show in the bottom left portion of the map,
  * but this can be changed by using the css selector `.ol-scale-line`.
+ * When specifying `scaleBar` as `true`, a scalebar will be rendered instead
```
tschaub:
```suggestion
 * When specifying `bar: true`, a scalebar will be rendered instead
```

src/ol/control/ScaleLine.js
```diff
@@ -123,6 +131,24 @@ class ScaleLine extends Control {
 
     this.setUnits(/** @type {Units} */ (options.units) || Units.METRIC);
 
+    /**
+     * @private
+     * @type {boolean}
+     */
+    this.scaleBar_ = options.scaleBar || false;
```
tschaub:
```suggestion
    this.scaleBar_ = options.bar || false;
```

src/ol/control/ScaleLine.js
```diff
@@ -123,6 +131,24 @@ class ScaleLine extends Control {
...
+
+    /**
+     * @private
+     * @type {number}
+     */
+    this.scaleBarSteps_ = options.scaleBarSteps || 4;
```
tschaub:
```suggestion
    this.scaleBarSteps_ = options.steps || 4;
```

examples/scale-line.js
```diff
@@ -4,12 +4,27 @@ import {defaults as defaultControls, ScaleLine} from '../src/ol/control.js';
...
 
-const scaleLineControl = new ScaleLine();
-
+function scaleControl() {
+  if (scaleType === 'scaleline') {
+    control = new ScaleLine();
+    return control;
+  }
+  control = new ScaleLine({
+    scaleBar: true,
```
tschaub:
```suggestion
    bar: true,
```

examples/scale-line.js
```diff
@@ -4,12 +4,27 @@ import {defaults as defaultControls, ScaleLine} from '../src/ol/control.js';
...
+    scaleBarSteps: scaleBarSteps,
```
tschaub:
```suggestion
    steps: scaleBarSteps,
```

examples/scale-line.js
```diff
@@ -4,12 +4,27 @@ import {defaults as defaultControls, ScaleLine} from '../src/ol/control.js';
...
+    scaleBarText: scaleBarText,
```
tschaub:
```suggestion
    text: scaleBarText,
```

src/ol/control/ScaleLine.js
```diff
@@ -274,8 +306,125 @@ class ScaleLine extends Control {
...
+  /**
+   * @private
+   * @param {number} width The current width of the scalebar.
+   * @param {number} scale The current scale.
+   * @param {string} suffix The suffix to append to the scale text.
+   * @returns {string} The stringified HTML of the scalebar.
+   */
+  createScaleBar(width, scale, suffix) {
+    let mapScale = Math.round(this.getScaleForResolution());
+    mapScale = '1 : ' + mapScale.toLocaleString().replace(/,/g, '.');
```
tschaub: I don't think we want to always replace `,` with `.` here.

For example, for my current locale:
```js
> (1e3 + 1e-2).toLocaleString().replace(/,/g, '.')
1.000.01
```

Instead, these two lines could be:
```js
const mapScale = '1 : ' + Math.round(this.getScaleForResolution()).toLocaleString();
```

examples/scale-line.js
```diff
@@ -4,12 +4,27 @@ import {defaults as defaultControls, ScaleLine} from '../src/ol/control.js';
...
+let scaleBarText = true;
+let control;
 
-const scaleLineControl = new ScaleLine();
-
+function scaleControl() {
+  if (scaleType === 'scaleline') {
+    control = new ScaleLine();
+    return control;
+  }
```
tschaub: I think you'll also need to set the units here.  As you can see in the [hosted example](https://488-4723248-gh.circle-artifacts.com/0/examples/scale-line.html), when you change from line to bar (or back), the units aren't right if they have been changed.

tschaub:
Thanks for the contribution @weskamm!  My suggestions above are mostly about shorter option names.

I'm curious what you and others think about making the default style look a bit more like the current scaleline style.

ThomasG77:
I'm not sure to understand the purpose of this PR as ol-ext has already an extension http://viglino.github.io/ol-ext/examples/canvas/map.canvas.control.html
Should it be maintain in the core?

tschaub:
> However, we could also think this the other way around and make the default theme for other controls look more black/white/grey.

I like that idea.  Probably best for a separate PR.

weskamm:
Thnanks for the reviews!
I adressed the issues mentioned by @tschaub (renamed props, fixed example, localization).

Regarding the length of the scalebar: I use a non-default minWidth in the example, because the default width of 64 pixels is a bit small for a scalebar (but ok for scaleline). Thats why the sizes differ.

Regarding the style: I would like to keep the current black and white look, as this is the way a scalebar typically looks like.

src/ol/control/ScaleLine.js
```diff
@@ -69,7 +76,8 @@ class ScaleLine extends Control {
 
     const options = opt_options ? opt_options : {};
 
-    const className = options.className !== undefined ? options.className : 'ol-scale-line';
+    const className = options.className !== undefined ? options.className :
+      options.bar !== undefined ? 'ol-scale-bar' : 'ol-scale-line';
```
tschaub:
```suggestion
      options.bar ? 'ol-scale-bar' : 'ol-scale-line';
```

Without this change, `bar: false` would result in `ol-scale-bar` for a class name, which is not what we want.

weskamm: right, fixed

PR #9023, opened 2018-11-30
Fix webgl PointsLayer not rendering anything

### LINKED ISSUE #9014: Issues with ol/layer/Heatmap
**Describe the bug**
Looks like the current implementation of `ol/layer/Heatmap` does not work. I tested in many different browsers, and I always get this error
```
[.WebGL-0x7f7f4c040e00]GL ERROR :GL_INVALID_OPERATION : glDrawElements: attempt to access out of range vertices in attribute 0
```

**To Reproduce**
1. Go to https://485-4723248-gh.circle-artifacts.com/0/examples/heatmap-earthquakes.html
2. Open the dev tools
3. See errors in the dev tools and no heatmap on the map

**Expected behavior**
A heatmap on the map, and no errors in the dev tools.

jahow:
Attributes were used in the shader but not bound to a buffer, which made the rendering fail in some implementations.

This should fix #9014 (to be confirmed)

Example visible at https://498-4723248-gh.circle-artifacts.com/0/examples/heatmap-earthquakes.html

### DIFF
--- a/src/ol/layer/Heatmap.js
+++ b/src/ol/layer/Heatmap.js
@@ -205,6 +205,7 @@ class Heatmap extends VectorLayer {
         attribute vec2 a_texCoord;
         attribute float a_rotateWithView;
         attribute vec2 a_offsets;
+        attribute float a_opacity;
         
         uniform mat4 u_projectionMatrix;
         uniform mat4 u_offsetScaleMatrix;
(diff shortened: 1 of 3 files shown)


tschaub:
Works for me in Chrome 70 and Safari 12 on macOS 10.14.  Thanks for the continued work on this @jahow.

PR #9024, opened 2018-12-02
Add prerender and postrender events for Image layer

### LINKED ISSUE #9001: render() in reproj.js ignores original context options
**Describe the bug**
One can customize canvas context in "precompose" event.
For instance, one can disable antialiasing:

```js
layer.on("precompose", evt => { 
    evt.context.imageSmoothingEnabled = false
})
```

When using ImageStatic source with projection, redner() in reproj.js is called for mapped triangulation.
In render(), new context for canvas is created inside, and everything user set in "precompose" event just ignored.

**To Reproduce**
1. Use Image layer with ImageStatic source with non-EPSG:3857 projection
2. DIsable antialiasing in "precompose" event for the layer
3. The static image is still antialiased

**Expected behavior**
The antialiasing in static image should be disabled.

ahocevar:
This pull request brings back render events for `ol/layer/Image`.

Fixes #9001.

### DIFF
--- a/src/ol/renderer/canvas/VectorImageLayer.js
+++ b/src/ol/renderer/canvas/VectorImageLayer.js
@@ -106,6 +106,16 @@ class CanvasVectorImageLayerRenderer extends CanvasImageLayerRenderer {
   }
 
   /**
+   * @override
+   */
+  preRender() {}
+
+  /**
+   * @override
+   */
+  postRender() {}
+
+  /**
    * @inheritDoc
    */
   forEachFeatureAtCoordinate(coordinate, frameState, hitTolerance, callback) {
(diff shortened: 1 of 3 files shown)


ahocevar:
Review question: do we want to do this, or should we make it so the image layer renderer just returns an image with a css transform in `renderFrame`, instead of drawing to a canvas?

src/ol/renderer/canvas/VectorImageLayer.js
```diff
@@ -105,6 +105,16 @@ class CanvasVectorImageLayerRenderer extends CanvasImageLayerRenderer {
...
 
+  /**
+   * @override
+   */
+  preRender() {}
+
+  /**
+   * @override
+   */
+  postRender() {}
```
tschaub: Not reading on a full sized computer, but won’t the vector image layer need to implement these to get events?

ahocevar: The vector image layer renderer relies on the vector layer renderer, which dispatches these events already. Without this noop override, the events would be triggered twice.

Issue #9006, opened 2018-11-26
ol.interaction.modify.insertVertexCondition should only be called if a vertex actually can be added


ludvigeriksson:
Currently, returning `true` from `ol.interaction.Modify.insertVertexCondition` does not mean a vertex will be created. It can be seen [here](https://github.com/openlayers/openlayers/blob/v5.3.0/src/ol/interaction/Modify.js#L826) that there are multiple other conditions that need to be true for this to happen. I think it would be more clear to put the call to `insertVertexCondition` last in the if-statement, since you then can rely on that returning `true` from this function will result in a vertex being created.

The [documentation ](https://github.com/openlayers/openlayers/blob/v5.3.0/src/ol/interaction/Modify.js#L85)for the function now states "whether a new vertex **can** be added to the sketch features". After this change, it could state "whether a new vertex **should** be added to the sketch features".

PR #9031, opened 2018-12-04
Call insertVertexCondition only when a vertex acually can be inserted

ludvigeriksson:
An implementation of #9006.


iakovagababian2020:
How to get the currently selected feature from the insertVertexCondition event?

PR #9058, opened 2018-12-12
Move all tile loading from the VectorImageTile to the source

ahocevar:
This pull request refactors tile loading for vector tile sources. The goal is to let the source manage all loading, instead of having load logic in the vector image tile.

This refactoring also results in a simplified tile load cycle, mostly by giving `ol/Tile` a new `hifi` property to indicate when a tile has data for its nominal (highest) resolution. 

This pull request is a first step towards a render tile that can get its data (i.e. render instructions) from a worker.

### DIFF
--- a/src/ol/source/VectorTile.js
+++ b/src/ol/source/VectorTile.js
@@ -3,15 +3,18 @@
  */
 
 import TileState from '../TileState.js';
-import VectorImageTile, {defaultLoadFunction} from '../VectorImageTile.js';
+import VectorImageTile from '../VectorImageTile.js';
 import Tile from '../VectorTile.js';
 import {toSize} from '../size.js';
 import UrlTile from './UrlTile.js';
-import {getKeyZXY} from '../tilecoord.js';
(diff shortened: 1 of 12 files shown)


ahocevar:
This is now ready for review.

src/ol/source/VectorTile.js
```diff
@@ -149,6 +166,140 @@ class VectorTile extends UrlTile {
...
+        covered = true;
+        empty = true;
+        sourceTileGrid.forEachTileCoord(extent, loadedZ, function(sourceTileCoord) {
+          const tileKey = getKey(sourceTileCoord);
+          let sourceTile;
+          if (tileKey in this.sourceTiles_) {
+            sourceTile = this.sourceTiles_[tileKey];
+            const state = sourceTile.getState();
+            if (state === TileState.LOADED || state === TileState.ERROR || state === TileState.EMPTY) {
+              empty = empty && sourceTile.getState() === TileState.EMPTY;
```
fredj: minor, `state` can be used instead of `sourceTile.getState()`

Issue #6514, opened 2017-02-19
OL4 - apparently ol.proj.transform doesn't "parseFloat()" coordinates


cjeanneret:
Hello,

I had a small headache with a proj transformation for a map center. While using `view.setCenter()` it kept pointing the [0,0] coordinates.

After some tests and struggles, it appears the float values weren't floats (coming from a JSON loaded via jquery) but string, and that I had to `parseFloat()` them while loading in the ol.proj.transform.
As I could `grep` parseFloat() in the ol code, I suspect its call isn't done everywhere.

Might be good to have a check on that.

Cheers,

C.

tschaub:
Please use parseFloat in your application.  The argument to view.setCenter() must be an array of numbers.

Issue #6111, opened 2016-11-11
ol.proj.toLonLat breaks when passing strings(sometimes?)


fredlllll:
okay so i was working on an app, and had to switch my webservice, that supplies coordinates from my dev environment to a test environment. test environment uses a slightly older version of mysql/mariadb and returns all results as strings, even though they are numbers, while dev does not. im then using these values in php, put them in an object, and return the json representation of that. so for dev, i get [number,number] in my app, and for test i get [string,string]. testing this in the console did not give me an error, but strangely in code it does mess up the latitude (after reconversion 48... to -41 and 1 to -108) when converting from lonlat.

as testcase i tried this in the console, but it worked as expected (these were the exact values that got screwed up)

ol.proj.fromLonLat(ol.proj.toLonLat([11.56202446466,48.141754842565]))
ol.proj.fromLonLat(ol.proj.toLonLat(["11.56202446466","48.141754842565"]))

i know coordinate is supposed to be a number[], but thanks to dynamic typing, javascript doesnt care for anything. does anyone else think it might be a good idea to add typechecks to some methods and throw errors if the wrong type is put in?

tschaub:
Thanks for the suggestion, but I think that at this time we are not going to put runtime type checks in places like this.

TDesjardins:
Hello Tim,

do you have a special reason for not doing this? In my opinion you could save a lot of time (yours and from users of ol3) when checking types. Have a look at these issues:

https://github.com/openlayers/ol3/issues/6086
https://github.com/openlayers/ol3/issues/6016

The cause is always the same and I am sure that these issues are not the last ones.

This is only a recommendation. For me this is not a problem because I use typesafe https://github.com/TDesjardins/gwt-ol3

Best regards

Tino

Issue #6016, opened 2016-10-24
view.setCenter() with coordinates as strings causes inconsistent behaviour


derekwatling:
When calling view.setCenter() with the coordinate array containing "numbers" as strings the map will change position to the correct center, but then some functinallity works while other functionality returns the view center as [NaN, NaN].

Specific issues I had were dragging the map and olView.calculateExtent(olMap.getSize()) returning [NaN,NaN,NaN,NaN]

ahocevar:
Please use the types listed in the API docs to avoid such issues. The argument of `ol.View#setCenter()` is `{Array.<number>}`.

TDesjardins:
Why not checking types with assertions to prevent bad arguments?

ahocevar:
We decided not to do so.

derekwatling:
I am using parseFloat() now that I know what was causing my problem. What was confusing is that it looked like it is working (moved the view center to the specified position) but then I couldn't drag the map.

TDesjardins:
What are the reasons for this decision? Is there new knowledge of good software design? I think you and the users of ol3 could save a lot of time when validating input arguments (type checking included). I suppose there will be a lot of more issues of this kind in future.

Best wishes

avbentem:
In the latest 4.0.1 (and in 3.20.1), using an array of string values (such as `["123", "456"]`, in my case badly parsed from a URL) causes a complete freeze of Chrome, Firefox, Safari and Internet Explorer.

Live example for 4.0.1 on https://codepen.io/anon/pen/jBOwjL

```
var view = new ol.View({
  center: [546000, 6868000],
  zoom: 13
});

var map = new ol.Map({
  target: 'map',
  layers: [
    new ol.layer.Tile({
      source: new ol.source.OSM()
    })
  ],
  view: view
});

// This will freeze the browser
view.setCenter(["546000", "6868000"]);
```

ahocevar:
Please supply numbers, as this is the expected and documented type.

blindfish3:
I just encountered this issue (by inadvertently passing strings from a database request to `View.setCenter()`) and have to say I'm very surprised at the decision not to do any input validation; particularly given the repercussions in this particular case.  If the result was silent failure I could almost accept it; but the symptoms I experienced look like a massive memory leak, crashing the browser and bringing my system to its knees.  That's very uncool.

I'd encourage you to either do type-checking; or, failing that, at least prevent memory leaks when the wrong type is passed...

[...]

Issue #4989, opened 2016-03-08
ol.proj.fromLonLat needs that latitude is a number (string not working)


rollingnet:
I have just discovered that ol.proj.fromLonLat function requires that the 2n component of array (i.e. latitude) is a number. If it is a string, it returns strange results.
The 1st argument (longitude) is correct even if it is a string.

ahocevar:
Yes, and it's documented - the first argument of [`ol.proj.fromLonLat`](http://openlayers.org/en/v3.14.2/apidoc/ol.proj.html#.fromLonLat) is an [`ol.Coordinate`](http://openlayers.org/en/v3.14.2/apidoc/ol.html#.Coordinate), which is an `Array.<number>`.

PR #9060, opened 2018-12-12
Avoid unexpected behavior when passing string coordinates

ahocevar:
As reported e.g. in #4989, #6111, #6514, #6016, #2500, accidently passing strings as coordinates causes issues that are very hard to track down. Assuming that not everyone uses a type checker, and given the fact that a fix is straightforward, I went ahead and changed the code to do explicit coercion to number rather than implicit coercion to string in two obvious cases.

### DIFF
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -40,8 +40,8 @@ import {padNumber} from './string.js';
  * @api
  */
 export function add(coordinate, delta) {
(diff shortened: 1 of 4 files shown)


marcjansen:
This will make many people happy, I guess.

ahocevar:
Force-pushed to trigger a CircleCI build.

PR #9070, opened 2018-12-17
Modify.removePoint returns true only when a vertex was removed

ludvigeriksson:
The [documentation](https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/Modify.js#L1019) states that `Modify.removePoint` returns true when a vertex was removed. This is currently not the case. This PR makes it return true only when a vertex was removed.

However, `MODIFYSTART` ([here](https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/Modify.js#L357)) and `MODIFYEND` ([here](https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/Modify.js#L1027)) events will still be dispatched, even though a vertex never was removed. Maybe a function `canRemoveVertex_` should exist and be called before `willModifyFeatures_`? This way the function will both return the correct value and not dispatch unnecessary modify events. I can implement this function if you think it sounds like a good idea.

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -1016,10 +1016,10 @@ class Modify extends PointerInteraction {
     if (this.lastPointerEvent_ && this.lastPointerEvent_.type != MapBrowserEventType.POINTERDRAG) {
       const evt = this.lastPointerEvent_;
       this.willModifyFeatures_(evt);
-      this.removeVertex_();
+      const removed = this.removeVertex_();
       this.dispatchEvent(new ModifyEvent(ModifyEventType.MODIFYEND, this.features_, evt));
       this.modified_ = false;
-      return true;
+      return removed;
     }
     return false;
   }


src/ol/interaction/Modify.js
```diff
@@ -1016,10 +1016,10 @@ class Modify extends PointerInteraction {
     if (this.lastPointerEvent_ && this.lastPointerEvent_.type != MapBrowserEventType.POINTERDRAG) {
       const evt = this.lastPointerEvent_;
       this.willModifyFeatures_(evt);
-      this.removeVertex_();
+      const removed = this.removeVertex_();
       this.dispatchEvent(new ModifyEvent(ModifyEventType.MODIFYEND, this.features_, evt));
```
ahocevar: You could displatch this event only when the vertex was removed:
```js
if (removed) {
  this.dispatchEvent(new ModifyEvent(ModifyEventType.MODIFYEND, this.features_, evt));
}
```

ludvigeriksson: Another thing I just thought of - it might be beneficial to have a public API to check whether a vertex can be removed. This could, for example, be used to hide a "remove vertex"-button in the UI if removal currently isn't possible.

ludvigeriksson: It could look something like [this](https://github.com/ludvigeriksson/openlayers/commit/c7154b1810cc99c194a2a9495566685f6dd9a9c1). It might need more thorough testing before merging, but it shows my thinking.

ahocevar: I think for the sake of library code simplicity, we'll leave things as in this pull request and let the user deal with modifystart/end events that do not result in any modification. Ok for you @ludvigeriksson?

ludvigeriksson: Sure thing

PR #9072, opened 2018-12-18
Remove 'layerStates' property from the FrameState

fredj:
We are currently maintaining two structures for the layers state in the frame state: `layerStates` and `layerStatesArray`. The pull request removes `layerStates`.

`layerStates` was only used in `ol/renderer/Map`.

~~An other change is that the layers removal function is now always scheduled.~~ see 1047fed

### DIFF
--- a/src/ol/renderer/Map.js
+++ b/src/ol/renderer/Map.js
@@ -280,7 +280,9 @@ class MapRenderer extends Disposable {
    * @protected
    */
   scheduleExpireIconCache(frameState) {
-    frameState.postRenderFunctions.push(/** @type {import("../PluggableMap.js").PostRenderFunction} */ (expireIconCache));
+    if (iconImageCache.canExpireCache()) {
+      frameState.postRenderFunctions.push(expireIconCache);
+    }
   }
 
   /**
--- a/src/ol/renderer/Map.js
+++ b/src/ol/renderer/Map.js
@@ -106,19 +106,14 @@ class MapRenderer extends Disposable {
     let result;
     const viewState = frameState.viewState;
     const viewResolution = viewState.resolution;
-    const managedLayers = frameState.layerStatesArray.map(function(state) {
-      if (state.managed) {
-        return getUid(state.layer);
-      }
-    });
 
     /**
+     * @param {boolean} managed Managed layer.
      * @param {import("../Feature.js").FeatureLike} feature Feature.
      * @param {import("../layer/Layer.js").default} layer Layer.
      * @return {?} Callback result.
      */
-    function forEachFeatureAtCoordinate(feature, layer) {
-      const managed = includes(managedLayers, getUid(layer));
+    function forEachFeatureAtCoordinate(managed, feature, layer) {
       if (!(getUid(feature) in frameState.skippedFeatureUids && !managed)) {
         return callback.call(thisArg, feature, managed ? layer : null);
       }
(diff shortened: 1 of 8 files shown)


src/ol/renderer/Map.js
```diff
@@ -105,14 +106,19 @@ class MapRenderer extends Disposable {
...
+    });
 
     /**
      * @param {import("../Feature.js").FeatureLike} feature Feature.
      * @param {import("../layer/Layer.js").default} layer Layer.
      * @return {?} Callback result.
      */
     function forEachFeatureAtCoordinate(feature, layer) {
-      const managed = frameState.layerStates[getUid(layer)].managed;
+      const managed = includes(managedLayers, getUid(layer));
```
tschaub: We could retain the previous performance by making a managed layers lookup object instead of an array.  (I haven’t confirmed this would be a performance improvement, but I’m guessing the property check would be cheaper than the includes call.)

fredj: See 5d1c27d
Instead of creating a structure for the managed layers, the value is now directly bounded to the callback function.

src/ol/renderer/Map.js
```diff
@@ -255,9 +257,12 @@ class MapRenderer extends Disposable {
    * @private
    */
   removeUnusedLayerRenderers_(map, frameState) {
-    for (const layerKey in this.layerRenderers_) {
-      if (!frameState || !(layerKey in frameState.layerStates)) {
-        this.removeLayerRendererByKey_(layerKey).dispose();
+    if (frameState) {
+      const layersUids = getLayersUids(frameState.layerStatesArray);
```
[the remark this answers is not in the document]
fredj: See a4fe067aad5ee19be3fca953e8232fde7c3a25c6

src/ol/renderer/Map.js
```diff
@@ -250,15 +251,14 @@ class MapRenderer extends Disposable {
...
    * @param {import("../PluggableMap.js").FrameState} frameState Frame state.
    * @private
    */
-  removeUnusedLayerRenderers_(map, frameState) {
-    for (const layerKey in this.layerRenderers_) {
-      if (!frameState || !(layerKey in frameState.layerStates)) {
-        this.removeLayerRendererByKey_(layerKey).dispose();
-      }
+  removeUnusedLayerRenderers_(layerKeys, map, frameState) {
+    for (const layerKey in layerKeys) {
```
tschaub: Looks like `layerKey` is going to be the array index here (instead of the layer key).

src/ol/renderer/Map.js
```diff
@@ -276,22 +276,26 @@ class MapRenderer extends Disposable {
...
-        frameState.postRenderFunctions.push(
-          /** @type {import("../PluggableMap.js").PostRenderFunction} */ (this.removeUnusedLayerRenderers_.bind(this))
-        );
-        return;
+      if (!(layerKey in layerStatesMap)) {
+        layerKeys.push(layerKey);
       }
     }
+    if (layerKeys.length > 0) {
+      frameState.postRenderFunctions.push(this.removeUnusedLayerRenderers_.bind(this, layerKeys));
```
tschaub: We could just get rid of the `removeUnusedLayerRenderers_` method altogether and push an anonymous function in the loop above.  Something like:
```js
if (!(layerKey in layerStatesMap)) {
  frameState.postRenderFunctions.push(function() {
    this.removeLayerRendererByKey(layerKey).dispose();
  }.bind(this));
}
```

fredj: yes, that's a very good idea

PR #9078, opened 2018-12-20
Reorder requestFullScreen to match order in getChangeType

### LINKED ISSUE #9077: ol.control.FullScreen triggering incorrect event for latest Chrome
**Describe the bug**
In Chrome Version 71.0.3578.98, entering full screen (using full screen control) appears to no longer use webkitfullscreenchange event, but instead triggers fullscreenchange. I believe that Chrome is only triggering the event that corresponds with element.requestFullscreen() or element.webkitRequestFullscreen() whereas either requestFullscreen was unavailable previously, or it previously raised both events.

Possible fix 1: reorder the if statement within [requestFullScreen](https://github.com/openlayers/openlayers/blob/master/src/ol/control/FullScreen.js#L254) to match the order within [getChangeType](https://github.com/openlayers/openlayers/blob/master/src/ol/control/FullScreen.js#L14)

Possible fix 2: reorder the if condition in getChangeType to match the order in requestFullScreen.

I'm preparing a PR for the 1st approach.

**To Reproduce**
1. Using Chrome Version 71.0.3578.98 (latest as at 20th Dec 2018)
2. Go to http://openlayers.org/en/latest/examples/full-screen-source.html
3. Click on the Full Screen icon
4. See that the icon does not update to an X but is the original icon

**Expected behavior**
The full screen icon should be an X.

spazzarama:
Fixes #9077

### DIFF
--- a/src/ol/control/FullScreen.js
+++ b/src/ol/control/FullScreen.js
@@ -252,14 +252,14 @@ function isFullScreen() {
  * @param {HTMLElement} element Element to request fullscreen
  */
 function requestFullScreen(element) {
-  if (element.requestFullscreen) {
-    element.requestFullscreen();
+  if (element.webkitRequestFullscreen) {
+    element.webkitRequestFullscreen();
   } else if (element.msRequestFullscreen) {
     element.msRequestFullscreen();
   } else if (element.mozRequestFullScreen) {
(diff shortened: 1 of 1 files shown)


ahocevar:
Fixed with #9079.

PR #9090, opened 2018-12-28
Add texture & color to the WebGL points renderer

jahow:
This PR improves a bit the `ol/renderer/webgl/PointsLayer` class by adding the possibility to easily define a texture and change the color of points without writing custom shaders.

An example was added to showcase this, where a dataset of 80k points is rendered with variable icons and colors: https://592-4723248-gh.circle-artifacts.com/0/examples/icon-sprite-webgl.html

I think that the WebGL points renderer is now pretty mature, I still have a few examples in mind using custom shaders though.


jahow:
This PR is ready for review, thanks!

tschaub:
Looks great @jahow.

When we get closer to the 6.0 release, we should talk about a convenient API for a layer like the one in your example.

jahow:
Thanks @tschaub, I agree that there's probably a way to expose a cleaner API for this which does not require extending classes.

PR #9094, opened 2018-12-29
do not cluster features that return null from the geometry function b…

### LINKED ISSUE #9093: v3.20.1 don't hide features that are not clustered
Features that do return null from ol.source.Cluster.geometryFunction are ignored during the clustering process but could be part of a cluster as a neighbor within the distance of another feature that will be clustered and will disappear from the layer.

To reproduce just watch the unit test from v3.20.1 that in my opinion shouldn't be green. Before the cluster process, the layer contains 3 features: a Point, a LineString and a Polygon. After clustering the layer contains just 1 cluster feature containing the Point and the Polygon. The LineString feature has disappeared?
```javascript
it('clusters with a custom geometryFunction', function() {
  var source = new ol.source.Cluster({
    geometryFunction: function(feature) {
      var geom = feature.getGeometry();
      if (geom.getType() == 'Point') {
        return geom;
      } else if (geom.getType() == 'Polygon') {
        return geom.getInteriorPoint();
      }
      return null;
    },
      features: [
        new ol.Feature(new ol.geom.Point([0, 0])),
        new ol.Feature(new ol.geom.LineString([[0, 0], [1, 1]])),
        new ol.Feature(new ol.geom.Polygon(
            [[[-1, -1], [-1, 1], [1, 1], [1, -1], [-1, -1]]]))
      ]
    })
  });
  source.loadFeatures(extent, 1, projection);
  expect(source.getFeatures().length).to.be(1);
  expect(source.getFeatures()[0].get('features').length).to.be(2);
});
```

Features that do return null from ol.source.Cluster.geometryFunction should be ignored during the clustering process as a isolated feature and as a neighbor of other features so it still will be shown on the layer as an isolated not clustered feature.

coemans:
Fix #9093

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5065 @@
+{
+  "name": "openlayers",
+  "version": "3.20.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "JSONStream": {
+      "version": "1.3.5",
+      "resolved": "http://artifactory-pr.lb.cumuli.be:8081/artifactory/api/npm/acd-npm/JSONStream/-/JSONStream-1.3.5.tgz",
+      "integrity": "sha1-MgjB8I06TZkmGrZPkjArwV4RHKA=",
+      "requires": {
+        "jsonparse": "^1.2.0",
+        "through": ">=2.2.7 <3"
+      }
+    },
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "http://artifactory-pr.lb.cumuli.be:8081/artifactory/api/npm/acd-npm/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg="
+    },
+    "absolute": {
+      "version": "0.0.1",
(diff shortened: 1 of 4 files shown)


coemans:
@tschaub how can I fix the CircleCI build? Locally all tests ran successful.

mprins:
The master branch is at 5.3.x, you seem to target v3.20.x which is basically dead (>2yrs old), your PR has merge conflicts so it's unlikely to be merged.

in fact file `src/ol/source/cluster.js` no longer exists

coemans:
Thanks for the reply @mprins!

Unfortunately we can't upgrade to v5 in our project at the moment. Could you provide a new version v3.20.2 with this fix or will there never be new ol3 versions? In that case it is not a problem to depend on our fork branch with the fix.

kekel87:
Hi,

@coemans I looking quickly, the issue is still present on master: [Cluster.js#L166](https://github.com/openlayers/openlayers/blob/master/src/ol/source/Cluster.js#L166)

Any news about integration of this in the current version (v5)?

Thx

coemans:
@kekel87 no news yet.. We've integrated with a fork dependency where I fixed this issue. Maybe you should make a new pull request with the same fixe in the current version?

PR #9098, opened 2019-01-02
External types bugfix for jsdoc-plugin-typescript

ahocevar:
Previous versions of jsdoc-plugin-typescript raised an exception when external types were used. This is fixed in the latest version. This pull request just updates the `devDependencies` to use that fixed version.


ahocevar:
No code change, merging.

PR #9103, opened 2019-01-07
Make examples more user friendly

ahocevar:
This pull request makes the examples more user friendly:

- [x] Separate HTML and JS example code
- [x] Add `package.json` example code
- [x] Example live editing with codesandbox.io

### DIFF
--- a/examples/webpack/example-builder.js
+++ b/examples/webpack/example-builder.js
@@ -10,6 +10,7 @@ const RawSource = require('webpack-sources').RawSource;
 const readFile = promisify(fs.readFile);
 const isCssRegEx = /\.css$/;
 const isJsRegEx = /\.js(\?.*)?$/;
+const importRegEx = /^import .* from '(.*)';$/;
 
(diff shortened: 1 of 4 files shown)


examples/webpack/example-builder.js
```diff
@@ -83,6 +84,32 @@ function getJsSource(chunk, jsName) {
...
+  const dependencies = {
+    ol: pkg.version
+  };
+  for (let i = 0, ii = lines.length; i < ii; ++i) {
+    const line = lines[i];
+    const importMatch = line.match(importRegEx);
+    if (importMatch) {
+      const imp = importMatch[1];
+      if (!imp.startsWith('ol/') && imp != 'ol') {
+        const dep = imp.split('/')[0];
```
tschaub: If we want to handle scoped packages, this could be something like:
```js
const parts = imp.split('/');
let dep;
if (imp.startsWith('@')) {
  dep = parts.slice(0, 2).join('/');
} else {
  dep = parts[0];
}
```

ahocevar: Good suggestion. Done.

Issue #7795, opened 2018-02-08
HiDPI / retina / pixelRatio zoom levels for tiled image sources (Zoomify, etc.)


hyperknot:
Thank you for your interest in making OpenLayers better!

To keep this project manageable for maintainers, we ask you to please check all boxes below before submitting an issue.

- [x] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [x] I have searched GitHub to see if a similar bug or feature request has already been reported.
- [x] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [x] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

When viewing tiled raster images, I believe the right selection strategy for HiDPI screens is to always have close to 2x (or DEVICE_PIXEL_RATIO *) the number of pixels on the zoom level, except on the last zoom level, where 1x sampling should be allowed. 

Unlike on an OSM map with labels, a tiled image sources should underzoom by default on HiDPI screen. That is, an photograph or scanned document is automatically "Retina" or HiDPI if looked at a level lower.

I've almost finished porting the reference IIIF implementation from OpenSeaDragon to OpenLayers, but this one just puzzled me.

// Also, should I wait more before everything settles for 5.0.0 or is Zoomify.js considered stable now?

crubier:
This problem is still there on latest versions. And it's funny but I keep seeing its consequences  everywhere on the web. It is impacting:

 - The application I build at www.sterblue.com
 - Our competitor at www.scopito.com 
 - This astronomy web app https://scope.avocco.com/case/20/eWKcUiIXpuQU9V0z

On all of these, if you zoom in with a retina screen, the image is blurrier than it should be.

crubier:
And my PR #9110 did not fix it, unfortunately

guatedude2:
The problem from looking at the code in `src/ol/source/Zoomify.js` is on line 42 of`CustomTile`. when changing that line to:
```
this.tileSize_ = [512, 512]
```

It fixes the problem, I also tried removing the custom zoomify class on line 254 and it fixes the problem as well. I'm not entirely sure what the `CustomTile` does different than the `ZoomifyTileClass`
```
tileClass: ZoomifyTileClass,
```

crubier:
[...]

PR #9110, opened 2019-01-12
Add TilePixelRatio to Zoomify

crubier:
Fix a bug where you can't specify the pixel ratio of zoomify tiles.

I found this issue that may be related:
https://github.com/openlayers/openlayers/issues/7795

But I encountered the problem myself. I could create an issue but it is only one line of code.


crubier:
Bump ? Need an approver

crubier:
@ahocevar ok done

PR #9116, opened 2019-01-15
add flag unnesessary event in format.GeoJSON readFeatureFromObject

Jinkwon:
Add silent flag to 'setProperties' in readFeatureFromObject method. (format/GeoJSON.js)
Actually unnecessary events in every features.


fredj:
Thanks for the pull request @Jinkwon. In addition to GeoJSON, it can be done for the other format (GPX, KML, ...). See #9118

Jinkwon:
@fredj Thanks so much. 👍 👍

PR #9131, opened 2019-01-21
Don't use strict comparison with style.opacity

fredj:
The comparison is always false because the opacity is a number and style.opacity a string.

### DIFF
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -149,7 +149,7 @@ class CanvasImageLayerRenderer extends CanvasLayerRenderer {
     }
 
     const opacity = layerState.opacity;
-    if (opacity !== canvas.style.opacity) {
+    if (opacity !== parseFloat(canvas.style.opacity)) {
       canvas.style.opacity = opacity;
     }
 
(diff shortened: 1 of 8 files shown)


fredj:
There's an other issue with the opacity from the style: the value is rounded so the comparison could be false.

For example (tested on Chrome):
```js
const element = document.createElement('div');
element.style.opacity = 0.1499999999;
element.style.opacity === "0.15";
```

src/ol/renderer/canvas/ImageLayer.js
```diff
@@ -149,7 +149,7 @@ class CanvasImageLayerRenderer extends CanvasLayerRenderer {
     }
 
     const opacity = layerState.opacity;
-    if (opacity !== canvas.style.opacity) {
+    if (opacity != canvas.style.opacity) {
```
ahocevar: `canvas.style.opacity` is a string, so a strict comparison with an explicit `parseInt(canvas.style.opacity)` would be better.

fredj:
To workaround the float issue with the opacity from the style, the value from the layer state is now rounded: see af8a22d

src/ol/renderer/canvas/ImageLayer.js
fredj: in 8dbe8bc34e563729eba4d92ac773287726c24ea9

Issue #6940, opened 2017-06-20
Memory hunger and high CPU usage, eventually Aw, Snap out of memory


simon998yang:
Hi,

I'm doing a simple testing after I experienced out of memory issue in my project. Here is my testing code(just replaced 'url' to '????' as the server url is bit sensitive)
Just open the page, not doing anything else. In chrome task manager, I see the CPU is constantly high(go up down 15%-35%), and the memory usage spikes from time to time (from 400M to 1GB, 2GB,3GB), and could eventually reach 'Snap, out of memory'.  Please help ...

```
<!doctype html>
<html lang="en">
  <head>
    <link rel="stylesheet" href="https://openlayers.org/en/v4.2.0/css/ol.css" type="text/css">
    <style>
      .map {
        height: 400px;
        width: 100%;
      }
    </style>
    <script src="https://openlayers.org/en/v4.2.0/build/ol.js" type="text/javascript"></script>
    <title>OpenLayers example</title>
  </head>
  <body>
    <h2>My Map</h2>
    <div id="map" class="map"></div>
    <script type="text/javascript">
      var paras = {'LAYERS': '4,0,65,61','VERSION': '1.1.1','FORMAT': 'image/png'};
      var projection = new ol.proj.Projection ({'code': 'EPSG:4326'});
              
      var map = new ol.Map({
        target: 'map',
        layers: [
          new ol.layer.Image({
            source: new ol.source.ImageWMS({'url':'????','params':paras, 'projection':projection})
          })
        ],
        view: new ol.View({
          center: ol.proj.fromLonLat([130.9769230478005, -12.487059768379256]),
          zoom: 10
        })
      });
    </script>
  </body>
</html>
```

ahocevar:
Your map view is in a different projection than your layer source. This means OpenLayers will reproject all images it receives from your WMS. You could either display your map in EPSG:4326 by configuring your view with `projection: projection` and setting the center directly (without `ol.proj.fromLonLat`), or you could try if your WMS supports Web Mercator (usually EPSG:3857) by looking at its GetCapabilities response. Then you'd have to omit the `projection` config option on the ImageWMS source.

simon998yang:
Thanks! Your instruction helped to resolve the problem.

PR #9133, opened 2019-01-21
Reduce memory footprint of tiles and labels

### LINKED ISSUE #8956: iOS Total canvas memory use exceeds the maximum limit
We are developing an application with multiple layers (raster images). 
When I try to render our application on an Ipad if I try to display only tow layers and zoom in and zoom out i get 'Total canvas memory use exceeds the maximum limit (256 MB)'
I try with http://openlayers.org/en/latest/examples/canvas-tiles.html and i get the same error after zooming in and out some time.
I get the same result on my MacBook but it take more time because the limit is (2151 MB).
It's linked with the last webkit update -> https://github.com/WebKit/webkit/commit/5d5b478917c685e50d1032ccf761ca53fc8f1b74#diff-b411cd4839e4bbc17b00570536abfa8f
So is there a workaround ? (Using WebGL ? it's not really good with raster images. Limit cache size ?)

EDIT:
After some investigations it's seams linked with projection:
On this exemple -> https://openlayers.org/en/latest/examples/simple.html?q=simple I can move, zoom etc... when I check the safari dev console no other canvas are created.
But on this exemple -> https://openlayers.org/en/latest/examples/canvas-tiles.html all the canvas of the tile debug are created and make the app crash after some time.
So i think it's related with tiles reprojected.
Same with this exemple: https://codepen.io/unhurried/pen/KrKjKN if I move on the left map everything is fine but if I move on the right map there is a lot of canvas creation.

ahocevar:
This pull request drastically reduces the tile cache size. Previously, each raster tile layer got a maximum of 2048 tiles cached. With only a few layers, this can easily exceed the browser memory. Each vector tile layer previously got a maximum of 128 tiles cached, often causing the canvas memory to be exceeded in Safari.

With this change, the tile cache is set on a per-layer basis, taking into account the screen size and the tile size. It makes no sense to cache hundreds or thousands of tiles, because the browser caches requested data anyway. For vector tiles, it would be desirable to keep pre-rendered images around for a little bit longer, but recent changes in Safari reduced the available canvas memory, so we have to cache less tiles here too.

Fixes #8956, addresses #6940 and other issues about exceeding the browser memory.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -47,7 +47,7 @@ import {create as createTransform, apply as applyTransform} from './transform.js
  * @property {import("./size.js").Size} size
  * @property {!Object<string, boolean>} skippedFeatureUids
  * @property {TileQueue} tileQueue
- * @property {Object<string, Object<string, import("./TileRange.js").default>>} usedTiles
+ * @property {!Object<string, Object<string, boolean>>} usedTiles
  * @property {Array<number>} viewHints
  * @property {!Object<string, Object<string, boolean>>} wantedTiles
  */
--- a/src/ol/TileCache.js
+++ b/src/ol/TileCache.js
@@ -21,8 +21,7 @@ class TileCache extends LRUCache {
   expireCache(usedTiles) {
     while (this.canExpireCache()) {
       const tile = this.peekLast();
-      const zKey = tile.tileCoord[0].toString();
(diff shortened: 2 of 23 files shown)


soberdor:
Thank you very much for this!

ahocevar:
Thanks for testing @fredj. Looking into this now.

PR #9137, opened 2019-01-22
Refactor the way view constraints are handled & add a view extent constraint

jahow:
This replaces #9071.

The aim of this PR is to streamline the way view constraints are handled and shift the constraint responsibility from the interactions to the `ol.View` class, allowing for more possibilities. Specifically, this PR adds:
* a true extent constraint (previously was only applied to the view center)
* smoothing effects when the view extent/zoom goes out of bounds

Public API changes:
* To the `ol.View` constructor options:
  * *(breaking)* to get the previous behaviour, the flag `constrainCenterOnly` must be set to true
  * Added the option `smoothExtentConstraint`; must be set to false to disable the smoothing effect on the extent constraint; default: true
  * Added the option `smoothResolutionConstraint`; must be set to false to disable the smoothing effect on the zoom constraint; default: true
  * Added the option `constrainResolution`; must be set to false to allow staying on intermediary zoom levels (which was the case previously); if true, view will always snap to the nearest valid resolution
  * Added the method `setResolutionConstrained`
  * *(breaking)* Removed the `rotate` method
  * Added the `adjustRotation`, `adjustResolution` and `adjustZoom` methods, which all take an optional anchor parameter; these are now used by interactions to manipulate the view without having to compute the new center; this was required as interactions do not have access to the internal *target* values of the view
  * Added the `move` method to change the view center (same logic as above, only there is no anchor computation involved)
* *(breaking)* Interactions and the `ol.View#fit` method do not support the `constrainResolution` option anymore; this is now handled on the `ol.View` constructor (see above)

Internal API changes:
* To center, resolution and rotation constraints:
  * All constraints now receive an `opt_isMoving` argument which is true if an interaction/animation is underway; this allows the constraints to give a different result depending on the view state, for example allowing intermediary zoom levels or slightly going out of bounds
  * The constraints no longer receive a `delta` argument, which was in some case misleading; the general logic is now that they now simply process one input value into one output value
* To the `ol.View` class:
  * Private members `targetRotation_`, `targetCenter_` and `targetResolutions_` were added; these are required to be able to apply constraints which are not "repeatable" (ie applying a 0.5 ratio to a value)
  * Added the `applyParameters_` private method: this computes the actual parameters (center/rotation/resolution) of the view using the `target*` values, and taking into account whether an interaction/animation is underway
  * Added the `resolveConstraints_` private method: this will animate the view to a "rested" state if necessary, typically at the end of an interaction or a chain of animation
  * Added the `getValidResolution`, `getValidZoomLevel` and `getValidCenter` methods so that interactions and controls can get constrained values without having to compute anything
  * Added `beginInteraction` and `endInteraction`: this should be used instead of e.g. `setHint(ViewHint.INTERACTING, 1)`; this way, when an interaction ends, the view is able to call `resolveConstraints_` automatically

Generally speaking I've attempted to streamline the computation of zoom and resolution values. The view still has `getZoomForResolution` and `getResolutionForZoom` which should be the only way to convert one value to another. Internally, all computations are done using resolutions.

A new example was added to showcase the new constraint: `extent-constrained.html`
https://682-4723248-gh.circle-artifacts.com/0/examples/extent-constrained.html

This is still WIP but tests should be green and things work well. Remaining to do:
- [x] Fixes the `PinchZoom` and `DragRotateAndZoom` interactions to better handle the resolution constraint
- [x] Fixes the `DragPan` interaction and `fit` method to avoid going out of bounds too much
- [x] More tests inc. on mobile
- [x] Update documentation


ahocevar:
Thanks for your continued effort on this very nice improvement @jahow. At this point, without having looked at the code, I just wanted to add a few notes on the public API changes:

* I like the fact that `constrainResolution` is now set in one central place (the view), instead of having to configure it for `View#fit()` and every interaction.
* There is no need to rename `extent` to `maxExtent` - there is no other extent related configuration on the view, so it makes sense to stick with what we had.
* The default for the new `constrainResolution` option should be `false` instead of `true`, because it gives us the desirable behavior we currently have.
* I see no need to expose the `smoothResolutionConstraint` and `smoothExtentConstraint` options.

jahow:
Thanks for the input @ahocevar, this makes sense. `maxExtent` was simply an attempt to make the option clearer and prevent mixing it up with the `extent` options of layers.

tschaub:
This sounds great @jahow.  I agree with @ahocevar that unconstrained resolutions is nicer, but this may be most pronounced on devices with trackpads (that continue to send scroll events after user interaction ends).  If the case can be made that different defaults are better on different devices, then we could introduce an ‘auto’ default that picks based on a heuristic.

jahow:
> This sounds great @jahow. I agree with @ahocevar that unconstrained resolutions is a nicer default, but this may be most pronounced on devices with trackpads (that continue to send scroll events after user interaction ends). If the case can be made that different defaults are better on different devices, then we could introduce an ‘auto’ default that picks based on a heuristic.

This is actually something I was kinda stuck on. Current implementation allows unconstrained resolutions for some interactions but not for others. Transferring this setting to the View meant breaking either the first or the latter.

I like your idea of an `auto` option. Another thing could be using `view.setResolutionConstrained` on the interactions to switch the mode from constrained to unconstrained according to user actions. But this doesn't sound as nice.

jahow:
Documentation was updated, this PR is now ready for review. Thanks!

src/ol/View.js
```diff
@@ -184,46 +196,62 @@ const DEFAULT_MIN_ZOOM = 0;
...
+ * view can cause issues because the way constraints are applied implicitly.
+ * Typically, calling `setResolution(10)` does not guarantee that
+ * `getResolution()` will return `10`.
  *
-  * @api
+ * To avoid this, one can use `move`, `adjustRotation`, `adjustZoom` and
+ * `adjustResolution` in order to change the view state while letting the
+ * view do its internal computations. The `adjust*` methods also take an `opt_anchor`
+ * argument which allows specifying an origin for the transformation.
+ *
```
ahocevar: Does this mean that using the `adjust*` methods allows users to bypass constraints? What will be the behavior of the view when a constraint is broken this way?

jahow: They do not bypass constraints but simply apply the given delta to the corresponding `target*` value, handle the computation related to the anchor and eventually call the associated `set*` methods (which goes through the constraint system).

With this refactoring an external consumer should not do these computations by itself (ie get the current value, apply a delta, take into account an anchor, and call `set*` with the result) as the result might be incorrect because of contraints.

src/ol/View.js
```diff
@@ -184,46 +196,62 @@ const DEFAULT_MIN_ZOOM = 0;
...
  * The *center constraint* is determined by the `extent` option. By
- * default the center is not constrained at all.
+ * default the view center is not constrained at all.
+ *
+ * ### Changing the view state
+ *
+ * Using only zoom, resolution, center and rotation setters to manipulate the
+ * view can cause issues because the way constraints are applied implicitly.
+ * Typically, calling `setResolution(10)` does not guarantee that
+ * `getResolution()` will return `10`.
```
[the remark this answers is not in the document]
jahow: Yes, the important thing is that one should use `adjust*` methods when applying a delta, instead of doing `get*` and then `set*`. I'll reword that to make it clearer.

src/ol/View.js
```diff
@@ -1104,30 +1132,74 @@ class View extends BaseObject {
   }
 
   /**
-   * Rotate the view around a given coordinate.
-   * @param {number} rotation New rotation value for the view.
+   * Adds relative coordinates to the center of the view. Any extent constraint will apply.
+   * @param {import("./coordinate.js").Coordinate} deltaCoordinates Relative value to add.
+   * @api
+   */
+  move(deltaCoordinates) {
```
ahocevar: As already said, I think `adjustCenter` would be consistent with the other new methods.

src/ol/View.js
```diff
@@ -1142,39 +1214,166 @@ class View extends BaseObject {
...
+
+  /**
+   * Get a valid position for the view center according to the current constraints.
+   * @param {import("./coordinate.js").Coordinate|undefined} targetCenter Target center position.
+   * @param {number=} opt_targetResolution Target resolution. If not supplied, the current one will be used.
+   * This is useful to guess a valid center position at a different zoom level.
+   * @return {import("./coordinate.js").Coordinate|undefined} Valid center position.
+   * @api
+   */
+  getValidCenter(targetCenter, opt_targetResolution) {
```
ahocevar: Does this need to be `@api`? I'd rather keep this away from the user if possible. A better name (i.e. more consistent with what we have elsewhere) would be `getConstrainedCenter`

src/ol/View.js
```diff
@@ -1142,39 +1214,166 @@ class View extends BaseObject {
...
+  /**
+   * Get a valid zoom level according to the current view constraints.
+   * @param {number|undefined} targetZoom Target zoom.
+   * @param {number=} opt_direction Direction. Default is `0`. Specify `-1` or `1` to return
+   * the available value respectively lower or greater than the target one. Leaving `0` will simply choose
+   * the nearest available value.
+   * @return {number|undefined} Valid zoom level.
+   * @api
+   */
+  getValidZoomLevel(targetZoom, opt_direction) {
```
ahocevar: Same as above re `@api`, and a better/simpler name would be `getConstrainedZoom`.

src/ol/View.js
```diff
@@ -1142,39 +1214,166 @@ class View extends BaseObject {
...
+  /**
+   * Get a valid resolution according to the current view constraints.
+   * @param {number|undefined} targetResolution Target resolution.
+   * @param {number=} opt_direction Direction. Default is `0`. Specify `-1` or `1` to return
+   * the available value respectively lower or greater than the target one. Leaving `0` will simply choose
+   * the nearest available value.
+   * @return {number|undefined} Valid resolution.
+   * @api
+   */
+  getValidResolution(targetResolution, opt_direction) {
```
ahocevar: Same as above re `@api`, and a better name would be `getConstrainedResolution`.

src/ol/View.js
```diff
@@ -1195,7 +1394,8 @@ function animationCallback(callback, returnValue) {
  */
 export function createCenterConstraint(options) {
```
ahocevar: This is more a question or something for a separate pull request, but I always wondered what we have this function exported for when we don't have an API to apply a constraint we created with this function. Or do we have an API for that and I'm just missing something?

src/ol/resolutionconstraint.js
```diff
@@ -2,37 +2,76 @@
...
+ * @typedef {function((number|undefined), number, import("./size.js").Size, boolean=): (number|undefined)} Type
  */
 
+/**
+ * @param {number} resolution Resolution
+ * @param {import("./extent.js").Extent=} maxExtent Maximum allowed extent.
+ * @param {import("./size.js").Size} viewportSize Viewport size.
+ * @return {number} Capped resolution.
+ */
+function getCappedResolution(resolution, maxExtent, viewportSize) {
```
ahocevar: Do you mean `getClampedResolution`? I don't fully understand what this function does. Can you explain?

src/ol/resolutionconstraint.js
```diff
@@ -2,37 +2,76 @@
...
+  return Math.min(resolution, Math.min(xResolution, yResolution));
+}
+
+/**
+ * @param {number} resolution Resolution.
+ * @param {number} maxResolution Max resolution.
+ * @param {number} minResolution Min resolution.
+ * @return {number} Smoothed resolution.
+ */
+function getSmoothedResolution(resolution, maxResolution, minResolution) {
```
ahocevar: What does this function do? I'm failing to understand.

src/ol/resolutionconstraint.js
```diff
@@ -44,29 +83,74 @@ export function createSnapToResolutions(resolutions) {
...
+        if (opt_isMoving) {
+          const smooth = opt_smooth !== undefined ? opt_smooth : true;
+          if (!smooth) {
+            return clamp(resolution, minResolution, cappedMaxRes);
+          }
+          return getSmoothedResolution(resolution, cappedMaxRes, minResolution);
         }
-        return maxResolution / Math.pow(power, newLevel);
+
+        const offset = -direction * (0.5 - 1e-9) + 0.5;
```
ahocevar: What exactly does this calculation do? I'm particularly wondering about `1e-9`.

test/spec/ol/interaction/dragzoom.test.js
```diff
@@ -32,7 +32,8 @@ describe('ol.interaction.DragZoom', function() {
       view: new View({
         projection: 'EPSG:4326',
         center: [0, 0],
-        resolution: 1
+        resolution: 1,
+        maxResolution: 8
```
ahocevar: Why was `maxResolution: 8` added? Does the test pass without that addition too?

test/spec/ol/interaction/interaction.test.js
```diff
@@ -87,67 +86,4 @@ describe('ol.interaction.Interaction', function() {
 
   });
 
-  describe('zoomByDelta()', function() {
```
ahocevar: It would be good to bring this test coverage back, using functions from the new implementation. Or maybe you have added these tests elsewhere? If so, please point me to them.

test/spec/ol/map.test.js
```diff
@@ -601,21 +600,6 @@ describe('ol.Map', function() {
         const interactions = defaultInteractions(options);
         expect(interactions.getLength()).to.eql(1);
         expect(interactions.item(0)).to.be.a(PinchZoom);
-        expect(interactions.item(0).constrainResolution_).to.eql(false);
-      });
-    });
-
-    describe('set constrainResolution option', function() {
```
ahocevar: Have these tests also been moved elsewhere?

test/spec/ol/resolutionconstraint.test.js
```diff
@@ -12,30 +13,30 @@ describe('ol.resolutionconstraint', function() {
...
     });
 
-    describe('zoom out', function() {
+    describe('direction -1', function() {
       it('returns expected resolution value', function() {
-        expect(resolutionConstraint(1000, -1, 0)).to.eql(1000);
-        expect(resolutionConstraint(500, -1, 0)).to.eql(1000);
-        expect(resolutionConstraint(250, -1, 0)).to.eql(500);
-        expect(resolutionConstraint(100, -1, 0)).to.eql(250);
+        expect(resolutionConstraint(1000, -1)).to.eql(1000);
```
ahocevar: Why are the results different now?

test/spec/ol/resolutionconstraint.test.js
```diff
@@ -50,42 +51,42 @@ describe('ol.resolutionconstraint', function() {
...
       it('returns expected resolution value', function() {
-        expect(resolutionConstraint(1050, 1, 0)).to.eql(500);
-        expect(resolutionConstraint(950, 1, 0)).to.eql(500);
-        expect(resolutionConstraint(550, 1, 0)).to.eql(250);
-        expect(resolutionConstraint(450, 1, 0)).to.eql(250);
-        expect(resolutionConstraint(300, 1, 0)).to.eql(100);
-        expect(resolutionConstraint(200, 1, 0)).to.eql(100);
-        expect(resolutionConstraint(150, 1, 0)).to.eql(100);
-        expect(resolutionConstraint(50, 1, 0)).to.eql(100);
+        expect(resolutionConstraint(1050, 1)).to.eql(1000);
```
ahocevar: Why are the results different now?

test/spec/ol/resolutionconstraint.test.js
```diff
@@ -50,42 +51,42 @@ describe('ol.resolutionconstraint', function() {
...
-        expect(resolutionConstraint(1050, -1, 0)).to.eql(1000);
-        expect(resolutionConstraint(950, -1, 0)).to.eql(1000);
-        expect(resolutionConstraint(550, -1, 0)).to.eql(1000);
-        expect(resolutionConstraint(450, -1, 0)).to.eql(1000);
-        expect(resolutionConstraint(300, -1, 0)).to.eql(500);
-        expect(resolutionConstraint(200, -1, 0)).to.eql(500);
-        expect(resolutionConstraint(150, -1, 0)).to.eql(250);
-        expect(resolutionConstraint(50, -1, 0)).to.eql(250);
+        expect(resolutionConstraint(1050, -1)).to.eql(1000);
+        expect(resolutionConstraint(950, -1)).to.eql(500);
```
ahocevar: Same question re different results.

test/spec/ol/resolutionconstraint.test.js
```diff
@@ -96,54 +97,54 @@ describe('ol.resolutionconstraint', function() {
 
     beforeEach(function() {
       resolutionConstraint =
-          createSnapToPower(2, 1024, 10);
+          createSnapToPower(2, 1024, 1);
```
ahocevar: Has the meaning of the 3rd argument changed? Or why is 1 used instead of 10 now?

test/spec/ol/resolutionconstraint.test.js
```diff
@@ -96,54 +97,54 @@ describe('ol.resolutionconstraint', function() {
...
-        expect(resolutionConstraint(256, 1, 0)).to.eql(128);
-        expect(resolutionConstraint(128, 1, 0)).to.eql(64);
-        expect(resolutionConstraint(64, 1, 0)).to.eql(32);
-        expect(resolutionConstraint(32, 1, 0)).to.eql(16);
-        expect(resolutionConstraint(16, 1, 0)).to.eql(8);
-        expect(resolutionConstraint(8, 1, 0)).to.eql(4);
-        expect(resolutionConstraint(4, 1, 0)).to.eql(2);
-        expect(resolutionConstraint(2, 1, 0)).to.eql(1);
-        expect(resolutionConstraint(1, 1, 0)).to.eql(1);
+        expect(resolutionConstraint(1024, 1)).to.eql(1024);
```
ahocevar: Why do we have different results now?

test/spec/ol/resolutionconstraint.test.js
```diff
@@ -96,54 +97,54 @@ describe('ol.resolutionconstraint', function() {
...
-        expect(resolutionConstraint(256, -1, 0)).to.eql(512);
-        expect(resolutionConstraint(128, -1, 0)).to.eql(256);
-        expect(resolutionConstraint(64, -1, 0)).to.eql(128);
-        expect(resolutionConstraint(32, -1, 0)).to.eql(64);
-        expect(resolutionConstraint(16, -1, 0)).to.eql(32);
-        expect(resolutionConstraint(8, -1, 0)).to.eql(16);
-        expect(resolutionConstraint(4, -1, 0)).to.eql(8);
-        expect(resolutionConstraint(2, -1, 0)).to.eql(4);
-        expect(resolutionConstraint(1, -1, 0)).to.eql(2);
+        expect(resolutionConstraint(1024, -1)).to.eql(1024);
```
ahocevar: Same question re different results.

src/ol/View.js
```diff
@@ -793,6 +813,15 @@ class View extends BaseObject {
     this.applyOptions_(this.getUpdatedOptions_({minZoom: zoom}));
   }
 
+  /**
+   * Set whether the view shoud allow intermediary zoom levels.
+   * @param {boolean} enabled Whether the resolution is constrained.
+   * @api
+   */
+  setResolutionConstrained(enabled) {
```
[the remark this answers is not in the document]
jahow: Will change

examples/pinch-zoom.js
```diff
@@ -7,9 +7,7 @@ import OSM from '../src/ol/source/OSM.js';
 
 const map = new Map({
   interactions: defaultInteractions({pinchZoom: false}).extend([
```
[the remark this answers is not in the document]
jahow: Will change

src/ol/View.js
```diff
@@ -597,9 +640,14 @@ class View extends BaseObject {
...
       this.updateAnimationKey_ = requestAnimationFrame(this.updateAnimations_);
     }
+
+    if (!this.getAnimating()) {
+      setTimeout(this.resolveConstraints_.bind(this));
+    }
   }
 
   /**
+   * @private
```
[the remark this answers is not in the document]
jahow: Will remove

src/ol/View.js
```diff
@@ -1142,39 +1214,166 @@ class View extends BaseObject {
...
+
+  /**
+   * Recompute rotation/resolution/center based on target values.
+   * Note: we have to compute rotation first, then resolution and center considering that
+   * parameters can influence one another in case a view extent constraint is present.
+   * @param {boolean=} opt_doNotCancelAnims Do not cancel animations.
+   * @param {boolean=} opt_forceMoving Apply constraints as if the view is moving.
+   * @private
+   */
+  applyParameters_(opt_doNotCancelAnims, opt_forceMoving) {
```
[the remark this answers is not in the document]
jahow: True, this name could be better. Other possibilities: `applyPendingState_`or `applyTargetState_`?

src/ol/View.js
jahow: Agreed, will change (+ the others)

src/ol/View.js
jahow: No idea TBH. Probably of a remnant of a past design decision.

src/ol/resolutionconstraint.js
jahow: I'll clarify this (basically the resolution is clamped by taking into account the viewport extent)

src/ol/resolutionconstraint.js
jahow: This applies a "smooth" clamp, meaning the resolution is allowed to be slightly out of bounds. I'll clarify it

src/ol/View.js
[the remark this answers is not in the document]
ahocevar: `applyTargetState_` sounds good. Thanks!

src/ol/View.js
ahocevar: Probably, yes. Thanks for confirming.

test/spec/ol/resolutionconstraint.test.js
jahow: The test suite has changed a lot since the resolution constraint signature does not include a `delta` parameter any more.

Previously: `resolutionConstraint(resolution: number, delta: number, direction: number)`
Now: `resolutionConstraint(resolution: number, direction: number, ...)`

So the "delta 0", "zoom in" and "zoom out" cases do not make sense any more. Only the direction parameter is varying

test/spec/ol/resolutionconstraint.test.js
jahow: See comment above

test/spec/ol/resolutionconstraint.test.js
jahow: See comment above

test/spec/ol/resolutionconstraint.test.js
jahow: Yes, the third argument is now `minResolution` where it was before `maxZoomLevel`.
This was part of an effort to reduce mixing zoom levels and resolution values.

test/spec/ol/resolutionconstraint.test.js
jahow: See first comment on this file

test/spec/ol/interaction/interaction.test.js
jahow: You will find these tests in the `view.test` file now, including some more:
https://github.com/openlayers/openlayers/blob/7ca32cd40b0d3169e8b8a429446088939d9b5aca/test/spec/ol/view.test.js#L1606-L1713

test/spec/ol/map.test.js
jahow: Since the `constrainResolution` option does not exist on interactions anymore, I felt this test was rendered useless. I may be wrong though.

jahow:
Thanks for the time spent on this @ahocevar, I really appreciate it. I hope things are clearer now, esp. regarding tests. I've generally tried to maintain and augment the test coverage where possible, but of course I'm open to suggestions. Also the modifications to the constraints API were indeed pretty radical, but for the better IMO.

As for modifications to actual code I'll make another commit with them.

test/spec/ol/interaction/dragzoom.test.js
jahow: This is not needed, will remove

src/ol/resolutionconstraint.js
jahow: This is needed in the case where the direction is -1: the offset applied to the zoom level should be just below 1, otherwise even on a round value the zoom level will be increased at the following line:
https://github.com/openlayers/openlayers/blob/7ca32cd40b0d3169e8b8a429446088939d9b5aca/src/ol/resolutionconstraint.js#L116-L117

src/ol/View.js
jahow: Actually I lied: this is exported to be used in tests here:
https://github.com/openlayers/openlayers/blob/3da449d77e6fa76fcdceae7bdbdde61d429455bd/test/spec/ol/view.test.js#L2-L5

jahow:
I've updated the PR commits following @ahocevar suggestions. The tests are still the same, but the doc annotations are better & some function names have been changed for clarity (see last commit).

If anyone can give this a second (or third, or...) look, that would be absolutely great. Thanks!

ahocevar:
Thanks for your continued effort on this @jahow, and for the detailed explanations. This helps a lot, and I'll give this pull request another look over the weekend.

jahow:
@ahocevar your two last points are fixed with two new commits. I hope this is the end of the tunnel!

tschaub:
I might be seeing the same thing that @ahocevar [mentioned above](https://github.com/openlayers/openlayers/pull/9137#pullrequestreview-204536151).  In the [constrained extent example](https://845-4723248-gh.circle-artifacts.com/0/examples/extent-constrained.html), when I zoom out with the pointer over Annecy, I get some odd shifting back and forth:

constrained

It also looks like there is no limit to how far out you can zoom.  Not a huge deal, but it could be nice to cap things at min - 1 (or whatever it was before).

simple

jahow:
@tschaub you were right, the computations for the center & resolution constraints weren't optimal. I have added a new commit which makes them better: https://github.com/openlayers/openlayers/pull/9137/commits/1d177f9e8d4210e83aa95fe7ed41983c3bff5907

Updated example is here: https://867-4723248-gh.circle-artifacts.com/0/examples/extent-constrained.html

ahocevar:
Thanks @jahow. The only remaining issue now, like @tschaub mentioned, is that zooming out should be capped at `minZoom - 1` (or whatever it currently is). The way it is now, it is possible to zoom out far enough to make the map disappear entirely, without a chance of getting it back:
feb-26-2019 06-24-02

jahow:
> Thanks @jahow. The only remaining issue now, like @tschaub mentioned, is that zooming out should be capped at `minZoom - 1` (or whatever it currently is). The way it is now, it is possible to zoom out far enough to make the map disappear entirely, without a chance of getting it back:

Wait, I thought this was fixed. What are you using to zoom out?

ahocevar:
A trackpad.

jahow:
I can't reproduce it with chromium and firefox on debian. I've tried with safari using an emulator but it did not pick up the trackpad events.

ahocevar:
A trackpad has an inertia, so it sends many events when two-finger dragging fast. Maybe you can reproduce it with a mouse with very fast wheel movement.

jahow:
I reproduced it, it was obvious, sorry! I have amended the last commit to prevent the resolution from going too far away from bounds.

jahow:
@ahocevar @tschaub I think the resolution and center constraints now behave correctly. I have improved a bit more the center constraint to better handle the case where the view is zoomed out beyond the limit.

ahocevar:
Thanks @jahow. I'm going to give this a final review today.

examples/extent-constrained.html
```diff
@@ -0,0 +1,9 @@
+---
+layout: example.html
+title: Constrained Extent
+shortdesc: Example of a view with a constrained extent.
+docs: >
+  This map has a view that is constrained in an extent.  This is done using the `extent` view option. Please note that specifying `restrictOnlyCenter: true` would only apply the extent restriction to the view center.
```
ahocevar:
```suggestion
  This map has a view that is constrained in an extent.  This is done using the `extent` view option. Please note that specifying `constrainOnlyCenter: true` would only apply the extent restriction to the view center.
```

jahow:
@tschaub do you think you could give this PR a second look? I'm afraid merge conflicts might show up if it stays in limbo too long. Thanks!

ahocevar:
I'm going to merge this now. If @tschaub finds things that need to be changed, those can be handled in follow-up pull requests.

Thanks again for the great work!

tschaub:
Thanks for all the work on this @jahow – and for your review work @ahocevar.

PR #9144, opened 2019-01-23
Methods to create Degrees Decimal Minutes (DDM)

darkscript:
Methods to create Degrees Decimal Minutes (DDM), commonly system used in aviation

### DIFF
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -190,6 +190,33 @@ export function degreesToStringHDMS(hemispheres, degrees, opt_fractionDigits) {
 }
 
 
+
+/**
+ * @param {string} hemispheres Hemispheres.
+ * @param {number} degrees Degrees.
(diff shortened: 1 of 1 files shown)


ThomasG77:
What the difference with https://openlayers.org/en/latest/apidoc/module-ol_coordinate.html#.toStringHDMS ?

darkscript:
In Degrees Decimal Minutes (DDM) , we don't give seconds and instead give decimal minutes unlike in DMS where we give seconds, not very important but I got special users who rely on that 

> 
> 
> What the difference with https://openlayers.org/en/latest/apidoc/module-ol_coordinate.html#.toStringHDMS ?

ahocevar:
Thanks for the contribution! Please fix the code according to the [CI output](https://circleci.com/gh/openlayers/openlayers/663?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). It would also be great if you could add simple unit tests for the new code.

PR #9152, opened 2019-01-25
Vector tile optimizations

ahocevar:
This pull request adds a few optimizations for better (vector) tile rendering:

* Now that we do not have `loadTilesWhileAnimating` and `loadTilesWhileInteracting` options any more (#9101), we can ensure the best possible user experience on slow devices by automatically throttling tile loading when we are low on frame budget.
* Do not keep vector tiles without pre-rendered image in the tile image render queue. Since tiles that need an image render will be re-added to the queue anyway, we avoid using render frame budget for tiles that we do not need any more.
* With the removal of clipping to tile borders (#9150), we reduce canvas memory consumption, and can allow larger tile caches. The defaults we have now are good for maps with up to 5 vector tile layers.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -934,8 +934,9 @@ class PluggableMap extends BaseObject {
       if (frameState) {
         const hints = frameState.viewHints;
         if (hints[ViewHint.ANIMATING] || hints[ViewHint.INTERACTING]) {
-          maxTotalLoading = 8;
-          maxNewLoads = 2;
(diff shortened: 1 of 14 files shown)


ahocevar:
With these changes, the `mapbox-style.html` example with 5 vector tile layers and 1 image tile layer also works smoothly on iOS 12.1, despite its memory limitations (#8956): https://78-4974939-gh.circle-artifacts.com/0/examples/mapbox-style.html.

PR #9155, opened 2019-01-25
[WIP] [POC] Experiments with MVT rendering using web workers

gberaudo:
THIS FOR DISCUSSIONS, DO NOT MERGE.

You can test with:
- npm run serve-examples
- chromium-browser http://localhost:8080/mapbox-vector-tiles-custom-worker.html

Make sure to run it in chrome as it currently relies on Offscreen canvas (style uncoupling is a WIP).
Here is the workflow:
- mainThread renderer#getTile() IDLE -> LOADING
- worker renderer#prepareTileInWorker()
- worker renderer#renderTileImage()
- mainThread renderer#onWorkerMessageReceived() LOADING -> LOADED
- mainThread renderer#renderFrame() (compose the tiles, no change)

By changing the useWorker constant in mapbox-vector-tiles-custom-worker.js you can test the two modes:
- with worker;
- without worker (same as before).

The POC implements the logics for all the render modes.
Currently, it only works with image render mode.
As soon as style uncoupling is done it will work in hybrid mode.

I compared worker / non-worker and the performance gain is very clear.
The frame rate is always around 60 and it feels very smooth.
I am very satisfied with this finding, it validates we can expect a major gain by using web workers.
It also shows Offscreen canvas is a great feature we can use as an extra performance gain, if it is available.

I will continue working on the POC. Notably, I want to fix/investigate these issues:
- [x] after some time there are unrendered blocks on the screen;
- [x] browsers without offscreencanvas support (limit to polygon + line instructions creation and copy from worker).
- [ ] some tiles lack some symbols (looks like rendering happens before all resources are loaded);
- [ ] hybrid render mode;


gberaudo:
@ahocevar, I have reworked the PR following our discussions:
- create a dedicated custom MVT example;
- minimize changes in the library;
- continue loading on non-hifi tiles.

Could you please have a new look before I invest more time in cleaning the example itself and splitting the changes into smaller commits?

Regarding the library changes I:
- added setters to allow overriding exported functions (ES6 exports are immutable);
- put custom logics into a CustomVectorTileLayer and a CustomVectorTileLayerRenderer that inherits from their library counterparts; these files are in the examples directory;
- kept some small conditions where it made sense.

The code is now clearer but the issue with empty tiles appearing after some time has not been fixed. Do you have other ideas about what could produce this behaviour?

src/ol/loadImage.js
```diff
@@ -0,0 +1,46 @@
+
```
ahocevar: For the purpose of a single occurrence, this is quite a lot of code. Couldn't you solve this by shimming `Image` in your worker? Something like
```js
 class ImageShim {
  constructor(opt_width, opt_height) {
    this.width = opt_width;
    this.height = opt_height;
  }
  set src(value) {
    // implement worker image loading here
  }
}

self.Image = ImageShim;
```

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -72,18 +72,22 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
...
-    const style = container.style;
-    style.position = 'absolute';
-    style.width = '100%';
-    style.height = '100%';
+    let container = null;
 
-    container.appendChild(baseCanvas);
-    container.appendChild(overlayCanvas);
+    const overlayCanvas = this.overlayContext_.canvas;
+    if (overlayCanvas.style) {
```
ahocevar: When your `createCanvas` fallback has a `style`, you don't need this condition. The only change here that makes sense would be adding a `createContainer(baseCanvas, overlayCanvas)` method, which you can override in your custom renderer.
```js
  createContainer(baseCanvas, overlayCanvas) {
    const container = document.createElement('div');
    const style = container.style;
    style.position = 'absolute';
    style.width = '100%';
    style.height = '100%';

    container.appendChild(baseCanvas);
    container.appendChild(overlayCanvas);

    return container;
  }
```

gberaudo:
@ahocevar, I finally found out what caused the transparent tiles. I will update the PR with the fix and the changes you asked in comments.

src/ol/loadImage.js
gberaudo: The idea is very appealing and it would work well for instantiating an image-like object.
However consumption would not work: how could the shim be considered like a normal image by the browser? What API to use to return the data?

@ahocevar, I see there are 2 other places where we do `new Image()` in the code, if you are OK with introducing a `loadImage` utility function I can refactor to use it there too. What do you think?

ahocevar: @gberaudo You would only shim this in a worker. And if your ImageShim is a `bitmaprenderer` canvas (instead of a plain object), with a src setter added, you should be fine.
```js
class ImageShim extends OffscreeenCanvas {
  set src(value) {
    // implement worker image loading here
  }
}
```
I'd hold off on using a `loadImage` function everywhere, at least until @fredj's experiments with `Image#decode` are completed in #9130.

gberaudo:
@ahocevar, I implemented your fallback suggestions.
I refactored the code to have a cleaner `loadImage` function, I will think about your idea of extending an offscreen canvas.

src/ol/loadImage.js
gberaudo: @ahocevar, I implemented your idea and it gives great results but is not compatible with transpilation to ES5. Here is the error message I get when extending the `OffscreenCanvas`:

`Please use the 'new' operator, this DOM object constructor cannot be called as a function`.

So, are you OK to go with my initial proposal?

ahocevar: @gberaudo Please elaborate. Where exactly do you get this error? Which command sequence? I'm still not ok with your initial proposal, for the reasons already pointed out.

src/ol/loadImage.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ *
+ * @param {string} src Source.
+ * @param {Options} options Options.
+ * @param {function((HTMLImageElement|ImageBitmap)): any} onSuccess Success .
+ * @param {function(any): any} onError Error callback.
+ */
+export function loadImageUsingDom(src, options, onSuccess, onError) {
+  try {
+    const image = new Image();
```
ahocevar: Always using try/catch adds a significant performance penalty.

src/ol/loadImage.js
gberaudo: @ahocevar, in https://github.com/openlayers/openlayers/commit/824cd16644232e07908f159f75a453a96c77bac0#diff-64dc856a6008c886102c4a48e74cbd6aR73, the transpiler transforms:

```js
export default class WorkerImage extends OffscreenCanvas {

  /**
   * @param {number} [width] Canvas width.
   * @param {number} [height] Canvas height.
   */
  constructor(width, height) {
    super(width || 0, height || 0);
    this.width_ = width;
    this.height_ = height;
    this.previousListeners_ = {};
  }
```

to

```js
var WorkerImage = /*@__PURE__*/(function (OffscreenCanvas) {
  function WorkerImage(width, height) {
    OffscreenCanvas.call(this, width || 0, height || 0);
    this.width_ = width;
    this.height_ = height;
    this.previousListeners_ = {};
  }

  if ( OffscreenCanvas ) WorkerImage.__proto__ = OffscreenCanvas;
  WorkerImage.prototype = Object.create( OffscreenCanvas && OffscreenCanvas.prototype );
  WorkerImage.prototype.constructor = WorkerImage;
```

The browser throws the above error when the constructor is called as a plain method: `OffscreenCanvas.call(this, width || 0, height || 0);`. That is why it works with ES6 code but not when transpiled to ES5. See https://github.com/gberaudo/openlayers/tree/webworker_poc_image_shim, note that transpilation is disabled in that branch.

Regarding your initial concerns about:
- the amount of code, I have refactored the code and most of it is in the example now;
- the nested function calls does not look like a performance issue (I ran tests in chrome and Safari) but I can look more into it if you think it is an issue.

Let me know what you think about all this. I am comited to find a suitable solution.

src/ol/loadImage.js
gberaudo: @ahocevar, do you mean it is a performance penalty to have lots of code like the listeners registration in the try catch? I tested with chrome and safari using http://jsben.ch/9E3xn and I see actually the contrary, especially on safari.

src/ol/loadImage.js
ahocevar: Regarding Image shim I sent a [pull request](https://github.com/gberaudo/openlayers/pull/2) against your `webworker_poc_image_shim` branch.

Regarding try/catch, I meant that the try/catch should not be in a image loader that is used in places where no try/catch is needed. But I want you to remove the image load helper anyway, so this won't be an issue in the final pull request.

So when you switch this pull request to your `webworker_poc_image_shim` branch with my change merged, and maybe further simplify the image shim, I'll do a quick last review round and we should be good.

gberaudo: OK @ahocevar, lets do like you propose: I merged your PR and will continue from there.

PR #9161, opened 2019-01-28
Use type template for the source type of layers

fredj:
By using a generic type for the source type of a layer class (18c4c55), all the type casts for the `getSource` function can be removed (3d5a2e8).

(Thanks to @sbrunner for finding this!)

### DIFF
--- a/src/ol/layer/BaseVector.js
+++ b/src/ol/layer/BaseVector.js
@@ -61,6 +61,8 @@ const Property = {
  * property on the layer object; for example, setting `title: 'My Title'` in the
  * options means that `title` is observable, and has get/set accessors.
  *
+ * @template {import("../source/Vector.js").default|import("../source/VectorTile.js").default} VectorSourceType
+ * @extends {Layer<VectorSourceType>}
  * @api
  */
 class BaseVectorLayer extends Layer {
@@ -217,15 +219,6 @@ class BaseVectorLayer extends Layer {
   }
 
 }
-
-
-/**
- * Return the associated {@link module:ol/source/Vector vectorsource} of the layer.
- * @function
- * @return {import("../source/Vector.js").default} Source.
- * @api
- */
-BaseVectorLayer.prototype.getSource;
 
 
 export default BaseVectorLayer;
--- a/src/ol/layer/Vector.js
+++ b/src/ol/layer/Vector.js
@@ -17,6 +17,7 @@ import CanvasVectorLayerRenderer from '../renderer/canvas/VectorLayer.js';
  * property on the layer object; for example, setting `title: 'My Title'` in the
  * options means that `title` is observable, and has get/set accessors.
  *
+ * @extends {BaseVectorLayer<import("../source/Vector.js").default>}
  * @api
  */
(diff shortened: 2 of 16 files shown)


src/ol/layer/BaseVector.js
```diff
@@ -219,13 +221,4 @@ class BaseVectorLayer extends Layer {
 }
 
 
-/**
- * Return the associated {@link module:ol/source/Vector vectorsource} of the layer.
- * @function
- * @return {import("../source/Vector.js").default} Source.
- * @api
- */
-BaseVectorLayer.prototype.getSource;
```
ahocevar: Do we still get API docs for `#getSource` when this is removed?

src/ol/layer/Vector.js
```diff
@@ -17,6 +17,7 @@ import CanvasVectorLayerRenderer from '../renderer/canvas/VectorLayer.js';
  * property on the layer object; for example, setting `title: 'My Title'` in the
  * options means that `title` is observable, and has get/set accessors.
  *
+ * @extends {BaseVectorLayer<import("../source/Vector.js").default>}
```
ahocevar: Why was this and other `@extends` added? Typescript knows that from `class ... extends ...`, and the JSDoc plugin generates `@extends` information from there as well.

src/ol/layer/VectorTile.js
```diff
@@ -165,11 +166,4 @@ class VectorTileLayer extends BaseVectorLayer {
 }
 
 
-/**
- * Return the associated {@link module:ol/source/VectorTile vectortilesource} of the layer.
- * @function
- * @return {import("../source/VectorTile.js").default} Source.
- * @api
- */
-VectorTileLayer.prototype.getSource;
```
ahocevar: Same question as above. What about API docs? If we loose API docs by removing this, we should have a JSDoc plugin that brings this information back.

src/ol/layer/BaseVector.js
fredj: no, the api is broken. I will look into this

src/ol/layer/Vector.js
fredj: `@extends` is only needed to set the template type for the class. Without it, Typescript knows the parent class but not the template type

ahocevar: Ok, please make sure that the API docs can handle that too.

ahocevar: Looks like something is broken in TypeScript too here.

sbrunner: It's to give a value to the generic attribute :-)

src/ol/layer/BaseVector.js
fredj: JSDoc doesn't support having a generic type in the extends tag.
I've tried to fix it with: https://github.com/openlayers/jsdoc-plugin-typescript/pull/4

fredj: done, the jsdoc plugin will be updated in #9184

fredj:
Thanks for the review

PR #9179, opened 2019-02-04
Allow declutter with image render mode

gberaudo:
When selecting both `image` render mode and `declutter`, the mode was automatically switched to `hybrid`.

Using declutter in image render mode is legitimate: within a tile, declutter avoids symbol/text overlaps.

This PR uncouple the render mode and declutter option.


gberaudo:
Thanks @ahocevar. Indeed when we render we need to clear the declutter tree between tiles.
I updated the commit and tested locally.

changelog/upgrade-notes.md
```diff
@@ -4,6 +4,11 @@
 
 #### Backwards incompatible changes
 
+##### Allow declutter in image render mode
```
ahocevar: This is only a behavioral change, not a breaking change.

gberaudo: OK, I moved the text to the "other" section.

gberaudo:
@ahocevar, strange, that should work. Here is a  screenshot from my browser:
image

I tested with different pixel ratios and it always displayed correctly in my browser.
I also tested in Firefox and Chrome which both work.

What browser did you use? Could you test with another one? Do you have some special set up for your fonts?

gberaudo:
Hi @ahocevar, we looked into it with @fredj and here is the explanation:
- we render tile by tile:
- there is a feature (Mali) shared by 2 tiles;
- the left tile (the one displaying `Mal`) have no collision, the feature is drawn and cut at tile boundary;
- the right tile (the one displaying `Niger`) have a collision, it chose first to display `Niger` so it is rejecting `Mali`;
- `Mali` is only drawn to the left tile so appears cut.

To me this looks like the expected behaviour and to match user request.

My suggestion is to modify the examples to:
- allow switching between image and hybrid modes;
- add text explaining the difference between the 2 modes, notably this limitation.

Another approach would be to reject text/symbols not rendered 100% inside the borders of the tile.
That would eliminate completely the issue (but will create other issues).

What do you think? Have you a preference?

ahocevar:
@gberaudo Since it is probably much more effort to make this work correctly, and would also make the code harder to read, I'd be fine with what you currently have. Just make sure to update all documentation. And please no `{declutter: true, renderMode: 'image'}` in any of the existing examples, and no switches between the two modes. Just document the limitation properly.

changelog/upgrade-notes.md
```diff
@@ -120,6 +120,11 @@ The non API `getChecksum` functions of the style is also removed.
 
 #### Other changes
 
+##### Allow declutter in image render mode
+
+It is now possible to use `declutter: true` and `renderMode: 'image'` options of the vector tiles layer.
```
ahocevar:
```suggestion
It is now possible to configure vector tile layers with `declutter: true` and `renderMode: 'image'`. However, note that decluttering will be done per tile, resulting in labels and point symbols getting cut off at tile boundaries.
```
This limitation should also be added to the API docs.

PR #9190, opened 2019-02-05
Fix MultiPolygon area calculation

### LINKED ISSUE #9189: MultiPolygon#getArea() returns wrong result under certain conditions
**Describe the bug**

The following MultiPolygon leads to an incorrect area calculation:

```javascript
var multiPolygon = new ol.geom.MultiPolygon([
  [[[1, 3], [1, 2], [0, 2], [1, 3]]],
  [[[2, 1], [2, 0.5], [3, 1], [2, 1]]]
]);

multiPolygon.getArea(); // Expected: 0.75, Actual: 0.25
```

**To Reproduce**

See [codepen](https://codepen.io/anon/pen/GzMKYo?editors=1010) for full example and workaround.

**Expected behavior**

`multiPolygon.getArea()` should return 0.75.

romanzoller:
- Add failing test for `MultiPolygon#getArea`
- Rename misnamed functions in `geom/flat/orient`
  - Rename `linearRingIsOriented` => `linearRingsAreOriented`
    The function checks all linear rings of a Polygon, so the plural "rings" is more appropriate
  - Rename `linearRingsAreOriented` => `linearRingssAreOriented`
    The double s is appropriate because the check is done for all Polygons of a MultiPolygon
  - (this restores the function names from OpenLayers v4, they were changed (wrongly IMHO) in #7820.)
- Fix offset passed from linearRingssAreOriented to linearRingsAreOriented
  The offset needs to be set to the end of the previous Polygon, see `offset = orientLinearRings(...)` in function `orientLinearRingsArray`.

Fixes #9189

### DIFF
--- a/src/ol/geom/MultiPolygon.js
+++ b/src/ol/geom/MultiPolygon.js
@@ -16,7 +16,7 @@ import {deflateMultiCoordinatesArray} from './flat/deflate.js';
 import {inflateMultiCoordinatesArray} from './flat/inflate.js';
 import {getInteriorPointsOfMultiArray} from './flat/interiorpoint.js';
 import {intersectsLinearRingMultiArray} from './flat/intersectsextent.js';
-import {linearRingsAreOriented, orientLinearRingsArray} from './flat/orient.js';
+import {linearRingssAreOriented, orientLinearRingsArray} from './flat/orient.js';
 import {quantizeMultiArray} from './flat/simplify.js';
 
 /**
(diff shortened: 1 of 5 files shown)


src/ol/geom/MultiPolygon.js
```diff
@@ -251,7 +251,7 @@ class MultiPolygon extends SimpleGeometry {
   getOrientedFlatCoordinates() {
     if (this.orientedRevision_ != this.getRevision()) {
       const flatCoordinates = this.flatCoordinates;
-      if (linearRingsAreOriented(
+      if (linearRingssAreOriented(
```
pvalsecc: Why adding a second `s` at `linearRings`?

PR #9201, opened 2019-02-09
Only consider child range with drawable tiles

ahocevar:
Probably not very frequently, but it can happen that a child tile range found by the function created with `ol/layer/Renderer#createLoadedTileFinder` contains tiles that are not drawable (yet). This pull request simply changes the callback to only consider drawable tiles.


ahocevar:
@gberaudo Could be.

gberaudo:
@ahocevar, FYI, unfortunately that did not fixed it.

PR #9202, opened 2019-02-09
Fix TS-suggestions to work inside Angular

OSHistory:
The PR adds a simple suggestion to get Intellisense support inside of a newly created 
angular project. The current suggestion will result in a compilation error, because the 
angular project files placed in the `src/` folder will not be compiled anymore. 

Assuming that a lot of TS-related usage of OpenLayers will stem from angular users
trying to build an OL based app, the extra notification seems justified. 

Also note that the path inside the compiler Options does not need to be set. 

```js
    "paths": {
      "ol": ["node_modules/ol/src"],
      "ol/*": ["node_modules/ol/src/*"]
    }
```

Also using the `allowJs` option will not compile due to a conflict with the `declaration`
option. 

´´´
BUILD ERROR
error TS5053: Option 'allowJs' cannot be specified with option 'declaration'.
´´´

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,19 @@ The `ol` package contains a `src/` folder with JSDoc annotated sources. TypeScri
 }
 ```
 
+### Angular support 
+
+Assuming you are running Angular all you need to add to 
+your `tsconfig.json` is the path to the `ol/src` folder
+and your standard app-folder.
+
+```js
+  "include": [
+    "src/**/*",
+    "node_modules/ol/src/**/*.js"
+  ]
+```
+
 ## Supported Browsers
 
 OpenLayers runs on all modern browsers that support [HTML5](https://html.spec.whatwg.org/multipage/) and [ECMAScript 5](http://www.ecma-international.org/ecma-262/5.1/). This includes Chrome, Firefox, Safari and Edge. For older browsers and platforms like Internet Explorer (down to version 9) and Android 4.x, [polyfills](http://polyfill.io) for `requestAnimationFrame` and `Element.prototype.classList` are required, and using the KML format requires a polyfill for `URL`.


ahocevar:
Does this work with VS Code's auto-import feature? I mean do you incorrectly get an import like
```js
import {Map} from 'ol/src';
```
or do you get the correct
```js
import {Map} from 'ol';
```

OSHistory:
It works with the regular import, so either 

```ts
import {Map} from 'ol/Map';
```
or 
```ts
import Map from 'ol/Map';
```

ahocevar:
If what you're saying is correct, are you sure you meant `"node_modules/ol/src/**/*.js"` and not `"node_modules/ol/**/*.js"`?

OSHistory:
I was a little premature on 

```
import {Map} from 'ol/Map';
```

This import style will not work. However the following will:
 
```
import Map from 'ol/Map';
```

So maybe this should be included in the Angular section of the README.

In regards to the question of `node_modules/ol/**/*.js` vs. `node_modules/ol/src/**/*.js`
there seems to be neither a difference in the behaviour of intellisense, nor the generated output
size (I use https://github.com/OSHistory/ng7-ol5-demo to test 6.0.0-beta.1). It's a fairly small 
396 kb for the angular-application.

I figured, restricting to the `src` folder would limit the necessary file-inspection. I am not sure, 
if there is a benefit for the application from the `js.map` files included in the regular directory.

ahocevar:
> I figured, restricting to the `src` folder would limit the necessary file-inspection. I am not sure,
if there is a benefit for the application from the `js.map` files included in the regular directory.

The `src/` folder contains the sources, whereas the root folder contains files transpiled to ES5.

OSHistory:
I am aware. Thanks for the hint, though. I figured using  `node_modules/ol/**/*.js` would 
include `node_modules/ol/View.js` as well as `node_modules/ol/src/View.js`.
Intellisense support is quite brittle with either include. 

BTW: 
Using the suggested paths (current Readme) will result in a failed production build (runtime error). 

```
    "paths": {
      "ol": ["node_modules/ol/src"],
      "ol/*": ["node_modules/ol/src/*"]
    }
```

Using only `"ol": ["node_modules/ol/src"]` passes, so does omitting it alltogether. There are no differences between the builds (checked via md5sum)

ahocevar:
@OSHistory and other users of Angular and/or TypeScript, can you please report your findings on Microsoft/TypeScript#29824 instead?

ahocevar:
I have crated #9203 as an alternative. We can revisit once there has been made progress regarding Microsoft/TypeScript#29824.

OSHistory:
I was about to suggest the same thing (holding off). I do believe however, that if explicit TS-suggestions 
are to be included in the stable 6.0, they should work for angular based setups. Feels like a fair amount of issues are TS/Angular related, which I believe is unnecessary distraction from the work on this 
excellent geospatial library. Your effort in the matter is greatly appreciated, especially considering 
you are not using TS yourself.

PR #9178, opened 2019-02-03
Sources in npm package for TypeScript support

ahocevar:
This pull request adds a `src/` folder to the `ol` package, which contains the untranspiled sources. To keep the package download small, the source content is no longer included in the source maps of the transpiled files. To achieve that, I switched from `buble` to `tsc` for transpiling the module files.

The benefit of this pull request is that it is now possible to use TypeScript and IntelliSense in projects that depend on the ol package. This can be done by configuring `tsconfig.json` to include `node_modules/ol`, setting `"allowJs": true` and overriding the module paths with the `baseUrl` and `paths` config options, so they point to the `src/` directory of the package:

**Example tsconfig.json:**
```json
{
  "compilerOptions": {
    "allowJs": true,
    "checkJs": true,
    "noEmit": true,
    "baseUrl": "./",
    "paths": {
      "ol": ["node_modules/ol/src"],
      "ol/*": ["node_modules/ol/src/*"]
    }
  },
  "include": [
    "**/*.js",
    "node_modules/ol/**/*.js"
  ]
}
```
Fixes #6496
Fixes #7879
Fixes #8999 

Now the only remaining upstream issue before OpenLayers is fully TypeScript enabled is Microsoft/TypeScript#26883.

PR #9203, opened 2019-02-10
Only promise what we can deliver regarding IntelliSense

ahocevar:
This pull request updates the instructions for getting IntelliSense in VS Code, based on the discussion in #9178.

I have also created Microsoft/TypeScript#29824 in the hope that the situation can be improved for TypeScript users.


tschaub:
Thanks for all your efforts trying to make this work.

ahocevar:
Thanks for the review @tschaub. Actually I forgot to commit a local change, so I created #9204 with an addendum.

Issue #9135, opened 2019-01-22
Rendering artefacts when changing style of a tile layer


walkermatt:
**Describe the bug**

I've been working with [ECMWF](https://www.ecmwf.int/) who are moving an application to OpenLayers. The application displays weather forecast via a `TileWMS` source and  `Tile` layer; users are allowed to switch style for any given layer. With `v5.3.0` and current `master` the user sees a flash of the previous style before tiles for the current style load if they zoom in, change style then zoom out.

I believe this is caused by `interimTile`s for the previous style being displayed and `ol/source/Tile~Tile#forEachLoadedTile` not checking the `tile.key` when checking if tiles are loaded.

**To Reproduce**

* Open https://codesandbox.io/s/8yq67pv11j
* Once the tiles have loaded, zoom in
* Change to the alternative style
* Once the tiles have loaded, zoom out
    * At this point you will see a flash of the previous style before tile for the current style load

**Expected behaviour**

When a user has interacted with the map then changed style only tiles with the same key should be rendered avoiding flashes of previous styles.

This [patched example](https://wms-tiled-switch-style-gf9lsgswz.now.sh/) attempts to fix both issues and behaves as expected. I'm used a locally packaged version of the `ol` package based on `master`.

In the example I'm patching `ol/Tile~Tile#getInterimTile` to disabling `interimTile`s (for all tile sources ATM which isn't ideal but proves the point) and 
patching `ol/source/Tile~Tile#forEachLoadedTile` to ensure that only tiles with the same key as the source are considered loaded.

The source can be seen here: [src/index.js](https://zeit.co/AdxvSHVwRENbFyNqkf1Ut3DZ/wms-tiled-switch-style/gf9lsgswz/source?f=src/src/index.js)

I'd be very happy to contribute fixes for both issues. My current feeling is that there two changes:

* Implement the option to disable `interimTile`s for a given tile source
* Update `ol/source/Tile~Tile#forEachLoadedTile` as in the example to only consider a tile loaded if it's `key` matches the source key.

Would you be happy for me to open pull requests for each?

fredj:
See also #9136 (when the url is changed)

walkermatt:
@fredj the issues highlighted are improved in current `master`, when building against `v5.3.0` there is still a flash of the previous style even with `interimTile`s disabled and the change to `forEachLoadedTile`.

walkermatt:
[...]

PR #9210, opened 2019-02-12
Only check tiles with matching key in forEachLoadedTile

walkermatt:
Ensure that only tiles with the same key as the source are considered
loaded to avoid previously loaded tiles associated with another key
(different style or dimension value) being shown.

Partial fix for #9135 (1 of 2)


walkermatt:
@ahocevar moving the check to the canvas renderer's `isDrawableTile` works great and actually fixes both issues detailed in #9135.

I've updated the test map with my changes: https://wms-tiled-switch-style-ftnomtn6m.now.sh/

However it also effectively disables interim tiles, hence the test failures...

I think I understand the idea behind interim tiles (attempt to improve perceived performance by reusing previously loaded tiles) but in practice they appear to produce undesirable results in a lot of cases. Maybe interim tiles should be made optional and maybe default to disabled?

PR #9231, opened 2019-02-18
Set vector tile extent earlier

ahocevar:
We did not catch a problem with the vector tile extent (which is needed for transforming to view coordinates), because a default was returned in `ol/VectorTile#getExtent`. This pull request removes the default and lets the loader set the extent before the features, so it is ready before we transform coordinates.


ahocevar:
Note: I'm planning to simplify the vector tile transformation workflow, also to make it easier to load tiles in workers. So this code will be refactored soon, and in that process I'm going to add more tests for this.

PR #9237, opened 2019-02-18
Fix zoom after export to PDF is done

### LINKED ISSUE #9235: Export PDF examples sometimes does not reset map extend and zoom
## Description
The official [Export PDF example](https://openlayers.org/en/latest/examples/export-pdf.html) (tested with 5.3.1 - current as of time of writing) does not always reset the map to the previous extend and zoom.

## Reproduction
1. Go to the example https://openlayers.org/en/latest/examples/export-pdf.html
2. Zoom and move and then try to export a pdf.
3. Look at the map, if it is the same as before, try 2. again. Usually it takes a few zoom and move operations to see the problem and the map is not the same as before. It can be triggered in all cases after zooming and moving around enough.

Tested with chromium on Debian GNU/Linux Stretch
Architecture: amd64
Source: chromium-browser
Version: 71.0.3578.80-1~deb9u1

## Expected behavior
The map extend should always be the same after the pdf was generated.

## related issues
Search on Stackoverflow and Github issues and pull requests did not reveal a similiar report.

umbe1987:
This should fix #9235

### DIFF
--- a/examples/export-pdf.js
+++ b/examples/export-pdf.js
@@ -64,8 +64,8 @@ exportButton.addEventListener('click', function() {
       // Reset original map size
       map.setSize(size);
       map.getView().fit(extent,
-        {size: size,
-        constrainResolution: false});
+          {size: size,
+          constrainResolution: false});
       exportButton.disabled = false;
       document.body.style.cursor = 'auto';
     });
--- a/examples/export-pdf.js
+++ b/examples/export-pdf.js
@@ -64,7 +64,7 @@ exportButton.addEventListener('click', function() {
       // Reset original map size
       map.setSize(size);
       map.getView().fit(extent,
-	    {size: size,
+        {size: size,
         constrainResolution: false});
       exportButton.disabled = false;
       document.body.style.cursor = 'auto';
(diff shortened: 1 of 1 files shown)


examples/export-pdf.js
```diff
@@ -64,6 +65,7 @@ exportButton.addEventListener('click', function() {
       // Reset original map size
       map.setSize(size);
       map.getView().fit(extent, {size});
```
ahocevar: Wouldn't it be enough to just change this line to
```js
map.getView().fit(extent, {
  size: size,
  constrainResolution: false
});
```
Then you could remove the rest of the changes.

umbe1987: Totally right. Just provided your suggestion in the commit. Thanks.

ahocevar:
If you format your code like in the snippet I provided, the CI will be happy. You can also test locally by running `npm run lint`.

umbe1987:
> If you format your code like in the snippet I provided, the CI will be happy. You can also test locally by running `npm run lint`.

Just did it, thanks for making me discover lint!

ahocevar:
Thanks @umbe1987.

umbe1987:
With pleasure ;)

Issue #9243, opened 2019-02-20
Add stylus only and touch only mode in drawing a shape


huyngkh:
**What is my problem?**
I work on a project which requires to draw and annotate on a map. I have used Openlayers as the main library to fulfill the requirement. It is good to draw on a map using a PC with a mouse but it is quite frustrated to draw on a tablet device with a stylus. While I am naturally drawing a new shape using a stylus my pinky finger frequently disrupts the drawing by accidentally touching on the tablet's screen. The unintentional touching from my smallest finger is going to draw some unexpected shape and break my current drawing.

**My solution**:
I have checked and found that browsers support to distinguish a click event by a mouse, a touch or a pen while Openlayers currently support mouse only. Therefore, I just add a small patch with few lines of code to the library to support stylusOnly (or penOnly) and touchOnly which has solved my problem (to do so, I add two more variables in ol/events/condition.js). When users access my system on their tablet they can set the Openlayer map to styleOnly mode to prevent accidentally touching by hand.

Besides, I also want free-hand drawing could work with stylus only then I have changed the condition in ol/interaction/Draw.js too.

fredj:
fixed with #9244

PR #9244, opened 2019-02-20
add stylus only and touch only mode to drawing a shape

huyngkh:
**What is my problem?**
I work on a project which requires to draw and annotate on a map. I have used Openlayers as the main library to fulfill the requirement. It is good to draw on a map using a PC with a mouse but it is quite frustrated to draw on a tablet device with a stylus. While I am naturally drawing a new shape using a stylus my pinky finger frequently disrupts the drawing by accidentally touching on the tablet's screen. The unintentional touching from my smallest finger is going to draw some unexpected shape and break my current drawing.

**My solution**:
I have checked and found that browsers support to distinguish a click event by a mouse, a touch or a pen while Openlayers has currently supported mouse only. Therefore, I have just added a small patch with few lines of code to the library to support stylusOnly (or penOnly) and touchOnly which has solved my problem (to do so, I add two more variables in ol/events/condition.js). When users access to my system on their tablet they can set the Openlayer map to styleOnly mode to prevent accidentally touching by hand.

Besides, I also want free-hand drawing could work with stylus only then I have changed the condition in ol/interaction/Draw.js too.

Hope the feature is going to be accepted soon to help anyone who has the same issue. I have already created a new issue at #9243 

Thank you for reading.

### DIFF
--- a/src/ol/events/condition.js
+++ b/src/ol/events/condition.js
@@ -230,6 +230,19 @@ export const mouseOnly = function(mapBrowserEvent) {
   return pointerEvent.pointerType == 'mouse';
 };
 
+export const touchOnly = (mapBrowserEvent) => {
+  const pointerEvt = /** @type {import("../MapBrowserPointerEvent").default} */ (mapBrowserEvent).pointerEvent;
+  assert(pointerEvt !== undefined, 56); // mapBrowserEvent must originate from a pointer event
+  // see http://www.w3.org/TR/pointerevents/#widl-PointerEvent-pointerType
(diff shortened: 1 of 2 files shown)


ahocevar:
The new condition functions should be fine, but please don't use arrow functions and add a JSDoc annotation to each function.

The change in `ol/interaction/Draw` is likely to break something. So before we accept that, please try if the `freehandCondition` works for your use case:

```js
import Draw from 'ol/interaction/Draw';
import {penOnly} from 'ol/events/condition';

new Draw({
  // ...
  freehandCondition: penOnly
});

PR #9246, opened 2019-02-20
Add missing ol.css in mobile-full-screen example

fredj:
The `ol.css` CSS file is missing in the [mobile-full-screen](https://openlayers.org/en/master/examples/mobile-full-screen.html) example


fredj:
yes, it's the same for the other examples. The `mapbox-style` example already includes `ol.css`:
https://github.com/openlayers/openlayers/blob/master/examples/mapbox-style.html#L17

ahocevar:
> yes, it's the same for the other examples. The `mapbox-style` example already includes `ol.css`

The paths are different though. So maybe we need to copy ol.css over to another place?

fredj:
I've changed the path to the css file to be the same as the `mapbox-style` example. The other verbatim example (geolocation-orientation) already has the css link

PR #9250, opened 2019-02-20
Clearer behaviour of clear() and refresh() on sources

### LINKED ISSUE #2683: Reloading a vector source when URL changes
Hi,
following Bart's recommendation I open this ticket so the issue can be commented.

It starts with https://groups.google.com/forum/#!topic/ol3-dev/PjFBy5mxwgo and https://groups.google.com/forum/#!topic/ol3-dev/H6tZaFUUO4E topics on the forum.

Once a vector layer with its source pointing to some URL is created, there is no way to update the layer (really the source) content other than loading, cleaning and adding new features by hand.

Although manually loading is the most flexible option, I think it is a bit confusing. I a user loads data initially setting an `url` property, why we need to load content manually if we want update the source?

It could be easy to have a `setUrl()` that updates the `url`property and invokes the `loadFeaturesFromURL()` method.

Any opinions?

PS: I offer as voluntary to make the code changes. But I don't know when I have enough free time :)

<bountysource-plugin>

---

Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/4218070-reloading-a-vector-source-when-url-changes?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F79013&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #3600: ol.source.Vector#clear issue
Go to http://openlayers.org/en/master/examples/vector-layer.html?mode=raw and type:

``` javascript
vectorLayer.getSource().clear();
```

In the console.

All the feature should be removed but it's not the case

### LINKED ISSUE #9135: Rendering artefacts when changing style of a tile layer

ahocevar:
There has been some confusion about the differences between `clear()` and `refresh()`, especially for vector layers. This pull request tries to fix this confusion:

- On vector sources, `clear()` just removes all features, whereas `refresh()` removes all features and reloads data from the server.
- Vector sources get a new `setUrl()` method, which calls `setLoader()`. This is for convenience and makes sense with the new logic.
- Tile sources get a new `clear()` method which clears the tile cache. So when calling `clear()` before changing the tile url (e.g. by calling `updateParams()` on a WMS source, no interim tiles from the previous set of params can be used any more.

Replaces and fixes #7648.
Fixes #3600.
Fixes #2683
Will fix #9135 when the application calls `clear()` before `updateParams()` on the source.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -124,6 +124,12 @@ The removed classes and components are:
 Following the removal of the experimental WebGL renderer, the AtlasManager has been removed as well. The atlas was only used by this renderer.
 The non API `getChecksum` functions of the style is also removed.
 
+##### Change of the behavior of the vector source's clear() and refresh() methods
+
+The `ol/source/Vector#clear()` method no longer triggers a reload of the data from the server. If you were previously using `clear()` to refetch from the server, you now have to use `refresh()`.
+
+The `ol/source/Vector#refresh()` method now triggers a reload of the data from the server. If you were previously using the `refresh()` method to re-render a vector layer, you should instead call `ol/layer/Vector#changed()`.
+
 #### Other changes
 
 ##### Allow declutter in image render mode
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -128,7 +128,7 @@ The non API `getChecksum` functions of the style is also removed.
 
 The `ol/source/Vector#clear()` method no longer triggers a reload of the data from the server. If you were previously using `clear()` to refetch from the server, you now have to use `refresh()`.
 
-The `ol/source/Vector#refresh()` method now triggers a reload of the data from the server. If you were previously using the `refresh()` method to re-render a vector layer, you should instead call `ol/layer/Vector#changed()`.
+The `ol/source/Vector#refresh()` method now removes all features from the source and triggers a reload of the data from the server. If you were previously using the `refresh()` method to re-render a vector layer, you should instead call `ol/layer/Vector#changed()`.
 
 #### Other changes
 
--- a/src/ol/source/Source.js
+++ b/src/ol/source/Source.js
@@ -145,7 +145,7 @@ class Source extends BaseObject {
   }
 
   /**
-   * Refreshes the source and finally dispatches a 'change' event.
+   * Refreshes the source. Data from the server will be reloaded.
    * @api
    */
   refresh() {
--- a/src/ol/source/Source.js
+++ b/src/ol/source/Source.js
@@ -145,7 +145,7 @@ class Source extends BaseObject {
   }
 
   /**
-   * Refreshes the source. Data from the server will be reloaded.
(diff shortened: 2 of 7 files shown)


jbelien:
Hello everyone, 

Is this PR included in latest `5.3.2` release ?
I still have to `clear()` on my `VectorSource` to "refresh" the data.

Thanks.

seidlmic:
Hello, Is there some specific source event to catch when the data is completely loaded from server? Or I have to write my loader with this behaviour?

ahocevar:
@seidlmic Not really. You can listen for the generic `change` event, but that is also triggered when features are added, removed or modified. But the first time it is triggered is when the data is completely loaded from the server.

seidlmic:
@ahocevar Yes I tried this. But with v. 6.2.1 it fires after refresh() "change" event 3 times. It looks that only the last one is the right one. Data are loaded and added to layer. Can I rely on this?

ahocevar:
@seidlmic I would not count on it.

seidlmic:
Hm, I got it. So I will have to try to implement custom event. Thanks for your time.

PR #9254, opened 2019-02-21
Add tileSize option to ol/source/TileJSON

petrsloup:
The TileJSON spec does not specify the tile size and there is no TileJSON property specifying the value.
(There is possibility it will be added in some form in the future: mapbox/tilejson-spec#26, but it still does not seem to be clear.)

Many providers nowadays use 512x512 tiles with TileJSON.

This PR adds `tileSize` option to the ctor to allow this to be configured. (Instead of having to fall back to `ol/source/XYZ`.)


petrsloup:
@ahocevar I added a note about this next to the `tileSize` option -- I think that's probably the best place for people to find this easily.

PR #9264, opened 2019-02-25
add question about resizing map element

### LINKED ISSUE #9208: CSS causes wrong coordinates for mouse interaction
**Describe the bug**

In a fairly simple setup, where the map is placed in a container with constrained dimensions, the click interaction produces wrong coordinates

**To Reproduce**
This can be observed on the selection example
1. Go to https://openlayers.org/en/latest/examples/select-features.html?q=select
2. Select 'Hover' in the dropdown
1. Change the styles to narrow down the size of the topmost `div.container-fluid`. For example make it 500px wide
3. Observe how the highlighted country is off to the left
hover

**Expected behavior**
The selection should be precise when the map (its parent?) is resized

tpluscode:
Added a question and sample code which addresses problem with resizing `Map`

fixes #9208

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -21,7 +21,7 @@ Table of contents:
 * [Why aren't there any features in my source?](#why-aren-t-there-any-features-in-my-source-)
 * [How do I force a re-render of the map?](#how-do-i-force-a-re-render-of-the-map-)
 * [Why are my features not found?](#why-are-my-features-not-found-)
-
+* [Why is zooming or clicking off, inaccurate?](#user-content-why-is-zooming-or-clicking-off-inaccurate)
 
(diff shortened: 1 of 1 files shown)


doc/faq.md
```diff
@@ -371,3 +371,32 @@ const vectorLayer = new VectorLayer({
...
+import Map from 'ol/Map';
+import ResizeObserver from 'resize-observer-polyfill';
+
+const mapElement = document.querySelector('#map')
+const map = new Map({
+  target: mapElement
+})
+
+const sizeObserver = new ResizeObserver(() => {
+  this.map.updateSize()
```
fredj: should be `map.updateSize()` here

PR #9286, opened 2019-03-03
Disable transition when an interim tile is available

### LINKED ISSUE #9284: WMTS tile transitions broken
**Describe the bug**
The smooth WMTS tile transitions introduced with #5795 from #4455 are broken in v6.0.0-beta1. See  #4455 for an extensive discussion and motivation. This affects, among other things, the "WMTS Tile Transitions" example.

**To Reproduce**
1. Use npm run serve-examples
2. Visit http://localhost:8080/wmts-dimensions.html , drag the slider
3. Notice the map disappears when the dimension changes
4. Compare to the smooth behavior in the [v5.3.0 example](https://openlayers.org/en/v5.3.0/examples/wmts-dimensions.html).

**Expected behavior**
The tiles from the most-recently-acquired zoom setting should be rendered until the next tile is ready.  

I have created a [video showing the difference](https://scalgo.com/download/ol-v6.0.0-beta1-issue.mkv) between v6.0.0-beta1 and v5.3.0.

**What now?**

I understand there has been quite a lot of changes in v6.0.0 related to the internals of how layers are rendered. Does anyone have an idea about what happened, and how to fix it?

ahocevar:
This change avoids flickering when a tile that has an interim tile has finished loading.

I am not sure if this is the right fix. @tschaub, can you maybe have a look at this? Thanks!

Fixes #9284.
Fixes #9286

### DIFF
--- a/src/ol/Tile.js
+++ b/src/ol/Tile.js
@@ -170,6 +170,9 @@ class Tile extends EventTarget {
     // cleaned up by refreshInterimChain)
     do {
       if (tile.getState() == TileState.LOADED) {
+        // Show tile immediately instead of fading it in after loading, because
+        // the interim tile is in place already
+        this.transition_ = 0;
         return tile;
       }
       tile = tile.interimTile;
--- a/test/spec/ol/renderer/canvas/tilelayer.test.js
+++ b/test/spec/ol/renderer/canvas/tilelayer.test.js
@@ -39,13 +39,18 @@ describe('ol.renderer.canvas.TileLayer', function() {
       document.body.removeChild(target);
     });
 
-    it('properly handles interim tiles', function() {
+    it('properly handles interim tiles', function(done) {
       const layer = map.getLayers().item(0);
+      source.once('tileloadend', function(e) {
+        expect(e.tile.inTransition()).to.be(false);
+        done();
+      });
       source.updateParams({TIME: '1'});
       map.renderSync();
       const tiles = map.getRenderer().getLayerRenderer(layer).renderedTiles;
       expect(tiles.length).to.be(1);
       expect(tiles[0]).to.equal(tile);
+      expect(tile.inTransition()).to.be(true);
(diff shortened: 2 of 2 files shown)


thomasmoelhave:
Unfortunately the patch does not work for me. In fact I am not getting any flooding tiles at all in the WMTS Tile Transitions example (Chromium on Linux).

ahocevar:
@thomasmoelhave Maybe the tile server was down when you tried? This is what I'm seeing:

mar-04-2019 09-48-07

ThomasG77:
Just a feedback about my tests. CI url works on my side on Chromium, Firefox and Chrome on Ubuntu 18.04 x64

thomasmoelhave:
@ahocevar I hope not, we keep a close eye on our tile servers :). 

Kidding aside. I was mistaken in my previous comment. I was using Firefox on linux (v65.0 on Ubuntu 18.10), and there I am not seeing that any network requests to `ts2` at all.  

out



In Chromium it does, in fact, work.

ahocevar:
@thomasmoelhave For me Firefox 65 works. Also, the changes in this pull request do nothing that would affect tile loading. So if the CI example does not work for you on Firefox, then I guess that master won't work either. Can you check?

fredj:
@thomasmoelhave The WMTS tiles are loaded in http, that's maybe the reason they are not loading.

thomasmoelhave:
Strange. Master is also broken for me. I guess this is a separate issue then, it works fine on my colleague's Firefox 65.

thomasmoelhave:
Good point @fredj. I looked into it some more and it was my Privacy Badger that killed the requests. Sorry for the confusion.

fredj:
@tschaub Can you please take a look at this pull request?

ahocevar:
Merging this now. If @tschaub finds a better solution, we can revert this later.

PR #9289, opened 2019-03-06
Vector source geom

sbrunner:


sbrunner:
UP!

sbrunner:
Thanks @ahocevar :-)

PR #9292, opened 2019-03-06
Downgrades puppeteer to `~1.11.0`

### LINKED ISSUE #9288: Failing test (ol.control.ScaleLine)
**Tests - Failing**

Currently the `ol.control.ScaleLine` test is failing under the following requirements:
- Linux Mint 19.1 Tessa (Ubuntu 18.04)
- Node 8.11.3 or 10.15.0

```
Chrome 74.0.3723 (Linux 0.0.0) ol.control.ScaleLine projections affect the scaleline Projection's metersPerUnit affect scale for non-degree units FAILED
	Error: expected '0.000049999999999999996 nm' to equal '0.00005 nm'

```

IMHO it is a Bug in the Chrome-Version (74.0.3723) which gets [launched via puppeteer](https://github.com/openlayers/openlayers/blob/master/test/karma.config.js#L77).

Math.pow seems to be broken:

```js
// Chrome-Version (71.0.3578.98)
console.log(Math.pow(10, -5)); // 0.00001

// Chrome-Version (74.0.3723)
console.log(Math.pow(10, -5)); // 0.000009999999999999999
```

**To Reproduce**
1. Fresh checkout
2. `npm i`
3. `npm run test` or `npm run karma`

**Possible Solution**
The [current karma config](https://github.com/openlayers/openlayers/blob/master/test/karma.config.js#L77) uses the chrome binary which gets installed via puppeteer. This seems to be the canary/dev build and not the stable chrome version. I think it would be better to use the stable chrome version for the tests.
Of course this is not a bug within OpenLayers but it leads to failing tests which might be avoided.

**Expected behavior**
No test should fail.

KaiVolland:
This downgrades puppeteer to version `~1.11.0`.

Version 1.11.x will download the stable version of chromium (72.0.3582.0), while version 1.12.x downloads the beta version of chromium (73.0.3679.0).

Fixes #9288

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "pixelmatch": "^4.0.2",
     "pngjs": "^3.3.3",
     "proj4": "2.5.0",
-    "puppeteer": "^1.12.2",
+    "puppeteer": "~1.11.0",
     "serve-static": "^1.13.2",
     "shx": "^0.3.2",
     "sinon": "^7.2.3",


marcjansen:
This looks like the right change atm. we should carefully watch when greenkeeper wants to update this again.

ahocevar:
Sorry for being late to the party, but I think the issue we were seeing here with Chromium 74 is something that's there to stay, and the issue will come up again when Chromium 74 is on the stable channel. This is also why I'm not opposed to use the dev version for running tests, because it helps us catch problems before the majority of users gets exposed to them.

Floating point arithmetics can always cause issues like this, so maybe doing a `toFixed(Math.abs(exponent))` after `Math.pow(base, exponent)` for numbers lower than 1 with a negative `exponent` in the `ScaleLine` code would be appropriate?

marcjansen:
Fine for me, @KaiVolland can probably come up with a PR that addresses this. Thanks for the input.

PR #9300, opened 2019-03-07
Fix documentation about zoomDelta for ol.interaction.default function

### LINKED ISSUE #9299: Fractionnal zoom missing for mouse
Default MouseWheel Interaction always zoom by 1 level. 

The [documentation]( https://openlayers.org/en/latest/apidoc/module-ol_interaction.html#.defaults) actually mentions :
> zoomDelta - number - Zoom level delta when using keyboard or mousewheel zoom.

But after taking look at the `ol/interaction.js`, we don't use zoomDelta in constructor : 

```javascript
 const mouseWheelZoom = options.mouseWheelZoom !== undefined ?
    options.mouseWheelZoom : true;
  if (mouseWheelZoom) {
    interactions.push(new MouseWheelZoom({
      condition: options.onFocusOnly ? focus : undefined,
      constrainResolution: options.constrainResolution,
      duration: options.zoomDuration
    }));
}
```

As mentionned here #8423, i can't really change zoomFactor.

fredj:
fixes #9299

### DIFF
--- a/src/ol/interaction.js
+++ b/src/ol/interaction.js
@@ -54,8 +54,7 @@ export {default as Translate} from './interaction/Translate.js';
  * @property {boolean} [dragPan=true] Whether drag pan is desired.
  * @property {boolean} [pinchRotate=true] Whether pinch rotate is desired.
  * @property {boolean} [pinchZoom=true] Whether pinch zoom is desired.
- * @property {number} [zoomDelta] Zoom level delta when using keyboard or
- * mousewheel zoom.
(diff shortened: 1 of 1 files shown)


ahocevar:
Won't this change anyway with #9137?

jahow:
No, the `zoomDelta` value should still be used as previously for keyboard and double-click interactions (will double check just to be sure...).

Walleza1:
Is there any plans to add fractional zoom feature to MouseWheel Interaction ?

PR #9305, opened 2019-03-08
Fixes failing tests for Chrome v74

### LINKED ISSUE #9288: Failing test (ol.control.ScaleLine)

KaiVolland:
This fixes failing test (ol.control.ScaleLine) which failed in Chrome v74. See #9288.

The previous solution (#9292) is reverted by updating the puppeteer to `~1.13.0`. 
The new solution is to compensate the inaccurate calculation by using `toFixed()` [as proposed](https://github.com/openlayers/openlayers/pull/9292#issuecomment-470460362).

Fixes #9288


KaiVolland:
Ouch. Seems like i some rendering tests need to be fixed, too.

src/ol/control/ScaleLine.js
```diff
@@ -262,8 +262,10 @@ class ScaleLine extends Control {
       Math.log(this.minWidth_ * pointResolution) / Math.log(10));
     let count, width;
     while (true) {
-      count = LEADING_DIGITS[((i % 3) + 3) % 3] *
-          Math.pow(10, Math.floor(i / 3));
+      const decimalCount = Math.floor(i / 3);
+      const decimal = Number(Math.pow(10, decimalCount)
+        .toFixed(Math.abs(decimalCount)));
+      count = LEADING_DIGITS[((i % 3) + 3) % 3] * decimal;
```
ahocevar: It's better to apply `toFixed()` to the number when it gets finally rendered, instead of converting it to string and back to number, and then continue doing calculations with it.

KaiVolland: As the error/inaccuracy occurs right in `Math.pow`, wouldn't it be better to correct it right afterwards before continuing to calculate with the wrong value?

ahocevar: No, because we might be running into other floating point issues further down the line. Since we calculate this only for displaying the scale, it's best to do the formatting when displaying the scale.

KaiVolland: Ok.

KaiVolland: Resolve due to force push.

src/ol/control/ScaleLine.js
```diff
@@ -274,11 +275,13 @@ class ScaleLine extends Control {
       }
       ++i;
     }
-
     let html;
     if (this.scaleBar_) {
       html = this.createScaleBar(width, count, suffix);
     } else {
+      if (count % 1 !== 0) {
```
ahocevar: You could play this save and replace the if-block with something that also works for numbers where you don't want decimal places:
```js
html = count.toFixed(decimalCount < 1 ? -decimalCount : 0) + ' ' + suffix;
```

Issue #9291, opened 2019-03-06
Canvas elements not garbage collected on iOS 12


esseb:
**Describe the bug**
Safari on iOS 12 has a bug where it doesn't properly garbage collect canvas elements unless you explicitly give the canvas elements a width and height of `0` when they are no longer meant to be used. See https://bugs.webkit.org/show_bug.cgi?id=195325 and the reduced testcase I attached to that bug report.

I previously described this bug in https://github.com/openlayers/openlayers/issues/9166 but I'm creating a new issue now that I have more information since my bug is a bit different.

**To Reproduce**
1. Open https://dist-gkmkekevph.now.sh on an iPhone running iOS 12 (I'm using iOS 12.1.4 on an iPhone X)
2. If possible, connect the iPhone to a Mac and debug Safari on the iPhone via Safari on the Mac using devtools
3. Click the `Destroy map` button and then the `Create map` button
4. Repeat step 3 several times (it takes me about 14 destroy/create clicks before I get the error)
5. Safari either fails with the message "This web page was reloaded because a problem occured", or the map silently fails and becomes white and the console tab in devtools (if you did step 2) will show the message "Total canvas memory use exceeds the maximum limit (288 MB)."

**Expected behavior**
The map should continue to work fine no matter how many times you destroy and recreate the map in step 3.

**Further information**
The testcase uses ol@6.0.0-beta.2 and the source is available on https://github.com/esseb/openlayers-ios-canvas-memory

The testcase may make it seem like this bug is extremely obscure and unlikely to occur in the wild, but the severity of the bug depends on how developers use the map. My testcase doesn't create or hide a lot a of layers in order make it as simple as possible. I can hit the bug very quickly on https://www.yr.no/en/map/weather/1-72837/Norway/Oslo/Oslo/Oslo if I go from the "Map" tab to the "Overview" tab and back to the "Map" again a single time on an iPhone X.

Based on the reduced testcase I created for the bug report I sent to WebKit (see the description at the top) I believe OpenLayers has to explicitly give canvas elements a width and height of 0 when they are no longer necessary.

esseb:
No problems on iOS 11 (tested on browserstack, I don't know the exact OS version since they don't list it).

No problems on Chrome v72.0.3626.121 on Samsung S9.

[...]

PR #9306, opened 2019-03-08
Set the canvas size to 0 on dispose

fredj:
related to #9291


fredj:
The ones in the renderers are called when the map's is set to `null`:
https://github.com/openlayers/openlayers/blob/master/src/ol/PluggableMap.js#L993

PR #9308, opened 2019-03-10
Simplify vector tile projection handling

### LINKED ISSUE #9293: tileLoadFunction of format MVT throws error in ol 6.0.0
When creating a custom `tileLoadFunction` of `ol/source/VectorTile` in ol v6.0.0, the Projection, the Extent and the Features of the tiles have to be set in this exact order (see [documentation](https://openlayers.org/en/master/apidoc/module-ol_source_VectorTile-VectorTile.html))

According to the documentation, the extent of the MVT-tile should be set with `format.getLastExtent()`, resulting in an error, since `format.getLastExtent()`only returns the extent if the Features were read beforehand.

I have also tried to change the order to set the Features of the tile first and the extent after, which breaks all of the tiles. 

My workaround for now is calling a `format.readFeatures()` for no value, which most likely is not the intended use.

ahocevar:
This pull request moves the responsibility of reprojecting vector tile features from the renderer to the loader, which brings significant simplifications.

This is the refactoring I mentioned in https://github.com/openlayers/openlayers/pull/9231#issuecomment-464690118.

Fixes #9293.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -99,6 +99,58 @@ Due to the constraint above (layers can only be added to a single map), the over
 
 Previously, a graticule was not a layer.  Now it is.  See the graticule example for details on how to add a graticule layer to your map.
 
+##### `ol/format/Feature` API change
+
+The `getLastExtent()` method, which was required for custom `tileLoadFunction`s in `ol/source/Vector`, has been removed because it is no longer needed (see below).
+
+##### `ol/VectorTile` API changes
+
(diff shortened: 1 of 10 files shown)


fredj:
In `src/ol/VectorTile.js`, `getFeatures` function the sentence `Geometries will be in the view projection.` can now be removed from the jsdoc tag.

ahocevar:
@fredj:
> In src/ol/VectorTile.js, getFeatures function the sentence Geometries will be in the view projection. can now be removed from the jsdoc tag.

But they *are* in the view projection.

fredj:
> But they _are_ in the view projection.

yes they are! Sorry for the noise (I need more :coffee:)

PR #9315, opened 2019-03-11
Change CSS class for tooltips in measure example

### LINKED ISSUE #9290: Measure tooltip disappearing (Google Chrome update)
**Describe the bug**
When using the measure interaction in the most recent Google Chrome browser update (version 72.0.3626.121), the tooltip intermittently disappears when moving the cursor around the map.

**To Reproduce**
1. Open the [Measure OpenLayers example](https://openlayers.org/en/latest/examples/measure.html) in Google Chrome v72.0.3626.121.

**Expected behavior**
The tooltip should persist as in previous versions of Google Chrome and other web browsers. Clearly this is a result of changes to the Google Chrome browser and could affect other functionality which uses tooltips or labels.

fredj:
fixes #9290 

To avoid mixing rules with the `tooltip` CSS rules from [Bootstrap](https://getbootstrap.com/docs/3.3/javascript/#tooltips)

### DIFF
--- a/examples/measure.css
+++ b/examples/measure.css
@@ -1,4 +1,4 @@
-.tooltip {
+.ol-tooltip {
   position: relative;
   background: rgba(0, 0, 0, 0.5);
   border-radius: 4px;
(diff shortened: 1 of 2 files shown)


fredj:
Thanks for the review

PR #9320, opened 2019-03-12
enhance tile-debug readability

RobertOrthofer:
As the `ol/source/TileDebug` is used for debugging only, the usability should be top priority. Unfortunately the tile coordinates can become difficult to read at aerial imagery with high visual noise:

Screenshot 2019-03-12 at 11 51 52

This pull request tries to fix this issue by adding a canvas `strokeText` to the current text:
Screenshot 2019-03-12 at 12 03 58

Additionally, the tile coordinates are simply cut off at very high zoom levels with long coordinates:
Screenshot 2019-03-12 at 12 20 56

This issue is fixed by adding a `maxWidth` to the text. The text then becomes condensed when needed, trading a bit of aesthetics for usability:

Screenshot 2019-03-12 at 12 20 10

### DIFF
--- a/src/ol/source/TileDebug.js
+++ b/src/ol/source/TileDebug.js
@@ -55,10 +55,12 @@ class LabeledTile extends Tile {
       context.strokeRect(0.5, 0.5, tileSize[0] + 0.5, tileSize[1] + 0.5);
 
       context.fillStyle = 'grey';
+      context.strokeStyle = 'white';
       context.textAlign = 'center';
       context.textBaseline = 'middle';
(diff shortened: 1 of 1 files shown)


RobertOrthofer:
Of course, i missed the option of a broader stroke width. I updated the style and returned to the normal font weight:
Screenshot 2019-03-13 at 09 28 44

ssshem:
I want to make the TileDebug.LabeledTile.getImage asynchronous, What should I do? Thanks!

Issue #9332, opened 2019-03-14
codesandbox not working for version 6


fredj:
https://codesandbox.io/s/884vv79vxj
This is the simple example with the `ol` version changed to `6.0.0-beta.3` in `package.json`.

The map's div height is 0.

fredj:
Looks like everything in the `head` tag of the `index.html` page is ignored by codesandbox

ahocevar:
Maybe a temporary issue, because the examples used to work fine there. Can you please create a ticket at https://github.com/CompuIves/codesandbox-client?

fredj:
> Looks like everything in the `head` tag of the `index.html` page is ignored by codesandbox

That's because we're using the default codesandbox template (`create-react-app`) and this template doesn't load the head tag.

patch to come

fredj:
#9335 doesn't fix the issue, the map's height is still 0

Adding the following in the head fixes the issue.
```css
.map {
  width: 100%;
  height: 400px;
}
```

I think something changed between version 5 and 6

PR #9335, opened 2019-03-14
Create codesandbox using the 'parcel' template

fredj:
See #9332

The default template is `create-react-app` and it ignores the `head` tag defined in out `index.html`


tschaub:
Should this close #9332?

Issue #9296, opened 2019-03-06
Incorrect vector tile rendering when source tile grid and render tile grid do not align


ahocevar:
**Describe the bug**
When a vector tile source has a tile grid that does not align with the standard Web Mercator tile grid, tiles are rendered incorrectly (i.e. with gaps or overlaps)

**To Reproduce**
1. Use the `mapbox-style` example as a starting point
2. Change the code to something like
```js
import apply from 'ol-mapbox-style';
import TileGrid from '../src/ol/tilegrid/TileGrid';
import {getTopLeft} from '../src/ol/extent';

apply('map', 'https://api.mapbox.com/styles/v1/mapbox/bright-v9?access_token=[my-token]').then(function(map) {
  const resolutions = [];
  for (let i = 0; i < 36; ++i) {
    resolutions.push(78271.51696402048 / Math.pow(1.5, i));
  }
  map.getLayers().item(0).getSource().tileGrids_['EPSG:3857'] = new TileGrid({
    origin: getTopLeft(map.getView().getProjection().getExtent()),
    resolutions: resolutions
  });
});
```
3. See error

![untitled](https://user-images.githubusercontent.com/211514/53909128-41ae9580-4051-11e9-9af1-b9415651f758.png)

**Expected behavior**
Map with tiles that align properly, because the renderer was design to work with render tile grids that are completely different from source tile grids.

ahocevar:
@jahow Did you get a chance to look at this already? Let me know if you feel comfortable taking this on.

ahocevar:
@jahow Working on this now, assuming you're swamped with other work anyway since you haven't replied.

ahocevar:
This issue only occurs when `wrapX` is set to `true` on the source, and is caused by a til grid that does not align with the world.

jahow:
@ahocevar yes, sorry, I've looked into it. This is actually caused by 3 separate issues (at least):
* TileImage renderer not offsetting the tiles correctly for "mirrored worlds" (it assumes the tile grid is regular, ie only composed of full tiles)
* wrapX computation in tilegrid.js is wrong, also assuming the tile grid is regular
* Vector features are wrongly positioned by the VectorTileLayer, meaning the position of the points and labels is also wrong

Not trivial to fix but doable for sure. Will update this issue as I go

ahocevar:
Ok @jahow, I came more or less to the same conclusion, but did not observe the third issue you mentioned. As long as `wrapX` is set to false, the results are good.

[...]

PR #9336, opened 2019-03-14
Use more resolutions for vector tile rendering

ahocevar:
@jahow This is what I had in mind for #9296.

With a debug tile grid:
Mar-14-2019 17-15-50

But it has also issues with finding correct child and parent tile ranges. So it might well be that what you suggest in https://github.com/openlayers/openlayers/issues/9296#issuecomment-472919402 is a better solution. In the end, I think the solution should not bloat the code base. If that's not possible, we'll find another way to get vector tiles rendered properly with styles for fractional zoom levels.


ahocevar:
Closing in favor of #9347.

PR #9347, opened 2019-03-18
[WIP] Render vector tiles at the exact view resolution

ahocevar:
This pull request changes vector tile rendering so tiles are rendered at the exact view resolution, instead of using the resolution according to the render tile's zoom level.

With this change, we need to re-render more often, so I added a performance improvement to compensate for that: instruction and declutter group execution is bypassed if we're only rendering queued image tiles.

To avoid too frequent re-renders, the frame and view state is now created only once and updated. This way, long queued tiles will be rendered at the most recent view resolution, reducing the chance that they need to be re-rendered again.

To avoid misuse of the render frame for instruction creation, instruction creation is run in post render functions. Post render functions now allow to request another render frame by returning true.

With this change, the render quality of vector tiles is greatly improved, so noone should be missing the vector render mode that we dropped a few months ago any more.

Fixes openlayers/ol-mapbox-style#138.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -54,6 +54,7 @@ import {create as createTransform, apply as applyTransform} from './transform.js
 
 
 /**
+ * Post render function. When `true` is returned, a render frame will be requested
  * @typedef {function(PluggableMap, ?FrameState): any} PostRenderFunction
  */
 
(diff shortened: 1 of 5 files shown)


src/ol/View.js
```diff
@@ -943,7 +949,8 @@ class View extends BaseObject {
       resolution: resolution,
       rotation: rotation,
       zoom: this.getZoom()
-    };
+    });
+    return state;
```
tschaub: Not clear to me why we want to retain this state (also only reviewing on a small device).

ahocevar: The idea was that when waiting for a vector tile to be loaded, the resolution of the view may still change. So when the tile has finished loading, it is more likely that the instructions will be created for the correct resolution, instead of having to recreate them for a different resolution immediately.

ahocevar: Actually I just removed this commit. I do see the risk that came with it. The alternative is to build instruction groups on demand, instead of in a load listener. I've implemented that in the last commit.

ahocevar:
I need to look at the tests before this is ready for review again.

ahocevar:
I'm going to take another stab at this. Determining what needs to be rendered in a render frame, making the changes after rendering, and then re-rendering does not sound like a good thing to do.

ahocevar:
See #9357 for the improved replacement of this pull request.

Issue #9354, opened 2019-03-20
How to add or remove the Layer from map in ol@6.0.0-beta.3


vampire1995:
In ol@5.2.0, I used "removeLayer()" and "addLayer()" to hide or show Layer ,and everyting is ok! </br>But,in ol@6.0.0-beta.3 ,the above two methods dont work any more ,and report an error sometimes:</br>
**Uncaught TypeError: Cannot read property 'dispose' of undefined</br>
    at CompositeMapRenderer.<anonymous> (Map.js:281)
    at Map.PluggableMap.handlePostRender (PluggableMap.js:959)**

```
/**
   * @param {!import("../PluggableMap.js").FrameState} frameState Frame state.
   * @protected
   */
  scheduleRemoveUnusedLayerRenderers(frameState) {
    const layerStatesMap = getLayerStatesMap(frameState.layerStatesArray);
    for (const layerKey in this.layerRenderers_) {
      if (!(layerKey in layerStatesMap)) {
        frameState.postRenderFunctions.push(function() {
         this.removeLayerRendererByKey_(layerKey).dispose(); (Cannot read property 'dispose' of undefined)
        }.bind(this));
      }
    }
  }
}
```
</br>
I konw "layerStates" have been removed  From   [](https://github.com/openlayers/openlayers/pull/9072)  ,is it caused? And what should I do，I am a rookie about ol .........

ahocevar:
ping @fredj

fredj:
@vampire1995 what type of layer are you using ?

fredj:
The layer renderer could be undefined if it has been removed between the time the postRender function is created and executed. See #9355

vampire1995:
> The layer renderer could be undefined if it has been removed between the time the postRender function is created and executed. See #9355
Perfect!  It solved my problem： </br>
```
const layerRenderer = this.removeLayerRendererByKey_(layerKey);
          if (layerRenderer) {
            layerRenderer.dispose();
          }
```
</br>
And will it merge to  ol@6.0.0-beta.3 ?

ahocevar:
@vampire1995 Please provide a test case or a code pen or similar so we can reproduce the issue. The fix in #9355 is only a workaround for something that might be severely broken, and we want to find that out.

fredj:
I've tried to create a test case by adding and removing vector layers (with a mix of managed and unmanaged layer) but I was not able to reproduce the issue.

vampire1995:
[...]

PR #9355, opened 2019-03-20
Check if the layer renderer exists before calling dispose

fredj:
see #9354

### DIFF
--- a/src/ol/renderer/Map.js
+++ b/src/ol/renderer/Map.js
@@ -278,7 +278,10 @@ class MapRenderer extends Disposable {
     for (const layerKey in this.layerRenderers_) {
       if (!(layerKey in layerStatesMap)) {
         frameState.postRenderFunctions.push(function() {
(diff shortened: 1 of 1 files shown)


tschaub:
Is something else calling dispose on the render in this case?

ahocevar:
> Is something else calling dispose on the render in this case?

I suspect that the reason is simply a `removeLayer()` call before the layer added with `addLayer()` was rendered. The renderer won't be created before `Layer#render()` is called for the first time.

fredj:
I cannot reproduce the issue from #9354 using the minimal example: https://codesandbox.io/s/9z91jo8j5y

I propose to close this pull request

ahocevar:
Ok sounds good. I ran into a similar issue that I'm investigating right now. I can create a new pull request if it turns out that it's not an application issue.

ahocevar:
My issue is also fixed with master.

Issue #9343, opened 2019-03-18
Make zDirection controllable for image tiles


lutzhelm:
**Is your feature request related to a problem? Please describe.**
When image quality is more important than speed or low bandwidth use, it might be preferable to display tiles of the next higher resolution instead of the closest resolution if the current view resolution does not exactly match any tile source resolution. If I am not mistaken, it is currently not possible to set `zDirection` via any TileLayer or TileGrid option or via any other way.

https://github.com/openlayers/openlayers/blob/3f7c52c29f0fd667ee3356300cecfc53e88ff48a/src/ol/tilegrid/TileGrid.js#L523-L534

To demonstrate the problem, here are two screenshots of the same example of the [IIIF tile source](https://github.com/lutzhelm/openlayers/blob/iiif/src/ol/source/IIIF.js) I've been working on (see #9323), one with default `zDirection = 0` and one with `zDirection = -1`. The behavior will probably be the same for other image tile sources.

Closest resolution:
![Closest resolution](https://lutzhelm.github.io/iiif-zdirection-closestres.png)

Next higher resolution:
![Next higher resolution](https://lutzhelm.github.io/iiif-zdirection-higherres.png)

**Describe the solution you'd like**
`zDirection` could be an option of either `ol/tilegrid/TileGrid~TileGrid` or `ol/layer/Tile~TileLayer`. The latter case is probably preferable. `ol/renderer/canvas/TileLayer~CanvasTileLayerRenderer` could then initialize it's `zDirection` property according to the layer's `zDirection`.

https://github.com/openlayers/openlayers/blob/3f7c52c29f0fd667ee3356300cecfc53e88ff48a/src/ol/renderer/canvas/TileLayer.js#L64

ahocevar:
Adding a `zDirection` option to tile layers sounds reasonable. Pull request accepted.

lutzhelm:
Simply adding `zDirection` to the tile layer would have some unwanted side effects. I'm not too familiar with the OpenLayers code, but I identified some spots in the code that make me hesitant.

### `ol/tilegrid/TileGrid~TileGrid#getTileCoordForXYAndResolution`

`ol/tilegrid/TileGrid~TileGrid#getZForResolution` is also used by `ol/tilegrid/TileGrid~TileGrid#getTileCoordForXYAndResolution` via `ol/tilegrid/TileGrid~TileGrid#getTileCoordForXYAndResolution_`.
[...]

PR #9361, opened 2019-03-21
 Add a zDirection option for Zoomify - implements #9343

lutzhelm:
Via `zDirection`, the renderer's selection of a tile source's resolution can be now controlled if the view's resolution does not match any tile source resolution. Currently the `zDirection` option is only available for `Zoomify`. https://github.com/openlayers/openlayers/issues/9343#issuecomment-474341033 describes why caution is advised when using zDirection for other sources.

ol/source/Tile~TileSource#getZDirection has been added to avoid having the getZDirection method solely in the Zoomify source - just to document the possibility.

For testing I have not seen a better possibility than to test the renderer result.


src/ol/source/Zoomify.js
```diff
@@ -253,6 +257,19 @@ class Zoomify extends TileImage {
       transition: options.transition
     });
 
+    /**
+     * @protected
```
ahocevar: The renderer reads this value with every render call, which means it can be changed on the source and the change will be reflected in the rendered result. So instead of marking it `@protected` and using a getter, you can access the member directly, and get rid of the getter.

lutzhelm: You're right. Having `getZDirection` in the API doc would also be a little confusing.

PR #9362, opened 2019-03-21
Use local font "Ubuntu Regular" for rendering tests

KaiVolland:
Set local font for all rendering tests including TextStyles to enhance comparability on different systems.


ahocevar:
I could think of two reasons for the differences:

1. The fonts are not yet loaded when the image is taken. To make sure the font is loaded, you could listen to the `'clear'` event on `ol/render/canvas.labelCache`.
2. The font is not defined for 12px, so browsers scale them differently. Maybe 16px works better? Also make sure to not define a fallback font, i.e. `'16px "Ubuntu Regular"'` instead of `'12px "Ubuntu Regular", sans-serif'`.

KaiVolland:
> 2. The font is not defined for 12px, so browsers scale them differently. Maybe 16px works better? Also make sure to not define a fallback font, i.e. '16px "Ubuntu Regular"' instead of '12px "Ubuntu Regular", sans-serif'.
I tried to fix it as propsed. 

But i think the font is loaded correctly as the shape of the letters is identical.
What differs is weight, height and maybe spacing.

ahocevar:
> But i think the font is loaded correctly as the shape of the letters is identical.
What differs is weight, height and maybe spacing.

Right. It will be necessary to find the exact size and weight that the Ubuntu Regular fonts you added are defined for. So maybe `16px Ubuntu` and `italic bold 16px Ubuntu` would work as `font` property for better cross-platform results.

ahocevar:
I also noticed that the `pixelRatio` configuration on the map is missing in some of the tests. This is important when running tests on devices that have a different pixel ratio.

KaiVolland:
I finally satisfied the CI by decreasing the font sizes which leads to comparing more white pixels. Not sure if  I'm satisfied with theses results, too. But at least the test are green now.

ahocevar:
Thanks for your continued effort on this @KaiVolland. Most of the changes are good, but the 2px fonts are too small. The risk of breaking renderer changes to fall through the tolerance is too high here. Maybe going back to a larger font size for those works, in combination with also setting a line height. Something like `'normal 400 12px/1 Ubuntu'`

marcjansen:
Thanks @KaiVolland & @ahocevar for the collaborative effort! :clinking_glasses:

PR #9364, opened 2019-03-22
Replace clip.test.js with new tests

KaiVolland:
Transforms the old rendering test for layer clipping to the new rendering test approach.


rendering/cases/layer-clipping/main.js
```diff
@@ -0,0 +1,62 @@
...
+]).transform('EPSG:4326', 'EPSG:3857');
+
+const style = new Style({
+  stroke: new Stroke({
+    width: 2,
+    color: 'blue'
+  })
+});
+
+layer.on('precompose', function(event) {
```
ahocevar: There are no 'precompose' and 'postcompose' events any more. Use 'prerender' and 'postrender' instead.

ahocevar:
```suggestion
layer.on('prerender', function(event) {
```

rendering/cases/layer-clipping/main.js
```diff
@@ -0,0 +1,62 @@
...
+  context.save();
+
+  const vectorContext = event.vectorContext;
+  vectorContext.setStyle(style);
+  vectorContext.drawGeometry(geometry);
+
+  context.clip();
+});
+
+layer.on('postcompose', function(event) {
```
ahocevar:
```suggestion
layer.on('postrender', function(event) {
```

KaiVolland:
Adapted the code and force pushed. Seems to be good now.

Issue #9366, opened 2019-03-22
Unexpected animation with constrained zoom


ahocevar:
**Describe the bug**
When starting to interact with a map that has a view with a `minZoom` that's higher than the initial `zoom` on the map, an unexpected zoom-out occurs.

**To Reproduce**
1. Go to https://codesandbox.io/s/98kw3q2vro
2. Pan the map
3. See the map zoom out

The minimal view configuration to reproduce this is:
```js
new View({
  center: [0, 0],
  zoom: 2,
  minZoom: 3
})
```

**Expected behavior**
Panning the map with the above setup does not lead to an unexpected zoom-out.

ahocevar:
@jahow, can you please take a look at this?

jahow:
Will do ASAP, thanks for the repro

ahocevar:
This not only happens with the initial view configuration. It also happens every time the application sets e.g.
```js
map.getView().setZoom(2);
```

PR #9368, opened 2019-03-22
Show only one world

ahocevar:
This pull request changes the view so no more than one world is shown. To go back to the previous behavior, `multiWorld` can be set to `true` on the view.


ahocevar:
The resolution constraint is not updated when the size of the viewport changes. I'll revisit this after #9366 is resolved, because that might fix this too.

tschaub:
I’ll try this out when I get back to a full size display, but this looks different than what I would expect (and have implemented elsewhere).  I would assume that multiWorld: false would still let you pan east or west from one wold to the next, but would never show more that one works at a time.

ahocevar:
That‘s what it does.

tschaub:
Looks great.  Thanks @ahocevar.

PR #9374, opened 2019-03-25
Fix view initialization with min/max resolution constraint

### LINKED ISSUE #9366: Unexpected animation with constrained zoom

jahow:
Previously the view could be initialized with a `resolution` or `zoom` parameter outside of the allowed range (either specified by `resolutions` array or `maxResolution`, `minResolution`, `minZoom`, `maxZoom`).

This PR refactors the initialization to 1/ be simpler and 2/ take into account the allowed resolution range. A specific test was added for this, which adds for verbosity in `view.test.js` but IMO makes the expected behaviour more explicit.

This fixes #9366

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -12,7 +12,7 @@ target values are provided and constraints are applied on these to determine the
 
 ##### Removal of the `constrainResolution` option on `View.fit`, `PinchZoom`, `MouseWheelZoom` and `ol/interaction.js`
 
-The `constrainResolution` option is now only supported by the `View` class. A `View.setResolutionConstrained` method was added as well.
+The `constrainResolution` option is now only supported by the `View` class. A `View.setConstrainResolution` method was added as well.
 
 Generally, the responsibility of applying center/rotation/resolutions constraints was moved from interactions and controls to the `View` class.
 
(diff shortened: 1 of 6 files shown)


jahow:
Yes, I thought of a better approach after creating the PR which should address that. Basically the idea is to apply all constraints at the end of `applyOptions_`.

There is also an issue where constraints are not applied when the map is resized, I'll try to address this as well (otherwise the initial extent constraint is moot anyway, as the map container does not always have a size).

jahow:
@KaiVolland hi, to make this PR pass I had to remove the `extent` view parameter you used in the Zoomify rendering tests. To my understanding this was not needed anyway, could you please confirm? Thanks a lot

lutzhelm:
@jahow Confirmed, still works as intended.

Issue #9376, opened 2019-03-25
Enhancements to module exports


KaiVolland:
**Enhance exports**
Currently modules/classes can only be imported from the single file they are exported from.

Users may benefit from importing the whole library:
```js
import ol from 'ol';
const layer = new ol.layer.Tile();
```
... or a single namespace:
```js
import control from 'ol/control'; 
const OverviewMap = new control.OverviewMap();
```

>Also the naming of default exports could be improved (i.e. append the name of the containing directory') so IntelliSense finds them more easily, e.g.
```js
class OSMSource extends XYZ {}; // instead of `class OSM extends XYZ`
export default OSMSource
```

See suggestion of @ahocevar for reference: https://github.com/openlayers/openlayers/issues/9358#issuecomment-475170787

ahocevar:
From https://github.com/openlayers/openlayers/issues/9358#issuecomment-476391827:

KaiVolland:
Ok and as `import control from 'ol/control'; ` could be written as `import * as control from 'ol/control'` the exports are fine as is?

Renaming the classes is the remaining TODO here right?

ahocevar:
> the exports are fine as is?

A quick sanity check could not hurt, to make sure that all exports that are marked `@api` are re-exported.

> Renaming the classes is the remaining TODO here right?

Right.

KaiVolland:
@ahocevar should the classname be changed too? Or only the default export?


```js
// export
class BingMapsSource extends TileImage
[...]
export default BingMapsSource;

// import
import {BingMapsSource} from '../src/ol/source.js';
import BingMapsSource from '../src/ol/source/BingMaps.js';
```
vs
```js
// export
[...]
const BingMapsSource = BingMaps;
export default BingMapsSource;

// import
import {BingMaps} from '../src/ol/source.js';
import BingMapsSource from '../src/ol/source/BingMaps.js';
```

Renaming the default export name is just relevant for intellisense, while renaming the whole class will have a big impact on existing code.

ahocevar:
Only the default export. But the change is simpler than what you have above:
```diff
diff --git a/src/ol/source/BingMaps.js b/src/ol/source/BingMaps.js
index 3d4ea68ab..24abe7a74 100644
--- a/src/ol/source/BingMaps.js
+++ b/src/ol/source/BingMaps.js
@@ -117,7 +117,7 @@ const TOS_ATTRIBUTION = '<a class="ol-attribution-bing-tos" ' +
[...]

PR #9381, opened 2019-03-26
[WIP] Renames class names to be more explicit

KaiVolland:
Changes the classnames to be more explicit.

This should be helpful when using intellisense or autmatic import functionality.

Addresses #9376

### DIFF
--- a/src/ol/control/Attribution.js
+++ b/src/ol/control/Attribution.js
@@ -43,7 +43,7 @@ import {visibleAtResolution} from '../layer/Layer.js';
  *
  * @api
  */
-class Attribution extends Control {
+class AttributionControl extends Control {
 
   /**
    * @param {Options=} opt_options Attribution options.
@@ -326,7 +326,7 @@ class Attribution extends Control {
 /**
  * Update the attribution element.
  * @param {import("../MapEvent.js").default} mapEvent Map event.
- * @this {Attribution}
+ * @this {AttributionControl}
  * @api
  */
 export function render(mapEvent) {
@@ -334,4 +334,4 @@ export function render(mapEvent) {
 }
 
 
-export default Attribution;
+export default AttributionControl;
--- a/src/ol/control/FullScreen.js
+++ b/src/ol/control/FullScreen.js
@@ -40,7 +40,7 @@ const events = ['fullscreenchange', 'webkitfullscreenchange', 'MSFullscreenChang
  *
  * @api
  */
-class FullScreen extends Control {
+class FullScreenControl extends Control {
 
   /**
    * @param {Options=} opt_options Options.
@@ -262,4 +262,4 @@ function exitFullScreen() {
   }
 }
 
-export default FullScreen;
+export default FullScreenControl;
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -49,7 +49,7 @@ const COORDINATE_FORMAT = 'coordinateFormat';
  *
  * @api
  */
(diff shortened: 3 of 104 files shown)


tschaub:
I’m not clear on the justification for this.  For consistency, are we also planning to rename the exports from modules like ol/style.js?  If so, this is a breaking change and should have a clear justification.

ahocevar:
> I’m not clear on the justification for this

An example would be to avoid IntelliSense trying to do this:
```js
import VectorTile from `ol/source/VectorTile`;
import VectorTile /* ! */ from `ol/layer/VectorTile`;
```

> For consistency, are we also planning to rename the exports from modules like ol/style.js?

I'd be hesitant to do this, but I agree it would be good for consistency.

tschaub:
Maybe we have a different setup or workflow.  Here is how things currently work for me:
import

I don't have a problem with this, which is why I'm having a hard time understanding the problem that motivates this change.

This is personal preference, but I also like the readability of this (and feel that the longer names are clunky):
```js
import {Fill, Stroke, Style} from 'ol/style';
```

We're also talking about breaking everything like this (which I think would be unfortunate):
```js
import Map from 'ol/Map';
import {Tile} from 'ol/layer';
import {OSM, XYZ} from 'ol/source';

const url = 'https://example.com/{z}/{x}/{y}';

new Map({
  layers: [
    new Tile({source: new OSM()}),
    new Tile({source: new XYZ({url})})
  ]
});
```

tschaub:
Seeing "IntelliSense auto import", I'm wondering if you are talking about a problem with a workflow using a specific plugin for VS Code.  I don't use an auto-import plugin, but I tried installing the two most popular results here https://marketplace.visualstudio.com/search?term=auto%20import&target=VSCode&category=All%20categories&sortBy=Relevance and couldn't get either to auto-import modules from `node_modules`.  Are you talking about a problem with one of these plugins?  Do you have a specific configuration?

Is the problem with

ahocevar:
Ok. Sorry for the noise then. I‘m not using any plugins, just typing e.g. `new VectorTile(/* Auto-Import works */{source: new VectorTile( /* auto-import does not work */`.

@KaiVolland and @tschaub, sorry for wasting your time, looks like this change does not improve anything.

ahocevar:
> We're also talking about breaking everything like this (which I think would be unfortunate):

None of those would break though.

tschaub:
I appreciate the work you've done here @KaiVolland and definitely agree with @ahocevar's intent of making the library easier to use (with nice tooling like auto-import).

Right now we have some consistency in these things:

 * our file names (e.g. `ol/source/OSM.js`)
 * our default export names (`OSM`)
 * our named exports from "package" modules (e.g. `OSM` from `ol/source.js`)

I think this adds a bit of niceness to the dev story because things are guessable (if I see `ol/source/OSM.js` I can guess that I can `import {OSM} from 'ol/source'` and if I see an import named `OSM` I can guess that I can find a file named `OSM.js`).

This change looks like it is just about the default export names.  Which I agree doesn't break the API  and it sounds like it might fix something.  If making our default export names unique makes auto-import work for application developers, that sounds nice.  I can't reproduce the problem or verify the fix right now, but I don't doubt it.

If we make this change, I imagine the next suggestion (or one in the future) might be to regain the consistency we have now (between default export name, file name, and convenience named exports).  Maybe this won't happen, or maybe we won't do it, or maybe we will and it isn't a big deal (either way, slippery slope fallacy).

How about we leave this up here and I work on getting auto-imports working (I'd like that anyway)?

PR #9385, opened 2019-03-28
PointsLayer added clear of the render Buffers on Source Change

Kai-W:
The Renderbuffers don't got cleared if the dataScource changes. The geometry of the  features gets pushed in the back of the render Buffer simly ccreating a larger Buffer with old and new state.
The proposed fix clears the Buffers before adding the changed Features.


ahocevar:
ping @jahow

PR #9394, opened 2019-04-03
Introduces import snippets for static functions

KaiVolland:
Adds snippets for import of modules and their exported functions to the API documentation.

Example output for `ol/sphere`:

127 0 0 1_8080_apidoc_module-ol_sphere html


marcjansen:
Thanks for this, @KaiVolland.

PR #9398, opened 2019-04-04
Introduces getGetLegendGraphic method

### LINKED ISSUE #9163: GetLegend requests
SLD enabled WMS servers can support GetLegendGraphic requests. I will need this for a project. If desired I can implement a function similar to the getGetFeatureInfoUrl of the ImageWMS class for openlayers. That would be a getGetLegendGraphicUrl

Some more information on the topic can be found here: http://suite.opengeo.org/ee/docs/4.5/geoserver/services/wms/get_legend_graphic/legendgraphic.html and in the sld specification chapter 10: http://portal.opengeospatial.org/files/?artifact_id=22364

KaiVolland:
Adds the `getGetLegendGraphicUrl` method to `ImageWMS` and `TileWMS` source.
Also adds an example and corresponding tests.

closes #9163

### DIFF
--- a/src/ol/source/ImageWMS.js
+++ b/src/ol/source/ImageWMS.js
@@ -192,6 +192,40 @@ class ImageWMS extends ImageSource {
   }
 
   /**
+   * Return the GetLegendGraphic URL for the passed resolution.
+   * Return `undefined` if the GetLegendGraphic URL cannot be constructed.
+   * @param {number} resolution Resolution.
+   * @param {!Object} params GetLegendGraphic params. Default `FORMAT` is
+   *     `image/png`. `VERSION` should not be specified here.
+   * @return {string|undefined} GetLegendGraphic URL.
+   * @api
+   */
+  getGetLegendGraphicUrl(resolution, params) {
+    const layers = this.params_.LAYERS;
+    const isSingleLayer = !Array.isArray(layers) || this.params_['LAYERS'].length === 1;
+    if (this.url_ === undefined || !isSingleLayer) {
+      return undefined;
+    }
+
+    const mpu = this.getProjection() ? this.getProjection().getMetersPerUnit() : 1;
+    const dpi = 25.4 / 0.28;
+    const inchesPerMeter = 39.37;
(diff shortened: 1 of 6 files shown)


src/ol/source/ImageWMS.js
```diff
@@ -191,6 +191,41 @@ class ImageWMS extends ImageSource {
...
+    const isSingleLayer = !Array.isArray(layers) || this.params_['LAYERS'].length === 1;
+    if (this.url_ === undefined || !isSingleLayer) {
+      return undefined;
+    }
+
+    const units = this.getProjection() && this.getProjection().getUnits();
+    const dpi = 25.4 / 0.28;
+    const mpu = METERS_PER_UNIT[units || 'm'];
+    const inchesPerMeter = 39.37;
+    const scale = resolution * mpu * inchesPerMeter * dpi;
```
KaiVolland: I'm not sure about the transformation of `resolution` to scale.
Couldn't find a code block where this was solved already and i think the pixelratio might be relevant here, too.
Advice appreciated.

KaiVolland: Thx. I adapted the code a little to use `getMetersPerUnit`.

simonseyock: But there is a problem: you actually need the projection of the view not the projection of the source as the resolution is tied to the view projection.

KaiVolland: Oh of course. Hmm. Maybe i should change the parameters to pass a view instead of a resolution.
This way i can get the resolution and view-projection at once.

simonseyock: I f you compare with https://github.com/openlayers/openlayers/blob/master/src/ol/source/ImageWMS.js#L159 you see that normally you would pass the resolution and projection as independent parameters. I think that is for greater usability in a context where you might have the needed parameters but not the view and more general less dependencies.

simonseyock: also scale is not a mandatory parameter for legendgraphic requests. I don't know if it should be possible to call the method without these parameters? What do you think?

PR #9399, opened 2019-04-04
Update constraints when setting center, zoom or rotation

### LINKED ISSUE #9396: Unexpected animation when starting to interact with the map
**Describe the bug**
This issue is similar to #9366. When starting to interact with a map that has a view whose zoom and center were set with `View#setZoom()` and `View#setCenter()` or `View#animate()` and are outside the restricted range, an unexpected zoom-out occurs.

**To Reproduce**
1. Go to https://codesandbox.io/s/3n2o1n5qm (uses `View#setZoom()` and `View#setCenter()` or https://codesandbox.io/s/o51l6l6qrz (uses `View#animate()`)
2. Pan the map
3. See the map zoom out

The minimal map configuration to reproduce this is:
```js
const map = new Map({
  layers: [new TileLayer({ source: new OSM() })],
  target: 'map'
});
const view = map.getView();
view.animate({
  center: [0, 0],
  zoom: 0
});
// or, instead of view.animate:
//view.setCenter([0, 0]),
//view.setZoom(0);
```

The same happens when the map is initially already configured with a view with `zoom` and `center`.

**Expected behavior**

ahocevar:
Fixes #9396.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -378,7 +378,6 @@ class View extends BaseObject {
     } else if (options.zoom !== undefined) {
       this.setZoom(options.zoom);
     }
-    this.resolveConstraints(0);
 
     this.setProperties(properties);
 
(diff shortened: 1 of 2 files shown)


ahocevar:
Closing in favor of #9404.

PR #9400, opened 2019-04-04
Changes to parsing of KML displayName

### LINKED ISSUE #9265: KML readFeatures incorrectly parses displayName
**Describe the bug**
If I try to parse/read kml file

    <kml xmlns="http://www.opengis.net/kml/2.2">
    <Document>
      <name>Kml Attribute Sample 1: Simple</name>
      <Placemark>
        <name>Some Point</name>
        <ExtendedData>
          <Data name="myNumber">
            <value>123</value>
          </Data>
          <Data name="myString">
		<displayName>Alias to myString</displayName>
            <value>the string value</value>
          </Data>
        </ExtendedData>
        <Point>
            <coordinates>16.334122,48.163808</coordinates>
        </Point>
      </Placemark>
        </Document>
    </kml>

 with 

        const features = new KML().readFeatures(data, {featureProjection: 'EPSG:3857', extractStyles: true});
       
it does not parse the displayName correctly

**To Reproduce**
1. Use the example from 'https://openlayers.org/en/latest/examples/kml.html'
2. use the KML file (in attachment) and call: 
    "const features = new KML().readFeatures(data, {featureProjection: 'EPSG:3857', extractStyles: true});"
3. readFeatures returns: 
image

**Expected behavior**
DisplayName is parsed/read as key for property like in google maps -
If you drag on drop KML file on google maps, it displays

image

and I expect OL to parse it the same way

KaiVolland:
To avoid loss of information the behaviour of parsing the displayName (if a name attribute is also given) is changed.

```xml
<kml xmlns="http://earth.google.com/kml/2.2">
  <Placemark xmlns="http://earth.google.com/kml/2.2">
    <ExtendedData>
      <Data>
        <displayName>capital</displayName>
        <value>London</value>
      </Data>
      <Data name="country">
        <displayName>Country</displayName>
        <value>United-Kingdom</value>
      </Data>
    </ExtendedData>
   </Placemark>
</kml>;
```

Previous behaviour was:
```js
feature.get('capital'); //'London'; 
feature.get('country'); //'United-Kingdom';
```

New behaviour is:
```js
feature.get('capital'); //'London'; 
feature.get('country'); //{value: 'United-Kingdom', displayName: 'Country'}
```

closes #9265

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -682,9 +682,9 @@ class KML extends XMLFeature {
       const localName = n.localName;
       if (includes(NAMESPACE_URIS, n.namespaceURI) &&
           (localName == 'Document' ||
-           localName == 'Folder' ||
-           localName == 'Placemark' ||
-           localName == 'kml')) {
+            localName == 'Folder' ||
+            localName == 'Placemark' ||
+            localName == 'kml')) {
         const name = this.readNameFromNode(n);
         if (name) {
           return name;
@@ -748,8 +748,8 @@ class KML extends XMLFeature {
       const localName = n.localName;
       if (includes(NAMESPACE_URIS, n.namespaceURI) &&
           (localName == 'Document' ||
-           localName == 'Folder' ||
-           localName == 'kml')) {
+            localName == 'Folder' ||
+            localName == 'kml')) {
         extend(networkLinks, this.readNetworkLinksFromNode(n));
       }
     }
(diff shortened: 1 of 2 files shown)


ahocevar:
Not a review, just a note that this is a breaking change and needs to be explained in the upgrade notes.

marcjansen:
Since this does only need a proper review and upgrade notes, I'd find it sad if this got closed without a merge.

marcjansen:
Again reopening.

ahocevar:
I took the liberty to add a commit that makes this change a non-breaking change. @marcjansen, do you also want to give this a final look before I merge?

KaiVolland:
@marcjansen could you have a quick look at this?

marcjansen:
Thx for the collaboration here.

PR #9404, opened 2019-04-05
View / apply constraints when an interaction starts

### LINKED ISSUE #9396: Unexpected animation when starting to interact with the map

jahow:
Previously, an interaction could begin while target values (center/resolution) were out of the allowed range, causing a glitch where the view zoom/position would jump suddenly.

@ahocevar I've reused your test from #9399 & expanded on it.

Fixes #9396.

### DIFF
--- a/src/ol/control/Zoom.js
+++ b/src/ol/control/Zoom.js
@@ -114,7 +114,6 @@ class Zoom extends Control {
       // upon it
       return;
     }
-    view.resolveConstraints(0);
     const currentZoom = view.getZoom();
     if (currentZoom !== undefined) {
       const newZoom = view.getConstrainedZoom(currentZoom + delta);
--- a/src/ol/control/Zoom.js
+++ b/src/ol/control/Zoom.js
@@ -114,6 +114,7 @@ class Zoom extends Control {
       // upon it
       return;
     }
+    view.resolveConstraints(0);
     const currentZoom = view.getZoom();
     if (currentZoom !== undefined) {
       const newZoom = view.getConstrainedZoom(currentZoom + delta);
--- a/src/ol/interaction/Interaction.js
+++ b/src/ol/interaction/Interaction.js
@@ -135,6 +135,7 @@ export function zoomByDelta(view, delta, opt_anchor, opt_duration) {
   const newZoom = view.getConstrainedZoom(currentZoom + delta);
   const newResolution = view.getResolutionForZoom(newZoom);
 
+  view.resolveConstraints(0);
   if (view.getAnimating()) {
     view.cancelAnimations();
   }
--- a/src/ol/interaction/Interaction.js
+++ b/src/ol/interaction/Interaction.js
@@ -111,7 +111,7 @@ export function pan(view, delta, opt_duration) {
   const currentCenter = view.getCenter();
   if (currentCenter) {
     const center = [currentCenter[0] + delta[0], currentCenter[1] + delta[1]];
-    view.animate({
+    view.animate_({
       duration: opt_duration !== undefined ? opt_duration : 250,
       easing: linear,
       center: view.getConstrainedCenter(center)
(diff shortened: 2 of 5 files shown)


src/ol/control/Zoom.js
```diff
@@ -114,6 +114,7 @@ class Zoom extends Control {
       // upon it
       return;
     }
+    view.resolveConstraints(0);
```
ahocevar: I'm wondering if a better place for this new line this would be ´ol/View#animate()`, before the last line.

src/ol/interaction/Interaction.js
```diff
@@ -135,6 +135,7 @@ export function zoomByDelta(view, delta, opt_anchor, opt_duration) {
   const newZoom = view.getConstrainedZoom(currentZoom + delta);
   const newResolution = view.getResolutionForZoom(newZoom);
 
+  view.resolveConstraints(0);
```
ahocevar: Is this needed? Or will it still be needed if `ol/View#animate()` handles it (like I suggested in the previous comment)?

test/spec/ol/view.test.js
```diff
@@ -1801,6 +1801,52 @@ describe('ol.View', function() {
...
+  beforeEach(function() {
+    map = new Map({
+      target: createMapDiv(512, 256)
+    });
+  });
+  afterEach(function() {
+    disposeMap(map);
+  });
+
+  it.only('works when initialized with #setCenter() and #setZoom()', function(done) {
```
ahocevar: The `.only` should be removed.

test/spec/ol/view.test.js
```diff
@@ -1801,6 +1801,52 @@ describe('ol.View', function() {
...
+    view.setZoom(0);
+    view.beginInteraction();
+    view.endInteraction();
+    setTimeout(function() {
+      expect(callCount).to.be(1);
+      done();
+    }, 500);
+  });
+
+  it.only('works when initialized with #animate()', function(done) {
```
ahocevar: `.only` also should be removed here.

src/ol/control/Zoom.js
jahow: I've thought about that as well but didn't manage to get it to work: it was either breaking the "smoothly return view in allowed extent/resolution" effect at the end of an interaction, or causing regressions in other places, depending on where I'd put it.

I agree this does not feel right. To have a cleaner system would mean some rework of the `animate` method IMO, which is a tricky task. The ideal solution would be to get rid of all these `applyTargetState_` and `resolveConstraints` calls and only translate target values into actual view parameters on the main render loop (ans as such apply constraints only once per frame).

test/spec/ol/view.test.js
jahow: fixed

src/ol/interaction/Interaction.js
jahow: See other comment

jahow:
TL;DR: could not find a better way with the time & means available to me, sorry!

ahocevar:
@jahow I finally wrapped my head around the new constraints system, and I think I came up with a good solution. Can you please let me know what you think about 070c1ec?

jahow:
Works fine! well done

PR #9418, opened 2019-04-09
Re-export GeometryCollection in ol/geom

ahocevar:
This export was currently missing in `ol/geom`.

### DIFF
--- a/src/ol/geom.js
+++ b/src/ol/geom.js
@@ -4,6 +4,7 @@
 
 export {default as Circle} from './geom/Circle.js';
 export {default as Geometry} from './geom/Geometry.js';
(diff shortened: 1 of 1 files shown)


src/ol/geom.js
```diff
@@ -4,6 +4,7 @@
 
 export {default as Circle} from './geom/Circle.js';
 export {default as Geometry} from './geom/Geometry.js';
+export {default as GeometryCollection} from './geom/GeometryCollection';
```
fredj:
```suggestion
export {default as GeometryCollection} from './geom/GeometryCollection.js';
```
Adding `.js` like the others exports

PR #9430, opened 2019-04-12
Adds IIIF Image API tile source

### LINKED ISSUE #9323: IIIF Image API tile source
For the digital library framework [Kitodo.Presentation](https://github.com/kitodo/kitodo-presentation) that uses OpenLayers to display tiled images from IIP and Zoomify servers as well as static images, I need an additional tile source for [International Image Interoperability Framework](https://iiif.io/) (IIIF) [Image API](https://iiif.io/api/image/) services. Kitodo.Presentation is not only used by several scientific libraries in Germany, it's also the basis for the [DFG Viewer](https://github.com/slub/dfg-viewer), the reference implementation of the German Research Foundation's (DFG) digitization guidelines.

IIIF is a set of API specifications that are primarily implemented and used by GLAM institutions (galleries, libraries, archives, museums and others; see list of active [IIIF community](https://iiif.io/community/#participating-institutions) member institutions) to present digital representations of works such as books, newspapers, scrolls, maps, manuscripts etc. (or other digital images) primarily to scholars and researchers. The IIIF APIs in detail are currently the Image API, Presentation API, Content Search API and Authentication API:
* The [IIIF Presentation API](https://iiif.io/api/presentation/) describes ways to present a digital work to a user, regarding the arrangement of (multiple) images and audio/visual resources, comments, transscriptions, descriptions etc.
* The [IIIF Image API](https://iiif.io/api/image/) describes how to serve/retrieve full or tiled images or specific parts of an image at different scale levels.
* The [IIIF Content Search API](https://iiif.io/api/search/) enables lookup of IIIF resources.
* The [IIIF Authentication API](https://iiif.io/api/auth/) describes how make use of existing authentication services to authenticate users of IIIF services.

In the context of this issue, the API of interest is the IIIF Image API. Support for IIIF would offer an additional way to use custom image tile sources besides IIP and Zoomify. [IIPSrv](https://github.com/ruven/iipsrv), the server side implementation of [IIP](https://iipimage.sourceforge.io/) which is already supported in OpenLayers via the Zoomify tile source does already offer IIIF image support, and there are more IIIF server implementations around.

While implementing an OpenLayers IIIF tile source could be done without changes to OpenLayers itself via a separate project, it would be great to have the support of the large OpenLayers community.

With https://github.com/klokantech/iiifviewer there is already a great IIIF implementation for OpenLayers 4, but it is no longer maintained, and it has several problems (a resolution / zoom level matching bug klokantech/iiifviewer#23, no support for [IIIF level 0 compliant](https://iiif.io/api/image/2.1/compliance/#level-0-compliance) image services, no support for non square tiles, does not use [canonical](https://iiif.io/api/image/2.1/#canonical-uri-syntax) URL syntax).

I already started [an implementation](https://github.com/lutzhelm/openlayers/blob/iiif/src/ol/source/IIIF.js) which is kind of based on the Zoomify tile source. The linked [branch](https://github.com/lutzhelm/openlayers/tree/iiif) ("iiif") comes with a [working example](https://github.com/lutzhelm/openlayers/blob/iiif/examples/iiif.html). There is still a lot to do (tests, documentation, getting around [rounding issues](https://www.jack-reed.com/2016/10/14/rounding-strategies-used-in-iiif.html)), but it should already support any Image API 1 and 2 services on any compliance level.

I would continue to work on that source and would appreciate any input / feedback.

 @hyperknot mentioned porting the OpenSeaDragon implementation to OpenLayers in #7795. I don't know if they finished this work but it would be an alternative to my suggestion.

lutzhelm:
Introduces a [IIIF Image API](https://iiif.io/api/image/) tile source with a parser for [IIIF image information responses](https://iiif.io/api/image/2.1/#image-information) and a working example. Closes #9323. More context about IIIF can be found in the [original issue](https://github.com/openlayers/openlayers/issues/9323).

This introduces:
* the image tile source `ol/source/IIIF~IIIF` that extends `ol/source/TileImage~TileImage`
* `ol/format/IIIFInfo~ImageInfo` to parse IIIF image information responses of any version into more Openlayers-like options for the tile source
* testcases for both classes
* an example that let's you display any IIIF image
* two new error codes

The IIIF tile source
* supports versions 1.0 through 2.1.1 and should already be working with services of the not yet released [version 3.0](https://iiif.io/api/image/3.0/)
* uses canonical URIs for image requests if supported
* supports non square tiles
* supports region calculation by absolute coordinates and by percentages, size calculation by width, by height, by width and height and by percentage
* supports [IIIF Compliance Level 0](https://iiif.io/api/image/2.1/compliance/#level-0-compliance), i.e. image services with pregenerated tiles and/or scaled full images

What's left to be done (later, partly):
* ~~:exclamation: For version 2, `IIIFInfo` adds `license` and `attribution` to the options' `attributions`. This allows introducing HTML content and it should be sanitized before a merge - or be removed. Sanitizing would probably require introducing a new dependency for that purpose which I am hesitant to do.~~
https://github.com/lutzhelm/openlayers/blob/1ebce1de128fca64dec1cab1404da79bae6aefcf/src/ol/format/IIIFInfo.js#L134-L145
* I'm not sure wether to keep `reprojectionErrorThreshold` and `projection` in the tile source options.
* Neither `ol/source/IIIF~IIIF` nor `ol/format/IIIFInfo~ImageInfo` do currently care about maximum width, maximum height or maximum area that might be given in the image information response. In real world applications this should mostly matter for full image requests, not for tiled images. This should not stop a merge.
* Version 3 functionality needs completion and more testing once the API is final.

### DIFF
--- a/examples/iiif.js
+++ b/examples/iiif.js
@@ -8,9 +8,10 @@ const layer = new TileLayer(),
     map = new Map({
       layers: [layer],
       target: 'map'
-    });
-
-const notifyDiv = document.getElementById('iiif-notification');
+    }),
+    notifyDiv = document.getElementById('iiif-notification'),
+    urlInput = document.getElementById('imageInfoUrl'),
+    displayButton = document.getElementById('display');
 
 function refreshMap(imageInfoUrl) {
   fetch(imageInfoUrl).then(function(response) {
@@ -38,12 +39,8 @@ function refreshMap(imageInfoUrl) {
   });
 }
 
-const urlInput = document.getElementById('imageInfoUrl');
-const displayButton = document.getElementById('display');
-
 displayButton.addEventListener('click', function() {
-  const imageInfoUrl = urlInput.value;
-  refreshMap(imageInfoUrl);
+  refreshMap(urlInput.value);
 });
 
 refreshMap(urlInput.value);
--- a/examples/iiif.js
+++ b/examples/iiif.js
@@ -1,0 +1,48 @@
+import Map from '../src/ol/Map.js';
+import View from '../src/ol/View.js';
+import TileLayer from '../src/ol/layer/Tile.js';
+import IIIF from '../src/ol/source/IIIF.js';
+import IIIFInfo from '../src/ol/format/IIIFInfo.js';
+
+const layer = new TileLayer(),
+    map = new Map({
+      layers: [layer],
+      target: 'map'
+    });
+
+const notifyDiv = document.getElementById('iiif-notification');
+
+function refreshMap(imageInfoUrl) {
+  fetch(imageInfoUrl).then(function(response) {
+    response.json().then(function(imageInfo) {
+      const options = new IIIFInfo().readFromJson(imageInfo);
+      if (options === undefined || options.version === undefined) {
+        notifyDiv.textContent = 'Data seems to be no valid IIIF image information.';
+        return;
+      }
+      const extent = [0, -options.size[1], options.size[0], 0];
+      layer.setSource(new IIIF(options));
+      map.setView(new View({
+        resolutions: layer.getSource().getTileGrid().getResolutions(),
+        extent: extent,
+        constrainOnlyCenter: true
+      }));
+      map.getView().fit(extent);
+      notifyDiv.textContent = '';
+    }).catch(function(body) {
+      notifyDiv.textContent = 'Could not read image info json. ' + body;
+    });
+  }).catch(function() {
+    notifyDiv.textContent = 'Could not read data from URL.';
+  });
+}
+
+const urlInput = document.getElementById('imageInfoUrl');
+const displayButton = document.getElementById('display');
+
+displayButton.addEventListener('click', function() {
+  const imageInfoUrl = urlInput.value;
+  refreshMap(imageInfoUrl);
+});
+
+refreshMap(urlInput.value);
--- a/examples/iiif.js
+++ b/examples/iiif.js
@@ -16,7 +16,7 @@ const layer = new TileLayer(),
 function refreshMap(imageInfoUrl) {
   fetch(imageInfoUrl).then(function(response) {
     response.json().then(function(imageInfo) {
-      const options = new IIIFInfo().readFromJson(imageInfo);
+      const options = new IIIFInfo().readOptionsFromJson(imageInfo);
       if (options === undefined || options.version === undefined) {
         notifyDiv.textContent = 'Data seems to be no valid IIIF image information.';
         return;
--- a/examples/iiif.js
+++ b/examples/iiif.js
@@ -21,15 +21,15 @@ function refreshMap(imageInfoUrl) {
         notifyDiv.textContent = 'Data seems to be no valid IIIF image information.';
         return;
       }
-      const extent = [0, -options.size[1], options.size[0], 0];
       options.zDirection = -1;
-      layer.setSource(new IIIF(options));
+      const iiifTileSource = new IIIF(options);
+      layer.setSource(iiifTileSource);
       map.setView(new View({
-        resolutions: layer.getSource().getTileGrid().getResolutions(),
-        extent: extent,
+        resolutions: iiifTileSource.getTileGrid().getResolutions(),
+        extent: iiifTileSource.getTileGrid().getExtent(),
         constrainOnlyCenter: true
       }));
-      map.getView().fit(extent);
+      map.getView().fit(iiifTileSource.getTileGrid().getExtent());
       notifyDiv.textContent = '';
     }).catch(function(body) {
       notifyDiv.textContent = 'Could not read image info json. ' + body;
--- a/examples/iiif.js
+++ b/examples/iiif.js
@@ -21,6 +21,7 @@ function refreshMap(imageInfoUrl) {
         return;
       }
       const extent = [0, -options.size[1], options.size[0], 0];
+      options.zDirection = -1;
       layer.setSource(new IIIF(options));
       map.setView(new View({
         resolutions: layer.getSource().getTileGrid().getResolutions(),
--- a/examples/iiif.js
+++ b/examples/iiif.js
@@ -16,7 +16,7 @@ const layer = new TileLayer(),
 function refreshMap(imageInfoUrl) {
   fetch(imageInfoUrl).then(function(response) {
     response.json().then(function(imageInfo) {
-      const options = new IIIFInfo().readOptionsFromJson(imageInfo);
+      const options = new IIIFInfo(imageInfo).getTileSourceOptions();
       if (options === undefined || options.version === undefined) {
         notifyDiv.textContent = 'Data seems to be no valid IIIF image information.';
         return;
--- a/src/ol/format/IIIFInfo.js
+++ b/src/ol/format/IIIFInfo.js
@@ -278,7 +278,7 @@ class IIIFInfo {
    * @param {PreferredOptions} opt_preferredOptions Optional options for preferred format and quality.
    * @returns {import("../source/IIIF.js").Options} IIIF tile source ready constructor options.
    */
-  readFromJson(imageInfo, opt_preferredOptions) {
+  readOptionsFromJson(imageInfo, opt_preferredOptions) {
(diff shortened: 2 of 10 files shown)


mejackreed:
With regard to further work around `IIIFInfo`, `license`, and `attribution`, I think this will likely be removed in IIIF Image API v3 . See https://github.com/IIIF/trc/issues/12

So I would be 👍 for removing that, and let other IIIF viewers handle that work. Hopefully that simplifies the approach.

lutzhelm:
CircleCI fails with
> /home/circleci/repo/node_modules/ol-mapbox-style: Appears to be a git repo or submodule.

... which has nothing to do with changes from this pull request, or at least it's not obvious to me. What can be done to make CircleCI pass?

tschaub:
@lutzhelm, if you rebase your work on the latest from the openlayers master (`git pull --rebase openlayers master`), you should get the fix for the issue above.  You’ll need to force push your branch after the rebase (`git push --force origin master`).

lutzhelm:
@tschaub Thanks, that did work.

examples/iiif.js
```diff
@@ -0,0 +1,46 @@
...
+      if (options === undefined || options.version === undefined) {
+        notifyDiv.textContent = 'Data seems to be no valid IIIF image information.';
+        return;
+      }
+      const extent = [0, -options.size[1], options.size[0], 0];
+      options.zDirection = -1;
+      layer.setSource(new IIIF(options));
+      map.setView(new View({
+        resolutions: layer.getSource().getTileGrid().getResolutions(),
+        extent: extent,
```
tschaub: It looks like it would be useful to get the extent from the tile grid instead of having to know how to generate it in this example.  I'd be in support of adding an `@api` annotation to [the `getExtent` method](https://github.com/openlayers/openlayers/blob/e4f377fb4697907fd68a39f7bfcd6497c38c251a/src/ol/tilegrid/TileGrid.js#L237-L240).

If you do that in this branch, you should be able to remove the `extent = [0, -options.size[1], options.size[0], 0]` bit above.

src/ol/format/IIIFInfo.js
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * Supported image formats, qualities and supported region / size calculation features
+ * for different image API versions and compliance levels
+ * @const
+ * @type {Object<string, Object<string, SupportedFeatures>}
+ */
+const IIIF_PROFILE_VALUES = {
+  version1: {
+    level0: {
```
tschaub: These unquoted properties are subject to renaming by a compiler or minifier.  Since elsewhere you look up these properties using the `Version` enum below, you should construct this object using the same enum values.

In addition, if the `level0` etc values are parsed from remote data (it looks like they are), they should be quoted when used as properties to avoid renaming.

E.g.
```js
const IIF_PROFILE_VALUES = {};
IIF_PROFILE_VALUES[Versions.VERSION1] = {
  'level0': {} // etc
};
```

src/ol/format/IIIFInfo.js
```diff
@@ -0,0 +1,368 @@
...
+
+const COMPLIANCE_VERSION1 = new RegExp('^https?\:\/\/library\.stanford\.edu\/iiif\/image-api\/(1\.1\/)?compliance\.html#level[0-2]$');
+const COMPLIANCE_VERSION2 = new RegExp('^https?\:\/\/iiif\.io\/api\/image\/2\/level[0-2](\.json)?$');
+const COMPLIANCE_VERSION3 = new RegExp('(^https?\:\/\/iiif\.io\/api\/image\/3\/level[0-2](\.json)?$)|(^level[0-2]$)');
+
+function generateVersion1Options(iiifInfo) {
+  let levelProfile = iiifInfo.getComplianceLevelSupportedFeatures();
+  // Version 1.0 and 1.1 do not require a profile.
+  if (levelProfile === undefined) {
+    levelProfile = IIIF_PROFILE_VALUES.version1.level0;
```
tschaub: Here you should use `Versions.VERSION1` instead of `version1`.

It may be that most minifiers are not aggressive enough to rename `level0`, but it could be subject to renaming as well.  Which suggests that you use an enum for levels as well (or quoted strings everywhere).

src/ol/format/IIIFInfo.js
```diff
@@ -0,0 +1,368 @@
...
+function generateVersion2Options(iiifInfo) {
+  const levelProfile = iiifInfo.getComplianceLevelSupportedFeatures(),
+      additionalProfile = Array.isArray(iiifInfo.imageInfo.profile) && iiifInfo.imageInfo.profile.length > 1,
+      profileSupports = additionalProfile && iiifInfo.imageInfo.profile[1].supports ? iiifInfo.imageInfo.profile[1].supports : [],
+      profileFormats = additionalProfile && iiifInfo.imageInfo.profile[1].formats ? iiifInfo.imageInfo.profile[1].formats : [],
+      profileQualities = additionalProfile && iiifInfo.imageInfo.profile[1].qualities ? iiifInfo.imageInfo.profile[1].qualities : [],
+      attributions = [];
+  if (iiifInfo.imageInfo.attribution !== undefined) {
+    // TODO potentially dangerous
+    attributions.push(iiifInfo.imageInfo.attribution);
```
tschaub: It isn't immediately clear to me as a first time reader what is potentially dangerous about this.  So a comment like this would be a source of confusion.  Please clarify or remove the comment.

src/ol/format/IIIFInfo.js
```diff
@@ -0,0 +1,368 @@
...
+ * Format for transforming IIIF Image API image information responses into
+ * IIIF tile source ready options
+ *
+ * @api
+ */
+class IIIFInfo {
+
+  /**
+   * @param {Object|string} imageInfo Deserialized image information JSON response
+   * object or JSON response as string
```
tschaub: Would it be possible to provide a typedef for this object?

lutzhelm: That would require three different typedefs as each IIIF Image API info.json has a different structure. I'll look into it.

lutzhelm: I added typedefs for all possible objects as well as their dependent objects.
One problem: IIIF Image information responses from version 1.1 on are JSON-LD. And I'm not sure how to reference `@id` or `@context` in the doc, e.g.:

https://github.com/lutzhelm/openlayers/blob/a2c1e9cb5bc48d63f6bd00ea0299394ec162f237/src/ol/format/IIIFInfo.js#L38-L39

src/ol/format/IIIFInfo.js
lutzhelm: Following @mejackreed's [advice](https://github.com/openlayers/openlayers/pull/9430#issuecomment-482610729), I removed the attribution construction completely, so this comment is no longer present.

src/ol/format/IIIFInfo.js
lutzhelm: Done.

src/ol/format/IIIFInfo.js
lutzhelm: Done.
Resolved the potential `level0` minifying problem by using quoted strings, not as enum, as they are currently not exported / used anywhere else.

examples/iiif.js
lutzhelm: Done.

src/ol/format/IIIFInfo.js
tschaub: I'm curious how these will be handled as well.  If you rebase your branch on the latest in master (`git pull --rebase openlayers master`), we'll get the API docs built and hosted in the CI job.

You can build the docs locally with `npm run apidoc` (the output will be in `build/apidoc`).

I'm also curious if `tsc` complains about those typedefs (`npm run typecheck` to find out).

lutzhelm: `npm run apidoc` worked once I fixed another issue.
`tsc` complains about every contained JSON-LD keyword: `Identifier expected.`

lutzhelm:
@tschaub The doc for that union type is actually useful if the first type is a simple type. See https://github.com/openlayers/openlayers/pull/9520/commits/80eb8b62ce8340e933fa4601f351addecb9bbc8a

PR #9450, opened 2019-04-19
Set the interacting flag on pointerdrag instead of pointerdown

### LINKED ISSUE #9449: icons blurry on mouse down (>= v6.0.0.-beta3)
An icon is becoming blurry when clicked anywhere on the map. Seems that it is related to Point features, as Point styles in general are getting blurry when clicked somewhere on the map. Reproduced on Desktop (Chrome, FF and Edge)

1. go to https://openlayers.org/en/master/examples/icon.html?q=icon
2. click anywhere on the map
3. see how the icon is moved a little bit and becomes blurry

Some more examples: 
- https://openlayers.org/en/master/examples/cluster.html?q=cluster
- https://openlayers.org/en/master/examples/draw-features.html?q=draw

It seems as #9137 did change the behavior. Taking a look at DragPan interaction, setting the INTERACTING state moved from pointer move to pointer down.

ahocevar:
Fixes #9449.

### DIFF
--- a/src/ol/interaction/DragPan.js
+++ b/src/ol/interaction/DragPan.js
@@ -73,6 +73,10 @@ class DragPan extends PointerInteraction {
    * @inheritDoc
    */
   handleDragEvent(mapBrowserEvent) {
+    if (!this.panning_) {
+      this.panning_ = true;
+      this.getMap().getView().beginInteraction();
+    }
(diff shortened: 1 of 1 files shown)


ahocevar:
I just verified that pointerdown still stops an animation. Thanks for the review!

PR #9452, opened 2019-04-19
Fewer calls to the WebGL color callback

tschaub:
We currently call the user supplied `colorCallback` sixteen times per feature.  I think it would make it nicer for users if we allowed the return of a color array rather than a single color component.  To avoid unnecessary array allocation, we can provide a reusable color array as the third arg.

@jahow - let me know if you had something else in mind that makes this change a bad idea.

I'm also curious about the use case for a different color per vertex.  It seems like we could get down to one call per feature without sacrificing too much flexibility (but I might be missing the use case for shading with multiple colors).

### DIFF
--- a/examples/filter-points-webgl.js
+++ b/examples/filter-points-webgl.js
@@ -37,15 +37,16 @@ updateStatusText();
 class WebglPointsLayer extends VectorLayer {
   createRenderer() {
     return new WebGLPointsLayerRenderer(this, {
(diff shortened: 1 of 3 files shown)


jahow:
Sure, the idea was to avoid object allocation on every callback, using a reusable array makes sense. Thanks!

tschaub:
Yeah, I can imagine a radial gradient, but am not sure there is a strong use case for a gradient from each corner.

PR #9472, opened 2019-04-26
Add text textplacement default documentation

simonseyock:


simonseyock:
somehow i was probably looking at an older version of the documentation.

PR #9476, opened 2019-04-28
Document events fired by interactions

### LINKED ISSUE #9471: Event name documentation missing
**Describe the bug**
In the api documentation the names of special events do not appear.

**To Reproduce**
See for example https://openlayers.org/en/latest/apidoc/module-ol_interaction_Draw-Draw.html
under `Fires:` only the type of the event is shown (`DrawEvent`) but not the names. (`drawstart` and `drawend`)

**Expected behavior**
Like with the generic `change` events this should appear as: `drawstart (DrawEvent)`

tschaub:
The event names for our interactions are not currently showing up in the docs.  It looks like the `longname` for these doclets is not being generated correctly if the `@event` annotation doesn't have the full module path.

~~It feels like there should be an opportunity to fix this another way (supporting the existing syntax).  I'll keep digging to see where that `longname` is generated - but I think this is an ok change to merge in too~~.

**Update**: I updated this branch to pull in the changes from openlayers/jsdoc-plugin-typescript#5

Here is what the draw interaction docs look like after this change:
image

Fixes #9471

### DIFF
--- a/src/ol/interaction/Extent.js
+++ b/src/ol/interaction/Extent.js
@@ -38,7 +38,7 @@ import {createEditingStyle} from '../style/Style.js';
 const ExtentEventType = {
   /**
    * Triggered after the extent is changed
-   * @event ExtentEventType#extentchanged
+   * @event ExtentEvent#extentchanged
    * @api
    */
   EXTENTCHANGED: 'extentchanged'
@@ -47,10 +47,10 @@ const ExtentEventType = {
 
 /**
  * @classdesc
- * Events emitted by {@link module:ol/interaction/Extent~ExtentInteraction} instances are
+ * Events emitted by {@link module:ol/interaction/Extent~Extent} instances are
  * instances of this type.
  */
-class ExtentInteractionEvent extends Event {
+class ExtentEvent extends Event {
 
   /**
    * @param {import("../extent.js").Extent} extent the new extent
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Extent.js
```diff
@@ -38,7 +38,7 @@ import {createEditingStyle} from '../style/Style.js';
 const ExtentEventType = {
   /**
    * Triggered after the extent is changed
-   * @event ExtentEventType#extentchanged
+   * @event module:ol/interaction/Extent~ExtentEvent#extentchanged
```
tschaub: Even if we do make another change to address the `longname` generation, we need to make a change here too.  We want the docs pointing to `ExtentEvent` instead of `ExtentEventType`.

tschaub:
It looks like this might be the code that is supposed to be generating full module longnames:

https://github.com/openlayers/jsdoc-plugin-typescript/blob/a2c2c3e5fd12653038681da85f4afcc15ee51971/index.js#L201-L215

tschaub:
See openlayers/jsdoc-plugin-typescript#5 for a change in the plugin package.  After that is published, I can update this branch to pull in the new version and make the one additional change mentioned above (https://github.com/openlayers/openlayers/pull/9476#pullrequestreview-231487224).

src/ol/interaction/Extent.js
tschaub: Addressed in 9c6dbe5a339744e990bbb85c775515de6cf8bebb.

PR #9484, opened 2019-05-01
Build API docs in CI job

tschaub:
For pull requests that change doc annotations or our doc building tools (like #9476), this will let us add links to the hosted API docs in the PR for easier review.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:latest-browsers
+      - image: circleci/node:10-browsers
 
     working_directory: ~/repo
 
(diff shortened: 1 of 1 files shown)


tschaub:
Trying to debug the [build failure](https://circleci.com/gh/openlayers/openlayers/1198).  Not able to reproduce it here yet.

tschaub:
We're hitting the same issue described here: https://github.com/jsdoc3/jsdoc/issues/1530#issuecomment-487311288

Running `npm run apidoc` on Node 10, 11, and 12 works for me locally.  Maybe an issue with case sensitive filesystems.

tschaub:
Here is an example of the hosted API docs: https://1199-4723248-gh.circle-artifacts.com/0/apidoc/index.html

It is unfortunate to have to keep the build image pegged at Node 10, but I think it is nice to have the API docs generated in CI (so we know about failures, can see the result).  When we need to move beyond Node 10, if there isn't a solution to hegemonic/requizzle#6, we'll have to come up with an alternative doc solution.

PR #9489, opened 2019-05-03
Fix Zoomify to display retina tiles

crubier:
This is a small PR that changes a few lines in zoomify source:

  - allows to use retina tiles in Zoomify
  - fixes https://github.com/openlayers/openlayers/issues/7795
  - adds an option in the Zoomify example to display retina tiles

### DIFF
--- a/examples/zoomify.html
+++ b/examples/zoomify.html
@@ -11,5 +11,6 @@ tags: "zoomify, deep zoom, IIP, pixel, projection"
   <select id="zoomifyProtocol">
     <option value="zoomify">Zoomify</option>
     <option value="iip">IIP</option>
+    <option value="zoomifyretina">Zoomify Retina</option>
   </select>
 </div>
--- a/examples/zoomify.js
+++ b/examples/zoomify.js
@@ -12,6 +12,8 @@ const iipUrl = 'http://vips.vtech.fr/cgi-bin/iipsrv.fcgi?FIF=' + '/mnt/MD1/AD00/
 
 const layer = new TileLayer({
   source: new Zoomify({
+    tileSize: 256,
+    tilePixelRatio: 1,
     url: zoomifyUrl,
     size: [imgWidth, imgHeight],
     crossOrigin: 'anonymous'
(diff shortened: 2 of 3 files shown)


crubier:
Before:

Screenshot 2019-05-03 at 23 51 30

After:

Screenshot 2019-05-03 at 23 51 37

crubier:
This is a small PR, how long can we expect before review, merge and release ? Thank you :-) 🙏

tschaub:
@crubier - minor point, but it would be nice to squash all those commits into fewer that represent sensible chunks of work.  It looks like this work could be a single commit.

As a general practice, if you `git pull --rebase openlayers master` you can have your work rebased on the latest from the hosted master branch without introducing merge commits.  There are some good [docs online](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) about squashing commits if you haven't done that before.

crubier:
@tschaub Ok thank you, done

crubier:
@tschaub up :-)

crubier:
👍

jameshoward:
Just in case anyone runs into the same issue as me, the only reason the example works when setting `tileSize` to `256 / pixelRatio` is because the example image server effectively ignores the tile group directory. For example, http://vips.vtech.fr/cgi-bin/iipsrv.fcgi?zoomify=/mnt/MD1/AD00/plan_CHU-4HD-01/FOND.TIF/TileGroup1/2-0-0.jpg and http://vips.vtech.fr/cgi-bin/iipsrv.fcgi?zoomify=/mnt/MD1/AD00/plan_CHU-4HD-01/FOND.TIF/TileGroup100/2-0-0.jpg return the same image.

If you try to use the same method using an image from another Zoomify source (e.g. the converter), you will likely get 404 errors when trying to load tiles because the tile grouping calculations are different between the generated image and the `ZoomifySource`.

crubier:
@jameshoward did you find a solution to this? I was not aware of this problem, only tried on the demo (The server of the demo image is down now actually...)

ahocevar:
The remaining issue was fixed with #10562.

jameshoward:
@crubier No, I didn't find a solution other than to tweak the generator that was creating my tiles to output them at 256x256. The solution is #10562 looks great though!

PR #9493, opened 2019-05-05
Upgrade jsdoc to v3.6.1

ahocevar:
This pull request brings in `jsdoc-plugin-typescript#2.0.0`, because earlier versions are incompatible with jsdoc v3.6.x. It also converts the links in the API documenation's `index.md` to html links, because markdown links do not work inside html tables any more with recent `marked` versions.

Closes #9491.
Closes #9494.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:10-browsers
+      - image: circleci/node:latest-browsers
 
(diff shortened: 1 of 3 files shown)


tschaub:
We should also be able to change back to the `node:latest-browsers` image in Circle CI (reverting https://github.com/openlayers/openlayers/commit/284c1dff83fc2ef9e03361320b403b356c13dbe3).  That can also be done separately if you want to merge this as is.

PR #9500, opened 2019-05-07
KML string validation with extendedData

### LINKED ISSUE #9496: KML string validation with extendedData
**Describe the bug**
When creating a KML String with KML.writeFeatures(), with extendedData, the format isn't validated correctly.

**To Reproduce**
1. Create an array of features, e.g. Line with style and custom property 'lineDash'.
2. Write a KML string with:
`featString = new KML({
extractStyles: true,
defaultStyle: [kmlStyle],
}).writeFeatures(exportFeatures);`

Output:
< kml
  xmlns="http://www.opengis.net/kml/2.2" xmlns:gx="http://www.google.com/kml/ext/2.2" 
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
  xsi:schemaLocation="http://www.opengis.net/kml/2.2 
  https://developers.google.com/kml/schema/kml22gx.xsd">
  < Document>
    < name>Editor< /name>
    < Placemark>
      < ExtendedData>
        < Data name="lineDash">< value>1,10</value>< /Data>
      < /ExtendedData>
      < Style>
        < LineStyle>< color>ff0000ff< /color>< width>4< /width>< /LineStyle>
      </Style>
      < LineString>
        < coordinates>
          7.969207763671874,46.941824493511206 8.666839599609375,47.246610279847005 
          8.735504150390627,47.039246258977784
        < /coordinates>
      < /LineString>
    < /Placemark>
  < /Document>
< /kml>

3. Validate the KML string against the KML schema `kml22gx.xsd` fails:
[https://developers.google.com/kml/schema/kml22gx.xsd](https://developers.google.com/kml/schema/kml22gx.xsd)

**Expected behavior**
The tag < ExtendedData > should be located after the <Style> tag in the KML string.

edellucien:
Pull Request to face issue:
Fix #9496

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -2674,22 +2674,12 @@ function writePlacemark(node, feature, objectStack) {
   }
 
   // serialize properties (properties unknown to KML are not serialized)
-  const properties = feature.getProperties();
+  let properties = feature.getProperties();
 
   // don't export these to ExtendedData
   const filter = {'address': 1, 'description': 1, 'name': 1, 'open': 1,
     'phoneNumber': 1, 'styleUrl': 1, 'visibility': 1};
   filter[feature.getGeometryName()] = 1;
-  const keys = Object.keys(properties || {}).sort().filter(function(v) {
-    return !filter[v];
-  });
-
-  if (keys.length > 0) {
-    const sequence = makeSequence(properties, keys);
-    const namesAndValues = {names: keys, values: sequence};
-    pushSerializeAndPop(context, PLACEMARK_SERIALIZERS,
-      EXTENDEDDATA_NODE_FACTORY, [namesAndValues], objectStack);
-  }
 
   const styleFunction = feature.getStyleFunction();
   if (styleFunction) {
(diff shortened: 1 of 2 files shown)


src/ol/format/KML.js
```diff
@@ -2713,6 +2703,18 @@ function writePlacemark(node, feature, objectStack) {
   pushSerializeAndPop(context, PLACEMARK_SERIALIZERS,
     OBJECT_PROPERTY_NODE_FACTORY, values, objectStack, orderedKeys);
 
+  properties = feature.getProperties();
```
ahocevar: Instead of reading the properties again here, you could keep the order as it was, and only call `pushSerializeAndPop(context, PLACEMARK_SERIALIZERS, EXTENDEDDATA_NODE_FACTORY, [namesAndValues], objectStack)` here instead of above.

ahocevar: Thanks for your work on this! It would be great if you could address the above comment.

edellucien: Thanks for the review and the reactivity! :smile:

Issue #9139, opened 2019-01-23
master 'Icon Symbolizer'


jingyebuchen:
**Describe the bug**

**To Reproduce**
3. See error

**Expected behavior**
![icon](https://user-images.githubusercontent.com/41276803/51580042-230b9980-1efe-11e9-8a4c-5e7e4974e349.png)
Under the latest master, the Overlay position of this example is not fixed.

ahocevar:
Can you provide more context on the platform/browser you are using? I cannot reproduce.

jingyebuchen:
> Can you provide more context on the platform/browser you are using? I cannot reproduce.

Platform:win10
Browser: Google Chrome 68.0.3440.75
When the mouse scrolls to zoom out

fredj:
As suspected by Andreas, this is an issue with Chrome and the `will-change` rule.
See https://bugs.chromium.org/p/chromium/issues/detail?id=960953

PR #9513, opened 2019-05-10
Remove absolute position on ol-overlaycontainer divs

### LINKED ISSUE #9467: Popup in example no longer moves with the map
**Describe the bug**
In the [popup example](https://openlayers.org/en/master/examples/popup.html) the overlap no longer pans with the map.

**To Reproduce**
1. Go to https://openlayers.org/en/master/examples/popup.html
2. Click on the map to open the popup
3. Pan the map and see the popup not move

popup

**Expected behavior**
The popup should pan with the map.  Just like in the [overlay example](https://openlayers.org/en/master/examples/overlay.html).

fredj:
work in progress, trying to fix #9467 and #9139

The `absolute` position have been added in 898c349fbfed798d68bb570f4be55cc8c6c290d3

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -236,10 +236,6 @@ class PluggableMap extends BaseObject {
      * @type {!HTMLElement}
      */
     this.overlayContainer_ = document.createElement('div');
-    this.overlayContainer_.style.position = 'absolute';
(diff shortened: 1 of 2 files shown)


fredj:
not working, I'll try to come up with a cleaner solution

PR #9519, opened 2019-05-13
Use synchronous render in Mapbox example

jahow:
This uses an undocumented method to trigger a synchronous redraw instead of using the standard schedule from Mapbox. The CSS previously used to make the OL and MB views match is not necessary anymore.

As such, the glitches when zooming/rotating the view are gone.

Reference: https://github.com/mapbox/mapbox-gl-js/issues/7893#issue-408992184

Link to updated example: https://1269-4723248-gh.circle-artifacts.com/0/examples/mapbox-layer.html

### DIFF
--- a/examples/mapbox-layer.js
+++ b/examples/mapbox-layer.js
@@ -63,17 +63,6 @@ class Mapbox extends Layer {
       this.mbmap.getContainer().querySelector('.mapboxgl-control-container').remove();
     }.bind(this));
 
-    this.mbmap.on('render', function() {
-      // Reset offset
-      if (this.centerNextRender) {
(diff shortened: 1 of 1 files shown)


jahow:
Well the synchronous redraw API discussed in the link issue seems like it would be a perfect fit for us. I think this PR is fine for an example but ideally we'd need a stable API before implementing an actual Mapbox layer in OL.

Besides the example already overwrites an undocumented method: `_setupContainer`

ahocevar:
Ok merged for now, and added my +1 to the linked mapbox-gl issue. You should do the same maybe.

PR #9525, opened 2019-05-13
Support a custom render function for the Layer class

jahow:
The `ol/layer/Layer` class now accepts a `render` function as option, which will take the frame state as argument and is expected to return an HTML element. When provided, this function will be called instead of the default rendering code.

This effectively provides an easier way for doing custom rendering logic, which is demonstrated in the updated [Mapbox example](https://1280-4723248-gh.circle-artifacts.com/0/examples/mapbox-layer.html): all the logic for syncing with Mapbox & triggering a repaint is done in the provided `render` callback, and a custom class is not needed anymore.

### DIFF
--- a/examples/mapbox-layer.js
+++ b/examples/mapbox-layer.js
@@ -45,7 +45,7 @@ const mbMap = new mapboxgl.Map({
   style: 'https://maps.tilehosting.com/styles/bright/style.json?key=' + key,
   attributionControl: false,
   boxZoom: false,
-  center,
+  center: center,
   container: map.getTargetElement(),
   doubleClickZoom: false,
   dragPan: false,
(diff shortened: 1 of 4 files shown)


examples/mapbox-layer.js
```diff
@@ -1,83 +1,88 @@
...
+const view = map.getView();
+const center = toLonLat(view.getCenter(), view.getProjection());
+const key = 'ER67WIiPdCQvhgsUjoWK';
 
-    this.mbmap = new mapboxgl.Map(options);
+const mbMap = new mapboxgl.Map({
+  style: 'https://maps.tilehosting.com/styles/bright/style.json?key=' + key,
+  attributionControl: false,
+  boxZoom: false,
+  center,
```
fredj: minor: we should not use the es6 object key value shorthand
```suggestion
  center: center,
```

PR #9534, opened 2019-05-14
Remove unnecessary parts from the Mapbox layer example

tschaub:
As far as I could tell, these view settings are not required by [the Mapbox layer example](https://1296-4723248-gh.circle-artifacts.com/0/examples/mapbox-layer.html):
 * `minZoom: 1`
 * `extent: [-Infinity, -20048966.10, Infinity, 20048966.10]`
 * `smoothExtentConstraint: false`
 * `smoothResolutionConstraint: false`

In addition, since the render function has access to `frameState.viewState`, I removed the `map.getView()` call.  Since this will only work with a `EPSG:3857` view, I removed the `view.getProjection()` call.    And since the focus of this example is the Mapbox layer, I moved the vector layer to the bottom of the example.

### DIFF
--- a/examples/mapbox-layer.js
+++ b/examples/mapbox-layer.js
@@ -1,11 +1,67 @@
 import Map from '../src/ol/Map.js';
 import View from '../src/ol/View.js';
 import Layer from '../src/ol/layer/Layer';
-import {toLonLat} from '../src/ol/proj';
+import {toLonLat, fromLonLat} from '../src/ol/proj';
 import {Stroke, Style} from '../src/ol/style.js';
 import VectorLayer from '../src/ol/layer/Vector.js';
 import VectorSource from '../src/ol/source/Vector.js';
 import GeoJSON from '../src/ol/format/GeoJSON.js';
+
+const center = [-98.8, 37.9];
+const key = 'ER67WIiPdCQvhgsUjoWK';
+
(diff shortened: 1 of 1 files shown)


examples/mapbox-layer.js
```diff
@@ -100,5 +63,26 @@ const mbLayer = new Layer({
...
+    format: new GeoJSON()
+  }),
+  style: style
+});
+
+const map = new Map({
+  target: 'map',
+  view: new View({
+    center: fromLonLat(center),
+    zoom: 4
```
jahow: I'm pretty sure removing the `extent` constraint will break when the user pans out of the world north/south bounds. Although admittedly this constraint could simply be the default configuration!

tschaub: [I see](https://1296-4723248-gh.circle-artifacts.com/0/examples/mapbox-layer.html).  Yeah, let's make this the default behavior (not to allow panning off the top/bottom edge of the world.

tschaub: I added the magic view extent back in and opened #9544.

tschaub:
If #9545 is merged first, I'll remove the `extent: [-Infinity, -20048966.10, Infinity, 20048966.10]` from the view.  If this gets in first, I'll update the other branch to remove the extent (I confirmed that the example works without the extent after the changes in #9545).

PR #9536, opened 2019-05-14
More font cache hits

ahocevar:
This pull request drastically increases the font cache hit rate. Instead of rendering chunks of text as they follow a path, we now either render the whole text if it does not need to bend, or we cache and render individual characters.

For users of RTL text or languages that rely on ligatures, it is still possible to get correct rendering by configuring `maxAngle: 0` on a text style.


ahocevar:
This pull request reveals an issue with cut off labels for italic text, which I'm going to address separately.

PR #9537, opened 2019-05-14
WebGL / Points renderer refactoring

### LINKED ISSUE #9479: Precision Problem with WebGLPointsLayerRenderer and large Coordinates
In the Following exampel im using EPSG:3857 and a coordiante at 7 *10^6. When Mooving the Map the point is jumping in y direction.
https://codesandbox.io/embed/qv8zlqpzkq
This is effect appears to be a floating point precison problem with large numbers. In the x direction with a smaller coordinate the effect is present but smaller.

jahow:
What this PR does:
* generalizing the way webgl buffers are generated from geometries (still only supporting points for now) ; a base `WebGLLayerRenderer` class was created, which receives common logic for rendering
* making it so that the points renderer rebuilds the buffers during navigation; also coordinates are now expressed in the view space instead of world space to avoid precision loss at high zoom levels (fixes #9479)
* improve test coverage on existing code
* allow using custom attributes for rendering geometries: *this is not yet part of the public API for the points renderer*

The buffer rebuild operation will have to be done in a worker eventually, considering its impact on performance and the fact it has to be redone regularly. This will be done in another PR.

Also now internally geojson features are used as input to create the webgl buffers for rendering. Although serializable, geojson features are not transferrable to workers so the next step is compiling the render instructions straight into typed arrays and skip the geojson feature part (will be done in another PR too).

All this was made with the objective to render tile-based geometries in mind.

Example using the refactored code: https://1298-4723248-gh.circle-artifacts.com/0/examples/filter-points-webgl.html

### DIFF
--- a/src/ol/renderer/webgl/Layer.js
+++ b/src/ol/renderer/webgl/Layer.js
@@ -1,0 +1,74 @@
+/**
+ * @module ol/renderer/webgl/Layer
+ */
+import LayerRenderer from '../Layer.js';
+import WebGLHelper from '../../webgl/Helper';
+
+
+/**
+ * @typedef {Object} PostProcessesOptions
+ * @property {number} [scaleRatio] Scale ratio; if < 1, the post process will render to a texture smaller than
+ * the main canvas that will then be sampled up (useful for saving resource on blur steps).
+ * @property {string} [vertexShader] Vertex shader source
+ * @property {string} [fragmentShader] Fragment shader source
+ * @property {Object.<string,import("../../webgl/Helper").UniformValue>} [uniforms] Uniform definitions for the post process step
+ */
+
+/**
+ * @typedef {Object} Options
+ * @property {Object.<string,import("../../webgl/Helper").UniformValue>} [uniforms] Uniform definitions for the post process steps
+ * @property {Array<PostProcessesOptions>} [postProcesses] Post-processes definitions
+ */
+
+/**
+ * @classdesc
+ * Base WebGL renderer class.
+ * Holds all logic related to data manipulation & some common rendering logic
+ */
+class WebGLLayerRenderer extends LayerRenderer {
+
+  /**
+   * @param {import("../../layer/Layer.js").default} layer Layer.
+   * @param {Options=} [opt_options] Options.
+   */
+  constructor(layer, opt_options) {
+    super(layer);
+
+    const options = opt_options || {};
+
+    this.helper_ = new WebGLHelper({
+      postProcesses: options.postProcesses,
+      uniforms: options.uniforms
+    });
+  }
+
+  /**
+   * @inheritDoc
+   */
+  disposeInternal() {
+    super.disposeInternal();
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


src/ol/renderer/webgl/Layer.js
```diff
@@ -0,0 +1,162 @@
...
+ * feature will be ignored.
+ * @param {import("../../webgl/Buffer").default} vertexBuffer WebGL buffer in which new vertices will be pushed.
+ * @param {import("../../webgl/Buffer").default} indexBuffer WebGL buffer in which new indices will be pushed.
+ * @param {import("../../format/GeoJSON").GeoJSONFeature} geojsonFeature Feature in geojson format, coordinates
+ * expressed in EPSG:4326.
+ * @param {Array<string>} [opt_attributes] Custom attributes. An array of properties which will be read from the
+ * feature and pushed in the buffer in the given order. Note: attributes can only be numerical! Any other type or
+ * NaN will result in `0` being pushed in the buffer.
+ */
+export function pushFeatureInBuffer(vertexBuffer, indexBuffer, geojsonFeature, opt_attributes) {
```
ahocevar: Should probably be renamed to `pushFeatureToBuffer`.

src/ol/renderer/webgl/Layer.js
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Pushes a quad (two triangles) based on a point geometry
+ * @param {import("../../webgl/Buffer").default} vertexBuffer WebGL buffer
+ * @param {import("../../webgl/Buffer").default} indexBuffer WebGL buffer
+ * @param {import("../../format/GeoJSON").GeoJSONFeature} geojsonFeature Feature
+ * @param {Array<string>} [opt_attributes] Custom attributes
+ * @private
+ */
+function pushPointGeomInBuffer_(vertexBuffer, indexBuffer, geojsonFeature, opt_attributes) {
```
ahocevar: `pushPointFeatureToBuffer_`?

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -337,24 +348,54 @@ class WebGLPointsLayerRenderer extends LayerRenderer {
...
+
+    this.verticesBuffer_.getArray().length = 0;
+    this.indicesBuffer_.getArray().length = 0;
+
+    // saves the projection transform for the current frame state
+    this.helper_.makeProjectionTransform(frameState, this.renderTransform_);
+    makeInverseTransform(this.invertRenderTransform_, this.renderTransform_);
+
+    // loop on features to fill the buffer
+    const features = vectorSource.getFeatures();
```
ahocevar: I understand that this pull request is only a snapshot of a bigger refactoring. But just as note to self: here only features for the current extent and resolution should be fetched from the source.

jahow: Sure, the logic here is very basic but requesting *all* features all the time has the advantage of giving a better user experience when zooming in/out. Obviously this would not work for even bigger datasets (although for them a tile based approach would make more sense).

PR #9545, opened 2019-05-15
Avoid panning off the edge of the world

### LINKED ISSUE #9544: Make it so you can't pan off the edge of the world
The change in #9368 was a nice usability improvement, showing the user only one world at a time.  However, it is still too easy to pan off the (top or bottom) edge of the world.

edge

It would be nice if people didn't have to know magic things like `extent: [-Infinity, -20048966.10, Infinity, 20048966.10]` to keep users from panning off the edge of the world.

tschaub:
This makes it so the default view extent keeps the user from going off the edge.

no-edge

To restore the previous behavior, the view can be configured with `multiWorld: true`.

Fixes #9544.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1409,11 +1409,19 @@ function animationCallback(callback, returnValue) {
  */
 export function createCenterConstraint(options) {
   if (options.extent !== undefined) {
-    return createExtent(options.extent, options.constrainOnlyCenter,
-      options.smoothExtentConstraint !== undefined ? options.smoothExtentConstraint : true);
-  } else {
-    return centerNone;
+    const smooth = options.smoothExtentConstraint !== undefined ? options.smoothExtentConstraint : true;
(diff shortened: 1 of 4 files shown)


src/ol/View.js
```diff
@@ -1409,11 +1409,19 @@ function animationCallback(callback, returnValue) {
...
   if (options.extent !== undefined) {
-    return createExtent(options.extent, options.constrainOnlyCenter,
-      options.smoothExtentConstraint !== undefined ? options.smoothExtentConstraint : true);
-  } else {
-    return centerNone;
+    const smooth = options.smoothExtentConstraint !== undefined ? options.smoothExtentConstraint : true;
+    return createExtent(options.extent, options.constrainOnlyCenter, smooth);
+  }
+
+  const projection = createProjection(options.projection, 'EPSG:3857');
```
jahow: This could go in the `if` block maybe?

PR #9550, opened 2019-05-15
Setup for building workers

### LINKED ISSUE #9547: Make build setup ready for webworkers
Main thread is often too busy on rendering tasks, specially for vector rendering.

It should be great to introduce webworkers to lighten main thread processing during heavy rendering tasks.
We would like to use a first webworker in the WebGL point renderer, that would be responsible of the creation of the buffers from the rendering instructions.

Introducing webworkers in openlayers build system can be quite challenging, we should list here the requirements and the implementation suggestions:
- webworkers should be loaded in serve-examples mode
- webworkers should be part of the build-package output
- it should be transparent for a third application that use openlayers code that depends on a worker (no specific action on the build system of the application)
- webworkers could make the build grow consequently, we need to optimize them so they don't fetch too much dependencies
- we should put all workers in one file so we load dependencies only once
- rollup looks like a good options to build and provide the worker
- shall we add a preapre-worker phase task during build system ?

tschaub:
This adds a task that transforms modules in the `src/ol/worker` directory into modules that export a `create` function that returns a new worker with the source of the original module.  When serving the examples, we configure a custom loader that transforms the worker modules (without minification).  Same goes for running the tests.  When the rendering tests depend on a worker, we can update the webpack config there to use the same loader.

I added an example worker that posts messages with the current version identifier.  This is not meant to stick around, but it demonstrates a worker with dependencies.  [The example](https://1342-4723248-gh.circle-artifacts.com/0/examples/worker.html) and test should also be deleted before a release, but they can serve as a starting point for developing more full featured workers.

With this setup, we can decide how many worker factories we want.  It may make sense to have only one initially, but eventually we may want separate workers so we don't pile all dependencies into one.

I anticipate that we'll also want to add in some utilities for message handling.  We can do this by adding utility modules that are dependencies of either the worker itself or the module that creates the worker.  I think we can do this without complicating the way workers are serialized.

Fixes #9547.


src/ol/worker/version.js
```diff
@@ -0,0 +1,13 @@
...
+ */
+import {VERSION} from '../util';
+
+onmessage = event => {
+  console.log('version worker received message:', event.data); // eslint-disable-line
+  // @ts-ignore
+  postMessage(`version: ${VERSION}`);
+};
+
+export let create;
```
tschaub: This is here to satisfy our eslint/import rule.  After the worker is "built" it exports a named `create` function.

tasks/serialize-workers.js
```diff
@@ -0,0 +1,96 @@
...
+    },
+    common(),
+    resolve(),
+    babel({
+      'presets': [
+        [
+          '@babel/preset-env',
+          {
+            'modules': false,
+            'targets': 'last 2 version, not dead'
```
tschaub: We can control which browsers to support with our workers here.

tasks/serialize-workers.js
```diff
@@ -0,0 +1,96 @@
...
+    {
+      name: 'serialize worker and export create function',
+      renderChunk(code) {
+        return `
+          const source = ${JSON.stringify(code)};
+          const blob = new Blob([source], {type: 'application/javascript'});
+          const url = URL.createObjectURL(blob);
+          export function create() {
+            return new Worker(url);
+          }
```
tschaub: We may end up wanting to move the blob and url creation into the `create` function.  Profiling with more representative worker sources should clarify which way to go.

test/spec/ol/worker/version.test.js
```diff
@@ -0,0 +1,32 @@
...
+  afterEach(function() {
+    if (worker) {
+      worker.terminate();
+    }
+    worker = null;
+  });
+
+  describe('messaging', function() {
+    it('responds with the version', function(done) {
+      worker.addEventListener('error', done);
```
gberaudo: @tschaub, shouldn you fail the test before calling `done`?

src/ol/worker/version.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @module ol/worker/version
+ * A worker that responds to messages by posting a message with the version identifer.
+ */
+import {VERSION} from '../util';
+
+onmessage = event => {
+  console.log('version worker received message:', event.data); // eslint-disable-line
+  // @ts-ignore
```
fredj: This `@ts-ignore` can be removed if `webworker` is added in `tsconfig.json`, see https://github.com/fredj/openlayers/commit/bb022050edeef790c9cb505cce255702295fde53

test/spec/ol/worker/version.test.js
ahocevar: @gberaudo The `done` callback supports an `Error` as argument. When passed, the test will fail. See https://mochajs.org/#asynchronous-code.

src/ol/worker/version.js
tschaub: Feel free to add this commit (now or later) if this fixes the warning.  I was running into an issue with a d.ts file that had an outdated signature.  Can’t find the github issue now, but it looked like it was fixed upstream.

jahow:
Hi,

I've tried adding a webgl worker based on this branch and it dit work really well! Setting up a new worker is straightforward, it just requires adding `export default create;` at the end and call that same `create` function when instantiating a worker. Very simple and elegant IMO.

Can't wait to dig more into optimization now!

src/ol/worker/version.js
fredj: I've pushed the commit, thanks

PR #9551, opened 2019-05-16
Use HTMLImageElement.decode if available

fredj:
Unlike the event system, the image are not decoded in the main thread. See this article for a detailed explanation: https://medium.com/dailyjs/image-loading-with-image-decode-b03652e7d2d2

browser support:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decode#Browser_compatibility


fredj:
Yes, right: the handlers will always be called. I've searched on how to cancel a Promise but it doesn't seems to be easily doable. I'll keep searching on a solution

PR #9557, opened 2019-05-16
Be able to cancel image.decode operations

fredj:
Thanks @ahocevar  for the `hasListener` idea !

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -5,7 +5,9 @@ import ImageBase from './ImageBase.js';
 import ImageState from './ImageState.js';
 import {listenOnce, unlistenByKey} from './events.js';
 import EventType from './events/EventType.js';
+import EventTarget from './events/Target.js';
 import {getHeight} from './extent.js';
+import {VOID} from './functions.js';
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/ol/Image.js
```diff
@@ -155,8 +157,27 @@ class ImageWrapper extends ImageBase {
 export function listenImage(image, loadHandler, errorHandler) {
   const img = /** @type {HTMLImageElement} */ (image);
   if (img.decode) {
-    img.decode().then(loadHandler).catch(errorHandler);
-    return null;
+    // create fake event target to be able to cancel the image decoding.
+    const target = new EventTarget();
```
ahocevar: I think this is a bit overkill. Why not do something like
```diff
diff --git a/src/ol/Image.js b/src/ol/Image.js
index 45d0cf07c..283427747 100644
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -155,7 +155,15 @@ class ImageWrapper extends ImageBase {
 export function listenImage(image, loadHandler, errorHandler) {
   const img = /** @type {HTMLImageElement} */ (image);
   if (img.decode) {
-    img.decode().then(loadHandler).catch(errorHandler);
+    img.decode().then(function() {
+      if (!img.getAttribute('data-ol-ignore')) {
+        loadHandler();
+      }
+    }).catch(function() {
+      if (!img.getAttribute('data-ol-ignore')) {
+        errorHandler();
+      }
+    });
     return null;
   } else {
     const listenerKeys = [
@@ -174,6 +182,8 @@ export function unlistenImage(listenerKeys) {
   if (listenerKeys) {
     listenerKeys.forEach(unlistenByKey);
     listenerKeys = null;
+  } else {
+    this.getImage().setAttribute('data-ol-ignore', true);
   }
 }
```

ahocevar: Or even nicer: make listenImage() a member function, then you can use the instance to store a flag when the image is not being needed any more.

tschaub:
I didn't see this before putting up #9558.  I think we can simplify things by just returning an `unlisten` function in all cases.

fredj:
#9558

PR #9558, opened 2019-05-16
Stop listening for image decoding

tschaub:
See https://github.com/openlayers/openlayers/pull/9551#pullrequestreview-238330100

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -121,7 +121,7 @@ class ImageWrapper extends ImageBase {
       this.state = ImageState.LOADING;
       this.changed();
       this.imageLoadFunction_(this, this.src_);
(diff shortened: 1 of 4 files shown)


fredj:
I've pushed some tests in this branch

src/ol/Image.js
```diff
@@ -142,39 +142,47 @@ class ImageWrapper extends ImageBase {
...
+    };
+    promise.then(function() {
+      if (listening) {
+        loadHandler();
+      }
+    }).catch(function(error) {
+      if (listening) {
+        errorHandler();
+      }
+    });
```
fredj: We should add a `finally` and call `unlisten` to make sure we have the same behavior as the `listenOnce` function.

PR #9560, opened 2019-05-16
Simplify loading detection loop

ahocevar:
This pull request simplifies the loop that detects loading layers for the 'RENDERCOMPLETE' event.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -764,6 +764,21 @@ class PluggableMap extends BaseObject {
   }
 
   /**
+   * @return {boolean} Layers have sources that are still loading.
+   */
+  getLoading() {
+    const layerStatesArray = this.getLayerGroup().getLayerStatesArray();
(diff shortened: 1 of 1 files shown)


src/ol/PluggableMap.js
```diff
@@ -1398,23 +1413,3 @@ function createOptionsInternal(options) {
...
 export default PluggableMap;
-
-/**
- * @param  {Array<import("./layer/Base.js").default>} layers Layers.
- * @return {boolean} Layers have sources that are still loading.
- */
-function getLoading(layers) {
-  for (let i = 0, ii = layers.length; i < ii; ++i) {
-    const layer = layers[i];
-    if (typeof /** @type {?} */ (layer).getLayers === 'function') {
```
jahow: This was supposed to handle layer groups, right? Will they still be handled with the updated code?

ahocevar: The `layerStatesArray` is already flattened and contains all layers from all groups of the root group.

PR #9562, opened 2019-05-16
Optimize the WebGL points renderer using a worker

jahow:
The utilities in `ol/renderer/webgl/Layer` have been reworked (again!) with the objective to work exclusively with typed arrays.

The principle is:
* an `instructions` float-typed array receives the values that describe *how* a feature (in our case, points) should be rendered: position, size, color, rotate with view, etc.
  This is done with `writePointFeatureInstructions`.
* this array is used to write to both an index buffer (integer-typed array) and a vertex buffer (float-typed array) which are then ready to be sent to the GPU for rendering
  This is done with `writePointFeatureToBuffers` and is actually run in a worker.

Other minor improvements:
* Rendering tests now support web workers - @tschaub might have a better idea for handling this, for now I only reused the custom loader from the examples

Basing everything on typed array manipulation is a good way to improve performance and also enables using workers with almost no transfer overhead. Writing to index/vertex buffers is pretty quick for points but the gain will be even larger with lines & polygons (when running the earcut algorithm).

There are still performance bottlenecks which can be felt when navigating in the webgl examples (there is a noticeable lag when rebuilding the buffers). IMO the next optimization steps are:
* ~improve the `WebGLBuffer` class to not be based on plain arrays: this does not make sense as we might as well do everything with typed arrays; currently we fill typed arrays, then create arrays from typed arrays, then create typed arrays from arrays... all on the main thread. There's room for improvement!~
  Done, the `WebGLBuffer` class now only works with typed arrays, which brought a significant improvement.
* apparently the "styling" callbacks used by the points renderer have a significant cost in performance; this should be looked into a bit more and the API of the renderer should probably evolve at some point -> any ideas @ahocevar?

~I'm leaving this PR as draft if anyone wants to pick up the work (ping @fgravin).~

This PR is now ready for review, IMO performance is acceptable although there probably still are optimizations to make. On the filter webgl example, rebuilding the buffers takes around 50~70ms on the main thread for several thousand features, and this is only done when the view is not moving so the lag is hardly noticeable.

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -291,6 +291,8 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
         // saves the projection transform for the current frame state
         this.renderTransform_ = projectionTransform;
         makeInverseTransform(this.invertRenderTransform_, this.renderTransform_);
+
+        this.renderInstructions_ = new Float32Array(event.data.renderInstructions);
       }
     }.bind(this));
   }
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -383,7 +383,7 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
 
     const features = vectorSource.getFeatures();
     const totalInstructionsCount = POINT_INSTRUCTIONS_COUNT * features.length;
-    if (this.renderInstructions_.length !== totalInstructionsCount) {
+    if (!this.renderInstructions_ || this.renderInstructions_.length !== totalInstructionsCount) {
       this.renderInstructions_ = new Float32Array(totalInstructionsCount);
     }
 
@@ -427,6 +427,7 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
     message['projectionTransform'] = projectionTransform;
 
     this.worker_.postMessage(message, [this.renderInstructions_.buffer]);
+    this.renderInstructions_ = null;
   }
 
 
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -2,7 +2,7 @@
  * @module ol/renderer/webgl/PointsLayer
  */
 import WebGLArrayBuffer from '../../webgl/Buffer.js';
-import {DYNAMIC_DRAW, ARRAY_BUFFER, ELEMENT_ARRAY_BUFFER, FLOAT, EXTENSIONS as WEBGL_EXTENSIONS} from '../../webgl.js';
+import {DYNAMIC_DRAW, ARRAY_BUFFER, ELEMENT_ARRAY_BUFFER, FLOAT} from '../../webgl.js';
 import {DefaultAttrib, DefaultUniform} from '../../webgl/Helper.js';
 import GeometryType from '../../geom/GeometryType.js';
 import WebGLLayerRenderer, {
@@ -19,7 +19,6 @@ import {
   apply as applyTransform
 } from '../../transform.js';
 import {create as createWebGLWorker} from '../../worker/webgl.js';
-import {includes} from '../../array.js';
 
 const VERTEX_SHADER = `
   precision mediump float;
@@ -421,7 +420,7 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
     const message = {
       type: WebGLWorkerMessageType.GENERATE_BUFFERS,
       renderInstructions: this.renderInstructions_.buffer,
-      useShortIndices: !includes(WEBGL_EXTENSIONS, 'OES_element_index_uint')
+      useShortIndices: !this.helper.getElementIndexUintEnabled()
     };
     // additional properties will be sent back as-is by the worker
     message['projectionTransform'] = projectionTransform;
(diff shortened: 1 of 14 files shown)


ahocevar:
Thanks @jahow! This is a good pull request to test some concepts. You are right that it is time to look at how we want the user facing API to look like. I’ll think about this more after this week. @tschaub May also have ideas here. One thing to think ahead is how we want to do styling I think.

src/ol/worker/webgl.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * @module ol/worker/webgl
+ * A worker that does cpu-heavy tasks related to webgl rendering
```
tschaub: Minor, but the description needs to go above the `@module` annotation.  See #9563.

jahow:
@tschaub the rendering tests fail on the CI although they (sometime) pass on my machine. The error given is related to workers:
```
> node rendering/test.js "--force"

(node:494) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_ARG_VALUE]: The argument 'path' must be a string or Uint8Array without null bytes. Received '\u0000rollupPluginBabelHelpers.js'
    at Object.stat (fs.js:829:3)
    at Object.stat (/home/circleci/repo/node_modules/graceful-fs/polyfills.js:281:19)
    at Storage.provide (/home/circleci/repo/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:78:3)
    at CachedInputFileSystem.stat (/home/circleci/repo/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:231:21)
    at /home/circleci/repo/node_modules/webpack/lib/CachePlugin.js:59:10
    at symbolEach (/home/circleci/repo/node_modules/neo-async/async.js:2439:9)
    at Object.each (/home/circleci/repo/node_modules/neo-async/async.js:2845:16)
    at /home/circleci/repo/node_modules/webpack/lib/CachePlugin.js:56:14
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/circleci/repo/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:20:1)
    at /home/circleci/repo/node_modules/webpack/lib/Compiler.js:278:19
(node:494) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:494) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Navigation Timeout Exceeded: 60000ms exceeded
```

I must have done something wrong in the rendering tests webpack config but can't figure out what.

src/ol/worker/webgl.js
jahow: fixed

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -418,7 +419,8 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
     /** @type import('./Layer').WebGLWorkerGenerateBuffersMessage */
     const message = {
       type: WebGLWorkerMessageType.GENERATE_BUFFERS,
-      renderInstructions: this.renderInstructions_.buffer
+      renderInstructions: this.renderInstructions_.buffer,
+      useShortIndices: !includes(WEBGL_EXTENSIONS, 'OES_element_index_uint')
```
fredj: `this.helper_.hasOESElementIndexUint` could be used here

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -420,7 +419,7 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
     const message = {
       type: WebGLWorkerMessageType.GENERATE_BUFFERS,
       renderInstructions: this.renderInstructions_.buffer,
-      useShortIndices: !includes(WEBGL_EXTENSIONS, 'OES_element_index_uint')
+      useShortIndices: !this.helper.hasOESElementIndexUint
```
jahow: Good idea, although I'd look for an actual API on the helper side, could be something specific to this extension such as `this.helper.isElementIndexUintAllowed()` or something.

fredj: And we should move the function out of the WebGLHelper class so it can be used without an instance

jahow: Then if not on an instance the original implementation is fine IMO. The `WEBGL EXTENSIONS` array contains all available extensions, so:
- either we assume that *available extensions are always enabled on webgl contexts* in which case we only need to check against this array
- or we do not do this assumption and we check *on the helper instance* whether the extension was enabled

I went for the assumption initially, considering that there is no reason to not enable an extension in my knowledge. Anyway, keep in mind that there are dozens of extensions so writing specialized wrappers for each may become too verbose.

jahow:
I've resolved the conflicts for this PR and everything works fine. Could anyone give this a full review so we can move on to implementing hit detection & custom attributes on the webgl renderer?

Thanks!

jahow:
I realized the handling for the `element_index_uint` extension generates a good amount of complexity in this PR, while it is apparently supported by almost everything (~98%): https://webglstats.com/webgl/extension/OES_element_index_uint

Should we get rid of this to simplify code?

ahocevar:
@jahow Taking `element_index_uint` for granted and removing the code that handles the extension sounds good to me.

jahow:
> @jahow Taking `element_index_uint` for granted and removing the code that handles the extension sounds good to me.

Aaand it's done.

marcjansen:
Thanks @jahow, code looks great, I'll leave meaningful reviews to others that are better suited. 👌

jahow:
Thanks all for your feedback, onward to interactivity!

PR #9568, opened 2019-05-17
Clip high res tiles out of low res tiles only once

ahocevar:
The code that clips the space for higher resolution tiles out of lower resolution tiles when rendering points and labels for vector tiles had a bug: it performed the clip with every executor group, regardless of whether a clip for the current tile z was done already.

This pull request fixes that.


ahocevar:
Looks like this can be simplified even more. I'll be adding another commit.

ahocevar:
I'm closing this for now. More tests would be needed to ensure this does what it is supposed to do.

PR #9571, opened 2019-05-17
Small examples improvements

fredj:


examples/icon-color.js
```diff
@@ -25,7 +25,7 @@ rome.setStyle(new Style({
   image: new Icon({
     color: '#8959A8',
     crossOrigin: 'anonymous',
-    src: 'data/dot.png'
```
tschaub: Curious why we would change this for Rome only.

fredj: It's only to have a mix of different image type

PR #9584, opened 2019-05-21
Reuse render target

### LINKED ISSUE #9367: Critical (and mindblowing) performance bug with empty-source layers
**Describe the bug**

Windows: 10 
Ram: 16Gb
CPU: i7
Browser: Chrome 73

**When we have layers using an empty source (no features!), performance and FPS drop in an unexpected and dramatic way.**

We're measuring both **ZOOM** and **PAN** actions,

Surprisingly, if we just add 1 simple (painted!) feature to every layer, PAN performance works as expected hitting 60fps.

Using that very same 1 feature on every layer, ZOOM interaction improves but not with the scale that PAN does.

**To Reproduce**
1. Go to https://codesandbox.io/s/7m0434on76
2. Pan (works fine)
3. Zoom works but it's kinda clunky (with very few features, 1/layer)
4. Now remove the 'olFeature' from the source array inside the loop and....boom!
5. Pan works incredibly slow with a huge drain on FPS (with 0 features painted/registered)
6. Zoom works even worse than before

**Expected behavior**
Regardless the number of layers, an empty source should not affect or impact performance since it's not really painting anything.

Indeed, it's always expected to get worse performance when more features are painted.

We reviewed the call tree and the method **IntermediateCanvas.drawImage** was consuming (scripting) our resources.

This is a (very) critical bug.

### LINKED ISSUE #9457: Ol 6 beta performance problem with multiple layers
**Describe the bug**
When creating a map with multiple layers in the old version of openlayers the performance was okay now its not working any more panning and zooming is really slow

**To Reproduce**
create a map like this:
And when you start adding layers from the attached zip file you can see it gets slower and slower.
ol 6.0.0-beta6 https://codesandbox.io/s/14kkpqkxjl
ol 5.3.0  https://codesandbox.io/s/mjv0r8138y?fontsize=14

and then you can use the attached zip to add layers.

[gpx-files.zip](https://github.com/openlayers/openlayers/files/3101158/gpx-files.zip)
ol-6b6 performance
ol 6b6-code

**Expected behavior**
The map is as fast a the pervious versions of Openlayers

ahocevar:
This pull request changes the composite rendering to minimize the creation of viewport size canvas objects.

To achieve this, the following changes were made:
* Each canvas layer renderer renders to one or two canvases which are wrapped in a div.
* The composite map renderer passes the container of a rendered layer to the next layer for rendering.
* By calling a `useContainer` method, layer renderers can reuse that previous layer's render container, or decide to create their own.
* Opacity is now handled on the canvas where possible. Vector layers with a layer opacity cannot reuse a previous layer's container, and such containers cannot be reused by other layers. Raster tile layers can only apply tile opacity transitions when no opacity is set on the layer.
* In order to avoid flicker when rendering decluttered content, the vector tile layer renderer needed a bit of an overhaul: Now it uses a more straightforward queueing strategy for rendering tile images: only when in a load listener, we queue them. Otherwise we render them immediataly. With this change, the rendered result is visible earlier, and we have less composition and decluttering work to do, so we don't need to [bail out early from `renderFrame`](https://github.com/openlayers/openlayers/blob/335648d6132696350571033cb04da27b60ff5012/src/ol/renderer/canvas/VectorTileLayer.js#L394-L397). This could be achieved without sacrificing responsiveness, because performance was improved by fixing a declutter bug where [features outside the visible extent were considered for decluttering](https://github.com/openlayers/openlayers/blob/335648d6132696350571033cb04da27b60ff5012/src/ol/render/canvas/Executor.js#L368).

Fixes #9367 
Fixes #9457

### DIFF
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -79,7 +79,7 @@ class CanvasLayerRenderer extends LayerRenderer {
   useContainer(target, transform, opacity) {
     const layerClassName = this.getLayer().getClassName();
     let container, context;
-    if (target) {
+    if (target && target.style.opacity === '') {
       const canvas = target.firstElementChild;
       if (canvas instanceof HTMLCanvasElement) {
         context = canvas.getContext('2d');
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -24,6 +24,12 @@ class CanvasLayerRenderer extends LayerRenderer {
 
     /**
      * @protected
+     * @type {HTMLElement}
+     */
+    this.container = null;
+
+    /**
+     * @protected
      * @type {number}
      */
     this.renderedResolution;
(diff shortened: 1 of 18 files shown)


ahocevar:
The mapbox-style example benefits a lot from this change.

Before:
before

After:
after

This should make the canvas memory issue on iOS devices go away.

ahocevar:
I need to take another look at the rework of the vector tile render queue. Looks like we don't render all tiles at the moment.

ahocevar:
I was able to improve the reworked code so we render all needed tiles, but skip those we do not need.

src/ol/renderer/canvas/Layer.js
```diff
@@ -55,20 +61,58 @@ class CanvasLayerRenderer extends LayerRenderer {
...
+    let container, context;
+    if (target && target.style.opacity === '') {
+      const canvas = target.firstElementChild;
+      if (canvas instanceof HTMLCanvasElement) {
+        context = canvas.getContext('2d');
+      }
+    }
+    if (context && context.canvas.style.transform === transformToString(transform)) {
+      // Container of the previous layer renderer can be used.
+      target.classList.add(layerClassName);
```
fredj: This breaks the [semi-transparent-layer example](https://248-4974939-gh.circle-artifacts.com/0/examples/semi-transparent-layer.html) where the CSS filter should only be applied to the OSM layer.

We could either take the class name into account in the if above or simply revert 22eb96637ff87ec62d2d9268854f9453bd934167

ahocevar: I decided to do the former.

jeroenvheel:
@ahocevar when can we test this in a new beta version?

ahocevar:
I just published v6.0.0-beta.9

jeroenvheel:
@ahocevar the speed of the map is perfect right now 🙌

jeroenvheel:
@ahocevar the speed for Vector layers in now spot on but the speed if the VectorImage layer is still slow is this a thing that is also going to be fixed in ol6 ?

PR #9591, opened 2019-05-22
Use the layer state from the frameState to render TileLayers

### LINKED ISSUE #9587: setOpacity doesn't work on ol/layer/Group
**Describe the bug**
I have a couple layers displayed on my map, most react well to `setOpacity` and I have a reluctant one that doesn't change its transparency when `setOpacity` is applied

**To Reproduce**
https://s.geo.admin.ch/820917c58e and try to change opacity on both visible layers, the one with buildings and dwellings doesn't work when the other does (same code and same layer type `ol/layer/Tile`)

That's something new with OL6 betas as it's working fine on 5.3 (see https://s.geo.admin.ch/82092c7fd6)
I don't really know where to look for clues, we are using `setOpacity()` to set this property in our code.
layer creation here https://github.com/geoadmin/mf-geoadmin3/blob/845b8d131b909fa34116504c9cfcf2387303724b/src/components/map/LayersService.js#L540-L549
opacity set here https://github.com/geoadmin/mf-geoadmin3/blob/845b8d131b909fa34116504c9cfcf2387303724b/src/components/map/DefinePropertiesForLayerService.js#L50-L57

**Expected behavior**
Same behaviour when changing opacity.

fredj:
fixes #9587


fredj:
The issue is that the opacity of a layer group is not taken into account when `getLayerState` is called on the children layers.

My idea is to use the layer states from the `frameState` as it was done in 5.x:
https://github.com/openlayers/openlayers/blob/v5.3.2/src/ol/renderer/canvas/Map.js#L149-L160

ahocevar:
What about something like this:

The map renderer could set a `layerIndex` property on the `frameState` before calling the layer renderer. Then layer renderers could get the layer state with `frameState.layerStatesArray[frameState.layerIndex]`, and we'd no longer need to call `renderFrame` on layer renderers with a `layerState`.

fredj:
> What about something like this:
> 
> The map renderer could set a `layerIndex` property on the `frameState` before calling the layer renderer. Then layer renderers could get the layer state with `frameState.layerStatesArray[frameState.layerIndex]`, and we'd no longer need to call `renderFrame` on layer renderers with a `layerState`.
> 
> The JSDoc for `getLayerState()` on the layer should clearly state that this method is not meant to be called by layers or layer renderers.

started here #9592

PR #9592, opened 2019-05-22
Remove `layerState` param from `prepareFrame` and `renderFrame` function

### LINKED ISSUE #9587: setOpacity doesn't work on ol/layer/Group

fredj:
replaces #9591

Instead of passing the `layerState` to the `prepareFrame` and `renderFrame` functions, the state is now taken from the `frameState` by using the new `layerIndex` property:
```
frameState.layerStatesArray[frameState.layerIndex];
```

fixes #9587


src/ol/layer/Base.js
```diff
@@ -83,6 +83,9 @@ class BaseLayer extends BaseObject {
   }
 
   /**
+   * This method is not meant to be called by layers or layer renderers because the state
+   * is iccorect if the layer is included in a layer group.
```
tschaub:
```suggestion
   * is incorrect if the layer is included in a layer group.
```

fredj:
Thanks for the review

PR #9596, opened 2019-05-23
Vector source / prevent adding features with duplicate id in the collection

### LINKED ISSUE #6183: ol.source.Vector with spatial index and ol.Collection for features is broken
I got a simple map that request geojson features dynamically based on extent of the map from a WFS (geoserver). 

For some reason, I need to only load the features on the extent. So after map pan (map.on('moveend')) I clear a ol.Collection and repopulate the Collection with the new requested features. 

The problem happen on the first clear (move the map). I get the error message:

    Cannot read property 'forEach' of undefined

The error seams to come from the event propagation of the pop event and append in the function

    ol.source.Vector.prototype.removeFeatureInternal

Here is a sample code and the jsbin of the reproductible error

```javascript
var myCollection = new ol.Collection();

var myLayer = new ol.layer.Vector({
        features: myCollection
    })
})

function myUpdateFunction() {
    // var url = 'https://raw.githubusercontent.com/johan/world.geo.json/master/countries/USA/WY.geo.json'
    var url = "https://raw.githubusercontent.com/johan/world.geo.json/master/countries.geo.json";
    $.get(url).then(function(result) {
        var format = new ol.format.GeoJSON({
            defaultDataProjection: "EPSG:4326",
            featureProjection: "EPSG:3857"
        });
        var features = format.readFeatures(result);

        myCollection.clear();
        myCollection.extend(features);
    });
}

var map = new ol.Map({
    layers: [
        new ol.layer.Tile({source: new ol.source.OSM()}),
        myLayer
    ],
    target: 'map',
    controls: ol.control.defaults({
        attributionOptions:({collapsible: false})
    }),
    view: new ol.View({center: [0, 0],zoom: 2})
});

map.on('moveend', function(e) {
    myUpdateFunction();
});

```
https://jsbin.com/tivezamufi/1/edit?html,js,console,output

jahow:
Previously two features with the same id could be pushed manually in the features collection and stay there.

This would cause an error when clearing the source (related to change events not registered for the features with duplicate id).

This only happened when manipulating the source directly (either through `extend()` or `push()`).

Fixes #6183.

Note: the added test fails without the change in the source code.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -282,6 +282,9 @@ class VectorSource extends Source {
    * instead. A feature will not be added to the source if feature with
    * the same id is already there. The reason for this behavior is to avoid
    * feature duplication when using bbox or tile loading strategies.
+   * Note: this also applies if an {@link module:ol/Collection} is used for features,
+   * meaning that if a feature with a duplicate id is added in the collection, it will
+   * be removed from it right away.
    * @param {import("../Feature.js").default} feature Feature to add.
    * @api
    */
@@ -300,6 +303,9 @@ class VectorSource extends Source {
     const featureKey = getUid(feature);
 
     if (!this.addToIndex_(featureKey, feature)) {
+      if (this.featuresCollection_) {
+        this.featuresCollection_.remove(feature);
+      }
       return;
     }
 
--- a/test/spec/ol/source/vector.test.js
+++ b/test/spec/ol/source/vector.test.js
@@ -719,6 +719,19 @@ describe('ol.source.Vector', function() {
       expect(source.getFeatures().length).to.be(0);
     });
 
+    it('prevents adding two features with a duplicate id in the collection', function() {
+      source = new VectorSource({
+        features: new Collection()
+      });
+      const feature1 = new Feature();
+      feature1.setId('1');
+      const feature2 = new Feature();
+      feature2.setId('1');
+      const collection = source.getFeaturesCollection();
+      collection.push(feature1);
+      collection.push(feature2);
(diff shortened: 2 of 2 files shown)


jahow:
This fix is more of a workaround really, as we can't prevent adding features with the same id directly in the collection. I've tested it with the code provided in the original issue and it works well.

Thanks for any review

src/ol/source/Vector.js
```diff
@@ -299,12 +299,12 @@ class VectorSource extends Source {
   addFeatureInternal(feature) {
     const featureKey = getUid(feature);
 
+    this.setupChangeEvents_(featureKey, feature);
```
ahocevar: This change will leave change events registered on features that could not be added to the source (i.e. with a spatial index).

src/ol/source/Vector.js
```diff
@@ -299,12 +299,12 @@ class VectorSource extends Source {
...
+
     if (!this.addToIndex_(featureKey, feature)) {
       return;
```
ahocevar: It would be better to remove the feature from the collection here, if there is a collection.

jahow: Problem is, `addFeatureInternal` can be called from a collection `add` event, so it would mean removing an element while adding it... would that be possible?

See: https://github.com/openlayers/openlayers/blob/ddb7da37336f803c635d3becfd316047615944a6/src/ol/source/Vector.js#L451

ahocevar: That should work.

jahow:
I've updated the PR following @ahocevar's suggestion, and it looks better indeed. Thanks

PR #9600, opened 2019-05-24
Fix addEventListener call in PluggableMap

bartvde:
We found an issue in PluggableMap that can cause issues in a minified environment, this PR fixes the call to addEventListener to be explicit for window, which fixes the issue in our case


ahocevar:
@bartvde To make the CI pass, please rebase this on top of master.

src/ol/PluggableMap.js
```diff
@@ -1031,7 +1031,7 @@ class PluggableMap extends BaseObject {
 
       if (!this.handleResize_) {
         this.handleResize_ = this.updateSize.bind(this);
-        addEventListener(EventType.RESIZE, this.handleResize_, false);
+        window.addEventListener(EventType.RESIZE, this.handleResize_, false);
```
ahocevar: Would `global` instead of `window` also work?

bartvde: yes that should also do the trick, but we had other occurences of window in the source so I went with that

bartvde:
I'll open up a clean PR off of master, didn't realize the same code is still in master

src/ol/PluggableMap.js
bartvde: I verified that global also works, see #9601 for the update

PR #9601, opened 2019-05-24
Use global addEventListener in PluggableMap

bartvde:
replaces #9600


tschaub:
See https://github.com/openlayers/openlayers/pull/9603#pullrequestreview-241717833.  Same question here.

PR #9602, opened 2019-05-24
Fix addeventlistener

bartvde:
Same as #9601 but for v5.3.x


bartvde:
pushed to the wrong branch in my fork, #9603 is the right one

PR #9603, opened 2019-05-24
Use global scope for addEventListener

bartvde:
Same as #9601 but for v5.3.x

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1031,7 +1031,7 @@ class PluggableMap extends BaseObject {
 
       if (!this.handleResize_) {
         this.handleResize_ = this.updateSize.bind(this);
-        addEventListener(EventType.RESIZE, this.handleResize_, false);
+        window.addEventListener(EventType.RESIZE, this.handleResize_, false);
       }
     }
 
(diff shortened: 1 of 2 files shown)


ahocevar:
Ok sorry for killing your first pull request on this, I hadn’t seen it was against 5.x. The failing test is caused by a puppeteer update. I think using v1.10.0 should fix it.

bartvde:
I will change back to ```window``` @tschaub it's what I originally had ;-)
@ahocevar agreed?

bartvde:
PR updated @tschaub

bartvde:
I'm unsure what is up with the CI @ahocevar for the v5.3.x branch. I tried adding pupeteer as a dev dependency but the result is the same

tschaub:
We didn’t use Puppeteer in the 5.x series, right?  I think the Karma confit sets the browser to Chrome in CI (only reviewing on a small device though).

tschaub:
@bartvde the failure I see is about precision in an assertion in the scaleline tests.  I assume there is a change in floating point math in a Chrome update.  To get this passing, I would lower the precision of the equality check in the scaleline test.

bartvde:
@tschaub correct, I pushed a fix for the test. Let me know if this would also make sense on master or not

ahocevar:
@bartvde:

> Let me know if this would also make sense on master or not

In master, this was already fixed with #9305.

bartvde:
ah right, that makes sense @ahocevar

Issue #9609, opened 2019-05-27
Support usage without a Bundler


mogoh:
I JavaScript bundler like Parcel is recommended by the documentation. In fact not using a bundler leads to some errors.

While using a bundler has many advantages, sometimes it is useful to not having to use a bundler.

* Configuring a bundler add an additional layer.
* A bundler needs to be installed and configured. A bundler can has it own errors.
* Bundled JavaScript is more complicated to debug if everything gets wrapped in a IIFE.
* While bundling files for production, in development, the opposite can be true. Recompling after every change of a file can slow down development.

Over all, having the option to not use a bundler would be nice. A documentation for that would be great.

I guess this takes minimal code changes. If I am not mistaken, ol should not use "bare" imports like 
`import { easeOut } from './easing';` but `import { easeOut } from './easing.js';`. But I do not really know what there is to do to achieve that.

ThomasG77:
> Over all, having the option to not use a bundler would be nice.
If you want the option with `import` syntax and without bundler, not ready yet (at least for production). This answer is for overall JavaScript ecosystem and not particular to OpenLayers.

I've wrote an answer about various ways to use OpenLayers (with bundlers, jspm and standalone) https://stackoverflow.com/questions/51093964/why-examples-dont-work-a-struggle-with-imports/51110511#51110511

tschaub:
@mogoh, for what it’s worth, the library already uses proper relative paths in our import specifiers (explicit, not implied file extensions; no “directory” imports with implied index files).  So, the library is ready for unbundled use, but as @ThomsaG77 points out, the same is not true of browsers and the JS modules ecosystem.

The library has dependencies that are not all published with the same rules we follow (proper relative paths).  Even if this were the case, bundling and minification is [still recommended](https://developers.google.com/web/fundamentals/primers/modules#bundle).

> Our general recommendation is to continue using bundlers before deploying modules to production. In a way, bundling is a similar optimization to minifying your code: it results in a performance benefit, because you end up shipping less code. Bundling has the same effect! Keep bundling.

While it adds extra overhead to the development process, getting comfortable using a bundler has benefits.

mogoh:
Thank you for your help. I am not that experienced this JavaScript so bear with me, if I say something stupid.

> @mogoh, for what it’s worth, the library already uses proper relative paths in our import specifiers (explicit, not implied file extensions; no “directory” imports with implied index files).

@tschaub 
[...]

PR #9615, opened 2019-05-28
Include full filename in import specifier

tschaub:
We are not consistent with our use of file extensions in import specifiers.  We were originally trying to use extensions consistently, but as @mogoh points out, we have gotten inconsistent (see https://github.com/openlayers/openlayers/issues/9609#issuecomment-496586788).

This updates the version of `eslint-config-openlayers` that we use to a version that treats this inconsistency as an error.

image

Closes #9614.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "copy-webpack-plugin": "^5.0.3",
     "coveralls": "3.0.3",
     "eslint": "^5.16.0",
-    "eslint-config-openlayers": "^11.0.0",
+    "eslint-config-openlayers": "^12.0.0",
     "expect.js": "0.3.1",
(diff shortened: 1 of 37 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "url": "https://github.com/openlayers/openlayers/issues"
   },
   "dependencies": {
+    "eslint-config-openlayers": "^12.0.0",
```
fredj: Is it correct to move the package from `devDependencies` to `dependencies` ?

tschaub: Thanks for the catch.  Fixed in deb00f20fe0c3b9bab4768d9cefe48c4de09a9fb.

PR #9622, opened 2019-05-30
Fix #9510: EsriJSON readGeometry should not mutate input

### LINKED ISSUE #9510: EsriJSON readGeometry mutate input
**Describe the bug**
Hello,
I read geometry from Esri back-end using EsriJSON `readGeometry()` function.

I found that the original geometry was modified (either by reading the geometry again or by a read only error message from `deep-freeze`).

We observe mutating code in `EsriJSON.js` : [l.259](https://github.com/openlayers/openlayers/blob/master/src/ol/format/EsriJSON.js#L259), [l.262](https://github.com/openlayers/openlayers/blob/master/src/ol/format/EsriJSON.js#L262) among others.

**Expected behavior**
The input geometry should not be mutated.

**Workaround**
I clone the input geometry, but I find it dirty.

kekel87:
Fix: #9510

### DIFF
--- a/src/ol/format/EsriJSON.js
+++ b/src/ol/format/EsriJSON.js
@@ -256,10 +256,10 @@ function readGeometry(object, opt_options) {
     const rings = convertRings(esriJSONPolygon.rings, layout);
     if (rings.length === 1) {
       type = GeometryType.POLYGON;
-      object['rings'] = rings[0];
+      object = Object.assign({}, object, {['rings']: rings[0]});
     } else {
       type = GeometryType.MULTI_POLYGON;
-      object['rings'] = rings;
+      object = Object.assign({}, object, {['rings']: rings});
     }
   }
   const geometryReader = GEOMETRY_READERS[type];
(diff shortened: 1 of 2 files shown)


src/ol/format/EsriJSON.js
```diff
@@ -256,10 +256,10 @@ function readGeometry(object, opt_options) {
     const rings = convertRings(esriJSONPolygon.rings, layout);
     if (rings.length === 1) {
       type = GeometryType.POLYGON;
-      object['rings'] = rings[0];
+      object = Object.assign({}, object, {rings: rings[0]});
```
tschaub: The `rings` property was quoted here to stop minifiers from renaming it.  It could be worth validating that this is still necessary, but without doing that validation, I would suggest leaving it quoted.

kekel87: That's good, I put quotes.

tschaub:
Thanks @kekel87.

PR #9624, opened 2019-05-31
Remove typecasts in ol/control/OverviewMap

fredj:
The `options` type was not correctly detected by the typescript compiler.
And also remove and unnecessary `bind`

### DIFF
--- a/src/ol/control/OverviewMap.js
+++ b/src/ol/control/OverviewMap.js
@@ -76,7 +76,7 @@ class OverviewMap extends Control {
    */
   constructor(opt_options) {
 
-    const options = opt_options ? opt_options : {};
+    const options = opt_options ? opt_options : /** @type {Options} */ ({});
 
     super({
       element: document.createElement('div'),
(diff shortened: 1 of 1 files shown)


src/ol/control/OverviewMap.js
```diff
@@ -76,7 +76,7 @@ class OverviewMap extends Control {
    */
   constructor(opt_options) {
 
-    const options = opt_options ? opt_options : {};
+    const options = opt_options ? opt_options : /** @type {Options} */ ({});
```
tschaub: No need to change this, but I’m curious if it works to add the type annotation above the options declaration.  I know we have a lot of typecasts, but I keep hoping we could find a way to get rid of them so we can use prettier.

fredj: I've tried and the compiler complains that:
``
Type 'Options | {}' is not assignable to type 'Options'.
  Property 'layers' is missing in type '{}' but required in type 'Options'
``

fredj:
Thanks for the review

src/ol/control/OverviewMap.js
tschaub: That makes sense.  It looks like we require the property in the typedef but then [we don’t actually require it](https://github.com/fredj/openlayers/blob/ede864c823205426bd150f72cb20cbad3f882c72/src/ol/control/OverviewMap.js#L161) in the constructor.  So I think the compiler is making the right complaint.

Instead of forcing the typecast (an empty object is not the same type as an object with a required layers property), we should update the type to reflect that layers is not actually required.  Or, if we do want to make layers required, we would need to change the code.

fredj: Yes, that makes sense once you actually read the error message :-)

See #9633

PR #9634, opened 2019-06-03
Remove old Android workaround

ahocevar:
This pull request removes an old workaround that prevents the `focus` condition to work properly on contemporary devices. On pointer event devices, the [interaction-options.html](https://openlayers.org/en/master/examples/interaction-options.html) example does not allow page srcolling when touching the map, even when the map does not have the focus.


ahocevar:
https://257-4974939-gh.circle-artifacts.com/0/examples/interaction-options.html shows the fixed bahavior on touch devices.

PR #9636, opened 2019-06-03
Remove and move code from ol/webgl

fredj:
Remove some leftover WebGL code from ol/webgl. The `EXTENSIONS`array was moved to the webgl helper class.

@jahow can you please take a look?

### DIFF
--- a/src/ol/webgl.js
+++ b/src/ol/webgl.js
@@ -7,56 +7,6 @@
  * Constants taken from goog.webgl
  */
 
-
-/**
- * @const
- * @type {number}
- */
-export const ONE = 1;
-
-
-/**
- * @const
- * @type {number}
- */
-export const SRC_ALPHA = 0x0302;
-
-
-/**
- * @const
- * @type {number}
- */
-export const COLOR_ATTACHMENT0 = 0x8CE0;
-
-
-/**
- * @const
- * @type {number}
- */
-export const COLOR_BUFFER_BIT = 0x00004000;
-
-
-/**
- * @const
- * @type {number}
- */
-export const TRIANGLES = 0x0004;
-
-
-/**
- * @const
- * @type {number}
- */
-export const TRIANGLE_STRIP = 0x0005;
-
-
-/**
- * @const
- * @type {number}
- */
-export const ONE_MINUS_SRC_ALPHA = 0x0303;
-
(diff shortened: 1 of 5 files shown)


jahow:
Hi,

Removing the `DEBUG` constant is nice since it does not make sense any more. For the rest of the changes, I'm not really 100% for it:
* max texture size and `HAS` aren't used for now but they are still valuable info; the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle would dictate their removal but it's really not that much code and ideally these should be checked at some point (ie giving feedback if webgl is not supported)
* the extensions support does not fit in the Helper class IMO, it is not linked to one specific GL context and as such has its place in the general  `ol/webgl` module

This assumes we're ok with leaving "not used yet" code in there, which is admittedly a not so great practice.

tschaub:
The only value I see in leaving this unused code around is as a source of confusion :).

jahow:
Hm, thinking of it again, webgl support and max texture size may not be that relevant for modern browsers... We might as well get rid of them.
Still, I'd leave the extension code in `ol/webgl`. Thanks @fredj!

fredj:
> Still, I'd leave the extension code in `ol/webgl`. Thanks @fredj!

done, but I'll wait for #9562 before merging

fredj:
@jahow can you please take a second look at this pull request? Thanks

src/ol/webgl.js
```diff
@@ -105,168 +55,12 @@ export const STATIC_DRAW = 0x88E4;
...
- * @type {number}
- */
-export const SCISSOR_TEST = 0x0C11;
-
-
-/**
- * @const
- * @type {number}
- */
-export const UNSIGNED_BYTE = 0x1401;
```
jahow: You should keep `UNSIGNED_INT`, `UNSIGNED_BYTE` and `UNSIGNED_SHORT` too, as they can be used by the helper API here:
https://github.com/openlayers/openlayers/blob/fac16d75c146c8a1ce88bc15c8d445dac9f461c4/src/ol/webgl/Helper.js#L704-L722

fredj: done

fredj:
Thanks for the review

PR #9637, opened 2019-06-03
Detach label cache on Map#setTarget()

ahocevar:
Since our recommended way of cleaning up an `ol/Map` instance is to use `setTarget(null)`, we need to call `detachLabelCache()` in `setTarget()` instead of `dispose()`.


src/ol/PluggableMap.js
```diff
@@ -1326,6 +1325,9 @@ class PluggableMap extends BaseObject {
    */
   setTarget(target) {
     this.set(MapProperty.TARGET, target);
+    if (!target) {
```
fredj: This could go to the `handleTargetChanged_` function instead

ahocevar: Done. Can you take another look please?

PR #9644, opened 2019-06-04
Use compose function from ol/transform

fredj:
Use `compose` function instead of calling `scale`, `rotate` and `translate`


fredj:
I've rounded to numbers to the 15th decimal place and the test is now passing

fredj:
Thanks for the review

PR #9655, opened 2019-06-06
Add hit detection on the WebGL points renderer

jahow:
This PR adds a straightforward implementation for `forEachFeatureAtCoordinate` on the WebGL points renderer:

* when the renderer rebuilds a set of render instructions (list of points with x/y, color, size etc.) it now also generates a set of *hit detection* instructions
* these instructions are almost identical to the rendering ones, except they contain the feature unique ids (given by `getUid`) in the color field; this encoding is done on 4 bytes (RGBA) which allows values up to 4 billion+
* the hit detection instructions are processed by the worker just like the rendering ones, where they are converted to gpu-ready buffers
* on each frame, the hit detection instructions are drawn to an offscreen texture called *render target*; this is done using the new utility class [`RenderTarget`](https://1524-4723248-gh.circle-artifacts.com/0/apidoc/module-ol_webgl_RenderTarget-WebGLRenderTarget.html), which hopefully will be self-explanatory enough
* then, whenever a hit test is required, the renderer simply has to read one pixel inside this render target and decode the uid from color to get the feature from the source.
  Note: the `VectorSource` class was modified to index *all* features based on their unique id; the previous behaviour was only indexing based on the `id` field and on the uid as a fallback

The [icon-sprite-webgl](https://1522-4723248-gh.circle-artifacts.com/0/examples/icon-sprite-webgl.html) example has been updated to allow inspecting features.

Caveat:
* this method only works for detecting the *topmost feature*; there might be ways to find the other features by re-drawing the instructions while skipping the previous hits, but I'm not 100% sure it would work out well; for this use case, using a spatial index is IMO the right way to go (or do backend calls)
* `hitTolerance` isn't supported for now

Some notes on performance:
* Drawing the hit detection instructions on each frame may sound futile but it's actually pretty cheap and much simpler to make sure the hit detection render target contains the right data
* The real performance hit comes from `gl.readPixels` which blocks the main thread to get back the rendered data from the GPU; it usually takes at least 10ms and often more than 20~30ms, and there is no way around it
* The good news though is that this performance cost is fixed and not related to the amount of features :)
* To accommodate for this, it is advised to avoid doing hit detection while the view is moving/animating;

Thanks for any review!


src/ol/source/Vector.js
```diff
@@ -241,11 +241,11 @@ class VectorSource extends Source {
     this.idIndex_ = {};
 
     /**
-     * A lookup of features without id (keyed by getUid(feature)).
+     * A lookup of features by uid (using getUid(feature)).
      * @private
      * @type {!Object<string, import("../Feature.js").default<Geometry>>}
      */
-    this.undefIdIndex_ = {};
+    this.uidIndex_ = {};
```
ahocevar: If you manage to change the color lookup for hit detection to array index, you should not need this change.

Although I like the idea of unconditionally indexing features by uid, I fear that we use too many uids if we do that. Before we decide to make this change, I'd like to see some tests with our vector source examples to see if we end up generating more uids after this change. It might well be that the features have a uid already, but I'm not sure about that.

jahow:
Thanks a lot for the review @ahocevar. I understand your point about encoding array index instead of uid (and hadn't realized that 4 bytes might not be enough).

Currently: *color-encoded uid* is used to retrieve the *feature* object from the source
Possibly: *color-encoded array index* might be used to:
* read the *uid* from the draw instructions array (which does not have a 4 bytes limitation)? this could work but still requires the change on the vector source
* or retrieve the *feature* from the source using `source.getFeatures()[index]`: this would not work as the order is not stable especially with a spatial index (unless I'm mistaken)

I guess I'll go with the first solution by encoding the feature uid in the opacity value, since it's not used for rendering. Also I'll check whether the change on the vector source leads to many more uid's being generated.

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -400,19 +469,43 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
+        tmpCoords[0],
+        tmpCoords[1],
+        u0,
+        v0,
+        u1,
+        v1,
+        size,
+        opacity,
+        rotateWithView,
+        colorEncodeId(parseInt(getUid(feature)), tmpColor)
```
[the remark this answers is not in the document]
jahow: Done: 4340f99792f05a5fd7a8662b49ef566ebe028bf2

src/ol/renderer/webgl/Layer.js
```diff
@@ -81,6 +82,21 @@ class WebGLLayerRenderer extends LayerRenderer {
   getShaderCompileErrors() {
     return this.helper.getShaderCompileErrors();
   }
+
+  /**
+   * @inheritDoc
+   */
+  forEachFeatureAtCoordinate(coordinate, frameState, hitTolerance, callback, declutteredFeatures) {
+    return undefined;
```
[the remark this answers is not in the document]
jahow: Done (amended b4ea8fc0177c52c3047a516f167ba5d5dfbcec02)

examples/icon-sprite-webgl.js
```diff
@@ -121,3 +121,18 @@ new Map({
...
 });
+
+const info = document.getElementById('info');
+map.on('pointermove', function(evt) {
+  if (map.getView().getInteracting()) {
+    return;
+  }
+  const pixel = map.getEventPixel(evt.originalEvent);
+  info.innerText = '';
+  map.forEachFeatureAtPixel(pixel, function(feature) {
```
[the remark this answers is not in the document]
jahow: Done (amended 93d9c68)

jahow:
@ahocevar I've run some tests and logged the `uidCounter_` values in the following examples:
* icon-sprite-webgl: ~160000
* earthquake-clusters: ~3400
* synthetic-points: ~40000

These do not vary significantly with or without the PR. I think it's safe to move forward with the indexing change.

Also with 4340f99 the limit on how many features we can hit detect is not related to the range of generated uids any more.

src/ol/source/Vector.js
jahow: Tests done, see below

jahow:
I've resolved the conflicts. Could someone please give this another look? @fredj maybe? Thanks :)

src/ol/renderer/webgl/Layer.js
```diff
@@ -81,6 +82,14 @@ class WebGLLayerRenderer extends LayerRenderer {
   getShaderCompileErrors() {
     return this.helper.getShaderCompileErrors();
   }
+
+  /**
+   * @inheritDoc
+   */
+  hasFeatureAtCoordinate(coordinate, frameState) {
```
fredj: I think this function is never called: `ol/renderer/Map` has a `hasFeatureAtCoordinate` function that loops over all the renderers and calls `forEachFeatureAtCoordinate`.

fredj: What about: https://github.com/fredj/openlayers/commit/3bf05265d5951fa21f56e7621a594807716b61e0 ?

jahow: Nice catch! I'm adding your commit to the branch, and also remove the implementation in the base layer renderer. Thanks a lot

jahow:
Thanks a lot for taking the time to review.

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -422,12 +518,80 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
+    const width = frameState.size[0];
+    const height = frameState.size[1];
+    const size = [width, height];
+
+    this.hitRenderTarget_.setSize(size);
+
+    this.helper.useProgram(this.hitProgram_);
+    this.helper.prepareDrawToRenderTarget(frameState, this.hitRenderTarget_, true);
 
+    this.helper.makeProjectionTransform(frameState, this.currentTransform_);
```
fredj: this is probably minor but I think this line is not needed: the `this.currentTransform_` is already set in the `prepareFrame` function

jahow: Good catch, thanks
See: https://github.com/openlayers/openlayers/pull/9729

PR #9658, opened 2019-06-06
Don't use HTMLImageElement.decode method on Safari

### LINKED ISSUE #9628: Icons in mapbox-vector-tiles example broken
The icons in the [mapbox-vector-tiles.html](https://openlayers.org/en/master/examples/mapbox-vector-tiles.html) examples, which should appear at high zoom levels, are no longer drawn.

`git bisect` shows that 2e34dd0fafdc40a0bdd2023db55514083b1da87c is the first bad commit.

/cc @fredj

fredj:
The decode function is supported by Safari but not when the image is a svg file.
Upstream bug report: https://bugs.webkit.org/show_bug.cgi?id=198527

fixes #9628

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -6,6 +6,7 @@ import ImageState from './ImageState.js';
 import {listenOnce, unlistenByKey} from './events.js';
 import EventType from './events/EventType.js';
 import {getHeight} from './extent.js';
+import {SAFARI} from './has.js';
 
 
 /**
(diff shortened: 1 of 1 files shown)


src/ol/Image.js
```diff
@@ -157,7 +158,10 @@ class ImageWrapper extends ImageBase {
  */
 export function listenImage(image, loadHandler, errorHandler) {
   const img = /** @type {HTMLImageElement} */ (image);
-  if (img.decode) {
+
+  // The decode function is supported by Safari but not when the image is a svg file.
+  // FIXME: remove `!SAFARI` in the test when this bug is fixed upstream.
```
tschaub: It could be useful to link to the bug here.

PR #9659, opened 2019-06-06
Take the view rotation into account in box-selection example

### LINKED ISSUE #5496: Box Selection example does not work correctly if map is the rotated
1. Open [Box Selection example](http://openlayers.org/en/latest/examples/box-selection.html)
2. Rotate the map (ALT+SHIFT+Drag)
3. Make a selection (CTRL+Drag)
   boxselect1
4. Some features outside of box might be selected
   boxselect2

### LINKED ISSUE #9656: Handle rotation in the Box Selection example
**Describe the bug**
The stale bot has closed the very old issue #5496 but it would be nice ti have an example which handled rotation in a way most might expect it to.

**To Reproduce**
See #5496

**Expected behavior**
https://codesandbox.io/s/boxselection-5cc8v

fredj:
Fixes #5496
Fixes #9656

And mark the rotation anchor as optional in ol/geom/Geometry class. The center of the geometry extent as the default value was already implemented for the `scale` method.

### DIFF
--- a/examples/box-selection.js
+++ b/examples/box-selection.js
@@ -45,8 +45,10 @@ map.addInteraction(dragBox);
 dragBox.on('boxend', function() {
   // features that intersect the box are added to the collection of
   // selected features
-  const extent = dragBox.getGeometry().getExtent();
-  vectorSource.forEachFeatureIntersectingExtent(extent, function(feature) {
+  const geometry = dragBox.getGeometry().clone();
+  geometry.rotate(-map.getView().getRotation());
+
+  vectorSource.forEachFeatureIntersectingExtent(geometry.getExtent(), function(feature) {
     selectedFeatures.push(feature);
   });
 });
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -150,10 +150,11 @@ class Geometry extends BaseObject {
    * coordinates in place.
    * @abstract
    * @param {number} angle Rotation angle in radians.
-   * @param {import("../coordinate.js").Coordinate} anchor The rotation center.
+   * @param {import("../coordinate.js").Coordinate=} opt_anchor The rotation center.
+   *     Defaults to the center of the geometry extent.
    * @api
    */
-  rotate(angle, anchor) {
+  rotate(angle, opt_anchor) {
     abstract();
   }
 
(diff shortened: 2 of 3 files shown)


jahow:
Hi,

I'm not sure your implementation is the right one, see behaviour below:
Peek 07-06-2019 09-27


Maybe the intersection check should be done using the actual selection box instead of its extent?

mike-000:
> Hi,
> 
> I'm not sure your implementation is the right one, see behaviour below:
> Peek 07-06-2019 09-27
> 
> Maybe the intersection check should be done using the actual selection box instead of its extent?

In #9656 I rotate the DragBox geometry so its extent is minimised (i.e. equal to the geometry with no excess due to rotation) but it is also essential to rotate each of the country geometries (after cloning) around the same point before checking if the they intersect the revised extent.  That part is missing from here.

jahow:
> In #9656 I rotate the DragBox geometry so its extent is minimised (i.e. equal to the geometry with no excess due to rotation) but it is also essential to rotate each of the country geometries (after cloning) around the same point before checking if the they intersect the revised extent. That part is missing from here.

Indeed, although it sounds like creating that many object on every selection might have a heavy cost depending on the amount of features. How about using `selectionBoxGeometry.intersectsExtent` on each feature in the source? At least we won't be cloning anything.

mike-000:
> > In #9656 I rotate the DragBox geometry so its extent is minimised (i.e. equal to the geometry with no excess due to rotation) but it is also essential to rotate each of the country geometries (after cloning) around the same point before checking if the they intersect the revised extent. That part is missing from here.
> 
> Indeed, although it sounds like creating that many object on every selection might have a heavy cost depending on the amount of features. How about using `selectionBoxGeometry.intersectsExtent` on each feature in the source? At least we won't be cloning anything.

Using the extent of the countries geometry instead of the geometry would produce many unexpected results.  e.g. most of Canada, Mexico and the Caribbean are inside the extent of the USA (because of Alaska and Hawaii, and most of Sweden, Finland and Estonia are inside the extent of Noway because of Norway's shape, so even when selecting well outside their territory countries like USA, Norway, Brazil, Russia, etc. would be incorrectly selected.

mike-000:
I've created a new and more efficient codesandbox as it is not necessary to clone every feature in the source.  The features selected in the unadjusted extent of the box will include all the valid features plus, when the view is rotated, some incorrect candidates.  Only those need cloning for a more accurate check.  https://codesandbox.io/s/box-selection-krl5z

PR #9672, opened 2019-06-12
Fix for workers in a leagacy build

Kai-W:
Fix for the new workers and the leagacy build.

An alternative would be:
return new Worker(url, {type: 'module'});
But for now this is in Chrome hidden behind the --enable-experimental-web-platform-features flag

### DIFF
--- a/tasks/serialize-workers.js
+++ b/tasks/serialize-workers.js
@@ -52,7 +52,7 @@ async function build(input, {minify = true} = {}) {
   });
 
   const bundle = await rollup.rollup({input, plugins});
-  const {output} = await bundle.generate({format: 'es'});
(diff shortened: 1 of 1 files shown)


tschaub:
Not in a place where I can dig further, but it feels to me like this is an issue with generating the legacy build, not an issue with generating the worker module(s).

Kai-W:
If im understand the worker creation correctly, it probably can be done with a custom webpack loader. 
The loader has to load the prepared Workers, get the Inline String, transpile it, and build the worker again.
I'm quite new to rollup/webpack so i tried to fix the first generation for every use case.

ahocevar:
I agree with @tschaub that the legacy build should be the place to fix this.

PR #9677, opened 2019-06-14
Add support for tile pixel projection in feature formats

ahocevar:
This pull request fixes the geojson-vt example, which is currently broken. To achieve this, I added support for tile pixel projections in `ol/format/Feature`, like we already have and use for `ol/format/MVT`.


src/ol/format/Feature.js
```diff
@@ -3,7 +3,7 @@
  */
 import {assign} from '../obj.js';
 import {abstract} from '../util.js';
-import {get as getProjection, equivalent as equivalentProjection, transformExtent} from '../proj.js';
+import {get as getProjection, equivalent as equivalentProjection, transformExtent, get} from '../proj.js';
```
tschaub: Looks like we already import `get` here.

ahocevar: Good catch. Fixed.

PR #9684, opened 2019-06-17
List complete classes in doc navigation

ahocevar:
#9681 removed all class members and methods from the navigation bar. This pull request adds them back, by putting the classes in a hierarchy below the modules. This change makes the search field work again for class members and methods.


ahocevar:
API docs after this pull request: https://271-4974939-gh.circle-artifacts.com/0/apidoc/index.html

PR #9686, opened 2019-06-17
Don't listen for rotation in OverviewMap control

fredj:
Code simplification in the OverviewMap control.

Instead of listening for rotation change events, rotate the view in the `updateBox_`function


fredj:
need more work

PR #9694, opened 2019-06-19
change import syntax to support client, ES6 javascript import 'nobare'

JerYme:
changes allows the js to run directly in any webbrowser (avoiding issues on the import bare syntax not correctly supported on the client browser side)

- change import syntax to support client, ES6 javascript import 'nobare'
- include external lib rbush converted to ES6 style


tschaub:
See https://github.com/openlayers/openlayers/pull/9615.

For bare imports, I’m interested in seeing if [import maps](https://wicg.github.io/import-maps/) can help.

PR #9699, opened 2019-06-21
Fix MapTiler API usage

petrsloup:
This PR "fixes" the usage of MapTiler API (previously called TileHosting):

- Fixed "get your own key" links -- the examples use MapTiler infrastructure, but were saying "get your own Mapbox API key", but the Mapbox keys are obviously not usable with the MapTiler API URLs :-)
- New URLs (api.maptiler.com instead of maps.tilehosting.com) (https://www.maptiler.com/blog/2019/04/maptiler-cloud-infrastructure-upgrade.html). The old URLs are now obsolete and redirect to the new domain.

@ahocevar I see the example now use your personal free key (which has 100k requests/month limit). If needed, maybe we could provide a special key with higher limits for the OpenLayers examples.

### DIFF
--- a/examples/mapbox-layer.html
+++ b/examples/mapbox-layer.html
@@ -3,13 +3,13 @@ layout: example.html
 title: Mapbox-gl Layer
 shortdesc: Example of a Mapbox-gl-js layer integration.
 docs: >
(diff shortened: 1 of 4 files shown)


tschaub:
Thanks @petrsloup.  Having a dedicated key for the examples would be great.

PR #9701, opened 2019-06-21
Remove will-change CSS rule for Overlay container

### LINKED ISSUE #9139: master 'Icon Symbolizer'

### LINKED ISSUE #9467: Popup in example no longer moves with the map

fredj:
Fixes #9467
Fixes #9139

Chrome doesn't force a composition if the `will-change` CSS rule is different from `transform` or `opacity`.
See https://bugs.chromium.org/p/chromium/issues/detail?id=960953

I propose to remove the `will-change` rule for the overlays,


tschaub:
Thanks for figuring this out @fredj.

Issue #9692, opened 2019-06-19
Fix TypeScript check errors


ahocevar:
We currently have TypeScript check errors in the following files:
```
    "control/FullScreen.js",
    "format/EsriJSON.js",
    "format/GMLBase.js",
    "format/GPX.js",
    "format/GML.js",
    "format/GML2.js",
    "format/GML3.js",
    "format/GML32.js",
    "format/WMSGetFeatureInfo.js",
    "format/IIIFInfo.js",
    "format/KML.js",
    "format/OSMXML.js",
    "format/OWS.js",
    "format/TopoJSON.js",
    "format/WMSCapabilities.js",
    "format/WMTSCapabilities.js",
    "interaction/DragPan.js",
    "interaction/MouseWheelZoom.js",
    "renderer/canvas/VectorImageLayer.js",
    "renderer/webgl/**/*.js",
    "source/IIIF.js"
```
I know that some of those in XML formats cannot be fixed due to TypeScript issues, but I'd appreciate any community help in fixing the others. Any takers? @KaiVolland maybe?

ahocevar:
To see the issues, run `npm run typecheck`.

lutzhelm:
I will take another look at my contributions in `format/IIIFInfo.js` and `source/IIIF.js`.

Part of the problem in `format/IIIFInfo.js` is that I don't know any other way to specify JSON-LD keywords like `@id` or `@context` as object `@property` in any other way than with quotation marks.

KaiVolland:
I'll have a quick look at it.

KaiVolland:
I was able to solve some smaller issues: #9700 
Sadly i have no solution for the `T` issuess.

lutzhelm:
Problems in `source/IIIF.js` can be completely solved with #9702.
The five remaining problems in `formats/IIIFInfo.js` are JSON-LD keyword related as described [above](#issuecomment-504330304).

PR #9702, opened 2019-06-21
Fix several TypeScript type check problems

lutzhelm:
Solves several type check problems mentioned in #9692 for `source/IIIF.js`.


tschaub:
Thanks @lutzhelm.

PR #9710, opened 2019-06-24
Additional Typescript compliance fixes

jahow:
Related to #9692

Some more typing fixes for:
* `MouseWheelZoom` interaction: using `any` because we rely on [`MouseWheelEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MouseWheelEvent) which is non standard
* `CanvasVectorImageLayerRenderer`
* web workers: the `webworker` lib was removed from tsconfig.json because it was causing conflicts; also, in worker sources, the type of the context had to be explicitly defined to avoid conflict with the `window.postMessage` function

With this PR the only remaining typecheck errors are on the formats (unless I missed something).

### DIFF
--- a/src/ol/worker/version.js
+++ b/src/ol/worker/version.js
@@ -4,9 +4,12 @@
  */
 import {VERSION} from '../util.js';
 
-onmessage = event => {
+/** @type {any} */
+const worker = self;
+
+worker.onmessage = event => {
   console.log('version worker received message:', event.data); // eslint-disable-line
-  postMessage(`version: ${VERSION}`);
+  worker.postMessage(`version: ${VERSION}`);
 };
 
 export let create;
--- a/src/ol/worker/webgl.js
+++ b/src/ol/worker/webgl.js
@@ -10,7 +10,10 @@ import {
 } from '../renderer/webgl/Layer.js';
 import {assign} from '../obj.js';
 
-onmessage = event => {
+/** @type {any} */
+const worker = self;
+
+worker.onmessage = event => {
   const received = event.data;
   if (received.type === WebGLWorkerMessageType.GENERATE_BUFFERS) {
     const renderInstructions = new Float32Array(received.renderInstructions);
@@ -41,7 +44,7 @@ onmessage = event => {
       renderInstructions: renderInstructions.buffer
     }, received);
 
-    postMessage(message, [vertexBuffer.buffer, indexBuffer.buffer, renderInstructions.buffer]);
+    worker.postMessage(message, [vertexBuffer.buffer, indexBuffer.buffer, renderInstructions.buffer]);
   }
 };
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     /* Basic Options */
     "target": "ES2017",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
-    "lib": ["es2017", "dom", "webworker"],    /* Specify library files to be included in the compilation. */
+    "lib": ["es2017", "dom"],                 /* Specify library files to be included in the compilation. */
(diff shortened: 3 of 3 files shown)


src/ol/renderer/canvas/VectorImageLayer.js
```diff
@@ -103,9 +103,13 @@ class CanvasVectorImageLayerRenderer extends CanvasImageLayerRenderer {
       });
 
       listen(image, EventType.CHANGE, function() {
+        /**
+         * @type {CanvasVectorImageLayerRenderer}
+         */
+        const this_ = this;
```
fredj: instead of this, we could use `bind(this)`:
```js
listen(image, EventType.CHANGE, function() {
  if (image.getState() === ImageState.LOADED) {
    this.image_ = image;
    this.skippedFeatures_ = skippedFeatures;
  }
}.bind(this);
```
I think it's prettier than creating this `this_` variable

sbrunner: We can also use an @this to type the this...

jahow: I've tried but typescript did not seem to take this into account.

src/ol/worker/version.js
```diff
@@ -4,9 +4,14 @@
  */
 import {VERSION} from '../util.js';
 
-onmessage = event => {
+/**
+ * @type {Worker}
+ */
+const worker = /** @type {any} */ (self);
```
sbrunner: For me it's a like hackish but required, Shouldn't we add a comment like:?
```
Workaround, see https://github.com/Microsoft/TypeScript/issues/20595 form more information
```

fredj: The problem is that now `worker` type is `any` (at least for intellisense).
What about this: https://github.com/openlayers/openlayers/compare/master...fredj:webworker_jsconfig ?

It move the `webworker` lib only for the files inside the `src/ol/worker` directory

jahow: Thanks! I've tried that but to no avail. Will check your branch & update the PR.

fredj: not sure if it was clear but the commit in my branch replaces this one

jahow: > It move the webworker lib only for the files inside the src/ol/worker directory

Did not work on my side. I've turned the problem around many times but just could not get tsc to use the webworker lib only for the `worker` folder, and the dom lib for the rest.. In the end I went with a manual extraction of the worker scope typings, see: https://github.com/openlayers/openlayers/pull/9710/commits/ae48809b252248bbc64f062faca90a7307921586

Hope that'll be fine, otherwise you're welcome to try other things on this branch ;)

jahow:
Thanks for your feedback @fredj and @sbrunner. I have tried to improve this PR, hopefull we'll get rid of those typecheck errors soon.

jahow:
Hi, maybe @sbrunner and @fredj could have a quick look at this PR again and confirm that this is good to go? Thanks a lot

fredj:
@jahow I don't understand why `webworker.d.ts` was added to the types directory. It is not used in the workers because `self` is cast to `any`.

What not just keep the `any` cast, remove `webworker.d.ts`and revisit it when https://github.com/Microsoft/TypeScript/issues/20595 is addressed?

jahow:
> @jahow I don't understand why `webworker.d.ts` was added to the types directory. It is not used in the workers because `self` is cast to `any`.

It is cast to `any`, and then to `DedicatedWorkerGlobalScope`, this in order to still kinda benefit from typechecking in the library code. Although...

> What not just keep the any cast, remove webworker.d.tsand revisit it when microsoft/TypeScript#20595 is addressed?

I agree that it's not worth it to clutter the code base with a temporary workaround, for something that is not really that critical. I'll modify the PR to only keep the any cast and remove the definition type.

KaiVolland:
@jahow Are you going to pick up the work with this?

jahow:
@fredj this PR is now rebased and much simpler. Can you please give it a last review? Thanks!

fredj:
@jahow what about the fixes for `MouseWheelZoom` and `CanvasVectorImageLayerRenderer`?

jahow:
Not required anymore, these parts of the code evolved along the way :)

PR #9719, opened 2019-06-25
Use special MapTiler API key dedicated for OpenLayers examples

petrsloup:
("continuation" of #9699)

This PR replaces @ahocevar's personal MapTiler API key with a special one with higher limits.
The key is intended only for OpenLayers examples so it clearly indicates "get your own" right in the key itself to make it even more clear to potential copy-pasters.


tschaub:
Awesome.  Thanks @petrsloup!

PR #9723, opened 2019-06-27
Pass the opt_direction value to the getConstrainedResolution function

fredj:
The parameter was not used

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1472,7 +1472,7 @@ class View extends BaseObject {
    */
   getConstrainedZoom(targetZoom, opt_direction) {
     const targetRes = this.getResolutionForZoom(targetZoom);
-    return this.getZoomForResolution(this.getConstrainedResolution(targetRes));
+    return this.getZoomForResolution(this.getConstrainedResolution(targetRes, opt_direction));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/ol/View.js
```diff
@@ -1372,7 +1372,7 @@ class View extends BaseObject {
    */
   getConstrainedZoom(targetZoom, opt_direction) {
     const targetRes = this.getResolutionForZoom(targetZoom);
-    return this.getZoomForResolution(this.getConstrainedResolution(targetRes));
+    return this.getZoomForResolution(this.getConstrainedResolution(targetRes, opt_direction));
```
jahow: I'm unsure how this parameter is supposed to be used. Maybe it should be inverted, since the zoom level increases when the resolution decreases? But it's not tested so not 100% positive on this.

fredj: And this `opt_direction` param is never used in the library so maybe it should be removed?

jahow: Depends if it was part of the API in 5.x, and also it makes sense to have such an option for this method... sorry for ruining the fun!

ahocevar:
Related to #9812. At least you can copy/paste a meaningful description for the option from there.

KaiVolland:
@fredj do you have time to work on this or would you prefer if someone else would finish this up?

fredj:
I've pushed a commit that changes the description for the `opt_direction` param.

PR #9729, opened 2019-06-28
Remove useless matrix operation in Webgl points renderer

jahow:
Remove a useless generation of a projection matrix on hit detection render pass.

Fix for this comment: https://github.com/openlayers/openlayers/pull/9655/files/933a6297bb0cad35026f9e3e1f8ea2a4738059c2#r298574950


fredj:
Maybe a miss something but I don't see were the "hit" transforms are used: I've tried to remove them and it still works, see https://github.com/openlayers/openlayers/commit/bf026c26db747014d8690146f951cb304c2ca6aa

fredj:
see #9748

PR #9745, opened 2019-07-02
Simplify heatmap's shaders

fredj:
Some small simplifications to the heatmap's shaders.


fredj:
> Thanks a lot! now I feel bad for leaving all this behind.

please don't feel bad for this !

Issue #9736, opened 2019-06-29
WebGL hit detection transform does not take pixel ratio into account


ahocevar:
When trying WebGL hit detection with the icon-sprite-webgl.html example on a Retina display (pixel ratio: 2), I noticed that hovered icons and hit detections do not align. With a pixel ratio of 1, hit detection works as expected.

jahow:
This probably has to do with the fact we're not looking up the right pixel here:
https://github.com/openlayers/openlayers/blob/32d1d9214cf8510b9b9870955eec2c4800c3d205/src/ol/renderer/webgl/PointsLayer.js#L545

Will try to look into this ASAP. Thanks for the issue

PR #9748, opened 2019-07-03
Remove hit detection transforms from ol/renderer/webgl/PointsLayer

fredj:
replaces #9729


tschaub:
Should these transforms be used to address #9736 (or is this change otherwise related to that problem)?

jahow:
> Should these transforms be used to address #9736 (or is this change otherwise related to that problem)?

I think transforms already take the pixel ratio into account so the fix for this lies elsewhere IMO.

PR #9751, opened 2019-07-03
Add minArea typedef and ctor prop in DragZoom

### LINKED ISSUE #9746: Acces to DragBox's minArea property from DragZoom
Hello,

For my current project, i use the DragZoom object. I noticed that the dragEnd event is fired only when the drag zone is bigger than the minArea property. But as DragZoom extends DragBox, this property is not accessible from the DragZoom object because this object's constructor had only 3 parent 's properties that can be set.. 
It is an wanted behavior ? if yes can you explain me why please ? else can i change it by just add this property to the DragZoom constructor's options ?

https://github.com/openlayers/openlayers/blob/ac3072d88836a1d8510c484aceff6e0a5f56100f/src/ol/interaction/DragZoom.js#L43

Thank you

logan-bounet:
close #9746


fredj:
Thanks for your contribution @logan-bounet !

Issue #9665, opened 2019-06-10
Export PDF example only showing part of expected results


mike-000:
**Describe the bug**
When using Chrome the Export PDF example is only showing the top left part of the original view extent on the pdf, as if html-to-image is using the original size instead of the one set using setSize().  I seem to remember it working correctly since the change to html-to-image so it might be a browser specific issue.

**To Reproduce**
Simply open the example and click on Export

![image](https://user-images.githubusercontent.com/49240900/59207973-0bc50400-8ba0-11e9-94ae-1147a5a5f9e2.png)

Then view the PDF

![image](https://user-images.githubusercontent.com/49240900/59208003-1ed7d400-8ba0-11e9-9a57-6da822a43436.png)

**Expected behavior**
The output should include all of the view extent.

mike-000:
The map div isn't resizing when the map resizes, so `html-to-image` on the map div doesn't work as `dom-to-image` did on the layers container (where excess from rotated canvas needs to be cropped).  However `html-to-image` isn't supported by all browsers so a more browser compatible approach may be to combine only the canvases to get a canvas only export similar to OL5 (without an option for the Null Island overlay as in the OL6 PNG Map Export example) https://codesandbox.io/s/export-pdf-4osj4

PR #9754, opened 2019-07-04
Export PDF example extent corrections

### LINKED ISSUE #9460: Export PDF extent problems when view is rotated
**Describe the bug**
There is another issue with the Export PDF example in addition to https://github.com/openlayers/openlayers/issues/9436 which can also be seen in earlier versions.  When the view is rotated the extent exported doesn't match the view and the view isn't correctly restored to its original state.  It's because `calculateExtent()` is returning a **bounding** extent which for `fit()` is the extent **to be bounded**.  Turning off rotation, applying the fit then reapplying rotation will fix both the version 5 and version 6 examples:

```
  var rotation = map.getView().getRotation();
  map.getView().setRotation(0);
  var extent = map.getView().calculateExtent(size);
 
  map.once('rendercomplete', function () {

    ...
    
      // Reset original map size
      map.setSize(size);
      map.getView().setRotation(0);
      map.getView().fit(extent, {size: size});
      map.getView().setRotation(rotation);

    ...

  });

  // Set print size
  var printSize = [width, height];
  map.setSize(printSize);
  map.getView().fit(extent, {size: printSize});
  map.getView().setRotation(rotation);
```

mike-000:
Fixes #9460 and #9665

Replace `calculateExtent` and `fit` with `getResolution` and `setResolution` as the former do not produce the desired result when the view is rotated.

Export the viewport instead of the map div (which doesn't resize) and specify the width and height in the export options

This is a minimal fix for the two issues.  Further enhancement including a fallback to export layer canvases where the browser doesn't support html-to-image could be done separately if necessary.

### DIFF
--- a/examples/export-pdf.js
+++ b/examples/export-pdf.js
@@ -65,18 +65,18 @@ exportButton.addEventListener('click', function() {
   const width = Math.round(dim[0] * resolution / 25.4);
   const height = Math.round(dim[1] * resolution / 25.4);
   const size = map.getSize();
-  const extent = map.getView().calculateExtent(size);
+  const viewResolution = map.getView().getResolution();
 
   map.once('rendercomplete', function() {
-    toJpeg(map.getTargetElement(), exportOptions).then(function(dataUrl) {
+    exportOptions.width = width;
+    exportOptions.height = height;
+    toJpeg(map.getViewport(), exportOptions).then(function(dataUrl) {
       const pdf = new jsPDF('landscape', undefined, format);
       pdf.addImage(dataUrl, 'JPEG', 0, 0, dim[0], dim[1]);
       pdf.save('map.pdf');
       // Reset original map size
       map.setSize(size);
-      map.getView().fit(extent, {
-        size: size
-      });
+      map.getView().setResolution(viewResolution);
       exportButton.disabled = false;
       document.body.style.cursor = 'auto';
     });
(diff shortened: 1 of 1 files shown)


ahocevar:
As far as I can tell, the dpi are not correct any more after your changes. Compare the generated PDF from https://openlayers.org/en/master/examples/export-pdf.html and https://1808-4723248-gh.circle-artifacts.com/0/examples/export-pdf.html, without changing the center, zoom or rotation.

mike-000:
> As far as I can tell, the dpi are not correct any more after your changes. Compare the generated PDF from https://openlayers.org/en/master/examples/export-pdf.html and https://1808-4723248-gh.circle-artifacts.com/0/examples/export-pdf.html, without changing the center, zoom or rotation.

Compare also with https://openlayers.org/en/latest/examples/export-pdf.html and set dpi to 300 in all three examples to make the different behaviours more obvious (but make other changes).

The OL5 example prints slightly more than the initial view area in both dimensions (due to the `.fit` not specifying `constrainResolution: false`) - slightly wrong and subsequently fixed but never made a full release before becoming obsolete.

The current master prints much less than the initial view area (due to #9665) - at 72dpi it might not be obvious but at 300dpi it's very wrong.   I initially though #9665 might be a browser issue but I get the same result with Chrome and Edge.

My version prints either the exact height (if the view aspect ratio is narrower than the paper) or width (if view aspect raio is wider than paper) of the initial view area plus whatever extra of the other dimension is needed to fill the paper - equivalent to an unconstrained `fit` but done using `setResolution` as `fit` doesn't work as desired when there is rotation).

ahocevar:
Thanks for the explanation!

PR #9762, opened 2019-07-08
Get legend request

### LINKED ISSUE #9163: GetLegend requests

simonseyock:
Fixes #9163 

It is based on @KaiVolland s work #9398

### DIFF
--- a/src/ol/source/ImageWMS.js
+++ b/src/ol/source/ImageWMS.js
@@ -192,6 +192,40 @@ class ImageWMS extends ImageSource {
   }
 
   /**
+   * Return the GetLegendGraphic URL for the passed resolution.
+   * Return `undefined` if the GetLegendGraphic URL cannot be constructed.
+   * @param {number} resolution Resolution.
+   * @param {!Object} params GetLegendGraphic params. Default `FORMAT` is
+   *     `image/png`. `VERSION` should not be specified here.
+   * @return {string|undefined} GetLegendGraphic URL.
+   * @api
+   */
+  getGetLegendGraphicUrl(resolution, params) {
+    const layers = this.params_.LAYERS;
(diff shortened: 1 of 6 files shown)


simonseyock:
@ahocevar what do you think of this pull request?

src/ol/source/ImageWMS.js
```diff
@@ -191,6 +191,43 @@ class ImageWMS extends ImageSource {
       1, sourceProjectionObj || projectionObj, baseParams);
   }
 
+  /**
+   * Return the GetLegendGraphic URL for the passed resolution.
+   * Return `undefined` if the GetLegendGraphic URL cannot be constructed.
+   * @param {!number} resolution Resolution. If set to undefined `SCALE`
```
jahow: Optional parameters should be using the `{Type} [name]` syntax now, I know this is not consistent with the rest of the file but this is the correct JSDoc syntax nowadays.

See for example this recently merged PR: https://github.com/openlayers/openlayers/pull/9751

marcjansen: I'll do this in a follow-up. Thanks for you review.

marcjansen:
Thanks for the work on this @KaiVolland , @simonseyock and @jahow.

src/ol/source/ImageWMS.js
marcjansen: See https://github.com/openlayers/openlayers/pull/9792

PR #9774, opened 2019-07-14
Fix observable properties documentation

### LINKED ISSUE #9607: Missing events in API documentation for View
**Describe the bug**
`change:center` `change:resolution` and `change:rotation` are no longer listed in the master documentation https://openlayers.org/en/master/apidoc/module-ol_View-View.html As they are still working and it's inconsistent with similar change events in https://openlayers.org/en/master/apidoc/module-ol_Map-Map.html it's presumably unintended and maybe related to https://github.com/openlayers/openlayers/issues/9471.  Also "An View" is a grammar error.

### LINKED ISSUE #9773: Observable properties missing from from some classes in API docs
The API documentation is missing observable properties and `change:*` events for getters and setters marked as `@observable`. The strange thing is that some classes (e.g. `ol/Map~Map`) show them, but some (e.g. `ol/View~View`) don't.

ahocevar:
Like with other "fixes" around our JSDoc setup, this is merely a workaround for obscure things that happen in JSDoc internally.

Fixes #9773. 
Fixes #9607.

### DIFF
--- a/config/jsdoc/api/plugins/api.js
+++ b/config/jsdoc/api/plugins/api.js
@@ -31,6 +31,24 @@ function hasApiMembers(doclet) {
 }
 
 function includeAugments(doclet) {
+  // Make sure that `observables` and `fires` are taken from an already processed `class` doclet.
+  // This is necessary because JSDoc generates multiple doclets with the same longname.
+  const cls = classes[doclet.longname];
+  if (cls.observables && !doclet.observables) {
+    doclet.observables = cls.observables;
+  }
+  if (doclet.fires && cls.fires) {
+    for (let i = 0, ii = cls.fires.length; i < ii; ++i) {
+      const fires = cls.fires[i];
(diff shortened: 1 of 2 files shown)


config/jsdoc/api/plugins/api.js
```diff
@@ -148,7 +166,7 @@ exports.handlers = {
         // constructor from the docs.
         doclet._hideConstructor = true;
         includeAugments(doclet);
-      } else if (doclet.undocumented !== false && !doclet._hideConstructor && !(doclet.kind == 'typedef' && doclet.longname in types)) {
+      } else if (!doclet._hideConstructor && !(doclet.kind == 'typedef' && doclet.longname in types)) {
```
ahocevar: This change was only for readability, because `doclet.undocumented` is either `true` or `undefined`, but never `false`.

PR #9778, opened 2019-07-16
Return existing source tiles if at target resolution

ahocevar:
This pull request reduces the number of vector tile requests by using the existing source tiles unconditionally when they have the target resolution already. It is also not necessary to check for an url tile coord, because we no longer enter the `getSourceTiles()` function in this case.


ahocevar:
Recommended review setting: https://github.com/openlayers/openlayers/pull/9778/files?utf8=✓&diff=unified&w=1 (whitespaces hidden).

PR #9375, opened 2019-03-25
Replaces reproj image.tests.js and tile.tests.js

KaiVolland:
Transforms the old rendering tests for `ReprojImage` and `ReprojTile` to the new rendering test approach.

PR #9794, opened 2019-07-23
Fix linting error (import extension missing)

marcjansen:
This makes linting work again. It was broken in the aftermath of #9375 

Before:
```
$ npm run lint

> ol@6.0.0-beta.11 lint /home/mjansen/code/openlayers
> eslint tasks test rendering src/ol examples config


/home/mjansen/code/openlayers/rendering/cases/reproj-image/main.js
  7:8  error  Missing file extension "js" for "../../../src/ol/proj"  import/extensions

✖ 1 problem (1 error, 0 warnings)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! ol@6.0.0-beta.11 lint: `eslint tasks test rendering src/ol examples config`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the ol@6.0.0-beta.11 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/mjansen/.npm/_logs/2019-07-23T09_41_35_294Z-debug.log
```

This PR removes the above error.

Please review.

### DIFF
--- a/rendering/cases/reproj-image/main.js
+++ b/rendering/cases/reproj-image/main.js
@@ -26,4 +26,6 @@ new Map({
   })
 });
 
-render();
+render({
+  tolerance: 0.001
+});
--- a/rendering/cases/reproj-image/main.js
+++ b/rendering/cases/reproj-image/main.js
@@ -4,7 +4,7 @@ import Static from '../../../src/ol/source/ImageStatic.js';
 import {
   get as getProjection,
   transformExtent
-} from '../../../src/ol/proj';
+} from '../../../src/ol/proj.js';
 import ImageLayer from '../../../src/ol/layer/Image.js';
 
 const source = new Static({


marcjansen:
It seems as if the build now fails due to the changes of the rendering tests from #9375. Hmm.

ahocevar:
case ./cases/reproj-image/main.js': mismatch 0.000

Looks like the `reproj-image/` render test case needs a minimal tolerance, e.g.
```js
render({
   tolerance: 0.001
});

marcjansen:
> ```
> case ./cases/reproj-image/main.js': mismatch 0.000
> ```
> 
> Looks like the `reproj-image/` render test case needs a minimal tolerance, e.g.
> 
> ```js
> render({
>    tolerance: 0.001
> });
> ```

Thanks, but this isn't entirely fixing the build for me. I'll try to make sense of this.

ahocevar:
@marcjansen I took the liberty to add a commit with the change I suggested, and as you can see the build passes now.

If it does not pass on your local machine, that's not unusual: different browser render engines result in different pixel results. We need to keep the tolerances as small as possible, otherwise true rendering problems may slip through. So all we should do is make the tests pass on the CI, and you can always look at the images there in the artifacts section on CircleCI.

marcjansen:
Thanks for the clarification, the suggestion and the commit😊

PR #9804, opened 2019-07-26
Replacement for the Shared Views example

### LINKED ISSUE #9799: Resurrect the Shared View example
The previous example was lost as a result of https://github.com/openlayers/openlayers/commit/7fa2189fe997a87f28e9551242a71e1da211e5cc#diff-bfebe34154a0dfd9fc7b447fc9ed74e9  Obviously WebGL is no longer applicable but a side by side example of the use of a shared view such as for road and aerial would be useful.

mike-000:
Fixes #9799

A replacement side-by-shared Shared Views example derived from the previous canvas/webGL maps example but substituting Road and Aerial maps from the Bing Maps example.

### DIFF
--- a/examples/side-by-side.css
+++ b/examples/side-by-side.css
@@ -1,0 +1,7 @@
+@media (min-width: 800px) {
+  .half {
+    padding: 0 10px;
+    width: 50%;
+    float: left;
+  }
+}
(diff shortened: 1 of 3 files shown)


marcjansen:
Thanks for this.

marcjansen:
I just now see that I cannot combine / squash all the commits into one single commit.

I'll merge this, but for future PRs (of which I hope there will be some) would ask you to combine many commits into one single meaningful commit.

Thanks nonetheless.

PR #9807, opened 2019-07-29
Store the state of the ScaleLine control

### LINKED ISSUE #9725: Scale Line getCurrentScale Function
I would like to display the current scale outside of the Scale line control. This way I can create a grid layer that creates lines based on the scale line control. I could recreate the logic inside the scale line control to do that, but it seems much easier to just have access to that number. 

I would like to add a getter function that gets the scale distance from the scale line control to be used elsewhere.

fredj:
Provides the state of the scale line control: width in pixels, count and unit suffix

Fixes #9725

### DIFF
--- a/src/ol/control/ScaleLine.js
+++ b/src/ol/control/ScaleLine.js
@@ -102,6 +102,15 @@ class ScaleLine extends Control {
     this.viewState_ = null;
 
     /**
+     * @type {{width: number, count: number, suffix: string}}
+     */
+    this.state = {
+      width: undefined,
+      count: undefined,
+      suffix: undefined
+    };
+
+    /**
      * @private
      * @type {number}
      */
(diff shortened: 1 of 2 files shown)


ahocevar:
Without this being documented, I think it is not very useful. Even if documented, the width of the scale bar does not seem to be of great value to me, but I may be wrong.

Eitehr way, the OP of #9725 could achieve what they want with
```js
const width = parseFloat(document.querySelector('.ol-scale-bar-inner').style.width)
```

mike-000:
count and units can also be obtained from element's `textContent` https://codesandbox.io/s/scalerings-41dzr

jahow:
@fredj what should we do with this PR? looks like it may not be necessary to merge it.

fredj:
it was for a customer project but it's not needed anymore. closing

PR #9823, opened 2019-08-06
Enable regeneration of image in ol.style.RegularShape

### LINKED ISSUE #9820: Enable regeneration of image in ol.style.RegularShape
### The problem
There's semantic error with `ol.style.RegularShape`.
(refer to #7398, note that `Circle` is subclass of `RegularShape`)

### Example
Although you can change a color of `Fill` of `RegularShape`, it doesn't regenerate image. This made some developers like me debug whole day long.

### Detail
`RegularShape` has no way to receive event when properties of its `Fill` or `Stroke` instance has changed.

### Considerations (these are just my thoughts)
- Making `Fill` and `Stroke` extend `BaseObject` (for event subscription) might cause slowdown.
- Internally calling `render` function whenever a property of `RegularShape` has changed (already implemented at [ol.style.Circle.setRadius](https://github.com/openlayers/openlayers/blob/c767091faf7de8340aa3eb06ede5930e519c01ac/src/ol/style/Circle.js#L61)) might cause slowdown under some circumstances.
- Creating new instance of `RegularShape` (which is a working solution) would not solve this bug-prone developing situation.

### The solution
Let developers choose when to regenerate image, by exposing `RegularShape.render` method from protected to public. Also, I suggest adding setter methods to `RegularShape` for dynamically changing its properties. I'll submit PR on this issue when ready.

dnsdhrj:
Closes #9820 
Note that `ol.style.Circle.getRadius` no longer calls `render`, which makes this commit backwards incompatible.

### DIFF
--- a/src/ol/style/RegularShape.js
+++ b/src/ol/style/RegularShape.js
@@ -299,7 +299,8 @@ class RegularShape extends ImageStyle {
   unlistenImageChange(listener, thisArg) {}
 
   /**
-   * @protected
+   * Regenerate the shape.
+   * @api
    */
   render() {
     let lineCap = defaultLineCap;


dnsdhrj:
No, I think the example still needs to generate all styles. Every point must have its own style because radius are not same, whereas this commit allows _dynamic_ change of style, or _mutating_ style.

mike-000:
> No, I think the example still needs to generate all styles. Every point must have its own style because radius are not same, whereas this commit allows _dynamic_ change of style, or _mutating_ style.

Style functions frequently update properties of a style for each feature e.g. `style.getText.setText(feature.get('name'); `for labels.  Creating a complete new style for every labelled feature would be very inefficient.  Radius is no different and it works in OL5 for CircleStyle https://codesandbox.io/s/icon-color-cgc7t

dnsdhrj:
Oh, I see your point. Then the example code will be optimized.

ahocevar:
> So I marked render as API.

I was hoping we could keep this unofficial, because we'll eventually be reworking the way we handle regular polygon styles and move all rendering to the renderer. Ok if you undo that change?

dnsdhrj:
If `render` is not documented, someone will open issue with title such as "Angle does not change after calling `setAngle` method" after merging this. So I suggest either:
- drop commit b8ef1e5 (remove setter APIs)
- add documentation to new setter APIs, such as "Shape should be regenerated." or "Currently the actual shape does not change."

What do you think @ahocevar ?

ahocevar:
If `ol/style/Fill`, `ol/style/Stroke` and `ol/style/Circle` had a `toString()` method, consumers could easily create keys that represent the full style. This would be useful elsewehre too. We had something like that already, but called it checksum: https://github.com/openlayers/openlayers/commit/0ec769c234f32c3bbd88343ef5e2f3988a6e6edd#diff-8799e1ec31b8cf6732a8f0bda23c70d3.

With something like that, we could lazily render regular shapes at render time, and only render when the rendered string representation is different from the current string representation:

```js
render() {
  const currentStyleKey = this.toString();
  if (currentStyleKey !== this.renderedStyleKey_) {
    this.render_();
    this.renderedStyleKey_ = currentStyleKey;
  };
}
```

dnsdhrj:
Upon inspecting some code for , I figured out that there's no 'invoke rerender' functionality for Image styles.
And implementing that functions looks like quite an overhaul, such as this:

The function below has to be changed, since `ol.ImageState` semantics looks like irreversible state (can't go from `LOADED` to `IDLE`):
https://github.com/openlayers/openlayers/blob/9cb96b23dbdd88347924a1225eb22fc45ebe7bc9/src/ol/renderer/vector.js#L100-L119
And from what I stated in #9820 :
> * Internally calling `render` function whenever a property of `RegularShape` has changed (already implemented at [ol.style.Circle.setRadius](https://github.com/openlayers/openlayers/blob/c767091faf7de8340aa3eb06ede5930e519c01ac/src/ol/style/Circle.js#L61)) might cause slowdown under some circumstances.

other simple solutions like event listener are not good solution.

I can't afford this job, so I will just

> * drop commit [b8ef1e5](https://github.com/openlayers/openlayers/commit/b8ef1e5401f9bcdfa62eeedf978ca63bddca651c) (remove setter APIs)

, and leave other works to renderer reworking.
Therefore, this PR will just pull `@protected` out from `render` function.

dnsdhrj:
Here are commits before drop: https://github.com/openlayers/openlayers/compare/0c8775b3c8b65b4c30c130f26fc03f14ca280435...dnsdhrj:regularshape-render-orig

PR #9824, opened 2019-08-06
Support zoom limits for layers

tschaub:
This adds support for `minZoom` and `maxZoom` options for limiting layer visibility.  The `minZoom` is exclusive and the `maxZoom` is inclusive.  If `maxResolution` or `minResolution` are set, they will take precedence.

### DIFF
--- a/examples/layer-zoom-limits.html
+++ b/examples/layer-zoom-limits.html
@@ -1,0 +1,13 @@
+---
+layout: example.html
+title: Layer Zoom Limits
+shortdesc: Using minZoom and maxZoom to control layer visibility.
+docs: >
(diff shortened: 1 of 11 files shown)


bojko108:
Great, I'm managing layers visibility using scale values as they are more user friendly (1:25 000, 1:50 000 etc.). From them I'm calculating min and max resolutions to pass them to the layers, but your solution is even better.

ahocevar:
I added a commit to fix the failing tests.

tschaub:
Ack.  Thanks for fixing up the tests @ahocevar.  I made a commit to do the same while offline and force pushed when I got back online before looking at this PR.

PR #9831, opened 2019-08-07
getGetLegendRequestUrl accepts a LAYER parameter

### LINKED ISSUE #9829: getGetLegendRequest should take a layer name as parameter
Follow up to #9762

While using the getGetLegendRequest method I found that it would be very useful if it was possible to set the layer for which a legend is requested instead of just aborting the execution.

https://github.com/openlayers/openlayers/blob/ec10cda088cdfc447b0e94fce8fccf6d9d5d9ae0/src/ol/source/ImageWMS.js#L210

I would suggest it is possible to set the `LAYER` property of params which will be checked for existence and only abort if either `LAYER` is set or the source has more than one or no wms layers.

simonseyock:
Fixes #9829

### DIFF
--- a/src/ol/source/ImageWMS.js
+++ b/src/ol/source/ImageWMS.js
@@ -197,15 +197,15 @@ class ImageWMS extends ImageSource {
    *
    * @param {number} [resolution] Resolution. If set to undefined, `SCALE`
    *     will not be calculated and included in URL.
-   * @param {Object} [params] GetLegendGraphic params. Default `FORMAT` is
-   *     `image/png`. `VERSION` should not be specified here.
+   * @param {Object} [params] GetLegendGraphic params. If `LAYER` is set, the
+   *     request is generated for this wms layer, else it will try to use the
+   *     configured wms layer. Default `FORMAT` is `image/png`.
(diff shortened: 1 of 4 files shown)


tschaub:
Unrelated to this change, but I wonder what folks would think of renaming `getGetLegendGraphicUrl` to `getLegendUrl`.

simonseyock:
The name was chosen to be similar to getGetFeatureInfoUrl. I am in favour of renaming, it is really clunky now. But then the other method should be renamed as well.

marcjansen:
This is what we major releases for. Plus one on renaming

jahow:
Once the renaming is done this PR is good to merge IMO. Thanks!

As for renaming `getGetFeatureInfoUrl`, what would you call it @simonseyock ?

simonseyock:
Sorry for answering this late I am on vacation right now. I would rename it to `getFeatureInfoUrl` . But I would prefer a different PR for renaming.

simonseyock:
But I can open another one first and then rebase this one.

simonseyock:
Rebasing done.

tschaub:
Thanks @simonseyock.

PR #9853, opened 2019-08-13
Remove accidentally added package-lock.json

fredj:


ahocevar:
You may also want to wait for @tschaub's opinion on this.

tschaub:
Greenkeeper keeps the lock file updated.  See https://github.com/openlayers/openlayers/commit/2d0d06842d6607fae82373c1d730fe4ef25ac15b.  Having the lock file in the repo makes sense to me.

tschaub:
If we think different non versions will lead to different lock file changes, and we don’t want people adding these when there are not dependency updates, then ignoring is ok by me too.  But I also don’t mind having it in there.

Issue #5244, opened 2016-04-18
Cross-origin authenticated layers with xhr option "withCredentials"


juristr:
Hi,

I'm having an application running under host, say, http://maps.myservice.com. Through ol3 I'm registering 3 different endpoints to overcome the usual browser connection limit:

- `a-maps.myservice.com`
- `b-maps.myservice.com`
- `c-maps.myservice.com`

Now the problem is that XHR requests don't send authentication information to the backend when they're going cross-domain, obviously (we use Kerberos authentication, not forms or basic).  
To overcome this limitation, a possible strategy is to set the `withCredentials: true` option on the xhr object. [See here for more information](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials).

**My question is** whether currently there is some possibility to augment the internal XHR object that's being used by ol3 with such custom properties.

Thx for any hint/help!

tsauerwein:
Have you seen https://github.com/openlayers/ol3/issues/4213?

juristr:
@tsauerwein Nope, searched the issues, but couldn't find it. But that's awesome and should definitely help :+1:. I'll verify and then close the issue, thx!

juristr:
@tsauerwein ok, seems to work, although I'd prefer to like get an instance of an `XMLHttpRequest` object handed over by ol3 which I can then augment with custom props rather then overriding the complete implementation which forces me to handle also the association of image tiles and potentially other logic..

ahocevar:
@juristr There is no `XMLHttpRequest` involved in fetching image tiles. It's just setting the `src` of an `Image`.

juristr:
ah I see, thx for answering @ahocevar

PR #9865, opened 2019-08-16
Set xhr.withCredentials = true in loadFeaturesXhr()

mstenta:
This pull request is related to #5244.

It adds `xhr.withCredentials = true;` to `loadFeaturesXhr()` so that cookies are sent when pulling GeoJSON from a cross-origin domain.

Currently, adding a GeoJSON source from another domain that is authenticated via a session cookie returns a 403 response, while opening the same URL in another tab returns a 200.

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -62,6 +62,7 @@ export function loadFeaturesXhr(url, format, success, failure) {
       if (format.getType() == FormatType.ARRAY_BUFFER) {
         xhr.responseType = 'arraybuffer';
       }
+      xhr.withCredentials = true;
       /**
        * @param {Event} event Event.
        * @private
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -62,7 +62,9 @@ export function loadFeaturesXhr(url, format, success, failure) {
       if (format.getType() == FormatType.ARRAY_BUFFER) {
         xhr.responseType = 'arraybuffer';
       }
-      xhr.withCredentials = true;
+      if (window.withCredentials) {
+        xhr.withCredentials = window.withCredentials;
+      }
       /**
        * @param {Event} event Event.
        * @private
@@ -131,4 +133,15 @@ export function xhr(url, format) {
         /** @type {import("./source/Vector").default} */ (sourceOrTile).addFeatures(features);
       }
     }, /* FIXME handle error */ VOID);
+}
+
+
+/**
+ * Sets a global withCredentials variable that will be used in
+ * loadFeaturesXhr() above.
+ * @param {bool} bool Boolean that will be used in xhr.withCredentials.
+ * @api
+ */
+export function setWithCredentials(bool) {
+  window.withCredentials = bool;
 }


ahocevar:
This change may not be suitable for all users. It can cause credential information to be exposed to third party sites where it is not intended.

In your scenario, it would be better to either use custom loaders, or use a global XMLHttpRequest replacement that adds the `withCredentials` property to every request.

mstenta:
Thanks @ahocevar - we actually did implement a custom loader already, in case this PR isn't accepted. I would prefer to use OpenLayers' native loader function, of course, to reduce our custom code footprint.

> It can cause credential information to be exposed to third party sites where it is not intended.

Is this true?  Maybe my understanding of `withCredentials` isn't complete, but I thought it only sends credentials for the domain that the request is being made to?  In other words, the server will only receive cookies for its own domain.

(Eg: if my Openlayers script is on `http://mydomain.com` and it is requesting GeoJSON from `https://yourdomain.com`, then it will *only* include cookies for `yourdomain.com` in the request.)

And even then, the requesting domain (`mydomain.com`) still can't access the cookies.  The browser handles sending them automatically without exposing them to the script making the request.

It is the server's responsibility to set CORS response headers to restrict cross-origin requests.

So I'm not sure if there is a risk including `withCredentials: true` in Openlayers.  But I would welcome input/references from someone more knowledgeable than me. :-)

ahocevar:
You're right, `withCredentials` only affects credentials for the domain that the request is being sent to. I do see the benefit of your pull request, but I think it should be configurable. Can you maybe change it so `withCredentials` can be enabled through a global setting, something like
```js
import setWithCredentials from 'ol/featureloader';
setWithCredentials(true);
```

mstenta:
Oh that's a good idea! Yes, I can try to get something together...

mstenta:
Pushed one additional commit - **this is not tested yet** - I just wanted to get something up to confirm that this is what you had in mind @ahocevar .

Do you think there's any alternative to setting a global `window.withCredentials` variable? I don't see this pattern used elsewhere in OpenLayers, and I don't love using a global variable. But on the other hand, I can't see any other way of making this configurable. Unless we added an argument to the `loadFeaturesXhr()` function... but that would require more extensive changes elsewhere in the code I imagine (which we should probably avoid).

What do you think? Is this the best way to do it?

jahow:
Hi, just dropping by to help clarify things (hopefully). Indeed I don't think adding a global variable on `window` was what Andreas had in mind, more something like:
```js

let useWithCredentials = false;

// this is not part of the api but still exported
export function getWithCredentials() {
  return useWithCredentials;
}

// @api something something
export function setWithCredentials(b) {
  useWithCredentials = b;
}

// and in another library file
import { getWithCredentials } from '...'
```

Although I'm not 100% sure this would work, it has the benefit of not relying on a global scope.

ahocevar:
Thanks for chiming in @jahow. Yes, this is what I had in mind. But we don't need the getter function.

KaiVolland:
Hey @mstenta, do you have time to work on this or would you prefer if someone else would finish this up?

mstenta:
@KaiVolland If you want to jump in be my guest! Otherwise I will give it another try when I have a chance, although that may be a few weeks realistically. We ended up adding a loader function to our GeoJSON code to work around this, but I'd love to remove that when this is available!

KaiVolland:
I'll have a look at it.

PR #9869, opened 2019-08-17
Only get squared tolerance once per render

MoonE:
The resolution and pixelRatio variables do not change in the loop, thus it should be safe to only calculate the squared render tolerance once per call to renderFeature.

### DIFF
--- a/src/ol/renderer/canvas/VectorLayer.js
+++ b/src/ol/renderer/canvas/VectorLayer.js
@@ -362,10 +362,10 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
     }
     let loading = false;
     if (Array.isArray(styles)) {
+      const squaredTolerance = getSquaredRenderTolerance(resolution, pixelRatio);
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks for the contribution @MoonE!  You are right that we don't need to get the squared tolerance as often when rendering vector features.  It looks like we should only need to do it once per render instead of once per style per feature per render.

I pushed another commit to your branch with the additional change.

Note that for future contributions, it may be easier to create a feature branch with the changes you would like merged (instead of committing directly to your master branch).  That way you can delete the feature branch when your pull request is closed, and you can branch again from master for your next feature.  A typical workflow might look like this:

```bash
# add upstream remote
git remote add openlayers git@github.com:openlayers/openlayers.git

# get the latest from upstream
git fetch openlayers

# start a new feature branch
git checkout -b my-new-feature openlayers/master

# make commits in your feature branch and then push
git push origin my-new-feature-branch
```

tschaub:
Thanks again for the contribution @MoonE.

Issue #9403, opened 2019-04-05
Major stuttering on Edge when panning


FINDarkside:
**Describe the bug**
There seems to be a big performance issue on Edge. When panning the map, the panning isn't smooth at all and stutters for long periods of time. I've done performance profiling on Edge and there's 3 things that stand out:
### pointermove event is handled very often. 
Sometimes multiple times in a millisecond. In the image you can see, how it's supposedly handled 29 times in just 50ms.  
![Pic](https://i.gyazo.com/9653dcca1bd4873ead1058d1ae73c8be.PNG)

### Paint and animation frame callback happens rarely

Despite (or because of) the huge number of these pointermove events, paint and animation frame callback is fired very rarely, about 11 times a second when I profiled it.

### Big empty gaps
Now this is not very scientific as I did manually pan the map. But given that the gaps are on consistent intervals I'm suspect it might not be because I would have stopped panning for a short period of time. In addition these gaps are usually either preceded or followed by unusually long pointermove event.
![Image](https://i.gyazo.com/4920711f0252be003557e8f58c52edf5.png)
![Closeup](https://i.gyazo.com/edb59a079251210d54a73801f1ebe085.png)
This might not be related to the issue though, since the repaint would have still been about 11 times a second even without these gaps.

**To Reproduce**
This issue can be produced on [demo page](https://openlayers.org/en/latest/examples/mobile-full-screen.html) with Edge 42 on Windows 10.

I also profiled the same map on Chrome, and on Chrome animation frame and repaint happened every time after pointermove event.

ahocevar:
Can you please try if this also happens with the [v6 beta version](https://openlayers.org/en/latest/examples/mobile-full-screen.html) of the example? The renderers got a significant overhaul in the meantime.

FINDarkside:
Yes, happens in the link provided as well.

jeroenvheel:
I have alsp problems with the speed of the map with the latest beta verion (6).
i created a new Map with VectorImageLayers.and with drag and drop i add layers 30 layers based on (gpx) files and with each layer the performace of the map get worse until you can't pan the map
![ol-6b6 performance](https://user-images.githubusercontent.com/228655/56460137-34016500-639e-11e9-90b4-57afb6617c54.png)

FINDarkside:
I doubt these are related. The original issue should be reproducible on demo pages without many layers, and is only problem on Edge.

FINDarkside:
Bump. I can still reproduce this.

[...]

PR #9883, opened 2019-08-26
Do not precalculate coordinate and pixel on MapBrowserEvent

ahocevar:
With this breaking change we save a lot of useless work pre-calculating pixel and coordinate of a `MapBrowserEvent`. Instead, we now provide `getCoordinate()` and `getPixel()` functions on the event object.

See #9403.

### DIFF
--- a/examples/custom-interactions.js
+++ b/examples/custom-interactions.js
@@ -51,13 +51,13 @@ class Drag extends PointerInteraction {
 function handleDownEvent(evt) {
   const map = evt.map;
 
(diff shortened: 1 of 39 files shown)


fredj:
Would it be possible to use es6 getters instead of functions? typescript is able to transpile it to es5 but I haven't checked jsdoc

ahocevar:
I'm going to replace this breaking change with a different approach to resolving the reported performance issues.

PR #9888, opened 2019-08-29
Use PointerEvent polyfill instead of our own abstraction

ahocevar:
This pull request gets rid of our Polymer based pointer event abstraction, and replaces it with a package dependency of its successor, the [pepjs](https://www.npmjs.com/package/pepjs) polyfill.

As a side effect of this change, the zoom slider control could also be simplified a bit.

Once pointer events have landed in all evergreen browsers, which should be the case when [Safari 13](https://caniuse.com/#feat=pointer) is released, we can get rid of the polyfill.


src/ol/control/ZoomSlider.js
```diff
@@ -158,7 +154,6 @@ class ZoomSlider extends Control {
    * @inheritDoc
    */
   disposeInternal() {
```
fredj: the function can be removed

PR #9895, opened 2019-08-30
Fix WebGL warnings (closes #9730)

### LINKED ISSUE #9730: Heatmap layer WebGL warnings
Warnings are displayed in the dev console when loading the heatmap-earthquakes example.
I guess it's related the the hit detection added in #9655

The warnings are:
```
WebGL: INVALID_OPERATION: useProgram: program not valid
WebGL: INVALID_OPERATION: getUniformLocation: program not linked
WebGL: INVALID_OPERATION: getAttribLocation: program not linked
WebGL-0x34f454258100]GL ERROR :GL_INVALID_OPERATION : glDrawElements: Source and destination textures of the draw are the same.
Helper.js:666 WebGL: too many errors, no more errors will be reported to the console for this context.
```

Maybe we should have a mechanism to deactivate the hit detection for a layer?

fjorgemota:
Hey.

This, at least here on my tests on both Chrome and Firefox, fix the WebGL warnings reported by #9730. I don't know if this is the optimal solution, even because I don't know WebGL a lot....but given the firefox error (reported below) I was able to fix the problem.

image

I'm not sure about how to add regression/tests so this warning never happens again. I accept ideas here. :)

### DIFF
--- a/src/ol/layer/Heatmap.js
+++ b/src/ol/layer/Heatmap.js
@@ -192,6 +192,8 @@ class Heatmap extends VectorLayer {
 
         varying vec2 v_texCoord;
         varying float v_opacity;
+        varying vec4 v_color;
+
 
         void main(void) {
           vec4 offsets = u_offsetScaleMatrix * vec4(a_offsets, 0.0, 0.0);


jahow:
Hey! Thanks a lot for tackling this. I've looked at this on my side too and I think it is better to fix this in the points renderer instead of at the layer level.

The error is caused by the fact that the hit detection pipeline uses a custom fragment shader, which is not exposed in the renderer options, and which expects a `varying v_color` to come from the vertex shader.

Since the vertex shader can be overwritten (and is in the case of the heatmap layer), it means there is currently an undocumented requirement to have a `varying v_color` in the vertex shader even if it is of no use in the render pipeline.

I have created #9896 for another, less error-prone approach. Hope that makes sense. Also I haven't really found a way to add unit tests for this, maybe we should somehow check the absence of warning in rendering tests?

fjorgemota:
> Hey! Thanks a lot for tackling this. I've looked at this on my side too and I think it is better to fix this in the points renderer instead of at the layer level.

Hey! No problem. I'm always happy to help and, as I said on the PR description, I didn't think my proposed solution was the best anyway. :)

> I have created #9896 for another, less error-prone approach. Hope that makes sense. Also I haven't really found a way to add unit tests for this, maybe we should somehow check the absence of warning in rendering tests?

Seems like a nice alternative...and about the unit tests, I think that really checking for the absense of warnings would be great. I just don't have any ideas about how to do that on the OpenLayers current test suite.

I'll close this PR in favour of #9896 

Thanks!

PR #9897, opened 2019-08-31
Use pointer events everywhere

ahocevar:
This pull request cleans up the code base, removing listeners and event types for mouse and touch events.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -866,6 +866,11 @@
         "@openlayers/doctrine": "^2.2.0"
       }
     },
+    "@openlayers/pepjs": {
+      "version": "0.5.3",
+      "resolved": "https://registry.npmjs.org/@openlayers/pepjs/-/pepjs-0.5.3.tgz",
+      "integrity": "sha512-Bgvi5c14BS0FJWyYWWFstNEnXsB30nK8Jt8hkAAdqr7E0gDdBBWVDglF3Ub19wTxvgJ/CVHyTY6VuCtnyRzglg=="
+    },
     "@sinonjs/commons": {
       "version": "1.4.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.4.0.tgz",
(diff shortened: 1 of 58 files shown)


ahocevar:
@tschaub Thanks for testing. Which mobile device did you use? I just tried the linked example on an iPhone SE (iOS 10.14.1), and it works with pointer events turned off and on in the experimental settings.

tschaub:
On an iPhone Xs (12.4), it looks like pointer events are off by default.

In this mode, the zoom +/- buttons (and rotate north and attribution) don’t work for me.  

When I enable pointer events in the experimental settings, the controls work.  In both cases, the zoom/pan/etc interactions work (only the controls do not).

ahocevar:
Hm. Actually I have the same OS version as you, just a different phone. I also had pointer events turned off (default) when I tested. Strange that the controls work for me and don't work for you.

Since I cannot reproduce the problem, do you think you can investigate this a bit and come up with a solution?

ahocevar:
I looked at the code, and the controls rely on plain old `click` listeners. Interesting how they would be affected by the PointerEvent polyfill. Browsing the issues in https://github.com/jquery/PEP, I also don't find anything that looks related.

ahocevar:
Nevermind @tschaub, I was now able to reproduce the issue on an iPhone 8 with iOS 12.4.

ahocevar:
I added a commit (b9caafb47b12566c8e0ee60bbf4f74a6c67d26bc) that simplifies our whole event system a fair bit. Previously, we stored listeners twice, and we also stored listeners that we registered on DOM elements. This is all removed now.

@tschaub Does this solve the iPhone issue? At least it should make it easier to debug.

tschaub:
It looks like things work if we stop setting the `touch-action` attribute to `'none`` in the map browser event handler.  Deleting https://github.com/ahocevar/openlayers/blob/more-pointer-events/src/ol/MapBrowserEventHandler.js#L64 makes the problem go away for me.

ahocevar:
The problem we're facing is in pepjs. I submitted jquery/PEP#388 and published https://npmjs.com/package/@openlayers/pepjs to use with OpenLayers.

@tschaub Can you please pull the latest and confirm that the problem is fixed now? Thanks!

tschaub:
That did the trick.

touch

tschaub:
Very nice!

PR #9903, opened 2019-09-04
Respect multiWorld: false when an explicit extent constraint would be respected

### LINKED ISSUE #9577: multiWorld: false is ignored when constrainResolution: true is set
**Describe the bug**
If `constrainResolution: true` is set more than one world can be seen despite `multiWorld: false` being set

**To Reproduce**
https://codesandbox.io/s/simple-ygzz5

**Expected behavior**
View is restricted to integer zoom levels which satisfy  `multiWorld: false`  as happens when an extent constraint is set https://codesandbox.io/s/simple-2see3

### LINKED ISSUE #9937: Extent constraint can be exceeded when combined with a constrained resolutions array
I noticed an expected result while adding tests for #9903 and checked if the same happened with regular extent constraints, which it does.

It can be seen in https://codesandbox.io/s/simple-busdx where the maximum resolution which complies with the extent constraint is 78271.51696402048 but it is possible to zoom out to resolution 100000 because the nearest integer zoom level is being selected instead of the next (i.e. lower resolution).  If you change 100000 to 110000 (or 50000 to 60000) it can be seen working correctly.

If this can be fixed the [160000, 75000, 40000, ... resolutions array in the test can be [160000, 80000, 40000, ...  and the test result will be 40000.

mike-000:
Fixes #9577
Fixes #9937

The multiWorld: false constraint is currently not taken into account when constrainResolution is set or when an explicit resolutions array is specified, while an explicit extent constraint is respected in the same situations.  This pull request takes a global projection extent into consideration in that same situations that an explicit extent would be.  Also includes fix for #9937 which affected this but isn't specific to multiWorld: false

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -198,7 +198,7 @@ const DEFAULT_MIN_ZOOM = 0;
  *
  * ### The view states
  *
- * An View is determined by three states: `center`, `resolution`,
+ * A View is determined by three states: `center`, `resolution`,
  * and `rotation`. Each state has a corresponding getter and setter, e.g.
  * `getCenter` and `setCenter` for the `center` state.
  *
@@ -1452,6 +1452,15 @@ export function createResolutionConstraint(options) {
   const smooth =
       options.smoothResolutionConstraint !== undefined ? options.smoothResolutionConstraint : true;
 
+  const projection = createProjection(options.projection, 'EPSG:3857');
+  const projExtent = projection.getExtent();
+  let constrainOnlyCenter = options.constrainOnlyCenter;
+  let extent = options.extent;
+  if (!multiWorld && !extent && projection.isGlobal()) {
+    constrainOnlyCenter = false;
+    extent = projExtent;
+  }
+
   if (options.resolutions !== undefined) {
     const resolutions = options.resolutions;
     maxResolution = resolutions[minZoom];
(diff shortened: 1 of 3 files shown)


tschaub:
Thanks for the proposed fix @mike-000.  I haven't dug into this yet, but I'd feel more comfortable doing a review if there were some tests that made it clear what the expected behavior was.  Maybe @jahow would be interested in taking a look.

mike-000:
These simple examples from #9577 should prove the basics.  Currently specifying `constrainResolution` or a resolutions array (even when not constrained) causes the view to behave as if `multiWorld: true` was set even when explicitly set false https://codesandbox.io/s/simple-1y971  https://codesandbox.io/s/simple-mgi30  The new expected behaviour should be similar to that seen when the projection extent is set as a constraint along with `constrainResolution` or a resolutions array ( as in https://codesandbox.io/s/simple-2see3 and https://codesandbox.io/s/simple-bdljs ) **except that panning across the dateline must be possible as long as only one world is visible**.  I noticed some previous changes used `test/spec/ol/view.test.js` which reviewers might wish to use to check for any undesired side effects (and it is possible that some old tests and examples might now need `multiWorld: true` to maintain their previous intended behaviour).

jahow:
Hi @mike-000, thanks a lot for the proposal, I think it's all good to merge. Could you maybe try to simplify the commit history a bit, if that is possible? Something like one commit per issue, to help make things clearer? Thanks

mike-000:
> Hi @mike-000, thanks a lot for the proposal, I think it's all good to merge. Could you maybe try to simplify the commit history a bit, if that is possible? Something like one commit per issue, to help make things clearer? Thanks

Yes, it's grown to two commits on each component.  I'm not sure of how to merge commits, but I could do a separate single commit PR for #9937 which would be a prerequisite (due to the tests needing it) for a cleaned up version for #9577

jahow:
You can use interactive rebase (`git rebase -i`) to squash the commits and reword their message, that really is all there is to it. No need to create another PR :)

PR #9919, opened 2019-09-10
Revise and correct OverviewMap rotation

### LINKED ISSUE #9910: OverviewMap shows incorrect box when view is rotated
This can be seen either of the examples:

image

Implementing the "TODO: This should rotate the extent rectangle instead of the overview map's view" comment in the code is easy, and a rotateWithView option/setter/getter can be used to produce either that behaviour or the current behaviour (but with the correct box), as in this demo https://codepen.io/mike-000/pen/QWLQKQP (debug full builds can still be useful!)
 
Is there a preferred default value for rotateWithView?

mike-000:
Fixes #9910

Correct the box display when the main map view is rotated and add an option to rotate either the box or the overview map view


tschaub:
Thanks for the work on this @mike-000!

PR #9920, opened 2019-09-10
Update OverviewMap rotation tests to use rotatable controls

mike-000:
#9919 makes `rotateWithView` optional so controls used for rotation tests must set to be rotatable


mike-000:
commit now added to #9919 so this is no longer needed

PR #9929, opened 2019-09-13
Opelline/draw interaction extend polygons

opelline:
The Draw interaction has only support for extending the drawing of a previous LineString Feature (the extend function). Appending coordinates from a feature to a drawing has not been supported.

This branch contains a new appendCoordinates function, which supports appendign both when drawing Polygons and LineStrings, by appending linear coordinates, such as coordinates from a LineString feature or a LinearRing of a Polygon feature. The coordinates are appended to the set of previously drawn coordinates.

I have also created an example for demonstration.
(Edit: Updated the description according to the separation of the appendCoordinates and extend functions)


jahow:
Hi,

I personally have a hard time seeing a use case for this. Isn't it prone to confusion to be able to clone the geometry in the same place and modify the new one? Besides, it looks like the new function you implemented is basically the equivalent of cloning the feature & extending it with the draw interaction. Sorry if I'm missing something here!

opelline:
Hi,

Thanks for the comment: this was built for an actual use case, as the user needs to be able to follow the borders on a vector layer, without snapping each edge point by hand. This could be a border of a region, or some part of a neighbouring feature. This of course requires some app specific handling for doing the snapping and slicing of those line segments before they can be added to the current geometry.

The extend feature doesn't allow adding points in the other end of a segment, and it also doesn't work on polygons. I was first trying to implement this using the extend feature, but I got confused and thought it was buggy. It was only after runnign the automated build tests that I actually understood what the extend feature was for, and why it was working the way it was. So now I've clarified the feature documentation to separate the two.

One might think all this could be achieved by just taking the coordinate arrays to be joined, concatenate them as you like, and creating a new feature, which then could be used in the extend feature. However, this doesn't work with polygons, as it will break the draw interaction.

opelline-sito:
@jahow Please tell me if you have any other comments or feedback!

opelline:
@jahow Please tell me if you have any further questions or comments on this!

jahow:
Hi @opelline, thanks, I understand better now the limitations of the `extend` method. It indeed does make sense to be able to inject an arbitrary series of coordinates in the currently drawn feature, that would allow for nice user experience when drawing.

I think the API description could be a bit more clearer, ie:
```js
  /**
   * Append coordinates to the end of the geometry that is currently being drawn.
   * This can be used when drawing LineStrings or Polygons. Coordinates will
   * either be appended to the current LineString or the outer ring of the current
   * Polygon.
   */
```

Also some remarks on the actual code:
* this method will need a few dedicated tests
* I'd be in favor or reusing `addToDrawing_` and rework this private method a little to accept a coordinate instead of an event; this would insure consistency with the other behaviours of the interaction & avoid duplicating code/logic
* you should also rebase this PR on master (the draw examples behave weirdly on this branch, the sketch feature is not visible)

But other than that I think this contribution is definitely valuable and interesting. Can someone else maybe confirm this should be pushed forward? @fredj maybe?

Thanks a lot @opelline

opelline:
Hi @jahow , I've been quite busy for the past few weeks, but now I had the time to do the requested changes, and I'd like to reopen this.

PR #9930, opened 2019-09-13
Correct extent clipping in ImageLayer

### LINKED ISSUE #9585: Rotation broken on Image layers with extent
**Describe the bug**
Image layers with an `extent` set disappear (at least partially) on rotated views

**To Reproduce**
1. Go to https://openlayers.org/en/master/examples/wms-image.html
2. Rotate the map (e.g. with alt-shift-drag)
3. See the wms overlay disappear

**Expected behavior**
The wms overlay should be properly displayed when rotated.

mike-000:
Fixes #9585

Clipping in image layers does not need to consider view rotation

### DIFF
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -121,7 +121,7 @@ class CanvasImageLayerRenderer extends CanvasLayerRenderer {
           !containsExtent(extent, frameState.extent) &&
           intersects(extent, frameState.extent);
     if (clipped) {
-      this.clip(context, frameState, extent);
+      this.clipUnrotated(context, frameState, extent);
     }
 
     const img = image.getImage();


tschaub:
Thanks for the fix @mike-000.

It would be great to add a render test to catch this type of regression.  The existing [image layer test](https://github.com/openlayers/openlayers/blob/master/rendering/cases/layer-image/main.js) could be copied and rotation added.

mike-000:
> Thanks for the fix @mike-000.
> 
> It would be great to add a render test to catch this type of regression. The existing [image layer test](https://github.com/openlayers/openlayers/blob/master/rendering/cases/layer-image/main.js) could be copied and rotation added.

In this case rotation on its own wasn't the problem, it was rotation in combination with a layer extent being set, and only then when the current view extent wasn't fully contained by the layer extent.  It would certainly be possible to test that combination, but to properly guard against regression issues caused by any further major redesigns of how layers are rendered several possible combinations of rotation and extents would need to be tested (and not just for ImageLayer).

tschaub:
Thanks for the additional detail.  My point is just that each time we add a fix for a regression, it is nice if we can add a test.  That will give us more confidence when making changes in the future.

It is awkward to ask someone who is already volunteering time to do the additional work of writing tests for an issue that they are fixing, but we really should stick with a policy of only accepting changes that come with tests (whether the changes are features or fixes).

Again, thanks for the time you have spent debugging and fixing these issues.

PR #9931, opened 2019-09-13
Custom Overview Map CSS fix and rotateWithView option

### LINKED ISSUE #9911: CSS for the Custom Overview Map example not working outside the example template
If you open the OL6 example via the Edit link it is very apparent the custom CSS isn't working correctly.  It works inside the examples template because that links ol.css but importing ol/ol.css as in codesandbox breaks it.  I found that using a mapcontainer div as in the Advanced View Positioning example fixes it https://codesandbox.io/s/overviewmap-custom-q5i81 (but can anyone suggest a better way?) and I could do that along with updating the example for #9910

mike-000:
Fixes #9911

Use CSS similar to the Advanced View Positioning example to give correct display outside the examples template.

Add a demonstration of the `rotateWithView` option added by #9919

### DIFF
--- a/examples/overviewmap-custom.css
+++ b/examples/overviewmap-custom.css
@@ -1,32 +1,36 @@
-.ol-custom-overviewmap,
-.ol-custom-overviewmap.ol-uncollapsible {
+.mapcontainer {
+  position: relative;
+}
+
+div.ol-custom-overviewmap,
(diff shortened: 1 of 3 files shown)


tschaub:
The [hosted example](https://1964-4723248-gh.circle-artifacts.com/0/examples/overviewmap-custom.html) looks good.

PR #9942, opened 2019-09-18
Renaming getGetFeatureInfoUrl

simonseyock:
As discussed in #9831 here is a renaming of the `getGetFeatureInfoUrl` and `getGetLegendGraphicUrl` methods to `getFeatureInfoUrl` and `getLegendUrl`.

I found that some headlines in the upgrade notes were not formatted correctly so I changed that. If I should make a different PR for this, just say.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -4,7 +4,7 @@
 
 #### Backwards incompatible changes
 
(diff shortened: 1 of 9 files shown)


tschaub:
Thanks @simonseyock!

PR #9951, opened 2019-09-20
New test for Image layer with extent and rotation

mike-000:
Follow up for #9585/#9930


tschaub:
Nice!  Thanks @mike-000.

PR #9954, opened 2019-09-22
Use next lower resolution in constrained array if snap value exceeds extent cap

### LINKED ISSUE #9937: Extent constraint can be exceeded when combined with a constrained resolutions array

mike-000:
Fixes #9937

Separated from #9903 so that commits for that can be simplified.  Merging this change is a prerequisite for a simplified version of #9903 as the tests require this fix.

### DIFF
--- a/src/ol/resolutionconstraint.js
+++ b/src/ol/resolutionconstraint.js
@@ -82,6 +82,9 @@ export function createSnapToResolutions(resolutions, opt_smooth, opt_maxExtent) 
 
         const capped = Math.min(cappedMaxRes, resolution);
         const z = Math.floor(linearFindNearest(resolutions, capped, direction));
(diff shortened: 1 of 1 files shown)


mike-000:
Included in #9903

PR #9960, opened 2019-09-23
Always dispatch pointermove events, but calculate event pixel and coordinate lazily

ahocevar:
With the changes made in #9889, it is no longer possible to set a condition on an interaction that involves a `pointermove` event.

This change makes it so we return to always dispatching `pointermove` events, but lazily calculate the coordinate and pixel of the event, which was the main reason for not always dispatching `pointermove`.

To avoid the need for pixel and coordinate calculation, I also removed the frame state's `focus` property, which was only used to prioritize tile loading. We now use the map center instead.


ahocevar:
A next step would be to get rid of `MapBrowserEvent` entirely, but we can do that separately after the 6.0 release.

PR #9961, opened 2019-09-23
SelectInteraction removal

simonseyock:
This PR removes the `SelectInteraction` as it is inherently broken, continues to confuse people and does not add much value. 

It provides 2 new examples which show how to achieve similar results with very basic code.

In consequence the skipFeature logic will be removed, too.


simonseyock:
How the replacement of the select interaction looks like in most examples:

https://github.com/openlayers/openlayers/pull/9961/commits/c02d9f1e68d822045816d0c5afaf1f119b50c3ca#diff-3f5ff1564401bac296826ec159b7c595

tschaub:
Thanks for this huge effort @simonseyock!  Looks great to me.

PR #9966, opened 2019-09-23
Document how to circumvent the canvas reuse optimization

jahow:
The optimization was introduced in https://github.com/openlayers/openlayers/pull/9584

This PR documents how to get the original functionality of `map.forEachLayerAtPixel`, without getting false positives. Setting a `className` different on each layer will do the trick.

Also I've added the `className` property in the API doc of the `Layer` class (previously it was only on the `BaseLayer` one).

This is intended to help with openlayers#9720

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -4,6 +4,23 @@
 
 #### Backwards incompatible changes
 
+##### Usage of `Map.forEachLayerAtPixel`
+
+Due to performance considerations, the layers in a map will sometimes be rendered into one
+single canvas instead of separate elements.
+This means `Map.forEachLayerAtPixel` will bring up false positives.
+
+The easiest solution to avoid that is to assign different `className` properties to each layer like so:
+```js
+new Layer({
+   // ...
+   className: 'my-layer'
+})
+```
+
+Please note that this may incur a significant performance loss when dealing with many layers and/or
+targetting mobile devices.
+
 ##### Removal of `TOUCH` constant from `ol/has`
 
 If you were previously using this constant, you can check if `'ontouchstart'` is defined in `window` instead.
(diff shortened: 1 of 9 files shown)


changelog/upgrade-notes.md
```diff
@@ -4,6 +4,14 @@
 
 #### Backwards incompatible changes
 
+##### Usage of `Map.forEachLayerAtPixel`
+
+Due to performance optimizations, this method might bring up false positives unless the
+canvas reuse optimization is disabled by calling `disableCanvasReuse()` from the `ol/render` module.
+
+Please note that this may have incur a significant performance loss when dealing with many layers and/or
```
marcjansen:
```suggestion
Please note that this may incur a significant performance penalty when dealing with many layers and/or
```

ahocevar:
The way it is now, there is no way back after calling `disableCanvasReuse()`. Why not instead have a function `disableCanvasReuse(boolean)`, so you can also call `disableCanvasReuse(false)` to get back the optimizations?

ahocevar:
The preferred way to disable this optimization is to set a `className` on the layer. The optimiziation is only applied to layers with the same `className`, the default is `ol-layer`. I think this should be documented, because I had also forgotten about it.

jahow:
> The preferred way to disable this optimization is to set a `className` on the layer. The optimiziation is only applied to layers with the same `className`, the default is `ol-layer`. I think this should be documented, because I had also forgotten about it.

Nice, I've reworked the PR to only be about documentation. Thanks

src/ol/layer/Layer.js
```diff
@@ -35,6 +35,7 @@ import SourceState from '../source/State.js';
  * @property {import("../PluggableMap.js").default} [map] Map.
  * @property {RenderFunction} [render] Render function. Takes the frame state as input and is expected to return an
  * HTML element. Will overwrite the default rendering for the layer.
+ * @property {string} [className='ol-layer'] A CSS class name to set to the layer element.
```
ahocevar: This should be added to the options typedefs of all layers, like `TileLayer`, `VectorLayer` etc.

PR #9974, opened 2019-09-24
Introduces the _withCredentials flag

KaiVolland:
This introduces the `_withCredentials` flag and the corresponding setter `setWithCredentials` to the `featureloader`.

It is an adaption of #9865 created by @mstenta.
Thx for your work. :muscle: 

Closes #9865

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -5,6 +5,13 @@ import {VOID} from './functions.js';
 import FormatType from './format/FormatType.js';
 
 /**
+ *
+ * @type {boolean} withCredentials Compare https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/
+ * @private
+ */
+let withCredentials = false;
+
+/**
  * {@link module:ol/source/Vector} sources use a function of this type to
  * load features.
  *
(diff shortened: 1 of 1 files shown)


src/ol/featureloader.js
```diff
@@ -131,3 +139,13 @@ export function xhr(url, format) {
       }
     }, /* FIXME handle error */ VOID);
 }
+
+/**
+ * Setter for the withCredentials configuration for the XHR.
+ *
+ * @param {boolean} withCredentials The value of withCredentials to set.
+ * Compare https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/
+ */
```
jahow: This is missing an @api tag I believe ;)

src/ol/featureloader.js
```diff
@@ -4,6 +4,13 @@
 import {VOID} from './functions.js';
 import FormatType from './format/FormatType.js';
 
+/**
+ *
+ * @type {boolean} withCredentials Compare https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/
+ * @private
+ */
+let _withCredentials = false;
```
tschaub: Super minor, but we don't really need to prefix this with `_` (naming it `withCredentials` is fine).

KaiVolland: I removed the `_` and renamed the variable in the setter.

Issue #9977, opened 2019-09-24
Simplified geometry cache grows without bound


tschaub:
When rendering, we call `geometry.getSimplifiedGeometry()` with a squared tolerance based on the view resolution.  If this results in a simplification, we [keep the result](https://github.com/openlayers/openlayers/blob/b97647d39b2d382881711bbae11de4f6084a3a65/src/ol/geom/SimpleGeometry.js#L118) in a cache on the geometry.  The only time we remove items from this cache is [when the geometry revision changes](https://github.com/openlayers/openlayers/blob/b97647d39b2d382881711bbae11de4f6084a3a65/src/ol/geom/SimpleGeometry.js#L99) (e.g. if the coordinates are modified).  This means these cached simplified geometries just keep piling up in applications that allow non-integer zoom levels and render at resolutions that result in geometry simplification (a common scenario).

We should be limiting the size of this cache.  Limiting to 1 would make things somewhat efficient for panning.  Something more sophisticated might be worthwhile, but could also be overkill.

tschaub:
Fixed by #9981.

PR #9981, opened 2019-09-24
Remove simplified geometry cache

simonseyock:
Removed simplified geometry cache because it would bloat memory usage if used with non-integer zoom levels.


tschaub:
Fixes #9977.  Thanks @simonseyock.

For others following along, the `geometry.simplify()` method no longer returns a cached result.  Within the library, we now call `geometry.simplifyTransformed()`, which is a memoized function - caching the result from a single call.

Issue #8120, opened 2018-04-24
TypeScript (Help Wanted)


tschaub:
I'd like to see if we can publish TypeScript definitions for the 5.0 release.  My understanding is that these can be published as [`@types/ol`](https://www.npmjs.com/package/@types/ol) or in the `ol` package itself.  I assume we'll need more than a single `index.d.ts` since we do not make everything exportable from a single module.  I'm hoping we can automatically generate the TypeScript definitions from existing JSDoc annotations in the code.  Perhaps this will be a migration path away from JSDoc annotations to authoring in TypeScript, but initially I think our goal should be to publish TypeScript definitions that are in sync with the code.

I'll be doing some more reading, but if anybody has experience and knowledge that they can spread, please let us know.

marcjansen:
Ping @kaivolland @atd-schubert

tschaub:
@yairtawil - given your work on DefinitelyTyped/DefinitelyTyped#22894, I'm curious what you would suggest in terms of automating TypeScript definition files.

atd-schubert:
As long as you can also insert your TypeScript type definitions into your own module you should not apply them to DefinitelyTyped (DT). In addition there is [already a definition on DT](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ol).

My recommendation is to insert a new definition based on the existing one with the help of the current definition author (@yairtawil). After that you can set the current type definition on DT as deprecated and maintain it in your module directly.

I can also help you a litte bit, if you wish.

marcjansen:
Please help, @atd-schubert. Your feedback is much appreciated.

chrismayer:
Wouldn't it make sense to have more members defined in the TypeScript types? In the Definition Files at https://github.com/DefinitelyTyped/DefinitelyTyped/ all classes are modelled like this

```
import * as ol from 'openlayers';

export default ol.layer.Vector;
``` 

Does it make sense to have the members like ``extent`` and ``visible`` defined there as well?

I just started with this TypeScript things, so sorry if I am wrong.

chrismayer:
> Maybe another topic for our FOSSGIS Hacking-Event in Essen 

Good idea ...

beginor:
maybe it is time to switch to typescript 😄

jumpinjackie:
I've already made such a JSDoc plugin to auto-generate `d.ts` files from JSDoc documentation.

https://github.com/jumpinjackie/jsdoc-typescript-plugin

[...]

PR #10000, opened 2019-09-25
Adopt JSDoc annotations to allow '.d.ts' generation

KaiVolland:
This introduces some smaller changes to the JSDoc annotations.
In order to auto-generate `.d.ts` files with [types-ol](https://github.com/hanreev/types-ol) from @hanreev as stated [here](https://github.com/openlayers/openlayers/issues/8120#issuecomment-534981112) some smaller changes to the docs are necessary.

This:
- removes a `@fires` annotation from `LabelCache.js`.
- refactors the `TEXT_ALIGN` enum in the `TextBuilder.js`
- fixes an invalid description of a `@type` anotation

### DIFF
--- a/src/ol/render/canvas/TextBuilder.js
+++ b/src/ol/render/canvas/TextBuilder.js
@@ -15,17 +15,17 @@ import TextPlacement from '../../style/TextPlacement.js';
  * @enum {number}
  */
 export const TEXT_ALIGN = {
-  left: 0,
-  end: 0,
-  center: 0.5,
-  right: 1,
-  start: 1,
-  top: 0,
-  middle: 0.5,
-  hanging: 0.2,
-  alphabetic: 0.8,
-  ideographic: 0.8,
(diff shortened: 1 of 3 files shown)


src/ol/render/canvas/TextBuilder.js
```diff
@@ -14,18 +14,19 @@ import TextPlacement from '../../style/TextPlacement.js';
...
+  end: 0,
+  center: 0.5,
+  right: 1,
+  start: 1,
+  top: 0,
+  middle: 0.5,
+  hanging: 0.2,
+  alphabetic: 0.8,
+  ideographic: 0.8,
+  bottom: 1
```
tschaub: At some point we should check to see whether this is still true for any minifiers, but in the past, we would quote these so they would not be minified.  Since we get values from this enum based on user-provided strings, it is important that these properties are not renamed by a minifier.

PR #10003, opened 2019-09-25
Add back Select interaction without overlay

tkohr:
This PR adds back the deleted Select interaction with the difference that the interaction displays selections by changing the style of the feature instead of adding an overlay to the map. It is based on the work done in #5279.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -275,14 +275,6 @@ The `ol/source/Vector#refresh()` method now removes all features from the source
 
 The `getGetFeatureInfoUrl` of `ol/source/ImageWMS` and `ol/source/TileWMS` is now called `getFeatureInfoUrl`.
 
-##### Removal of `SelectInteraction`
-
-The `SelectInteraction` is removed. There are two examples ([Select Features by Hover](https://openlayers.org/en/master/examples/select-hover-features.html) and [Select multiple Features](https://openlayers.org/en/master/examples/select-multiple-features.html) which show how similar results can be achieved by using more basic methods.
-
-##### `getFeaturesAtPixel` always returns an array
-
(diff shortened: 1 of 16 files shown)


changelog/upgrade-notes.md
```diff
@@ -275,14 +275,6 @@ The `ol/source/Vector#refresh()` method now removes all features from the source
 
 The `getGetFeatureInfoUrl` of `ol/source/ImageWMS` and `ol/source/TileWMS` is now called `getFeatureInfoUrl`.
 
-##### Removal of `SelectInteraction`
-
-The `SelectInteraction` is removed. There are two examples ([Select Features by Hover](https://openlayers.org/en/master/examples/select-hover-features.html) and [Select multiple Features](https://openlayers.org/en/master/examples/select-multiple-features.html) which show how similar results can be achieved by using more basic methods.
-
-##### `getFeaturesAtPixel` always returns an array
```
simonseyock: this upgrade note should not be removed

src/ol/interaction/Select.js
```diff
@@ -0,0 +1,493 @@
...
+ * @property {import("../Collection.js").default<import("../Feature.js").default>} [features]
+ * Collection where the interaction will place selected features. Optional. If
+ * not set the interaction will create a collection. In any case the collection
+ * used by the interaction is returned by
+ * {@link module:ol/interaction/Select~Select#getFeatures}.
+ * @property {FilterFunction} [filter] A function
+ * that takes an {@link module:ol/Feature} and an
+ * {@link module:ol/layer/Layer} and returns `true` if the feature may be
+ * selected or `false` otherwise.
+ * @property {boolean} [wrapX=true] Wrap the world horizontally on the selection
```
simonseyock: this property is not used anymore i think

PR #10006, opened 2019-09-25
Check hits for wrapped around geometries

### LINKED ISSUE #9939: Polygon that straddles the international dateline
**Describe the bug**
I am using map.forEachFeatureAtPixel to highlight polygons as I move the cursor around.  It works on half of the polygon but as I move across the International Date Line it stops working.  

**To Reproduce**
1. Create a map using EPSG:3857 and a vector layer with wrapX = true
2. Create a EPSG:3857 polygon that straddles the Date Line (if it were in lat/lon it would have X coordinates from say -160 to -200)
3. Listen for 'pointermove' or 'singleclick' on map
4. call map.forEachFeatureAtPixel using event.pixel and a callback(feature,layer)
5. in Map.js there is a function forEachFeatureAtCoordinate that calculates worldsaway.  If the coordinate calculated by the pixel location is outside the bounds of the projection (ie -180 o 180 but in EPSG:3857) then that translates the coordinate to be on the opposite side of the world but it is not and thus does not return a hit for that location.

**Expected behavior**
I would expect that as I move across a polygon that it would continue to highlight even if I cross the International Date Line.

simonseyock:
Check one world before and after for wrapped around geometries.
If the geometries are off more than one world they are not detected.

A new parameter `checkWrapped` for the `*AtPixel` methods disables the execution of the hitDetection for more than one world.

Fixes #9939

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -75,6 +75,8 @@ import {toUserCoordinate, fromUserCoordinate} from './proj.js';
  * will be tested for features. By default, all visible layers will be tested.
  * @property {number} [hitTolerance=0] Hit-detection tolerance in pixels. Pixels
  * inside the radius around the given position will be checked for features.
+ * @property {boolean} [checkWrapped=true] Check-Wrapped Will check for for wrapped geometries inside the range of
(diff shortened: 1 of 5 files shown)


tschaub:
Thanks for the fix @simonseyock!

PR #10011, opened 2019-09-26
Fix a type checking error on the WebGL points renderer

jahow:
I guess someone needed to prove the new typecheck step was effective!


ahocevar:
Although it looks like there is still something wrong.

PR #10012, opened 2019-09-26
Use the user projection for the map event coordinate

tschaub:
Our map browser event has a `coordinate` property.  This branch makes it so that coordinate is in the user projection when one is set.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1237,6 +1237,17 @@ class View extends BaseObject {
    * @api
    */
   adjustResolution(ratio, opt_anchor) {
+    const anchor = opt_anchor && fromUserCoordinate(opt_anchor, this.getProjection());
+    this.adjustResolutionInternal(ratio, anchor);
+  }
+
+  /**
+   * Multiply the view resolution by a ratio, optionally using an anchor. Any resolution
+   * constraint will apply.
+   * @param {number} ratio The ratio to apply on the view resolution.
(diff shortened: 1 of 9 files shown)


src/ol/View.js
```diff
@@ -1237,6 +1237,21 @@ class View extends BaseObject {
    * @api
    */
   adjustResolution(ratio, opt_anchor) {
+    let anchor = opt_anchor;
+    if (anchor) {
+      anchor = fromUserCoordinate(opt_anchor, this.getProjection());
```
jahow: How about:
```js
let anchor = opt_anchor && fromUserCoordinate(opt_anchor, this.getProjection());
```

src/ol/View.js
```diff
@@ -1448,6 +1463,21 @@ class View extends BaseObject {
    * @api
    */
   endInteraction(opt_duration, opt_resolutionDirection, opt_anchor) {
+    let anchor = opt_anchor;
+    if (anchor) {
+      anchor = fromUserCoordinate(anchor, this.getProjection());
```
jahow: Same as for `beginInteraction`.

src/ol/View.js
tschaub: Done in 5d4e77151c3eb2c8cd509969a291f4d6c9e3402c.

src/ol/View.js
tschaub: 5d4e77151c3eb2c8cd509969a291f4d6c9e3402c

PR #10014, opened 2019-09-26
Do not run only one test

ahocevar:
Now let's see which tests are really failing.


ahocevar:
Closing, because @jahow is working on this as well.

PR #10015, opened 2019-09-26
Fix hit detection test for wrapped geometries

jahow:
Tests were added in https://github.com/openlayers/openlayers/pull/10006 but were using untransformed coordinates.


tschaub:
Thanks for fixing this up @jahow.

PR #10028, opened 2019-09-26
Fix zoomByDelta test

ahocevar:
Fixes a test that is currently failing on master.


ahocevar:
@tschaub has a better fix for this. Closing.

PR #10031, opened 2019-09-26
Cache starting coordinates and add mapBrowserEvent to translateEvent

flexjoly:
If the starting-coordinates in Translate and TranslateEvent are cached, they can easily be used to undo the translate.

I also added mapBrowserEvent just like it is done in (some) other interactions (ie. Select).

I have tested it in my own application. But I don't have the required test-environment (and I have no experience with that). But I hope it is ok. 😬


flexjoly:
I have seen the errors, working on them!

PR #10035, opened 2019-09-26
Support building the legacy build on windows

### LINKED ISSUE #9891: Legacy build and Worker generation
The generation of Workers dosen't work for the legacy builds.
Related: https://github.com/openlayers/openlayers/pull/9672

tschaub:
This should make it so people can build the legacy build on Windows.  Ideally, people that still use the legacy build use a hosted version of the legacy build instead of building it themself.  But it sounds like that isn't happening.

Fixes #9891.  See also #9672.

### DIFF
--- a/tasks/serialize-workers.js
+++ b/tasks/serialize-workers.js
@@ -14,7 +14,7 @@ async function build(input, {minify = true} = {}) {
         if (id !== input) {
           return null;
         }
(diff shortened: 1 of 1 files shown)


ahocevar:
I think this can be merged. I just tried a minimal example with the legacy build I created.

Issue #9958, opened 2019-09-23
Add Boolean 'isSnapped' to event or anything so you can find out whether a feature has been 'snapped'


flexjoly:
**Is your feature request related to a problem? Please describe.**
Our markers are styled as a small circle with a text-label. Snap is the only interaction which only matches with the circle/point. 
interaction/Select or map.getFeatureAtPixel also select a feature when clicking/hovering at the text-label. 
So far I could not find any other simple solution/workaround for this. 

Snap does exactly what we need. Only I could not find how to know if Snap found a feature. For it only changed the pixels/coordinate.

**Describe the solution you'd like**
As far as I can see, adding a boolean parameter to current event would solve this. This could be done in de handleEvent():
```
  Snap.prototype.handleEvent = function handleEvent (evt) {
    var result = this.snapTo(evt.pixel, evt.coordinate, evt.map);
    evt.isSnapped = false;
    if (result.snapped) {
      evt.coordinate = result.vertex.slice(0, 2);
      evt.pixel = result.vertexPixel;
      evt.isSnapped = true;
    }
    return PointerInteraction.prototype.handleEvent.call(this, evt);
  };
```
Of course it is better if such a boolean is initiated with the event itself, like the boolean 'dragging' in the MapBrowserEvent.

Another solution would be a new 'snapped' event... but that is more complicated and costs more performance, I think. 

Thanks in advance!

jahow:
The Snap interaction works with geometries whereas hit detection uses the final render, which explains the difference. I'd say everything is working as intended here, maybe you could put the text labels in a different layer which would use the same source and exclude this layer from the hit detection?

It may be easier if you could set up a small reproduction case. Thanks a lot!

flexjoly:
Hi Jahow,

Thanks for your fast reply.

>  hit detection uses the final render

sorry for my newby-reply on this: is there another, better hit-detection except interaction/Select or map.getFeatureAtPixel?
Or do I need to check the hit-dectection somewhere else?


> I'd say everything is working as intended here, 
yes it does!

Therefor I started this as feature-request 😉 

> maybe you could put the text labels in a different layer which would use the same source and exclude this layer from the hit detection?
I looked into that also, but I hoped to I did not need this workaround. Because I need to rewrite and decouple my nice Feature-object 😏 

> 
> It may be easier if you could set up a small reproduction case. Thanks a lot!
If needed I can do that, but my code is a bit large, so I hope it is clear enough without it 😬

[...]

PR #10037, opened 2019-09-26
new SnapEvent and snapped boolean

flexjoly:
With this code Snap will trigger a new SnapEvent. See issue #9958 
It uses a snapCondition like seen in Select. The condition defaults to singleClick.

When a feature is 'snapped' ánd the condition is met, then the SnapEvent is triggered from within the central handleEvent() function. This way it is not limited only to a mouseclick or -move etc.

Also a boolean 'snapped' is set in the MapBrowserEvent. 

### Question
The trigger does NOT handle the return from handleEvent(). I don't know if it should or not. 

**_Is handleEvent() ok like this ?_** :
```
  handleEvent(evt) {
    const result = this.snapTo(evt.pixel, evt.coordinate, evt.map);
    evt.snapped = result.snapped;
    if (result.snapped) {
      evt.coordinate = result.vertex.slice(0, 2);

      if (this.snapCondition_(evt)) {
        this.dispatchEvent(
          new SnapEvent(SnapEventType.SNAP,
            result, evt.map.getFeaturesAtPixel(evt.pixel), evt));
      }
    }
    return super.handleEvent(evt);
  }
```
**_or should it be 'return this.dispatchEvent_**' ? : 
```
       return this.dispatchEvent(
          new SnapEvent(SnapEventType.SNAP,
            result, evt.map.getFeaturesAtPixel(evt.pixel), evt));
```

PS.
(sorry for the extra edited lines because of spaces/tabs; I had a hard time to get my editor (netbeans on windows) to the requested format)

### DIFF
--- a/src/ol/MapBrowserEvent.js
+++ b/src/ol/MapBrowserEvent.js
@@ -50,6 +50,13 @@ class MapBrowserEvent extends MapEvent {
      */
     this.dragging = opt_dragging !== undefined ? opt_dragging : false;
 
+    /**
+     * Indicates if a feature is currently snapped. Default is `false`.
+     *
+     * @type {boolean}
+     */
+    this.snapped_ = false;
+
   }
 
   /**
@@ -65,6 +72,18 @@ class MapBrowserEvent extends MapEvent {
   }
   set pixel(pixel) {
     this.pixel_ = pixel;
+  }
+
+  /**
+   * Indicates if a feature is currently snapped. Default is `false`.
+   * @type {boolean}
+   * @api
+   */
+  get snapped() {
+    return this.snapped_;
+  }
+  set snapped(snapped) {
+    this.snapped_ = snapped;
   }
 
   /**
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -2,6 +2,7 @@
  * @module ol/interaction/Snap
  */
 import {getUid} from '../util.js';
+import Collection from '../Collection.js';
 import CollectionEventType from '../CollectionEventType.js';
 import {distance as coordinateDistance, squaredDistance as squaredCoordinateDistance, closestOnCircle, closestOnSegment, squaredDistanceToSegment} from '../coordinate.js';
 import {listen, unlistenByKey} from '../events.js';
@@ -86,7 +87,7 @@ class SnapEvent extends Event {
   /**
    * @param {SnapEventType} type The event type.
    * @param {Result} result Snap result array.
-   * @param {Array<import("../Feature.js").default>} features Array with the snapped features.
+   * @param {Collection<import("../Feature.js").default>} features The snapped features.
    * @param {import("../MapBrowserEvent.js").default} mapBrowserEvent Associated
    *     {@link module:ol/MapBrowserEvent}.
    */
@@ -101,8 +102,8 @@ class SnapEvent extends Event {
     this.result = result;
 
     /**
-     * Snapped features array.
-     * @type {Array<import("../Feature.js").default>}
+     * Snapped features collection.
+     * @type {Collection<import("../Feature.js").default>}
      * @api
      */
     this.features = features;
@@ -321,9 +322,10 @@ class Snap extends PointerInteraction {
       evt.pixel = result.vertexPixel;
 
       if (this.snapCondition_(evt)) {
+        const features = new Collection([evt.map.getFeaturesAtPixel(evt.pixel)]);
         this.dispatchEvent(
           new SnapEvent(SnapEventType.SNAP,
-            result, evt.map.getFeaturesAtPixel(evt.pixel), evt));
+            result, features, evt));
       }
     }
     return super.handleEvent(evt);
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -135,6 +135,7 @@ class SnapEvent extends Event {
  *       source: source
  *     });
  *
+ * @fires SnapEvent
  * @api
  */
 class Snap extends PointerInteraction {
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -54,6 +54,11 @@ const SnapEventType = {
  * @property {number} [pixelTolerance=10] Pixel tolerance for considering the pointer close enough to a segment or
  * vertex for snapping.
  * @property {import("../source/Vector.js").default} [source] Snap to features from this source. Either this option or features should be provided
+ * @property {import("../events/condition.js").Condition} [snapCondition] A function that
+ * takes an {@link module:ol/MapBrowserEvent~MapBrowserEvent} and returns a
+ * boolean to indicate whether that event should be handled. This is the event
+ * for the snapped features as a whole. By default, this is
+ * {@link module:ol/events/condition~singleClick}. Clicking on a snapped feature will trigger the SnapEvent
  */
 
 
@@ -78,7 +83,7 @@ function getFeatureFromEvent(evt) {
 class SnapEvent extends Event {
   /**
    * @param {SnapEventType} type The event type.
-   * @param {Array} result Snap result array.
+   * @param {Result} result Snap result array.
    * @param {Array<import("../Feature.js").default>} features Array with the snapped features.
    * @param {import("../MapBrowserEvent.js").default} mapBrowserEvent Associated
    *     {@link module:ol/MapBrowserEvent}.
@@ -88,7 +93,7 @@ class SnapEvent extends Event {
 
     /**
      * Snapped result array.
-     * @type {Array}
+     * @type {Result}
      * @api
      */
     this.result = result;
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -5,8 +5,10 @@ import {getUid} from '../util.js';
 import CollectionEventType from '../CollectionEventType.js';
 import {distance as coordinateDistance, squaredDistance as squaredCoordinateDistance, closestOnCircle, closestOnSegment, squaredDistanceToSegment} from '../coordinate.js';
 import {listen, unlistenByKey} from '../events.js';
+import Event from '../events/Event.js';
 import EventType from '../events/EventType.js';
 import {boundingExtent, createEmpty} from '../extent.js';
+import {singleClick} from '../events/condition.js';
 import {TRUE, FALSE} from '../functions.js';
 import GeometryType from '../geom/GeometryType.js';
 import {fromCircle} from '../geom/Polygon.js';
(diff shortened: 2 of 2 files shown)


simonseyock:
Hello, thank you very much for the effort!

Here are some thoughts I have about this:

I don't think the `Snap` interaction should fire `SnapEvent` at all. It's responsibility is to change the `MapBrowserEvent`s coordinates for all following interactions but not actually to interact with directly.

If you would want to get snapped features you could use a `Select` interaction afterwards.

The `snapped` booelan is maybe interesting for following interactions though. Maybe it's also reasonable to save the unchanged coordinates somewhere? Maybe instead of the boolean (you could than just check these coordinates on equality)?

@jahow what do you think about this?

flexjoly:
Hi @simonseyock ,

The snapEvent was an idea of @jahow 😜 

> I don't think the `Snap` interaction should fire a `SnapEvent`. It's responsibility is to change the `MapBrowserEvent`s coordinates for all following interactions but not actually to interact with directly.
> If you would want to get snapped features you could use a `Select` interaction afterwards.

As stated in the issue #9958 the difference between Select and Snap is that Snap only finds the Vertex while Select also matches on textLabels etc. 

I do agree that the SnapEvent has an overlap with Select. 
So that is why my first FR was only the snapped-boolean, so it could work together with Select. 

At the other hand it seemed that the next interactions do not always get the correct snapped-boolean. Because every change of the mouse fires a new mapBrowserEvent.
And because the SnapEvent was suggested, I did not look into this anymore.

> The `snapped` boolean is maybe interesting for following interactions though. Maybe it's also reasonable to save the unchanged coordinates somewhere? Maybe instead of the boolean (you could than just check these coordinates on equality)?

Nice idea. Although I don't see much purpose for it. Is it to undo the Snap? 
Is that really needed? 
For if not then checking a simple boolean costs less performance than checking the coordinates each time.

Greetz, flexjoly

simonseyock:
Oh I did not see that discussion 🙃 

I am not convinced that the snap interaction is the right way to solve this problem. But i agree it is sensible to expose some more information over snapping.

A feature collection would not be neccessary though, because you could still use a select for this or not?

simonseyock:
-> I probably should continue any comments in the issue

flexjoly:
> A feature collection would not be neccessary though, because you could still use a select for this or not?

**Pros for the feature collection**:
- Snap can be used on its own, and don't need select
- Other interactions also add the features to their corresponding event

**Cons**: 
- The `map.getFeaturesAtPixel` can also be done by the listener to the snapEvent..., since the found pixel is also put into the event, and mapBrowserEvent has the pixel also.
- Costs a little performance

**map.getFeaturesAtPixel in eventHandler()**
I was just using/debugging it (finding a minor error again 😖 )

I don't like the fact that I use `map.getFeaturesAtPixel` in eventhandler() to make the feature-collection. It would be better to let toSnap() add the snapped-features to result.  (lol, I just see that you mention the same in response to the issue)

But toSnap() is a bid complex, so I don't trust myself to add the features there.

flexjoly:
would this be the correct way to add the snapped feature to the result array at the end of the [snapTo() method](https://github.com/openlayers/openlayers/blob/40f85100837dcda52d1c753ac592c86ee913674f/src/ol/interaction/Snap.js#L459)? 
```
    if (snapped) {
      vertexPixel = [Math.round(vertexPixel[0]), Math.round(vertexPixel[1])];
      feature = closestSegmentData.feature;
    }
```

NB!! I don't know anything about more complex features and cannot debug them.

This is the map I making. 
It combines at least Draw, Translate and Snap and gets modals add info to the markers and is going to be used on mobile most of the time.
osm testing

simonseyock:
This looks good to me! If you only find one feature ever you could let the snapevent only have a `feature` property instead of a `features` property.

simonseyock:
I don't think it is neccessary to add the `Result` object to the snapevent. The vertices should not be of that much interest to the user. It is possible to find the point in the geometry of the features anyhow.
Also the `get snapped` and `set snapped` is not needed as ahocevar mentioned before.

I find it kind of strange that the `snapCondition` not handles if the interactions snaps or not but only if the event gets fired. I think the interaction as a whole should only snap if the `snapCondition` is met. If it is desired to turn the `SnapEvent` off a boolean flag would be sufficient. Then I would also set the condition to `module:ol/events/condition.always` to maintain the current behaviour.

flexjoly:
> The code in the `snapTo` looks good to me! If you only find one feature ever you could let the snapevent only have a `feature` property instead of a `features` property.

Ok! thanks for checking!

One feature... I prefer this also, but (almost?) all other scripts make an array or collection, so I figured Snap should do the same.

> I don't think it is neccessary to add the `Result` object to the snapevent. The vertices should not be of that much interest to the user. It is possible to find the point in the geometry of the features anyhow.
> Also the `get snapped` and `set snapped` is not needed as ahocevar mentioned before.

ok! I will remove the Result-object and the get/set

> I find it kind of strange that the `snapCondition` not handles if the interactions snaps or not but only if the event gets fired. 

I am not sure if I understand you correctly

My idea was:
- Snap works the same as before
- SnapEvent should only be fired if the snapCondition is met, for example when the user clicked

In my case I need Snap for Translate AND the SnapEvent at a SingleClick. 
But someone else might need the SnapEvent when hovering over to show a tooltip with info.

> I think the interaction as a whole should only snap if the `snapCondition` is met. If it is desired to turn the `SnapEvent` off a boolean flag would be sufficient. Then I would also set the condition to `module:ol/events/condition.always` to maintain the current behaviour.

In this case I suggest to add 'Condition' next to 'SnapCondition' like Select has:
- Condition works on the snapping itself. 
- SnapCondition handles the firing of SnapEvent

I like the boolean idea, since SnapEvent is extra indeed; but WITH the snapCondition.

'SnapCondition' is not the best name, I did think about a different name like EventCondition.... but I did not like that too. Any suggestions?

simonseyock:
> One feature... I prefer this also, but (almost?) all other scripts make an array or collection, so I figured Snap should do the same.

An array would suggest that there is a variable amount of features. But it is always exactly one.

>In this case I suggest to add 'Condition' next to 'SnapCondition' like Select has:
>    Condition works on the snapping itself.
>    SnapCondition handles the firing of SnapEvent
> I like the boolean idea, since SnapEvent is extra indeed; but WITH the snapCondition.
>
> 'SnapCondition' is not the best name, I did think about a different name like EventCondition.... but I did not like that too. Any suggestions?

I think it would be enough to have just one `Condition` for the whole thing and no boolean either. If you need something on `click` and something on `pointermove` you could still just use two interactions.

jahow:
Hi,

I generally agree with @simonseyock here, we have to keep this contribution as simple as possible and not deviate too much from the original intention of the interaction. The `snap` event should have exactly one feature and be dispatched when the `snapped` state switches from false to true. IMO it would make sense to have another `snap` event dispatched when the `snapped` state goes to false again (ie the cursor detaches from the feature), with the `feature` property undefined or null.

flexjoly:
> I generally agree with @simonseyock here, we have to keep this contribution as simple as possible and not deviate too much from the original intention of the interaction. The `snap` event should have exactly one feature and be dispatched when the `snapped` state switches from false to true. IMO it would make sense to have another `snap` event dispatched when the `snapped` state goes to false again (ie the cursor detaches from the feature), with the `feature` property undefined or null.

So to summarize:
- Snap has one `Condition` (default: `module:ol/events/condition.always`) to activate snapping
- when `snapped` > fire `SnapEvent` and remember it has `snapped` a feature
- after de-snapping > fire `SnapEvent` with `feature`=null/undefined
- `SnapEvent` has one or no `feature` attached (captured in `snapTo`)
- `SnapEvent` has no `Result` attached
- remove get/set for `snapped` in `MapBrowserEvent`

simonseyock:
I would remember the last snapped feature. If it stays the same, do nothing. If it changes fire `desnap` (or `release`) and then `snap`. Desnap could than include the feature it was snapped to last.

What also can happen is that it snaps to another vertex of the same feature ...

simonseyock:
Also you could fire just a new `snap` instead of both events on change.

ahocevar:
Firing events comes with a cost, so it would be good to assess what the original requirement is: knowing when a vertex was snapped. Right? To that purpose, whenever the snap interaction modifies `event.coordinate` and `event.pixel`, setting an `event.snapped` property to true should be enough.

To get the original coordinate, you can always do `map.getEventCoordinate(event.originalEvent)`. Or for the original pixel: `map.getEventPixel(event.originalEvent)`.

If I read this pull request and the underlying issue correctly, this is also the original intention of @flexjoly. Please correct me if I'm wrong or missing something.

jahow:
> If I read this pull request and the underlying issue correctly, this is also the original intention of @flexjoly. Please correct me if I'm wrong or missing something.

You're right, I was the one suggesting adding an event for this snapped state, since it would make it much easier to react to it by e.g. showing a popup, changing a style and whatnot.

You're probably right though, modifying the event should be enough since then the application code can react to it accordingly.

flexjoly:
So you suggest not making a `SnapEvent`, but only `mapBrowserEvent.snapped` boolean?

As @simonseyock mentioned before: `map.getFeaturesAtPixel` can give a different result then what `Snap` found. 
Changing the boolean to `mapBrowserEvent.snappedFeature' would solve this.

flexjoly:
hhhmmmm Just looking into `map.getFeaturesAtPixel`. Deeper into the code it is called `getFeaturesAtCoordinate` 

I would like to say that those are 2 different things to me 😀 
- `getFeaturesAtPixel` should search for any aspect (like bigger marker or textlabel) of a `feature` at that `pixel`
- getFeaturesAtCoordinate should only match at/within the `Coordinates` of a `Feature`.... with some `hitTolerance`

jahow:
`getFeaturesAtPixel` and hit detection in general is always made against a render result in OL. Without this, how would you for example be able to select a linestring? The render width has to be taken into account, otherwise you'll just have to come up with an arbitrary hit tolerance.

> So you suggest not making a SnapEvent, but only mapBrowserEvent.snapped boolean?

Pretty much.

> As @simonseyock mentioned before: map.getFeaturesAtPixel can give a different result then what Snap found.
> Changing the boolean to mapBrowserEvent.snappedFeature would solve this.

I'd personally agree to have this on the event, and then we can call it a day. @ahocevar what do you think of this?

flexjoly:
> If you can elaborate a bit on what exactly you tried to achieve and what did not work, we should be able to either come up with a solution using existing features, or provide guidance on what should be added.

I thought it is better to answer this in [the issue ](https://github.com/openlayers/openlayers/issues/9958#issuecomment-536995766)

flexjoly:
Problem solved by: setting `renderBuffer` in `VectorLayer`, [See answer](https://github.com/openlayers/openlayers/issues/9958#issuecomment-537112018)

PR #10042, opened 2019-09-27
Better typing

fredj:
Small typing fixes found by setting `strict` to `true` in `tsconfig.json`

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -232,7 +232,7 @@ class PluggableMap extends BaseObject {
 
     /**
      * @private
-     * @type {Array<import("./events.js").EventsKey>}
+     * @type {?Array<import("./events.js").EventsKey>}
      */
     this.layerGroupPropertyListenerKeys_ = null;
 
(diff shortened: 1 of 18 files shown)


fredj:
The strict mode yields around 3500 error but I hope most of them are easily fixable.

fredj:
For example with this code:
```js
/**
 * The map pixel relative to the viewport corresponding to the original browser event.
 * @type {import("./pixel.js").Pixel}
 */
this.pixel_ = null;
```
typescript complains that _Type 'null' is not assignable to type 'number[]'.ts(2322)_
We could change the type to be `@type {null|import("./pixel.js").Pixel}`

fredj:
I've added two commit that adds more typecast and mark properties as nullable

src/ol/PluggableMap.js
```diff
@@ -604,7 +604,7 @@ class PluggableMap extends BaseObject {
     }
     const options = opt_options || /** @type {AtPixelOptions} */ ({});
```
tschaub: Not related to your change here, but I think we should also address type casts like the one above.  If an empty object is not assignable to `AtPixelOptions`, we shouldn’t be using it as one.

fredj: You're right: these typecast are not necessary. Typescript is able to deduce that an empty object is compatible with `AtPixelOptions`.

pull request to come.

fredj: See #10076

PR #10043, opened 2019-09-27
User coordinates during snapping

tschaub:
Following up on #10009, this makes the snap interaction work when a user projection is set.

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -14,6 +14,7 @@ import PointerInteraction from './Pointer.js';
 import {getValues} from '../obj.js';
 import VectorEventType from '../source/VectorEventType.js';
 import RBush from '../structs/RBush.js';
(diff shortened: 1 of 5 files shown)


src/ol/interaction/Snap.js
```diff
@@ -613,17 +622,4 @@ class Snap extends PointerInteraction {
...
- * @param {SegmentData} a The first segment data.
- * @param {SegmentData} b The second segment data.
- * @return {number} The difference in distance.
- * @this {Snap}
- */
-function sortByDistance(a, b) {
-  const deltaA = squaredDistanceToSegment(this.pixelCoordinate_, a.segment);
-  const deltaB = squaredDistanceToSegment(this.pixelCoordinate_, b.segment);
-  return deltaA - deltaB;
-}
```
tschaub: Previously we were sorting the full list of segments but only using the first.  Now we go through the list once to get the closest instead of sorting the whole list.

PR #10047, opened 2019-09-27
Apidoc changes

KaiVolland:
This introduces some changes to the API doc navigation.
The target is to improve the user experience when using the API-Doc.

The attempt is to improve the UI to let the user find relevant informations faster.
Steps done so far are:

- Removes modules from the apidoc navigation --> Still reachable as sublink of classes (`↳ module`)
- Simplifies classnames in the Navigation
- Adds a weight paramater to the search to be more accurate on searchresults --> Entering "Map" shows the `ol/Map~Map` class first

This is currently WIP so please share your thoughts.


jahow:
Hey,

I think it's a very good start. How about writing the name of the module more explicitly, something like:
```
Map
 ↳ part of ol/Map
```

marcjansen:
@tschaub and/or @ahocevar would be noice if you could take a look at this anf give feedback. TIA

jahow:
Maybe a pointless question, but weren't the `.less` files used for building the css? I see they were removed

KaiVolland:
> Maybe a pointless question, but weren't the `.less` files used for building the css? I see they were removed

I didn't see any relation or any process that used the less files so i removed them. As what i have seen the styling is all done in the `jaguar.css`. ¯\_(ツ)_/¯

PR #10048, opened 2019-09-27
Early EMPTY state for VectorRenderTile

ahocevar:
The tile queue and loading logic was meant to know if a tile is EMPTY (i.e. not provided by the source) before loading. Previously we tried to change that logic to handle an IDLE -> EMPTY transition during loading, but that did not work properly.

This pull request reverts that change and changes the VectorTile source to set the EMPTY state upon tile creation.

### DIFF
--- a/src/ol/TileQueue.js
+++ b/src/ol/TileQueue.js
@@ -112,12 +112,10 @@ class TileQueue extends PriorityQueue {
       if (state === TileState.ABORT) {
         abortedTiles = true;
       } else if (state === TileState.IDLE && !(tileKey in this.tilesLoadingKeys_)) {
+        this.tilesLoadingKeys_[tileKey] = true;
+        ++this.tilesLoading_;
+        ++newLoads;
         tile.load();
-        if (tile.getState() === TileState.LOADING) {
-          this.tilesLoadingKeys_[tileKey] = true;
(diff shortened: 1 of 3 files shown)


src/ol/TileQueue.js
```diff
@@ -112,12 +112,10 @@ class TileQueue extends PriorityQueue {
...
       } else if (state === TileState.IDLE && !(tileKey in this.tilesLoadingKeys_)) {
+        this.tilesLoadingKeys_[tileKey] = true;
+        ++this.tilesLoading_;
+        ++newLoads;
         tile.load();
-        if (tile.getState() === TileState.LOADING) {
-          this.tilesLoadingKeys_[tileKey] = true;
-          ++this.tilesLoading_;
-          ++newLoads;
-        }
```
tschaub: Reviewing this last time, I liked that we would never count a tile as loading if the state wasn’t loading.  I trust that this change improves things, but I admit I can’t explain why we would increment the count if the state is not loading.  

Maybe the new logic updates the state after incrementing the count.

ahocevar:
We now never get here when the state is not LOADING. The tile loading sequence is now either IDLE-LOADING-LOADED/ERROR or EMPTY.

PR #10056, opened 2019-09-28
Remove called assert extension

tschaub:
As part of simplifying our test setup (and making it easier to swap out the parts), this removes the `called` extension to expect.js.


tschaub:
Thanks for the late night review!

PR #10067, opened 2019-09-29
fixed jsdoc type annotations

### LINKED ISSUE #10044: Pure option typedefs cause problems in the documentation
It just shows an link to the options instead.

See https://openlayers.org/en/master/apidoc/module-ol_layer_Image-ImageLayer.html

simonseyock:
Fixes `Event` type annotations to `BaseEvent` and removed an empty Option typedef for `ol/image/Layer` which caused the options to not be displayed inline in the docs.

Fixes #10044

### DIFF
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -231,7 +231,7 @@ class Feature extends BaseObject {
    * styles. If it is `null` the feature has no style (a `null` style).
    * @param {import("./style/Style.js").StyleLike} style Style for this feature.
    * @api
-   * @fires module:ol/events/Event~Event#event:change
+   * @fires module:ol/events/Event~BaseEvent#event:change
    */
   setStyle(style) {
     this.style_ = style;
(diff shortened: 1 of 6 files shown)


tschaub:
We should use a less fragile naming convention.  Instead of using the local identifier, we should be using the exported identifier.  This way there is nothing special about the default export - it is just named `default`.

For example, if we had `ol/example.js`
```
const foo = 'foo';
const bar = 'bar';

export {foo as bar}
export default bar;
```

From the outside, these could be referred to as `module:ol/example~bar` (the string `'foo'`) and `module:ol/example~default` (the string `'bar'`).  The local module identifiers are not important, only the exported names (and the default export is really just a named export with the name `default` - there is nothing special about this except it lets people use the shorthand syntax for importing).

simonseyock:
I was not aware that `~default` works. I agree with you, the importing file should not need to know how the default export is named.

simonseyock:
I just tried it and it does not work, but this might be another problem of jsdoc-plugin-typescript

tschaub:
This may also be a problem with how JSDoc comes up with long names.

simonseyock:
Yes, at the moment we convert `import("./example.js/").default` to `module:ol/example~bar` inside of jsdoc-plugin-typescript. Probably because JSDoc cannot handle `module:ol/example~default`.
We could add another rewrite for `module:ol/example~default` though.
Or adding a rewrite so we can use `import("./example.js/").default` everywhere.

ahocevar:
I have investigated this before, and we're running into a JSDoc limitation here.

PR #10074, opened 2019-10-01
Remove offsets from mouse position calculation

### LINKED ISSUE #10073: Overview control reacting strange in v6
**Describe the bug**
When moving the box in the overview the the position of the map is out of place.
The map is moved to the place of the pointer instead of the center of the box.

**To Reproduce**
Open in the example page the overview example and move it around.
As you can see the the map snaps to the extent where the pointer of cursor was not then center of the box

**Expected behavior**
It will move the map to the center of the box

mike-000:
Fixes #10073

Remove unneeded offsets from mouse position calculation as box position was centered in #9919

### DIFF
--- a/src/ol/control/OverviewMap.js
+++ b/src/ol/control/OverviewMap.js
@@ -217,7 +217,7 @@ class OverviewMap extends Control {
     const computeDesiredMousePosition = function(mousePosition) {
       return {
         clientX: mousePosition.clientX,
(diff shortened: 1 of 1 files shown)


tschaub:
The example looks better now: https://2298-4723248-gh.circle-artifacts.com/0/examples/overviewmap.html

fredj:
nice, thanks @mike-000

PR #10078, opened 2019-10-02
Resolve constraints using anchor when cancelling animation

### LINKED ISSUE #10071: Continuous ZoomOnScroll looses the mouse position when using constrainResolution true
**Describe the bug**
on the map with constrainResolution, when zooming with scroll, the position of the map start to change like crazy, doesnt matter if its zoom in or zoom out.

**To Reproduce**
1. set View constrainResolution  true
2. place the mouse far from the map center
3. scroll fast to zoom (when zoom changes more then 2 levels, the map position of the mouse snaps )

**Expected behavior**
keep the mouse position on the map

mike-000:
Fixes #10071

Resolve constraints using anchor when cancelling an active animation with anchor.  This prevents a subsequent default resolve constraints fixing the resolution from the view center which can move the coordinates under the pointer.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -584,13 +584,21 @@ class View extends BaseObject {
    */
   cancelAnimations() {
     this.setHint(ViewHint.ANIMATING, -this.hints_[ViewHint.ANIMATING]);
+    let anchor;
     for (let i = 0, ii = this.animations_.length; i < ii; ++i) {
       const series = this.animations_[i];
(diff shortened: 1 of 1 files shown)


mike-000:
This isn't ready yet.  The View Animation example zoomed out to world extent, flying to Bern and clicking to cancel when slightly more than one world is visible recovers without this code but has a problem with it.

mike-000:
Now replaced by #10120

PR #10083, opened 2019-10-03
webgl points layer example fixes

MoonE:
This PR improves the webgl-points-layer example.
Bugs:
- Remove 'Predefined styles' option from input. When selected it displays undefined as style.
- Catch errors when input string cannot be parsed as JSON.
- Catch errors when input string is not a valid style.
- Do not remove current point layer if the new style is invalid.
- Show 'success' instead of 'invalid' message when the user reverts an invalid style to the previously selected one.

Improvements:
- Also pre-select the circles style in the select input.
- Add a dummy tag that reserves the height for the parsing status message.
- Hide this message when the style is changed via the select input.

### DIFF
--- a/src/ol/layer/WebGLPoints.js
+++ b/src/ol/layer/WebGLPoints.js
@@ -74,10 +74,7 @@ class WebGLPointsLayer extends Layer {
      * @private
      * @type {import('../webgl/ShaderBuilder.js').StyleParseResult}
      */
-    this.parseResult_ = parseSymbolStyle(
-      /** @type {import('../style/LiteralStyle.js').LiteralStyle} */
-      (options.style).symbol
-    );
+    this.parseResult_ = parseSymbolStyle(options.style.symbol);
   }
 
   /**
--- a/src/ol/layer/WebGLPoints.js
+++ b/src/ol/layer/WebGLPoints.js
@@ -4,7 +4,6 @@
 import {assign} from '../obj.js';
 import WebGLPointsLayerRenderer from '../renderer/webgl/PointsLayer.js';
 import {getSymbolFragmentShader, getSymbolVertexShader, parseSymbolStyle} from '../webgl/ShaderBuilder.js';
-import {assert} from '../asserts.js';
 import Layer from './Layer.js';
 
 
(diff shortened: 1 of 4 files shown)


MoonE:
Is this what you had in mind?
Now the style is parsed in the ol.layer.WebGLPoints class' constructor.
Previously it was only parse on the first rendering during an animation frame which is asynchronous.

src/ol/layer/WebGLPoints.js
```diff
@@ -72,24 +71,24 @@ class WebGLPointsLayer extends Layer {
...
     /**
-     * @type {import('../style/LiteralStyle.js').LiteralStyle}
+     * @private
+     * @type {import('../webgl/ShaderBuilder.js').StyleParseResult}
      */
-    this.style = options.style;
-
-    assert(this.style.symbol !== undefined, 65);
+    this.parseResult_ = parseSymbolStyle(
+      /** @type {import('../style/LiteralStyle.js').LiteralStyle} */
```
fredj: This typecast is not necessary: typescript already knows the type of `options.style`. can you please change to:
```js
this.parseResult_ = parseSymbolStyle(options.style.symbol);
```

src/ol/layer/WebGLPoints.js
```diff
@@ -72,24 +71,24 @@ class WebGLPointsLayer extends Layer {
     super(baseOptions);
 
     /**
-     * @type {import('../style/LiteralStyle.js').LiteralStyle}
+     * @private
+     * @type {import('../webgl/ShaderBuilder.js').StyleParseResult}
      */
-    this.style = options.style;
-
-    assert(this.style.symbol !== undefined, 65);
```
fredj: @jahow this is not directly related to this pull request but the symbol property is marked as optional [in the typedef](https://github.com/openlayers/openlayers/blob/v6.0.1/src/ol/style/LiteralStyle.js#L9).

would it make sense to make it as mandatory?

jahow: Sure! it might change when/if we handle other kind of geometries. Thanks

jahow:
> Is this what you had in mind?
> Now the style is parsed in the ol.layer.WebGLPoints class' constructor.
> Previously it was only parse on the first rendering during an animation frame which is asynchronous.

Yes! Well done. Have you tried showing the exception message under the style text area if any, to give some more indication to the user? Thanks a lot

jahow:
Awesome, thanks a lot @MoonE!

PR #10086, opened 2019-10-03
Update mapbox-style example

petrsloup:
This PR updates the `mapbox-style` example to use the standard example layout (with visible code etc.) instead of the "verbatim" one and minor description fixes.


ahocevar:
What @tschaub says is correct. As long as we have no better way to measure performance, we rely on this example.

petrsloup:
Ok, I wasn't aware of that.

I "reverted" that part of the change, so this PR is now just a really minor docs update.

PR #10087, opened 2019-10-03
Add example showing vector tiles in EPSG:4326

petrsloup:
New example showing how to initialize TileGrid for EPSG:4326 vector tiles and style them using ol-mapbox-style.

The tiles are loaded from `api.maptiler.com` (using the same example-only key as the other examples).

### DIFF
--- a/examples/vector-tiles-4326.html
+++ b/examples/vector-tiles-4326.html
@@ -1,0 +1,13 @@
+---
+layout: example.html
+title: Vector tiles in EPSG:4326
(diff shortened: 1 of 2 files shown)


ahocevar:
@petrsloup Thanks for the new example! I added a commit which simplifies it. Please let me know if you agree with that change.

petrsloup:
I see you've taken a little different approach, but I don't have any issue with that :-) I agree with merging this. Thanks!

ahocevar:
Thanks @petrsloup!

PR #10093, opened 2019-10-04
Set label text for Graticule option styles

### LINKED ISSUE #10092: Graticule doesn't display label when using custom label style
**Describe the bug**
Graticule doesn't display label when using custom label style

**To Reproduce**

Use Graticule, notice that labels are shown. Add LabelStyle, refresh, and notice that labels don't show at all. E.g.
```
new Graticule({
    lonLabelStyle: new Text({
        font: '12px Calibri,sans-serif',
        textAlign: 'start',
        textBaseline: 'bottom',
        fill: new Fill({
            color: 'rgba(0,0,0,1)',
        }),
        stroke: new Stroke({
            color: 'rgba(255,255,255,1)',
            width: 3,
        }),
    }),
}),
```

I checked source and it it suspicious that `text: label` is not passed anywhere when custom style is used [here](https://github.com/openlayers/openlayers/blob/master/src/ol/layer/Graticule.js#L337).

**Expected behavior**
Either apply custom style, or remove config value.

mike-000:
Fixes #10092

Set label text for lonLabelStyle and latLabelStyle options

### DIFF
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -318,10 +318,23 @@ class Graticule extends VectorLayer {
         options.latLabelPosition;
 
       /**
-       * @type {Object.<string,Style>}
+       * @type {Style}
        * @private
        */
-      this.lonLabelStyleCache_ = {};
+      this.lonLabelStyleBase_ = new Style({
+        text: options.lonLabelStyle !== undefined ? options.lonLabelStyle.clone() :
+          new Text({
+            font: '12px Calibri,sans-serif',
+            textBaseline: 'bottom',
+            fill: new Fill({
+              color: 'rgba(0,0,0,1)'
+            }),
+            stroke: new Stroke({
+              color: 'rgba(255,255,255,1)',
+              width: 3
+            })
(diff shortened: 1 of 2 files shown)


src/ol/layer/Graticule.js
```diff
@@ -379,6 +378,7 @@ class Graticule extends VectorLayer {
                 })
               })
           });
+          this.latLabelStyleCache_[label].getText().setText(label);
```
ahocevar: With this addition, it is no longer necessary to cache or clone the style.

You could remove the `lonLabelStyleCache_` and `latLabelStyleCache_` members entirely, and create `lonLabelStyle_` and `latLabelStyle_` members instead, which either get the provided `options.lonLabelStyle` and `options.latLabelStyle` assigned, or the default that's created when these are not provided.

mike-000: Yes, I had the same thought but concentrated on the minimum change to fix the bug.  I have kept a single clone of each style we are updating to guard against the possibility of a user using the supplied definition elsewhere.

Issue #9950, opened 2019-09-20
Chromium 77 changes to getCanvasData breaks selection performance.


ablakey:
**Describe the bug**
Changes to getCanvasData (which drives `forEachFeatureAtPixel`) has reduced performance of OpenLayers selection by a few orders of magnitude.

**To Reproduce**
1. Open Chrome 77
2. Play with the Select Features example and monitor performance (https://openlayers.org/en/latest/examples/select-features.html)

**Expected behavior**
50-100x faster performance in my experience. My personal app now takes 800ms to select a feature.

It looks like the Chromium team might try to mitigate this, but they have also indicated it was known and intentional.  It's possible we need to re-work how OpenLayers does selection. Is there a way to limit how many calls to `getCanvasData` it has to make? I see it making dozens for a single selection. I imagine this is related to testing against many candidate features in the r-tree?

tschaub:
Thanks for the report @ablakey.  Support for some sort of `willReadFrequently` option when getting the context would be useful, but hard to say if that will be implemented.  We could look into using `addHitRegion` for our hit detection.  There are likely other ways to rework things to improve the performance.

ahocevar:
I just trried the https://openlayers.org/en/latest/examples/select-features.html example with Chrome Canary, and the issue seems to be resolved there.

ablakey:
@ahocevar thanks for checking on this! I don't have a non-Linux environment to install Canary on.  What difference in performance are you getting between the two? Are you inspecting performance in the developer tab and comparing the reported time for `getCanvasData` ?

ahocevar:
It thought it felt much faster yesterrday, but now with Chrome 79 (79.0.3923.0) compared to 77, the numbers don't necessarily reflect that.

Chrome 76:
<img width="628" alt="Untitled" src="https://user-images.githubusercontent.com/211514/65531182-96d0cc80-def9-11e9-8cc9-821632dfdd77.png">

Chrome 77:
<img width="628" alt="Untitled 6" src="https://user-images.githubusercontent.com/211514/65530584-8ec45d00-def8-11e9-8eaf-a5dc1db6b8bd.png">

Chrome 79:
<img width="628" alt="Untitled 7" src="https://user-images.githubusercontent.com/211514/65530613-9ab01f00-def8-11e9-9eb6-964a8a291040.png">

I'm investigating a different approach for hit detection right now, which would also simplify the code and reduce the memory footprint. I'll be reporting progress here.

[...]

PR #10094, opened 2019-10-06
New Layer#getFeatures method with fast hit detection for VectorLayer

ahocevar:
This pull request introduces an alternative to our current hit detection implementation.

See #9950.


ahocevar:
This is now ready for review.

ahocevar:
Anyone interested in reviewing this?

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -379,6 +394,69 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
...
+              if (sourceTileCoord.toString() === sourceTiles[j].tileCoord.toString()) {
+                sourceTile = sourceTiles[j];
+                break;
+              }
+            }
+          }
+          break;
+        }
+      }
+      const corner = getTopLeft(tileGrid.getTileCoordExtent(sourceTile.tileCoord));
```
jahow: I'm getting the following error on this line:
```
Uncaught (in promise) TypeError: Cannot read property 'tileCoord' of undefined
    at e.<anonymous> (VectorTileLayer.js:435)
    at new Promise (<anonymous>)
    at e.getFeatures (VectorTileLayer.js:401)
    at e.getFeatures (Layer.js:201)
    at e.getFeatures (BaseVector.js:153)
    at e.getFeatures (VectorTile.js:134)
    at e.<anonymous> (vector-tile-selection.js:48)
    at e.dispatchEvent (Target.js:110)
    at e.handleMapBrowserEvent (PluggableMap.js:946)
    at e.dispatchEvent (Target.js:110)
```

This happens randomly when clicking around the map in [the selection example](https://399-4974939-gh.circle-artifacts.com/0/examples/vector-tile-selection.html). Not very easy to reproduce, let me know if you need more precise instructions.

src/ol/render/canvas/Immediate.js
```diff
@@ -962,3 +973,139 @@ class CanvasImmediateRenderer extends VectorContext {
...
+export function createHitDetectionImageData(size, transforms, features, styleFunction, extent, resolution, rotation) {
+  const width = size[0] / 2;
+  const height = size[1] / 2;
+  const context = createCanvasContext2D(width, height);
+  context.imageSmoothingEnabled = false;
+  const canvas = context.canvas;
+  const renderer = new CanvasImmediateRenderer(context, 0.5, extent, null, rotation);
+  const featureCount = features.length;
+  // Stretch hit detection index to use the whole available color range
+  const indexFactor = Math.ceil((256 * 256 * 256) / featureCount);
```
jahow: Why is that necessary?

ahocevar: Because of the Canvas 2D anti-aliasing, we may get incorrect results in areas where the borders of rendered geometries overlap. By stretching the indexing to the available range, we improve the accuracy and increase the likelihood that we do not detect an incorrect features in such cases.

ahocevar:
Thanks for the thorough review @jahow. I have addressed your comments.

src/ol/renderer/canvas/VectorTileLayer.js
ahocevar: I was able to reproduce this - it was due to tiles that were not loaded yet, when we should actually resolve with an empty array.

PR #10101, opened 2019-10-08
Add hit detection support on the Heatmap layer

### LINKED ISSUE #10100: When a map have a heatmap layer, funciton hasFeatureAtPixel will throw a error
On demo https://openlayers.org/en/latest/examples/heatmap-earthquakes.html

Call function hasFeatureAtPixel, will throw a error

The Error message is "Uncaught AssertionError: Assertion failed. See https://openlayers.org/en/v6.0.1/doc/errors/#66 for details."

But I can't find a way to solve it

On 5.3.3 it's OK

jahow:
With this PR the Heatmap layer now supports hit detection again. Fixes #10100 

Note: hopefully soon enough we will be able to use the `ShaderBuilder` utilities to generate these shaders including for hit detection, instead of having to write mostly duplicate shader code.

### DIFF
--- a/src/ol/layer/Heatmap.js
+++ b/src/ol/layer/Heatmap.js
@@ -220,6 +220,51 @@ class Heatmap extends VectorLayer {
           float alpha = smoothstep(0.0, 1.0, value) * v_weight;
           gl_FragColor = vec4(alpha, alpha, alpha, alpha);
         }`,
+      hitVertexShader: `
+        precision mediump float;
+        uniform mat4 u_projectionMatrix;
+        uniform mat4 u_offsetScaleMatrix;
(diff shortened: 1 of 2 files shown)


ahocevar:
The tests added by this example sometimes fail. I could not get #10112 to pass, and I do not see how my changes there could be related to the failing tests.

jahow:
Okay, this is most likely due to the timeout here: https://github.com/openlayers/openlayers/pull/10101/files#diff-62e4a48ea375e4cf8e7da22eae166edeR70
This is intended to let the worker generate the buffers for rendering.
I'll see if I can get it to work without setTimeout.

PR #10103, opened 2019-10-09
Increase default attribution size

systemed:
The current default of 0.7rem is anomalously tiny and unlikely to fulfil typical licence requirements (e.g. OpenStreetMap's requirement that attribution is "reasonably calculated to make any Person that... views... the Produced Work"). See https://www.cyclinguk.org/route/great-north-trail-full-route-cape-wrath for an example of this attribution in the wild: unreadably small on Safari, Chromium and Firefox (macOS).

Increasing to 80% of the lineheight, i.e. 1.1rem, takes up no more vertical screen space but makes the attribution readable.

Note that the OL attribution example at https://openlayers.org/en/latest/examples/attributions.html overrides the default attribution to have a height of 1rem.

### DIFF
--- a/src/ol/ol.css
+++ b/src/ol/ol.css
@@ -182,7 +182,7 @@
 .ol-attribution ul {
   margin: 0;
   padding: 0 .5em;
-  font-size: .7rem;
+  font-size: 1.1rem;
   line-height: 1.375em;
   color: #000;
   text-shadow: 0 0 2px #fff;


simonpoole:
And just as miniscule on FF and Chrome on W10.

fredj:
@systemed what about this: https://github.com/openlayers/openlayers/compare/attribution_font_size ?
This removes all the text size for the attribution element.

systemed:
Looks good to me (though I'm not a CSS expert!).

fredj:
> Looks good to me (though I'm not a CSS expert!).

By removing the sizes in `ol.css`, the default font-size of the document will be used.
This is what https://www.cyclinguk.org/route/great-north-trail-full-route-cape-wrath will look like:

Screenshot from 2019-10-09 14-30-34

systemed:
Much better!

fredj:
fixed with #10104

PR #10108, opened 2019-10-10
Fix documentation for handleMoveEvent

### LINKED ISSUE #9814: PointerInteraction handleMoveEvent documentation bug
[`handleMoveEvent`](https://github.com/openlayers/openlayers/blob/v5.3.0/src/ol/interaction/Pointer.js#L143) function is not called when dragging. However, the [documentation](https://openlayers.org/en/latest/apidoc/module-ol_interaction_Pointer-PointerInteraction.html) says it is:
> Function handling "move" events. This function is called on "move" events, **also during a drag sequence (so during a drag sequence both the handleDragEvent function and this function are called)**.

Documentation should be corrected.

bobobobo:
This fixes #9814


tschaub:
Thanks @bobobobo.

PR #10118, opened 2019-10-11
Document the OSM crossOrigin default

mike-000:
This might help someone using an alternative url in ol/source/OSM understand why they get a CORS error


tschaub:
Thanks @mike-000.

PR #10119, opened 2019-10-12
Add crossOrigin option to ol/format/KML for icons

mike-000:
Add a crossOrigin option to ol/format/KML to make the crossOrigin setting for icons images configurable.

Fixes one of several FIXMEs in ol/format/KML and provides an simpler solution to situations such as https://stackoverflow.com/questions/58330075/kml-with-image-that-from-source-that-does-not-have-access-control-allow-origin-h

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -1121,6 +1121,7 @@ const STYLE_MAP_PARSERS = makeStructureNS(
 
 
 /**
+ * @this {KML}
  * @param {Element} node Node.
  * @param {Array<*>} objectStack Object stack.
  * @return {Array<Style>|string|undefined} StyleMap.
@@ -1146,6 +1147,7 @@ const ICON_STYLE_PARSERS = makeStructureNS(
 
 
 /**
+ * @this {KML}
  * @param {Element} node Node.
  * @param {Array<*>} objectStack Object stack.
  */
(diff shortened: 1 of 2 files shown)


mike-000:
This method working correctly is dependent on scope having been passed to readStyle either via .call or pushParseAndPop at each step.  Depending on the KML structure are at least three routes to reach readStyle and I initially missed some resulting in a "cannot read property crossOrigin of undefined" error, hence an icon style test for most of the permutations already in the tests.  I don't feel that passing scope like this is an ideal solution, it would be better if readStyle was a method of the class, hence #10400 is a reworking of the module where that error would be impossible, so if we were to go with #10400 instead I would be happy to remove from #10400 the multiple tests I copied from here.

jahow:
Ok, I agree with you on the scope issue. I'll review #10440 and maybe if it can be merged first this PR can be simplified. Thanks!

mike-000:
I could remove duplication of tests where it is only necessary to check the icon setup doesn't crash because of the changes.  Perhaps a failsafe in the code such as `crossOrigin: this ? this.crossOrigin_ ? default,` would better - and keep only the test which doesn't default?

jahow:
How about always specifying a `crossOrigin` value in the tests you added (read shared style from a folder, apply shared style to a feature, style maps) instead of having one test without value and another test with one? then we still make sure the scope is passed correctly in all cases.

Also adding `@this {KML}` in the JSDoc of `iconStyleParser` would make things more clear I think.

Thanks!

jahow:
Thanks!

PR #10120, opened 2019-10-12
Resolve constraints using anchor if following a cancelled animation

### LINKED ISSUE #10071: Continuous ZoomOnScroll looses the mouse position when using constrainResolution true

mike-000:
Fixes #10071

Resolve constraints using anchor if view state is the result of cancelling an animation with anchor

**This is an alternative to #10078**  
Unlike that PR this does not change the timing of the call, so a user clicking on an animating view will continue to see it freeze until a new interaction is started, when the resolution will use anchor of cancelled animation.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -312,6 +312,12 @@ class View extends BaseObject {
      */
     this.targetRotation_;
 
+    /**
+     * @private
+     * @type {import("./coordinate.js").Coordinate|undefined}
+     */
+    this.cancelAnchor_ = undefined;
+
     if (options.center) {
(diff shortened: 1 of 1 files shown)


jahow:
Ok, I'm merging this as it's been here for a while and no better solution was found. Thanks @mike-000

Issue #10158, opened 2019-10-21
Name collision of target_ property in Control and Target


ejn:
It looks like #10024 introduced a problem which leads to `ObjectEvents` dispatched on a `Control` with a custom target (parent element) having this parent element (`HTMLElement`) as their target instead of the `Control`.

The problem is caused by
 1. the `Target` class storing the target for the Event in the `target_` property, since #10024
 2. the `Control` class having always stored a custom parent element in the `target_` property
 3. `Control` inherits from `Target`. Calling `setTarget()` on a `Control` (or passing the `target` to the constructor) therefore also sets the `target` for Events generated by this `Control`, which I do not believe is the intended behaviour and is certainly a change since 5.3.x

To reproduce (e.g. with 6.0.1):
 1. Create a Control with a custom target, e.g. `new MousePosition({target: document.getElementById("pos")});`
 2. Add a `propertychange`-listener to this `Control`
 3. Inspect the `target`-property of the `ObjectEvent` received by the listener: This is the `HTMLElement` and not the `MousePosition`.

I guess the (private) `target_` property either in `Target` (introduced in #10024) or in `Control` should be renamed. Since the property in `Target` is newer, renaming this (e.g. to `eventTarget_`) would seem preferable.

ahocevar:
Good find, thanks for reporting this. Would you be able to provide a pull request with your suggested fix?

ejn:
Yes, I can provide a PR - just wanted to check that the suggested solution would be acceptable first.

fredj:
fixed with #10159, thanks @ejn

PR #10159, opened 2019-10-22
Fix target name collision in event.Target and Control

ejn:
Fix for Issue #10158

### DIFF
--- a/test/spec/ol/control/control.test.js
+++ b/test/spec/ol/control/control.test.js
@@ -54,3 +54,25 @@ describe('ol.control.Control\'s target', function() {
     });
   });
 });
+
+describe('ol.control.Control\'s event target', function() {
+  it('is the Control when the Control uses the default target', function() {
+    const ctrl = new Control({element: document.createElement('div')});
+    ctrl.on('test-event', function(e) {
+      expect(e.target).to.be(ctrl);
+    });
+    ctrl.dispatchEvent('test-event');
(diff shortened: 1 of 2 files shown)


test/spec/ol/control/control.test.js
```diff
@@ -54,3 +54,25 @@ describe('ol.control.Control\'s target', function() {
...
+  });
+  it('is the Control when the Control has a custom target', function() {
+    const ctrl = new Control({
+      element: document.createElement('div'),
+      target: document.createElement('div')
+    });
+    ctrl.on('test-event', function(e) {
+      expect(e.target).to.be(ctrl);
+    });
+    ctrl.dispatchEvent('test-event');
```
tschaub: Thanks for adding these tests.

As written, this would pass if the event listener were never called.  And if for some reason the listener were called asynchronously, the assertion would be made after the test was complete.  To more reliably make the same assertion, you could declare a `target` before registering the listener, assign to it in the listener, and then make an assertion about the target after dispatching the event.

tschaub:
Looks like I wasn’t quick enough with the review :)

test/spec/ol/control/control.test.js
ejn: I'd briefly considered that, and decided that it was very unlikely that they'd get called asynchronously. Nevertheless: #10160

PR #10160, opened 2019-10-22
Improve test in case of async call

ejn:
As suggested by @tschaub at https://github.com/openlayers/openlayers/pull/10159#discussion_r337371216, take account of the fact that the listener callbacks could (currently completely theoretically) be run asynchrously.

Based on pattern in eventtarget.test.js


tschaub:
Thanks @ejn.

PR #10168, opened 2019-10-22
Webgl / Add support for filtering and color interpolation in style expressions

jahow:
This PR significantly reworks the `ol/style/LiteralStyle` and `ol/webgl/ShaderBuilder` modules to add the following:
* a `filter` expression can be specified in a literal style in order to exclude features from the render
* several logical operators have been added to support this: `>`, `>=`, `<`, `<=`, `between`
* the literal styles now handle different types instead of only numbers; for now this is only used for the `color` style property; all types will be checked when the style is parsed and exceptions will be thrown if there is a mismatch
* a new `interpolate` operator has been added, which takes in two color values and a numerical ratio
* several new math operators have been added: `/`, `-`, `mod`, `pow`
* a new `time` operator was added, which simply returns the time since layer creation in seconds
* the literal styles can also receive a `variables` object; each key in the object is a variable that can be referenced in the style expressions using the `var` operator; **this object is meant to be mutated** and all changes will be reflected in real time on the map

See the [reworked filter-points-webgl example](https://2445-4723248-gh.circle-artifacts.com/0/examples/filter-points-webgl.html) for a showcase of the added features.

Also the `ol/webgl/ShaderBuilder` module now has a `ShaderBuilder` class which actually implements a builder pattern and is used to generate shaders, hopefully making things easier to read & understand...

### DIFF
--- a/src/ol/webgl/ShaderBuilder.js
+++ b/src/ol/webgl/ShaderBuilder.js
@@ -18,23 +18,32 @@ export function formatNumber(v) {
 /**
  * Will return the number array as a float with a dot separator, concatenated with ', '.
  * @param {Array<number>} array Numerical values array.
- * @returns {string} The array as string, e. g.: `1.0, 2.0, 3.0`.
+ * @returns {string} The array as a vector, e. g.: `vec3(1.0, 2.0, 3.0)`.
  */
 export function formatArray(array) {
-  return array.map(formatNumber).join(', ');
+  if (array.length < 2 || array.length > 4) {
+    throw new Error('`formatArray` can only output `vec2`, `vec3` or `vec4` arrays.');
+  }
+  return `vec${array.length}(${array.map(formatNumber).join(', ')})`;
 }
 
 /**
- * Will normalize and converts to string a color array compatible with GLSL.
+ * Will normalize and converts to string a `vec4` color array compatible with GLSL.
  * @param {string|import("../color.js").Color} color Color either in string format or [r, g, b, a] array format,
- * with RGB components in the 0..255 range and the alpha component in the 0..1 range. Note that if the A component is
- * missing, only 3 values will be output.
- * @returns {string} The color components concatenated in `1.0, 1.0, 1.0, 1.0` form.
+ * with RGB components in the 0..255 range and the alpha component in the 0..1 range.
+ * Note that the final array will always have 4 components.
+ * @returns {string} The color expressed in the `vec4(1.0, 1.0, 1.0, 1.0)` form.
  */
 export function formatColor(color) {
-  return asArray(color).map(function(c, i) {
-    return i < 3 ? c / 255 : c;
-  }).map(formatNumber).join(', ');
+  const array = asArray(color).slice();
+  if (array.length < 4) {
+    array.push(1);
+  }
+  return formatArray(
+    array.map(function(c, i) {
+      return i < 3 ? c / 255 : c;
+    })
+  );
 }
 
 /**
--- a/src/ol/webgl/ShaderBuilder.js
+++ b/src/ol/webgl/ShaderBuilder.js
@@ -1,5 +1,5 @@
 /**
- * Utilities for generating shaders from literal style objects
+ * Classes and utilities for generating shaders from literal style objects
  * @module ol/webgl/ShaderBuilder
  */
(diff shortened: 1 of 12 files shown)


jahow:

src/ol/webgl/ShaderBuilder.js
```diff
@@ -18,237 +18,669 @@ export function formatNumber(v) {
...
+        checkNumber(v[1]);
+        break;
+      case 'between':
+        checkNumber(v[1]);
+        checkNumber(v[2]);
+        checkNumber(v[3]);
+        break;
+
+      case 'interpolate':
+        checkNumber(v[1]);
```
tschaub: I think we should use more context when checking/coercing types.  For example, when we have `color: ['interpolate', ...args]`, then we would pass along the expectation that values are color type, and when we have `size: ['interpolate', ...args]`, then we would pass along the expectation that values are numeric (I also think that we could collapse `stretch` and `interpolate` into a single `interpolate` expression a la https://docs.mapbox.com/mapbox-gl-js/style-spec/#expressions-interpolate).

src/ol/webgl/ShaderBuilder.js
```diff
@@ -18,237 +18,669 @@ export function formatNumber(v) {
...
+ * @returns {string} Assignment string.
  */
-export function getSymbolVertexShader(parameters) {
-  const offsetMatrix = parameters.rotateWithView ?
-    'u_offsetScaleMatrix * u_offsetRotateMatrix' :
-    'u_offsetScaleMatrix';
+export function parse(value, attributes, attributePrefix, variables, typeHint) {
+  check(value);
+
+  const v = value;
```
tschaub: Why?

src/ol/webgl/ShaderBuilder.js
```diff
@@ -18,237 +18,669 @@ export function formatNumber(v) {
...
+  function p(value) {
+    return parse(value, attributes, attributePrefix, variables);
+  }
+  function pC(value) {
+    return parse(value, attributes, attributePrefix, variables, ValueTypes.COLOR);
+  }
+
+  // operator
+  if (Array.isArray(v) && typeof value[0] === 'string') {
```
tschaub: `v` or `value`?

tschaub:
Really cool to see this getting fleshed out @jahow.  Consider the comments above non-blocking.

src/ol/webgl/ShaderBuilder.js
jahow: I made the assumption that each operator would only expect one type of value, in order to simplify code. Looking at the `interpolate` operator, I realized it would not really make sense to interpolate *numbers* anyway (and strings neither), so I went for colors only.

But looking again at the `interpolate` operator of Mapbox style, I guess it would indeed be a good match! If that's alright, I'll work towards this in another PR. Thanks

src/ol/webgl/ShaderBuilder.js
jahow: To keep the lines shorter!

jahow:
Thanks a lot for the swift review @tschaub! The next step will be adding support for string attributes and migrating the `icon-sprite-webgl` example to use a literal style as well. Still not sure how to do that though...

src/ol/webgl/ShaderBuilder.js
tschaub: I also like the idea of having typed operators, but working with colors already suggests a variadic operator (if we accept a string or rgb(a) array).  Choosing the type of interpolation (linear, exponential) can also be useful.  For example, scaling a circle radius with zoom level:

```js
['interpolate', ['exponential', 2], ['zoom'], 0, 1, 18, 12500]
```

(This could also be `linear` with `['resolution']`, but I think people have an easier time working with zoom compared to resolution.)

Anyway, I agree it makes sense to work on this separately.

jahow: Definitely agree. `interpolate` can become a cornerstone of style expressions if done correctly (pretty much like Mapbox did it IMO). I think we should organize the code a bit differently to allow for more logic in each separate operator, e.g. to support different input/output kinds, variable argument count, etc. This would allow us to have more powerful operators, and hopefully less of them!

Issue #10166, opened 2019-10-22
Add condition to extent interaction


Chris1234567899:
The extent interaction should allow a condition function as argument like other interactions do (e.g. draw, modify, move etc).

Would be good for consistency if this kind of interactions work similar.

fredj:
@Chris1234567899  Would you be willing to create a pull request?

PR #10174, opened 2019-10-23
adding condition to extent interaction

Chris1234567899:
This adds an optional condition function as an argument to the extent interaction. Similar to the other interactions.
See issue: 
https://github.com/openlayers/openlayers/issues/10166

### DIFF
--- a/src/ol/interaction/Extent.js
+++ b/src/ol/interaction/Extent.js
@@ -230,8 +230,8 @@ class Extent extends PointerInteraction {
    */
   handlePointerMove_(mapBrowserEvent) {
     if (!this.condition_(mapBrowserEvent)) {
-      this.removePointFeature_()
-      return
+      this.removePointFeature_();
+      return;
     }
     const pixel = mapBrowserEvent.pixel;
     const map = mapBrowserEvent.map;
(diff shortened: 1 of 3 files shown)


src/ol/interaction/Extent.js
```diff
@@ -293,6 +318,9 @@ class Extent extends PointerInteraction {
    * @inheritDoc
    */
   handleDownEvent(mapBrowserEvent) {
+    if (!this.condition_(mapBrowserEvent)) {
+      return;
```
ahocevar: `return false`  here to indicate that the event has not been consumed by this interaction.

src/ol/interaction/Extent.js
```diff
@@ -218,6 +229,10 @@ class Extent extends PointerInteraction {
    * @private
    */
   handlePointerMove_(mapBrowserEvent) {
+    if (!this.condition_(mapBrowserEvent)) {
+      this.removePointFeature_();
+      return;
```
ahocevar: `return false` here to indicate that the event has not been consumed by this interaction.

src/ol/interaction/Extent.js
```diff
@@ -272,6 +287,16 @@ class Extent extends PointerInteraction {
     return vertexFeature;
```
ahocevar: You do not need the `removePointFeature_` function when you work on the return value here: `return mapBrowserEvent.dragging` might do the trick, but I have not tested that.

PR #10177, opened 2019-10-23
Experimental KML IconStyle test

mike-000:
This adds an IconStyle test from #10119 to determine if there is an existing issue with either ol/format/KML or the test setup.  It is unlikely to need merging.


mike-000:
Closing this as there is clearly an existing issue unrelated to #10177 where something in a preceding test is preventing a "clean" IconStyle being read in a subsequent test..

PR #10178, opened 2019-10-23
Make ol/interaction/Select#handleEvent return always true.

### LINKED ISSUE #9945: why multiple select interaction does not work?
I have multiple layers in my openlayers application. So I need layer based specific select interaction. Each layer will have its own select interaction. This is my problem. I do not want to add layers in one select interaction, because it is not my solution.

[Working application is here.](https://jsfiddle.net/veung7hc/)

TieSKey:
Fix #9945 https://github.com/openlayers/openlayers/issues/9945 

Make ol/interaction/Select#handleEvent return always true regardless 
of event type.
Previously it was returning true only if the event was a pointerMove
which prevented any but the last Select objects with other event types 
from working.

### DIFF
--- a/examples/select-features.js
+++ b/examples/select-features.js
@@ -1,6 +1,6 @@
 import Map from '../src/ol/Map.js';
 import View from '../src/ol/View.js';
-import {click, altKeyOnly} from '../src/ol/events/condition.js';
+import {click, pointerMove, altKeyOnly} from '../src/ol/events/condition.js';
 import GeoJSON from '../src/ol/format/GeoJSON.js';
 import Select from '../src/ol/interaction/Select.js';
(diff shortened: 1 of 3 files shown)


TieSKey:
About the pointerMove, automated tools/test complaint it becoming an unused variable. I'm not familiar with the toolchain, is there a way to avoid this check?

ahocevar:
@TieSKey You probably misunderstood the warning. It was only about the import of `pointerMove` in `ol/interaction/Select`, not about the export.

ahocevar:
Thanks!

TieSKey:
Thank you for your time reviewing and helping me with this minuscule (but necessary) change.

PR #10187, opened 2019-10-24
Set touch-action to allow native touch gestures

ahocevar:
Native page scrolling on touch devices only works when `touch-action: none` is not set. So we need to put some effort into not setting that when the user might want page scrolling (which we expect to be the case when a `tabindex` attribute is set on the map target element).

Note that this change could have been done entirely in css if we'd rely on always having a stylesheet, and not need the PEP polyfill. To make it always work, I decided to implement this change in JavaScript.

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -61,7 +61,6 @@ class MapBrowserEventHandler extends EventTarget {
     this.down_ = null;
 
     const element = this.map_.getViewport();
-    element.setAttribute('touch-action', 'none');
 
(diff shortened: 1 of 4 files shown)


ahocevar:
To provide some context: this pull request fixes the broken interaction-options example on touch devices. Page scrolling is broken here:

https://openlayers.org/en/v6.0.1/examples/interaction-options.html

and fixed here:

https://418-4974939-gh.circle-artifacts.com/0/examples/interaction-options.html

ahocevar:
Thanks in advance for any review.

PR #10188, opened 2019-10-24
Document that map render events do not have a context set

### LINKED ISSUE #10186: rendercomplete event
Map event rendercomplete returns context undefined.

ahocevar:
Fixes #10186.

### DIFF
--- a/src/ol/render/Event.js
+++ b/src/ol/render/Event.js
@@ -33,8 +33,8 @@ class RenderEvent extends Event {
     this.frameState = opt_frameState;
 
     /**
-     * Canvas context. Only available when a Canvas renderer is used, null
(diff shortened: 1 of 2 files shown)


src/ol/render/Event.js
```diff
@@ -33,8 +33,8 @@ class RenderEvent extends Event {
     this.frameState = opt_frameState;
 
     /**
-     * Canvas context. Only available when a Canvas renderer is used, null
-     * otherwise.
+     * Canvas context. Not available when the event is dispatched by the map. Only available
+     * when a Canvas renderer is used, null otherwise.
```
fredj: The comment was already false; it should be _undefined otherwise_

Issue #10193, opened 2019-10-25
Bbox filter should throw error if extent is wrong.


MrSoUndso:
The bbox filter should throw an error if the extent is wrong. e.g. an array with less than four elements.

**To Reproduce**
Create a new bbox filter with an extent like this: [0,0,100]

**Expected behavior**
This should throw an error, but fails silently.

PR #10194, opened 2019-10-25
Added check to make sure extent array has four objects.

MrSoUndso:

### DIFF
--- a/src/ol/format/filter/Bbox.js
+++ b/src/ol/format/filter/Bbox.js
@@ -30,7 +30,11 @@ class Bbox extends Filter {
     /**
      * @type {import("../../extent.js").Extent}
      */
-    this.extent = extent;
+      if(extent.length === 4) {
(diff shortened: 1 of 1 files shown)


MrSoUndso:
This pull request should close issue #10193

src/ol/format/filter/Bbox.js
```diff
@@ -30,7 +30,11 @@ class Bbox extends Filter {
     /**
      * @type {import("../../extent.js").Extent}
      */
-    this.extent = extent;
```
tschaub: So that the type annotation applies to this property, you should leave this line here.

You can throw in a condition below (without needing an `else`):

```js
if (extent.length !== 4) {
    throw new Error('Expected an extent with four values ([minX, minY, maxX, maxY])';
}
```

tschaub:
Thanks @MrSoUndso

PR #10196, opened 2019-10-25
Webgl / add support for string and arrays in style expressions

jahow:
This PR is a followup of #10168 and again brings a lot of changes to the style expressions system:
* A new `ol/style/expressions` module has been created, which holds type definitions as well as type checking and conversion to GLSL logic for style expressions; I contemplated putting this in `ol/webgl/*` but eventually went with this as it holds all the expressions logic regardless of output

* Value types in style expressions are now handled by binary flags: `ValueTypes.NUMBER`, `ValueTypes.BOOLEAN`, `ValueTypes.COLOR` and `ValueTypes.STRING`; this means an operator can have a return type of `ValueTypes.NUMBER | ValueTypes.NUMBER_ARRAY`

* Style expression operators now all have their own declaration, composed of two methods:
  * `getReturnType` returns one or more value types and is used to recursively check types in style expressions
  * `toGlsl` takes in the parsing context and operator arguments, and outputs a GLSL-compatible string (by recursively parsing its arguments)
  Each operator can now have a different logic, take in variable amount of arguments, etc.
  The parsing context is used to store variables and attributes required by the expressions.

* Several new operators are added:
  * `match` compares an input to a list of matches and returns the associated output, with a fallback value (just like the [corresponding Mapbox style operator](https://docs.mapbox.com/mapbox-gl-js/style-spec/#expressions-match))
  * `interpolate` has been extended to work like the [corresponding Mapbox style operator](https://docs.mapbox.com/mapbox-gl-js/style-spec/#expressions-interpolate), and now allows interpolating through several stops with either a linear or exponential rate. Also handles numbers in addition to colors.
    Consequently, the `stretch` operator has been removed as it was essentially doing the same thing.
    Note: the implementation here is much simpler than in Mapbox-gl-js
  * `color` and `array` both take in number values and output their respective types
  * `zoom` and `resolution` simply return the current zoom and resolution values
  * `pow` has been renamed to `^` (again, matching the [Mapbox style spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/#expressions-^))
  * the `get` operator can now handle string attributes; these get mapped to number values at parse time, so that they can be used in the vertex/fragment shaders; the string/number map is stored in the parsing context.

* The `ShaderBuilder` class is now able to output shaders fit for hit detection

* The `WebGLPointsLayer` class now takes an optional `disableHitDetection` flag which can save some performance when hit detection is not needed (hit detection is enabled by default).

As a result, the [icon-sprite-webgl](https://2485-4723248-gh.circle-artifacts.com/0/examples/icon-sprite-webgl.html) example was rewritten to use a literal style and not handcrafted shaders. Also the [webgl-points-layer](https://2485-4723248-gh.circle-artifacts.com/0/examples/webgl-points-layer.html) has been expanded with new sample styles! 

Thanks a lot for any review. With this in, I think we have a solid foundation for using style expressions.

### DIFF
--- a/examples/icon-sprite-webgl.js
+++ b/examples/icon-sprite-webgl.js
@@ -4,11 +4,9 @@ import TileLayer from '../src/ol/layer/Tile.js';
 import TileJSON from '../src/ol/source/TileJSON.js';
 import Feature from '../src/ol/Feature.js';
 import Point from '../src/ol/geom/Point.js';
-import VectorLayer from '../src/ol/layer/Vector.js';
 import {Vector} from '../src/ol/source.js';
 import {fromLonLat} from '../src/ol/proj.js';
-import WebGLPointsLayerRenderer from '../src/ol/renderer/webgl/PointsLayer.js';
-import {colorToGlsl, numberToGlsl} from '../src/ol/style/expressions.js';
+import WebGLPointsLayer from '../src/ol/layer/WebGLPoints.js';
 
 const key = 'pk.eyJ1IjoidHNjaGF1YiIsImEiOiJjaW5zYW5lNHkxMTNmdWttM3JyOHZtMmNtIn0.CDIBD8H-G2Gf-cPkIuWtRg';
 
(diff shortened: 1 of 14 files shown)


examples/icon-sprite-webgl.js
```diff
@@ -17,161 +15,41 @@ const vectorSource = new Vector({
...
+      1950, oldColor,
+      2013, newColor
+    ],
+    rotateWithView: false,
+    offset: [
+      0,
+      9
+    ],
+    textureCoord: [
+      'match',
```
fredj: very nice !

jahow: I know! :) this really turned out well.

src/ol/style/expressions.js
```diff
@@ -0,0 +1,612 @@
...
+    return 'u_resolution';
+  }
+};
+Operators['*'] = {
+  getReturnType: function(args) {
+    return ValueTypes.NUMBER;
+  },
+  toGlsl: function(context, args) {
+    assertArgsCount(args, 2);
+    assertNumber(args[0]);
```
fredj: See https://github.com/fredj/openlayers/commit/94127ca28adf9fff33e2574c3e4bef6a2872aa4d for a small simplification: I've added a `assertNumbers` function that takes an array and check that every item is a number.

jahow: Well done, I'll add this commit to the PR

PR #10198, opened 2019-10-25
WebGL renderer / use the specified loading strategy for the vector data

### LINKED ISSUE #10173: Heatmap issue with GeoJSON and loadingstrategy as tile
Hi 

I am trying to create a Heatmap with a GeoJSON source and loadingstrategy.tile. I am running the code below and openlayers is not displaying anything or crashes the browser.

var source = new ol.source.Vector({
                            format: new ol.format.GeoJSON({
                                dataProjection: projectionCode,
                                geometryName: geometryName
                            }),
                            loader: function(...),
                            strategy: ol.loadingstrategy.tile(new ol.tilegrid.createXYZ({extent: getExtent() })),
                            projection: projectionCode
                        });

var vector = new ol.layer.Heatmap({
              source: source,
              blur: 25,
              radius: 15
            });

vector.getSource().on('addfeature', function(event) {
              event.feature.set('weight', 1);
});

I managed to get round the problem by using loadingstrategy.tile with a ol.layer.Vector and using loader function to fill Heatmap source.

jahow:
Previously the WebGL points renderer would specify an infinite extent when loading features, bypassing the loading strategy and potentially crashing the browser. It now specifies the current view extent to `vectorSource.loadFeatures`.

Fixes #10173 

@mike-000 could you please take a look? thanks!

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -7,7 +7,7 @@ import {AttributeType, DefaultUniform} from '../../webgl/Helper.js';
 import GeometryType from '../../geom/GeometryType.js';
 import WebGLLayerRenderer, {colorDecodeId, colorEncodeId, WebGLWorkerMessageType} from './Layer.js';
 import ViewHint from '../../ViewHint.js';
-import {createEmpty, equals} from '../../extent.js';
+import {buffer, createEmpty, equals} from '../../extent.js';
 import {
   apply as applyTransform,
   create as createTransform,
@@ -18,6 +18,7 @@ import {create as createWebGLWorker} from '../../worker/webgl.js';
 import {getUid} from '../../util.js';
 import WebGLRenderTarget from '../../webgl/RenderTarget.js';
 import {assert} from '../../asserts.js';
+import BaseVector from '../../layer/BaseVector.js';
 
 /**
  * @typedef {Object} CustomAttribute A description of a custom attribute to be passed on to the GPU, with a value different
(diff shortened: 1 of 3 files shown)


jahow:
I have updated the PR to fix the behaviour of the renderer. Now the features loading is called when:
* either the source revision has changed, or the current extent
* the view is not moving

Previously the source was only reloaded when its revision was changed. New tests were added to reflect that. Thanks for any review!

mike-000:
With the latest change it looks good (the calls to the loader are matching those of the normal vector layer).

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -296,20 +297,22 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
+    if (viewNotMoving && (extentChanged || sourceChanged)) {
       const projection = viewState.projection;
       const resolution = viewState.resolution;
-      vectorSource.loadFeatures([-Infinity, -Infinity, Infinity, Infinity], resolution, projection);
-    }
 
-    const viewNotMoving = !frameState.viewHints[ViewHint.ANIMATING] && !frameState.viewHints[ViewHint.INTERACTING];
-    const extentChanged = !equals(this.previousExtent_, frameState.extent);
-    if ((sourceChanged || extentChanged) && viewNotMoving) {
+      const renderBuffer = layer instanceof BaseVector ? layer.getRenderBuffer() : 0;
```
fredj: did you use `instanceof` because of an error from the type checker?

jahow: Actually the renderer can use layers that do not inherit from the `BaseVector` class, as seen here: https://github.com/openlayers/openlayers/blob/master/src/ol/layer/WebGLPoints.js

The only requirement is having a vector source.

fredj: ok, got it now.

jahow:
Thanks to both for the reviews!

PR #10214, opened 2019-10-28
WebGL / Add 'case' operator for style expressions

jahow:
This PR adds the `case` operator, allowing if/else control flow in expressions. The implementation should be compliant with [the Mapbox style spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/#expressions-case).

Also added the `!=` operator and renamed `mod` to `%` to be more in line with the afore-mentioned spec.

Finally, a few fixes were made to the existing operators and the `ShaderBuilder` class to allow filtering on string attributes. To showcase this, the icon-sprite-webgl example was modified to allow filtering UFO sightings based on shape.

### DIFF
--- a/examples/filter-points-webgl.js
+++ b/examples/filter-points-webgl.js
@@ -18,7 +18,7 @@ const period = 12; // animation period in seconds
 const animRatio =
   ['^',
     ['/',
-      ['mod',
+      ['%',
         ['+',
           ['time'],
           [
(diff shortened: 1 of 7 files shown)


jahow:
I'd be keen to hear @tschaub's opinion on where this style expressions system is headed, thanks!

tschaub:
Looks great @jahow.

ahocevar:
Merging, so we have this in the 6.1.0 release. Thanks @jahow!

PR #10216, opened 2019-10-28
Update changelog script to recognize dependabot instead of greenkeeper

ahocevar:
Should be merged after the 6.1.0 release.


haxgibsons:
what is dependabot?

PR #10224, opened 2019-10-29
Vector tile hit detection fixes

ahocevar:
This pull request fixes two issues with vector tile hit detection:

* The source tiles of a loaded vector tile can have an ERROR state, so we need to check the state of a source tile before using it.
* The scaling of the tile hit pixel and the resolution for the render transform was incorrect for overzoomed vector tiles.

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -437,10 +437,12 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
         resolve([]);
         return;
       }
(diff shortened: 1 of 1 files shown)


tschaub:
Seems like tiles in an error state have been coming up a lot.  I know it isn't related to this change, but I'm curious when this happens.

ahocevar:
> Seems like tiles in an error state have been coming up a lot. I know it isn't related to this change, but I'm curious when this happens.

This usually comes from vector tile sets that do not cover the whole world or a specified extent, i.e. due to 404 errors. I use tippecanoe a lot for creating tile sets for a specific region, and that only creates tiles that have data.

PR #10226, opened 2019-10-30
ABORT reproj tiles properly

### LINKED ISSUE #10199: White Tiles on bing maps
**Describe the bug**
When using Bing maps as a basemap as with following code:
`       var basemap = new TileLayer({
            visible: true,

            source: new BingMaps({
                imagerySet: 'Road',
                key: 'code',
                wrapX: true
            })
        });
`
I get the following white space when it is rendered:
image

If I zoom in further the white space gets filled in.

**To Reproduce**
1. use bing for basemap
2. Zoom in at least 3 levels in the Caribbean region
3. See image above.

**Expected behavior**
All tiles rendered correctly or at least provide zoom levels for which the Bing maps do work correctly. However, I did not see this issue before using Openlayers 6.01

Update: The error is not Bing maps, but rather the vector layer on top of it.
image

ahocevar:
When the tile cache gets ride of a tile, its state needs to be set to ABORT. This is missing in `reproj/Tile`. But instead of adding it there, I restructured things a bit to set the ABORT state in the base class instead, to make sure every tile class has this behavior.

In version 5, this tile life cycle issue (along with others that we fixed already) usually did not reveal itself, because we had an incredibly big tile cache default size there. In v6, with smaller tile caches, these issues became more apparent.

Fixes #10199.

### DIFF
--- a/src/ol/ImageTile.js
+++ b/src/ol/ImageTile.js
@@ -69,8 +69,6 @@ class ImageTile extends Tile {
     if (this.interimTile) {
       this.interimTile.dispose();
     }
-    this.state = TileState.ABORT;
-    this.changed();
     super.disposeInternal();
   }
 
--- a/src/ol/Tile.js
+++ b/src/ol/Tile.js
@@ -150,6 +150,13 @@ class Tile extends EventTarget {
   }
 
   /**
+   * @inheritDoc
+   */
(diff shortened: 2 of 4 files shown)


bearman987:
When a new release can be expected?

PR #10229, opened 2019-10-30
Update Draw.js

neylsongularte:
fix remove last point in draw interaction


jahow:
Hi,

Thanks for contributing. I'm sorry but this will need a bit more context on what you're trying to achieve with this change!

neylsongularte:
Hi, removeLastPoint function is not working properly, next to last item in sketchLineCoords_ array but last not updated

neylsongularte:
the last two coordinates are repeated

neylsongularte:
I believe this only happens on mobile

jahow:
Sorry, that's still not clear enough. Please refer to [the contributing guide](https://github.com/openlayers/openlayers/blob/master/CONTRIBUTING.md) as to what your PR shoud look like. Thanks!

ahocevar:
Fixed with #11698

PR #10235, opened 2019-10-31
Various optimizations and fixes for the WebGL points renderer

jahow:
This PR addresses several points:
* The `rebuildBuffers` operation of the WebGL points renderer can be CPU-intensive and has to be done on the main thread; this PR reduces the time it takes by around 50-60%, by minimizing garbage collection and maintaining a cache of geometries to render instead of calling `source.getFeatures()`, `feature.getGeometry()` etc.
  On the `icon-sprite-webgl` example it should hopefully be down to 30-60ms on most machines, which makes the stutter less noticeable than before
* When a `WebGLPointsLayer` object is removed and `dispose()` is called, the underlying renderer and WebGL helper will be disposed as well, and the associated WebGL context will be garbage collected; this was not the case before as some references were kept (which meant on the `webgl-points-layer` example an error would pop up in the console after modifying the style more than ~15 times)
* Fixes an issue where hit detection on a WebGL layer would bring up false positives
* The canvas used to render hit detection data is now half the width/height of the main canvas, which still gives sufficient precision while using up less memory

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -19,7 +19,7 @@ import {getUid} from '../../util.js';
 import WebGLRenderTarget from '../../webgl/RenderTarget.js';
 import {assert} from '../../asserts.js';
 import BaseVector from '../../layer/BaseVector.js';
-import {listen} from '../../events.js';
+import {listen, unlistenByKey} from '../../events.js';
 import VectorEventType from '../../source/VectorEventType.js';
 
 /**
@@ -288,9 +288,11 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
     this.featureCache_ = {};
 
     const source = this.getLayer().getSource();
-    listen(source, VectorEventType.ADDFEATURE, this.handleSourceFeatureChanged_, this);
-    listen(source, VectorEventType.CHANGEFEATURE, this.handleSourceFeatureChanged_, this);
-    listen(source, VectorEventType.REMOVEFEATURE, this.handleSourceFeatureDelete_, this);
+    this.sourceListenKeys_ = [
+      listen(source, VectorEventType.ADDFEATURE, this.handleSourceFeatureChanged_, this),
+      listen(source, VectorEventType.CHANGEFEATURE, this.handleSourceFeatureChanged_, this),
+      listen(source, VectorEventType.REMOVEFEATURE, this.handleSourceFeatureDelete_, this)
+    ];
     source.getFeatures().forEach(function(feature) {
       const uid = getUid(feature);
       this.featureCache_[uid] = {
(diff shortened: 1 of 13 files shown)


jahow:

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -269,6 +279,52 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
+     */
+    this.featureCache_ = {};
+
+    const source = this.getLayer().getSource();
+    this.sourceListenKeys_ = [
+      listen(source, VectorEventType.ADDFEATURE, this.handleSourceFeatureChanged_, this),
+      listen(source, VectorEventType.CHANGEFEATURE, this.handleSourceFeatureChanged_, this),
+      listen(source, VectorEventType.REMOVEFEATURE, this.handleSourceFeatureDelete_, this)
+    ];
+    source.getFeatures().forEach(function(feature) {
```
tschaub: Seems like you could call `source.forEachFeature()` here.

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -343,45 +399,44 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
   rebuildBuffers_(frameState) {
-    const layer = this.getLayer();
-    const vectorSource = layer.getSource();
-
     // saves the projection transform for the current frame state
     const projectionTransform = createTransform();
     this.helper.makeProjectionTransform(frameState, projectionTransform);
 
-    const features = vectorSource.getFeatures();
+    const featureUids = Object.keys(this.featureCache_);
```
tschaub: Doesn't need to happen here, but since you only want the count of features, we could avoid the array allocation and cache the count when you cache the features.

src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -343,45 +399,44 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
+    let featureUid, featureCache, geometry;
     const tmpCoords = [];
     const tmpColor = [];
     let renderIndex = 0;
     let hitIndex = 0;
     let hitColor;
-    for (let i = 0; i < features.length; i++) {
-      feature = features[i];
-      if (!feature.getGeometry() || feature.getGeometry().getType() !== GeometryType.POINT) {
+    for (let i = 0; i < featureUids.length; i++) {
```
tschaub: And if you did https://github.com/openlayers/openlayers/pull/10235/files#r341259256, then this would be `for (const key in this.featuresCache_)`.

src/ol/renderer/webgl/PointsLayer.js
jahow: Done in 8aa606170975062dd733e75f19965978bc7e2cbf

src/ol/renderer/webgl/PointsLayer.js
jahow: Good find, done in 8aa606170975062dd733e75f19965978bc7e2cbf

src/ol/renderer/webgl/PointsLayer.js
jahow: Done in 7da86ae71fdedc41c334c7f882fb458f8ab49e08

jahow:
Thanks for the suggestions @tschaub.

Issue #10096, opened 2019-10-07
Features in decluttered layer are above other layer, ignoring zIndex


M393:
**Describe the bug**
Feature in decluttered layer is above feature of other layer with higher zIndex.

**To Reproduce**
https://codesandbox.io/s/priceless-hugle-3fxkb
I set both the layer zIndex and the feature style zIndex to the same value for each layer.
for the layer with decluttering zIndex is set to 1, for the one without it is set to 2.
The feature from the layer with decluttering is drawn above the feature of the other layer with higher zIndex.

**Expected behavior**
Decluttering should not influence the zIndex order and higher zIndex should be above lower zIndex.

ahocevar:
This behavior is caused by an optimization where OpenLayers tries to combine multiple layers on a single canvas. To opt out of this behavior, configure at least one of the layers with a `className` other than `ol-layer`. See my updated codesandbox, where I set the `className` of `otherLayer` to `otherLayer`: https://codesandbox.io/s/exciting-feather-klyyl.

M393:
Thank you, this was indeed the problem.

Is this behaviour documented somewhere? 
I did not see it in the ol.layer.Vector documentation.

jahow:
It has been documented in `ol/layer/Layer`:
https://openlayers.org/en/latest/apidoc/module-ol_layer_Layer-Layer.html

Agreed, it'd be nice to have something mentioned in `ol/layer/Vector` specifically for declutter.

mike-000:
It should also be mentioned in ol/layer/VectorTile https://stackoverflow.com/questions/58442141/openlayers-6-adding-second-vector-tile-layer-causes-base-layer-labels-to-disap

Issue #9970, opened 2019-09-24
VectorTile Label Always Rendered on Topmost level


GISerWYJ:
**Describe the bug**
 After  #9957 , now the vectorTile label is always rendered on the **top level**.

**To Reproduce**
https://codesandbox.io/s/mapbox-vector-tiles-yfncf?fontsize=14
![image](https://user-images.githubusercontent.com/12789732/65475373-f1ded100-deb1-11e9-9b9d-9a2fdec19a6b.png)

two canvas elements exisit in the ol-layer element.
![image](https://user-images.githubusercontent.com/12789732/65476548-de356980-deb5-11e9-8c00-39a63a092191.png)

**Expected behavior**
vector tile label should rendered the same as the host vectorLayer.

ahocevar:
Vector labels rendered on top of everything is by design.

ahocevar:
What you can do to render labels next to their layer is to configure the layer with a className:

    new VectorTileLayer({
      className: ‘base‘
    })

GISerWYJ:
add a 'className' worked for me ! Thank you for your reply .

PR #10239, opened 2019-10-31
Simpler container reuse

ahocevar:
This pull request changes vector tile layer rendering so a single target container is sufficient. Previously, a separate container was used for text and image styles.

This significantly simplifies the container reuse code, improves rendering performance, and reduces memory consumption. However, there is a drawback: text and images on rotated map views will be a bit blurry on devices with a pixel ratio of 1.

In the process of making this change, I also noticed that CSS matrix transforms have rounded matrix components, which tend to be not equal to our generated CSS matrix string when comparing for container reuse. So instead of comparing a generated CSS matrix string with an applied CSS matrix style, we now apply the generated string to a test canvas to get consistent results.

This change significantly improves rendering performance on rotated maps, which now reuse containers the same way as unrotated maps. This is most significant when rotating the `mapbox-style.html` example.

This pull request should also improve the situation with some of the issues that have been reported since v6.x, regarding unexpected order of rendered features, where setting a `className` on the layer was a workaround, but not an intuitive solution. See #9970 and #10096.


tschaub:
Thanks for your continued efforts on this.

PR #10243, opened 2019-11-01
Store correct font value

### LINKED ISSUE #10242: Labels are incorrectly positioned when setting textAlign: Left in OL6
**Describe the bug**
If you set textAlign:left on a text label in OL6 the labels are now shifted noticeably to the right. I can reproduce this issue on iOS with Google Chrome and Ubuntu with Google Chrome. It doesn't seem to happen on Android with Google Chrome

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/vector-labels.html and set Align to "Left" for points and click Refresh. 
2. Then go to https://openlayers.org/en/v5.3.0/examples/vector-labels.html and set Align to "Left" for points and click Refresh. 
3. The label position is shifted way to the right in OpenLayers 6.1.0 unlike in 5.3.0.

**Expected behavior**
The labels should be positioned in the same place in OL6 and not be shifted to the right.

ahocevar:
Fixes #10242.


greeeg:
@ahocevar Thanks for your reactivity, can't wait to see this fix in the next release 🧙‍♂️

src/ol/render/canvas.js
```diff
@@ -333,7 +333,8 @@ export const measureTextHeight = (function() {
 export function measureTextWidth(font, text) {
   const measureContext = getMeasureContext();
   if (font != measureFont) {
-    measureContext.font = measureFont = font;
+    measureContext.font = font;
+    measureFont = measureContext.font;
```
tschaub: Are we doing this because the getter doesn’t always return what we set above?  This is very subtle.

Somewhat unrelated, but I’d be in favor of enforcing https://eslint.org/docs/rules/no-multi-assign (I know it is a stylistic preference).

ahocevar: > Are we doing this because the getter doesn’t always return what we set above? This is very subtle.

Yes indeed. Similar to what I recently encountered in 3de05c249bc761e7e3b0193c7e08f94e12110885.

> Somewhat unrelated, but I’d be in favor of enforcing https://eslint.org/docs/rules/no-multi-assign (I know it is a stylistic preference).

Agreed. I can create a follow-up pull request with that.

PR #10257, opened 2019-11-06
Allow View.adjust* methods to take a null opt_anchor

brianhelba:
The methods `View.adjustRotation`, `View.adjustZoom` and `View.adjustResolution` optionally take an `opt_anchor` parameter. When `opt_anchor` is `undefined`, the methods work properly. When it is `null`, they currently attempt to access the `null` value as though it is a `Coordinate` object, and throw:
```
TypeError: Cannot read property '0' of null
```

This change allows `opt_anchor` to safely be `null`.

Existing code paths already expect this to be supported. For example, when `MouseWheelZoom.useAnchor_` is `false`, then [`MouseWheelZoom.lastAnchor_` remains `null`](https://github.com/openlayers/openlayers/blob/77bc6897dd4f4dd47d7f6d4dc34fb9d4c5a34a29/src/ol/interaction/MouseWheelZoom.js#L166); thus, this `null` value [is passed to `View.adjustZoom` as `opt_anchor`](https://github.com/openlayers/openlayers/blob/77bc6897dd4f4dd47d7f6d4dc34fb9d4c5a34a29/src/ol/interaction/MouseWheelZoom.js#L209).

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1252,7 +1252,7 @@ class View extends BaseObject {
     const size = this.getSizeFromViewport_(this.getRotation());
     const newResolution = this.constraints_.resolution(this.targetResolution_ * ratio, 0, size, isMoving);
 
-    if (opt_anchor !== undefined) {
+    if (opt_anchor) {
       this.targetCenter_ = this.calculateCenterZoom(newResolution, opt_anchor);
     }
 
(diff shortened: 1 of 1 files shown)


brianhelba:
@jahow Thanks for the quick review!

Issue #8846, opened 2018-10-23
Make it possible to import the library outside a browser


tschaub:
It should be possible to import `ol` modules outside a browser (e.g. in Node) without needing a DOM/HTML implementation.  This does not mean that all functionality in the library needs to work outside a browser, just that importing a module should not require DOM access.  This should be possible if all DOM related access is done lazily.

fredj:
See also #7914

KaiVolland:
I'll have a look at this, lets see if i can do something.

KaiVolland:
Im not sure which way to go here.
1. Are we looking for actual code changes that add all kind of checks to the modules? 
2. Or is the usage of a polyfill package – like [jsdom](https://www.npmjs.com/package/jsdom) – what we want?

I think 1. will blow up the code but 2. might increase the build size.

ahocevar:
@KaiVolland The goal is not to add any checks or a compatibility layer. It is simply to use functionality that depends on the DOM lazily. Not sure if many of those that are relevant are left. Good examples are `ol/has.js` and `ol/pointer/PointerEvent.js`, which access the DOM as side effects. A positive side effect of such a change would also be that these side effects get removed.

Example from `ol/pointer/PointerEvent`:
```js
/**
 * Checks if the `buttons` property is supported.
 */
(function() {
  try {
    const ev = new MouseEvent('click', {buttons: 1});
    HAS_BUTTONS = ev.buttons === 1;
  } catch (e) {
    // pass
  }
})();
```
could be changed to something like
```js
let HAS_BUTTONS;
/**
 * Checks if the `buttons` property is supported.
 */
function hasButtons() {
  if (HAS_BUTTONS !== undefined) {
    return HAS_BUTTONS;
  }
  try {
    const ev = new MouseEvent('click', {buttons: 1});
    HAS_BUTTONS = ev.buttons === 1;
  } catch (e) {
    // pass
  }
};
```
Consumers would then access `hasButtons()` instead of `HAS_BUTTONS`.

KaiVolland:
Thx @ahocevar for the clarification. I did some further investigations and found some other relevant passages. As i'm offline until April 16 – and someone likes to work on this in the meantime – i'll leave my notes in this ticket:

[...]

PR #10262, opened 2019-11-06
Rework to make Document and XMLSerializer lazy and injectable

bjornharrtell:
Intends to support registering external implementations for Document and XMLSerializer so that XML based formats can be used outside browser environments.

Related to https://github.com/openlayers/openlayers/issues/8846.

### DIFF
--- a/src/ol/dom.js
+++ b/src/ol/dom.js
@@ -133,7 +133,7 @@ let domImplementation_ = undefined;
  * @api
  */
 export function registerDOMImplementation(domImplementation) {
-  domImplementation_ = document;
+  domImplementation_ = domImplementation;
 }
 
 let document_ = undefined;
--- a/src/ol/dom.js
+++ b/src/ol/dom.js
@@ -123,31 +123,26 @@ export function replaceChildren(node, children) {
   }
 }
 
-let domImplementation_ = undefined;
+let document_ = undefined;
 
 /**
- * Register an external DOMImplementation. Can be used to supply a DOMImplementation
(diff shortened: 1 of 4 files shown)


src/ol/dom.js
```diff
@@ -122,3 +122,32 @@ export function replaceChildren(node, children) {
...
+
+/**
+ * Register an external DOMImplementation. Can be used to supply a DOMImplementation
+ * in non-browser environments.
+ *
+ * @param {DOMImplementation} domImplementation A DOMImplementation.
+ * @api
+ */
+export function registerDOMImplementation(domImplementation) {
+  domImplementation_ = document;
```
tschaub: Do you mean to use the provided argument?

bjornharrtell: Yes, that was a blunder and due to that my first iteration was registerDocument. Will fix.. but so would it be preferable to inject Document instead of DOMImplementation?

The reason for using the name `register` instead of `set` is because I thought this to be a bit similar to proj4 injection which is done with method `register` in the `proj4` module.

bjornharrtell:
Reworked to make both Document and XMLSerializer injectable. Successfully tested using the WFS format in Node 10 by registering implementations from https://www.npmjs.com/package/jsdom and https://www.npmjs.com/package/w3c-xmlserializer.

bjornharrtell:
@jahow makes sense, I'll try to revisit this in the holidays.. :)

bjornharrtell:
@jahow reworked and green now. :)

PR #10269, opened 2019-11-08
Replace the BNG layer source in the Raster Reprojection example

mike-000:
The MapTiler MiniScale source used for the BNG layer is no longer working.  The data continues to be available via ESRI UK https://www.arcgis.com/home/item.html?id=5b5ff94ca7414a26b75e0d337a023d38 which is best set up using WMTS Capabilities due to the irregular resolutions of the tilegrid.

### DIFF
--- a/examples/reprojection.js
+++ b/examples/reprojection.js
@@ -55,15 +55,6 @@ proj54009.setExtent([-18e6, -9e6, 18e6, 9e6]);
 
 const layers = {};
 
-layers['bng'] = new TileLayer({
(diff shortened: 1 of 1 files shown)


mike-000:
Although Esri UK describe this as a free service https://www.esriuk.com/content-services/online-content-services/free-services covered by the OS OpenData license if you would prefer not to use Esri I have also downloaded the original Mini Scale data from OS OpenData and created a my own grid https://codesandbox.io/s/reprojection-ykxe3

Issue #3899, opened 2015-07-13
ol.control.MousePosition reports invalid lat/lon on wrapped maps.


pconnell99:
When I implement a 'many world's wrapX implementation with my vector layer and a tiled map layer, the mouseposition control reports invalid coordinates

Example is at https://jsfiddle.net/smtxcdeo/ - put your mouse over world +1 or -1 and the coordinates system reported is greater that 180/-180.

```
map.addControl(new ol.control.MousePosition({
className: "mapCoords",
projection: "EPSG:4326",
            coordinateFormat: function(coordinate) {
                return ol.coordinate.format(coordinate, '{y}\', {x}\'', 4);
            },
            target: 'coordinates',
            undefinedHTML: ""
}));
```

Is this by design? Is this a bug? How would I get it to report the correct lat/lon (mouseover of a feature on the other 'worlds' returns a co-ordinate within world 1 as expected - this causes other issues with popups etc.)

jeansouza:
I've created a function to resolve the "problem" in the app that I'm building, maybe useful for somebody:

  ```
/**
   * @author Jean Souza [jeancfsouza@gmail.com] - Paulo Remoli [paulo002@gmail.com]
   *
   * @function normalizeLongitude
   */
  var normalizeLongitude = function(longitude) {
    var correctedLongitude = longitude;
    var originalLongitude = longitude;

    if(longitude > 180 || longitude < -180) {
      longitude = longitude < 0 ? longitude * -1 : parseFloat(longitude);

      var isDivisionResultEven = (parseInt(longitude / 180) % 2) === 0;
      var divisionResult = isDivisionResultEven ? parseInt(longitude / 180) : Math.ceil(longitude / 180);

      if(divisionResult > 0)
        correctedLongitude = (originalLongitude < 0) ? originalLongitude + (divisionResult * 180) : longitude - (divisionResult * 180);
    }

    return correctedLongitude;
  };
```

**Comment by chander** [comment]
+1 - this bug is annoying.  However, the fix would need to also account for projections other than EPSG:4326 that might also wrap, which might be harder than this...

Won't this do the same thing, but without the complexity (when your control is defined to use the EPSG:4326 projection):

```
/**
 * @function correctLongitude
 * Take a longitude that is wrapped around the IDL multiple times and unwrap it to a value
 * that is between -180 and 180.
 */
function correctLongitude(value) {
    // Remove wraps after the first wrap.
    v= v % 360;
[...]

PR #10274, opened 2019-11-09
Fix longitudes that wrap around the IDL when converting from EPSG:3857 to EPSG:4326

mstenta:
This adds a bit of logic to the `toEPSG4326()` function in `src/ol/proj/epsg3857.js` (for transforming coordinates from `EPSG:3857` to `EPSG:4326`), so that longitudes cannot be less than -180 or greater than +180. This issue occurs in various circumstances (described in #3899), including when drawing shapes after the map has been panned across the international date line.

See #3899

### DIFF
--- a/src/ol/proj/epsg3857.js
+++ b/src/ol/proj/epsg3857.js
@@ -140,6 +140,14 @@ export function toEPSG4326(input, opt_output, opt_dimension) {
   }
   for (let i = 0; i < length; i += dimension) {
     output[i] = 180 * input[i] / HALF_SIZE;
+    // fix longitudes that wrap around the international date line.
+    output[i] %= 360;
+    if (output[i] < -180) {
+      output[i] += 360;
+    }
+    if (output[i] > 180) {
+      output[i] -= 360;
+    }
     output[i + 1] = 360 * Math.atan(
       Math.exp(input[i + 1] / RADIUS)) / Math.PI - 90;
   }


mstenta:
Looking into the failing tests now...

mstenta:
Thank you @mike-000 for taking the time to look at this and provide feedback!

The "drawing across dateline" consideration is not something that occurred to me - and @chander just made the same point in #3899 where this started. That makes sense.

In your opinion, would this change to `toEPSG4326()` be sensible in a limited use-case where we are don't expect shapes to cross the date line? I made a proof-of-concept Gist that overrides the `toEPSG4326()` function without patching the OL library code itself: https://gist.github.com/mstenta/3d7e6a02c37bbf9dd246116712c6bd8b

In my case, I'm developing an application for farmers, which involves drawing polygons around field boundaries, so (I hope) they will never cross the date line. Or do you think this would cause other problems I'm not foreseeing?

Thanks again for your feedback!

ahocevar:
@mstenta I think you need to consider the whole geometry, not just the current coordinate for adjusting the coordinate. In OpenLayers, you could do that with code like this:
```js
/**
 * Adjusts the geometry's coordinates to be in the range of -180 to +180. Geometries that are
 * crossing the date line are not modified.
 * @param {import("ol/geom/Geometry.js").default} geometry Geometry in EPSG:4326 projection
 */
function adjustWorld(geometry) {
  let extent = geometry.getExtent();
  while (extent[0] < -180 && extent[2] < -180) {
    geometry.translate(180, 0);
    extent = geometry.getExtent();
  }
  while (extent[0] > 180 && extent[2] > 180) {
    geometry.translate(-180, 0);
    extent = geometry.getExtent();
  }
}
```

mike-000:
You might also want to look into splitting geometries so all coordinates lie between -180 and +180.  There's a codesandbox example using turf.js in #10289

mstenta:
Thank you @ahocevar and @mike-000! I appreciate your feedback (and ALL the work you do)!

I'll look into your suggestions for our specific use-case.

PR #10275, opened 2019-11-09
fix type

arszp10:


ahocevar:
Can you please provide some context for these changes? Could it be that you wanted to work around microsoft/TypeScript#33575?

PR #10285, opened 2019-11-12
Fix WMS GetLegendGraphic example codesandbox error

mike-000:
`src=""` is not needed in the html and causes an error when the example is opened in codesandbox


tschaub:
Thanks @mike-000.

PR #10301, opened 2019-11-17
Create hit detection data per layer and without requestAnimationFrame

ahocevar:
Calculating hit detection data in an animation frame properly would require more effort, to avoid creating it multiple time if multiple calls to `getFeatures()` were made before the animation frame. Looking at the performance profile, I decided to calculate it immediately, but only when we are not animating or interacting. For vector layers, this was the case already.

With vector tile layers, the previous code did extra work when a source was shared among multiple layers. By storing hit detection data on the render tile for each layer that uses the render tile (like we do for the executor groups), we can improve that.

Fixes #10250

### DIFF
--- a/src/ol/VectorRenderTile.js
+++ b/src/ol/VectorRenderTile.js
@@ -61,9 +61,9 @@ class VectorRenderTile extends Tile {
     this.errorSourceTileKeys = {};
 
     /**
-     * @type {ImageData}
+     * @type {Object<number, ImageData>}
      */
-    this.hitDetectionImageData = null;
+    this.hitDetectionImageData = {};
 
     /**
      * @private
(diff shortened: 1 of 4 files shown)


ahocevar:
I need to revisit the hit detection image data part. Looks like we’re sharing render tiles between layers that use the same source, which we should not be doing.

ahocevar:
I just confirmed, the way we key hit detection image data by layer uid is ok and consistent with what we do for the canvas context and the executor groups.

I'm going to add a few more tests, and report back when this is ready for review.

ahocevar:
I added more tests now - this is ready for review.

ahocevar:
Thanks for any review.

PR #10309, opened 2019-11-18
Bring back vector render mode for vector tile layers

ahocevar:
The vector render mode for vector tiles (i.e. no images for the render tiles are generated) was removed before v6. This pull request brings it back, because it is useful for creating highlight/selection layers for vector tile layers. I updated the vector-tile-selection example to show that.


ahocevar:
Note that hover selection performance in the updated https://454-4974939-gh.circle-artifacts.com/0/examples/vector-tile-selection.html example will be significantly improved when #10301 is merged.

ahocevar:
Thanks for any review.

examples/vector-tile-selection.html
```diff
@@ -5,7 +5,11 @@
 docs: >
   <p>
     Click a rendered vector-tile feature to highlight it on the map. Click on an empty spot (ocean) to reset the selection.
-    By changing the action type to "Multi Select" you can select multiple features at a time.
+    By changing the action type to "Multi Select" you can select multiple features at a time. With "Single Sslect on hover",
```
fredj:
```suggestion
    By changing the action type to "Multi Select" you can select multiple features at a time. With "Single Select on hover",
```

PR #10310, opened 2019-11-18
Ensure that OverviewMap respects the initial rotation of attached Maps

brianhelba:
Without this fix, the updated tests here will fail.

### DIFF
--- a/src/ol/control/OverviewMap.js
+++ b/src/ol/control/OverviewMap.js
@@ -304,6 +304,8 @@ class OverviewMap extends Control {
    */
   bindView_(view) {
     view.addEventListener(getChangeEventType(ViewProperty.ROTATION), this.boundHandleRotationChanged_);
+    // Sync once with the new view
+    this.handleRotationChanged_();
   }
 
   /**
--- a/test/spec/ol/control/overviewmap.test.js
+++ b/test/spec/ol/control/overviewmap.test.js
@@ -35,7 +35,7 @@ describe('ol.control.OverviewMap', function() {
       const view = new View({
         center: [0, 0],
         zoom: 0,
-        rotation: 0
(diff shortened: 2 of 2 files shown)


brianhelba:
@mike-000 @tschaub This code was recently touched by you in #9919.

src/ol/control/OverviewMap.js
```diff
@@ -304,6 +304,8 @@ class OverviewMap extends Control {
    */
   bindView_(view) {
     view.addEventListener(getChangeEventType(ViewProperty.ROTATION), this.boundHandleRotationChanged_);
+    // Sync once with the new view
+    this.handleRotationChanged_();
```
brianhelba: I was a little uncertain about where to put this call. Ideally, it would have been nice to keep `bindView_` "clean", so it only does event binding. However, it is simpler to re-sync the rotation once here, rather than in every other place where the Map or Map's View are changed (especially to protect against future code changes, where changes could happen in even more places).

DaniEll-AT: It is a common pattern to connect event-listeners and perform the first sync of state at once.
How else would one guarantee that there is no initial inconsistency?

mike-000: I agree it does make sense to sync on binding when the rotation can be already set.  This seems to have been a problem ever since version 4 when rotate with view was the only option, and it has not been noticed because none of the tests started with an already rotated view, so some amended tests would also be desirable.

brianhelba: @mike-000 Indeed, this PR adds testing for already-rotated views, and should now detect any regressions of this bug.

Thanks for the feedback, @DaniEll-AT @mike-000

PR #10315, opened 2019-11-18
Replace Bing sources in some examples

### LINKED ISSUE #10294: Bing based examples fail to show map
**Describe the bug**
Bing based example on OpenLayers website show gray map zone with nothing inside.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/index.html?q=bing
2. Open any of the filtered examples
3. Map zone only show + and – zoom buttons

**Expected behavior**
Map with Bing map background?

mike-000:
Fixes #10294

Replace Bing sources with Esri World Streets and Esri World Imagery in some examples where Bing is not specifically required,

### DIFF
--- a/examples/feature-move-animation.html
+++ b/examples/feature-move-animation.html
@@ -6,10 +6,10 @@ docs: >
   This example shows how to use <b>postrender</b> events and a <b>vector context</b> to
   animate a marker feature along a line. In this example an encoded polyline
   is being used.
-tags: "animation, feature, postrender, polyline"
+tags: "animation, feature, postrender, polyline, maptiler"
 cloak:
-  - key: As1HiMj1PvLPlqc_gtM7AqZfBL8ZL3VrjaS3zIb22Uvb9WKhuJObROC-qUpa81U5
-    value: Your Bing Maps Key from http://www.bingmapsportal.com/ here
+  - key: get_your_own_D6rA4zTHduk6KOKTXzGB
+    value: Get your own API key at https://www.maptiler.com/cloud/
 ---
 <div id="map" class="map"></div>
 <label for="speed">
--- a/examples/feature-move-animation.js
+++ b/examples/feature-move-animation.js
@@ -4,7 +4,7 @@ import View from '../src/ol/View.js';
 import Polyline from '../src/ol/format/Polyline.js';
 import Point from '../src/ol/geom/Point.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
-import BingMaps from '../src/ol/source/BingMaps.js';
+import XYZ from '../src/ol/source/XYZ.js';
 import VectorSource from '../src/ol/source/Vector.js';
 import {Circle as CircleStyle, Fill, Icon, Stroke, Style} from '../src/ol/style.js';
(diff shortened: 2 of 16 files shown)


ahocevar:
Can you please provide a link to the terms of use of the Esri layers, to make sure we can use them this way in the examples?

ahocevar:
Looks like I'm right and we cannot use these layers. See the [Terms of use link](http://downloads2.esri.com/ArcGISOnline/docs/tou_summary.pdf) at the bottom of https://www.arcgis.com/home/item.html?id=a2e7c99be14d421abac4f002d6c301f5. Even worse, I believe that Esri's world imagery is just a wrapped Bing Maps service, and the linked terms do not apply to Bing Maps. In that case, Bing Maps terms apply, and we're back to where we were.

I would suggest for all examples that do not need imagery data to use the OSM or Stamen source.

mike-000:
All those examples use imagery (two also use streets).  I had assumed it would be OK as the xyz-esri and xyz-esri-4326-512 examples already used Esri sources but you may also need to review those.

ahocevar:
Good point. @bartvde, I think you did the work on Esri sources back then. Do you remember if there was any agreement with Esri, or if the terms of use allow us to use these layers?

ahocevar:
@petrsloup @klokan Given the current situation, would it be ok for you guys if we switch to MapTiler Cloud raster maps as Bing replacement for the examples in this pull request?

klokan:
Yes. We are very happy to provide the satellite layer for demonstration purposes at openlayers.org domain with proper attribution links in place. Please use the previously provided key.

mike-000:
> Everything is fine, however `preload.html` sample is also using Bing map.

It uses Aerial with Labels - as does feature-move-animation (and of course bing-maps).  Neither Esri or MapTiler have a single source equivalent for that.   But removing Bing from 6 examples will free up API key quota for the remaining 3 examples.

klokan:
Aerial/Satelite with burned-in labels: https://cloud.maptiler.com/maps/hybrid/
Pure Aerial/Satelite: https://cloud.maptiler.com/tiles/satellite/
BTW we recommend to use 512px tiles / tilgrid (less tile requests=more efficiency).

tschaub:
Thanks @klokan!

mike-000:
@klokan Is pure satellite available as 512px tiles?

bartvde:
> Good point. @bartvde, I think you did the work on Esri sources back then. Do you remember if there was any agreement with Esri, or if the terms of use allow us to use these layers?

I don't think there was any specific agreement, I am not sure if the terms of use have changed since then

bartvde:
@mpriour do you maybe know more about the Esri terms of use?

emcastro:
> @mpriour do you maybe know more about the Esri terms of use?

Not formally forbiden, not formally authorized.

Summary:
http://downloads2.esri.com/ArcGISOnline/docs/tou_summary.pdf

Complete terms of use:
https://www.esri.com/en-us/legal/terms/full-master-agreement

Involved ressources:
* World Street Map: https://www.arcgis.com/home/item.html?id=3b93337983e9436f8db950e38a8629af
* World Imagery: https://www.arcgis.com/home/item.html?id=10df2279f9684e4a9f6a7f08febac2a9

PR #10318, opened 2019-11-19
Use a minimalist pointer events polyfill

ahocevar:
I've been quite unhappy with size vs. quality of the PEP polyfill for pointer events. I wanted to create a minimal polyfill myself, but before I did I fortunately found the awesome [elm-pep](https://github.com/mpizenberg/elm-pep) polyfill, which does enough for OpenLayers to work properly on browsers that do not support pointer events (which are a diminishing fraction).

Another nice thing about this change is that builds are now 17kB smaller.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "url": "https://github.com/openlayers/openlayers/issues"
   },
   "dependencies": {
-    "@openlayers/pepjs": "^0.5.3",
+    "elm-pep": "^1.0.2",
     "pbf": "3.2.1",
     "pixelworks": "1.1.0",
     "rbush": "^3.0.1"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -81,6 +81,7 @@
     "loglevelnext": "^3.0.1",
     "marked": "0.7.0",
     "mocha": "6.2.2",
+    "napa": "^3.0.0",
```
DaniEll-AT: why add napa, it seems to not be required for elm-pep?

ahocevar: Right, that's an accidental leftover from a previous attempt trying to integrate https://github.com/mpizenberg/elm-pep directly from github, which does not have a package.json. Removed.

illogicz:
elm-pep is causing issues with internet explorer due to arrow functions not being transpiled.

ahocevar:
Thanks for reporting, I’ll update the package.

PR #10321, opened 2019-11-20
Support touch events for DragBox interaction 

### LINKED ISSUE #10290: Support touch events for DragBox interaction
It would be beneficial for the DragBox interaction to support touch events. I use the DragBox interaction for numerous other map interactions and it would be nice to easily support these on mobile.

The DragBox currently supports a condition filter that defaults to always. In addition to applying that condition, it also artificially limits the events it handles to mouseOnly and mouseActionButton.
https://github.com/openlayers/openlayers/blob/8b036ba96cb3d113c7115fd24884af55ec874af4/src/ol/interaction/DragBox.js#L220-L226

I think it would be preferable for DragBox to not limit to mouseOnly and mouseActionButton conditions and only use the conditions filter. I believe the conditions filter could default to mouseActionButton to keep current compatibility and others like myself could override this default to support DragBox functionality on touch events. 

If there is agreement on this change/improvement, I would be happy to submit a PR with the described changes.

greggian:
Update DragBox to use only condition for filtering
- Remove mouseOnly and mouseActionButton hardcoded filters
- Default condition to mouseActionButton

Resolves #10290

### DIFF
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -3,7 +3,7 @@
  */
 // FIXME draw drag box
 import Event from '../events/Event.js';
-import {always, mouseOnly, mouseActionButton} from '../events/condition.js';
+import {mouseActionButton} from '../events/condition.js';
 import {VOID} from '../functions.js';
 import PointerInteraction from './Pointer.js';
 import RenderBox from '../render/Box.js';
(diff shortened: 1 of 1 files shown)


greggian:
Note, the user facing changes of this PR can be seen on the box-selection and drag-zoom examples. Both now support touch events when using the appropriate modifier key (control or shift keys respectively).

fredj:
Looks good, thanks

Please update the `typedef` and change the `condition` default value to `{@link ol/events/condition~mouseActionButton}` and remove `This interaction is only supported for mouse devices.`

greggian:
@fredj I updated the jsdoc accordingly. Sorry, I missed those references on the first pass.

PR #10330, opened 2019-11-23
Fix modifying circle geometries

### LINKED ISSUE #10316: modify interaction can not be used for 'Circle'
**Describe the bug**
see https://openlayers.org/en/latest/examples/draw-and-modify-features.html

**To Reproduce**
1. Draw a circle feature
2. Modify the feature's radius
3. The feature has no change, but the center can be changed

greggian:
The Snap interaction may have altered the event coordinate, so Modify
should use the event coordinate whenever available.

Resolves #10316

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -807,8 +807,8 @@ class Modify extends PointerInteraction {
     if (!this.condition_(evt)) {
       return false;
     }
-    this.handlePointerAtPixel_(evt.pixel, evt.map);
     const pixelCoordinate = evt.coordinate;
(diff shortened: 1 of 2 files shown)


greggian:
I would like to add a test to prevent this regression from happening again. There is already a test for modifying circles and it was passing because this bug appears to be specific to the modify interaction when the snap interaction is also in use.

I welcome advice on the best way to handle this sort of test. Should the test just be in the snap or modify test spec? Should there be a separate spec for using both interactions together? Thanks

ahocevar:
Thanks for fixing this! You can add that test in modify.test.js.

greggian:
Thanks @ahocevar. Test added.

PR #10343, opened 2019-11-26
Correct interactions with circle geometries when using user coordinates

### LINKED ISSUE #10328: Modify interaction with useGeographic does not interact at drawn circle circumference
**Describe the bug**
Additionally (as observed while investigating #10316) the modify interaction does not act on the circle center as it should.

**To Reproduce**
Draw a circle in this example https://codesandbox.io/s/draw-and-modify-features-voi73  (which will be yellow and circular in the view) and the code will add a red circular polygon based on geographical coordinates which is oval in the view.  Hover over both circles and see that the modify interaction is activated only over the red oval, but modifies the radius of the yellow circle.  Also it does not activate over the circle center.

However, I cannot see any obvious link between this problem and problems using modify in conjunction with snap in view coordinates in #10316 as there is no problem with using only modify as above in view coordinates.

mike-000:
Fixes #10328

Correct modify interaction at center and at drawn circle circumference when using user coordinates.
Corresponding update to snap interaction at circumference .

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -14,13 +14,14 @@ import {always, primaryAction, altKeyOnly, singleClick} from '../events/conditio
 import {boundingExtent, buffer as bufferExtent, createOrUpdateFromCoordinate as createExtent} from '../extent.js';
 import GeometryType from '../geom/GeometryType.js';
 import Point from '../geom/Point.js';
+import {fromCircle} from '../geom/Polygon.js';
 import PointerInteraction from './Pointer.js';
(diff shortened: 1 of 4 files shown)


ahocevar:
Thanks @mike-000 for this contribution! @tschaub, can you please take a look at this?

tschaub:
Bump.

tschaub:
Example with the fix applied https://2781-4723248-gh.circle-artifacts.com/0/examples/draw-and-modify-features.html

tschaub:
Thank you for your work on this, @mike-000!

PR #10344, opened 2019-11-26
Fix lazy UTFGrid loading

### LINKED ISSUE #10342: UTFGrid issue with preemptive false and tileJSON
**Describe the bug**

I try to use an UTFGrid source with the `preemptive` option set to `false` and a `tileJSON` configuration, which requires the use of `{x}+{y}+{z}` in the url.  These get automatically replaced by the tile url and mouseoverring the layer works properly.

Something else seems to break: the tile queue.  There are tiles from UTFGrid that have keys that never leave the queue and prevents the whole map from loading tiles.

In other words, in a map where you have 2 layers, one OSM and an UTFGrid configured like mentioned above, then trying to zoom in result in the OSM tiles to not load properly.

If I try to remove the `{x}+{y}+{z}` values from the tileJSON url, then it works.  So, it seems to be an issue related with the keys... but I haven't found out the cause yet.

**To Reproduce**
1. Clone this branch: https://github.com/adube/ol3/tree/utfgrid-issue-preemptive-false, (or apply this patch: https://github.com/adube/ol3/commit/201d6d52e3eeaf1c1e3b9d18d9d9769c6cdd7f2b)
2. run the example locally (npm run serve-examples, then visit the example "utfgrid.html")
3. zoom in a couple times
4. you'll see that the tiles from OSM stop from working --> I think the queue is full, i.e. not emptied properly

**Expected behavior**
The queue should be emptied properly, and OSM tiles properly loaded.

ahocevar:
Fixes #10342.

We cannot leave the tile IDLE. That causes the tile queue not to make room for other tiles that need to be loaded. By setting the tile state to EMPTY, we now tell the queue that there is nothing to do with the tile.

### DIFF
--- a/src/ol/source/UTFGrid.js
+++ b/src/ol/source/UTFGrid.js
@@ -143,7 +143,8 @@ export class CustomTile extends Tile {
    *                               The tile data is requested if not yet loaded.
    */
   forDataAtCoordinate(coordinate, callback, opt_request) {
-    if (this.state == TileState.IDLE && opt_request === true) {
+    if (this.state == TileState.EMPTY && opt_request === true) {
+      this.state = TileState.IDLE;
       listenOnce(this, EventType.CHANGE, function(e) {
         callback(this.getData(coordinate));
       }, this);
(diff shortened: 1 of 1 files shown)


adube:
Am I allowed to approve this PR?

PR #10363, opened 2019-12-01
Interactive SVG layer example

### LINKED ISSUE #4633: Unable to use SVG image as layer
Hi All,

I'm sorry to post it here but I did not get any good information on stackoverflow.

My question is simple: Can I use a SVG image as a layer with OpenLayers-3? I did not find anything on the website but because it is such a commonly used format I want to be sure.

For reference my question on stackoverflow: https://stackoverflow.com/questions/34416412/how-can-i-use-a-svg-image-as-layer-on-openlayers-3

### LINKED ISSUE #8105: SVG image as layer is blurry in OpenLayers 4
As shown here: https://jsfiddle.net/status420/mhfn9anm/2/
When you use a SVG file in a image layer it does not scale properly always being at the minimum scale making it useless.

This has apparently been an issue for [over a year](https://github.com/openlayers/openlayers/issues/4633), but no solution has appeared. Since that last issue is ignored because it is wrongly closed i am making a new one here.

ahocevar:
This pull request adds an example that shows how to use the new Layer `render()` API to create an interactive SVG layer.

To make working with transforms in custom layer renderers easier, I added a new `composeCssTransform` function to the `ol/transform` module.

To allow interacting with the SVG, I needed to add two css rules to the OpenLayers css.

See #4633, fixes #8105.


ahocevar:
> I'm unsure as to which CSS core rule is overridden to make the example work though.

That was only the case in a previous iteration of this pull request. I'm updating the description now.

PR #10384, opened 2019-12-05
Include Transform typedef in API

### LINKED ISSUE #10377: Documentation dead link
**Describe the bug**
There is a 404 error on the link to the  inversePixelTransform definition in 
https://openlayers.org/en/latest/apidoc/module-ol_render_Event-RenderEvent.html

mike-000:
Fixes #10377

It should be in the API as it is referenced in https://openlayers.org/en/latest/apidoc/module-ol_render_Event-RenderEvent.html


fredj:
thanks @mike-000

PR #10393, opened 2019-12-08
getFeaturesInExtent function for ol/source/VectorTile

ahocevar:
Many users want to get features or render features from a VectorTile source. Not that it is a very good idea in general, but there are valid rendering-centric use cases like e.g. for creating a dynamic legend that only shows entries for classes of features that are currently rendered.

In addition to adding a `getFeaturesInExtent()` function, this pull request also simplifies the API for the VectorTile source's `tileLoadFunction`: that function can now call `VectorTile#setFeatures()` synchronously, and does not need to call `VectorTile#setLoader()` any more. The need for this simplification came up when I wrote the tests for the `getFeaturesInExtent()` function.

### DIFF
--- a/src/ol/VectorTile.js
+++ b/src/ol/VectorTile.js
@@ -105,7 +105,9 @@ class VectorTile extends Tile {
     if (this.state == TileState.IDLE) {
       this.setState(TileState.LOADING);
       this.tileLoadFunction_(this, this.url_);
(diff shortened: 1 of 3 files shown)


ahocevar:
Conflicts resolved and rebased. So this is ready for review again.

PR #10400, opened 2019-12-09
Rework KML parsing to bring some parser constants and functions into scope

mike-000:
This is a potential replacement for #10119

#10119 showed it was difficult to cleanly handle options from the constructor in unscoped callbacks from the parsers.  Using `.call` provides a quick fix but isn't obvious for maintainers.  Bringing style related parser constants and functions into scope as private properties/methods of the class and where necessary binding them (similar to the existing `readDocumentOrFolder_` but for efficiency only setting the former constants once in the constructor) make it much easier to process constructor options such as icon crossOrigin after parsing.

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -87,42 +87,19 @@ const ICON_ANCHOR_UNITS_MAP = {
   'insetPixels': IconAnchorUnits.PIXELS
 };
 
+
 /**
  * @const
  * @type {Object<string, Object<string, import("../xml.js").Parser>>}
  */
 // @ts-ignore
-const PLACEMARK_PARSERS = makeStructureNS(
+const ICON_STYLE_PARSERS = makeStructureNS(
   NAMESPACE_URIS, {
-    'ExtendedData': extendedDataParser,
-    'Region': regionParser,
-    'MultiGeometry': makeObjectPropertySetter(
-      readMultiGeometry, 'geometry'),
-    'LineString': makeObjectPropertySetter(
-      readLineString, 'geometry'),
-    'LinearRing': makeObjectPropertySetter(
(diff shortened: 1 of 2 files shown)


jahow:
I have examined these changes and am personnally not in favor of merging these for the reasons below.

Currently there are no parser constants in the class scope. Adding some while keeping others in the module scope will add complexity and make the KML code even less readable than it is now.

Same thing for functions, this for example looks weird at first glance:
```js
    this.style_parsers_ = makeStructureNS(
      NAMESPACE_URIS, {
        'IconStyle': this.iconStyleParser_.bind(this),
        'LabelStyle': labelStyleParser,
        'LineStyle': lineStyleParser,
        'PolyStyle': polyStyleParser
      });
```

If you look at the GML formats, these are structured differently as all parser constants and functions are stored on the prototype of the class, with almost nothing in module scope. This is IMO the way the KML format should work, and that would generally solve scope issues like the one you encountered. Unfortunately, this would be a quite big refactor of the format so I'd understand if this goes out of this PR's scope.

Regarding the original improvement (`crossOrigin` option), I'd be in favor of merging #10119 (even if it is a bit quirky) instead of this PR, as #10119 gets the work done in relatively few lines of code and introduces less complexity.

I'll let others review if they want to give a different opinion. Thanks for the work on this @mike-000.

mike-000:
If I had looked at the GML code earlier I would have taken that approach to keep the constants looking like constants, and I agree an all or nothing approach makes sense - however there is still a lot reworking of constants and functions (mostly geometry related) needed to achieve that, so if you wish to go with #10119 for now please do.

mike-000:
Closed due to #10119.  A more complete rework similar to GML3 could be potentially based on that.

Issue #10412, opened 2019-12-13
Constrain the ZoomSlider's range to the View's constrained resolution


MoonE:
**Is your feature request related to a problem? Please describe.**
The ZoomSlider's range is the View's minResolution and maxResolution. If the View has a constraint set the maximum selected value is less than the maxResolution. This can be confusing for a user.

**Describe the solution you'd like**
Limit the ZoomSlider 's range to the View's minResolution and constrained maxResolution.

PR #10413, opened 2019-12-13
Limit the ZoomSlider's range to the view's constrained resolution.

MoonE:
Implements #10412

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -945,7 +945,7 @@ class View extends BaseObject {
    */
   getResolutionForValueFunction(opt_power) {
     const power = opt_power || 2;
-    const maxResolution = this.maxResolution_;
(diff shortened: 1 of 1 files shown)


ahocevar:
Looking at the changes in `ol/View`, I think this is something for @jahow to review.

MoonE:
The ZoomSlider should update when max / min resolution changes, but this is not handled atm.
The needed events are missing, see  #10435.

jahow:
I'll review this ASAP.

MoonE:
The *getResolutionForValueFunction* function is only used by the *ZoomSlider* and not part of the api. I removed the the new *opt_constrained* parameter again and have it always constrain the resolution. The *ZoomSlider* is now using the new behavior without an option to configure it.

PR #10417, opened 2019-12-15
Replace Bing layer with MapTiler

mike-000:
Update the Image Filters example to replace Bing imagery with MapTiler


tschaub:
Thanks @mike-000.

PR #10428, opened 2019-12-16
Draw interaction: Append coordinates to polygons and lines

opelline:
This was built for an actual use case, as the user needs to be able to follow the borders on a vector layer, without snapping each edge point by hand. This could be a border of a region, or some part of a neighbouring feature. This of course requires some app specific handling for doing the snapping and slicing of those line segments before they can be added to the current geometry.

The extend feature doesn't allow adding points in the other end of a segment, and it also doesn't work on polygons. I was first trying to implement this using the extend feature, but I got confused and thought it was buggy. It was only after runnign the automated build tests that I actually understood what the extend feature was for, and why it was working the way it was. So now I've clarified the feature documentation to separate the two.

One might think all this could be achieved by just taking the coordinate arrays to be joined, concatenate them as you like, and creating a new feature, which then could be used in the extend feature. However, this doesn't work with polygons, as it will break the draw interaction. 

The changes requested by @jahow have also been added to the branch, and the example was changed to better illustrate a typical use case.

(The previous pull request was closed by stale bot. I asked how to reopen it, but didn't get any response, so I'm opening a new one. Some discussions are found in the old one https://github.com/openlayers/openlayers/pull/9929)


opelline:
Ping @jahow can you have a look at the updates?

jahow:
Awesome, thanks! I'll look into this ASAP.

src/ol/interaction/Draw.js
```diff
@@ -888,10 +887,46 @@ class Draw extends PointerInteraction {
...
+    const ending = coordinates.pop();
+
+    // Append coordinate list
+    for (let i = 0; i < coordinateExtension.length; i++) {
+      this.addToDrawing_(coordinateExtension[i]);
+    }
+
+    // Duplicate last coordinate for sketch drawing
+    this.addToDrawing_(ending);
+    //this.addToDrawing_(coordinateExtension[coordinateExtension.length - 1]);
```
jahow: To remove

src/ol/interaction/Draw.js
```diff
@@ -888,10 +887,46 @@ class Draw extends PointerInteraction {
...
+  }
+
   /**
    * Extend an existing geometry by adding additional points. This only works
```
jahow: Maybe this could be reworded a bit to be clearer, something like:
> Initiate draw mode by starting from an existing geometry which will receive new additional points. This only works...

src/ol/interaction/Draw.js
```diff
@@ -888,10 +887,46 @@ class Draw extends PointerInteraction {
...
    * on features with `LineString` geometries, where the interaction will
    * extend lines by adding points to the end of the coordinates array.
+   * This will change the original feature, instead of drawing a copy.
+   *
+   * Extend can be used without having triggered the startDrawing_ function.
```
jahow: This line should be removed as it mentions an internal method

src/ol/interaction/Draw.js
```diff
@@ -888,10 +887,46 @@ class Draw extends PointerInteraction {
...
+   * The function will dispatch a DRAWSTART event.
```
jahow:
```suggestion
   * The function will dispatch a `drawstart` event.
```

src/ol/interaction/Draw.js
opelline: Removed

src/ol/interaction/Draw.js
opelline: reworded

src/ol/interaction/Draw.js
opelline: removed

src/ol/interaction/Draw.js
opelline: fixed

opelline:
@jahow @tschaub I'm thinking about how not to make the example too complicated (such as handling the clicking and slicing and adding a preview for any segment between two points, in the right direction, along the existing feature). 

One solution that could be relatively straight forward would be the following:
There would be one point on the feature from which one could follow the feature to any of the other points. By clicking the starting point, and then clicking any ending point, the user could append a custom segment along the existing feature in a way that would mimic a realistic use case. The segment to be added could perhaps be highlighted on mouseover for clarity.

opelline:
Ping @jahow & @tschaub 
Would you have any suggestions?

jahow:
Sorry @opelline, I've been meaning to take another look at your contribution but didn't have the time. Please bear with me a few more days!

jahow:
@opelline So, how about the following use case for the example:
* the map initially only contains one polygon with a "realistic" shape, ie an administrative border or shape of a building
* a draw interaction is active, along with a snap interaction with the initial polygon
* when the user draws vertices on two different points of the initial polygon, a subset of this polygon is appended to the currently drawn shape, effectively doing an "auto snapping" on several vertices on the polygon (for lack of a better word)

I'll have some time to work on this next week, please let me know if you're OK with me modifying the example in your branch. Thanks!

jahow:
I took the liberty to create another PR with a different example, hope that's alright!

tschaub:
It looks like this should be closed after #10632.  Please reopen if that is wrong.

PR #10429, opened 2019-12-16
Option to disable image smoothing for Tile layer sources (including reprojected sources)

### LINKED ISSUE #10273: Effectively avoid image smoothing in tile layers

In a project I'm using OpenLayers (btw, thanks for this great open source package!) to render satellite imagery and gridded climate datasets using tile layers. I want the imagery to be rendered without any interpolation between pixel values, so no image smoothing / blurring occurs when zooming into the map. Similar requests have been posted on OpenStack some time ago, see [here](https://stackoverflow.com/questions/35875270/turn-off-image-smoothing-in-openlayers-3) and [here](https://stackoverflow.com/questions/54083424/preventing-smoothing-of-tileimage-layer). The suggestion was altering canvas settings in a `precompose` (ol 4 and 5) or `prerender` (ol 6) event handler:

```ts
        layer.on('prerender', (event: RenderEvent) => {
            const ctx = event.context;
            ctx.imageSmoothingEnabled = false;
            const ctxAny = ctx as any;
            ctxAny.webkitImageSmoothingEnabled = false;
            ctxAny.mozImageSmoothingEnabled = false;
            ctxAny.msImageSmoothingEnabled = false;
        });
```
but that doesn't have any effect on my layers with ol 4, 5, 6.1.1. If this is still the recommended way, I'd be glad if you could share any ideas why [it does not work in my case](https://github.com/dcs4cop/xcube-viewer/blob/f94b70948baf737fd7fecd79e7c8c9c13eb57d31/src/components/ol/layer/Tile.tsx#L41)?  

I'd like to have a tile layer options that effectively avoids image smoothing/blurring, i.e. forces nearest neighbor resampling.

I reported [something similar](https://github.com/AnalyticalGraphicsInc/cesium/issues/5846) for the Cesium project; I provided a pull request to them and [here](https://sandcastle.cesium.com/?src=Imagery%20Layers%20Texture%20Filters.html) is the result in their sandbox. Unfortunately, I currently do not have the time to do the same for OpenLayers.

mike-000:
Fixes #10273

Provides a means to set canvas context properties such as `ctx.imageSmoothingEnabled = false;` in the reprojection process for reprojected tile sources (TileImage and any subclasses which support the `reprojectionErrorThreshold` option).  Adds an example of a reprojected DEM layer with image smoothing disabled.

This could easily be extented in a separate PR to include Image sources and provide a complete fix for #9001 (prerender events alone do not fix the problem with reprojection described there).


mike-000:
I've checked these changes in conjunction with those proposed in #10463 and used a ol/source/Raster to highlight any obviously corrupt DEM values occurring at reprojection edges.  The removal of the triangulation overlap reduces that considerably although using a `lighter` operation makes it harder to detect (as corrupt values can be fully opaque) but where necessary that operation could overridden by `opt_contextOptions`.  If #10463 is to be merged it would be best to delay merging this PR until then, rebase and make any changes needed to integrate with it.

There is also potential for a future enhancement to apply the source's `reprojectionContextOptions` automatically in layer rendering avoiding the need to use prerender events.  That need not apply just to reprojected sources so perhaps a name change is desirable?

mike-000:
waiting for #10463

mike-000:
Carrying the options (renamed to `contextOptions`?) into the main rendering should be straightforward, as noted in my earlier comment.  I had used context options instead of a boolean to keep it as generic as possible, in case someone needed to set other context options for some reason, but it is now seems to get a 100% accurate DEM reprojection some special handling for `imageSmoothingEnabled: false` and the IE equivalent will be needed (which might not be suitable for other context options) as any form of built-in diagonal clipping can distort the color along the edge unless the #10463 brokenDiagonalRendering workaround is used.

ahocevar:
What other context settings would you think of? If this is only about image smoothing, I would not expose more. An `imageSmoothing` option is easy to find and convenient.

If we want users to be able to modify the context, we could think of a way of exposing reprojection rendering hooks at a lower level.

tschaub:
@mike-000, if you ever need to rework a commit, you can commit with the `--amend` flag and then push with the `--force` flag.

You'll likely find a lot of suggestions online to avoid rewriting git history (or force pushing), but on this project, we embrace force pushes as a way to make things easier for reviewers.  A branch with a few clear, logically separated commits can be a lot less intimidating to review than a branch with many commits that may or may not be significant.  If you are working on a shared branch with others, it can require extra communication to make sure force pushes are ok with others (no work should be lost as long as everyone knows how to `git pull --rebase some-remote shared-branch` or equivalent to get the latest from a shared branch).  When you are working individually on a branch, you can force push with impunity.

If you haven't done it before, the workflow would look something like this.
```bash
# assuming your remote is named "origin" and your branch is "my-branch"
# if you `git push origin my-branch` and then you realize you want to modify the last commit
git add path/to/the/changed/file.js # with the --patch option if you like to review your changes before adding
git commit --amend
# edit the commit message in your configured `git config --global core.editor`
git push --force origin my-branch
```

And then a complimentary set of commands if you want to rebase your changes on the latest from OpenLayers:
```bash
# assuming the upstream remote is named "openlayers"
git checkout my-banch
# fix any conflicts (these would be the same conflicts if you had merged, but the fixes will be applied to the appropriate commits)
git push --force origin my-branch
```

If you already know all this, I hope it doesn't come off as patronizing.

ahocevar:
@mike-000: Next time, to avoid merge commits like dfcc649 in a feature branch, please do

instead of merging master into your branch. See https://github.com/openlayers/openlayers/pull/10429#issuecomment-608003854.

PR #10430, opened 2019-12-17
Offset regular shape

Razi91:
While working on QGIS's SLD parser, I had to add option to make an offset for RegularShape (displacement on PointSymbolizer). I could do that with `geometry` option in shape, but I don't think it would be good idea (or maybe it is?)

Anyway, this commit modifies `anchor` on RegularShape, offsetting it to given position. Add optional `offset` option for RegularShape constructor. Tests added, current tests passing (except those with font rendering, but it's system font rendering issue, master doesn't pass either). Doesn't collide with current implementations.

### DIFF
--- a/examples/regularshape.html
+++ b/examples/regularshape.html
@@ -5,7 +5,9 @@ shortdesc: Example of some Regular Shape styles.
 docs: >
   This example shows how several regular shapes
   or symbols (representing `x`, `cross`, `star`,
-  `triangle` and `square`) can be created.
+  `triangle`, `square` and `stacked`) can be created.
+
+  Style `stacked` represents possility to stack multiple shapes with offset
 tags: "vector, symbol, regularshape, style, square, cross, star, triangle, x"
 ---
 <div id="map" class="map"></div>
--- a/examples/regularshape.js
+++ b/examples/regularshape.js
@@ -68,7 +68,7 @@ const styles = {
         points: 4,
         radius: 5,
         angle: Math.PI / 4,
-        offset: [0, 10]
+        displacement: [0, 10]
       })
     }),
     new Style({
--- a/examples/regularshape.js
+++ b/examples/regularshape.js
@@ -59,18 +59,39 @@ const styles = {
       radius2: 0,
       angle: Math.PI / 4
     })
-  })
+  }),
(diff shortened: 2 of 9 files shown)


mike-000:
If I understand this PR correctly the offset has a similar purpose to `offsetX` and `offsetY` in ol/style/Text?  If so should those option names also be used here?

Razi91:
I think it's good idea. I see that Icon already has offset.

Razi91:
In SLD nomenclature it's called `displacement`:
```xml
          <se:PointSymbolizer>
            <se:Graphic>
              <se:Mark>
                <se:WellKnownName>line</se:WellKnownName>
                <se:Fill>
                  <se:SvgParameter name="fill">#0000ff</se:SvgParameter>
                </se:Fill>
                <se:Stroke>
                  <se:SvgParameter name="stroke">#0053d0</se:SvgParameter>
                  <se:SvgParameter name="stroke-width">1</se:SvgParameter>
                </se:Stroke>
              </se:Mark>
              <se:Size>4</se:Size>
              <se:Displacement>
                <se:DisplacementX>0</se:DisplacementX>
                <se:DisplacementY>-5</se:DisplacementY>
              </se:Displacement>
            </se:Graphic>
          </se:PointSymbolizer>
```

ahocevar:
An array named `displacement` instead of `offsetX` and `offsetY` sounds good.

Razi91:
Rebased and renamed `offset` to `displacement`, added test to `ol/style/Icon`. For this class it just displaces final anchor by given displacement.

Razi91:
Could someone review that?

Razi91:
What do you mean? `stacked` is only in examples, to show displaced style. I just noticed that it wasn't updated to `displacement`. I'll fix it when I'm home.

ahocevar:
@Razi91 Sorry for that, I had misread the diff. Everything is fine with the `stacked` style. Waiting for your final fix now Note that I rebased and force-pushed your pull request, so you need to do `git pull --rebase` before you make more changes.

Razi91:
I did that fix, but I messed that `-rebase`. I'm not too much experienced with git and multiple remote repositories

PR #10438, opened 2019-12-20
Avoid accessing the DOM to read the viewport size too often

### LINKED ISSUE #10431: View constraints computation can cause a layout reflow
The following method from the `ol/View` class reads the computed style of the viewport element, which triggers a layout reflow:
https://github.com/openlayers/openlayers/blob/2b45bc054ee2d0e95a8c4b0e98abe42cfb7da482/src/ol/View.js#L729-L745

See: https://gist.github.com/paulirish/5d52fb081b3570c81e3a#getcomputedstyle

This method is called very often now that we use the viewport size to apply view constraints.

A caching mechanism for element dimensions could be used to avoid this, provided this gives a performance/experience improvement.

jahow:
Fixes #10431

This optimization prevents very frequent DOM access for view constraints computation. It should generally improve performance (`getComputedStyle` will trigger a layout reflow) although the gains were pretty subtle in my tests.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1,7 +1,6 @@
 /**
  * @module ol/PluggableMap
  */
-import {getUid} from './util.js';
 import Collection from './Collection.js';
 import CollectionEventType from './CollectionEventType.js';
 import MapBrowserEvent from './MapBrowserEvent.js';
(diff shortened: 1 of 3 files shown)


src/ol/PluggableMap.js
```diff
@@ -1103,6 +1103,9 @@ class PluggableMap extends BaseObject {
     const view = this.getView();
     if (view) {
       this.viewport_.setAttribute('data-view', getUid(view));
```
ahocevar: This line is now obsolete and can be removed. Also, the JSDoc for `ol/View#calculateExtent`'s optional `size` parameter can be changed now:

> If not provided, the size of the map that uses this view will be used.

PR #10439, opened 2019-12-20
WebGL points layer / add support for symbol rotation

### LINKED ISSUE #10381: How to custom special point in webgl point layer
Hello anyone,

I have a request how to custom special point in WebGL Point Layer
In my understand WebGL just support LiteralStyle with some symbol type so I would like to draw special one like as RegularShape

Example for special icon like this
https://www.marinetraffic.com/en/ais/home/centerx:-28.3/centery:17.1/zoom:6

Please correct me if I'm wrong

Thanks!

jahow:
This PR adds support for rotated symbols using the `rotation` property of the layer style object. This property supports expression.

The `rotation` property is consistent with the main style API: it is expressed in radians, and the pivot point is the center of the point (even with an offset applied to the symbol).

A new predefined style was added in the webgl-points-layer example with rotating bars (waiting for link).

Partially solves #10381

### DIFF
--- a/examples/webgl-points-layer.html
+++ b/examples/webgl-points-layer.html
@@ -13,6 +13,8 @@ docs: >
        Note: those will be taken from the attributes provided to the renderer
      * `['var', 'varName']` fetches a value from the style variables, or 0 if undefined
      * `['time']` returns the time in seconds since the creation of the layer
+     * `['zoom']` returns the current zoom level
+     * `['resolution']` returns the current resolution
 
    * Math operators:
      * `['*', value1, value2]` multiplies `value1` by `value2`
@@ -57,7 +59,6 @@ docs: >
      * `['color', red, green, blue, alpha]` creates a `color` value from `number` values; the `alpha` parameter is
        optional; if not specified, it will be set to 1.
        Note: `red`, `green` and `blue` components must be values between 0 and 255; `alpha` between 0 and 1.
-
    Values can either be literals or another operator, as they will be evaluated recursively.
    Literal values can be of the following types:
     * `boolean`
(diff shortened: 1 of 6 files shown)


fredj:
Example: https://2832-4723248-gh.circle-artifacts.com/0/examples/webgl-points-layer.html

comictvn:
`rotation: ['match', ["get", "heading"], 1, 0.8, 0],`
I can't use literal style to get heading information when apply rotation

jahow:
> `rotation: ['match', ["get", "heading"], 1, 0.8, 0],`
> I can't use literal style to get heading information when apply rotation @jahow

What exactly are you trying to achieve? The `match` operator works like a series of if/else blocks, maybe simply fetching the `heading` attribute with `['get', 'heading']` will do it?

comictvn:
my feature format like  this 
`
{ 
  type: "Feature",
  properties: {
    heading: 360, 
    color: "primary", 
    type: "vessel", 
    amount: "2", 
    vessel_id: 16572
  },
  geometry: {
    type: "Point",
    coordinates: [9.59427, 64.1319483333]
  }
}
`
when I use `rotation: ['match', ["get", "heading"], 360, 6.28319, 0]` 
Expected: rotation will get the value is 6.28319 but it always return 0 

I also try `rotation: ['get', 'heading']` but it doesnt' work 
Please correct me if i'm wrong
Thank you
@jahow

jahow:
Rotations are expressed in radians. If you can setup a reproduction case showing the unexpected behaviour, please open an issue with it. Thanks!

hexadeciman:
Hi @jahow I have written a small app to show the issue:

https://stackblitz.com/edit/react-openlayers

If you change the rotation to be 1 in the style symbol ( App.js line 19) the triangle rotates.

But if you change the heading value to 1 in the data.js geoJSONData constant, it doesn't change, the triangle rotation stays at 0.

You can also note that the binding works with size (if you change the size in geoJSONData constant the triangle size changes.

Same issue with image symbol.

Is there a way to bind the value of rotation from the feature properties ?

jahow:
Thanks @hexadeciman for giving a reproduction case, https://github.com/openlayers/openlayers/pull/10758 fixes the issue.

hexadeciman:
I tested it, and I confirm that it works, when can I expect it to be released ?

PR #10452, opened 2019-12-24
Add silent mode to update feature style

ccsldya:
Add silent mode to update feature style;I have below scenario:
I need update thousands of points on the map every 10s, each feature have alone style, I create a new source then added those features, when update their style one by one ,each setStyle method cost 4 ms, totally 4s by 1000 feature,  caused by the "changed" event called too many times ; I need a silent mode, those style will apply when I add the new source to map layer.


jahow:
Hi! Thanks for trying to improve the library. I think your use case does not justify an evolution such as this one; you can probably achieve your goal by using a style function.

PR #10463, opened 2019-12-28
Fix issue with reprojection and double drawing pixels.

pjsg:
This fixes the current triangulation during reprojection so that there is no overlap of pixels, and hence reprojecting translucent layers works as expected. It turns out that the pixels along the edge of the triangle are partially written with variable translucency. The next triangle will exactly bring the color and translucency correct (now that the compositing mode is correct).

The good news about this PR is that the amount of code is actually reduced.

This PR does not solve all the issues -- in particular, layers can be requested to be rendered with infinite extent, and there are still missing corners in the world at some zoom levels.

[This is a work in progress as I'd like to fix the other issues as well.]


pjsg:
The tests don't pass locally (however, the same three test failures also happen on the v6.1.1 tagged item)
```
case ./cases/layer-vector-multigeometry-decluttering/main.js': mismatch 0.0217742919921875
case ./cases/text-style/main.js': mismatch 0.02166748046875
case ./cases/text-style-overlap/main.js': mismatch 0.024383544921875
```

pjsg:
There is one open issue that I can't (currently) figure out. At low zoom levels, a canvas layer in EPSG:3857 doesn't get rendered at all and the render call of that object is called to render a zero size canvas. However, the other issues are much improved.

pjsg:
It turns out that this still draws lines on Windows 8 with Firefox and IE. More work required....

pjsg:
I have filed a bug with Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1606976 which includes a jsfiddle to reproduce the issue without needing to include openlayers!

pjsg:
I have now added a workaround for browsers that fail to hangle diagonal clip regions correctly. The workaround uses a different shape of clip region.

pjsg:
There is still an issue with Chrome, but this turns out to be a Chrome bug. See https://jsfiddle.net/philipjsg/kxmtfc8e/21/ for the example, and compare with Firefox. Chrome defect: https://bugs.chromium.org/p/chromium/issues/detail?id=1041204&can=2

ahocevar:
@pjsg What is the status of this pull request? Ready for review, or more to be done?

pjsg:
This is the code that I have been using on my website for a couple of months with no obvious bad effects. Chrome did fix their bug, so that is good news as well.

As far as I am concerned, it is ready to go...

ahocevar:
@pjsg Would you be able to add a rendering test that verifies the correct visual result, and does not pass before applying your changes? Also, I'm failing to see where the amount of code is reduced by this pull request.

ahocevar:
This is great work and I'd really like to see your improvements in the library @pjsg. Would you be able to address the comments in this pull request and in #10531?

pjsg:
Let me get to this now...

pjsg:
I have added a test that takes a uniform translucent image (actually 1x1 pixel) and reprojects it into a much larger canvas. The test verifies that all the pixels are (essentially) the same. This test fails with the current master, and passes on my branch.

pjsg:
Hmm -- these look the same to me -- which browser are you using on which platform? I was using Chrome 80 on OSX.

ahocevar:
Latest Safari on latest MacOS.

pjsg:
Yes indeed -- I see it too with Safari.

pjsg:
I have reverted that piece of code to the original version (from master). The change that I had made was an optimization and tried to draw less of the image.

PR #10473, opened 2020-01-01
Fix for undefined source in Vector layer

### LINKED ISSUE #10464: Error at VectorLayer.js:333:12after  if (vectorSource.getWrapX() && viewState.projection.canWrapX()
After switching from openlayers v 4.6.4 to v.6.1.1 I encounter the following error:
r is null  and points to the VectorLayer.js:333:12
When the page is loaded the screen stays blank (controls are visible) the error is dispayed in the web console. The map is drawn, when I pan or zoom an the error is repeated with every panning and zooming.

The page has a lot of layers, but even when I disable them all (blank page without background map) the error persists.

mike-000:
Fixes #10464

The documentation implies leaving a source undefined and setting it later is valid but that causes an error in OL6 for Vector layers.  For a vector layer a source left or set null or undefined is equivalent to an empty source so treat it as such to prevent errors in `prepareFrame`

### DIFF
--- a/src/ol/renderer/canvas/VectorLayer.js
+++ b/src/ol/renderer/canvas/VectorLayer.js
@@ -11,16 +11,6 @@ import CanvasLayerRenderer from './Layer.js';
 import {defaultOrder as defaultRenderOrder, getTolerance as getRenderTolerance, getSquaredTolerance as getSquaredRenderTolerance, renderFeature} from '../vector.js';
 import {toString as transformToString, makeScale, makeInverse, apply} from '../../transform.js';
 import {createHitDetectionImageData, hitDetect} from '../../render/canvas/hitdetect.js';
-import VectorSource from '../../source/Vector.js';
-
-
-/**
- * @const
- * @type {VectorSource}
- * @private
- */
-const emptySource = new VectorSource();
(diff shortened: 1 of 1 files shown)


src/ol/renderer/canvas/VectorLayer.js
```diff
@@ -328,7 +338,7 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
    */
   prepareFrame(frameState) {
     const vectorLayer = this.getLayer();
-    const vectorSource = vectorLayer.getSource();
+    const vectorSource = vectorLayer.getSource() || emptySource;
```
tschaub: What happens if we return early here instead of proceeding with an empty source?

mike-000: Returning false seems to have no adverse effect (it simply stops the render).  I had been concerned that if the source was cleared using `setSource(null)` the renderer's properties would not get reset but without any rendering that should not be a problem, and it also avoids a similar error in the renderFrame function if interacting when `setSource(null)` is called.

tschaub:
Thanks, @mike-000.

PR #10490, opened 2020-01-07
Select style multiple select interactions removed

### LINKED ISSUE #10486: Select style is not removed when using two Select interactions
**Describe the bug**
We upgraded our application from OL 4.6.5 to 6.1.1. This was surprisingly smooth. However, we notice that selected features get the selection style when they're selected, but they don't revert to the default style when they're deselected.

**To Reproduce**
1. Add a `Select` with a custom style to a map. 
2. Remove the `Select`
3. Add another `Select` with a custom style to a map

**Expected behavior**
When a deselecting a feature (with shift-click if multiple selection is allowed or simple click if not), the feature should revert to its initial style.

**Code observations**
We were not sure if the bug was due our usage of OL or in OL itself, so we had a hard look with a debugger.

What happens is that when a feature is selected, first its original style (if any) is stored in `featureStyleAssociation_`, then the select style is set. However, this happens twice because the `addFeature_` function is called twice. As a result, the second time the select style (and not the default style) is copied to the `featureStyleAssociation_` map. 

So, when the feature is deselected, the style is "reverted" to the selection style, not the default style.

The `addFeature_` is called twice because the added event listener (`addEventListener` in the constructor or `Select`) is not unregistered.

We do call `removeInteraction` in our code and we do observe `setMap` being called with a `null` argument, but that does not appear to remove the `addFeature_` event listener.

bepremeg:
This fixes #10486 by removing the event listeners when an
interaction is removed from a map.

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -169,6 +169,16 @@ class Select extends Interaction {
 
     /**
      * @private
+     */
+    this.boundAddFeature_ = this.addFeature_.bind(this);
+
+    /**
+     * @private
+     */
+    this.boundRemoveFeature_ = this.removeFeature_.bind(this);
+
+    /**
+     * @private
      * @type {import("../events/condition.js").Condition}
      */
     this.condition_ = options.condition ? options.condition : singleClick;
@@ -250,9 +260,8 @@ class Select extends Interaction {
      */
     this.featureLayerAssociation_ = {};
 
-    const features = this.getFeatures();
-    features.addEventListener(CollectionEventType.ADD, this.addFeature_.bind(this));
-    features.addEventListener(CollectionEventType.REMOVE, this.removeFeature_.bind(this));
+    this.features_.addEventListener(CollectionEventType.ADD, this.boundAddFeature_);
+    this.features_.addEventListener(CollectionEventType.REMOVE, this.boundRemoveFeature_);
   }
 
   /**
@@ -322,6 +331,10 @@ class Select extends Interaction {
     super.setMap(map);
     if (map && this.style_) {
       this.features_.forEach(this.applySelectedStyle_.bind(this));
+    }
+    if (!map) {
+      this.features_.removeEventListener(CollectionEventType.ADD, this.boundAddFeature_);
+      this.features_.removeEventListener(CollectionEventType.REMOVE, this.boundRemoveFeature_);
     }
   }
 
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -333,8 +333,7 @@ class Select extends Interaction {
       if (this.style_) {
         this.features_.forEach(this.applySelectedStyle_.bind(this));
       }
-    }
-    if (!map) {
+    } else {
       this.features_.removeEventListener(CollectionEventType.ADD, this.boundAddFeature_);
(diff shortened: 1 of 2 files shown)


jahow:
Thanks for your contribution! I think a new test would be in order, although your code seems like it should fix the problem. Please let me know if you need indications.

bepremeg:
I had a look at the existing tests. I'll probably find a way how to test this. But I'm in doubt what the best place is. a top-level `describe('replace Select')`, something under `describe('selecting a polygon')`, or another place altogether?

jahow:
I'd say a top level `describe('clear event listeners on interaction removal')` would make sense, so at the same level as this:
https://github.com/openlayers/openlayers/blob/6e8f3407ff29bd9525f83c25596833712d9c26cc/test/spec/ol/interaction/select.test.js#L122

Thanks

bepremeg:
All right. I had some trouble finding the right way to test this. I couldn't get my test to fail when I reverted my fix :-(
Turns out that the issue also requires an external feature collection. Obvious in hindsight.
This also requires that the activation of the event listeners only happens when the interaction is added to the map. Otherwise the `addFeature` is invoked regardless of the interaction being associated with a map or not.
This change also makes the event handler association more symmetrical.

bepremeg:
@jahow Could you have a look at the test result? There is one failing test  for `TileQueue` on Circleci, but it works on my local test environment and I don't see how my change could impact that test.

bepremeg:
I now have exactly the same test code as where I started but the build succeeds now.
As far as I'm concerned, this is patch ready.

mike-000:
> @jahow Could you have a look at the test result? There is one failing test for `TileQueue` on Circleci, but it works on my local test environment and I don't see how my change could impact that test.

Unrelated failed tests seem to be quite common, for example see https://github.com/openlayers/openlayers/issues/7437#issuecomment-532164316  Any change even an edit to a comment will rerun the tests as will squashing and force pushing which also tidies the commit history.

bepremeg:
@jahow Indeed. The tests are passing now, even if the code is the same. I made a few commits that cancel each other.

src/ol/interaction/Select.js
```diff
@@ -322,8 +328,16 @@ class Select extends Interaction {
       this.features_.forEach(this.removeSelectedStyle_.bind(this));
     }
     super.setMap(map);
-    if (map && this.style_) {
-      this.features_.forEach(this.giveSelectedStyle_.bind(this));
+    if (map) {
+      this.features_.addEventListener(CollectionEventType.ADD, this.boundAddFeature_);
+      this.features_.addEventListener(CollectionEventType.REMOVE, this.boundRemoveFeature_);
+
```
michalzielanski:
```javascript
if (!currentMap) {
    // ...
}
```
is needed.
For such a code:
```javascript
import Map from 'ol/Map';
import Select from 'ol/interaction/Select';

var map1 = new Map(/* ... */);
var map2 = new Map(/* ... */);

var selectInteraction = new Select();
// Attach interaction to a map1.
selectInteraction.setMap(map1);
// Change map for interaction to map2.
selectInteraction.setMap(map2);
```
there will be two listeners per event.

ahocevar:
Rebased to fix merge conflicts.

PR #10495, opened 2020-01-08
typo in view.fit

danielklim:
view.fit had a typo where the variable `geometry` was used instead of `geometryOrExtent`.


mike-000:
@DaniEll-AT  Extents have already been identified by `if (Array.isArray(geometryOrExtent)) {` so `geometryOrExtent` in the else block will always be a geometry.

Issue #6948, opened 2017-06-23
Overlay - Element events (click) pass through to Map


ralphstodomingo:
I have a popup in React made as an Overlay. The problem is, events meant for this React component (clicking a button) pass through to the map, which means sometimes features get (de)selected because the buttons happen to be over them.

Setting the overlay with `stopEvent: true` causes the React component to not receive any click events at all, while `false` causes the map to be also clicked as well.

Is this intended behavior? Does 'stopEvent' prevent the overlay element from having any events?

bartvde:
yeah I've dealt with this as well, there are multiple threads on it such as:

https://stackoverflow.com/questions/35454014/clicks-on-reactjs-components-not-firing-on-an-openlayers-overlay

I've also used the approach of setting interactions to not active when the popup is open such as here: 

https://github.com/boundlessgeo/sdk/blob/3ee2e70c6c35d3390be4804058ee9c0521da7806/src/components/EditPopup.js#L223:L228

See also: https://github.com/openlayers/openlayers/issues/4400

tschaub:
I think we'd agree that setting `stopEvent: true` should only cause the map to ignore events that target the overlay.  We should be able to handle this in the library, so I think it is a bug (instead of intended behavior) that other listeners can't respond to events on the overlay.

ralphstodomingo:
For my particular use-case my workaround involved setting `stopEvent: true` and doing something like this inside my React component (which I took from @bartvde's answer in that linked SO question):

```
if (this.closeButtonRef.button.button.onclick === null) {
   this.closeButtonRef.button.button.onclick = this.setFeaturePopupVisibleAsFalse;
}
```
 
This is because I was using `material-ui` which comes with its own buttons. A normal button will likely involve `this.buttonElement.onclick = this.testFunc` or something similar.
I also had to bind this in `componentWillReceiveProps` instead of the constructor or `componentDidMount` since the element in question relied on a prop triggering before appearing.
This looks hacky at best, and I had to do this for every action button on that popup. Still, hoping this helps anyone who stumbles with it, while there isn't a fix yet.

bartvde:
Yep would be great if we could handle this in the library instead, I agree those workaround are wonky ;-)

marcjansen:
We have deployed such wonky workarounds as well, I would also favor a library solution.

bassarisse:
Any news on this? Looking for a better solution, at least without having to patch each component =/

hwbllmnn:
[...]

PR #10502, opened 2020-01-10
Stop events that originate with a removed target

walkermatt:
As discussed in https://github.com/openlayers/openlayers/issues/6948#issuecomment-565375694

The check to see if the target is within the "page" uses the viewport as
the MapBrowserEventHandler instance adds it's listeners to the viewport.

Using Node.contains appears to have a slight performance benefit
over manually walking the DOM.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -934,7 +934,7 @@ class PluggableMap extends BaseObject {
     }
     const target = /** @type {Node} */ (mapBrowserEvent.originalEvent.target);
     if (!mapBrowserEvent.dragging) {
-      if (this.overlayContainerStopEvent_.contains(target) || !this.viewport_.contains(target)) {
+      if (this.overlayContainerStopEvent_.contains(target) || !document.body.contains(target)) {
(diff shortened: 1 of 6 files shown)


walkermatt:
It looks as though a number of tests fail due to the simulated events not setting a target. I'll investigate and update.

walkermatt:
@ahocevar sorry for being a bit premature with this one! I think it's ready for review now. In summary this PR:

* Stops events that originate with a removed target - this resolves #6948
* The check to see if the target is within the "page" tests to see if the `target` is contained within `document.body` instead of the map viewport. This is due to some events originating outside the map viewport such as keyboard events which originate with the element specified by `keyboardEventTarget` which could be `document.body`. [`Node.contains()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) is inclusive so a `target` of `document.body` will return true.
* Several of the tests simulate pointer events and did not previously specify a `target` property. I've updated the event simulation to mock a `PointerEvent` with a plain object that includes a `target` property and the other essential properties used internally.

Issue #10503, opened 2020-01-10
Image quality degrades at rotation with 90,180 or 270 degrees


danzaharescu:
**Describe the bug**
If the View has the rotation set to Math.PI / 2 or Math.PI or 3 * Math.PI / 2, then apparently the quality of the image suffers (depending on image and/or image resolution, either the rendering becomes blurred - as in the attached picture, either it becomes aliased - text and lines broken) . Dragging the edge of the image (for example, the separator in the codesandbox) sometimes temporarily fixes the image, but additional dragging make the image again degraded and so on. I must mention, the quality is already much better than older OpenLayers versions (example, 4.*). The degradation appearing during the rotation leads to the image having the same rendering quality as in 4.*

**To Reproduce**
I believe the following screenshot will make it clear (I have used a rotation of 2 * Math.PI - is ALMOST EQUIVALENT TO 0, but not exactly, apparently, just to better view the differences):
![rotation](https://user-images.githubusercontent.com/17156861/72151867-105dcf00-33b2-11ea-9dc7-7154c8ecfb4f.png)

Please take into account that in other cases the image may become 'aliased' (text and lines broken)

3. See error

**Expected behavior**
The same quality image should be observable at any rotation.
As a side note, I am thinking of the problem being in the area that Math.PI is not exactly the real PI (that cannot be represented). Therefore, I am thinking that maybe supporting rotations in degrees could be helpful.

mike-000:
Possibly browser specific, I cannot see any difference between 0 and 2*Math.PI.  However I noticed that when rotation is set to 2*Math.PI clicking the reset north button has no effect unless `duration: 0` is set in the Rotate control options.

mike-000:
I can see the effect when resizing the map.  It is because **any** non-zero rotation uses a canvas larger the the viewport (a square with sides sqrt(2) times the largest dimension of the viewport) which is centered on the center of the viewport and rotated as necessary.  Making a special case for non-oblique rotations would need a lot of extra extra code for relatively rare scenarios.

danzaharescu:
[...]

PR #10504, opened 2020-01-10
Fix for reset north when rotation is 360 degrees

mike-000:
Fixes the Rotate control glitch observed in https://github.com/openlayers/openlayers/issues/10503#issuecomment-573031268
Animated reset doesn't work and isn't needed if rotation is 360 degrees (or a multiple of 360)

### DIFF
--- a/src/ol/control/Rotate.js
+++ b/src/ol/control/Rotate.js
@@ -131,8 +131,9 @@ class Rotate extends Control {
       // upon it
       return;
     }
-    if (view.getRotation() !== undefined) {
-      if (this.duration_ > 0) {
+    const rotation = view.getRotation();
+    if (rotation !== undefined) {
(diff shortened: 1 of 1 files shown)


danzaharescu:
Hello, I see you already implemented a fix. 
Just want to re-iterate (in case maybe not clear yet), problem occurs not only for 360 rotation, but also, for 90,180,270. Also, I have reproduced the issue when rotating with and without animation... 
Thanks.

mike-000:
@danzaharescu This only fixes the resetting of the Rotate control, it doesn't affect the rendering (I have not been able to reproduce your problem).

danzaharescu:
The problem is not always reproducible. It seems that on edge is not reproducible, or much harder. But on Chrome, if you have the image rotated and drag the separator (in the codesandbox) between the code and the rendering area (maybe you must try a few more times), I am almost sure it is reproducible...the image may also recover after additional drags.

PR #10509, opened 2020-01-11
Ensure that effective min resolution is never higher than max resolution

### LINKED ISSUE #10508: View glitches when conflicting constraint are set
**Describe the bug**
When an `extent` and `minResolution` are set and they can't both be satisfied the `View` misbehaves.

**To Reproduce**
1. Go to https://codesandbox.io/s/laughing-keldysh-x241y
2. Zoom / pan in the maps.
3. The view jumps / locks to different resolutions.

**Expected behavior**
If one of the constraints can't be satisfied prioritize one constraint.

MoonE:
Fix #10508

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1051,7 +1051,7 @@ class View extends BaseObject {
   getResolutionForValueFunction(opt_power) {
     const power = opt_power || 2;
     const maxResolution = this.getConstrainedResolution(this.maxResolution_);
-    const minResolution = this.minResolution_;
+    const minResolution = Math.min(this.minResolution_, maxResolution);
     const max = Math.log(maxResolution / minResolution) / Math.log(power);
     return (
       /**
@@ -1084,7 +1084,7 @@ class View extends BaseObject {
   getValueForResolutionFunction(opt_power) {
     const logPower = Math.log(opt_power || 2);
     const maxResolution = this.getConstrainedResolution(this.maxResolution_);
-    const minResolution = this.minResolution_;
+    const minResolution = Math.min(this.minResolution_, maxResolution);
     const max = Math.log(maxResolution / minResolution) / logPower;
     return (
       /**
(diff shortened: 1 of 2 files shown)


MoonE:
I agree that it is kind of ugly to sprinkle this check everywhere but I'm not sure your suggestion would work, as the `View`'s `multiWorld`/`extent` options resolve to different `maxResolution` depending on the map size and that can be contolled by the user by resizing the browser window.

If you resize the map in the provided codesandbox the first map's `View` works fine once the map is small enough and the world does fit within the resolution constraint.

ahocevar:
@MoonE Thanks for the explanation. I understand the problem now. What about clamping the resolutoin to the range from `this.minResolution_` to `this.maxResolution_` in `View#getConstrainedResolution`?

To verify the result, a unit test would also be great.

MoonE:
The way I implemented it here is that  `extent` / `multiWorld` is the 'winning' constraint if there is a conflict, because I think it is more important if set. With your suggestion there would be some wrap around / parts outside the configured extent visible on conflict, which could be seen as a bug itself.

MoonE:
I implemented your suggestion here (with added example from codesandbox), though it is not enough to fix the problem:
https://github.com/MoonE/openlayers/tree/view-zoom-glitch2

ahocevar:
Thanks for that branch @MoonE. After #10865, the situation is different now. Maybe this allows for a more straightforward fix?

tschaub:
It might also be worth looking into #10892 when addressing this.

tschaub:
#10862 introduced a number of conflicts with these changes.  I've pushed a rebased, conflict-free version of your commits in 64a2f3681f32c21a2bda78b46ab9e2b1d749c7e9.

If you want to use this commit, you can do this:
```
git fetch openlayers
git reset --hard 64a2f3681f32c21a2bda78b46ab9e2b1d749c7e9
git push --force origin view-zoom-glitch
```

MoonE:
Thanks @tschaub

Issue #10460, opened 2019-12-28
Dead links in the apidocs


MoonE:
**Describe the bug**
There are multiple dead links in the api docs. See also #10213

**To Reproduce**
1. Run the following script after configuring the `src` and `www` variables. 
It recursively checks for dead links in the api docs
```bash
#!/bin/bash
dir=build/apidoc
src=~/openlayers/$dir
www=http://localhost:8080/$dir

for link in $(\
    wget --spider --no-directories --no-verbose --recursive --no-parent $www \
        |& grep --before-context=1 -- "-- broken link!!!" \
        | grep --only-matching --perl-regexp "(?<=/${dir}/).*\.html" \
        | sort \
); do
    printf "\n${link#:} 404, linked from:\n"
    grep --recursive --files-with-matches "${link}" "${src}" \
        | grep --only-matching --perl-regexp "(?<=${src}/).*" \
        | sort \
        | sed 's/^/  /'
done
```
2. current output 0519e908d55fbeb95408a06644930adfe2256cda:
```
module-ol_Image-ImageWrapper.html 404, linked from:
  module-ol_Image.html
  module-ol_source_Image.ImageSourceEvent.html

module-ol_ObjectEventType.html 404, linked from:
  module-ol_Geolocation.html
  module-ol_Map.html
  module-ol_Overlay.html
  module-ol_View.html
  module-ol_layer_Base.html

module-ol_TileCache-TileCache.html 404, linked from:
  module-ol_source_OGCVectorTile-OGCVectorTile.html
  module-ol_source_VectorTile-VectorTile.html

module-ol_centerconstraint.html 404, linked from:
  module-ol_View.html

module-ol_expr_expression.html 404, linked from:
  module-ol_style_flat.html

module-ol_resolution.html 404, linked from:
  module-ol_Image.html

module-ol_resolutionconstraint.html 404, linked from:
  module-ol_View.html

module-ol_rotationconstraint.html 404, linked from:
  module-ol_View.html

module-ol_style_literal.html 404, linked from:
  module-ol_style_expressions.html
```
**Expected behavior**
There should be no dead links in the api docs.

PR #10512, opened 2020-01-11
Fix two missing apidoc links

MoonE:
Add IconAnchorUnits and IconOrigin to the api.

See #10460


ahocevar:
enums should appear in the API docs without `@api` annotation. So there must be an issue somewhere else.

Issue #10506, opened 2020-01-10
Flag constrainCenterOnly seems to be missing


fgubler:
In the upgrade notes to version 6 it is written that:

> The view extent option now applies to the whole viewport
Previously, this options only constrained the view center. This behaviour can still be obtained by specifying constrainCenterOnly in the view options.

However, I cannot find a place to set the constrainCenterOnly flag: It is not part of the ViewOptions interface and according to my webstorm-search, the only place it is ever mentioned in your code-base are the aforementioned upgrade notes.

Am I missing something? I would like to get the previous behavior again...

fgubler:
Hmm it might only be a typo in the upgrade notes: a flag of the name "constrainOnlyCenter" instead of "constrainCenterOnly" does seem to exist. Will investigate further

fredj:
Thanks for reporting this @fgubler ! Would you be able to provide a pull request to change `constrainCenterOnly` to `constrainOnlyCenter` in `changelog/upgrade-notes.md` ?

fgubler:
Sure: will provide it shortly.

PR #10513, opened 2020-01-13
fixed flag-name in upgrade notes

fgubler:
fixed flag-name in upgrade notes: "constrainOnlyCenter" instead of "constrainCenterOnly"


fredj:
fixes #10506

fredj:
Thanks @fgubler !

PR #10525, opened 2020-01-13
Fix the z-index of the Overlay-Container

fgubler:
Setting the z-index of 'ol-overlaycontainer-stopevent' to '0' makes the following use-case impossible:

 - I have a map in the background
 - I have some controls for user-interaction in front of the map
 - I want popups / overlays from the map to be in front of the controls

With z-index '0' on 'ol-overlaycontainer-stopevent', the popups / overlays are always behind the controls, even if I give them a higher z-index.
With z-index 'auto' on 'ol-overlaycontainer-stopevent', I can control the position of popups by setting their z-index to an appropriate value.

Before revision 898c349, the z-index was (implicitly) set to 'auto'. I expect the problem discribed above to be an unintended side-effect of that change.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -253,7 +253,7 @@ class PluggableMap extends BaseObject {
      */
     this.overlayContainer_ = document.createElement('div');
     this.overlayContainer_.style.position = 'absolute';
-    this.overlayContainer_.style.zIndex = '0';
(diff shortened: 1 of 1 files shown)


joserwan:
@fgubler I had the same issue, but I forgot to include `ol.css` as mentionned here : 
 #10663

Issue #10500, opened 2020-01-10
MultiPolygons don't display on map when loading mapbox vector tile with MVT format which created with ol.Feature rather than RenderFeature


heavis:
I use MVT with ol.Feature to load mapbox vector tile including some polygon or multiploygon geometries, but some multiploygons failed render on map.
![有问题](https://user-images.githubusercontent.com/8612147/72131064-2f5e6000-33b6-11ea-8b89-b4ced20c93aa.png)
Above picture is the render result with ol.Feature class, some multiplygon geometries don't display.
![没问题](https://user-images.githubusercontent.com/8612147/72131286-bdd2e180-33b6-11ea-848c-88c2c9cf471b.png)

Above picture is the render result with ol.RenderFeature class, all geometries display properly.

ol version is 5.1.3

ahocevar:
Can you please try the latest 6.x version and see if the issue is still present?

heavis:
> Can you please try the latest 6.x version and see if the issue is still present?

I have tried the lastest version 6.1.1, but this problem still exists.

PR #10526, opened 2020-01-13
Update MVT.js

heavis:
descriptions of slice method of Array: 
@param end The end of the specified portion of the array. This is exclusive of the element at the index 'end'. 
initial value of prevEndIndex and i both are 0 so that slice(0, 0)  return a empty array. it lead to a new issue that when I load a multiplygon geometry, endss array's first item is [] but correct value should be the last coordinate index of first polygon of multiplygon. the description of issue: https://github.com/openlayers/openlayers/issues/10500

### DIFF
--- a/src/ol/format/MVT.js
+++ b/src/ol/format/MVT.js
@@ -201,8 +201,8 @@ class MVT extends FeatureFormat {
         for (let i = 0, ii = ends.length; i < ii; ++i) {
           const end = ends[i];
           if (!linearRingIsClockwise(flatCoordinates, offset, end, 2)) {
-            endss.push(ends.slice(prevEndIndex, i));
-            prevEndIndex = i;
+            endss.push(ends.slice(prevEndIndex, i + 1));
+            prevEndIndex = i + 1;
           }
           offset = end;
         }


ahocevar:
@heavis The code path you changed in this pull request is not entered when `featureClass` is set to `RenderFeature`. So I am not sure if this change fixes your issue.

If the situation was the other way around (i.e. the code works for `Feature` features, but not for `RenderFeature`, then I would suspect your tiles to have geometries with the wrong winding order.

That said, if you still think this fix is correct, please add a test case.

heavis:
> @heavis The code path you changed in this pull request is not entered when `featureClass` is set to `RenderFeature`. So I am not sure if this change fixes your issue.
> 
> If the situation was the other way around (i.e. the code works for `Feature` features, but not for `RenderFeature`, then I would suspect your tiles to have geometries with the wrong winding order.
> 
> That said, if you still think this fix is correct, please add a test case.

I am sure that my data format is right because it can properly display when I use mapbox with same data.
I need interaction effect with geometry so I have set Feature type rather than RenderFeature and modified codes will be executed under Feature class.

heavis:
> I was also going to suggest that we need some tests here to clearly understand what is broken. It is kind of scary that this change doesn't break any existing tests (but I haven't looked closely).

[z3_x5_y2.pbf.zip](https://github.com/openlayers/openlayers/files/4069193/z3_x5_y2.pbf.zip)
Attach file is a tile and z,x,y are 3, 5, 2 respectively. you can take a test with this pbf file.
Original codes can't display geometry but modified codes can display properly.

PR #10527, opened 2020-01-13
Remove label cache, render text directly to target canvas

### LINKED ISSUE #10414: Vector-Tiles: Font becomes blurred after rotating the map
**Describe the bug**
As of Openlayers 6.x, the font becomes blurred when the map is rotated.

**To Reproduce**
The Bug can be reproduced with the official examples (see https://openlayers.org/en/latest/examples/mapbox-vector-tiles.html)

1. Go to https://openlayers.org/en/latest/examples/mapbox-vector-tiles.html
2. Rotate the map
3. The country names gets blurry

**Expected behavior**
The font remains sharp.

ahocevar:
This pull request gets rid of the label cache. Labels are now rendered directly to the target canvas.

This is something we moved away from a while ago, before we had decluttering. Now, with decluttering, we do not have to render so many labels any more, and the time spent in `fillText` and `strokeText` is manageable again.

This pull request brings the following advantages:
* Previously, `fillText` and `strokeText` was called for all labels, even when they were never rendered due to decluttering. They were also cached. Now, `fillText` and `strokeText` is only called for labels that are actually used.
* Fixes #10414


src/ol/render/canvas/Executor.js
```diff
@@ -270,11 +288,11 @@ class Executor {
    * @param {Array<*>} fillInstruction Fill instruction.
    * @param {Array<*>} strokeInstruction Stroke instruction.
    */
-  replayImage_(
+  replayImageOrLabel_(
```
tschaub: Oh.  Perhaps because of this.

ahocevar: Yes indeed. A clear separation of the two code paths might make sense, but that would be a much more verbose change.

PR #10531, opened 2020-01-14
Dynamically chose the number of subdivisions based on the size of the Image to reproject

### LINKED ISSUE #10511: Reprojecting a Tile and an Image should use the same size of triangles
**Describe the bug**
The number of triangles that a Tile and an Image can be split into is defined by the Triangulation class. This is *independent* of the size of the Image (or Tile). 

**To Reproduce**
Look at the code.

**Expected behavior**
I would expect that the number of triangles would be related to the area of the screen covered rather than predetermined. If the Tiles are standard 256x256 tiles and the Image is 1024x1024 then they both have max_subdivision set to 10. I think that the image ought to be using max_subdivision = 14 so that the size of the triangles are comparable. This will make reprojected overlays look better on top of reprojected tiles.

pjsg:
This fixes #10511 by making the number of subdivisions be large when the object being reprojected is larger. It does this by figuring out the area of the object to be reprojected and seeing how much larger it is than one tile. This is then used to compute the number of subdivisions.

### DIFF
--- a/src/ol/reproj/Triangulation.js
+++ b/src/ol/reproj/Triangulation.js
@@ -49,9 +49,9 @@ class Triangulation {
    * @param {import("../extent.js").Extent} targetExtent Target extent to triangulate.
    * @param {import("../extent.js").Extent} maxSourceExtent Maximal source extent that can be used.
    * @param {number} errorThreshold Acceptable error (in source units).
-   * @param {number} destinationResolution The (optional) resolution of the destination.
+   * @param {?number} opt_destinationResolution The (optional) resolution of the destination.
    */
-  constructor(sourceProj, targetProj, targetExtent, maxSourceExtent, errorThreshold, destinationResolution) {
+  constructor(sourceProj, targetProj, targetExtent, maxSourceExtent, errorThreshold, opt_destinationResolution) {
 
     /**
      * @type {import("../proj/Projection.js").default}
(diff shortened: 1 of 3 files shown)


ahocevar:
@petrsloup, could you take a look at this please?

src/ol/reproj/Triangulation.js
```diff
@@ -138,11 +139,16 @@ class Triangulation {
     const sourceBottomRight = this.transformInv_(destinationBottomRight);
     const sourceBottomLeft = this.transformInv_(destinationBottomLeft);
 
+    const maxSubdivision = MAX_SUBDIVISION + (opt_destinationResolution ?
+      Math.max(0, Math.ceil(Math.log2(getArea(targetExtent) /
+        (opt_destinationResolution * opt_destinationResolution * 256 * 256))))
```
ahocevar: Why 256?

petrsloup: I guess maybe this calculation assumes the `MAX_SUBDIVISION` value was estimated for the 256x256 tiles, so this calculates how much "larger" this extent is than that?

Anyway, it would be great to describe this calculation in more detail (or maybe separate into more variables with more descriptive names?)..

pjsg:
I added a comment to explain the 256.

PR #10534, opened 2020-01-15
[GeoJSON] Read projection from CRS type EPSG

wussup:
CRS also may have type 'EPSG' and then the code is reading from property 'code' in object 'properties'

### DIFF
--- a/src/ol/format/GeoJSON.js
+++ b/src/ol/format/GeoJSON.js
@@ -162,6 +162,8 @@ class GeoJSON extends JSONFeature {
     if (crs) {
       if (crs['type'] == 'name') {
         projection = getProjection(crs['properties']['name']);
+      } else if (crs['type'] === 'EPSG') {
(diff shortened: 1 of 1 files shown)


src/ol/format/GeoJSON.js
```diff
@@ -162,6 +162,8 @@ class GeoJSON extends JSONFeature {
     if (crs) {
       if (crs['type'] == 'name') {
         projection = getProjection(crs['properties']['name']);
+      } else if (crs['type'] === 'EPSG') {
```
tschaub: This syntax was removed in draft 6 of the community standard 11 years ago.  I’m curious where you are seeing it used.

wussup: > This syntax was removed in draft 6 of the community standard 11 years ago. I’m curious where you are seeing it used.

Unfortunately, it still exists in few of Poland's governments WFS's :|

tschaub:
Thanks for the contribution @wussup.

PR #10537, opened 2020-01-17
Correct resolution used for scale bar and add dpi option

### LINKED ISSUE #10529: Scale line shows incorrect sizes.
The scale line control (ol/control/ScaleLine), assumes a pixel size of 0.28mm.
This creates very large errors in the scale.
On mobile devices we tested we found the average error being more that 50%.

Passing the screen density as an option when creating the control should give an easy solution.

In addition the function converts mm to inch (X25.4) and then inch to meters(X39.37), since these are only used internally it seems redundant to introduce these errors instead of converting mm to meters directly.

  getScaleForResolution() {
    const resolution = this.getMap().getView().getResolution();
    const dpi = **25.4** / 0.28;
    const mpu = this.viewState_.projection.getMetersPerUnit();
    const inchesPerMeter = **39.37**;
    return parseFloat(resolution.toString()) * mpu * inchesPerMeter * dpi;
  }

mike-000:
Fixes #10529

Correct the resolution used for scale. Add `dpi `option for output devices such as printers.
Also add `setDpi` and `setMinWidth` methods

Add a Print to scale example

### DIFF
--- a/examples/print-to-scale.js
+++ b/examples/print-to-scale.js
@@ -8,8 +8,6 @@ import {register} from '../src/ol/proj/proj4.js';
 import WMTS, {optionsFromCapabilities} from '../src/ol/source/WMTS.js';
 import proj4 from 'proj4';
 
-import {toJpeg} from 'html-to-image';
-
 
 proj4.defs('EPSG:27700', '+proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 ' +
     '+x_0=400000 +y_0=-100000 +ellps=airy ' +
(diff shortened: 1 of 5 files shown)


mike-000:
The purpose of the dpi option was to produce a scale bar suitable for printing, so the example wouldn't relate to the change if the map was printed without it.  Also the terms of many sources require an attribution to be visible on screenshots and printed output.  To demonstrate that it's either going to need a browser dependent export (it could use a cdn library to avoid a project dependency) or the controls would need a considerable rework to use canvas elements - and where attributions require clickable links or images that isn't an simple option.

examples/print-to-scale.js
```diff
@@ -0,0 +1,130 @@
...
+      map.updateSize();
+      map.getView().setResolution(viewResolution);
+      exportButton.disabled = false;
+      document.body.style.cursor = 'auto';
+    });
+  });
+
+  // Set print size
+  scaleLine.setDpi(resolution);
+  scaleLine.setMinWidth(resolution * 4 / 2.54); // 4cm
```
ahocevar, line 124: Wouldn't it make sense to adjust widths automatically, so their units change from "pixels" to "pixels at screen resolution (25.4/28 dpi assumed)"? Then users would not have to worry about changing the minWidth when changing the dpi, and there would not be a need to add a `setMinWidth` method.

PR #10544, opened 2020-01-18
Apidoc add experimental tag

MoonE:
This adds a 'experimental' notice to the api documentation similar to the one that is shown for examples using experimental features.
It is shown for module pages when the module is tagged with `@experimental` and also for class pages belonging to such a module.

Add the `ol/layer/WebGLPoints`, `ol/style/LiteralStyle.js` and `ol/style/expressions.js` as experimental api and thus fixes one dead link from the `WebGLPoints` module as seen in #10460

### DIFF
--- a/config/jsdoc/api/conf.json
+++ b/config/jsdoc/api/conf.json
@@ -20,7 +20,8 @@
     "config/jsdoc/api/plugins/inheritdoc",
     "config/jsdoc/api/plugins/events",
     "config/jsdoc/api/plugins/observable",
-    "config/jsdoc/api/plugins/api"
+    "config/jsdoc/api/plugins/api",
+    "config/jsdoc/api/plugins/experimental"
   ],
   "typescript": {
     "moduleRoot": "src"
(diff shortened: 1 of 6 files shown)


jahow:
Thanks! This is looking great. Originally the agreement was that anything with the `@api` tag was part of the official API, hence why the webgl modules were left out. @tschaub might have an opinion on this but IMO this is a great solution.

Also the `useGeographic` API is experimental too (an easy way to find out is to look for "experimental: true" in the examples folder) so this should be tagged as well.

Akoneto9:
Good one

tschaub:
We handled this previously by supplying a value with the `@api` annotation (e.g. `@api stable` or `@api experimental`).  The benefit of the old approach is that we could add an experimental function to the API without marking the whole module as experimental.  With the approach in this branch, it doesn't look like we could do that.  So we couldn't mark `useGeographic()` from the `ol/proj.js` module as experimental, for example.

tschaub:
Thanks for your work on this, @MoonE.  I'm going to close it for now since I don't think we want to classify "experimental" things at the module level.

We're working on getting a large, formatting only, conflict-prone change in (see #10862).  After that is in, we can revisit the idea of more clearly identifying non-stable parts of the API in our docs.

PR #10558, opened 2020-01-20
Avoid unnecessary cloning of text style in KML

mike-000:
Although #10545 required cloning an ol/style/Style object that action will also clone its text style, so rearranging the code will avoid the need for a separate cloning of that.


mike-000:
Change is now included in #10580

Issue #10539, opened 2020-01-17
Zoomify example not working


mike-000:
The tile source has been unresponsive every time I have tried this in recent weeks

crubier:
The server it was using has stop serving the files

Solved in https://github.com/openlayers/openlayers/pull/10562  by switching to another server

ahocevar:
Fixed with #10562.

PR #10562, opened 2020-01-21
Zoomify: Separate the service pixel ratio and the device pixel ratio

crubier:
The long-standing issue https://github.com/openlayers/openlayers/issues/7795 was partially solved by https://github.com/openlayers/openlayers/pull/9489 by myself.

However, as @jameshoward mentioned (https://github.com/openlayers/openlayers/pull/9489#issuecomment-518706270) there was still a bug in the case where there are several tile groups, which was not the case of the example dataset.

The example dataset is not available anymore so the example is actually broken currently on the OpenLayers website ( https://openlayers.org/en/latest/examples/zoomify.html ). So I replaced the example with an image I uploaded on a public server, and this image has several tile groups, which is a more complicated test case.

Everything works as expected in the MR, so I think it actually finally solves https://github.com/openlayers/openlayers/issues/7795 in the general case. 

Bonus: a (zoomed in) screenshot of the difference between retina and non-retina, on a retina screen, from the example image.

Screenshot 2020-01-21 at 20 37 04

### DIFF
--- a/examples/zoomify.html
+++ b/examples/zoomify.html
@@ -10,7 +10,6 @@ tags: "zoomify, deep zoom, IIP, pixel, projection"
 <div class="controls">
   <select id="zoomifyProtocol">
     <option value="zoomify">Zoomify</option>
-    <option value="iip">IIP</option>
     <option value="zoomifyretina">Zoomify Retina</option>
   </select>
 </div>
--- a/examples/zoomify.js
+++ b/examples/zoomify.js
@@ -8,15 +8,27 @@ const imgHeight = 3000;
 
 const zoomifyUrl = 'https://ol-zoomify.surge.sh/zoomify/';
 
-const layer = new TileLayer({
-  source: new Zoomify({
-    url: zoomifyUrl,
-    size: [imgWidth, imgHeight],
-    crossOrigin: 'anonymous'
-  })
+const source = new Zoomify({
+  url: zoomifyUrl,
+  size: [imgWidth, imgHeight],
+  crossOrigin: 'anonymous',
+  zDirection: -1 // Ensure we get a tile with the screen resolution or higher
+});
+const extent = source.getTileGrid().getExtent();
+
+const retinaPixelRatio = 2;
+const retinaSource = new Zoomify({
+  url: zoomifyUrl,
+  size: [imgWidth, imgHeight],
+  crossOrigin: 'anonymous',
+  zDirection: -1, // Ensure we get a tile with the screen resolution or higher
+  tilePixelRatio: retinaPixelRatio, // Display retina tiles
+  tileSize: 256 / retinaPixelRatio // from a higher zoom level
 });
 
-const extent = [0, -imgHeight, imgWidth, 0];
+const layer = new TileLayer({
+  source: source
+});
 
 const map = new Map({
   layers: [layer],
@@ -35,20 +47,9 @@ const control = document.getElementById('zoomifyProtocol');
 control.addEventListener('change', function(event) {
   const value = event.currentTarget.value;
   if (value === 'zoomify') {
-    layer.setSource(new Zoomify({
-      url: zoomifyUrl,
-      size: [imgWidth, imgHeight],
-      crossOrigin: 'anonymous',
-      zDirection: -1 // Ensure we get the most precise tile in any case
-    }));
+    layer.setSource(source);
   } else if (value === 'zoomifyretina') {
-    layer.setSource(new Zoomify({
-      url: zoomifyUrl,
-      size: [imgWidth, imgHeight],
-      crossOrigin: 'anonymous',
-      zDirection: -1, // Ensure we get the most precise tile in any case
-      tilePixelRatio: 2 // Display retina tiles
-    }));
+    layer.setSource(retinaSource);
   }
 });
 
--- a/examples/zoomify.js
+++ b/examples/zoomify.js
@@ -3,17 +3,13 @@ import View from '../src/ol/View.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import Zoomify from '../src/ol/source/Zoomify.js';
 
-const imgWidth = 9911;
-const imgHeight = 6100;
+const imgWidth = 4000;
+const imgHeight = 3000;
 
-const zoomifyUrl = 'http://vips.vtech.fr/cgi-bin/iipsrv.fcgi?zoomify=' +
-    '/mnt/MD1/AD00/plan_CHU-4HD-01/FOND.TIF/';
-const iipUrl = 'http://vips.vtech.fr/cgi-bin/iipsrv.fcgi?FIF=' + '/mnt/MD1/AD00/plan_CHU-4HD-01/FOND.TIF' + '&JTL={z},{tileIndex}';
+const zoomifyUrl = 'https://ol-zoomify.surge.sh/zoomify/';
 
 const layer = new TileLayer({
   source: new Zoomify({
-    tileSize: 256,
-    tilePixelRatio: 1,
     url: zoomifyUrl,
     size: [imgWidth, imgHeight],
     crossOrigin: 'anonymous'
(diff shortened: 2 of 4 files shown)


crubier:
Solves https://github.com/openlayers/openlayers/issues/10539

crubier:
@ahocevar here is a better explanation of my understanding of the zoomify tile problem:

## Before #9110

`tilePixelRatio` was not an option for zoomify tiles. Zoomify tiles always looked bad on retina, with no way of changing this. It was bad enough that we had users complain.

## After #9110, Before #9489 

`tilePixelRatio` was taken into account in the "openlayers standard way". But it means that it impacted *both* the resolution of tiles on the screen*and* the assumption about tile resolution served by the server. 

Basically with a `tilePixelRatio=2` and `tileSize=256`, OpenLayers displayed tiles with the right retina resolution *but* expected the server to serve JPGs of 512px, which can be inconvenient in many cases because I don't know of any zoomify server serving such tiles. 

This led to this situation in the case where the server was not retina-compatible:

  - Either you picked the right `tileSize` and got obviously buggy display where the tiles took only 1/4 of their supposed size, but had the correct stride between them, leaving gaping holes in the display.
  - Or you picked half the nominal `tileSize`, and couldn't load any tiles because it messed up with the zoomify url function.

## After #9489, Before #10562:

It was now possible (through the hack introduced in #9489) to use `tilePixelRatio` to display tiles with the right pixelRatio on screen *while* not needing bigger tiles from the server. 

The big inconvenience was that the extent of the image on screen was influenced by the pixel ratio. 

And then @jamshoward found a bug in this hacky implementation in the case where there are more than one tileGroup (I have no idea why zoomify has this notion of tileGroup btw, seems like unnecessary complexity) https://github.com/openlayers/openlayers/pull/9489#issuecomment-518706270  . This bug led to missing tiles at certain zoom levels. 

After #10562: 

It is now possible to dissociate the resolution you want tiles to be displayed (`tilePixelRatio`) from the size of tiles on the server (`tilePixelRatioOriginal`). This is what I need, and it solves all problems on my side. 

However I understand that it is not the "standard" openlayers way. There are two possibilities

  - Either zoomify is a strange format that actually necessitates this dissociation.
  - Or it's just that the initial implementation of zoomify was problematic in another way, and the problem could be solved differently.

One big question I have that could help me understand better is: 

> Is Openlayers supposed to automatically select the display pixel ratio of a layer? Like if you want openlayers to work on a retina screen, is it done automatically, or do you have to input the pixelRatio somewhere in the map or layer configuration?

- If the display pixel ratio is computed automatically, where can I access it? Is it kosher to just use `window.devicePixelRatio` directly in the layer code, or does the `map` or some other component provide it? In this case we could use this value instead of `tilePixelRatio` in this MR, and rename `tilePixelRatioOriginal` to just `tilePixelRatio`
- If the display pixel ratio is to be set per layer, then I am afraid that we have no choice but to keep theese two pixel ratio specification in parallel, for the reasons above.

ahocevar:
@crubier The screen pixel ratio used for rendering is available in `frameState.pixelRatio`. And the `tilePixelRatio` source option is intended for what you introduced as `tilePixelRatioOriginal`.

So I'd suggest you change this pull request along these lines: `ol/source/Zoomify` needs to override the `getTile` function, which receives the screen pixel ratio as 4th argument. Now before returning the tile, that function sets the `CustomTile`'s `tileSize_` accordingly. The `CustomTile`'s constructor does not need a `tilePixelRatio` option any more then.

crubier:
@ahocevar I basically reverted `ol/source/Zoomify.js` to its state before #9489 and it mostly solves the problem... Don't ask me why, but now it is solved, as shown in the example file you can chose to display the same tileset as either retina or not retina.

However, I think it does not behave as you say it should, because:

  - When on a retina screen with tiles from a non-retina server and `tilePixelRatio=1` the tiles show as non-retina.
  - When on a retina screen with tiles from a non-retina server and `tilePixelRatio=2` the tiles show as retina.

From what I understand of your message, supporting retina screens or not should not involve any parameter change from the point of view of the user of the OpenLayers library.

But here to support retina screen I need to change the source tilePixelRatio, and it does not impact at all what happens on the network side of things, the same tiles are fetched, it only affects the display side of things.

Let me know if this is what is expected.

ahocevar:
@crubier I pushed a commit to your branch, which fixes the tile grid setup and the example configuration. Please take a look to see if things behave now like you want them to behave.

The configuration works like this:

* If you have a tile set with true retina tiles (image size 512px for a 256px tile), you just configure the source with `tilePixelRatio: 2`.
* If you have a standard tile set, but want to display it with a pixel ratio of 2, you configure the source with `tilePixelRatio: 2` and adjust the tile size (128px for an image size of 256px).

Now that's the same thing that works for all tile layers, so it's consistent behavior.

crubier:
@ahocevar well done! It works as expected on my side 👍 This can be merged!

Screenshot 2020-01-27 at 12 35 22

Screenshot 2020-01-27 at 12 35 15

jameshoward:
This probably solves my original problem much better than having to make the generator output 512x512 tiles. However, as I am still using 512x512 tiles, I think there's a problem in the `createFromTemplate` method as I am now back to getting 404 errors because it's looking for tiles in the wrong `TileGroup` directory. 

I've confirmed that generating 256x256 tiles from an image 50% the size creates exactly the same tile group folders, so I think the tiles are correct.

It might be a simple as changing 

```
var tileGroup = ((tileIndex + tileCountUpToTier[tileCoordZ]) / tileWidth) | 0;
```

to

```
var tileGroup = ((tileIndex + tileCountUpToTier[tileCoordZ]) / tileSize) | 0;
```

Because I think the group calculations don't need to factor in the pixel ratio, because they would be the same groups at 256x256 as at 512x512 (or indeed any size)

ahocevar:
@jameshoward When your tiles are 512x512 and you want a tilePixelRatio of 2, you need to configure your source with `tileSize: 256; tilePixelRatio: 2`? This was the intention of the pull request.

jameshoward:
Yes, that was my expectation and they are the settings I used but it still
causes the 404s. I think the problem is that tile group isn’t dependent on
the pixel ratio, but the code is taking it into account.

Still not sure what the tile group directories are for, it seems to cause
more problems than it would solve. :)
-- 
James Howard
ReetGood

Website: http://www.reetgood.co.uk
Twitter: @reetgoodyork <http://twitter.com/#!/reetgoodyork>

ahocevar:
>  I think the problem is that tile group isn’t dependent on
the pixel ratio, but the code is taking it into account.

i don't think that's the problem. it's all about setting the correct tile size. Can you share your tile set?

jameshoward:
The retina (512x512) tiles are at http://ygt.reet.gd/tiles/

Non-retina version is at http://ygt.reet.gd/tiles-regular/. The directory structure and tile file names are identical, it's just that they are 256x256 instead of 512x512.

Using the ~oomify example as a starting point I created 3 sources that, I believe, should all work. However, on the server-retina source you can only zoom to a certain level before you start seeing 404s for the tiles.

```
const imgWidth = 8181;
const imgHeight = 19859;

const zoomifyUrl = 'http://ygt.reet.gd/tiles-regular/';
const zoomifyUrlRetina = 'http://ygt.reet.gd/tiles/';

const retinaPixelRatio = 2;

// Standard Zoomify source using 256x256 tiles from the server.
const source = new Zoomify({
  url: zoomifyUrl,
  size: [imgWidth, imgHeight],
  crossOrigin: 'anonymous',
  zDirection: -1 // Ensure we get a tile with the screen resolution or higher
});

// Retina but using 256x256 tiles from the server and scaling the tile size.
const retinaSource = new Zoomify({
  url: zoomifyUrl,
  size: [imgWidth, imgHeight],
  crossOrigin: 'anonymous',
  zDirection: -1, // Ensure we get a tile with the screen resolution or higher
  tilePixelRatio: retinaPixelRatio, // Display retina tiles
  tileSize: 256 / retinaPixelRatio // from a higher zoom level
});

// Retina using double size (512x512) provided by the server.
const retinaServerSource = new Zoomify({
  url: zoomifyUrlRetina,
  size: [imgWidth, imgHeight],
  crossOrigin: 'anonymous',
  zDirection: -1, // Ensure we get a tile with the screen resolution or higher
  tilePixelRatio: retinaPixelRatio, // Display retina tiles
  tileSize: 256
});
```

ahocevar:
You are right. I was not aware of the impact of the tile group calculation. But we can fix this in a convenient way for the user, without the need for a new option. In #10785, I added a retina detection that simply tries to load a tile for the highest resolution. If the tile url is incorrect, the configuration will be changed to assume server-side retina.

Issue #10541, opened 2020-01-17
Ability to create merged canvas for exporting in V6


ChrisHSandN:
**Is your feature request related to a problem? Please describe.**
Now OL6 uses multiple `<canvas>` elements it has lost the very useful ability to export the current view using the `'rendercomplete' + event.context.canvas` technique.

The example
https://openlayers.org/en/latest/examples/export-map.html
has now been changed for OL6 to use 'html-to-image' but that can only support Chrome / FF which is a severe limitation.

**Describe the solution you'd like**
While testing the idea that I could go through each `<canvas>` element in the DOM in order and write it to a new master canvas using successive calls to `drawImage()` I encountered the issue that CSS `opacity:` and ` transform: matrix()` are applied to the container `<div class="ol-layer">` element and affects how the canvas has to be drawn onto the master.

If OL internally knows how each source canvas in the DOM could be written to a master destination canvas could it:

- do this itself and expose some kind of `ol/Map.toCanvas()`?
- expose enough details (`canvas element`, `sx`, `sy`, `sWidth`, `sHeight`, `dx`, `dy`, `dWidth`, `dHeight`, `opacity`) that this could be done in userland code for an export?

mike-000:
It can be done by users, there's an example of the code needed in https://github.com/openlayers/openlayers/issues/9938#issuecomment-532226277 (doesn't yet handle opacity/globalAlpha) but it's not very user friendly.  Potential candidate for the API method you suggested?

jahow:
This has been brought up several times recently, I'm going to mark this as PR accepted, if anyone feels up to the task... Thanks.

ldv-dev:
i have been using the .catch(function(error) {} from the example mike-000 gave above, and works fine, the only issue is with some amount of  layers (20, 30, 40+) detecting when the maps is fully rendered after a SetResolution().

mike-000:
Here's an update which handles opacity https://codesandbox.io/s/export-pdf-gq5j5

ahocevar:
I have created #10563 as a source of inspiration for a generic map export component. Such a component could be developed and published as a [third-party library](https://openlayers.org/3rd-party/).

PR #10563, opened 2020-01-21
More compatible way of exporting a map as pdf

ahocevar:
As a starting point for a generic map-to-image component, this pull request updates the export-pdf example to use simple canvas composition instead of the `html-to-image` library to export a map to a pdf.

See #10541.


examples/export-pdf.js
```diff
@@ -4,8 +4,6 @@ import WKT from '../src/ol/format/WKT.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
 import {OSM, Vector as VectorSource} from '../src/ol/source.js';
 
-import {toJpeg} from 'html-to-image';
```
tschaub: Can we remove this as a dependency?

ahocevar: No, it is still used in the export-map example.

Issue #4125, opened 2015-09-17
ol.control.FullScreen doesn't fire any event


jonataswalker:
Trying to answer this question http://gis.stackexchange.com/q/162136/50718 I realized that the only way to listen to `fullscreenchange` on webkit browsers for instance is:

``` javascript
document.addEventListener("webkitfullscreenchange", function(evt) {
    console.log(evt);
});
```

Should the docs be changed or the event be listened on `ol.control.FullScreen` instead of [`goog.global.document`](https://github.com/openlayers/ol3/blob/d3c8880b22c61b617989f4e3b5d9eaba1364385c/src/ol/control/fullscreencontrol.js#L70)?

jonataswalker:
Since there's a great effort to get free of Closure Library dependency and `ol.control.FullScreen` itself has no event:
```javascript
//src/ol/control/fullscreencontrol.js
  goog.events.listen(goog.global.document,
      goog.dom.fullscreen.EventType.CHANGE,
      this.handleFullScreenChange_, false, this);
```
Is it worthy writing own fullscreen controller?

azerbini:
@jonataswalker is there any reason why you closed this ticket? As far as I know, FullScreen still has no event.

sookoll:
@jonataswalker, this issue should be reopen, since no event has fired in 4.2.0

jonataswalker:
Well ... reopened.

fredj:
The standard `fullscreenchange` event must be used, see:
 https://developer.mozilla.org/en-US/docs/Web/Events/fullscreenchange

rasa2k:
This is still an issue I think

philip-firstorder:
Can someone please merge @rasa2k changes?

ahocevar:
Do we have a pull request with those changes?

devel0:
Sure, opened PR #10564

PR #10564, opened 2020-01-21
enterfullscreen and leavefullscreen map events, should address #4125.

devel0:


ahocevar:
@fredj would you agree to such a change? In https://github.com/openlayers/openlayers/issues/4125#issuecomment-408875849 you suggested that users should listen for the browser's fullscreen* events instead.

devel0:
I did such patch because using browser fullscreen event didn't worked for specific configuration I used, in other words but I didn't remember now what's exactly was, there was a timing problem using fullscreen browser event so that I cannot set some component ( toolbar, etc in the map ) because browser fullscreen fired before effective openlayer map availability.
Instead using proposed events from the map everthing worked in sync as expecting.

fredj:
Hi, sorry for the late response.

I'm not against it but the event should originated from the fullscreen control, not the map.
@devel0 will #10683 work for you?

devel0:
It should work because it will sync to the map state even originating from the fullscreen control and for sure it's a step forward to add requested feature, we could see further if it will fullfill certain usage scenario but I suppose it will work as expected. So definitively I happy if such addition will available.

fredj:
see #10683

PR #10570, opened 2020-01-24
Use correct extent for the vector image

ahocevar:
Currently the `imageRatio` for `ol/layer/VectorImage` is not respected. To fix this, we need to pass the correct extent to the image frame state.


ahocevar:
Compare panning

* Before: https://openlayers.org/en/v6.1.1/examples/image-vector-layer.html
* After: https://505-4974939-gh.circle-artifacts.com/0/examples/image-vector-layer.html

Just pay attention to when you fall off the edge of the map.

PR #10572, opened 2020-01-24
Assert each layer is only added to the map once

### LINKED ISSUE #10565: Performance regression in ol6 vs ol5
Since we switched to OL6, the FPS of our map tanked, from 25 FPS to 5 FPS. Scrolling and zooming is rally laggy when it used to be super smooth.

Our use case is really simple, just one tile layer.

There seems to be something weird going on but I can't tell what. An interesting thing is the memory usage in the profiler: it looks like scrolling panning and zooming triggers a lot of Garbage collection, and the memory usage stays really constant with of a lot of garbage collecting, when it used to vary a lot more in ol5, Garbage collecting much less frequently. 

OL5 profile when panning, 2s sample:

Screenshot 2020-01-22 at 16 40 18

OL6 profile when panning, 2s sample:

Screenshot 2020-01-22 at 16 40 07

ahocevar:
Fixes #10565.

### DIFF
--- a/doc/errors/index.md
+++ b/doc/errors/index.md
@@ -245,3 +245,7 @@ Layer opacity must be a number.
 
 `forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has not been enabled.
 This is done by providing adequate shaders using the `hitVertexShader` and `hitFragmentShader` properties of `WebGLPointsLayerRenderer`.
+
+### 67
+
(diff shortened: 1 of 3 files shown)


jahow:
Not sure what is going on with the CI...

crubier:
It works, it throws as expected.

Screenshot 2020-01-24 at 15 47 29

PR #10573, opened 2020-01-24
Replace Bing layer with MapTiler in example

mike-000:
Another example which wasn't caught in #10315


fredj:
Thanks for you work on that @mike-000 !
You have updated `drag-and-drop-image-vector.js` and `drag-and-drop.html` but both examples (js and html) must be changed.

PR #10574, opened 2020-01-25
Text rendering improvements

### LINKED ISSUE #10555: Smooth rendering ol.style.Text stroke
**Describe the bug**
The backgroundstroke in ol.style.Text is rendered smoothed, i.e with no sharp edges.

**To Reproduce**
https://codepen.io/mattti/pen/OJPagag
It is specially visible when zooming. Sometimes the edges are sharp, on different zoom level it is smoothed.

**Expected behavior**
I would expect the edges to be sharp without smoothing.

I tried to apply this suggestions but to no avail (I had to replace the "precompose" with "prerender" because the context is missing in precompose handler):
https://stackoverflow.com/questions/35875270/turn-off-image-smoothing-in-openlayers-3

ahocevar:
This pull request makes sure that label rendering does not change properties on the rendering context. It also improves performance of label rendering by storing the label dimensions and instructions on the executor. That reduces garbage and avoids repeated `measureText` calls on the context.

Fixes #10555.

### DIFF
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -4,7 +4,6 @@
 import {getFontParameters} from '../css.js';
 import {createCanvasContext2D} from '../dom.js';
 import {clear} from '../obj.js';
-import {create as createTransform} from '../transform.js';
 import {executeLabelInstructions} from './canvas/Executor.js';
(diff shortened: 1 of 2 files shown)


ahocevar:
Thanks for any review.

PR #10580, opened 2020-01-27
Fix KML Polystyle outline 0 conflict with Linestyle for linestrings

### LINKED ISSUE #4466: KML: <outline> from <PolyStyle>
Hi all,

If I load the attached KML, OL3 doesn't show it because the KML has a PolyStyle with outline=0. This KML only has a LineString (MultiGeometry), so PolyStyle shouldn't affect to LineString appearance, I think. If the value of outline is 1, or if I delete the outline element, OL3 loads the KML right and shows the LineString. Any idea about this?

Thanks
[doc.txt](https://github.com/openlayers/ol3/files/40087/doc.txt)

### LINKED ISSUE #6028: Reading KML features: retrieved styles
Hi,

It seems that there's a bug with the ol.format.KML.readFeature method. In the case of a KML style defined in this way:

```
<Style id="LineStyle37">
        <LabelStyle>
            <color>00000000</color>
            <scale>0</scale>
        </LabelStyle>
        <LineStyle>
            <color>ffff7000</color>
            <width>1.5</width>
        </LineStyle>
        <PolyStyle>
            <color>00000000</color>
            <outline>0</outline>
        </PolyStyle>
    </Style>
```

the style of the feature retrieved by readFeature() is invalid, because the line color is black (the PolyStyle color). The way to fix this, so the style be the same as LineStyle, is to modify the KML in the following ways:
- Changing PolyStyle.outline to 1,
- Or commenting/deleting the PolyStyle element

In other cases, the line color is the value of the element PolyStyle.color.

Regards

### LINKED ISSUE #10538: Polystyle outline affects Linestyle width in KML
[example.txt](https://github.com/openlayers/openlayers/files/4076654/example.txt)

In the attached simplified code, the PolyStyle setting <outline>0</outline> is causing the linestring not to render in OpenLayers. It will however render in Google Earth. 

This seems to be an ongoing issue and was mentioned historically in the following
https://github.com/openlayers/openlayers/issues/4466 
https://github.com/openlayers/openlayers/issues/6028 

can be tested in https://openlayers.org/en/latest/examples/drag-and-drop.html (alter attached file extension to .kml)

Altering the KML is not necessarily convenient as our KMLs are uploaded to product directly by our customers. 

Any help on this would be appreciated

### LINKED ISSUE #10571: KML imports correctly, but data doens't show
Our generated .KML file imports to our Open Layers application, but the data doesn't appear in the layer.

File included in zip
[layer.zip](https://github.com/openlayers/openlayers/files/4108893/layer.zip)

mike-000:
Fixes #10571
Fixes #10538
Fixes #6028
Fixes #4466

Return separate style objects for polygon geometries (including multipolygon and polygons in geometry collections) and other geometries if Polystyle outline is defined as 0.

Update Polystyle outline 0 tests

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -882,32 +882,32 @@ class KML extends XMLFeature {
  * @return {Style} style Style.
  */
 function createNameStyleFunction(foundStyle, name) {
-  let textStyle = null;
   const textOffset = [0, 0];
   let textAlign = 'start';
-  if (foundStyle.getImage()) {
-    let imageSize = foundStyle.getImage().getImageSize();
+  const imageStyle = foundStyle.getImage();
+  if (imageStyle) {
+    let imageSize = imageStyle.getImageSize();
     if (imageSize === null) {
       imageSize = DEFAULT_IMAGE_STYLE_SIZE;
     }
     if (imageSize.length == 2) {
-      const imageScale = foundStyle.getImage().getScale();
-      // Offset the label to be centered to the right of the icon, if there is
-      // one.
+      const imageScale = imageStyle.getScale();
+      // Offset the label to be centered to the right of the icon,
+      // if there is one.
       textOffset[0] = imageScale * imageSize[0] / 2;
       textOffset[1] = -imageScale * imageSize[1] / 2;
       textAlign = 'left';
     }
   }
-  if (foundStyle.getText() !== null) {
+  let textStyle = foundStyle.getText();
+  if (textStyle) {
     // clone the text style, customizing it with name, alignments and offset.
     // Note that kml does not support many text options that OpenLayers does (rotation, textBaseline).
-    const foundText = foundStyle.getText();
-    textStyle = foundText.clone();
-    textStyle.setFont(foundText.getFont() || DEFAULT_TEXT_STYLE.getFont());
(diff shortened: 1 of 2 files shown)


mike-000:
> Thanks for your ongoing effort on this. Can you please indicate when this pull request is ready for review? As a rule of thumb, it is better to address one issue with one pull request, if possible.

It is ready.  It can be seen in action using the KML from https://github.com/openlayers/openlayers/issues/6028#issuecomment-256590611 with `<outline>1</outline>` changed to `<outline>0</outline>`.  #10582 still requires a fix but I will do that as a follow up.

src/ol/format/KML.js
```diff
@@ -1766,13 +1768,57 @@ function readStyle(node, objectStack) {
...
+            );
+          } else if (type !== GeometryType.POLYGON && type !== GeometryType.MULTI_POLYGON) {
+            return geometry;
+          }
+        },
+        fill: fillStyle,
+        image: imageStyle,
+        stroke: strokeStyle,
+        text: textStyle,
+        zIndex: undefined // FIXME
```
ahocevar: What is supposed to be fixed here?

mike-000: I copied it from the original code.  If anyone knows what is means there will also need fixing in the new geometry specific section.

Issue #10586, opened 2020-01-29
Add option to use larger resolution constraint for a View


jeremy-smith-maco:
If the resolutions of the max extent do not exceed the max resolution in `resolutionconstraint.js` it returns the lower value of the resolution values for the extent. I want to add an option where instead if the resolution values for the extent are lower than the max resolution then it will return the large of the two values.

PR #10587, opened 2020-01-29
Adds option to View for using larger resolution value when clamping #10586

jeremy-smith-maco:
Adds functionality for #10586

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -129,8 +129,9 @@ import {createMinMaxResolution} from './resolutionconstraint.js';
  * @property {boolean} [smoothResolutionConstraint=true] If true, the resolution
  * min/max values will be applied smoothly, i. e. allow the view to exceed slightly
  * the given resolution or zoom bounds.
- * @property {boolean} [largerResolutionConstraint=false] If true, the constrained
- * resolution values will use the larger value to do so.
+ * @property {boolean} [constrainOneAxis=false] If true, the extent constraint can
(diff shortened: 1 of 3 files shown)


mike-000:
The proposed description in a API 

> If true, the constrained resolution values will use the larger value to do so.

wouldn't be easily understood by a user

Looking at you code I think the effect could be better described as

> If true, the extent constraint can be exceeded along one but not both axes

and something like `constrainOneAxis` would be a more understandable option name

jeremy-smith-maco:
Wasn't sure how to best explain it. I have updated it. Hopefully it is more concise.

tschaub:
This looks like nice functionality.  What do you think about calling the option `showFullExtent`?  I’m trying to think how someone might find the option if they don’t know it exists.  And maybe the documentation could be something like this:

Allow the view to be zoomed out to show the full configured extent.  By default, when a view is configured with an extent, users will not be able to zoom out so the viewport exceeds the extent in either dimension.  This means the full extent may not be visible if the viewport is taller or wider than the aspect ratio of the configured extent.  If `showFullExtent` is true, the user will be able to zoom out so that the viewport exceeds the height or width of the configured extent, but not both, allowing the full extent to be shown.

jeremy-smith-maco:
Makes more sense. Updated.

ahocevar:
Thanks @tschaub for your valuable input. I was hesitant to review this pull request because neither the option name nor the documentation made sense to me. Now everything should be clear. Thanks @jeremy-smith-maco for your continued effort on this.

PR #10588, opened 2020-01-29
Sort doc navigation properly and add missing modules

ahocevar:
This pull request improves the user experience when navigating the API doc, by applying a more natural sort order.

It also adds missing modules (e.g. for classes that also have static functions, e.g. `ol/geom/Polygon` to accommodate static functions like `fromExtent`.

### DIFF
--- a/config/jsdoc/api/template/publish.js
+++ b/config/jsdoc/api/template/publish.js
@@ -188,6 +188,12 @@ function attachModuleSymbols(doclets, modules) {
   });
 }
 
+function getPrettyName(longname) {
+  return longname
+    .split('~')[0]
+    .replace('module:', '');
+}
+
 /**
  * Create the navigation sidebar.
  * @param {object} members The members that will be used to create the sidebar.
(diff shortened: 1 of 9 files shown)


ahocevar:
Link for testing the docs: https://511-4974939-gh.circle-artifacts.com/0/apidoc/index.html

tschaub:
The new sort order looks great.

I notice this adds duplicate entries in the table of contents for modules with a class as a default export:

image

tschaub:
Not sure how straightforward this would be, but I think it would be sufficient if each entry in the table of contents were a module (instead of a module or a class).  Then if you expanded that entry, it would show the exports from a module – classes, functions, enum/types (so you wouldn't be able to expand a class and see its methods in the TOC).

So the TOC entry for the `ol/interaction/Draw.js` module might look something like this:

image

tschaub:
But then I guess we are back to two clicks to get to things like the `ol/Map` constructor.  Feels like we've been there before.

ahocevar:
Right, we had classes nested under modules previously already.

ahocevar:
I added two commits that reduce the amount of modules created in the navigation. @tschaub Please see if this looks better to you now.

PR #10598, opened 2020-02-01
Remove inheritDoc to work around JSDoc issue

### LINKED ISSUE #10317: Missing documentation of VectorTileSource#refresh
VectorTileSource inherits the class `ol/source/Tile`, which has the method `refresh`. This is missing from the documentation.

ahocevar:
Fixes #10317 

See https://github.com/openlayers/openlayers/issues/10317#issuecomment-581038671.

### DIFF
--- a/src/ol/source/Tile.js
+++ b/src/ol/source/Tile.js
@@ -317,9 +317,6 @@ class TileSource extends Source {
     this.tileCache.clear();
   }
 
-  /**
-   * @inheritDoc
-   */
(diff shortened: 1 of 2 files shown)


MoonE:
There is also the jsdoc of  `ol/source/Vector`'s refresh method that prevents it from being displayed for `ol/source/Cluster`.

ahocevar:
Thanks @MoonE, I just removed that one as well.

PR #10613, opened 2020-02-05
Show KML name labels for Points in MultiGeometry

### LINKED ISSUE #10582: KML labels for Points in MultiGeometry
Google Earth displays state names for https://github.com/openlayers/openlayers/blob/master/test/spec/ol/format/kml/states.kml but OpenLayers does not as labels are only drawn if `geometry.getType() === GeometryType.POINT` in `ol/format/KML`'s `createFeatureStyleFunction`.

Is this an intentional restriction or should the check be extended to find Points in MultiGeometry?

mike-000:
Fixes #10582

Ensures labels are shown for Points regardless of whether they are a single Point geometry or part of a MultiGeometry as with Google Earth.

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -918,7 +918,11 @@ function createNameStyleFunction(foundStyle, name) {
 
   const nameStyle = new Style({
     image: imageStyle,
-    text: textStyle
+    text: textStyle,
+    // although nameStyle will be used only for Point geometries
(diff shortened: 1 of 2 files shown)


mike-000:
This is ready and can be seen in action with https://github.com/openlayers/openlayers/blob/master/test/spec/ol/format/kml/states.kml   Two temporary lines are used to maintain compatibility with the existing writeStyles functionality which as noted in #10614 needs revising and can be removed again when that is done.

PR #10615, opened 2020-02-06
Mark GeometryLayout as public API

bjornharrtell:
Missing `ol.geom.GeometryLayout` in full build at https://cdn.rawgit.com/openlayers/openlayers.github.io/master/en/v6.1.1/build/ol.js

But it's listed as part of the API in the docs (assuming the docs show the public API?) https://openlayers.org/en/latest/apidoc/module-ol_geom_GeometryLayout.html

### DIFF
--- a/src/ol/geom/GeometryLayout.js
+++ b/src/ol/geom/GeometryLayout.js
@@ -7,6 +7,7 @@
  * or measure ('M') coordinate is available. Supported values are `'XY'`,
  * `'XYZ'`, `'XYM'`, `'XYZM'`.
  * @enum {string}
+ * @api
  */
 export default {
   XY: 'XY',


mike-000:
It seems to be an inconsistency, not a bug.  It's similar to ol/geom/GeometryType., ol/style/TextPlacement, etc. where users have been expected to supply the string values as in https://openlayers.org/en/latest/examples/draw-features.html

Also :ol/style/IconOrigin ol/style/IconAnchorUnits ol/style/IconOrigin but those are dead links in https://openlayers.org/en/latest/apidoc/module-ol_style_Icon-Icon.html

bjornharrtell:
Ah you are right @mike-000.

PR #10632, opened 2020-02-10
Draw interaction: Append coordinates to polygons and lines (reworked)

jahow:
This PR is a follow-up of #10428. The example showcasing the new feature was rewritten, and the code was rebased on the master branch. Most original changes to the library code are left unchanged.

I thought the example in the original PR did not give a clear enough impression of what could be achieved with the new feature, so now the example shows a kind of "tracing" feature (similar to Qgis) which makes drawing around an existing geometry while preserving topology easier. There is a bit of computation involved but IMO the example is still understandable.

[Link to the example](https://3122-4723248-gh.circle-artifacts.com/0/examples/tracing.html)

Thanks again @opelline for the original contribution.


src/ol/interaction/Draw.js
```diff
@@ -903,9 +902,44 @@ class Draw extends PointerInteraction {
...
-   * extend lines by adding points to the end of the coordinates array.
+   * Append coordinates to the end of the geometry that is currently being drawn.
+   * This can be used when drawing LineStrings or Polygons. Coordinates will
+   * either be appended to the current LineString or the outer ring of the current
+   * Polygon.
+   * @param {!LineCoordType} coordinateExtension Linear coordinates to be appended into
+   * the coordinate array.
+   * @api
+   */
+  appendCoordinates(coordinateExtension) {
```
ahocevar: Minor, but naming the argument `coordinates` would make the API easier to read.

src/ol/interaction/Draw.js
```diff
@@ -903,9 +902,44 @@ class Draw extends PointerInteraction {
...
+
+  /**
+   * Initiate draw mode by starting from an existing geometry which will
+   * receive new additional points. This only works on features with
+   * `LineString` geometries, where the interaction will extend lines by adding
+   * points to the end of the coordinates array.
+   * This will change the original feature, instead of drawing a copy.
+   *
+   * The function will dispatch a `drawstart` event.
+   *
```
ahocevar: I'm wondering why this is limited to linestrings. Shouldn't this work the same way for polygons?

examples/tracing.js
```diff
@@ -0,0 +1,306 @@
...
+
+
+// layers definition
+
+const raster = new TileLayer({
+  source: new OSM()
+});
+
+// Idaho state GeoJSON
+const baseFeature = new GeoJSON().readFeature({
```
ahocevar: Having this outside the example would allow the user to focus more on the code. You could do something like this instead:
```js
const baseVector = new VectorLayer({
  source: new VectorSource({
    format: new GeoJSON(),
    url: 'https://ahocevar.com/geoserver/wfs?service=wfs&request=getfeature&typename=topp:states&cql_filter=STATE_NAME=\'Idaho\'&outputformat=application/json'
  })
});
```

src/ol/interaction/Draw.js
ahocevar: Actually I see now that what you added as `appendCoordinates` api function is more or less a replacement for this old `extend`  function. It also looks like this function is not used in any example, and looking at the new example I don't think it is very useful Couldn't this be marked `@deprecated` and removed in v7?

ahocevar: For reference: `Draw#extend` was introduced in #4044. It still might be useful. @jahow, do you think you can quickly check if this works for polygons out of the box, and update the docs if this is the case?

jahow: The `extend` method gives errors if used on polygons, or in polygon drawing mode. Even when extending an existing line, there is an error when finishing the draw. I'll create an issue with a reproduction case, maybe this API needs some more work or to be deprecated in favor of something more generic?

Anyway, the `appendCoordinates` API doesn't fill exactly the same purpose so I think both can coexist for the time being.

src/ol/interaction/Draw.js
jahow: Agreed, changed in https://github.com/openlayers/openlayers/pull/10632/commits/8d4c3b28875500484372bea736e1706b9e3dbca1

examples/tracing.js
jahow: Good idea! Much simpler. Done in https://github.com/openlayers/openlayers/pull/10632/commits/6641832621add9c4a8de2bd18151596adeb0e731

jahow:
Thanks for the review @ahocevar!

opelline:
Thank you @jahow for the improved example, and thanks for the review & approval @ahocevar !

ahocevar:
I noticed that the `appendCoordinates()` method does not work as expected on touch devices. This can also be seen in the tracing.html example: it takes another click after calling the method to see its result.

@jahow or @opelline, could either of you take a look and try to fix that? Thanks in advance.

ahocevar:
Actually I was able to fix that as part of another fix I'd been working on. See https://github.com/openlayers/openlayers/pull/11800/commits/f239a20b596966643122066398d504685e04ea4b. Maybe you could give that a review, @jahow?

PR #10639, opened 2020-02-11
Improve new mousewheel zoom

### LINKED ISSUE #10638: Issue with constraints for new mousewheel zoom
**Describe the bug**
The smooth resolution constraint allows zooming out slightly more than 

**To Reproduce**
1. Go to https://openlayers.org/en/v6.2.0/examples/drag-rotate-and-zoom.htmland compare the zoom out behaviour to https://openlayers.org/en/v6.1.1/examples/drag-rotate-and-zoom.html
2. with v6.2.0 you can zoom out ~1 zoom level past the constraint, with v6.1.1 smooth constraints doesn't seem to have an effect on zoom.

**Expected behavior**
According to the description for `smoothResolutionConstraint`  of `ol/View` it should be possible to exceed it slightly. 
> (defaults to true) | If true, the resolution min/max values will be applied smoothly, i. e. allow the view to exceed slightly the given resolution or zoom bounds.

This sounds like the old *and* the new behaviors are wrong.

MoonE:
Fixes #10638

This reduces the allowed amount of over-zoom allowed in the view
to provide faster feedback to the user that the limit is reached.

And it reduces the MouseWheelZoom's event gap to remove the small
pause after which the view snaps back to the constrained resolution.

### DIFF
--- a/src/ol/interaction/MouseWheelZoom.js
+++ b/src/ol/interaction/MouseWheelZoom.js
@@ -96,7 +96,7 @@ class MouseWheelZoom extends Interaction {
      * interactions.
      * @type {number}
      */
-    this.eventGap_ = 400;
+    this.eventGap_ = 250;
 
     /**
      * @type {?}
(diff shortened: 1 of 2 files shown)


ahocevar:
Preview of this change: https://3138-4723248-gh.circle-artifacts.com/0/examples/simple.html

I am not 100% sure if I like the new behavior better when using a trackpad. A trackpad has much more inertia than a mouse wheel, which now causes a lack of feedback while zooming out further than the min zoom.

What do others think?

MoonE:
Unfortunately I do not have a track pad to test with.

PR #10654, opened 2020-02-13
deltaX and deltaY as strings

wussup:
The arguments deltaX and deltaY may be numbers represented as strings (for example, values from input forms), in that case, coordinates are summarised as strings and that's very bad.

### DIFF
--- a/src/ol/geom/flat/transform.js
+++ b/src/ol/geom/flat/transform.js
@@ -108,8 +108,8 @@ export function translate(flatCoordinates, offset, end, stride, deltaX, deltaY, 
   const dest = opt_dest ? opt_dest : [];
   let i = 0;
   for (let j = offset; j < end; j += stride) {
-    dest[i++] = flatCoordinates[j] + deltaX;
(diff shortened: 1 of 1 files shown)


wussup:
> This change is problematic, because it adds extra function calls on a super hot path. Note that coordinates in OpenLayers **never** accept strings. Also not in the translate function. See the API docs for its signature, which contains
> 
> ```js
> /**
>  ...
>  * @param {number} deltaX Delta X.
>  * @param {number} deltaY Delta Y.
>  ...
>  */
> ```

Yes, you are right, but I was confused when I pass by accident deltaX and deltaY as numbers represented by strings and translate not worked without any errors or warnings.
When I added ex. 1000 to coordinate 102030.22 I log coordinates and see something like "102030.221000", just then I understood what is going on.

tschaub:
Thanks for your proposed contribution, @wussup.  I agree with @ahocevar that we don’t want to accept strings in addition to numbers in this function - especially since it is an internal function and not part of the API.

Please parse your string values as numbers before passing them to the API.

PR #10669, opened 2020-02-14
More browser compatible Export Map example

### LINKED ISSUE #10643: Export Map example does not work in Safari
The example to export map is at https://openlayers.org/en/latest/examples/export-map.html

Expected Behaviour:
1. Click PNG
2. Download PNG
3. Open PNG and it contains the map viewport

Actual Behaviour in Safari Version 13.0.5 (15608.5.11) (works ok with Firefox):
1. Click PNG
2. Download PNG
3. Open PNG and it either contains no data, a completely black PNG or random data in the PNG.

map-2

mike-000:
Fixes #10643

Removes the overlay and  `html-to-image`  to support browsers as much as possible.

Removes the `html-to-image` dependency.

IE and Edge export only the canvas and use msSaveBlob as the link download attribute method does not work.  Chrome exports using the download link.  Please can reviewers check Safari.

### DIFF
--- a/examples/export-map.css
+++ b/examples/export-map.css
@@ -1,6 +1,0 @@
-.overlay {
-  background-color: yellow;
-  border-radius: 6px;
-  padding: 4px;
-  white-space: nowrap;
-}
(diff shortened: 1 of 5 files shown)


mike-000:
I don't use Safari but silent fails would explain the black exports

PR #9565, opened 2019-05-17
Make mousewheel zoom behave like trackpad zoom.

dbrnz:
This resolves issue #9564.

PR #10672, opened 2020-02-15
Nicer mousewheel and trackpad zooming

### LINKED ISSUE #10641: Map pans with mousewheel zoom
**Describe the bug**
The map pans when the mouse wheel is used during the snap-back animation of the view.

**To Reproduce**
1. Go to https://openlayers.org/en/v6.2.0/examples/drag-rotate-and-zoom.html
2. Zoom out farther than the resolution constraint allows
3. Wait a new hundred milliseconds until it snaps back.
4. Use the mouse wheel during the snap-back animation.
5. If the mouse was positioned to the left of the map center it jumps to the left, otherwise to the right. The more it was over-zoomed the farther the jump is.

**Expected behavior**
The map coordinate at the mouse pointer should not change.

### LINKED ISSUE #10659: the intermediary zoom levels are still allowed when constrainResolution is true
**Describe the bug**
I'm using v6.2.1 and v6.2.0 also has this issue.

From the document:  
constrainResolution: If true, the view will always animate to the closest zoom level after an interaction;false means intermediary zoom levels are allowed.

Does it mean intermediary zoom levels shouldn't allowed when constrainResolution is true?

Because there are still intermediary zoom levels after I scroll on the map, it just animate to the closest zoom level

**To Reproduce**
I just set up a map normally
```
self.olMap = new Map({
                target: 'map',
                logo: false,
                interactions: interactions.extend([dragAndDropInteraction]),
                moveTolerance: 2,
                view: new View({
                    projection: newProj,
                    constrainResolution: true,
                    resolution:280,
                    resolutions: Config.Resolutions,
                    center: [474200, 5350629],
                    extent: Config.EXTENT,
                    zoom: 1
                })
            });
```

**Expected behavior**
Mouse scroll with the resolutions I set up to the view.

Thanks

ahocevar:
This pull request reverts the changes from #9565, and replaces them with a smaller change that gives the MouseWheelZoom interaction two different behaviors when using a mouse:
* When no zoom constraint is set on the view, one wheel tick will zoom the map only a fraction of a zoom level. The faster the wheel is moved, the more the movement of a single tick will be.
* When a zoom constaint is set on the view, the behavior is the same as it was before v6.2 - one tick zooms by 1 integer zoom.

For trackpad use, two improvements were made:
* The rebound behavior has been improved to feel more like the page scroll rebound behavior of browsers.
* When the view has a resolution constraint, the mousewheel behavior is used instead of trackpad mode.

Fixes #10659.
Fixes #10641.
Replaces and closes #10639.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -794,6 +794,13 @@ class View extends BaseObject {
   }
 
   /**
+   * @return {boolean} Resolution constraint is set
+   */
+  getConstrainResolution() {
+    return this.options_.constrainResolution;
+  }
+
+  /**
    * @param {Array<number>=} opt_hints Destination array.
    * @return {Array<number>} Hint.
    */
--- a/src/ol/interaction/MouseWheelZoom.js
+++ b/src/ol/interaction/MouseWheelZoom.js
@@ -4,7 +4,17 @@
 import {always, focus} from '../events/condition.js';
 import EventType from '../events/EventType.js';
 import {DEVICE_PIXEL_RATIO, FIREFOX} from '../has.js';
-import Interaction from './Interaction.js';
+import Interaction, {zoomByDelta} from './Interaction.js';
+import {clamp} from '../math.js';
+
+
+/**
+ * @enum {string}
+ */
+export const Mode = {
+  TRACKPAD: 'trackpad',
+  WHEEL: 'wheel'
+};
 
 
 /**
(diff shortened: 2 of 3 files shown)


ahocevar:
This is now ready for review. Example page for testing: https://535-4974939-gh.circle-artifacts.com/0/examples/simple.html

ahocevar:
Thanks for the review @jahow.

PR #10676, opened 2020-02-16
Document that overviewmap view must use same projection as main map

### LINKED ISSUE #10670: Overview map wrong unless using same projection than main map.
**Describe the bug**
The OverviewMap control assumes the overview map uses the same projection than the main map. However it defaults to using EPSG:3857 regardless of the main map's view.
Thusly, the overview map shows the wrong location whenever the view projections mismatch.

**To Reproduce**
Working example where main map and overview map use EPSG:2157 as compiled for #10668 is in https://stackblitz.com/edit/typescript-liwrsh

Broken example where main map shows Ireland but overview map shows Nigeria: https://stackblitz.com/edit/typescript-iwnndr

The difference between the examples is the specification of the projection for the overview map (or lack thereof).

**Expected behavior**
Overview map should:
1.  default to using projection of main map instead of EPSG:3857,
2. transform coordinates if projections of main map and overview map differ.

The transformation could maybe done by simply transforming the center coordinate in https://github.com/openlayers/openlayers/blob/v6.2.1/src/ol/control/OverviewMap.js#L429 but this is just a guess. I didn't look into it in detail.

Similar issues raised earlier:
* #2998
* #5188

mike-000:
Fixes #10670

Notes the current limitation (which could be removed in future if extra functionality can be added)

### DIFF
--- a/src/ol/control/OverviewMap.js
+++ b/src/ol/control/OverviewMap.js
@@ -60,7 +60,8 @@ class ControlledMap extends PluggableMap {
  * to be rendered outside of the map's viewport.
  * @property {string} [tipLabel='Overview map'] Text label to use for the button tip.
  * @property {import("../View.js").default} [view] Custom view for the overview map. If not provided,
- * a default view with an EPSG:3857 projection will be used.
+ * a default view with an EPSG:3857 projection will be used. The main map and the overviewmap must use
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks, @mike-000.

mike-000:
@tschaub  Had you intended to close without merging?

PR #10678, opened 2020-02-16
Add maxResolution option to ol/tilegrid.createXYZ() and ol/source/XYZ

### LINKED ISSUE #10677: Tile Grid based on height of extent
Most EPSG:4326 tile grids have two tiles at level zero, requiring either a custom tile grid or url function.  A createXYZ (and sources which use it) option to calculate a default tilegrid based on the height of the extent would make things much simpler.  It might also be useful for custom extents where the {-y} placeholder is used.

Similar to this https://codesandbox.io/s/xyz-esri-4326-512-dzpb9 but also allow useExtentHeight as an option in source constructors so createXYZ does not need to be called explicitly in user code.

mike-000:
Fixes #10677

Add a maxResolution option to ol/tilegrid.createXYZ() and ol/sourceXYZ to simplify creation of tile grids which are one tile high but more than one wide such as many EPSG4326 setups and/or avoid the need for custom tile url functions.

Simplify the **ArcGIS REST with 512x512 Tiles** example which had used a custom tile url function to work around an oversized default tile grid,

### DIFF
--- a/examples/xyz-esri-4326-512.js
+++ b/examples/xyz-esri-4326-512.js
@@ -3,26 +3,18 @@ import View from '../src/ol/View.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import XYZ from '../src/ol/source/XYZ.js';
 
-// The tile size supported by the ArcGIS tile service.
-const tileSize = 512;
-
-const urlTemplate = 'https://services.arcgisonline.com/arcgis/rest/services/' +
-    'ESRI_Imagery_World_2D/MapServer/tile/{z}/{y}/{x}';
-
 const map = new Map({
   target: 'map',
   layers: [
     new TileLayer({
       source: new XYZ({
         attributions: 'Copyright:© 2013 ESRI, i-cubed, GeoEye',
-        maxZoom: 16,
+        url: 'https://services.arcgisonline.com/arcgis/rest/services/' +
+            'ESRI_Imagery_World_2D/MapServer/tile/{z}/{y}/{x}',
+        maxZoom: 15,
         projection: 'EPSG:4326',
-        tileSize: tileSize,
-        tileUrlFunction: function(tileCoord) {
-          return urlTemplate.replace('{z}', (tileCoord[0] - 1).toString())
-            .replace('{x}', tileCoord[1].toString())
-            .replace('{y}', tileCoord[2].toString());
-        },
+        tileSize: 512, // the tile size supported by the ArcGIS tile service
+        maxResolution: 180 / 512, // Esri's tile grid fits 180 degrees on one 512 px tile
         wrapX: true
       })
     })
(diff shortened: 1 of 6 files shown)


src/ol/tilegrid.js
```diff
@@ -124,8 +127,8 @@ function resolutionsFromExtent(extent, opt_maxZoom, opt_tileSize) {
 
   const tileSize = toSize(opt_tileSize !== undefined ?
     opt_tileSize : DEFAULT_TILE_SIZE);
-  const maxResolution = Math.max(
-    width / tileSize[0], height / tileSize[1]);
+  const maxResolution = opt_maxResolution > 0 ? opt_maxResolution :
```
ahocevar:
```suggestion
  const maxResolution = opt_maxResolution !== undefined ? opt_maxResolution :
```

mike-000: That was intentional,  Zero (or a negative value) is not a valid resolution.

Note I have also added the option to ol/source/VectorTile as it is equally valid for sources such as that used in the **Vector tiles in EPSG:4326** example if that were styled directly via OpenLayers.

ahocevar: Ok! I'm surprised that TypeScript understands it. So fine by me.

examples/xyz-esri-4326-512.js
```diff
@@ -3,26 +3,18 @@ import View from '../src/ol/View.js';
...
+        maxZoom: 15,
         projection: 'EPSG:4326',
-        tileSize: tileSize,
-        tileUrlFunction: function(tileCoord) {
-          return urlTemplate.replace('{z}', (tileCoord[0] - 1).toString())
-            .replace('{x}', tileCoord[1].toString())
-            .replace('{y}', tileCoord[2].toString());
-        },
+        tileSize: 512, // the tile size supported by the ArcGIS tile service
+        maxResolution: 180 / 512, // EPSG:4326 tile grid is 1 tile high, 2 tiles wide at level 0
```
ahocevar: This code/comment combination is a bit misleading. Maybe something like
```suggestion
        maxResolution: 180 / 512, // Esri's tile grid fits 180 degrees on one 512 px tile
```

test/spec/ol/tilegrid/tilegrid.test.js
```diff
@@ -404,16 +404,17 @@ describe('ol.tilegrid.TileGrid', function() {
     it('respects configuration options', function() {
       const tileGrid = createXYZ({
         extent: [10, 20, 30, 40],
+        maxResolution: 10 / 128,
         minZoom: 1,
         maxZoom: 2,
         tileSize: 128
       });
-      expect(tileGrid.getExtent()).to.eql([10, 20, 30, 40]);
```
ahocevar: Why was the extent assertion removed?

mike-000: Seems I overwrote it instead of adding new line

mike-000: The handlebars error is getting rather repetitive!

ahocevar:
Thanks! That CI error has nothing to do with handlebars. Puppeteer cannot launch because it does not find a display. That's an issue on CircleCI's end. For now, I triggered a re-run of the tests.

Issue #10664, opened 2020-02-14
WebGLPointsLayer slows when using larger texture maps


jellyedwards:
**Describe the bug**
The WebGLPointsLayer sprite demo has a small icon set with a png size of 128 x 64. If you need many more icons and your png is larger, say 2048 x 2048 in my case, performance noticeably dips:

![10fpsWebGL](https://user-images.githubusercontent.com/14051992/74529773-734a0500-4f20-11ea-8e3c-df78333f48db.gif)

**To Reproduce**
1. Change the image in the sprite demo to a much larger texture
2. Run the demo and for Chrome I go into developer tools (F12)->Console, hit the vertical ellipsis and select Rendering, then enable the FPS counter:
![image](https://user-images.githubusercontent.com/14051992/74529937-cd4aca80-4f20-11ea-9872-b8bfef4ffddc.png)
3. You should see a low FPS (< 60FPS) as in the image above.

**Expected behavior**
FPS shouldn't be so harshly affected by a bigger texture (in my case it's still < 1MB).

I believe this is because webgl\Helper.js reloads the image onto the graphics card for every frame (line 557 onwards):

```javascript
      // apply value based on type
      if (value instanceof HTMLCanvasElement || value instanceof HTMLImageElement || value instanceof ImageData) {
        // create a texture & put data
        if (!uniform.texture) {
          uniform.texture = gl.createTexture();
        }
        gl.activeTexture(gl[`TEXTURE${textureSlot}`]);
        gl.bindTexture(gl.TEXTURE_2D, uniform.texture);
        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);

        const imageReady = !(value instanceof HTMLImageElement) || /** @type {HTMLImageElement} */(value).complete;
        if (imageReady) {
          gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, value);
        }

        // fill texture slots by increasing index
        gl.uniform1i(this.getUniformLocation(uniform.name), textureSlot++);

      }
```

I've tried a simple only-reload-if-different approach:

```javascript
...
[...]

PR #10684, opened 2020-02-17
perf: only do expensive reload when texture changes

jellyedwards:
This change is aimed to solve #10664 - only upload the texture to the GC if the texture has changed as it's an expensive operation and leads to reduced performance.

### DIFF
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -558,6 +558,7 @@ class WebGLHelper extends Disposable {
       if (value instanceof HTMLCanvasElement || value instanceof HTMLImageElement || value instanceof ImageData) {
         // create a texture & put data
         if (!uniform.texture) {
+          uniform.prevValue = undefined;
           uniform.texture = gl.createTexture();
         }
         gl.activeTexture(gl[`TEXTURE${textureSlot}`]);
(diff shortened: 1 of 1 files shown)


jellyedwards:
Sorry, just noticed a test failed:

Chrome 80.0.3987 (Windows 10.0.0) HTML Image loading handles error event FAILED      
        Error: Uncaught Error: expected false to equal true 
Chrome 80.0.3987 (Windows 10.0.0) HTML Image loading handles error event FAILED
        Error: Uncaught Error: expected false to equal true 

I'll look into it...

fredj:
@jellyedwards is it this test that is failing ? 
https://github.com/openlayers/openlayers/blob/master/test/spec/ol/image.test.js#L35-L44

jellyedwards:
@fredj yes that looks like it, I'll just try it on the original to double-check it wasn't there before my fix for the issue...

jellyedwards:
@fredj yeah that test failed before my changes too. This is my first time contributing, but I could look into why that test is failing. Just to speed things up how do you run a single test?

fredj:
This test is only failing on Windows and it's related to the timeout value; I've changed the value from 200 to 500ms and it is passing:
```diff
diff --git a/test/spec/ol/image.test.js b/test/spec/ol/image.test.js
index bf6e19ef0..b6f75abf4 100644
--- a/test/spec/ol/image.test.js
+++ b/test/spec/ol/image.test.js
@@ -40,7 +40,7 @@ describe('HTML Image loading', function() {
       expect(handleLoad.called).to.be(false);
       expect(handleError.called).to.be(true);
       done();
-    }, 200);
+    }, 500);
   });

   it('handles cancelation', function(done) {
```

fredj:
@jahow I'm creating a PR for the failing test. This one can be merged.

PR #10061, opened 2019-09-29
Use TypeScript for .d.ts genertion

ahocevar:
This is a work-in-progress experiment to get .d.ts generation working with TypeScript >= 3.7.

PR #10688, opened 2020-02-17
Publish type definition files

### LINKED ISSUE #8120: TypeScript (Help Wanted)

ahocevar:
Thanks to recent improvements of TypeScript, we can now use `tsc` to generate type definitions (`.d.ts` files) for OpenLayers.

With the changes in this pull request, `npm run build-package` will create `.d.ts` files. This means that once this pull request is merged, any future release will contain the definitions as part of the npm package.

One thing that @tschaub will like in particular: Thanks to the `"declarationMap": true` option in the TypeScript configuration `.d.ts.map` files will also be created. So clicking "Go to definition" in VS Code will take you to the source code in the `.js` file, and not the `.d.ts` file.

Because TypeScript v3.8.1-rc still lacks proper support for JavaScript enums (microsoft/TypeScript#33575), we cannot use that version for our type checks. But it does create proper type definitions, and for type checking w continue to use v3.5.3.

Fixes #8120
Replaces and closes #10061

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "build-index": "npm run build-package && node tasks/generate-index",
     "build-legacy": "shx rm -rf build && npm run build-index && webpack --config config/webpack-config-legacy-build.js && cleancss --source-map src/ol/ol.css -o build/legacy/ol.css",
     "copy-css": "shx cp src/ol/ol.css build/ol/ol.css",
-    "transpile": "shx rm -rf build/ol && shx mkdir -p build/ol && shx cp -rf src/ol build/ol/src && node tasks/serialize-workers && tsc --project config/tsconfig-build.json",
+    "transpile": "shx rm -rf build/ol && shx mkdir -p build/ol && shx cp -rf src/ol build/ol/src && node tasks/serialize-workers && npx --package typescript@3.8.1-rc tsc --project config/tsconfig-build.json",
     "typecheck": "tsc --pretty",
     "apidoc": "shx rm -rf build/apidoc && jsdoc -R config/jsdoc/api/index.md -c config/jsdoc/api/conf.json -P package.json -d build/apidoc"
   },
--- a/README.md
+++ b/README.md
@@ -41,30 +41,6 @@ See the following examples for more detail on bundling OpenLayers with your appl
  * Using [Parcel](https://github.com/openlayers/ol-parcel)
  * Using [Browserify](https://github.com/openlayers/ol-browserify)
 
-## IntelliSense support and type checking for VS Code
-
-The `ol` package contains a `src/` folder with JSDoc annotated sources. TypeScript can get type definitions from these sources with a `jsconfig.json` config file in the project root:
-```js
-{
-  "compilerOptions": {
-    "checkJs": true,
-    // Point to the JSDoc typed sources when using modules from the ol package
-    "baseUrl": "./",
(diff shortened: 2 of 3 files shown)


jahow:
I have tested this by building the package locally and using `npm link` in a typescript application project to point to this package, it just seems to work perfectly. Jumping straight to `.js` files works well in VSCode. This is great!

package.json
```diff
@@ -22,7 +22,7 @@
     "build-index": "npm run build-package && node tasks/generate-index",
     "build-legacy": "shx rm -rf build && npm run build-index && webpack --config config/webpack-config-legacy-build.js && cleancss --source-map src/ol/ol.css -o build/legacy/ol.css",
     "copy-css": "shx cp src/ol/ol.css build/ol/ol.css",
-    "transpile": "shx rm -rf build/ol && shx mkdir -p build/ol && shx cp -rf src/ol build/ol/src && node tasks/serialize-workers && tsc --project config/tsconfig-build.json",
+    "transpile": "shx rm -rf build/ol && shx mkdir -p build/ol && shx cp -rf src/ol build/ol/src && node tasks/serialize-workers && npm install --no-save typescript@3.8.1-rc && tsc --project config/tsconfig-build.json && npm install",
```
tschaub: I think this could be `npx --package typescript @3.8.1-rc tsc --project config/tsconfig-build.json` (`npx` comes with `npm@5.2+`.  Not sure that's really better or worse though.

tschaub:
Nice work @ahocevar.  I'm sure this will be really appreciated.

package.json
ahocevar: I like it. Force-pushed.

gimerstedt:
not sure if this is the right place to ask but go to definition does not go to the definition, it goes to the source (just like you stated). attempting to explicitly go to the type definition also goes to the source. do i have to remove the source maps to get this working correctly?

alievrenkut:
Is there any plan when this commit will be released?

ahocevar:
You can already `npm install ol@dev` to get this change. A new release with it will be out later this month.

boeckMt:
With what kind of TypeScript version and compiler options should the types from ol@dev be used?

I now have OpenLayers version ol@6.2.2-dev.1583332915528 and TypeScript  version 3.7.5 and get a view errors like: **Cannot use namespace 'GeometryLayout' as a type.ts(2709)**

```
// node_modules/ol/geom/SimpleGeometry.d.ts
...
declare class SimpleGeometry extends Geometry {
    /**
     * @protected
     * @type {GeometryLayout}
     */
    protected layout: GeometryLayout;
...
```

ahocevar:
The types are currently being generated with TypeScript v3.8.1-rc.

PR #10699, opened 2020-02-18
Make Select interaction work when there are multiple instances

### LINKED ISSUE #10501: Multiple select interactions on same features corrupt style
**Describe the bug**
It is not possible to have more than one select interaction activated on the same set of features. The select style is mistaken for the default style with the second interaction.

This has the same root cause as https://github.com/openlayers/openlayers/issues/10486, but is not solved by that fix.

**To Reproduce**
Have 2 select interactions on the same layer. E.g. one for click (`condition: ol.events.condition.click`) and one for hover (`condition: ol.events.condition.pointerMove`).

**Expected behavior**
When the pointer moves over the feature, it should get the selected style and when the pointer moves away, it should acquire the unselected style again.

ahocevar:
Fixes #10501

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -133,6 +133,12 @@ class SelectEvent extends Event {
 
 }
 
+/**
+ * Original feature styles to reset to when features are no longer selected.
+ * @type {Object.<number, import("../style/Style.js").default|Array.<import("../style/Style.js").default>|import("../style/Style.js").StyleFunction>}
+ */
+const originalFeatureStyles = {};
+
 
 /**
  * @classdesc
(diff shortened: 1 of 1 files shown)


src/ol/interaction/Select.js
```diff
@@ -345,28 +343,44 @@ class Select extends Interaction {
...
    * @param {import("../Feature.js").default} feature Feature
    * @private
    */
-  removeSelectedStyle_(feature) {
+  restorePreviousStyle_(feature) {
     const key = getUid(feature);
-    feature.setStyle(this.featureStyleAssociation_[key]);
-    delete this.featureStyleAssociation_[key];
+    const selectInteractions = /** @type {Array<Select>} */ (this.getMap().getInteractions().getArray().filter(function(interaction) {
+      return interaction instanceof Select && interaction.getFeatures().getArray().indexOf(feature) !== -1;
```
michalzielanski: Thanks @ahocevar for this PR. It seems to me that the filter condition should check if the feature has style.
```javascript
return interaction instanceof Select && interaction.getStyle() && interaction.getFeatures().getArray().indexOf(feature) !== -1;
```

ahocevar: @michalzielanski `interaction.getStyle() !== false` to be precise:

> `style`: Style for the selected features. By default the default edit style is used (see module:ol/style). If set to false the selected feature's style will not change.

Can you create a follow-up pull request?

michalzielanski: In other places there is no strong type checking. There is only `if (this.style_)`, so `null` instead of `false` will also work. Using `interaction.getStyle() !== false` only here will cause a malfunction if someone uses `null` instead of `false`.
Creating PR would take me half a day, you will do it quicker :)

ahocevar: See #10708. Feel free to buy me a beer @michalzielanski ;-)

michalzielanski: :)

PR #10702, opened 2020-02-19
Position graticule labels relative to the viewport when view is rotated

### LINKED ISSUE #10673: Graticule labels misplaced in rotated view
**To Reproduce**
Open the Map Graticule example and rotate the view.  Only one label per axis is fully visible, the others are positioned at the hidden edge of the rotated convas instead of the edge of the viewport.

**Expected behavior**
Labels should be positioned where meridians and parallels intersect the edge of the viewport.

mike-000:
Fixes #10673

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -16,15 +16,16 @@ let withCredentials = false;
  * load features.
  *
  * This function takes an {@link module:ol/extent~Extent} representing the area to be loaded,
- * a `{number}` representing the resolution (map units per pixel) and an
- * {@link module:ol/proj/Projection} for the projection  as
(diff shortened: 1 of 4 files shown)


mike-000:
I noticed that as well, but as soon as the map is panned it fixes itself..  I presume the loader isn't being called at every change in rotation and the previously calculated positions are still being used.

ahocevar:
You are right @mike-000. The loader is only called as the image size increases, until it has reached the maximum size it can get from rotating.

I'm wondering if it wouldn't be better to draw the labels in a postrender function. That way, it would also not be necessary to pass the rotation to the loader.

PR #10720, opened 2020-02-21
Apidoc better search

MoonE:
Searching feels rather slow, and the result relevance is not always great.
I set out to fix this.
There were  also a few bugs, that are fixed  with this:
- Clearing the input box after searching doesn't restore the original state instead there are the subtitles visible, and clicking the icon no longer expands members
- Entering an invalid regexp throws an exception
- When hovering a class entry the cursor changes to indicate it can be expanded, but actually only the glyph icon is clickable
- when searching for a class name all members are also matched, but not events because the search is done on the longname of each item / member, and events belong to different classes; searching for 'module' matches everything
- some entries are displayed with a wrong name e.g `ol/interaction/DragBox~DragBoxEvent` is also listed as `ol/interaction/DragBox`
- the keyup event used on the search input also captures keys like CTRL and arrow keys
- some events like e.g. boxdrag are not highlighted when clicked because a jquery error is thrown

Besides making searching faster and showing more relevant results I also did some small improvements (imho):
- If the search box is not empty on page load a search is started to show the corresponding results, Firefox leaves the input intact when reloading a page / going back to a previous page
- The prettyname includes the class name if it is different from the module name `ol/source/Vector~VectorSource` instead of only `ol/source/Vector`
- To better display the longer names a zero-width space is inserted before '~' and '.' in class and module names to allow word wrapping when needed
- The search is only done on the string that is displayed to the user - that's the prettyname for classes / modules and the name for members / events.
- The icon and class name are displayed in a two-column flexbox layout so wrapped lines are indented nicely
- When search results are displayed clicking a class toggles between the matched members and all members instead of collapsing all

Results are more relevant when
- any word of the path starts with the search term +
- the actual short class / module name starts with the search term ++
- the start *and* end of the name are both at word boundaries +++
- the whole name matches from start to finish ++++

additionally there is a small preference to
- shorter names
- classes over modules

The speed improvements are achieved by
- *not* using jquery with each item in the search loop 
- keeping track of the matched items from the previous search

Feel free to try it and leave a comment.

### DIFF
--- a/config/jsdoc/api/plugins/api.js
+++ b/config/jsdoc/api/plugins/api.js
@@ -106,6 +106,20 @@ function includeTypes(doclet) {
   }
 }
 
+const defaultExports = {};
+const path = require('path');
+const moduleRoot = path.join(process.cwd(), 'src');
+
+// Tag default exported Identifiers because their name should be the same as the module name.
+exports.astNodeVisitor = {
+  visitNode: function(node, e, parser, currentSourceName) {
+    if (node.type === 'Identifier' && node.parent.type === 'ExportDefaultDeclaration') {
+      const modulePath = path.relative(moduleRoot, currentSourceName).replace(/\.js$/, '');
+      defaultExports['module:' + modulePath + '~' + node.name] = true;
+    }
+  }
+};
+
 exports.handlers = {
 
   newDoclet: function(e) {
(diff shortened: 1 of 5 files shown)


MoonE:
> - some entries are displayed with a wrong name e.g ol/interaction/DragBox~DragBoxEvent is also listed as ol/interaction/DragBox

This was part of the reason why I changed it.

I now changed it to only keep the class name for non-default exports. So
`ol/interaction/DragBox~DragBoxEvent` is still listed like this, but it is `ol/source/Vector` for both module and class instead of `ol/source/Vector~VectorSource` for the class.

PR #10722, opened 2020-02-22
fix: handle layer clear event in case clear(true) called

### LINKED ISSUE #10721: clear(true) doesn't work for WebGLPoints layer
**Describe the bug**
If you clear(true) on a vector source of a WebGLPoints layer, nothing happens as the clear event isn't handled.

(If you clear without opt_fast = true it works as the delete feature event is handled.)

**To Reproduce**
1. Add a few features to a VectorSource
2. Add a WebGLPointsLayer to the map using the VectorSource
3. Call vectorSource.clear(true) maybe on a timeout and notice that the layer doesn't clear as expected

```
...
    var vectorSource = new Vector();
...
    vectorSource.addFeatures(features);

    mapRef.current.addLayer(
      new WebGLPointsLayer({
        source: vectorSource,
        style: style 
      })
    );
...
    // it should clear but nothing happens after 3s
    setTimeout(() => vectorSource.clear(true), 3000);
```

**Expected behavior**
Layer should clear with opt_fast = true;

I think I have it fixed so I'll create a PR for this...

jellyedwards:
Should fix #10721 - tests pass.

Handle layer clear event in case clear(true) called as individual delete feature events aren't generated and the layer remains the same.

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -290,7 +290,8 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
     this.sourceListenKeys_ = [
       listen(source, VectorEventType.ADDFEATURE, this.handleSourceFeatureAdded_, this),
       listen(source, VectorEventType.CHANGEFEATURE, this.handleSourceFeatureChanged_, this),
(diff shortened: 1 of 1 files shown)


jellyedwards:
@jahow I made an absolute mess of the rebase, I think I've sorted it now, does this look ok to you?

jahow:
This is looking great, thanks a lot.

PR #10736, opened 2020-02-25
Allow cluster source to unlisten from its source

### LINKED ISSUE #10728: Cannot cleanly dispose of ol/source/Cluster
**Describe the bug**
The `ol/source/Cluster` adds an event listener to its source to be notified of changes.
But there is no way to unlisten when the `ol/source/Cluster` is disposed.

**To Reproduce**
1. Go to https://codesandbox.io/s/musing-blackwell-1dksx
I added a console output for the `ol/source/Cluster`'s refresh method which is called when the vecorsource changes.
2. The x button toggles between a `ol/source/Vector` or a new `ol/source/Cluster` with the same source
I create a new `ol/source/Cluster` because the clustering is done on refresh and thats expensive to do if it isn't needed.
3. The console always prints messages for all created `ol/source/Cluster`s because the eventlisteners cannot be removed.

**Expected behavior**
There should be a `setSource` method that accepts `null` to cleanly dispose of a `ol/source/Cluster`.

M393:
This adds a setSource method to change or remove the cluster source's source.
Fixes #10728

When the source is changed the cluster source and the features array are cleared and refresh is called.
It allows to create the cluster source without a source (not sure if that's useful).

### DIFF
--- a/src/ol/source/Cluster.js
+++ b/src/ol/source/Cluster.js
@@ -141,7 +141,7 @@ class Cluster extends VectorSource {
 
   /**
    * Replace the wrapped source.
-   * @param {VectorSource|null} source The new source for this instance.
+   * @param {VectorSource} source The new source for this instance.
    * @api
    */
   setSource(source) {
--- a/src/ol/source/Cluster.js
+++ b/src/ol/source/Cluster.js
@@ -29,7 +29,7 @@ import VectorSource from './Vector.js';
  * ```
  * See {@link module:ol/geom/Polygon~Polygon#getInteriorPoint} for a way to get a cluster
  * calculation point for polygons.
- * @property {VectorSource} [source=null] Source.
+ * @property {VectorSource} [source] Source.
  * @property {boolean} [wrapX=true] Whether to wrap the world horizontally.
  */
 
@@ -150,7 +150,7 @@ class Cluster extends VectorSource {
     }
     this.source = source;
     if (source) {
-      this.sourceListenKey_ = source.addEventListener(EventType.CHANGE, this.boundRefresh_);
+      source.addEventListener(EventType.CHANGE, this.boundRefresh_);
     }
     this.refresh();
   }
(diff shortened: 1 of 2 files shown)


src/ol/source/Cluster.js
```diff
@@ -132,7 +140,23 @@ class Cluster extends VectorSource {
   }
 
   /**
-   * handle the source changing
+   * Replace the wrapped source.
+   * @param {VectorSource|null} source The new source for this instance.
```
ahocevar: `null` is implicitly allowed, so no need to add it to the type.

src/ol/source/Cluster.js
```diff
@@ -29,7 +29,7 @@ import VectorSource from './Vector.js';
  * ```
  * See {@link module:ol/geom/Polygon~Polygon#getInteriorPoint} for a way to get a cluster
  * calculation point for polygons.
- * @property {VectorSource} source Source.
+ * @property {VectorSource} [source=null] Source.
```
ahocevar: The type is already correct without your change: `[source]` means that it is optional.

src/ol/source/Cluster.js
```diff
@@ -132,7 +140,23 @@ class Cluster extends VectorSource {
...
+   * @param {VectorSource|null} source The new source for this instance.
+   * @api
+   */
+  setSource(source) {
+    if (this.source) {
+      this.source.removeEventListener(EventType.CHANGE, this.boundRefresh_);
+    }
+    this.source = source;
+    if (source) {
+      this.sourceListenKey_ = source.addEventListener(EventType.CHANGE, this.boundRefresh_);
```
ahocevar: The way you register the listener, you do not need the `sourceListenKey_` to unregister.

M393: I first used on and unByKey to add and remove the listeners and this slipped through.

src/ol/source/Cluster.js
M393: I added null because it didn't show it as an allowed value in the apidoc otherwise.

src/ol/source/Cluster.js
M393: It should be changed to `[source]`, right?

ahocevar: Right.

src/ol/source/Cluster.js
ahocevar: There is no need to document that here, but maybe it should be documented somewhere else that `null` can be used to unset things, because that works everywhere in OpenLayers.

Issue #10734, opened 2020-02-24
Need to document tabindex behavior for MouseWheelZoom


matthias-ccri:
First of all, thank you for your work on openlayers!

I ran into this issue where I had to click the map before scroll-zoom worked. This had me scratching my head for a long time.

The [`MouseWheelZoom` documentation](https://openlayers.org/en/latest/apidoc/module-ol_interaction_MouseWheelZoom-MouseWheelZoom.html) says that the default condition is `module:ol/events/condition~always`. However, the code has a special case: (MouseWheelZoom.js#L122-L124)[https://github.com/openlayers/openlayers/blob/v6.2.1/src/ol/interaction/MouseWheelZoom.js#L122-L124]
```js
  conditionInternal_(mapBrowserEvent) {
    let pass = true;
    if (mapBrowserEvent.map.getTargetElement().hasAttribute('tabindex')) {
      pass = focus(mapBrowserEvent);
    }
    return pass && this.condition_(mapBrowserEvent);
  }
```
Basically, if a map has a `tabIndex`, then it must be focused before MouseWheelZoom works. For me, this meant that I had to 1) click inside the map, and 2) scroll to zoom. 

It looks like this PR ([tabindex without focus condition #10291](https://github.com/openlayers/openlayers/pull/10291)) introduced the special case, and DragPan interaction is also affected. I don't know if other interactions have the special case.

I would like to contribute this fix to the docs! Should the `options.condition` documentation be modified like this?: 
> A function that takes a `module:ol/MapBrowserEvent~MapBrowserEvent` and returns a boolean to indicate whether that event should be handled. Default is `module:ol/events/condition~always`. In addition, if there is a `tabindex` attribute on the map element, `module:ol/events/condition~focus` will also be applied.

jipexu:
HI
i dont know if i understand well 
but this can be usefull ??

         interactions: new ol.interaction.defaults({
            constrainResolution: true, onFocusOnly: true
         }),

matthias-ccri:
Hi @jipexu my case was the opposite. I did not want "on focus only", and I did not have `onFocusOnly: true`. The documentation states "default false". But basically, the default is _not_ false if your map has a `tabindex` attribute.

jipexu:
hi @matthias-ccri oups sorry i misunderstood ...

ahocevar:
@matthias-ccri Thanks for reporting this. A pull request with your suggested documentation fix would be much appreciated.

PR #10740, opened 2020-02-27
Document tabindex behavior for MouseWheelZoom and DragPan

matthias-ccri:
Addresses https://github.com/openlayers/openlayers/issues/10734


tschaub:
Thanks, @matthias-ccri.

PR #10744, opened 2020-02-27
Ensure the Modify Features Test example opens at correct zoom

mike-000:
The default projection isn't ideal for this example, but its side affects can be averted by setting `multiWorld: true` so the view opens at zoom 2 regardless of display width.


mike-000:
Perhaps the same should apply to the Dynamic Data and GeoJSON examples? (although only for visual effect as there is no interaction outside the opening view).

PR #10746, opened 2020-02-28
Fix building apidoc on windows

MoonE:
The apidocs cannot be built on windows because the file path is not normalized to have `/` as path separator.

Sorry for that.


MoonE:
Thank you for your time!

PR #10748, opened 2020-02-28
Fix "Cannot read property 'anchor' of undefined" in ol/View

mike-000:
This fixes a timing issue in #10120 where all animations have completed when there is a cancellation (in this case caused by using setRotation to keep an EPSG:3857 view aligned to EPSG:27700 grid north if there is a significant change while panning) resulting in an error.  Using more robust code will fix it.

image


tschaub:
Thanks, @mike-000.

PR #10758, opened 2020-03-03
Allow using feature attributes for symbol rotation in WebGL layers

jahow:
Issue was reported [here](https://github.com/openlayers/openlayers/pull/10439#issuecomment-593476693).

Previously when using feature attributes in a rotation expression, the resulting rotation would always be zero.


jahow:
The fix can be tried here:
https://3337-4723248-gh.circle-artifacts.com/0/examples/webgl-points-layer.html

With the following style:
```json
{
  "symbol": {
    "symbolType": "square",
    "rotation": [
     "get", "population"
    ],
    "size": [
      "array",
      4,
      [
        "interpolate",
        [
          "linear"
        ],
        [
          "get",
          "population"
        ],
        20000,
        4,
        300000,
        28
      ]
    ],
    "color": [
      "interpolate",
      [
        "linear"
      ],
      [
        "get",
        "population"
      ],
      20000,
      "#ffdc00",
      300000,
      "#ff5b19"
    ],
    "offset": [
      "array",
      0,
      [
        "interpolate",
        [
          "linear"
        ],
        [
          "get",
          "population"
        ],
        20000,
        2,
        300000,
        14
      ]
    ]
  }
}
```

PR #10763, opened 2020-03-04
TypeScript: Fix inconsistent optionality in various APIs

### LINKED ISSUE #10762: TypeScript: Inconsistent optionality of various APIs
**Describe the bug**

The following `ImageSource` sub-classes have **required** `options.projection` which contradicts the optionality of `options.projection` in the parent `ImageSource` constructor itself.

 * `ImageCanvas`
 * `ImageMapGuide`
 * `ImageStatic`

Also the `GeometryFunction` typedef generates an invalid function signature in TypeScript:

`ol/interaction/Draw.d.ts`

```
export type GeometryFunction = (arg0: number[] | number[][] | number[][][], arg1?: import("../geom/SimpleGeometry.js").default, arg2: import("../proj/Projection.js").default) => import("../geom/SimpleGeometry.js").default;
```

Note how `arg2` is required after the optional `arg1`

**To Reproduce**

`new` any of the above image sources. You are *required* by the TypeScript compiler to specify a value for the `projection` option.

**Expected behavior**

This property should be optional.

jumpinjackie:
Fixes #10762

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -110,7 +110,7 @@ import {fromUserCoordinate, getUserProjection} from '../proj.js';
  * geometry is the geometry that is returned when the function is called without
  * a second argument.
  * @typedef {function(!SketchCoordType, import("../geom/SimpleGeometry.js").default=,
- *     import("../proj/Projection.js").default):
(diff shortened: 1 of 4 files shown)


tschaub:
Thanks, @jumpinjackie.

PR #10787, opened 2020-03-09
Improved projection extent in the "Reprojection with EPSG.io Search" example

mike-000:
The enhanced extent transforms introduced in #10723 allow the previously _"very approximate calculation of projection extent"_ (which for some polar projections resulted in extents with zero width or height) to be made more realistic.

Also after basing a test bed for the graticule updates on this example I think a graticule can help to illustrate a projection so I have added a "Show graticule" checkbox option in addition to the "Render edges" option.


tschaub:
Thanks, @mike-000.

Issue #10747, opened 2020-02-28
Crash on touch events on IOS since 6.2.0


sosmo:
**Describe the bug**
Panning the map or using gesture zoom on IOS 12 causes OL to crash.

> Cannot read property 'defaultPrevented' of undefined

at MapBrowserEventHandler https://github.com/openlayers/openlayers/blob/88b8b2f7cb95fe6c728767dc5d1194bc04139b60/src/ol/MapBrowserEventHandler.js#L278

The bug was introduced at version 6.2.0. The same doesn't happen on Chrome on PC or Android. It's likely related to commit https://github.com/openlayers/openlayers/commit/88b8b2f7cb95fe6c728767dc5d1194bc04139b60?diff=unified

MapBrowserEventHandler expects "this.originalPointerMoveEvent_" to be set when "handleTouchMove_" gets called, but initially the variable is uninitialized. It should get set at "relayEvent_", but it only gets called after a pointermove event. It seems on IOS Safari "handleTouchMove_" gets called before the listener that fires "relayEvent_", and thus "originalPointerMoveEvent_" is still undefined. "handleTouchMove_" does get called later, it's just the initial call that crashes the app. Thus adding a null check before using "originalPointerMoveEvent_" at this line could be enough to fix the issue

https://github.com/openlayers/openlayers/commit/88b8b2f7cb95fe6c728767dc5d1194bc04139b60?diff=unified#diff-0833750a228e1622e02bcc1779738e7aR278

ahocevar:
@sosmo Thanks for sharing your observation. I don't have an iOS 12 device to test. Can you please create a pull request with your suggested fix?

sosmo:
Thank you for the answer! I tried exploring whether the null check would be enough to cover all bases, but now it seems there's other issues too

Everything works perfectly when I try running the local examples from the repo with `npm start`, even the latest 6.2.2 version. I wasn't even able to reproduce the iOS crash when running from the source js files directly. However, when I tried installing 6.2.1 as an npm module again, it caused a crash on iOS again, and then I noticed that panning the map doesn't work, even on desktop Chromium or Firefox. This didn't happen when I first reported the bug!

Here's a minimal example: https://codesandbox.io/s/loving-resonance-wqcqm [wqcqm.zip](https://github.com/openlayers/openlayers/files/4308074/wqcqm.zip)

[...]

PR #10788, opened 2020-03-09
Add safeguard to handleTouchMove

sosmo:
Fixes https://github.com/openlayers/openlayers/issues/10747

The related issue describes other, possibly related problems. As such this may be just a patch rather than a proper fix

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -275,7 +275,8 @@ class MapBrowserEventHandler extends EventTarget {
    * @private
    */
   handleTouchMove_(event) {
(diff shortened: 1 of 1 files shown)


src/ol/MapBrowserEventHandler.js
```diff
@@ -275,7 +275,8 @@ class MapBrowserEventHandler extends EventTarget {
    * @private
    */
   handleTouchMove_(event) {
-    if (this.originalPointerMoveEvent_.defaultPrevented) {
+    // In some environments (such as iOS 12), this handler can get called once before the `originalPointerMoveEvent_` is initialized. This prevents crashes in those situations (but may not work correctly if `originalPointerMoveEvent_` lags behind the current event?).
```
ahocevar:
```suggestion
    // Due to https://github.com/mpizenberg/elm-pep/issues/2, `this.originalPointerMoveEvent_`
    // may not be initialized yet when we get here on a platform without native pointer events.
```

Issue #10796, opened 2020-03-11
ol/source/Vector: removeFeature fails even though hasFeature returns true


icholy:
**Describe the bug**

If two features have the same id, it's impossible to correctly use `hasFeature` with `removeFeature`. `hasFeature` checks for both feature id and uid while `removeFeature` only uses the feature uid.

**To Reproduce**

``` ts
let a = new Feature();
a.setId("123");

let b = new Feature();
b.setId("123");

source.addFeature(a);

if (source.hasFeature(b)) {
  source.removeFeature(b);
  // Uncaught TypeError: Cannot read property 'forEach' of undefined
}
```

**Expected behavior**

* `hasFeature` only returns true if the feature uid matches.
* ~~have a separate `hasFeatureId` method.~~

icholy:
``` js
function hasFeature(source, feature) {
  let id = feature.getId();
  if (id) {
    return source.getFeatureById(id) === feature;
  }
  return source.hasFeature(feature);
}
```

icholy:
One way to work around this would be to do an equality check against the feature returned by `getFeatureById`:

``` js
function hasFeature(source, feature) {
  let id = feature.getId();
  if (id) {
    return source.getFeatureById() === feature;
  }
  return source.hasFeature(feature);
}
```

ahocevar:
Two distinct features should not have the same ids. This is enforced in the source's `addFeature` method, but that restriction can be circumvented by using `feature.setId`. The proper fix would be for the source to throw an exception upon handling a feature's change event when the id is set to the id of an existing feature.

icholy:
@ahocevar In my use-case I have the same feature geojson arriving from two separate places. Even though they're the "same" feature, they're not the same instance of `ol.Feature`.

ahocevar:
@icholy If the two GeoJSONs are loaded into the same source, then features that have the same id as a previous feature will not be added again.

icholy:
@ahocevar I wanted the opposite behaviour, so I'm doing this:

``` js
function addFeature(source, feature) {
  let id = feature.getId();
  if (id !== undefined) {
    let duplicate = source.getFeatureById(id);
    if (duplicate) {
      source.removeFeature(duplicate);
    }
  }
  source.addFeature(feature);
}
```

However, now the removed (original) feature has no way of knowing if it's in the source.

icholy:
Another option is to prevent `removeFeature` from throwing an exception if the source doesn't contain the `io.Feature` instance.

PR #10797, opened 2020-03-11
Add hasFeatureId to the Vector source

icholy:
#10796


icholy:
Now that I think about it, `hasFeatureId` can be trivially implemented using `getFeatureById`.

ahocevar:
You can also just *use* getFeatureById.

icholy:
@ahocevar yes, that's what I meant to say. My original reasoning was to re-add the functionality removed from `hasFeature`.

Issue #10741, opened 2020-02-27
Expose Overlay.panIntoView as an API Method


ejn:
**Is your feature request related to a problem? Please describe.**
It is possible to make the map pan so that an overlay is contained in the viewport when the position is changed (#3256), but it is not possible to pan the map to show the overlay at a later time without some form of workaround (e.g. changing the overlay position, creating a new overlay on the same position, etc.). This would also provide a reasonable solution to #7083 by allowing the application to request a pan explicitly when the overlay content / element size has been changed.

**Describe the solution you'd like**
1. Upgrade the `panIntoView()` method on `Overlay` to be an API method
2. Modify `panIntoView()` to accept custom values for the animation and the margin as parameters (`panIntoView(panOptions: PanOptions, margin: number)`), defaulting to `20` and `{}` (same defaults as the `autoPan` defaults) - the `autoPan`-settings will however not be used here.
3. Add a new protected `autopan()` convenience wrapper method to call `panIntoView()` with `autoPanAnimation` and  `autoPanMargin`
4. Set `handlePositionChanged()` to call the new `autopan()` wrapper instead of `panIntoView()`

I'm happy to hack up a PR for this if the suggested solution sounds acceptable.

ejn:
Implemented in #10800 (now merged - thanks @ahocevar!)

PR #10800, opened 2020-03-12
Make Overlay.panIntoView an API method

ejn:
* Add parameters for custom animation parameters and margin

 * Add protected wrapper method for previous autoPan functionality

See #10741

### DIFF
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -378,7 +378,7 @@ class Overlay extends BaseObject {
   }
 
   /**
-   * Pan the map so that the overlay is entirely visisbly in the current viewport
+   * Pan the map so that the overlay is entirely visisble in the current viewport
    * (if necessary) using the configured autoPan parameters
    * @protected
    */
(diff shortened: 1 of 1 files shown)


tschaub:
It could be nice to have the function take a single `options` object.  This would have `animation` and `margin` properties and would allow more options to be added in the future without an awkwardly long signature.

To do this, you would create a new type as is done with the existing constructor options.

ejn:
I agree. It would also then make sense to alter the autoPan options to use the new type rather than being two separate parameters to the constructor (and two internal properties).

ejn:
so autoPanOptions in the constructor options instead of autoPanAnimation and autoPanMargin, and then carry the change through the whole class? autoPanAnimation and autoPanMargin could be left in as deprecated options for backward compatibility to avoid breaking the API.

ejn:
@tschaub the test failures (rendering test) appear to be unrelated to this change: Could you force the build to run again?

ejn:
@ahocevar Nice and obvious idea - should have thought of that myself!

PR #10805, opened 2020-03-13
make ImageSourceEventType available for consumers

regileeso:


regileeso:
`@types/ol` have it available, so `tsc` says everything's ok. The actual build fails because the type is just not exported.

regileeso:
Fixes #10814

PR #10807, opened 2020-03-14
Handle Graticule wrapX without calculating excess meridians

### LINKED ISSUE #10799: Missing meridians in zoomed in wrapped graticule
**Describe the bug**
I noticed the problem while testing #10795 but it does not appear to be related to that or other recent changes and can also be seen in version 6.2.1.

**To Reproduce**
Open the example https://openlayers.org/en/latest/examples/graticule.html in code sandbox and comment out both `wrapX: false` lines.  Pan from the "normal" world to an adjacent world and zoom in.  Instead of the interval continuing to reduce similar to the parallels the meridians disappear below an interval of 1 degree.  In the "normal" world the meridians do not disappear.

**Expected behavior**
Meridians should appear at all intervals in all worlds.

mike-000:
Fixes #10799

When wrapped the renderer was calling the loader function with the full width of the projection resulting in the graticule attempting to calculate meridians for the whole world even when zoomed in on a small region.  This exceeded the `maxLines` limit and resulted in missing meridians, unnecessarily long parallels and poor performance.  It is fixed by the vector layer renderer always calling the loader with the frameState extent for Graticule layers and handling wrapping in the graticule.

### DIFF
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -415,7 +415,7 @@ export function toStringXY(coordinate, opt_fractionDigits) {
  */
 export function wrapX(coordinate, projection) {
   const projectionExtent = projection.getExtent();
-  if (projection.canWrapX() && (coordinate[0] < projectionExtent[0] || coordinate[0] > projectionExtent[2])) {
+  if (projection.canWrapX() && (coordinate[0] < projectionExtent[0] || coordinate[0] >= projectionExtent[2])) {
     const worldWidth = getWidth(projectionExtent);
     const worldsAway = Math.floor((coordinate[0] - projectionExtent[0]) / worldWidth);
     coordinate[0] -= (worldsAway * worldWidth);
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -407,7 +407,8 @@ export function toStringXY(coordinate, opt_fractionDigits) {
 
 /**
  * Modifies the provided coordinate in-place to be within the real world
- * extent.
+ * extent. The lower projection extent boundary is inclusive, the upper one
+ * exclusive.
  *
  * @param {Coordinate} coordinate Coordinate.
  * @param {import("./proj/Projection.js").default} projection Projection
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -3,6 +3,7 @@
  */
 import {modulo} from './math.js';
 import {padNumber} from './string.js';
+import {getWidth} from './extent.js';
 
 
 /**
@@ -401,4 +402,23 @@ export function toStringHDMS(coordinate, opt_fractionDigits) {
  */
 export function toStringXY(coordinate, opt_fractionDigits) {
   return format(coordinate, '{x}, {y}', opt_fractionDigits);
+}
+
+
+/**
+ * Modifies the provided coordinate in-place to be within the real world
+ * extent.
+ *
+ * @param {Coordinate} coordinate Coordinate.
+ * @param {import("./proj/Projection.js").default} projection Projection
+ * @return {Coordinate} The coordinate within the real world extent.
+ */
+export function wrapX(coordinate, projection) {
+  const projectionExtent = projection.getExtent();
+  if (projection.canWrapX() && (coordinate[0] < projectionExtent[0] || coordinate[0] > projectionExtent[2])) {
+    const worldWidth = getWidth(projectionExtent);
+    const worldsAway = Math.floor((coordinate[0] - projectionExtent[0]) / worldWidth);
+    coordinate[0] -= (worldsAway * worldWidth);
+  }
+  return coordinate;
 }
--- a/src/ol/extent.js
+++ b/src/ol/extent.js
@@ -813,3 +813,25 @@ export function applyTransform(extent, transformFn, opt_extent, opt_stops) {
   }
   return _boundingExtentXYs(xs, ys, opt_extent);
 }
+
+
+/**
+ * Modifies the provided extent in-place to be within the real world
+ * extent.
+ *
+ * @param {Extent} extent Extent.
+ * @param {import("./proj/Projection.js").default} projection Projection
+ * @return {Extent} The extent within the real world extent.
+ */
+export function wrapX(extent, projection) {
+  const projectionExtent = projection.getExtent();
+  const center = getCenter(extent);
+  if (projection.canWrapX() && (center[0] < projectionExtent[0] || center[0] > projectionExtent[2])) {
+    const worldWidth = getWidth(projectionExtent);
+    const worldsAway = Math.floor((center[0] - projectionExtent[0]) / worldWidth);
+    const offset = (worldsAway * worldWidth);
+    extent[0] -= offset;
+    extent[2] -= offset;
+  }
+  return extent;
+}
--- a/src/ol/extent.js
+++ b/src/ol/extent.js
@@ -295,6 +295,18 @@ export function equals(extent1, extent2) {
       extent1[1] == extent2[1] && extent1[3] == extent2[3];
 }
 
+/**
+ * Determine if two extents are approximately equivalent.
+ * @param {Extent} extent1 Extent 1.
+ * @param {Extent} extent2 Extent 2.
+ * @param {number} tolerance Tolerance in extent coordinate units.
+ * @return {boolean} The two extents differ by less than the tolerance.
+ */
+export function approximatelyEquals(extent1, extent2, tolerance) {
+  return Math.abs(extent1[0] - extent2[0]) < tolerance && Math.abs(extent1[2] - extent2[2]) < tolerance &&
+  Math.abs(extent1[1] - extent2[1]) < tolerance && Math.abs(extent1[3] - extent2[3]) < tolerance;
+}
+
 
 /**
  * Modify an extent to include another extent.
--- a/src/ol/extent.js
+++ b/src/ol/extent.js
@@ -826,7 +826,7 @@ export function applyTransform(extent, transformFn, opt_extent, opt_stops) {
 export function wrapX(extent, projection) {
   const projectionExtent = projection.getExtent();
   const center = getCenter(extent);
-  if (projection.canWrapX() && (center[0] < projectionExtent[0] || center[0] > projectionExtent[2])) {
+  if (projection.canWrapX() && (center[0] < projectionExtent[0] || center[0] >= projectionExtent[2])) {
     const worldWidth = getWidth(projectionExtent);
     const worldsAway = Math.floor((center[0] - projectionExtent[0]) / worldWidth);
     const offset = (worldsAway * worldWidth);
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -18,7 +18,7 @@ import {
   applyTransform,
   containsCoordinate,
   containsExtent,
-  equals,
+  approximatelyEquals,
   getCenter,
   getHeight,
   getIntersection,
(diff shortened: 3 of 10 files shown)


ahocevar:
The changes to the renderer are problematic. If the Graticule layer handles wrapX, wouldn't it be enough to configure its source always with `wrapX: false`?

mike-000:
The graticule only handles wrapX in as much as it moves the provided extent to a normal world (possibly overlapping another world) to avoid unnecessary loader processing.  The renderer's replay groups and renderFrame are not affected and will need a wrapX source to render that onto other worlds.

ahocevar:
I see @mike-000. In this case, it would make sense to move the +/-540° logic from the renderer to `ol/source/Vector`'s `loadFeatures()` method, so it would not affect the Graticule if it uses a customized vector source.

mike-000:
Even custom loaders are called from `loadFeatures` so that would be a breaking change for anything which used a custom loader.  Maybe instead of hardcoding a test for a specific layer type a new source property is needed to indicate which extent is passed to loaders when wrapped.  It might also be useful for loading WFS features in small areas such as Fiji which would otherwise request data from across the world.

ahocevar:
@mike-000 That sounds good. What about a `getLoadExtent()` method on the source?

ahocevar:
@mike-000 Thanks for your continued effort on this. I'm inclined to always pass an extent with a real world center to the loader, regardless of whether the source is a graticule or a plain vector layer. If you agree, I'll add a commit to your branch to show you what I mean.

mike-000:
Yes, the loader having the majority of the extent in a real world makes sense even if it wraps onto another, and if entirely in one world it wouldn't need to be concerned about which one.

ahocevar:
@mike-000 Ok, committed.

One thing I noticed (but that might be an issue beyond the scope of this pull request) is that when `wrapX` is set to true in the `graticule.html` example, no meridian labels will be shown outside the real world.

mike-000:
That is because this PR was started before #10795 was merged.  These changes are not related and should be no conflict with that.

mike-000:
I suspect the removal of `loadWrapX` is going to be a breaking change. In this case always basing the load extent on the frameStateExtent would result in an approximate request for either -184 to -176 or +176 to +184 (depending on the exact position of the center).  Will all data sources return results from both sides of the dateline for such requests?  If not the `loadWrapX` is still needed to get all results from -180 to +180 (which might results in a lot of excess data from a global source) or the loader would need to make two requests, one on each side of the dateline.
image

ahocevar:
That's a valid concern @mike-000. I added another commit so now two extents are loaded in these cases. You may want to take another final look at your changes to make sure everything still works.

src/ol/layer/Graticule.js
```diff
@@ -479,11 +479,25 @@ class Graticule extends VectorLayer {
...
+    if (this.projection_) {
+      const center = getCenter(extent);
+      const projectionExtent = this.projection_.getExtent();
+      const worldWidth = getWidth(projectionExtent);
+      if (this.getSource().getWrapX() && this.projection_.canWrapX() && !containsExtent(projectionExtent, extent)) {
+        const worldsAway = Math.floor((center[0] - projectionExtent[0]) / worldWidth);
+        realExtent[0] -= (worldsAway * worldWidth);
+        realExtent[2] -= (worldsAway * worldWidth);
+      }
+      realExtent[0] = Math.round(realExtent[0] * 1e6) / 1e6;
```
ahocevar: Better use the resolution to round to pixels, instead of hard-coding `1e6`.

src/ol/layer/Graticule.js
```diff
@@ -479,11 +479,25 @@ class Graticule extends VectorLayer {
...
-    if (this.loadedExtent_ && !equals(this.loadedExtent_, extent)) {
+    // extents may be passed in different worlds, to avoid endless loop we use only one
+    const realExtent = extent.slice();
+    if (this.projection_) {
+      const center = getCenter(extent);
+      const projectionExtent = this.projection_.getExtent();
+      const worldWidth = getWidth(projectionExtent);
+      if (this.getSource().getWrapX() && this.projection_.canWrapX() && !containsExtent(projectionExtent, extent)) {
+        const worldsAway = Math.floor((center[0] - projectionExtent[0]) / worldWidth);
+        realExtent[0] -= (worldsAway * worldWidth);
```
ahocevar: We're doing this same operation a lot. I think a `toRealWorld(extent, projection)` function in `ol/extent` could be nice.

src/ol/layer/Graticule.js
mike-000: I don't think it is related to resolution. There is a javascript rounding problem when adding or subtracting multiples of the width which could happen at any resolution breaking the equals check and therefore not fixing the endless loop
image

src/ol/layer/Graticule.js
mike-000: Once in the renderer and once in the graticule.  Is it being done anywhere else?

ahocevar: I added another commit that adds wrapX functions for extent and coordinate. With that commit, I also simplified the solution your rounding problem like I suggested.

Can you please take another final look?

mike-000: The wrapX functions needed to allow `worldsAway` to be calculated when the center is at +180 otherwise the graticule loops when the view is centered on the dateline.  The strategy must test for `!equal` as any real extent change needs the graticule to be updated.  With `!containsExtent` the graticule intervals do not update when zooming it.  I cannot see any alternative to rounding as the problem is similar to JavaScript 0.2 + 0.1 and not related to resolution.

ahocevar: I know that it is not *related* to resolution, but *using* resolution for rounding is nicer than arbitrary numbers. In the end what you need is pixel exact extents, and that can be achieved with rounding by resolution (i.e. 1 pixel).

src/ol/extent.js
```diff
@@ -826,7 +826,7 @@ export function applyTransform(extent, transformFn, opt_extent, opt_stops) {
 export function wrapX(extent, projection) {
   const projectionExtent = projection.getExtent();
   const center = getCenter(extent);
-  if (projection.canWrapX() && (center[0] < projectionExtent[0] || center[0] > projectionExtent[2])) {
+  if (projection.canWrapX() && (center[0] < projectionExtent[0] || center[0] >= projectionExtent[2])) {
```
ahocevar: Please elaborate why you think this is necessary.

src/ol/layer/Graticule.js
```diff
@@ -484,9 +483,11 @@ class Graticule extends VectorLayer {
     // extents may be passed in different worlds, to avoid endless loop we use only one
     const realExtent = extent.slice();
     if (this.projection_ && this.getSource().getWrapX()) {
-      wrapX(realExtent, this.projection_);
+      wrapExtentX(realExtent, this.projection_);
     }
-    if (this.loadedExtent_ && !containsExtent(bufferExtent(this.loadedExtent_, resolution / 2), realExtent)) {
+    realExtent[0] = Math.round(realExtent[0] * 1e8) / 1e8;
+    realExtent[2] = Math.round(realExtent[2] * 1e8) / 1e8;
```
ahocevar:
```suggestion
    realExtent[0] = Math.round(realExtent[0] * resolution) / resolution;
    realExtent[2] = Math.round(realExtent[2] * resolution) / resolution;
```
This would round the extent to pixels, which should be accurate enough.

src/ol/extent.js
mike-000: If you have two extents with centers at -180 and +180 they should be moved to the same real world (as +540 would be) but the with > that didn't happen, the graticule thought it had different extents and went into a loop.

src/ol/layer/Graticule.js
```diff
@@ -484,9 +483,11 @@ class Graticule extends VectorLayer {
...
+    if (this.loadedExtent_ && !equals(this.loadedExtent_, realExtent)) {
```
ahocevar:
```suggestion
    if (this.loadedExtent_ && !equals(this.loadedExtent_, realExtent.map(function(n) { return Math.round(n * resolution) / resolution; })) {
```
In combination with the above suggestion, this would round the extent to pixels, which should be accurate enough and would avoid the arbitrary `1e8` figure.

src/ol/coordinate.js
```diff
@@ -415,7 +415,7 @@ export function toStringXY(coordinate, opt_fractionDigits) {
  */
 export function wrapX(coordinate, projection) {
   const projectionExtent = projection.getExtent();
-  if (projection.canWrapX() && (coordinate[0] < projectionExtent[0] || coordinate[0] > projectionExtent[2])) {
+  if (projection.canWrapX() && (coordinate[0] < projectionExtent[0] || coordinate[0] >= projectionExtent[2])) {
```
ahocevar: Fine for the extent's center, but here I think we do not want to include the right edge.

src/ol/layer/Graticule.js
mike-000: I also tried 

    if (this.loadedExtent_ && (
      !containsExtent(bufferExtent(this.loadedExtent_, resolution / 2), realExtent) ||
      !containsExtent(bufferExtent(realExtent, resolution / 2), this.loadedExtent_)
    )) {


but that fails when resolutions become so small relative to large coordinate values that buffering has no effect.

Any rounding has the potential for a different last digit.  A more general approach which would avoid that and might be useful elsewhere would be a near equality method similar to that used in the tests which checks values are within JavasScript's approx 15 digit accuracy e.g.

`Math.abs(a - b) <= Math.abs((a + b) * 1e-15)`

src/ol/extent.js
ahocevar: Ok, thanks for the explanation. I think this fine for extent centers, but not for coordinates.

src/ol/layer/Graticule.js
ahocevar: What do you use to test this pull request? I.e. where do you check for the infinite loops you mentioned several times? Which example or test case?

src/ol/coordinate.js
mike-000: There is a similar problem with these extents [-360, -90, 0, 90] [0, -90, 360, 90] [360, -90, 720, 90] as with the center.  They should be all produce the same real world extent but [0, -90, 360, 90] doesn't.

ahocevar: But `ol/coordinate#wrapX()` is not involved in `ol/extent#wrapX`, and also not in `ol/layer/Graticule`. So I wonder why you insist on this change in `ol/coordinate`.

ahocevar: I am fine with your change in `ol/extent#wrapX`, if that was not clear.

mike-000: `ol/coordinate#wrapX` is used in `prepareFrame` where it should be consistent with the extent

src/ol/layer/Graticule.js
mike-000: I set up a sandbox https://codesandbox.io/s/simple-11f0s patched with the changes and logging the extents used in the strategy function to debug the looping.  The opening at +180 problem is easily reproduced by changing `>=` to `>`  The rounding problems occur intermittently if the rounding is commented out.

src/ol/coordinate.js
ahocevar: Good point.

src/ol/layer/Graticule.js
ahocevar: If you could turn that into a unit or rendering test, it would be great.

ahocevar:
I think this is good now. I added an `approximatelyEquals` function for extents and had to rebase this branch because of a conflict.

Note that I set the tolerance for the `approximatelyEquals` function to the resolution (i.e. the allowed tolerance is be 1 pixel). That worked fine with your sandbox. If there are still concerns, please add tests (unit or rendering) that show the problem.

If this looks good to you now, we can merge it.

mike-000:
All looks good.  I hadn't noticed your change to loader when I tested in the sandbox so rearranged the strategy to ensure the loader can use a simple equals comparison.

Issue #10826, opened 2020-03-22
Send lon & lat to lonLabelFormatter & latLabelFormatter in ol/Graticule


Sagie501:
The current map Graticule options lonLabelFormatter and latLabelFormatter functions get only one argument: lon or lat respectively.
Sometimes we would like to have both of the arguments for formatting the lon / lat.
For example, if we want to use different coordinate system for displaying the labels etc.

alaluf99:
Great solution.

SharonShrits:
Very helpful, thank you!

PR #10827, opened 2020-03-22
Add the ability to get both lon and lat arguments

Sagie501:
Add the ability to get both longitude and latitude arguments to lonLabelFormatter & latLabelFormatter function in Graticule.
Issue #10826


mike-000:
How does the second argument relate to the label position?  If you need more information on a label it would be simpler and more accurate to pass the `textpoint` calculated in `drawLabels_` by making the second property of `GraticuleLabelDataType` a `number` and deferring the calls to the formatters until `drawLabels_`

PR #10840, opened 2020-03-25
Remove inherit doc

### LINKED ISSUE #10761: FYI: @inheritDoc produces incorrect method signatures in TypeScript
**Describe the bug**

Trying out the new typings bundled with OL `6.2.2-dev.1583236579811`

Any derived class method decorated with `@inheritDoc` that involves optional parameters are generated incorrectly by the TypeScript compiler (the optionality is lost on the derived class methods)

**To Reproduce**

```ts
import LayerGroup from 'ol/layer/Group';

const group = new LayerGroup();
group.getLayersArray(); //TS error: The "opt_array" argument is required despite being documented as optional in the parent
group.getLayersArray(undefined); //Current workaround

```

**Expected behavior**

The optionality should be preserved. I've identified this as a problem in the TypeScript compiler's JSDoc processing and have filed an appropriate issue there: https://github.com/microsoft/TypeScript/issues/37210

I don't know if this issue can be reliably fixed/worked-around downstream.

This issue is more a reminder if/when this issue is fixed upstream, that you should adopt the next version of TypeScript that has the fix incorporated.

fredj:
fixes #10761 

This pull request remove the `@inheritDoc` tags.

I had to change some `{Node}` to `{Element}` to make the checker happy. There's also a small change in ESRIJson `readFeatureFromObject`: the `objectIdFieldName` property is now passed as a new parameter instead of being added to the read options


tschaub:
@fredj - this seems like it has good potential.  How much more work do you think is needed to get it ready for review?

I want to make sure we don't loose good work by creating nasty conflicts in #10862.  But if this is not going to be ready without a lot more work, that would be good to know.

fredj:
@tschaub this is not 100% done but I think it can be merged as it. I can open a new PR after that to finish the work,

tschaub:
Thanks for all the work on this, @fredj!

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -167,7 +171,8 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
   }
 
   /**
-   * @inheritdoc
```
tschaub, line 170: Tricky spelling!

tschaub:
It looks like the plugin for `@inheritDoc` can be removed too.  Feel free to cherry pick https://github.com/tschaub/openlayers/commit/6a8f49904ec3b5c2a3c7648f3f6d690b6b5a39f0, to add your own commit, or to merge as is and we can remove it later.  The generated docs look the same with and without the plugin.

fredj:
Thanks for the review @tschaub

PR #10845, opened 2020-03-26
Fix createHitDetectionImageData error for features with no size

### LINKED ISSUE #10839: createHitDetectionImageData error for features with no size
```
this.layer.getFeatures(pixel).then((features) => { });
```

triggers an error in `createHitDetectionImageData` when there is a feature with an image with no size. 

for some reason the array here is undefined:

https://github.com/openlayers/openlayers/blob/5ea0b52ba8774e8170dc18fb6453affdef43eafc/src/ol/render/canvas/hitdetect.js#L69

gedaiu:
fix #10839


ahocevar:
Thanks, @gedaiu

Issue #10823, opened 2020-03-19
Geoserver Filter Functions


glenselle:
**Is your feature request related to a problem? Please describe.**
Some of the WMS layers we work with have data represented in comma-delimited lists which we allow users to filter by. We currently support WMS filtering by using a `cql_filter` query param that looks something like the following (slightly simplified for example purposes):

```
cql_filter=strMatches(tags, '^mycustomtag( )??,.*') = TRUE OR strMatches(tags, '.*,( )??mycustomtag( )??,.*') = TRUE OR strMatches(tags , '.*,( )??mycustomtag( )??$') = TRUE OR tags = 'mycustomtag'
```

This looks atrocious in code but works really well for users. Features on the layer have an attribute called `tags` which then has a value like `tag1, something2, mycustomtag`. The layer can then be filtered by any of those tags.

However, when we attempt to do the same thing with a WFS `GetFeature` request we have to create an XML payload with a `<Function>` inside of a `<Filter>`. OpenLayers ships with a set of filters but `<Function>` doesn't appear to be in the list or the source (unless we missed it). Interestingly, OpenLayers v2 seems to have this feature: http://dev.openlayers.org/docs/files/OpenLayers/Filter/Function-js.html

**Describe the solution you'd like**
We're interested in adding support for OGC filter functions within OpenLayers and are willing to submit a PR to add this, if the maintainers think this is reasonable & there's not currently a way to do this. Here's how I was envisioning this work:

```javascript
import filter from 'ol/format/filter'

const customFilter = filter.or(
  filter.like(filter.function('strMatches', 'tags', '.*,( )??mycustomtag( )??$'), true, '*', '.', '!'),
  filter.like(filter.function('strMatches', 'tags', '.*,( )??mycustomtag( )??,.*'), true, '*', '.', '!')
)
```
would then create:
```xml
<Or>
  <PropertyIsLike wildCard="*" singleChar="." escape="!">
    <Function name="strMatches">
      <PropertyName>tags</PropertyName>
      <Literal>.*,( )??mycustomtag( )??$</Literal>
    </Function>
    <Literal>true</Literal>
  </PropertyIsLike>
  <PropertyIsLike wildCard="*" singleChar="." escape="!">
    <Function name="strMatches">
      <PropertyName>tags</PropertyName>
      <Literal>.*,( )??mycustomtag( )??,.*</Literal>
    </Function>
    <Literal>true</Literal>
  </PropertyIsLike>
</Or>
```

[...]

PR #10846, opened 2020-03-26
Initial work to add filter function support for IsLike

glenselle:
This PR is first take on the work needed to make https://github.com/openlayers/openlayers/issues/10823 possible. In order to add support for this internally on our project we created a new WFS subclass which made these changes. I would like to make the check when it writes the filter more robust and also add tests.

### DIFF
--- a/src/ol/format/WFS.js
+++ b/src/ol/format/WFS.js
@@ -1091,8 +1091,18 @@ function writeIsLikeFilter(node, filter, objectStack) {
   if (filter.matchCase !== undefined) {
     node.setAttribute('matchCase', filter.matchCase.toString());
   }
-  writeOgcPropertyName(node, filter.propertyName);
-  writeOgcLiteral(node, '' + filter.pattern);
+  // TODO -- make this a safer check
+  if (typeof filter.propertyName === 'object') {
+    const child = createElementNS(OGCNS, 'Function');
+
+    child.setAttribute('name', filter.propertyName.name);
+    writeOgcLiteral(child, `${filter.propertyName.expression}`);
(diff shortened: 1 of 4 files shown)


glenselle:
It appears taking this approach is going to be really messy with Typescript enforcing types. And probably error-prone anyway? Is there another design pattern that would be better suited here? Also, our focus (as described in the linked issue) was on `IsLike` -- but I imagine `Function` will be used with other operators.

tschaub:
#10862 brought in some new formatting rules that will require a small change to your commit.  I've pushed a rebased, reformatted version of your commit in f988fc42a9b9b3617015e396df6a21f834e21a68.  You can either do the rebase and reformatting yourself or use this commit.

If you want to rebase and reformat yourself, you can do something like this:
```
npm run lint -- --fix
```

If you want to use the commit above, you can do this:
```
git fetch openlayers
git reset --hard f988fc42a9b9b3617015e396df6a21f834e21a68
```

In either case, you'll want to force push a change to your branch after making the changes:
```
git push --force origin issue-10823
```

glenselle:
I hate to create a stale PR on the repo so I'm going to close for now but leave the issue open. As I mentioned, not sure how best to proceed but totally willing to work on this if the maintainers have a specific way they'd like to handle this.

ahocevar:
@glenselle Thanks for your contribution! If you are able to finish this pull request with tests in a timely manner, it would be appreciated. Feel free to do the refactoring you suggested, or stick with passing around `propertyName` as string. If you need help for type checking, I'll be able to help. If you rebase your existing work on master, this will be easier than it used to be, because I made some type checking related changes to the XML formats.

Issue #6416, opened 2017-01-30
Optimization proposal for overlays


ogo-adp:
When a map contains a lot of overlays, panning could be made faster by implementing updateRenderedPosition() (in overlay.js) a bit differently. 

Currently it does position the overlay using top/left style properties. But by using translate3d() css transform, it could be made faster because this transform does use the gpu. Other mapping frameworks make use of this trick.

Here is a very crude implementation using translate3d() to play with it :

`ol.Overlay.prototype.updateRenderedPosition = function (pixel, mapSize) {`
     `  var offset = this.getOffset();`
         `   var left = Math.round(pixel[0] + offset[0]);`
         `   var top = Math.round(pixel[1] + offset[1] - mapSize[1]);`
          `  var newTransform = 'translate3d(' + left + 'px,' + top + 'px,0px)';`
       `     var style = this.element_.style;`
           ` if (style.transform !== newTransform)`
           `     style.transform = newTransform;`
          `  this.setVisible(true);`
`        };`

ahocevar:
Just out of curiosity: what is your reason to prefer `ol.Overlay` objects and not use `ol.layer.Vector`?

ogo-adp:
We use ol.overlay to generate "stickers" for moving objects (which are based on on.layer.Vector). We need HTML/CSS to display mixed text, icons, colors (and probably some CSS animations in the near future). It would be very hard to do this with only ol.layer.Vector.

ahocevar:
Pull request welcome. But please ensure that this also works on IE9.

PR #10848, opened 2020-03-27
Speed up Overlay element positioning using CSS translate()

horsenit:
A similar proposal was made a few years ago (#6416) with no pull request.

This makes panning with, zooming, or updating Overlays feel much more responsive.

https://codesandbox.io/s/overlays-sy7xy has an example (using the current method) with support for many overlays.

In my case, while currently using Chrome 80, panning 5000 overlays with the transform method feels more responsive than 1000 with the current method. There's also a slightly smaller improvement in performance in Firefox and IE11.

I don't have IE9 available, but this method did work with IE11 in IE9 compatibility mode.

### DIFF
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -526,7 +526,8 @@ class Overlay extends BaseObject {
     if (this.rendered.transform_ != transform) {
       this.rendered.transform_ = transform;
       style.transform = transform;
-      style.msTransform = transform; // IE9
+      // @ts-ignore IE9
+      style.msTransform = transform;
     }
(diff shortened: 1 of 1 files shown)


aalbericio:
Hello all,

It really is an improvement compared to what we are experiencing in 5.3.3 (yeah, it's quite old). Thanks for the sandbox!

Btw, is there anyway in OL to avoid this "position calculation" when the overlay's location (coordinate) is outside the current extent? Correct if I'm wrong but that's what the canvas is doing with default markers/features?

Thanks in advance.

PR #10850, opened 2020-03-27
API docs for enums

ahocevar:
This pull request is mostly based on @MoonE's work in #10512. After seeing the results of #10802 and the review of #10805, I remembered how we decided to document enums a long time ago: only the values that are documented as options somewhere are marked as `@api`, but not the enums themselves. It also makes no sense to have enums in the docs that are only used internally.

With this change, consistency should be restored.

### DIFF
--- a/config/jsdoc/api/plugins/api.js
+++ b/config/jsdoc/api/plugins/api.js
@@ -73,9 +73,6 @@ function includeAugments(doclet) {
           });
         }
         cls._hideConstructor = true;
-        if (!cls.undocumented) {
-          cls._documented = true;
-        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


ahocevar:
This is not ready for review yet.

MoonE:
These two commits should probably not be reverted:
058ca59 Also correctly detect default exports without an identifier
f247762 Remove redundant check if doclet kind is typedef

They are not directly related to the enums in the apidoc.

ahocevar:
@MoonE Thanks, I figured that out already.

This is now ready for review.

ahocevar:
Thanks for any review.

fredj:
Result: https://598-4974939-gh.circle-artifacts.com/0/apidoc/index.html

Issue #10838, opened 2020-03-23
Clipping layer with a vector source is not working anymore?


SDaron:
Trying to use a vector layer for clipping, I can't have a working example with the last build.

The [last codepen](https://github.com/openlayers/openlayers/issues/8216#issuecomment-391416492) proposed for v4.6.5 is not working anymore. I've made [a code pen here](https://codepen.io/sdaron/pen/XWbPQWz), which is an exact copy of the v4.6.5 exemple but with the last v6.2.1, showing a black vector.

Maybe I'm missing something, but the clipping layer propose in example is not really helpfull to understand how to use this. Is there some documentation to make this work with the v6?

ahocevar:
The one thing you might be missing is that `renderMode: 'image'` has been removed. You now have to use `ol/layer/VectorImage` with a vector source.

SDaron:
Thanks for your answer. As I read you, this should work and maybe this is a misunderstanding and not a bug. Sorry for that. I've updated the [codepen](https://codepen.io/sdaron/pen/XWbPQWz), replacing ol/layer/Image && renderMode: "image" with an ol/layer/VectorImage but it still not working, here is the relevant part:
```
var style = new ol.style.Style({
  fill: new ol.style.Fill({
    color: "black"
  })
});
var clipLayer = new ol.layer.VectorImage({
  style: style,
  source: new ol.source.Vector({
    url:
      "https://raw.githubusercontent.com/1Map/geojson_public/master/custom.geo.json",
    format: new ol.format.GeoJSON()
  })
});

clipLayer.on("precompose", function(e) {
  e.context.globalCompositeOperation = "destination-in";
});

clipLayer.on("postcompose", function(e) {
  e.context.globalCompositeOperation = "source-over";
});
```

mike-000:
The method won't work in OpenLayers 6 where each layer has its own canvas.  But it can be done by drawing the polygon to the postrender vector context https://codesandbox.io/s/vector-layer-m12el

SDaron:
Thanks. So this is due to a design change. If I made a PR with an new example for the doc with this, would it be considered usefull?

ahocevar:
@SDaron An example pull request would be much appreciated. Thanks in advance!

fredj:
fixed with #10859

PR #10859, opened 2020-04-01
Add an example of clipping layer based on a vector source

SDaron:
Following the discussion on #10838. This PR add an example of clipping layer based on a vector source.


tschaub:
Thanks, @SDaron!

Issue #10705, opened 2020-02-19
Support zIndex on WebGLPointsLayer


jellyedwards:
**Is your feature request related to a problem? Please describe.**
The non-WebGL layers and the previous WebGL implementation supported zIndex ordering the features so some would always appear on top of others. For example I'd like the green icon here to appear on top of the black icons:

![image](https://user-images.githubusercontent.com/14051992/74857591-ca464480-533b-11ea-8f41-15d82cc45af5.png)

However I don't see any way of specifying the zIndex in the LiteralStyle...

**Describe the solution you'd like**
Something like:

```
const webGLStyle = {
...
    zIndex: ['get', 'zIndex']
...
};
```

on the literal style to allow zIndex ordering of points.

Thanks!!

jahow:
Duplicate of https://github.com/openlayers/openlayers/issues/10634

jellyedwards:
Sorry I did search but didn't find that one

Issue #10634, opened 2020-02-11
z index & WebGL points renderer


sign0:
I'm looking for a solution to assign a z index to the feature itself.
Is it possible with the WebGL points renderer ?

jahow:
Not yet, and that would be a great improvement. I'll mark this as PR accepted if some one wants to try their luck on this.

In my opinion the simplest way to do this would be to use a depth buffer (which is currently disabled in the points renderer) when drawing points. The downside being that this will cause transparency glitch if the rendered points are not fully opaque. Sorting the features before rendering them is feasible as well but will degrade performance even more when rebuilding the buffers.

jellyedwards:
Hey @jahow,

I've tried to implement depth buffer/z-ordering in WebGL but it doesn't seem to have any effect. 

Here's the branch on my forked openlayers: https://github.com/jellyedwards/openlayers/tree/WebGL-zIndex if you'd like to have a look and you might see what I'm doing wrong.

I've changed the icon-sprite-webgl.js example to make fireballs red and have a higher z-index than the other sprite types (I also increased the size of the icons to make ordering easier to see I tried a few of the options (GEQUAL/GLESS/NEVER, disable/enable DEPTH_TEST, etc.) but nothing seems to work.

The items may have to be ordered on the CPU side first, see the answer here: https://stackoverflow.com/questions/185536/in-opengl-how-can-you-get-items-to-draw-back-to-front.

Cheers - if this isn't the right place to discuss let me know where to go.

jahow:
Hi @jellyedwards, thanks for working on this! I have tested your code and it seems that somehow the depth buffer doesn't get the right value. Not entirely sure why, but IMO this has to do with this line:
https://github.com/jellyedwards/openlayers/commit/6c325a8a48f3d4355da5a32476a07413413a98f7#diff-6a6ff35db3348550b898217b3fe6b343R349

The z value should be set here in "world" space (so not in the -1..1 range but between whatever near and far plane we chose). There must be something wrong with the projection matrix though, so I guess you should lookup how an orthographic projection matrix is built.

Other than that, I think this is great progress!

PR #10861, opened 2020-04-01
z index on WebGL points renderer

jellyedwards:
This should hopefully sort #10634 (and my duplicate #10705); zIndex sorting on the WebGL PointsLayer.

For example, here every second icon has a zIndex of 1 (vs 0):

```
  var style = {
    symbol: {
      symbolType: "image",
...
      zIndex: ["get", "zIndex"],
    } 
  };

  features = [];
  const total = 5000;
  const side = 70;
  for (let i = 0; i < total; i++) {
    var coords = fromLonLat([j * 90 + (i % side), Math.floor(i / side)]);

    features.push(
      new Feature({
...
        zIndex: i % 2,
        geometry: new Point(coords)
      })
    );
  }

  var vectorSource = new Vector({ features });

  map.addLayer(
    new WebGLPointsLayer({
      source: vectorSource,
      style: style
    })
  );
```

image

I welcome any feedback on this. One part I'm particularly unsure of is that to keep things consistent I have a zIndex in the LiteralStyle so I can retrieve values from each feature using the same methods (e.g. ['get', 'z']) but the zIndex isn't used in the shader. As a result there's a spare a_zIndex in there that's unused.

### DIFF
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -17,6 +17,7 @@ import {
 import {
   applyTransform,
   containsCoordinate,
+  containsExtent,
   equals,
   getCenter,
   getHeight,
@@ -622,9 +623,9 @@ class Graticule extends VectorLayer {
   drawLabels_(event) {
     const rotation = event.frameState.viewState.rotation;
     const extent = event.frameState.extent;
-    let rotationCenter, rotationExtent;
+    const rotationCenter = getCenter(extent);
+    let rotationExtent = extent;
     if (rotation) {
-      rotationCenter = getCenter(extent);
       const width = getWidth(extent);
       const height = getHeight(extent);
       const cr = Math.abs(Math.cos(rotation));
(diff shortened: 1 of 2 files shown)


jahow:
Hey! Thanks a lot for looking into this, your approach is interesting (grouping features by zIndex, hopefully not having to sort all the features this way). I'm not sure that would work out with an expression (id `['get', 'z']`) in the zIndex field, since it isn't evaluated on the GPU.

For this feature I think it'd be better to use GPU depth testing to draw features on top of others, without having to actually sort the rendering order. This has a strong limitation though, which is that if points have transparency this will cause visual anomalies. In that case, I think it would be reasonable to fall back to actually sorting the features array on the CPU, which is something that can be done either on the application code or by the source object.

In any case, I really think the renderer code should *not* be complexified, especially `rebuildBuffers_` which already takes a huge amount of time and causes hiccups when a source has more than 10'000 features.

What do you think?

jellyedwards:
Hi @jahow, yeah you're probably right. I'll scrap it and if I get time I might try the depth buffer way.

For my application I have transparency so was hoping to avoid those issues!

BTW the ['get', 'z'] zIndex thing seemed to work fine as it was just used on the CPU side when it gets values before passing to the GPU (PointsLayer.js:330):

```
this.zIndexAttribute_.callback(feature, properties)
```

which is the same method as (PointsLayer.js:552)

```
        // pushing custom attributes
        let value;
        for (let j = 0; j < this.customAttributes.length; j++) {
          value = this.customAttributes[j].callback(featureCache.feature, featureCache.properties);
...
```

PR #10862, opened 2020-04-01
Use prettier

tschaub:
This updates our ESLint config to use Prettier.  This wasn't possible previously because Prettier didn't allow JSDoc-style type casts (see prettier/prettier#7791 for the fix).

Contributors don't need to do any extra setup or install additional editor integrations beyond what they already use with ESLint.  In fact, contributors _should not_ set up their editors to use Prettier directly – instead it is used as an ESLint plugin.

If we merge this, all existing pull requests will likely need a rebase, and in some cases the conflicts will be nasty.  So we should get in any larger pull requests before this one.  After doing a rebase (or at any time), contributors can auto-fix formatting issues with this:

```
npm run lint -- --fix
```


config/jsdoc/api/plugins/markdown.js
```diff
@@ -63,7 +63,7 @@ function parse(source) {
 
   result = marked(source, {renderer: markedRenderer})
     .replace(/\s+$/, '')
-    .replace(/&#39;/g, '\'');
+    .replace(/&#39;/g, "'");
```
ahocevar, line 66: This one is interesting, but I have to admit I like it.

src/ol/format/WMSGetFeatureInfo.js
```diff
@@ -111,31 +104,43 @@ class WMSGetFeatureInfo extends XMLFeature {
...
+          this.gmlFormat_
+        );
         const parsersNS = makeStructureNS(
-          [context['featureNS'], null], parsers);
+          [context['featureNS'], null],
+          parsers
+        );
         layerElement.setAttribute('namespaceURI', this.featureNS_);
         const layerFeatures = pushParseAndPop(
           // @ts-ignore
```
tschaub, line 124: It [looks like](https://app.circleci.com/pipelines/github/openlayers/openlayers/3491/workflows/6e8f1da1-bc7e-40e2-a13b-4f41d0f33b2a/jobs/3566) this comment needs to get pushed down a couple lines.  I'll update in a new commit.

tschaub: Addressed in ad851e779f4545bc0a156820e0943f9a7bca0dbc.

config/jsdoc/api/plugins/api.js
```diff
@@ -2,19 +2,18 @@
  * Define an @api tag
  * @param {Object} dictionary The tag dictionary.
  */
-exports.defineTags = function(dictionary) {
+exports.defineTags = function (dictionary) {
```
tschaub, line 5: I think this one is going to take the most getting used to for me.  Lots of reading about it on prettier/prettier#3903 - but I still don't really get the justification.  But I'm just treating it as another example of a strong belief/convention/habit that there is really no reason to hang on to.  It is really kind of liberating to just stop caring about things like this – while still getting satisfaction knowing that the ultimate goal of improved legibility due to consistency is being achieved.

ahocevar: It reminds me of this very same discussion we had years ago in OpenLayers 2. And for what it's worth, this is the setting I have definitely disabled in my project configuration, but don't really care about it any more now that I see the change here.

ahocevar:
The only thing I am concerned about is the potential merge conflicts we get with open pull requests. Let's really put some effort into reviews of open pull requests before merging this.

tschaub:
> The only thing I am concerned about is the potential merge conflicts we get with open pull requests. Let's really put some effort into reviews of open pull requests before merging this.

Agreed.  I'll leave it open and rebase as needed.  When we feel ok about the size of the open PR list (maybe zero), we can follow up here and I'll merge.

PR #10876, opened 2020-04-04
Add image style support for LineString geometries

jvain:
This PR adds image style rendering for LineString geometries. This is needed for example rendering highway or interstate shields to LineString roads.

### DIFF
--- a/src/ol/render/canvas/ImageBuilder.js
+++ b/src/ol/render/canvas/ImageBuilder.js
@@ -168,6 +168,37 @@ class CanvasImageBuilder extends CanvasBuilder {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mike-000:
This would be a breaking change for many existing application which use a single style which includes both stroke and image similar to the default https://openlayers.org/en/latest/apidoc/module-ol_style_Style-Style.html as images intended only for Points would appear on LineStrings.

Images can already be easily added to LineString styling using a geometry function which returns Point or MultiPoint geometry as in https://openlayers.org/en/latest/examples/polygon-styles.html

jvain:
Thanks, the geometry function looks just like the thing that's needed.

Issue #10871, opened 2020-04-03
Included typings do not work correctly


MichalK6677:
I've been using types from DefinitelyTyped 
(https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ol) and all worked fine.

It wants to use new OL included typings in 6.3.0 and ignores the DefinitelyTyped typings. The included types does not work correctly.

Example of included types:
![image](https://user-images.githubusercontent.com/3100059/78376584-d37b3000-75ce-11ea-8d60-f1074b4f083a.png)

Comparsion between OL types and DefinitelyTyped types
![image](https://user-images.githubusercontent.com/3100059/78376799-1e954300-75cf-11ea-85cb-39650cede9db.png)

How can I use the OL types in this example where DefinitelyTyped typings are used:
```
this.draw = new Draw({
	source: this.source,
	type: GeometryType.POINT,
	snapTolerance: 6,
	stopClick: true
});
```
There are many types like this.

How can I exclude/ignore the included types and use types from https://github.com/DefinitelyTyped? 
I have angular application with tsconfig.json. I've tried ignore section in tsconfig.json but it does not help.
![image](https://user-images.githubusercontent.com/3100059/78377481-f528e700-75cf-11ea-8647-38b6121b2364.png)

jeroenvheel:
I also have the same issue I can’t compile the code with 6.3.0

mhosman:
Same issue here!

ahocevar:
Try to add the following to the `compilerOptions` section of `tsconfig.json`:
```json
  "skipLibCheck": true

**Comment by mhosman** [comment]
Hey @ahocevar this is a workaround or a required option from now on?

**Comment by MichalK6677** [comment]
@ahocevar 
I've tried 
```
"skipLibCheck": true,
```
and
```
"skipDefaultLibCheck": true,
```

It did not help. There are still errors in imports because it still does not take it from DefinitelyTyped.
![image](https://user-images.githubusercontent.com/3100059/78400599-51ebc800-75f7-11ea-80e6-09fb279c939d.png)

ahocevar:
@mhosman It depends on the TypeScript version you are using.

@MichalK6677 The inofficial types from DefinitelyTyped were manually created for older versions of OpenLayers, so some of them will very likely be incompatible. My advice would be to update your code to work with the types that ship with OpenLayers when upgrading to v6.3.

MichalK6677:
[...]

PR #10877, opened 2020-04-04
Remove .d.ts files from the package

### LINKED ISSUE #10875: type definitions errors
There are multiple errors reported by typescript compiler when using openlayers 6.3.0 in typescript. These error make `ol` unusable with framework like angular as compilation errors prevent successful project build.
Attached repro project has dependency only on `ol` (v6.3.0) and `typescript` (v3.8.3) and contains sample.ts file that creates a few ol objects.

**Steps To Reproduce**
1. get the attached repro project [ol_sample.zip](https://github.com/openlayers/openlayers/files/4431962/ol_sample.zip)
1. execute `npm install`
1. execute `npm run build`
3. See errors from typescript compilation

**Expected Behaviour**
there should be no errors reported

Note: if additional types were used in sample.ts, additional errors might be reported

ahocevar:
Now that the included type definitions got exposed to more TypeScript users with the 6.3.0 release, it turns out that there are several issues (#10875, #10871) with them.

This pull request removes the .d.ts file generation step.

I'll publish a patch release when this is merged.

Fixes #10875.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,31 @@ OpenLayers appreciates contributions of all kinds.  We especially want to thank 
 
 See our [Open Collective](https://opencollective.com/openlayers/contribute/sponsors-214/checkout) page if you too are interested in becoming a regular sponsor.
 
+## IntelliSense support and type checking for VS Code
+
+The ol package contains a src/ folder with JSDoc annotated sources. TypeScript can get type definitions from these sources with a `jsconfig.json` config file in the project root:
+
+```json
+{
+  "compilerOptions": {
+    "checkJs": true,
+    "baseUrl": "./",
+    "paths": {
+      "ol": ["node_modules/ol/src"],
+      "ol/*": ["node_modules/ol/src/*"]
+    }
+  },
+  "include": [
+    "**/*.js",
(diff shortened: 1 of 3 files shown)


mediafreakch:
@ahocevar Just stumbled over this. I want to use openlayers in my project that is using typescript. Is there a workaround to get typings for >6.3.1 as long as you wait for that typescript bug to be fixed?

ahocevar:
@mediafreakch There is an inofficial `@types/ol` package that you could use.

namedgraph:
@ahocevar why does the README claim ["TypeScript support"](https://github.com/openlayers/openlayers#typescript-support) then? It's confusing!

tschaub:
> @ahocevar why does the README claim ["TypeScript support"](https://github.com/openlayers/openlayers#typescript-support) then? It's confusing!

The change mentioned in https://github.com/openlayers/openlayers/pull/12430#pullrequestreview-703487781 brought the type declarations back in.  So the README is up to date.  This issue is out of date.  The change that brought type definitions back in (9e88134286bb9bb9523b93ecf6a91d025d3ddb40) is included in releases 6.6.0 and higher.

PR #10878, opened 2020-04-05
Avoid page scrolling when freehand drawing on mobile

### LINKED ISSUE #10804: Freehand drawing on map ends by itself after start drawing.
**Describe the bug**
Freehand drawing on map ends by itself after start drawing.

**To Reproduce**
1. https://openlayers.org/en/latest/examples/draw-freehand.html?q=draw
2. go in developer tools, mobile view
3. try to draw

**Expected behavior**
it should stop drawing when touch is done.

I already found a way to fix this issue:
touch-action: None;
seems to be removed in mobile view.

ahocevar:
Fixes #10804.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -523,7 +523,11 @@ class Draw extends PointerInteraction {
     } else if (move) {
       pass = event.type === MapBrowserEventType.POINTERMOVE;
       if (pass && this.freehand_) {
-        pass = this.handlePointerMove_(event);
+        this.handlePointerMove_(event);
+        if (this.shouldHandle_) {
+          // Avoid page scrolling when freehand drawing on mobile
+          event.preventDefault();
+        }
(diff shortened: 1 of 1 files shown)


src/ol/interaction/Draw.js
```diff
@@ -523,7 +523,11 @@ class Draw extends PointerInteraction {
     } else if (move) {
       pass = event.type === MapBrowserEventType.POINTERMOVE;
       if (pass && this.freehand_) {
-        pass = this.handlePointerMove_(event);
+        this.handlePointerMove_(event);
+        if (this.shouldHandle_) {
```
tschaub, line 527: Are there cases where `this.shouldHandle_` is true and `this.handlePointerMove_` wasn’t returning `false`?

ahocevar: `this.handlePointerMove` always returned true, so I removed the return.

PR #10882, opened 2020-04-05
Apidoc - Fix toggle state when there are no hidden members

MoonE:
Trying to expand an element in the apidoc sidebar does not trigger an icon transition the first time if the clicked class/module has no hidden members.

Examples are `ol/webgl/Buffer`, or searching for 'stableSort' and then expaning the `ol/array` module


tschaub:
Thanks, @MoonE .

PR #10885, opened 2020-04-06
Add @api tag to SelectEvent

ejn:
SelectEvent is otherwise missing in the generated .d.ts files

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -96,6 +96,7 @@ const SelectEventType = {
  * @classdesc
  * Events emitted by {@link module:ol/interaction/Select~Select} instances are instances of
  * this type.
+ * @api
  */
 class SelectEvent extends Event {
   /**


ahocevar:
We are no longer generating .d.ts files, at least until the remaining TypeScript issues are fixed.

That said, we would need to rework the way we document events in the API docs, and the types of the listener function arguments. Currently our event types are typed as `string`, and the arguments to the listener function is `ol/events/Event`.

Changing this would involve adding `@template`s to all our Observable subclasses that emit public events, and to the respecitve `on` and `un` functions. If someone wants to take on this effort, a pull request would be appreciated. I think it would make sense to start a separate pull request for that, so I'm closing this one.

ejn:
OK @ahocevar - I'd missed #10877. Certainly from my viewpoint at least the specialised Event classes should be exposed as the individual properties are important in the handlers.

PR #10893, opened 2020-04-06
Fix focus condition

### LINKED ISSUE #10888: Can't pan/mousewheel zoom the map on IE 11
**Describe the bug**
SInce the change of behavior with the focus/tabindex/scroll stuff added in 6.2.0, if the map element has a tabIndex, it's impossible to pan the map on IE 11 .

**To Reproduce**
1. Open https://openlayers.org/en/latest/examples/simple.html on IE 11
2. Add a tabindex="0" to the map element in the dev tools.
3. Try to pan the map
---> nothing happens because the focus is on the layer element, not on the map element.

**Expected behavior**
The user can pan the map.

If someone has a miraculous workaround. I take it.

ahocevar:
Fixes #10888

### DIFF
--- a/src/ol/events/condition.js
+++ b/src/ol/events/condition.js
@@ -58,7 +58,7 @@ export const altShiftKeysOnly = function(mapBrowserEvent) {
  * @api
  */
 export const focus = function(event) {
-  return event.target.getTargetElement() === document.activeElement;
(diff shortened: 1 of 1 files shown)


oterral:
@ahocevar awesome !! thx for your reactivity!!!

PR #10934, opened 2020-04-15
Update TypeScript to v3.8

ahocevar:
This pull request makes several fixes related to types, which allows us to upgrade to the latest TypeScript version.

Most issues were around private/protected/public scope problems. I resolved some of them by refactoring the way we pass functions to the contstructor (e.g. `tileUrlFunction`). Now we simply rely on inheritance for that. A broader field of problems were the XML formats. Due to the way these allow matching and mixing properties and functions from other formats, I had to remove private/protected restrictions from these properties and functions.

This pull request also contains one commit that works around microsoft/TypeScript#33575, b0f20d6 to make `npm run typecheck` pass. We might be able to revert that commit when the TypeScript issue is fixed. Also, until that TypeScript issue is fixed, enum types in applications still do not work correctly.

When these changes are in, TypeScript projects can also benefit from the types that are directly derived from the sources in the package's `src/` folder.

### DIFF
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -290,7 +290,7 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
 
     // set scale transform for calculating tile positions on the canvas
     composeTransform(
-      this.tempTransform_,
+      this.tempTransform,
       width / 2,
       height / 2,
       canvasScale,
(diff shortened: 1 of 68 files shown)


tschaub:
Wow.  Thanks for all the careful work on this, @ahocevar.

Odd not to see a CircleCI pipeline running for this.  I wonder if you need to push a new commit.

src/ol/renderer/canvas/TileLayer.js
```diff
@@ -290,7 +290,7 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
 
     // set scale transform for calculating tile positions on the canvas
     composeTransform(
-      this.tempTransform_,
+      this.tempTransform,
```
tschaub, line 293: I'm happy to see all these `_` suffixed properties go.

PR #10937, opened 2020-04-16
Use the tile coordinate as a tile pseudo URL in geojson-vt example

### LINKED ISSUE #10933: Wrong tiles generation in geojson-vt example
**Describe the bug**
In the geojson-vt example when zooming in inside a polygon some tiles are not rendered.
You can notice this only when you give a fill color to the style.

**To Reproduce**
1. Go to 'https://codesandbox.io/s/geojson-vt-7ghcj?file=/index.js'
3. Some tiles will be empty instead of solid green

**Expected behavior**
Polygons should be filled.

The issue comes from this line of code:
 if (this.sourceTileCache.containsKey(tileUrl)) in ol/source/VectorTile
and it's because all tileUrls for a specific feature when the tile is completely contained in the polygon will look similar to this:
data:application/json;charset=UTF-8,{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[-64,4160],[-64,-64],[4160,-64],[4160,4160],[-64,4160]]]},"properties":{"name":"Italy"}}]} 
It's not really a bug in OpenLayers but it's still not working properly.

As a workaround I just need to set a property unique for differentiate same features in different tiles. I use the z/x/y combined string for this.

bonus question: I couldn't figure it out why using Math.random() in the replacer doesn't render any tile at all.

If you think there is nothing to fix here, maybe it could just be mentioned in the example for future reference

mike-000:
Fixes #10933

The issue with solid fill seen in #10933 can also be seen although less obvious in the semi-opaque default fill in the live example.

Using the unique tile coordinate as a tile pseudo URL instead of a GeoJSON data URL which may not be unique prevents problems with tile caching,  The construction of the GeoJSON is instead done in a custom tileLoadFunction.

### DIFF
--- a/examples/geojson-vt.js
+++ b/examples/geojson-vt.js
@@ -82,6 +82,16 @@ fetch(url)
         }),
       }),
       tileUrlFunction: function (tileCoord) {
+        // Use the tile coordinate as a pseudo URL for caching purposes
+        return JSON.stringify(tileCoord);
+      },
+    });
+    // For tile loading obtain the GeoJSON for the tile coordinate
+    // before calling the default tileLoadFunction
+    const defaultTileLoadFunction = vectorSource.getTileLoadFunction();
+    vectorSource.setTileLoadFunction(
+      function (tile, url) {
+        const tileCoord = JSON.parse(url);
         const data = tileIndex.getTile(
           tileCoord[0],
           tileCoord[1],
@@ -94,9 +104,12 @@ fetch(url)
           },
           replacer
         );
-        return 'data:application/json;charset=UTF-8,' + geojson;
-      },
-    });
+        defaultTileLoadFunction(
+          tile,
+          'data:application/json;charset=UTF-8,' + geojson
+        );
+      }
+    );
     const vectorLayer = new VectorTileLayer({
       source: vectorSource,
     });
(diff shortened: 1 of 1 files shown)


mike-000:
> If you choose to use a custom tileLoadFunction, then why go through a data uri? Then you can use the minimal tileLoadFunction as shown in the docs. Would make the example easier to understand.

I had attempted that and only succeeded in getting features at [0, 0], but I have it working now.

examples/geojson-vt.js
```diff
@@ -72,16 +72,21 @@ fetch(url)
...
+      dataProjection: new Projection({
+        code: 'TILE_PIXELS',
+        units: 'tile-pixels',
+        extent: [0, 0, 4096, 4096],
       }),
+    });
+    const vectorSource = new VectorTileSource({
       tileUrlFunction: function (tileCoord) {
+        // Use the tile coordinate as a pseudo URL for caching purposes
+        return JSON.stringify(tileCoord);
```
ahocevar, line 86:
```suggestion
        return tileCoord.toString();
```
would be a bit lighter. Feel free to commit or ignore.

PR #10958, opened 2020-04-24
Wrap transforms between global projections defined using proj4

### LINKED ISSUE #10928: Transforms between global projections using proj4 don't wrap as they do with predefined transforms
**Describe the bug**
I checked this as part of #10881 and for example a Miller projection does not always display vector layers correctly when using geographic mode because the transformed view extent does not reflect wrapping.  For the same reason even in standard mode meridians are sometimes missing on a graticule.

**To Reproduce**
Pan the geographic mode map https://codesandbox.io/s/simple-ubsq7 and sometimes the vector layer disappears or only some countries appear.  Pan the standard mode map https://codesandbox.io/s/simple-ixyex and sometimes meridians are missing

**Expected behavior**
If it is not too much overhead `transform` and `transformExtent` and the transform returned by `getTransform` should work consistently regardless of whether the transforms were predefined or obtained from proj4.  If both projections are global the world offsets in the source coordinates should be reflected to the destination coordinates.

mike-000:
Fixes #10928

If both projections are set to allow wrapX return an updated transform to handle that similar to the predefined projections, and which is used by default.  `getTransform` has an optional `wrapX` parameter which if false sets the transform to merely normalise the source coordinates before transforming - this might be useful when loading data into a single world of a destination projection where the world is offset from the data source world (a potential follow up might a `readWrapX()` method for formats which support `readProjection()` so features on a wrapping source load into its normal world).

`ol/layer/Graticule` is updated to handle offset projections which were previously assumed to be always non-wrapping.

### DIFF
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -1179,25 +1179,34 @@ class Graticule extends VectorLayer {
       this.toLonLatTransform_ = toLonLatTransform;
     } else {
       const split = this.minLon_ + this.maxLon_ / 2;
-      this.maxLon_ += 360;
-      this.toLonLatTransform_ = function (
-        coordinates,
-        opt_output,
-        opt_dimension
-      ) {
-        const dimension = opt_dimension || 2;
-        const lonLatCoordinates = toLonLatTransform(
+      if (projection.canWrapX()) {
+        if (split > 0) {
+          this.minLon_ -= 360;
+        } else {
+          this.maxLon_ += 360;
+        }
+        this.toLonLatTransform_ = toLonLatTransform;
+      } else {
+        this.maxLon_ += 360;
+        this.toLonLatTransform_ = function (
           coordinates,
           opt_output,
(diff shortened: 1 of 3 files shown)


mike-000:
proj4 and mproj seem very similar.  While they do typically understand wrapped source coordinates the output is almost always in the destination real world (unlike the built in EPSG:4326/EPSG:3857 transforms whose behaviour I am trying to replicate).

ahocevar:
Instead of trying the transform function to see what it does, it would be nicer to add options to `ol/proj.addCoordinateTransforms` to indicate what it does. Something like
```js
addCoordinateTransforms('FOO', 'BAR', forward, inverse, {
  forwardWrapsX: false,
  inverseWrapsX: true
});
```

The `proj4` module can then add transforms with the appropriate options.

mike-000:
Currently has to wait until getTransform() is called as the projection doesn't have an extent or global setting until after the it is registered.

PR #10960, opened 2020-04-25
Improve Vector Clipping Layer example with a background layer and intersect

SDaron:
This is a small update on vector layer clipping example, which, I think doesn't need to open an issue.
The PR porpose to use 'prerender' and 'postrender' directly on the vector source. The code is then easier to read and to understand.

### DIFF
--- a/examples/layer-clipping-vector.js
+++ b/examples/layer-clipping-vector.js
@@ -6,7 +6,6 @@ import {OSM, Stamen, Vector as VectorSource} from '../src/ol/source.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
 import {fromLonLat} from '../src/ol/proj.js';
 import {getVectorContext} from '../src/ol/render.js';
-import {intersects} from '../src/ol/extent.js';
 
 //A distinct className is required to use another canvas for the background
 const background = new TileLayer({
@@ -28,6 +27,11 @@ const clipLayer = new VectorLayer({
   }),
 });
 
+//Giving the clipped layer an extent is necessary to avoid rendering when the feature is outside the viewport
+clipLayer.getSource().on('addfeature', function () {
+  base.setExtent(clipLayer.getSource().getExtent());
+});
+
 const style = new Style({
   fill: new Fill({
     color: 'black',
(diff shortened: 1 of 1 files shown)


SDaron:
Actually, testing deeply, this is not working with a MultiPolygon feature, and so it's a regression.

Also actually the clipping layer has two weaknesses: 
1) the baselayer is not clipped anymore when you move out of the feature
2) not only the base layer is clipped, but the whole map, and it's not possible to use transparency to display another layer on the background.

I've try to improve this, without success until now.

mike-000:
For issue 2 the layers to be clipped and not clipped may need a different `className` property to ensure the clipped layer doesn't share a canvas with another layer.

SDaron:
Indeed, layers with distinct className have distinct canvas. This is not documented but works as expected then.

The only way I've found to have the layer clipped even when the feature is outside is to detect the feature presence with an intersect, see below.
The solution works but looks a bit tricky. Is their a cleaner way to detect that a feature/layer is rendered?
getVisible() returns true even is the layer is not rendered.
```
base.on('postrender', function (e) {

  const vectorContext = getVectorContext(e);  
  e.context.globalCompositeOperation = 'destination-in';  
  clipLayer.getSource().forEachFeature(function (feature) {
    vectorContext.drawFeature(feature, style);
  });  
  if(olExtent.intersects(e.frameState.extent, clipLayer.getSource().getExtent())){
    e.context.globalCompositeOperation = 'source-over';
  }
  
});
```
I still don't understand why MultiPolygons and Polygons does not have the same behaviour when using postrender and prerender events on the vectorLayer, however.

mike-000:
It would be better to always reset the `globalCompositeOperation` to its default value.  A dummy fill could be used trigger a `destination-in` when `drawFeature` would have no effect.

```
base.on('postrender', function (e) {

  const vectorContext = getVectorContext(e);  
  e.context.globalCompositeOperation = 'destination-in';  
  if(olExtent.intersects(e.frameState.extent, clipLayer.getSource().getExtent())){
    clipLayer.getSource().forEachFeature(function (feature) {
      vectorContext.drawFeature(feature, style);
    });  
  } else {
    e.context.fillStyle = 'transparent';
    e.context.fillRect(0, 0, 1, 1);
  }
  e.context.globalCompositeOperation = 'source-over';
  
});
```

SDaron:
Ok I've updated the example according to this and add few comments. As I've roll the code back to the previous clipping strategy, another PR is maybe better. Let me know.

SDaron:
Sorry, I don't understand why the checks failed. Am I supposed to do something?

```
> node rendering/test.js "--force"

Failed to launch the browser process!
Fontconfig warning: "/etc/fonts/fonts.conf", line 100: unknown element "blank"
[780:780:0426/113016.350058:ERROR:browser_main_loop.cc(1485)] Unable to open X display.

```

mike-000:
Rendering test failures are an ongoing issue, see  #10931  They are relevant for changes to examples but all test are run for any change.

SDaron:
All right so my part of the job is done :-)

ahocevar:
Thanks for the contribution @SDaron.

Drawing an empty fill looks like a dirty hack to me. Giving the vector source an extent would be a cleaner solution. Then it will only be rendered when it is inside the viewport.

mike-000:
> Thanks for the contribution @SDaron.
> 
> Drawing an empty fill looks like a dirty hack to me. Giving the vector source an extent would be a cleaner solution. Then it will only be rendered when it is inside the viewport.

The vector source isn't being rendered directly, but limiting the OSM layer to its extent would work

```
clipLayer.getSource().on('addfeature', function() {
  base.setExtent(clipLayer.getSource().getExtent());
});
```

SDaron:
Giving the base layer an extent works indeed perfectly and looks cleaner.

PR #10975, opened 2020-04-28
Do not render tile when there is nothing to render

ahocevar:
When a VectorTile source is used by more than one layer, an executor group may not exist for a layer that has just been switched from `visible: false` to `visible: true`. That is because the `wantedTiles` frame state is keyed by source, not by layer, and vector tiles have an executor group per layer.

This pull request avoids `cannot read length of undefined` errors when trying to render a tile that has no executor group for the current layer.


ahocevar:
Actually there must be something else wrong that causes this behavior. Need to investigate. Closing for now.

ahocevar:
Reopening. I think this fix is goo.

ahocevar:
Something does not work with CircleCI here. Closing.

PR #10977, opened 2020-04-28
Only render vector tile when there are executor groups

ahocevar:
This pull request prevents errors when there is no executor group for a tile. This can happen when a vector tile layer is configured with `renderMode: vector` and the source is being used by more than one layer.

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -210,9 +210,10 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
   isDrawableTile(tile) {
     const layer = this.getLayer();
     return (
-      (super.isDrawableTile(tile) &&
-        layer.getRenderMode() === VectorTileRenderType.VECTOR) ||
-      tile.hasContext(layer)
+      super.isDrawableTile(tile) &&
+      (layer.getRenderMode() === VectorTileRenderType.VECTOR
+        ? getUid(layer) in tile.executorGroups
+        : tile.hasContext(layer))
     );
   }
 


ahocevar:
I cannot figure out why CircleCI is not able to run the test suite with this pull request. Giving up for today, maybe someone else can take a look at this.

ahocevar:
Ha, looks like me giving up was enough for CircleCI to behave nicely again ;-)

ahocevar:
I removed the CI fix, and created a separate pull request for that: #10983. Will rebase when that is merged.

ahocevar:
Rebased, all green now.

ahocevar:
I'll see if I can handle this in a more consistent way by changing the `isDrawableTile` logic.

ahocevar:
Ready for review now, which a much cleaner fix.

ahocevar:
Thanks for any review.

tschaub:
Do you think there is a place we could throw an error the a user would see?  I know we’re not consistent with this kind of thing.  I was just pondering this for another case though.

tschaub:
Or is the right behavior just not to render (maybe I don’t understand the issue).

ahocevar:
> Or is the right behavior just not to render (maybe I don’t understand the issue).

The right behavior is not to render, because the tile is not ready for rendering yet. It will be rendered in the next cycle. That's the same in hybrid or image mode. The only difference is that `renderQueuedTileImages_` also creates an executor group directly after the tile has finished loading, and in vector mode this only happens in `prepareFrame`.

That said, the whole vector and vector tile rendering code could use a refactoring or rewrite, i.e. simplify executor groups, improve decluttering etc. - but I guess that effort will have to wait a bit.

PR #10981, opened 2020-04-29
Calculate WMTS TileGrid extent using TileMatrixSetLimit when available

### LINKED ISSUE #10979: WMTS optionsFromCapabilities ignores WGS84BoundingBox when creating WMTS Source
Like #10974, another regression with WMTS optionsFromCapabilities between version OL 6.1.1 and 6.3.1.

In v6.1.1, the WMTS.optionsFromCapabilities creates an WMTS Source/TileGrid that respects the WGS84BoundingBox:

https://github.com/openlayers/openlayers/blob/e4be8309f7880c322c9850a634dab8b6adc30c62/src/ol/source/WMTS.js#L380-L398

In v6.3.1, the extent of the TileGrid is calculated based on the first TileMatrix.

https://github.com/openlayers/openlayers/blob/af9f26b9d343e5496ea44a729265c4002616b240/src/ol/source/WMTS.js#L381-L400

The change seems to be introduced in commit https://github.com/openlayers/openlayers/commit/98e8bec3709364b5fd14dd71a4e4e72dcf4ae1c9

This can easily create a bounding box that is too large. For example a WMTS layer containing Europe with WGS84BoundingBox

 ```xml
<ows:WGS84BoundingBox>
<ows:LowerCorner>-36.0 15.0</ows:LowerCorner>
<ows:UpperCorner>43.0 90.0</ows:UpperCorner>
</ows:WGS84BoundingBox>
```

In v6.1.1, the tile grid will have a correct extent of [-36, 15, 43, 90]. 
In v6.3.1, the tile grid will have an incorrect extent of [-180, -90, 180, 90].  The calculation is based on TileMatrixLimits level 0, which spans the whole world. 

Because of this change, there are unnecessary HTTP requests to WMTS tiles that do not exist outside the WMTS bounding box area.

nielsmeijer:
Fixes #10979 

I've changed the extent calculation to use the TileMatrixSetLimit of the WMTS Capabilities when available. The most detailed level is used to get the closest possible representation of the extent. 
Extent calculation is unchanged when no TileMatrixSetLimit is defined.

Also, i fixed a potential issue where the switching of XY was only applied for EPSG:4326. Instead, now   ```projection.getAxisOrientation().substr(0, 2) == 'ne' ``` is used, as there are more NE coordinate systems (for example LAEA Lambert / EPSG:3035)

The changed case is covered by a test.  

<sup><sup>(Hopefully the lint won't be upset with me this time 😊, thanks for the pointer in my previous PR 👍, I had some failing tests locally though regarding scale lines. Hopefully this is a local glitch in my development environment )</sup></sup>

### DIFF
--- a/src/ol/source/WMTS.js
+++ b/src/ol/source/WMTS.js
@@ -451,20 +451,39 @@ export function optionsFromCapabilities(wmtsCap, config) {
   const wrapX = false;
   const switchOriginXY = projection.getAxisOrientation().substr(0, 2) == 'ne';
 
-  const matrix0 = matrixSetObj.TileMatrix[0];
+  let matrix = matrixSetObj.TileMatrix[0];
+
+  // create default matrixLimit
+  let selectedMatrixLimit = {
+    MinTileCol: 0,
+    MinTileRow: 0,
+    // substract one to end up at tile top left
+    MaxTileCol: matrix.MatrixWidth - 1,
+    MaxTileRow: matrix.MatrixHeight - 1,
+  };
+
+  //in case of matrix limits, use matrix limits to calculate extent
+  if (matrixLimits) {
+    selectedMatrixLimit = matrixLimits[matrixLimits.length - 1];
+    matrix = matrixSetObj.TileMatrix.find(
+      (value) => value.Identifier === selectedMatrixLimit.TileMatrix
+    );
+  }
+
   const resolution =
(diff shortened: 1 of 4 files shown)


nielsmeijer:
Ah, I thought I might add to the XML that I added for the other issue. Sure, I'll add another XML..

nielsmeijer:
@ahocevar I've added another XML for the unit test.

PR #10983, opened 2020-04-29
Trying to fix CI

ahocevar:
Looks like recent versions of Puppeteer do not work with node 14, which is what CircleCI is using now. This pull request fixes the CI, which is broken for all new pull requests since some time yesterday.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1849,16 +1849,6 @@
         "@types/geojson": "*"
       }
     },
-    "@types/yauzl": {
-      "version": "2.9.1",
-      "resolved": "https://registry.npmjs.org/@types/yauzl/-/yauzl-2.9.1.tgz",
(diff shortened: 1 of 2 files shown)


tschaub:
It looks like a bug in node that affected puppeteer has been fixed https://github.com/nodejs/node/pull/32968.  No reason to rush it, but it may be safe to upgrade again.  See also https://github.com/puppeteer/puppeteer/issues/5719.

ahocevar:
@tschaub Done already by @dependabot, in #11002.

tschaub:
#11002 brought in puppeteer@3.0.2, which only logged an error in Node 14.  It looks like #11032 will bring in puppeteer@3.0.4, which is the version that works with Node 14.1 and doesn't log a warning (see puppeteer/puppeteer#5816).

PR #10987, opened 2020-04-30
Use brokenDiagonalRendering reprojection processing if image smoothing is disabled

mike-000:
Apply the brokenDiagonalRendering reprojection processing if image smoothing is disabled to avoid the color distortion along reprojection edges which happens with both `source-over` and `lighter` globalCompositeOperation settings.

The distortion can be seen when highlighted by an ol/source/Raster showing odd and even values
image

### DIFF
--- a/src/ol/reproj.js
+++ b/src/ol/reproj.js
@@ -1,7 +1,7 @@
 /**
  * @module ol/reproj
  */
-import {assign} from './obj.js';
+import {assign, isEmpty} from './obj.js';
 import {
   containsCoordinate,
   createEmpty,
@@ -330,7 +330,7 @@ export function render(
     context.save();
     context.beginPath();
 
-    if (isBrokenDiagonalRendering()) {
+    if (isBrokenDiagonalRendering() || !isEmpty(opt_contextOptions)) {
       // Make sure that everything is on pixel boundaries
       const u0r = pixelRound(u0);
       const v0r = pixelRound(v0);
--- a/src/ol/reproj.js
+++ b/src/ol/reproj.js
@@ -1,7 +1,8 @@
 /**
  * @module ol/reproj
  */
-import {assign, isEmpty} from './obj.js';
+import {IMAGE_SMOOTHING_DISABLED} from './source/common.js';
+import {assign} from './obj.js';
 import {
   containsCoordinate,
   createEmpty,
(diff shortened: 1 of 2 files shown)


src/ol/reproj.js
```diff
@@ -330,7 +330,7 @@ export function render(
     context.save();
     context.beginPath();
 
-    if (isBrokenDiagonalRendering()) {
+    if (isBrokenDiagonalRendering() || !isEmpty(opt_contextOptions)) {
```
ahocevar, line 333: You could make this more specific and check for the imageSmooting option, instead of just checking whether there are contextOptions. One thing less to worry about in the future, in case other options get added.

mike-000: That is going to need some rework anyway.  It seems the latest version of Chrome behaves unexpectedly if `msImageSmoothingEnabled` (which is still needed by IE) is set so instead of passing an object containing both there will need to be browser specific code.

src/ol/source/common.js
```diff
@@ -12,7 +13,10 @@ export const DEFAULT_WMS_VERSION = '1.3.0';
  * Context options to disable image smoothing.
  * @type {Object}
  */
-export const IMAGE_SMOOTHING_DISABLED = {
-  imageSmoothingEnabled: false,
-  msImageSmoothingEnabled: false,
-};
+export const IMAGE_SMOOTHING_DISABLED =
+  FIREFOX || SAFARI || WEBKIT
```
ahocevar, line 17: Not a good idea to assume IE when it's not FIREFOX or SAFARI or WEBKIT. Better to use feature detection, e.g.
```js
'imageSmoothingEnabled' in CanvasRenderingContext2D.prototype
```
But that should not be exported here. Better to create that object in the TileImage source constructor when needed, to avoid problems when OpenLayers is loaded in a non-browser environment.

mike-000: I am going to revert that change as the fix doesn't work.

mike-000: I have now realised `msImageSmoothingEnabled`  isn't relevant to the problem. While testing locally without it I had been applying `imageSmoothingEnabled: false` to the reprojection of all layers which did fix the problem, but only if you want all layers on both both maps to have smoothing disabled, so it looks like a bug in the current version of Chrome.  The original purpose of the PR to fix the reprojection edges which also applies other browsers is complete and the PR is ready for merging.

PR #10990, opened 2020-05-01
Include context options in tile cache key

### LINKED ISSUE #10984: Chrome issue with disable image smoothing reprojected tile example
**Describe the bug**
I have checked https://openlayers.org/en/master/examples/disable-image-smoothing.html to see if it works correctly with the #10463 changes.  There is a new issue with Chrome which may or may not be related (I have already checked and it does NOT appear to be related to using `globalCompositeOperation = 'lighter'`).

**To Reproduce**
Open https://openlayers.org/en/master/examples/disable-image-smoothing.html in Chrome.  On two different PCs I get some consistent erroneous altitudes 

image

but the same point is correct when using Edge

image

While investigating I noticed that the problem could be temporarily fixed by turning reprojection off and on again.  That can be seen in https://codesandbox.io/s/disable-image-smoothing-sy0re by editing view projection to EPSG:3857 then back to EPSG:4326

**Expected behavior**
As in the original development there are very thin diagonal  lines of erroneous altitudes In all browsers along the reprojection edges which can hopefully be fixed now by using the #10463 `brokenDiagonalRendering` workaround, but there should be no differences between browsers.

mike-000:
Fixes #10984

It seems two sources using the same URL but different options (e..g. smoothing, cross origin) might be sharing the same internal cache (although that affect should have been seen in all browsers) so a follow up investigation the internal cache can be duplicated if necessary without the performance overhead of overriding browser cache.  The original Chrome issue now appears to be a browser cache problem and is possibly related to hardware acceleration.

### DIFF
--- a/examples/disable-image-smoothing.js
+++ b/examples/disable-image-smoothing.js
@@ -35,7 +35,7 @@ const enabledLayer = new TileLayer({
   source: new XYZ({
     attributions: attributions,
     url:
-      'https://api.maptiler.com/tiles/terrain-rgb/{z}/{x}/{y}.png?key=' + key;
+      'https://api.maptiler.com/tiles/terrain-rgb/{z}/{x}/{y}.png?key=' + key,
     maxZoom: 10,
     crossOrigin: '',
   }),
--- a/examples/disable-image-smoothing.js
+++ b/examples/disable-image-smoothing.js
@@ -35,9 +35,7 @@ const enabledLayer = new TileLayer({
   source: new XYZ({
     attributions: attributions,
     url:
-      'https://api.maptiler.com/tiles/terrain-rgb/{z}/{x}/{y}.png?key=' +
-      key +
-      '&map=2', // dummy parameter added to avoid shared cache
+      'https://api.maptiler.com/tiles/terrain-rgb/{z}/{x}/{y}.png?key=' + key;
     maxZoom: 10,
     crossOrigin: '',
   }),
--- a/examples/disable-image-smoothing.js
+++ b/examples/disable-image-smoothing.js
@@ -35,7 +35,8 @@ const enabledLayer = new TileLayer({
   source: new XYZ({
     attributions: attributions,
     url:
-      'https://api.maptiler.com/tiles/terrain-rgb/{z}/{x}/{y}.png?key=' + key,
+      'https://api.maptiler.com/tiles/terrain-rgb/{z}/{x}/{y}.png?key=' + key +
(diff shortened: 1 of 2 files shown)


mike-000:
I have already tried defining the sources using `ol/source/TileImage` where the `key` option is supported (I'm not sure why `ol/source/XYZ` doesn't support the `key` option when it also supports `tileUrlFunction`, or why key is only used in conjunction with `tileUrlFunction`) and it did not fix the problem.

ahocevar:
@mike-000 I don't think this can be handled on the application level. Overriding the `getKey()` method and adding a representation of the context options in `ol/reproj/Tile` should do the trick.

mike-000:
Using `ol/source/TileImage` distinct keys were being passed set in ReprojTile, but I have now added stringified context options to those values.  I have even added some extra `./` in the tile url in one of the url functions so the string is different but represents the same url, but all with no effect https://codesandbox.io/s/disable-image-smoothing-m20n3

 The only change that did work was one url with `?key=key&map=1` and the other with `?map=1&key=key`

ahocevar:
Very suspicious. Also the fact that you're saying this only happens in the latest version of Chrome (which one btw? v81?).

mike-000:
Currently on 81.0.4044.129 on Win10, there has been one update since I first noticed this., Also on  80.0.4044.117 on Android but it doesn't happen on an outdated phone where updates haven't been available since version 71.  I also developed this on Chrome between December and March and would have noticed if it was happening then.

ahocevar:
Regardless of whether this is a different issue or not, including the context options in the cache key of the tiles does make sense.

mike-000:
I agree, the potential for a problem existed.  The Chrome issue looks to be browser cache related as a change to the url of the enabled source requires several refreshes to fix the problem, but a change to the disabled source url fixes it immediately.

ahocevar:
Ok I see now. There is no reprojection involved in that example. Anyway, better to override `getKey()` in `ol/source/TileImage` instead of appending to the key you retrieve using `getKey()`. Otherwise this will always be error prone.

ahocevar:
I'd also vote against the workaround to append a query param to the url. No obscure workarounds please. Better to isolate the problem and find out why it only happens in Chrome 81. Experiments like disabling hardware acceleration and checking if the problem persists or the such.

mike-000:
Yes, I just turned off hardware acceleration on Win10 and I have not been able to reproduce the problem.

PR #11011, opened 2020-05-05
Set background fill style for each declutter label

### LINKED ISSUE #10949: backgroundFill color not respected when declutter is used
**Describe the bug**
All labels are displayed with the same backgroundFill color if declutter is specified.

**To Reproduce**
Without declutter even numbers have red backgroundFill, odd number blue
https://codesandbox.io/s/cluster-dq6it

If declutter is specified all numbers appear with the same color, either red or blue
https://codesandbox.io/s/cluster-cox1t

**Expected behavior**
The backgroundFill of the individual style should be used regardless of decluttering.

mike-000:
Fixes #10949

fillStyle needs to be set for each label to handle declutter groups containing different styles

### DIFF
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -262,6 +262,7 @@ class Executor {
    * @param {import("../../coordinate.js").Coordinate} p4 4th point of the background box.
    * @param {Array<*>} fillInstruction Fill instruction.
    * @param {Array<*>} strokeInstruction Stroke instruction.
+   * @param {boolean} declutter Declutter.
    */
   replayTextBackground_(
     context,
@@ -270,7 +271,8 @@ class Executor {
     p3,
     p4,
     fillInstruction,
-    strokeInstruction
+    strokeInstruction,
+    declutter
   ) {
     context.beginPath();
     context.moveTo.apply(context, p1);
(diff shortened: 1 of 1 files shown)


src/ol/render/canvas/Executor.js
```diff
@@ -280,6 +280,7 @@ class Executor {
     context.lineTo.apply(context, p1);
     if (fillInstruction) {
       this.alignFill_ = /** @type {boolean} */ (fillInstruction[2]);
+      context.fillStyle = /** @type {import("../../colorlike.js").ColorLike} */ (fillInstruction[1]);
```
ahocevar, line 283: Without a context.save/restore, doesn't this affect subsequent polygon fills?

mike-000: It would if the the color isn't set each time, but `.fill_()` seems to rely on that as it is already doing save and restore for rotation https://github.com/openlayers/openlayers/blob/master/src/ol/render/canvas/Executor.js#L1038

ahocevar: Well, since this seems to be an issue only when decluttering, I guess you are right. But then again, wouldn't it be anough to set the `fillStyle` here only when we're decluttering?

PR #11019, opened 2020-05-07
Option to disable image smoothing for Image layer sources

mike-000:
This extends the functionality added in #10429 to Image layer sources (ImageStatic and those with a `resolutions` option).  The base getContextOption() method is moved from ol/source/Tile to ol/source/Source to accomodate VectorImage layer using the Image layer renderFrame  Add with and without reprojection tests.  A checkbox is added to the Image Reprojection example to demonstrate a source with and without smoothing.


mst2020:
Hello & thanks all of you for your great work. :+1: 

@mike-000 In my opinion this does not work as i & maybe others will expect.

I'm using 6.4.2.
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.122 Safari/537.36

I added an example of my 3 color layer (green, orange, red), as you can see there are pixels added with `imageSmoothing: false`

### Working: Using global canvas setting: (=original image data)
```
HTMLCanvasElement.prototype.__oldGetContext =
  HTMLCanvasElement.prototype.getContext;

HTMLCanvasElement.prototype.getContext = function(type, options) {
  var context = this.__oldGetContext(type, options);
  if (type === "2d") {
    context.imageSmoothingEnabled = false;
    context.webkitImageSmoothingEnabled = false;
    context.mozImageSmoothingEnabled = false;
    context.msImageSmoothingEnabled = false;
  }
  return context;
};
```
imagesmoothgood
(Perfect)

### Not working: ImageStatic/imageSmoothing: false
```
new ol.source.ImageStatic({
                        url: url,
                        imageSmoothing: false,
                    });
```
imagesmoothbad
(Where does those new pixels come from?)

### And as reference the default behavior
imagesmoothdefault


Doesn't change if reprojection is used or not.

Am i misinterpreting something? Shouldn't it be like the "working" version with canvas?

Greetings. :)

P.S.: You'll see this also in the example: [(Image Reprojection)](https://openlayers.org/en/latest/examples/reprojection-image.html)

mike-000:
This is the Image Reprojection example viewed with Chrome

image

Only `imageSmoothingEnabled` and `msImageSmoothingEnabled` (needed for Internet Explorer) are set.  Are you using another browser which still requires a vendor prefix?  (apart from IE I thought all current versions now supported `imageSmoothingEnabled`?)

I am aware of a problem with Chrome when hardware acceleration is enabled in https://openlayers.org/en/latest/examples/disable-image-smoothing.html (but it doesn't happen if the second map is removed).

mst2020:
Thanks for your response!

Okay, my bad, the Example with the British national grid is not a good one because the original image has anti-aliasing, so it's hard to see. :)

"my" browser needs only `context.imageSmoothingEnabled = false;` set.
https://jsfiddle.net/dbtpm0c9 (works for me)

i've build some jsfiddle for you, if you don't see differences between 1) und 2) it is a problem with my browser (vivaldi), but its just using the chrome engine.
`Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.122 Safari/537.36`


1) imageSmoothing: false: (does **not** work)
https://jsfiddle.net/9mjv7csy/1/

2) using canvas before: (**= how it should be!**)
https://jsfiddle.net/9mjv7csy/2/

3) default:
https://jsfiddle.net/9mjv7csy

### Original image:
Untitled

mike-000:
Yes, I see the difference.  Where there is no reprojection the `imageSmoothing: false` setting is automatically applying the previous method which used to have to be coded in the `prerender` event as in https://openlayers.org/en/latest/examples/sea-level.html  But when set in all contexts via the getContext method the image is even sharper (and setting `imageSmoothing: false` in the OSM layer context or removing the OSM doesn't have any effect), so it looks like there is another canvas involved.

PR #11023, opened 2020-05-09
Add MapBrowserPointerEvent to apidoc

MoonE:
This adds a page for the `ol/MapBrowserPointerEvent` class which is linked to from a few different pages as detailed in #10460.

### DIFF
--- a/src/ol/MapBrowserPointerEvent.js
+++ b/src/ol/MapBrowserPointerEvent.js
@@ -3,6 +3,10 @@
  */
 import MapBrowserEvent from './MapBrowserEvent.js';
 
+/**
+ * @classdesc
+ * @api
+ */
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks, @MoonE.

ahocevar:
Sorry for being late to the party, but I'm not sure if we want the `MapBrowserPointerEvent` constructor to show up as API in the docs. We don't even do that for the `MapBrowserEvent`.

I think there must be a bug in the API doc generation. It should be documented like `MapBrowserEvent` if it is referenced by something API. I'm going to investigate this now.

ahocevar:
In #11025, I removed `MapBrowserPointerEvent` entirely.

tschaub:
I had a feeling this wasn't actually necessary.  Thanks for fixing it up, @ahocevar.

PR #11024, opened 2020-05-09
Make events work when inside a web component

### LINKED ISSUE #10980: Map on click event is fired when clicking on controls like the zoom-buttons
**Describe the bug**
Situation: Polymer 3 project
When using map.on('click', functionxxx) the function is executed when clicking on a control like th zoom-button

**To Reproduce**
1. Unzip [olinpolymer.zip](https://github.com/openlayers/openlayers/files/4552128/olinpolymer.zip)
2. Go to the unpacked folder
3. npm install
4. npm run start
5. Click on a zoom-button
6. The clicked pixel is logged in the browser console
**Expected behavior**
Not firing the event/not executing the function

ahocevar:
Fixes #10980.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1016,15 +1016,21 @@ class PluggableMap extends BaseObject {
       // coordinates so interactions cannot be used.
       return;
     }
-    const target = /** @type {Node} */ (mapBrowserEvent.originalEvent.target);
     if (!mapBrowserEvent.dragging) {
+      const rootNode = this.viewport_.getRootNode
+        ? this.viewport_.getRootNode()
+        : document;
+      const originalEvent = /** @type {PointerEvent} */ (mapBrowserEvent.originalEvent);
+      const target =
+        rootNode === document
+          ? /** @type {Node} */ (originalEvent.target)
+          : /** @type {ShadowRoot} */ (rootNode).elementFromPoint(
+              originalEvent.clientX,
+              originalEvent.clientY
+            );
       if (
         this.overlayContainerStopEvent_.contains(target) ||
-        !(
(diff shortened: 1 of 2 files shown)


ahocevar:
Actually this does not fix the issue. Needs more work.

ahocevar:
Force-pushed with a slightly more involved fix.

src/ol/PluggableMap.js
```diff
@@ -1016,15 +1016,21 @@ class PluggableMap extends BaseObject {
...
     if (!mapBrowserEvent.dragging) {
+      const rootNode = this.viewport_.getRootNode
+        ? this.viewport_.getRootNode()
+        : document;
+      const originalEvent = /** @type {PointerEvent} */ (mapBrowserEvent.originalEvent);
+      const target =
+        rootNode === document
+          ? /** @type {Node} */ (originalEvent.target)
+          : /** @type {ShadowRoot} */ (rootNode).elementFromPoint(
+              originalEvent.clientX,
```
tschaub, line 1028: This is subtle to have these two conditionals (here and line 1022).  Are there cases where `getRootNode` returns `document` and we don’t want to call `elementFromPoint`?

If not, it seems like the code would be clearer with just one `if` to handle the case where `getRootNode` exists.

ahocevar: `getRootNode()` returns `document`, except when in a web component, where it returns the shadow root. The check for `getRootNode` is only needed because Internet Explorer does not have a `getRootNode` function on a `HTMLElement`.

tschaub: Thanks for the explanation.

BHDBvde:
Tested for 10980 and it works.

PR #11025, opened 2020-05-10
Get rid of MapBrowserPointerEvent

ahocevar:
From a code perspective, It makes no sense to distinguish between `MapBrowserEvent` and `MapBrowserPointerEvent`. We had to do this a long time ago to entertain the Closure Compiler. But since we got rid of that, we can now make the original event we consume in `MapBrowserEvent` generic, and remove `MapBrowserPointerEvent`.


ahocevar:
Fixed TypeScript types, now ready for review.

Issue #10817, opened 2020-03-18
Flip feature Icon


hdlol:
Hello,

I want any method or help to be able to flip (vertically and/or horizontally) the feature's icon as it is possible for rotation.

I use OL 6.1.1

mike-000:
Currently you would need to load and flip the image (e.g. via a canvas element) before creating a new style.  There was some work started in #6100 which included allowing the scale to be an array.  If the relevant part could be revived maybe setting a negative x or y value could be used to flip?

hdlol:
Flip the image via canvas element is not a good solution (performance issue with thousands of features). Negative values for scales for example would be great.

hdlol:
Is it possible to add the flip feature icon ?

mike-000:
> Is it possible to add the flip feature icon ?

#11037 allows two dimensional scales.  A negative scale for either dimension will flip the icon in that dimension https://codesandbox.io/s/icon-scale-5mzvs

PR #11037, opened 2020-05-13
Allow icon and text styles to be scaled in two dimensions

mike-000:
This revives one of the aims of #6100 and also provides a fix for #10817 if a negative scale is used (images or text are flipped around their anchor point).  Similar to tileSize the API interface is `number | module:ol/size~Size` to be backward compatible, although scales are always stored as arrays which can be accessed using `getScaleArray()` for rendering.

A new example demonstrates an icon and label being scaled in both dimensions in a vector layer and in a vector context..


mike-000:
My aim was to be able to dynamically scale icons (in a style function if necessary) using `setScale()` while maintaining backward compatibility so that anyone who had set a single scale value continued to get a number returned by `getScale()` .  But instead of simply setting
    `this.scale_ = options.scale;`
I could add
    `this.scaleArray_ = toSize(options.scale);`
and use a non-API `getScaleArray()` method when rendering

 I had not initially intended to extent this to text but as point placement text shared the same logic it was easy to include (I hadn't even considered line placement until type checking picked up I had missed it).  I am not attempting to support RTL line placement either by flip or real RTL in this PR, but I think a `drawTextOnPathRTL()` would be needed as both would require right hand anchor points.

If there is going to be a scale option on text having it two dimensional would surely make it more useful, and it you are going to scale (stretch or squash) an icon you might want to make a corresponding change to the associated text?  I would be more likely to change the font size instead of using the same scale in both dimensions.

tschaub:
@mike-000 - you'll need to rebase on the latest `openlayers/master` to get a new script for the `deploy/netlify` check:
```
```

Apologies for the churn.

mike-000:
`toSize()` is now called only in the style constructors and `setScale` methods.  New arrays will be created during render only where scales are adjusted for pixel ratio (even that could be reduced by testing `pixelRatio == 1` and adding a new constant `defaultScale = [1, 1]` to  `ol/render/canvas.js`).  I have dropped the addition of scale to the labels example in favour of a new simple example showing both icon and text.

tschaub:
@mike-000, are you able to rebase these commits on master?  That would get rid of the merge commits an other unrelated history above.

If your "upstream" remote (the one for https://github.com/openlayers/openlayers.git or git@github.com:openlayers/openlayers.git) is named `openlayers`, a rebase would look like this:
```bash
```
If your remote is named something other than `openlayers`, you'd need to substitute that name.  Running `git remote --verbose` will list the names of your remotes.

After rebasing, if you force push, it should remove the extraneous commits above.
```bash
git push --force origin patch-11
```
(Substitute `origin` for the name of the remote corresponding to https://github.com/mike-000/openlayers.git or git@github.com:mike-000/openlayers.git).

ahocevar:
Thanks for your continued effort on this, @mike-000.

PR #11045, opened 2020-05-16
Install headless Chrome dependencies

tschaub:
The latest advice from CircleCI on trying to fix the issue launching headless Chrome is to add the dependencies used by the [threetreeslight/puppeteer orb](https://circleci.com/orbs/registry/orb/threetreeslight/puppeteer) (see the [troubleshooting doc](https://github.com/puppeteer/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-circleci)).

If this doesn't improve the situation, it might be worth revisiting the decision to run in headless mode in CI (see 8c89e17618e1b0486e7297fb52e35e0eec74dc0e and be17bfe85af859a506862f66cf641873299d0029).  If we do need to run in "headfull" mode, we could also try switching to GitHub actions and using something like [puppeteer-headful](https://github.com/marketplace/actions/puppeteer-headful).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,16 @@ jobs:
           - v2-dependencies-{{ checksum "package-lock.json" }}
 
       - run:
+          name: Install Headless Chrome Dependencies
+          command: |
+            sudo apt-get install -yq \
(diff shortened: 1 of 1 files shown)


tschaub:
This didn't make the problem go away: https://app.circleci.com/pipelines/github/openlayers/openlayers/3881/workflows/fe7deb16-881a-4c6a-8fa9-a0bc78dddb80/jobs/3999/steps

PR #11054, opened 2020-05-17
Add a code of conduct

tschaub:
This adds a code of conduct to the repo.


tschaub:
The failing checks will stop failing when #11053 is in.  They can be safely ignored here.

PR #11074, opened 2020-05-20
Copy properties when cloning geometries

### LINKED ISSUE #11071: Cloned geometry loses additional properties.
When we clone a geometry that has additional attributes, the clone does not contain these attributes.

```js
var geom = new ol.geom.Point([0, 0]);
geom.setProperties({foo: 'bar'});

console.log(geom.getProperties()); // {foo: 'bar'}
console.log(geom.clone().getProperties()); // return: {}, expected: {foo: 'bar'}
```

icholy:
Fixes #11071

### DIFF
--- a/src/ol/geom/Circle.js
+++ b/src/ol/geom/Circle.js
@@ -37,7 +37,13 @@ class Circle extends SimpleGeometry {
    * @api
    */
   clone() {
-    return new Circle(this.flatCoordinates.slice(), undefined, this.layout);
+    const circle = new Circle(
+      this.flatCoordinates.slice(),
+      undefined,
+      this.layout
+    );
+    circle.setProperties(this.getProperties(), true);
+    return circle;
   }
 
   /**
--- a/src/ol/geom/GeometryCollection.js
+++ b/src/ol/geom/GeometryCollection.js
@@ -69,6 +69,7 @@ class GeometryCollection extends Geometry {
   clone() {
     const geometryCollection = new GeometryCollection(null);
     geometryCollection.setGeometries(this.geometries_);
+    geometryCollection.setProperties(this.getProperties(), true);
     return geometryCollection;
   }
 
--- a/src/ol/geom/LineString.js
+++ b/src/ol/geom/LineString.js
@@ -87,7 +87,12 @@ class LineString extends SimpleGeometry {
    * @api
    */
   clone() {
-    return new LineString(this.flatCoordinates.slice(), this.layout);
+    const lineString = new LineString(
+      this.flatCoordinates.slice(),
+      this.layout
+    );
+    lineString.setProperties(this.getProperties(), true);
+    return lineString;
   }
 
   /**
--- a/src/ol/geom/LinearRing.js
+++ b/src/ol/geom/LinearRing.js
@@ -58,7 +58,12 @@ class LinearRing extends SimpleGeometry {
    * @api
    */
   clone() {
-    return new LinearRing(this.flatCoordinates.slice(), this.layout);
+    const linearRing = new LinearRing(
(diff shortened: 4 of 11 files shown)


tschaub:
This will generate a fair bit of unnecessary garbage for people who don't use their geometries to store arbitrary data (`geometry.getProperties()` always creates a new object which is no longer referenced after `geometry.setProperties()`).  We clone geometries during render, so it doesn't feel like a great place to do unnecessary allocation.

Some options:
 * add an option to the clone method to include properties, default is false
 * add something like an `apply` method to objects that could be called during clone
 * document that people have to `clone = geometry.clone(); clone.setProperties(geometry.getProperties())` if they tack arbitrary properties on to their geometries

tschaub:
@icholy - hope my comment above didn't come across the wrong way.  If we want to get something like this in, I think there are workable alternatives.

icholy:
@tschaub what you said makes sense.

michalzielanski:
@tschaub, it seems to me that `BaseObject` should never create an object to store values in advance. He should do this only when needed.

Something like that:
```diff
diff --git a/src/ol/Feature.js b/src/ol/Feature.js
index 590094057..facb98d06 100644
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -128,7 +128,7 @@ class Feature extends BaseObject {
    * @api
    */
   clone() {
-    const clone = new Feature(this.getProperties());
+    const clone = new Feature(this.hasProperties() ? this.getProperties() : undefined);
     clone.setGeometryName(this.getGeometryName());
     const geometry = this.getGeometry();
     if (geometry) {
diff --git a/src/ol/Object.js b/src/ol/Object.js
index a462129c1..5ea80d6b0 100644
--- a/src/ol/Object.js
+++ b/src/ol/Object.js
@@ -97,7 +97,7 @@ class BaseObject extends Observable {
      * @private
      * @type {!Object<string, *>}
      */
-    this.values_ = {};
+    this.values_;
 
     if (opt_values !== undefined) {
       this.setProperties(opt_values);
@@ -112,7 +112,7 @@ class BaseObject extends Observable {
    */
   get(key) {
     let value;
-    if (this.values_.hasOwnProperty(key)) {
+    if (this.values_ && this.values_.hasOwnProperty(key)) {
       value = this.values_[key];
     }
     return value;
@@ -124,7 +124,7 @@ class BaseObject extends Observable {
    * @api
    */
   getKeys() {
-    return Object.keys(this.values_);
+    return this.values_ && Object.keys(this.values_) || [];
   }
 
   /**
@@ -133,7 +133,14 @@ class BaseObject extends Observable {
    * @api
    */
   getProperties() {
-    return assign({}, this.values_);
+    return this.values_ && assign({}, this.values_) || {};
+  }
+
+  /**
+   * @return {boolean}
+   */
+  hasProperties() {
+    return !!this.values_;
   }
 
   /**
@@ -156,11 +163,12 @@ class BaseObject extends Observable {
    * @api
    */
   set(key, value, opt_silent) {
+    const values = this.values_ || (this.values_ = {});
     if (opt_silent) {
-      this.values_[key] = value;
+      values[key] = value;
     } else {
-      const oldValue = this.values_[key];
-      this.values_[key] = value;
+      const oldValue = values[key];
+      values[key] = value;
       if (oldValue !== value) {
         this.notify(key, oldValue);
       }
@@ -187,7 +195,7 @@ class BaseObject extends Observable {
    * @api
    */
   unset(key, opt_silent) {
-    if (key in this.values_) {
+    if (this.values_ && key in this.values_) {
       const oldValue = this.values_[key];
       delete this.values_[key];
       if (!opt_silent) {

```

PR #11082, opened 2020-05-22
Fix outside window dragging

### LINKED ISSUE #11081: Panning broken when pointerup outside the browser window
Panning is currently broken when pointerup occurs outside the browser window. When panning again, the map jumps back to the start location of the previous drag. This issue can be seen in all examples, when making the browser window smaller and ending the pan sequence outside the browser window.

This is a regression introduced by #11024.

ahocevar:
Fixes #11081.

Before #11024, which introduced the #11081 regression, we checked for
```js
this.viewport_.getRootNode().contains(target)
```

In #11024, this was changed to

```js
this.viewport_.contains(target)
```

This is obviously not the same, and the reason for the problem. Now that we have a `rootNode` calculated already, the fix is easy.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1030,14 +1030,17 @@ class PluggableMap extends BaseObject {
               originalEvent.clientY
             );
       if (
-        this.overlayContainerStopEvent_.contains(target) ||
-        !(document.body.contains(target) || rootNode.contains(target))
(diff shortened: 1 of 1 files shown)


ahocevar:
I added another commit which resolves the same issue for maps in web components (not a regression, never worked). Because that if block is quite involved, I placed more comments so it is easier to understand.

PR #11106, opened 2020-05-26
Defer ZoomSlider initialization until its size is available

### LINKED ISSUE #11104: ZoomSlider initializaion wrong when not visible when added
**Describe the bug**
When the ZoomSlider element is added to the map and the map and/or the ZoomSlider are not displayed the element size evaluates to 0x0 and this is never corrected. 

**To Reproduce**
1. Go to https://codesandbox.io/s/zoomslider-init-kgl9o
2. The zoomslider is initially not displayed with css and will become visible after 1sec
3. The slider element is not usable because the calculated height is zero

**Expected behavior**
Do not initialize the slider when its size is not yet available.

M393:
Fixes #11104

### DIFF
--- a/src/ol/control/ZoomSlider.js
+++ b/src/ol/control/ZoomSlider.js
@@ -179,14 +179,16 @@ class ZoomSlider extends Control {
    * direction_ and also constrain the dragging of the thumb to always be within
    * the bounds of the container.
    *
+   * @return {boolean} Initialization successful
    * @private
    */
   initSlider_() {
     const container = this.element;
-    const containerSize = {
-      width: container.offsetWidth,
-      height: container.offsetHeight,
-    };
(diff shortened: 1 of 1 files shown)


src/ol/control/ZoomSlider.js
```diff
@@ -366,8 +368,8 @@ class ZoomSlider extends Control {
     if (!mapEvent.frameState) {
       return;
     }
-    if (!this.sliderInitialized_) {
-      this.initSlider_();
+    if (!this.sliderInitialized_ && !this.initSlider_()) {
+      return;
```
ahocevar, line 372: The slider can still get out of sync with the map, e.g. when a view with different resolution settings is assigned to the map.

PR #11146, opened 2020-06-06
Update JSTS example to use JSTS 2.3.0

bjornharrtell:


tschaub:
Thanks, @bjornharrtell.

PR #11148, opened 2020-06-07
Vectortile label flicker

### LINKED ISSUE #10169: ol6  vector tile  label  flicker
when  I use ol6 ， zoom the map,  the vector tile label  flicker,why this hapeen,Any help from you will be greatly appreciated!

ahocevar:
The goal of this pull request is to fix #10169.

First of all, this pull request fixes a bug: some declutter groups were previously added more than once, and empty declutter groups were also added to the list of declutter items. That issue caused some flickers during panning.

Also to reduce flickers during panning, we now consider an area around the viewport for decluttering. We start with the `renderBuffer` of the layer, and then we look at the width and height of the boxes of declutter candidates and increase the buffer dynamically.

To make decluttering more stable also during zooming, we now look at individual boxes of combined declutter items (i.e. text along lines, images combined with text). This leaves more options for declutter item placement, because no-go boxes more closely resemble the rendered content. This is especially true for streets that are not straight in horizontal or vertical direction.

There was also a fix in ol-mapbox-style, which caused labels to disappar. That issue is fixed with v6.1.2, which we also use in ol since yesterday.


ahocevar:
This is now ready for review.

mapbox-vector-layer example before this change:

before

mapbox-vector-layer example after this change:

after

Issue #11102, opened 2020-05-25
Mouse interactions with tabIndex set on map container do not work as expected


dellisd:
**Describe the bug**
Because of the change made in #10291, if a map's container element has a `tabIndex` attribute set the `DragPan` and `MouseWheelZoom` interactions no longer work until the container element has focus. While this fixes some behaviours for mobile/touch devices, the behaviour for mouse input seems (in my opinion) broken.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/accessible.html
2. Try to zoom by scrolling, or panning by dragging _without the map being focused_
3. Interactions do not occur

**Expected behavior**
Mouse input behaviour should behave as before where the map can be interacted with using the mouse without the map needing to first be focused when it has a `tabIndex` attribute set.

ahocevar:
The behavior you are seeing is intentional. The reason is the followig: Consider a map on a scrollable page, e.g. in a blog post. You should always be able to scroll the page. Only after clicking on the map, you should be able to interact with the map.

I'd be curious to hear your use case, where the current behavior feels inappropriate.

dellisd:
My use case is an application centered around a map with no scrolling involved. Given the way my application is laid out, there would never be an instance where a scroll or drag behaviour would interfere with what the user expected and so requiring the map to be clicked first isn't desirable. 

There are other things in the application such as text fields that could be selected by the user and cause the map to lose focus and require the map to be clicked again. Re-focusing the map any time an element is blurred might be a workaround, but it's not a great solution.

I understand the use case you described there and I think the current behaviour is useful, but I do think it should also be possible to disable it.

ahocevar:
Good point. Since #10291, the `onFocusOnly` option of `ol/interaction#defaults` has no effect any more. #11135 restores that option, so you will be able to turn off the focus behavior again.

PR #11165, opened 2020-06-09
Use drag-pan default condition with onFocusOnly

### LINKED ISSUE #11162: Rotation not working correctly in dev version
**Describe the bug**
Using alt shift drag to rotate the view in the dev version is also panning the map instead of simply rotating around the view center.

**To Reproduce**
Open https://openlayers.org/en/latest/examples/simple.html zoom in 2 levels to avoid any conflict with multiWorld: false and use alt shift drag to rotate.  The view rotates around the center as expected.  Resetting the rotation restores the view to its original setting.

Repeat with https://openlayers.org/en/master/examples/simple.htm  The view rotates apparently around a random position as it also pans.  Resetting the rotation does not restore the view to its original position.

**Expected behavior**
alt shift drag should not pan the view.

ahocevar:
Fixes #11162.

I think the most convenient way to conditionally apply different conditions is to have a `chain(...conditions)` function. This pull request introduces such a function internally, and it also adds `onFocusOnly` options to the `DragPan` and `MousewheelZoom` interactions. This should make it easier for users to configure these interactions properly, in cases like the one described in #11102.

### DIFF
--- a/src/ol/events/condition.js
+++ b/src/ol/events/condition.js
@@ -14,6 +14,29 @@ import {assert} from '../asserts.js';
  */
 
 /**
+ * Creates a condition function that is only fulfilled when a chain of conditions pass.
+ * @param {...Condition} var_args Conditions to check.
+ * @return {Condition} Condition function that checks a chain of conditions.
+ */
+export function chain(var_args) {
+  const conditions = arguments;
+  /**
+   * @param {import("../MapBrowserEvent.js").default} event Event.
+   * @return {boolean} All conditions passed.
+   */
+  return function (event) {
+    let pass = true;
+    for (let i = 0, ii = conditions.length; i < ii; ++i) {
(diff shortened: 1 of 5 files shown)


src/ol/events/condition.js
```diff
@@ -13,6 +13,29 @@ import {assert} from '../asserts.js';
  * @typedef {function(this: ?, import("../MapBrowserEvent.js").default): boolean} Condition
  */
 
+/**
+ * Creates a condition function that is only fulfilled when a chain of conditions pass.
+ * @param {...Condition} var_args Conditions to check.
+ * @return {Condition} Condition function that checks a chain of conditions.
+ */
+export function chain(var_args) {
```
tschaub, line 21: If we want to make this part of the API in the future, I'd be tempted to name it `all` or `and`.

PR #11188, opened 2020-06-15
Late object initialization

michalzielanski:
This pull request solves the problem mentioned by @tschaub in a [comment](/openlayers/openlayers/pull/11074#issuecomment-631787430) to PR #11074.

It seems to me that BaseObject should never create an object to store values in advance. He should do this only when needed.

### DIFF
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -128,7 +128,9 @@ class Feature extends BaseObject {
    * @api
    */
   clone() {
-    const clone = new Feature(this.getProperties());
+    const clone = new Feature(
+      this.hasProperties() ? this.getProperties() : undefined
+    );
     clone.setGeometryName(this.getGeometryName());
     const geometry = this.getGeometry();
     if (geometry) {
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -129,7 +129,7 @@ class Feature extends BaseObject {
    */
   clone() {
     const clone = new Feature(
-      this.hasProperties() ? this.getProperties() : undefined
+      this.hasProperties() ? this.getProperties() : null
     );
     clone.setGeometryName(this.getGeometryName());
     const geometry = this.getGeometry();
--- a/src/ol/Object.js
+++ b/src/ol/Object.js
@@ -95,7 +95,7 @@ class BaseObject extends Observable {
 
     /**
      * @private
-     * @type {?Object<string, *>}
+     * @type {Object<string, *>}
      */
     this.values_ = null;
 
--- a/src/ol/Object.js
+++ b/src/ol/Object.js
@@ -4,7 +4,7 @@
 import Event from './events/Event.js';
 import ObjectEventType from './ObjectEventType.js';
 import Observable from './Observable.js';
-import {assign} from './obj.js';
+import {assign, isEmpty} from './obj.js';
 import {getUid} from './util.js';
 
 /**
@@ -94,10 +94,10 @@ class BaseObject extends Observable {
     getUid(this);
 
     /**
+     * @name values_
      * @private
      * @type {!Object<string, *>}
      */
-    this.values_ = {};
 
     if (opt_values !== undefined) {
       this.setProperties(opt_values);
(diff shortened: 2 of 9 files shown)


src/ol/Object.js
```diff
@@ -94,10 +94,10 @@ class BaseObject extends Observable {
     getUid(this);
 
     /**
+     * @name values_
      * @private
      * @type {!Object<string, *>}
      */
-    this.values_ = {};
```
tschaub, line 100: Rather than adding `@name`, this would be more consistent with the rest of the library:
```js
/**
 * @private
 * @type {!Object<string, *>}
 */
this.values_ = null;
```

michalzielanski: When I add the declaration `this.listeners_;` in the [Target class](/michalzielanski/openlayers/blob/517c9e6b52e1f88ad74bf9b998f7d4d75094a34a/src/ol/events/Target.js#L68) an _"TS2339: Property 'indexOf' does not exist on type '{}'"_ error appears on the line:
```js
    if (listeners.indexOf(listener) === -1) {
```
Adding types in the comments:
```js
let listeners = this.listeners_ || (this.listeners_ = /** @type {Object<string, Array<import("../events.js").Listener>>} */ {});
listeners = listeners[type] || (listeners[type] = /** @type {Array<import("../events.js").Listener>} */ []);
```
does not solve the problem. Is there a better solution than `@ts-ignore`?

ahocevar: There is:
```js
/**
 * @private
 * @type {Object<string, *>}
 */
this.values_ = null;

Comment (michalzielanski) [reply src/ol/Object.js:100]: I have it this way (no `!` sign). The whole code looks like this:
```js
  constructor(opt_target) {
    // (...)

    /**
     * @private
     * @type {Object<string, Array<import("../events.js").Listener>>|undefined}
     */
    this.listeners_;
  }

  addEventListener(type, listener) {
    if (!type || !listener) {
      return;
    }
    let listeners =
      this.listeners_ ||
      (this.listeners_ = /** @type {Object<string, Array<import("../events.js").Listener>>} */ {});
    listeners =
      listeners[type] ||
      (listeners[
        type
      ] = /** @type {Array<import("../events.js").Listener>} */ []);
    if (listeners.indexOf(listener) === -1) {
      listeners.push(listener);
    }
  }
```
but the error occurs.


I also tried:
```js
/**
 * @private
 * @type {Object<string, Array<import("../events.js").Listener>>} // wo undefined
 */
this.listeners_;

/**
 * @private
 * @type {Object<string, Array<import("../events.js").Listener>>}
 */
this.listeners_ = null; // null assignment

/**
 * @private
 * @type {?Object<string, Array<import("../events.js").Listener>>} // question mark
 */
this.listeners_ = null;
```
and `@type {!` here:
```js
let listeners = this.listeners_ || (this.listeners_ = /** @type {!Object<string, Array<import("../events.js").Listener>>} */ {});
listeners = listeners[type] || (listeners[type] = /** @type {!Array<import("../events.js").Listener>} */ []);
```
Nothing helps.
All that remains is to use `@ts-ignore`?

ahocevar: TypeScript does not like re-assigning a variable to a value with a different type. I took the liberty to make a change that fixes the type error, in addition to removing all `@name` annotations.

src/ol/Feature.js
```diff
@@ -128,7 +128,9 @@ class Feature extends BaseObject {
    * @api
    */
   clone() {
-    const clone = new Feature(this.getProperties());
+    const clone = new Feature(
+      this.hasProperties() ? this.getProperties() : undefined
```
tschaub, line 132: We could also make it so `getProperties` is efficient without adding a new API method `hasProperties`.

src/ol/Object.js
```diff
@@ -133,7 +133,14 @@ class BaseObject extends Observable {
    * @api
    */
   getProperties() {
-    return assign({}, this.values_);
+    return (this.values_ && assign({}, this.values_)) || {};
```
tschaub, line 136: I see that to maintain backwards compatibility here we should return `{}` instead of `undefined` (so never mind https://github.com/openlayers/openlayers/pull/11188/files#r444566590).

michalzielanski:
@tschaub, @ahocevar, please review the changes from 235babf.

PR #11190, opened 2020-06-16
Correct vector layer clipping with rotation

### LINKED ISSUE #11186: Vector layer extent option does not rotate with the view
**Describe the bug**
If an extent option is specified on a vector layer the output is clipped by the unrotated extent when the view is rotated.

**To Reproduce**
Open https://codesandbox.io/s/vector-layer-stsxf and rotate the view.

**Expected behavior**
The clipping extent should rotate with the view.  VectorImage layers seem to behave much as expected but there appears to be some excess clipping in some situations.

mike-000:
Fixes #11186

This fixes another regression introduced in v6.0.0 with similar cause to #9585


tschaub:
Thanks for the fix and the test, @mike-000.

PR #11194, opened 2020-06-17
ol/geom/LineString#getCoordinateAt() to return Z and M values if available

### LINKED ISSUE #11177: ol/geom/LineString#getCoordinateAt() does not return Z and M values
**Describe the bug**
It seems easy for `ol/geom/flat/interpolate#interpolatePoint` to return Z and M values if included in the stride, but unlike `getCoordinateAtM()` and `getClosestPoint()` it does not.  A workaround to get the interpolated elevation and time at km `i `of an `n` km route is to use `lineString.getClosestPoint(lineString.getCoordinateAt(i / n)) `but that is very inefficient.

**Expected behavior**
`ol/geom/flat/interpolate#interpolatePoint` should interpolate all values in the stride to make `getCoordinateAt()` consistent with other methods

mike-000:
Fixes #11177

Add option for `ol/geom/flat/interpolate#interpolatePoint` to interpolate all dimensions in the stride to make `getCoordinateAt()` consistent with other API methods.  Keep the default behaviour of `interpolatePoint` as returning 2 dimensions for compatibility with internal methods such as `getFlatMidpoints`.

### DIFF
--- a/src/ol/geom/LineString.js
+++ b/src/ol/geom/LineString.js
@@ -211,7 +211,8 @@ class LineString extends SimpleGeometry {
       this.flatCoordinates.length,
       this.stride,
       fraction,
-      opt_dest
+      opt_dest,
+      this.stride
     );
   }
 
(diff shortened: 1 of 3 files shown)


mike-000:
This is ready for review (test failed due an unrelated tileQueue error)

PR #11195, opened 2020-06-17
getExtent() to return getEmpty() if geometry is unprojectable

### LINKED ISSUE #11175: ol/geom/Geometry#getExtent() should return createEmpty() if geometry contains NaN coordinates
**Describe the bug**
Coordinates are quite reasonably transformed to NaN if they are impossible for the projection (e.g. the opposite side of the world in an orthographic projection).  However if any coordinate in any feature is NaN the entire layer fails to render as the geometry extent, and consequently the vector source extent, is returned as [NaN, NaN, NaN, NaN].  A simple fix would be for ol/geom/Geometry#getExtent() to return an empty extent if invalid.  https://codesandbox.io/s/orthographic-2e8vl

ol/layer/Graticule is also failing in an orthographic projection and may need to validate further.

**To Reproduce**
See https://codesandbox.io/s/orthographic-2e8vl for a patched example which displays visible countries correctly.  Remove the patch and reload to see the problem.

**Expected behavior**
Out-of-range features should not affect rendering of in-range features.

mike-000:
Fixes #11175

Return getEmpty() if unprojectable geometry coordinates have been correctly transformed to NaN to prevent the entire source extent being "tainted" and to allow projectable features to be rendered.

### DIFF
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -8,7 +8,12 @@ import {
   compose as composeTransform,
   create as createTransform,
 } from '../transform.js';
-import {createEmpty, getHeight, returnOrUpdate} from '../extent.js';
+import {
+  containsExtent,
+  createEmpty,
+  getHeight,
+  returnOrUpdate,
+} from '../extent.js';
 import {get as getProjection, getTransform} from '../proj.js';
 import {memoizeOne} from '../functions.js';
 import {transform2D} from './flat/transform.js';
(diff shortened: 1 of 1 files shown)


src/ol/geom/Geometry.js
```diff
@@ -170,6 +175,14 @@ class Geometry extends BaseObject {
...
     if (this.extentRevision_ != this.getRevision()) {
       this.extent_ = this.computeExtent(this.extent_);
+      if (
+        !containsExtent(
+          [-Infinity, -Infinity, Infinity, Infinity],
+          this.extent_
+        )
+      ) {
+        this.extent_ = createEmpty();
+      }
```
ahocevar, line 185:
```suggestion
      const extent = this.computeExtent(this.extent_);
      if (extent[0] === NaN || extent[1] === NaN) {
        createOrUpdateEmpty(extent);
      }
```
would be more efficient and create less garbage.

mike-000: done
(although `=== NaN` doesn't work as `NaN === NaN` returns false)

ahocevar: Right, sorry for that. I meant `isNaN(extent[0])`

PR #11197, opened 2020-06-17
Added '&&' logical operator to expressions

RydingM:
This PR adds support for an '&&' operator in the styling expressions. It will only return true if both value1 and value2 are true. This can be used when you only want to show features that conditions are both true, rather then the '==' operator which will show the features if values 1 and 2 are the same. I did this because I ran into an issue where I need to have more complex filtering that couldn't be achieved with the current set.

### DIFF
--- a/src/ol/style/expressions.js
+++ b/src/ol/style/expressions.js
@@ -53,6 +53,8 @@ import {asArray, isStringColor} from '../color.js';
  *   * `['==', value1, value2]` returns `true` if `value1` equals `value2`, or `false` otherwise.
  *   * `['!=', value1, value2]` returns `true` if `value1` does not equal `value2`, or `false` otherwise.
  *   * `['!', value1]` returns `false` if `value1` is `true` or greater than `0`, or `true` otherwise.
+ *   * `['all', value1, value2, ...]` returns `true` if all the inputs are `true`, `false` otherwise.
+ *   * `['any', value1, value2, ...]` returns `true` if any of the inputs are `true`, `false` otherwise.
  *   * `['between', value1, value2, value3]` returns `true` if `value1` is contained between `value2` and `value3`
  *     (inclusively), or `false` otherwise.
  *
@@ -494,7 +496,6 @@ Operators['^'] = {
     )})`;
   },
 };
-
 Operators['>'] = {
   getReturnType: function (args) {
     return ValueTypes.BOOLEAN;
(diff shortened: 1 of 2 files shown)


RydingM:
Cheers for the comments, I am more then happy to make the changes suggested!

src/ol/style/expressions.js
```diff
@@ -590,6 +592,52 @@ Operators['!'] = {
...
+      );
+    }
+
+    return `(${expressionToGlsl(context, args[0], type)} && ${expressionToGlsl(
+      context,
+      args[1],
+      type
+    )})`;
+  },
+};
```
jahow, line 621: this should be removed I think

src/ol/style/expressions.js
```diff
@@ -590,6 +592,52 @@ Operators['!'] = {
...
+    }
+
+    return `(${expressionToGlsl(context, args[0])} && ${expressionToGlsl(
+      context,
+      args[1]
+    )})`;
+  },
+};
+
+Operators['any'] = getEqualOperator('||');
```
jahow, line 639: This should use the same code as the above operator, instead of the equality ones (which can only take two arguments).

RydingM:
Did you want me to rebase the linting commits into one? I had problems with my prettier in my IDE.

jahow:
> Also based on the mapbox documentation it seems the and is called all, https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/?size=n_10_n#all, happy to rename it to this.

You're absolutely right, sorry for giving wrong indications... It should be named `all`.

I've looked at your new commits, maybe I wasn't clear enough but for me both these operators have almost similar logic. However, you only accept 2 arguments for the `and`(`all`) operator instead of two and more. Besides, the code for both can probably be factorized. Lastly, I think it would be nice to specify that these operators can take 2 or more arguments in the documentation [here](https://github.com/openlayers/openlayers/pull/11197/files#diff-26e460563a3a3adcd21d6eb7626fce29R56-R57).

> Did you want me to rebase the linting commits into one? I had problems with my prettier in my IDE.

Ideally all your commits should be squashed into one. Thanks!

jahow:
@RydingM do you think you can keep working on this? Otherwise I'll try to pick it up when I have time. Thanks!

RydingM:
> @RydingM do you think you can keep working on this? Otherwise I'll try to pick it up when I have time. Thanks!

Apologies I have been busy recently, I have got the all the code ready to push up. Can do this tonight!

jahow:
That would be amazing, thanks a lot.

RydingM:
> Great! I've added 2 suggestions but they're optional, I'm leaving it up to you to integrate them. You will have to squash all your commits into one before merging this PR though. Let me know if you need indications.

Its ready for merge however I cannot merge it due to netlify/ol-site/deploy-preview being expected but no status has been reported.

src/ol/style/expressions.js
```diff
@@ -590,6 +591,29 @@ Operators['!'] = {
...
+      result = args
+        .map((arg) => expressionToGlsl(context, arg))
+        .join(` ${operator} `);
+      result = `(${result})`;
+      return result;
+    },
+  };
+}
+
+Operators['and'] = getDecisionOperator('&&');
```
jahow, line 615: Looks like we're back with the name `and` instead of `all`... did something go wrong when squashing the commits?

jahow:
Thanks a lot for your contribution @RydingM!

PR #11201, opened 2020-06-19
Prepare for a new default branch name

tschaub:
This updates things in preparation for a new default branch name.


sbrunner:
For my general knowledge, why do you make this name change?

ahocevar:
@sbrunner See https://twitter.com/mislav/status/1270388510684598272.

PR #11202, opened 2020-06-20
Branch name changes

tschaub:
This changes a few links so we point to the latest release.  And it makes it so we publish `dev` releases on pushes to `main`.


tschaub:
Merging so we can get site builds on pull requests.

PR #11203, opened 2020-06-21
fix linearRingIsClockwise check in format/MVT

### LINKED ISSUE #8148: mvt vectortile with multiPolygon not displayed
I want to dispaly some MVT vectortile ,but I found some polygon is not displayed like this.The white background color should be same as others.
image

I debug the source code and found that

image

`ends` does not match `endss`.  I think ` endss.push(ends.slice(prevEndIndex, i))` has some problems ， when `i=0` and `prevEndIndex=0` , `ends[0]` has to be `[]` .

I modify` endss.push(ends.slice(prevEndIndex, i))` to `endss.push(ends.slice(prevEndIndex))`,the background color is displayed correctly.
image

Is this a bug?

songyumeng:
fixes #8148

judge whether it is multiPolygon or Polygon:
[https://github.com/mapbox/vector-tile-js/blob/a9a9102/lib/vectortilefeature.js#L195-L223](https://github.com/mapbox/vector-tile-js/blob/a9a9102/lib/vectortilefeature.js#L195-L223)

image

### DIFF
--- a/src/ol/format/MVT.js
+++ b/src/ol/format/MVT.js
@@ -19,7 +19,6 @@ import RenderFeature from '../render/Feature.js';
 import Units from '../proj/Units.js';
 import {assert} from '../asserts.js';
 import {get} from '../proj.js';
-import {linearRingIsClockwise} from '../geom/flat/orient.js';
 
 /**
  * @typedef {Object} Options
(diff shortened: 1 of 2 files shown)


ahocevar:
Thanks @songyumeng. I would like to add more tests for this. Would you mind sharing the geometries you show in your screenshots ad geojson or mvt for that purpose?

src/ol/format/MVT.js
```diff
@@ -440,4 +445,28 @@ function getGeometryType(type, numEnds) {
...
 
+/**
+ * Determines Flat coordinates is clockwise in MVT.
+ * @param {Array<number>} flatCoordinates Flat coordinates.
+ * @param {number} offset Offset.
+ * @param {number} end End.
+ * @param {number} stride Stride.
+ * @return {boolean} Is clockwise in MVT.
+ */
+function linearRingIsClockwise(flatCoordinates, offset, end, stride) {
```
ahocevar, line 456: Couldn't you just `NOT` the result of the imported `linearRingIsClockwise` function, instead of defining a new one here?

songyumeng: I have thought about "imported linearRingIsClockwise function",But I am confused about the case where it is equal to 0. when the result is 0,the `orient.linearRingIsClockwise ` will return `false`,but it is not counter-clockwise in mvt either.

songyumeng:
> Thanks @songyumeng. I would like to add more tests for this. Would you mind sharing the geometries you show in your screenshots ad geojson or mvt for that purpose?

[example](https://codesandbox.io/s/vectortile-zw6pc)
mvt:
[7-64-55.zip](https://github.com/openlayers/openlayers/files/4810639/7-64-55.zip)

PR #11219, opened 2020-06-25
Make proj4 transforms behave like built-in transforms

### LINKED ISSUE #10928: Transforms between global projections using proj4 don't wrap as they do with predefined transforms

ahocevar:
This pull request adds a `createSafeCoordinateTransform()` function, which shims a transform function so it supports wrapping x coordinates and clamping input coordinates to the destination projection's validity extent. That way, transform functions e.g. from proj4 behave the same way as the built-in transform functions between EPSG:4326 and EPSG:3857.

The added tests make use of the fact that proj4 does not overwrite existing projections (EPSG:4326 and EPSG:3857), but registers projections for codes that are not registered yet (WGS84 and GOOGLE). So we can compare the proj4 transforms with the built-in transforms.

Replaces and closes #10958
Fixes #10928

### DIFF
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -410,17 +410,22 @@ export function toStringXY(coordinate, opt_fractionDigits) {
  * @return {Coordinate} The coordinate within the real world extent.
  */
 export function wrapX(coordinate, projection) {
-  const projectionExtent = projection.getExtent();
-  if (
-    projection.canWrapX() &&
-    (coordinate[0] < projectionExtent[0] ||
-      coordinate[0] >= projectionExtent[2])
-  ) {
-    const worldWidth = getWidth(projectionExtent);
-    const worldsAway = Math.floor(
-      (coordinate[0] - projectionExtent[0]) / worldWidth
-    );
-    coordinate[0] -= worldsAway * worldWidth;
+  const worldsAway = getWorldsAway(coordinate, projection);
+  if (worldsAway) {
+    coordinate[0] -= worldsAway * getWidth(projection.getExtent());
   }
   return coordinate;
+}
+
+export function getWorldsAway(coordinate, projection) {
+  const projectionExtent = projection.getExtent();
+  let worldsAway = 0;
+  if (
+    projection.canWrapX() &&
+    (coordinate[0] < projectionExtent[0] || coordinate[0] > projectionExtent[2])
+  ) {
+    const worldWidth = getWidth(projectionExtent);
+    worldsAway = Math.floor((coordinate[0] - projectionExtent[0]) / worldWidth);
+  }
+  return worldsAway;
 }
--- a/src/ol/coordinate.js
+++ b/src/ol/coordinate.js
@@ -406,26 +406,35 @@ export function toStringXY(coordinate, opt_fractionDigits) {
  * exclusive.
  *
  * @param {Coordinate} coordinate Coordinate.
- * @param {import("./proj/Projection.js").default} projection Projection
+ * @param {import("./proj/Projection.js").default} projection Projection.
  * @return {Coordinate} The coordinate within the real world extent.
  */
 export function wrapX(coordinate, projection) {
-  const worldsAway = getWorldsAway(coordinate, projection);
+  const worldWidth = getWidth(projection.getExtent());
+  const worldsAway = getWorldsAway(coordinate, projection, worldWidth);
   if (worldsAway) {
-    coordinate[0] -= worldsAway * getWidth(projection.getExtent());
+    coordinate[0] -= worldsAway * worldWidth;
   }
   return coordinate;
 }
-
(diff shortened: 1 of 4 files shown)


src/ol/proj.js
```diff
@@ -625,6 +626,52 @@ export function fromUserExtent(extent, destProjection) {
...
+      worldsAway = getWorldsAway(coord, sourceProj);
+      const sourceExtent = sourceProj.getExtent();
+      if (worldsAway && sourceExtent) {
+        x = coord[0] - worldsAway * getWidth(sourceExtent);
+      }
+    }
+    const destExtent = destProj.getExtent();
+    if (destExtent) {
+      const clampMin = inverse(destExtent.slice(0, 2));
+      const clampMax = inverse(destExtent.slice(2, 4));
```
tschaub, line 658: These may not actually represent the min/max y-values (right side could be lower than left, left higher than right, maybe bottom inverts to top).  Do we care?

src/ol/coordinate.js
```diff
@@ -410,17 +410,22 @@ export function toStringXY(coordinate, opt_fractionDigits) {
...
  */
 export function wrapX(coordinate, projection) {
+  const worldsAway = getWorldsAway(coordinate, projection);
+  if (worldsAway) {
+    coordinate[0] -= worldsAway * getWidth(projection.getExtent());
+  }
+  return coordinate;
+}
+
+export function getWorldsAway(coordinate, projection) {
```
tschaub, line 420: If this were something like `getWorldOffset` and returned `worldsAway * extentWidth`, we could avoid calculating the extent width again in the places it is called.

tschaub: I see now that it is also used on the transformed coordinate.  It could take an extent width instead of a projection, but perhaps not as nice.

src/ol/proj.js
ahocevar: Right, that's a problem, e.g. with polar projections. I'm currently looking into how to solve this.

src/ol/coordinate.js
ahocevar: I solved this now by optionally providing an already calculated world width.

src/ol/proj.js
ahocevar: Changed the code now so we only clamp source coordinates to the transformed destination extent when transforming the original coordinates fails.

ahocevar: The new code also handles the case where bottom and top are inverted.

src/ol/proj.js
```diff
@@ -625,6 +626,66 @@ export function fromUserExtent(extent, destProjection) {
...
+  forward,
+  inverse
+) {
+  return function (coord) {
+    let x, y, worldsAway;
+    if (sourceProj.canWrapX()) {
+      const sourceExtent = sourceProj.getExtent();
+      const sourceExtentWidth = getWidth(sourceExtent);
+      worldsAway = getWorldsAway(coord, sourceProj, sourceExtentWidth);
+      if (worldsAway && sourceExtent) {
```
tschaub, line 652: It looks like the `sourceExtent` here is not needed (covered by `canWrapX()` and implied by `getWidth(sourceExtent)`).

src/ol/proj.js
```diff
@@ -625,6 +626,66 @@ export function fromUserExtent(extent, destProjection) {
...
+        const y2 = corner2[2];
+        if (isFinite(x1) && isFinite(x2)) {
+          x = clamp(
+            x == undefined ? coord[1] : x,
+            Math.min(x1, x2),
+            Math.max(x1, x2)
+          );
+        }
+        if (isFinite(y1) && isFinite(y2)) {
+          y = clamp(coord[1], Math.min(y1, y2), Math.max(y1, y2));
```
tschaub, line 676: I think you'd need to transform all four corners to determine which resulted in the min/max y.  As this is written, if the top-left corner transforms to a higher y-value than the top-right corner, the max y will be smaller than it should.

ahocevar:
Thanks for the continued review effort @tschaub. I have addressed your comments.

tschaub:
Nice usability improvement, @ahocevar.

PR #11224, opened 2020-06-26
Quote . in RegExp

mohawk2:
I just updated OpenLayers.js in https://github.com/metacpan/metacpan-web/pull/2329. CodeQL [points out](https://github.com/metacpan/metacpan-web/pull/2329/checks?check_run_id=812921165) that your RegExp for host-matching includes `.` where you obviously meant `\.`. This fixes that.

### DIFF
--- a/src/ol/format/IIIFInfo.js
+++ b/src/ol/format/IIIFInfo.js
@@ -145,15 +145,9 @@ IIIF_PROFILE_VALUES['none'] = {
   },
 };
 
-const COMPLIANCE_VERSION1 = new RegExp(
-  '^https?://library.stanford.edu/iiif/image-api/(1.1/)?compliance.html#level[0-2]$'
-);
-const COMPLIANCE_VERSION2 = new RegExp(
-  '^https?://iiif.io/api/image/2/level[0-2](.json)?$'
-);
-const COMPLIANCE_VERSION3 = new RegExp(
-  '(^https?://iiif.io/api/image/3/level[0-2](.json)?$)|(^level[0-2]$)'
-);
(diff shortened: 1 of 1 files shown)


mohawk2:
Your CI is failing this PR because `prettier` is rejecting `RegExp` with `\.` in. That probably needs fixing.

MoonE:
I believe you need to double the backslashes when constructing the regexp from a string.
```js
const COMPLIANCE_VERSION1 = new RegExp(
  '^https?://library\\.stanford\\.edu/iiif/image-api/(1\\.1/)?compliance\.html#level[0-2]$'
);
```
But you should be able to use a regexp literal here instead as it is not dynamic. With this you only need to escape dots once, but also need to escape the slashes.
```js
const COMPLIANCE_VERSION1 = 
  /'^https?:\/\/library\.stanford\.edu\/iiif\/image-api\/(1\.1\/)?compliance\.html#level[0-2]$/;
```
And while you are at it you could also use a non-capturing group for the optional groups as these are only used to check if it is a match:
```js
const COMPLIANCE_VERSION1 = 
  /'^https?:\/\/library\.stanford\.edu\/iiif\/image-api\/(?:1\.1\/)?compliance\.html#level[0-2]$/;
```

Best of luck.

ahocevar:
Better to not use the RegExp constructor, and there is more to escape than just the dots:
```js
const COMPLIANCE_VERSION1 = /^https?:\/\/library\.stanford\.edu\/iiif\/image-api\/(1\.1\/)\?compliance\.html#level[0-2]$/;
```

mohawk2:
Overnight I realised you were quite right, and also concluded it should be an RE literal. Great minds!

PR #11225, opened 2020-06-28
Use 'source-over' to render reprojection edges

mike-000:
Reprojection edges are only visible in the development examples https://openlayers.org/en/main/examples/reprojection.html and https://openlayers.org/en/main/examples/reprojection-by-code.html where there are no tiles, or there is partial transparency as in the United States overlay.  This is a consequence of the globalCompositeOperation having been set to 'lighter'.

### DIFF
--- a/src/ol/reproj.js
+++ b/src/ol/reproj.js
@@ -398,6 +398,7 @@ export function render(
   if (opt_renderEdges) {
     context.save();
 
+    context.globalCompositeOperation = 'source-over';
     context.strokeStyle = 'black';
     context.lineWidth = 1;
 


ahocevar:
@pjsg, can you take a look at this please? What was the reason for setting the composite operation to `'lighter'` in the first place?

pjsg:
The quick response is that source-over doesn't work right when the tile is reprojected into something which is not aligned on pixel boundaries and (as I recall) the layer being drawn is partially transparent. The real problem is that (IMHO) we are missing compositing modes which deal with transparency in different ways. This is made more complex when the tile is mapped to a shape that has sloped edges -- then the rendering engine needs to figure out how to composite two different pixels (in separate operations) into a single pixel (which already has the layer beneath in it).

Having said all that, it turned out that some browsers have sufficiently bad bugs in their rendering engines that you can't deal with partial pixels. In this case, the current reprojection code switches to an alternate algorithm which using clipping regions which are directly mapped onto target pixels rather than the default approach which is to use triangles (which necessarily have partial pixels in them). 

I did add a rendering test when I was making these changes -- but whether they would fail or not in your case, I do not know.

I'll take a detailed look over the next couple of days....

pjsg:
Actually, now I've actually read your message more carefully, I suspect that your fix is good. When I was working trying to resolve the reprojection issues, I never used the render_edges option....

mike-000:
I have no problem with the use of `'lighter'` for diagonal stitching and unless the `renderReprojectionEdges` option is used there is no need to reset it as the context has no other purpose.

PR #11253, opened 2020-07-06
Only clamp to source extent for safe transforms

### LINKED ISSUE #11238: Layer reprojection broken by commit bef4d8a4
**Describe the bug**

I tried to update my app to use the current 'dev' version from npmjs. I found that it introduced rendering problems for the Azimuthal Equal Area projection in areas at the edge of the world (the world is projected into a circle with severe distortion at the edge of the circle). 

If I revert 
```
diff --git a/src/ol/proj/proj4.js b/src/ol/proj/proj4.js
index ca7c17c..f4fb255 100644
--- a/src/ol/proj/proj4.js
+++ b/src/ol/proj/proj4.js
@@ -60,8 +61,18 @@ export function register(proj4) {
           addCoordinateTransforms(
             proj1,
             proj2,
-            transform.forward,
-            transform.inverse
+            createSafeCoordinateTransform(
+              proj1,
+              proj2,
+              transform.forward,
+              transform.inverse
+            ),
+            createSafeCoordinateTransform(
+              proj2,
+              proj1,
+              transform.inverse,
+              transform.forward
+            )
           );
         }
       }
```

then it works correctly again. I'm not sure of what the use of `createSafeCoordinateTransform` is, but it seems to prevent transforms from behaving correctly at the edge of their validity. There are some transforms that repeat in `x` coordinates and I suspect that that is the case that `createSafeCoordinateTransform` handles. However, not all transforms behave like that. The reprojection code makes use of the fact that some points in the reprojected plane (i.e. the canvas) do not map to *any* valid point on the globe.

I can try and put together a jsfiddle if needed....

ahocevar:
This pull request makes the coordinate clamping to the validity extent in `ol/proj.createSafeCoordinateTransform` more straightforward: it only clamps the source coordinate to the validity extent of the source extent.

Fixes #11238.


ahocevar:
I updated this pull request to a much simpler approach that merely ensures sanity of the source coordinate.

PR #11267, opened 2020-07-11
Fix upright labels on vector tiles

### LINKED ISSUE #11264: Line placement labels don't always adjust to rotation since 6.1.0
**Describe the bug**
**To Reproduce**
See the example https://codepen.io/mike-000/pen/RwrryMv in https://github.com/openlayers/ol-mapbox-style/issues/279  Rotate the view and the street labels do not flip as they pass vertical.  It happens regardless of the declutter setting.  But change the OpenLayers build to 6.1.0 or earlier and repeat it works as expected, 6.1.1 or later is does not.  The version of olms does not affect it, and the style json works as expected with Mapbox GL, so it looks like an OpenLayers regression under some conditions as the Street Labels example works correctly in all releases.

**Expected behavior**
Line placement street labels should flip as the view is rotated so the text is not inverted.

ahocevar:
Fixes #11264


src/ol/render/canvas/Executor.js
```diff
@@ -957,7 +957,8 @@ class Executor {
               Math.abs(textScale[0]),
               measureAndCacheTextWidth,
               font,
-              cachedWidths
+              cachedWidths,
+              viewRotationFromTransform ? 0 : this.viewRotation_
```
jahow, line 961: Out of curiosity, in which situation does `viewRotationFromTransform` happen to be non zero?

ahocevar: in the case of layer renderers that rotate the executor groups, e.g. the vector layer renderer. As opposed to the vector tile layer renderer, which renders executor groups upright.

PR #11269, opened 2020-07-12
Register and unregister events in set target handler

### LINKED ISSUE #11231: MapBrowserEventHandler needs to be destroyed for listeners to be unregistered
**Describe the bug**
`MouseWheelZoom.conditionInternal_` throws errors when the map target is null due to the `hasAttribute` check.
I see this is changed on master but I believe this bug will still exist (https://github.com/openlayers/openlayers/commit/2cac6c4f39521470a899f1c6338aa47baea05392#diff-a08a90960184e04e1b7384a480a81d6a).

**To Reproduce**
1. Go to https://codesandbox.io/s/react-openlayers-xni5f?file=/src/App.js which contains a simple map example reloading every 5s and setting the previous map target to null (`myMap.setTarget(null);`)
2. Click on the map to drag but don't release for 5s.
3. Next mouse movements will trigger an error:
image
image

What I've found so far is
- Events need a map target or they crash
- Events are not deregisterd without `map.dispose` which propagates to `disposeInternal` of `MapBrowserEventHandler` where it unlistens avoiding the events to be called.

**Expected behavior**
I am not sure which solution is preferable:
- Have events check if they have a target before checking the `hasAttribute`
- Making `map.dispose` needed for destroy (https://github.com/openlayers/openlayers/issues/10856#issuecomment-643601271 seems like it is not needed). I believe this is cleaner to avoid memory leaks everywhere. Could you tell me the reasoning to not make the dispose methods public?

Let me know what you prefer @tschaub

ahocevar:
Fixes #11231

This pull request moves code from the `PluggableMap` constructor and `disposeInternal()` method to the set target handler. This avoids stale event listeners when the target is changed.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -282,17 +282,13 @@ class PluggableMap extends BaseObject {
      * @private
      * @type {MapBrowserEventHandler}
      */
-    this.mapBrowserEventHandler_ = new MapBrowserEventHandler(
-      this,
-      options.moveTolerance
-    );
-    const handleMapBrowserEvent = this.handleMapBrowserEvent.bind(this);
-    for (const key in MapBrowserEventType) {
-      this.mapBrowserEventHandler_.addEventListener(
-        MapBrowserEventType[key],
-        handleMapBrowserEvent
-      );
-    }
+    this.mapBrowserEventHandler_ = null;
+
+    /**
+     * @private
+     * @type {number}
+     */
+    this.moveTolerance_ = options.moveTolerance;
 
     /**
      * @private
@@ -305,18 +301,6 @@ class PluggableMap extends BaseObject {
      * @type {?Array<import("./events.js").EventsKey>}
      */
     this.keyHandlerKeys_ = null;
-
-    const handleBrowserEvent = this.handleBrowserEvent.bind(this);
-    this.viewport_.addEventListener(
-      EventType.CONTEXTMENU,
-      handleBrowserEvent,
-      false
-    );
-    this.viewport_.addEventListener(
-      EventType.WHEEL,
-      handleBrowserEvent,
-      PASSIVE_EVENT_LISTENERS ? {passive: false} : false
-    );
 
     /**
      * @type {Collection<import("./control/Control.js").default>}
@@ -550,19 +534,6 @@ class PluggableMap extends BaseObject {
    * Clean up.
    */
   disposeInternal() {
-    this.mapBrowserEventHandler_.dispose();
-    this.viewport_.removeEventListener(
-      EventType.CONTEXTMENU,
-      this.boundHandleBrowserEvent_
-    );
-    this.viewport_.removeEventListener(
-      EventType.WHEEL,
-      this.boundHandleBrowserEvent_
-    );
-    if (this.handleResize_ !== undefined) {
-      removeEventListener(EventType.RESIZE, this.handleResize_, false);
-      this.handleResize_ = undefined;
-    }
     this.setTarget(null);
     super.disposeInternal();
   }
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for any review.

EvertEt:
Updating the above codesandbox to `6.3.2-dev.1595058168339` still results in some error, will see if I can find the cause
image

I am also wondering if we call `mapBrowserEventHandler_.dispose()` during setTarget(null), shouldn't we also create it in setTarget instead of only once?

src/ol/PluggableMap.js
```diff
@@ -1162,12 +1137,36 @@ class PluggableMap extends BaseObject {
...
+        removeEventListener(EventType.RESIZE, this.handleResize_, false);
+        this.handleResize_ = undefined;
+      }
     } else {
       targetElement.appendChild(this.viewport_);
       if (!this.renderer_) {
         this.renderer_ = this.createRenderer();
       }
 
+      const handleBrowserEvent = this.handleBrowserEvent.bind(this);
```
EvertEt, line 1159: Curious as to why we bind again instead of using `boundHandleBrowserEvent_`. Will a different bind be able to get matched over `addEventListener` and `removeEventListener`?

ahocevar:
@EvertEt This has not been merged yet, so it won't be in the latest dev version yet.

EvertEt:
My bad! Is it possible to test it online or should I build from this commit?

ahocevar:
@EvertEt It's best to build from this branch.

src/ol/PluggableMap.js
ahocevar: I just realized I did not push everything I had locally. Hold on, update to come.

ahocevar:
@EvertEt I just updated the pull request with the latest I had locally.

src/ol/PluggableMap.js
```diff
@@ -1168,6 +1153,27 @@ class PluggableMap extends BaseObject {
         this.renderer_ = this.createRenderer();
       }
 
+      this.mapBrowserEventHandler_ = new MapBrowserEventHandler(
+        this,
+        this.moveTolerance_
+      );
+      for (const key in MapBrowserEventType) {
+        this.mapBrowserEventHandler_.addEventListener(
```
EvertEt, line 1161: Will these get removed by `this.mapBrowserEventHandler_.dispose();`? Otherwise it might cause leaks?

src/ol/PluggableMap.js
```diff
@@ -1139,11 +1111,24 @@ class PluggableMap extends BaseObject {
...
+        EventType.CONTEXTMENU,
+        this.boundHandleBrowserEvent_
+      );
+      this.viewport_.removeEventListener(
+        EventType.WHEEL,
+        this.boundHandleBrowserEvent_
+      );
+      if (this.handleResize_ !== undefined) {
+        removeEventListener(EventType.RESIZE, this.handleResize_, false);
+        this.handleResize_ = undefined;
```
EvertEt, line 1129: Not sure if we can remove the same logic in the `(!targetElement)` branch below? I couldn't think of a case where it would still trigger. (Only 2 subsequent `setTarget(null)` but then there wouldn't be an `this.handleResize_`)

EvertEt:
Can confirm it fixes #11231 by changing the `simple` example to include the following
```js
setTimeout(() => map.setTarget(null), 5000);
setTimeout(() => map.setTarget('map'), 7000);
```

src/ol/PluggableMap.js
ahocevar: Good catch, thanks! That was a copy/paste mistake.

src/ol/PluggableMap.js
ahocevar: Yes, these will get removed.

EvertEt: Mind to explain me how? just interested :)

PR #11270, opened 2020-07-12
Fix handling of events originating from overlayContainerStopEvent children

### LINKED ISSUE #11251: Controls stealing touch gesture key-up events
**Describe the bug**
When pinch zooming gesture ends on top of control, the gesture control gets messed up. Somehow there seems to be now an extra finger in the position where the last finger was released on top of the control and attempted panning will produce zooming or rotation

**To Reproduce**
1. Go to 'https://openlayers.org/en/latest/examples/bing-maps.html' with a touch enabled device. (tested with a phone, tablet and Win 10 touch enabled desktop PC running Firefox)
2. Pinch zoom-in on the map so that one finger slides on top of control such as zoom buttons. Lift up this finger last.
3. After lifting both fingers, try to pan the map with one finger.

**Expected behavior**
After zooming, the map operation would be normal

The behaviour seems to be somehow related to timing and it seems to be easier to replace with slightly heavier tile maps such as satellite maps or high resolution terrain maps.
The feature is very annoying on small screens, such as mobile phones, when you have multiple controls such as zoom slider, attributions and scale line visible on the edges of the screen. In such situations this issue will hit you very quickly. I have done most of my testing with zoom slider, but managed to repeat the issue with zoom buttons and attributions area.

Setting touch-events to none on the control's css (such as attribution or scale line), the problem can be avoided but of course this does not work for interactions.

ahocevar:
Fixes #11251.

We should not pass pointerdown, wheel and keydown events to interactions. Similar to like we did before #8982, but without using `stopPropagation`.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -13,6 +13,7 @@ import MapEvent from './MapEvent.js';
 import MapEventType from './MapEventType.js';
 import MapProperty from './MapProperty.js';
 import ObjectEventType from './ObjectEventType.js';
+import PointerEventType from './pointer/EventType.js';
 import RenderEventType from './render/EventType.js';
 import TileQueue, {getTilePriority} from './TileQueue.js';
 import View from './View.js';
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for any review.

PR #11277, opened 2020-07-14
Have high resolution regular shape

sbrunner:
Be able to have the regular shape in a high resolution.

Visible on a smartphone on the example https://openlayers.org/en/latest/examples/regularshape.html


M393:
I like it.

Have you thought about applying the same to `ol/style/Icon`s with the color attribute set?
Would be pretty neat for svg icons.
https://openlayers.org/en/master/examples/icon-color.html

sbrunner:
As I see it's not directly possible for the SVG, for the SVG you should create a bigger icon, and use the parameter scale to get the wanted size...

ahocevar:
Nice work @sbrunner. I look forward to when this is ready for review.

> As I see it's not directly possible for the SVG, for the SVG you should create a bigger icon, and use the parameter scale to get the wanted size...

It should be doable to fix that with this pull request too, because regular shapes and icons go through the same code path in the end. All that needs to be done is modify the `replaceColor()` method in `ol/style/IconImage` to multiply the canvas width and height by the pixel ratio, and modify the drawImage call to provide both source and destination size.

sbrunner:
OK @ahocevar I will try :-)

sbrunner:
Good solution, now it's working like a charm :-)

src/ol/style/IconImage.js
```diff
@@ -188,7 +199,7 @@ class IconImage extends EventTarget {
     if (this.imageState_ == ImageState.IDLE) {
       this.imageState_ = ImageState.LOADING;
       try {
-        /** @type {HTMLImageElement} */ (this.image_).src = this.src_;
+        /** @type {HTMLImageElement} */ (this.image_).srcset = this.src_;
```
M393, line 202: The srcset attribute is not supported in Internet Explorer 11. Is this change intended if only one src is set?

examples/icon-color.js
```diff
@@ -19,13 +19,18 @@ const london = new Feature({
...
+const berlin = new Feature({
+  geometry: new Point(fromLonLat([13.3884, 52.5169])),
+});
 
 rome.setStyle(
   new Style({
     image: new Icon({
       color: '#8959A8',
       crossOrigin: 'anonymous',
-      imgSize: [20, 20],
```
M393, line 28: The imgSize is necessary for svg images in Internet Explorer 11

src/ol/style/IconImage.js
```diff
@@ -152,18 +159,22 @@ class IconImage extends EventTarget {
    * @return {HTMLImageElement|HTMLCanvasElement} Image element.
    */
   getHitDetectionImage(pixelRatio) {
-    if (!this.hitDetectionImage_) {
-      if (this.isTainted_()) {
+    if (!this.hitDetectionImage_[pixelRatio]) {
+      if (this.color_) {
```
M393, line 163: The isTainted check should not be removed. If the image is loaded from an external server or is and svg in internet explorer the canvas data will not be accessible.

src/ol/style/IconImage.js
sbrunner: It was a test that => will be removed...

examples/icon-color.js
sbrunner: +1, Thanks :-)

src/ol/style/IconImage.js
sbrunner: Then I also modify the replaceColor function to have all the canvas function in the isTainted if, Thanks :-)

sbrunner:
> In which case is it expected to update the pixel ratio? Changing the browser zoom level keeps the original icons, so the icons are blurred when zooming in. Is this expected?

To test it I use the Firefox debugger switch to Mobile device (Galaxy S9), click on take a screenshot, open the given image with Gimp and here we can relay see the details of the hidpi image.

When we change the browser zoom it's also working, but we should reload the page, It's probably also doable but probably deeper in OpenLayers :-)
In fact, it's set here, https://github.com/openlayers/openlayers/blob/main/src/ol/PluggableMap.js#L169-L172
And the DEVICE_PIXEL_RATIO is used in 10 different files...

ahocevar:
Changing the pixel ratio dynamically is beyond the scope of this pull request. It is doable (with a listener), but that's a separate issue - and it could be handled on the application level too.

src/ol/style/IconImage.js
```diff
@@ -75,27 +75,23 @@ class IconImage extends EventTarget {
...
-      }
+      const context = createCanvasContext2D(1, 1);
+      context.drawImage(this.image_, 0, 0);
       try {
         context.getImageData(0, 0, 1, 1);
-        this.tainted_ = false;
-      } catch (e) {
         this.tainted_ = true;
+      } catch (e) {
+        this.tainted_ = false;
```
M393, line 94: I don't think this change is intentional, is it?

src/ol/style/IconImage.js
```diff
@@ -152,18 +159,22 @@ class IconImage extends EventTarget {
...
+          width * pixelRatio,
+          height * pixelRatio
+        );
+        context.scale(pixelRatio, pixelRatio);
         context.fillRect(0, 0, width, height);
-        this.hitDetectionImage_ = context.canvas;
+        this.hitDetectionImage_[pixelRatio] = context.canvas;
       } else {
-        this.hitDetectionImage_ = this.image_;
+        this.hitDetectionImage_[pixelRatio] = this.image_;
```
M393, line 174: The case when it is tainted but no color is set needs to also fill the canvas, but it does not need to be scaled.

src/ol/style/IconImage.js
sbrunner: Good question, yes it is but for me, its relay a strange change, If I don't do that it's not working on my Firefox, did I change the logic somewhere?

For me, this function should return true if the Canvas is working and some other..., and with the old code it's the inverse. I'm probably wrong somewhere but where?

src/ol/style/IconImage.js
sbrunner: I don't understand why we need it... no problem to do this change :-)

src/ol/style/IconImage.js
ahocevar: @sbrunner The `catch` block is meant to handle the the case where the canvas cannot be accessed because it is tainted. Better debug the try-block to see if you're running into the catch lock for a different reason. But the content of the catch-block should not be changed.

src/ol/style/IconImage.js
sbrunner: I probably just understand, it's to be able to do the get the color on a pixel :-)

M393: Try loading an image from an external server, then the canvas is 'tainted' and this line will throw an error.
https://github.com/openlayers/openlayers/blob/2dd9342a982e14c068cdd8c6492d2c6f79e8f81f/src/ol/style/IconImage.js#L95
This is a browser security measure.

The fallback is to fill the entire icon with one color for hit detection.

sbrunner:
> Changing the pixel ratio dynamically when the browser zoom changes is beyond the scope of this pull request. It is doable (with a listener), but that's a separate issue - and I would suggest handling that on the application level.

+1 that's out of the scope of this pull request

But since DEVICE_PIXEL_RATIO is a constant I think that's not possible to do it at application level...

src/ol/style/IconImage.js
```diff
@@ -125,10 +121,11 @@ class IconImage extends EventTarget {
     if (this.size_) {
       this.image_.width = this.size_[0];
       this.image_.height = this.size_[1];
+    } else {
+      this.size_ = [this.image_.width, this.image_.height];
     }
-    this.size_ = [this.image_.width, this.image_.height];
     this.unlistenImage_();
-    this.replaceColor_();
+    this.replaceColor_(1);
```
M393, line 128: I think this call is not needed as the call is done with the correct pixel ratio when the image is rendered.

sbrunner: I agree, I will remove it :-)

src/ol/style/IconImage.js
sbrunner: Effectively, I get it back and it's working :-) Thanks :-)

src/ol/style/IconImage.js
sbrunner: +1, Tested and working :-)

M393:
This call should use the canvas width and height. Otherwise only the top-left corner is colored.
https://github.com/openlayers/openlayers/blob/2dd9342a982e14c068cdd8c6492d2c6f79e8f81f/src/ol/style/IconImage.js#L236-L241

And maybe the canvas size should be rounded up if it is not an integer. The circle in Paris looks like it is cut off at the left.

sbrunner:
@M393 just changed, thanks.

For the `ceil` I already tested that, and as I see:
* With Firefox I always have an issue...
* With Chrome I never have an issue...

Now I added it because it looks like stronger :-)

sbrunner:
I just see, I miss some places, now it's working well :-)

src/ol/style/IconImage.js
```diff
@@ -75,22 +75,18 @@ class IconImage extends EventTarget {
...
    * @return {boolean} The image canvas is tainted.
    */
-  isTainted_(context) {
+  isTainted_() {
     if (this.tainted_ === undefined && this.imageState_ === ImageState.LOADED) {
-      if (!context) {
-        context = createCanvasContext2D(1, 1);
-        context.drawImage(this.image_, 0, 0);
-      }
+      const context = createCanvasContext2D(1, 1);
```
ahocevar, line 88: Now that you do not pass an existing canvas any more, it would be better to have a dedicated canvas that you can reuse as long as it is not tainted. Especially in Safari, cleanup of unused canvases takes a while, so creating too many canvases should be avoided.

sbrunner: +1, just done :-)

src/ol/style/IconImage.js
```diff
@@ -85,12 +90,15 @@ class IconImage extends EventTarget {
    */
   isTainted_() {
     if (this.tainted_ === undefined && this.imageState_ === ImageState.LOADED) {
```
M393, line 92: Doesn't this check ensure that there is only one canvas created. Once the `tainted_` is set this will no longer be executed.

sbrunner: +1 my last commit can be removed, do you agree @ahocevar?

ahocevar: The way you made the last commit does not bring an improvement. I was suggesting to have a static context, not a member context. Just replace `this.taintedTestContext_` with `taintedTestContext`, and initialize that outside the class with
```js
const taintedTestContext = null;
```

ahocevar:
@sbrunner I'm adding this to the 6.4.0 milestone. Would be great if you could address my remaining comment.

ahocevar:
@sbrunner Any update on this?

sbrunner:
Thanks @ahocevar :-)

PR #11286, opened 2020-07-16
Allow line breaks before slashes in apidoc navigation

MoonE:
Firefox allows line breaks before slashes but chrome needs a little extra help.
This gets rid of the horizontal scrollbar in the apidoc navigation in chrome.

### DIFF
--- a/config/jsdoc/api/template/tmpl/navigation.tmpl
+++ b/config/jsdoc/api/template/tmpl/navigation.tmpl
@@ -59,7 +59,7 @@ function listContent(item, title, listItemPrinter) {
         <li class="item item-<?js= item.type ?>" data-longname="<?js= item.longname ?>" data-name="<?js= item.prettyname.toLowerCase() ?>">
             <span class="title toggle">
                 <span class="glyphicon <?js= getItemCssClass(item.type) ?>"></span>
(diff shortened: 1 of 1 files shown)


config/jsdoc/api/template/tmpl/navigation.tmpl
```diff
@@ -59,7 +59,7 @@ function listContent(item, title, listItemPrinter) {
         <li class="item item-<?js= item.type ?>" data-longname="<?js= item.longname ?>" data-name="<?js= item.prettyname.toLowerCase() ?>">
             <span class="title toggle">
                 <span class="glyphicon <?js= getItemCssClass(item.type) ?>"></span>
-                <span><?js= self.linkto(item.longname, item.prettyname.replace(/[.~]/g, '\u200b$&')) ?></span>
+                <span><?js= self.linkto(item.longname, item.prettyname.replace(/[.~/]/g, '\u200b$&')) ?></span>
```
ahocevar, line 62: Shouldn't this be

```suggestion
                <span><?js= self.linkto(item.longname, item.prettyname.replace(/[.~\/]/g, '\u200b$&')) ?></span>
```
?

MoonE: Both work, I think the interpreter is smart enough to know that the regex literal cannot end inside a character group.

PR #11300, opened 2020-07-20
Drag and Drop to update source and fire event only if features added

### LINKED ISSUE #11299: Drag and Drop Interaction fires addfeatures event when no features have been added
**Describe the bug**
**To Reproduce**
Drag a random file (e.g. a binary file such as .zip) onto https://openlayers.org/en/main/examples/drag-and-drop.html  Check the console.  The application has attempted to zoom to the source extent when there are no features, but has quite reasonably done so in response to an `addfeatures` event.

**Expected behavior**
The event should only be fired when features have been added.

mike-000:
Fixes #11299

Do not update the source or fire the addfeatures event if dropping out of the formatConstructors loop without features being parsed.

### DIFF
--- a/src/ol/interaction/DragAndDrop.js
+++ b/src/ol/interaction/DragAndDrop.js
@@ -149,7 +149,9 @@ class DragAndDrop extends Interaction {
       if (features && features.length > 0) {
         if (this.source_) {
           this.source_.clear();
-          this.source_.addFeatures(/** @type {Array<import("../Feature.js").default>} */ (features));
+          this.source_.addFeatures(
+            /** @type {Array<import("../Feature.js").default>} */ (features)
+          );
         }
         this.dispatchEvent(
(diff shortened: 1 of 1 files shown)


src/ol/interaction/DragAndDrop.js
```diff
@@ -215,11 +214,13 @@ class DragAndDrop extends Interaction {
    * @param {string} text Text.
    * @param {import("../format/Feature.js").ReadOptions} options Read options.
    * @private
-   * @return {Array<import("../Feature.js").FeatureLike>} Features.
+   * @return {Array<import("../Feature.js").default>} Features.
    */
   tryReadFeatures_(format, text, options) {
     try {
-      return format.readFeatures(text, options);
+      return
```
ahocevar, line 221: Beware! This means `undefined` will be returned.

PR #11304, opened 2020-07-21
Reuse empty canvas

### LINKED ISSUE #11303: Empty canvas elements added to map with empty source
**Describe the bug**
Each `ol/layer/Vector` with empty `ol/source/Vector` adds one empty (0 x 150 px) canvas if no other layer with the same `className` has features.

**To Reproduce**
1. Go to https://codesandbox.io/s/empty-canvas-for-layer-without-features-s204b
2. On load there are two empty layers added to the map with the same className which results in two canvas elements with width of 0px

**Expected behavior**
Only one (or none at all) canvas should be created if no features are to be drawn.

ahocevar:
Fixes #11303

### DIFF
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -95,7 +95,11 @@ class CanvasLayerRenderer extends LayerRenderer {
         context = canvas.getContext('2d');
       }
     }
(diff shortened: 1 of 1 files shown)


ahocevar:
The failing tests will be fixed when #11309 is merged.

PR #11306, opened 2020-07-22
Replace/remove usage of tileserver.maptiler.com

petrsloup:
Some examples are still using tiles from https://tileserver.maptiler.com/, which we plan to shut down relatively soon.

This PR replaces the tiles in those examples with tiles from MapTiler Cloud (using the same key as other examples).

### DIFF
--- a/examples/reprojection.js
+++ b/examples/reprojection.js
@@ -5,7 +5,7 @@ import View from '../src/ol/View.js';
 import WMTS, {optionsFromCapabilities} from '../src/ol/source/WMTS.js';
 import WMTSCapabilities from '../src/ol/format/WMTSCapabilities.js';
 import proj4 from 'proj4';
-import {OSM, TileImage, TileWMS, XYZ} from '../src/ol/source.js';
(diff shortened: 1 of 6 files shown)


examples/reprojection.js
```diff
@@ -146,18 +146,6 @@ fetch(urlB)
     layers['bng'].setSource(new WMTS(options));
   });
 
-layers['grandcanyon'] = new TileLayer({
```
ahocevar, line 149: This should also be removed from the dropdown in the html.

petrsloup: Thanks for the quick reaction!

I think I did remove it: https://github.com/openlayers/openlayers/pull/11306/files#diff-016b1c92fe0f04f440228e4c4d91b53cL24 -- this should be the only reference to that layer, is it not?

ahocevar: Oh sorry @petrsloup, I had overlooked that. Thanks, all good!

petrsloup:
Thanks!

PR #11313, opened 2020-07-23
Add direct link to other examples to tags

MoonE:
This PR adds a tag-count next to each tag on every example page and a bootstrap modal to go directly to the related tag page.

To add the tag count I had to split the processing of the pages into two steps to  first collect all tags and then write the pages with the numbers.


sbrunner:
Thanks @ahocevar :-)

PR #11315, opened 2020-07-25
Examples fix html

MoonE:
Some fixes mostly related to the example pages.


MoonE:
The alt atribute is mainly for screen readers. As there is already the OpenLayers name next to it  it does not make sense to have it twice. An empty tag will be ignored by screen readers so it is only read once. If the logo included the name it would make sense to add it to the alt tag.
Source: https://www.drupal.org/project/panopoly/issues/2905972

jipexu:
yep but some screen reader will probably read the name of the file screen reader argument but  for the bot , reference and SEO or even if the image is not displayed (for some reason) a non empty alt image attribut  take more sens ...

examples/index.html
```diff
@@ -5,7 +5,6 @@
     <meta http-equiv="X-UA-Compatible" content="chrome=1">
     <meta name="viewport" content="initial-scale=1.0, user-scalable=no, width=device-width">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" type="text/css">
-    <link rel="stylesheet" href="../css/ol.css" type="text/css">
```
mike-000, line 8: It was there because when run in the template the examples were not importing ol.css (which was causing inconsistencies in some examples with custom css when viewed on the example page and when run stand-alone or in codesandbox).  Has that changed?  If not it should be included, but for better consistency with the stand-alone behaviour it might be best placed below the `{{{ css.tag }}}`

MoonE: The index.html file is this: https://openlayers.org/en/main/examples/, as there is no map there should be no need for the ol.css.

ahocevar:
@MoonE I have a rebased version of your pull request locally, for reviewing. Do you want me to push the rebased version, or have you made other changes in the meantime?

MoonE:
Thanks @ahocevar, I also just rebased,  pushed it now.

MoonE:
@jipexu, When an icon conveys the same meaning as text next to it an empty alt attribute should be set.
You can find the (very detailed) explanation with examples here in the html standard specification:
https://html.spec.whatwg.org/multipage/images.html#a-short-phrase-or-label-with-an-alternative-graphical-representation:-icons,-logos

jipexu:
@MoonE  yes thanks i agree with this in this context for the navigation or command icons !
I just say in some case (for SEO) - if we want the image indexed - is better to have the alt attribut even if the next text is the same meaning.
For example the openlayers logo : all the first images of openlayers in the images search engines come from wikipedia (where the image have an alt attribut with "openlayers logo...") if we had an attribut in our menu for the logo it will be also indexed from openlayers.org ...

PR #11323, opened 2020-07-26
Release v6.4.0

ahocevar:


MoonE:
Looks like https://github.com/openlayers/openlayers/pull/11315 did not make it into the current  website build?

ahocevar:
@MoonE It should be included in the v6.4.0 build. What makes you think otherwise?

MoonE:
If I go to one of the example pages, the changes from that PR are not there.
https://openlayers.org/en/latest/examples/box-selection.html
The line numbers are missing, the ol.css is included, the source headings are the old ones, etc.

ahocevar:
It looks like you are right @MoonE. I thought I had merged #11315 before cutting the release, but it looks like I did not.

Anyway, we have a few bugfixes already that warrant a patch release, so I assume I will be releasing v6.4.1 tomorrow, which will update the website too.

PR #11339, opened 2020-07-27
Use bootstrap 4.5.0 in all examples

MoonE:
Some of the examples still used the old bootstrap / jquery versions. 
This caused an error with the call to destroy and the background went pitch black when clicking on the new tag count buttons.


ahocevar:
The template could now be modified to not load Bootstrap again, because it is already loaded for all examples.

PR #11340, opened 2020-07-27
Examples fix / improvement

MoonE:
- Use the same jquery version (3.5.1) in all examples.
- Don't add bootstrap js twice
- Remove unused extraResources in build process
- fix tooltip in kml-earthquake example

### DIFF
--- a/examples/webpack/example-builder.js
+++ b/examples/webpack/example-builder.js
@@ -9,9 +9,11 @@ const promisify = require('util').promisify;
 const RawSource = require('webpack-sources').RawSource;
 
 const readFile = promisify(fs.readFile);
-const isCssRegEx = /\.css$/;
+const isCssRegEx = /\.css(\?.*)?$/;
 const isJsRegEx = /\.js(\?.*)?$/;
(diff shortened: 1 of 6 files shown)


MoonE:
bootstrap js requires jquery and it is used to display the list of other examples for a tag.

examples/webpack/example-builder.js
```diff
@@ -12,6 +12,8 @@ const readFile = promisify(fs.readFile);
 const isCssRegEx = /\.css$/;
 const isJsRegEx = /\.js(\?.*)?$/;
 const importRegEx = /^import .* from '(.*)';$/;
+const isTemplateJs = /\/(jquery|(bootstrap(\.bundle)?))(\.min)?\.js$/;
```
ahocevar, line 15: To find the jquery scripts as we include them, the regex should look like this:
```suggestion
const isTemplateJs = /\/(jquery(-[0-9.]+)?|(bootstrap(\.bundle)?))(\.min)?\.js$/;
```

MoonE:
Please merge, thanks for asking.

jipexu:
thanks !

PR #11346, opened 2020-07-28
Fix text instruction flat coordinates when stride is not 2

### LINKED ISSUE #11344: Issues in KML example
ff980077eed459c6a2a9e3525e0a555f9d9ffeae broke the `kml.html` example. Looks like that change broke something for rendering of features that are not decluttered: some features disappear and reappear as the user zooms in or out.

### LINKED ISSUE #11347: features with style function disappear while panning after update to v6.4.0
After update to v.6.4.0, features with style function, disappear while panning the map,
and reappear after pan stop.
This includes pointer pan and View aniamte().
Other features remain on the map and move with it.
The feature disappear including its label.

ahocevar:
Fixes #11347 

Text instructions (i.e. labels) for features with 3 or more dimensions caused the whole feature to disappear when interacting with the map. This pull request fixes that.

### DIFF
--- a/src/ol/render/canvas/TextBuilder.js
+++ b/src/ol/render/canvas/TextBuilder.js
@@ -174,7 +174,7 @@ class CanvasTextBuilder extends CanvasBuilder {
     const geometryType = geometry.getType();
     let flatCoordinates = null;
     let end = 2;
-    let stride = 2;
+    let stride = geometry.getStride();
     let i, ii;
 
     if (textState.placement === TextPlacement.LINE) {
@@ -183,7 +183,6 @@ class CanvasTextBuilder extends CanvasBuilder {
       }
       let ends;
       flatCoordinates = geometry.getFlatCoordinates();
-      stride = geometry.getStride();
       if (geometryType == GeometryType.LINE_STRING) {
         ends = [flatCoordinates.length];
       } else if (geometryType == GeometryType.MULTI_LINE_STRING) {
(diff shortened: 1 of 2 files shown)


src/ol/render/canvas/TextBuilder.js
```diff
@@ -240,7 +239,6 @@ class CanvasTextBuilder extends CanvasBuilder {
         case GeometryType.POINT:
         case GeometryType.MULTI_POINT:
           flatCoordinates = /** @type {import("../../geom/MultiPoint.js").default} */ (geometry).getFlatCoordinates();
-          end = flatCoordinates.length;
```
tschaub, line 243: I hadn’t yet approved because I was curious to know why we still need to set `end` for MultiLineString and why we still set `stride` for Polygon.  But I haven’t yet made the time to understand it.

ahocevar: Thanks for the review @tschaub!

For polygons, the `getFlatMidpoint()` function always uses a stride of 3, with the available length at the midpoint at index 2. That is why we set the `stride` to 3.

For the `getFlatMidpoints` result in the MultiLineString, I'm going to add another test case. Looks like the fact that we still set `ends` there might be causing issues.

ahocevar: Thanks for raising concerns @tschaub. There were indeed more issues that I fixed now. I also added tests for all geometry types.

ahocevar: We need to set `stride = 2` for MultiLineString because `getFlatMidpoints()` returns 2-dimensional coordinates. And for MultiPolygon, because we create a 2-dimensional `flatCoordinates` array ourselves. For all other geometry types, we only receive a single point in `flatCoordinates`, of which only the first two values are used anyway because `end = 2` is set.

PR #11348, opened 2020-07-29
Add all layers to layer export

M393:
Three of the layer types are not listed in `ol/layer` module. I guess they just were forgotten?


tschaub:
Thanks, @M393.

PR #11354, opened 2020-07-30
Fix line breaks in changelog

ahocevar:


ahocevar:
Text-only change, merging.

PR #11361, opened 2020-07-30
Add nominalScale option to ScaleBar

### LINKED ISSUE #11349: ScaleBar on v6.4.0
**Describe the bug**
The numeric value of the ScaleBar may not be correct. 1 : xxxxxxxx

**To Reproduce**
The new ScaleBar allows the use of dynamic dpi values which should be great (didn't test it enough) and uses ol.proj.getPointResolution() to correct the ScaleBar when we are away from projection center (EPSG:3857), both in ScaleBar WIDTH and Text value.

Shouldn't it correct the WIDTH of the bar only?

In my testing i created 2 maps, 1 with zoom levels other with resolutions[] based on rounded scale values (1:10000, 1:5000, 1:2000, ...), in both maps when drag the map without zooming, the ScaleBar updates based on the distance from projection center, but my view resolution keeps the same (as intended i think).

If i am on a resolution corresponding to ex: 1:100000 my ScaleBar should give me 1:100000 (at projection center) and as i drag the map or move away from projection center the ScaleBar WIDTH should update but not the text (i think).

All this because i MUST use a Scale based Map so i can make sure what i see is on a fixed scale value in other to analyze and export data.

Am i too off on this issue?

Thanks

mike-000:
Fixes #11349 

Adds a nominalScale boolean option to the ScaleLine control to allow the scale text above the scalebar to be based on nominal projection units without adjustment for point resolution.

### DIFF
--- a/examples/scale-line.html
+++ b/examples/scale-line.html
@@ -30,3 +30,7 @@ tags: "scale-line, openstreetmap"
 <div id="showScaleTextDiv" style="display:none">
   <input type="checkbox" id="showScaleText" checked>Show scale text
 </div>
+
+<div id="nominalScaleDiv" style="display:none">
+  <input type="checkbox" id="nominalScale">Nominal scale
+</div>
--- a/examples/scale-line.js
+++ b/examples/scale-line.js
@@ -9,10 +9,13 @@ const typeSelect = document.getElementById('type');
 const stepsSelect = document.getElementById('steps');
 const scaleTextCheckbox = document.getElementById('showScaleText');
 const showScaleTextDiv = document.getElementById('showScaleTextDiv');
+const nominalScaleCheckbox = document.getElementById('nominalScale');
+const nominalScaleDiv = document.getElementById('nominalScaleDiv');
 
 let scaleType = 'scaleline';
 let scaleBarSteps = 4;
 let scaleBarText = true;
+let nominalScale = false;
 let control;
 
 function scaleControl() {
@@ -27,6 +30,7 @@ function scaleControl() {
     bar: true,
     steps: scaleBarSteps,
     text: scaleBarText,
+    nominalScale: nominalScale,
     minWidth: 140,
   });
   return control;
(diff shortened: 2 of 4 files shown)


ahocevar:
@mike-000 Can you please pay more attention to making sensible commits? Our [pull request guidelines[(https://github.com/openlayers/openlayers/blob/main/CONTRIBUTING.md#pull-request-guidelines) should give you an idea. Example: commits like "fix typos" should not be necessary when you have appropriate tooling locally, or run `npm run lint` before committing changes. You can also rewrite history in your feature branches, e.g. using `git commit --amend` or `git rebase --interactive`.

Also, if you are planning to add more commits to a pull request, set the pull request to "Draft". Otherwise it is hard to know when the pull request is ready for review.

mike-000:
I am using the online editor, I'll squash the commits when finished.

mike-000:
My own exceptions were simply to preserve the existing functionality where maps of the sane physical area show the same scale regardless of underlying projection and to provide an option to return to the previous behaviour where they did not:, i.e.

> the Scalebar text takes in consideration the current view resolution only

Before version 6 a constant true scale at the center of a web mercator projection could be achieved in the application by dynamically updating the entries in the view's resolutions array as the point resolution changed, e.g. to simulate British National Grid where reprojection was not possible due to a vector tile overlay layer.

image

image

image

In OL6 that interferes with the zoom interaction.  Setting resolution constraints might be incompatible with printing where resolutions required for print dpi would differ from those for screen dpi.

ahocevar:
https://github.com/openlayers/openlayers/issues/11349#issuecomment-670975981 explains how a map with a constant scale can be created. The working demo there already makes use of the fix from #11395.

mike-000:
> [#11349 (comment)](https://github.com/openlayers/openlayers/issues/11349#issuecomment-670975981) explains how a map with a constant scale can be created. The working demo there already makes use of the fix from #11395.

Mousewheel zoom in the codesandbox is still affected.  My code was similar except my view was constrained to a resolutions array, so I was updating each entry in the array used to construct it:

                if (Math.abs((viewScalingNew/viewScaling)-1) > 0.01) {
                    viewScaling = viewScalingNew;
                    var zoom = view.getZoom();
                    for (var i = 0; i < viewResolutions.length; i++) {
                        viewResolutions[i] = osgbResolutions[i] * viewScaling;
                    }
                    view.setZoom(zoom);
                }


That worked in OL5 but in OL6 even if updating only when the change was more than 1% mousewheel zoom is almost unusable.

PR #11363, opened 2020-07-30
Fix wrap x for custom projections without extent defined

Razi91:
Version 6.4.0 changed the way `wrapX` works, it stopped checking if given projection can be wrapped and extent exists. This PR restores it, returns unchanged coordinate if doesn't.

BTW. should this function mutate original array?


mike-000:
Duplicate of #11359
`canWrapX` includes a check for extent (it cannot wrap without it).

PR #11367, opened 2020-07-30
Use https for mapbox tiles

MoonE:
Browsers are going to block non-secure content on https pages at some point in the future.

Is there a reason why some examples use v3 api and some v4? 
And why does the v3 api not need access tokens?


ahocevar:
@MoonE 

> Is there a reason why some examples use v3 api and some v4?

Historic reasons, I would say. 

> And why does the v3 api not need access tokens?

I cannot answer this.

Issue #11282, opened 2020-07-15
Overlay example rendering issue


tschaub:
**Describe the bug**

The overlay example renders the word “Vienna” over the popup.

![6B57292A-ED1C-4B13-9763-E9403A5BBE1D](https://user-images.githubusercontent.com/41094/87562180-e9db6200-c67a-11ea-93af-3254ab084556.jpeg)

**To Reproduce**
1. Go to https://openlayers.org/en/v6.3.1/examples/overlay.html
2. Click on the map below the Vienna marker
3. See the overlay rendered under the Vienna label

**Expected behavior**
The popup should render over the Vienna label as it did in https://openlayers.org/en/v6.2.1/examples/overlay.html.

mike-000:
In 6.2.1 if you pan the map the overlay passes over the controls.  That can still be achieved by specifying `insertFirst: false`

ahocevar:
I do see @tschaub's point. The problem is that in v6.2.1 it was possible to use css `z-index` to control stacking of overlays, which is what the Bootstrap popover does because of the `placement: top` option. Now this is not possible any more.

keyro90:
> In 6.2.1 if you pan the map the overlay passes over the controls. That can still be achieved by specifying `insertFirst: false`

IMHO it's not a solution, because if you've more than one overlay, you can't decide priorities.

PR #11376, opened 2020-07-31
Overlay Eample - Show the popup in front of the other overlays

MoonE:
This fixes the rendering issue for the example mentioned here: #11282


tschaub:
Fixes #11282.

PR #11391, opened 2020-08-03
Simplify scale calculation used for WMS getLegendUrl()

mike-000:
Not only is converting from meters to inches and back to meters inefficient but as noted in https://github.com/openlayers/openlayers/issues/11349#issuecomment-668024841 using the international conversion forward and the old US conversion back is introducing a small error (if US units are being used for the projection the difference between US and imperial/international is already taken into account in the mpu).


tschaub:
Thanks, @mike-000.

PR #11395, opened 2020-08-04
Correct inchesPerMeter and add tests for ScaleBar text

mike-000:
This PR contains a fix and tests from #11361 which were not specific to the nominalScale proposal.  The issue with using an imprecise inches per meter value can also be seen in the Print to scale example if 1:500000 is selected (the text in the pdf is 1:499999)

If #11361 is not merged this replaces #11361

### DIFF
--- a/test/spec/ol/control/scaleline.test.js
+++ b/test/spec/ol/control/scaleline.test.js
@@ -617,9 +617,7 @@ describe('ol.control.ScaleLine', function () {
       expect(element).to.be.a(HTMLDivElement);
       const text = element.innerText;
       expect(text.slice(0, 4)).to.be('1 : ');
-      expect(text.replace(/^1|\D/g, '')).to.eql(
-        Math.round(map.getView().getResolution() / 0.00028)
-      );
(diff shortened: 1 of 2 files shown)


test/spec/ol/control/scaleline.test.js
```diff
@@ -596,4 +596,53 @@ describe('ol.control.ScaleLine', function () {
...
+        })
+      );
+      map.renderSync();
+      const element = document.querySelector('.ol-scale-text', map.getTarget());
+      expect(element).to.not.be(null);
+      expect(element).to.be.a(HTMLDivElement);
+      const text = element.innerText;
+      expect(text.slice(0, 4)).to.be('1 : ');
+      expect(text.replace(/^1|\D/g, '')).to.eql(
+        Math.round((map.getView().getResolution() * 0.5) / 0.00028)
```
ahocevar, line 644: For test readability, either expect the actual value instead of the calculation, or use the point resolution in the calculation. I'd prefer the former, also for the test above.

PR #11433, opened 2020-08-13
Fix typo in module name

### LINKED ISSUE #11432: update6.4.3 Problem
When I upgraded from 6.3.2 to 6.4.3 and rebuilt the index file of ol.js, the package /ol/render/canvas/hitdetet could not be found. I don't know what went wrong?
QQ图片20200812175122

fredj:
fixes #11432


fredj:
> It should be possible to add a rule to https://github.com/openlayers/eslint-plugin to catch this.

I've created an issue in eslint-plugin

PR #11455, opened 2020-08-23
Apidoc improvements

MoonE:
This PR add some improvements to the apidocs.

- The source code links to github are created in the template, not on page load and if the version ends in -dev it links to the main branch instead of generating an invalid link.
- It is recommended to have the charset meta tag within the first 1024 bytes of the document.
- pressing enter in the search field will navigate to the first result
- adjust the margin around the horizontal line between elements to have equal space below  and above
- remove some empty nameContainer tags, this also  gets rid of some horizontal lines for those empty tags
- add required `dl` tags around `dd` / `dt` tags.


ahocevar (approved):
Thanks, @MoonE!

PR #11467, opened 2020-08-24
Style renderer example

ahocevar:
This pull request adds an example that shows how to use a custom style renderer.


ahocevar:
See https://deploy-preview-11467--ol-site.netlify.app/examples/style-renderer.html for a preview of the new example.

Issue #11463, opened 2020-08-24
WebGL-renderer: implement preRender and postRender


MichaelLangbein:
### Description

Currently, Heatmaps don't fire  `prerender` or `postrender` events. This is because a.t.m. only `CanvasLayerRenderer` implements the `preRender` and `postRender`  methods. 

Here's a minimum working example: note how only the OSM's render-events are being logged:

```js
const layers = [
    new TileLayer({
        source: new OSM(),
    }),
    new Heatmap({
        source: new VectorSource({
            features: new GeoJSON().readFeatures({
                "type": "FeatureCollection",
                "features": [{
                    "type": "Feature",
                    "properties": {},
                    "geometry": {
                      "type": "Point",
                      "coordinates": [13.38134765625, 52.50953477032727]
                    }
                }]
              })
        })
    })
];

const view = new View({
    center: [15, 52],
    zoom: 7,
    projection: 'EPSG:4326'
});


const map = new Map({
    layers, view, target
});


for (const layer of layers) {
    layer.on('postrender', (event) => {
        console.log(event);
    });
}
```

### Solution

I believe we can literally just copy-paste the source code of `CanvasLayerRenderer`'s `preRender` and `postRender`  methods into `WebGLLayerRenderer`. We'd then only have to call `preRender` and `postRender` from `WebGLPointsLayerRenderer` inside of `renderFrame`. 

Would this be appropriate? It is my understanding that WebGl-support is still somewhat experimental (is that still true?). But if this is something we'd like to see, I'd like to try myself on a pull-request over the weekend. 

Keep up the great work and thanks a lot! :)

jahow:
That sounds absolutely appropriate. Please go ahead with a PR if you can, and don't hesitate to ask here for indications if needed. Thanks!

MichaelLangbein:
Thanks a lot! Working on this now. But I could indeed need a little pointer here.

This is the new `WebGLLayerRenderer.dispatchRenderEvent_` method I've created (You can find the full current state [here](https://github.com/openlayers/openlayers/compare/main...MichaelLangbein:webGlPostRender)): 

```js
  /**
   * @param {import("../../render/EventType.js").default} type Event type.
   * @param {WebGLRenderingContext} context Context.
   * @param {import("../../PluggableMap.js").FrameState} frameState Frame state.
   * @private
   */
  dispatchRenderEvent_(type, context, frameState) {
    const layer = this.getLayer();
    if (layer.hasListener(type)) {
      /**
[...]

PR #11471, opened 2020-08-25
Feature: added preRender and postRender methods to WebGLLayerRenderer…

MichaelLangbein:
…. Calling them from WebGLPointsLayerRenderer. This way users can get notified of `prerender` and `postrender` events

This PR handles [issue 11463](https://github.com/openlayers/openlayers/issues/11463).

### Description

Currently, Heatmaps don't fire  `prerender` or `postrender` events. This is because a.t.m. only `CanvasLayerRenderer` implements the `preRender` and `postRender`  methods. 

Here's a minimum working example: note how only the OSM's render-events are being logged:

```js
const layers = [
    new TileLayer({
        source: new OSM(),
    }),
    new Heatmap({
        source: new VectorSource({
            features: new GeoJSON().readFeatures({
                "type": "FeatureCollection",
                "features": [{
                    "type": "Feature",
                    "properties": {},
                    "geometry": {
                      "type": "Point",
                      "coordinates": [13.38134765625, 52.50953477032727]
                    }
                }]
              })
        })
    })
];

const view = new View({
    center: [15, 52],
    zoom: 7,
    projection: 'EPSG:4326'
});


const map = new Map({
    layers, view, target
});


for (const layer of layers) {
    layer.on('postrender', (event) => {
        console.log(event);
    });
}
```

### Solution

We basically just copy-pasted the source code of `CanvasLayerRenderer`'s `preRender` and `postRender`  methods into `WebGLLayerRenderer`. We then only had to call `preRender` and `postRender` from `WebGLPointsLayerRenderer` inside of `renderFrame`. 

A notable change: `RenderEvent`s usually provide the user with a `context` and an `inversePixelTransform`. After [a little discussion](https://github.com/openlayers/openlayers/issues/11463#issuecomment-679876602), it was decided that those two attributes are less useful in a WebGL context and can therefore be set to `null`.

### Ceterum censeo ...
As always, I am grateful for all constructive criticism and thanks to all for the great work :)

### DIFF
--- a/test/spec/ol/renderer/webgl/pointslayer.test.js
+++ b/test/spec/ol/renderer/webgl/pointslayer.test.js
@@ -4,6 +4,7 @@ import Point from '../../../../../src/ol/geom/Point.js';
 import VectorLayer from '../../../../../src/ol/layer/Vector.js';
 import VectorSource from '../../../../../src/ol/source/Vector.js';
 import ViewHint from '../../../../../src/ol/ViewHint.js';
+import WebGLPointsLayer from '../../../../../src/ol/layer/WebGLPoints.js';
 import WebGLPointsLayerRenderer from '../../../../../src/ol/renderer/webgl/PointsLayer.js';
 import {WebGLWorkerMessageType} from '../../../../../src/ol/renderer/webgl/Layer.js';
 import {
@@ -614,8 +615,13 @@ describe('ol.renderer.webgl.PointsLayer', function () {
         }),
       });
 
-      layer = new VectorLayer({
+      layer = new WebGLPointsLayer({
         source,
+        style: {
+          symbol: {
+            symbolType: 'square',
+          },
+        },
       });
 
       renderer = new WebGLPointsLayerRenderer(layer, {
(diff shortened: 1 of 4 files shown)


jahow:
Hey! thanks for taking the time to make this contribution. I think it's looking good, I'm just not sure about the unit test you wrote: you use a regular `VectorLayer` which itself uses a canvas renderer, so I'm guessing the test would pass even without your changes (since the canvas renderer already emits the events)... right?

I'll do some more test and make a complete review asap.

test/spec/ol/renderer/webgl/pointslayer.test.js
```diff
@@ -592,4 +593,81 @@ describe('ol.renderer.webgl.PointsLayer', function () {
...
+              'geometry': {
+                'type': 'Point',
+                'coordinates': [13, 52],
+              },
+            },
+          ],
+        }),
+      });
+
+      layer = new VectorLayer({
```
jahow, line 617: I would suggest using a webgl-based layer here, the result is exactly the same but it simply is a bit less confusing IMO.

```suggestion
      layer = new WebGLPointsLayer({
        source,
        style: {
          symbol: {
            symbolType: 'square',
          },
        },
      });
```

jahow (approved):
Thanks! This is looking good. I just made a suggestion to make the test a bit less confusing (use a webgl-based layer instead of a canvas-based one), I'll let you decide whether you integrate it.

test/spec/ol/renderer/webgl/pointslayer.test.js
MichaelLangbein: Of course, thank you!
Also thanks for checking the unit test - after reading your comment, I really wasn't sure anymore, either. 
Commit comming up in a second!

jahow:
All good, thanks again @MichaelLangbein

PR #11474, opened 2020-08-25
fix: modify rbush package import

### LINKED ISSUE #11473: Module import bug
**Describe the bug**
When using the OL package, in some circumstances (used as a module), importing the RBush package results in an error.

**Error messages in browser**

```
Uncaught SyntaxError: import not found: default Executor.js:5:7
```
and

```
Uncaught SyntaxError: import not found: default RBush.js:4:7
```

**Bad codes**

https://github.com/openlayers/openlayers/blob/82328fe2c1d336bae3744526e18ffa4726be9fbf/src/ol/render/canvas/Executor.js#L5

and

https://github.com/openlayers/openlayers/blob/139b048197f705ccd26919813a6728496423b4be/src/ol/structs/RBush.js#L4

**Correctly**

The following changes will correct the error.

In the Executor.js:

```js
import RBush from 'rbush';
```

and in the RBush.js:

```js
import RBush_ from 'rbush';
```

haasz:
Fixes #11473

### DIFF
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -2,7 +2,7 @@
  * @module ol/render/canvas/Executor
  */
 import CanvasInstruction from './Instruction.js';
-import RBush from 'rbush/rbush.js';
+import RBush from 'rbush';
 import {TEXT_ALIGN} from './TextBuilder.js';
 import {WORKER_OFFSCREEN_CANVAS} from '../../has.js';
 import {
--- a/src/ol/structs/RBush.js
+++ b/src/ol/structs/RBush.js
@@ -1,7 +1,7 @@
 /**
  * @module ol/structs/RBush
  */
-import RBush_ from 'rbush/rbush.js';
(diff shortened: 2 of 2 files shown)


ahocevar (changes requested):
Good catch, thanks @haasz!

ahocevar (approved):

walkermatt:
Hi, I've created https://github.com/mourner/rbush/pull/114 to transpile ES6 features except import/ export in the rbush module. Currently the package specifies `module: 'index.js'`, `index.js` includes ES6 syntax such as `class` which is what prompted #10385.

If https://github.com/mourner/rbush/pull/114 is merged then we shouldn't need any changes in OpenLayers but we will gain IE11 support.

ahocevar:
> If https://github.com/mourner/rbush/pull/114 is merged then we shouldn't need any changes in OpenLayers but we will gain IE11 support.

This is obsolete now that we don't support IE any more and ship untranspiled modules.

PR #11481, opened 2020-08-26
Adding original properties to cloned geometry

### LINKED ISSUE #11071: Cloned geometry loses additional properties.

michalzielanski:
This PR fix #11071 issue. If the original geometry has additional properties, they are copied to the cloned geometry.

### DIFF
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -110,6 +110,17 @@ class Geometry extends BaseObject {
   }
 
   /**
+   * Import properties from another geometry.
+   * @param {Geometry} geometry Source geometry.
+   * @protected
+   */
+  importPropertiesFrom(geometry) {
+    if (geometry.hasProperties()) {
+      this.setProperties(geometry.getProperties(), true);
+    }
+  }
+
+  /**
    * @abstract
    * @param {number} x X.
    * @param {number} y Y.
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -110,17 +110,6 @@ class Geometry extends BaseObject {
   }
 
   /**
-   * Import properties from another geometry.
-   * @param {Geometry} geometry Source geometry.
-   * @protected
-   */
-  importPropertiesFrom(geometry) {
-    if (geometry.hasProperties()) {
-      this.setProperties(geometry.getProperties(), true);
-    }
-  }
-
-  /**
    * @abstract
    * @param {number} x X.
    * @param {number} y Y.
--- a/src/ol/Object.js
+++ b/src/ol/Object.js
@@ -189,6 +189,18 @@ class BaseObject extends Observable {
   }
 
   /**
+   * Apply any properties from another object without triggering events.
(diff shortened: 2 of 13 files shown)


src/ol/geom/Geometry.js
```diff
@@ -109,6 +109,17 @@ class Geometry extends BaseObject {
...
+  /**
+   * Import properties from another geometry.
+   * @param {Geometry} geometry Source geometry.
+   * @protected
+   */
+  importPropertiesFrom(geometry) {
+    if (geometry.hasProperties()) {
+      this.setProperties(geometry.getProperties(), true);
+    }
+  }
```
tschaub, line 121: This method still allocates an unnecessary object and triggers an event for each property assigned.

We could instead have a protected `applyProperties` method on the `BaseObject` class.  This could look something like this:

```js
  /**
   * Apply any properties from another object without triggering events.
   * @param {BaseObject} source The source object.
   * @protected
   */
  applyProperties(source) {
    if (!source.values_) {
      return;
    }
    if (!this.values_) {
      this.values_ = {};
    }
    for (const key in source.values_) {
      this.values_[key] = source.values_[key];
    }
  }
```

Then this would be called in the `clone` method of each geometry.

michalzielanski: This method does not allocate the object when it is not needed. The `hasProperties()` returns `false` when `BaseObject` has no properties.

```js
import Point from 'ol/geom/Point';
let p = new Point([0,0]);
p.setProperties({foo: 1});
p.unset('foo')
console.log(p.values_, p.hasProperties()) // null, false
p = new Point([0,0]);
p.setProperties({});
console.log(p.values_, p.hasProperties()) // null, false
```

tschaub: Right.  But the extra object is still not needed when there are properties.

I realize this sounds pedantic.  Nobody is likely to be able to measure the difference.

However, an `applyProperties` method on the `BaseObject` would be useful to all other classes that inherit from it and want to use this same functionality in a `clone` method.  By contrast, this `importPropertiesFrom` method is implemented only for geometries, but it really has nothing to do with geometry.

As an added benefit, if we implement the `applyProperties` method on `BaseObject`, we can iterate directly over keys in the `values_` object – no need to copy all properties to an unnecessary intermediate object as is the case when calling `getProperties`.

michalzielanski: You're right. `getProperties` creates an intermediate object that is not used later.

tschaub:
Thanks, @michalzielanski.

Unrelated, the [security scan](https://github.com/openlayers/openlayers/pull/11481/checks?check_run_id=1035284013) is failing because this branch is not based on 849c8eafb46172486cbffbd79fe5634a572dfa4c.

PR #11486, opened 2020-08-28
guard for null

### LINKED ISSUE #11478: Overlay: null return value from PluggableMap not handled causes crash
Over in pluggable map, line 772 in getCoordinateFromPixelInternal there is a return null

https://github.com/openlayers/openlayers/blob/62842d550839d5fc1cdb5bfa2dfe5bca051408fe/src/ol/PluggableMap.js#L772

Which is not compensated for in Overlay.js

https://github.com/openlayers/openlayers/blob/62842d550839d5fc1cdb5bfa2dfe5bca051408fe/src/ol/Overlay.js#L460

In Overlay.js panIntoView(),  It is assumed that map.getCoordinateFromPixelInternal() will not return null, as in line 461 it immediately dereferences the return value with a [0] - this needs a guard for the murphy's law coincidence when this function gets called before the animation states are initialized.

Experimenting casually, merely refraining from any further action at this cycle/animation time finding a null value there seems to work around the issue, but I file this as bug report rather than pull request as I'm unsure if there are other side effects I might not be aware of...

davidiamaf:
against map.getPixelFromCoordinateInternal() returning null

Fixes #11478

### DIFF
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -458,6 +458,9 @@ class Overlay extends BaseObject {
           .getView()
           .getCenterInternal());
         const centerPx = map.getPixelFromCoordinateInternal(center);
+        if (!centerPx) {
+          return;
+        }
         const newCenterPx = [centerPx[0] + delta[0], centerPx[1] + delta[1]];
 
         const panOptions = panIntoViewOptions.animation || {};


ahocevar (approved):
Thanks, @davidiamaf

PR #11488, opened 2020-08-28
"Layer Z-Index" example bugfix

### LINKED ISSUE #11100: Bug in the example "Layer Z-Index"
There is an error in the [example](https://openlayers.org/en/v6.3.1/examples/layer-z-index.html).
Currently the code looks like this:
```js
var layer0 = createLayer([40, 40], styles['star'], 0);
```
and it should look like this (unmanaged layer should not have z-index):
```js
var layer0 = createLayer([40, 40], styles['star']);
```
to make the code match the description:
> The Z-index determines the rendering order; with {square: 1, triangle: 0, star: unmanaged} indices, the rendering order is triangle, square and star on top.

michalzielanski:
This PR fix #11100 issue. The code did not match the description. The star layer was assigned a z-index.

### DIFF
--- a/examples/layer-z-index.js
+++ b/examples/layer-z-index.js
@@ -56,7 +56,7 @@ function createLayer(coordinates, style, zIndex) {
   return vectorLayer;
 }
 
-const layer0 = createLayer([40, 40], styles['star'], 0);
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @michalzielanski

PR #11490, opened 2020-08-28
iconUrlFunction option for ol/format/KML

### LINKED ISSUE #4503: kml icon with url relative path
I have a kml file,a part of this:

```
<IconStyle><Icon><href>GroundStation.png</href></Icon></IconStyle>
```

this kml file is in my 'data' floader:

```
var vector = new ol.layer.Vector({
    url: 'data/facilities.kml',
    format: new ol.format.KML()
  })
});
```

and i put my GroundStation.png in my 'data' floader too.

But in this way ,my icon can't be requested rightly,because the url is 'GroundStation.png' but not expected 'data/GroundStation.png'.

### LINKED ISSUE #11415: iconUrlFunction option for ol/format/KML

Firstly this would provide a solution to #4503 but perhaps a more significant reason for having the option is it would allow applications to easily subclass the KML format to handle KMZ.

In the case of a KML containing relative paths if a iconUrlFunction was provided, for example

```
const format = new KML({
  iconUrlFunction: function (href) {
    let url = href;
    let path = window.location.href;
    path = path.slice(0, path.lastIndexOf('/'));
    if (href.indexOf(path) === 0) {
      url = href.replace(path, dataPath)
    }
    return url;
  }
});
```

Then at https://github.com/openlayers/openlayers/blob/main/src/ol/format/KML.js#L1285 using
` src: this.iconUrlFunction_(src),`
where a default function would return the href unaltered.

In the case of a custom format for KMZ (with the application using either jszip-sync or JSZip 2.x) the custom format would pass a function to use the relative file name from the kmz:

```
const zip = new JSZip();

function getKMLData (buffer) {
  let kmlData;
  zip.sync(function() {
    zip.loadAsync(buffer).then(function(zip) {});
    const kmlFile = zip.file(/.kml$/i)[0];
    if (kmlFile) {
      kmlFile.async('text').then(function(result) {kmlData = result}); 
    }
  });
  return kmlData;
}

function getKMLImage (href) {
  let url = href;
  let path = window.location.href;
  path = path.slice(0, path.lastIndexOf('/'));
  if (href.indexOf(path) === 0) {
    zip.sync(function() {
      const regexp = new RegExp(href.replace(path , '') + '$','i');
      const kmlFile = zip.file(regexp)[0];
      if (kmlFile) {
        const extension = href.slice(href.lastIndexOf('.'));
        kmlFile.async('base64').then(function(result) {
          url = 'data:image/' + (
            extension === '.png' ? 'png' : 'jpeg'
          ) + ';base64,' + result;
        }); 
      }
    });
  }
  return url;
}

class KMZ extends KML {
  constructor(opt_options) {
    const options = opt_options || {};
    options.iconUrlFunction = getKMLImage;
    super(options);
  }
  getType() {
    return 'arraybuffer';
  }
  readFeature(source, options) {
    const kmlData = getKMLData(source);
    return super.readFeature(kmlData, options);
  }
  readFeatures(source, options) {
    const kmlData = getKMLData(source);
    return super.readFeatures(kmlData, options);
  }
}
```
It is probably not worth making KMZ part of the library because it would add a dependency, but it could be added as a example using jsdelivr.

mike-000:
Fixes #11415
Fixes #4503

Adds a means to revise icon paths or to use icons from a KMZ (when using a synchronous method such as jszip-sync or JSZip 2.x to extract a data url from a preloaded buffer).

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -379,6 +379,24 @@ function createStyleDefaults() {
 let TEXTAREA;
 
 /**
+ * A function that takes a url `{string}` and returns a url `{string}`.
+ * Might be used to change an icon path or to substitute a
+ * data url obtained from a KMZ array buffer.
+ *
+ * @typedef {function(string):string} IconUrlFunction
+ * @api
+ */
+
+/**
+ * Function that returns a url unchanged.
+ * @param {string} href Input url.
+ * @return {string} Output url.
+ */
+function defaultIconUrlFunction(href) {
+  return href;
+}
+
+/**
  * @typedef {Object} Options
  * @property {boolean} [extractStyles=true] Extract styles from the KML.
  * @property {boolean} [showPointNames=true] Show names as labels for placemarks which contain points.
@@ -387,6 +405,8 @@ let TEXTAREA;
  * @property {boolean} [writeStyles=true] Write styles into KML.
  * @property {null|string} [crossOrigin='anonymous'] The `crossOrigin` attribute for loaded images. Note that you must provide a
  * `crossOrigin` value if you want to access pixel data with the Canvas renderer.
+ * @property {IconUrlFunction} [iconUrlFunction] Function that takes a url string and returns a url string.
+ * Might be used to change an icon path or to substitute a data url obtained from a KMZ array buffer.
  */
 
 /**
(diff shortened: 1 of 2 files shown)


mike-000:
Ready for review.

ahocevar (approved):
Thanks, @mike-000

PR #11500, opened 2020-09-01
New format method for GetProjectSettings service QGIS

ilpise:
Added a parameter in WMSCapabilities.js to get layers visibility status from a GetProjetcSettings request on a QGIS server.


ahocevar:
@ilpise I think this is decent work, but being specific to QGIS and not the part of the WMS specification, it would probably be a better fit for a separate package. We'd be happy to list that package on our [3rd party libraries](https://openlayers.org/3rd-party/) page.

PR #11501, opened 2020-09-01
Empty this.canvas_ for all pixel ratios when render() is called

### LINKED ISSUE #11497: Custom GPS accuraccy (circle style) broken från 6.3.1 -> 6.4.1
**Describe the bug**
I have a custom gps accuracy feature and style. Up until 6.3.1 it have worked great but in 6.4.1 the gps accuracy circle jumps (i.e moves to a different location in the map) when zooming. I have not pinpointed to exactly where the problem is.

**To Reproduce**
Working in 6.3.1: https://jsfiddle.net/se2k3t6y/
Broken in 6.4.1: https://jsfiddle.net/mscb4qgd/

An image of how it looks like in 6.4.1 when zooming in:
bild

**Expected behavior**
As in 6.3.1

How it should look like:
bild

mike-000:
Fixes #11497

Empty this.canvas_ for all pixel ratios when render() is called
Use pixelRatio 2 in rendering test

### DIFF
--- a/rendering/cases/circle-style/main.js
+++ b/rendering/cases/circle-style/main.js
@@ -52,7 +52,7 @@ const style = new Style({
 });
 
 new Map({
-  pixelRatio: 1,
+  pixelRatio: 2,
   layers: [
     new VectorLayer({
       source: vectorSource,
--- a/src/ol/style/RegularShape.js
+++ b/src/ol/style/RegularShape.js
@@ -404,6 +404,7 @@ class RegularShape extends ImageStyle {
 
     this.draw_(renderOptions, context, 0, 0, 1);
 
+    this.canvas_ = {};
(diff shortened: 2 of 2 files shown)


sbrunner (approved):
Looks good to me :-)

ahocevar (commented):
Thanks, @mike-000! Do you think you could add a new test that fails before and passes after your code change? The test you modified in this pull request also passes before the code change.

mike-000:
With pixelRatio set to 2 the old code produced three circles with radius 10
image
How does that pass?

ahocevar:
My bad, @mike-000 - I had only run the unit tests, not the complete test suite.

ahocevar (approved):

Issue #11508, opened 2020-09-03
urn:x-ogc:def:crs:EPSG:6.6:4326 is not reconized as WGS84 proj identifier


fgravin:
**Describe the bug**
Seems that the code `urn:x-ogc:def:crs:EPSG:6.6:4326` could be a valid `srsName` to identify WGS84 projection (eg: `EPSG:4326`). It is not set as an equivalent identifier for WGS84 projection in actual Openlayers code.

**To Reproduce**
```js
//proj.js
get('urn:x-ogc:def:crs:EPSG:6.6:4326')
null
```

**Expected behavior**
`get('urn:x-ogc:def:crs:EPSG:6.6:4326')` should return the 4326 proj object.

PR #11509, opened 2020-09-03
Add urn:x-ogc:def:crs:EPSG:6.6:4326 as WGS84 proj identifier

fgravin:
Fixes https://github.com/openlayers/openlayers/issues/11508

Add `urn:x-ogc:def:crs:EPSG:6.6:4326` as an equivalent identifier for WGS84 projection.

### DIFF
--- a/src/ol/proj/epsg3857.js
+++ b/src/ol/proj/epsg3857.js
@@ -64,8 +64,6 @@ export const PROJECTIONS = [
   new EPSG3857Projection('EPSG:102100'),
   new EPSG3857Projection('EPSG:102113'),
   new EPSG3857Projection('EPSG:900913'),
-  new EPSG3857Projection('urn:ogc:def:crs:EPSG:6.18:3:3857'),
-  new EPSG3857Projection('urn:ogc:def:crs:EPSG::3857'),
   new EPSG3857Projection('http://www.opengis.net/gml/srs/epsg.xml#3857'),
 ];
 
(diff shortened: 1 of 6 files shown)


ahocevar (commented):
I'm wondering if it wouldn't be better to use regular espression replacements in `ol/proj.get()`:
```js
return typeof projectionLike === 'string'
    ? getProj(/** @type {string} */ (projectionLike.replace(/urn:(x-)?ogc:def:crs:(\w+):(.*:)?(\w+)$/, '$2:$4')))
    : /** @type {Projection} */ (projectionLike) || null;
```
We do something like this in `ol/source/WMTS` already, but the regular expression there does not cover strings with `x-ogc` in the identifier:
https://github.com/openlayers/openlayers/blob/3a01e083c65f201ffb89ea814da8136ac8ef5c71/src/ol/source/WMTS.js#L374

Such replacements can be removed if we move the logic to `ol/proj.get`. And projection aliases could also be removed, saving a few bytes of build size.

fgravin:
@ahocevar i've implemented the solution you suggest, i have a concern about those 2 projections
- `urn:ogc:def:crs:OGC:1.3:CRS84`
- `urn:ogc:def:crs:OGC:2:84`

is it accetable that the projection code would then be ?
- `OGC:CRS84`
- `OGC:S84`

Otherwise, it might be preferable to put the regexp logic in `projections.js#get` method instead, what do you think ?

ahocevar:
@fgravin In this case I'd say let's use the regex just for EPSG codes. This would mean to change it to
```js
projectionLike.replace(/urn:(x-)?ogc:def:crs:EPSG:(.*:)?(\w+)$/, 'EPSG:$3')

fgravin:
There is still an issue in the  `proj.register` function.
If I do manually register `urn:ogc:def:crs:EPSG::26713`, then there is an error cause it tries to get `EPSG:26713` while this code does not exist.

If we put the regexp test in `projections.js#get` then it could check for both codes

ahocevar:
@fgravin Moving the regex replacement to `projections.js#get` sounds good. Thanks for your continued effort on this.

fgravin:
@ahocevar Thanks for following my work up.
I think the PR is ready for a review.

ahocevar (approved):
Thanks @fgravin!

Issue #11512, opened 2020-09-03
GML Linearring should not have srsName attribute


fgravin:
**Describe the bug**
The OpenLayers GML formats have an issue serializing a geometry to GML: `<gml:LinearRing>` element should not have `srsName` attribute.

Reference: http://www.datypic.com/sc/niem21/e-gml32_LinearRing.html

**To Reproduce**
1. Create an ol geometry
2. Write features from GML format of this geometry
3. Output GML (both version 2 and 3) will have an `srsName`

```xml
<gml:exterior>
  <gml:LinearRing srsName="urn:x-ogc:def:crs:EPSG:4326">
    <gml:posList srsDimension="2">38.5571 -75.7071 38.4631 -75.6991 38.4551 -75.3501
```

**Expected behavior**
`<gml:LinearRing>` does not have `srsName` attribute.
```xml
<gml:exterior>
  <gml:LinearRing>
    <gml:posList srsDimension="2">38.5571 -75.7071 38.4631 -75.6991 38.4551 -75.3501
```

mprins:
well that depends on the version of GML
for GML2 a geometry is defined as:
```xml
<complexType name="AbstractGeometryType" abstract="true">
<annotation>
<documentation>
All geometry elements are derived from this abstract supertype; a geometry element may have an identifying attribute (gid). It may be associated with a spatial reference system.
</documentation>
</annotation>
<complexContent>
<restriction base="anyType">
<attribute name="gid" type="ID" use="optional"/>
<attribute name="srsName" type="anyURI" use="optional"/>
</restriction>
</complexContent>
</complexType>
```
see http://schemas.opengis.net/gml/2.1.2/geometry.xsd

while for gml 3.2.1 a geometry is defined as:
```xml
<complexType name="AbstractGeometryType" abstract="true">
<annotation>
<documentation>
All geometry elements are derived directly or indirectly from this abstract supertype. A geometry element may have an identifying attribute (gml:id), one or more names (elements identifier and name) and a description (elements description and descriptionReference) . It may be associated with a spatial reference system (attribute group gml:SRSReferenceGroup). The following rules shall be adhered to: - Every geometry type shall derive from this abstract type. - Every geometry element (i.e. an element of a geometry type) shall be directly or indirectly in the substitution group of AbstractGeometry.
</documentation>
</annotation>
<complexContent>
<extension base="gml:AbstractGMLType">
<attributeGroup ref="gml:SRSReferenceGroup"/>
</extension>
[...]

PR #11513, opened 2020-09-03
GML: remove srsName from linearring

fgravin:
Fixes https://github.com/openlayers/openlayers/issues/11512

Remove `srsName` attribute in GML `linearring` serializers.

Reference: http://www.datypic.com/sc/niem21/e-gml32_LinearRing.html

### DIFF
--- a/src/ol/format/GML2.js
+++ b/src/ol/format/GML2.js
@@ -560,10 +560,6 @@ class GML2 extends GMLBase {
    */
   writeLinearRing(node, geometry, objectStack) {
     const context = objectStack[objectStack.length - 1];
-    const srsName = context['srsName'];
-    if (srsName) {
-      node.setAttribute('srsName', srsName);
-    }
     const coordinates = this.createCoordinatesNode_(node.namespaceURI);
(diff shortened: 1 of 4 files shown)


mprins:
this is just wrong, linearring being derived from abstractgeometry it can have srsname, 
see https://github.com/openlayers/openlayers/issues/11512#issuecomment-686620502

ahocevar (commented):
@fgravin I'd be curious to hear what this pull request fixes. `srsName` is an optional attribute (see schema links posted by @mprins in #11512) and should not do any harm if present.

tschaub:
As mentioned on the other ticket, a linear ring is not a geometry.  A polygon is (and can contain a linear ring in its exterior or interior elements).

ahocevar (changes requested):
In order to accept this pull request, it would need to be changed so when serializing a `LinearRing` that stands on its own (i.e. from an `ol/geom/LinearRing`), it should still have a `srsName` attribute. For the rings of a polygon, the `srsName` attribute can be removed.

PR #11517, opened 2020-09-04
Make clockwise check work when coordinates are duplicated

### LINKED ISSUE #11503: Negative value of the calculated area
**Describe the bug**
Calculating the area of the polygon at the time of duplicate vertices produces a negative but correct value.
It does not always occur, it can be observed while drawing when the point we are drawing is aligned with the previous one.

It has been available since version 6.4.x.
Verified on EPSG:2180 and restored on 3857

Initially I thought it was supposed to indicate a geometry error, but it doesn't always occur

**To Reproduce**
Example:
[link](https://codesandbox.io/s/red-wood-dpfw4)

ahocevar:
Fixes #11503 by improving the clockwise ring check algorithm to support duplicated coordinated not only for first and last vertex of a ring, but anywhere. Additional tests are added to avoid future regressions.

### DIFF
--- a/src/ol/geom/flat/orient.js
+++ b/src/ol/geom/flat/orient.js
@@ -77,7 +77,7 @@ function findUniqueVertex(
     i += direction * stride;
     if (i >= end) {
       i = offset;
-    } else if (i <= offset) {
+    } else if (i < offset) {
       i = end - stride;
     }
     x = flatCoordinates[i];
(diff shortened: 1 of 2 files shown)


src/ol/geom/flat/orient.js
```diff
@@ -16,44 +16,85 @@ import {coordinates as reverseCoordinates} from './reverse.js';
...
+) {
+  let previousX, previousY, x, y;
+  let i = start;
+  while (x === previousX && y === previousY) {
+    previousX = flatCoordinates[i];
+    previousY = flatCoordinates[i + 1];
+    i += direction * stride;
+    if (i >= end) {
+      i = offset;
+    } else if (i <= offset) {
```
tschaub, line 80: I think this should be `i < offset`, right?

ahocevar: Good find, you are absolutely right. Thanks!

ahocevar:
```suggestion
    } else if (i < offset) {
```

tschaub (approved):

PR #11519, opened 2020-09-06
Do not draw multipoints outside render extent

### LINKED ISSUE #11518: Too many instructions generated for MultiPoint
**Describe the bug**
When a MultiPoint is rendered and only some of the points are outside, render instructions for some of those are still generated.
This is because the appendFlatCoordinates algorithm is also used for lines and in this case needs to add one additional point outside to draw the line from / to there.
This is not needed for `MultiPoint` and polygon labels.

**To Reproduce**
1. Go to https://codesandbox.io/s/falling-snow-ukbrh?file=/src/index.js
2. Position the map so most of the Points are outside the map, stop interacting
3. Drag the points back in view, keep the mouse button pressed
4. There are points draw that were way outside the view

**Expected behavior**
There should only be points drawn that were at most 100px outside the map, as the render instructions are not rebuild until the interaction ends.

MoonE:
Previously the algorithm used is also used for line strings and
therefor the line has to be drawn to the next point outside the
rendered extent, this is not needed for `Multipoint` and text labels of `MultiPoint` and `MultiPolygon`.

It seems to work now as far as i can tell.
This changes the `TextBuilder` and `ImageBuilder` behavior, the others should remain unchanged.

Fixes #11518

### DIFF
--- a/src/ol/render/canvas/Builder.js
+++ b/src/ol/render/canvas/Builder.js
@@ -186,7 +186,8 @@ class CanvasBuilder extends VectorContext {
     closed,
     skipFirst
   ) {
-    let myEnd = this.coordinates.length;
+    const coordinates = this.coordinates;
+    let myEnd = coordinates.length;
     const extent = this.getBufferedMaxExtent();
     if (skipFirst) {
       offset += stride;
@@ -203,15 +204,15 @@ class CanvasBuilder extends VectorContext {
       nextRel = coordinateRelationship(extent, nextCoord);
       if (nextRel !== lastRel) {
         if (skipped) {
-          this.coordinates[myEnd++] = lastXCoord;
-          this.coordinates[myEnd++] = lastYCoord;
+          coordinates[myEnd++] = lastXCoord;
+          coordinates[myEnd++] = lastYCoord;
+          skipped = false;
         }
-        this.coordinates[myEnd++] = nextCoord[0];
-        this.coordinates[myEnd++] = nextCoord[1];
-        skipped = false;
+        coordinates[myEnd++] = nextCoord[0];
+        coordinates[myEnd++] = nextCoord[1];
       } else if (nextRel === Relationship.INTERSECTING) {
-        this.coordinates[myEnd++] = nextCoord[0];
-        this.coordinates[myEnd++] = nextCoord[1];
+        coordinates[myEnd++] = nextCoord[0];
+        coordinates[myEnd++] = nextCoord[1];
         skipped = false;
       } else {
         skipped = true;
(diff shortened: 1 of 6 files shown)


MoonE:
This may now generate a builder without coordinates. Doesn't seem to cause any problems, though it may be possible to discard the builder/generate no instructions entirely?

ahocevar:
I'm wondering if it wouldn't be better to not run multipoints through the topology simplifier at all, instead of making changes in there. For multipoints, the only check to make is really whether it intersects the desired extent or not.

MoonE:
Thanks for the suggestion, @ahocevar, It does make it a bit simpler.

Do you know what the problem is with the failed rendering test?
If I run it locally I get a bunch of failed tests, but not that one. 

The failed test for me are exactly the same with the main branch and this branch:
```
case ./cases/image-stretched-disable-smoothing/main.js': mismatch 0.013702392578125
case ./cases/layer-mapbox-vector/main.js (Mapbox vector layer renders)': mismatch 0.0278167724609375
case ./cases/layer-vector-decluttering/main.js': mismatch 0.00860595703125
case ./cases/layer-vector-multigeometry-decluttering/main.js': mismatch 0.021820068359375
case ./cases/layer-vectorimage-decluttering/main.js': mismatch 0.0097808837890625
case ./cases/reproj-image-stretched-disable-smoothing/main.js': mismatch 0.0102996826171875
case ./cases/reproj-tile-none-square/main.js': mismatch 0.0062408447265625
case ./cases/text-style/main.js': mismatch 0.026397705078125
```

MoonE:
I'm getting an error when a text style with placement 'line' is used with a Point/MultiPoint.  (Also happening in main branch)
`ends` is undefined in that case:
https://github.com/openlayers/openlayers/blob/6b0471c2dc32453122c4acff94e6529407a55f4e/src/ol/render/canvas/TextBuilder.js#L184-L205
Maybe it should default to point placement in that case?

src/ol/render/canvas/Builder.js
```diff
@@ -306,6 +328,19 @@ class CanvasBuilder extends VectorContext {
         renderer,
         inflateCoordinates,
       ]);
+    } else if (type == GeometryType.MULTI_POINT) {
+      flatCoordinates = geometry.getFlatCoordinates();
+      builderEnd = this.appendFlatPointCoordinates(flatCoordinates, stride);
+      if (builderEnd > builderBegin) {
```
MoonE, line 334: This skips the instruction if all points are filtered, though the begin and end geometry instructions are still added.

ahocevar:
First of all, don't worry about rendering tests failing locally. That's because of different renderer engines on different platforms.

You are right about label placement for points and multi points - 'line'  placement should be ignored there, defaulting to 'point'.

ahocevar (approved):
Thanks, @MoonE.

PR #11523, opened 2020-09-06
Fix decluttering of MultiPoint

### LINKED ISSUE #11507: Missing points when MultiPoint is used with declutter on layer
**Describe the bug**
When a `MultiPoint` is decluttered the decluttering is not stable, with multiple `Point`s it is.
Some points are not drawn even though there is no other point close to it.
It only happens if some of the points are outside the current view.

**To Reproduce**
1. Go to https://codesandbox.io/s/falling-snow-ukbrh?file=/src/index.js
2. Pan the map
3. Some `Point`s appear / disappear. When clicking the 'Render' button the decluttering changes after panning.

**Expected behavior**
The decluttering should be the same as with multiple `Point`s, change the features in line 58 to compare.

MoonE:
Fixes #11507

Labels are not rendered when zoomed out (nearly) all the way for https://deploy-preview-11523--ol-site.netlify.app/examples/vector-label-decluttering. Probably caused by wrapping x but I think that's not related to this issue?

Is there a case when more than two items are in a declutter group, I only know that icons and texts are grouped together when in the same style object.
The way it was handled by increasing the index by a fraction of the length did only work correctly for up to 5 labels, after that you get rounding issues.
```js
new Array(10).fill().map((_, x) => {
  let sum = 0;
  for (let i = x; i > 0; --i) {
    sum += 1 / x;
  }
  return sum;
});
```

Please have a look and tell me if there is something I overlooked / broke with this change.


MoonE:
Below is the code I used for testing, put it in examples/icon.js.
It triggers the original bug, and it should now render correctly with the changes in this PR.
```js
import Feature from '../src/ol/Feature.js';
import MultiPoint from '../src/ol/geom/MultiPoint.js';
// import Point from '../src/ol/geom/Point.js';
import VectorSource from '../src/ol/source/Vector.js';
import {Icon, Stroke, Style, Text} from '../src/ol/style.js';
import {Vector as VectorLayer} from '../src/ol/layer.js';
// import { Circle, Polygon } from '../src/ol/geom.js';

// The test feature positions:
// 2
//            3
//
//     4 5
//
// 0           1
//
const points = [
  [0, 0], // 0
  [0, 1], // 2
  [0.5, 0.5], // 5
  [0.9, 0.85], // 3
  [1, 0], // 1
  [0.3, 0.5], // 4
];

const vectorLayer = new VectorLayer({
  declutter: true,
  renderBuffer: 0,
  source: new VectorSource({
    features: [new Feature(new MultiPoint(points))],
    // features: points.map((p) => new Feature(new Point(p))).concat(new Feature(new Polygon([[ ...points, points[0].slice() ]]))),
    // features: [new Feature(new Polygon([[...points, points[0].slice()]]))],
  }),
  style: new Style({
    image: new Icon({
      anchor: [0.5, 46],
      anchorXUnits: 'fraction',
      anchorYUnits: 'pixels',
      src: 'data/icon.png',
      // scale: [1, -1],
    }),
    text: new Text({
      text: '.',
      // scale: [-1, -1],
      // padding: [10, 10, 10, 10],
      font: 'italic 20px Arial',
      stroke: new Stroke({
        color: 'red',
        width: 20,
      }),
    }),
  }),
});

const map = new Map({
  layers: [vectorLayer],
  target: document.getElementById('map'),
  view: new View({
    center: [-1.5, 0],
    resolution: 0.008828409901106019,
  }),
});
window.m = map;
```

ahocevar:
@MoonE This is a tough one, thanks for taking it on!

I like the changes you made, they make the code easier to read. Unfortunately I also don't understand why the changes here cause labels to disappear as soon as more than one world is displayed.

ahocevar:
@MoonE I tried to find the cause for the disappearing labels when multiple worlds are in the viewport, but I could not figure it out. I *think* the cause was different before this pull request, but I couldn't figure it out either.

That said, if you have any clue or idea, and are able to fix it as part of this pull request, that would be awesome!

If it helps: I do not think there are other reasons for having more than one item in a declutter group than combining point symbol and text.

MoonE:
Thanks for looking into it, @ahocevar.

It seems like using one Style object with text, fill and stroke has a slightly different (better?) behavior than the currently configured style where it is an array with fill, stroke separate from the text style.

I'll take a deeper look at the weekend.

ahocevar:
Thanks, @MoonE!

> With a canvas width of 478 I have x-coordinates for the labels between ~1908 and ~2386. So all coordinates are outside the canvas and no labels are drawn.

Those features are probably the ones from the world right of the real world, which gets drawn as soon as a part of that world is visible on the right edge of the map.

ahocevar:
Ok, after more investigation I come to the conclusion that the disappearing labels issue with this pull request is indeed different from the one before. And it is caused by executor groups being executed more than once when multiple worlds are visible. In this case the declutter groups get messed up. This might have to do with the removal of this logic:
```js
            if (
               rendered &&
               declutterGroup &&
               declutterGroups[declutterGroups.length - 1] !== declutterGroup
             ) {
               declutterGroups.push(declutterGroup);
             }
```

ahocevar:
I was able now to trigger the exact same behaviour we have with disappearing labels here with master. So I'm back at thinking that problem is not related to this pull request. But still, help with this problem would be much appreciated @MoonE.

MoonE:
When multiple worlds need to be rendered all labels before the last world are discarded here ...
https://github.com/openlayers/openlayers/blob/beefeaad1a4ba3233771e6ed36d17642a93dc253/src/ol/render/canvas/Executor.js#L653-L662
... because instructions are not added when the executor stays the same here:
https://github.com/openlayers/openlayers/blob/beefeaad1a4ba3233771e6ed36d17642a93dc253/src/ol/render/canvas/ExecutorGroup.js#L478-L488

I gave it a shot with https://github.com/MoonE/openlayers/commit/819f8902e073b5935537adb5591a7d1be7ca2e1a but that doesn't seem quite right 
As it is now the same label from multiple worlds is added to the same declutterGroup. I believe this to be the cause of label flickering in the declutter vector layer example.

ahocevar:
Thanks for your continued effort on this, @MoonE. The fix definitely goes in the right direction. And like you say, the remaining problem is that a separate declutter group should be created for each replayed world.

ahocevar:
@MoonE Thanks for your excellent findings! I took a slightly different approach attempting to fix the disappearing labels. Do you want to give abce10388e349af580279647e69be86247242812 a try?

ahocevar:
@MoonE If you want to cherry-pick a17eb43 or add a similar fix for the disappearing labels on multi-world views, then I think we can merge this. Or do you see any issues with that fix?

MoonE:
There is one problem remaining. The Fiji label at the date line overlaps with a different Fiji label from the next world.
Fiji is a three MultiPolygon with three Polygons, the first two are in the east next to the date line, the third is in the west next to the date line.
In the first execute call the first two are added, in the execute call for the next world only the last gets added, but it is added to the first group which contains the first polygon.

The overlapping labels get drawn because they are loaded together here https://github.com/openlayers/openlayers/blob/2de80e04e4c16799c43cb2b9b2e88b62db396ab1/src/ol/render/canvas/Executor.js#L553-L576
The different labels get grouped together because they are the first group of each world that gets rendered

Fiji:
```json
[
  [
    [
      [178.3736,-17.33992],
      [178.71806,-17.62846],
      [178.55271,-18.15059],
      [177.93266,-18.28799],
      [177.38146,-18.16432],
      [177.28504,-17.72465],
      [177.67087,-17.38114],
      [178.12557,-17.50481],
      [178.3736,-17.33992]
    ]
  ],
  [
    [
      [179.364143,-16.801354],
      [178.725059,-17.012042],
      [178.596839,-16.63915],
      [179.096609,-16.433984],
      [179.413509,-16.379054],
      [180,-16.067133],
      [180,-16.555217],
      [179.364143,-16.801354]
    ]
  ],
  [
    [
      [-179.917369,-16.501783],
      [-180,-16.555217],
      [-180,-16.067133],
      [-179.79332,-16.020882],
      [-179.917369,-16.501783]
    ]
  ]
]
```

So after writing this up, I believe I found the solution with b0a0de3 ...

MoonE:
Thanks for your help so far, @ahocevar.
If you have an idea on how to solve that issue any help would be greatly appreciated.

ahocevar:
I need to give this a more thorough investigation. 2cb01f47b69390b391b0905b3f26ea59be8819a5 is not a good fix, because it makes the declutter groups grow infinitely when replaying the same executor group over and over (e.g. during panning).

ahocevar:
I'm tempted to rework and simplify the whole grouping code. Initially the groups were only meant for keeping image and label together, but then, with different handling of multi geometries, it became overly complicated. I'm going to propose a commit here once I have it.

ahocevar:
I created a new decluttering implementation in #11592, which should also address the problem this pull request was meant to solve.

PR #11537, opened 2020-09-08
Add rotation and rotateWithView to ol/style/Circle clone method and options

### LINKED ISSUE #11520: ol/style/Circle .clone() does not set rotation or rotateWithView
**Describe the bug**
Although rotation and rotateWithView are not useful for a pure circle, if it is scaled in two dimensions those properties are relevant for the resulting ellipse.  But they is not set when ol/style/Circle is cloned.

**To Reproduce**
https://codesandbox.io/s/quirky-mccarthy-12jgq?file=/main.js
The style and its clone should both be rotated and rotate with the view.  However the cloned style has neither property.

**Expected behavior**
As a minimum these properties should be set in the clone in the same manner as opacity.  A better option would be to set them as constructor options, with a note in the API descriptions that they are meaningful only when used in conjunction with a two dimensional scale.

mike-000:
Fixes #11520

Add rotation and rotateWithView to clone method and options

Add these to clone test
Also test that scale and displacement arrays are cloned


ahocevar (approved):
Thanks, @mike-000

PR #11540, opened 2020-09-09
Avoid unnecessary transform in the MVT format

gberaudo:
When reading features using `ol.render.Feature` a mandatory geometry transform was applied.
The reprojection now only occurs when an `option.featureProjection` is defined.

With this change it is now easy again to use the MVT format as a standalone way to read features in the native tile projection.
Since there is no reprojection, there is no need either to define an extent and a world extent, for that use-case.

@ahocevar, I require this change to implement an experimental MapBoxVectorTile imagery provider for Cesium, in the OL-Cesium project.

### DIFF
--- a/src/ol/format/MVT.js
+++ b/src/ol/format/MVT.js
@@ -194,7 +194,7 @@ class MVT extends FeatureFormat {
         values,
         id
       );
-      feature.transform(options.dataProjection, options.featureProjection);
+      feature.transform(options.dataProjection);
     } else {
       let geom;
       if (geometryType == GeometryType.POLYGON) {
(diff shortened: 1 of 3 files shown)


ahocevar:
@gberaudo So you want the coordinates in tile-relative pixel coordinates? Wouldn't it be enough to have an if-check like in your description, i.e. only transform when a `featureProjection` is defined?
```js
if (options.featureProjection) {}
```

ahocevar (changes requested):
Since what `ol/render/Feature#transform()` cannot reproject, but only transform from pixel to map coordinates, I think it is better to remove `featureProjection` from the equation. Feel free to use a7c3e0deb0. Your test still passes with that change.

ahocevar (approved):

PR #11543, opened 2020-09-09
wmts matrixset null pointer fix and improved handling of Identifier

sweco-dkjesh:
Nullpointer bug was introduced in 6.4.0 when fixing issue #10981. Commit #9cab1215c08aaf813ab9ee3be1116ef074aca1ea.

Fix nullpointer error when a matrixSet is not found in the matrixsetobj when matrix limits exist.
Also check for matrix match by prepending matrixSetObj.Identifier to support more wmts services.


ahocevar (approved):
Thanks, @sweco-dkjesh

PR #11559, opened 2020-09-14
Only draw visible worlds

MoonE:
With a multi-world source and projection currently every world until world zero or from world zero is rendered depending on if the current world(s) are left or right of world zero. So if worlds -100 and -99 are in the current view all worlds between are also rendered. It should be enough to only render the visible two worlds.

This PR changes the order of world rendering, but I don't think it matters, does it?
Before: 
- world zero (always), 
- negative worlds increasing order (if to the left)
-  positive worlds decreasing order (if to the right)

Now:
- always increasing from left to right

### DIFF
--- a/src/ol/renderer/canvas/VectorLayer.js
+++ b/src/ol/renderer/canvas/VectorLayer.js
@@ -219,84 +219,37 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
       viewHints[ViewHint.ANIMATING] || viewHints[ViewHint.INTERACTING]
     );
 
-    const transform = this.getRenderTransform(
-      center,
-      resolution,
-      rotation,
-      pixelRatio,
-      width,
-      height,
-      0
-    );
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
This is a nice simplification, and the order does not matter indeed. Thanks, @MoonE!

ahocevar (approved):

ahocevar:
I'm going to check the failing rendering test to see if there aren't any unexpected changes.

MoonE:
It rendered one world too many. I think the failed render test was related to #11523.

MoonE:
I did not account for the offset of the projection which does not start at zero.

ahocevar (approved):
Thanks, @MoonE!

MoonE:
I think I messed up with the last change I made 
https://github.com/openlayers/openlayers/compare/6e01f2139965f0d32ff62a27c3b85fab199a0a37..85499d7daf9ac5871d2ef3849be1fc1a0859ee86
It now only works correctly if the projection extents' x-midpoint is at zero.

PR #11561, opened 2020-09-15
Constrain resolution when fit called without duration

### LINKED ISSUE #11560: View constraint not resolved after call to fit
**Describe the bug**
After a call to `fit` the `View` can be outside its configured `extent` when no `duration` is set.
This is because the unconstrained `center` is used here:
https://github.com/openlayers/openlayers/blob/2de80e04e4c16799c43cb2b9b2e88b62db396ab1/src/ol/View.js#L1306-L1311

**To Reproduce**
1. Go to https://codesandbox.io/s/unruffled-hawking-2lod7?file=/src/index.js
2. See the white border at the left of the map which is outside the view's extent
3. Pan the map and the view is constrained again.

**Expected behavior**
The view should be constrained to its configured extent.

M393:
Fixes #11560

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1290,14 +1290,14 @@ class View extends BaseObject {
     centerRotY += ((padding[0] - padding[2]) / 2) * resolution;
     const centerX = centerRotX * cosAngle - centerRotY * sinAngle;
     const centerY = centerRotY * cosAngle + centerRotX * sinAngle;
-    const center = [centerX, centerY];
+    const center = this.getConstrainedCenter([centerX, centerY], resolution);
     const callback = options.callback ? options.callback : VOID;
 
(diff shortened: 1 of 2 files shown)


jahow:
Thanks a lot for proposing a fix. The changes makes sense to me. Do you think you can add a test that is failing without the fix, and passing with it?

M393:
Some of the other View tests use a size of [200, 200], but here:
https://github.com/openlayers/openlayers/blob/2de80e04e4c16799c43cb2b9b2e88b62db396ab1/src/ol/View.js#L1714-L1716
the default size ([100, 100]) is used regardless.
I have not investigated if this is a problem, just found it a bit strange.

jahow (approved):
Perfect, thanks a lot.

PR #11563, opened 2020-09-15
Fix end world calculation if projection is not symmetric

MoonE:
I do hope this is correct now, please merge if you agree that the current behaviour could be wrong as outlined in https://github.com/openlayers/openlayers/pull/11559#issuecomment-693023339


ahocevar (commented):
Thanks for your continued effort on this, @MoonE. Can you please add a rendering test to avoid regressions with this? Do you have a projection an mind that would trigger the wrong behavior?

MoonE:
Actually I do not know a projection like that, which is probably because I so far only used wgs84 and the projection used by OpenLayers ...
If you know how to create one like that I'd appreciate it.

ahocevar:
@MoonE: The code below defines a projection like EPSG:4327, but with a center coordinate of `[20037508.342789244, 0]`:
```js
import {
  Projection,
  addCoordinateTransforms,
  fromLonLat,
} from '../src/ol/proj.js';
import {fromEPSG4326, toEPSG4326} from '../src/ol/proj/epsg3857.js';

const projection = new Projection({
  code: 'custom',
  units: 'm',
  global: true,
  extent: [0, 0, 40075016.68557849, 40075016.68557849],
});

addCoordinateTransforms(
  'EPSG:4326',
  projection,
  function (coord) {
    const converted = fromEPSG4326(coord);
    converted[0] += 20037508.342789244;
    return converted;
  },
  function (coord) {
    return toEPSG4326([coord[0] - 20037508.342789244, coord[1]]);
  }
);
```
However, I do not notice a difference on rendered maps before and after this pull request, when using the above projection as view projection.

MoonE:
Thank you, the projection in your comment it may draw one world more.
With the zero coordinate to the right the last world may be missing as long as more than one world is drawn.
```js
const projection = new Projection({
  code: 'custom',
  units: 'm',
  global: true,
  extent: [-40075016.68557849, -20037508.342789244, 0, 20037508.342789244],
});

addCoordinateTransforms(
  'EPSG:4326',
  projection,
  function (coord) {
    const converted = fromEPSG4326(coord);
    converted[0] -= 20037508.342789244;
    return converted;
  },
  function (coord) {
    return toEPSG4326([coord[0] + 20037508.342789244, coord[1]]);
  }
);
```

ahocevar:
Ah you're right @MoonE. Your code snippet has the correct projection extent. Are you able to create a rendering test with your projection definition, to verify that the correct number of worlds is drawn?

ahocevar:
@MoonE Let me know if you need further guidance with how to create rendering tests.

MoonE:
I added a test with 2a81a7a, is it ok or should I test a different case too?

ahocevar:
That looks great @MoonE, and testing this special case should be enough also for the generic case.

ahocevar (approved):
Thanks, @MoonE!

Issue #11412, opened 2020-08-07
Add GIF in icon


GeorgeCodeHub:
I'm trying to put animated icons all over my map but for some reason, it keeps them only in the first frame. Is there a certain setting that I need to add? In OpenLayers 2 it was able to do it no problem but in 6 it doesn't for some reason.

Below are two links. One with an example and the other with the link to the gif.

Example:  https://codesandbox.io/s/fervent-keller-wydlg?file=/main.js

Icon: https://i.pinimg.com/originals/a9/02/16/a902167cae8538769ad56ec7514ce73c.gif

Note: I need them to be as features and not as markers because I want to have click events on them.

Thank you for your time.

mike-000:
Only one frame of a GIF can be rendered to canvas, so they cannot animate as features.  You could create them as features, hide the vector layer under your base map and for each feature create an overlay which will animate.  If `stopEvent` is false clicks will reach the map so you can detect the hidden features https://codesandbox.io/s/festive-bardeen-g2j3s?file=/main.js

ahocevar:
Actually there is a way to render animated gif styles. You need a gif decoder library, e.g. [gifler](https://themadcreator.github.io/gifler/). Then you can set up an Icon style with the gifler controlled canvas, and call `render()` on the map when gifler renders a frame. See https://codesandbox.io/s/optimistic-bush-scgk3?file=/main.js for a working example.

The relevant code is:
```js
const gifUrl = "./data/a902167cae8538769ad56ec7514ce73c.gif";
const gif = gifler(gifUrl);
gif.frames(
  document.createElement("canvas"),
  (ctx, frame) => {
    if (!iconFeature.getStyle()) {
      iconFeature.setStyle(
        new Style({
          image: new Icon({
            img: ctx.canvas,
            imgSize: [frame.width, frame.height]
          })
        })
      );
    }
    ctx.drawImage(frame.buffer, frame.x, frame.y);
    map.render();
  },
  true
);
```

mike-000:
Nice. That should be added as an example. There have been questions on here and stackoverflow and none have an answer using a feature with icon style.  When I tried it the dancing banana used in a lot of questions there was a bad ghosting effect, but adding `ctx.clearRect(0, 0, frame.width, frame.height);` before the `drawImage` fixes it.

jipexu:
@ahocevar  so cool ! as @mike-000  said it will be a really a usefull example ...?

GeorgeCodeHub:
[...]

PR #11565, opened 2020-09-16
Animated GIF example

mike-000:
Adds an animated GIF example similar to that in https://github.com/openlayers/openlayers/issues/11412#issuecomment-685783562 using a different icon and Gifler from CDN to avoid a dependency.


marcjansen:
Cool thanks, @mike-000.

For reference, here is [the online example](https://deploy-preview-11565--ol-site.netlify.app/examples/animated-gif.html).

Where is ghe gif coming from? Did you create it from scratch?

The gif is huge on my mobile, can we somehow scale it?

Would be awesome if you could enhance this a bit and please answer the question wrt the source of the gif.

mike-000:
The size is much as I expected, so it allows the hit detection to be easily seen with pointermove, but I can reduce it to be more mobile friendly.  It is based on https://webqoe.telecom-paristech.fr/ although watermarks indicate that originated from PicMix where at least 4 people claim to be authors.  There are also many monochrome versions (without watermarks) online so I converted it to single color using https://ezgif.com/

marcjansen:
Hm. Any chance we can include a CC-BY globe? And simply attribute it in the text of the example?

I'd much prefer that.

Thanks for your work, and I hope you understand I only want a clearly usable image to be included in the library.

The size issue is really not that important.

marcjansen:
Cool thanks again.

marcjansen (approved):

Issue #11564, opened 2020-09-16
Data URLs in example codesandboxes


mike-000:
**Describe the bug**
Data URLs for binary resources are clearly not working as intended.  I have experimented with CodeSandbox's own example sandbox https://codesandbox.io/s/define-api-post-request-forked-4ptte?file=/index.js and it seems that the `isBinary` setting is ignored when the content does not begin with `http`.  That would suggest the problem could be fixed by using a full path url in `common.js`, for binary data, for example

      if (isImage) {
        resolve ({
          isBinary: true,
          content: new URL(resource, window.location.href).href
        })
      } else {
        const xhr = new XMLHttpRequest();

while keeping the current setup for text data (although as the example shows all types of resource could be loaded externally).

PR #11566, opened 2020-09-16
Fix Vector Clipping Layer geojson url so it is imported into codesandbox

mike-000:
Fixes the Vector Clipping Layer example which does not work in codesandbox when edit is clicked as an unneeded `./` in the geojson url prevents it being treated as a resource.


tschaub (commented):
Alternatively, `(\.\/)?` could be added to the pattern that matches these to make it more robust.

mike-000:
Closed as #11564 makes this change unnecessary

PR #11568, opened 2020-09-17
Fix raster source crash bug when using multiple threads

### LINKED ISSUE #11178: RasterSource threads >1 throws `Cannot read property 'slice' of undefined`
**Describe the bug**
Settings threads to `2` throws `Cannot read property 'slice' of undefined`.
https://github.com/openlayers/openlayers/blob/254306496ca5ecd9334aa1e7cae6df29f30eedd1/src/ol/source/Raster.js#L286 Buffers seems to be an array of length 1 instead of an array of length `threads`

**To Reproduce**
1. https://openlayers.org/en/latest/examples/raster.html and open example or go to (different tiles and threads added) https://codesandbox.io/s/raster-0pprs?file=/index.js 
2. See error

**Expected behavior**
No error and multiple working workers

rgroothuijsen:
This PR fixes #11178 by correcting the assignment of buffer slices to workers. Previously, if a raster source was processed by multiple threads, this would instead result in an error message. This is caused by the use of a presumably incorrect loop variable, which will crash the application if the number of threads exceeds the number of buffers.

### DIFF
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -283,7 +283,7 @@ export class Processor extends Disposable {
           const offset = i * segmentLength;
           const slices = [];
           for (let j = 0, jj = buffers.length; j < jj; ++j) {
-            slices.push(buffers[i].slice(offset, offset + segmentLength));
+            slices.push(buffers[j].slice(offset, offset + segmentLength));
           }
           this._workers[i].postMessage(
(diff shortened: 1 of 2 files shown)


tschaub (approved):

tschaub:
Thanks for the for, @rgroothuijsen!

PR #11569, opened 2020-09-17
Fix binary data exports to codesandbox in examples

### LINKED ISSUE #11564: Data URLs in example codesandboxes

mike-000:
Fixes #11564

Replaces data URLs for binary resources with uploads from the full path which will work as intended in codesandbox.

Also handles `./` type relative paths as suggested in https://github.com/openlayers/openlayers/pull/11566#pullrequestreview-490241841

Restore IE compatibility to the Edit button (codesandbox will not run in IE but a link is opened which can be used by another browser)

### DIFF
--- a/examples/resources/common.js
+++ b/examples/resources/common.js
@@ -36,11 +36,12 @@
     codepenButton.href = form.action;
     codepenButton.addEventListener('click', function (event) {
       event.preventDefault();
-      const html = document.getElementById('example-html-source').innerText;
-      const js = document.getElementById('example-js-source').innerText;
+      const innerText = document.documentMode ? 'textContent' : 'innerText';
+      const html = document.getElementById('example-html-source')[innerText];
+      const js = document.getElementById('example-js-source')[innerText];
       const workerContainer = document.getElementById('example-worker-source');
-      const worker = workerContainer ? workerContainer.innerText : undefined;
-      const pkgJson = document.getElementById('example-pkg-source').innerText;
+      const worker = workerContainer ? workerContainer[innerText] : undefined;
+      const pkgJson = document.getElementById('example-pkg-source')[innerText];
 
       const unique = new Set();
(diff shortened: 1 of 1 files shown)


mike-000:
It is apparent that there is inconsistent behaviour between browsers when using the `.innerText` method, which can also be seen in the example in https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText

The expected behaviour as seen in most browsers is
image

In what looks like a bug in the current version of Chrome it produces the same result as `.textContext`
image
(row numbers are not included in the export using either method in Chrome, while they result in extra empty line in some other browsers if `.textContext` is used.

IE is different from both the above
image
Linebreaks are not included the export using `.innerText` but `.textContent` is similar to Chrome.  Missing linebreaks at the end of comments is major problem and requires a workaround.

mike-000:
This is ready for review.  Note that codesandbox cannot import binary files without CORS access (e.g. it cannot work with deploy-preview examples).  I have set up CORS enabled mock-ups saved from latest examples but using the new minified `resources/common.js` at
https://mikenunn.net/data/test/icon.html
https://mikenunn.net/data/test/layer-clipping-vector.html
https://mikenunn.net/data/test/kml.html

ahocevar (approved):
Thanks, @mike-000

PR #11571, opened 2020-09-19
Replace countryflags.io with flagcdn.com in the Style Renderer example

mike-000:
https://flagcdn.com/ has much better quality images than https://www.countryflags.io/ (up to 2560px wide, although that would affect performance and 320px is a good compromise), and they will fit correctly without cropping.


tschaub (approved):

tschaub:
Looks nice.  Thanks, @mike-000.

PR #11579, opened 2020-09-21
Replace Animated GIF example globe gif

mike-000:
Use better quality image based on public domain landmask image at https://ian.macky.net/pat/map/globe.html and data at http://www.shadedrelief.com/natural3/pages/extra.html


tschaub (approved):

tschaub:
Thanks, @mike-000.

PR #11587, opened 2020-09-24
Determine orientation by actual text start and end x

### LINKED ISSUE #11584: Upside-down text with placement: line
**Describe the bug**
Upside down text with `placement:  'line'`

**To Reproduce**
1. Go to https://codesandbox.io/s/line-placement-upside-down-text-wwu5z?file=/src/index.js
2. The text is upside down

**Expected behavior**
The text should be upright for better legibility.

The check if it should be reversed uses the begin and end of the entire possible placement path instead of only the section where the text is being placed. https://github.com/openlayers/openlayers/blob/705e328fa096ee3f63fa3c609e55a32315c50c0e/src/ol/geom/flat/textpath.js#L54
To get the correct result it should check the first x1 and last x2 (which would need to be calculated beforehand).
https://github.com/openlayers/openlayers/blob/705e328fa096ee3f63fa3c609e55a32315c50c0e/src/ol/geom/flat/textpath.js#L74-L82

MoonE:
Fixes #11584 
Instead of using the first and last x coordinate of the entire line it interpolates the x for the exact begin and end of the text.
Characters on the same line segment are drawn with one instruction, see the `it('renders angled text', function () {` test for an example.

Todo:
- ~Text rendering along a path is broken with a negative x scale, not related to this change though~ no change with this PR
- ~I do not yet understand in which case the rotation is set, please tell me if you know~

### DIFF
--- a/src/ol/geom/flat/textpath.js
+++ b/src/ol/geom/flat/textpath.js
@@ -34,71 +34,89 @@ export function drawTextOnPath(
   cache,
   rotation
 ) {
-  const result = [];
+  let x2 = flatCoordinates[offset];
+  let y2 = flatCoordinates[offset + 1];
+  let x1 = 0;
+  let y1 = 0;
+  let segmentLength = 0;
+  let segmentM = 0;
+
+  function advance() {
+    x1 = x2;
+    y1 = y2;
+    offset += stride;
+    x2 = flatCoordinates[offset];
+    y2 = flatCoordinates[offset + 1];
+    segmentM += segmentLength;
+    segmentLength = Math.sqrt((x2 - x1) * (x2 - x1) + (y2 - y1) * (y2 - y1));
+  }
+  do {
+    advance();
+  } while (offset < end - stride && segmentM + segmentLength < startM);
+
+  let interpolate = (startM - segmentM) / segmentLength;
+  const beginX = lerp(x1, x2, interpolate);
+  const beginY = lerp(y1, y2, interpolate);
+
+  const startOffset = offset - stride;
+  const startLength = segmentM;
+  const endM = startM + measureAndCacheTextWidth(font, text, cache);
+  while (offset < end - stride && segmentM + segmentLength < endM) {
+    advance();
+  }
+  interpolate = (endM - segmentM) / segmentLength;
+  const endX = lerp(x1, x2, interpolate);
+  const endY = lerp(y1, y2, interpolate);
 
   // Keep text upright
   let reverse;
   if (rotation) {
-    const rotatedCoordinates = rotate(
-      flatCoordinates,
-      offset,
-      end,
-      stride,
-      rotation,
-      [flatCoordinates[offset], flatCoordinates[offset + 1]]
-    );
-    reverse =
-      rotatedCoordinates[0] >
-      rotatedCoordinates[rotatedCoordinates.length - stride];
+    const flat = [beginX, beginY, endX, endY];
+    rotate(flat, 0, 4, 2, rotation, flat, flat);
+    reverse = flat[0] > flat[2];
   } else {
-    reverse = flatCoordinates[offset] > flatCoordinates[end - stride];
+    reverse = beginX > endX;
   }
 
-  const numChars = text.length;
-
-  let x1 = flatCoordinates[offset];
-  let y1 = flatCoordinates[offset + 1];
-  offset += stride;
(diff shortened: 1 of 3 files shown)


MoonE:
You can test it here:
https://deploy-preview-11587--ol-site.netlify.app/examples/text-along-path-test

ahocevar:
@MoonE:

> * I do not yet understand in which case the rotation is set, please tell me if you know

Basically `drawTextOnPath` will be called with a rotation when rendering a vector layer, and without a rotation when rendering a vector tile layer. So for testing with rotation, you could use the `street-labels` example, and for testing without rotation the `mapbox-style` example when zoomed in.

ahocevar:
@MoonE Do you need help to finalize this pull request? I'd like to see this fixed. Let me know where I can help.

MoonE:
Thank you for checking in, @ahocevar.

I was struggling with negative x scale values, but I guess it isn't too important, and didn't work well from the start anyway.

Looking at the failing test, I'm not too sure if they were correct to begin with.
E. g. the first test expects `[ [ 40, 0, 5, 0, 'foo' ] ]` while the function now returns `[ [ 50, 0, 15, 0, 'foo' ] ]`
The `40` and `5` in the expected result actually come from the first character position and width instead of from the whole word.
https://github.com/openlayers/openlayers/blob/f102e11bbe8304a9c8efa8caa78eb9e630a98daa/src/ol/geom/flat/textpath.js#L104-L106
The resulting position seems to be about the same.

src/ol/geom/flat/textpath.js
```diff
@@ -34,74 +34,114 @@ export function drawTextOnPath(
...
       let delta = angle - previousAngle;
-      angleChanged = angleChanged || delta !== 0;
-      delta +=
-        delta > Math.PI ? -2 * Math.PI : delta < -Math.PI ? 2 * Math.PI : 0;
+      delta += delta > PI ? -2 * PI : delta < -PI ? 2 * PI : 0;
       if (Math.abs(delta) > maxAngle) {
         return null;
       }
     }
     previousAngle = angle;
```
MoonE, line 119: I would have preferred this line to be after the if block with the `continue`  below to measure the angle change between characters instead of the angle change of each line segment. When there are multiple angle changes between two successive characters they can exceed the configured `maxAngle`, but this cannot be taken into account when the line sub segment is determined where the text should be placed here:
https://github.com/openlayers/openlayers/blob/f102e11bbe8304a9c8efa8caa78eb9e630a98daa/src/ol/geom/flat/straightchunk.js#L14

ahocevar (commented):
Thanks for your continued effort on this @MoonE! The results look good, I only noticed two problems by playing with the [vector-labels.html] example:

* For some lines, when `align` is set to `left`, text starts on the first line segment and does not follow the other segments. Instead, it exceeds the first segment:
   Untitled
* When `align` is not set and `overflow` is `true`, some labels start in the middle of a segment instead of being centered on that segment, like it was the case previously:
   Untitled

Maybe both are related? I don't have an explanation for this. Maybe this makes more sense to you, @MoonE ?

MoonE:
I can't seem to reproduce your first example. For me the old and new placements look the same.
Maybe you forgot to hit the refresh button for the line layer? But I'm not sure what combination would cause this.

Your second example might be caused by the max angle beeing set to 45°.

The example not updating was something frustrating me for my testing and I modified it in https://github.com/MoonE/openlayers/tree/example-vector-labels. I need to make it a bit prettier before making a PR for this, but should be a better testing experience.

ahocevar:
Thanks for checking again @MoonE. I did hit "Refresh" in the first screenshot, and also double-checked in different browsers. In https://openlayers.org/en/main/examples/vector-labels.html, with the same settings, the lines look like this:
Untitled

The second example, with the same settings, looks like this:
Untitled

Did you try with https://deploy-preview-11587--ol-site.netlify.app/examples/vector-labels.html, or with your local version? Maybe you forgot to push a change?

MoonE:
I can reproduce it with a pixelRatio of 2.

ahocevar:
Thanks @MoonE! My device does indeed have a pixel ratio of 2.

MoonE:
Both cases you reported seem to be fixed now. The text text length was not scaled correctly.

ahocevar (approved):
Looks great now. Thanks, @MoonE!

PR #11588, opened 2020-09-24
Fix text alignment for start / end options

### LINKED ISSUE #11586: Text alignment wrong for start and end
**Describe the bug**
`start` and `end` text-align are swapped.
https://github.com/openlayers/openlayers/blob/705e328fa096ee3f63fa3c609e55a32315c50c0e/src/ol/render/canvas/TextBuilder.js#L31-L36
**To Reproduce**
1. Copy+paste the data url to get a test page (CTRL + U to view source):
`data:text/html;base64,PCFET0NUWVBFIGh0bWw+CjxodG1sPgo8aGVhZD4KCTx0aXRsZT5UZXh0IGFsaWdubWVudDwvdGl0bGU+Cgk8bWV0YSBjaGFyc2V0PSJVVEYtOCIgLz4KPC9oZWFkPgo8Ym9keT4KCTxkaXYgc3R5bGU9InRleHQtYWxpZ246IGxlZnQiPmxlZnQ8L2Rpdj4KCTxkaXYgc3R5bGU9InRleHQtYWxpZ246IHN0YXJ0Ij5zdGFydDwvZGl2PgoJPGRpdiBzdHlsZT0idGV4dC1hbGlnbjogcmlnaHQiPnJpZ2h0PC9kaXY+Cgk8ZGl2IHN0eWxlPSJ0ZXh0LWFsaWduOiBlbmQiPmVuZDwvZGl2Pgo8L2JvZHk+CjwvaHRtbD4=`
2. Compare the start / left and end / right values in the page with the code in the `TEXT_ALIGN` enum and / or this example:
https://openlayers.org/en/main/examples/vector-labels.html  

**Expected behavior**
The start should be to left, end to the right, at least for normal ltr text which is the default.

MoonE:
Fixes #11586

### DIFF
--- a/src/ol/render/canvas/TextBuilder.js
+++ b/src/ol/render/canvas/TextBuilder.js
@@ -30,10 +30,10 @@ import {matchingChunk} from '../../geom/flat/straightchunk.js';
  */
 export const TEXT_ALIGN = {
   'left': 0,
-  'end': 0,
+  'start': 0,
   'center': 0.5,
   'right': 1,
-  'start': 1,
+  'end': 1,
   'top': 0,
   'middle': 0.5,
   'hanging': 0.2,


ahocevar (commented):
See https://github.com/openlayers/openlayers/issues/11586#issuecomment-701228040 for the historic reasons why `start` and `end` behave like they currently do.

I do not know enough about RTL text, but ideally `start` and `end` would have the correct meaning for both LTR and RTL text, not only RTL.

ahocevar:
@MoonE I think #11722 should be a good replacement for this one. I'd be grateful if you could review that. Thanks!

PR #11590, opened 2020-09-25
Export all the Event sub class

sbrunner:
Actually only some are exported

This is useful to type the received events


ahocevar (approved):
Thanks, @sbrunner

PR #11591, opened 2020-09-25
Make permalink example less complex

kannes:
https://openlayers.org/en/latest/examples/permalink.html used hard to understand math to do some number formatting, I switched it to .toFixed() instead. It might be a minor thing for experienced JS developers but had me confused for a moment and feels totally unnecessary.

I also took the chance to fix fractional zooming not being stored as such.


ahocevar (approved):
Nice improvement, thanks!

PR #11594, opened 2020-09-27
Display current OS copyright date in the Raster Reprojection example

mike-000:
Display current OS copyright date to keep up with source updates,
see https://www.arcgis.com/home/item.html?id=7561029c176d43d09a5aa1180ada309a

Also increase the EPSG:27700 extent used in the example beyond the usual definition to the visible extent of the OS Overview layer


ahocevar (approved):
Thanks, @mike-000

PR #11597, opened 2020-09-27
Improved performance when removing feature with only one Select

### LINKED ISSUE #11582: select.getFeatures().clear() takes a very long time to complete
**Describe the bug**
The time taken by the select.getFeatures().Clear() function grows exponentially with the number of objects:
- 1000 (1s)
- 10000 (3s)
- 30000 (>120s)

If the layer is visible or not the execution time is the same. No events are attached to select or features collection.
All features shares same style instance via same styleFunction.

**To Reproduce**
Select 1000, 10000 and 30000 features and use select.getFeatures().Clear()

**Expected behavior**
select.getFeatures().Clear() should be completed within a reasonable time.

MoonE:
Fixes #11582

This implements the suggested change by @bojanbizjak from https://github.com/openlayers/openlayers/issues/11582#issuecomment-699668573
There was a bug in the suggested change when accessing the previous style and the feature was not added to some of the other selects which I fixed.
As a further improvement I stop searching for the feature once it is found once and do not search the current Select.

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -391,25 +391,23 @@ class Select extends Interaction {
    * @private
    */
   restorePreviousStyle_(feature) {
-    const key = getUid(feature);
-    const selectInteractions = /** @type {Array<Select>} */ (this.getMap()
-      .getInteractions()
-      .getArray()
-      .filter(function (interaction) {
-        return (
-          interaction instanceof Select &&
-          interaction.getStyle() &&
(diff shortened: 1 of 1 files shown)


MoonE:
This change may be a problem when a feature is added back during the collection's remove event though...

ahocevar:
> This change may be a problem when a feature is added back during the collection's remove event though...

I think we can live with that.

ahocevar (approved):

PR #11607, opened 2020-09-28
Reset attributions control to collapsible when no uncollapsible sources are visible

### LINKED ISSUE #11595: Attribution control does not return to collapsible when no attributionsCollapsible: false sources are visible
**Describe the bug**
Unless collapsible is specified for the control it should be collapsible when no `attributionsCollapsible: false` sources are visible.  However it does not reset when all such layers are hidden or removed.

**To Reproduce**
See the Raster Reprojection example.  When the OSM layer is replaced by WMS the attributions should be collapsible, but are not.  If the example is changed to open with the WMS layer the attributions are collapsible, then if the OSM layer is the used the attributions become non-collapsible as expected, but switching the layers back again has no effect.

**Expected behavior**
The process should be reversible.

mike-000:
Fixes #11595

When there is no override reset the attributions to collapsible when no sources requiring uncollapsible attributions are visible

### DIFF
--- a/src/ol/control/Attribution.js
+++ b/src/ol/control/Attribution.js
@@ -70,6 +70,12 @@ class Attribution extends Control {
      * @private
      * @type {boolean}
      */
+    this.userCollapsed_ = this.collapsed_;
+
+    /**
+     * @private
+     * @type {boolean}
+     */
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Nice, thanks @mike-000!

Issue #11608, opened 2020-09-28
Allow modify interaction to modify only one of several features at a time


M393:
**Is your feature request related to a problem? Please describe.**
Currently the Modify interaction will modify more than one vertex if
- multiple Features have the same coordinates
- a new vertex is inserted on identical segments of the same / different geometries (That might be classifiable as a bug)

**Describe the solution you'd like**
If multiple lines geometries share a corner, I would like it to only modify one of the features at a time or allow a callback function to filter which will be modified.

ahocevar:
I agree that the latter is a bug. The former can easily be achieved by setting only the feature you're modifying in the interaction's `source` or `features`.

M393:
Thanks for your reply, @ahocevar 
I do want to allow the user to modify multiple features, but only one at a time, how can I achieve that?

ahocevar:
> I do want to allow the user to modify multiple features, but only one at a time, how can I achieve that?

One way would be to let the user click the feature they want to modify before actually modifying it. That way there won't be any surprises, especially if you highlight the feature that is currently being modified.

M393:
Thanks, @ahocevar, this will do fine.

PR #11609, opened 2020-09-28
Modify interaction insert only one vertex

M393:
Modify interaction now only inserts one vertex if there are multiple overlapping segments of the same feature.
See the second point in #11608

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -1070,6 +1070,7 @@ class Modify extends PointerInteraction {
       if (dist <= this.pixelTolerance_) {
         /** @type {Object<string, boolean>} */
         const vertexSegments = {};
+        vertexSegments[getUid(closestSegment)] = true;
 
         if (
           node.geometry.getType() === GeometryType.CIRCLE &&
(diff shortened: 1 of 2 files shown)


ahocevar (commented):
This change looks good. Would you be able to add a test? I know writing tests for the Modify interaction is a bit of an effort, but worth it in the long term. With copying and pasting one of the existing tests you should have a good starting point.

M393:
Adding the test did find a bug, the closest segment was always added afterwards ...

ahocevar (approved):
Thanks, @M393! There is another bug (not related to your changes), which could maybe be easily fixed along the way: Editing of multi geometries is broken. Try to edit the multipolygon at the top of https://deploy-preview-11609--ol-site.netlify.app/examples/modify-test.html: it is not possible to add a vertex.

Let me know if you're willing to address that, otherwise this is good to merge.

M393:
I think that is working as inteded:
https://github.com/openlayers/openlayers/blob/038739955466ee3f6a81d8524a07aa04d6522e26/examples/modify-test.js#L318-L328

ahocevar:
@M393 I meant something else:

Sep-29-2020 15-57-51

 But maybe it is better to handle that in a separate pull request, because the issue is not related to this pull request. I just thought it might be a low hanging fruit now that you are familiar with that code.

M393:
That's what the `insertVertexCondition` is for.
As it is now it prevents inserting new vertexes for any of the Polygon types.
Inserting a new vertex works if the option is removed.

ahocevar:
Thanks, @M393. You are absolutely right - I had missed that `insertVertexCondition` is configured in that example. Sorry for the noise, and thanks again for this improvement. Merging now.

Issue #11612, opened 2020-10-01
Invalid mouse inputs for map in different window


elnabo:
**Describe the bug**
When you render the map in a different window (like a popup window) the mouse event are not working properly since they are binding on `document`

**To Reproduce**
1. Render a map in a `window.open()`
2. Interact with it,
3. See lack of interaction or invalid interaction

**Expected behavior**
Works as in base window

PR #11613, opened 2020-10-01
Allow mouse input if map is rendered in an other window

elnabo:
This propose a fix for #11612 by attempting to lisen on the `targetElement` `ownerDocument` instead of the default `document`.

An example as been added.

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -239,19 +239,15 @@ class MapBrowserEventHandler extends EventTarget {
     this.down_ = pointerEvent;
 
     if (this.dragListenerKeys_.length === 0) {
+      const doc = this.map_.getOwnerDocument();
       this.dragListenerKeys_.push(
         listen(
-          document,
+          doc,
           MapBrowserEventType.POINTERMOVE,
           this.handlePointerMove_,
           this
         ),
-        listen(
-          document,
-          MapBrowserEventType.POINTERUP,
-          this.handlePointerUp_,
-          this
-        ),
+        listen(doc, MapBrowserEventType.POINTERUP, this.handlePointerUp_, this),
         /* Note that the listener for `pointercancel is set up on
          * `pointerEventHandler_` and not `documentPointerEventHandler_` like
          * the `pointerup` and `pointermove` listeners.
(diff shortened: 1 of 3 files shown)


jahow:
Hi, thanks for your contribution!

Have you checked that this does not break other features? AFAIK there are many places where the global `document` is used without additional check.

elnabo:
Hello,

I'm unsure about the other places due to my general lack of knowledge of OpenLayers. We have tested those changes with a few features (selection, d&d, click, zoom) with success.

Some features might still be broken if the map is on an secondary window but this shouldn't have any effect if its on the main document

jahow:
After a bit more code looking I'd say the following things might break as well:
* zoom slider (uses mouse events as well)
* full screen control (uses the `document` global extensively)

Fixing the full screen control might be a bit overkill for this usage, but the zoom slider should be easily doable.

Also I'd suggest creating a new (public but not API) method on the `PluggableMap` class, something like `getOwnerDocument` to help factorize the logic you added in your commit.

Also you missed a `document` in the following line (which is required if anyone ever feels like using a web component inside a popup...):
https://github.com/openlayers/openlayers/blob/main/src/ol/MapBrowserEventHandler.js#L277

Finally, I'm not sure an example adds anything here to be honest, especially since the stylesheet is missing in the popup. Simply mentioning on the map `targetElement` option that this can be an element from another document is enough IMO. See:
https://github.com/openlayers/openlayers/blob/aa69c5d3aad8103224be989f6c459e860cc2ed7b/src/ol/PluggableMap.js#L130-L133

Apart from that, I think this contribution is good to be merged! I have tested it with the web component example just to be sure, since its changes are very close to https://github.com/openlayers/openlayers/pull/11024, but noticed no regression.

elnabo:
I've done what you suggested/asked.

Do I remove the 2 example files ?

jahow:
> Should I remove the 2 example files ?

I would advise so, yes. To be honest having an example might be interesting but IMO it should showcase some kind of exchange between the main page and the popup, either controlling the map from the main page or receiving feature info from the popup, things like that.

I'd be curious to hear other people thoughts on this though.

jahow:
> I've done what you suggested/asked.

Changes are looking good now, thanks! could you please rebase all your commits into one before merging?

elnabo:
Squash is done.

jahow (approved):

jahow:
Great, thanks again @elnabo!

PR #11618, opened 2020-10-03
Update the Marker Animation example

mike-000:
The reference to jsFiddle  is outdated and there is now no reason why the polyline cannot be fetched (as it would in a real application interfacing a directions API) as CodeSandbox does not have that limitation.

The original example could not animate smoothly as it predated the introduction of the `getCoordinateAt()` method.  Basing the animation on distance instead of coordinate indexes produces a much smoother effect.


ahocevar (commented):
Thanks, @mike-000! The change would be even nicer if the `route.json` file was GeoJSON.

mike-000:
It's the only example tagged Polyline so presumably the only example of using `ol/format/Polyline`? route.json uses the same structure that Mapbox and Openrouteservice directions APIs use to return a route.

ahocevar (approved):
Good point, I had missed that tag. Thanks, @mike-000.

PR #11620, opened 2020-10-03
Add charset to the examples html below the map

MoonE:
The vector-labels example uses the wrong encoding when opened with codesandbox. The '°' does not show correctly.


marcjansen (approved):

marcjansen:
Thanks

PR #11621, opened 2020-10-04
Do not draw empty text styles

### LINKED ISSUE #11619: Flickering points when decluttering points with empty text style
**Describe the bug**
Points with a fill / stroke and a text style with empty text flicker when the map is moved

**To Reproduce**
1. Go to https://codesandbox.io/s/declutter-points-with-empty-text-style-3z9ne?file=/main.js
2. Move the map
3. The points are only drawn some frames thus creating a flicker.
I suspect that the points may be drawn twice every other frame in this case. It can be seen sometimes when the map is moved around.

**Expected behavior**
An empty text for the text style should behave the same as no text style.

MoonE:
Fixes #11619

### DIFF
--- a/src/ol/renderer/vector.js
+++ b/src/ol/renderer/vector.js
@@ -75,7 +75,7 @@ function renderCircleGeometry(builderGroup, geometry, style, feature) {
     circleReplay.drawCircle(geometry, feature);
   }
   const textStyle = style.getText();
-  if (textStyle) {
+  if (textStyle && textStyle.getText()) {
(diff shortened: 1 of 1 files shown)


simonseyock (approved):
Every mention of the textStyle in `ol/renderer` is respected. In `ol/render/canvas/immediate` empty text styles already will not render.
For me this looks good.

ahocevar:
Thanks, @MoonE! I cherry-picked this into #11592, with the suggestions you made there.

PR #11628, opened 2020-10-06
VectorSource load events

### LINKED ISSUE #9975: Why feature sources do not contain loadstart and loadend events?
Tile and Image soruces contain "tileloadstart" and "tileloadend" events. But, feature sources do not containd any event to get "loadstart" and "loadend", so we need to customize the feature sources using ajax calls. Do you have a plan to add load events in feature sources?

simonseyock:
Resolves #9975

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -27,7 +27,7 @@ let withCredentials = false;
  *           import("./extent.js").Extent,
  *           number,
  *           import("./proj/Projection.js").default,
- *           function(): void=,
+ *           function(Array<import("./Feature.js").default>): void=,
  *           function(): void=): void} FeatureLoader
  * @api
  */
@@ -134,14 +134,14 @@ export function xhr(url, format) {
    * @param {import("./extent.js").Extent} extent Extent.
    * @param {number} resolution Resolution.
    * @param {import("./proj/Projection.js").default} projection Projection.
-   * @param {function(): void} success Success
+   * @param {function(): void=} success Success
    *      Function called when loading succeeded.
-   * @param {function(): void} failure Failure
+   * @param {function(): void=} failure Failure
    *      Function called when loading failed.
    * @this {import("./source/Vector").default}
    */
   return function (extent, resolution, projection, success, failure) {
-    const sourceOrTile = /** @type {import("./source/Vector").default} */ (this);
+    const source = /** @type {import("./source/Vector").default} */ (this);
     loadFeaturesXhr(
       url,
       format,
@@ -154,9 +154,12 @@ export function xhr(url, format) {
        * projection.
        */
       function (features, dataProjection) {
-        sourceOrTile.addFeatures(features);
+        if (success !== undefined) {
+          success(features);
+        }
+        source.addFeatures(features);
       },
-      /* FIXME handle error */ VOID
+      /* FIXME handle error */ failure ? failure : VOID
     );
   };
 }
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -16,15 +16,19 @@ let withCredentials = false;
  * load features.
  *
  * This function takes an {@link module:ol/extent~Extent} representing the area to be loaded,
- * a `{number}` representing the resolution (map units per pixel) and an
- * {@link module:ol/proj/Projection} for the projection  as
+ * a `{number}` representing the resolution (map units per pixel), an
+ * {@link module:ol/proj/Projection} for the projection and success and failure callbacks as
  * arguments. `this` within the function is bound to the
  * {@link module:ol/source/Vector} it's called from.
  *
  * The function is responsible for loading the features and adding them to the
  * source.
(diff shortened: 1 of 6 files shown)


simonseyock:
This ready for review

src/ol/source/VectorEventType.js
```diff
@@ -34,4 +34,25 @@ export default {
...
+  /**
+   * Triggered when features finishes loading.
+   * @event module:ol/source/Vector.VectorSourceEvent#featureloadend
+   * @api
+   */
+  FEATURELOADEND: 'featureloadend',
+
+  /**
+   * Triggered if feature loading results in an error.
+   * @event module:ol/source/TVector.VectorSourceEvent#featureloaderror
```
ahocevar, line 54:
```suggestion
   * @event module:ol/source/Vector.VectorSourceEvent#featureloaderror
```

src/ol/featureloader.js
```diff
@@ -130,25 +130,38 @@ export function loadFeaturesXhr(url, format, success, failure) {
...
+      /**
+       * @param {Array<import("./Feature.js").default>} features The loaded features.
+       * @param {import("./proj/Projection.js").default} dataProjection Data
+       * projection.
+       */
+      function (features, dataProjection) {
+        if (success !== undefined) {
+          success();
+        }
+        sourceOrTile.addFeatures(features);
```
ahocevar, line 160: You'd still have to check for availabilty of `addFeatures`, because vector tiles don't have that function. But I have a better idea: Now that you have a `success` callback, this function could be improved if it wouldn't deal with adding features at all any more. Then the tile or source would take care of adding features in the `success` callback. Should be a straightforward change, if you change the `success` callback to accept `features` as argument.

src/ol/source/VectorEventType.js
```diff
@@ -34,4 +34,25 @@ export default {
    * @api
    */
   REMOVEFEATURE: 'removefeature',
+
+  /**
+   * Triggered when features starts loading.
+   * @event module:ol/source/Vector.VectorSourceEvent#featureloadstart
+   * @api
+   */
+  FEATURELOADSTART: 'featureloadstart',
```
ahocevar, line 43: Maybe just a matter of taste, but I think the event names could be shorter: `loadstart`, `loadend`, `loaderror`.

ahocevar (changes requested):
Thanks for your work on this, @simonseyock! I added a few comments. Would be great if you could address them.

src/ol/featureloader.js
simonseyock: The xhr function is not actually used in the vector tiles. There exists a `defaultLoadFunction` which does this. That's why I proposed in https://github.com/openlayers/openlayers/issues/9975#issuecomment-704092256 that the functionality could be moved into a similar default function in the `source/Vector` file.

I agree that this could be done with a callback as well. I abstained from doing it, because it would break existing custom load functions. But if that's ok, I can do that.

src/ol/source/VectorEventType.js
simonseyock: I made it this way, because it is similar to the `tileloadstart` and `imageloadstart`. But I don't see why these have to be named differently in the first place. We could also introduce these events in `source/Source` if we implemented them in `source/UTFGrid` as well. They would not be fired in `source/Tile` and `source/Source` then, but those are abstract anyways.

simonseyock: I noticed that the other load events always contain the image or tile. This way the names make sense and they cannot be generalized that easily.

It could make sense to provide the loaded features in the `loadend` event for other than the default loading strategy.

src/ol/featureloader.js
ahocevar: Let's not break existing custom load function. Good point!

Thanks for the explanations! I had missed that the function is not used by the VectorTile source any more. I don't think it's necessary to do the refactoring you proposed in https://github.com/openlayers/openlayers/issues/9975#issuecomment-704092256. I think the only thing that's left to do here is rename the `sourceOrTile` variable to `source`, and change its type accordingly.

src/ol/source/VectorEventType.js
ahocevar: The reason why I started thinking about it was because we're loading `features`, not a `feature`. `featuresloadstart` would be a long event name, but it would be the correct name. And yes, it might make sense to provide the loaded features in the `featuresloadend` event.

simonseyock:
I added the changes. I introduced a new property `features` on the `VectorSourceEvent`. This looks a little ugly with the other property being `feature`, I think the proper way would be to remove `feature` and just provide an array with just one element (similiar to the modify event). But again this would be a breaking change. Or should I introduce a new event type `VectorSourceLoadedEvent`?

ahocevar (commented):
No need to do further renaming of event properties or make a breaking change. I think this looks good now. I only added two suggestions for the docs.

simonseyock:
done.

ahocevar (approved):
Thanks, @simonseyock

PR #11646, opened 2020-10-13
Ensure a unique tile key for each tile coordinate

### LINKED ISSUE #11626: TileImage/tileLoadFunction: (slow) sequentially tile loading / pause between tileloadend/start
Hello!

I'm not sure if this is a bug or i'm the first one who cares about that. :)

**Describe the bug**
When loading tiles with **tileLoadFunction**, there is a variable "pause" between two tiles (**tileloadend** and the next **tileloadstart**).
And because the tiles are loaded sequentially, it takes "too long" to render a layer.

I googled a lot and didn't find an answer or how i can change this behavior.

**To Reproduce**
See console output on https://jsfiddle.net/hdkpvrac (zoom in)
my browser: _Chrome/85.0.4183.123_

```
[3, 8, 9], 1601925605154, "tileloadstart"
[3, 8, 9], 1601925605154, "tileloadend"
"pause was (ms)", 41
[3, 8, 10], 1601925605195, "tileloadstart"
[3, 8, 10], 1601925605195, "tileloadend"
"pause was (ms)", 27
[3, 8, 11], 1601925605222, "tileloadstart"
[3, 8, 11], 1601925605223, "tileloadend"
"pause was (ms)", 23
[3, 8, 12], 1601925605246, "tileloadstart"
[3, 8, 12], 1601925605247, "tileloadend"
"rendercomplete, pause sum (ms)", 710
```

Sure, you could say that's not that much, but if you have multiple layers with many tiles, it adds up to a non user-friendly behavior.

**Expected behavior**
- Faster tile loading. :)
- There should'nt be a pause between two tiles (?)
- async/concurrent **tileloadstart**

Thanks for your time. :)

ahocevar:
Instead of providing a `getKey()` method for each subclass of `ol/Tile`, we can use the one from `ol/Tile` if we set the `key` on the instance.

In particular, this pull request fixes issues with custom `tileLoadFunction`s in applications, in which case there is no need to configure the tile source with a unique `url` for each tile coordinate. The `ol/Tile` `getKey()` function ensures that the tile coordinate is part of the key, to make it unique. The unique key is needed for the tile queue.

Fixes #11626

### DIFF
--- a/src/ol/ImageTile.js
+++ b/src/ol/ImageTile.js
@@ -39,6 +39,8 @@ class ImageTile extends Tile {
      */
     this.src_ = src;
 
+    this.key = src;
+
     /**
      * @private
      * @type {HTMLImageElement|HTMLCanvasElement}
@@ -68,13 +70,6 @@ class ImageTile extends Tile {
    */
   getImage() {
     return this.image_;
-  }
-
-  /**
-   * @return {string} Key.
-   */
-  getKey() {
-    return this.src_;
   }
 
   /**
--- a/src/ol/VectorTile.js
+++ b/src/ol/VectorTile.js
@@ -63,6 +63,8 @@ class VectorTile extends Tile {
      * @type {string}
      */
     this.url_ = src;
+
+    this.key = src;
   }
 
   /**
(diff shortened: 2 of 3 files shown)


tschaub (approved):

tschaub:
Thanks for the additional explanation (wasn’t obvious looking at the diff alone).

Issue #5458, opened 2016-06-10
DWithin filter


bartvanandel:
I know support for OGC filtering is still experimental, but are there any plans to include the DWithin filter? Example filter:

``` xml
<ogc:DWithin>
    <ogc:PropertyName>Location</ogc:PropertyName>
    <gml:Point srsName="EPSG:4326" xmlns:gml="http://www.opengis.net/gml">
        <gml:pos>123.45 54.321</gml:pos>
    </gml:Point>
    <ogc:Distance units="http://www.opengeospatial.org/se/units/metre">15</ogc:Distance>
</ogc:DWithin>
```

My first (and only) attempt for now was to patch it in externally to a version loaded from npm. Besides being a very ugly hack, I'm probably using the wrong methods, but I did not have much time to dive deep.

``` js
import ol from 'openlayers';

// Attempt to patch DWithin filter into loaded npm package
const DWithin = function(propertyName, point, opt_srsName, distance, opt_units) {
    goog.base(this, 'DWithin');
    this.propertyName = propertyName;
    this.point = point;
    this.srsName = opt_srsName;
    this.distance = distance;
    this.units = opt_units || 'http://www.opengeospatial.org/se/units/metre';
};
goog.inherits(DWithin, ol.format.ogc.filter.Filter);

ol.format.WFS.GETFEATURE_SERIALIZERS_['http://www.opengis.net/ogc']['DWithin'] =
    ol.xml.makeChildAppender(function(node, filter, objectStack) {
        ol.format.WFS.writeOgcPropertyName_(node, filter.propertyName);
        ol.format.GML3.prototype.writeGeometryElement(node, filter.point, objectStack);

        const distanceNode = ol.xml.createElementNS('http://www.opengis.net/ogc', 'Distance');
        distanceNode.setAttribute('units', filter.units);
        ol.format.XSD.writeStringTextNode(distanceNode, filter.distance+'');
        node.appendChild(distanceNode);
    });
```

Anyone with more experience on the `openlayers` internals?

NB: at this point it would actually be fine if we could just insert a raw piece of XML as one of the filters, so we can combine custom, (still) unsupported filter types with existing filters.

tsauerwein:
> NB: at this point it would actually be fine if we could just insert a raw piece of XML as one of the filters, so we can combine custom, (still) unsupported filter types with existing filters.

`ol.format.WFS()#writeGetFeature()` returns a XML node which you could modify. Something like this:

```js
[...]

PR #11648, opened 2020-10-14
Dwithin filter

jbo023:
- added support for distance within filter
- fixed wrong filter encoding namespace handling for WFS 2.0 requests.

this was also requested some years ago:  https://github.com/openlayers/openlayers/issues/5458

I also removed some duplicated code and combined the writeContainsFilter, writeDisjointFilter, writeIntersectsFilter and writeWithinFilter functions in a new writeSpatialFilter function. 

```
<fes:Filter
   xmlns:fes="http://www.opengis.net/fes/2.0"
   xmlns:gml="http://www.opengis.net/gml/3.2"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://www.opengis.net/fes/2.0
   http://schemas.opengis.net/filter/2.0/filterAll.xsd
   http://www.opengis.net/gml/3.2
   http://schemas.opengis.net/gml/3.2.1/gml.xsd">
   <fes:DWithin>
      <fes:ValueReference>geometry</fes:ValueReference>
      <gml:Point gml:id="P1" srsName="http://www.opengis.net/def/crs/epsg/0/4326">
         <gml:pos>43.716589 -79.340686</gml:pos>
      </gml:Point>
      <fes:Distance uom="m">10</fes:Distance>
   </fes:DWithin>
</fes:Filter>
```

### DIFF
--- a/test/spec/ol/format/wfs.test.js
+++ b/test/spec/ol/format/wfs.test.js
@@ -687,7 +687,7 @@ describe('ol.format.WFS', function () {
       expect(serialized.firstElementChild).to.xmleql(parse(text));
     });
 
-    it('creates a dwithin filter for WFS 1.x', function () {
+    it('creates a dwithin filter', function () {
       const text =
         '<wfs:Query xmlns:wfs="http://www.opengis.net/wfs" ' +
         '    typeName="area" srsName="EPSG:4326" ' +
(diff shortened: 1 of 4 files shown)


test/spec/ol/format/wfs.test.js
```diff
@@ -686,13 +687,55 @@ describe('ol.format.WFS', function () {
...
+    });
+
     it('creates During property filter', function () {
       const text =
         '<wfs:Query xmlns:wfs="http://www.opengis.net/wfs" ' +
         '    typeName="states" srsName="EPSG:4326">' +
         '  <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">' +
         '    <ogc:During>' +
-        '      <fes:ValueReference xmlns:fes="http://www.opengis.net/fes">date_prop</fes:ValueReference>' +
+        '      <ogc:PropertyName>date_prop</ogc:PropertyName>' +
```
ahocevar, line 738: Ideally, you'd modify this test case to use WFS 2.0 (and continue using ValueReference), and add another using WFS 1.x (with PropertyName).

ahocevar (commented):
Thanks for your work on this @jbo023! I made one comment about the tests. Would be great if you could address that. Looks great otherwise!

test/spec/ol/format/wfs.test.js
jbo023: Thanks @ahocevar  for reviewing, 

The during test for wfs 2.0 is already there.

The during implementation was a bit weird the filter is only specified in the 2.0.0 filter encoding. But the filter was already there for the wfs 1.x format but with a mix of the ogc and fes namespace. 

To handle this there are three options
- remove during from wfs 1.0
- change wfs 1.0 during implementation to use ogc:propertyName for consistency ( This is already done)
- keep wfs 1.0 during implementation as it was with using fes:valueReference instead of ogc:propertyName for backwards compatibility

What implementation do you prefer also see: https://github.com/openlayers/openlayers/pull/6786

ahocevar: Thanks for the explanations @jbo023. I think users might be relying on the existing implementation, so if it doesn't mean adding weird code, I'd opt for keeping the WFS 1.x `During` implementation as it was for backwards compatibility.

jbo023: @ahocevar ok, i changed the behaviour so that the during filter with wfs 1.0 keeps the old behaviour

ahocevar (approved):
Thanks, @jbo023!

PR #11657, opened 2020-10-17
Use https to avoid security warnings

mike-000:
The WMTS Tile Transitions example causes not secure and mixed content browser warnings which are avoidable as SCALGO supports `https`.  Also use `target="_blank"` to avoid closing map when clicking on attributions.


marcjansen (approved):
Thanks.

PR #11668, opened 2020-10-19
Don't try to get image data for zero sized images

### LINKED ISSUE #11656: Exception with raster source when zoomed out
**Describe the bug**
When zooming out and the layer gets so small that it's not visible, the following exception is thrown:
Uncaught DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The source width is 0.
    at getImageData
    at RasterSource.push../node_modules/ol/source/Raster.js.RasterSource.processSources_
    at RasterSource.push../node_modules/ol/source/Raster.js.RasterSource.getImage
    at CanvasImageLayerRenderer.push../node_modules/ol/renderer/canvas/ImageLayer.js.CanvasImageLayerRenderer.prepareFrame
    at ImageLayer.push../node_modules/ol/layer/Layer.js.Layer.render
    at CompositeMapRenderer.push../node_modules/ol/renderer/Composite.js.CompositeMapRenderer.renderFrame
    at Map.push../node_modules/ol/PluggableMap.js.PluggableMap.renderFrame_
    at Map.<anonymous>

This happens with ol/layer/Image and ol/source/Raster, but not with ol/layer/Tile and ol/source/TileWMS.

Maybe Raster#getImageData should return null if width or height is zero?

**To Reproduce**
Sorry, no demo.

**Expected behavior**
Nothing is shown on map and there is no exception.

tapioko:
Fixes #11656

### DIFF
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -832,6 +832,9 @@ function getImageData(layer, frameState) {
   }
   const width = frameState.size[0];
   const height = frameState.size[1];
+  if (width === 0 || height === 0) {
+    return null;
+  }
   const container = renderer.renderFrame(frameState, null);
   let element;
   if (container) {


ahocevar (approved):
Thanks, @tapioko

PR #11673, opened 2020-10-20
Do not emulate dblclick and slingleclick on multi-touch

### LINKED ISSUE #11670: DoubleClickZoom also fires when using multitouch
I have a simple map with the following interaction added:

```javascript
map.addInteraction(new ol.interaction.DoubleClickZoom({delta:1}));
```

But when testing on iOS, I noticed that the map also zooms in when I simply place two fingers on the map. I think behind the scenes it registers two 'presses' (one of each finger), and therefore fires this event.

Is there anything I can do to prevent this from happening?

ahocevar:
Fixes #11670.

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -32,6 +32,12 @@ class MapBrowserEventHandler extends EventTarget {
     this.clickTimeoutId_;
 
     /**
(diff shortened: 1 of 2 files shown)


ahocevar:
I added another commit with improved tests that revealed the same problem with emulated `singleclick` events, which I also fixed now.

This is ready for review.

fredj (approved):

Issue #11672, opened 2020-10-20
Rendering of reprojected VectorImage layers seems broken from v6.3.1 -> v6.4.3


geometix:
**Describe the bug**
Rendering of Vector sources on a VectorImage layer seems to be broken in v6.4.3, when the vector source has EPSG:4326 projection covering the whole globe, and the map is in EPSG:3857. It seems that the reprojection goes wrong, as EPSG:4326 exceeds the extent of EPSG:3857 (at high latitudes). It results in several layers of false polygons that are drawn on top of the correct ones, sometimes covering the whole map.

**To Reproduce**
1.) create a Vector source from a GeoJSON file containing polygons in EPSG:4326 covering the whole world (-180 to 180W and -90 to 90N).
2.) create a VectorImage layer with the Vector source
3.) add that layer to a map that is created with a view with projection EPSG:3857
4.) zoom and pan around and see flickering false polygons covering the map

**Expected behavior**
Not seeing any false polygons, like it is the case if you did add the same layer to a map with a view in EPSG:4326 (not reprojecting the source).

geometix:
It seems that most of the country polygons in the example are within the extent of EPSG:3857. I think the problem becomes more evident if a lot of features are outside of EPSG:3857. If you pan the map in the example to the southern boundary, you can at least see that the polygon of Antarctica is self-intersecting near the edge of the extent. I don't know whether this has to do with the problem. I will try to reproduce the bug either as you suggested in the example or with another minimalistic setup tomorrow.

mike-000:
There is a data issue with Antarctica.  To display correctly polygons should be split at the dateline as they are in Russia and Fiji.  That does not happen in Antarctica resulting in a line across the world `[180,-84.71338],[-179.942499,-84.721443]` just above the base of the projection extent.

geometix:
I have put up a most minimalistic setup where the problem can easily be seen and compared between three different scenarios. I did: setup a map with two layers, an OSM tile layer as background and a weather data layer on top, derived from a geojson file coded in EPSG:4326, covering the whole world from -90 to +90N. The polygons are drawn semi-transparent, as the effect is even more evident in these cases.

1.) map view in EPSG:3856 and ol v6.3.1 -> everything is fine. polygons are drawn only once and background ist clearly visible, no random color changes when zooming or panning as expected.
[...]

PR #11676, opened 2020-10-21
Fix countries.geojson Antarctica polygon so it wraps correctly

mike-000:
As observed in https://github.com/openlayers/openlayers/issues/11672#issuecomment-713182724 there is a problem with the Antarctica polygon used in the examples in global projections wrapped at the dateline.

Replacing `[180,-84.71338],[-179.942499,-84.721443]`
with `[180,-84.71338],[180,-90],[-180,-90],[-180,-84.71338],[-179.942499,-84.721443]`
gives the correct polygon in these projections.

`countries-110m.geojson` already has a similar setup.


ahocevar (approved):
Thanks, @mike-000.

PR #11693, opened 2020-10-26
Use W3C Modifier Keys Attributes value to detect ctrl key on Firefox

### LINKED ISSUE #11692: Firefox issue with evt.ctrlKey on keypress events
**Describe the bug**
On Firefox: On listen keydown/keyup, sometimes the detection of `event.ctrlKey` fail. (Work everytime on Chrome).
As `ol.events.conditions.js` rely on the detection of `event.ctrlKey` a bug may append.

Update: I've opened a corresponding issue on bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1673362

**To Reproduce**
I've made a codepen example: https://codepen.io/ger-benjamin/pen/ZEOeyRb (without any library). Open it with Firefox and press repeatedly the ctrl key (it append ~10% of the time on my machine.).

**Expected behavior**
I propose to use the Use W3C Modifier Keys Attributes value to detect ctrl key on Firefox.
I'll propose a PR => #11693

ger-benjamin:
Fix #11692

We may use the key attribute value also for the `alt key` and the `shift key` to be coherent.

### DIFF
--- a/src/ol/events/condition.js
+++ b/src/ol/events/condition.js
@@ -37,6 +37,18 @@ export function all(var_args) {
 }
 
 /**
+ * Return `true` if the platform-modifier-key (the meta-key on Mac, ctrl-key otherwise) is
+ * pressed, `false` otherwise.
+ * @param {import("../MapBrowserEvent.js").default} mapBrowserEvent Map browser event.
+ * @return {boolean} True if the platform modifier key is pressed.
+ */
+export const isPlatformModifierKey = function (mapBrowserEvent) {
+  const originalEvent = /** @type {KeyboardEvent} */ (mapBrowserEvent.originalEvent);
+  // Do not use key.ctrl (or key.meta on Mac) Because Firefox sometimes doesn't assign the ctrlKey.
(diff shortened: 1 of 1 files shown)


ahocevar (changes requested):
> We may use the key attribute value also for the alt key and the shift key to be coherent.

Yes, we should be consistent and use the same API for all modifier keys.

ger-benjamin:
Thanks @ahocevar .
The issue is updated on bugzilla. It looks that this issue append only on Ubuntu (Ubuntu 18 ?) on Firefox.
I think it's still a good idea to use `originalEvent.key`. What's your opinion. Should I do the change ?

ger-benjamin:
Mhh, I've tried to update the code.
But I realized that the key string always only contains one key. That means, if you press ctrl and alt at the same time, there is still two event, one with key `ctrl`, one with key `alt`. The first have only one flag to true, the second have both flag activated.

So, I can't use this workaround. I close and hope they fix the bug in Firefox for Ubuntu.

Issue #11084, opened 2020-05-23
No doc for WebGLPoints, is it deprecated ?


FractalWire:
I've noticed the [WebGL points layer example](https://openlayers.org/en/latest/examples/webgl-points-layer.html)
I wanted to dig a bit more into this and check out the [api doc](https://openlayers.org/en/latest/apidoc/index.html).
However, as far as I can tell, there is nothing about this layer in the doc.

So is it deprecated or is the doc lagging behind ?

jahow:
Some experimental features are currently not part of the API doc. [This PR](https://github.com/openlayers/openlayers/pull/10544) by @MoonE was fixing that, but it was postponed because of a project-wide formatting change (prettier). I guess it would be good to revisit this PR!

FractalWire:
That would be a good idea I think.

Another question.
Is WebGL still considered an experimental feature in Openlayers at this point ?

I can see WebGLLayerRenderer is part of the API but there are no examples using it and it's hard to figure out how to use it.

jahow:
> Is WebGL still considered an experimental feature in Openlayers at this point ?

Yes. There are a few examples marked as experimental which you can use as reference:
https://openlayers.org/en/latest/examples/?q=webgl

PR #11702, opened 2020-10-31
Add @api tag to layer/WebGLPoints

ustiloff:
To include this class in legacy build and documentation


MoonE:
The webgl points layer  in its current state is considered experimantal and not yet part of the api and therefore also not included in the legacy build.
https://github.com/openlayers/openlayers/issues/11084
This PR was aiming at adding it to the api, but I have been occupied with other things in the meantime.
https://github.com/openlayers/openlayers/pull/10544

PR #11704, opened 2020-10-31
Apidoc experimental tag

MoonE:
Replaces / closes #11702, #10544
@jahow Please have a look if the correct functions / classes are tagged or any are missing.

### DIFF
--- a/config/jsdoc/api/plugins/api.js
+++ b/config/jsdoc/api/plugins/api.js
@@ -4,12 +4,11 @@
  */
 exports.defineTags = function (dictionary) {
   dictionary.defineTag('api', {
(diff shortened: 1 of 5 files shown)


ahocevar:
I'm still not convinced we want to re-introduce api stability levels, since we removed them for good reason.

An alternative would be to provide a legacy build that exposes the whole library, not just symbols marked `@api`. Something like `ol-full.js`.

ahocevar:
@MoonE If you agree, I'd like to see #11715 as a replacement for this. #11715 does not address the part of the problem that experimental features are not documented. But I'd see that as motivation to improve the experimental features so they don't need to be called "experimental" any more.

PR #11716, opened 2020-11-04
Treat svg as binary resource

mike-000:
The large svg in the SVG Layer example, when embedded as text, is causing an error when attempting to open the example in CodeSandbox.  It (and any other svg data) can however be successfully imported into CodeSandbox in the same way as binary image files.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #11719, opened 2020-11-05
Clamp EPSG:3857 y to lowest/highest possible value instead of validity extent

### LINKED ISSUE #11672: Rendering of reprojected VectorImage layers seems broken from v6.3.1 -> v6.4.3

ahocevar:
Although the validity extent of EPSG:3857 in y direction only ranges from ~-85 to ~85 degrees, lower/higher values can be handled. To avoid geometries getting corrupted topologically when transforming values from high/low latitudes, we can clamp to the lowest/highest safe value instead of the validity extent.

Improves (but does not fix) #11672.

### DIFF
--- a/src/ol/proj/epsg3857.js
+++ b/src/ol/proj/epsg3857.js
@@ -36,7 +36,7 @@ export const WORLD_EXTENT = [-180, -85, 180, 85];
  * @const
  * @type {number}
  */
-export const MAX_SAFE_Y = RADIUS * Math.log(Math.tan((Math.PI * 180) / 360));
+export const MAX_SAFE_Y = RADIUS * Math.log(Math.tan(Math.PI / 2));
 
 /**
  * @classdesc
--- a/src/ol/proj/epsg3857.js
+++ b/src/ol/proj/epsg3857.js
@@ -32,6 +32,13 @@ export const EXTENT = [-HALF_SIZE, -HALF_SIZE, HALF_SIZE, HALF_SIZE];
 export const WORLD_EXTENT = [-180, -85, 180, 85];
 
 /**
+ * Maximum safe value in y direction
+ * @const
+ * @type {number}
+ */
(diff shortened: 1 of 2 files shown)


src/ol/proj/epsg3857.js
```diff
@@ -31,6 +31,13 @@ export const EXTENT = [-HALF_SIZE, -HALF_SIZE, HALF_SIZE, HALF_SIZE];
  */
 export const WORLD_EXTENT = [-180, -85, 180, 85];
 
+/**
+ * Maximum safe value in y direction
+ * @const
+ * @type {number}
+ */
+export const MAX_SAFE_Y = RADIUS * Math.log(Math.tan((Math.PI * 180) / 360));
```
MoonE, line 39: This looks a bit weird calculating with radians.
```suggestion
export const MAX_SAFE_Y = RADIUS * Math.log(Math.tan(Math.PI / 2));
```

tschaub (approved):

Issue #11714, opened 2020-11-03
KML with empty z coordinates results in features with null geometry


mike-000:
**Describe the bug**
ol/format/KML does not produce a geometry when placemark coordinates are in the format `<coordinates>6.28240681,49.50119400,</coordinates>`  A workaround is to load the file and replace each `,</coordinates>` by `</coordinates>` or `,0</coordinates>` but that should not be necessary as they are handled correctly by Google Earth.

Additionally ol/format/KML produces the default style instead of  #wronline when defined as 
```
<StyleMap id="wronline"><Pair><key>normal</key><styleUrl>n_wronline</styleUrl></Pair><Pair><key>highlight</key><styleUrl>h_wronline</styleUrl></Pair></StyleMap>
```
while a workaround to insert # 
```
<StyleMap id="wronline"><Pair><key>normal</key><styleUrl>#n_wronline</styleUrl></Pair><Pair><key>highlight</key><styleUrl>#h_wronline</styleUrl></Pair></StyleMap>
```
does work.

`**To Reproduce**
Use this example
```
<?xml version="1.0"?>
[...]

PR #11721, opened 2020-11-06
Handle empty Z coordinates in KML

mike-000:
Fixes the null geometry problem in #11714 (the styleUrl problem needs a separate fix)
Add tests for reading XY and empty Z coordinates


ahocevar (approved):
Thanks, @mike-000

PR #11722, opened 2020-11-06
Correct meaning of 'start' and 'end' text align for LTR text

### LINKED ISSUE #11586: Text alignment wrong for start and end

ahocevar:
The Text style's `textAlignment` option takes - among others - `start` and `end` as values. Previously, these values only made sense for right-to-left (RTL) text. With this change, `start` means start of the text for both left-to-right (LTR) and RTL text, and `end` means end.

Replaces #11588
Fixes #11586


src/ol/render/canvas/Executor.js
```diff
@@ -75,6 +75,20 @@ function getDeclutterBox(replayImageOrLabelArgs) {
   return replayImageOrLabelArgs[3].declutterBox;
 }
 
+const rtlRegEx = /[\u0591-\u07FF]/;
```
MoonE, line 78:
```suggestion
const rtlRegEx = new RegExp(
  /* eslint-disable prettier/prettier */
  '[' +
    String.fromCharCode(0x00591) + '-' + String.fromCharCode(0x008ff) +
    String.fromCharCode(0x0fb1d) + '-' + String.fromCharCode(0x0fdff) +
    String.fromCharCode(0x0fe70) + '-' + String.fromCharCode(0x0fefc) +
    String.fromCharCode(0x10800) + '-' + String.fromCharCode(0x10fff) +
    String.fromCharCode(0x1e800) + '-' + String.fromCharCode(0x1efff) +
  ']'
  /* eslint-enable prettier/prettier */
);
```
Apparently it is  not possible to use 5-digit unicode escape codes in js ...

The data I used to look it up is this:
https://github.com/openlayers/openlayers/blob/e7a23d619adc41f8ebe7b57ac5e42fe528533c1c/unicode14.ods?raw=true
sourced from: https://www.unicode.org/Public/14.0.0/ucd/UnicodeData-14.0.0d3.txt

RTL characters are of type `AL` or `R` (orange in the spreadsheet),  LTR characters are of type `L` (blue in the spreadsheet), see: https://www.unicode.org/reports/tr9/#Table_Bidirectional_Character_Types

rendering/cases/rtl-text-align/main.js
```diff
@@ -0,0 +1,192 @@
...
+import View from '../../../src/ol/View.js';
+
+const vectorSource = new VectorSource();
+let feature;
+
+// Latin - end (right)
+feature = new Feature({
+  geometry: new Point([-10, 50]),
+});
+feature.setStyle(
```
MoonE, line 20: Setting a style function on the `VectorSource` and the `text` / `textAlign` on the `Feature`s might make it easier to see what's being tested here.

src/ol/render/canvas/Executor.js
ahocevar: Thanks for the thorough research on this!

rendering/cases/rtl-text-align/main.js
ahocevar: Good suggestion! Implemented with 1db2c4bc45.

rendering/cases/rtl-text-align/main.js
```diff
@@ -0,0 +1,95 @@
...
+      geometry: new Point([10, 0]),
+      text: 'עִברִית',
+      textAlign: 'end',
+    }),
+
+    // Arabic - end (left)
+    new Feature({
+      geometry: new Point([10, -50]),
+      text: 'عربى',
+      font: '24px Ubuntu',
```
MoonE, line 54: This line can be removed.

MoonE (approved):
Thanks @ahocevar looks good to me now.

PR #11723, opened 2020-11-06
Drag and Drop Interaction support for formats that read ArrayBuffer sources

### LINKED ISSUE #11301: Drag and Drop Interaction passing file content as ArrayBuffer to support custom formats
Allow custom formats requiring ArrayBuffer (e.g. KMZ, or zipped SHP + DBF) to be integrated in the sequence of formats for the Drag and Drop interaction.

This would require the interaction (either optionally or by default) to pass an ArrayBuffer to the readFeatures methods, and for formats which support String to check for ArrayBuffer and decode it.

mike-000:
Fixes #11301

Adds support for formats that read ArrayBuffer sources in the Drag and Drop Interaction.
Also accept constructed formats as well as constructors allowing constructor options to be used.

Add Custom Drag and Drop examples with sample data including a KMZ with custom icons and MVT preview.
Add sample data and KML extractStyles option to original example.


examples/drag-and-drop-custom.html
```diff
@@ -0,0 +1,32 @@
+---
+layout: example.html
+title: Custom Drag-and-Drop
+shortdesc: Example of using the drag-and-drop interaction with custom formats.
+docs: >
+  Example of using the drag-and-drop interaction with custom formats.
+  In addition to the formats used in the [Drag-and-Drop](drag-and-drop.html) example a custom format (subclassing KML) is used to handle KMZ files, the standard KML format is pre-constructed to not extract styles, and individual MVT tiles can be previewed.
+  There is no projection transform support, so this will only work with data in EPSG:4326 and EPSG:3857.
```
tschaub, line 8: There is a lot of nice stuff in this example.  Rather than one example that does three things (read KMZ, read KML without styles, view MVT tiles), I think it would be simpler to have three examples that demonstrate one thing each.

This would avoid users being confused about something like the MVT tile coordinate inputs when viewing KML, for example.

tschaub (approved):

tschaub:
Very nice, @mike-000.

PR #11724, opened 2020-11-07
Add transition option to OSM and CartoDB sources, document default

MoonE:
Add transition option to OSM and CartoDB sources, and document the default transition of 250 ms for those sources using the TileImage class exclusively. That is XYZ and all its sub classes except TileDebug.


tschaub (approved):

tschaub:
Thanks, @MoonE

PR #11732, opened 2020-11-10
Replace readURI with readStyleURL for KML StyleURLs

mike-000:
Fixes incorrect style from pairs in #11714

New function `readStyleURL` ensures StyleURL begins with # if it does not contain one.


ahocevar (approved):
Thanks, @mike-000!

PR #11737, opened 2020-11-12
Write the correct SRS code in EsriJSON

### LINKED ISSUE #11734: EsriJson writes wrong wkid?
**Describe the bug**

The EsriJSON feature format produces a geometry in EPSG:4326, but with a wkid of 3857.

I'm not confident that I understand the intended behaviour, but this seems wrong. By setting `dataProjection` to `EPSG:4326` and `featureProjection` to `EPSG:3857`, it converts features from the default projection to EPSG:4326 but with the wrong wkid.

**To Reproduce**

```
draw = new ol.interaction.Draw({
    source: new ol.source.Vector({wrapX: false}),
    type: 'Polygon',
});
map.addInteraction(draw);
draw.on('drawend', e => {
    const geometry = new ol.format.EsriJSON()
        .writeFeatureObject(e.feature, {
        dataProjection: 'EPSG:4326',
        featureProjection: 'EPSG:3857'
    }).geometry;
    console.log(geometry.spatialReference.wkid); //  3857
});
```

**Expected behavior**

Wkid of 4326.

ahocevar:
The previous behavior was incorrect, because the `featureProjection` ist the projection the features are in, whereas `dataProjection` is the projection of the serialized EsriJSON. So the correct behavior is to get the `wkid` from the `dataProjection` if it is set, and fall back to the `featureProjection`, in which case the features will not be transformed. 

Fixes #11734.

### DIFF
--- a/examples/vector-esri-edit.js
+++ b/examples/vector-esri-edit.js
@@ -15,9 +15,8 @@ import {fromLonLat} from '../src/ol/proj.js';
 import {tile as tileStrategy} from '../src/ol/loadingstrategy.js';
 
 const serviceUrl =
-  'https://services.arcgis.com/rOo16HdIMeOBI4Mb/arcgis/rest/' +
-  'services/PDX_Pedestrian_Districts/FeatureServer/';
-const layer = '0';
+  'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire/FeatureServer/';
+const layer = '2';
 
 const esrijsonFormat = new EsriJSON();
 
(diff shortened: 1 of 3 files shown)


ahocevar:
I added another test to make things clearer. This is now ready for review.

DaniEll-AT (approved):

ahocevar:
While working on this fix, I noticed that the `vector-esri-edit.html` example uses a service that is no longer available without token, so I updated the example to use a different service that works.

tschaub (approved):

PR #11739, opened 2020-11-12
Fix: Check for layer extent in CanvasLayerRenderer.getDataAtPixel 

### LINKED ISSUE #10099: map.forEachLayerAtPixel still return image layer even though clicked outside polygon
I am having an issue using ol version 6.0.1, as I clicked outside imagewms polygon, its still catch that layer and when multiple layers were visibled, it will detect the top layer even if I click outside the top layer's feature. I already set transparent to true. This is not happening when I use 5.3.3 version. Please kindly help me!

boeckMt:
## Current behavior

If the [CanvasLayerRenderer](https://github.com/openlayers/openlayers/blob/v6.4.3/src/ol/renderer/canvas/Layer.js#L301) gets cross-origin image data, it returns an empty array so `map.forEachLayerAtPixel` will detect a layer  even if it is clicked outside of the layer extent.

Fixes #10099

## New behavior
The `CanvasLayerRenderer` checks for each pixel if it is in the layer extent, when not it will return `null` in `getDataAtPixel`.

### DIFF
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -9,13 +9,14 @@ import {
   compose as composeTransform,
   create as createTransform,
 } from '../../transform.js';
-import {createCanvasContext2D} from '../../dom.js';
 import {
+  containsCoordinate,
   getBottomLeft,
   getBottomRight,
   getTopLeft,
   getTopRight,
 } from '../../extent.js';
+import {createCanvasContext2D} from '../../dom.js';
 import {rotateAtOffset} from '../../render/canvas.js';
 
 /**
@@ -283,6 +284,20 @@ class CanvasLayerRenderer extends LayerRenderer {
       pixel.slice()
     );
     const context = this.context;
+
+    const layer = this.getLayer();
+    const layerExtent = layer.getExtent();
+    if (layerExtent) {
+      const renderCoordinate = applyTransform(
+        frameState.pixelToCoordinateTransform,
+        pixel.slice()
+      );
+
+      /** get only data inside of the layer extent */
+      if (!containsCoordinate(layerExtent, renderCoordinate)) {
+        return null;
+      }
+    }
 
     let data;
     try {
--- a/test/spec/ol/renderer/canvas/imagelayer.test.js
+++ b/test/spec/ol/renderer/canvas/imagelayer.test.js
@@ -66,6 +66,189 @@ describe('ol.renderer.canvas.ImageLayer', function () {
     });
   });
 
+  describe('#forEachLayerAtPixel Image CORS', function () {
+    let map,
+      target,
+      imageExtent,
+      projection,
+      sourceCross,
+      source,
+      imageLayer,
+      imageLayerCross;
+    beforeEach(function (done) {
+      projection = new Projection({
+        code: 'custom-image',
+        units: 'pixels',
+        extent: [0, 0, 200, 200],
+      });
+      target = document.createElement('div');
+      target.style.width = '100px';
+      target.style.height = '100px';
+      document.body.appendChild(target);
+      imageExtent = [0, 0, 20, 20];
+      source = new Static({
+        url: `https://openlayers.org/assets/theme/img/logo70.png`,
+        projection: projection,
+        imageExtent: imageExtent,
+      });
+      imageLayer = new ImageLayer({
+        source: source,
+      });
+      sourceCross = new Static({
+        url: `https://openlayers.org/assets/theme/img/logo70.png`,
+        projection: projection,
+        imageExtent: imageExtent,
+        crossOrigin: 'anonymous',
+      });
+      imageLayerCross = new ImageLayer({
+        source: sourceCross,
+      });
+      map = new Map({
+        pixelRatio: 1,
+        target: target,
+        layers: [imageLayer, imageLayerCross],
+        view: new View({
+          projection: projection,
+          center: [10, 10],
+          zoom: 1,
+          maxZoom: 8,
+        }),
+      });
+      let loadedCount = 0;
+      [source, sourceCross].forEach(function (source) {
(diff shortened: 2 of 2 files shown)


boeckMt:
@ahocevar I will further add a test for this fix, but currently it looks like the async `beforeEach` in [ol.renderer.canvas.ImageLayer#forEachLayerAtCoordinate](https://github.com/openlayers/openlayers/blob/main/test/spec/ol/renderer/canvas/imagelayer.test.js#L46)  is not called so the test case fails. Maybe I have to check if my local development environment is set up correctly.

src/ol/renderer/canvas/Layer.js
```diff
@@ -284,6 +285,20 @@ class CanvasLayerRenderer extends LayerRenderer {
...
+    const layer = this.getLayer();
+    const layerExtent = layer.getExtent();
+    if (layerExtent) {
+      const renderCoordinate = applyTransform(
+        frameState.pixelToCoordinateTransform,
+        pixel.slice()
+      );
+
+      /** get only data inside of the layer extent */
+      if (!containsXY(layerExtent, renderCoordinate[0], renderCoordinate[1])) {
```
ahocevar, line 297: You could also use `containsCoordinate(layerExtent, renderCoordinate)` here.

boeckMt: @ahocevar thanks for the hint, have missed that this function is available.

I also have created a temporal branch on my repo with an [example](https://github.com/boeckMt/openlayers/blob/canvas-layer-renderer-test/examples/hitdetect-image.js) so it is easier to discuss the Issue.

What is supposed to happen when there are layers without extent and one layer runs in the catch block of  [getDataAtPixel](https://github.com/boeckMt/openlayers/blob/canvas-layer-renderer-test/src/ol/renderer/canvas/Layer.js#L318)?
I think then it is not possible to detect this layer, because there is no extent and  `context.getImageData` is forbidden.

Would it then be fine to do the following check and don't return directly from the catch block.
```js
...
} catch (err) {
      if (err.name === 'SecurityError') {
        // tainted canvas, we assume there is data at the given pixel (although there might not be)
        // return new Uint8Array(); // do not return here
        data = new Uint8Array();
      }
      // return data;  // if there is an error getting the getImageData, pass it further to the next check
    }

    if (data[3] === 0 || !data.length) {
      return null;
    }
return data;
...
```
Or should `getDataAtPixel` return the empty array and then the check for data length should be happen in `forEachLayerAtPixel` like you suggested at the beginning.

> if (data && data.length) {
> https://github.com/openlayers/openlayers/issues/10099#issuecomment-725463303

ahocevar: I don't think we need to change anything in addition. If we have a layer with a tainted canvas, our best guess (when this fix is merged) is to assume there is data when we hit the layer inside its extent. When there is no extent, we assume that there is data everywhere on the layer, like the inline comment says.

boeckMt: That's fine, I further added two tests for this fix. Hope the usage of the openlayers Image https://openlayers.org/assets/theme/img/logo70.png for testing cross-data is ok.

test/spec/ol/renderer/canvas/imagelayer.test.js
```diff
@@ -79,81 +86,74 @@ describe('ol.renderer.canvas.ImageLayer', function () {
...
-        layers: [],
+        layers: [imageLayer, imageLayerCross],
         view: new View({
           projection: projection,
           center: [10, 10],
           zoom: 1,
           maxZoom: 8,
         }),
       });
+      source.on('imageloadend', function () {
```
ahocevar, line 119: You have to wait until both `source` and `sourceCross` are loaded, e.g.
```js
let loadedCount = 0;
[source, sourceCross].forEach(function (source) {
  source.once('imageloadend', function () {
    loadedCount++;
    if (loadedCount === 2) {
      done();
    }
  });
});
```

ahocevar (commented):
I added a comment which might help you fix the remaining test problem.

test/spec/ol/renderer/canvas/imagelayer.test.js
boeckMt: @ahocevar, really thanks for this hint!

Unfortunately it is not working, still get the test failing where it looks for pixels outside of the layer extent, which should return the empty array if `crossOrigin` is not set https://github.com/boeckMt/openlayers/runs/1403599911?check_suite_focus=true

Here is an Image which shows the [example](https://github.com/boeckMt/openlayers/blob/canvas-layer-renderer-test/examples/hitdetect-image.js) I have created. It is a Map with `100px X 100px` so normally the [pixel values](https://github.com/boeckMt/openlayers/commit/b005f785e2626d72a486ea40c80dbc148ce62185#diff-bc239e9a86b1a23567f4a66ea27677761f98ca67d12dea959bbaa679913abfabR147) **[50, 50] for true** and **[10, 10] for false** which I use to test should work.
image

I came to the consideration of whether it is a good decision to use the logo as a test image, since this might not always be available and the test then fails. So I tried to only test `getDataAtPixel()` https://github.com/boeckMt/openlayers/blob/test-fix-layer-at-pixel/test/spec/ol/renderer/canvas/imagelayer.test.js#L166
but there also the test fails.

Unfortunately, I have the problem that the tests do not work locally on my machine , because the done function is never called, so I tried to use the GitHub Actions on my repository.

I hope I can fix everything tomorrow and then rebase it to clean commits.

ahocevar:
@boeckMt I re-ran the CI, and now all tests pass. The last failure you saw was unrelated to your pull request. You can mark the pull request "Ready for review" now, unless there is something else you want to address before.

ahocevar (approved):
Thanks, @boeckMt!

boeckMt:
@ahocevar, thanks also for merging this fix!

PR #11740, opened 2020-11-13
Wfs bbox per feature type

### LINKED ISSUE #11187: Support unique bbox per feature types in WFS GetFeature
**Currently**

Currently, when using the WFS format to create a GetFeature XML, you can provide a list of feature type names and a single bbox as options.  As a result, we get two Query objects, one for each feature types, and the bbox ends up in the two of them.

**Required feature**

It would be nice to be able to define a unique bbox parameter for each feature type names.

The reason behind this is this: when querying using WFS onto point layers using a "mocked click", (i.e. we create a box with a buffer where the user clicked), we want to consider the size of the symbols as an extra "buffer" for the query. The symbols don't have the same size for each queryable layers, therefore it would make sense to have different bbox for each layer.

**Proposed solution**

We are aware of the "filter" option, but in our case only having the possibility to set a "bbox" for each feature types.

At the "options" level, here's how I see it.  We could either:

 - A) implement a new `[featureTypesBBOX]` option, like this:

```
var featureRequest = new WFS().writeGetFeature({
  srsName: 'EPSG:3857',
  featureNS: 'http://server.org',
  featurePrefix: 'server',
  featureTypes: ['points1', 'points2'],
  featureTypesBBOX: [
    [-8900000, 5380000, -8890000, 5390000],
    [-8901000, 5379000, -8891000, 5391000]
  ]
  outputFormat: 'application/json'
});
```

 - B) implement something more oriented to have focus on each feature type:

```
var featureRequest = new WFS().writeGetFeature({
  srsName: 'EPSG:3857',
  featureNS: 'http://server.org',
  featurePrefix: 'server',
  featureTypes: [{
    'name': 'points1',
    'bbox':  [-8900000, 5380000, -8890000, 5390000]
  }, {
    'name': 'points2',
    'bbox': [-8901000, 5379000, -8891000, 5391000]
  }],
  outputFormat: 'application/json'
});
```

So, doing the above would end up supporting two different type definition for the featureTypes: strings and objects.

**Thoughts**

Please, let me know what you think about this Feature Request.

ger-benjamin:
Fixes #11187 

Add a possibility to provide one specific bbox per feature type
on WFS writeGetFeature. This option results to one query node
per featureTypesBbox item. One query node, for one feature type,
will have a specific bbox filter and every query node will share
the same others filters (if a filter option is defined).

Also improve the doc for `WFS writeGetFeature bbox` option and remove a useless `if` in the same method.

### DIFF
--- a/src/ol/format/WFS.js
+++ b/src/ol/format/WFS.js
@@ -473,42 +473,40 @@ class WFS extends XMLFeature {
     node.setAttribute('service', 'WFS');
     node.setAttribute('version', this.version_);
     let filter;
-    if (options) {
-      if (options.handle) {
-        node.setAttribute('handle', options.handle);
-      }
-      if (options.outputFormat) {
-        node.setAttribute('outputFormat', options.outputFormat);
-      }
-      if (options.maxFeatures !== undefined) {
-        node.setAttribute('maxFeatures', String(options.maxFeatures));
-      }
-      if (options.resultType) {
-        node.setAttribute('resultType', options.resultType);
-      }
-      if (options.startIndex !== undefined) {
-        node.setAttribute('startIndex', String(options.startIndex));
-      }
-      if (options.count !== undefined) {
-        node.setAttribute('count', String(options.count));
-      }
-      if (options.viewParams !== undefined) {
-        node.setAttribute('viewParams', options.viewParams);
-      }
-      filter = options.filter;
-      if (options.bbox) {
-        assert(options.geometryName, 12); // `options.geometryName` must also be provided when `options.bbox` is set
-        const bbox = bboxFilter(
-          /** @type {string} */ (options.geometryName),
-          options.bbox,
-          options.srsName
-        );
-        if (filter) {
-          // if bbox and filter are both set, combine the two into a single filter
-          filter = andFilter(filter, bbox);
-        } else {
-          filter = bbox;
-        }
+    if (options.handle) {
+      node.setAttribute('handle', options.handle);
+    }
+    if (options.outputFormat) {
+      node.setAttribute('outputFormat', options.outputFormat);
+    }
+    if (options.maxFeatures !== undefined) {
+      node.setAttribute('maxFeatures', String(options.maxFeatures));
+    }
+    if (options.resultType) {
+      node.setAttribute('resultType', options.resultType);
+    }
+    if (options.startIndex !== undefined) {
+      node.setAttribute('startIndex', String(options.startIndex));
+    }
+    if (options.count !== undefined) {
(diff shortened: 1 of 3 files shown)


src/ol/format/WFS.js
```diff
@@ -143,13 +143,23 @@ const TRANSACTION_SERIALIZERS = {
  * @property {number} [count] Number of features to retrieve when paging. This is a
  * WFS 2.0 feature backported to WFS 1.1.0 by some Web Feature Services. Please note that some
  * Web Feature Services have repurposed `maxfeatures` instead.
- * @property {import("../extent.js").Extent} [bbox] Extent to use for the BBOX filter.
+ * @property {import("../extent.js").Extent} [bbox] Extent to use for the BBOX filter. The `geometryName`
+ * option must be set.
+ * @property {Array<FeatureTypeBbox>} [featureTypesBbox] Define a unique bbox filter per feature type
```
ahocevar, line 148: I think `featureTypes` would be a better name for the property. And the `FeatureTypeBbox` type could be renamed to `FeatureType`.

ahocevar (changes requested):
Thanks for your work on this, @ger-benjamin. I added a few comments. Would be great if you could address them.

examples/vector-wfs-getfeature.js
```diff
@@ -51,6 +51,9 @@ const featureRequest = new WFS().writeGetFeature({
   featureNS: 'http://openstreemap.org',
   featurePrefix: 'osm',
   featureTypes: ['water_areas'],
+  // bbox: [-20026376.39, -20048966.10, 20026376.39, 20048966.10],
+  // bbox: [-8874337.74061099, -8943436.814180791, 5374274.369608698, 5389561.775265733],
+  // geometryName: 'the_geom',
```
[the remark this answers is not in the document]
ger-benjamin: It's a mistake, I'll remove them, thanks

src/ol/format/WFS.js
ger-benjamin: The problem is that there is already a property named `featureTypes` in this object.
```
* @property {Array<string>} featureTypes The feature type names.
```
What should I do ?

(By the way, `featureTypes` is now optional, I've to write it `[featureTypes]` here)

ahocevar: Should not be a problem. `featureTypes` can either be an `Array<string>` or an `Array<FeatureType>`.

ger-benjamin: Done thanks :-)

ahocevar (approved):
Thanks, @ger-benjamin. This looks good to me now.

PR #11748, opened 2020-11-16
Add fetch and TextDecoder to polyfill list in README

mike-000:
#11723 updated the examples template and tutorials but omitted to update the README

#10996 added fetch to the library (previously only needed for examples).
#11723 uses TextDecoder under some conditions.


ahocevar (approved):
Thanks, @mike-000!

PR #11749, opened 2020-11-17
Document that snapTolerance must be greater than 0

ahocevar:
~A `snapTolerance` of `0` is a falsey value, so it won't be set on the instance. This pull request fixes that.~

*Edit*: A `snapTolerance` of `0` would not work, and it also would not be set on the instance because when a falsey value is provided, the default of `12` will be used. So we should just document that the value must be greater than `0`.


MoonE:
With a tolerance of 0 I'm not able to finish a LineString by clicking twice at the same point because the
calculated distance is not exactly zero but something like 2.5421149729252077e-13.
Maybe the intention was to not allow 0?

ahocevar:
Very good observation, thanks @MoonE. I re-purposed this pull request and just added a note in the API docs that the value must be greater than `0`.

marcjansen (approved):

PR #11750, opened 2020-11-18
Update TMS tileUrlFunction example in ol/source/XYZ documentation

mike-000:
The custom TMS `tileUrlFunction` example had not been updated to reflect the y coordinate change made in v6.0.0.  The reference to using `ol/source/TileImage` seems outdated as `ol/source/XYZ` also supports a `tileUrlFunction`.


ahocevar (approved):
Thanks, @mike-000!

PR #11753, opened 2020-11-19
Use Mode.LINE_STRING in Draw interaction for Circle geometries

### LINKED ISSUE #11713: Undo when drawing Circles and using geometry functions
**Describe the bug**
The updated example from #11698 highlights a difference between ability to undo between Circle geometry (cannot undo center) and other geometries which can be undone back to the first point.  Drawing a circle is basically the same as using a 2 point LineString with a geometry function to produce a circle https://codesandbox.io/s/pensive-einstein-05es0?file=/main.js which can be undone as long as the geometry function handles it.  The same would apply to `createRegularPolygon` and `createBox` functions (attempting to undo even a Circle draw using these functions currently results in an error).  Custom geometry functions such as star and ellipse (based on a 3 point LineString) can also be undone as long as the function is set up to handle it (although in the codesandbox the fill re-renders immediately while the stroke does not update until there is a pointer event).

**Expected behavior**
If Circles drawn using a 2 point LineString can be undone the same should apply if they are drawn as a Circle.
There should be no errors if drawing is undone when using the predefined geometry functions. 
Fill and Stroke should always show the same geometries.

mike-000:
Fixes #11713

Use `Mode.LINE_STRING` with a default `maxPoints` of 2 when the type is `Circle`.  This allows use of `removeLastPoint`.
Update all geometry functions to handle empty or single coordinates which may result from that.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -743,6 +743,31 @@ class Draw extends PointerInteraction {
   }
 
   /**
+   * @param {import("../geom/Polygon.js").default} geometry Polygon geometry.
+   * @private
+   */
+  createOrUpdateCustomSketchLine_(geometry) {
+    if (!this.sketchLine_) {
+      this.sketchLine_ = new Feature();
+    }
+    const ring = geometry.getLinearRing(0);
+    let sketchLineGeom = this.sketchLine_.getGeometry();
+    if (!sketchLineGeom) {
+      sketchLineGeom = new LineString(
+        ring.getFlatCoordinates(),
+        ring.getLayout()
+      );
+      this.sketchLine_.setGeometry(sketchLineGeom);
+    } else {
+      sketchLineGeom.setFlatCoordinates(
+        ring.getLayout(),
+        ring.getFlatCoordinates()
+      );
+      sketchLineGeom.changed();
+    }
+  }
+
+  /**
    * Start the drawing.
    * @param {import("../MapBrowserEvent.js").default} event Event.
    * @private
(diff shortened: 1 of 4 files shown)


ahocevar (commented):
Thanks for your work on this, @mike-000. I like the way circle drawing is simplified by removing CIRCLE draw mode.

Tha additional `opt_midpoint` argument to the `createRegularPolygon` function and drawing stretched shapes are beyond the scope of fixing #11713, and I see problems with both:

* Unfortunately the `createRegularPolygon` function has two optional arguments already. Elsewhere we have a single object literal with the options in such a case. Now add another option does not make the function look nicer.
* (Mis)using `createRegularPolygon` with a `Draw` interaction and a `minPoints: 3` configuration is not very intuitive from an API perspective. And the user experience for drawing stretched shapes is not intuitive either.

That said, I'd leave out both of these features for now. Maybe keep them in a separate branch for future use, after finding a better API and UX.

ahocevar (changes requested):

ahocevar:
@mike-000 Thanks! Would you be able to squash the commits before I give it a final review?

src/ol/interaction/Draw.js
```diff
@@ -905,11 +912,23 @@ class Draw extends PointerInteraction {
         }
       }
       this.geometryFunction_(coordinates, geometry, projection);
+
+      if (geometry.getType() === GeometryType.POLYGON && this.sketchLine_) {
```
ahocevar, line 916: Can you please explain what this `if`-block is for? It looks to me like it just writes the coordinates of the sketch line again. Why would that be necessary? Why wasn't it necessary before?

ahocevar (approved):

src/ol/interaction/Draw.js
mike-000: You can see this issue in the codesandbox in #11713  Start drawing a Star, move the mouse over the undo button and click.  The fill disappears but the sketch line does not update until the mouse moves back over the viewport.

ahocevar: `removeLastPoint` should be a noop, also when using the `CIRCLE` type (currently only when using the `LINE_STRING` type. If you fix that, you can remove this block and also the additional `if (coordinates.length >= 2)` blocks you added.

ahocevar (changes requested):

src/ol/interaction/Draw.js
mike-000: It is not specific to Circle geometry.  The Star example could equally well use LineString geometry with `maxPoints: 2` and this example https://jsfiddle.net/j0vbs2fc/ has no limit on points.

The `coordinates.length >= 2` I added fixes an existing problem.  Start drawing a polygon in https://openlayers.org/en/main/examples/draw-features.html then open the console and see the errors  being logged as you click undo.

ahocevar: Ok, I see. However, I'm still worried about recreating the sketch line over and over during modification. This could make the interaction really slow when modifying a polygon with many vertices. Also, what if you're modifying an inner ring (as can be tested in the modify-test example)?

I'd really like to see this solved differently somehow.

mike-000: The extra call only happens when the application calls `removeLastPoint` and the mode is LineString and geometry type is Polygon (i.e. when a geometry function is specified), so it is not going to be something which happens often.  It is based on existing code at https://github.com/openlayers/openlayers/blob/main/src/ol/interaction/Draw.js#L826 which is unaltered and is already called frequently when geometry functions are specified.  There is no obvious performance impact even with over 30 points defining a bezier curve polygon (which is likely to have thousands of vertices in its ring) and that is probably an extreme case.

ahocevar: Thanks for the explanation, @mike-000. In this case, it would probably make sense to factor that code out into a separate function, and call that function from both places where it is used.

This is the last remark I have. Great work!

ahocevar (changes requested):
This looks great! Would you be able to address my [last minor remark](https://github.com/openlayers/openlayers/pull/11753#discussion_r529933156)?. Thanks in advance!

ahocevar (approved):
Thanks, @mike-000

PR #11755, opened 2020-11-22
Add padding option for View

ahocevar:
In an application that covers the left half of the map viewport almost entirely with a collapsible overlay element, I needed a way to fit the map to elements so they are not covered by that overlay. I was able to handle that with `View.fit()` with a `padding`. But In addition, the app uses the History API and adds the map's center and zoom to the URL. This is what is neither covered well by `fit()` with a `padding`, nor by using the `centerOn()` method. I wanted something where I just set the `padding` and don't need to worry about how to fit or center anything.

This pull request adds what I came up with:
```js
const view = new View({
  //...
  padding: [0, 0, 0, 200];
});
```
My initial configuration is for the case where the overlay with a width of 200px is expanded. Everything on the map will be centered so it won't be covered by the overlay, because the viewport center is shifted so it is in the center of the remaining visible area.

Now when I collapse the overlay, all I need to do is
```js
view.padding[3] = 0;
```
Subsequent operations that center the map will use the center of the viewport again.

### DIFF
--- a/examples/center.html
+++ b/examples/center.html
@@ -11,6 +11,11 @@ docs: >
   view's <code>centerOn</code> method is used to position a coordinate (Lausanne)
   at a specific pixel location (the center of the black box).
   <p>Use <code>Alt+Shift+Drag</code> to rotate the map.</p>
+  <p><b>Note:</b> This example does not shift the view center. So the zoom 
+  controls and rotating the map will still use the center of the viewport as anchor.
+  To shift the whole view based on a padding, use the `padding` option on the view,
+  as shown in the <a href="view-padding.html">view-padding.html</a> example.
+  </p>
 tags: "center, rotation, openstreetmap"
 ---
 <div class="mapcontainer">
--- a/examples/view-padding.css
+++ b/examples/view-padding.css
@@ -1,0 +1,54 @@
+.mapcontainer {
+  position: relative;
+  margin-bottom: 20px;
+}
+.map {
+  width: 1000px;
+  height: 600px;
+}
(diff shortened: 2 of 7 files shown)


ahocevar:
This is ready for review. To demonstrate what's new:

New example: https://deploy-preview-11755--ol-site.netlify.app/examples/view-padding.html
Docs for the new `padding` option: https://deploy-preview-11755--ol-site.netlify.app/apidoc/module-ol_view-view#padding

mike-000:
Should `extent` and `multiWorld: false` constraints apply to the full viewport or padded region?  Currently `multiWorld: false` appears to use the size of the full viewport but is centered on the padded region.  When zoomed out panning up and down reveals an area north of the projection extent, but part of Antarctica cannot be reached.

ahocevar:
@mike-000 I don't think the padding should have an effect on constraints. It's really just an aid for helping to center the map properly when parts are covered by overlays. Similar to this in Google Maps, where I searched for 'Vienna' and the info box about Vienna covers parts of the map, so the viewport center is shifted to show all of Vienna:

Untitled

ahocevar:
Maybe the solution to the remaining issue would be to make the center constraint aware of the padding. This may be necessary because what this pull request does is simply to ask the renderers for a different center than what the view thinks the center is.

ahocevar:
This is ready for review again. The center constraint is now handled properly.

jipexu:
really a very interesting possibility !

tschaub (approved):
Looks great.

PR #11767, opened 2020-11-23
Improve documentation of geometry rotation

### LINKED ISSUE #11751: Angle inconsistencies and documentation
In `ol/style/RegularShape` an `angle` value of 0 will have one of the shape's points facing up.  `angle` and `rotation` both have positive rotation clockwise.  These are documented and consistent with most other meanings of angle and rotation in the library.

In `ol/geom/Polygon.fromCircle` an `angle` value of 0 will have one of the shape's points facing right, and positive angle values result in counter-clockwise rotation from that position.  Neither of these behaviours is documented.

`ol/interaction/Draw.createRegularPolygon` is similar to `ol/geom/Polygon.fromCircle`.  It is documented that an angle of 0 means East, although in practice 0 is a falsey value which results in the default behaviour based on current pointer position, and the documented behaviour needs an angle of `2 * Math.PI`.  Positive angle values result in counter-clockwise rotation from that position, which is not documented.

In `ol/geom/SimpleGeometry.rotate` positive angle values result in counter-clockwise rotation, which is not documented.

In this example https://codesandbox.io/s/vigorous-rhodes-46hog?file=/main.js all triangle features (`RegularShape` in red, `fromCircle`, `.rotate()` and drawn in default style) use an angle of `Math.PI / 8` .

All behaviours should be documented, and ideally be consistent (0 at top, positive rotation clockwise) even if that is a breaking change.  0 is a valid angle and should be distinguished from other falsey values such as `undefined` and `null` where appropriate.

mike-000:
Fixes #11751

Updates the documentation of geometry rotation which has not already be done in #11753

Further changes for consistency with style rotation are probably not desirable as rotation is geometry is justifiably different and is associated with transforms and polygon winding.

### DIFF
--- a/src/ol/geom/Circle.js
+++ b/src/ol/geom/Circle.js
@@ -227,7 +227,7 @@ class Circle extends SimpleGeometry {
   /**
    * Rotate the geometry around a given coordinate. This modifies the geometry
    * coordinates in place.
-   * @param {number} angle Rotation angle in radians.
+   * @param {number} angle Rotation angle in counter-clockwise radians.
    * @param {import("../coordinate.js").Coordinate} anchor The rotation center.
    * @api
    */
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #11769, opened 2020-11-24
Modify interaction developer experience improvements

### LINKED ISSUE #7404: Modify interaction "modifyend" event, return all feature, not only the modifyed feature
when i modify one feature, Modify interaction "modifyend" event
return all feature in source, not only the modifyed feature.

https://codepen.io/anon/pen/qVdEdo
var modify = new ol.interaction.Modify({
  source: vector.getSource()
});
modify.on('modifyend',function(evt){
console.log(evt.features.getArray().length);
})
image
 it's difficult to get modifyed feature in 179 features

no get modifyed feature method in api
http://openlayers.org/en/latest/apidoc/ol.interaction.Modify.html

### LINKED ISSUE #8242: request for new events "stylestart" and "styleend" on ol.interaction.Modify
Hi, 

I would like to catch the events when the modify interaction style is applied and disapplied on the feature. I want to be able to change the css cursor on "grab" or "grabbing". Events modifystart and modifyend are not good for this because the style is applied on feature with pixelTolerance settings in mind. It could work with map.forEachFeatureAtPixel but this is a bit of hack, overhead and is susceptible  to changes(e.g. pixelTolerance changes or some algorithm in ol.interaction.Modify itself changes). I could make a new grab icon, but then I still would have to disable the native browser cursor as well.

Thanks.

ahocevar:
Working with the Modify interaction in an application, I noticed several flaws regarding developer experience:

* When modifying points with a style that is different from a circle with 10px radius, it is hard to grab the symbol for dragging because the vertex selection is geometry based. I would expect being able to grab the point for dragging anywhere (like it can currently be done with the Translate interaction). This could easily be achieved by adding hit detection.
* When using the `modifystart` and `modifyend` events, the `features` in the event object are all features on the modify `source`. But I would expect the features to be only those actually being modified.
* When configuring a `style` for the interaction, the vertex feature has no indication about which feature(s) it belongs to. So it is impossible to do conditional styling (e.g. only render a drag vertex for lines, not for points). I'd expect the feature to have a `features` property, so I can apply conditional styling based on the features the vertex belongs to.

This pull request fixes these flaws.

To take advantage of hit detection for points, the `Modify` interaction has to be configured with the new `layer` option.

Fixes #7404 
Replaces #8871
Fixes #8242 (thanks to the added hit detection, modifystart/modifyend can now be used)

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -108,8 +108,11 @@ const ModifyEventType = {
  * @property {number} [pixelTolerance=10] Pixel tolerance for considering the
  * pointer close enough to a segment or vertex for editing.
  * @property {import("../style/Style.js").StyleLike} [style]
- * Style used for the features being modified. By default the default edit
- * style is used (see {@link module:ol/style}).
+ * Style used for the modification point. For linestrings and polygons, this will
+ * be the affected vertex, for circles a point along the circle, and for points the actual
+ * point. If not configured, the default edit style is used (see {@link module:ol/style}).
+ * When using a style function, the point feature passed to the function will have a `features`
+ * property - an array whose entries are the features that are being modified.
  * @property {VectorSource} [source] The vector source with
  * features to modify.  If a vector source is not provided, a layer or feature collection
  * must be provided with the `layer` or `features` option.
@@ -765,10 +768,11 @@ class Modify extends PointerInteraction {
 
   /**
    * @param {import("../coordinate.js").Coordinate} coordinates Coordinates.
+   * @param {Array<Feature>} features The features being modified.
    * @return {Feature} Vertex feature.
    * @private
    */
-  createOrUpdateVertexFeature_(coordinates) {
+  createOrUpdateVertexFeature_(coordinates, features) {
     let vertexFeature = this.vertexFeature_;
     if (!vertexFeature) {
       vertexFeature = new Feature(new Point(coordinates));
@@ -778,6 +782,7 @@ class Modify extends PointerInteraction {
       const geometry = vertexFeature.getGeometry();
       geometry.setCoordinates(coordinates);
     }
+    vertexFeature.set('features', features);
     return vertexFeature;
   }
 
@@ -830,9 +835,14 @@ class Modify extends PointerInteraction {
       evt.coordinate[0] + this.delta_[0],
       evt.coordinate[1] + this.delta_[1],
     ];
+    const features = [];
     for (let i = 0, ii = this.dragSegments_.length; i < ii; ++i) {
       const dragSegment = this.dragSegments_[i];
       const segmentData = dragSegment[0];
+      const feature = segmentData.feature;
+      if (features.indexOf(feature) === -1) {
+        features.push(feature);
+      }
       const depth = segmentData.depth;
       const geometry = segmentData.geometry;
       let coordinates;
@@ -912,7 +922,7 @@ class Modify extends PointerInteraction {
         this.setGeometryCoordinates_(geometry, coordinates);
       }
     }
-    this.createOrUpdateVertexFeature_(vertex);
+    this.createOrUpdateVertexFeature_(vertex, features);
   }
 
   /**
@@ -1160,7 +1170,7 @@ class Modify extends PointerInteraction {
           node.index === CIRCLE_CIRCUMFERENCE_INDEX
         ) {
           this.snappedToVertex_ = true;
-          this.createOrUpdateVertexFeature_(vertex);
+          this.createOrUpdateVertexFeature_(vertex, [node.feature]);
         } else {
           const pixel1 = map.getPixelFromCoordinate(closestSegment[0]);
           const pixel2 = map.getPixelFromCoordinate(closestSegment[1]);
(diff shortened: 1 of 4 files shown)


ahocevar:
This pull request is now ready for review.

ahocevar:
I added another commit that simplifies both code and API for hit detection mode, and an example to show the new `layer` option. This should make it easier to review.

MoonE (commented):
-If I set a negative y-scale on the icon image the hit still works on the unmodified icon.-
I did not properly change the scale, it works fine...

src/ol/interaction/Modify.js
```diff
@@ -314,27 +332,44 @@ class Modify extends PointerInteraction {
...
+        features = new Collection(this.source_.getFeatures());
+        this.source_.addEventListener(
+          VectorEventType.ADDFEATURE,
+          this.handleSourceAdd_.bind(this)
+        );
+        this.source_.addEventListener(
+          VectorEventType.REMOVEFEATURE,
+          this.handleSourceRemove_.bind(this)
+        );
+      }
```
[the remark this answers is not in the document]
ahocevar: Thanks for raising this question. It made me re-think the API, and I decided to abandon the new `layer` option. Instead, I added a new `hitDetection` option, which can be configured with a boolean (for detection of the feature on any layer) or a layer (for detection of the feature on a specific layer).

ahocevar:
Thanks for the review, @MoonE. I have addressed your comments.

src/ol/interaction/Modify.js
```diff
@@ -376,14 +408,27 @@ class Modify extends PointerInteraction {
...
+        if (features.indexOf(feature) === -1) {
+          this.featuresBeingModified_.push(feature);
+        }
+      }
+
       this.dispatchEvent(
-        new ModifyEvent(ModifyEventType.MODIFYSTART, this.features_, evt)
+        new ModifyEvent(
+          ModifyEventType.MODIFYSTART,
+          new Collection(features),
```
MoonE, line 429: This should be `this.featuresBeingModified_` instead of a new Collection.

ahocevar: Thanks, fixed.

src/ol/interaction/Modify.js
```diff
@@ -108,14 +108,24 @@ const ModifyEventType = {
...
+ * property - an array whose entries are the features that are being modified, and a `geometries`
+ * property - an array whose entries are the geometries that are being modified. Both arrays are
+ * in the same order. The `geometries` are only useful when modifying geometry collections, where
+ * the geometry will be the particular geometry from the collection that is being modified.
  * @property {VectorSource} [source] The vector source with
  * features to modify.  If a vector source is not provided, a feature collection
- * must be provided with the features option.
+ * must be provided with the `features` option.
+ * @property {boolean|import("../layer/BaseVector").default} [hitDetection] When configured, point
+ * features will considered for modification based on their visual appearance, instead of being within
```
MoonE, line 123:
```suggestion
 * features will be considered for modification based on their visual appearance, instead of being within
```

MoonE (changes requested):
If hit detection is enabled with a point and line feature and the hit is detected on the point feature a point is inserted on the line if that point is closer to the clicked position.
I updated the new example here to show the issue. If you grab the icon at the top it will instead insert a new vertex into the line.
https://github.com/MoonE/openlayers/tree/modify-dx-bug

ahocevar:
Thanks for testing this, @MoonE. I have addressed your comment - in your example, no line vertex won't be inserted any more when the icon is hit detected.

MoonE:
Thanks you, @ahocevar, unfortunately I found another bug. I updated the example in https://github.com/MoonE/openlayers/tree/modify-dx-bug.

If you drag the icon it moves both of the  green line at [0, 0] it should only move one.
If you then alt click the point feature to delete, it only deletes one of the points for the green line, but seems to delete both from its internal rbush. The internal representation is out of sync from then on.

I also updated the code for the `pointermove` event in the example to better work with hitdetection and polygons. It should only change the cursor when it is near one of the lines not when it is anywhere over the polygon. It is probably better to use this in the example because this should always work.

ahocevar:
@MoonE I just rebased this pull request to bring in your changes from #11803. So the other bug you noticed should be fixed now?

I also added a commit, taking inspiration from your change to get a besser cursor feedback when hovering over modification candidates, removing the overhead of another hit detection cycle.

MoonE:
Yes, this seems to work now.

Something else I noticed:
If you modify a point feature it sometimes gets grabbed and dragged wherever it was clicked and sometimes it is placed directly under the pointer. This depends on the distance of the click from the point coordinate
Maybe it should always use the new behavior to not center the vertex under the cursor? It feels better to me and would be more consistent.

ahocevar:
Good suggestion, @MoonE. Changed with 3e3cd08.

ahocevar:
For the previous change, I had to increase a few test tolerances where circle radiuses are calculated from pointer locations, because now the pointer location counts, instead of the vertex geometry. Ready for review again.

test/spec/ol/interaction/modify.test.js
```diff
@@ -1000,7 +1084,7 @@ describe('ol.interaction.Modify', function () {
       simulateEvent('pointerdrag', 30, -5, null, 0);
       simulateEvent('pointerup', 30, -5, null, 0);
 
-      expect(circleFeature.getGeometry().getRadius()).to.equal(25);
+      expect(circleFeature.getGeometry().getRadius()).to.roughlyEqual(25, 1e9);
```
MoonE, line 1087: The tolerance is way too high here, either should be 1e-9 or 0.1 like the others.

ahocevar: Right, thanks for catching this! Fixed.

MoonE (approved):
Great work, @ahocevar. Everything looks good now.

ahocevar:
Thanks a lot @MoonE for your sharp eye and all the help on this!

PR #11775, opened 2020-11-26
Enable drawing MultiPoint geometry with more than one point

mike-000:
Currently the Draw interaction casts a single Point as a MultiPoint geometry.  This change enables MultiPoint geometry with more than one point to be drawn, while keeping the existing behaviour as the default for backward compatibility.

Add MultiPoint to the Draw Features example.

Add tests for MultiPoint drawing when maxPoints is set at more than one.

### DIFF
--- a/examples/draw-features.js
+++ b/examples/draw-features.js
@@ -29,9 +29,17 @@ let draw; // global so we can remove it later
 function addInteraction() {
   const value = typeSelect.value;
   if (value !== 'None') {
+    let maxPoints;
+    if (value === 'MultiPoint') {
+      // By default MultiPoint drawing casts a single Point
+      // as a MultiPoint geometry. Override this to allow
+      // the user to choose when to finish drawing.
+      maxPoints = Infinity;
+    }
     draw = new Draw({
       source: source,
       type: typeSelect.value,
+      maxPoints: maxPoints,
     });
     map.addInteraction(draw);
   }
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -52,11 +52,13 @@ import {squaredDistance as squaredCoordinateDistance} from '../coordinate.js';
  * @property {boolean} [stopClick=false] Stop click, singleclick, and
  * doubleclick events from firing during drawing.
  * @property {number} [maxPoints] The number of points that can be drawn before
- * a polygon ring or line string is finished. By default there is no
- * restriction.
+ * a polygon ring, line string or multi-point geometry is finished. By default there
+ * is no restriction for polygon rings or line strings. The default is `1` for
+ * multi-point geometries.
  * @property {number} [minPoints] The number of points that must be drawn
- * before a polygon ring or line string can be finished. Default is `3` for
- * polygon rings and `2` for line strings.
+ * before a polygon ring, line string or multi-point geometry can be finished.
+ * Default is `3` for polygon rings, `2` for line strings, and `1` for
+ * multi-point geometries.
  * @property {import("../events/condition.js").Condition} [finishCondition] A function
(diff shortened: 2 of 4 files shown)


src/ol/interaction/Draw.js
```diff
@@ -630,7 +641,7 @@ class Draw extends PointerInteraction {
       if (this.shouldHandle_) {
         if (!this.finishCoordinate_) {
           this.startDrawing_(event);
-          if (this.mode_ === Mode.POINT) {
+          if (this.mode_ === Mode.POINT || this.maxPoints_ < 2) {
```
ahocevar, line 644: Instead of deciding this based on `this.maxPoints_`, The `finishCondition` should be tested here, like for the other geometry types.

examples/draw-features.js
```diff
@@ -29,9 +29,17 @@ let draw; // global so we can remove it later
 function addInteraction() {
   const value = typeSelect.value;
   if (value !== 'None') {
+    let maxPoints;
+    if (value === 'MultiPoint') {
+      // By default MultiPoint drawing casts a single Point
+      // as a MultiPoint geometry. Override this to allow
+      // the user to choose when to finish drawing.
+      maxPoints = Infinity;
```
ahocevar, line 37: In a real life use case, I think drawing an infinite number of points in a single MultiPoint geometry is not very common. A typical UI will use a `finishCondition` or a button connected with a `finishDrawing()` call when multipoint drawing is done.

ahocevar (commented):
Thanks for your work on this @mike-000. See my comments. This pull request adds a feature for multipoints which is not available for other geometry types, and maybe not very useful at all. Just so you know, the reason for supporting multi geometries here is that some WFS or similar services use multi geometries, and fail when single geometries are uploaded. It was not the intention to be able to draw multi geometries with multiple entities.

examples/draw-features.js
mike-000: Setting Infinity should make it behave like a drawing a LineString.  2 clicks on the same point will finish it.  I've not tested with a `finishCondition` but it goes through the same code.

src/ol/interaction/Draw.js
mike-000: When `maxPoints` is 1 that test makes the behaviour exactly the same as a single Point draw (and the current MultiPoint).   How could a `finishCondition` be relevant it that situation?

examples/draw-features.js
ahocevar: The code of this interaction is becoming increasingly hard to understand. I know I approved the use of `LINE_STRING` mode for drawing circles. But now that `LINE_STRING` mode is also being used to draw multi points, I regret that. Thinking more about it now, I wish there was still a `CIRCLE` mode. Even if it is more or less just an alias for `LINE_STRING`, it could help avoid if-clauses like the ones that were introduced in https://github.com/openlayers/openlayers/pull/11753/files#diff-91329c8a038835b6a5e7bae5e13f15e7fa5499b3414f0e094a5bfe347efc5a32R957.

For multipoints, the code would also be easier to understand if the mode was still `POINT`. And I'm still not convinced that this pull request is serving a useful purpose. Because if I want multipoints, I can as well draw points and use the `drawend` event to take the point and add it to a multipoint geometry. Same as for other multi geometries.

mike-000: It can also be easily done using a geometry function  https://codesandbox.io/s/draw-features-forked-yhmv7?file=/main.js which does the same as the PR without adding anything extra to the interaction code, so if it has limited use I will close this.

ahocevar: Thanks @mike-000. Please close this. If you'd be able to implement the code readability improvements I [suggested](https://github.com/openlayers/openlayers/pull/11775#discussion_r531200283), it would be much appreciated.

PR #11778, opened 2020-11-27
No need to adjust hitTolerance for pixel ratio twice

ahocevar:
I think the behavior of the `hitTolerance` option of `Map#getFeaturesAtPixel()` and friends is buggy. Its units should be css pixels, not canvas pixels, so users don't have to configure different behavior depending on the device pixel ratio. This is what I mean:

devicePixelRatio === 2:
Nov-27-2020 12-31-55

devicePixelRatio === 1:
Nov-27-2020 12-30-46

As you can see, the tolerance is different, and I would expect it to always be the same as the area preview at the bottom.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,20 @@
 ## Upgrade notes
 
+### Next version
+
+#### Units of the `hitTolerance` option fixed
+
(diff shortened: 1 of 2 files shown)


ahocevar:
I updated the changelog, so users are not surprised by this bugfix. This is now ready for review.

MoonE:
I'm pretty sure the file `src/ol/interaction/Modify.js.orig` should not be there.

ahocevar:
You are absolutely right @MoonE. Removed and force-pushed.

MoonE (approved):
Effectively the tolerance was multiplied twice with the pixelRatio.
With a pixelRatio of 2 a configured tolerance of 10 should become 20 screen pixel, but due to this bug it became 40 px instead.
Maybe you could this in the change log like that.

The change itself is looking good.

ahocevar:
Thanks for the review, @MoonE! I updated the changelog.

PR #11780, opened 2020-11-27
Add a cancel event to the DragBox interaction

M393:
Currently there is only a boxstart and boxend event but when the box is canceled it is silent.

This adds a boxcancel event in the case where no boxend event is dispatched.

I have an overlay in the map and on boxstart I hide the overlay because when the mouse is above the overlay the event is not passed through to the map. When the interaction is finished it should become visible again, but this is currently possible because of the missing event.


ahocevar (approved):
Thanks, @M393.

PR #11781, opened 2020-11-27
Pass geometry to forEachFeatureAtPixel callback

ahocevar:
Working on #11769, I realized that when editing a geometry collection, it is useful to not only get the feature that a hit was detected on, but also the exact geometry from the collection. Since this information is available at the time a hit is detected, this change is only about passing that information all the way to the map's `forEachFeatureAtPixel()` method.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -538,7 +538,7 @@ class PluggableMap extends BaseObject {
    * callback with each intersecting feature. Layers included in the detection can
    * be configured through the `layerFilter` option in `opt_options`.
    * @param {import("./pixel.js").Pixel} pixel Pixel.
-   * @param {import("./renderer/vector.js").FeatureCallback<T>} callback Feature callback. The callback will be
+   * @param {function(import("./Feature.js").FeatureLike, import("./layer/Layer.js").default, import("./geom/SimpleGeometry.js").default): T} callback Feature callback. The callback will be
    *     called with two arguments. The first argument is one
    *     {@link module:ol/Feature feature} or
    *     {@link module:ol/render/Feature render feature} at the pixel, the second is
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -538,8 +538,7 @@ class PluggableMap extends BaseObject {
    * callback with each intersecting feature. Layers included in the detection can
    * be configured through the `layerFilter` option in `opt_options`.
    * @param {import("./pixel.js").Pixel} pixel Pixel.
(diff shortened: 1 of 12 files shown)


src/ol/PluggableMap.js
```diff
@@ -538,8 +538,7 @@ class PluggableMap extends BaseObject {
    * callback with each intersecting feature. Layers included in the detection can
    * be configured through the `layerFilter` option in `opt_options`.
    * @param {import("./pixel.js").Pixel} pixel Pixel.
-   * @param {function(this: S, import("./Feature.js").FeatureLike,
```
MoonE, line 541: The description should be updated to include the third geometry parameter.
The `FeatueCallback<T>` type is not available for the apidocs, becomes a dead link.

src/ol/renderer/canvas/VectorLayer.js
```diff
@@ -423,13 +423,14 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
...
        * @param {import("../../Feature.js").FeatureLike} feature Feature.
+       * @param {import("../../geom/SimpleGeometry.js").default} geometry Geometry.
        * @return {?} Callback result.
        */
-      const featureCallback = function (feature) {
+      const featureCallback = function (feature, geometry) {
         const key = getUid(feature);
         if (!(key in features)) {
           features[key] = true;
-          return callback(feature, layer);
```
MoonE, line 432: The `forEachFeatureAtPixel` callback will only be called with the first detected geometry, but multiple geometries of the same feature could be at that pixel.
This should probably also be documented.

src/ol/PluggableMap.js
ahocevar: I intentionally did not document that third parameter, because I am not sure it is relevant for the API. The reason why I added it was to be able to use the Modify interaction with hit detection and geometry collections after #11781, where this third parameter is used internally.

I did update the type to avoid the dead link, so the documentation should now be the same as before this pull request.

ahocevar:
I updated the documentation now at least for internal use according to your suggestions. Thanks, @MoonE.

MoonE:
If this is used with the new layer option of the Modify interaction MutliGeometries will not be supported because there is just one geometry object. Is this expected?

MoonE (approved):
If the point in my previous comment is not an issue this looks good to me.

ahocevar:
> If this is used with the new layer option of the Modify interaction MutliGeometries will not be supported because there is just one geometry object. Is this expected?

I know it is unfortunate, but it is expected. For geometry collections, this change was a low hanging fruit. For multi geometries, this would be much harder to achieve.

PR #11782, opened 2020-11-27
Start from hit circle center when hitTolerance is configured

ahocevar:
When configuring a `hitTolerance` for `Map#getFeaturesAtPixel()` and friends, the user experience is somewhat strange. The reason is that hits on the hit canvas are checked starting from the top-left corner of the hit canvas. The user experience can be fixed by starting from the center of the hit canvas.

### DIFF
--- a/src/ol/render/canvas/ExecutorGroup.js
+++ b/src/ol/render/canvas/ExecutorGroup.js
@@ -230,8 +230,12 @@ class ExecutorGroup {
     function featureCallback(feature) {
       const imageData = context.getImageData(0, 0, contextSize, contextSize)
         .data;
-      for (let i = 0; i < contextSize; i++) {
-        for (let j = 0; j < contextSize; j++) {
+      for (let ci = 0; ci < contextSize; ci++) {
+        const i =
+          ci % 2 ? (hitTolerance - 0.5 * ci) | 0 : hitTolerance + 0.5 * ci;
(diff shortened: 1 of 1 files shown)


MoonE:
It would make sense to cache the indexes in the required order like this:
https://github.com/MoonE/openlayers/tree/hitdetect-indexes

Instead of iterating over the two-dimensional boolean array for each hit detection, the indexes can be stored in an array in the order they should be checked.
I more or less copied your changes to the getCircleArray function, so the end result should be the same.

MoonE:
The order in which pixels are checked is not quite correct.
It starts at the center pixel, but then checks complete lines, though it should kind of spiral outwards instead.
This is the order in which pixels would be checked with a tolerance of 3:
```json
[
  [3,3],
  [3,2],
  [3,4],
  [3,1],
  [3,5],
  [2,3],
  [2,2],
  [2,4],
  [4,3],
  [4,2],
  [4,4],
  [1,3],
  [5,3]
]
```

What I also noticed is that the circle array always seems to have a 1px border (unrelated to this change)
e. g. with zero tolerance it checks the center of a 1x1 array, with 1 it checks only the center pixel of a 3x3 array.

ahocevar:
Thanks @MoonE! Can you please create a pull request from your `hitdetect-indexes` branch to replace my pull request?

PR #11783, opened 2020-11-28
Cache hit detect indexes and check closest pixels first.

MoonE:
replaces / closes #11782

@ahocevar Could you explain what kind of user experience problems you were trying to solve by checking the center first?
After looking at this some more as far as I understand it the order in which features are checked is determined by the render order, not by how close the user clicked to a feature.

Using an array with indexes still makes sense and looking at the closest pixels first will probably find a hit sooner (if it exists) but I think the user experience will be the same.

### DIFF
--- a/examples/hit-tolerance-priority.html
+++ b/examples/hit-tolerance-priority.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: Hit tolerance priority
+shortdesc: Shows bad behavior of hit detection with hit tolerance.
+docs: >
+  Hover over the map and observe how the small circles get a black outline as you hover over them. Is the expected feature getting highlighted?
+tags: "simple, openstreetmap"
+---
+<div id="map" class="map"></div>
--- a/examples/hit-tolerance-priority.js
+++ b/examples/hit-tolerance-priority.js
@@ -1,0 +1,84 @@
+import CircleStyle from '../src/ol/style/Circle.js';
+import Feature from '../src/ol/Feature.js';
+import Map from '../src/ol/Map.js';
+import VectorLayer from '../src/ol/layer/Vector.js';
+import VectorSource from '../src/ol/source/Vector.js';
+import View from '../src/ol/View.js';
+import {Fill, Stroke, Style} from '../src/ol/style.js';
+import {Point} from '../src/ol/geom.js';
+
+const map = new Map({
+  target: 'map',
+  view: new View({
+    center: [0, 0],
+    resolution: 1,
+    resolutions: [1],
+  }),
+});
+
+const vectorLayer = new VectorLayer({
+  source: new VectorSource({
+    features: [
+      new Feature({
+        geometry: new Point([0, 0]),
+        color: 'white',
+      }),
+      new Feature({
+        geometry: new Point([-10, 0]),
+        color: 'fuchsia',
+      }),
+      new Feature({
+        geometry: new Point([-10, -10]),
+        color: 'orange',
+      }),
+      new Feature({
+        geometry: new Point([-10, 10]),
+        color: 'cyan',
+      }),
+    ],
+  }),
+  style: (feature) => {
+    return new Style({
+      image: new CircleStyle({
+        radius: 5,
+        fill: new Fill({
+          color: feature.get('color'),
+        }),
+        stroke: new Stroke({
+          color: 'gray',
+          width: 1,
+        }),
+      }),
+    });
+  },
+});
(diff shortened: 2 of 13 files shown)


ahocevar:
Thanks for your research @MoonE! You are right, my thinking was wrong. What I wanted to achieve is that the feature closest to the pointer gets highlighted when using `forEachFeatureAtPixel()` to get the first (i.e. most relevant) feature. Like you found out correctly, the way hit tolerance handling is implemented, this won't be the case. I created an example that shows the problem: https://github.com/ahocevar/openlayers/commit/7119412caeb72b27d4c8232f33a327c0f28b0b74.

Regardless of that, I think this pull request is a good improvement.

While we're at it, do you see a good way to fix this problem? I think we could collect all matching features and sort them in `ExecutorGroup#forEachFeatureAtCoordinate`, before executing the callback passed to that function. But that could potentially complicate things a fair bit.

MoonE:
I tried your suggestions in  https://github.com/MoonE/openlayers/tree/hitdetect-closest.
This works per layer at the moment The matches would need to be collected one level above to make it work across all layers.
Would need to be implemented in a similar way for the other renderers.
Please let me know if this looks promising.

For features with direct hits the callback is called immediately, the other features are collected and sorted by distance before calling the callback with them. 
If a truthy value is returned from the callback for a feature that was directly hit the hit detection is stopped early, like before.

ahocevar:
This looks great, @MoonE! Less complicated than I thought. It would be awesome if you could finish that as you proposed - by making it work across layers and for the other renderers. I'd also be willing to take on some of the remaining tasks. Just let me know.

This is a significant usability improvement, thanks a lot for your effort!

MoonE:
This should now work for `VectorLayerRenderer` and `VectorTileLayerRenderer`.
`CanvasLayerRenderer.getDataAtPixel` and  `WebGLPointsLayerRenderer.forEachFeatureAtCoordinate` ignore the hitTolerance, so I did not change anything there.
Should hitTolerance be used in the those two?

`MapRenderer.hasFeatureAtCoordinate` is kind of inefficient now when no direct hit is made because it still collects all features.

I did not see more places that need to be changed.

MoonE:
Could the executor group be skipped entirely if `tileContainsCoordinate` is false (line 438)? Its value only changes per tile.
Or is it broken with declutter and the check should be removed (see line 411)? 
https://github.com/openlayers/openlayers/blob/8cc78acde90c9e4b1751d16aa388b8d1cd073224/src/ol/renderer/canvas/VectorTileLayer.js#L405-L454

ahocevar:
> CanvasLayerRenderer.getDataAtPixel and WebGLPointsLayerRenderer.forEachFeatureAtCoordinate ignore the hitTolerance, so I did not change anything there.
Should hitTolerance be used in the those two?

No, that's ok to leave it unchanged.

ahocevar:
> Could the executor group be skipped entirely if tileContainsCoordinate is false (line 438)? Its value only changes per tile.
Or is it broken with declutter and the check should be removed (see line 411)?

Thanks for spotting this flaw! Now that I look at it, I would say the executor group can be skipped entirely in that case.

MoonE:
OK, thanks. Now it only checks for tiles containing the provided coordinate.

Another question:
Shouldn't the check if the tile contains the coordinate account for the hitTolerance?

ahocevar:
> Shouldn't the check if the tile contains the coordinate account for the hitTolerance?

Good question. Thinking about it, I would say it cannot hurt. Usually vector tiles use a buffer where features are repeated on neighboring tiles, so it should not strictly be necessary. But we cannot take that for granted, because we don't know how the tiles we process were generated.

MoonE:
This is ready for review. Thank you for your help with this so far, @ahocevar.

ahocevar:
This looks excellent, @MoonE! Unfortunately the merging of #11781 caused conflicts in this pull request. Would you be able to resolve those, or do you want me to look into it?

MoonE:
I added the template argument to the new HitMatch typedef and got an error that T|void is not asignable to T.
I then changed the signature of the forEachFeatureAtPixel function from T|void to T|undefined and added this also to the inline featureCallback functions. I hope this makes sense.

ahocevar:
This makes perfect sense. Thanks again for your effort on this, @MoonE!

ahocevar (approved):
Great work, @MoonE. Thanks and please merge!

MoonE:
Thanks for your help with this, @ahocevar.

PR #11790, opened 2020-12-01
Fix attributions links for WMS layers from swisstopo

### LINKED ISSUE #11766: Attribution links not working
**Describe the bug**
The links in the attribution in the **WMS without Projection** and **Single Image WMS with Proj4js** examples are no longer working.  They also lack `target="_blank"` and overwrite the map. 

**Expected behavior**
The English language home page now appears to be https://www.geo.admin.ch/en/home.html  The public WMS service is described in https://www.geo.admin.ch/de/geo-dienstleistungen/geodienste/darstellungsdienste-webmapping-webgis-anwendungen/web-map-services-wms.html but is not available in English..

fredj:
fixes #11766

The attribution link is now the layer description page at Swisstopo. Because the "National parks" layer has no English page, I've changed it to "Flood Alert" layer.

### DIFF
--- a/examples/reprojection.js
+++ b/examples/reprojection.js
@@ -92,8 +92,8 @@ layers['wms4326'] = new TileLayer({
 layers['wms21781'] = new TileLayer({
   source: new TileWMS({
     attributions:
-      '© <a href="http://www.geo.admin.ch/internet/geoportal/' +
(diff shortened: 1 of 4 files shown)


ahocevar (approved):
Thanks, @fredj

PR #11405, opened 2020-08-06
Fixes extent calculation in WMTS optionsFromCapabilities when BoundingBox exists

amir-ba:
As mentioned  in #11350 when creating the extent, the WMTS optionsFromCapabilities dose not consider the Wgs84BoundingBox. This pull request attempts to fix this by using the Wgs84BoundingBox extent in cases where it exists.

PR #9590, opened 2019-05-22
Calculate tile grid extent from extent of bottom-level tile matrix

mloskot:
Prefers extent derived from the tile matrix set in the capabilities over default projection extent.

-----

It's not really a new feature or a bug fix, but a small tweak that improves WMTS experience:

It augments tile grid extent based on extent calculated from parameters of the bottom-level matrix as reported by capabilities. This works around some issues when using custom tile matrix sets, etc.

I'm no JavaScript programmer or experienced OpenLayers user, so apologies if I abused the language/library in any way. I'll be thankful for any feedback about how to improve my patch.

PR #11792, opened 2020-12-01
Do not set projection extent in WMTS optionsFromCapabilities unless matrix set projection is also supported

### LINKED ISSUE #11738: Failed optionsFromCapabilities call corrupts the projection
**Describe the bug**
Calling `optionsFromCapabilities` with a projection that is not supported by the WMTS server will throw an error. This is expected and documented behaviour, so I tried to use it to select a suitable projection by wrapping the call in a try-catch and calling it with different projections until one works. The problem is that those failed `optionsFromCapabilities` calls seem to somehow "corrupt" the projections. For example, creating an `ImageCanvasSource` with a projection that was used for a failed `optionsFromCapabilities` call causes the canvas function to receive invalid parameters.

**To Reproduce**

**_Edit: It seems that the WMTS server I used in the sample was closed, it returns 404 now. But this can most likely be reproduced with other WMTSs and projections as well_**

1. Clone and run the sample: https://github.com/janruo/ol-issue
2. See that it works, a red rectangle is drawn on the map.
3. See the console output, canvasFunction parameters are valid.
4. Uncomment line 101 of index.js, comment out line 104.
5. See that it no longer works, no rectangle.
6. See the console output, canvas function received invalid extent and size parameters.

**Expected behavior**
Failed `optionsFromCapabilities` calls should not cause the projection to become unusable.

mike-000:
Fixes #11738

If the matrix set does not contain a useable projection an error is thrown by ol/tilegrid/WMTS.createFromCapabilitiesMatrixSet (where an assertion might be nicer than `Cannot read property 'getMetersPerUnit' of undefined`)  It seems inappropriate to set the extent of the projection supplied in the WMTS optionsFromCapabilities config options to the matrix extent unless it is going to be used.

### DIFF
--- a/src/ol/source/WMTS.js
+++ b/src/ol/source/WMTS.js
@@ -2,6 +2,7 @@
  * @module ol/source/WMTS
  */
 
+import Projection from '../proj/Projection.js';
 import TileImage from './TileImage.js';
 import WMTSRequestEncoding from './WMTSRequestEncoding.js';
 import {appendParams} from '../uri.js';
@@ -502,15 +503,26 @@ export function optionsFromCapabilities(wmtsCap, config) {
       wgs84BoundingBox[2] === wgs84ProjectionExtent[2];
   }
 
-  if (projection.getExtent() === null) {
-    projection.setExtent(extent);
-  }
-
   const tileGrid = createFromCapabilitiesMatrixSet(
     matrixSetObj,
     extent,
     matrixLimits
   );
+
+  // if projection has no extent do not change it,
+  // use a new projection with the matrix extent
+  if (projection.getExtent() === null) {
+    projection = new Projection({
+      code: projection.getCode(),
+      units: projection.getUnits(),
+      extent: extent,
+      axisOrientation: projection.getAxisOrientation(),
+      global: projection.isGlobal(),
+      metersPerUnit: projection.getMetersPerUnit(),
+      worldExtent: projection.getWorldExtent(),
+      getPointResolution: projection.getPointResolutionFunc(),
+    });
+  }
 
   /** @type {!Array<string>} */
   const urls = [];
--- a/test/spec/ol/format/wmts/os.xml
+++ b/test/spec/ol/format/wmts/os.xml
@@ -1,0 +1,859 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Capabilities
+	xmlns="http://www.opengis.net/wmts/1.0"
+	xmlns:ows="http://www.opengis.net/ows/1.1"
+	xmlns:xlink="http://www.w3.org/1999/xlink"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:gml="http://www.opengis.net/gml"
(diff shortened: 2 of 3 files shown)


ahocevar (changes requested):
This fix is a bit problematic, because it only makes the underlying problem less obvious. A better fix would be if `optionsFromCapabilities()` would not set an extent on the projection at all. For example by creating a new `Projection` object, which is returned as `projection` property in the options output.

ahocevar (commented):
Thanks, @mike-000. This looks good now. Would you be able to add a test to make it perfect?

ahocevar (commented):
Instead of adding a tile matrix set to `test/spec/ol/format/wmts/ogcsample.xml`, could you please create a new wmts capabilities doc for the new tests? Maybe a real-life one from Ordnance Survey?

mike-000:
I inadvertently chose a layer with TileMatrixSetLimits for a test and found the limits were reflected in the projection extent.  Beyond the scope of this PR but perhaps a new issue as that might not be desirable, for example if multiple layers were to be used as in https://labs.os.uk/public/os-data-hub-examples/os-maps-api/zxy-example-change-style (but using EPSG:27700 and WMTS with the Leisure layer) the outcome could depend on the order the layers were defined.

ahocevar:
> I inadvertently chose a layer with TileMatrixSetLimits for a test and found the limits were reflected in the projection extent. Beyond the scope of this PR but perhaps a new issue as that might not be desirable, for example if multiple layers were to be used

Absolutely right @mike-000, that's not desirable.

ahocevar:
And thinking about it more, I feel like having the `optionsFromCapabilities` function set an extent on a projection at all is wrong. Since `optionsFromCapabilities` uses `WMTSTileGrid#createFromCapabilitiesMatrixSet` to create a suitable tilegrid (with the proper extent), I see no reason why #9590 added this if-block in the first place:

https://github.com/mloskot/openlayers/blob/98e8bec3709364b5fd14dd71a4e4e72dcf4ae1c9/src/ol/source/WMTS.js#L396-L398

@mloskot Can you maybe shed some light on this? All tests continue to pass with that block removed.

@mike-000 I'm leaving this pull request open until we hear back from @mloskot.

mike-000:
Creating new projections will mitigate against setting the extent of a single shared projection too small, but I agree that setting projection extents based on the tile grid can be meaningless.  Esri's tile grid for EPSG:27700 as used in the Raster Reprojection example is much larger than the normal projection extent with the OS Overview map reduced to a few pixels in one corner of the 0/0/0 tile.  If used as the limits for reprojection that would probably cause errors.

image

mloskot:
@ahocevar TBH, it's been a while and I can't remember exactly. I think I added that as a fallback, a safety net.

mike-000:
@ahocevar How to you wish to proceed?  The current code still sets an extent on the primary projection object which seems wrong in any situation.  Either a new PR to remove that line and always leave it to the application to set extents, or this best attempt at setting an extent on a new copy of the projection which may not be perfect but perhaps better than doing nothing?  A rebase was needed due to #11405 but that means my tests will need to be revised.

ahocevar:
I'd say leave setting the extent to the application. If I'm not mistaken, that extent on the projection is only relevant for `wrapX`.

PR #11800, opened 2020-12-03
Fix removeLastPoint when removing last point

ahocevar:
This pull request fixes an issue with the Draw interaction's `removeLastPoint` method. Previously, when no vertex of a polygon or linestring was left to undo, it took two more calls of `removeLastPoint()` to abort the drawing. Now it only takes one more, which I think is the expected behavior.

This issue can be seen in the draw-features.html example, when drawing a polygon or linestring and repeatedly clicking undo to get rid of the entire geometry.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -59,7 +59,8 @@ import {squaredDistance as squaredCoordinateDistance} from '../coordinate.js';
  * polygon rings and `2` for line strings.
  * @property {import("../events/condition.js").Condition} [finishCondition] A function
  * that takes an {@link module:ol/MapBrowserEvent~MapBrowserEvent} and returns a
- * boolean to indicate whether the drawing can be finished.
+ * boolean to indicate whether the drawing can be finished. Not used when drawing
+ * POINT or MULTI_POINT geometries.
  * @property {import("../style/Style.js").StyleLike} [style]
  * Style for sketch features.
  * @property {GeometryFunction} [geometryFunction]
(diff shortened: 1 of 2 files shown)


ahocevar:
I see now why the `removeLastPoint()` function behaved like it did: to make it work with the `appendCoordinates()` method like in the `tracing.html` example. To not break any tests, I had to improve the `appendCoordinates()` method to also work when no drawing is active.

This is now ready for review.

src/ol/interaction/Draw.js
```diff
@@ -937,7 +936,7 @@ class Draw extends PointerInteraction {
       this.geometryFunction_(this.sketchCoords_, geometry, projection);
     }
 
-    if (coordinates.length === 0) {
+    if (coordinates.length === 1) {
```
ahocevar, line 939: This one-line change is what I intended. The rest of the changes is to not break tests and to keep the `tracing.html` example work like it used to. Useful changes though, because `appendCoordinates()` can now also be used to start a new drawing.

mike-000: There are several `if (coordinates.length) {` checks in the geometry functions which may no longer be needed as a result of that.

ahocevar: Thanks! I knew those checks shouldn't be there.

ahocevar:
Working on that fix revealed several issues with the appendCoordinates() method, which I also fixed now. That can be seen in the [tracing.html](https://deploy-preview-11800--ol-site.netlify.app/examples/tracing.html) example: On a touch device, the trace is now immediately visible. [Previously](https://openlayers.org/en/latest/examples/tracing.html] it was only visible after another click.

jahow (approved):
Thanks! I have tested the tracing example on mobile, it's indeed much better now. Still behaving a bit weirdly (the last clicked point is not the one highlighted), but the trace is at least correctly rendered. Maybe this has to be improved in the example code and not in the library.

ahocevar:
Thanks for the review @jahow. I was able to fix the sketch position problem you noticed post-review.

PR #11803, opened 2020-12-06
Modify fix vertex insertion

MoonE:
If there are multiple same segments it should only insert one vertex per geometry for all features.
If one feature had a duplicate segment this blocked the insertion of a vertex for other features because the loop was left.

I'm not quite sure if the correct fix is this or if the break should simply be removed.


ahocevar (approved):
Good find, this is exactly the right fix. Thanks, @MoonE. Please merge!

PR #11811, opened 2020-12-07
Better getPointResolution default when no transform available

mike-000:
The default getPointResolution function does not give a meaningful result when proj4 is not used and there is no EPSG:4326 transform defined (as can be seen from the scalebar in https://codesandbox.io/s/wild-resonance-9350t) as the default identity transform is only meaningful when both projections units are degrees.  This change is equivalent to having specified
`getPointResolution: function (resolution) { return resolution; }`
in the Projection options in that situation

Added a ScaleLine to the **WMS without Projection** example as it is now meaningful, although it may lack the 100% accuracy of the **Single Image WMS with Proj4js** or **Custom Tiled WMS** examples.

### DIFF
--- a/examples/wms-no-proj.js
+++ b/examples/wms-no-proj.js
@@ -4,6 +4,7 @@ import Projection from '../src/ol/proj/Projection.js';
 import TileWMS from '../src/ol/source/TileWMS.js';
 import View from '../src/ol/View.js';
 import {Image as ImageLayer, Tile as TileLayer} from '../src/ol/layer.js';
+import {ScaleLine, defaults as defaultControls} from '../src/ol/control.js';
 
 const layers = [
   new TileLayer({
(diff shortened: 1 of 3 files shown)


ahocevar (commented):
Not proceeding with transforms when `toEPSG4326` equals the `identityTransform` is a good change, but I wonder why the whole change would not be as simple as
```diff
--- a/src/ol/proj.js
+++ b/src/ol/proj.js
@@ -202,16 +202,20 @@ export function getPointResolution(projection, resolution, point, opt_units) {
     }
   } else {
     const units = projection.getUnits();
-    if ((units == Units.DEGREES && !opt_units) || opt_units == Units.DEGREES) {
+    const toEPSG4326 = getTransformFromProjections(
+      projection,
+      get('EPSG:4326')
+    );
+    if (
+      toEPSG4326 === identityTransform ||
+      (units == Units.DEGREES && !opt_units) ||
+      opt_units == Units.DEGREES
+    ) {
       pointResolution = resolution;
     } else {
       // Estimate point resolution by transforming the center pixel to EPSG:4326,
       // measuring its width and height on the normal sphere, and taking the
       // average of the width and height.
-      const toEPSG4326 = getTransformFromProjections(
-        projection,
-        get('EPSG:4326')
-      );
       let vertices = [
         point[0] - resolution / 2,
         point[1],
```
Especially since you said

> This change is equivalent to having specified
`getPointResolution: function (resolution) { return resolution; }`
in the Projection options in that situation

mike-000:
That would not handle the case where `opt_units` is not degrees but is different to the projection units.

ahocevar:
In this case I'd need tests to see what your intention and the expected results are.

ahocevar (approved):
Thanks, @mike-000

PR #11815, opened 2020-12-08
Remove unnecessary line that breaks drawing with multi-touch

ahocevar:
I accidentally added an unneeded line in #11800. It needs to be removed again, because otherwise an active Draw interaction during pinch-zooming will add a vertex on pointerup of one of the pinch fingers.


ahocevar:
I added a regression test to avoid surprises with this in the future.

marcjansen (approved):
Thanks, looks good to me! Please merge.

Issue #11816, opened 2020-12-09
Allow Select interaction to stop event propagation


greggian:
**Is your feature request related to a problem? Please describe.**
I think there are multiple use cases for the Select interaction to be able to stop propagation on the click event in some circumstances.

One use case I have in mind is when multiple Select interactions are active on the map. I would like to be able to give one Select interaction priority so that if it handles the click event, then it stops the click event from propagating so it doesn't 'fall through' to the other Select interaction(s).

**Describe the solution you'd like**
I think the ideal way to support this might be to update PluggableMap to respect MapBrowserEvent.stopPropagation/propagationStopped in handleMapBrowserEvent: 
https://github.com/openlayers/openlayers/blob/f3ad86e8e4808da510303aa637d4243a748eb076/src/ol/PluggableMap.js#L1031-L1034

That would allow listeners to "select" on a Select interaction to call stop propagation on the mapBrowserEvent:
```
let select = new Select();
select.on('select', function(e) {
     // stop downstream listeners from handling this event
     e.mapBrowserEvent.stopPropagation();
});
map.addInteraction(select);
```
Of course this would also allow other MapBrowserEvent listeners to also use stopPropagation, which I think would be expected behavior. I think this could make the event handling more consistent throughout.

If this sounds like a nice enhancement, I would be happy to submit a PR for it.

greggian:
For completeness, I'll mention... I also considered that we could have the Select interaction return the propagation value from dispatchEvent here:
https://github.com/openlayers/openlayers/blob/fb9c239d726cb002c4f97af2aceaf18b291f36da/src/ol/interaction/Select.js#L512-L522

Something like:
```
 if (selected.length > 0 || deselected.length > 0) { 
   const propagate = this.dispatchEvent( 
     new SelectEvent( 
       SelectEventType.SELECT, 
       selected, 
       deselected, 
       mapBrowserEvent 
     ) 
   );
   return propagate !== false;
 } 
 return true; 
```

This option would also fulfill the use case specifically for the Select interaction. I thought the changes in PluggableMap were the better option. it wasn't clear to me that stopping propagation on the Select event should stop the underlying mapBrowserEvent.

PR #11817, opened 2020-12-09
stopPropagation support for PluggableMap

greggian:
Update PluggableMap to respect MapBrowserEvent.propagationStopped in handleMapBrowserEvent. This allows Select interaction and others to support stop propagation from listeners.

Fixes openlayers/openlayers#11816

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1029,7 +1029,7 @@ class PluggableMap extends BaseObject {
           continue;
         }
         const cont = interaction.handleEvent(mapBrowserEvent);
-        if (!cont) {
+        if (!cont || mapBrowserEvent.propagationStopped) {
           break;
         }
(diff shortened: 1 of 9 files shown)


ahocevar (changes requested):
This is a good idea and a nice change. Thanks! There are probably a few more places where `event.preventDefault()` needs to be replaced with `event.originalEvent.preventDefault()`. A quick scan of the library reveals at least two places in `ol/interaction/Draw`. Could you please double-check if there are more, and change those accordingly?

greggian:
Thanks @ahocevar. I had missed some instances in Draw and MouseWheelZoom. I appreciate the pointer. I did a more complete check and believe I got all the appropriate cases now.

ahocevar (approved):
This looks good now. Thanks, @greggian!

Issue #11380, opened 2020-08-02
Please reopen #8313


cazacugmihai:
Hi!

Please reopen [#8313](https://github.com/openlayers/openlayers/issues/8313).

Thank you!

tschaub:
Reopened.

PR #11821, opened 2020-12-10
Call the finishCondition when drawing points / circles

### LINKED ISSUE #8313: finishCondition not triggered for some types
I've seen that when using `finishCondition` property in `ol/interaction/draw`, will not trigger the event for points and boxes (type `circle`). [See codepen](https://codepen.io/anon/pen/ERpLdm?&editors=1010).

I would like to have this feature for all the types.

An example use case:

I need to draw features, and check if intersects or it's contained by other feature.

BTW, I workaround this in `drawend` event, by checking if it intersects and removing the feature.

Why finishCondition is not called for these types?
It would be nice if finishCondition were called for these types in order to avoid this duplicated code.

Thanks in advance.

MoonE:
Fixes #8313, #11380

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -625,19 +625,22 @@ class Draw extends PointerInteraction {
       this.handlePointerMove_(event);
 
       if (this.shouldHandle_) {
-        if (!this.finishCoordinate_) {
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks for fixing this, @MoonE!

PR #11825, opened 2020-12-11
Add es6 modules to README

### LINKED ISSUE #11236: Build instructions for normal modules.
It seems all the build instructions are for various build tools, but I'm not using any for my project. I'd like some build instructions that allow me to use this with normal JS modules (in particular, inside web components).

Expected interface to import would be something like:
```
import("./ol.js").then((ol) => {
    ol.Map(...);
});
```

Trying this with the pre-built dist version doesn't work. I'd also like to explicitly choose what components should be included in the built version, in order to have a slim production build.

Any clues on how to build such a bundle would be great.

Dreamsorcerer:
Fixes #11236 

I've finally got working instructions for creating ES6 modules, and would like to propose adding them to the documentation.

First step would be to copy https://github.com/Dreamsorcerer/ol-parcel-es6 to https://github.com/openlayers/ol-parcel-es6

Then this link can be merged in.

Note there are some details that I'll update as upstream bugs are resolved (such as minify is broken, and we need nightly parcel until there's a beta-2 release).

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ See the following examples for more detail on bundling OpenLayers with your appl
  * Using [Webpack](https://github.com/openlayers/ol-webpack)
  * Using [Parcel](https://github.com/openlayers/ol-parcel)
  * Using [Browserify](https://github.com/openlayers/ol-browserify)
+ * Using [ES6 modules](https://github.com/openlayers/ol-parcel-es6)
 
 ## Sponsors
 


Dreamsorcerer:
Still waiting for a review...

tschaub:
@Dreamsorcerer - I don't think we'll include your example as it isn't significantly different than the existing ol-parcel example (it still uses Parcel to [generate a bundle](https://github.com/Dreamsorcerer/ol-parcel-es6/blob/e12eb357ab2c78eb6d68004a7d13326fd9bc92ad/package.json#L7)).

Dreamsorcerer:
Perhaps, but there are no instructions or anything in the parcel example (or any hint in the documentation that this is the tool that should be used) to generate this module. It took me atleast a couple of days work to figure out how to actually do this... I doubt others will want to repeat that process.

Dreamsorcerer:
Although it appears there's a section about ES modules being directly available now.
https://github.com/openlayers/openlayers/blob/main/README.md#es-modules

So, if that's the recommended approach, then we can leave it with that.

PR #11838, opened 2020-12-15
Custom circle render

### LINKED ISSUE #11837: Custom circle render
I need a gradient circle, but the circle feature style render option doesn't have a fire render function.

I implement this feature and use it like this:

```javascript
const circleFeature = new Feature({
  geometry: new Circle([12127398.797692968, 4063894.123105166], 50),
});
circleFeature.setStyle(
  new Style({
    renderer(coordinate, state) {
      console.log('This circle is rendered by the code below.', Date.now());
      const [x, y, x1, y1] = coordinate;
      const ctx = state.context;
      const dx = x1 - x;
      const dy = y1 - y;
      const radius = Math.sqrt(dx * dx + dy * dy);

      const innerRadius = 0;
      const outerRadius = radius * 1.4;

      const gradient = ctx.createRadialGradient(
        x,
        y,
        innerRadius,
        x,
        y,
        outerRadius
      );
      gradient.addColorStop(0, 'rgba(255,0,0,0)');
      gradient.addColorStop(0.6, 'rgba(255,0,0,0.2)');
      gradient.addColorStop(1, 'rgba(255,0,0,0.8)');
      ctx.beginPath();
      ctx.arc(x, y, radius, 0, 2 * Math.PI, true);
      ctx.fillStyle = gradient;
      ctx.fill();

      ctx.arc(x, y, radius, 0, 2 * Math.PI, true);
      ctx.strokeStyle = 'rgba(255,0,0,1)';
      ctx.stroke();
    },
  })
);
```

gif2

changqingom:
Fixes #11837

### DIFF
--- a/examples/custom-circle-render.html
+++ b/examples/custom-circle-render.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: Custom Circle Render
+shortdesc: Example of a custom circle render.
+docs: >
+  This example demonstrates the use of 'ol/style/Style' render option function to render circle feature.
+tags: "circle, feature, vector, render, custom"
+---
+<div id="map" class="map"></div>
--- a/examples/custom-circle-render.js
+++ b/examples/custom-circle-render.js
@@ -12,8 +12,6 @@ const circleFeature = new Feature({
 circleFeature.setStyle(
   new Style({
     renderer(coordinates, state) {
-      // eslint-disable-next-line no-console
-      console.log('This circle is rendered by the code below.', Date.now());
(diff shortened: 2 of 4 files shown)


ahocevar (changes requested):
Thanks for your contribution, @changqingom! Great idea to allow circles for custom rendering. However, the code could be simplified a fair bit. See c3dbb7162d. Feel free to cherry-pick if you agree.

Also, it would be good to remove the console.log from the example.

ahocevar (approved):
Thanks for this useful addition and the example, @changqingom!

Issue #11765, opened 2020-11-23
Add every other filter operators from Mapbox


Pipoupi:
**Is your feature request related to a problem? Please describe.**
I'm trying to implement a dynamic multiple filter and I need to have the IN operator implemented. 

Hi all, 

As I said I'm looking for an implementation of IN operator. I saw in some of your PRs that you're trying to be as close as possible to Mapbox spec, so I'm wondering why not every operators are implemented in Openlayers? Is it planned to be added or operators will stay like it is at this time ? https://github.com/openlayers/openlayers/blob/main/examples/webgl-points-layer.html

Following issue #11308 

Thanks !

mjjurkoic:
I've created a pull request which would add the in operator according to the Mapbox specs, it's PR #11839.

Pipoupi:
Thanks for your job and for noticing me. I'll close if PR is approved

mjjurkoic:
If this PR is approved I'll look into adding some of the other missing Mapbox operators as well.

PR #11839, opened 2020-12-15
Added 'in' operator as a valid WebGL expression.

mjjurkoic:
Added the 'in' operator for use in WebGL expressions to address issue #11765.

Code was written to follow Mapbox specifications, so the expression syntax is `['in', keyword, input]`, where `keyword` can be any type, and `input` is either an array or a string.  The expression returns true if `keyword` is contained within `input`.

Added an example to demonstrate the usage of the new operator.


examples/filter-with-mapbox-in-operator.js
```diff
@@ -0,0 +1,152 @@
...
+  for (let i = 1; i < 10; i++) {
+    yearArr.push(yearVal + i);
+  }
+  style.variables.decade = yearArr;
+  style.filter = ['in', ['get', 'year'], yearArr];
+  refreshLayer();
+  updateStatusText();
+}
+function updateStatusText() {
+  // eslint-disable-next-line no-console
```
tschaub, line 120: Looks like this could be removed.

src/ol/style/expressions.js
```diff
@@ -784,3 +785,58 @@ Operators['case'] = {
...
+    const haystack = [];
+    const needle = [];
+    let result;
+    if (getValueType(args[1]) & ValueTypes.NUMBER_ARRAY) {
+      const keyword = expressionToGlsl(context, args[0]);
+      const input = args[1];
+      assertNumbers(input);
+      let i = 0;
+      let inputArr = ['array'];
+      // @ts-ignore
```
tschaub, line 804: Are these `@ts-ignore` necessary?

src/ol/style/expressions.js
```diff
@@ -784,3 +785,58 @@ Operators['case'] = {
     return result;
   },
 };
+Operators['in'] = {
+  getReturnType: function (args) {
+    return ValueTypes.BOOLEAN;
+  },
+  toGlsl: function (context, args) {
```
tschaub, line 792: I see that there are limited tests in place already, but it would be good to start adding tests for things like this.

src/ol/style/expressions.js
mjjurkoic: I had them there because, if I don't include them, my IDE marks the `input.includes()` and the `input.forEach()` as errors.  It's because `input` can be one of a few different types.  The `if` statements and `assert` statements should insure that those functions are never invoked on a type that doesn't support them, and the code still runs fine without the `@ts-ignore`.

Is there a cleaner way to handle those errors?

src/ol/style/expressions.js
mjjurkoic: I added some unit tests for the 'in' expression.

src/ol/style/expressions.js
MoonE, line 804:
```suggestion
      const input = /** @type {Array<number>} */ (args[1]);
      let i = 0;
      /** @type {Array<string|number>} */
      let inputArr = ['array'];
```
You can typecast the variables instead of using ts-ignore.
The assert here is unnecessary because the getValueType function already checks the type.

src/ol/style/expressions.js
```diff
@@ -784,3 +785,58 @@ Operators['case'] = {
...
+      for (i = 0; i < haystack.length; i++) {
+        preliminary.push([`any(equal(${haystack[i]}, ${needle[i]}))`]);
+      }
+      result = preliminary.join(` || `);
+    } else if (getValueType(args[1]) & ValueTypes.STRING) {
+      const keyword = args[0];
+      const input = args[1];
+      assertString(input);
+      assertString(keyword);
+      // @ts-ignore
```
MoonE, line 832:
```suggestion
      assertString(args[0]);
      const keyword = /** @type {string} */ (args[0]);
      const input = /** @type {string} */ (args[1]);
```
Same as above, assertString is not necessary for input because it has already been checked and casting the variables gets rid of the ts warnings.

examples/filter-with-mapbox-in-operator.js
```diff
@@ -0,0 +1,151 @@
...
+
+function updateDecade() {
+  const yearVal = parseInt(decadeInput.value);
+  const yearArr = [yearVal];
+  for (let i = 1; i < 10; i++) {
+    yearArr.push(yearVal + i);
+  }
+  style.variables.decade = yearArr;
+  style.filter = ['in', ['get', 'year'], yearArr];
+  refreshLayer();
```
MoonE, line 116: Recreating the layer each time the filter changes is not a good use case for the in operator.
Maybe you can think of a better example use?

mjjurkoic: There's a good example case on the Mapbox website that shows a good use of the in operator.  I'll try to recreate that in OpenLayers.

mjjurkoic:
I changed the example to avoid destroying and recreating the layer in order to update the filter.  However, I can't figure out how to make the layer actually apply the new filter.  Any tips on how to force the filter to apply to the WebGLPointsLayer?

mjjurkoic:
I rebased my fork onto the current OpenLayers main branch.

MoonE:
As I understand it mapbox-gl styles allow for needle and haystack to be variables `['get', '...']` that access feature data. Do you have know how it is implemented in mapbox? Do they evaluate the condition in javascript?

As it is now the string version isn't really useful as neither parameter can be a variable so it evaluates to a constant boolean for all features.

I also don't see a way to update the style a webglpoints layer in its current state, apart from recreating it.
Maybe @jahow could share some insight here?

mjjurkoic:
Mapbox-gl implements its style expressions quite differently.  It evaluates everything in the context of JavaScript objects, storing all.feature variables and such in instances of custom classes, while OpenLayers translates everything to glsl and does the evaluation when the outputted glsl code is run.

As my implementation stands currently, the needle variable can be obtained with `['get', ...]` and it will work fine.  I should be able to change the haystack variable to work with feature variables, but it would be limited to arrays of length 4 or shorter, since that's currently the longest allowed by an `['array', ...]` expression, and I couldn't figure out a good way to access arrays in the style variables.  Using `['var', ...]` returns a variable name in glsl, so I can't really access the original array directly.

Regarding strings, glsl doesn't support string operations, so I was limited to what could be done in the immediate JavaScript context.  String values can't be assigned as variables in glsl, so the only strings that would be encountered would be provided directly in the style expression.

I hope that helps explain things, and I'd appreciate any suggestions on how to improve this PR.

jahow:
Thanks @mjjurkoic for working on this. Indeed, in Mapbox-gl-js style expressions are evaluated in Javascript, whereas in OpenLayers they are translated to GLSL code. This comes with several limitations:
* strings are treated as tokens and cannot be transformed or compared other than strict equality
* arrays of variable length other than 2, 3 or 4 cannot be handled

There may be workarounds for these limitations but I personally haven't been able to come up with some.

Regarding the `in` operator, implementing it in OpenLayers is quite hard because of the above limitations. Specifying an array of variable length in the `filter` is not possible for example. Filters _can only be literal values, not expressions_. Remember, they are fed to the shader through uniforms, which are numerical values. Similarly, something like `['in', 'foo', 'foobar']` will not work either. All things considered, we're pretty much back to the `match` operator (comparing an input with a fixed amount of values).

To fit the use case showed in your example, you could either:
* Use a filter based on variable minX/maxX/minY/maxY values, since you're drawing a box (admittedly, this isn't super satisfying), or
* Recreate the layer with a style where expressions are dynamically generated; note that recreating a layer sounds cumbersome but is actually quite fast since the data source does not change; it can cause a flicker on the map though

Regarding the current implementation of the WebGL layer in OL, I think it's reasonable to keep challenging its limitations and leave the door open for a refactoring to remove them. This is why the feature is still experimental!

jahow:
> I also don't see a way to update the style a webglpoints layer in its current state, apart from recreating it.

There is currently no other way indeed. Being able to change the style on the fly would be awesome and probably reduce the need for having all the missing Mapbox operators implemented.

mjjurkoic:
@jahow Thanks for the feedback. I'm going to run with that idea of having a variable number of inputs, and I'll rework the example a bit to give a good idea of how to use that.

Am I correct in thinking that style variables (ones that an expression can access with `['var', name]`) can only be numbers? There doesn't currently seem to be a way to store and access arrays or strings via expressions.

If the input array could be stored as a variable, it wouldn't be necessary to recreate the layer when the input array changes, I don't think. To that end, I'm mulling over a way to rework the way expressions are integrated with the GLSL fragment builder that would allow more flexibility with things like arrays. If I make any headway I'll make a new PR for that.

tschaub:
Even if we have a current limitation around the number of elements that would be allowed, I think it makes sense to try to stick to the Mapbox style spec for now.  That would allow us to update the implementation and remove current limitations without requiring that people change their style objects.

mjjurkoic:
@tschaub The major choke point right now is the limitation on the number of elements allowed in an array expression. The reason for that limitation is that array expressions are implemented in GLSL as vecs, and the reason they are implemented as vecs is because the expression has to be translated into a single line of GLSL code, that is, there is no room for the loop logic that would be necessary to handle GLSL arrays.

I think a potential solution could be to change where in the GLSL code output the expressions evaluate to. Currently the Mapbox-style expressions, after they are evaluated to a GLSL string, get dropped into line 400 in the WebGL [ShaderBuilder](https://github.com/openlayers/openlayers/blob/main/src/ol/webgl/ShaderBuilder.js). The `this.discardExpression` property in that line could be replaced with a GLSL variable, so the line might look something like `if (discardVertex) { discard; }`. Then `discardVertex` could be declared earlier in the GLSL code, which would allow implementations of Mapbox-style expressions that could cover several lines, which would give enough space and flexibility for the loop logic necessary to deal with true arrays. Then array expressions could allow more than 4 elements.

I can make a new issue to continue the discussion about refactoring, if necessary, since it is somewhat outside the scope of this PR.

tschaub:
In case it wasn't clear, I was suggesting that we try to conform with the spec even if the current implementation cannot handle arbitrarily sized arrays.

So instead of `['in', ['get', 'name'], 'Madrid', 'Berlin', 'London']` we would take `['in', ['get', 'name'], ['Madrid', 'Berlin', 'London']]` and we could throw if the length were different than 2, 3, or 4.

mjjurkoic:
Yes I understood that.  Sorry if there was any confusion.

I don't think the current array expression implementation supports strings though, does it?

mjjurkoic:
I reworked the expression and tests to work with the current array size restrictions of 2, 3, or 4.

Currently, I'm not sure if we can make a meaningful string version of the 'in' operator, since object attributes and style variables can't really be strings, because GLSL has no support for string operations.  Thoughts?

PR #11844, opened 2020-12-19
Style expressions error message

MoonE:
The error message for the `assertArgsOdd` function was the same as for `asserArgsEven` 
Also only call `getValueType` once in `expressionToGlsl` instead of in each if condition.


ahocevar (approved):
Thanks, @MoonE!

PR #11845, opened 2020-12-19
Call finishCondition for all types for Draw interaction

MoonE:
This time for real ... with tests.
 #11821

I added an additional commit that makes the draw interaction immediately throws an error if it is created with an invalid type.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -1227,23 +1227,21 @@ export function createBox() {
  * @return {Mode} Drawing mode.
  */
 function getMode(type) {
-  let mode;
-  if (type === GeometryType.POINT || type === GeometryType.MULTI_POINT) {
-    mode = Mode.POINT;
-  } else if (
-    type === GeometryType.LINE_STRING ||
-    type === GeometryType.MULTI_LINE_STRING
-  ) {
-    mode = Mode.LINE_STRING;
-  } else if (
-    type === GeometryType.POLYGON ||
-    type === GeometryType.MULTI_POLYGON
-  ) {
-    mode = Mode.POLYGON;
-  } else if (type === GeometryType.CIRCLE) {
-    mode = Mode.CIRCLE;
+  switch (type) {
+    case GeometryType.POINT:
+    case GeometryType.MULTI_POINT:
+      return Mode.POINT;
+    case GeometryType.LINE_STRING:
+    case GeometryType.MULTI_LINE_STRING:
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Draw.js
```diff
@@ -632,9 +632,10 @@ class Draw extends PointerInteraction {
               break;
             }
           // eslint-disable-next-line no-fallthrough
-          case this.freehand_ ||
-            (this.atFinish_(event.pixel) && this.finishCondition_(event)):
-            this.finishDrawing();
+          case this.freehand_ || this.atFinish_(event.pixel):
+            if (this.freehand_ || this.finishCondition_(event)) {
+              this.finishDrawing();
+            }
```
ahocevar, line 638: When reading this change, it is not obvious at first glance why the previous version did not work. Instead of falling through from the previous `case`, wouldn't it be possible to get rid of this `case`? I have not tested the suggestion below, but I think something like this could work:
```suggestion
            if (this.freehand_ ||
             (this.atFinish_(event.pixel) && this.finishCondition_(event))) {
              this.finishDrawing();
            }
```

ahocevar (approved):
Thanks @MoonE for revisiting this, and for the additional tests.

MoonE:
Thanks for your review, I had another look at the changes and realized that my code did also not quite behave as expected as for Points the atFinish method is not being called. Not really a problem, because it would return true every time in that case. 

Trying to fix it brought me more or less back to the orignal solution:
https://github.com/openlayers/openlayers/blob/6cfe9148411753d04a256c63cbfcbbdabb8e92c3/src/ol/interaction/Draw.js#L628-L641

The solution with the switch would have been more complicated:
```js
        switch (true) {
          case !this.finishCoordinate_:
            this.startDrawing_(event.coordinate);
            if (this.mode_ !== Mode.POINT) {
              break;
            }
          // eslint-disable-next-line no-fallthrough
          case this.freehand_ || this.atFinish_(event.pixel):
            if (
              this.freehand_ ||
              (this.mode_ !== Mode.POINT &&
                this.atFinish_(event.pixel) &&
                this.finishCondition_(event))
            ) {
              this.finishDrawing();
            }
            break;
          case !this.freehand_:
            this.addToDrawing_(event.coordinate);
            break;
          default:
            break;
        }
```

MoonE:
Apologies for the delay.
This should work now.

ahocevar (approved):
Thanks, @MoonE!

Issue #11847, opened 2020-12-20
MapboxVectorLayer does not work with non-hosted Mapbox compatible styles


mike-000:
**Describe the bug**
There seems no obvious reason why MapboxVectorLayer should not be used with other compatible styles such as hosted by MapTiler. However if not a `mapbox://` path  https://github.com/openlayers/openlayers/blob/main/src/ol/layer/MapboxVector.js#L101 uses the TileJSON url as the OpenLayers source's tile url, resulting in an invalid pbf error when attempting to use the layer (which was created without errors).

**Expected behavior**
If not a `mapbox://` path load the TileJSON and use its tile url(s) as in ol-mapbox-style's `setupVectorLayer`.

ahocevar:
Yes, this is currently a limitation. To properly fix that, TileJSON parsing would have to be made available separate from `ol/layer/TileJSON` (which is only for raster tiles), e.g. in an `ol/format/TileJSON`. If someone wants to take that on, it would be a nice addition.

In the meantime, you could use [ol-mapbox-style](https://npmjs.com/package/ol-mapbox-style), which is used by `ol/layer/MapboxVector` anyway, directly. Something like this:
```js
import Map from `ol/Map`;
import addLayers from `ol-mapbox-style`;

const map = new Map({
  // map configuration, layers etc.
});
addLayers(map, 'path/to/your/style.json').then(() => {
  // work with the Mapbox layers that are now added to the map`s `layers` collection.
});

mike-000:
Yes, there is certainly much more which could be done to improve TileJSON both for raster and vector. Some of the MapTiler examples hard code custom tile grids, but they could be obtained from the JSON in the same way as in this Ordnance Survey example https://github.com/OrdnanceSurvey/OS-Data-Hub-API-Demos/blob/master/OSVectorTileAPI/OpenLayers/map.js#L52-L71  Mapbox however simply assumes a standard EPSG:3857 grid, so for mapbox-gl.js compatible sources it should be sufficient to fetch the TileJSON url and use its `tiles` property to set the VectorTileSource's url(s).

PR #11856, opened 2020-12-21
Use tilegrid extent as default attributionExtent for TileJSON

### LINKED ISSUE #11846: TileJSON uses wrong projection extent for attributions
**Describe the bug**
If a TileJSON does not specify bounds the EPSG:4326 extent is used instead, regardless of the source extent.  In the case of an EPSG:3857 source this is a 360 x 180 meter box around the "null island".

**To Reproduce**
In https://codesandbox.io/s/sharp-sammet-fmsr2 pan or zoom to the area of coverage and see the attributions control disappear when null island is outside the frame.

**Expected behavior**
In https://github.com/openlayers/openlayers/blob/main/src/ol/source/TileJSON.js#L188 if there is no extent available from the bounds the tilegrid extent would make a better alternative.  Also, while not applicable to vector tiles, for raster tiles which can be reprojected the extent used in the subsequent `intersects` call should be transformed to the view projection if different.

mike-000:
Fixes #11846

Use a more appropriate default extent.

This does not attempt to handle reprojection issues mentioned in #11846 as transforms involving global extents and local projections could be invalid, and conflicts between reprojection and attributions also seem to affect BingMaps sources, so these would be best dealt with as a separate issue.

### DIFF
--- a/src/ol/source/TileJSON.js
+++ b/src/ol/source/TileJSON.js
@@ -171,10 +171,11 @@ class TileJSON extends TileImage {
       extent = applyTransform(tileJSON['bounds'], transform);
     }
 
+    const gridExtent = extentFromProjection(sourceProjection);
     const minZoom = tileJSON['minzoom'] || 0;
     const maxZoom = tileJSON['maxzoom'] || 22;
     const tileGrid = createXYZ({
-      extent: extentFromProjection(sourceProjection),
+      extent: gridExtent,
       maxZoom: maxZoom,
       minZoom: minZoom,
       tileSize: this.tileSize_,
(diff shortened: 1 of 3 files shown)


mike-000:
Ready for review, the failed test seems unrelated.

ahocevar (commented):
It would be nice to have a test for this.

mike-000:
The same unrelated failed test, but ready for review again.

I noticed the code was duplicated in `ol/source/UTFGrid` so I made the same change there. As noted in #11847 a common method (which could handle the Esri and MapTiler extensions to the Mapbox TileJSON specification for non-standard tile grids) would definitely be a good idea.  A new format might not be needed, an `optionsFromTileJSON` method might be sufficient.

ahocevar:
A static `optionsFromTileJSON()` function in `ol/source/TileJSON` would be good to avoid duplicated code in `ol/source/UTFGrid`. And yes, it could also handle Esri and MapTiler extensions. @mike-000, if you'd be willing to take that on, it would be appreciated if you'd create a separate pull request for that.

ahocevar (approved):
Thanks, @mike-000

PR #11858, opened 2020-12-22
Correct documented event names for VectorSourceEvent

MoonE:
An `s`is missing in the event name documentation.


ahocevar (approved):
Thanks, @MoonE!

PR #11859, opened 2020-12-24
Do not fire duplicate postrender events

MoonE:
Multiple `rendercomplete` may be fired when the `setTimout` for `handlePostRender` does not run immediatly between two animation frames or `renderSync` is called.

Instead of the order of function calls for `PluggableMap` being 
`renderFrame_` -> `handlePostRender` -> `renderFrame_` -> `handlePostRender` 
it may be 
`renderFrame_` -> `renderFrame_` -> `handlePostRender` -> `handlePostRender`

With this PR no more than one `handlePostRender` will occur between any two `renderFrame_` calls.


ahocevar (approved):
A good improvement, thanks!

PR #11863, opened 2020-12-25
update to latest jsPDF in Export PDF example

jipexu:
point to the latest version and need to change the name of the global variable ... ?

it's seems work for me please re-check ...


tschaub:
@jipexu - you’ll need to base your branch on a more recent commit to get the above failing security check to pass.

You can do this in a number of ways.  Here is one:
```
git pull --rebase openlayers main
git push --force origin main
```

This assumes you have named this remote `openlayers` and your own remote `origin`.

jipexu:
@tschaub thanks ! Definitely I still have a lot of trouble to work effectively with this tool ... I always forget something ... Well, I am trying to do something cleaner.

mike-000:
It is usually good practice to use a defined stable version and update periodically to new versions after checking compatibility.  `latest` links can change at any time and break applications - for example if the latest `mapbox-gl.js` was used in the Mapbox-gl Layer example a Mapbox API key would now be needed in addition to the MapTiler key, and the latest `jszip` would not work in the KMZ Drag-and-Drop example.

ahocevar:
Another (maybe better) option would be to add jspdf to devDependencies (`npm install --save-dev jspdf`) and `import jspdf from 'jspdf'` instead of using the cdn version.

jipexu:
> It is usually good practice to use a defined stable version and update periodically to new versions after checking compatibility. `latest` links can change at any time and break applications - for example if the latest `mapbox-gl.js` was used in the Mapbox-gl Layer example a Mapbox API key would now be needed in addition to the MapTiler key, and the latest `jszip` would not work in the KMZ Drag-and-Drop example.

@mike-000 thanks you are right  i hesitated to point the latest or the 2.2.0 ... so point to the 2.2.0 is better

jipexu:
> Another (maybe better) option would be to add jspdf to devDependencies (`npm install --save-dev jspdf`) and `import jspdf from 'jspdf'` instead of using the cdn version.

@ahocevar  sorry but this is  much more than my technical level ... ;-(  so i can not help more for this option

jipexu:
close by #12417  thanks @ahocevar

PR #11865, opened 2020-12-27
Use ol/array#includes instead of Array.prototype.includes()

mike-000:
Fixes browser compatibility in the IIIF Image API example which currently doesn't work with IE and perhaps some other old browsers.

### DIFF
--- a/src/ol/format/IIIFInfo.js
+++ b/src/ol/format/IIIFInfo.js
@@ -3,6 +3,7 @@
  */
 
 import {assert} from '../asserts.js';
+import {includes} from '../array.js';
 
 /**
  * @typedef {Object} PreferredOptions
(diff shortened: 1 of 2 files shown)


MoonE:
Please also change the uses of includes in `ol/source/IIIF.js`.

mike-000:
> Please also change the uses of includes in `ol/source/IIIF.js`.
> 
> Maybe it would be reasonable to add `includes` / `find` / `findIndex` to the list of required polyfills instead?

If those (and all other built-in polyfill) methods exists they should be used consistently, not duplicated.  I noticed the problem with `ol/format/IIIFInfo` on the v6.4.3 example but a Mapbox update now means none of the dev or 6.5.0 examples currently work with IE so I was unable to test if further changes were needed.

MoonE:
The examples should work again in ie11 with #11871.
Please rebase on the latest version in the main branch.

MoonE:
You are of course correct that the available methods should be used consistently. 
Though in my optinion it would make sense to rely on polyfills for these functions instead as it is a bit of a hassle for every contributor to remember which functions should not be used.

MoonE (approved):
Thanks, @mike-000.

PR #11871, opened 2020-12-29
Fix examples build for ie11

MoonE:
The examples work now with ie11 and the spread operator is gone from the legacy build.

I added `Number.isInteger` to the polyfills because that function is not available in ie11.

Building the examples takes twice as long now and I have to say that I am not at all an webpack expert...

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ For older browsers and platforms (Internet Explorer, Android 4.x, iOS v12 and ol
 * [`element.prototype.classList` (`add`/`remove`)](https://caniuse.com/#feat=classlist): Available from [polyfill.io](https://polyfill.io/).
 * [`URL` API](https://caniuse.com/#feat=url): Available from [polyfill.io](https://polyfill.io/).
 * [`TextDecoder`](https://caniuse.com/#feat=textencoder): Available from [polyfill.io](https://polyfill.io/).
+* [`Number.isInteger`](https://caniuse.com/#feat=isInteger): Available from [polyfill.io](https://polyfill.io/).
 * [Pointer events](https://caniuse.com/#feat=pointer): Use [elm-pep](https://npmjs.com/package/elm-pep) (lightweight) or [pepjs](https://npmjs.com/package/pepjs) (for really, really old browsers).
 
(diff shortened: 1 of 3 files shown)


ahocevar (commented):
buble, which is currently used for transpiling examples, can handle the spread operator, and does so for our own library code already. So instead of adding a babel-loader for everything in node_modules, it should be enough to include the dependency that causes the problem. Something like this:
```diff
--- a/examples/webpack/config.js
+++ b/examples/webpack/config.js
@@ -31,6 +31,14 @@ module.exports = {
         include: [
           path.join(__dirname, '..', '..', 'src'),
           path.join(__dirname, '..'),
+          path.join(
+            __dirname,
+            '..',
+            '..',
+            'node_modules',
+            '@mapbox',
+            'mapbox-gl-style-spec'
+          ),
         ],
       },
       {
```

MoonE:
Buble complains about for ... of loops which it may not handle properly:

> ERROR in ../node_modules/@mapbox/mapbox-gl-style-spec/dist/index.es.js
> Module build failed (from ../node_modules/buble-loader/index.js):
> BubleLoaderError: 
> 7760 :         callback(style.sources[k]);
> 7761 :     }
> 7762 : }
> 7763 : function eachLayer(style, callback) {
> 7764 :     for (const layer of style.layers) {
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> Transforming for-of statements is not fully supported. Use `transforms: { forOf: false }` to skip transformation and disable this error, or `transforms: { dangerousForOf: true }` if you know what you're doing. (7764:4)
>     at Object.module.exports (/home/moon/Programming/openlayers/node_modules/buble-loader/index.js:24:69)

ahocevar:
@MoonE The for..of loop that buble complains about should be safe, so the `{transforms: {dangerousForOf: true}}` option can be enabled. I did the same for the legacy build in #11872.

ahocevar (approved):
Thanks, @MoonE

PR #11872, opened 2020-12-29
Legacy build for legacy browsers

ahocevar:
Makes the legacy build work in IE11.


MoonE (approved):
Works for me on ie11 with a very simple map.

PR #11873, opened 2020-12-30
Update Mapbox-gl Layer example to mapbox-gl-js version 1.13.0 and add note about why an old version is used

mike-000:
mapbox-gl-js version 2.x is definitely not compatible with IE and also requires a Mapbox access token even when the source is not hosted by Mapbox, so I have updated to the final compatible version with a note about why the version is used.

### DIFF
--- a/examples/mapbox-layer.html
+++ b/examples/mapbox-layer.html
@@ -3,12 +3,14 @@ layout: example.html
 title: Mapbox-gl Layer
 shortdesc: Example of a Mapbox-gl-js layer integration.
 docs: >
-  Show how to add a mapbox-gl-js layer in an openlayers map. **Note**: Make sure to get your own API key at https://www.maptiler.com/cloud/ when using this example. No map will be visible when the API key has expired.
+  Show how to add a mapbox-gl-js layer in an OpenLayers map.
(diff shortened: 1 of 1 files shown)


MoonE:
Why do you think mapbox-gl-js v1.13 does not work with ie11? Seems to work when I use it.

mike-000:
> Why do you think mapbox-gl-js v1.13 does not work with ie11? Seems to work when I use it.

Different application but using the same method to put a mapbox-gl-js map inside an OpenLayers map, with version 1.12.0 it works on all browsers, but IE throws an error with version 1.13.0 while the same code works with Chrome (as it does with version 2.x with an Mapbox access token).

image

MoonE:
Looks like mapbox-gl-js will only develop the v2 branch, but there's fork at https://github.com/maplibre/maplibre-gl-js. 
Would it make sense to switch to that? If you report the issue there they might be able to fix it.

mike-000:
My problem with object urls is not relevant to this "normal use" example so I have updated to v1.13.0.  Using Maplibre may be worth considering in the future if it becomes a popular alternative.

MoonE (approved):
Makes sense. Thanks @mike-000.

PR #11874, opened 2020-12-30
Hit lines even if they are dashed

### LINKED ISSUE #11805: Dashed line hit detection?
Hi, 

What if my map's primary objects are dashed lines? Is there a nice hit detection solution? If I click on that empty area between dashes of a line I want a hit for that line object to occur. It seems the hit doesn't occur now...

I have figured out one solution to have "fake" duplicate line symbol with 0.01 color opacity underneath a dashed line, but the problem is that I can see that ghost line with my eyes... And it seems I can't set lower opacity like 0.001

Any suggestions, guys? :)

tschaub:
I think the expected hit detection behavior is to consider lines hit even if they are dashed.

This change makes it so we ignore any line dash (as we do with color) when doing hit detection in both `map.forEachFeatureAtPixel()` and `layer.getFeatures()`.  The new tests fail without the source changes.

Fixes #11805.

### DIFF
--- a/src/ol/render/canvas/LineStringBuilder.js
+++ b/src/ol/render/canvas/LineStringBuilder.js
@@ -6,6 +6,7 @@ import CanvasInstruction, {
   beginPathInstruction,
   strokeInstruction,
 } from './Instruction.js';
+import {defaultLineDash, defaultLineDashOffset} from '../canvas.js';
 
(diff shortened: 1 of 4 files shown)


ahocevar (approved):
Makes sense! Thanks, @tschaub.

PR #11886, opened 2021-01-05
Throw an error if shaders fail to compile or program fails to link

tschaub:
Currently, if GL shader compilation or program linking fails, nothing happens (and the user would have to figure out that they need to add a call to `layer.getShaderCompileErrors()` and log the output).  Instead, we should fail early and in an obvious way when either shader compilation fails or program linking fails.

### DIFF
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -709,8 +709,7 @@ class WebGLHelper extends Disposable {
   }
 
   /**
-   * Create a program for a vertex and fragment shader. The shaders compilation may have failed:
-   * use `WebGLHelper.getShaderCompileErrors()`to have details if any.
+   * Create a program for a vertex and fragment shader.  Throws if shader compilation fails.
    * @param {string} fragmentShaderSource Fragment shader source.
    * @param {string} vertexShaderSource Vertex shader source.
    * @return {WebGLProgram} Program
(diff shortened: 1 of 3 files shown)


src/ol/webgl/Helper.js
```diff
@@ -723,39 +722,41 @@ class WebGLHelper extends Disposable {
...
       )}`;
+      throw new Error(message);
     }
-    if (gl.getShaderInfoLog(vertexShader)) {
-      this.shaderCompileErrors_ =
-        (this.shaderCompileErrors_ || '') +
-        `Vertex shader compilation failed:\n${gl.getShaderInfoLog(
-          vertexShader
-        )}`;
+    gl.deleteShader(fragmentShader);
```
tschaub, line 742: Nothing else has a reference to this shader, so we can free up resources.

src/ol/webgl/Helper.js
```diff
@@ -723,39 +722,41 @@ class WebGLHelper extends Disposable {
...
+
+    if (!gl.getShaderParameter(vertexShader, gl.COMPILE_STATUS)) {
+      const message = `Vertex shader compilation failed: ${gl.getShaderInfoLog(
+        vertexShader
+      )}`;
+      throw new Error(message);
     }
+    gl.deleteShader(vertexShader);
```
tschaub, line 750: Same as above.

test/spec/ol/webgl/helper.test.js
```diff
@@ -354,7 +358,7 @@ describe('ol.webgl.WebGLHelper', function () {
         uniform float u_test;
 
         void main(void) {
-          gl_Position = vec4(u_test, a_test, 0.0, 1.0);
+          gl_Position = vec4(u_test, attr3, 0.0, 1.0);
```
tschaub, line 361: `a_test` was an undeclared identifier in the shader used in this test.  This now throws an error.

ahocevar (approved):
Good usability improvement and cleanup, thanks @tschaub!

jahow:
Hey, sorry to be late on this! This was something I also considered, but what happened was that Firefox would sometimes report compilation "errors" even on a working shader. See: https://github.com/openlayers/openlayers/pull/10261

I've checked again, and it seems to work fine in Firefox now: https://deploy-preview-11886--ol-site.netlify.app/examples/heatmap-earthquakes.html

So, all good! Thanks @tschaub!

PR #11890, opened 2021-01-07
Fix WFS writeFilter

walkermatt:
Following an upgrade to ol v6.5.0 we (@AstunTechnology) had a number of tests fail in a project due to the addition of the mandatory `version` argument to `ol/format/WFS~writeFilter`. While investigating I found that `writeFilter` doesn't currently support writing WFS 2.0.0 filters due to the wrong namespace being assigned to the filter child node.

This PR:
* Adds support for writing WFS 2.0.0 filters
* Makes the `version` argument optional, defaulting to WFS 1.1.0
* Adds tests for writing a WFS 2.0.0 filter and calling `writeFilter` without a `opt_version` argument

The reason for making the `version` argument optional is to avoid v6.5.x introducing a breaking change to the existing API which I think would strictly trigger a major version release according to semver rules.

### DIFF
--- a/src/ol/format/WFS.js
+++ b/src/ol/format/WFS.js
@@ -1328,7 +1328,7 @@ function writeTimeInstant(node, time) {
  * @api
  */
 export function writeFilter(filter, version) {
-  const child = createElementNS(OGCNS[version], 'Filter');
+  const child = createElementNS(getFilterNS(version), 'Filter');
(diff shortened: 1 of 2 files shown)


tschaub (approved):

tschaub:
Thanks for the fix and additional tests, @walkermatt.

Issue #11860, opened 2020-12-24
Style clone function doesn't copy the renderer function


changqingom:
I find this:

The new style options don't have a renderer property.

https://github.com/openlayers/openlayers/blob/1a952e7a41a030791780af6086ca4ad91e30e6f0/src/ol/style/Style.js#L213-L226

mjjurkoic:
Someone correct me if I'm wrong, but wouldn't this require a deep copy of the renderer property?

The obvious solution, of course, would be to just add `renderer: this.getRenderer() ? this.getRenderer() : undefined,` into the properties of the new Style, but that would create issues because it would only pass the renderer function by reference.

To properly clone the renderer function, I think a deep copy of the renderer function object would be needed.  Core JavaScript doesn't provide a good way of creating a deep copy of an object that contains functions, so a true solution would require either a custom deep copy function to be added to OpenLayers, or the use of a third-party library that provides a deep copy function.  A third-party library would add another dependency, since I don't believe any of the current OpenLayers dependencies provide deep copy functions.

changqingom:
I've wondered before whether a deep copy of a function is necessary.If a function is defined and used according to the standard specification, there is no need for a deep copy.I understand that many circumstances need to be considered for the convenience of deep copy of the function. In this case, I suggest some supplementary instructions on style.clone in the API documentation to make it clear that no copy will be carried out for the renderer function.

mike-000:
There seems to be no attempt to deep clone functions in the library.  If `geometry` is a function there is also no deep clone.  The `style` of a feature is not deep cloned, regardless of whether it is a function.

mjjurkoic:
In that case, taking a cue from how `clone()` handles the geometry property, we should be able to do `renderer: this.getRenderer()` to get the same kind of copy as geometry would get if it were a function.

Would that be an appropriate fix for this issue, or would it be better to leave `clone()` as is and simply state in the API that the renderer function is not copied?

mjjurkoic:
I created PR #11893 to move the issue forward.

PR #11893, opened 2021-01-08
Pass the renderer function to the cloned style

mjjurkoic:
Addresses Issue #11860.

Makes Style.clone() more thorough by passing the renderer function to the new Style object created by the clone function.

### DIFF
--- a/src/ol/style/Style.js
+++ b/src/ol/style/Style.js
@@ -219,6 +219,7 @@ class Style {
       geometry: geometry,
       fill: this.getFill() ? this.getFill().clone() : undefined,
       image: this.getImage() ? this.getImage().clone() : undefined,
+      renderer: this.getRenderer() ? this.getRenderer() : undefined,
       stroke: this.getStroke() ? this.getStroke().clone() : undefined,
       text: this.getText() ? this.getText().clone() : undefined,
       zIndex: this.getZIndex(),
--- a/src/ol/style/Style.js
+++ b/src/ol/style/Style.js
@@ -219,7 +219,7 @@ class Style {
       geometry: geometry,
       fill: this.getFill() ? this.getFill().clone() : undefined,
       image: this.getImage() ? this.getImage().clone() : undefined,
-      renderer: this.getRenderer() ? this.getRenderer() : undefined,
+      renderer: this.getRenderer(),
       stroke: this.getStroke() ? this.getStroke().clone() : undefined,
       text: this.getText() ? this.getText().clone() : undefined,
       zIndex: this.getZIndex(),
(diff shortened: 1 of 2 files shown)


src/ol/style/Style.js
```diff
@@ -219,6 +219,7 @@ class Style {
       geometry: geometry,
       fill: this.getFill() ? this.getFill().clone() : undefined,
       image: this.getImage() ? this.getImage().clone() : undefined,
+      renderer: this.getRenderer() ? this.getRenderer() : undefined,
```
sbrunner, line 222: Same as
```suggestion
      renderer: this.getRenderer(),
```

simonseyock:
It would be nice to have a test for this if it should get merged. See https://github.com/openlayers/openlayers/blob/main/test/spec/ol/style/style.test.js

test/spec/ol/style/style.test.js
```diff
@@ -91,12 +102,16 @@ describe('ol.style.Style', function () {
       expect(original.getGeometry()).not.to.be(clone.getGeometry());
       expect(original.getFill()).not.to.be(clone.getFill());
       expect(original.getImage()).not.to.be(clone.getImage());
+      expect(original.getRenderer()).not.to.be(clone.getRenderer());
```
mjjurkoic, line 105: This test currently fails, since both the original and the clone reference the same renderer function.  This is expected in the absence of a deep clone, as was discussed in the issue #11860 thread.  Should I change line 105 to `expect(original.getRenderer()).to.be(clone.getRenderer());`, or delete that test entirely?

simonseyock: I would remove this line.

test/spec/ol/style/style.test.js
```diff
@@ -108,6 +123,9 @@ describe('ol.style.Style', function () {
       expect(original.getImage().getScale()).not.to.eql(
         clone.getImage().getScale()
       );
+      expect(original.getRenderer().toString()).not.to.eql(
+        clone.getRenderer().toString()
+      );
```
simonseyock, line 128: Here I think the string representation could theoretically be the same, but the reference should be different. In this specific case it is indeed different. But I would use the reference nevertheless.

test/spec/ol/style/style.test.js
```diff
@@ -64,6 +68,9 @@ describe('ol.style.Style', function () {
       expect(original.getImage().getRadius()).to.eql(
         clone.getImage().getRadius()
       );
+      expect(original.getRenderer().toString()).to.eql(
+        clone.getRenderer().toString()
+      );
```
simonseyock, line 73: I would assume they are not only identical by string association but also by reference. So in my opinion
```
expect(original.getRenderer()).to.eql(clone.getRenderer());
```
would be correct.

simonseyock (approved):

ahocevar (approved):
Thaks @mjjurkoic and @simonseyock

PR #11899, opened 2021-01-10
Update HERE Map Tile API example to use API key

### LINKED ISSUE #11895: HERE Map Tile API example not working
**Describe the bug**
The app_id app_code pair seems to be out of date.

**To Reproduce**
See the example and errors on the console.

**Expected behavior**
Up to date app_id app_code, or better to generate an apiKey which would require the url to be updated to a new path as in the latest HERE documentation

```
  'https://{1-4}.{base}.maps.ls.hereapi.com' +
  '/{type}/2.1/maptile/newest/{scheme}/{z}/{x}/{y}/256/png' +
  '?apiKey={apiKey}';.
```

mike-000:
Fixes #11895

The current app_id app_code pair are not working and new HERE accounts use API keys instead of key pairs.

### DIFF
--- a/examples/here-maps.html
+++ b/examples/here-maps.html
@@ -3,14 +3,12 @@ layout: example.html
 title: HERE Map Tile API
 shortdesc: Example of a map with map tiles from HERE.
 docs: >
-  <p><a href="https://developer.here.com/rest-apis/documentation/enterprise-map-tile">HERE Map Tile API</a> used with <code>ol/source/XYZ</code>.</p>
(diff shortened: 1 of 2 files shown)


ahocevar:
Where did you obtain the API key? Will it expire?

mike-000:
> Where did you obtain the API key? Will it expire?

I used a webmail address to create a free account (250K transactions per month) specifically for this example.
The documentation https://developer.here.com/documentation/map-tile/dev_guide/topics/resource-base-basetile.html states 
 "API Keys never expire but you can invalidate your API Keys at any time."

ahocevar (approved):
Thanks, @mike-000

PR #11900, opened 2021-01-10
Update pratique key to choisirgeoportail in the IGN WMTS example

mike-000:
The "pratique" key is no longer working in the IGN WMTS example.  There is a more recent free alternative "choisirgeoportail" for WMTS https://geoservices.ign.fr/blog/2018/09/06/acces_geoportail_sans_compte.html


ahocevar (approved):
Thanks, @mike-000

PR #11913, opened 2021-01-14
Handle null geometry in TopoJSON

### LINKED ISSUE #11910: null geometry type in TopoJSON
**Describe the bug**
See https://stackoverflow.com/questions/65680845/error-loading-topojson-file-in-openlayers-5

**Expected behavior**
`null` geometry type is supported in the TopoJSON specification but causes an error when read in OpenLayers.  Since the OpenLayers GeoJSON test checks it 'can read a feature with null geometry' there should be no undesired side effects if TopoJSON did the same.

mike-000:
Fixes #11910

Handle features will null geometry type similar to features will null geometry in GeoJSON.

### DIFF
--- a/src/ol/format/TopoJSON.js
+++ b/src/ol/format/TopoJSON.js
@@ -368,18 +368,18 @@ function readFeatureFromGeometry(
   name,
   opt_options
 ) {
-  let geometry;
+  let geometry = null;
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #11925, opened 2021-01-19
Mention unit of radius for Polygon.circular()

kannes:
Radius was using an unknown unit. This makes it clear that it will be interpreted as meters.


mike-000:
For a 1500 meter radius circle 

```
new Circle(
  [945751.80, 7154116.68],  // coordinates in EPSG:3857
  1500 / getPointResolution('EPSG:3857', 1, [945751.80, 7154116.68])
),

```
would be a very close approximation to a true circle.

ol/geom/Polygon.circular is most useful for large radius circles, especially in regions where the projection scale changes considerable with latitude, for example 500, 1000, 1500, 2000 km radius in Greenland

image

tschaub (approved):

tschaub:
Thanks, @kannes

PR #11926, opened 2021-01-19
Example about metric circles

kannes:
EPSG:3857 says it is using "meters" but they are not the real world meters people might expect.
This is a trap that way too many people and projects fall into.
This example provides an example for the case of circles at least, hopefully providing some guidance and best practise.

There will be another commit. Sorry about that but Github's web interface does not let me update without using external tools and my fork is outdated. Apparently this is a way to get around that for now. Sorry for the mess...

### DIFF
--- a/examples/metric-circle.html
+++ b/examples/metric-circle.html
@@ -1,0 +1,17 @@
+---
+layout: example.html
+title: Metric Circle
+shortdesc: Example of a circular polygon with a radius specified in meters
+docs: >
+  A circular geometry created via `ol/geom/Circle`'s `Circle` constructor will use
+  the unit of the given coordinate reference system for the radius. For the default
+  Pseudo Mercator system these are "fake" meters much different from actual real-world
+  meters pretty much everywhere on the planet. A circle with a radius of 1500 of
+  these "fake" meters is shown in red.
+  
+  To create a circle more true to actual geodesic measurements, `ol/geom/Polygon`
+  provides the `circular` method to create a circle approximation in WGS84 coordinates.
+  A circle with a radius of 1500 actual meters is shown in blue.
+tags: "circle, projection, wgs84, vector, geometry, geographic, buffer"
+---
+<div id="map" class="map"></div>
--- a/examples/metric-circle.js
+++ b/examples/metric-circle.js
@@ -1,0 +1,70 @@
+import 'ol/ol.css';
+import Feature from 'ol/Feature';
+import Map from 'ol/Map';
+import View from 'ol/View';
+import {Circle} from 'ol/geom';
+import {circular} from 'ol/geom/Polygon';
+import {OSM, Vector as VectorSource} from 'ol/source';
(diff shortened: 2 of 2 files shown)


kannes:
There it is, this should be all. Very open for critique and comments.
And sorry for not being consistent about wording, I guess "metric circle" is the most correct term.

kannes:
No idea about the errors and I will not set up any JS development stuff, sorry.

mike-000:
For a 1500 meter radius circle 

```
new Circle(
  [945751.80, 7154116.68],  // coordinates in EPSG:3857
  1500 / getPointResolution('EPSG:3857', 1, [945751.80, 7154116.68])
),

```
would be a very close approximation to a true circle.

ol/geom/Polygon.circular is most useful for large radius circles, especially in regions where the projection scale changes considerable with latitude, for example 500, 1000, 1500, 2000 km radius in Greenland

image

tschaub:
This isn’t so much about real vs. fake or metric vs. something else.  The difference here is about spherical vs. planar measures.  The `Polygon.circular` function gives you a polygon that represents points at a given distance from a center point on the surface of a sphere.  The `Circle` constructor gives you a geometry representing all points at a given distance from a center point on a plane.  Earth is neither a sphere or a plane, but both can be used to approximate portions of it.

kannes:
@mike-000: Nice, thanks! That looks like a good way for when you know the circles will be small in size. I'd suggest I add that method as well so that people can see the variety of options. Do you have a ballpark value on the largest radius that this would be reasonable to use?

@tschaub: So so many people see "meters" as unit for EPSG:3857 and think they can simply treat that CRS's coordinate values and measurements as actual meters. The example is a hopeful attempt to guide at least some of them towards understanding of that issue. I didn't make that clear, sorry!

kannes:
I tried to fix the failing tests but am confused about also appending the ".js" extensions. I do not see that for other examples like e.g. https://openlayers.org/en/latest/examples/custom-circle-render.html or https://openlayers.org/en/latest/examples/button-title.html.

Is that (and the "../src/") automatically replaced before the examples are published to the web?

MoonE:
The code is indeed changed for the published examples:
You can see the original source code here
https://github.com/openlayers/openlayers/blob/main/examples/custom-circle-render.js

kannes:
I give up, battling a linter on a per-commit basis where online linters do not highlight any errors and no suggestions for fixing are given is not how I want to spent my time.

I'd love to see this example online as I think it will help many people avoid mistakes.

mike-000:
> I give up, battling a linter on a per-commit basis where online linters do not highlight any errors and no suggestions for fixing are given is not how I want to spent my time.
> 
> I'd love to see this example online as I think it will help many people avoid mistakes.

I have created #11951 which builds on your code. Preview at https://deploy-preview-11951--ol-site.netlify.app/examples/circle-comparison.html

tschaub:
Closing now that #12239 is in.

PR #11937, opened 2021-01-22
Use actual meters in JTST example

### LINKED ISSUE #11927: JSTS example talks about meters but means EPSG:3857's fake meters
**Describe the bug**
https://openlayers.org/en/latest/examples/jsts.html shows buffers around lines. It says that those buffers have a radius of 40 meters.
In reality the buffers have a radius of 40 units of EPSG:3857 which are called meters but are not corresponding to actual, real world meters. They are highly distorted pretty much everywhere on the planet.

https://gis.stackexchange.com/questions/329213/using-meters-to-buffer-with-jsts mentioned this but sadly the user filed no bug.

They also provided an image: https://i.stack.imgur.com/tIoBJ.png

To fix this the example would need to use a projection like UTM.

kannes:
Fixes #11927

# Visual comparison
Before:
image

After:
image

I will also add a note to the HTML.

### DIFF
--- a/examples/jsts.html
+++ b/examples/jsts.html
@@ -5,6 +5,10 @@ shortdesc: Example on how to use JSTS with OpenLayers.
 docs: >
   Example showing the integration of <a href="https://github.com/bjornharrtell/jsts">JSTS</a>
   with OpenLayers.
+
+  As JSTS operates in cartesian coordinate space, special attention has to be taken when
+  performing geometric calculations. In this example the local factor between meters and
+  map units is used to make sure the buffer radius is calculated correctly.
 tags: "vector, jsts, buffer"
 resources:
   - https://unpkg.com/jsts@2.3.0/dist/jsts.min.js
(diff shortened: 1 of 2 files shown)


kannes:
Can I do anything about the failing tests?

MoonE:
You can see the individual errors with suggestions on how to fix them at https://github.com/openlayers/openlayers/pull/11937/files.
The import paths should be prefixed with `../src/`.

kannes:
Thanks! I fixed that but will also have to add ".js" extensions apparently.

Before I do that, any suggestion how to fix the "map" issue? It is referenced in the fetch-block, which seems to work but of course is not clean. Should/could I construct the map without layers first?

MoonE:
This can be fixed by moving the fetch call to the very end of the file.

kannes:
Woohoo, made it. Sorry for the commit mess but I am not a JS person and had to play high latency feedback loop with the tests on the pipeline instead of running them locally...
Should I make a clean PR with clean commits?

PR #11938, opened 2021-01-22
Take view padding into account for calculateExtent()

### LINKED ISSUE #11936: ol.View#calculateExtent() does not take the ol.View#padding into account
**Describe the bug**
A new property `ol.View#padding` was added recently (https://github.com/openlayers/openlayers/pull/11755), which is very useful. 
But the property is not taken into account in the function `ol.View#calculateExtent()`.
If `view.calculateExtent()` is called without any attribute, the result extent is wrong.

There is a workaround, the function can be called with the right size: `view.calculateExtent(view.getViewportSizeMinusPadding_());`, then the result is correct.

See an example here: https://codesandbox.io/s/view-padding-forked-ktldk?file=/index.html

**To Reproduce**
1. Open https://codesandbox.io/s/view-padding-forked-ktldk?file=/index.html
2. Click on 'Display view.calculateExtent()'
3. The returned extent is wrong, it hasn't taken the view padding and shifted map center into account

**Expected behavior**
There should be a possibility to get the right complete extent and also just the visible extent.

ahocevar:
Fixes #11936.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -921,7 +921,7 @@ class View extends BaseObject {
    * @return {import("./extent.js").Extent} Extent.
    */
   calculateExtentInternal(opt_size) {
-    const size = opt_size || this.getViewportSize_();
+    const size = opt_size || this.getViewportSizeMinusPadding_();
     const center = /** @type {!import("./coordinate.js").Coordinate} */ (this.getCenterInternal());
     assert(center, 1); // The view center is not defined
     const resolution = /** @type {!number} */ (this.getResolution());
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -399,15 +399,10 @@ class View extends BaseObject {
     this.resolutions_ = options.resolutions;
 
     /**
-     * Padding (in css pixels).
-     * If the map viewport is partially covered with other content (overlays) along
-     * its edges, this setting allows to shift the center of the viewport away from that
-     * content. The order of the values in the array is top, right, bottom, left.
-     * The default is no padding, which is equivalent to `[0, 0, 0, 0]`.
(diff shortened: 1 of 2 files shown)


ahocevar:
I added another commit to this pull request, because I noticed that when just setting a new `padding` on the view, the center will jump e.g. with the first pointermove event, because it is not adjusted properly. aa2c301 takes care of that, so now a changed padding will not change any position on the viewport.

src/ol/View.js
```diff
@@ -448,6 +443,36 @@ class View extends BaseObject {
...
+   * its edges, this setting allows to shift the center of the viewport away from that
+   * content. The order of the values in the array is top, right, bottom, left.
+   * The default is no padding, which is equivalent to `[0, 0, 0, 0]`.
+   * @type {Array<number>|undefined}
+   * @api
+   */
+  get padding() {
+    return this.padding_;
+  }
+  set padding(padding) {
```
tschaub, line 458: Were people trying to change the padding by setting `view.padding`?

Not opposed to the setter, just noticing that it is different than the rest of the API (aside from MapBrowserEvent).

tschaub: I see that `padding` was marked as part of the API in 6.5.0 (20ccdc56b8e6f7d5fcf2642dccb79d2777950ac7).

tschaub (approved):

src/ol/View.js
ahocevar: Yes, people were trying to change the padding by setting `view.padding` because I had marked it as part of the API. Back then I had not thought about the consequences thoroughly enough, so I had to use a setter now.

ahocevar:
Thanks for the review, @tschaub.

PR #11941, opened 2021-01-23
Fix loading flag for vector loading strategies

### LINKED ISSUE #11934: map rendercomplete event fires multiple times
**Describe the bug**
Map "rendercomplete" event _should_ only fire once when all layers have rendered (if I understand the docs as written). 
Not sure if this is specific to Esri service layers...
I'm seeing issues when using multiple types of Esri layers (MapServer/ImageArcGISRest source, and FeatureServer/VectorSource source).
Ultimately, I want a single event to fire when the map is completely rendered for all visible layers.
Currently the rendercomplete event for the map fires multiple times, on both initial load (when VectorLayer is present, and for subsequent map renders (when any Esri service layer is present).

**To Reproduce**
1. add a VectorLayer to the map
2. log console on map rendercomplete event
3. refresh map to view multiple rendercomplete log statements for a single map render action
or:
1. add an Esri Image REST source layer to the map
2. log console on map rendercomplete event
3. refresh map to view a single map rendercomplete event on initial load
4. move/pan/zoom the map and multiple map rendercomplete events are logged to console.

**Expected behavior**
The map rendercomplete should also support any layer type (including VectorLayers) so that we can determine when the map has completed rendering all layers (without having to hook up separate layer "rendercomplete" event handlers).

Here are working examples demonstrating the issue:
with VectorLayer: https://stackblitz.com/edit/openlayers-rendercomplete-issue-vectorlayer-only
with Esri REST Image layers: https://stackblitz.com/edit/openlayers-rendercomplete-issue-esri-mapservicelayers-only
with both types of layers: https://stackblitz.com/edit/openlayers-rendercomplete-issue-demo

It looks like this should have been addressed here:
https://github.com/openlayers/openlayers/issues/6447

ahocevar:
Fixes #11934.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -235,6 +235,12 @@ class VectorSource extends Source {
     this.loadedExtentsRtree_ = new RBush();
 
     /**
+     * @type {number}
+     * @private
+     */
+    this.loadingExtentsCount_ = 0;
+
+    /**
      * @private
      * @type {!Object<string, import("../Feature.js").default<Geometry>>}
      */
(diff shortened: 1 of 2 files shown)


MoonE (approved):
Thanks, @ahocevar.

PR #11947, opened 2021-01-25
Demonstrates how to get the coordinates been clustered.

yukihiratype2:


mike-000:
Examples should use the documented API methods if possible.  `boundingExtent` would be preferable to the internal `createOrUpdateFromCoordinates` method.

yukihiratype2:
> Examples should use the documented API methods if possible. `boundingExtent` would be preferable to the internal `createOrUpdateFromCoordinates` method. Also a `if (features.length > 1)` check would be nice to avoid zooming in to a single point.

Done. Plz review.

ahocevar (approved):
This is a nice addition. Thanks, @yukihiratype2. And thanks @mike-000 for your review.

PR #11948, opened 2021-01-26
Improve layer typings

simonseyock:
As I was using the typings from hanreev, I noticed that many layers are not generic even if they support different source types. This PR fixes this.


simonseyock:
The benefit that this provides is that `getSource` and `setSource` have the correct source type if the generic parameter is set,

tschaub (approved):

PR #11950, opened 2021-01-26
Make polygons in the GeoJSON example comply to GeoJSON specification

mike-000:
Although OpenLayers is tolerant of open rings in the example, in a standards compliant GeoJSON https://tools.ietf.org/html/rfc7946#section-3.1.6 the polygon rings should be closed and use the right-hand rule.


tschaub (approved):

tschaub:
Thanks, @mike-000.

PR #11951, opened 2021-01-26
Circle geometry and circular polygon comparison example

mike-000:
Replaces #11926

Add a circle geometry and circular polygon comparison example.  This builds on #11926 by adding a circle geometry adjusted for point resolution and revising the description to avoid terms such as fake and metric.

### DIFF
--- a/examples/circle-comparison.html
+++ b/examples/circle-comparison.html
@@ -1,0 +1,21 @@
+---
+layout: example.html
+title: Circle Comparison
+shortdesc: Comparison of circle geomtry and a circular polygon
+docs: >
+  A circle geometry created via `ol/geom/Circle` will use the unit of the coordinate
+  reference system for the radius. Except for local projections projection units cannot
+  accurately represent true distances across the extent of a projection as the relative
+  scale or point resolution can vary considerably, especially for global projections such
+  as web mercator. A circle with a radius of 1000 km in projection units is shown in red,
+  and another with 1000 km radius adjusted for point resolution is shown in green.
+
+  Adjustment for point resolution is sufficient for a small radius, but with a larger
+  radius the point resolution can also vary within the radius, so a circle on the
+  Earth's surface will not be a circle geometry on a flat projection. To overcome this
+  `ol/geom/Polygon` provides the `circular` method to create a circle approximation
+  based on spherical geometry. A polygon with a geodesics great-circle radius of 1000 km
+  is shown in blue.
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks for taking additional time with this example, @mike-000.  I agree it is an improvement to remove any language about fake or real meters.

I'm not sure that including the green example really adds anything but potential confusion (who knows what "point resolution" means?).  If we are going to add any example of this sort, I think we could talk about two things: distances on a plane and distances on a sphere (suggesting that a sphere can be used to better represent the earth for large distances).

mike-000:
Maybe that needs a separate example.  From StackOverflow and GIS stackExchange it seems using (or attempting to use) circle geometry to represent radius on the sphere over relatively small distances is something that is done quite often.  It works off the self with the draw interaction and can be easily modified using the modify interaction.  However users then find a circle with a specified radius does not reach as far as expected, or a drawn circle returns an unexpectedly large value from `getRadius()`.  A geodesic circle can be drawn using a geometry function https://codesandbox.io/s/draw-and-modify-features-forked-wz0fy?file=/main.js but the result is a polygon and cannot be modified in the same way as a circle geometry.  Also `ol/sphere` does not have a `getRadius()` method, although radius could be calculated from the area, and for circle geometry that method produces a result very close to that from applying point resolution to the geometry radius, although of course both results represent an average.

kannes:
Thank you very much for this cleaned up example. :) I agree that my terminology was too blurry.

I would much prefer the circles to be much smaller, maybe 1 km so that it is clear how the issue manifests itself even for very local regions.

My main motivation for the proposal was to prevent people from falsely assuming that, if a projection says its units are "meters", they would correspond in any reasonable accuracy to "real" meters as used in every day, non-geodesist language. OpenLayers is a highly advanced library, its examples usually full of tricks and magic, but on the other hand it is used by many non-experts and it would be great if common pitfalls were easily avoided.

So a perfect solution would be a page that people end up on when googling. Maybe the title could be "Creating metric circles"?

For the point resolution maybe a sentence in parenthesis would provide all the information needed? To be honest, I read https://openlayers.org/en/latest/apidoc/module-ol_proj.html#.getPointResolution and after reading I right now do not actually quickly grasp what its return value actually is. In this case I assume something about the length of a 1-unit-length geometry in EPSG:3857 and the length of a 1-actual-meter-length unit on the EPSG:4326 ellipsoid (function description says sphere?)?

tschaub:
Closing now that #12239 is in.

PR #11952, opened 2021-01-26
Draw 800km circles around both poles in Tissot example

mike-000:
The line of the 800km circle around the south pole can be seen in the current example when panning, but there was no corresponding circle around the north pole.


MoonE (approved):

MoonE:
Thanks, @mike-000

PR #11954, opened 2021-01-27
Simplify icon example

### LINKED ISSUE #11953: Popup opens on the wrong marker
**Describe the bug**
Popup opens on the wrong marker.

recording

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/icon.html
2. Move worldmap till you see the marker again
3. Click on the marker
4. Notice that the popup opens on the main marker

**Expected behavior**
Popup should open on the clicked marker

ahocevar:
Fixes #11953.

### DIFF
--- a/examples/icon.js
+++ b/examples/icon.js
@@ -56,7 +56,6 @@ const popup = new Overlay({
   element: element,
   positioning: 'bottom-center',
   stopEvent: false,
-  offset: [0, -50],
 });
 map.addOverlay(popup);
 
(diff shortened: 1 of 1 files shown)


mike-000:
The problem with simply using `evt.coordinate` is that the popup does not stay on the icon when you zoom in

image

ahocevar:
> The problem with simply using evt.coordinate is that the popup does not stay on the icon when you zoom in

Same with the existing example. This is because the Bootstrap Popover lives outside the div element that's automatically anchored to the coordinate by `ol/Overlay`.

mike-000:
Yes, there was a regression between version 6.3 and 6.4 which is fixed by going back to the previous version of bootstrap.

ahocevar:
> Yes, there was a regression between version 6.3 and 6.4 which is fixed by going back to the previous version of bootstrap.

It could also be fixed by creating the Bootstrap popover with a `container` property, like in the `overlay.html` example. But because the purpose of this example is to showcase working with icons, not popups, I added a commit that simply closes the popup when the map is moved.

MoonE (approved):
Thanks, @ahocevar.

PR #11965, opened 2021-01-30
Handle view rotation in d3 Integration example

### LINKED ISSUE #11964: d3 integration example no rotation for svg layer
Hi
In this example https://openlayers.org/en/latest/examples/d3.html 

when you rotate the map the svg layer doesnt rotate ...

I have a playground here https://observablehq.com/@jipexu/openlayers-d3js-map-svg-with-json where i try to make it work ..

any advice are welcome
Thank

mike-000:
Fixes #11964

Use d3 projection `.angle()` method to rotate rendered output with the view rotation.


MoonE (approved):
Thanks, @mike-000  and  @jipexu for your joint effort.

PR #11967, opened 2021-01-31
Simplify popup positioning in Icon Scale example

mike-000:
The Icon Scale example uses popups to indicate which feature has been detected when clicked.  These were based on the Icon Symbolizer example and suffer the same issues with wrapping and Bootstrap update which were fixed in #11954


ahocevar (approved):
Thanks, @mike-000

PR #11974, opened 2021-02-02
Fix polyfill inconsistencies

### LINKED ISSUE #11973: Typo in script tag at example
at
https://openlayers.org/en/latest/examples/vector-tiles-4326.html
it says
`    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=String.prototype.startsWith,Object.assign""></script>
`
should change the doublequote at the end to appear once, not twice (now it has a double doublequote ;-)

mike-000:
Fixes #11973

Remove obsolete polyfills previously needed by `ol-mapbox-style` from Vector tiles in EPSG:4326 example.
Include `Number.isInteger` in displayed html for all examples and in tutorial.

### DIFF
--- a/doc/tutorials/background.md
+++ b/doc/tutorials/background.md
@@ -19,7 +19,7 @@ OpenLayers is available as [`ol` npm package](https://npmjs.com/package/ol), whi
 
 By default, OpenLayers uses a performance optimized Canvas renderer.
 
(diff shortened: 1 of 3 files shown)


MoonE (approved):
Thanks, @mike-000

PR #11975, opened 2021-02-02
jsdoc linting

simonseyock:
I used `eslint-plugin-jsdoc` to harmonize the jsdoc in the code base. I decided to replace `@returns` by `@return` because it is used way more.

### DIFF
--- a/test/test-extensions.js
+++ b/test/test-extensions.js
@@ -332,10 +332,15 @@
   }
 
   /**
+   * @typedef {Object} xmleqlOptions
+   * @property {boolean} [includeWhiteSpace]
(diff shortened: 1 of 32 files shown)


tschaub (approved):

tschaub:
This is great.  Thanks, @simonseyock.

test/test-extensions.js
```diff
@@ -331,11 +331,16 @@
     }
   }
 
+  /**
+   * @typedef {Object} xmleqlOptions
```
tschaub, line 335: Minor, but in #11977 I suggested we change this to `XMLEqlOptions` for consistency with other custom types.

PR #11977, opened 2021-02-02
More consistent JSDoc syntax

tschaub:
Following up on #11975, this adds a bit more consistency with our JSDoc syntax.

 * Arrays use `Array<ItemType>` instead of `Array.<ItemType>`
 * Objects use `Object<KeyType, ValueType>` instead of `Object.<KeyType, ValueType>`
 * All custom types are UpperCamelCase


MoonE (approved):
Thanks, @tschaub.

simonseyock:
Looks great!

PR #11979, opened 2021-02-03
Jsdoc types syntax

simonseyock:
With these commits the `check-types` and `check-syntax` rules of `eslint-plugin-jsdoc` pass.

The first commit contains some type changes and second commit only replaces the `{type=} name` syntax with `{type} [name]`.


tschaub (approved):

tschaub:
Thanks for the continued work on this, @simonseyock.

PR #11987, opened 2021-02-06
Respect opacity for vector instructions

### LINKED ISSUE #11982: Mapbox Vector Layer opacity is not applied to labels
Change opacity of a Mapbox Vector Layer with declutter false doesn't change labels opacity

1. Go to https://codesandbox.io/s/mapbox-vector-layer-forked-pthnb?file=/main.js
2. Opacity doesn't change for labels

**Expected behavior**
Opacity for labels should also be applied.

ahocevar:
Fixes #11982.

Layer opacity was not applied when vector instructions were rendered for VectorTile layers. The result was that with `renderMode: 'vector'`, layer opacity had no effect at all, and with `renderMode: 'hybrid'`, images and text did not get the layer opacity applied.


src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -604,6 +604,7 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
    * @param {import("../../PluggableMap.js").FrameState} frameState Frame state.
    */
   renderDeclutter(frameState) {
+    this.getLayer().getOpacity();
```
MoonE, line 607: This does nothing, it should set and reset globalAlpha in the same way as in the renderFrame method.

MoonE (changes requested):
I believe there should be a test case with labels and `declutter: true`.

Am I wrong in assuming the two expected.png files should be the same? Because the `renderMode: 'vector'` is kind of darker. The opacity is 0.1 in both.

Below are both color-adjusted images:
image

src/ol/renderer/canvas/VectorTileLayer.js
ahocevar: Thanks, I had forgotten to finish that. I just added another commit with a test for decluttering.

Regarding your question why the vector mode example is darker: this is because overlapping geometries are each drawn with the layer opacity, so the overall opacity will be higher. For vector layers, this was solved with quite an expensive workaround (render to a separate canvas, and dim that down with the layer opacity - see #6981), but I think this is a corner case and does not justify the extra rendering work.

MoonE (approved):
Looks good now, thanks @ahocevar .

PR #11988, opened 2021-02-06
Warn when map container's width or height are zero

### LINKED ISSUE #11984: Map no longer renders unless map div height specified explicitly
Bug spotted when upgrading from OL 5 to 6.

I did notice someone had opened a previous issue #10327 but it has gone stale. I think we do either need a fix or an explanation as it is not obvious what the issue is.

For an example using version 6 see 
https://test.floodassist.co.uk/flood-warnings/flood-area-info/cambridgeshire/055fwfpmne08/areas-near-the-river-nene-from-elton-to-wansford

For the working version (OL 5)  see
https://floodassist.co.uk/flood-warnings/flood-area-info/cambridgeshire/055fwfpmne08/areas-near-the-river-nene-from-elton-to-wansford

If I use chrome tools to change the page-map-size css class from min-height:600px to height:600px the map suddenly renders correctly.

ahocevar:
Fixes #11984.

I think it makes sense to let the user know why no map is visible in the non obvious case where the container's width or height is `0`.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1500,18 +1500,25 @@ class PluggableMap extends BaseObject {
       this.setSize(undefined);
     } else {
       const computedStyle = getComputedStyle(targetElement);
-      this.setSize([
+      const width =
         targetElement.offsetWidth -
-          parseFloat(computedStyle['borderLeftWidth']) -
-          parseFloat(computedStyle['paddingLeft']) -
-          parseFloat(computedStyle['paddingRight']) -
-          parseFloat(computedStyle['borderRightWidth']),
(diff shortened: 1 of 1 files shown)


MoonE (approved):
Thanks,  @ahocevar, not seeing a map because the container had zero height is probably a common problem.

AndreaSGI:
> Thanks, @ahocevar, not seeing a map because the container had zero height is probably a common problem.

How to handle a case when the Map container is hidden (in my case it is not visible because it is in another tab), and the initialization gives this error?

MoonE:
This should befixed in the next version by #12632, until then you can simply ignore it.

PR #11991, opened 2021-02-06
Fix some typings in format and format/WKT

simonseyock:
I turned on `strictNullChecks` for the tsc to find issues in the openlayers code. I looked at the wkt format in particular and fixed some types here.

The `EMPTY` geometry handling was causing some problems because some functions would return `null`. I checked the OGC specifications and figured out that `EMPTY` can only occur immediately after a geometry type. This way I was able to move these checks out of the functions and into `parseGeometry_`.

### DIFF
--- a/src/ol/format/Feature.js
+++ b/src/ol/format/Feature.js
@@ -67,15 +67,15 @@ class FeatureFormat {
   constructor() {
     /**
      * @protected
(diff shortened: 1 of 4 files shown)


tschaub (approved):
Thanks, @simonseyock

PR #11992, opened 2021-02-06
Improve measure example ux

MoonE:
- Pass click events on measure tooltips through to the map
- Put the tooltip for  the currently drawn geometry on top
- Disable text selection in the tooltips so clicking to draw does not select


ahocevar (approved):
Thanks, @MoonE.

PR #11995, opened 2021-02-07
Fix hitdetect inaccuracy for VectorLayer's getFeatures

MoonE:
If the mouse is at the last pixel on the right side of the map the feature at the opposite side was detected as a hit because the calculated x after rounding was larger than hit detect canvas width.

Basically the hit detection currently is off by one pixel over the whole canvas and wraps around at the end.


ahocevar (commented):
Good find, @MoonE. Would you be able to add a test for this? Thanks in advance.

MoonE:
The pixel coordinates can have decimals, so the correct thing to do is round them to the nearest pixel, then scale it down to the hit canvas using floor.

MoonE:
Now the pixel coordinates are also clamped to the canvas dimensions because the pixel coordinate may actually slightly exceed the canvas  width when the pixel ratio is not an integer or for some other reason.

ahocevar (approved):
Thanks, @MoonE. This looks great now.

PR #12012, opened 2021-02-09
Fix documentation for account parameter in ol/source/CartoDB

### LINKED ISSUE #12011: CartoDB Source account documentation wrong
`account` parameter in documentation is wrongly described as: 

> If using named maps, this will be the name of the template to load.

This is wrong, it should be something like:

> username as used to access public Carto dashboard at https://{username}.carto.com/.

geraldo:
Fixing description for parameter `account` in ol/source/CartoDB. Fixes #12011


tschaub (approved):

tschaub:
Thanks, @geraldo

Issue #12014, opened 2021-02-09
MapboxVectorLayer does not correctly set additional options


roemhildtg:
**Describe the bug**

If I pass an additional property to a MapboxVectorLayer, layer.get('customProp') does not return it.

For example: 

```

    const layer = new MapboxVector({
      type: 'ol/layer/MapboxVector',
      styleUrl: 'mapbox://styles/mapbox/light-v10',
      accessToken: 'xxxx',
      id: 'mapbox-gray',
    });
    
    layer.get('id') // -> undefined
```

**Expected behavior**

Most (if not all) other openlayers layer objects allow this. The use case is that I can populate the layer with some additional metadata that lets me find the layer in the future. I think that to ensure consistency in the API the mapbox layer should also allow this. 

I am able to contribute a pull request if this is an acceptable solution.

tschaub:
I think we need to decide what direction we want to go with this.  I know that there is inconsistency in that some things allow arbitrary properties to be set.  It would be nice to have more consistency.

It seems like people like the flexibility of adding arbitrary properties to things like layers.

One of the drawbacks of this is that it is a liability if the library ever chooses to add properties (that may have different meaning, types, etc than how people are using them).

Another drawback of allowing arbitrary properties is that people then assume that the library will do things like ensuring they are applied to a clone (see #11071).

Disallowing arbitrary properties makes it possible to improve type checking within the library and for application developers (see #11960).

My personal opinion is that it was a mistake to make everything extend the common `BaseObject`, to mix user set properties with things that the library depends on, and to trigger events on all property changes.

I'm curious what others have to say about it.

ahocevar:
I totally agree with @tschaub here, and I'd be in favor of making a breaking change, getting rid of `BaseObject`.

mike-000:
As arbitrary user properties are accessed using `getProperties()` and `setProperties()` then in constructors instead of

`name: 'My Layer',`

it would be more logical to use

`properties: {name: 'My Layer'},`

In the case of features it would make OpenLayers features more consistent with GeoJSON features and avoid ambiguities in the use of geometry, id and style.

roemhildtg:
I'll add my two cents here and thoughts.

> It would be nice to have more consistency.

Totally agree. I'm not opposed to removing the custom properties being assigned to the class level properties if it would improve consistency and help future proof the API. I would, however, like a way to pass custom properties to be stored _somewhere_ in the class. See below.

[...]

PR #12015, opened 2021-02-09
ensure additional properties are set correctly

roemhildtg:
https://github.com/openlayers/openlayers/issues/12014


tschaub:
Thanks for the proposed change, @roemhildtg.  Let's discuss on #12014 first.

PR #12016, opened 2021-02-09
Fix VectorLayer hitdetect problems

MoonE:
This fixes some issues with the image hitdetection with the `getFeatures` method.
- There were two canvas elements created per icon, one is enough
- icon opacity should be always set to 1 otherwise the color won't match the feature
- If no zIndex is set `NaN` was used as the array key which messes with the whole order in which features are drawn because it evaluates to false for any compare operation
- added a constant for the hit detect canvas scale, as 0.5, like before
- ~if no image size was set the whole style was skipped instead of just ignoring the image~ Not quite sure if that's a problem...


ahocevar (approved):
Thanks, @MoonE!

PR #12020, opened 2021-02-11
Fix WebGL buffer deletion

tschaub:
The WebGL helper's `deleteBuffer()` method would previously throw since it was trying to call `gl.deleteBuffer()` with an object that is not a [`WebGLBuffer`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLBuffer).

This was part of #12008.  I'm pulling out small changes to make review easier.

### DIFF
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -410,8 +410,8 @@ class WebGLHelper extends Disposable {
     const gl = this.getGL();
     const bufferKey = getUid(buf);
     const bufferCacheEntry = this.bufferCache_[bufferKey];
-    if (!gl.isContextLost()) {
-      gl.deleteBuffer(bufferCacheEntry.buffer);
+    if (bufferCacheEntry && !gl.isContextLost()) {
+      gl.deleteBuffer(bufferCacheEntry.webGlBuffer);
     }
     delete this.bufferCache_[bufferKey];
   }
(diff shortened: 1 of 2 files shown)


src/ol/webgl/Helper.js
```diff
@@ -410,8 +410,8 @@ class WebGLHelper extends Disposable {
     const gl = this.getGL();
     const bufferKey = getUid(buf);
     const bufferCacheEntry = this.bufferCache_[bufferKey];
-    if (!gl.isContextLost()) {
-      gl.deleteBuffer(bufferCacheEntry.buffer);
+    if (bufferCacheEntry && !gl.isContextLost()) {
+      gl.deleteBuffer(bufferCacheEntry.webGlBuffer);
```
tschaub, line 414: The `buffer` property on the cache entry is an instance of a `WebGLArrayBuffer` from the `ol/webgl/Buffer.js` module.  The `webGlBuffer` property is an instance of a [`WebGLBuffer`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLBuffer).  The [`gl.deleteBuffer()`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/deleteBuffer) method expects the former.

test/spec/ol/webgl/helper.test.js
```diff
@@ -240,6 +242,18 @@ describe('ol/webgl/WebGLHelper', function () {
...
 
+    describe('deleteBuffer()', function () {
+      const helper = new WebGLHelper();
+      const buffer = new WebGLArrayBuffer(ARRAY_BUFFER, STATIC_DRAW);
+      buffer.fromArray([0, 1, 2, 3]);
+      helper.flushBufferData(buffer);
+      const bufferKey = getUid(buffer);
+      expect(helper.bufferCache_).to.have.property(bufferKey);
+
+      helper.deleteBuffer(buffer);
```
tschaub, line 253: Without the above change, this call throws.

ahocevar (approved):
Thanks @tschaub for taking the time to pull out small pieces of #12008 for easier reviewing, and for explaining the fix in detail. This looks great.

PR #12021, opened 2021-02-11
Allow source projection to be set after construction

tschaub:
It is useful to be able to create a source that is not immediately "ready" after construction.  The TileJSON source is an example that configures itself asynchronously before setting its state to ready.  In the case of the TileJSON source, the tile grid, tile URL function, attributions, and other properties are not known at construction.

This change adds the source projection to the list of things that can be set after construction.  I added a `setProjection` method to provide the convenience of being able to pass a projection id instead of a `Projection` instance.  But we could also rename the `projection_` property to `projection` and make that protected instead of private if we want (this would be consistent with `tileGrid` and other post-construction-settable properties.

### DIFF
--- a/src/ol/source/Source.js
+++ b/src/ol/source/Source.js
@@ -51,10 +51,10 @@ class Source extends BaseObject {
     super();
 
     /**
-     * @private
+     * @protected
      * @type {import("../proj/Projection.js").default}
      */
-    this.projection_ = getProjection(options.projection);
+    this.projection = getProjection(options.projection);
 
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
I agree it would be somewhat nicer to rename `this.projection_` to `this.projection` and make it `@protected`. Feel free to change before merging, or merge as-is.

tschaub:
I renamed `projection_` to `projection` and made it protected (which I agree is nicer).

PR #12022, opened 2021-02-11
Throw when WebGL style parsing encounters an unexpected type

tschaub:
Currently, the `expressionToGlsl` function returns `undefined` if the passed value doesn't match the passed type hint.  This means that the string `"undefined"` ends up being serialized in a shader - leading to failure later at shader compilation or (worse) in mysterious rendering issues.

This change makes it so we throw as soon as possible when parsing a style if we encounter something unexpected.

### DIFF
--- a/src/ol/style/expressions.js
+++ b/src/ol/style/expressions.js
@@ -260,24 +260,35 @@ export function expressionToGlsl(context, value, typeHint) {
     }
     return operator.toGlsl(context, value.slice(1), typeHint);
   }
+
   const valueType = getValueType(value);
   if ((valueType & ValueTypes.NUMBER) > 0) {
     return numberToGlsl(/** @type {number} */ (value));
-  } else if ((valueType & ValueTypes.BOOLEAN) > 0) {
+  }
+
+  if ((valueType & ValueTypes.BOOLEAN) > 0) {
     return value.toString();
-  } else if (
+  }
+
+  if (
     (valueType & ValueTypes.STRING) > 0 &&
     (typeHint === undefined || typeHint == ValueTypes.STRING)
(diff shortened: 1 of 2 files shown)


src/ol/style/expressions.js
```diff
@@ -260,24 +260,35 @@ export function expressionToGlsl(context, value, typeHint) {
     }
     return operator.toGlsl(context, value.slice(1), typeHint);
   }
+
   const valueType = getValueType(value);
   if ((valueType & ValueTypes.NUMBER) > 0) {
     return numberToGlsl(/** @type {number} */ (value));
-  } else if ((valueType & ValueTypes.BOOLEAN) > 0) {
```
tschaub, line 266: I've also removed the `else` after `return` to make this function a bit more readable.  I'll follow at some point by proposing that we enforce this with eslint.

src/ol/style/expressions.js
```diff
@@ -260,24 +260,35 @@ export function expressionToGlsl(context, value, typeHint) {
...
   ) {
     return colorToGlsl(/** @type {Array<number> | string} */ (value));
-  } else if ((valueType & ValueTypes.NUMBER_ARRAY) > 0) {
+  }
+
+  if ((valueType & ValueTypes.NUMBER_ARRAY) > 0) {
     return arrayToGlsl(/** @type {Array<number>} */ (value));
   }
+
+  throw new Error(`Unexpected expression ${value} (expected type ${typeHint})`);
```
tschaub, line 291: This is the only significant change - we now throw instead of returning `undefined`.

ahocevar (approved):
Thanks, @tschaub!

PR #12023, opened 2021-02-11
Avoid collisions between user variables and internal names in WebGL shaders

tschaub:
Currently, when users supply `variables` in a WebGL layer's style, we serialize the variable names as uniforms by prefixing them with `u_`.  This means that user supplied variables will collide with the uniforms that we use internally - so variable names like `time`, `resolution`, `zoom`, etc. would cause problems.

This change proposes that we prefix user supplied variables with `u_var_` when creating uniform names.  This will avoid conflicts with uniforms that we use internally (as long as we remember not to name them `u_var_something`.

PS - We should do the same with feature attributes that get encoded as attributes in a vertex shader, but I'm not taking that on here.

### DIFF
--- a/src/ol/style/expressions.js
+++ b/src/ol/style/expressions.js
@@ -378,6 +378,15 @@ Operators['get'] = {
   },
 };
 
+/**
+ * Get the uniform name given a variable name.
+ * @param {string} variableName The variable name.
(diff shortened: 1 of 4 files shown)


jahow (approved):
Looking good, thanks! Nothing to add

PR #12024, opened 2021-02-11
Add a tile grid method for getting a tile range given a tile coord and zoom level

tschaub:
Tile grids have a `forEachTileCoordParentTileRange` method.  This calls a callback with a tile range for each zoom level that is lower than the tile coord's zoom level.  This function is awkward to use since the callback inevitably needs variables from the calling scope (including `this`).  Where we use it, we have to have another function that creates this callback function (see [`createLoadedTileFinder`](https://github.com/openlayers/openlayers/blob/98ac74e0bf6ebea00a95ddca5f479b30ac09eddc/src/ol/renderer/Layer.js#L80-L102)).

Since the caller already knows the tile coord's zoom level and the tile grid's min zoom level, it is more convenient to write a regular old `for` loop iterating through parent zoom levels.  What is missing is a function to get a tile range for a tile coord and an arbitrary zoom level.  The change here adds that function.

This could be used to simplify the loaded tile finding logic in the canvas tile renderers, but I haven't taken that on here.  I added the function as part of #12008, and I've pulled it out here to make reviewing easier.

### DIFF
--- a/src/ol/tilegrid/TileGrid.js
+++ b/src/ol/tilegrid/TileGrid.js
@@ -360,6 +360,48 @@ class TileGrid {
   }
 
   /**
+   * @param {import("../tilecoord.js").TileCoord} tileCoord Tile coordinate.
+   * @param {number} z Integer zoom level.
+   * @param {import("../TileRange.js").default=} opt_tileRange Temporary import("../TileRange.js").default object.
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @tschaub!

PR #12028, opened 2021-02-13
Overview map no initial target

### LINKED ISSUE #12010: Initially expanded overview is empty until first collaps
**Describe the bug**
This is an almost duplicate of a stale bug (#9274)
The stale bug refers to OpenLayers 4.x
I am using 6.5.0

https://jsfiddle.net/mmomtchev/ouebkx0z/6/

```js
const map = new ol.Map({
  controls: ol.control.defaults().extend([
    new ol.control.OverviewMap({collapsed: false, layers: [new ol.layer.Tile({source: new ol.source.OSM()})]})
  ]),
  layers: [
    new ol.layer.Tile({
      source: new ol.source.OSM()
    })
  ],
  view: new ol.View({
    center: ol.proj.fromLonLat([37.41, 8.82]),
    zoom: 4
  })
});

map.setTarget('map');
```

**To Reproduce**
Create a map with an overview control without a target and a collapsed=false option
Call setTarget after the map is created

**Expected behavior**
The minimap is immediately displayed

MoonE:
Fixes #12010

Also prevents `[NaN, NaN]` as map size when the target element is not in the document.
The styles from `getComputedStyle` are empty strings when the map  is not in the DOM. `parseFloat` converts them to `NaN`.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1496,9 +1496,8 @@ class PluggableMap extends BaseObject {
   updateSize() {
     const targetElement = this.getTargetElement();
 
-    if (!targetElement) {
-      this.setSize(undefined);
-    } else {
+    let size = undefined;
+    if (targetElement) {
       const computedStyle = getComputedStyle(targetElement);
       const width =
         targetElement.offsetWidth -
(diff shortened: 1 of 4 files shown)


tschaub (approved):

MoonE:
I changed the for the invalid size a bit because it did not work when the map is added to a shadow root.
I changed the size check to using the `hasArea` function because the calculated size may actually become negative.

ahocevar (approved):
The additional changes look good to me. Thanks, @MoonE.

PR #12057, opened 2021-02-22
TileQueue tests with events instead of setTimeout

ahocevar:
This pull request fixes the TileQueue tests, which sometimes do not pass. The reason was that the timeout of 20ms was sometimes too short to get all tiles loaded. So now, we rely on change events of each tile to detect when all tiles are loaded, and then finish the tests immediately in the next cycle (`setTimeout(finish, 0)`);


tschaub (approved):
Thanks for looking into this.

PR #12058, opened 2021-02-22
Fix map loading for basemap.at and add attributions

M393:
https://openlayers.org/en/latest/examples/wmts-hidpi.html does not load any tiles.
https://www.basemap.at/ redirects to https://basemap.at/ but there are no CORS headers set on the former.
Add attribution for the map tiles, see: https://basemap.at/#lizenz


ahocevar (approved):
Thanks, @M393

PR #12062, opened 2021-02-23
Fix kml-timezone example tooltip

M393:
The tooltip did not work at all for https://openlayers.org/en/latest/examples/kml-timezones.html


marcjansen (approved):
Thanks, @M393

PR #12066, opened 2021-02-24
Always return Promise from getFeatures

### LINKED ISSUE #12065: getFeatures returns error when the layer is added but not yet rendered
**Describe the bug**
Calling getFeatures on a `ol/layer/VectorTile` that is not yet rendered but visible throws an error because this.renderer_ is not yet set.
https://github.com/openlayers/openlayers/blob/b6e7d617b5860aba7dd4dc8835085c415c64d731/src/ol/layer/Layer.js#L212-L214
This likely also happens with `ol/layer/Vector`.

**To Reproduce**
1. Go to https://codesandbox.io/s/getfeatures-error-after-load-qs14b?file=/main.js
2. After page is loaded move the mouse over the map until the layer is added
3. When the layer is added 3 seconds later you get an error

**Expected behavior**
There should be no error, it should always return a Promise.

M393:
Fixes #12065
Also changed `VectorImageLayer`'s getFeatures function to follow the same pattern, no functional change there.

### DIFF
--- a/src/ol/layer/Layer.js
+++ b/src/ol/layer/Layer.js
@@ -210,6 +210,9 @@ class Layer extends BaseLayer {
    * an array of features.
    */
   getFeatures(pixel) {
+    if (!this.renderer_) {
+      return new Promise((resolve) => resolve([]));
+    }
     return this.renderer_.getFeatures(pixel);
   }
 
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @M393

src/ol/layer/Layer.js
```diff
@@ -210,6 +210,9 @@ class Layer extends BaseLayer {
    * an array of features.
    */
   getFeatures(pixel) {
+    if (!this.renderer_) {
+      return new Promise((resolve) => resolve([]));
```
tschaub, line 214: Minor, but this could be `return Promise.resolve([])`.

M393: Thanks @tschaub, I did not know this. Should I create a PR?

tschaub: Not critical.

PR #12079, opened 2021-03-02
Change IGN WMTS example to use PLANIGNV2 layer 

mike-000:
Support of the `pratique` and `choisirgeoportail` keys has changed again https://geoservices.ign.fr/blog/2021/01/29/Maj_Cles_Geoservices.html and neither now support the `GEOGRAPHICALGRIDSYSTEMS.MAPS` layer.  While `GEOGRAPHICALGRIDSYSTEMS.MAPS` remains supported by free personal keys obtained (temporarily) via https://www.sphinxonline.com/surveyserver/s/etudesmk/Geoservices_2021/questionnaire.htm#1 these require security settings such as referer which would make them unsuitable for users editing the example in codesandbox, so it seems appropriate to use the new `GEOGRAPHICALGRIDSYSTEMS.PLANIGNV2` layer instead.

### DIFF
--- a/examples/wmts-ign.js
+++ b/examples/wmts-ign.js
@@ -19,7 +19,7 @@ const matrixIds = [];
 const proj3857 = getProjection('EPSG:3857');
 const maxResolution = getWidth(proj3857.getExtent()) / 256;
 
-for (let i = 0; i < 18; i++) {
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12080, opened 2021-03-02
Add a minimum distance config to the cluster source

### LINKED ISSUE #12078: Cluster doesn't respect minimum distance
**Describe the bug**
Features are closer together than specified minimum distance.
Version 6.5.0

**To Reproduce**
Open Cluster example from documentation, potentially zoom in/out or slightly adjust distance to see overlap:
https://openlayers.org/en/v6.5.0/examples/cluster.html

Features have a radius of 10 pixels. Distance is at 40 pixels. Still some features overlap:
image

**Expected behavior**
My expectation would be that the centers of the clusters are at minimum `distance` pixels apart, as stated in the documentation:
https://openlayers.org/en/v6.5.0/apidoc/module-ol_source_Cluster-Cluster.html

MoonE:
Fixes #12078
This adds a minimum distance config to `ol/source/Cluster`
As it is in the current version clusters may still overlap which kind of defeats the purpose of the cluster source.
This new config should guarantee the configured distance between each cluster. It is meant to be set to the used icon width or height. This way icons won't overlap and the position shift away from the center of a cluster's features is at a minimum.


src/ol/source/Cluster.js
```diff
@@ -9,13 +9,19 @@ import Point from '../geom/Point.js';
...
+  getCenter,
+} from '../extent.js';
 import {getUid} from '../util.js';
 
 /**
  * @typedef {Object} Options
  * @property {import("./Source.js").AttributionLike} [attributions] Attributions.
- * @property {number} [distance=20] Minimum distance in pixels between clusters.
+ * @property {number} [distance=20] Distance within which features will be clustered
+ * together.
```
ahocevar, line 24: Please add back the units of the distance (pixels).

src/ol/source/Cluster.js
```diff
@@ -29,6 +35,11 @@ import {getUid} from '../util.js';
  * ```
  * See {@link module:ol/geom/Polygon~Polygon#getInteriorPoint} for a way to get a cluster
  * calculation point for polygons.
+ * @property {number} [minDistance=0] Minimum distance between clusters. Will be capped
+ * at the configured distance.
+ * By default no minimum distance is guaranteed. This config can be used to avoid
+ * overlapping icons. As a tradoff, the cluster feature's position will no longer be
+ * the center of all its features.
```
ahocevar, line 42: Please indicate the unit of the minDistance. Pixels?

examples/cluster.js
```diff
@@ -13,7 +13,11 @@ import {Cluster, OSM, Vector as VectorSource} from '../src/ol/source.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
 import {boundingExtent} from '../src/ol/extent.js';
 
-const distance = document.getElementById('distance');
+const distanceInput = document.getElementById('distance');
+const distanceNode = document.getElementById('distance-info');
+const minDistanceInput = document.getElementById('min-distance');
+const minDistanceNode = document.getElementById('min-distance-info');
+const numClustersNode = document.getElementById('num-clusters');
```
ahocevar, line 20: I am always a bit hesitant to approve pull requests that add a lot of code to examples to make config options configurable through forms, so users see how the options affect the result.

My suggestion: rename this example to something like `cluster-configuration`, and create a new `cluster` example that just shows a cluster source with the defaults.

ahocevar (changes requested):
Thanks for your work on this, @MoonE. The code looks good. Please consider my suggestions regarding docs and example.

examples/cluster.js
MoonE: Would you say it is better now? The output nodes aren't really necessary.

ahocevar: Definitely better, thanks!

ahocevar (approved):

PR #12083, opened 2021-03-03
Simplify event handling for custom elements

### LINKED ISSUE #12029: Error in handleMapBrowserEvent for keydown event with map in shadowroot
**Describe the bug**
An error is thrown If the map is inside a shadow root element and a `keydown` event occurs inside with the map having focus.

The `clientX` and `clientY` properties are not available for key events.
https://github.com/openlayers/openlayers/blob/27b45393607bb25e03a82e4e35a2a065b450c3b0/src/ol/PluggableMap.js#L990-L1005
```
Uncaught TypeError: ShadowRoot.elementFromPoint: Argument 1 is not a finite floating-point value.
    handleMapBrowserEvent PluggableMap.js:1001
    handleBrowserEvent PluggableMap.js:976
```

ahocevar:
Fixes #12029.

This pull request removes the special handling of events originating from a map in a web component. Now it is no longer necessary to use `elementFromPoint()`. The root of the problem that was meant to be fixed in #11024 came from the emulated click events in `MapBrowserEventHandler`, which stores a `pointerdown` event instead of just its relevant properties.

### DIFF
--- a/src/ol/MapBrowserEvent.js
+++ b/src/ol/MapBrowserEvent.js
@@ -7,7 +7,7 @@ import MapEvent from './MapEvent.js';
  * @classdesc
  * Events emitted as map browser events are instances of this type.
  * See {@link module:ol/PluggableMap~PluggableMap} for which events trigger a map browser event.
- * @template {UIEvent} EVENT
+ * @template {UIEvent|import("./MapBrowserEventHandler").PointerEventData} EVENT
  */
 class MapBrowserEvent extends MapEvent {
   /**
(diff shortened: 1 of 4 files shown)


src/ol/index.modern.js
```diff
@@ -0,0 +1,46 @@
+/**
```
tschaub, line 1: Did you mean to include this file?

ahocevar: That was an accident. Thanks for your sharp eye! I force-pushed a version without that file.

tschaub (approved):

PR #12090, opened 2021-03-06
Use WGS84BoundingBox from WMTSCapabilities.xml, set wrapX

### LINKED ISSUE #11350: WMTS optionsFromCapabilities ignores WGS84BoundingBox
**Describe the bug**
WGS84BoundingBox from WMTSCapabilities.xml is ignored.
There exists a test to check it is read correctly but it is not used anywhere in the source.
Related issues: #10979.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/wmts-hidpi.html
2. Look in network manager when the view is outside the bounding box there are lots of 404 status codes because it tries to load tiles that don't exist.

**Expected behavior**
Only tiles inside the bounding box should be loaded.

MoonE:
Fixes #11350
This (more or less) adds the code back that used the `WGS84BoundingBox` to determine the source extent. As noted in #10979
https://github.com/openlayers/openlayers/blob/e4be8309f7880c322c9850a634dab8b6adc30c62/src/ol/source/WMTS.js#L380-L398

If a `WGS84BoundingBox` is set the used extent is the intersection of the tile matrix extent and the `WGS84BoundingBox`.
This also adds back the `wrapX` detection based on the bounding box. Before `wrapX` was always set to `false`. Now it is set to true if the projection can wrap x and the extent bounding box spans the entire world horizontally.

### DIFF
--- a/src/ol/source/WMTS.js
+++ b/src/ol/source/WMTS.js
@@ -6,9 +6,10 @@ import TileImage from './TileImage.js';
 import WMTSRequestEncoding from './WMTSRequestEncoding.js';
 import {appendParams} from '../uri.js';
 import {assign} from '../obj.js';
+import {containsExtent, getIntersection} from '../extent.js';
 import {createFromCapabilitiesMatrixSet} from '../tilegrid/WMTS.js';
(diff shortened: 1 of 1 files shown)


MoonE:
Fixed by #11405

PR #12106, opened 2021-03-13
improve text width calculation

### LINKED ISSUE #12101: Text Style: placement 'line' not considering multi-line strings
**Describe the bug**
Text Style: when placement is set to 'line', multi-line strings are not considered when calculating the width of the text, affecting the chosen location for the label

**To Reproduce**
1. Go to [https://8nlub.csb.app](https://8nlub.csb.app)
2. Notice how the label for the line that is most to the north is not centred. This happens when the text has more than one line - 'Roussel\nRoussel' in the example.

**Expected behavior**
For one line text the label is centred on the best segment. Same would be expected for text with multiple lines.

Problem appears to be in measureAndCacheTextWidth() in canvas.js - it's not considering multi-line text
Calculating the width using the lines in the text seems to fix the issue:
```
    const lines = text.split('\n');
    const width = measureTextWidths(font, lines, []);
```

Can create a PR if you think this seems the right fix.

luiscamachopt:
Fixes #12101
now considers new lines to match rendering

### DIFF
--- a/src/ol/geom/flat/textpath.js
+++ b/src/ol/geom/flat/textpath.js
@@ -104,6 +104,9 @@ export function drawTextOnPath(
     return result;
   }
 
+  // rendering across line segments
+  text = text.replace(/\n/g, ' '); // ensure rendering in single-line as all calculations below don't handle multi-lines
+
   for (let i = 0, ii = text.length; i < ii; ) {
     advance();
     let angle = Math.atan2(y2 - y1, x2 - x1);
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -362,7 +362,9 @@ export function measureAndCacheTextWidth(font, text, cache) {
   if (text in cache) {
     return cache[text];
   }
-  const width = measureTextWidth(font, text);
+  const width = text
+    .split('\n')
+    .reduce((prev, curr) => Math.max(prev, measureTextWidth(font, curr)), 0);
   cache[text] = width;
   return width;
 }
--- a/test/node/ol/geom/flat/textpath.test.js
+++ b/test/node/ol/geom/flat/textpath.test.js
@@ -242,4 +242,59 @@ describe('ol/geom/flat/drawTextOnPath.js', function () {
     expect(instructions[0][3]).to.be((45 * Math.PI) / 180);
     expect(instructions.length).to.be(1);
   });
+
+  it('renders multi-line in one segment', function () {
+    const foo = 'foo';
+    const bar = 'bar';
+    const text = foo + '\n' + bar;
+    const pathLength = lineStringLength(angled, 2, angled.length, 2);
+    const textLength = measureAndCacheTextWidth('', bar);
+    const textAlign = 0.5;
+    const startM = (pathLength - textLength) * textAlign;
+    const instructions = drawTextOnPath(
+      angled,
+      0,
+      angled.length,
+      2,
+      text,
+      startM,
+      Infinity,
+      1,
+      measureAndCacheTextWidth,
+      '',
+      {}
+    );
+    expect(instructions[0][3]).to.be((45 * Math.PI) / 180);
+    expect(instructions[0][4]).to.be(text);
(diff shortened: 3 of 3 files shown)


luiscamachopt:
Would appreciate feedback on the PR if possible

ahocevar (commented):
This only fixes label placement in cases where lines are not bent, and will lead to even more unexpected behavior when they are. As @MoonE suggested in #12101, I think a better fix would be to replace `'\n'` with `' '` when `placement: 'line'` is set.

timmayo-ds:
Would really be good to have a fix for this problem as it is a major limitation for our application

ahocevar:
The solution in this pull request is not ideal, because it modifies the `measureTextWidth()` function to handle multi-line text, instead of changing the text-along-line rendering code to handle multi-line texts. This is a performance problem, because other renderer code is written in a way to use `measureTextWidth()` when single line text is to be measured, and `measureTextWidths()` for multi line text.

luiscamachopt:
@ahocevar thank you for the additional feedback

>  and `measureText()` for more complex measurements.

Did you mean `measureTextWidths()` here? 

In that case, would only modifying `measureAndCacheTextWidth()` to have 

```
    const lines = text.split('\n');
    const width = measureTextWidths(font, lines, []);
```
instead of
```  
    const width = measureTextWidth(font, text);
```

work for you? This is the method used by the rendering code.
Modifying the rendering code to do the line splitting when calculating the text width (as done in `measureTextWidths()`) would probably reduce performance as the line splitting would have to be done each time before hitting the cache.

I'll update the PR if this seems better

ahocevar:
There is no `measureTextWidths()` function any more, the code that your pull request changed is much different now after the introduction of rich text labels with #13410.

I still think that a better solution to this problem would be to replace `"\n"` with `" "` in texts with line placement. The use case you're describing could be solved using one text style per line, and work with offsets. That would also work (to some extent) with bent lines.

luiscamachopt:
> There is no `measureTextWidths()` function any more, the code that your pull request changed is much different now after the introduction of rich text labels with #13410.

Thank you, I've checked the new code and updated the PR for it.
Change is now in `measureAndCacheTextWidth`. For single line strings the only additional work is checking the string for instances of `"\n"` and this is cached so don't think there are performance implications and it fixes the width calculation to match the rendering.
 
> I still think that a better solution to this problem would be to replace `"\n"` with `" "` in texts with line placement. The use case you're describing could be solved using one text style per line, and work with offsets. That would also work (to some extent) with bent lines.

IMHO, this would add a fair degree of complexity (and probably worse performance?) to someone with the use case (multi-line text along straight lines - think graphs or schematics)
Might be better, as you say, for someone trying to use labels with multi-line text on bent lines (with angles below the threshold)

ahocevar:
I do see the straight lines use case you mention (graphs or schematics). To avoid surprises for the user, maybe we could find a way to combine your approach with the "replace `\n` with ` `" one: when `maxAngle` is set to `0`, use multi-line rendering, and single-line rendering otherwise.

@luiscamachopt Do you think you could extend your pull request to handle things that way?

luiscamachopt:
> I do see the straight lines use case you mention (graphs or schematics). To avoid surprises for the user, maybe we could find a way to combine your approach with the "replace `\n` with ` `" one: when `maxAngle` is set to `0`, use multi-line rendering, and single-line rendering otherwise.
> 
> @luiscamachopt Do you think you could extend your pull request to handle things that way?

@ahocevar I experimented with it and believe I found a good solution that doesn't even require setting `maxAngle` to `0` for the straight lines use case.
It now renders in single-line when the code determines that it needs to render across segments but leaves as multi-line if it all fits in one segment.

ahocevar (commented):
This is a very elegant solution. Thanks, @luiscamachopt. Do you think you'd be able to add tests for the new behavior?

ahocevar (approved):

ahocevar:
Thanks @luiscamachopt for this elegant solution!

PR #12135, opened 2021-03-22
Update webpack

tschaub:
This branch reworks things to work with webpack@5 and the latest rollup plugins.

The `examples/webpack/example-builder.js` plugin needed to be updated to conform with the new webpack plugin API.

The `tasks/serialize-workers.js` script needed to be updated to work with the latest babel plugin for rollup.

I've confirmed that examples are properly built and served with `npm start`, that the production build of the examples with `npm run build-examples` look good, that the legacy build with `npm run build-legacy` still works, and that `npm run build-package` still works.

The examples take about 20 seconds longer to build in production mode compared to building with webpack 4 (building examples in dev mode takes about 5 seconds longer).  I think it is a known issue that webpack 5 is slower (webpack/webpack#12102).  Maybe we can expect improvements in future releases.

I don't think we have to migrate things right now if we don't want to.  But I wanted to see what was involved in updating our plugins.

Closes #12133.


ahocevar (approved):
I think it makes sense to upgrade. Thanks for the investigation, and feel free to merge.

PR #12168, opened 2021-03-31
VectorSource#getFeatures() consistently returns a new array

ahocevar:
I just stumbled across an issue where I had a `Vector` source configured with `useSpatialIndex: false`. When such a source is used as `source` of a `Modify` interaction, the interaction will modify the array of the source's features `Collection`, bringing the collection out of sync with its `length`.

To avoid such issues, we should document what is returned by `VectorSource#getFeatures()`, and make sure we don't return the internal array of the features `Collection`.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -653,7 +653,7 @@ class VectorSource extends Source {
   }
 
   /**
-   * Get all snapshot of the features currently on the source in random order. The returned array
+   * Get a snapshot of the features currently on the source in random order. The returned array
    * is a copy, the features are references to the features in the source.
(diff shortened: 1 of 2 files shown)


src/ol/source/Vector.js
```diff
@@ -653,14 +653,15 @@ class VectorSource extends Source {
   }
 
   /**
-   * Get all features on the source in random order.
+   * Get all snapshot of the features currently on the source in random order. The returned array
```
marcjansen, line 656: ~~all~~ a

marcjansen (approved):
Looks good to me. I can you come up with a test verifying that you'll get a new array everytime, not a reference. That'd be appreciated.

Please merge with or without the test. Thanks!

PR #12175, opened 2021-04-03
Refresh available examples before each compilation

MoonE:
This makes it possible to pickup new examples and prevents compilation
errors when an example is removed. This is especially handy when switching
branches. Otherwise the dev server has to be restarted each time.


ahocevar (approved):
Excellent, thanks @MoonE!

PR #12178, opened 2021-04-04
Add WKB (Well-Known Binary) format

kikuchan:
Implements Well-Known Binary parser and writer.
This also supports EWKB used in PostGIS for example.


jahow:
Just wanted to say: very clean contribution, thanks!

mike-000:
Just one comment on the documentation.  You document the `hex` option as being used for controlling output, but it has a second purpose through its use in `getType()` which should also be documented, as that will determine whether the loader fetches a url or file as text or ArrayBuffer, for example in

```
new VectorSource({
   format: new WKB(),
   url: 'xxxxxxxx'
})
```

or when used with the `DragAndDrop` interaction.

src/ol/format/WKB.js
```diff
@@ -0,0 +1,866 @@
...
+    super();
+
+    const options = opt_options ? opt_options : {};
+
+    this.splitCollection = Boolean(options.splitCollection);
+
+    this.viewCache_ = null;
+
+    this.hex = options.hex !== false;
+    this.littleEndian = options.littleEndian !== false;
```
tschaub, line 666: Minor, but if these names are suffixed with `_` (e.g. `littleEndian_`) then they will be available for minification.

src/ol/format/WKB.js
```diff
@@ -0,0 +1,866 @@
...
+      }
+    }
+
+    writer.writeGeometry(
+      transformGeometryWithOptions(geometry, true, options),
+      SRID
+    );
+    const buffer = writer.getBuffer();
+
+    return this.hex !== false ? encodeHexString(buffer) : buffer;
```
tschaub, line 820:
```suggestion
    return this.hex ? encodeHexString(buffer) : buffer;
```

src/ol/format/WKB.js
```diff
@@ -0,0 +1,866 @@
...
+      ewkb: this.ewkb,
+
+      nodata: {
+        Z: this.nodataZ,
+        M: this.nodataM,
+      },
+    });
+
+    // extract SRID from `dataProjection`
+    let SRID = Number.isInteger(this.SRID) ? Number(this.SRID) : null;
```
tschaub, line 802: If we want IE support, you could check `typeof this.SRID === 'boolean'` instead.

mike-000:
For consistency with option naming conventions elsewhere in the library `srid` should be lower case in the same way as we use `url` as the option name when specifying a URL.

src/ol/format/WKB.js
mike-000: `Number.isInteger` is included the the polyfills (`ol/source/IIIF` uses it to check that numbers are integer values).

src/ol/format/WKB.js
kikuchan: Thanks! I renamed all of them.

kikuchan:
@mike-000 Thank you. It is really helpful!

tschaub (approved):

tschaub:
This is such a great contribution.  Thank you, @kikuchan.

kikuchan:
@tschaub Thank you so much!

PR #12180, opened 2021-04-04
Fix rendering of MultiPolygon text styles

MoonE:
When overflow is false and some of a MultiPolygon's polygons are
outside the rendered extent the coordinates and geometryWidths arrays are
not kept in sync. Therefore the width check will filter on wrong data and
some texts may not be rendered.

You can see the problem in https://openlayers.org/en/latest/examples/vector-layer.html
The text 'United States of America' on the main land disappears when you zoom in far enough.


ahocevar (approved):
Thanks for this fix, @MoonE! A regression test would be nice if you have time.

ahocevar (approved):

Issue #12160, opened 2021-03-29
OffscreenCanvas getFeatures document not defined


vittrup88:
**Describe the bug**
I'm running Vector Tiles in offscreencanvas and it is working. I'm running getFeatures for VectorTile layer and when only lines it works as expected. When trying to getFeatures over a point (with circle style) then I get the following error:

Uncaught (in promise) ReferenceError: document is not defined
    at createHitDetectionImageData (hitdetect.js:69)
    at CanvasVectorTileLayerRenderer.<anonymous> (VectorTileLayer.js:413)
    at new Promise (<anonymous>)
    at CanvasVectorTileLayerRenderer../node_modules/ol/renderer/canvas/VectorTileLayer.js.CanvasVectorTileLayerRenderer.getFeatures (VectorTileLayer.js:364)
    at VectorTileLayer../node_modules/ol/layer/Layer.js.Layer.getFeatures (Layer.js:201)
    at VectorTileLayer../node_modules/ol/layer/BaseVector.js.BaseVectorLayer.getFeatures (BaseVector.js:161)
    at VectorTileLayer../node_modules/ol/layer/VectorTile.js.VectorTileLayer.getFeatures (VectorTile.js:145)
    at getFeatures (vector-tile.worker.js:152)
    at vector-tile.worker.js:257

DOM is not accessible from worker.

**To Reproduce**
Will try to setup a simple example.

mike-000:
It seems to be related to decluttered items.  The dev version example https://openlayers.org/en/main/examples/offscreen-canvas.html is also broken.

ahocevar:
I don't think these two are related. The one that @mike-000 reports is fixed with #12164, the original poster's issue is related to the extra canvas that is created for the circle style's circle. More shimming would be needed to make hit detection work with offscreen canvas.

vittrup88:
I was looking at the code in hit detection and I see creating the document, setting width and height and getting the 2d context.
Could this simply be replaced by createCanvasContext2D ?
I have not done any testing of this yet but could/will do.
@ahocevar

ahocevar:
@vittrup88 If you can come up with a solution, a pull request would be much appreciated! Using `createCanvasContext2D` could definitely be part of the solution.

vittrup88:
@ahocevar 
I tried this out and it seems to work. Hope maybe you could help review the pull request. It is my first.

PR #12196, opened 2021-04-11
Hitdetection offscreencanvas worker

vittrup88:
This pull request removes dependency to document in hitdetecting. This enables getFeatures from offscreen canvas worker. #12160 
Please comment - thanks


examples/offscreen-canvas.worker.js
```diff
@@ -128,6 +127,20 @@ const maxTotalLoading = 8;
 const maxNewLoads = 2;
 
 worker.addEventListener('message', (event) => {
+  if (event.data.action === 'requestFeatures') {
+    layers.forEach((layer) => {
+      if (inView(layer.getLayerState(), frameState.viewState)) {
+        layer.getFeatures(event.data.pixel).then((res) => {
+          worker.postMessage({
+            action: 'getFeatures',
+            features: JSON.parse(stringify(res)),
```
MoonE, line 136: When the features are converted to and from json the class information is lost it would make sense to call getProperties() here and only pass that information, otherwise you have to rely on private properties to get any information out of that object.

examples/offscreen-canvas.js
```diff
@@ -120,6 +131,8 @@ worker.addEventListener('message', (message) => {
       );
     });
     image.src = event.data.src;
```
MoonE, line 133: Please include this change, the window.event is deprecated.
```suggestion
    image.src = message.data.src;
```

examples/offscreen-canvas.js
```diff
@@ -120,6 +131,8 @@ worker.addEventListener('message', (message) => {
...
+  } else if (message.data.action === 'getFeatures') {
+    console.log(message.data.features); // eslint-disable-line
```
MoonE, line 135: Would be nice if you could display the hit information in the actual web page in / below the map instead of logging it.
Maybe just the information if a feature is at the pointer position, or the type of feature to keep it simple.

MoonE (changes requested):
Your changes make sense to me and it runs when I try it locally though not in the deploy preview.
Please revert the changes to `package-lock.json` and squash your branch to one or two meaningful commits.
The changes to the package-lock.json happen because you are using an older version of npm, Upgrade to NodeJS >=15 with  npm 7 if you are able to.

examples/offscreen-canvas.js
vittrup88: I now added info window from example vector tile info

vittrup88:
I have messed up in git. so i'm creating a new pull request - sorry tried to squash the commits

PR #12198, opened 2021-04-12
Add attributionsCollapsible option to source/TileWMS and source/WMTS

### LINKED ISSUE #12197: source/TileWMS does not support the attributionsCollapsible option
`source/TileWMS` currently does not support setting the `attributionsCollapsible` option. This currently prevents developers from disabling collapsable attributions in favor of the static attribution text when using TileWMS. 

It should be possible to set the `attributionsCollapsible` option on an TileWMS source.

**Notes**
- The same issue also affects me when using `source/WMTS` (with an identical solution) for another map source. Should this be a separate issue?
- As the option is simply passed down to `source/TileImage` (which does support `attributionsCollapsible`), fixing it requires only a very minor change, so I've already applied it on my fork to allow me to continue working on my project. I'd be happy to make a PR out of it (or two, if you prefer a separate issue for WMTS).

sbx320:
Fixes #12197

### DIFF
--- a/src/ol/source/TileWMS.js
+++ b/src/ol/source/TileWMS.js
@@ -20,6 +20,7 @@ import {hash as tileCoordHash} from '../tilecoord.js';
 /**
  * @typedef {Object} Options
  * @property {import("./Source.js").AttributionLike} [attributions] Attributions.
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @sbx320!

PR #12214, opened 2021-04-14
Do not set invalid style

### LINKED ISSUE #12213: backgroundFill is bugged when body is flex
**Describe the bug**
When the body element of the page has the style "display: flex" the backgroundFill option for text becomes very big. Having flex on anything else doesn't produce the problem. I'm using 6.4.5 but other versions I've tested have the same issue.

**To Reproduce**
1. Create a page with a map
2. Add some text
3. Add "display: flex" to the body

image

**Code example**
Here is a jsfiddle demonstrating it. Remove the flex from body and if you want add a new div inside body with flex so you can se it doesn't matter if the map is encapsulated by a flex, it's just the body that gives problems.

https://jsfiddle.net/Doskii/29hxu4bd/

ahocevar:
We need to use block display for measuring text heights.

Fixes #12213.

### DIFF
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -289,7 +289,7 @@ export const measureTextHeight = (function () {
   /**
    * @type {HTMLDivElement}
    */
-  let div;
+  let measureElement;
   return function (fontSpec) {
     let height = textHeights[fontSpec];
     if (height == undefined) {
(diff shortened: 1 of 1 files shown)


MoonE (changes requested):
The current version did not werk because `!important` is not allowed for styles set directly on an element. The patch below should work.
```patch
diff --git a/src/ol/render/canvas.js b/src/ol/render/canvas.js
index 61780050a..e5f793324 100644
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -306,10 +306,10 @@ export const measureTextHeight = (function () {
         if (!div) {
           div = document.createElement('div');
           div.innerHTML = 'M';
-          div.style.margin = '0 !important';
-          div.style.padding = '0 !important';
-          div.style.position = 'absolute !important';
-          div.style.left = '-99999px !important';
+          div.style.margin = '0';
+          div.style.padding = '0';
+          div.style.position = 'absolute';
+          div.style.left = '-99999px';
         }
         div.style.font = fontSpec;
         document.body.appendChild(div);
```

ahocevar:
Wow, good find, @MoonE. You're absolutely right, I just force-pushed a commit with your suggestion, and it works indeed.

MoonE:
The pushed commit isn't quite my suggestion.
Margin can be removed and left can be removed, but border could be added in case it was set by a user style sheet.
See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth

Anyway, to be extra sure, it might as well be
```patch
diff --git a/src/ol/render/canvas.js b/src/ol/render/canvas.js
index 61780050a..3d5ea7a0e 100644
--- a/src/ol/render/canvas.js
+++ b/src/ol/render/canvas.js
@@ -306,10 +306,13 @@ export const measureTextHeight = (function () {
         if (!div) {
           div = document.createElement('div');
           div.innerHTML = 'M';
-          div.style.margin = '0 !important';
-          div.style.padding = '0 !important';
-          div.style.position = 'absolute !important';
-          div.style.left = '-99999px !important';
+          div.style.minHeight = '0';
+          div.style.maxHeight = 'none';
+          div.style.height = 'auto';
+          div.style.padding = '0';
+          div.style.border = 'none';
+          div.style.position = 'absolute';
+          div.style.display = 'block';
         }
         div.style.font = fontSpec;
         document.body.appendChild(div);

```
Might be hyper correct now.
Styles only need to be reset in case the user overides them in a style sheet, normally you wouldn't set these with a super general rule, but who knows...

ahocevar:
`left` was added intentionally to render the element outside the visible viewport. Feel free to modify my branch directly. Thanks!

ahocevar:
Thanks, looks good to me now. Do you agree, @MoonE ?

MoonE (approved):

Issue #11960, opened 2021-01-29
Improve overall typing of openlayers


simonseyock:
At the moment I am looking into how to improve the typing of openlayers to improve a) typechecking (also in dependent projects) and b) generation of typescript typings.

Options
----------

One huge issue that I have with the typing of openlayers are the `Options` objects of the layers. They all have a certain amount of strictly typed properties and can also accept arbitrary properties that get passed to the `BaseObject`. For typing purposes this bad.

What I would like to see is that all `Options` objects would get a property `properties` which would get passed to the `BaseObject`. I would like to look into this if you like the idea. This would be a breaking change and would only get included in v7. This mainly would affect the api of the layer types.

Nevertheless I found a way to type this properly in JSDoc, namely using
```
* @param {Options & Object<string, *>} opt_options Options.
``` 
Here is a PR for this: https://github.com/openlayers/openlayers/pull/11959

BaseObject
----------------

Maybe it is possible to incorporate `@template` Variables for the properties ( https://github.com/openlayers/openlayers/issues/11572 ).

Generic source types for layers
-----------------------------------------

I created a PR for improved generic source types of layers here: https://github.com/openlayers/openlayers/pull/11948
The `Layer` and `BaseVectorLayer` were already using it but all other types are missing it. This mainly improves the types of `getSource` and `setSource`, but also improves assignability. For example assignig an `ImageLayer` to a variable of a type `Layer<ImageSource>`.

I will continue to gather some issues here.

simonseyock:
Sadly `{Options & Object<string, *>}` is no valid JSDoc but is a valid typescript type and is therefore supported by the tsc (See https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#type ).

Would it be possible to remove this in `jsdoc-plugin-typescript` or would there need to be an adjustment to `valid-tsdoc` of eslint as well?

tschaub:
Another idea is to get rid of BaseObject altogether.  I would be in favor of doing this in a v7.

simonseyock:
Yes, it could be replaced with a property `properties` with `getProperties()`, `setProperties()` and a change event. It can than be easily of a generic type.

simonseyock:
This would only be relvant for layers and features.

simonseyock:
[...]

PR #12217, opened 2021-04-14
Fix building of examples with latest copy-webpack-plugin

MoonE:
This fixes the error when building examples observed in #12216

PR #12219, opened 2021-04-15
Add 'properties' to Layer constructors

ahocevar:
This pull request suggests a simple solution to the `BaseObject` problem lined out in #11960. It simply adds a `properties` property to constructors of `ol/layer/Base` and subclasses. This allows setting arbitrary observable properties that can later be retrieved with `#get()` or modified with `#set()` in a typed environment.

Example:
```js
const layer = new VectorLayer({
  properties: {
    name: 'My vector layer'
  },
  source: vectorSource
});

console.log(layer.get('name');
> "My vector layer"
```
would be equivalent to
```js
const layer = new VectorLayer({
  name: 'My vector layer', // type error
  source: vectorSource
});

console.log(layer.get('name');
> "My vector layer"
```

### DIFF
--- a/src/ol/layer/Base.js
+++ b/src/ol/layer/Base.js
@@ -27,6 +27,7 @@ import {clamp} from '../math.js';
  * visible.
  * @property {number} [maxZoom] The maximum view zoom level (inclusive) at which this layer will
  * be visible.
+ * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  */
 
 /**
@@ -50,6 +51,10 @@ class BaseLayer extends BaseObject {
      * @type {Object<string, *>}
      */
     const properties = assign({}, options);
+    if (typeof options.properties === 'object') {
+      delete properties.properties;
+      assign(properties, options.properties);
+    }
 
     properties[LayerProperty.OPACITY] =
       options.opacity !== undefined ? options.opacity : 1;
--- a/src/ol/layer/BaseImage.js
+++ b/src/ol/layer/BaseImage.js
@@ -27,6 +27,7 @@ import Layer from './Layer.js';
  * temporary layers. The standard way to add a layer to a map and have it managed by the map is to
  * use {@link module:ol/Map#addLayer}.
  * @property {import("../source/Image.js").default} [source] Source for this layer.
+ * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  */
 
 /**
--- a/src/ol/layer/BaseTile.js
+++ b/src/ol/layer/BaseTile.js
@@ -32,6 +32,7 @@ import {assign} from '../obj.js';
  * temporary layers. The standard way to add a layer to a map and have it managed by the map is to
  * use {@link module:ol/Map#addLayer}.
  * @property {boolean} [useInterimTilesOnError=true] Use interim tiles on error.
+ * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  */
 
 /**
--- a/src/ol/layer/BaseVector.js
+++ b/src/ol/layer/BaseVector.js
@@ -53,6 +53,7 @@ import {
  * batches will be recreated when no animation is active.
  * @property {boolean} [updateWhileInteracting=false] When set to `true`, feature batches will
  * be recreated during interactions. See also `updateWhileAnimating`.
+ * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  */
 
 /**
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -164,6 +164,7 @@ const INTERVALS = [
  * [30, 10]
  * ```
  * @property {boolean} [wrapX=true] Whether to repeat the graticule horizontally.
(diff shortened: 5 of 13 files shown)


ahocevar:
The examples build will be fixed when #12217 is merged.

simonseyock (approved):
I like this, this is much cleaner than clobbering the options with arbitrary user properties!

src/ol/Object.js
```diff
@@ -184,7 +184,11 @@ class BaseObject extends Observable {
    */
   setProperties(values, opt_silent) {
     for (const key in values) {
-      this.set(key, values[key], opt_silent);
+      if (key === 'properties') {
+        this.setProperties(values[key], opt_silent);
```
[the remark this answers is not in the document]
simonseyock: or maybe remove this logic from setProperties and change the call in the constructor instead? I think having this in the setProperties function is not really needed.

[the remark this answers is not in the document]
simonseyock:
```
if (opt_values !== undefined) {
  this.setProperties(opt_values);
}
```
to
```

if (opt_values !== undefined) {
  if (opt_values.properties !== undefined) {
    this.setProperties(opt_values.properties);
    delete opt_values.properties;
  }
  this.setProperties(opt_values);
}
```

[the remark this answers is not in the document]
ahocevar: I had something like that already, but was hesitant because usually we don't modify the object that gets passed to the constructor. But you convinced me that in this case it makes sense to do that, because it's just a workaround for typed environments.

src/ol/control/ZoomSlider.js
```diff
@@ -26,6 +26,7 @@ const Direction = {
  * @property {string} [className='ol-zoomslider'] CSS class name.
  * @property {number} [duration=200] Animation duration in milliseconds.
  * @property {function(import("../MapEvent.js").default):void} [render] Function called when the control
+ * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  * should be re-rendered. This is called in a `requestAnimationFrame` callback.
```
MoonE, line 30: This line belongs to the previous property description

src/ol/control/ZoomToExtent.js
```diff
@@ -15,6 +15,7 @@ import {fromExtent as polygonFromExtent} from '../geom/Polygon.js';
  * Instead of text, also an element (e.g. a `span` element) can be used.
  * @property {string} [tipLabel='Fit to extent'] Text label to use for the button tip.
  * @property {import("../extent.js").Extent} [extent] The extent to zoom to. If undefined the validity
+ * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  * extent of the view projection is used.
```
MoonE, line 19: This line also belongs to the previous property description

src/ol/control/ZoomSlider.js
ahocevar: Good find, thanks! Fixed.

src/ol/control/ZoomToExtent.js
ahocevar: Fixed.

MoonE (changes requested):
There's a bunch of cases where the options object with the properties config is not passed to the BaseObject constructor call, either adding the config or removing the jsdoc would be fine I guess.

ahocevar:
Thanks for the thorough review, @MoonE! Unfortunately I had prepared a commit already that fixes all the parent constructor calls without arguments, but I got dragged away from this pull request and must have lost the stash somehow. I'm hoping to be able to re-add these changes soon.

ahocevar:
I reworked this pull request, making the `properties` option only available to `ol/layer/Base` and its subclasses. Currently, these constructors are the only ones that support passing arbitrary properties in the constructor.

src/ol/layer/Base.js
```diff
@@ -50,6 +51,10 @@ class BaseLayer extends BaseObject {
      * @type {Object<string, *>}
      */
     const properties = assign({}, options);
+    if (typeof options.properties === 'object') {
+      delete properties.properties;
+      assign(properties, options.properties);
+    }
```
MoonE, line 57: Would it make sense to reorder the assignments so that the values int the properties object will not overwrite the configured options? (Did not test this:)
```suggestion
    const properties = assign({}, options.properties, options);
    if (typeof options.properties === 'object') {
      if ('properties' in options.properties) {
        properties.properties = options.properties.properties;
      } else {
        delete properties.properties;
      }
    }
```

simonseyock: The configured options are not affect by the original code, the `options` object is not affected. It actually is the other way around, a little further down the properties are overwritten with certain values.
Maybe we could add a note that it is not possible to set a property that also is an option.

simonseyock (approved):
I think this is a nice immediate step to support better automatically generated typing.

I was wondering if it is ok that anyone who was using a property `properties` will have problems. That is another problem of the base object properties mixed with the actual options.

ahocevar:
I added another test to show that users can supply a `properties` property under `properties`. As for the other raised concerns, these also exist without introducing a `properties` object, e.g. whether a user supplied property gets overridden by an internal property depends on the order.

I'd appreciate a pragmatic mindset when reviewing this non-breaking change. In the medium to long term, I'd still suggest we get rid of the base object and observable properties, as a breaking change.

src/ol/layer/Base.js
MoonE: You are right, this is not needed.

MoonE (approved):
Looking good! Thanks @ahocevar.

PR #12223, opened 2021-04-16
generic source parameters for options

simonseyock:
This allows automatic inference of source types if provided in options.
```
const layer = new ImageLayer({
  source: new ImageWMS()
});
// layer is now of type `ImageLayer<ImageWMS>`
```

Also in TypeSript this disallows construction with incorrect source types.
```
const layer = new ImageLayer<ImageWMS>({
  source: new ImageSource()
});
// was allowed before, should result in a type error now
```

Here is a picture of the resulting jsdoc. It is problematic that the type of the source is now seen as the generic parameter (`ImageSourceType`). Also the generic parameters are not documented.

It would be possible to replace `ImageSourceType` with the type that it has to be at least (`module:ol/source/Image`).

generic_options

### DIFF
--- a/config/jsdoc/plugins/inline-options.cjs
+++ b/config/jsdoc/plugins/inline-options.cjs
@@ -5,7 +5,48 @@
  * Inlines option params from typedefs
  */
 
-const properties = {};
+const docletInfos = {};
+
+/**
+ * This parses the comment for `@template` annotations and returns an object with name / type pairs for all template
+ * values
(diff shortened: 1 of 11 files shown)


simonseyock:
I added a simple parsing of the comments for template value name / type pairs that will result in docs that look like before. Still missing any template parameter, though.

generic_options2

ahocevar (approved):
Thanks for this improvement, @simonseyock!

Issue #12226, opened 2021-04-16
PointerEvent constructor call not supported in ie11


MoonE:
**Describe the bug**
after fixing  #12225 an error is thrown in 
https://github.com/openlayers/openlayers/blob/1554bc4cc27ed087027e23071bc0e90d643937f7/src/ol/MapBrowserEventHandler.js#L247
> Object doesn't support this action

**To Reproduce**
1. Apply patch in #12225
2. Go to box-selection.html
3. Click in map

**Expected behavior**
It should handle events without error.

ahocevar:
Could be fixed by using `MSPointerEvent` for IE11.

MoonE:
That's strange, it does not seem to work for me, even though it is listed as supported...

I tried the example code on the page you linked and get an error with 
`new PointerEvent(...)` `new MSPointerEvent(...)` and `document.createEvent(....)`
```js
var downEvent = new PointerEvent("pointerdown",
   {pointerId: 1,
    bubbles: true,
    cancelable: true,
    pointerType: "touch",
    width: 100,
    height: 100,
    isPrimary: true
   });
```

ahocevar:
I'm wondering if it wouldn't be time to stop supporting Internet Explorer.

mike-000:
Another polyfill may be needed.  Using https://cdn.jsdelivr.net/npm/events-polyfill@2.1.2 fixes the MDN example.

MoonE:
That's better, no errors with this. But the simulated events like doubleclick still don't work. It looks like the target property can't be set here.
https://github.com/openlayers/openlayers/blob/1554bc4cc27ed087027e23071bc0e90d643937f7/src/ol/MapBrowserEventHandler.js#L247-L251

mike-000:
#12319 gets it working in IE, but what about devices needing the elm-pep polyfill?  Is the event ever dispatched or any methods such as preventDefault/stopPropagation called?  If never dispatched an object containing the required properties and void methods typecast as PointerEvent might be sufficient.

ahocevar:
I created #12330 as a simpler alternative to #12319.

PR #12227, opened 2021-04-16
Fix event handling for map in shadow root

### LINKED ISSUE #12225: document.contains not supported in ie11
**Describe the bug**
`document.contains` not supported in ie11
https://developer.mozilla.org/en-US/docs/Web/API/Node/contains#browser_compatibility
https://github.com/openlayers/openlayers/blob/1554bc4cc27ed087027e23071bc0e90d643937f7/src/ol/PluggableMap.js#L990-L1000

**To Reproduce**
1. Go to https://openlayers.org/en/main/examples/box-selection.html
2. Click in the map
3. See error

**Expected behavior**
This seems to work, not sure if there is a case where this gives a different result.
```patch
index 7c35c4374..fc8d9be34 100644
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -996,7 +996,7 @@ class PluggableMap extends BaseObject {
         // It's possible for the target to no longer be in the page if it has been removed in an
         // event listener, this might happen in a Control that recreates it's content based on
         // user interaction either manually or via a render in something like https://reactjs.org/
-        !this.getOwnerDocument().contains(target)
+        this.getOwnerDocument().body.contains(target)
       ) {
         return;
       }
```

MoonE:
Fixes #12225 
#12226 still needs to be fixed so it is usable in ie11.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -987,6 +987,10 @@ class PluggableMap extends BaseObject {
       eventType === EventType.WHEEL ||
       eventType === EventType.KEYDOWN
     ) {
+      const doc = this.getOwnerDocument();
+      const rootNode = this.viewport_.getRootNode
+        ? this.viewport_.getRootNode()
+        : doc;
       const target = /** @type {Node} */ (originalEvent.target);
       if (
         // Abort if the target is a child of the container for elements whose events are not meant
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
Unfortunately this change breaks the [`es2015-custom-element.html` example](https://deploy-preview-12227--ol-site.netlify.app/examples/es2015-custom-element.html) and ShadowRoot handling.

MoonE:
That example was broken since 83ee87581f04a4687e0bab6fba524b74e893fb0d. I reverted that change from there.

ahocevar (approved):
Great work @MoonE, and sorry for my mess. Please merge.

PR #12229, opened 2021-04-18
Prevent TypeScript Compiler from erroring out due to lib checks

ahocevar:
Fixes a build issue when running `npm run build-package` caused by dependencies:
```
node_modules/@types/eslint/index.d.ts:302:41 - error TS2724: '"/Users/ahocevar/projects/openlayers/node_modules/@types/estree/index"' has no exported member named 'ChainExpression'. Did you mean 'ThisExpression'?

302         ChainExpression?: (node: ESTree.ChainExpression & NodeParentExtension) => void;
                                            ~~~~~~~~~~~~~~~

node_modules/@types/eslint/index.d.ts:325:42 - error TS2694: Namespace '"/Users/ahocevar/projects/openlayers/node_modules/@types/estree/index"' has no exported member 'ImportExpression'.

325         ImportExpression?: (node: ESTree.ImportExpression & NodeParentExtension) => void;
                                             ~~~~~~~~~~~~~~~~


Found 2 errors.
```

### DIFF
--- a/config/tsconfig-build.json
+++ b/config/tsconfig-build.json
@@ -51,11 +51,12 @@
     // "sourceRoot": "",                      /* Specify the location where debugger should locate TypeScript files instead of source locations. */
     // "mapRoot": "",                         /* Specify the location where debugger should locate map files instead of generated locations. */
     // "inlineSourceMap": true,               /* Emit a single file with source maps instead of having a separate file. */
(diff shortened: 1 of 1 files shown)


tschaub:
I think this might happen if you are using npm < 7.  In that case, I think we don't get the right version of `@types/estree` installed.

Actually, now that I'm looking at the `package-lock.json`, perhaps this isn't the issue - but it feels familiar and maybe related.

ahocevar:
I am using npm 7, but I may have messed up package-lock.json in #12224.

PR #12238, opened 2021-04-23
Use proxy url as interim fix for mixed content in the MapGuide example

### LINKED ISSUE #12233: Example "MapGuide" not working due to "Mixed Content" (https/http)
Not sure if this is an issue - please feel free to delete this issue.

This example page

https://openlayers.org/en/latest/examples/mapguide-untiled.html

does not work, because the protocol of the individual map agent uses "http" only.

Would love to know and see what it does. Also: there is a "Library://" protocol – never seen that before. Is that an official protocol or for internal usage only?

mike-000:
Fixes #12233

Keeps the example working in modern browsers until @jumpinjackie is able to update the demo server to an https domain.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12239, opened 2021-04-23
Add example showing how to Draw and Modify geodesic circles

mike-000:
Perhaps are more useful demonstration of the differences between Circle geometry and circular polygon than #11951/#11926


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12241, opened 2021-04-25
Use modules everywhere

### LINKED ISSUE #10821: package.json type module
**Describe the bug**
Using the new (but still exprimental) node esm loading mechanism does not work with the current version of  ol's package.json.

**To Reproduce**
1. `npm i ol`
2. in index.js 
  `import { createXYZ } from 'ol/tilegrid.js';`
3. `node --experimental-modules index.js` gives the following error:
> Cannot use import statement outside a module

**Expected behavior**
The module should be imported.

**Fix**
Add `"type": "module"` to the package.json

### LINKED ISSUE #12186: Make OpenLayers Node.js friendly
The OpenLayers is packed as ES module.
Despite recent Node.js can load ES modules directly, the Node cannot load it.
Because the Node requires `"type": "module"` should be placed in the package's `package.json` to load ES module.

I think this is the root cause of #12148, but the title should be more accurate.

Add `"type": "module"` to the `package.json` in the package.

I've found that modifying `package.json` in the repo break toolchains used in everywhere else, so I suggest this should be done in the `task/prepare-package.js` like below, for example.
```
--- tasks/prepare-package.js
+++ tasks/prepare-package.js
@@ -12,6 +12,9 @@ const utilSrc = fs
   .replace(versionRegEx, `var VERSION = '${pkg.version}';`);
 fs.writeFileSync(utilPath, utilSrc, 'utf-8');

+// Make it Node.js friendly
+pkg.type = 'module';
+
 // write out simplified package.json
 delete pkg.scripts;
 delete pkg.devDependencies;
```

This patch works perfectly on my side so far.

I have no idea whether it breaks other products or bundlers, but I believe it is not so harmful to them.

tschaub:
This updates our `package.json` to use modules everywhere.

## Server-side use in Node.js

This branch makes it possible for people to use (untranspiled) modules from the `ol` package directly in Node.  Node 12.17 and beyond can `import` ES modules directly without a flag.  It was not possible to `import` or `require` (untranspiled) modules from the `ol` package previously in Node because we don't have a `"type": "module"` in our published `package.json`.  The default behavior in Node is to treat packages as CommonJS packages, and we use `import` in our package (which doesn't work in Node with CommonJS modules).

## Client-side use with a bundler

This branch doesn't change anything for people who were using `import` and the `ol` package with a bundler like Rollup, Parcel, webpack, or esbuild.

See the examples below:
 * with esbuild openlayers/ol-esbuild#2
 * with Rollup openlayers/ol-rollup#5
 * with Parcel openlayers/ol-parcel#1
 * with webpack openlayers/ol-webpack#3

It is also still possible to use `require` to import `ol` modules with webpack, Rollup, etc.  I'm not sure if people are doing that, but before and after this change, you can still do things like `const Map = require('ol/Map').default;`.

## Developing in the OpenLayers repo

There are a few things to understand about how this changes scripts and tooling within the OpenLayers repo.  By adding `"type": "module"` to our `package.json`, we can no longer use `require` to import `*.js` scripts in the repo.  And the same goes for other dev tooling.  We _can_ use `require` if we need to by using the `*.cjs` extension on scripts.  Alternatively, we can place `package.json` files with `"type": "commonjs"` in the repo in directories where we need to author CommonJS modules instead of ES modules.  One of these two strategies is required to work with tooling that doesn't yet support ES modules.

### Webpack ES module support

As mentioned above, for client-side module bundling webpack works well with ES modules.  However, webpack has limited support for ES modules in things like loaders or plugins (see webpack/webpack#13233).  We can work around this by naming our custom loader(s) with a `*.cjs` extension.

### Karma ES module support

Karma doesn't support authoring config files as ES modules (see karma-runner/karma#3677).  As above, we can work around this by naming our config file `karma.config.cjs`.

### JSDoc ES module support

JSDoc doesn't support authoring plugins or templates with ES modules.  Because JSDoc assumes that a template will have a `publish.js` file, we cannot work around this by renaming our files to have a `*.cjs` extension.  Instead, we add a `package.json` with `"type": "commonjs"` to the root of our JSDoc config directory.

### Mocha ES module support

For in-browser tests, Mocha works as it did before - with the webpack transpiled modules.  This branch also adds the ability to run tests directly in Node.  For the Node-based tests, Mocha works with ES modules in single-run mode but not in watch mode (see mochajs/mocha#4374).  The tests run fast enough in Node that I don't think this is a problem, but it will be nice when there is a solution.

I've added a `test/node` directory where we can place tests that are not browser-specific.  For now, I've just copied in a handful of the `test/browser` tests.  We could decide that all tests that don't do DOM manipulations should run directly in Node (this is a much simpler setup than the Karma/webpack setup).  If others think this is a good idea, I could continue porting some of the existing Karma tests to run directly in Node.

You can try out the results with `npm install ol@module`.

Fixes #10821.
Fixes #12186.


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
tschaub, line 1: I added this to [configure the version of Node](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript) used in our Netlify builds.  If we don't want this file in the repo, we can instead set a `NODE_VERSION` env var before building the site.

A potential benefit of having this file here is to set the version of Node for others that use `nvm`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ol",
-  "version": "6.5.1-dev",
+  "version": "6.5.1-module.1",
```
tschaub, line 3: I'll change this back if we decide to merge.  I've just done this here to test out `ol@module` tagged packages.

tschaub:
I'll look into the failing `ol.renderer.webgl.PointsLayer` test later.  It also looks like I'll need to restore sourcemaps for the Karma tests.

tschaub:
Next puzzle is to figure out why the rendering tests never finish now.  Last successful run was https://github.com/openlayers/openlayers/runs/2444136446 (for be806f6c9f5753a4782755f7827c69fea3d527d5).  The run https://github.com/openlayers/openlayers/runs/2464262655 for a rebased version of the same commit (0b2ec1e747f11db32d119051985e01a32cbceed1) appears to be hung.

tschaub:
Porting over a handful of tests from `test/browser` to `test/node`, it looks like the test runtime for the Karma/webpack tests is reduced by about 15 seconds (down to about 2 minutes) and the Node test runtime is about 3 seconds.

ahocevar (approved):
Very impressive work @tschaub, thanks for your effort on this! I like the idea of having tests that do not rely on the DOM or other browser APIs in `test/node`. Moving more tests to `test/node` could also be done in a separate pull request and should not be blocking this, if you want to get this pull request in.

PR #12250, opened 2021-04-27
Do not set projection extent in WMTS optionsFromCapabilities

### LINKED ISSUE #11738: Failed optionsFromCapabilities call corrupts the projection

mike-000:
Fixes #11738
Replaces #11792

Leave setting projection extents to the application.

### DIFF
--- a/src/ol/source/WMTS.js
+++ b/src/ol/source/WMTS.js
@@ -502,10 +502,6 @@ export function optionsFromCapabilities(wmtsCap, config) {
       wgs84BoundingBox[2] === wgs84ProjectionExtent[2];
   }
 
-  if (projection.getExtent() === null) {
-    projection.setExtent(extent);
-  }
-
   const tileGrid = createFromCapabilitiesMatrixSet(
     matrixSetObj,
     extent,


ahocevar (approved):
Thanks, @mike-000!

lennerd:
Any plans for releasing this bugfix?

ahocevar:
@lennerd You can always `npm install ol@dev` to get the latest package from the main branch.

PR #12251, opened 2021-04-27
Test reorganization

tschaub:
This organizes the `rendering` and `browser` tests under the `test` directory, making room for `node` tests.

Since the diff here is pretty unbearable, here is what the two commits do:
 * Organize tests
   * moves things from `test` to `test/browser`
   * moves things from `rendering` to `test/rendering`
   * updates all import paths to point to the source
   * removes the unused `webpack.resolve.alias` from the `karma.config.js`
   * removes the `/rendering/` proxy from the `karma.config.js`
 * Remove unused rendering assertion from browser tests
   * removes pixelmatch and the associated assertions from the browser test extensions


ahocevar:
I looked into the issue with the two failing rendering tests. It seems that Puppeteer fails silently on those, because the resulting image is empty. The same happens for me locally. There are no errors, so I'm not sure how to investigate this further.

tschaub:
Thanks for looking into it.  Other tests were failing initially due to broken links when I moved the rendering dir (earthquake data was linked from examples).  I'll debug the remaining ones.

(the map is also blank when you load the test cases in a regular browser with `--interactive` - so not a Puppeteer issue)

tschaub:
That was a puzzle!  My overeager deleter got me in trouble again.  Looks like `ol-mapbox-style` requires the `ol` alias that I had removed from the webpack config.  That is restored in d72368667d3f6ecd72891dec5041ddeb851b4c08.

ahocevar:
Ok, that makes sense. Since `ol-mapbox-style` has `ol` as peer dependency and imports modules from `ol`, we need that alias. Thanks for the investigation and the fix!

ahocevar (approved):

marcjansen:
888 files changed, wow. Thanks

PR #12264, opened 2021-05-02
Run selected format tests in Node

tschaub:
This moves some of the format tests to run in Node.


ahocevar (approved):
Thanks for your continued effort on this, @tschaub!

PR #12266, opened 2021-05-02
Fix legacy build

### LINKED ISSUE #12265: Legacy build broken
**Describe the bug**
The legacy build generation is broken since #12241.

Something like this should do, but I'm getting errors for classes that do not have a default export, before it generated only warnings ...
```patch
diff --git a/tasks/generate-index.js b/tasks/generate-index.js
index c3e12f5ac..51c00f1f4 100644
--- a/tasks/generate-index.js
+++ b/tasks/generate-index.js
@@ -24,13 +24,13 @@ function getImport(symbol, member) {
   const defaultExport = symbol.name.split('~');
   const namedExport = symbol.name.split('.');
   if (defaultExport.length > 1 && defaultExport[0].indexOf('.') === -1) {
-    const from = defaultExport[0].replace(/^module\:/, './') + '.js';
+    const from = defaultExport[0].replace(/^module\:/, './');
     const importName = from.replace(/[.\/]+/g, '$');
-    return `import ${importName} from '${from}';`;
+    return `import ${importName} from '${from}.js';`;
   } else if (namedExport.length > 1 && member) {
-    const from = namedExport[0].replace(/^module\:/, './') + '.js';
+    const from = namedExport[0].replace(/^module\:/, './');
     const importName = from.replace(/[.\/]+/g, '_');
-    return `import {${member} as ${importName}$${member}} from '${from}';`;
+    return `import {${member} as ${importName}$${member}} from '${from}.js';`;
   }
 }
 ```
Before #12241 webpack generated warnings
```
WARNING in ./build/index.js 847:11-20
export 'default' (imported as '$ol$color') was not found in './ol/color' (possible exports: asArray, asString, fromString, isStringColor, normalize, toString)
```
After #12241 it generates errors and does not finish the build
```
ERROR in ./build/index.js 847:11-20
export 'default' (imported as '$ol$color') was not found in './ol/color.js' (possible exports: asArray, asString, fromString, isStringColor, normalize, toString)
```

**To Reproduce**
1. Run `npm run build-index`
2. The imported names in build/index.js do not match the names assigned to the `ol` object.

**Expected behavior**
It should generate a working legacy build.

MoonE:
Fixes #12265 
The build process failed because webpack does not allow failed imports
when it is run as a module.
This detects modules with default exports and only generates import
statements for default exports where they are available.


MoonE:
I added the github workflows changes from #12267

tschaub (approved):

tschaub:
Thanks for identifying the issue and for the fix, @MoonE.

tschaub (approved):

Issue #12281, opened 2021-05-07
View.calculateExtent is wrong with view padding


M393:
**Describe the bug**
`View` with `padding` returns wrong extent for `calculateExtent`.

**To Reproduce**
1. Go to https://codesandbox.io/s/calculate-extent-view-padding-kt76i?file=/main.js
3. See logged extent in console: `[-270, -90, 90, 90]`

**Expected behavior**
The calculated extent should be `[-180, -90, 180, 90]` (or alternatively `[-180, -90, 0, 90]` for the non-padded part?)

MoonE:
Looks like this was fixed by #11938, it returns `[-180, -90, 0, 90]` now.

PR #12282, opened 2021-05-07
Fix DragZoom with view padding

### LINKED ISSUE #12279: DragZoom does not work correctly with view padding
**Describe the bug**
When using DragZoom with view padding the resolution after zooming is wrong

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/view-padding.html
2. Use DragZoom (Shift + drag) around Switzerland
3. The padded view is centered on Switzerland, but the resolution is wrong (it matches the unpadded view size)

**Expected behavior**
The drag-zoom extent should fit into the padded view.

M393:
Fixes #12279 
The failing test is due to #12281

### DIFF
--- a/examples/view-padding.js
+++ b/examples/view-padding.js
@@ -50,24 +50,26 @@ const map = new Map({
   view: view,
 });
 
-const zoomtoswitzerland = document.getElementById('zoomtoswitzerland');
-zoomtoswitzerland.addEventListener(
-  'click',
-  function () {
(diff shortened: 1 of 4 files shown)


ahocevar (approved):
Thanks @M393 and @MoonE. This is a good fix! The changed behavior of `getResolutionForExtent` makes sense and is in line with the according previous change to `calculateExtent`.

@MoonE, please merge.

PR #12284, opened 2021-05-08
Layers should inherit group z-index

### LINKED ISSUE #6654: Documentation for zIndex in regards to ol.layer.Group needs clarification

### LINKED ISSUE #12059: LayerGroup zIndex is not inherited by the child layers
**Describe the bug**
I noticed that the zIndex of the LayerGroup is not inherited by the child layer added to it. Is this expected behaviour or a bug? Reason for this question is that i found an old bug report from years back of the same thing and there was a comment that it should be inherited. There was also PR made, but never merged?

https://github.com/openlayers/openlayers/issues/6654

**To Reproduce**
Example code can be provided if needed.

**Expected behavior**
Based on the LayerGroup documentation, zIndex of the group should be inherited by the child layers.

MoonE:
This will let layers in a layer group inherit the group's zIndex if no own zIndex is defined.

Somewhat based on #6661
Fixes #6654, fixes #12059

### DIFF
--- a/src/ol/layer/Group.js
+++ b/src/ol/layer/Group.js
@@ -205,12 +205,16 @@ class LayerGroup extends BaseLayer {
   }
 
   /**
-   * @param {Array<import("./Layer.js").State>} [opt_states] Optional list of layer states (to be modified in place).
+   * Get the layer states list and use this groups z-index as the default
+   * for all layers in this and nested groups, if it is unset at this point.
+   * If opt_states is not provided and this group's z-index is undefined
+   * 0 is used a the default z-index.
+   * @param {Array<import("./Layer.js").State>} [opt_states] Optional list
+   * of layer states (to be modified in place).
    * @return {Array<import("./Layer.js").State>} List of layer states.
    */
   getLayerStatesArray(opt_states) {
     const states = opt_states !== undefined ? opt_states : [];
-
     const pos = states.length;
 
     this.getLayers().forEach(function (layer) {
@@ -218,6 +222,10 @@ class LayerGroup extends BaseLayer {
     });
 
     const ownLayerState = this.getLayerState();
+    let defaultZIndex = ownLayerState.zIndex;
+    if (!opt_states && ownLayerState.zIndex === undefined) {
+      defaultZIndex = 0;
+    }
     for (let i = pos, ii = states.length; i < ii; i++) {
       const layerState = states[i];
       layerState.opacity *= ownLayerState.opacity;
@@ -241,6 +249,9 @@ class LayerGroup extends BaseLayer {
         } else {
           layerState.extent = ownLayerState.extent;
         }
+      }
+      if (layerState.zIndex === undefined) {
+        layerState.zIndex = defaultZIndex;
       }
     }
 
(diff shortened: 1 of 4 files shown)


test/browser/spec/ol/layer/group.test.js
```diff
@@ -497,6 +501,67 @@ describe('ol.layer.Group', function () {
...
+        ],
+      });
+
+      const layerStatesArray = group.getLayerStatesArray();
+      expect(layerStatesArray[0].zIndex).to.be(0);
+
+      disposeHierarchy(group);
+    });
+
+    it('returns undefined zindex in call to getLayerStatesArray when nested', function () {
```
tschaub, line 552: This description (“when nested”) looks more appropriate for the case above.

src/ol/layer/Group.js
```diff
@@ -205,19 +205,27 @@ class LayerGroup extends BaseLayer {
...
     const pos = states.length;
 
     this.getLayers().forEach(function (layer) {
       layer.getLayerStatesArray(states);
     });
 
     const ownLayerState = this.getLayerState();
+    let defaultZIndex = ownLayerState.zIndex;
+    if (!opt_states && ownLayerState.zIndex === undefined) {
+      defaultZIndex = 0;
```
tschaub, line 227: It isn’t clear to me why this is necessary to solve the problem.

test/browser/spec/ol/layer/group.test.js
MoonE: nested in this case referred to the `getLayerStatesArray([])` call with passed states array which indicates that it is called from a parent layer group..

I updated the tests, this was indeed confusing.

src/ol/layer/Group.js
tschaub: > At the end of the `getLayerStatesArray` call chain all layer states should have a defined zIndex

Kind of subtle that `getLayerState` will return `zIndex: undefined` and `getLayerStates` will return `zIndex: 0`.  I don't have my head enough into this to know if this is what we want or not.  Just that it tripped me up when reviewing the test changes.

tschaub (approved):

ahocevar:
Did you mean to merge this, @MoonE?

MoonE:
I do, thanks for the reminder.

PR #12306, opened 2021-05-12
Fix layer canvas reuse

### LINKED ISSUE #12305: Canvas never reused for layers with pixelRatio of 3
**Describe the bug**
When the canvas transform has more than 6 decimal places it is rounded by the browser. The comparison at
https://github.com/openlayers/openlayers/blob/fdfdc500aa041e2921c059298c04e7bad45619cf/src/ol/renderer/canvas/Layer.js#L99-L103
will fail because the transform matrix values `matrix(0.3333333333333333, 0, 0, 0.3333333333333333, 0, 0)` are rounded to `matrix(0.333333, 0, 0, 0.333333, 0, 0)`
**To Reproduce**
1. Go to https://codesandbox.io/s/pixelratio-canvas-reuse-ltz2u
2. See the number of canvas elements below the map

**Expected behavior**
The transform comparison should handle rounded numbers. Don't create additional canvas elements with different pixelRatio.

M393:
Fixes #12305

### DIFF
--- a/src/ol/transform.js
+++ b/src/ol/transform.js
@@ -1,6 +1,7 @@
 /**
  * @module ol/transform
  */
+import {WORKER_OFFSCREEN_CANVAS} from './has.js';
 import {assert} from './asserts.js';
 
 /**
@@ -265,11 +266,24 @@ export function determinant(mat) {
 }
 
 /**
- * A string version of the transform.  This can be used
+ * @type {HTMLElement}
+ * @private
+ */
+let transformStringDiv;
+
+/**
+ * A rounded string version of the transform.  This can be used
  * for CSS transforms.
  * @param {!Transform} mat Matrix.
  * @return {string} The transform as a string.
  */
 export function toString(mat) {
-  return 'matrix(' + mat.join(', ') + ')';
+  const transformString = 'matrix(' + mat.join(', ') + ')';
+  if (WORKER_OFFSCREEN_CANVAS) {
+    return transformString;
+  }
+  const node =
+    transformStringDiv || (transformStringDiv = document.createElement('div'));
+  node.style.transform = transformString;
+  return node.style.transform;
 }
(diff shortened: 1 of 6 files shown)


ahocevar (changes requested):
This change would break rendering in a worker, where `document.createElement()` is not available. For this reason, `ol/layer/renderer/canvas/TileLayer.js` and `ol/layer/renderer/canvas/ImageLayer.js` call `createTransformString()` before calling `useContainer()`. But this is missing for `ol/layer/renderer/canvas/VectorLayer.js`.

I agree that the right place to normalize the transform string would be `toString()` like in your pull request, but the implementation should be changed to handle the case where we are in a worker, like in `createTransformString()`. If you can change `toString()` this way, the `createTransformString()` function and its call sites can probably be removed.

MoonE:
Return with assignment will return the original value.

src/ol/transform.js
```diff
@@ -265,11 +266,23 @@ export function determinant(mat) {
 }
 
 /**
- * A string version of the transform.  This can be used
+ * @type {HTMLElement}
+ * @private
+ */
+let dummyNode;
```
ahocevar, line 272:
```suggestion
let transformStringDiv;
```

src/ol/transform.js
```diff
@@ -265,11 +266,23 @@ export function determinant(mat) {
...
  * @param {!Transform} mat Matrix.
  * @return {string} The transform as a string.
  */
 export function toString(mat) {
-  return 'matrix(' + mat.join(', ') + ')';
+  const transformString = 'matrix(' + mat.join(', ') + ')';
+  if (WORKER_OFFSCREEN_CANVAS) {
+    return transformString;
+  }
+  const node = dummyNode || (dummyNode = document.createElement('div'));
```
ahocevar, line 285:
```suggestion
  const node = transformStringDiv || (transformStringDiv = document.createElement('div'));
```

ahocevar (changes requested):
I'd suggest to rename `dummyNode` to `transformStringDiv`. Other than that, this pull request looks good.

ahocevar (approved):

ahocevar:
Thanks, @M393

PR #12315, opened 2021-05-13
Improve attribution accessibility

### LINKED ISSUE #12043: [accessibility] Attribution control accessibility improvements
After opening the collapsed attribution, the attribution links can not be navigated to in a logical order using a keyboard:

<table><td width="400"><video src="https://user-images.githubusercontent.com/26493779/108219710-44b7c900-7136-11eb-8fa3-f3468c46ff85.mp4"></video></td></table>

This is problematic for keyboard-only users and relates to <span title="Web Content Accessibility Guidelines">WCAG 2.1</span> Success Criterion [2.4.3 Focus Order](https://www.w3.org/TR/WCAG21/#focus-order) and [1.3.2 Meaningful Sequence](https://www.w3.org/TR/WCAG21/#meaningful-sequence).

Can the attribution be placed after the `<button>` attribution control in the DOM order?

Additionally, it'd be great if the [state](https://www.w3.org/TR/WCAG21/#dfn-states) was exposed to users of assistive technology via [`aria-expanded="(true|false)"`](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) on `<button>`.

MoonE:
Tab order of attribution text should be after the toggle button
Add aria-expanded state to the toggle button.

If the attribution is wider than the map it now stays to the left of the toggle button instead of jumping above:
https://deploy-preview-12315--ol-site.netlify.app/examples/bing-maps.html vs https://openlayers.org/en/v6.5.0/examples/bing-maps.html

Fixes #12043

### DIFF
--- a/src/ol/control/Attribution.js
+++ b/src/ol/control/Attribution.js
@@ -147,7 +147,7 @@ class Attribution extends Control {
      */
     this.toggleButton_ = document.createElement('button');
     this.toggleButton_.setAttribute('type', 'button');
-    this.toggleButton_.setAttribute('aria-expanded', String(false));
+    this.toggleButton_.setAttribute('aria-expanded', String(!this.collapsed_));
     this.toggleButton_.title = tipLabel;
     this.toggleButton_.appendChild(activeLabel);
 
--- a/src/ol/control/Attribution.js
+++ b/src/ol/control/Attribution.js
@@ -140,12 +140,18 @@ class Attribution extends Control {
 
     const activeLabel =
       this.collapsible_ && !this.collapsed_ ? this.collapseLabel_ : this.label_;
-    const button = document.createElement('button');
-    button.setAttribute('type', 'button');
-    button.title = tipLabel;
-    button.appendChild(activeLabel);
 
-    button.addEventListener(
+    /**
+     * @private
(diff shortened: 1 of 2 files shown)


Malvoz (approved):

src/ol/control/Attribution.js
```diff
@@ -140,12 +140,18 @@ class Attribution extends Control {
...
-    button.appendChild(activeLabel);
 
-    button.addEventListener(
+    /**
+     * @private
+     * @type {HTMLElement}
+     */
+    this.toggleButton_ = document.createElement('button');
+    this.toggleButton_.setAttribute('type', 'button');
+    this.toggleButton_.setAttribute('aria-expanded', String(false));
```
marcjansen, line 150:
```suggestion
    this.toggleButton_.setAttribute('aria-expanded', String(!this.collapsed_));
```

marcjansen (approved):
One tiny comment, looks good otherwise. Thanks.

src/ol/control/Attribution.js
MoonE: Makes sense to use the collapsed_ state for this especially as it can be initialized differently.

PR #12319, opened 2021-05-16
Include IE fallback for PointerEvent constructor

### LINKED ISSUE #12226: PointerEvent constructor call not supported in ie11

mike-000:
Fixes #12226

Include IE fallback for PointerEvent constructor and target property

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -252,7 +252,7 @@ class MapBrowserEventHandler extends Target {
       });
     } catch (error) {
       // fallback for IE11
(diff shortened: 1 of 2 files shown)


mike-000:
Closed in favour of #12330 as this only fixed IE and not other old browsers (e.g. old Android).

PR #12330, opened 2021-05-17
Do not require PointerEvent constructor

### LINKED ISSUE #12226: PointerEvent constructor call not supported in ie11

ahocevar:
Fixes #12226.

It is enough for the `down_` member in `MapBrowserEventHandler` to hold a copy of all properties of the original down event. The `preventDefault` and `stopPropagation` functions (and other functions on the event copy, if any) can be noop, because the event has been handled already.

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -8,6 +8,7 @@ import MapBrowserEventType from './MapBrowserEventType.js';
 import PointerEventType from './pointer/EventType.js';
 import Target from './events/Target.js';
 import {DEVICE_PIXEL_RATIO, PASSIVE_EVENT_LISTENERS} from './has.js';
+import {VOID} from './functions.js';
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
@ahocevar or @M393 - should we revisit this now that we are dropping IE support?  Doesn't seem urgent, but it would be nice to avoid warnings like the one described in #13862.

ahocevar:
Yes I think that would make sense. It should just be a matter of reverting the changes from this pull request.

PR #12334, opened 2021-05-19
More consistent example for ol/layer/MapboxVector

### LINKED ISSUE #12333: [API] Import different from usage
Not sure if this is a big deal. But it seams that in this example:

https://openlayers.org/en/latest/apidoc/module-ol_layer_MapboxVector-MapboxVectorLayer.html

the import is named differently - `MapboxVectorLayer ` - than its actual usage: `MapboxVector`.

(For the sake of consistency I would call it the same.)

MoonE:
Fixes #12333

### DIFF
--- a/src/ol/layer/MapboxVector.js
+++ b/src/ol/layer/MapboxVector.js
@@ -229,7 +229,7 @@ const SourceType = {
  *         zoom: 1
  *       }),
  *       layers: [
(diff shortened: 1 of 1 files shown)


rowild (approved):
Look good to me.

ahocevar (approved):

PR #12338, opened 2021-05-23
Do not use __proto__ to restore instances after main thread - worker transfer

### LINKED ISSUE #12317: Offscreen canvas example does not work in production mode
**Describe the bug**
There seem to be different minifications for the projection class with the webworker and normal example code.
These two objects cannot be compared:
image

It's caused by the changes to the example webpack config in #12176

**To Reproduce**
1. Go to https://deploy-preview-12313--ol-site.netlify.app/examples/offscreen-canvas.html
2. See the failed assertion
3. Go to the previous stack frame, as in the picture above

**Expected behavior**
The projections compared here should be equivalent.

ahocevar:
Fixes #12317.

The problem is that using `__proto__` to restore instances does not work here, because private members with `_` suffix are mangled, and are mangled differently in the main thread and the worker code. Since using `__proto__` isn't really good practice anyway, I now avoid its use by specifying the known projection directly.


MoonE (approved):
Thanks, @ahocevar

PR #12339, opened 2021-05-23
Fix confusing import in Mapbox Style example

mike-000:
Importing the default `olms` as `apply` is likely to be confused with importing `{apply}`.  Anyone doing so (e.g. with legacy build) would still get a map but no FullScreen control as the result would not support the `.then()` method.


ahocevar (approved):
Thanks, @mike-000

PR #12340, opened 2021-05-23
Fix raster example tooltips

MoonE:
This fixes the raster example's tooltips und upgrades d3 to the latest version.


ahocevar (approved):
Thanks, @MoonE

PR #12349, opened 2021-05-24
Document the input to an attribution function

mike-000:
It might be helpful to know what the function should use to produce its output


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12355, opened 2021-05-25
document descending order of resolutions array

### LINKED ISSUE #12354: If `resolutions` is set on view, `resolution` is ignored
**Describe the bug**
If a map is created with `resolutions`, the value of `resolution` is ignored. Instead `resolution` is set to the lowest value of `resolutions`. If `resolution` is not set, the map is not rendered.

**To Reproduce**
See console output of https://codesandbox.io/s/simple-forked-mhm4e?file=/main.js

**Expected behavior**
I would expect that the view would initialize with the set resolution.

simonseyock:
Fixes #12354

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -166,9 +166,11 @@ import {fromExtent as polygonFromExtent} from './geom/Polygon.js';
  * alternative to setting this is to set `zoom`. Layer sources will not be
  * fetched if neither this nor `zoom` are defined, but they can be set later
  * with {@link #setZoom} or {@link #setResolution}.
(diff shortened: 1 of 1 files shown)


mike-000:
Perhaps there should be an explanation of why, such as  "If specified the resolutions define the view zoom levels and must be in descending order."

simonseyock:
Yeah, maybe rephrase the whole thing, because the main purpose of `resolutions` is not the constraint, but setting the zoom levels. Good point!

tschaub (approved):

tschaub:
Thanks, @simonseyock.

PR #12357, opened 2021-05-26
Use calculateElevation function in Shaded Relief example

mike-000:
The Shaded Relief example does not explain how elevation is calculated, and as the source uses a non-standard encoding, this makes it difficult to use with other DEM sources.  Moving the calculation to a function and including an explanation of how the data is decoded and other common formats which might be encountered should make it more reusable.

### DIFF
--- a/examples/shaded-relief.js
+++ b/examples/shaded-relief.js
@@ -41,6 +41,18 @@ function shade(inputs, data) {
     aspect,
     cosIncidence,
     scaled;
+  function calculateElevation(pixel) {
+    // The method used to extract elevations from the DEM.
+    // In this case the format used is
+    // red + green * 2 + blue * 3
+    //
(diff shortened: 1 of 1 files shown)


examples/shaded-relief.js
```diff
@@ -41,6 +41,18 @@ function shade(inputs, data) {
     aspect,
     cosIncidence,
     scaled;
+  function calculateElevation(pixel) {
+    // The method used to extract elevations from the DEM.
+    // In this case the format used is
+    // red + green * 2 + blue * 3
+    //
+    // Other freqquently used methods include the Mapbox format
```
EvertEt, line 49:
```suggestion
    // Other frequently used methods include the Mapbox format
```

fredj (approved):

PR #12359, opened 2021-05-26
Fix broken link in docs

### LINKED ISSUE #12360: Broken external links in docs
**Describe the bug**
The links to https://developer.mozilla.org/en-US/docs/Web/API/ImageData in https://openlayers.org/en/latest/apidoc/module-ol_source_Raster.html#~Operation are broken as %7D is being added to the url.  The problem persists despite attempts to fix in #12359.

mike-000:
Fixes #12360

ImageData links were not working


MoonE (approved):
Thanks, @mike-000

PR #12375, opened 2021-06-03
Fix Vector Tile Selection example on old browsers

mike-000:
Replace `string.startsWith(x)` with fully supported `string.indexOf(x) === 0`


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12378, opened 2021-06-06
Correct .split() calls to only split numeric srid for ArcgisRest

### LINKED ISSUE #12377: Allow custom WKT string projection parameters (BBOXSR and IMAGESR) for ImageArcGISRest sources

My problem is dealing with rendering an Esri "dynamic map image service" with custom LayerDefinitions. I'm struggling to get these layers to draw Pacific-spanning Esri service datasets across the antimeridian / 180 longitude / "dateline". By default the Esri services use the 3857 web mercator projection, and they don't support any "wrapping" across the edge meridians. I found a work around that works in the Esri REST API Export image test page, but it would require me to pass custom projection parameters. I've been able to use a WKT projection string with an alternate central meridian defined. The Esri service then sends back an image that fully crosses the antimeridian centered on wherever I wanted the image centered. The parameters in question are IMAGESR and BBOXSR. 

However, the OpenLayers implementation of the ImageArcGISRest layer source does not appear to allow a custom IMAGESR or BBOXSR. It only seems to allow passing an EPSG/SRID code to the Esri service as a parameter.

Here is the documentation that indicates that the BBOXSR and IMAGESR are set "dynamically" and can't be passed to the Esri REST service as custom parameters:
https://openlayers.org/en/latest/apidoc/module-ol_source_ImageArcGISRest-ImageArcGISRest.html
"ArcGIS Rest parameters. ... **BBOX, SIZE, BBOXSR, and IMAGESR will be set dynamically.** "

It seems you can force an alternate EPSG/SRID code to be sent as a parameter to the Esri Image Service, using the ImageArcGISRest source "projection" property, but that seems to only accept a "projection-like" object.

I'd like to send a custom projection string (in WKT format) as an optional parameter that would override these "dynamically" set REST parameters (BBOXSR and IMAGESR). 

One approach is to first check to use custom BBOXSR and IMAGESR parameters (if present) before falling back to the default BBOXSR and IMAGESR settings derived from the "projection" optional property.
This might happen in the "getRequestUrl_" method here?
https://github.com/openlayers/openlayers/blob/de81d2b7629e4331d5455c809c51a99d57202319/src/ol/source/ImageArcGISRest.js#L239

Alternatively, if there is an work around to get an Esri dynamic map service image request to span the Pacific basin crossing the antimeridian, I'm very interested!

mike-000:
Fixes #12377

Ensuring the split only works for numeric srid has the advantage that the projection code can be set to a custom Arcgis SpatialReference definition (which would contain a `:`) without it being broken.

### DIFF
--- a/src/ol/source/ImageArcGISRest.js
+++ b/src/ol/source/ImageArcGISRest.js
@@ -242,7 +242,7 @@ class ImageArcGISRest extends ImageSource {
     // ArcGIS Server only wants the numeric portion of the projection ID.
     // (if there is no numeric portion the entire projection code must
     // form a valid ArcGIS SpatialReference definition).
-    const srid = projection.getCode().split(/:(?=\d+$)/)).pop();
+    const srid = projection.getCode().split(/:(?=\d+$)/).pop();
 
     params['SIZE'] = size[0] + ',' + size[1];
     params['BBOX'] = extent.join(',');
--- a/src/ol/source/ImageArcGISRest.js
+++ b/src/ol/source/ImageArcGISRest.js
@@ -28,6 +28,8 @@ import {containsExtent, getHeight, getWidth} from '../extent.js';
  * https://developers.arcgis.com/rest/services-reference/export-map.htm
  * for further reference.
  * @property {import("../proj.js").ProjectionLike} [projection] Projection. Default is the view projection.
+ * The projection code must contain a numeric end portion separated by :
+ * or the entire code must form a valid ArcGIS SpatialReference definition.
  * @property {number} [ratio=1.5] Ratio. `1` means image requests are the size of the map viewport,
  * `2` means twice the size of the map viewport, and so on.
  * @property {Array<number>} [resolutions] Resolutions. If specified, requests will be made for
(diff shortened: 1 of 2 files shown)


ahocevar:
Currently this pull request only changes whitespace and adds newlines. Did you mean to do something else?

mike-000:
My intention was to prevent arbitrarily splitting custom codes which are meaningful to Arcgis but do not have a numeric srid (although they do contain `:`) https://codesandbox.io/s/arcgis-image-forked-kuo1u?file=/main.js

murdocha:
Does the new regex in this PR update allow a spatial reference param to be something as complex as:
```
{"wkt":"PROJCS[\"WGS_1984_Web_Mercator_Auxiliary_Sphere\",GEOGCS[\"GCS_WGS_1984\",DATUM[\"D_WGS_1984\",SPHEROID[\"WGS_1984\",6378137.0,298.257223563]],PRIMEM[\"Greenwich\",0.0],UNIT[\"Degree\",0.0174532925199433]],PROJECTION[\"Mercator_Auxiliary_Sphere\"],PARAMETER[\"False_Easting\",0.0],PARAMETER[\"False_Northing\",0.0],PARAMETER[\"Central_Meridian\",-71.68417519999998],PARAMETER[\"Standard_Parallel_1\",0.0],PARAMETER[\"Auxiliary_Sphere_Type\",0.0],UNIT[\"Meter\",1.0]]"}
```
(a WMAS projection with a custom central meridian around Massachusetts, US)
? 
That was hard for me to visually make out without testing the code against a string like this...

Currently only a "real" EPSG code can be passed in to the OL code for ImageArcGISRest sources.
That was the initial issue I reported earlier here:
#12377

ahocevar (approved):
Thanks, @mike-000.

PR #12394, opened 2021-06-09
Ensure VectorSource getUrl() corresponds to last setUrl()

### LINKED ISSUE #12046: A getUrl immediately following a setUrl will return the old url for some sources
The following code will produce
```
http://url1
http://url1
```
```js
const source = new ol.source.Vector({url: 'http://url1', format: new ol.format.GeoJSON()});
console.log(source.getUrl());
source.setUrl('http://url2');
console.log(source.getUrl());
```

<https://jsfiddle.net/mmomtchev/v2mo1as3/6/>

It is not really a problem outside of unit testing since the url will eventually get modified, but still, it is a puzzling behaviour. Not all sources do it - `Vector` does it.

mike-000:
Fixes #12046

Might be useful if someone needs to keep track of multiple sources which are being dynamically updated.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -1054,6 +1054,7 @@ class VectorSource extends Source {
    */
   setUrl(url) {
     assert(this.format_, 7); // `format` must be set when `url` is set
+    this.url_ = url;
     this.setLoader(xhr(url, this.format_));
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks, @mike-000

PR #12405, opened 2021-06-15
VectorTile getSourceTiles behaviour with loading source tiles

### LINKED ISSUE #12396: Vector tile layer wrapping regression
**Describe the bug**
There is a problem with vector tile layers not initially wrapping when a map opens with `multiWorld` set true..

**To Reproduce**
Compare https://openlayers.org/en/main/examples/vector-tile-selection.html with the correctly wrapping version https://openlayers.org/en/latest/examples/vector-tile-selection.html when opening on a wide screen.  In the dev version only one world is initially visible, but others can be revealed by panning or zooming out.  Looking at the deploy previews it seems this was introduced by #12364  There seems to be no problem if the vector tile layer shares a canvas with another correctly wrapped layer - comment out the osm `className` in https://codesandbox.io/s/vector-tile-info-forked-4v8l7 and the vector tile layer will also wrap correctly

**Expected behavior**
The `dev` version should wrap on opening as the `latest` version did..

ahocevar:
Fixes #12396

This pull request makes sure that source tiles are also considered when they are already/still loading. It also fixes a potential issue that could lead to unnecessary source tile creation due to too aggressive cache expiration on the source tile cache.

### DIFF
--- a/src/ol/source/VectorTile.js
+++ b/src/ol/source/VectorTile.js
@@ -471,6 +471,9 @@ class VectorTile extends UrlTile {
    */
   updateCacheSize(tileCount, projection) {
     super.updateCacheSize(tileCount * 2, projection);
+    this.sourceTileCache.highWaterMark = this.getTileCacheForProjection(
+      projection
+    ).highWaterMark;
   }
 }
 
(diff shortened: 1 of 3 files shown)


ahocevar:
I added another commit which fixes a related issue that might have gone unnoticed for a long time: the source tile cache size was set to `0` (which means infinite), and never adjusted properly.

This is now ready for review.

ahocevar:
@openlayers/core Thanks for any review on this. It fixes a severe issue which, in my opinion, is blocking the 6.6.0 release.

tschaub (approved):

ahocevar:
Thanks for the review, @tschaub!

Issue #12235, opened 2021-04-22
Modify interaction with hitDetection and features does not work correctly


M393:
**Describe the bug**
When a Modify interaction is configured with a hit detection layer and a features collection the vertex feature is created for features that are not in the collection and modifystart / end events are fired with an empty features collection.

**To Reproduce**
1. Go to https://codesandbox.io/s/modify-with-hitdetection-htcoe?file=/src/index.js
2. Drag the point in the map
3. The point can't be dragged, but there is a modifystart event fired and the vertex feature is visible.

**Expected behavior**
If the feature is not in the provided collection / source it should be ignored.

mike-000:
Events with empty features collections are also fired if you drag between vertices when inserting vertices is disabled https://codesandbox.io/s/modify-with-hitdetection-forked-4ku2p

M393:
One more problem I noticed is if you configure the interaction with `hitDetection: true` and you have a layer with `RenderFeature`s, an error will be thrown when `getCoordinates` is called on the feature because this method does not exist.
https://github.com/openlayers/openlayers/blob/7a2f87caca9ddc1912d910f56eb5637445fc11f6/src/ol/interaction/Modify.js#L1135-L1141

mike-000:
Another oddity, with hitDetection the interaction will detect Point geometries created by style functions https://codesandbox.io/s/modify-with-hitdetection-forked-h3pv5  If the feature geometry is not a point hitDetection should not detect it or fire an event (but does), if the feature geometry is a Point should it operate on that?

M393:
With `hitDetection: true` it detects hits on all features regardless of configured collection or source.
It properly ignores non-point features, but point features that are not in the source / collection will be added to the `nodes` array.
https://github.com/openlayers/openlayers/blob/7a2f87caca9ddc1912d910f56eb5637445fc11f6/src/ol/interaction/Modify.js#L1135-L1141
This is fixed by adding `includes(this.features_.getArray(), feature)` to the condition

PR #12408, opened 2021-06-17
Do not fire modifystart when nothing is being modified

ahocevar:
Fixes the problem described in https://github.com/openlayers/openlayers/issues/12235#issuecomment-825086303.


MoonE (approved):
Thanks @ahocevar

PR #12413, opened 2021-06-19
Fix hitdetection image invalidation

### LINKED ISSUE #11994: VectorLayer may use outdated hitdetect image
**Describe the bug**
The hitdetect image may not be updated when the map is moved horizontally and wrapX is enabled.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/hitdetect-vector.html
2. Open devtools
3. enable keyboard input with `document.getElementById('map').tabIndex = 0`
4. put focus on map by clicking on it
5. use left / right keys to  move map
6. move mouse to test hit detection

**Expected behavior**
It should hitdetect features under the mousepointer only.

MoonE:
Fixes #11994

### DIFF
--- a/src/ol/renderer/canvas/VectorLayer.js
+++ b/src/ol/renderer/canvas/VectorLayer.js
@@ -30,6 +30,7 @@ import {
   getSquaredTolerance as getSquaredRenderTolerance,
   renderFeature,
 } from '../vector.js';
+import {equals} from '../../array.js';
 import {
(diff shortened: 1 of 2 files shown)


ahocevar:
Good find and good fix, @MoonE! Could you maybe add a test? Thanks!

ahocevar (approved):
Looks good, if possible please add a test.

Issue #12412, opened 2021-06-18
heatmap-earthquakest example broken in production build


MoonE:
**Describe the bug**
In production builds the earthquake  heatmap example does not run.

**To Reproduce**
1. Go to https://openlayers.org/en/main/examples/heatmap-earthquakes.html
2. The map remains empty

The error happens in this line, not sure why, `getChangeEventType` seems to be missing
https://github.com/openlayers/openlayers/blob/e82a86f56337cfa394235df7528da032a5afbf30/src/ol/layer/Heatmap.js#L89-L92

**Expected behavior**
The map should load and show the heatmap clusters.

PR #12414, opened 2021-06-19
Replace getChangeEventType() with add/removeChangeListener methods

ahocevar:
Although the reason for #12412 is probably a Terser problem (which I was unable to isolate), I think we can make an improvement, which also makes that problem go away. I see no reason for caching event types, and instead of having to use the `getChangeEventType()` function, I like it better to have dedicated `addChangeListener()` and `removeChangeListener()` methods.

### DIFF
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -1,7 +1,7 @@
 /**
  * @module ol/Feature
  */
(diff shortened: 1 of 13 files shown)


MoonE (approved):
Thanks, @ahocevar. Looks better like this.

PR #12416, opened 2021-06-20
Use generics to limit event types in on(), once() and un()

ahocevar:
This pull request adds another improvement regarding type safety: The `ol/Observable` methods `on()`, `once()` and `un()` now only accept the event types that are actually available for the respective class.

In a code-aware editor, registering a listener on a `Modify` interaction would now look like this:

Untitled

My medium term goal with changes like this is to get rid of the `*EventType` enums (which are not part of the API anyway), and use string constant union types instead. Another possible improvement along these lines would be to use generics to show the correct signature of the listener function in `on()`, `once()` and `un()`, i.e. the correct `*Event` of the listener argument.

@simonseyock I think you might want to take a look at this pull request, and maybe you are also interested in working towards the goals I mentioned above.

### DIFF
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -59,7 +59,7 @@ export class CollectionEvent extends Event {
  * Collection as a whole.
  *
  * @fires CollectionEvent
- *
+ * @extends BaseObject<'add'|'remove'|'change:length'>
  * @template T
  * @api
  */
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -58,6 +58,7 @@ import {listen, unlistenByKey} from './events.js';
  *
  * @api
  * @template {import("./geom/Geometry.js").default} Geometry
+ * @extends BaseObject<'change:geometry'>
  */
 class Feature extends BaseObject {
   /**
--- a/src/ol/Geolocation.js
+++ b/src/ol/Geolocation.js
@@ -82,6 +82,7 @@ class GeolocationError extends BaseEvent {
  *     });
  *
  * @fires module:ol/events/Event~BaseEvent#event:error
+ * @extends BaseObject<'change:accuracy'|'change:accuracyGeometry'|'change:altitude'|'change:altitudeAccuracy'|'change:heading'|'change:position'|'change:projection'|'change:speed'|'change:tracking'|'change:trackingOptions'>
  * @api
  */
 class Geolocation extends BaseObject {
--- a/src/ol/Object.js
+++ b/src/ol/Object.js
@@ -78,6 +78,8 @@ export class ObjectEvent extends Event {
  * object.unset('foo').
  *
  * @fires ObjectEvent
+ * @template {string} EventTypes
+ * @extends Observable<'propertychange'|EventTypes>
  * @api
  */
 class BaseObject extends Observable {
--- a/src/ol/Observable.js
+++ b/src/ol/Observable.js
@@ -6,6 +6,11 @@ import EventType from './events/EventType.js';
 import {listen, listenOnce, unlistenByKey} from './events.js';
 
 /**
+ * @template ChildTypes
+ * @typedef {ChildTypes|'change'|'error'|Array<ChildTypes|'change'|'error'>} EventTypes
+ */
+
+/**
  * @classdesc
  * Abstract base class; normally only used for creating subclasses and not
  * instantiated in apps.
@@ -14,6 +19,7 @@ import {listen, listenOnce, unlistenByKey} from './events.js';
  * {@link module:ol/Observable~Observable#changed}.
  *
  * @fires import("./events/Event.js").default
+ * @template {string} ChildTypes
  * @api
  */
 class Observable extends EventTarget {
@@ -48,7 +54,7 @@ class Observable extends EventTarget {
 
   /**
    * Listen for a certain type of event.
-   * @param {string|Array<string>} type The event type or array of event types.
+   * @param {EventTypes<ChildTypes>} type The event type or array of event types.
    * @param {function(?): ?} listener The listener function.
    * @return {import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} Unique key for the listener. If
    *     called with an array of event types as the first argument, the return
@@ -70,7 +76,7 @@ class Observable extends EventTarget {
 
   /**
    * Listen once for a certain type of event.
-   * @param {string|Array<string>} type The event type or array of event types.
+   * @param {EventTypes<ChildTypes>} type The event type or array of event types.
    * @param {function(?): ?} listener The listener function.
    * @return {import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} Unique key for the listener. If
    *     called with an array of event types as the first argument, the return
@@ -94,7 +100,7 @@ class Observable extends EventTarget {
 
   /**
    * Unlisten for a certain type of event.
-   * @param {string|Array<string>} type The event type or array of event types.
+   * @param {EventTypes<ChildTypes>} type The event type or array of event types.
    * @param {function(?): ?} listener The listener function.
    * @api
    */
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -101,6 +101,7 @@ const Property = {
  *     popup.setPosition(coordinate);
  *     map.addOverlay(popup);
  *
+ * @extends BaseObject<'change:element'|'change:map'|'change:offset'|'change:position'|'change:positioning'>
  * @api
  */
 class Overlay extends BaseObject {
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -132,6 +132,7 @@ import {removeNode} from './dom.js';
  * @fires import("./render/Event.js").default#precompose
  * @fires import("./render/Event.js").default#postcompose
  * @fires import("./render/Event.js").default#rendercomplete
+ * @extends BaseObject<'change:layergroup'|'change:size'|'change:target'|'change:view'|'precompose'|'singleclick'|'click'|'dblclick'|'pointerdrag'|'pointermove'|'postrender'|'movestart'|'moveend'|'postcompose'|'rendercomplete'>
  * @api
  */
 class PluggableMap extends BaseObject {
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -288,6 +288,7 @@ const DEFAULT_MIN_ZOOM = 0;
  * the snap angle is 0), an animation will be triggered at the interaction end to
  * put back the view to a stable state;
  *
+ * @extends BaseObject<'change:center'|'change:resolution'|'change:rotation'>
  * @api
  */
 class View extends BaseObject {
--- a/src/ol/control/Control.js
+++ b/src/ol/control/Control.js
@@ -41,6 +41,8 @@ import {removeNode} from '../dom.js';
  * You can also extend this base for your own control class. See
  * examples/custom-controls for an example of how to do this.
  *
+ * @template {string} EventTypes
+ * @extends BaseObject<EventTypes>
  * @api
  */
 class Control extends BaseObject {
--- a/src/ol/control/FullScreen.js
+++ b/src/ol/control/FullScreen.js
@@ -66,6 +66,7 @@ const FullScreenEventType = {
  *
  * @fires FullScreenEventType#enterfullscreen
  * @fires FullScreenEventType#leavefullscreen
+ * @extends Control<'enterfullscreen'|'leavefullscreen'>
  * @api
  */
 class FullScreen extends Control {
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -50,6 +50,7 @@ const COORDINATE_FORMAT = 'coordinateFormat';
  * On touch devices, which usually do not have a mouse cursor, the coordinates
  * of the currently touched position are shown.
  *
+ * @extends Control<'change:coordinaetFormat'|'change:projection'>
  * @api
  */
 class MousePosition extends Control {
--- a/src/ol/control/ScaleLine.js
+++ b/src/ol/control/ScaleLine.js
@@ -69,6 +69,7 @@ const DEFAULT_DPI = 25.4 / 0.28;
  * When specifying `bar` as `true`, a scalebar will be rendered instead
  * of a scaleline.
  *
+ * @extends Control<'change:units'>
  * @api
  */
 class ScaleLine extends Control {
--- a/src/ol/interaction/DragAndDrop.js
+++ b/src/ol/interaction/DragAndDrop.js
@@ -82,6 +82,7 @@ export class DragAndDropEvent extends Event {
  * combinnation of formats read both text string and ArrayBuffer sources. Older browsers such
  * as IE which do not support this will need a TextDecoder polyfill to be loaded before use.
  *
+ * @extends Interaction<'addfeatures'>
  * @api
  *
  * @fires DragAndDropEvent
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -103,6 +103,7 @@ export class DragBoxEvent extends Event {
  * {@link module:ol/interaction/DragRotateAndZoom}).
  *
  * @fires DragBoxEvent
+ * @extends PointerInteraction<'boxcancel'|'boxdrag'|'boxend'>
  * @api
  */
 class DragBox extends PointerInteraction {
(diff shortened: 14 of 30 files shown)


simonseyock (approved):
I really like this! I think it is more readable and clearer.

simonseyock:
The correct signature would be great. I liked that about working with the `@hanreev/ol-types`. There for each event exists a method overload in the declaration somewhat like this:
```
on('singleclick', (e: MapBrowserEvent) => void);
on('postcompose', (e: RenderEvent) => void);
```
This might not possible at all with jsdoc, as the overloading capabilities are different in declaration and implementation mode in typescript.
If it is possible it is probably not possible to do this by generic parameters.

The best alternative I see at the moment is probably a second generic parameter that defines the actual types of the event objects. This could then result in something like this:
```
on('singleclick' | 'postcompose', (e: MapBrowserEvent | RenderEvent) => void);
```
or (maybe better as you maybe  would not need a guard inside the function):
```
on('singleclick' | 'postcompose', (e: MapBrowserEvent) => void |  (e: RenderEvent) => void);
```

ahocevar:
Thanks for sharing your experience here @simonseyock. What's missing in TypeScript's JSDoc support is function overloads. One thing I have in mind that will make it easier is a refactoring of how we handle events. I'm envisioning that each class that fires events has a dedicated event object (does not need to be a class like it is now), instead of the map having `MapEvent`, `MapBrowserEvent`, `ObjectEvent` etc.. This will also allow us to remove a lot of custom JSDoc plugin code.

simonseyock:
Yes I can definitely do that. I played around a little with typescript typing, and something like this is possible:
image

simonseyock:
But I don't think this will really help as there still is no way to iterate over multiple of these tuples and then generate overloaded functions out of that.

simonseyock:
If there is just one type for every class, the typing would get fuzzier or not? We would need to have many optional properties on the events and subsequently have a check for every property in strictly typed code.

ahocevar:
> If there is just one type for every class, the typing would get fuzzier or not? We would need to have many optional properties on the events and subsequently have a check for every property in strictly typed code.

Fuzzier for sure. Not sure about the checks for the properties though. But maybe not such a good idea anyway, as it just shifts the fuzziness to a different place.

simonseyock:
I looked a little bit into how to type this, and this form probably is not really posssible:
```
on('singleclick' | 'postcompose', (e: MapBrowserEvent) => void |  (e: RenderEvent) => void);
```
(maybe with more compicated stuff like this: https://stackoverflow.com/questions/59535995/parameters-generic-of-overloaded-function-doesnt-contain-all-options/59538756#59538756 )

Overall from a strictly typed perspective It would be nice to have distinctive methods for every event. But I am not sure if I really like that:
```
onSingleClick(listener: (e: MapBrowserEvent) => void));
onPostCompose(listener: (e: RenderEvent) => void));
```
This would also need then the corresponding `un` variants. Internally it could be implemented by calling a protected register function that works with strings like before. Or this methods could be additions.

simonseyock:
> Fuzzier for sure. Not sure about the checks for the properties though. But maybe not such a good idea anyway, as it just shifts the fuzziness to a different place.

I was thinking about the user code.
```
map.on('singleclick', (e: MapEvent) => {
  const coordinate = e?.coordinate;
  const dragging = e?.dragging;
});
```
But worse in my opinion is that the IDE cannot be sure anymore that `e` has either `coordinate` and `dragging` or `frameState` and `context` but never a mixture of those.

ahocevar:
@simonseyock Maybe it is possible to define overloads for our `on()`, `once()` and `un()` signatures. https://github.com/microsoft/TypeScript/issues/25590#issuecomment-561493756 appears to be very related.

simonseyock:
Ah that looks promising indeed!
I will look into that.
One question arises immediately is how or if this will affect the produced documentation.

ahocevar:
Just a quick try, and it appears to work correctly:
```diff
diff --git a/src/ol/PluggableMap.js b/src/ol/PluggableMap.js
index 578d924f98..939754837b 100644
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -126,13 +126,17 @@ import {removeNode} from './dom.js';
  * {@link module:ol/Map~Map#setView}.
  */
 
+/** @typedef {(type:import("./MapBrowserEventType").Types, listener:(event:MapBrowserEvent) => boolean) => import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} MapBrowserEventSignature */
+/** @typedef {(type:import("./MapEventType").Types, listener:(event:MapEvent) => boolean) => import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} MapEventSignature */
+/** @typedef {MapBrowserEventSignature & MapEventSignature} EventSignature */
+
 /**
  * @fires import("./MapBrowserEvent.js").MapBrowserEvent
  * @fires import("./MapEvent.js").MapEventl/PluggableMap.js
  * @fires import("./render/Event.js").default#precompose
  * @fires import("./render/Event.js").default#postcompose
  * @fires import("./render/Event.js").default#rendercomplete
- * @extends BaseObject<'change:layergroup'|'change:size'|'change:target'|'change:view'|'precompose'|'singleclick'|'click'|'dblclick'|'pointerdrag'|'pointermove'|'postrender'|'movestart'|'moveend'|'postcompose'|'rendercomplete'>
+ * @extends BaseObject<'change:layergroup'|'change:size'|'change:target'|'change:view'|'precompose'|'postcompose'|'rendercomplete'|import("./MapBrowserEventType").Types|import("./MapEventType").Types>
  * @api
  */
 class PluggableMap extends BaseObject {rEventType").Types, listener:(event:MapBrowserEvent) @@ -144,6 +148,9 @@ class PluggableMap extends BaseObject {("./events.js").EventsKey>} MapBro serEventSignature */
     const optionsInternal = createOptionsInternal(options);r:(event:MapEvent) => boolean) => import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} MapEventSignature */
+    /** @type {EventSignature} */ature & MapEventSignature} EventSignature */
+    this.on;
+/**
     /** @private *//MapBrowserEvent.js").MapBrowserEvent
     this.boundHandleBrowserEvent_ = this.handleBrowserEvent.bind(this);
  * @fires import("./render/Event.js").default#precompose
@@ -448,6 +455,10 @@ class PluggableMap extends BaseObject {
     );res import("./render/Event.js").default#rendercomplete
   }
 
+  /** @type {MapBrowserEventSignature} */
+  /** @type {MapEventSignature} */
+  /** @function on */
+
   /**
    * @abstract
    * @return {import("./renderer/Map.js").default} The map renderer
```
I think with this approach, we can even revert this pull request.

To make it work with the documentation, no change should be needed. Just use
```js
/***
 */
```
syntax (3 asterisks) for all added comments from the above diff, then JSDoc will ignore the comments, but TypeScript will still respect them.

simonseyock:
Nice.

I would use then some helper type like this (should work, not tested):
```
/**
 * @template Types, Event
 * @typedef {(Types, listener:(event: Event) => boolean) => import("./events.js").EventsKey|Array<import("./events.js").EventsKey>}
 */

**Comment by ahocevar** [comment]
@simonseyock Sounds good to me. Would be great if you could start a pull request. No problem is the first commit is `git revert b6c143deb294a12c49a2b76e9643e3bc37d9e169 -m1`.

**Comment by simonseyock** [comment]
I encountered a problem with this. If the base class already contains a method with the same name but broader signature, it won't create declarations for the new signatures. The autocompletion seems to work, though. But this is probably not true for a typescript project that would just import the declaration file.

This produces a `.d.ts` that has the desired typings:

[Example 1](https://www.typescriptlang.org/play?strictFunctionTypes=false&strict=true&filetype=js#code/MYGwhgzhAECyCeBRAbgUwHYBdoG8CwAUIQL6GGiQwIDCArhJgPYC2KG2qAHphgCZVI0WXGQKkiBClDjwA8pgAWqAE50GLNsK490-GZuz4J4wgHoAVJcLRz0AAI9mAB3A9cDZQEt0Ac2LQAFXgnVGtbB1RnV1RcBAN-AzD7TGDUXlQAM1wAChSQgC5A1IAaaBBPBgwVQuzUIUxCgwBKaABeAD5oACNGRhBUMHQWjuh0WmYulQT6gGVPH3QwTFplUIIbU1FycGkA1AYRdehj4EZ0D1pgJmVslqNjh4eLc2tHx-C8mJwDOYWlldQAB4AORgLrAYGlGj0JiseqdABk0B+80Wy1WIM48AAXpCZPIlKoYRp4eI3u9NkdyYoKgA6M4Ablex3EJkkZwO2Fao1QAHdAvtMLcmRIthJMPT0MLCEA)

While this doesn't:

[Example 2](https://www.typescriptlang.org/play?strictFunctionTypes=false&strict=true&filetype=js#code/MYGwhgzhAECyCeBRAbgUwHYBdoG8CwAUIQL6GGiQwIDCArhJgPYC2KG2qAHphgCZVI0WXGQKkiBClDjwA8pgAWqAE50GLNsK490-GZuz4J4wgHoAVJcLRz0AAI9mAB3A9cDZQEt0Ac2LQAFXgnVGtbB1RnV1RcBAN-AzD7TGDUXlQAM1wAChSQgC5A1IAaaBBPBgwVQuzUIUxCgwBKaABeAD5oACNGRhBUMHQWjuh0WmYulQT6gGVPH3QwTFplUIIbU1EpGAAhSBijaCPoC3NrY5t7JzBlMGZ3TC9ffzy1i8u7a9v7nFrCuPqw06PT6A3Q-nKlXQKnOx3Mm3Wx0Y6FyJTKFR0Khah3eR1Wy2U6GgABYANyw6DiEyScDSAKoBjQbR8Xb7ESIo7AZEeWjAJjKbLYikXCC0EICprkiS4o6nYUXcKvXAGOYLJYrVAAHgA5GAusBtaUaPQmKx6p0AGTQFXzRYErXazjwABehpk8iUqhNGnN4hlR3h8qOigqADpkVKLlSttzsNhWqNUAB3QIMzCCqWiLMETDhlGSwhAA)

I will think more about this later.

**Comment by simonseyock** [comment]
I tried this in a typescript project and this is true - as the typescript project will only consider the declaration but not the implementation.

A solution for this would be to rename the `on` function in `Observable` and make it protected. And then work like this:

```
constructor() {
  /** @type {Signature} */
  this.on = (type, listener) => this._on(type, listener);
}
```

But then the ApiDoc is missing.

**Comment by simonseyock** [comment]
Is there a possibility to insert apidoc for a function that is only declared in the constructor like that?

**Comment by ahocevar** [comment]
Not without a custom plugin, I'm afraid.

**Comment by simonseyock** [comment]
I am still thinking about the `onSingleClick` syntax. It would also make documenting certain events easier. I think it would be a very straight forward approach with easy type syntax while producing boiler plate code.

I will think about a custom plugin.

**Comment by ahocevar** [comment]
@simonseyock To document something like
```js
/** @type {Signature} */
this.on = (type, listener) => this._on(type, listener);
```
you could try something like
```js
/**
 * @type {Signature}
 * @param {string|Array<string>} type Type.
 * @param {ListenerFunction} listener Listener.
 * @return {EventsKey|Array<EventsKey>} Event key for `Observable.unByKey`.
 */
this.on = (type, listener) => this._on(type, listener);
```
I agree that dedicated methods like `onSingleClick()` would make everything more straightforward, but I have mixed feelings about actually doing it. It would be better if we could make it work with the function overload approach.

simonseyock:
I was thinking maybe adding some plugin that adds methods to the apidoc for annotations like this:

```js
class SomeClass {
  /**
   * @function on
   * @param {import("./MapBrowserEventType").Types|import("./MapEventType").Types} type Type.
   * @param {(e: MapBrowserEvent) => boolean|(e: MapEvent) => boolean} listener Listener.
   * @return {EventsKey|Array<EventsKey>} Event key for `Observable.unByKey`.
   */
}
```
This would a function to the apidoc without needing to have a corresponding symbol in the actual code.

Then it might even be possible to add the same function multiple times to document them differently:

```js
class SomeClass {
  /**
   * @function on
   * @param {'singleclick'|'click'} type Type.
   * @param {(e: MapBrowserEvent) => boolean} listener Listener.
   * @return {EventsKey|Array<EventsKey>} Event key for `Observable.unByKey`.
   */

  /**
   * @function on
   * @param {'movestart'|'moveend'} type Type.
   * @param {(e: MapEvent) => boolean} listener Listener.
   * @return {EventsKey|Array<EventsKey>} Event key for `Observable.unByKey`.
   */
}
```

simonseyock:
@ahocevar we would need a plugin for your approach as well to have that appear in the apidoc.

simonseyock:
> A solution for this would be to rename the on function in Observable and make it protected.

This actually does not really solve the problem as we still can have have base classes that would define an `on` function like `Interaction`. And then again the `.d.ts` files only pick up the typing of the base class.

See [Example 3](https://www.typescriptlang.org/play?strictFunctionTypes=false&filetype=js#code/MYGwhgzhAECyCeBRAbgUwHYBdoG8CwAUIQL6GGiQwIDCArhJgPYC2KG2qAHphgCZVI0WXGQKkiBClDjwA8pgAWqAE50GLNsK490-GZuz4J4wgHoAVJcLRz0AAI9mAB3A9cDZQEt0Ac2LQAFXgnVGtbB1RnV1RcBAN-AzD7TGDUXlQAM1wAChSQgC5A1IAaaBBPBgwVQuzUIUxCgwBKaABeAD5oACNGRhBUMHQWjuh0WmYulQT6gGVPH3QwTFplUIIbU1EpGAAhSBijaCPoYEZ0D1pgJmVslsPjh+sH44tzJ+ej8LyDgzmFpZWqAAPAByMBdYAg0o0ehMVj1driD7HcybdbIxQVAB0ZwA3O9oOITJJwNIAqgGNBtHxdvsROijqdzphlJdrrd6cjoBBaCEbk18RIua8Cc8vqlcL95otlqtQeDIdD4Go4QZOgAyaBS-6y4EgzjwABeUJk8iUqlhGgRSK5qNFx0xEBx6EFDyJWzOlOwrVGqAA7oEKZhboLRGGCJhnSHCEA)

simonseyock:
> I agree that dedicated methods like onSingleClick() would make everything more straightforward, but I have mixed feelings about actually doing it. It would be better if we could make it work with the function overload approach.

I don't love it either. This would be a really big change in the API. But it might be possible to deprecate `on` before removing it in a later version.

ahocevar:
> This actually does not really solve the problem as we still can have have base classes that would define an on function like Interaction. And then again the .d.ts files only pick up the typing of the base class.

I would consider this a bug in the declaration generation of TypeScript. Would you mind creating an issue on https://github.com/microsoft/typescript with a reproducible test case?

ahocevar:
> This would a function to the apidoc without needing to have a corresponding symbol in the actual code.

The `@function on` approach works already, you just have to add an `@api` annotation.

simonseyock:
> I would consider this a bug in the declaration generation of TypeScript. Would you mind creating an issue on https://github.com/microsoft/typescript with a reproducible test case?

https://github.com/microsoft/TypeScript/issues/44693

This actually started to appear somewhere around v3.9.7

simonseyock:
This actually works:

```js
/**
 * @function
 * @param {import("./MapBrowserEventType").Types} type Type.
 * @param {import("./events").ListenerFunction<import("./MapBrowserEventType").default>} listener Listener.
 * @return {EventsKey|Array<EventsKey>} Event key for `Observable.unByKey`.
 * @api
 *//**
 * @function
 * @param {import("./MapEventType").Types} type Type.
 * @param {import("./events").ListenerFunction<import("./MapEventType").default>} listener Listener.
 * @return {EventsKey|Array<EventsKey>} Event key for `Observable.unByKey`.
 * @api
 */
PluggableMap.prototype.on;
```

image

Should I create this multiple entries?

ahocevar:
Should be enough to remove the `@api` annotation from the existing one.

ahocevar:
Also note that the above syntax for overloading is only understood by JSDoc, not by TypeScript.

simonseyock:
> Also note that the above syntax for overloading is only understood by JSDoc, not by TypeScript.

Yeah, I wanted to use that fact. But if I indeed assume that the overriding in the declaration file is a bug, then we can just keep the documentation on `Observable`.

But I just want to note that this would be possible as well:

```js
    /**
     * @param {import("./MapBrowserEventType").Types} type Type.
     * @param {function(import("./MapBrowserEventType").default): (void|boolean)} listener Listener.
     * @return {import("./events").EventsKey|Array<import("./events").EventsKey>} Event key for `Observable.unByKey`.
     * @api
     *//**
     * @param {import("./MapEventType").Types} type Type.
     * @param {function(import("./MapEventType").default): (void|boolean)} listener Listener.
     * @return {import("./events").EventsKey|Array<import("./events").EventsKey>} Event key for `Observable.unByKey`.
     * @api
     *//***
     * @type {EventSignature}
     */
    this.on;
```

PR #12417, opened 2021-06-20
Use html2canvas instead of dom-to-image

ahocevar:
The `print-to-scale.html` example never worked in Safari because dom-to-image does not support Safari. By switching to html2canvas, which does not use svg to render the dom to a canvas, but pure JavaScript, we can circumvent this limitation.

While at it, I also updated jsPDF to the most recent version for both examples that use it.


MoonE (approved):
Thanks @ahocevar

jipexu:
@ahocevar  thanks i closed 11863

PR #12430, opened 2021-06-22
Event listener typing

simonseyock:
This PR adds type overloads to the `.on` functions as discussed in https://github.com/openlayers/openlayers/pull/12416

This would produce signatures like this (written in typescript for simplicity):
```js
  on(type: 'singleclick'|'doubleclick'|('singleclick'|'doubleclick')[], (e: MapBrowserEvent) => (boolean|void)): ...;
  on(type: 'movestart'|'moveend'|('movestart'|'moveend')[], (e: MapEvent) => (boolean|void)): ...;
```

The typing is already picked up by IDEs, but not yet properly propagated to the declaration file because of this bug: https://github.com/microsoft/TypeScript/issues/44693

[solved] passing arrays of event types to `on`
------------------------------------------------

It would be a typing error for now if event types that will result in different Events are grouped together:
```js
  map.on(['singleclick', 'movestart'], console.log); // -> error
```
This could maybe get circumvented by adding an additional signature like this:
```js
  on(type: ('singleclick'|'doubleclick'|'movestart'|'moveend')[], (e: Event) => (boolean|void)): ...;
```

[solved] Typing of `on` in `Observable`
------------------------------------------

The typing of `on` directly on `Observable` will not have the overloaded typing with `'change'|'error'` because it also contains the generic `on` function.

This also results in not an error if a class that directly extends `Observable` will not overload `on` with a type that handles `'change'|'error'`.


simonseyock:
@ahocevar I found a way to work around the above mentioned 'Typing of on in Observable' problem.

I don't know if it is worth it:

```js
class Observable {
  constructor () {
    /***
     * @type {OnSignature<EventTypes, import("./events/Event.js").default>}
     */
    this.on = this._on;
  }

  _on(type, listener) {
    // on implementation
  }
}

/**
 * Listen for a certain type of event.
 * @function
 * @param {string|Array<string>} type The event type or array of event types.
 * @param {function(?): ?} listener The listener function.
 * @return {import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} Unique key for the listener. If
 *     called with an array of event types as the first argument, the return
 *     will be an array of keys.
 * @api
 */
Observable.prototype.on;
```

ahocevar:
@simonseyock I ended up trying something similar before, but for me it failed as soon as I had another ineritance step, e.g. `Observable` -> `BaseObject` -> `PluggableMap`. In that case I could not get it to work for `PluggableMap`. Does this work for you?

simonseyock:
Yes, If I leave out the first line here:
https://github.com/openlayers/openlayers/pull/12430/files#diff-18abda95f23b680101919fcce9ba86e936e8fa974b668c4d05f424e4475d3ac5R140
npm run typecheck will throw an error

ahocevar:
Well if this workaround works, I'm fine with it for now. Thanks for your continued effort on this, @simonseyock.

simonseyock:
I think this PR is ready for now. The first issue 'passing arrays of event types to on' in the is not addressed right now.

ahocevar (dismissed):
Very nice! I have only one suggestion: the `Observable._on()` method should be renamed to `Observable.onInternal()` and marked `@protected`, to be consistent with other places where we do similar things.

src/ol/Observable.js
```diff
@@ -62,7 +62,7 @@ class Observable extends EventTarget {
   }
 
   // the jsdoc api annotation follows below the class declaration.
-  _on(type, listener) {
+  onInternal(type, listener) {
```
ahocevar, line 65: Maybe add a `@protected` annotation here?

ahocevar (commented):
This structure looks good. Now you'd need to add the same for `once` and `un`.

simonseyock:
I added a `CombinedOnSignature` that gets all event types as a generic parameter and enables event types of different types to be grouped together in an array. Then the event in the listener will have the type `BaseEvent`.

simonseyock:
@ahocevar I was wondering if `un` really needs this typing. `un` should never get passed a method that was not passed to `on` as well. So it should always have the correct typing or not?

If we had this typing this would prevent slip ups where the function was combined with the wrong event type for another event class. But not if it was the wrong function for the correct event type.

simonseyock:
@ahocevar another question is if the `ListenerFunction` type from `event.js` really needs the type
```
function((Event|import("./events/Event.js").default)): (void|boolean)
```
or is it in reality not rather:
```
function(import("./events/Event.js").default): (void|boolean)
```

simonseyock:
The type is used for dom elements as well. So it does not really make sense to reuse it here. But for `on` the listeners will always get a `BaseEvent` if I am not mistaken.

simonseyock:
I added the signatures for once.

simonseyock:
About my second question: `dispatchEvent` could technically dispatch a `Event`. So I will include it.

simonseyock:
Why I am a little hesitant to type `un` is that it requires another signature (it does not return am `EventsKey`) and I do not see a way to to not duplicate these complicated expressions and then for every class we need to assure that both are in sync.

ahocevar:
> @ahocevar another question is if the `ListenerFunction` type from `event.js` really needs the type
> 
> ```
> function((Event|import("./events/Event.js").default)): (void|boolean)
> ```
> 
> or is it in reality not rather:
> 
> ```
> function(import("./events/Event.js").default): (void|boolean)
> ```

On this level you are correct. But you'll need to create a new typedef. The existing one works on a lower level too (EventTarget).

ahocevar:
> Why I am a little hesitant to type `un` is that it requires another signature (it does not return am `EventsKey`) and I do not see a way to to not duplicate these complicated expressions and then for every class we need to assure that both are in sync.

Good point. A pragmatic approach would be to change `un` so it returns the `EventKey` too.

simonseyock:
Yes, or what do you think about this:

> @ahocevar I was wondering if `un` really needs this typing. `un` should never get passed a method that was not passed to `on` as well. So it should always have the correct typing or not?
> 
> If we had this typing this would prevent slip ups where the function was combined with the wrong event type for another event class. But not if it was the wrong function for the correct event type.

ahocevar:
@simonseyock That sounds good - if you can make such a typing possible.

simonseyock:
I was just thinking to leave the typing as before, because it would not provide much benefits to have the more specific typing with overloaded methods for `un`.

src/ol/Observable.js
```diff
@@ -75,15 +92,13 @@ class Observable extends EventTarget {
...
-   *     called with an array of event types as the first argument, the return
-   *     will be an array of keys.
-   * @api
+   * @param {string|Array<string>} type Type.
+   * @param {function(?): (void|boolean)} listener Listener.
+   * @return {import("./events.js").EventsKey|Array<import("./events.js").EventsKey>} Event key.
+   * @protected
    */
-  once(type, listener) {
+  // the jsdoc api annotation follows below the class declaration.
```
ahocevar, line 100: Doesn't this comment prevent TypeScript from picking up the signature and the `protected` flag?

simonseyock: No, it does seem to count as an empty line (which is also possible). As far as I know jsdoc comments always attach to the next javascript symbol (if they are not standalone).

simonseyock: I can remove that comment if wanted.

src/ol/interaction/DragAndDrop.js
```diff
@@ -82,7 +82,6 @@ export class DragAndDropEvent extends Event {
  * combinnation of formats read both text string and ArrayBuffer sources. Older browsers such
  * as IE which do not support this will need a TextDecoder polyfill to be loaded before use.
  *
- * @extends Interaction<'addfeatures'>
```
ahocevar, line 85: I think `DragAndDrop` should also be equipped with that whole event type enchilada.

ahocevar:
@simonseyock I took the liberty to push a commit that adds typed signatures to all `un()` methods, without the need to duplicate type definitions.

simonseyock:
Good idea! That the nicer way for sure.

ahocevar:
@simonseyock Do you have time to add the type signatures for `ol/interaction/DragAndDrop`? If so, I'd wait for that before merging.

simonseyock:
Done.

ahocevar:
@simonseyock Unfortunately the `Observable.prototype.on` workaround only works one inheritance level down from Observable. We do have types exported in `Object`, but not in `PluggableMap`, which inherits from `Object`.

Do you have an idea for another workaround?

ahocevar:
Acutally I don't even get `on`, `once` and `un` types in `Object.d.ts`. Do you, @simonseyock?

ahocevar:
Oh, I see now. The workaround you suggested is only for JSDoc to behave properly. For TypeScript, we're still blocked by TypeScript#44693.

simonseyock:
The fix was to have the proper typescript type on observable. But this actually works against us, because of microsoft/TypeScript#44693.

Now the resulting declaration allows only the stuff that is defined on Observable, which results in unusable declarations.

If we undo the fix we have the widest signature `(type: string, event: Event|import("./events/Event").default): Result` on Observable.

The correct typing will then only be used if working directly with the js code, not if used via the declarations. And it will not be correct for observable.

ahocevar:
So, the options we have are
* wait for microsoft/TypeScript#44693 to be fixed. Unlikely unless we do it ourselves. Would you be able to investigate which commit broke our setup? With `git bisect` on the typescript repo?
* Use an older typescript version for generating types, which might not work in other places
* Use the JSDoc types throughout, with a proper tsconfig.json recommendation. Would you be able to come up with such a configuration, which works in true typescript environments?

Any other ideas?

simonseyock:
>     * wait for [Overriden properties in jsdoc annotation are not picked up in declaration file microsoft/TypeScript#44693](https://github.com/microsoft/TypeScript/issues/44693) to be fixed. Unlikely unless we do it ourselves. Would you be able to investigate which commit broke our setup? With `git bisect` on the typescript repo?

I give this a try.
 
>     * Use an older typescript version for generating types, which might not work in other places

At the moment it seems to generate the correct declarations. We would need 2 different typescript versions.

>     * Use the JSDoc types throughout, with a proper tsconfig.json recommendation. Would you be able to come up with such a configuration, which works in true typescript environments?

I give this a try. I don't know if this is possible, but maybe `checkJs` works.

simonseyock:
`git bisect` is a little problematic as I am not able to build older typescript versions. Before v4.0.0 or so the repository did not contain a `package-lock.json` and it might be that I just have some wrong dependency version installed.

ahocevar:
@simonseyock I think for now the only viable, but temporary, solution is to generate the types with a different version. I added a commit with that change, and also updated the README a bit so users see that using third-party types won't be possible any more in the near future. Why? Because to my knowledge, there is no way to use third party types any more once `.d.ts` files are included next to the `.js` files in the package, which - I assume - is what we're aiming for.

simonseyock:
I agree that this is the only solution that will work in a forseeable future. I was under the impression that it was possible to configure alternative types via the tsconfig - instead of using `@types` in the `package.json`. Or will the definitions actually get merged? I think my ide (webstorm) did not always handle it correctly in path completions, but the tsc did.

ahocevar:
It would be great if you could come up with a way to define alternative types. I tried with `typeRoots`, but did not find a way that works. My goal would be to ship with type definitions next to the transpiled .js files, which means by default users would benefit from the types we ship with. But for those who do not agree with our types, how would they be able to continue using `@types/ol` or other types?

simonseyock:
I think I was using the `paths` in the compiler options:
```
// file: tsconfig.json

{
  "compilerOptions": {
    ...
    "baseUrl": "./",
    "paths": {
      "ol": ["node_modules/@hanreev/types-ol/ol"],
      "ol/*": ["node_modules/@hanreev/types-ol/ol/*"]
    },
    ...
  }
}
```

But I can't say for sure because I also deleted directories at some points. And I can't verify because I am on vacation :D

ahocevar:
Ok, thanks for the pointer @simonseyock. I'll play with that and see if I can come up with a solution.

ahocevar (approved):
I think this is ready for prime time. Thanks @simonseyock for your work on this"

I added another commit which adds the type declarations directly to the `ol` directory of the package, along with upgrade notes explaining how to opt out of the included types. This way, the included types will get more attention, and users do not need to set up anything to benefit from IntelliSense in VS Code, no matter whether they are authoring in TypeScript or JavaScript.

PR #12431, opened 2021-06-22
Add tsc-generated type definitions for opt-in

ahocevar:
Our type definitions and the TypeScript compiler feature to generate `*.d.ts` files from them are quite mature now, but not yet ready for prime time. Other than last time we started including types, where users were no longer able to use `@types/ol` instead, I'd like to include them in a separate directory now, so users can opt in to use them.

Opt-in is easy and explained in the README:

> When authoring in TypeScript, we recommend you try out the types that are included in the `ol` package. To use these types, make sure you have the following entry for `@types/ol` in your `package.json`'s `devDependencies` section:
> ```
> {
> ...
>   "devDependencies": {
>     ...
>     "@types/ol": "file:node_modules/ol/types",
>     ...
>   }
> }
> ```

This way, we can encourage users to start working with the built-in types, and when we feel our type definitions are complete, we can include them directly along with the modules.

I also updated the README with a paragraph about applications that want to use our untranspiled ES6 code.


README.md
```diff
@@ -59,9 +59,13 @@ OpenLayers appreciates contributions of all kinds.  We especially want to thank
 
 See our [GitHub sponsors page](https://github.com/sponsors/openlayers) or [Open Collective](https://opencollective.com/openlayers/contribute/sponsors-214/checkout) if you too are interested in becoming a regular sponsor.
 
-## IntelliSense support and type checking for VS Code
+## Building ES6
```
tschaub, line 62: Pretty petty issue, but the ECMAScript 6th edition (ES6) was never published.  Instead it was renamed to ES2015.  Rather than pinning ourselves to a specific version, I think we could use "ES Modules."

And then instead of "Building ES6," since this section is about using the untranspiled sources directly, maybe the heading could be "Using ES Modules" or just "ES Modules."

tschaub:
```suggestion
## ES Modules
```

README.md
```diff
@@ -59,9 +59,13 @@ OpenLayers appreciates contributions of all kinds.  We especially want to thank
...
 
-The ol package contains a src/ folder with JSDoc annotated sources. TypeScript can get type definitions from these sources with a [`jsconfig.json`](https://gist.github.com/ahocevar/9a7253cb4712e8bf38d75d8ac898e36c#file-jsconfig-json) (when authoring in JavaScript) or [`tsconfig.json`](https://gist.github.com/ahocevar/ad7b52a2fa0f6c5495193cd695ab3780#file-tsconfig-json) (when authoring in TypeScript) config file in the project root:
+The `ol` package contains a `src/` folder with the sources, authored in ES6. To use these untranspiled sources, either import modules from `ol/src` instead of `ol`, or configure your bundler with an alias pointing to `ol/src` for the `ol` package.
```
tschaub, line 64:
```suggestion
The `ol` package contains a `src/` folder with the sources, authored as [ES Modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules). To use these untranspiled sources, either import modules from `ol/src` instead of `ol`, or configure your bundler with an alias pointing to `ol/src` for the `ol` package.
```

tschaub (approved):
Thanks for the continued improvements on this, @ahocevar.

README.md
```diff
@@ -59,9 +59,13 @@ OpenLayers appreciates contributions of all kinds.  We especially want to thank
...
+
+## IntelliSense (VS Code) support
+
+The untranspiled sources in the `src/` are JSDoc type annotated. For applications authored in JavaScript, VS Code can get type definitions from these sources with a properly configured `jsconfig.json` in the project root:
```
simonseyock, line 68:
```suggestion
The untranspiled sources in the `src/` folder are JSDoc type annotated. For applications authored in JavaScript, VS Code can get type definitions from these sources with a properly configured `jsconfig.json` in the project root:
```

README.md
```diff
@@ -84,33 +88,22 @@ The ol package contains a src/ folder with JSDoc annotated sources. TypeScript c
...
+    "@types/ol": "file:node_modules/ol/types",
+    ...
   }
 }
 ```
-
-</details>
-
-TypeScript users may want to use a [third-party types package](https://github.com/hanreev/types-ol) instead.
+These are auto-generated with the TypeScript compiler, and will be the default in future versions. Alternatively, you can use third-party types from [Definitely Typed](https://definitelytyped.org) (`npm install @types/ol`).
```
simonseyock, line 106: Last time I checked (maybe around december last year) the `@hanreev/types-ol` were providing better typing then the `@types/ol`. Maybe mentioning both?

simonseyock:
```suggestion
These are auto-generated with the TypeScript compiler, and will be the default in future versions. Alternatively, you can use third-party types from [Definitely Typed](https://definitelytyped.org) (`npm install @types/ol`) or from [hanreev/types-ol](https://github.com/hanreev/types-ol).
```

package.json
```diff
@@ -20,11 +20,12 @@
     "start": "npm run serve-examples",
     "serve-examples": "webpack serve --config examples/webpack/config.js --mode development",
     "build-examples": "webpack --config examples/webpack/config.js --mode production",
-    "build-package": "npm run transpile && npm run copy-css && node tasks/prepare-package.js",
+    "build-package": "npm run transpile && npm run copy-css && npm run move-types && node tasks/prepare-package.js",
     "build-index": "shx rm -f build/index.js && npm run build-package && node tasks/generate-index.js",
     "build-legacy": "shx rm -rf build/legacy && npm run build-index && webpack --config config/webpack-config-legacy-build.js && cleancss --source-map src/ol/ol.css -o build/legacy/ol.css",
     "build-site": "npm run build-examples && npm run apidoc && mkdir -p build/site && cp site/index.html build/site && mv build/apidoc build/site/apidoc && mv build/examples build/site/examples",
     "copy-css": "shx cp src/ol/ol.css build/ol/ol.css",
+    "move-types": "shx rm -rf build/types && shx mkdir -p build/types && shx cp -Rf build/ol/* build/types/ && shx rm 'build/ol/**/*.d.ts*' && shx rm 'build/types/**/*.js*' && shx mv build/types build/ol/",
```
simonseyock, line 28: maybe use `--emitDeclaration=false` and `--emitDeclaration=true` with `--outDir`?

simonseyock: sorry wrong syntax. but there was a way to do this via cli.

ahocevar: Maybe. I also need to touch the declaration maps so they point to the correct source files. The way it is now, users will be taken to the type definition instead of the source when they click "Go to definition" in VS Code.

simonseyock: with these options it should be possible to create to different commands, one for typechecking the other for creating the declaration. Maybe `declaration` needs to be set to false in the `tscconfig` then:

`tsc --declaration --emitDeclarationOnly --noEmit false --outDir xyz`

simonseyock: ah well. Just turn off `declaration` and `declarationMap` again and use:
`tsc --declaration --declarationMap --outDir xyz`

package.json
```diff
@@ -20,11 +20,12 @@
     "start": "npm run serve-examples",
     "serve-examples": "webpack serve --config examples/webpack/config.js --mode development",
     "build-examples": "webpack --config examples/webpack/config.js --mode production",
-    "build-package": "npm run transpile && npm run copy-css && node tasks/prepare-package.js",
+    "build-package": "npm run transpile && npm run copy-css && npm run generate-types && node tasks/prepare-package.js",
     "build-index": "shx rm -f build/index.js && npm run build-package && node tasks/generate-index.js",
     "build-legacy": "shx rm -rf build/legacy && npm run build-index && webpack --config config/webpack-config-legacy-build.js && cleancss --source-map src/ol/ol.css -o build/legacy/ol.css",
     "build-site": "npm run build-examples && npm run apidoc && mkdir -p build/site && cp site/index.html build/site && mv build/apidoc build/site/apidoc && mv build/examples build/site/examples",
     "copy-css": "shx cp src/ol/ol.css build/ol/ol.css",
+    "generate-types": "tsc --project config/tsconfig-build.json --declaration --declarationMap --emitDeclarationOnly --outdir build/ol/types",
```
simonseyock, line 28: just found out there is a `declarationDir` option in the tsconfig

simonseyock: https://www.typescriptlang.org/tsconfig/#declarationDir

ahocevar: Doesn't appear to work. I'll leave what I have now, it works as expected.

simonseyock:
Thanks for the work on this @ahocevar! It is great that finally again openlayers can provide automatically generated declarations - not perfect, but usable. And this will probably soon reveal any major issues about the generated declarations.

simonseyock (approved):

PR #12441, opened 2021-06-26
Style-based measure example

mike-000:
The existing Measure example is an excellent example of using overlays (and should be retained) - with OpenLayers 3 there was no alternative.  It is however relatively difficult to set up and then to clear, and does not support modify.  With better styling available in later versions of OpenLayers it would be easier to implement using vector styling, modify would be supported, and additional features such as the lengths of individual segments/sides would also be easy to implement.

### DIFF
--- a/examples/measure-style.js
+++ b/examples/measure-style.js
@@ -1,0 +1,266 @@
+import Map from '../src/ol/Map.js';
+import View from '../src/ol/View.js';
+import {
+  Circle as CircleStyle,
+  Fill,
+  RegularShape,
+  Stroke,
+  Style,
+  Text,
+} from '../src/ol/style.js';
+import {Draw, Modify} from '../src/ol/interaction.js';
+import {LineString, Point} from '../src/ol/geom.js';
+import {OSM, Vector as VectorSource} from '../src/ol/source.js';
+import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
+import {getArea, getLength} from '../src/ol/sphere.js';
+
+const typeSelect = document.getElementById('type');
+const showSegments = document.getElementById('segments');
+const clearPrevious = document.getElementById('clear');
+
+const style = new Style({
+  fill: new Fill({
+    color: 'rgba(255, 255, 255, 0.2)',
+  }),
+  stroke: new Stroke({
+    color: 'rgba(0, 0, 0, 0.5)',
+    lineDash: [10, 10],
+    width: 2,
+  }),
+  image: new CircleStyle({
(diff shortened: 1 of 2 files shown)


jipexu:
i like this example ! the behavior of the dashed line  when zooming is funny it's become as animated ... 👍

marcjansen:
https://deploy-preview-12441--ol-site.netlify.app/examples/measure-style.html looks great to me.

Thanks!

marcjansen (approved):
Please merge!

examples/measure-style.js
```diff
@@ -0,0 +1,266 @@
...
+
+typeSelect.onchange = function () {
+  map.removeInteraction(draw);
+  addInteraction();
+};
+
+addInteraction();
+
+showSegments.onchange = function () {
+  vector.changed();
```
MoonE, line 265: If you are currently drawing the labels of that feature are not immediatly updated.
```suggestion
  vector.changed();
  draw.getOverlay().changed();
```

MoonE (commented):
If you complete drawing without moving the mouse afterwards the modify tooltip is at last position where the interaction was active.
This patch would prevent drawing of the tip for the next frame,
I'm not sure I like it because it is more a hack...
Maybe you have a better idea? 

```patch
diff --git a/examples/measure-style.js b/examples/measure-style.js
index e561ec7e5..898f150da 100644
--- a/examples/measure-style.js
+++ b/examples/measure-style.js
@@ -156,7 +156,17 @@ const raster = new TileLayer({
 
 const source = new VectorSource();
 
-const modify = new Modify({source: source, style: modifyStyle});
+let skip = false;
+const modify = new Modify({
+  source: source,
+  style: function () {
+    if (skip) {
+      skip = false;
+      return null;
+    }
+    return modifyStyle;
+  },
+});
 
 function styleFunction(feature, segments, drawType, tip) {
   const styles = [style];
@@ -248,6 +258,7 @@ function addInteraction() {
   });
   draw.on('drawend', function () {
     modify.setActive(true);
+    skip = true;
     tip = idleTip;
   });
   modify.setActive(true);
```

examples/measure-style.js
mike-000: The draw labels are now updating immediately, and by temporarily overriding the modify style geometry I was able to immediately display the tip in the expected position.

jipexu:
Hi
Is it a way to remove a point alone ? I can not find...

MoonE:
You can use ALT + click to remove points after you finished drawing, @jipexu

ahocevar (approved):

PR #12442, opened 2021-06-27
Fix layer render extent when it is not at all in view

### LINKED ISSUE #11651: Layer extent not working properly on Vector Layers
If an extent is set to a vector layer, the vector layer is not rendered outside of the extent as long as the extent is in the current viewport. As soon as the extent moves out of the viewport, all features are visible/rendered outside of the layers extent.

To reproduce the issue you can open the example https://openlayers.org/en/latest/examples/extent-interaction.html
and add following line at the end:
`map.getLayers().getArray()[1].setExtent([-860986.686604,4500612.225431,3287403.712489,7631472.903992]);`
That puts an extent around Europe and features should only be visible in Europe. Zooming in to Asia however will show the features.

Expected behavior is, that features outside of the layers extent should never be visible/rendered.

MoonE:
Fixes #11651 for `ol/layer/Vector` and `ol/layer/Image`.

Also removes `ol/renderer/canvas/Layer#clip` which wasn't  used anywhere.

### DIFF
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -6,7 +6,7 @@ import ViewHint from '../../ViewHint.js';
 import {ENABLE_RASTER_REPROJECTION} from '../../reproj/common.js';
 import {assign} from '../../obj.js';
 import {compose as composeTransform, makeInverse} from '../../transform.js';
-import {containsExtent, intersects} from '../../extent.js';
+import {containsExtent, intersects as intersectsExtent} from '../../extent.js';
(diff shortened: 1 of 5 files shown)


ahocevar:
Are you still working on this, @MoonE? What is missing to make it ready for review?

MoonE:
I will try to add a test, any suggestions?

ahocevar:
> I will try to add a test, any suggestions?

Good question, @MoonE. Maybe just ensure that no rendering is done when the specified extent is outside the viewport, by using a `sinon.spy` on a function involved in rendering?

ahocevar (approved):
Thanks, @MoonE!

PR #12453, opened 2021-06-29
Allow multiple LinearRings per innerBoundaryIs

### LINKED ISSUE #12407: Reading Polygon with multiple holes from KML only keeps last one
**Describe the bug**
Reading Polygon with multiple holes from KML only keeps last one 

**To Reproduce**
1. Read a KML file that contains a polygon with one outer ring and multiple inner rings into a Document
2. new KML().readFeatures(document);
3. Check the output that got read

**Expected behavior**
A polygon with as many holes as there are inner rings in the source KML file

**Actual behavior**
A polygon with one single hole (always the last one specified in the file)

MoonE:
Fixes #12407

Allows multiple LinearRing nodes per innerBoundaryIs, not allowed by spec, but Google Earth will read it too.

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -2206,7 +2206,9 @@ function lodParser(node, objectStack) {
  */
 // @ts-ignore
 const INNER_BOUNDARY_IS_PARSERS = makeStructureNS(NAMESPACE_URIS, {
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @MoonE

PR #12463, opened 2021-07-02
Improve animation examples

MoonE:
Update marker animation example
- Continually animate between start and end position
- Allow changing speed during animation
- Stay at current position when animation stops

Improve feature-animation example
- Keep features within tile bounds
- Skip drawing fully transparent style
- Don't clone the geometry each frame

Prefer Date.now() when possible instead of creating new Date objects.


ahocevar (approved):
Thanks, @MoonE!

Issue #12353, opened 2021-05-25
Make zDirection option available for all tile sources


mike-000:
**Is your feature request related to a problem? Please describe.**
At some intermediate zoom levels the text in some sources such as OSM can almost illegible as the tile images are compressed.  This could be improved by specifying `zDirection: 1` but the option is only supported when using the XYZ constructor. https://codesandbox.io/s/side-by-side-forked-kpqlc

**Describe the solution you'd like**
`zDirection` option to be available in all tile sources.  Better still support for any fractional value between -1 and +1 to give precise control over when changeover between tile z levels occurs.

mike-000:
Fixed by #12464

PR #12464, opened 2021-07-03
Make zDirection option available for all tile sources

### LINKED ISSUE #12351: Description of zDirection
**Describe the bug**
The description of the `zDirection` option in tile sources and direction option for the `getZForResolution` method of tilegrids seems to be wrong.

> If 1, the nearest lower resolution will be used. If -1, the nearest higher resolution will be used.

**To Reproduce**
https://codesandbox.io/s/simple-forked-g88d1

A direction of 1 results in the next lower z level which has a higher resolution

`input:  1 8.5 432.3931282735686  result: 8 611.49622628141`

A direction of -1 results in the next higher z level which has a lower resolution

`input: -1 8.5 432.3931282735686  result: 9 305.748113140705`

**Expected behavior**
The behaviour seems to be correct, but the description should reflect the behaviour.

mike-000:
Fixes #12351 and #12353

Update description of `zDirection` to clarify higher resolution (lower Z) and lower resolution (higher Z).
Add `zDirection` option to tile sources which do not already have it.

Also allow fractional values while keeping the existing default as 0. The midpoint between resolutions is a fractional zoom level z + 0.41503 with a zoom factor of 2, while most mapping clients change z level at z + 0.5. This can be achieved by specifying a `NearestDirectionFunction` to compare the resolution with one calculated for the zoom midpoint (or any other offset as required). This could be used where consistency with other client behaviour is desirable (e.g. raster tile sources in ol-mapbox-style) https://codesandbox.io/s/mapbox-raster-side-by-side-forked-zgsdt or where the content of tiles from a higher z viewed at the previous z + 0.41503 is not clearly legible while always enlarging lower z tiles would be excessive.

Add tests for fractional direction values.

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -60,11 +60,26 @@ export function includes(arr, obj) {
 }
 
 /**
- * @param {Array<number>} arr Array.
+ * {@link module:ol/tilegrid/TileGrid~TileGrid#getZForResolution} can use a function
+ * of this type to determine which nearest resolution to use.
+ *
+ * This function takes a `{number}` representing a value between two array entries,
+ * a `{number}` representing the value of the nearest higher entry and
+ * a `{number}` representing the value of the nearest lower entry
+ * as arguments and returns a `{number}`. If a negative number or zero is returned
+ * the lower value will be used, if a positive number is returned the higher value
+ * will be used.
+ * @typedef {function(number, number, number): number} NearestDirectionFunction
+ * @api
+ */
+
+/**
+ * @param {Array<number>} arr Array in desccending order.
  * @param {number} target Target.
- * @param {number} direction 0 means return the nearest, > 0
- *    means return the largest nearest, < 0 means return the
- *    smallest nearest.
+ * @param {number|NearestDirectionFunction} direction
+ *    0 means return the nearest,
+ *    > 0 means return the largest nearest,
+ *    < 0 means return the smallest nearest.
  * @return {number} Index.
  */
 export function linearFindNearest(arr, target, direction) {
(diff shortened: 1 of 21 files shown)


ahocevar (commented):
I really like this new feature, but "Fractional values give intermediate behavior." should be expanded to a sentence that reflects the gist of the pull request description, i.e. the formula and the recommended value to achieve a tile zoom level change at z + 0.5.

ahocevar:
Looking at this closer, I find the calculation of the desired `zDirection` value quite unintuitive, due to the range of `-1` to `+1`. What if `linearFindNearest` would also accept a function `zDirection`:
```js
zDirection: (value, high, low) => value > low * Math.SQRT2 ? high : low
```
What do you think, @mike-000?

Also, it would maybe be better to document how this works in `TileGrid.getZForResolution`, and link to that from the sources where `zDirection` can be configured.

mike-000:
Yes, I think a function would be useful as it could better handle non-standard tile grids with resolutions derived from paper maps such as 10, 25, 50, 100, etc. where the midpoint of the zooms would be slightly different at steps with zoom factor 2.5

```
const factor = (high - low) / low;
const midzoom = low * Math.sqrt(factor);
return value > midzoom ? high : low;
```

and in a situation like that much more understandable.

ahocevar (commented):
I think it would be more intuitive if the function returned the same values/types as when provided a constant value, i.e. values greater than `0` for `high`, smaller than `0` for `low`, and maybe even `0` for the closest:
```js
function(value, high, low) {
  return value - low * Math.sqrt(high / low);
}
```

Regarding documentation, I think for the sources it could be shortened to something like
```js
/**
 * @property {number|import("../array.js").NearestDirectionFunction} [zDirection=0]
 * Choose whether to use tiles with a higher or lower zoom level when in between integer
 * zoom levels. See {@link module:ol/tilegrid/TileGrid~TileGrid#getZForResolution}.
 */
```

ahocevar (approved):
Thanks for your effort on this, @mike-000

PR #12466, opened 2021-07-04
Fix getVectorContext on TileLayer with pixelRatio != 1

MoonE:
If the transform matrix on the canvas does not match the frameState.pixelRatio it draws at the wrong size.


ahocevar (approved):
Thanks, @MoonE

PR #12480, opened 2021-07-06
Fix gap in tiles due to floating point math

MoonE:
Due to floating point inaccuracies gaps between tiles may occur:
https://codesandbox.io/s/floating-point-tile-gap-vomuj?file=/main.js
In the example above there is a vertical gap at the center of the map.
In one iteration nextX `840.4999999999999` is rounded to `840`, and in the next  iteration `x` `840.5` is rounded to `841`.
This can be fixed by reordering the calculation.


ahocevar (approved):
Very nice, thanks @MoonE!

Issue #12486, opened 2021-07-08
Can't run any test


Razi91:
**Describe the bug**
Can't run locally any `test-browser`

**To Reproduce**
Clone repository, `npm install`, then `npm test`. Tested on node 10 and 14.

**Expected behavior**
Tests should work. It seems there are some problems with `ol-mapbox-style` package.

OS: KDE Neon (Linux 5.8)
Node versions: 10 and 14.

logs: https://pastebin.com/gVum8VNL

ahocevar:
With node 10, our build environment does not work. With node 16, everything should work fine. Anyway, I updated `ol-mapbox-style` to also use `"type": "module"` in the package, and that makes it work with node 14 too. So #12489 fixes this issue.

Razi91:
After those merges works fine, thanks

PR #12487, opened 2021-07-08
Custom cluster feature creation function

Razi91:
Adds `hydrate` function option into Cluster constructor. It allows to for example fill cluster's feature with attributes, set custom style etc.

I'm not sure if `hydrate` is right name here. It's called after Feature is created, before `features` attribute is set.

I'm not able to run tests locally (due to https://github.com/openlayers/openlayers/issues/12486)

example usage:
```ts
      const source = new Cluster({
        source: new VectorSource({
          features: [feature1, feature2],
        }),
        hydrate: (clusterFeature, features) => {
          clusterFeature.set('sum', features.reduce((acc, ft) => acc + ft.get('value'), 0);
          clusterFeature.set('hasAllValidItems', features.every((ft) => acc + ft.get('isValid'));
        },
      });
```

It might be useful when you receive events elsewhere and you want to have for example the same attribute in every feature on map, whether it's clustered or not

// EDIT
Changed to  `createClusterFeature: (clusterPoint: Point, featuresInCluster: Feature[]): Feature`, it will be more elastic. Now inserting features is optional and can be done under different attribute name. Also we can use different Feature implementation (if anyone needs it), reuse etc.

### DIFF
--- a/src/ol/source/Cluster.js
+++ b/src/ol/source/Cluster.js
@@ -38,10 +38,20 @@ import {getUid} from '../util.js';
  *   return feature.getGeometry();
  * }
  * ```
- * @property {function(Feature, Array<Feature>):any} [hydrate] Function called after creating
- * cluster feature
  * See {@link module:ol/geom/Polygon~Polygon#getInteriorPoint} for a way to get a cluster
  * calculation point for polygons.
+ * @property {function(Point, Array<Feature>):Feature} [createClusterFeature]
+ * Function that takes cluster's ceter {@link module:ol/geom/Point} and array
+ * of {@link module:ol/Feature} included in this cluster, must return
+ * {@link module:ol/Feature} that will be used to render. Default implementation is:
+ * ```js
+ * function(point, features) {
+ *   return new Feature({
+ *     geometry: point,
+ *     features: features
+ *   });
+ * }
+ * ```
  * @property {VectorSource} [source] Source.
  * @property {boolean} [wrapX=true] Whether to wrap the world horizontally.
  */
@@ -111,10 +121,10 @@ class Cluster extends VectorSource {
       };
 
     /**
-     * @type {function(Feature, Array<Feature>):any}
+     * @type {function(Point, Array<Feature>):Feature}
      * @protected
      */
-    this.hydrate = options.hydrate;
+    this.createClusterFeature = options.createClusterFeature;
 
     /**
      * @type {VectorSource}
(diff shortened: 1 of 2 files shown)


mike-000:
`aggregate` might be a better name as it is intended for property value aggregation

ahocevar:
What about allowing the user complete control over creation of the cluster feature, with an createClusterFeature option? The current behavior (default) would be
```js
  createClusterFeature: (clusterPoint, featuresInCluster) => {
    return new Feature({
      geometry: clusterPoint,
      features: featuresInCluster
    });
  }
```

Razi91:
Sounds good @ahocevar , AFAIR `'features'` attribute is not used anywhere in OpenLayers itself, it's just for application developers. That way we'll have full control over how this Feature will look like.

Razi91:
Changed and rebased to new version. Should I squash those commits?

ahocevar (commented):
Please add better documentation. It would be a good idea to add the default code as snippet to the docs, so users get a clear picture of what this function could look like.

Razi91:
Added better documentation, I was considering also adding an example or modify one, but I don't have any idea right now about it :(

src/ol/source/Cluster.js
```diff
@@ -40,6 +40,18 @@ import {getUid} from '../util.js';
  * ```
  * See {@link module:ol/geom/Polygon~Polygon#getInteriorPoint} for a way to get a cluster
  * calculation point for polygons.
+ * @property {function(Point, Array<Feature>):Feature} [createClusterFeature]
+ * Function that takes cluster's ceter {@link module:ol/geom/Point} and array
+ * of {@link module:ol/Feature} included in this cluster, must return
```
ahocevar, line 45:
```suggestion
 * Function that takes the cluster's center {@link module:ol/geom/Point} and an array
 * of {@link module:ol/Feature} included in this cluster. Must return a
```

src/ol/source/Cluster.js
```diff
@@ -180,6 +198,28 @@ class Cluster extends VectorSource {
     this.updateDistance(this.distance, minDistance);
   }
 
+  /**
+   * Current function used to create cluster features
+   * @return {function(Point, Array<Feature>):Feature | undefined} cluster creation function
+   * @api
+   */
+  getCreateClusterFeature() {
+    return this.createClusterFeature;
```
ahocevar, line 207: If you provide this getter, it should also return the default if no custom function is configured.

Razi91: I'm not sure if it should return default. This way we can check if any is provided, or default is used. If we return the default, we'll need to check if it's the one we want or not (or if it's default)

ahocevar: If you export the default function, you can easily check if it's the default:
```js
if (cluster.getCreateClusterFeature() === defaultCreateClusterFeature)
```
But I'd be inclined to not have a `getCreateClusterFeature()` getter at all, and maybe not even a setter. What would be a use case for needing those? It's cheap to create a new cluster source.

Razi91:
Ok, removed getter/setter, fixed docs.

ahocevar (approved):
Thanks @Razi91, this looks great now!

PR #12488, opened 2021-07-08
MousePosition: allow rendering of empty string

### LINKED ISSUE #12482: MousePosition Control: Allow setting empty string if coordiante is undefined
MousePosition Control:
- [Docs v4.6.5](https://openlayers.org/en/v4.6.5/apidoc/ol.control.MousePosition.html)
- [Docs v6.5.0](https://openlayers.org/en/v6.5.0/apidoc/module-ol_control_MousePosition-MousePosition.html)

I upgraded an application from OL4 to OL6 and discovered the following: In OL4 the default value for an undefined coordinate was an empty string. in OL6 the default value is a whitespace. This is a problem for me, because my application relies on having an empty string when the coordinate is undefined. 

--> the problem is that in OL6 it does not seem to be possible to set an empty string via `undefinedHTML`, because it recognizes an empty string as "falsy" value. This causes the return of the last rendered position, which is not always desired.

An option to make it possible to get an empty string in case the coordinate is undefined.

JakobMiksch:
fixes #12482

 **ol.control.MousePosition** 
This PR makes it possible to enter an empty string `''` as `undefinedHTML` value. This will finally add an empty string into the target HTML element. The behavior before was that an empty string was recognized as falsey and therefore the default value - a whitespace - was used in the target HTML element.

### DIFF
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -35,8 +35,8 @@ const COORDINATE_FORMAT = 'coordinateFormat';
  * @property {string} [undefinedHTML='&#160;'] Markup to show when coordinates are not
  * available (e.g. when the pointer leaves the map viewport).  By default, the last position
  * will be replaced with `'&#160;'` (`&nbsp;`) when the pointer leaves the viewport.  To
- * retain the last rendered position, set this option to something falsey (like an empty
- * string `''`).
+ * retain the last rendered position, set this option to something falsey. An exception is an
+ * empty string `''`. It does not count as falsey in this case.
  */
 
 /**
@@ -89,7 +89,8 @@ class MousePosition extends Control {
      * @private
      * @type {boolean}
      */
-    this.renderOnMouseOut_ = !!this.undefinedHTML_;
+    this.renderOnMouseOut_ =
+      this.undefinedHTML_ === '' || !!this.undefinedHTML_;
 
     /**
      * @private
--- a/test/browser/spec/ol/control/mouseposition.test.js
+++ b/test/browser/spec/ol/control/mouseposition.test.js
@@ -105,7 +105,30 @@ describe('ol/control/MousePosition', function () {
         expect(element.innerHTML).to.be('&nbsp;');
       });
 
-      it('retains the mouse position when undefinedHTML is falsey and mouse moves outside the viewport', function () {
+      it('retains the mouse position when undefinedHTML is false and mouse moves outside the viewport', function () {
+        const ctrl = new MousePosition({
+          undefinedHTML: false,
+        });
+        ctrl.setMap(map);
+        map.renderSync();
+
+        const element = document.querySelector(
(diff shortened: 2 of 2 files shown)


tschaub:
The tests make it look like the current behavior is intentional: https://github.com/openlayers/openlayers/blob/bfe5afe0698d2c48e19807e58716acc5b7f0eefc/test/browser/spec/ol/control/mouseposition.test.js#L108

MoonE:
See https://github.com/openlayers/openlayers/pull/8109 for the PR introducing the current behavior.

tschaub:
An alternative would be to render an empty element instead of a non-breaking space by default.  This would not break applications that are relying on the current documented/tested behavior with respect to `undefinedHTML: ''`.  But it would potentially break styling if an application is relying on the element always having text content (which I think is why the non-breaking space was added).

tschaub:
Thanks for the reminder, @MoonE .  After reading through #8109 and #2914, I think this pull request makes sense (maybe a breaking change, but maybe a fix).

If you want to clear the control on mouseout, use the default options (same as `undefinedHTML: '&nbsp;'`).

If you want to retain the position on mouseout, pass `undefinedHTML: undefined`.

If you want to clear the position by rendering an empty element instead of a non-breaking space, pass `undefinedHTML: ''`.

Kind of subtle and awkward (but I think that is mostly due to the `undefinedHTML` name).

tschaub:
@JakobMiksch can you update the tests to cover the new behavior (change existing test for empty string and add test for `undefined`)?

marcjansen:
While we are at it, maybe also change the name from `undefinedHTML` to sth. less confusing?

tschaub:
> While we are at it, maybe also change the name from `undefinedHTML` to sth. less confusing?

I had `placeholder` in mind (inspired by [`<input placeholder="foo">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefplaceholder)).  But we might want to wait for a major release for that.  Or we could add `placeholder`, deprecated `undefinedHTML`, and then later remove it.

tschaub:
> @JakobMiksch can you update the tests to cover the new behavior (change existing test for empty string and add test for `undefined`)?

I pushed new tests in 9e7f5a9b40480d7aa60f225a20116191b5b3c140.

tschaub (approved):

tschaub:
Thanks for the contribution, @JakobMiksch

tschaub:
> While we are at it, maybe also change the name from `undefinedHTML` to sth. less confusing?

@marcjansen - see #12491.  The change there restores the previous (awkward) behavior of `undefinedHTML` and adds a new `placeholder` option.  This `placeholder` option can be set to the empty string to render the empty string when the mouse position is not available.

JakobMiksch:
@tschaub thanks for finalizing this PR. I would have tried soon, but you were faster 👍

PR #12491, opened 2021-07-09
Add a new placeholder option and deprecate undefinedHTML for the mouse position control

tschaub:
This builds on the changes in #12488.

The new `placeholder` option replaces the deprecated `undefinedHTML` option.

I haven't changed the behavior of the `undefinedHTML` option – this branch restores its behavior to how it was before #12488.

The new `placeholder` option can be set to an empty string (`''`) to render an empty string when the mouse position is not available (fixing #12482 without breaking existing behavior).  The same `placeholder` option can be set to `false` to retain the previous mouse position when the mouse leaves the viewport.

In a future release, we can remove `undefinedHTML` and we can change the default behavior of the control to render the previous position when the mouse leaves the viewport by default.  This will make it so we can change the type of the `placeholder` option to be `string` instead of `string|boolean`.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -14,6 +14,17 @@ If desired, e.g. when you don't want to adjust your code after upgrading from a 
       "ol/*": ["node_modules/@types/ol/*"]
     },
 ```
+#### Deprecation of `undefinedHTML` option for the MousePosition control
+
+The `undefinedHTML` option for the MousePosition control has been deprecated and will be removed in a future release.  Use the new `placeholder` option instead.
+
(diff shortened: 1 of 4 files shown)


ahocevar (approved):

marcjansen (approved):
Looks great. Since the options are mutually exclusive, we might want to add a note, that people should not set both at a time. You decide whether this would help.

Please merge with or without adding the note.

tschaub:
> Since the options are mutually exclusive, we might want to add a note, that people should not set both at a time.

The latest commit includes this detail in the upgrade notes.

PR #12498, opened 2021-07-12
CanvasLayerRenderer: Forward LayerType type parameter to LayerRenderer

bryantevans00:
This change fixes a minor issue with the generated TypeScript definitions, where `CanvasLayerRenderer` does not currently forward its `LayerType` type parameter to the `LayerRenderer` superclass. This results in the following declaration:

```typescript
class CanvasLayerRenderer<LayerType extends Layer = Layer> extends LayerRenderer<any> { ... }
```
which results in type errors when `CanvasLayerRenderer<T>` is extended in application code. By forwarding the type parameter correctly, the declaration becomes

```typescript
class CanvasLayerRenderer<LayerType extends Layer = Layer> extends LayerRenderer<LayerType> { ... }
```
which fixes the type errors.

### DIFF
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -21,6 +21,7 @@ import {createCanvasContext2D} from '../../dom.js';
 /**
  * @abstract
  * @template {import("../../layer/Layer.js").default} LayerType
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @bryantevans00!

PR #12506, opened 2021-07-12
Fix adding controls with map config

M393:
If a control like [ol.control.Bar](https://viglino.github.io/ol-ext/doc/doc-pages/ol.control.Bar.html) by @Viglino adds sub controls when added to a map it fails during map creation because the listeners are added after the control was added to the map.

Adding listeners before adding controls / interactions / overlays fixes this

See https://codesandbox.io/s/ol-ext-bar-map-config-vl1o0?file=/src/index.js for an example

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -383,16 +383,6 @@ class PluggableMap extends BaseObject {
     // is "defined" already.
     this.setProperties(optionsInternal.values);
 
-    this.controls.forEach(
-      /**
-       * @param {import("./control/Control.js").default} control Control.
-       * @this {PluggableMap}
-       */
-      function (control) {
-        control.setMap(this);
-      }.bind(this)
-    );
-
     this.controls.addEventListener(
       CollectionEventType.ADD,
       /**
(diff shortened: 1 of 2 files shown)


ahocevar (commented):
I'm not sure we want this. Although it solves the issue you are describing, it also adds overhead to the initial creation of a map with interactions and controls.

Couldn't this be solved in [ol.control.Bar](https://viglino.github.io/ol-ext/doc/doc-pages/ol.control.Bar.html) instead?

M393:
I don't think it adds any overhead.
The listeners are only dispatched for controls that are newly added when the map is set on the already configured controls.
If the collections don't change, it behaves the same.

ahocevar:
Ok, then I misread your change. Can you add a test that failed before and passes after this change? That should make clear what you were doing.

M393:
I added the test for the control config, I can add one for overlays and interactions if you like, @ahocevar.
I hope this clears up the intent of this change.

ahocevar (approved):
Ok, thanks for the test. This change makes sense and does not add overhead through additional collection events.

PR #12513, opened 2021-07-13
Draw a sketch point when adding a vertex programmatically

### LINKED ISSUE #12118: Can't call removeLastPoint after using appendCoordinates to reopen a polygon
**Describe the bug**
After calling appendCoordinates on DrawInteraction, if you then try to call removeLastPoint on the newly opened geometry, openlayers errors to the console.

**To Reproduce**
1. Go to https://codesandbox.io/s/agitated-snyder-glhwb?file=/main.js
2. Draw a polygon
3. Press any key to reopen that polygon
4. Press any key again once it is opened
5. Get the error: `Cannot read property 'setGeometry' of null`

**Expected behavior**
The last point is removed the same as if you were calling removeLastPoint whilst initially drawing the polygon

ahocevar:
Fixes #12118.

By calling `createOrUpdateSketchPoint_` instead of setting the sketch point's geometry directly, we no longer fail when there is no sketch point, which won't happen anymore anyway, because we now add a sketch point when adding coordinates with `appendCoordinates()`.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -708,7 +708,7 @@ class Draw extends PointerInteraction {
     if (this.finishCoordinate_) {
       this.modifyDrawing_(event.coordinate);
     } else {
-      this.createOrUpdateSketchPoint_(event);
+      this.createOrUpdateSketchPoint_(event.coordinate.slice());
     }
   }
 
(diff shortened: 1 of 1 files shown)


MoonE (approved):
Thanks, @ahocevar

PR #12518, opened 2021-07-14
VectorEventType: Add 'changefeature' in @typedef

DaniEll-AT:
All other events (addfeature, clear, removefeature, featuresloadstart, featuresloadend, featuresloaderror) are there, but 'changefeature' is missing.


ahocevar (approved):
Good catch. Thanks for the fix, @DaniEll-AT.

PR #12527, opened 2021-07-16
Treat custom loaders without success/fail handling as if they were a void loader

### LINKED ISSUE #12515: Map rendercomplete event is not fired when Graticule is added in v6.6.0
**Describe the bug**
Map `rendercomplete` event is not fired when Graticule is added. This only occurs in v6.6.0.

**To Reproduce**
1. Go to https://stackblitz.com/edit/ol-660?devtoolsheight=33&file=index.ts
2. See console at bottom right

**Expected behavior**
`rendercomplete` event should be fired as in v6.5.0. Here the same code using v6.5.0: https://stackblitz.com/edit/ol-650?devtoolsheight=33&file=index.ts

mike-000:
Fixes #12515

#11941 broke subsequent events when custom loaders such as `ol/layer/Graticule` do not handle success/fail.  Treating such loaders in the same way as a void loader fixes it.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -983,7 +983,7 @@ class VectorSource extends Source {
       }
     }
     this.loading =
-      this.loader_ === VOID ? false : this.loadingExtentsCount_ > 0;
+      this.loader_.length < 4 ? false : this.loadingExtentsCount_ > 0;
   }
 
   refresh() {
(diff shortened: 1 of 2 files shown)


ahocevar (commented):
Thanks for this clever fix, @mike-000. Would you be able to add a test to avoid future regressions on this? Would be much appreciated.

ahocevar (approved):
Thanks, @mike-000

PR #12528, opened 2021-07-17
Convert the hit tolerance priority example into a test

MoonE:
See https://github.com/openlayers/openlayers/pull/11783#commitcomment-45529702


tschaub (approved):
Thanks for following up on this, @MoonE

PR #12549, opened 2021-07-23
Add `change:layers` event.

simonseyock:
This adds a missing event to the `on` signature.


ahocevar (approved):
Thanks, @simonseyock!

PR #12565, opened 2021-07-27
Handle named colors as string in equal operator

### LINKED ISSUE #12564: Webgl style parsing incorrectly parses string as color in conditional operator
**Describe the bug**
If you use strings that are colors (e.g. `red`) in conditional operators, the parsing of the expression fails.

**To Reproduce**
Have a webgl point layer with a style that includes a conditional like:
```js
['==', ['get', 'someProperty'], 'red'],
```

You get an error like: `Unexpected expression 'red' (expected type 6)`. Even though 'red' is exactly type 6 (meaning string or color)
The same would be true for number arrays of length 3 or 4.
The issue is that the `expressionToGlsl` function in `src/ol/style/expressions.js` doesn't deal with ambiguous typeHints. 

**Expected behavior**

The `expressionToGlsl` could either make a deterministic choice when this happens. E.g. prefer strings over colors if something can be both a string or a color.
Alternatively the syntax of the conditionals could be expanded to haven an optional 4th operator where the user can force a certain value type.

sebakerckhof:
Fixes #12564

Alternatively, I could change the equal operator to take an additional argument to overrule the type, but afaik it never makes sense to have color values here. So I assume we can always treat named strings as strings here.

### DIFF
--- a/test/browser/spec/ol/style/expressions.test.js
+++ b/test/browser/spec/ol/style/expressions.test.js
@@ -250,6 +250,9 @@ describe('ol.style.expressions', function () {
       expect(expressionToGlsl(context, ['==', 10, ['get', 'attr4']])).to.eql(
         '(10.0 == a_attr4)'
       );
+      expect(expressionToGlsl(context, ['==', 'red', ['get', 'attr4']])).to.eql(
+        '(0.0 == a_attr4)'
+      );
       expect(expressionToGlsl(context, ['!=', 10, ['get', 'attr4']])).to.eql(
         '(10.0 != a_attr4)'
       );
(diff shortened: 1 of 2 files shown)


test/browser/spec/ol/style/expressions.test.js
```diff
@@ -250,6 +250,9 @@ describe('ol.style.expressions', function () {
       expect(expressionToGlsl(context, ['==', 10, ['get', 'attr4']])).to.eql(
         '(10.0 == a_attr4)'
       );
+      expect(expressionToGlsl(context, ['==', 'red', ['get', 'attr4']])).to.eql(
+        '(0.0 == a_attr4)'
```
ahocevar, line 254: To be safe, maybe also check for correct string mapping:
```suggestion
        `(${stringToGlsl(context, 'red')} == a_attr4)`
```

ahocevar (commented):
Thanks for your work on this, @sebakerckhof. I added two suggestions you might want to consider.

ahocevar (approved):

ahocevar:
Thanks, @sebakerckhof

PR #12577, opened 2021-07-31
Use shx for the build-site script, clean before run

MoonE:
The build-site script does not use shx to execute its copy / move / mkdir commands and thus cannot be run in e. g. windows cmd.
The script also fails to copy the examples / apidoc when it is run a second time because folders cannot be move when there they already exist in the destination.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "build-package": "npm run transpile && npm run copy-css && npm run generate-types && node tasks/prepare-package.js",
     "build-index": "shx rm -f build/index.js && npm run build-package && node tasks/generate-index.js",
     "build-legacy": "shx rm -rf build/legacy && npm run build-index && webpack --config config/webpack-config-legacy-build.mjs && cleancss --source-map src/ol/ol.css -o build/legacy/ol.css",
-    "build-site": "npm run build-examples && npm run apidoc && mkdir -p build/site && cp site/index.html build/site && mv build/apidoc build/site/apidoc && mv build/examples build/site/examples",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,12 +23,12 @@
...
+    "build-site": "shx rm -rf build/site && npm run build-examples && npm run apidoc && shx mkdir -p build/site && shx cp site/index.html build/site/ && shx mv build/apidoc build/examples build/site/",
     "copy-css": "shx cp src/ol/ol.css build/ol/ol.css",
     "generate-types": "npx --package=typescript@3.8.3 -y -- tsc --project config/tsconfig-build.json --declaration --declarationMap --emitDeclarationOnly --outdir build/ol",
     "transpile": "shx rm -rf build/ol && shx mkdir -p build/ol && shx cp -rf src/ol build/ol/src && node tasks/serialize-workers.cjs && npx --package=typescript@4.3.5 -y -- tsc --project config/tsconfig-build.json",
     "typecheck": "npx --package=typescript@4.3.5 -y -- tsc --pretty",
-    "apidoc-debug": "shx rm -rf build/apidoc && node --inspect-brk=9229 ./node_modules/jsdoc/jsdoc.js -R config/jsdoc/api/index.md -c config/jsdoc/api/conf.json -P package.json -d build/apidoc",
+    "apidoc-debug": "shx rm -rf build/apidoc && node --inspect-brk=9229 ./node_modules/.bin/jsdoc -R config/jsdoc/api/index.md -c config/jsdoc/api/conf.json -P package.json -d build/apidoc",
```
ahocevar, line 31: Is this change necessary? I seem to remember that there were problems with dot-directories in Windows under some circumstances.

MoonE: It's not necessary.
As I understand it the .bin directory is where links to all the executable files are added to make it easier to run them.
Thought it make sense to use it, but I wasn't aware of potential problems.

ahocevar (approved):

PR #12578, opened 2021-08-01
Update rendering test reference image

ahocevar:
Looks like truetype font rendering has improved in our rendering test setup. This is the reason why one of the rendering tests started failing.

This pull request only updates the reference image to make the tests pass again.


ahocevar:
Looks like the previous truetype rendering is restored in our build environment. Closing.

PR #12597, opened 2021-08-04
#12596 fix RasterSource does not end Tile transition

### LINKED ISSUE #12596: RasterSource does not end Tile transition
When using a TileSource like in the "Color Manipulation" example inside a RasterSource/ImageLayer the transition state of the loaded Tiles will never end (in contrast to directly using a TileLayer). This causes multiple effects:
- rendering loop although nothing changes
- wrong tile is rendered

**To Reproduce**
1. Go to 'https://openlayers.org/en/latest/examples/color-manipulation.html'
2. Zoom into the map (wait to load the tiles with a higher zoom level)
3. Zoom out of the map
4. You will see some tiles rendered with the wrong zoom level (one zoom level higher because of transition rendering mode) like in this screenshot:
grafik

**Expected behavior**
RasterSource should correctly end Tiles transition and FrameState animation mode.

**My analysis**
I debugged the code but unfortunately could not find a working solution.
The issue seems to be connected with impl. of RasterSource.updateFrameState_ where `frameState.time = Infinity` is used.
This will cause Tile.getAlpha never to return 1 in CanvasTileLayerRenderer.renderFrame leading to rendering childTileRange (wrong tiles over the already loaded correct tiles) and to never stop transition. 
This will also lead to RasterSource.getImage to call requestAnimationFrame (after zooming/loading additional tiles) forever (consuming cpu resources).

mwerlitz:
fixes #12596

### DIFF
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -783,7 +783,6 @@ class RasterSource extends ImageSource {
     this.dispatchEvent(
       new RasterSourceEvent(RasterEventType.AFTEROPERATIONS, frameState, data)
     );
-    
     requestAnimationFrame(this.changed.bind(this));
   }
 
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -639,7 +639,7 @@ class RasterSource extends ImageSource {
     frameState.extent = extent.slice();
     frameState.size[0] = Math.round(getWidth(extent) / resolution);
     frameState.size[1] = Math.round(getHeight(extent) / resolution);
-    frameState.time = Infinity;
+    frameState.time = Date.now();
 
     const viewState = frameState.viewState;
     viewState.center = center;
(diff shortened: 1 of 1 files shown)


mwerlitz:
Maybe it's also possible to replace line RasterSource:780 `this.changed();` with `requestAnimationFrame(this.changed.bind(this));`

src/ol/source/Raster.js
```diff
@@ -783,6 +783,8 @@ class RasterSource extends ImageSource {
     this.dispatchEvent(
       new RasterSourceEvent(RasterEventType.AFTEROPERATIONS, frameState, data)
     );
+
```
ahocevar, line 786:
```suggestion

```

ahocevar (commented):
This fix makes sense. Thanks!

I added one suggestion for a whitespace fix to make the CI happy. Please commit that suggestion, then we can merge this.

mwerlitz:
Ops, sorry for the whitespace problem. IDE setup was not complete yet.

ahocevar (approved):

ahocevar:
Thanks, @mwerlitz!

PR #12600, opened 2021-08-05
Make attribution getters public.

simonseyock:
In some applications it is nice to have a custom attribution control that for example can merge attributions for multiple layers. With the current API this is quite cumbersome.

This PR makes the `getAttributions` and `getAttributionsCollapsible` methods of `ol/source/Source` public to enable a custom attribution handling.

### DIFF
--- a/src/ol/source/Source.js
+++ b/src/ol/source/Source.js
@@ -95,6 +95,7 @@ class Source extends BaseObject {
   /**
    * Get the attribution function for the source.
    * @return {?Attribution} Attribution function.
+   * @api
    */
   getAttributions() {
     return this.attributions_;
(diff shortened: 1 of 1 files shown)


tschaub:
This sounds ok to me, but I’m wondering if there is something that could be done to make the attribution control more flexible.  @simonseyock could the logic for your custom control be handled be a more flexible control in the library?

simonseyock:
The control would need to have an aggregation function that ideally not only gets the attribution but also the layer passed.

Something like:
```
attributionReducer: (attributions: string[], attribution: string, layer: Layer) => string[]
```

simonseyock:
I can add this if wanted, but I still think that the `getAttribution` function should be public.

tschaub (approved):

PR #12605, opened 2021-08-06
Some typing improvements

simonseyock:
This PR includes:
* `Feature.clone` returns now `Feature<Geometry>` instead of `Feature<any>` (in this case `Geometry` is the generic parameter)
* `FeatureLike` uses `Feature<Geometry>` instead of `Feature<any>`
* `Layer` is now always used as `Layer<Source>` instead of `Layer<any>` (hopefully i did catch everything)

This improves typing as `layer.getSource().someRandomName` now is an error which is not the case if layer has type `Layer<any>`.

* The Feature Options are now typed.

This enables the feature to be properly typed if an object gets passed. i.e.:
```
const feature = new Feature({ geometry: new Point() });
```
will now have type `Feature<Point>` instead of `Feature<any>`. In the apidoc the type of the options object is now missing the `Object<string, *>` part, but it is described in the text:

image

### DIFF
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -22,6 +22,11 @@ import {listen, unlistenByKey} from './events.js';
  *     |'change:geometry', Return>} FeatureOnSignature
  */
 
+/***
+ * @template Geometry
+ * @typedef {Object<string, *> & { geometry?: Geometry }} ObjectWithGeometry
+ */
+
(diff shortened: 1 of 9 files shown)


tschaub (approved):

tschaub:
Thanks, @simonseyock.

PR #12607, opened 2021-08-08
Don't add color as possible type if it was not set

MoonE:
Using xor will remove the color type only if it was set, if not it will be added which is wrong.
See #12565


ahocevar (approved):
Thanks, @MoonE.

Issue #12566, opened 2021-07-28
Layer disappearing when coming into vector layer with opacity set


EngineerCoding:
**Describe the bug**
I have two layers setup, one which is the OSM tilelayer and one is a VectorLayer with a plain VectorSource. The former layer has zIndex of 0 and the vectorlayer a zIndex of 500. To the vector layer I am dynamically adding features, for now just points with an icon as style. 

When I change the opacity to any value between 0 and 1 (but not 0 and 1 themselves), the OSM layer (and other layers which I can configure in my app, but probably not relevant) disappears to the background color when the point in the vectorlayer is in the rendered extent. When I move out of the rendered extent, the OSM layers shows up again.

I have no clue why this is happening nor am I unable to reproduce this issue in a trimmed down sample application which shows this issue in action. Any ideas on why this might be happening would be great, as this seems like a proper bug.

**To Reproduce**
Unfortunately no sample app available, as I am unable to reproduce it in a isolated app. The issue is consistent though.

**Expected behavior**
The opacity be set without the underlying OSM layer disappearing (the background color seems to be showing up)

ahocevar:
Without a sample app this will be hard to investigate. But one thing you should double-check is that you are using the correct type and value for the opacity, i.e. no string and correct comma separator:

* Correct: `0`, `1`, `0.5`
* Incorrect: `"0"`, `"1"`, `"0,5"`, `"0.5"`.

EngineerCoding:
I have validated that, and the opacity is working for the icon itself. When I toggle the visiblity of the layer itself, the OSM layer is visible again. So that indicates to me that the layer is drawing the background itself, and it is not omitting the other layers.

Also given that it only happens when the icon is in the rendered extent, might this something to do with feature rendering?

mike-000:
What you describe is very similar to what happens in https://openlayers.org/en/latest/examples/layer-clipping-vector.html if an extent is not set on the OSM layer.  Are there any prerender or postrender events on either of the layers?  Giving one of the the layers a unique `className` might be a workaround.

EngineerCoding:
I don't see how clipping is an issue, as I am able to do this in a sample app. However, I did notice something odd, when I use `.addFeatures` as opposed to `.addFeature`, I get different results:

Only showing polygon: https://codepen.io/engineercoding/pen/bGWvmxo
Showing both line and polygon: https://codepen.io/engineercoding/pen/mdmxzQO
[...]

PR #12619, opened 2021-08-10
Additional default .ol-viewport css settings

### LINKED ISSUE #12439: More robust css
Would it be practical to make the css more robust and set appropriate values within the ol classes to override anything from the application or other libraries it may use which might interfere with the correct operation of the map?

For example, an OpenLayers map should not be affected by canvas size rules needed elsewhere on the page as in https://gis.stackexchange.com/questions/401776/mis-alignment-of-gpx-data-layer-on-os-map

mike-000:
Fixes #12439 and #12566

Adds default settings to `.ol-viewport` to reset some settings observed to break or give poor user experience if set by third party libraries.

### DIFF
--- a/src/ol/ol.css
+++ b/src/ol/ol.css
@@ -68,12 +68,25 @@
   display: none;
 }
 .ol-viewport, .ol-unselectable {
+  background: transparent;
   -webkit-touch-callout: none;
   -webkit-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
   -webkit-tap-highlight-color: rgba(0,0,0,0);
+}
+.ol-viewport canvas {
+  background: transparent;
+  max-height: none;
+  max-width: none;
+  min-height: 0;
+  min-width: 0;
+}
+.ol-viewport div {
+  background: transparent;
+  font-size: medium;
+  font-weight: normal;
 }
 .ol-selectable {
   -webkit-touch-callout: default;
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
I'm wondering if it wouldn't be enough to set
```css
.ol-viewport, .ol-viewport div, .ol-viewport canvas {
  all: initial;
}
```

mike-000:
That would be nicer, although I had avoided using initial due to it not working in IE.  Starting with IE fallbacks for the most obvious potential problems, followed by the `all: initial` as suggested would be an alternative, as the fallbacks could be easily removed if support is dropped at a future release.

ahocevar:
I think we can start ignoring Internet Explorer specifics, especially when they are easy to work around on the application level.

Also, I'd set `all: initial` on the respective elements programmatically when they are created, to make things work correctly even when `ol.css` is not used.

mike-000:
The effects of using initial for divs are too far reaching, and I have now noticed that even just the clearing the background has broken the control button white borders in the deploy-preview examples.  However css which sets the background of divs by type selector only is unlikely to be encountered.  It is also difficult to know if fonts have been set intentionally, as in the examples `layout.css`, or as an unwanted effect of another library as in the Mapbox-gl Layer example.  That leaves canvas, where it would be easy to set the style directly on the element in `createCanvasContext2D`.

mike-000:
Replaced by #12626

PR #12626, opened 2021-08-15
Set canvas style to override problem 3rd party css

### LINKED ISSUE #12439: More robust css

mike-000:
Fixes #12439 and #12566
Replaces #12619

Set new canvas styles to `all: initial;` to override problems caused by third party css,  This is possible as OpenLayers only ever sets styles on canvas directly.

### DIFF
--- a/src/ol/dom.js
+++ b/src/ol/dom.js
@@ -19,12 +19,16 @@ export function createCanvasContext2D(
   opt_canvasPool,
   opt_Context2DSettings
 ) {
-  const canvas =
+  const canvas = /** @type {HTMLCanvasElement} */ (
     opt_canvasPool && opt_canvasPool.length
       ? opt_canvasPool.shift()
       : WORKER_OFFSCREEN_CANVAS
       ? new OffscreenCanvas(opt_width || 300, opt_height || 300)
-      : document.createElement('canvas');
+      : document.createElement('canvas')
+  );
+  if (canvas.style) {
+    canvas.style.all = 'initial';
+  }
   if (opt_width) {
     canvas.width = opt_width;
   }
@@ -32,9 +36,7 @@ export function createCanvasContext2D(
     canvas.height = opt_height;
   }
   //FIXME Allow OffscreenCanvasRenderingContext2D as return type
-  return /** @type {CanvasRenderingContext2D} */ (
-    canvas.getContext('2d', opt_Context2DSettings)
-  );
+  return canvas.getContext('2d', opt_Context2DSettings);
 }
 
 /**
--- a/src/ol/dom.js
+++ b/src/ol/dom.js
@@ -19,14 +19,14 @@ export function createCanvasContext2D(
   opt_canvasPool,
   opt_Context2DSettings
 ) {
-  const canvas = /** @type {HTMLCanvasElement} */ (
-    opt_canvasPool && opt_canvasPool.length
-      ? opt_canvasPool.shift()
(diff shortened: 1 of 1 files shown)


src/ol/dom.js
```diff
@@ -19,22 +19,24 @@ export function createCanvasContext2D(
...
     opt_canvasPool && opt_canvasPool.length
       ? opt_canvasPool.shift()
       : WORKER_OFFSCREEN_CANVAS
       ? new OffscreenCanvas(opt_width || 300, opt_height || 300)
-      : document.createElement('canvas');
+      : document.createElement('canvas')
+  );
+  if (canvas.style) {
+    canvas.style.all = 'initial';
+  }
```
ahocevar, line 31: With the type cast you added above, it should not be necessary to check for `canvas.style` beforehand:

```suggestion
  canvas.style.all = 'initial';
```

ahocevar (changes requested):

ahocevar:
Thinking about this more, I realize that this will at least break right-to-left text support.

src/ol/dom.js
```diff
@@ -19,22 +19,24 @@ export function createCanvasContext2D(
   opt_canvasPool,
   opt_Context2DSettings
 ) {
-  const canvas =
+  const canvas = /** @type {HTMLCanvasElement} */ (
```
ahocevar, line 22: An `OffscreenCanvas` is not a `HTMLCanvasElement`. So I think you'll need to enter two different paths for `OffscreenCanvas` and `HTMLCanvasElement`.

mike-000: `OffscreenCanvas` does not have a style object which is is the reason why the `if (canvas.style) {` test is needed to prevent an error.  Apart from that the properties and methods are the same as the end result was already typecast as `CanvasRenderingContext2D`.  It is one typecast instead of three.

ahocevar: I was envisioning a change without any additional type cast. Like this:
```diff
--- a/src/ol/dom.js
+++ b/src/ol/dom.js
@@ -19,12 +19,16 @@ export function createCanvasContext2D(
   opt_canvasPool,
   opt_Context2DSettings
 ) {
-  const canvas =
-    opt_canvasPool && opt_canvasPool.length
-      ? opt_canvasPool.shift()
-      : WORKER_OFFSCREEN_CANVAS
-      ? new OffscreenCanvas(opt_width || 300, opt_height || 300)
-      : document.createElement('canvas');
+  /** @type {HTMLCanvasElement|OffscreenCanvas} */
+  let canvas;
+  if (opt_canvasPool && opt_canvasPool.length) {
+    canvas = opt_canvasPool.shift();
+  } else if (WORKER_OFFSCREEN_CANVAS) {
+    canvas = new OffscreenCanvas(opt_width || 300, opt_height || 300);
+  } else {
+    canvas = document.createElement('canvas');
+    canvas.style.all = 'initial';
+  }
   if (opt_width) {
     canvas.width = opt_width;
   }
```

ahocevar (approved):
Thanks, @mike-000

PR #12632, opened 2021-08-16
Only warn of zero size when map should be visible

### LINKED ISSUE #12625: OverviewMap control issues warning about zero height and width when collapsed
**Describe the bug**
When a OverviewMap control is present and is collapsed, it issues warnings about zero height and width of map.

**To Reproduce**
1. Go to 'https://openlayers.org/en/latest/examples/overviewmap.html'
2. Open the developer console
3. See the warnings

**Expected behavior**
No warnings when overview map is not collapsed. Try setting `collapsed: false` and there will be no error.

mike-000:
Fixes #12625

Solution based on https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom/33456469#33456469


ahocevar (approved):
Thanks, @mike-000

Issue #8136, opened 2018-04-28
No hit detection when using ol.style.Style.renderer


strikehawk:
Thank you for your interest in making OpenLayers better!

To keep this project manageable for maintainers, we ask you to please check all boxes below before submitting an issue.

- [x] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [x] I have searched GitHub to see if a similar bug or feature request has already been reported.
- [x] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [x] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

Hi,

I am using [ol.style.Style.renderer](http://openlayers.org/en/latest/apidoc/ol.style.Style.html) to do a custom rendering of features, such as in this CodePen example : https://codepen.io/anon/pen/PeWYZM.

Hit detection fails for these feature, as it can be seen when hovering the red square: cursor is not changing (while it works for the feature using a regular icon style). The same behavior can be seen when using other types of geometry (such as line strings or polygons).

Am I missing something? 

From what I see in the code of [ol.render.canvas.Replay](https://github.com/openlayers/openlayers/blob/master/src/ol/render/canvas/Replay.js), custom draw instructions are correctly added to the `instructions` array, but not to the `hitDetectionInstructions` one.
I guess this explains the absence of hit detection.

I can propose two ways of fixing the problem:

**1. Adding the same custom instruction to both `instructions` and `hitDetectionInstructions` arrays:**
```
...
if (type == GeometryType.POINT) {
    flatCoordinates = geometry.getFlatCoordinates();
    this.coordinates.push(flatCoordinates[0], flatCoordinates[1]);
    replayEnd = this.coordinates.length;
    this.instructions.push([CanvasInstruction.CUSTOM,
      replayBegin, replayEnd, geometry, renderer]);
    this.hitDetectionInstructions.push([CanvasInstruction.CUSTOM,
      replayBegin, replayEnd, geometry, renderer]);
  }
...
```

I tried this fix, and it seems to be working (but I had no time to conduct extensive tests).

**2. Adding an optional `hitDetectionRenderer` property to [ol.style.Style](http://openlayers.org/en/latest/apidoc/ol.style.Style.html):**
[...]

PR #12646, opened 2021-08-20
Support for hit detection in styles with custom rendering

ashchurova:
This fix is for [Issue 8136](https://github.com/openlayers/openlayers/issues/8136). As per @ahocevar request defined new optional property for `ol.style.Style` type called `hitDetectionRenderer` and it is used if specified. If it is not specified logic in `CanvasBuilder.drawCustom` falls back to passed renderer. This is my first time contributing so I am new to the process.  @ahocevar you did mention in your comments that you wanted use hitDetectionRenderer when no custom renderer is specified. Still researching how to do that. Because that would mean recoupling drawing from hit detection instructions.

### DIFF
--- a/examples/custom-hit-detection-renderer.js
+++ b/examples/custom-hit-detection-renderer.js
@@ -71,19 +71,17 @@ circleFeature.setStyle(
   })
 );
 
-const vectorLayer = new VectorLayer({
-  source: new VectorSource({
-    features: [circleFeature],
-  }),
-});
-
 const map = new Map({
   layers: [
     new TileLayer({
       source: new OSM(),
       visible: true,
     }),
-    vectorLayer,
+    new VectorLayer({
+      source: new VectorSource({
+        features: [circleFeature],
+      }),
+    }),
   ],
   target: 'map',
   view: new View({
--- a/examples/custom-hit-detection-renderer.js
+++ b/examples/custom-hit-detection-renderer.js
@@ -1,0 +1,111 @@
+import Feature from '../src/ol/Feature.js';
+import Map from '../src/ol/Map.js';
+import View from '../src/ol/View.js';
+import {Circle} from '../src/ol/geom.js';
+import {OSM, Vector as VectorSource} from '../src/ol/source.js';
+import {Style} from '../src/ol/style.js';
+import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
+import {fromLonLat} from '../src/ol/proj.js';
+
+const columbusCircleCoords = fromLonLat([-73.98189, 40.76805]);
+const labelTextStroke = 'rgba(120, 120, 120, 1)';
+const labelText = 'Columbus Circle';
+
+let pointerOverFeature = null;
+
+const renderLabelText = (ctx, x, y, stroke) => {
+  ctx.fillStyle = 'rgba(255,0,0,1)';
+  ctx.strokeStyle = stroke;
+  ctx.lineWidth = 1;
+  ctx.textAlign = 'center';
+  ctx.textBaseline = 'middle';
+
+  ctx.font = `bold 30px verdana`;
+  ctx.moveTo(x, y);
+  ctx.lineTo(x, y);
+  ctx.filter = 'drop-shadow(7px 7px 2px #e81)';
+  ctx.fillText(labelText, x, y);
+  ctx.strokeText(labelText, x, y);
+};
+
+const circleFeature = new Feature({
+  geometry: new Circle(columbusCircleCoords, 50),
+});
+
+circleFeature.set('label-color', labelTextStroke);
+
+circleFeature.setStyle(
+  new Style({
+    renderer(coordinates, state) {
+      const [[x, y], [x1, y1]] = coordinates;
+      const ctx = state.context;
+      const dx = x1 - x;
+      const dy = y1 - y;
+      const radius = Math.sqrt(dx * dx + dy * dy);
+
+      const innerRadius = 0;
+      const outerRadius = radius * 1.4;
+
+      const gradient = ctx.createRadialGradient(
+        x,
+        y,
+        innerRadius,
+        x,
+        y,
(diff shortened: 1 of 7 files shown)


ahocevar (commented):
As far as I can tell, this is exactly what I had in mind! Maybe add a test and an example, then this should be good.

ashchurova:
@ahocevar OK :) Sorry for the delayed response, had a long labor day weekend. So I will add tests and and example to your samples library. I need to think about what would be a good case to have different renderer for hit detection.

ashchurova:
@ahocevar Hi Andreas, I am having trouble starting `serve-examples` while working on the example.  When I run `npm run serve-examples` I get an error: 

```
[webpack-cli] Failed to load 'C:\Dev\openlayers_my_fork\examples\webpack\config.mjs' config
[webpack-cli] Error: Not supported
```

Any suggestion on how to fix this one?

ashchurova:
@ahocevar Hi Andrea, added a sample for custom hit renderer, I kind of based it on the existing custom renderer sample for circle and showed that only label is used for hit detection. Let me know what you think. Thanks.

examples/custom-hit-detection-renderer.js
```diff
@@ -0,0 +1,111 @@
...
+  const featureOver = map.forEachFeatureAtPixel(evt.pixel, (feature) => {
+    feature.set('label-color', 'rgba(255,255,255,1)');
+    return feature;
+  });
+
+  if (pointerOverFeature && pointerOverFeature != featureOver) {
+    pointerOverFeature.set('label-color', labelTextStroke);
+  }
+  pointerOverFeature = featureOver;
+  vectorLayer.getSource().changed();
```
MoonE, line 110: This line can be removed, otherwise the layer will be redrawn even when the pointerOverFeature does not change.
A redraw is triggered when setting the label color.

examples/custom-hit-detection-renderer.js
```diff
@@ -0,0 +1,111 @@
...
+      );
+      gradient.addColorStop(0, 'rgba(255,0,0,0)');
+      gradient.addColorStop(0.6, 'rgba(255,0,0,0.2)');
+      gradient.addColorStop(1, 'rgba(255,0,0,0.8)');
+      ctx.beginPath();
+      ctx.arc(x, y, radius, 0, 2 * Math.PI, true);
+      ctx.fillStyle = gradient;
+      ctx.fill();
+
+      ctx.arc(x, y, radius, 0, 2 * Math.PI, true);
```
MoonE, line 65: This line can be removed as the same arc has already been created for the fill.

examples/custom-hit-detection-renderer.js
```diff
@@ -0,0 +1,111 @@
...
+const renderLabelText = (ctx, x, y, stroke) => {
+  ctx.fillStyle = 'rgba(255,0,0,1)';
+  ctx.strokeStyle = stroke;
+  ctx.lineWidth = 1;
+  ctx.textAlign = 'center';
+  ctx.textBaseline = 'middle';
+
+  ctx.font = `bold 30px verdana`;
+  ctx.moveTo(x, y);
+  ctx.lineTo(x, y);
```
MoonE, line 25: The moveTo / lineTo seem to be not needed.

examples/custom-hit-detection-renderer.js
```diff
@@ -0,0 +1,111 @@
...
+  ctx.filter = 'drop-shadow(7px 7px 2px #e81)';
```
MoonE, line 26: Seems like Firefox does not draw a drop-shadow when the stoke / fill is outside the canvas, which means there is no hitdetection on the drop-shadow in Firefox. Doesn't matter much for this example though, just something I noticed.

examples/custom-hit-detection-renderer.js
ashchurova: Removed. Thanks for pointing this out.

examples/custom-hit-detection-renderer.js
ashchurova: Done.

examples/custom-hit-detection-renderer.js
ashchurova: You are correct, since I am updating feature properties it would force re-drawing.

examples/custom-hit-detection-renderer.js
ashchurova: Ok, I have tested this only with Chrome. Thanks.

ashchurova:
@ahocevar Hi Andreas, no sure why these unit tests are failing on the server.  All unit tests pass on my machine. Could it be that there is a race condition some sort? Thank you.

ashchurova:
> @ahocevar Hi Andreas, no sure why these unit tests are failing on the server. All unit tests pass on my machine. Could it be that there is a race condition some sort? Thank you.

Now Browser tests ran passed. I interesting.. @MoonE I have addressed all your requested changes.

ashchurova:
@ahocevar any update on when this PR will be merged or any other feedback? Thanks.

ahocevar:
Hey @ashchurova, sorry this is taking so long. Unfortunately I have been too busy with my day job and FOSS4G preparation. I'll be looking into it as soon as possible.

ahocevar (approved):
Excellent work, @ashchurova! Thanks a lot for this contribution, and thanks for adding an example that shows custom style renderers in action.

ahocevar:
@MoonE, I think all your comments were properly addressed, but please double-check and approve when you're happy too.

MoonE (approved):
The requested changes are good, thanks @ashchurova.

ashchurova:
@ahocevar @MoonE Yeay!! When do you plan to publish this change? Thank you.

ahocevar:
@ashchurova It is already published on the dev tag. `npm install ol@dev` will pull that for you. It will of course be in the next release. We just published v6.8, and we publish releases approximately every 2 months.

Issue #12638, opened 2021-08-19
"ol.interaction.Translate" specifies the "features" attribute is invalid


imzhy:
version: 6.6.1

The specified "Feature"(data.editFeature) has been created and added to the map, but it cannot be dragged and there is no gripper

```javascript
data.translate = new ol.interaction.Translate({
  features: new ol.Collection([data.editFeature]),
  hitTolerance: 5
});
data.map.addInteraction(data.translate);
```

The same way works fine in "ol.interaction.Modify"

```javascript
data.modify= new ol.interaction.Modify({
  features: new ol.Collection([data.editFeature])
});
data.map.addInteraction(data.modify);
```

ahocevar:
I can confirm this. The expected behavior would be that at construction time, if the `features` collection contains features, these should be draggable. To make it consistent with the Modify interaction, it should also have a `source` option as alternative to the `features` option. If anyone would be willing to provide a pull request, it would be appreciated.

mike-000:
I cannot reproduce this as described.  There are already 3 filters available in the options for Translate - `features`, `filter` and `layer`.  Any or none of the 3 options can be used.  `layer` does not do anything `filter` could not do, and `filter` could filter by source as in `translate1` in https://codesandbox.io/s/translate-features-forked-grwj5?file=/main.js

Modify does not have an option to use features except from a collection or source, so unlike Translate at least one of the 2 options is required, plus they are mutually exclusive.  I don't think Translate needs any more options and whatever is done to make the interactions consistent is likely to be a breaking change.

**However** if the Translate collection contains multiple features as in `translate2` in the example they are all translated, even though the interaction only works when the pointer is over a feature satisfying the filter.  **This seems like a separate bug.**

mike-000:
Looking at this further it looks like the primary purpose of the `features` option is to allow multiple features to be translated together and would normally be used in conjunction with a Select interaction or similar.  Therefore it does not serve the same purpose as the `features` option in Modify (which will only operate on multiple features if they share vertices) so a `source` option is not needed for consistency.  There is no easy way to validate all features supplied in the `features` collection against filters involving layer, so if `features` is supplied both the `filter` and `layers` options should be ignored and default to true, similar to `features` and `source` being mutually exclusive in Modify.  If a Select interaction or similar is used to control the `features` collection it must be responsible for filtering.  The behaviour needs to be documented.

`filter` and `layers` options for Modify might be a nice-to-have alternative to `features` or `source`.

[...]

PR #12647, opened 2021-08-21
Remove Translate option conflicts and update documentation

mike-000:
Fixes https://github.com/openlayers/openlayers/issues/12638#issuecomment-902219976

May also fix the original issue which I cannot reproduce.


ahocevar (approved):
Thanks, @mike-000

PR #12671, opened 2021-08-27
Configure cache on the layer instead of the source

tschaub:
The WebGL tile layer renderer maintains a cache of tile textures - since it knows which are needed, maintaining it here makes sense.  Sources used by WebGL tile layers should not have to have a separate cache of tiles.  In the case of data tiles, we can set the cache size to (near) zero.

This branch fixes an issue with the WebGL tile layer renderer where the tile cache on the source is never pruned (since we never add the source key to the `usedTiles` lookup on the frame state.

This branch also adds a `cacheSize` option to WebGL tile layers.  This is passed to the renderer and defaults to 512.

### DIFF
--- a/src/ol/renderer/webgl/TileLayer.js
+++ b/src/ol/renderer/webgl/TileLayer.js
@@ -56,6 +56,8 @@ const attributeDescriptions = [
   },
 ];
 
+const empty = {};
+
 /**
  * Transform a zoom level into a depth value ranging from -1 to 1.
  * @param {number} z A zoom level.
@@ -103,6 +105,7 @@ function getRenderExtent(frameState) {
  * @property {Object<string, import("../../webgl/Helper").UniformValue>} [uniforms] Additional uniforms
  * made available to shaders.
  * @property {string} [className='ol-layer'] A CSS class name to set to the canvas element.
+ * @property {number} [cacheSize=512] The texture cache size.
  */
 
 /**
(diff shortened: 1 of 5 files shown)


src/ol/renderer/webgl/TileLayer.js
```diff
@@ -468,23 +472,17 @@ class WebGLTileLayerRenderer extends WebGLLayerRenderer {
...
 
     // TODO: let the renderers manage their own cache instead of managing the source cache
-    if (tileSource.canExpireCache()) {
-      /**
-       * @param {import("../../PluggableMap.js").default} map Map.
-       * @param {import("../../PluggableMap.js").FrameState} frameState Frame state.
-       */
-      const postRenderFunction = function (map, frameState) {
-        const tileSourceKey = getUid(tileSource);
-        if (tileSourceKey in frameState.usedTiles) {
```
tschaub, line 478: This was never true since the renderer never adds the tile source key to the `usedTiles` lookup.  Things would be simpler if we removed the tile cache from sources altogether (and instead maintained them on renderers).

ahocevar (approved):
Thanks for this, @tschaub!

src/ol/renderer/webgl/TileLayer.js
ahocevar: Thanks for fixing this. I agree it would simplify things to let the renderer cache tiles, instead of the source. Definitely something to keep in mind for a refactoring in the near future.

PR #12684, opened 2021-08-31
Fix view resolutions in example

mike-000:
Presumably the shift key was not working (a corresponding change to initial zoom may be needed depending on what was intended).


ahocevar (approved):
Thanks, @mike-000

PR #12696, opened 2021-09-04
Use union types instead of enums

### LINKED ISSUE #12432: Get rid of string enums, use union types instead
Our string enums, like those used for `*EventType`, are code that is redundant and only used for type checking. Instead, we could use union types that ensure types just as well, without adding payload to the build.

The easiest way to get documented event types in JSDoc while making TypeScript turn them into union types is a `@typedef` in JSDoc with a `keyof` operator in TypeScript:
```js
/**
 * @typedef {Object} MapBrowserEventTypeKeys
 * @property {string} singleclick A true single click with no dragging and no double click. Note that this
 * event is delayed by 250 ms to ensure that it is not a double click.
 * @property {string} click A click with no dragging. A double click will fire two of this.
 * @property {string} dblclick A true double click, with no dragging.
 * @property {string} pointerdrag Triggered when a pointer is dragged.
 * @property {string} pointermove Triggered when a pointer is moved. Note that on touch devices this is
 * triggered when the map is panned, so is not the same as mousemove.
 */

/**
 * @typedef {keyof MapBrowserEventTypeKeys} MapBrowserEventType
 */

/** @type {MapBrowserEventType} */
const a = 'pointerdrag';
```

See also https://blog.bam.tech/developer-news/should-you-use-enums-or-union-types-in-typescript.

ahocevar:
This pull request is a proof of concept for the idea I described in #12432.

`SourceState` is an enum that is used quite often, so it is a good candidate for analysing the effect of such a change on the build size. The difference is not significant. I think the most common scenario is that OpenLayers modules are part of a bundle with quite standard configurations. In this case, the gzipped size is a bit smaller with this change than without. In other scenarios (non gzipped, build with mangling enabled for properties suffixed with `_` like the legacy build), the build size is a bit larger after this change.

* full build, gzipped, without mangling for properties suffixed with `_`:
   before: 331,756 bytes
   after: 331,740 bytes
* full build, without mangling for properties suffixed with `_`:
   before: 1,301,295 bytes
   after: 1,301,407 bytes
* full build, gzipped, with mangling for properties suffixed with `_`:
   before: 326,060 bytes
   after: 326,078 bytes
* full build, with mangling for properties suffixed with `_`:
   before: 1,242,378 bytes
   after: 1,242,490 bytes

I'm putting this up for discussion. I like working with union types better than with string enums. What do others think? Is mangling common in build configurations? If so, is the marginal difference in build sizes something we should consider?

No need to merge this if others are not in favour of such changes. In that case we should maybe also close #12432.


tschaub:
Thanks for putting this up for discussion, @ahocevar.  It seems like this difference in build sizes is negligible.  Maybe you could provide a few benefits/drawbacks.  I assume it makes no difference from an app developer perspective.  If so, what are the differences in the lib dev experience?  I wonder about things like auto-complete or editor suggestions.

tschaub:
I also see nice suggestions in the editor:
image

I'm sold.

tschaub:
I added a commit that removes the `GeometryType` enum in favor of a union type.  If others like this direction, we could make changes across the board.

tschaub:
I stopped before reworking the event types, since we might want to have more discussion on the alternatives for those - mostly around their presentation in the docs (see https://github.com/openlayers/openlayers/issues/12432#issuecomment-913205829).

jahow:
Giving my opinion for what it's worth: I think union types are preferable to enums because they require less code and incur less cognitive load to the reader. Besides, enums are a bit awkward IMO, since they exist at runtime but are only useful during static typechecking. All in all, union types makes our lives better and we should use them!

fredj:
@ahocevar I did the change for `extent/Corner` and `ol/source/WMTSRequestEncoding`:
 * https://github.com/fredj/openlayers/commit/ef1309fd88694be746365fe9cdabd1cd45afab49
 * https://github.com/fredj/openlayers/commit/1f448e1ab6e8b9849841e7541d182ccd546a3b82

I you want I can push the commits to your branch

ahocevar:
Thanks, @fredj. Yes, please add these commits to my branch.

MoonE:
The union types play nicer with the typescript type definitions, so I got a little too excited and tried to replace the ImageState enum which of course does not work as it is not a string enum and using the numbers directly is not really a good idea if we care about readability.

Do you have an idea on how to better deal with those?

tschaub:
I've wished those image states were strings many times.  Since they are not part of the API, we could change them to strings.  What do others think?

Rezorl:
I need it. 👍

I am upgrading ol library from 5.3.0 to 6.11.0 and I have a lot of errors.

dkastl:
Regarding TypeScript there is also quite some effort put into another project (repository):
https://github.com/hanreev/types-ol

Wouldn't it make sense to join efforts?

tschaub:
> Regarding TypeScript there is also quite some effort put into another project (repository): https://github.com/hanreev/types-ol
> 
> Wouldn't it make sense to join efforts?

We would welcome contributions from @hanreev or others on improving the type annotations in this repo.  I think that it is more sustainable to manage the types in the same place we manage the source.

tschaub:
@ahocevar - do you think there is work here that can be salvaged, or should we let this go?

ahocevar:
I think it's better to start over.

ahocevar:
Spending some time now to see if this can be salvaged.

ahocevar:
I think this should be good again.

simonseyock (approved):
This looks very nice! Any chance to get that merged at some point?

ahocevar:
Finishing this now.

ahocevar:
Rebased, conflicts resolved, ready for review again.

src/ol/renderer/Layer.js
```diff
@@ -196,11 +195,7 @@ class LayerRenderer extends Observable {
    */
   renderIfReadyAndVisible() {
     const layer = this.getLayer();
-    if (
-      layer &&
-      layer.getVisible() &&
-      layer.getSourceState() == SourceState.READY
-    ) {
+    if (layer && layer.getVisible() && 'ready') {
```
MoonE, line 198:
```suggestion
    if (layer && layer.getVisible() && layer.getSourceState() == 'ready') {
```
The correct comparison was lost here.

ahocevar: Good find, @MoonE, thanks! I created #13835 to fix this, and also added a regression test.

PR #12697, opened 2021-09-05
Move description above type tag

tschaub:
This removes a warning from JSDoc about `@type` tags not being able to have a description.


MoonE (approved):
Thanks, @tschaub

PR #12698, opened 2021-09-05
Transitive dependency updates

tschaub:
This gets rid of two npm warnings.


tschaub:
Was thinking this would make a difference in the release, but it won’t.  We’ll just wait for dependabot.

PR #12709, opened 2021-09-06
Support rendering of GeoTIFF images in pixel coordinates

### LINKED ISSUE #12701: Cloud Optimized GeoTiffs (COG) without georeference can't be displayed.
**Describe the bug**
Cloud Optimized GeoTiffs without an explicit geoference can't be added to a map. Projection is ignored (it's set to null inside `GeoTiff.js` ) and since GeoKeys can't be set, internal methods for boundingbox and origin fails with a reference to a missing "affine transformation". 

It should be possible to use COGs like a static-image as in https://openlayers.org/en/latest/examples/static-image.html , with the "functionality" of https://openlayers.org/en/latest/examples/zoomify.html

In the images "FileDirectory" resides information about both tilesizes, resolution and dimensions so creating a TileGrid should be possible without the geokeys, if we allow projection to be local as in the "static-image" example. 

**To Reproduce**
1. Use this example COG: https://tmp-asger-public.s3.eu-central-1.amazonaws.com/2017_84_40_1_0015_00008062.cog.tif in the COG example
2. See that the layer is not added, and that the "error" property on the source complains about a missing affine transformation
3. See error

**Expected behavior**
I would expect Cloud Optimized Geotiffs to display with a local coordinate system as in the static-image example, if no georeference exists. I would expect that I would be able to show the COG in a pure local coordinate system or with a supplied projection. 

Side info: 
We have verified that this is definitely possible in an own branch, by passing a flag to `GeoTIFF.js` and gathering the information needed from the `FileDirectory` property of the tiff images, and by setting the origin point in the upper left corner `[0, h]`.

See: https://github.com/Septima/openlayers/commit/d85585275c2a56e124c6aff83e228cd1e86f2e79 for a working "hacked" example.

tschaub:
This adds support for rendering GeoTIFF images without an affine transform.

Fixes #12701.

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -25,12 +25,70 @@ import {toSize} from '../size.js';
  * infrared band, configure `bands: [3]`.
  */
 
+/**
+ * @typedef {Object} GeoTIFFImage
+ * @property {Object} fileDirectory The file directory.
+ * @property {Object} geoKeys The parsed geo-keys.
+ * @property {boolean} littleEndian Uses little endian byte order.
+ * @property {Object} tiles The tile cache.
+ * @property {boolean} isTiled The image is tiled.
+ * @property {function():Array<number>} getBoundingBox Get the image bounding box.
+ * @property {function():Array<number>} getOrigin Get the image origin.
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @tschaub!

PR #12711, opened 2021-09-06
Use band numbers starting with one

tschaub:
The `band` operator in raster style expressions works with band numbers starting with 1.  This is for consistency with the use of band numbers in GDAL and related software.  This branch updates the `bands` option for the GeoTIFF source to work with one-based band numbers as well (as opposed to a zero-based band index).  This is likely to trip people up, but I think we should at least be consistent between the `band` operator and the `bands` option.


ahocevar (approved):
Agreed! I had thought about this too.

PR #12714, opened 2021-09-06
Avoid creating duplicate projections in COG examples

tschaub:
This simplifies the COG examples by removing the duplicate projection creation.  Subtle, but the projection (registered with proj) had units but no extent, and the projection on the view had an extent but no units.


tschaub:
Merging as an example-only change.

PR #12726, opened 2021-09-09
Make removeFeature consistent with other remove methods

### LINKED ISSUE #12723: removeFeature() error inconsistent with other remove methods
**Describe the bug**
Calling the VectorSource `removeFeature()` method with a feature which is not in the source causes an error.  This is inconsistent with other remove methods such as `removeLayer()`.

**To Reproduce**
https://codesandbox.io/s/icon-negative-forked-ofl8c

**Expected behavior**
Do nothing if there is nothing to be removed..

mike-000:
Fixes #12723

Similar to `removeLayer`, `removeInteraction`, etc. avoid errors if the feature is not defined or has already been removed.
Return the removed feature if it is successfully removed.


ahocevar (commented):
I agree with most of the changes, but I'm wondering if it's really necessary to handle the case where no feature is passed to `removeFeature()`.

mike-000:
I was thinking of consistency with situations similar to adding/replacing interactions or layers where same function can be used for both first and subsequent calls as it does not matter if the "previous" value is undefined at the first call, e.g.:

```
let draw;
function addInteraction() {
  map.removeInteraction(draw);
  const value = typeSelect.value;
  if (value !== 'None') {
    draw = new Draw({
      source: source,
      type: typeSelect.value,
    });
    map.addInteraction(draw);
  }
}
```

ahocevar:
Well, I wouldn't encourage such coding style, but I understand if others want do do things like that.

src/ol/source/Vector.js
```diff
@@ -544,7 +544,10 @@ class VectorSource extends Source {
       }
     } else {
       if (this.featuresRtree_) {
-        this.featuresRtree_.forEach(this.removeFeatureInternal.bind(this));
+        // use Array forEach to ignore return
+        this.featuresRtree_
+          .getAll()
```
ahocevar, line 549: This makes things slower, and I do not see a need for it. What was the rationale behind this change?

mike-000: The Rtree f`orEach` method calls `getAll()`
https://github.com/openlayers/openlayers/blob/main/src/ol/structs/RBush.js#L154
but exits if the callback returns a truthy value (i.e. the feature)
https://github.com/openlayers/openlayers/blob/main/src/ol/structs/RBush.js#L178
That broke the `.clear()` test https://github.com/openlayers/openlayers/blob/main/test/browser/spec/ol/source/vector.test.js#L278 as although all the features were removed only one `removefeature` event was fired.  It is only going to be slower if Array `forEach` is slower than a `for` loop.

ahocevar: Oh no, that `structs/RBush` wrapper could use some love.

But thanks for the explanation! This means adding a return value to `removeFeatures()` could be a breaking change. Is the return value really useful?

mike-000: Yes it would break `map.forEachFeatureAtPixel(pixel, source.removeFeature.bind(source));`  I have never used the return values from the other remove methods so maybe that could wait until a full release,

PR #12727, opened 2021-09-09
Add missing 'boxstart' event to OnSignature

### LINKED ISSUE #12725: Typings of Dragbox are missing boxstart
**Describe the bug**
The Typings `DragBox.d.ts` are missing `boxstart` event.

So if I want to use something like  
```ts
const dragBox = new DragBox(dragBoxOptions);
dragBox.on('boxstart', (evt) => {
        ...
}

dragBox.on('boxend', (evt) => {
        ...
});
```
`boxend` is working but I get a TypeError for `boxstart` 

> No overload matches this call.
>  The last overload gave the following error.
>    Argument of type 'string' is not assignable to parameter of type '("error" | "change" | "propertychange" | "change:active" | >"boxdrag" | "boxend" | "boxcancel")[]'.ts(2769)
> DragBox.d.ts(66, 440): The last overload is declared here.

**To Reproduce**
You can also reproduce this on https://codesandbox.io/s/wk0bs?file=/main.js 

1. write `dragBox.on('` and the completion don't has any `boxstart`.

**Expected behavior**
The `boxstart` event should also be in `DragBoxOnSignature`

**Version**
- ol: 6.6.1

simonseyock:
Fixes #12725

### DIFF
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -98,7 +98,7 @@ export class DragBoxEvent extends Event {
  * @typedef {import("../Observable").OnSignature<import("../Observable").EventTypes, import("../events/Event.js").default, Return> &
  *   import("../Observable").OnSignature<import("../ObjectEventType").Types|
  *     'change:active', import("../Object").ObjectEvent, Return> &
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
Thanks, @simonseyock

Issue #10881, opened 2020-04-05
Issues with vector loaders when using useGeographic()


mike-000:
**Describe the bug**
I experimented to see if the Map Graticule example was compatible with useGeographic() and unsurprisingly it isn't.  The Graticule can only work correctly if it is passed the view projection and extent.  When useGeographic() is specified it receives an EPSG:4326 extent but a view projection resolution.  I also checked the Sphere Mollweide example and while `countries-110m.geojson` loads correctly with an EPSG:3857 view in the Geographic Editing example it doesn't with Sphere Mollweide (even without the graticule).

**To Reproduce**
Specify useGeographic() in the examples.

**Expected behavior**
Either loaders should always be passed the view projection and extent and it should be the loaders responsibility to handle user projections, or there should be a layer option for the renderer to always pass view projection and extent.  The Sphere Mollweide issue (a quick fix would be `toUserExtent` passing an `opt_stops` value to `transformExtent`, but it would be better to use the real view extent) suggests that even default loaders might need to handle user projections themselves.  Regardless of the solution the resolution should be in the same units as the projection and extent.

ahocevar:
Wouldn't it be enough to pass the user projection's resolution to the loader, instead of the view projection's resolution?

mike-000:
Yes, if the projection and extent are user projection the resolution should also be in user projection.  Currently it is always in view projection.  With non-parallel projections such as polar and Mollweide there can be some inaccuracy in extent transforms (even using the stops option) which might cause a reduction in extent, so an option to use view projection throughout in conjunction with a custom loader would be useful, especially for the graticule which would need to transform user projection extents back to view projection.

ahocevar:
Thinking about this more, the correct behavior for a loader would be to work with the user projection anyway. Not optionally, but always.

mike-000:
Currently it does work with user projection if one is set.  I was suggesting an option to use view projection regardless of user projection.

ahocevar:
I don't think we want that. A configured user projection really means that the user projection is used everywhere in the public API.

mike-000:
The problem with that is transforms between view projection and user projection introduce errors.  e.g. when bottom of the view projection is off the world the transform to user projection places it at -61 degrees so Antarctica is not rendered

![image](https://user-images.githubusercontent.com/49240900/78567600-d8332480-7818-11ea-8e4c-b94272693c9d.png)


[...]

PR #12731, opened 2021-09-09
Handle resolution and tile loadingstrategy with user projection

ahocevar:
This pull request fixes two problems with vector tile loaders:
* Resolutions need to be adjusted when units of user and view projection are different. This also fixes one of the problems described in #10881.
* Assuming we always interact with tile grids in view projection, the tile grid passed to `ol/loadingstrategy/tile` needs to be in view projection.

The pull request adds a new example that shows how to use a tile loading strategy with `useGeographic()`.

### DIFF
--- a/examples/vector-wfs-geographic.html
+++ b/examples/vector-wfs-geographic.html
@@ -1,13 +1,13 @@
 ---
 layout: example.html
-title: WFS
+title: WFS with geographic coordinates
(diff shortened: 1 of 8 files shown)


mike-000:
The example needs Geographic in the title and tags to distinguish it from the existing example

tschaub (approved):
Thanks for the work on this, @ahocevar

PR #12732, opened 2021-09-10
Add <meta charset="utf-8"> to Quick Start

mike-000:
While this is not currently needed with the jsdelivr build (when it is working) since v6.7.0 it is needed when a downloaded legacy build is used, and is good practice in all cases.


ahocevar (approved):
Thanks, @mike-000

Issue #12730, opened 2021-09-09
Home page map missing, legacy build issues


mike-000:
**Describe the bug**
The small map at the top of the home page https://openlayers.org/ is not appearing as there is no legacy build in the v6.7.0 folder.  Also jsdeliver is not serving the link suggested in the Quick Start.  When changed to use a downloaded `ol.js` it needs `<meta charset="utf-8">` adding to display the zoom out button correctly in Chrome, so that should be added to the example.  IE fails to load due to an arrow function.

ahocevar:
I re-deployed the release. Now the map on the home page works again.

mike-000:
It looks like geotiff.js introduced a considerable number of arrow functions and backtick quotes.  I presume the latter are necessary and cannot be transpiled, in which case the minimum browser requirement is now ES6.

ahocevar:
@mike-000 If properly transpiled, none of these should end up in the legacy build. I suspect the worker is built without transpilation, but I might be wrong.

PR #12734, opened 2021-09-10
Continue supporting IE11 in examples and legacy build

ahocevar:
This pull request modifies the webpack configurations for the example and legacy build to transpile for IE11, including all dependencies. Because geotiff brings in dependencies that publish code with async/await, we can no longer use buble for transpiling. So this pull request also gets rid of the buble dev dependency, and uses babel instead, which we have as dependency already for the worker build.

This pull request fixes one of the problems described in #12730.

### DIFF
--- a/config/webpack-config-legacy-build.mjs
+++ b/config/webpack-config-legacy-build.mjs
@@ -1,8 +1,5 @@
 import TerserPlugin from 'terser-webpack-plugin';
-import path, {dirname} from 'path';
-import {fileURLToPath} from 'url';
-
-const baseDir = dirname(fileURLToPath(import.meta.url));
+import path from 'path';
 
 export default {
   entry: './build/index.js',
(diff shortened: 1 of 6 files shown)


ahocevar:
@mike-000, can you please test the examples and the legacy build from this pull request in IE11? I don't have access to such an old browser.

mike-000:
Do you have a download link available?  The deploy preview examples do not work in any browser.

ahocevar:
@mike-000 The examples preview should work again now, and I added a legacy build test map to the preview entry page (https://deploy-preview-12734--ol-site.netlify.app).

mike-000:
Examples and legacy build working in both Chrome and IE (when copied and polyfills used)

ahocevar:
@mike-000 I added the polyfills to https://deploy-preview-12734--ol-site.netlify.app. Can you please test again if it works out of the box now in IE?

ahocevar:
I just found a rusty Windows virtual machine with Internet Explorer on it. I could verify that both the legacy build preview and the preview examples work fine there. So this is now ready for review.

mike-000:
Yes, it is working.  The new WebGL examples have highlighted some more polyfills are needed, e.g. Array.from()

ahocevar:
Thanks for testing, @mike-000.

> Yes, it is working. The new WebGL examples have highlighted some more polyfills are needed, e.g. Array.from()

I'll leave that to you to document/add those, if that's ok for you. Thanks in advance!

tschaub (approved):

PR #12736, opened 2021-09-10
OnSignatures return more specific type

### LINKED ISSUE #12733: Typescript signatures for .on() methods don't distinguish between listening to single and multiple events
The type signatures for the .on() methods (e.g. in Observable and all subclasses) don't allow correctly to distinguish between calling with a single event name (`.on('change', (e) => {...})`) which returns a single `EventsKey` and calling with multiple event names (`.on(['change', 'error'], (e) => {...})`), returning an array of `EventsKey`.

This makes using the result in Typescript somewhat inelegant as it is necessary to always cast to the correct type (`... as EventsKey` or `... as EventsKey[]` respectively).

(may also apply to other ways of attaching listeners which return one or more `EventsKey`).

Using a "pure" Typescript definition it would be fairly easy to correctly represent the overridden signature, similar to the following:
```ts
declare class Observable extends EventTarget {
    on(type: string, listener: (e: any) => any): EventsKey;
    on(type: string[], listener: (e: any) => any): EventsKey[];
    on(type: string|string[], listener: (e: any) => any): EventsKey|EventsKey[] {
        ...
    }

    ...
}
```

Is it possible to produce such overridden signatures using the JSDoc? I couldn't find a trivial way to do it, but maybe someone with more knowledge of the JSDoc -> Typescript declarations has a good solution?

simonseyock:
If `on` is called without an array it now returns a `EventsKey` and if it is called with an array it returns `EventsKey[]`.

Also fixes a bug in the `OnSignature` of `Observable` and removes the `OnReturn` typedef which is not needed anymore.

image

Fixes #12733

### DIFF
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -81,12 +81,12 @@ class Collection extends BaseObject {
     super();
 
     /***
-     * @type {CollectionOnSignature<import("./Observable.js").OnReturn>}
+     * @type {CollectionOnSignature<import("./events").EventsKey>}
      */
     this.on;
 
     /***
-     * @type {CollectionOnSignature<import("./Observable.js").OnReturn>}
+     * @type {CollectionOnSignature<import("./events").EventsKey>}
      */
     this.once;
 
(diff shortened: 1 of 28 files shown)


tschaub (approved):

tschaub:
Thanks for the continued work on this, @simonseyock

PR #12739, opened 2021-09-10
Change canvas style from initial to unset

### LINKED ISSUE #12735: Cannot set cursor style
**Describe the bug**
I updated openlayers to 6.7 version and my `map.getViewport().style.cursor = 'some cursor'` stops working. Now, layers canvases have `cursor: initial` style, which force cursor to be `pointer`, cause of canvases are on top of viewport.

mike-000:
Fixes #12735

Unset achieves the same results as #12626 over unwanted css such as min-width without interfering with desired inherited properties.


ahocevar (approved):
Thanks for fixing this regression, @mike-000

PR #12740, opened 2021-09-11
Various fixes for browser compatibility issues

mike-000:
Various fixes for browser compatibility issues seen in the examples.

In the library change `Math.log2` to use `ol/math/log2` and add a fallback for `TypedArray.from()`

Ensure examples which use the `input` event for range sliders also listen for `change` (as already done in Heatmap and WebGL filtering examples) as there are inconsistencies between browsers.

It seems unrealistic to expect the GeoTiff examples to run on old browsers because the workers require iterator support.

### DIFF
--- a/examples/cluster.js
+++ b/examples/cluster.js
@@ -77,13 +77,17 @@ const map = new Map({
   }),
 });
 
-distanceInput.addEventListener('input', function () {
+const distanceHandler = function () {
   clusterSource.setDistance(parseInt(distanceInput.value, 10));
-});
+};
+distanceInput.addEventListener('input', distanceHandler);
+distanceInput.addEventListener('change', distanceHandler);
 
-minDistanceInput.addEventListener('input', function () {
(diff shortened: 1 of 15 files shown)


tschaub:
Are both `input` and `change` listeners needed?  If one will do, that would be preferable.  Also, the more explicit you can be about what issue you are trying to address (and in what browsers), the better.  When people revisit these changes in the future, they will appreciate the detail.

mike-000:
`input` usually fires while the slider is being dragged, so provides an immediate response.  `change` usually fires when only a move is complete.  This is quite an old answer https://stackoverflow.com/a/19067260/10118270 but IE11 still does not fire `input` (but does fire `change` when `input` would be expected) and Chrome now behaves as expected.  I have not tried Firefox or Safari, but their behaviour may affect https://openlayers.org/en/latest/examples/fractal.html which is the sole example which uses only the `change` event.

tschaub (approved):

PR #12743, opened 2021-09-13
Remove docs reference to unmanaged layer and Select

### LINKED ISSUE #12742: Features from Select are not shown on top
Features styled using Select are not on top as expected from "features added to an internal unmanaged layer".

See this demo, slightly adjusted [Layer Z-Index example](https://openlayers.org/en/latest/examples/layer-z-index.html):
https://gist.githubusercontent.com/jacekkow/68cb37d17f686b51a7ad1c75af3cb9a4/raw/52ff63d951a9086501d5a698389fed75e6d58061/ol_select.html

Clicking on the triangle does not bring it to the foreground - feature stays in the background, but according to docs of [ol.interaction.Select](https://openlayers.org/en/latest/apidoc/module-ol_interaction_Select-Select.html):
> Selected features are added to an internal unmanaged layer.

and per [ol.layer.Layer](https://openlayers.org/en/latest/apidoc/module-ol_layer_Layer-Layer.html) docs:
> Components like module:ol/interaction/Select\~Select use unmanaged layers internally. These unmanaged layers are associated with the map using module:ol/layer/Layer\~Layer#setMap instead.
> (...)
> a zIndex of 0 is assumed for layers that are added to the map's layers collection, or **Infinity** when the layer's setMap() method was used.

So the triangle (becomes a circle after selection due to the custom style on Select) should be in front of everything else.

As far as I can tell there is no unmanaged layer at all in Select since commit https://github.com/openlayers/openlayers/commit/2ccaed0acec464c378cae0ba90be820124b0d7c5 in 2019, which broke the documented behavior.

mike-000:
Fixes #12742

Remove unmanaged layer reference from Select documentation, and change Layer description to use Draw

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -154,8 +154,6 @@ const originalFeatureStyles = {};
  * `toggle`, `add`/`remove`, and `multi` options; a `layers` filter; and a
  * further feature filter using the `filter` option.
  *
- * Selected features are added to an internal unmanaged layer.
- *
  * @fires SelectEvent
  * @api
  */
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #12759, opened 2021-09-15
#12758: Allow undefined to be passed to control setMap(). Also fix type of getMap()

### LINKED ISSUE #12758: TypeScript: ol/control/Control setMap does not allow for undefined
**Describe the bug**
The `setMap` method of the `Control` class does not allow passing in `undefined` in TypeScript.

If you look at the underlying code in this method, it is clear that `this.map_` being set/get can be `undefined`

**To Reproduce**

1. Attempt to pass `undefined` to `setMap` of any `Control` instance.

**Expected behavior**

It should be allowed, instead the TypeScript compiler will error: `Argument of type 'undefined' is not assignable to parameter of type 'PluggableMap'`

jumpinjackie:
Fixes #12758

### DIFF
--- a/src/ol/control/Control.js
+++ b/src/ol/control/Control.js
@@ -98,7 +98,7 @@ class Control extends BaseObject {
 
   /**
    * Get the map associated with this control.
-   * @return {import("../PluggableMap.js").default} Map.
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @jumpinjackie

PR #12760, opened 2021-09-15
Fix opacity handling for tile and vector layers

ahocevar:
This pull request fixes two issues with opacity handling:

* Tile layers currently multiply the layer opacity with the tile transition opacity, and apply the result to the tile. This causes flicker when lower resolution tiles overlay with target resolution tiles during tile loading on layers that have an opacity other than `1` set. I'm surprised noone has noticed this since v6.0:
   Sep-15-2021 19-31-11
   The flickering (ocean temporarily appearing in a darker blue) can be seen in the above animation.

* Vector layers currently set the opacity on the container on every render frame, instead of only when it has changed. This adds unnecessary DOM processing overhead.

This fix sets the tile layer opacity on the container (same as for vector layers), and only applies the tile transition opacity on individual tiles. For all layers, the new `cssOpacity()` function makes sure that for an opacity of 1, `''` is compared with `''` instead of `1`. For correct container reuse, the `useContainer` function also gets a little improvement so tile layers with the same opacity can share a container.

### DIFF
--- a/src/ol/css.js
+++ b/src/ol/css.js
@@ -114,3 +114,11 @@ export const getFontParameters = function (fontSpec) {
   style.families = style.family.split(/,\s?/);
   return style;
 };
+
+/**
+ * @param {number} opacity Opacity (0..1).
+ * @return {string} CSS opacity.
+ */
+export function cssOpacity(opacity) {
+  return opacity === 1 ? '' : String(Math.round(opacity * 100) / 100);
+}
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -16,6 +16,7 @@ import {
   getIntersection,
   getTopLeft,
 } from '../../extent.js';
+import {cssOpacity} from '../../css.js';
 import {fromUserExtent} from '../../proj.js';
 import {getUid} from '../../util.js';
 import {numberSafeCompareFunction} from '../../array.js';
@@ -410,8 +411,7 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
           w,
           h,
           tileGutter,
-          transition,
-          layerState.opacity
+          transition
         );
         if (clips && !inTransition) {
           context.restore();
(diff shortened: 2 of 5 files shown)


ahocevar:
This is now ready for review.

src/ol/renderer/canvas/TileLayer.js
```diff
@@ -465,16 +470,14 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
...
-   * @param {number} opacity Opacity.
    */
-  drawTileImage(tile, frameState, x, y, w, h, gutter, transition, opacity) {
+  drawTileImage(tile, frameState, x, y, w, h, gutter, transition) {
     const image = this.getTileImage(tile);
     if (!image) {
       return;
     }
     const uid = getUid(this);
-    const tileAlpha = transition ? tile.getAlpha(uid, frameState.time) : 1;
```
tschaub, line 476: I agree with the linter that `tilerAlpha` is no longer defined (though used below).

src/ol/css.js
```diff
@@ -114,3 +114,11 @@ export const getFontParameters = function (fontSpec) {
...
   return style;
 };
+
+/**
+ * @param {number} opacity Opacity (0..1).
+ * @return {string} CSS opacity.
+ */
+export function cssOpacity(opacity) {
+  return opacity === 1 ? '' : String(opacity);
+}
```
MoonE, line 124: The css opacity may be rounded when it is set on a html node. E. g. `1 / 3` becomes `"0.333333"`.
To allow container reuse in that case it might make sense to do something similar to the toString transform function. Lines 285-288
https://github.com/openlayers/openlayers/blob/cb6995d71ae84bfa41e93852d23dae5019814b0b/src/ol/transform.js#L274-L289

src/ol/renderer/canvas/TileLayer.js
ahocevar: Oh, this is what you get from last minute force-pushes. Sorry for having marked this ready for review without double-checking.

src/ol/css.js
ahocevar: I decided to use simple rounding instead of going through the DOM, because this function would be called with every render frame, and the required DOM operation is probably almost as expensive as setting the opacity on every render frame, which I wanted to avoid.

tschaub (approved):

Issue #12763, opened 2021-09-16
Typedef wrong for ImageInformationResponse in module ol/format/IIIFInfo


webwitcher:
**Describe the bug**
Response object for IIIF API version 2 and 3 not assignable to type `ImageInformationResponse`. `tiles` property of response object does not match any type, because it is an array of `TileInfo`. The bug has been introduced by commit: https://github.com/openlayers/openlayers/commit/bb0b94fedf0c29b30e9a3bea71f47cf557584bc7#diff-e804e2aec4eb823cdae866ee86eafb3af1ae2f190c412f698dbacebd0e344ec2

**To Reproduce**
n/a

**Expected behavior**
Typedef `ImageInformationResponse` should conform to `{Object<string,string|number|Array<number|string|IiifProfile>|Object<string, number>|Array<TileInfo>>}` in https://github.com/openlayers/openlayers/blob/15cc849febe9226e32b16ee8b08a74c6a85c7303/src/ol/format/IIIFInfo.js#L43

marcjansen:
Can you come up with a PR, @webwitcher ? Would @simonseyock be available for a review?

tschaub:
Fixed by #12764

PR #12764, opened 2021-09-16
Fix ImageInformationResponse typedef

webwitcher:
`TileInfo` should be an array to conform to https://iiif.io/api/image.

For more information see: https://github.com/openlayers/openlayers/issues/12763

### DIFF
--- a/src/ol/format/IIIFInfo.js
+++ b/src/ol/format/IIIFInfo.js
@@ -40,7 +40,7 @@ import {includes} from '../array.js';
  */
 
 /**
- * @typedef {Object<string,string|number|Array<number|string|IiifProfile>|Object<string, number>|Array<TileInfo>>}
+ * @typedef {Object<string,string|number|Array<number|string|IiifProfile|Object<string, number>|TileInfo>>}
  *    ImageInformationResponse
  */
 
--- a/src/ol/format/IIIFInfo.js
+++ b/src/ol/format/IIIFInfo.js
@@ -40,7 +40,7 @@ import {includes} from '../array.js';
  */
 
 /**
- * @typedef {Object<string,string|number|Array<number|string|IiifProfile>|Object<string, number>|TileInfo>}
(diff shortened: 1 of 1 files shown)


simonseyock (approved):
Very nice! Thank you for your contribution!

src/ol/format/IIIFInfo.js
```diff
@@ -40,7 +40,7 @@ import {includes} from '../array.js';
  */
 
 /**
- * @typedef {Object<string,string|number|Array<number|string|IiifProfile>|Object<string, number>|TileInfo>}
+ * @typedef {Object<string,string|number|Array<number|string|IiifProfile>|Object<string, number>|Array<TileInfo>>}
```
simonseyock, line 43:
```suggestion
 * @typedef {Object<string,string|number|Array<number|string|IiifProfile|Object<string, number>|TileInfo>>}
```

Comparing with the state before the faulty commit this might be the desired typing. `sizes` seemed to be an array as well, before.

simonseyock:
Hey, after looking into the original code I found another probable bug that could be fixed in one go.

simonseyock (changes requested):

src/ol/format/IIIFInfo.js
simonseyock: This seems to be correct after the cited spec as well: https://iiif.io/api/image/3.0/#53-sizes

simonseyock: The type would be a little stricter in your version, though. You could also keep your version and just add another `Array` around `Object<string,number>` because they can not get mixed.

webwitcher: I am good with your proposal. Thank you.

tschaub (approved):

tschaub:
Thanks for the fix @webwitcher and @simonseyock.

PR #12761, opened 2021-09-15
Add Float Textures for GL Rendering / DataTiles

theduckylittle:
1. Adds the ability to use a `Float32Array` data type in the texture.
2. Fixes bug with > 4 band imagery textures not being set properly.
3. Allows `DataTile` to accept a variable number of bands through an option.
4. Includes an exmaple of using a 5-band floating point source with NumpyTiles.

PR #12766, opened 2021-09-16
Use only Uint8Array with gl.UNSIGNED_BYTE for browser compatibility

mike-000:
Fixes a compatibility issue with can be seen on the Data Tiles example with IE11 and some less old Chromium based Android browsers

image
https://bugs.chromium.org/p/chromium/issues/detail?id=961658

### DIFF
--- a/src/ol/webgl/TileTexture.js
+++ b/src/ol/webgl/TileTexture.js
@@ -70,7 +70,7 @@ function uploadDataTexture(gl, texture, data, size, bandCount) {
     0,
     format,
     gl.UNSIGNED_BYTE,
-    data
+    data instanceof Uint8Array ? data : new Uint8Array(data.buffer)
   );
 }
 


tschaub:
This will conflict with #12761.

mike-000:
It will need changing to `texType != gl.UNSIGNED_BYTE || data instanceof Uint8Array ? data : new Uint8Array(data.buffer)`

It might be easier to add that to #12761.

tschaub:
If the problem is just with using `Uint8ClampedArray` in the data tile example, I'd rather see us change that example (to convert to `Uint8Array`).

mike-000:
That is how I initially checked to confirm the problem, but the function could potentially be called from anywhere.  If there is no check for mis-matched gl and data types made in `ol/webgl/TileTexture.js` the typedef imported from `ol/DataTile.js` should not include `Uint8ClampedArray` and `DataView` and the calling code should ensure only `Uint8Array` and `Float32Array` are passed.

tschaub:
[We document](https://openlayers.org/en/latest/apidoc/module-ol_DataTile.html#~Data) that the `DataTile` expects `Uint8Array`, `Uint8ClampedArray`, `DataView`, or `Float32Array` (as of #12761).  This is type checked for people that use `tsc`.  It is true that users could provide other garbage and that things would fail.

This data is ultimately passed to the [`gl.texImage2D()` function](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D) that accepts [`ArrayBufferView` type](https://developer.mozilla.org/en-US/docs/Web/API/ArrayBufferView) data.  As you pointed out, Chrome didn't support `Uint8ClampedArray` until version 76 (when https://bugs.chromium.org/p/chromium/issues/detail?id=961658 was addressed).

I gather from what you are saying that IE 11 also expects 8-bit data to be `Uint8Array` and not `Uint8ClampedArray`.  I don't think it makes sense to have a workaround in the library for IE 11 on this.  I do think it could make sense to update the `data-tile.js` example so it works in IE 11.  And if we want to add additional documentation to clarify what data types the `DataTile` loader expects, that could also make sense.

mike-000:
If the only browser affected was IE this would only affect the Data Tiles example as there is no way GeoTIFF can work without browser support for Symbol and Promise used by the workers.  

Chromium based browsers should support GeoTIFF from version 38 (subject to a main thread polyfill for `Promise.allSettled` being needed up to version 76).  With `Uint8ClampedArray` converted to `Uint8Array` I can get the COG example working in Chrome 71 on an old Android tablet - but without that conversion there is simply a black screen similar to current Data Tiles example on IE.

tschaub:
Thanks for the additional detail, @mike-000.  I opened #12770 with a proposed fix.

PR #12770, opened 2021-09-16
Use Uint8Array instead of Uint8ClampedArray for increased browser compatibility

tschaub:
As reported by @mike-000 in #12766, IE 11 and some Chromium browsers do not accept `Uint8ClampedArray` in `gl.texImage2D()`.  [The issue](https://bugs.chromium.org/p/chromium/issues/detail?id=961658) has been fixed in Chromium, but we can use `Uint8Array` instead where we create 8-bit arrays that get passed as texture data for more compatibility.

Closes #12766.


ahocevar (approved):

mike-000:
The Data Tiles example is now working on IE but I am unable to test the GeoTIFF example with Chrome 71 due to #12771

PR #12773, opened 2021-09-17
Enable WebGL extensions when required

### LINKED ISSUE #12771: Excessive use of assertion 63
**Describe the bug**
Since #12761 no WebGL examples (e.g. Heatmap) are working on two Android tablets.  I cannot debug on the devices but I presume the examples are throwing assertion 63 due to lack of device support for either `OES_texture_float` or `OES_texture_float_linear` even through these are not used by the application.  (A script to display the supported extensions confirms `OES_texture_float_linear` is not supported in both cases).

**To Reproduce**
Try https://deploy-preview-12761--ol-site.netlify.app/examples/heatmap-earthquakes.html on various devices.

**Expected behavior**
Separate assertions for each extension and thrown only if the application must use it.

tschaub:
This branch makes it so we don't try to enable WebGL extensions (with `gl.getExtension()`) until we actually need it.  The returned extensions are cached on the WebGL helper for reuse.

Fixes #12771.

### DIFF
--- a/doc/errors/index.md
+++ b/doc/errors/index.md
@@ -233,10 +233,6 @@ Cannot determine IIIF Image API version from provided image information JSON.
 
 A `WebGLArrayBuffer` must either be of type `ELEMENT_ARRAY_BUFFER` or `ARRAY_BUFFER`.
 
-### 63
-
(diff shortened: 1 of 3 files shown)


mike-000:
That is a nicer solution.  I tried it on a device where everything except `OES_texture_float_linear` is supported and the only example not working is NumptyTiles, as expected.

ahocevar (approved):

PR #12774, opened 2021-09-17
Fix text decluttering when segment length is zero

### LINKED ISSUE #12752: Decluttering is broken for specific Text style options
**Describe the bug**
When using a `Text` style with `placement: "line"`, `overflow: true`, and `textAlign: "center"`, the labels are not decluttered above a certain zoom level.

**To Reproduce**
1. Go to https://codesandbox.io/s/declutter-text-labels-5gd9f?file=/index.js 
2. Zoom map out 
3. Observe decluttering not working

https://user-images.githubusercontent.com/461416/133225186-1ea4562b-1001-4d56-b218-c923edff3d21.mov

4. Comment out any of the `placement`, `overflow`, or `textAlign` options and see the expected behaviour.

**Expected behavior**
The Text labels to be properly decluttered.

MoonE:
Fixes #12752

When the segment length is zero there is a division by zero, which ultimately results in NaN values being inserted in the declutter tree. This causes all checks for overlapping labels to fail.

In the example in #12752 the second LineString of the first MultiLineString has the same start and end coordinates. At a low enough zoom level this gets simplified to just the start and end coordinates, thus resulting in a zero length LineString.

### DIFF
--- a/src/ol/geom/flat/textpath.js
+++ b/src/ol/geom/flat/textpath.js
@@ -54,7 +54,8 @@ export function drawTextOnPath(
     advance();
   } while (offset < end - stride && segmentM + segmentLength < startM);
 
-  let interpolate = (startM - segmentM) / segmentLength;
+  let interpolate =
+    segmentLength === 0 ? 0 : (startM - segmentM) / segmentLength;
   const beginX = lerp(x1, x2, interpolate);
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
Nice work, @MoonE

PR #12778, opened 2021-09-18
Do not render vector tiles unless they are being used

ahocevar:
This pull request removes code that causes quite a bit of overhead in the vector tile renderer: rendering a tile immediately after loading bypasses our logic that makes sure we only render tiles that are actually being used (`frameState.usedTiles`). That code originally was meant for better making use of a render frame, but in practice, it causes more work because it renders tiles during/after zoom animations that are not needed for rendering the target resolution.

This pull request is best reviewed [without looking at whitespace changes](https://github.com/openlayers/openlayers/pull/12778/files?diff=unified&w=1).

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -4,7 +4,6 @@
 import CanvasBuilderGroup from '../../render/canvas/BuilderGroup.js';
 import CanvasExecutorGroup from '../../render/canvas/ExecutorGroup.js';
 import CanvasTileLayerRenderer from './TileLayer.js';
-import EventType from '../../events/EventType.js';
 import ReplayType from '../../render/canvas/BuilderType.js';
 import TileState from '../../TileState.js';
 import VectorTileRenderType from '../../layer/VectorTileRenderType.js';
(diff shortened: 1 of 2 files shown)


src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -173,29 +159,16 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
     const projection = viewState.projection;
     const layer = this.getLayer();
     const tile = layer.getSource().getTile(z, x, y, pixelRatio, projection);
-    if (tile.getState() < TileState.LOADED) {
+    const viewHints = frameState.viewHints;
+    const hifi = !(
+      viewHints[ViewHint.ANIMATING] || viewHints[ViewHint.INTERACTING]
+    );
+    if (hifi || !tile.wantedResolution) {
       tile.wantedResolution = resolution;
```
tschaub, line 177: Do we not need to know `tile.wantedResolution` any more?

tschaub (approved):

src/ol/renderer/canvas/VectorTileLayer.js
ahocevar: We do need to know it, and we still set it in `getTile()`.

PR #12787, opened 2021-09-20
fix(KML): add support for reading <Camera> tags in KML format

### LINKED ISSUE #12786: Support <Camera> tags in KML
Currently [ol/format/KML](https://github.com/openlayers/openlayers/blob/v6.7.0/src/ol/format/KML.js) has no methods for reading View properties (rotation, center, etc.) from KML. 
The KML specification for map views is the [Camera](https://developers.google.com/kml/documentation/cameras) tag.

The methods for reading <Region> were copied and adapted for Camera tags. readCamera returns an array containing objects for each <Camera> in the file.

danji90:
Fixes #12786

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -163,6 +163,21 @@ const LINK_PARSERS = makeStructureNS(NAMESPACE_URIS, {
  * @type {Object<string, Object<string, import("../xml.js").Parser>>}
  */
 // @ts-ignore
+const CAMERA_PARSERS = makeStructureNS(NAMESPACE_URIS, {
+  Altitude: makeObjectPropertySetter(readDecimal),
+  Longitude: makeObjectPropertySetter(readDecimal),
+  Latitude: makeObjectPropertySetter(readDecimal),
+  Tilt: makeObjectPropertySetter(readDecimal),
+  AltitudeMode: makeObjectPropertySetter(readString),
+  Heading: makeObjectPropertySetter(readDecimal),
+  Roll: makeObjectPropertySetter(readDecimal),
+});
+
+/**
+ * @const
+ * @type {Object<string, Object<string, import("../xml.js").Parser>>}
+ */
+// @ts-ignore
 const REGION_PARSERS = makeStructureNS(NAMESPACE_URIS, {
   'LatLonAltBox': latLonAltBoxParser,
   'Lod': lodParser,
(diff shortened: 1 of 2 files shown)


oterral:
@ahocevar   this stale bot is a bit painful :D   It closes the PR and closes the ticket, but the bug is still there. So it makes no chance to have this merged. Any chance to improve this ?

ahocevar:
@oterral If the bot adds a comment that the issue is stale, just add a comment that you're still interested in (solving) the issue.

src/ol/format/KML.js
```diff
@@ -888,6 +903,74 @@ class KML extends XMLFeature {
...
+  /**
+   * @param {Element} node Node.
+   * @return {Array<Object>} Cameras.
+   * @api
+   */
+  readCameraFromNode(node) {
+    const cameras = [];
+    for (let n = node.firstElementChild; n; n = n.nextElementSibling) {
+      if (
+        includes(NAMESPACE_URIS, n.namespaceURI) &&
```
MoonE, line 952:
```suggestion
        NAMESPACE_URIS.includes(n.namespaceURI) &&
```

src/ol/format/KML.js
```diff
@@ -888,6 +903,74 @@ class KML extends XMLFeature {
...
+        n.localName === 'Camera'
+      ) {
+        const obj = pushParseAndPop({}, CAMERA_PARSERS, n, []);
+        cameras.push(obj);
+      }
+    }
+    for (let n = node.firstElementChild; n; n = n.nextElementSibling) {
+      const localName = n.localName;
+      if (
+        includes(NAMESPACE_URIS, n.namespaceURI) &&
```
MoonE, line 962:
```suggestion
        NAMESPACE_URIS.includes(n.namespaceURI) &&
```

MoonE (commented):
The `inlcudes` function has been removed when  support for IE11 was dropped. You can use the `includes` function of the Array prototype now.

src/ol/format/KML.js
danji90: Thanks, for the hint. My fork was outdated, I synced the repo and updated the methods.

danji90:
> Thanks for this contribution, @danji90. Would you be able to address the problems found by the GitHub action?

The issues have been fixed

danji90:
> Thanks, @danji90. The code looks good now. As a bonus, it would be great if you could squash the commits. Also, to make sure that what you added does not break in the future, you should consider adding a test.
> 
> Let me know if you're unable to do this now, then we can also merge it as-is.

Yeah good call, I will add some tests

danji90:
@ahocevar A test for the KML reader has been added and the commits were squashed. I think we are done here.

src/ol/format/KML.js
```diff
@@ -879,6 +894,71 @@ class KML extends XMLFeature {
     return regions;
   }
 
+  /**
+   * Read the cameras of the KML.
+   *
+   * @param {Document|Element|string} source Source.
+   * @return {Array<Object>} Cameras.
```
ahocevar, line 901: Since this is part of the official API, I think the type should be more specific. Would you be able to add a `KMLCamera` typedef and use `{Array<KMLCamera>}` as return type?

ahocevar (changes requested):

src/ol/format/KML.js
danji90: Done. I guess this should be done for other KML methods as well (e.g. region), but I do not have the time to spare atm

PR #12790, opened 2021-09-21
Allow maps to be configured with a promise for view props

### LINKED ISSUE #12644: Auto view
I'm wondering if it would be a useful feature to be able to construct a map without a view and to have the view auto-generated based on layer or source information.  For example, a GeoTIFF source may provide metadata on view-related properties like projection and extent.

To create a map that would show an entire GeoTIFF, we could make it possible to do this:

```js
const layer = new TileLayer({
  source: new GeoTIFF({
    sources: [{url: 'https://example.com/cog-1.tif'}],
  }),
});

// map with no view will get one when information is available
const map = new Map({
  target: 'map',
  layers: [layer],
});
```

Then to later switch the view to render a different source, the application could do this:

```js
map.setView(null);

// this will eventually provide the map with a new view
layer.setSource(
  new GeoTIFF({
    sources: [{url: 'https://example.com/cog-2.tif'}],
  })
);
```

Before discussing details of the implementation, it would be good to decide whether this would be a worthwhile feature.  I can imagine it would be useful in situations where someone wants to build an application that lets users browse through a catalog of imagery (that would be rendered as a single layer on a map).

In terms of the implementation, I can imagine making it so that a map without a view listens (once) for a new "viewready" type event.  In the example above, this event would be fired by the source and relayed through the layer to the map.

I'm curious to hear others' thoughts.

tschaub:
This allows maps to be configured with a promise for view properties.  Where appropriate, sources can resolve view-related properties.  The `source.getView()` method is currently only marked as part of the API in the GeoTIFF source - where we can get relevant projection, center, and resolution information from the source.

Fixes #12644.

### DIFF
--- a/examples/cog-math-multisource.js
+++ b/examples/cog-math-multisource.js
@@ -1,7 +1,6 @@
 import GeoTIFF from '../src/ol/source/GeoTIFF.js';
 import Map from '../src/ol/Map.js';
 import TileLayer from '../src/ol/layer/WebGLTile.js';
-import View from '../src/ol/View.js';
 
 const source = new GeoTIFF({
   sources: [
@@ -57,9 +56,5 @@ const map = new Map({
       source,
     }),
   ],
-  view: new View({
-    center: [1900000, 6100000],
-    zoom: 13,
-    minZoom: 10,
-  }),
+  view: source.getView(),
 });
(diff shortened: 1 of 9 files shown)


ahocevar (approved):
Did you mean to commit the changes to the `simple` example? It is broken with these changes, so I think you'll want to remove the changes there.

Other than that, this looks great! Please merge after fixing the `simple` example.

PR #12793, opened 2021-09-21
Add a method to set the map layers

tschaub:
This adds a `map.setLayers()` method that lets people set map layers.  Any existing layers are cleared when setting new layers.


ahocevar (approved):
Oh yes, this adds a lot of convenience. Thanks!

PR #12798, opened 2021-09-23
More cleanup in the WebGL tile layer's dispose method

tschaub:
This branch makes it so for all layers, the `layer.dispose()` method calls `renderer.dispose()` – this wasn't happening before.  In addition, for the WebGL tile layer renderer, we now delete underlying WebGL resources in the renderer's dispose method.  Most significantly, we call `gl.deleteTexture()` for all of the textures in the renderer's tile texture cache.  The garbage collector should eventually get to cleaning up these resources itself, but these additional calls should make that happen sooner.

### DIFF
--- a/src/ol/layer/Layer.js
+++ b/src/ol/layer/Layer.js
@@ -348,6 +348,11 @@ class Layer extends BaseLayer {
    * Clean up.
    */
   disposeInternal() {
+    if (this.renderer_) {
+      this.renderer_.dispose();
+      delete this.renderer_;
+    }
+
     this.setSource(null);
     super.disposeInternal();
   }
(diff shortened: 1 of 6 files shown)


jahow (approved):
Thanks! Note that this was done in a non-generic way here:
https://github.com/openlayers/openlayers/blob/2ebbee23407f8af91e0a1d24523e0e269c02c83b/src/ol/layer/WebGLPoints.js#L121

So this line is probably obsolete now.

tschaub:
> So this line is probably obsolete now.

I've removed that from the WebGL points layer in the latest commit.

I also decided it might be worth trying the mysterious ["WEBGL_lose_context" extension](https://www.khronos.org/registry/webgl/extensions/WEBGL_lose_context/).  It sounds to me like this extension only simulates a context being lost, but it sounds like it is up to implementors to decide if it should actually free up resources (see also the [WebGL Public mailing list](https://www.khronos.org/webgl/public-mailing-list/public_webgl/1611/msg00013.php)).

PR #12802, opened 2021-09-23
Enqueue tiles at the next step in the animation

tschaub:
This change makes it so the WebGL tile layer renderer enqueues tiles at the next step in an animation series.  In cases where an animation covers several zoom levels or is outside the initial extent, this makes it so destination tiles have a chance to start loading before the animation finishes.


ahocevar (approved):
This looks excellent, thanks @tschaub!

PR #12803, opened 2021-09-23
Add support for arbitrary GLSL functions

theduckylittle:
Add support for calling a function in the GLSL pipeline.

 * Adds the `func` operator to call the function in the style.
 * Adds a `functions` list to append function text to the fragment shader.
 * Includes an example which allows the user to blur the data.

### DIFF
--- a/examples/webgl-custom-func.html
+++ b/examples/webgl-custom-func.html
@@ -1,0 +1,17 @@
+---
+layout: example.html
+title: Add a custom function to the fragment shader
+shortdesc: Renders a DataTile layer with a custom WebGL function.
+docs: >
+  This example uses a <code>ol/source/DataTile</code> source to load multi-byte raster data in the
+  <a href="https://github.com/planetlabs/numpytiles-spec/">NumpyTile</a> format.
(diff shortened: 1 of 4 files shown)


tschaub:
This is a cool idea, @theduckylittle.  Since the WebGL tile layer rendering implementation is still so new, we want to avoid getting locked in to too many of the specifics right now.  If we have an API that allows people to work with the names of the uniforms, varyings, etc that the library uses, it means we break their applications (and API compatibility) if we decide to rework parts of the implementation.  I don't think we want to commit ourselves to every detail of the current implementation yet.  So I don't think we'll take this contribution as part of the supported API.

It might also be worthwhile to talk a bit more about what specific functionality you'd like to see supported.  If this were just about doing things like blurring, then I think we could find a solution that didn't involve hand-authored GLSL (e.g. styles could accept a filter or kernel array).

theduckylittle:
Totally understood! I'm open minded to an overall solution. This just seemed like a useful shortcut for those that were comfortable writing bespoke shader code. Passing in a kernel to be applied may, ultimately, be a better solution. Will close for now.

PR #12805, opened 2021-09-24
Do not replace icon color if image not loaded

mike-000:
Attempting to set an icon color before the image has loaded, as can happen with the immediate renderer https://codesandbox.io/s/icon-color-forked-sgbvf creates an invalid zero size canvas.


mike-000:
Closed due to incorrect rebase

PR #12807, opened 2021-09-24
Updated styles

tschaub:
This branch makes minor adjustments to the style applied to our controls.

style

The box rendered by the drag zoom interaction now has a fill and the stroke is no longer `blue`.  The line-height for all controls is controlled by other styles on the page - we no longer bump it up to `1.14em`.  Instead of overriding the anchor style of attributions just for our examples, we now do it it in `ol.css` (to match the style of other control style).  Flex items in the attribution control are now centered (perpendicular to the flex direction) instead of being aligned to the flex end.  Collapsible controls now use ‹ and › as collapse/expand labels instead of « and ».

### DIFF
--- a/examples/resources/layout.css
+++ b/examples/resources/layout.css
@@ -122,12 +122,14 @@ ol.inline>li {
   background: url(map-background.jpg) repeat;
   margin-bottom: 10px;
 }
+
+.ol-control {
+  line-height: normal;
+}
+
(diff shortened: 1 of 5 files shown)


ahocevar (approved):
Nice, I think everyone embedding an OpenLayers map in their webpage with customized styles will appreciate this simplification!

PR #12830, opened 2021-09-27
bugfix: Fix type of `layers` option in `OverviewMap`

ejn:
* The foundation `Layer` class is `BaseLayer` and not `Layer`, it is also possible to use e.g. a `LayerGroup` which does not inherit from `Layer`


ahocevar (approved):
Thanks, @ejn

PR #12831, opened 2021-09-27
Fix rendering VectorImage with no features in view extent

### LINKED ISSUE #12816: VectorImage layer does not remove features
**Describe the bug**
A vector feature (Circle or Polygon) is still rendered in map when using VectorImage layer after the feature was removed.

**To Reproduce**
1. Go to https://jsfiddle.net/z87pejso/
2. Click on "Click to remove circle" button
3. See that the circle is still rendered in the map + pan or zoom the map and the feature changes its position.

**Expected behavior**
The feature should be gone.

MoonE:
Fixes #12816

When no features are to be rendered the previous image was kept around, but it should have been removed.


ahocevar (approved):
Thanks, @MoonE!

PR #12835, opened 2021-09-29
updated mapbox-layer example

FreeYeti:
added mapbox-gl canvas to an ol-layer container. To solve the problem during adding another vector layer.

### DIFF
--- a/examples/mapbox-layer.js
+++ b/examples/mapbox-layer.js
@@ -56,7 +56,12 @@ const mbLayer = new Layer({
     }
     mbMap._render();
 
-    return canvas;
+    const div = document.createElement("div");
+    div.className = "ol-layer"
+    div.style = "position: absolute; width: 100%; height: 100%;"
+    div.appendChild(canvas);
+
+    return div;
   },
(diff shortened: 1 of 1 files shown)


MoonE:
Thank you for contributing.
Could you explain what problem this commit solves? The example seems to work just the same with or without this commit.

FreeYeti:
In our React application, I add a mapbox-gl layer when the component did mount. Then I try to add a vector layer with geojson source or WKT source by an onClick event, but this layer can't display. I checked elements and finally realized that the cause is my mapbox-gl layer doesn't have a layer's container, so when I try to add another layer, the newer one will be under the previous mapbox-gl layer.

FreeYeti (commented):
replace quotes

mike-000:
A container should not be needed.  OpenLayers own WebGL based layers do not have containers, for example the Heatmap layer in https://openlayers.org/en/latest/examples/heatmap-earthquakes.html  The base layer 2d canvas is in a container while the heatmap layer WebGL canvas has no container

image

PR #12836, opened 2021-09-29
Support a normalize option on the GeoTIFF source

### LINKED ISSUE #12834: GeoTiff style avoiding normalization.
The style of the GeoTiff need values normalized betwen [0,1] (Is It true?).
for example: https://openlayers.org/en/latest/examples/cog-math.html

GeoTiff files usually have values of any kind and in this way the developer has to translate the legend in a (normalized) style that sometime can also be unaccurate.  

It will be possible to avoid the normalization?

so in order to define the style like a silly legend:

```
const source = new GeoTIFF({
                sources: [
                    {
                        url: "dem.tif",  //digital elevation model 
                        nodata: -9999,
                        min: 0,
                        max: 8848.86    //Everest
                    },
                ],
})
```

```
const dem = new TileLayer({
                    source: source,
                    style: {
                      color: [
                        'interpolate',
                        ['linear'],
                        ['band', 1],
                        0,[0,0,0,0],
                        10,[0,0,64],
                        100,[0,0,128],
                        1000,[0,0,255],
                        /*....*/
                        8848.86,[255,255,255],
                       ],
                    },
                }),
```

tschaub:
By default, we normalize values in GeoTIFF sources so that style expressions on the layer work with values between 0 and 1.  This change makes it so the GeoTIFF source accepts a `normalize` option to control this behavior.  If `normalize` is set to `false`, then the raw GeoTIFF values will be make available to style expressions on the layer.

The current implementation uses 8-bit values for normalized data and 32-bit otherwise.  In the future, we may change this to support higher bit depth for normalized values as well.  But this implementation detail is not exposed to users.

I haven't had any luck finding a CORS-friendly source of elevation data hosted as GeoTIFF.  Both the AWS [Terrain Tiles](https://registry.opendata.aws/terrain-tiles/) and [Copernicus DEM data](https://registry.opendata.aws/copernicus-dem/) are not CORS-friendly.  But here is a movie of a elevation tile from the Terrain Tiles with a CORS proxy in front:

elevation

Fixes #12834.

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -273,6 +273,10 @@ function getMaxForDataType(array) {
  * @property {boolean} [convertToRGB = false] By default, bands from the sources are read as-is. When
  * reading GeoTIFFs with the purpose of displaying them as RGB images, setting this to `true` will
  * convert other color spaces (YCbCr, CMYK) to RGB.
+ * @property {boolean} [normalize=true] By default, the source data is normalized to values between
+ * 0 and 1 with scaling factors based on the `min` and `max` properties of each source.  If instead
+ * you want to work with the raw values in a style expression, set this to `false`.  Setting this option
+ * to `false` will make it so any `min` and `max` properties on sources are ignored.
  * @property {boolean} [opaque=false] Whether the layer is opaque.
  * @property {number} [transition=250] Duration of the opacity transition for rendering.
  * To disable the opacity transition, pass `transition: 0`.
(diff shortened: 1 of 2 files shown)


ahocevar (approved):

mike-000:
AWS Terrain Tiles are CORS friendly as long as you use `https://elevation-tiles-prod.s3.amazonaws.com/ `(`https://s3.amazonaws.com/elevation-tiles-prod/ `is not CORS enabled)

https://codesandbox.io/s/webgl-tiles-forked-yx38uh?file=/main.js

Issue #12833, opened 2021-09-28
forEachLayerAtPixel gives error for WebGLTile Layer


m-mohr:
**Describe the bug**

I receive the following error when implementing a singleclick event for a map with a WebGLTile layer + GeoTiff source:

```
util.js:9 Uncaught Error: Unimplemented abstract method.
    at util.js:9
    at abstract (util.js:10)
    at WebGLTileLayerRenderer.push../node_modules/ol/renderer/Layer.js.LayerRenderer.getDataAtPixel (Layer.js:133)
    at CompositeMapRenderer.push../node_modules/ol/renderer/Composite.js.CompositeMapRenderer.forEachLayerAtPixel (Composite.js:167)
    at Map.push../node_modules/ol/PluggableMap.js.PluggableMap.forEachLayerAtPixel (PluggableMap.js:560)
    at Map.<anonymous> (MapViewer.vue:179)
    at Map.push../node_modules/ol/events/Target.js.Target.dispatchEvent (Target.js:116)
    at Map.push../node_modules/ol/PluggableMap.js.PluggableMap.handleMapBrowserEvent (PluggableMap.js:897)
    at MapBrowserEventHandler.push../node_modules/ol/events/Target.js.Target.dispatchEvent (Target.js:116)
    at MapBrowserEventHandler.<anonymous> (MapBrowserEventHandler.js:129)
```

**To Reproduce**

Can't provide a codesandbox right now due to the txml issue, but use the COG basic example and add something like:

```
			map.on('singleclick', evt => {
				const pixel = this.map.getEventPixel(evt.originalEvent);
				map.forEachLayerAtPixel(pixel, layer => {
					console.log(pixel);
					console.log(layer);
				});
			});
```

OpenLayers version 6.8.1.

**Expected behavior**
Be able to get pixel values.

mike-000:
This currently affects all WebGL layers including Heatmap.  To return an RGBA array in the callback similar to 2d canvas layers would need the WebGL canvas to be created with the `preserveDrawingBuffer: true` option (which is also required to use WebGL layers in map exports, see #11014).  That is inefficient and should be an option.  As the docs say "will be null for layer types that do not currently support this argument" I think unsupported layers should be ignored, and not throw an error.

m-mohr:
Thanks for the details. I agree that this should not throw an error.

But that means that there's no way to get the pixel value (e.g. on a click) for a GeoTiff source right now?

mike-000:
[...]

PR #12837, opened 2021-09-30
forEachLayerAtPixel return null for unsupported layer types

mike-000:
Fixes the error in #12833

Ensure the getDataAtPixel return is null unless the layer is supported as the documentation states "will be null for layer types that do not currently support this argument" instead of throwing an error.

WebGL layers could be supported in the same way as 2d canvas layers if the context is created `preserveDrawingBuffer: true`  That would be best done via an option in the Heatmap, WebGLTile and WebGLPoints layer constructors instead of patching the HTMLCanvasElement prototype.  Since it is also necessary to export maps, both by html2canvas and the Export Map method, adding that is beyond the scope of this fix.

### DIFF
--- a/src/ol/renderer/Layer.js
+++ b/src/ol/renderer/Layer.js
@@ -130,7 +130,7 @@ class LayerRenderer extends Observable {
    *    returned, and empty array will be returned.
    */
   getDataAtPixel(pixel, frameState, hitTolerance) {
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
Thanks, @mike-000

Issue #12842, opened 2021-10-01
Reading very large TopoJSON files results in RangeError: Maximum call stack size exceeded


karimbahgat:
**Describe the bug**
When loading very large TopoJSON files, the TopoJSON format fails on `readFeatures(data)` with the following error:

```
Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at YS (TopoJSON.js:182)
    at MultiPolygon (TopoJSON.js:276)
    at qS (TopoJSON.js:325)
    at WS (TopoJSON.js:300)
    at e.readFeaturesFromObject (TopoJSON.js:121)
    at e.readFeatures (JSONFeature.js:61)
    at loadFromTopoJSON (internalData.js:193)
    at internalData.js:278
```

After some debugging, it was clear this was happening specifically for very large files topojson files and specifically when processing very long arc coordinate sequences (arcs containing somewhere between 100,000 to 200,000 coordinates). The error happens in the `concatenateArcs` function at https://github.com/openlayers/openlayers/blob/78a4efab8cf53515ff44724551b3f0843e204dec/src/ol/format/TopoJSON.js#L203

Further research reveals that this happens when concatenating two arrays using `push.apply` in cases where the other array contains too many elements: https://stackoverflow.com/questions/61740599/rangeerror-maximum-call-stack-size-exceeded-with-array-push. 

The error can be fixed simply by using a for-loop to push each coordinate individually instead of using `push.apply`. I don't see any unintended consequences of this, and according to [this benchmark](https://www.measurethat.net/Benchmarks/Show/7699/0/array-concat-vs-spread-vs-push-spread-loop-apply) the for-loop is going to be just as fast as `push.apply`. The TopoJSON format reader appears to be alone in using `push.apply`, the GeoJSON format reader uses normal for-loops when extract/adding coordinates. 

Submitting a PR for this shortly. 

I know that most TopoJSON's will likely not contain arcs that long, but for noncontiguous geometry collections with only a few or no shared boundary segments, entire polygon rings often get stored in a single arc (rather than getting chopped into multiple shared arc segments). Large and complex geometries are therefore more likely to result in the maximum call stack error. 

**To Reproduce**
The equivalent of the following code is what produced the error:

```
var data = JSON.parse('read some very large topojson string')
var format = new ol.format.TopoJSON({});
var features = format.readFeatures(topoj, {
                                                dataProjection: 'EPSG:4326',
                                                featureProjection: 'EPSG:3857'
                                            }
[...]

PR #12843, opened 2021-10-01
Fix maximum call stack error when parsing very long arc arrays

karimbahgat:
This fixes a bug caused when concatenating arc segments using `push.apply()` for arcs that are very large (100,000-200,000 points), see eg: https://stackoverflow.com/questions/61740599/rangeerror-maximum-call-stack-size-exceeded-with-array-push

See #12842 for more details. 

I have only tested that the fix works by copying the relevant functions to a separate js script, not within the context of the entire OpenLayers library.

### DIFF
--- a/src/ol/format/TopoJSON.js
+++ b/src/ol/format/TopoJSON.js
@@ -200,7 +200,9 @@ function concatenateArcs(indices, arcs) {
       // reverse arc
       arc = arcs[~index].slice().reverse();
     }
-    coordinates.push.apply(coordinates, arc);
+    for (p of arc) {
+      coordinates.push(p);
+    }
   }
   // provide fresh copies of coordinate arrays
   for (let j = 0, jj = coordinates.length; j < jj; ++j) {
--- a/src/ol/format/TopoJSON.js
+++ b/src/ol/format/TopoJSON.js
@@ -186,7 +186,7 @@ const GEOMETRY_READERS = {
 function concatenateArcs(indices, arcs) {
   /** @type {Array<import("../coordinate.js").Coordinate>} */
   const coordinates = [];
-  let index, arc;
+  let index, arc, vertex;
   for (let i = 0, ii = indices.length; i < ii; ++i) {
     index = indices[i];
     if (i > 0) {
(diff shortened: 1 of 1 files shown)


karimbahgat:
Any takers to help merge this? This would help out a great deal for those working with TopoJSON. Much appreciated!

mike-000:
A regular for `loop` is more efficient than `for ... of` in most browsers https://jsfiddle.net/qwofaujr/

src/ol/format/TopoJSON.js
```diff
@@ -200,7 +200,9 @@ function concatenateArcs(indices, arcs) {
       // reverse arc
       arc = arcs[~index].slice().reverse();
```
ahocevar, line 201:
```suggestion
      arc = arcs[~index].slice();
```

src/ol/format/TopoJSON.js
```diff
@@ -200,7 +200,9 @@ function concatenateArcs(indices, arcs) {
...
     }
-    coordinates.push.apply(coordinates, arc);
+    for (vertex of arc) {
+      coordinates.push(vertex);
+    }
```
ahocevar, line 205:
```suggestion
    for (let a = arc.length - 1; a >= 0; --a) {
      coordinates.push(arc[a]);
    }
```

ahocevar (changes requested):
@karimbahgat I added two suggestions. Could you commit them please?

src/ol/format/TopoJSON.js
mike-000: Avoiding `reverse()` makes sense but won't you now need separate forward and reverse `for` loops?  Isn't the `slice()` also redundant?

ahocevar: Yes, hard to see that in the diff view.

karimbahgat:
Thanks for pointing this out @mike-000. Would be glad to merge your suggestions @ahocevar, but see you already created a new PR and merged the fix for this into main in https://github.com/openlayers/openlayers/pull/13165. Thanks for your quick response both, my app will be much happier come next release :)

PR #12845, opened 2021-10-01
Preserve drawing buffer

mike-000:
Adds a `preserveDrawingBuffer` option to the Heatmap, WebGLPoints and WebGLTile layer options to make the canvases exportable (and compatible with a future PR to support `forEachLayerAtPixel`).  Add a simple Heatmap layer to the Export Map example.

### DIFF
--- a/examples/export-map.js
+++ b/examples/export-map.js
@@ -1,8 +1,12 @@
 import GeoJSON from '../src/ol/format/GeoJSON.js';
 import Map from '../src/ol/Map.js';
 import View from '../src/ol/View.js';
+import {
+  Heatmap as HeatmapLayer,
+  Tile as TileLayer,
+  Vector as VectorLayer,
+} from '../src/ol/layer.js';
 import {OSM, Vector as VectorSource} from '../src/ol/source.js';
-import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
 
 const map = new Map({
   layers: [
@@ -15,6 +19,19 @@ const map = new Map({
         format: new GeoJSON(),
       }),
       opacity: 0.5,
+    }),
+    new HeatmapLayer({
+      source: new VectorSource({
+        url: 'data/geojson/world-cities.geojson',
+        format: new GeoJSON(),
+      }),
+      weight: function (feature) {
+        return feature.get('population') / 1e7;
+      },
+      radius: 15,
+      blur: 15,
+      opacity: 0.5,
+      preserveDrawingBuffer: true,
     }),
   ],
   target: 'map',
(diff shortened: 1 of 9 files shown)


examples/export-map.js
```diff
@@ -16,6 +22,18 @@ const map = new Map({
...
+        features: [
+          new Feature(new Point([0, 1000])),
+          new Feature(new Point([-1000, 0])),
+          new Feature(new Point([1000, 0])),
+          new Feature(new Point([0, -1000])),
+        ],
+      }),
+      opacity: 0.5,
+      preserveDrawingBuffer: true,
+    }),
```
tschaub, line 36: We should avoid mixing tests and examples.  This change is making the example feel more like a test (potentially confusing to users looking to learn how to export a map).

mike-000:
Lost contexts seem to happen when multiple tabs are open, which is beyond the application's control, but sharing might help.  Also in intensive applications such as the example for #12803 with hardware acceleration on (it can hang the browser without hardware acceleration).  Those may be cases where performance is most impacted by using `preserveDrawingBuffer`.

I kept the original option name in the same way as the original `crossOrigin` name is used for raster Tile and Image sources, with a similar explanation of its purpose.  It is also consistent with the option name used by Mapbox https://docs.mapbox.com/mapbox-gl-js/api/map/

If exporting I presume an application would want to use `preserveDrawingBuffer` for all WebGL layers.  If they were different we could either take a "one tainted all tainted" approach as in 2d and share regardless, or not share contexts if the options are different.  With context sharing `forEachLayerAtPixel` would have similar issues to layers sharing 2d canvas, so a different `className` would be needed to prevent sharing.

src/ol/renderer/webgl/Layer.js
```diff
@@ -40,6 +40,7 @@ export const WebGLWorkerMessageType = {
  * @property {string} [className='ol-layer'] A CSS class name to set to the canvas element.
  * @property {Object<string,import("../../webgl/Helper").UniformValue>} [uniforms] Uniform definitions for the post process steps
  * @property {Array<PostProcessesOptions>} [postProcesses] Post-processes definitions
+ * @property {Object} [contextAttributes] WebGL context attributes.
```
fredj, line 43: This can be typed as `{WebGLContextAttributes}` instead of `{Object}`

jahow:
Setting `preserveDrawingBuffer: true` by default for all contexts would solve many issues:
* opening the possibility for `forEachLayerAtPixel` to work with webgl
* allowing reading/exporting the canvas contents
* no headache when/if contexts are to be shared at some point

A well documented opt-in performance improvement could then be made available e.g. like this:
```js
import { disablePreserveDrawingBuffer } from 'ol/webgl';

disablePreserveDrawingBuffer();

// then initialize the layers
```

This way the impact on the API would be minimal. But of course assessing whether the performance hit from `preserveDrawingBuffer: true` is relevant or not is a priority, otherwise it can just be enabled all the time!

mike-000:
Shared 2d canvases have the same problem, but specifying a `className` can be used to control what is shared.  `className` is already applied to WebGL canvases.

jahow:
> My concern is that `forEachLayerAtPixel` would not work with a shared context. If the main benefit of `preserveDrawingBuffer` is to enable `forEachLayerAtPixel`, we have to weigh that against the drawback of not being able to share a context (assuming this would break `forEachLayerAtPixel`).

I think the main benefit is just that any application code can access the layers canvas' without surprises. I have found that it is really not that uncommon for people to do that, either for exporting in pdf/png or any other use case. Although no explicit API is helping in that regard, well, I guess maps are just images after all, so why not use them as such?

This is especially true if webgl renderers become increasingly used. For example, when migrating from a canvas2d tiled layer to a webgl one, an error related to preserveDrawingBuffer might be a serious show stopper. On the other hand we're expecting largely improved performance with webgl, so why not sacrifice a very small part of it to make sure we're not breaking any use cases (i.e. enable `preserveDrawingBuffer` by default)?

I hope that makes sense!

mike-000:
> > My concern is that `forEachLayerAtPixel` would not work with a shared context. If the main benefit of `preserveDrawingBuffer` is to enable `forEachLayerAtPixel`, we have to weigh that against the drawback of not being able to share a context (assuming this would break `forEachLayerAtPixel`).
> 
> I think the main benefit is just that any application code can access the layers canvas' without surprises. I have found that it is really not that uncommon for people to do that, either for exporting in pdf/png or any other use case. Although no explicit API is helping in that regard, well, I guess maps are just images after all, so why not use them as such?
> 
> This is especially true if webgl renderers become increasingly used. For example, when migrating from a canvas2d tiled layer to a webgl one, an error related to preserveDrawingBuffer might be a serious show stopper. On the other hand we're expecting largely improved performance with webgl, so why not sacrifice a very small part of it to make sure we're not breaking any use cases (i.e. enable `preserveDrawingBuffer` by default)?
> 
> I hope that makes sense!

I have now discovered via #13117 a side effect of the very efficient #12965 opacity implementation is that with `preserveDrawingBuffer: true` WebGL layer opacity can only be increased, and not decreased.  It also causes a ghosting effect when panning the map.  Regardless of whether `preserveDrawingBuffer` is set as the default or as an option, or even by overriding the `HTMLCanvasElement` prototype, it will be needed to export a map or to access pixel values, but if using it means reverting to a less efficient canvas reuse strategy that is a strong case for it not being the default or a global setting.  Is there any way efficient opacity handling could be made to work with `preserveDrawingBuffer` enabled?

mike-000:
Replaced by #13126

Issue #12838, opened 2021-09-30
GeoTiff image is pasted over a zero-buffer-tile when the image is smaller than tile size.


valluzzi:
**Describe the bug**
GeoTiff image is pasted over a zero-buffer-tile when the image is smaller than tile size.

Assuming that ZERO is not always NODATA, but most of the time is a valid value, When ( at lower zooms)  the image size is SMALLER the the TILE_SIZE the GeoTiff is stiched over  a ZERO-BUFFER tile.
In this way is impossible to distinguish zero values from nodata also using alpha band (because alpha band is derived from sourceband) 

![image](https://user-images.githubusercontent.com/10461494/135442588-7c9585fb-769e-41c0-beb7-b4bbaf0fb52f.png)

**To Reproduce**
```
const source = new GeoTIFF({
                sources: [
                    {
                        projection: EPSG,
                        //url: process.env.PUBLIC_URL+"/WD.tif", //digital elevation model
                        url: process.env.PUBLIC_URL+"/rimini.tif", //digital elevation model
                        //url: process.env.PUBLIC_URL+"/cologne.tif", //digital elevation model
                        //url: process.env.PUBLIC_URL+"/pamplona.tif", //digital elevation model
                        nodata: -9999,
                        min:0,
                        max:30
                    },
                ],
            })
```

```
const dem =    new TileLayer({
                    source: source,

                    style: {
                      color: [
                        'interpolate',
                        ['linear'],
                        ["case", ['==',['band', 2],1], ['band', 1], -9999],  //alpha band to make nodata transparent
                        -9999,[255,0,0,0],//transparent
                        0,[255,0,0], //red
                        1,[0,0,255],

                      ],
                    },
                }),
```

**Expected behavior**
![image](https://user-images.githubusercontent.com/10461494/135443087-1dc74626-af00-42c8-a0d8-8e70ef443bf3.png)

**Suggestion**

In ol/source/GeoTiff.js

My suggestion is to identify  (w,h) the real size of the image and  turn noimage-pixels into nodata in the pixels loop.

```
var image = this.sourceImagery_[sourceIndex][z];

//(w,h) is the real (useful data) tile size --------------------------------------------------------//
var m = size[1]
var n = size[0]
var w = (image.getWidth()%n  < n && ((x+1)*n)>image.getWidth() )?image.getWidth()%n:n;
var h = (image.getHeight()%m < m && ((y+1)*m)>image.getHeight())?image.getHeight()%m:m;
[...]

PR #12846, opened 2021-10-01
Use nodata values in the GeoTIFF headers for fill value

tschaub:
We currently use the `nodata` value from a source config object when reading rasters with geotiff.js.  If this config property is set, it is passed as the `fillValue` property to `readRasters` - so areas outside the image extent get filled with the user-configured `nodata` value.

We can make it easier to provide the appropriate `fillValue` by using the header-parsed `nodata` values.  This means people shouldn't need to configure a source with `nodata` unless the value is not provided in the image headers.

See #12838 for some more detail.


ahocevar (approved):
Thanks, @tschaub! I think it could make sense to add a note to the `nodata` documentation that the value from the geotiff metadata is used by default.

Issue #12860, opened 2021-10-05
Two suggestions for documentation and examples that would have saved me a lot of time.


debaser75:
First, I hope it's OK for me to post issues with the documentation here. If not, please let me know where I should. 

Suggestion number one: While following the steps on [this page](https://openlayers.org/en/latest/doc/tutorials/bundle.html), I received an error with the message "could not find commit hash for HEAD". After spending a considerable amount of time troubleshooting and Googling, I eventually found this [issue/post](https://github.com/openlayers/create-ol-app/issues/2). Installing git fixed the issue. Here's my brief suggestion for the "Application Setup" section: "If you receive an error message along the lines of 'Could not find commit hash for HEAD', make sure that git is properly installed on your system." 

Suggestion 2: After I got my Node dev environment running correctly, I was encountering CORS errors when attempting to add some ArcGIS data sources.  Eventually, I found a post on StackOverflow that suggested adding `crossOrigin: 'anonymous'`. My updated code is below.
`var usgstopo = new TileLayer({
  source: new XYZ({
    url:
      'https://basemap.nationalmap.gov/arcgis/rest/services/USGSTopo/MapServer/tile/{z}/{y}/{x}',
    crossOrigin: 'anonymous',  
  })
});`

I would suggest adding this [here](https://openlayers.org/en/latest/examples/arcgis-image.html), [here](https://openlayers.org/en/latest/examples/arcgis-image.html), [here](https://openlayers.org/en/latest/examples/xyz-esri-4326-512.html), and [here](https://openlayers.org/en/latest/examples/xyz-esri-4326-512.html). The suggestion could go in the brief text above the code, or as a comment in the code. I think it should also go in the API docs for modules that involve ArcGIS services, such as [here](https://openlayers.org/en/latest/apidoc/module-ol_source_TileArcGISRest-TileArcGISRest.html) and [here](https://openlayers.org/en/latest/apidoc/module-ol_source_ImageArcGISRest-ImageArcGISRest.html). Those two pages do mention the requirement to add a cross origin if accessing the pixel data is necessary. In my case, I had to add it just to **display** the layers in my app. 

Huge thanks to everyone who built OpenLayers into the amazing software that it is today. I wouldn't be able to do some non-profit GIS work if FOSS4G hadn't started thriving.

ahocevar:
[...]

PR #12865, opened 2021-10-06
Add a note about installing git before using create-ol-app

tschaub:
The `create-ol-app` program depends on `git`.  This branch adds a link from our docs to the Git installation docs.

See #12860.


ABCooClay:
use default terminal instead zsh is ok (if you have git on ur computer)

PR #12873, opened 2021-10-08
Add a style convenience function

tschaub:
This adds a new `style` function as a convenience for constructing `Style` objects.

```js
// before
import {Stroke, Fill, Circle, Style} from 'ol/style.js';

const s = new Style({
  fill: new Fill({color: 'orange'}),
  stroke: new Stroke({width: 15, color: 'red'}),
  image: new Circle({
    radius: 42,
    stroke: new Stroke({width: 10, color: 'yellow'}),
    fill: new Fill({color: 'white'}),
  }),
});
```

```js
// after
import {style} from 'ol/style.js';

const s = style({
  fill: {color: 'orange'},
  stroke: {width: 15, color: 'red'},
  circle: {
    radius: 42,
    stroke: {width: 10, color: 'yellow'},
    fill: {color: 'white'},
  },
});
```

Alternatively, we could support a completely flat object:

```js
// alternatively (not implemented)
import {style} from 'ol/style.js';

const s = style({
  fillColor: 'orange',
  strokeWidth: 15,
  strokeColor: 'red',
  circleRadius: 42,
  circleStrokeWidth: 10, 
  circleStrokeColor: 'yellow',
  circleFillColor: 'white',
});
```

The flat style would require more duplication in terms of types and documentation.  And whenever we modified the options for the traditional style constructors, we'd also need to remember to update the options taken by the new `style` function.  This is probably manageable, but I wanted to try out the simpler implementation first.

We could also make it so this `style` function was called internally.  Then people could just return object literals from their style functions, for example.  (The downside of doing this is that it encourages a lot of garbage creation, but maybe people are already creating new `Style` objects on every call.)

### DIFF
--- a/examples/geojson.js
+++ b/examples/geojson.js
@@ -3,78 +3,65 @@ import Feature from '../src/ol/Feature.js';
 import GeoJSON from '../src/ol/format/GeoJSON.js';
 import Map from '../src/ol/Map.js';
 import View from '../src/ol/View.js';
-import {Circle as CircleStyle, Fill, Stroke, Style} from '../src/ol/style.js';
 import {OSM, Vector as VectorSource} from '../src/ol/source.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
-
-const image = new CircleStyle({
-  radius: 5,
-  fill: null,
-  stroke: new Stroke({color: 'red', width: 1}),
-});
+import {style} from '../src/ol/style.js';
 
 const styles = {
-  'Point': new Style({
-    image: image,
+  Point: style({
+    circle: {radius: 10, stroke: {color: 'red', width: 1}},
   }),
-  'LineString': new Style({
-    stroke: new Stroke({
+  LineString: style({
+    stroke: {
       color: 'green',
-      width: 1,
-    }),
+    },
   }),
-  'MultiLineString': new Style({
-    stroke: new Stroke({
+  MultiLineString: style({
+    stroke: {
       color: 'green',
-      width: 1,
-    }),
+    },
   }),
-  'MultiPoint': new Style({
-    image: image,
-  }),
-  'MultiPolygon': new Style({
-    stroke: new Stroke({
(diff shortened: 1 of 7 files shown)


silvester-pari:
This looks really great! We have some usecases where a style is created by a JSON object, and this change would make thinks much more straightforward.

ahocevar:
I like the flat object style better. And maybe it would make sense to use the same property names as in the Mapbox Style spec, where possible. Also with dashes between words instead of camel case.

jipexu:
👍 oh it will be a great simplification of the code ! A neophyte question it will be the same syntaxe when we using : 
 new ol.x.X({  ??

```
...
                          let style = new ol.style.Style({
                                text: new ol.style.Text({
                                    text: "\uf007",
                                    font: "900 24px 'Font Awesome 5 Free'",
                                    bottom: "Bottom",
                                    fill: fill,
                                    stroke: stroke
                                })
                            });
...
```

chrismayer:
Very cool :+1: Like @ahocevar I also like the flat object style better. A couple of time ago we did such a convenience layer (with minimal set of properties) in a project and it worked out very well (with all downsides of maintaining and so on). So I like the idea of having this in OL directly.

pilatte:
It looks like this PR will never be merged. I've made [this npm package](https://github.com/pilatte/ol-json-styles), it does exactly what is described in the PR. 

Except the syntax is `image: { Circle: ... }` instead of directly `circle: ...`. A benefit of this is that it stays close to the documentation. But it also makes most fields feel a bit redundant `"fill": { "Fill": { "color": "#ff000050" } }`.

```js
import { loadStyle } from "ol-json-style";

const myStyle = loadStyle({
    "Style": {
        "image": {
            "Circle": {
                "radius": 10,
                "fill": {
                    "Fill": {
                        "color": "#ff000050"
                    }
                },
                "stroke": {
                    "Stroke": {
                        "color": "#ff0000ff"
                    }
                }
            }
        }
    }
})
```
This is my first npm package 🥳.

Issue #12872, opened 2021-10-08
Type definition for Attribution is not consistent with the documentation


MatthijsBon:
**Describe the bug**
The documentation clearly states that the `label` can either be a string or an element (like `span`) can be used. However, the Type Definition for `Options.label = string`. I think it should be types the same as `Options.collapseLabel`.

**To Reproduce**
See `ol/attribution/Attribution.js`

**Expected behavior**
The solution is quite easy: 
```
...
* @property {string|HTMLElement} [label='i'] Text label to use for the
* collapsed attributions button.
...
```

tschaub:
Thanks for the report, @MatthijsBon.  A pull request to fix this would be great.

PR #12885, opened 2021-10-11
Fix typeDefs for several Control modules

MatthijsBon:
Pull request to fix https://github.com/openlayers/openlayers/issues/12872


tschaub:
Thanks, @MatthijsBon.

PR #12887, opened 2021-10-12
Exposing clearLoadedExtents in Vector source

Dominik-Reinert:
In very large datasets, it might be needed to refetch data based on
certain conditions. Such a condition could be a change in zoom level on
the map, leading to more finegrained, server-side clustered data.

Right now, it is not possible to reload data with a loader function
after a zoom change, as the loadedExtentsRtree would see the requested
BBox as already loaded, not forwarding the request to the loader
function.

In order to be able to listen to map events outside the vector source
and refetch data based on custom event conditions, the
clearLoadedExtents function was added, so the user can manually decide
when to reset the already fetched extents.

See this stackoverflow issue for reference:
https://stackoverflow.com/questions/68543574/openlayers-slow-performance-with-140k-features/68555759#68555759

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -1015,7 +1015,7 @@ class VectorSource extends Source {
    * Removes all extents from the list of loaded extents.
    */
   clearLoadedExtents() {
-    if(this.loadedExtentsRtree_) {
+    if (this.loadedExtentsRtree_) {
       this.loadedExtentsRtree_.clear();
     }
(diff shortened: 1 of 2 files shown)


mike-000:
The useful application for this might be in custom bbox and tile strategies where the features in the extents when zoomed out exceed the maximum features per request limit of the server https://codesandbox.io/s/vector-esri-forked-r3w1b?file=/main.js

PR #12893, opened 2021-10-13
Allow map target to be an external window

andrewcoder002:
Possibility to setTarget to DIV located in another browser window.


MoonE:
While the event listener to `this.getViewport().getRootNode().defaultView` will be added, it doesn't seem to be removed when the target is changed again.
It might be necessary to store a reference to the current `defaultView` to be able to do

MoonE:
I pushed a commit to make the example work with our example build process.
I added an unload handler to close the window when the user navigates away or reloads the page

To run the example locally you need to install [NodeJS](https://nodejs.org/). Once installed you should be able to run the following commands within the OpenLayers directory.
```bash
npm i
npm run serve-examples
```
The examples should then be available at http://localhost:8080/external-map.html

MoonE:
If I understand it correctly getRootNode should be called with `{composed: true}` to get the document when inside a shadow root. See: https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

`getRootNode` is not available in Internet Explorer, and either needs to be [polyfilled](https://github.com/foobarhq/get-root-node-polyfill/blob/9cfa82dd2f8477734c81ee082e10dd2a28ad4393/index.js#L27) or in case it is not available always use the window object, and the map can't be set to a new window in browsers that don't support it.

I added tabindex="0" to the external map to be able to use keyboard input, but this breaks the mouse interactions. I suspect it could be related to this function: https://github.com/openlayers/openlayers/blob/bbc99e729b4069ad74ce213b6edeb778ff430514/src/ol/MapBrowserEventHandler.js#L236

andrewcoder002:
I faught with it a whole day. In fact I was not very succesfull.

I'm not able to distinguish load/reload of _map_ window from the _master_ window. Proposed `window.performance.navigation.type` does not work from _master_ window properly. I had to put the script directly into html code of _map_ window. It works but it is not very nice.

I also tried to solve mentioned FullScreen issue. I tried to trace event handling but `MapBrowserEventHandler` script is still complicated for me. May be somebody with deeper knowlage can help. Thanks

MoonE:
Have you tried adding `mapWindow.close();` to the `beforeunload` listerner of the external window? It seems to work for me.

Putting the `beforeunload` listener outside the load listener seems to handle the case where the window is closed before it is loaded, is there a reason to have it inside the load handler? As far as I understand it the `unload` event can happen without a `load` event firing first so handling them independently would be better.

MoonE:
I made a few more changes, most notable using the `pagehide` event works in both chrome and firefox without having to rely on having code in the external html page. Adding a timeout to check if the window successfully opened will reset button in case the window did not successfully open or load.

I still need to fix a few tests and I would like to include the `external-map-map.html` source in the example page below the other code.

andrewcoder002:
Can someone from OpenLayers team review proposed changes? Is there anything that must be done to finally commit and close this request?

MoonE:
@andrewcoder002 I rebased this branch on the openlayers/main branch to make the preview builds work again.
To update your local branch you can use these commands:
```sh
# If you have any changes be sure to save (stash/commit) them.
# If you are unsure make a backup copy of your working directory just in case.
git stash
# Assuming your fork of openlayers is named: origin
git fetch origin
git rebase origin/main
# If you had any changes, otherwise skip this
git stash pop
```
I adapted the change you made to disable the map when the main window is not visible to use the `visibilitychange` which works more reliably.

@marlo22 Thanks for your investigation, it would seem you are correct. With the patch below the map will refresh even if the main window is not visible. Though handling that case completely will require some more work which I agree with @andrewcoder002 would be best to do in a separate PR.
```patch
index 81c38e070..dbaf34429 100644
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -1332,7 +1332,10 @@ class PluggableMap extends BaseObject {
    */
   render() {
     if (this.renderer_ && this.animationDelayKey_ === undefined) {
-      this.animationDelayKey_ = requestAnimationFrame(this.animationDelay_);
+      const defaultView = this.getOwnerDocument().defaultView;
+      this.animationDelayKey_ = defaultView.requestAnimationFrame(
+        this.animationDelay_
+      );
     }
   }
```

MoonE:
This PR is ready for a review.

examples/offscreen-canvas.html
```diff
@@ -6,10 +6,11 @@
   The map in this example is rendered in a web worker, using `OffscreenCanvas`. **Note:** This is currently only supported in Chrome and Edge.
 tags: "worker, offscreencanvas, vector-tiles"
 experimental: true
+sources:
+  - offscreen-canvas.worker.js as worker.js
```
tschaub, line 10: If we wanted to stick with structured data here, this could be
```yml
sources:
  - path: offscreen-canvas.worker.js
    as: worker.js
```

examples/external-map.html
```diff
@@ -0,0 +1,13 @@
+---
+layout: example.html
+title: External map
+shortdesc: Move a map to a seperate window.
+docs: >
+  Move a map to a seperate window.
+tags: "external, window"
+sources:
+  - resources/external-map-map.html
```
tschaub, line 9:
```suggestion
  - path: resources/external-map-map.html
```

examples/offscreen-canvas.html
tschaub, line 10:
```suggestion
  - path: offscreen-canvas.worker.js
    as: worker.js
```

examples/webpack/example-builder.js
```diff
@@ -279,80 +271,80 @@ export default class ExampleBuilder {
...
-        source: workerSource,
-      };
-      assets[workerName] = workerSource;
+    let jsSources = jsSource;
+    if (data.sources) {
+      data.extraSources = await Promise.all(
+        data.sources.map(async (fileName) => {
+          const as = fileName.split(/ +as +/);
+          fileName = as[0];
+          const extraSourcePath = path.join(data.dir, fileName);
```
tschaub, line 320:
```suggestion
        data.sources.map(async (extraSource) => {
          const extraSourcePath = path.join(data.dir, extraSource.path);
```

examples/webpack/example-builder.js
```diff
@@ -279,80 +271,80 @@ export default class ExampleBuilder {
...
-      data.worker = {
-        source: workerSource,
-      };
-      assets[workerName] = workerSource;
+    let jsSources = jsSource;
+    if (data.sources) {
+      data.extraSources = await Promise.all(
+        data.sources.map(async (fileName) => {
+          const as = fileName.split(/ +as +/);
+          fileName = as[0];
```
tschaub, line 319:
```suggestion
        data.sources.map(async (extraSource) => {
          const fileName = extraSource.path;
```

examples/webpack/example-builder.js
```diff
@@ -279,80 +271,80 @@ export default class ExampleBuilder {
...
+            ext = 'js';
+          }
+          if (ext === 'js') {
+            source = this.transformJsSource(source);
+            jsSources += '\n' + source;
+          }
+          source = this.cloakSource(source, data.cloak);
+          assets[fileName] = source;
+          return {
+            name: as[1] || fileName,
```
tschaub, line 333:
```suggestion
            name: extraSource.as || extraSource.path,
```

tschaub:
Is the external map supposed to be usable in fullscreen mode?  I'm seeing "map not usable" after going into fullscreen mode.  If the map is not supposed to be usable in fullscreen mode, then I think it would make sense to remove the fullscreen control.

MoonE:
The usability of the map in the external window currently depends on `document.visibilityState === 'visible'`  of the original document for the animation frames to be updated.
Depending on the browser / window maximize / minimize / second screen it may or may not be usable.
It should be usable if it is maximized on a second screen while the original page is visible.

tschaub (approved):

andrewcoder002:
Thanks @tschaub  and @MoonE for doing the last necessary work. I'm still not familiar with processes on GitHub. Will be this fork merged to base branch automatcally or should I do some additional step?

MoonE:
Just merged it.

Thanks for your contribution, @andrewcoder002 and thanks for the review, @tschaub.

PR #12914, opened 2021-10-22
Build site preview as a GitHub action

tschaub:
We are limited to 1,000 build minutes in our Netlify plan and are exceeding this with our pull request deploy previews.  Instead of running our site builds on Netlify, we can do this as part of our GitHub workflows.

I've stopped the integrated deploys from Netlify while we test this alternative method.

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -1,0 +1,22 @@
+name: Build Preview
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 16
+          registry-url: https://registry.npmjs.org/
+      - run: npm ci
+      - run: npm run build-site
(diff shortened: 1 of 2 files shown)


tschaub:
This is complicated by a few things:
 * secrets (like our Netlify auth token) are not available to workflows that are triggered from events on forks - like this pull request
 * so to be able to deploy, we can use the newish [`pull_request_target` event](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target) that will have access to secrets, but will run on the target instead of the base branch (i.e. it will run using workflow/code in `main`)
 * in the deploy workflow that runs on `pull_request_target` we need to wait on the unprivileged build job that runs on `pull_request` events - but github actions don't allow dependencies between workflows so we use the [`lewagon/wait-on-check` action](https://github.com/lewagon/wait-on-check-action)
 * and since github actions don't allow downloading artifacts from jobs in different workflows, we use the [`dawidd6/action-download-artifact` action](https://github.com/dawidd6/action-download-artifact)

This might not work, and we won't know until the workflow is in `main` whether or not it does.  So I'm going to experiment a bit and may have to back out the changes later.

Issue #12913, opened 2021-10-22
Select interaction crashes with empty vector layer when `declutter: true`


perliedman:
**Describe the bug**

If you have a `VectorLayer` with `declutter: true` and its `VectorSource` has no features, attempting to select with a `Select` interaction throws an exception.

**To Reproduce**
1. https://codesandbox.io/s/exciting-tdd-ywt0c
2. Click the map
3. A `TypeError` is shown ("Cannot read properties of null (reading 'all')")

**Expected behavior**
Selecting from an empty layer should work without exceptions even when declutter is set to true.

masihsadatyyc:
@ahocevar is there any ETA for when this will be released? We're also currently facing this issue.

MoonE:
You may want to use [patch-package](https://www.npmjs.com/package/patch-package) in the meantime.

ahocevar:
We publish packages from the main branch automatically, using the `dev` tag. So you can simply `npm install ol@dev` to get a version which includes that change.

ahocevar:
Fixed with #12935.

masihsadatyyc:
Just wanted to confirm that this is fixed. Thank you!

DrEppen:
I see that this issue is still open and I would like to know the status.

I'm still having this issue in VectorTileLayer.js, where frameState.declutterTree is null:

        const declutteredFeatures =
          executorGroups === declutterExecutorGroups
            ? frameState.declutterTree.all().map((item) => item.value)
            : null;

ahocevar:
@DrEppen This is a separate issue, which was fixed with #15028. Please install `ol@dev` to a version which contains the fix.

PR #12935, opened 2021-10-29
fix bug #12913

mrpan:
fix bug when select interaction crashes with empty vector layer when declutter: true

### DIFF
--- a/src/ol/renderer/canvas/VectorLayer.js
+++ b/src/ol/renderer/canvas/VectorLayer.js
@@ -484,8 +484,9 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
         rotation,
         hitTolerance,
         featureCallback,
-        executorGroup === this.declutterExecutorGroup
-          ? frameState.declutterTree&&frameState.declutterTree.all().map((item) => item.value)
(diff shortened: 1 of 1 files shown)


src/ol/renderer/canvas/VectorLayer.js
```diff
@@ -485,7 +485,7 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
         hitTolerance,
         featureCallback,
         executorGroup === this.declutterExecutorGroup
```
ahocevar, line 487:
```suggestion
        executorGroup === this.declutterExecutorGroup && frameState.declutterTree
```

src/ol/renderer/canvas/VectorLayer.js
```diff
@@ -485,7 +485,7 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
...
-          ? frameState.declutterTree.all().map((item) => item.value)
+          ? frameState.declutterTree&&frameState.declutterTree.all().map((item) => item.value)
```
ahocevar, line 488:
```suggestion
          ? frameState.declutterTree.all().map((item) => item.value)
```

ahocevar (approved):

PR #12939, opened 2021-10-29
Example that demonstrates a color expression using variables

tschaub:
While it isn't currently supported to have arrays as style variables for WebGL layers, color expressions can include the use of variables by having explicit variables for red, green, and blue values.

This branch adds [an example](https://deploy-preview-12939--ol-site.netlify.app/examples/cog-colors.html) that demonstrates the use of color expressions based on user-controlled inputs.  In addition, a change to the WebGL tile layer constructor allows `layer.updateStyleVariables()` to be called before the layer is rendered (which is done in the example).

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -3,6 +3,7 @@
     "$": false,
     "arc": false,
     "common": false,
+    "chroma": false,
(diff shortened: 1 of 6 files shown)


ahocevar (approved):
Beautiful!

Issue #12940, opened 2021-10-30
Update country names and borders in example


dzach:
**The problem**
Standard policy of opensource projects regarding international borders has always been to follow what the UN recognizes as such.

In the OpenLayers example [Image Vector Layer](https://openlayers.org/en/latest/examples/image-vector-layer.html), Cyprus, the island, appears divided to two countries, Cyprus and Northern Cyprus, which is incorrect. There is no country named North Cyprus which is internationally recognized as such, or by the UN.

Following the policy mentrioned above, in the same OpenLayers example, Crimea correctly appears to be part of Ukraine, as is internationally recognized.

On the same line, there is no country internationally recognized as 'Macedonia'. The correct name is 'North Macedonia'.

The issue exists in file [countries.geojson](https://openlayers.org/en/latest/examples/data/geojson/countries.geojson)

Attached is a corrected version of the file, without the false international border line, and country name.

[countries.zip](https://github.com/openlayers/openlayers/files/7447209/countries.zip)
.
Thank you.

dzach:
I can submit the changes with a pull request, if necessary.

tschaub:
@dzach - Thanks for the report.  I haven't yet looked at your `countries.zip`, but it would be good to have an updated `countries.geojson` used throughout the examples.  It should be simplified in a similar way or otherwise represent the same level of generalization as the current dataset.

If you are able to create a pull request, please include details about the source for the updated dataset.  It would be good to have details in the git log about the provenance.

mike-000:
Swaziland should be renamed to Eswatini.  It would also be useful to include the ISO A2 code in the country properties to avoid needing a third party geojson for the Style renderer example.

dzach:
@mike-000 - I verified that Eswatini is the correct name with wikipedia and will add it to this pull request.

dzach:
@MoonE : Thank you for reviewing this PR and for pointing to the https://github.com/nvkelso/natural-earth-vector/issues/391#issuecomment-678230303 thread. Regretfully but necessarily, this is a long reply:

**In principle:**

[...]

PR #12941, opened 2021-10-30
Fix for issue #12940: Wrong country borders in example

dzach:
Issue #12940 
Fixed country names:'Swaziland' to 'Eswatini', 'Macedonia' to 'North Macedonia'.
Removed 'North Cyprus' and a wrong border line within 'Cyprus'.

Sources:
All: https://www.un.org/en/about-us/member-states
Eswatini: https://en.wikipedia.org/wiki/Eswatini
North Macedonia: https://en.wikipedia.org/wiki/North_Macedonia
Cyprus: https://en.wikipedia.org/wiki/Cyprus


MoonE:
Thanks for your work on this PR and opening the issue, @dzach

The countries.geojson file is base on data from https://github.com/nvkelso/natural-earth-vector. The name changes you suggest are already included in the latest commit there. I opened #12942 to have these changes included in OpenLayers.

North Cyprus is still a separate country in this version, but I would prefer you take this issue to https://github.com/nvkelso/natural-earth-vector because changing the data manually will make it harder to update in the future.

Here is a comment thread on why the data is like this https://github.com/nvkelso/natural-earth-vector/issues/391#issuecomment-672085221

dzach:
@MoonE : Thank you for reviewing this PR and for pointing to the https://github.com/nvkelso/natural-earth-vector/issues/391#issuecomment-678230303 thread.

As I read in that thread, if I get it correctly, changing the data maually in OL will be necessary, since that project will offer 'worldviews' for users to decide what to use from their dataset. Hence the issue may anavoidably become politicized.

Therefore it all falls back to OL, and what principle to apply in situations like this. If the non internationally recognized 'Northern Cyprus' remains in the dataset, it would be an indication of good faith to provide a reason why.

Please see my long reply in https://github.com/openlayers/openlayers/issues/12940#issuecomment-955683848

PR #12955, opened 2021-11-02
Make MapboxVector layer work in more access key scenarios

ahocevar:
`MapboyVector` layer with Mapbox layers uses a special `mapbox://` url with an `accessToken`. With MapTiler layers, the access token is provided in the `styleUrl`, and the MapTiler server returns a style document with the token in all resources linked from the style document.

But with Esri, a style document requested with an access token does not apply that token to the linked resources. This pull request makes it so Esri style URLs also work when the access token is provided by the user, and appends the token to the tile url (the sprite url does not require a token).

### DIFF
--- a/src/ol/layer/MapboxVector.js
+++ b/src/ol/layer/MapboxVector.js
@@ -99,9 +99,12 @@ export function normalizeStyleUrl(url, token) {
 export function normalizeSourceUrl(url, token, tokenParam) {
   const mapboxPath = getMapboxPath(url);
   if (!mapboxPath) {
-    return token && url.indexOf(token) === -1
-      ? `${url}${url.indexOf('?') === -1 ? '?' : '&'}${tokenParam}=${token}`
-      : url;
+    if (!token) {
+      return url;
+    }
+    const urlObject = new URL(url, location.href);
+    urlObject.searchParams.set(tokenParam, token);
+    return decodeURI(urlObject.href);
   }
   return `https://{a-d}.tiles.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}.vector.pbf?access_token=${token}`;
 }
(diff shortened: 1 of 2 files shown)


src/ol/layer/MapboxVector.js
```diff
@@ -92,13 +92,16 @@ export function normalizeStyleUrl(url, token) {
...
  * @private
  */
-export function normalizeSourceUrl(url, token) {
+export function normalizeSourceUrl(url, token, tokenParam) {
   const mapboxPath = getMapboxPath(url);
   if (!mapboxPath) {
-    return url;
+    return token && url.indexOf(token) === -1
+      ? `${url}${url.indexOf('?') === -1 ? '?' : '&'}${tokenParam}=${token}`
+      : url;
```
tschaub, line 104:
```suggestion
    if (!token) {
        return url;
    }
    url = new URL(url);
    url.searchParams.set(tokenParam, token);
    return url.href;
```

Maybe we can safely assume that the token doesn't include reserved characters.  But, it feels pretty easy to avoid hand concatenating URLs.

tschaub: Hmm.  Maybe this doesn't cover a case where the source is configured with an `accessToken` but the style URL is not a Mapbox type URL.  Not sure if that is supposed to be handled or not.

src/ol/layer/MapboxVector.js
```diff
@@ -155,8 +159,8 @@ const SourceType = {
  * style created with Mapbox Studio and hosted on Mapbox, this will look like
  * 'mapbox://styles/you/your-style'.
  * @property {string} [accessToken] The access token for your Mapbox style. This has to be provided
- * for `mapbox://` style urls. For `https://` and other urls, access keys must be part of the style
- * url.
+ * for `mapbox://` style urls. For `https://` and other urls, any access key must be the last query
+ * parameter of the style url.
```
tschaub, line 163: This answered my question above (`accessToken` only works for `mapbox://` style URLs).

tschaub (approved):
Take or leave the suggestion as you like.

ahocevar:
Thanks for the review and the good suggestion, @tschaub.

PR #12956, opened 2021-11-03
Preserve drawing buffer

mike-000:
Replaces #12845

Sets `preserveDrawingBuffer: true`.as default in WebGL context attributes to make the canvases exportable and compatible with `forEachLayerAtPixel` along with a `disablePreserveDrawingBuffer()` fallback method as suggested in https://github.com/openlayers/openlayers/pull/12845#issuecomment-939544606

Add a simple Heatmap layer to the Export Map example showing changes to the export needed to handle WebGL canvases..

### DIFF
--- a/examples/export-map.js
+++ b/examples/export-map.js
@@ -1,8 +1,12 @@
 import GeoJSON from '../src/ol/format/GeoJSON.js';
 import Map from '../src/ol/Map.js';
 import View from '../src/ol/View.js';
+import {
+  Heatmap as HeatmapLayer,
+  Tile as TileLayer,
+  Vector as VectorLayer,
+} from '../src/ol/layer.js';
 import {OSM, Vector as VectorSource} from '../src/ol/source.js';
-import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
 
 const map = new Map({
   layers: [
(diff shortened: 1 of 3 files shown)


tschaub:
Here are my quick thoughts on this.  I think we should

1. first deal with shared WebGL contexts before adding functionality that may only work without a shared context
2. next establish some rendering benchmarks
3. then determine how preserving the drawing buffer affects performance (different devices, layer setups, etc)
4. only preserve the drawing buffer by default if the performance hit is negligible, or failing that
5. only preserve the drawing buffer when we need to (e.g. instead of on every single frame, only when exporting, only when someone asks for data at a pixel, etc), or if that proves to be impossible to achieve
6. expose a method for people to toggle the behavior, experimental at first, eventually part of the API if we really can't do any better

I'm not in favor of this pull request because it changes the default behavior to something that might negatively affect rendering performance (without helping to establish whether or not or where this might be the case) and adds an API method that only allows opting out of this behavior (and doesn't allow opting back in).

mike-000:
Context sharing possibilities will surely be affected by layer `zIndex` if there are interspersed 2d and WebGL layers?  And any change of `zIndex` might change how they are interspersed and require shared contexts to be recreated.  For `forEachLayerAtPixel` to work for a single layer that layer needs a unique `className` applied to the canvas element (which already happens for both 2d and WebGL) preventing it being shared with other layers.  I think interspersed layer types and updateable `zIndex` will pose a greater problem than respecting a non-updateable `className` within that grouping.

PR #12958, opened 2021-11-04
Reuse temporary canvas in getDataAtPixel()

mike-000:
It seems inefficient to create a new one pixel canvas and context for every call to `getDataAtPixel()` in the CanvasLayerRenderer.  Once created a canvas and context can be reused unless it becomes tainted.


ahocevar (approved):
Thanks, @mike-000!

PR #12976, opened 2021-11-08
Handle ReprojTile in ol/layer/WebGLTile

### LINKED ISSUE #12975: WebGL tile layer renderer does not handle ReprojTile
**Describe the bug**
The WebGL tile layer renderer does not handle ReprojTile.

**To Reproduce**
Open the OpenStreetMap Reprojection example and change `ol/layer/Tile` to `ol/layer/WebGLTile`.  The map opens with the background visible.  Zooming produces a black canvas.  Panning results in an error as a `ReprojTile` is treated as if it is a `DataTile`.

**Expected behavior**
A `ReprojTile` tile should be handled in the same way as an `ImageTile` in the WebGL Tiles example.

mike-000:
Fixes #12975

The WebGL tile layer renderer handles instances of `ImageTile` as in the WebGL Tiles example but if the tile source is reprojected the tiles are instances of `ReprojTile` which should be handled in the same way.  Also as the tile cache is always that for the view projection tile grid, that is what should be expired, if it is different to the tile source projection that cache is needed for further reprojection.

### DIFF
--- a/src/ol/renderer/webgl/TileLayer.js
+++ b/src/ol/renderer/webgl/TileLayer.js
@@ -503,7 +503,7 @@ class WebGLTileLayerRenderer extends WebGLLayerRenderer {
      * @param {import("../../PluggableMap.js").FrameState} frameState Frame state.
      */
     const postRenderFunction = function (map, frameState) {
-      tileSource.expireCache(tileSource.getProjection(), empty);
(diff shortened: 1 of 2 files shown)


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12978, opened 2021-11-09
Add null to style jsdoc of VectorImage

EvertEt:
Similar to BaseVector

### DIFF
--- a/src/ol/layer/VectorImage.js
+++ b/src/ol/layer/VectorImage.js
@@ -39,9 +39,9 @@ import {assign} from '../obj.js';
  * @property {boolean} [declutter=false] Declutter images and text on this layer. The priority is defined
  * by the `zIndex` of the style and the render order of features. Higher z-index means higher priority.
  * Within the same z-index, a feature rendered before another has higher priority.
- * @property {import("../style/Style.js").StyleLike} [style] Layer style. See
- * {@link import("../style/Style.js").default the style docs} for the default style that will be used if
- * this is not defined.
+ * @property {import("../style/Style.js").StyleLike|null} [style] Layer style. When set to `null`, only
+ * features that have their own style will be rendered. See {@link module:ol/style} for default style
+ * which will be used if this is not set.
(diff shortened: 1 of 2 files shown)


MoonE:
There certainly is an inconsistency though I'm unsure if allowing null is the wanted solution here.
There quite a few classes needing a style configuration and some allow null, some don't.
For most config options it is either type or undefined allowed.

```js
/**
src/ol/interaction/Draw.js:
  61   * POINT or MULTI_POINT geometries.
  62:  * @property {import("../style/Style.js").StyleLike} [style]
  63   * Style for sketch features.

src/ol/interaction/Modify.js:
  108   * pointer close enough to a segment or vertex for editing.
  109:  * @property {import("../style/Style.js").StyleLike} [style]
  110   * Style used for the modification point or vertex. For linestrings and polygons, this will

src/ol/interaction/Select.js:
  56   * selectable.
  57:  * @property {import("../style/Style.js").StyleLike|null} [style]
  58   * Style for the selected features. By default the default edit style is used

src/ol/layer/BaseVector.js:
  47   * higher priority.
  48:  * @property {import("../style/Style.js").StyleLike|null} [style] Layer style. When set to `null`, only
  49   * features that have their own style will be rendered. See {@link module:ol/style} for default style

src/ol/layer/VectorImage.js:
  41   * Within the same z-index, a feature rendered before another has higher priority.
  42:  * @property {import("../style/Style.js").StyleLike} [style] Layer style. See
  43   * {@link import("../style/Style.js").default the style docs} for the default style that will be used if

src/ol/layer/VectorTile.js:
  64   * higher priority.
  65:  * @property {import("../style/Style.js").StyleLike} [style] Layer style. See
  66   * {@link import("../style/Style.js").default the style docs} for the default style that will be used if

src/ol/layer/WebGLTile.js:
  39   * @typedef {Object} Options
  40:  * @property {Style} [style] Style to apply to the layer.
  41   * @property {string} [className='ol-layer'] A CSS class name to set to the layer element.
*/
```

MoonE (approved):
Thanks @EvertEt.

PR #12987, opened 2021-11-10
Browser test config update

### LINKED ISSUE #12966: Remove or rework test coverage tool
The Istanbul coverage instrumentation breaks source maps in the tests.  This makes debugging the browser tests more painful than it should be.  When I need to debug the tests, I disable the Istanbul coverage stuff first.  It would be nice not to have to modify the webpack config just to improve debugging.

Istanbul has been deprecated for some time.  It looks like the `karma-coverage` project might have been updated to work with non-deprecated libs (see karma-runner/karma-coverage#377).

tschaub:
I'm hoping to improve the situation with the browser-based tests.  I'm never able to have a successful run of all the tests, and debugging the tests is painful.

For example, here is the output of a clean run of `npm run test-browser` on `main`:
```
Chrome 93.0.4577.0 (Mac OS 10.15.7): Executed 2049 of 2049 (1 FAILED) (41.479 secs / 31.408 secs)
TOTAL: 1 FAILED, 2048 SUCCESS
```
This time I got pretty lucky, only one test failed.  It doesn't even matter which one.  Each time it is something different.

Quickly identifying the source of a problem is hard because Karma is not using the sourcemaps.  For example, if I introduce this intentional test failure:
```diff
--- a/test/browser/spec/ol/Map.test.js
+++ b/test/browser/spec/ol/Map.test.js
@@ -34,11 +34,12 @@ import {createXYZ} from '../../../../src/ol/tilegrid.js';
 import {defaults as defaultInteractions} from '../../../../src/ol/interaction.js';
 import {tile as tileStrategy} from '../../../../src/ol/loadingstrategy.js';
 
-describe('ol/Map', function () {
+describe.only('ol/Map', function () {
   describe('constructor', function () {
     it('creates a new map', function () {
       const map = new Map({});
       expect(map).to.be.a(Map);
+      expect('stack trace').to.be('useful');
     }); 
```

The output of `npm run test-browser` looks like this:
```
Chrome 93.0.4577.0 (Mac OS 10.15.7) ol/Map constructor creates a new map FAILED
	Error: expected 'stack trace' to equal 'useful'
	    at Assertion.assert (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:96:13)
	    at Assertion.be.Assertion.equal (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:216:10)
	    at Assertion.<computed> [as be] (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:69:24)
	    at Context.<anonymous> (/var/folders/ct/jqy2463935j10d2t3_v21nf00000gn/T/_karma_webpack_330121/commons.js:649910:26)
Chrome 93.0.4577.0 (Mac OS 10.15.7) ol/Map constructor creates a new map FAILED
	Error: expected 'stack trace' to equal 'useful'
	    at Assertion.assert (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:96:13)
	    at Assertion.be.Assertion.equal (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:216:10)
	    at Assertion.<computed> [as be] (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:69:24)
	    at Context.<anonymous> (/var/folders/ct/jqy2463935j10d2t3_v21nf00000gn/T/_karma_webpack_330121/commons.js:649910:26)
```
Two puzzling/frustrating things about this: why does it look like things failed twice and how am I supposed to figure out what source location `/var/folders/ct/jqy2463935j10d2t3_v21nf00000gn/T/_karma_webpack_330121/commons.js:649910:26` corresponds with?

If you try to attach another browser and debug (e.g. from http://localhost:9876/debug.html), you find that the sourcemaps are useless and finding any of the sources is painful or not possible.

One of the problems with the browser tests is the way they are instrumented with Istanbul.  The [`istanbul` package](https://www.npmjs.com/package/istanbul) we depend on is deprecated, and the related `coveralls` package adds dependencies to the deprecated [`request`](https://www.npmjs.com/package/request) and [`har-validator`](https://www.npmjs.com/package/har-validator) packages.

Rather than try to find non-deprecated alternatives, I have opted to remove this instrumentation altogether.  The result is that we no longer get output like this at the end of a test run:
```
=============================== Coverage summary ===============================
Statements   : 24.89% ( 5328/21402 )
Branches     : 18.51% ( 1952/10541 )
Functions    : 26.25% ( 835/3180 )
Lines        : 25.1% ( 5218/20785 )
================================================================================
```

I don't find any real value in the above output (is "Branches: 18.51%" good or bad?), so I'm inclined to go without the coverage instrumentation and reporting.  I like keeping track of test coverage in theory.  But in practice with this library, I haven't liked it (though I have had better experiences elsewhere).

The changes in this branch remove the coverage instrumentation.  This makes it possible to debug tests in a browser.

In addition, this branch adds [`karma-source-map-support`](https://www.npmjs.com/package/karma-source-map-support) to the Karma config.  This give more meaningful stack traces when tests fail.  For example, with the same intentionally added test failure from above, the stack trace (singular) looks like this:

```
Chrome Headless 93.0.4577.0 (Mac OS 10.15.7) ol/Map constructor creates a new map FAILED
	Error: expected 'stack trace' to equal 'useful'
	    at Assertion.assert (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:96:13)
	    at Assertion.be.Assertion.equal (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:216:10)
	    at Assertion.<computed> [as be] (/Users/tschaub/projects/openlayers/node_modules/expect.js/index.js:69:24)
	    at Context.<anonymous> (/var/folders/ct/jqy2463935j10d2t3_v21nf00000gn/T/_karma_webpack_93512/webpack:/ol/test/browser/spec/ol/Map.test.js:42:1)
```

The improvement is that you can see the failure location at `ol/test/browser/spec/ol/Map.test.js:42:1`.

The other change is to use Headless Chrome instead of Chrome by default.  With these changes, a full run of `npm run test-browser` succeeds with this:

```
Chrome Headless 93.0.4577.0 (Mac OS 10.15.7): Executed 2049 of 2049 SUCCESS (37.433 secs / 26.377 secs)
```

Fixes #12966.

If someone would like to bring back coverage reporting, I think it is important to ensure that we still

 * get useful stack traces when things fail
 * can use source files in a debugger
 * have all tests pass


tschaub:
image

jahow:
The coverage reporting did also produce a full HTML report file by file (and line by line), although I don't know if that is still working. But I agree, the real value of a coverage report would be to indicate whether a branch increases or decreases coverage and where, not much more. Having an absolute value at fixed point in time is not incredibly useful.

Also, what was the rationale for removing the package-lock? I had the feeling it was required for CI runs.

tschaub:
> Also, what was the rationale for removing the package-lock?

This branch doesn’t remove `package-lock.json`.  It does remove lines from that file representing the removed dependencies.

jahow:
> This branch doesn’t remove `package-lock.json`. It does remove lines from that file representing the removed dependencies.

Right sorry! I assumed such a large diff meant that the file was removed. Looks like `istanbul` brought a lot of clutter with it!

ahocevar:
I like this change. Since we split the tests into browser and node tests, the test coverage report was not useful any more anyway, because it covered only the browser tests.

In the medium term, we might want to switch to a setup that runs all tests in node and the browser (the former filtered to only run tests that work in node), but that would be a greater effort, which would involve changing things to get rid of Karma, and either run the mocha tests directly in a browser, or switch to something like Jasmine, which makes it easier to run tests in node and the browser.

ahocevar (approved):

PR #12988, opened 2021-11-11
Setting default view reference

marlo22:
It's my attempt to fix a bug described [here](https://github.com/openlayers/openlayers/pull/12893#issuecomment-948284350). What do you think about this approach? Is it acceptable?

### DIFF
--- a/examples/external-map.js
+++ b/examples/external-map.js
@@ -2,29 +2,14 @@ import Map from '../src/ol/Map.js';
 import OSM from '../src/ol/source/OSM.js';
 import TileLayer from '../src/ol/layer/Tile.js';
 import View from '../src/ol/View.js';
-import {
-  Control,
-  FullScreen,
-  defaults as defaultControls,
-} from '../src/ol/control.js';
+import {FullScreen, defaults as defaultControls} from '../src/ol/control.js';
 import {fromLonLat} from '../src/ol/proj.js';
-
-class UnusableMask extends Control {
-  constructor() {
-    super({
-      element: document.createElement('div'),
-    });
-    this.element.setAttribute('hidden', 'hidden');
-    this.element.className = 'ol-mask';
-    this.element.innerHTML = '<div>Map not usable</div>';
-  }
-}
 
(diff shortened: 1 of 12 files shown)


MoonE:
Seems to be working nicely.

Storing the default view in a global object seems problematic when more than one map is added to a page, though I'm unsure how to solve this for the calls in the source classes.
Maybe the map property could be propagated further to the source as described in https://github.com/openlayers/openlayers/pull/12965#issue-1046568946 by @tschaub 

You can also remove this css that was used for the mask to disable the external map.
https://github.com/openlayers/openlayers/blob/924b85f4cae0d6a6da05adfd7c051eaa23f98081/examples/resources/external-map-map.html#L12-L24

ahocevar:
Since a `Source` can be shared among multiple layers which can be on different maps, what @MoonE suggests above will not work. As it is now, I'm not in favor of this pull request, because the implemented approach will only work for a single map, and I'm unsure about collateral damage in terms of memory leaks. If this approach is persued further, the way we request animation frames will probably have to be entirely reworked. I'm imagining it could be done by including information about the source window in the frame state, and then every caller of `requestAnimationFrame` would need the frame state, which might be possible without too much effort.

An alternative approach would be if applications that need popout maps prevent the main app from entering sleep mode by using the Screen Wake Lock API, as described in https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API#examples. Unfortunately, this is currently not supported by many browsers.

marlo22:
@ahocevar, you're totally right, I think. I'm closing this PR for now. :wink: Thank you for the answers.

NoriSte:
> An alternative approach would be if applications that need popout maps prevent the main app from entering sleep mode by using the Screen Wake Lock API, as described in [developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API#examples](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API#examples). Unfortunately, this is currently not supported by many browsers.

FYI: I tried it and it does not work 😊 thanks for suggesting!

PR #12990, opened 2021-11-11
Fix for image smoothing disabled / hardware acceleration / reprojection issue

mike-000:
The issue with hardware acceleration and Chrome noted in https://github.com/openlayers/openlayers/pull/10990#issuecomment-623567917 is still there and now affects Edge as well.

This happens when all of these are simultaneously true:

- Chromium-based browser with hardware acceleration enabled
- Same tile in browser cache in two or more OpenLayers source caches.
- Tile sources are reprojected.
- Used with image smoothing enabled, then disabled.

I have found that drawing the image unscaled to a temporary canvas before drawing that scaled to the stitchContext produces consistent results with smoothing disabled.  The pixels chosen when smoothing is disabled do differ between browsers and their settings, including hardware acceleration, but the values are in the expected range (there are no depressions deeper than  the Dead Sea on the slopes of Mont Blanc).

### DIFF
--- a/src/ol/reproj.js
+++ b/src/ol/reproj.js
@@ -248,20 +248,36 @@ export function render(
 
   const stitchScale = pixelRatio / sourceResolution;
 
+  let imageContext;
+  if (opt_contextOptions === IMAGE_SMOOTHING_DISABLED) {
+    imageContext = createCanvasContext2D(1, 1);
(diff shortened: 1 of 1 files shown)


mike-000:
Issue fixed by #13097

PR #12994, opened 2021-11-13
Add a map property to layers

tschaub:
Layers can only be rendered on a single map.  Layer renders have access to the layer, but not the corresponding map.  This makes it awkward for renderers to clean up after themselves when a layer is removed from a map, for example.

By adding a "map" property to layers, renderers can listen for changes in this property and clean up when necessary.

This was originally part of #12965.  I pulled it out to make for easier review.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -5,7 +5,8 @@ import BaseObject from './Object.js';
 import Collection from './Collection.js';
 import CollectionEventType from './CollectionEventType.js';
 import EventType from './events/EventType.js';
-import LayerGroup from './layer/Group.js';
+import Layer from './layer/Layer.js';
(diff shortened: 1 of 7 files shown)


src/ol/PluggableMap.js
```diff
@@ -524,6 +555,14 @@ class PluggableMap extends BaseObject {
     layers.push(layer);
   }
 
+  /**
+   * @param {import("./layer/Group.js").GroupEvent} event The layer add event.
+   * @private
+   */
+  handleLayerAdd_(event) {
+    setLayerMapProperty(event.layer, this);
+  }
```
tschaub, line 564: We could dispatch an `addlayer` event here on the map if we wanted others to be able to leverage this event (and `removelayer` below).  I don't plan on doing that as part of this change.

ahocevar (approved):

PR #12996, opened 2021-11-13
Align labels with the Select elements in the Raster Reprojection example

mike-000:
Center aligned labels do not work well in columns above left aligned Select elements on a wide screen where the column widths may be much wider than the Selects.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #12998, opened 2021-11-14
Fix `ol/source/Cluster#setSource` type annotation

MoonE:
Type definitions currently don't allow removing the source and Don't indicate that `getSource` may not return a source object

I tried to match the behaviour of `ol/Feature#setGeometry` which also allows undefined.

### DIFF
--- a/src/ol/source/Cluster.js
+++ b/src/ol/source/Cluster.js
@@ -52,7 +52,7 @@ import {getUid} from '../util.js';
  *   });
  * }
  * ```
- * @property {VectorSource} [source] Source.
+ * @property {VectorSource} [source=null] Source.
  * @property {boolean} [wrapX=true] Whether to wrap the world horizontally.
  */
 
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
Not sure if I'm understanding correctly what is currently broken, so I added a code comment.

src/ol/source/Cluster.js
```diff
@@ -63,7 +63,7 @@ import {getUid} from '../util.js';
  * considered for clustering, a custom `geometryFunction` can be defined.
  *
  * If the instance is disposed without also disposing the underlying
- * source `setSource(null)` has to be called to remove the listener reference
+ * source `setSource(undefined)` has to be called to remove the listener reference
```
ahocevar, line 66: Did `setSource(null)` work previously, i.e. did it remove the source? If so, then this is a breaking change.

EvertEt:
Looking at https://github.com/openlayers/openlayers/pull/13301, it would be more consistent to just add null to the types here, I could open a PR for this if preferred.

src/ol/source/Cluster.js
MoonE: Yes, passing null removed the source, I changed it again to accept null

MoonE:
@ahocevar The function did not allow passing null when strict type checking is enabled.
@EvertEt Thanks for the offer. I updated this branch instead.

ahocevar (approved):

PR #13009, opened 2021-11-15
Use workflow_run event to deploy pull request previews

### LINKED ISSUE #12991: Deploy preview fails for new contributors
**Describe the bug**
The deploy preview task fails on PRs of new contributors before/without approval

**To Reproduce**
1. I've seen it in #12988 and #12985
2. See error

**Expected behavior**
It should either not start the deploy preview task or complete successfully

tschaub:
This branch updates our **Deploy Preview** workflow so that it is triggered by the [`workflow_run` event](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run).  Previously, this workflow was triggered by the `pull_request_target` event.  Both events run based on the workflow configured in the default branch and have access to secrets (since they do not run on PR-submitters code).  The benefit of the `workflow_run` event is that it is triggered after other workflows run - so we can wait for the **Build Preview** workflow to complete before deploying.  The previous solution relied on polling for the build step to complete.  The shortcoming of this is that the deploy workflow would time out while the build workflow was awaiting approval for first time contributors.

I've also updated the deploy preview workflow so that it adds a comment to a pull request with links to the deployed examples and API docs.  I think this is easier to find than the current link at the bottom of the list of checks for a pull request.  The comment should only appear once (even though the deploy is updated for each new commit pushed after the PR is opened).

image

Fixes #12991.

We won't see this new comment on this pull request since the deploy preview workflow in the default branch is the one being run.  So unfortunately we can't fully validate this until it is in `main` and we open another pull request.  If I've made any typos, I may make changes directly in `main` to correct them.

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -9,14 +9,21 @@ jobs:
   build-preview:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
         with:
-          node-version: 16
-          registry-url: https://registry.npmjs.org/
+          node-version: '16'
       - run: npm ci
       - run: npm run build-site
       - uses: actions/upload-artifact@v2
         with:
           name: site
           path: build/site
+      - name: Store pull request number for later use
+        run: |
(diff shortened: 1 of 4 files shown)


ahocevar (approved):
Thanks for your efforts on this, @tschaub.

PR #13012, opened 2021-11-16
Remove unused dependencies

tschaub:
This removes a few dev dependencies that are no longer in use.


package.json
```diff
@@ -72,10 +72,8 @@
     "express": "^4.17.1",
     "front-matter": "^4.0.0",
     "fs-extra": "^10.0.0",
-    "glob": "^7.1.5",
```
tschaub, line 75: This was added in 8b8d94c78a9c996c2da6916b2e883c6bb7f0641d for `tasks/transform-types.js`.  It should have been removed in 7c6755d3ecca774a332390c1e271bac5c0f363a9 when that task was removed.

package.json
```diff
@@ -72,10 +72,8 @@
...
     "globby": "^12.0.0",
     "handlebars": "4.7.7",
-    "istanbul": "0.4.5",
```
tschaub, line 78: This should have been removed as part of #12987.

package.json
```diff
@@ -99,9 +97,7 @@
     "serve-static": "^1.14.0",
     "shx": "^0.3.2",
     "sinon": "^12.0.1",
-    "terser-webpack-plugin": "^5.1.1",
```
tschaub, line 102: This is a dependency of webpack.

package.json
```diff
@@ -99,9 +97,7 @@
...
     "threads": "^1.6.5",
-    "url-polyfill": "^1.1.5",
```
tschaub, line 104: This was added in 097c7b689a6e8b8ddff526897df9b478b139e500 for legacy browser testing and is no longer needed.

ahocevar (approved):

PR #13018, opened 2021-11-17
Add projection selector to Disable Image Smoothing example

mike-000:
While #12990 might make results more meaningful for some browsers, there remains considerable browser inconsistency over the way `ctx.imageSmoothingEnabled = false` works in conjunction context transforms.  With IE only pixel values seen in the original image were used, which was the desired result, but Chromium based browsers have some form of interpolation, even through that usually preserves meaningful numeric results.  I think it would be best to mention that the results can differ between browsers and add a projection selector to the example so reprojection can be turned off/on.

### DIFF
--- a/examples/disable-image-smoothing.html
+++ b/examples/disable-image-smoothing.html
@@ -5,7 +5,8 @@ shortdesc: Example of disabling image smoothing
 docs: >
   Example of disabling image smoothing when using raster DEM (digital elevation model) data.
   The <code>imageSmoothing: false</code> setting is used to disable canvas image smoothing during
-  reprojection and rendering. Elevation data is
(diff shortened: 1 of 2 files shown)


mike-000:
I think the apparent interpolation may have been due to tiles from different zoom level being used in the reprojection.  In the example (as long as #12990 is applied) there are typically 4 pixels in the original tile for each pixel in the reprojected tile - one of those will correspond to the reprojection value, but which one depends on the browser and whether hardware acceleration is enabled.

mike-000:
Correcting the tile size in #13097 fixes the inconsistencies, and a projection selector such as this which maintains scale would be best demonstrated in a separate example which includes a scale bar.

PR #13021, opened 2021-11-18
Add the renderer type to generic layer template variables

tschaub:
This change makes it so the `layer.getRenderer()` return type is specific for the layers that create specific renderers.


tschaub:
Despite the queued CodeQL check (see https://github.com/openlayers/openlayers/pull/13022#issuecomment-973643394), this is ready for review.

ahocevar (approved):
Very nice, thanks!

ahocevar:
This will probably be fixed with #13190.

PR #13022, opened 2021-11-18
Update the security scan task

tschaub:
It looks like the **Security Scan** workflow is creating a new queued commit status and never updating it (see https://github.com/openlayers/openlayers/pull/13021/checks?check_run_id=4257816265).  This branch is an effort to troubleshoot that.


tschaub:
I've opened a support request about the permanently queued CodeQL check.

tschaub:
I pushed another commit that removes the CodeQL workflow altogether.  If we aren't able to figure out the issue with the permanently queued CodeQL check, we can remove the whole thing.

tschaub:
We can resurrect this workflow later if CodeQL things start working again.

PR #13025, opened 2021-11-19
Add background to MapboxVector layer

### LINKED ISSUE #13023: MapboxVector does not render dark map style
**Describe the bug**
When using  the [`mapbox/dark-v10`](https://api.mapbox.com/styles/v1/mapbox/dark-v10.html?title=true&access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4M29iazA2Z2gycXA4N2pmbDZmangifQ.-g_vE53SD2WrJ6tFX7QHmA#1.07/0/0)  style with `MapboxVector` the style render with light colors on lands, in place of dark ones.

**To Reproduce**
1. Go to https://codesandbox.io/s/mapbox-vector-layer-forked-z7g4q?file=/main.js
2. Replace `YOUR_ACCESS_TOKEN` with your Mapbox access token
3. See that the colors on lands in light

**Expected behavior**
When using Mapbox GL JS, the style render correctly : https://jsfiddle.net/pj2s491b/

ahocevar:
This pull request makes it so the `MapboxVector` layer picks up the background from the style. This is achieved by adding a polygon as first feature to each tile, which covers the whole tile extent and is filled with the background's color and opacity.

Fixes #13023.

### DIFF
--- a/examples/mapbox-vector-layer.css
+++ b/examples/mapbox-vector-layer.css
@@ -1,3 +1,0 @@
-.map {
-  background: #f8f4f0;
-}
--- a/src/ol/layer/MapboxVector.js
+++ b/src/ol/layer/MapboxVector.js
@@ -3,11 +3,17 @@
  */
 import BaseEvent from '../events/Event.js';
 import EventType from '../events/EventType.js';
+import GeometryType from '../geom/GeometryType.js';
 import MVT from '../format/MVT.js';
+import RenderFeature from '../render/Feature.js';
 import SourceState from '../source/State.js';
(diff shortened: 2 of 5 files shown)


tschaub (approved):

tschaub:
At some point, we could add support for a `background` option on vector (or all) layers.  Not sure how often it would get used, but it would be pretty straightforward to handle it at the layer renderer level.

ahocevar:
Actually I was thinking the same. I'probably start working on a `background` option soon, now that you also suggest it.

Issue #13024, opened 2021-11-19
Question about use of "object naming" ?


jipexu:
Hi
Sorry this is just a question for my knowledge ...
Before with the prior version < 6.6 ? i had some code that stoped to work when i use > 6.6 and 6.9 it seems due to a change of the "object naming"?
for example
 idIndex_ become Tu
values_ become A
_id become W
array_ become R
strokeStyle_.color_ become Kh.Gs
... 

so my question is somebody can explain when and why this had changed  > 6.6 ? and is it a documentation ? or give me some links to understand better ?

Many thanks

TomazicM:
Mike answered this here: https://gis.stackexchange.com/questions/416701/web-console-result-is-not-comprehensible-when-switched-to-new-openlayers-library

jipexu:
Thanks for the links !
Effectively the mike answer is shorter as the new private property names ... and sure the final code is less human readable... 
example

```            
graticule.strokeStyle_.color_="rgba(0,0,0,0.2)";
become
graticule.Kh.Gs="rgba(0,0,0,0.2)";
```
good it doesn’t do anything the main thing is to know it ...

ahocevar:
Well, private means you shouldn't be using it anyway. If you want to use the legacy build know what you're doing, you can always create your own build with something like https://github.com/phpmyadmin/phpmyadmin/pull/16959. If you use a bundler and know what you are doing, you can access private properties. The latter is quite easy, e.g. with https://www.npmjs.com/package/create-ol-app, and gives you much nicer debugging thanks to source map support, nicer than with the pre-6.7 legacy build.

jipexu:
@ahocevar many thank for your answer even if I do not understand everything (but I will try!)... I know that the applications I maintain are coded according to my knowledge and that I do not use the right method recommended ... In the above example I don’t know how to write this other than using private property to make it work... I will try to understand the two links you mentioned. I searched in the documentation and in some js files to know where and when these names are assigned but I did not find ...

ahocevar:
Whenever you think you need to access private properties, there is a chance you're doing things wrong. In your above example, the API way to do it would be to create a graticule with the desired style
```js
new Graticule({
  strokeStyle: new Stroke({
    color: 'rgba(0,0,0,0.2)',
  });
});
```
If that isn't sufficient, you could file a feature request to make the stroke style settable after instantiation.

ahocevar:
[...]

PR #13026, opened 2021-11-19
Fix strokeStyle documentation

ahocevar:
See #13024. It seems the documentation for the `strokeStyle` constructor option of `ol/layer/Graticule` is misleading.

### DIFF
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -86,9 +86,13 @@ const INTERVALS = [
  * appropriate for conformal projections like Spherical Mercator. If you
  * increase the value, more lines will be drawn and the drawing performance will
  * decrease.
- * @property {Stroke} [strokeStyle='rgba(0,0,0,0.2)'] The
- * stroke style to use for drawing the graticule. If not provided, a not fully
- * opaque black will be used.
+ * @property {Stroke} [strokeStyle] The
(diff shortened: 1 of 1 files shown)


src/ol/layer/Graticule.js
```diff
@@ -86,9 +86,14 @@ const INTERVALS = [
  * appropriate for conformal projections like Spherical Mercator. If you
  * increase the value, more lines will be drawn and the drawing performance will
  * decrease.
- * @property {Stroke} [strokeStyle='rgba(0,0,0,0.2)'] The
+ * @property {Stroke} [strokeStyle] The
  * stroke style to use for drawing the graticule. If not provided, a not fully
- * opaque black will be used.
+ * opaque black will be used. If not provided, the following stroke will be used:
```
tschaub, line 91: This now reads "If not provided, a not fully opaque black will be used. If not provided, the following stroke will be used."

I think you probably meant to change it to "If not provided, the following stroke will be used."

tschaub:
Funny.  I just left another comment on the support request about permanently queued **Code scanning results / CodeQL** check - and then returned here to find they were now marked as complete!

src/ol/layer/Graticule.js
ahocevar: Ah, thanks @tschaub. You're absolutely right.

tschaub (approved):

PR #13027, opened 2021-11-19
Restore the security scan workflow

tschaub:
It looks like the issues with the **Code scanning results / CodeQL** status check have been fixed.

Reverts openlayers/openlayers#13022


tschaub:
image

I'm assuming this "1 analysis not found" and base branch not found are because there wasn't a security scan for the parent commit.  But the whole CodeQL thing is a mysterious black box (of dubious value) to me.

ahocevar (approved):

PR #13030, opened 2021-11-20
Correctly set popover position in geographic example

### LINKED ISSUE #13029: geographic example - strange longitude
Hi
In this example (https://openlayers.org/en/latest/examples/geographic.html)
i known this is an experimental example ..
when you drag the map until a "parrallel world" to the east or to the west
- the result of the get center give wrong longitude (+|- 180 +++ )
- and the click on the red marker open the popup in the original world ...

Thanks

MoonE:
Fixes #13029

Popover is only initialized once, calling `popover` again does not change the content unless the popover was disposed.
Not necessary for this example as there is only one feature, but feels like a good idea to have it anyways.

### DIFF
--- a/examples/geographic.js
+++ b/examples/geographic.js
@@ -63,10 +63,15 @@ map.on('moveend', function () {
 });
 
 map.on('click', function (event) {
+  $(element).popover('dispose');
+
   const feature = map.getFeaturesAtPixel(event.pixel)[0];
   if (feature) {
     const coordinate = feature.getGeometry().getCoordinates();
-    popup.setPosition(coordinate);
+    popup.setPosition([
+      coordinate[0] + Math.round(event.coordinate[0] / 360) * 360,
+      coordinate[1],
+    ]);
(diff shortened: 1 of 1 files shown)


jipexu:
Thanks
what do you think for the latitude values ?

MoonE:
I think it makes sense to _not_ keep the center longitude to the range of `[ -180, 180 )`. 
The example shows how to work with geographic coordinates in OpenLayers and coordinates won't wrap automatically, so I think it makes sense to display it this way.

tschaub (approved):

jipexu:
Ok but for me it is not comprehensible and quite confusing that a longitude can display such results in a geographic application... A visitor playing with the example will say at first glance that something is wrong ... It is not the correctness or logic of the code that causes me the problem but what is shown in this example...

MoonE:
I understand you concerns from a user's point of view, but these are the values provided by OpenLayers for developers.

PR #13045, opened 2021-11-24
Fix tiles and background handling

ahocevar:
This pull request fixes two issues with the MapboxVector layer:
* Error on styles without `background` because of a faulty boolean expression
* Error on sources with `tiles` but without `url`, because we always expected a `url`

I added regression tests for both fixes.

### DIFF
--- a/src/ol/layer/MapboxVector.js
+++ b/src/ol/layer/MapboxVector.js
@@ -405,10 +405,7 @@ class MapboxVectorLayer extends VectorTileLayer {
     }
 
     const source = this.getSource();
-    if (
-      styleSource.url.indexOf('mapbox://') === 0 ||
-      styleSource.url.indexOf('{z}') !== -1
-    ) {
+    if (styleSource.url && styleSource.url.indexOf('mapbox://') === 0) {
       // Tile source url, handle it directly
       source.setUrl(
         normalizeSourceUrl(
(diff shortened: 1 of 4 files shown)


src/ol/layer/MapboxVector.js
```diff
@@ -405,10 +405,7 @@ class MapboxVectorLayer extends VectorTileLayer {
     }
 
     const source = this.getSource();
-    if (
-      styleSource.url.indexOf('mapbox://') === 0 ||
-      styleSource.url.indexOf('{z}') !== -1
```
ahocevar, line 410: I removed this because according to the spec, a `url` in the source can never contain a tile url template.

test/browser/spec/ol/layer/MapboxVector.test.js
```diff
@@ -141,7 +141,7 @@ describe('ol/layer/MapboxVector', () => {
               version: 8,
               sources: {
                 'foo': {
-                  url: '/spec/ol/data/{z}-{x}-{y}.vector.pbf',
+                  tiles: ['/spec/ol/data/{z}-{x}-{y}.vector.pbf'],
```
ahocevar, line 144: This change not only makes the style valid, it also turns this test into a regression test for sources without a `url`.

ahocevar:
Some of the confusion with the MapboxVector layer tests came from the fact that our test styles provided the tile url template in the `url` property instead of the `tiles` property. I fixed the test styles in b61969861b831efce608fbfe5a0458f65498bd6b.

tschaub (approved):

PR #13054, opened 2021-11-29
Updates for newer version of the OGC API spec draft

### LINKED ISSUE #13048: OGC API - Tiles demo pages not working after removing obsolete tileMatrixSetDefinition and links from server
**Describe the bug**
Both the OGC Map & Vector Tiles demo pages no longer work after we updated our demo server.

**To Reproduce**
1. Visit https://openlayers.org/en/latest/examples/ogc-vector-tiles.html and https://openlayers.org/en/latest/examples/ogc-map-tiles.html
2. Notice that nothing shows up

**Expected behavior**
The maps should show up.

**Additional details**
We have removed the obsolete _tileMatrixSetDefinition_ property and obsolete links from our tilesets ([NE countries](https://maps.ecere.com/ogcapi/collections/NaturalEarth:cultural:ne_10m_admin_0_countries/tiles/WebMercatorQuad) and [Blue Marble](https://maps.ecere.com/ogcapi/collections/blueMarble/tiles/WebMercatorQuad)).

Obsolete elements that were removed include:
- The _tileMatrixSetDefinition_ property in the TileSet. That has been replaced by a link with the http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme relation type inside the `links` array. (I believe this to likely be the cause of the problem, since the demo pages points directly to the individual tilesets)
- Links with relation type http://www.opengis.net/def/rel/ogc/1.0/tilesets from the collection page (there is now a dedicated relation type for _tilesets-vector_, _tilesets-coverage_ and _tilesets-map_)
- The links from the tilesets list (at `/tiles/`) now use the `self` relation type, rather than the http://www.opengis.net/def/rel/ogc/1.0/tileset relation type.

These [changes](https://github.com/opengeospatial/2D-Tile-Matrix-Set/commit/b26857dc01e9d1901f59e00b794b709ed0b7f128#diff-7b795ea89a27c3ad024f3b12c3bcd512e5d5e781af7f47258a17aa43c2eb67acL81) to the TileSet metadata were done around August. See also https://github.com/opengeospatial/2D-Tile-Matrix-Set/issues/59

At this point, the TileSet schema should be fairly stable apart from addressing comments received during the [comment period](https://www.ogc.org/pressroom/pressreleases/4598) (including 5 of my own at https://github.com/opengeospatial/2D-Tile-Matrix-Set/issues) for the version 2.0 of 2DTMS introducing this new TileSet metadata, which closes next week.

Please let me know if any other issue with our server is found to prevent the demos from working. Thank you!

@tschaub

ahocevar:
This pull request addresses the spec updates listed in #13048, and thereby fixes the OGC API examples.

Fixes #13048

### DIFF
--- a/src/ol/proj/epsg4326.js
+++ b/src/ol/proj/epsg4326.js
@@ -63,7 +63,7 @@ export const PROJECTIONS = [
   new EPSG4326Projection('EPSG:4326', 'neu'),
   new EPSG4326Projection('urn:ogc:def:crs:OGC:1.3:CRS84'),
   new EPSG4326Projection('urn:ogc:def:crs:OGC:2:84'),
-  new EPSG4326Projection('http://www.opengis.net/def/crs/OGC/1.3/CRS84', 'neu'),
+  new EPSG4326Projection('http://www.opengis.net/def/crs/OGC/1.3/CRS84'),
   new EPSG4326Projection('http://www.opengis.net/gml/srs/epsg.xml#4326', 'neu'),
   new EPSG4326Projection('http://www.opengis.net/def/crs/EPSG/0/4326', 'neu'),
 ];
--- a/src/ol/source/ogcTileUtil.js
+++ b/src/ol/source/ogcTileUtil.js
@@ -373,11 +373,18 @@ function parseTileSetMetadata(sourceInfo, tileSet) {
     );
   }
 
-  if (!tileSet.tileMatrixSetDefinition) {
-    throw new Error('Expected tileMatrixSetDefinition or tileMatrixSet');
+  const tileMatrixSetLink = tileSet.links.find(
+    (link) =>
+      link.rel === 'http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme'
+  );
+  if (!tileMatrixSetLink) {
+    throw new Error(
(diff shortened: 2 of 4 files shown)


src/ol/proj/epsg4326.js
```diff
@@ -63,7 +63,7 @@ export const PROJECTIONS = [
   new EPSG4326Projection('EPSG:4326', 'neu'),
   new EPSG4326Projection('urn:ogc:def:crs:OGC:1.3:CRS84'),
   new EPSG4326Projection('urn:ogc:def:crs:OGC:2:84'),
-  new EPSG4326Projection('http://www.opengis.net/def/crs/OGC/1.3/CRS84', 'neu'),
+  new EPSG4326Projection('http://www.opengis.net/def/crs/OGC/1.3/CRS84'),
```
ahocevar, line 66: According to https://docs.opengeospatial.org/pol/09-048r5.html#_names_for_epsg_definitions, `http://www.opengis.net/def/crs/OGC/1.3/CRS84` is equivalent to the previous `CRS:84` identifier, which uses the standard `enu` axis order.

tschaub: Interesting.  I think I added this after using an early GeoServer example.  But could be remembering wrong.

tschaub (approved):

Issue #13053, opened 2021-11-29
Zoom out interaction breaks web map with GeoTIFF


drnextgis:
If zoom out a web map with GeoTIFF layer added on it then at some point in time all layers are gone. JSFiddle with reproducible example: https://jsfiddle.net/7kyfL3za/1/ 

https://user-images.githubusercontent.com/866124/143831348-b4aaad00-b483-4aaa-bdd6-bf5978694db9.mp4

ahocevar:
This was fixed with the context sharing added in #12965.

drnextgis:
Great! Looking forward to the new release!

Can I please ask one more question here. Is there a way to disable GeoTIFF rendering outside of the image boundaries for rasters without nodata value?

tschaub:
@drnextgis if you want to, you can try out the context sharing now on the `ol@dev` release.

Regarding nodata values, if they are advertised in the GeoTIFF headers, they should be transparent.  Otherwise you can include a `nodata` property in the source configuration (next to the `url`).

tschaub:
I still see a problem when zooming out using the latest.  We'll leave this open to investigate.

drnextgis:
Thanks @tschaub! It is very clear how nodata works. My main concern was regarding the areas outside of the image boundaries. Is there a way to prevent OL to render them? In some cases the nodata value is not advertised and we don't know it advance to set it manually. It is ok to have these dark areas inside the image extent but not outside of it.

tschaub:
Here is what the example looks like when I add `nodata: 0` to the source configuration:
![image](https://user-images.githubusercontent.com/41094/143903106-2c83ce9c-5a51-4bb9-99ad-e998d3d0fb93.png)

If you are seeing something different, it may have been addressed by https://github.com/geotiffjs/geotiff.js/pull/248.

drnextgis:
So it turns out that preventing pixels be rendered outside of the image boundaries can be achieved only via `nodata` property? The thing is I don't know `nodata` in advance and I would like OL to render it like this (by respecting boundaries of the raster):
![image](https://user-images.githubusercontent.com/866124/143904563-8853d8a7-9631-42c7-af3c-f7760b60b577.png)

tschaub:
Ideally the nodata value is present in the GeoTIFF header, so you don't need to configure it at all - but unfortunately we cannot rely on this either.

I agree it would be nice to discard pixels outside the image extent altogether.  This is another thing to look into.

[...]

PR #13060, opened 2021-11-29
Avoid fetching data outside the grid extent

tschaub:
This branch makes it so the WebGL tile renderer doesn't fetch data outside the configured tile grid extent (see https://github.com/openlayers/openlayers/issues/13053#issuecomment-981795559).

A follow-up enhancement would be to mask out areas outside the layer extent (and perhaps the tile grid extent).  I'm not taking that on here.


src/ol/source/DataTile.js
```diff
@@ -31,7 +31,7 @@ import {toPromise} from '../functions.js';
  * @property {boolean} [opaque=false] Whether the layer is opaque.
  * @property {import("./State.js").default} [state] The source state.
  * @property {number} [tilePixelRatio] Tile pixel ratio.
- * @property {boolean} [wrapX=true] Render tiles beyond the antimeridian.
+ * @property {boolean} [wrapX=false] Render tiles beyond the antimeridian.
```
tschaub, line 34: This is a documentation bug.  There was no default applied by this class, and the parent class defaults to false.

src/ol/source/Tile.js
```diff
@@ -34,7 +34,7 @@ import {scale as scaleSize, toSize} from '../size.js';
  * @property {import("../proj.js").ProjectionLike} [projection] Projection.
  * @property {import("./State.js").default} [state] State.
  * @property {import("../tilegrid/TileGrid.js").default} [tileGrid] TileGrid.
- * @property {boolean} [wrapX=true] WrapX.
+ * @property {boolean} [wrapX=false] WrapX.
```
tschaub, line 37: This is also a documentation bug.  The tile source doesn't apply a default and the parent defaults to false 
https://github.com/openlayers/openlayers/blob/79b522990811e4ada0bb845e8c0920e20a772a89/src/ol/source/Source.js#L88-L92

test/rendering/cases/webgl-mixed-layers/main.js
```diff
@@ -46,6 +46,7 @@ new Map({
     }),
     new TileLayer({
       source: new DataTile({
+        wrapX: true,
```
tschaub, line 49: The WebGL tile layer renderer now will not fetch data outside the tile grid extent if `wrapX` is false.

ahocevar (approved):

PR #13070, opened 2021-12-01
Use source minzoom if not configured otherwise

ahocevar:
For vector tile sources that do not provide tiles from zoom 0, it is good practice to not show the layer when viewed at a lower zoom then the source's minzoom. Otherwise too many tiles will be loaded.

This pull request makes it so if neither `minZoom` nor `maxResolution` is specified on the `MapboxVector` layer, a `maxResolution` that matches the source's `minzoom` will be configured.

### DIFF
--- a/src/ol/layer/MapboxVector.js
+++ b/src/ol/layer/MapboxVector.js
@@ -197,9 +197,11 @@ const SourceType = {
  * @property {number} [minResolution] The minimum resolution (inclusive) at which this layer will be
  * visible.
  * @property {number} [maxResolution] The maximum resolution (exclusive) below which this layer will
- * be visible.
- * @property {number} [minZoom] The minimum view zoom level (exclusive) above which this layer will be
- * visible.
+ * be visible. If neither `maxResolution` nor `minZoom` are defined, the layer's `maxResolution` will
+ * match the style source's `minzoom`.
+ * @property {number} [minZoom] The minimum view zoom level (exclusive) above which this layer will
+ * be visible. If neither `maxResolution` nor `minZoom` are defined, the layer's `minZoom` will match
+ * the style source's `minzoom`.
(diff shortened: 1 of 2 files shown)


src/ol/layer/MapboxVector.js
```diff
@@ -197,9 +197,11 @@ const SourceType = {
  * @property {number} [minResolution] The minimum resolution (inclusive) at which this layer will be
  * visible.
  * @property {number} [maxResolution] The maximum resolution (exclusive) below which this layer will
- * be visible.
+ * be visible. If neither `maxResolution` nor `minZoom` are defined, the layer's `maxResolution` will
+ * match the source's `minZoom`
  * @property {number} [minZoom] The minimum view zoom level (exclusive) above which this layer will be
- * visible.
+ * visible. If neither `maxResolution` nor `minZoom` are defined, the layer's `maxResolution` will
```
tschaub, line 203:
```suggestion
 * visible. If neither `maxResolution` nor `minZoom` are defined, the layer's `minZoom` will
```
maybe?  seems fine either way

src/ol/layer/MapboxVector.js
```diff
@@ -197,9 +197,11 @@ const SourceType = {
  * @property {number} [minResolution] The minimum resolution (inclusive) at which this layer will be
  * visible.
  * @property {number} [maxResolution] The maximum resolution (exclusive) below which this layer will
- * be visible.
+ * be visible. If neither `maxResolution` nor `minZoom` are defined, the layer's `maxResolution` will
+ * match the source's `minZoom`
```
tschaub, line 201:
```suggestion
 * match the source's `minZoom`.
```

only if you're already making other changes

tschaub (approved):

PR #13083, opened 2021-12-04
Support expressions for band arguments

tschaub:
This makes it so arguments to the `band` operator can be expressions - allowing for things like style variables to be used in picking bands.


ahocevar (approved):
Very slick, thanks @tschaub!

PR #13095, opened 2021-12-06
Render is only complete when no tiles are queued

### LINKED ISSUE #13086: rendercomplete event fires too early
**Describe the bug**
When using a tile layer the rendercomplete event may fire before all tiles are drawn.
This happens when the last tile is loaded between the call to `renderFrame_` and the call to `handlePostRender`

**To Reproduce**
1. Go to https://codesandbox.io/s/vectortile-rendercomplete-s8dkl
2. Click on the map to zoom in
3. Check if some tiles are blurry when the alert is shown
4. Repeat a few times until it happens

**Expected behavior**
Tiles that are loaded after the map is rendered should not cause the `rendercomlete` event to fire.
The easiest solution might be to call `handlePostRender` without a timeout.

ahocevar:
Fixes #13086.

@M393, can you please confirm that this fixes the issue? If so, I'll try to add a regression test.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -208,6 +208,12 @@ class PluggableMap extends BaseObject {
 
     const optionsInternal = createOptionsInternal(options);
 
+    /**
+     * @private
+     * @type {boolean}
+     */
+    this.renderComplete_;
+
     /** @private */
     this.boundHandleBrowserEvent_ = this.handleBrowserEvent.bind(this);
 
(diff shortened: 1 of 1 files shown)


M393:
This does not work. The `rendercomplete` is fired prematurely because all tiles are loaded but because `handleRenderComplete` is called asynchronously, the tile queue state at that time does not match the state with which the last frame was rendered.

Saving the tile queue state during `renderFrame_` seems to work:
```js
frameState.hasTilesQueued = this.tileQueue_.getTilesLoading() > 0 || !this.tileQueue_.isEmpty();
```

ahocevar:
@M393 Thanks for the feedback and suggestion. I updated the pull request along these lines. Can you please test again?

M393 (approved):
Seems to work :+1:

ahocevar:
Unfortunately I was unable to create a regression test for this, so I'm marking this ready for review.

tschaub (approved):

PR #13101, opened 2021-12-09
Fix View set center to undefined

M393:
Calling `setCenter` with `undefined` failed if the view had a center because `equals` expects an array.


tschaub (approved):

tschaub:
Thanks, @M393

Issue #13074, opened 2021-12-02
When I turn around the world, shapes may disappear on zoom


Spencehiko:
When I turn around the world 1 or more times(to the left or the right doesn't matter), shapes may disappear on zoom in.

**To Reproduce**
[In this link](https://openlayers.org/en/latest/examples/draw-and-modify-features.html) you can reproduce. Just draw a shape after you turn around the world and zoom in.
![tmglMy0br1](https://user-images.githubusercontent.com/40501852/144380410-b50570d9-3564-4454-bf11-95a167af32fa.gif)

Is there a way to fix it?

Thanks in advance

jipexu:
yes a lot of strange and unstable behaviors in this example ....

M393:
You should only draw Features on the initially visible world. When you scroll to the next or previous world and draw there the checks whether a feature intersects the view may fail which causes it to not be rendered.

Maybe coordinates should by transformed by the Draw/Modify interaction to wrap within the configured view projection.

ahocevar:
OpenLayers can handle coordinates with latitudes from -540 to +540. To avoid surprises in the above example, you could configure the view with a restricted extent:
```js
const extent = get("EPSG:3857").getExtent().slice();
extent[0] += extent[0];
extent[2] += extent[2];
const map = new Map({
  layers: [raster, vector],
  target: "map",
  view: new View({
    center: [-11000000, 4600000],
    zoom: 4,
    extent
  })
});
```
This way, you can only draw geometries that will work in OpenLayers.

jipexu:
@ahocevar  thank for your light ! True if we stay in one world the example work fine ! 
However for me when we play with this example it remains confusing and therefore raises questions. Could we specify the limits you specify in the example documentation ?

ahocevar:
@jipexu I think it would make sense to modify that example to use a restricted extent like in my example above. Would you be willing to create a pull request?

jipexu:
good ! ok i will try...

jipexu:
arghh it start bad ... the example not work in codesandbox ...

jipexu:
https://deploy-preview-13077--ol-site.netlify.app/

Spencehiko:
@ahocevar thank you for your answer. 
>  you could configure the view with a restricted extent:

What if we use a different map service in OpenLayers? How can we restrict the map in this case?

Spencehiko:
Ok I got the solution

Spencehiko:
[...]

PR #13102, opened 2021-12-09
update line-arrows example

jipexu:
as same behavior as  #13074

Co-Authored-By: Andreas Hocevar <211514+ahocevar@users.noreply.github.com>


jipexu:
perhaps it will be good to modify some other example who show some drawing actions
as https://openlayers.org/en/latest/examples/measure.html and more ... ?

ahocevar (approved):

PR #13118, opened 2021-12-15
Add a layer opacity example

tschaub:
This adds an example showing how `layer.setOpacity()` can be used to adjust the layer opacity.

### DIFF
--- a/examples/layer-opacity.html
+++ b/examples/layer-opacity.html
@@ -1,0 +1,17 @@
+---
+layout: example.html
+title: Layer Opacity
+shortdesc: Adjust layer opacity based on user input
+docs: >
+  The `layer.setOpacity()` method can be called to adjust the opacity of a layer.
+tags: "opacity"
+cloak:
+  - key: get_your_own_D6rA4zTHduk6KOKTXzGB
+    value: Get your own API key at https://www.maptiler.com/cloud/
+---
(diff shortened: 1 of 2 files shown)


examples/layer-opacity.html
```diff
@@ -0,0 +1,17 @@
...
+tags: "opacity"
+cloak:
+  - key: get_your_own_D6rA4zTHduk6KOKTXzGB
+    value: Get your own API key at https://www.maptiler.com/cloud/
+---
+<div id="map" class="map"></div>
+<label>
+  Layer opacity
+  <input id="opacity-input" type="range" min="0" max="1" step="0.01" value="1" />
+  <span id="opacity-output"></span>
```
marcjansen, line 16:
```suggestion
  <span id="opacity-output">1.00</span>
```

marcjansen (approved):
Thanks. Initially the opacity isn't displayed, which is super minor. Please merge with or without taking the suggestion.

examples/layer-opacity.html
tschaub: Thanks, @marcjansen.  I was bugged by the same and added a call to `update` in the latest commit.

marcjansen: Same effect and fine for me.

PR #13120, opened 2021-12-15
Populate view's object properties

ahocevar:
Often it is useful to be able to create a new view with some properties from an existing view. Currently, calling `View#getProperties()` often returns an empty object, so this is not really possible. This pull request makes it so the the properties are populated with the options that the view currently maintains in the private `options_` member (which is removed), so users can obtain all options the view was constructed with, updated to the current view state.

### DIFF
--- a/test/browser/spec/ol/View.test.js
+++ b/test/browser/spec/ol/View.test.js
@@ -617,6 +617,19 @@ describe('ol/View', function () {
       expect(options.zoom).to.eql(10);
     });
 
+    it('returns the current properties with getProperties()', function () {
+      const view = new View({
+        center: [0, 0],
+        minZoom: 2,
+        zoom: 10,
+      });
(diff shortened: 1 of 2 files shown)


test/browser/spec/ol/View.test.js
```diff
@@ -617,6 +617,19 @@ describe('ol/View', function () {
       expect(options.zoom).to.eql(10);
     });
 
+    it('returns the current properties with getProperties()', function () {
```
tschaub, line 620: Is this really going to return the "current" properties?  It seems like instead it will return a mix of the initial and the current properties.  Like the initial `zoom` and the current `resolution`.

ahocevar: Good point! I changed things now so `zoom` is also updated as view property.

tschaub (approved):

PR #13126, opened 2021-12-18
Preserve the drawing buffer by default for WebGL layers

tschaub:
This should make it possible to get the canvas data for maps that have WebGL layers.  I haven't added any API methods to disable/enable this.  I think we should just enable it by default and see if we get any feedback that suggests that it should be easier to disable (if it has a negative performance impact on mobile, for example).  And I think it makes sense to have a release with this functionality before we build any other API methods on top of it (like `forEachLayerAtPixel`).

If `preserveDrawingBuffer` is true, we need to manually clear the canvas on the first draw.  A new lookup is added to the frame state to determine if the renderer is the first to draw to a canvas.

Without clearing the canvas first, here is what we get from the new rendering test (double rendering, improper opacity handling):

actual

And here is what we get when we do clear the canvas:

expected

Credit to @mike-000 for advocating for this.


ahocevar (dismissed):

ahocevar (approved):
Great work. I initially thought it breaks opacity handling while loading tiles, but everything works smoothly. Awesome, thanks!

PR #13127, opened 2021-12-18
Fix for range slider on IE

mike-000:
#13126 seems to be working well on all platforms for the heatmap, but the opacity example was not working on IE.


tschaub (approved):

tschaub:
We've discussed removing IE 11 workarounds ahead of the [June 15, 2022 end of support date](https://blogs.windows.com/windowsexperience/2021/05/19/the-future-of-internet-explorer-on-windows-10-is-in-microsoft-edge/).  So be aware that at some point ahead of that we'll stop accepting new workarounds.

Issue #13128, opened 2021-12-19
WebGLTile layer does not work with lazy source


mike-000:
**To Reproduce**
Change `ol/layer/Tile` to `ol/layer/WebGLTile` in the Lazy Source example, or any other example such as Print to Scale which uses a lazy source.

**Expected behavior**
Similar to `ol/layer/Tile`.

PR #13129, opened 2021-12-19
Use WebGL to avoid html2canvas rotation problem

### LINKED ISSUE #13125: html2canvas does not work correctly with rotated view
**Describe the bug**
Using `html2canvas` as an alternative to `dom-to-image-more` does not work with a rotated view.

**To Reproduce**
Open https://openlayers.org/en/v6.5.0/examples/print-to-scale.html and https://openlayers.org/en/latest/examples/print-to-scale.html rotate the views and export.  Compare the results.

**Expected behavior**
An export including the specified controls, and the rotated canvas correctly positioned.

mike-000:
Fixes #13125

WebGL canvases are not rotated with the view and now that `preserveDrawingBuffer` is set are a practical alternative to using `ol/layer/Tile` with `html2canvas`.  The code is restructured to avoid the lazy source issue #13128

### DIFF
--- a/examples/print-to-scale.js
+++ b/examples/print-to-scale.js
@@ -1,5 +1,5 @@
 import Map from '../src/ol/Map.js';
-import TileLayer from '../src/ol/layer/Tile.js';
+import TileLayer from '../src/ol/layer/WebGLTile.js';
 import View from '../src/ol/View.js';
 import WMTS, {optionsFromCapabilities} from '../src/ol/source/WMTS.js';
 import WMTSCapabilities from '../src/ol/format/WMTSCapabilities.js';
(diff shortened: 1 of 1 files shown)


ahocevar:
A more sustainable approach would be to try to fix this in html2canvas. I did not have time yet to investigate this closer, but it's interesting that a canvas with css rotation works in Safari, but does not work in Chrome.

ahocevar:
Looking at this closer, I tend to think it is a Chrome bug. I inspected the rotated canvas in the Dev Tools, went to the `transform: matrix(...` style, and modified one of the matrix values. From that moment on, the translation of the transform is lost, even when changing back to the original value. The result is the same as in the resulting pdf.

ahocevar:
And after removing the `canvas.style.all = 'unset';` line in `ol/dom.js#L30`, it works in Chrome too.

PR #13142, opened 2021-12-20
Include transpacific flights in Flight Animation example

mike-000:
The current example does not show any transpacific flights as it does not handle the multi-part geometries returned by arcGenerator, or animations in adjacent worlds.  This adds them.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #13144, opened 2021-12-20
Replace deprecated syntax in example

mike-000:
The `ol/Overlay` documentation describes `autoPan: true` as deprecated, but it is still being used in the Popup example.


ahocevar (approved):
Thanks, @mike-000

PR #13151, opened 2021-12-21
Avoid error accessing attributes when context is lost

mike-000:
This error occurs in codesandbox with latest version when another example is opened causing context to be lost (I saw similar error when testing the `getDataAtPixel` method in #13000).
image


tschaub (approved):

tschaub:
I think this just delays the failure.  I imagine we'll have more work to do to handle lost contexts.

PR #13154, opened 2021-12-23
Update Export Map example to handle WebGL

mike-000:
This is the example update extracted from #13000 as separate PR.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #13156, opened 2021-12-24
Ensure that tile range covers all pixels

tschaub:
This change makes it so the `tileGrid.getTileRangeForExtentAndZ(extent, z)` method generates a tile range that covers all the pixels in the provided extent.  The old logic did not cover all fractional pixels.  The difference becomes apparent when you have a source with a limited number of zoom levels and zoom way beyond this level.

The new rendering test demonstrates the change.  The map background is red.  The tiles are green.  Ideally, we see no red.  The green source only has tiles at levels 0 and 1.  The map is zoomed into level 8.

Before this change, when the edge of a tile approaches the edge of the viewport, the adjacent tiles are not rendered.  This is because the "covered pixel" calculation is performed at the source resolution.  When the view is zoomed way beyond the source max zoom, a single source-resolution pixel may be many pixels at the rendered resolution.  So we want to include a tile if any fraction of the source pixel is included in the rendered extent.

Before

before

After

expected


ahocevar (approved):

tschaub:
Having some deja vú writing this.  This change essentially reverts 700903ca5c37147ab0a61ef639b8f3d2a7c87c9f to restore the logic added in #207.  The difference this time is that we use the limited precision floor/ceil rounding functions.

PR #13158, opened 2021-12-26
Add example for dynamic clusters

ahocevar:
This pull request adds an example showing functionality similar to Leaflet's [Leaflet.markercluster](http://leaflet.github.io/Leaflet.markercluster/) plugin.


ahocevar:
https://deploy-preview-13158--ol-site.netlify.app/examples/clusters-dynamic.html is the new example.

tschaub:
clusters

It looks like the convex hulls render under the cluster markers.  Just want to confirm this is what it is supposed to look like.

examples/clusters-dynamic.html
```diff
@@ -0,0 +1,16 @@
+---
+layout: example.html
+title: Dynamic clusters
+shortdesc: Clusters with zoom-to-cluster, hull view, and uncluttering of overlapping features.
+docs: >
+  <p>This example shows open data of subsidized photovoltaic installations in Vienna. Different style functions
+    are used for cluster display, single feature display, convex hull of a cluster, and an uncluttered view of
+    overlapping features. Hovering over a cluster shows its convex hull. Clicking on a cluster zooms to the
+    extent of the contained features. Clicking on a cluster consisting of features that are very close to each other reveals an uncluttered view
```
tschaub, line 9: Did you mean "unclustered" instead of "uncluttered" (above as well)?

tschaub:
> It looks like the convex hulls render under the cluster markers. Just want to confirm this is what it is supposed to look like.

I see the same on the https://leaflet.github.io/Leaflet.markercluster/example/marker-clustering-realworld.388.html example.  And I see it looks nicer when zoomed out more than in my example above.

examples/clusters-dynamic.html
ahocevar: Not really "unclustered". It's a view for features that share the same location (within one pixel), or are still clustered at the view's `maxZoom`. How would you describe that?

Untitled

The convex hulls are supposed to be rendered below the cluster circle.

ahocevar:
I made the cluster circles a bit transparent, so the convex hulls shine through.

examples/clusters-dynamic.html
tschaub: Ah, cool.  How about an “expanded” view?

ahocevar: Good suggestion, thanks! I changed all occurrences to "expanded".

tschaub (approved):

tschaub:
Nice example.

jipexu:
👍 so good ! thanks

PR #13171, opened 2021-12-29
Correct scale bar in EPSG:4326

### LINKED ISSUE #13167: Scale bar shows incorrect scale when not centered on equator in EPSG:4326
**Describe the bug**
The scale bar's use of point resolution in degrees and meters per degree displays an unexpected scale value in EPSG:4326.

**To Reproduce**
Open https://codesandbox.io/s/scale-line-forked-8bfmx?file=/main.js and use the projection selector to change between projections while maintaining the same point resolution in meters,  When the projection changes the scale bar length remains unchanged as expected but the scale displayed above it does change, which is unexpected as length and scale are interdependent.

**Expected behavior**
Uncomment to the patch for `getScaleForResolution` which calculates scale based on point resolution directly in meters instead of resolution in degrees multiplied by meters per degree.  Change projections again and the scale changes only after the fifth digit, a small difference which might be expected considering the transforms and spherical distance calculations needed to obtain point resolution for EPSG:4326 (there is a small difference between the radius 6370997 used in `ol/proj/Units` and 6371008.8 used in `ol/sphere` which might contribute to this).

mike-000:
Fixes #13167

Corrects the incorrect scale shown on scale bar in EPSG:4326 and adds a new example demonstrating how to maintain scale when changing projection.

As noted in https://github.com/openlayers/openlayers/issues/13167#issuecomment-1002588857 there are small differences due to the accuracy of calculations in `getPointResolution` but that should be dealt with as a separate issue where changes could have consequences elsewhere.

### DIFF
--- a/examples/projection-and-scale.html
+++ b/examples/projection-and-scale.html
@@ -1,0 +1,18 @@
+---
+layout: example.html
+title: Projection and Scale
+shortdesc: Example of maintaining scale when changing projection.
+docs: >
+  Example of maintaining scale when changing projection.
+  `getPointResolution()` is used to calculate the resolution for the
+  new projection which corresponds to that for the old projection. 
+tags: "projection, scale"
+---
+<div id="map" class="map"></div>
+<div>
+  <label for="view-projection">View projection</label>
+  <select id="view-projection">
+    <option value="EPSG:3857">Spherical Mercator (EPSG:3857)</option>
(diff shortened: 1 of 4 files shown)


ahocevar (commented):
The code change looks good. Thanks for your work on this, @mike000. I only wonder if the example could be simplified to just show a projection change. No switching of scale bar to scale line etc.. Also, a unit test would be important to avoid future regressions.

ahocevar (approved):
Thanks, @mike-000

PR #13173, opened 2021-12-30
Refine offset used for point resolution calculation

mike-000:
Fixes the inconsistencies noted in https://github.com/openlayers/openlayers/issues/13167#issuecomment-1002588857

Calculating point resolution as the average resolution across a distance offset from the point is more precise when the offset is closer to the point.  An offset of 0.5 pixel (distance range of 1 pixel) gives a results consistent to better than 5 digits, but an offset of 0.01 pixel gives a consistent result in a scale bar at OSM zoom level 0 (about 1 : 500,000,000)

### DIFF
--- a/src/ol/proj.js
+++ b/src/ol/proj.js
@@ -216,20 +216,44 @@ export function getPointResolution(projection, resolution, point, opt_units) {
         // no transform is available
         pointResolution = resolution * projection.getMetersPerUnit();
       } else {
-        let vertices = [
-          point[0] - resolution / 2,
-          point[1],
-          point[0] + resolution / 2,
-          point[1],
-          point[0],
-          point[1] - resolution / 2,
-          point[0],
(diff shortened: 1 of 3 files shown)


ahocevar:
I doubt that using such a small distance would work well for high zooms. Since scale is a concept that does not work well on web maps anyway, I need more to be convinced that this is a good change, otherwise I'd be in favor of leaving things as they are. At least a test with a scale confirmed with a spherical measurement on a high zoom level, e.g. 22, would be good.

mike-000:
There is no problem at high zoom levels as point resolution does not change over those distances, so I think it would be better to base the reduction on distance.  Assuming that spherical measurement remains reasonably accurate down to 1km, and that point resolution does not change significantly across 1km, if both dimensions of the box exceed that reduce the offset so the smallest dimension is around 1km.

PR #13174, opened 2021-12-30
Use dynamic import for geotiff

### LINKED ISSUE #13046: Cloud Optimized GeoTIFF example is not working on codesandbox
Cloud Optimized GeoTIFF [example](https://openlayers.org/en/latest/examples/cog.html) is not working on [codesandbox](https://codesandbox.io/s/owk4e). I [filed](https://github.com/codesandbox/codesandbox-client/issues/6313) an issue at codesandbox repository too because it's not clear on which side the problem is.

ahocevar:
This pull request is an attempt to avoid treeshaking problems on
```js
import { AnySource } from 'ol/source';
```
I'm hoping that this fixes #13046, at least for the case where `ol/source/GeoTIFF` is not actually used.

### DIFF
--- a/src/ol/worker/geotiff-decoder.js
+++ b/src/ol/worker/geotiff-decoder.js
@@ -1,16 +1,14 @@
 import 'regenerator-runtime/runtime.js';
 
-// BEGIN copied from geotiff/src/decoder.worker.js
 import {Transfer, expose} from 'threads/worker';
-import {getDecoder} from 'geotiff/src/compression';
 
 async function decode(fileDirectory, buffer) {
+  const {getDecoder} = await import('geotiff/src/compression');
   const decoder = await getDecoder(fileDirectory);
   const decoded = await decoder.decode(fileDirectory, buffer);
   return Transfer(decoded);
 }
 
 expose(decode);
-// END copied from geotiff/src/decoder.worker.js
 
 export let create;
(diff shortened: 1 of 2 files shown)


src/ol/worker/geotiff-decoder.js
```diff
@@ -1,16 +1,14 @@
 import 'regenerator-runtime/runtime.js';
 
-// BEGIN copied from geotiff/src/decoder.worker.js
 import {Transfer, expose} from 'threads/worker';
-import {getDecoder} from 'geotiff/src/compression';
 
 async function decode(fileDirectory, buffer) {
+  const {getDecoder} = await import('geotiff/src/compression');
```
ahocevar, line 6: This change does not directly relate to the goal of this pull request, but it reduces the size of the worker a bit. (90 kB vs. previously 92 kB).

ahocevar:
I'm not sure this change is something we really want, so I won't complain if it does not get approved.

ahocevar:
I'm closing this. I don't think it is the right approach for dealing with problems caused by a dependency. I could think of three alternatives:
1. Import untranspiled sources from `geotiff`. The disadvantage is that users need extra configuration when targeting legacy browsers.
2. Repackage a transpiled version of only what we need from `geotiff`, and make that part of the package.
3. Contribute to `geotiff` to make the package easier to use with node and browsers, like we did in the past.

In the long run, 3. will probably be the best approach.

PR #13176, opened 2021-12-31
Add `attributions` and `attributionsCollapsible` options to `ol/source/DataTile`

### LINKED ISSUE #13170: `ol/source/DataTile` attributions
`ol/source/DataTile` does not support an `attributions` option in the constructor, so attributions must be set using `setAttributions()`.  Is that an oversight or intentional?

mike-000:
Fixes #13170


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #13177, opened 2021-12-31
Use ecoregions data

### LINKED ISSUE #12940: Update country names and borders in example

tschaub:
This branch replaces the use of the current countries dataset with data from the [RESOLVE Ecoregions 2017](https://academic.oup.com/bioscience/article/67/6/534/3102935) dataset.

See the updated examples.  I'm curious if people think these load or render too slowly.
 * [vector layer](https://deploy-preview-13177--ol-site.netlify.app/examples/vector-layer.html)
 * [box selection](https://deploy-preview-13177--ol-site.netlify.app/examples/box-selection.html)
 * [export map](https://deploy-preview-13177--ol-site.netlify.app/examples/export-map.html)
 * [geojson-vt](https://deploy-preview-13177--ol-site.netlify.app/examples/geojson-vt.html)
 * [select features](https://deploy-preview-13177--ol-site.netlify.app/examples/select-features.html)
 * [hit detection](https://deploy-preview-13177--ol-site.netlify.app/examples/hitdetect-vector.html)
 * [image vector](https://deploy-preview-13177--ol-site.netlify.app/examples/image-vector-layer.html)
 * [Sphere Mollweide](https://deploy-preview-13177--ol-site.netlify.app/examples/sphere-mollweide.html)

The new https://github.com/openlayers/data repo is a place where we can host data for our examples.  Yes, data in git is not fun, but these are the benefits:

 * Commits to that repo are hosted at https://openlayers.org/data/*
 * CORS
 * gzip
 * We can make tweaks to the data before hosting

Fixes #12940
Closes #12941
Closes #12942


mike-000:
In a map export there is no reason (apart from a very un-obvious assertion if you don't set a `className` on one of the layers) why you couldn't have a semi-opaque vector layer with `background` over another layer as in https://codesandbox.io/s/export-map-forked-cmdk1?file=/main.js  The `backgroundColor` should therefore be applied to the `mapContext` after the `globalAlpha` is set.  I think it would be helpful to keep a tile layer to demonstrate that.

tschaub:
> The `backgroundColor` should therefore be applied to the `mapContext` after the `globalAlpha` is set.

Good point.  I adjusted the example to do this.

> I think it would be helpful to keep a tile layer to demonstrate that.

I don't agree with this part.  I think it is unfortunate that our examples tend to get more complex and uglier over time, as people add new functionality or fix issues and tack features on to existing examples as a way to demonstrate the change to themselves.  While the changes likely make sense to the author, I don't think this makes the examples more useful.

I can imagine it would be useful to add a `map.export({type, quality})` method (that would throw in cases where we cannot generate a data URL).  Then we could push the complexity from this example into that method.  And we could have tests that assert that things work with layer background, opacity, etc.

ahocevar (approved):
Thanks for this huge effort, @tschaub!

Issue #13114, opened 2021-12-14
package.json: type module, but no exports map


dummdidumm:
**Describe the bug**
The ol package specifies in its `package.json` that it's of type module. This means that any folder-style imports like `import Map from 'ol/Map';` are prohibited when using it inside a project that itself is of type module. According to the documentation, they should work.

**To Reproduce**
1. Clone https://github.com/maxicarlos08/sveltekit-ol
2. `npm ci`
3. `npm run build`
4. See error about an invalid directory import

**Expected behavior**
Folder imports should work.

To make them work, you need to add an exports map to your `package.json`. Beware though, this means that anything not listed in the exports map is no longer accessible from the outside, so strictly speaking this would be a breaking change.

cc @maxicarlos08 who opened the issue over at https://github.com/sveltejs/kit/issues/3037

tschaub:
OpenLayers doesn’t do any “folder” imports (or at least it shouldn’t).  It looks like your `ol/Map` import without a file extension might be interpreted as a folder import.  What is the error if you use the appropriate file extension?

dummdidumm:
Yes it's interpreted as a folder import, because if one uses `"type": "module"` in its own `package.json` one has to explicitely add file extensions to the import (and shortcuts for `index.js` imports also don't work). This is then likely more of a documentation issue or a gotcha that may need to be added somewhere.
I don't use OpenLayers myself, rather I opened this issue on behalf of @maxicarlos08 who opened an issue over at SvelteKit, where we think it's an issue with OpenLayers or wrong usage of it. It seems that when adding the file extension one gets the following error: https://github.com/openlayers/openlayers/issues/10821#issuecomment-991738508

maxicarlos08:
Might this also be an issue with `geotiff` (which got an upgrade recently)?

tschaub:
As mentioned in sveltejs/kit#3037, here is a working `index.svelte` route with no modifications to the `ol` package (or any other):
```html
<script>
  import "ol/ol.css"
  import Map from 'ol/Map.js';
  import View from 'ol/View.js';
  import OSM from 'ol/source/OSM.js';
  import Tile from 'ol/layer/Tile.js';
  import {browser} from '$app/env';
  import {onMount} from 'svelte';

  let target;
  let map;

  onMount(() => {
    if (browser) {
      const view = new View({center: [0, 0], zoom: 2});
      const layer = new Tile({source: new OSM()});
      map = new Map({target, view, layers: [layer]});
    }
[...]

PR #13180, opened 2022-01-01
Always use ES modules from geotiff.js

ahocevar:
When geotiffjs/geotiff.js#267 is released and both this pull request and #13179 are merged, server side rendering scenarios and other uses of OpenLayers with Node will work even with imports from `ol/source.js`. See https://github.com/openlayers/openlayers/issues/13114#issuecomment-994134162.


tschaub (approved):

tschaub:
Thanks for digging into this, @ahocevar.

PR #13186, opened 2022-01-03
Add getDataAtPixel() method for WebGL

mike-000:
This is the getDataAtPixel() update extracted from #13000 plus some corresponding tests.


tschaub:
Thanks for proposing this, @mike-000.  I thought it made sense to see how things performed using the WebGL context's `readPixels` method instead of this approach of using a second 2d context.  I opened #13192 with the alternative approach.

tschaub (approved):

Issue #13189, opened 2022-01-04
Load GeoTiff from Blob


m-mohr:
**Is your feature request related to a problem? Please describe.**
I'd like to show a GeoTiff that I only have available Blob (i.e. a dynamically generated GeoTiff that is returned as a result of a HTTP POST request). I see that fromBlob is available in geotiff.js so I'm wondering whether it could also be made available here.

**Describe the solution you'd like**
Instead of passing `url` to the GeoTiff source, I'd like to pass a Blob, e.g. in a new parameter called `blob`, and display it on a Map.

marcjansen:
I would love to have this in the library, but would like to hear other peoples opinion. I am not so sure how tight we are coupling geotiff.js and OpenLayers (not that this was introduced here, just a random observation)

An example and a test would be nice.

m-mohr:
I've created a draft PR with more details and an example for now: https://github.com/openlayers/openlayers/pull/13191

ahocevar:
Is there a reason for not simply configuring the GeoTIFF source with `url: URL.createObjectUrl(blob)`?

m-mohr:
@ahocevar I must admit I directly found fromBlob in the geotiff.js source and didn't even think about the object URLs.
I've tested them and it also works for me, but I can't really judge whether there are serious pros and cons (e.g. for larger blobs regarding memory management). What I've seen in the dev console is that for `URL.createObjectUrl(blob)` multiple "requests" appear in the network pane, but only one for the PR'ed solution.  Also, MDN seems to indicate that URL.createObjectUrl was later implemented in Browsers than the FileReader that is used by geotiff.js.

Other than that, OpenLayers should probably give an example or improve docs a bit as this potential solution is not very obvious, I'd say.

m-mohr:
Another difference I found is that createObjectUrl seems to be synchronous and memory needs to be freed with revokeObjectURL while FileReader is asynchronous.

mike-000:
> Also, MDN seems to indicate that URL.createObjectUrl was implemented in Browsers later than the FileReader that is used by geotiff.js.

I don't think that is significant.  For example, GeoTIFF needs a minimum of Chrome 45 for the workers to run.  `FileReader` and `URL.createObjectUrl` were both supported long before that.

yonda-yonda:
I'm looking forward to this feature.
Glad to see the discussion is active again.

ahocevar:
[...]

PR #13191, opened 2022-01-04
Load GeoTiff from Blob

m-mohr:
Implements loading GeoTiffs from Blobs, implements #13189

Any feedback is appreciated.

PS: This is my first contribution, I hope I did not make any obvious mistakes or so...


mike-000:
There is a problem with your hosted file used in the example, it does not work even when used as a url in the current COG example - the same assertion as logged in https://deploy-preview-13191--ol-site.netlify.app/examples/cog-blob.html   However, the file used in the original example does work with your update when loaded as a blob, but takes several seconds longer to load and process than when specified as a url.

m-mohr:
Ah, thanks for the hint. Forgot the s in the protocol which made it load only locally. Fixed.

mike-000:
The new file downloads at 460KB which would be small enough to be included to the `example/data` folder if you prefer not to host it.  Those files get copied to CodeSandbox when the Edit button in the examples is used.  `.tiff` is supported but the regex at https://github.com/openlayers/openlayers/blob/main/examples/resources/common.js#L12 would need to be `tiff?` to include `.tif`.

m-mohr:
Thanks, I've updated the commit to include the example .tif file.

mike-000:
Did you intend to leave the `Loading...` in the map div? https://deploy-preview-13191--ol-site.netlify.app/examples/cog-blob.html

m-mohr:
That was a very naive attempt to not just show an empty area by default, but that did not work as expected so I'll remove that tomorrow.

examples/cog-blob.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+
+    const map = new Map({
+      target: 'map',
+      layers: [
+        new TileLayer({
+          source: source,
+        }),
+      ],
+      view: source.getView(),
```
MoonE, line 23: Looks like the example is suffering from #10508 because an extent is set and only one resolution is configured.
```suggestion
      view: source.getView().then(function (viewConfig) {
        viewConfig.showFullExtent = true;
        return viewConfig;
      }),
```
Alternatively you could remove the `resolutions` config:
```patch
-      view: source.getView(),
+      view: source.getView().then(function (viewConfig) {
+        delete viewConfig.resolutions;
+        return viewConfig;
+      }),
```

m-mohr: Yeah, I was wondering where the flickering comes from. Thanks for the hint, will fix the example tomorrow.

m-mohr: Fixed the example.

m-mohr:
I fixed the example. It's not containing the "Loading..." anymore and I also removed the "flickering" thanks to the comments from @MoonE.

marcjansen (approved):

marcjansen:
I really like this, thanks for this @m-mohr & @MoonE. I'd really like a second opinion from either @ahocevar or @tschaub, if possible.

tschaub:
Looks good to me as well.  I wanted to tweak [the example](https://deploy-preview-13191--ol-site.netlify.app/examples/cog-blob.html) a bit to make it possible to zoom in a bit - but haven't gotten to it yet.

m-mohr:
@ahocevar raised an interesting point for consideration here: https://github.com/openlayers/openlayers/issues/13189#issuecomment-1006752097
Maybe someone has a better idea of the differences between the two methods?

examples/cog-blob.js
tschaub: I haven't tried, but assume we could instead just add more resolutions.

```js
source.getView().then(config => (
  {
    ...config
    resolutions: new Array(4).fill(0).map((v, i) => config.resolutions[0] / Math.pow(2, i))
  }
))
```

Or similar

tschaub:
Thanks for the proposed contribution, @m-mohr.  Let's revisit this if we find a case that is not handled well with the current options.

src/ol/source/GeoTIFF.js
```diff
@@ -24,7 +25,8 @@ import {fromCode as unitsFromCode} from '../proj/Units.js';
 /**
  * @typedef {Object} SourceInfo
  * @property {string} url URL for the source GeoTIFF.
- * @property {Array<string>} [overviews] List of any overview URLs.
+ * @property {Blob} blob Blob containing the source GeoTIFF. `blob` and `url` are mutually exclusive.
```
ahocevar, line 28: Both `url` and `blob` should now be optional.

m-mohr: Adding the [ and ] brackets was enough, right? Is now part of the new PR #13724

PR #11255, opened 2020-07-07
GetDataAtPixel performance improvement

ahocevar:
Closes #11235

Issue #11235, opened 2020-07-03
CanvasLayerRenderer.prototype.getDataAtPixel - better performance result


Shekenix:
This is my second task about this feature. Please check it.
You should improve your existing CanvasLayerRenderer.prototype.getDataAtPixel function.
In the function you get image data by whole context. This solution decrease performance due to scanning of all pixels. 

``` javascript
CanvasLayerRenderer.prototype.getDataAtPixel = function (pixel, frameState, hitTolerance) {
        var renderPixel = applyTransform(this.inversePixelTransform, pixel.slice());
        var context = this.context;
        var data;
        try {
            data = context.getImageData(Math.round(renderPixel[0]), Math.round(renderPixel[1]), 1, 1).data;
        }
        catch (err) {
            if (err.name === 'SecurityError') {
                // tainted canvas, we assume there is data at the given pixel (although there might not be)
                return new Uint8Array();
            }
            return data;
        }
        if (data[3] === 0) {
            return null;
        }
        return data;
    };
```

To achieve a better performance, first you should crop canvas to 1px image (drawing new 1px image) and finally get that pixel using getDataImage:

``` javascript
CanvasLayerRenderer.prototype.getDataAtPixel = function (pixel, frameState, hitTolerance) {
        var renderPixel = applyTransform(this.inversePixelTransform, pixel.slice());
        var context = this.context;
        var data;
        try {
            var x = Math.round(renderPixel[0]);
            var y = Math.round(renderPixel[1]);
            var newCanvas = document.createElement("canvas");
            var newContext = newCanvas.getContext('2d');
            newCanvas.width = 1;
            newCanvas.height = 1;
            newContext.clearRect(0, 0, 1, 1);
            newContext.drawImage(context.canvas, x, y, 1, 1, 0, 0, 1, 1);
            data = newContext.getImageData(0, 0, 1, 1).data;
            //data = context.getImageData(Math.round(renderPixel[0]), Math.round(renderPixel[1]), 1, 1).data;
        }
        catch (err) {
            if (err.name === 'SecurityError') {
                // tainted canvas, we assume there is data at the given pixel (although there might not be)
                return new Uint8Array();
            }
            return data;
        }
        if (data[3] === 0) {
            return null;
        }
        return data;
    };
```

Please, improve this function. Performance result is much better.

Regards.

ahocevar:
[...]

PR #13192, opened 2022-01-05
Support getDataAtPixel for WebGL layer renderers

tschaub:
This makes it possible to get data from WebGL layers using `map.forEachLayerAtPixel()`.

As noted in the docs, this method will generally *not* return data for a single layer (for Canvas or WebGL layers) when there is more than one layer on a map.  I think this makes the method a bit of a hazard.  We could have an event-based mechanism for getting data for one layer at a time that would disable context sharing or otherwise return reliable results.  If we do add this, I think it would make sense to remove the `map.forEachLayerAtPixel()` method in the next major release.  Until then, people need to stumble across the doc that suggests that they add a unique `className` to each layer (not very guessable).

This is an alternative implementation to #13186, using the [`readPixels`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/readPixels) method on the WebGL rendering context instead of drawing to a secondary 2d context and then calling [`getImageData`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData).  I'm not sure what kind of profiling went into #11235 or #11255, but it looks like a different approach makes sense for the WebGL renderer.

Here is the average time for `forEachLayerAtPixel` in this branch and in the branch for #13186 over 342000 iterations.

| method  | time (ms) |
|---------|-----------|
| `getImageData` (#13186) | 0.69657 |
| `readPixels` (this branch) | 0.04389 |

Both fast enough really, but it looks like the `readPixels` approach is 15X faster.

Here is the CPU profile for the `getImageData` approach (in #13186)

getImageData

And here is the CPU profile for the `readPixels` approach (this branch)

readPixels

Credit to @mike-000 for pushing for this and for the tests.  I removed the test that overrides the `HTMLCanvasElement.prototype.getContext` method to set `preserveDrawingBuffer: false` since we don't provide an API method to do that.


mike-000:
I had also done a comparison https://codesandbox.io/s/getdataatpixel-tests-tpofm?file=/main.j and found the `drawImage` method to be consistently faster.  The larger the map the slower the `readPixels` methods becomes:

With the test opened in a maximized new window on 1920px wide monitor, Chrome on Windows 10, times in milliseconds for 1024 lookups:
```
drawImage 2095
readPixels 10948
drawImage 1762
readPixels 10889
drawImage 1803
readPixels 10724
drawImage 1747
readPixels 10735
drawImage 1729
readPixels 11068
drawImage 1796
readPixels 11051
drawImage 1738
readPixels 10994
drawImage 1786
readPixels 10958
drawImage 1674
readPixels 10743
drawImage 1721
readPixels 10822
```

The same applies with `getImageData()` on large 2d canvases which was the reason for #11255

Although I tested by setting `preserveDrawingBuffer: false` my check for unreadable data also handles lost context  as `getContextAttributes()` returns null when context is lost.

tschaub:
Interesting.  Maybe a hardware issue.  I'm seeing `readPixels` consistently ~15X faster than `drawImage/getImageData` independent of the canvas size.

Here are results from a 2560 x 1336 canvas with 213760 samples

| method | elapsed | average |
|---------|---------|---------|
| `getImageData` | 152.4 s | 0.7130 ms |
| `readPixels` | 9.0 s | 0.0422 ms |

And the associated profiles

For the `getImageData` method

image

And for the `readPixels` method

image

This is on Chrome 96 on an M1 Mac.

Here is the example I'm using to generate these profiles:
`get-data.html`
```html
---
layout: example-verbatim.html
title: Get Data Example
shortdesc: Profiling the map.forEachLayerAtPixel method.
docs: >
  In the console, run the `time()` function.
tags: "data tiles"
---
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="chrome=1">
    <title>Data Tiles</title>
    <link rel="stylesheet" href="../css/ol.css" type="text/css">
    <style type="text/css">
      html, body, .map {
        margin: 0;
        padding: 0;
        width: 100%;
        height: 100%;
      }
    </style>
  </head>
  <body>
    <script src="common.js"></script>
    <script src="get-data.js"></script>
  </body>
</html>
```

`get-data.js`
```js
import DataTile from '../src/ol/source/DataTile.js';
import TileLayer from '../src/ol/layer/WebGLTile.js';

const size = 256;

const canvas = document.createElement('canvas');
canvas.width = size;
canvas.height = size;

const context = canvas.getContext('2d');
context.strokeStyle = 'white';
context.lineWidth = 4;
context.textAlign = 'center';
context.font = '24px sans-serif';
const lineHeight = 30;

const map = new Map({
  target: 'map',
  layers: [
    new TileLayer({
      source: new DataTile({
        interpolate: true,
        loader: function (z, x, y) {
          const half = size / 2;
          context.clearRect(0, 0, size, size);
          const gradient = context.createRadialGradient(
            half,
            half,
            0,
            half,
            half,
            half
          );
          gradient.addColorStop(0, 'red');
          gradient.addColorStop(1, 'green');
          context.fillStyle = gradient;
          context.fillRect(0, 0, size, size);
          context.fillStyle = 'white';
          context.fillText(`z: ${z}`, half, half - lineHeight);
          context.fillText(`x: ${x}`, half, half);
          context.fillText(`y: ${y}`, half, half + lineHeight);
          context.strokeRect(0, 0, size, size);
          const data = context.getImageData(0, 0, size, size).data;
          // converting to Uint8Array for increased browser compatibility
          return new Uint8Array(data.buffer);
        },
        // disable opacity transition to avoid overlapping labels during tile loading
        transition: 0,
      }),
    }),
  ],
  view: new View({
    center: [0, 0],
    zoom: 0,
  }),
});

function time() {
  const size = map.getSize();
  const px = [0, 0];
  const noop = () => {};
  let count = 0;
  console.profile('forEachLayerAtPixel');
  const start = performance.now();
  for (let x = 0; x < size[0]; x += 4) {
    for (let y = 0; y < size[1]; y += 4) {
      px[0] = x;
      px[1] = y;
      map.forEachLayerAtPixel(px, noop);
      ++count;
    }
  }
  const end = performance.now();
  console.profileEnd('forEachLayerAtPixel');
  const elapsed = end - start;
  console.log('count', count);
  console.log('elapsed (ms)', elapsed);
  console.log('average (ms)', elapsed / count);
}

window.time = time;
```

Then I switch between the branch for #13186 and this branch, load the example, and run `time()` in the console.

mike-000:
t seems to be a Windows or hardware, and browser combination (Chromium-based, as IE is not affected) possibly related to https://stackoverflow.com/questions/54076117/readpixels-performance-what-makes-it-so-slow  The Mac average of 0.7ms for `drawImage/getImageData` beats my PC's 1.7ms for Chrome (IE is faster) which is still acceptable for use with `pointermove` events.  On a very small map `readPixels` would be faster with any combination, but with very large maps average times up to 15ms would be unusable with `pointermove`.

**Chrome 13186**

map size 256,256
count 1024
elapsed (ms) 1777.7000000476837
average (ms) 1.7360351562965661

map size 2560,1336
count 1024
elapsed (ms) 1749.5
average (ms) 1.70849609375

**Chrome 13192**

map size 256,256
count 1024
elapsed (ms) 503.90000009536743
average (ms) 0.49208984384313226

map size 2560,1336
count 1024
elapsed (ms) 12014.399999976158
average (ms) 11.732812499976717

**Edge 13186**

map size 256,256
count 1024
elapsed (ms) 1674.3999999761581
average (ms) 1.635156249976717

map size 2560,1336
count 1024
elapsed (ms) 1742.2000000476837
average (ms) 1.7013671875465661

**Edge 13192**

map size 256,256
count 1024
elapsed (ms) 506.10000002384186
average (ms) 0.49423828127328306

map size 2560,1336
count 1024
elapsed (ms) 15866.5
average (ms) 15.49462890625

**IE 13186**

map size 256,256
count 1024
elapsed (ms) 703.3000000000011
average (ms) 0.6868164062500011

map size 2560,1336
count 1024
elapsed (ms) 706.3000000000011
average (ms) 0.6897460937500011

**IE 13192**

map size 256,256
count 1024
elapsed (ms) 277.7000000000007
average (ms) 0.2711914062500007

map size 2560,1336
count 1024
elapsed (ms) 284.10000000000036
average (ms) 0.27744140625000035

ahocevar:
For me, on an 2017 Intel MacBook Pro, with both @mike-000's and @tschaub's benchmark, in Both Chrome and Safari, this pull request yields slower results than #13186 on maps with a size of approximately 1200x700 pixels.

mike-000:
One option would be to determine the best method for the system when WebGL is first required, then selectively continue to use that.  The initial overhead should be less than 200ms.

```
export const getDataAtPixelMethod {
  READPIXELS: 1,
  DRAWIMAGE: 2,
};

let getDataAtPixelMethod_;

export function getGetDataAtPixelMethod () {
  if (getDataAtPixelMethod_ === undefined) {
    const iMax = 1024;
    const glCanvas = document.createElement('canvas');
    glCanvas.width = iMax;
    glCanvas.height = iMax;
    const gl = getContext(glCanvas);
    gl.clearColor(255.0, 255.0, 255.0, 1.0);
    gl.clear(gl.COLOR_BUFFER_BIT);

    const pixelCanvas = document.createElement('canvas');
    pixelCanvas.width = 1;
    pixelCanvas.height = 1;
    const pixelContext = pixelCanvas.getContext('2d');
      
    let data;
    const ii = Math.floor(iMax / 5);

    const startReadPixels = Date.now();
    for (let i = ii; i < iMax; i += ii) {
      data = new Uint8Array(4);
      gl.readPixels(i, i, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, data);
    }
    const timeReadPixels = Date.now() - startReadPixels;

    const startDrawImage = Date.now();
    for (let i = ii; i < iMax; i += ii) {
      pixelContext.clearRect(0, 0, 1, 1);
      pixelContext.drawImage(gl.canvas, i, i, 1, 1, 0, 0, 1, 1);
      data = pixelContext.getImageData(0, 0, 1, 1).data;
    }
    const timeDrawImage = Date.now() - startDrawImage;

    getDataAtPixelMethod_ = timeReadPixels < timeDrawImage
      ? getDataAtPixelMethod.READPIXELS
      : getDataAtPixelMethod.DRAWIMAGE;
  }
  return getDataAtPixelMethod_;
}
```

tschaub:
I was thinking the same, @mike-000.  But it may be too fiddly to get right.  I imagine that eventually, `readPixels` will be faster.  But based on your experience and @ahocevar, it sounds like we should go with #13186.  I’d like to apply the doc change from this branch.

PR #13200, opened 2022-01-08
Use interpolate option in Sea Level example

mike-000:
The setting of `imageSmoothingEnabled` in a layer `prerender` event pre-dates the `imageSmoothing`/`interpolate` option which would make the example simpler and consistent with other examples which now use that option.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #13202, opened 2022-01-08
API method to generate single canvas for map exports

mike-000:
Adds a `map.getCompositeCanvas()` method, initially based on the existing examples (plus one bug fix), to make map exports simpler and more maintainable, similar to those in OpenLayers 5.


ahocevar:
Do we really want to commit to this way of exporting a map? I have made very good experience with [html2canvas](https://html2canvas.hertzen.com), and did not run into any limitations with any OpenLayers map, including controls and overlays. We also use it in the `print-to-scale.html` example.

Wouldn't it make sense to convert the other examples that export maps to also use `html2canvas`?

ahocevar:
That's what the `print-to-scale.html` example does. What I usually do is just call `setTarget()` on the map to an offscreen div with the correct size, and once `rendercomplete` is triggered I do the export and move the map back to the original target.

bojanbizjak:
Is it meant that the getCompositeCanvas function will also work when the map has a combination of WebGLTile, VectorImage and Vector layers?

mike-000:
Yes, it should be able to handle any combination of layers.  It uses the same code as the Export example https://openlayers.org/en/latest/examples/export-map.html (except the `backgroundColor` is set after applying the transform to handle the case where a semi-opaque layers overlays another layer).

Amirh0sseinHZ:
Hi,
This [commit ](https://github.com/openlayers/openlayers/pull/13351/commits/1acf62b6e6ffe18a200c58cf46323dd86e99e16c)might be relevant to this PR too.

PR #13217, opened 2022-01-11
Fix MVT incorrect feature data handling

### LINKED ISSUE #13213: Cannot read properties of null (reading 'getStyleFunction')
**Describe the bug**
I've got vector tiles generated from QGIS which I have successfully loaded Mapbox GL but when I load them in a simple OL setup I get the error "Cannot read properties of null (reading 'getStyleFunction')". 

https://codesandbox.io/s/vector-tile-info-forked-hbivu?file=/main.js

I'm new to OL so its quite possible I'm doing something wrong. I've tried applying a style property to the VectorTileLayer without success. Any recommendations would be greatly appreciated.

EvertEt:
Closes #13213

This would have been caught by having `strictNullChecks` enabled, which might be good to consider in the future.

### DIFF
--- a/src/ol/format/MVT.js
+++ b/src/ol/format/MVT.js
@@ -164,7 +164,7 @@ class MVT extends FeatureFormat {
    * @param {PBF} pbf PBF
    * @param {Object} rawFeature Raw Mapbox feature.
    * @param {import("./Feature.js").ReadOptions} options Read options.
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @EvertEt! Agreed on `strictNullChecks`, but that would be a really, really huge effort. Definitely something to keep in mind, e.g. for a code sprint.

PR #13221, opened 2022-01-11
Draw VectorTiles with interpolate in hybrid mode again

### LINKED ISSUE #13218: Interpolate default changed for VectorTile source
**Describe the bug**
VectorTile sources have interpolate disabled by default since v6.10.0

**To Reproduce**
Interpolate enabled: https://openlayers.org/en/v6.9.0/examples/osm-vector-tiles.html
Interpolate disabled: https://openlayers.org/en/v6.10.0/examples/osm-vector-tiles.html

Interpolate is set to false by default here:
https://github.com/openlayers/openlayers/blob/e98607e3448338c08e6e4e11ec963a622c1cef4b/src/ol/source/Source.js#L100
and here:
https://github.com/openlayers/openlayers/blob/e98607e3448338c08e6e4e11ec963a622c1cef4b/src/ol/Tile.js#L132

**Expected behavior**
Vector tiles should be rendered with image smoothing enabled by default.

M393:
Fixes #13218

### DIFF
--- a/src/ol/source/VectorTile.js
+++ b/src/ol/source/VectorTile.js
@@ -124,6 +124,7 @@ class VectorTile extends UrlTile {
       attributions: options.attributions,
       attributionsCollapsible: options.attributionsCollapsible,
       cacheSize: options.cacheSize,
+      interpolate: true,
       opaque: false,
       projection: projection,
       state: options.state,


ahocevar (approved):
Thanks, @M393

PR #13227, opened 2022-01-11
Default CORS mode for WebGL rendered sources

tschaub:
This makes it so images loaded by sources used with WebGL tile layers have their [request mode](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode) set to CORS and their [credentials mode](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) set to same-origin.  Using no-cors mode (the previous default) does not work with images used as textures.


ahocevar (approved):
Very nice, thanks!

PR #13229, opened 2022-01-11
Load api navigation dynamically to reduce needed disk space

MoonE:
Load the apidoc navigation via fetch to reduce overall disk space used from 268 MB to 14 MB.


ahocevar (approved):
Excellent work, thanks @MoonE!

PR #13236, opened 2022-01-13
Re-enable image smoothing after rendering

### LINKED ISSUE #13234: Interpolation of TileLayer does not work in Firefox after adding VectorImage layer
**Describe the bug**
This only happens in Firefox, Chrome seems to be OK.

**To Reproduce**
1. Open Firefox (I used 95.0.1 under Linux Mint)
2. Go to https://jsfiddle.net/oxdLycwn/1/
3. See the map interpolated nicely ("blur effect")
4. Click on the button to add VectorImage layer
5. Move the map a little by dragging the map
6. The map is not interpolated now and you can see individual pixels

**Expected behavior**

mike-000:
Fixes #13234

Reset interpolation after rendering Image and TileImage sources as the canvas may be shared.

Set interpolation for Vector sources as these may be rendered by VectorImage layers.

### DIFF
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -4,7 +4,7 @@
 import CanvasLayerRenderer from './Layer.js';
 import ViewHint from '../../ViewHint.js';
 import {ENABLE_RASTER_REPROJECTION} from '../../reproj/common.js';
-import {IMAGE_SMOOTHING_DISABLED} from './common.js';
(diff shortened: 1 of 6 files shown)


ahocevar (commented):
Thanks, @mike-000. Would you be able to either add a rendering test or a regression test that ensures the correct `imageSmoothingEnabled` property on the canvas context?

ahocevar (approved):
Thanks, @mike-000.

PR #13237, opened 2022-01-14
Describe use of className with declutter

### LINKED ISSUE #13235: MapboxVector layers zIndex doesn't affect labels/text
**Describe the bug**
When adding vector layer on top of a MapboxVector layer the labels/text of the MapboxVector layer are always on top of the vector layer no mater the zIndex.  This seems to be the case in V6.8.0 and up and is not the case in V6.7.0.

**To Reproduce**
example of this can be created here.  Just add an API key and then draw over the labels and you will see that the drawing ends up under the label even though its on top of everything else. 
https://codesandbox.io/s/draw-features-forked-jf2u2?file=/main.js

**Expected behavior**
Setting the zIndex of a Vector Layer greater than that set for a MapboxVector layer would put the items in the Vector Layer over all items in the MapboxVector including labels/text.

mike-000:
Fixes #13235

Update docs based on https://github.com/openlayers/openlayers/issues/10096#issuecomment-538962484

### DIFF
--- a/src/ol/layer/BaseVector.js
+++ b/src/ol/layer/BaseVector.js
@@ -45,6 +45,11 @@ import {
  * is defined by the z-index of the layer, the `zIndex` of the style and the render order of features.
  * Higher z-index means higher priority. Within the same z-index, a feature rendered before another has
  * higher priority.
+ *
+ * As an optimization decluttered features from layers with the same `className` are rendered above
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Thanks for this clarifying doc improvement, @mike-000.

PR #13259, opened 2022-01-18
Add crossOrigin option to LiteralSymbolStyle

### LINKED ISSUE #13256: WebGL points layer cross origin symbols not supported
**Describe the bug**
The Icon Sprites example is currently reporting a CORS issue on the sprite when opened in CodeSandbox (which may be due to changes in CodeSandbox).  However that highlights the lack of a `crossOrigin` option in `LiteralSymbolStyle` (or an `'anonymous'` default).

**To Reproduce**
https://codesandbox.io/s/icon-sprite-webgl-forked-8lwcg

**Expected behavior**
A `crossOrigin` option, which should default to  `'anonymous'` .

mike-000:
Fixes #13256

Add `crossOrigin` option to `LiteralSymbolStyle` and apply it (defaulting to anonymous) when loading icon or sprite.  The example will not need updating but should work again in CodeSandbox which redirects to the original url for images.

### DIFF
--- a/src/ol/webgl/ShaderBuilder.js
+++ b/src/ol/webgl/ShaderBuilder.js
@@ -548,6 +548,8 @@ export function parseLiteralStyle(style) {
 
   if (symbStyle.symbolType === 'image' && symbStyle.src) {
     const texture = new Image();
+    texture.crossOrigin =
+      symbStyle.crossOrigin === undefined ? 'anonymous' : symbStyle.crossOrigin;
     texture.src = symbStyle.src;
     builder
       .addUniform('sampler2D u_texture')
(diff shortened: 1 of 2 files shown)


src/ol/webgl/ShaderBuilder.js
```diff
@@ -548,6 +548,7 @@ export function parseLiteralStyle(style) {
 
   if (symbStyle.symbolType === 'image' && symbStyle.src) {
     const texture = new Image();
+    texture.crossOrigin = symbStyle.crossOrigin || 'anonymous';
```
ahocevar, line 551:
```suggestion
    texture.crossOrigin = symbStyle.crossOrigin === undefined ? 'anonymous' : symbolStyle.crossOrigin;
```
In case a user demands `''` as `crossOrigin` value.

ahocevar (changes requested):

ahocevar (approved):
Thanks, @mike-000

src/ol/webgl/ShaderBuilder.js
tschaub: [My understanding](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) is that `''` is interpreted as `'anonymous'`.

PR #13261, opened 2022-01-19
add @api comment on getAllLayers method

XiaofengZeng:
This is my first commit. When using Map class, I found that getAllLayers() method can be used, but the API documentation did not find the description of this method. Would love to be able to merge my commits.


ahocevar (approved):
Thanks, @XiaofengZeng

Issue #13270, opened 2022-01-22
utfgrid.html example  working ?


jipexu:
**Describe the bug**
In this example i get 
404 error(s) 
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (5.grid.json, line 0)
https://b.tiles.mapbox.com/v4/mapbox.geography-class/3/4/5.grid.json?access_token=pk.eyJ1IjoiYWhvY2V2YXIiLCJhIjoiY2t0cGdwMHVnMGdlbzMxbDhwazBic2xrNSJ9.WbcTL9uj8JPAsnT9mgb7oQ
....

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/utfgrid.html
2. Must we see a grid  ??
3. See 404 error on console  

**Expected behavior**
See a grid ??
or i dont understand the example ...

ah i see we have no key for this example ....??? 
I understand the management problem of keys in our examples, but an example must be an example and work!... so what to do ...(mapbox might provide a key to the openlayers.org domain name?? )

... and the example not work in sandbox from the example (sandbox error(s))

mike-000:
The example is working for me now.   The API key is hidden and is probably restricted to not work on other domains.  But CodeSandbox is throwing a "Cannot import an empty path" error which I have not seen in any other example - that is not related to the API key.

jipexu:
@mike-000  so i have probably not understand the example ... i expected for a grid displayed ....
yes same error for the sandbox ...

mike-000:
The name comes from the text grid https://github.com/mapbox/utfgrid-spec/blob/master/1.3/utfgrid.md#example-utfgrid-json-file used to index the country data in the tiles.

The CodeSandbox error is due to the invalid initial setting of `src` in `<img id="country-flag" src="" />`.  Remove `src=""` and the error goes away.

jipexu:
thank @mike-000 and sorry for my ignorance

PR #13271, opened 2022-01-22
Remove unneeded `src=""`

mike-000:
Fixes https://github.com/openlayers/openlayers/issues/13270#issuecomment-1019285139

Using an unneeded `src=""` on an `img` tag causes an obscure error in CodeSandbox


ahocevar (approved):
Thanks, @mike-000

PR #13290, opened 2022-01-25
Allowing to pass additional options to the geotiff.js source

### LINKED ISSUE #13289: GeoTIFF source additional options
In some cases it is necessary to pass additional options to the GeoTIFF source, e.g authorization headers.

An additional option in the GeoTIFF construction to be passed to the source.

constantinius:
Fixes #13289

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -207,14 +207,15 @@ function getImagesForTIFF(tiff) {
 
 /**
  * @param {SourceInfo} source The GeoTIFF source.
+ * @param {object} options Options for the GeoTIFF source.
  * @return {Promise<Array<GeoTIFFImage>>} Resolves to a list of images.
  */
-function getImagesForSource(source) {
+function getImagesForSource(source, options) {
   let request;
   if (source.overviews) {
-    request = tiffFromUrls(source.url, source.overviews);
+    request = tiffFromUrls(source.url, source.overviews, options);
   } else {
-    request = tiffFromUrl(source.url);
+    request = tiffFromUrl(source.url, options);
   }
   return request.then(getImagesForTIFF);
 }
@@ -308,6 +309,7 @@ function getMaxForDataType(array) {
  * sources, one with 3 bands and {@link import("./GeoTIFF.js").SourceInfo nodata} configured, and
  * another with 1 band, the resulting data tiles will have 5 bands: 3 from the first source, 1 alpha
  * band from the first source, and 1 band from the second source.
+ * @property {object} [sourceOptions] Additional options to be passed to the underlying geotiff.js source.
  * @property {boolean} [convertToRGB = false] By default, bands from the sources are read as-is. When
  * reading GeoTIFFs with the purpose of displaying them as RGB images, setting this to `true` will
  * convert other color spaces (YCbCr, CMYK) to RGB.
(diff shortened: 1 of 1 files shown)


src/ol/source/GeoTIFF.js
```diff
@@ -308,6 +309,7 @@ function getMaxForDataType(array) {
  * sources, one with 3 bands and {@link import("./GeoTIFF.js").SourceInfo nodata} configured, and
  * another with 1 band, the resulting data tiles will have 5 bands: 3 from the first source, 1 alpha
  * band from the first source, and 1 band from the second source.
+ * @property {object} [sourceOptions] Additional options to be passed to the underlying geotiff.js source.
```
ahocevar, line 312: For the user it is not obvious what `object` to pass as `sourceOptions`. This should be replaced with a type defined by a `@typedef`, which lists and explains the available options. To keep the API lean, only the options that we think are necessary (e.g. setting request headers) should be listed there.

constantinius: @ahocevar Updated. Could you have a look?

src/ol/source/GeoTIFF.js
```diff
@@ -298,6 +299,20 @@ function getMaxForDataType(array) {
   return 255;
 }
 
+/**
+ * @typedef {Object} GeoTIFFSourceOptions
+ * @property {boolean} [forceXHR=false] Whether to force the usage of the browsers XMLHttpRequest API.
+ * @property {Object} [headers] additional headers to be passed with each request.
```
ahocevar, line 305: I assume this is an object literal of key-value pairs?
```suggestion
 * @property {Object<string, string>} [headers] additional key-value pairs of headers to be passed with each request. Key is the header name, value the header value.
```

src/ol/source/GeoTIFF.js
```diff
@@ -308,6 +323,7 @@ function getMaxForDataType(array) {
  * sources, one with 3 bands and {@link import("./GeoTIFF.js").SourceInfo nodata} configured, and
  * another with 1 band, the resulting data tiles will have 5 bands: 3 from the first source, 1 alpha
  * band from the first source, and 1 band from the second source.
+ * @property {GeoTIFFSourceOptions} [sourceOptions] Additional options to be passed to the underlying geotiff.js source.
```
ahocevar, line 326:
```suggestion
 * @property {GeoTIFFSourceOptions} [sourceOptions] Additional options to be passed to [geotiff.js](https://geotiffjs.github.io/geotiff.js/module-geotiff.html)'s `fromUrl` or `fromUrls` methods.
```

ahocevar (commented):
I added two more suggestions. If you can address these, then this is good to go. Thanks!

src/ol/source/GeoTIFF.js
constantinius: Yes, right! Will include this.

src/ol/source/GeoTIFF.js
constantinius: Thanks, thats better

ahocevar (approved):
Thanks, @constantinius

tschaub:
Would anybody object if we just added an `options` property to the existing source info object (as a sibling of `url` and `overviews`)?  This would mean you would have to repeat options share among multiple sources.  But it would also mean you could have separate options for each of that were ever needed.

I can prepare a pull request if it sounds ok.

PR #13291, opened 2022-01-26
Defaults for generic types

### LINKED ISSUE #13230: Generic type 'Layer<SourceType, RendererType>' requires 2 type argument(s).
**Describe the bug**
After upgrading [our lib](https://github.com/hslayers/hslayers-ng) to use ol@6.10.0 or ol@6.11.0, the build fails with the above-mentioned errors.
```
Error: projects/hslayers/src/components/utils/layer-utils.service.ts:408:27 - error TS2314: Generic type 'Layer<SourceType, RendererType>' requires 2 type argument(s).

408   isLayerClustered(layer: Layer<Source>): boolean {
```

etc.

Our lib is based on Angular (i.e. TypeScript). 

There is a question from my side, rather then a direct bug report, whether there was some undocumented breaking change happening between versions 6.9 and 6.10?

**To Reproduce**
1. Use openlayers in a TS application
2. Use `Layer<Source>` as a type
3. Compile the app and see error

**Expected behavior**
No breaking changes when upgrading between minor-releases.

ahocevar:
This pull request updates TypeScript to v4.6.0-beta, which allows specifying defaults for JSDoc type generics.

The generics for Layer, Source and Feature are updated to use a default, which makes them optional for consumers.

Fixes #13230.


ahocevar:
The type changes in the XML parsers were necessary for TypeScript v4.6.0 to not complain.

tschaub (approved):

sbrunner (approved):

PR #13292, opened 2022-01-27
Update to geotiff@2

### LINKED ISSUE #13046: Cloud Optimized GeoTIFF example is not working on codesandbox

### LINKED ISSUE #13285: 6.10.1-dev.1641117808939 *seems* to break ts import from index.ts for jest test files.
When we upgrade from 6.10.1-dev.1641117746220 to 6.10.1-dev.1641117808939, our exports from our index.ts file are not recognized in our jest test files. I am assuming this is a weird transitive dependency. It is a problem for all such imports -- not just OL. I checked the jest version -- it is 27.4.6 in both working and non-working versions.

I discovered this strange behavior when upgrading from 6.10.0 to 6.12.0 and then narrowed it down. 

1. Install Jest
2. create a very simple Jest test
>>>>>>
 import { Util } from '../ts'

describe('convertMultiworldLongitude', () => {
  it('converts out of bounds longitudes', () => {
    const posLong = Util.convertMultiworldLongitude(1224)
  })
})
>>>>>>
3. Under the directory '../ts' there is an index.ts which exports Util. This export is recognized in the working version and not recognized in the other version.
4. When the export is not recognized, the test of course fails. I can then replace the import statement with an explicit import -- import { Util } from '../ts/util/Util'. This succeeds

I *believe* the error message is not of interest. However, I do include it below. The error has nothing to do with GeoTiFFImage.

I am hoping that the specificity of the version will be helpful. Really like OL a lot -- this is the first problem we have had in ages.

<<<<<<<<
 Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    ...

    /node_modules/geotiff/src/geotiff.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import GeoTIFFImage from './geotiffimage';
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (node_modules/ol/src/source/GeoTIFF.js:7:1)
<<<<<<<<<

ahocevar:
This pull request updates geotiff.js to v2. It is now no longer necessary to create our own worker for the GeoTIFF decoder pool.

Should fix #13046.
Should fix #13285.


src/ol/source/GeoTIFF.js
```diff
@@ -471,7 +447,7 @@ class GeoTIFFSource extends DataTile {
       for (let imageIndex = 0; imageIndex < imageCount; ++imageIndex) {
         const image = images[imageIndex];
         const nodataValue = image.getGDALNoData();
-        metadata[sourceIndex][imageIndex] = image.getGDALMetadata();
+        metadata[sourceIndex][imageIndex] = image.getGDALMetadata(0);
```
ahocevar, line 450: This was a breaking change in geotiff@2: `GeoTIFFImage#getGDALMetadata()` now has to be called with a sample index, otherwise only metadata that is not associated with a sample will be returned. In v1, we got the metadata for the last sample, now we get the metadata for the first sample (`0`).

I *think* what we actually want is the metadata for each `wantedSample`, but making such a change would be more involved and probably deserve a separate pull request.

tschaub: Yeah, we could add another dimension to the values we store.  We could also wait until we run into a case where the nodata values are different per-band.  Curious to know if people do that.

ahocevar: I'd also say we eait until we run into such a case. Just mentioned this breaking change here as a reminder that we currently assume all bands to have the same min/max values, like we and geotiff.js do for nodata.

tschaub (approved):
Thanks for all the work on this, @ahocevar.  I meant to approve yesterday, but then got distracted looking into the related upstream changes in geotiff.js.  Nice work!

PR #13293, opened 2022-01-28
Re-assign style variables on setStyle()

### LINKED ISSUE #13287: WebGLTileLayer: updateStyleVariables is broken after using setStyle
**Describe the bug**
When using a `WebGLTileLayer`s `setStyle` function, the `updateStyleVariables` does not seem to work anymore and the 

**To Reproduce**
1. Go to https://geotiffjs.github.io/dem-app/
2. Use any slider, which should update the visualization (this is using `updateStyleVariables` underneath)
3. Set a different color ramp or visualization mode (this is using `setStyle` underneath)
4. Use the sliders again, and no changes in the visualization is to be observed

**Expected behavior**
Using `updateStyleVariables` should work after using `setStyles`

ahocevar:
With this change, when `setStyle()` is called on a `WebGLTile` layer, the style variables will also be reassigned. Previously, subsequent calls to `updateStyleVariables()` modified the old style variables, which were no longer bound to the new style.

Fixes #13287.

### DIFF
--- a/src/ol/layer/WebGLTile.js
+++ b/src/ol/layer/WebGLTile.js
@@ -456,10 +456,13 @@ class WebGLTileLayer extends BaseTileLayer {
   /**
    * Update the layer style.  The `updateStyleVariables` function is a more efficient
    * way to update layer rendering.  In cases where the whole style needs to be updated,
-   * this method may be called instead.
+   * this method may be called instead.  Note that calling this method will also replace
+   * any previously set variables, so the new style also needs to include new variables,
+   * if needed.
    * @param {Style} style The new style.
    */
   setStyle(style) {
+    this.styleVariables_ = style.variables || {};
     this.style_ = style;
     const parsedStyle = parseStyle(this.style_, this.getSourceBandCount_());
     const renderer = this.getRenderer();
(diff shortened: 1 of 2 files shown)


src/ol/layer/WebGLTile.js
```diff
@@ -474,7 +477,7 @@ class WebGLTileLayer extends BaseTileLayer {
 
   /**
    * Update any variables used by the layer style and trigger a re-render.
-   * @param {Object<string, number>} variables Variables to update.
+   * @param {Object<string, (string|number)>} variables Variables to update.
```
ahocevar, line 480: Previously, the initial variables allowed strings and numbers as values, which is also the documented behavior, but `updateStyleVariables()` only allowed numbers. Now both accept strings and numbers.

ahocevar: @tschaub, is this desired? Or should it be the other way around, i.e. only accept numbers everywhere?

ahocevar: Nevermind. I'm removing this change. A quick glance on how the `stringLiteralsMap` works, it looks like variable updates really only work with numbers.

constantinius:
This seems to fix the issue I'm experiencing. With this PR I found this workaround, which also solves the issue:

```javascript
        this.demLayer.styleVariables_ = this.renderVariables;
        this.demLayer.setStyle({
          color: this.renderStyle,
          variables: this.renderVariables,
        });
```

tschaub (approved):
Thanks for fixing this up.

PR #13297, opened 2022-01-28
Add function to convert RenderFeature to Feature

MoonE:
See https://github.com/openlayers/openlayers/pull/13195#issuecomment-1023672163

Should the `geometryFromRenderFeature` function be placed somewhere in the `ol/geom`?

### DIFF
--- a/src/ol/render/Feature.js
+++ b/src/ol/render/Feature.js
@@ -373,12 +373,6 @@ export function toGeometry(renderFeature) {
       return endss.length > 1
         ? new MultiPolygon(flatCoordinates, GeometryLayout.XY, endss)
         : new Polygon(flatCoordinates, GeometryLayout.XY, ends);
-    case GeometryType.MULTI_POLYGON:
-      return new MultiPolygon(
-        renderFeature.getFlatCoordinates(),
-        GeometryLayout.XY,
-        /** @type {Array<Array<number>>} */ (renderFeature.getEndss())
-      );
     default:
       throw new Error('Invalid geometry type:' + geometryType);
   }
--- a/src/ol/render/Feature.js
+++ b/src/ol/render/Feature.js
@@ -1,7 +1,17 @@
 /**
  * @module ol/render/Feature
  */
+import Feature from '../Feature.js';
+import GeometryLayout from '../geom/GeometryLayout.js';
 import GeometryType from '../geom/GeometryType.js';
+import {
+  LineString,
+  MultiLineString,
+  MultiPoint,
+  MultiPolygon,
+  Point,
+  Polygon,
+} from '../geom.js';
 import {
   compose as composeTransform,
   create as createTransform,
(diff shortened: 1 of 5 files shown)


MoonE:
Not sure if MultiPolygons are handled correctly.
Should the winding order / contains be checked for each ring to determine what kind of polygon it is?
https://github.com/openlayers/openlayers/blob/9e4d943239a373e12ab985b95502544f0d9bdda9/src/ol/format/MVT.js#L450-L454

EvertEt:
This looks amazing @MoonE ! Thank you 🎉 

It looks functional when I try to use it in a vectorTile dragbox selection test 🙌

ahocevar:
> Should the geometryFromRenderFeature function be placed somewhere in the ol/geom?

Actually I think because this function depends on many geometry types, it would be best to place both in the `ol/render/Feature` module. These features are created only by the MVT format, which has the same dependencies.

In this case, the names for these functions should probably be `ol/render/Feature.toFeature()` and `ol/render/Feature.toGeometry()`:
```js
import {toFeature, toGeometry} from 'ol/render/Feature';

const feature = toFeature(renderFeature);
const geometry = toGeometry(renderFeature);
```
The geometry types these functions need to support are
* LineString
* MultiLineString
* MultiPoint
* MultiPolygon
* Point
* Polygon

src/ol/render/Feature.js
```diff
@@ -326,4 +337,76 @@ RenderFeature.prototype.getEndss = RenderFeature.prototype.getEnds;
...
+        /** @type {Array<number>} */ (renderFeature.getEnds())
+      );
+    case GeometryType.POLYGON:
+      const flatCoordinates = renderFeature.getFlatCoordinates();
+      const ends = /** @type {Array<number>} */ (renderFeature.getEnds());
+      const endss = inflateEnds(flatCoordinates, ends);
+      return endss.length > 1
+        ? new MultiPolygon(flatCoordinates, GeometryLayout.XY, endss)
+        : new Polygon(flatCoordinates, GeometryLayout.XY, ends);
+    case GeometryType.MULTI_POLYGON:
```
ahocevar, line 376: You can remove this case. There is no `GeometryType.MULTI_POLYGON` for `ol/render/Feature`. See `ol/format/MVT.getGeometryType()`. For rendering, all that matters is whether the ring is counterclockwise (outer ring) or clockwise (inner ring). Because of the order of the rings, we know that all clockwise rings following a counterclockwise ring are inner rings of that counterclockwise ring. When the next counterclockwise ring is encountered, it is the outer ring of another polygon.

test/node/ol/render/Feature.test.js
```diff
@@ -0,0 +1,247 @@
...
+          [
+            [-0, -0],
+            [-5, -0],
+            [-5, -5],
+            [-0, -0],
+          ],
+        ],
+      ]);
+      const renderFeature = new RenderFeature(
+        geometry.getType(),
```
ahocevar, line 182: There is no MULTI_POLYGON type for a render feature. You can remove this test case.

ahocevar (commented):
Look great, thanks for your effort, @MoonE ! There is no case where a Render Feature gets a MULTI_POLYGON type. I tried to explain this in an inline comment. I hope that clarifies how it works.

src/ol/render/Feature.js
MoonE: I wasn't sure whether a MultiPolygon was used anywhere, because it should work, though less efficiently.
It is removed now.

ahocevar (approved):
Thanks, @MoonE!

PR #13298, opened 2022-01-29
Document geometry type

### LINKED ISSUE #13295: Heatmap doesn't render (v6.12)
**Describe the bug**
Heatmap layers do not appear on the screen, despite being present as layers on a map object

I should note that the heatmap DOES appear when developing on localhost. I'm using Webpack to bundle a react app - but even when I normalized the differences between the webpack = production and webpack = development environments the same behaviour occurs (not visible on production build)

Since there are so many differences between developing on localhost vs on a https domain, and using webpack with development/production mode, I thought it worth posting here as a starting point to figure out the problem. Hope that's okay!

Looking at the site - nccrd.saeon.ac.za the map clearly works as do other layers. It's just the heatmap that does not. I have also tried running `npm install` instead of `npm run ci` to see if the problem was maybe related to the lock file (it is not - this did not help)

**To Reproduce**
1. Go to https://nccrd.saeon.ac.za. There should be a yellowish heat map on top of the base map
2. Go to the developer console. the map, and other relevant objects have been attached to the window object for debugging

```js
// window objects
window.VectorSource // (import VectorSource from 'ol/source/Vector')
window.WKT // (import WKT from 'ol/format/WKT')
window.Heatmap  // import Heatmap from 'ol/layer/Heatmap'

// (1) Defined the layer making function
makeLayer = data => {
  const format = new WKT()
  return new Heatmap({
    gradient: ['#893448', '#d95850', '#eb8146', '#ffb248', '#f2d643', '#ebdba4'],
    source: new VectorSource({
      features: data.POINT_LOCATIONS.data.map(({ xy, normalizedBudget }) => {
        const feature = format.readFeature(xy, {
          dataProjection: 'EPSG:4326',
          featureProjection: 'EPSG:4326',
        })

        feature.set('normalizedBudget', normalizedBudget)
        return feature
      }),
    }),
    blur: 100,
    radius: 80,
  })
}

// (2) Create the layer
heatMapLayer = makeLayer(data)

// (3) Add the layer
map.addLayer(heatMapLayer)
```

By interrogating that map object, you can see that the heatmap layer just ... doesn't show:

```js
// Confirm that the map object is the map that is on the screen
map.getAllLayers()[0].set('opacity', 0.3) // it is!

// There are 3 layers - the 3rd is the heatmap that we just added above. delete it
map.removeLayer(map.getAllLayers()[2])

// Now there are 2 layers. The base layer and the invisible heatmap
heatmap = map.getAllLayers()[1]

// And I can see that all the features are there as supposed to be
heatmap.getSource().getFeatures() // This looks the same as on localhost
```

Also, looking at the DOM output, I can see that the `<canvas>` element seems to be correctly placed and sized. I have no idea why Heatmap layers just don't seem to appear on maps.

**Expected behavior**
I expected the heatmap to render!

ahocevar:
Document that we expect a source of Point geometries.

Fixes #13295.

### DIFF
--- a/src/ol/layer/Heatmap.js
+++ b/src/ol/layer/Heatmap.js
@@ -33,7 +33,7 @@ import {createCanvasContext2D} from '../dom.js';
  * @property {string|function(import("../Feature.js").default):number} [weight='weight'] The feature
  * attribute to use for the weight or a function that returns a weight from a feature. Weight values
  * should range from 0 to 1 (and values outside will be clamped to that range).
- * @property {import("../source/Vector.js").default} [source] Source.
+ * @property {import("../source/Vector.js").default<import("../geom/Point.js").default>} [source] Point source.
  * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
  */
 
--- a/src/ol/layer/WebGLPoints.js
+++ b/src/ol/layer/WebGLPoints.js
@@ -7,7 +7,7 @@ import {assign} from '../obj.js';
 import {parseLiteralStyle} from '../webgl/ShaderBuilder.js';
 
 /**
- * @template {import("../source/Vector.js").default} VectorSourceType
+ * @template {import("../source/Vector.js").default<import("../geom/Point.js").default>} VectorSourceType
  * @typedef {Object} Options
  * @property {import('../style/literal.js').LiteralStyle} style Literal style to apply to the layer features.
  * @property {string} [className='ol-layer'] A CSS class name to set to the layer element.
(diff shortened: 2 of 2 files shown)


src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -563,8 +563,13 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
...
         featureCache.geometry
       );
-      if (!geometry || geometry.getType() !== GeometryType.POINT) {
+      if (!geometry) {
         continue;
+      } else if (geometry.getType() === GeometryType.POINT) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          `Expected a Point geometry, but got a ${geometry.getType()}`
+        );
```
mike-000, line 572: That is warning when the geometry is correct

ahocevar: Ok, let's go with just the documentation change for now. To handle the warning better, it should probably be done when the featureCache_ is populated, and not during rendering.

tschaub (approved):

PR #13300, opened 2022-01-30
Fix 81 "Property 'X' is used before being assigned" TS errors

EvertEt:
Resulting from Observable type narrowing.

The results of this will be noticeable when strictNullChecks are enabled in the future.


EvertEt:
This does seem to have some unwanted side-effects of overwriting these properties so not a good solution

PR #13301, opened 2022-01-30
Improve some null types to prepare for strictNullChecks

EvertEt:
The results of this will be noticeable when strictNullChecks are enabled in the future.

### DIFF
--- a/src/ol/layer/Base.js
+++ b/src/ol/layer/Base.js
@@ -119,7 +119,7 @@ class BaseLayer extends BaseObject {
      * @private
      */
     this.className_ =
-      properties.className !== undefined ? options.className : 'ol-layer';
+      properties.className !== undefined ? properties.className : 'ol-layer';
     delete properties.className;
 
     this.setProperties(properties);
@@ -288,7 +288,7 @@ class BaseLayer extends BaseObject {
   }
 
   /**
-   * Sets the backgrlound color.
+   * Sets the background color.
    * @param {BackgroundColor} [opt_background] Background color.
    */
   setBackground(opt_background) {
--- a/src/ol/source/Image.js
+++ b/src/ol/source/Image.js
@@ -127,7 +127,7 @@ class ImageSource extends Source {
 
     /**
      * @private
-     * @type {Array<number>}
+     * @type {Array<number>|null}
      */
     this.resolutions_ =
       options.resolutions !== undefined ? options.resolutions : null;
@@ -146,7 +146,7 @@ class ImageSource extends Source {
   }
 
   /**
-   * @return {Array<number>} Resolutions.
+   * @return {Array<number>|null} Resolutions.
    */
   getResolutions() {
     return this.resolutions_;
--- a/src/ol/worker/webgl.js
+++ b/src/ol/worker/webgl.js
@@ -29,7 +29,7 @@ worker.onmessage = (event) => {
     const indexBuffer = new Uint32Array(indicesCount);
     const vertexBuffer = new Float32Array(verticesCount);
 
-    let bufferPositions = null;
+    let bufferPositions;
     for (let i = 0; i < renderInstructions.length; i += instructionsCount) {
       bufferPositions = writePointFeatureToBuffers(
         renderInstructions,
--- a/src/ol/control/ZoomToExtent.js
+++ b/src/ol/control/ZoomToExtent.js
@@ -38,7 +38,7 @@ class ZoomToExtent extends Control {
     });
 
     /**
-     * @type {?import("../extent.js").Extent}
+     * @type {?import("../extent.js").Extent|null}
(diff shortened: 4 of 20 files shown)


src/ol/layer/Base.js
```diff
@@ -288,11 +289,11 @@ class BaseLayer extends BaseObject {
   }
 
   /**
-   * Sets the backgrlound color.
+   * Sets the background color.
    * @param {BackgroundColor} [opt_background] Background color.
    */
   setBackground(opt_background) {
-    this.background_ = opt_background;
+    this.background_ = opt_background !== undefined ? opt_background : false;
```
tschaub, line 296: It might make sense to change the zero value to an empty string (instead of `false`) here.  Either way, it looks like this might impact other code.

EvertEt: Changing the default would be a bigger change since the false value seems documented for MapboxLayer: https://openlayers.org/en/latest/apidoc/module-ol_layer_MapboxVector-MapboxVectorLayer.html. 
Maybe this change should be reverted and tackled as a whole in another PR (and also looking at all `getBackground` and `BackgroundColor` usages. Possibly adding `undefined` to BackgroundColor). What do you think about this?

src/ol/source/Image.js
```diff
@@ -127,10 +127,9 @@ class ImageSource extends Source {
 
     /**
      * @private
-     * @type {Array<number>}
+     * @type {Array<number>|undefined}
      */
-    this.resolutions_ =
-      options.resolutions !== undefined ? options.resolutions : null;
+    this.resolutions_ = options.resolutions;
```
tschaub, line 132: Why `undefined` instead of `null` here?

src/ol/source/Image.js
```diff
@@ -127,10 +127,9 @@ class ImageSource extends Source {
 
     /**
      * @private
-     * @type {Array<number>}
+     * @type {Array<number>|undefined}
```
tschaub, line 130: Why `undefined` instead of `null` here?

src/ol/worker/webgl.js
```diff
@@ -29,7 +29,7 @@ worker.onmessage = (event) => {
     const indexBuffer = new Uint32Array(indicesCount);
     const vertexBuffer = new Float32Array(verticesCount);
 
-    let bufferPositions = null;
+    let bufferPositions = undefined;
```
tschaub, line 32:
```suggestion
    let bufferPositions;
```

src/ol/source/Image.js
EvertEt: Because then it was an easy match with the `getResolutions` defined in the parent class https://github.com/openlayers/openlayers/blob/main/src/ol/source/Source.js#L155

src/ol/worker/webgl.js
EvertEt: Ok. I did this because of a `Variable might not have been initialized ` warning, but I see it's just a webstorm warning, not TS or linter.

src/ol/source/Image.js
tschaub: If you instead changed the type of the return on the parent to `Array<string>|null`, then you wouldn't need to change the code below.

In addition, `Array<string>|null` would be consistent with the other type annotation changes you are making here.

I think it would be preferable if this pull request were restricted to type annotation changes (comments) instead of code changes.  If there are bugs revealed in changing the type annotations, then that would justify code changes.  Otherwise, including unnecessary code changes is more likely to introduce new bugs.

EvertEt: I agree null would fit closer to the code style here.
Should we leave https://github.com/openlayers/openlayers/blob/main/src/ol/View.js#L449 like it is, or also switch undefined for null, but then requiring a code change again for the option parsing (which is internal so I think it's not a problem)

tschaub: I'd be tempted to see how far you can get with comment-only changes toward supporting `strictNullChecks: true`.  So not changing any of the behavior, just the type checking.

Later we could work on making things consistently `Something|null` or `Something|undefined`.   I guess my preference would be to choose whatever minimizes unnecessary guards, ternary expressions, and extra code.  It feels like `Something|undefined` would be simpler.  But I think this would probably be a breaking change.

Anybody else have any opinions?

ahocevar: Totally agree. It means reviewing the whole code base to see where we actually allow `null` and where we don't. Regarding `null` vs `undefined`, I think we have several places in the code where `null` has a different effect than `undefined`. So there will definitely be breaking changes, but I agree that the old Closure Library advice to use `null` for `Object`s and `undefined` for simple types when they are not set may not hold true any more.

tschaub:
Thanks for the continued work on this, @EvertEt

PR #13319, opened 2022-01-31
fix: fix casing for z-index

OSHistory:
The CSS-property z-index had some faulty casings. Only occurs in src/ol/ol.css


ahocevar (approved):
Thanks, @OSHistory

PR #13321, opened 2022-01-31
Avoid canvas memory overflow on iOS devices, to avoid the 'Total canvas memory use exceeds the maximum limit' warning 

xsfavisf:
Workaround to avoid canvas memory overflow on iOS devices, to avoid the 'Total canvas memory use exceeds the maximum limit (xxx MB)' warning that causes VectorTile layers to stop painting elements on the map correctly.
Tests have been done on 2 iPads with  Safari:

- iOS 12 with max canvas memory:224MB 
- iOS 15 with max canvas memory:480MB

My solution is based on maintaining the reference to all the canvases created by the TileCache and on their correct release in memory when they are no longer needed and when total canvas memory usage exceeds 200MB (after that the rendering becomes unstable and sometimes the application crashes). This link has helped me a lot to understand how canvas works in iOS:
https://pqina.nl/blog/total-canvas-memory-use-exceeds-the-maximum-limit/

This is a problem that Apple should have fixed many years ago, they are not managing the canvas garbage collector correctly in their SO.

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -37,7 +37,7 @@ import {fromUserCoordinate, toUserCoordinate} from './proj.js';
 import {getUid} from './util.js';
 import {hasArea} from './size.js';
 import {listen, unlistenByKey} from './events.js';
-import {removeNode} from './dom.js';
+import {releaseAllCanvasMemory, removeNode} from './dom.js';
 
 /**
  * State of the current frame. Only `pixelRatio`, `time` and `viewState` should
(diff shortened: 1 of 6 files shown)


ahocevar:
Thanks for your work on this, @xsfavisf. I think your pull request does more than what needs to be done. The `VectorRenderTile` tiles are managed in a canvas pool already, to make sure that no more tiles than needed are created. The same could be done for `ol/reproj/Tile`. The other tiles in the tile cache are images, not canvases, so your pull request won't reduce the pressure on canvas memory.

There is another place where canvas memory consumption can be reduced, and that is `ol/renderer/canvas/ImageLayer.js`. That issue is ticketed as #9578.

PR #13323, opened 2022-02-01
Fix rendercomplete for WebGLPoints layer and subclasses

ahocevar:
This pull request fixes the problem described in https://github.com/openlayers/openlayers/pull/13294#issuecomment-1026023330.

It introduces a `ready` flag on the layer renderer. For the WebGLPoints layer, this will be set to `false` whenever the renderer is waiting for a message from the worker, in which case it cannot render the latest data.

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -137,6 +137,8 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
       postProcesses: options.postProcesses,
     });
 
+    this.ready = false;
+
     this.sourceRevision_ = -1;
 
     this.verticesBuffer_ = new WebGLArrayBuffer(ARRAY_BUFFER, DYNAMIC_DRAW);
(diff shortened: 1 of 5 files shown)


src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -320,7 +322,7 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
               event.data.renderInstructions
             );
           }
-
+          this.ready = true;
```
MoonE, line 325: Shouldn't this only be set if the received data is *not* for hit detection?
Or after receiving both, if hitdetection is enabled?

MoonE:
It should also discard outdated worker messages when buffers are rebuild multiple times in a row.
Something like https://github.com/openlayers/openlayers/commit/78c6701c2d0ea0f8ac3dc21d49e5991e2edbc7be

ahocevar:
Good suggestions, @MoonE. Please feel free to commit those directly to this pull request. It would be great if the case of a discarded message could also be tested somehow. Do you see a way to add a test which does that?

ahocevar:
Very nice, thanks! No, hit detection is not relevant.

MoonE (approved):
Ok, looks good to me.

ahocevar:
Thanks for your help on this, @MoonE!

PR #13328, opened 2022-02-04
Improve some nullable map types

### LINKED ISSUE #13326: Layer.setMap(null) does not work in typescript strict mode
**Describe the bug**
To remove an unmanaged layer from openlayers we are supposed to use `Layer.setMap(null)` as per the docs. Since the jsdocs for the `setMap` function in `src/ol/layer/Layer.js` is written as 
```js
   * @param {import("../PluggableMap.js").default} map Map.
   * @api
   */
  setMap(map) {
```
the types generated does not allow `null` to be passed to setMap. Which in turn makes typescript not compile.

**To Reproduce**
Make a layer and try to unset the map with `.setMap(null)` with typescript strict mode enabled

**Expected behavior**
The jsdoc for generating typescript should be defined as
```js
   * @param {import("../PluggableMap.js").default | null} map Map.
   * @api
   */
  setMap(map) {
```

EvertEt:
Closes #13326

I left out the interactions for now, they could probably also be improved.

### DIFF
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -378,8 +378,8 @@ class Overlay extends BaseObject {
 
   /**
    * Set the map to be associated with this overlay.
-   * @param {import("./PluggableMap.js").default|undefined} map The map that the
-   * overlay is part of.
+   * @param {import("./PluggableMap.js").default|undefined|null} map The map that the
+   * overlay is part of. Pass `null` to just remove the overlay from the current map.
    * @observable
    * @api
    */
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -249,14 +249,14 @@ class Overlay extends BaseObject {
 
   /**
    * Get the map associated with this overlay.
-   * @return {import("./PluggableMap.js").default|undefined} The map that the
+   * @return {import("./PluggableMap.js").default|null} The map that the
    * overlay is part of.
    * @observable
    * @api
    */
   getMap() {
-    return /** @type {import("./PluggableMap.js").default|undefined} */ (
-      this.get(Property.MAP)
+    return /** @type {import("./PluggableMap.js").default|null} */ (
+      this.get(Property.MAP) || null
     );
   }
 
@@ -378,7 +378,7 @@ class Overlay extends BaseObject {
 
   /**
    * Set the map to be associated with this overlay.
-   * @param {import("./PluggableMap.js").default|undefined|null} map The map that the
+   * @param {import("./PluggableMap.js").default|null} map The map that the
    * overlay is part of. Pass `null` to just remove the overlay from the current map.
    * @observable
    * @api
--- a/src/ol/control/Control.js
+++ b/src/ol/control/Control.js
@@ -69,7 +69,7 @@ class Control extends BaseObject {
 
     /**
      * @private
-     * @type {import("../PluggableMap.js").default|undefined|null}
+     * @type {import("../PluggableMap.js").default|null}
      */
     this.map_ = null;
 
@@ -98,7 +98,7 @@ class Control extends BaseObject {
 
   /**
    * Get the map associated with this control.
-   * @return {import("../PluggableMap.js").default|undefined|null} Map.
+   * @return {import("../PluggableMap.js").default|null} Map.
    * @api
    */
   getMap() {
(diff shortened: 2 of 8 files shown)


src/ol/control/Control.js
```diff
@@ -124,14 +124,14 @@ class Control extends BaseObject {
     if (this.map_) {
       const target = this.target_
         ? this.target_
-        : map.getOverlayContainerStopEvent();
+        : this.map_.getOverlayContainerStopEvent();
```
ahocevar, line 127: This change and the others here that change `map` to `this.map_` is not necessary, because of the assignment in line 123.

ahocevar: Did you mean to add `| null` in line 112?

ahocevar (changes requested):

src/ol/control/Control.js
EvertEt: Indeed, this is the same behaviour but TS can't figure that out. `map` will not be narrowed down by the if check where `this.map` will be. 
I did this change indeed because I wanted to add null here, but I reverted that part since I wasn't sure since it already supported undefined in this function.

src/ol/Overlay.js
```diff
@@ -378,7 +378,7 @@ class Overlay extends BaseObject {
 
   /**
    * Set the map to be associated with this overlay.
-   * @param {import("./PluggableMap.js").default|undefined} map The map that the
+   * @param {import("./PluggableMap.js").default|undefined|null} map The map that the
```
EvertEt, line 381: Preferably we remove undefined here to keep it consistent but that would be a breaking change so I didn't do it yet.

src/ol/control/Control.js
```diff
@@ -98,7 +98,7 @@ class Control extends BaseObject {
 
   /**
    * Get the map associated with this control.
-   * @return {import("../PluggableMap.js").default|undefined} Map.
+   * @return {import("../PluggableMap.js").default|undefined|null} Map.
```
EvertEt, line 101: Preferably we remove undefined here to keep it consistent but that would be a breaking change so I didn't do it yet.

src/ol/control/Control.js
```diff
@@ -121,7 +121,7 @@ class Control extends BaseObject {
     }
     this.listenerKeys.length = 0;
     this.map_ = map;
-    if (this.map_) {
+    if (map) {
```
EvertEt, line 124: Better TS type inference inside the if

src/ol/Overlay.js
ahocevar: Feel free to remove it now, and also make the argument mandatory where it is currently optional. It's just type improvements.

src/ol/control/Control.js
ahocevar: Feel free to remove it now. It's not a code change, so it is not considered a breaking change according to our API change policy.

src/ol/Overlay.js
```diff
@@ -120,7 +120,9 @@ class Overlay extends BaseObject {
    * @param {Options} options Overlay options.
    */
   constructor(options) {
-    super();
+    super({
+      [Property.MAP]: null,
+    });
```
EvertEt, line 125: Is this the best way to initialize it to null?

ahocevar: I don't think this would help, because tsc wouldn't know what to do with it. Better change the `getMap()` method to `return this.get(Property.MAP) || null;`

src/ol/Overlay.js
```diff
@@ -249,14 +249,14 @@ class Overlay extends BaseObject {
...
+   * @return {import("./PluggableMap.js").default|null} The map that the
    * overlay is part of.
    * @observable
    * @api
    */
   getMap() {
-    return /** @type {import("./PluggableMap.js").default|undefined} */ (
-      this.get(Property.MAP)
+    return /** @type {import("./PluggableMap.js").default|null} */ (
+      this.get(Property.MAP) || null
```
ahocevar, line 259: Although this looks like a breaking change, it actually only fixes the inconsistency that when previously`Map#removeOverlay()` was called, a subsequent call of `Overlay#getMap()` returned `null`, but when called after `Overlay` creation, it returned `undefined`.

ahocevar (approved):
Thanks, @EvertEt

PR #13332, opened 2022-02-05
Avoid duplicate imports

tschaub:
This pulls in the latest `eslint-config-openlayers` config and fixes the duplicate imports.


ahocevar (approved):

tschaub:
Hmm.  Some trouble with actions.

tschaub:
> Hmm. Some trouble with actions.

Our actions failed to run at the same time this incident was occurring: https://www.githubstatus.com/incidents/vvthhf8gxt80.  I assumed that was the reason the latest commit didn't run.  Instead (or in addition), it looks like missing yaml quotes were causing trouble.  I fixed that in 459cd51ae2ae7fded1f1ede5215248f9537f4e7c.

PR #13335, opened 2022-02-06
Get data from tiles

tschaub:
This changes the implementation of the `getDataAtPixel` method for the WebGL tile renderer.  The implementation in #13186 returns a Uint8 array representing the color rendered for all layers at the requested pixel.  The implementation in this branch returns the pixel data value from a single layer.  The data type returned depends on the tile type.  For data tiles, the returned value is the same type as the data tile.  For image tiles, a Uint8 array is returned.

I've removed the test that sets the context's `preserveDrawingBuffer: false` because we no longer require this attribute to be set in order to get pixel data.

The other change in behavior (seen in the tests) is that the method works for pixel values where the fourth band is zero.  Previously, the method returned `null` in this case.  We now return the full pixel array.

The [NDVI example](https://deploy-preview-13335--ol-site.netlify.app/examples/cog-math.html) demonstrates how NDVI values (based on original pixel values) can be displayed on `pointermove` or `click`.  This is not possible (in general) with the current implementation (no guarantee that you can invert a color value to retrieve the original data values, and layer composition and alpha handling is "lossy" in terms of original data values).

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,16 @@
 ## Upgrade notes
 
+### Next
+
+#### New behavior of map.forEachLayerAtPixel() for WebGL tile layers
+
+The `map.forEachLayerAtPixel()` method now returns original pixel values for WebGL-rendered tile layers.  Previously,
+the value returned was based on the RGBA composite pixel based on rendering all layers.  Additional data types have been added to the signature of the callback passed to this method.  In addition to being called with a `Uint8ClampedArray` or `Uint8Array`, the method may now be called with a `Float32Array` or `DataView` depending on the underlying data type of the tiles used.
+
+For example, when rendering a GeoTIFF source that loads floating point values (with `normalize: false`), the `map.forEachLayerAtPixel()` callback will be called with a `Float32Array` representing pixel values.
+
+Previously, when rendering data tile sources with zero pixel values in the fourth band, the `map.forEachLayerAtPixel()` callback would be called with `null`.  Now, the full pixel array will be provided (even if the fourth band is zero).
+
 ### v6.12.0
 
 No special changes are required when upgrading to the 6.12.0 release.
--- a/examples/cog-math.html
+++ b/examples/cog-math.html
@@ -3,9 +3,12 @@ layout: example.html
 title: NDVI from a Sentinel 2 COG
 shortdesc: Calculating NDVI and applying a custom color map.
 docs: >
-  The GeoTIFF layer in this example draws from two Sentinel 2 sources: a red band and a near infrared band.
+  The GeoTIFF layer in this example draws from two Sentinel 2 sources: a red band and a near-infrared band.
   The layer style includes a `color` expression that calculates the Normalized Difference Vegetation Index (NDVI)
   from values in the two bands.  The `interpolate` expression is used to map NDVI values to colors.
+  The `map.forEachLayerAtPixel()` method can be used to retrieve pixel values from the GeoTIFF.  Move your mouse
+  or tap on the map to see calculated NDVI values based on the red and near-infrared pixel values.
 tags: "cog, ndvi"
 ---
 <div id="map" class="map"></div>
+<div>NDVI: <span id="output"></span></div>
--- a/examples/cog-math.js
+++ b/examples/cog-math.js
@@ -79,3 +79,17 @@ const map = new Map({
   ],
   view: source.getView(),
 });
+
+const output = document.getElementById('output');
+function displayPixelValue(event) {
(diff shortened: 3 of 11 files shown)


mike-000:
Could this method be a option for `forEachLayerAtPixel` instead of a change?  I can see there will be cases where the raw data may be required, but it is also useful to be able to get styled or interpolated values from the screen pixels, e.g. in https://deploy-preview-13232--ol-site.netlify.app/examples/webgl-interpolation.html the elevation shown changes smoothly as the raw data is interpolated as a numeric value and then restyled back to a readable pixel (in a real application the layer would be concealed below another and linked to a MousePosition control or similar). But with this change the values displayed would revert to the inferior pixelated values seen in the canvas example (which cannot be interpolated as the RGB bands would be interpolated independently).

tschaub:
@mike-000 - I'm struggling to think of any valid use case, but I think you are describing what should be a different method.  If what you want to get is the rendered pixel value after composing all layers, you should not be calling `map.forEachLayerAtPixel()`.  I think it is a bug that the `map.forEachLayerAtPixel()` method behaves this way in some cases and not others.  I'm not advocating for adding another method, because I think we need to establish a valid use case first.

> But with this change the values displayed would revert to the inferior pixelated values

I'm not sure I agree with this.  The new behavior will return values from the original data.  I wouldn't characterize those as "inferior."

mike-000:
I don't want the value of all layers, in the example I linked I would specifically like the value of one layer as it is styled after interpolation, i.e.

image

which can only be done with WebGL instead of

image

which is the best achievable with 2d canvas.

mike-000:
> I think it is a bug that the `map.forEachLayerAtPixel()` method behaves this way in some cases and not others

It already behaves differently with canvas layers depending on whether interpolation is enabled or not.  In the case of canvas interpolation can only be applied to RGB bands and would corrupt values calculated from RGB encoding, but WebGL can decode the encoded value then interpolate from it.

tschaub:
@mike-000 - It sounds like you are specifically interested in being able to interpolate the data value at a position in the case where you are dealing with oversampled data.  To support this, we could make use of the `hitTolerance` option passed to the `map.forEachLayerAtPixel()` method (or we could add a new option).  By default, a single pixel value is passed to the callback.  We could make it possible to pass an array of pixel values (for example, 9 pixels representing a 3x3 neighborhood).  Then you could calculate any derived value (like elevation) and interpolate.  The callback would need to be passed an additional argument (like `position`) representing the normalized position of the original location within the returned neighborhood (to support the interpolation).

```js
function callback(layer, data, position) {
  // data represents a 3x3 neighborhood of pixel values
  // position is 2 element array of 0..1 values representing the location of the original pixel within the neighborhood
}
map.forEachLayerAtPixel(pixel, callback, {hitTolerance: 1});
```

I think it is important to point out that we are only talking here about interpolating between oversampled pixels here.  Going through the renderer to achieve this is not a good solution in my opinion (mixing with other layers, lossy behavior with alpha).

I don't especially like the above idea (awkward signatures, added complexity to maintain, not commonly used).  I'm also not really that fond of the `map.forEachLayerAtPixel()` method (weird behavior, inconsistently implemented, requires you know about renderer details to use safely).  If others think we should leave that method as is (returning rendered RGBA values instead of source data values), I'll add a new method that is about getting data values.

tschaub:
Abandoning `map.forEachLayerAtPixel()` in favor of `layer.getData()` in #13338 (though I could go either way).

mike-000:
> you are dealing with (extremely) oversampled data

Extremely oversampled data seems to be normal when using DEM sources - e.g. this https://labs.os.uk/prototyping/vector-tiles/height/dynamic-hillshade.html shaded relief example in Mapbox GL where the maxzoom of the DEM source is 9, while the view maxZoom is 14.  Using the original Ordnance Survey EPSG:27700 data downloaded from https://osdatahub.os.uk/downloads/open/Terrain50 with a single tilegrid resolution of 50 with raster tiles at resolution 1.75 (and overzoomed to 0.875) the interpolated elevations and contours typically closely match the surveyed contours on the raster tiles - except when close to high vertical cliffs, an obvious limitation of the data https://codesandbox.io/s/os-contours-forked-nhylw?file=/main.js  There is also some distortion at tile edges as there is no interpolation across the edges (the number of edges present could be reduced by using larger tile sizes).

I could not find a way of producing thin contour lines from oversampled data directly in WebGL as xOffset and yOffset for bands refer to the source data, so zooming in widens styled contour lines.  I obtained the thin lines from the rendered WebGL using ol/source/Raster after applying a simple fix for that to support WebGL layers.  ol/source/Raster would also allow rendered pixels values to be retrieved if forEachLayerAtPixel was deprecated.

PR #13336, opened 2022-02-06
Dispose of webgl contexts

MoonE:
Now properly disposes of the webgl context when running test.

Issues I found (and hopefully fixed) while doing this:
- Calling dispose on `oll/layerWebGLTile` parses the style again when the source is set to null. Updating the style when no source is set should be unnecessary.

- the helper in `ol/renderer/webgl/TileLayer` may be undefined and calling dispose then caused an error.
- the `tileTextureCache` should be cleared every time the helper is destroyed as the textures hold a reference to the helper.
https://github.com/openlayers/openlayers/blob/5fec74c089834caf4b269587ab75041a987863c3/src/ol/renderer/webgl/Layer.js#L186-L192

### DIFF
--- a/src/ol/layer/WebGLTile.js
+++ b/src/ol/layer/WebGLTile.js
@@ -390,7 +390,9 @@ class WebGLTileLayer extends BaseTileLayer {
    * @private
    */
   handleSourceUpdate_() {
-    this.setStyle(this.style_);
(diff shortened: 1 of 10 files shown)


tschaub (approved):

tschaub:
Thanks, @MoonE

PR #13350, opened 2022-02-08
Remove warning for `@type` annotation by jsdoc

MoonE:
> WARNING: The `@type` tag does not permit a description; the description will be ignored. File: Layer.js, line: 20
> WARNING: The `@type` tag does not permit a description; the description will be ignored. File: Layer.js, line: 23


ahocevar (approved):
Thanks for fixing this, @MoonE! This was an oversight on my end.

PR #13351, opened 2022-02-08
Reset globalAlpha back to its initial after mutating it for drawing layers

Amirh0sseinHZ:
The `mapContext` ends up having the opacity of the last layer iterated over.
Since some users might like to pass this `context` around to add their own stuff on top of it (scale, branding, copyright, etc.), I believe it's better to reset it back to its initial `1` value after we're done with the drawings.

### DIFF
--- a/examples/export-map.js
+++ b/examples/export-map.js
@@ -98,6 +98,7 @@ document.getElementById('export-png').addEventListener('click', function () {
         }
       }
     );
+    mapContext.globalAlpha = 1;
     if (navigator.msSaveBlob) {
       // link download attribute does not work on MS browsers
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @Amirh0sseinHZ

mike-000:
If you are resetting the globalAlpha would you also not want to reset the transform which might have been changed to handle layer specific requirements for pixel ratio and/or view rotation?

Amirh0sseinHZ:
@mike-000, That's a good call! Especially since I saw you're trying to extract a method that would return the map canvas similar to ol v5.

PR #13352, opened 2022-02-09
Removing an unnecessary word from DEVELOPING.md

Amirh0sseinHZ:
As obvious forking can be for more experienced folks, it might be the opposite for juniors who are about to make their first-ever contribution.
I think the guide shouldn't assume that it is `obvious` generally.


ahocevar (approved):
Good suggestion, thanks @Amirh0sseinHZ

PR #13363, opened 2022-02-10
Add floor, round and ceil to style expressions

mike-000:
This includes the `floor` operator extracted from #13232 and, for completeness, corresponding `round` and `ceil` operators. As `round` is not fully supported in WebGL it is implemented as `floor(value1 + 0.5)`.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #13365, opened 2022-02-11
Contour Interpolation example

mike-000:
A more practical application of WebGL interpolation than #13232 which avoids the use of `forEachLayerAtPixel` and the complexities of conflicting requirements of interpolation and data reprojection.


mike-000:
This is working as expected for me when opened in codesandbox https://codesandbox.io/s/webgl-interpolation-forked-etfl1?file=/main.js but the contours do not appear on https://deploy-preview-13365--ol-site.netlify.app/examples/webgl-interpolation.html with Chrome.

PR #13382, opened 2022-02-14
Fix wrong types when using `"skipLibCheck": false`

seravifer:
Fix wrong types when compile an Angular app with `"skipLibCheck": false'` an `"allowSyntheticDefaultImports": true` on `tsconfig.json`.

Without this fixes build:
Captura de pantalla 2022-02-14 a las 13 05 28

### DIFF
--- a/src/ol/style/literal.js
+++ b/src/ol/style/literal.js
@@ -29,7 +29,7 @@ export const SymbolType = {
 
 /**
  * @typedef {Object} LiteralSymbolStyle
- * @property {ExpressionValue|Array<ExpressionValue, ExpressionValue>} size Size, mandatory.
+ * @property {ExpressionValue|Array<ExpressionValue>} size Size, mandatory.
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Thanks, @seravifer!

src/ol/style/literal.js
```diff
@@ -29,7 +29,7 @@ export const SymbolType = {
 
 /**
  * @typedef {Object} LiteralSymbolStyle
- * @property {ExpressionValue|Array<ExpressionValue, ExpressionValue>} size Size, mandatory.
+ * @property {ExpressionValue|Array<ExpressionValue>} size Size, mandatory.
```
tschaub, line 32: I think the intention here was to either allow a single expression or an array of two expressions (and not a longer array).  I wonder if the right tuple syntax would have been `[ExpressionValue, ExpressionValue]`.

seravifer: Yes, i think you are correct. The size can be a value or a tuple. I can correct it in another PR.

PR #13383, opened 2022-02-14
User projection API

ahocevar:
This pull request suggests adding `setUserProjection()` and `useGeographic()` to the API. It also adds a one-time warning when there is a chance that a user gets started with incorrect coordinates.

### DIFF
--- a/src/ol/proj.js
+++ b/src/ol/proj.js
@@ -539,18 +539,17 @@ let userProjection = null;
 
 /**
  * Set the projection for coordinates supplied from and returned by API methods.
- * Note that this method is not yet a part of the stable API.  Support for user
- * projections is not yet complete and should be considered experimental.
+ * This includes all API methods except for those interacting with tile grids.
  * @param {ProjectionLike} projection The user projection.
+ * @api
  */
 export function setUserProjection(projection) {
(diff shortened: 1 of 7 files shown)


src/ol/proj.js
```diff
@@ -598,6 +609,20 @@ export function toUserCoordinate(coordinate, sourceProjection) {
...
+      !equals(coordinate, [0, 0]) &&
+      coordinate[0] >= -180 &&
+      coordinate[0] <= 180 &&
+      coordinate[1] >= -90 &&
+      coordinate[1] <= 90
+    ) {
+      showCoordinateWarning = false;
+      // eslint-disable-next-line no-console
+      console.warn(
+        'Call ol/proj#useGeographic() once to work with [longitude, latitude] coordinates.'
```
tschaub, line 623: Maybe “Call useGeorgraphic() from ol/proj …” instead?

tschaub (approved):

ahocevar:
@mike-000 These problems are not specific to the case where `useGeographic()` is set. I agree we should add a note to our geometry methods that they assume cartesian coordinates and constant map units. But that's for a separate pull request.

PR #13390, opened 2022-02-16
Fix clone of icon loses imgSize when cache is full

MoonE:
When an `ol/style/Icon` is cloned and the `ol/IconImage` is not in the shared IconImageCache then the configured imgSIze is lost.

This stores the imgSize provided in the constructor as a property of the icon and uses it again for cloning.


ahocevar (approved):
Thanks, @MoonE

PR #13393, opened 2022-02-17
Fix setting View resolution or center to undefined

MoonE:
Setting the resolution to undefined resulted in an invalid center `[NaN, NaN]`.

The check for the warning from #13383 broke setting `ol/View` center to `undefined`, though it arguably never should have called that function with `undefined`. It worked once `showCoordinateWarning` was `false`.
https://github.com/openlayers/openlayers/blob/63fc00902faff8c017c600fbd8c1f04791add5d4/src/ol/proj.js#L610-L614

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1636,7 +1636,9 @@ class View extends BaseObject {
    * @api
    */
   setCenter(center) {
-    this.setCenterInternal(fromUserCoordinate(center, this.getProjection()));
+    this.setCenterInternal(
+      center ? fromUserCoordinate(center, this.getProjection()) : center
+    );
   }
 
   /**
--- a/src/ol/centerconstraint.js
+++ b/src/ol/centerconstraint.js
@@ -18,52 +18,54 @@ export function createExtent(extent, onlyCenter, smooth) {
   return (
     /**
      * @param {import("./coordinate.js").Coordinate|undefined} center Center.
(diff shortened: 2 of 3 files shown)


ahocevar:
@MoonE I think I'm failing to understand the motivation for this pull request. Can you explain what #13383 broke exactly? There is a test for `setCenter(undefined)`, which still passes: https://github.com/openlayers/openlayers/blob/main/test/browser/spec/ol/View.test.js#L566-L573. And I don't think we want to allow `setResolution(undefined)`.

ahocevar (approved):
Thanks for the explanation, @MoonE. This makes perfect sense.

Utkrisht:
@MoonE can you please help me with an issue? I'm working with Openlayers and in this line `var viewResolution = (view.getResolution());` I'm getting an error which says 
```
Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.
```
Can you please help me solve  this?

MoonE:
Looks like the example code you provided is incomplete as it should not produce an error in that case.

If you are sure a resolution has been set you can cast it to to `number`.

If you are writing TypeScript:
```ts
(view.getResolution() as number)
```
with plain JavaScript do:
```js
/** @type {number} */ (view.getResolution())
```

Utkrisht:
Okay, thanks a lot!

Issue #13397, opened 2022-02-19
Static Image is not cleared when change projection


yonda-yonda:
**Describe the bug**
Please watch this movie.
![154806128-df2bc1de-eef7-4bf0-8dee-bfb1dc50b05a](https://user-images.githubusercontent.com/7313904/154807195-824606e1-23d4-41b5-a9f0-a44e943418dd.gif)

**To Reproduce**
1. Set Image Source, projection is EPSG:4326, extent is [0,0,10,10].
2. First, render with EPSG:3857. Displayed at the image at correct position.
3. Change map projection to EPSG:3413. At this time, center is [0, 0], zoom is 4 (Avoid including point of [lon, lat]=[10, 10] in screen).
4. The image remain in the center, should disappear. 

**Expected behavior**
Image is cleared when prepare frame.

**sample**
I checked ol@v6.12.0.

``` js
import Map from 'ol/Map';
import {
  Image as ImageLayer,
  Tile as TileLayer
} from 'ol/layer';
import Static from 'ol/source/ImageStatic';
import View from 'ol/View';
import proj4 from 'proj4';
import {
  OSM,
} from 'ol/source';
import {
  getCenter,
} from 'ol/extent';
import {
  get as getProjection,
} from 'ol/proj';
import {
  register
} from 'ol/proj/proj4';
proj4.defs(
  'EPSG:3413',
  '+proj=stere +lat_0=90 +lat_ts=70 +lon_0=-45 +k=1 ' +
  '+x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs'
);
register(proj4);
const proj3413 = getProjection('EPSG:3413');
proj3413.setExtent([-4194304 * 10, -4194304 * 10, 4194304 * 10, 4194304 * 10]);

const layers = {};
layers['osm'] = new TileLayer({
  source: new OSM(),
});
layers['osm'].setProperties({
  name: "osm"
})

layers['static'] = new ImageLayer({
  source: new Static({
    url: 'https://openlayers.org/assets/theme/img/logo70.png',
    crossOrigin: '',
    projection: 'EPSG:4326',
    imageExtent: [0, 0, 10, 10],
    interpolate: true,
  })
})

const map = new Map({
  layers: [layers['osm'], layers['static']],
  target: 'map',
  view: new View({
    projection: 'EPSG:3857',
    center: [0, 0],
    zoom: 4,
  }),
});
const viewProjSelect = document.getElementById('view-projection');

[...]

PR #13398, opened 2022-02-19
Clear image when source's image is not ready

yonda-yonda:
In some cases, Static Image should disappear is not cleared.

Set renderer image_ to null when source's image is not ready, for example, the status is empty.

close https://github.com/openlayers/openlayers/issues/13397


ahocevar (approved):
Thanks, @yonda-yonda

PR #13415, opened 2022-02-24
Check for graticule resolution change and fix unrotated size

### LINKED ISSUE #13400: Graticule labels are misplaced when the map is rotated 45 degrees
**Describe the bug**

It seems like the graticule labels are misplaced when the map is rotated exactly at 45 degrees -- see [this example](https://codesandbox.io/s/graticule-forked-3qz4fe). Attached is a screenshot that shows how the graticule labels are drawn on my machine (macOS 12.1, Firefox 97). Changing the rotation to 44.9 or 45.1 degrees is okay. Other angles that are problematic: -45 degrees, -135 degrees and 135 degrees.

Screenshot 2022-02-21 at 10 37 55

**To Reproduce**

1. Open [this code sandbox](https://codesandbox.io/s/graticule-forked-3qz4fe)
2. Note that the graticule labels are misplaced (they are in the middle of the map, not at the edges)

**Expected behavior**

The graticule labels should be placed at the edges, just like with 44.9 or 45.1 degrees or any other angle that is not 45, -45, 135 or -135 degrees.

**Additional info**

I tried to fix this myself but I haven't figured it out yet; one thing that I've noticed is that the `extent` argument passed into `Graticule.getParallelPoint_` does not seem like a valid extent because `extent[0] > extent[2]`. This happens only for 45, -45, 135 or -135 degrees.

### LINKED ISSUE #13413: Graticule interval does not reset when zooming out if extent is set
**Describe the bug**
The graticule interval does not reset correctly when zooming out if an extent is set.

**To Reproduce**
Set an extent on the Map Graticule example so only the 5 and 10 degrees meridians and 45 and 50 degree parallels are visible https://codesandbox.io/s/graticule-forked-fhqdpt?file=/main.js  Zoom in until more lines are visible at 2 and then 1 degree intervals.  Then zoom out again - the interval remains at 2 degrees.

**Expected behavior**
The extent should not affect the interval which should depend only on resolution, and interval should be the same whether the resolution is reached by zooming or in by zooming out.

mike-000:
Fixes #13400
Fixes #13413 

Calculating the "unrotated" width and height from the frame size and resolution is simpler than trigonometry and avoids random values when results used in division are close to zero.

Also check for changes to resolution as well as extent when the rendered extent fits inside the frame to ensure the interval is updated when zooming out.

### DIFF
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -20,7 +20,6 @@ import {
   containsExtent,
   equals,
   getCenter,
-  getHeight,
   getIntersection,
   getWidth,
   intersects,
@@ -484,8 +483,15 @@ class Graticule extends VectorLayer {
 
     /**
      * @type {?import("../extent.js").Extent}
+     * @private
      */
     this.renderedExtent_ = null;
+
+    /**
+     * @type {?number}
+     * @private
+     */
+    this.renderedResolution_ = null;
 
     this.setRenderOrder(null);
   }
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #13434, opened 2022-03-01
Fix rendercomplete with invisible WebGLPoints layer

### LINKED ISSUE #13423: New `getLoadingOrNotReady` method since v6.13 failing when `getRenderer` does not return an object
**Describe the bug**
OL 6.13 introduced a new method `getLoadingOrNotReady` in `PluggableMap` that expects `layer.getRenderer()` to always return an object.

The basic `ol/layer/Layer`, however returns `null`, which causes errors.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/mapbox-layer.html
2. Open dev tools
3. See error

**Expected behavior**
The library either should not expect `layer.getRenderer()` to always return an object or improve the checking in `getLoadingOrNotReady`.

M393:
When a WebGLPoints layer is not yet visible no rendercomplete event fires.

I already included the commit from #13424 to avoid a merge conflict.
Replaces #13424
Fixes #13423

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -950,13 +950,16 @@ class PluggableMap extends BaseObject {
   getLoadingOrNotReady() {
     const layerStatesArray = this.getLayerGroup().getLayerStatesArray();
     for (let i = 0, ii = layerStatesArray.length; i < ii; ++i) {
-      const layer = layerStatesArray[i].layer;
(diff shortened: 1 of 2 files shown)


M393:
@ahocevar Test added.

ahocevar (approved):

ahocevar:
Thanks for the added test, @M393

PR #13437, opened 2022-03-02
Guard PluggableMap against null renderer

### LINKED ISSUE #13436: Cannot read properties of null (reading 'forEachFeatureAtCoordinate')
**Describe the bug**
`forEachFeatureAtPixel` errored in our application. `Cannot read properties of null (reading 'forEachFeatureAtCoordinate')`
I would expect (not 100% sure) that it was on a map where the target had just been removed. 

https://github.com/openlayers/openlayers/blob/main/src/ol/PluggableMap.js#L635
**To Reproduce**
```
map.renderSync();
map.setTarget(undefined);
map.forEachFeatureAtPixel([1, 2], () => {});
```

**Expected behavior**
No error. Would it make sense to guard again a null `renderer_` in PluggableMap or is it unsupported to call functions on the map when it has no target?

EvertEt:
Closes #13436

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -391,7 +391,7 @@ class PluggableMap extends BaseObject {
     this.overlayIdIndex_ = {};
 
     /**
-     * @type {import("./renderer/Map.js").default}
+     * @type {import("./renderer/Map.js").default|null}
      * @private
      */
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @EvertEt

PR #13439, opened 2022-03-03
WIP: Snap to intersection

langpavel:
This adds new `intersection` option to Snap interaction.


src/ol/interaction/Snap.js
```diff
@@ -68,8 +69,6 @@ function getFeatureFromEvent(evt) {
   }
 }
 
-const tempSegment = [];
```
tschaub, line 71: This is not global, instead it is local to this module.  Declaring and assigning it here avoids the unnecessary allocation on every iteration in the loops below.

langpavel: Ok, got it, I will drop first commit

PR #13443, opened 2022-03-03
Clarify docs on color values

tschaub:
This adds to the documentation for the `color` expression, clarifying that rgb values are 0-255 and alpha is 0-1.


tschaub:
Wow, I didn't even read the sentence below that already notes this.

PR #13444, opened 2022-03-04
Ensure FullScreen button has classname set on render.

themoffster:
When the FullScreen icon is first rendered, the button within it is not having the inactive classname set. The class name is set on toggling fullscreen on/off - just not being set on render.

**After render (classname not set)**

after render

**After entering full screen (classname correctly set)**

after full screen click

**After existing fullscreen (classname correctly set)**

after exiting fullscreen


ahocevar (approved):
Thanks, @themoffster

PR #13455, opened 2022-03-07
Improve Snap interaction performance

MoonE:
This corrects a small bug with the extent of multi point segments, which was always set to the full multi point extent and it now loads all segments of a single feature in one load operation into the rbush which is a faster when loading and should also be a bit faster for querying.
https://github.com/mourner/rbush#bulk-inserting-data

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -625,7 +625,7 @@ class Snap extends PointerInteraction {
         feature: feature,
         segment: [point],
       };
-      this.rBush_.insert(geometry.getExtent(), segmentData);
+      this.rBush_.insert(boundingExtent(segmentData.segment), segmentData);
     });
   }
 
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -181,18 +181,18 @@ class Snap extends PointerInteraction {
     /**
      * @const
      * @private
-     * @type {Object<string, function(import("../Feature.js").default, import("../geom/Geometry.js").default): void>}
+     * @type {Object<string, function(Array<Array<import('../coordinate.js').Coordinate>>, import("../geom/Geometry.js").default): void>}
      */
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Looks very solid. Thanks, @MoonE!

src/ol/interaction/Snap.js
```diff
@@ -553,134 +568,101 @@ class Snap extends PointerInteraction {
...
   }
 
   /**
-   * @param {import("../Feature.js").default} feature Feature
+   * @param {Array<Array<import('../coordinate.js').Coordinate>>} segments Segments
    * @param {import("../geom/GeometryCollection.js").default} geometry Geometry.
    * @private
    */
-  writeGeometryCollectionGeometry_(feature, geometry) {
+  segmentGeometryCollectionGemetry_(segments, geometry) {
```
ddanny, line 580: This method has a typo (missing "o" in Geometry)

src/ol/interaction/Snap.js
```diff
@@ -553,134 +568,101 @@ class Snap extends PointerInteraction {
...
   }
 
   /**
-   * @param {import("../Feature.js").default} feature Feature
+   * @param {Array<Array<import('../coordinate.js').Coordinate>>} segments Segments
    * @param {import("../geom/LineString.js").default} geometry Geometry.
    * @private
    */
-  writeLineStringGeometry_(feature, geometry) {
+  segmentLineStringGemetry_(segments, geometry) {
```
ddanny, line 595: Same here with "Gemetry". Was the use of this typo intended? I see multiple methods using the same naming scheme.

MoonE: I used a regular expression to change all of the method names, `write(LineString)Geometry` -> `segment$1Gemetry` thats why all are missing the o in Geometry.

So yeah, it makes sense to fix the typo.

Issue #13414, opened 2022-02-24
Vector tile performance - compared to MapBox


MortenHofft:
I struggle with performance when rendering my vector tile point layers. It isn't terrible at all, but when compared to MapBox there is a notable difference: MapBox (gl 2.7) feels smooth, OpenLayers (6.12) feels a bit more flickering and unresponsive. Edges are jagged when zooming. Tiles slower to show. Are there any settings or similar I can do to improve the performance?

My dream scenario is to use OpenLayers (especially for the projection support), but have as smooth a rendering as MapBox.

Example comparing the two: https://codepen.io/hofft/pen/dyZKezB?editors=0010

Thank you for any help and advice

ahocevar:
Mapbox GL uses a very efficient WebGL renderer, which we cannot beat with OpenLayers's Canvas renderer. You may be able to improve things a bit by configuring the `MapboxVector` layer with `renderMode: 'vector'`. This will result in a bit more lag during panning, but better updates during zooming.

MortenHofft:
Thank you for your quick response. I have a follow up question, feel free to say if this is better directed at StackOverflow.

The examples and API docs mention webgl often e.g. https://openlayers.org/en/latest/examples/webgl-tiles.html

I also notice [ol/layer/WebGLPoints example](https://openlayers.org/en/latest/examples/webgl-points-layer.html) using a GeoJSON - but I'm not having any luck when I try to modify it to use a vector tile source instead. Is webgl support something that is coming? If I'm reading [the docs](https://openlayers.org/en/latest/apidoc/module-ol_layer_WebGLTile-WebGLTileLayer.html) correctly, then it is currently only for raster sources?

ahocevar:
The WebGLPoints layer is for vector sources only, and the WebGLTile layer for tile sources. There is no WebGL based renderer that can render vector tiles.

MortenHofft:
Feel free to close this issue, or leave it open as a feature request for a "WebGL based renderer that can render vector tiles"

Issue #12868, opened 2021-10-08
WebGL Line Layer


lujiayou123:
Could you provide WebGL Lines Layer? something like webglPointsLayers?

Issue #12658, opened 2021-08-24
webglpointlayer feature


abigstupidbear:
hi. can webglpointlayer support these？
1 draw line and polygon.
2 draw text 
3 support wrapx

if not . do you have any plan to achieve these？
wating for your reply.

jahow:
> hi. can webglpointlayer support these？
> 1 draw line and polygon.
> 2 draw text
> 3 support wrapx

Not yet. For 1 I have planned work on an experimental renderer in the following months (no promise). 2 is more unlikely considering how much trouble text rendering in webgl is. 3 is definitely doable right now, although there is no plan for it that I know of.

PR #13461, opened 2022-03-09
WebGL vector renderer for polygons, lines and points

jahow:
This PR introduces a new WebGL-based renderer able to handle polygons, lines and points, building up and expanding on the existing `WebGLPointsLayerRenderer` class.

Partially addresses #13414, https://github.com/openlayers/openlayers/issues/12868, https://github.com/openlayers/openlayers/issues/12658

Important caveats with this WIP branch:
* ~Lots of tests still need to be written :sob:~
* No hit detection on shapes
* No support for circle geometries
* No text!
* This does not include a higher-level layer type with styling options; instead, the renderer expects to be fed raw GLSL programs for each type of geometry
* All the webgl rendering logic assumes that geometries have a stride of 2; I honestly don't know how far this might have consequences

https://deploy-preview-13461--ol-site.netlify.app/examples/webgl-vector-layer.html

webgl-ol

A lot of time was actually spent on the `MixedGeometryBatch` class, because it has to track all geometries and store them in the appropriate "buckets" (lines, points, polygons) while keeping track of the vertices count, rings count, etc. This is then used to generate render instructions by the `BatchRenderer` classes.

Once render instructions are done, they are transformed into WebGL buffers by the worker, same as points. This is especially interesting here since we can have costly operations (triangulating polygons, computing lines offsets) off the main thread. Although to be fair, [@mapbox/earcut](https://github.com/mapbox/earcut) is so fast that the different for polygons is probably negligible.


jahow:
This is now ready for review. The `WebGLVectorLayerRenderer` is still quite simple and not tested yet (either unit or rendering), I think it will be subject to change later on so I didn't invest too much time on it.

For now I feel like this PR lays solid foundations for a more generic use of WebGL with vector data, so it makes sense as it is IMO. A vector tile renderer should be able to instantiate one `MixedGeometryBatch` per tile and use that to generate buffers and render each tile in three draw calls (polygons, lines, points).

I'm still planning to do another PR specifically for lines to showcase what will be possible with this new rendering logic (and because the ecoregions example really doesn't give them justice!).

PS: there's a weird error on the build about an invalid import, no idea where that comes from (probably from the worker), if anyone has an idea please share :slightly_smiling_face:

src/ol/renderer/webgl/VectorLayer.js
```diff
@@ -0,0 +1,344 @@
...
+ *
+ * Please note that the main shader output should have premultiplied alpha, otherwise visual anomalies may occur.
+ *
+ * Polygons are broken down into triangles using the @mapbox/earcut package.
+ * Lines are rendered into strips of quads.
+ *
+ *
+ * This uses {@link module:ol/webgl/Helper~WebGLHelper} internally.
+ *
+ * @api
```
tschaub, line 93: I think we should keep this "experimental" (not tagged with `@api`) for now.

src/ol/webgl/Helper.js
```diff
@@ -805,22 +798,20 @@ class WebGLHelper extends Disposable {
   }
 
   /**
-   * Use a program.  If the program is already in use, this will return `false`.
+   * Set up a program for use. The program will be set as the current one. Then, the uniforms used
+   * in the program will be set based on the current frame state and the helper configuration.
    * @param {WebGLProgram} program Program.
-   * @return {boolean} Changed.
+   * @param {import("../PluggableMap.js").FrameState} frameState Frame state.
    * @api
```
tschaub, line 805: Here is an example of where it was premature to add this to the `@api`.  This is technically a breaking change (adding a required argument).

I think that the helper and WebGL layer renderers should be internal for now.  This gives us the ability to continue improving the higher level API (layers and sources) without breaking applications.

src/ol/renderer/webgl/VectorLayer.js
jahow: Totally agree, I'll change it; guess I wasn't really paying attention to API consistency while working on this, as it is obviously in the experimental field for now

jahow:
I've made a pass on the new renderer as well as all `@api` tags in the webgl utilities (`ol/webgl/*`) and removed them all. The following classes were part of the API:
* WebGLHelper
* WebGLRenderTarget
* WebGLArrayBuffer
* WebGLPostProcessingPass

(Note that I didn't touch the other renderers)

Not sure why these were included in the API originally, it looks like I was a little too enthusiastic back then. To me these are clearly too low level to be part of the API and it was a mistake. Hopefully we can now backpedal on this and let the few people (if any) that rely on it know that function signatures might change in the future?

ahocevar:
By removing `@api` annotations from functions that were `@api` in releases already, we have to bump the major version just the same way as when we make a breaking change to something that has an `@api` annotation.

src/ol/webgl/Buffer.js
```diff
@@ -35,7 +35,6 @@ export const BufferUsage = {
  * Note:
  * See the documentation of [WebGLRenderingContext.bufferData](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bufferData)
  * for more info on buffer usage.
- * @api
```
ahocevar, line 38: Did your pull request change anything in WebGLArrayBuffer? If not you could leave this `@api` annotation here, and we can remove it for the next major release.

src/ol/webgl/Helper.js
```diff
@@ -302,8 +304,6 @@ function releaseCanvas(key) {
  *   ```
  *
  * For an example usage of this class, refer to {@link module:ol/renderer/webgl/PointsLayer~WebGLPointsLayerRenderer}.
- *
- * @api
```
ahocevar, line 306: Do existing applications continue work with the changes you made to the WebGL helper, e.g. without passing a `frameState` to `useProgram()`? If so, this `@api` annotation and those for the methods can stay for now.

ahocevar (commented):
@jahow I added a few questions regarding the API. My goal is to figure out if we need to bump the major version before merging this pull request.

Also, would you be able to resolve merge conflicts to keep this pull request up to date?

Like @tschaub already said, if we keep the API experimental, we can start with what you have, like we did for the WebGLPoints layer. I think that worked well. It gives everyone a chance to get involved with the new code more easily, and learn about how the WebGL specifics work while playing with applications that make use of it. Like @tschaub and I did for the FOSS4G-NA workshop in Dan Diego.

ahocevar:
@jahow Currently the webgl-vector-layer example does not render before I interact with the map in some way (pan/zoom). Can you maybe take a look at this after removing the conflicts?

src/ol/webgl/Helper.js
ahocevar: Or we keep it like you have it (with the @api annotations removed), and publish v7.0. What do others think?

jahow:
> @jahow Currently the webgl-vector-layer example does not render before I interact with the map in some way (pan/zoom). Can you maybe take a look at this after removing the conflicts?

Defintely, I picked up work on it this week to address that and then rebase the branch. There is also one last issue that I identified with this PR: on some systems (specifically mobile), the lines do not render. I suspect that this is because of the amount of primitives which are required, but not 100% sure and we need a way to handle that gracefully anyway.

src/ol/webgl/Buffer.js
jahow: yep, that could work

src/ol/webgl/Helper.js
jahow: Unfortunately now, this change to the helper is a breaking change, and a necessary one (although it may be worth it to investigate some more)

examples/webgl-vector-layer.js
```diff
@@ -0,0 +1,182 @@
...
+          float v = a_index == 0.0 || a_index == 1.0 ? 0.0 : 1.0;
+          v_texCoord = vec2(u, v);
+        }`,
+      pointFragmentShader: `
+        precision mediump float;
+        varying vec2 v_texCoord;
+
+        void main(void) {
+            gl_FragColor = vec4(1.0, 1.0, 1.0, 1.0);
+        }`,
```
tschaub, line 149: What do you think about having the shaders above be the defaults?  And we could have dark gray `color` attribute by default below.  Then people could try out the experimental renderer without having to copy/paste the shader code from this example.  And then they could experiment with their own color attribute as a first modification.

This feels like it would be a gentler introduction to the new functionality.  And since it is all experimental, we could still change things as we work on the higher level layer styling API.

jahow: That would make sense; maybe having a `width` parameter as well for the lines.

jahow:
> Can you maybe take a look at this after removing the conflicts?

This has been fixed with the latest commit; also the branch was rebased.

bozhidarc:
Hi @jahow, I was running your example on iOS device (15.4) and I notice that the borderers are not shown.
I'm not sure if this is a known issue, so I'm letting you know.

IMG_8B7724398CC0-1

jahow:
Thanks for the report @bozhidarc, this is currently something that I'm looking into!

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,204 @@
...
+    this.helper_.drawElements(0, renderCount);
+  }
+
+  /**
+   * Rebuild rendering instructions based on the provided frame state
+   * This is specific to the geometry type and has to be implemented by subclasses.
+   * @param {import("./MixedGeometryBatch.js").GeometryBatch} batch Geometry batch
+   * @protected
+   */
+  generateRenderInstructions_(batch) {
```
tschaub, line 118: The current naming pattern is to only use the `_` suffix for `@private` names.

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,204 @@
...
+    abstract();
+  }
+
+  /**
+   * Rebuild internal webgl buffers for rendering based on the current rendering instructions;
+   * This is asynchronous: webgl buffers wil _not_ be updated right away
+   * @param {import("./MixedGeometryBatch.js").GeometryBatch} batch Geometry batch
+   * @param {import("../../geom/GeometryType.js").default} geometryType Geometry type
+   * @param {function(): void} callback Function called once the render buffers are updated
+   * @protected
```
tschaub, line 128: I don't see this being called elsewhere.  Should it be `@private` instead?

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,204 @@
...
+    /** @type {import('./constants.js').WebGLWorkerGenerateBuffersMessage} */
+    const message = {
+      id: messageId,
+      type: messageType,
+      renderInstructions: batch.renderInstructions.buffer,
+      renderInstructionsTransform: batch.renderInstructionsTransform,
+      customAttributesCount: this.customAttributes_.length,
+    };
+    // additional properties will be sent back as-is by the worker
+    message['projectionTransform'] = batch.renderInstructionsTransform;
```
tschaub, line 157: Curious about the bracket notation here.  Are the other message properties getting renamed?

jahow:
> Hi @jahow, I was running your example on iOS device (15.4) and I notice that the borderers are not shown.
I'm not sure if this is a known issue, so I'm letting you know.

This was a pixel ratio issue, fixed with the latest commit.

tschaub:
> This was a pixel ratio issue, fixed with the latest commit.

Does the shader also need an update so that the stroke width is normalized (to a fixed CSS pixel width)?

stroke-width

The stroke is currently wider on low device pixel ratio.

jahow:
> Does the shader also need an update so that the stroke width is normalized (to a fixed CSS pixel width)?

Good point, the shader needs to be adapted indeed. I'm going to add a `u_pixelRatio` uniform to the shaders, originally I wanted to avoid having this information in the shaders but since we're relying on the `gl_FragCoord` built-in vector I guess there's no way around it.

src/ol/render/webgl/BatchRenderer.js
jahow: Good catch, this was a useless leftover from the points layer renderer (originally I used a bracket notation for type checking reasons, i.e. setting a property that was not part of the defined type).

Parzavil3:
Im blown away by this

examples/webgl-vector-layer.js
jahow: Done in the latest commits, there is now a separate module providing default shaders expecting attributes such as color, opacity and width. This has greatly simplified the example.

Hopefully the process of providing custom shaders isn't too obscure and error-prone, and relying on default ones should let people try this renderer without too much trouble!

jahow:
This is ready for final review. I've tried to improve the API of the renderer class to allow relying more on default shaders and attributes, @tschaub please let me know if that fits what you had in mind, thanks!

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
+/**
+ * @module ol/renderer/webgl/shaders
+ */
+import {asArray} from '../../color.js';
+
+/**
+ * Attribute names used in the default shaders.
+ * @enum {string}
+ */
+export const DefaultAttributes = {
```
ahocevar, line 10: It would be nice to not introduce new string enums. Using a union type like
```js
/** @typedef {'color'|'opacity'|'width'} DefaultAttributes */
```
would also make the example look nicer, if the shader`s attribute were typed like
```js
/** @typedef {Object<DefaultAttributes, (import("ol/Feature").default, Object<string, *>) => number} ShaderAttributes */
```

examples/webgl-vector-layer.js
```diff
@@ -0,0 +1,68 @@
...
+            return packColor(color);
+          },
+          [DefaultAttributes.OPACITY]: function () {
+            return 0.6;
+          },
+        },
+      },
+      lineStringShader: {
+        attributes: {
+          [DefaultAttributes.COLOR]: function (feature, properties) {
```
ahocevar, line 33: Wouldn't it make more sense to have a `function (feature, resolution)` here, to allow for resolution based styling? Or at least a `function (feature)`, because the properties could be retrieved from the feature with e.g. `feature.get('COLOR')`? Or are the `properties` different from the feature's properties?

jahow: Having `resolution` here wouldn't be really useful since this is only evaluated when the webgl buffers are rebuilt (typically on source change); instead, the `u_resolution` uniform is available in the shaders.

As for the `properties` object, it's there to save some time by doing `properties['my-prop']` instead of `feature.get('my-prop')`. Maybe this can be made more explicit?

src/ol/render/webgl/MixedGeometryBatch.js
```diff
@@ -0,0 +1,368 @@
...
+   * @param {import("../../Feature").default} feature Feature
+   * @return {GeometryBatchItem} Batch item added (or existing one)
+   * @private
+   */
+  addFeatureEntryInPointBatch_(feature) {
+    const uid = getUid(feature);
+    if (!(uid in this.pointBatch.entries)) {
+      this.pointBatch.entries[uid] = {
+        feature: feature,
+        properties: feature.getProperties(),
```
ahocevar, line 165: `Feature.getProperties()` (inherited from `Object.getProperties()` is a bit inefficient, because it returns a copy of the object's `valuees_`. Since this is only used to pass to the shader callback as 2nd argument, without the guarantee that the user needs it at all, I'd suggest removing that second argument and let the user call `feature.get('myProperty')` in the callback.

examples/webgl-vector-layer.js
ahocevar: Ok, thanks for the explanation. I added another in-line comment to explain why I don't think that passing the properties as second argument makes much sense. As for the resolution, i think it would make sense to have an example that shows how to use that, even if it is just for now until we have a better API. If users can see how to do property and resolution based styling, we've already achieved a lot.

jahow:
> introducing basic style expressions like for the WebGL Points renderer would be a low hanging fruit, and would be nicer from a user's perspective than passing the configuration for a PolygonShader (which means fill style) and a LineStringShader (which means line style)

Hmm, I'm not sure how to offer the possibility to provide custom shaders while also offering these simple style parameters. Will give this some thought, please don't hesitate to throw any ideas!

ahocevar:
@jahow If nothing else, I think it would make styling clearer if `polygonShader` would be renamed to `fill`, and `lineStringShader` to `line` to start with. To make clear we're not dealing with geometries, but with symbolizers here.

Later on, there would also be a need to control the rendering order regardless of symbolizer type.

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,201 @@
...
+   * @param {string} vertexShader Vertex shader
+   * @param {string} fragmentShader Fragment shader
+   * @param {Array<CustomAttribute>} customAttributes List of custom attributes
+   */
+  constructor(helper, worker, vertexShader, fragmentShader, customAttributes) {
+    /**
+     * @type {import("../../webgl/Helper.js").default}
+     * @protected
+     */
+    this.helper_ = helper;
```
tschaub, line 41: Since this is not used elsewhere, it could be `@private`.  I pushed an additional commit to change this.

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * @type {Worker}
+     * @protected
+     */
+    this.worker_ = worker;
```
tschaub, line 47: Since this is not used elsewhere, it could be `@private`.  I pushed an additional commit to change this.

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * @type {WebGLProgram}
+     * @protected
+     */
+    this.program_ = this.helper_.getProgram(fragmentShader, vertexShader);
```
tschaub, line 53: Since this is not used elsewhere, it could be `@private`.  I pushed an additional commit to change this.

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * A list of attributes used by the renderer.
+     * @type {Array<import('../../webgl/Helper.js').AttributeDescription>}
+     * @protected
+     */
+    this.attributes_ = [];
```
tschaub: Since this is used elsewhere, it should not be named with a `_` suffix.  I pushed an additional commit to change this.

src/ol/render/webgl/BatchRenderer.js
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * @type {Array<CustomAttribute>}
+     * @protected
+     */
+    this.customAttributes_ = customAttributes;
```
tschaub: Since this is used elsewhere, it should not be named with a `_` suffix.  I pushed an additional commit to change this.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+  OPACITY: 'opacity',
+  WIDTH: 'width',
+};
+
+/**
+ * Packs red/green/blue channels of a color into a single float value; alpha is ignored.
+ * This is how DefaultAttributes.COLOR is expected to be computed.
+ * @param {import("../../color.js").Color|string} color Color as array of numbers or string
+ * @return {number} Float value containing the color
+ * @api
```
tschaub, line 21: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+  fract(floor(a_color / 256.0 / 256.0) / 256.0),
+  fract(floor(a_color / 256.0) / 256.0),
+  fract(a_color / 256.0)
+);`;
+
+/**
+ * Default polygon vertex shader.
+ * Relies on DefaultAttributes.COLOR and DefaultAttributes.OPACITY.
+ * @type {string}
+ * @api
```
tschaub, line 41: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+  void main(void) {
+    gl_Position = u_projectionMatrix * vec4(a_position, 0.0, 1.0);
+    v_color = ${DECODE_COLOR_EXPRESSION}
+    v_opacity = a_opacity;
+  }`;
+
+/**
+ * Default polygon fragment shader.
+ * @type {string}
+ * @api
```
tschaub, line 61: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+
+  void main(void) {
+    gl_FragColor = vec4(v_color, 1.0) * v_opacity;
+  }`;
+
+/**
+ * Default linestring vertex shader.
+ * Relies on DefaultAttributes.COLOR, DefaultAttributes.OPACITY and DefaultAttributes.WIDTH.
+ * @type {string}
+ * @api
```
tschaub, line 76: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+    v_segmentEnd = worldToPx(a_segmentEnd);
+    v_color = ${DECODE_COLOR_EXPRESSION}
+    v_opacity = a_opacity;
+    v_width = a_width;
+  }`;
+
+/**
+ * Default linestring fragment shader.
+ * @type {string}
+ * @api
```
tschaub, line 142: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+    vec2 v_currentPoint = gl_FragCoord.xy / u_pixelRatio;
+    gl_FragColor = vec4(v_color, 1.0) * v_opacity;
+    gl_FragColor *= segmentDistanceField(v_currentPoint, v_segmentStart, v_segmentEnd, v_width);
+  }`;
+
+/**
+ * Default point vertex shader.
+ * Relies on DefaultAttributes.COLOR and DefaultAttributes.OPACITY.
+ * @type {string}
+ * @api
```
tschaub, line 173: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/shaders.js
```diff
@@ -0,0 +1,213 @@
...
+    float v = a_index == 0.0 || a_index == 1.0 ? 0.0 : 1.0;
+    v_texCoord = vec2(u, v);
+    v_color = ${DECODE_COLOR_EXPRESSION}
+    v_opacity = a_opacity;
+  }`;
+
+/**
+ * Default point fragment shader.
+ * @type {string}
+ * @api
```
tschaub, line 204: Since this is experimental, I removed this `@api` annotation in an additional commit.

src/ol/renderer/webgl/VectorLayer.js
tschaub: I pushed an additional commit to remove this `@api` annotation.

tschaub:
> @jahow If nothing else, I think it would make styling clearer if `polygonShader` would be renamed to `fill`, and `lineStringShader` to `line` to start with. To make clear we're not dealing with geometries, but with symbolizers here.
> 
> Later on, there would also be a need to control the rendering order regardless of symbolizer type.

I agree.  I pushed another commit that renames `polygonShader` to `fill` and `lineStringShader` to `stroke`.

src/ol/render/webgl/MixedGeometryBatch.js
tschaub: I agree it makes sense to keep the attribute callback signature minimal for now.  I pushed a commit that removes the second argument and removes this `properties` object from the `GeometryBatchItem` type.

src/ol/renderer/webgl/shaders.js
tschaub: I agree.  I've pushed another commit that makes this change.

examples/webgl-vector-layer.js
tschaub: I pushed a change that removes the `properties` argument from the attribute getter.  I agree it will be nice to do resolution-dependent styling, but since that will require customization to the shader (instead of just calculating an attribute value in JS), I think we should wait until we have support for compiling shaders from style expressions.

examples/webgl-vector-layer.html
```diff
@@ -0,0 +1,11 @@
...
+layout: example.html
+title: WebGL Vector Layer
+shortdesc: Example of a vector layer rendered using WebGL
+docs: >
+  The ecoregions are loaded from a GeoJSON file.
+tags: "vector, geojson, webgl"
+experimental: true
+---
+<div id="map" class="map"></div>
+<div id="info">&nbsp;</div>
```
tschaub: This was unused, so I pushed a commit that removed it.

examples/webgl-vector-layer.js
```diff
@@ -0,0 +1,65 @@
...
+import VectorSource from '../src/ol/source/Vector.js';
+import View from '../src/ol/View.js';
+import WebGLVectorLayerRenderer from '../src/ol/renderer/webgl/VectorLayer.js';
+import {asArray} from '../src/ol/color.js';
+import {packColor} from '../src/ol/renderer/webgl/shaders.js';
+
+class WebGLLayer extends Layer {
+  createRenderer() {
+    return new WebGLVectorLayerRenderer(this, {
+      className: this.getClassName(),
```
tschaub: The `className` option is unused, so I pushed a commit that removes it from the example.

tschaub (approved):
Thanks for all your work putting together this foundation for WebGL vector rendering, @jahow.  This is a huge contribution to the library.  I'm looking forward to helping build a nice layer styling API on top of it and seeing some new examples that really show off the capabilities.

If you agree with the additional changes I've made here, please merge.

jahow:
Wow, thanks @tschaub, I guess you've addressed everything remaining. No objection at all, it looks much better now!

PR #13484, opened 2022-03-15
Include GeometryCollection in hit detection

e-salmon:


ahocevar:
Thanks for this contribution, @e-salmon. Would you be able to add a test (or extend an existing one) to verify the change?

PR #13492, opened 2022-03-19
Workaround for Safari WebGL issue

### LINKED ISSUE #13486: iOS 15.4 Webgl via Metal issue
**Describe the bug**
The Webgl layer is not panning / zooming with the map.
In iOS 15.4 the Webgl via Metal is enabled by default.

**To Reproduce**
1. Go to https://viewer.bomenwacht.nl/?code=935771DB2CB1AF76AD30B1B4345405E2 with a iOS device.
2. Try to pan / zoom te map

**Expected behavior**
When you turn off Webgl via Metal in the experimental features the Webgl layer is panning / zooming again with the map.

tschaub:
This is a workaround for a WebGL rendering issue in Safari (https://bugs.webkit.org/show_bug.cgi?id=237906).

Fixes #13486.

### DIFF
--- a/src/ol/has.js
+++ b/src/ol/has.js
@@ -20,6 +20,17 @@ export const FIREFOX = ua.indexOf('firefox') !== -1;
 export const SAFARI = ua.indexOf('safari') !== -1 && ua.indexOf('chrom') == -1;
 
 /**
+ * https://bugs.webkit.org/show_bug.cgi?id=237906
+ * @type {boolean}
+ */
+export const SAFARI_BUG_237906 =
+  SAFARI &&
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
This fixes the problem for me. Thanks for your work on this, @tschaub.

tschaub:
It sounds like the Safari fix is [included in version 15.5](https://bugs.webkit.org/show_bug.cgi?id=237906#c17), so we shouldn't have to modify the workaround to handle additional versions.  After 15.4 falls out of use, we should be able to remove the workaround.

PR #13496, opened 2022-03-22
Improve Projection and Scale example calculations

mike-000:
Replaces #13173

Changes to `getPointResolution` can be avoided by ensuring that the Projection and Scale example does not use very large resolutions (1 degree) across which calculating an accurate point resolution would be affected by the curvature of the earth.  By using meters per unit all calculations can be done more accurately at a resolution of approximately 1 meter.  Differences in scale which appeared at the 6th digit are gone, only very small differences at the 8th digit can now be seen. 

https://deploy-preview-13496--ol-site.netlify.app/examples/projection-and-scale.html


ahocevar (approved):
Thanks, @mike-000

PR #13506, opened 2022-03-25
Change imageSmoothing to interpolate in tests

mike-000:
Update rendering tests which still use the depreciated `imageSmoothing` to use `interpolate`.


ahocevar (approved):
Thanks, @mike-000

tschaub:
I amended the commit above to get the status checks to run - they appeared to be stuck on the last commit.

PR #13521, opened 2022-03-29
Include tile gutter in offsets for getData() methods

### LINKED ISSUE #13520: layer.getData() method returns incorrect value if tiles have a gutter
**Describe the bug**
The `getData()` methods for `ol/layer/Tile` and `ol/layer/WebGLTile` return incorrect values if tiles have a gutter.

**To Reproduce**
See https://codesandbox.io/s/wms-tiled-forked-xdu8e6?file=/main.js  The center of the view is red but the logged result shows green.  Comment out (or change to 0) the `gutter` and red is logged.  Change `ol/layer/Tile` to `ol/layer/WebGLTile` and the same set to results are seen.

**Expected behavior**
The gutter need to be added to the tile pixel row and column offsets in both renderers as in https://codesandbox.io/s/wms-tiled-forked-3w4jbd?file=/main.js.

mike-000:
Fixes #13520

### DIFF
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -208,7 +208,11 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
             tileCoord[2] * tileSize[1])
       );
 
-      return this.getImageData(tile.getImage(), col, row);
+      const gutter = Math.round(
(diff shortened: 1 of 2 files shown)


tschaub:
Thanks for the fix, @mike-000.  It looks like tiles with gutters are under-tested.  Please add a test here so we can start improving the coverage.

tschaub (approved):

Issue #12937, opened 2021-10-29
Add a property containing the Feature type (= e.g. GML Element name) to the Feature class


ejn:
# Problem

When parsing a GML document containing mixed feature types - which are not known in advance, e.g. from a GetFeatureInfo request - it is impossible to tell what FeatureType (XML Element) produced each of the resulting features as only the properties of the Feature are recorded, but not the Element name. This is a problem when presenting the GetFeatureInfo results as it cannot be seen from which Feature type the properties come, making the interpretation somewhat difficult.

# Ideas

One easy approach would be to add a special key e.g. `_featureType_` to the properties (similar to the use of object key _content_ when parsing complex features), but this would need extra documentation, and to be handled when writing features.

I propose instead adding a non-observable property to the Feature class, which can be accessed through `getFeatureType()` and set with `setFeatureType()` , similar to `getId()`/`setId()`. The reader methods of the Feature formats may then set the Feature type where appropriate. Additionally, when writing Features, this may be used to set the element (local) name of the written Feature elements (for GML, overriding the feature type defined in the writer, which becomes the default type for features not containing an explicit type) or other properties indicating the feature type in other formats.

# Summary

I'd be happy to knock up PR for this, but before I put the effort in then I'd like to get a bit of feedback as to whether the approach described seems reasonable, or whether anyone has a better suggestion.

PR #12936, opened 2021-10-29
Bugfix for GML parsing with multiple property elements with XML attributes

ejn:
This PR fixes a problem with parsing GML Features containing multiple properties with XML attibutes, e.g.
```xml
<ns:Feature gml:id="f1">
<ns:Link xlink:href="http://example.com/a">
<ns:Link xlink:href="http://example.com/b">
</ns:Feature>
```
Such constructs are found often in INSPIRE themes and in the German AAA model.

I included a test case with some simplified synthetic data. The bug was spotted in the wild whilst attempting to parse the result of a [GetFeatureInfo-Request to a WMS providing AAA-structured responses](
https://www.geodaten-mv.de/dienste/alkis_wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetFeatureInfo&FORMAT=image%2Fpng&TRANSPARENT=true&QUERY_LAYERS=adv_alkis_gesetzl_festlegungen%2Cadv_alkis_weiteres%2Cadv_alkis_gebaeude%2Cadv_alkis_flurstuecke&LAYERS=adv_alkis_gesetzl_festlegungen%2Cadv_alkis_weiteres%2Cadv_alkis_gebaeude%2Cadv_alkis_flurstuecke&INFO_FORMAT=text%2Fxml%3Bsubtype%3Dgml%2F3.2.1&I=50&J=50&CRS=EPSG%3A25833&STYLES=&WIDTH=101&HEIGHT=101&BBOX=293849.5404071358%2C6004185.0609577745%2C293918.4150123862%2C6004253.935563024) (particularly the repeated `adv:gehoertZu`-properties of the `adv:AX_LagebezeichnungOhneHausnummer` object - the depth of the resulting nested array before this fix was causing recursion problems in the browser whilst processing).

PR #13527, opened 2022-03-31
Add featureType property to features

ejn:
This PR implements (most of) the functionality described in #12937:
 - adds a non-observable feature type property to features (comparable to feature ID), with getters/setters `getFeatureType()`/`setFeatureType()`
 - when reading GML, the (local) element name will be parsed to this property

Not (yet?) implemented is the use of the defined individual feature type when writing features, which is somewhat more complicated.

Motivation for this change are service query results (e.g. [this GetFeatureInfo request](https://www.geodaten-mv.de/dienste/alkis_wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetFeatureInfo&FORMAT=image%2Fpng&TRANSPARENT=true&QUERY_LAYERS=adv_alkis_gesetzl_festlegungen%2Cadv_alkis_weiteres%2Cadv_alkis_gebaeude%2Cadv_alkis_flurstuecke&LAYERS=adv_alkis_gesetzl_festlegungen%2Cadv_alkis_weiteres%2Cadv_alkis_gebaeude%2Cadv_alkis_flurstuecke&INFO_FORMAT=text%2Fxml%3Bsubtype%3Dgml%2F3.2.1&I=50&J=50&CRS=EPSG%3A25833&STYLES=&WIDTH=101&HEIGHT=101&BBOX=293849.5404071358%2C6004185.0609577745%2C293918.4150123862%2C6004253.935563024)) containing multiple feature types where a differentiation (not just based on the contained properties) between them is required.

### DIFF
--- a/src/ol/Feature.js
+++ b/src/ol/Feature.js
@@ -98,6 +98,13 @@ class Feature extends BaseObject {
     this.un;
 
     /**
+     * Name of type of feature (e.g. GML Feature element name)
+     * @type {string|undefined}
+     * @api
+     */
+    this.featureType = undefined;
+
+    /**
      * @private
      * @type {number|string|undefined}
      */
@@ -156,6 +163,7 @@ class Feature extends BaseObject {
     const clone = /** @type {Feature<Geometry>} */ (
       new Feature(this.hasProperties() ? this.getProperties() : null)
     );
+    clone.featureType = this.featureType;
     clone.setGeometryName(this.getGeometryName());
     const geometry = this.getGeometry();
     if (geometry) {
--- a/src/ol/format/GMLBase.js
+++ b/src/ol/format/GMLBase.js
@@ -354,6 +354,7 @@ class GMLBase extends XMLFeature {
       if (fid) {
         feature.setId(fid);
       }
+      feature.featureType = node.localName;
       return feature;
     }
   }
--- a/test/browser/spec/ol/format/gml.test.js
+++ b/test/browser/spec/ol/format/gml.test.js
@@ -43,6 +43,10 @@ describe('ol.format.GML2', function () {
     it('reads all features', function () {
       expect(features.length).to.be(3);
     });
+
+    it('reads the names of feature types', function () {
+      expect(features[0].featureType).to.be('water_areas');
+    });
   });
 
   describe('#readGeometry', function () {
@@ -1442,6 +1446,10 @@ describe('ol.format.GML3', function () {
       expect(features[0].getId()).to.equal('states.1');
     });
 
+    it('reads the names of feature types', function () {
+      expect(features[0].featureType).to.be('states');
+    });
+
     it('can reuse the parser for a different featureNS', function () {
       const text =
         '<gml:featureMembers xmlns:gml="http://www.opengis.net/gml">' +
@@ -1504,6 +1512,10 @@ describe('ol.format.GML3', function () {
 
     it('creates the right id for the feature', function () {
       expect(features[0].getId()).to.equal('states.1');
+    });
+
+    it('reads the names of feature types', function () {
+      expect(features[0].featureType).to.be('states');
     });
 
     it('writes back features as GML', function () {
@@ -1682,6 +1694,11 @@ describe('ol.format.GML3', function () {
     it('reads all features', function () {
       expect(features.length).to.be(12);
     });
+
+    it('reads the names of feature types', function () {
+      expect(features[0].featureType).to.be('planet_osm_line');
+      expect(features[11].featureType).to.be('planet_osm_polygon');
+    });
   });
 
   describe('when parsing multiple feature types', function () {
@@ -1727,6 +1744,11 @@ describe('ol.format.GML3', function () {
 
     it('reads all features', function () {
       expect(features.length).to.be(2);
+    });
+
+    it('reads the names of feature types', function () {
+      expect(features[0].featureType).to.be('states');
+      expect(features[1].featureType).to.be('roads');
     });
   });
 
--- a/test/node/ol/Feature.test.js
+++ b/test/node/ol/Feature.test.js
@@ -358,12 +358,21 @@ describe('ol/Feature.js', function () {
     });
   });
 
+  describe('featureType', function () {
+    it('is initially undefined', function () {
+      const feature = new Feature();
+
+      expect(feature.featureType).to.be(undefined);
+    });
+  });
+
   describe('#clone', function () {
     it('correctly clones features', function () {
       const feature = new Feature();
       feature.setProperties({'fookey': 'fooval'});
       feature.setId(1);
       feature.setGeometryName('geom');
+      feature.featureType = 'Foo';
       const geometry = new Point([1, 2]);
       feature.setGeometry(geometry);
       const style = new Style({});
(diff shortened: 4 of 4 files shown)


ahocevar (changes requested):
Thanks for this contribution! Instead of adding a new `featureType` member on `ol/Feature`, which is quite specific to GML, I'd suggest giving the GML format a `featureTypeProperty` option, and `feature.set(featureTypeProperty, featureType)` when set.

ejn:
> Thanks for this contribution! Instead of adding a new `featureType` member on `ol/Feature`, which is quite specific to GML, I'd suggest giving the GML format a `featureTypeProperty` option, and `feature.set(featureTypeProperty, featureType)` when set.

I touched on this in #12937, which I hoped would result in some discussion before I implemented anything. Since it didn't then I just went ahead anyway.

The problem I see with a feature type property is that it then becomes a "true" property of the feature (albeit a special one), which would then be output in all formats. In GML it would have some special handling and be filtered out / converted to element name, but otherwise it would just get output.

I note that the feature ID already has a special-case handling and is not set as a normal property.

If I go down the route of a `featureTypeProperty` for the GML format, then should this have a default value - which would be a noticable (arguably breaking) change for all GML parsing - or should this be optional?

ahocevar:
Thanks for the explanation, @ejn. I think if you change the private `featureType_` property in your pull request to `featureType` and remove the getters and setters, so the format sets and reads the property directly, I could live with this pull request. I don't think it is necessary to fire a change event when the featureType changes.

ejn:
Thanks for the feedback @ahocevar - will make the changes as you suggest. Agree that the change event is unnecessary - had thought about it in case the rendering should be updated, but changing the property after adding the feature to the map is not really a sensible use-case.

src/ol/Feature.js
```diff
@@ -97,6 +97,13 @@ class Feature extends BaseObject {
      */
     this.un;
 
+    /**
+     * Name of type of feature (e.g. GML Feature element name)
+     * @type {string|undefined}
+     * @api
```
ahocevar, line 103: Does this need to be part of the API? If I understand you correctly, this is only for not losing information between GML reader and writer.

ahocevar (commented):
Didn't you want to also have the GML writer pick up the `featureType`?

Thinking about this more: what is the use case for this really? Can you provide a test that shows where/how this is actually being used? Are you intending to use different feature types within a single GML document? If not, you can use the configurations available for the GML reader and writer to maintain the featureType.

ahocevar:
I'm still struggling to find this change useful, and still thinking about what alternatives we have. Two come to mind:
1. GeoServer always makes the feature type part of the feature id, e.g. `myFeatureType.1`. This could be achieved by passing the GML reader an `idProperty` option (similar to `ol/format/MVT`, which could also be a function. The GML writer could then also be changed so `featureType` accepts a function.
2. The GML reader could have a `readFeatureTypes` method, which would return the feature types found in the passed document. Then, with a new `featureType` reader option, reading features could be restricted to that one `featureType`. Then you'd call the reader once for each found featureType, with the respective `featureType` configured.

src/ol/Feature.js
ejn: > Does this need to be part of the API? If I understand you correctly, this is only for not losing information between GML reader and writer.

The information should also be available for use e.g. in generic interprtation of GetFeatureInfo-results (see below)

ejn:
> Didn't you want to also have the GML writer pick up the `featureType`?

Ultimately yes. Since the writing is slightly more complicated - and my immediate requirement is fulfilled by being able to read the feature type as part of a generic WMS client - then I would implement the writing as a second PR

> I'm still struggling to find this change useful, and still thinking about what alternatives we have. Two come to mind:
> 
>     1. GeoServer always makes the feature type part of the feature id, e.g. `myFeatureType.1`. This could be achieved by passing the GML reader an `idProperty` option (similar to `ol/format/MVT`, which could also be a function. The GML writer could then also be changed so `featureType` accepts a function.

This is very much implementation-specific and is not something which could be relied on. My "standard" example for the need to read multiple feature types is [this GetFeatureInfo-request](https://www.geodaten-mv.de/dienste/alkis_wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetFeatureInfo&FORMAT=image%2Fpng&TRANSPARENT=true&QUERY_LAYERS=adv_alkis_flurstuecke&LAYERS=adv_alkis_flurstuecke&INFO_FORMAT=text%2Fxml%3Bsubtype%3Dgml%2F3.2.1&I=50&J=50&CRS=EPSG%3A25833&STYLES=&WIDTH=101&HEIGHT=101&BBOX=293849.5404071358%2C6004185.0609577745%2C293918.4150123862%2C6004253.935563024) which requests only one layer but receives two (complex) features of different types (`AX_Flurstueck`, `AX_LagebezeichnungOhneHausnummer`) in response. For a generic client it is necessary to show the user not only the properties of the features found, but also what type of features they are (e.g. feature type name as a heading followed by a table containing the feature properties). In general there is also no way to discover in advance through the WMS interface exactly what a GML-formatted GetFeautreInfo-Response will contain. I agree that such responses are not very common (although similar services to my example are freely available in many German states), and similar examples of GetFeatureInfo-responses containing multiple feature types may also be found in INSPIRE services.

>     2. The GML reader could have a `readFeatureTypes` method, which would return the feature types found in the passed document. Then, with a new `featureType` reader option, reading features could be restricted to that one `featureType`. Then you'd call the reader once for each found featureType, with the respective `featureType` configured.

I agree that this would be a possible strategy for fulfilling the aim of reading multiple feature types - but does not open the possibility to write multiple feature types in a single collection. It would however fulfill my immediate need for presenting GetFeatureInfo-results (until it becomes necessary to also offer an export function!).

ahocevar:
@ejn Can you share the code you use to read the GML you linked above from the GetFeatureInfo request? I'd like to set up a test case.

ejn:
@ahocevar Sure, I'll try and cut it down to a reasonable size as it's part of a viewer application. Note that #12936 also resulted from parsing such data!

Although the above link is freely available, it is not strictly under an open data licence. [This link](https://www.wms.nrw.de/geobasis/wms_nw_alkis?SERVICE=WMS&REQUEST=GetFeatureInfo&FORMAT=image/png&TRANSPARENT=TRUE&STYLES=Farbe&VERSION=1.3.0&LAYERS=adv_alkis_flurstuecke&WIDTH=563&HEIGHT=614&CRS=EPSG:25832&BBOX=356465.3636101475,5645201.37707637,356633.46569215134,5645384.706878058&QUERY_LAYERS=adv_alkis_flurstuecke&INFO_FORMAT=text%2Fxml%3Bsubtype%3Dgml%2F3.2.1&I=282&J=307) produces identically structured data (with the bonus of an additional feature type `AX_LagebezeichnungOhneHausnummer`) from a different state (NRW) who publish the data under the [Data licence Germany - Zero](https://www.govdata.de/dl-de/zero-2-0), which would be ideal for including in a test-case.

ejn:
@ahocevar Thanks for your patience on this up until now - will try and and some more tests etc. It looks like I won't have time to work on this the next few days though, so might be a while until the next activity here.

Issue #11131, opened 2020-06-02
WebGLPointsLayer wrapX support


landonsilla:
**Is your feature request related to a problem? Please describe.**
`'ol/source/Vector'` supports `wrap: true` for lines, polygons and tiles.  So when you pan around the world left or right those features remain persistent in the correct location.  However,WebGLPointsLayer does not support that.  You pan "around the world" once and all the points do not appear.  You can plainly see this on any of the [examples](https://openlayers.org/en/latest/examples/webgl-points-layer.html) using WebGLPoints

**Describe the solution you'd like**
Being able to enable `wrap: true` on either the WebGLPointsLayer, or the underlying source would be awesome, and appreciated.  (Btw, I'm new here, this project is great).

jahow:
Marking this as PR accepted, this would be an awesome improvement.

IMO a good way to approach this would be to modify the webgl renderer to do multiple renders instead of just one [here](https://github.com/openlayers/openlayers/blob/f7b4f756e80a817111091735a9077e1097fbadfc/src/ol/renderer/webgl/PointsLayer.js#L409), shifting each render on the X axis based on the current rendered extent.

Hit detection will have to be modified to handle this as well.

KoAi:
Sure, it will be an awesome feature, I am waiting for it too

bochkov-vi:
this problem is urgent, it is not possible to work with the Pacific Ocean

oshawa-connection:
Hello! I'd be interested in attempting to fix this. 

Sorry if this is a stupid question, but I can't see a way to set an offset with this.helper.drawElements(0, renderCount). By the time its called, the vertices should be already set up in webGL?

PR #13528, opened 2022-04-01
WebGLPointsLayer wrapX support - partially addressing #11131

burleight:
WebGLPointsLayer renderFrame and renderHitDetection will now draw multiple worlds if the source and projection support wrapX - #11131

image

Hit detection needs additional improvement. It stops working after more than one wrap around the world. If 0 is the middle world, then the hit detection works for worlds -1, 0, and -1, but not for worlds > 2 or < -2.

The example has hit detection enabled, demonstrated with a colour change on hover for the circle styles. When moving the mouse, the hit detection is unreliable and flickers on/off. This needs improvement.

The webgl-points renderer test has been updated.


ahocevar (approved):
Thanks, @burleight

PR #13531, opened 2022-04-01
Add optional maxWidth for ScaleLine control

bartvde:
Add an optional maxWidth config option for the ScaleLine control, so that applications can limit the size of the ScaleLine

### DIFF
--- a/src/ol/control/ScaleLine.js
+++ b/src/ol/control/ScaleLine.js
@@ -51,6 +51,8 @@ const DEFAULT_DPI = 25.4 / 0.28;
  * @property {string} [className='ol-scale-line'] CSS Class name.
  * @property {number} [minWidth=64] Minimum width in pixels at the OGC default dpi. The width will be
  * adjusted to match the dpi used.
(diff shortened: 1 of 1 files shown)


src/ol/control/ScaleLine.js
```diff
@@ -51,6 +51,8 @@ const DEFAULT_DPI = 25.4 / 0.28;
  * @property {string} [className='ol-scale-line'] CSS Class name.
  * @property {number} [minWidth=64] Minimum width in pixels at the OGC default dpi. The width will be
  * adjusted to match the dpi used.
+ * @property {number|undefined} [maxWidth=undefined] Maximum width in pixels at the OGC default dpi. The width will be
```
ahocevar, line 54:
```suggestion
 * @property {number} [maxWidth] Maximum width in pixels at the OGC default dpi. The width will be
```
This is a shorter equivalent for an optional property, which implies `undefined`.

ahocevar (approved):
Looks good! If you're able to, it would be great if you could add a test.

bartvde:
@ahocevar I just added the tests

ahocevar (approved):
Awesome, please merge!

Issue #2700, opened 2014-09-09
Draw and modify interaction need to accept geometry layout


tschaub:
The draw and modify interactions create new point geometries.  The interactions need to be configured with a geometry layout in order to be used with anything but XY layouts.

ahocevar:
I think you're right, @drnextgis

drnextgis:
@ahocevar I've made a PR to fix that. Please review.

PR #13546, opened 2022-04-08
Add geometryLayout property to Draw interaction

drnextgis:
Fixes https://github.com/openlayers/openlayers/issues/2700

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -5,6 +5,7 @@ import Circle from '../geom/Circle.js';
 import Event from '../events/Event.js';
 import EventType from '../events/EventType.js';
 import Feature from '../Feature.js';
+import GeometryLayout from '../geom/GeometryLayout.js';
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Draw.js
```diff
@@ -393,6 +413,46 @@ class Draw extends PointerInteraction {
           } else {
             geometry = new Constructor(coordinates);
           }
+          const stride = getStrideForLayout(this.geometryLayout_);
```
ahocevar, line 416: I'm wondering if this block could be simplified. Maybe a better approach would be to work with the specified coordinate layout's stride throughout the draw process, i.e. starting with the `sketchCoord`?

drnextgis: Thanks for reviewing. I've updated the PR.

ahocevar (approved):
Very beautiful. Thanks, @drnextgis!

PR #13551, opened 2022-04-09
Do not add second interaction in Pinch Zoom example

mike-000:
Before version 6 the Pinch Zoom example replaced the default interaction with one specifying `constrainResolution`.  In version 6 the constraint is moved to the view, but the example has gained an unnecessary second default interaction https://github.com/openlayers/openlayers/pull/9137/files#diff-c42b5ce0fa11ec19b452214fbb1845c75c0934c8961225d186bb020b0a23c4ed


ahocevar (approved):
Thanks, @mike-000

PR #13552, opened 2022-04-09
Updates for ol-mapbox-style v8

ahocevar:
This pull request makes the necessary changes for the breaking changes in ol-mapbox-style v8, and updates the [vector-tiles-4326.js](https://github.com/openlayers/openlayers/compare/main...ahocevar:ol-mapbox-style-8?expand=1#diff-97d89635db79fdc975294642548a387103c0246be1b93eb029d54015361b379c) example to make use of the improved API.

`ol/layer/MapboxVector` can now be simplified. In part, this is achieved by moving the Mapbox url utilities to `ol-mapbox-style`. This also avoids some circular dependencies between `ol` and `ol-mapbox-style`. The tests related to these utilities are now also in `ol-mapbox-style`.

### DIFF
--- a/examples/vector-tiles-4326.js
+++ b/examples/vector-tiles-4326.js
@@ -1,55 +1,36 @@
-import MVT from '../src/ol/format/MVT.js';
-import TileGrid from '../src/ol/tilegrid/TileGrid.js';
+import VectorTileLayer from '../src/ol/layer/VectorTile.js';
 import VectorTileSource from '../src/ol/source/VectorTile.js';
-import View from '../src/ol/View.js';
-
-import olms from 'ol-mapbox-style';
-import {defaultResolutions} from 'ol-mapbox-style/dist/util.js';
+import {Map, View} from '../src/ol/index.js';
+import {applyBackground, applyStyle} from 'ol-mapbox-style';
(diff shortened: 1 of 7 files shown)


mike-000:
In the example the tilegrid could be simplified to

```
    createXYZ({
      extent: [-180, -90, 180, 90],
      tileSize: 512,
      maxResolution: 180 / 512,
      maxZoom: 13
    }),
```

examples/vector-tiles-4326.js
```diff
@@ -1,55 +1,36 @@
-import MVT from '../src/ol/format/MVT.js';
 import TileGrid from '../src/ol/tilegrid/TileGrid.js';
-import VectorTileSource from '../src/ol/source/VectorTile.js';
-import View from '../src/ol/View.js';
+import {Map, View} from '../src/ol/index.js';
```
tschaub, line 3: Curious if there is a reason for this blank line (like forcing a certain import order).  If so, a comment would clarify.

ahocevar: This was just an oversight on my end, I removed the line break while including @mike-000's suggested simplification.

tschaub (approved):

Issue #7475, opened 2017-11-15
Add boolean ignorePlacement option to ol.style.Image and ol.style.Text


tomchadwin:
- [x] I am submitting a bug or feature request, not a usage question. Go to https://stackoverflow.com/questions/tagged/openlayers for questions.
- [x] I have searched GitHub to see if a similar bug or feature request has already been reported.
- [x] I have verified that the issue is present in the latest version of OpenLayers (see 'LATEST' on https://openlayers.org/).
- [ ] If reporting a bug, I have created a [CodePen](https://codepen.io) or prepared a stack trace (using the latest version and unminified code, so e.g. `ol-debug.js`, not `ol.js`) that shows the issue.

The new `declutter` option in 4.5.0 is a huge development. However, I would ask for one enhancement to be considered. If a layer's style function returns both imagery and text (eg point markers and labels), it would be great to have the option to declutter only text. To me, that seems like a more common requirement. I appreciate that this can be worked around, but changing `declutter` from a boolean to something like `none`, `text`, `geometry`, and `both` (or `all`) would work well.

ahocevar:
Wouldn't it add more flexibility to add boolean `ignorePlacement` options to `ol.style.Image` and `ol.style.Text`? That's equivalent to [icon-ignore-placement](https://www.mapbox.com/mapbox-gl-js/style-spec/#layout-symbol-icon-ignore-placement) and [text-ignore-placement](https://www.mapbox.com/mapbox-gl-js/style-spec/#layout-symbol-text-ignore-placement) in the Mapbox Style specification.

tomchadwin:
If `ignorePlacement` includes the ability to be `declutter`ed, yes, this would work more granularly. Thanks!

congying2:
yes ,draw lessons from mapbox ，Good things

bartvde:
Talked to @ahocevar about this and we agree that probably allowOverlap is a better term here, we did find https://github.com/mapbox/mapbox-gl-js/pull/486 which gives an insight into the two different options, but we think a single option will suffice for now and allowOverlap is the better term for disabling the decluttering.

tomchadwin:
@bartvde Would this replace `declutter`, or be used in conjunction with it?

bartvde:
Good question @tomchadwin I'm not 100% sure, I think the option we are talking about is at the style level, and declutter is currently at the layer level. So I'm guessing this is in conjunction. @ahocevar can you confirm?

[...]

PR #13566, opened 2022-04-12
Decluttering mode by style

cns-solutions-admin:
As requested e.g. in #7475, this improves decluttering by being able to state per (image) style, how this style should be decluttered:
* declutter => declutter as before
* obstacle => draw image/text, but use it as obstacle for styles with mode "declutter"
* none => do not declutter

Decluttering still only happens, if the layer also has declutter = true.


cns-solutions-admin:
I named the property `declutterMode`, because 
* it strongly hints that it is only used, if the layer's `declutter = true`
* it is easier to understand than mapbox's allowOverlap and ignorePlacement
* we can easily add other modes later on

I have also implemented support for decluttering in ol-mapbox-style. PR will follow, when this is accepted.

cns-solutions-admin:
Build errors (doc type links) fixed.

ahocevar (changes requested):
Thanks for your work on this, @cns-solutions-admin! 

Instead of introducing a `DeclutterMode` text enum, you could simply create a typedef with a union type. Something like:
```js
/** @typedef {"none" | "obstacle" | "declutter"} DeclutterMode */
```
Eventually, we'll be changing all string enums in the library to union types.

To make sure that this new feature can be maintained, it would be important to add tests. In this case, two rendering tests (one for each new mode) would be appropriate.

Of course it would be nice to have this new option also for Text styles, but that does not need to be part of this pull request.

Finally, would you be able to squash the commits in a way that the pull request does not contain any merge commits? A single commit for the whole pull request would also work.

cns-solutions-admin:
I succeeded in squashing everything into one commit.
I added a rendering test for both modes based on the other decluttering test.
I left the DeclutterMode enum to be consistent with the other enums - unless everyone uses typescript, this way is safer.

ahocevar:
Thanks for your continued effort on this, @cns-solutions-admin.

> I left the DeclutterMode enum to be consistent with the other enums - unless everyone uses typescript, this way is safer.

We'll be removing these enums, see #12696. We do have the tooling in place (`npm run typecheck`)  to ensure correct values with union types, and the generated d.ts files do not work with the examples. For example, the `getDeclutterMode()` API method you introduced, will be broken from the start if you stick with the enum.

ahocevar:
The rendering test does not test the difference between `none` and `obstacle` mode. If there were e.g. other features rendered on top of the red and blue circles, it could be shown that with `none` mode (on top of blue) the other features would be drawn, but not with `obstacle` mode (on top of red).

cns-solutions-admin:
Based on your comments "you could..." and "eventually, ...", I assumed that it was better to use the enum for now. 
It's changed now and the test is also improved.

ahocevar (commented):
In https://github.com/ahocevar/openlayers/commit/bbd58eae6b3dee6e22b3261030a2899e120feb1f, I tried to make the test more readable. And by doing so, I saw something I have no explanation for:
actual
Why are the red circles on top of the `c-orange` label?

test/rendering/cases/layer-vector-decluttering-extended/main.js
```diff
@@ -0,0 +1,159 @@
...
+    },
+  })
+);
+
+const source2 = new VectorSource();
+source2.addFeatures([
+  new Feature({
+    geometry: new Point([center[0], center[1]]),
+    text: 'center',
+    zIndex: 2,
```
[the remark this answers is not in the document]
cns-solutions-admin: You introduced a typo: "zindex" instead of "zIndex". Thus the red circles were not used for the orange labels, as those were on top of  the red layer.

But this highlighted that I somehow did not commit the continue in the executor: the obstacles are used in the declutter executor to be added to the declutter tree, but rendered in the normal executor - so that they are behind the labels (in the test they were also rendered in the declutter executor thus being drawn on top of the label).

Fixed the executor and updated the test with your version + fix + additional blue circle/text to test
* that objects (red) cannot be obstacles for layers with a higher z index (blue)
* that the obstacles (red) are drawn behind layers with a higher z index (blue)

test/rendering/cases/layer-vector-decluttering-extended/main.js
```diff
@@ -0,0 +1,188 @@
...
+    geometry: new Point([center[0] - 540, center[1]]),
+    text: 'w-orange',
+  }),
+  new Feature({
+    geometry: new Point([center[0] + 540, center[1]]),
+    text: 'e-orange',
+  }),
+]);
+// circles are always drawn,
+// texts are decluttered against each other and the layer 1 circles/texts
```
ahocevar, line 117: For readability, please refer to the color instead of the layer number everywhere.

test/rendering/cases/layer-vector-decluttering-extended/main.js
```diff
@@ -0,0 +1,188 @@
...
+        }),
+      });
+    },
+  })
+);
+
+const sourceRed = new VectorSource();
+sourceRed.addFeatures([
+  new Feature({
+    geometry: new Point([center[0] + 1000, center[1] + 1000 - 200]),
```
ahocevar, line 61: The previous version with the overlapping red circles was better, to show that they do serve as obstacle, but are rendered even if they overlap.

ahocevar (commented):
Yes, the new behavior is what I would expect. I added two more comments regarding the tests. When these are addressed, I think this is ready for prime time.

test/rendering/cases/layer-vector-decluttering-extended/main.js
cns-solutions-admin: ok, reverted zoom, moved orange circles' text to left with xOffset to show at least one label.

test/rendering/cases/layer-vector-decluttering-extended/main.js
```diff
@@ -0,0 +1,189 @@
...
+
+const sourceBlue = new VectorSource();
+sourceBlue.addFeatures([
+  new Feature({
+    geometry: new Point([center[0] + 1000 + 540, center[1] + 900 - 600]),
+    text: 'top-blue',
+  }),
+]);
+// circles are always drawn, but serve as obstacles,
+// texts are decluttered against each other and the circles
```
ahocevar, line 31: I think this comment is not applicable. `declutterMode` `declutter` here, so if you had more blue circles overlapping this one, they wouldn't be rendered.

test/rendering/cases/layer-vector-decluttering-extended/main.js
cns-solutions-admin: comments improved.

test/rendering/cases/layer-vector-decluttering-extended/main.js
cns-solutions-admin: removed when improving comments.

ahocevar (approved):
Thanks for your effort on this, @cns-solutions-admin

ahocevar:
Now we have a rendering test mismatch. The actual result does not show a label for the cyan layer:
actual

cns-solutions-admin:
This is weird. Locally the test works.

ahocevar:
I doubt that a difference in font rendering would cause this label to disappear, but I'll try in interactive mode to see what happens when zooming in/out a bit.

ahocevar:
This is really weird. The results are quite stable when zooming in/out locally.

ahocevar:
@cns-solutions-admin Maybe you could cherry-pick 134462cb8d, so we can see if the tests pass then. Or you could grant maintainers access to this pull request's branch, then I could push that change here directly.

cns-solutions-admin:
Done. Seems that the font on my computer is a bit smaller/slimmer.

ahocevar:
Now it looks good. You have to update the reference image in your branch with the one from the CI, or give me access to this pull request's branch, or cherry-pick 9d17d92fc2.

cns-solutions-admin:
CI image downloaded, committed and pushed.
Seems that giving access to branches is not possible for organization accounts.

ahocevar (approved):
Thanks for your continued effort on this, @cns-solutions-admin

marcjansen:
Very nice collaboration here.

Issue #13481, opened 2022-03-14
Decouple Label Position from textAlign


rycgar:
# Decouple Label Position from `textAlign`

The [ol/style/Text](https://openlayers.org/en/latest/apidoc/module-ol_style_Text-Text.html) offers the `textAlign` option,
however, `textAlign` also affects the label position. You can see it in this image:

![textAlign-affects-position](https://user-images.githubusercontent.com/51781722/158227779-43265774-c425-42e8-bc1d-beb6eb0c30ff.png)

## Proposal for Point Placement

Separate `textAlign` into two option.
1. `textAlign`: only text alignment. The values stay the same: {`'left'`, `'center'`, `'right'`} (_case: point placement_)
2. `position`: introduce new option with extended values:

### Position Values
- center
- left
- left-top
- top
- right-top
- right
- right-bottom
- bottom
- left-bottom

### Few Position Examples

Few position examples, whereat, text align is set to default value `'center'`.

![label-position-examples](https://user-images.githubusercontent.com/51781722/158227562-01b5be2a-2a2d-4fba-85e2-b2596ab4a5ff.png)

## Contribution

If the proposed API is good enough, I would like to implement the solution and create a pull request. Improvements and/or concerns are welcome too.

ahocevar:
Thanks for the suggestion! i agree the current API lacks positioning options, but this would be a breaking change. I think we could also handle this with a non-breaking change for now. At some point, we have more work to do to consolidate the style API, and when that time has come we should make all breaking changes altogether.

I think a viable solution would be to add a `justify` option. When not set, everything is as it is now, the meaning would be "justify text towards the `textAlign` anchor". When set to `left`, `center` or `right`, text justification within the box aligned by `textAlign` will change. This would also avoid confusion with the (poorly chosen) `textBaseline` option. So the examples you posted in the above image would then be:

* center: `textAlign: "center", textBaseline: "middle"`
* right: `textAlign: "left", textBaseline: "middle", justify: "center"`
* left-top: `textAlign: "right", textBaseline: "bottom", justify: "center"`
* bottom: `textAlign: "center", textBaseline: "top"`

Would this be something you could live with?

rycgar:
For now, I agree on introducing the new option `justify`, because `position` makes less sense in case of `placement: line` and so it requires more discussion.

I will start implementing `justify`, if it is fine for you.
If there are some important things to consider while doing that, please comment.

[...]

PR #13571, opened 2022-04-13
Add `justify` option for text style

rycgar:
See discussion about these changes here: #13481 

Note:
The `justify` options is ignored in case (1) `placement: 'line'` or for (2) immediate rendering

1. Although multi-line text is available for line placement, it does not make any sense to use `justify` here, so it is ignored
2. As noted under the `text` option in the API-doc for `ol/style/Text`: _rich text is not supported for the immediate rendering API_. Multi-line text is not available here, and so `justify` is ignored here as well.


rycgar:
I cannot reproduce failing rendering test for the new test case **rich-text-justify-style**
```
$ node test/rendering/test.js --match rich-text-justify-style
<i> [webpack-dev-middleware] wait until bundle finished: /cases/rich-text-justify-style/main.js
135 assets
1008 modules
webpack 5.72.0 compiled successfully in 9943 ms
```

ahocevar:
Thanks for this contribution, @rycgar! I updated the reference image for the failing rendering test with that from the CI - font rendering is very different depending on the platform and used font rendering libraries, so all rendering tests with text are very fragile. Now the test will probably fail for you locally, but don't worry about that.

Reviewing now.

src/ol/render/canvas/Immediate.js
```diff
@@ -946,6 +946,8 @@ class CanvasImmediateRenderer extends VectorContext {
     const textAlign = textState.textAlign
       ? textState.textAlign
       : defaultTextAlign;
+    // Ignore `textState.justify` here,
+    // since rich-text is not supported for the immediate rendering API.
```
ahocevar, line 950: Maybe replace with
```suggestion
    //FIXME Implement `justify`
```
or remove the comment.

ahocevar (changes requested):
Very elegant solution, thanks @rycgar. Please address my inline comments, but code-wise this is in great shape.

src/ol/render/canvas/Immediate.js
rycgar: Thx. I'll remove the comment, because, as far as I know, having multi-line text for immediate rendering is not possible yet.

rycgar:
I adjusted the comment, they should be okay now.

Thanks for your help with the rendering test and for the fast actions. :+1:

ahocevar:
I took the liberty to simplify and beautify the example a bit.

ahocevar:
I took the liberty to simplify and beautify the example a bit.

ahocevar (approved):
Thanks, @rycgar

PR #13572, opened 2022-04-13
Add Geodesic Measurement

theduckylittle:
Adds a drop-in alternative to ol/sphere that
utilizes geographiclib for geodesic based measurement.

The sphere approximation is fast to calculate but overestimates
the area near the browser.


ahocevar:
I'm hesitant to add another dependency to the library. The work in this pull request looks solid, and I think it would be a perfect candidate for a [3rd party library](https://openlayers.org/3rd-party/). If you agree to make this geodesy module a 3rd party library, it might still be useful to add an example to this repository that shows how to use it.

krishnaglodha:
https://github.com/Rotten-Grapes-Pvt-Ltd/ol-measurement-geographiclib

PR #13575, opened 2022-04-15
Add premultiplied option to WebGLTileLayer getData() method

### LINKED ISSUE #13570: `layer.getData()` RGB values corrupted by alpha-premultiplication
**Describe the bug**
`layer.getData()` may not return the correct RGB values for tile sources due to alpha-premultiplication in the 2d canvas used to obtain the pixel value..

**To Reproduce**
This can be seen in the Tilezen normal format tiles https://github.com/tilezen/joerd/blob/master/docs/formats.md#normal where at high elevations the alpha value is low (or even zero).

See this example https://codesandbox.io/s/simple-forked-jjly5g?file=/main.js  The `getData()` result is shown together with the true value obtained by styling a layer with full alpha and then reading the rendered result, either with `forEachLayerAtPixel()` or by wrapping the layer in an `ol/source/Raster` and calling `getData()` on that.

**Expected behavior**
This would not be a problem when using canvas layers as OpenLayers would never have access to the original values, but with WebGL layers that data is available and can be used, so it should be reflected by `getData()`.

That can be achieved by reading directly from the textures https://codesandbox.io/s/simple-forked-z1p5q4?file=/main.js although I am uncertain of what (potentially browser or device dependent) overheads that might have..  The differences in the example are due to `getData()` needing to floor the offsets while `forEachLayerAtPixel()` uses the closest rendered pixel - when the source is overzoomed the results match.

mike-000:
Fixes #13570

Add `premultiplied` optional parameter which defaults to true, which is the current efficient but, for some image tiles which contain data, potentially inaccurate behavior.  Add test to get such data not premultiplied.


src/ol/layer/BaseTile.js
```diff
@@ -150,11 +150,13 @@ class BaseTileLayer extends Layer {
    * });
    * ```
    * @param {import("../pixel").Pixel} pixel Pixel.
+   * @param {boolean} [opt_premultiplied=true] If `false` return the RGBA data for image tiles before
+   * alpha premultiplication (when used frequently this may incur a significant performance loss).
```
tschaub, line 154: If we are going to support this, we should accept an `options` object.  The non-api methods don’t need to change.

PR #13594, opened 2022-04-23
Do not reload data tiles if already loaded or loading

mike-000:
The API states "Load not yet loaded URI." but the `load()` method for data tiles does reload already loaded or loading tiles unless the application uses the non-API `getState()` method to prevent it.  Code which works for Image tiles results in continuously loading data tiles hanging the application.

### DIFF
--- a/src/ol/DataTile.js
+++ b/src/ol/DataTile.js
@@ -75,21 +75,26 @@ class DataTile extends Tile {
    * @api
    */
   load() {
-    this.state = TileState.LOADING;
-    this.changed();
+    if (this.state == TileState.ERROR) {
+      this.state = TileState.IDLE;
+    }
+    if (this.state == TileState.IDLE) {
+      this.state = TileState.LOADING;
+      this.changed();
 
-    const self = this;
-    this.loader_()
-      .then(function (data) {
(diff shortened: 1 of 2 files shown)


src/ol/DataTile.js
```diff
@@ -75,21 +75,26 @@ class DataTile extends Tile {
...
    */
   load() {
-    this.state = TileState.LOADING;
-    this.changed();
+    if (this.state == TileState.ERROR) {
+      this.state = TileState.IDLE;
+    }
+    if (this.state == TileState.IDLE) {
+      this.state = TileState.LOADING;
+      this.changed();
```
tschaub, line 83: This seems a bit mysterious.

If all you want to do is prevent loading if not idle, I would add these lines to the start of the function:
```js
if (this.state !== TileState.IDLE) {
    return;
}
```

tschaub: Never mind, I see the behavior is similar for image tiles.

tschaub:
I simplified the change and added a test in 5084d5566a427b177f3828e01b7c475e734e6e6c.

tschaub (approved):

Issue #13061, opened 2021-11-30
Images of WebGL Tile Layer get resized


hackermd:
**Describe the bug**
With the current HEAD of the `main` branch (commit a788c9bd1ebc93761660b3c5c290b9fbd6c8d42f), images of WebGL tile layers at individual resolution levels are resized (stretched along the row or column direction) for some tile grids. That leads to distortions of the raster graphics and incorrect display of corresponding vector graphics.

I encountered this problem with the following tile grid:

![Screen Shot 2021-11-30 at 9 44 24 AM](https://user-images.githubusercontent.com/6855297/144069291-2bac02f2-a2c8-4ffa-99d0-db26d376c170.png)

This is the image at the top of the pyramid (zoom level 0):

![Screen Shot 2021-11-30 at 9 48 04 AM](https://user-images.githubusercontent.com/6855297/144069220-81bb1178-5081-4a9f-8ffe-741e0ead1ffb.png)

This is the image one level below the top of the pyramid (zoom level 1):

![Screen Shot 2021-11-30 at 9 46 39 AM](https://user-images.githubusercontent.com/6855297/144069219-ba865e21-b971-4982-98a2-c369b85e9f82.png)

One can clearly see that pixels are non-square in the first image and that the vector graphic (orange line) is displayed at the wrong spatial location.

The images display correctly with an earlier version of the app that uses Canvas tile layers of release v6.9.0.

mike-000:
The tiles have a standard multiple of 8 width so there should be no problems related to #12995 as long as the supplied height (which is used to calculate the data bytes per row from the total data bytes) is correct.

hackermd:
It appears `frameState.size` is constant across resolutions levels and is not adjusted when frames at a given level get rendered by the WebGL renderer via `renderFrame()`:
https://github.com/openlayers/openlayers/blob/c57c8983814451813002fe29f479cae6ed4fc11a/src/ol/renderer/webgl/TileLayer.js#L404-L405

tschaub:
@hackermd - can you provide an example that reproduces this issue (ideally, a link to a CodeSandbox like https://codesandbox.io/s/data-tiles-flv3p?file=/main.js)?

Punzo:
Hi @tschaub recently wee encountered again this issuue and we managed to reproduuce the issue in a simple example: 

https://codesandbox.io/s/openlayer-issue-13061-f9guyd?file=/main.js

[...]

PR #13603, opened 2022-04-25
Handle rotation with non-square tiles

tschaub:
The WebGL tile renderer doesn't properly handle rotation with non-square tiles.

Here is what things look like currently if you try rotating a map with non-square tiles:
bad

Instead of being able to compose the transform in one step, we need to apply the scaling due to the non-square aspect ratio after the rotation.

Here is what things look like after this change:
good

This was discovered as part of trying to add support for non-square pixels (see #13593).


ahocevar (approved):

hackermd:
@tschaub this PR also fixes #13061

PR #13609, opened 2022-04-27
Adjust canvas getData() offsets and rendered gutter for served image size

### LINKED ISSUE #13605: layer.getData() returns incorrect result if image is scaled to fit tile size
**Describe the bug**
The image tile equivalent of the non-square pixels in #13593 is generally not a problem as with both canvas and WebGL renderers images are scaled to fit the tile size (the exception being Zoomify/IIIF where a custom tile class crops or pads to fit).  However the `layer.getData()` method does not consider this.

**To Reproduce**
See https://codesandbox.io/s/data-tiles-forked-0ft1es?file=/main.js  512 x 256 images are stretched to fit the 512 x 512 tile grid.  No data is detected in the lower half of the tiles, and incorrectly detected in the top half.

**Expected behavior**
`layer.getData()` should take the actual image size into consideration when calculating pixel offsets.

mike-000:
Fixes #13605 for canvas rendering

Calculate `layer.getData()` offsets based on the actual image size in case image is scaled to fit specified tile size.

Also fix the rendering so gutter is adjusted for image size.  Previously canvas rendering only handled gutter for TileWMS where the image was requested to fit the tile grid.  Canvas rendering is now consistent with WebGL rendering applying gutter to scaled textures.

To avoid conflicts a separate commit in #13575 applies the change to WebGL.


ahocevar:
Wouldn't it simplify things in a lot of places if we handled the corner case of a `gutter` (which I think no-one really uses, except some very rare cases with WMS tiles) the same way as we handle cropped tiles in the Zoomify source, i.e. by rendering the image we get from the source to a canvas that is a correctly sized tile?

mike-000:
No, gutter is now also available for XYZ (currently we have no examples).  When rendering with WebGL without a gutter there are always tile edge effects when interpolating, sometimes quite obvious as below.  Mapbox and Nextzen have guttered tiles which avoid that, but to work the gutter must be clipped during rendering

image

image

Issue #13606, opened 2022-04-26
GeoTIFF not working with legacy build


mike-000:
**Describe the bug**
GeoTIFFs are not working with legacy build since v6.13.0.

**To Reproduce**
Open simple example in https://jsbin.com/xomuhus/edit?html,js,output It works with version 6.12. Change `v6.12.0/build/ol.js` to a later release and observe the error in the console.

**Expected behavior**
Legacy build on latest browsers (I was using Chrome 100 and 101 and Edge 100) should support API methods.

ahocevar:
Ok, a chunk loading problem. We've had that before. Need to check what's going on here.

ahocevar:
It turns out this is not a problem with the legacy build, but with jsdelivr.net. The requests for the dynamic import chunks (e.g. https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.14.1/build/231.ol.js) return a 403 with

> Package size exceeded the configured limit of 50 MB. Try https://github.com/openlayers/openlayers.github.io/tree/master/en/v6.14.1/build/231.ol.js instead.

If you copy the legacy build to your local server, it should work fine.

mike-000:
Same problem.  I do have `ol.js` but not `231.ol.js`  Why would the application expect worker script to be created on the server?
![image](https://user-images.githubusercontent.com/49240900/165482383-c15dd41e-08ad-4b97-b597-15af9bfb0f1d.png)

ahocevar:
Oh, I see our scripts that create ol-dist.zip and ol.zip for the releases do not include these extra chunks indeed. I created https://github.com/openlayers/openlayers.github.io/issues/124 for this.

ahocevar:
Release artifacts on https://github.com/openlayers/openlayers/releases/tag/v6.14.1 are updated and now include the dynamic chunks.

ahocevar:
The other reported problem here is https://github.com/openlayers/openlayers.github.io/issues/113.

ahocevar:
Redeployed v6.14.1, works now from cdn.

PR #13469, opened 2022-03-12
Fix legacy build

ahocevar:
Fixes #13468

Issue #13468, opened 2022-03-12
GeoTIFF not working with v6.13.0 legacy build


mike-000:
**Describe the bug**

**To Reproduce**
Create an html file by copying the quick start example, replacing the JavaScript with

```
const source = new ol.source.GeoTIFF({
  sources: [
    {
      url: 'https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/2020/S2A_36QWD_20200701_0_L2A/TCI.tif',
    },
  ],
});

const map = new ol.Map({
  target: 'map',
  layers: [
    new ol.layer.WebGLTile({
      source: source,
    }),
  ],
  view: source.getView(),
});
```

Change the version in the cdn path to 6.12.0 and open with a browser - it works as expected.
Repeat with version 6.13.0 and observe the error in the console.

**Expected behavior**
Legacy build should continue to support documented API methods.

PR #13610, opened 2022-04-27
Fix dynamic import paths for legacy build

ahocevar:
I know I went a bit back and forth on this, but the correct configuration was introduced with #12870, and #13469 broke it again. The real fix for  #13468, which #13606 is a duplicate of, is https://github.com/openlayers/openlayers.github.io/pull/125.


ahocevar:
Closing. This does not work in jsbin either. I think the problem is that the scripts there are loaded through a wrapper, so `document.currentScript` does not work.

PR #13613, opened 2022-04-28
Add information whether vertex was snapped to edge

manisandro:
Rationale: I'd like to add visual feedback when snapping occurs, I do this by extending the snap interaction and overriding `handleEvent`, and then style an overlay layer feature if `snapTo` returns a snap result. To be able to style the overlay feature differently based on whether snapping occured on a point or a segment, I'd like the snapping result to return this information.

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -28,6 +28,7 @@ import {listen, unlistenByKey} from '../events.js';
  * @typedef {Object} Result
  * @property {import("../coordinate.js").Coordinate|null} vertex Vertex.
  * @property {import("../pixel.js").Pixel|null} vertexPixel VertexPixel.
+ * @property {boolean} snappedToEdge Whether vertex was snapped to an edge.
  */
 
 /**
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
Nice work, and with a little effort it could be made even nicer: What about calling the new property `segment`, and return `null` when snapped to vertex and `segmentData.segment` when snapped to segment?

And even more bonus for adding a unit test.

manisandro:
Would it be an idea to add snapTo result to the returned event? Like

    handleEvent(evt) {
        const result = this.snapTo(evt.pixel, evt.coordinate, evt.map);
        if (result) {
          evt.coordinate = result.vertex.slice(0, 2);
          evt.snapResult = result;
        }
        return super.handleEvent(evt);
      }

ahocevar:
@manisandro I'm not sure. If you get the snapped vertex (if any) and the snapped segment (if any), I think you get more useful information for styling and other purposes.

manisandro:
My idea was to be able to subclass ol.interaction.Snap and override handleEvent without having to copy the actual implementation, say:

     class SnapInteraction extends ol.interaction.Snap {

        handleEvent(evt) {
            const evt = super.handleEvent(evt);
            if(evt.snapResult) {
                /* ... */
            }
            return evt;
        }
    }

ahocevar:
I may be missing something , but the `snapResult` only contains the vertex and the vertex's pixel, and you have that information in the event already. And for subclassing, this does not make a difference.

manisandro:
I meant the "new" extended result object which also includes the segment information. Or just adding the segment information directly to the event also works clearly.

ahocevar:
@manisandro If you're still interested in finishing this: let me know if you understood my suggestion and are fine with it. If you recreate this pull request, I'd be happy to review again.

manisandro:
Thanks, I've reopened it in https://github.com/openlayers/openlayers/pull/15306

PR #13625, opened 2022-04-30
Add getMetadata() method

### LINKED ISSUE #13612: Get GeoTIFF GDALMetadata values
I need original GeoTIFF data values as I encoded population values into them. I can't figure out how to access internal Type Definitions for STATISTICS_MINIMUM and STATISTICS_MAXIMUM values to use them for manual normalization.

STATISTICS_MINIMUM and STATISTICS_MAXIMUM are used in GeoTIFF module in case source min and max values are not set. So it would be enough to overwrite source min and max values to get access to them.

mike-000:
Fixes #13612

It seems reasonable to make metadata accessible when it is available.
As the structure is used by the loader returning a clone is safer.

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -440,6 +440,18 @@ class GeoTIFFSource extends DataTile {
   }
 
   /**
+   * Returns metadata from the sources, which might be helpful for manual normalization
+   * or other purposes.
+   * @return {Array<Array<GDALMetadata>>|undefined} Metadata
+   * @api
+   */
+  getMetadata() {
+    return this.metadata_
+      ? JSON.parse(JSON.stringify(this.metadata_))
+      : undefined;
+  }
+
+  /**
    * Configure the tile grid based on images within the source GeoTIFFs.  Each GeoTIFF
    * must have the same internal tiled structure.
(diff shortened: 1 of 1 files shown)


ahocevar:
This exposes 3rd party library structures to the OpenLayers API. Instead, I'd recommend using [geotiff.js](https://npmjs.com/package/geotiff)'s API to get the metadata, which is straightforward:
```js
import {fromUrl as tiffFromUrl} from 'geotiff';

tiffFromUrl(geotiffUrl).then(geotiff => geotiff.getImage()).then(image => {
  console.log(image.getGDALMetadata(0))
});
```

m-mohr:
@ahocevar But doesn't that do a lot of work twice (reading the file [hopefully from cache], parsing, etc.)?

ahocevar:
I'd be interested to hear what others think. The overhead of my suggested approach would be minimal, because (depending on when you read the url) the relevant metadata range will be cached already, and there is no heavy parsing work involved.

m-mohr:
I've added my actual use-case here: https://github.com/openlayers/openlayers/issues/13643#issuecomment-1118402880 
Thinking more about it I don't think getMetadata() as proposed here really solves it for me yet.

PR #13626, opened 2022-05-01
Wait for icons to be loaded before firing rendercomplete event

MoonE:
The rendercomplete event should wait for icons to load before it is being fired.
As there is already similar behavior for the webgl renderers with the `ready` property, this uses the same property.

This also fixes an error when an icon is loaded after the renderer has been disposed.

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -87,12 +87,6 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
 
     /**
      * @private
-     * @type {boolean}
-     */
-    this.dirty_ = false;
-
-    /**
-     * @private
      * @type {number}
      */
     this.renderedLayerRevision_;
@@ -289,7 +283,6 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
             builderGroup,
             declutterBuilderGroup
           );
-          this.dirty_ = this.dirty_ || dirty;
           builderState.dirty = builderState.dirty || dirty;
         }
       };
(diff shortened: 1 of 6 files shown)


ahocevar (approved):
Thanks, @MoonE!

MoonE:
Yes, I believe the `VectorTileLayer`'s `ready` is currently set to the last tile's state, I think I fixed it in my local branch, but I'd like to write a test to confirm it.

ahocevar:
Excellent, thanks!

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -255,7 +250,6 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
       const bufferedExtent = equals(sourceTileExtent, sharedExtent)
         ? null
         : builderExtent;
-      builderState.dirty = false;
```
MoonE, line 258: Looks like this did only take the last source tile's dirty state into account.

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -289,7 +283,6 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
             builderGroup,
             declutterBuilderGroup
           );
-          this.dirty_ = this.dirty_ || dirty;
```
MoonE, line 292: `this.dirty_` was never reset to `false` , though it was also unused.

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -673,10 +666,12 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
     const tiles = this.renderedTiles;
     const clips = [];
     const clipZs = [];
+    let ready = true;
     for (let i = tiles.length - 1; i >= 0; --i) {
       const tile = /** @type {import("../../VectorRenderTile.js").default} */ (
         tiles[i]
       );
+      ready = ready && !tile.getReplayState(layer).dirty;
```
MoonE, line 674: `ready` is true when the currently drawn tiles have no loading icons. Tile state is tracked separately.

ahocevar (approved):
Thank for the explanations and the thorough work on this!

PR #13627, opened 2022-05-01
Fix hitdetection for icon with offset and pixelratio != 1

### LINKED ISSUE #13608: Icon hit detection on retina screens
Describe the bug
I discovered a bug where hit detection does not work under a specific set of circumstances:

I provide a horizontal offset and size to my icon style
My monitor has a pixel ratio > 1 (e.g. a retina display on a macbook)
I give a custom color to my icon style
This seems related to an issue with styling I had before under similar circumstances: https://github.com/openlayers/openlayers/issues/12682

To Reproduce
I made a reproduction here:
https://codepen.io/sebakerckhof/pen/zYzrzNO

Expected behavior

Clicking on the feature should select it. This works on my linux machine (even if I increase display scaling), but not on my macbook pro.

MoonE:
Fixes #13608 
Depends on #13626 for the test to work.

The originX/originY value was always multiplied with the pixel ratio, that's wrong as hit detection is always done with a pixel ratio of 1.

### DIFF
--- a/src/ol/render/canvas/ImageBuilder.js
+++ b/src/ol/render/canvas/ImageBuilder.js
@@ -129,8 +129,8 @@ class CanvasImageBuilder extends CanvasBuilder {
       this.anchorY_ * this.imagePixelRatio_,
       Math.ceil(this.height_ * this.imagePixelRatio_),
       this.opacity_,
-      this.originX_,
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Thanks, @MoonE

PR #13636, opened 2022-05-02
Support non-uniform tile pixel ratios

tschaub:
This is an alternative to #13593 for supporting non-uniform pixel resolutions.

In #13593, the tile grid is updated to know about non-uniform resolutions.  In this branch, the source and tile textures are updated to support non-uniform pixel ratios, modifying the tile sizes returned by the tile grid.

### DIFF
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -169,13 +169,13 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
     const viewState = frameState.viewState;
     const source = layer.getRenderSource();
     const tileGrid = source.getTileGridForProjection(viewState.projection);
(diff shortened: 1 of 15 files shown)


mike-000:
I don't think is it necessary to allow `tilePixelRatio` arrays in anything other than DataTile/GeoTIFF sources.  The only use in image tiles is to request larger images for WMS and ArcGISRest and to render to larger image canvas.  Single numbers could be converted internally to arrays using `toSize()` where arrays are expected.

mike-000:
This also introduces conflicts with #13609 which fixes incorrect dependence of pixel ratios in favour of using the real image size.

tschaub:
I opened this to demonstrate (to myself at least) that the `tilePixelRatio` approach was not a good one.  And I'm convinced.

PR #13641, opened 2022-05-03
Fix typos in upgrade notes

marcjansen:
Title says it all, please review.

Introduced in https://github.com/openlayers/openlayers/pull/13637

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,7 +2,7 @@
 
 ### v6.15.0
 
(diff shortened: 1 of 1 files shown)


MoonE:
There is also a missing quote after both of the `degree` strings
https://github.com/openlayers/openlayers/blob/9a6ff5a321cc64e28a915632fdb46537a06823c5/changelog/upgrade-notes.md?plain=1#L11

marcjansen:
Thanks @MoonE & @ahocevar

Issue #13639, opened 2022-05-03
RGB COG with float values not shown?


m-mohr:
**Describe the bug**
It looks like COGs that contain floats can not always be rendered correctly (at least it is the only difference I see right now to COGs that work for me and contain ints).

**To Reproduce**
1. Go to https://codesandbox.io/s/cog-forked-pt8yxd?file=/main.js
2. It doesn't show anything. I also can't get pixel values on pointer move.

**Expected behavior**
I'd expect to render similarly to what QGIS shows for the same COG with the same min/max values:
![image](https://user-images.githubusercontent.com/8262166/166424546-9ebd0125-dac9-4d5b-a6ea-55572a3e2199.png)

mike-000:
It may be a problem with the tile grid.  https://codesandbox.io/s/cog-forked-wptn6m?file=/main.js  There is only 1 zoom level and I would not expect so many tiles at level 0.

m-mohr:
@mike-000 Are you saying that I should generate the COG differently (e.g. multiple zoom levels?) or that this looks like an issue in OL?

mike-000:
Tiles were being requested but going the `TileState.ERROR`, as geotiff.js is throwing an error which can be logged by adding `.catch` to the code https://codesandbox.io/s/cog-forked-nhb37m?file=/main.js  Not sure if that is due to the data or the requests being invalid.

m-mohr:
Interesting, it says "Unsupported data format/bitsPerSample". I'll try to check why this might be thrown...

m-mohr:
The error occurs in geotiffimage.js, function arrayForType. The arguments passed are format = undefined, bitsPerSample = undefined, size = 65536. So it's weird that the values are undefined.

The undefined values originate from getArrayForSample, where for the sampleIndex'es 0 and 1 it tries to read from the fileDirectory, but only 0 is defined in there.

If I change two functions slightly in geotiffimage.js, it at least doesn't throw the error any more, but not so sure whether it is meaningful:

```js


  getSampleFormat(sampleIndex = 0) {
    if (!this.fileDirectory.SampleFormat || this.fileDirectory.SampleFormat.length === 0) {
      return 1;
    }

    return typeof this.fileDirectory.SampleFormat[sampleIndex] !== 'undefined'
      ? this.fileDirectory.SampleFormat[sampleIndex] : this.fileDirectory.SampleFormat[0];
  }

  getBitsPerSample(sampleIndex = 0) {
[...]

PR #13645, opened 2022-05-04
Output GeoTIFF tile load errors to console

mike-000:
#13639 highlighted that while GeoTIFF source load errors are output for debugging tile load errors are handled silently by the parent class.  This catches and outputs the error before rethrowing for handling by the parent.

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -770,6 +770,11 @@ class GeoTIFFSource extends DataTile {
       }
 
       return data;
+    })
+    .catch(function (error) {
+      // output then rethrow
+      console.error(error); // eslint-disable-line no-console
+      throw error;
     });
   }
 }


ahocevar:
We do have a built-in way to get source errors, but it is not visible in the API docs:

https://github.com/openlayers/openlayers/blob/c2639f89f23edaecbd2a29c1818eadfbe1b2cc4d/src/ol/source/GeoTIFF.js#L427-L440

But maybe you're right, @mike-000, that it is easier for the user to see the error in the console. What do others think?

mike-000:
> We do have a built-in way to get source errors, but it is not visible in the API docs:

As that documented as a source loading error I did not update the `.error_` property for a tile loading error.  However the error is also output to the console when that is set:

https://github.com/openlayers/openlayers/blob/c2639f89f23edaecbd2a29c1818eadfbe1b2cc4d/src/ol/source/GeoTIFF.js#L419-L423

ahocevar:
So instead of adding an additional log in a separate place, this pull request should be changed to ensure that GeoTIFF errors are handled by the same `catch` block.

mike-000:
It is a different part of the code.  In #13639 the source had loaded successfully to `SourceState.READY` after construction without triggering any errors, the errors were occurring later when loading tiles at the request of the renderer where they were silently caught to generate `TileState.ERROR`.

Issue #13500, opened 2022-03-23
WebGL interpolation on Android devices


mike-000:
**Describe the bug**
WebGL interpolation behaves differently on Android and Windows devices.

**To Reproduce**
Run https://codesandbox.io/s/webgl-sea-level-forked-bsd855 on Windows and Android devices (I tried Android 12 and 4.4, Chrome and Opera, both give similar results).
On Windows the "sea" has a smooth interpolated edge.
![image](https://user-images.githubusercontent.com/49240900/159714726-1bb6ee20-70a7-434b-aafc-aab637523555.png)
On Android at some sea levels and locations the edge may be repeated or jagged
![image](https://user-images.githubusercontent.com/49240900/159715105-8b7b9ade-291a-4599-9335-5d3921ac00d2.png)
This does not appear to be related to pixel ratio as setting the map pixel ratio to 1 does make any noticeable difference.  The output from code used in rendering tests such as `webgl-data-tile-interpolate-true`  and `source-raster-webgl` does appear as expected when run on the same Android devices.

**Expected behavior**
Some differences between hardware might be expected but this seems more basic as both new and quite old Android devices produce the same unexpected result.

mike-000:
The problem seems related to the calculation of elevation from RGB values in the style.  If the RGB tiles are converted to Float32 data tiles the interpolation produces the same smooth edge as in Windows https://codesandbox.io/s/webgl-sea-level-forked-bkyl0f. To handle all scenarios, such as `ReprojTile` and `gutter` a convenience method for `ol/source/TileImage` which returns promise of a canvas which can be used to populate the data tile might be desirable.

tschaub:
Was this fixed by #13505?

mike-000:
> Was this fixed by #13505?

No, but #13505 will allow the data tile method described above to be used on all devices if necessary, albeit without interpolation on older devices which do not support `OES_texture_float_linear` (similar to the canvas Sea Level example).

I have only checked this with Windows and Android - is anyone able to check https://bsd855,csb.app produces a smooth edge on any other systems, and if not is it fixed in https://bkyl0f,csb.app?

mike-000:
Checks on BrowserStack and LambdaTest show this affects almost all mobile devices, both Android and IOS.  It seems to be a precision issue despite `highp` being used - with DEM tiles it typically occurs close a change in the green cbannel.

[...]

PR #13647, opened 2022-05-05
Contours interpolation using Terrarium tiles with gutter

mike-000:
Replaces #13365

As noted in https://github.com/openlayers/openlayers/pull/13609#issuecomment-1118340667 the example in #13365 suffers noticeably from tile edge effects and also produces poor results on mobile devices due to #13500.  The former is overcome by using Terrarium tiles with gutter, and the latter by creating, where supported, Float32 data tiles from the image tiles.

I found the Nextzen tile source with gutter to be unreliable so instead dynamically created one buffered tile from nine adjacent tiles loaded from AWS. https://deploy-preview-13647--ol-site.netlify.app/en/latest/examples/contour-interpolation.html

### DIFF
--- a/examples/contour-interpolation.html
+++ b/examples/contour-interpolation.html
@@ -1,0 +1,24 @@
+---
+layout: example.html
+title: Interpolating Contours from a DEM
+shortdesc: Example of WebGL data interpolation
+docs: >
+  <p>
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Thanks, @mike-000. A complex, but useful example!

Issue #13651, opened 2022-05-07
Draw shapes example don't work with parcel


arekgotfryd:
**Describe the bug**
Currently if I copy code from below example 
[https://openlayers.org/en/latest/examples/draw-shapes.html](https://openlayers.org/en/latest/examples/draw-shapes.html)
and set it up to run locally on my machine I would get following error

🚨 Build failed.

@parcel/transformer-js: Browser scripts cannot have imports or exports.

 example\main.js:1:1
  > 1 | import 'ol/ol.css';
  >   | ^^^^^^^^^^^^^^^^^^^
    2 | import Draw, {
    3 |   createBox,

  example\index.html:7:5
    6 |     
  > 7 |     <script src="https://unpkg.com/elm-pep@1.0.6/dist/elm-pep.js"></script>
  >   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ The environment was originally created here
    8 |     
    9 |     <script src="https://cdn.polyfill.io/v3/polyfill.min.js?features=fetch,requestAnimationFrame,Element.prototype.classList,TextDecoder"></script> 

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/draw-shapes.html
2. Copy code from that page and create three files:
main.js, index.html and package.json under common directory
3. Run npm install to install dependencies
4. Run npm start

**Expected behavior**
I would expect this example to run without an issue in my local environment after copying 
code from the page and installing dependencies. I believe that this might be the case for all of the examples
and I'm willing to provide pull request which fix that issue for all of them at once.

ahocevar:
@arekgotfryd Thanks for willing to fix this! Recent versions of parcel do not bundle in dev mode, but use modules as supported by the browser. Therefore, the fix would be to change a single line of code, from
```js
<script src="main.js"></script>
```
to
```js
<script type="module" src="main.js"></script>
```
Would be great if you could submit a pull request that changes the template accordingly.

PR #13656, opened 2022-05-08
Fix for parcel error while building examples

arekgotfryd:
Fixes https://github.com/openlayers/openlayers/issues/13651

### DIFF
--- a/examples/templates/example.html
+++ b/examples/templates/example.html
@@ -208,7 +208,7 @@
 {{#if css.source}}{{ indent css.source spaces=6 }}{{/if}}    &lt;/style&gt;
   &lt;/head&gt;
   &lt;body&gt;
-{{ indent contents spaces=4 }}    &lt;script src="main.js"&gt;&lt;/script&gt;
+{{ indent contents spaces=4 }}    &lt;script type="module" src="main.js"&gt;&lt;/script&gt;
   &lt;/body&gt;
 &lt;/html&gt;</code></pre>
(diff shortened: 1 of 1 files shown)


examples/templates/example.html
```diff
@@ -198,6 +197,9 @@ <h5 class="source-heading">index.html</h5>
     &lt;script src="https://cdn.polyfill.io/v3/polyfill.min.js?features=fetch,requestAnimationFrame,Element.prototype.classList,TextDecoder"&gt;&lt;/script&gt;
     &lt;script src="https://cdnjs.cloudflare.com/ajax/libs/core-js/3.18.3/minified.js"&gt;&lt;/script&gt;{{#if extraHead.remote}}
 {{ indent extraHead.remote spaces=4 }}{{/if}}
+    &lt;style&gt;
+    @import "node_modules/ol/ol.css";
+    &lt;/style&gt;
```
ahocevar, line 202: To make this work in codesandbox.io ("Edit" button), you'll have to change it to
```html
&lt;link rel="stylesheet" href="node_modules/ol/ol.css"&gt;
```

ahocevar (changes requested):
Thanks @arekgotfryd! See my comment about changing how the css is included. Looks good otherwise.

examples/templates/example.html
arekgotfryd: Thanks @ahocevar! I just made suggested change.

PR #13672, opened 2022-05-13
fix currentClip == null

### LINKED ISSUE #13671: Type Error: clip is null
**Describe the bug**
Error during rendering of vector tiles in renderer/canvas/VectorTileLayer.js:
that clip is null (because currentClip was null before).

**To Reproduce**
Assumption:
* MVT tiles
* Mapbox layer style that includes a "filter" expression
* when zooming in, a tile is processed that has no (rendered) features because of the filter

**Expected behavior**
As ExecutorGroup.getClipCoords can definitely return null, ignore the null currentClip.

cns-solutions-admin:
Fixes #13671.
As ExecutorGroup.getClipCoords can definitely return null, we need to check the return value.


ahocevar (approved):
Thanks, @cns-solutions-admin

PR #13673, opened 2022-05-13
Add  null return type

ahocevar:
Follow up on #13672: add `null` as return type for `getClipCoords()`, in preparation for the `strictNullChecks: true` TypeScript configuration.

### DIFF
--- a/src/ol/render/canvas/ExecutorGroup.js
+++ b/src/ol/render/canvas/ExecutorGroup.js
@@ -287,7 +287,7 @@ class ExecutorGroup {
 
   /**
    * @param {import("../../transform.js").Transform} transform Transform.
-   * @return {Array<number>} Clip coordinates.
+   * @return {Array<number>|null} Clip coordinates.
    */
   getClipCoords(transform) {
     const maxExtent = this.maxExtent_;


MoonE (approved):

src/ol/render/canvas/ExecutorGroup.js
```diff
@@ -287,7 +287,7 @@ class ExecutorGroup {
 
   /**
    * @param {import("../../transform.js").Transform} transform Transform.
-   * @return {Array<number>} Clip coordinates.
+   * @return {Array<number>|null} Clip coordinates.
```
EvertEt, line 290: Would it be good to also add a guard in the `ExecutorGroup.clip` already?

ahocevar: @EvertEt I did not do that, because it would be inefficient. The caller function already guards against the condition that would make `getClipCoords()` return `null`. So some refactoring would be in order there.

EvertEt: I see, since we did check it, I would normally consider casting (with an optional comment) but I am not sure the syntax keeps it readable and since it's not always guaranteed I am doubtful. But we will stumble across again when we enable strictNullChecks
```
    const flatClipCoords = /** @type {Array<number>} */ (
      this.getClipCoords(transform)
    );
```

PR #13676, opened 2022-05-15
Test improvements

MoonE:
- Use `document.head` instead of `document.getElementByTagName('head')`, it is well supported: [Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Document/head#browser_compatibility)
- Fix the two font re-rendering tests when the font familiy is already available as a system font. In that case no re-rending takes place and the test fails. It always failed for me as I have 'Droid Sans' installed.
It now uses a font-face declaration with a custom font family name to avoid any overlap with system fonts.
- The `DataTileSource.loader` function expects a `TypedArray` as a return type, not the `ImageData` itself. Didn't seem to be relevant for the tests though as `rendercomplete` event fires as long as the tile is in either loaded or error state and the style update test does not care about the data.
- The changed KML test used a 200 ms timeout to wait for an icon to load, this is  unnecessary long in most cases and may fail when the loading takes unusually long, e.g. when debugging the test. Listening for image state change events solves this.
- The MapboxVector tests unexpectedly generated an error in https://github.com/openlayers/ol-mapbox-style/blob/f01442d77fc9c3ded7672146defa96fab0449d3c/src/util.js#L67 after it succeeded because `fetch` is expected to always resolve with an object.


tschaub (approved):
Thanks, @MoonE

PR #13688, opened 2022-05-20
Flip extent coordinates for projections with ne* axis order

ahocevar:
This pull request fixes a problem with extents defined in WMTS Capabilities Tile Matrix Sets, when a projection with inverted axis order is used.

To solve this, the same technique we use for flipping origin coordinates already is used.

### DIFF
--- a/test/browser/spec/ol/source/WMTS.test.js
+++ b/test/browser/spec/ol/source/WMTS.test.js
@@ -475,6 +475,34 @@ describe('ol/source/WMTS', function () {
       expectDelta(extent[3], expectedMatrixSetExtend[3]);
     });
   });
+
+  describe('when creating options from epsg:4326 capabilities with BoundingBox', function () {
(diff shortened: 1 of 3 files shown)


test/browser/spec/ol/source/WMTS.test.js
```diff
@@ -475,6 +475,34 @@ describe('ol/source/WMTS', function () {
...
+      );
+    });
+
+    it('returns correct bounding box when the layer has BoundingBox', function () {
+      const options = optionsFromCapabilities(capabilities, {
+        layer: 's2cloudless-2020',
+      });
+
+      const extent = options.tileGrid.getExtent();
+      expect(extent).to.eql([-180, -90, 180, 90]);
```
ahocevar, line 502: Without the fix from this pull request, the extent would be `[-90, -180, 90, 180]`.

tschaub (approved):

PR #13689, opened 2022-05-20
Link interaction

tschaub:
This adds a link interaction for synchronizing map state and the URL.

By default, `x`, `y` (center coords), `z` (zoom level), `r` (rotation), and `l` (layer visibility) search parameters are read and written.  A `prefix` option can be used to prefix these values.

By default, history entries are added for changes in the map state.  The `replace` option can be used to replace state instead of pushing a new entry.

By default, navigating back through history animates transitions in the map.  The `animate` option can be used to control this (`false` to disable animation or `{duration: 100}` to speed it up, for example).

Example link: https://deploy-preview-13689--ol-site.netlify.app/examples/link.html?x=-12361408.89207&y=5729886.94037&z=10.42345&r=0&l=1

### DIFF
--- a/src/ol/PluggableMap.js
+++ b/src/ol/PluggableMap.js
@@ -605,6 +605,9 @@ class PluggableMap extends BaseObject {
    * Clean up.
    */
   disposeInternal() {
+    this.controls.clear();
+    this.interactions.clear();
+    this.overlays_.clear();
     this.setTarget(null);
     super.disposeInternal();
   }
--- a/src/ol/interaction/MouseWheelZoom.js
+++ b/src/ol/interaction/MouseWheelZoom.js
@@ -155,7 +155,11 @@ class MouseWheelZoom extends Interaction {
    */
   endInteraction_() {
     this.trackpadTimeoutId_ = undefined;
-    const view = this.getMap().getView();
+    const map = this.getMap();
+    if (!map) {
+      return;
+    }
+    const view = map.getView();
     view.endInteraction(
       undefined,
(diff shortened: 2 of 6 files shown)


src/ol/PluggableMap.js
```diff
@@ -605,6 +605,9 @@ class PluggableMap extends BaseObject {
    * Clean up.
    */
   disposeInternal() {
+    this.controls.clear();
+    this.interactions.clear();
+    this.overlays_.clear();
```
tschaub, line 610: The link interaction modifies the URL when added to a map.  The interaction removes its search parameters from the URL when it is removed from a map.  The tests make use of `map.dispose()` to clean up, and they rely on the URL not having extra junk in it, so `map.dispose()` now removes interactions (and controls and overlays) so that additional cleanup can be done.

src/ol/interaction/MouseWheelZoom.js
```diff
@@ -155,7 +155,11 @@ class MouseWheelZoom extends Interaction {
    */
   endInteraction_() {
     this.trackpadTimeoutId_ = undefined;
-    const view = this.getMap().getView();
+    const map = this.getMap();
+    if (!map) {
+      return;
+    }
+    const view = map.getView();
```
tschaub, line 162: This method is called in a timeout, with no guarantees that the interaction still has a map when it runs.  In the tests, we now remove interactions from the map in test cleanup, and this method was being called after the tests complete.

ahocevar (approved):
A very useful new feature. Thanks for the thorough work on this!

Issue #13690, opened 2022-05-21
Vector source getFeaturesInExtent and projection parameter and support canWrapX


burleight:
I work primarily with maps and data in the Pacific covering either side of longitude +/- 180. The following code does not behave as expected: 

`vectorSource.getFeaturesInExtent(view.getExtent())`

Scroll the map to the east of 180 longitude or west of -180 longitude, and `getFeaturesInExtent` won't all of the visible features.

I'd like to add an optional `projection` parameter to `vectorSource.getFeaturesInExtent` so I could do the following to retrieve the expected features:

`
vectorSource.getFeaturesInExtent(view.getExtent(), view.getProjection())
`

PR #13691, opened 2022-05-21
#13690 VectorSource#getFeaturesInExtent add projection parameter

burleight:
Addresses #13690 

Adds optional projection parameter to `vectourSource.getFeaturesInExtent` and adds the interal function wrapAndSliceX to fit the extent into the projection extent, slicing it in two if necessary

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -14,7 +14,7 @@ import VectorEventType from './VectorEventType.js';
 import {TRUE, VOID} from '../functions.js';
 import {all as allStrategy} from '../loadingstrategy.js';
 import {assert} from '../asserts.js';
-import {containsExtent, equals} from '../extent.js';
+import {containsExtent, equals, wrapAndSliceX} from '../extent.js';
 import {extend} from '../array.js';
 import {getUid} from '../util.js';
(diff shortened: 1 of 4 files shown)


src/ol/source/Vector.js
```diff
@@ -736,12 +736,23 @@ class VectorSource extends Source {
    * features.
    *
    * @param {import("../extent.js").Extent} extent Extent.
+   * @param {import("../proj/Projection.js").default} [projection] Projection.
    * @return {Array<import("../Feature.js").default<Geometry>>} Features.
    * @api
    */
-  getFeaturesInExtent(extent) {
+  getFeaturesInExtent(extent, projection) {
```
ahocevar, line 743: The new `projection` argument should be optional. The documentation should state the purpose of this argument (i.e. needed for extents that cross the -180/180 line).

ahocevar (changes requested):
Thanks for your contribution, @burleight. Please address my comment, otherwise this looks good.

src/ol/source/Vector.js
burleight: Thanks. I've renamed it `opt_projection` and added to the documentation

ahocevar (approved):
Thanks, @burleight!

PR #13701, opened 2022-05-23
Update ESLint config and plugins

tschaub:
This brings in the latest `eslint-config-openlayers` with changes that allow us to update ESLint itself.

The `eslint-plugin-jsdoc` rules now prefer `object` to `Object` as a type.  We have some inconsistency in the library, but we spell it `Object` the majority of the time.  I've configured the settings in the `eslint-config-openlayers` package to prefer `Object` to `object` (see https://github.com/openlayers/eslint-config-openlayers/pull/36).  When we want to go the other way, we can change the setting there.

Closes #13693.

### DIFF
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -1189,8 +1189,8 @@ class Executor {
           }
           ++i;
           break;
-        default:
-          ++i; // consume the instruction anyway, to avoid an infinite loop
+        default: // consume the instruction anyway, to avoid an infinite loop
+          ++i;
           break;
       }
     }
--- a/src/ol/style/Style.js
+++ b/src/ol/style/Style.js
@@ -37,8 +37,7 @@ import {assert} from '../asserts.js';
  * 1. The pixel coordinates of the geometry in GeoJSON notation.
  * 2. The {@link module:ol/render~State} of the layer renderer.
  *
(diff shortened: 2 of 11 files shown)


src/ol/style/Style.js
```diff
@@ -37,8 +37,7 @@ import {assert} from '../asserts.js';
  * 1. The pixel coordinates of the geometry in GeoJSON notation.
  * 2. The {@link module:ol/render~State} of the layer renderer.
  *
- * @typedef {function((import("../coordinate.js").Coordinate|Array<import("../coordinate.js").Coordinate>|Array<Array<import("../coordinate.js").Coordinate>>),import("../render.js").State): void}
- * RenderFunction
+ * @typedef {function((import("../coordinate.js").Coordinate|Array<import("../coordinate.js").Coordinate>|Array<Array<import("../coordinate.js").Coordinate>>),import("../render.js").State): void} RenderFunction
```
tschaub, line 40: We don't currently run the `no-undefined-types` rule from `eslint-plugin-jsdoc`, but when we do, the name of the typedef will have to be on the same line as the `@typedef` tag.

src/ol/render/canvas/Executor.js
```diff
@@ -1189,8 +1189,8 @@ class Executor {
           }
           ++i;
           break;
-        default:
-          ++i; // consume the instruction anyway, to avoid an infinite loop
+        default: // consume the instruction anyway, to avoid an infinite loop
```
tschaub, line 1192: Prettier now wants these commends on the same line as `default`.

PR #13712, opened 2022-05-28
Fix tile pyramid getData()

### LINKED ISSUE #13706: getData() issue with GeoTIFF tile pyramid
**Describe the bug**
In case of GeoTIFF tile pyramid, when more than one tile represents the layer at the given zoom level, getData() only returns values for one of the tiles. 

**To Reproduce**
1. Go to https://codesandbox.io/s/cog-pyramid-forked-rc4e29?file=/index.html
2. Wait until the browser output tab loads (map should appear)
3. Hover mouse over any part of land in the map in the browser output, and observe the color lookup values (directly under map) change 
3. Zoom in over any land near 0th meridian (West Africa, West Europe)
4. At the highest zoom level with two tiles stacked side by side, getData() only returns values for the right tile (presumably because it was loaded last?)

**Expected behavior**
getData() should return values for all tiles representing the layer at a given zoom level, not just one tile.

**Notes**
1) This example is derived from OpenLayers GeoTIFF tile pyramid example (https://openlayers.org/en/latest/examples/cog-pyramid.html), by straightforward addition of the getData() functionality from https://openlayers.org/en/latest/examples/cog-math.html example. 
2) The GeoTIFF tile pyramid in this example has only one tile per zoom level for all lower zoom levels, and then getData() works fine. But the highest zoom level contains two tiles stacked side by side. getData() works only for one of them (the right one representing the Eastern hemisphere, likely due to it being loaded last). 
3) I have confirmed this behavior in another example with even more (eight) tiles at a higher zoom level, still, only one (bottom right) tile is seen by getData().

mike-000:
Fixes #13706

Instead of using `.getRenderSource()` which returns the last source used anywhere on the frame the renderer needs to determine the last source suitable for rendering at the coordinate.  Add test with minimal DataTile pyramid,


ahocevar (approved):
Thanks, @mike-000

PR #13713, opened 2022-05-28
PointerInteraction: handle pointer cancel events when tracking

### LINKED ISSUE #13238: While dragging, rotating and zooming a while, I cant handle the map
This is related only to touch screen devices!
I found a bug in updateTrackedPointers_ (File: Pointer.js Line: 191)
updateTrackedPointers_ does not listen to pointercancel-Event. Sometimes pointer events stay in trackedPointers_ list.
This happens, when the browser does not fire a pointerup event for an associated pointerdown event.
In updateActivePointers_ (File: MapBrowserEventHandler.js Line: 140) you do so.

luiscamachopt:
As explained by @bjoernboldt in #13238:

"updateTrackedPointers_ does not listen to pointercancel-Event. Sometimes pointer events stay in trackedPointers_ list.
This happens, when the browser does not fire a pointerup event for an associated pointerdown event.
In updateActivePointers_ (File: MapBrowserEventHandler.js Line: 140) you do so."

This PR updates `updateTrackedPointers_()` to also listen to `MapBrowserEventType.POINTERCANCEL`

Fixes #13238

### DIFF
--- a/src/ol/interaction/Pointer.js
+++ b/src/ol/interaction/Pointer.js
@@ -193,7 +193,10 @@ class PointerInteraction extends Interaction {
       const event = mapBrowserEvent.originalEvent;
 
       const id = event.pointerId.toString();
-      if (mapBrowserEvent.type == MapBrowserEventType.POINTERUP) {
+      if (
+        mapBrowserEvent.type == MapBrowserEventType.POINTERUP ||
+        mapBrowserEvent.type == MapBrowserEventType.POINTERCANCEL
+      ) {
         delete this.trackedPointers_[id];
       } else if (mapBrowserEvent.type == MapBrowserEventType.POINTERDOWN) {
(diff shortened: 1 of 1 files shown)


src/ol/interaction/Pointer.js
```diff
@@ -193,7 +193,10 @@ class PointerInteraction extends Interaction {
       const event = mapBrowserEvent.originalEvent;
 
       const id = event.pointerId.toString();
-      if (mapBrowserEvent.type == MapBrowserEventType.POINTERUP) {
+      if (
+        mapBrowserEvent.type == MapBrowserEventType.POINTERUP ||
```
ahocevar, line 197: This change won‘t make a difference, because `isPointerDraggingEvent()` prevents this code path from being entered on `pointercancel`.

Issue #13714, opened 2022-05-30
The hit detection in styles with custom rendering at Retina display does not work well


hangbiaozhang:
**Describe the bug**
The hit detection in styles with custom rendering at Retina display does not work well.
`window.devicePixelRatio === 2`

**To Reproduce**
1.  Add a feature by custom rendering style and point geometry To map
2. Draw a fill color circle at renderer, center point 
3. Add 'click' listener To map
4. Get feature by call `map.forEachFeatureAtPixel()` at the listener function, `hitTolerance = 0`
5. At Retina display click out of the circle can still get the feature, the active area is double the radius

**Expected behavior**
The active area only in circle

**Debug place**
At `ol.render.canvas.ExecutorGroup.forEachFeatureAtCoordinate` function
`var transform = composeTransform(this.hitDetectionTransform_, hitTolerance + 0.5, hitTolerance + 0.5, 1 / resolution, -1 / resolution, -rotation, -coordinate[0], -coordinate[1])`
Is it necessary to consider `frameState.pixelRatio` when calculating transform
Just a personal opinion, hope it helps

ahocevar:
@hangbiaozhang Thanks for reporting this! Do you think you could create a pull request with your suggested fix?

hangbiaozhang:
@ahocevar I created a pull request #13722

hangbiaozhang:
@ahocevar 
Thanks for reply
I find another way to fix the problem, change the custom renderer function. So I close the pull request and issue.

```
feature.setStyle(
  new Style({
    renderer(coordinates, state) {
      const scalePixel = function (value) {
        if (
          state.context.canvas.width === 1 &&
          state.context.canvas.height === 1
        ) {
          return value;
        } else {
          return state.pixelRatio * value;
        }
      };
      const [x, y] = coordinates;
      const ctx = state.context;
      ctx.beginPath();
      ctx.arc(x, y, scalePixel(30), 0, 2 * Math.PI, true);
      ctx.fillStyle = '#FFFF00';
      ctx.fill();
      ctx.strokeStyle = 'rgba(255,0,0,1)';
      ctx.stroke();
    },
  })
);
```

PR #13722, opened 2022-06-01
Fixes the hit detection does not work well at Retina display

hangbiaozhang:
The hit detection in styles with custom rendering at Retina display does not work well #13714

### DIFF
--- a/src/ol/render/canvas/ExecutorGroup.js
+++ b/src/ol/render/canvas/ExecutorGroup.js
@@ -158,6 +158,7 @@ class ExecutorGroup {
    * @param {import("../../coordinate.js").Coordinate} coordinate Coordinate.
    * @param {number} resolution Resolution.
    * @param {number} rotation Rotation.
+   * @param {number} pixelRatio The pixel ratio of the frame.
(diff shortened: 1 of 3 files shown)


ahocevar:
Thanks for this attempt to fix the problem. It looks like this breaks hit detection for standard image styles, see e.g. https://deploy-preview-13722--ol-site.netlify.app/examples/icon.html.

hangbiaozhang:
Thanks for reply
I find another way to fix the problem, change the custom renderer function. So I close the pull request and issue.
```
feature.setStyle(
  new Style({
    renderer(coordinates, state) {
      const scalePixel = function (value) {
        if (
          state.context.canvas.width === 1 &&
          state.context.canvas.height === 1
        ) {
          return value;
        } else {
          return state.pixelRatio * value;
        }
      };
      const [x, y] = coordinates;
      const ctx = state.context;
      ctx.beginPath();
      ctx.arc(x, y, scalePixel(30), 0, 2 * Math.PI, true);
      ctx.fillStyle = '#FFFF00';
      ctx.fill();
      ctx.strokeStyle = 'rgba(255,0,0,1)';
      ctx.stroke();
    },
  })
);
```

PR #13745, opened 2022-06-09
Fix modifying polygons with overlapping vertices

### LINKED ISSUE #13736: Dragging an overlapping vertex of a polygon results in an invalid geometry (open exterior ring).
**Describe the bug**
When dragging (ol/interaction/modify) a vertex that overlaps with the first vertex of the exterior ring of the same polygon, the exterior ring of the polygon isn't closed anymore. It appears that the final vertex in the exterior ring isn't updated to match the first vertex and the ring becomes open. In my case, the two vertices are overlapping because they snapped to each other.

**To Reproduce**
1. Go to https://codesandbox.io/s/draw-and-modify-features-forked-vdzj3e?file=/main.js
2. Drag the two overlapping vertices. 
3. The first and last vertices are logged to the console at `modifystart` and `modifyend`. In `modifyend`, they're not the same anymore and so the exterior ring of the polygon isn't closed.

**Expected behavior**
The exterior ring of the polygon to remain closed after the interaction. When dragging two overlapping vertices, I would expect either:

- only one of the vertices to be dragged at a time **or**
- if the first vertex in the ring is one of the vertices being dragged, the final vertex to be updated to match the first one.

I tried to find the source of the issue and I believe it's [these lines](https://github.com/openlayers/openlayers/blob/main/src/ol/interaction/Modify.js#L1032). When the polygon is being modified, only two segments for the polygon are being tracked, whereas for two overlapping vertices, there are four segments that need to be updated. Usually this isn't an issue as when overlapping vertices are dragged, the two segments connect to the same vertex. But in the case of the first vertex, it finds one segment for it, and then a segment for the overlapping vertex.

I would be happy to open a PR for this, so that more than two segments can be tracked.

hargasinski:
This PR fixes #13736. It adds a check to make sure that when you are modifying the first vertex of a polygon, the other segment selected contains the closing vertex. I thought about allowing more than two segments to be dragged for overlapping vertices, but that would change the current behaviour (for overlapping vertices that don't include the first vertex, modifying them only allows one vertex to be dragged at a time). I tried to follow the style in `removeVertex_` for this.

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -1042,15 +1042,32 @@ class Modify extends PointerInteraction {
           coordinatesEqual(segment[1], vertex) &&
           !componentSegments[uid][1]
         ) {
-          // prevent dragging closed linestrings by the connecting node
           if (
-            (segmentDataMatch.geometry.getType() === GeometryType.LINE_STRING ||
-              segmentDataMatch.geometry.getType() ===
-                GeometryType.MULTI_LINE_STRING) &&
             componentSegments[uid][0] &&
             componentSegments[uid][0].index === 0
           ) {
-            continue;
+            let coordinates = segmentDataMatch.geometry.getCoordinates();
+            switch (segmentDataMatch.geometry.getType()) {
+              // prevent dragging closed linestrings by the connecting node
+              case GeometryType.LINE_STRING:
+              case GeometryType.MULTI_LINE_STRING:
+                continue;
+              // if dragging the first vertex of a polygon, ensure the other segment
+              // belongs to the closing vertex of the linear ring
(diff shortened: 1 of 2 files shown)


hargasinski:
The failing test doesn't appear to be related to the PR. It's one of the CORs tests and it's saying the test timed out, so maybe the request took too long? I tried running the tests again locally, and they're passing, so any idea on what could be happening?

ahocevar:
@hargasinski Thanks for your work on this! I triggered a re-run of the test. Let's see what happens now.

hargasinski:
@ahocevar that appears to have fixed it, thanks for your help!

ahocevar (approved):
Beautifully solved. Thanks, @hargasinski!

PR #13747, opened 2022-06-09
Add setFill and setStroke to Shapes

theduckylittle:
This rounds out the shape style sto be more
consistent and make the `setFill` and `setStroke`
methods available.


ahocevar (commented):
Thanks, @theduckylittle. I think it would make sense to add a test to to make sure the `setFill()` and `setStroke()` methods do what they are supposed to, now that they are part of the public API. Preferably a rendering test, because given the fact that an image is created at RegularShape construction time, I'm not sure changing fill and stroke later on works at all.

MoonE:
Reusing RegularShape like this would be rather inefficient as @ahocevar noted. The currently available setters all work without a need to re-render the cached canvas-image.

tschaub:
I haven't looked that closely, but it may enough to call `this.render()` after updating properties (like fill or stroke) that would invalidate previously cached versions.

theduckylittle:
- Updated the regular shapes example that demonstrates changing the stroke/fill of a regular shape style.
- Added a rendering test for both `setFill` and `setStroke`.
- @tschaub was spot-on, calling `this.render` did invalidate the old cached images and they re-rendered properly. (https://github.com/openlayers/openlayers/pull/13747/files#diff-e3750617700415f875eb42b59eb88fae43039d3037b66e989ffa9572c81e34f4R210)

theduckylittle:
> Reusing RegularShape like this would be rather inefficient as @ahocevar noted. The currently available setters all work without a need to re-render the cached canvas-image.

This really does round-out the setter functionality that is common across other classes. While there may be other approaches that can be more performant in specific environments, this is still a worthwhile rounding-out of the API that does address a set of missing functionality.

tschaub (approved):

tschaub:
Thanks for this addition, @theduckylittle

PR #13763, opened 2022-06-13
Fix single selection bug when holding modifier keys

pilatte:
## The problem

- A Layer has two features A and B.
- A Select interaction is active on the layer with `multi` **disabled**
- **You select A** (A is now selected)
- **You hold SHIFT and select B** (B is now selected)
- **You hold SHIFT and select A** (Nothing is selected)

## Result

The current SelectEvent holds no features

## Expected behavior

In single selection mode (with multi disabled), the last clicked feature should always be in the SelectEvent

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -478,7 +478,7 @@ class Select extends Interaction {
     const features = this.getFeatures();
     const deselected = [];
     const selected = [];
-    if (set) {
+    if (set || !this.multi_) {
       // Replace the currently selected feature(s) with the feature(s) at the
       // pixel, or clear the selected feature(s) if there is no feature at
       // the pixel.


pilatte:
The expected default behavior is described in doc as such:

> Clicking on a feature selects that feature and removes any that were in the selection.

pilatte:
I think I don't understand multiple selection then. I thought it was the map equivalent of radio-buttons vs checkboxes.

In your example, whether `multi` is true or false, `SHIFT` is required to make multiple selection.

And setting `multi` to false doesn't prevent multiple selection.

Is `multi` only used to select single features when they overlap ?

ahocevar:
> I think I don't understand multiple selection then. I thought it was the map equivalent of radio-buttons vs checkboxes.

That's the toggle behavior, which can be controlled with the `toggleCondition`.

> In your example, whether multi is true or false, SHIFT is required to make multiple selection.

Correct. If you want multiple selection active without modifier key, configure the interaction with `toggleCondition: () => true`.

> And setting multi to false doesn't prevent multiple selection.

Correct. If you want to prevent multiple selection, you have to configure the control with `toggleCondition: () => false`

> Is multi only used to select single features when they overlap ?

Correct.  When features overlap, or when you click the border between to areas in the example above, setting `multi: true` will cause both areas to be selected, instead of just the one that lies on top.

pilatte:
Ok, thanks for taking the time to clarify. 👍  Closing this now

Issue #13753, opened 2022-06-13
Be able to zoom out with double tap


oterral:
**Is your feature request related to a problem? Please describe.**
Mapbox and google maps provides a way to zoom out using double tap.  It would be nice to also have this gesture in ol.

**Describe the solution you'd like**
Add an interaction that zoom out when the user tap the screen with 2 fingers

PR #13752, opened 2022-06-13
Add DblClickDragZoom interaction

oterral:
Proposal for https://github.com/openlayers/openlayers/issues/13751 .

You can test the behavior in the [drag-rotate-and-zoom.html example](https://deploy-preview-13752--ol-site.netlify.app/examples/drag-rotate-and-zoom.html)

PR #13765, opened 2022-06-14
Add DblPointerClickZoom interaction

oterral:
Proposal for https://github.com/openlayers/openlayers/issues/13753

You can test the behavior in the [drag-rotate-and-zoom.html example](https://deploy-preview-13752--ol-site.netlify.app/examples/drag-rotate-and-zoom.html)

One problem with this interaction is that it could be in conflict with DragPan interaction if it is not added the first.

@ahocevar  I guess your comment in https://github.com/openlayers/openlayers/pull/13752 is relevant for this PR.


oterral:
Other proposal  https://github.com/openlayers/openlayers/pull/13766

PR #13771, opened 2022-06-15
Better fix for changing pointer ids on event target change

ahocevar:
This reverts the changes made in #13770, and suggests a better fix on a lower level of the event handling chain.

Fixes https://github.com/openlayers/openlayers/issues/11465
Fixes https://github.com/openlayers/openlayers/issues/13746
Fixes https://github.com/openlayers/openlayers/issues/13238
Closes https://github.com/openlayers/openlayers/pull/13713

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -73,7 +73,7 @@ class MapBrowserEventHandler extends Target {
     this.activePointers_ = 0;
 
     /**
-     * @type {!Object<number, boolean>}
+     * @type {!Object<number, Event>}
      * @private
      */
     this.trackedTouches_ = {};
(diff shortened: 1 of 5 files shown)


src/ol/MapBrowserEventHandler.js
```diff
@@ -169,14 +169,26 @@ class MapBrowserEventHandler extends Target {
...
     } else if (event.type == MapBrowserEventType.POINTERDOWN) {
-      this.trackedTouches_[event.pointerId] = true;
+      this.trackedTouches_[id] = event;
+    } else if (event.type == MapBrowserEventType.POINTERMOVE) {
+      for (const pointerId in this.trackedTouches_) {
+        if (this.trackedTouches_[pointerId].target !== event.target) {
+          // Some platforms assign a new pointerId when the target changes.
+          // If this happens, delete one tracked pointer. If there is more
+          // than one tracked pointer for the old target, it will be cleared
+          // by subsequent POINTERUP events from other pointers.
```
ahocevar, line 187:
```suggestion
          // by subsequent pointermove events from other pointers.
```

luiscamachopt:
This sounds related to bug #13238 and corresponding proposed fix in #13713.
Just mentioning in case you haven't seen those @ahocevar

ahocevar:
Thanks @luiscamachopt, the fix there is similar to my first fix, which was not so good. Better to fix this on a lower level.

ahocevar:
I'll need to dig a bit deeper to fix this. Currrently, `PointerInteraction` does mot make use of the `MapBrowserEvent`s we have for such purposes. It has its own pointer tracking code, which is an unnecessary code duplication. In addition, the `MapBrowserEvent` would be useful because it provides a higher level `pointerdrag` event.

My plan is to rework the `PointerInteraction` to use `MapBrowserEvent`s, and fix the pointer handling in `MapBrowserEventHandler` like proposed in this pull request.

src/ol/MapBrowserEventHandler.js
```diff
@@ -303,29 +324,36 @@ class MapBrowserEventHandler extends Target {
...
    * @private
    */
-  relayEvent_(pointerEvent) {
+  relayMoveEvent_(pointerEvent) {
     this.originalPointerMoveEvent_ = pointerEvent;
     const dragging = !!(this.down_ && this.isMoving_(pointerEvent));
     this.dispatchEvent(
-      new MapBrowserEvent(pointerEvent.type, this.map_, pointerEvent, dragging)
+      new MapBrowserEvent(
+        MapBrowserEventType.POINTERMOVE,
```
ahocevar, line 352: For readability: renamed method from `relayEvent_` to `relayMoveEvent_`, and use a onstant event type, because this is only called for `pointermove`.

tschaub (approved):

rossholdway:
Thank you for looking in to this. 🚀 
How soon can we expect to see this in a release?

ahocevar:
This fix is included in the v6.15.x release.

Issue #13781, opened 2022-06-20
this example not work in Codesandbox ?


jipexu:
Hi
**Describe the bug**
https://openlayers.org/en/latest/examples/d3.html
this exemple work well but not in codesandbox ?
**To Reproduce**
1. Go to '...' https://openlayers.org/en/latest/examples/d3.html
2. Click on '....' edit

**Expected behavior**
work as in openlayers examples....
It's only for me ?

Thanks

PR #13784, opened 2022-06-21
d3 version update

jipexu:
d3 version : 7.4.4


MoonE:
Does this solve #13781?

Would be great if you could also update [the second example where d3 is used](https://deploy-preview-13784--ol-site.netlify.app/examples/raster.html).

jipexu:
@MoonE 
no not solve the codesansbox probleme...
yes sure i will update the second example too !

MoonE (approved):
Thanks, @jipexu

PR #13787, opened 2022-06-22
Checkcontenteditable

### LINKED ISSUE #13785: Keyboard events from "contenteditable" elements trigger map controls
**Describe the bug**
Pressing + or - on a "content editable" div zooms the map.

**To Reproduce**
1. Go to any example like https://openlayers.org/en/latest/examples/bing-maps.html
2. Run this on console to create a content editable div on map
```
var div = document.createElement('div');
div.textContent = 'Edit me +- ';
var input = document.createElement('input');
input.value = 'This works +-';
div.appendChild(input);
div.setAttribute('contenteditable', true);
div.style = 'background-color: red; position: fixed; top: 40%; left: 40%;';
document.getElementById('map').appendChild(div);
```
3. Focus on the input next to the red text and press + or - (input value changes and map isn't zoomed)
4. Click on the the text on the red background to edit it and press + or - (text changes but map is zoomed)

**Expected behavior**
Map isn't zoomed when editing the input field content nor when editing the editable text.

I think the fix would be to add a check for content editable on this line: https://github.com/openlayers/openlayers/blob/v6.14.1/src/ol/events/condition.js#L253
```
  return tagName !== 'INPUT' && tagName !== 'SELECT' && tagName !== 'TEXTAREA' && !originalEvent.target.isContentEditable;
```
I don't have the dev env set up so I can't test this change though. This problem occurs on our app that uses the https://draftjs.org editor as part of an app with OpenLayers map.

jipexu:
check also the contenteditable attribut ...
close #13785


jipexu:
worst ...

src/ol/events/condition.js
```diff
@@ -250,7 +250,12 @@ export const targetNotEditable = function (mapBrowserEvent) {
     mapBrowserEvent.originalEvent
   );
   const tagName = /** @type {Element} */ (originalEvent.target).tagName;
-  return tagName !== 'INPUT' && tagName !== 'SELECT' && tagName !== 'TEXTAREA';
+  return (
+    tagName !== 'INPUT' &&
+    tagName !== 'SELECT' &&
+    tagName !== 'TEXTAREA' &&
+    !originalEvent.target.isContentEditable;
```
M393, line 257:
```suggestion
    !originalEvent.target.isContentEditable
```

jipexu:
thank @M393 i retry ...

src/ol/events/condition.js
```diff
@@ -250,7 +250,12 @@ export const targetNotEditable = function (mapBrowserEvent) {
     mapBrowserEvent.originalEvent
   );
   const tagName = /** @type {Element} */ (originalEvent.target).tagName;
-  return tagName !== 'INPUT' && tagName !== 'SELECT' && tagName !== 'TEXTAREA';
+  return (
+    tagName !== 'INPUT' &&
+    tagName !== 'SELECT' &&
+    tagName !== 'TEXTAREA' &&
+    !originalEvent.target.isContentEditable
```
M393, line 257:
```suggestion
    // `isContentEditable` is only available on `HTMLElement`, but it may also be a
    // different type like `SVGElement`.
    // @ts-ignore
    !originalEvent.target.isContentEditable
```

src/ol/events/condition.js
```diff
@@ -239,7 +239,7 @@ export const shiftKeyOnly = function (mapBrowserEvent) {
 
 /**
  * Return `true` if the target element is not editable, i.e. not a `<input>`-,
- * `<select>`- or `<textarea>`-element, `false` otherwise.
+ * `<select>`-, `<textarea>`-element or `contentEditable` -attribut at true, `false` otherwise.
```
tschaub, line 242: I would change this to be “not an `input`, `select`, or `textarea` element and no `contenteditable` attribute.”

jipexu: @tschaub thanks ==> done

MoonE (approved):
Thanks for your work on this, @jipexu.

PR #13799, opened 2022-06-28
Do not refresh use time for tiles when collecting used source tiles

### LINKED ISSUE #13797: VectorTile infinite loading loop
**Describe the bug**
The vector tile layer may load tiles forever.

I'm exporting a map as a multi-page pdf, but for one specific task the export never completes.
The specific steps required to reproduce are in the linked codesandbox.

**To Reproduce**
1. Go to https://codesandbox.io/s/vectortilelayer-tile-loading-bug-ppqluq?file=/main.js
2. The map will get stuck in a rendering loop, tiles loading never completes

**Expected behavior**
The map should load tiles once and fire a rendercomplete event at the end.

M393:
Fixes #13797

`tileCache.get(cacheKey)` changes the least recently used state. That's bad when done just for tile expiration.

~Iterating the entire cache may be slower when the cache size is large, but there is currently no way to access an entry by key without updating the used time.~

### DIFF
--- a/src/ol/source/VectorTile.js
+++ b/src/ol/source/VectorTile.js
@@ -240,8 +240,9 @@ class VectorTile extends UrlTile {
     const tileCache = this.getTileCacheForProjection(projection);
     const usedSourceTiles = Object.keys(usedTiles).reduce((acc, key) => {
       const cacheKey = getCacheKeyForTileKey(key);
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @M393

PR #13800, opened 2022-06-29
Change WKB readFeature(s) return type to Feature

mike-000:
WKB `readFeature()`/`readFeatures()` has no option to return `RenderFeature` so the return type could be more precisely defined as `Feature`.  See https://stackoverflow.com/q/72798555/10118270


MoonE (approved):
Thanks, @mike-000

PR #13803, opened 2022-06-30
Include displacement and declutterMode in Icon style clone

### LINKED ISSUE #13802: Cloning an ImageStyle doesn't clone Icon displacement
**Describe the bug**

When cloning a Style containing an image, the `displacement` property of the Icon isn't cloned with it:

```js
const style = new Style({
   image: new Icon({
      src: 'path/to/svg',
      scale: 1.2,
      displacement: [10, 50],
   }),
   zIndex: 11,
})
```

Seems to be [this line](https://github.com/openlayers/openlayers/blob/v6.14.1/src/ol/style/Icon.js#L220)

**To Reproduce**
- 

**Expected behavior**

The displacement should be cloned from the original style.

mike-000:
Fixes #13802

Include displacement and declutterMode in Icon style clone and add to test

### DIFF
--- a/src/ol/style/Icon.js
+++ b/src/ol/style/Icon.js
@@ -238,6 +238,8 @@ class Icon extends ImageStyle {
       scale: Array.isArray(scale) ? scale.slice() : scale,
       size: this.size_ !== null ? this.size_.slice() : undefined,
       src: this.getSrc(),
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #13819, opened 2022-07-07
Remove HERE Maps example

### LINKED ISSUE #13215: HERE Maps example not working
**Describe the bug**
API key no longer working.  When I checked the account created for #11899 the status is suspended and requires payment details to reactivate, so has presumably exceeded the free limit.

mike-000:
Fixes #13215

Remove the example as it is no longer viable to use a free API key due to limits imposed.


ahocevar (approved):
Thanks, @mike-000

PR #13832, opened 2022-07-15
Base vector tile render tile grid on the source grid 

### LINKED ISSUE #13187: TilematrixSets must match to render maps performant
As we found out, Openlayers renders tiled vector layers faster if the TileMatrixSet of the view (or the projection of the view) and the vektorlayer are identical. 

So in Openlayers 6.5.0 we overloaded 
`ol.source.VectorTile.prototype.getTileGridForProjection` 
with
```js
= function(projection) {
  const code = projection.getCode();
  let tileGrid = this.tileGrids_[code];
  if (!tileGrid) {
      if (!projection.getDefaultTileGrid()) {
          const sourceTileGrid = this.tileGrid;
          tileGrid = ol.tilegrid.createForProjection(
          projection,
          undefined,
          sourceTileGrid
              ?  sourceTileGrid.getTileSize(sourceTileGrid.getMinZoom())
              : undefined
          );
          this.tileGrids_[code] = tileGrid;
      } else {
          tileGrid = projection.getDefaultTileGrid();

      }
      this.tileGrids_[code] = tileGrid;
  }
  return tileGrid;
}
```
With Openlayers 6.9.0, however, the corresponding functions and variables became private.
Is there another way to replace the projections tileMatrixSet? 
Since the poor performance makes our app unusable, especially under MS Windows, I would like to know how Openlayers can be made acceptably fast again.

mike-000:
Fixes #13187

Base the vector tile render tile grid on the source grid, with a factor 2 pyramid below that for overzooming.

Ensure the specified `zDirection` is used throughout.  This is especially important for grids with irregular tile sizes such as those in https://github.com/openlayers/openlayers/issues/13187#issuecomment-1005434098

This includes a fix to the `tileSizes` typedef for tile grids.  Numbers are clearly valid in the array (`createFromCapabilitiesMatrixSet` passes them where possible) but the typedef only permitted `ol/size` arrays.

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -510,7 +510,9 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
         let hitDetectionImageData = tile.hitDetectionImageData[layerUid];
         if (!hitDetectionImageData && !this.animatingOrInteracting_) {
           const tileSize = toSize(
-            tileGrid.getTileSize(tileGrid.getZForResolution(resolution))
+            tileGrid.getTileSize(
+              tileGrid.getZForResolution(resolution, source.zDirection)
+            )
           );
           const rotation = this.renderedRotation_;
           const transforms = [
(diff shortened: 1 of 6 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #13835, opened 2022-07-16
Recover from incorrect rebase after removal of string enums

ahocevar:
Fixes the problem encountered by @MoonE after #12696.


test/browser/spec/ol/renderer/layer.test.js
```diff
@@ -7,15 +7,27 @@ import View from '../../../../../src/ol/View.js';
...
+    layer = new Layer({});
     renderer = new LayerRenderer(layer);
   });
 
+  describe('#renderIfReadyAndVisible', function () {
+    it('updates revision when data is ready and layer is visible', function () {
+      layer.setVisible(true);
+      layer.getSourceState = function () {
+        return 'ready';
+      };
```
MoonE, line 24: To test for this regression it should return something other than `'ready'` and check whether the revision did not change. But it makes sense to test both cases.

PR #13836, opened 2022-07-16
Replace more enums with typedefs

MoonE:


src/ol/layer/VectorTile.js
```diff
@@ -127,12 +127,6 @@ class VectorTileLayer extends BaseVectorLayer {
      */
     this.un;
 
-    if (options.renderMode === VectorTileRenderType.IMAGE) {
-      //FIXME deprecated - remove this check in v7.
-      //eslint-disable-next-line
-      console.warn('renderMode: "image" is deprecated. Option ignored.')
-      options.renderMode = undefined;
-    }
```
jahow, line 135: I wonder if the following assertion is also not relevant any more with this change: https://github.com/openlayers/openlayers/blob/ecadbc7d5c7d4eef2c31c6a711dcbfdaf7e5bf41/src/ol/layer/VectorTile.js#L131-L135

This feels a little unnecessary now, doesn't it?

src/ol/layer/WebGLPoints.js
```diff
@@ -9,7 +9,7 @@ import {parseLiteralStyle} from '../webgl/ShaderBuilder.js';
 /**
  * @template {import("../source/Vector.js").default<import("../geom/Point.js").default>} VectorSourceType
  * @typedef {Object} Options
- * @property {import('../style/literal.js').LiteralStyle} style Literal style to apply to the layer features.
+ * @property {LiteralStyle} style Literal style to apply to the layer features.
```
MoonE, line 12: Apidoc generation failed with an error without this change presumable because without any exports `ol/style/literal.js` is not a module. Though with this change no apidocs are generated for the typedefs in there.
Maybe the file content could be added somewhere else?

src/ol/proj/Units.js
```diff
@@ -65,18 +28,26 @@ export function fromCode(code) {
...
 
+/**
+ * @typedef {Object} MetersPerUnitLookup
+ * @property {number} radians Radians
+ * @property {number} degrees Degrees
+ * @property {number} ft  Feet
+ * @property {number} m Meters
+ * @property {number} us-ft US feet
+ */
+
```
MoonE, line 39: Adding this typedef makes the METERS_PER_UNIT lookup object below typesafe.

src/ol/CollectionEventType.js
```diff
@@ -9,13 +9,11 @@ export default {
   /**
    * Triggered when an item is added to the collection.
    * @event module:ol/Collection.CollectionEvent#add
-   * @api
    */
   ADD: 'add',
   /**
    * Triggered when an item is removed from the collection.
    * @event module:ol/Collection.CollectionEvent#remove
```
MoonE, line 16: Should we add a typedef version of this enum for the apidoc and keep using the enum? Using only typedef for these does not catch typos.

ahocevar: I'm envisioning a broader change to handle event types. For the `on`, `once` and `un` API methods, we added type safety already a while ago with separate typedefs - in this case in https://github.com/openlayers/openlayers/blob/main/src/ol/Collection.js#L47-L54.

I started a branch a few day ago to bring this kind of type safety also to the internal methods. I created a draft pull request for this now: #13840.

PR #13838, opened 2022-07-17
Remove deprecated Overlay options

MoonE:
Is this how it is supposed to work now? The `autoPanOptions` was already entirely ignored.
Or should `autoPan` be a `boolean` option and `autoPanOptions` be kept.

### DIFF
--- a/src/ol/Overlay.js
+++ b/src/ol/Overlay.js
@@ -43,20 +43,7 @@ import {outerHeight, outerWidth, removeChildren, removeNode} from './dom.js';
  * probably set `insertFirst` to `true` so the overlay is displayed below the
  * controls.
  * @property {PanIntoViewOptions|boolean} [autoPan=false] Pan the map when calling
- * `setPosition`, so that the overlay is entirely visible in the current viewport?
- * If `true` (deprecated), then `autoPanAnimation` and `autoPanMargin` will be
- * used to determine the panning parameters; if an object is supplied then other
- * parameters are ignored.
- * @property {PanOptions} [autoPanAnimation] The animation options used to pan
- * the overlay into view. This animation is only used when `autoPan` is enabled.
- * A `duration` and `easing` may be provided to customize the animation.
- * Deprecated and ignored if `autoPan` is supplied as an object.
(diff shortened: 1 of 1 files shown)


ahocevar:
There is no `autoPanOptions` configuration option. It is `autoPan`. The `true` value of `autoPan` is deprecated. Expected values for `autoPan` are either `false` (the default), or an object consisting of `animation` (from the deprecated `autoPanAnimation` and `margin` (from the deprecated `autoPanMargin`).

MoonE:
I updated it according to your comment. 
Though I do wonder if the default should just be `undefined` without allowing `false`.

ahocevar:
I agree, it makes no sense to set this to `explicitly`. When no `autoPan` is configured, then no `autoPan` will happen.

src/ol/Overlay.js
```diff
@@ -180,18 +167,11 @@ class Overlay extends BaseObject {
...
-    if (autoPan && 'object' !== typeof autoPan) {
-      autoPan = {
-        animation: options.autoPanAnimation,
-        margin: options.autoPanMargin,
-      };
-    }
     /**
      * @protected
-     * @type {PanIntoViewOptions|false}
+     * @type {PanIntoViewOptions|undefined}
```
ahocevar, line 172: The `|undefined` here should be redundant, because you already set the `autoPan` option in `Options` optional.

But looking at the code, I see that any truthy value will enable auto pan with defaults, so it might make sense to set `boolean` as allowed type in addition to `PanIntoViewOptions` in the `Options`.

MoonE: In a non-strict context `| undefined` can be omitted because `null` / `undefined` is implicitly allowed.
If autoPan were a local variable its type could be inferred from context.
Class member types cannot be inferred because the variable could be set to something else in a different function.

Without `| undefined` and `strict` enabled you will get an error:
> Type 'PanIntoViewOptions | undefined' is not assignable to type 'PanIntoViewOptions'.
> Type 'undefined' is not assignable to type 'PanIntoViewOptions'.ts(2322)

Makes sense to allow `true` for auto-pan with default options.

ahocevar (approved):

PR #13840, opened 2022-07-18
No string enums and type safety for observable properties and event types

ahocevar:
This pull request aims to remove string enums for event types, and add type safety for observable properties and event types by adding union type generics (defaulting to `string`) to Event, Observable and Object.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "handlebars": "4.7.7",
         "jquery": "3.6.0",
         "jsdoc": "3.6.10",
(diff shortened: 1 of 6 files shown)


ahocevar:
API doc generation currently fails because JSDoc does not recognize dynamic types like `change:${ObservableProperties}`. This could be fixed by converting such types in `jsdoc-plugin-typescript`.

ahocevar:
Closing this for now. Let's see where the string enum removal leads to.

I'd really like to see our event handling work without anything special (event type enums, JSDoc plugin), and ideally we could even get rid of `Object` and observable properties.

PR #13858, opened 2022-07-19
Replace enums with typedef

MoonE:
This is a subset of #13836 that is ready to be reviewed.


ahocevar (approved):
Thanks for this effort, @MoonE

PR #13863, opened 2022-07-21
Fix reloading tiles in case of an error with tile.load()

ahocevar:
This has gone unnoticed for a long time, and maybe it never really worked. But the idea of calling `tile.load()` - among other things - was to be able to reload the tile. Something like
```js
const retryCodes = [408, 429, 500, 502, 503, 504];
const retries = {};
source.setTileLoadFunction((tile, src) => {
  const image = tile.getImage();
  fetch(src)
    .then((response) => {
      if (retryCodes.includes(response.status)) {
        retries[src] = (retries[src] || 0) + 1;
        if (retries[src] <= 3) {
          setTimeout(() => tile.load(), retries[src] * 1000);
        }
        return Promise.reject();
      }
      return response.blob();
    })
    .then((blob) => {
      const imageUrl = URL.createObjectURL(blob);
      image.src = imageUrl;
      setTimeout(() => URL.revokeObjectURL(imageUrl), 5000);
    })
    .catch(() => tile.setState(3 /* error */));
});
```
This pull request fixes that. To make this work consistently, I also removed code from the renderer that modifies the tile's state when `useInterimTilesOnError` is set to false. This should not be needed, because we only consider an error tile drawable when `useInterimTilesOnError` is `false`: 
https://github.com/openlayers/openlayers/blob/1c68d5093d10ebbe0ced545b8bb30a0b1e4063a5/src/ol/renderer/canvas/TileLayer.js#L107-L116

### DIFF
--- a/src/ol/ImageTile.js
+++ b/src/ol/ImageTile.js
@@ -112,7 +112,40 @@ class ImageTile extends Tile {
   }
 
   /**
-   * Load not yet loaded URI.
+   * Load the image or retry if loading previously failed.
+   * Loading is taken care of by the tile queue, and calling this method is
+   * only needed for preloading or for reloading in case of an error.
+   *
+   * To retry loading tiles on failed requests, use a custom `tileLoadFunction`
+   * that checks for error status codes and reloads only when the status code is
+   * 408, 429, 500, 502, 503 and 504, and only when not too many retries have been
+   * made already:
+   *
+   * ```js
+   * const retryCodes = [408, 429, 500, 502, 503, 504];
+   * const retries = {};
+   * source.setTileLoadFunction((tile, src) => {
+   *   const image = tile.getImage();
+   *   fetch(src)
+   *     .then((response) => {
+   *       if (retryCodes.includes(response.status)) {
+   *         retries[src] = (retries[src] || 0) + 1;
+   *         if (retries[src] <= 3) {
+   *           setTimeout(() => tile.load(), retries[src] * 1000);
+   *         }
+   *         return Promise.reject();
+   *       }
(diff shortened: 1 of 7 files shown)


ahocevar:
Looking into the failing rendering tests now.

bartvde:
For the snippet, you are missing one evt.tile (it reads tile), so:

```
let retries = {};
source.on('tileloaderror', (evt) => {
  const key = evt.tile.getTileCoord().toString();
  retries[key] = (retries[key] || 0) + 1;
  if (retries[key] <= 3) {
    setTimeout(() => evt.tile.load(), retries * 1000)
  }
});
```

ischneider:
Hi @ahocevar ! This is good stuff though I'm concerned about non-retryable status codes. For example, a 400 implies the client should not retry the request without modification.

A google search suggests that 408, 429, 500, 502, 503 and 504 are retryable.

Additionally, might it be useful to head the advice of any Retry-After header?

ahocevar:
A very valid concern, @ischneider. I'll update the snippet accordingly, and also add that to the API documentation.

ahocevar:
@bartvde I updated the original snippet to address @ischneider's concern. Please try the new snippet - it makes sure we only reload tiles when the response code indicates that a reload might bring a cure.

tschaub (approved):
Any idea if this works with the WebGL tile renderer?  We can fix it later if not, just curious.

ahocevar:
@tschaub I just tried. Yes, this does work with the WebGL tile renderer.

bartvde:
thanks @ahocevar !

PR #13868, opened 2022-07-24
Remove deprecated method PluggableMap#forEachLayerAtPixel ...

MoonE:
... and a few more methods that are no longer used because of this.


mike-000:
In the Interpolation example both maps would return the same values with `getData()` regardless of interpolation if the source was not reprojected. even though the pixelation or interpolation would be visible to the user.  To clarify that `a second map with interpolation enabled` should now be `a second map reprojected with interpolation enabled`.  Reading interpolated values without reprojection would now require an `ol/source/Raster` as in https://deploy-preview-13365--ol-site.netlify.app/examples/webgl-interpolation.html

MoonE:
I do get different values for both maps, but as far as I understand it tiles are created for the current zoom level even if the source does not provide it. `getData` may not return  the actual screen pixel when the map zoom level is not an integer. Not sure why the result would be different with reprojection.

To always get the actual screen pixel value you would indeed have to use `ol/source/Raster`.

tschaub (approved):

mike-000:
> Not sure why the result would be different with reprojection.

In this case `getData()` is using reproj tiles, not the original source tiles.  Interpolation is applied or suppressed during the creation of the tiles.

PR #13877, opened 2022-07-25
Queue reprojection source tiles

### LINKED ISSUE #13866: Reprojection example can overload WMS server
**Describe the bug**
Reprojecting the EPSG:4326 tile WMS source to EPSG:3857 is slow and may overload the server.

**To Reproduce**
Open https://openlayers.org/en/latest/examples/reprojection.html on a wide screen, select WMS EPSG:4326 as the base map, the equatorial and mid latitude reproj tiles appear quite quickly.  Pan north or south and the rendering gets very slow as the number of source tiles requested increases exponentially with latitude.  Eventually most of the tiles are rendered, but there may be some gaps and load errors on the console.

**Expected behavior**
The example should be reasonably fast and not request more tiles than the server can handle.  Possible solutions might be to restrict the maximum width of the map, to specify a tile grid to limit the zoom level of source tiles, or to use ImageWMS or an alternative cached XYZ source (e.g. https://github.com/openlayers/openlayers/issues/13782#issuecomment-1161720327) which is more able to handle the load - or better queue management.

mike-000:
Fixes #13866

Add a simple global queue for source tiles needed by reproj tiles in the main queue to restrict the number of simultaneous requests on a server.  This is not configurable - although configurable queues could be assigned to individual sources a default queue seems sufficient as simultaneous reproj tiles could be restricted in the main queue by reducing the map's `maxTilesLoading` if necessary.

### DIFF
--- a/src/ol/reproj/Tile.js
+++ b/src/ol/reproj/Tile.js
@@ -5,8 +5,10 @@ import {ERROR_THRESHOLD} from './common.js';
 
 import EventType from '../events/EventType.js';
 import Tile from '../Tile.js';
+import TileQueue from '../TileQueue.js';
 import TileState from '../TileState.js';
 import Triangulation from './Triangulation.js';
+import {VOID} from '../functions.js';
 import {
   calculateSourceExtentResolution,
   canvasPool,
@@ -20,6 +22,10 @@ import {releaseCanvas} from '../dom.js';
 /**
  * @typedef {function(number, number, number, number) : import("../Tile.js").default} FunctionType
  */
+
+const sourceTileQueue = new TileQueue(function () {
+  return 1;
+}, VOID);
 
 /**
  * @classdesc
@@ -298,6 +304,7 @@ class ReprojTile extends Tile {
       this.changed();
 
       let leftToLoad = 0;
+      const maxQueue = 16;
 
       this.sourcesListenerKeys_ = [];
       this.sourceTiles_.forEach(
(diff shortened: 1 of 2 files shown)


ahocevar:
Caution. I don't think this fixes anything. The reprojected tiles that cause source tiles to be requested are controlled by the map's tile queue already. As described in #13866, the number of source tiles per reprojected increase when e.g. reprojecting from EPSG:4326 to EPSG.3857. Introducing an extra tile queue for these cases may reduce the load on the server, but will not improve overall performance much, because way too many pixels need to be handled by the reprojection code. In this case, it is better to configure the source's `reprojectionErrorThreshold` appropriately.

mike-000:
`reprojectionErrorThreshold` can improve performance in extreme cases - if zooming in to the northern tip of Greenland with the default `reprojectionErrorThreshold` the number of source tile per reproj tile rises to 48 - with `reprojectionErrorThreshold` set to 2 that reduces to 16.

However, the overload situation observed was not as extreme, just a wide screen with world map 8 reproj tiles wide.  When panned towards the pole the top row requires 16 source tiles each, the row below 12 each.  There will be some sharing within columns, but not in rows as the columns are aligned in both projection.  So allowing for that even with reproj tiles queued that is almost 200 simultaneous source tile requests.  Changing `reprojectionErrorThreshold` to 2 does not reduce that at all.

As the test which broke demonstrates reproj tiles can handle almost 8000 source tiles each when there is no server involved.  Queueing the tiles in that very extreme situation did increase the render time by several seconds.

ahocevar:
Thanks for your work on this and the investigation, @mike-000. Curious to hear what others think, but I'm not in favor of this pull request. The problem it aims to solve with added complexity is a corner case resulting from poor configuration and a server that was not made to handle high loads. As an alternative, I created #13880 to improve the configuration of the WMS source.

mike-000:
Closing in favour of #13880

In less specific cases where all else fails a second queue can always be avoided by using `setTileGridForProjection` so the main queue manages smaller reproj tiles.

PR #13879, opened 2022-07-26
Codesandbox and bundler improvements for examples

### LINKED ISSUE #13865: Examples using proj4 do not load correctly in CodeSandbox
**Describe the bug**
Most examples which use proj4 have `"proj4": "^2.7.5"` in the package.json.  The `^` is not understood by CodeSandbox and must be edited out before the example will load.

In some cases (e.g. Print to Scale) proj4 is completely missing.

**To Reproduce**
Open an example in CodeSandbox and see error.

**Expected behavior**
Examples should load without needing edits.

ahocevar:
Fixes #13865.

And use vite instead of parcel.

### DIFF
--- a/examples/webpack/example-builder.js
+++ b/examples/webpack/example-builder.js
@@ -146,6 +146,8 @@ function getDependencies(jsSource, pkg) {
       const dep = imp.startsWith('@') ? parts.slice(0, 2).join('/') : parts[0];
       if (dep in pkg.devDependencies) {
         dependencies[dep] = pkg.devDependencies[dep];
+      } else if (dep in pkg.dependencies) {
+        dependencies[dep] = pkg.dependencies[dep];
       }
     }
   }
--- a/examples/webpack/example-builder.js
+++ b/examples/webpack/example-builder.js
@@ -13,7 +13,7 @@ const baseDir = dirname(fileURLToPath(import.meta.url));
 
 const isCssRegEx = /\.css(\?.*)?$/;
 const isJsRegEx = /\.js(\?.*)?$/;
-const importRegEx = /(?:^|\n)import .* from '(.*)';(?:\n|$)/g;
+const importRegEx = /\simport .* from '([^']+)';\s/g;
 const isTemplateJs =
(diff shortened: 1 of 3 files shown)


jipexu:
@ahocevar  hi sorry for this example  https://deploy-preview-13879--ol-site.netlify.app/examples/d3.html  is still not work (and with some errors this time  )?
Error
Could not fetch dependencies, please try again in a couple seconds:

examples/webpack/example-builder.js
```diff
@@ -13,7 +13,7 @@ const baseDir = dirname(fileURLToPath(import.meta.url));
 
 const isCssRegEx = /\.css(\?.*)?$/;
 const isJsRegEx = /\.js(\?.*)?$/;
-const importRegEx = /(?:^|\n)import .* from '(.*)';(?:\n|$)/g;
+const importRegEx = /\simport .* from '([^']+)';\s/g;
```
MoonE, line 16: This new regex does not match an import in the first or last line of a file
```suggestion
const importRegEx = /\bimport [^']+ from '([^']+)'/g;
```

As a test:
```js
function match(re, text) {
  let imports = [];
  let m;
  while ((m = re.exec(text))) {
    imports.push(m[1])
  }
  return imports;
}
match(
 	/\bimport [^']+ from '([^']+)'/g,
  `import assert from 'assert';import frontMatter from 'front-matter';
import fs from 'fs'
import fse from 'fs-extra';`
)
```

ahocevar:
@jipexu You need to set the OpenLayers version to one that is actually published. But note that this pull request does not address the problems with the d3 example, as the issue it solves is not about the s3 example.

jipexu:
@ahocevar ah ok thanks

examples/webpack/example-builder.js
MoonE: Did you forget to push the fix?

ahocevar: Indeed, pushed now. Thanks for the reminder. I used a slightly different regex than the one you suggested, but it passes your test.

MoonE (approved):
Thanks @ahocevar

PR #13880, opened 2022-07-26
Improve reprojection WMS config for better performance

### LINKED ISSUE #13866: Reprojection example can overload WMS server

ahocevar:
Fixes #13866
Closes #13877

Other than #13877, this pull request configures the WMS in the example with an appropriate tile size and reprojection error threshold, instead of introducing the complexity and problems of a second tile queue next to the existing queue.

### DIFF
--- a/examples/reprojection.js
+++ b/examples/reprojection.js
@@ -6,6 +6,7 @@ import WMTS, {optionsFromCapabilities} from '../src/ol/source/WMTS.js';
 import WMTSCapabilities from '../src/ol/format/WMTSCapabilities.js';
 import proj4 from 'proj4';
 import {OSM, TileImage, TileWMS} from '../src/ol/source.js';
+import {createXYZ} from '../src/ol/tilegrid.js';
 import {getCenter, getWidth} from '../src/ol/extent.js';
 import {get as getProjection, transformExtent} from '../src/ol/proj.js';
 import {register} from '../src/ol/proj/proj4.js';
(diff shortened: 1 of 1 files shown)


mike-000:
There are still some 500 errors on a wide screen.  An additional problem is the tilegrid is based on EPSG:3857.  It must be aligned with the EPSG:4326 extent otherwise the WMS requests at the edges will be invalid, as can be seen when viewed without reprojection
image

ahocevar:
@mike-000 Thanks, tile grid configuration update.

mike-000:
The number of source tiles per reproj tile is now no more than 4, but server is still overloading at times.

Using smaller reproj tiles managed by the main queue looks promising, no more than 2 source tiles per reproj tile, and many of those will already be cached

`layers['wms4326'].getSource().setTileGridForProjection('EPSG:3857', createXYZ({tileSize: 128}));`

ahocevar:
I run that server. It has a 2-minute coffee break every full hour, because I have no interest in it being used for anything but demo purposes. I'm not going to change to such a small tile size. Instead, I changed it back to the default of 256, making use of GeoWebCache to reduce the server load.

tschaub (approved):

Issue #7679, opened 2018-01-05
Should feature.clone() clone the style too?


raiyni:
It can create weird situations when you are  modifying the cloned feature for some reason. Doesn't seem intuitive if you don't already know it doesn't clone the style.

https://codepen.io/anon/pen/opojBB

```
var f2 = f1.clone();


f2.getStyle().setImage(new ol.style.Icon({
          color: '#ff0000',
          crossOrigin: 'anonymous',
          src: 'https://openlayers.org/en/v4.6.4/examples/data/dot.png'
        }));
```

See how both are now red

cs09g:
I think it shouldn't be sync with target and cloned feature. its behavior seems like Object.assign(target) instead Object.assign({}, target) though I didn't check the code yet.

simonseyock:
If feature.clone would clone the style, it would be more in line with the rest of the code (the geometry is cloned also and style.clone is deep, too). But on the other hand styles are meant to be shared between features. We should clarify the documentation either way.

Cavva79:
Would be nice to be able to choose with a parameter to clone styles and eventually parameters too.

PR #13882, opened 2022-07-26
Propagate style changes to features and layers

tschaub:
Although we have methods to modify styles, we also require that application developers manually call `layer.changed()` to get the map to render after a style is modified.  Instead, changes to styles should propagate to their parent layer or feature.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,10 @@
 
 ### Next version
 
+#### Style changes propagate to layers and features
+
+It is no longer necessary to call `layer.changed()` or `feature.changed()` if you modify the layer or feature's style with one of the setter methods.  Vector layers with a style or features with a style will now automatically dispatch change events when those styles are modified (resulting in a map render).
+
 ### 6.15.0
 
 #### Deprecated `tilePixelRatio` option for data tile sources.
--- a/test/browser/spec/ol/graticule.test.js
+++ b/test/browser/spec/ol/graticule.test.js
@@ -187,8 +187,6 @@ describe('ol.layer.Graticule', function () {
       graticule.createGraticule_(extent, [0, 0], resolution, squaredTolerance);
       expect(graticule.meridiansLabels_[0].text).to.be('lon: 0');
       expect(graticule.parallelsLabels_[0].text).to.be('lat: 0');
-      expect(graticule.lonLabelStyle_(feature).getText()).to.eql(lonLabelStyle);
-      expect(graticule.latLabelStyle_(feature).getText()).to.eql(latLabelStyle);
       feature.set('graticule_label', graticule.meridiansLabels_[0].text);
       expect(graticule.lonLabelStyle_(feature).getText().getText()).to.be(
         'lon: 0'
--- a/test/node/ol/Feature.test.js
+++ b/test/node/ol/Feature.test.js
@@ -3,6 +3,7 @@ import Point from '../../../src/ol/geom/Point.js';
 import Style from '../../../src/ol/style/Style.js';
 import expect from '../expect.js';
 import sinon from 'sinon';
+import {Stroke} from '../../../src/ol/style.js';
 import {isEmpty} from '../../../src/ol/obj.js';
 
 describe('ol/Feature.js', function () {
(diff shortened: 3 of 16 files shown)


test/node/ol/Feature.test.js
```diff
@@ -379,7 +391,8 @@ describe('ol/Feature.js', function () {
       const coordinates = geometryClone.getFlatCoordinates();
       expect(coordinates[0]).to.be(1);
       expect(coordinates[1]).to.be(2);
-      expect(clone.getStyle()).to.be(style);
+      expect(clone.getStyle()).to.be.a(Style);
+      expect(clone.getStyle()).not.to.be(style);
```
tschaub, line 395: Previously, cloned features shared a style.  I think this was a bug.

test/browser/spec/ol/graticule.test.js
```diff
@@ -187,8 +187,6 @@ describe('ol.layer.Graticule', function () {
       graticule.createGraticule_(extent, [0, 0], resolution, squaredTolerance);
       expect(graticule.meridiansLabels_[0].text).to.be('lon: 0');
       expect(graticule.parallelsLabels_[0].text).to.be('lat: 0');
-      expect(graticule.lonLabelStyle_(feature).getText()).to.eql(lonLabelStyle);
-      expect(graticule.latLabelStyle_(feature).getText()).to.eql(latLabelStyle);
```
tschaub, line 191: This was asserting that cloned features shared style instances.  This is no longer the case.  A cloned feature will get a cloned style.

test/node/ol/Feature.test.js
mike-000: That was raised in https://github.com/openlayers/openlayers/issues/7679#issuecomment-355902011  `styles are meant to be shared between features` is valid, so maybe it should be an option as suggested in the reply?

ahocevar (commented):
Although I like the convenience that this adds, I fear the overhead that comes with it: When using style functions, it is common practice to reuse style objects and call their property setters to change a value before returning the style. This is e.g. done very frequently on the style functions that ol-mapbox-style creates, and many applications out there do that too. It used to be less overhead than creating new style objects with every style function call.

I did not measure performance, but e.g. in https://deploy-preview-13882--ol-site.netlify.app/examples/mapbox-style.html I notice a significantly longer freeze when trying to pan again after zooming than in https://openlayers.org/en/latest/examples/mapbox-style.html.

If we make the change suggested here, it should not come without a change that styles returned by a style function can be plain object literals, which is easy for polygon and line symbolizers, but challenging for point symbolizers because of the rendering work that currently happens in these styles. Or something else that allows bypassing the Observable overhead.

ahocevar:
One suggestion for an alternative way to achieve what you want to achieve in this pull request (since it only addresses the case where a `Style` or and `Array<Style>` is configured as style on the feature or layer, and not the case of style functions): what if a `Style` (and `Fill`, `Stroke` etc) had a `trackChanges()` method which would make them observable, instead of creating them as `Observable` subclass? That way, `Feature` and `BaseVector`'s `setStyle()` methods could make styles and style arrays observable, and style functions would not suffer from the overhead described in my previous comment.

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,10 @@
 
 ### Next version
 
+#### Style changes propagate to layers and features
+
+It is no longer necessary to call `layer.changed()` or `feature.changed()` if you modify the layer or feature's style with one of the setter methods.  Vector layers with a style or features with a style will now automatically dispatch change events when those styles are modified (resulting in a map render).
```
ahocevar, line 7: Maybe also mention that this is not the case for style functions.

tschaub:
A rough benchmark shows that in the `main` branch, you consume about half the frame budget by by modifying style symbolizers a few million times.  In this branch, you blow half the frame budget doing the same about ten thousand times.  In cases where there are thousands of features and tens of modifications may be made to symbolizers on zoom level changes (for example), this is a problem.

It feels once again that working toward flat style objects (or style expressions) is the way to go.  And that calling `layer.setStyle(newStyle)` should be the only API method for changing a style (instead of having mutable nested symbolizers).

Issue #13883, opened 2022-07-26
Revert IE workarounds


tschaub:
For the next major release, we could remove support for IE.  There are years of IE workarounds in the library, so we may not get them all out, but we could remove some of the obvious ones.

I'll use this ticket to link to candidate reverts:

 - [x] #12734
 - [x]  #10989
 - [x]  #10824
 - [x]  #10612
 - [x]  #12891
 - [x]  #12785
 - [x]  #12740
 - [x]  #11865

mike-000:
Plus other workarounds in ol/math and ol/obj, and any references to `msSaveBlob`, `msToBlob` and `msImageSmoothingEnabled`.

tschaub:
We’ll likely find more, but I’ll close this one as done.

PR #13884, opened 2022-07-26
Remove workaround for Array.prototype.includes

tschaub:
This removes the `includes` function from `ol/array.js` in favor of Array.prototype.includes.

See #13883.

### DIFF
--- a/src/ol/interaction/Select.js
+++ b/src/ol/interaction/Select.js
@@ -4,12 +4,13 @@
 import Collection from '../Collection.js';
 import CollectionEventType from '../CollectionEventType.js';
 import Event from '../events/Event.js';
+import Feature from '../Feature.js';
 import Interaction from './Interaction.js';
 import VectorLayer from '../layer/Vector.js';
 import {TRUE} from '../functions.js';
 import {clear} from '../obj.js';
 import {createEditingStyle} from '../style/Style.js';
-import {extend, includes} from '../array.js';
+import {extend} from '../array.js';
 import {getUid} from '../util.js';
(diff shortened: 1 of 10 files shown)


src/ol/interaction/Select.js
```diff
@@ -526,12 +526,23 @@ class Select extends Interaction {
          */
         function (feature, layer) {
           if (this.filter_(feature, layer)) {
-            if ((add || toggle) && !includes(features.getArray(), feature)) {
+            if (
+              (add || toggle) &&
+              !features
+                .getArray()
+                .includes(
+                  /** @type {import("../Feature.js").default} */ (feature)
```
MoonE, line 534: Isn't this showing a problem with the type returned by `getFeatures`, and its return type should be `Array<FeatureLike>`?

mike-000:
`find` and `findIndex` in that module are also IE workarounds.  However some caution is needed as `ol/array#find` returns null where `Array.prototype.find` returns undefined.

src/ol/interaction/Select.js
tschaub: Currently digging in that can of worms.  Not clear to me where this interaction actually works with render features.  There are parts that assume a full fledged feature.

tschaub: I think that ae87cad7ef65dbc8b7cf490809e17ee891961a4b added a few too many `FeatureLike` types.  The select interaction only really works with `Feature` instances.  I pushed a new commit that makes this clear.

ahocevar (approved):

PR #13887, opened 2022-07-27
Remove find and findIndex from array module

tschaub:
This removes the `find` and `findIndex` functions from `ol/array.js` in favor of Array.prototype.find and Array.prototype.findIndex.

See #13883.


MoonE (approved):
Thanks @tschaub

PR #13889, opened 2022-07-27
Use Object.values

tschaub:
This removes the `getValues` function from `ol/obj.js` in favor of `Object.values()`.

See #13883.


MoonE (approved):
Thanks @tschaub

PR #13890, opened 2022-07-27
Use Math.cosh and Math.log2

tschaub:
This removes the `cosh` and `log2` functions from the `ol/math.js` module.

See #13883.


MoonE (approved):
Thanks @tschaub

PR #13891, opened 2022-07-27
Publish untranspiled sources

tschaub:
This removes the transpilation step prior to publishing the `ol` package.  Previously, we used `tsc` to transpile module sources with the [`es5` target](https://www.typescriptlang.org/tsconfig#target).  This was largely about transpiling away the `class` keyword for IE.  This branch makes it so we publish the sources for the modules as authored instead.

See #13883.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,19 +80,15 @@ OpenLayers appreciates contributions of all kinds.  We especially want to thank 
 
 See our [GitHub sponsors page](https://github.com/sponsors/openlayers) or [Open Collective](https://opencollective.com/openlayers/contribute/sponsors-214/checkout) if you too are interested in becoming a regular sponsor.
 
-## ES Modules
-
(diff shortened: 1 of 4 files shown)


tschaub:
I've published a version of the package with this branch.  If anybody wants to try it out:

```
npm install ol@7.0.0-pr.13891.1
```

This change removes 5 MB from the installed package size.  The package tarball is 917 KB as opposed to 1.6 MB.

ahocevar (approved):
Looks good, but I seem to remember we mention the separate src/ folder in the package somewhere in the README or the docs. That should be removed as well, but this can also be done in a separate pull request.

tschaub:
> Looks good, but I seem to remember we mention the separate src/ folder in the package somewhere in the README or the docs.

I was looking for the same but couldn't find it.  I'll dig some more.

tschaub:
Oh, I guess I didn't look very hard.

PR #13892, opened 2022-07-27
Remove unused worker

tschaub:
The `ol/worker/version.js` module was meant as a demonstration in #9550.  It isn't meant to actually be used and can be removed.


MoonE (approved):
Thanks @tschaub

PR #13897, opened 2022-07-29
Align graticule default intervals to minutes and seconds

mike-000:
A follow on from #13886 and #13893   If formatting whole minutes it makes little sense to use intervals such as 36 seconds but not 1 minute, or to use fractional values such as 7.2 seconds.  Instead use the same 30, 20, 10, 5, 2, 1 interval range for degrees, minutes and seconds.


tschaub (approved):
Looks great.  Thanks, @mike-000.

PR #13899, opened 2022-07-29
Wait until first load to create icon image

tschaub:
This shifts image creation from the constructor to the load method.

Extracted from #12873 and will be used in testing upcoming style changes.


MoonE:
Will this return `undefined` when getImage is called when it is not loading or loaded? Should it also be initialized in the `getImage` method?

tschaub:
> Will this return `undefined` when getImage is called when it is not loading or loaded? Should it also be initialized in the `getImage` method?

Good point.  I initialized the image in a couple more places.

MoonE (approved):
Thanks @tschaub

Issue #13896, opened 2022-07-29
New vector style API


tschaub:
The current vector layer styling API is awkward in that style and symbolizer instances are mutable, but changes to them don’t result in a layer being re-rendered (see also #13882).  In addition, the current API limits us in terms of efficiently evaluating styles and building rendering instructions (hard to pass all the context to a worker, not practical to do in a shader).  That and the style and nested symbolizer constructors feel pretty verbose (see also #12873).

Updating a symbolizer property (and getting a layer to re-render) currently looks like this:

```js
style.getText().getStroke().setWidth(3);
layer.changed();
```

Instead, we could limit the API to just `layer.setStyle()` and the above would look like this:

```js
layer.setStyle({
  …style,
  'text-stroke-width': 3, // property names tbd
});
```

To support this, the `layer.setStyle()` function would accept a new style literal.

In addition to values like `3` for `text-stroke-width`, we could also support expressions for property values.  So you could vary the text stroke width based on zoom with something like this:

```js
layer.setStyle({
  ...style,
  'text-stroke-width': ['interpolate', ['linear'], ['zoom'], 3, 1, 10, 5],
});
```

Expressions like this would take the place of the current style functions.  As a result, we would have more control over where styles were parsed and evaluated with feature attributes (in a worker, in a shader).

Even before we add support for expressions (eventually replacing style functions), I think the object literal syntax and simplified API will be easier for users.

I can imagine getting there from here in a few milestones:

 1. Accept style literals (and arrays of the same) in `layer.setStyle()` calls.  These will be transformed into style and symbolizer instances by the layer, so no renderer changes needed.
 2. Rework the layer/renderer to handle style object literals directly (without needing to create style and symbolizer instances).  At this point, style functions could return style object literals and the style and symbolizer constructors could be removed.
 3. Add support for expressions in style object literals.  Remove support for style functions.

 I've started experimenting with a style object literal syntax in a [`flat-style` branch](https://github.com/openlayers/openlayers/compare/main...tschaub:openlayers:flat-style).  When that is working, I'll open a draft pull request and we can discuss the property names (`text-stroke-width` vs `textStrokeWidth` etc.) or other parts of the syntax.

ajkopinga:
[...]

PR #13900, opened 2022-07-29
Accept an object literal for static layer styling

tschaub:
This change updates the `layer.setStyle()` method to accept an object literal or an array of object literals for layer styling.  This is an alternative to using style and symbolizer constructors for static styles.

I've converted over a number of the examples and rendering tests that were using static styling to use the new object literal syntax.

This is part 1 of changes to the styling API (see #13896).

### DIFF
--- a/config/jsdoc/api/conf.json
+++ b/config/jsdoc/api/conf.json
@@ -14,6 +14,7 @@
     ]
   },
   "plugins": [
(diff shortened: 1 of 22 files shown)


ahocevar (approved):
Beautiful! Thanks for your effort on this, @tschaub.

PR #13902, opened 2022-07-29
Improve Collection type-safety

MoonE:
The `Collection` will no longer fire `remove` events when the collection remains unchanged
It will throw when attempting to insert values at index larger than the current length and it. I don't think this mechanic was very useful and it was not used in the library itself.

`CollectionEvent` will now inherit the Collection's type via generics.
The `CollectionEvent` arguments are no longer optional. The index was always set anyways and if the element  may be undefined the `Collection` should be of a type that allows this.

The `ol/source/Vector.Options` object is now templated and inherits the source's type.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -14,6 +14,10 @@ The `toStringHDMS` function from the `ol/coordinate.js` module now formats longi
 
 The default intervals now align with integer minutes and seconds better suited to the default label formatter.  If formatting in decimal degrees you may wish to specify custom intervals suited to that format.
 
+#### ol/Collection
+
+Inserting with `setAt` or `insertAt` beyond the current length used to create a sparse Collection with `undefined` inserted for any missing indexes.  This will now throw an error instead.
+
 ### 6.15.0
 
 #### Deprecated `tilePixelRatio` option for data tile sources.
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -254,6 +254,9 @@ class Collection extends BaseObject {
    * @api
    */
   removeAt(index) {
+    if (index < 0 || index >= this.getLength()) {
+      return undefined;
+    }
     const prev = this.array_[index];
     this.array_.splice(index, 1);
     this.updateLength_();
--- a/src/ol/Collection.js
+++ b/src/ol/Collection.js
@@ -22,10 +22,10 @@ const Property = {
 export class CollectionEvent extends Event {
   /**
    * @param {import("./CollectionEventType.js").default} type Type.
-   * @param {*} [opt_element] Element.
-   * @param {number} [opt_index] The index of the added or removed element.
+   * @param {*} element Element.
+   * @param {number} index The index of the added or removed element.
    */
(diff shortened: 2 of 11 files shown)


src/ol/Collection.js
```diff
@@ -278,14 +288,18 @@ class Collection extends BaseObject {
...
+        /** @type {CollectionEvent<T>} */ (
+          new CollectionEvent(CollectionEventType.ADD, elem, index)
+        )
       );
     } else {
-      for (let j = n; j < index; ++j) {
-        this.insertAt(j, undefined);
+      if (index > n) {
+        throw new Error('Index out of bounds: ' + index);
       }
```
tschaub, line 303: Maybe worth mentioning in the upgrade notes that you can no longer insert elements at index values beyond the length of the collection (leaving a sparse collection).

tschaub (approved):

changelog/upgrade-notes.md
```diff
@@ -14,6 +14,10 @@ The `toStringHDMS` function from the `ol/coordinate.js` module now formats longi
 
 The default intervals now align with integer minutes and seconds better suited to the default label formatter.  If formatting in decimal degrees you may wish to specify custom intervals suited to that format.
 
+#### ol/Collection
+
+Inserting with`setAt` or `insertAt` beyond the current length used to create a sparse Collection with `undefined` inserted for any missing indexes.  This will now throw an error instead.
```
tschaub, line 19:
```suggestion
Inserting with `setAt` or `insertAt` beyond the current length used to create a sparse Collection with `undefined` inserted for any missing indexes.  This will now throw an error instead.
```

(missing space)

MoonE: Thanks, fixed it.

bartvde:
We are now getting

```
TS2322: Type '(Feature<Point> | Feature<LineString>)[]' is not assignable to type 'Feature<Point>[] | Collection<Feature<Point>> | undefined'.
  Type '(Feature<Point> | Feature<LineString>)[]' is not assignable to type 'Feature<Point>[]'.
    Type 'Feature<Point> | Feature<LineString>' is not assignable to type 'Feature<Point>'.
      Type 'Feature<LineString>' is not assignable to type 'Feature<Point>'.
        Type 'LineString' is not assignable to type 'Point'.
          The types returned by 'getCoordinates()' are incompatible between these types.
            Type 'Coordinate[]' is not assignable to type 'Coordinate'.
              Type 'Coordinate' is not assignable to type 'number'.
    65 |       });
    66 |       const source = new VectorSource({
  > 67 |         features: showScale ? [lineFeature, arrow1, arrow2, markerFeature] : [markerFeature],
       |         ^^^^^^^^
    68 |       });
    69 |       const layer = new VectorLayer({
    70 |         source,
```

wonder if this related to this change?

MoonE:
Looks like when you combine features with different geometry types you need to explicitly set the source type.
`new VectorSource<Geometry>` or `new VectorSource<Point|LineString>()`

bartvde:
@MoonE thanks for the pointer, will try that out tomorrow

PR #13904, opened 2022-07-30
Fail PointsLayer tests faster and with error message

MoonE:
Currently the error is not reported back to the test runner and it times out on failure. This should report the error immediately.


MoonE:
Actually this is unnecessary, mocha registers a global error listener that should catch any errors.

PR #13905, opened 2022-07-30
Replace enums with typedefs

MoonE:
These are some more changes from #13836
If you prefer the assertion in `ol/Layer/VectorTile` removed I don't mind either way.
```js
assert(renderMode == 'hybrid' || renderMode == 'vector', 28);
```
Should the typedefs be  tagges wiht `@api`, they seem to show up in the docs  already without it?


tschaub:
> Should the typedefs be tagges wiht @api, they seem to show up in the docs already without it?

In #13900, I found that I needed to add a single `@api` annotation so that documentation for the `ol/style/flat.js` module would get included in the docs.  Maybe typedefs only get documented if the module has at least one other `@api` annotation.

tschaub (approved):

PR #13906, opened 2022-07-30
Re-export the link interaction from the interactions module

tschaub:
This makes it possible to `import {Link} from 'ol/interactions.js';`.


MoonE (approved):
Thanks @tschaub

PR #13908, opened 2022-07-30
Uniformly grayscale controls

tschaub:
Following up on https://github.com/openlayers/openlayers/pull/9013#issuecomment-442613882, this updates the style for controls to be uniformly grayscale.

image

### DIFF
--- a/src/ol/ol.css
+++ b/src/ol/ol.css
@@ -1,8 +1,16 @@
+:root {
+  --ol-background-color: white;
+  --ol-foreground-color: #333333;
+  --ol-subtle-foreground-color: #666666;
+  --ol-subtle-background-color: rgba(128, 128, 128, 0.25);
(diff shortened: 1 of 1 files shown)


MoonE:
These css variables should be prefixed with `ol-` to avoid any name clashes.

tschaub:
> These css variables should be prefixed with `ol-` to avoid any name clashes.

Good point.  Done.

MoonE:
The controls kind of disappear against a white background.
image

Maybe use a darker color for the subtle background?
`--ol-subtle-background-color: rgba(128, 128, 128, .25);`
image

tschaub:
> The controls kind of disappear against a white background.

Agreed.  I darkened the background.  I also reduced the margin on the controls so the background isn't as big.

image

I think this looks tighter than the 2px background.

ahocevar (approved):
So beautiful!

PR #13912, opened 2022-07-31
Remove IconImage color fallback for IE

MoonE:
All current browsers should support `globalCompositeOperation` `multiply` and `destination-in`.


tschaub (approved):
Thanks, @MoonE.

See #13883.

PR #13913, opened 2022-07-31
Use a ResizeObserver on the map target to update its size

MoonE:
This makes it unnecessary to call `updateSize` in most cases. It should now only be needed when `setSize` was used before to set it to something else.

[Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#browser_compatibility) is [ok](https://browserslist.dev/?q=Q2hyb21lIDwgNjQsIEVkZ2UgPCA3OSwgRmlyZWZveCA8IDY5LCBPcGVyYSA8IDUxLCBTYWZhcmkgPCAxMy4xLCBhbmRfY2hyIDwgNjQsIGFuZF9mZiA8IDc5LCBvcF9tb2IgPCA0NywgaW9zX3NhZiA8IDEzLjQsIHNhbXN1bmcgPCA5LjA%3D), Safari was the last to support it with 13.1 and 13.4 on iOS in early 2020.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -26,6 +26,10 @@ The control will now by default keep displaying the last mouse position when the
 
 The `PluggableMap` class has been removed.  If you want to create a custom map class, extend the `Map` class instead.
 
+#### ol/Map
+
+A [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) is added to the map target element and therefore `updateSize` no longer needs to be called when the map's target has been resized.
+
 ### 6.15.0
 
 #### Deprecated `tilePixelRatio` option for data tile sources.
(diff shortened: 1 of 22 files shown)


changelog/upgrade-notes.md
```diff
@@ -22,6 +22,10 @@ Inserting with `setAt` or `insertAt` beyond the current length used to create a
 
 The control will now by default keep displaying the last mouse position when the mouse leaves the viewport.  With `placeholder: '&#160;'` you can keep the old behaviour.  The `placeholder` option no longer accepts `false` as a valid value, instead simply omit the option.  The `undefinedHTML` option has been removed. You should use `placeholder` instead.
 
+#### ol/Map and ol/PluggableMap
```
tschaub, line 25:
```suggestion
#### ol/Map
```

PR #13922, opened 2022-08-01
Remove reference to ES6 as ES2017 already listed

mike-000:
No need for specific version for GeoTIFF is a later version is already given.


tschaub (approved):

tschaub:
Thanks, @mike-000

PR #13932, opened 2022-08-03
Remove input type="range" `change` event listeners where `input` events are also used

mike-000:
These were only needed to support IE


ahocevar (approved):
Thanks, @mike-000

PR #13933, opened 2022-08-03
Remove check for context.setLineDash

MoonE:
IE 11 actually already supported this, just not IE 10.

I also added a few null type annotations for line dash.


tschaub (approved):
Looks good

PR #13937, opened 2022-08-04
Use nicer functions, remove old code

MoonE:
Remove the left-over `array.includes` function  #13884
Use `includes` and `startsWith` and `test` functions where possible
Remove some deprecated access of `RegExp` match via `$1` and `$3` properties.
The print-to-scale example worked fine because the attributions were not collapsible.


tschaub (approved):
Looks good, @MoonE

PR #13939, opened 2022-08-05
Render vector tile vectors in correct sequence for the postrender event

### LINKED ISSUE #13938: VectorTileLayer render mode `vector` is rendered after the `postrender` event
**Describe the bug**
With render mode `vector` the `postrender` is firing too soon.

**To Reproduce**
In a layer swipe it is necessary to place the `ctx.restore` in a timeout as in https://codesandbox.io/s/layer-swipe-forked-fdtie7?file=/main.js  Remove the timeout and it does not work.

**Expected behavior**
Possibly related to #11606 but this example only styles polygons - there are no declutter items, so no obvious reason to render after the event

ahocevar:
Fixes #13938

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -617,18 +617,16 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
   }
 
   /**
-   * Render the layer.
+   * Render the vectors for this layer.
+   * @param {CanvasRenderingContext2D} context Target context.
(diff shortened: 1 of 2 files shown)


ahocevar:
I now added a test to ensure the correct behavior. Without the fix, the sequence of context calls and events is
```js
[
  'prerender', 'clearRect', 'postrender',
  'save', 'beginPath', 'moveTo',  'lineTo', 'lineTo', 'lineTo', 'clip', 'restore'
] 
```
With the fix, `postrender` comes last, as expected.

tschaub (approved):

PR #13941, opened 2022-08-06
Fix KML default icon anchor with https icon url

MoonE:
Standard icons were not anchored correctly when the .icon url was with https.


ahocevar (approved):
Thanks, @MoonE

tschaub (approved):

PR #13943, opened 2022-08-06
Scaleline

MoonE:
This add some improvements to the ScaleLine class and its example page.

- ScaleLine is now configured with `pointer-events: none;` so it is now possible to e. g. drag the map at the position of the ScaleLine
- For very high zoom levels the scale bar now displays `step : 1` instead of `1 : step` (step always >= 1)
- The scale bar has unnecessary nesting div containing `scaleBarText` and `scaleSteps` removed
- The example has a new option to allow inverting the bar colors and now uses a range input for the number of steps

### DIFF
--- a/src/ol/control/ScaleLine.js
+++ b/src/ol/control/ScaleLine.js
@@ -334,12 +334,9 @@ class ScaleLine extends Control {
       previousDecimalCount = decimalCount;
       ++i;
     }
-    let html;
-    if (this.scaleBar_) {
-      html = this.createScaleBar(width, count, suffix);
-    } else {
-      html = count.toFixed(decimalCount < 0 ? -decimalCount : 0) + ' ' + suffix;
-    }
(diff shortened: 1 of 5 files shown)


src/ol/control/ScaleLine.js
```diff
@@ -500,7 +460,7 @@ class ScaleLine extends Control {
     );
     const dpi = this.dpi_ || DEFAULT_DPI;
     const inchesPerMeter = 1000 / 25.4;
-    return parseFloat(resolution.toString()) * inchesPerMeter * dpi;
```
MoonE, line 503: If this did something useful I don't understand it.
It seems to convert the resolution from number to string to number?

tschaub: I agree it is not useful (oversight in review of #9013).

tschaub (approved):
Thanks, @MoonE

PR #13944, opened 2022-08-07
Fix WebGL points layer flaky test

jahow:
Added a `unByKey` call to unregister an event listener in the `rendercomplete` test.


ahocevar (approved):
Thanks, @jahow

PR #13955, opened 2022-08-08
Do not apply #12467 change to Icons

### LINKED ISSUE #13954: ol.style.Icon doubles when using vector Context.drawFeature
I was trying to make a ripple animation around a point with style.icon applied on the point. In order for my pulsation to be under the icon, I draw the pulsations as it should in postrender first and then overlay the icon on top. But my icon, instead of being the original view, is slightly reduced and shifted relative to the center of the point, from which we see 2 icons

1. Go to http://jsfiddle.net/upm2sxbq/20/
2. You will see how the icon shrinks and shifts relative to the center

I expected that vectorContext.drawFeature would just overlay its icon on top and everything would look acceptable.

mike-000:
Fixes #13954

Also fixed the rendering test which might have picked this up, but was not using the correct pixel ratio.

Compare the fixed version https://deploy-preview-13955--ol-site.netlify.app/examples/icon-scale.html
with https://openlayers.org/en/v6.15.1/examples/icon-scale.html
on a mobile device or with browser zoom

### DIFF
--- a/src/ol/render/canvas/Immediate.js
+++ b/src/ol/render/canvas/Immediate.js
@@ -1041,19 +1041,24 @@ class CanvasImmediateRenderer extends VectorContext {
       this.image_ = null;
       return;
     }
+    const imagePixelRatio = imageStyle.getPixelRatio(this.pixelRatio_);
     const imageAnchor = imageStyle.getAnchor();
     const imageOrigin = imageStyle.getOrigin();
     this.image_ = imageStyle.getImage(this.pixelRatio_);
-    this.imageAnchorX_ = imageAnchor[0] * this.pixelRatio_;
-    this.imageAnchorY_ = imageAnchor[1] * this.pixelRatio_;
-    this.imageHeight_ = imageSize[1] * this.pixelRatio_;
+    this.imageAnchorX_ = imageAnchor[0] * imagePixelRatio;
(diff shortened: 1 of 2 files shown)


src/ol/render/canvas/Immediate.js
```diff
@@ -1041,19 +1042,29 @@ class CanvasImmediateRenderer extends VectorContext {
...
       return;
     }
+    let pixelRatio = this.pixelRatio_;
+    const imageScale = imageStyle.getScaleArray();
+    if (imageStyle instanceof IconStyle) {
+      this.imageScale_ = [
+        pixelRatio * imageScale[0],
+        pixelRatio * imageScale[1],
+      ];
+      pixelRatio = 1;
```
ahocevar, line 1052: Not sure if this is the correct fix. It would be batter if icons would be created with the pixel ratio of the destination context, then they would also appear crisper on high dpi devices.

mike-000: Yes, images with `color` applied are available at higher dpi.  The ImageBuilder uses the image style's `getPixelRatio()` method to determine what is available, so that can also be used here.

mike-000: Updated to use high dpi images where available.

ahocevar (approved):
Much better now. Thanks, @mike-000

PR #13916, opened 2022-07-31
Remove Google Analytics and cookie consent

tschaub:
This removes the Google Analytics tracking.

PR #13956, opened 2022-08-08
Fix js error in api pages on load

MoonE:
The `fa` variable was moved out of scope in # #13916


ahocevar (approved):

tschaub:
Thanks for fixing this up, @MoonE.

PR #13957, opened 2022-08-08
replace VERSION in un-transpiled source

MoonE:
The regex no longer matched because in the un-transpiled sources the declaration is with `const` instead of `var`.
See #13891


tschaub (approved):
Thank you

PR #13963, opened 2022-08-09
Remove polyfills for IE and Android 4

tschaub:
The core-js polyfill was added for IE in #12891.

Removing polyfills for `fetch`, `requestAnimationFrame`, `Element.prototype.classList`, and `TextDecoder` removes support for Android 4.  With ~0.4% usage, I think this is acceptable.

I'm leaving in the pointer events polyfill.  It looks like [about 2%](https://caniuse.com/?search=pointer%20events) of browsers could use that.  Maybe 1% should be the cut off.

See #13883.

### DIFF
--- a/examples/templates/example.html
+++ b/examples/templates/example.html
@@ -11,8 +11,6 @@
     <link rel="stylesheet" href="./css/ol.css" type="text/css">
     <link rel="stylesheet" href="./css/site.css" type="text/css">
     <script src="https://unpkg.com/elm-pep@1.0.6/dist/elm-pep.js"></script>
-    <script src="https://cdn.polyfill.io/v3/polyfill.min.js?features=fetch,requestAnimationFrame,Element.prototype.classList,TextDecoder"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/core-js/3.18.3/minified.js"></script>
     {{{ extraHead.local }}}
     {{{ css.tag }}}
     <title>{{ title }}</title>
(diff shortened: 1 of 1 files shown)


mike-000:
The default Android 4.2 browser has not worked correctly for some time.  Other Chromium based browsers (Chrome, Opera, etc) run Chromium 71, which is ECMAScript 2017 but does not support `Promise.allSettled` and possibly some others which the code uses but are not documented.

tschaub:
@mike-000 - What Android version still uses Chromium 71?  I thought after Android 5 it was supposed to be updatable.

mike-000:
Android 4.2 Chromium 71
Android 4.4 Chromium 81
Android 5   Chromium 95
I think Chrome is no longer on GooglePlay on those versions, but Opera is.

tschaub:
It looks like Android 4.4 fell below 1% [in December of 2020](https://gs.statcounter.com/android-version-market-share/mobile/worldwide/2020).

image

And Android 4.2 was below 1% [in December of 2018](https://gs.statcounter.com/android-version-market-share/mobile/worldwide/2018).

image

I'm in favor of removing the workarounds for these browsers (and any others that are below 1% global usage).

mike-000:
So the minimum requirement will be ES 2020, not ES 2017 ?

ahocevar (approved):

tschaub:
> So the minimum requirement will be ES 2020, not ES 2017 ?

I think we should say we strive to support modern browsers with at least 1% global usage.  I've updated the docs to reflect this.

PR #13965, opened 2022-08-10
Remove workaround and docs for IE 11

tschaub:
This removes workarounds in the icon-color example for IE 11 and updates docs for the icon `imgSize` property.

See #13883.


MoonE (approved):
Thanks @tschaub

PR #13967, opened 2022-08-10
Remove circular dependency

tschaub:
Currently, `ol/Map.js` depends on `ol/control.js`, `ol/control.js` depends on `ol/control/OverviewMap.js`, and `ol/control/OverviewMap.js` depends on `ol/Map.js`.

This branch moves the `defaults` function to `ol/control/defaults.js` to break the circular dependency.  The interactions get the same treatment.


src/ol/control/defaults.js
```diff
@@ -0,0 +1,62 @@
...
+ * @api
+ */
+
+/**
+ * Set of controls included in maps by default. Unless configured otherwise,
+ * this returns a collection containing an instance of each of the following
+ * controls:
+ * * {@link module:ol/Zoom~Zoom}
+ * * {@link module:ol/Rotate~Rotate}
+ * * {@link module:ol/Attribution~Attribution}
```
MoonE, line 31:
```suggestion
 * * {@link module:ol/control/Zoom~Zoom}
 * * {@link module:ol/control/Rotate~Rotate}
 * * {@link module:ol/control/Attribution~Attribution}
```

src/ol/control/defaults.js
```diff
@@ -0,0 +1,62 @@
...
+ * @typedef {Object} DefaultsOptions
+ * @property {boolean} [attribution=true] Include
+ * {@link module:ol/control/Attribution~Attribution}.
+ * @property {import("./Attribution.js").Options} [attributionOptions]
+ * Options for {@link module:ol/control/Attribution~Attribution}.
+ * @property {boolean} [rotate=true] Include
+ * {@link module:ol/control/Rotate~Rotate}.
+ * @property {import("./Rotate.js").Options} [rotateOptions] Options
+ * for {@link module:ol/control/Rotate~Rotate}.
+ * @property {boolean} [zoom] Include {@link module:ol/Zoom~Zoom}.
```
MoonE, line 19:
```suggestion
 * @property {boolean} [zoom] Include {@link module:ol/control/Zoom~Zoom}.
```

MoonE (commented):
In ol/Map the `controls` and `interactions` options description should point to the new location of the defaults functions.

Strangely the import example is only available for the interaction defaults function:
https://deploy-preview-13967--ol-site.netlify.app/en/latest/apidoc/module-ol_interaction_defaults.html
https://deploy-preview-13967--ol-site.netlify.app/en/latest/apidoc/module-ol_control_defaults.html

image

src/ol/control/defaults.js
tschaub: Thanks, addressed in the latest commit.

src/ol/control/defaults.js
tschaub: Addressed in the latest commit.

tschaub:
> In ol/Map the `controls` and `interactions` options description should point to the new location of the defaults functions.

Done.

> Strangely the import example is only available for the interaction defaults function: https://deploy-preview-13967--ol-site.netlify.app/en/latest/apidoc/module-ol_interaction_defaults.html https://deploy-preview-13967--ol-site.netlify.app/en/latest/apidoc/module-ol_control_defaults.html

Good catch, @MoonE.  This was due to the condition that splits on `ol/` here: https://github.com/openlayers/openlayers/blob/c327740108c4904d3df3866aceef2fccc88871da/config/jsdoc/api/template/tmpl/method.tmpl#L37

So import statements for methods with more than one `ol/` in the name (like `ol/control/defaults.defaults`) were not being written.

Maybe @KaiVolland remembers what condition this was guarding against in 9860699d5f72bfacd70053b36fa0da0d48fd7643.

I remove that condition in the latest commit.  In a quick check of the published doc, I don't see any bad import statements written, but there may be some I haven't found yet.

MoonE:
Should the `defaults` methods be the default export in their respective modules?

I pushed a commit that corrects the generated import statement in case you think that's better.

tschaub:
> Should the `defaults` methods be the default export in their respective modules?

Our convention is to only have a `default` export for modules with UppercaseFirstLetters that export a constructor (or an enum historically).  Modules that start with a lowercase first letter always only have named exports.

MoonE (approved):
Fine by me. Thanks @tschaub

PR #13970, opened 2022-08-11
Nicer links in the API docs

tschaub:
This shortens `module:ol/foo/Bar~Bar` type names wherever possible to `Bar` where we link to the given type.

Here is an example of what the map options look like after this change:

image

And the same options before:

image

For people who don't know JSDoc's bespoke `module:foo/Bar~Thing#method` syntax, I think the API docs are much more readable without the extra characters.

We should be including sensible text in the `@link` tags.  When we don't do that, this change shortens the text for the links.

### DIFF
--- a/config/jsdoc/api/index.md
+++ b/config/jsdoc/api/index.md
@@ -73,8 +73,8 @@
     <div class="card h-100 bg-light">
       <div class="card-body">
         <h4 class="card-title">Projections</h4>
-          <p>All coordinates and extents need to be provided in view projection (default: EPSG:3857). To transform coordinates from and to geographic, use <a href="module-ol_proj.html#.fromLonLat">ol/proj#fromLonLat()</a> and <a href="module-ol_proj.html#.toLonLat">ol/proj#toLonLat()</a>. For extents and other projections, use <a href="module-ol_proj.html#.transformExtent">ol/proj#transformExtent()</a> and <a href="module-ol_proj.html#.transform">ol/proj#transform()</a>.<p>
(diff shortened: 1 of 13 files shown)


MoonE:
I've also thought about getting rid of the `module:` prefix for some time. Only showing the base name makes it look even nicer as the full context isn't needed most of the time.

Is the method return type parsed differently? Because no spaces are added there around `|`

tschaub:
> Is the method return type parsed differently? Because no spaces are added there around `|`

I added a commit that gives union types in the return signature a bit more breathing room.  Let me know if you were thinking about going the other way.  Although when complex types end up getting stringified by Catharsis, we get spaces around `|` (and we can't really control this).  So I think space around `|` is the most consistent.

MoonE (approved):
With spaces it is more legible, great work. :+1:

Theres a few instances where the links are already shortened manually, sometimes lowercased `(\{@link +[^ }]+?)(\w+) +\2(\(\))?\}` -> `$1$2}` if you feel like replacing them.

tschaub:
> Theres a few instances where the links are already shortened manually, sometimes lowercased `(\{@link +[^ }]+?)(\w+) +\2(\(\))?\}` -> `$1$2}` if you feel like replacing them.

Those cases revealed that `:` needs to be among the characters that we consider a delimiter for extracting the short name.  I added a commit to cover that and updated the instances that used the short name (and potentially `()`) as the link text.

Thanks for the review.

marcjansen:
Very nice, thxanks.

PR #13972, opened 2022-08-12
Remove opt_ prefix

tschaub:
This removes the `opt_` prefix from parameter and property names.  This was a legacy convention from the Closure Compiler.


ahocevar (approved):
Nice!

PR #13973, opened 2022-08-12
Fix bandcount per texture for 8, 12, 16, etc. bands

mike-000:
It is currently not possible to load the final texture for DataTiles with 8, 12, 16, bands, etc. as the number of bands in the final texture is calculated as zero.  It is not possible to style any band in the final 4.  https://codesandbox.io/s/data-tiles-forked-i9qh4y?file=/main.js

Reduce the array to 7 entries everything works as expected.  With 6 entries band 6 uses band 5 data due to #13704 which seems unrelated.


src/ol/webgl/TileTexture.js
```diff
@@ -278,7 +278,7 @@ class TileTexture extends EventTarget {
       this.textures.push(texture);
 
       const bandCount =
-        textureIndex < textureCount - 1 ? 4 : this.bandCount % 4;
+        textureIndex < textureCount - 1 ? 4 : ((this.bandCount - 1) % 4) + 1;
```
tschaub, line 281: I agree this looks like the right fix.  Thanks for identifying the issue and the fix.  Looks like we don't have a test that covers this.  Would you be able to add one?

mike-000: Done.  The problem #13704 identified with 6, 10, 14, etc. bands also affects sources with just 2 bands, so that is presumably a rendering issue related to `LUMINANCE_ALPHA`?

tschaub (approved):
Nice.  Thanks, @mike-000

PR #13977, opened 2022-08-13
Fix some errors in examples

MoonE:
- A few examples still used the old path to `ol.css`.
- Use the non-symlinked `ol.css` when building the examples so webpack watching files will pick up changes
The symlink in `site/src/theme` maybe could be deleted with this change
- The topolis example did not show error messages with `toastr` becaue jquery has to be loaded first
This moves all scripts  at the end of the `body` to load them in the order they were specified for the example (jquery and bootstrap will be loaded first as they  are always included, but should be no problem)
- Remove `clipboad.js` script as `prism` No longer seems to depend on it
- Only load `lz-string` when the user wants to edit an example
- `es2015-custom-component` and `example-map` still don't load `ol.css` correctly when editing on codesandbox, though that never worked. Unless someone has a smart idea on how to handle these I'd keep it that way for now.

### DIFF
--- a/site/src/download/index.hbs
+++ b/site/src/download/index.hbs
@@ -18,7 +18,7 @@ layout: default.hbs
       <p>
         If you want to try out OpenLayers without downloading anything (<b>not recommended for production</b>), include the following in the head of your html page:
 <pre><code class="language-html">&lt;script src="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/{{ version }}/build/ol.js"&gt;&lt;/script&gt;
-&lt;link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/{{ version }}/css/ol.css"&gt;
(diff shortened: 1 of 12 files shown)


site/src/download/index.hbs
```diff
@@ -18,7 +18,7 @@ layout: default.hbs
       <p>
         If you want to try out OpenLayers without downloading anything (<b>not recommended for production</b>), include the following in the head of your html page:
 <pre><code class="language-html">&lt;script src="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/{{ version }}/build/ol.js"&gt;&lt;/script&gt;
-&lt;link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/{{ version }}/css/ol.css"&gt;
+&lt;link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/{{ version }}/theme/ol.css"&gt;
```
MoonE, line 21: Not quite sure if the changed path will be correct.

tschaub (approved):
Thanks, @MoonE

site/src/download/index.hbs
tschaub: Yeah, we may revisit this.

PR #13982, opened 2022-08-14
Fix kml-timezone example calculations

MoonE:
- The parsed minutes had the wrong sign when the offset is negative.
- When calculating the local time, the minutes should be subtracted instead of added
- Fix a data error in the kml, a fullstop was used instead of a colon.


ahocevar (approved):
Thanks, @MoonE

PR #13984, opened 2022-08-14
Build and deploy the website

tschaub:
This adds a new `tasks/build-website.sh` script for building the examples, api docs, and rest of the website.  It takes optional `-v` and `-l` arguments for the current **v**ersion and the **l**atest release version.  The script is meant to be run in a few different scenarios:

 * commits to `main` - The website needs to be rebuilt with the current version set to `main` and the latest version set to the highest release (sorted according to semver).  The new `deploy` job runs `./tasks/build-website.sh -l ${latest}` where `latest` is set to the highest release.  This updates the `en/main` examples and API docs in addition to the home page and other doc pages.  The `deploy` job then creates a commit in the `openlayers.github.io` repo.  Commits to the default branch there result in a deploy of the website.
 * release tags - The website needs to be rebuilt with "latest" set to the tag - only if the tag is greater than the latest release (in semver terms).  This branch doesn't yet add a CI job to do this.  This job will be added separately, and will act like the one above.
 * pull requests - The "preview" version of the website needs to be built.  For this preview, the `-l` and `-v` arguments are set to `dev` and the results are deployed to Netlify.

As an example of how things will work after this is merged, I temporarily changed the condition of the new `deploy` job to run on commits to the `deploy` branch.  The [`deploy` job](https://github.com/openlayers/openlayers/runs/7832735815?check_suite_focus=true) for the resulting commit in this repo created this commit in the website repo: https://github.com/openlayers/openlayers.github.io/commit/f5e859abb0b81a2ed0a11327066b556c9f685107.  That commit triggered the website repo's [`deploy` job](https://github.com/openlayers/openlayers.github.io/runs/7832773168?check_suite_focus=true), and the GitHub pages deploy was updated (see https://openlayers.org/, examples and API docs won't get the updated theme until the next release).

We could speed things up and simplify everything in the future by just having a storage bucket with the website content.  Then the `deploy` job from this repo could just push there (rather than creating a commit and deploying everything all over again with GitHub pages).


ahocevar (approved):
Looks great, thanks!

PR #13988, opened 2022-08-15
Fix copying of non-linked ol.css, clean example build

### LINKED ISSUE #13983: Extra theme directory created when building examples
When running `npm run build-examples`, there is a `theme` directory that gets created at the root of the repo.  I assume this came with #13977.

MoonE:
Watching `ol.css` worked because it copied the symlinked file when the non-symlinked file changed.
I also added a `rm -rf build/examples` to `build-examples` to have a clean build.
Fixes #13983


tschaub (approved):
Thanks, @MoonE

PR #13989, opened 2022-08-15
Minor adjustments to the website style

tschaub:
This removes a bit of whitespace and adjusts some links on the home page.


MoonE (approved):
It would make sense to also change the link underline for the individual examples [here](https://deploy-preview-13989--ol-site.netlify.app/en/latest/examples/), something like:
```css
a.example {
  text-decoration: none;
}
a.example:hover > .mainlink > strong {
  text-decoration: underline;
}
```

tschaub:
> It would make sense to also change the link underline for the individual examples [here](https://deploy-preview-13989--ol-site.netlify.app/en/latest/examples/), something like:
> 
> ```css
> a.example {
>   text-decoration: none;
> }
> a.example:hover > .mainlink > strong {
>   text-decoration: underline;
> }
> ```

I added another commit that I think gets at what you're suggesting.

MoonE (approved):
Yeah, that was it. Thanks, @tschaub

PR #13996, opened 2022-08-16
Update to bootstrap 5.2

MoonE:
~I'll provide a separate pull request to remove jquery for some parts shortly.~ 
jQuery is removed except a few pages.


examples/templates/example.html
```diff
@@ -3,100 +3,101 @@
...
+                <a class="dropdown-item" href="/doc/tutorials/"><i class="fa fa-book fa-fw me-2 fa-lg"></i>Tutorials</a>
+                <a class="dropdown-item" href="/workshop/"><i class="fa fa-graduation-cap fa-fw me-2 fa-lg"></i>Workshop</a>
                 <div class="dropdown-divider"></div>
-                <a class="dropdown-item" href="https://stackoverflow.com/questions/tagged/openlayers"><i class="fab fa-stack-overflow fa-fw mr-2"></i>Ask a Question</a>
+                <a class="dropdown-item" href="https://stackoverflow.com/questions/tagged/openlayers"><i class="fab fa-stack-overflow fa-fw me-2"></i>Ask a Question</a>
             </div>
           </li>
           <li class="nav-item"><a class="nav-link" href="/en/latest/examples/">Examples</a></li>
-          <li class="nav-item"><a class="nav-link" href="/en/latest/apidoc/"><i class="fa fa-sitemap mr-1"></i>API</a></li>
+          <li class="nav-item"><a class="nav-link" href="/en/latest/apidoc/"><i class="fa fa-sitemap me-1"></i>API</a></li>
```
tschaub, line 43:
```suggestion
          <li class="nav-item"><a class="nav-link" href="../apidoc/"><i class="fa fa-sitemap me-1"></i>API</a></li>
```
Unrelated to this change, but the link should keep you on a specific version.  I can submit a separate PR if you don’t want to include this here.

tschaub:
You may be working on this, but it looks like there are some spacing issues around the news item.

With these changes:
49BEAA80-DF0F-491D-BE5A-88E23B2EFDDF

Before:
C502DBAD-D8BD-432E-8845-CE0D8EAB89BE

jipexu:
Yes it was very necessary ! @MoonE good job !
- in the deploy the dropdown of thedropdown of the code menu is misaligned ...

MoonE:
Thanks, @jipexu, I'm aware of that. I'll provide a new PR to remove jquery once this is merged.

jipexu:
ah ok good ! update bs is always a lot of work ...

examples/index.html
```diff
@@ -61,29 +59,29 @@
           </div>
```
tschaub, line 59: It looks like the markup for the input group should change (no more `input-group-append`): https://getbootstrap.com/docs/5.2/forms/input-group/

There is no longer CSS for the `.input-group>.input-group-append>.input-group-text` border radius.

image

tschaub:
I think I like the nav links without the underline.  It looks like this change adds them on hover/active/focus:

image

What do you think?

tschaub:
This is looking really good.  Thanks for all your work on the upgrade, @MoonE!

site/src/theme/site.css
```diff
@@ -510,4 +516,8 @@ div.code-toolbar>.toolbar button:before {
 
 .nameContainer .anchor:target+h4 {
   background-color: inherit;
-}
\ No newline at end of file
+}
+
+.container {
+  max-width: 1140px;
+}
```
tschaub, line 523: This overrides the `max-width` for all of the breakpoints - making it less responsive.  Did you mean something like 
```css
@media (min-width: 1400px) {
  .container {
    max-width: 1140px;
  }
}
```

MoonE: That was my intention, thanks.

MoonE:
> @MoonE in this example https://deploy-preview-13996--ol-site.netlify.app/en/latest/examples/button-title.html if we remove jquery we have to call the tooltip in another way ...
Sorry, I may have misundertstood. Even with jQueryand  bootstrap 5 these tooltips no longer worked, though there are the normal browser tooltips from the `tooltip` attribute.

I udpated the examples to work again with bootstrap 5, hopefully everything is working now.

examples/cluster.html
```diff
@@ -5,19 +5,21 @@
 docs: >
   This example shows how to do clustering on point features.
 tags: "cluster, vector"
+resources:
+  - https://cdn.jsdelivr.net/npm/bootstrap@5.2.0/dist/css/bootstrap.min.css
```
tschaub, line 9: Should this go in the example template instead?

tschaub: I guess I'm not clear on which examples need it.

MoonE: It is in the examples template, but when editing an example in codesandbox it needs to be loaded there too.

jipexu:
> > @MoonE in this example https://deploy-preview-13996--ol-site.netlify.app/en/latest/examples/button-title.html if we remove jquery we have to call the tooltip in another way ...
> 
> @jipexu Sorry, I may have misundertstood. Even with jQueryand bootstrap 5 these tooltips no longer worked, though there are the normal browser tooltips from the `tooltip` attribute.
> 
> I udpated the examples to work again with bootstrap 5, hopefully everything is working now.

@MoonE 
yes because it is 2 problems : 
if we remove jquery we can not use $ in our examples
if we dont remove jquery doesnt work too because the way to call tooltip has changed in bs5 ...

examples/cluster.html
tschaub: Got it.  Thanks.

jipexu:
@MoonE 
concerning the change of the case of utf-8 and DOCTYPE in the examples
as the meta charset is case insensitive we can use both
but for the doctype who is also  case insensitive  in html5,  i will let in upper case for a better compliance with xhtml or polyglot documents  ? .... 
it’s just a thought i dont know what is better ...

MoonE:
@jipexu Honestly it doesn't really matter, either works without problems. I simply prefer it to be consistent.

site/src/doc/quickstart.md
```diff
@@ -41,6 +41,10 @@ Open the `index.html` file in a text editor.  It should look something like this
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Quick Start</title>
+    <style>
+      #map { height: 400px; }
```
tschaub, line 46: This conflicts with the rule in the style.css.  Not important to change, but I also don’t think it needs to be added.

tschaub (approved):
Thanks for the huge effort, @MoonE!

site/src/doc/quickstart.md
MoonE: You are correct, I didn't notice the css imported from the js file.

PR #13997, opened 2022-08-16
Avoid append only cache in WebGL tile layers

### LINKED ISSUE #13927: Tiles/canvas not released by WebGL renderer?
**Describe the bug**
Image and reproj tiles and reproj tile canvas appear to be not released by WebGL renderer.

**To Reproduce**
See the Codesandbox from #13921 https://codesandbox.io/s/simple-forked-ijj5xm?file=/main.js  It shows tiles being read and released.  But change `ol/layer/Tile` to `ol/layer/WebGLTile` and no releases are logged  Another Codesandbox https://codesandbox.io/s/simple-forked-qfs79z?file=/main.js shows reproj tiles being read and released.  Again change `ol/layer/Tile` to `ol/layer/WebGLTile` and no releases are logged.  So presumably the canvas is also not released?

**Expected behavior**
Some evidence that tiles and their canvas are released.

tschaub:
The WebGL tile renderer maintains its own cache of tile textures.  When a layer is configured with an explicit `cacheSize`, this is passed to the renderer's cache.  And the intention is that the source's cache remains near empty.  In #12671, the cache size for the data tile source was set to 0.1 to make sure it stays empty.  But for all other tile sources, the source's tile cache has a size of 0 or unlimited.

The Canvas tile renderer calls `source.updateCacheSize()` to manage the cache on its layers' sources.  This change makes the WebGL tile renderer do the same.  (Having the renderer reach through the layer and the source to manage the tile cache is further support for moving the cache to the renderer instead of the source.)

Fixes #13927.

### DIFF
--- a/src/ol/renderer/webgl/TileLayer.js
+++ b/src/ol/renderer/webgl/TileLayer.js
@@ -674,6 +674,7 @@ class WebGLTileLayerRenderer extends WebGLLayerRenderer {
      * @param {import("../../Map.js").FrameState} frameState Frame state.
      */
     const postRenderFunction = function (map, frameState) {
+      tileSource.updateCacheSize(0.1, frameState.viewState.projection);
       tileSource.expireCache(frameState.viewState.projection, empty);
     };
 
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks!

ahocevar:
> (Having the renderer reach through the layer and the source to manage the tile cache is further support for moving the cache to the renderer instead of the source.)

If we do that, we might consider doing it in a way that the same source can still be used in more than one layer, without having to have its own cache for each layer renderer.

PR #14007, opened 2022-08-18
Replace the Icon Scale example

mike-000:
Make the example interactive to better illustrate how icon scale, anchor, displacement and rotation, and their text equivalents, interact together.

https://deploy-preview-14007--ol-site.netlify.app/en/latest/examples/icon-scale.html


ahocevar (changes requested):
I like this example because it can help understand some of the style properties. Normally I would say it is too complex and does not help users in writing their own application, but here I think the educational purpose justifies the complexity.

Only I'd remove the vector context switch and implementation, and the note about the lack of multi-line text support, because it does not help the educational purpose. Instead, please create an issue that reminds us about this missing functionality for the vector context API.

ahocevar (approved):
Thanks, @mike-000

PR #14010, opened 2022-08-18
Refresh layer style on input change for icon-scale example

MoonE:
Styles only update after the map view changed.


ahocevar:
@MoonE I think this was also fixed by #14012.

MoonE:
Indeed.

PR #14023, opened 2022-08-20
Do not display TileJSON attributions if empty string

mike-000:
Generally the attributions control is not displayed if the attributions are specified with a falsey value such as empty string https://github.com/openlayers/openlayers/blob/main/src/ol/source/Source.js#L228
However TileJSON tests for undefined resulting in a an attributions control with no content as in https://openlayers.org/en/latest/examples/tilejson.html


ahocevar (approved):
Thanks, @mike-000

PR #14025, opened 2022-08-20
New URL to check for latest release

### LINKED ISSUE #14024: package.json not found
**Describe the bug**
The examples and apidocs try to load the current version from https://raw.githubusercontent.com/openlayers/openlayers.github.io/build/package.json 
but it returns a `404` error.
If you are browsing an older version you no longer get the info to switch to the latest 

**To Reproduce**
1. Go to https://openlayers.org/en/v6.15.1/examples/accessible.html
2. No info about outdated version
3. See error in console

**Expected behavior**
The `package.json` should be available at the URL above.

tschaub:
This updates how we check for the latest release in the examples and API docs.  After this is in, I'll make changes to the openlayers.github.io content for the releases that are already there.

Fixes #14024.

### DIFF
--- a/config/jsdoc/api/template/static/scripts/main.js
+++ b/config/jsdoc/api/template/static/scripts/main.js
@@ -278,11 +278,11 @@ $(function () {
 
   // warn about outdated version
   const currentVersion = document.getElementById('package-version').innerHTML;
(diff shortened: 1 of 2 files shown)


MoonE (approved):
Thanks @tschaub

PR #14046, opened 2022-08-23
Support tracing with the draw interaction

tschaub:
This adds a `trace` option to the draw interaction.  If true (or a condition that evaluates to true), clicking on a target feature will start tracing.  A second click stops tracing.  The `traceSource` option can be used to provide an alternate source for target features to trace.  By default, features from the `source` will be used.

The draw interaction is already a complicated component.  And this adds more nuance to it.  But I think the enhanced functionality is useful.  The behavior of the draw interaction in trace mode differs from the current tracing example in a few ways.  The draw interaction updates the current sketch with the traced coordinates (while the tracing example shows the trace separately from the sketch).  The draw interaction also has support for multiple candidate tracing targets.  At some point, I'll update the example to demonstrate this.  In addition, the draw interaction doesn't force the traced path to be the shortest path around the traced feature.  This allows you to trace more than half of the perimeter of a polygon, for example.  Later, I'll add some logic to make it easier to switch trace directions after a trace has been started.

### DIFF
--- a/examples/tracing.html
+++ b/examples/tracing.html
@@ -1,15 +1,14 @@
 ---
 layout: example.html
 title: Tracing around a polygon
-shortdesc: Example of setting up a draw interaction to easily snap to an existing feature.
+shortdesc: Using the draw interaction to trace around features.
 docs: >
-  This example showcases how the draw interaction API can be set up to make snapping along
-  an existing geometry easier while preserving topology, which is sometimes called "tracing".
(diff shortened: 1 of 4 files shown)


ahocevar (approved):
This is just great! Thanks for this effort, @tschaub.

larssoe74:
Great work @tschaub. We implemented the previous tracing example. The new commit is a big improvement.

PR #14057, opened 2022-08-25
Consider both tracing directions when within snap tolerance

tschaub:
The draw interaction doesn't always support switching trace direction currently (it only works if you start snapped to a vertex).  This change makes it always possible to switch trace directions.

I've also updated the tracing example and added types for flat styles on the draw and modify interactions (this was already supported, but just not included in the type annotation).


ahocevar (approved):
Nice!

PR #14062, opened 2022-08-27
Use preferred tile.openstreetmap.org URL

Firefishy:
Use the preferred URL for tile.openstreetmap.org, the old aliases are no longer recommended now that tile supports HTTP/2 + HTTP/3.

Signed-off-by: Grant Slater <git@firefishy.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ new Map({
   layers: [
     new TileLayer({
       source: new XYZ({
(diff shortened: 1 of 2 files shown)


MoonE (approved):
Thanks @Firefishy

Firefishy:
Link for reference ;-) See: https://github.com/openstreetmap/operations/issues/737

PR #14067, opened 2022-08-28
Include the full build in the ol package

tschaub:
This adds the "legacy" full build to the `ol` package.

Here is [an example](https://codesandbox.io/s/ol-full-build-basic-g1x466) using the full build in a `<script>` tag:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Full Build</title>
    <link
      rel="stylesheet"
      href="https://cdn.jsdelivr.net/npm/ol@7.0.1-pr.14067.6/ol.css"
      type="text/css"
    />
    <style>
      html,
      body {
        margin: 0;
        height: 100%;
      }
      #map {
        position: absolute;
        top: 0;
        bottom: 0;
        width: 100%;
      }
    </style>
  </head>
  <body>
    <div id="map"></div>
    <script src="https://cdn.jsdelivr.net/npm/ol@7.0.1-pr.14067.7/dist/ol.js"></script>
    <script>
      new ol.Map({
        target: "map",
        layers: [
          new ol.layer.Tile({
            source: new ol.source.XYZ({
              url: "https://{a-c}.tile.openstreetmap.org/{z}/{x}/{y}.png"
            })
          })
        ],
        view: new ol.View({
          center: [0, 0],
          zoom: 1
        })
      });
    </script>
  </body>
</html>
```

The source for these script tags is now versioned (e.g. https://unpkg.com/ol@7.0.1-pr.14067.7/dist/ol.js or https://cdn.jsdelivr.net/npm/ol@7.0.1-pr.14067.7/dist/ol.js), so we shouldn't run into the problems we've had with people linking to the head of the default branch in the openlayers.github.io repo (where the default branch name or contents may change).

When we cut the next release, we need to update the `site/src/download/index.hbs` page to point to the new URL.  We should also add a suggestion that people download the script and deploy it with their application if they don't want to rely on these 3rd party services.

The geotiff and ol-mapbox-style dependencies are not bundled into the full build.

Here are examples using the full build with external libraries:
 * [Full build with geotiff](https://codesandbox.io/s/ol-full-build-cog-1e659n?file=/index.html)
 * [Full build with ol-mapbox-style](https://codesandbox.io/s/ol-full-build-mapbox-style-y28ogp?file=/index.html)


ahocevar:
I pushed a commit that changes the location to `dist/ol.js`, adds a source map, and makes the full build iife only, which allows us to use global externals not only for geotiff.js (which I fixed), but also for ol-mapbox-style. I added notes to `ol/source/GeoTIFF` and `ol/layer/MapboxVector` about this additional requirement when using the full build.

config/rollup-config-legacy-build.js
```diff
@@ -0,0 +1,38 @@
...
+    name: 'ol',
+    format: 'umd',
+    globals: {
+      geotiff: 'geotiff',
+      'ol-mapbox-style': 'olms',
+    },
+    sourcemap: true,
+  },
+  external: ['geotiff', 'ol-mapbox-style'],
+};
```
tschaub, line 38: Is this unused?

config/rollup-full-build.js
```diff
@@ -1,22 +1,24 @@
 import commonjs from '@rollup/plugin-commonjs';
-import esbuild from 'rollup-plugin-esbuild';
```
tschaub: Curious if you saw an improvement using terser over esbuild.  If we don’t use this plugin, it can be removed from package.json.

ahocevar: Yes, the build gets 11k smaller with terser. Removing the esbuild plugin now.

tschaub:
Thanks for the additional effort on this, @ahocevar.  Great to see this working with external geotiff and ol-mapbox-style.  I updated the description with links to sandboxes that show examples adding these two external libs.

ahocevar (approved):

PR #14077, opened 2022-08-30
Document and mark the ResourceId operator as public api

bjornharrtell:
Useful to make feature id queries with fx. WFS format/protocol.


ahocevar (commented):
No objection from my side, but could you add a test that shows how you use this operator, also to avoid future regressions?

bjornharrtell:
I had to redo my clone so https://github.com/openlayers/openlayers/pull/14169 replaces this. It's in draft until do the tests.

PR #14079, opened 2022-08-31
Fix control styles on Full Screen Mobile example

mike-000:
Correct the css path

### DIFF
--- a/examples/geolocation-orientation.html
+++ b/examples/geolocation-orientation.html
@@ -11,8 +11,8 @@ tags: "fullscreen, geolocation, orientation, mobile"
     <meta name="viewport" content="initial-scale=1.0, user-scalable=no, width=device-width">
     <title>Mobile Geolocation Tracking with Orientation</title>
     <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.0/dist/css/bootstrap.min.css" crossorigin="anonymous">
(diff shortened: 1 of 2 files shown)


tschaub:
The other examples use the css at `/theme/site.css` instead.

mike-000:
I copied the setup from https://github.com/openlayers/openlayers/blob/main/examples/geolocation-orientation.html  I presumed from that `site.css` was not a full replacement for `ol.css` but site specific customisations?

mike-000:
I think at least one more example may be need changing and/or files may need de-duping.

Most examples use `https://openlayers.org/theme/ol.css` **and** `https://openlayers.org/theme/site.css`

Geolocation Tracking with Orientation is using `https://openlayers.org/en/latest/examples/theme/site.css` **and** `https://openlayers.org/en/latest/examples/theme/ol.css`

while Full Screen Mobile  was attempting to use `https://openlayers.org/en/latest/theme/ol.css` which is a 404.

Should the examples not be version specific?

tschaub:
We decided to have the hosted examples use the css at /theme/.  The other linked style sheets are there for the dev setup.

mike-000:
I have now made `Full Screen Mobile` and `Geolocation Tracking with Orientation` consistent with the other examples.

tschaub (approved):

tschaub:
Thanks, @mike-000

PR #14086, opened 2022-09-01
Updates for ol-mapbox-style v9.2

ahocevar:
Updates the tests for background, because the underlying implementation in ol-mapbox-style has changed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "earcut": "^2.2.3",
         "geotiff": "2.0.4",
-        "ol-mapbox-style": "9.1.0",
+        "ol-mapbox-style": "^9.2.0",
         "pbf": "3.2.1",
         "rbush": "^3.0.1"
       },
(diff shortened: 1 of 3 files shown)


tschaub (approved):

tschaub (changes requested):
Actually, it looks like there might be an unwanted rendering issue introduced by this.

From [this branch](https://deploy-preview-14086--ol-site.netlify.app/en/latest/examples/mapbox-style.html):

image

From the [latest release](https://openlayers.org/en/latest/examples/mapbox-style.html):

image

Not sure which is right.

ahocevar:
Ok, thanks for spotting this, @tschaub. I'm going to investigate this.

ahocevar:
Actually this branch fixes the appearance of the land areas. See https://github.com/openlayers/ol-mapbox-style/pull/671. The latest release suffered from a problem with the changed background rendering. For comparison, see https://openlayers.org/en/v6.15.1/examples/mapbox-style.html.

tschaub (approved):

tschaub:
I see.  Thanks for confirming.

PR #14099, opened 2022-09-08
Workaround for unwanted data interpolation in Firefox

mike-000:
Firefox, since version 56, has chosen to interpolate the result of `drawImage()` when the scale in both dimension is less than 1 as the specification for `imageSmoothingEnabled = false` only describes behaviour when scaling up.  No other current browsers appear to do that.  This is undesirable when reprojecting data, see https://github.com/openlayers/openlayers/pull/13654#discussion_r965219302, especially for bytes which are to be combined as a float.

For example in a non-normalized version of the Multiple COGs example https://codesandbox.io/s/multiple-cogs-forked-u6yyc0?file=/main.js the lower (Firefox) image shows random color pixels along the river, which are not present on the upper (Chrome) image.
image

For affected browsers when downscaling would otherwise occur the workaround removes some rows or columns from the image canvas to avoid downscaling in one dimension, thereby avoiding interpolation.

Without this workaround the browser test for non-parallel DataTile reprojection would fail in Firefox or other Gecko browsers.  As the tests run in Chrome the workaround is partially tested by forcing the workflow in Chrome.

Although the workaround will also apply to the reprojection of image tiles when `interpolate: false` is specified, there is no change to the canvas renderer, where downscaling might also occur.


mike-000:
The problem does not only affect non-parallel projections, as the workaround was being triggered elsewhere (with noticeable performance impact).  This is screenshot (magnified x3) of central Asia from uncorrected https://deploy-preview-13654--ol-site.netlify.app/en/latest/examples/geotiff-reprojection.html Chrome above, Firefox below

image

As the same (or very close) scaling may be used for multiple triangulations of the same stitch context downscaled contexts are now cached to reduce the impact on performance. https://deploy-preview-14099--ol-site.netlify.app/en/latest/examples/geotiff-reprojection.html

mike-000:
An alternative to having a built-in workaround would be for applications to use a polyfill as required to make Gecko browser behavior consistent with other browsers (there would be no caching of downscaled downsized contexts)  https://codesandbox.io/s/geotiff-reprojection-forked-cnzquc?file=/index.html

sweepline:
I have been trying to use this MR, but I am having some problems with reprojecting some Datatiles in firefox. This patch works a lot better, but produces a lot of artifacts still. 

It seems to me the error still happens somewhere within `ol/reproj/render()`, but i cannot pinpoint what is going wrong.

The datatiles in the following images contain 1-banded float values with elevation in each pixel. I have a custom webgl layer to render the tiles.

Firefox Standard OL 9
image
Firefox with polyfill
image
Firefox with this MR
image
Chromium
image

mike-000:
> I have no such issues when reprojection datatiles with u8's.

Reprojection of Uint8 does fail when reprojecting between two UTM projections as in https://github.com/openlayers/openlayers/blob/main/test/browser/spec/ol/reproj/DataTile.test.js#L218

The values returned after reprojection are not found in the original tiles and, given the relationship between bands 1 and 2, and 3 and 4 based on the algorithm used to create the tiles, should not even be possible.  The polyfill fixes that.

mike-000:
> On firefox after calling `putImageData` the resulting canvas has mostly `[191, 0, 0, 255]` but some places it will be slightly pertubated like `[190, 0, 0, 255]` or `[191, 1, 0, 255]`

What is the operating system?  I found an old report https://stackoverflow.com/questions/51089924/firefox-canvas-getimagedata-values-differ-from-original-image-pixel-values  of Firefox doing that on Mac even if alpha is always 255.  It might also depend on the hardware and hardware acceleration settings.

sweepline:
> > On firefox after calling `putImageData` the resulting canvas has mostly `[191, 0, 0, 255]` but some places it will be slightly pertubated like `[190, 0, 0, 255]` or `[191, 1, 0, 255]`
> 
> What is the operating system? I found an old report https://stackoverflow.com/questions/51089924/firefox-canvas-getimagedata-values-differ-from-original-image-pixel-values of Firefox doing that on Mac even if alpha is always 255. It might also depend on the hardware and hardware acceleration settings.

The tests were done on Firefox 125.0.2 (64-bit) on Arch Linux with a Ryzen 9 7900 using the iGPU.

I have also conducted some testing where i rewrote the construction of the sources in `reproj/DataTile` to use a webgl shader. This seems to fix the issue there (no pertubations), but `stitchContext.drawImage` seems to reintroduce the pertubated values. I might try rewriting that part of the projection as well. 

> > I have no such issues when reprojection datatiles with u8's.
> 
> Reprojection of Uint8 does fail when reprojecting between two UTM projections as in https://github.com/openlayers/openlayers/blob/main/test/browser/spec/ol/reproj/DataTile.test.js#L218
> 
> The values returned after reprojection are not found in the original tiles and, given the relationship between bands 1 and 2, and 3 and 4 based on the algorithm used to create the tiles, should not even be possible. The polyfill fixes that.

I went back and double checked. I get the same mini-pertubations on firefox with u8 data. There seems to be no difference between using the polyfill or not though.
The left side is EPSG:26914 and the right side EPSG:26915. The view is in EPSG:26915.
image

---

I was thinking of trying to rewrite the code using webgl for reprojection. It should be possible to do 4 bytes per pixel as well then.

sweepline:
I have now proposed a separate fix using webgl in #15860.

PR #14101, opened 2022-09-08
Layer for rendering STAC item assets

tschaub:
This adds a STAC layer for rendering assets from an item.  It currently creates a tile layer with a GeoTIFF source, but eventually could do other things too.

Examples
 * Using the [Registry of Open Data on AWS](https://deploy-preview-14101--ol-site.netlify.app/en/latest/examples/stac.html)
 * Using [Planetary Computer](https://deploy-preview-14101--ol-site.netlify.app/en/latest/examples/planetary-computer.html)

I'm opening this up to gather ideas on what would make it more convenient.  If we rely just on the STAC core, I don't think we can add all that much convenience (feels just as easy to create a GeoTIFF source).  If we rely on extensions, I think we will have a hard time drawing a line (there are many extensions) and maintaining compatibility (core and all extensions are versioned independently).  Ideally we could provide just the right "hooks" or options for users to not have to write too much application code to do what they want.

Here is a description of the current options/methods:

The `getSourceOptions` option lets users provide a function that returns options for the source given item metadata.  This can also return a promise in cases where async work needs to be performed.  The [Planetary Computer example](https://deploy-preview-14101--ol-site.netlify.app/en/latest/examples/planetary-computer.html) uses this option to [sign](https://planetarycomputer.microsoft.com/docs/concepts/sas/) the asset URLs before the source is created.

The `getItem` method can be used to get the item metadata after the source is ready.  The examples use this method in a `sourceready` event handler to fit the view to the item's bbox.

The `assets` option is used to provide an array of keys that map to assets for the source.  This can also be a function that picks the assets from the lookup.  If no `assets` option is provided, we try to rank the assets based on the type and roles - though this is probably only marginally useful (often there are many `data` and many `overview` COG assets, hard to guess which one(s) would be best).

Band stats would be nice to make use of (if the [raster extension](https://github.com/stac-extensions/raster#statistics-object) is being used).  I also have a branch going that extracts and updates stats from the tiles as they load.  I want to settle on something that works in both cases (stats available in metadata and stats only available after rendering).

Curious to get thoughts from others (cc @m-mohr).

### DIFF
--- a/examples/planetary-computer.html
+++ b/examples/planetary-computer.html
@@ -1,0 +1,14 @@
+---
+layout: example.html
+title: Planetary Computer
+shortdesc: Rendering a STAC Item from Planetary Computer.
+docs: >
+  This example uses the STAC layer to render an item from Planetary Computer.
+  Planetary Computer requres that asset URLs are signed with a token.  The example
+  uses the `getSourceOptions` callback to sign the asset URL.
+tags: "stac, geotiff, cog"
+cloak:
+  - key: get_your_own_D6rA4zTHduk6KOKTXzGB
+    value: <Your API key from https://www.maptiler.com/cloud/>
+---
+<div id="map" class="map"></div>
--- a/examples/planetary-computer.js
+++ b/examples/planetary-computer.js
@@ -1,0 +1,61 @@
+import Map from '../src/ol/Map.js';
+import STAC from '../src/ol/layer/STAC.js';
+import TileLayer from '../src/ol/layer/WebGLTile.js';
+import View from '../src/ol/View.js';
+import XYZ from '../src/ol/source/XYZ.js';
+import proj4 from 'proj4';
+import {register} from '../src/ol/proj/proj4.js';
+import {transformExtent} from '../src/ol/proj.js';
+
+register(proj4); // required to support source reprojection
+
+/**
+ * Get a Shared Access Signature Token to authorize asset requests.
+ * See https://planetarycomputer.microsoft.com/docs/concepts/sas/
+ * @param {string} href The unsigned URL.
+ * @return {Promise<string>} A promise for the signed URL.
+ */
+async function sign(href) {
+  const params = new URLSearchParams({href});
+  const response = await fetch(
+    `https://planetarycomputer.microsoft.com/api/sas/v1/sign?${params}`
+  );
+  const body = await response.json();
+  return body.href;
+}
+
+const layer = new STAC({
+  url: 'https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2B_MSIL2A_20220909T185929_R013_T10TES_20220910T222807',
+  assets: ['visual'],
+  async getSourceOptions(item, options) {
+    for (const source of options.sources) {
+      source.url = await sign(source.url);
+    }
+    return options;
+  },
+});
+
+const key = 'get_your_own_D6rA4zTHduk6KOKTXzGB';
+const background = new TileLayer({
+  visible: false,
+  source: new XYZ({
+    url: 'https://api.maptiler.com/maps/pastel/{z}/{x}/{y}.png?key=' + key,
+    tileSize: 512,
+  }),
+});
+
+const map = new Map({
+  target: 'map',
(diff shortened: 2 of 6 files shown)


m-mohr:
Thank you @tschaub, I did not expect this so quickly. :-) I'm traveling for the next two weeks, but I'll try to have a look at this as soon as possible.

For now, the relevant extensions are:
- raster (for min, max, nodata) - you can also read the nodata value from file:nodata, but I think usually you'd use raster
- proj
  - be aware that neither WKT2 nor PROJJSON are not supported by proj4js yet: https://github.com/proj4js/proj4js/issues/400 
  - proj:epsg only works when it is defined upfront via proj4js

I don't think you need to implement any other extension.

Be aware that the fields could be present in the Item Properties or in the Item Assets. In principle, you could also have them in Collection Assets.

tschaub:
> * raster (for min, max, nodata)

@m-mohr - do you have any example items that use the raster extension for stats?  Curious to see if the same metadata is available in the GeoTIFF in those cases.

marcjansen:
Is this post-FOSS4G goodness? Eagerly anticipating this.

m-mohr:
Hey @tschaub, 
sorry for the slow reply. Here are some examples:

1. All items in these collections have statistics via the raster extension:
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/noaa-mrms-qpe-24h-pass2
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/noaa-mrms-qpe-1h-pass1
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/planet-nicfi-analytic
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/planet-nicfi-visual
2. The items in https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/esa-worldcover also use the raster extension for nodata, but the actual value range can be read from the classification extension (the classes with color codes may also result in nice visualization opportunities)
3. Here's a collection where the classes are stored in the file extension (deprecated though): https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/nrcan-landcover:

The file example makes me think whether OL should use [stac-migrate](https://www.npmjs.com/package/@radiantearth/stac-migrate) as it could cater for these cases where an "upgrade" is needed so that OL only need to cater for the latest version.

m-mohr:
Hey @tschaub, 
sorry for the slow reply. Here are some examples:

1. All items in these collections have statistics via the raster extension:
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/noaa-mrms-qpe-24h-pass2
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/noaa-mrms-qpe-1h-pass1
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/planet-nicfi-analytic
    - https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/planet-nicfi-visual
2. The items in https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/esa-worldcover also use the raster extension for nodata, but the actual value range can be read from the classification extension (the classes with color codes may also result in nice visualization opportunities)
3. Here's a collection where the classes are stored in the file extension (deprecated though): https://radiantearth.github.io/stac-browser/#/external/planetarycomputer.microsoft.com/api/stac/v1/collections/nrcan-landcover:

The file example makes me think whether OL should use [stac-migrate](https://www.npmjs.com/package/@radiantearth/stac-migrate) as it could cater for these cases where an "upgrade" is needed so that OL only need to cater for the latest version.

m-mohr:
I think this can be closed in favor of https://ol-stac.moregeo.it/

PR #14122, opened 2022-09-15
Add support for fetching proj4 definitions from epsg.io

tschaub:
This adds a `fromEPSGCode` function to the `ol/proj/proj4.js` module.  This is called with an EPSG code, fetches the definition from epsg.io, registers the definition with proj4, and returns a promise for a projection.  The `register` function needs to be called first to inject the proj4 dependency.

```js
import proj4 from 'proj4';
import {fromEPSGCode, register} from 'ol/proj/proj4.js';

register(proj4);

async function example() {
  const projection = await fromEPSGCode('EPSG:32721');
}
```

I've added other functions to the same module for setting the lookup function and cleaning up.  These are used by the tests, but not marked as part of the API.  It could be useful to add the `setEPSGLookup` to the API as well.  This would let people provide an alternate way of getting the definitions (e.g. from a db, if epsg.io goes away, etc.).


ahocevar:
What about a signature with an options object, e.g.
```js
fromEPSGCode('EPSG:32721',{loader: loadFromEPSGio});
```
? With This would give users an easy way to use a differnt API, or even a local EPSG database. The `loader` would be Promise that resolves to a projection definiton understood by proj4, which is currently proj4 or WKT v1 format.

src/ol/proj/proj4.js
```diff
@@ -19,18 +31,20 @@ import {get as getTransform} from './transforms.js';
  * registry, e.g. after calling `proj4.defs()`. Existing transforms will not be
  * modified by this function.
  *
- * @param {?} proj4 Proj4.
+ * @param {import("proj4")} proj4 Proj4.
```
tschaub, line 34: The previous version of jsdoc-plugin-typescript didn't like this syntax.  I've released a new version that does (see https://github.com/openlayers/jsdoc-plugin-typescript/pull/19).

tschaub:
> What about a signature with an options object, e.g.`fromEPSGCode('EPSG:32721',{loader: loadFromEPSGio});`

My thinking behind the `setEPSGLookup` function is that a user would call that and then the library could make calls to `fromEPSGCode` with the user provided lookup.  If we instead require that users pass a loader to the `fromEPSGCode` function, then we can't call the same function in the library with the user's lookup.

ahocevar:
Thanks for the explanation, @tschaub. In this case, I think the `setEPSGLookup` function should be `@api` from the start.

tschaub:
I added the `@api` annotations and some additional tests.

ahocevar (approved):

src/ol/proj/proj4.js
```diff
@@ -66,3 +87,72 @@ export function register(proj4) {
     }
   }
 }
+
+/**
+ * @param {number} code The EPSG code.
+ * @return {Promise<string>} The proj4 definition.
+ */
+let epsgLookup = async function (code) {
+  const response = await fetch(`https://epsg.io/${code}.proj4`);
```
rdewit, line 96: Sorry for the late comment on this closed PR, but the results here can be unexpected. See the following:

```
code = 1;
response = await fetch(`https://epsg.io/${code}.proj4`);
```

`response.ok` results in `true`

Also when using `code = undefined` for example.

Unfortunately epsg.io returns a `200` header code, even though the page content says `500: internal server error`. Example: https://epsg.io/1.proj4

Shall I file a separate bug report for this?

rdewit:
The [error report](https://github.com/maptiler/epsg.io/issues/188) mentioned above relates to my [comment about epsg.io error code issues](https://github.com/openlayers/openlayers/pull/14122/files#r973643964).

Ideally the MapTiler team would fix this on their end and no further work would be required here.

AliFlux:
This is fixed at epsg. https://epsg.io/1.proj4 returns 404 now

src/ol/proj/proj4.js
marcjansen: Better even would be a PR with a fix :-)

PR #14126, opened 2022-09-16
Make track properties in Link interaction configurable

ahocevar:
This pull request adds a `trackProperties` option to `ol/interaction/Link`. With that, users can e.g. opt out of tracking the visible layers.

In addition, this pull request fixes an issue with history state when it is not purely managed by the Link interaction. This is done by passing the current `history.state` as first argument to `history.pushState()` or `history.replaceState()`

### DIFF
--- a/src/ol/interaction/Link.js
+++ b/src/ol/interaction/Link.js
@@ -52,9 +52,13 @@ function differentArray(a, b) {
   return differentNumber(a[0], b[0]) || differentNumber(a[1], b[1]);
 }
 
+/** @typedef {'x'|'y'|'z'|'r'|'l'} Params */
+
 /**
  * @typedef {Object} Options
  * @property {boolean|import('../View.js').AnimationOptions} [animate=true] Animate view transitions.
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Link.js
```diff
@@ -52,9 +52,13 @@ function differentArray(a, b) {
   return differentNumber(a[0], b[0]) || differentNumber(a[1], b[1]);
 }
 
+/** @typedef {'x'|'y'|'z'|'r'|'l'} TrackProperties */
+
 /**
  * @typedef {Object} Options
  * @property {boolean|import('../View.js').AnimationOptions} [animate=true] Animate view transitions.
+ * @property {Array<TrackProperties>} [trackProperties=['x', 'y', 'z', 'r', 'l']] Properties to track. Default is to track
```
tschaub, line 60: Minor nit, when reading this below, I keep thinking of "track" the noun, like a racetrack or a GPS track.  Feels like `properties` or `params` (maybe overused, but I think people know "query params") would be more natural.

tschaub (approved):

ahocevar:
`params` sounds good - changed accordingly.

Issue #13396, opened 2022-02-19
Potentially avoidable duplication of WebGL layers


mike-000:
**Describe the bug**
In #13365 I had initially hoped to use a single layer instead of `elevation1`, `elevation2` and `dataLayer` as that should reduce the network traffic and WebGL contexts needed.  Although one layer can be shared by two `ol/source/Raster` there were some rendering issues.  Tiles at the edge of the display are sometimes not included in one of the raster sources - a similar effect to that seen when `maxTilesLoading` is exceeded due to unhandled errors.  I could not replicate when using a `ol/layer/Tile` as source (even though duplication would have a much smaller overhead) so perhaps a timing issue or related to the layer vs source tile caching difference?

A further duplication was needed as `layer.getData()` returns null unless the layer is added directly to the map (that does also apply when using `ol/layer/Tile` but again the overhead is less).

**To Reproduce**
With a single layer https://codesandbox.io/s/webgl-interpolation-forked-1knkv0?file=/main.js notice how - especially near the edges when full page - **Elevation** is sometimes missing after zooming in/out or panning.  It will usually appear after a subsequent small move.  **Data value** never appears as it is always null.

**Expected behavior**
Although using WebGL layers as a source in `ol/source/Raster` may be unusual WebGL contexts are a limited resource and should not have to be duplicated if it can be avoided.

mike-000:
Closed as it is probably no longer possible to share any layer between two raster sources following the improvements made in #14856.  Network traffic can be reduced by layers sharing sources, and the need for WebGL contexts can be reduced by appropriate layer ordering to avoid interleaving WebGL and canvas and layers.

PR #14140, opened 2022-09-25
Tile source in multiple tile queues

### LINKED ISSUE #14139: Tiles sometimes not rendered when source is loaded from two TileQueues
**Describe the bug**
Tiles are only added to the TileQueue when they are in 'idle' state. If one queue already started loading a tile that is subsequently also needed by the same source but via a different TileQueue no tile loading events reach the second layer and therefore the layer may not completely render.

**To Reproduce**
1. Go to the raster.html example (disable tile transitions for easier reproduction)
2. Pan the map and wait for all tiles to load
3. Sometimes the green raster layer overlay has missing tiles until the map is rerendered.

Also may applies to the overviewmap.html example where the source is reused for the OverviewMap

**Expected behavior**
Both layers using the same source should be notified about tiles loading.

MoonE:
Fixes #14139


ahocevar:
Just one thought before you dig too deep into this @MoonE: wouldn't it be another option to just have a single tile queue, i.e. a singleton that each map instance "connects" to?

MoonE:
I also thought about that. I wasn't sure how to consolidate tile priorities from the different queues and how the `maxLoadingTiles` config of Maps should be handled.

ahocevar:
Good question. Other options: For the OverviewMap, it would probably be enough to assign it the map's tile queue in the `setMap()` method. And for the raster source, simply get rid of its tile queue. Reprojection tiles are also fetched without a tile queue.

mike-000:
> And for the raster source, simply get rid of its tile queue. Reprojection tiles are also fetched without a tile queue.

Then you could have the same problem if the unqueued tiles start to load before they are requested for a queued layer.  Also a raster source can use a reprojected source, so potentially a very large number of requests if neither is queued.

tschaub:
Each subscriber to the global queue could register a priority function (and callback).  The queue’s internal priority function would run all subscribers’ priority functions and return the max.

MoonE:
> Good question. Other options: For the OverviewMap, it would probably be enough to assign it the map's tile queue in the setMap() method.

What about using the same source in multiple maps?

> And for the raster source, simply get rid of its tile queue. Reprojection tiles are also fetched without a tile queue.

While the Reprojection tile loads its source tiles without a queue, itself is queued. As @mike-000 said, removing the tile queue for the raster source could load a lot of tiles at once.

mike-000:
#13396 seems related to this.  In a simplified version of the example (obtaining data from a source within a raster source is unrelated) elevation data from the first raster source is always available, but contours from the second raster source which uses the same WebGL tile layer is sometimes missing or incomplete
https://codesandbox.io/s/webgl-interpolation-h350tk?file=/main.js  In this case there is one layer cache of textures, but queued by two raster sources.

PR #14165, opened 2022-10-03
Correct return type for ol/layer/VectorTile~getFeatures()

MoonE:
`ol/layer/VectorTile` may also return `ol/render/Feature`. Currently the return type is always `ol/Feature`.


src/ol/render/canvas/hitdetect.js
```diff
@@ -192,5 +193,5 @@ export function hitDetect(pixel, features, imageData) {
       resultFeatures.push(features[i / indexFactor - 1]);
     }
   }
-  return resultFeatures;
+  return /** @type {Array<F>} */ (resultFeatures);
```
MoonE, line 196: Not sure why it doesn't work without this cast. Its simply copying features from the input array to the returned array ....

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -434,97 +434,97 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
...
+            0
+          ),
         ];
-        const features = tile
-          .getSourceTiles()
-          .reduce(function (accumulator, sourceTile) {
-            return accumulator.concat(sourceTile.getFeatures());
-          }, []);
-        let hitDetectionImageData = tile.hitDetectionImageData[layerUid];
-        if (!hitDetectionImageData && !this.animatingOrInteracting_) {
```
MoonE, line 496: `this.animatingOrInteracting_` does not exist in `ol/renderer/canvas/VectorTileLayer`, so I removed it. Should be ok to ignore interactions as the hitdetect image is per tile. Not sure about animations.

tschaub (approved):

PR #14169, opened 2022-10-04
Document and mark the ResourceId operator as public api

bjornharrtell:
Useful to make feature id queries with fx. WFS format/protocol.


bjornharrtell:
Set to draft because tests should be added before this is ready.

PR #14170, opened 2022-10-04
Use arrow functions instead of binding anonymous functions

tschaub:
This makes use of arrow functions instead of creating an unnecessary anonymous function and calling bind on it.  It appears that TypeScript does a better job interpreting these arrow functions, so this change required a few updates to types.

### DIFF
--- a/src/ol/reproj/DataTile.js
+++ b/src/ol/reproj/DataTile.js
@@ -116,7 +116,7 @@ class ReprojDataTile extends DataTile {
 
     /**
      * @private
-     * @type {!Array<import("../Tile.js").default>}
+     * @type {!Array<DataTile>}
      */
     this.sourceTiles_ = [];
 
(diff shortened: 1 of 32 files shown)


src/ol/reproj/DataTile.js
```diff
@@ -116,7 +116,7 @@ class ReprojDataTile extends DataTile {
 
     /**
      * @private
-     * @type {!Array<import("../Tile.js").default>}
+     * @type {!Array<DataTile>}
```
tschaub, line 119: The `reproject_` method below assumes these are data tiles, but TypeScript didn't notice with the previous syntax.

src/ol/reproj/Tile.js
```diff
@@ -103,7 +103,7 @@ class ReprojTile extends Tile {
 
     /**
      * @private
-     * @type {!Array<import("../Tile.js").default>}
+     * @type {!Array<ReprojTile|import("../ImageTile.js").default>}
```
tschaub, line 106: Use of `tile.getImage()` below doesn't work with the base tile class.

src/ol/webgl/PostProcessingPass.js
```diff
@@ -46,7 +46,7 @@ const DEFAULT_FRAGMENT_SHADER = `
 /**
  * @typedef {Object} UniformInternalDescription
  * @property {import("./Helper").UniformValue} value Value
- * @property {number} location Location
+ * @property {WebGLUniformLocation} location Location
```
tschaub, line 49: TypeScript recognized that this type was wrong after removing bind on anonymous function below.

ahocevar (approved):

src/ol/reproj/Tile.js
MoonE: Can source tiles be `ReprojTile`s?

I believe the return type of `getTileInternal` should be `TileImage` only.
https://github.com/openlayers/openlayers/blob/66fb11b31679e865711884c19fd48eeeee8b4ca7/src/ol/source/TileImage.js#L356-L365
`createTile_` always returns `TileImage` and the cache should also only contains this type.

tschaub: @MoonE - Thanks.  I think #14174 addresses your suggestion.

tschaub: Ooops.  I missed #14173.

PR #14172, opened 2022-10-04
Translate interaction can't work with RenderFeature

MoonE:
As `ol/render/Feature` is readonly it can't be translated.

This discards any `ol/render/Feature` and changes the `FilterFunction` feature type to only accept `ol/Feature`.


src/ol/interaction/Translate.js
```diff
@@ -337,20 +338,26 @@ class Translate extends PointerInteraction {
...
    * @private
    */
   featuresAtPixel_(pixel, map) {
     return map.forEachFeatureAtPixel(
       pixel,
+      /**
+       * @param {import("../Feature.js").FeatureLike} feature Feature.
+       * @param {import("../layer/Layer.js").default} layer Layer.
+       * @return {Feature|undefined} Feature that should be handled if found.
+       */
```
tschaub, line 352: Is this comment block necessary?

MoonE: It helps with type-safety. Without it `feature` and `layer` are of type `any` and no type-checking can happen.

MoonE: Hm, actually the parameter types seem to be inferred automatically, so could be removed.

tschaub: I found that the types were `any` when we were calling `bind` on an anonymous function, but that they are inferred without the bind.

mike-000:
> As `ol/render/Feature` is readonly, it can't be translated.

It cannot be modified either, should the Modify interaction should also be included in this PR?

M393:
I think @mike-000 is correct. 
Modify hit detection with RenderFeature was fixed with #12237, but actually modifying such a feature doesn't make sense.
Maybe the solution could be to disallow modifying VectorTileSources at all?

tschaub (approved):

PR #14173, opened 2022-10-04
Only TileImage as ReprojTile source tile type

MoonE:
`ol/source/ImageTile~createTile_` only returns `ImageTile` and `getTileInternal` only gets tiles by calling `createTile_` or getting tiles out of the cache that were also set in this method. So `ReprojTile` should be removed from `getTileInternal` return type.


tschaub (approved):
Thanks, @MoonE

PR #14178, opened 2022-10-05
Pixel as fixed length array

### LINKED ISSUE #14167: Pixel type should be [number, number]
**Describe the bug**
The Pixel type is number[], however it would more accurately be [number, number]

tschaub:
This updates the `Pixel` type to be a fixed length array.

TypeScript doesn't try to be clever about calls to `slice`.  Given a two element array, it treats the return from `pixel.slice()` as having unknown length.  In the same way (and more understandably), the return from `coordinate.slice(0, 2)` has unknown length.  This requires that we cast the return to retain the fixed length type.  I've added `clone` and `fromCoordinate` function to the `ol/pixel.js` module for this purpose.

In addition, TypeScript gives up on keeping track of array literals as fixed length after the first statement where they appear.  

For example, if the `clone` method expects a two element array, this does not work:
```js
const p = [1, 2];
const c = clone(p);
```

However, this does work:
```js
const c = clone([1, 2]);
```

If the array literal cannot be passed directly as an argument, this is an alternative:
```js
/** @type {import("./pixel.js").Pixel} */
const p = [1, 2];
const c = clone(p);
```

In TypeScript, the use of `as` is required in this case.

I've shifted a few expressions around to pass array literals directly as arguments.  If instead we want to have the type annotations, I can change that.

I'm not attached to these changes.  Mostly just wanted to see how much work would be involved.

Fixes #14167.

### DIFF
--- a/src/ol/Map.js
+++ b/src/ol/Map.js
@@ -40,6 +40,7 @@ import {fromUserCoordinate, toUserCoordinate} from './proj.js';
 import {getUid} from './util.js';
 import {hasArea} from './size.js';
 import {listen, unlistenByKey} from './events.js';
+import {fromCoordinate as pixelFromCoordinate} from './pixel.js';
 import {removeNode} from './dom.js';
 
 /**
@@ -988,7 +989,7 @@ class Map extends BaseObject {
     }
     return applyTransform(
       frameState.coordinateToPixelTransform,
-      coordinate.slice(0, 2)
+      pixelFromCoordinate(coordinate)
     );
   }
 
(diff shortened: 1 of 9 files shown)


marcjansen:
Can we somehow make use of https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser

A wild guess, tbh.

tschaub:
That looks like a nice parser, @marcjansen.  Seems like it could be useful at some point.  I'm going to close this since I'm not motivated to put additional effort into making it work – not sure I can see enough benefit to justify the work.

If someone else wants to pick this up, feel free to start with the commit here.

PR #14180, opened 2022-10-06
Fix mouse/touch interaction anchors when using css transform scale()

### LINKED ISSUE #13283: Mouse events end up in the wrong location when using scaled containers
**Describe the bug**
Openlayers version: 6.9.0

When using css directive `transform: scale` on the map container the mouse events end up en an incorrect place. This behavior has been described earlier in #5747 #9669 and #10211 with no real explanation why it has not been considered for development. Also, it actually affects all kinds of mouse events.

We have created a high resolution print preview and in order to display the entire resulting map on the visible screen, we scale the map container using `transform: scale`. Doing so the map events end up in the wrong place, which prevents us from having interactions active in print preview. 

**To Reproduce**
1. Add  `transform: scale(0.5)` to the map's div container
2. Use mouse wheel to scroll
3. The the map does not zoom where the mouse pointer is
4. Drag the map
5. The map pans slower than the mouse moves
6. Set up a eventhandler on the map object
7. Click on the map
8. The coordinate in the click event does not correspond the the actual coordinate where the mouse was clicked.

**Expected behavior**
The mouse event should occur at the coordinate that corresponds to the coordinate where the mouse event occurred.

### LINKED ISSUE #14179: CSS transform scale interferes with interactions (mouse/touch events)
**Describe the bug**
If the map, or a ancestor, is scaled using css transform scale(), mouse and touch events get offset.
It is like the anchor coordinate for the interaction (eg DoubleClickZoom, MouseWheelZoom, PinchZoom) is calculated using the original size.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/zoom-constrained.html
2. Open dev tools, find the `id="map"` element and add `transform: scale(0.5)` to it (to effectively scale the map to half its size)
3. Hover the map and double click or mouse wheel zoom and see that the anchor point used when zooming is not directly under the mouse cursor.

**Expected behavior**
Zooming should use an anchor point directly under the mouse cursor.

didkan:
closes #14179 
closes #13283


didkan:
@mike-000 
I found some issues with my first approach and rewrote the PR to use `View.getViewportSize_()`, but since it is a private function of `View` i moved the helper function into `View.getViewportScale()`.

I have tested this on the example implementations and also on a custom implementation in another project and it works fine for me.

Not sure I have covered ALL interactions possibly affected though.

ahocevar:
Thanks for your effort on this, @didkan. I'm not sure this is the right approach. Any transform (e.g. also a rotation, or - more generally - a transform matrix on a dom element) can cause this kind of distortion, not just a scale.

I think this could be solved by using `offsetX` and `offsetY` instead of `clientX` and `clientY`. Such a change would be a major rework, but would simplify the code base in some places, because we wouldn't have to deal with `getBoundingClientRect()`.

PR #14203, opened 2022-10-15
Create hit and taint detection canvas with willReadFrequently set to true

### LINKED ISSUE #14175: Latest Chromium warns about canvas's willReadFrequently
**Describe the bug**
Getting a warning from latest Chromium:
> Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently

https://github.com/openlayers/openlayers/blob/daf604b0a83e841d0ebbf9e79d7f6a7a18a540ce/src/ol/render/canvas/ExecutorGroup.js#L222-L228

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/custom-hit-detection-renderer.html
2. Move the mouse over the map
3. See warning in console

**Expected behavior**
Maybe its worth investigating a performance gain here.

Also warns at this line:
https://github.com/openlayers/openlayers/blob/daf604b0a83e841d0ebbf9e79d7f6a7a18a540ce/src/ol/style/IconImage.js#L111

ahocevar:
Fixes #14175.

### DIFF
--- a/src/ol/render/canvas/ExecutorGroup.js
+++ b/src/ol/render/canvas/ExecutorGroup.js
@@ -181,7 +181,9 @@ class ExecutorGroup {
     if (newContext) {
       this.hitDetectionContext_ = createCanvasContext2D(
         contextSize,
-        contextSize
+        contextSize,
+        undefined,
+        {willReadFrequently: true}
       );
     }
     const context = this.hitDetectionContext_;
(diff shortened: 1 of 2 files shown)


mike-000:
The warning is also appearing when using `layer.getData()`, e.g. the WMS GetFeatureInfo examples with `pixelContext` in `ol/renderer/canvas/Layer.js` (and in `ol/webgl/TileTexture.js` if the example is changed to use a WebGLTile layer).

tschaub (approved):

PR #14206, opened 2022-10-16
Worker for loading and composing GeoTIFF tiles

tschaub:
The GeoTIFF source does a lot of work on the main thread.  This leads to dropped frames.

image

Most of the long tasks are due to the `composeTile_` function.  For example, the calls above took > 60 ms.

This branch moves the GeoTIFF loader into a worker.  With all image fetching and tile composing done in a worker, interaction with the map remains smooth on the main thread.

I'm opening this in draft mode because I may try another approach that doesn't include all of geotiff.js in the worker.  It may be that we can still get good performance by just doing the `composeTile_` work in a worker.  And this could potentially be done with a general purpose tile worker.

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -3,59 +3,15 @@
  */
 import DataTile from './DataTile.js';
 import TileGrid from '../tilegrid/TileGrid.js';
-import {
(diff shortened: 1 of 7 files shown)


BeGj:
Is this still something that should be worked on?

PR #14213, opened 2022-10-17
Do not use .getImageData where .createImageData can be used

mike-000:
This avoids Chrome's `willReadFrequently` warning and should also be more efficient.


ahocevar (approved):
Thanks, @mike-000

PR #14216, opened 2022-10-17
Pass transition option along to reproj data tile

tschaub:
This makes it so reprojected data tiles respect the `transition` option passed to a source.


test/browser/spec/ol/reproj/DataTile.test.js
```diff
@@ -46,15 +47,51 @@ describe('ol/reproj/DataTile', () => {
...
-    mapR.setTarget(null);
+    if (map) {
+      map.setTarget(null);
+    }
+    if (mapR) {
+      mapR.setTarget(null);
+    }
     document.body.removeChild(target);
     document.body.removeChild(targetR);
     delete proj4.defs['EPSG:27700'];
```
mike-000, line 58: Unrelated this this PR but EPSG:27700 is not used, this should be
```
 delete proj4.defs['EPSG:32632'];
 delete proj4.defs['EPSG:32636'];
```

ahocevar (approved):

PR #14217, opened 2022-10-18
Minor fixes to ReprojDataTile tests

mike-000:
Delete the correct proj4 defs, and make test data values specific to each source tile


tschaub:
Thanks, @mike-000

PR #14232, opened 2022-10-24
Add wrapX support to the MousePosition control

### LINKED ISSUE #3899: ol.control.MousePosition reports invalid lat/lon on wrapped maps.

### LINKED ISSUE #14231: MousePosition control reports invalid coordinates when world wraps
**Describe the bug**
MousePosition control reports invalid coordinates when world wrap is on (default).

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/mouse-position.html
2. Pan the map horizontally. Notice how longitude exceeds valid values(-180 to 180) when map wraps.
Capture

**Expected behavior**
The control should report correct coordinates regardless of world wrapping.

**Possible regression of https://github.com/openlayers/openlayers/issues/3899** ?

ahocevar:
This pull request adds convenience for those who want to display coordinates in the expected range of the projection for global projections:
```js
new MousePosition({wrapX: true});
```

Fixes #3899
Fixes #14231

### DIFF
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -138,7 +138,7 @@ class MousePosition extends Control {
      * @private
      * @type {boolean}
      */
-    this.wrapX_ = options.wrapX;
+    this.wrapX_ = options.wrapX === false ? false : true;
   }
 
   /**
--- a/src/ol/control/MousePosition.js
+++ b/src/ol/control/MousePosition.js
@@ -11,6 +11,7 @@ import {
   identityTransform,
 } from '../proj.js';
 import {listen} from '../events.js';
+import {wrapX} from '../coordinate.js';
 
 /**
  * @type {string}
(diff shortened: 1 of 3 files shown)


src/ol/control/MousePosition.js
```diff
@@ -46,6 +47,8 @@ const COORDINATE_FORMAT = 'coordinateFormat';
  * initially and the last position is retained when the mouse leaves the viewport.
  * When a string is provided (e.g. `'no position'` or `''` for an empty string) it is used as a
  * placeholder.
+ * @property {boolean} [wrapX=true] Wrap the world horizontally on the projection's antimeridian, if it
```
tschaub, line 50: Looks like the current default is `false`.  Do you think it should be `true`?

ahocevar: I initially meant to make it `true` by default, but then I decided to make it `false` for backwards compatibility. What would be your preference?

tschaub (approved):

PR #14233, opened 2022-10-24
Change loading spinner to use map viewport.

paul121:
Just a little improvement to the load-events example. Because the map viewport is already positioned relative to the map wrapper implementing code should not need to change the positioning of their wrapper. This makes it easier to implement the loading spinner from this example.

The `position: relative;` is required so that the spinner displays in the center of the map. Otherwise if the map wrapper is not `position: relative;` then the spinner will always be in the middle of the browser window, not necessarily the center of the map.

Although this does work, I'm curious, are there any downsides to using the map viewport for this?

### DIFF
--- a/examples/load-events.css
+++ b/examples/load-events.css
@@ -1,10 +1,8 @@
 .map {
   background: #85ccf9;
-  position: relative;
 }
 #map {
   height: 400px;
-  position: relative;
 }
 
 @keyframes spinner {
--- a/examples/load-events.js
+++ b/examples/load-events.js
@@ -24,8 +24,8 @@ const map = new Map({
 });
 
 map.on('loadstart', function () {
-  map.getTargetElement().classList.add('spinner');
+  map.getViewport().classList.add('spinner');
 });
 map.on('loadend', function () {
-  map.getTargetElement().classList.remove('spinner');
+  map.getViewport().classList.remove('spinner');
 });


tschaub:
@paul121 - Though I'm not sure this is explicitly documented anywhere, the idea is that the application "owns" the target element and that the library "owns" the viewport element.  So the library does things like add inline styles or class names to the viewport element.  Since the library creates and destroys the viewport element, the application doesn't really have any guarantees about its stability.  The library doesn't add any inline style or class names to the target element.  It is the responsibility of the application to do that.

I think the example is currently consistent with the responsibilities above.  There may be other examples that contradict what I've said above, but if not, it might be best to keep the example the way it is.  Others may have other opinions.

paul121:
>  the idea is that the application "owns" the target element and that the library "owns" the viewport element.

Thanks @tschaub, that makes sense. Maybe irrelevant to the examples then, but the motivating use-case for me is our farmOS-map *library* that is a wrapper around openlayers. Looking at how to include a loading spinner as a default feature of the farmOS-map library where the wrapper is not required to have the position applied.

Maybe this isn't recommended for examples but will work fine for us

ahocevar:
I agree with @tschaub here. I don't know what the reason was to make `getViewport()` part of the API in #1941. I think the other few examples that use `getViewport()` could be changed to use `getTargetElement()` instead.

paul121:
Thanks all. I'll close this :-)

PR #14239, opened 2022-10-27
Use wkt definitions in EPSG.io Search example

### LINKED ISSUE #14235: Reprojection with EPSG.io Search not working
**Describe the bug**
The Reprojection with EPSG.io Search example is not working with many projections.
In any version since 6.6.0 projections such as 27700 and 3031 are not working.  Presumably something has changed in the epsg.io response as later v6 versions used to work.  Earlier versions do still work.

mike-000:
Fixes #14235

It seems impractical to download large grid shift files on demand, in a real application these would be preloaded, so for the purposes of the example use the wkt definition which contains datum and works with proj4..

### DIFF
--- a/examples/reprojection-by-code.js
+++ b/examples/reprojection-by-code.js
@@ -107,7 +107,7 @@ function search(query) {
           if (result) {
             const code = result['code'];
             const name = result['name'];
-            const proj4def = result['proj4'];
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Good enough for a workaround.

PR #14241, opened 2022-10-28
Hitdetect tolerance for WebGLPoints layer

### LINKED ISSUE #14154: hitTolerance option not works in WebGLPointsLayerRenderer.forEachFeatureAtCoordinate
hitTolerance not used in WebGLPointsLayerRenderer.forEachFeatureAtCoordinate method code,
because of this, the select component does not work correctly for WebGLPointsLayer

MoonE:
Fixes #14154

This implements hit-tolerance for the WebGLPoints layer. It will still only find maximum one feature.
Finding more than one feature would work with hit-tolerance enabled, but it would only find some features when they are overlapping.

### DIFF
--- a/src/ol/renderer/Map.js
+++ b/src/ol/renderer/Map.js
@@ -164,9 +164,7 @@ class MapRenderer extends Disposable {
     const order = 1 / matches.length;
     matches.forEach((m, i) => (m.distanceSq += i * order));
     matches.sort((a, b) => a.distanceSq - b.distanceSq);
-    matches.some((m) => {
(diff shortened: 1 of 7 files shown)


jahow (commented):
Thanks for tackling this! I have tested it and it works fine.

It seems to me like we could return several features with a hit tolerance > 0, this wouldn't catch features that are exactly overlapping but I think most of the time it would give good results. But I'm not familiar with the existing system for the canvas-based layers, so I couldn't get it to work properly.

What was the reason for only returning one feature?

pedrolopes10:
I've tested this and works perfectly!

PR #14251, opened 2022-11-01
Update ArcGIS MapServer examples service url

mike-000:
Esri's sampleserver1 is no longer accessible https://community.esri.com/t5/arcgis-api-for-javascript-questions/sampleserver1-arcgisonline-com-returning-503s/m-p/1200099  There is a similar service available on sampleserver6.


ahocevar (approved):
Thanks, @mike-000

PR #14236, opened 2022-10-25
Update COG path

mike-000:
Fixes #14234

The `TCI` file seems unavailable via the `/2020/` path but is available via the `/36/Q/WD/2020/7/` path.
https://deploy-preview-14236--ol-site.netlify.app/en/latest/examples/cog.html

(The corresponding `B04` and `B08` files used in other examples continue to be available via the shorter path).

PR #14252, opened 2022-11-01
Update more COG paths

mike-000:
The `B04` and `B08` file short paths which were still working in #14236 are no longer working, but can be accessed using the same path as needed by the `TCI` file.


tschaub (approved):
Thanks again, @mike-000

PR #14253, opened 2022-11-01
Replace discontinued ArcGIS sample Feature Service with open data service

mike-000:
Replaces the discontinued sampleserver1 service used in the ArcGIS REST Feature Service example with an open data service.

https://deploy-preview-14253--ol-site.netlify.app/en/latest/examples/vector-esri.html

### DIFF
--- a/examples/vector-esri.js
+++ b/examples/vector-esri.js
@@ -3,6 +3,7 @@ import Map from '../src/ol/Map.js';
 import VectorSource from '../src/ol/source/Vector.js';
 import View from '../src/ol/View.js';
 import XYZ from '../src/ol/source/XYZ.js';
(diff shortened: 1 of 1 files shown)


mike-000:
As with the original example the styles are copied from the Arcgis layer Drawing Info.  It might be better to fetch the layer definition json and create the styles dynamically.

mike-000:
Was there a historical reason for using a loader function and jquery?  It would work with just a url function and format.

ahocevar:
> Was there a historical reason for using a loader function and jquery? It would work with just a url function and format.

Yes, that example was there before the url function. Please update to use a url function.

ahocevar:
Replaced by #14296.

PR #14256, opened 2022-11-04
Add a PMTiles example

tschaub:
This adds an example inspired by the [PMTiles MapLibre example](https://protomaps.github.io/PMTiles/examples/maplibre_raster_dem.html).

https://deploy-preview-14256--ol-site.netlify.app/en/latest/examples/pmtiles-elevation.html

Since PMTiles [encodes its raster data](https://github.com/protomaps/PMTiles/blob/master/spec/v3/spec.md#header-design) as PNG, JPG, or WEBP, it makes more sense to use an image tile source rather than the data tile source, but I'm partial to the API of the data tile source.  So instead of hacking things together with a `tileUrlFunction` and a `tileLoadFunction`, I opted to go with a data source and hack my way to the typed arrays that we want via a canvas.

Ideally, an image source would accept a `loader` that was called with `z, x, y` and returned a promise for an image.  I'd like to add that as an option for the existing source, but it feels like it might be easier to start from scratch with a new image source.

### DIFF
--- a/examples/pmtiles-elevation.css
+++ b/examples/pmtiles-elevation.css
@@ -1,0 +1,7 @@
+table.controls td {
+  padding: 2px 5px;
+}
+table.controls td:nth-child(3) {
+  text-align: right;
+  min-width: 3em;
+}
--- a/examples/pmtiles-elevation.html
+++ b/examples/pmtiles-elevation.html
@@ -1,0 +1,37 @@
+---
+layout: example.html
+title: PMTiles Shaded Relief
+shortdesc: Elevation data from PMTiles.
+docs: >
+  This example shows a shaded relief rendering of elevation data from
+  a [PMTiles](https://github.com/protomaps/PMTiles) source.
+resources:
+  - https://unpkg.com/pmtiles@2.4.0/dist/index.js
+tags: "pmtiles, elevation, data tiles"
+---
+<div id="map" class="map"></div>
+<table class="controls">
+  <tr>
+    <td>elevation</td>
+    <td colspan="2" id="elevationOut"></td>
+  </tr>
+  <tr>
+    <td>location</td>
+    <td colspan="2" id="locationOut"></td>
+  </tr>
+  <tr>
+    <td><label for="vert">vertical exaggeration:</label></td>
+    <td><input id="vert" type="range" min="5" max="50" value="10"/></td>
+    <td><span id="vertOut"></span> x</td>
+  </tr>
+  <tr>
+    <td><label for="sunEl">sun elevation:</label></td>
+    <td><input id="sunEl" type="range" min="0" max="90" value="45"/></td>
+    <td><span id="sunElOut"></span> °</td>
+  </tr>
+  <tr>
+    <td><label for="sunAz">sun azimuth:</label></td>
+    <td><input id="sunAz" type="range" min="0" max="360" value="45"/></td>
+    <td><span id="sunAzOut"></span> °</td>
+  </tr>
+</table>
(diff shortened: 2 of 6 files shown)


bdon:
Woohoo! It's been a while since I've done OL, so I was thinking about how to abstract a PMTiles source across vector or raster sources (or `DataTile`). Here are links to my OL examples:

Vector: https://github.com/protomaps/PMTiles/blob/master/js/examples/openlayers.html#L28
Raster: https://github.com/protomaps/PMTiles/blob/master/js/examples/openlayers_raster.html#L28

Those regex-based approaches are hacky, is there a way to re-use the much cleaner `loader` for the non-`DataTile` use cases?

I do feel like any OL-PMTiles glue deserves a nice abstraction, to handle tile loading states as well as cases not handled in the above (getting `undefined` back from `getZxy` might mean you're beyond `maxzoom` or the tile simply does not exist in the archive, which is valid for "holey" vector datasets).

And source `maxzoom`/`minzoom` can technically be derived from the PMTiles header, but this would need to be async, so might be overkill for any OL integration.

tschaub:
> is there a way to re-use the much cleaner `loader` for the non-`DataTile` use cases?

There definitely should be.  I'll probably make a separate pull request, but it would be nice if image tile sources accepted a `loader` option.

I nearly gave up when I was reminded that you have to go through those `tileUrlFunction` and `tileLoadFunction` options – I see you persisted in your examples.  Instead I opted just to abuse the data tile source and go through a canvas.

> I do feel like any OL-PMTiles glue deserves a nice abstraction, to handle tile loading states as well as cases not handled in the above (getting `undefined` back from `getZxy` might mean you're beyond `maxzoom` or the tile simply does not exist in the archive, which is valid for "holey" vector datasets).

I agree it would be nice to handle that explicitly.  With the data tile loader, tiles end up in an "error" state if the loader throws (e.g. like when the example code tries to access undefined response).  Might be better to represent that as an empty tile.

> And source `maxzoom`/`minzoom` can technically be derived from the PMTiles header, but this would need to be async, so might be overkill for any OL integration.

Sources can start in a "loading" state while they do whatever configuration work needs to be done.  For example, the GeoTIFF source starts in a "loading" state until it reads information about data type, block size, overviews, projections, etc. from the GeoTIFF metadata.  Layer rendering is skipped until a source changes to the "ready" state.  So if we wanted to have a dedicated PMTiles source, it could fetch things like the max zoom level from the header before advertising itself as ready.

Really amazing work on Protomaps and PMTiles, @bdon.

bdon:
Another thing that might be relevant is `getZxy` can take an `AbortSignal` as an argument. This is used internally by Leaflet and MapLibre GL to cancel inflight requests if the tile goes offscreen. This does improve the user experience noticeably as there's less work clogging up the CPU and network for tiles that will never be shown. Is there any machinery inside OL to support this? 

example: https://github.com/protomaps/PMTiles/blob/master/js/adapters.ts#L144

ahocevar (approved):
Beautiful!

ahocevar:
@bdon Regarding `AbortSignal`: OpenLayers handles all tile loading in a tile queue on the map. It currently only removes tiles from the queue when they are no longer needed (because outside the viewport or the zoom has changed). But loading is not aborted. It should be straightforward to add that, thanks for the good suggestion! I'm going to create a separate pull request for that.

@tschaub Regarding a `loader` function for the `TileImage` source, I think that would be a great API improvement. If you prefer to start from scratch, I think making a breaking change to remove `tileUrlFunction` and `tileLoadFunction` would be worth considering. Changing existing application code should be easy:
```js
loader(z, x, y) {
  const url = tileUrlFunction([z, x, y]);
  const image = new Image();
  tileLoadFunction({getImage: () => image});
  return new Promise((resolve, reject) => {
    image.addEventListener('error', reject);
    image.addEventListener('load', resolve); 
  });
}
```

tschaub:
I'll merge this as is.  Then we can work separately on improving the API for asynchronously generating (and loading) the image/canvas used by image tiles.

PR #14258, opened 2022-11-04
Add support for images in data tiles

tschaub:
This makes it so data tiles can contain image data in addition to array data.

Adding support for images in data tiles makes sense as part of a larger plan to render all raster tiles with the WebGL tile renderer.  The plan could look something like this:
 1. Add support for images in data tiles (this branch).
 2. Add a new `ImageTile` (or similarly named) source that extends the `DataTile` source.  This would take a `url` option that would be a convenience for constructing a `loader`.  The `url` could be `string|Array<string>|function(number, number, number):string` (the function form would return a url given z, x, and y tile coordinates.  Deprecate the (canvas) tile layer, the `TileImage` source, and lots of other things.
 3. In a major release, rename the `WebGLTile` layer to `Tile` and remove all the deprecated things.  The `DataTile` things could be renamed to `RasterTile` (so `RasterTile` and `VectorTile` would be the two tile classes in the API).

### DIFF
--- a/examples/pmtiles-elevation.js
+++ b/examples/pmtiles-elevation.js
@@ -20,20 +20,13 @@ function loadImage(src) {
   });
 }
 
-const size = 256;
-const canvas = document.createElement('canvas');
-canvas.width = size;
-canvas.height = size;
-const context = canvas.getContext('2d', {willReadFrequently: true});
-
 async function loader(z, x, y) {
   const response = await tiles.getZxy(z, x, y);
   const blob = new Blob([response.data]);
   const src = URL.createObjectURL(blob);
-  const img = await loadImage(src);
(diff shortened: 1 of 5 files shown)


ahocevar:
I agree on most parts of the plan, except for the deprecation of the Canvas tile layer. The reason is canvas reuse on maps that mix raster tile and vector (tile) layers. For such scenarios, it is better to stay within a renderer type (canvas or WebGL).

An alternative to deprecating the canvas tile layer (renderer) could be to make the future image (data) tiles work with the canvas renderer too.

ahocevar (approved):

tschaub:
Good point about the canvas renderer.  I put up #14262 with a work in progress implementation of a new `ImageTile` source.

PR #14259, opened 2022-11-04
Use local files for GeoTIFF with Overviews example

### LINKED ISSUE #13535: GeoTIFF with Overviews example 404 error
**Describe the bug**
The example is not working, the console shows 404 errors on all three source urls.

mike-000:
Fixes #13535

Use small public domain COGs with overviews downloaded from https://dataverse.nl/dataset.xhtml?persistentId=hdl%3A10411%2FGO2HVB (see also https://research.vu.nl/files/105031068/Simulating_past_land_use_patterns_The_impact_of_the_Romans_on_the_LowerRhine_delta_in_the_first_century_AD.pdf) to produce a simple but working example of using overviews:

https://deploy-preview-14259--ol-site.netlify.app/en/latest/examples/cog-overviews.html


tschaub:
Thanks for reporting and proposing a fix for #13535, @mike-000.

It is arguable that any example is better than a broken example, but I think it would be ideal to have a nicer looking replacement.  Ideally we would have a nice looking image that demonstrated the use of the overview property and nothing else.

Since a nice looking GeoTIFF is probably going to be large, and adding large datasets to this repository will make daily git use more painful, using the https://github.com/openlayers/data repository for example data makes sense.  If we can find any nice imagery with an overview or another more visually compelling dataset, we could add it to https://github.com/openlayers/data and update the example to point to it.

PR #14260, opened 2022-11-05
Avoid creating tiles with one row each

tschaub:
When the GeoTIFF source is used with imagery that doesn't have tiled layout, we end up creating one tile (and one HTTP request) per strip.  In cases where there is only one row per strip, this is a lot of extra overhead.  This change makes it so we try to fetch larger blocks of data, organizing things internally as 256x256 tiles if the image has does not have a tiled layout.

I've included an additional commit in this branch that ensures that the GeoTIFF source `getView` method resolves to a view with at least three zoom levels.  When using GeoTIFFs that only have a single overview or have no overviews, the map is jumpy and frustrating to interact with.  This is mostly due to the constraint added due to the `extent` in the resolved view.  I think it would be a breaking change to remove the `extent` from the resolved view, so in this branch I'm only ensuring that there are at least three zoom levels to make things less frustrating.


src/ol/source/GeoTIFF.js
```diff
@@ -108,6 +108,8 @@ function readRGB(preference, image) {
 const STATISTICS_MAXIMUM = 'STATISTICS_MAXIMUM';
 const STATISTICS_MINIMUM = 'STATISTICS_MINIMUM';
 
+const defaultTileSize = 256;
```
ahocevar, line 111: Is `256` an arbitrary value, or did you choose this purposely over e.g. `512`?

ahocevar (approved):
Looks great, I just dropped a question regarding default tile size.

src/ol/source/GeoTIFF.js
tschaub: We probably need to see how this works out in practice.  If we make it too large, and the image is very wide, it could take a long time to load and parse - since the range requests will be for batches of rows that span the whole image width.

PR #14262, opened 2022-11-05
New image tile source

tschaub:
This adds a new `ImageTile` source that extends the `DataSource` as a convenience for working with image tiles.  This is part 2 in the plan mentioned in #14258.

The [simplest example](https://deploy-preview-14262--ol-site.netlify.app/en/latest/examples/webgl-tiles.html) of its use looks like this:
```js
import Layer from 'ol/layer/WebGLTile.js';
import Map from 'ol/Map.js';
import Source from 'ol/source/ImageTile.js';

const map = new Map({
  target: 'map',
  layers: [
    new Layer({
      source: new Source({
        url: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
      }),
    }),
  ],
  view: new View({
    center: [0, 0],
    zoom: 0,
  }),
});
```

The source is purely a convenience for creating an image loader given a URL.  The `url` option can be a single URL template, an array of templates, or a function that returns a URL.  The resulting loader returns a promise that creates an image, sets its `src`, and resolves when the image loads.  For more complex image loading scenarios, the source can be passed a `loader` option.

When a data tile is disposed, we now dispatch an abort event on an abort controller's signal.  The GeoTIFF source shows how the abort signal can be passed to fetch.

There is probably more to do to support all the functionality of the current image tile stuff, but I think this makes for a simpler foundation than what we have currently.  If we continue with this, we could eventually remove things like:

 * `ol/source/UrlTile.js`
 * `ol/source/TileImage.js`
 * `ol/source/XYZ.js`
 * `ol/ImageTile.js`
 * `ol/tileurlfunction.js`

Still todo:
 - [ ] deal with interim tiles


src/ol/tileurlfunction.js
```diff
@@ -26,17 +22,16 @@ export function createFromTemplate(template, tileGrid) {
...
-          const z = tileCoord[0];
-          const range = tileGrid.getFullTileRange(z);
-          assert(range, 55); // The {-y} placeholder requires a tile grid with extent
-          const y = range.getHeight() - tileCoord[2] - 1;
-          return y.toString();
-        });
+      let maxY;
+      const z = tileCoord[0];
+      if (tileGrid) {
+        // TODO: fix places where we call this without a tile grid (tests fail without the above condition)
```
tschaub, line 28: Here are example test failures without this condition: https://github.com/openlayers/openlayers/actions/runs/3401521368/jobs/5656684976

With the previous code, there would be an unhandled error if the template included a `{-y}` placeholder - which is probably why we didn't notice.  Now the error happens at source construction.

tschaub:
No rush, @ahocevar.  I would also like to do the same.  One significant difference is that the data tile source doesn’t cache tiles.  So this will mean adding a cache to the canvas renderer.

src/ol/renderer/canvas/VectorTileLayer.js
```diff
@@ -96,21 +96,25 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
...
+    this.updateExecutorGroup_(
+      tile,
+      frameState.pixelRatio,
+      frameState.viewState.projection
+    );
+    if (this.tileImageNeedsRender_(tile)) {
+      this.renderTileImage_(tile, frameState);
     }
-    return render;
+    super.drawTileImage(tile, frameState, x, y, w, h, gutter, transition);
```
tschaub, line 117: I wonder if you meant to add more to this commit.  The renderer still calls `prepareTile` here: https://github.com/openlayers/openlayers/blob/a4b4998b78c5f72dc0fb922335cbe338f3a24e27/src/ol/renderer/canvas/VectorTileLayer.js#L144

ahocevar: I don't remember, but obviously yes. I pushed another commit that completes the previous one, and takes into account your new `drawTile_()` method, which I had to make `@protected`.

tschaub:
I've squashed these changes in order to make it easier to deal with the conflicts.  New pull request #15568.

PR #14263, opened 2022-11-05
Include file extensions in examples and docs

tschaub:
This updates the import paths in our docs and examples to include the filename extensions.  Excluding the extension works when a module loader has file system access, guesses, or can be configured to add a file extension.  Including the extension works everywhere.

### DIFF
--- a/site/src/doc/quickstart.md
+++ b/site/src/doc/quickstart.md
@@ -58,21 +58,22 @@ Open the `main.js` file in a text editor.  It should look something like this:
 
 ```js
 import './style.css';
-import {Map, View} from 'ol';
-import TileLayer from 'ol/layer/Tile';
-import OSM from 'ol/source/OSM';
+import Map from 'ol/Map.js';
+import OSM from 'ol/source/OSM.js';
+import TileLayer from 'ol/layer/Tile.js';
+import View from 'ol/View.js';
 
 const map = new Map({
   target: 'map',
   layers: [
     new TileLayer({
-      source: new OSM()
-    })
+      source: new OSM(),
+    }),
   ],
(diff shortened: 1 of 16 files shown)


ahocevar (approved):

site/src/doc/tutorials/background.md
```diff
@@ -36,12 +36,12 @@ For convenience, these are also available as named exports, e.g.
...
 ```js
 import {Map, View} from 'ol';
-import {Tile, Vector} from 'ol/layer';
+import {Tile, Vector} from 'ol/layer.js';
 ```
 
 In addition to these re-exported classes, modules with lowercase names also provide constants or functions as named exports:
 
 ```js
 import {getUid} from 'ol';
```
MoonE, line 45: Should these two imports of ol also be changed? It was changed in quickstart.md.

tschaub: I’m not seeing what you are referring to.  But pleased follow up with a ticket or pull request if there is more to change here.

site/src/doc/quickstart.md
```diff
@@ -58,21 +58,22 @@ Open the `main.js` file in a text editor.  It should look something like this:
 
 ```js
 import './style.css';
-import {Map, View} from 'ol';
-import TileLayer from 'ol/layer/Tile';
-import OSM from 'ol/source/OSM';
+import Map from 'ol/Map.js';
```
MoonE, line 61: I was wondering if this change here was related to the problem this PR addresses.
```patch
-import {Map, View} from 'ol';
+import Map from 'ol/Map.js';
+import View from 'ol/View.js';
```

PR #14282, opened 2022-11-08
Add resolutions option to Raster source

ahocevar:
This pull request adds a `resolutions` option (which also exists on other Image sources) to force rendering to the specified resolutions. This allows for better results when input sources are tile sources, by setting the Raster source resolutions to the input source's tile resolutions.

I updated the [shaded-relief example](https://deploy-preview-14282--ol-site.netlify.app/en/latest/examples/shaded-relief.html) to make use of that. With the view's `maxZoom` restriction removed, and a sensible `maxZoom` on the input source, the shaded relief is rendered much more nicely when zoomed in.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -2,6 +2,12 @@
 
 ### Next version
 
+#### Rendered resolutions of `ol/source/Raster`
+
+Previously, `ol/source/Raster` processed input sources at the current view resolution, which caused interpolation artefacts in cases where input sources were up- or downsampled. Now, `ol/source/Raster` picks up the resolutions from the first input source that has resolutions configured (either implicitly through a tile grid in the case of tile sources, or directly when configured through the `resolutions` constructor option). This improves the rendered output in most cases.
+
+If the previous behavior is desired, configure the source with `resolutions: null`.
+
 #### Fixed `wrapX` behavior of `ol/control/MousePosition`
 
 Previously, `ol/control/MousePosition` always displayed coordinates as-is. Now it has a `wrapX` option,
--- a/examples/shaded-relief.js
+++ b/examples/shaded-relief.js
@@ -133,8 +133,6 @@ const raster = new Raster({
   sources: [elevation],
   operationType: 'image',
   operation: shade,
-  // avoid interpolation artefacts by only rendering native elevation resolutions
-  resolutions: elevation.getTileGrid().getResolutions(),
 });
 
 const map = new Map({
--- a/examples/shaded-relief.js
+++ b/examples/shaded-relief.js
@@ -126,12 +126,15 @@ function shade(inputs, data) {
 const elevation = new XYZ({
   url: 'https://{a-d}.tiles.mapbox.com/v3/aj.sf-dem/{z}/{x}/{y}.png',
   crossOrigin: 'anonymous',
+  maxZoom: 13,
 });
 
 const raster = new Raster({
   sources: [elevation],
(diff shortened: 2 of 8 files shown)


examples/shaded-relief.js
```diff
@@ -126,12 +126,15 @@ function shade(inputs, data) {
...
   crossOrigin: 'anonymous',
+  maxZoom: 13,
 });
 
 const raster = new Raster({
   sources: [elevation],
   operationType: 'image',
   operation: shade,
+  // avoid interpolation artefacts by only rendering native elevation resolutions
+  resolutions: elevation.getTileGrid().getResolutions(),
```
tschaub, line 137: Do you think the first source's resolutions would make a good default?

tschaub: Maybe that would be a breaking change.  But also maybe something to consider later.

tschaub:
The rendering does look way nicer.

Although, it looks like changing the sliders in [the example](https://deploy-preview-14282--ol-site.netlify.app/en/latest/examples/shaded-relief.html) no longer works.

examples/shaded-relief.js
ahocevar: I was also thinking about using the first source's resolutions as default, but then thought it might complicate things for the user when multiple sources are involved.

ahocevar:
Thanks @tschaub for catching the problem with the sliders. I pushed another commit, which updates the extent comparison to account for the resolution adjustment, and fixes that problem.

tschaub:
@ahocevar - it looks like you might not have pushed the new commit

ahocevar:
Indeed! Pushed now.

ahocevar:
@tschaub I added another commit, so resolutions are now picked from the first input source that has them configured. I can merge with or without that commit, just let me know.

tschaub (approved):

tschaub:
Looks good, @ahocevar

src/ol/source/Raster.js
```diff
@@ -589,6 +592,17 @@ class RasterSource extends ImageSource {
       this.layers_[i].addEventListener(EventType.CHANGE, changed);
     }
 
+    if (options.resolutions !== null) {
+      for (let i = 0, ii = this.layers_.length; i < ii; ++i) {
+        const source = this.layers_[i].getSource();
+        const resolutions = source.getResolutions();
```
mike-000, line 598: This returns the original tilegrid resolutions, but the source could be rendered with a difference grid (e.g. vector tiles or reprojected raster tiles) whose resolutions would be more appropriate here.  Maybe `source.getResolutions()` should take an optional projection to handle that?

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,12 @@
 
 ### Next version
 
+#### Rendered resolutions of `ol/source/Raster`
+
+Previously, `ol/source/Raster` processed input sources at the current view resolution, which caused interpolation artefacts in cases where input wources were up- or downsampled. Now, `ol/source/Raster` picks up the resolutions from the first input source that has resolutions configured (either implicitly through a tile grid in the case of tile sources, or directly when configured through the `resolutions` constructor option). This improves the rendered output in most cases.
```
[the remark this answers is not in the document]
ahocevar: Fixed. Thanks for spotting this, @mike-000

src/ol/source/Raster.js
ahocevar: Good suggestion, implemented. The rendering tests now also work without the previous change.

PR #14286, opened 2022-11-10
Update example showing COG with external overviews

### LINKED ISSUE #13535: GeoTIFF with Overviews example 404 error

tschaub:
This updates the data used in the [external overview example](https://deploy-preview-14286--ol-site.netlify.app/en/latest/examples/cog-overviews.html).

Fixes #13535.

I notice that mobile Safari is issuing an `OPTIONS` preflight request for this example, and GitHub pages responds with a 405.  The example should work when deployed, but this looks like it will be a limitation for mobile Safari in some cases with cross-origin data requests (it looks like a bug to me given that the request headers are in the [CORS safe list](https://fetch.spec.whatwg.org/#cors-safelisted-request-header).)

**Update:** the example works in the latest mobile Safari, so it looks like the issue above was an upstream bug.

### DIFF
--- a/examples/cog-overviews.html
+++ b/examples/cog-overviews.html
@@ -4,9 +4,7 @@ title: GeoTIFF with Overviews
 shortdesc: Rendering a GeoTIFF with external overviews as a layer.
 docs: >
   In some cases, a GeoTIFF may have external overviews.  This example uses the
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Beautiful!

PR #14294, opened 2022-11-11
Upscale to avoid unwanted data interpolation in Firefox

mike-000:
A simpler alternative to #14099 which should catch the majority of cases by upscaling the initial draw of a reproj tile canvas by an integer factor for affected browsers where interpolation is not required, then eliminating the excess rows and columns, and hopefully with less performance impact.

Unlike #14099 this method may not catch all cases and will have a constant overhead per reprojection in Firefox, while the overhead in #14099 will depend on the number of triangulations needing it.


mike-000:
#14099 has been simplified and performance improved considerably since it was first proposed, so this alternative has no advantage.

PR #14296, opened 2022-11-12
Update ArcGIS REST Feature Service example

mike-000:
A more comprehensive update which can replace #14253

As well as replacing the broken sample server source, styles are obtained from the service json drawing info, and the default loader is used instead of JSONP (there is still an example of a JSONP custom loader in the Editable ArcGIS REST Feature Service example).

https://deploy-preview-14296--ol-site.netlify.app/en/latest/examples/vector-esri.html


ahocevar (changes requested):
The addition of getting styles from the service adds complexity beyond the scope of the example. I'd prefer to keep the scope of the previous version, maybe even replace the conditional style with a simple static style.

ahocevar (approved):
Thanks, @mike-000

PR #14304, opened 2022-11-16
Use ResizeObserver instead of resize event

### LINKED ISSUE #14247: Automatically call updateSize when target element is rendered in shadowDom slot
When a target element is is added to webcomponent slot, the map remains blank.
The reason for this is that the map starts to render when target element is added to DOM at its original location (not in the slot); however, the target element is then copied into the shadowRoot of the webcomponent and remains blank.
To prevent this issue, you have to call updateSize() after listening to the `slotchange` event of the webcomponent.

Example code HTML where a openlayers target element is a assigned to a slot of the `webcomponent-map` webcomponent:
```html
<webcomponent-map id="map-webcomponent">
    <div id="openlayers-map-target" slot="map-slot"></div>
</webcomponent-map>
```

JS code to update the map by listening to slotchange event:
```js
this.querySelector("#map-webcomponent").shadowRoot.addEventListener('slotchange', () => {
    this.map.updateSize();
});
```

It would be a nice feature if OpenLayers would be able to check whether the target element is rendered by listening to the `slotchange` event.
I don't know if that's possible but I think it would solve some issues for many users.
Other users seem to set a timeout (See: https://stackoverflow.com/questions/60775376/i-cant-display-the-openlayer-map-properly-until-after-a-window-resize), which would not be optimal as you don't know how when the slot element is done rendering.
It took us quite some time to figure out what caused the timing issue and come up with the solution to listen to the `slotchange` event.
Otherwise, I think it would be great if this behaviour (and the slotchange solution) would be described in the documentation as webcomponents are becomming more and more popular.

ahocevar:
This pull request should make it unnecessary to call `map.updateSize()` manually in most cases, even when the map is inside a shadow DOM.

Closes #13913 (although it may not cover everything that @MoonE had in mind there)
Fixes #14247

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ OpenLayers appreciates contributions of all kinds.  We especially want to thank 
 [![ela-compil logo](./sponsor-logos/ela-compil.png)](https://ela.pl/)
 
 > We develop leading Physical Security Information Management (PSIM) software.
-> OpenLayers is the core of our map engine and we love it! 
+> OpenLayers is the core of our map engine and we love it!
 > https://elacompil.recruitee.com/
 
 <br>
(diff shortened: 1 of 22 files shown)


ahocevar:
Accidently pushed wrong branch, closing.

PR #14305, opened 2022-11-16
Use ResizeObserver to detect resizes, also when in Shadow DOM

ahocevar:
This pull request should make it unnecessary to call map.updateSize() manually in most cases, even when the map is inside a shadow DOM.

Closes https://github.com/openlayers/openlayers/pull/13913 (although it may not cover everything that @MoonE had in mind there)
Fixes https://github.com/openlayers/openlayers/issues/14247


tschaub (approved):

tschaub:
I assume we could have a rendering test that put a map in a shadow root, resized stuff, and confirmed that the rendered output was properly rescaled.  Not saying it needs to happen here, but it would be nice at some point.

PR #14319, opened 2022-11-18
Ensure `change:position` event is fired if identify transform used

### LINKED ISSUE #14317: Geolocation do not trigger event change:position if projection setting is omitted
If creating a new Geolocation without the projection setting set, then the event 'change:position' is only fired once.

mike-000:
Fixes #14317

`slice()` the projected position coordinates array to ensure a change event is triggered as an identity transform, either from an undefined projection or one without transforms, will return `this.position_` which will be the same array despite having changed contents.


marcjansen (approved):
Nice catch. Thanks and please merge.

ahocevar (commented):
Thanks, @mike-000

PR #14320, opened 2022-11-20
Assign variable to work around swc minification problem

### LINKED ISSUE #14295: Openlayer tiles do not render on build
I am using React 18, Next 13, OL 7.1.0

It works when I execute the code "yarn run dev" like this

image

But when I deploy on Vercel which is Next.js fron-end server. The tile don't load like this. 

image

How can I solve it?

Link: https://stackoverflow.com/questions/74411056/openlayer-tile-dont-render-on-next-js-vercel-server
Link: https://stackoverflow.com/questions/74396848/openlayer-tiles-do-not-render-on-build

ahocevar:
Fixes #14295, and makes the code easier to read.

### DIFF
--- a/src/ol/tilegrid/TileGrid.js
+++ b/src/ol/tilegrid/TileGrid.js
@@ -433,18 +433,13 @@ class TileGrid {
    * @return {import("../TileRange.js").default} Tile range.
    */
   getTileRangeForExtentAndZ(extent, z, tempTileRange) {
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
Were you able to see what this function looked like when minified with swc before and after this change?

ahocevar:
No, unfortunately.

PR #14321, opened 2022-11-20
Support better detection of projection in GeoTIFF files

### LINKED ISSUE #13708: GeoTIFF `getProjection()` is too simplistic
**Describe the bug**
When loading a geotiff file, openlayers determines the projection in a rather simplistic way, looking only at the `ProjectedCSTypeGeoKey` and `GeographicTypeGeoKey` keys.    In particular, when the key exists but is set to 32767 (user-defined), the result is nonsense.

**Expected behavior**
Openlayers doesn't necessarily need to implement a full geokeys parser - that is most likely better left to the client.  But it should expose an interface so that clients can read the geokeys, work out the correct projection, and can then somehow set the right projection on the source.  (Perhaps there is already a way to do the latter, but I could not figure it out from the documentation.)

In particular, when the `ProjectedCSTypeGeoKey` or `GeographicTypeGeoKey` key is set to 32767, it should be ignored.

ahocevar:
Replaces #14314, which accidently got overwritten and closed.

@otfried, thanks for your contribution. Next time, please don't use your main branch as source for a pul request, then it is easier for maintainers to auto-fix lint. Or, even better, run `npm run lint -- --fix` before pushing your changes, then we don't even have to do that.

Fixes #13708.

### DIFF
--- a/src/ol/source/GeoTIFF.js
+++ b/src/ol/source/GeoTIFF.js
@@ -181,7 +181,10 @@ function getProjection(image) {
     return null;
   }
 
-  if (geoKeys.ProjectedCSTypeGeoKey && geoKeys.ProjectedCSTypeGeoKey !== 32767) {
+  if (
+    geoKeys.ProjectedCSTypeGeoKey &&
+    geoKeys.ProjectedCSTypeGeoKey !== 32767
+  ) {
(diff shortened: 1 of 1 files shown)


otfried:
Thanks for the tips!  This was my first pull request ever, next time I'll do it right :slightly_smiling_face:

PR #14333, opened 2022-11-23
Fix immediate renderer image and text rotation in rotated canvas

mike-000:
Fixes the inconsistent rotations observed in https://github.com/openlayers/openlayers/issues/14308#issuecomment-1321258691

Subtract canvas rotation from image or text rotation so it does not automatically rotate with the view (or double that rotation if `rotateWithView` is specified).

Replace circle style with triangle in rendering tests so any incorrect image rotation can be seen.


src/ol/render/canvas/Immediate.js
```diff
@@ -83,6 +84,14 @@ class CanvasImmediateRenderer extends VectorContext {
      */
     this.transform_ = transform;
 
+    /**
+     * @private
+     * @type {number}
+     */
+    this.transformRotation_ = transform
```
ahocevar, line 91: Something looks wrong here. `transform` is a transform matrix, not a number.

mike-000: The rotation is the number.  I first check that the transform is not undefined or null (which is should never be) as several browser tests which create a CanvasImmediateRenderer without using `drawImages_` or `drawText_` fail if they attempt to access entries from a missing matrix.

src/ol/render/canvas/Immediate.js
```diff
@@ -83,6 +84,14 @@ class CanvasImmediateRenderer extends VectorContext {
      */
     this.transform_ = transform;
```
ahocevar, line 85: Here you apply the transform...

src/ol/render/canvas/Immediate.js
ahocevar, line 91: ... and here you apply the same transform, but the type says `number`. If you want the rotation here, you need to extract it from the transform, or assign null initially and note that in the type.

mike-000: That is not the end of the statement.  I check if the transform exists before extracting from it.

ahocevar: Oh I see. Sorry, my bad.

ahocevar (approved):
Thanks, @mike-000

PR #14344, opened 2022-11-29
Calculate snap tolerance extent allowing for rotation

### LINKED ISSUE #14343: Cursor doesn't snap to features when the view is rotated 45 degrees
**Describe the bug**

Cursor doesn't snap to features if the view is rotated 45/135/225/315 degrees.

**To Reproduce**

Demo: https://codesandbox.io/s/ol-rotation-snap-problem-btt58d?f=

1. Draw two points on the map
2. Try drawing a linestring between the two points
3. → Cursor doesn't properly snap to the points

**Expected behavior**

Cursor should snap to features if the view is rotated 45 degrees.
See the original snap interaction demo: https://openlayers.org/en/latest/examples/snap.html

mike-000:
Fixes #14343

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -7,7 +7,7 @@ import PointerInteraction from './Pointer.js';
 import RBush from '../structs/RBush.js';
 import VectorEventType from '../source/VectorEventType.js';
 import {FALSE, TRUE} from '../functions.js';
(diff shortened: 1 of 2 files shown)


ahocevar (commented):
This looks like a good fix, @mike-000. Would you be able to add a test?

ahocevar (approved):
Thanks, @mike-000

PR #14349, opened 2022-11-29
Convert closest point on geometry edge to user coordinates for snap

### LINKED ISSUE #14345: Snap does not snap to LineString or Polygon edges when using geographic coordinates
**Describe the bug**
An additional snap issue observed when investigating #14343/#14344.

**To Reproduce**
Draw a LineString or Polygon in https://openlayers.org/en/latest/examples/snap.html and the cursor will snap to the geometry edges as well as the vertices.

But when using a geographic coordinates version of that example as in https://codesandbox.io/s/snap-forked-rt4kpl?file=/main.js the cursor does not snap to the edges.  Additionally, if two overlapping circles are drawn, the snap sometimes does not snap to the closest.

**Expected behavior**
Similar behaviour for both coordinate systems.

mike-000:
Fixes #14345

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -500,14 +500,9 @@ class Snap extends PointerInteraction {
               .clone()
               .transform(userProjection, projection);
           }
-          vertex = toUserCoordinate(
-            closestOnCircle(
-              projectedCoordinate,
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #14354, opened 2022-12-01
Mark OGC tile sources as stable

tschaub:
The OGC API – Tiles core v1 spec [has been published](https://opengeospatial.github.io/ogcna-auto-review/20-057.html), so we can mark the vector and map tile sources as part of the API now.


ahocevar (approved):
Nice, thank you!

Issue #13508, opened 2022-03-26
Need a way to scale an ol.style.Icon to a destination size rather than a percentage of the original


justdave:
In `ol.style.Icon` we should have a way to describe the destination size of the image to scale it to, without knowing the original size of the image (because it may be coming from multiple sources which are different sizes, and we want them to all be the same size on the map).

The existing `size` property crops the icon rather than scaling it, and that appears to be the intended behavior.  The `scale` property requires you to know the original size of the image in order to scale it to the correct destination size, because it scales to a percentage of the original size (whatever that size was).

This probably needs a new property such as `destSize` or somesuch to describe the target size of the scaling instead of the percentage to scale.

See also #2720 for previous (lengthy) discussion on this topic which was marked as stale and closed because nobody ever picked it up to fix it.

mike-000:
There is already something similar to that internal to the KML format where an icon scale of 1 is based on a width of 32 pixels regardless of the native size of the image.  It works by waiting for the image to load (if not already cached) and then rescaling, similar to 

https://codesandbox.io/s/icon-forked-o1bfll?file=/main.js

justdave:
Does this comment count as activity?  This feature is still needed.

mike-000:
You can do it as in the CodeSandbox above.  For something more convenient for frequent use you could create a custom class, e.g. https://codesandbox.io/s/icon-forked-uof9by?file=/main.js

KaiVolland:
I really think this would be a nice enhancement to the IconSymbolizer. In a lot of projects i worked on i came to this point and didn't found a proper solution.

During development and use of the GeoStyler we came across this problem again:
https://github.com/geostyler/geostyler-openlayers-parser/pull/549
https://github.com/geostyler/geostyler-openlayers-parser/issues/624

I could imagine `width` and `height` (px) properties on the IconSymbolizers that behave similar to CSS (keep aspect-ratio when only one of them is given, etc.).

What do you think about this @ahocevar @tschaub @mike-000 @MoonE ?

As the Sandbox example above shows this shouldn't be to hard to implement.

ahocevar:
Sounds good, @KaiVolland. Would you be able to provide a pull request?

KaiVolland:
Yes i can give it a try. Not sure when i will be able to find the time for this but i'll do my best.

PR #14364, opened 2022-12-06
Introduce `width` and `height` to Icon style

KaiVolland:
This introduces the `width` and `height` property to the `Icon` style.

It includes an example.

openlayers_width_height

The code is highlighy influenced by https://codesandbox.io/s/icon-forked-uof9by?file=/main.js so kudos to @mike-000 

addresses #13508

### DIFF
--- a/src/ol/AssertionError.js
+++ b/src/ol/AssertionError.js
@@ -60,6 +60,7 @@ const messages = {
   66: '`forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has not been enabled. This is done by providing adequate shaders using the `hitVertexShader` and `hitFragmentShader` properties of `WebGLPointsLayerRenderer`',
   67: 'A layer can only be added to the map once. Use either `layer.setMap()` or `map.addLayer()`, not both',
   68: 'A VectorTile source can only be rendered if it has a projection compatible with the view projection',
+  69: '`width` or `height` cannot be provided together with `scale`',
 };
 
 /**
--- a/src/ol/AssertionError.js
+++ b/src/ol/AssertionError.js
@@ -60,7 +60,7 @@ const messages = {
   66: '`forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has not been enabled. This is done by providing adequate shaders using the `hitVertexShader` and `hitFragmentShader` properties of `WebGLPointsLayerRenderer`',
   67: 'A layer can only be added to the map once. Use either `layer.setMap()` or `map.addLayer()`, not both',
   68: 'A VectorTile source can only be rendered if it has a projection compatible with the view projection',
-  69: '`width` or `height` cannot be provided together with `scale` or `size`',
+  69: '`width` or `height` cannot be provided together with `scale`',
 };
 
 /**
--- a/src/ol/AssertionError.js
+++ b/src/ol/AssertionError.js
@@ -60,7 +60,7 @@ const messages = {
   66: '`forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has not been enabled. This is done by providing adequate shaders using the `hitVertexShader` and `hitFragmentShader` properties of `WebGLPointsLayerRenderer`',
   67: 'A layer can only be added to the map once. Use either `layer.setMap()` or `map.addLayer()`, not both',
   68: 'A VectorTile source can only be rendered if it has a projection compatible with the view projection',
-  69: '`width` or `height` cannot be provided together with `scale` or `imgSize`',
+  69: '`width` or `height` cannot be provided together with `scale` or `size`',
 };
 
 /**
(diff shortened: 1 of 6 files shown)


mike-000:
The `size` option defines the `image.width` and `image.height` when the image is obtained from a sprite.  It need not prevent the `width` and `height` options being used to define the displayed size.

src/ol/AssertionError.js
```diff
@@ -60,6 +60,7 @@ const messages = {
   66: '`forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has not been enabled. This is done by providing adequate shaders using the `hitVertexShader` and `hitFragmentShader` properties of `WebGLPointsLayerRenderer`',
   67: 'A layer can only be added to the map once. Use either `layer.setMap()` or `map.addLayer()`, not both',
   68: 'A VectorTile source can only be rendered if it has a projection compatible with the view projection',
+  69: '`width` or `height` cannot be provided together with `scale` or `size`',
```
marcjansen, line 63:
```suggestion
  69: '`width` or `height` cannot be provided together with `scale`',
```

src/ol/style/Icon.js
```diff
@@ -43,6 +43,8 @@ import {getUid} from '../util.js';
  * @property {Array<number>} [displacement=[0, 0]] Displacement of the icon in pixels.
  * Positive values will shift the icon right and up.
  * @property {number} [opacity=1] Opacity of the icon.
+ * @property {number} [width] The width of the icon in pixels. This can't be used together with `scale` or `size`.
+ * @property {number} [height] The height of the icon in pixels. This can't be used together with `scale` or `size`.
```
marcjansen, line 47:
```suggestion
 * @property {number} [width] The width of the icon in pixels. This can't be used together with `scale`.
 * @property {number} [height] The height of the icon in pixels. This can't be used together with `scale`.
```

src/ol/style/Icon.js
```diff
@@ -163,6 +165,15 @@ class Icon extends ImageStyle {
     }
     assert(src !== undefined && src.length > 0, 6); // A defined and non-empty `src` or `image` must be provided
 
+    // `width` or `height` cannot be provided together with `scale` or `size`
+    assert(
+      !(
+        (options.width !== undefined || options.height !== undefined) &&
+        (options.scale !== undefined || this.size_ !== undefined)
```
marcjansen, line 172:
```suggestion
        options.scale !== undefined
```

KaiVolland:
> Should `width`, `height` and `scale` always be in sync, so that setting one will ensure that all getters return actual values?
> 
> We should have a test that ensures that the image is effectively scaled in the aspect ratio, when only one of `width` or `height` is being set. 

I tried to create tests for both of theses cases but unfortunately the used image has no dimensions in the test suite (e.g. the basse64 src). The dimensions of the image are used to calculate the scale in [`updateScaleFromWidthAndHeight`](https://github.com/openlayers/openlayers/pull/14364/files#diff-7352598be22ae8538928c44c863406baf10c31b1225d47774f09b1373280aca9R293).

> The `size` option defines the `image.width` and `image.height` when the image is obtained from a sprite. It need not prevent the `width` and `height` options being used to define the displayed size.

Ok. You're right. I'll see who it currently behaves and adapt the assertion.

KaiVolland:
@mike-000 I adapted the assertion and checked how the example behaves when used with `size` (and `offset`) and it's all fine i would say.

KaiVolland:
I couldn't manage to add a test for the scale. But i added the scale to the example to verify everthing works as expected.

But ofcourse a test would be much better as a broken example won't be detected by CI. If some has an idea how to fix the issue with the undefined dimension i could add the regarding tests. I could imagine that the 'load' eventlistener in the constructor has some timing issues with the testsuite. But as state i couldn't handle it.

marcjansen:
@KaiVolland What do you think about https://github.com/marcjansen/openlayers/commit/24e3a53ea421093d2066529104cd8b88ac52ef41 which uses `imgSize` when testing. I am not sure whether this counts as cheating, though. `npm run test-browser` passes locally with these changes.

Feel free to cherry-pick if you want.

marcjansen:
> The image will not load until required by the renderer unless you force a load as in https://github.com/openlayers/openlayers/blob/main/test/browser/spec/ol/style/icon.test.js#L29-L40

This looks like a saner approach. Maybe @KaiVolland can pick and amend as suggested. That'd be great.

KaiVolland:
I added some code and tests to ensure that `scale` and `width`/`height` are in sync.

src/ol/style/Icon.js
marcjansen: Comment still applies.

marcjansen (approved):
On documentation comment, otherwise looks great to me.

marcjansen (approved):
Great. Thanks for the nice collaboration, everyone.

gmerciel:
First of all, thanks all the developers for this excellent library!

I'm seeing a bug with this commit. You can not _clone_ the Style or the Icon (needed to get clicked features), since it will try to set both height and width, and the scale, thus giving an AssertionError when creating the new Icon.

> AssertionError: `width` or `height` cannot be provided together with `scale`
>     at assert (asserts.js:12:11)
>     at new Icon (Icon.js:169:5)
>     at Icon.clone (Icon.js:259:12)
>     at Style.clone (Style.js:236:48)
>     at createHitDetectionImageData (hitdetect.js:74:35)

IMHO it would be better if this assertion is removed, since the properties are in sync, so they shouldn't cause a conflict (or maybe the assert should check if you specify both, that they are in sync.

tschaub:
Thanks for the report, @gmerciel.  Can you open a dedicated issue describing the bug?

If we want to keep the assertion as is, one option would be to update the clone method to not include the scale of height and width are defined.  Whatever the fix, tests should be added to cover the case.

gmerciel:
Thanks @marcjansen!

Issue #13529, opened 2022-04-01
NumpyTile example not working on Android (chrome)


fabry24:
**Describe the bug**
Android version: 10
Chrome version: 100.0.4896.58
Device: Samsung Galaxy A6

In this example https://openlayers.org/en/latest/examples/numpytile.html is unexpected behaviour.
WebGL tile layer does not show in map and in console are warnings (img).

We think that it is caused by using _Float32Array_ data type, because another examples with data tiles and webgl are correct.
We found out that extension _OES_texture_float_ is not supported on device.
We reproduced this bug on multiple devices.
![img](https://user-images.githubusercontent.com/78112061/161223214-901fb2e4-a1a2-4e08-8f50-9c43c259d63d.jpg)

mike-000:
It depends on the WebGL capabilities of the device, not the version of Android or Chrome.  It works for me on old Samsung Galaxy Tab 2 (Android 4.2 and Chrome 71) and LG G Pad (Android 4.4 and Chrome 81), as well as modern Nokia X20 (Android 12 and Chrome 100).

fabry24:
It seems in GeotiffSource it is already solved ((https://openlayers.org/en/latest/examples/cog-math-multisource.html). Should it be possible to solve this problem also for WebGL with DataTileSource?

mike-000:
The GeoTIFF examples have 8 bit bands so do not need Float32 support.

mike-000:
It should be possible to fallback to half float if Float32 is not supported.  This https://deploy-preview-13534--ol-site.netlify.app/examples/numpytile.html works for me if I force half float on devices which also support Float32 but I do not have anything to test on which does not support Float32.  Does it work on your Galaxy A6?

fabry24:
Thank for your possible solution, but it doesn't work on my device. It is still the same warning.

mike-000:
I have checked similar devices (such as Galaxy A8) using BrowserStack and LambdaTest and in all cases where there was no support for `OES_texture_float` there was also no support for `OES_texture_half_float`, so there appears to be no workaround

![image](https://user-images.githubusercontent.com/49240900/161588450-b8b1d983-6bb5-41f4-acdb-b2495becf205.png)

PR #14366, opened 2022-12-07
webgl2-support && tests changed

fabry24:
Possible fix for #13529.

I have made a devices investigation and find out there are multiple newer devices not supporting OES_texture_float and many more not supporting OES_texture_float_linear, which makes NumpyTiles example and many more applications of this concept unuseful. However, many of these devices supports WebGL2, which natively support working with floating point by default. Also, there always could be devices not supporting OES_texture_float_linear, but we can fallback then to gl.NEAREST - as already implemented.

 Could you please make a revision of my merge request to get to main OL branch to support many more devices?

### DIFF
--- a/src/ol/webgl/TileTexture.js
+++ b/src/ol/webgl/TileTexture.js
@@ -9,7 +9,7 @@ import ImageTile from '../ImageTile.js';
 import ReprojTile from '../reproj/Tile.js';
 import TileState from '../TileState.js';
 import WebGLArrayBuffer from './Buffer.js';
-import {ARRAY_BUFFER, STATIC_DRAW} from '../webgl.js';
+import {ARRAY_BUFFER, RGB32F, RGBA32F, STATIC_DRAW} from '../webgl.js';
 import {createCanvasContext2D} from '../dom.js';
 import {toSize} from '../size.js';
 
@@ -80,9 +80,13 @@ function uploadDataTexture(
   }
 
   let format;
+  let internalFormat;
   switch (bandCount) {
     case 1: {
       format = gl.LUMINANCE;
+      textureType = gl.UNSIGNED_BYTE;
+      data = new Uint8Array(Array.prototype.slice.call(data));
+
       break;
     }
     case 2: {
@@ -90,10 +94,12 @@ function uploadDataTexture(
       break;
     }
     case 3: {
+      internalFormat = data instanceof Float32Array ? RGB32F : gl.RGB;
       format = gl.RGB;
       break;
     }
     case 4: {
+      internalFormat = data instanceof Float32Array ? RGBA32F : gl.RGBA;
       format = gl.RGBA;
       break;
     }
@@ -104,10 +110,11 @@ function uploadDataTexture(
 
   const oldUnpackAlignment = gl.getParameter(gl.UNPACK_ALIGNMENT);
   gl.pixelStorei(gl.UNPACK_ALIGNMENT, unpackAlignment);
+
   gl.texImage2D(
     gl.TEXTURE_2D,
     0,
-    format,
+    internalFormat ? internalFormat : format,
     size[0],
     size[1],
     0,
@@ -115,6 +122,7 @@ function uploadDataTexture(
     textureType,
     data
   );
+
   gl.pixelStorei(gl.UNPACK_ALIGNMENT, oldUnpackAlignment);
 }
 
--- a/examples/numpytile.js
+++ b/examples/numpytile.js
@@ -29,7 +29,7 @@ function numpyTileLoader(z, x, y) {
           dataTile[px * 5 + 2] = numpyData.data[bandSize * 2 + y * 256 + x];
           dataTile[px * 5 + 3] = numpyData.data[bandSize * 3 + y * 256 + x];
           dataTile[px * 5 + 4] =
-            numpyData.data[bandSize * 4 + y * 256 + x] > 0 ? 1.0 : 0;
+            numpyData.data[bandSize * 4 + y * 256 + x] > 0 ? 255 : 0;
         }
       }
       return dataTile;
--- a/src/ol/webgl.js
+++ b/src/ol/webgl.js
@@ -76,6 +76,18 @@ export const UNSIGNED_INT = 0x1405;
  */
 export const FLOAT = 0x1406;
 
+/**
+ * @const
+ * @type {number}
+ */
+export const RGB32F = 0x8815;
+
+/**
+ * @const
+ * @type {number}
+ */
+export const RGBA32F = 0x8814;
+
 /** end of goog.webgl constants
  */
 
@@ -83,12 +95,18 @@ export const FLOAT = 0x1406;
  * @const
  * @type {Array<string>}
  */
-const CONTEXT_IDS = ['experimental-webgl', 'webgl', 'webkit-3d', 'moz-webgl'];
+const CONTEXT_IDS = [
+  'webgl2',
+  'experimental-webgl',
+  'webgl',
+  'webkit-3d',
+  'moz-webgl',
+];
 
 /**
  * @param {HTMLCanvasElement} canvas Canvas.
  * @param {Object} [attributes] Attributes.
- * @return {WebGLRenderingContext} WebGL rendering context.
+ * @return {WebGLRenderingContext | WebGL2RenderingContext} WebGL rendering context.
  */
 export function getContext(canvas, attributes) {
   attributes = Object.assign(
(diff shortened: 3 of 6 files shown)


src/ol/webgl/TileTexture.js
```diff
@@ -104,17 +110,19 @@ function uploadDataTexture(
 
   const oldUnpackAlignment = gl.getParameter(gl.UNPACK_ALIGNMENT);
   gl.pixelStorei(gl.UNPACK_ALIGNMENT, unpackAlignment);
+
   gl.texImage2D(
     gl.TEXTURE_2D,
     0,
-    format,
+    internalFormat ? internalFormat : format,
```
mike-000, line 117: Float32 should be supported with WebGL as well as WebGL2. Many PCs now running Windows 10 (upgraded from Windows 7) have no WebGL2 support but do support GL_OES_texture_float and GL_OES_texture_float_linear.

src/ol/webgl/TileTexture.js
```diff
@@ -80,20 +80,26 @@ function uploadDataTexture(
...
+      data = new Uint8Array(Array.prototype.slice.call(data));
+
       break;
     }
     case 2: {
       format = gl.LUMINANCE_ALPHA;
       break;
     }
     case 3: {
+      internalFormat = data instanceof Float32Array ? RGB32F : gl.RGB;
```
mike-000, line 97: Many PCs now running Windows 10 (upgraded from Windows 7) have no WebGL2 support but do support `GL_OES_texture_float` and `GL_OES_texture_float_linear`.

fabry24: Thank you, I have made a fix of it. Now it supports both webgl and webgl2.

tschaub:
@fabry24 - Thanks for your work on this.  The [rendering test failure](https://github.com/openlayers/openlayers/actions/runs/3639294257/jobs/6162723576) is from the [webgl-palette case](https://github.com/openlayers/openlayers/tree/main/test/rendering/cases/webgl-palette).

This is the expected output:

expected

And here is the actual output in this branch:

actual

You can run this single rendering test in "interactive" mode with the following:
```
node test/rendering/test.js --match webgl-palette --interactive
```

That will start the test server and use Chromium to render that case.  While the test server is running, you can open that same page with a different browser (http://127.0.0.1:3000/cases/webgl-palette/) for easier debugging.

src/ol/webgl/TileTexture.js
```diff
@@ -80,20 +80,32 @@ function uploadDataTexture(
   }
 
   let format;
+  let internalFormat;
   switch (bandCount) {
     case 1: {
       format = gl.LUMINANCE;
+      textureType = gl.UNSIGNED_BYTE;
+      data = new Uint8Array(Array.prototype.slice.call(data));
```
mike-000, line 88: The failing test uses single band Float32 data but you are you are converting that to UInt8.  Since single band Float32 data is valid and can have values beyond the range of UInt8 that seems wrong.  Should the internal formats for 1 and 2 bands be R32F and RG32F for WebGL2 if the data is Float32?

tschaub:
I'm not that excited about the idea of trying to support both WebGL 1 and 2 if it means spreading fragile conditions across the library.

@fabry24 - Can you provide a bit more detail on what you have found regarding float texture and WebGL 2 support.  Ideally, a list of browsers/devices that we would gain.

I think it is worth considering a switch to WebGL 2 without providing a fallback.  It would be nice to have a list of what browsers/devices we would lose in this case.

The [OES_texture_float_linear](https://caniuse.com/mdn-api_oes_texture_float_linear) and [WebGL 2](https://caniuse.com/webgl2) coverage numbers make it look like we would drop from 97% to 94%.  But I'm not sure if that is the only way to look at it.

mike-000:
Support is more related to hardware and drivers than to browsers.  On one desktop PC (Windows 7 upgraded to Windows 10) Chrome `chrome://gpu` and Edge `edge://gpu` report WebGL2 as unavailable and a WebGL2 context cannot be created.  On the same system Firefox shows WebGL2 in `about:support` and can use WebGL2.  But Chrome and Edge now dominate on Windows PCs and Firefox is rarely used.  As noted in https://github.com/openlayers/openlayers/issues/13500#issuecomment-1094976674 support for `OES_texture_float_linear` on older mobile devices is limited, although support on recent devices is good.  https://mikenunn.net/demo/testext.html will show a simple list of available extensions useful for screenshots via BrowserStack or LambdaTest.

fabry24:
We have found that there are mobile devices that support webgl2, which automatically includes support for oes_texture_float, but still do not support this library within webgl1. We tested the devices with the Chrome browser.
The list of non-functioning devices tested ( the numpytiles example from the openlayers examples did not work on these devices), but would work correctly within webgl2: 
Samsung A53,
Samsung Galaxy S21 Ultra,
Oppo Reno5 Z,
Xiaomi 11T,
Samsung Galaxy S10,
Samsung Galaxy S21,
Samsung Galaxy S21+

mike-000:
> I'm not that excited about the idea of trying to support both WebGL 1 and 2 if it means spreading fragile conditions across the library.

It would simplify things a bit if LUMINANCE_ALPHA was never used with WebGL2

      internalFormat = webGL2 ? float ? gl.RG32F : gl.RG8 : gl.LUMINANCE_ALPHA;
      format = webGL2 ? gl.RG : gl.LUMINANCE_ALPHA;

Then we need to check for WebGL2 support anywhere else a `ol/has` property could be used.

fabry24:
I have other devices that are not working properly
Samsung S8, S8+, S10e S20 and S21 Ultra

mike-000:
Tests with Samsung devices can give variable results.  Samsung used either Mali or Adreno GPUs depending on the sales region, etc.  Only one of those has a problem with `OES_texture_float`.

fabry24:
I tried to follow your instructions in supporting webgl2 to support also devices from other regions.
I have recognized problem with mocking HtmlCanvasElement.prototype.getContext, because of ol/has provides WEBGL2 support as constant - it is the same problem as DEVICE_PIXEL_RATIO constant. If I mock getContext function in test, there is already WEBGL2 constant evaluated as true - so actual state does not match ol/has constant. Should I make WEBGL2 constant as function?

mike-000:
I had in mind something similar to the previous has `WEBGL` value https://github.com/openlayers/openlayers/commit/ef10834eb3816a7e489d18d711580debd4dbd0a6 which was not set until the first call was made.  Unfortunately the initial styles are parsed before the renderer is constructed, which would either mean constructing the renderer with an empty style then updating the style, which might be inefficient or, perhaps better, adding a new uniform for WebGL version so the `bandIndex` could be defined as

`bandIndex = '(u_version == 1) ? 3 : 1';`

(assuming that is valid GLSL syntax).

fabry24:
in my last commit is fixed webgl2 support with fallback to webgl1 also with support of LUMINANCE and LUMINANCE_ALPHA. I have used Uniforms as you suggested. I added new tests for 6-band float32, also with HtmlCanvasElement.prototype.getContext() mocking test.
could you check it please?

mike-000:
That looks good as it avoids breaking styles for UInt8 data.  However, 1 and 2 band float data with values in the 0 to 1 range such https://codesandbox.io/s/simple-forked-l6fyhn which will work with WebGL would still need explicit styling to work with WebGL2.  To avoid explicit styling a default color style could be set, which would also work for UInt8 data.  How to other reviewers feel about adding defaults?

On browsers which do not support WebGL2 `WebGL2RenderingContext` is undefined, so testing for it would throw an error.  A safer way to test for WebGL2 is when the context is not `WebGLRenderingContext`.

fabry24:
I've changed the WebGL2RenderingContext check, also added default styles for single band and 2-bands inspired by your commit with test. Could I have anything more to do?

mike-000:
The `RGB32F` and `RGBA32F` constants are not needed as they are also properties of the context similar to `gl.R32F` and `gl.RG32F`.

'internalFormat ? internalFormat : format,' could simply be `internalFormat,` as all cases now set a value for `internalFormat`.

An upgrade note might be needed.  With UInt8 data if the `bandCount` was left defaulted to 4 but only 1 or 2 bands provided in the data (as in https://github.com/openlayers/openlayers/blob/main/test/rendering/cases/webgl-data-tile-interpolate-false/main.js) it did and will still style correctly as 4 bands.  But with float data the correct `bandCount` will now need to be specified otherwise the output will differ depending on whether WebGL or WebGL2 is used.

mike-000:
I think the original API description for `bandCount` remains sufficient.  But as a potential breaking change it should be mentioned in https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md

The deleted comment originally at https://github.com/openlayers/openlayers/blob/main/src/ol/style/expressions.js#L507 is still relevant and would be helpful for anyone maintaining the code.

yurivlasov:
Do you have any idea in which version you can incorporate these changes? I agree with the discussion that it is a problem on many devices. Thank you

fabry24:
Hello @mike-000 ,
Do you have any new news about this?
Are there any further adjustments needed?
Do you plan to include our changes in a release?

PR #14395, opened 2022-12-21
Fix types and docs for GeolocationError

### LINKED ISSUE #14390: Export class GeolocationError
The class **GeolocationError** should be exported, so that the "code"-property be accessible in Typescript.
The issue has already been discussed here: #10905

ahocevar:
Fixes #14390.


marcjansen (approved):
Thanks, please merge.

PR #14400, opened 2022-12-22
Add repeat property for text along lines

ahocevar:
This pull request adds a new `repeat` property to `ol/style/Text`. When a value `n` is provided, it means that text along a line (i.e. when `placement` is set to `'line'`) will be repeated every `n` pixels.


ahocevar:
Tested in combination with https://github.com/openlayers/ol-mapbox-style/pull/768, this is now ready for review.

tschaub (approved):

PR #14398, opened 2022-12-21
Use custom query function to replace missing taffydb query features

ahocevar:
Fixes #14397

PR #14402, opened 2022-12-22
Fix build index

### LINKED ISSUE #14401: What happened to the function "ol.proj.fromLonLat([longitude, latitude])"?
Hi there,

after re-opening my small openlayers prototype after a year I noticed that my code is not running anymore due to problems with the Content Deliver Network (CDN) (HTTP response 403 "Forbidden"). I fixed this issue by replacing the following two lines

```
    <script src="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.1.1/build/ol.js"></script>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.1.1/css/ol.css" type="text/css">
```
with
```
	<script src="https://cdn.jsdelivr.net/npm/ol@v7.2.1/dist/ol.js"></script>
	<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/ol@v7.2.1/ol.css">
```
Next I faced an issue with the OpenLayers API. Weirdly, the function "ol.proj.fromLonLat()" is not defined anymore?! o.O According to the official documentation it should - however - still be usable as before, see here: https://openlayers.org/en/latest/apidoc/module-ol_proj.html#.fromLonLat

**Describe the bug**
Importing OpenLayers (Javascript and CSS) from a CDN and trying to execute "**ol.proj.fromLonLat([11.1297433, 49.4230742])**" in an interactive javascript console (for example the one available in Firefox or Chrome) will lead to the following error

`Uncaught TypeError: ol.proj.fromLonLat is not a function`

Why is this not working anymore, while the other api calls (i.e. "ol.View", "ol.control.Attribution", etc.) are still working fine?

**To Reproduce**
1. Create a file 'index.html'
2. Add the following code in the file:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Quick Start</title>
	
	
	<script src="https://cdn.jsdelivr.net/npm/ol@v7.2.1/dist/ol.js"></script>
	<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/ol@v7.2.1/ol.css">
	
	
  </head>
  <body>
    <div id="map"></div>
    <script>
	alert(ol.proj.fromLonLat([11.1297433, 49.4130742]));
	</script>
  </body>
</html>
```
4. See error message in javascript console (in Firefox: right click - "Inspect"): "Uncaught TypeError: ol.proj.fromLonLat is not a function"

**Expected behavior**
I would have expected to be able to use this method just as before - in this case to see the output of the conversion from Longitude / Latitude to the WebMercator coordinate system. However, an error is thrown instead.

As far as my search went I need to downgrade to a previous version. I will try to do that - however, I would love to be able to use the most recent version of course.

Thank you for your time and help in advance.

Kind regards,
Adam

ahocevar:
Fixes #14401.

The previous fix in #14398 did not restore all of the full build. It looks like I got the old taffydb query wrong there. Going back in the history of that file, it looks like we use `doc.define` in the other conditions anyway, so there should not be a need to filter out docs where `define` is not set.

### DIFF
--- a/config/jsdoc/info/publish.js
+++ b/config/jsdoc/info/publish.js
@@ -25,7 +25,6 @@ exports.publish = function (data, opts) {
       return true;
     }
     return (
-      typeof this.define === 'object' &&
       !['file', 'event', 'module'].includes(this.kind) &&
       this.meta &&
       this.meta.path &&


ahocevar:
If someone wants to create a new patch release with this fix, please go ahead. I won't be available for the next 8 hours or so.

tschaub (approved):

ahocevar:
Thanks, @tschaub. Creating a patch release now.

marcjansen:
Great

PR #14405, opened 2022-12-22
Add rendering test for the full build

ahocevar:
To avoid future regressions with the full build.


ahocevar:
I'm changing this to use a tile we have locally available, to avoid network traffic.

ahocevar:
Done.

marcjansen:
How about also having https://github.com/openlayers/openlayers/pull/14407 What do you think?

marcjansen (approved):
This is great, thanks.

PR #14407, opened 2022-12-22
Add a task to validate files from the build

marcjansen:
This is open for discussion an might help when validating the full build.

What do people think?

Also any input on the actual checking script is greatly appreciated


ahocevar:
@marcjansen Did you see #14405?

ahocevar:
I don't think this adds much benefit. The first fix I made for v7.2.1 would have passed your check here.

marcjansen:
Yeah, fine for me.

PR #14418, opened 2022-12-28
Raster source interpolation

### LINKED ISSUE #14415: `interpolate: false,` settings of sources no longer respected in `ol/source/Raster`
**Describe the bug**
The `interpolate: false,` settings of sources are no longer respected in `ol/source/Raster`.

**To Reproduce**
In v7.1 the `interpolate: false` setting of the source within the RasterSource was always respected as the setting was used to render the source tiles to the view resolutions used by the RasterSource https://codesandbox.io/s/rastersource-7-1-0-kiiw3e

In 7.2 that only happens if `resolutions: null` is specified.  Otherwise the interpolate setting of the source becomes irrelevant as there is no scaling of the tile within the RasterSource, and it is not used when the RasterSource output is rendered to the map https://codesandbox.io/s/rastersource-7-1-0-kiiw3e

Also, regardless of what was specified. unless `resolutions: null` is specified `getData()` now always returns uninterpolated results.

**Expected behavior**
The interpolate setting of the source now also needs to be used when rendering the RasterSource output.  If the RasterSource has multiple sources with differing interpolate settings there would be a conflict which could be avoided by specifying `resolutions: null`.  The documentation for `resolutions` should mention that.

mike-000:
Fixes #14415

Set the `interpolate` property of a RasterSource to that of its first internal source as the RasterSource output image can now be scaled when rendered to the map.


ahocevar (commented):
Not sure I like this approach. Wouldn't it be more straightforward to just let the user configure the raster source with an `interpolate` option? This would avoid the guessing logic and the unintuitive connection with the `resolutions` setting.

mike-000:
Many RasterSources have only one source, and I cannot think of any reason for the interpolate settings of that and the parent to be different.  If there are multiple sources with conflicting interpolation requirements the only way to handle that to to set `resolutions: null` which avoids scaling when rendering to the map and makes interpolation at that step irrelevant.  It would be possible to check all sources for conflicting requirements and set `this.useResolutions_` false to keep that transparent to the user.

ahocevar:
@mike-000. I can think of cases where users want to set the interpolation option of input and output separately. For example, if you want to classify input rgb values into a reduced color palette, but you don't want a blurry output. In this case you'd want your input source with `interpolate: false`, and the output with `interpolate: true`.

mike-000:
That would be needed only if the operation was being done at a resolution which was neither the native resolution of the source nor the view resolution.  If there is no scaling there will be no interpolation regardless of the setting.  Can that happen if there is a single source?  Multiple sources with conflicting native resolutions perhaps, but then the first source could be `interpolate: true`.

ahocevar:
That can happen also with a single input source, when the view resolution is higher than the highest input source resolution.

mike-000:
OK, a user option would be needed for that (and also where retina tiles are used on a low res display) but I think using the source setting as default would better than defaulting to `true`.  And if the sources have conflicting settings and no user option is provided to only use view resolutions.

ahocevar (changes requested):
@mike-000 Don't you agree that reading the docs makes it clear that just having an `interpolate` option on the Raster source, without any default derived from the input source(s), would make for an API that is much easier to understand?

I'm not going to approve this pull request as it is now.

mike-000:
I think having to specify `interpolate: false` once for input and again for output is confusing.  If we have a user option for output which defaults to `true` regardless of what was used by the input then the docs should make it clear that specifying `resolutions: null` allows the input source interpolation settings to be used throughout.

mike-000:
Whichever approach is taken one example which may need attention is canvas Sea Level.  The `interpolate: false` option in the tile source now has little purpose as there would be no interpolation at tile resolution.  It was needed to prevent interpolation of red and green pixels which caused thin lines of sea close to 25.6 meter intervals and a more noticeable one at 3107 meters.  However as the `maxZoom` of the DEM is much smaller than the base map there is now a very blurry appearance when zooming in.

tschaub:
I think we need to review `interpolate` before spreading it around any more.  Ideally we would have a list of example uses cases that were clearly described and as simple as possible.

hweri69:
First of all, many thanks for the brilliant work of all contributors of OL.
Probably I can give you an use case for different interpolate settings: 
For meterological maps, eg temperature, wind speed or what so ever, the rgb values of the input source are representing the value. This source is modified with raster-source manipulation to set an individual colouring sheme. For this case it would be a benefit to have the option of using the source with interpolation: false and setting output interpolation: true, to avoid artefacts in the output due to interpolation of the source.

mike-000:
@hweri69 The input source interpolation setting is always respected.  Currently the output is always interpolated, and the only way to avoid that is to specify `resolutions: null` so the output is not scaled.

hweri69:
@mike-000 Thank you for the quick reply. Anyway I have to admit, I can't follow you. Probably I'm doing something wrong.

```
var mapextent = [-20037508.3428,-20037508.3428,20037508.3428,20037508.3428];
var mapext31287 = [19500,219500,720500,620500];

var map = new ol.Map({
  maxTilesLoading: 256,
  pixelRatio: 1,
  layers: [
    new ol.layer.Tile({
        className: 'grey',
        source: new ol.source.OSM()
      })
  ],
  target: 'map',
  view: new ol.View({
    projection: availproj.EPSG31287,
    center: ol.extent.getCenter(mapext31287),
    extent: mapextent,
    resolutions:[2048,1024,512,256,128,64,32,16,8,4,2,1],
    zoom: 3
  })
});

var xlsource = new ol.source.ImageStatic({
        url: './raster.png',
        interpolate: true,
        projection: availproj.EPSG31287,
        imageExtent: mapext31287
      });

var dingsda1 = new ol.source.Raster({
    resolutions: [1],
    sources: [xlsource],
    operation: function(pixels) {
      var opa = pixels[0][3];
      var r = pixels[0][0];
      var g = pixels[0][1];
      var b = pixels[0][2];
      var pixel;
      pixel = [r,g,b,opa];
      return pixel;
    },
    threads: 0
  });

var xl = new ol.layer.Image({
      zIndex: 2,
      opacity: 0.9,
      source: dingsda1
    });

map.addLayer(xl);
```
The output depends only from interpolate option of the ImageStatic Source, any manipulation of the raster resolution option makes no difference, so if input interpolation is set to false, the output is without any interpolation.

PR #14419, opened 2022-12-28
Add .babelrc and .prettierrc to CodeSandbox export

### LINKED ISSUE #14417: v7.2.2 examples with custom classes not working in CodeSandbox
**Describe the bug**
 Examples with custom classes are not working in CodeSandbox due to transpiling.

**To Reproduce**
Open the Custom Controls example in CodeSandbox
Observe `Class constructor Control cannot be invoked without 'new'`

**Expected behavior**
v7.1.0 included `"@babel/core": "latest"` in the package.json.  This is now missing but appears to be required.

mike-000:
Fixes #14417

Adds an empty `.babelrc` entry to prevent transpiling.
Also adds a `.prettierrc` entry so the linting is preserved when saving changes.

### DIFF
--- a/examples/resources/common.js
+++ b/examples/resources/common.js
@@ -78,7 +78,25 @@
             'index.html': {content: html},
             'main.js': {content: js},
             'package.json': {content: pkgJson},
-            'sandbox.config.json': {content: '{"template": "parcel"}'}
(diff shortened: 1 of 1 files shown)


tschaub (approved):
Thanks, @mike-000

Issue #14421, opened 2022-12-30
Impossible to stop/cancel looping view animations


ericcorriel:
**Describe the bug**
```view.cancelAnimations()``` does not stop looping animations. Specifically, if ```view.animate()``` is wrapped by a function that has itself as a callback (to create a looping animation), the animation can never be stopped because ```view.cancelAnimations()``` did not account for this use case.

**To Reproduce**

1. Go to ```examples/animation.html``` and replace the html with the following to add buttons for a looping animation and ```view.cancelAnimations()```:
```
<div id="map" class="map"></div>
<button id="rotate-left" title="Rotate clockwise">↻</button>
<button id="rotate-right" title="Rotate counterclockwise">↺</button>
<button id="pan-to-london">Pan to London</button>
<button id="elastic-to-moscow">Elastic to Moscow</button>
<button id="bounce-to-istanbul">Bounce to Istanbul</button>
<button id="spin-to-rome">Spin to Rome</button>
<button id="fly-to-bern">Fly to Bern</button>
<button id="rotate-around-rome">Rotate around Rome</button>
<button id="tour">Take a tour</button>
<!-- Two new buttons here -->
<button id="never-ending-tour">Take never ending tour</button>
<button id="cancel-animation">Cancel Animation</button>
```
2. Go to ```examples/animation.js``` and add the following to create a looping animation and a call to ```view.cancelAnimations()```. Notice ```neverEndingTour``` has itself as a callback to create a looping animation.

```
function neverEndingTour() {
  const locations = [london, bern, rome, moscow, istanbul];
  const duration = 2000;
  view.animate(
      { center: london, duration: duration },
      { center: bern, duration: duration },
      { center: rome, duration: duration },
      { center: moscow, duration: duration },
      { center: istanbul, duration: duration },
      neverEndingTour // <-- callback function calling itself creates a looping animation
      );
}

// Event handlers for new buttons 
onClick('never-ending-tour', neverEndingTour);

onClick('cancel-animation', function () {
  view.cancelAnimations();
});
```
3. Load ```animation.html``` in the browser and start any animation (**Spin to Rome**, for example) and then click **Cancel Animation** and notice the animation stops 👍
4.  Click **Take Never Ending Tour**, notice the tour loops 👍
5. Click **Cancel Animation** and notice the tour doesn't stop, it only restarts from the beginning 👎
6. In ```animation.js```, replace the callback ```neverEndingTour``` with ```function(){ flyTo(bern, function () {}) }```
[...]

PR #14423, opened 2022-12-31
adding optional parameter to cancelAnimations to fix use case where c…

ericcorriel:
Addresses issue https://github.com/openlayers/openlayers/issues/14421

Adds an optional parameter ```cancelCallback``` (defaults to ```false```) to ```view.cancelAnimations(cancelCallback)``` that allows for the following use cases:
- preserves current default behavior: *animation series* is canceled but callback function still fires
- adds option to cancel callback function as well, which is useful to cancel looping animations, for example

To demo:
1. Open ```examples/animation.html``` in browser
2. Click **Never ending tour**
3. Click **Cancel Animation** to demo current behavior
4. Notice animation never stops since only the current animation series is canceled and the callback, ```neverEndingTour``` is still called
5. Refresh the page and restart the **Never ending tour**
6. Click **Cancel Animation And Callback**
7. Notice the animation stops

I believe adding this parameter adds conceptual clarity on how to cancel more types of animations while also preserving default behaior.

### DIFF
--- a/examples/animation.html
+++ b/examples/animation.html
@@ -17,3 +17,6 @@ tags: "animation"
 <button id="fly-to-bern">Fly to Bern</button>
 <button id="rotate-around-rome">Rotate around Rome</button>
 <button id="tour">Take a tour</button>
+<button id="never-ending-tour">Take never ending tour</button>
+<button id="cancel-animation">Cancel Animation</button>
+<button id="cancel-animation-and-callback">Cancel Animation And Callback</button>
--- a/examples/animation.js
+++ b/examples/animation.js
@@ -4,6 +4,7 @@ import TileLayer from '../src/ol/layer/Tile.js';
 import View from '../src/ol/View.js';
 import {easeIn, easeOut} from '../src/ol/easing.js';
 import {fromLonLat} from '../src/ol/proj.js';
+import {never} from "ol/events/condition.js";
 
 const london = fromLonLat([-0.12755, 51.507222]);
 const moscow = fromLonLat([37.6178, 55.7517]);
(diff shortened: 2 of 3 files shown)


ahocevar (commented):
Thanks for your contribution, @ericcorriel. I think you've added something that is already possible with the existing API.

The callback you can pass as last argument to `animate()` has a `completed` argument. When `false`, the animation was cancelled, and you can handle that case. See https://codesandbox.io/s/animation-forked-1qv3cv?file=/main.js

ericcorriel:
thx @ahocevar – i didn't think of that! Maybe it's more obvious than I thought since @mike-000 also pointed it out to me [here](https://github.com/openlayers/openlayers/issues/14421#issuecomment-1368192658).

That said, I do think this proposed method is conceptually clearer (certainly to newcomers such as myself) and avoids a function signature mismatch, which could prove problematic down the road if the library is ever ported to a more strongly typed language such as typescript.

Either way, just throwing it out there!

ericcorriel:
@ahocevar also, correct me if I'm wrong, but I don't the current method will work if the function wrapping ```view.animate()``` already takes a parameter (which is the case in my application). See ```neverEndingTour()``` [here](https://codesandbox.io/s/animation-forked-umt032?file=/main.js).

mike-000:
You can embed the `neverEndingTour` call inside a callback instead using it directly https://codesandbox.io/s/animation-forked-e9sl31?file=/main.js

ericcorriel:
@mike-000 nicely done–that'll do it!

I do still have questions about how intuitive that solution is though…Of course I think my proposal is cleaner and more intuitive, but I understand the motivation not to make changes to a large, widely distributed library when a workaround can be found…

ahocevar:
@ericcorriel I think the current API is very clear. We have one callback that gets called when an animation is done, and the callback knows whether the animation was finished or cancelled.

ericcorriel:
@ahocevar 👍

Issue #14389, opened 2022-12-19
TypeError parsing GML (target.addEventListener is not a function)


ghschier:
In the Method **GMLBase.readFeatureElementInternal** the nodevalue is put in a wrapper-object, when the node has attributes.
The actual nodevalue has the key "\_content\_" in the wrapper-object, the attributes are pushed as key-value-pairs.

The "getGeometry"-Method of the Feature doesn't take care about that, it gives back the wrapper instead of the geometry.

This results in a TypeError ("target.addEventListener is not a function") in **Feature.handleGeometryChanged_**, because the
first argument given to the listen-method is not a geometry.

ahocevar:
Can you create a test case that shows the problem?

ahocevar:
Thanks, I see. Do you think you'd be able to create a pull request with a fix?

ghschier:
I solved the problem for me by modifying `Feature.getGeometry`, so that it returns the geometry instead of the wrapper. IMO that would be the best solution.

> Do you think you'd be able to create a pull request with a fix?

I didn't do that before. How do I create a pull request?

ahocevar:
This should be fixed in the format, not in `ol/Feature`. See https://github.com/openlayers/openlayers/blob/main/CONTRIBUTING.md#contributing-code for instructions. Thanks in advance.

PR #14430, opened 2023-01-09
Fix for issue  #14389

ghschier:
Fixes  openlayers/openlayers#14389

Do not use a wrapper-object, if value is a geometry. This causes a TypeError in Feature.handleGeometryChanged_

### DIFF
--- a/src/ol/format/GMLBase.js
+++ b/src/ol/format/GMLBase.js
@@ -5,6 +5,7 @@
 // of GEOMETRY_PARSERS_ and methods using GEOMETRY_PARSERS_ do not expect
 // envelopes/extents, only geometries!
 import Feature from '../Feature.js';
+import Geometry from '../geom/Geometry.js';
 import LineString from '../geom/LineString.js';
 import LinearRing from '../geom/LinearRing.js';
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
Thanks for this fix, @ghschier. Do you think you'd also be able to add a simple regression test for this?

ghschier:
I don't know. I didn't do this before and I don't know how time consuming it is.

ahocevar:
@ghschier In this case it should be neither difficult nor time consuming. Just add a `describe()`  function in `test/browser/spec/ol/format/gml.test.js`, add a GML snippet that fails before your change and passes after, and test for the result. If you look at the other tests in that file, the process should be pretty much self describing.

ahocevar:
@ghschier Do you feel confident adding the test? If not, I can merge as is, but then there's no guarantee it will break again if someone makes other changes.

ghschier:
I do not feel confident, I'm not familiar with the test syntax and do not know how to run the tests.
If you want, I can send you a GML snippet for testing.

ahocevar (approved):

PR #14444, opened 2023-01-19
Store images as canvas for more efficient reuse

### LINKED ISSUE #13362: Android Chrome Performance issue when rendering a large ImageLayer with a static source
**Describe the bug**
Currently, OpenLayer's performance on an Android phone using chrome is much worse than using Firefox when rendering a large ImageLayer with a static source, say a 10,000 x 7,000 pixels image.

**To Reproduce**
1. Open the example project reproducing the issue, which was forked from the static image example, found [here](https://codesandbox.io/s/static-image-forked-prrz5?file=/main.js) on both Chrome and Firefox on Android (Tested on Oneplus 7 Pro running Android 11)
2. Run the example
3. Try pinching to zoom and panning
4. Notice that on Chrome, pinching to zoom will be smooth for a second, and then completely freeze the page for a couple seconds, back and forth, while on Firefox, the freezing doesn't occur

**Expected behavior**
Both Chrome and Firefox to be able to pinch to zoom and pan without freezing every few seconds.

### LINKED ISSUE #14436: Performance drops with many Wms layers
We are seeing drastic, nonlinear performance reduction ("off a cliff") when panning / zooming / rotating an Openlayers map as soon as it contains more than 15~16 Wms image layers, depending on screen size.

With 18 Wms image layers in FullHD on Openlayers 7.2.2, single animation frames can take more than a second. With 17 layers, performance doubles to ~500ms per frame. With 15 layers or less, performance is basically fine (all frames faster than 200ms).

## To reproduce
@ThorstenHi and I have prepared a single-file demo that should show this nicely: https://gist.github.com/werrolf/911bbd428aa453b021d0735547a40f0f

Open this HTML in your browser of choice, wait a few seconds until all layers have finished intial requests, then zoom in one step or try panning.

You can step down the number of layers by adding e.g. `?layers=15` to your url.

Because this definitely depends on the number of pixels rendered, the map is forced to a somewhat FullHDish pixel size.

Note that if you use Chrome Inspector or similar, it should be undocked and ideally placed on a different screen. Otherwise it will improve performance :)

This demo uses the Openlayers 7.2.2 official build from the CDN.

Note that even though these Wms servers are all public access, their operators may not expect high traffic. Please be gentle :)

## Affected versions
This has originally been observed on Openlayers 6.14. Demonstrating this without an official build has proven difficult though.

6.15 and 7.x. Changelogs have mentioned canvas performance / memory usage improvements, so we stepped through these. All in all 7.2.2 is ~10% faster than 6.14. This is impressive in itself, but the nonlinear dropoff remains, and at some point map interactions still get too slow for reasonable use.

As for browsers, I've tried Chromium 83 up to Google Chrome 108 and also Firefox 108. There was no discernable difference unfortunately.

## Is this a bug?
Probably not. Chrome performance capture shows that two thirds of the CPU time is spent not in Openlayers code, but in the [canvas drawImage method](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage), and more than half of that time goes into decodeImage. The nonlinear aspect is highly suspect though.

We are somewhat hopeful that there are some ways in which the opaque canvas API may be used in a more efficient way. I am totally not an expert on canvas :]

We would also appreciate any guidance on how to push the performance cliff further out with different Openlayers usage on our end :)

ahocevar:
Fixes #14436.
Probably also fixes #13362.

This pull request changes the Image wrapper so it can store the loaded image in the canvas of a provided context. This speeds up `renderFrame()` on maps with many image layers on large screens, because the image does not have to be decoded over and over during `drawImage()` to the target canvas.

To avoid storing the image twice (once as image, once as canvas), the Image wrapper will replace the `image_` with the provided context's canvas on the first `getImage()` call. To avoid creating a new context whenever the extent changes, which is problematic on iOS due to the limited canvas memory, the sources that use the Image wrapper get a new `context_` member, so only a single context is created.

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -34,6 +34,9 @@ class ImageWrapper extends ImageBase {
    * @param {string} src Image source URI.
    * @param {?string} crossOrigin Cross origin.
    * @param {LoadFunction} imageLoadFunction Image load function.
+   * @param {CanvasRenderingContext2D} [context] Canvas context. When provided, the image will be
+   *    drawn into the context's canvas, and `getImage()` will return the canvas once the image
+   *    has finished loading.
    */
   constructor(
     extent,
@@ -41,7 +44,8 @@ class ImageWrapper extends ImageBase {
     pixelRatio,
     src,
     crossOrigin,
-    imageLoadFunction
+    imageLoadFunction,
+    context
   ) {
     super(extent, resolution, pixelRatio, ImageState.IDLE);
 
@@ -59,6 +63,12 @@ class ImageWrapper extends ImageBase {
     if (crossOrigin !== null) {
       this.image_.crossOrigin = crossOrigin;
     }
+
+    /**
+     * @private
+     * @type {CanvasRenderingContext2D}
+     */
+    this.context_ = context;
 
     /**
      * @private
@@ -84,6 +94,17 @@ class ImageWrapper extends ImageBase {
    * @api
    */
   getImage() {
+    if (
+      this.state == ImageState.LOADED &&
+      this.context_ &&
+      !(this.image_ instanceof HTMLCanvasElement)
+    ) {
+      const canvas = this.context_.canvas;
+      canvas.width = this.image_.width;
+      canvas.height = this.image_.height;
+      this.context_.drawImage(this.image_, 0, 0);
+      this.image_ = this.context_.canvas;
+    }
     return this.image_;
   }
 
--- a/src/ol/renderer/canvas/ImageLayer.js
+++ b/src/ol/renderer/canvas/ImageLayer.js
@@ -44,7 +44,7 @@ class CanvasImageLayerRenderer extends CanvasLayerRenderer {
    * @return {HTMLCanvasElement|HTMLImageElement|HTMLVideoElement} Image.
    */
   getImage() {
-    return !this.image_ ? null : this.image_.getImage();
+    return this.image_ ? this.image_.getImage() : null;
   }
 
   /**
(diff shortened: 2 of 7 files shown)


ahocevar:
The changes in [src/ol/renderer/canvas/ImageLayer.js](https://github.com/openlayers/openlayers/pull/14444#diff-d63a4fd4c80f2dafc266d6c1a32961c8adca7a4777e1b94d7642542b930aae19) are not necessary for the proposed fix, but I added them for more clarity during the investigation phase before making the changes for this pull request.

tschaub:
Nice work, @ahocevar.

Here are the before/after examples (extracted from #14436)
 * [before](https://codesandbox.io/s/issue-14436-before-pr-1444-i2m9gb) ([full size](https://i2m9gb.csb.app/))
 * [after](https://codesandbox.io/s/issue-14436-after-pr-1444-bu1t59) ([full size](https://bu1t59.csb.app/))

tschaub:
I think this is a good approach given the current image source/layer/renderer setup.  As you mention, it avoids hanging on to both an image and a canvas element for each image.

However, this approach is specific to the canvas renderer.  In the future, when the webgl renderer renders image layers, it could be nicer to adopt an approach similar to the (evolving) tile source/layer/renderer setup.  In the future setup, sources would not hang on to copies of the image (whether a tile or a non-tile).  The source just generates the tile/image when asked.  The renderer knows about what format it makes sense to cache (e.g. a canvas element or a webgl texture), so it manages the cache.

tschaub (approved):

werrolf:
The performance difference is staggering. >= 6x faster with profiling instrumentation on, so realisitically probably even better if you don't try to measure it :)

Many many thanks @ahocevar :)

PR #14445, opened 2023-01-20
WebGL vector tile renderer

jahow:
This PR is a first attempt at getting a webgl-based vector tile renderer running. The approach is as follows:
* Rely on `MixedGeometryBatch` and `BatchRenderer` used in the webgl vector renderer; here, each tile will have its own geometry batch renderer separately
* Move as much logic as possible from `TileTexture` and `WebGLTileLayerRenderer` into base classes, so as to reuse logic for vector tiles; these base classes were called `BaseTileRepresentation` and `WebGLBaseTileLayerRenderer`
* Create a new `WebGLVectorTileLayerRenderer` which tries to do as few as possible, relying on the geometry batch utilities and base classes to draw tiles on screen

See here: https://deploy-preview-14445--ol-site.netlify.app/en/latest/examples/webgl-vector-tiles.html

Peek 25-01-2023 09-47

Known limitations:
- no alpha transition
- no text rendering

Remaining:
- [x] add rendering tests
- [x] add dedicated example

Discussion: https://github.com/openlayers/openlayers/discussions/14446

*This work is sponsored by BSW-Hamburg and Landesbetrieb Geoinformation und Vermessung Hamburg.*


jahow:
This is ready for review!

ahocevar:
I haven't looked at the code yet, but a first glance at the [example](https://deploy-preview-14445--ol-site.netlify.app/en/latest/examples/webgl-vector-tiles.html) in Safari 16.3 and 16.4 suggests that there are still a few problems:


https://user-images.githubusercontent.com/211514/220097901-4fdfd0d8-1c1d-4936-8865-d9afc5013ec8.mov

jahow:
> I haven't looked at the code yet, but a first glance at the [example](https://deploy-preview-14445--ol-site.netlify.app/en/latest/examples/webgl-vector-tiles.html) in Safari 16.3 and 16.4 suggests that there are still a few problems:

Thanks for the feedback, there was indeed an issue with pixel ratios > 1, also fixed an issue related to tile gutters not being correctly cropped.

jahow:
> edit: I still have issues on mobile... investigating there's a precision issue when specifying a render extent for a tile on older devices; i'll push a separate fix on the branch

This is fixed in the latests commits, I added automatic precision detection as per [MDN indications](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#be_precise_with_glsl_precision_annotations).

Note that on devices which do not support `highp` precision for floats, tile gutters will not be cropped. This should concern very few devices and still allow showing something, even though it might have glitches; having a proper fallback for these devices isn't trivial so I'm leaving this aside for now.

tschaub:
Thanks for your work on this, @jahow.  It is encouraging to see this coming together.  As part of updating the canvas tile renderer (in #14262), I discovered a duplicate tile rendering issue in the webgl tile renderer.  I finally got around to correcting that in #14550.  I think it makes sense to get that fix in first so this work doesn't inherit the same problem.

After #14550 is in, it would be good if we could break this down into more easily reviewable changes.  Would you be willing to put up a branch that separates out the webgl tile renderer and tile texture logic into shared base classes (without TileGeometry and VectorTileLayer classes)?

jahow:
> After https://github.com/openlayers/openlayers/pull/14550 is in, it would be good if we could break this down into more easily reviewable changes. Would you be willing to put up a branch that separates out the webgl tile renderer and tile texture logic into shared base classes (without TileGeometry and VectorTileLayer classes)?

That sounds good! I'll keep that PR opened and open a new one specifically for the tile renderer/tile representation indirection

mike-000:
> See here: https://deploy-preview-14445--ol-site.netlify.app/en/latest/examples/webgl-vector-tiles.html

The CodeSandbox configuration currently used by the examples does not support the `?.` operator.  It can be seen working in some other CodeSandbox examples found via Google search, so with some configuration changes it should be able to support this valid example.

jahow:
> Would you be willing to put up a branch that separates out the webgl tile renderer and tile texture logic into shared base classes (without TileGeometry and VectorTileLayer classes)?

Done in https://github.com/openlayers/openlayers/pull/14594

jahow:
This has been rebased on main.

tschaub:
This looks great, @jahow.  Thanks for your continued work on it.

In working on the particle flow renderer (#14491), I noticed some issues with alpha blending and had to adjust some things to avoid flickering and some opacity issues.  I notice some flashing during rendering in this branch and wonder if it could be related.  After this is in, I look forward to working together to making the webgl rendering work well for the various use cases we are targeting.

I'll break #14491 into smaller parts and we can see how they might affect the vector tile rendering after you merge this.

jahow:
Thanks @tschaub for the multiple reviews!

PR #14448, opened 2023-01-20
Add more browsers that support OffscreenCanvas

ahocevar:
Good news, OffscreenCanvas will soon be supported in all mainstream browsers. Firefox already supports it, and it has only recently also landed in Safari Technology Preview.


MoonE (approved):
Thanks, @ahocevar

PR #14476, opened 2023-02-01
New isVisible() and getAttributions() methods on Layer

### LINKED ISSUE #14453: Make `collectSourceAttributions_` of `ol/control/Attribution` accessible for custom attribution controls
To write own attribution controls it would be nice if the `collectSourceAttributions_` method would be accessible so the code does not need to get duplicated.

For me the nicest solution would be if the method would be exported as a function that accepts a `FrameState` and returns an object containing the attributions and a collapsible property.

i.e.:
```
/**
 * Collect a list of visible attributions and determines the collapsible state.
 * @param {import("../Map.js").FrameState} frameState Frame state.
 * @return {{
 *   attributions: Array<string>,
 *   collapsible: boolean
 * }} Attribution texts and collapsible state.
 */
export function collectSourceAttributions(frameState) {
// ...
```

The only 'major' code change in the control would then be to check `this.overrideCollapsible_` outside of this function call.

ahocevar:
This pull request adds something that I've been wishing for often when working with layers - an `isVisible(view)` method on the layer. And a `getAttributions(view)` method which makes use of the former, to make it easier to collect attributions from all or some layers of the map. This new method also simplifies the `Attribution` control's `collectSourceAttributions_()` method.

Replaces and closes #14454.
Fixes #14453.

### DIFF
--- a/src/ol/control/Attribution.js
+++ b/src/ol/control/Attribution.js
@@ -5,7 +5,6 @@ import Control from './Control.js';
 import EventType from '../events/EventType.js';
 import {CLASS_COLLAPSED, CLASS_CONTROL, CLASS_UNSELECTABLE} from '../css.js';
 import {equals} from '../array.js';
-import {inView} from '../layer/Layer.js';
 import {removeChildren, replaceNode} from '../dom.js';
 
 /**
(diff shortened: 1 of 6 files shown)


src/ol/control/Attribution.js
```diff
@@ -191,60 +190,22 @@ class Attribution extends Control {
...
-        }
-      }
-    }
+    const collapsible = !this.getMap()
+      .getAllLayers()
+      .find(
+        (layer) =>
+          layer.getSource() &&
+          layer.getSource().getAttributionsCollapsible() === false
+      );
```
simonseyock, line 208:
```suggestion
    const collapsible = !this.getMap()
      .getAllLayers()
      .some(
        (layer) =>
          layer.getSource() &&
          layer.getSource().getAttributionsCollapsible() === false
      );
```

simonseyock:
This looks very handy! I just had some minuscule remarks.

simonseyock (approved):

tschaub (approved):

Issue #14485, opened 2023-02-06
Map weather layers


Bahaa-Fadhil:
### Questions:
Does anyone have experience in developing weather map layers such as [windy.com ](https://www.windy.com/)and [ventusky.com ](https://ventusky.com/) with OpenLayers v7.2 and can share their knowledge with us?

Exs:
![image](https://user-images.githubusercontent.com/32929383/216940889-06ead231-2a9d-4ccf-9288-bc22b7a6a276.png)

jahow:
I know of an experiment done by a colleague of mine, @gberaudo
https://github.com/gberaudo/olwind

Definitely looking forward to better support for "flow-like" layers in OL, especially with the new webgl rendering capabilities.

cc @Tomcariello for common interest!

gberaudo:
Hi, olwind POC basically works with raster data, which is styled:
- in WebGL for the temparature gradients;
- on the CPU for the wind particles.

Time has passed since I wrote that POC. And OL has evolved a lot regarding WebGL and GeoTIFF rasters. It would be interesting to discuss ideas for a wind rendering making use of the underlying features now available in OL.

Also, regarding the data standards / resources / pipelines which are necessary to make it useful for many persons.

jahow:
I think this is a good candidate for a [discussion](https://github.com/openlayers/openlayers/discussions).

Bahaa-Fadhil:
Hi @jahow & @gberaudo and thank you for your replay!

Yes, I have seen and tried many of the solutions out there, but none of them work like the example above. Many that these projects are either out of date or do not work with OpenLayers 7.2.

I think that it is a good feature for OpenLayers and will add good value to the OpenLayers library, since it is well-developed and has good documentation compared to other maps!

tschaub:
I have been experimenting with a branch implementing a particle flow renderer for some time.  I just put up a draft pull request that links to [a wind example](https://deploy-preview-14491--ol-site.netlify.app/en/latest/examples/wind.html).  Still plenty of things to discuss and work out.

I hadn't seen @gberaudo's work when I started on that.  Will be interesting to dig in more and compare implementations.

Bahaa-Fadhil:
Hi @tschaub, I see you did a great job here with the new layer. I hope that idea comes to fruition in the future, then it will attract more users to OpenLayers. I already know many people who want to build maps with many weather layers, such as windy.com's weather map forecast.

Not only that, but I hope that we can use a lot of open API data and forecast model such as:

[...]

PR #14491, opened 2023-02-06
Particle flow renderer

tschaub:
This adds a new layer type for rendering velocity fields.  The implementation is inspired by and borrows from the excellent description of the technique described by [Vladimir Agafonkin](https://blog.mapbox.com/how-i-built-a-wind-map-with-webgl-b63022b5537f) for building a wind map.

The [wind example](https://deploy-preview-14491--ol-site.netlify.app/en/latest/examples/wind.html) shows the new "flow" layer (final names TBD).  The layer takes a style object similar to the other WebGL layers, and the example demonstrates a linear color ramp using the new `speed` expression.  The source for the flow layer is a data tile layer where pixel values are assumed to represent x and y velocities, and the layer renders particles moving through the velocity field.  The example still does quite a bit of work to smoothly interpolate the coarse resolution input data.  I haven't yet decided how much flexibility we should provide in terms of input data types and expressions for converting input data to velocities.


jahow:
Impressive work!! I had pretty much the same idea, using style expressions to compute the velocity field. Other parameters could be used, e.g. to customize the particle colors, lifetime, density etc.

I was also hoping to look more seriously into using WebGL 2, specifically because of the limitations mentioned in Vladimir's blog post:

> A new GL feature called [compute shaders](https://www.khronos.org/opengl/wiki/Compute_Shader) (in OpenGL ES 3.1 and equivalent WebGL 2.0 specifications) allows you to run shader code on arbitrary data (without any rendering). Unfortunately, support for the new specifications across browsers and mobile devices is marginal, and so we’re left with only one practical option: textures.

The flow particles use case might very well be the "killer app" for taking this bridge.

Thanks for sharing!

Tomcariello:
Following up on your comment from [14485](https://github.com/openlayers/openlayers/issues/14485#issuecomment-1443870684), here are some links to sample data that we are looking to implement in our application. These geoJSON files reflect our current approach to the data structure; we can re-format if required/preferred.

There are a number of files which represent different swaths (each swath represents a different time - there is more info in the [readme](https://nasa-gibs.github.io/gibs-web-examples/data/vector-samples/ascat/readme.md)). Each file is in directory `https://nasa-gibs.github.io/gibs-web-examples/data/vector-samples/ascat/`

```
ascat_20200101_005400_metopb_37810_eps_o_250_3202_ovw.l2.json
ascat_20200101_023600_metopb_37811_eps_o_250_3202_ovw.l2.json
ascat_20200101_041800_metopb_37812_eps_o_250_3202_ovw.l2.json
ascat_20200101_060000_metopb_37813_eps_o_250_3202_ovw.l2.json
ascat_20200101_073900_metopb_37814_eps_o_250_3202_ovw.l2.json
ascat_20200101_092100_metopb_37815_eps_o_250_3202_ovw.l2.json
ascat_20200101_110300_metopb_37816_eps_o_250_3202_ovw.l2.json
ascat_20200101_124500_metopb_37817_eps_o_250_3202_ovw.l2.json
ascat_20200101_142700_metopb_37818_eps_o_250_3202_ovw.l2.json
ascat_20200101_160600_metopb_37819_eps_o_250_3202_ovw.l2.json
ascat_20200101_174800_metopb_37820_eps_o_250_3202_ovw.l2.json
ascat_20200101_193000_metopb_37821_eps_o_250_3202_ovw.l2.json
ascat_20200101_211200_metopb_37822_eps_o_250_3202_ovw.l2.json
ascat_20200101_225100_metopb_37823_eps_o_250_3202_ovw.l2.json
```

Please let me know if you have any questions/feedback or if there is anything that I can do to assist in this effort. Very excited to see where this goes! Thanks!

Tomcariello:
I appreciate the effort that has gone into this already & I feel bad bumping this thread, but we are revisiting particle flow in our project & need to find a path forward in the coming months.  Is there a roadmap to get this over the finish line & into OL? Is there room for me/others to assist towards the remaining challenges? 

Thanks.

tschaub:
Thanks for the motivation, @Tomcariello.  I resurrected the branch and got it working against the latest changes in the rest of the lib.  So [the example](https://deploy-preview-14491--ol-site.netlify.app/en/latest/examples/wind.html) is up and running again.

My last effort stalled out after I started working on a utility source for interpolating a vector field given point data.  I wanted to do the interpolation (and data tile generation) in a worker, but this would require a lot of redundant code in the worker for loading and parsing feature data.  In some future version of the library, we may have things set up to always load and parse data in a worker - and then could have smaller specialized workers to operate on that data before generating tiles.  Instead of waiting for that future, I think I'll back out of the idea of including the interpolation source here.

I still want to do a couple things to change how the flow source is configured and how it handles different data tile types.  I'll try to carve out some time in the coming days to do this.  If I don't have these changes together in a week, let's discuss alternatives for getting this in.

tschaub:
I've marked this as ready for review now.  I've removed the `@api` annotation from the flow layer, meaning it is "experimental" and we can continue to have it evolve without issuing a major release.

The flow layer works with gridded velocity data, so it needs to be configured with a data tile source with 3 or 4 bands in a `Float32Array`.  Band 1 is the eastward velocity component (`u`) and band 2 is the northward velocity component (`v`).

Most of the example code involves interpolating the 360x180 png-encoded velocities into data tiles.  This logic will be different for different data sources.

I put together [an example](https://github.com/tschaub/openlayers/blob/interpolation/examples/point-interpolation.js) using the ASCAT data, but the interpolation method that I used there isn't really that performant.

I'd like to make it simpler to work with point data.  I can imagine doing this by triangulating the point dataset, using those as vertices when rendering, and letting the gpu interpolate the velocity as varyings.  This won't be nice for a very coarsely sampled dataset, but it could work well for a regular gridded set of points.  At some point I may experiment with this.  If it works, the layer could be updated to accept a data tile source or a vector data source.  But for now I think it makes sense to get this in and see if others can make use of it.

As mentioned in the example, the source must be configured with `transition: 0`.  There is an issue with alpha blending that I will work on addressing later (see #16119).

In addition, I'd like to simplify the post-processing pass logic (see #16120).  This may end up yielding a solution for the alpha blending issue above.

ahocevar (approved):
Impressive work, @tschaub. Thanks!

robitaillej:
Thanks a lot for this amazing work. I got it to work (still some tweaks to make it better) with transferring data from a remote GeoServer with the WCS protocol. 
The thing is that I'm not sure if currently there is a possibility to increase the size of the particles? They are rather small by default, and we would prefer having bigger, fewer of them. 
Thanks for the update.

Issue #14508, opened 2023-02-14
non-georeferenced COG-compatible TIFF not loading in 7.2.2


gravatite:
**Describe the bug**
I have TIFFs that are compatible with the COG format (tiled, overview layers, directory structure at the beginning) but are without georeference information. I would like to use these directly with the GeoTIFF layer to get zoomify-like functionality in a projectionless space.  I found issue #12701 where this was made to work, and then I found another issue #13564 where this broke and was fixed again.  With OL 7.2.2, I get an error with the example code from those two referenced issues using WebGLTile, and I get a blank map using just Tile.

**To Reproduce**
I used the code from the earlier referenced issue to see if the problem was just with my TIFF, but I get the same results.  I can get the OL COG example to work with Tile, so I at least know geotiff.js and its dependencies are loading.  Here is the non-working code:

  ```
const extent = [0, 0, 13470, 8670];
  
  const projection = new Projection({
    code: 'custom',
    units: 'pixels',
    extent: extent,
  });
  
  const map = new Map({
    target: 'map',
    layers: [
      new TileLayer({
        source: new GeoTIFF({
          convertToRGB: true,
          sources: [
            {
              url: 'https://tmp-asger-public.s3.eu-central-1.amazonaws.com/2017_84_40_1_0015_00008062.cog.tif',
              nodata: 0,
            },
          ],
        }),
      }),
    ],
    view: new View({
      projection: projection,
      center: getCenter(extent),
      extent: extent,
      zoom: 1,
    }),
  });
```

**Expected behavior**
A zoomify-like view of the TIFF through the geotiff.js tiled/pyramid support.

mike-000:
The code is using an incorrect property for image height and consequently calculating tile height as `NaN`.  #14511 will fix this.

gravatite:
That seems to work!  I applied the patch from #14511 to my local copy, and it came right up with both the sample code from the older ticket and using my own local TIFF.   I will watch the other ticket so I can see when the fix gets merged in.  Thanks for your help!

tschaub:
This was fixed by #14511 (thanks @spoyd).

PR #14511, opened 2023-02-14
issue-14510 use correct GeoTIFF value for image height in getResolutions

spoyd:
This commit fixes Issue 14510.


mike-000:
This will also fix #14508

tschaub (approved):

tschaub:
Thanks for this fix, @spoyd.  I added an additional commit that makes use of the `image.getWidth()` and `image.getHeight()` functions.

Issue #14515, opened 2023-02-18
The performance of remove Snap interaction is poor


XingGuoZ:
**Describe the bug**
The performance of remove Snap interaction is poor, resulting in obvious page lag

**To Reproduce**
1. My project has several Vector Source and Vector Layer, when editing one of these Layers, need to add Snap Interaction
2. When switching to edit another layer, remove Snap Interaction and add a new Snap Interaction(The purpose of this is to reset the Snap source). 
3. The remove operation is slow to respond, and the page will be noticeably stuck.

**Expected behavior**
I expected it responds quickly. It's best not to feel the page stutter

PR #14516, opened 2023-02-18
Improve the performance of remove Snap interaction

XingGuoZ:
#14515 This commit fixes Issue 14515.
When the Snap Source has 13,000 features，the time of remove Snap Interaction will reduced from 6.5s to 150ms

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -244,14 +244,6 @@ class Snap extends PointerInteraction {
   }
 
   /**
-   * @param {import("../Feature.js").default} feature Feature.
-   * @private
-   */
-  forEachFeatureAdd_(feature) {
-    this.addFeature(feature);
-  }
-
-  /**
    * @return {import("../Collection.js").default<import("../Feature.js").default>|Array<import("../Feature.js").default>} Features.
    * @private
    */
@@ -411,7 +403,7 @@ class Snap extends PointerInteraction {
           )
         );
       }
-      features.forEach(this.forEachFeatureAdd_.bind(this));
+      features.forEach((feature) => this.addFeature(feature));
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Snap.js
```diff
@@ -381,7 +381,8 @@ class Snap extends PointerInteraction {
     if (currentMap) {
       keys.forEach(unlistenByKey);
       keys.length = 0;
-      features.forEach(this.forEachFeatureRemove_.bind(this));
```
ahocevar, line 384: The `forEachFeatureRemove_` method can now be removed - this was the only call site.

src/ol/interaction/Snap.js
```diff
@@ -381,7 +381,8 @@ class Snap extends PointerInteraction {
...
+      this.rBush_.clear();
+      features.forEach(this.unlistenFeatureChange_.bind(this));
```
ahocevar, line 385: You could speed this up even more by getting rid of the new `unlistenFeatureChange_` method and this change:
```suggestion
      Object.values(this.featureChangeListenerKeys_).forEach(unlistenByKey);
      this.featureChangeListenerKeys_ = {};
```

ahocevar (changes requested):
Thanks for your contribution, @XingGuoZ. Can you please address my suggestions. A bonus would be if you could add a test too.

src/ol/interaction/Snap.js
XingGuoZ: The` forEachFeatureRemove_` method has been removed

src/ol/interaction/Snap.js
XingGuoZ: Nice! It has been modified in this way

XingGuoZ:
> Thanks for your contribution, @XingGuoZ. Can you please address my suggestions. A bonus would be if you could add a test too.

Just now, I learned about `test`. Please let me know if there is any problem, thank you!

ahocevar (approved):
Thanks, @XingGuoZ. I took the liberty to complete the test.

PR #14524, opened 2023-02-20
Update ESRI World Basemap url in examples

mike-000:
The ESRI World Basemap v1 url is returning 400 Invalid URL.  Update it to the v2 url.


ahocevar (approved):
Thanks, @mike-000

PR #14531, opened 2023-02-23
Fix a padding setter bug of ol.View when using useGeographic()

kikuchan:
Setting `padding` of `View` instance after creation corrupts center coordinates when using `useGeographic()`.
```js
useGeographic();
const view = new View({
  center: [135, 35],
  zoom: 6,
});

const center = view.getCenter();
// [ 135, 34.999999999999986 ]

view.padding = [0, 0, 0, 0];

const point = view.getCenter();
// [ 0.001212725633561354, 0.0003144103494321371 ]
```
This should be `center == point` roughly because there is no change about the padding.

This PR fixes the bug, and includes a test for it.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -507,7 +507,7 @@ class View extends BaseObject {
   set padding(padding) {
     let oldPadding = this.padding_;
     this.padding_ = padding;
-    const center = this.getCenter();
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @kikuchan.

Issue #14542, opened 2023-02-28
UTFGrid wrapX anomoly


mike-000:
**Describe the bug**
As can be seen in the UTFGrid example the imagery does wrap worlds in a global projection.  However `ol/source/UTFGrid` subclasses `ol/source/Tile` without specifying `wrapX`, and `getWrapX()` returns false, which is to be expected given the default value.  Data and attributions are not handled in wrapped worlds, but that could be easily fixed to match imagery if `getWrapX()` reflected the behavior of the imagery.

**To Reproduce**
https://openlayers.org/en/latest/examples/utfgrid.html
The imagery wraps, data could be accessed in wrapped worlds using normalised coordinates but isn't, and the attribution control is only hidden when the normal world is not visible (although to be consistent with #14023 the control should never be shown for an empty string).

**Expected behavior**
Consistent behavior reflecting `getWrapX()` for imagery, data and attributions.  It could be made configurable.

PR #14543, opened 2023-02-28
Handle attributions in wrapped worlds and reprojected sources

### LINKED ISSUE #14526: Attribution functions in wrapped worlds and reprojected sources
**Describe the bug**
The Bing source attribution function does not handle wrapped worlds or reprojection.  The TileJSON attribution function also does not attempt to handle wrapped worlds.

**To Reproduce**
Open https://openlayers.org/en/latest/examples/bing-maps.html and select the aerial source, expand the attribution control and observe the attribution.  Zoom out, observe it change, then pan to an adjacent wrapped world and zoom back into London.  Observe only the terms of use are now displayed.  This could also affect TileJSON and UTFGrid sources.

When reprojected to EPSG:4326 the Bing attributions applicable when zoomed in to the Gulf of Guinea are displayed for all locations and resolutions.  As well as the view projection not being used to transform the bbox, the resolution used to calculate the tile grid zoom is incorrect but this could be best guested by using the point resolution of the view center in the source projection.

**Expected behavior**
Handle wrapped worlds.  Although reprojection can use tiles from multiple tile grid zoom levels a best guess as to which is being used can be made.

mike-000:
Fixes #14526

Addresses #14542 by setting `wrapX; true` in `super()`.  This could be made configurable if wrapping of the imagery can be controlled.

For Bing sources the calculation of source tilegrid resolution in reprojected sources is based on the **Projection and Scale** example.


ahocevar:
Before you're digging too deep into this @mike-000, the code could be simplified significantly by looking at the used tile extents, instead of doing transformation gymnastics with the view extent.

mike-000:
All the code changes are done.  Additional tests are still be needed for Bing attributions, very similar to the TileJSON tests, but choosing some with limited zoom ranges.

mike-000:
The view extent is not transformed, only normalised.  Attribution extents which were previously transformed to the source projection are now transformed to the view projection as these can be different.  The source resolution of reprojected sources is obtained from the method used by reprojection, other methods such as used in the Projection and Scale example do not reflect how tiles are selected for reprojection.

mike-000:
> the code could be simplified significantly by looking at the used tile extents, instead of doing transformation gymnastics with the view extent

The reproj tile extents would not be useful - if the reproj tiles held a list of source tile coordinates those could be useful.  Reproj tiles do temporarily maintain an array of source tiles (the full objects) but that is cleared when no longer required.

Issue #14544, opened 2023-03-01
Change how viewNotMoving is calculated for PointsLayer


jeremy-smith-maco:
**Describe the bug**
I was implementing heatmaps that change with live values in our project and it would seemingly randomly stop updating with live values and when moving features around it would not update too. I did a bunch of debugging and such and found out that in the `PointsLayer` class, the `viewNotMoving` variable in the `prepareFrameInternal` method was being calculated as `false` every render. 

The root of the cause was to do with my interactions as they were calling handleUp methods of other interactions which calls `endInteraction` on the `View` of the map even though the `beginInteraction` was not called which causes the `viewHints[ViewHint.INTERACTING]` count to go negative. The `View` class handles negative values in the `getAnimating` and `getInteracting` methods by checking if the count is above 0.

https://github.com/openlayers/openlayers/blob/main/src/ol/renderer/webgl/PointsLayer.js#L519

Obviously I should just handle the interactions better but I think when calculating the `viewNotMoving` variable, it should either use the result of the `getAnimating` and `getInteracting` methods and put them into the `frameState`, or instead of just checking `!frameState.viewHints...` to instead check if they are <= 0. It could potentially save future people from having to dig down the rabbit hole, or maybe users should just handle their interactions properly. Not sure.

**To Reproduce**
1. Have heatmap layer
2. Call `endInteraction` on the map's view
3. Heatmap no longer updates

Example: https://codesandbox.io/s/snap-forked-2em51i?file=/main.js
(although in this example it also seems to break the modify as well, will re-render when you move the map though)

**Expected behavior**
The heatmap should continue to update even when the interaction count is a negative number.

ahocevar:
@jahow Can you take a look at this please?

jahow:
@jeremy-smith-maco I submitted a PR which should fix the issue in your codesandbox. Thanks a lot for the preliminary research!

jeremy-smith-maco:
Much appreciated, thanks guys!

PR #14545, opened 2023-03-01
View / safeguard against negative hint values to avoid unexpected behaviour

jahow:
This could happen e.g. in case `view.endInteraction()` was used too many times, which then might cause unexpected behaviours in dependent logic such as the webgl points renderer.

Fixes https://github.com/openlayers/openlayers/issues/14544

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1854,8 +1854,10 @@ class View extends BaseObject {
    * @param {import("./coordinate.js").Coordinate} [anchor] The origin of the transformation.
    */
   endInteractionInternal(duration, resolutionDirection, anchor) {
+    if (!this.getInteracting()) {
+      return;
+    }
     this.setHint(ViewHint.INTERACTING, -1);
-
     this.resolveConstraints(duration, resolutionDirection, anchor);
   }
 
--- a/test/browser/spec/ol/View.test.js
+++ b/test/browser/spec/ol/View.test.js
@@ -2010,6 +2010,13 @@ describe('ol/View', function () {
       view.endInteraction();
       expect(view.getHints()[1]).to.be(0);
     });
+
+    it('does not allow hint value to become negative', function () {
+      view.beginInteraction();
+      view.endInteraction();
+      view.endInteraction();
+      expect(view.getHints()[1]).to.be(0);
+    });
   });
 
   describe('#getConstrainedZoom()', function () {


ahocevar (commented):
thanks @jahow for taking a stab at this. It looks like your fix is a cure for the symptoms, but not for the root of the problem. In the first place. the view hints should never have a negative value. I'm not sure if it is enough to change `endInteractionInternal()` to not go below zero, or if the observed behavior is a sign of a more severe problem.

For now, it would probably be ok to safeguard the hints against going below zero, but I'd sleep better if interaction handling would work in a way that the hints would never go below zero without a safeguard.

jahow:
IIRC the `endInteraction` method wasn't meant to be called from the application code, but only used by actual interactions such as zoom, pan etc. to signal to the view that it could go back to a stable state. Maybe a better, non public API should be used for this? I don't have much ideas for now.

I agree that this fix isn't great -- especially because other parts of the lib also rely on view hints being specifically zero instead of zero and less.

ahocevar:
Ah, I see now! I hadn't read the original issue closely enough. So for this case, I think it is enough to safeguard `endInteraction()` against going below zero. I agree that it shouldn't be an API method, but now that it is, that would be the most sensible fix I think.

tschaub (approved):

jahow:
I modified the PR to add the safeguard at the `View` level which is indeed a much more generic fix.

ahocevar (commented):
I think a better place to add a safeguard would be the API method `endInteraction()`, by means of checking for `getInteracting()` . Internally, I hope/think we do things correctly, and we could add a test to ensure that the hint is always `>= 0`. Otherwise I fear this is prone to problems that are hard to debug.

jahow:
@ahocevar went with your suggestion! I think this contributes to make the view API safer to use.

M393:
OpenLayers could output a warning in this case. It's a bug in application code, but maybe shouldn't be ignored silently.

ahocevar (approved):
To me, this looks great now. Thanks, @jahow !

jahow:
> OpenLayers could output a warning in this case. It's a bug in application code, but maybe shouldn't be ignored silently.

I think making it more permissive is not unreasonable, people might end up writing custom interactions which use begin/endInteraction liberally and calling endInteraction too many times shouldn't cause any serious issue with this PR.

tschaub:
Apologies for approving without reading the existing discussion.  I think this ended up in a good place.

PR #14550, opened 2023-03-05
Avoid rendering a tile more than once

tschaub:
While reworking the canvas tile layer renderer in #14262, I discovered that the WebGL tile renderer does quite a bit of duplicate rendering.  This was was not apparent until porting the logic to the canvas tile rendering because of the difference in alpha blending – with the canvas renderer, partially transparent tiles were rendered with too much opacity.

The duplication occurs when adding tile textures to the lookup by zoom level.  Previously, we unconditionally added a tile texture to the lookup.  With the change here, we only add it to the lookup if it has not already been added.  This avoids duplicates when considering tiles at alt zoom levels when the target zoom level tiles have not yet loaded.  Previously, if two adjacent tiles at the target zoom level were not yet loaded and shared the same alt tile from a higher zoom level, that alt tile would be rendered twice.  There is also potential for duplication when animating a view transition – tiles from the target extent may be rendered more than once if they also intersect the source extent.

### DIFF
--- a/src/ol/renderer/webgl/TileLayer.js
+++ b/src/ol/renderer/webgl/TileLayer.js
@@ -83,17 +83,41 @@ function depthForZ(z) {
 }
 
 /**
+ * @typedef {Object} TileTextureLookup
+ * @property {Set<string>} tileIds The set of tile ids in the lookup.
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks for digging into this, @tschaub !

PR #14568, opened 2023-03-13
Update CONTRIBUTING.md

### LINKED ISSUE #14567: Contributing page has link to what looks like defunct mailing list
**Describe the bug**
The `CONTRIBUTING.md` page has a link to join the dev mailing list here - https://groups.google.com/g/openlayers-dev - This appears to have not been used properly for a number of years and the most recent messages are all spam.

**Expected behaviour**
Page should not have the link on. If there is a separate place where these kinds of discussions take place, that should be linked to, otherwise it should just be removed.

I am happy to quickly do this if someone can confirm that this list is now defunct and if there is an alternative.

RobQuincey:
Remove reference to defunct mailing list

Fixes #14567


ahocevar (approved):
Thanks, @RobQuincey

PR #14572, opened 2023-03-14
Fix issue with external map example

### LINKED ISSUE #14569: External map example does not render a map
**Describe the bug**
The [External Map example](https://openlayers.org/en/latest/examples/external-map.html) does not render a map in the external window it opens

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the [External Map example](https://openlayers.org/en/latest/examples/external-map.html) 
2. Click on 'Open external map' button
3. A new browser window is opened, but there is no map in it

**Expected behaviour**
Map should be rendered. I believe this is just a CSS issue. The `map` class for the external map has a height of 100%, I believe this could be fixed by simply switching this to use viewport units `vh` and `vw`.

Happy to open a PR for this if you confirm the issue.

RobQuincey:
Sets the map class in the external map example to use vh and vw units, rather than percentage.

Fixes #14569

### DIFF
--- a/examples/resources/external-map-map.html
+++ b/examples/resources/external-map-map.html
@@ -9,7 +9,8 @@
         margin: 0;
       }
       .map {
-        height: 100%;
+        height: 100vh;
+        width: 100vw;
       }
       .map.unusable .ol-mask {
         height: 100%;


MoonE (approved):
Thanks, @RobQuincey

PR #14576, opened 2023-03-16
Automatically update URL of examples page based on search query

### LINKED ISSUE #14574: Update URL on examples page so your search is remembered
When using the [Examples page](https://openlayers.org/en/latest/examples/), I will often type in a keyword in the search box, then click a relevant example. If the example is not what I wanted, I click back, but my search is lost, and I have to re-type the keyword.

You can already include the search in the query string of the URL by adding `?q=search`, e.g. https://openlayers.org/en/latest/examples/?q=layer

I propose that, on typing a search, we use the History API to update the URL to include the 'q=' automatically, so when a user goes to an example and hits 'back', their search is retained. It also makes sharing a specific set of examples related to a keyword easier.

Related to this, I propose that the 'type' of the input is changed from 'text' to 'search'. This allows for easier clearing of the search via either the `ESC` key or the little 'x' that is added by most browsers. I don't think this would have any negative effects.

I have a little proof of concept I'm happy to tidy up into a full PR if you think this is a good idea.

RobQuincey:
This PR updates the examples page to automatically update the URL when typing keywords in the search bar using the History API. This allows users' searches to be remembered when they go back to the examples page, and makes it slightly easier to share the URL of a set of examples.

This also changes the type of the input to `search`, which allows users to more easily clear their search with the 'x' the browser adds to search inputs, rather than having to delete the text manually.

The implementation uses [History.replaceState](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) to continually update the URL without adding lots of items to the history. It will retain additional query parameters and only replaces/removes the 'q' parameter, so if additional parameters are used/added in the future, this will not effect it.

There are probably slightly cleaner/cleverer ways of doing the `updateHistoryState` function I added, but this works and is relatively simple. Suggestions welcome.

Closes #14574

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -52,7 +52,7 @@
       <nav class="collapse navbar-collapse" id="olmenu">
         <form role="search">
           <div class="input-group">
-            <input name="q" type="text" id="keywords" class="form-control search-query" placeholder="Search" aria-label="Examples" aria-describedby="count" autocomplete="off" autofocus>
+            <input name="q" type="search" id="keywords" class="form-control search-query" placeholder="Search" aria-label="Examples" aria-describedby="count" autocomplete="off" autofocus>
             <span class="input-group-text text-white bg-transparent" id="count"></span>
           </div>
         </form>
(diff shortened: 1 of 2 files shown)


MoonE (approved):
Thanks, @RobQuincey

Issue #14582, opened 2023-03-17
Feature Request: Support for GetFeatureInfo on a WMTS layer using template from WMTS capabilities


mbosecke:
I'm using Geoserver as a backend to serve WMTS layers, which are built on the frontend using "optionsFromCapabilities". Within the capabilities document, I see URL templates on how to implement GetFeatureInfo that look like this:

`<ResourceURL format="text/xml" resourceType="FeatureInfo" template="https://mitchell.internal.forcorp.com/geoserver/gwc/service/wmts/rest/public:fma_simplified/{style}/{TileMatrixSet}/{TileMatrix}/{TileRow}/{TileCol}/{J}/{I}?format=text/xml"/>
<ResourceURL format="text/html" resourceType="FeatureInfo" template="https://mitchell.internal.forcorp.com/geoserver/gwc/service/wmts/rest/public:fma_simplified/{style}/{TileMatrixSet}/{TileMatrix}/{TileRow}/{TileCol}/{J}/{I}?format=text/html"/>
<ResourceURL format="application/json" resourceType="FeatureInfo" template="https://mitchell.internal.forcorp.com/geoserver/gwc/service/wmts/rest/public:fma_simplified/{style}/{TileMatrixSet}/{TileMatrix}/{TileRow}/{TileCol}/{J}/{I}?format=application/json"/>`

I would like a feature that helps build GetFeatureInfo URLs, using the templates provided from a WMTS capabilities document.

It looks like this pull request was on the right track but never completed: https://github.com/openlayers/openlayers/pull/3150

mbosecke:
I've added a small bounty to this feature request with a 1 month expiration in case there's anyone who can help: 

https://app.bountysource.com/issues/118117964-feature-request-support-for-getfeatureinfo-on-a-wmts-layer-using-template-from-wmts-capabilities

mike-000:
I don't think this would be a method of the WMTS source as it requires information not available to the source.  It is however easy to deduce the url needed for an info request given the parsed capabilities, the options used to create the source and the format required for the info, together with the coordinate and resolution, in the application using existing API methods. For example replacing the WMS with WMTS in the GeoFeatureInfo example https://codesandbox.io/s/getfeatureinfo-tile-forked-4418ms?file=/main.js

PR #14589, opened 2023-03-20
Add latest typescript to CodeSandbox devDependencies

mike-000:
While adding an empty `.babelrc` in #14419 prevented unwanted transpiling by CodeSandbox after #14279 it appears that the default parcel template in CodeSandbox uses an old version of Babel for parsing the application JavaScript and as noted in https://github.com/openlayers/openlayers/pull/14445#issuecomment-1475247856 modern syntax such as optional chaining (which would have been useful in the example in https://github.com/openlayers/openlayers/issues/14582#issuecomment-1474852682) is rejected.  If the unneeded `'@babel/core': 'latest',` removed from live example by #14279 is injected into the `package.json` supplied to CodeSandbox it fixes the problem.  This is an external use so it does not need to be an OpenLayers dependency.

### DIFF
--- a/examples/resources/common.js
+++ b/examples/resources/common.js
@@ -56,7 +56,10 @@
       const js = document.getElementById('example-js-source').innerText;
       const workerContainer = document.getElementById('example-worker-source');
       const worker = workerContainer ? workerContainer.innerText : undefined;
-      const pkgJson = document.getElementById('example-pkg-source').innerText;
+      let pkgJson = document.getElementById('example-pkg-source').innerText;
+      pkgJson = JSON.parse(pkgJson);
+      pkgJson['devDependencies']['typescript'] = 'latest';
+      pkgJson = JSON.stringify(pkgJson, undefined, 2);
 
       const unique = new Set();
       const localResources = (js.match(/'(?:\.\/)?(?:data|resources)\/[^']*'/g) || [])


tschaub:
It looks like a few alternatives to adding babel would be to
 * use a Cloud Sandbox (e.g. https://codesandbox.io/p/sandbox/mutable-currying-f98wu6?file=%2Fmain.js)
 * avoid the use of optional chaining (only used in the [`icon-width.js` example](https://github.com/openlayers/openlayers/blob/feb80f804c378f880fcd47c9ea145d141859503b/examples/icon-width.js#L85) currently)
 * add Typescript as a sandbox dev dependency instead of Babel

I'm not sure how much would have to change if we used Cloud Sandboxes instead of Web Sandboxes (I think this is the term they use to differentiate).  Also don't know if Cloud Sandboxes will provide a better or worse experience for people trying to edit our examples (not sure if they will be faster or slower, etc.).

It feels pretty easy to just change the `icon-width.js` example to avoid using optional chaining.  But there could be other things that break the CodeSandbox/Parcel parser as well.

I don't have a strong preference regarding Typescript or Babel, but Typescript feels closer to being a first class dev dependency of ours, so I might choose that over Babel.

mike-000:
> But there could be other things that break the CodeSandbox/Parcel parser as well.

I looks like syntax changes in ES2018 such as `let { x, y, ...z } = { x: 1, y: 2, a: 3, b: 4 };` and later ES versions all cause problems.

Adding `"typescript": "latest"` as a fix works equally as well as `"@babel/core": "latest"`.

tschaub (approved):
Thanks, @mike-000

PR #14594, opened 2023-03-21
WebGL / Introduce a base tile renderer

jahow:
This PR is a refactoring aimed at facilitating https://github.com/openlayers/openlayers/pull/14445 

This factorizes as much logic as possible from the `WebGLTileLayerRenderer` class into a `WebGLBaseTileLayerRenderer` class.

Similarly, some logic from the `ol/webgl/TileTexture` class was factorized into an `ol/webgl/BaseTileRepresentation` class.

This shouldn't affect anything.


src/ol/renderer/webgl/TileLayerBase.js
```diff
@@ -0,0 +1,748 @@
...
+     * @type {import("../../structs/LRUCache.js").default<TileRepresentation>}
+     * @protected
+     */
+    this.tileRepresentationCache_ = new LRUCache(cacheSize);
+
+    /**
+     * @protected
+     * @type {import("../../Map.js").FrameState|null}
+     */
+    this.frameState_ = null;
```
tschaub, line 207: I think we are still consistently using the `_` suffix to signal that a property is private.  Could you rename these protected properties to conform with that?

(At some point we should add a linter rule about this if we want to keep the convention.)

tschaub (approved):
Thanks for separating this out, @jahow.  One minor comment about the private naming convention.  Otherwise this looks good to me.

jahow:
Thanks for the review!

PR #14492, opened 2023-02-07
Use event.key instead of event.charCode

WebKieth:
Thanks for [approving](https://github.com/openlayers/openlayers/issues/14490#issuecomment-1419664753) to change unstable code of interaction. Build Checked, feel free to review =)

PR #14598, opened 2023-03-21
Update deprecated KeyboardEvent properties

mike-000:
In addition to `KeyboardEvent.charCode` updated by #14492 MDN https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#obsolete_properties now lists `KeyboardEvent.keyCode` and `KeyboardEvent.which` as deprecated and recommends using `KeyboardEvent.key` instead.


test/browser/spec/ol/interaction/keyboardpan.test.js
```diff
@@ -31,22 +31,22 @@ describe('ol.interaction.KeyboardPan', function () {
         preventDefault: Event.prototype.preventDefault,
       });
 
-      event.originalEvent.keyCode = 40; // DOWN
+      event.originalEvent.key = 'ArrowDown';
```
tschaub, line 34: Can we import and use `Key.DOWN` here?  Same below.

mike-000: Is that normal for tests?  If so should we not also be using `KEYDOWN`, etc. from `ol/events/EventType` in tests?  (A few tests do use `EventType.CHANGE` - in one case inconsistently using both that and the string - but most do not).

tschaub: Good point.  Small screen and I didn’t pay attention to which file this was in.

tschaub (approved):
Thanks, @mike-000

PR #14607, opened 2023-03-25
Do not use non-API methods in example

ahocevar:
This pull request changes the icon-width example to only use API methods. `Icon#setWidth()` and `Icon.setHeight()` are not API methods.

On a related note, I feel like this example falls more under the category "test" than "example", and would be a candidate for removal. What do others think?


tschaub (approved):

tschaub:
I also agree it could be removed.

Issue #11681, opened 2020-10-23
Line crossing anti-meridian stops rendering when it is clipped


ashchurova:
We use OL to plot vessel movement history on the map, Since OL doesn't support lines crossing anti-meridian (or if it does I am not aware of it) when vessel movement line crosses anti-meridian we adjust coordinates based on direction of the movement, so for ex if it is crossing from west to east  we adjust east hemisphere coordinates by adding -360 to coordinate's longitude, and if line is crossing from east to west we add +360 to the longitude of the west hemisphere coordinates of the line.    

So while panning or zooming line stops rendering on the screen. See attached screen shots. I have added points layer based on line coordinates to help you see where the line should be rendered :

![with_line](https://user-images.githubusercontent.com/15365676/97006444-e3784400-150d-11eb-86a9-f64236480ede.png)
![without line 2](https://user-images.githubusercontent.com/15365676/97006457-e96e2500-150d-11eb-889b-455e9b27a9d0.png)
![without line](https://user-images.githubusercontent.com/15365676/97006470-ed9a4280-150d-11eb-8a51-0e70ef3b2e71.PNG)
 
The sample code that can be used to reproduce this issue can be found here: https://codepen.io/ashchurova/pen/rwXjzP

Thank you.

MoonE:
Your line crosses the date line from the left, if the world this line belongs to is no longer in the current render extent it will not be drawn. To avoid this you can split it at each date line intersection and use a `ol/geom/MultiLineString`.
```js
function splitAtDateLine(coords) {
  const lineStrings = [];
  let lastX = Infinity
  let lineString;
  for (let i = 0, ii = coords.length; i < ii; ++i) {
    const coord = coords[i];
    const x = coord[0];
    if (Math.abs(lastX - x) > 180) { // Crossing date line will be shorter
      if (lineString) {
        const prevCoord = coords[i - 1];
        const w1 = 180 - Math.abs(lastX);
        const w2 = 180 - Math.abs(x);
        const y = (w1 / (w1 + w2)) * (coord[1] - prevCoord[1]) + prevCoord[1];
        if (Math.abs(lastX) !== 180) {
          lineString.push(ol.proj.fromLonLat([lastX > 0 ? 180 : -180, y]));
        }
        lineStrings.push(lineString = []);
        if (Math.abs(x) !== 180) {
          lineString.push(ol.proj.fromLonLat([x > 0 ? 180 : -180, y]));
        }
      } else {
[...]

PR #14608, opened 2023-03-25
Handle wrapped worlds in Box Selection example

### LINKED ISSUE #14605: Box Selection example broken
**Describe the bug**
Drawing a box selection can fail to select the expected geometries.

**To Reproduce**
I found some buggy behavior in the OpenLayers [Box Selection][1] example.

1. translate the map by clicking and dragging to the right so China is on the left and the USA is on the right.

image

2. Hold down the Command (or Ctrl key on other OSs) and drag a selection box around the USA. The selection will work and will look like:

image

3. Do the same, but try drawing the selection box around China. It will not work.

If in (1), one clicks and drags to the left to get a similar view, selecting both the USA and China will work.

I believe what is going on is the extent of the box has had it's latitude shifted by 360deg and that can no longer intersect with the China coordinates which remain in the range of -180deg to 180deg.

**Expected behavior**
The expected geometries which intersect with the selection box should be selected.

mike-000:
Fixes #14605

https://deploy-preview-14608--ol-site.netlify.app/en/latest/examples/box-selection.html

### DIFF
--- a/examples/box-selection.js
+++ b/examples/box-selection.js
@@ -84,7 +84,11 @@ dragBox.on('boxend', function () {
 
     const boxFeatures = vectorSource
       .getFeaturesInExtent(extent)
-      .filter((feature) => feature.getGeometry().intersectsExtent(extent));
+      .filter(
+        (feature) =>
+          !selectedFeatures.getArray().includes(feature) &&
+          feature.getGeometry().intersectsExtent(extent)
+      );
 
     // features that intersect the box geometry are added to the
     // collection of selected features
@@ -127,12 +131,9 @@ dragBox.on('boxstart', function () {
 const infoBox = document.getElementById('info');
 
 selectedFeatures.on(['add', 'remove'], function () {
-  const names = selectedFeatures
-    .getArray()
-    .filter((feature, index, array) => array.indexOf(feature) === index)
-    .map((feature) => {
-      return feature.get('ECO_NAME');
-    });
+  const names = selectedFeatures.getArray().map((feature) => {
+    return feature.get('ECO_NAME');
+  });
   if (names.length > 0) {
     infoBox.innerHTML = names.join(', ');
   } else {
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
This example has a lot of complexity. Previously it was the rotation that made it complicated, now antimeridian handling is added. What if the API of `ol/interaction/DragBox` had a `getGeometries()` method, or the `dragend` event had a `geometries` property, where one geometry for each world is provided? Then this example would not get even more complicated. Later, this example could be further simplified by using turf for geometry level intersections, then nothing would have to be rotated.

mike-000:
That requirement is not specific to this example or DragBox.  A method for converting wrapped geometries into normal world MultiPolygons or MultiLineStrings could be very useful elsewhere. https://github.com/openlayers/openlayers/issues/11681#issuecomment-715583978 has the algorithm for LineStrings, which would need to be adapted for LinearRings.  It should also work with any global projection coordinates as splitting a long straight line in EPSG:4326 would produce a kink when displayed in EPSG:3857.

ahocevar:
@mike-000 Something like a `Geometry#getGeometries(projection)` method? That would indeed be nice! The algorithms wouldn't be trivial though, because multiple antimeridian crossings could be possible. But still simpler than a generic split algorithm. However, I think this is beyond the scope of this pull request.

A simple fix for #14608, without making this example more complicated, would be to simply limit panning to one world on the view.

mike-000:
> A simple fix for #14608, without making this example more complicated, would be to simply limit panning to one world on the view.

That would just be hiding the issue.  It is a legitimate requirement for anyone dealing with the Pacific region.

mike-000:
> What if the API of `ol/interaction/DragBox` had a `getGeometries()` method, or the `dragend` event had a `geometries` property, where one geometry for each world is provided

Even with multiple geometries there would still need to be separate `feature.getGeometry().intersectsExtent(extent)` as we would not want to use the extent of a MultiPolygon, and splitting an extent is not difficult (`wrapAndSliceX` does it but is not API so not suitable for the example).  If it wan't for the complication of `geometry.translate(-world * worldWidth, 0);` when handling rotation the multiple `feature.getGeometry().intersectsExtent(extent)` calls could be done in a single `filter` operation.

ahocevar (approved):
I'm not a fan of examples that try to be full featured applications. In my opinion, an example should show a single piece of functionality, which gives it an educational value. It seems you're not sharing this view, and it seems I'm unable to convince you. Anyway, as an appreciation for your efforts, I'm going to merge this now. But I'll continue to remind you.

Issue #14596, opened 2023-03-21
The map `loadend` event is sometimes not fired in combination with `Heatmap` layer.


simonseyock:
**Describe the bug**

Sometimes the `loadend` event on the map is not fired even though all requests are finished.

**To Reproduce**

The heatmap layer has to have a `maxZoom` that is lower then the initial zoom of the map.

https://codesandbox.io/s/heatmap-earthquakes-forked-mljy3v?file=/main.js

Reload the Browser window in the sandbox to see the effect.

Debugging gave me the following insights:

At the point where `renderFrame_` is called and `this.renderComplete_` would be set to `true`, the `renderer` of the `Heatmap` is not ready (in `getLoadingOrNotReady`). This will cause `this.renderComplete_` to remain `false` and no `loadend` event is fired.

Soon after the renderer changes it state to 'ready' but will not cause a new call to `renderFrame_`. Therefore `this.renderComplete_` stays false and no `loadend` event is fired. This can be tested with a `setInterval`.

**Expected behavior**

`loadend` should be dispatched if all layers finished loading.

Maybe the Implementation of `getLoadingOrNotReady` can be changed to not check for for layers that are not in the current resolution.

Also it might be worthwhile to check why the renderer of the `Heatmap` is not ready in the first place.

ahocevar:
@jahow Can you please take a look at this?

jahow:
Thanks for the reproduction case @simonseyock, I'll look into it

PR #14611, opened 2023-03-26
WebGL Points / mark the renderer as initially ready

jahow:
This simply aligns the WebGL points renderer with other renderers by having its `ready` property set to `true` initially. In the case where the renderer is initially not needed because e.g. the layer is not being rendered, then it will not prevent the `loadend` event of the map to be dispatched.

Also contains a tiny bit of boyscoutting.

Fixes https://github.com/openlayers/openlayers/issues/14596


ahocevar (approved):
Thanks for fixing this, @jahow!

PR #14622, opened 2023-03-29
Clear size when target is detached from map

ahocevar:
This pull request fixes a regression caused by #14305: When detaching a map from a `target` using `map.setTarget(null)` and re-attaching it later using `map.setTarget(target)`, no map will be visible. The reason is that when no target is attached, the logic expects the map's `size` to be unset.


ahocevar:
`.only` removed, tests pass now.

tschaub (approved):

PR #14628, opened 2023-03-31
Helper method to bind a texture

tschaub:
This is one small change from #14491, adding a helper function that prepares a program to use a texture.

The current rendering tests cover this refactor.


jahow (approved):
:ok_hand:

PR #14631, opened 2023-04-01
Upgrade dependencies

bjornharrtell:
Except rollup because I think that requires some changes.


tschaub:
Hi, @bjornharrtell.  I'm curious on the motivation for this change (not that it is unreasonable, just wanted to understand the motivation).

We use dependabot to update dependencies.  Where we use ["caret" dependency ranges](https://nodesource.com/blog/semver-tilde-and-caret/), this generally means that dependabot updates our lockfile but doesn't bump the version in the `package.json` file for minor and patch releases (example #14633, counter-example #14632).  We aren't entirely consistent about where we use caret vs pegged dependencies.

The potential benefit of having our `package.json` reflect the minimum version that we depend on is that it provides more potential for deduplication (in cases where packages that depend on us also depend on a version of a dependency that is lower than the version in our `package.json`).

bjornharrtell:
@tschaub not much more than scratching an itch. I agree with your arguments, especially for a library. I was worried openlayers was lagging behind dependencies but then it's really only rollup that might be worth spending some time upgrading to? I also had a look at geotiff.js which does seem in more need of getting a fair amount of dependencies upgraded.

PR #14656, opened 2023-04-11
Remove experimental description from getUserProjection() API method

mike-000:
This description was removed from other methods when they were added to the API so was presumably overlooked here.

In addition to tile grids user projections to not apply to coordinates in `frameState` and `viewState` which are also part of the API.  Should that be mentioned either here in or in the descriptions for those objects?


ahocevar:
Thanks, @mike-000. I think for `frameState` and `viewState`, that should be mentioned both here and there - to make sure it gets noticed.

ahocevar (approved):
Thanks, @mike-000

PR #14660, opened 2023-04-11
WebGL / Add support for stroke and fill in the ShaderBuilder class

jahow:
This is a preparatory step to have classes like `WebGLVectorLayer` and `WebGLVectorTileLayer` similar to the `WebGLPointsLayer` and `WebGLTileLayer`, taking a style object as input instead of passing lower level shaders and attributes to their respective renderers.

I tried to make the code of the `ShaderBuilder` class as well as the logic for parsing a literal style more readable.

No significant changes, the only potential impacts are:
* the `u_sizePx` uniform which was available in stroke/fill shaders is now called `u_viewportSizePx` for clarity

Note that I also started transcribing the style format in `ol/style/flat` to something equivalent that could be used for the webgl layers. This means that eventually we might drop the current symbol style format for the `WebGLPointsLayer`, but for now this format is still used.


tschaub (approved):

jahow:
Thanks for the quick review!

PR #14672, opened 2023-04-15
Handle Circle geometry resulting from non-parallel transforms

### LINKED ISSUE #14666: Translate interaction with geographic coordinates changes geometry shape
**To Reproduce**
In https://codesandbox.io/s/translate-features-forked-dj5tfn select a state a translate it up or down.  Notice the shape changes with latitude.  That does not happen when not using geographic coordinates.  If using geographic coordinates with a polar view projection the results could be very unexpected. 

**Expected behavior**
As with Draw, Modify and Snap the interaction should work internally with view coordinates and transform any changes back to the user projection (and as with those interactions Circle geometry might need special attention).

mike-000:
Fixes https://github.com/openlayers/openlayers/issues/14666#issuecomment-1509210484

Update Circle geometry to allow for coordinates transformed from a non-parallel projection so the radius is not defined as `[cx + r, cy]` in methods beyond `getRadius`.  `setCenter` and `getExtent` should handle them, while `setRadius` should fully reset them.  Also ensure the `transform` method returns a `Circle` geometry as documented which does not need further typecasting.

Handle non-parallel user projection transforms of Circle geometry in Modify interaction.
Always return `[cx + r, cy]` compliant Circle geometry in user projections in Draw interaction.


mike-000:
To get reliable dragging of a Circle circumference with Modify in a non-parallel user projection the circle extent is calculated using an outer circle which encloses the normal circle as in

image

This does not increase the extent in parallel projections but allows for closest points being detected in non-parallel projections which might be outside the extent produced using points on only the inner circle.

ahocevar:
Can we close this, @mike-000 ?

mike-000:
> Can we close this, @mike-000 ?

No, I also identified significant issues with Circle geometry and the Modify with fully non-parallel projections (i.e. not simply non-parallel meridians as in the Mollweide tests).

ahocevar:
I have no idea what you're talking about, @mike-000. So if you're intending to continue working on this, please start with a rendering test and show a comparison of the expected vs rendered result.

mike-000:
> I have no idea what you're talking about, @mike-000. So if you're intending to continue working on this, please start with a rendering test and show a comparison of the expected vs rendered result. No codesandbox, but a rendering test.

As described in https://github.com/openlayers/openlayers/issues/14666#issuecomment-1509210484

> The modify interaction is also affected by this as in https://codesandbox.io/s/modify-features-forked-k5euf0?file=/main.js where modifying the Circle center changes the radius in view projection and modifying the radius it is not aligned with the pointer.

The codesandbox simply illustrates the issue but browser tests are more appropriate for testing an interaction.

As well as being an internal problem for the modify interaction, more generally a geometry resulting from an interaction should behave as expected, e.g. with geographic coordinates in a UTM view following translation or modification of the center two circles with the same radius would be expected to display the same circumference, but when geographic parallels are non-parallel curves that does not happen

```
const feature = new Feature(new Circle([-20, 50], 5));
const geom = feature.getGeometry();
geom.transform(userProjection, projection);
geom.translate(3000000, 0);
geom.transform(projection, userProjection);
vector.getSource().addFeature(feature.clone());
geom.setRadius(geom.getRadius());
source.addFeature(feature);
```

image

ahocevar:
Like I said already, a `Circle` geometry is always a circle when rendered in the view. This means that, when working with a user projection different from the view projection, that circle won't have the shape of a circle in the user projection, because that's not the rendered projection. So if this is the problem you are trying to explain, it is not a problem. But what do you mean exactly with

> As well as being an internal problem for the modify interaction

If it is impossible to grab the circle for modification on its perimeter (is that what you're referring to?), then the problem is the coordinate system of the Modify interaction's RTree - it is probably because it is the user projection instead of the view projection (haven't checked though).

mike-000:
In that example both have the shape of a circle in the view projection, but with different radius.  Flat coordinates `[10, 50, 10 + Math.sqrt(5), 50 + Math.sqrt(5)]` (which might result from a transform from view projection after a translate in non-parallel view projection) and `[10, 50, 15, 50]` (which would result from setting the radius) are both radius 5 in geographic coordinates, but will have a different radius if viewed in another (non-parallel) projection.

Additionally

`circle.setRadius(circle.getRadius())` should always give same result as `circle.setCenterAndRadius(circle.getCenter(), circle.getRadius())` in any single projection

`circle.setCenter([x, y])` should always give same result as `circle.translate([x - circle.getCenter()[0], y - circle.getCenter()[1]])` in any single projection

`circle.getExtent()` should always give same result as `new Circle(circle.getCenter(), circle.getRadius()).getExtent()` in any single projection

Currently they do not if the flat coordinates are in the form `[10, 50, 10 + Math.sqrt(5), 50 + Math.sqrt(5)]` as those methods to not allow for that possibility, and Modify needs these to work instead as expected of using those workarounds.

> If it is impossible to grab the circle for modification on its perimeter (is that what you're referring to?), then the problem is the coordinate system of the Modify interaction's RTree - it is probably because it is the user projection instead of the view projection (haven't checked though).

It is because the closest point on the circumference is slightly outside the extent of the circle as calculated from 32 points obtained by `ol/Polygon/fromCircle` when transformed from the other non-parallel projection - I discovered that because it broke the radius modification test using `4, -30` (while `6, -30` which was the same distance away worked).  Increasing the number of points to a very large number would fix it, but applying a buffer as illustrated would be more efficient and, I think, more reliable.

ahocevar:
I think calculating a circular polygon from a Circle geometry in the Modify interaction, for the purpose of being able to grab the perimeter, like you introduced in #10343, is not the best solution. Knowing the radius (in view projection), it should be possible to determine whether a circle is grabbable just by checking if the center of any circle is `r` units away (plus/minus the configured tolerance).

mike-000:
Leaving the Circle extent as infinite in the RBush seems to work.  I have noticed that there is also a problem with projections where only the meridians are non-parallel, e.g. drag the Circle center away from the equator in this Mollweide example https://codesandbox.io/s/sphere-mollweide-forked-or8gqu?file=/main.js the view projection radius decreases as it remains a 5 degrees, which is inconsistent with what happens without a user projection, so I am closing this in favour of an approach similar to that now used for Translate.

mike-000:
> I think calculating a circular polygon from a Circle geometry in the Modify interaction, for the purpose of being able to grab the perimeter, like you introduced in #10343, is not the best solution.

More generally I think the internal use of extents in user projection is not a good solution for any geometry type.  An extent in geographic coordinates is meaningless in a polar projection, or where a local projection crosses the antimeridian (e.g. US National Atlas and western Aleutians). The Modify interaction (and vector sources in general) do not work well in those situations.

ahocevar:
I think it does not make too much sense to spent time on this - the primary use case of `setUserProjection()` is `useGeographic()`, i.e. with EPSG:4326 as user projection and EPSG:3857 as view projection. And for this case it works well. If users work with local projections, they will likely be using them throughout. A corner case is sophisticated global projections like Sphere Mollweide, but I don't think there will be many use cases for modiying geometries in that projection.

PR #14674, opened 2023-04-16
OSMXML: support reading geometry from ways

simon04:
The output mode `out geom`  of Overpass API adds `lon`, `lat` to the `<nd>` elements of `<way>`.

Ref: https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#out

Example: https://overpass-api.de/api/interpreter?data=way(423692233);out+geom;


ahocevar (approved):
Thanks, @simon04!

Issue #13741, opened 2022-06-08
Expressions list in WebGL points layer example is incomplete


mike-000:
The expressions list in https://openlayers.org/en/latest/examples/webgl-points-layer.html was started before https://openlayers.org/en/latest/apidoc/module-ol_style_expressions.html#~ExpressionValue was available in the API and has not been updated as new operators have been added to the API.  Would it be better to provide a link to the API instead of attempting to duplicate it?

PR #14685, opened 2023-04-17
WebGL / add support for more types in style expressions

jahow:
This PR expands the style expressions system to allow:

* multiplying colors:
  ```js
  color: ['*', 'red', 'rgba(255, 255, 255, 0.3)'] // gives a translucent red
  ```
* reading colors or arrays from attributes and variables
  ```js
  color: ['get', 'COLOR'], // reads the attribute value and computes a color out of it
  ```
  ```js
  size: ['var', 'iconSize', 'number_array'] // a type hint can be provided to disambiguate in the case where either
                                            // a number or a number_array is expected
  ```
* the `WebGLVectorLayerRenderer` class has been reworked quite a bit to provide a simpler API and better handle custom attributes; I also wrote test for most of it, which was lacking. Note that there's still pending work on the vector tile equivalent, both APIs are not really consistent and might change again once we introduce `WebGLVectorLayer` and `WebGLVectorTileLayer` classes

Also contains some fixes relating to alpha premultiplication and slightly better error messages when parsing a style.

As an illustration, the `webgl-vector-layer` example has been reworked to rely on the following style:
```js
const style = {
  'stroke-color': ['*', ['get', 'COLOR'], [220, 220, 220]],
  'stroke-width': 1.5,
  'fill-color': ['*', ['get', 'COLOR'], [255, 255, 255, 0.6]],
};
```

### DIFF
--- a/examples/webgl-vector-layer.js
+++ b/examples/webgl-vector-layer.js
@@ -13,24 +13,20 @@ class WebGLLayer extends Layer {
   createRenderer() {
     return new WebGLVectorLayerRenderer(this, {
       fill: {
-        attributes: {
-          color: function (feature) {
-            const color = [...asArray(feature.get('COLOR') || '#eee')];
-            color[3] = 0.6;
-            return packColor(color);
-          },
+        color: function (feature) {
+          const color = [...asArray(feature.get('COLOR') || '#eee')];
+          color[3] = 0.6;
+          return packColor(color);
         },
       },
       stroke: {
-        attributes: {
-          color: function (feature) {
-            const color = [...asArray(feature.get('COLOR') || '#eee')];
-            color.forEach((_, i) => (color[i] = Math.round(color[i] * 0.75))); // darken slightly
-            return packColor(color);
-          },
-          width: function () {
-            return 1.5;
-          },
+        color: function (feature) {
+          const color = [...asArray(feature.get('COLOR') || '#eee')];
+          color.forEach((_, i) => (color[i] = Math.round(color[i] * 0.75))); // darken slightly
+          return packColor(color);
+        },
+        width: function () {
+          return 1.5;
         },
       },
     });
--- a/examples/webgl-vector-layer.js
+++ b/examples/webgl-vector-layer.js
@@ -15,12 +15,9 @@ class WebGLLayer extends Layer {
       fill: {
         attributes: {
           color: function (feature) {
-            const color = asArray(feature.get('COLOR') || '#eee');
-            color[3] = 0.85;
+            const color = [...asArray(feature.get('COLOR') || '#eee')];
+            color[3] = 0.6;
             return packColor(color);
-          },
-          opacity: function () {
-            return 0.6;
           },
         },
       },
@@ -33,9 +30,6 @@ class WebGLLayer extends Layer {
           },
           width: function () {
             return 1.5;
-          },
-          opacity: function () {
-            return 1;
           },
         },
       },
--- a/examples/webgl-vector-layer.js
+++ b/examples/webgl-vector-layer.js
@@ -6,29 +6,29 @@ import TileLayer from '../src/ol/layer/WebGLTile.js';
 import VectorSource from '../src/ol/source/Vector.js';
 import View from '../src/ol/View.js';
 import WebGLVectorLayerRenderer from '../src/ol/renderer/webgl/VectorLayer.js';
-import {asArray} from '../src/ol/color.js';
-import {packColor} from '../src/ol/webgl/styleparser.js';
+import {parseLiteralStyle} from '../src/ol/webgl/styleparser.js';
+
+/** @type {import('../src/ol/style/literal.js').LiteralStyle} */
+const style = {
+  'stroke-color': ['*', ['get', 'COLOR'], [220, 220, 220]],
+  'stroke-width': 1.5,
+  'fill-color': ['*', ['get', 'COLOR'], [255, 255, 255, 0.6]],
+};
 
 class WebGLLayer extends Layer {
   createRenderer() {
+    const parseResult = parseLiteralStyle(style);
     return new WebGLVectorLayerRenderer(this, {
       fill: {
-        color: function (feature) {
-          const color = [...asArray(feature.get('COLOR') || '#eee')];
-          color[3] = 0.6;
-          return packColor(color);
-        },
+        vertexShader: parseResult.builder.getFillVertexShader(),
+        fragmentShader: parseResult.builder.getFillFragmentShader(),
       },
       stroke: {
-        color: function (feature) {
-          const color = [...asArray(feature.get('COLOR') || '#eee')];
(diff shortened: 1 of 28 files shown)


examples/webgl-vector-layer.js
```diff
@@ -6,39 +6,29 @@ import TileLayer from '../src/ol/layer/WebGLTile.js';
 import VectorSource from '../src/ol/source/Vector.js';
 import View from '../src/ol/View.js';
 import WebGLVectorLayerRenderer from '../src/ol/renderer/webgl/VectorLayer.js';
-import {asArray} from '../src/ol/color.js';
-import {packColor} from '../src/ol/renderer/webgl/shaders.js';
+import {parseLiteralStyle} from '../src/ol/webgl/styleparser.js';
+
+/** @type {import('../src/ol/style/literal.js').LiteralStyle} */
+const style = {
+  ['stroke-color']: ['*', ['get', 'COLOR'], [220, 220, 220]],
```
tschaub, line 13: Curious why the brackets here?

src/ol/style/expressions.js
```diff
@@ -21,15 +21,18 @@ import {asArray, fromString, isStringColor} from '../color.js';
  *     of bands, depending on the underlying data source and
  *     {@link import("../source/GeoTIFF.js").Options configuration}. `xOffset` and `yOffset` are optional
  *     and allow specifying pixel offsets for x and y. This is used for sampling data from neighboring pixels.
- *   * `['get', 'attributeName']` fetches a feature attribute (it will be prefixed by `a_` in the shader)
+ *   * `['get', 'attributeName', typeHint]` fetches a feature attribute (it will be prefixed by `a_` in the shader)
  *     Note: those will be taken from the attributes provided to the renderer
+ *     A type hint can optionally be specified, in case the resulting expression contains a type uncertainty which
+ *     will make it invalid. Type hints can be one of: 'string', 'color', 'number', 'boolean', 'number_array'
```
tschaub, line 27: I think `number_array` might be the first introduction of snake_case to the API.

What do you think about these alternatives:
 * `numbers`
 * `numberArray`

examples/webgl-vector-layer.js
jahow: no reason, they're not needed indeed!

src/ol/style/expressions.js
jahow: I would go with `numberArray`; on the other hand I considered `number[]`, it's shorter and quite explicit for more people (because of typescript).

tschaub: I think `number[]` sounds good.

src/ol/style/expressions.js
```diff
@@ -21,15 +21,18 @@ import {asArray, fromString, isStringColor} from '../color.js';
...
  *   * `['resolution']` returns the current resolution
  *   * `['time']` returns the time in seconds since the creation of the layer
- *   * `['var', 'varName']` fetches a value from the style variables, or 0 if undefined
+ *   * `['var', 'varName', typeHint]` fetches a value from the style variables; will throw an error if variable is undefined
+ *     A type hint can optionally be specified (see 'get' operator)
```
tschaub, line 31: Can we infer the type from the variable value during parsing if needed?

src/ol/style/expressions.js
```diff
@@ -366,40 +423,81 @@ function assertArgsMaxCount(args, count) {
...
+      });
+    } else if (outputType !== existing.type) {
+      throw new Error(
+        `The following attribute was used in different places with incompatible types: ${name}
+Types were: ${printTypes(existing.type)} and ${printTypes(outputType)}`
+      );
     }
     const prefix = context.inFragmentShader ? 'v_' : 'a_';
-    return prefix + value;
+    return prefix + name;
```
tschaub, line 500: Doesn’t need to be in this branch, but we should prefix these with something that avoids conflicts with names that we might want to use (e.g. `a_user_` or similar).

src/ol/style/expressions.js
```diff
@@ -21,15 +21,18 @@ import {asArray, fromString, isStringColor} from '../color.js';
  *     of bands, depending on the underlying data source and
  *     {@link import("../source/GeoTIFF.js").Options configuration}. `xOffset` and `yOffset` are optional
  *     and allow specifying pixel offsets for x and y. This is used for sampling data from neighboring pixels.
- *   * `['get', 'attributeName']` fetches a feature attribute (it will be prefixed by `a_` in the shader)
+ *   * `['get', 'attributeName', typeHint]` fetches a feature attribute (it will be prefixed by `a_` in the shader)
```
tschaub, line 24: While updating this doc, we could use the word "property" instead of "attribute" (and I would remove any promise about how things will be spelled in the shader).

src/ol/style/expressions.js
jahow: I considered it but couldn't find a way to do it that didn't feel hacky (like just moving the variables object around). I'll look into it again

jahow: Addressed: the style being parsed is now present in the parsing context, which makes sense in the general sense I think.

src/ol/style/expressions.js
jahow: The doc was clarified

jahow:
Also addressed https://github.com/openlayers/openlayers/issues/13741 while taking care of the expressions documentation.

src/ol/style/expressions.js
jahow: Leaving this for another PR as this one is already quite full! thanks for the suggestion, definitely makes sense

tschaub (approved):

tschaub:
Great to see this coming together.  Thanks for your ongoing work on it, @jahow.

I've been envisioning a migration path for styles that includes adding support for evaluating flat style objects with expressions in the Canvas renderer.  Ideally we could deprecate the `Style` and symbolizer constructors before the WebGL vector rendering were as capable as the Canvas renderer.  In addition, I'm interested in using the same syntax for expressions that aren't strictly about styling (for applying transforms to data and generating data tiles in a worker).

The "parsing" logic for style expressions right now is mainly about serializing shader code.  I'm not sure logic we might be able to share between a parser for WebGL rendering and a parser for other purposes (Canvas rendering or arbitrary data transforms).  But at a minimum we should have a single syntax.  I'm not suggesting doing anything here about it, just letting you know my thoughts.  Something we can dig into more next week.

jahow:
Thanks for all the quick reviews @tschaub! I think the expressions system is ripe for a refactoring. Things I had in mind:
* Implement a distinct consistency check logic which would gather all errors in the expressions instead of throwing when the first one is encountered; this also means the serialization to GLSL or anything else would be much less cluttered by assertions
* add a `toFunction` method or something like that to serialize expressions to runnable Javascript

This would probably help towards the goals you mention!

tschaub:
> Things I had in mind

I like both of these ideas.  Looking forward to working on them.

PR #14692, opened 2023-04-23
Use MapTiler Coordinates API to find proj4 transforms without `+nadgrids` for `fromEPSGCode()`

### LINKED ISSUE #14691: EPSG.io proj4 definition is unreliable in `fromEPSGCode`
**Describe the bug**
The default `fromEPSGCode` lookup is unreliable for the same reason as in #14235.

**To Reproduce**
Compare
https://codesandbox.io/s/fromepsgcode-1dnzpw?file=/main.js
with
https://codesandbox.io/s/fromepsgcode-xmyomh?file=/main.js

**Expected behavior**
EPSG.io lookup is reliable, but only if its `wkt` string is used in proj4 definitions instead of the `proj4` string.

mike-000:
Fixes #14691 
Add `epsgLookupMapTiler` function using MapTiler Coordinates API to find proj4 transforms without `+nadgrids` which can be handled more easily "on demand".

Add an example of using `fromEPSGCode()` via a GeoTIFF source view options.
https://deploy-preview-14692--ol-site.netlify.app/en/latest/examples/cog-projection.html

### DIFF
--- a/src/ol/proj/proj4.js
+++ b/src/ol/proj/proj4.js
@@ -156,3 +156,69 @@ export async function fromEPSGCode(code) {
 
   return get(epsgCode);
 }
+
+/**
+ * Generate an EPSG lookup function which uses the MapTiler Coordinates API to find projection
+ * definitions which do not require proj4 to be configured to handle `+nadgrids` parameters.
+ * Call {@link module:ol/proj/proj4.setEPSGLookup} use the function for lookups
+ * `setEPSGLookup(epsgLookupMapTiler('{YOUR_MAPTILER_API_KEY_HERE}'))`.
+ *
+ * @param {string} key MapTiler API key.  Get your own API key at https://www.maptiler.com/cloud/.
+ * @return {function(number):Promise<string>} The EPSG lookup function.
+ * @api
+ */
+export function epsgLookupMapTiler(key) {
+  return async function (code) {
+    const response = await fetch(
+      `https://api.maptiler.com/coordinates/search/code:${code}.json?transformations=true&exports=true&key=${key}`
+    );
+    if (!response.ok) {
+      throw new Error(
+        `Unexpected response from maptiler.com: ${response.status}`
+      );
+    }
+    return response.json().then((json) => {
+      const results = json['results'];
+      if (results?.length > 0) {
+        const result = results.filter(
+          (r) => r['id']?.['authority'] === 'EPSG' && r['id']?.['code'] === code
+        )[0];
+        if (result) {
+          const transforms = result['transformations'];
+          if (transforms?.length > 0) {
+            // use default transform if it does not require grids
+            const defaultTransform = result['default_transformation'];
+            if (
+              transforms.filter(
+                (t) =>
+                  t['id']?.['authority'] === defaultTransform?.['authority'] &&
+                  t['id']?.['code'] === defaultTransform?.['code'] &&
+                  t['grids']?.length === 0
+              ).length > 0
+            ) {
+              return result['exports']?.['proj4'];
+            }
+            // otherwise use most accurate alternative without grids
+            const transform = transforms
+              .filter(
+                (t) =>
+                  t['grids']?.length === 0 &&
+                  t['target_crs']?.['authority'] === 'EPSG' &&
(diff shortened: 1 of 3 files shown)


ahocevar:
It has been discussed here already, but I'm unable to find the issue at the moment. The epsg.io API allows to customize what kind of definition you're requesting, and you can say you don't want grid shift transforms. Using WKT instead of a proj string is just an ugly hack in my opinion, so I'm not in favor of this pull request as it is.

mike-000:
It was discussed in #14235.as described in #14691.  For EPSG:27700 the solution would be to use https://epsg.io/27700-1314.proj4 instead of https://epsg.io/27700.proj4  To discover that was needed you would first need to run https://epsg.io/?q=27700&format=json&trans=1 and filter the returned transforms to those with `target_crs` 4326 and `grids_list` length 0.  There  is nothing documented in the API https://github.com/maptiler/epsg.io which could do it in a single lookup.

ahocevar:
I think two lookups would be acceptable. But with help from @klokan, we could do even better: what if epsg.io could provide a download API for the the NTv2 files referenced in the `+nadgrids` param?

mike-000:
It would certainly be useful to have a central repository for general use with proj4.  However, some are quite large files (e.g. OSTN15_NTv2_OSGBtoETRS.gsb is almost 15MB).  Using a `+towgs84` definition would be quicker if dynamically defining projections.

klokan:
+1
We could perhaps make it streamable with HTTP Range request - to download into the client only blocks which are relevant for the required transformation - in parts?

klokan:
BTW The EPSG Coordinates API is well documented at https://docs.maptiler.com/cloud/api/coordinates/ - this is the one which is supported also for future and which we could improve to provide additional outputs.

src/ol/proj/proj4.js
```diff
@@ -93,11 +93,45 @@ export function register(proj4) {
...
+      if (result) {
+        const transforms = result['trans'];
+        if (transforms?.length > 0) {
+          const transform = transforms.filter(
+            (t) =>
+              t['grids_list']?.length === 0 &&
+              t['target_crs'] === 4326 &&
+              t['deprecated'] === 0 &&
+              t['usable'] === true
+          )[0]?.['code_trans'];
```
ahocevar, line 125: You could sort the results by accuracy, and use the one with the highest accuracy instead of the first.

ahocevar (changes requested):

ahocevar:
@mike-000: According to @klokan's comment, I think it would make sense to switch to the mentioned Coordinates API.

mike-000:
> @mike-000: According to @klokan's comment, I think it would make sense to switch to the mentioned Coordinates API.

That would be a breaking change as it needs an API key.

ahocevar:
Good point, @mike-000. But using an undocumented API isn't a good solution either. I think the best solution would be to leave things as they are, but provide a ready-made function factory to call with the API key, which users can then pass to `setEPSGLookup()`.

mike-000:
The example continues to use the default lookup without an API key as the UTM projection does use `+nadgrids`.  I could add another such as https://codesandbox.io/s/fromepsgcode-maptiler-jn73tq?file=/main.js if one is needed.

ahocevar (changes requested):
Thanks for your continued effort on this, @mike-000.

Two things that I would change:
* Replace the current example with one that actually uses the `epsgLookupMapTiler()` function
* Use the view returned from the COG as-is, i.e.
   ```js
   setEPSGLookup(epsgLookupMapTiler(key));
   
   // ...
   const map = new Map({
     // ...
     view: cogSource().getView()
   }); 
   ```

mike-000:
The minimum needed would be

```
  view: cogSource()
    .getView()
    .then((viewConfig) =>
      fromEPSGCode(viewConfig.projection.getCode()).then(() => viewConfig)
    ),
```

otherwise the projection will not be registered with `proj4` and any other layers will not be reprojected

e.g. https://codesandbox.io/s/fromepsgcode-maptiler-forked-5vcmy9?file=/main.js

ahocevar (changes requested):

ahocevar (approved):
Thanks, @mike-000

Issue #14705, opened 2023-04-25
Raster reprojection without tiled source?


jasontk19:
I am trying to reproject WMS image layers from EPSG:4326 into EPSG:3857.  My map view is configured to use 3857: 
```
  const mapObject = new Map({
    view: new View({
      projection: 'EPSG:3857',
      center: [0, 0],
      zoom: 0,
    }),
  });
```

and my layer source is configured to use 4326:
```
const params = new URLSearchParams({
      STYLES: style.name,
      FORMAT: 'image/png',
      TRANSPARENT: true,
      EXCEPTIONS: 'application/vnd.ogc.se_inimage',
      TIME: timeStep,
      DIM_reference_time: issuance,
      dataset,
      layers: layerName,
      ratio: 1,
    });

const source = new ImageWMS({
      projection: 'EPSG:4326',
      url: WMS_URL,
      params: Object.fromEntries(params),
      resolutions,
      imageLoadFunction(img, src) {
        if (getLoadedSteps().length !== allTimeSteps.length) {
          // eslint-disable-next-line no-param-reassign
          img.getImage().src = src;
        }
      },
    });
    const subLayer = new ImageLayer({
      source,
    });
```

The raster reprojection docs show an example using tiled layers, which in my experimentation works fine.  However, if I try to use an ImageLayer with an ImageWMS source, the WIDTH, HEIGHT, and BBOX params sent in the WMS request are wrong. As you can see the image requested is just 1x1 px and the BBOX coordinates are in 3857 format.

<img width="630" alt="Screen Shot 2023-04-25 at 10 02 10 AM" src="https://user-images.githubusercontent.com/1480000/234336059-32402475-cec7-4ba2-828e-e2dec412f7f3.png">

Please advise if I am going about this the wrong way or misunderstand how this should work.  Does raster reprojection only support tiled sources?  We cannot use tiled WMS for performance reasons.

mike-000:
It works as expected here https://codesandbox.io/s/wms-image-forked-h1xspr?file=/main.js  Are you `resolutions` meaningful for EPSG:4326?

jasontk19:
Thanks for the pointer Mike, the resolutions were definitely the issue.  

For some background and clarification:  Our WMS service handles meridian/antimeridian crossing requests that are made for 4326 but does not handle 3857 properly.  So my objective was to keep the map in 3857 while making the WMS requests in 4326 and then reprojecting them.

[...]

PR #14706, opened 2023-04-26
Do not restrict image reprojection extent width for global projections

mike-000:
Fixes https://github.com/openlayers/openlayers/issues/14705#issuecomment-1522416020

Image wrapping is not directly supported by OpenLayers but works if the server supports wrapped coordinates.  Reprojection is currently assuming there is no wrapping and restricting the extent to projection extents and therefore not handling the image produced by the server.  While restricting the x axis is needed in non-global projections and restricting the y axis is essential in all cases to avoid going beyond the poles, there is no need to limit the x coordinates of global projections beyond those of the view extent, or the layer extent if one is specified.

### DIFF
--- a/src/ol/reproj/Image.js
+++ b/src/ol/reproj/Image.js
@@ -11,7 +11,13 @@ import {
   calculateSourceResolution,
   render as renderReprojected,
 } from '../reproj.js';
-import {getCenter, getHeight, getIntersection, getWidth} from '../extent.js';
+import {
+  getCenter,
+  getHeight,
+  getIntersection,
+  getWidth,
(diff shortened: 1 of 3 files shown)


mike-000:
There is a separate existing issue when requests are made for images in the form `&WIDTH=NaN&HEIGHT=NaN&BBOX=NaN,NaN,NaN,NaN` if projection extents are exceeded - I presume this was not intentional.  It no longer happens if both projections are global, but continues to in other cases.

ahocevar (approved):
Thanks, @mike-000

PR #14712, opened 2023-04-28
Support FlatStyle in Feature setStyle() method

mike-000:
If the vector layer `setStyle()` method supports flat styles the method for Feature should also be able to.

Also the Feature `createStyleFunction()` static method was historically different to that for layers, but it now does the same as the `ol/style/Style` `toFunction()` method and could be deprecated in favor of that.


src/ol/Feature.js
```diff
@@ -265,13 +266,24 @@ class Feature extends BaseObject {
...
    */
   setStyle(style) {
-    this.style_ = style;
-    this.styleFunction_ = !style ? undefined : createStyleFunction(style);
+    let styleLike;
+    if (
+      !style ||
+      typeof style === 'function' ||
+      style instanceof Style ||
+      (Array.isArray(style) && style.every((s) => s instanceof Style))
```
mike-000, line 279: This prevents test breaking unnecessary cloning of style arrays which was introduced in layer styles in #13900.  The check could be moved to `toStyleLike` to fix the problem at source.

jahow:
In my understanding the flat style format is not intended to be applied to a single feature. Applying styles to single features is actually something that the expression system intends to replace.

mike-000:
Yes, feature styles might go eventually (KML could be a problem) but currently there are places where flat styles can be used (layers, Draw and Modify interactions) while they cannot be used on features and Select interaction, or in the immediate renderer, which is unintuitive, and inconvenient when flat styles are simpler to code and self-documenting.

ahocevar:
We're not going to add short-lived API methods that take us in the opposite direction of where we're headed with flat styles. All the places that @mike-000 mentioned apply the style to layers. The Select interaction is something we should not have re-added after we had removed it a couple of years ago. And the immediate renderer won't benefit from adding flat style support for `Feature#setStyle()`, because it has a `drawFeature(feature, style)` method which takes the style as separate argument.

mike-000:
If feature styles and style objects are to be deprecated I presume an API compliant workaround such as
```
feature.setStyle(new VectorLayer({style}).getStyleFunction())
```
to share flat styles used elsewhere will continue be useable until then?

(I only mentioned the immediate renderer because it has a `setStyle` method not included in the PR)

daboudii:
> We're not going to add short-lived API methods that take us in the opposite direction of where we're headed with flat styles. All the places that @mike-000 mentioned apply the style to layers. The Select interaction is something we should not have re-added after we had removed it a couple of years ago. And the immediate renderer won't benefit from adding flat style support for `Feature#setStyle()`, because it has a `drawFeature(feature, style)` method which takes the style as separate argument.

So do you advice not to use `Select` interaction ?

ahocevar:
> So do you advice not to use Select interaction ?

@daboudii Yes, that would be my advice. In most scenarios (e.g. [vector tiles](https://openlayers.org/en/latest/examples/vector-tile-selection.html)), there are more efficient ways to style selected features differently than what the Select interaction does.

PR #14714, opened 2023-04-30
Handle user projections in Graticule

### LINKED ISSUE #13650: Geographic coordinates and graticule
**Describe the bug**
A repeat of part of #10881.  While most of that was fixed the problem with the graticule remains when using non-parallel projections..

**To Reproduce**
Select 27700 in https://codesandbox.io/s/reprojection-by-code-forked-jnsgom?file=/main.js and display the graticule.  All lines are straight.  The parallels, which should be clearly curved, are too far north over the center of the view.  The meridians should also be slightly curved.

**Expected behavior**
The output should be the same regardless of whether geographic coordinates are used..  I think the only way to produce the correct curved lines would be to generate the graticule using view projection coordinates internally.

mike-000:
Fixes #13650


src/ol/layer/Graticule.js
```diff
@@ -518,9 +532,15 @@ class Graticule extends VectorLayer {
    * resolution.
    * @param {import("../extent.js").Extent} extent Extent.
    * @param {number} resolution Resolution.
+   * @param {import("../proj/Projection.js").default} projection Projection.
+   * @param {import("../featureloader.js").FeatureLoaderInternal} [internal] Internal properties.
    * @return {Array<import("../extent.js").Extent>} Extents.
    */
-  strategyFunction(extent, resolution) {
+  strategyFunction(extent, resolution, projection, internal) {
```
ahocevar, line 539: We have all this additional information in the frame state, so there should be no need to pass it to the strategy function.

mike-000: ol/source/Vector which calls the strategy function does not have the frameState, but at that point it does have the internal properties which will be passed to the loader.  Also the internal extent is not taken directly from the frameState as the renderer handles wrapping.

PR #14724, opened 2023-05-08
Update Maptiler topo map to latest version

ahocevar:
This pull request updates the Maptiler topo map to the latest version (v2). This is necessary because in the previous version, the used hillshade tiles are not available for all zoom levels any more, which can be seen in the small map on https://openlayers.org/.

### DIFF
--- a/examples/mapbox-style.js
+++ b/examples/mapbox-style.js
@@ -3,7 +3,7 @@ import olms from 'ol-mapbox-style';
 
 olms(
   'map',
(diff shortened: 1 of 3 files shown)


jahow (approved):

jahow:
Wait, sorry but I think the offscreen canvas example is broken now? No error but the map stays blank:
image

ahocevar:
I'll create a separate pull request for that. Thanks for noticing!

PR #14734, opened 2023-05-08
Details on conforming with the style guidelines

tschaub:
This adds documentation on the project's style guidelines – including configuring an editor with an ESLint integration.

I've removed the `.editorconfig` file as this does not represent all of the project's style guidelines.  A proper ESLint integration is preferred.


jahow:
Thanks, always good to have some doc refreshing!

PR #14743, opened 2023-05-12
Handle user projection in Translate interaction

### LINKED ISSUE #14666: Translate interaction with geographic coordinates changes geometry shape

mike-000:
Fixes #14666

A basic fix which handles all geometry types but may result in non `[cx + r, cy]` compliant Circle geometry if a geographic user projection Circle geometry is translated across a non-parallel view projection such as transverse mercator.  Problems caused by the Circle `setCenter` method not working in non-compliant geometries are avoided by removing the unneeded override of the Circle `translate` method and using flat coordinates (although these will still need fixing for the Modify interaction to work correctly in a similar situation).

### DIFF
--- a/src/ol/geom/Circle.js
+++ b/src/ol/geom/Circle.js
@@ -4,7 +4,7 @@
 import SimpleGeometry from './SimpleGeometry.js';
 import {createOrUpdate, forEachCorner, intersects} from '../extent.js';
 import {deflateCoordinate} from './flat/deflate.js';
-import {rotate, translate} from './flat/transform.js';
+import {rotate} from './flat/transform.js';
 
 /**
  * @classdesc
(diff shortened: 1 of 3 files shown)


ahocevar (approved):
Thanks, @mike-000. The code makes sense now. However, I had to read the part about Circle geometries in the description of the pull request 10 times, and still failed to understand what you are trying to say.

PR #14744, opened 2023-05-14
Handle Circle geometry with non-parallel user projections in Modify interaction

mike-000:
Replaces #14672


src/ol/geom/Circle.js
```diff
@@ -172,10 +172,12 @@ class Circle extends SimpleGeometry {
    */
   setCenter(center) {
     const stride = this.stride;
-    const radius = this.flatCoordinates[stride] - this.flatCoordinates[0];
+    const radiusX = this.flatCoordinates[stride] - this.flatCoordinates[0];
+    const radiusY = this.flatCoordinates[stride + 1] - this.flatCoordinates[1];
```
ahocevar, line 176: y should be the same for the center and the point right of the center on the perimeter. So if we have different y values here, it's a bug that needs to be fixed before making any other changes.

mike-000: It will not be the same if the geometry has been transformed between projections where the x axes are not parallel.

ahocevar: Then that's the bug that needs to be fixed.

mike-000: The API https://openlayers.org/en/latest/apidoc/module-ol_geom_Circle-Circle.html#transform explains

> This `transform` function just transforms these two points.

We need to handle that internally, but when I suggested we should avoid returning coordinates in that format to the user you did not consider it important.

ahocevar: The more changes you're trying to get in here, the more I see how deep the rabbit hole goes. And fixing the transform, which is an easy fix, will make a ton of other things easier in return.

> This transform function just transforms these two points.

is cited out of context, and maybe not well expressed. It should say "This transform function just transforms the center and the radius of the circle". The overall meaning is that a circle is not a polygon with the shape of a circle, but an artificial construct with just a center and a radius.

mike-000: > Transforming a `Circle` from one coordinate system to another means: 1. transform the center (first coordinate), 2. convert the radius from source crs units to destination crs units, 3. encode the radius in the second coordinate.

It is not as simple as that as the reference points will be on different alignments, so extrapolation may be needed as in https://github.com/openlayers/openlayers/pull/14672/files#diff-90b8ad0e1e436e034c530c6de89696519c2f2308e8b90d9795325599f34a8a4fR294 to produce a visually similar result.

ahocevar:
The test I'd expect to pass is:
```js
const original = new Circle(fromLonLat([16, 48], 100);
const transformed = original.clone().transform('EPSG:3857', 'EPSG:4326');
const transformedBack = transformed.clone().transform('EPSG:4326', 'EPSG:3857');
expect(transformedBack.getCenter()).to.eql(original.getCenter()); // a floating point tolerance is ok
expect(transformedBack.getRadius()).to.eql(original.getRadius()); // a floating point tolerance is ok
```

mike-000:
That already works in all projections.  However, in non-parallel projections (e.g. replace 'EPSG:3857' with 'EPSG:32633') you will have `radiusY !== 0` in `transformed`.  I was suggesting we leave `transform` as documented but use `transformToNormalCircle` to ensure the result of an interaction returned in user coordinates always has `radiusY == 0` and the user does not get any suprises when `circle.setRadius(circle.getRadius())` moves the reference point onto a different alignment causing a visual change in the view projection as in https://github.com/openlayers/openlayers/pull/14672#issuecomment-1546630205.

This simply approach (using the existing `transform`) does **not** work

```
const original = new Circle(fromLonLat([16, 48], 'EPSG:32633'), 100);
const temp1 = original.clone().transform('EPSG:32633', 'EPSG:4326');
const transformed = new Circle(temp1.getCenter(), temp1.getRadius());
const temp2 = transformed.clone().transform('EPSG:4326', 'EPSG:32633');
const transformedBack = new Circle(temp2.getCenter(), temp2.getRadius());
```

This does work, but will have `radiusY !== 0` in `transformedBack` 

```
const original = new Circle(fromLonLat([16, 48], 'EPSG:32633'), 100);
const transformed = original.clone().transformToNormalCircle('EPSG:32633', 'EPSG:4326');
const transformedBack = transformed.clone().transform('EPSG:4326', 'EPSG:32633');
```

ahocevar:
Are you going to fix this or should I? Making the above test pass is within what's documented, but the documentation could be improved as I suggested above. I also added two expectations to the test above.

mike-000:
Although not self-reversing the `transformToNormalCircle` method is reversible with a standard transform.  Therefore a reversible method is possible if the appropriate logic is used depending on which projection is source and destination.  The built in EPSG:3857 and EPSG:4326 and their equivalents can be identified, so if they are involved that is possible.  In the case of a transform between two custom projections the existing default transform would still be used, and the inherited `applyTransform` method would also continue do exactly as the documentation says.

ahocevar:
#14766 addresses the underlying problem with the violated `[cx, cy, cx +r, cy]` coordinate pattern by circle geometry transforms.

PR #14760, opened 2023-05-17
Accept custom templates with only `{x}`, `{y}` and `{z}` placeholders for Zoomify

### LINKED ISSUE #14758: Zoomify source with only {x}, {y}, and {z} placeholders not handled
**Describe the bug**
If only `{x}`, `{y}`, and `{z}` placeholders are required for a Zoomify url a workaround involving a tileUrlFunction is needed.  That should not be necessary.

**To Reproduce**
The tiles in https://codesandbox.io/s/simple-forked-ghl6o5?file=/main.js need to use `ol/source/Zoomify` for the tiles at the right and bottom edges of the grid to be rendered correctly.  However that source currently expects a url to contain either `{TileGroup}` or `{tileIndex}`.  `{x}`, `{y}`, and `{z}` are always needed with `{TileGroup}`, but in this case will they will also work alone.

**Expected behavior**
Check for

      !(url.includes('{x}') && url.includes('{y}') && url.includes('{z}')) &&
      !url.includes('{tileIndex}')

instead of

      !url.includes('{TileGroup}') && !url.includes('{tileIndex}')

before resorting to adding a default extension to the supplied url.

mike-000:
Fixes #14758

### DIFF
--- a/src/ol/source/Zoomify.js
+++ b/src/ol/source/Zoomify.js
@@ -93,6 +93,7 @@ export class CustomTile extends ImageTile {
  * `http://my.zoomify.info/IMAGE.TIF/`. A URL template must include
  * `{TileGroup}`, `{x}`, `{y}`, and `{z}` placeholders, e.g.
  * `http://my.zoomify.info/IMAGE.TIF/{TileGroup}/{z}-{x}-{y}.jpg`.
+ * Custom templates with only the `{x}`, `{y}`, and `{z}` placeholders are accepted.
  * Internet Imaging Protocol (IIP) with JTL extension can be also used with
  * `{tileIndex}` and `{z}` placeholders, e.g.
  * `http://my.zoomify.info?FIF=IMAGE.TIF&JTL={z},{tileIndex}`.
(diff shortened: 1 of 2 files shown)


ahocevar (commented):
As a library user, it would never occur to me to use a Zoomify source when I have a XYZ source with tiles that don't have the full size in the last row or column.

I'd be interested to hear which tool creates such tile sets.

mike-000:
I encountered this in https://gis.stackexchange.com/questions/459802/tiled-image-with-pixel-coordinates-projection-by-openlayers-7  I have found these related sites https://github.com/tilemapjp https://code4history.dev/maplat_flyer.pdf https://github.com/code4history

Also DZI sources (an old Microsoft specification now supported by OpenSeadragon) have similar tiles and differ from Zoomify only in the definition of zoom level 0, which is a single pixel instead of a single tile.  They could be easily supported by OpenLayers with a `minSize` (or similar name) options to override `tileSize` when determining the grid zoom levels as in https://codesandbox.io/s/dzi-c88ui3?file=/main.js  A .xml or .dzi file (e.g. https://s3.amazonaws.com/dzchattzoom/1907_west/dzc_output.xml) defines the sizes, the OpenLayers Zoomify custom tile takes care of the overlaps.

PR #14761, opened 2023-05-18
WIP: Declutter vector layers by group

scadergit:
Allows declutter to be organized by groups to avoid writing over layers higher in the layer order. This is a proof of concept, the implementation may need refinement.

Refer to https://github.com/openlayers/ol-mapbox-style/issues/687. Requires a pull request for ol-mapbox-style to apply the `declutterGroup` property to layers created by vector styling.

Every layer can be defined with the declutterGroup parameter. For mapbox vector layers, it can be passed in the `apply` options parameter, or inherit from the Group layer if applied to a layer group, e.g.:

```
new Group({
  declutterGroup: "basemap"
});
```

```
let LayerGroup = new Group();
apply(LayerGroup, 'https://api.maptiler.com/maps/topo-v2/style.json?key=get_your_own_D6rA4zTHduk6KOKTXzGB', {
  declutterGroup: 'basemap-option'
});
```
If not defined by any layer, declutter by groups will be ignored. If defined to only some of the layers, proceeding layers will be grouped together.

The example is a WIP with both examples, along with a temporary tiled layer to NWS radar to demonstrate.


scadergit:
@ahocevar if this pull request is merged, will it only go into 7 and not 6? We use VueLayers, and their most recent is not compatible yet with 7

ahocevar:
I'm going to review this in the next couple of days. Thanks for your effort, @scadergit. If it gets merged, it will be in 7.x. We do not provide backports.

src/ol/renderer/Composite.js
```diff
@@ -120,6 +120,18 @@ class CompositeMapRenderer extends MapRenderer {
         continue;
       }
 
+      if (previousDeclutterGroup !== null && previousDeclutterGroup !== layer.get('declutterGroup')) {
+        console.log(`>>> DECLUTTERING group '${previousDeclutterGroup}'`, `before switching to '${layer.get('declutterGroup')}' group`);
```
tschaub, line 124:
```suggestion
```

src/ol/renderer/Composite.js
```diff
@@ -120,6 +120,18 @@ class CompositeMapRenderer extends MapRenderer {
...
+        for (let i = declutterLayers.length - 1; i >= 0; --i) {
+          declutterLayers[i].renderDeclutter(frameState);
+        }
+        declutterLayers = []    
+      }
+      if (layer.get('declutterGroup') && layer.get('declutterGroup') !== previousDeclutterGroup) {
+        previousDeclutterGroup = layer.get('declutterGroup')
+        console.log('declutterGroup', previousDeclutterGroup)
```
tschaub, line 132:
```suggestion
```

src/ol/renderer/Composite.js
```diff
@@ -134,9 +146,13 @@ class CompositeMapRenderer extends MapRenderer {
         );
       }
     }
-    for (let i = declutterLayers.length - 1; i >= 0; --i) {
-      declutterLayers[i].renderDeclutter(frameState);
+    if (declutterLayers.length > 0) {
+      console.log('-------->>> DECLUTTERING REMAINING');
```
tschaub, line 150:
```suggestion
```

src/ol/renderer/Composite.js
```diff
@@ -134,9 +146,13 @@ class CompositeMapRenderer extends MapRenderer {
...
+      for (let i = declutterLayers.length - 1; i >= 0; --i) {
+        declutterLayers[i].renderDeclutter(frameState);
+      }
     }
+    console.log('[layerStates:finish]')
```
tschaub, line 155:
```suggestion
```

ahocevar:
@scadergit To make it easier to review, would you be able to remove the lint by running `npm run lint -- --fix` and commit the changes? Thanks!

scadergit:
@tschaub removed the comments, I left them in only temporary to explain thought, but guess that would be a better case for a code comment

ahocevar (changes requested):
Given the fact that a `declutterGroup` is just a hint to the renderer to declutter what has been rendered already when its value changes, I think it could make sense to rename the new property to `startDeclutterGroup` and make it `boolean`. So whenever the render loop in the composite renderer stumbles across a layer with `startDeclutterGroup: true`, it will declutter what has been rendered so far or since the last `startDeclutterGroup: true`, and zero the length of the `declutterLayers` array. This would also avoid confusion, because it is not possible to add to a previous `declutterGroup` that has already been rendered before another `dececlutterGroup`

I think this is moving in the right direction, so it would be good to also add tests.

examples/declutter-group.html
```diff
@@ -0,0 +1,16 @@
...
+docs: >
+  This shows how to specify the frequency that vector layers are decluttered. If not specified, decluttering
+  will overlay all layers despite layer order. When specified, decluttering will occur every time the 
+  `declutterGroup` changes between layers. The same group can contain any number of layers. For vector layers,
+  the `declutterGroup` can be set in the `apply` options, or inherit from the parent Group if applied to a group.
+
+tags: "mapbox, declutter, vector"
+cloak:
+  - key: pk.eyJ1IjoiYWhvY2V2YXIiLCJhIjoiY2t0cGdwMHVnMGdlbzMxbDhwazBic2xrNSJ9.WbcTL9uj8JPAsnT9mgb7oQ
+    value: Your Mapbox access token from https://mapbox.com/ here
```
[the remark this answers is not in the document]
scadergit: oops, thanks

scadergit:
That approach definitely simplifies everything and removes the `ol-mapbox-style` change. I've pushed an update to code, documentation, and the example. I've also simplified the example using a blue polygon overlay so that the same can be used for a render test that can be reproduced.

Speaking of tests, I presume this should have browser tests for setting the option (on all layer subclasses?), and then also a render test? 

Several render tests are failing locally, even on tests I haven't modified--is that expected? 
```
case ./cases/layer-mapbox-vector/main.js (Mapbox vector layer renders)': mismatch 0.0323028564453125
case ./cases/layer-vector-decluttering/main.js': mismatch 0.0085296630859375
case ./cases/layer-vector-decluttering-extended/main.js': mismatch 0.0098876953125
case ./cases/layer-vector-multigeometry-decluttering/main.js': mismatch 0.0217742919921875
case ./cases/layer-vectorimage-decluttering/main.js': mismatch 0.0098114013671875
case ./cases/layer-vectortile-icon-label/main.js (Vector tile layer declutters image with text correctly)': mismatch 0.0288848876953125
case ./cases/linestring-style-text-repeat/main.js (Vector tile layer renders)': mismatch 0.025238037109375
case ./cases/rich-text-justify-style/main.js': mismatch 0.0704345703125
case ./cases/rich-text-style/main.js': mismatch 0.0566558837890625
case ./cases/rtl-text-align/main.js': mismatch 0.0111236572265625
case ./cases/text-style/main.js': mismatch 0.025390625
case ./cases/text-style-linestring-nice/main.js': mismatch 0.0251617431640625
RENDERING TESTS FAILED
```
The failure looks related to different ways systems render text; for example, using layer-vector-decluttering, the font has slightly more kern and space between itself and the circles. Is that supposed to fail? Is this only likely to pass within your build system?
image

scadergit:
@ahocevar thoughts on my render tests question? Would like this wrap this up, but unsure how to proceed when existing tests are failing.

ahocevar:
@scadergit These rendering tests failing locally are nothing to worry about. It's like you said - different font rendering systems on different platforms. Sorry this is taking so long for a final review, I've been too busy recently. Will try to give this a final review later today.

src/ol/layer/Base.js
```diff
@@ -48,6 +48,9 @@ import {clamp} from '../math.js';
  * @property {BackgroundColor} [background] Background color for the layer. If not specified, no background
  * will be rendered.
  * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
+ * @property {boolean} [startDeclutterGroup] Signals to the renderer to start a new declutter group, which causes previous vector(tile)
+ * layers with the declutter option set true to declutter before rendering the current layer. May be set on multiple layers to manage
+ * multiple declutter efforts.
```
ahocevar, line 53: The new option won't work on layer groups (because they are not included in the `layerStatesArray`), so it should not be added on `Group`'s base here either.

src/ol/layer/Group.js
```diff
@@ -68,6 +68,9 @@ export class GroupEvent extends Event {
  * be visible.
  * @property {Array<import("./Base.js").default>|Collection<import("./Base.js").default>} [layers] Child layers.
  * @property {Object<string, *>} [properties] Arbitrary observable properties. Can be accessed with `#get()` and `#set()`.
+ * @property {boolean} [startDeclutterGroup] Signals to the renderer to start a new declutter group, which causes previous vector(tile)
+ * layers with the declutter option set true to declutter before rendering the current layer. May be set on multiple layers to manage
+ * multiple declutter efforts.
```
ahocevar, line 73: The new option won't work on layer groups (because they are not included in the `layerStatesArray`), so it should not be added here.

examples/declutter-group.js
```diff
@@ -0,0 +1,84 @@
...
+  target: 'map',
+  view: new View({
+    center: [-10203186.115192635, 4475744.563386],
+    zoom: 4,
+  }),
+  layers: [LayerGroup, overlay],
+});
+
+function update(event) {
+  overlay.set('startDeclutterGroup', event.target.value === 'true');
```
ahocevar, line 75: If you make `startDeclutterGroup` an observable property, you'll still need to add a getter and setter (`getStartDeclutterGroup()` and `setStartDeclutterGroup()`). But I'd just make it a private member property on `Layer`, and add a public getter and setter.

ahocevar (changes requested):

examples/declutter-group.js
scadergit: This sent me on a reading exercise as I never knew the scope of `set` in regards to being observable. I am going with private member getter/setter, but I'm curious if they should be included in the API docs. Looking at the code, I thought it was based on usefulness outside other classes, but that line is blurry...should I include these in the API docs?

src/ol/layer/Base.js
scadergit: Whoops, carry-over from the previous approach, thanks for catching.

examples/declutter-group.js
ahocevar: If the setter is not part of the API, there's also no need to add a setter in the first place.

src/ol/layer/Base.js
scadergit: I was adding the new property to Layer.test.js and need to clarify that `startDeclutterGroup` should only be a private member of `Layer` and not included in the `layerState` defined in `Base.js` --is that correct? I'm a little fuzzy on the role of the state.

examples/declutter-group.js
scadergit: Our use-case, weather raster layers over vector basemaps, could find the ability to toggle this feature convienent, but perhaps that is not the best way to achieve that behavior? Is that enough reason to keep a setter and add it to the API? Otherwise, a configure and forget wouldn't need a setter.

ahocevar: Configure and forget sounds good. A setter can be added later if someone need's it, but I think we're good without.

scadergit: I updated the branch, a couple of thoughts:
1. Without a setter, I had to set the private property directly in the constructor from options--I assume that is best practice
2. There are unrelated browser test failures, curious to see how that pans out in tests ran by the workflow

scadergit:
Not sure why rendering failed with 404's, I only get noted miss-match issues locally:
```webpack 5.82.0 compiled successfully in 19464 ms
case ./cases/layer-vector-decluttering/main.js': mismatch 0.0085296630859375
case ./cases/layer-vector-decluttering-extended/main.js': mismatch 0.0098876953125
case ./cases/layer-vector-multigeometry-decluttering/main.js': mismatch 0.0217742919921875
case ./cases/layer-vectorimage-decluttering/main.js': mismatch 0.0098114013671875
case ./cases/layer-vectortile-icon-label/main.js (Vector tile layer declutters image with text correctly)': mismatch 0.0288848876953125
case ./cases/linestring-style-text-repeat/main.js (Vector tile layer renders)': mismatch 0.025238037109375
case ./cases/rich-text-justify-style/main.js': mismatch 0.072113037109375
case ./cases/rich-text-style/main.js': mismatch 0.043670654296875
case ./cases/rtl-text-align/main.js': mismatch 0.0111236572265625
case ./cases/text-style/main.js': mismatch 0.025390625
case ./cases/text-style-linestring-nice/main.js': mismatch 0.0251617431640625
RENDERING TESTS FAILED
```

ahocevar:
@scadergit These tests are sensitive to different font rendering subsystems on different platforms, and usually only pass on Debian style Linux systems. So no need to worry if they fail for you locally. What counts is the test results in the CI. We had to set the tolerances on font rendering tests very low to avoid false passes.

scadergit:
Yeah, this morning the CI showed failed due to missing files, not just from tolernaces...but now they passed. /shrug

scadergit:
Hi, @ahocevar, just making sure there's nothing left for me to change.

ahocevar:
@scadergit Sorry for the delay on this. I think you've addressed everything, I only need to find some time to do a final review.

scadergit:
I agree that aligns better and refelects how flexible the application is. This would also make it easier to toggle.

I presume this line on your new fork was copied over unintentionally? https://github.com/openlayers/openlayers/pull/14971/files#diff-aa03714c5c755515d314c2928d5519593d119e2c046c05cdd6bb95bcb6c99329R18

ahocevar:
> I presume this line on your new fork was copied over unintentionally?

Indeed, thanks for catching that. I removed the line.

PR #14765, opened 2023-05-21
WebGL / Add support for multiple styles in vector & vector tile renderers

jahow:
This PR implements support for an array of styles in `WebGLVectorLayerRenderer` and `WebGLVectorTileLayerRenderer`.

To do so, a new class called `VectorStyleRenderer` is introduced; this class is responsible for:
1. generating buffers for one style
2. rendering those buffers

The former `BatchRenderer` classes were merged into this new class which provides a simpler API: `generateBuffers` and `render`.

This also allows harmonizing the APIs of the Vector and Vector Tile renderers: each of them now take a `style` property which can be an object or array of objects; each individual style can either be a `LiteralStyle` or a set of custom shaders. The `VectorStyleRenderer` will handle both cases, taking a lot of responsibility away from the layer renderers.

The aim of this work is to eventually be able to transform mapbox style objects into an array of styles compatible with WebGL renderers, thus enabling a conversion from one to another.

Note: the `render/webgl/shaders.js` module with default shaders has been removed, since it would have required to be maintained in parallel with the `ShaderBuilder` class. Now that literal styles can be provided for vector and vector tile layers, I figured this module wasn't really useful anymore.


src/ol/render/webgl/VectorStyleRenderer.js
```diff
@@ -0,0 +1,440 @@
...
+  getCustomAttributesSize,
+} from './renderinstructions.js';
+import {parseLiteralStyle} from '../../webgl/styleparser.js';
+
+const WEBGL_WORKER = createWebGLWorker();
+let workerMessageCounter = 0;
+
+/**
+ * Names of attributes made available to the vertex shader.
+ * Please note: changing these *will* break custom shaders!
```
tschaub, line 26: Small note that probably deserves a longer discussion elsewhere:  I don't think we should try to support custom shaders.  If we did allow people to provide arbitrary shaders, it would impose some serious constraints on how we could continue evolving the rendering implementation.  We likely couldn't do things like upgrade from WebGL 1 to 2, or migrate to WebGPU, for example.

examples/webgl-vector-tiles.js
```diff
@@ -6,39 +6,62 @@ import View from '../src/ol/View.js';
...
 
+const result = parseLiteralStyle({
+  'fill-color': ['get', 'fillColor'],
+  'stroke-color': ['get', 'strokeColor'],
+  'stroke-width': ['get', 'strokeWidth'],
+  symbol: {
+    symbolType: 'circle',
+    size: 8,
+    color: '#777',
+  },
```
tschaub, line 22: Not important to change anything here, but this is what this style literal would look like when passed to the layer:
```js
{
  'fill-color': ['get', 'fillColor'],
  'stroke-color': ['get', 'strokeColor'],
  'stroke-width': ['get', 'strokeWidth'],
  'circle-radius': 8,
  'circle-fill-color': '#777',
}
```

tschaub (approved):

tschaub:
The changes here look great, @jahow.  Really nice to see this progressing.

I opened #14780 with a proposal for the new styling API.  I'm sure those ideas will continue evolving, but wanted to give visibility into what I'm thinking.  I'd like to get that to a point where it is a replacement (more or less) for style functions and the style/symbolizer constructors in the Canvas renderer.  And want to make sure it feels reasonable on the WebGL side as well.

examples/webgl-vector-tiles.js
jahow: Yes, hopefully the webgl renderers can align with this format soon! Thanks for the hint

src/ol/render/webgl/VectorStyleRenderer.js
jahow: Good point, I realize now that my intentions were more to offer the possibility to use the ShaderBuilder instead of a literal style, and then be able to write custom expressions for color, width, etc. thus offering much more freedom while still providing an abstraction layer on top of GLSL. This could ease things up when moving to WebGL2 for example.

Maybe the ShaderBuilder could be made more usable and become a first class citizen of the API as an alternative to literal styles?

jahow:
Thanks for the quick review @tschaub! I'll look into https://github.com/openlayers/openlayers/pull/14780 right next.

PR #14766, opened 2023-05-22
Ensure circle coordinates are always [cx, cy, cx + r, cy]

ahocevar:
This pull request fixes the way `transform()` and `applyTransform()` work on `Circle` geometries. The internal representation of a circle geometry is two points, one being the point's center `[cx, cy]`, and the other being a point "right" of the center, by the amount of the circle's radius, i.e. `[cx + r, cy]`. Previously, the `transform()` and `applyTransform()` functions caused this contract to break.

Addresses https://github.com/openlayers/openlayers/pull/14744#discussion_r1193215641.

### DIFF
--- a/src/ol/geom/Circle.js
+++ b/src/ol/geom/Circle.js
@@ -2,6 +2,7 @@
  * @module ol/geom/Circle
  */
 import SimpleGeometry from './SimpleGeometry.js';
+import {METERS_PER_UNIT, transform} from '../proj.js';
 import {createOrUpdate, forEachCorner, intersects} from '../extent.js';
 import {deflateCoordinate} from './flat/deflate.js';
 import {rotate} from './flat/transform.js';
@@ -171,15 +172,7 @@ class Circle extends SimpleGeometry {
    * @api
    */
   setCenter(center) {
-    const stride = this.stride;
-    const radius = this.flatCoordinates[stride] - this.flatCoordinates[0];
-    const flatCoordinates = center.slice();
-    flatCoordinates[stride] = flatCoordinates[0] + radius;
-    for (let i = 1; i < stride; ++i) {
-      flatCoordinates[stride + i] = center[i];
-    }
-    this.setFlatCoordinates(this.layout, flatCoordinates);
-    this.changed();
+    this.setCenterAndRadius(center, this.getRadius());
   }
 
   /**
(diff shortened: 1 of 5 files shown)


mike-000:
A projection unit in one projection is not the same as that unit in another projection.  For a small radius circle I would expect

`new Circle(fromLonLat([16, 60], 'EPSG:32633'), 100)`

to look similar to

`new Circle(fromLonLat([16, 60], 'EPSG:3857'), 200)`

when rendered in their respective projections.

The current transform does that.  If this does not it is a significant breaking change.

ahocevar:
@mike-000 What you're saying does not make sense. You're showing Circle constructors in your code snippet, but are talking about transforms. The parameters of a `Circle` geometry are a cartesian coordinate, i.e. the point's center, and a distance, i.e. the point's radius. Distances in `ol/geom/*` are always cartesian distances on a plane, and as you know, neither Transverse Mercator nor Mercator projections are suitable for maintaining distance. So anything you get after a transform is not a correct distance, but just something you may have expected or not. Since we did not have any documented or tested expectation, this is not a breaking change.

mike-000:
Relative distances **are** generally maintained over small distances (e.g. 20 km local delivery radius) regardless of the projections unless the x and y axes have different scales such as in EPSG:4336.

These currently look similar when rendered in EPSG:3857

```
new Circle(fromLonLat([16, 60], 'EPSG:32633'), 100).transform('EPSG:32633', 'EPSG:3857')

fromCircle(new Circle(fromLonLat([16, 60], 'EPSG:32633'), 100)).transform('EPSG:32633', 'EPSG:3857')

```
As the radius defines the circumference I think most people would expect that, and continue to without having to convert another geometry type - if it works for the polygon it should work for the circle.

mike-000:
That is equivalent to

`new Circle(fromLonLat([16, 60], 'EPSG:32633'), 100).transform('EPSG:32633', 'EPSG:3857')
`

and

`new Circle(fromLonLat([16, 60]), 100)
`

and no, I do not think anyone would expect them to look similar in EPSG:3857 at that latitude.

mike-000:
Expect behavior when transforming all features including a circle between a local projection and EPSG:3857 at a northern latitude
https://codesandbox.io/s/projection-and-scale-forked-6tg0uc?file=/main.js

ahocevar:
@mike-000 What would be the benefit? The bigger the radius, the more inaccurate this will get.

mike-000:
> @mike-000 What would be the benefit? The bigger the radius, the more inaccurate this will get.

If you had a large radius circle you should be using spherical geometry.  There has never been any problem with using small radius Circle geometry (typically set up using `new Circle(center, trueRadius / getPointResolution('EPSG:3857', 1, center))`) and transforming it seemlessly if required.  At 10 km radius as in the example there is currently minimal difference between the two projections, and I presume that has been the case ever since version 3.

PR #14780, opened 2023-05-25
Expression based styling

tschaub:
## Overview

This is a branch working toward a new unified styling API.  The goal is provide a consistent and flexible way to style vector and raster layers that allows for efficient implementations with our current and future renderers.

Examples demonstrating style expressions:
 * [Style Expressions](https://deploy-preview-14780--ol-site.netlify.app/en/latest/examples/style-expressions.html)
 * [Vector Layer](https://deploy-preview-14780--ol-site.netlify.app/en/latest/examples/vector-layer.html)

See also the updated [API docs](https://deploy-preview-14780--ol-site.netlify.app/en/latest/apidoc/module-ol_style_flat).

## Background

Things we want to improve upon:
 * The current API for styling vector data is very flexible (allowing arbitrary work in a style function), but is difficult to make efficient (style functions need to be run on the main thread in the same context as the rest of the app).
 * The recently added "flat" style syntax is a nice terse alternative to using style and symbolizer constructors, but they lack the ability to evaluate expressions or use feature properties.  And the flat syntax only works when all features can be styled in the same way.

## Proposal

This branch proposes to:
 * Build upon the flat style syntax, adding support for expressions as used by the WebGL tile and vector renderers.
 * Add the ability to specify filters so that styles can be applied to subsets of features (without repeating conditional expressions in all property values).
 * Share parsing logic and style structures between the Canvas and WebGL renderers.

The idea is that vector and some raster layers will accept a similarly shaped `style` object literal.  This object will be "built" in two steps: parse and compile.

Logic for the parse step can be shared between the different renderers.  The result of parsing an encoded style is an instance of a class not exposed in the API and a parsing context (with information about things like properties and variables accessed in the expressions).

The compile step will be renderer-specific.  The Canvas renderer will compile style instances into functions that will be evaluated during rendering (like the current style functions).  Initially, building and evaluating these functions will be done on the main thread.  Eventually, this can be done in a worker (unlike the current style functions).

The compile step for the WebGL renderer will result in shader code (as it does for current style expressions).

When evaluating a built style, the evaluation context will include just the required properties and variables (available from the parsing context).  So a complete feature or additional context are not needed during evaluation.

Parsing and compiling happens infrequently (at layer construction, when calling `layer.setStyle()`).  Evaluation happens frequently (each render frame for each feature).

## Example Syntax

A style can be specified with a single symbolizer, an array of symbolizers, or an array of rules.

A symbolizer is an object literal that includes properties for strokes, fills, markers, and labels.  Property values may be literals (e.g. a number for stroke width, a string for a fill color, etc.) or expressions that are evaluated to determine the property value.

Here is an example symbolizer with literal property values:

```js
{
  'stroke-color': 'orange',
  'stroke-width': 10,
  'fill-color': [200, 200, 0],
}
```
Here is an example symbolizer that uses expressions to determine the property values:

```js
{
  'stroke-color': 'orange'
  'stroke-width': ['*', 2, ['get', 'lanes']],
}
```

Expressions can also be used to represent conditions:

```js
{
  'stroke-color': ['match', ['get', 'type'], 'highway', 'orange', 'black'],
  'stroke-width': ['match', ['get', 'type'], 'highway', 5, 2],
}
```

For a small number of properties and conditions, the `match` expression above works well.  However, it makes it difficult to represent multiple conditions and requires a lot of repetition for many related symbolizer properties.

To address this, we can support rules.

A rule is an object with a `filter` property and a `style` property.  The `style` is a symbolizer as described above.  The `filter` is an expression that is evaluated to determine if the symbolizer should be applied.  Rules are like CSS rules, where the filter is like a CSS selector.

```js
[
  {
    filter: ['==', ['get', 'type'], 'highway'],
    style: {
      'stroke-color': 'orange',
      'stroke-width': 5,
    },
  },
  {
    filter: ['!=', ['get', 'type'], 'highway'],
    style: {
      'stroke-color': 'black',
      'stroke-width': 2,
    },
  },
]
```

A rule could have an `else` property to indicate that it applies if the above set of rules (up to the previous `else`) do not apply:

```js
[
  {
    filter: ['==', ['get', 'type'], 'highway'],
    style: {
      'stroke-color': 'orange',
      'stroke-width': 5,
    },
  },
  {
    else: true,
    style: {
      'stroke-color': 'black',
      'stroke-width': 2,
    },
  },
]
```

Arrays of symbolizers can be used to render a feature multiple times.  This can be used to create cased lines, for example.  After evaluating rules, all symbolizers are applied in painter's order: all features are rendered first with the first symbolizer, then again with the second symbolizer, and so on.

```js
[
  {
    filter: ['==', ['get', 'type'], 'highway'],
    style: [
      {
        'stroke-color': 'orange',
        'stroke-width': 5,
      },
      {
        'stroke-color': 'red',
        'stroke-width': 3,
      },
    ]
  },
  {
    else: true,
    style: {
      'stroke-color': 'black',
      'stroke-width': 2,
    },
  },
]
```

See also the [new rendering test](https://github.com/openlayers/openlayers/blob/c7c08401a97912abc484efc5f8cfcee50f33e26b/test/rendering/cases/vector-style-rules/main.js) demonstrating a set of rules like the one above.


jahow:
Thanks for all the work on this @tschaub! To me this style format looks great but I was convinced from the start.

About the parse step, I was wondering if instead of throwing errors whenever something is not as expected we could instead list errors or "problems" encountered while parsing and store them in the context? Then at the end of this step we would throw a specific `StyleParseError` or similar containing all problems encountered with some information on each (like maybe line/col in the formatted JSON).

tschaub:
I've marked this as ready for review.  There is more work to do, but I think the foundation is solid.  If it sounds ok to others, I think this could be merged and we could continue to iterate on it with separate pull requests.

I haven't made any changes to WebGL rendering.  For Canvas vector layer rendering, I've added support for constructing a layer with a flat style, an array of flat styles, or an array of rules.  Expressions can be used for symbolizer properties where we currently have setters.  For example, since we have `icon.setAnchor()`, the `icon-anchor` property can be an expression, but since we don't have `icon.setSrc()`, the `icon-src` property cannot be an expression.

For the Canvas renderer, compiled expressions are wrapped in a style function that conforms with the existing interface.  This means that the function doesn't have access to things like zoom level.  And all of the parse, compile, and evaluation work is still done on the main thread.  If we later drop support for style functions, we can move this work to a worker if appropriate.

I've added support for `get`, `var`, `==`, `!=`, `>`, `>=`, `<`, `<=`, `+`, `-`, `*`, and `/` operators.  It should be straightforward to continue adding to this list based on the existing patterns.

@jahow - I agree that it could be nice to compile a list of all errors while parsing and throw some composite error, but I've left this for later (for now, we fail fast on the first parsing error).

ahocevar (approved):
Thanks @tschaub for walking me through the code, for the final touches, and for this solid foundation for expression based styling!

jahow:
Insane work, thank you both. Looking forward to get that to work for webgl rendering as well!

MoonE:
> I've added support for `get`, `var`, `==`, `!=`, `>`, `>=`, `<`, `<=`, `+`, `-`, `*`, `/`, `clamp`, `%`, `^`, `abs`, `floor`, `ceil`, `round`, `sin`, `cos`, `atan`, `sqrt`, and `match` operators. It should be straightforward to continue adding to this list based on the existing patterns.

@tschaub How does one use a `var` expression with the canvas renderer?
Seems like there is no way to provide tem to the evaluation context.

Properties and resolution are set here, but not variables?
https://github.com/openlayers/openlayers/blob/70e6da27c98c762e77abb93e0d557012e717a843/src/ol/render/canvas/style.js#L71-L80
https://github.com/openlayers/openlayers/blob/70e6da27c98c762e77abb93e0d557012e717a843/src/ol/render/canvas/style.js#L101-L115

tschaub:
> @tschaub How does one use a `var` expression with the canvas renderer? Seems like there is no way to provide tem to the evaluation context.

Yeah.  I started wondering about doing this at the map level instead of per-layer and then never decided.  I opened #15146 to consider alternatives.

PR #14790, opened 2023-06-02
Miscellaneous rich text documentation fixes

mike-000:
Document that rich text is not supported for `placement: 'line'` (and use quotes in `'text-placement': 'line'` as required by JavaScript elsewhere in docs).  Remove inappropriate tag from example which does not use rich text.


ahocevar (approved):
Thanks, @mike-000

PR #14791, opened 2023-06-03
Fix ufos example with type hint

MoonE:
https://openlayers.org/en/latest/examples/icon-sprite-webgl.html no longer works with 7.4.0 because of a required type hint.


ahocevar (approved):
Thanks, @MoonE

jahow:
Thanks for the fix @MoonE , I also noticed very recently that the example was broken. Ideally a type hint should not be required here since the type can be inferred from the variable initial value, but addressing that is more involved so this fix makes sense in the meantime. I'm working on improving type inference for the `var` operator, that should come in another PR soon.

Issue #13739, opened 2022-06-08
Add More Style Expression Operator like 'sqrt', 'distance'


linjiacheng:
I have a geotiff with two bands, u and v bands to calculate wind speed, can you provide more operator like 'sqrt', so I can use expression like
`['sqrt', ['+', ['^', ['band', 1], 2], ['^', ['band', 2], 2]]]`

ref:  https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#sqrt

mike-000:
Where not supported `sqrt` can be implemented as
`['^', ['+', ['^', ['band', 1], 2], ['^', ['band', 2], 2]], 0.5]`

linjiacheng:
Silly me

jahow:
Hi, I think such operator would make a lot of sense and help simplify expressions/make them more readable. Besides, we're still trying to be as compatible as possible with mapbox expressions.

@linjiacheng would you be willing to help with such a contribution? A good reference would be this PR: https://github.com/openlayers/openlayers/pull/13363

Issue #11308, opened 2020-07-22
Style Expression Operator 'in'


RydingM:
Trying to use an array variable in the literal style and for filtering what to be able to check a features property is a value within the variable array. I couldn't see any operator that would allow me to this currently. I had a quick go at trying to add it into my local, following mapbox documentation ( https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#in ), but ran into issues. 

It would be great to see this as an option for the literal style.

Pipoupi:
Did you find a way to do it? I'd like to implement a search box allowing multiple requests, and I can't find a way to perform it.

RydingM:
> Did you find a way to do it? I'd like to implement a search box allowing multiple requests, and I can't find a way to perform it.

After trying for a couple of hours I was unsuccessful, due to my time constraints I ended up having to work around the issue using an inefficient way of filtering out the points from the display (especially when there's a lot of points). 

The workaround I ended up doing for the time being was adding a new property to the features called "show" which is a boolean and added the logic for if they should be shown in the literal style based off this. To alter the display I would then loop through all the features and change the property "show" to true or false.

RydingM:
> Did you find a way to do it? I'd like to implement a search box allowing multiple requests, and I can't find a way to perform it.

I didn't manage to get this functionality into the expressions and due to time constraints I wasn't able to continue with it. I just used a workaround which isn't great as it involves updating the properties on the features called "show" that is boolean and have logic in the style if it should be displayed. But I have seen from other issue posts that updating the features causes a full rebuild of the buffers.

PR #14803, opened 2023-06-06
Add support for 'in' and 'sqrt' operators in expressions

jahow:
This PR implements two new operators taken from the MapBox style spec:
* [`in`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#in) takes two arguments: a "needle" and a "haystack", and will return true if the haystack contains the needle; currently has the following limitations:
  * cannot be used to check the presence of a substring in a string
  * **only supports array literals as haystack**:
    * for instance, this is valid since it can be identified as a number array:  
    `['in', ['get', 'factor'], [2, 5, 10]]`  
    * this will fail because a string array can either be an array literal or an expression:  
    `['in', ['get', 'type'], ['road', 'path', 'street']]`  
    * to disambiguate a string array literal, [use the `literal` operator](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#types-literal):
    `['in', ['get', 'type'], ['literal', ['road', 'path', 'street']]]`  
    * lastly, this will throw an error since dynamic haystacks are not yet supported:  
    `['in', ['get', 'type'], ['var', 'supportedTypes']]`  

* [`sqrt`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#sqrt): self explanatory

The current limitations on the `in` operator will be really hard to work around with the current system. On the other hand, the approach introduced in #14780 will give much more flexibility and should help remove this limitations. As a result I figured I would not invest too much efforts in the current system, and the `in` operator as it is should already be very helpful for static filtering.

Addresses https://github.com/openlayers/openlayers/issues/11765, https://github.com/openlayers/openlayers/issues/11308, https://github.com/openlayers/openlayers/issues/13739

Replaces https://github.com/openlayers/openlayers/pull/11839

### DIFF
--- a/src/ol/style/expressions.js
+++ b/src/ol/style/expressions.js
@@ -74,6 +74,10 @@ import {asArray, fromString, isStringColor} from '../color.js';
  *   * `['any', value1, value2, ...]` returns `true` if any of the inputs are `true`, `false` otherwise.
  *   * `['between', value1, value2, value3]` returns `true` if `value1` is contained between `value2` and `value3`
  *     (inclusively), or `false` otherwise.
+ *   * `['in', input, [value1, value2, ...valueN]]` returns `true` if `input` equals any of the provided values, and
+ *     `false` otherwise; only supports number, string or boolean values.
+ *     Note that the second argument must be a fixed-size array literal and cannot be the result of an expression
+ *     (this is a known limitation as arrays of size > 4 are currently not supported in expressions).
  *
  * * Conversion operators:
  *   * `['array', value1, ...valueN]` creates a numerical array from `number` values; please note that the amount of
@@ -252,6 +256,15 @@ function printTypes(valueType) {
  * @property {Array<PaletteTexture>} [paletteTextures] List of palettes used by the style.
  * @property {import("../style/literal").LiteralStyle} style The style being parsed
  */
+
+/**
+ * @param {string} operator Operator
+ * @param {ParsingContext} context Parsing context
+ * @return {string} A function name based on the operator, unique in the given context
+ */
+function computeOperatorFunctionName(operator, context) {
+  return `operator_${operator}_${Object.keys(context.functions).length}`;
+}
 
 /**
  * Will return the number as a float with a dot separator, which is required by GLSL.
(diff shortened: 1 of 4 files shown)


jahow:
The `layer-mapbox-vector` rendering test was failing, it was apparently close to the threshold for a while. I assume this is because of various improvements related to style parsing in `ol-mapbox-style`, so I updated the expected render. @ahocevar does that make sense to you?

src/ol/style/expressions.js
```diff
@@ -868,6 +882,17 @@ Operators['atan'] = {
...
 };
 
+Operators['sqrt'] = {
+  getReturnType: function () {
+    return ValueTypes.NUMBER;
+  },
+  toGlsl: function (context, args) {
+    assertArgsCount(args, 1);
+    assertNumbers(args);
+    return `sqrt(${expressionToGlsl(context, args[0])})`;
```
gberaudo, line 892: If the value is always static, an alternative is to compute the expression here once and for all instead of passing it to the shader where it will be computed many times.

```suggestion
    return expressionToGlsl(context, Math.sqrt(args[0]))`;
```

mike-000: See #13739 for requirement

gberaudo: Thanks for the inputs.

jahow:
I re-read https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#type-system and I'm realizing that we're drifting away from the type system used in mapbox styles. For example in this PR the `in` operator shouldn't just assume that the haystack is a literla array, there should be a `literal` operator to declare that explicitly, otherwise later on we will not be able to support dynamically-computed haystacks (e.g. `['in', ['get', 'type'], ['var', 'allowedTypes']]`).

Would be curious to hear other's opinions on this!

gberaudo:
> I re-read https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#type-system and I'm realizing that we're drifting away from the type system used in mapbox styles. For example in this PR the `in` operator shouldn't just assume that the haystack is a literla array, there should be a `literal` operator to declare that explicitly, otherwise later on we will not be able to support dynamically-computed haystacks (e.g. `['in', ['get', 'type'], ['var', 'allowedTypes']]`).
> 
> Would be curious to hear other's opinions on this!

My opinion is that we should require that `literal` operator, just like plain Mapbox style.
As you said, that is a requirement to support haystacks defined by an expression.

jahow:
> My opinion is that we should require that `literal` operator, just like plain Mapbox style.
> As you said, that is a requirement to support haystacks defined by an expression.

Changed the implementation to require disambiguation in case of a string array literal, using `literal` as per the Mapbox spec. This is all a little bit hackish for now, as I said I'm looking forward to switching to the approach in #14780 to have greater flexibility in implementing operators.

tschaub (approved):

tschaub:
Thanks for the work on this.

PR #14805, opened 2023-06-06
Update reference image for changed type system on CI

ahocevar:
This pull request updates a reference image that was affected by changes to the typography system on the docker image used by our CI for the rendering tests.


jahow (approved):
That's the reason!

PR #14806, opened 2023-06-06
Rework webgl points hit detection

gberaudo:
Together with @jahow we are in the process of refactoring/implementing hit detection in all the WebGL renderers.
This PR is the first step, focused on refactoring the existing WebGL points hit detection mechanism by simplifying and optimizing it.

Before this PR, plain and hit detection rendering had redundant code paths: 2 array buffers, 2 vertex shaders, 2 fragment shaders, 2 test cases... This made it difficult to maintain and also consummed more resource than striclty necessary.

With this PR we introduce a unified approach for WebGL points rendering, driven by a single uniform.

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -50,8 +50,7 @@ import {listen, unlistenByKey} from '../../events.js';
  * Please note that these can only be numerical values.
  * @property {string} vertexShader Vertex shader source, mandatory.
  * @property {string} fragmentShader Fragment shader source, mandatory.
- * @property {string} [hitVertexShader] Vertex shader source for hit detection rendering.
- * @property {string} [hitFragmentShader] Fragment shader source for hit detection rendering.
+ * @property {boolean} [hitDetectionEnabled] Whether shader is hit detection aware.
  * @property {Object<string,import("../../webgl/Helper").UniformValue>} [uniforms] Uniform definitions for the post process steps
  * Please note that `u_texture` is reserved for the main texture slot and `u_opacity` is reserved for the layer opacity.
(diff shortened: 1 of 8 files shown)


src/ol/renderer/webgl/PointsLayer.js
```diff
@@ -639,29 +562,14 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
       id: ++this.lastSentId,
       type: WebGLWorkerMessageType.GENERATE_POINT_BUFFERS,
       renderInstructions: this.renderInstructions_.buffer,
-      customAttributesSize: this.customAttributes.length,
+      customAttributesSize: singleInstructionLength - 2,
     };
     // additional properties will be sent back as-is by the worker
     message['projectionTransform'] = projectionTransform;
+    message['hitDetection'] = this.hitDetectionEnabled_;
```
jahow, line 569: I think we don't need that anymore?

src/ol/webgl/Helper.js
```diff
@@ -713,6 +714,15 @@ class WebGLHelper extends Disposable {
     ]);
   }
 
+  /**
+   * Sets the `u_hitDetection` uniform.
+   * @param {boolean} enabled Whether to enable the hit detection code path
+   */
+  applyHitDetectionUniform(enabled) {
+    const loc = this.getUniformLocation('u_hitDetection');
```
jahow, line 722: maybe we should use `DefaultUniform.HIT_DETECTION` here?

jahow (approved):
Thanks, that approach brings so much simplification! I think it's great. We could probably use `packColor` on the hit detection color attribute in order to use only two numbers instead of four, but that can be done later on.

I have only a couple of minor remarks, nothing else to add.

gberaudo:
Hi @jahow, thanks for the review. I pushed the changes you asked for. If it looks good to you then please merge because I don't have write access.

PR #14810, opened 2023-06-07
send 'loadend' event on layer image load failure

### LINKED ISSUE #14808: Map never sends 'loadend' event when loading one of its layer images fails
**Describe the bug**
In a map with a WMS layer (other layer types might be affected, too), if there is an error loading the layer image (e.g. because of a network issue), the map never sends the 'loadend' event.

**To Reproduce**
1. Create a map with a WMS layer.
2. Simulate an error loading the layer image by using a custom imageLoadFunction, e.g.
_imageWMSSource.setImageLoadFunction(image.getImage().dispatchEvent(new CustomEvent('error'));_
3. The map never finishes loading. If you have some loading indicator (e.g. a spinner) that waits for the "loadend" event of the map, it never goes away.

**Expected behavior**
The map finishes loading (sends the 'loadend' event) when all of its layers have finished loading, no matter if the layers could be loaded successfully or if there was an error in one or more layers.

I have a merge request prepared in case you agree that it is a bug. Changing the line
_if (image.getState() === ImageState.LOADED) {_
in renderer/Layer.js/handleImageChange_(event) to
_if (image.getState() === ImageState.LOADED || image.getState() === ImageState.ERROR) {_
fixes the problem.

Tobias43:
Before this fix the map's 'loadend' event was never fired when an error loading a WMS image occured.

The check whether the map finished loaded happens in Map.renderFrame_. That method gets triggered every time something relevant changes in a layer, e.g. when a WMS image finishes loading. But what was missing was the triggering of the rendering when the loading of a WMS image fails.

fixes #14808

### DIFF
--- a/src/ol/renderer/Layer.js
+++ b/src/ol/renderer/Layer.js
@@ -152,7 +152,7 @@ class LayerRenderer extends Observable {
    */
   handleImageChange_(event) {
     const image = /** @type {import("../Image.js").default} */ (event.target);
-    if (image.getState() === ImageState.LOADED) {
+    if (image.getState() === ImageState.LOADED || image.getState() === ImageState.ERROR) {
       this.renderIfReadyAndVisible();
     }
   }
--- a/src/ol/renderer/Layer.js
+++ b/src/ol/renderer/Layer.js
@@ -152,7 +152,10 @@ class LayerRenderer extends Observable {
    */
   handleImageChange_(event) {
     const image = /** @type {import("../Image.js").default} */ (event.target);
-    if (image.getState() === ImageState.LOADED || image.getState() === ImageState.ERROR) {
+    if (
+      image.getState() === ImageState.LOADED ||
+      image.getState() === ImageState.ERROR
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @Tobias43

src/ol/renderer/Layer.js
```diff
@@ -152,7 +152,7 @@ class LayerRenderer extends Observable {
    */
   handleImageChange_(event) {
     const image = /** @type {import("../Image.js").default} */ (event.target);
-    if (image.getState() === ImageState.LOADED) {
+    if (image.getState() === ImageState.LOADED || image.getState() === ImageState.ERROR) {
```
ahocevar, line 155:
```suggestion
    if (
      image.getState() === ImageState.LOADED ||
      image.getState() === ImageState.ERROR
    ) {
```

Tobias43:
Thank you very much for the quick merge @ahocevar !

PR #14816, opened 2023-06-09
Remove pointer events polyfill from examples

### LINKED ISSUE #14703: DragPan example doesn't work with Safari on iOS 16.4.1
**Describe the bug**
DragPan with Safari on Apples iPhone iOS 16.4.1 doesn't work any more. Previous version of iOS or other browsers on Android aren't affected. Example works on iPadOS with 16.4.1 too.
Using two fingers (put on the same time) to drag the map, the map isn't moved.
Using two fingers (put with a delay), the map can be dragged.

**To Reproduce**
1. Go to example [Panning and page scrolling](https://openlayers.org/en/latest/examples/two-finger-pan-scroll.html)
2. Drag with two fingers
3. Panning not working on iOS 16.4.1

**Expected behavior**
DragPan with two fingers works on iOS 16.4.1 to drag map.

ahocevar:
Maybe fixes #14703 

I think it's now been long enough to wait for browsers without pointer events support to disappear.


mike-000:
I presume you did not intend to add the Circle example via this PR?

ahocevar:
Thanks, @mike-000. This was indeed unintentional.

jahow (approved):
Looks like Safari was the last to add support and that was 4 years ago so, makes sense!

By the way `elm-pep` is mentioned here: https://github.com/openlayers/openlayers/blob/b80fc198cf26a158485b9365856d899faee15415/src/ol/MapBrowserEventHandler.js#L362-L363

I guess it doesn't hurt to keep this workaround but maybe rephrasing the comment would make sense since elm-pep is not used anymore in the examples?

Issue #14541, opened 2023-02-28
Make Bing "placeholder" tiles optional


mike-000:
**Is your feature request related to a problem? Please describe.**
Currently `Aerial` and `Ordnance Survey` styles show placeholder tiles when there is no data because the metadata image url does not include `&n=z`.  The `Aerial with Labels` image url includes `&n=z` which produces an empty response instead of a placeholder, and when `useInterimTilesOnError` and `preload` layer options are set this causes the nearest available preloaded zoom level to be used instead.  If `&n=z` is removed from the tile requests `Aerial with Labels` will also display placeholder tiles.

**Describe the solution you'd like**
This is similar to setting the `blankTile=false` url option in ESRI XYZ sources to produce an error response instead of a "no information to display" tile.  It would be nice if it was as easily configurable for Bing sources (in the same was as `&dpi=d1` is added to tile requests for the `hidpi` option) if a `placeholder` option was explicitly set `true` or `false` (otherwise the metadata defaults would continue to be used).

PR #14832, opened 2023-06-15
Optionally show or hide Bing 'placeholder' tiles

Shane98c:
Fixes https://github.com/openlayers/openlayers/issues/14541

Placeholder tiles are shown by default for different styles as described in the issue. The behavior is controlled by the presence of the `n=z` url param. This PR adds an optional parameter allowing users to decide whether to show these tiles or, alternatively, overzoom the best available resolution. If the parameter is not set, the default behavior for the style is used.

This is a fairly big improvement from the method of using maxzoom to hide placeholder tiles mentioned in the example because Bing's zoom level coverage differs across the globe, so while a maxzoom of x works to hide the tiles in the US, a different one is required in, eg, Spain.

### DIFF
--- a/src/ol/source/BingMaps.js
+++ b/src/ol/source/BingMaps.js
@@ -266,21 +266,20 @@ class BingMaps extends TileImage {
               tileCoord[2],
               quadKeyTileCoord
             );
-            let url = imageUrl;
+            const url = new URL(
+              imageUrl.replace('{quadkey}', quadKey(quadKeyTileCoord))
+            );
+            const params = url.searchParams;
             if (hidpi) {
-              url += '&dpi=d1&device=mobile';
+              params.set('dpi', 'd1');
+              params.set('device', 'mobile');
             }
-            url = url.replace('{quadkey}', quadKey(quadKeyTileCoord));
-            const uri = new URL(url);
-            const params = uri.searchParams;
-            if (placeholderTiles === true && params.has('n')) {
+            if (placeholderTiles === true) {
               params.delete('n');
-              url = uri.toString();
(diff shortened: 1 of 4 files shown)


src/ol/source/BingMaps.js
```diff
@@ -261,7 +270,17 @@ class BingMaps extends TileImage {
             if (hidpi) {
               url += '&dpi=d1&device=mobile';
             }
-            return url.replace('{quadkey}', quadKey(quadKeyTileCoord));
+            url = url.replace('{quadkey}', quadKey(quadKeyTileCoord));
+            const uri = new URL(url);
+            const params = uri.searchParams;
```
jahow, line 275: Even though the implementation works, I think it is a bit awkward: we start by manipulating the URL as a string, then switch to the `URL` class. I would suggest working with a `URL` object from the start (except for the `{quadkey}` token replacement), and return `url.toString()` in the end. Would that make sens to you?

jahow (commented):
Thank you for the contribution! That feature is definitely useful. I've made a couple of remarks regarding the implementation, because I think it can be simplified.

Do you think you'd be able to add a test covering this feature here? https://github.com/openlayers/openlayers/blob/b80fc198cf26a158485b9365856d899faee15415/test/browser/spec/ol/source/BingMaps.test.js#L25

Thanks again

Shane98c:
Thanks for the thoughtful review @jahow! All makes sense.

I've made the suggested changes and added a few basic tests. I'm new to using karma, so any pointers on that front would be very welcome.

jahow (approved):
I think it's better now, thanks!

PR #14837, opened 2023-06-16
Expressions / add `geometry-type` operator, fix exponential `interpolate`

jahow:
This PR:
* adds a new [`geometry-type`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#geometry-type) operator; this operator returns the current geometry type of the feature, either `Point`, `Polygon` or `LineString`; other types (multi*, collections) are resolved to one of these three (see doc)
* reworks the [`interpolate`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#interpolate) computation when the interpolation type is `exponential`; the computation is now compliant with the MapBox style spec and allows for example scaling values according to the current zoom using `['exponential', 2]`. The linear interpolation is unchanged.

**Note: the change on the `interpolate` operator can be considered as breaking since any style using it might notice a difference in the final result.** The difference can be quite significant depending on the range. Here is a visual representation of the curve before and after for an interpolation of type `['exponential', 2]`, input range 5..15, output range 0..100:

Before:
image
   
After:
image

On the other hand it is now compliant to the reference implementation so this can also be considered as fixing an erroneous behavior.


tschaub (approved):
Looks great, @jahow

jahow:
thanks for the quick review @tschaub!!

PR #14839, opened 2023-06-16
Use image resize function from ImageStatic when ImageWrapper loads

mike-000:
Alternative to #14826 which uses a single draw to canvas taking both `imageSize` and `imageExtent` into account instead of resizing the image twice (which has quality implications).  This is more like the behavior before #14444 but always produces a canvas.


ahocevar:
I think it was a mistake to make the changes I made in #14444. I'm currently thinking about a different way to create Image layers with images that don't need to be decoded on every render frame.

mike-000:
A canvas would still be needed if the aspect ratios so not match, and scaling down to reduce canvas use or up see more detail in a smaller natural size SVG will still be useful in that case.  Although for huge SVGs with a lot a detail I do not think ImageStatic is the best approach even if you want to render to canvas.  It would be easy to tile in both canvas layers https://codesandbox.io/s/svg-flag-canvas-tiles-292mps?file=/main.js and WebGL https://codesandbox.io/s/svg-flag-data-tiles-5nyscm?file=/main.js

ahocevar:
The original issue #14826 tried to fix is that due to the size of the used SVG image, nothing was rendered because the resulting canvas was too big. I only liked the idea of giving the `imageSize` property a purpose, but that's not something we really need.

I agree that `ImageStatic` is not a good way to work with SVG images. But the same can happen with non-SVG images, so the state of image handling after #14444 is problematic.

mike-000:
Replaced by #14945

PR #14853, opened 2023-06-21
Avoid missing tiles for Raster source using Tile layers or sources

ahocevar:
This pull request fixes an issue with the Raster source, which can be seen in the map at the top of https://openlayers.org/: Sometimes not all tiles of the hillshade are displayed. To see this bug, simple reload the website several times - it will reveal itself eventually.

A second chunk of this change removes an unnecessary `layer.changed()` call, when we're waiting for an animation of dependent sources to finish.

### DIFF
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -587,9 +587,12 @@ class RasterSource extends ImageSource {
      */
     this.layers_ = createLayers(options.sources);
 
-    const changed = this.changed.bind(this);
     for (let i = 0, ii = this.layers_.length; i < ii; ++i) {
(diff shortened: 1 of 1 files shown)


tschaub (approved):

tschaub:
Thanks for digging into this, @ahocevar.

I notice that I still see the elevation tiles not render periodically when refreshing https://deploy-preview-14853--ol-site.netlify.app/.

tschaub:
With your latest commit, I think I'm always seeing the shaded relief render.  Looks good!

ahocevar:
I think so too. Thanks, @tschaub!

PR #14855, opened 2023-06-22
Mark image as changed in the next animation frame

ahocevar:
#14853 improved the situation regarding missing tiles for raster sources, and was a step in the right direction. With the additional change here, I think the issue is finally fixed.


ahocevar:
No, unfortunately repeatedly reloading https://deploy-preview-14855--ol-site.netlify.app/ still has missing tiles sometimes.

PR #14856, opened 2023-06-22
Another attempt to avoid missing tiles with the Raster source

ahocevar:
I'm opening this as work-in-progress. Locally, the issue does not occur, so I'm misusing the deploy preview to finish what I started in #14853.

### DIFF
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -599,9 +599,7 @@ class RasterSource extends ImageSource {
      * @private
      * @type {import("../TileQueue.js").default}
      */
-    this.tileQueue_ = new TileQueue(function () {
-      return 1;
-    }, this.changed.bind(this));
+    this.tileQueue_ = new TileQueue(() => 1, changed);
 
     /**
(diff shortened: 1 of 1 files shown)


ahocevar:
Finally I was able to fix this, by introducing a Promise that resolves when the workers have completed. The promise is used on the overridden `changed()` function, which calles the super's `changed()` function when the promise resolves. This should also improve performance, because it avoids the previous ping-pong between the source's `changed()` calls that trigger a `getImage()` call when `getImage()` does not have the data to render yet.

src/ol/source/Raster.js
```diff
@@ -815,18 +823,19 @@ class RasterSource extends ImageSource {
       if (imageData) {
         imageDatas[i] = imageData;
       } else {
-        return;
+        return Promise.resolve();
```
tschaub, line 826: Minor, but here and above it seems like we could reuse a single promise from `Promise.resolve()` created at the module level.

tschaub (approved):

Issue #14866, opened 2023-06-27
Opacity of the VectorImageLayer is applied to features individually instead of the whole layer


MrLynx93:
**Describe the bug**
Opacity of the VectorImageLayer is applied to features individually instead of the whole layer. When features individually are 100% opaque and the layer itself is semi-transparent, then:
- **On first render**, the layer is 100% opaque and features with the same color "merge" on the map.
- When you zoom/pan the map, then features become semi-transparent and they do not "merge" anymore - in the place where they overlap they are darker.
![before-pan](https://github.com/openlayers/openlayers/assets/3768864/573780b3-b1b9-4575-93a9-e1b0498f7007)
![after-pan](https://github.com/openlayers/openlayers/assets/3768864/5a12dc69-7cd6-4e4d-9a00-40993b54aa58)

**To Reproduce**

You can also check the effect here: https://jsfiddle.net/MrLynx93/08bcvygt/
1. Create a VectorImageLayer with 2 overlapping geometries. Set their style to be opaque
2. Create the layer's opacity to 0.5
3. Look at the map - it's fully opaque.
4. Pan/zoom the map - it becomes semi-transparent

```
const features = [
  new ol.Feature({
    geometry: new ol.geom.Circle([-1000, -1000], 2000),
    name: 'the buffer'
  }),
  new ol.Feature({
    geometry: new ol.geom.Circle([1000, 1000], 2000),
    name: 'the buffer',
  })
];
features.forEach(feature => feature.setStyle(
  new ol.style.Style({
    fill: new ol.style.Fill({
    color: 'green'
    })
  }))
);
const source = new ol.source.Vector();
source.addFeatures(features);

const vectorImageLayer = new ol.layer.VectorImage({
  source: source,
  opacity: 0.5
});

const map = new ol.Map({
  view: new ol.View({
    center: [0, 0],
    resolution: 20
  }),
  target: 'map',
  layers: [ vectorImageLayer ]
});
```

**Expected behavior**
The layer should be semi-opaque as a whole and features with the same color should "merge" there - same as they would if you used VectorLayer instead.

PR #14867, opened 2023-06-27
Fix VectorImageLayer opacity

MrLynx93:
Fixing the issue Issue #14866:
At the point context.globalAlpha was set, the context is the context of vectorRenderer, not global context for the whole render.

### DIFF
--- a/src/ol/renderer/canvas/VectorImageLayer.js
+++ b/src/ol/renderer/canvas/VectorImageLayer.js
@@ -107,7 +107,7 @@ class CanvasVectorImageLayerRenderer extends CanvasImageLayerRenderer {
       vectorRenderer.useContainer(null, null);
       const context = vectorRenderer.context;
       const layerState = frameState.layerStatesArray[frameState.layerIndex];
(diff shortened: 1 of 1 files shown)


src/ol/renderer/canvas/VectorImageLayer.js
```diff
@@ -107,7 +107,7 @@ class CanvasVectorImageLayerRenderer extends CanvasImageLayerRenderer {
       vectorRenderer.useContainer(null, null);
       const context = vectorRenderer.context;
       const layerState = frameState.layerStatesArray[frameState.layerIndex];
-      context.globalAlpha = layerState.opacity;
+      this.globalAlpha = layerState.opacity;
```
ahocevar, line 110: This change essentially means that no layer opacity will be applied at all, because there is no `globalAlpha` member property. Removing the line would have the same effect, and this would result in the same opacity behavior that we have on `ol/layer/Vector`. Is that your intention?

ahocevar (approved):
Thanks, @MrLynx93

PR #14870, opened 2023-06-30
Remove ol/layer/MapboxVector to avoid a circular dependency

ahocevar:
This pull request removes the circular `ol` <-> `ol-mapbox-style` dependency by removing `ol/layer/MapboxVector`. Since the creation of the MapboxVector layer, the `ol-mapbox-style`'s API has advanced a lot, so now it is very easy to consume a Mapbox-hosted vector style directly with `ol-mapbox-style`'s `applyStyle()` function. The `mapbox-vector-layer.html` example has been updated accordingly, and I added a minimal example for the required change in the upgrade notes.

This is a breaking change.


MoonE (commented):
There's a small error in the upgrade notes example code.

jahow (approved):
Thanks! One thing that may be worth mentioning is that listening to the `error` and `sourceready` events on such a layer will not work anymore?

tschaub:
Would it make sense to have something like the `MapboxVector` available in `ol-mapbox-style`?

ahocevar:
@tschau Oh yes! I'm going to do that.

ahocevar:
@tschaub Updated the pull request to use the brand new MapboxVectorLayer from `ol-mapbox-style@11`.

ahocevar:
@tschaub Is it ok for you if I merge this? ol-mapbox-style now has a `MapboxVectorLayer` named export, which is basically a copy of the removed `ol/layer/MapboxVector`.

tschaub (approved):
Thanks, @ahocevar

tschaub:
Looks like a rendering test is failing: https://github.com/openlayers/openlayers/actions/runs/5916085353/job/16042971717#step:5:43

Not in a place where I can check the artifacts.  Maybe an unrelated/ephemeral failure, but I noticed the same failed for another PR.

ahocevar:
@tschaub Looks unrelated, in both cases. I'm going to investigate.

ahocevar:
@tschaub #15032 should fix that rendering test issue.

PR #14883, opened 2023-07-06
WebGL / Rework symbol rendering to include circles with outlines, shapes and icons

jahow:
This PR significantly reworks how symbols are handled in the WebGL renderers. Symbol styles now use a format that is compatible with the "flat" style format used by Canvas renderers, and which looks like this:

```js
const iconStyle = {
  'icon-src': '../data/icon.png',
  'icon-width': 40,
  'icon-height': 80,
  'icon-offset': [5, 10],
  'icon-size': [20, 40],
};

const circleStyle = {
  'circle-radius': ['get', 'size'],
  'circle-fill-color': 'rgba(120, 140, 255, 0.6)',
  'circle-stroke-width': 3,
  'circle-stroke-color': 'blue',
};

const starStyle = {
  'shape-points': 4,
  'shape-radius1': 12,
  'shape-radius2': 5,
  'shape-fill-color': 'orange',
};
```

**The previous symbol style format is not supported anymore**: using it will simply not show anything on screen.

New styling possibilities include:
- circles with outline
- regular shapes or stars with or without outlines
- icons specified as path or `Image`, with displacement, offset, anchor etc.

image

Note: this will break existing applications using `WebGLPointsLayer` with the previous style format. I figured that since we're moving towards a new major version, it was the right time to introduce a breaking change (which only affects parts of the API marked as experimental anyway).

I added several rendering tests to better guarantee the output of the rendering in WebGL.

Remaining to do:
- [x] fix heatmap layer
- [x] fix remaining broken tests
- [x] update the change log

*This work is sponsored by BSW-Hamburg and Landesbetrieb Geoinformation und Vermessung Hamburg.*


jahow:
This is ready for review, please let me know if I need to subdivide some commits further to make this easier to read, thanks in advance!

tschaub (approved):

tschaub:
@jahow - this looks great!  I meant to review it earlier, but then lost track of it.

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,89 @@
 
 ### Next version
 
+#### [Breaking] Change of the symbol style format in `ol/layer/WebGLPointsLayer`
```
tschaub, line 5: I added a "Backwards incompatible changes" heading to [the upgrade notes](https://github.com/openlayers/openlayers/blob/fe8c2ebd5dfa9ba0e38af9bcc3352c0b617931b8/changelog/upgrade-notes.md).  Doesn't need to be in this change, but we should either group all the breaking changes under that header or add a prefix as you have done.

jahow: rebased and moved this to the top of the upgrade notes

PR #14895, opened 2023-07-12
Fix dynamic background color

### LINKED ISSUE #14894: Dynamic background color change doesn't work
**Describe the bug**
Dynamic background color change doesn't work as expected for vector layers.

**To Reproduce**
1. Run the following code
2. Zoom in/out
3. You'll see the background color is not changed as expected

```js
import './style.css';
import {Map, View} from 'ol';
import Graticule from 'ol/layer/Graticule';

function background(resolution) {
  const color = resolution < 10000 ? '#88f' : '#f88';
  console.log(resolution, color);
  return color;
}

const map = new Map({
  target: 'map',
  layers: [
    new Graticule({
      background,
    })
  ],
  view: new View({
    center: [0, 0],
    zoom: 0,
  })
});
```

**Expected behavior**
The background color is changed when the zoom level (resolution) is changed.

kikuchan:
Try to fix #14894

### DIFF
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -177,7 +177,7 @@ class CanvasLayerRenderer extends LayerRenderer {
       this.container = null;
       this.context = null;
       this.containerReused = false;
-    } else if (this.container && this.container.style) {
+    } else if (this.container) {
       this.container.style.backgroundColor = null;
     }
     if (!this.container) {
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -177,6 +177,8 @@ class CanvasLayerRenderer extends LayerRenderer {
       this.container = null;
       this.context = null;
       this.containerReused = false;
+    } else if (this.container && this.container.style) {
+      this.container.style.backgroundColor = null;
     }
     if (!this.container) {
       container = document.createElement('div');
--- a/test/browser/spec/ol/renderer/canvas/vectortilelayer.test.js
+++ b/test/browser/spec/ol/renderer/canvas/vectortilelayer.test.js
@@ -328,12 +328,17 @@ describe('ol/renderer/canvas/VectorTileLayer', function () {
     });
 
     it('changes background when function returns a different color', function (done) {
-      let first = true;
+      let count = 0;
       layer.setBackground(function (resolution) {
+        const backgrounds = [
+          undefined,
+          'rgba(255, 0, 0, 0.5)',
+          'rgba(0, 0, 255, 0.5)',
+          undefined,
+        ];
+
         expect(resolution).to.be(map.getView().getResolution());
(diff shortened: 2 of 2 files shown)


kikuchan:
I'm not sure what should happen if the background function returns `undefined` **again** in the test code, btw.

If you want something like this;
```diff
--- test/browser/spec/ol/renderer/canvas/vectortilelayer.test.js
+++ test/browser/spec/ol/renderer/canvas/vectortilelayer.test.js
@@ -334,6 +334,7 @@ describe('ol/renderer/canvas/VectorTileLayer', function () {
           undefined,
           'rgba(255, 0, 0, 0.5)',
           'rgba(0, 0, 255, 0.5)',
+          undefined,
         ];

         expect(resolution).to.be(map.getView().getResolution());
@@ -349,6 +350,8 @@ describe('ol/renderer/canvas/VectorTileLayer', function () {
         expect(layer.getRenderer().container.style.backgroundColor).to.be(
           'rgba(0, 0, 255, 0.5)'
         );
+        map.renderSync();
+        expect(layer.getRenderer().container.style.backgroundColor).to.be('');
         done();
       });
     });
```

The patch would look like this;
```patch
--- src/ol/renderer/canvas/Layer.js
+++ src/ol/renderer/canvas/Layer.js
@@ -195,12 +195,8 @@ class CanvasLayerRenderer extends LayerRenderer {
       this.container = container;
       this.context = context;
     }
-    if (
-      !this.containerReused &&
-      backgroundColor &&
-      !this.container.style.backgroundColor
-    ) {
-      this.container.style.backgroundColor = backgroundColor;
+    if (!this.containerReused && this.container && this.container.style) {
+      this.container.style.backgroundColor = backgroundColor || null;
     }
   }

```

ahocevar:
Thanks for your attempt to fix this, @kikuchan. The solution you chose changes the behavior in the case where multiple layers in the layer stack have a `background` configured. So I would suggest setting the style to `null` before the container is passed to the first layer renderer, and leave the rest of the code as it was before your changes.

kikuchan:
Thank you for your reply!
How about this?

src/ol/renderer/canvas/Layer.js
```diff
@@ -177,6 +177,8 @@ class CanvasLayerRenderer extends LayerRenderer {
       this.container = null;
       this.context = null;
       this.containerReused = false;
+    } else if (this.container && this.container.style) {
```
ahocevar, line 180: I think this can be simplified to
```suggestion
    } else if (this.container) {
```

ahocevar (commented):
Excellent, thanks for your continued effort. I added one minor suggestion, which you might want to consider.

kikuchan:
Thank you for the suggestion.
I had to change a test code because it used dummy container directly, and it failed with the change.

ahocevar (approved):
Excellent, thanks @kikuchan!

kikuchan:
Thank you!

Issue #13545, opened 2022-04-08
Stop PinchZoom when page scaled by pinch


ssugden:
On touch devices the OpenLayers PinchZoom interaction changes the zoom/resolution of the map. This works well unless the user gets into the following situation of 'no escape':

- User visits a page with an OpenLayers map, on say a mobile device
- The user uses the pinch gesture to scale the page (not the map) so that now only the map is now visible on the page
- The user tries to pinch again to scale the page back to normal, but as only the map is visible and any pinch gesture changes the zoom/resolution on the map and not the page, the user gets stuck in the map and cannot get back to the page 

The above can be seen in the https://openlayers.org/en/latest/examples/pinch-zoom.html example.

One option is to disable the map PinchZoom interaction whenever the user uses the pinch gesture to scale the page. You may have a better solution?

How do we detect the above situation and prevent it from happening?

Thank you.

mike-000:
That is opposite problem to #13495.  If the page must be able to zoom https://stackoverflow.com/a/995967/10118270 might help with the detection.

mike-000:
This works on Android with Chrome - disable the interaction if the map size exceeds more than half the visual viewport https://codesandbox.io/s/pinch-zoom-forked-r7lk4r?file=/main.js

ssugden:
Many thanks. This looks like a something that I can work with.

PR #14906, opened 2023-07-17
Use `touch-action: none;` for elements in viewport

### LINKED ISSUE #13495: `user-scalable=no` can be defeated in some browsers
**Describe the bug**
Preventing browser scaling is important in full screen mobile applications, but `user-scalable=no` as used in the examples can be defeated as it does apply to some elements.

**To Reproduce**
Open the Full-Screen Mobile example (I used Chrome on Android 12).  Place one finger on a zoom control button, the other on the center of the map, and spread apart.  It is possible to zoom so all the controls go out of the viewport.  Release and touch again and this time it is the map panning and there is no way to get the controls back.  That might be hard to reproduce accidently but if there are input elements such as Geocoder or Layer Switcher on the map it is easily done accidently 

**Expected behavior**
`touch-action: none;`  https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#disabling_all_gestures seems to be the current recommended method, e.g.:

```
#map * {
  touch-action: none;
}
```

### LINKED ISSUE #14893: ol.layer.Layer maxZoom property causes zooming bug on android devices
Setting property maxZoom on ol.layer.Layer causes zooming the whole page instead of the map itself on most of the android devices.

1. Go to https://openlayers.org/en/latest/examples/layer-zoom-limits.html.
2. Pinch zoom over threshold zoom level 14.
3. From this point, instead of the map the whole page is zoomed and the view is broken.

There is no problem on iPhones and some android devices. We observed that the bug can be reproduced on approximately 80% of the android devices at least on google chrome and samsung internet browser.

Problem: Samsung S21 Ultra 5G, Samsung Note 20 Ultra, Samsung Galaxy Note 9
No problem: Samsung s10e, Xiaomi Redmin Note 9 pro

I am using the library version 7.1.0, but the problem is also in the newest version. I think, that I saw this issue even years ago on one android phone. It really might be a very old problem, although I can't imagine it has been unobserved and unresolved till now.

mike-000:
Fixes #13495
Fixes #14893

Prevents touch actions on elements within the viewport scaling the page (but not actions on the viewport itself if interactions are disabled as in #13545).

### DIFF
--- a/src/ol/ol.css
+++ b/src/ol/ol.css
@@ -111,7 +111,7 @@
   all: unset;
 }
 
-.ol-viewport * {
+.ol-viewport {
   touch-action: none;
 }
 
--- a/src/ol/ol.css
+++ b/src/ol/ol.css
@@ -111,6 +111,10 @@
   all: unset;
 }
 
+.ol-viewport * {
+  touch-action: none;
+}
+
 .ol-selectable {
   -webkit-touch-callout: default;
   -webkit-user-select: text;


src/ol/ol.css
```diff
@@ -111,6 +111,10 @@
   all: unset;
 }
 
+.ol-viewport * {
```
ahocevar, line 114: Why not just
```suggestion
.ol-viewport {
```
?

mike-000: The setting is not inherited and in #13495 the main problem is large controls within the viewport.  I'm not entirely sure what is causing the problem in #14893 but I suspect transient empty elements as the layer visibility changes.

ahocevar: I think the `*` is too invasive for a library setting to make.

mike-000: `.ol-viewport` alone will fix #14893 and the full screen mobile example in #13495 as one finger is going to be on the viewport.  If an application has controls such as layer switcher or geocoder then either the control or application will need to suppress those from scaling the page.

ahocevar: I'd say this is the responsibility of the application. Because as a user, I might expect to be able to zoom the output of a layer switcher or geocoder.

In the first place
html```
```
in the document's head should prevent user zooming on mobile devices.

mike-000:
That is reasonable enough if a user pinches with one finger on a control or overlay, but in #14893 it can happen unexpectedly.

ahocevar:
> Pinching on the map should zoom the map. not the page.

It does - on a fullscreen map, and on any map when both fingers start pinching on the map.

On other pages, this may not be the desired behaviour: it depends on the layout of the page, and how central the role of the map is on that page. Users can always opt into avoiding page zoom when by setting `touch-action: none` on the map's target element.

ahocevar:
Ok, now that's finally a decent enough explanation of the original problem. This sounds like it's related to #13771. #14916 might be an alternative fix then.  Can anyone verify that with https://deploy-preview-14916--ol-site.netlify.app/en/latest/examples/layer-zoom-limits.html?

mike-000:
#14916 does not fix it.

ahocevar:
Thanks for the feedback, @mike-000. Still wondering if there isn't a more plausible fix for this problem. In some way, it must have to do with the target of tracked pointers changing when the ol-layers container has no children.

mike-000:
I tried the Lazy Source example, if you tap the buttons while zooming in the page briefly zooms, but the map quickly takes over again before it completely fills the viewport (unlike Layer Zoom Limits).

ahocevar:
I did some more investigation, and it seems the extra rendering effort when the layers change in the the `layer-zoom-limits` example blocks JavaScript execution to an extent where the browser handles the event instead of OpenLayers. With this in mind, I think the change that @mike-000 suggests here is the least invasive solution we have available at the moment.

ahocevar (approved):

PR #14907, opened 2023-07-17
Remove reference to ES5

mike-000:
The examples are no longer transpiled so remove reference from readme


MoonE:
In this case the Note and es2015 tag can be removed,  too.
https://github.com/openlayers/openlayers/blob/a798d81cfb6a914e78266228f82f948ec61ff5bf/examples/es2015-custom-element.html#L6-L7
Maybe also rename the example file? Though that would break  any existing links.

ahocevar:
> In this case the Note and es2015 tag can be removed, too.
> 
> https://github.com/openlayers/openlayers/blob/a798d81cfb6a914e78266228f82f948ec61ff5bf/examples/es2015-custom-element.html#L6-L7
> 
> 
> Maybe also rename the example file? Though that would break any existing links.

I think it's ok to keep this as-is, since web components are a feature that was added with ES2015.

jahow (approved):
Thanks for the catch @mike-000!

PR #14910, opened 2023-07-18
Add the drawing style settings

celia1020:


ahocevar:
We try to provide some really basic examples. To keep this examples easy to read, please create a new example, and use the [flat style](https://openlayers.org/en/latest/apidoc/module-ol_style_flat.html) syntax.

celia1020:
yes

Issue #14909, opened 2023-07-17
Cluster source does not draw if a point has invalid geometry


roemhildtg:
**Describe the bug**

https://github.com/openlayers/openlayers/blob/main/src/ol/source/Cluster.js#L118

This line above produces an error if a geometry is null. 

**To Reproduce**
Steps to reproduce the behavior: simply passing a feature with a null geometry produces the error. 

https://codesandbox.io/s/cluster-forked-5k3my5?file=/main.js

Here's also some sample geojson that reproduces the issue:

<details>

<summary>Sample geojson</summary>

```json
{
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "id": 2551,
            "geometry": null,
            "geometry_name": "geom",
            "properties": {
                "id": 2551,
                "type": "projects-files",
                "typeId": "0",
                "programId": "47"
            }
        },{
            "type": "Feature",
            "id": 2571,
            "geometry": {
                "type": "Point",
                "coordinates": [
                    -84.466466667,
                    33.883755556
                ]
            },
            "geometry_name": "geom",
            "properties": {
                "id": 2571,
                "type": "projects-files",
                "typeId": "0",
                "programId": "47"
            }
        }
    ],
    "totalFeatures": 2,
    "numberMatched": 2,
    "numberReturned": 2,
    "crs": {
        "type": "name",
        "properties": {
            "name": "urn:ogc:def:crs:EPSG::4326"
        }
    }
}
```
</details>

**Expected behavior**

I'm not sure, but I think that openlayers should work with features that have a null geometry, since its possible and common. 

I'm happy to open a MR if this is considered a bug and someone can recommend the appropriate fix - here's what I think would work:

```js


    /**
     * @param {Feature} feature Feature.
     * @return {Point} Cluster calculation point.
     * @protected
     */
    this.geometryFunction =
      options.geometryFunction ||
      function (feature) {
        const geometry = /** @type {Point} */ (feature.getGeometry());
        if(!geometry){ return new Point([]); } // <-- added this check here
        // assert(geometry.getType() == 'Point', 10); // The def ault `geometryFunction` can only handle `Point` geometries
        return geometry;
      };

```

mike-000:
`assert(geometry?.getType() === 'Point', 10);` would be sufficient to include null or undefined geometry in the assertion.  The documentation already specifies that a custom geometry function is required if the geometry is not Point.

MoonE:
[...]

PR #14915, opened 2023-07-18
Fix cluster source handling of points with null geometry

roemhildtg:
closes https://github.com/openlayers/openlayers/issues/14909


MoonE (approved):
Thanks, @roemhildtg
I added a test to prevent it from regressing again.

roemhildtg:
Awesome, thank you for the extra text!

PR #14916, opened 2023-07-18
Delete all pointers with outdated target

### LINKED ISSUE #14893: ol.layer.Layer maxZoom property causes zooming bug on android devices

ahocevar:
Maybe fixes #14893

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -171,14 +171,11 @@ class MapBrowserEventHandler extends Target {
       event.type == MapBrowserEventType.POINTERCANCEL
     ) {
       delete this.trackedTouches_[id];
+      // Some platforms assign a new pointerId when the target changes.
+      // If this happens, delete pointers with the old target.
       for (const pointerId in this.trackedTouches_) {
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for the feedback, @mike-000. Still wondering if there isn't a more plausible fix for this problem. In some way, it must have to do with the target of tracked pointers changing when the `ol-layers` container has no children.

PR #14918, opened 2023-07-18
Avoid line dash flicker in Measure examples

mike-000:
When a polygon is drawn clockwise the outline stroke line dash matches the offset of the sketch line stroke line dash, but if drawn counter clockwise the offsets to not match and cause a flickering effect when drawing.  This can be avoided as when drawing a polygon for measurement the sketch line does not need to be styled.


MoonE (approved):
Thanks, @mike-000

PR #14919, opened 2023-07-19
Call requestAnimationFrame in the appropriate place

ahocevar:
This is a follow-up on #14856, and fixes one remaining issue with the raster source's workflow of handling the response from the worker.


ahocevar:
Closing, not the right fix.

PR #14912, opened 2023-07-18
add an example of drawing features style setting

celia1020:

PR #14920, opened 2023-07-19
Draw features style example

ahocevar:
See #14912 for the history of this. I accidentally closed #14912.


ahocevar:
Thanks, @celia1020

PR #14922, opened 2023-07-19
Fix infinite tile rendering loop and incomplete images in Raster source

ahocevar:
This pull request fixes an infinite rendering loop that was introduced with #14856, by reverting the changes from that pull request. I did an in-depth investigation of the problem with incomplete images when tiled sources are used for a Raster source, and I found a way to fix it: by calling `processSources_()` instead of `changed()` in the tile queue's loaded callback.

### DIFF
--- a/src/ol/source/Raster.js
+++ b/src/ol/source/Raster.js
@@ -601,7 +601,7 @@ class RasterSource extends ImageSource {
      */
     this.tileQueue_ = new TileQueue(function () {
       return 1;
-    }, this.changed.bind(this));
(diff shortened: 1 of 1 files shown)


ahocevar:
Note: other than I previously thought, the map at the top of the website preview does not use the OpenLayers version from the pull request, but the one referenced in the ol-mapbox-style dependency. The best example to test this is to do a lot of zooming and panning in the https://deploy-preview-14922--ol-site.netlify.app/mapbox-style.html example.

PR #14925, opened 2023-07-20
Add typedef, use underscore for private methods

mike-000:
It seems wrong that a property does not have a typedef and that private methods do not use underscore.  I am not familiar with Angular but I suspected the former is causing the issue in https://stackoverflow.com/q/76728961/10118270  As in `ol/interaction/MouseWheelZoom` a timeout id is declared as `{?}`.

### DIFF
--- a/src/ol/interaction/DblClickDragZoom.js
+++ b/src/ol/interaction/DblClickDragZoom.js
@@ -59,6 +59,12 @@ class DblClickDragZoom extends Interaction {
     this.handlingDoubleDownSequence_ = false;
 
     /**
+     * @type {?}
+     * @private
+     */
+    this.doubleTapTimeoutId_ = undefined;
+
+    /**
      * @type {!Object<string, PointerEvent>}
      * @private
      */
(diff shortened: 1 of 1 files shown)


src/ol/interaction/DblClickDragZoom.js
```diff
@@ -58,6 +58,12 @@ class DblClickDragZoom extends Interaction {
      */
     this.handlingDoubleDownSequence_ = false;
 
+    /**
+     * @type {?}
```
ahocevar, line 62:
```suggestion
     * @type {ReturnType<typeof setTimeout>}
```

ahocevar (commented):
Thanks, @mike-000. I added a suggestion for a better type.

mike-000:
> Thanks, @mike-000. I added a suggestion for a better type.

That could also be updated in `ol/MapBrowserEventHandler.js` and `ol/interaction/MouseWheelZoom.js`

ahocevar (approved):
Thanks, @mike-000

ahocevar:
> That could also be updated in `ol/MapBrowserEventHandler.js` and `ol/interaction/MouseWheelZoom.js`

True, that would be a a nice fix in a separate pull request.

PR #14927, opened 2023-07-20
STAC support (a STAC layer group)

m-mohr:
This PR proposes to add STAC capabilities to OpenLayers. It adds a STAC LayerGroup.

The layer group creates various layers depending on the given options and what the STAC entity offers, e.g. footprints, COGs and thumbnails. What usually is the source is the stac-js library here.

Examples: https://deploy-preview-14927--ol-site.netlify.app/en/latest/examples/?q=stac

The work is based on PR #14101 from @tschaub (he's aware of this work).
The interface is mostly aligned with the stac-layer library for Leaflet (<https://github.com/stac-utils/stac-layer>), which is also based on stac-js and offers pretty much the same functionality.

An important question is whether this is something for OL core or whether I should publish this as an extension.

The functionality implemented here includes:
- [x] Bounding box / geometry rendering with custom styles
- [x] COG rendering (client-side or server-side)
    - it reads various information from the STAC metadata to find a good default rendering
    - if the client-side rendering errors (e.g. due to CORS), it can fallback to a XYZ tile-server
- [x] Rendering of thumbnails if no COG is available
- [x] Several options to customize the behavior, e.g. specific bands, which assets to display, tile-server,  styles, ...
- [x] Supports STAC Items, Collections, API Collections, and API Items
- [x] Supports older STAC versions (via stac-migrate)
- [x] Supports several events: assetsready, sourceready, error
- [x] Multiple examples showing various ways of usage
- [x] The async method `getSourceOptions` (implemented by @tschaub) to provide a function that returns options for the source given item metadata. The Planetary Computer example uses this option to [sign](https://planetarycomputer.microsoft.com/docs/concepts/sas/) the asset URLs before the source is created.
- [x] ~Adds a `projection` option to the `GeoTiff` source to allow overriding the auto-detected projection with what STAC provides~ Moved to #15036 

Todos:
- [x] Projection handling for GeoJSON (see https://github.com/openlayers/openlayers/pull/14927#discussion_r1282024469 )
- [x] Support passing HTTP headers with the requests (via the GeoTiffSourceOptions)
- [x] (Basic) support for STAC web-map-links extension
- [x] Easy retrieval of STAC entities for events (see https://github.com/openlayers/openlayers/pull/14927#discussion_r1282171332)
- [ ] Tests (I'll start this once I know whether this will live in the OL repo or not)

Curious to get thoughts from others and maybe point me to things that can be improved. 

PS: The work on this was funded by Planet and EOX.

### DIFF
--- a/src/ol/events/ErrorEvent.js
+++ b/src/ol/events/ErrorEvent.js
@@ -1,0 +1,25 @@
+/**
+ * @module ol/events/ErrorEvent
+ */
+import BaseEvent from './Event.js';
+import EventType from './EventType.js';
+
+/**
+ * @classdesc
+ * Event emitted on configuration or loading error.
+ */
+class ErrorEvent extends BaseEvent {
+  /**
+   * @param {Error} error error object.
+   */
+  constructor(error) {
+    super(EventType.ERROR);
+
+    /**
+     * @type {Error}
+     */
+    this.error = error;
+  }
+}
+
+export default ErrorEvent;
--- a/src/ol/layer/STAC.js
+++ b/src/ol/layer/STAC.js
@@ -1,0 +1,780 @@
+/**
+ * @module ol/layer/STAC
+ */
+import ErrorEvent from '../events/ErrorEvent.js';
+import GeoJSON from '../format/GeoJSON.js';
+import GeoTIFF from '../source/GeoTIFF.js';
+import ImageLayer from './Image.js';
+import Layer from './Layer.js';
+import LayerGroup from './Group.js';
+import StaticImage from '../source/ImageStatic.js';
+import TileJSON from '../source/TileJSON.js';
+import TileLayer from './Tile.js';
+import VectorLayer from './Vector.js';
+import VectorSource from '../source/Vector.js';
+import WMS from '../source/TileWMS.js';
+import WMTS, {optionsFromCapabilities} from '../source/WMTS.js';
+import WebGLTileLayer from './WebGLTile.js';
+import XYZ from '../source/XYZ.js';
+import create, {Asset, STAC} from 'stac-js';
+import {
+  defaultBoundsStyle,
+  defaultCollectionStyle,
+  getBoundsStyle,
+  getGeoTiffSourceInfoFromAsset,
+  getProjection,
+  getSpecificWebMapUrl,
+  getWmtsCapabilities,
+} from './stacUtil.js';
+import {toGeoJSON} from 'stac-js/src/geo.js';
+import {transformExtent} from '../proj.js';
+
+/**
+ * @typedef {import("../extent.js").Extent} Extent
+ */
+/**
+ * @typedef {import("../source/GeoTIFF.js").Options} GeoTIFFSourceOptions
+ */
+/**
+ * @typedef {import("../source/ImageStatic.js").Options} ImageStaticSourceOptions
+ */
+/**
+ * @typedef {import("stac-js").Link} Link
+ */
+/**
+ * @typedef {import("stac-js").STACObject} STACObject
+ */
+/**
+ * @typedef {import('../style.js').Style} Style
+ */
+/**
+ * @typedef {import("../source/XYZ.js").Options} XYZSourceOptions
+ */
+
+/**
+ * @typedef {Object} Options
+ * @property {string} [url] The STAC URL. Any of `url` and `data` must be provided.
+ * Can also be used as url for data, if it is absolute and doesn't contain a self link.
+ * @property {STAC|Asset|Object} [data] The STAC metadata. Any of `url` and `data` must be provided.
+ * `data` take precedence over `url`.
+ * @property {Array<string|Asset>|null} [assets=null] The selector for the assets to be rendered,
+ * only for STAC Items and Collections.
+ * This can be an array of strings corresponding to asset keys or Asset objects.
+ * null shows the default asset, an empty array shows no asset.
+ * @property {Array<number>} [bands] The (one-based) bands to show.
+ * @property {function(GeoTIFFSourceOptions, Asset):(GeoTIFFSourceOptions|Promise<GeoTIFFSourceOptions>)} [getGeoTIFFSourceOptions]
+ * Optional function that can be used to configure the underlying GeoTIFF sources. The function can do any additional work
+ * and return the completed options or a promise for the same. The function will be called with the current source options
+ * and the STAC Asset.
+ * @property {function(ImageStaticSourceOptions, (Asset|Link)):(ImageStaticSourceOptions|Promise<ImageStaticSourceOptions>)} [getImageStaticSourceOptions]
+ * Optional function that can be used to configure the underlying ImageStatic sources. The function can do any additional work
+ * and return the completed options or a promise for the same. The function will be called with the current source options
+ * and the STAC Asset or Link.
+ * @property {function(XYZSourceOptions, (Asset|Link)):(XYZSourceOptions|Promise<XYZSourceOptions>)} [getXYZSourceOptions]
+ * Optional function that can be used to configure the underlying XYZ sources that displays imagery. The function can do any
+ * additional work and return the completed options or a promise for the same. The function will be called with the current
+ * source options and the STAC Asset or Link.
+ * @property {boolean} [displayGeoTiffByDefault=false] Allow to choose non-cloud-optimized GeoTiffs as default image to show,
+ * which might not work well for larger files or larger amounts of files.
+ * @property {boolean} [displayPreview=false] Allow to display images that a browser can display (e.g. PNG, JPEG),
+ * usually assets with role `thumbnail` or the link with relation type `preview`.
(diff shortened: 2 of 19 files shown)


jahow:
Impressive work @m-mohr, thanks!! since you're asking for feedback here are a few things that came to my mind when reading your PR. Note that some aspects of the PR may have already been discussed and agreed on so this is just my opinion!

* The fact that the STAC layer can create other layers (e.g. in case of a collection of items) is uncommon for layers in the library; this might make things awkward for users e.g. for specifying layer options (should layer options be applied to all created layer?) or getting back these layers without searching through the layer tree
* More generally, it feels like the STAC layer is doing too much magic: using a XYZ or Geotiff source depending of where tiles are computed, using a vector source for the footprint and a static source for the preview... I think the API would be more explicit and straightforward with utilities like `readStacItem` and `readStacCollection` (naming probably dumb) that would then allow the user to explicitly create the type of layer and source they want according to the assets found; it might be significantly more verbose though
* I would externalize STAC-related utilities to make the STAC layer file lighter (see for example https://github.com/openlayers/openlayers/blob/main/src/ol/source/ogcTileUtil.js)
* Some examples have logic for reading layer titles as well, but this doesn't show up visually (it looks like it's expecting a layer tree ?)

Again, this is just my opinion and not a definitive review. And it's great to see first-class STAC support coming in OpenLayers!

m-mohr:
Thank you for your feedback @jahow, I appeciate it!

>     * The fact that the STAC layer can create other layers (e.g. in case of a collection of items) is uncommon for layers in the library; this might make things awkward for users e.g. for specifying layer options (should layer options be applied to all created layer?) or getting back these layers without searching through the layer tree

This is something we could improve, of course, e.g. by having options like `geotiffOptions`, `xyzOptions` that are passed through. We could also specify a property for each layer for identification purposes.

>     * More generally, it feels like the STAC layer is doing too much magic: using a XYZ or Geotiff source depending of where tiles are computed, using a vector source for the footprint and a static source for the preview... I think the API would be more explicit and straightforward with utilities like `readStacItem` and `readStacCollection` (naming probably dumb) that would then allow the user to explicitly create the type of layer and source they want according to the assets found; it might be significantly more verbose though

That's why it is a LayerGroup, of course. The LayerGroup is meant to make life very easy and give a consistent rendering behavior for STAC. But I can understand that this might be too much magic. In that case another option could also be to just include the change to the GeoTiff (and maybe the STAC source) in this PR, and then keep the magic/LayerGroup as extension outside of this repo. Also fine with me. Just let me know what the maintainers think about it.

>     * I would externalize STAC-related utilities to make the STAC layer file lighter (see for example https://github.com/openlayers/openlayers/blob/main/src/ol/source/ogcTileUtil.js)

Sure, I can do that. For example, the ErrorEvent is also used in MapboxVector.js, so that's the first thing to centralize. Parts of the GeoTIFF logic can probably be externalized. The default styles can also be external. Other than that, I'd need to dig a bit deeper to see what makes sense. If you have something obvious in mind, please let me know.

>     * Some examples have logic for reading layer titles as well, but this doesn't show up visually (it looks like it's expecting a layer tree ?)

Yeah, I think it was for debugging purposes at some point. I could add a layer switcher there, which could help visualize the layer tree structure. Not sure whether that's something you would want in an example though?

src/ol/layer/STAC.js
```diff
@@ -0,0 +1,625 @@
...
+    if (data.isItemCollection() || data.isCollectionCollection()) {
+      geojson = toGeoJSON(data.getBoundingBox());
+    } else {
+      geojson = data.toGeoJSON();
+    }
+
+    if (geojson) {
+      const source = new VectorSource({
+        features: new GeoJSON().readFeatures(geojson, {
+          featureProjection: 'EPSG:3857', // TODO: revisit this
```
m-mohr, line 515: This is something I'd like to hear thoughts on. I need to load the GeoJSON and provide a target projection, but I likely don't know the map projection at this this. So this is hardcoded to the default, which is not ideal.

src/ol/layer/STAC.js
```diff
@@ -0,0 +1,625 @@
...
+      const features = new GeoJSON().readFeatures(geojson, {
+        featureProjection: 'EPSG:3857', // TODO: revisit this
+      });
+      const source = new VectorSource({features});
+      const vectorLayer = new VectorLayer({
+        source,
+        style: getBoundsStyle(this.boundsStyle_, this),
+      });
+      vectorLayer.set('bounds', true);
+      features.forEach((feature) => feature.set('stac', data));
```
ahocevar, line 522: A few thoughts here:
* If you configure the source with a `loader` instead of `features`, you can use the view projection, which is passed to the loader as 3rd argument. Then you do not need to hard-code the projection.
* Instead of spreading metadata across assets that the STAC layer group creates, I'd suggest making them available through the STAC instance (e.g. `getBoundsLayer()` instead of setting a `bounds` boolean on the layer. Same for the `stac` data, which should not be set on each feature.

src/ol/layer/stacUtil.js
```diff
@@ -0,0 +1,121 @@
...
+  }),
+});
+
+/**
+ * Get the STAC objects associated with this event, if any. Excludes API Collections.
+ * @param {import('../MapBrowserEvent.js').default} event The asset to read the information from.
+ * @return {Array<STAC>} A list of STAC objects
+ */
+export function getStacObjectsForEvent(event) {
+  const features = event.map
```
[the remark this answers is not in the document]
ahocevar: Please add a test case in `test/ol/browser` to provide some context and something we can use to look at it and, if need be, fix it.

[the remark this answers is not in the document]
m-mohr: The STAC ItemCollection in this PR has an example. But I can also look into how writing tests tomorrow.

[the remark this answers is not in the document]
m-mohr: I couldn't make my code work but also couldn't replicate the issue in tests. So I went for a different approach which seems to work just fine.

src/ol/layer/STAC.js
m-mohr: Thanks, good pointers!

1. Switched to use a loader, which seems to work perfectly.
2. I removed the stac metadata for the features, this is not needed anymore (due to the change for https://github.com/openlayers/openlayers/pull/14927#discussion_r1282171332). I added a `getBoundsLayer()` method, but the metadata on the assets is still valuable due to the nesting of the LayerGroups, I think. In case you go through all layers, it is helpful to be able to retrieve the relevant STAC entities. Using that for example for customite a LayerSwitcher or for the "click detection" in stacUtils / the STAC Item Collection example.

src/ol/layer/STAC.js
m-mohr: Resolved via https://github.com/openlayers/openlayers/pull/14927#discussion_r1282188137

src/ol/source/GeoTIFF.js
```diff
@@ -382,7 +384,7 @@ class GeoTIFFSource extends DataTile {
     super({
       state: 'loading',
       tileGrid: null,
-      projection: null,
+      projection: options.projection || null,
```
m-mohr, line 387: I've extracted the changes in this file to #15036 so that we could also release this PR separately, if required.

src/ol/events/ErrorEvent.js
```diff
@@ -0,0 +1,25 @@
+/**
```
m-mohr, line 1: This was a shared event between MapboxVector and STAC, but now that MapboxVector was removed I could also keep it outside of the events folder and copy it into stacUtils.js or STAC.js. Whatever you prefer.

m-mohr:
Hey there, 
to be able to get this into our projects via npm etc, I (temporarily?) released this in a separate repo: https://m-mohr.github.io/ol-stac/
I used the OL repo as a template with all the nice boilerplate for website, examples, docs etc. I'm wondering if this is fine with you/the OL maintainers? Shall I change the design more to make it look too less like an "official" openlayers thing?

ahocevar:
Thanks @m-mohr for your effort on this, and having this as a separate library makes perfect sense! No objections from me on how the website looks.

Would you mind creating a pull request that adds your library to the list of 3rd party libraries in `site/src/3rd-party/index.hbs`?

Also, to make this more prominent in OpenLayers, I think it would be nice to have an example here that uses your library.

m-mohr:
Thanks, @ahocevar. If you still want STAC be part of OL, I'm also happy to also have it live in OpenLayers directly. Up to the maintainers :-) But I wanted to get this out so people can start using it and give feedback.

> Would you mind creating a pull request that adds your library to the list of 3rd party libraries in site/src/3rd-party/index.hbs?

Sure, see #15041

> Also, to make this more prominent in OpenLayers, I think it would be nice to have an example here that uses your library.

I tried this, but couldn't figure out how it works with the dependencies. How can I reference external dependencies? If I just add the ol-stac import, it can't resolve it. Any hints? devDependencies? See #15042

m-mohr:
Whoops, forgot to close. The OpenLayers plugin can now be found here: https://m-mohr.github.io/ol-stac/

PR #14929, opened 2023-07-21
WebGL helper improvements

ahocevar:
I think because getting the appropriate WebGL context from a canvas implies a little overhead, it is better to store the context instead of the canvas, like we also do for 2d contexts.

I also added missing `dispose()` calls on the helper in the WebGL tests.

### DIFF
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -114,7 +114,7 @@ export const AttributeType = {
 
 /**
  * @typedef {Object} CanvasCacheItem
- * @property {HTMLCanvasElement} canvas Canvas element.
+ * @property {WebGLRenderingContext} context The context of this canvas.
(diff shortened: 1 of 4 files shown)


src/ol/webgl/Helper.js
```diff
@@ -174,12 +177,14 @@ function releaseCanvas(key) {
...
-  const canvas = cacheItem.canvas;
-  const gl = getContext(canvas);
+  const gl = cacheItem.context;
   const extension = gl.getExtension('WEBGL_lose_context');
   if (extension) {
     extension.loseContext();
   }
+  const canvas = gl.canvas;
+  canvas.width = 1;
+  canvas.height = 1;
```
ahocevar, line 187: This helps to avoid exceeding the available canvas memory in Safari.

jahow (approved):
Thanks!

PR #14934, opened 2023-07-21
Better types for more timeouts

mike-000:
As suggested in https://github.com/openlayers/openlayers/pull/14925#issuecomment-1644530238


ahocevar (approved):
Thanks, @mike-000

ahocevar (approved):

PR #14937, opened 2023-07-22
Update TileJSON source url

mike-000:
`https://a.tiles.mapbox.com/v3/aj.1x1-degrees.json` used in several examples is no longer working and the alternative v4 version needs an API key which is inconvenient in CodeSandbox when it is only needed to provide a background context to what the example is demonstrating.  An alternative to is directly reference the TileJSON endpoint of an OGC Web Mercator raster source which does not require an API key.


ahocevar (changes requested):
It looks weird to enforce the TileJSON format for a data source that supports OGC API Tiles, which we have a dedicated source for. I think it would be nice to use that instead, except for the `tilejson` example.

ahocevar (approved):
Thanks, @mike-000! It looks like the examples that use elevation tiles from Mapbox's v3 API are also broken. If you have a replacement source at hand, a fix for those would be appreciated too.

PR #14938, opened 2023-07-23
Apply dashed line hit detection to all geometry types

### LINKED ISSUE #14936: Extend #11805 / #11874 hit detection to polygons and circles
Although a transparent fill can be used to hit detect a polygon that is not useful if polygons overlap meaning they must be selected by their outlines.  If the outlines are dashed that can be difficult as in https://codesandbox.io/s/layer-clipping-vector-forked-drzyx8  It should be consistent with LineString hit detection as in #11805 / #11874.

mike-000:
Fixes #14936

Previously only LineString geometry was hit detected between dashes. A MultiLineString with similar coordinates would not (although using the previous test for LineString would have not have noticed that as it was hitting the rounded end of a dash).

This applies the same logic to LineString and MultiLineString, plus Polygon, MutiPolygon and Circle geometry outlines, and sets the test resolution and geometry coordinates to match the dash intervals to ensure a meaningful test with hits midway between dashes.

Vector layer hit detection already had this behaviour, so map and vector layer hit detection will now be consistent.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -6,6 +6,10 @@
 
 The `start` and `end` behavior previously was equivalent to `right` and `left`. Now it takes the text direction into account, so it will mean `left` and `right` for left-to-right text.
 
+#### Consistent hit detection of dashed lines
+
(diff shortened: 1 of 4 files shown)


ahocevar (changes requested):
Please add a note about this change to the upgrade notes

ahocevar (approved):
Thanks, @mike-000

Issue #14924, opened 2023-07-20
When the opacity of the feature is set to 0, its click event will not be triggered


Lazydd:
This is an example：https://codesandbox.io/s/overlay-forked-hhtfm6?file=/main.js, And when the map is moving, the feature will no longer flash. How can I solve this problem.

tanhh326:
https://openlayers.org/en/latest/examples/?q=animation

Lazydd:
> https://openlayers.org/en/latest/examples/?q=animation

不太行，写了半天没写出来

mike-000:
To update the style while panning use `updateWhileInteracting: true` in the layer options.
To hit detect an invisible feature you will need to add a Circle or RegularShape with transparent fill to the style.
https://codesandbox.io/s/overlay-forked-twymhj?file=/main.js

Lazydd:
> To update the style while panning use `updateWhileInteracting: true` in the layer options. To hit detect an invisible feature you will need to add a Circle or RegularShape with transparent fill to the style. https://codesandbox.io/s/overlay-forked-twymhj?file=/main.js

Thanks

mike-000:
> can you add an option to control the hidden things would also work

@ahocevar maybe that should be the default?  It seems inconsistent that a zero opacity icon can cause a visible icon to be decluttered, but it is not hit detected.  Does opacity really need to be included in hit detection instructions?  Does anyone use opacity to avoid hit detection?  How it is different to transparency or the intervals between line dash which are hit detected?  If any change is made text and image behaviour should be consistent.

ahocevar:
@mike-000 Yes, to be consistent with what was decided and implemented for polygons in #7750, an `Icon` style with `opacity: 0` should be hit detected.

For `Text` styles, the situation is a bit different because `Text` does not have an opacity. Instead, `Text` has `Fill` and `Stroke` that can have a color with an alpha of zero. I honestly don't know if that is already handled the same way as wehen rendering polygons or not.

mike-000:
#9986 and #10007 extended #7750 to Circle geometry, Circle style and RegularShape but not to text fill.  Transparent fill in text is current not hit detected, and because text fill defaults if undefined it is necessary to specify `fill: null` to use no fill (which is not documented in the API).

PR #14939, opened 2023-07-23
Hit detect images even if zero opacity

mike-000:
Fixes the image opacity as discussed in https://github.com/openlayers/openlayers/issues/14924#issuecomment-1646545130 plus ensures feature styles are not ignored in vector layer hit detection when the layer style is null.

(text opacity would need a separate PR).


ahocevar (changes requested):
Please add a note about this change to the upgrade notes.

test/browser/spec/ol/layer/vector.test.js
```diff
@@ -193,6 +193,49 @@ describe('ol.layer.Vector', function () {
       });
     });
 
+    it('detects zero opacity images', function (done) {
```
ahocevar, line 196: Split this into two tests: one for zero opacity images, one for feature styles with a layer style of null.

test/browser/spec/ol/layer/vector.test.js
```diff
@@ -193,6 +193,49 @@ describe('ol.layer.Vector', function () {
...
+
+      const style = new Style({
+        image: new Circle({
+          radius: 4,
+          fill: new Fill({
+            color: 'black',
+          }),
+        }),
+      });
+      style.getImage().setOpacity(0);
```
ahocevar, line 218: Use an `Icon` style instead of the `Circle`. For circle and regular shape, the recommended way of adding opacity is to use the nested `Fill` and `Stroke` style.

ahocevar (approved):
Thanks, @mike-000

PR #14943, opened 2023-07-24
Hit detect transparent fill and backgroundFill in Text styles

mike-000:
Makes Text fill and backgroundFill hit detection consistent with RegularShape fill as discussed in https://github.com/openlayers/openlayers/issues/14924#issuecomment-1646569165

### DIFF
--- a/src/ol/render/canvas/TextBuilder.js
+++ b/src/ol/render/canvas/TextBuilder.js
@@ -337,7 +337,6 @@ class CanvasTextBuilder extends CanvasBuilder {
         );
         if (textState.backgroundFill) {
           this.updateFillStyle(this.state, this.createFill);
-          this.hitDetectionInstructions.push(this.createFill(this.state));
         }
         if (textState.backgroundStroke) {
           this.updateStrokeStyle(this.state, this.applyStroke);
@@ -405,6 +404,12 @@ class CanvasTextBuilder extends CanvasBuilder {
         geometryWidths,
       ]);
       const scale = 1 / pixelRatio;
+      // Set default fill for hit detection background
+      const currentFillStyle = this.state.fillStyle;
+      if (textState.backgroundFill) {
+        this.state.fillStyle = defaultFillStyle;
+        this.hitDetectionInstructions.push(this.createFill(this.state));
+      }
       this.hitDetectionInstructions.push([
         CanvasInstruction.DRAW_IMAGE,
         begin,
@@ -433,6 +438,11 @@ class CanvasTextBuilder extends CanvasBuilder {
         this.textOffsetY_,
         geometryWidths,
       ]);
+      // Reset previous fill
+      if (textState.backgroundFill) {
+        this.state.fillStyle = currentFillStyle;
+        this.hitDetectionInstructions.push(this.createFill(this.state));
+      }
 
       this.endGeometry(feature);
     }
--- a/src/ol/render/canvas/TextBuilder.js
+++ b/src/ol/render/canvas/TextBuilder.js
@@ -95,6 +95,7 @@ class CanvasTextBuilder extends CanvasBuilder {
      * @type {!Object<string, import("../canvas.js").FillState>}
      */
     this.fillStates = {};
+    this.fillStates[defaultFillStyle] = {fillStyle: defaultFillStyle};
 
     /**
      * @private
@@ -427,7 +428,7 @@ class CanvasTextBuilder extends CanvasBuilder {
         this.text_,
         this.textKey_,
         this.strokeKey_,
-        this.fillKey_,
+        this.fillKey_ ? defaultFillStyle : this.fillKey_,
         this.textOffsetX_,
         this.textOffsetY_,
         geometryWidths,
@@ -524,7 +525,7 @@ class CanvasTextBuilder extends CanvasBuilder {
       end,
       baseline,
       textState.overflow,
-      fillKey,
+      fillKey ? defaultFillStyle : fillKey,
       textState.maxAngle,
       1,
       offsetY,
--- a/test/browser/spec/ol/renderer/map.test.js
+++ b/test/browser/spec/ol/renderer/map.test.js
@@ -205,6 +205,174 @@ describe('ol.renderer.Map', function () {
       expect(hit.geometry).to.be(geometry);
     });
 
+    it('hits Text stroke, transparent fill and background fill', function () {
+      let hit;
+      const geometry = new Point([0, 0]);
+      const feature = new Feature(geometry);
+      const layer = new VectorLayer({
+        source: new VectorSource({
+          features: [feature],
+        }),
+      });
+      map.addLayer(layer);
+
+      layer.setStyle({
+        'text-value': 'X',
+        'text-font': 'bold 100px sans-serif',
+        'text-baseline': 'top',
+        'text-offset-y': -50,
+        'text-stroke-width': 20,
+        'text-stroke-color': 'black',
+        'text-fill-color': null,
+      });
+      map.renderSync();
+      hit = map.forEachFeatureAtPixel([50, 50], (feature, layer, geometry) => ({
+        feature,
+        layer,
+        geometry,
+      }));
+      expect(hit).to.be.ok();
+      expect(hit.feature).to.be(feature);
+      expect(hit.layer).to.be(layer);
+      expect(hit.geometry).to.be(geometry);
+
+      layer.setStyle({
+        'text-value': 'X',
+        'text-font': 'bold 100px sans-serif',
+        'text-baseline': 'top',
+        'text-offset-y': -50,
+        'text-stroke-width': 1,
+        'text-stroke-color': 'black',
+        'text-fill-color': null,
+      });
+      map.renderSync();
+      hit = map.forEachFeatureAtPixel([50, 50], (feature, layer, geometry) => ({
+        feature,
+        layer,
+        geometry,
+      }));
+      expect(hit).to.be(undefined);
+
+      layer.setStyle({
+        'text-value': 'X',
+        'text-font': 'bold 100px sans-serif',
+        'text-baseline': 'top',
+        'text-offset-y': -50,
+        'text-stroke-width': 1,
+        'text-stroke-color': 'black',
+        'text-fill-color': 'transparent',
+      });
+      map.renderSync();
+      hit = map.forEachFeatureAtPixel([50, 50], (feature, layer, geometry) => ({
+        feature,
+        layer,
+        geometry,
+      }));
+      expect(hit).to.be.ok();
+      expect(hit.feature).to.be(feature);
+      expect(hit.layer).to.be(layer);
+      expect(hit.geometry).to.be(geometry);
+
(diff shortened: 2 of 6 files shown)


src/ol/render/canvas/TextBuilder.js
```diff
@@ -429,7 +436,7 @@ class CanvasTextBuilder extends CanvasBuilder {
         this.text_,
         this.textKey_,
         this.strokeKey_,
-        this.fillKey_,
+        this.fillKey_ ? defaultFillKey : this.fillKey_,
```
ahocevar, line 439: This should be either the `defaultFillKey` or `null`, right? The above can also result in `undefined`. I don't think this is desired.

src/ol/render/canvas/TextBuilder.js
```diff
@@ -526,7 +533,7 @@ class CanvasTextBuilder extends CanvasBuilder {
       end,
       baseline,
       textState.overflow,
-      fillKey,
+      fillKey ? defaultFillKey : fillKey,
```
ahocevar, line 536: Same as above.

ahocevar (changes requested):

src/ol/render/canvas/TextBuilder.js
mike-000: falsy should always be an empty string

ahocevar: Can you point to a test or a section of code that backs this?

mike-000: There are only two places in the module where `this.fillKey_` is set.  It is always a string.  `fillKey` is copied from `this.fillKey_`.  In any case the executor only checks for truthy/falsy.

src/ol/render/canvas/TextBuilder.js
```diff
@@ -41,6 +41,11 @@ export const TEXT_ALIGN = {
   'bottom': 1,
 };
 
+const defaultFillKey =
+  typeof defaultFillStyle == 'string'
+    ? defaultFillStyle
+    : '|' + getUid(defaultFillStyle);
+
```
ahocevar, line 48: You know the type of the `defaultFillStyle`, and there is no way for the user to change it.

mike-000: Defensive coding.  Where `defaultFillStyle` is set anywhere else a key is obtained by the normal route, so any future change to the default would not break any code.

ahocevar: I'd prefer a code base where we have tests rather than defensive coding.

src/ol/render/canvas/TextBuilder.js
```diff
@@ -336,6 +339,7 @@ class CanvasTextBuilder extends CanvasBuilder {
         );
         if (textState.backgroundFill) {
           this.updateFillStyle(this.state, this.createFill);
+          this.state.fillStyle = defaultFillStyle;
```
ahocevar, line 342: This looks scary, and unrelated to the pull request. Can you explain the rationale behind this added line?

src/ol/render/canvas/TextBuilder.js
```diff
@@ -39,6 +39,8 @@ export const TEXT_ALIGN = {
   'bottom': 1,
 };
 
+const defaultFillKey = /** @type {string} */ (defaultFillStyle);
```
ahocevar, line 42: Instead of the type cast here, you could change the type of `defaultFillStyle` to `string`. This would serve the purpose of your initial defensive coding intent, without the need to run extra code at run time.

src/ol/render/canvas/TextBuilder.js
mike-000: `backgroundFill` does not have a key for either rendering or hit detection, it uses the fillStyle directly, so that is needed to override transparent so transparent can be hit.

ahocevar: This brings state and instructions out of sync. Did you mean to use `state.currentFillStyle`?

ahocevar:
`DRAW_CHARS` instructions will also need to be updated if the intention here is to make things consistent.

src/ol/render/canvas/TextBuilder.js
ahocevar: Anyway, this does not look right. Text background is unfortunately not so well tested (my bad), and I'd like to avoid side effects. Better not implement hit-detection for transparent background fill, than risking a regression.

ahocevar: To do this properly, 3 hit detection instructions would be needed:
* SET_FILL_STYLE with a non-transparent fill
* DRAW_IMAGE
* SET_FILL_STYLE restoring the previous fill style

mike-000: Is the restore best done before or after the END_GEOMETRY ?

ahocevar: Do it right after DRAW_IMAGE.

src/ol/render/canvas/TextBuilder.js
```diff
@@ -336,7 +337,10 @@ class CanvasTextBuilder extends CanvasBuilder {
         );
         if (textState.backgroundFill) {
           this.updateFillStyle(this.state, this.createFill);
+          const fillStyle = this.state.fillStyle;
+          this.state.fillStyle = defaultFillStyle;
```
ahocevar, line 341: This won't create a `SET_FILL_STYLE` instruction. Calling `updateFillStyle()` will.

mike-000:
> `DRAW_CHARS` instructions will also need to be updated for transparent hit detection if the intention here is to make things consistent.

It was already updated.  I have added another test.

src/ol/render/canvas/TextBuilder.js
```diff
@@ -427,11 +431,14 @@ class CanvasTextBuilder extends CanvasBuilder {
         this.text_,
         this.textKey_,
         this.strokeKey_,
-        this.fillKey_,
+        this.fillKey_ ? defaultFillStyle : this.fillKey_,
         this.textOffsetX_,
         this.textOffsetY_,
         geometryWidths,
       ]);
+      if (textState.backgroundFill) {
```
ahocevar, line 439: The change in fills state and the additional SET_FILL_STYLE instruction for hit detection would be easier to read if added right above line 411. This would also make a nice symmetry with the reset here.

ahocevar (changes requested):

src/ol/render/canvas/TextBuilder.js
```diff
@@ -404,6 +404,13 @@ class CanvasTextBuilder extends CanvasBuilder {
         geometryWidths,
       ]);
       const scale = 1 / pixelRatio;
+      // Set default fill for hit detection background
+      if (textState.backgroundFill) {
+        const fillStyle = this.state.fillStyle;
+        this.state.fillStyle = defaultFillStyle;
+        this.hitDetectionInstructions.push(this.createFill(this.state));
+        this.state.fillStyle = fillStyle;
+      }
```
ahocevar, line 413:
```suggestion
      const currentFillStyle = this.state.fillStyle;
      if (textState.backgroundFill) {
        this.state.fillStyle = defaultFillStyle;
        this.hitDetectionInstructions.push(this.createFill(this.state));
      }
```

src/ol/render/canvas/TextBuilder.js
```diff
@@ -427,11 +434,15 @@ class CanvasTextBuilder extends CanvasBuilder {
...
         this.textKey_,
         this.strokeKey_,
-        this.fillKey_,
+        this.fillKey_ ? defaultFillStyle : this.fillKey_,
         this.textOffsetX_,
         this.textOffsetY_,
         geometryWidths,
       ]);
+      // Reset previous fill
+      if (textState.backgroundFill) {
```
ahocevar, line 443:
```suggestion
      if (textState.backgroundFill) {
        this.state.fillStyle = currentFillStyle;
```

ahocevar (changes requested):

ahocevar:
@mike-000 Since you keep making changes, please indicate when this is ready for review.

mike-000:
It is ready.  After including your review I corrected a typo in the upgrade notes.

ahocevar (approved):
Thanks, @mike-000

tschaub:
For the flat style at least, I would prefer that we support `transparent` rather than distinguishing between `null` and `undefined`.

```js
const style = {
  text-fill-color: 'transparent',
  text-stroke-width: 1,
};
```

This would align with [CSS colors](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#transparent).  It looks like Mapbox style supports `transparent` as well.

ahocevar:
@tschaub `'transparent'` is already supported, and means a transparent, but hit detectable fill. I think what you're asking for (and I agree), is to use `'none'` instead of `null`.

tschaub:
Thanks, @ahocevar.  I didn’t look closely enough.  Just saw `null` being given special meaning in flat styles and wanted an alternative.  It does seem like `none` could work.

test/browser/spec/ol/renderer/map.test.js
```diff
@@ -205,6 +205,174 @@ describe('ol.renderer.Map', function () {
...
+        'text-placement': 'line',
+        'text-overflow': true,
+      });
+      map.renderSync();
+      hit = map.forEachFeatureAtPixel([50, 50], (feature, layer, geometry) => ({
+        feature,
+        layer,
+        geometry,
+      }));
+      expect(hit).to.be.ok();
```
ahocevar, line 329: @mike-000 This keeps failing for me locally. Any chance you could look into this again please?

mike-000: There no reason why pixel [50, 50] would not hit detect for me 
image
(taken from https://codesandbox.io/s/simple-forked-p7mmhh?file=/main.js)  I do see a small difference in the offset of the character if remove  `'text-placement': 'line',` but the font is so big that should not break it (and I presume the point feature test worked for you?).  If fonts differ between systems you might need a wider stroke to reach the center of the character.

test/browser/spec/ol/renderer/map.test.js
```diff
@@ -205,6 +205,174 @@ describe('ol.renderer.Map', function () {
...
```
ahocevar, line 370: Sorry @mike-000, I made the comment on the wrong line. This one is failing.

mike-000: You have a 200 x 200 image from a 100px x 100px map.  The tests do not specify a pixel ratio and you have found a pre-existing bug with line placement text when the pixel ratio is not 1.  Hit detection in v7.4.0 https://codesandbox.io/s/simple-forked-hyx359?file=/main.js did not match the rendered text.

ahocevar: The pixel ratio is indeed 2 on my Mac. Your comment sounds like you already know what's wrong. Do you think you could create a pull request with a fix, @mike-000 ?

PR #14956, opened 2023-07-29
fix Style RenderFunction type

tanhh326:
When rendering a `MultiPolygon`, it should be a coordinate data structure of a three-level nested array


ahocevar (approved):
Thanks, @tanhh326

PR #14957, opened 2023-07-29
Use Terrarium tiles in Shaded Relief examples

mike-000:
As noted in https://github.com/openlayers/openlayers/pull/14937#pullrequestreview-1552731586  the Mapbox v3 tiles used in the Shaded Relief examples no longer work and while v4 tiles are available these require an access token.  The terrarium tiles as used in the ol-mapbox-style example give better output in that region and do not need an API key.

In addition to changes necessary to handle the Terrarium format some further simplification of expressions made possible by the improvement to the `'*'` and `'+'` style operators in #14256 and the addition of `'sqrt'` is included and the change to the `'*'` and `'+'` operators documented, along with corresponding simplifications in the PMTiles and WebGL Sea Level examples.


tschaub (approved):
Looks great.  Thanks, @mike-000

PR #14962, opened 2023-08-01
Cannot create polygon from empty extent

### LINKED ISSUE #14961: ol/proj transformExtent returns bad data
**Describe the bug**
Transforming specific extents using proj.transformExtent near Antarctica from EPSG:4326 to EPSG:3031 (or to EPSG:4326) generates `[Infinity, Infinity, -Infinity, -Infinity]`. This behavior is inconsistent and certain extents will fail to be transformed while other nearby extents will transform without issue.

**To Reproduce**
```
import { transformExtent } from 'ol/proj

console.log(transformExtent([-157.6938, -72.6479, 82.6547, -80.291], 'EPSG:4326', 'EPSG:3031')) // returns [Infinity, Infinity, -Infinity, -Infinity]

console.log(transformExtent([-157.6938, -72.6479, 82.6547, -80.291], 'EPSG:4326', 'EPSG:4326')) // returns [Infinity, Infinity, -Infinity, -Infinity]

console.log(transformExtent([[-141.5768, -78.6318, 46.8319, -77.539], 'EPSG:4326', 'EPSG:3031')) // returns [-844611.0275765229, -1064747.955111334, 991233.1220461084, 929791.8549741994]
```

**Expected behavior**
It is expected that the extents should be transformed to the specified projection-like. If the input extent cannot be transformed to the target projection-like then an error should be thrown. Transforming an extent where the source and destination projection-likes are the same should return the input extent unchanged.

**Workaround**
```
import { fromExtent } from 'ol/geom/Polygon'

const extentPolygon = fromExtent([-157.6938, -72.6479, 82.6547, -80.291]);
const crsCorrectedExtentPolygon = extentPolygon.transform('EPSG:4326', 'EPSG:3031');
const crsCorrectedExtent = crsCorrectedExtentPolygon.getExtent();
```

mike-000:
Fixes #14961

Assert that the extent supplied to `ol/geom/Polygon#fromExtent` is not empty.

### DIFF
--- a/src/ol/geom/Polygon.js
+++ b/src/ol/geom/Polygon.js
@@ -5,7 +5,7 @@ import LinearRing from './LinearRing.js';
 import Point from './Point.js';
 import SimpleGeometry from './SimpleGeometry.js';
 import {arrayMaxSquaredDelta, assignClosestArrayPoint} from './flat/closest.js';
-import {closestSquaredDistanceXY, getCenter} from '../extent.js';
+import {closestSquaredDistanceXY, getCenter, isEmpty} from '../extent.js';
 import {deflateCoordinatesArray} from './flat/deflate.js';
 import {extend} from '../array.js';
 import {getInteriorPointOfArray} from './flat/interiorpoint.js';
@@ -438,6 +438,9 @@ export function circular(center, radius, n, sphereRadius) {
  * @api
  */
 export function fromExtent(extent) {
+  if (isEmpty(extent)) {
+    throw new Error('Cannot create polygon from empty extent');
+  }
   const minX = extent[0];
   const minY = extent[1];
   const maxX = extent[2];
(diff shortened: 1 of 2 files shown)


src/ol/AssertionError.js
```diff
@@ -61,6 +61,7 @@ const messages = {
   67: 'A layer can only be added to the map once. Use either `layer.setMap()` or `map.addLayer()`, not both',
   68: 'A VectorTile source can only be rendered if it has a projection compatible with the view projection',
   69: '`width` or `height` cannot be provided together with `scale`',
+  70: 'Cannot create polygon from empty extent',
```
ahocevar, line 64: We have deprecated `ol/AssertionError`. Please throw an exception directly in `ol/geom/Polygon.js`.

ahocevar (changes requested):

ahocevar (approved):

PR #14972, opened 2023-08-03
fix #14968

### LINKED ISSUE #14968: Docs: BaseVector.getFeatures shows incorrect return value
The documentation has a small quirk:

grafik

It shows `PromiseFeatureLike>>` instead of `Promise<Array<FeatureLike>>`

tanhh326:
fix #14968

> In the debugging process, I also found a lot of methods missed @Api annotations, resulting in the documentation is not rendered, wait for the follow-up and then unified change it, right?


ahocevar:
@tanhh326 If something is not marked `@api`, it is not meant to be part of the API at all, or experimental. Which symbols are you referring to?

ahocevar:
Yes. like I said, those are intentionally not part of the API.

PR #14977, opened 2023-08-04
Fix test for keyboard zoom on a Mac

fjellfly:
The test always failed because the `Ctrl` key isn't considered on a Mac.


ahocevar (approved):
Thanks, @fjellfly

PR #14982, opened 2023-08-07
Reorganise old no fill / transparent fill tests

mike-000:
Moves the no fill / transparent fill tests added in #9986 to a more appropriate place with related tests


tschaub (approved):
Thanks, @mike-000

PR #14983, opened 2023-08-08
WebGLTile: Properly render semi-transparent tiles

puckipedia:
When using the WebGL raster tile renderer with semi-transparent tiles (e.g. overlay layers), I noticed that the tiles aren't properly getting clipped by the other zoom layers, like happens on the canvas renderer: (Sample used: https://gist.github.com/puckipedia/a2a191a6684c5d9d238eed4f78133756)

<details>
<summary>Video of the situation before this PR</summary>

https://github.com/openlayers/openlayers/assets/488734/bfb2dc91-9d86-4619-a241-b0b02d6f0de6
</details>

Not only the layers involved in the fade, but also lower zoom levels, are visible, causing a bunch of visual noise that is not necessary.

This pull request solves that in a bit of a heavy-handed way: It adds proper depth testing to the WebGL helper, as well as the WebGL tile renderer:

<details>
<summary>Video of the situation after this PR</summary>

https://github.com/openlayers/openlayers/assets/488734/b628fb39-4da9-4e81-b2db-0b78e56243b0
</details>

As you can see, compared to the original situation, there's always at most two tiles being blended.

The PR first introduces the depth buffer to the `PostProcessingPass` and `RenderTarget`, as well as allowing enabling depth testing (with the default OpenGL settings) to the WebGL `Helper` . This tries to avoid any conflict with existing code using the helper, and as such is entirely opt-in. If depth isn't enabled, the depth renderbuffer will be entirely ignored.

463b5b59b7d4302be7ac113ff2989b7f2658552e then adjusts the renderer to render from the highest zoom level to the lowest, making use of the depth buffer to clip any lower zoom levels to only render where higher zoom level tiles haven't been rendered. Afterwards, it renders all the tiles with an alpha lower than 1, which allows them to blend with the existing tiles. This is necessary to preserve proper transparency behavior.

### DIFF
--- a/src/ol/array.js
+++ b/src/ol/array.js
@@ -50,6 +50,17 @@ export function ascending(a, b) {
 }
 
 /**
+ * Compare function sorting arrays in descending order.  Safe to use for numeric values.
+ * @param {*} a The first object to be compared.
+ * @param {*} b The second object to be compared.
+ * @return {number} A negative number, zero, or a positive number as the first
+ *     argument is greater than, equal to, or less than the second.
+ */
+export function descending(a, b) {
+  return a < b ? 1 : a > b ? -1 : 0;
+}
+
+/**
  * {@link module:ol/tilegrid/TileGrid~TileGrid#getZForResolution} can use a function
  * of this type to determine which nearest resolution to use.
  *
--- a/src/ol/layer/WebGLTile.js
+++ b/src/ol/layer/WebGLTile.js
@@ -274,10 +274,6 @@ function parseStyle(style, bandCount) {
 
       ${pipeline.join('\n')}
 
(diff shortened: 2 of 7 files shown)


puckipedia:
Looks like an actual regression, woops! I didn't have chrome locally, and couldn't easily use the puppeteer binary, so ended up testing with Firefox and i think i missed this test.. Will look later.

mprins:
expected: https://github.com/openlayers/openlayers/blob/main/test/rendering/cases/webgl-vectortile/expected.png

expected

actual: (from https://github.com/openlayers/openlayers/suites/14913615489/artifacts/850052736)
actual

it seems to primarily be an opacity problem

tschaub (approved):

PR #14984, opened 2023-08-08
Render VectorLayer when ReplayGroup is empty but render listener exists

M393:
Currently a `prerender` or `postrender` event is not dispatched when no features are rendered on a VectorLayer. So you can't reliably use it for drawing using a vector context.

Now it dispatches the events even when no features are rendered and adds the container to the map as long as a listener has been added.


tschaub (approved):
Thanks, @M393

PR #14989, opened 2023-08-10
Reflect that Stamen styles are hosted by Stadia Maps 

### LINKED ISSUE #14963: Update Stamen source
Stamen tiles are moving: https://docs.stadiamaps.com/guides/migrating-from-stamen-map-tiles/

This will be a breaking change since the tiles will require an API key or registered domain to work.

ianthetechie:
Closes #14963, and additionally adds all other styles available from Stadia Maps + an example.

### DIFF
--- a/examples/animated-gif.js
+++ b/examples/animated-gif.js
@@ -3,7 +3,7 @@ import Map from '../src/ol/Map.js';
 import Point from '../src/ol/geom/Point.js';
 import View from '../src/ol/View.js';
 import {Icon, Style} from '../src/ol/style.js';
-import {Stamen, Vector as VectorSource} from '../src/ol/source.js';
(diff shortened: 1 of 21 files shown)


tschaub:
Thank you for your work on this, @ianthetechie.  It looks great.  I'll follow up separately with a change to the [upgrade notes](https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md) and will likely ask for your 👀

tschaub (approved):

ianthetechie:
Sure thing! Tag me where needed @tschaub and I'll be happy to write it up in whatever style is needed.

tschaub:
@ianthetechie see https://github.com/openlayers/openlayers/pull/14997/files - let us know if you have any suggested changes.

PR #14992, opened 2023-08-11
Use none to avoid hit detection of text fills

tschaub:
Following up on https://github.com/openlayers/openlayers/pull/14943#issuecomment-1672441870, this changes syntax for avoiding hit detection on text fills when using the flat literal syntax.

We're adding support for expressions in styles in #14780.  Currently, the types supported by expressions do not include `null` or `undefined`.  We may need to add support for this, but I'm going to avoid it until there is a compelling case.

Instead of `'text-fill-color': null`, this change makes it so `'text-fill-color': 'none'` will results in a text where the fill is not hit detected.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -4,7 +4,55 @@
 
 #### Hit detection with Text fill
 
-Hit detection of fill for Text styles is now consistent with that for Circle and RegularShape styles. Transparent `fill` and `backgroundFill` is detected, with no hit detection over unfilled shapes. To get the previous behavior in Text styles where a transparent fill may have been used to avoid the default fill specify `fill: null`. If using FlatStyle notation specify `'text-fill-color': 'none'`.
+Previously, text labels with transparent fills were not hit detected.  Now, you can control whether a transparent fill in a text label is hit detected or not.
+
+To create a text style with a transparent fill that will be hit detected, simply exclude the `fill` or use a fill with `'transparent'` as the color.
+```js
+// transparent fill, will be hit detected
+const style = Style({
+  text: new Text({
+    fill: new Fill({
+      color: 'transparent',
+    }),
+    stroke: new Stroke({
+      color: 'red',
+      width: 2,
+    }),
+  }),
+});
+```
+Or, if using the flat literal style syntax:
+```js
+// transparent fill, will be hit detected
+const style = {
+  'text-fill-color': 'transparent',
+  'text-stroke-color': 'red',
+  'text-stroke-width': 2,
+}
+```
+
+By contrast, if you want a transparent fill that will not be hit detected, do the following:
+```js
+// absent fill, will not be hit detected
+const style = Style({
+  text: new Text({
+    fill: null,
+    stroke: new Stroke({
+      color: 'red',
+      width: 2,
+    }),
+  }),
+});
+```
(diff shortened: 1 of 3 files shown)


changelog/upgrade-notes.md
```diff
@@ -4,7 +4,7 @@
 
 #### Hit detection with Text fill
 
-Hit detection of fill for Text styles is now consistent with that for Circle and RegularShape styles. Transparent `fill` and `backgroundFill` is detected, with no hit detection over unfilled shapes. To get the previous behavior in Text styles where a transparent fill may have been used to avoid the default fill specify `fill: null`. If using FlatStyle notation specify `'text-fill-color': null`.
+Hit detection of fill for Text styles is now consistent with that for Circle and RegularShape styles. Transparent `fill` and `backgroundFill` is detected, with no hit detection over unfilled shapes. To get the previous behavior in Text styles where a transparent fill may have been used to avoid the default fill specify `fill: null`. If using FlatStyle notation specify `'text-fill-color': 'none'`.
```
tschaub, line 7: These upgrade notes are not clear to me.  Would it be true to say the following:

Previously, text labels with transparent fills were not hit detected.  Now, you can control whether a transparent fill in a text label is hit detected or not.

To create a text style with a transparent fill that will be hit detected, simply exclude the `fill` or use a fill with `'transparent'` as the color.
```js
// implicitly transparent fill, will be hit detected
const style = Style({
  text: new Text({
    stroke: new Stroke({
      color: 'red',
      width: 2,
    }),
  }),
});
```
Or, if using the flat literal style syntax:
```js
// implicitly transparent fill, will be hit detected
const style = {
  'text-stroke-color': 'red',
  'text-stroke-width': 2,
}
```

By contrast, if you want a transparent fill that will not be hit detected, do the following:
```js
// explicitly absent fill, will not be hit detected
const style = Style({
  text: new Text({
    fill: null,
    stroke: new Stroke({
      color: 'red',
      width: 2,
    }),
  }),
});
```
Or, if using the flat literal style syntax:
```js
// explicitly absent fill, will not be hit detected
const style = {
  'text-fill-color': 'none',
  'text-stroke-color': 'red',
  'text-stroke-width': 2,
}
```

ahocevar: Yes, having this in the upgrade notes instead of what's currently there would make it much clearer.

ahocevar (approved):
Thanks, @tschaub! Your suggestion for new wording in the release notes is very good. Please change.

changelog/upgrade-notes.md
```diff
@@ -4,7 +4,55 @@
 
 #### Hit detection with Text fill
 
-Hit detection of fill for Text styles is now consistent with that for Circle and RegularShape styles. Transparent `fill` and `backgroundFill` is detected, with no hit detection over unfilled shapes. To get the previous behavior in Text styles where a transparent fill may have been used to avoid the default fill specify `fill: null`. If using FlatStyle notation specify `'text-fill-color': null`.
+Previously, text labels with transparent fills were not hit detected.  Now, you can control whether a transparent fill in a text label is hit detected or not.
+
+To create a text style with a transparent fill that will be hit detected, simply exclude the `fill` or use a fill with `'transparent'` as the color.
```
mike-000, line 9: excluding the fill will default to #333, not transparent

Issue #14166, opened 2022-10-03
Coordinate type could be more explicit


rooby:
**Describe the bug**
The Coordinate type is `number[]`, however it would more accurately be `[number, number]`

MoonE:
In many cases Coordinate can be also three or four numbers.
http://localhost:8080/apidoc/module-ol_geom_Geometry.html#~GeometryLayout

Maybe change it to something like `[number, number] | [number, number, number] | [number, number, number, number]` in the meantime and have different typedefs for coordinates with only xy and coordinates with more number in the long run.

rooby:
@MoonE  Oh yeah, I forgot about that. Thanks.

tschaub:
Nothing in the library limits the number of dimensions in a coordinate.   I’m not sure it makes sense for the type to do so.

rooby:
Ah ok, no worries. Thanks for the quick reply.

Issue #9122, opened 2019-01-19
Type of ol.coordinate, Typescript?


michaelletzgus:
ol.Coordinate is defined as a 2-tupel [number, number] for the XY format.

But when reading features from GPX files feature.getGeometry().getCoordinates() returns coordinates as XYZ or XYZM. When using TypeScript accessing a c[3] gives an error since c is a 2-tupel.

So, in fact, the behavior of getCoordinates() is wrong - or its signature.

Maybe ol.Coordinate should always be a 4-tupel?

ahocevar:
In OpenLayers, it is defined as `Array<number>` (plain JavaScript with JSDoc types). Are you using external types? Those are not maintained by OpenLayers developers.

michaelletzgus:
I tried to use [the npm package](https://www.npmjs.com/package/ol) which suggested to install [@types/ol](https://www.npmjs.com/package/@types/ol):

`Try `npm install @types/ol` if it exists or add a new declaration (.d.ts) file containing `declare module 'ol';`

As I understood the ongoing discussions regarding TypeScript the `.d`-files are created from JSDoc.
But @types/ol (which depends on @types/openlayers) defines:

`export type Coordinate = [number, number]`

Maybe the `@types/ol` is build upon an outdated version 4.6.3?

ahocevar:
The [@types/ol](https://npmjs.com/package/@types/ol) package is not maintained by OpenLayers. So it will probably be better to create an issue there.

Note that the "Try npm install `@types/ol` if it exists or add a new declaration (.d.ts) file containing declare module 'ol';`" message does not come from OpenLayers.

PR #14994, opened 2023-08-11
Improve the description of the Coordinate type

Isti115:
As discussed previously in #9122, #14166 and partially #14167, coordinates can have more that two dimensions. Not including this possibility in the description is misleading and makes it seem like the `Array<number>` type is incorrect. (Which it actually is, but not in such a blatant way.)


tschaub (approved):
Thanks for the doc contribution, @Isti115

PR #14999, opened 2023-08-11
Doc updates

tschaub:
Currently, comments with `@fileoverview` (or `@file`) get ignored (they end up `undocumented: true` due to the api plugin).  If a comment with `@fileoverview` is given an `@api` tag, the comment is included inline in the index page ([we concatenate](https://github.com/openlayers/openlayers/blob/f68564107019688658f6387a067f61124ec7fdf7/config/jsdoc/api/template/publish.js#L624-L647) all `file` type tags with the index docs).

Since we don't currently use the `@fileoverview` and `@api` tag combo anywhere, I repurposed this to serve as a way to add a description for a module.

In this branch, I've also updated the templates so the table for typedefs looks more like the table for constructor options.

I've updated the [docs for the `ol/style/flat.js` module](https://deploy-preview-14999--ol-site.netlify.app/en/latest/apidoc/module-ol_style_flat) to demonstrate the change.  Here is an animation of the before/after change (after has the module description, example, and typedef table changes).

doc-change

### DIFF
--- a/config/jsdoc/api/template/publish.js
+++ b/config/jsdoc/api/template/publish.js
@@ -621,8 +621,8 @@ exports.publish = function (taffyData, opts, tutorials) {
     generate('Global', [{kind: 'globalobj'}], globalUrl);
   }
 
-  // index page displays information from package.json and lists files
-  const files = find({kind: 'file'});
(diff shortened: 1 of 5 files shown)


ahocevar (approved):
Great improvement, thanks!

PR #15001, opened 2023-08-12
Fix text placement line hit detection for pixelRatio !== 1

ahocevar:
~~This pull request makes the recently added tests for hit detection of line placement texts better readable.~~

This pull request fixes the pixel ratio problem identified by the new text fill tests.


test/browser/spec/ol/renderer/Map.test.js
```diff
@@ -507,82 +507,121 @@ describe('ol/renderer/Map.js', function () {
...
+            );
+            expect(hit).to.be(undefined);
+            done();
+          } catch (e) {
+            done(e);
+          }
+        });
+      });
+
+      it('with transparent fill', function (done) {
```
ahocevar, line 593: @mike-000 This test is failing on MacOS, and to this point I'm unable to find out why.

mike-000:
First test has no fill but detects wide stroke
Second test has no fill and thin stroke, detects nothing
Third test detects transparent fill

For the tests to pas use `pixelRatio: 1` in the map options.

ahocevar:
@mike-000 This looks fishy. Only the third test fails. If there is something wrong with pixel ratio handling for text hit detection, I would expect all three tests to fail on a device with a pixel ratio of 2.

mike-000:
> @mike-000 This looks fishy. Only the third test fails. If there is something wrong with pixel ratio handling for text hit detection, I would expect all three tests to fail on a device with a pixel ratio of 2.

The wide stroke in the first test would extend outwards as well as into the center of the character - just enough to get a hit a [50, 50] even though the hit detection image was misplaced.  Increase the pixel ratio beyond 2 the misplacement increases and it also would fail.

ahocevar:
@mike-000 Nevermind. I'm looking into a fix for the pixel ratio issue now.

ahocevar:
This is now ready for review. For those who have been following: I re-purposed this pull request to not only make the tests more readable, but fix the pixel ratio issue I encountered when running the tests on a system with a device pixel ratio other than 1.

tschaub (approved):
Thanks for your work on this, @ahocevar

PR #15002, opened 2023-08-13
Do not override `caption-side` in featureInfo styles

### LINKED ISSUE #14988: GetFeatureInfo output is reordered in examples
**Describe the bug**
css appears to be reordering the output in the GetFeatureInfo examples.

**To Reproduce**
Compare the output when clicking in https://openlayers.org/en/v6.3.1/examples/getfeatureinfo-tile.html and https://openlayers.org/en/latest/examples/getfeatureinfo-tile.html  The `ne_10m_admin_0_countries` caption used to appear above the table, as it is served.  It now appears below the table, presumably due to a css change.

**Expected behavior**
Output to be displayed in the order it is served.

mike-000:
Fixes #14988

https://deploy-preview-15002--ol-site.netlify.app/en/latest/examples/getfeatureinfo-tile.html

### DIFF
--- a/site/src/theme/site.css
+++ b/site/src/theme/site.css
@@ -525,3 +525,7 @@ div.code-toolbar>.toolbar button:before {
     max-width: 1140px;
   }
 }
+
+table.featureInfo {
+  caption-side: initial;
+}


ahocevar (approved):

ahocevar (approved):

tschaub:
Would an alternative here be to not use the `caption-side` class?

mike-000:
> Would an alternative here be to not use the `caption-side` class?

I presume you mean `caption-top` https://bootstrapshuffle.com/classes/tables/caption-top. That would have to be set on the table, setting the class on the info div has no effect.  The table is returned by the WMS server which styles it with a `featureInfo` class which does not specify `caption-side`.

tschaub:
Got it, I misinterpreted what was going on.

PR #15003, opened 2023-08-13
Styling with logical operators and type assertions

tschaub:
Building on #14780, this adds support for additional expression operators: `any`, `all`, `!`, `resolution`, `string`, and `number`.  The logical operators (`any`, `all`, `!`) and type assertions (`string`, `number`) behave like those in the [Mapbox style spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/).  We should be able to add support for `zoom` instead of (or in addition to) `resolution`, but I haven't done that yet (sticking with the existing style function interface).

I've updated a few examples.  I'll leave a comment inline about a few specific issues.

### DIFF
--- a/examples/osm-vector-tiles.js
+++ b/examples/osm-vector-tiles.js
@@ -3,58 +3,60 @@ import TopoJSON from '../src/ol/format/TopoJSON.js';
 import VectorTileLayer from '../src/ol/layer/VectorTile.js';
 import VectorTileSource from '../src/ol/source/VectorTile.js';
 import View from '../src/ol/View.js';
-import {Fill, Stroke, Style} from '../src/ol/style.js';
 import {fromLonLat} from '../src/ol/proj.js';
 
 const key = 'uZNs91nMR-muUTP99MyBSg';
 
-const roadStyleCache = {};
-const roadColor = {
-  'major_road': '#776',
-  'minor_road': '#ccb',
-  'highway': '#f39',
-};
-const buildingStyle = new Style({
-  fill: new Fill({
-    color: '#666',
-    opacity: 0.4,
-  }),
-  stroke: new Stroke({
-    color: '#444',
-    width: 1,
-  }),
-});
-const waterStyle = new Style({
-  fill: new Fill({
-    color: '#9db9e8',
-  }),
-});
-const roadStyle = function (feature) {
-  const kind = feature.get('kind');
-  const railway = feature.get('railway');
-  const sort_key = feature.get('sort_key');
-  const styleKey = kind + '/' + railway + '/' + sort_key;
-  let style = roadStyleCache[styleKey];
-  if (!style) {
-    let color, width;
-    if (railway) {
-      color = '#7de';
-      width = 1;
-    } else {
-      color = roadColor[kind];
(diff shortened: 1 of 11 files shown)


examples/osm-vector-tiles.js
```diff
@@ -3,58 +3,60 @@ import TopoJSON from '../src/ol/format/TopoJSON.js';
...
-const roadStyleCache = {};
-const roadColor = {
-  'major_road': '#776',
-  'minor_road': '#ccb',
-  'highway': '#f39',
-};
-const buildingStyle = new Style({
-  fill: new Fill({
-    color: '#666',
-    opacity: 0.4,
```
tschaub, line 20: The Fill constructor doesn't have an `opacity` option, so this had no effect.  In the updated style, I've set the opacity in the `fill-color` and `stroke-color`, so the rendered output looks different.

examples/osm-vector-tiles.js
```diff
@@ -3,58 +3,60 @@ import TopoJSON from '../src/ol/format/TopoJSON.js';
...
+      'stroke-color': [
+        'match',
+        ['get', 'kind'],
+        'major_road',
+        '#776',
+        'minor_road',
+        '#ccb',
+        'highway',
+        '#f39',
+        'none',
```
tschaub, line 39: I decided to build on the logic for handling `none` as a fill color.  The canvas style builder will now return `null` for a stroke if the color evaluates to `none`.

src/ol/render/canvas/style.js
```diff
@@ -179,7 +179,11 @@ export function buildRuleSet(rules, context) {
       }
       someMatched = true;
       for (const styleEvaluator of compiledRules[i].styles) {
-        styles.push(styleEvaluator(context));
+        const style = styleEvaluator(context);
+        if (!style) {
+          continue;
+        }
+        styles.push(style);
```
tschaub, line 186: This change handles the `null` styles that can be generated if all of the symbolizers are `null` below.

examples/vector-layer.js
```diff
@@ -10,20 +10,9 @@ const vectorLayer = new VectorLayer({
...
-    },
-    {
-      // otherwise, use this style
-      else: true,
-      style: {'fill-color': '#eee'},
-    },
-  ],
+  style: {
+    'fill-color': ['string', ['get', 'COLOR'], '#eee'],
+  },
```
tschaub, line 15: It was a bit of a hack to use a `filter` to evaluate the color and then have a fallback style.  The `string` type assertion with a fallback is well suited for this.

tschaub:
It looks like the key we use for tile.nextzen.org tiles doesn't work for the branch preview (not related to this change).  The example renders locally and I assume will work when deployed to openlayers.org.

Here is a before/after view of the example:

osm-tiles

In the after frame, the buildings are lighter due to the opacity issue mentioned above.  In addition, only railway, major_road, minor_road, and highway features are rendered from the roads layer.  In the before case, other lines get the default stroke.  In the after case, nothing gets rendered due to the `none` stroke color.

ahocevar (approved):
Beautiful! Thanks @tschaub, especially for the very well annotated pull request.

PR #15005, opened 2023-08-14
Fix return type of Layer#renderFrame

### LINKED ISSUE #15004: Type incompatibility between CanvasVectorLayerRenderer and LayerRenderer<any>
**Describe the bug**
When updating ol from 7.4.0 to 7.5.0 we get the following type error:
image
Apparently in one case the method renderFrame() is allowed to return null and in the other not.

**Expected behavior**
The code should compile without errors as in 7.4.0

MoonE:
Fixes #15004

VectorLayer#renderFrame would return null, but it wasn't annotated until #14984

### DIFF
--- a/src/ol/renderer/Layer.js
+++ b/src/ol/renderer/Layer.js
@@ -69,8 +69,8 @@ class LayerRenderer extends Observable {
    * Render the layer.
    * @abstract
    * @param {import("../Map.js").FrameState} frameState Frame state.
-   * @param {HTMLElement} target Target that may be used to render content to.
-   * @return {HTMLElement} The rendered element.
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @MoonE

mprins:
Will this be released as 7.5.1?

tschaub:
@mprins - I’ll prepare a patch release later today.  Let me know if there is anything else you think should be included in it (we’ve already begun making changes toward 8 in main, so I would just apply this single change to a 7.5 branch).

mprins:
@tschaub I've only run into this specific issue, so AFAIC just this PR

tschaub:
This fix has been included in the 7.5.1 release.  Thanks, @MoonE.

PR #15015, opened 2023-08-14
Fix OGCVectorTile source types

### LINKED ISSUE #15014: Typescript Error in 7.5.0
the 7.5.0 Version has a problem with the Typescript definitions in the `OGCVectorTile.d.ts`

image

In the d.ts. file VectorTile is used correctly but imported as VectorTile_1. 

image

This is after i fixed the following Bug manually in the node_modules folder: https://github.com/openlayers/openlayers/issues/15004

@tschaub  Maybe you can also take a look for the 7.5.1 Bugfix Release ?

MoonE:
Fixes #15014

I suspect the problem is that there is `ol/source/VectorTile` and also `ol/VectorTile` which had the same name.

### DIFF
--- a/src/ol/source/OGCVectorTile.js
+++ b/src/ol/source/OGCVectorTile.js
@@ -2,7 +2,7 @@
  * @module ol/source/OGCVectorTile
  */
 
-import VectorTile from './VectorTile.js';
+import VectorTileSource from './VectorTile.js';
 import {getTileSetInfo} from './ogcTileUtil.js';
 import {error as logError} from '../console.js';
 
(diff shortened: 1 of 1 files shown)


ahocevar (approved):

MoonE:
New types: https://deploy-preview-15015--ol-site.netlify.app/en/latest/ol/source/OGCVectorTile.d.ts

tschaub:
Thanks for tracking this down and fixing it, @MoonE.

Is there anything else we can change here?  It feels fragile to rely on nobody repeating a local name in a default import declaration.  Ideally, these local names could be anything and it wouldn't break anything.

MoonE:
I reported it to the TypeScript project https://github.com/microsoft/TypeScript/issues/55375

The issue seems to be specific to the usage of the  class as a type in the Options object, and only  when it is an optional property. So it should quite rare.

tschaub:
This fix has been included in the 7.5.1 release.  Thanks, @MoonE.

PR #15021, opened 2023-08-17
Fix various typos

luzpaz:
Found via `codespell -q 3 -S "./node_modules" -L coordinatess,nd`

### DIFF
--- a/test/browser/spec/ol/format/ows.test.js
+++ b/test/browser/spec/ol/format/ows.test.js
@@ -59,7 +59,7 @@ describe('ol.format.OWS 1.1', function () {
         'xmlns:xlink="http://www.w3.org/1999/xlink" >' +
         '<ows:ServiceIdentification>' +
         '<ows:Title>Web Map Tile Service</ows:Title>' +
-        '<ows:Abstract>Service that contrains the map access interface ' +
+        '<ows:Abstract>Service that constrains the map access interface ' +
         'to some TileMatrixSets</ows:Abstract>' +
         '<ows:Keywords>' +
         '<ows:Keyword>tile</ows:Keyword>' +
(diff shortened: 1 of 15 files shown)


ahocevar (approved):
Thanks, @lupaz

test/browser/spec/ol/format/ows.test.js
```diff
@@ -79,7 +79,7 @@ describe('ol.format.OWS 1.1', function () {
     const serviceIdentification = obj.ServiceIdentification;
     expect(serviceIdentification).to.be.ok();
     expect(serviceIdentification.Abstract).to.eql(
-      'Service that contrains the map access interface to some TileMatrixSets'
+      'Service that contains the map access interface to some TileMatrixSets'
```
tschaub, line 82: Looks like this also needs to be corrected in these places:
 * https://github.com/openlayers/openlayers/blob/0519e908d55fbeb95408a06644930adfe2256cda/test/browser/spec/ol/format/wmts/ogcsample.xml#L5
 * https://github.com/openlayers/openlayers/blob/0519e908d55fbeb95408a06644930adfe2256cda/test/browser/spec/ol/format/ows.test.js#L62

mike-000: Should it be `contains` or `constrains` ?

ahocevar: Should indeed be `constrains`.

tschaub: @luzpaz - in case it helps, it doesn't really matter what the content is, it just needs to be the same where we pass it to the [`parse` function](https://github.com/openlayers/openlayers/blob/b03ba31affe77dfb1bf190118a5454455504a043/test/browser/spec/ol/format/ows.test.js#L62) and where we pass it to the [`eql` function](https://github.com/openlayers/openlayers/blob/b03ba31affe77dfb1bf190118a5454455504a043/test/browser/spec/ol/format/ows.test.js#L82).

(The `<ows:AccessConstraints>` element is not really related to the content of the `<ows:Abstract>`, but the original author probably did mean to write "Service that constrains the map access interface..." for the abstract.)

luzpaz: Understood. Here are the requested revisions: 1bc0f6a

tschaub (approved):

tschaub:
Thanks, @luzpaz

PR #15027, opened 2023-08-18
Make BingMaps source 'placeholderTiles' optional

### LINKED ISSUE #15026: Fix type definition of placeholderTiles options of BingMaps source
**Describe the bug**
A `placeholderTiles` options has been recently introduced in #14832

Its type definition is `boolean`, i.e. it is typed as a required parameter to set among the options.  However, it seems showcased as an "optional option", i.e. the code and example assume that it can be left undefined. That's all fine.

When used in a TypeScript project, since the property is not optional, we are force to set it, which does not allow to let the default behaviour of leaving it undefined to be a possible option.

**To Reproduce**
It is a minor technical issue, and it is not really reproducible.

**Expected behavior**
It should be possible to leave the option undefined.

adube:
Fixes #15026 

Makes the `placeholderTiles` option of the BingMaps source object optional.

This fix is just about fixing the type definition to make the option optional.  Without it, the option is forced to be defined as boolean when used in a TypeScript project.

### DIFF
--- a/src/ol/source/BingMaps.js
+++ b/src/ol/source/BingMaps.js
@@ -69,7 +69,7 @@ const TOS_ATTRIBUTION =
  * Choose whether to use tiles with a higher or lower zoom level when between integer
  * zoom levels. See {@link module:ol/tilegrid/TileGrid~TileGrid#getZForResolution}.
  * @property {boolean} [placeholderTiles] Whether to show BingMaps placeholder tiles when zoomed past the maximum level provided in an area. When `false`, requests beyond
- * the maximum zoom level will return no tile. When `true`, the placeholder tile will be returned. When no set, the default behaviour of the imagery set takes place,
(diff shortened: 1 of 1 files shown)


adube:
@Shane98c if you could take a quick look at this and make sure that it matches your original intentions, it would be nice.  It's a pretty small patch.

Thanks a lot for your contribution.

Shane98c (approved):
This makes total sense! Great catch.

mike-000:
This **is** changing the default behavior, which is to use the Bing's default for the imagery set.  Aerial and Ordnance Survey default to placeholders, Aerial with Labels and Roads allow do not, so allow overzooming.  The change makes overzooming the default for for all imagery sets.  The only change to make it optional in typescript is to change

`@property {boolean} placeholderTiles
`

to

`@property {boolean} [placeholderTiles]`

adube:
Understood.  I'm on it.

tschaub (approved):

tschaub:
Thanks, @adube

PR #15028, opened 2023-08-19
Keep declutter tree for hit detection

ahocevar:
This pull request fixes a regression that was introduced with #14971. Hit detection of vector tile layers does not work anymore when `declutter` is `true`, because hit detection of decluttered items relies on a declutter tree that contains all decluttered items. The new `flushDeclutter` method deleted the declutter tree, which is good for rendering performance, but breaks hit detection.

This pull request restores the previous behavior, i.e. that the `declutterTree` is kept after the end of the render cycle.

### DIFF
--- a/src/ol/renderer/Composite.js
+++ b/src/ol/renderer/Composite.js
@@ -161,7 +161,6 @@ class CompositeMapRenderer extends MapRenderer {
     for (let i = layers.length - 1; i >= 0; --i) {
       layers[i].renderDeclutter(frameState);
     }
-    frameState.declutterTree = null;
     layers.length = 0;
   }
 }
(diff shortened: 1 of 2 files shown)


rdewit:
@ahocevar , I ran into this problem with v7.5 and could not reproduce it with the existing examples! I worked around it for the time being by setting `declutter: false`. Please let me know if you want me to try out this PR to see if it solves my situation. Otherwise, I'll wait for it to end up in `dev`.

ahocevar:
@rdewit If you can, please try it out. But it sounds like you're running into the same issue that made me encounter this regression.

tschaub (approved):

tschaub:
@rdewit let us know if a patch release would be helpful after this is in or if the dev releases will work - there are some breaking changes in main already.

rdewit:
@ahocevar, I can confirm that this fix solves the issue on my end too! 

@tschaub, if nobody else encounters this issue, I'm happy to hold off upgrading until v8 has been released. Thanks for offering to create a patch release though! 🙏

bartvde:
@rdewit just pointed me to this issue as we are also running into this with 7.5.0/7.5.1 so a patch release would really be appreciated @tschaub or @ahocevar

ahocevar:
I'm going to create a patch release right now, @bartvde.

bartvde:
@ahocevar thanks Andreas, that did the trick

PR #15030, opened 2023-08-19
Add `params` option for any additional customization parameters accepted by Bing imagery sets

### LINKED ISSUE #15019: Add MKT and UR parameters to Bing Map Api
Hello, at the moment the bing map api take into consideration the culture param only, is it possible to add the mkt (market parameter) and UR ( user region ), hence those are very important too to provide a much suitable experience to the user, i am forced to appended it to the api call each time. ( that would be a great addition :D)

in the ol.js

`const i="https://dev.virtualearth.net/REST/v1/Imagery/Metadata/"+this.imagerySet_+"?uriScheme=https&include=ImageryProviders&key="+this.apiKey_+"&mkt=ma-Ma&countryRegion=Morocco&ur=ma&c="+this.culture_;`

Thanks

mike-000:
Fixes #15019


nabilelmahiri (approved):

ahocevar (changes requested):
Thanks @mike-000 for your effort on this!

What about the previous suggestion to add two options, `style` and `userRegion`, instead of a generic `params` option? Then it would maybe also be possible to add types or typedefs for better type safety of the new options.

Regarding the example, I'd prefer showing this functionality in a separate dedicated example, instead of making an existing complex example even more complex.

mike-000:
> What about the previous suggestion to add two options, `style` and `userRegion`, instead of a generic `params` option? Then it would maybe also be possible to add types or typedefs for better type safety of the new options.

I though I would keep it generic as with other sources which use a `params` or `dimension` option because there are more addition parameters available than just those two, e.g. `mapLayer: 'trafficFlow'`.

PR #15032, opened 2023-08-21
Wait until font loading has finished

ahocevar:
This pull request fixes an ephemeral issue with the layer-vectortile-icon-label rendering test: Sometimes the test fails because the map uses the fallback font until the desired font has finished loading.


tschaub (approved):
Thanks, @ahocevar

Issue #14663, opened 2023-04-12
Projection issue with Equi7 and GeoTiff source


m-mohr:
**Describe the bug**
It looks like there's an issue with projection detection or reprojection for the Equi7 projection when I load a COG.

**To Reproduce**
Sandbox: https://codesandbox.io/s/geotiff-reprojection-forked-p3utjd?file=/main.js

The url for the epsg4326 variable is shown over the OSM baselayer.
If you change to the url for the equi7 variable, it's shown "somewhere" but definitively not over the baselayer.
The files though are at the same location, just one COG is in EPSG:4326 and one is in Equi7 (see https://github.com/TUW-GEO/Equi7Grid). 
Loading them in QGIS shows them at the same location.

**Expected behavior**
The Equi7 COG should be shown at the correct location (and over the baselayer). Not sure what the best way to achieve this is.
The baselayer should stay in the default projection (EPSG:3857) and the COGs should be reprojected.

I have the correct proj string (`+proj=aeqd +lat_0=53 +lon_0=24 +x_0=5837287.81977 +y_0=2121415.69617 +datum=WGS84 +units=m +no_defs`), but I'm not sure how I can assign it to the source. In other OL sources such as GeoJSON it seems I can set a source projection, but it seems for GeoTiff source I can't do that?!

fyi @tschaub :-)

mike-000:
`view: source.getView()` will set the view to to COG projection and reproject the baselayer.

To display the COG in the baselayer projection see https://codesandbox.io/s/geotiff-reprojection-forked-21zp8q

or for multiple COGs see https://openlayers.org/en/latest/examples/multiple-cogs.html

The Equi7 COG is returning EPSG:4326 as its projection but the coordinates are not EPSG:4326

![image](https://user-images.githubusercontent.com/49240900/231586359-035675d0-a635-4e0a-9e4e-3e8325dd03ad.png)

I tried overriding that https://codesandbox.io/s/geotiff-reprojection-forked-umgdey but there is no RGB data visible

m-mohr:
Thanks! 

Ah yes, indeed, I shouldn't have set the View from the source. I guess I was too tired.

@SerRichard can you say why the COG is returning EPSG:4326 as projection?

mike-000:
[...]

PR #15036, opened 2023-08-21
Allow to set GeoTiff projection #14663 #14927

m-mohr:
Allow to set the GeoTiff projection, for example if the read projection by OL/geotiff.js is not correct.

Provides a solution for #14663 and extracts the change from PR #14927, for which this is also needed.


tschaub:
Thanks, @m-mohr

PR #15042, opened 2023-08-22
Add OL STAC example

m-mohr:
Adds an OL STAC example as requested in https://github.com/openlayers/openlayers/pull/14927

It does complain about missing dependencies. Do I need to add it do devDependencies? Edit: Yes, that seems to work. Hope that's the right way of doing it.

### DIFF
--- a/examples/stac-item.html
+++ b/examples/stac-item.html
@@ -1,0 +1,14 @@
+---
+layout: example.html
+title: STAC support
+shortdesc: Rendering a STAC Item with geometry and COG.
+docs: >
+  This example uses the npm package [`ol-stac`](https://m-mohr.github.io/ol-stac/) which
+  adds [STAC](https://stacspec.org) support to OpenLayers.
+
+  It can for example show geometries and assets from SpatioTemporal Asset Catalog (STAC) Items as a layer group.
+
+  For more examples visit <https://m-mohr.github.io/ol-stac/en/latest/examples/>.
+tags: "stac, geotiff, cog, layergroup"
+---
+<div id="map" class="map"></div>
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "metalsmith": "^2.5.0",
     "mocha": "10.2.0",
     "ol-mapbox-style": "^11.0.1",
+    "ol-stac": "^1.0.0-beta.6",
     "pixelmatch": "^5.1.0",
     "pngjs": "^7.0.0",
     "proj4": "2.9.0",
--- a/examples/stac-item.js
+++ b/examples/stac-item.js
@@ -1,0 +1,31 @@
+import Map from '../src/ol/Map.js';
+import OSM from '../src/ol/source/OSM.js';
+import STAC from 'ol-stac';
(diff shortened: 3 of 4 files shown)


mike-000:
Yes, I have sometimes found it necessary to add a nodata setting to force an alpha band to be added for non-parallel reprojection, but doing that unnecessarily overrides any genuine nodata.  It might be better if any supplied setting was appended instead of simply overriding.

mike-000:
#15048 will ensure alpha can be set in regions outside the original source extent without overriding the metadata.  The metadata would handle alpha inside the original extent.

examples/stac-item.html
```diff
@@ -0,0 +1,12 @@
+---
+layout: example.html
+title: STAC support
+shortdesc: Rendering a STAC Item with geometry and COG.
+docs: >
+  This example uses the npm package [`ol-stac`](https://m-mohr.github.io/ol-stac) which
```
tschaub, line 6:
```suggestion
  This example uses the npm package [`ol-stac`](https://m-mohr.github.io/ol-stac/) which
```

Minor, but this change avoids the unnecessary redirect.

package.json
```diff
@@ -93,6 +93,7 @@
     "metalsmith": "^2.5.0",
     "mocha": "10.2.0",
     "ol-mapbox-style": "^11.0.1",
+    "ol-stac": "^1.0.0-beta.5",
```
tschaub, line 96: Adding this to our dev dependencies will make it available when building the examples.  However, we'll likely get into a situation with duplicate version of `ol` since [your package lists `ol`](https://github.com/m-mohr/ol-stac/blob/v1.0.0-beta.5/package.json#L31) as a direct dependency.

When we release 8.0.0, for example, since that version doesn't satisfy your `^7.0.0` constraint, there will be multiple versions of the `ol` package installed and bundled with the examples.

I know that package versioning can be a mess, but my preference when publishing a "plugin" type package is to _not_ include it in `dependencies`.  Instead, I think it works ok if you have a very generous version constraint listed in `peerDependencies` and then you can have a more specific/constrained version in `devDependencies`.

As an example, here is a popular looking Vue component library that doesn't list `vue` in its `dependencies` but does list it in its `devDependencies`: https://github.com/vuetifyjs/vuetify/blob/v3.3.14/package.json

And a similarly popular React component library that doesn't list `react` in its `dependencies` but does list it in its `peerDependencies`: https://github.com/mui/material-ui/blob/v5.14.6/packages/mui-base/package.json

m-mohr: Thanks, this makes sense, I just need to check whether that has any side effects.

m-mohr: Done, seems to work just fine.

m-mohr, line 96:
```suggestion
    "ol-stac": "^1.0.0-beta.6",
```

To be changed once ol-stac has been rereleased. Makes OL a peer dependency of ol-stac.

m-mohr:
Updated and re-released ol-stac with ol as peer dependency.

tschaub (approved):
Thanks, @m-mohr

PR #15044, opened 2023-08-23
Set correct maxZoom in OSM Vector Tiles example

mike-000:
The source is not serving tiles above zoom level 16, consequently blurred interim tiles are seen with higher view zoom levels instead of sharp render tiles based on tiles served at the correct maxZoom.


ahocevar (approved):
Thanks, @mike-000

PR #15048, opened 2023-08-24
Add alpha band if `nodata` is specified as `null`

mike-000:
As mentioned in #15042 an alpha band is needed in non-parallel reprojections of GeoTIFFs otherwise zero values used to fill regions outside the source tile grid will appear black in the reprojection tiles instead of transparent.  Specifying `nodata: NaN` would achieve that but when used in a general purpose application where metadata values may or may not exist that would override any which did exist.  Distinguishing between `null` and `undefined` ensures an alpha band will be added without overriding the metadata.

Updated existing test to use `null` instead of `NaN` to add alpha, and added new test to show `null` does not override metadata.


mike-000:
Reprojection does not know the purpose of the bands, it just reprojects whatever is in them and leaves regions outside the source coverage as zeros (which will be black or transparent depending on whether one of the bands is alpha).  The documented purpose of `nodata` is to add an alpha band.  The alternative would be an additional boolean option such as `addAlpha` which is considered in addition to any `nodata` setting or metadata.

PR #15057, opened 2023-08-29
example: export map: add retina support

jlecordier:


jlecordier:
Here is the site side POUR : https://github.com/openlayers/openlayers.github.io/pull/132

mike-000:
It works with that example because both layers canvases use the device pixel ratio, but if you add a layer which does the transform used for that layer canvas is wrong.  e.g. a tile layer with OSM source and rotate the view with device pixel ratio 2

Screenshot

image

Export

image

ahocevar (changes requested):
Changing the transform obtained from the canvas element is not a good idea. The reason why you're not seeing a retina canvas is because `mapCanvas` is created from the size obtained from `map.getSize()`, which is in css pixels. The appropriate way to create a higher resolution `mapCanvas` is to set a bigger width and height (like you do already in this pull request), and `scale()` the existing transform instead of modifying the scale of the existing transform.

mike-000:
The final scale was correct but the OSM layer was misplaced.  You need to upscale the entire transform, not just the scale components, and using the map pixel ratio ensures a consistent ratio is used throughout.  https://codesandbox.io/s/export-map-forked-4w2l62?file=/main.js

PR #15059, opened 2023-08-29
Remove ol/style/Icon's imgSize property

ahocevar:
This pull request removes the `imgSize` property and its flat style `icon-img-size` counterpart, changing the loading sequence so they are no longer needed.

The image is now loaded using the simple `ol/Image.load()` function instead of the more involved `listenImage()` one, because the way `listenImage` decodes the image was only guaranteed to provide the decoded image for a single render frame, which does not bring a real performance benefit. We could also use `ol/Image.decode()`, but that would make SVG images, which are frequently used as icons, blurry when scaled or on retina displays.

### DIFF
--- a/src/ol/style/IconImage.js
+++ b/src/ol/style/IconImage.js
@@ -7,8 +7,8 @@ import EventType from '../events/EventType.js';
 import ImageState from '../ImageState.js';
 import {asString} from '../color.js';
 import {createCanvasContext2D} from '../dom.js';
+import {decodeFallback} from '../Image.js';
 import {shared as iconImageCache} from './IconImageCache.js';
-import {listenImage} from '../Image.js';
 
 /**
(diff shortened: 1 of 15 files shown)


src/ol/style/IconImage.js
```diff
@@ -8,7 +8,7 @@ import ImageState from '../ImageState.js';
 import {asString} from '../color.js';
 import {createCanvasContext2D} from '../dom.js';
 import {shared as iconImageCache} from './IconImageCache.js';
-import {listenImage} from '../Image.js';
+import {load} from '../Image.js';
```
mike-000, line 11: Wouldn't `decodeFallback` be slightly more efficient, while still keeping SVGs scaleable?

jahow (approved):
Tested the changes on the webgl style and it works great, thanks!

PR #15064, opened 2023-08-30
Revert "Remove lineCap, it is useless on closed paths"

MoonE:
https://github.com/openlayers/openlayers/pull/15058#issuecomment-1699898738


tschaub (approved):
See original change in #12507

MoonE:
@mike-000 Thanks, I didn't consider the canvas' default lineCap.

PR #15066, opened 2023-08-31
Recreate WebGL helper when context is lost and restored

puckipedia:
For any reason, the WebGL context might be lost, causing the map to stop rendering. This adjusts the code to set a flag indicating that any resources used have to be recreated on the WebGL context.

This fix piggybacks on the fact that recreating the helper (because of e.g. a canvas switch) recreates all resources, since different canvases may have different WebGL contexts.

### DIFF
--- a/src/ol/renderer/webgl/Layer.js
+++ b/src/ol/renderer/webgl/Layer.js
@@ -167,7 +167,11 @@ class WebGLLayerRenderer extends LayerRenderer {
       const canvasCacheKey =
         'map/' + frameState.mapId + '/group/' + groupNumber;
 
-      if (!this.helper || !this.helper.canvasCacheKeyMatches(canvasCacheKey)) {
+      if (
+        !this.helper ||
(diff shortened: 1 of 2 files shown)


src/ol/renderer/webgl/Layer.js
```diff
@@ -167,7 +167,11 @@ class WebGLLayerRenderer extends LayerRenderer {
       const canvasCacheKey =
         'map/' + frameState.mapId + '/group/' + groupNumber;
 
-      if (!this.helper || !this.helper.canvasCacheKeyMatches(canvasCacheKey)) {
+      if (
+        !this.helper ||
+        !this.helper.canvasCacheKeyMatches(canvasCacheKey) ||
+        this.helper.needsToBeRecreated()
+      ) {
```
tschaub, line 174: Doesn’t need to happen now, but it could be nice to collect these checks into a single function.

tschaub (approved):

tschaub:
Thanks again, @puckipedia

PR #15075, opened 2023-09-03
Avoid Chrome/Edge deprecated feature warning

mike-000:
Avoids the browser warning reported in  https://stackoverflow.com/questions/77031326/getting-css-overflow-error-while-working-with-openlayers-v7-5-2


ahocevar (approved):
Thanks, @mike-000

PR #15094, opened 2023-09-07
Do not attempt to decode image if `src` is not set

### LINKED ISSUE #15093: ImageWMS imageLoadFunction asynchronous loading is broken
**Describe the bug**
The imageLoadFunction of ol/source/ImageWMS was useful to load images with post requests or - in general - through XHR requests, where the src of the image element would be set to a data URI when the content is loaded. This does no longer work in version 8 of OpenLayers.

**To Reproduce**
1. Setup an ImageWMS Source with a custom imageLoadFunction.
2. Use any asynchronous code to set the image.src to a data URI.
3. This results in the WMS image not being rendered on the map. The console shows the error message: "DOMException: The source image cannot be decoded."

**Stackblitz Example**
Here is an example implementation for OpenLayers version 7 and 8:
Working V7: https://stackblitz.com/edit/vite-abvayd?file=main.js
Broken V8: https://stackblitz.com/edit/vite-gkegv7?file=main.js

**Expected behavior**
According to the current documentation of the imageLoadFunction (https://openlayers.org/en/latest/apidoc/module-ol_Image.html#~LoadFunction) it should be possible to use a custom http request to load the image and apply it as a data URI.
This setup worked in version 7 of Openlayers and broke after we updated our codebase.

mike-000:
Fixes #15093

Calling `image.decode()` if `image.src` is not set will throw an error, so we need to fall back to using `load`.

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -296,7 +296,7 @@ export function decodeFallback(image, src) {
   if (src) {
     image.src = src;
   }
-  return IMAGE_DECODE
+  return image.src && IMAGE_DECODE
     ? new Promise((resolve, reject) =>
         image.decode().then(() => resolve(image), reject)
       )
@@ -316,7 +316,7 @@ export function decode(image, src) {
   if (src) {
     image.src = src;
   }
-  return IMAGE_DECODE && CREATE_IMAGE_BITMAP
+  return image.src && IMAGE_DECODE && CREATE_IMAGE_BITMAP
     ? image.decode().then(() => createImageBitmap(image))
     : decodeFallback(image);
 }


ahocevar (commented):
This is not the right fix. Like the documentation says, `src` needs to be provided as 2nd argument when it is not set already. If `decode()`, `decodeFallback()` or `load()` get called with an image that has no `src` set and no `src` is provided as 2nd argument, the call site needs to be changed.

mike-000:
Despite the docs `load` can handle images without src because it uses `load` event listeners.  The alternative would be to check for a custom `imageLoadFunction` in `ol/source/ImageWMS`, `ImageArcGISRest` and `ImageMapGuide`
 
        load: (image, src) => {
          this.image.setImage(image);
          this.imageLoadFunction_(this.image, src);
          return this.imageLoadFunction_ === defaultImageLoadFunction ? decode(image) : load(image);
        },

ahocevar (approved):
I just saw `decode()` and friends only accept an `HTMLImageElement` (no `HTMLCanasElement` or `ImageBitmap` as input, so the change you suggest in this pull request is safe. Thanks, @mike-000.

ahocevar:
@mike-000 Maybe just update the JSDoc of `decode` and `decodeFallback` to explain what happens when a src is not provided *and* not set on the image.

PR #15095, opened 2023-09-08
More realistic canvas gradient/pattern example

mike-000:
The current Canvas Gradient example looks strange from user's perspective, the gradient appears at seemingly arbitrary positions depending on the view center and resolution and does not align consistently with the features which are sometimes all red or all purple.  Repeating patterns based on real data would make a more realistic example.

### DIFF
--- a/examples/canvas-gradient-pattern.html
+++ b/examples/canvas-gradient-pattern.html
@@ -1,11 +1,12 @@
 ---
 layout: example.html
 title: Styling feature with CanvasGradient or CanvasPattern
-shortdesc: Example showing a vector layer styled with a gradient.
+shortdesc: Example showing a vector layer styled with a gradient pattern.
 docs: >
-  This example creates a [`CanvasGradient`](https://developer.mozilla.org/en/docs/Web/API/CanvasGradient).
-  The vector data is loaded from a file and features are filled with the gradient.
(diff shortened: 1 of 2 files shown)


tschaub:
> The current Canvas Gradient example looks strange from user's perspective

I agree.  And I was hopeful when I read this.  If we can’t come up with a sensible looking example for a feature, then I think it is worth questioning the value of the feature.

Then I clicked on [the example](https://deploy-preview-15095--ol-site.netlify.app/en/latest/examples/canvas-gradient-pattern.html) linked above

image

Perhaps this is a matter of taste, but I found this example somewhat shocking.  I don’t think this qualifies as “a more realistic example.”  I can say I have never seen anything like it actually.

ahocevar:
"More realistic", in my opinion, would be something like different hatches for different attribute values, or a solid color and a hatch depending on an attribute value. Both would be a pattern rather than a gradient, but hey, that's also more realistic, no?

mike-000:
A gradient alone does not render well as it can be randomly positioned, so both options in the example are patterns containing gradients, allowing two properties of a feature to be represented in a single fill.

Hatched patterns (single color plus transparent) for a single property are supported in styles for the ArcGIS REST Feature service

image

https://www.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/RbMX0mRVOFNTdLzd/ArcGIS/rest/services/hosted_symbology_test/FeatureServer/0

and those could be replicated in OpenLayers.

PR #15102, opened 2023-09-10
Add getFeatureInfoUrl and getLegendUrl functions to ol/wms

ahocevar:
This pull request adds functions to get WMS GetFeatureInfo and GetLegendGraphic URLs without the need for an ol/source/ImageWMS (or ol/source/TileWMS) instance.

See #15093.


jahow (approved):
Looking good, thanks!

PR #15103, opened 2023-09-10
Return loaded image when decode fails

### LINKED ISSUE #15097: ImageWMS source error after upgrading from v7.5.2 to v8.0.0
I have recently tried to upgrade OpenLayers from version 7.5.2 to 8.0.0.

I face an error when OpenLayers tries to load the images from the WMS source. The `load` function in `Image.js` catches an exception in line 190:

```
Image.js:190 DOMException: The source image cannot be decoded.
```

Strange thing is when I look at the request to my GeoServer, it comes back with `200 OK`, and I even see the PNG data in the response. Without changing anything else, and going back to v7.5.2, everything works fine. I also tried v8.1.0, the result is the same.

A have carefully studied the migration notes, and found nothing that could be related to this error.

ahocevar:
This pull returns a loaded image when decode() failed.

Fixes #15097

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -298,7 +298,12 @@ export function decodeFallback(image, src) {
   }
   return image.src && IMAGE_DECODE
     ? new Promise((resolve, reject) =>
-        image.decode().then(() => resolve(image), reject)
+        image
+          .decode()
+          .then(() => resolve(image))
+          .catch((e) =>
+            image.complete && image.width ? resolve(image) : reject(e)
+          )
       )
     : load(image);
 }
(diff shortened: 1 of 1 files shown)


src/ol/Image.js
```diff
@@ -317,7 +322,14 @@ export function decode(image, src) {
     image.src = src;
   }
   return image.src && IMAGE_DECODE && CREATE_IMAGE_BITMAP
-    ? image.decode().then(() => createImageBitmap(image))
+    ? image
+        .decode()
+        .then(() => createImageBitmap(image))
+        .catch((e) =>
+          image.complete && image.width
+            ? Promise.resolve(image)
```
tschaub, line 330:
```suggestion
            ? image
```

src/ol/Image.js
```diff
@@ -317,7 +322,14 @@ export function decode(image, src) {
...
+            : Promise.reject(e)
```
tschaub, line 331:
```suggestion
            : throw e
```

tschaub (approved):

mike-000:
Does the wrapping Promise in `decodeFallback` have any purpose?  Why not

    ? image
        .decode()
        .then(() => image)
        .catch((e) => {
          if (image.complete && image.width) {
            return image;
          }
          throw e;
        })

PR #15104, opened 2023-09-10
Catch `image.decode()` errors and return image if loaded

### LINKED ISSUE #15097: ImageWMS source error after upgrading from v7.5.2 to v8.0.0

mike-000:
Fixes the regression noted in https://github.com/openlayers/openlayers/issues/15097#issuecomment-1712819423

Probably also fixes #15097

### DIFF
--- a/src/ol/Image.js
+++ b/src/ol/Image.js
@@ -311,7 +311,7 @@ export function decodeFallback(image, src) {
           return image;
         }
         listener();
-        return Promise.reject(e.message);
+        return Promise.reject(e);
       });
   }
   return load(image);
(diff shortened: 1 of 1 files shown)


mike-000:
Duplicate of #15103

PR #15113, opened 2023-09-11
Make inflateEnds work regardless of the y coordinate orientation

ahocevar:
Currently, the `inflateEnds()` function only works with coordinate systems where y coordinates increase downwards. To make it work for arbitrary y coordinate orientations, this pull request changes the function from expecting a specific winding order to remembering the winding order of the first ring to determine the start of a new polygon within a multipolygon, which will be the case when a ring has the same winding order as the first ring.


jahow (approved):
Perfect, thanks!

PR #15116, opened 2023-09-12
Support updating and tracking extra params in the link interaction

tschaub:
This makes it possible to synchronize other parts of the application state in the URL using the link interaction.  By default, the link interaction only synchronizes parts of the map state with the URL.  The `track` method can be used to add a listener for changes to other params.  And the `update` method can be called to update the URL with new values.

### DIFF
--- a/src/ol/interaction/Link.js
+++ b/src/ol/interaction/Link.js
@@ -55,6 +55,10 @@ function differentArray(a, b) {
 /** @typedef {'x'|'y'|'z'|'r'|'l'} Params */
 
 /**
+ * @typedef {function(string):void} Callback
+ */
+
+/**
  * @typedef {Object} Options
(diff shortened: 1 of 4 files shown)


ahocevar (approved):

src/ol/interaction/Link.js
```diff
@@ -342,6 +368,46 @@ class Link extends Interaction {
...
+    this.trackedValues_[key] = value;
+    return value;
+  }
+
+  /**
+   * Update the URL with a new search parameter value.  If the value is null, it will be
+   * deleted from the search parameters.
+   *
+   * @param {string} key The URL search parameter.
+   * @param {string} [value] The updated value (or null to remove it from the URL).
```
M393, line 394: This allows `undefined` and omitting the value parameter.
Should this be `{string|null} value`?

tschaub: Good catch.  Updated in the latest commit.

PR #15117, opened 2023-09-12
Make tooltip / hint  on hover example

arekgotfryd:
This pull request adds example of showing tooltip/hint with text information about hovered feature.

### DIFF
--- a/examples/tooltip-on-hover.js
+++ b/examples/tooltip-on-hover.js
@@ -18,10 +18,8 @@ const vector = new VectorLayer({
     format: new GeoJSON(),
   }),
   background: 'white',
-  style: function (feature) {
-    const color = feature.get('COLOR') || '#eeeeee';
-    style.getFill().setColor(color);
-    return style;
+  style: {
+    'fill-color': ['string', ['get', 'COLOR'], '#eeeeee'],
   },
 });
 
--- a/examples/tooltip-on-hover.js
+++ b/examples/tooltip-on-hover.js
@@ -64,4 +64,5 @@ map.on('click', function (evt) {
 
 map.getTargetElement().addEventListener('pointerleave', function () {
   currentFeature = undefined;
+  info.style.visibility = 'hidden';
 });
--- a/examples/tooltip-on-hover.js
+++ b/examples/tooltip-on-hover.js
@@ -35,12 +35,8 @@ const map = new Map({
 });
 
 const info = document.getElementById('info');
+info.style.visibility = 'hidden';
 info.style.pointerEvents = 'none';
-info.style.backgroundColor = 'red';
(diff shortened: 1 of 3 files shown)


examples/tooltip-on-hover.js
```diff
@@ -0,0 +1,77 @@
...
+  source: new VectorSource({
+    url: 'https://openlayers.org/data/vector/ecoregions.json',
+    format: new GeoJSON(),
+  }),
+  background: 'white',
+  style: function (feature) {
+    const color = feature.get('COLOR') || '#eeeeee';
+    style.getFill().setColor(color);
+    return style;
+  },
```
tschaub, line 25:
```suggestion
  style: {
    'fill-color': ['string', ['get', 'COLOR'], '#eeeeee'],
  },
```

We're moving away from using the style and symbolizer constructors.  Can you update the example to use a style like the one above and remove the call to the `Style` constructor above?

tschaub:
Thanks for this nice looking contribution, @arekgotfryd.

examples/tooltip-on-hover.js
arekgotfryd: Thanks for quick review. Suggested changes applied.

examples/tooltip-on-hover.js
```diff
@@ -0,0 +1,75 @@
...
+import Style from '../src/ol/style/Style.js';
+import VectorLayer from '../src/ol/layer/Vector.js';
+import VectorSource from '../src/ol/source/Vector.js';
+import View from '../src/ol/View.js';
+
+const style = new Style({
+  fill: new Fill({
+    color: '#eeeeee',
+  }),
+});
```
tschaub, line 13: This will need to go away too.  And the `import`s for `Fill` and `Style` above.

arekgotfryd: Ah sorry for that. Done.

examples/tooltip-on-hover.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+  } else {
+    info.style.visibility = 'hidden';
+  }
+  currentFeature = feature;
+};
+
+map.on('pointermove', function (evt) {
+  if (evt.dragging) {
+    info.style.display = 'none';
```
tschaub, line 53:
```suggestion
    info.style.visibility = 'hidden';
```

It looks like if you drag the map, the tooltip is not shown again on hover.  I think you'll either want to use `display` or `visibility` consistently.

arekgotfryd: Switched to visibility for consistency.

tschaub (approved):

tschaub:
Thanks for the work on this, @arekgotfryd.  Looks great.

Issue #14811, opened 2023-06-07
useGeographic() and WebGLVectorLayerRenderer are not fully compatible


vitalus:
Simply,

        const layer = new WebGLLayer({
            source: new VectorSource({
              url...
              format: new GeoJSON(),
            })
          }

is working, but

        const layer: WebGLLayer = new WebGLLayer({
            source: new VectorSource({
                features
            })
        },
        style);

is not.

Tested  with same JSON in EPSG:4326, loaded by OL (first case) and preloaded and passed as Feature[] (second case).
In first case data comes to the renderer in EPSG:3857 (internal reprojection is done when data is loaded), in second case it comes in EPSG:4326 and rendering does not work at all.

mike-000:
Also applies to WebGLPointsLayer.  The first case is incorrect as the feature geometry should be in the user projection, not the view projection - it is only working because of the issue in the second case.  Both WebGLVectorLayer and WebGLPointsLayer are still experimental, so full support might not be expected.

But the Heatmap layer which uses WebGLPointsLayer is not experimental and therefore would be expected to support features in a user projection - but it does not https://codesandbox.io/s/heatmap-earthquakes-forked-es3mvf?file=/main.js

tschaub:
After #15118, you can call `useGeographic()` and the WebGL vector renderer will transform geometries with geographic coordinates before rendering.

PR #15118, opened 2023-09-12
Add support for user projections to WebGL rendering

tschaub:
This makes it possible to transform vector data to the user projection (set by calling `useGeographic()` for example) before rendering in WebGL.


jahow (approved):
Thanks! That should fix https://github.com/openlayers/openlayers/issues/14811 as well

PR #15120, opened 2023-09-12
Add hit detection support to WebGL vector layer

gberaudo:
Add hit detection to the WebGL vector layer renderer.
This is similar to the points layer:
- when hit detection is enabled (the default), compute a color for each feature and add it to the vertex buffer;
- render the frame twice: for the main framebuffer and for the hit detection buffer;
- use the hit detection buffer to implement the `forEachFeatureAtCoordinate` method.

### DIFF
--- a/examples/webgl-vector-layer.js
+++ b/examples/webgl-vector-layer.js
@@ -42,3 +42,48 @@ const map = new Map({
     zoom: 1,
   }),
 });
+
+const featureOverlay = new WebGLLayer({
+  source: new VectorSource(),
+  map: map,
+  style: {
+    'stroke-color': 'rgba(255, 255, 255, 0.7)',
+    'stroke-width': 2,
+  },
+});
+
+let highlight;
+const displayFeatureInfo = function (pixel) {
+  const feature = map.forEachFeatureAtPixel(pixel, function (feature) {
+    return feature;
+  });
+
+  const info = document.getElementById('info');
+  if (feature) {
+    info.innerHTML = feature.get('ECO_NAME') || '&nbsp;';
+  } else {
+    info.innerHTML = '&nbsp;';
+  }
+
+  if (feature !== highlight) {
+    if (highlight) {
+      featureOverlay.getSource().removeFeature(highlight);
+    }
+    if (feature) {
+      featureOverlay.getSource().addFeature(feature);
+    }
+    highlight = feature;
+  }
+};
+
+map.on('pointermove', function (evt) {
+  if (evt.dragging) {
+    return;
+  }
+  const pixel = map.getEventPixel(evt.originalEvent);
+  displayFeatureInfo(pixel);
+});
+
+map.on('click', function (evt) {
+  displayFeatureInfo(evt.pixel);
+});
--- a/src/ol/render/webgl/MixedGeometryBatch.js
+++ b/src/ol/render/webgl/MixedGeometryBatch.js
@@ -6,7 +6,7 @@ import {getUid} from '../../util.js';
 import {inflateEnds} from '../../geom/flat/orient.js';
 
 /**
- * @typedef {import("../../Feature").default} Feature
+ * @typedef {import("../../Feature.js").default} Feature
  */
 /**
  * @typedef {import("../../geom/Geometry.js").Type} GeometryType
@@ -19,6 +19,7 @@ import {inflateEnds} from '../../geom/flat/orient.js';
  * @property {number} [verticesCount] Only defined for linestring and polygon batches
  * @property {number} [ringsCount] Only defined for polygon batches
  * @property {Array<Array<number>>} [ringsVerticesCounts] Array of vertices counts in each ring for each geometry; only defined for polygons batches
+ * @property {number} [ref] The reference in the global batch (used for hit detection)
  */
 
 /**
@@ -70,6 +71,29 @@ import {inflateEnds} from '../../geom/flat/orient.js';
  */
 class MixedGeometryBatch {
   constructor() {
+    this.globalCounter_ = 0;
+    /**
+     * Refs are used as keys for hit detection.
+     * @type {Map<number, Feature|RenderFeature>}
+     * @private
+     */
+    this.refToFeature_ = new Map();
+
+    /**
+     * Features are split in "entries", which are individual geometries. We use the following map to share a single ref for all those entries.
+     * @type {Map<string, number>}
+     * @private
+     */
+    this.uidToRef_ = new Map();
+
+    /**
+     * The precision in WebGL shaders is limited.
+     * To keep the refs as small as possible we maintain an array of returned references.
+     * @type {Array<number>}
+     * @private
+     */
+    this.freeGlobalRef_ = [];
+
     /**
      * @type {PolygonGeometryBatch}
      */
@@ -126,6 +150,7 @@ class MixedGeometryBatch {
 
   /**
    * @param {Feature|RenderFeature} feature Feature
+   * @return {GeometryBatchItem|void} the cleared entry
    * @private
    */
   clearFeatureEntryInPointBatch_(feature) {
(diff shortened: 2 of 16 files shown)


examples/webgl-vector-layer.js
```diff
@@ -18,6 +18,7 @@ const style = {
 class WebGLLayer extends Layer {
   createRenderer() {
     return new WebGLVectorLayerRenderer(this, {
+      // disableHitDetection: true,
```
tschaub, line 21: Is this intentional?

examples/webgl-vector-layer.js
```diff
@@ -42,3 +43,20 @@ const map = new Map({
...
 });
+
+map.on('pointermove', function (evt) {
+  if (evt.dragging) {
+    return;
+  }
+  map.forEachFeatureAtPixel(
+    evt.pixel,
+    function (feature) {
+      window.console.log(feature);
```
tschaub, line 54: It would be nice to have this update the page with some detail about the feature (like the [Canvas vector layer example](https://openlayers.org/en/latest/examples/vector-layer.html)) rather than logging.

src/ol/render/webgl/MixedGeometryBatch.js
```diff
@@ -70,6 +71,26 @@ import {inflateEnds} from '../../geom/flat/orient.js';
  */
 class MixedGeometryBatch {
   constructor() {
+    this.globalCounter = 0;
+    /**
+     * Refs are used as keys for hit detection.
+     * @type {Map<number, Feature|RenderFeature>}
```
tschaub, line 77: It looks like this and the properties below could be `@private` so other parts of the library don't mistakenly think they can mess with them.

src/ol/renderer/webgl/VectorTileLayer.js
```diff
@@ -98,7 +98,7 @@ class WebGLVectorTileLayerRenderer extends WebGLBaseTileLayerRenderer {
    */
   createRenderers_() {
     this.styleRenderers_ = this.styles_.map(
-      (style) => new VectorStyleRenderer(style, this.helper)
+      (style) => new VectorStyleRenderer(style, this.helper, false) // FIXME: implement
```
tschaub, line 101: I think these `FIXME` are a bit mysterious (I can't tell if they were meant to be addressed during development of this feature or if they are instructions to others).  If this is not going to be implemented in this branch, could you please remove the comment (or a more detailed comment might be easier for future readers to interpret - maybe "when we get around to implementing hit detection for the vector tile renderer, we will pass along an option from the constructor..." or something).

src/ol/renderer/webgl/worldUtil.js
```diff
@@ -0,0 +1,29 @@
+import {getWidth} from '../../extent.js';
+
+/**
+ * Compute world params
+ * @private
```
tschaub, line 5: I'm unclear on what the impact of a `@private` annotation on an exported function is.  Does this mean other modules cannot or should not import?

src/ol/render/webgl/MixedGeometryBatch.js
```diff
@@ -412,13 +465,44 @@ class MixedGeometryBatch {
...
 
+  /**
+   * @param {string} featureUid Feature uid
+   * @param {GeometryBatchItem} entry The entry to add
+   * @return {GeometryBatchItem} the added entry
+   * @private
+   */
+  addRefToEntry_(featureUid, entry) {
+    const currentRef = this.uidToRef.get(featureUid);
+    const ref = currentRef || this.freeGlobalRef.pop() || ++this.globalCounter;
```
jahow, line 476:
```suggestion
    const ref = currentRef || this.freeGlobalRef.pop() || this.globalCounter++;
```

Otherwise refs are 1-based, isn't that a little counter-intuitive?

src/ol/renderer/webgl/VectorLayer.js
```diff
@@ -409,6 +455,28 @@ class WebGLVectorLayerRenderer extends WebGLLayerRenderer {
     callback,
     matches
   ) {
+    assert(
+      this.hitDetectionEnabled_,
+      '`forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has not been enabled. This is done by providing adequate shaders using the `hitVertexShader` and `hitFragmentShader` properties of `WebGLPointsLayerRenderer`'
```
jahow, line 460:
```suggestion
      '`forEachFeatureAtCoordinate` cannot be used on a WebGL layer if the hit detection logic has been disabled using the `disableHitDetection: true` option.'
```

src/ol/renderer/webgl/VectorLayer.js
```diff
@@ -409,6 +455,28 @@ class WebGLVectorLayerRenderer extends WebGLLayerRenderer {
...
+    );
+    // FIXME: properly check that we have some buffer
+    if (!this.styleRenderers_.length || !this.hitDetectionEnabled_) {
+      return undefined;
+    }
+
+    // showDebugCanvas(this.hitRenderTarget_);
```
jahow, line 467: to delete

jahow (approved):

jahow (commented):
Also for firefox another workaround could be:

```diff
--- a/src/ol/render/webgl/VectorStyleRenderer.js
+++ b/src/ol/render/webgl/VectorStyleRenderer.js
@@ -504,6 +504,9 @@ class VectorStyleRenderer {
     frameState,
     preRenderCallback
   ) {
+    if (indicesBuffer.getArray().length === 0) {
+      return;
+    }
     this.helper_.useProgram(program, frameState);
     this.helper_.bindBuffer(verticesBuffer);
     this.helper_.bindBuffer(indicesBuffer);
```

Here we're simply skipping rendering if no primitives are found in the buffer. This way we don't add complexity to the ShaderBuilder API.

src/ol/render/webgl/MixedGeometryBatch.js
gberaudo: As discussed, starting from 1 allows to distinguish from the white color. I am adding a comment.

gberaudo:
> Also for firefox another workaround could be:
> 
> ```diff
> --- a/src/ol/render/webgl/VectorStyleRenderer.js
> +++ b/src/ol/render/webgl/VectorStyleRenderer.js
> @@ -504,6 +504,9 @@ class VectorStyleRenderer {
>      frameState,
>      preRenderCallback
>    ) {
> +    if (indicesBuffer.getArray().length === 0) {
> +      return;
> +    }
>      this.helper_.useProgram(program, frameState);
>      this.helper_.bindBuffer(verticesBuffer);
>      this.helper_.bindBuffer(indicesBuffer);
> ```
> 
> Here we're simply skipping rendering if no primitives are found in the buffer. This way we don't add complexity to the ShaderBuilder API.

Thanks, this looks like the proper thing to do.
I commited it and updated the tests (it was actually missing a few features in order to render correctly).

jahow (approved):
This looks perfect now, thanks a lot!!

guldbran:
Excellent news, this is a crucial feature for our project @gberaudo 🙌🏻. Looking forward to 8.2 now @tschaub - keep up the good work!

tschaub:
@gberaudo - sorry to be the bearer of bad news, but I'm not seeing any feature info displayed on this example https://deploy-preview-15120--ol-site.netlify.app/en/latest/examples/webgl-vector-layer.html on Chrome 116 or Safari 16.4 on a Mac (macOS 13, M1 chip).  Has anybody else seen it working on a Mac?

ahocevar:
Unfortunately I'm getting the same results as @tschaub .

jahow:
Could reproduce as well on iOS systems only, no errors in the console. The hit detection render target looks completely empty. I suspect this line to make things behave inconsistently: https://github.com/openlayers/openlayers/blob/5075214a1669407ab90c9e4f722b54768d9f7d3b/src/ol/render/webgl/renderinstructions.js#L24

Here the `value` equals typically `[0, 0, 0, 0.12345]` for a hit color with a low id, so `value[0]` is falsy. This line then assigns the whole `value` to a slot in the `Float32Array`, which results in a `NaN` value. The fact that it works on certain systems is actually surprising. @gberaudo I hope that helps!

PR #15128, opened 2023-09-14
Named color parsing

tschaub:
This change updates the `ol/color.js` module so it doesn't depend on the DOM.

I added a dependency on the `color-name` package.  This is [a simple lookup](https://github.com/colorjs/color-name/blob/38ff46b4a0742bbc0c467ec1405a36af2fc4a0e6/index.js) of CSS named colors to RGB values.  We could include this lookup directly in the `ol/color.js` module if we don't want the dependency.  But it may eventually be useful to depend on https://github.com/colorjs/color-parse or other packages that depend on the same.

We didn't previously have tests that checked that `fromString()` worked with named colors.  I added tests to cover this and moved all color tests to demonstrate that they pass without the DOM.

### DIFF
--- a/src/ol/color.js
+++ b/src/ol/color.js
@@ -1,6 +1,7 @@
 /**
  * @module ol/color
  */
+import colorNames from 'color-name';
 import {clamp} from './math.js';
 
 /**
@@ -42,20 +43,20 @@ export function asString(color) {
 }
 
 /**
- * Return named color as an rgba string.
+ * Return named color as an rgb(a) string.
  * @param {string} color Named color.
- * @return {string} Rgb string.
+ * @return {string} RGB(A) string.
  */
 function fromNamed(color) {
-  const el = document.createElement('div');
-  el.style.color = color;
-  if (el.style.color !== '') {
-    document.body.appendChild(el);
-    const rgb = getComputedStyle(el).color;
-    document.body.removeChild(el);
-    return rgb;
+  const name = color.toLowerCase();
+  if (!colorNames.hasOwnProperty(name)) {
+    if (name === 'transparent') {
+      return 'rgba(0,0,0,0)';
+    }
+    return '';
   }
-  return '';
+  const rgb = colorNames[name];
+  return 'rgb(' + rgb[0] + ',' + rgb[1] + ',' + rgb[2] + ')';
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/ol/color.js
```diff
@@ -42,20 +43,20 @@ export function asString(color) {
...
-    document.body.appendChild(el);
-    const rgb = getComputedStyle(el).color;
-    document.body.removeChild(el);
-    return rgb;
+  const name = color.toLowerCase();
+  if (!colorNames.hasOwnProperty(name)) {
+    if (name === 'transparent') {
+      return 'rgba(0,0,0,0)';
+    }
+    return '';
```
simonseyock, line 56: is this a good default value? It is a string but no valid rgb(a) string. most likely i would suspect this should not cause an error, because normally if colors are allowed, an empty strings is also ok.

Also this technically differs from the previous implementation because the browsers would actually accept any color name and just parse it randomly. I think this can be neglected as it is an unspecified case - also browsers handle this differently.

test/node/ol/color.test.js
```diff
@@ -1,85 +1,116 @@
...
+        ['oops', 'Invalid color'],
+      ];
+      for (const c of cases) {
+        it(`works for ${c[0]}`, () => {
+          const expected = c[1];
+          if (typeof expected === 'string') {
+            expect(() => {
+              fromString(c[0]);
+            }).to.throwException(expected);
+            return;
```
simonseyock, line 82: Why does this throw an exception? I would expect an empty string.

simonseyock (approved):
had some questions but answered them myself :D looks good to me!

src/ol/color.js
tschaub: The previous behavior in this case was not really predictable.  If you `el.style.color = 'oops'`, then `getComputedStyle(el).color` depends on what other CSS rules might apply.  For example, on this page, that returns `rgb(31, 35, 40)`.  On `about:blank`, it returns `'rgb(0, 0, 0)'`.  I thought that I previously saw a case where the result was `''`, but cannot reproduce that now.

I think that the new behavior with the exported methods is probably more "correct" now.  Calling `asArray('oops')` now throws `new Error('Invalid color')` whereas previously I think the behavior would depend on CSS rules that might be on the page.

tschaub:
Good comments.  Thanks for the review.  Maybe we will want to revisit the behavior of `fromNamed` in the future (perhaps it should throw on invalid colors instead of returning `''`).  But for now I think the exported methods behave as we want them to.

PR #15129, opened 2023-09-14
Debugging a rendering test in CI

tschaub:
Using this branch to see if we can get the rendering test from #15127 to pass in CI.


jahow:
Interestingly the rendering test failure was caused by using `RenderFeature`s in the `tileLoadFunction`. Switching to `Feature` made the test pass right away.

This problem wasn't happening locally on two different computers. Closing this, thanks for the assistance @tschaub :)

PR #15121, opened 2023-09-12
Add a GLSL compiler for expressions

jahow:
This PR adds the possibility to compile expressions for a GPU target; the end result is then GLSL code snippets, that can then be fed into the `ShaderBuilder` class.

This was mostly a large refactoring to move the logic and tests from the old `ol/style/expressions.js` module to the new `ol/expr/gpu.js` one.

This also contains several improvements to expressions parsing, including:
* Better handling of type hints
* Parsing of the `in` operator
* Added the `band` and `palette` operators
* Better test coverage by bringing many tests from `ol/style/expressions.js`

Also addresses https://github.com/openlayers/openlayers/issues/15099

Note that with this PR *everything should work exactly like before*.

PR #15130, opened 2023-09-14
New expression operators, allow chain of args validator

tschaub:
Introduced the following operators:

*    geometry-type
*    between
*    interpolate
*    case
*    in
*    array
*    color

Only parsers have been introduced, no compiling yet.

Also add support for type hint when parsing expressions

---

Note that this is a cherry-picked version of a commit from #15121, rebased on the changes in #15128 so that tests pass.

Since this is @jahow's commit, I'll review and merge this (if I approve) on this pull request instead of waiting for #15121 to be complete (since I want to base another branch of work on this commit).


tschaub (commented):
I approve

PR #15131, opened 2023-09-14
Fix the type for VectorTile features

ahocevar:
This is a small fix for the type of the features used by `ol/VectorTile`. I'm surprised this went unnoticed for so long.


jahow (approved):
Thanks!

PR #15135, opened 2023-09-14
Parse hsl/hsla colors

mike-000:
#15128 has caused a regression of #10195

Although HSL is relatively rare some Mapbox styles e.g. https://github.com/OrdnanceSurvey/OS-Open-Zoomstack-Stylesheets/blob/master/Vector%20Tiles/Mapbox%20GL%20Styles/OS%20Open%20Zoomstack%20-%20Outdoor.json use it heavily, so it would now be more efficient to parse instead of reverting to using `getComputedStyle`.  The algorithm is based on the older longer version in https://stackoverflow.com/questions/36721830/convert-hsl-to-rgb-and-hex which gives the same result as `getComputedStyle` in Chrome, which the new shorter version does not.

### DIFF
--- a/src/ol/color.js
+++ b/src/ol/color.js
@@ -27,7 +27,7 @@ const HEX_COLOR_RE_ = /^#([a-f0-9]{3}|[a-f0-9]{4}(?:[a-f0-9]{2}){0,2})$/i;
  * @type {RegExp}
  * @private
  */
-const NAMED_COLOR_RE_ = /^([a-z]*)$|^hsla?\(.*\)$/i;
+const NAMED_COLOR_RE_ = /^([a-z]*)$/i;
 
 /**
  * Return the color as an rgba string.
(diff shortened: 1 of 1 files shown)


jahow (commented):
This looks quite good, thanks! I think we'll need to add some test cases covering this to avoid further similar regressions.

tschaub:
Thanks, @mike-000.  We should have requested a test in #10195.  Without one, we can expect more regressions in the future (anything that is not tested should not really be trusted).

src/ol/color.js
```diff
@@ -126,6 +126,43 @@ export function asArray(color) {
...
+  function hueToRgb(p, q, t) {
+    t = t < 0 ? t + 1 : t > 1 ? t - 1 : t;
+    return t < 1 / 6
+      ? p + (q - p) * 6 * t
+      : t < 1 / 2
+      ? q
+      : t < 2 / 3
+      ? p + (q - p) * (2 / 3 - t) * 6
+      : p;
+  }
```
tschaub, line 149: The `hueToRgb` function could be moved to the module scope instead of being declared each time the `hslaToRgba` function is run.

PR #15136, opened 2023-09-14
Switch to WebGL 2.0

gberaudo:
As discussed during the OL code sprint in Munich, we would like to move our code base to WebGL2.

This will allow us to write cleaner and more performant code, thanks to the new rendering techniques introduced in this new version.

Closes: #14377
Closes: #14366


mike-000:
Without a fallback as in #14366/#14377 this will break all existing WebGL functionality on Chrome and Edge on many Windows 10 systems (especially those which have been upgraded from Windows 7).

jahow:
@mike-000 could you please indicate where you get that data from? https://caniuse.com/webgl2 shows all green for chrome and edge (albeit with a slightly lower compatibility of 93% instead of 95% for webgl1).

mike-000:
@jahow As in https://github.com/openlayers/openlayers/pull/14366#issuecomment-1344685794 it is not a problem of browser support, it is Chrome and Edge will not work with older hardware (even though Firefox will work with it) so support for a WebGL fallback would be needed until October 2025 (I presume all Windows 11 compatible systems will be OK).  Both of my ex Windows 7 systems have the same issue, while an ex Windows 8 system has no problem.

image

tschaub:
When we want to move to WebGL 2, the Canvas renderer provides an alternative for people who want to support older hardware.

mike-000:
There is no problem requiring WebGL2 for new functionality which is currently experimental.  Supporting only WebGL2 will break existing API functionality such as WebGL tile and Heatmap on the default browsers of many Windows 10 systems which still have 2 years support.  The only alternative to ensure compatibility will be to not upgrade to newer OpenLayers versions or switch to alternative platforms.

malaretv:
With Windows 10 support ended, maybe worth revisiting this now?

Are there other concerns with WebGL 2?

M393:
Windows 10 is still widely used, and covered by extended support which was/is basically free in the first year.
https://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide
https://www.engadget.com/computing/microsoft-has-ended-windows-10-support-but-heres-how-to-get-an-extra-year-for-free-125118044.html

PR #15137, opened 2023-09-15
Parse more variety of color strings

tschaub:
This removes the custom color parsing logic in `ol/color.js` and replaces it with color parsing from https://github.com/colorjs/color-parse.

Previously, our color parsing didn't support `rgb(255 122 127 / 80%)` and similar new syntaxes.  This change brings support for that.

Closes #15135.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "url": "https://opencollective.com/openlayers"
   },
   "dependencies": {
-    "color-name": "^2.0.0",
+    "color-rgba": "^2.4.0",
     "earcut": "^2.2.3",
     "geotiff": "^2.0.7",
     "pbf": "3.2.1",
--- a/test/rendering/cases/webgl-vectortile-masking/main.js
+++ b/test/rendering/cases/webgl-vectortile-masking/main.js
@@ -46,7 +46,7 @@ class WebGLVectorTileLayer extends VectorTile {
       className: this.getClassName(),
       style: {
         'fill-color': '#eee',
-        'stroke-color': 'rgb(136,136,136, 0.5)',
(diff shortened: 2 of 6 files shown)


test/rendering/cases/webgl-vectortile-masking/main.js
```diff
@@ -46,7 +46,7 @@ class WebGLVectorTileLayer extends VectorTile {
       className: this.getClassName(),
       style: {
         'fill-color': '#eee',
-        'stroke-color': 'rgb(136,136,136, 0.5)',
+        'stroke-color': 'rgba(136,136,136, 0.5)',
```
tschaub, line 49: This looks like a typo in the rendering test.  The previous parser parsed this as an `rgba()` color.  The new parser ignores the fourth value in the `rgb()` list.  Changing the test to use `rgba()` matches the expected output (where the stroke has an alpha value applied).

jahow: Thanks, good catch!

tschaub: There was one other case that I updated as well.  Rendering tests are passing now.  Let me know if you approve.  I'd like to base some color interpolation work on this.

package.json
```diff
@@ -44,7 +44,7 @@
     "url": "https://opencollective.com/openlayers"
   },
   "dependencies": {
-    "color-name": "^2.0.0",
+    "color-rgba": "^2.4.0",
```
tschaub, line 47: The `color-rgba` package depends on `color-name`.  So the behavior with named colors is retained from #15128..

ahocevar (approved):

PR #15141, opened 2023-09-16
Interpolation support for Canvas styles

tschaub:
This adds support for using the `interpolate` operator with numbers and colors for flat styles on the Canvas renderer.

The new [populated places example](https://deploy-preview-15141--ol-site.netlify.app/en/latest/examples/populated-places.html) shows how `interpolate` can be used to create a color ramp and scale points.

PR #15142, opened 2023-09-16
Throw an error when building flat styles that result in an empty style

tschaub:
Currently, if you make a typo in a flat style or fail to include a required property for some symbolizers (e.g. a circle requires `circle-radius`), either nothing renders or you get an obscure error from the renderer (`Uncaught TypeError: Cannot read properties of null (reading 'getGeometryFunction')`).

This change makes it so if you pass a flat style that results in a style with no stroke, fill, point, or label symbolizers, an error is thrown.

I pulled this out of #15141 (it was discovered there while trying to make the example work).

### DIFF
--- a/src/ol/render/canvas/style.js
+++ b/src/ol/render/canvas/style.js
@@ -18,6 +18,7 @@ import {
   newParsingContext,
 } from '../../expr/expression.js';
 import {buildExpression, newEvaluationContext} from '../../expr/cpu.js';
+import {isEmpty} from '../../obj.js';
 import {toSize} from '../../size.js';
 
 /**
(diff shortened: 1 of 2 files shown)


mike-000:
The docs also need some revision, e.g. for FlatIcon

> Icon style properties applied to point features. One of `icon-src` or `icon-img` must be provided to render points with an icon.

as `icon-img` was removed in version 8  `icon-src` is always required

Also what about Circles or Shapes which have neither fill or stroke, or Shapes without at least one radius?

tschaub:
> The docs also need some revision, e.g. for FlatIcon
> 
> > Icon style properties applied to point features. One of `icon-src` or `icon-img` must be provided to render points with an icon.
> 
> as `icon-img` was removed in version 8 `icon-src` is always required
> 
> Also what about Circles or Shapes which have neither fill or stroke, or Shapes without at least one radius?

These all sound like good improvements for a separate pull request.  This pull request is just about `Style` instances without a stroke, fill, image, or text symbolizer.

PR #15143, opened 2023-09-16
Properly hit detect circles after updating size

tschaub:
After calling `circle.setRadius()` with a new radius, circles are no longer properly hit detected.  This is because we keep around the old hit detection image.

This change makes it so we clear the hit detection image in the `render` call ~~if the size doesn't match the previous size~~.  The new test demonstrates that the hit detection image is properly updated when requested after a `setRadius` call.

I pulled this out of #15141 after discovering it while trying to make that example.

### DIFF
--- a/src/ol/style/RegularShape.js
+++ b/src/ol/style/RegularShape.js
@@ -505,6 +505,7 @@ class RegularShape extends ImageStyle {
     this.renderOptions_ = this.createRenderOptions();
     const size = this.renderOptions_.size;
     this.canvas_ = {};
+    this.hitDetectionCanvas_ = null;
     this.size_ = [size, size];
   }
 
--- a/test/browser/spec/ol/style/Circle.test.js
+++ b/test/browser/spec/ol/style/Circle.test.js
@@ -2,7 +2,7 @@ import CircleStyle from '../../../../../src/ol/style/Circle.js';
 import Fill from '../../../../../src/ol/style/Fill.js';
 import Stroke from '../../../../../src/ol/style/Stroke.js';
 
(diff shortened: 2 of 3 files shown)


MoonE:
This should also be done when the fill or stroke style is changed.
Might be best to reset the hit detection canvas unconditionally?
For simplicity move the setRadius function to RegularShape (and add a setRadius2 function?

tschaub:
> This should also be done when the fill or stroke style is changed. Might be best to reset the hit detection canvas unconditionally? For simplicity move the setRadius function to RegularShape (and add a setRadius2 function?

I considered adding an arg to `render` that would let callers decide if the hit detection image was invalid.  That would let `setStroke` compare width or maybe `setFill` compare null/non-null changes.  But I didn't really want to write the hit detection tests to validate this all worked as expected and was really only facing a change due to circle size, so I opted for a fix of just that issue.

But if I find more time/energy, I may push another change.  If anybody else wants to push another commit that fixes other related issues, feel free.

tschaub:
> This should also be done when the fill or stroke style is changed. Might be best to reset the hit detection canvas unconditionally?

The latest commit unconditionally removes the hit detection canvas in the `render` call (this method should be renamed).

> For simplicity move the setRadius function to RegularShape (and add a setRadius2 function?

I'm going to leave this as an improvement for hit detecting circles.  This also improves the situation for hit detecting all regular shapes in cases where stroke or fill changes should change the hit detection behavior.  I'll leave it for a separate change to make it possible to update the radius for all regular shapes.

MoonE (approved):
Thanks, @tschaub

PR #15164, opened 2023-09-22
Add properties to WebGLTileLayer options type

sebakerckhof:
Properties was missing from the WebGLTileLayer Options type


tschaub (approved):
Thanks, @sebakerckhof

PR #15165, opened 2023-09-22
Create new loader if view projection changes

### LINKED ISSUE #15161: Image WMS does not consider projection after view changed
**Describe the bug**
After view changed (with a new projection), an ImageWMS source generate request with the old projection

**To Reproduce**
Instantiate a map, with a view with EPSG:4326 projection.
Add a Image layer with a ImageWMS source to the map
Instantiate a new view with EPSG:3857 projection
Zoom/pan on the map, and check the layer's request

**Expected behavior**
the layer has to generate a request with the EPSG:3857 projection

This is a regression of the v8.0.0

mike-000:
Fixes #15161

### DIFF
--- a/src/ol/source/ImageArcGISRest.js
+++ b/src/ol/source/ImageArcGISRest.js
@@ -118,6 +118,12 @@ class ImageArcGISRest extends ImageSource {
      * @type {number}
      */
     this.ratio_ = options.ratio !== undefined ? options.ratio : 1.5;
+
+    /**
+     * @private
+     * @type {import("../proj/Projection.js").default}
+     */
+    this.loaderProjection_ = null;
   }
 
   /**
@@ -141,8 +147,9 @@ class ImageArcGISRest extends ImageSource {
     if (this.url_ === undefined) {
       return null;
     }
-    if (!this.loader) {
+    if (!this.loader || this.loaderProjection_ !== projection) {
       // Lazily create loader to pick up the view projection and to allow `params` updates
+      this.loaderProjection_ = projection;
       this.loader = createLoader({
         crossOrigin: this.crossOrigin_,
         params: this.params_,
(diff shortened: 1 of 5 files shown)


src/ol/source/ImageArcGISRest.js
```diff
@@ -141,8 +147,9 @@ class ImageArcGISRest extends ImageSource {
     if (this.url_ === undefined) {
       return null;
     }
-    if (!this.loader) {
+    if (!this.loader || this.loaderProjection_ !== projection) {
```
ahocevar, line 150: The `equivalent()` function should be used here

src/ol/source/ImageMapGuide.js
```diff
@@ -141,8 +147,9 @@ class ImageMapGuide extends ImageSource {
     if (this.url_ === undefined) {
       return null;
     }
-    if (!this.loader) {
+    if (!this.loader || this.loaderProjection_ !== projection) {
```
ahocevar, line 150: Same

src/ol/source/ImageWMS.js
```diff
@@ -195,8 +201,9 @@ class ImageWMS extends ImageSource {
     if (this.url_ === undefined) {
       return null;
     }
-    if (!this.loader) {
+    if (!this.loader || this.loaderProjection_ !== projection) {
```
ahocevar, line 204: Same

ahocevar (changes requested):

src/ol/source/ImageArcGISRest.js
mike-000: The code replicates the behavior before version 8 - if the view projection was EPSG:3857 the image was requested for CRS EPSG:3857, if the view projection was changed to EPSG:900913 the image was requested for CRS EPSG:900913.  Ditto for EPSG:4326/CRS:84, etc.

ahocevar (approved):

ahocevar:
Obviously a test would be good to avoid future regressions on this.

ahocevar (approved):

ahocevar:
Thanks, @mike-000

KyllianGautier:
Thanks you very much guys ❤️

PR #15169, opened 2023-09-24
RegularShape cache

### LINKED ISSUE #15147: Icon cache for RegularShapes
With the new style expressions, should there be some kind of IconCache for RegularShapes?

An example like https://openlayers.org/en/latest/examples/regularshape.html that is [implemented with style expressions](https://github.com/MoonE/openlayers/commit/f96b1a4b5789324542d1cfd0e288c36ed0e6b9fc) will re-render the RegularShape canvas for every drawn feature.

Even if the RegularShape is only re-rendered when the resulting style actually changed that will not help  much when the features are in a random order.

MoonE:
Fixes #15147
Missing some test, but seems to be working ...

- Caching gradients and patterns is done by putting those in a WeakMap with an id (could also be a getUid call).
- There's an inconsistency for the declutter mode properties. For icons it was `'declutter'`, for shapes `'declutter-mode'`


ahocevar:
@MoonE Thanks for your work on this! Is there anything I can help with to get this ready for review?

PR #15179, opened 2023-09-25
Make rules documentation example syntax valid

mike-000:
The rules example script in https://openlayers.org/en/latest/apidoc/module-ol_style_flat.html has syntax errors and would not work, so change it to something usable as in  https://codesandbox.io/s/webgl-points-layer-forked-9s97tl?file=/main.js


tschaub (approved):
Thanks, @mike-000

PR #15181, opened 2023-09-25
Strict type checks for select files

MoonE:
When a file passes  strict type checking it can be removed from the exclude list in `config/tsconfig-strict.json` to avoid regressing.


tschaub (approved):
Fun!

PR #15188, opened 2023-09-27
Cluster example duplicate if condition removed

### LINKED ISSUE #15187: Incorrect Conditional Statements in Dynamic Clusters Example of OpenLayers v6.15.1
**Describe the bug**
In the OpenLayers v6.15.1 example titled "Dynamic clusters" there is an error in line 265 where the if conditions are incorrectly written as getWidth && getWidth. The correct conditions, as per the new documentation, should be getWidth && getHeight. This issue affects the functionality and readability of the example

**To Reproduce**
1. Go to the OpenLayers v6.15.1 example page: https://openlayers.org/en/v6.15.1/examples/clusters-dynamic.html
2. Navigate to the source code or use the search functionality to find line no 265 in the code. Check the if condition its repeated 
    same thing twice  (getWidth(extent) < resolution && getWidth(extent) < resolution)

**Expected behavior**
The if conditions in line 265 of the code should be correctly written as
 if (view.getZoom() === view.getMaxZoom() || (getWidth(extent) < resolution && getHeight(extent) < resolution))  instead of 
 (getWidth(extent) < resolution && getWidth(extent) < resolution)  in accordance with the updated documentation. This change will ensure that the code functions as intended and is consistent with the latest best practices.

ShanteshSindgi:
Fixes #15187  The intention was to check for width and height of the extent of the cluster in v6.15.

### DIFF
--- a/examples/clusters-dynamic.js
+++ b/examples/clusters-dynamic.js
@@ -13,7 +13,7 @@ import {
 import {Cluster, Vector as VectorSource, XYZ} from '../src/ol/source.js';
 import {LineString, Point, Polygon} from '../src/ol/geom.js';
 import {Tile as TileLayer, Vector as VectorLayer} from '../src/ol/layer.js';
-import {createEmpty, extend, getWidth} from '../src/ol/extent.js';
+import {createEmpty, extend, getHeight, getWidth} from '../src/ol/extent.js';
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @ShanteshSindgi

PR #15189, opened 2023-09-27
More strict type checks

MoonE:


src/ol/tilegrid/TileGrid.js
```diff
@@ -262,8 +265,8 @@ class TileGrid {
     }
     while (z >= this.minZoom) {
       if (this.zoomFactor_ === 2) {
-        x = Math.floor(x / 2);
-        y = Math.floor(y / 2);
+        x = Math.floor(/** @type {number} */ (x) / 2);
+        y = Math.floor(/** @type {number} */ (y) / 2);
```
ahocevar, line 269: I'm wondering if these type casts could be avoided when the wrapping if clause is changed from
```js
if (this.zoomFactor_ === 2) {
```
to
```js
if (x !== undefined && y !== undefined) {
```

MoonE: That works.

tschaub (approved):

ahocevar (approved):

PR #15190, opened 2023-09-27
Require Geometry array in GeometryCollection constructor

MoonE:
Allowing null instead of an empty array doesn't make much sense, it was only used internally in the GeometryCollection class, but is not needed.
This prevents errors where it was assumed the geometries_ property is always an array and it was not checked.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,5 +1,9 @@
 ## Upgrade notes
 
+### Next Release
+
+ * `ol/geom/GeometryCollection` can no longer be created without providing a Geometry array. Emtpy arrays are still valid.
+
 ### 8.0.0
 
 #### Removal of deprecated properties and methods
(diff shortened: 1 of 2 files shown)


MoonE:
Technically this is a breaking change, as when `listenGeometriesChange_` is now called from the constructor with undefined it throws an exception.
Maybe the should be handled gracefully  until the next major release?

ahocevar:
Nice API simplification! Looks like you have another breaking change lined up already (#15169). I have no problem with frequent major version bumps. Just add a note about this backwards incompatible change to the upgrade notes, give the pull request the "breaking change" label, and we'll bump the major on the next release.

mike-000:
Is this not similar to https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md#v510   I presume as with that change an empty array can be used instead of null?

MoonE:
@mike-000 It is similar, and possibly nobody is affected, but still a breaking change.

ahocevar (approved):

PR #15191, opened 2023-09-28
Remove radius1 for regular shapes, use radius instead

MoonE:
Having two optional parameters where one is required is bad for type-checking.

Current handling is inconsistent. RegularShape constructor treated radius and radius1 as aliases while the webgl style parser used radius exclusively for polygons and radius1 exclusively for stars.

### DIFF
--- a/changelog/upgrade-notes.md
+++ b/changelog/upgrade-notes.md
@@ -1,8 +1,12 @@
 ## Upgrade notes
 
-### Next Release
+### 9.0.0
 
- * `ol/geom/GeometryCollection` can no longer be created without providing a Geometry array. Emtpy arrays are still valid.
+ Removed the `ol/style/RegularShape`'s `radius1` property. Use `radius` for regular polygons or `radius` and `radius2` for stars.
+
+ Removed the `shape-radius1` property from `ol/style/flat~FlatShape`. Use  `shape-radius` instead.
+
+ `ol/geom/GeometryCollection` can no longer be created without providing a Geometry array. Emtpy arrays are still valid.
 
 ### 8.0.0
 
--- a/examples/earthquake-clusters.js
+++ b/examples/earthquake-clusters.js
@@ -46,7 +46,7 @@ function createEarthquakeStyle(feature) {
   return new Style({
     geometry: feature.getGeometry(),
     image: new RegularShape({
-      radius1: radius,
+      radius: radius,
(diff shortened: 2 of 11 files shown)


changelog/upgrade-notes.md
```diff
@@ -1,5 +1,10 @@
 ## Upgrade notes
 
+### Next release
```
jahow, line 3: I think you should explicitly mark that as a breaking change

jahow (commented):
I really like this simplification, thanks! And as we are changing property names anyway, could we use something like `radiusInner` instead of `radius2`?

MoonE:
For RegularShape there is no requirement for `radius` to be larger than `radius2`. So calling it inner / outer doesn't make sense there.
For webgl this unfortunately doesn't hold true, but if this is improved upon in the future it would be a misnomer.
See https://openlayers.org/en/latest/examples/webgl-points-layer.html with a style like:
```json
{
  "shape-points": 5,
  "shape-radius1": ["+", ["*", ["sin", ["time"]], 10], 20],
  "shape-radius2": 15,
  "shape-fill-color": "red",
  "shape-stroke-color": "black",
  "shape-stroke-width": 2
}
```

changelog/upgrade-notes.md
MoonE: Isn't the upgrade notes all about breaking changes? I don't see any previous changes marked explicitly as breaking there.
What do you have in mind?

MoonE:
I added a commit that fixes allows radius2 to be larger than radius for webgl rendering.

changelog/upgrade-notes.md
jahow: For version 8 the breaking changes were put in separate section: https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md#backwards-incompatible-changes

This is what I had in mind

jahow:
> For RegularShape there is no requirement for `radius` to be larger than `radius2`. So calling it inner / outer doesn't make sense there. For webgl this unfortunately doesn't hold true, but if this is improved upon in the future it would be a misnomer

Thanks, you're totally right!

jahow (approved):
Thanks for the simplification and fixing the webgl style parser in the process! I'm not sure when would be the best time for merging this though, but still, at least this is approved.

PR #15203, opened 2023-10-02
Introuded  opacity field in Fill class and handled opacity while rendering in maps

### LINKED ISSUE #15202: Introduce Opacity field in Fill Class
As of now opacity can be set for layers and some styles like Icon, Circle but there is no way we can set opacity
for features like Polygon,Circle,etc. There can be many usecases where we need to use only fill color as a style ,either user has to to specify color in rgba.
Introduce a new attribute opacity in Fill class.
Now if the opacity is present it will add opacity to the given color.

ShanteshSindgi:
Resolves #15202 

1. Introced new field opacity for Fill class, default opacity is 1.
2.If the opacity is provided, parsed the given colors such as normal,hex & rgb to add opacity to them.


ShanteshSindgi:
merge conflicts,raising new PR

Issue #9936, opened 2019-09-16
DragPan & Scroll behaviour mobile/desktop


isthisstackoverflow:
**Is your feature request related to a problem? Please describe.**
I found it to be quite difficult to implement a scenario where all these requirements are fulfilled:

- On desktop, DragPan works as it currently does ✔️
- Mouse wheel scrolls past map ✔️ (easy with the focus condition, or just removing the interaction altogether)
- On mobile, scroll past map with 1 finger ❌
- On mobile, drag pan map with 2 fingers ❌

However, such a scenario is slowly becoming expected by users and customers due to Google Maps acting like that. (I personally like it better how OL does it, but oh well ...)

**Describe the solution you'd like**

Could the DragPan be changed to allow for the described behaviour? I think ideally a new flag would change the behaviour so that the same interaction works differently for mobile and desktop environments as described.

However, I have no idea how to distinguish between desktop and mobile environments properly, and my personal attempts of writing a new interaction as specified haven't worked.

isthisstackoverflow:
I came up with this:

```js
var map = new ol.Map({
  //   ...
  interactions: ol.interaction.defaults({ dragPan: false, mouseWheelZoom: false })
});

map.addInteraction(new ol.interaction.DragPan({
  condition: function(e) {
    return ol.events.condition.noModifierKeys(e) && (!/Mobi|Android/i.test(navigator.userAgent) || this.targetPointers.length === 2)
  }
}));
```

However, with this "solution" ([jsfiddle](https://jsfiddle.net/tnk0ohm3/2/)) scrolling the page with 1 finger on the map is not possible. (Also, I should probably not use this.targetPointers?)

I am not sure entirely sure why, but it seems `.ol-viewport` having `touch-action: none` as element style is the problem. In https://github.com/openlayers/openlayers/issues/6767#issuecomment-378929233 it is suggested to use

```css
.ol-viewport {
  touch-action: auto !important;
}
```

, which works fine in Firefox, but does not work in Chrome, where it ends panning/pinchzooming very quickly after starting.

I'm stuck and assume OL is simply not supporting what I'm trying to do.

isthisstackoverflow:
[...]

PR #15208, opened 2023-10-05
Allow `touch-action: pan-x pan-y;` on viewport

mike-000:
Possible fix for https://github.com/openlayers/openlayers/issues/9936#issuecomment-1747866325

Explicitly setting `touch-action: pan-x pan-y;` seems to fix the issue described on Android Chrome, but unlike `touch-action: none;` it can cause a pull-to-refresh to be triggered, which may be unexpected.

### DIFF
--- a/src/ol/ol.css
+++ b/src/ol/ol.css
@@ -113,7 +113,7 @@
 }
 
 .ol-viewport {
-  touch-action: none;
+  touch-action: pan-x pan-y;
 }
 
 .ol-selectable {


ahocevar:
This fixes the problem for me on iOS.

mike-000:
Since v7.0.0 all versions of the examples have been using ol.css from the website theme folder, which is a copy of the latest version css.   It could become incompatible with those versions and makes it difficult to identity when problems were introduced.

ahocevar (approved):
Thanks, @mike-000

PR #15210, opened 2023-10-06
Use version specific ol.css in examples

mike-000:
As noted in https://github.com/openlayers/openlayers/pull/15208#issuecomment-1749563333 all versions of the examples since 7.0.0 are using the same ol.css instead of the css for that version.  Therefore the behaviour of https://openlayers.org/en/main/examples/two-finger-pan-scroll.html does not yet reflect the changes in #15208, but at the next release that change will be picked up by old versions of the examples.  Examples which use specific versions of the library JavaScript should also be using the corresponding version of the css.


ahocevar (commented):
To avoid issues with that in the future, I think it could make sense to remove both `config/jsdoc/api/template/static/theme/ol.css` and `site/src/theme/ol.css`, and change all occurrences to use the version specific `ol.css`.

mike-000:
Perhaps not realistic to retrospectively change all examples from 7.0.0 to 8.1.0, but `theme/ol.css` should be frozen in its current state for those.  Apart from examples I think the only site change needed is for the map at the top of the home page?

PR #15214, opened 2023-10-07
Fix `fromUserResolution` and `toUserResolution` for `useGeographic()`

kikuchan:
The `fromUserResolution` and `toUserResolution` should use each projection's `getMetersPerUnit()` instead of direct `METERS_PER_UNIT[]` reference.
I also had to change tests because EPSG:4326 (designated by `useGeographic()`) uses 6378137 (WGS84's) as a radius to calculate meters per units, and the tests used 6370997 (Normal sphere's) instead.

I've found this bug when I try to use a `Cluster` with a custom projection that have `units: 'pixels'`, and the `metersPerUnit` option didn't work either.

### DIFF
--- a/test/node/ol/proj.test.js
+++ b/test/node/ol/proj.test.js
@@ -158,7 +158,7 @@ describe('ol/proj.js', function () {
   describe('fromUserResolution()', function () {
     it("adjusts a resolution for the user projection's units", function () {
       useGeographic();
-      const user = 1 / METERS_PER_UNIT.degrees;
+      const user = 1 / getProjection('EPSG:4326').getMetersPerUnit();
       const resolution = fromUserResolution(user, 'EPSG:3857');
       expect(resolution).to.roughlyEqual(1, 1e-9);
     });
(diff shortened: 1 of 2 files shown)


test/node/ol/proj.test.js
```diff
@@ -158,7 +158,7 @@ describe('ol/proj.js', function () {
   describe('fromUserResolution()', function () {
     it("adjusts a resolution for the user projection's units", function () {
       useGeographic();
-      const user = 1 / METERS_PER_UNIT.degrees;
+      const user = 1 / ((Math.PI * 6378137) / 180);
```
ahocevar, line 161: It would be better to use `getProjection('EPSG:4326).getMetersPerUnit()` here.

test/node/ol/proj.test.js
```diff
@@ -175,7 +175,7 @@ describe('ol/proj.js', function () {
       useGeographic();
       const dest = 1;
       const resolution = toUserResolution(dest, 'EPSG:3857');
-      expect(resolution).to.eql(1 / METERS_PER_UNIT.degrees);
+      expect(resolution).to.roughlyEqual(1 / ((Math.PI * 6378137) / 180), 1e-9);
```
ahocevar, line 178: Same as above.

ahocevar (changes requested):
Thanks for your contribution, @kikuchan. Please consider my suggestions.

test/node/ol/proj.test.js
kikuchan: Ok, thanks!

kikuchan:
sorry

ahocevar (approved):
Thanks, @kikuchan

PR #15239, opened 2023-10-16
Fix TileImage's transition is incorrect when re-projection

### LINKED ISSUE #15238: TileImage transition option not work when re-projection
**Describe the bug**
1. Use TileImage with `transition: 0` and projection 3857
2. Use View with 4326
3. The TileImage ReprojTile `transition` not be 0

**To Reproduce**
[Edit tile-transitions (forked)](https://codesandbox.io/s/tile-transitions-forked-38d8gj?fontsize=14&hidenavigation=1&theme=dark)

**Expected behavior**
The TileImage ReprojTile transition to be 0

yujinpan:
Fix #15238


ahocevar (approved):
Very nice fix, thanks @yujinpan

PR #15250, opened 2023-10-17
Style Expressions Concat Function

### LINKED ISSUE #15246: Flat Style Multiple String Expression Combining
I am attempting to combine the combination of two (possibly more) fields into one String Expression where I have multiple fields I want to combine into a single text. I'm not sure if its currently possible to combine the fields in the String Expression to do multiple getters like the following:

```
style: {
    'text-value': [['get', 'adm0name'], ' ', ['get', 'adm1name']],
    'text-font': '16px sans-serif',
    'text-fill-color': 'white',
    'text-stroke-color': 'gray',
    'text-stroke-width': 2,
 }
```
above example is using the [style expression](https://openlayers.org/en/latest/examples/style-expressions.html) data.

I have also created a quick [code sandbox](https://codesandbox.io/s/style-expressions-forked-cqtcl4?file=/main.js) just to quickly create an example to mess around with to try and get the "text-value" property to work. I know it works with a single ['get', 'adm0name'] but it doesn't with both as provided. Ideally, I would like to stick with just the flat style as I didn't want to provide a style function as it was going to increase complexity when I add the data and only need to adjust it slightly based on the field(s) I want to grab.

Any insight would be helpful!

MannyKSoSo:
Added the concat function which can take in 2+ arguments to use the String.concat function. The function can be utilized with other expressions or just regular strings. Arguments are just like the following (from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat)):

```javascript
const greetList = ["Hello", " ", "Venkat", "!"];
"".concat(...greetList); // "Hello Venkat!"
```

This should close #15246.

Note that this is only applied to the canvas expression list and not webgl. I was confused how to actually get the values and haven't worked with raw webgl before. This is also my first pull request so hopefully not too many changes are needed for this. I also updated the style expression example just to show this off. I can revert it back if requested.

### DIFF
--- a/src/ol/expr/cpu.js
+++ b/src/ol/expr/cpu.js
@@ -202,13 +202,10 @@ function compileAccessorExpression(expression, context) {
       return (context) => context.variables[name];
     }
     case Ops.Concat: {
-      return (context_) => {
-        return ''.concat(
-          ...expression.args.reduce((a, c) => {
-            return [...a, compileExpression(c, context)(context_).toString()];
-          }, [])
-        );
-      }
+      const args = expression.args.map((e) => compileExpression(e, context));
+      return (context) => ''.concat(
+        ...args.map((arg) => arg(context).toString())
+      );
     }
     default: {
       throw new Error(`Unsupported accessor operator ${expression.operator}`);
--- a/src/ol/expr/cpu.js
+++ b/src/ol/expr/cpu.js
@@ -203,9 +203,11 @@ function compileAccessorExpression(expression, context) {
     }
     case Ops.Concat: {
       return (context_) => {
-        return "".concat(...expression.args.reduce((a, c) => {
-          return [...a, compileExpression(c, context)(context_).toString()];
-        }, []));
+        return ''.concat(
+          ...expression.args.reduce((a, c) => {
+            return [...a, compileExpression(c, context)(context_).toString()];
+          }, [])
+        );
       }
     }
     default: {
--- a/src/ol/expr/cpu.js
+++ b/src/ol/expr/cpu.js
@@ -203,9 +203,8 @@ function compileAccessorExpression(expression, context) {
     }
     case Ops.Concat: {
       const args = expression.args.map((e) => compileExpression(e, context));
-      return (context) => ''.concat(
-        ...args.map((arg) => arg(context).toString())
-      );
+      return (context) =>
+        ''.concat(...args.map((arg) => arg(context).toString()));
     }
     default: {
       throw new Error(`Unsupported accessor operator ${expression.operator}`);
--- a/src/ol/expr/cpu.js
+++ b/src/ol/expr/cpu.js
@@ -105,9 +105,13 @@ function compileExpression(expression, context) {
       return compileAssertionExpression(expression, context);
     }
     case Ops.Get:
-    case Ops.Var:
-    case Ops.Concat: {
+    case Ops.Var: {
       return compileAccessorExpression(expression, context);
+    }
+    case Ops.Concat: {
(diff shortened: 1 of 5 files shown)


src/ol/expr/cpu.js
```diff
@@ -200,6 +201,13 @@ function compileAccessorExpression(expression, context) {
     case Ops.Var: {
       return (context) => context.variables[name];
     }
+    case Ops.Concat: {
+      return (context_) => {
+        return "".concat(...expression.args.reduce((a, c) => {
+          return [...a, compileExpression(c, context)(context_).toString()];
+        }, []));
```
MoonE, line 208:
```suggestion
      const args = expression.args.map((e) => compileExpression(e, context));
      return (context) => args.map((arg) => arg(context)).join('');
```

Compiling of the expression shoukd be done outside the evaluation. This way it will only be compiled once when the style is created.

src/ol/expr/expression.js
```diff
@@ -240,6 +241,11 @@ export const Ops = {
 const parsers = {
   [Ops.Get]: createParser(AnyType, withArgsCount(1, 1), withGetArgs),
   [Ops.Var]: createParser(AnyType, withArgsCount(1, 1), withVarArgs),
+  [Ops.Concat]: createParser(
+    AnyType,
```
MoonE, line 245:
```suggestion
    StringType,
```
Concat will always return string.

src/ol/expr/cpu.js
MannyKSoSo: Gotcha, I can move it outside of the context function. But since this uses the concat function as [String.concat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat) should I change it to the following?

```javascript
const args = expression.args.map((e) => compileExpression(e, context));
return (context) => ''.concat(...args.map((arg) => arg(context)));
```

I can change it to a 'Join' instead of the concat, but would just need to pop out which argument would be the join's character.

src/ol/expr/expression.js
MannyKSoSo: Will apply with next commit with the compile expression move to outside the evaluation.

src/ol/expr/cpu.js
MoonE: I guess the functionality is the same, so either would work.

src/ol/expr/cpu.js
```diff
@@ -105,7 +105,8 @@ function compileExpression(expression, context) {
       return compileAssertionExpression(expression, context);
     }
     case Ops.Get:
-    case Ops.Var: {
+    case Ops.Var:
+    case Ops.Concat: {
```
jahow, line 109: I think you can put the implementation of the expression here instead of in `compileAccessorExpression`, because `concat` cannot really be considered an accessor expression; I guess eventually we might have a `compileStringExpression` function if we have more than just concat :)

jahow (commented):
Thanks for working on this! Don't worry about implementing the operator in webgl, it's not even possible for now as strings do not exist in the GLSL language.

src/ol/expr/cpu.js
MannyKSoSo: Alright, I can move it. I personally can't see many more string expressions being a thing, but I could be missing some use cases that people have.

tschaub:
Thanks for the contribution, @MannyKSoSo

src/ol/expr/cpu.js
jahow: I was thinking of uppercase, lowercase, capitalize, substring etc.

Issue #15268, opened 2023-10-23
default `size` for the `fit` function behaves unintuitively


simonseyock:
**Is your feature request related to a problem? Please describe.**
If you work on an app that uses multiple openlayers map, the default size value is very unintuitive. The actual size default value is the size of the first map in the dom. One would expect that the size is from the map that uses the view. I know that the view does not have a reference to the map.

**Describe the solution you'd like**
I would suggest to move the size out of the options object and make it a parameter (`view.fit(extent, size, options)` . This will remind the developer to think about it.

This would be a breaking change.

Also I would consider removing the default. In every case I encountered while working with openlayers so far, the map was known at the point where `.fit` was called.

ahocevar:
Maybe the documentation needs an update. Since quite a while, `fit` does not use the size of the first map in the dom, but the actual size of the map the view is associated with. The `size` parameter is only necessary for a view that has not been associated with a map yet.

Maybe you want to create a pull request that fixes the documentation?

simonseyock:
Ok, then it is bugged. I will look into that.

simonseyock:
If `setTarget` is called from a react mount callback the call of `getComputedSize` returns a CSSStyleDeclaration that does not contain `width` and `height`.
That is no problem in `updateSize` because it uses the `offsetWidth` and calculates the actual size from there. But in `updateViewportSize_` the problem occurs.
This leads to the question: Why is `updateViewportSize_` in `Map.js` a different logic from `updateSize`, anyways? I don't see any reason for that.

I would suggest passing the size as a parameter to `updateViewportSize_`.

mike-000:
> the actual size of the map the view is associated with

A view can be associated with multiple maps.  In this example https://codesandbox.io/s/side-by-side-forked-qwqf8v?file=/main.js both `calculateExtent()` and `fit()` are defaulting to the size of the second map - if map sizes can be independently updated it could be either map.

simonseyock:
Ok, according to implementation then the size that would be used, would be the map where one of the following events occured last:
* setSize
* resize event
* setView
* setTarget

I would then write that the size can not be determined correctly if more than one map is used with one view. Is this features used often? I think it might be used for overview maps, but in any application I work with, we actually use a different view.

simonseyock:
See the linked PR for a formulation.

PR #15269, opened 2023-10-23
fix(Map): use same size calculation in `updateSize` and `updateViewportSize_`

simonseyock:
Fixes: #15268

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -75,10 +75,9 @@ import {fromExtent as polygonFromExtent} from './geom/Polygon.js';
 /**
  * @typedef {Object} FitOptions
  * @property {import("./size.js").Size} [size] The size in pixels of the box to
- * fit the extent into. If the view is associated with one map, it defaults to
- * the size of the map. If it is associated with multiple maps, the size can be
- * of any one of the maps, so the desired size should be specified. If no map
- * is associated it uses `[100, 100]` as the size.
+ * fit the extent into. Defaults to the size of the map the view is associated with.
+ * If no map or multiple maps are connected to the view, the desired box size
+ * (e.g. `map.getSize()`) has to be provided to get a usable result.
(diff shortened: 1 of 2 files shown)


src/ol/View.js
```diff
@@ -74,9 +74,11 @@ import {fromExtent as polygonFromExtent} from './geom/Polygon.js';
...
 /**
  * @typedef {Object} FitOptions
- * @property {import("./size.js").Size} [size] The size in pixels of the box to fit
- * the extent into. Default is the current size of the first map in the DOM that
- * uses this view, or `[100, 100]` if no such map is found.
+ * @property {import("./size.js").Size} [size] The size in pixels of the box to
+ * fit the extent into. If the view is associated with one map, it defaults to
+ * the size of the map. If it is associated with multiple maps, the size can be
+ * of any one of the maps, so the desired size should be specified. If no map
+ * is associated it uses `[100, 100]` as the size.
```
ahocevar, line 81:
```suggestion
 * @property {import("./size.js").Size} [size] The size in pixels of the box to
 * fit the extent into. Defaults to the size of the map the view is associated with.
 * If no map or multiple maps are connected to the view, the desired box size
 * (e.g. `map.getSize()`) should be provided.
```

ahocevar (changes requested):
Good simplification. Thanks, @simonseyock. I have added a suggestion for simpler wording that you may want to consider. Looks good otherwise.

src/ol/View.js
simonseyock: I changed `should be provided` to `has to be provided to get a usable result`.

simonseyock:
Just to clarify: This not only simplifies it, but also fixes a bug that was occurring in the removed code part.

simonseyock:
Also I updated the docs for `calculateExtent` with the same text.

ahocevar (approved):
Thanks, @simonseyock

PR #15270, opened 2023-10-24
Moved geometry type check for hit detection in Modify interaction

### LINKED ISSUE #15235: Modify interaction with hit detection treats all features as Points
**Describe the bug**
When a Modify interaction is configured with a hit detection layer, all features seem to be treated as Points.
This is caused by a recent change in Modify.js ol version 7.4.0:
https://github.com/openlayers/openlayers/blob/1f9250c0f7223dbf7c4f128d28649a9c534fbf33/src/ol/interaction/Modify.js#L1187-L1190

**To Reproduce**
1. Go to https://codesandbox.io/s/draw-and-modify-features-with-hit-detection-5x5948?file=/main.js
2. Draw a line on the map
3. Try modifying the line
4. Notice that the first point of the line is modified instead of the one you are trying to modify

**Expected behavior**
Obviously, the dragged point on line should be modified, instead of the 1st.

**Suggestion**
```
if (geometry && geometry.getType() === 'Point') {
	geometry = new Point(
		toUserCoordinate(geometry.getCoordinates(), projection)
	);
}
```

adhaca:
Fixes #15235

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -1184,14 +1184,13 @@ class Modify extends PointerInteraction {
       map.forEachFeatureAtPixel(
         pixel,
         (feature, layer, geometry) => {
-          if (geometry) {
+          if (geometry && geometry.getType() === 'Point') {
             geometry = new Point(
               toUserCoordinate(geometry.getCoordinates(), projection)
             );
           }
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @adhaca

PR #15277, opened 2023-10-28
Use local theme instead of root website theme for versioned examples

ahocevar:
Previously, the examples on the website used the theme of the website, which contains ol.css of the latest version. This pull request changes the links to the local theme of the versioned examples, so ol.css matches the version of the examples.

Replaces and closes #15210
Replaces and closes https://github.com/openlayers/openlayers.github.io/pull/133

### DIFF
--- a/examples/templates/example.html
+++ b/examples/templates/example.html
@@ -8,17 +8,17 @@
     <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@6.1.2/css/fontawesome.min.css" crossorigin="anonymous">
     <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@6.1.2/css/solid.css" crossorigin="anonymous">
     <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@6.1.2/css/brands.css" crossorigin="anonymous">
-    <link rel="stylesheet" type="text/css" href="/theme/ol.css">
-    <link rel="stylesheet" type="text/css" href="/theme/site.css">
+    <link rel="stylesheet" type="text/css" href="./theme/ol.css">
+    <link rel="stylesheet" type="text/css" href="./theme/site.css">
 {{#each css.local}}
     <link rel="stylesheet" type="text/css" href="{{{ . }}}">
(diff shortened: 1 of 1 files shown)


jahow (approved):
Thanks!

tschaub:
Ideally we could update things like the logo and the overall website theme and examples for older versions would be updated as well.  That was the reason for using a shared CSS file for the website.  But I understand that older versions may have CSS specific to the controls etc.  I haven't tried, but wonder if the website CSS can be "global" and the example CSS can be versioned.

examples/templates/example.html
```diff
@@ -8,17 +8,17 @@
...
-    <link rel="stylesheet" type="text/css" href="/theme/site.css">
+    <link rel="stylesheet" type="text/css" href="./theme/ol.css">
+    <link rel="stylesheet" type="text/css" href="./theme/site.css">
 {{#each css.local}}
     <link rel="stylesheet" type="text/css" href="{{{ . }}}">
 {{/each}}
-    <link rel="icon" type="image/svg+xml" href="/theme/img/logo-light.svg" media="(prefers-color-scheme: light)" />
-    <link rel="icon" type="image/svg+xml" href="/theme/img/logo-dark.svg" media="(prefers-color-scheme: dark)" />
+    <link rel="icon" type="image/svg+xml" href="./theme/img/logo-light.svg" media="(prefers-color-scheme: light)" />
+    <link rel="icon" type="image/svg+xml" href="./theme/img/logo-dark.svg" media="(prefers-color-scheme: dark)" />
```
tschaub, line 17: Is the problem about the logo?  Seems like it would be nice if this could be shared.

ahocevar: The problem was with the current `ol.css` being potentially incompatible with older versions. I thought it would make sense to take all static assets for the examples from the respective example build. But it would be easy to decide on a per-file basis in [examples/templates/example.html](https://github.com/openlayers/openlayers/pull/15277/files/5536dea2c4676368d18a4d1f263638e1125146b7#diff-5fcb569c5d8722abaa8bb95603c3a6e0b13e0e0401335d866bc7d5565804da28) whether to use `/theme/` (theme created upon website build) or `./theme/` (theme created upon examples build).

So it looks like what you're envisioning is to only keep `ol.css` version specific, and switch everything else back to website build specific. I created #15280 with this change. @tschaub Maybe you want to review that and see if it's what you're envisioning.

Issue #13367, opened 2022-02-11
Add support for WebGL2


DonIsaac:
# The Problem

The primary issue I'm encountering is not being able to use [anti-aliasing render buffers](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/renderbufferStorageMultisample). These would not be available without some other code changes, but the first step in this process is exposing access to [WebGL2](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext).

The secondary benefit is being able to pass uniforms all at once using a [uniform block](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/uniformBlockBinding). While a large number of use cases would not see a significant performance boost, applications using a large number of uniforms and rendering at 60fps would.

Additionally WebGL2 adds a bunch of [new features](https://webgl2fundamentals.org/webgl/lessons/webgl2-whats-new.html) which would be nice to have.

# Potential Solutions

A basic, backwards-compatible solution would be to add a feature flag, via a property on the existing options object, when creating a WebGL layer. By default, WebGL2 uses GLSL 3.0 (whereas WebGL uses GLSL2.1), which has a [large number of breaking changes](https://www.khronos.org/opengl/wiki/History_of_OpenGL#OpenGL_3.0_.282008.29). This is fine when developing a new application where you are writing your own shaders, but problematic for both existing applications and the shader builder code. The shader builder could be adapted to handle WebGL2, but that would be a much larger amount of work.

tschaub:
Can you describe a bit more about what you are trying to do?

DonIsaac:
For my specific use case, I'm making a map displaying a high number of pins on a massive display, specifically 9 4K screens in a 3x3 grid. Due to the size and resolution of this map, aliasing effects are highly visible. Pin images are either a SVG or a PNG (the difference isn't relevant as aliasing is visible on both). Pins are rendered and animated with custom shaders using a WebGLPointsRenderer.

ahocevar:
As an alternative, you could try setting a higher `pixelRatio` on your map, e.g.
```js
new Map({
  // ...
  pixelRatio: devicePixelRatio * 2
});
```
Maybe the result will be a canvas that is too big for the browser to handle, but at least it's worth a try.

DonIsaac:
This solution did work, at the expense of frame rate. However, the frame rate is still good enough for this to work for me, thank you!

[...]

PR #15279, opened 2023-10-28
WebGL / Add support for patterns in polygons and lines

jahow:
This PR adds support for fill and stroke patterns using the following properties of the flat WebGL-specific style:
* `fill-pattern-src` ~and `fill-pattern-img`~
* `fill-pattern-offset`,  `fill-pattern-offset-origin` and `fill-pattern-size`
* `stroke-pattern-src` ~and `stroke-pattern-img`~
* `stroke-pattern-offset`, `stroke-pattern-offset-origin` and `stroke-pattern-size`
* `stroke-pattern-spacing`

These properties work the same way as the `icon-` properties, letting the user specify an image either as a path ~or an `Image` or `Canvas` instance~; the offset properties let the user specify a image in a spritesheet.

I didn't reproduce the current API which accepts `CanvasPattern` and `CanvasGradient` because:
* it is tailor-made for the Canvas2D API and has limitations for WebGL (a texture can be created out of it, but the dimensions are not known)
* doesn't let the user pick a sprite in a spritesheet
* is not serializable

## Fill pattern

Fill patterns have an origin which can be [0,0] in case of vector layers, or the tile origin in case of vector tiles. Patterns are slightly upscaled/downscaled when zooming to match integer zoom levels:

Fill patterns are tinted by the fill color if specified.
Peek 28-10-2023 15-11
(note: I didn't touch the actual webgl vector example)

## Stroke pattern

Stroke patterns are scaled to match the line width. They are tinted by the line color if specified. I also worked on fixing the distance computation along the line when applying an offset; before this PR, the computation was off and symbols were clearly not lining up at the joins. Now the symbols line up, although the interpolation the joins still needs work:

Before:
Peek 28-10-2023 15-16

After:
Peek 28-10-2023 15-15

Note: providing `Canvas` or `Image` for fill patterns, stroke patterns or icons is not supported anymore. Use data URLs or object URLs instead.


ahocevar:
Nice! It should be easy to add support for fill pattern images to the Canvas 2D renderer. Stroke patterns like this will be a bit harder to support in Canvas 2D, because they need to consider the segment angle, but doable.

Regarding the new style properties, I'm wondering if we could get along without `fill-pattern-offset-origin` and `stroke-pattern-offset-origin`, because I think we'll be moving towards sprite sheets provided to the layer or map, separate from the styles, and the benefit of specifying the origin is quite minor, compared to always assuming `top-left`. With separate sprite sheets, we can eventually remove all of these new properties and just have `fill-pattern` and `stroke-pattern` instead.

mike-000:
Fill is not working in the deploy preview on some setups when there is no feature detected
image
(also hit detection reports "Faroe Islands boreal grasslands" for every feature - but that is not caused by this PR as it also happens in main).

jahow:
> Nice! It should be easy to add support for fill pattern images to the Canvas 2D renderer. Stroke patterns like this will be a bit harder to support in Canvas 2D, because they need to consider the segment angle, but doable.

Good to hear, feature parity would be awesome.

> Regarding the new style properties, I'm wondering if we could get along without `fill-pattern-offset-origin` and `stroke-pattern-offset-origin`, because I think we'll be moving towards sprite sheets provided to the layer or map, separate from the styles, and the benefit of specifying the origin is quite minor, compared to always assuming `top-left`. With separate sprite sheets, we can eventually remove all of these new properties and just have `fill-pattern` and `stroke-pattern` instead.

The `offset-origin` properties are definitely not essential, but I figured they might still be a useful feature, and they don't bring that much complexity. So I don't really have a strong opinion either way.

> Fill is not working in the deploy preview on some setups when there is no feature detected

@mike-000 could you please tell me on which platform you witnessed this? Thanks a lot!!

mike-000:
One of the systems mentioned in https://github.com/openlayers/openlayers/issues/13367#issuecomment-1780868465 where until now there were no problems with WebGL, only WebGL2.  In this case Firefox is also affected.  The same ANGLE backend change will fix Chrome and Edge, I am not sure what can be done for Firefox.

Results for https://webglreport.com/?v=1 with Chrome

default settings (with the issue)
image

and with Angle backend set to `D3D11on12` in flags (no issue but slow)
image

jahow:
Thanks @mike-000 I could reproduce the issue (false hit detection + missing fill) using an emulated win10 system running both FF and Chrome; it's not related to this PR though.

This looks like an ANGLE-related issue. I will look into this, see if there's anything that can be done on OL side.

mike-000:
I have seen missing fill in the example in main as well now, so cannot be related to this PR.  Both issues can also been seen in the deploy preview for #15121 so probably started with that.

src/ol/style/webgl.js
```diff
@@ -21,7 +20,15 @@
 
 /**
  * @typedef {Object} FillProps
- * @property {ColorExpression} [fill-color] The fill color.
+ * @property {ColorExpression} [fill-color] Fill color.
+ * @property {string} [fill-pattern-src] Fill pattern image source URI. If `fill-color` is defined as well, it will be used to tint this image.
+ * @property {HTMLImageElement|HTMLCanvasElement} [fill-pattern-img] Fill pattern image object. Can be provided as an alternative to `fill-pattern-src`.
```
tschaub, line 25: Accepting an image here is at odds with the goal of parsing styles and preparing attribute buffers in a worker.  But we could wait until we are doing that and then deal with this then.

src/ol/style/webgl.js
```diff
@@ -1,7 +1,6 @@
 /**
- * Literal style objects differ from standard styles in that they cannot
- * be functions and are made up of simple objects instead of classes.
- * @module ol/style/literal
+ * WebGL style objects slightly differ from standard flat styles for certain properties
+ * @module ol/style/webgl
```
tschaub, line 3: Feels like we should work toward having a single place for these types.  I think it would be preferable to have a single set of types and then to allow the implementations to differ (or be incomplete)  rather than having two sets of types.

tschaub (approved):

tschaub:
Really impressive work, @jahow.  I only left a couple general comments that I'm curious to hear your thoughts on, but think this looks good to go.

src/ol/style/webgl.js
jahow: Absolutely, and a procedurally-generated image could easily be made into an object URL that could then be read from a worker.

I really went with a conservative approach here and mimicked the existing icon properties (also true for the `origin-offset` properties which @ahocevar mentioned). But I'm considering removing the code path for `Image` objects entirely, as there's no point in introducing APIs which we know are not aligned with long-term goals anyway.

src/ol/style/webgl.js
jahow: This is definitely what I'm aiming for. Having a separate type for WebGL renderers is helpful to track of where we currently stand and what is the gap between the two style formats, but it's not a situation that we want to keep eternally.

There is still some work to close this gap though, mainly around the concept of rules; I plan on addressing this later on while optimizing the way features are filtered out before rendering in WebGL.

jahow:
@tschaub I added a commit to remove all traces of `-img` properties (for icons, fill patterns and stroke patterns), this bringing the webgl style closer to the flat one.

jahow:
> The docs describe the action of `fill-color`/`stroke-color` on patterns as "to tint". As in #15290 was it intentional for this to affect the alpha of the pattern?

Yes absolutely, being able to get a translucent pattern is one benefit of the "tinting" behavior. Maybe the doc could clarify that tinting means multiplying colors?

mike-000:
@jahow  That makes sense.  Applying `icon-color` of `[255, 0, 0, 0.25)` in canvas currently has the same effect as `[255, 191, 191, 1]` (which is the result of applying `[255, 0, 0, 0.25]` over a white canvas) so `[255, 191, 191, 1]` should always be used to achieve that, and the bug is `[255, 0, 0, 0.25]` not multiplying correctly in canvas styles.

mike-000:
Options to scale/resize patterns would be useful.  For example the Esri feature style linked in https://github.com/openlayers/openlayers/pull/15095#issuecomment-1775303924 (reproduced in https://codesandbox.io/s/vector-esri-forked-3ly3ry?file=/main.js) contains a data url for the icon pattern (which produces a 256 x 256 image) in the eastern-most polygon and specifies a smaller (64 pixel) width and height to be used in the pattern

This could be specified as

```
  'fill-pattern-src': 'data:' + symbol.contentType + ';base64,' + symbol.imageData,
  'fill-pattern-width': symbol.width,
  'fill-pattern-height': symbol.height,
```

but without support for that the data url would need be to be asynchronously loaded to an image, then scaled in canvas to produce another data url.

jahow:
I agree @mike-000, but with the limited resources available we're trying to aim for supporting Mapbox Style features, and the spec does not offer any option for scaling fill patterns.

PR #15280, opened 2023-10-29
Use ol.css from example build, logos and site css from the website

ahocevar:
Addresses https://github.com/openlayers/openlayers/pull/15277#issuecomment-1783996545


tschaub (approved):
Thanks, @ahocevar

PR #15295, opened 2023-11-04
Fix canvas `icon-color` alpha handling

### LINKED ISSUE #15290: `icon-color` inconsistency between canvas and WebGL
**Describe the bug**
Alpha in `icon-color` has a different effect in canvas and WebGL.

**To Reproduce**
Compare applying an icon-color of `#ff000040` in WebGL (left) and canvas (right)

image
In WebGL the alpha affects the icon opacity, in canvas it does not and merely tints the icon with a semi-opaque color.

**Expected behavior**
Same behavior in canvas and WebGL  Should a transparent "tint" make an icon transparent or have no effect?

mike-000:
Fixes #15290

Makes the handling of alpha in `icon-color` and icons in IconImage consistent with WebGL


tschaub:
The icon already accepts an `opacity`.  It seems like it would be simpler to parse the color and remove any alpha before rendering it.  Then the alpha could be applied as opacity.  And we need to decide how to handle cases when both opacity and alpha are supplied.

mike-000:
Opacity is an updateable setting for all ol/style/Image subclasses (which might also have alpha in their fill or stroke) and is handled by the canvas layer renderer, while `color` is applied when the IconImage is constructed.  Also an icon can have semi transparent regions (as in the test), which were also not handled consistently compared to WebGL.  Applying an opacity in addition to a color with alpha appears to give consistent results for canvas and WebGL in the test setup.

mike-000:
I have checked how well outputs from canvas and WebGL match in a setup similar to the rendering test, and with various combinations of `icon-color` alpha, `icon-opacity` and layer `opacity`, using `ol/source/Raster` to highlight any pixels which are not exactly the same

image

It is apparent that using 0.5 twice in those situations can produce more differences than using 0.25 once - this is because there is no such thing as a 0.5 alpha, it is either 127/255 or 128/255 - and using those together will produce a better match.  However specifying 127/255 as alpha for an `icon-color` does not work as expected as `ol/color#asString` rounds alpha to 2 places, which then produces the same result as 128/255.

mike-000:
The IE fallback removed in #13912 always worked correctly, and was probably faster even without the additional `drawImage` calls needed to make the `multiply` globalCompositeOperation work as expected, so is used again as long as the canvas is not tainted.

mike-000:
@tschaub @ahocevar Ready for review

mike-000:
@ahocevar any chance of someone reviewing this?
The main issue it is trying to fix is seen with an icon (or fill pattern) containing (from center outward) pure white, semi-transparent white, semi-transparent black, pure black, and pure transparency, as seen on the left.

image

When multiplied by pure red specifications such as KML expect the result **on the right** - both whites become red with alpha unchanged, and the blacks are unaffected.  The WebGL vector renderer does that, as did the old IE workaround.

Google Earth:
image

With the removal of the IE workaround the canvas renderer currently uses a multiply globalCompositeOperation which produces the blended result in the center.  This PR effectively restores the IE workaround and, for tainted canvas only, uses a lengthy fallback to produce the same result with globalCompositeOperations.

MapLibre uses an sdf operation where the color replaces fully opaque regions while semi opaque regions becomes fully transparent (the ol-mapbox-style code appears to do something similar) which is not relevant here.

@tschaub's suggestion to use the color's alpha as icon opacity is valid, but could be done independently to this in the Icon style at any time - it would not affect the output but could offer more configurability.  Even if Icon style removed alpha from color IconImage would still need to support colors with alpha for fill patterns.

PR #15306, opened 2023-11-07
Add segment information to snap result

manisandro:
Followup #13613

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "pixelmatch": "^5.1.0",
         "pngjs": "^7.0.0",
         "proj4": "2.9.2",
-        "puppeteer": "21.4.1",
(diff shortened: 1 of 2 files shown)


ahocevar:
Thanks for picking this up again, @manisandro. What do you think about also adding the `segment` to the `Snap` event? That way, users would be able to benefit from the new result property.

A bonus task would be to add a test to make sure the returned result is what you expect.

ahocevar:
@manisandro Thanks for your work on this. I had to clean up your commit history a bit, and added another commit to handle circles properly. See #15312.

Next time, please create a feature branch when creating a pull request, instead of creating a pull request from the `main` branch of your fork. This will make it easier for maintainers to modify the pull request before merging, because then we do not have to create a new pull request to do that.

PR #15309, opened 2023-11-08
Export TileGrid and WMTSTileGrid from ol/tilegrid

mike-000:
It seems inconsistent that
`import {Projection, fromLonLat} from 'ol/proj.js'
`
and
`import {Layer, Tile as TileLayer} from 'ol/layer.js'
`
will work,

while
`import {TileGrid, createXYZ} from 'ol/tilegrid.js'
`
or
`import {TileGrid, WMTS as WMTSTileGrid} from 'ol/tilegrid.js'
`
do not.

This adds the missing exports to support that

### DIFF
--- a/src/ol/tilegrid.js
+++ b/src/ol/tilegrid.js
@@ -13,6 +13,9 @@ import {
 } from './extent.js';
 import {toSize} from './size.js';
 
+export {TileGrid};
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #15316, opened 2023-11-10
Increase alpha precision in rgba strings created from color arrays

mike-000:
As noted in https://github.com/openlayers/openlayers/pull/15295#issuecomment-1804231066 2 decimal place alpha in rbga strings is not as precise as eight character hex notation, sometimes differing from the hex value by 2, and this means that color arrays cannot be reliably used to provide string values with precise alpha for canvas drawing.  3 places are needed to match the precision of hex format.


ahocevar (approved):
Thanks, @mike-000

PR #15317, opened 2023-11-10
WebGL / Fix pattern fill issues with vector tiles

jahow:
This PR addresses an issue that happened when adding a fill pattern to vector tiles in WebGL: the fill pattern was interfering with the tile mask texture and caused a visual bug.

Also contains several improvements to fill and stroke patterns:
* improve texture sample logic to avoid artifacts at the pattern borders
* inverse the stroke pattern y coordinate to make it so that the right-hand side of a line is "up"
* fix a shader compilation error in case both a stroke pattern and a fill-color were set
* Take into account view rotation for the fill pattern
* ~add a `fill-pattern-scale` option as suggested in https://github.com/openlayers/openlayers/pull/15279#issuecomment-1793595475~
  ~This was actually necessary to make sure that patterns line up well when using vector tiles (in case patterns do not have a size which is a power of two)~


src/ol/style/webgl.js
```diff
@@ -28,6 +28,8 @@
  * `top-left` or `top-right`.
  * @property {ExpressionValue|import("../size.js").Size} [fill-pattern-size] Fill pattern image size in pixel. Can be used together with `fill-pattern-offset` to define the
  * sub-rectangle to use from the origin (sprite) fill pattern image.
+ * @property {ExpressionValue|number|import("../size.js").Size} [fill-pattern-scale=1] Scaling of the fill pattern. Can be used to have seamless tiling with patterns that
+ * do not have a power-of-two size.
```
ahocevar, line 32: Again, looking at the perspective to replace individual image handling in styles with sprite sheets, I'm not sure if we should add this. It is easy to generate fill pattern images in a way that their size is a power of two. And sprite patterns used in maps following the Mapbox Style spec have that already.

However, the scaling capability (not the `fill-pattern-scale` property) could be useful to scale pattern images to adjust the pattern pixel ratio to the device pixel ratio, like Mapbox does. To prepare for that, I feel like a `fill-pattern-pixel-ratio` property would suit better until we get rid of image handling with style properties in favor of sprite sheets.

Something like this:

```suggestion
 * @property {ExpressionValue|number|import("../size.js").Size} [fill-pattern-pixel-ratio=1] Pixel ratio of the fill pattern. This is used to ensure same visual appearance
 * of the pattern regardless of the device pixel ratio.
```

jahow: In my mind the scale property could be used for both:
* adapting for pixel ratio and
* adjusting to ensure seamless tiling

I realized that sprites in a Mapbox style did not always have a power-of-two size; apparently Mapbox/Maplibre handles that seamlessly for the user, but in the context of OpenLayers I'm not sure how that could be done without the user providing a scale ratio to fix that.

ahocevar: > * adapting for pixel ratio and

In theory yes, but not with a static style, because you cannot pass the pixel ratio as value to the `fill-pattern-scale` property.

> I realized that sprites in a Mapbox style did not always have a power-of-two size

That's correct, and it's because most sprite images are only used as icon, where there is no requirement for them to have a power-of-two size. Sprite images meant to be used as pattern should always have a size that is a power of two.

ahocevar: @jahow Thinking about the use case that came up in https://github.com/openlayers/openlayers/pull/15318#issuecomment-1808451899, it might make sense to introduce a `fill-pattern-transform`, instead of just `fill-pattern-scale`. What do you think about this? It would also not solve the pixel ratio problem, but maybe that can be solved later elsewhere in combination with registered sprite sheets.

jahow: If we provide a `['pixel-ratio']` operator then a static style can handle that as well automatically!

About `fill-pattern-transform`, what would it look like, a matrix? I agree that in principle it makes sense to let users specify a translation offset as well as an angle offset for the pattern.

ahocevar: Yes, `fill-pattern-transform` would be an array of 6 values, like a DOMMatrix.

How would the `['pixel-ratio']` operator work if you want to use a different src for pixel ratios less than and greater than 1.5? I think it would only cover one part of the equation, i.e. which src to use for a given pixel ratio. It would not allow to tell the renderer what scaling to apply to the provided image. But all these questions are beyond the scope of this pull request. I think a `fill-pattern-transform` would be useful regardless, because it would allow to rotate images, e.g. to have hatches at different angles.

jahow: @ahocevar about `fill-pattern-transform`, wouldn't providing a array of 6 numbers be awkward for users? for example, to have both translation, rotation and scale one should provide:
```
fill-pattern-transform: [
  scaleX * cos(angle),
  scaleY * sin(angle),
  -scaleX * sin(angle),
  scaleY * cos(angle),
  translateX,
  translateY
]
```
I agree that this offers a very high degree of freedom, but I'm worried it might make the API too obtuse. The [DOMMatrix API](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix) looks quite powerful, but I'm not sure we can use it with expression-based values.

On the other hand something like this might cover most use-cases:
```
fill-pattern-rotation: angle,
fill-pattern-displacement: [translateX, translateY],
fill-pattern-scale: [scaleX, scaleY]
```
It's also similar to other APIs, although that argument might not weigh much.

ahocevar: I see your point, but the problem is that with rotate, translate and scale, the result depends on the order they are applied. With a transform matrix, you can use whatever tools you want to create the correct transform with operations like rotate, translate and scale in the correct order, but you won't get lucky with expressions to e.g. get a rotation from a feature attribute.

So the more I think about this, the more I agree with Mapbox - they only have a `fill-pattern`, so if users want different versions of a pattern (scales, rotations), they need to pre-generate them in a style sheet. Everything else will lead to a bloated API.

When we started this discussion, I was actually only concerned about the correct scaling in two ways:
* Allow for using different resolution version of an image depending on the device pixel ratio (in Mapbox done by providing 1x and 2x sprite sheets)
* Make patterns look the same on every screen regardless of the device pixel ratio (in Mapbox done by calculating the ratio between the available sprite sheet pixel ratio and the device pixel ratio).

I'd be happy to approve this pull request without the added `fill-pattern-scale` property, and I think we can answer the above questions without adding new style properties as soon as we introduce sprite sheets.
*

jahow: Got it, I'll let the `scale` option out and we can discuss spritesheets later on.

PR #15318, opened 2023-11-12
Fill pattern for Canvas renderer flat styles

ahocevar:
This pull request brings the `fill-pattern-src`, `fill-pattern-size` and `fill-pattern-offset` flat style properties, which were recently added to the WebGL renderer, to the Canvas 2d renderer. This is done by introducing a `PatternDescriptor` object structure, which can also be passed to `ol/style/Fill`'s `color` property instead of a `CanvasPattern`.

The behavior of a pattern created from a `PatternDescriptor` or flat style is the same as a `CanvasPattern` passed to `ol/style/Fill`'s `color` property, except for how pixel ratio is handled. Same as for the WebGL renderer, the pattern is scaled by the map's pixel ratio to ensure the same visual map output regardless of the pixel ratio.

Other than the WebGL renderer, there is no scaling between integer zoom levels. This is the reason why the rendering test result for `fill-pattern` is slightly different from `webgl-fill-pattern`.


mike-000:
`fill-pattern-src` used in conjunction with `fill-color` does give a significantly different result to WebGL if either has semi-transparency, but #15295 should fix that without any additional changes (although #15295 could be rebased after this is merged and a second rendering test added to confirm that).

StefanBrand:
Really cool that you have introduced this functionality so quickly! I have tried it out in our EOX AgriApp already. Here is a working usage example:

image

```js
function getLayerConfig() {
      const cnv = document.createElement("canvas");
      cnv.width = 16;
      cnv.height = 32;

      const ctx = cnv.getContext("2d");
      ctx.fillStyle = "#d95f02";
      ctx.fillRect(0, 0, cnv.width, cnv.height);

      ctx.beginPath();
      ctx.fillStyle = "#7fc97f";
      ctx.moveTo(0, 0);
      ctx.lineTo(16, 16);
      ctx.lineTo(0, 32);
      ctx.closePath();
      ctx.fill();

      // const pattern = ctx.createPattern(cnv, "repeat");
      // const matrix = new DOMMatrix();
      // pattern.setTransform(matrix.rotate(90).scale(1 * 0.75));

      const patternStyle = {
        "stroke-color": "yellow",
        "stroke-width": 3,
        "fill-pattern-src": cnv.toDataURL("png"),
      };

      return {
        type: "VectorTile",
        declutter: true,
        source: {
          type: "VectorTile",
          url: decodeURI(url.toString()),
          format: "MVT",
        },
        style: patternStyle,
      };
```

The only thing that I could not get to work in comparison with the `CanvasPattern` usage, is the transform on the actual pattern, using a `DOMMatrix`. While this can worked around by modifying the path instructions, scaling is not easily achievable without introducing artifacts (a canvas with 30x15 pixels already introduces artifacts at the tile boundaries).

ahocevar:
Thanks for the feedback, @StefanBrand. I need to [coordinate with @jahow](https://github.com/openlayers/openlayers/pull/15317#discussion_r1391362958), but maybe it makes sense to introduce a `fill-pattern-transform` property. If not: you could rotate and scale the context you draw to, then you don't need to change the paths. Don't you also see artefacts at tile boundaries when using a `CanvasPattern` with a transform matrix applied`? Pattern images should always have a width and height that matches a power of two, to avoid these artefacts.

mike-000:
Compared alongside perfected aligned WebGL rendering, when the pixel ratio is not 1 a canvas pattern is slightly misaligned when the geometry, position and resolution is set so it should be perfectly aligned.  Unlikely to be noticeable by a user but could be inconvenient in rendering tests https://jsfiddle.net/L4359chw/

StefanBrand:
I'm getting a weird traceback and artifacts on text when used in combination with the pattern:
```js
Uncaught TypeError: i is null
    re main-fb82107f.js:1040
    setTextStyle main-fb82107f.js:15348
    bD main-fb82107f.js:17526
    wD main-fb82107f.js:17358
    Gf main-fb82107f.js:17337
    renderFeature main-fb82107f.js:30093
    N main-fb82107f.js:29792
    updateExecutorGroup_ main-fb82107f.js:29805
    prepareTile main-fb82107f.js:29721
    getTile main-fb82107f.js:29732
    renderFrame main-fb82107f.js:29314
    render main-fb82107f.js:6316
    renderFrame main-fb82107f.js:6787
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    renderFrame_ main-fb82107f.js:9730
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    handleTileChange_ main-fb82107f.js:9551
    handleTileChange main-fb82107f.js:7563
    dispatchEvent main-fb82107f.js:837
    changed main-fb82107f.js:23388
    handleImageLoad_ main-fb82107f.js:23554
    node_modules chunk-vendors.js:59826
    promise callback*eO main-fb82107f.js:10806
    load main-fb82107f.js:23594
    loadMoreTiles main-fb82107f.js:7574
    handlePostRender main-fb82107f.js:9464
    node_modules chunk-vendors.js:58757
    setTimeout handler*renderFrame_ main-fb82107f.js:9737
    animationDelay_ main-fb82107f.js:9600
    render main-fb82107f.js:9624
    handleSizeChanged_ main-fb82107f.js:9483
    dispatchEvent main-fb82107f.js:837
    notify main-fb82107f.js:1103
    set main-fb82107f.js:1132
    setSize main-fb82107f.js:9758
    updateSize main-fb82107f.js:9805
    handleTargetChanged_ main-fb82107f.js:9545
    dispatchEvent main-fb82107f.js:837
    notify main-fb82107f.js:1103
    set main-fb82107f.js:1132
    setTarget main-fb82107f.js:9768
    firstUpdated main-fb82107f.js:84936
    _$AE main-fb82107f.js:267
    performUpdate main-fb82107f.js:258
    scheduleUpdate main-fb82107f.js:231
```


image

This is the style that should be applied (black text):

```js
      const defaultStyle = {
        "stroke-width": 0.3,
        "stroke-color": "black",
        "text-value": "foobar",
        "text-font": "14px sans-serif",
        "text-fill-color": "black",
        "text-stroke-color": "#fff",
        "text-stroke-width": 3,
        "text-overflow": true,
      };

      const patternStyle = {
        ...defaultStyle,
        "stroke-color": "yellow",
        "stroke-width": 3,
        "fill-pattern-src": cnv.toDataURL("png"),
      };
```

src/ol/render/canvas/style.js
```diff
@@ -298,11 +298,16 @@ export function buildStyle(flatStyle, context) {
...
  */
 function buildFill(flatStyle, prefix, context) {
-  const evaluateColor = colorLikeEvaluator(
-    flatStyle,
-    prefix + 'fill-color',
-    context
-  );
+  let evaluateColor;
+  if ('fill-pattern-src' in flatStyle) {
+    evaluateColor = patternEvaluator(flatStyle, 'fill-', context);
```
StefanBrand, line 303: I think this condition is also met when the text style is evaluated in 

https://github.com/openlayers/openlayers/blob/f757d988bea49cbd3e1296510f2e48a10d8fdb5d/src/ol/render/canvas/style.js#L454

This might be the cause for my issue reported in https://github.com/openlayers/openlayers/pull/15318#issuecomment-1810244483

src/ol/render/canvas/style.js
```diff
@@ -298,11 +298,16 @@ export function buildStyle(flatStyle, context) {
  * @return {FillEvaluator?} A function that evaluates to a fill.
  */
 function buildFill(flatStyle, prefix, context) {
-  const evaluateColor = colorLikeEvaluator(
-    flatStyle,
-    prefix + 'fill-color',
-    context
-  );
+  let evaluateColor;
+  if ('fill-pattern-src' in flatStyle) {
```
StefanBrand, line 302: This fixes the error and the artifacts that appear when combining pattern and text:

```suggestion
  if (
    prefix !== 'text-' &&
    prefix !== 'text-background-' &&
    'fill-pattern-src' in flatStyle
  ) {
```

image

ahocevar:
The problems mentioned above should now be fixed.

src/ol/colorlike.js
```diff
@@ -16,13 +31,70 @@ import {toString} from './color.js';
...
  * @api
  */
 export function asColorLike(color) {
+  if (!color) {
+    return null;
+  }
+  if (typeof color === 'object' && 'src' in color) {
+    return asCanvasPattern(color);
+  }
   if (Array.isArray(color)) {
```
tschaub, line 45: Not critical, but I would probably do this `Array.isArray` check first.

src/ol/colorlike.js
```diff
@@ -16,13 +31,70 @@ import {toString} from './color.js';
...
+ * PatternDescriptor was not found in the icon image cache.
+ */
+function asCanvasPattern(pattern) {
+  if (pattern.offset && pattern.size) {
+    const cacheKey = pattern.src + ':' + pattern.offset;
+    let canvasPattern = iconCache.getPattern(
+      cacheKey,
+      undefined,
+      pattern.color
+    );
```
tschaub, line 63: I'm not yet understanding this logic.  If `iconCache.getPattern(cacheKey, undefined, pattern.color)` is truthy, then we return `iconCache.getPattern(pattern.src, 'anonymous', pattern.color)`.

tschaub: Ah, I think I misread the diff.

Inverting some of the conditions and returning early when possible
```js
/**
 * @param {PatternDescriptor} pattern Pattern descriptor.
 * @return {CanvasPattern|null} Canvas pattern or null if the pattern referenced in the
 * PatternDescriptor was not found in the icon image cache.
 */
function asCanvasPattern(pattern) {
  if (!pattern.offset || !pattern.size) {
    return iconCache.getPattern(pattern.src, 'anonymous', pattern.color);
  }

  const cacheKey = pattern.src + ':' + pattern.offset;
  const canvasPattern = iconCache.getPattern(
    cacheKey,
    undefined,
    pattern.color
  );

  if (canvasPattern) {
    return canvasPattern;
  }

  const iconImage = iconCache.get(pattern.src, 'anonymous', null);
  if (iconImage.getImageState() !== ImageState.LOADED) {
    return null;
  }

  const patternCanvasContext = createCanvasContext2D(
    pattern.size[0],
    pattern.size[1]
  );
  patternCanvasContext.drawImage(
    iconImage.getImage(1),
    pattern.offset[0],
    pattern.offset[1],
    pattern.size[0],
    pattern.size[1],
    0,
    0,
    pattern.size[0],
    pattern.size[1]
  );
  getIconImage(
    patternCanvasContext.canvas,
    cacheKey,
    undefined,
    ImageState.LOADED,
    pattern.color,
    true
  );

  return iconCache.getPattern(cacheKey, undefined, pattern.color);
}
```

I think this is the same logic - just easier (for me) to make sense of things with shorter `if` blocks.

ahocevar: The former is a cached colored sub-image from a sprite sheet (`pattern.offset` and `pattern.size` are specified - see `if (...` in line 57 with `return ...` in line 97). The latter is a cached colored image directly from the URL specified in `pattern.src` (`return ...` in line 99).

ahocevar:
Thanks @tschaub for the suggestions, I added a commit which addresses them.

ahocevar:
Thanks for the thorough review, @tschaub!

PR #15336, opened 2023-11-15
Add information whether snapped to line extremity

manisandro:
This PR adds the information whether a snapped vertex is a line extremity to the snap result.


manisandro:
Needs to be reworked.

PR #15338, opened 2023-11-15
More specific return types in Format#readFeatures

### LINKED ISSUE #15334: OL8.2.0 GeoJSON readFeatures returns FeatureLike
In OpenLayers 7 GeoJSON readFeatures returns Feature[] and these can be added to a VectorSource.

In OpenLayers 8.2 a FeatureLike[] is returned and that can be both Feature and RenderFeature. 

We have to work around that in all places that need Feature (and not RenderFeature) in our project with:
const features: Feature[] = new GeoJSON().readFeatures(json).filter(feature => feature instanceof Feature) as Feature[];

Is this intended? Would it be simpler to use a specific RenderFeature reader?

ahocevar:
This pull request improves the types of the GeoJSON and MVT format so they can infer the return type of `readFeatures()` from the `featureClass` constructor option. This means that no type casts are needed any more in the code snippets below:
```js
/** @type {Array<Feature>} */
const features = new GeoJSON().readFeatures(source);

/** @type {Array<RenderFeature>} */
const renderFeatures = new GeoJSON({featureClass: RenderFeature}).readFeatures(source);
```

For all other formats, the `Feature` return type is used for `readFeatures()`, as before v8.2.0.

Fixes #15334.

### DIFF
--- a/src/ol/format/Feature.js
+++ b/src/ol/format/Feature.js
@@ -94,6 +94,11 @@ import {
  * @property {Object<string, *>} [properties] Properties.
  */
 
+/***
+ * @template {import("../Feature.js").FeatureClass} T
+ * @typedef {T extends typeof import("../render/Feature.js").default ? import("../render/Feature.js").default : import("../Feature.js").default} FeatureOrRenderFeature<T>
+ */
+
 /**
  * @classdesc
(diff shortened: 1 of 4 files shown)


tschaub (approved):
Nice!

PR #15340, opened 2023-11-16
WebGL / Fix hit detection for vector layers

jahow:
Hit detection for the `WebGLVectorLayerRenderer` was broken by #15121 and somehow we missed a test for it, which means it sadly went undetected.

This PR adds a test that is initially failing, as long as a fix which was simply a missing `prop` prefix for the features hit detection color.

Note that this problem was also causing rendering artifacts on some browsers and platforms; these should now be gone.

Addresses the issues mentioned in https://github.com/openlayers/openlayers/discussions/14446#discussioncomment-7578596 and https://github.com/openlayers/openlayers/pull/15279#issuecomment-1783888157

### DIFF
--- a/src/ol/renderer/webgl/PointsLayer.js
+++ b/src/ol/renderer/webgl/PointsLayer.js
@@ -196,7 +196,7 @@ class WebGLPointsLayerRenderer extends WebGLLayerRenderer {
 
     if (this.hitDetectionEnabled_) {
       this.attributes.push({
-        name: 'a_hitColor',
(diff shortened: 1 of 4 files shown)


tschaub (approved):

daboudii:
Hi @jahow, I'm working with ol@9.0.0-dev.1701030855118 and I still have the rendering artifacts on firefox. 
Here is a video to show how it looks like: https://vimeo.com/889054497

PR #15344, opened 2023-11-19
Create css matrix without using the DOM

ahocevar:
This pull request fixes a performance problem that occurs e.g. when the view is rotated or the browser viewport is scaled (by css or cmd/ctrl and +/- keys): in this case it can happen that the canvas layer renderer is unable to reuse the container of the previous layer, due to a css matrix that differs due to floating point inaccuracies.

Previously, we tried to work around this by creating the css transform matrix by applying the desired transform to a DOM element and using the returned css transform matrix. In this process, some rounding happened. It seems that modern browsers don't round the matrix parameters any more.

The change in this pull request changes things so no DOM element is used to create the css transform matrix. Instead, each element of the transform is rounded appropriately - the first 4 parameters which contain sin and cos values from the rotation to 6 decimal places, the last 2 parameters which represent the translation to half a pixel.

### DIFF
--- a/test/browser/spec/ol/renderer/canvas/Layer.test.js
+++ b/test/browser/spec/ol/renderer/canvas/Layer.test.js
@@ -1,0 +1,48 @@
+import Feature from '../../../../../../src/ol/Feature.js';
+import Map from '../../../../../../src/ol/Map.js';
+import TileLayer from '../../../../../../src/ol/layer/Tile.js';
+import VectorLayer from '../../../../../../src/ol/layer/Vector.js';
+import VectorSource from '../../../../../../src/ol/source/Vector.js';
+import View from '../../../../../../src/ol/View.js';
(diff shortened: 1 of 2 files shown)


test/browser/spec/ol/renderer/canvas/Layer.test.js
```diff
@@ -0,0 +1,48 @@
...
+                geometry: new Point([0, 0]),
+              }),
+            ],
+          }),
+        }),
+      ],
+      view: new View({
+        center: [0, 0],
+        zoom: 0,
+        rotation: 1e-8,
```
ahocevar, line 35: Rotation usually causes vector and tile layers not to be able to reuse the same container. This minimal invisible rotation is only used to show the problem.

tschaub (approved):

Issue #15363, opened 2023-11-22
WMTS Capabilities parser don't supports BoundingBox in Layer node


ltshb:
**Describe the bug**
Some WMTS server uses the `BoundingBox` instead of `WGS84‌Bounding‌Box` on the `Layer` node but the ol/format/WMTSCapabilites parser ignore `BoundingBox`. `BoundingBox` is valid as specified in the OGC standard.

**To Reproduce**

Here is an example of a server that uses `BoundingBox`

https://wmts.geo.bs.ch/1.0.0/WMTSCapabilities.xml

Using the answer of this server with the `ol/format/WMTSCapabilities`, the `BoundingBox` of the layers is then missing the parsed results.

**Expected behavior**
Also parse the `BoundingBox` as specified in the OGC Standard.

shimoncohen:
@ltshb Was this issue fully resolved? :)

ahocevar:
Yes, that should be fully resolved. Closing.

PR #15364, opened 2023-11-22
Fix WMTS Capabilities Layer.BoundingBox

ltshb:
In WMTS OGC standard, each Layer node might have `BoundingBox` children instead of the regular `WGS84BoundingBox`

This solve #15363


ahocevar (approved):
Thanks, @ltshb

PR #15366, opened 2023-11-23
Revert "Fix WMTS Capabilities Layer.BoundingBox"

ltshb:
Reverts openlayers/openlayers#15364

@ahocevar Sorry I just saw an issue with my PR, the `BoundingBox` element can contain a `crs` element that specify the CRS of the bounding box, see printscreen from the spec

image

The actual code to read the `BoundingBox` in the PR #15364 rely on the code to read `WGS84BoundingBox` that returns an array. For the `BoundingBox` we should not return an array but an object, containing the following keys:
- bbox (or extent): mandatory array
- crs: optional string
- dimensions: optional number

What do you think of my proposal for the object structure ? If you agree I would do another PR with a proper implementation.


ahocevar:
Whatever you prefer - that structure isn‘t being used anywhere else in the library anyway.

ahocevar (approved):

PR #15371, opened 2023-11-24
End drawing when appendCoordinates() hits maxPoints

### LINKED ISSUE #15367: Script crashes when on drawing LineString from existing coordinates
**Describe the bug**
Hello.
I try to draw a line that has only origin and destination. So I set the maxPoints : 2 (for options of the Draw class). 
But when I try to draw a line based on existing coordinates, I get the error ERROR TypeError: Cannot read properties of null (reading 'getGeometry'). 

ERROR TypeError: Cannot read properties of null (reading 'getGeometry')
    at Draw.modifyDrawing_ (Draw.js:1522:42)
    at Draw.appendCoordinates (Draw.js:1787:10)

**To Reproduce**
1. Go to this example https://codesandbox.io/p/sandbox/draw-and-modify-features-cxz3dk?file=%2Fmain.js
2. Replace the addInteractions function by the code below:

function addInteractions() {
  draw = new Draw({
    source: source,
    type: typeSelect.value,
    maxPoints: 2,
  });
  map.addInteraction(draw);
  snap = new Snap({source: source});
  map.addInteraction(snap);
  draw.appendCoordinates([
    [
        -12526802.69811565,
        5223970.689581122
    ],
    [
        -10985575.684457522,
        5247452.234244971
    ]
])
}

3. Select Geometry type - LineString
4. Open the browser console and observe errors.

**Expected behavior**
Script shouldn't crash

ahocevar:
Fixes #15367.

### DIFF
--- a/src/ol/interaction/Draw.js
+++ b/src/ol/interaction/Draw.js
@@ -1562,6 +1562,7 @@ class Draw extends PointerInteraction {
   /**
    * Add a new coordinate to the drawing.
    * @param {!PointCoordType} coordinate Coordinate
+   * @return {Feature<import("../geom/SimpleGeometry.js").default>} The sketch feature.
    * @private
    */
   addToDrawing_(coordinate) {
@@ -1600,8 +1601,9 @@ class Draw extends PointerInteraction {
     this.createOrUpdateSketchPoint_(coordinate.slice());
     this.updateSketchFeatures_();
     if (done) {
-      this.finishDrawing();
+      return this.finishDrawing();
     }
+    return this.sketchFeature_;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


tschaub (approved):

src/ol/interaction/Draw.js
```diff
@@ -1666,12 +1668,13 @@ class Draw extends PointerInteraction {
    * Stop drawing and add the sketch feature to the target layer.
    * The {@link module:ol/interaction/Draw~DrawEventType.DRAWEND} event is
    * dispatched before inserting the feature.
+   * @return {Feature<import("../geom/SimpleGeometry.js").default>|null} The drawn feature.
```
tschaub, line 1671: Hardly seems worth it, but we could add a note to the upgrade notes saying that `finishDrawing` now returns a feature.

PR #15379, opened 2023-11-27
Conditional return types for getFeatureById

ahocevar:
Fixes #15378

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -62,6 +62,11 @@ export class VectorSourceEvent extends Event {
 }
 
 /***
+ * @template {import("../Feature.js").FeatureLike} [T=import("../Feature.js").default]
+ * @typedef {T extends RenderFeature ? T|Array<T> : T} FeatureClassOrArrayOfRenderFeatures
+ */
+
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for any review.

src/ol/source/Vector.js
```diff
@@ -275,7 +280,7 @@ class VectorSource extends Source {
     /**
      * A lookup of features by id (the return from feature.getId()).
      * @private
-     * @type {!Object<string, FeatureClass|Array<RenderFeature>>}
+     * @type {!Object<string, import('../Feature.js').FeatureLike|Array<import('../Feature.js').FeatureLike>>}
```
jahow, line 283: I'm just curious why you didn't use `FeatureClassOrArrayOfRenderFeatures<FeatureClass>` here instead?

jahow (approved):
Thanks, interesting usage of conditional types!

src/ol/source/Vector.js
ahocevar: I had tried, but TypeScript didn't like it.

PR #15384, opened 2023-11-27
Heatmap dynamic radius and blur

bjornharrtell:
I'm trying to make radius and blur accept a function to be able to control these parameters similar to what is already possible for weight.

The proposed code change work for blur but not for radius. I've kept the `u_size` uniform in here until I can figure it out... it seems that for some reason `setSymbolSizeExpression`  will not produce expected result after using a varyings, i.e `.setSymbolSizeExpression('vec2(v_prop_size)')` for this PR. The result is invisible heatmap in the example.

I found that docs for `setSymbolColorExpression` mentions varyings but `setSymbolSizeExpression` but I'm not sure that is intentional. Looking at the `ShaderBuilder` code indicates that varyings are taken into consideration for `setSymbolSizeExpression` but I'm too much of a shader novice to be sure I understand it right or to debug.

### DIFF
--- a/src/ol/layer/Heatmap.js
+++ b/src/ol/layer/Heatmap.js
@@ -102,6 +102,20 @@ class Heatmap extends BaseVector {
       this.weightFunction_ = weight;
     }
 
+    const blur = options.blur ? options.blur : 15;
+    if (typeof blur === 'number') {
+      this.blurFunction_ = (feature) => blur;
+    } else {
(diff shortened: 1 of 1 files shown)


bjornharrtell:
@mike-000 that's a nice trick. On the weight function, in my experiments it did appear to reevaluate.. are you sure about that it's not really dynamic?

PR #15392, opened 2023-11-30
Fix WMTS Capabilities Layer.BoundingBox

ltshb:
In WMTS OGC standard, each Layer node might have one or more `BoundingBox` children instead of the regular `WGS84BoundingBox`. Those bounding box might have a `crs` attribute that specify the coordinate reference system of the box.

This solve https://github.com/openlayers/openlayers/issues/15363

### DIFF
--- a/test/browser/spec/ol/format/wmts/ogcsample.xml
+++ b/test/browser/spec/ol/format/wmts/ogcsample.xml
@@ -73,6 +73,14 @@ access interface to some TileMatrixSets</ows:Abstract>
 				<ows:LowerCorner>-180 -90</ows:LowerCorner>
 				<ows:UpperCorner>180 90</ows:UpperCorner>
 			</ows:WGS84BoundingBox>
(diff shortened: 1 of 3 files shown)


test/browser/spec/ol/format/wmts/ogcsample.xml
```diff
@@ -73,6 +73,14 @@ access interface to some TileMatrixSets</ows:Abstract>
 				<ows:LowerCorner>-180 -90</ows:LowerCorner>
 				<ows:UpperCorner>180 90</ows:UpperCorner>
 			</ows:WGS84BoundingBox>
+			<ows:BoundingBox crs="urn:ogc:def:crs:CRS::84">
```
ahocevar, line 76:
```suggestion
			<ows:BoundingBox crs="urn:ogc:def:crs:OGC:1.3:CRS84">
```

ahocevar (approved):
Thanks, @ltshb

Issue #7255, opened 2017-09-13
KML support for GroundOverlay and ScreenOverlay


raiyni:
The KML parser lacks support for https://developers.google.com/kml/documentation/kmlreference#groundoverlay and https://developers.google.com/kml/documentation/kmlreference#screenoverlay

GroundOverlay is an overlay image that moves with the map.

ScreenOverlay is an overlay image that stays fixed in view. 

It would be a job. Probably convert each one to a Control/Overlay but that could lead to export problems. Or a fixed image feature but that wouldn't solve the problem a screenoverlay creates.

bartvde:
#2941

raiyni:
Hmm, wonder why my search didn't return it. Will close this and work on that then since the PRA is already in place.

raiyni:
Wonder why my search didn't bring that up. Still leaves out ScreenOverlay but they are decently similar.

raiyni:
What would be the preferred method of doing this? 

Passing the map object into ol.format.KML.readFeatures so that an overlay/control can be attached? Or make it a second function call on the KML to return the overlay/control? Or something different like creating a new feature type so that it can be hidden with the layer and exported with KML export?

Skuriles:
Are there any news on this issue?

raiyni:
No immediate progress. Been occasionally working on it in my downtime.

skyval:
Are there any news on this issue?

raiyni:
I stopped progress once the transition to v5 started and in the meantime our mapping requirements at work have slowed down.

skyval:
Ok, I understand... I place image using ImageStatic and use coordinates from KML in imageExtent, but I can't rotate it. Is in there an any option to rotete ImageStatic?

eric-g-97477:
I would like to see this ticket reopened in the hopes that someone can resolve it.

mike-000:
@eric-g-97477  This should be possible with support for a transform matrix in projections as in #15402 making it much simpler https://jsfiddle.net/mthw1zjy/  than the method in https://github.com/openlayers/openlayers/issues/4949#issuecomment-495940972 which required integration with proj4.  All that would then be needed would be a `readGroundOverlay` method similar to `readCamera` using OpenLayers XML parsing to obtain the required properties of each overlay (name, west, south, east, north, href, rotation, drawOrder, color) to create a layer group of overlays.  

[...]

Issue #4949, opened 2016-03-02
ImageStatic and rotation?


ch08532:
First off, forgive me if this is not the appropriate place to ask questions about the API.

I was wondering if there is a way to rotate a ol.source.ImageStatic?  I am trying to support the KML GroundOverlay feature and part of the standard allows for rotation of the imageExtent.  I do not see this ability supported in ImageStatic.  Anybody know if this is possible?  Thank you.

ahocevar:
This is not directly supported, but it is possible: You can define a custom `imageLoadFunction` with the `ol.source.ImageStaticConstructor`, in which you can load the image, create a canvas, rotate the canvas, and use the canvas context's `drawImage()` method to place the image on the rotated canvas.

mike-000:
Another method is to use reprojection.  Ground overlays are intended to fit an equidistant cylindrical projection.  For the most accurate results the correct projection should be calculated so the image can be reprojected to the view projection.  If the image is rotated the projection can be customised to handle rotation.  Demo of that https://codepen.io/anon/pen/LomLYe derived from https://renenyffenegger.ch/notes/tools/Google-Earth/kml/index

garpenholm:
I also would appreciate a **rotate:** attribute in the ImageStatic module, like;

Just a suggestion:
```
var staticSource = new ol.source.ImageStatic({
                        url: 'images/myImage.png',
                        rotate: 16.35,
                        attributions: 'Copyright Me',
                        projection: myView.getProjection(),
                        imageExtent: myImageExtent
                });
```

mike-000:
> I also would appreciate a **rotate:** attribute in the ImageStatic module, like;
> 
> Just a suggestion:
> 
> ```
> var staticSource = new ol.source.ImageStatic({
>                         url: 'images/myImage.png',
>                         rotate: 16.35,
>                         attributions: 'Copyright Me',
>                         projection: myView.getProjection(),
>                         imageExtent: myImageExtent
>                 });
> ```

Copy the `rotateProjection` function from https://codesandbox.io/s/reprojection-image-z0uit

then use 

`                         projection: rotateProjection(myView.getProjection(), 16.35 * Math.PI/180, myImageExtent),
`

garpenholm:
> > I also would appreciate a **rotate:** attribute in the ImageStatic module, like;
> > Just a suggestion:
> > ```
> > var staticSource = new ol.source.ImageStatic({
> >                         url: 'images/myImage.png',
[...]

PR #15402, opened 2023-12-05
Handle rotated, skewed or flipped GeoTIFF tile grids

### LINKED ISSUE #15294: GeoTIFF layers do not support TIFFs with GeoTransform/rotation
**Describe the bug**
When trying to add a certain GeoTIFF to an OpenLayers map, it renders in an incorrect location. The GeoTIFF in question is https://unnerving-monsoon.ams3.cdn.digitaloceanspaces.com/Georgia/2023-10-30-06-47-55_UMBRA-04.tif. This is not reproduced in QGIS or other COG tools.

**To Reproduce**
1. Go to https://www.cogeo.org/map/ and use the URL above. The TIFF renders correctly, using Marblecutter.
2. Go to https://geotiffjs.github.io/cog-explorer/ and use the URL above. The TIFF renders in an incorrect location, using OpenLayers.

It can also be reproduced by adding the following source to an OpenLayers map:

```
const source = new GeoTIFF({
  normalize: true,
  bands: [1, 1, 1],
  sources: [
    {
      url: 'https://unnerving-monsoon.ams3.cdn.digitaloceanspaces.com/Georgia/2023-10-30-06-47-55_UMBRA-04.tif',
      bands: [1, 1, 1],
      max: 300,
    },
  ],
});
```

**Expected behavior**
The projection and extent is read and interpreted correctly.

### LINKED ISSUE #15370: GeoTIFF rotation
Hello,
I'm trying to load a TIF into OpenLayers but it doesn't seem to be rotating the image.
In QGIS it is displayed correctly.
The URL is: http://ftpcdd.cnig.es/Vuelos_2015_2016/Vuelos_2015/Baleares/1.VF/1.08_PNOA_L6_2015_BAL_25cm_VF_img8c_rgb_hu31/h50_0697_fot_002-0015.tif
The projection is EPSG:25831.

Example:
const sources = [new GeoTIFF({
   convertToRGB: 'auto',
   sources: [{
     url: 'https://ftpcdd.cnig.es/Vuelos_2022/Vuelos_2022/extremadura_2022/Sur/01.VF/01.08_PNOA_2022_EXT_S_22cm_VF_img8c_rgb_hu29/h50_0752_fot_02-3221_cog.tif',
   }, ],
})];

const layer = new TileLayer({
   sources: sources,
});

I attach a screenshot of how it is displayed in QGIS and OpenLayers.

all the best
QGIS
Ol

### LINKED ISSUE #15428: error if GeoTiffs origin is set to the lower left corner
**Describe the bug**
I create several GeoTiffs in python, where the origin was set to the lower left corner. The result in OpenLayers was, that all my cells had 0 as value.
When I retranspose my GeoTiff, so that the upper left corner is the origin, all is represented as expected.

**To Reproduce**
1. Download the test setup with a given GeoTiff: [tif_test.zip](https://github.com/openlayers/openlayers/files/13672169/tif_test.zip)
3. extract the content an cd into th directory
4. Install the npm packages: `npm i`
5. Start the vite server `npm start`
6. open a Browser and see the black image. If you open the console and click on a pixel, you will see the value

**Expected behavior**
The tif_layer should show data between 10-360 in greyscale, no matter how the origin is defined. At least as long as the pixel size is then set to (1,1) and not (1,-1)

mike-000:
Fixes #15294
Fixes #15370
Fixes #15428

Uses rotated/skewed/flipped variants of projections to handle GeoTIFF tile grids with rotation, skew or flip specified by ModelTransformation.

See https://deploy-preview-15402--ol-site.netlify.app/en/latest/examples/cog-modeltransformation.html

Uses a local alternative to https://github.com/geotiffjs/geotiff.js/pull/403 which could be removed if the replacement https://github.com/geotiffjs/geotiff.js/pull/413 (with corrected Y axis scale) is merged.  The code could be further simplified if https://github.com/geotiffjs/geotiff.js/pull/414 is merged.


mike-000:
Ready for review as I do not have any files small enough to use for a rendering test.  The Projection setMatrix method should work equally well for a rotated ImageStatic, etc. but I have not documented that as part of the API.

loganwilliams:
@mike-000 Thanks for this! I'm happy to provide more test files, though the smallest I have is ~80MB.

mike-000:
@loganwilliams There is a large selection of open data available via https://radiantearth.github.io/stac-browser/#/external/s3.us-west-2.amazonaws.com/umbra-open-data-catalog/stac/catalog.json typically around that size suitable for use in examples but too big to use as local test data.  I have now found a reduced version of the file used in the example at https://github.com/GeoTIFF/test-data which is under 4MB so might be suitable for another rendering test.

loganwilliams:
This is really great, thanks for your work on it @mike-000. Hope it gets merged soon.

src/ol/source/GeoTIFF.js
```diff
@@ -123,6 +125,49 @@ function getWorkerPool() {
   return workerPool;
 }
 
+/**
+ * Replacement for GeoTIFFImage.getResolution() to handle ModelTransformation.
+ * @param {GeoTIFFImage} image The image.
+ * @param {GeoTIFFImage} [referenceImage] The reference image.
+ * @return {Array<number>} The map x and y units per pixel.
+ */
+function getResolution(image, referenceImage) {
```
ahocevar, line 134: Wouldn't it make more sense to contribute this to geotiff.js?

src/ol/proj.js
```diff
@@ -467,6 +481,20 @@ export function getTransformFromProjections(
...
+    sourceMatrix = invertMatrix((sourceMatrix || createMatrix()).slice());
+    destinationMatrix = destinationMatrix || createMatrix();
+    const coordTransform = function (coordinate) {
+      return applyMatrix(
+        destinationMatrix,
+        transform(applyMatrix(sourceMatrix, coordinate.slice())),
+      );
+    };
+    transformFunc = createTransformFromCoordinateTransform(coordTransform);
+  }
```
ahocevar, line 497: I wouldn't bake matrix transforms into projections. It does not feel right. Instead, a matrix transform should be defined with custom transform functions using `addCoordinateTransforms()`, like e.g. shown in https://openlayers.org/en/latest/examples/wms-custom-proj.html.

ahocevar (commented):
The approach developed in this pull request is good. My suggestions are only about where parts of the implementation belong to, so more users can benefit from it.

src/ol/source/GeoTIFF.js
mike-000: I agree, if https://github.com/geotiffjs/geotiff.js/pull/413 and https://github.com/geotiffjs/geotiff.js/pull/414 are accepted and included in a release then these copies could be replaced by the real thing.

src/ol/proj.js
mike-000: The problem with that approach is transforms would need to be maintained between every matrix variation and every projection added (past and future) via proj4.  Given that this is a relatively unusual situation (although it could also be applied to KML ground overlays) I think setting an additional property on the projection and calculating the transforms dynamically as required would be easier.

ahocevar: When you create a transform function, you can mix in the matrix transform part from a separate function, so you can still handle matrix variations outside the once created transform function. Basically the same thing you're currently doing in src/ol/proj.js.

mike-000: The existing transform functions are indexed by projection codes - the matrixes are assigned to projection objects with those codes.  Doing it this way avoids adding complexity to the indexing of the primary transforms.

ahocevar: Can you point me to the code section where that happens?

mike-000: The new code calls `projection.setMatrix(matrix);` or `this.projection.setMatrix(matrix);` when determining the projection https://github.com/mike-000/openlayers/blob/7cc890b34747feb99173b974d2e6e143c88e0504/src/ol/source/GeoTIFF.js#L297
and if a matrix is detected when an override projection has been specified
https://github.com/mike-000/openlayers/blob/7cc890b34747feb99173b974d2e6e143c88e0504/src/ol/source/GeoTIFF.js#615

It is handled in existing reprojection code via `getTransform` and `transform` calls which need no changes as they are passed the projection objects (not codes) and therefore pass them on unchanged to `getTransformFromProjections`.

src/ol/source/GeoTIFF.js
ahocevar: Both are now merged and released.

mike-000:
@ahocevar  This is ready for further review.  Workarounds have been removed in favour of the latest geotiff.js methods.  The handling of matrixes in projections now uses entire transform functions (in series if necessary) instead of very inefficiently extracting a single coordinate transform from a projection transform function to use in a new transform function.  This has made the rendering noticeably faster.  In many cases (as in the two examples and current tests) the matrix transform is independent of projection transforms and could potentially be cached and/or used directly, e.g.

```
  if ((!sourceTransform || !destinationTransform) && !transformFunc) {
    return sourceTransform || destinationTransform;
  }
```

although I suspect there would be little further performance gain from this.

Matrix and projection transforms used in series will be needed for generic viewers such as attempted in https://radiantearth.github.io/stac-browser/#/external/s3.us-west-2.amazonaws.com/umbra-open-data-catalog/stac/2023/2023-02/2023-02-12/6d584f33-0489-47dd-9412-14d2c83532fc/6d584f33-0489-47dd-9412-14d2c83532fc.json where the rotated UTM is further transformed to 3857 as in  https://jsfiddle.net/w4qxs50k/ - a test is still needed for that (potentially using an ImageStatic as in https://github.com/openlayers/openlayers/issues/7255#issuecomment-1912131950).

src/ol/proj.js
```diff
@@ -463,11 +473,55 @@ export function getTransformFromProjections(
...
-  return transformFunc;
+  if (destinationMatrix) {
+    const matrix = destinationMatrix;
+    destinationTransform = createTransformFromCoordinateTransform(
+      function (coordinate) {
+        return applyMatrix(matrix, coordinate);
+      },
+    );
+  }
+  return (
```
mike-000, line 500: A single transform could be used directly here, but any performance gain from the extra code would be minimal
```
  if ((!sourceTransform || !destinationTransform) && !transformFunc) {
    return sourceTransform || destinationTransform;
  }
```

ahocevar (changes requested):
Thanks for your continued effort here, @mike-000. The one thing that still does not look right is that you're using `Projection` as container for the transform matrix, and the projection transform function to apply a transform matrix. Like I said already, that does not feel like the right place.

Tthinking about it more, my previous suggestion to use transform functions directly does not feel right.

The `GeoTIFF` source looks like the right place to maintain a mapping image -> matrix, and to make use of it in an overridden `getTile()` method.

mike-000:
Separating matrix and projection transforms has the potential for double reprojection.  I would regard the image as being in a rotated version of a standard projection - in the example case a UTM projection - defined by a standard projection code and a rotation matrix, so to render it in for example EPSG:3857 only one reprojection should be needed (no reprojection of reproj tiles) the transform function being a combination of the UTM to 3857 transform and the matrix transform which can be easily done by `getTransformFromProjections` using only its current two parameters (and if source and destination codes are the same only the matrix is used).  Having the overall projection defined by a code in the projection and a matrix in the source would be difficult for the current renderer to manage.

Similarly the technique could be used for KML ground overlays, by regarding them as being in rotated/scaled version of EPSG:4326, they could be easily rendered in any view projection which has a transform to EPSG:4326.

But I do think setting a matrix on a projection object should not be exposed in the API, it should be done only inside the GeoTIFF source code (and potentially inside a ground overlay parser) to avoid any future breaking changes.

ahocevar:
I agree that a single transform of raster data is preferable. But still, this can be done in `ol/DataTile`, if the ReporjDataTile is configured with a transform and a projection.

mike-000:
A solution which is not specific to DataTile would be preferable, so it could be easily reused elsewhere, e.g. as part of a fix for #4949/#7255 where rotation transforms can be combined projection transforms without the need to maintain large numbers of individual transforms for multiple projection code/rotation permutations.

Also reprojection inevitably reduces the output quality, sometimes as in #15579 that can be unacceptable, so it would nice to allow the view to be in the tiles native projection and reproject any less important background or vector data.  Currently `getView()` returns a standard projection familiar to the user, and the fact that the source may have a non-standard projection with the same code is not documented, but for best quality that can be used as the view projection as in https://jsfiddle.net/jo9wav6y/ something I think should be retained and documented.

ahocevar:
@mike-000 To make it easier for me to review, please answer one question: in Projection, is it now either matrix or transform function, or both? If both, then how does that work exactly?

ahocevar:
@mike-000 Thanks for the explanation. I wish it was the other way, i.e. either numeric transform or matrix transform. Now that it's not like that, what exactly does 

> getTransformFromProjections considers the proj4 transform between the codes together with each of the transform matrixes (if present) to return a complete transform between any pair of projections.

mean?

I stell feel that in this case, registering transforms with transform functions (without `ol/proj/proj4`) that do whatever you do now to achieve the rotation, skew or flip or whatever.

mike-000:
> what exactly does
> 
> > getTransformFromProjections considers the proj4 transform between the codes together with each of the transform matrixes (if present) to return a complete transform between any pair of projections.
> 
> mean?

The process necessary to display a rotated or flipped UTM projection tile grid as in the example or in #15923 in a generic viewer such as  http://app.geotiff.io/ which uses a global projection (or in some cases the opposite as in https://jsfiddle.net/jo9wav6y/ to reproject a background such as OSM to the camera projection).  When the source is loaded we know its projection code and ModelTransformation, but we do not know the view projection and may need look up the source projection proj4 definition to obtain that part of the transform.  The adjustments for ModelTransformation do not need a lookup and can easily be added on demand.

ahocevar:
Well, like you're saying - the whole transform needed here is a "process", not a matter of projection. If you manage to change your pull request to do the transform with a chain of transform functions (i.e. a process) instead of trying to squeeze this into `Projection`, I'll review again.

mike-000:
So how do you propose to do that?  Since all methods which take a source projection and destination projection eventually call `getTransformFromProjections` that seems the obvious place to do it.  If the matrix is not included in the projection then each method which takes source and destination projection would also need source and destination matrix options.

ahocevar:
I can answer that once I fully understand what you're aiming for, which is not the case. So one more question: why are you saying the transform matrix is applied *after* reprojecting? I would think that if a GeoTIFF is rotated, skewed or flipped, that applying a transform matrix would be the *first* thing to do. Then we're done, unless the result is still in a different projection than the `View`. If so, a source -> view projection transform will be applied.

mike-000:
Yes, that is the problem situation.  Consider an OpenLayers version of https://app.geotiff.io/ where the code might include

```
register(proj4);
const map = new Map({
  target: 'map',
  layers: [
    new TileLayer({
      source: new OSM(),
    }),
  ],
  view: new View({
    center: [0, 0],
    zoom: 0,
  }),
});

// call this when URL is entered in form

function addGeoTIFF(url) {
  const cogSource = new GeoTIFF({
    sources: [{url: url}],
  });

  cogSource
    .getView()
    .then((viewConfig) =>
      fromEPSGCode(viewConfig.projection.getCode()).then(() => {
        map.getView().fit(
          transformExtent(
            viewConfig.extent,
            viewConfig.projection,
            map.getView().getProjection(),
          ),
        );
        map.addlayer(
          new TileLayer({
            source: cogSource,
          }),
        );
      })
    );
}
```
image

To avoid reprojecting a reprojection, which would affect output quality, the tiles should be reprojected directly from the rotated UTM projection to the view projection - the overall transform for doing that is obtained by combining the matrix and proj4 transforms (only the proj4 transform is cached, transform functions for a matrix are easy to create as required).

That is more complex than in the example where only the matrix transform is used to reproject the tiles (to an unrotated reprojection tile grid) and the proj4 transform is only needed to reproject the background OSM to UTM.

ahocevar:
Ok, so the matrix transform comes first, then the projection transform. So once you have a transform function that combines both, you don't have to create a transformed canvas from a transformed canvas. Try to think less in what we currently have (i.e. Projection), think more in what would make most sense.

For example, a renderer could get data from a source, obtain a transform matrix from the source, get the projection from the view, and then do what we currently call "reprojection" using a function that combines the matrix transform and the reprojection transform functions.

mike-000:
I would consider the projection with matrix to be a projection in its own right as a tile grid must be aligned with the source projection axes, so it does make sense to me.  And as noted above

> If the matrix is not included in the projection then each method which takes source and destination projection would also need source and destination matrix options.

so any other method is likely to be more complex to implement (especially for something which will be relatively uncommon) and if done at renderer level would not be reusable for other layer types (e.g. a KML GroundOverlay could be easily implemented as an ImageStatic in a scaled and rotated EPSG:4326 based projection and reprojected into any view projection with minimal additional effect).

Also it would be desirable to be able use the rotated source projection directly at times as in https://jsfiddle.net/jo9wav6y/ as any reprojection will affect output quality and in a rotated version of a COG such as in #15579 changing the pixel alignment would make the output even more meaningless than the unwanted resizing of pixels.

ahocevar:
Again, I'm not talking about multiple transforms on a canvas. Just chaining of transform functions. And my suggestion is to separate affine transforms (source matrix transform) from projections (source to view projection). An API for the user could look like this:
```js
new Source({
  projection: sourceProjection, 
  transform: transformMatrix, // transform data into source projection
});
```
For GeoTIFF, both can of course be determined from the GeoTIFF metadata.

mike-000:
Why make it unnecessarily complicated for the rendererers?  `getTransformFromProjections` is already chaining transforms, and it does that using the two existing parameters which can now contain both the code for the projection and the transform matrix.

ahocevar:
We should aim for making it easy for the users. I don't see where this would make it more complicated in the renderer code.

The problem with your approach is that projections work based on a projection code lookup. And with an arbitrary matrix in addition to a projection that can uniquely be identified by a code, we get into a situation where we need to set up a code (or a projection object) for each transform+projection combination.

mike-000:
Yes, we need new projection objects for custom projections - before each call to `setMatrix()` the projection on which is based needs to be cloned (`ol/source/GeoTIFF` does that).  I do not see any problem with that as long as `setMatrix()` is not exposed in the API, but maybe a non-API `createProjectionWithMatrix` method would be safer?

ahocevar:
I think I have made valid suggestions and also shown a way how these could be implemented. Please let's bury your selected approach of trying to squeeze all that into Projection objects.

mike-000:
Your proposal works well when the source is reprojected to another projection, e.g. UTM source with a rotation matrix reprojected to 3857.   The matrix needs to be passed to `ReprojDataTile` and on to `Triangulation` to be included in the `transformInv`.  Ideally it should also be passed to `calculateSourceExtentResolution` and `getPointResolution` in case the matrix has significant scaling.  But when attempting to use reprojection with the same projection (source with matrix and view without matrix) it starts getting messy.  A reprojection must be forced in `getTile()` if the source has a matrix, but if the same `getTile()` is called from `ReprojDataTile` it must return a source tile and not recursively create reproj tiles.  I managed that but still cannot get any output.  I suspect the need for separate caches, tile grid and gutters for source and reproj tiles, indexed by projection object uid, may be the problem.

ahocevar:
Maybe some things can be simplified along the way:

* Reprojected tiles in view projection can use a standard tile grid
* Reprojected tiles in view projection do not need gutters
* Tiles in view projection are cached by the renderer since #15905, so no need to keep a lookup of caches by projection in the source any more.

It sounds like that would resolve most or your issues.

mike-000:
Yes, I have it mostly working now, by taking the source transform matrix into account in `getTileGridForProjection`, `getTileCacheForProjection` and `getGutterForProjection`.  One of the test samples is not rendering the full extent, which I don't understand as the overall transform should be the same regardless of where the components are obtained from:

image

ahocevar:
> Yes, I have it mostly working now, by taking the source transform matrix into account in getTileGridForProjection, getTileCacheForProjection and getGutterForProjection.

Interesting. Would you mind sharing your code? I'm unable to come up with a reason why the transform would matter for the view projection.

mike-000:
Replaced by #15962/#15988

PR #15403, opened 2023-12-06
Update link to terrain-rgb in sea-level example

fredj:


ahocevar (approved):

mike-000:
There is also a newer tile url available `https://api.maptiler.com/tiles/terrain-rgb-v2/{z}/{x}/{y}.webp` with maxZoom 14 (also applies to https://openlayers.org/en/latest/examples/interpolation.html although increasing the maxZoom in that would make the unwanted interpolation harder to see).

tschaub (approved):

PR #14048, opened 2022-08-24
Clone the WMS params option

mike-000:
Fixes #14044

Clone the WMS params option instead of using it directly to ensure the object is updateable.

This broke two tests which were setting params by bypassing `updateParams` - does that justify an upgrade note?

PR #15410, opened 2023-12-07
Fix `updateParams` in ImageMapGuide and ImageArcGISRest

### LINKED ISSUE #15407: openlayers8.2.0 when loading mapguide layer. ImageMapGuide as the data source. Call the updateParams method to update the parameters. The data in the scene is not automatically updated.
openlayers8.2.0 when loading mapguide layer. ImageMapGuide as the data source. Call the updateParams method to update the parameters. The data in the scene is not automatically updated.

Tried 1.model.getSource().updateParams({time: Date.now()})

Tried 2.model.getSource().refresh()

Can not complete the refresh

Please give the correct way

thank you.

mike-000:
Fixes #15407

Ensure `updateParams` takes immediate effect, resetting the inherited `this.image` instead of the unused `this.image_` by making the code consistent with ImageWMS.

Also apply the #14048 changes for WMS to these and TileArcGISRest (reapplying to ImageWMS to prevent a regression).

### DIFF
--- a/src/ol/source/ImageArcGISRest.js
+++ b/src/ol/source/ImageArcGISRest.js
@@ -93,7 +93,7 @@ class ImageArcGISRest extends ImageSource {
      * @private
      * @type {!Object}
      */
-    this.params_ = options.params || {};
+    this.params_ = Object.assign({}, options.params);
 
     /**
      * @private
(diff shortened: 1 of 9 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #15412, opened 2023-12-08
Fix return type for getOverlayById

### LINKED ISSUE #15411: `getOverlayById` has an incorrect return type
**Describe the bug**
The return type of `getOverlayById` is defined as `Overlay` in the JSDoc while it may be `null` if an overlay with this id does not exist.

**To Reproduce**
Use OpenLayers with TypeScript and add a call to `getOverlayById`

**Expected behavior**
Return type should be `Overlay | null`

helgee:
Fixes #15411


ahocevar (approved):
Thanks, @helgee

PR #15414, opened 2023-12-08
Implement `geometry-type` operator in CPU

jahow:
This PR implements the `geometry-type` operator on the CPU compiler.

Part of #15413

### DIFF
--- a/src/ol/expr/cpu.js
+++ b/src/ol/expr/cpu.js
@@ -32,6 +32,7 @@ import {
  * @property {Object} variables The values for variables used in 'var' expressions.
  * @property {number} resolution The map resolution.
  * @property {string|number|null} featureId The feature id.
+ * @property {string} geometryType Geometry type of the current object.
  */
 
 /**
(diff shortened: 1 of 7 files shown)


src/ol/expr/cpu.js
```diff
@@ -43,6 +44,7 @@ export function newEvaluationContext() {
     properties: {},
     resolution: NaN,
     featureId: null,
+    geometryType: null,
```
tschaub, line 47: It looks like the code below uses `''` as the zero value for geometry type (in `computeGeometryType`).  I guess TypeScript doesn't mind, but I think we should either initialize this to `''` or add `null` to the list of accepted types and use `null` to mean unset/unknown.  (I think using `''` is fine for this.)

tschaub (approved):

src/ol/expr/cpu.js
jahow: I changed it to `''` for consistency

PR #15415, opened 2023-12-08
Allow partial expression evaluation on the CPU

jahow:
This PR adds support for evaluation expressions on the CPU with an incomplete context (e.g. no feature, no resolution, etc.). An `UNKNOWN_VALUE` constant is introduced; it will be returned whenever an expression evaluates to something that is not defined, for instance trying to read a property if no feature properties have been provided in the context.

This will be instrumental in preprocessing some parts of a style object for WebGL rendering (e.g. filters, string operations).

Rebased on https://github.com/openlayers/openlayers/pull/15414

Part of https://github.com/openlayers/openlayers/pull/15413


jahow:

PR #15416, opened 2023-12-08
WebGL / Filter geometries before processing them for rendering

jahow:
This PR introduces a `filter()` method on the `MixedGeometryBatch` class. Before generating WebGL buffers from a geometry batch, the `VectorStyleRenderer` will apply the style filter and exclude from the batch any geometry that does not match this filter. The style filter expression is evaluated on the CPU.

Some filters might not be evaluated at buffer generation time (e.g. a filter depending on `time` or `resolution` which are dynamic parameters). This entails that the filter is _also_ still applied on the GPU (fragment shader) to make sure that we filter our all relevant features at any given frame.

This change typically makes vector tiles processing a little bit longer, but is essential when rendering styles composed of dozens of different rules where it provides a huge performance boost.

Rebased on https://github.com/openlayers/openlayers/pull/15415

Part of https://github.com/openlayers/openlayers/pull/15413

### DIFF
--- a/src/ol/render/webgl/VectorStyleRenderer.js
+++ b/src/ol/render/webgl/VectorStyleRenderer.js
@@ -1,6 +1,12 @@
 /**
  * @module ol/render/webgl/VectorStyleRenderer
  */
+import {buildExpression, newEvaluationContext} from '../../expr/cpu.js';
+import {
+  BooleanType,
+  computeGeometryType,
+  newParsingContext,
+} from '../../expr/expression.js';
 import {
   create as createTransform,
   makeInverse as makeInverseTransform,
@@ -120,8 +126,9 @@ class VectorStyleRenderer {
    * @param {import('../../style/flat.js').StyleVariables} variables Style variables
    * @param {import('../../webgl/Helper.js').default} helper Helper
    * @param {boolean} [enableHitDetection] Whether to enable the hit detection (needs compatible shader)
+   * @param {import("../../expr/expression.js").ExpressionValue} [filter] Optional filter expression
    */
-  constructor(styleOrShaders, variables, helper, enableHitDetection) {
+  constructor(styleOrShaders, variables, helper, enableHitDetection, filter) {
     /**
      * @private
      * @type {import('../../webgl/Helper.js').default}
@@ -213,6 +220,15 @@ class VectorStyleRenderer {
        * @private
        */
       this.symbolFragmentShader_ = asShaders.builder.getSymbolFragmentShader();
+    }
+
+    /**
+     * @type {function(import('../../Feature.js').FeatureLike): boolean}
+     * @private
+     */
+    this.featureFilter_ = null;
+    if (filter) {
+      this.featureFilter_ = this.computeFeatureFilter(filter);
     }
 
     const hitDetectionAttributes = this.hitDetectionEnabled_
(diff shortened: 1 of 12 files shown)


src/ol/expr/cpu.js
```diff
@@ -236,17 +251,43 @@ function compileAccessorExpression(expression, context) {
...
+ * @param {function(Array): ReturnType} evaluator Final evaluator taking in the evaluated args
+ * @return {function(EvaluationContext):ReturnType} the evaluator function; if any arg evaluated to UNKNOWN_VALUE, will return UNKNOWN_VALUE
+ * @template ReturnType
+ */
+function checkForUnknown(argEvaluators, evaluator) {
+  return (context) => {
+    const evaluatedArgs = new Array(argEvaluators.length);
+    for (let i = 0, ii = evaluatedArgs.length; i < ii; i++) {
+      const value = argEvaluators[i](context);
+      if (value === UNKNOWN_VALUE) {
```
tschaub, line 282: I'm wondering if we will need to distinguish between unknown and `null`.  For example, a coalesce-type operator like the `'string'` operator returns the first string in an array that may contain `null`.  We wouldn't want the whole expression to be considered unknown in this case.

Basically, I'm wondering if we want unknown be a more reliable sentinel value:
```
const UNKNOWN = {};
```

src/ol/render/webgl/VectorStyleRenderer.js
```diff
@@ -166,6 +178,31 @@ class VectorStyleRenderer {
...
+    /**
+     * @type {function(import('../../Feature.js').FeatureLike): boolean}
+     * @private
+     */
+    this.featureFilter_ = null;
+    if (filter) {
+      const parsingContext = newParsingContext();
+      const compiled = buildExpression(filter, BooleanType, parsingContext);
+      this.featureFilter_ = (feature) => {
+        const evalContext = newEvaluationContext();
```
tschaub, line 190: This could be created once and then reused - as long as all properties are updated - to avoid a bit of unnecessary garbage creation.

src/ol/render/webgl/VectorStyleRenderer.js
```diff
@@ -166,6 +178,31 @@ class VectorStyleRenderer {
...
     }
 
+    /**
+     * @type {function(import('../../Feature.js').FeatureLike): boolean}
+     * @private
+     */
+    this.featureFilter_ = null;
+    if (filter) {
+      const parsingContext = newParsingContext();
+      const compiled = buildExpression(filter, BooleanType, parsingContext);
```
tschaub, line 188: This could be done separately, but I was envisioning that we would also add things like `hasFill`, `hasStroke`, and `hasPoint` (or some other name) to the parsing context.  This would allow us to filter out features based on the style properties being used.  For example, if a style doesn't have any stroke-related properties, then all the linestring geometries could be skipped.

src/ol/expr/cpu.js
jahow: Very good point! I didn't think of that

jahow: Addressed in #15415

src/ol/render/webgl/VectorStyleRenderer.js
jahow: I have tried doing this because it sounds like an easy optimization, but after running it though a performance analysis it turns out it doesn't really speed up the render, and on the other hand generates a lot of garbage.

The VectorStyleRenderer uses a mixed geometry batch where geometries are already distributed in three "buckets": polygons, lines and points. If the style doesn't contain any fill instruction, the polygon geometry bucket will simply not be processed. So for this part, filtering beforehand based on geometry type will not bring any improvement.

On the other hand, filtering a geometry batch before processing it takes time because we loop over very feature, and it produces garbage (this might be improved though).

I think I'll leave that out for now, unless you see another opportunity for improving things?

jahow:
I added some missing tests, still very much open to suggestions and ideas on this! thanks!!

jahow:
This is now rebased on https://github.com/openlayers/openlayers/pull/16524

I will try to setup a benchmark to have a better idea of what this optimization brings.

jahow:
Using [the vector tiles benchmark](https://openlayers.org/bench/cases/vector-tiles-rendering/?x=0&y=0&z=4&r=0&l=1) this shows performance tracking on a vector tiles layer with 200 styles.

Each frame is represented by one vertical slice in the graph at the bottom; the red threshold is the 16ms budget for each frame to stay above 60fps.

Before optimization (main branch):

image
Barely runs.


After optimization (this PR):

image
Still sluggish but at least somewhat useable.

URL of the benchmark project: https://github.com/openlayers/bench

src/ol/expr/cpu.js
```diff
@@ -20,30 +20,37 @@ import {ColorType, LiteralExpression, Ops, parse} from './expression.js';
...
- * @property {Object} variables The values for variables used in 'var' expressions.
- * @property {number} resolution The map resolution.
- * @property {string|number|null} featureId The feature id.
- * @property {string} geometryType Geometry type of the current object.
+ * Each of these values can be set to UNKNOWN_VALUE, which means that they are not known in the current context.
+ * @property {Object|UNKNOWN_VALUE} properties The values for properties used in 'get' expressions.
+ * @property {Object|UNKNOWN_VALUE} variables The values for variables used in 'var' expressions.
+ * @property {number|UNKNOWN_VALUE} resolution The map resolution.
+ * @property {string|number|UNKNOWN_VALUE} featureId The feature id.
+ * @property {string|UNKNOWN_VALUE} geometryType Geometry type of the current object.
```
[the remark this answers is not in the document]
tschaub: Actually, I have the same question about all of these. I better keep reading below...

src/ol/expr/cpu.js
```diff
@@ -20,30 +20,37 @@ import {ColorType, LiteralExpression, Ops, parse} from './expression.js';
  * value.  The evaluator function should do as little allocation and work as possible.
  */
 
+export const UNKNOWN_VALUE = {};
```
tschaub, line 23: Minor nit, I think `UNKNOWN` would sufficiently communicate what this is about (no need for `_VALUE`).

src/ol/expr/cpu.js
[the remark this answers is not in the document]
jahow: At evaluation time on the CPU these are set to whatever feature is currently being rendered; trying to evaluate an expression that depends on feature data without providing a proper feature would result in UNKNOWN

src/ol/expr/cpu.js
```diff
@@ -281,6 +296,26 @@ function compileAccessorExpression(expression, context) {
   }
 }
 
+/**
+ * @param {Array<ExpressionEvaluator>} argEvaluators Argument evaluators
+ * @param {function(Array): ReturnType} evaluator Final evaluator taking in the evaluated args
+ * @return {function(EvaluationContext):ReturnType|UNKNOWN} the evaluator function; if any arg evaluated to UNKNOWN, will return UNKNOWN
+ * @template ReturnType
+ */
+function checkForUnknown(argEvaluators, evaluator) {
```
jahow, line 305: reading this again after a while, I find this function very confusing; I'll try to rework this logic to make it less obtuse for a new eye

jahow:
OK I've been trying to improve this PR but I still feel that it brings too much complexity to the CPU expression evaluator, and this for something that brings _nothing_ to the canvas renderers (and might even slow them down).

I'll reopen a PR with a different approach and I am still very much open to suggestions!

Issue #15426, opened 2023-12-13
Add new Conversion operators 'number' to ExpressionValue


daboudii:
**Is your feature request related to a problem? Please describe.**
I'm trying to filter Features in a WebGL layer with ExpressionValue style, but at the moment I receive some value that should be a number but they are sent as string..., so I can't use the 'between' Logical operators.

**Describe the solution you'd like**
Add a new Conversion operators `number` that would just convert a string to a number.

Would that operator would make sense in ExpressionValue ?

jahow:
Sure, it would definitely make sense. Implementing new operators is always encouraged: this one would be callsed `to-number` and comply with this spec: https://maplibre.org/maplibre-style-spec/expressions/#types-to-number

I'm marking this as PR accepted, and you'd be very welcome to give it a try.

Also, it would be greatly appreciated if new operators could be implemented both in the CPU and GPU compilers. Thanks in advance!

PR #15427, opened 2023-12-13
Add support for to-number operator on ExpressionValue/gpu/cpu

daboudii:
#15426 

Here is a start for support to `to-number`.

I have an issue in the `gpu.test.js` where the expression is not taken into account, no matter what number I set it always expect 13. 
@jahow  I saw that you fixed `mix of var and get operators` that had a similar problem. Your help would be welcome for that. 

Still need to add support for : "If the input is null or false, the result is 0. If the input is true, the result is 1." form https://maplibre.org/maplibre-style-spec/expressions/#types-to-number

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "ol-stac": "^1.0.0-beta.6",
         "pixelmatch": "^5.1.0",
         "pngjs": "^7.0.0",
-        "proj4": "2.9.2",
+        "proj4": "2.10.0",
         "puppeteer": "21.7.0",
         "rollup": "^4.1.4",
         "rollup-plugin-external-globals": "^0.9.0",
@@ -7330,9 +7330,9 @@
       }
     },
     "node_modules/proj4": {
-      "version": "2.9.2",
-      "resolved": "https://registry.npmjs.org/proj4/-/proj4-2.9.2.tgz",
-      "integrity": "sha512-bdyfNmtlWjQN/rHEHEiqFvpTUHhuzDaeQ6Uu1G4sPGqk+Xkxae6ahh865fClJokSGPBmlDOQWWaO6465TCfv5Q==",
+      "version": "2.10.0",
+      "resolved": "https://registry.npmjs.org/proj4/-/proj4-2.10.0.tgz",
+      "integrity": "sha512-0eyB8h1PDoWxucnq88/EZqt7UZlvjhcfbXCcINpE7hqRN0iRPWE/4mXINGulNa/FAvK+Ie7F+l2OxH/0uKV36A==",
       "dev": true,
       "dependencies": {
         "mgrs": "1.0.0",
@@ -15660,9 +15660,9 @@
       "dev": true
     },
     "proj4": {
-      "version": "2.9.2",
-      "resolved": "https://registry.npmjs.org/proj4/-/proj4-2.9.2.tgz",
-      "integrity": "sha512-bdyfNmtlWjQN/rHEHEiqFvpTUHhuzDaeQ6Uu1G4sPGqk+Xkxae6ahh865fClJokSGPBmlDOQWWaO6465TCfv5Q==",
+      "version": "2.10.0",
+      "resolved": "https://registry.npmjs.org/proj4/-/proj4-2.10.0.tgz",
+      "integrity": "sha512-0eyB8h1PDoWxucnq88/EZqt7UZlvjhcfbXCcINpE7hqRN0iRPWE/4mXINGulNa/FAvK+Ie7F+l2OxH/0uKV36A==",
       "dev": true,
       "requires": {
         "mgrs": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "ol-stac": "^1.0.0-beta.6",
     "pixelmatch": "^5.1.0",
     "pngjs": "^7.0.0",
-    "proj4": "2.9.2",
+    "proj4": "2.10.0",
     "puppeteer": "21.7.0",
     "rollup": "^4.1.4",
     "rollup-plugin-external-globals": "^0.9.0",


src/ol/expr/gpu.js
```diff
@@ -436,6 +436,7 @@ ${ifBlocks}
     const compiledIndex = compile(index, NumberType, context);
     return `texture2D(${paletteName}, vec2((${compiledIndex} + 0.5) / ${numColors}.0, 0.5))`;
   },
+  [Ops.ToNumber]: createCompiler(([value]) => value),
```
jahow, line 439: Ok so this is going to be more involved I'm afraid. A compiled string in GLSL is turned into a number using `getStringNumberEquivalent()`. So here, since `value` is the compiled argument, it will most likely be a arbitrary number which has nothing to the input string.

Essentially, the GPU compiler cannot do string manipulation, and that includes conversion.

I attempted to a solution in [this commit](https://github.com/openlayers/openlayers/pull/15413/commits/89fb58d7fe223bfa25c6c59c7166da87fe1a2572), but it's not done yet. I'm actively working on this topic so, so hopefully in the near future string manipulations will become easier in the GPU compiler.

daboudii: Can I skip the `gpu.js` part, add a TODO comment and just merge the `expression` and `cpu` ?

daboudii:
Will working on that, I spotted an error in `function parseArgsOfType(argType)` 

I tested this style : `'circle-opacity': ['-', '1.0', ['*', animRatio, 0.75]]`  and got this error `Unexpected type for argument 0 of - operation, got number but expected string` `number` and `string` should switch places.

Should I fixe that in this merge or create a new one ?

daboudii:
Also can I try to add the `between` operator in `cpu.js` with this merge ?

fredj:
> Still need to add support for : "If the input is null or false, the result is 0. If the input is true, the result is 1." form https://maplibre.org/maplibre-style-spec/expressions/#types-to-number

You code should already works with `null`, `false` and `true`:
```
Number(null) === 0
Number(false) === 0
Number(true) === 1
```

daboudii:
> > Still need to add support for : "If the input is null or false, the result is 0. If the input is true, the result is 1." form https://maplibre.org/maplibre-style-spec/expressions/#types-to-number
> 
> You code should already works with `null`, `false` and `true`:
> 
> ```
> Number(null) === 0
> Number(false) === 0
> Number(true) === 1
> ```

Yes I realize it soon after, thank you. I was still missing `BooleanType`    as `parseArgsOfType` and a test

src/ol/expr/gpu.js
daboudii: @jahow would it be ok ? Also I added 2 questions :)

jahow: > Can I skip the `gpu.js` part, add a TODO comment and just merge the `expression` and `cpu` ?

That seems reasonable, yes; string manipulations in the GPU compiler will come later

jahow:
> Also can I try to add the `between` operator in `cpu.js` with this merge ?

Yes please; ideally your PR should end up with two commits (one for `to-number` and one for `between`), can you achieve that by rebasing/combining commits? thanks!

daboudii:
> Yes please; ideally your PR should end up with two commits (one for `to-number` and one for `between`), can you achieve that by rebasing/combining commits? thanks!

I manage to rebase the commits and added `between`

src/ol/expr/cpu.js
```diff
@@ -279,6 +283,14 @@ function compileComparisonExpression(expression, context) {
     case Ops.GreaterThanOrEqualTo: {
       return (context) => left(context) >= right(context);
     }
+    case Ops.Between: {
+      const min = compileExpression(expression.args[1], context);
```
jahow, line 287:
```suggestion
      const min = left;
```

src/ol/expr/expression.js
```diff
@@ -102,6 +102,7 @@ import {isStringColor} from '../color.js';
  *     (e.g. `'#86A136'`), colors using the rgba[a] functional notation (e.g. `'rgb(134, 161, 54)'` or `'rgba(134, 161, 54, 1)'`),
  *     named colors (e.g. `'red'`), or array literals with 3 ([r, g, b]) or 4 ([r, g, b, a]) values (with r, g, and b
  *     in the 0-255 range and a in the 0-1 range) (WebGL only).
+ *   * `['to-number', value1]` converts `value1` to a `number`. (Flat only)
```
jahow, line 105: I just realized that the `to-number` operator in Maplibre [takes multiple fallback values as well](https://maplibre.org/maplibre-style-spec/expressions/#types-to-number). This is what the documentation should look like:

```suggestion
 *   * `['to-number', value, fallback1, fallback2, ...]` converts `value` to a number. If the input is null or false, the result is 0. If the input is true, the result is 1. If the input is a string, it is converted to a number if possible. If multiple values are provided, each one is evaluated in order until the first successful conversion is obtained. If none of the inputs can be converted, an error is thrown. (Canvas only)
```

src/ol/expr/cpu.js
```diff
@@ -570,3 +582,31 @@ function interpolateColor(base, value, input1, rgba1, input2, rgba2) {
...
+
+  const args = new Array(length);
+  for (let i = 0; i < length; ++i) {
+    args[i] = compileExpression(expression.args[i], context);
+  }
+  switch (op) {
+    case Ops.ToNumber: {
+      return (context) => {
+        if (isNaN(Number(args[0](context)))) {
+          throw new Error(`Parsing failure the value resulted in NaN`);
```
jahow, line 603:
```suggestion
          throw new Error(`Conversion to number failed`);
```

src/ol/expr/expression.js
```diff
@@ -1051,6 +1059,19 @@ function parsePaletteArgs(encoded, context) {
...
+/**
+ * @type ArgValidator
+ */
+function parseToNumberArgs(encoded, context, parsedArgs) {
+  const operation = encoded[0];
+  if (isNaN(Number(encoded[1]))) {
+    throw new Error(
+      `The passed argument for the ${operation} operation, resulted in NaN`
+    );
+  }
```
jahow, line 1071: Here you are checking the actual type of the encoded argument, so if it's an expression like `['get', 'id']` it will fail at compilation time.

You shouldn't need this new function, see my comment above.

src/ol/expr/expression.js
```diff
@@ -613,6 +615,12 @@ const parsers = {
     parseArgsOfType(NumberType)
   ),
   [Ops.Palette]: createParser(ColorType, withArgsCount(2, 2), parsePaletteArgs),
+  [Ops.ToNumber]: createParser(
+    NumberType,
+    withArgsCount(1, 1),
+    parseArgsOfType(StringType | BooleanType),
+    parseToNumberArgs
+  ),
```
jahow, line 623: This should be enough:

```suggestion
  [Ops.ToNumber]: createParser(
    NumberType,
    withArgsCount(1, Infinity),
    parseArgsOfType(StringType | BooleanType | NumberType)
  ),
```

Number arguments should be allowed since they can be useful in fallbacks.

jahow (changes requested):
Thanks for this contribution! I'm sorry for the long delay in the review.

The between operator looks fine. As for to-number, I think we should strive to be compliant with the Maplibre style spec when introducing new operators, which is the reason of most of my comments. I hope it'll make sense. Also I noticed you did a check for number conversion at compile time, but this is not appropriate as arguments can be any kind of expressions as well. Knowing at compile time if a conversion from string will succeed or fail is simply not possible.

PR #15432, opened 2023-12-17
Make VectorLayer#getFeatures(pixel) work with useGeographic()

ahocevar:
`VectorLayer#getFeatures(pixel)` fails when a user projection is set (e.g. with `useGeographic()`). This pull request fixes that.

### DIFF
--- a/src/ol/render/canvas/hitdetect.js
+++ b/src/ol/render/canvas/hitdetect.js
@@ -7,6 +7,11 @@ import {Icon} from '../../style.js';
 import {ascending} from '../../array.js';
 import {clamp} from '../../math.js';
 import {createCanvasContext2D} from '../../dom.js';
+import {
+  getTransformFromProjections,
+  getUserProjection,
+  toUserExtent,
+} from '../../proj.js';
 import {intersects} from '../../extent.js';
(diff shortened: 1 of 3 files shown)


src/ol/render/canvas/hitdetect.js
```diff
@@ -68,7 +77,7 @@ export function createHitDetectionImageData(
     for (let j = 0, jj = styles.length; j < jj; ++j) {
       const originalStyle = styles[j];
       const geometry = originalStyle.getGeometryFunction()(feature);
-      if (!geometry || !intersects(extent, geometry.getExtent())) {
+      if (!geometry || !intersects(userExtent, geometry.getExtent())) {
```
tschaub, line 80: Would it work to to add `projection` to the argument list (instead of changing `extent` to `userExtent` and adding `userTransform`)?  Then we could use `fromUserExtent(geometry.getExtent(), projection)` here?

ahocevar: Thanks for the good suggestion, @tschaub. Definitely looks better now!

tschaub (approved):

PR #15442, opened 2023-12-19
[WIP] Implement hit detection for WebGL Mapbox Vector Tiles

gberaudo:
This PR implements hit detection in the MVT WebGL renderer.
This is achieved by using the same technique has for the vector layer renderer, with some additionnal bits to handle the tiles.

This is an early PR (WIP) to share the current state of the work.

For testing, use http://localhost:8080/webgl-vector-tiles.html
This test has been modified to display the selected feature with a different styling, in a separate layer.

To investigate / fix:
- [ ] points does not seem to work;
- [ ] white countries polygons does not match anything (maybe OK: do the actually exist as features?);
- [ ] why only initial zoom level works?


gberaudo:
Hit detection framebuffer looks correct, whatever the zoom level:
image

image

jahow:
>     * [ ]  white countries polygons does not match anything (maybe OK: do the actually exist as features?);

In these tiles the countries are not polygons; the water bodies are.

jo-brue:
hey @gberaudo, thanks for your work first of all. I don't know if you're still working on this, but I implemented the changes in these commits and it works pretty good. At some point (after some zooming ) the hit detection does not work anymore. It does either return nothing or a wrong (old) feature. I already looked into the forEachFeatureAtCoordinate function but couldn't quite figure it out. I can see there that a ref is found but the tile does not have an entry for this ref...

 I guess this is more related to the way tiles are generated and features baked into the lookup texture?

Maybe you can give a hint where else to look, many thanks :)

Edit: okay just saw that this is already part of your checklist up top. Could for some reason this be wrong: frameState.coordinateToPixelTransform, which is used to calculate the pixel?

leweohlsen:
Hey! Thanks a lot @gberaudo for your work on this! I am instrested in bringing this forward and spent a bit of time to figure out what the problem was.

Turns out the hit detection cache isn't being invalidated between frames, causing `readPixel()` to return stale hit detection framebuffer data on every interaction after the first one. This broke hit detection immediately after the initial hover/click and after any map changes like zooming or panning.

My proposed solution would be to call `clearCachedData` on the hit detection render target right after rendering the hit detection tiles. I've tested this and it works. However, as far as I see this makes the cache on the hit detection render target kind of useless. Maybe that's not a problem though because we need fresh hit detection data on every interaction.

On my branch, to test it out, I have also changed the example to be a more MVT map with country polygons only. I will post a PR shortly and rebase it on the current OL version.

PR #15450, opened 2023-12-27
Add Graticule rendering test

mike-000:
New test (picked from #14714) as the existing tests did not prevent #15076/#15077 confusing the view projection world extent with the graticule projection (EPSG:4326) extent (both use degrees but are otherwise unrelated) and breaking graticule rendering in the **Reprojection with EPSG.io Search** example while still passing all existing tests.


ahocevar:
What about a rendering test for a case that is not broken? If I look at the reference image you added, I feel like the graticule should continue beyond 60°W, no?

mike-000:
If the right side meridian label was in view it would be 65° 41' 24" W as defined by the view projection world extent (based on the EPSG.io bbox for the projection in the example).  But that would need more than a 256 x 256 rendering test:

image

It should not attempt to draw beyond that, which happened with #15076 when any of the suggested local projections were used in the example.

ahocevar:
Thanks for the explanation, @mike-000

ahocevar (approved):

Issue #15451, opened 2023-12-30
Features for extending support to integrated circuit design applications


sunilkpai:
**Is your feature request related to a problem? Please describe.**
I am intending to build a Vue/Vite app using openlayers for integrated chip feature rendering. I think many of openlayers' features already satisfy the needs in this field, and it appears to be a perfect fit for such an application, despite the fact that most applications appear to focus on geographic analysis. To this end, there are som changes that I think could go a long way to enabling chip rendering, and I am happy to aid in these tasks if this is viewed as a fruitful change / useful application.

If this isn't seen as useful / a good 

**Describe the solution you'd like**
Below are a couple of features from the top of my head that I think would be ideal for chip-scale approaches:

1. Allow ScaleLine resolution down to nanometers (currently microns is the lowest).
2. Add support for grid lines for flat-world projections (similar to graticule except for flat maps).

There are probably many more things, but these seem the most relevant / concrete.

ahocevar:
Thanks for getting in touch, @sunilkpai. 

> Allow ScaleLine resolution down to nanometers (currently microns is the lowest).

A pull request that adds nanometers would be appreciated.

> Add support for grid lines for flat-world projections (similar to graticule except for flat maps).

That should be supported already. If you feel something is missing or broken, a minimal reproduction example would help to understand.

sunilkpai:
> A pull request that adds nanometers would be appreciated.

Will get started on this, thank you!

> That should be supported already. If you feel something is missing or broken, a minimal reproduction example would help to understand.

Could you clarify? I tried using graticule, but the issue is that intervals are not represented in microns as desired. Is there some other solution you are hinting at here?

ahocevar:
Would you mind sharing your attempt with Graticule?

sunilkpai:
Yes, here it is: https://codesandbox.io/p/sandbox/graticule-forked-tkvlnq

ahocevar:
Ok, I see. You're absolutely right. The graticule currently always displays a grid in geographic coordinates. I agree it would make sense to make the Graticule's projection configurable. The default should be EPSG:4326, i.e. what it currently only supports. If you'd be able to create a pull request for that too, it would be appreciated. Thanks in advance!

sunilkpai:
I managed to change some args to get closer to the desired behavior using:

```javascript
intervals: [10000, 1000, 100, 10, 1, 0.1, 0.01, 0.001, 0.0001],
lonLabelFormatter: function (r) {
[...]

PR #15453, opened 2023-12-30
Add nanometer to scaleline

sunilkpai:
Addresses #15451 item 1, adding nm to scaleline and the appropriate test. Let me know if anything else is required.


ahocevar (approved):
Thanks, @sunilkpai

PR #15460, opened 2024-01-02
Updated linter config

tschaub:
This updates the linter config to use Prettier 3.  That change brings a change to trailing commas (see https://github.com/prettier/prettier/pull/11479), and the fixes are applied here.


.github/workflows/test.yml
```diff
@@ -19,17 +19,14 @@ jobs:
     name: Pre-Test
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
```
tschaub, line 23: This `strategy` should have been removed in #12248 when the `matrix` was removed.

PR #15465, opened 2024-01-05
Google maps source

### LINKED ISSUE #15461: Option to integrate Google Maps Tiles (using the new Map Tiles API recently released, no hacks or licensing issues)
Dear team,
I noticed that Google recently released an API to get Google Maps Tiles, using an API-Key and thus **without** hacks or other workarounds that directly access their servers, thus violating their license.

[Map Tiles API overview](https://developers.google.com/maps/documentation/tile/overview)

That is why I wanted to ask if it would be possible to integrate this service as is already done with Bing Maps Tiles and other providers.

From the documentation I understand that, once the API-Key is obtained, the procedure consists of a POST-type request that returns a SESSION_TOKEN, to be used in subsequent requests to the Tiles server.

[Satellite Tiles](https://developers.google.com/maps/documentation/tile/satellite)

```
curl -X POST -d '{
  "mapType": "satellite",
  "language": "de-DE",
  "region": "DE"
}' \
-H 'Content-Type: application/json' \
"https://tile.googleapis.com/v1/createSession?key=YOUR_API_KEY"
```

`curl "https://tile.googleapis.com/v1/2dtiles/z/x/y?session=YOUR_SESSION_TOKEN&key=YOUR_API_KEY"`

Thank you in advance.

J.

tschaub:
This adds a new source using the Google [Map Tiles API](https://developers.google.com/maps/documentation/tile/overview).

Here is what the example looks like when a valid API key is entered:

image

```js
import Google from 'ol/source/Google.js';
import Layer from 'ol/layer/Tile.js';
import Map from 'ol/Map.js';

const map = new Map({
  layers: [
    new Layer({
      source: new Google({
        key: 'Your Google Maps Key',
        scale: 'scaleFactor2x',
        highDpi: true,
      }),
    }),
  ],
  target: 'map',
  view: new View({
    center: [0, 0],
    zoom: 2,
  }),
});
```

Closes #15461.

### DIFF
--- a/config/tsconfig-strict.json
+++ b/config/tsconfig-strict.json
@@ -212,6 +212,7 @@
     "../src/ol/source/Cluster.js",
     "../src/ol/source/DataTile.js",
     "../src/ol/source/GeoTIFF.js",
+    "../src/ol/source/Google.js",
     "../src/ol/source/IIIF.js",
     "../src/ol/source/Image.js",
     "../src/ol/source/ImageArcGISRest.js",
--- a/examples/google.html
+++ b/examples/google.html
@@ -1,0 +1,24 @@
+---
+layout: example.html
+title: Google Maps
+shortdesc: Map with tiles from Google Maps
+docs: >
+  This example demonstrates how to display tiles from Google's Map Tiles API in a map.  To use the Google Map Tiles API,
+  you need to set up a Google Cloud project and create an API key for your application.  See the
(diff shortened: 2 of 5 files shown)


tschaub:
@openlayers/core - This is now ready for review.

I haven't had any response yet about getting an API key that we can use with this project.  I've updated the example to render a dialog that takes an API key before adding the source.  This may not be what we want long term, but it feels like a decent solution for now (without over complicating the example).

ahocevar (approved):
Thanks for your work on this, @tschaub. Looks great!

PR #15475, opened 2024-01-09
Use and recommend spatialreference.org for projection lookup

ahocevar:
Now that https://spatialreference.org/ is being [maintained by the PROJ OSGeo project](https://lists.osgeo.org/pipermail/proj/2024-January/011216.html) and has been updated to a static website built from the https://github.com/OSGeo/spatialreference.org/ repository,, I think it is time to switch back from epsg.io to spatialreference.org, both in our internal library use and in what we reference in examples and docs.

The key advantage is that the returned results from spatialreference.org are the same as  other projects that use PROJ, which - in the case of projections where grid transforms are available - means the returned definition will not use a grid transform. This is good because proj4.js requires additional configuration to work with grid transforms, and it allows us to get rid of the `fromEPSGLookupMaptiler()` function.

I also went through the examples and was able to simplify some of them.

### DIFF
--- a/examples/reprojection-by-code.html
+++ b/examples/reprojection-by-code.html
@@ -1,12 +1,12 @@
 ---
 layout: example.html
-title: Reprojection with EPSG.io Search
+title: Reprojection with EPSG search
 shortdesc: Demonstrates client-side raster reprojection of OSM to arbitrary projection
 docs: >
   This example shows client-side raster reprojection capabilities from
   OSM (EPSG:3857) to arbitrary projection by searching
-  in <a href="https://epsg.io/">EPSG.io</a> database.
-tags: "reprojection, projection, proj4js, epsg.io, graticule"
+  for EPSG codes on <a href="https://spatialreference.org/">spatialreference.org</a>.
+tags: "reprojection, projection, proj4js, spatialreference.org, graticule"
 resources:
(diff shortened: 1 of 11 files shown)


examples/reprojection-by-code.html
```diff
@@ -15,7 +15,7 @@
   <span class="col-auto">
     <span class="input-group">
       <label class="input-group-text" for="epsg-query">Search projection:&nbsp</label>
-      <input type="text" id="epsg-query" placeholder="4326, 27700, 3031, US National Atlas, Swiss, France, ..." class="form-control" size="50" />
+      <input type="text" id="epsg-query" placeholder="4326, 27700, 3031, ..." class="form-control" size="50" />
```
mike-000, line 18: Maybe replace names with the codes 9311, 2154, 21781
More significantly 9311 does not work with the definition returned by the new lookup.

ahocevar: @mike-000 What do you mean by

> 9311 does not work

I tried and the result looks correct to me.

ahocevar:
Unfortunately I have to close this for now. spatialreference.org does not include towgs84 information in both WKT and proj4 strings, and proj4.js cannot determine those properly from the given datum at this time.

I'm going to improve proj4.js first, and then reopen this pull request, or create a modified version.

PR #15482, opened 2024-01-11
Update examples codesandbox linter config to match #15460

mike-000:
Update setup to retain trailing commas when the CodeSandbox is saved


tschaub (approved):
Thanks, @mike-000

Issue #15404, opened 2023-12-06
Snap Interaction: dispatch event when not snapped anymore


kad-sidera:
**Is your feature request related to a problem? Please describe.**
When I have activated DrawInteraction and SnapInteraction, I want a different mouse cursor or draw styling when snapped to a geometry. The SnapInteraction only fires an event when snapped and doesn't fire an event when not snapped anymore. 
At this moment, it is possible to change the draw style when snapped, but it is not possible to change the draw style back because there is no way of knowing when not snapped anymore.

**Describe the solution you'd like**
Also fire an event from SnapInteraction when not snapped anymore. In `handleEvent` of SnapInteraction, it only dispatches an event when `snapTo` returns a result. It should also dispatch an event when `snapTo` doesn't return a result.
Would a Pull Request be accepted for this issue?

ahocevar:
@kad-sidera Good suggestion, pull request accepted. Thanks!

PR #15493, opened 2024-01-19
Add unsnap event

eklauberg:
Resolve issue: https://github.com/openlayers/openlayers/issues/15404


src/ol/interaction/Snap.js
```diff
@@ -306,6 +307,8 @@ class Snap extends PointerInteraction {
           segment: result.segment,
         }),
       );
+    } else {
+      this.dispatchEvent(new UnsnapEvent());
```
tschaub, line 311: It looks like this will dispatch an unsnap event on every event that the interaction handles (unless snapped).  So as you move your mouse around the map, with every mousemove, there will be an unsnap event.  Ideally this event would only be dispatched on the single event where the state changes from snapped to unsnapped.

PR #15494, opened 2024-01-20
Add unsnap event

eklauberg:
Resolve issue: https://github.com/openlayers/openlayers/issues/15404

### DIFF
--- a/src/ol/events/SnapEvent.js
+++ b/src/ol/events/SnapEvent.js
@@ -13,6 +13,7 @@ export const SnapEventType = {
    * @api
    */
   SNAP: 'snap',
+  UNSNAP: 'unsnap',
 };
 
 /**
@@ -54,5 +55,11 @@ export class SnapEvent extends Event {
      * @api
      */
     this.segment = options.segment;
+  }
+}
+
+export class UnsnapEvent extends Event {
+  constructor() {
+    super(SnapEventType.UNSNAP);
   }
 }
--- a/src/ol/events/SnapEvent.js
+++ b/src/ol/events/SnapEvent.js
@@ -57,9 +57,3 @@ export class SnapEvent extends Event {
     this.segment = options.segment;
   }
 }
-
-export class UnsnapEvent extends Event {
-  constructor() {
-    super(SnapEventType.UNSNAP);
-  }
-}
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -7,7 +7,7 @@ import PointerInteraction from './Pointer.js';
 import RBush from '../structs/RBush.js';
 import VectorEventType from '../source/VectorEventType.js';
 import {FALSE, TRUE} from '../functions.js';
-import {SnapEvent, SnapEventType} from '../events/SnapEvent.js';
+import {SnapEvent, SnapEventType, UnsnapEvent} from '../events/SnapEvent.js';
 import {boundingExtent, buffer, createEmpty} from '../extent.js';
 import {
   closestOnCircle,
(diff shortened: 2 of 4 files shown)


src/ol/events/SnapEvent.js
```diff
@@ -56,3 +57,9 @@ export class SnapEvent extends Event {
     this.segment = options.segment;
   }
 }
+
+export class UnsnapEvent extends Event {
+  constructor() {
+    super(SnapEventType.UNSNAP);
+  }
+}
```
ahocevar, line 65: This event is not necessary - the `SnapEvent` can be constructed with `SnapEventType.UNSNAP` as first argument.

ahocevar (changes requested):
Thanks for your contribution, @eklauberg.  See my in-line comments for suggested changes.

src/ol/interaction/Snap.js
```diff
@@ -306,6 +314,9 @@ class Snap extends PointerInteraction {
           segment: result.segment,
         }),
       );
+    } else if (this.snapped_) {
+      this.snapped_ = false;
+      this.dispatchEvent(new UnsnapEvent());
```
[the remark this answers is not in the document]
eklauberg: It seems that these fields (vertex, vertex Pixel, feature, and segment) are all null in the context you mentioned. Should I consider storing the last snap event to send in these fields if needed?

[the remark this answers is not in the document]
ahocevar: Yes, you could store them in the new `snapped_` member. To make the event more useful, I think you'd also have to fire an `unsnap` event when you're still snapped, but to a different vertex, feature or segment. For that, the stored properties would also help.

src/ol/interaction/Snap.js
```diff
@@ -210,6 +219,13 @@ class Snap extends PointerInteraction {
      */
     this.rBush_ = new RBush();
 
+    /**
+     * @type {SnappedInfo|null}
+     * @private
+     * @description Holds information about the last snapped state.
```
ahocevar, line 225:
```suggestion
     * Holds information about the last snapped state.
     * @type {SnappedInfo|null}
     * @private
```

src/ol/interaction/Snap.js
```diff
@@ -288,25 +304,55 @@ class Snap extends PointerInteraction {
...
+   *
+   * @param {SnappedInfo} data1 The first snap data set.
+   * @param {SnappedInfo} data2 The second snap data set.
+   * @return {boolean} `true` if the data sets are equal, otherwise `false`.
+   * @private
+   */
+  areSnapDataEqual_(data1, data2) {
+    return (
+      data1.vertex[1] === data2.vertex[1] &&
+      data1.feature?.getId() === data2.feature?.getId()
```
ahocevar, line 319: Features do not need an id, but the feature reference will be the same if the features are the same
```suggestion
      data1.feature === data2.feature
```

src/ol/interaction/Snap.js
```diff
@@ -288,25 +304,55 @@ class Snap extends PointerInteraction {
...
       evt.pixel = result.vertexPixel;
-      this.dispatchEvent(
-        new SnapEvent(SnapEventType.SNAP, {
+
+      // Dispatch UNSNAP event if already snapped
+      if (this.snapped_ && !this.areSnapDataEqual_(this.snapped_, result)) {
+        this.dispatchEvent(new SnapEvent(SnapEventType.UNSNAP, this.snapped_));
+      }
+
+      if (!this.snapped_ || !this.areSnapDataEqual_(this.snapped_, result)) {
```
ahocevar, line 340: Instead of calling `areSnapDataEqual_()` again here, you could store it in a variable above.

src/ol/interaction/Snap.js
```diff
@@ -288,25 +304,55 @@ class Snap extends PointerInteraction {
...
+   * Compares the Y-coordinate of the `vertex` and the feature ID.
+   *
+   * @param {SnappedInfo} data1 The first snap data set.
+   * @param {SnappedInfo} data2 The second snap data set.
+   * @return {boolean} `true` if the data sets are equal, otherwise `false`.
+   * @private
+   */
+  areSnapDataEqual_(data1, data2) {
+    return (
+      data1.vertex[1] === data2.vertex[1] &&
```
ahocevar, line 318: I think you'll want to check equality for the segment instead. I'd expect an `unsnap` event only when snapping leaves a segment, not whenever I move the cursor.
```suggestion
      data1.segment === data2.segment &&
```

ahocevar (changes requested):

MoonE:
Resolved by #16352

PR #15496, opened 2024-01-21
feat(ol/feature): allow generics for `feature.getProperties<T>()`

### LINKED ISSUE #14868: TS allow generics for `getProperties()`
As a user of Openlayers with TypeScript, I want to be able to receive `feature` properties by passing a type as generic.
This allows to explicitly set the type of feature properties for further processing.

_Current Behavior_:

```ts
const stop = feature?.getProperties() as Stop
```

_Solution i'd like_:

```ts
const stop = feature?.getProperties<Stop>()
```

d-koppenhagen:
with this change it's possible to pass a generic type `T` when calling `feature.getProperties<T>()` and using with TypeScript.

```
// before: const stop = feature?.getProperties() as Stop
const stop = feature?.getProperties<Stop>()
console.log(typeof stop) // "Stop"
```

closes #14868


ahocevar (commented):
Thanks for your contribution! However, I'm failing to understand the benefit of this change. However,  the result is no better than the type cast you initially had - you still have to make an assumption on what the return type is, and TypeScript won't help you checking it because of the `@ts-ignore`.

I don't think there is a good easy alternative. A more involved alternative would be to define a generic PropertyType on the Feature constructor, and allow setting it also on VectorSource and FeatureFormat. If set on FeatureFormat, VectorSource should pick it up. Doable, but not easy.

d-koppenhagen:
@ahocevar you were totally right. I was on the wrong path I think. Please review again. Now it should be possible to infer the type.

ahocevar:
@d-koppenhagen Are you able to benefit from this? With your last changes, it looks like you only would if you created `RenderFeature` instances manually, instead of retrieving them from a FeatureFormat or a VectorTile source.

Guillaumedemoff:
I believe the get function should also be typed accordingly.

Something like this should do the trick.
```
  /**
   * Get a feature property by its key.
   * @template {keyof T} Key
   * @param {Key} key Key
   * @return {T[Key]} Value for the requested key.
   * @api
   */
```

PR #15497, opened 2024-01-21
Fix #14954 - Add `willReadFrequently: true` to CanvasLayerRenderer context generation settings

### LINKED ISSUE #14954: Canvas2D warning
I am using openlayers v 7.4.0 when I refresh the page this warning occur : 
**the Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: <URL>**

** const L = h.getImageData(0, 0, a, a).data; ** 
this line off code had error but still work
if I did something wrong , let me know or please fix this .
thanks a lot

Menecats:
Fixes #14954 by adding `willReadFrequently` in context generation settings for CanvasLayerRenderer.

### DIFF
--- a/src/ol/renderer/canvas/Layer.js
+++ b/src/ol/renderer/canvas/Layer.js
@@ -164,7 +164,9 @@ class CanvasLayerRenderer extends LayerRenderer {
     ) {
       const canvas = target.firstElementChild;
       if (canvas instanceof HTMLCanvasElement) {
-        context = canvas.getContext('2d');
+        context = canvas.getContext('2d', {
+          willReadFrequently: true,
+        });
       }
     }
     if (context && context.canvas.style.transform === transform) {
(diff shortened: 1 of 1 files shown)


mike-000:
This will disable hardware acceleration and reduce rendering performance.  What are you doing to produce the warning?

Menecats:
Hi @mike-000, I'm getting this message when reading the canvas image data directly inside a `postrender` event handler (see following *partial* code).

I've tought of adding a configuration option that would allow the user to specify the `getContext` options directly but I was unsure of where to place it. (my general idea was: `new ImageLayer({ source: [...], contextSettings: {willReadFrequently: true})`)

My code that produces the warning:
```typescript
const layer = new ImageLayer({
  source: [...]
});

layer.on('prerender', e => {
  // Draw mask that will block part of the view

  e.context.globalCompositeOperation = 'source-in';
});

layer.on('postrender', e => {
  e.context.globalCompositeOperation = 'source-atop';

  // Tint part of the view

  const imageData = e.context.getImageData(0, 0, canvasWidth, canvasHeight);

  // Send the image data via presentation API to a chromecast or other low power casting device
});
```

ahocevar:
In this case the warning issued by the Developer Console is not appropriate. Because you create content on the canvas, read it, and then create new content and read it again. I don't think the `willReadFrequently: true` flag would improve the overall performance here. It would be interesting to benchmark this use case.

PR #15523, opened 2024-01-31
Change Geometry.transform return type to 'this'

fredj:
Otherwise the return type is `Geometry` for subclasses instead of the actual class.


ahocevar (approved):
Good find, thank @fredj!

mike-000:
It might be possible to remove typecasts, e.g.
https://github.com/openlayers/openlayers/blob/main/src/ol/interaction/Modify.js#L1589
https://github.com/openlayers/openlayers/blob/main/src/ol/interaction/Modify.js#L1631

PR #15524, opened 2024-01-31
Check OGC `orderedAxes` before defaulting to projection axis orientation

### LINKED ISSUE #15521: Projections other than WebMercator are not working on OGCVectorTile
**Describe the bug**

The [OGCVectorTile](https://openlayers.org/en/latest/apidoc/module-ol_source_OGCVectorTile-OGCVectorTile.html) class supports a parameter called `Projection`, which describes the projection of the tile grid. According to the documentation, it supports a ProjectionLike object, which defaults to 'EPSG:3857'. 

The issue is that 3857 is the only projection supported by the class. Using any other projection will cause the layer to fail to display, silently.

**To Reproduce**

 This works:
```
                  new VectorTileLayer({
                    title: 'Natural Earth',
                    visible: true,
                    source: new OGCVectorTile({
                      projection: 'EPSG:3857',
                      url: 'https://maps.gnosis.earth/ogcapi/collections/NaturalEarth:cultural:ne_10m_admin_0_countries/tiles/WebMercatorQuad',
                      format: new MVT({
                        dataProjection: 'EPSG:3857',
                        featureProjection: 'EPSG:3857'
                       }),
                    }),
                  })  
```
This also works (using EPSG:3857 as default):

`                  new VectorTileLayer({
                    title: 'Natural Earth',
                    visible: true,
                    source: new OGCVectorTile({
                      url: 'https://maps.gnosis.earth/ogcapi/collections/NaturalEarth:cultural:ne_10m_admin_0_countries/tiles/WebMercatorQuad',
                      format: new MVT({
                       }),
                    }),
                  }) `

However, this does not work:

`                  new VectorTileLayer({
                    title: 'Natural Earth',
                    visible: true,
                    source: new OGCVectorTile({
                      projection: 'EPSG:4326',
                      url: 'https://maps.gnosis.earth/ogcapi/collections/NaturalEarth:cultural:ne_10m_admin_0_countries/tiles/WorldCRS84Quad',
                      format: new MVT({
                        dataProjection: 'EPSG:4326',
                        featureProjection: 'EPSG:4326'
                       }),
                    }),`

According to the network inspector, a request to "https://maps.gnosis.earth/ogcapi/tileMatrixSets/WorldCRS84Quad" is issued and it receives a 200 response, but nothing happens after that (no tiles are requested).

**Expected behavior**
We should see a map like in the case of WebMercator tiles, but projected in WGS84.

mercator-map

mike-000:
Fixes #15521

Also adds a geographic vector tiles example
https://deploy-preview-15524--ol-site.netlify.app/en/latest/examples/ogc-vector-tiles-geographic.html

### DIFF
--- a/examples/ogc-vector-tiles-geographic.html
+++ b/examples/ogc-vector-tiles-geographic.html
@@ -1,0 +1,9 @@
+---
+layout: example.html
+title: OGC Vector Tiles (Geographic)
+shortdesc: Rendering vector tiles from an OGC API – Tiles service.
+docs: >
+  The <a href="https://ogcapi.ogc.org/tiles/">OGC API – Tiles</a> specification describes how a service can provide vector tiles.
+tags: "ogc, vector"
+---
+<div id="map" class="map"></div>
--- a/examples/ogc-vector-tiles-geographic.js
+++ b/examples/ogc-vector-tiles-geographic.js
@@ -1,0 +1,29 @@
+import MVT from '../src/ol/format/MVT.js';
+import Map from '../src/ol/Map.js';
+import OGCVectorTile from '../src/ol/source/OGCVectorTile.js';
+import VectorTileLayer from '../src/ol/layer/VectorTile.js';
+import View from '../src/ol/View.js';
+
+const map = new Map({
+  target: 'map',
+  layers: [
+    new VectorTileLayer({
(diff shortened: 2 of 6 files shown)


ahocevar (approved):
Thanks, @mike-000

PR #15554, opened 2024-02-14
WebGL / Improve line antialiasing

jahow:
Antialiasing is now done on the `+0.5 -0.5` range; previously it was done on the `+0 -1` range, which meant line width was always 1px less that expected.

Before (line of 1px width):
image

After:
image

This PR also contains a workaround for the webgl fill and line patterns that sometimes fail because the images might not be ready for rendering.


test/rendering/cases/webgl-fill-pattern/main.js
```diff
@@ -104,6 +104,9 @@ new Map({
   }),
 });
 
-render({
-  message: 'renders four polygons with various pattern fills',
+requestAnimationFrame(() => {
+  map.renderSync();
+  render({
+    message: 'renders four polygons with various pattern fills',
+  });
```
ahocevar, line 111: I'm wondering if it wouldn't be better to do something like this instead:
```js
map.once('rendercomplete', () => {
  render({
    message: 'renders four polygons with various pattern fills',
  });
});
```

ahocevar (approved):
Thanks, @jahow! You may want to consider my comment regarding `renderSync()` vs. `rendercomplete` event. Looks good otherwise.

test/rendering/cases/webgl-fill-pattern/main.js
jahow: Amended the first commit with this change

Issue #15542, opened 2024-02-08
ScaleLine Control reports incorrect distance when not on equator


weskamm:
The `ol/control/ScaleLine` does not report correct measurements when not used on the equator.
The differences in measurement become huge, a feature with 10000 meters length is reported to have ~6000m when drawn near england.
I prepared a jsfiddle  here:

https://jsfiddle.net/43ev0nfc/12

There are two features on the map, both are rectangles with 10km side length. One feature is at the equator, one between france and england. The scaleline reports correct values for the feature at equator:

![image](https://github.com/openlayers/openlayers/assets/1381363/184a8d90-4c7c-45e9-ae4e-2f929e9a2c4b)

But the feature near england looks like this:

![image](https://github.com/openlayers/openlayers/assets/1381363/4b58c8d3-bb0f-4af5-a2b1-27326841b655)

The documentation states that the measurement is not exact, but those differences are really huge.
I identified the `setGetPointResolution` function to be the cause / or solution to this problem.

In the fiddle you can uncomment the last line to override that function to just return the input value:

`map.getView().getProjection().setGetPointResolution(r => r);`

When using this override, the measurement becomes correct for the "england" polygon:

![image](https://github.com/openlayers/openlayers/assets/1381363/9e3305ef-3021-485c-a66c-bf9e25ca6e6a)

The measurement at equator keeps correct with this override.
So to me it seems we always have to override the default function to just return the same value that has been given as argument.

The issue behaves the same for different projections, i tried 4326, 3857, 25832, 31467.
Everything works as expected with the override from above.

Seems like a bug to me?

ahocevar:
I'm surprised to read you're having trouble with EPSG:4326 and EPSG:3857, because these should work fine out of the box, without any modifications to the Projection instance. Do you have code examples for both that show the problem?

For other projections, you need to set up the projections's units correctly, and for non-cartesian projections also the `getPointResolution` function.

weskamm:
I thought the same :-) In the fiddly from above there is a projection switcher at the bottm (which seem to work only if you use it at least twice).

I thought that at least for the default projections, users of the scaleline would not need to setup custom complicated functions for the scalebar to work correctly?

M393:
The two polygons are bother 10000 * 10000 units in EPSG:3857, but in vertical direction these are only meters at the equator. At England latitude its off by about `cos(50°) ~ 0.643`. The point resolution used by ScaleLine is the average of the lon and lat resolutions.
[...]

PR #15556, opened 2024-02-16
Add note about cartesian measurements on scaleline

weskamm:
This adds a short note to the scaleline about cartesian measurements.

For disucussion, see https://github.com/openlayers/openlayers/issues/15542


ahocevar (approved):
Thanks, @weskamm

PR #15568, opened 2024-02-19
New image tile source

tschaub:
This is a new version of the changes in #14262.  I needed to squash the commits to deal with all the conflicts in the past couple years.

This adds a new `ol/source/ImageTile.js` module that is intended to replace the following:

 * `ol/source/UrlTile.js`
 * `ol/source/TileImage.js`
 * `ol/source/XYZ.js`
 * `ol/ImageTile.js`
 * `ol/tileurlfunction.js`

The new source accepts a `url` option that can be a single URL template, an array of URL templates, or a function that returns a URL.  The resulting loader returns a promise that creates an image, sets its src, and resolves when the image loads. For more complex image loading scenarios, the source can be passed a `loader` option.

Since we have some breaking changes already merged in, I thought it could make sense to deprecate a few things and merge this in.

When we are ready to remove `ol/source/TileImage.js`, we can update the the 9 sources that extend `TileImage` to instead extend the `ImageTile` source.  This will be a breaking change since some of the options no longer make sense (`tileLoadFunction`, `urls`, `cacheSize`).


thomasmoelhave:
Maybe it's too early for feedback here, if so I apologize :).

It's exciting to see some change in this part of OpenLayers. I tried comparing the two "WMTS Tile Transistions" examples. 

[current version](https://openlayers.org/en/latest/examples/wmts-dimensions.html)
[this PR](https://deploy-preview-15568--ol-site.netlify.app/en/latest/examples/wmts-dimensions.html)

Here's a video showing difference. First the version in this MR, and second the version from openlayers.org. 

[Screencast from 05-14-2024 02:34:03 PM.webm](https://github.com/openlayers/openlayers/assets/179143/84c9e769-02f1-497f-a0c7-bfe5930836a2)

I feel the new version has some problems compared to the current version, in particular in terms of added flickering.

This example, introduced in #4389 was made primarily to test how the interim tiles work, and how we render the most recently received tiles while waiting for updated images corresponding to a dimension change.

tschaub:
Replaced by #15905.

PR #15569, opened 2024-02-19
Avoid circular ol <-> ol-mapbox-style dependency from old versions

ahocevar:
Fixes the broken map at the top of the website, which is a result of multiple versions of `ol` and `ol-mapbox-style` being used.


ahocevar:
This is what I did to get rid of the duplicate packages:

    npm uninstall ol-stac ol-mapbox-style
    npm install --save-dev ol-stac ol-mapbox-style

ahocevar:
Converted to draft, because the map at the top is still broken.

ahocevar:
The remaining problem was that ol-mapbox-style and ol-pmtiles still pull in `ol@8` for their `ol` peer dependency. Although this problem will go away the moment we publish v9, I added `@rollup/plugin-alias` for the builder of the map at the top of the website, so it uses the local version instead of a peer dependency ol package, which will likely be an older version.

tschaub:
Thanks for tracking this down and finding a fix, @ahocevar.

I'm starting to feel like `*` is the only sensible version to put in a peer dependency (for "plugin" type libraries where you want to avoid multiple copies of the upstream lib being used).

tschaub (approved):

PR #15570, opened 2024-02-19
Also defer rendering of image layers when decluttering is needed

ahocevar:
This is a follow-up on #15566 - to keep all rendering in the correct order, also image layers need to be rendered deferred when decluttering is done on any layer.


ahocevar:
The incorrect render order can be seen in https://deploy-preview-15566--ol-site.netlify.app/en/latest/examples/mapbox-style.html (after #15566), the fixed version after this pull request is https://deploy-preview-15570--ol-site.netlify.app/en/latest/examples/mapbox-style.html. The difference is the missing hillshade on the first map. It is rendered before the vector tile layers. That's what this pull request fixes.

tschaub (approved):
Not sure how involved it would be to set up, but this feels like an example of where we could use a rendering test.

ahocevar:
Thanks for the review, @tschaub. Rendering test added.

PR #15591, opened 2024-02-27
feat: add to-string operator

fredj:
Only on cpu.


tschaub:
Thanks, @fredj.  I think the `to-string` operator makes sense.  I’m wondering about the use of it for converting an array to a color.  In some cases, I think we accept an array for a color.  And I think we also have a `color` operator (maybe wrong, not looking at the code right now).  Did you find that an array didn’t work for a color?

tschaub (approved):
Thanks, @fredj

PR #15596, opened 2024-02-28
Fix #15595 - Stop caching TileRepresentation based on source key and use interimTile until tiles are loaded.

### LINKED ISSUE #15595: WebGLBaseTileLayerRenderer causes flickering when source or source key changes.
**Describe the bug**
When using `WebGLBaseTileLayerRenderer` there is flickering if trying to use source keys to update dimensions smoothly like in the wmts-dimensions example. This is caused by `tileRepresentationCache` caching based on source key firstly, and secondly because WebGLBaseTileLayerRenderer does not use `interimTiles` correctly.

This bug report i think reports the same as #15310, but since it was closed I want to reopen it since i have proposed a fix in !PULL_REQUEST.

**To Reproduce**
1. Checkout a5bf0757539c7ac97547c7f79d51b0b16176d9b1
2. `npm run serve-examples`
3. Go to [localhost/examples/webgl-dimensions](http://localhost:8080/webgl-dimensions.html) and change threshold
4. See the map flickering in between tile loads.

**Expected behavior**
Same behaviour as [localhost/examples/wmts-dimensions](http://localhost:8080/wmts-dimensions.html) and change threshold

sweepline:
Fixes #15595 by stopping caching TileRepresentation based on source key and use interimTile until tiles are loaded.

I am new to contributing to this project so im not sure if this is the best way to fix the underlying issue. I think the main aspects are correct. For example I could not find a way to keep using `isDrawableTile_`, but the current way its used in  `WebGLBaseTileLayerRenderer` is also causing issues.

### DIFF
--- a/examples/webgl-dimensions.html
+++ b/examples/webgl-dimensions.html
@@ -1,0 +1,15 @@
+---
+layout: example.html
+title: WebGL Tile Transitions
+shortdesc: Example of smooth tile transitions when changing the dimension of a WebGL layer.
+docs: >
+  Demonstrates smooth reloading of layers when changing a dimension continuously. The reason this example exists is to show this feature working with the WebGL based renderer as compared to the default canvas based renderer shown in <a href="/en/latest/examples/wmts-dimensions.html">the WMTS Tile Transitions example</a>. The demonstration layer is a global sea-level computation (flooding computation from <a href="https://scalgo.com/">SCALGO</a>, underlying data from <a href="https://cgiarcsi.community/data/srtm-90m-digital-elevation-database-v4-1">CGIAR-CSI SRTM</a>) where cells that are flooded if the sea-level rises to more than <em>x</em> m are colored blue. The user selects the sea-level dimension using a slider.
+tags: "webgl, parameter, transition, grid"
+---
+
+<div id="map" class="map"></div>
+<label>
+  Sea-level
+  <input id="slider" type="range" value="10" max="100" min="-1">
+</label>
+<span id="theinfo"></span>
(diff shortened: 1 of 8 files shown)


sweepline:
Is there anything else I need to do to get a rerun of the workflow / get assigned a reviewer. I could not read anything in [DEVELOPING.md](https://github.com/openlayers/openlayers/blob/main/DEVELOPING.md) or [CONTRIBUTING.md](https://github.com/openlayers/openlayers/blob/main/CONTRIBUTING.md) about the review process other than I need to pass the CI tests. The ones failing on my local machine also fails on master so I think I need to run them in CI.

ahocevar:
@sweepline Thanks for your contribution. You have correctly followed the process, and the reason why you haven't gotten a review yet is simply that none of the potential reviewers (@tschaub in this case) has found time to look at it yet.

I have triggered a CI run, and the result looks good.

tschaub:
Thanks for putting time into diagnosing this issue and proposing a fix, @sweepline.

I have been working on another change that will let us remove interim tiles altogether (see #15568).  The goal there is to still use previously loaded cached tiles, but to do so without the current interim tile chain handling.  That is a breaking change, and will require discussion with (and likely assistance from) others if we are to get it in.  If the new tile source changes don't make it in, we can revisit these changes, but I'd like to give that a chance first.  In either case, the example you add here will provide a useful case for the behavior we want to retain (and in both cases, we need to decide what to do with the `isDrawableTile_` function).

sweepline:
Thanks for looking at it @tschaub. At SCALGO we are already running a patched version of Openlayers with this PR in it as we are using a slider to change dimension of a DataTileSource with a custom WebGL renderer. 
The patch works well for our use-case but we will keep an eye on #15568.

Issue #15589, opened 2024-02-26
The new Google Maps source in v9 doesn't comply with Google's policies by default (missing logo and attribution)


fnicollet:
**Describe the bug**
When using the new Google Maps source by default, it looks something like this:
![image](https://github.com/openlayers/openlayers/assets/1240481/da56c596-1138-4173-bfae-a20438c4447e)

The Google API Tile page:
https://developers.google.com/maps/documentation/tile/overview
has a link to the policies:
https://developers.google.com/maps/documentation/tile/policies
which states:
++++++++++++
_When you use the Map Tiles API to display Google Maps, you must also display the Google logo on the map, and you must not alter the logo in any way. You should place the Google logo in the bottom left-hand corner of the map, and you should place the attribution information in the right-hand corner of the map_
++++++++++++
I believe the Google OpenLayers source should by default embed the Google logo in the bottom-left corner.

This is how Cesium handles it by default for instance:
https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/GoogleMaps.js#L36

If you condider that it's the developper's responsability to add the Google logo, it should at least be properly showcased in the example, to enforce best practices

Also, Google policies state:
++++++++++++
_Data returned from the Map Tiles API requires the display of attribution and copyright information from the appropriate metadata or viewport information requests. You should display this information, in full as provided in the appropriate location, usually the bottom right corner of the displayed set of tiles, or in the 3D renderer view. Note that the attribution strings are variable, depending on the map data requested by the renderer's viewport._

_If it's infeasible to display data attributions in full due to viewport size constraints, consider adding a hover-over or clickable UI element labeled "Data sources", which opens within the map window to provide attribution information. Always aim to maintain good cartographic practices._
++++++++++++
At the moment, here is how the Google source displays the attribution:

![image](https://github.com/openlayers/openlayers/assets/1240481/79399a42-168b-432a-881d-cc719023fdfb)

I don't believe this agrees with Google policies by default. From what it says on Google's documentation, attributions depend on the current viewport (position and scale), So I believe the Google OpenLayers source should automatically request the current viewport attribution (with a bit of throttle/debounce to avoid having too many requests). 
I don't think the developper should do it in his own code (and I guess you would need the session token generated by the OL source for it). It is can't be done automatically, the official example should show how to properly request the attributions and display them,

To request the viewport attribution, there is a Google WebService described here:
[...]

PR #15598, opened 2024-02-28
Retrieve Google attributions automatically via their API and provide an example showing how to add the Google logo

fnicollet:
Fixes issue #15589 
To avoid adding to OL's bundle size, I decided not to include the Google logo in the source code. It also allows developpers to chose the logo they prefer/need (color/white and high DPI or not).
Instead I added an example of how to add a custom OL Control in the bottom-left corner, with a simple logo added as base64.

For the attributions, they are now retrieved via their API. Using ViewHint.ANIMATING / ViewHint.INTERACTING was still resulting in many identical API calls, so I added a little memory of the last API parameters requested, to avoid requesting the same viewport informations many time.

The new example can be tested easily by entering your own Google API Key. Then you need to zoom in somewhere to see the attributions change:
image

On Google's documentation, section "Displaying third-party data attributions":

https://developers.google.com/maps/documentation/tile/policies#displaying-third-party-data-attributions

It says:
+++++
_When you use the Map Tiles API to display a Google Maps data as a basemap and overlay third-party (non-Google) geospatial data, you may be required to display attribution from the third-party data provider. You must not overlap or obscure the Google data attribution in any way with the data attribution of the third-party data. The attribution of third-party data must clearly be disassociated from Google's data attributions. In addition, it must be clear that Google's logo and Google's data attribution are associated with the basemap and with each other._
+++++

So I added a "Google : " prefix, to avoid conflict with other source attributions.

To be able to retrieve attributions in an async way, I changed the way the setAttributions function works, adding the patch from @ahocevar in the original issue.

I also added more information in the text of the example, to explain how you can comply with Google's policies and linked back to documentation.

I tried running the linter but on Windows, it gave me thousands of errors:
image

So hopefully the formatting is OK


ahocevar:
@fnicollet Thanks for your effort on this! Can you please run `npm run lint -- --fix` to auto-fix the linter errors?

mike-000:
> So I added a "Google : " prefix, to avoid conflict with other source attributions.

By default attributions for all sources appear as a single line so adding a prefix might not help. Multiple attributions can be shown on separate lines by adding
```
.map .ol-attribution li {
  display: block;
}
```
to the application css.

mike-000:
To avoid the large base64 string in the example code you could link the logo https://developers.google.com/static/maps/documentation/images/google_on_white.png directly (in the same way as the IGN logo in https://openlayers.org/en/latest/examples/wmts-ign.html).

fnicollet:
Following @mike-000 comments, I removed the prefix (devs can add the CSS to display lines as block in its on code, if mixing with other sources) and linked to the image URL instead.

@ahocevar ran the linter, which fixed a couple of issues (didn't the the thousands of errors when using the --fix option)

mike-000:
> didn't the the thousands of errors when using the --fix option

It looks like your editor is changing line endings from LF to CRLF.

src/ol/source/Google.js
```diff
@@ -238,10 +252,48 @@ class Google extends TileImage {
...
+      frameState.viewHints[ViewHint.ANIMATING] ||
+      frameState.viewHints[ViewHint.INTERACTING]
+    ) {
+      return this.previousViewportAttribution_;
+    }
+    const transform = getTransformFromProjections(
+      this.getProjection(),
+      getProjection('EPSG:4326'),
+    );
+    const epsg4326Extent = applyTransform(frameState.extent, transform);
```
ahocevar, line 271: Using
```js
import { transform as transformExtent } from `ol/extent.js`;
```
you can simplify this to
```suggestion
    const epsg4326Extent = transformExtent(frameState.extent, this.getProjection(),
      getProjection('EPSG:4326'));
```

src/ol/source/Google.js
```diff
@@ -238,10 +252,48 @@ class Google extends TileImage {
...
     this.setState('ready');
   }
 
+  async fetchAttributions(frameState) {
+    if (
+      frameState.viewHints[ViewHint.ANIMATING] ||
+      frameState.viewHints[ViewHint.INTERACTING]
+    ) {
+      return this.previousViewportAttribution_;
+    }
```
ahocevar, line 266: I think if you also keep checking for `frameState.animating` here, you won't get duplicate calls for the same extent. 

If that does not work, you could bail out here already when the extent has not changed:
```suggestion
    if (
      frameState.viewHints[ViewHint.ANIMATING] ||
      frameState.viewHints[ViewHint.INTERACTING] ||
      equalsExtent(frameState.extent, this.previousViewportExtent_)
    ) {
      return this.previousViewportAttribution_;
    }
    this.previousViewportExtent_ = frameState.extent.slice(0);
```

src/ol/source/Google.js
```diff
@@ -238,10 +252,48 @@ class Google extends TileImage {
...
+    const north = epsg4326Extent[3];
+    const south = epsg4326Extent[1];
+    const east = epsg4326Extent[2];
+    const west = epsg4326Extent[0];
+    const viewportExtent = `zoom=${zoom}&north=${north}&south=${south}&east=${east}&west=${west}`;
+    // check if the extent or zoom has actually changed to avoid unnecessary requests
+    if (this.previousViewportExtent_ == viewportExtent) {
+      return this.previousViewportAttribution_;
+    }
+    this.previousViewportExtent_ = viewportExtent;
```
ahocevar, line 286: Not needed if the above works.

ahocevar (changes requested):
This is coming together nicely! I added a few more suggestions.

src/ol/source/Google.js
mike-000: The Bing issues in #14526 also apply here but are easier to fix

To handle wrapping the longitudes should be normalized https://developers.google.com/maps/documentation/tile/2d-tiles-overview#viewport_info_request (or set to-180 +180 if more than one world width is visible).

To handle reprojection the extent should be transformed to EPSG:4326 from the view projection instead of the source projection, in which case the resolution used to calculated the zoom would need to be obtained using `ol/reproj#calculateSourceExtentResolution`.

ahocevar: @mike-000 I think reprojection is beyond the scope of this pull request, and I doubt it is allowed to reproject a Google map image.

src/ol/source/Google.js
fnicollet: Tried it, it removed most of the duplicates (like I don't get duplicates when panning the map), but it gets me one duplicate when zooming (I guess there is an event when zoom starts and when it ends), so I still need a check to abort the unnecessary request. I added the `frameState.animate` check for early termination but kept the this.previousViewportExtent check that was there

src/ol/source/Google.js
ahocevar: Good point about the -180...+180 range. The easiest way to achieve this is to use `toLonLat()` instead of `transformExtent`:
```js
const [west, south] = toLonLat(getBottomLeft(frameState.extent), frameState.viewState.projection);
const [east, north] = toLonLat(getTopRight(frameState.extent), frameState.viewState.projection);
```

fnicollet: @ahocevar your suggestion don't seem to work. There is no transform function in ol/extent.js, only applyTransform (i copied the code from Bing Maps source)

fnicollet: Where is the toLonLat function from ? I can't find it in the source code

fnicollet: transformExtent is in proj, nevermind

ahocevar: toLonLat as well.

fnicollet: ok, got a bit confused as "transform as transformExtent" doesn't work when using proj, because the first argument is a coordinates, so it was projecting only the first 2 coordinates of the extent. Ended up using toLatLon to get the two corners like you suggested, should be good now

examples/google.js
```diff
@@ -16,8 +17,24 @@ function showMap(key) {
...
+        'https://developers.google.com/static/maps/documentation/images/google_on_white.png';
+      super({
+        element: element,
+      });
+    }
+  }
+
   const map = new Map({
     layers: [new Layer({source})],
+    controls: defaultControls().extend([new GoogleLogoControl()]),
```
ahocevar, line 37: I'm not sure, but the terms might require the attribution to be always visible. If that's the case, the controls should be configured appropriately:
```suggestion
    controls: defaultControls({
      attributionOptions: {
        collapsible: false
      }
    }).extend([new GoogleLogoControl()]),
```
What's your interpretation of the terms here?

fnicollet: Their doc states
++++
_If it's infeasible to display data attributions in full due to viewport size constraints, consider adding a hover-over or clickable UI element labeled "Data sources", which opens within the map window to provide attribution information._
++++
So you are right i think. When the map is not displayed on a small screen, there is room to display the attibutions, and it should be so. Best would be a "responsive" change of this collapsible property, but it might be a bit far fetched :) i'll commit your changes

mike-000: Would it be better to default `attributionsCollapsible` in the source super similar to OSM?

fnicollet: Sure, i'll add it the same way

examples/google.js
```diff
@@ -16,8 +17,28 @@ function showMap(key) {
...
+    }
+  }
+
   const map = new Map({
     layers: [new Layer({source})],
+    controls: defaultControls({
+      attributionOptions: {
+        collapsible: false,
+      },
+    }).extend([new GoogleLogoControl()]),
```
ahocevar, line 41:
```suggestion
    controls: defaultControls().extend([new GoogleLogoControl()]),
```

ahocevar (approved):
Great work! Thanks, @fnicollet

fnicollet:
Great, thanks both of you for your help on this!

tschaub:
Thank you @fnicollet for taking time to read the terms of service, open an issue, and contribute this enhancement.  And thanks @mike-000 and @ahocevar for the time spent reviewing.

PR #15600, opened 2024-02-29
Do not use arrays for non-consecutive zIndex keys

### LINKED ISSUE #15599: Default Styles for Modifying Feature Broke in Upgrade from 8.2 to 9.0
I noticed when I upgraded from 8.2 to 9.0 that my editing styles broke for my LineString. It took me a couple of hours, but in the documentation here: https://openlayers.org/en/latest/apidoc/module-ol_style_Style-Style.html  I noticed that it has the default style as:
` styles['Circle'] = styles['Polygon'].concat(
   styles['LineString']
 );`

when I believe it should be:
` styles['Circle'] = styles['Point'].concat(
            styles['LineString']
        );`

The latter bit of code, fixed the issue.

ahocevar:
Fixes #15599

### DIFF
--- a/src/ol/render/canvas/ExecutorGroup.js
+++ b/src/ol/render/canvas/ExecutorGroup.js
@@ -118,9 +118,9 @@ class ExecutorGroup {
     this.renderedContext_ = null;
 
     /**
-     * @type {Array<Array<import("./ZIndexContext.js").default>>}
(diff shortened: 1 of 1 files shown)


marcjansen (approved):
Looks good to me, please merge

PR #15602, opened 2024-03-01
Updates to the Google attribution

tschaub:
This makes a minor adjustment to the attribution handling added in #15598, exposing the `attributionsCollapsible` option on the source.  This makes it easier for developers to decide whether the attribution control is collapsed or not by default.  I think having the attribution be collapsed still complies with [the policies](https://developers.google.com/maps/documentation/tile/policies#displaying-googles-data-attributions):

> Data returned from the Map Tiles API requires the display of attribution and copyright information from the appropriate metadata or viewport information requests. You should display this information, in full as provided in the appropriate location, usually the bottom right corner of the displayed set of tiles, or in the 3D renderer view. Note that the attribution strings are variable, depending on the map data requested by the renderer's viewport.

> If it's infeasible to display data attributions in full due to viewport size constraints, consider adding a hover-over or clickable UI element labeled "Data sources", which opens within the map window to provide attribution information. Always aim to maintain good cartographic practices.

I think this allows for the attributions to be revealed based on user interaction

### DIFF
--- a/src/ol/source/Google.js
+++ b/src/ol/source/Google.js
@@ -25,6 +25,7 @@ const maxZoom = 22;
  * @property {Array<string>} [layerTypes] The layer types added to the map (e.g. `'layerRoadmap'`, `'layerStreetview'`, or `'layerTraffic'`).
  * @property {boolean} [overlay=false] Display only the `layerTypes` and not the underlying `mapType` (only works if `layerTypes` is provided).
  * @property {Array<Object>} [styles] [Custom styles](https://developers.google.com/maps/documentation/tile/style-reference) applied to the map.
+ * @property {boolean} [attributionsCollapsible=true] Allow the attributions to be collapsed.
  * @property {boolean} [interpolate=true] Use interpolated values when resampling.  By default,
  * linear interpolation is used when resampling.  Set to false to use the nearest neighbor instead.
  * @property {number} [cacheSize] Initial tile cache size. Will auto-grow to hold at least the number of tiles in the viewport.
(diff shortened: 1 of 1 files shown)


mike-000:
`transformExtent` does not work with wrapX

> east and west must be in the range (-180, 180). To express bounds crossing the antimeridian, west can be positive (for example, 170) and east can be negative (for example, -170).

mike-000:
As with OSM the `attributionsCollapsible: false` setting can be overridden using

`controls: defaults({attributionOptions:{collapsible: true}}),`

tschaub:
> As with OSM the `attributionsCollapsible: false` setting can be overridden

My opinion is that that `attributionsCollapsible: false` is a bad default and that forcing everyone to discover `controls: defaults({attributionOptions:{collapsible: true}})` is a bad user experience.

tschaub:
> `transformExtent` does not work with wrapX

Good point.  I removed the change to use `transformExtent`.  The north and south may not be correct for some transforms with the current logic - but perhaps not a problem in practice.

mike-000:
Having `false` as the default would be expected if the source terms state that should be normal practice, but I agree having an option to override that in the source would be more user friendly (for both Google and OSM).

ahocevar (approved):

mike-000:
> The north and south may not be correct for some transforms with the current logic

It will be correct except in some non-parallel reprojections (e.g. polar).  But to handle reprojection we cannot simply use the view resolution to obtain the source tile grid zoom level, we would first need to obtain the reprojection tile grid zoom level and from that calculate the zoom of the original tiles.

PR #15603, opened 2024-03-01
Listen for `change:size` event in Attributions example

mike-000:
The window `resize` event does not fire in Chrome when maximising/restoring the window, but the map `change:size` event does.


marcjansen (approved):
Thanks, please merge

ahocevar (approved):

PR #15604, opened 2024-03-02
Make UTF wrapX and empty string attribution handling consistent with TileJSON

### LINKED ISSUE #14542: UTFGrid wrapX anomoly

mike-000:
Fixes #14542

Separated from #14543 as that is now outdated and reprojection is a separate issue which now also affects Google sources.

The UTF data in the example now wraps together with the corresponding raster data and no attribution control is displayed for an empty string.

### DIFF
--- a/src/ol/source/UTFGrid.js
+++ b/src/ol/source/UTFGrid.js
@@ -263,6 +263,7 @@ export class CustomTile extends Tile {
  * If not provided, `url` must be configured.
  * @property {string} [url] TileJSON endpoint that provides the configuration for this source.
  * Request will be made through JSONP. If not provided, `tileJSON` must be configured.
+ * @property {boolean} [wrapX=true] Whether to wrap the world horizontally.
  * @property {number|import("../array.js").NearestDirectionFunction} [zDirection=0]
  * Choose whether to use tiles with a higher or lower zoom level when between integer
  * zoom levels. See {@link module:ol/tilegrid/TileGrid~TileGrid#getZForResolution}.
@@ -281,6 +282,7 @@ class UTFGrid extends TileSource {
     super({
       projection: getProjection('EPSG:3857'),
       state: 'loading',
+      wrapX: options.wrapX !== undefined ? options.wrapX : true,
       zDirection: options.zDirection,
     });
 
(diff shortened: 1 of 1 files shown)


src/ol/source/UTFGrid.js
```diff
@@ -263,6 +263,7 @@ export class CustomTile extends Tile {
  * If not provided, `url` must be configured.
  * @property {string} [url] TileJSON endpoint that provides the configuration for this source.
  * Request will be made through JSONP. If not provided, `tileJSON` must be configured.
+ * @property {boolean} [wrapX=true] Whether to wrap the world horizontally.
```
mprins, line 266: IMO horizontally is an ambiguous term when a projected plane is in play.

```suggestion
 * @property {boolean} [wrapX=true] Whether to wrap the map along the X-axis (equator) of the projection.
```

ahocevar: @mprins The equator is definitely wrong in this context.

mprins: indeed!

mike-000: I think that is a separate issue which affects all tile sources (except for DataTile and GeoTIFF where the option is used inconsistently).  Maybe the description used for vector tiles is better and should be used everywhere?

ahocevar (approved):
Thanks, @mike-000

PR #15620, opened 2024-03-09
Store previous transform in FrameState

### LINKED ISSUE #15593: Flickering map when using webview on a NOTHING phone
Hello,

When bumping OL version to V9, I found that, only on the Nothing phone, the map was flickering randomly when panning or zooming.
Most of the time, the map was perfect, but sometimes, it went completely black. This occurred mostly around zoom levels 3,4,5.

This occurred on 2 Nothing phones with different Android versions (V12 & V13) and any webview version.

This did not occurred on any other Android, Ios or Web platforms.

I was able to pinpoint this regression to commit #15344 .

When investigating, I found that the test `has.js#WORKER_OFFSCREEN_CANVAS ` is usually false on all my test systems, and so the `transform.js#toString()` was always using the DOM.

After this commit, this test is not performed anymore, and the new transform calculus makes this specific phone flickering for an unknown reason.

As a quick fix, I re-added the test, and so used the previous DOM transform. This solved the problem.

It looks like there are daemons lurking around there...

Michel

ahocevar:
This pull request stores the previous layer's transform on the `FrameState`, instead of reading the CSS matrix transform from the canvas. This allows some code simplifications, and instead of comparing CSS matrix transform strings, we now compare transform arrays with a tolerance, using the new `equals()` function of the `ol/transform` module.

Fixes #15593

### DIFF
--- a/src/ol/transform.js
+++ b/src/ol/transform.js
@@ -46,6 +46,26 @@ export function reset(transform) {
 }
 
 /**
+ * @param {!Transform} transform1 Transform 1.
+ * @param {!Transform} transform2 Transform 2.
+ * @return {boolean} Transform 1 and 2 are pixel equal.
+ */
+export function equals(transform1, transform2) {
+  if (transform1 === transform2) {
+    return true;
+  }
+  const tolerance = 1e-6;
+  return (
+    Math.abs(transform1[0] - transform2[0]) < tolerance &&
+    Math.abs(transform1[1] - transform2[1]) < tolerance &&
+    Math.abs(transform1[2] - transform2[2]) < tolerance &&
+    Math.abs(transform1[3] - transform2[3]) < tolerance &&
+    Math.abs(transform1[4] - transform2[4]) < tolerance &&
+    Math.abs(transform1[5] - transform2[5]) < tolerance
+  );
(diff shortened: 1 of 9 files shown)


src/ol/transform.js
```diff
@@ -45,6 +45,26 @@ export function reset(transform) {
...
+ * @param {!Transform} transform1 Transform 1.
+ * @param {!Transform} transform2 Transform 2.
+ * @return {boolean} Transform 1 and 2 are pixel equal.
+ */
+export function equals(transform1, transform2) {
+  if (transform1 === transform2) {
+    return true;
+  }
+  const tolerance = 1e-6;
+  return (
```
mike-000, line 58: This is efficient but ideally it should compare the differences relative to the values instead of using a fixed tolerance for each element.  Since browsers will round when applying transforms (e.g. Firefox uses Float32) should we not attempt something similar, e.g. `equalsArray(transform1.map(Math.fround), transform2.map(Math.fround))`?

ahocevar: Tried with `Math.fround`, but that breaks existing tests.

tschaub (approved):

Issue #15636, opened 2024-03-13
MultiPolygonGeometry with empty coordinates throws an undefined error


yphanikiran:
**Describe the bug**
When an MultiPolygonGeometry with empty coordinates is passed to readFeature() method, throws undefined error as shown in the below screenshot

![image](https://github.com/openlayers/openlayers/assets/145588190/2d04dae3-583e-4027-85e0-1a71f47134ac)

**To Reproduce**
1. Create an object with empty coordinates array 
```
const featureObject = {
    "type": "Feature",
    "id": "test",
    "properties": {},
    "geometry": {
        "type": "MultiPolygon",
        "coordinates": [
            [
                [],
                []
            ]
        ]
    }
}
```
2. pass it readFeature method
```
const geojsonFormat = new GeoJSON({
    dataProjection: 'EPSG:4326',
    featureProjection: 'EPSG:3857'
});
const feature = this.geojsonFormat.readFeature(featureObject) as Feature;
```
3. we will get an error in the console as below
![image](https://github.com/openlayers/openlayers/assets/145588190/0ad89f47-1060-48f9-9ce6-cca3d28ecf72)

**Expected behavior**
when a MultiPolygonGeometry with empty coordinates is passed to readFeature() method, it should handle it without throwing an error

ahocevar:
The GeoJSON you're throwing at the `GeoJSON` format here is invalid. You can verify that at https://geojsonlint.com.

At the very least, to get a valid GeoJSON, you'd have to remove the empty coordinate pairs:
```json
{
    "type": "Feature",
    "id": "test",
    "properties": {},
    "geometry": {
        "type": "MultiPolygon",
        "coordinates": [
            []
        ]
    }
}
```

PR #15637, opened 2024-03-13
allow empty GeoJSON coordinate for MultiPolygon

yphanikiran:
When an MultiPolygonGeometry with empty coordinates is passed to readFeature() method, throws undefined error, to handle this, conditional operator(?) is added while reading the coordinates to assign the value to **stride** in **readMultiPolygonGeometry** method

Fixes: [15636](https://github.com/openlayers/openlayers/issues/15636)


ahocevar:
Thanks for your effort, @yphanikiran. However, please see my comment on #15636. This change would allow parsing invalid GeoJSON, which we do not want.

PR #15638, opened 2024-03-13
Fix WFS property format writer for multiline strings

### LINKED ISSUE #10127: WFS property format writer handling of string values with new lines
Hello,
When updating a feature through WFS, if there are any newline characters in string properties, these are lost (for example, geoserver handles these newlines as whitespace)

Instead I think the WFS format writer should protect string properties in CDATA block when there are any newline characters in string values (or maybe for any string value?)
The attached patch solves the issue for me:

[openlayers-fix-WFS-write-property.patch.txt](https://github.com/openlayers/openlayers/files/3726595/openlayers-fix-WFS-write-property.patch.txt)

aboulan:
When updating a feature through WFS, if there are any newline
  characters in a string properties, these are lost (for example,
  geoserver handles these newlines as whitespace).

  To fix this, multiline string properties should be encoded as
  CDATA block.

Fix #10127

### DIFF
--- a/src/ol/format/WFS.js
+++ b/src/ol/format/WFS.js
@@ -25,6 +25,7 @@ import {get as getProjection} from '../proj.js';
 import {
   readNonNegativeIntegerString,
   readPositiveInteger,
+  writeCDATASection,
   writeStringTextNode,
 } from './xsd.js';
 
@@ -938,7 +939,20 @@ function writeProperty(node, pair, objectStack) {
         GML32.prototype.writeGeometryElement(value, pair.value, objectStack);
       }
     } else {
-      writeStringTextNode(value, pair.value);
+      if (typeof pair.value === 'string') {
+        // Generate CDATA section to preserve whitespaces.
+        pair.value.split(']]>').forEach((part, i, a) => {
+          if (i < a.length - 1) {
+            part += ']]';
+          }
+          if (i > 0) {
+            part = '>' + part;
+          }
+          writeCDATASection(value, part);
+        });
+      } else {
+        writeStringTextNode(value, pair.value);
+      }
     }
   }
 }


src/ol/format/WFS.js
```diff
@@ -938,7 +939,11 @@ function writeProperty(node, pair, objectStack) {
         GML32.prototype.writeGeometryElement(value, pair.value, objectStack);
       }
     } else {
-      writeStringTextNode(value, pair.value);
+      if (typeof pair.value === 'string' && pair.value.search('\n') >= 0) {
```
ahocevar, line 942:
```suggestion
      if (typeof pair.value === 'string' && pair.value.includes('\n') >= 0) {
```
The above should work as well, and yield better performance because no regular expression is created.

ahocevar (changes requested):

src/ol/format/WFS.js
rdewit: Did you mean to remove the ` >= 0` part as well, @ahocevar, like below?

```suggestion
      if (typeof pair.value === 'string' && pair.value.includes('\n')) {
```

ahocevar: Yes, thanks @rdewit

aboulan: When looking at the issue more closely, I wonder if it wouldn't be better to completely skip the test and always output a CDATA block.
Indeed, it is not only end of lines that are not correctly interpreted in a text node: any white space combination (space, tab, end of line) is interpreted as a single space (as per the html/xml rules), so that for example a string with multiple consecutive spaces is also not correctly handled.
But using CDATA blocks are not the perfect solution either: they cannot contain the end of block sequence ("]]>").
They also cannot contain arbitrary binary data (it must be a valid utf8 string) but this is probably not a problem since WFS property values are supposed to be valid xml text (but I don't know if the WF specs explicitly says something about this)
For simple use cases (multi-line string, which I needed initially), CDATA blocks are working great out of the box with geoserver (and any xml parser), but maybe the correct solution for the general case would be to have an API to let the user handle the serialization himself (base64 ?)
What do you think?

aboulan: The following may be the best compromise to catch any white space combination without using CDATA blocks for single words:
[ ' ', '\n', '\t' ].some(c => pair.value.includes(c)

ahocevar: Fine with me to always create a CDATA block.

M393:
If you have text with `]]>` end marker included you can write multiple CDATA sections by splitting the end marker.

```js
pair.value.split(']]>').forEach((part, i, a) => {
  if (i < a.length - 1) {
    part += ']]';
  }
  if (i > 0) {
     part = '>' + part;
  }
  writeCDATASection(value, part);
});
```

src/ol/format/WFS.js
```diff
@@ -938,7 +939,17 @@ function writeProperty(node, pair, objectStack) {
...
     } else {
-      writeStringTextNode(value, pair.value);
+      if (typeof pair.value === 'string') {
+        try {
+          // Generate a CDATA section to preserve whitespaces.
+          writeCDATASection(value, pair.value);
+        } catch {
+          // If value contains the CDATA closing marker ']]>', fallback to a text node.
+          writeStringTextNode(value, pair.value);
+        }
```
ahocevar, line 949: Not sure if it would work, but how about
```suggestion
      pair.value.split(']]>').forEach((text) => writeCDataSection(value, text));
```

src/ol/format/WFS.js
```diff
@@ -938,7 +939,20 @@ function writeProperty(node, pair, objectStack) {
...
+      if (typeof pair.value === 'string') {
+        // Generate CDATA section to preserve whitespaces.
+        pair.value.split(']]>').forEach((part, i, a) => {
+          if (i < a.length - 1) {
+            part += ']]';
+          }
+          if (i > 0) {
+            part = '>' + part;
+          }
+          writeCDATASection(value, part);
```
ahocevar, line 951: This won't work as expected. It will convert `"foo>]]bar"` to `<![CDATA[foo]]]]><![CDATA[>bar]]>` instead of `<![CDATA[foo>]]><![CDATA[]]bar]]>`.

I think this addition could benefit from a test case.

ahocevar (changes requested):

MelkorCC:
What is the status of this PR? Is there a way for me to support it?

ahocevar:
@MelkorCC If you'd be able to create a test according to my comment above, it would be great.

MelkorCC:
@ahocevar while working on test coverage for this i noticed a problem, the change suggested by this PR is actually not enough. It only covers cases in which `writeProperty` is actually used. This is for example NOT the case for Insert operations. So to support this i modified GML 2/3 `writeFeature`. But then i wondered about filters, while i personally only need CDATA support for properties in Update and Insert operations, other users could in theory also need CDATA support for literals used in filters. So i modified `writeLiteral` and `writeExpression`. But i figured that i should maybe ask you about your opinion about all of this before putting even more time into fixing testcases and creating new ones. Do you feel like this is the way to go? Or do you have a better idea? Or am i maybe missing something here and making stuff more complicated than it needs to be?

ahocevar:
Thanks for your effort on this, @MelkorCC. Thinking about what you reported, I'd say it would be best to change the `writeStringTextNode()` function in `ol/format/xsd.js` to check the string for characters not allowed in XML text nodes (should work with a single regular expression or a set of regular expressions), and automatically use a CDATA section when the string contains invalid characters. Then existing `writeCDATASection()` call sites could also be replaced with `writeStringTextNode()`. There are a couple in the KML format, and maybe also in other places.

MelkorCC:
@ahocevar sry for the delay, but i started working on this again with the approach you suggested.
However i hit a roadblock, it seems like Chromium's DOMParser handles CDATA sections in an incorrect way, the parsed XMLDocument contains a merged CDATA section instead of two CDATA sections:
image
In Firefox this is handled correctly:
image

Testcode if somebody wants to verify it:
```JS
const xml = '<test><![CDATA[test]]]]><![CDATA[>test]]></test>';
const node = new DOMParser().parseFromString(xml, 'application/xml');
node.childNodes[0].childNodes
```

My guess is that this is related to [this bug in chromium](https://issues.chromium.org/issues/327273585).

Since OpenLayers uses karma with headless chrome for testing my testcases fail because they currently utilize the DOMParser to parse the xml for the comparison even tho the serialization works as expected.
Do you suggest we delay this until the fix landed in Chromium? Alternatively i could also construct the XMLDocument programmatically instead of parsing it. I went with the parsing for now since this is what other tests (wfs, kml etc) use and since it makes the test much cleaner (everybody understands with one look how the xml should look like).

MelkorCC:
I thought about this and i think for the given testcases it is fine to construct the XMLDocument programmatically.
In all of my new testcases that are affected by this XMLDocuments are very small and creating them should only require 3 lines of code. No existing testcases are affected by this.
While i still believe it would be nicer to have the XML as a string like in other testcases, i think it makes not that much sense to wait until Chromium eventually managed to fix their bugs with regards to CDATA sections. As i already mentioned the Chromium bug is not a problem with regards to the functionality of this change, it only affects the parsing of the XML data for comparison in the testcases.

ahocevar:
Fixed with #16596

PR #15649, opened 2024-03-18
Avoid flickering by using transform matrix as returned from style

### LINKED ISSUE #15644: When adding draw interaction, it flickers while zooming in and out (Version 9.0.0)
This happens only with the newest version 9.0.0. Adding a draw interaction, like drawing polygon, the map flickers while zooming.
The polygon is still visible, but the map under it disappears and flickers. Sometimes it even is blank when I stop zooming.

Style:

```ts
  styleRedOutline: Style = new Style({
    fill: new Fill({
      color: [0, 0, 0, 0.1]
    }),
    stroke: new Stroke({
      color: [255, 0, 0, 0.5],
      width: 5
    })
  }
```

Initialization:

```ts
   const osmLayer = new TileLayer({
      source: new OSM()
    });

    let that = this;

    this.vectorLayer = new VectorLayer({
      source: this.source,
      style: function (feature) {
        return that.styleRedOutline;
      }
    });

    this.view = new View({
      center: [-472202, 7530279],
      zoom: 12
    });
    this.map = new olMap({
      layers: [
        osmLayer,
        this.vectorLayer
      ],
      view: this.view,
      controls: []
    });
```

Adding drawing interaction:

```ts
this.interaction = new Draw({
      type: 'Polygon',
      source: this.source
    });

    let draw: Draw = this.interaction;

    draw.on('drawend', evt => {
      console.log("drawEnd")
      this.lastSelectedFeature = evt.feature;
    });

    this.map?.addInteraction(this.interaction);
```

I tried to trigger layer change and other stuff, but it seems to be a problem with the version 9.0.0. With 8.2.0 it does not happen.

ahocevar:
Hopefully fixes #15644.

Instead of reverting to what we had before #15344, I decided to keep going with storing the current transform on the frame state, but use the css transform matrix as returned from the current style. This should avoid the extra canvases that were fixed with #15344, and avoid the flickering introduced therewith.

### DIFF
--- a/src/ol/transform.js
+++ b/src/ol/transform.js
@@ -1,6 +1,7 @@
 /**
  * @module ol/transform
  */
+import {WORKER_OFFSCREEN_CANVAS} from './has.js';
 import {assert} from './asserts.js';
 
 /**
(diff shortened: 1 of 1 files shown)


tschaub:
@ahocevar was your latest change pushed after https://github.com/openlayers/openlayers/issues/15644#issuecomment-2005137163 ?  Unsure if you saw that message or not.

ahocevar:
@tschaub I have not seen that message yet. A strange issue really, and probably worth to create a Chromium issue instead of trying to work around it here. But I’m unable to reproduce, so I’m unable to create a reproduction case without OpenLayers involved.

ahocevar:
Closing this in favor of #15652.

PR #15669, opened 2024-03-25
Update satellite tiles url

### LINKED ISSUE #15668: MapTiler satellite tiles in examples
**Describe the bug**
`https://api.maptiler.com/tiles/satellite/{z}/{x}/{y}.jpg` used in multiple examples is no longer listed in MapTiler documentation and returns status 204 beyond zoom level 6.

**Expected behavior**
The following are listed, any of which should work to zoom 22 and could be used instead
`https://api.maptiler.com/tiles/satellite-v2/{z}/{x}/{y}.jpg`
`https://api.maptiler.com/maps/satellite/{z}/{x}/{y}.jpg` (tileSize 512)
`https://api.maptiler.com/maps/satellite/256/{z}/{x}/{y}.jpg`

mike-000:
Fixes #15668

There seems little to be gained by increasing maxZoom to 22 as it does not improve image quality

### DIFF
--- a/examples/layer-spy.js
+++ b/examples/layer-spy.js
@@ -22,7 +22,8 @@ const roads = new TileLayer({
 const imagery = new TileLayer({
   source: new XYZ({
     attributions: attributions,
-    url: 'https://api.maptiler.com/tiles/satellite/{z}/{x}/{y}.jpg?key=' + key,
+    url: 'https://api.maptiler.com/maps/satellite/{z}/{x}/{y}.jpg?key=' + key,
+    tileSize: 512,
     maxZoom: 20,
   }),
 });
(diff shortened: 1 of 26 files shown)


examples/layer-spy.js
bolollo, line 1: Can you also change the map of the roads layer from streets to streets-v2?

examples/side-by-side.js
bolollo, line 1: Can you also change the map of the `roadLayer` layer from `streets` to **`streets-v2`**?

`streets` maps it's **deprecated**

Thanks

mike-000: There are some other examples which use streets.  Unlike the old satellite url the old streets url is still working, so I think could be handled as a separate issue.  

Also the hybrid source used in https://openlayers.org/en/latest/examples/feature-move-animation.html is no longer documented but again it still works.  There is no alternative documented for that.

bolollo (approved):

tschaub (approved):

PR #15671, opened 2024-03-26
Clear RenderBox on DragBox Deactivation

### LINKED ISSUE #15571: Clear RenderBox on DragBox Deactivation

Users usually interact map with `Dragbox` and that interaction could be canceled as it fires `boxcancel` event according to [document](https://openlayers.org/en/v7.5.2/apidoc/module-ol_interaction_DragBox.DragBoxEvent.html#event:boxcancel).

Developers might need to implement "canceling drag" case, such like "when user press Escape, stop to drag". But there's no functions to handle this except `setActive(false)`.

But when developer stops DragBox interaction with `setActive(false)`, the `RenderBox`, which indicates where user drags, happens to remain. The `RenderBox` should be cleared by calling its own [`disposeInternal`](https://github.com/openlayers/openlayers/blob/f2c05afbd128428035f51945bbc74dc00aeaed7b/src/ol/render/Box.js#L52) function.

I simply reproduced with [stackblitz](https://stackblitz.com/edit/vitejs-vite-zqw5mk?embed=1&file=main.js)

FYI) It's hard to access `_box` of `DragBox` for Developers because it's private field.

clear existing `RenderBox` when deactivate(`setActive(false)`) `Dragbox`.

I hope to work on this feature request if it's accepted to improve developer experience.

behoney:
Currently, when the DragBox interaction is deactivated by calling `setActive(false)`, the RenderBox (which visually represents the dragged area) remains on the map. This Pull Request should resolve #15571 by clearing the RenderBox when the DragBox is deactivated.

### Changes

In the `setActive` method of the `DragBox` class, a new line has been added to call the `disposeInternal` method of the `box_` (RenderBox) instance when active is set to `false`. This will ensure that the RenderBox is properly disposed and removed from the map.

### DIFF
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -269,6 +269,23 @@ class DragBox extends PointerInteraction {
    * @param {import("../MapBrowserEvent.js").default} event Event.
    */
   onBoxEnd(event) {}
+
+  /**
+   * Activate or deactivate the interaction.
+   * @param {boolean} active Active.
+   * @observable
+   * @api
+   */
+  setActive(active) {
+    if (!active) {
+      this.box_.setMap(null);
+      this.dispatchEvent(
+        new DragBoxEvent(DragBoxEventType.BOXCANCEL, this.startPixel_, null),
+      );
+    }
+
+    super.setActive(active);
+  }
 }
 
 export default DragBox;
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -205,6 +205,10 @@ class DragBox extends PointerInteraction {
    * @param {import("../MapBrowserEvent.js").default} mapBrowserEvent Event.
    */
   handleDragEvent(mapBrowserEvent) {
+    if (!this.startPixel_) {
+      return;
+    }
+
     this.box_.setPixels(this.startPixel_, mapBrowserEvent.pixel);
 
     this.dispatchEvent(
@@ -222,6 +226,10 @@ class DragBox extends PointerInteraction {
    * @return {boolean} If the event was consumed.
    */
   handleUpEvent(mapBrowserEvent) {
+    if (!this.startPixel_) {
+      return false;
+    }
+
     this.box_.setMap(null);
 
     const completeBox = this.boxEndCondition_(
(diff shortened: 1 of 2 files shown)


src/ol/interaction/DragBox.js
```diff
@@ -269,6 +277,24 @@ class DragBox extends PointerInteraction {
...
+   * @observable
+   * @api
+   */
+  setActive(active) {
+    if (!active) {
+      this.box_.disposeInternal();
+      this.startPixel_ = null;
+      this.dispatchEvent(
+        new DragBoxEvent(DragBoxEventType.BOXCANCEL, this.startPixel_, null),
+      );
```
ahocevar, line 293:
```suggestion
      if (this.startPixel_) {
        this.dispatchEvent(
          new DragBoxEvent(DragBoxEventType.BOXCANCEL, this.startPixel_, null),
        );
        this.startPixel_ = null;
      }
```

ahocevar (changes requested):
Thanks, @behoney! I think the BOXCANCEL event should only be fired when a box is currently being drawn, so I added a suggestion to change this. Looks good otherwise.

src/ol/interaction/DragBox.js
behoney: Thanks, @ahocevar 😃 
I pushed https://github.com/openlayers/openlayers/pull/15671/commits/07873fbde1465d13f6d2852be132bf373967b2a0 which includes suggested changes.

src/ol/interaction/DragBox.js
```diff
@@ -269,6 +277,26 @@ class DragBox extends PointerInteraction {
...
+
+  /**
+   * Activate or deactivate the interaction.
+   * @param {boolean} active Active.
+   * @observable
+   * @api
+   */
+  setActive(active) {
+    if (!active) {
+      this.box_.disposeInternal();
```
M393, line 289:
```suggestion
      this.box_setMap(null);
```
`disposeInternal` is called by `dispose` to permanently destroy an object. While it would do the job here it makes more sense to me to call `this.setMap(null)` directly here.

behoney: I see.. I missed what `dispose` strictly means, and I agree with calling `this.setMap(null)` instead.

https://github.com/openlayers/openlayers/pull/15671/commits/bcd923f7fa62f3b9eb3ff2a341078036b857ff7b includes this suggestion. Thanks!

ahocevar (approved):
Thanks, @behoney

PR #15688, opened 2024-03-30
Update MapTiler streets url

mike-000:
As noted in https://github.com/openlayers/openlayers/pull/15669#discussion_r1544490546 multiple examples are using an outdated url.


ahocevar (approved):
Thanks, @mike-000

PR #15713, opened 2024-04-09
Fix VectorImageLayer options types

### LINKED ISSUE #15703: VectorImageLayer options don't accept flat styles
**Describe the bug**

VectorLayer options type accepts a `StyleLike | FlatStyleLike | null | undefined`.  
VectorImageLayer options type accepts a `StyleLike | null | undefined`.

https://github.com/openlayers/openlayers/blob/474b20135ea630c49a4cc6a640d78e233c54bee6/src/ol/layer/BaseVector.js#L51
https://github.com/openlayers/openlayers/blob/474b20135ea630c49a4cc6a640d78e233c54bee6/src/ol/layer/VectorImage.js#L43

However, VectorImageLayer seems to actually support flat styles, so it looks like `FlatStyleLike` is just missing from the type definition.

**To Reproduce**

```ts
new VectorImageLayer({
    style: {
        "fill-color": "#0000ff11",
        "stroke-color": "blue"
    } // as any <- forcing it to accept the style works
});
```
`Object literal may only specify known properties, and '"fill-color"' does not exist in type 'StyleLike'.ts(2353)`

**Expected behavior**
Both VectorLayer and VectorImageLayer options should accept flat styles.

janruo:
VectorImageLayer supports mostly the same options as VectorLayer, because they both extend from BaseVectorLayer.

`declutter`, `style` (were too narrow) and `background` (was missing) have been synced from the base class.

Resolves #15703

### DIFF
--- a/src/ol/layer/VectorImage.js
+++ b/src/ol/layer/VectorImage.js
@@ -35,14 +35,16 @@ import CanvasVectorImageLayerRenderer from '../renderer/canvas/VectorImageLayer.
  * this layer in its layers collection, and the layer will be rendered on top. This is useful for
  * temporary layers. The standard way to add a layer to a map and have it managed by the map is to
  * use [map.addLayer()]{@link import("../Map.js").default#addLayer}.
- * @property {boolean} [declutter=false] Declutter images and text on this layer. The priority
- * is defined by the `zIndex` of the style and the render order of features. Higher z-index means higher
- * priority. Within the same z-index, a feature rendered before another has higher priority. Items will
- * not be decluttered against or together with items on other layers with the same `declutter` value. If
- * that is needed, use {@link import("../layer/Vector.js").default} instead.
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @janruo

janruo:
@ahocevar The declutter option description I copied from the base layer is not quite accurate after all. I only tested it superficially (i.e. looks about right and doesn't crash). But now I noticed that multiple VectorImageLayers won't declutter together like VectorLayers do, when given the same declutter value. It only declutters inside the layer.

Is that a bug in VectorImageLayer or is that how it's supposed to work?

ahocevar:
That's because a VectorImageLayer creates its own frameState for the vector renderer. I think the behavior could be changed to do what the docs say after your change without too much effort. So if you want to give it a try, please go ahead. Otherwise, please submit another pull request to fix the docs.

PR #15724, opened 2024-04-11
Fix VectorImageLayer declutter description

janruo:
Restores the previous description with the addition of "Any truthy value will enable decluttering" due to the type being wider now.

https://github.com/openlayers/openlayers/pull/15713#issuecomment-2047383187

I think it would make more sense if it worked the same way as VectorLayer, but I don't know the code well enough to try that.


ahocevar (approved):
Thanks, @janruo

PR #15730, opened 2024-04-12
Add Snap Data: feature

### LINKED ISSUE #15728: Snap interaction option edges doesn't return feature
**Describe the bug**
Snap interaction option edges doesn't return the closest feature like option vertex does.

**To Reproduce**
Here is the code : 
`const snapInteraction = new Snap({
        source: vectorSource,
        vertex: false,
        edge: true,
      });`
`snapInteraction.on("snap", (event) => {
      console.log(event);
    });`

When I snap to a polygon, it returns no feature : 
image

**Expected behavior**
If I set vertex to true, it returns the snapped feature : 
image

**Solution proposal**
In /ol/interaction/Snap.js file, if vertex, the closestFeature is set but if edge, the closestFeature is not set : 
image
image
Is it possible to add closestFeature for edge option ?

CarolineDelerue:
Fixes #15728

~~Add closestSegment for vertex option.~~
Add closestFeature for edge option.

### DIFF
--- a/src/ol/interaction/Snap.js
+++ b/src/ol/interaction/Snap.js
@@ -510,6 +510,7 @@ class Snap extends PointerInteraction {
               closestVertex = vertex;
               minSquaredDistance = delta;
               closestFeature = segmentData.feature;
+              closestSegment = segmentData.segment;
             }
           });
         }
@@ -554,6 +555,7 @@ class Snap extends PointerInteraction {
                 ? null
                 : segmentData.segment;
             minSquaredDistance = delta;
+            closestFeature = segmentData.feature;
           }
         }
       }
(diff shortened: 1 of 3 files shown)


src/ol/interaction/Snap.js
```diff
@@ -510,6 +510,7 @@ class Snap extends PointerInteraction {
               closestVertex = vertex;
               minSquaredDistance = delta;
               closestFeature = segmentData.feature;
+              closestSegment = segmentData.segment;
```
ahocevar, line 513: I don't think you want to return the closest segment when you've snapped to a vertex. Some users need a distinction between a vertex snap and a segment snap, and if you return the closest segment on a vertex snap it will be much harder to detect that. And it would be an undesired breaking change.

ahocevar (approved):

src/ol/interaction/Snap.js
CarolineDelerue: Ok I have removed it.

ahocevar:
Thanks, @CarolineDelerue. I updated the tests and added an upgrade note about this change.

PR #15738, opened 2024-04-14
Use outdoor instead of streets style for site and mapbox-style example

ahocevar:
Makes the site's top banner map more appealing again, and brings the mapbox-style example closer to what it was before #15725.


ahocevar:
@bolollo Ok for you if we make this change?

tschaub (approved):

PR #15762, opened 2024-04-22
Show how to combine a layer with a render function and a standard layer

ahocevar:
See #15758. This example change should make life easier for users who want to combine standard layers with a layer with a custom render function.

### DIFF
--- a/examples/svg-layer.js
+++ b/examples/svg-layer.js
@@ -1,6 +1,10 @@
+import Feature from '../src/ol/Feature.js';
 import Layer from '../src/ol/layer/Layer.js';
 import Map from '../src/ol/Map.js';
+import VectorSource from '../src/ol/source/Vector.js';
 import View from '../src/ol/View.js';
+import {Point} from '../src/ol/geom.js';
+import {Vector} from '../src/ol/layer.js';
 import {composeCssTransform} from '../src/ol/transform.js';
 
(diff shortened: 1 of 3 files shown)


ahocevar:
Preview of the updated example: https://deploy-preview-15762--ol-site.netlify.app/en/latest/examples/svg-layer.html

examples/svg-layer.js
```diff
@@ -47,8 +52,21 @@ map.addLayer(
         center[1] / svgResolution - height / 2,
       );
       svgContainer.style.transform = cssTransform;
-      svgContainer.style.opacity = this.getOpacity();
+      svgContainer.style.opacity = this.getOpacity() || 1;
```
MoonE, line 55: Why the ` || 1`? 
`this.getOpacity()`  always returns 1 here; and what if it opacity is zero?

MoonE (approved):
Works, just not sure what's behind the opacity change.

examples/svg-layer.js
ahocevar: This change avoids that the style ends up being the string `'undefined'`, which is not valid CSS.

ahocevar: You were absolutely right, @MoonE . I undid the opacity change. Thanks for the sharp eye.

PR #15765, opened 2024-04-23
Remove unused layer property

M393:
This wasn't accessed anywhere else.


ahocevar (approved):
Good find, thanks @M393

PR #15767, opened 2024-04-23
Better generics for vector and vectortile layers

### LINKED ISSUE #15764: Potential type bug in vector layer types in combination with render features
**Describe the bug**
I'm looking into ways to improve performance of large geojson sets.
I found the RenderFeature in the API docs and wanted to give it a try, but although I can create a vector source with it, the types of the vector layer do not accept this source.

**To Reproduce**
In a typescript project, create:
```
const source = new VectorSource<RenderFeature>();
const layer = new VectorLayer({ source });
```

Then you'll get a type error: `Type 'VectorSource<RenderFeature>' is not assignable to type 'VectorSource<Feature<Geometry>>'.`

Either I'm not understanding how to use the RenderFeature, or the type is too strict.

ahocevar:
Fixes #15764. In addition, it also makes `vectorTileLayer.getSource().getFeatures()` infer the correct type (Feature or RenderFeature)

### DIFF
--- a/src/ol/layer/BaseVector.js
+++ b/src/ol/layer/BaseVector.js
@@ -13,7 +13,7 @@ import {
 } from '../render/canvas/style.js';
 
 /**
- * @template {import("../source/Vector.js").default|import("../source/VectorTile.js").default} VectorSourceType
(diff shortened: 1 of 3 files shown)


MoonE (approved):
Thanks, @ahocevar

PR #15768, opened 2024-04-23
Rebuild executors when declutter state changes

### LINKED ISSUE #15766: map.removeLayer() on a VectorLayer with `declutter` causes other vector layers to disappear
**Describe the bug**

As of recent builds calling `map.removeLayer()` on a VectorLayer when `declutter` is enabled causes all vector layers to disappear.

**To Reproduce**
Steps to reproduce the behavior
(Apologies for not linking directly to an example as CodeSandbox won't let users create their own drafts anymore without paying)

1. Go to https://openlayers.org/en/latest/examples/vector-labels.html
2. Edit codesandbox example to add `declutter: true` to vectorPoints layer definition. Then add the following at the end of the example code:
```js
 setTimeout(() => {
  map.removeLayer(vectorPoints)
}, 2000);
```
3. Run example, and watch layers disappear after 2 seconds when `map.removeLayer` is called

**Expected behavior**
Only the requested layer should disappear from map

ahocevar:
Fixes #15766

### DIFF
--- a/src/ol/renderer/canvas/VectorLayer.js
+++ b/src/ol/renderer/canvas/VectorLayer.js
@@ -128,6 +128,12 @@ class CanvasVectorLayerRenderer extends CanvasLayerRenderer {
 
     /**
      * @private
+     * @type {boolean}
+     */
+    this.renderedFrameDeclutter_;
+
+    /**
+     * @private
      * @type {import("../../render/canvas/ExecutorGroup").default}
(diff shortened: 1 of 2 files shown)


M393 (approved):
This fixes the issue I have, Thanks, @ahocevar

MoonE (approved):
Looks fine.

PR #15774, opened 2024-04-25
Add feature type to vector source signature

sebakerckhof:
Found another case where the feature type generic was not taken into account


ahocevar (approved):
Thanks, @sebakerckhof

ahocevar (changes requested):
Actually I think this will not work. The `this.on` and `this.once` signatures would need to include the FeatureClass template of the class, otherwise it cannot be inferred.

ahocevar (approved):
Actually it does work. Thanks!

PR #15779, opened 2024-04-29
Accept render features source in WebGLPointsLayer

sebakerckhof:
Seems like the WebGLPointsLayer was not included here: https://github.com/openlayers/openlayers/pull/15767 . But AFAICT this should also work with sources with render features.


ahocevar (approved):
Thanks, @sebakerckhof

PR #15785, opened 2024-04-30
Fix WebGL VectorLayer wrapX renderer

### LINKED ISSUE #15674: WebGL Vector Layer wrapX support
Hi guys, as in a merged pull request that WebGLPointsLayer wrapX support would be nice to have the same using WebGL Vector Layer.

Example:
https://openlayers.org/en/latest/examples/webgl-vector-layer.html

Thanks

sakitam-fdd:
Fixes #15674


ahocevar (approved):
Thanks, @sakitam-fdd

PR #15789, opened 2024-05-01
Improve performance of deferred rendering

### LINKED ISSUE #15772: VectorTile renderMode vector Firefox regression
**Describe the bug**
Firefox gets pretty much stuck when using VectorTile renderMode vector. It looks like GPU utilization goes 100% and won't come down. Tested on Windows and Mac. Chrome, Edge and Safari seems to work ok but maybe get a little bit slower. Issue is present on OL 9.1.0 and it was working fine in OL 8.2.0.

**To Reproduce**
1. See attachment zip -> Extract -> npm install -> npm start (It is fork from https://openlayers.org/en/latest/examples/canvas-tiles-tms.html with added renderMode:vector)
2. Zoom all the way up -> zoom all the way in -> zoom all up -> zoom all in -> Try to move map
3. Browser becomes unresponsive

**Expected behavior**
Map should work without jamming browser with normal usecase

[ol-910-firefox-mvt-rendermode-bug.zip](https://github.com/openlayers/openlayers/files/15104296/ol-910-firefox-mvt-rendermode-bug.zip)

ahocevar:
This pull request improves performance of deferred rendering:
* Avoid duplicate rendering of the same zIndex: when collecting the executors of all tiles, a zIndex entry was added for each tile, instead of just one
* Reduce garbage creation by reusing the instructions arrays in `ZIndexContext`
* When no decluttering needs to be done, run the executor directly in the ZIndexContext, instead of creating canvas instructions

Fixes #15772

### DIFF
--- a/src/ol/render/canvas/Executor.js
+++ b/src/ol/render/canvas/Executor.js
@@ -1175,13 +1175,9 @@ class Executor {
           dd = /** @type {number} */ (instruction[2]);
           x = pixelCoordinates[d];
           y = pixelCoordinates[d + 1];
-          roundX = (x + 0.5) | 0;
-          roundY = (y + 0.5) | 0;
-          if (roundX !== prevX || roundY !== prevY) {
-            context.moveTo(x, y);
-            prevX = roundX;
-            prevY = roundY;
-          }
+          context.moveTo(x, y);
+          prevX = (x + 0.5) | 0;
+          prevY = (y + 0.5) | 0;
           for (d += 2; d < dd; d += 2) {
             x = pixelCoordinates[d];
             y = pixelCoordinates[d + 1];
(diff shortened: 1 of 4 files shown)


src/ol/render/canvas/Executor.js
```diff
@@ -1175,13 +1175,9 @@ class Executor {
...
-          roundX = (x + 0.5) | 0;
-          roundY = (y + 0.5) | 0;
-          if (roundX !== prevX || roundY !== prevY) {
-            context.moveTo(x, y);
-            prevX = roundX;
-            prevY = roundY;
-          }
+          context.moveTo(x, y);
+          prevX = (x + 0.5) | 0;
+          prevY = (y + 0.5) | 0;
```
ahocevar, line 1180: In addition to what I wrote in the description, this removes an if condition that was never `false`.

tschaub (approved):

Issue #15791, opened 2024-05-03
Support OGC API Tiles "Collections Selection" Conformance Class


jansule:
**Is your feature request related to a problem? Please describe.**

The OGC API Tiles Part 1 - Core describes the requirement class "Collections Selection" which allows for selecting collections via query parameter, when requesting tiles from a service.

See also https://docs.ogc.org/is/20-057/20-057.html#toc49

**Describe the solution you'd like**

It would be great to add support for this to the [OGCVectorTile Source](https://openlayers.org/en/v8.1.0/apidoc/module-ol_source_OGCVectorTile-OGCVectorTile.html)

jansule:
Solved in https://github.com/openlayers/openlayers/pull/15792

PR #15792, opened 2024-05-03
feat: add collections option to OGCVectorTile source

jansule:
This adds support for the OGC API Tiles  Part 1 - Core "Collections Selection" conformance class for the OGCVectorTile source. 

This solves https://github.com/openlayers/openlayers/issues/15791

### DIFF
--- a/src/ol/source/ogcTileUtil.js
+++ b/src/ol/source/ogcTileUtil.js
@@ -194,14 +194,7 @@ export function getVectorTileUrlTemplate(
     // commata in the identifiers of the `collections` query parameter
     // need to be URLEncoded, while the commata separating the identifiers
     // should not.
-    let url;
-    let isAbsolute = true;
-    try {
-      url = new URL(tileUrlTemplate);
-    } catch (e) {
-      isAbsolute = false;
-      url = new URL(tileUrlTemplate, 'https://example.com');
-    }
+    const url = new URL(tileUrlTemplate, window.location.href);
 
     if (url.pathname.split('/').includes('collections')) {
       logError(
@@ -213,12 +206,7 @@ export function getVectorTileUrlTemplate(
         .join(',');
 
       url.searchParams.append('collections', encodedCollections);
-      let baseUrl;
-      if (isAbsolute) {
-        baseUrl = decodeURI(url.toString()).split('?')[0];
-      } else {
-        baseUrl = decodeURIComponent(url.pathname);
-      }
+      const baseUrl = url.origin + decodeURIComponent(url.pathname);
       const queryParams = decodeURIComponent(url.searchParams.toString());
       tileUrlTemplate = `${baseUrl}?${queryParams}`;
     }
--- a/src/ol/source/ogcTileUtil.js
+++ b/src/ol/source/ogcTileUtil.js
@@ -194,7 +194,14 @@ export function getVectorTileUrlTemplate(
     // commata in the identifiers of the `collections` query parameter
     // need to be URLEncoded, while the commata separating the identifiers
     // should not.
-    const url = new URL(tileUrlTemplate, window.location.href);
+    let url;
+    let isAbsolute = true;
+    try {
+      url = new URL(tileUrlTemplate);
+    } catch (e) {
+      isAbsolute = false;
+      url = new URL(tileUrlTemplate, 'https://example.com');
+    }
 
     if (url.pathname.split('/').includes('collections')) {
       logError(
@@ -206,7 +213,12 @@ export function getVectorTileUrlTemplate(
         .join(',');
 
       url.searchParams.append('collections', encodedCollections);
-      const baseUrl = url.origin + decodeURIComponent(url.pathname);
+      let baseUrl;
+      if (isAbsolute) {
+        baseUrl = decodeURI(url.toString()).split('?')[0];
+      } else {
+        baseUrl = decodeURIComponent(url.pathname);
+      }
       const queryParams = decodeURIComponent(url.searchParams.toString());
       tileUrlTemplate = `${baseUrl}?${queryParams}`;
     }
--- a/src/ol/source/ogcTileUtil.js
+++ b/src/ol/source/ogcTileUtil.js
@@ -6,6 +6,7 @@ import TileGrid from '../tilegrid/TileGrid.js';
 import {getJSON, resolveUrl} from '../net.js';
 import {get as getProjection} from '../proj.js';
 import {getIntersection as intersectExtents} from '../extent.js';
+import {error as logError} from '../console.js';
 
 /**
  * See https://ogcapi.ogc.org/tiles/.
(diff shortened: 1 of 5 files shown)


src/ol/source/ogcTileUtil.js
```diff
@@ -184,6 +188,42 @@ export function getVectorTileUrlTemplate(
...
+    // need to be URLEncoded, while the commata separating the identifiers
+    // should not.
+    let url;
+    let isAbsolute = true;
+    try {
+      url = new URL(tileUrlTemplate);
+    } catch (e) {
+      isAbsolute = false;
+      url = new URL(tileUrlTemplate, 'https://example.com');
+    }
```
ahocevar, line 204: Instead of using try/catch and differentiating between relative and absolute URLs, you could simplify this to
```suggestion
    const url = new URL(tileUrlTemplate, window.location.href);
```

src/ol/source/ogcTileUtil.js
```diff
@@ -184,6 +188,42 @@ export function getVectorTileUrlTemplate(
...
+        .map((c) => encodeURIComponent(c))
+        .join(',');
+
+      url.searchParams.append('collections', encodedCollections);
+      let baseUrl;
+      if (isAbsolute) {
+        baseUrl = decodeURI(url.toString()).split('?')[0];
+      } else {
+        baseUrl = decodeURIComponent(url.pathname);
+      }
```
ahocevar, line 221: And now that we always have a correct absolute URL, you can simplify this too:
```suggestion
      const baseUrl = url.origin + decodeURIComponent(url.pathname);
```

ahocevar (changes requested):

jansule:
Good points @ahocevar. Applied the suggestions.

jansule:
While having another look at it, it looks like the suggested approach does change the semantics a little in the sense that previously we did not add a domain for the case that `link.href` did not provide one. A href without domain should be prepended by the domain name of the service and not `window.location`. The service domain will also already be prepended by https://github.com/openlayers/openlayers/blob/main/src/ol/source/ogcTileUtil.js#L341.

So I think the initial approach should be used instead.

ahocevar:
@jansule Can you elaborate on what the difference would be? If `link.href` is a href without an origin, the resulting requests would be exactly the same as `new URL(link.href, window.location.href)`.

The reason for the failing test is that Node does not have a `window` object. So you would have to move that test to the browser tests.

jansule:
> @jansule Can you elaborate on what the difference would be? If link.href is a href without an origin, the resulting requests would be exactly the same as new URL(link.href, window.location.href).

If `link.href` is without an origin, it relates to the document it belongs to (the service) and not to the document our JavaScript code is running. E.g. when my OL application is running on `https://my-app.com` and I want to include a OGCVectorTile from https://maps.gnosis.earth/ogcapi/collections/Daraa/tiles/WebMercatorQuad?f=json , which provides following link item:

```
{
  rel: "item",
  type: "application/vnd.mapbox-vector-tile",
  title: "WebMercatorQuad multi-layer vector tiles for Daraa (as Mapbox Vector Tiles)",
  href: "/ogcapi/collections/Daraa/tiles/WebMercatorQuad/{tileMatrix}/{tileRow}/{tileCol}.mvt",
  templated: true
}
``` 
The url should always be resolved to `https://maps.gnosis.earth/ogcapi/collections/Daraa/tiles/WebMercatorQuad/{tileMatrix}/{tileRow}/{tileCol}.mvt` and not to `https://my-app.com/ogcapi/collections/Daraa/tiles/WebMercatorQuad/{tileMatrix}/{tileRow}/{tileCol}.mvt`.

Resolving urls without a domain in relation to its service will already be done by https://github.com/openlayers/openlayers/blob/main/src/ol/source/ogcTileUtil.js#L341.

So using `window.location.href` does not make sense here, as it points to the wrong reference (the application, not the service).

ahocevar:
Got it, thanks for the explanation. Then instead of `window.location.href`, you just have to use the document url (`sourceInfo.url` in `getTileSetInfo()`) as second argument to the URL constructor.

One more thing: isn't the collections option also specified for OGCMapTile?

ahocevar:
But you're right, @jansule. Looking at the rest of the OGC API code, this should also work with local files, so using the URL constructor like in my suggestion indeed isn't the best thing to do.

jansule:
> One more thing: isn't the collections option also specified for OGCMapTile?

Yes it is. I thought it would be better to do separate PRs for that.

> Looking at the rest of the OGC API code, this should also work with local files, so using the URL constructor like in my suggestion indeed isn't the best thing to do.

So should I just revert the last commit, or should I check again to see if I can simplify the code a little?

ahocevar:
> So should I just revert the last commit, or should I check again to see if I can simplify the code a little?

I think you should at least get rid of the try/catch. Maybe not using the URL constructor at all would yield simpler code.

Regarding OGCMapTile: since it's the same code, you might add that in this pull request, and create a common function instead of duplicating the code there.

ahocevar:
@jansule I think something like this might be a good solution:
```js
  if (collections && collections.length) {
    // According to conformance class
    // http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/collections-selection
    // commata in the identifiers of the `collections` query parameter
    // need to be URLEncoded, while the commata separating the identifiers
    // should not.
    const url = new URL(tileUrlTemplate, 'file:/');
    if (url.pathname.split('/').includes('collections')) {
      logError(
        'The "collections" query parameter cannot be added to collection endpoints',
      );
      return tileUrlTemplate;
    }
    const encodedCollections = collections
      .map((c) => encodeURIComponent(c))
      .join(',');

    url.searchParams.append('collections', encodedCollections);
    const baseUrl = tileUrlTemplate.split('?')[0];
    const queryParams = decodeURIComponent(url.searchParams.toString());
    return `${baseUrl}?${queryParams}`;
  }
```
It does use the URL constructor, but takes the `baseUrl` directly from the original template URL.

jansule:
Yes, came to the same solution :grinning: Do you know any example service that provides dataset tilesets (`/map/tiles`) that I can use for the OGCMapTile tests? The example services on https://tiles.developer.ogc.org/ only provide geoData tilesets (`/collections/{collectionId}/map/tiles`).

src/ol/source/ogcTileUtil.js
```diff
@@ -96,14 +97,50 @@ const knownVectorMediaTypes = {
...
+ * @param {Array<string>} collections List of collections to include as query parameter.
+ * @return {string} The tile URL template with appended collections query parameter.
+ */
+export function appendCollectionsQueryParam(tileUrlTemplate, collections) {
+  if (!collections.length) {
+    return tileUrlTemplate;
+  }
+
+  // making sure we can always construct a URL instance.
+  const url = new URL(tileUrlTemplate, 'file://');
```
ahocevar, line 114: Minor but one slash is enough:

```suggestion
  const url = new URL(tileUrlTemplate, 'file:/');
```

ahocevar (approved):
Thanks, @jansule !

src/ol/source/ogcTileUtil.js
ahocevar: Oops. I forgot to wait for this. Maybe something for a quick follow-up fix.

jansule: Created a follow up PR: https://github.com/openlayers/openlayers/pull/15809

Issue #15801, opened 2024-05-06
Regression: Match operator input type not correctly inferred since v8.2.0, when using strings that happen to be colors as well


sebakerckhof:
**Describe the bug**
In openlayers 8.1.0, if I have a literal (webgl) style that includes:
```
 'icon-color': [
    'match',
    ['get', 'foo'],
    'red',
    [245, 33, 45, 0.5],
    'green',
    [83, 196, 26, 0.5],
    [255, 255, 255, 0.5],
  ],
```
This worked as expected since the input type of the 'foo' variable was inferred to be a string by looking at the possible values (red/green -> string | color) and the accepted type (string | number | boolean). 
Relevant code: https://github.com/openlayers/openlayers/blob/v8.1.0/src/ol/style/expressions.js#L1205-L1206

Since openlayers 8.2.0, this does not work anymore.  More specifically since commit de0c8d8afaf6b0594f13328b668cef7b0c4eb0ce , which was part of this PR: https://github.com/openlayers/openlayers/pull/15121 .

The reason for that is that although a check was introduced ( https://github.com/openlayers/openlayers/blob/v9.1.0/src/ol/expr/expression.js#L861-L862 ), it no longer forces the type in case of it being ambiguous. And ambiguous types pass the overlap test. But since we know the type can not be a color, it must be a string and we should treat it as such.

PR #15802, opened 2024-05-06
fix: webgl style regression - force ambiguous match input type to an accepted type

sebakerckhof:
Fix for https://github.com/openlayers/openlayers/issues/15801 , which contains more context


ahocevar (approved):
Thanks, @sebakerckhof

PR #15808, opened 2024-05-08
fix: handle errors in response parsing in xhr loader

sebakerckhof:
Although I'm not sure if/what the OWS spec says about this, but at least the popular geoserver will return status 200 even when the result is an OWS exception report.

E.g. If I make a WFS call that fails in geoserver. I'll get a status 200 with an ows exception report XML.
e.g.
```xml
<?xml version="1.0" encoding="UTF-8"?><ows:ExceptionReport xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ows="http://www.opengis.net/ows" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0.0" xsi:schemaLocation="http://www.opengis.net/ows localhost/geoserver/schemas/ows/1.0.0/owsExceptionReport.xsd">
  <ows:Exception exceptionCode="NoApplicableCode">
    <ows:ExceptionText>We have had issues trying to flip axis of GEOGCS["GCS_WGS_1984", ...</ows:ExceptionText>
  </ows:Exception>
</ows:ExceptionReport>
```

The problem is that the xhr loader does not account for this and will throw an uncaught error when trying to parse it as json.
Therefore the failure/success callbacks are never called and the source stays in loading state (and in turn rendercomplete is not called)

This PR tries to solve this by putting wrapping the entire data parsing in a try-catch

### DIFF
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -89,18 +89,10 @@ export function loadFeaturesXhr(
       try {
         /** @type {Document|Node|Object|string|undefined} */
         let source;
-        if (type == 'json') {
-          source = JSON.parse(xhr.responseText);
-        } else if (type == 'text') {
+        if (type == 'text' || type == 'json') {
           source = xhr.responseText;
         } else if (type == 'xml') {
-          source = xhr.responseXML;
-          if (!source) {
-            source = new DOMParser().parseFromString(
-              xhr.responseText,
-              'application/xml',
-            );
-          }
+          source = xhr.responseXML || xhr.responseText;
         } else if (type == 'arraybuffer') {
           source = /** @type {ArrayBuffer} */ (xhr.response);
         }
--- a/src/ol/featureloader.js
+++ b/src/ol/featureloader.js
@@ -86,35 +86,39 @@ export function loadFeaturesXhr(
     // status will be 0 for file:// urls
     if (!xhr.status || (xhr.status >= 200 && xhr.status < 300)) {
       const type = format.getType();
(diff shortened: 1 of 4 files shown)


src/ol/featureloader.js
```diff
@@ -86,35 +86,39 @@ export function loadFeaturesXhr(
...
+            /** @type {Array<FeatureType>} */
+            (
+              format.readFeatures(source, {
+                extent: extent,
+                featureProjection: projection,
+              })
+            ),
+            format.readProjection(source),
           );
+        } else {
```
sebakerckhof, line 118: Striclty speaking this `if-else` is not needed anymore, since a falsy source would throw an error which is then caught and calls `failure();`, but it seemed cleaner to keep it.

ahocevar (commented):
@sebakerckhof A slightly related note: for GeoServer, when requesting WFS in JSON format, you should configure it to also return exceptions in JSON. See https://docs.geoserver.org/latest/en/user/services/wfs/reference.html#exceptions. That does not change the response status code though.

It would be good if you could change the test to use the JSON version of the exception.

sebakerckhof:
Ah, I didn't know about that option, thanks. But indeed, it doesn't change the need for the try-catch that this PR adds.

I added a test for the json exception, but kept the xml test as well, since it's not wrong for the server to send an xml report from what I got from the spec.

The redundant parsing is removed now.

ahocevar (approved):
Thanks, @sebakerckhof

PR #15809, opened 2024-05-08
refactor: remove unneeded '/' for URL construction

jansule:
This is a very minor follow up of https://github.com/openlayers/openlayers/pull/15792 that removes the unneeded second `/` when constructing the helper URL instance for appending the collections parameter on OGC API Layers. See also https://github.com/openlayers/openlayers/pull/15792#discussion_r1592822864


ahocevar:
Thanks, @jansule !

ahocevar (approved):

PR #15813, opened 2024-05-09
Support nested properties with the 'get' expression

### LINKED ISSUE #15505: Accessing nested properties in flat style definitions

We're using OpenLayers extensively in our project. First of all: thank you for this amazing library!

At the moment we're exploring options where our users may be able to customize or define their own styles for layers. We're looking into defining all our available layers, as well as custom data, in a database, and sending flat styles over the wire from our API backend instead of hardcoding `new Style(...)` in code.

Things are looking promising, but we have quite a few layers with properties which are actually arbitrarily deeply nested objects. For instance:

```typescript
const connection = new Feature();
connection.set('demand', '100KW');
connection.set('year_of_construction', '1927');
connection.set('address', {
  country: 'The Netherlands',
  province: 'Gelderland',
  municipality: 'Nijmegen',
  // etc
});
```

Say we want to use `address.province` in a `filter` expression, or a `get` expression. As far as I can tell, this is currently not supported, and only expressions where `feature.get('key')` which return a JS literal can be used.

Hopefully I'm wrong and this is possible. If not, I've been diving into the code in `src/ol/expression.js`, and I propose a new `nested` expression. Given the previous `address` feature, something like:

```typescript
const layer = new VectorLayer({
  style: {
    'text-value': ['get', ['nested', 'address', 'province']]
  }
});
```

Or a bit more formally:

```
['nested', key1, key2 ... keyN]: Traverses an object of type T where key1 extends keyof T, key2 extends keyof T[key1], keyN extends T[keyN-1] and returns the value at path N.
```

`Op.Nested`, when walked down the JS object path, should return a `LiteralExpression` (I think). This *should* also work out of the box in any other expressions or operators where a literal is expected.

Please let me know if this is something you may wish to pursue. I think, with a bit more time spent browsing through the code base in `expression.js` and `cpu.js`, I should be able to implement this and file a Pull Request.

ahocevar:
Fixes #15505.


tschaub:
Nice!  Thanks for taking this on.

I realize this may sound stubborn, but I’d like to see a real example where type hints are needed before introducing another operator that we might just remove later.  By “a real example” I mean an expression used in a style object.

ahocevar:
@tschaub Actually it's the ambiguity of size and scale properties, which use `NumberType | NumberArrayType` as type. This could be fixed by introducing a `SizeType`. Once we have that, we can get rid of the type hints. I'm going to create a separate pull request with that change.

ahocevar:
> @tschaub Actually it's the ambiguity of size and scale properties, which use `NumberType | NumberArrayType` as type. This could be fixed by introducing a `SizeType`. Once we have that, we can get rid of the type hints. I'm going to create a separate pull request with that change.

See #15818.

ahocevar:
Tests will pass once #15818 is merged and this pull request is rebased.

ahocevar:
Rebased again, without a merge commit. Thanks for any review.

changelog/upgrade-notes.md
```diff
@@ -2,6 +2,18 @@
 
 ### 9.2.0
 
+#### Flat styles: Removal of Type hints in `'get'` expressions
```
tschaub, line 5: Looks like this should go under a "next release" heading instead of 9.2.0.

ahocevar: Fixed.

changelog/upgrade-notes.md
```diff
@@ -1,6 +1,18 @@
...
+
+These type hints are no longer needed. If you were previously using a type hint in a get expression, you can change the expression from e.g.
+```js
+['get', 'foo', 'number[]']
+```
+to
+```js
+['get', 'foo']
+```
+For the Canvas renderer, additional arguments to the `'get'` call expression now mean access to nested properties. Property access will only be attempted when the `typeof` the target is `'object'`, so existing applications will not break.
```
tschaub, line 16:
```suggestion
### 9.2.0

```

src/ol/expr/cpu.js
```diff
@@ -256,7 +256,22 @@ function compileAccessorExpression(expression, context) {
   const name = /** @type {string} */ (nameExpression.value);
   switch (expression.operator) {
     case Ops.Get: {
-      return (context) => context.properties[name];
+      return (context) => {
+        const args = expression.args;
+        let value = context.properties[name];
+        for (let i = 1, ii = args.length; i < ii; ++i) {
+          if (typeof value !== 'object') {
```
tschaub, line 263: Should we continue on `null`?

ahocevar: Good point, I‘ll add a null check tomorrow.

src/ol/expr/cpu.js
```diff
@@ -256,7 +256,22 @@ function compileAccessorExpression(expression, context) {
...
+        const args = expression.args;
+        let value = context.properties[name];
+        for (let i = 1, ii = args.length; i < ii; ++i) {
+          if (value === null || typeof value !== 'object') {
+            break;
+          }
+          const keyExpression = /** @type {LiteralExpression} */ (args[i]);
+          const key = /** @type {string|number} */ (keyExpression.value);
+          if (typeof key !== 'number' && Array.isArray(value)) {
+            break;
```
M393, line 269: Looks like with properties `{"a": 1}` and `["get", "a", "b"]` it would currently return `1`?
Same for properties `{"a": [1]}` and `["get", "a", "b"]` would return `[1]`?

Should this return `undefined` instead?

ahocevar: Returning `undefined` in such cases was my initial approach, but then I thought it might be useful to allow data structures with variable nesting depth - and be backwards compatible with the removed type hint as last argument.

If people think there is no practical use of this, and it rather adds confusion, I'll be happy to go back to the initial approach and make this a breaking change.

changelog/upgrade-notes.md
```diff
@@ -1,5 +1,20 @@
...
+### 9.2.0
```
tschaub, line 16:
```suggestion
```

We'll get it right one of these times :)

src/ol/expr/cpu.js
tschaub: I think we should move on from the type hint.

ahocevar: Ok, updated. Now it's a breaking change.

src/ol/expr/expression.js
```diff
@@ -668,19 +634,29 @@ const parsers = {
...
+        );
+      }
+      continue;
+    }
+    if (typeof arg.value !== 'string') {
+      throw new Error(
+        'Expected the attribute name of a get operation to be a string',
+      );
+    }
+    context.properties.add(String(arg.value));
```
tschaub, line 657: We aren't leveraging the `properties` of the parsing context right now in Canvas, but here is the idea.  As we parse style expressions, we build up a lookup of all the data that is needed from a feature (feature id, geometry, properties).  Then during the evaluation phase, we could supply a minimal set of data in the evaluation context.  

Currently, we don't do this in Canvas.  Instead we add [all the feature properties](https://github.com/openlayers/openlayers/blob/72c3b7ff61921aa7b0551f3c833ea8c4e68336a3/src/ol/render/canvas/style.js#L78) to the evaluation context.  If we were passing the data around to/from workers, we might instead want to encode a subset of the data for that purpose.

In the WebGL renderer, the `properties` from the parsing context is passed to the evaluation context [and then used](https://github.com/openlayers/openlayers/blob/72c3b7ff61921aa7b0551f3c833ea8c4e68336a3/src/ol/render/canvas/style.js#L78) when building shaders and uploading attributes to the gpu.

To support working with a subset of feature properties in the Canvas renderer in the future, we want to keep track of these "property paths" that are used in style expressions.  Now that we support deeply nested paths, instead of the parsing context properties being `Set<string>` it could be `Array<Array<string>>` (where a single entry might be `['path', 'to', 'property']`.  Since deeply nested paths aren't supported in WebGL yet, this would break things on that side.  For now, since we aren't using the parsing context `properties` in Canvas, we could just store a dot delimited path to all the properties (e.g. `path.to.properties`).  This isn't the best long-term solution if people do crazy things like put dots in their property names, but it could be fine for now.

The tl;dr is that we don't want to add 'path', 'to', and 'property' to this `context.properties` set.  Instead we could add 'path.to.property' (and then later change the type when we want to fully support nested properties in both Canvas and WebGL).

tschaub:
Thanks for your work on this, @ahocevar.  It feels like a simpler and more functional solution now (together with #15818).

Hope my [comment above](https://github.com/openlayers/openlayers/pull/15813#pullrequestreview-2074810984) makes sense.  Since we aren't currently using the parsing context `properties` set on the Cavnas side, it doesn't break anything to leave the code as you have it.  So feel free to merge with or without the suggested change.

tschaub (approved):

ahocevar:
Thanks for the thorough review, @tschaub. Your suggested change sounds interesting. I think it makes sense as a follow-up change at a later time, and maybe it also makes sense to solve the problems a dot delimited properties path would impose before doing that.

santilland:
Great feature for flat styles, thanks for this!

tschaub:
> I think it makes sense as a follow-up change at a later time

I'm working on support for accessing nested properties in WebGL here: https://github.com/openlayers/openlayers/compare/main...tschaub:openlayers:nested (this reworks the property handling as [described above](https://github.com/openlayers/openlayers/pull/15813#discussion_r1612190109)).

PR #15815, opened 2024-05-09
Restrict ReprojDataTile triangulation to source projection extent

mike-000:
Potential fix for #15804


mike-000:
Replaced by #15825

PR #15818, opened 2024-05-10
Resolve NumberType | NumberArrayType ambiguity with a new SizeType

ahocevar:
This pull request removes the only real-life need for type hints in `"get"` call expressions, which was needed due to an ambiguity between `NumberType` and `NumberArrayType` for `-scale` style properties.

With the new `SizeType`, it would also be possible to support styles like `'icon-size': 24` instead of `'icon-size': [24, 24]`, but I did not make this change here.

### DIFF
--- a/src/ol/expr/gpu.js
+++ b/src/ol/expr/gpu.js
@@ -10,6 +10,7 @@ import {
   NumberArrayType,
   NumberType,
   Ops,
+  SizeType,
   StringType,
   computeGeometryType,
   isType,
@@ -19,6 +20,7 @@ import {
 } from './expression.js';
 import {Uniforms} from '../renderer/webgl/TileLayer.js';
 import {asArray} from '../color.js';
+import {toSize} from '../size.js';
 
 /**
  * @param {string} operator Operator
@@ -70,6 +72,16 @@ export function colorToGlsl(color) {
     (array[2] / 255) * alpha,
     alpha,
   ]);
+}
+
+/**
+ * Normalizes and converts a number or array toa `vec2` array compatible with GLSL.
+ * @param {number|import('../size.js').Size} size Size.
+ * @return {string} The color expressed in the `vec4(1.0, 1.0, 1.0, 1.0)` form.
+ */
+export function sizeToGlsl(size) {
+  const array = toSize(size);
+  return arrayToGlsl(array);
 }
 
 /** @type {Object<string, number>} */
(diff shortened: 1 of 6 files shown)


test/browser/spec/ol/webgl/styleparser.test.js
```diff
@@ -898,12 +898,12 @@ describe('ol.webgl.styleparser', () => {
           ],
           'circle-radius': 8,
           'circle-fill-color': ['get', 'color'],
-          'circle-scale': ['get', 'iconSize', 'number[]'],
+          'circle-scale': ['get', 'iconSize'],
```
ahocevar, line 901: This is the actual achievement - no type hint needed any more!

ahocevar:
This is now ready for review.

src/ol/expr/gpu.js
```diff
@@ -485,6 +497,12 @@ function compile(expression, returnType, context) {
     return arrayToGlsl(/** @type {Array<number>} */ (expression.value));
   }
 
+  if ((expression.type & SizeType) > 0) {
```
tschaub, line 500: Since anything that includes the `NumberType` is already [handled above](https://github.com/openlayers/openlayers/pull/15818/files#diff-152e8cab0a3f88df63530d1530a373f48e08c10bf47096ced43afd1047020223L466), isn’t this only testing that `NumberArrayType` is included?

I’m only reviewing on a small device, so probably missing context, but it isn’t clear to me why we aren’t checking for more narrow types in all these `if`s.

tschaub: Oh, I must be missing something.  I see `NumberArrayType` is also checked right above.

ahocevar: @tschaub Looks like you're not missing anything, but that code is missing tests. Looking into it now.

ahocevar: I think `SizeType` should be a distinct type, like `ColorType`, instead of a broadened type. But I haven't been able to fully wrap my head around the types involved in the parsing and compilation phases to figure out how that works. So @tschaub, if you have an idea how that could be achieved, any help would be appreciated.

ahocevar: Nevermind, I think I figured it out, @tschaub. Before I commit, I'm adding more tests.

ahocevar:
@tschaub Ready for review - tests added, `SizeType` is now a distinct type.

tschaub (approved):
Thanks for the work on this, @ahocevar.

PR #15826, opened 2024-05-14
Align getStyle() behavior and documentation

ahocevar:
Currently, the documentation for the BaseVector layer's `getStyle()` method says:

> Get the `style` for features. This returns whatever was passed to the `style` option at construction or to the `setStyle` method.

This is not correct:
1. When no style is configured that way, the default style is returned. This is even covered by an existing test for `setStyle()`.
2. When a flat style is configured that way, the style function created for that flat style is returned.

This pull request makes it so that 1. is properly documented, and that for 2., the passed flat style is returned by `getStyle()`, as documented.

### DIFF
--- a/src/ol/layer/BaseVector.js
+++ b/src/ol/layer/BaseVector.js
@@ -112,7 +112,7 @@ class BaseVectorLayer extends Layer {
 
     /**
      * User provided style.
-     * @type {import("../style/Style.js").StyleLike}
+     * @type {import("../style/Style.js").StyleLike|import("../style/flat.js").FlatStyleLike}
      * @private
      */
     this.style_ = null;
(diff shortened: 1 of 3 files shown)


santilland:
This is really helpful to be able to work more easily with flat styles, great! thanks!

tschaub (approved):

alecarn:
Should this type of change should be considered as breaking change and what is your versioning policy?

In our case, our apps and library broke because we allowed the minor version to be automatically updated.

ahocevar:
Sorry if this change broke your app. According to our semantic versioning policy, It was a bug fix (function did not do what the API documentation said), so a minor version increase. Were you using `getStyle()` instead of `getStyleFunction()` to get the style function?

alecarn:
Maybe the Typescript signature changes should be considered a breaking change?
Because this is not the only errors of type where our application failed. We were wondering if we should fix to the patch version of `ol` to ensure more stability and let's our maintainer catch this kind of changes up front.

PR #15827, opened 2024-05-14
Fix removal of render features from vector source

sebakerckhof:
Render features were not fully removed because they don't have change keys and can be arrays.
I look at the uindex now instead of the change keys

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -1116,6 +1116,10 @@ class VectorSource extends Source {
    */
   removeFeatureInternal(feature) {
     const featureKey = getUid(feature);
+    if (!this.uidIndex_.hasOwnProperty(featureKey)) {
+      return;
+    }
+
     if (featureKey in this.nullGeometryFeatures_) {
       delete this.nullGeometryFeatures_[featureKey];
     } else {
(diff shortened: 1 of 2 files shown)


src/ol/source/Vector.js
```diff
@@ -1116,22 +1116,36 @@ class VectorSource extends Source {
    */
   removeFeatureInternal(feature) {
     const featureKey = getUid(feature);
+    if (!this.uidIndex_.hasOwnProperty(featureKey)) {
```
ahocevar, line 1119: Elsewhere we do this like
```suggestion
    if (!(featureKey in this.uidIndex_)) {
```

src/ol/source/Vector.js
```diff
@@ -1116,22 +1116,36 @@ class VectorSource extends Source {
...
     } else {
       if (this.featuresRtree_) {
         this.featuresRtree_.remove(feature);
       }
     }
+
     const featureChangeKeys = this.featureChangeKeys_[featureKey];
-    if (!featureChangeKeys) {
-      return;
+    if (featureChangeKeys) {
```
ahocevar, line 1132: We usually keep the happy path aligned as far left as possible, so it would be preferable to keep the logic that returns early when there ar no `featureChangeKeys`.

ahocevar (changes requested):
Thanks, @sebakerckhof. I added a few suggestions.

src/ol/source/Vector.js
sebakerckhof: But we can't return early because renderfeatures don't have change keys.
I could check for `!featureChangeKeys && feature instanceof Feature` though if you prefer

ahocevar: Yes please, that would make the code easier to read.

sebakerckhof: Actually, an early return also doesn't make sense here since `!featureChangeKeys && feature instanceof Feature` can never happen anymore due to the introduction of the `uidIndex` check at the top.
So I just removed the if clause and added optional chaining on the `featureChangeKeys.foreach`.

ahocevar (approved):
Thanks, @sebakerckhof

PR #15831, opened 2024-05-16
Do not make the source type of a vector layer generic

### LINKED ISSUE #15830: Type VectorSource<Feature<Geometry>> is not assignable to type VectorSource<FeatureLike>
this ts issue not happen in ol 9.1.0 happen in 9.2.0

image

ahocevar:
Fixes #15830.

This pull request changes the TypeScript signature of `ol/layer/Vector`, `ol/layer/VectorImage` and `ol/layer/Heatmap` to accept the feature type instead of the source type as type parameter. This change makes sense because the previous signature was redundant, and did not work in some cases (see #15830).

Example: Previously, we had:
```ts
const layer: VectorLayer<VectorSource<Feature<Geometry>>> = new VectorLayer({
  source: new VectorSource({ // no source type other than VectorSource is accepted anyway
    features: [
      new Feature(new Point([0, 0])) // RenderFeature or Feature with different geometry type are also possible
    ]
  })
});
```
And now we have:
```ts
const layer: VectorLayer<Feature<Geometry>> = new VectorLayer({
  source: new VectorSource({ // no source type other than VectorSource is possible anyway
    features: [
      new Feature(new Point([0, 0])) // RenderFeature or Feature with different geometry type are also possible
    ]
  })
});
```

### DIFF
--- a/src/ol/layer/Graticule.js
+++ b/src/ol/layer/Graticule.js
@@ -178,7 +178,7 @@ const INTERVALS = [
  * Note that the view projection must define both extent and worldExtent.
  *
  * @fires import("../render/Event.js").RenderEvent
- * @extends {VectorLayer<import("../source/Vector.js").default>}
+ * @extends {VectorLayer<Feature>}
  * @api
  */
 class Graticule extends VectorLayer {
--- a/src/ol/layer/Heatmap.js
+++ b/src/ol/layer/Heatmap.js
@@ -8,6 +8,7 @@ import {clamp} from '../math.js';
 import {createCanvasContext2D} from '../dom.js';
 
 /**
+ * @template {import("../Feature.js").FeatureLike} FeatureType
  * @typedef {Object} Options
  * @property {string} [className='ol-layer'] A CSS class name to set to the layer element.
  * @property {number} [opacity=1] Opacity (0, 1).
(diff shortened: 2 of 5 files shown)


ahocevar:
Added another commit to also fix the [mentioned issue with Cluster sources](https://github.com/openlayers/openlayers/issues/15830#issuecomment-2115180041).

tschaub (approved):
Thanks, @ahocevar

MarcelCoding:
I think there are now some broken tnernal typings:
```
1. Type 'VectorSource<FeatureType>' does not satisfy the constraint 'VectorSource<FeatureLike> | VectorTile<FeatureLike>'.
     Type 'VectorSource<FeatureType>' is not assignable to type 'VectorSource<FeatureLike>'.
       Types of property 'setLoader' are incompatible.
         Type '(loader: FeatureLoader<FeatureType>) => void' is not assignable to type '(loader: FeatureLoader<FeatureLike>) => void'.
           Types of parameters 'loader' and 'loader' are incompatible.
             Types of parameters 'arg4' and 'arg4' are incompatible.
               Type '((arg0: FeatureType[]) => void) | undefined' is not assignable to type '((arg0: FeatureLike[]) => void) | undefined'.
                 Type '(arg0: FeatureType[]) => void' is not assignable to type '(arg0: FeatureLike[]) => void'.
                   Types of parameters 'arg0' and 'arg0' are incompatible.
                     Type 'FeatureLike[]' is not assignable to type 'FeatureType[]'.
                       Type 'FeatureLike' is not assignable to type 'FeatureType'.
                         'FeatureType' could be instantiated with an arbitrary type which could be unrelated to 'FeatureLike'. [2344]
```
                         
                         
image

ahocevar:
@MarcelCoding Have you tried v9.2.2?

MarcelCoding:
Yeah, sorry for not explicitly mentioned it, take a look at my node modules path. Also this worked before I've updated.

ahocevar:
@MarcelCoding I cannot reproduce. Can you share a reproduction repository, or at least your tsconfig.json?

PR #15834, opened 2024-05-16
Add test cases for typescript use

tschaub:
Inspired by #15830 and other similar issues, I think it makes sense to add test cases covering TypeScript use.


ahocevar (approved):
Good idea, @tschaub. Thanks!

Issue #15840, opened 2024-05-17
Modify interaction options and cluster source options optional


KyllianGautier:
Modify interaction options are required in the constructor, but all others interactions have optional options.
Same case for the cluster source.

**Describe the solution you'd like**
Make the Modify interaction options and the cluster source options optional.

mike-000:
Modify requires either `source` or `features` and throws an error "The modify interaction requires features, a source or a layer" if not supplied (the reference to layer seems to be outdated).  Options should not be optional for Draw, a valid `type` is required.  The docs for ClusterSource are incorrect as it also throws a runtime error if source is missing.

KyllianGautier:
@mike-000 Oh that's true, thanks, I fixed on my side.

mike-000:
I think the ClusterSource runtime error is a bug and the docs are correct as the error also occurs if the documented `setSource(null)` is used.  That means no options are mandatory and the `options` parameter can be optional.  I have created #15842 to fix that.

PR #15842, opened 2024-05-17
Avoid throwing error if ClusterSource source is null

mike-000:
Fixes the runtime error mentioned in https://github.com/openlayers/openlayers/issues/15840#issuecomment-2117715783

Since `source` can be null `options` can be optional.


tschaub (approved):
Thanks, @mike-000

PR #15851, opened 2024-05-21
Fix execute() arguments

### LINKED ISSUE #15849: VectorTileLayer with no declutter causes error when another VectorTileLayer has declutter enabled
**Describe the bug**
Having a `VectorTileLayer` with `declutter:false` and another `VectorTileLayer` with `declutter:true` causes the below console error and the map is blank. The error is not caused when all `VectorTileLayer`'s are `declutter:false`. 

The bug appears to have been introduced in `9.2.0` as it does not cause an error with `9.1.0`

```
helpers.js:122 Uncaught TypeError: declutterTree.collides is not a function
    at Executor.execute_ (Executor.js:960:32)
    at Executor.execute (Executor.js:1262:10)
    at ExecutorGroup.execute (ExecutorGroup.js:412:20)
    at CanvasVectorTileLayerRenderer.postRender (VectorTileLayer.js:765:23)
    at CanvasVectorTileLayerRenderer.renderFrame (TileLayer.js:536:10)
    at VectorTileLayer.render (Layer.js:384:28)
    at CompositeMapRenderer.renderFrame (Composite.js:142:29)
    at Map.renderFrame_ (Map.js:1595:20)
    at Map.animationDelay_ (Map.js:1449:10)
    at sentryWrapped (helpers.js:100:17)
```

**To Reproduce**
1. Go to this [sandbox](https://codesandbox.io/p/sandbox/mapbox-vector-tiles-forked-cwc5dr)
2. Add your Mapbox api key
3. Change one of the `VectorTileLayer`'s to `false`
4. See the error in the console.

If the sandbox link does not work, it is a fork of the [Mapbox Vector Tiles](https://openlayers.org/en/latest/examples/mapbox-vector-tiles.html) example, with the layer duplicated. 

**Expected behavior**
No error and the map to be rendered.

ahocevar:
Fixes #15849

### DIFF
--- a/src/ol/renderer/canvas/VectorTileLayer.js
+++ b/src/ol/renderer/canvas/VectorTileLayer.js
@@ -779,7 +779,7 @@ class CanvasVectorTileLayerRenderer extends CanvasTileLayerRenderer {
           rotation,
           hifi,
           replayTypes,
-          frameState.declutter === null ? undefined : frameState.declutter,
+          frameState.declutter?.[declutter],
         );
       }
       if (contextSaved) {
--- a/test/browser/spec/ol/renderer/canvas/vectortilelayer.test.js
+++ b/test/browser/spec/ol/renderer/canvas/vectortilelayer.test.js
@@ -806,4 +806,48 @@ describe('ol/renderer/canvas/VectorTileLayer', function () {
       });
     });
   });
+
+  describe('mixed declutter settings', () => {
+    let map;
+    beforeEach((done) => {
+      const extent = [
+        1824704.739223726, 6141868.096770482, 1827150.7241288517,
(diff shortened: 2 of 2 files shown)


tschaub (approved):

tschaub:
@ahocevar - I can cut another release with this fix.  Curious to hear your thoughts on including #15818.  Easy enough to cherry pick this change into a release branch if we think #15818 should have more release notes or if it is a breaking change.

ahocevar:
@tschaub I'd prefer to get #15813 in too, that pull request also contains the required upgrade notes. For the canvas renderer, the changes are definitely not breaking changes, for WebGL I don't know. So it would either be a minor or major version bump.

PR #15859, opened 2024-05-23
Use offset from start instead of full coordinate to improve getArea() precision

### LINKED ISSUE #15856: Polygon area calculation is inaccurate with large coordinate values
**Describe the bug**
When calling `polygon.getArea()` for a small polygon with large coordinate values, the result is inaccurate (floating point precision issue maybe?)

**To Reproduce**
1. Go to https://codesandbox.io/p/sandbox/measure-forked-q37j9x
2. The map is initially centered on (0,0) at maximum zoom level. Draw a small polygon. The area calculation works fine.
3. Zoom out and go somewhere far away like Greenland.
4. Zoom all they way in, start drawing a small polygon again.
5. While drawing, the calculated area keeps oscillating between a few discrete values like `0`, `-39.063`, `39.063`, `78.125` etc 

**Expected behavior**
The area calculation should return accurate results regardless of location.

**Workaround**
The issue can be fixed by first translating the polygon closer to origin and then calling `getArea()`. It would be good if `getArea()` did something like that automatically.

mike-000:
Fixes #15856

The current method is subject to rounding errors when subtracting the results of multiplying large numbers (especially in wrapped worlds).

### DIFF
--- a/src/ol/geom/flat/area.js
+++ b/src/ol/geom/flat/area.js
@@ -11,14 +11,16 @@
  */
 export function linearRing(flatCoordinates, offset, end, stride) {
   let twiceArea = 0;
-  let x1 = flatCoordinates[end - stride];
-  let y1 = flatCoordinates[end - stride + 1];
+  const x0 = flatCoordinates[end - stride];
+  const y0 = flatCoordinates[end - stride + 1];
+  let dx1 = 0;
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @mike-000

Issue #15579, opened 2024-02-22
GeoTiff reprojection from WGS84 to Mercator 


maxschmi:
**Describe the bug**

Openlayers reprojection of a GeoTiff TileLayer from WGS84 (EPS:4326) to Mercator (EPSG:3857) is not working as expected. 
The resulting grid has different cell widths. They somehow togle from wide to thin. So one column is wide and the next one is thin. 

See the screenshot:
<img src="https://github.com/openlayers/openlayers/assets/46047581/304fadb1-01dc-4f49-b6e0-edb162394e27" width="700px">

**To Reproduce**

Go to [codesandbox of minimal example](https://codesandbox.io/p/github/maxschmi/bug_tests_openlayer_reprojection/master?file=%2Findex.html)  or checkout the [minimal example repository](https://github.com/maxschmi/bug_tests_openlayer_reprojection)

**Expected behavior**

The grid cells should have the same width as seen in this screenshot from QGIS of the same layer and the same map projection:
<img src="https://github.com/openlayers/openlayers/assets/46047581/c70e602d-c899-41c8-a53e-3e4b83e5497c" width="700px">

mike-000:
You need to prevent excessive overzooming of the original tiles during reprojection by limiting the `maxZoom` of the reprojected tiles, for example

`tif_layer.getSource().setTileGridForProjection('EPSG:3857', createXYZ({maxZoom: 5}))`

works in your CodeSandbox.

You could attempt to calculate an appropriate setting based on `tif_view.resolutions`.

maxschmi:
Thanks for the quick response. At first the sugested solution looked good, but having a closer look I realized it interpolates between cells to get the pixel value.
This should absolutly not be the case, as I really want to show the grid cells value in my application.

I forked the minimal example and added @mike-000's sugested solution. 
Furthermore I added a hover to show the pixels value on mouse hover. The values in the test GeoTiffs are all integers, but as openalyer interpolates the values, they become float numbers.
See the [codeSandbox](https://codesandbox.io/p/github/maxschmi/bug_tests_openlayer_reprojection/fix_tileGrid_with_hover?file=%2Fmain.js) for the updated version.

Unfortunatly I don't really get how setTileGridForProjection and createXYZ work, so I couldn't fix this problem. I could only play around and change the values, but it never happened to work out as excpected.

>You could attempt to calculate an appropriate setting based on tif_view.resolutions.

How could I calculate this?

mike-000:
[...]

PR #15860, opened 2024-05-23
Reproject DataTiles using WebGL instead of 2d canvas.

sweepline:
I have made an implementation of the `ol/src/reproj:render()` function using webgl instead of 2d canvas. This solves three problems:

* Doing 4band reprojection in a single step instead of 3-bytes per pixel.
* Linear interpolation of data instead of only nearest neighbor.
* [An issue in firefox described here](https://github.com/openlayers/openlayers/pull/14099#issuecomment-2072301864
) means that a few pixels are mangled when using a 2d-canvas for reprojection.

Since this will be used for datatiles, it should be okay to do it in webgl, as datatiles are used for webgl based rendering anyway.

It does all the work for a datatile reprojection using a single `WebGLRenderingContext`. Whereas the current implementation computes an affine transformation from the source to the target explicitly, this is not needed when doing the work in WebGL as the transformation is built-in. 

This makes it possible to do up 4band f32/u8 reprojection in 1 pass <s>which means that you never need to do multiple passes as that is what a tiletexture can hold as well as far as I understand. It produces the same output in examples/numpytile for example which is supposedly 5-bands. I have not A/B tested the speed difference.</s>
(Some things like geotiff does 5-band reprojection, but it is working now) 

This means you can do interpolation when reprojection instead of always having to do nearest neighbour. It is even possible for f32 data if the webgl extension is available. Below is a comparison on some elevation-data. You can still see some artifacts because reprojected tiles have no gutter (also on the main branch), which I intend to fix in a future pull-request.
withartifacts
withoutartifacts

This fixes the same reprojection artifacts in Firefox as #14099, as well as the ones that could not be fixed in that PR.


tschaub:
Awesome!  Thank you for this contribution, @sweepline.  It has been on my list for a while to rework our raster reprojection.  I look forward to making time to dig into this, but just wanted to express my gratitude for your work.

src/ol/WebGLCanvas.js
```diff
@@ -0,0 +1,226 @@
...
+`;
+
+/** @typedef {import("./transform.js").Transform} Matrix */
+
+/**
+ * Canvas-like operations implemented in webgl.
+ */
+export class WebGLCanvas {
+  /**
+   * @param {WebGLRenderingContext} [gl] Context to render in.
```
tschaub, line 49:
```suggestion
   * @param {WebGLRenderingContext} gl Context to render in.
```

mike-000:
The `webgl-reproj-no-wrap` rendering test is passing at the `center` and `resolution` used.  However if you move to the pole and zoom in slightly an error occurs in your code:

image

This seems unrelated to the failing tests where `source.getTileSize(z)` does not match `source.getTileGrid().getTileSize(z)`.

sweepline:
Sorry for the late reply, some other things came onto my stack.

> The `webgl-reproj-no-wrap` rendering test is passing at the `center` and `resolution` used. However if you move to the pole and zoom in slightly an error occurs in your code:
> 
> image
> 

I cannot seem to reproduce this issue. I'm running `npm run test-rendering -- --interactive --match webgl-reproj-no-wrap` and panning around. Maybe a webgl context was dropped?

> This seems unrelated to the failing tests where `source.getTileSize(z)` does not match `source.getTileGrid().getTileSize(z)`.
> 

Which failing tests are that?

> Also https://deploy-preview-15860--ol-site.netlify.app/en/latest/examples/geotiff-reprojection.html and https://deploy-preview-15860--ol-site.netlify.app/en/latest/examples/multiple-cogs.html are not working for me (Windows 10 and Android) due to `Too many active WebGL contexts. Oldest context will be lost.` warnings. If I restrict the first example to 4 bands including added alpha by specifying `bands: [1, 2, 3],` it does work.

The active webgl context error is me not returning early enough when there is more than 4 bands, I have fixed that now. I guess the code should still be made robust with regards to webgl contexts though.

Then the actual issue though, is that I have not written support for more than 4 bands as i thought it was not really used and other webgl code in openlayers also does not support 4 bands e.g. `src/ol/webgl/TileTexture.js`. I had not realised all GeoTIFF's basically use 5 bands because of the "nodata-band". 

I will work on some code to support multiple bands, I guess I can do what the old code did with multiple reprojections.

mike-000:
> Which failing tests are that?

Failing rendering tests:
image

The first test uses a COG with non-square pixels, as does the second (the output should, but does not, look similar so that probably also loses context).

The third is a different issue as the output is not clipped as expected at the top or left.

Zipped actual outputs be been downloaded via https://github.com/openlayers/openlayers/actions/runs/9209372373/artifacts/1531447766

mike-000:
> I cannot seem to reproduce this issue. I'm running `npm run test-rendering -- --interactive --match webgl-reproj-no-wrap` and panning around. Maybe a webgl context was dropped?

I can reproduce reliably with

    center: fromLonLat([0, 90], 'stereo-sib'),
    resolution: 7000,

throwing `WebGL: INVALID_VALUE: texImage2D: width or height out of range` at line 244 in glreproj.js

sweepline:
> The first test uses a COG with non-square pixels, as does the second (the output should, but does not, look similar so that probably also loses context).

Thanks, I can see the flaw having something to with pixel ratios, looking to fix that this week.
 
> The third is a different issue as the output is not clipped as expected at the top or left.

Thanks, I fixed that now.

> I can reproduce reliably with
> 
> ```
> center: fromLonLat([0, 90], 'stereo-sib'),
> resolution: 7000,
> ```
> 
> throwing `WebGL: INVALID_VALUE: texImage2D: width or height out of range` at line 244 in glreproj.js

image
 I am just seeing this.
 
 ---
 
 I also fixed the examples now in the newest commits (I squashed sorry). It was indeed just supporting multiple bands.
 
 So i need to fix the last two tests and it should be all.

mike-000:
> I am just seeing this.

I have tried on more devices and get different results depending on the device.

mike-000:
That is probably an improvement as it was necessary to stretch in canvas without interpolation to avoid corrupting floats, see #15885.  Maybe add a new test **webgl-reproj-float-interpolate-false** which should produce something similar to the old results.  Might this also improve `interpolate: false` situations such as #15579?

mike-000:
> > I am just seeing this.
> 
> I have tried on more devices and get different results depending on the device.

In a polar projection source tiles from the entire width of the world may need to be included in the `stitchWidth` to produce a reproj tile covering the pole.  Although I initially noticed the problem on an old Windows PC where MAX_TEXTURE_SIZE was 8196, it is worst on a relatively new Android 13 device where MAX_TEXTURE_SIZE is only 4096.  If these stats https://web3dsurvey.com/webgl/parameters/MAX_TEXTURE_SIZE are reliable most Android devices and 24% of all devices would have this issue.  Would it be practical in your solution to downsize the `stitchTexture` if either dimension exceeds `gl.getParameter(gl.MAX_TEXTURE_SIZE)`?

sweepline:
> Would it be practical in your solution to downsize the `stitchTexture` if either dimension exceeds `gl.getParameter(gl.MAX_TEXTURE_SIZE)`?

Yeah I think, so thanks for digging into this.

sweepline:
@mike-000 I have tested this by artificially reducing the max texture size in js, could you see if this works on your devices? I don't really have access to anything with low max texture resolution.

sweepline:
The issue concerning max texture size with stitch images should be fixed now.

mike-000:
> The issue concerning max texture size with stitch images should be fixed now.

Yes, some artefacts are visible at times because of the reduced quality, but that is better than black tiles:

image

sweepline:
I was wondering if there is anything else I can to do in order to move the review process along?

tschaub (approved):

tschaub:
Thank you for your work on this, @sweepline. I added a single commit that moved the `ol/WebGLCanvas.js` module to `ol/webgl/Canvas.js` so that it would be adjacent to other similar WebGL helper code (and ideally encourage some reuse).

I am seeing black/blank spots rendered in the geotiff-reprojection.html example with the changes in this branch. But I am also seeing the same in the default branch, so it feels like there are existing issues that could be improved.

reproj

For future work work on reprojection, we want to move away from the "reproj tiles" altogether. Ideally, the renderer would be working with regular tiles and would handle the warping itself.

mike-000:
@tschaub The console suggests that might be a network problem with that source file.  I do not see any problems with Multiple COG Sources or STAC Support example which also reproject.  However after rebasing #16227 it seems that there is a regression in either #15825 or #15484 which can be seen in the example:

image

compare
https://deploy-preview-16227--ol-site.netlify.app/en/latest/examples/cog-modeltransformation.html
with
https://deploy-preview-15988--ol-site.netlify.app/en/latest/examples/cog-modeltransformation.html

PR #15869, opened 2024-05-25
Consistent generics for vector(tile) layers, sources, formats

### LINKED ISSUE #15858: Can't run/compile angular/typescript application with OL > 9.0.0
**Describe the bug**
I've just upgraded to OL 9.2.3 from OL 8.1.0 and now OL can't compile due to what I think is internal import errors.
Previously I've been using OL together with package @types/ol for typedefinitions, but now it seems like OL has it's own.

After installing OL >9 i get errors like this  when trying to run/compile my application:

Multiple export errors like this for GeoJSON:
> Error: node_modules/ol/format/GeoJSON.d.ts:2:36 - error TS2307: Cannot find module 'geojson' or its corresponding type declarations.
> 2 export type GeoJSONObject = import("geojson").GeoJSON;

Errors like this:

> Error: node_modules/ol/layer/Heatmap.d.ts:89:99 - error TS2344: Type 'VectorSource<FeatureType>' does not satisfy the constraint 'VectorSource<FeatureLike> | VectorTile<FeatureLike>'.
>   Type 'VectorSource<FeatureType>' is not assignable to type 'VectorSource<FeatureLike>'.
>     Types of property 'setLoader' are incompatible.
>       Type '(loader: FeatureLoader<FeatureType>) => void' is not assignable to type '(loader: FeatureLoader<FeatureLike>) => void'.
>         Types of parameters 'loader' and 'loader' are incompatible.
>           Types of parameters 'arg4' and 'arg4' are incompatible.
>             Type '((arg0: FeatureType[]) => void) | undefined' is not assignable to type '((arg0: FeatureLike[]) => void) | undefined'.
>               Type '(arg0: FeatureType[]) => void' is not assignable to type '(arg0: FeatureLike[]) => void'.
>                 Types of parameters 'arg0' and 'arg0' are incompatible.
>                   Type 'FeatureLike[]' is not assignable to type 'FeatureType[]'.
>                     Type 'FeatureLike' is not assignable to type 'FeatureType'.
>                       'FeatureLike' is assignable to the constraint of type 'FeatureType', but 'FeatureType' could be instantiated with a different subtype of constraint 'FeatureLike'.
>                         Type 'RenderFeature' is not assignable to type 'FeatureType'.
>                           'RenderFeature' is assignable to the constraint of type 'FeatureType', but 'FeatureType' could be instantiated with a different subtype of constraint 'FeatureLike'.
> 
> 89 declare class Heatmap<FeatureType extends import("../Feature.js").FeatureLike> extends BaseVector<import("../source/Vector.js").default<FeatureType>, WebGLPointsLayerRenderer> {

> Error: node_modules/ol/layer/VectorImage.d.ts:165:35 - error TS2344: Type 'VectorSource<FeatureType>' does not satisfy the constraint 'VectorSource<Feature<Geometry>>'.
>   Types of property 'setLoader' are incompatible.
>     Type '(loader: FeatureLoader<FeatureType>) => void' is not assignable to type '(loader: FeatureLoader<Feature<Geometry>>) => void'.
>       Types of parameters 'loader' and 'loader' are incompatible.
>         Types of parameters 'arg4' and 'arg4' are incompatible.
>           Type '((arg0: FeatureType[]) => void) | undefined' is not assignable to type '((arg0: Feature<Geometry>[]) => void) | undefined'.
>             Type '(arg0: FeatureType[]) => void' is not assignable to type '(arg0: Feature<Geometry>[]) => void'.
>               Types of parameters 'arg0' and 'arg0' are incompatible.
>                 Type 'Feature<Geometry>[]' is not assignable to type 'FeatureType[]'.
>                   Type 'Feature<Geometry>' is not assignable to type 'FeatureType'.
>                     'Feature<Geometry>' is assignable to the constraint of type 'FeatureType', but 'FeatureType' could be instantiated with a different subtype of constraint 'Feature<Geometry>'.
> 
> 165     constructor(options?: Options<import("../source/Vector.js").default<FeatureType>> | undefined);

I'm maintaining a very large OpenLayers based application as well as a inner-source library based on OpenLayers, and I need som information that can point me in the right direction to fix this.

**To Reproduce**
Upgrade application from OL < 9 to OL > 9

**Expected behavior**
The application should compile/run

ahocevar:
Fixes #15858

In addition to fixing several type issues, this pull request sets defaults for all type parameters to layers, sources and formats - this way no generics need to be specified any more when the defaults are used.

### DIFF
--- a/src/ol/format/Feature.js
+++ b/src/ol/format/Feature.js
@@ -95,7 +95,7 @@ import {
  */
 
 /***
- * @template {Feature|RenderFeature} T
+ * @template {import('../Feature.js').FeatureLike} T
  * @typedef {T extends RenderFeature ? typeof RenderFeature : typeof Feature} FeatureToFeatureClass
  */
 
@@ -113,7 +113,7 @@ import {
  * {@link module:ol/Feature~Feature} objects from a variety of commonly used geospatial
  * file formats.  See the documentation for each format for more details.
  *
- * @template {import('../Feature.js').FeatureClass} [T=typeof import('../Feature.js').default]
+ * @template {import('../Feature.js').FeatureClass} [FeatureClassType=import('./Feature.js').FeatureToFeatureClass<import("../Feature.js").default>]
  * @abstract
  * @api
  */
@@ -133,9 +133,9 @@ class FeatureFormat {
 
     /**
      * @protected
-     * @type {T}
+     * @type {FeatureClassType}
      */
-    this.featureClass = /** @type {T} */ (Feature);
+    this.featureClass = /** @type {FeatureClassType} */ (Feature);
 
     /**
      * A list media types supported by the format in descending order of preference.
@@ -206,7 +206,7 @@ class FeatureFormat {
    * @abstract
    * @param {Document|Element|Object|string} source Source.
    * @param {ReadOptions} [options] Read options.
-   * @return {import("../Feature.js").FeatureLike|Array<import("../render/Feature.js").default>} Feature.
+   * @return {FeatureClassToFeature<FeatureClassType>|Array<FeatureClassToFeature<FeatureClassType>>} Feature.
    */
   readFeature(source, options) {
     return abstract();
@@ -218,7 +218,7 @@ class FeatureFormat {
    * @abstract
    * @param {Document|Element|ArrayBuffer|Object|string} source Source.
    * @param {ReadOptions} [options] Read options.
-   * @return {Array<import('../Feature.js').FeatureLike|FeatureClassToFeature<T>>} Features.
+   * @return {Array<FeatureClassToFeature<FeatureClassType>>} Features.
    */
   readFeatures(source, options) {
     return abstract();
--- a/test/typescript/cases/cluster-source.ts
+++ b/test/typescript/cases/cluster-source.ts
@@ -5,7 +5,19 @@ import Point from '../../../src/ol/geom/Point.js';
 import VectorLayer from '../../../src/ol/layer/Vector.js';
 import VectorSource from '../../../src/ol/source/Vector.js';
 
-export const layer: VectorLayer<Feature<Geometry>> = new VectorLayer({
+export const layer1: VectorLayer<Feature<Geometry>> = new VectorLayer({
+  source: new ClusterSource({
+    distance: 10,
+    source: new VectorSource({
+      features: [new Feature(new Point([0, 0]))],
+    }),
+  }),
+});
+
+export const layer2: VectorLayer<
+  Feature,
+  ClusterSource<Feature<Point>>
+> = new VectorLayer({
   source: new ClusterSource({
     distance: 10,
     source: new VectorSource({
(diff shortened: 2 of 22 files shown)


ahocevar:
Marked ready for review. I think this is as good as it gets. I experimented with extracting the FeatureType from the SourceType‘s FestureType type argument, but could not get it to work.

RobertOrthofer:
This seems to have fixed the issues I had with `VectorLayers` and `VectorSources`.

ahocevar:
Added a test to make sure we continue to pass `tsconfig`s with `skipLibCheck: false` in the future.

Thanks for any review.

src/ol/format/Feature.js
```diff
@@ -113,7 +113,7 @@ import {
  * {@link module:ol/Feature~Feature} objects from a variety of commonly used geospatial
  * file formats.  See the documentation for each format for more details.
  *
- * @template {import('../Feature.js').FeatureClass} [T=typeof import('../Feature.js').default]
+ * @template {import('../Feature.js').FeatureClass} [FeatureClassType=import('./Feature.js').FeatureToFeatureClass<import("../Feature.js").default>]
```
tschaub, line 116: Maybe you already tried this and it was a dead end, but it seems like it would be simpler if the template here were just this.

```
/**
 * @template {import('../Feature.js').default} T
 */
```

Or maybe it needs to be optional, not sure.

Then below, we could have this:
```js
    /**
     * @protected
     * @type {import('../Feature.js').FeatureClass}
     */
    this.featureClass = Feature;
```

I'm wondering if this would keep us from having to use `import('./Feature.js').FeatureToFeatureClass<import("../Feature.js").default>` in so many other places.

test/typescript/cases/cluster-source.ts
```diff
@@ -5,7 +5,19 @@ import Point from '../../../src/ol/geom/Point.js';
...
+    source: new VectorSource({
+      features: [new Feature(new Point([0, 0]))],
+    }),
+  }),
+});
+
+export const layer2: VectorLayer<
+  Feature,
+  ClusterSource<Feature<Point>>
+> = new VectorLayer({
```
tschaub, line 20: I imagine TS people would expect to be able to use something like
```js
const layer: VectorLayer<ClusterSource<Feature<Point>>> = // ...
```

Meaning the `Feature` type parameter would not need to be repeated.

There is a [stack overflow post](https://stackoverflow.com/questions/49102449/nested-generics-in-function-definition-in-typescript) that makes it look like `infer` might be a way to have a single `@template` in the layer constructor representing the source type and to extract the named feature type.  I toyed around with this but could not get the right syntax (or make sense of the docs really).

I'm not sure if this is a limitation of the JSDoc support, but it seems like TypeScript would allow.

ahocevar: That's what I tried, but failed (see https://github.com/openlayers/openlayers/pull/15869#issuecomment-2132397184)

src/ol/format/Feature.js
ahocevar: If we did that, we'd lose the ability to infer the return type of `readFeatures()` etc. to infer the feature type from the configured `featureClass`.

PR #15876, opened 2024-05-27
Fix case expression with numeric size types

### LINKED ISSUE #15868: Previously working WebGLPointsLayer styles crash in 9.2.3
**Describe the bug**  
We are using WebGLPointsLayer with 'flat styles' (dont quite work the same way for webGLPointslayer as far as i can tell).  
One of our layer styles sets the scale of the shape rendered by looking up if one of a number of attributes is set on a feature.

```
'shape-scale': [
	'case',
	['==', ['get', 'A'], 'true'], 1,
	['==', ['get', 'B'], 'true'], 2,
	3
],
                
```

This works fine up to ol 9.2.2

Once you use 9.2.3, it crashes giving this error:

```
Error: Could not find a common output type for the following case operation: ["case",["==",["get","A"],"true"],1,["==",["get","B"],"true"],2,3]
```

I have looked at the changes in 9.2.3 but i don't have enough insight into ol's inner working to understand why they would have any influence on this.

**To Reproduce**  

I made i minimal reproducable example: https://codepen.io/Innok-Louis/pen/qBGaOpN

1.  open example
2.  open dev tools to see error
3.  change version to 9.2.2 to see the error disappear

**Expected behavior**  
Some insight on why this happens.   
For now we moved our dependency from ^9.0.0 to 9.2.2 to prevent any more production systems grabbing the version that crashes.  
Happy to provide any more info to help find the root cause.

M393:
Fixes #15868

### DIFF
--- a/src/ol/expr/expression.js
+++ b/src/ol/expr/expression.js
@@ -951,7 +951,7 @@ function parseInterpolateArgs(encoded, context, parsedArgs, typeHint) {
  * @type ArgValidator
  */
 function parseCaseArgs(encoded, context, parsedArgs, typeHint) {
-  const fallback = parse(encoded[encoded.length - 1], context);
+  const fallback = parse(encoded[encoded.length - 1], context, typeHint);
   let outputType =
     typeHint !== undefined ? typeHint & fallback.type : fallback.type;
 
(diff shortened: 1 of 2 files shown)


tschaub (approved):

tschaub:
Thanks, @M393!

PR #15880, opened 2024-05-28
VectorSource improvements

M393:
This simplifies the code and (presumably) improves the performance when removing features or when a feature's id was changed.

### DIFF
--- a/src/ol/source/Vector.js
+++ b/src/ol/source/Vector.js
@@ -1103,7 +1103,7 @@ class VectorSource extends Source {
   /**
    * Remove feature without firing a `change` event.
    * @param {FeatureType} feature Feature.
-   * @return {boolean} True if the feature was removed, false if it was not added.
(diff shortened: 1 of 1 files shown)


src/ol/source/Vector.js
```diff
@@ -1098,23 +1094,22 @@ class VectorSource extends Source {
...
       this.changed();
     }
   }
 
   /**
    * Remove feature without firing a `change` event.
    * @param {FeatureType} feature Feature.
-   * @return {FeatureType|undefined} The removed feature
-   *     (or undefined if the feature was not found).
+   * @return {boolean} True if the feature was removed, false if it was not added.
```
ahocevar, line 1106:
```suggestion
   * @return {boolean} True if the feature was removed, false if it was not found.
```

ahocevar (approved):
Thanks, @M393

PR #15881, opened 2024-05-28
fix: avoid useless array allocation in removeFeatures

sebakerckhof:
A small perf improvement when removing large amounts of features.


sebakerckhof:
Seems like @M393 had the same idea at the same time

PR #15891, opened 2024-06-03
Sentinel Hub source

tschaub:
This adds a source that uses the Sentinel Hub [Processing API](https://docs.sentinel-hub.com/api/latest/api/process/) to render tiles.  Use of the API requires an OAuth client id and secret (or [an access token](https://docs.sentinel-hub.com/api/latest/api/overview/authentication/) retrieved by other means).  In case people don't have an account, here is what [one of the examples](https://deploy-preview-15891--ol-site.netlify.app/en/latest/examples/sentinel-hub-custom-script) looks like.

image

### DIFF
--- a/config/tsconfig-strict.json
+++ b/config/tsconfig-strict.json
@@ -231,6 +231,7 @@
     "../src/ol/source/OGCVectorTile.js",
     "../src/ol/source/OSM.js",
     "../src/ol/source/Raster.js",
+    "../src/ol/source/SentinelHub.js",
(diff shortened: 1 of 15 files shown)


bojanbizjak:
Is it possible to use SentinelHub source right now e.g. by pointing to the specific version of ol in package.json.

Thank you.

ahocevar (approved):
Thanks for this great addition to the library, @tschaub. Sad though that the new examples aren't really fun without having a Sentinel Hub account.

PR #15905, opened 2024-06-10
New image tile source

ahocevar:
Duplicate of #15568, to make it work with the changes in main since the creation of that pull request.

Still missing:

- [x] Stale key handling (make wmts-dimensions.html example work properly)


ahocevar:
I think this is in good shape now. Along with the changes comes at least one improvement: tile transitions now also work on layers with an opacity set.

@tschaub, since this is based on #15568, do you want to review my additional changes here? Your changes from #15568, which are fully included here, look good to me.

src/ol/renderer/canvas/TileLayer.js
```diff
@@ -74,62 +168,109 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
...
+     * @type {number}
+     */
+    this.renderedSourceRevision_;
+
     /**
      * @private
      * @type {boolean}
      */
-    this.newTiles_ = false;
+    //this.newTiles_ = false;
```
tschaub, line 187: Should this be removed or uncommented?

ahocevar: Since the rest of the code `this.newTiles_` is commented out, I left it like this. But all that could be removed - it's no longer needed.

tschaub (approved):

tschaub:
Thanks for all the work resurrecting this branch, dealing with all the conflicts, and making additional improvements, @ahocevar.  This was a huge effort.

ahocevar:
Thanks for your hard work and the good collaboration on this, @tschaub!

PR #15906, opened 2024-06-11
Provide expected type when parsing expressions

tschaub:
This makes it so the `parse` function for expressions requires and consistently makes use of the expected type.


src/ol/color.js
```diff
@@ -102,11 +102,11 @@ export function fromString(s) {
 
   const color = parseRgba(s);
   if (color.length !== 4) {
-    throw new Error('Failed to parse "' + s + '" as color');
+    throw new Error('failed to parse "' + s + '" as color');
```
tschaub, line 105: I changed these to lowercase so that they read more naturally when wrapping the messages with more detail (see below).  I could change them back if we want.

src/ol/expr/cpu.js
```diff
@@ -92,14 +85,7 @@ export function newEvaluationContext() {
...
 export function buildExpression(encoded, type, context) {
-  const expression = parse(encoded, context);
-  if (!overlapsType(type, expression.type)) {
-    const expected = typeName(type);
-    const actual = typeName(expression.type);
-    throw new Error(
-      `Expected expression to be of type ${expected}, got ${actual}`,
-    );
-  }
+  const expression = parse(encoded, type, context);
```
tschaub, line 88: Parsing used to try to infer types from literal values and then assert that the expected type overlaps with the inferred type.  The new strategy is to pass down the expected type from the top - casting or coercing values to the expected type where appropriate.

src/ol/expr/expression.js
```diff
@@ -256,65 +268,101 @@ export function newParsingContext() {
...
+      if (isType(expectedType, BooleanType)) {
+        return new LiteralExpression(BooleanType, !!encoded);
+      }
+      if (isType(expectedType, StringType)) {
+        return new LiteralExpression(StringType, encoded.toString());
+      }
+      if (!includesType(expectedType, NumberType)) {
+        throw new Error(`got a number, but expected ${typeName(expectedType)}`);
+      }
+      return new LiteralExpression(NumberType, encoded);
```
tschaub, line 301: When parsing literals in expressions, we cast values when appropriate to the expected type.

ahocevar (approved):
Very impressive work. Thanks, @tschaub

PR #15916, opened 2024-06-15
Correct grammatically incorrect use of "an" in documentation

mike-000:
Previously `an ol/Feature` (for example) in the documentation would have been grammatically correct, but it now appears as `an Feature` with is incorrect unless an is followed by a vowel.


tschaub (approved):
Thanks, @mike-000

Issue #15910, opened 2024-06-13
webgl memory leak


752841728:
在官网的[例子](https://openlayers.org/en/latest/examples/webgl-draw-line.html)中，我似乎发现了内存泄漏，每次平移或者缩放时，内存就会增加，特别是数据量较大的时候，浏览器会因为内存的增加而变得卡顿

下面是我测试时候，内存的占用情况
![image](https://github.com/openlayers/openlayers/assets/59192839/cefd9c2d-3e4c-4588-8172-a039e4fcf335)

jahow:
Could you please test again with the main branch and report here whether the issue still happens? Thank you!

leizhangc:
Hi @jahow .  about VectorLayer.js——renderer.generateBuffers(this.batch_, transform).
       If submit a big batch of features to render, webworker of webgl always report "RangeError: Invalid array length".  
       If cut the big batch to some small batch, then submit the small batches, webgl can show much more features, but the features can't be selected rightly.  Click feature A, but return Feature B. 
       How to cut the big batch into many small batch to render and click, aim to load much more features.  thank you.

leizhangc:
Hi @jahow .  about VectorLayer.js——renderer.generateBuffers(this.batch_, transform).
       If submit a big batch of features to render, webworker of webgl always report "RangeError: Invalid array length".  
       If cut the big batch to some small batch, then submit the small batches, webgl can show much more features, but the features can't be selected rightly.  Click feature A, but return Feature B. 
       How to cut the big batch into many small batch to render and click, aim to load much more features.  thank you.

Issue #15850, opened 2024-05-21
openlayers performance


nipeone:
Using openlayers and geojs to display **90,000** features, when zooming or panning, the openlayers will lag for 1~2 seconds, but geojs works smoothly. Is the usage of openlayers not correct? Here is my reference code:https://github.com/nipeone/geojs_vs_openlayers

behoney:
Performance might be improved by setting [`renderBuffer`](https://openlayers.org/en/latest/apidoc/module-ol_layer_Vector-VectorLayer.html) to recommended value(the size of the largest symbol, line width or label; default is 100)

ahocevar:
Another option is to use an ImageVector layer instead of a Vector layer.

nipeone:
> Another option is to use an ImageVector layer instead of a Vector layer.

Thanks for your reply, do you mean the class ol.layer.VectorImage? I am using this class and it does perform better than ol.layer.Vector. But still not as good as geojs performance.

ahocevar:
@nipeone I don't think the `renderBuffer` advice applies here.

ahocevar:
@nipeone If you want to use the WebGL renderer (which might improve performance), you need to write different code. See e.g. https://openlayers.org/en/latest/examples/webgl-vector-layer.html. Looks like what you have now comes from an OpenLayers v3 example.

nipeone:
that's

> @nipeone If you want to use the WebGL renderer (which might improve performance), you need to write different code. See e.g. https://openlayers.org/en/latest/examples/webgl-vector-layer.html. Looks like what you have now comes from an OpenLayers v3 example.

@ahocevar Thanks again, performance improves dramatically with the above example, but if increase the number of features to 360000, an error will be reported like this: `RangeError: Invalid array length`

jia-mx:
> @nipeone If you want to use the WebGL renderer (which might improve performance), you need to write different code. See e.g. https://openlayers.org/en/latest/examples/webgl-vector-layer.html. Looks like what you have now comes from an OpenLayers v3 example.

When using WebGLLayer to render massive vector features, zooming or moving the map causes the memory to increase sharply until the web page crashes. In which version can this problem be solved?

nipeone:
@jia-mx you can try my example code, i did not test max limitation, when len(features)=90000, everything is ok.if len(features)=360000 will report error

jia-mx:
[...]

Issue #15132, opened 2023-09-14
WebGL vector layer example crashes iOS Safari


tschaub:
**Describe the bug**
It looks like the WebGL vector layer crashes mobile Safari after a short period of use.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/webgl-vector-layer.html
2. Land and zoom around for a while
3. See “a problem repeatedly occurred…”

![image](https://github.com/openlayers/openlayers/assets/41094/3d110d72-361d-4380-84d1-4eb3697786bd)

**Expected behavior**
Ideally, the example would not crash the mobile browser.

jahow:
Reproduced on my Android phone as well, the example gets slower and slower while interacting and then crashes

jahow:
After investigating, it turns out that the webgl rendering logic is really intensive on garbage collection for the lines and polygons. Array buffers are also not reused between two rebuilding of the webgl buffers, so panning/zooming around will generate _a lot_ of garbage to collect.

See the following places when arrays/typed arrays are unnecessarily allocated:
https://github.com/openlayers/openlayers/blob/10a0f23de9c90ad05182a9e9bc672e79c2b18597/src/ol/worker/webgl.js#L136-L137
https://github.com/openlayers/openlayers/blob/10a0f23de9c90ad05182a9e9bc672e79c2b18597/src/ol/render/webgl/renderinstructions.js#L72-L77

jeroenvheel:
We also have a problem with an iPad with the 5th generation where the website craches
We are using OL 8.1.0

garthhenning:
I can reproduce this error regularly with a textured polygon in OpenLayers. As of iOS15, Apple reportedly limited the WebGL texture size to only 4096x4096. If a polygon is textured, the webmap will slow as the polygon gets larger, then the iOS Safari browser tab will crash when the polygon hits 4096 pixels wide or high. The tab gives the "A problem repeated occurred..." error every time. Since an iPhone screen is usually around 400 pixels wide, that means a polygon zoomed to be 10x the width of the iPhone screen will cause the crash. Ideally, iOS would give a user-friendly error, but instead it crashes the browser tab quite thoroughly. This bug took forever to find because my individual texture image for the polygon is only 256x256, but the bug is triggered when the resulting textured polygon rendered by OpenLayers hits 4096x4096.

jahow:
[...]

PR #15943, opened 2024-06-25
WebGL Vector Layer: Clear existing buffers before allocating new ones

jahow:
This should fix a longstanding memory leak reported with the WebGL Vector Layer Renderer.

Reported in #15910 #15850 #15132 

With this change I don't see the WebGL Vector Layer example crash on iOS, and the memory footprint looks stable locally. Could someone please try out this branch and confirm that this indeed fixes it? Thanks in advance.


jahow:
On now I was just bit by the flaky tests... will work on this next (as well as the flashing bug in the webgl vector example)

tschaub (approved):
Thanks, @jahow

PR #15946, opened 2024-06-26
Fix polyfills link

sanak:
Hi maintainers,

We noticed the following `polyfill.io` security news today.
https://sansec.io/research/polyfill-supply-chain-attack

And we noticed that there are 2 links in this repository's documents.

This PR replaces the link to safer ones ([Fastly](https://polyfill-fastly.io) or [Cloudflare](https://cdnjs.cloudflare.com/polyfill)), so reviewing this is helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ The [ol package](https://npmjs.com/package/ol) includes auto-generated TypeScrip
 
 ## Supported Browsers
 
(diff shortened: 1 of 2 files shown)


mike-000:
I suspect few people will need polyfills with version 7 or later as it also uses modern browser syntax which cannot be polyfilled.

A more immediate risk may be the examples accessible via the OLDER VERSIONS examples links for version 6.15.1 and earlier on https://openlayers.org/ which embed links to scripts on the polyfill.io domain.

tschaub (approved):

PR #15962, opened 2024-07-02
Handle rotated, skewed or flipped GeoTIFF tile grids

### LINKED ISSUE #15294: GeoTIFF layers do not support TIFFs with GeoTransform/rotation

### LINKED ISSUE #15370: GeoTIFF rotation

### LINKED ISSUE #15428: error if GeoTiffs origin is set to the lower left corner

mike-000:
Fixes #15294
Fixes #15370
Fixes #15428

Replaces #15402

An update to `getTileCoordForTileUrlFunction` was needed to use the correct grid and fix the issue in https://github.com/openlayers/openlayers/pull/15402#issuecomment-2197354410.  I made this specific to `ol/source/DataTile` but it might be possible to apply the change to `ol/source/Tile` if other tile sources are compatible.


ahocevar (changes requested):

mike-000:
Browser tests are failing because the source does not have a tile grid, which is now required if a view projection is not specified in `getTileCoordForTileUrlFunction`.  In real conditions is it ever possible for a tile source to not have a tile grid?

ahocevar:
If a tile source does not have a tile grid, `ol/tilegrid#getForProjection()` should be used. This can be seen in the `getTileGridForProjection(projection)` methods of the respective TileSource classes.

mike-000:
> If a tile source does not have a tile grid, `ol/tilegrid#getForProjection()` should be used. This can be seen in the `getTileGridForProjection(projection)` methods of the respective TileSource classes.

While a GeoTIFF source should always have a tile grid I can see it would be possible for a DataTile source to not have one, and as a `transformMatrix` will cause reprojection I have also updated `getReprojTile_` to allow for that.  There is currently no API method to set a `transformMatrix` but it might be useful for debug tiles.

mike-000:
Also updated rendering test to show ModelTransformation can work in COG without a projection, which would not have been possible in #15402.

PR #15964, opened 2024-07-05
Handle a KML icon scale of zero correctly

pakb:
let the name go through without calculating any anchor or offset, it otherwise leads to a [NaN, NaN] offset and the name isn't displayed.

I have a couple user's KML that have `<Placemark>` with an `<IconStyle>` with only `<scale>0</scale>` in them. (They only want to show a label, without the default yellow pin coming from GoogleEarth). The label style is ignored/mishandled because an offset is calculated on the scale of zero, leading to a [NaN, NaN] result.
I'm trying to fix this issue.

Here's the rendering of the KML on Google Earth
image

Here on current OL
image

Here with the fix
image

### DIFF
--- a/src/ol/format/KML.js
+++ b/src/ol/format/KML.js
@@ -1027,8 +1027,13 @@ function createNameStyleFunction(foundStyle, name) {
   const imageStyle = foundStyle.getImage();
   if (imageStyle) {
     const imageSize = imageStyle.getSize();
-    if (imageSize && imageSize.length == 2) {
-      const imageScale = imageStyle.getScaleArray();
+    const imageScale = imageStyle.getScaleArray();
+    if (
+      imageSize &&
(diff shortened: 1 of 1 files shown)


ahocevar (commented):
Are you sure this is the correct fix? You're checking for `imageScale` being truthy (which is redundant, because it will always be an `Array`). Then you're checking if either dimension of `imageScale` is greater than zero. But in the `textOffset` calculation below, a zero `imageScale` will work fine, because the result will be `0`. Where did you get a `[NaN, NaN]` textOffset?

To see what's going on, I think this pull request would benefit from a test.

ahocevar:
@pakb I'm closing this. If you get a chance to look into why `imageStyle.getAnchor()` returns `[NaN, NaN]` in this context, it would be appreciated.

PR #15979, opened 2024-07-09
Better generics for rbush

MoonE:


ahocevar (approved):
Thanks @MoonE for this type improvement!

PR #15983, opened 2024-07-11
Cleanup code and fix type declaration

MoonE:
z-index is no longer undefined in the layer state since #12284


tschaub (approved):
Thanks, @MoonE

PR #15989, opened 2024-07-13
No implicit this

MoonE:
This fixes all typescript implicit-this warnings. Most were straight-forward, in `xml.js` it could by any of the `XMLFormat` sub-classes, not sure if something else., but also in the affected methods it doesn't really matter because the this-type is just passed through. so just annotate them as `*`.

Two small improvements:
- remove the `called` variable in `memoizeOne`, it has been called if `lastArgs` is not `undefined`
- remove the call to `bind` in `listen` when `once` is `true`

### DIFF
--- a/src/ol/layer/Layer.js
+++ b/src/ol/layer/Layer.js
@@ -450,7 +450,7 @@ class Layer extends BaseLayer {
       this.mapPrecomposeKey_ = listen(
         map,
         RenderEventType.PRECOMPOSE,
-        function (evt) {
+        (evt) => {
           const renderEvent =
             /** @type {import("../render/Event.js").default} */ (evt);
           const layerStatesArray = renderEvent.frameState.layerStatesArray;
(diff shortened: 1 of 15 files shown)


src/ol/layer/Layer.js
```diff
@@ -450,7 +450,7 @@ class Layer extends BaseLayer {
       this.mapPrecomposeKey_ = listen(
         map,
         RenderEventType.PRECOMPOSE,
-        function (evt) {
+        (evt) => {
```
ahocevar, line 453: With an arrow function, you could use `addEventListener()` instead of `listen()` and get rid of the `mapPrecomposeKey_`.  Then instead you would store the `mapPrecomposeFunction_` for later use in `removeEventListener()` instead of `unlisten()`.

ahocevar (approved):
Looks good! I added a suggestion about getting rid of `listen()` / `unlinsten()` pairs and replacing them with `addEventListener()` and `removeEventListener()`. There are more places than `Layer.js` where this could be done. It does not have to happen as part of this pull request though, only if you feel like it.

PR #15990, opened 2024-07-13
Fix Raster source attributions bugs

MoonE:
Two bugs with Raster source attributions.

- when attributsions is configured as a function returning a string there was an error
- when the combined attributions array of the raster source is empty it returned `null` which was converted to an array with null as its only element here:
https://github.com/openlayers/openlayers/blob/5cfa6a7ab5e1b7872570b0800ba6754e1c37400b/src/ol/layer/Layer.js#L352-L356

### DIFF
--- a/test/browser/spec/ol/source/raster.test.js
+++ b/test/browser/spec/ol/source/raster.test.js
@@ -47,7 +47,7 @@ where('Uint8ClampedArray').describe('ol.source.Raster', function () {
     greenSource = new Static({
       url: green,
       imageExtent: extent,
-      attributions: ['green raster source'],
+      attributions: (frameState) => 'green raster source',
     });
 
     blueSource = new VectorImageLayer({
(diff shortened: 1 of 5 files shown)


test/browser/spec/ol/source/raster.test.js
```diff
@@ -47,7 +47,7 @@ where('Uint8ClampedArray').describe('ol.source.Raster', function () {
     greenSource = new Static({
       url: green,
       imageExtent: extent,
-      attributions: ['green raster source'],
+      attributions: (frameState) => 'green raster source',
```
ahocevar, line 50: Since the array of string flavour should still work, it would be better to add another source to this test case or create a separate test case for the function flavour.

ahocevar (changes requested):
Thanks @MoonE. Looks good, but please consider my suggestion for the tests.

test/browser/spec/ol/source/raster.test.js
MoonE: `redSource` above still uses an array of strings, and they are used in combination, so I was hoping it'd be ok.

ahocevar: Oh, sorry! I just looked at the diff, not the context. In this case it's fine.

PR #15996, opened 2024-07-16
Remove last remaining `@inheritDoc`

MoonE:
One instance of `@inheritDoc` survived #10840

Overriding the method in src/ol/renderer/canvas/TileLayer.js is not needed as the `getImage` method does not accept any arguments.


ahocevar (approved):
Thanks, @MoonE !

PR #15997, opened 2024-07-16
Enable no implicit override check

MoonE:
https://www.typescriptlang.org/tsconfig/#noImplicitOverride


ahocevar (approved):
Thanks, @MoonE

PR #15998, opened 2024-07-16
Improve listen calls

MoonE:
@ahocevar In most cases the listen call is made with a class method and therefore the bind step cannot be skipped.
Not quite sure which style is better
- storing an arrow function for each instance of a class in the constructor with addEventListener
- a private method with storing the keys returned by listen
- a private method, storing a bound function in the constructor with addEventListener

Besides these changes I only found some in src/ol/reproj, though I'm curious in [`ol/repoj/Image`](https://github.com/openlayers/openlayers/blob/main/src/ol/reproj/Image.js) the listeners are removed on dispose
https://github.com/openlayers/openlayers/blob/c201f4c62d4367bdde2fd2bc828f366faedb9544/src/ol/reproj/Image.js#L162-L164
In [`src/ol/reproj/DataTile`](https://github.com/openlayers/openlayers/blob/main/src/ol/reproj/DataTile.js) and [`ol/reproj/Tile`](https://github.com/openlayers/openlayers/blob/main/src/ol/reproj/Tile.js) this is not done. Should it be added here?
I assume this skips the expensive reproject step in case the tile is no longer needed.

### DIFF
--- a/src/ol/layer/Layer.js
+++ b/src/ol/layer/Layer.js
@@ -446,23 +446,30 @@ class Layer extends BaseLayer {
       this.mapPrecomposeKey_ = listen(
         map,
         RenderEventType.PRECOMPOSE,
-        (evt) => {
-          const renderEvent =
-            /** @type {import("../render/Event.js").default} */ (evt);
-          const layerStatesArray = renderEvent.frameState.layerStatesArray;
-          const layerState = this.getLayerState(false);
-          assert(
-            !layerStatesArray.some(function (arrayLayerState) {
-              return arrayLayerState.layer === layerState.layer;
-            }),
-            'A layer can only be added to the map once. Use either `layer.setMap()` or `map.addLayer()`, not both.',
-          );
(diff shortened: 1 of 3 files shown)


ahocevar:
> In [src/ol/reproj/DataTile](https://github.com/openlayers/openlayers/blob/main/src/ol/reproj/DataTile.js) and [ol/reproj/Tile](https://github.com/openlayers/openlayers/blob/main/src/ol/reproj/Tile.js) this is not done. Should it be added here?
I assume this skips the expensive reproject step in case the tile is no longer needed.

Good question, but since we haven't noticed any memory leaks with reprojected tiles, I assume it's okay to leave it as it is.

ahocevar:
> * storing an arrow function for each instance of a class in the constructor with addEventListener

At least that's what I've been preferring recently.

ahocevar (approved):
It will be good to have these improvements in, even if we leave reproj etc. like it is. Thanks for your efforts, please merge!

PR #16003, opened 2024-07-18
Fix Linked list concatenateion

MoonE:
Though maybe it should be removed completely, as it is not used anymore which I discovered a bit late ...


ahocevar:
Good find that it isn't used anywhere. Removing it would indeed be the preferred option then. Thanks!

ahocevar (approved):

PR #16021, opened 2024-07-24
Fix cluster geometryFunction return type

### LINKED ISSUE #16020: Cluster's geometryFunction return type does not allow "null"
**Describe the bug**
In a _TypeScript_ project you will get an error if you try to return `null` from `ol/source/Cluster`'s geometryFunction. Currently, only `ol/geom/Point` is allowed as return type.

In regards to the [documenation](https://openlayers.org/en/latest/apidoc/module-ol_source_Cluster-Cluster.html), the geometryFunction should return `null` if a feature should not be considered.

```
Type '(feature: Feature<Geometry>) => Point | null' is not assignable to type '(arg0: Feature<Geometry>) => Point'.
  Type 'Point | null' is not assignable to type 'Point'.
    Type 'null' is not assignable to type 'Point'.
```

**To Reproduce**
```typescript
const clusterSource = new ClusterSource({
  source: new VectorSource(),
  geometryFunction: (feature) => {
    const geometry = feature.getGeometry();
    if (geometry && geometry.getType() === "Point") {
      return geometry as Point;
    } else {
      return null;
    }
  },
});
```

**Expected behavior**
`null` should be a valid return type of the geometryFunction.

NiclasSchmitz:
The geometryFunction can return `null` if a feature should not be considered.

Fixes #16020

### DIFF
--- a/src/ol/source/Cluster.js
+++ b/src/ol/source/Cluster.js
@@ -27,7 +27,7 @@ import {getUid} from '../util.js';
  * By default no minimum distance is guaranteed. This config can be used to avoid
  * overlapping icons. As a tradoff, the cluster feature's position will no longer be
  * the center of all its features.
- * @property {function(FeatureType):(Point)} [geometryFunction]
+ * @property {function(FeatureType):(Point|null)} [geometryFunction]
  * Function that takes a {@link module:ol/Feature~Feature} as argument and returns a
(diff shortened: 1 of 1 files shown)


ahocevar (approved):
Thanks, @NiclasSchmitz

Issue #8570, opened 2018-08-31
Night and Day Layer


pedrolopes10:
Is there a way to accomplish the following layer:

![preview](https://user-images.githubusercontent.com/5683649/44931877-0ad28c80-ad5b-11e8-9414-10e921bee443.jpg)

https://github.com/rossengeorgiev/nite-overlay

I've tried to adapt with regular Circles ( https://jsfiddle.net/5mvk1oc9/ ) and also with circular polygons but with no success.

Thanks for this great library, it has been a pleasure working with it!

RalphLAtGitHub:
I have not found a simple solution until today. From my point of view you will have to calculate the actual solar position, compute a circle (in parametric form) around the opposite point with the different radii for civil twilight, nautical twilight etc., create a polygon from each circle with a suitable increment, split the polygons if necessary at the antimeridian, transform the polygons to your projection and add them as features to a vector source. Any better idea is welcome!

pedrolopes10:
Hi @RalphLAtGitHub! The first part is all done on the library 'nite-overlay'. If you check the fiddle https://jsfiddle.net/5mvk1oc9/ I've done the same but with perfect circles. I couldn't apply a new projection or do a circular polygon with the calculated points.

RalphLAtGitHub:
Hi @pedrolopes10, ol.geom.Circle generates the perfect circle in the Mercator projection. You need a circle on the sphere (i.e. see here:   [https://math.stackexchange.com/a/1349100](url)) which is transformed to the Mercator projection. Depending on the position of the sun, this results in an ellipse or a sinusoid.

pjsg:
I have a custom rendered day/night layer that works reasonably well in ol5. It does use WebGL to do the calculations. Check out https://pskreporter.info/pskmap for an example. As a gimmick, it also draws the shadow of the moon during solar eclipses.

ThomasG77:
I've more or less done a library to generate a GeoJSON of night & days and then display it in OpenLayers. Not as good as mentioned Google maps "nite-overlay" that display civil twilight, nautical twilight, astronomical twilight and night but it does display night & days.
See https://github.com/webgeodatavore/GeoJSON.Terminator/ and associated [OpenLayers demo ](https://rawgit.com/webgeodatavore/GeoJSON.Terminator/master/demo/index.html)

TagLoomis:
[...]

PR #16028, opened 2024-07-26
Add WebGL Night Tile Layer

kikuchan:
This new layer renders a night shadow, accomplish #8570.

image

It renders a tile from a tile source as usual, but it only renders an image as a shadow by changing alpha value of the pixel depending on the shadow darkness. The tile would be opaque for where the midnight, transparent for where the daytime.

In this way, you can easily mix two layers, one for day and the other for night.
If no sources are specified, it uses a new `ol/source/DummyTile` as a fallback to render a simple matt shadow on the map.

The key idea is `ol/source/CoordinateTile`. It embeds projected latitude/longitude into a floating point texture tile, so that we can lookup it's location on shaders.

I had to change webgl helper a bit, and had to modify WebGLTileLayer and WebGLTileLayerRenderer.

The most tricky part is a `renderTile()` in `ol/renderer/webgl/NightTileLayer`. It manually `load()` the corresponding coordinate tile for the image tile to be render. It works, but if you know a better way, please let me know.

Thank you.


kikuchan:
Sorry for the subsequent force pushes.
I realized that I need a `CompositeTile` to mix `DataTile` and `ImageTile`, allowing us to expand the band count of existing tile sources. In this way, I can use the existing tile loading mechanism to load additional data tiles. Essentially, it's the same method that `ReprojTile` uses.

How does the implementation I pushed this time look?
Comments are welcome!
Thank you.

Issue #16025, opened 2024-07-25
Add static attribution to attribution control


RobQuincey-DC:
The attribution control uses the visible layers 'attribution' property to determine what attributions to show. There are instances where you may want a particular attribution to be always shown, such as attribution information for the app, or for OpenLayers itself. 

Some have used workarounds ([example SO question](https://stackoverflow.com/questions/60673165/openlayer-static-attribution-on-the-bottom-left-of-the-map)) that involve manually forcing this attribution into every layers attribution array, meaning it will always display. This feels sub-optimal.

I propose that a new option is added to the Attribution control, along the links of `staticAttribution`. If this is defined, it is always added the attributions list regardless of the current layers.

I have a draft implementation ready to go if you believe something like this is a good idea.
[Screen capture of potential implementation](https://github.com/user-attachments/assets/df703fa7-64e0-479f-b47d-dd1890efc14f)

mike-000:
The simplest workaround is to add an empty layer `new VectorLayer({source: new VectorSource({attributions: 'Static attribution'})})` but that is still a workaround, an option in the control constructor would be more logical.

RobQuincey-DC:
Closed via #16031

PR #16031, opened 2024-07-29
Add static `attributions` to attribution control constructor options

RobQuincey-DC:
This PR adds an optional `staticAttribution` option to the attribution control.

A static attribution is one that always remains in the attributions list, and is not linked to a layer source. This can be used to attribute the application itself, or as a way of attributing OpenLayers or other services. 

The tests have been updated to take into account static attributions, and an example has been added.

Linked issue: #16025


src/ol/control/Attribution.js
```diff
@@ -192,10 +200,16 @@ class Attribution extends Control {
    */
   collectSourceAttributions_(frameState) {
     const layers = this.getMap().getAllLayers();
-    const visibleAttributions = Array.from(
+    let visibleAttributions = Array.from(
       new Set(layers.flatMap((layer) => layer.getAttributions(frameState))),
     );
 
+    if (this.staticAttribution_ !== undefined) {
```
RobQuincey-DC, line 207: I'm not entirely happy with how this works (recreating the array again). I feel there must be a way to do it in one go, but simply pushing to the `visibleAttributions` array didn't work as the de-duplication of using a `Set` is then lost when adding the static attribution. Suggestions welcome

test/browser/spec/ol/control/attribution.test.js
```diff
@@ -31,6 +31,7 @@ describe('ol.control.Attribution', function () {
         new Attribution({
           collapsed: false,
           collapsible: false,
+          staticAttribution: 'static',
```
RobQuincey-DC, line 34: Should I be adding another test to handle having a static attribution vs not having one. I think this is covered as part of the other tests, but I can add another if you think its needed.

src/ol/control/Attribution.js
```diff
@@ -192,10 +200,16 @@ class Attribution extends Control {
...
+      visibleAttributions = Array.from(
+        new Set([...visibleAttributions, this.staticAttribution_]),
+      );
+    }
```
M393, line 211: Just skip the Array.from all and store the Set?
```suggestion
    const visibleAttributions = new Set(layers.flatMap((layer) => layer.getAttributions(frameState)));
    if (this.staticAttribution_ !== undefined) {
      visibleAttributions.add(this.staticAttribution_);
    }
```

and at the end:
```js
    return Array.from(visibleAttributions);
```

src/ol/control/Attribution.js
```diff
@@ -33,6 +33,8 @@ import {toPromise} from '../functions.js';
  * @property {function(import("../MapEvent.js").default):void} [render] Function called when
  * the control should be re-rendered. This is called in a `requestAnimationFrame`
  * callback.
+ * @property {string|undefined} [staticAttribution] Optional attribution that will always be
```
M393, line 36: Probably should also allow an array of strings to allow deduplicating multiple attributions.

RobQuincey-DC: Good point, hopefully that's done now

src/ol/control/Attribution.js
RobQuincey-DC: Since I've now changed it to handle an array of strings now, this has been refactored, taking your advice.

RobQuincey-DC:
I'm not sure why the browser test failed, it appears to have nothing to do with what I did

ahocevar:
I triggered a re-run of the tests, and they pass now. The failing test had indeed nothing to do with your change, that's something WebGL related and @jahow wanted to look into it.

ahocevar (changes requested):
Thanks for your contribution, @RobQuincey-DC.

It would be good if you could create a new test that tests if static attribution works at all, and then a second one to ensure it is not duplicated. And it would be best to not do that in an existing test, but to create a new one as well.

Other than that, the code looks good. A suggestion would be to rename the new option from `staticAttribution` to `attributions`, and accept an array of attributions. I think that would make it more consistent with the `attribution` option on `Source`.

RobQuincey-DC:
@ahocevar I've made your suggested changes. I hope the tests make sense, tried to cover every base, but if I missed anything let me know

ahocevar (commented):
Thanks, @RobQuincey-DC. The tests look good, but it looks like you need to update the example now to use the `attributions` property instead of `staticAttribution`. Also, it would be good if you could rebase your pull request with

    git pull --rebase openlayers main

assuming you have set up `git@github.com:openlayers/openlayers.git` as remote `openlayers`.

examples/static-attribution.js
```diff
@@ -0,0 +1,29 @@
+import Map from '../src/ol/Map.js';
+import OSM from '../src/ol/source/OSM.js';
+import TileLayer from '../src/ol/layer/Tile.js';
+import View from '../src/ol/View.js';
+import {Attribution, defaults as defaultControls} from '../src/ol/control.js';
+
+const attribution = new Attribution({
+  collapsible: false,
+  staticAttribution: `<a href="https://openlayers.org">I'm a static attribution. I never disappear</a>`,
```
ahocevar, line 9:
```suggestion
  attributions: `<a href="https://openlayers.org">I'm a static attribution. I never disappear</a>`,
```

ahocevar (approved):

ahocevar:
Thanks, @RobQuincey-DC

RobQuincey-DC:
@ahocevar Thanks for sorting that, I was on leave last week so couldn't get round to it, so I appreciate you stepping in and sorting it for me and for your helpful reviews!

PR #16041, opened 2024-07-29
Add v9.2.4 to the API version selector and bump 8.1.0 to 8.2.0.

tplunket:
Documentation update.

Updating "latest" to version 10 left version 9 orphaned in the UI so entries to the version 9.2.4 API were added. During this work, it was found that the latest version 8 reference was to version 8.1.0, so that was bumped to 8.2.0 as well.


ahocevar (approved):
Thanks, @tplunket

PR #16044, opened 2024-07-30
Do not render when no tiles for current z

### LINKED ISSUE #16038: Rendering error when tile source provides only a sub extent of the view projection
**Describe the bug**
When a tile layer's source extent is outside the current view extent an error occurs during rendering at this line because `tilesByZ[z]` is undefined.
https://github.com/openlayers/openlayers/blob/ff9fa42a28ffc710ccfe602fcd176e1359c1c233/src/ol/renderer/canvas/TileLayer.js#L636

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/wmts-hidpi.html
2. Pan Austria outside the view
3. See error

**Expected behavior**
No error, just don't render the layer.

ahocevar:
Fixes #16038

### DIFF
--- a/config/tsconfig-strict.json
+++ b/config/tsconfig-strict.json
@@ -174,6 +174,7 @@
     "../src/ol/render/canvas/PolygonBuilder.js",
     "../src/ol/render/canvas/style.js",
     "../src/ol/render/canvas/TextBuilder.js",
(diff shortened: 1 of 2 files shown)


tschaub (approved):
Thanks, @ahocevar

PR #16051, opened 2024-08-02
Fix canvas group change on webgl layers

### LINKED ISSUE #15698: Features rendered with WebGLVectorLayerRenderer disappear when WebGLTile layer is added to the map
**Describe the bug**
I discovered a bug with a certain combination of GL and non-GL layers (see example below). OL somehow makes the features rendered with experimental *WebGLVectorLayerRenderer* disappear if you add/remove another GL tile layer to the map, it is needed to pan/zoom the map a little to make them visible again. The bug disappears if you remove the classic vector layer from the map (named `vectors` in the example - line 59 - it contains the green line). This happens in FF and also in Chrome.

**To Reproduce**
- I made this fiddle: https://jsfiddle.net/ghx0eb2u/
- Click the "Add or remove GL tile layer" button
- see that the red line disappears when GL tile layer is added/removed

**Expected behavior**
The red line should be visible all the times.

**Use case**
I use normal tile layer for background map, WebGL tile for shading, vectors for some features and I wanted to use *WebGLVectorLayerRenderer* for some special lines with image pattern. The terrain shading layer can be turned on/off and that makes the GL lines invisible.

kikuchan:
Dynamically inserting a normal layer between tandem WebGL-based layers causes a separation of the underlying WebGL canvas, which in turn causes a recreation of `this.helper` in `WebGLLayerRenderer` because their `groupNumber` will differ.

Almost everything is handled properly, but the WebGL layers need to upload their contents again to the newly assigned WebGL context.

For example, setting up layers `[glBaseLayer, layerGroup, glOverlay]` and inserting/removing a normal `new VectorLayer()` into/from `layerGroup` by using `layerGroup.getLayers().push()`/`.pop()` affects the `glOverlay` visuals, causing it to disappear or flicker.

This PR try to resolve/mitigate the issue.

Additionally, this also fixes #15698.


kikuchan:
Updated.
- Fix an error on test/rendering
- Add tests for this issue

kikuchan:
This fixes #15698.

ahocevar (approved):
Thanks, @kikuchan

kikuchan:
Thank you!

PR #16054, opened 2024-08-04
Fix #16039 webgl vector layer flashes

### LINKED ISSUE #16039: WebGL Vector Layer example flashes on hover or click
**Describe the bug**
The OSM TileLayer in [this example](https://openlayers.org/en/latest/examples/webgl-vector-layer.html) goes blank when hovering over features with a mouse (desktop; Chrome/Edge/Firefox) and when tapping via mobile (only tested on Chrome).

**Expected behavior**
I expect all layers to be stable when interacting with the map.

kikuchan:
This PR try to fix #16039.

### DIFF
--- a/src/ol/renderer/webgl/VectorLayer.js
+++ b/src/ol/renderer/webgl/VectorLayer.js
@@ -156,6 +156,12 @@ class WebGLVectorLayerRenderer extends WebGLLayerRenderer {
      */
     this.buffers_ = [];
 
+    /**
+     * @type {Array<import('../../render/webgl/VectorStyleRenderer.js').WebGLBuffers>}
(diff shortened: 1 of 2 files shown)


jahow:
Hi @kikuchan, thank you for looking into this! I tried your fix and found that there were still errors from time to time. I'm not 100% sure why but basically the buffers can sometimes be bound for rendering while being also marked for deletion, and that causes INVALID_OPERATION errors and flashing.

Could you take a look at #16055 and let me know what you think about it? thanks!!

kikuchan:
Hi @jahow, thank you for reviewing this.

And indeed, your solution looks perfect to me!

PR #16055, opened 2024-08-04
Fix flashing on the WebGL vector example

### LINKED ISSUE #16039: WebGL Vector Layer example flashes on hover or click

jahow:
This fixes #16039 and is an alternative to #16054

Calling `gl.deleteBuffer` was marking the buffer for deleting which caused INVALID_OPERATION errors, see: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bindBuffer#exceptions

We can omit the call safely as the buffers are garbage collected automatically once they are not bound anymore.

### DIFF
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -527,12 +527,8 @@ class WebGLHelper extends Disposable {
    * @param {import("./Buffer.js").default} buf Buffer.
    */
   deleteBuffer(buf) {
-    const gl = this.gl_;
     const bufferKey = getUid(buf);
-    const bufferCacheEntry = this.bufferCache_[bufferKey];
-    if (bufferCacheEntry && !gl.isContextLost()) {
-      gl.deleteBuffer(bufferCacheEntry.webGlBuffer);
-    }
+    // Note: gl.deleteBuffer is not called here since we let WebGL garbage collect it automatically
     delete this.bufferCache_[bufferKey];
   }
 


kikuchan:
LGTM :+1:

tschaub:
@jahow - any idea what operation is using the buffer after deletion?

jahow:
I am not 100% certain, in theory the buffers were only deleted once they were not needed anymore. This being said, rendering happens asynchronously in webgl so what might happen is that the draw call was happening slightly later than the `renderFrame` call, thus ending up relying on deleted buffers. I'm guessing that's why @kikuchan seemed to fix the issue almost perfectly (by simply delaying the deletion of 1 frame), but it wasn't completely error-proof.

kikuchan:
> any idea what operation is using the buffer after deletion?

Actually, that's what I'm really wondering about...

FWIW the following warning messages in Firefox console when the flicker occurs:
```
WebGL warning: drawElementsInstanced: Vertex attrib array 1 is enabled but has no buffer bound. 16
WebGL warning: drawArraysInstanced: Vertex attrib array 1 is enabled but has no buffer bound.
WebGL warning: drawElementsInstanced: Vertex attrib array 4 is enabled but has no buffer bound.
```

I tried hooking the helper functions to detect suspicious activity such as "double-deletion" and/or "bind after deletion", but I didn't find any sign with the vector layer example. Besides, the MDN document in `bindBuffer` says:
> A WebGLBuffer which has been marked for deletion with deleteBuffer cannot be (re-)bound. An attempt to do so will generate an INVALID_OPERATION error, and the current binding will remain untouched.

and, I've never seen INVALID_OPERATION for this example in my environment. So I believe we've already handled the buffer lifetime properly, as long as with the helper.

I also tried to find a case for missing `bindBuffer` call by inserting `gl.bindBuffer(..., null);` everywhere, but no effects.

A delay after commands sent? But `gl.flush()` didn't work.

Really weird, but it would make sense if there were a delay between the buffer binding and rendering, becasue the buffer swap happens in a Promise callback.

If that's the case, this PR will be really effective.

kikuchan:
I'm still trying to track the issue down... It's really hard.

I'm sorry, but I didn't know that `gl.getError()` must be called to get the `INVALID_OPERATION` error. The error surely happens on my system too.
As a side note, it can be helpful to see what's happening by setting the `className` on the `featureOverlay` layer to create its own independent GL context.

It feels like deleting one buffer might be corrupting another.

Here's a stack trace when the issue happens:
image
If the buffer needs to be kept alive until the final `PostProcessPasses.apply` call, that makes sense. It could be interrupted by an async microtask, so there's a chance the active buffer might be deleted in the meantime.


Oh btw, do we need to turn `preserveDrawingBuffer` off in `getContext()` within `ol/webgl.js` if we don't explicitly delete the buffer perhaps?

tschaub:
My assumption is that this issue occurs when we call something in the WebGL API after calling `deleteBuffer` (as opposed to some asynchronous rendering outside our control causing the issue).  It makes sense to me that the issue goes away if we stop calling `deleteBuffer`.  But that also might just be masking another issue with our use of the buffers.

kikuchan:
Eureka! I'm sorry I was totally wrong!
Thank you for pointing it out.

I've found a solution to avoid this issue by the following patch:
```
diff --git a/src/ol/webgl/Helper.js b/src/ol/webgl/Helper.js
index d94ef71e0f..4a0c99c738 100644
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -1047,6 +1047,13 @@ class WebGLHelper extends Disposable {
   enableAttributes(attributes) {
     const stride = computeAttributesStride(attributes);
     let offset = 0;
+    const gl = this.gl_;
+
+    const maxAttributes = gl.getParameter(gl.MAX_VERTEX_ATTRIBS);
+    for (let i = 0; i < maxAttributes; i++) {
+      gl.disableVertexAttribArray(i);
+    }
+
     for (let i = 0; i < attributes.length; i++) {
       const attr = attributes[i];
       this.enableAttributeArray_(
diff --git a/src/ol/webgl/PostProcessingPass.js b/src/ol/webgl/PostProcessingPass.js
index 6b7901dc36..a5b7dffe28 100644
--- a/src/ol/webgl/PostProcessingPass.js
+++ b/src/ol/webgl/PostProcessingPass.js
@@ -335,6 +335,10 @@ class WebGLPostProcessingPass {
     gl.bindBuffer(gl.ARRAY_BUFFER, this.renderTargetVerticesBuffer_);

     gl.useProgram(this.renderTargetProgram_);
+    const maxAttributes = gl.getParameter(gl.MAX_VERTEX_ATTRIBS);
+    for (let i = 0; i < maxAttributes; i++) {
+      gl.disableVertexAttribArray(i);
+    }
     gl.enableVertexAttribArray(this.renderTargetAttribLocation_);
     gl.vertexAttribPointer(
       this.renderTargetAttribLocation_,
```

As you can see, disabling previously enabled vertex attribute arrays appears to be necessary.

Unintentionally enabled vertex attribute arrays could lead to accessing data outside the buffer.
If we call `deleteBuffer`, this would invalidate regions outside the buffer, potentially causing an error.
If we don't, referencing invalid regions would occur less frequently.

jahow:
Oh my god @kikuchan very well found, disabling all attributes before enabling new ones indeed works!! I would never have found that. I just hope that this won't have a hit on performance, since we're doing this clearing on every render frame.

jahow:
Ok @kikuchan @tschaub hear me out, as this is becoming a can of worms:

I think we should stick to my previous solution and remove the call to `deleteBuffer`.

1. `deleteBuffer` is _optional_, as clearly stated in the docs. WebGL will garbage collect unused buffers on its own.
2. calling multiple times `disableVertexAttribArray` _has_ a cost, these calls amount to multiple milliseconds across frames. And that makes sense: any `gl.*` call has a cost, be it for reading, setting, clearing etc.

To make up my opinion I have searched through [twgl.js](http://twgljs.org/) codebase; it's not an absolute reference but I'm pretty sure it's battle-tested and reliable. There is **no call** to `gl.deleteBuffer` and the only calls to `gl.disableVertexAttribArray` are done when setting up a new program (shader) for rendering; so at no point is it disabling previously enabled attribs.

Another post that makes me think that is this one: https://stackoverflow.com/a/36301555 (pretty old but still)

Basically I think the best approach here in terms of perf is: leaving old buffers still dangling and enabled; they won't be used by the other shaders anyway, so they don't hurt, but at least their indices will still point to a valid buffer. They will get garbage collected automatically (I have checked repeatedly that removing `gl.deleteBuffer` will _not_ increase the memory consumption).

Note that WebGL2 has the concept of VAO (vertex attrib objects) which replaces all this tedious stateful gl mess.

kikuchan:
I agree.
Let's merge this! 😄

jahow:
@tschaub could you please give this a final look and let me know if that was convincing enough? Thank you very much for your time

kikuchan:
I tried some experiments:
* disabling **all** attributes before enabling attributes, works.
* disabling **all** attributes immediately after deleting the buffer, works.
* disabling **enabled** attributes before enabling attributes, **doesn't work**.
* disabling **enabled** attributes immediately after deleting the buffer, **works**.

It's really weird.

The last option is the most efficient (as long as prefer disabling attributes), but deleting the buffer and disabling attributes seem unrelated at first glance. Besides, it doesn't make sense that the 3rd option doesn't work while the 1st option does. (Disabling already disabled attributes affects something...?)

Anyway, controlling unpredictable timing is a nightmare...

Here's a patch just in case:
```patch
diff --git a/src/ol/webgl.js b/src/ol/webgl.js
index 72144bb0cc..b6ab193653 100644
--- a/src/ol/webgl.js
+++ b/src/ol/webgl.js
@@ -130,3 +130,13 @@ export function getSupportedExtensions() {
   }
   return supportedExtensions;
 }
+
+/**
+ * @param {WebGLRenderingContext} gl WebGL rendering context.
+ */
+export function disableAllAttributes(gl) {
+  const maxAttributes = gl.getParameter(gl.MAX_VERTEX_ATTRIBS);
+  for (let i = 0; i < maxAttributes; i++) {
+    gl.disableVertexAttribArray(i);
+  }
+}
diff --git a/src/ol/webgl/Helper.js b/src/ol/webgl/Helper.js
index d94ef71e0f..fbaa307eb2 100644
--- a/src/ol/webgl/Helper.js
+++ b/src/ol/webgl/Helper.js
@@ -9,6 +9,7 @@ import {
   UNSIGNED_BYTE,
   UNSIGNED_INT,
   UNSIGNED_SHORT,
+  disableAllAttributes,
   getContext,
 } from '../webgl.js';
 import {clear} from '../obj.js';
@@ -446,6 +447,12 @@ class WebGLHelper extends Disposable {
      * @private
      */
     this.startTime_ = Date.now();
+
+    /**
+     * @type {Object<string, number>}
+     * @private
+     */
+    this.enabledAttributes_ = {};
   }

   /**
@@ -532,10 +539,23 @@ class WebGLHelper extends Disposable {
     const bufferCacheEntry = this.bufferCache_[bufferKey];
     if (bufferCacheEntry && !gl.isContextLost()) {
       gl.deleteBuffer(bufferCacheEntry.webGlBuffer);
+      this.disableAttributes();
     }
     delete this.bufferCache_[bufferKey];
   }

+  /**
+   * @private
+   */
+  disableAttributes() {
+    //disableAllAttributes(this.gl_);
+    const gl = this.gl_;
+    for (const l in this.enabledAttributes_) {
+      gl.disableVertexAttribArray(this.enabledAttributes_[l]);
+    }
+    this.enabledAttributes_ = {};
+  }
+
   /**
    * Clean up.
    * @override
@@ -1036,6 +1056,7 @@ class WebGLHelper extends Disposable {
     }
     this.gl_.enableVertexAttribArray(location);
     this.gl_.vertexAttribPointer(location, size, type, false, stride, offset);
+    this.enabledAttributes_[location] = location;
   }

   /**
@@ -1045,6 +1066,8 @@ class WebGLHelper extends Disposable {
    * @param {Array<AttributeDescription>} attributes Ordered list of attributes to read from the buffer
    */
   enableAttributes(attributes) {
+    //this.disableAttributes();
+
     const stride = computeAttributesStride(attributes);
     let offset = 0;
     for (let i = 0; i < attributes.length; i++) {
```

jahow:
Option 3 is unreliable because there might be multiple helpers working with the same context, so the already enabled attributes cannot be tracked on the helper.

Option 4 sounds like it could definitely work. I still think that we should avoid these calls as it will have an impact on performance.

kikuchan:
Yeah, that's one of the reasons why we shouldn't call `deleteBuffer` manually.
I think there's almost no benefit to calling `disableVertexAttribArray` precisely as it incurs a cost.

kikuchan:
> multiple helpers working with the same context
Oh sorry, right, and I'm really relief because it makes sense. I've updated #16069 just in case, and it works fine as I expected.

However, as you can see, it's too much lines of code, isn't it?
Also, it costs to detect unused vertex attributes every frame indeed, as you mentioned.

tschaub:
Thanks for all your work digging into this, @kikuchan and @jahow

tschaub (approved):

PR #16056, opened 2024-08-04
WebGL / fix flaky tests relying on patterns

### LINKED ISSUE #15942: Flaky WebGL style parsing test
The `test/browser/spec/ol/webgl/styleparser.test.js` tests often fail.  See https://github.com/openlayers/openlayers/actions/runs/9643433220/job/26593423765?pr=15934 for example.

It looks like the generated shader code will differ depending on whether an image is loaded at parse time or not: https://github.com/openlayers/openlayers/blob/744ba5bbb3fc5ee2bcbe03ee4e32082da863aa98/src/ol/webgl/styleparser.js#L212-L221

We should either update the tests so that they tolerate this or change the code that generates the shader.

cc @jahow

jahow:
This PR make it so that image sizes will always be provided asynchronously using a uniform in WebGL shaders.

Fixes #15942 

The previous logic was not using a uniform if an image was be ready synchronously. This would happen in very rare cases and in an unreliable way; if we don't want to make the style parsing asynchronous we have to always use a uniform for the image size.

This logic was probably an insignificant optimization anyway, as uniforms are not costly in shaders.

### DIFF
--- a/src/ol/webgl/styleparser.js
+++ b/src/ol/webgl/styleparser.js
@@ -12,14 +12,13 @@ import {
   newParsingContext,
 } from '../expr/expression.js';
 import {ShaderBuilder} from './ShaderBuilder.js';
+import {asArray} from '../color.js';
 import {
-  arrayToGlsl,
   buildExpression,
   getStringNumberEquivalent,
   stringToGlsl,
   uniformNameForVariable,
 } from '../expr/gpu.js';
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @jahow

jahow:
Looks like it's better now :) thanks @ahocevar

PR #16057, opened 2024-08-05
Add a coordinate tile source

kikuchan:
Add a new source, `CoordinateTileSource`, to provide coordinate information (not limited to latitude and longitude) for every pixel within a tile.
This floating-point tile is intended to be accessed via the `band` operator of the `EncodedExpression` in the `WebGLTileLayer` and styled accordingly.

The most notable application is drawing the "Earth's shadow", which is demonstrated in the example.
It calculates the Sun's altitude for every geographic location of each pixel, rendering it as a tile with an alpha channel, allowing it to be overlaid on a map.

This was part of #16028, but I've separated it to simplify things.

### DIFF
--- a/config/tsconfig-strict.json
+++ b/config/tsconfig-strict.json
@@ -211,6 +211,7 @@
     "../src/ol/source/BingMaps.js",
     "../src/ol/source/CartoDB.js",
     "../src/ol/source/Cluster.js",
+    "../src/ol/source/CoordinateTile.js",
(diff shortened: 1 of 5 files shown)


kikuchan:
Updated.
* Always utilize view projection to prevent reprojection within `CoordinateTileSource`.
* Update the example to refine the calculation of the subsolar position. 😉
https://deploy-preview-16057--ol-site.netlify.app/en/latest/examples/coordinate-tile-source.html

PR #16063, opened 2024-08-06
expression: has

bkuster:
# `has` expression

This PR provides a `has` expression to ol/expr which returns a boolean based on the context properties having a certain key (or nested key).

## Reasoning

It is not trivial to get a `has` or `defined` operation with the current expressions. Things like `["any", ["get", "bar"], ["==", ["get", "bar"], ""], ["==", ["get", "bar"], false], ["==", ["get", "bar"], 0]]` would work, but aren't very convenient.

## Solution

Similar to [mapbox has](https://docs.mapbox.com/style-spec/reference/expressions/#has) this expression will return `true` if the key was found in the properites, making the above expression reduce to `["has", "bar"]`. The syntax is identical to the `get` expression, where you can provide more then one key for nested properties.

### DIFF
--- a/src/ol/expr/expression.js
+++ b/src/ol/expr/expression.js
@@ -91,6 +91,8 @@ import {toSize} from '../size.js';
  *   * `['!', value1]` returns `false` if `value1` is `true` or greater than `0`, or `true` otherwise.
  *   * `['all', value1, value2, ...]` returns `true` if all the inputs are `true`, `false` otherwise.
  *   * `['any', value1, value2, ...]` returns `true` if any of the inputs are `true`, `false` otherwise.
+ *   * `['has', attributeName, keyOrArrayIndex, ...]` returns `true` if feature properties include the (nested) key `attributeName`,
+ *     `false` otherwise.
  *   * `['between', value1, value2, value3]` returns `true` if `value1` is contained between `value2` and `value3`
  *     (inclusively), or `false` otherwise.
  *   * `['in', needle, haystack]` returns `true` if `needle` is found in `haystack`, and
@@ -414,6 +416,7 @@ export const Ops = {
   Band: 'band',
   Palette: 'palette',
   ToString: 'to-string',
+  Has: 'has',
 };
 
 /**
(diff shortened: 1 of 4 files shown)


bkuster:
@jahow this is ready for review. im not a webgl expert, but from my understanding of shader programming, there is no `undefined` in glsl. if you have a proposal to get `has` to work for webgl as well, id happily give it a shot

jahow:
thanks! indeed in WebGL we need to introduce a special "undefined" value somehow, not sure yet how that would work out. I'll give this a review asap

src/ol/expr/expression.js
```diff
@@ -91,6 +91,8 @@ import {toSize} from '../size.js';
  *   * `['!', value1]` returns `false` if `value1` is `true` or greater than `0`, or `true` otherwise.
  *   * `['all', value1, value2, ...]` returns `true` if all the inputs are `true`, `false` otherwise.
  *   * `['any', value1, value2, ...]` returns `true` if any of the inputs are `true`, `false` otherwise.
+ *   * `['has', attributeName, keyOrArrayIndex, ...]` returns `true` if feature properties include the (nested) key `attributeName`,
+ *     `false` otherwise.
```
tschaub, line 95:
```suggestion
 *     `false` otherwise.  If the property exists but the value is `null` or `undefined`, the expression evaluates to `false`.
```

This behavior might be different than what people expect (different from [`Object.hasOwn()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) for example).  I think it should be documented at least.

tschaub:
I'm reworking our accessor expression handling (`get`, `var`) in https://github.com/openlayers/openlayers/compare/main...tschaub:openlayers:accessor-expressions.  This is going to involve changing the WebGL style parsing and evaluation, ideally sharing the evaluation of accessor expressions on the CPU.  I'll explain more when opening a pull request for that branch.  Just mentioning it here because I don't think it is worth trying to include `has` handling in WebGL as part of this pull request.

src/ol/expr/expression.js
bkuster: @tschaub I hadn't thought of this, you are right. Initially, I wanted to name the operation `defined`, but @jahow pointed out the `has` API and we decided to name it this way. But you are right, the mapbox `has` API tests `...for a presence of a property`, the big question being, if `undefined` or `null` count as present (same as they do with `hasOwn`).

I'm open to reverting to my initial idea of naming it `defined`, which would be a more exact description of the current expression. Additionally, if you see the use case, introduce a second operation `has` and use `Object.hasOwn` to test for `presence`.

tschaub: @bkuster - I don't think things would be clearer with `defined`.  For example, I can define a property on an object whose value is `undefined` (`{example: undefined}` or `Object.defineProperty({}, 'example', {value: undefined})`).

I think the semantics of `has` are easier to guess/understand.  The documentation that you included in your commit describes `has` well.  I pushed another commit that updates the implementation to match the documentation.

tschaub (approved):

PR #16066, opened 2024-08-07
Fix #15926: Check property values for undefined

### LINKED ISSUE #15926: WFS#writeTransaction() fails for undefined properties
The [`writeTransaction()`](https://openlayers.org/en/latest/apidoc/module-ol_format_WFS-WFS.html#writeTransaction) method of the `ol/format/WFS` class fails for features having `undefined` property values.

As an example, consider this snippet:

```js
import WFS from "ol/format/WFS";
import GML3 from "ol/format/GML3";
import Feature from "ol/Feature";

const wfs = new WFS({
    version: "2.0.0",
    gmlFormat: new GML3()
});

const feature = new Feature({
    someProperty: undefined
});

const transaction = wfs.writeTransaction([feature], [], [], {
    featurePrefix: "sf",
    featureNS: "http://www.openplans.org/spearfish",
    featureType: "bugsites",
    nativeElements: []
});
console.log("Transaction:", transaction);
```

Running the code fails with an error saying:

```
GML3.js:867 Uncaught TypeError: Cannot read properties of undefined (reading 'getSimplifiedGeometry')
    at wx.writeFeatureElement (GML3.js:867:67)
    at yT (WFS.js:806:21)
    at xml.js:273:16
    at np (xml.js:479:58)
    at rp (xml.js:523:3)
    at dT (WFS.js:724:3)
    at wb.format.WFS.writeTransaction (WFS.js:625:7)
```

Here's an excerpt from the [`writeFeatureElement()`](https://github.com/openlayers/openlayers/blob/main/src/ol/format/GML3.js#L844) method of the `GML3` class:

```js
writeFeatureElement(node, feature, objectStack) {
  /* ... */
  const properties = feature.getProperties();
  for (const key in properties) {
    const value = properties[key];
    if (value !== null) {
      keys.push(key);
      values.push(value);
      if (
        key == geometryName ||
        typeof (/** @type {?} */ (value).getSimplifiedGeometry) ===
          'function'
      ) {
        if (!(key in context.serializers[featureNS])) {
          context.serializers[featureNS][key] = makeChildAppender(
            this.writeGeometryElement,
            this,
          );
        }
      } else {
        if (!(key in context.serializers[featureNS])) {
          context.serializers[featureNS][key] =
            makeChildAppender(writeStringTextNode);
        }
      }
    }
  }
  /* ... */
}
```

Shouldn't the `value !== null` check be changed to `value != null`?

This occurs in OpenLayers v9.2.4.

m-elseberg:
This pull request fixes issue #15926.


ahocevar (commented):
Thanks, @m-elseberg. To avoid future regressions on this, would you be able to add a test?

m-elseberg:
@ahocevar I tried adding a test inside a new file called `/src/test/node/ol/format/WFS.test.js`, which does nothing but ensure the `writeTransaction()` method does not throw in the given scenario. (I hope that it meets your coding standards.)

However, since I am new to contributing to OpenLayers, I was not able to actually run the test due to some NPM problems: 

```sh
$ npm install
[...]

changed 752 packages, and audited 928 packages in 1m

182 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$ npm test

> ol@10.0.1-dev pretest
> npm run lint && npm run typecheck && npm run typecheck-strict && npm run build-package


> ol@10.0.1-dev lint
> eslint tasks test src/ol examples config


Oops! Something went wrong! :(

ESLint: 8.57.0

Error: Cannot find module 'callsites'
Require stack:
[...]

$ node -v
v20.10.0

$ npm -v
10.2.3
```

Could you provide me with guidance on how to resolve this missing-module error, or perhaps just run the test for me?

Beyond that, I noticed that your linter [does not seem to like](https://github.com/openlayers/openlayers/actions/runs/10282341763/job/28455066131?pr=16066) the `==` operator:

```
/home/runner/work/openlayers/openlayers/src/ol/format/GML2.js
Error:   221:13  error  Use '===' to compare with null  no-eq-null

/home/runner/work/openlayers/openlayers/src/ol/format/GML3.js
Error:   852:13  error  Use '===' to compare with null  no-eq-null
```

I thus replaced the two `value != null` checks by `value !== undefined && value !== null`.

ahocevar:
Thanks for your continued effort, @m-elseberg. The error you're getting when running the tests looks strange, and I cannot reproduce it with the same setup. However, the problem seems to come from `eslint`, so you should at least be able to `npm run test-node`.

The tests are well designed, it only seems that you have a typo in the imports. See my separate suggestion comment for a fix.

test/node/ol/format/WFS.test.js
```diff
@@ -0,0 +1,28 @@
+import Feature from '../../../../src/ol/Feature.js';
+import GML3 from '../../../../src/ol/format/GML2.js';
```
ahocevar, line 2:
```suggestion
import GML2 from '../../../../src/ol/format/GML2.js';
```

ahocevar (changes requested):

m-elseberg:
@ahocevar Thanks, running `npm run test-node` does in fact work on my machine. I adopted your suggestion on the import.

However, in this state, the test would still fail: The `createElementNS()` function defined in [`src/ol/xml.js`](https://github.com/openlayers/openlayers/blob/9f7630b1f3e0c9d3c6b5c6fc26ca5b508582caad/src/ol/xml.js#L34) would throw a `TypeError` due to the `getDocument()` function returning `undefined`.

I was able to resolve this problem by introducing a mocked `document` object using the [`jsdom`](https://www.npmjs.com/package/jsdom) library. I [committed](https://github.com/openlayers/openlayers/pull/16066/commits/cb872f00bd4250d856830849bfeac9aa48e10e31) this workaround, but then [reverted](https://github.com/openlayers/openlayers/pull/16066/commits/10194a588baad98b7e4696a196e4f87be36951a9) it, as I wasn't sure as to whether this would be the correct way of solving the problem.

With the workaround in place though, I was able to confirm that the test does succeed and, likewise, that the test fails when the fix proposed by this pull request is removed.

ahocevar:
@m-elseberg You could move the whole test file to `test/browser`, where `Document` will be available. Then run `npm run karma` to check if it works. Sorry, it was an oversight on my part, I should have thought about the missing `Document` in node.

m-elseberg:
@ahocevar Thanks, I integrated the test into the existing `/test/browser/spec/ol/format/wfs.test.js` file, and it seems to be working. Could you confirm that?

ahocevar (approved):
Looks good now! Thanks @m-elseberg.

PR #16067, opened 2024-08-07
Reset layer style when new GeoTIFF source is ready

### LINKED ISSUE #16065: Single-band GeoTIFFs behave as if they have three bands in style expressions
**Describe the bug**
For the purpose of style expressions, single-band GeoTIFFs behave as if they have 3 identical bands, which unexpectedly places the alpha band generated by the `nodata` source parameter into the fourth band.

**To Reproduce**
1. Construct a `GeoTIFF` source to a single-band GeoTIFF. In case it's important, the file that exhibited the issue for me has the following `gdalinfo` output:
    ```
    Band 1 Block=512x512 Type=Float32, ColorInterp=Gray
    NoData Value=nan
    Overviews: 38016x46208, 19008x23104, 9504x11552, 4752x5776, 2376x2888, 1188x1444, 594x722, 297x361
    ```
I constructed my GeoTIFF source as follows:
```javascript
const source = new GeoTIFF({
      normalize: false,
      interpolate: false,
      sources: [{ url: <no public source but I suspect any single-band GeoTIFF with a no-data value will work>, nodata: NaN, bands: [1] }],
});
cogLayer.setSource(source);
```
`cogLayer` is a `WebGLTileLayer()`.

Style the layer using something like
```javascript
cogLayer.setStyle({
      color: ["color", 255, 0, 0, ["band", 2]], // <-- Expecting this to be the nodata alpha band
    });
```

**Expected behavior**
I expected the above snippets to result in a GeoTIFF layer that displays bright red wherever there is data, and be transparent elsewhere, like this:

image

But the actual behavior is that the transparency mirrors the data band:
image 

[The docs for the `'band'` operator say](https://openlayers.org/en/latest/apidoc/module-ol_expr_expression.html#~ExpressionValue):
> [TileImage](https://openlayers.org/en/latest/apidoc/module-ol_source_TileImage-TileImage.html) sources have 4 bands: red, green, blue and alpha. [DataTileSource](https://openlayers.org/en/latest/apidoc/module-ol_source_DataTile-DataTileSource.html) sources can have any number of bands, depending on the underlying data source and [configuration](https://openlayers.org/en/latest/apidoc/module-ol_source_GeoTIFF.html#~Options).

Since `GeoTIFF` [inherits from `DataTileSource`](https://openlayers.org/en/latest/apidoc/module-ol_source_GeoTIFF-GeoTIFFSource.html), I took this to mean that the number of bands accessible by the `'band'` operator would be equal to the number of bands in the source, and so `['band', 1]` would be the data band and `['band', 2]` would return the alpha nodata mask. But that is not what actually happens; from within the style expression, `["band", 1]`, `["band", 2]`, and `["band", 3]` all return the same value, which is the same as the GeoTIFF's single band. The alpha value, on the other hand, is accessible via `["band", 4]`.

I hope this is clear, but I'm happy to provide any other info that would be helpful if it isn't!

mike-000:
Fixes #16065

If there is already a renderer when a new GeoTIFF source is set using `setSource()` it is necessary to wait until the source is ready before parsing the style with the correct bands.


ahocevar (approved):
Thanks, @mike-000

PR #16073, opened 2024-08-09
Clear the RenderBox when the DragBox is removed from the Map

### LINKED ISSUE #16072: Clear the RenderBox when the DragBox is removed from the Map.
If the DragBox is removed from the Map while drawing a box, the box that was being drawn will remain on the map.
So I'm filing an issue to address this issue.

If the DragBox interaction is removed from the Map while drawing a box, the box should disappear as well.

dlog6649:
Fixes #16072

DragBox interaction calls setMap when it is removed from the map. So I added code to clean up the RenderBox at that time. Additionally, if the box was being drawn, I also dispatched a cancel event.

### DIFF
--- a/src/ol/interaction/DragBox.js
+++ b/src/ol/interaction/DragBox.js
@@ -231,8 +231,6 @@ class DragBox extends PointerInteraction {
       return false;
     }
 
-    this.box_.setMap(null);
-
     const completeBox = this.boxEndCondition_(
       mapBrowserEvent,
       this.startPixel_,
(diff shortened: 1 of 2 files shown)


ahocevar (commented):
Thanks @dlog6649! Do you think you'd be able to add tests for the functionality you added?

dlog6649:
> Thanks @dlog6649! Do you think you'd be able to add tests for the functionality you added?

I'll check it out

dlog6649:
I checked it out, but it seems difficult for me.

ahocevar (approved):
Thanks @dlog6649. I added tests.

PR #16075, opened 2024-08-11
Fix color style for WebGL tile layer

### LINKED ISSUE #16074: Premultiplied-alpha applied twice on WebGLTileLayer
**Describe the bug**
Styling a tile by `color` operator with alpha value on WebGLTile layer, darken the color unexpectedly.

**To Reproduce**
```js
import Map from 'ol/Map.js';
import OSM from 'ol/source/OSM.js';
import TileLayer from 'ol/layer/WebGLTile.js';

const map = new Map({
  layers: [
    new TileLayer({
      source: new OSM(),
      style: {
        color: ['color', 255, 255, 255, 0.5],
      },
    }),
  ],
  target: 'map',
  view: new View({
    center: [0, 0],
    zoom: 2,
  }),
});
```
**Expected behavior**
The tile should be rendered as if `rgba(255, 255, 255, 0.5);` (transparent white), but it is rendered as if `rgba(128, 128, 128, 0.5);` (transparent gray) instead.

kikuchan:
This PR moves alpha premultiplication from the WebGL expression compiler to `styleparser`.

This change is necessary because alpha premultiplication strategies can differ between shaders.
The shader in `WebGLTileLayer` employs a lazy strategy, while the shader in `styleparser` uses an eager strategy.
Therefore, it is more appropriate to avoid applying alpha premultiplication within the `color` operator.

Fixes #16074.


src/ol/webgl/ShaderBuilder.js
```diff
@@ -28,10 +28,13 @@ uniform mediump int u_hitDetection;
...
 
+vec4 premultiplyAlpha(vec4 color) {
+  return vec4(color.rgb * color.a, color.a);
+}
+
 // this used to produce an alpha-premultiplied color from a texture
 vec4 samplePremultiplied(sampler2D sampler, vec2 texCoord) {
-  vec4 color = texture2D(sampler, texCoord);
-  return vec4(color.rgb * color.a, color.a);
+  return premultiplyAlpha(texture2D(sampler, texCoord));
```
jahow, line 37: I'm wondering if we can just get rid of this `samplePremultiplied` function altogether now

src/ol/webgl/ShaderBuilder.js
```diff
@@ -112,9 +115,9 @@ export class ShaderBuilder {
      * @type {string}
      * @private
      */
-    this.symbolColorExpression_ = colorToGlsl(
+    this.symbolColorExpression_ = `premultiplyAlpha(${colorToGlsl(
```
jahow, line 118: I think these should go in the final shader outputs in ShaderBuilder.js:

```diff
@@ -582,7 +582,7 @@ void main(void) {
   float c = cos(v_angle);
   float s = sin(v_angle);
   coordsPx = vec2(c * coordsPx.x - s * coordsPx.y, s * coordsPx.x + c * coordsPx.y);
-  gl_FragColor = ${this.symbolColorExpression_};
+  gl_FragColor = premultiplyAlpha(${this.symbolColorExpression_});
   if (u_hitDetection > 0) {
     if (gl_FragColor.a < 0.05) { discard; };
     gl_FragColor = v_prop_hitColor;
```
```diff
@@ -846,7 +846,7 @@ void main(void) {
   float currentLengthPx = max(0., min(dot(segmentTangent, startToPoint), segmentLength)) + v_distanceOffsetPx; 
   float currentRadiusPx = abs(dot(segmentNormal, startToPoint));
   float currentRadiusRatio = dot(segmentNormal, startToPoint) * 2. / v_width;
-  vec4 color = ${this.strokeColorExpression_} * u_globalAlpha;
+  vec4 color = premultiplyAlpha(${this.strokeColorExpression_}) * u_globalAlpha;
   float capType = ${this.strokeCapExpression_};
   float joinType = ${this.strokeJoinExpression_};
   float segmentStartDistance = computeSegmentPointDistance(currentPoint, v_segmentStart, v_segmentEnd, v_width, v_angleStart, capType, joinType);
```
```diff
@@ -954,7 +954,7 @@ void main(void) {
   }
   #endif
   if (${this.discardExpression_}) { discard; }
-  gl_FragColor = ${this.fillColorExpression_} * u_globalAlpha;
+  gl_FragColor = premultiplyAlpha(${this.fillColorExpression_}) * u_globalAlpha;
   if (u_hitDetection > 0) {
     if (gl_FragColor.a < 0.1) { discard; };
     gl_FragColor = v_prop_hitColor;
```

This way `premultiplyAlpha` doesn't have to be called in too many places.

src/ol/webgl/ShaderBuilder.js
```diff
@@ -150,9 +153,9 @@ export class ShaderBuilder {
      * @type {string}
      * @private
      */
-    this.strokeColorExpression_ = colorToGlsl(
+    this.strokeColorExpression_ = `premultiplyAlpha(${colorToGlsl(
```
jahow, line 156: same as above

src/ol/webgl/ShaderBuilder.js
```diff
@@ -189,9 +192,9 @@ export class ShaderBuilder {
      * @type {string}
      * @private
      */
-    this.fillColorExpression_ = colorToGlsl(
+    this.fillColorExpression_ = `premultiplyAlpha(${colorToGlsl(
```
jahow, line 195: same as above

src/ol/webgl/styleparser.js
```diff
@@ -304,21 +304,21 @@ function parseCircleProperties(
   // FILL COLOR
   let fillColor = null;
   if ('circle-fill-color' in style) {
-    fillColor = expressionToGlsl(
+    fillColor = `premultiplyAlpha(${expressionToGlsl(
```
jahow, line 307: I think we can leave this call out, see my comments on ShaderBuilder

src/ol/webgl/styleparser.js
```diff
@@ -304,21 +304,21 @@ function parseCircleProperties(
...
       ColorType,
-    );
+    )})`;
   }
 
   // STROKE COLOR
   let strokeColor = null;
   if ('circle-stroke-color' in style) {
-    strokeColor = expressionToGlsl(
+    strokeColor = `premultiplyAlpha(${expressionToGlsl(
```
jahow, line 317: same as above

src/ol/webgl/styleparser.js
```diff
@@ -422,21 +422,21 @@ function parseShapeProperties(
   // FILL COLOR
   let fillColor = null;
   if ('shape-fill-color' in style) {
-    fillColor = expressionToGlsl(
+    fillColor = `premultiplyAlpha(${expressionToGlsl(
```
jahow, line 425: same as above

src/ol/webgl/styleparser.js
```diff
@@ -422,21 +422,21 @@ function parseShapeProperties(
...
       ColorType,
-    );
+    )})`;
   }
 
   // STROKE COLOR
   let strokeColor = null;
   if ('shape-stroke-color' in style) {
-    strokeColor = expressionToGlsl(
+    strokeColor = `premultiplyAlpha(${expressionToGlsl(
```
jahow, line 435: same as above

src/ol/webgl/styleparser.js
```diff
@@ -506,7 +506,7 @@ function parseIconProperties(
   // COLOR
   let color = 'vec4(1.0)';
   if ('icon-color' in style) {
-    color = expressionToGlsl(fragContext, style['icon-color'], ColorType);
+    color = `premultiplyAlpha(${expressionToGlsl(fragContext, style['icon-color'], ColorType)})`;
```
jahow, line 509: same as above

src/ol/webgl/styleparser.js
```diff
@@ -628,7 +628,7 @@ function parseStrokeProperties(
 ) {
   if ('stroke-color' in style) {
     builder.setStrokeColorExpression(
-      expressionToGlsl(fragContext, style['stroke-color'], ColorType),
+      `premultiplyAlpha(${expressionToGlsl(fragContext, style['stroke-color'], ColorType)})`,
```
jahow, line 631: same as above

src/ol/webgl/styleparser.js
```diff
@@ -793,7 +793,7 @@ function parseFillProperties(
 ) {
   if ('fill-color' in style) {
     builder.setFillColorExpression(
-      expressionToGlsl(fragContext, style['fill-color'], ColorType),
+      `premultiplyAlpha(${expressionToGlsl(fragContext, style['fill-color'], ColorType)})`,
```
jahow, line 796: same as above

jahow (commented):
Thanks for this PR @kikuchan, after trying it out a little locally I think it makes sense. I honestly can't remember why I added alpha premultiplication in the `colorToGlsl` function back then, I mean it works but it seems way more straightforward to do this premultiplication on the final color in the shader.

I'm leaving a few comments that will hopefully make it simpler in the end, and then we can see if the rendering tests still pass! thanks!

kikuchan:
Hi @jahow, thank you for your review and suggestions!

Unfortunately, I tried rewriting the `styleparser`, but it didn't work because of the `mix` function.
The function doesn't work well with regular alpha blendings as you know, and I think that's why you implemented the current method previously.
Furthermore, since vector tile styles allow users to use `interpolate` operator, color values should be premultiplied by their alpha value before the operator is applied.

To address the issue of multiple of `premultiplyAlpha`, I've implemented a new approach: introducing a `premultiplyAlpha` context flag that tells the compiler whether or not to perform premultiplication.
Additionally, this fixes the `array`, `interpolate`, and some other operators within the color context by applying premultiplications as soon as possible during compilation.

What are your thoughts on this approach?
Please let me know, and I'm very happy to change my code!

Thanks!

jahow:
Ok, I think there is a way to make the alpha premultiplication much less of a concern throughout the code. See my branch here: https://github.com/jahow/openlayers/tree/fix-webgl-color-style

Alpha premultiplication is now only done at the end of each fragment shader. I kept all your changes throught the code to not multiply everything by alpha all the time, and also added some changes in the style parser to adapt the color expressions accordingly. The rendering tests pass, so it's a matter of adapting the unit tests now. What do you think? I feel like this would be the bast way forward.

Thank you for your continued efforts!

kikuchan:
Oh! It surely works! It's great!
I'm sorry I was wrong, I was missing some alpha premultiplications that need to be avoided in my WIP code.
Thank you for your code! It's been very helpful to compare it with mine.

I've sent a rebased and revised version.
Also, I added you as a 'Co-authored-by' in the commit message.

Thank you!

src/ol/webgl/ShaderBuilder.js
```diff
@@ -580,6 +574,7 @@ void main(void) {
   float s = sin(v_angle);
   coordsPx = vec2(c * coordsPx.x - s * coordsPx.y, s * coordsPx.x + c * coordsPx.y);
   gl_FragColor = ${this.symbolColorExpression_};
+  gl_FragColor.rgb *= gl_FragColor.a;
```
kikuchan, line 581: BTW, currently `u_globalAlpha` is not applied here. Is this OK for a symbol...?

jahow (approved):
That looks perfect, thank you for sticking to this!!

src/ol/webgl/ShaderBuilder.js
jahow: Oh, right, maybe that was missing. I think this should be done in another PR, I have the feeling this isn't tested anywhere

PR #16082, opened 2024-08-12
Add BaseEvent export from event.js entry file

oxyeez:
In order to fix some issues related to jest test runner that is not able to find the BaseEvent when imported from it's definition file for some reason.

### DIFF
--- a/src/ol/events.js
+++ b/src/ol/events.js
@@ -29,6 +29,8 @@ import {clear} from './obj.js';
  * @typedef {ListenerFunction|ListenerObject} Listener
  */
 
+export {default as BaseEvent } from './events/Event.js';
+
 /**
  * Registers an event listener on an event target. Inspired by
  * https://google.github.io/closure-library/api/source/closure/goog/events/events.js.src.html


ahocevar:
Interesting. Can you point to a minimal reproduction repository where this problem occurs?

oxyeez:
I can't because it's in a project of the company I work for, but I can describe it.
We are developing an internal angular library that wraps openlayers and makes it easier to use it in the final products. For a specific tool I had to create a custom event extending BaseEvent.
While testing it directly from the library jest works well.
But once built, and used within a project, when unit testing the project, jest could not find the BaseEvent and produces this error : 
    TypeError: Class extends value [object Object] is not a constructor or null.
And the error comes from the line where my custom event extending the BaseEvent is declared.

If you really need a reproduction repository I could do it, but it forces to create both a minimal library, and a minimal repository using the library, because the error appears only when the library is built and used from an other project.

ahocevar:
This looks more like an issue with the build tools you use for your library, so I don't think the change you suggest here is the best solution.

oxyeez:
Maybe, but we are working with nx, and I've tried the different nx/angular packager availables, but I have the same problem with all of them.
I've try many things and the only solution that worked was to export the BaseEvent from a file more at the root of the openlayers library, like it's done for many other elements of openlayers.

ahocevar:
The reason why BaseEvent is not re-exported from `events.js` is because it is not part of the official API.

PR #16085, opened 2024-08-12
Add `@types/rbush` as dependency

### LINKED ISSUE #16083: Regression: new required dependecy "rbush"
Building project that required openlayers does not longer work without also including rbush dependecy:

image

Previously it wat not required, is that a regression ?

MoonE:
Fixed #16083


ahocevar (approved):
Thanks, @MoonE

PR #16090, opened 2024-08-15
Add a CompositeTile source

kikuchan:
I want to composite a tile image (served by OSM, for example) with a floating-point data tile, so that I can access both at the same time via the `band` operator of WebGL style expressions.
This requires multiple textures to be available within a shader at the same time. However, it seems that `ol/DataTile` and `ol/webgl/TileTexture` currently don't support uploading multiple data types at once.

This PR adds support for this functionality.

Additionally, it introduces a new virtual utility source, `ol/source/CompositeTile`.
```js
new WebGLTileLayer({
  source: new CompositeTile({
    sources: [new OSM(), new SomeOtherSource()],
  }),
}),
```

This allows tile sources to be combined as if they were a single source, making the data equally accessible via the `band` operator.


mike-000:
It looks like you intend to accept reproj tiles in the composite instead of reprojecting composite sources?  In that case should the composite projection be the view projection and the tilegrid be `sources[0].getTileGridForProjection(viewProjection)`?  Note there may be conflicts with changes proposed in #15860 and #15988.

kikuchan:
Hi @mike-000, thank you for your comment!

Ah that's right, intended to accept a reprojected tile as an image.
Since the `CompositeTileSource` just hooks and proxies on `getTile` function, is it possible to set `projection` and `tileGrid` to simply `null` maybe?

Oh, and I'll take a look at these PRs.
Thanks.

mike-000:
Passing the projection from the renderer's `getTile` request onto the component sources is correct.  Projection and tilegrid should be set null in the composite source setup.  The renderer will call the composite source's `getTileGridForProjection` method so an override may be needed for that to return the result from a component source (and as in GeoTIFF if two component sources have incompatible grids throw an error?).

kikuchan:
Indeed, thank you!
So, how about this way?
It collects and redistributes projection and tileGrid mappings.

mike-000:
Be aware that currently if the requested projection is equivalent to `source.getProjection()` `getTileGridForProjection()` returns `source.getTileGrid()` (if set) regardless of what was set for the projection using `setTileGridForProjection()`.

#15988 would change that for a DataTile source if a `transformMatrix` is set - and a `[1, 0, 0, 1, 0, 0]` matrix (or some other override) could be used to "reproject" to a new grid with no rotating or flipping (following on from #15988 I would like to see a similar change for other raster tile sources as there is currently no way to use TileDebug for rotated grids).

kikuchan:
Yeah, I see. Thank you.
I'll keep an eye on that.

kikuchan:
`layer.getData()` now returns both the extended data and the color data of the pixel, even for canvas layers.
https://deploy-preview-16090--ol-site.netlify.app/en/latest/examples/composite-tile-source-pmtiles-elevation.html

Is this somewhat useful...? 🤔

src/ol/source/CompositeTile.js
```diff
@@ -0,0 +1,194 @@
...
+
+    const wrapX = sources.every((source) => source.getWrapX());
+    const attributions = (
+      /** @type {import('../Map.js').FrameState} */ frameState,
+    ) =>
+      sources
+        .map((source) => source.getAttributions())
+        .filter(Boolean)
+        .map((func) => func(frameState))
+        .join(' / ');
```
mike-000, line 56: Avoid hardcoding a separator.  If you push each attribution to an array the css will format it according to any application specific css.

e.g. custom separator

```
.ol-attribution li:not(:last-child):after {
  content: " / ";
}
```

or separate lines

```
.ol-attribution li {
  display: block;
}
```

The code in ol/source/Raster may be useful.

kikuchan: Ah! Thank you!
I'll change it to:
```patch
-        .map((func) => func(frameState))
-        .join(' / ');
+        .flatMap((func) => func(frameState));
```
looks reasonable?

mike-000:
In the example it looks like the OSM is not interpolated.  Is it possible to handle the different interpolate settings of the original sources?

PR #16094, opened 2024-08-16
Focus for maps placed in ShadowRoot in custom HTML elements

### LINKED ISSUE #16076: tabindex on map leading to non-working interactions when map el in shadowRoot
**Describe the bug**
I have the following situation:

- OL map's target element is inside of a `shadowRoot` of a custom DOM component
-  `tabindex="0"` is set on the map's target element in order to gain accessibility according to the corresponding[ official OL example](https://openlayers.org/en/latest/examples/accessible.html)
- Now the map interactions, like pan and zoom stop to work

I did some investigations and for my understanding the following happens:

-  `defaultInteractions` on the map have `onFocusOnly: true` as default
- [focusWithTabindex](https://github.com/openlayers/openlayers/blob/main/src/ol/events/condition.js#L97C14-L97C31) detects that there is a tabindex attribute on map's target and calls the `focus` method
- [focus](https://github.com/openlayers/openlayers/blob/main/src/ol/events/condition.js#L85) method tries to detect if the map target has the focus, but the current approach does not detect map focus correctly if in inside `shadowRoot`, since the `contains` functions used there does not reflect elements as `shadowRoot` children.

A possible solution could be using an extended `contains` functions, which iterates down to `shadowRoot` elements or there might be another approach to detect focus at all.

**To Reproduce**
1. Create an OL map and set its target to an element inside a [shadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) 
2. Ensure map target has an `tabindex="0"` attribute
3. Check map and see that map interactions do not work anymore.

**Expected behavior**
The  map interactions should work if the map has the focus, so the map can be panned and zoomed with mouse and keyboard.

chrismayer:
This ensures that focus detection and keyboard navigation for maps placed inside a `ShadowRoot` is working properly. So in case of having a custom HTML element holding an OpenLayers map inside its ShadowRoot, the `tabindex` attribute can be placed on the custom element itself and the keyboard-based interactions (and all others of course) are still working:

`<ol-map tabindex="0" id="map" class="map"></ol-map>`

Therefore the `focusWithTabindex` and `focus` conditions were extended and the `MapBrowserEvent` propagation in the `Map` component is adapted accordingly by allwoing the host of the custom HTML element to be the `keyboardEventTarget_`. Also added an example and some tests.

The code for the changeset in `src/ol/events/condition.js` was provided by @ahocevar in [#16076](https://github.com/openlayers/openlayers/issues/16076#issuecomment-2286387515). Thanks for this :+1: 

Closes #16076

### DIFF
--- a/src/ol/Map.js
+++ b/src/ol/Map.js
@@ -148,6 +148,9 @@ import {warn} from './console.js';
  * element itself or the `id` of the element. If not specified at construction
  * time, {@link module:ol/Map~Map#setTarget} must be called for the map to be
  * rendered. If passed by element, the container can be in a secondary document.
+ * In case your `target` element is inside a Shadow DOM you have to ensure that
+ * a possible `tabindex` attribute has to be placed on the target's host
+ * element, so it can be focused for key events to be emitted.
  * **Note:** CSS `transform` support for the target element is limited to `scale`.
  * @property {View|Promise<import("./View.js").ViewOptions>} [view] The map's view.  No layer sources will be
  * fetched unless this is specified at construction time or through
@@ -1668,6 +1671,9 @@ class Map extends BaseObject {
 
   /**
    * Set the target element to render this map into.
+   * In case the given `target` is inside a Shadow DOM you have to
+   * ensure that a possible `tabindex` attribute has to be placed on the
+   * target's host element, so it can be focused for key events to be emitted.
    * @param {HTMLElement|string} [target] The Element or id of the Element
    *     that the map is rendered in.
    * @observable
--- a/src/ol/Map.js
+++ b/src/ol/Map.js
@@ -1129,6 +1129,14 @@ class Map extends BaseObject {
         : doc;
       const target = /** @type {Node} */ (originalEvent.target);
 
+      const currentDoc =
+        rootNode instanceof ShadowRoot
+          ? rootNode.host === target
+            ? rootNode.host.ownerDocument
+            : rootNode
+          : rootNode === doc
+            ? doc.documentElement
+            : rootNode;
       if (
         // Abort if the target is a child of the container for elements whose events are not meant
         // to be handled by map interactions.
@@ -1137,12 +1145,7 @@ class Map extends BaseObject {
         // It's possible for the target to no longer be in the page if it has been removed in an
         // event listener, this might happen in a Control that recreates it's content based on
         // user interaction either manually or via a render in something like https://reactjs.org/
-        // skip check for maps place in ShadowRoot element structures
-        !(rootNode instanceof ShadowRoot)
-          ? !(rootNode === doc ? doc.documentElement : rootNode).contains(
-              target,
-            )
-          : false
+        !currentDoc.contains(target)
       ) {
         return;
       }
--- a/src/ol/Map.js
+++ b/src/ol/Map.js
@@ -1671,9 +1671,9 @@ class Map extends BaseObject {
 
   /**
    * Set the target element to render this map into.
-   * In case the given `target` is inside a Shadow DOM you have to
-   * ensure that a possible `tabindex` attribute has to be placed on the
-   * target's host element, so it can be focused for key events to be emitted.
+   * For accessibility (focus and keyboard events for map navigation), the `target` element must have a
+   *  properly configured `tabindex` attribute. If the `target` element is inside a Shadow DOM, the
+   *  `tabindex` atribute must be set on the custom element's host element.
    * @param {HTMLElement|string} [target] The Element or id of the Element
    *     that the map is rendered in.
    * @observable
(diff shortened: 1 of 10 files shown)


src/ol/Map.js
```diff
@@ -1136,7 +1137,12 @@ class Map extends BaseObject {
         // It's possible for the target to no longer be in the page if it has been removed in an
         // event listener, this might happen in a Control that recreates it's content based on
         // user interaction either manually or via a render in something like https://reactjs.org/
-        !(rootNode === doc ? doc.documentElement : rootNode).contains(target)
+        // skip check for maps place in ShadowRoot element structures
+        !(rootNode instanceof ShadowRoot)
+          ? !(rootNode === doc ? doc.documentElement : rootNode).contains(
+              target,
+            )
+          : false
```
ahocevar, line 1145: Why was this change necessary? I'm not sure if it's enough to just skip this check with a `ShadowRoot`. I think it would be better to use an equivalent logic in the `ShadowRoot` case, i.e. check if the shadowRoot's `documentElement.contains(target)`.

ahocevar (commented):
Thanks for your efforts, @chrismayer. I think some additional documentation explaining the different `tabindex` setup for custom elements would be good.

src/ol/Map.js
chrismayer: It becomes necessary because `target` is the custom component element (`<ol-map>`) in the `ShadowRoot` case. Therefore `rootNode.contains(target)` would be false. IMHO it is necessary to check is part of the `documentElement`. I'd suggest for the following:

```js
      // check if event target is a child of the container that is no longer
      // in the page
      let targetInPage;
      if (rootNode instanceof ShadowRoot) {
        targetInPage = doc.documentElement.contains(target);
      } else {
        targetInPage = (
          rootNode === doc ? doc.documentElement : rootNode
        ).contains(target);
      }

      if (
        // Abort if the target is a child of the container for elements whose events are not meant
        // to be handled by map interactions.
        this.overlayContainerStopEvent_.contains(target) ||
        // Abort if the event target is a child of the container that is no longer in the page.
        // It's possible for the target to no longer be in the page if it has been removed in an
        // event listener, this might happen in a Control that recreates it's content based on
        // user interaction either manually or via a render in something like https://reactjs.org/
        // skip check for maps place in ShadowRoot element structures
        !targetInPage
      ) {
        return;
      }
```

What do you thinks @ahocevar ? Should I go for it or do you prefer an other check?

chrismayer:
> I think some additional documentation explaining the different tabindex setup for custom elements would be good.

Where should it be placed @ahocevar? In the description of the example or somewhere else?

src/ol/Map.js
ahocevar: I think something simpler should work too, maintaining the previous logic:
```js
const currentDoc = rootNode instanceof ShadowRoot
  ? rootNode.host
  : rootNode === doc ? doc.documentElement : rootNode)
if (!currentDoc.contains(target)) {
  return;
}

Comment (chrismayer) [reply src/ol/Map.js]: Thanks for this @ahocevar, I picked this and added slightly extended in [03b95fa](https://github.com/openlayers/openlayers/pull/16094/commits/03b95fade0e00f96c877a0ca3285e68d22a81551)

**Comment by chrismayer** [comment]
API docs for `target` property and `setTarget` added with 515133d.

File: src/ol/Map.js [hunk 2]
```diff
@@ -148,6 +148,9 @@ import {warn} from './console.js';
  * element itself or the `id` of the element. If not specified at construction
  * time, {@link module:ol/Map~Map#setTarget} must be called for the map to be
  * rendered. If passed by element, the container can be in a secondary document.
+ * In case your `target` element is inside a Shadow DOM you have to ensure that
+ * a possible `tabindex` attribute has to be placed on the target's host
+ * element, so it can be focused for key events to be emitted.
```
ahocevar, line 153:
```suggestion
 * For accessibility (focus and keyboard events for map navigation), the `target` element must have a
 *  properly configured `tabindex` attribute. If the `target` element is inside a Shadow DOM, the
 *  `tabindex` atribute must be set on the custom element's host element.
```

src/ol/Map.js
```diff
@@ -1651,6 +1671,9 @@ class Map extends BaseObject {
 
   /**
    * Set the target element to render this map into.
+   * In case the given `target` is inside a Shadow DOM you have to
+   * ensure that a possible `tabindex` attribute has to be placed on the
+   * target's host element, so it can be focused for key events to be emitted.
```
ahocevar, line 1676:
```suggestion
   * For accessibility (focus and keyboard events for map navigation), the `target` element must have a
   *  properly configured `tabindex` attribute. If the `target` element is inside a Shadow DOM, the
   *  `tabindex` atribute must be set on the custom element's host element.
```

ahocevar (approved):

ahocevar:
Thanks, @chrismayer!

chrismayer:
Thanks for your input and the reviews @ahocevar.

PR #16099, opened 2024-08-17
Do not transform M dimension

### LINKED ISSUE #16098: Time in XYM coordinates incorrectly passed to proj4 as elevation
**Describe the bug**

Elevation in XYZ coordinates may affect the XY results  in some proj4 transforms, for example

` new Point([-4.004431525245309, 50.74081267230213, 0], 'XYZ').transform('EPSG:4326', 'EPSG:27700').getCoordinates()
`
 **[258665.59205862257, 95383.43722020995, 0]**

` new Point([-4.004431525245309, 50.74081267230213, 1723304052], 'XYZ').transform('EPSG:4326', 'EPSG:27700').getCoordinates()
`
 **[258574.18090452763, 95454.2295165936, 1723304052]**

Assuming the proj4 transform is correct at such extreme elevations that might be expected.

The same value as a time in XYM coordinates has the same effect on the XY coordinates

` new Point([-4.004431525245309, 50.74081267230213, 1723304052], 'XYM').transform('EPSG:4326', 'EPSG:27700').getCoordinates()
`
 **[258574.18090452763, 95454.2295165936, 1723304052]**

which is incorrect as the elevation is 0.

**Expected behavior**

If proj4 cannot distinguish elevation and time only elevation should be passed to it.

ahocevar:
Fixes #16098.

### DIFF
--- a/test/node/ol/geom/Point.test.js
+++ b/test/node/ol/geom/Point.test.js
@@ -1,10 +1,12 @@
 import Point from '../../../../src/ol/geom/Point.js';
 import expect from '../../expect.js';
+import proj4 from 'proj4';
 import sinon from 'sinon';
 import {
   get as getProjection,
   getTransformFromProjections,
 } from '../../../../src/ol/proj.js';
+import {register, unregister} from '../../../../src/ol/proj/proj4.js';
 
 describe('ol/geom/Point.js', function () {
   it('cannot be constructed with a null geometry', function () {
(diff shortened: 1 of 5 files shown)


test/node/ol/geom/Point.test.js
```diff
@@ -219,7 +222,19 @@ describe('ol/geom/Point.js', function () {
     });
   });
 
-  describe('#transform()', function () {
+  describe.only('#transform()', function () {
```
tschaub, line 225:
```suggestion
  describe('#transform()', function () {
```

tschaub (approved):

PR #16100, opened 2024-08-17
Fix canvas group changes on webgl layers (revised)

kikuchan:
This is a follow-up PR for #16051.

It properly fixes `WebGLVectorLayerRenderer` and `WebGLPointsLayerRenderer` by re-uploading their buffers, instead of triggering a source change event.
This resolves the flickering on group changes that occurred as a result of the mitigation.


tschaub (approved):
Thanks, @kikuchan

kikuchan:
Thank you!

PR #16115, opened 2024-08-21
Upload all palette textures before binding tile textures

### LINKED ISSUE #15471: WebGLTile layer rendering issue after setStyle operation with combination of alpha band matching
**Describe the bug**
WebGLTile layer rendering issue after setStyle operation with combination of alpha band matching.

Purpose: we are using palette rendering with index based on computation ( in this case we used only modulo operation ). layer style cannot be changed only by using variables and we would like to preserve alpha band of source geotiff.
We tried checking last band with multiple operators such as: case, match, >=, <= .
None of them works correctly after use of setStyle operation.

image

**To Reproduce**
We modified existing example to reproduce the behavior:
Example: https://codesandbox.io/p/sandbox/numpytile-forked-47mv36?file=%2Fmain.js%3A27%2C29

Please change the select value to Mix color and you can see invalid tile rendered after setStyle operation.
Rendering issue will be fixed after map move or zoom operation.
If only variables change is needed, layer.setStyleVariables renders correctly.

**Expected behavior**
I would like to change layer style in effective way without loading new data.

image

tschaub:
Currently, if you use `transition: 0` on a data tile source where the layer uses the `palette` operator, on the first render of the first tile, the palette is not used correctly.  This change fixes that by uploading data for all palettes before binding and rendering tile textures.

The `webgl-palette` rendering test demonstrates the issue if `transition: 0` is added.  This animation shows what that test looks like before and after the fix.

palette-fix
(the first render of the first tile (top-left) doesn't use the palette)

Fixes #15471.

### DIFF
--- a/src/ol/renderer/webgl/TileLayer.js
+++ b/src/ol/renderer/webgl/TileLayer.js
@@ -146,6 +146,11 @@ class WebGLTileLayerRenderer extends WebGLBaseTileLayerRenderer {
         this.fragmentShader_,
         this.vertexShader_,
       );
+      const gl = this.helper.getGL();
+      for (const paletteTexture of this.paletteTextures_) {
+        // upload the texture data
+        paletteTexture.getTexture(gl);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @tschaub

Issue #16122, opened 2024-08-23
MouseWheelZoom glitch when calling `view.setCenter()`


matthias-ccri:
**Describe the bug**
If you call `view.setCenter()` after the `MouseWheelZoom` interaction has handled the `wheel` event (in [handleEvent](https://github.com/openlayers/openlayers/blob/7a68da647c293e2009dea9af160a43c428f8bf77/src/ol/interaction/MouseWheelZoom.js#L175)) but before its `setTimeout` has a chance to finish and run `handleWheelZoom_` ([here](https://github.com/openlayers/openlayers/blob/7a68da647c293e2009dea9af160a43c428f8bf77/src/ol/interaction/MouseWheelZoom.js#L249-L252)), then the zoom interaction get corrupted.

Here's our use case. We have a global map and our users view regions that span the IDL (e.g. the pacific ocean), so we've run into the classic IDL problems. One solution we've put in place is "world jumping" (not sure if there's a different term...). This is: when the view crosses the IDL into a longitude past `[-180..180]`, we seamlessly "jump" the user to the other side of the world by adding or subtracting `360` to their longitude. For example, if they pan to `190` lon, we jump them to `-170` lon. To the user, they're still in the same place, but they've been moved to the other side of the globe.

We encountered a bug where _sometimes_ when a user is zooming across the IDL, they get transported to a random different longitude. Usually somewhere in the center of the world. 

### The bug
I tracked it down to the `MouseWheelZoom` class. Not sure if this is considered a bug, but I don't think it's fixable in userland without forking MouseWheelZoom. 

The bug happens when these things happen:
- The user does a wheel zoom, causing `MouseWheelZoom.handleEvent` to run
- `handleEvent` sets `this.lastAnchor_` to the event's coordinate.
- `handleEvent` then reaches a `setTimeout` that delays for a variable amount of time (`timeLeft`).
- While the timeout is waiting, our "world jumping" code receives a map `moveend` event, detects that the view is out of range, and calls `view.setCenter()`.
- Then the `setTimeout` in `handleEvent` fires, calling `handleWheelZoom_`, which references `this.lastAnchor_`. As a result, the zoom is calculated as if the mouse position is on the other side of the map, 360 degrees away from its actual position.

Is this a valid bug, or is it unsupported to change the view at the same time that a zoom interaction is happening?

[...]

PR #16123, opened 2024-08-23
Mark the type of InteractionOptions.handleEvent as optional

matthias-ccri:
A small change to mark `handleEvent` as optional. The property is optional based on its usage in `Interaction` as well as in the `MouseWheelZoom` subclass.

I ran into this while implementing a custom `MouseWheelZoom` interaction in typescript (working around this bug: https://github.com/openlayers/openlayers/issues/16122).


ahocevar (approved):
Thanks, @matthias-ccri

PR #16124, opened 2024-08-25
Store anchor pixel instead of coordinate

### LINKED ISSUE #16122: MouseWheelZoom glitch when calling `view.setCenter()`

ahocevar:
Fixes #16122

### DIFF
--- a/src/ol/interaction/MouseWheelZoom.js
+++ b/src/ol/interaction/MouseWheelZoom.js
@@ -104,7 +104,7 @@ class MouseWheelZoom extends Interaction {
 
     /**
      * @private
-     * @type {?import("../coordinate.js").Coordinate}
+     * @type {?import("../pixel.js").Pixel}
      */
     this.lastAnchor_ = null;
 
@@ -161,7 +161,7 @@ class MouseWheelZoom extends Interaction {
     view.endInteraction(
       undefined,
       this.lastDelta_ ? (this.lastDelta_ > 0 ? 1 : -1) : 0,
-      this.lastAnchor_,
+      this.lastAnchor_ ? map.getCoordinateFromPixel(this.lastAnchor_) : null,
     );
   }
 
@@ -188,7 +188,7 @@ class MouseWheelZoom extends Interaction {
     wheelEvent.preventDefault();
 
     if (this.useAnchor_) {
-      this.lastAnchor_ = mapBrowserEvent.coordinate;
+      this.lastAnchor_ = mapBrowserEvent.pixel;
     }
 
     // Delta normalisation inspired by
@@ -236,7 +236,10 @@ class MouseWheelZoom extends Interaction {
         this.endInteraction_.bind(this),
         this.timeout_,
       );
-      view.adjustZoom(-delta / this.deltaPerZoom_, this.lastAnchor_);
+      view.adjustZoom(
+        -delta / this.deltaPerZoom_,
+        this.lastAnchor_ ? map.getCoordinateFromPixel(this.lastAnchor_) : null,
+      );
       this.startTime_ = now;
       return false;
     }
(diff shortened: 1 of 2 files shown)


tschaub (approved):

matthias-ccri:
Unfortunately, the `pixel` fix doesn't fix the whole issue. The glitch is still present in ol version `10.1.1-dev.1725276208693`

See [codesandbox](https://codesandbox.io/p/sandbox/epsg-4326-forked-6h97kt).

https://github.com/user-attachments/assets/05c7fbe6-7f55-469c-82e9-7a536559a77e

PR #16141, opened 2024-08-28
Clear layer tile cache when view projection reset

### LINKED ISSUE #16136: Map not update when reprojected.
**Description**
When setting a new projection from the code, the map is not updated for the new projection.

**To Reproduce**
1. Go to [this ol example](https://openlayers.org/en/latest/examples/reprojection-by-code.html)
2. Select a new projection ex: EPSG:32661 (UPS North).
3. The map is well updated.
4. Select another projection ex: EPSG:32761 (UPS South).
5. The map is not updated. To update it correctly we need to make an explicit action triggering the refresh of it, like "Render reprojection edges" for exemple.

**Expected behavior**
In the previous versions, on change of the projection in the view, or set of a new view with different projection, the map was updated for the new projection without the need to force its refresh.

mike-000:
Fixes #16136


src/ol/renderer/canvas/TileLayer.js
```diff
@@ -206,6 +206,12 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
      */
     this.tileCache_ = new LRUCache(cacheSize);
 
+    /**
+     * @private
+     * @type {import("../../proj/Projection.js").default}
+     */
+    this.projection_ = undefined;
```
ahocevar, line 213: Maybe rename this to `renderedProjection_`?

ahocevar (changes requested):
Thanks, @mike-000. I have one minor suggestion.

ahocevar (approved):

ahocevar:
Thanks, @mike-000

PR #16157, opened 2024-09-04
Dispose of expired tiles

### LINKED ISSUE #16156: V10 seems to introduce a detached tile image memory leak
If zooming some and panning around a lot in this example:
https://openlayers.org/en/v9.2.4/examples/draw-features.html
Then the following detached elements can be found in chrome debugger:

Screenshot 2024-09-04 172921

There are both new and deleted detached images.

But if doing the same for a long time in this example (v10.1.0):
https://openlayers.org/en/latest/examples/draw-features.html
Then the following detached elements can be found in chrome debugger:

Screenshot 2024-09-04 173629

There are no deleted images at all! Only 22000+ new detached images.

I guess this might be a new raised caching limit, but then this limit has been raised alot between version 9 and 10. 

(A lot of IOS users have started getting map crashes in my application also, which might be due to this. It could be something else also)

ahocevar:
Fixes #16156.

By nullifying tile images when we expire tiles from the cache, we encourage the garbage collector to get rid of the detached `HTMLImageElement`s.

For vector tiles, we get rid of the `sourceTileCache` and unreference source tiles when they are not used any more by any render tile. Because `expireCache()` is no longer called on the source, the cleanup is now triggered by the tile's `release()` method.

### DIFF
--- a/src/ol/VectorRenderTile.js
+++ b/src/ol/VectorRenderTile.js
@@ -145,9 +145,11 @@ class VectorRenderTile extends Tile {
    * @override
    */
   release() {
-    releaseCanvas(this.context_);
-    canvasPool.push(this.context_.canvas);
-    this.context_ = null;
+    if (this.context_) {
+      releaseCanvas(this.context_);
+      canvasPool.push(this.context_.canvas);
+      this.context_ = null;
+    }
     super.release();
   }
 }
--- a/src/ol/VectorRenderTile.js
+++ b/src/ol/VectorRenderTile.js
@@ -88,9 +88,10 @@ class VectorRenderTile extends Tile {
     this.getSourceTiles = getSourceTiles.bind(undefined, this);
 
     /**
-     * @type {!function():void}
+     * @type {!function(VectorRenderTile):void}
+     * @private
      */
-    this.removeSourceTiles = removeSourceTiles.bind(undefined, this);
+    this.removeSourceTiles_ = removeSourceTiles;
 
     /**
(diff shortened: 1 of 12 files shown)


ahocevar:
For image tiles, the fix seems to be correct. However, I'm still experiencing increasing heap snapshot sizes with vector tiles. Investigating those now, hoping to be able to add another commit to fix those too.

ahocevar:
The memory leak of vector tiles is now also removed.

ahocevar:
I added tests, because the regressions that caused these memory leaks came from the fact that the involved code was widely untested.

This is now ready for review.

src/ol/VectorRenderTile.js
```diff
@@ -81,6 +87,11 @@ class VectorRenderTile extends Tile {
      */
     this.getSourceTiles = getSourceTiles.bind(undefined, this);
 
+    /**
+     * @type {!function():void}
+     */
+    this.removeSourceTiles = removeSourceTiles.bind(undefined, this);
```
tschaub, line 93: Not that important, but I'm curious why `bind` here instead of calling like `removeSourceTiles(this)` below? Was this a typescript thing?

tschaub (approved):

src/ol/VectorRenderTile.js
ahocevar: Only to make it consistent with `getSourceTiles()`. But you're right, it adds unnecessary overhead. I'm going to add a commit to change as you suggest.

PR #16167, opened 2024-09-08
Delete orphaned file

mike-000:
The interpolation example was (mostly) removed in #15905, which makes sense as an ImageTile source is now a better way of avoiding unwanted interpolation during reprojection because it is reprojected as a DataTile.  While the html and css were removed an orphaned js file remained.


ahocevar (approved):
Thanks, @mike-000

PR #16180, opened 2024-09-11
Fix getData() on reprojected tile layers

mike-000:
`getData()` is not working on reprojected canvas tile layers since version 10.0.0.

ReprojTile is not a subclass of ImageTile (unlike ReprojDataTile and DataTile) so both must compared when checking for tile type.

### DIFF
--- a/test/browser/spec/ol/layer/Tile.test.js
+++ b/test/browser/spec/ol/layer/Tile.test.js
@@ -69,6 +69,33 @@ describe('ol/layer/Tile', function () {
       expect(data[3]).to.be(255);
     });
 
+    it('gets reprojected pixel data', (done) => {
+      layer.setSource(
+        new XYZ({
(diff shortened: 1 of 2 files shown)


test/browser/spec/ol/layer/Tile.test.js
```diff
@@ -40,6 +40,8 @@ describe('ol/layer/Tile', function () {
       layer = new TileLayer({
         source: new XYZ({
           url: 'spec/ol/data/osm-0-0-0.png',
+          maxZoom: 0,
+          interpolate: false,
```
ahocevar, line 44: This affects the other tests within this `describe()` function. Can you please change the new test to create a TileLayer with these settings, instead of changing it for all tests here?

ahocevar (changes requested):

ahocevar (approved):
Thanks, @mike-000.

PR #16194, opened 2024-09-14
Use `@types/rbush` v3.0.3

### LINKED ISSUE #16138: Namespace RBush has no exported member 'default'
**Describe the bug**
when compiling the angular project, the angular compiler complains the rbush namespace has no 'default' member.
This most likely occurs because of the newly added dependency to @types/rbush, which does not present a default member.

image

**To Reproduce**
1. use newest ol in your angular project
3. compile the angular project
5. See error

ahocevar:
Fixes #16138

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "10.1.1-dev",
       "license": "BSD-2-Clause",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


MoonE (approved):
Thanks, @ahocevar

PR #16203, opened 2024-09-16
No double caching, distinct tile keys for queue

### LINKED ISSUE #16205: Error when zooming in and out multiple times in tile layers
**Describe the bug**
When zooming in and out multiple times on a map with tile layers, eventually tiles will stop loading and the following error message will appear in the console:
```
TypeError: null is not an object (evaluating 't.getImage().src=e')
```

**To Reproduce**
1. Go to https://openlayers.org/en/main/examples/wmts-hidpi.html
2. Zoom in fast until you reach street level. While tiles still load zoom out until you see a white border everywhere. Zoom in to a different are until you reach street level, and out again until you see a white border everywhere. Then zoom in and out to approximately the first location again. Repeat multiple times.
3. Tiles stop loading and the console shows the mentioned error.

Untitled

This issue was not present in v10.1.

**Expected behavior**
All tiles load, no error message.

ahocevar:
This pull request fixes two issues with tile rendering with the new TileLayer renderer:
* When using an `ol/source/TileImage`, tiles were cached by the source and the renderer. This resulted in disposed tiles being returned from the source cache - tiles that were disposed when expiring the renderer cache. The result was that when the queue tried to load such an expired tile, there was an error setting the src on the removed image. This pull request makes it so the source cache is bypassed in this case.
* When using multiple layers with `ol/source/TileImage`, the tile keys were the same for the same tile coordinate in each source. This caused the TileQueue to not queue tiles from subsequent layers. This pull request adds the source's uid as default prefix to the tile key to make it unique.

Fixes #16205

### DIFF
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -250,6 +250,7 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
         y,
         frameState.pixelRatio,
         frameState.viewState.projection,
+        true,
       );
       if (!tile) {
         return null;
--- a/src/ol/renderer/canvas/TileLayer.js
+++ b/src/ol/renderer/canvas/TileLayer.js
@@ -250,7 +250,6 @@ class CanvasTileLayerRenderer extends CanvasLayerRenderer {
         y,
         frameState.pixelRatio,
         frameState.viewState.projection,
-        true,
       );
       if (!tile) {
         return null;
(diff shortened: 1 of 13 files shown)


ahocevar:
I marked this as draft. I'm working on a commit that gets rid of the source's tile cache entirely.

ahocevar:
This was straightforward. All tests pass, all examples appear to work correctly. Ready for review.

jahow (approved):
Thanks! Could reproduce the issue and its resolution.

ahocevar:
Thanks for the review, @jahow!

tschaub:
Thank you for identifying this issue and fixing it, @ahocevar.

PR #16206, opened 2024-09-19
Fix getResolutionsForZoom() for views with only one resolution

ahocevar:
On a map with a view with just a single resolution configured, nothing will be rendered for tile layers. See https://codesandbox.io/p/sandbox/x529ls:
```js
  new View({
    center: [0, 0],
    zoom: 0,
    resolutions: [39135.75848201024],
  }),
```
This pull request fixes the `View#getResolutionsForZoom()` method, which is responsible for this problem, because it returns `0` in this case, which is not a resolution the `enqueueTiles()` method in the TileLayer renderer can work with.

### DIFF
--- a/src/ol/View.js
+++ b/src/ol/View.js
@@ -1310,9 +1310,9 @@ class View extends BaseObject {
    * @api
    */
   getResolutionForZoom(zoom) {
-    if (this.resolutions_) {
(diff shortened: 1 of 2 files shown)


jahow (approved):
Thanks!

PR #16208, opened 2024-09-21
Experimental Globe View

kikuchan:
This is an experimental (and somewhat hacky) implementation of a globe view with tilt.

Since it's achieved through post-processing, there are some limitations, including:
- Limited drawn area on the screen (due to extent mismatch).
- No support for rotation (yet).
- No support for extrusion.
- Only works on WebGL layers, and a canvas vector layer (so far).

It introduces concepts such as:
- **Subjective projection**: Projected coordinates depend on the current view (typically the center).
- **Final projection**: Lifts the restriction of affine transform when converting intermediate (flat) coordinates to pixel coordinates.

And, some notes:
- Had to use double-float arithmetic in GLSL for better precision at z > 18.
- Built on top of #16165.

It's still in progress, and any feedbacks are welcome!

Thank you.


kikuchan:
https://deploy-preview-16208--ol-site.netlify.app/en/latest/examples/globe.html

kikuchan:
Thank you for your interest!

Regarding double-float arithmetics, there are some downsides:

- We have to write expressions in a function call style.
- It’s hard to trick an optimizer for all devices.
- It’s slow, especially due to the Taylor series summation.

For expressions, we can use S-expressions, for example, to mitigate this, but I've yet to find solutions for the other issues.

If these issues don’t matter for the case, then yes, it could be possible!

PR #16221, opened 2024-09-24
Get rid of reprojection tile caches

### LINKED ISSUE #16218: Tile reprojection broken in v10.2.0
**Describe the bug**
Tile reprojection is broken in v10.2.0. Maybe this is tilecache issue or something. Reprojection kind of works but when zooming out the map has empty tiles.

[ol-1020-tile-reproj-bug.zip](https://github.com/user-attachments/files/17108654/ol-1020-tile-reproj-bug.zip)
ol-1020-tile-bug-example

**To Reproduce**
1. See attached zip project -> extract -> npm install -> npm run dev
2. Zoom in with scrollwheel and then start to zoom out (It might need few tries but for me it happens almost every time)
3. See empty tiles on map

**Expected behavior**
Reprojected tiles should work as it did with previous versions (Even on v10.1.0).

ahocevar:
With the renderer responsible for caching tiles, we don't need projection specific tile caches on the sources any more.

Fixes #16218.

### DIFF
--- a/src/ol/renderer/webgl/TileLayerBase.js
+++ b/src/ol/renderer/webgl/TileLayerBase.js
@@ -39,11 +39,6 @@ export const Uniforms = {
 };
 
 /**
- * @type {Object<string, boolean>}
- */
-const empty = {};
-
-/**
  * Transform a zoom level into a depth value; zoom level zero has a depth value of 0.5, and increasing values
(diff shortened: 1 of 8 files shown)


tschaub (approved):
Thank you, @ahocevar

PR #16230, opened 2024-09-27
Update URL and attributions of wmts-ign example

cvagner:
IGN has done a lot of work on its new public data infrastructure Géoplateforme. The switch was made a few months ago and the temporary server-side redirects will be removed on 09/30/2024.

API key are not necessary anymore for this service and logo has moved.

See https://geoservices.ign.fr/bascule-vers-la-geoplateforme


jahow (approved):
Thank you for thinking of this before the redirection ends!

Issue #16228, opened 2024-09-27
Base TileDebug on `ol/source/ImageTile` instead of `ol/source/XYZ`


mike-000:
#16227 introduces a `transformMatrix` property to `ol/source/DataTile`.  This cannot be easily debugged as there is no such property in `ol/source/TileImage` subclasses where, apart from using TileDebug for DataTile sources, it is unlikely to ever be needed.  To better support DataTile source debugging it might be better if `TileDebug` was implemented as subclass of `ol/source/ImageTile`.

PR #16231, opened 2024-09-28
Add `zDirection` option to ol/source/ImageTile

mike-000:
Adds the `zDirection` option to `ol/source/ImageTile` (passed via `ol/source/DataTile`) to make it more consistent with `ol/source/XYZ` where a custom `zDirection` function can be used to change tile zoom levels at view zooms z.5 instead of the OpenLayers default of z.415.

This is also a prerequisite for #16228


tschaub (approved):
Thanks, @mike-000

PR #16233, opened 2024-09-29
Fix ImageTile source `{-y}` placeholder handling

mike-000:
A TMS source such as

    new TileLayer({
      source: new XYZ({
        url: 'https://maps{1-4}.nyc.gov/tms/1.0.0/carto/basemap/{z}/{x}/{-y}.jpg',
      }),
    }),

works as expected when centered on New York City.  But if `XYZ` is changed to `ImageTile` it fails silently.
A `maxY` property is expected in the LoaderOptions but is never set.  Providing it using the same logic as used in a `tileUrlFunction` fixes the problem.


ahocevar (approved):
Thanks, @mike-000

PR #16242, opened 2024-09-30
Only track pointers as touches if they have fired a POINTERDOWN event

### LINKED ISSUE #16225: Drag event generation not compatible with (Apple) pencil hovering
**Describe the bug**

In the current version of MapBrowserEventHandler.js, dragging logic is incompatible with Apple Pencil hover (and possibly other stylus devices).

**Pointer events with Apple Pencil**
Apple Pencil with hover enabled will fire
- pointerover with pointerId 1 when entering the hover range,
- pointermove with pointerId 1 while moving without touching the screen,
- pointerover, pointerdown, pointermove, pointerup, pointerout events with a unique pointerId anytime the user touches the screen,
- pointerout with pointerId 1 when leaving the hover range.

**MapBrowserEventHandler.js dragging logic**

MapBrowserEventHandler.js currently tracks dragging by
1. waiting for a single pointerdown event
2. starting to listen to all pointermove and pointerup events
3. tracking all additional pointers encountered in pointermove and pointerdown events
4. ending the dragging condition after all tracked pointers have fired pointerup events

During the transition from touch to hover, the "hover pointer" (ID 1) may fire its first move event before the "touch pointer" fires its up event. This will implicitly add the hover pointer to the active pointers (3.) and never end the drag interaction because the hover pointer never fires pointerup (4.). Any subsequent touch will be added to the set alongside the hover pointer and cause unexpected pinch behavior.

**To Reproduce**
On an iPad/Pencil with hover enabled, touch the map, then keep hovering and move the stylus around and see the map pan / zoom out of control when touching again.

**Expected behavior**
Only track drags for pointers that have fired a down event before. This might be as simple as changing https://github.com/openlayers/openlayers/blob/f34586cb28c508d2333457d65db15bf13b0bf344/src/ol/MapBrowserEventHandler.js#L188-L191
such that a pointermove event is only put into `trackedTouches_` if the object already has its ID e.g.

```js
else if (
  event.type == MapBrowserEventType.POINTERDOWN ||
  (event.type == MapBrowserEventType.POINTERMOVE && id in this.trackedTouches_)
) {
  this.trackedTouches_[id] = event;
}
```

pgorczak:
This fixes #16225. Drag interactions now work on iPad without interference from the pencil hover-pointer.

I'm not sure about the comment above:

    Some platforms assign a new pointerId when the target changes.

This might constitute a case where we want to track a pointermove without a preceding pointerdown as a touch. Is there a known platform we can use to test against a possible regression?

### DIFF
--- a/src/ol/MapBrowserEventHandler.js
+++ b/src/ol/MapBrowserEventHandler.js
@@ -186,7 +186,8 @@ class MapBrowserEventHandler extends Target {
       }
     } else if (
       event.type == MapBrowserEventType.POINTERDOWN ||
-      event.type == MapBrowserEventType.POINTERMOVE
+      (id in this.trackedTouches_ &&
+        event.type == MapBrowserEventType.POINTERMOVE)
     ) {
       this.trackedTouches_[id] = event;
     }


ahocevar:
Thanks for the fix! This might indeed interfere with the commented code section you mentioned.

The problems related to that comment were reported with iOS and older Android devices. See https://github.com/openlayers/openlayers/pull/13771 for the list of associated issues.

ahocevar:
On the other hand, the only thing that could happen would be that a tracked pointer gets deleted on pointerup or pointercancel, so that should not matter.

It would be good if you could add a test with the event sequence you reported in #16225, to make this pull request complete.

ahocevar:
Oh, I understand now what you probably meant. With your change, we might indeed be losing a drag sequence when the pointer id changes due to a target change. Unfortunately I was not able to reproduce the issues myself back then.

ahocevar:
@pgorczak Would you be interested in finishing this according to my comment above?

PR #16245, opened 2024-10-01
Make TileDebug a subclass of ImageTile instead of XYZ

### LINKED ISSUE #16228: Base TileDebug on `ol/source/ImageTile` instead of `ol/source/XYZ`

mike-000:
Fixes #16228

Making TileDebug a subclass of ImageTile source will use the more efficient reprojection introduced in #15860, and will also allow easy support for any new options which might be added to DataTile source subclasses such as ModelTransformation support.

Although not included in this PR if #16227 is merged it would be easy to support debugging tile grid with transform matrix by providing a new TileDebug option and a getter for the GeoTIFF source

```
  view: cogSource.getView().then((viewConfig) => {
    viewConfig.projection = 'SR-ORG:97019';
    map.addLayer(
      new TileLayer({
        source: new TileDebug({
          projection: viewConfig.projection,
          tileGrid: cogSource.getTileGrid(),
          transformMatrix: cogSource.getTransformMtatrix(),
        }),
      })
    );
    return viewConfig;
  }),
```

image

### DIFF
--- a/src/ol/source/TileDebug.js
+++ b/src/ol/source/TileDebug.js
@@ -2,9 +2,9 @@
  * @module ol/source/TileDebug
  */
 
-import XYZ from './XYZ.js';
+import ImageTile from './ImageTile.js';
 import {createCanvasContext2D} from '../dom.js';
-import {toSize} from '../size.js';
+import {renderXYZTemplate} from '../uri.js';
 
 /**
  * @typedef {Object} Options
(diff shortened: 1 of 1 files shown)


tschaub (approved):
Thanks, @mike-000

PR #16249, opened 2024-10-02
Fix #15484 regression

### LINKED ISSUE #16248: #15484 regression not detected by rendering tests
**Describe the bug**
As noted in https://github.com/openlayers/openlayers/pull/15860#issuecomment-2384133322 #15860 appeared to have introduced a regression to #15484.  This can confirmed from the **webgl-data-tile-clip-extent-reproj** rendering test actual output which, despite being different from the expected, did not cause the test to fail.

**To Reproduce**
The expected and actual outputs from the final rendering test for #15860 clearly show a difference, with the actual output not being clipped as expected.
Expected
image
Actual
actual

**Expected behavior**
Although the test could be made more robust by increasing the `lineWidth` the regression should have been detected.  The regression also needs a fix.

mike-000:
Fixes #16248

Make the #15484 rendering test more robust, so until the fix is complete one should fail.

Update code to include `clipExtent` in source objects passed to `renderProjected` (initially without code to apply these to the `stitchTexture`)


ahocevar (approved):
Thanks, @mike-000

PR #16251, opened 2024-10-03
Fix Flow layer rotation

kikuchan:
This PR fixes the flux direction of the `Flow` layer when the view is rotated.

The issue is found and described in https://github.com/openlayers/openlayers/discussions/16222


tschaub (approved):

kikuchan:
Thank you!

PR #16254, opened 2024-10-03
Use fflate instead of jszip

ahocevar:
Addresses https://github.com/openlayers/openlayers/pull/16252#issuecomment-2391011253.

@MoonE This was more straightforward than making the example work with async data. Also, fflate is faster than jszip, and has no dependencies.


MoonE (commented):
The example description needs to be changed, too. Maybe just delete this line.
https://github.com/openlayers/openlayers/blob/3a3bb5505ead1156fc8d42ccbc0f333cee8c0942/examples/drag-and-drop-custom-kmz.html#L9

ahocevar:
@MoonE Good point, done.

MoonE (approved):
Thanks, @ahocevar, fflate looks like a good choice

Issue #16261, opened 2024-10-07
Performance issues with Translate interaction together with VectorTileLayer


MrLynx93:
When having a map with Translate interaction and 2 layers: 
- one layer with a few features that can be translated
- VectorTileLayer with a lot of features below that
The performance of dragging the feature is very poor. Just moving the mouse causes multiple calls of:
`Translate.handleMoveEvent -> Translate.js::featuresAtPixel_ -> map.forEachFeatureAtPixel` 
The problem is that the Translate interaction is created with `features` parameters provided and therefore `Translate.layerFilter_` is always `TRUE`. `Translate.featuresAtPixel_` always passes `layerFilter: undefined` to `map.forEachFeatureAtPixel`. This causes `handleMoveEvent` to search through all features in all layers.

1. Create a map:
2. Add a VectorTileLayer (or probably any layer with a lot of features)
3. Add a simple VectorLayer with one feature
4. Create Translate interaction for the VectorLayer. Pass in `features` parameter to the constructor.
5. Just move the mouse or drag the feature.

**Expected behavior**
The mousemove should not search through all the features in all visible layers, only the ones needed. One option would be to allow to pass in `layers` and `filter` to the `Translate` interaction, so that only the features from the translated layer are checked in handleMoveEvent.

![perfomance_mousemove](https://github.com/user-attachments/assets/817ae1db-154e-4f71-a68d-e3bb3784f59b)
![performance2](https://github.com/user-attachments/assets/0c81a6fa-88cc-4089-8834-858b43eb06a5)

PR #16262, opened 2024-10-07
#16261: Allow to pass in filter and layers parameter to Translate interaction

MrLynx93:
Allow to pass in filter and layers parameter to Translate interaction. Layers are used in `Translate.handleMoveEvent` (passed in as `this.layerFilter_` to `map.forEachFeatureAtPixel`


MrLynx93:
Thanks for the quick response @ahocevar . My problem is that I do pass in `features` and `layers` both. But `layers` are ignored because of this line: `if (options.layers && !this.features_)` which assigns `layerFilter_` only if you **didn't** pass in `features`.

ahocevar:
So what's the point of using both `features` and `layers`? Why don't you just pass `features`, e.g. from a source you created with `useSpatialIndex: false`?

PR #16268, opened 2024-10-08
Do not call postrender functions when there is no frameState

### LINKED ISSUE #16266: TypeError: Cannot read properties of null (reading 'wantedTiles')
**Describe the Bug:**

After upgrading to OpenLayers 10, we have experienced the following error:

**Error Message:**

TypeError: Cannot read properties of null (reading 'wantedTiles')
    at Array.postRenderFunction (TileLayer.js:847:38)
    at Map.handlePostRender (Map.js:1233:29)
    at Map.js:1643:14
    at timer (zone.js:1716:37)
    at _ZoneDelegate.invokeTask (zone.js:398:33)
    at core.mjs:14556:55
    at AsyncStackTaggingZoneSpec.onInvokeTask (core.mjs:14556:36)
    at _ZoneDelegate.invokeTask (zone.js:397:38)
    at Object.onInvokeTask (core.mjs:14869:33)
    at _ZoneDelegate.invokeTask (zone.js:397:38)

_Downgrading to OpenLayers 7 makes the error go away._

**To Reproduce:**
_Steps to reproduce the behavior:_

Show OpenLayers (OL) tiles.

Hide OL tiles.

Show OL tiles again.

See the error.

**Expected Behavior:**

The expected behavior would be that of OpenLayers 7 where "wantedTiles" does not cause an error.

ahocevar:
When the map has no size, the `renderFrame()` method does not do much. In this case, the `frameState` will be `null`. At the end of the  `renderFrame()` method, postrender functions are called. These functions expect the `frameState` as 2nd argument. If nothing was rendered, and if there is no frame state, I think we can safely say that we do not want to call postrender functions.

Fixes #16266.

### DIFF
--- a/src/ol/Map.js
+++ b/src/ol/Map.js
@@ -1229,8 +1229,10 @@ class Map extends BaseObject {
     }
 
     const postRenderFunctions = this.postRenderFunctions_;
-    for (let i = 0, ii = postRenderFunctions.length; i < ii; ++i) {
-      postRenderFunctions[i](this, frameState);
+    if (frameState) {
+      for (let i = 0, ii = postRenderFunctions.length; i < ii; ++i) {
+        postRenderFunctions[i](this, frameState);
+      }
     }
     postRenderFunctions.length = 0;
   }


tschaub (approved):

FunnBobby:
@ahocevar Will this be in a branch that's available soon?

ahocevar:
@FunnBobby You can `npm install ol@dev` and try it right away.

FunnBobby:
@ahocevar  Thank you so much! ...and thanks for all the work you do on Open Layers!

FunnBobby:
@ahocevar is this change in ol 10.2.1?  Maybe a better question is, how will I know what branch this is in?

M393:
Github shows the tags for a commit [here](https://github.com/openlayers/openlayers/pull/16268/commits/77d507421e65bd6427af4d776b1f4412269b0525), this commit does not yet belong to a release.

As a comparison [here](https://github.com/openlayers/openlayers/pull/16203/commits/bb8f92b0dabdcf110fbaddf547a79fc5fa8aba58) is an older commit which is in the v10.2.1 and v10.2.0 release.
image

FunnBobby:
@ahocevar @M393 @tschaub Hey Guys, I hate to revisit this but my company is asking when this might be a branch that we can use for our January release?  Currently we are using the DEV version in dev and it's working.  Any help on the typical timeline for something like this would help.  Thank you!

ahocevar:
@FunnBobby The plan is to publish a release just before or during the FOSS4G conference, i.e. in the first week of December.

Issue #16278, opened 2024-10-11
Incorrect retina tile setup in example, and incorrect reprojections


mike-000:
**Describe the bug**
The XYZ Retina Tiles example is not set up to indicate the tile pixel ratio to the source. If the view projection to set to EPSG:4326 it does not reproject correctly in 10.2.1 or dev versions.

**To Reproduce**
According to the docs the ImageTile source should be configured with
```
        tileSize: 512,
        tileGrid: createXYZ({tileSize: 256}),
```
However that does not fix the reprojection in either version.  Additionally in the dev version hidpi DataTile sources as in https://github.com/openlayers/openlayers/tree/main/test/rendering/cases/webgl-data-tile-tilepixelratio2 do not reproject correctly, (may be related to #15860).  It only seems to work if for one of the dimensions the texture and rendered sizes match while for the other is the texture is the same or smaller than the rendered size (which is usually true for GeoTIFF, therefore no tests failed).

**Expected behavior**
Fix the reprojection issues and add test(s), update the example if necessary.

PR #16279, opened 2024-10-12
Add `source` option to TileDebug

mike-000:
Adds a `source` option to TileDebug to allow easy use of settings from another tile source.
Adds a non-API getter to GeoTIFF to obtain to the transform matrix for any ModelTransformation.
Updates the `COG with ModelTransformation` example to show tile coordinates.

~~The PR currently uses one commit cherry picked from #16280 to enable correct reprojection of TileDebug.~~

Browser tests added, and one ModelTransformation rendering test updated to include TileDebug.


ahocevar:
I'm not in favour of exposing the transform matrix to the public API for now. It's good to have it internally, but let's wait a few release cycles and see if there's really a need for it to be used externally.

mike-000:
Yes, without the matrix the debug tiles would still appear, just not in the same place as the GeoTIFF.  The code setting tile (texture) sizes to integer values is still needed as many GeoTIFFs (including the one in the example) return a tile grid with fractional sizes which don't work with reprojection.  However any integer values should be valid, not just the rounded down sizes set here, so that could be moved to one fix for #16278 (which looks like two separate bugs).

mike-000:
I think a better way would be for TileDebug to take a `source` option.  `projection`, `tileGrid`, `wrapX`, `zDirection` and `transformMatrix` could be obtained from that without exposing any additional getters to the API.  The existing options, if set, could be used to override those obtained from `source`.  If not ready `source` would be disregarded, or an error thrown.

ahocevar:
Yes, that would make for a nice API. And it could also work when the source is not yet ready, by waiting until the source is ready before the TileDebug source is configured.

mike-000:
> Yes, that would make for a nice API. And it could also work when the source is not yet ready, by waiting until the source is ready before the TileDebug source is configured.

Done, and ready for another review.

src/ol/source/GeoTIFF.js
```diff
@@ -1017,6 +1017,14 @@ class GeoTIFFSource extends DataTile {
...
     return data;
   }
+
+  /**
+   * Return affine transformation matrix derived from the GeoTIFF ModelTransformation, if present.
+   * @return {import("../transform.js").Transform|null} Affine transformation matrix.
+   */
+  getTransformMatrix() {
+    return this.transformMatrix?.slice() || null;
+  }
```
ahocevar, line 1027: Since this is only used internally, it would be better to remove the `@protected` flag from `ol/source/DataTile#transformMatrix`, and get rid of this getter. On a related note, a getter returning a copy of something is not really a getter.

src/ol/source/TileDebug.js
```diff
@@ -36,32 +43,87 @@ class TileDebug extends ImageTile {
...
+      this.zDirection =
+        options.zDirection !== undefined
+          ? options.zDirection
+          : source !== undefined
+            ? source.zDirection
+            : this.zDirection;
+
+      if (source instanceof GeoTIFF) {
+        this.transformMatrix = source.getTransformMatrix();
+      }
```
ahocevar, line 80: The instanceof check should check for `DataTile` instead of `GeoTIFF`, and a copy of the member property instead of the getter should be used when assigning `this.transformMatrix`.

examples/cog-modeltransformation.html
```diff
@@ -6,6 +6,13 @@
...
   A ModelTransformation may rotate or skew the tile grid relative to the reported projection.
   This can be reprojected to that or any other supported projection.
+  To apply the ModelTransformation to `ol/source/TileDebug` use the `source` option.
 tags: "cog, projection, ModelTransformation"
 ---
 <div id="map" class="map"></div>
+<form>
+  <div class="form-check">
+    <input class="form-check-input" type="checkbox" id="show-tiles" checked />
+    <label class="form-check-label" for="show-tiles">Show tile coordinates</label><br>
```
ahocevar, line 16:
```suggestion
    <label class="form-check-label" for="show-tiles">Show source tile coordinates</label><br>
```

ahocevar (changes requested):

examples/cog-modeltransformation.js
```diff
@@ -12,11 +13,26 @@ register(proj4);
...
 
+const invert = (band, power) => ['-', 1, ['^', ['band', band], power]];
+const showTilesCheckbox = document.getElementById('show-tiles');
+const debugLayer = new TileLayer({
+  source: new TileDebug({source: cogSource}),
+  style: {
+    color: ['array', invert(1, 4), invert(2, 4), 0, ['band', 4]],
+  },
+  visible: showTilesCheckbox.checked,
+});
```
ahocevar, line 31:
```suggestion
const showTilesCheckbox = document.getElementById('show-tiles');
const debugLayer = new TileLayer({
  source: new TileDebug({source: cogSource}),
  visible: showTilesCheckbox.checked,
});
```
Nice way to change the colors, but let's try to keep the example's focus on what's in the title and description.

ahocevar (approved):
Thanks, @mike-000

PR #16280, opened 2024-10-12
Fix ImageTile/DataTile source hidpi reprojection

### LINKED ISSUE #16278: Incorrect retina tile setup in example, and incorrect reprojections

mike-000:
Fixes #16278

Fixes the ImageTile source problems described in #16278, which dated from #15905.
Updates the source setup in the example so it would work if reprojected.

Also fixes the hidpi DataTile problem described in #16278 which was introduced by #15860, adds a test and update results for two others where more pixels are made available in reprojection.

Incorrect reprojection when a hidpi ImageTile source is not defined as hidpi is also fixed.


src/ol/DataTile.js
```diff
@@ -148,13 +148,13 @@ class DataTile extends Tile {
    * @return {import('./size.js').Size} Tile size.
    */
   getSize() {
-    if (this.size_) {
-      return this.size_;
-    }
     const imageData = asImageLike(this.data_);
```
mike-000, line 151: This avoids having to accurately specify tile sizes when using oversized tiles.  But that might make it incompatible with gutter unless the tiles have a gutter property.

mike-000: This has been moved to #16299.

mike-000:
@ahocevar This is ready for review as it fixes the immediate bugs.

In `ReprojDataTile` the pixel ratio was not included in the returned image dimensions (resulting in incomplete images).  In some cases inconsistent rounding of fractional values resulted in corrupted images.

In `glreproj` the pixel ratio was not being included in the stitchTexture sizing, resulting in reduced quality (in the extreme case of half the rows and columns of black and white pixels being removed without interpolation the result was all black or all white).

mike-000:
In common with other hidpi rendering tests a27fba6 only demonstrates that when correctly configured the output is correct in a standard dpi screenshot (i.e. all of the source tile is used and all the reprojected data is exported in the reproj tile image).  There are currently no rendering tests which demonstrate the map canvas is genuinely hidpi.

If `pixelRatio: 2` is specified in the map options this is the map canvas export without d983fa2 blurred because the reprojection has not considered the tile pixel ratio in the stitch texture
image

and much sharper with d983fa2 particularly noticeable with the labels in the top right quadrant.
image

A regular rendering test would not detect the difference in a standard dpi screenshot.  A test which scaled the map div might work if it was compliant with the limitations of #14199
```
const style = document.getElementById('map').style;
style.position = 'relative';
style.left = '-128px';
style.top = '128px';
style.transform = 'scale(2)';
```

examples/xyz-retina.js
```diff
@@ -20,6 +21,8 @@ const map = new Map({
         url:
           'https://api.maptiler.com/maps/outdoor-v2/256/{z}/{x}/{y}@2x.png?key=' +
           key,
+        tileSize: 512,
+        tileGrid: createXYZ({tileSize: 256, maxZoom: 22}),
```
ahocevar, line 25: Why is this needed?

mike-000: It is equivalent to specifying `tilePixelRatio: 2` in `ol/source/XYZ`.  It is not needed in that example any more than `tilePixelRatio` was when the example used `ol/source/XYZ` as there is no reprojection, but as then it is the correct way to specify the source in case it is reprojected.

Since DataTile sources can have different pixel ratios in each dimension there is no `tilePixelRatio` option, see #13648.  However, I do think having a `tilePixelRatio` option for `ol/source/ImageTile` would be a nice future enhancement, where unless both `tileSize` and `tileGrid` were specified they could be calculated and passed to `super()`.

mike-000: I have now added a `tilePixelRatio` option so the source can be set up in a simpler way consistent with `ol/source/XYZ` (based on the code from #16279 which ensures data sizes are integer even if tile grid sizes are not).  The new hidpi map rendering test is using it.

Issue #16292, opened 2024-10-15
Set source projection from OGC TileMatrixSet crs


kudlav:
**Describe the bug**
 The following example throws correctly an error: `Unsupported CRS: "http://www.opengis.net/def/crs/EPSG/0/5514"`

```js
new Map({
  layers: [
    new TileLayer({
      source: new OGCMapTile({
        url: 'https://api-staging.maptiler.com/tiles/satellite/tiles?key=4dvaVugXIRQT6kn8BEnp',
      }),
    }),
    new TileLayer({
      source: new OGCMapTile({
        url: 'https://api-staging.maptiler.com/tiles/cz-satellite-5514/tiles?key=4dvaVugXIRQT6kn8BEnp',
      }),
    }),
  ],
  target: 'map',
  view: new View({
    center: [0, 0],
    zoom: 2,
  }),
});
```
When I define the projection, the error is fixed but the projection is not used:
```js
const projName = 'http://www.opengis.net/def/crs/EPSG/0/5514';
const projDef = '+proj=krovak +lat_0=49.5 +lon_0=24.8333333333333 +alpha=30.2881397527778 +k=0.9999 +x_0=0 +y_0=0 +ellps=bessel +towgs84=589,76,480,0,0,0,0 +units=m +no_defs +type=crs';
proj4.defs(projName, projDef);
register(proj4);
```
![obrazek](https://github.com/user-attachments/assets/534f9501-36e3-464c-94f8-ed5f25972cf6)

I have to set the projection of the source manually:
```js
source: new OGCMapTile({
  projection: projName
  url: 'https://api-staging.maptiler.com/tiles/cz-satellite-5514/tiles?key=4dvaVugXIRQT6kn8BEnp',
}),
```
![obrazek](https://github.com/user-attachments/assets/ea65a3c1-31a2-4986-abe9-2ceb5bcb5ba0)

But it should be set automatically, otherwise, what's the purpose of checking the existence of the projection?

**Expected behavior**
Propagate the projection from TileMatrixSet projection to the Source.
Users can still override the projection using the Source `projection` option.
Created PR with a fix: https://github.com/openlayers/openlayers/pull/16293

PR #16293, opened 2024-10-15
Set source projection from OGC TileMatrixSet crs

kudlav:
Propagate the projection from TileMatrixSet projection to the Source.
Users can still override the projection using the Source `projection` option.

Fixes https://github.com/openlayers/openlayers/issues/16292


tschaub (approved):

tschaub:
Thanks, @kudlav. I pushed an additional commit with some tests.

PR #16296, opened 2024-10-15
Improve the Modify interaction's API for programmatic actions

ahocevar:
This pull request improves the API of the Modify interaction for use cases that involve UI elements like e.g. a context menu, where uses can select actions like "Remove vertex" or "Insert vertex" when modifying linestrings or polygons:

* The existing `removePoint()` method now takes an optional `coordinate` as argument. In combination with the new `getPoint()` method, this can be used to remove a point later, without an immediate pointer action that matches the `deleteCondition`.
* A new `insertPoint()` method which takes an optional `coordinate` can be used to programmatically insert a vertex to a linestring or polygon, without an immediate pointer action that matches the `insertVertexCondition`.
* New `canInsertPoint()` and `canRemovePoint()` methods can be used to decide whether additional actions to insert or remove a vertex should be offered e.g. in a context menu.

### DIFF
--- a/src/ol/interaction/Modify.js
+++ b/src/ol/interaction/Modify.js
@@ -89,6 +89,10 @@ const ModifyEventType = {
  */
 
 /**
+ * @typedef {[SegmentData, number]} DragSegment
+ */
+
+/**
  * @typedef {Object} Options
  * @property {import("../events/condition.js").Condition} [condition] A function that
  * takes a {@link module:ol/MapBrowserEvent~MapBrowserEvent} and returns a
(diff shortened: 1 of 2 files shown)


src/ol/interaction/Modify.js
```diff
@@ -439,27 +444,24 @@ class Modify extends PointerInteraction {
...
       for (let i = 0, ii = segments.length; i < ii; ++i) {
-        const segment = segments[i];
-        for (let s = 0, ss = segment.length; s < ss; ++s) {
-          const feature = segment[s].feature;
-          if (feature && !features.includes(feature)) {
-            this.featuresBeingModified_.push(feature);
-          }
+        const feature = segments[i].feature;
+        if (feature && !features.includes(feature)) {
+          this.featuresBeingModified_.push(feature);
```
ahocevar, line 464: This change is not directly related - in trying to understand (again) what the Modify interaction does, I added the `DragSegment` typedef above, and using that type for the `dragSegments_` member revealed a problem with the code. While the nested loop did work, it made the code unnecessarily hard to read, and required nested arrays to be passed as the `willModifyFeatures_`'s `segments` argument.

Issue #16300, opened 2024-10-17
Event rendercomplete not firing on WebGLTileLayer


christophfriedrich:
_After learning in #16298 that I couldn't use those events as I thought, I fiddled with `rendercomplete` and that too doesn't work as I expected..._

**Describe the bug**
Using a WebGLTileLayer with a GeoTIFF source, the `rendercomplete` event is not fired by the layer despite being documented as doing so in https://openlayers.org/en/latest/apidoc/module-ol_layer_WebGLTile-WebGLTileLayer.html

**To Reproduce**
1. Clone the `cog` example: https://openlayers.org/en/latest/examples/cog.html
2. Define the layer separately and add simple event listeners to both the map and the layer:
```js
const layer = new TileLayer({
  source: source,
});

layer.on('rendercomplete', () => console.log('rendercomplete layer'));

const map = new Map({
  target: 'map',
  layers: [
    layer
  ],
  view: source.getView(),
});

map.on('rendercomplete', () => console.log('rendercomplete map'));
```
3. Start example application and see that only the "map" handler prints to the console, but not the "layer" handler.

**Expected behavior**
The event fires on both the layer and the map.

M393:
That looks like a bug in the documentation. Only the Map fires the `rendercomplete` event.

christophfriedrich:
Interesting, it's documented for `Heatmap` and `Graticule` too

![grafik](https://github.com/user-attachments/assets/2db795ed-c194-4e61-9ed5-15382519529a)

ahocevar:
That's because of incorrect `@fires` annotations in the respective layers. Thanks for finding that!

It should be
```js
 * @fires import("../render/Event.js").RenderEvent#prerender
 * @fires import("../render/Event.js").RenderEvent#postrender
```
not
```js
 * @fires import("../render/Event.js").RenderEvent

tschaub:
I corrected the `@fires` annotations for a few layers in #16301.

@christophfriedrich - I imagine `prerender` and `postrender` aren't really going to help you. If you can describe what you are trying to accomplish, maybe someone can point you in the right direction.

christophfriedrich:
Too bad it was just a documentation issue, but I'm glad I could at least help improving that 😂

[...]

PR #16301, opened 2024-10-17
Correct annotation for render events

tschaub:
This updates the `@fires` annotation for a few layers. Previously, the docs made it look like these layers fired the `rendercomplete` event. After this change, the docs show that they fire `prerender` and `postrender` events instead.

Thanks @ahocevar for pointing out the fix (see https://github.com/openlayers/openlayers/issues/16300#issuecomment-2419701370).


ahocevar (approved):
Thanks for fixing this, @tschaub!

PR #16302, opened 2024-10-17
Add a keepUpright parameter to control text labels rotation

leyan:
This PR adds a `keepUpright` parameter for text styling, in order to control whether text can be rendered upside-down or not when rendering along a line or polygon. Currently, OpenLayers hardcode a rotation to always keep the text "facing up". However, having a text label following the line direction even if it ends up upside down can be useful, for example when the label is a symbol arrow showing the direction of the line element.

The default behavior is not changed, it is only a facultative additional option providing more control.

This setting already exists in Mapbox Style Spec, so ol-mapbox-style will be able to use this parameter to follow more closely some style definitions. I first tried a workaround in ol-mapbox-style but @ahocevar advised me to add the option in OpenLayers instead (cf https://github.com/openlayers/ol-mapbox-style/pull/1206#pullrequestreview-2371577882).

I tried to copy how other options were passed, but I am not entirely sure about the different instructions arrays in TextBuilder.js, it was not clear to me why some have more instructions than others and whether my setting should go in them or not.


ahocevar (approved):
Thanks, @leyan. This looks great!

ahocevar (approved):

Issue #16244, opened 2024-09-30
COG with Projection Lookup example not working


mike-000:
**Describe the bug**
The COG with Projection Lookup example not working.  This seems to be due to the MapTiler API returning the same exports for all EPSG:27700  transforms, so transforms which supposedly have no grids return a proj4 definition containing a `+nadgrids` extry.

**To Reproduce**

For example `"grids": [],` suggests the proj4 definition should be useable, but it fails because it contains `+nadgrids=OSTN15_NTv2_OSGBtoETRS.gsb`:

        {
          "id": {
            "authority": "EPSG",
            "code": 1314
          },
          "name": "OSGB36 to WGS 84 (6)",
          "exports": {
            "proj4": "+proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 +x_0=400000 +y_0=-100000 +ellps=airy +nadgrids=OSTN15_NTv2_OSGBtoETRS.gsb +units=m +no_defs +type=crs",
            "wkt": "PROJCS[\"OSGB36 / British National Grid\",GEOGCS[\"OSGB36\",DATUM[\"Ordnance_Survey_of_Great_Britain_1936\",SPHEROID[\"Airy 1830\",6377563.396,299.3249646],EXTENSION[\"PROJ4_GRIDS\",\"OSTN15_NTv2_OSGBtoETRS.gsb\"]],PRIMEM[\"Greenwich\",0,AUTHORITY[\"EPSG\",\"8901\"]],UNIT[\"degree\",0.0174532925199433,AUTHORITY[\"EPSG\",\"9122\"]],AUTHORITY[\"EPSG\",\"4277\"]],PROJECTION[\"Transverse_Mercator\"],PARAMETER[\"latitude_of_origin\",49],PARAMETER[\"central_meridian\",-2],PARAMETER[\"scale_factor\",0.9996012717],PARAMETER[\"false_easting\",400000],PARAMETER[\"false_northing\",-100000],UNIT[\"metre\",1,AUTHORITY[\"EPSG\",\"9001\"]],AXIS[\"Easting\",EAST],AXIS[\"Northing\",NORTH],AUTHORITY[\"EPSG\",\"27700\"]]"
          },
          "unit": null,
          "accuracy": 2,
          "area": "United Kingdom (UK) - Great Britain - England and Wales onshore, Scotland onshore and Western Isles nearshore including Sea of the Hebrides and The Minch; Isle of Man onshore.",
[...]

PR #16314, opened 2024-10-24
Simplify the GeoTIFF reprojection example

### LINKED ISSUE #16311: Poorly named example "geotiff-reprojection"
**Describe the bug**
The "geotiff-reprojection" example includes a lot of functionality that has nothing to do with reprojection, and the reprojection itself is quite boring (4326 to 3857) and only implicit.

**To Reproduce**
1. Go to https://openlayers.org/en/latest/examples/geotiff-reprojection.html
2. See all the styling-related code and UI
3. See that terms like `proj` don't even appear in the code

**Expected behavior**
A rather minimal example (without fancy styling options) that showcases how to deal with a file that has a more exotic projection that requires explicit handling. Maybe showcased on a Map that additionally has a non-standard projection for its View.

**Suggestion**
The current example could be renamed to "geotiff-dynamic-rgb-stretching" or something, as it _is_ a cool example, but people wouldn't look for it under its current name.

**Additional remark**
The description says "GeoTIFF source with 20 bytes of raw values per pixel", but in fact it has 4 bands with 16 bits each (as the filename already suggests), so only 4*16/8 = 8 bytes per pixel.

tschaub:
This updates the `geotiff-reprojection.html` example so it shows how a GeoTIFF source can be reprojected. The example demonstrates using `proj4` to transform the image center to the view CRS before recentering the view.

The UTM -> Web Mercator transform works without fetching additional CRS info. The `cog-projection.html` example shows how additional CRS info can be fetched to support more transforms. That example is broken right now (see #16244).

Fixes #16311.


ahocevar (approved):

mike-000:
The example is not reprojecting (as can be seen if you add a background layer https://stackblitz.com/edit/js-17xvad?file=index.html,package.json,index.js - it should be on the Nile).  UTM projections are not predefined, so you need to call `fromEPSGCode` as in https://openlayers.org/en/main/examples/cog-modeltransformation.html otherwise it uses `identityTransform`.

ahocevar:
proj4@2.13.0 (https://github.com/proj4js/proj4js/pull/482) adds the UTM zones as predefined CRSs. I decided to add those for convenience.

mike-000:
> proj4@2.13.0 ([proj4js/proj4js#482](https://github.com/proj4js/proj4js/pull/482)) adds the UTM zones as predefined CRSs. I decided to add those for convenience.

It might still be nice to have a background layer to give the reprojection some context.  Close to the center of the zone and the equator it is not immediately obvious there is a reprojection (unlike https://openlayers.org/en/latest/examples/multiple-cogs.html where the two sources are noticeably non-parallel).

tschaub:
Even though proj4 can handle UTM transforms, I was wrong assuming that `register` was enough to make this magically happen. The transforms are not registered, so the identity transform is indeed used. There were a number of things that were surprising to me about this, so i'm going to work on making it less surprising.

mike-000:
> Even though proj4 can handle UTM transforms, I was wrong assuming that `register` was enough to make this magically happen.

The example is currently using proj4 version 2.12.1.  I have tried 2.13.0 and `register` will be sufficient.

christophfriedrich:
> so i'm going to work on making it less surprising

I like this thinking 👍👍👍 

Is there a preview of the current state of this PR? I'm not sure the link posted by the bot above reflects the most recent version.

IMO the example should include a situation where `proj4.defs(...)` **_is_** necessary, as that is the most complex situation, where one is especially grateful for guidance. A situation where it is _not_ necessary could be included as a second layer, or by adding a comment in the code like `// can be omitted if it's EPSG:...`

PR #16315, opened 2024-10-24
Add text-keep-upright to flat style attributes

leyan:
This is a followup to #16302, as I realized additional code is needed to make the attribute available to the "flat style" mode of style definition.


leyan:
@ahocevar I had a look in the test/ folder but I did not find any test dedicated to flat style definition that I could copy or build upon, how should I go about it?

ahocevar (approved):
Thanks @leyan, I changed the rendering test to a regression test.

PR #16318, opened 2024-10-25
Use epsg.io proj4 instead of proj4 from MapTiler API exports

### LINKED ISSUE #16244: COG with Projection Lookup example not working

mike-000:
Fixes #16244

This provide a workaround because the MapTiler API now longer exports more than the default transform.  The MapTiler API continues to be required to determine the best compatible transform to use, but the transform id is then used to obtain the proj4 definition from epsg.io.  This could be reverted if the MapTiler API reverts to its original behaviour.

Also rebased to use  proj4 v2.14, changed `fromEPSGCode` to use spatialrefence.org, updated the example to use that, and deprecated `epsgLookupMapTiler`.

The example is working as expected https://deploy-preview-16318--ol-site.netlify.app/en/latest/examples/cog-projection.html

### DIFF
--- a/examples/cog-projection.js
+++ b/examples/cog-projection.js
@@ -3,12 +3,7 @@ import Map from '../src/ol/Map.js';
 import TileLayer from '../src/ol/layer/WebGLTile.js';
 import XYZ from '../src/ol/source/XYZ.js';
 import proj4 from 'proj4';
-import {
-  epsgLookupMapTiler,
-  fromEPSGCode,
-  register,
-  setEPSGLookup,
-} from '../src/ol/proj/proj4.js';
+import {fromEPSGCode, register} from '../src/ol/proj/proj4.js';
 
 const key = 'get_your_own_D6rA4zTHduk6KOKTXzGB';
 const attributions =
@@ -16,7 +11,6 @@ const attributions =
   '<a href="https://www.openstreetmap.org/copyright" target="_blank">&copy; OpenStreetMap contributors</a>';
 
 register(proj4);
-setEPSGLookup(epsgLookupMapTiler(key));
 
 const cogSource = new GeoTIFF({
   sources: [


ahocevar:
In addition to this change, I'd be in favor of depecating the `epsgLookupMapTiler()`, and use the resurrected spatielrefence.org API for the default `epsgLookup()` function, using `https://spatialreference.org/ref/epsg/${code}/proj4.txt`.

mike-000:
The spatialreference definition for EPSG:27700 gives results very close to those for the most accurate non-grid transform if `Ordnance_Survey_of_Great_Britain_1936` is changed to `OSGB36` (without the change most results are over 100 meters out) so if an alias was supported that would work.

mike-000:
I do not think maintaining aliases may be necessary.  Many of the WKT definitions place the datum code in in the GEOGCS name, not in the DATUM name (which holds a description).  I found patching the following code from Proj.js

```
  if (json.datumCode && json.datumCode !== 'none') {
    var datumDef = match(Datum, json.datumCode);
    if (datumDef) {
      json.datum_params = json.datum_params || (datumDef.towgs84 ? datumDef.towgs84.split(',') : null);
      json.ellps = datumDef.ellipse;
      json.datumName = datumDef.datumName ? datumDef.datumName : json.datumCode;
    }
  }
```

to

```
  if (json.datumCode && json.datumCode !== 'none') {
    var datumDef = match(Datum, json.datumCode);
    if (!datumDef && json.GEOGCS && json.GEOGCS.name) {
      var geogcsName = json.GEOGCS.name.toLowerCase();
      datumDef = match(Datum, geogcsName);
      if (datumDef) {
        json.datumCode = geogcsName;
      }
    }
    if (datumDef) {
      json.datum_params = json.datum_params || (datumDef.towgs84 ? datumDef.towgs84.split(',') : null);
      json.ellps = datumDef.ellipse;
      json.datumName = datumDef.datumName ? datumDef.datumName : json.datumCode;
    }
  }
```
produces accurate results.  Otherwise it falls back to WGS84, which in the case of EPSG:27700 is equivalent the tourist with GPS  standing on the line at the Greenwich observatory.  Ideally the WKT parser should set `json.geogcsName` to avoid proj4 having the access the GEOGCS object.  It might also make checks for names such as `new_zealand_geodetic_datum_1949` in the WKT parser unnecessary.

ahocevar:
That's a bit of a hack. Better to allow looking up datums by the datum name, which proj4 v2.14 does.

TomPohys:
Hi, sorry for the inconvenience. We will fix the Coordinates API bug (returning default transformation `proj4` and `wkt` for all the transformations) on the MapTiler side as soon as possible.

mike-000:
> In addition to this change, I'd be in favor of depecating the `epsgLookupMapTiler()` function, and eventually even use the resurrected https://spatialrefence.org/ API for the default `epsgLookup()` function, using `https://spatialreference.org/ref/epsg/${code}/ogcwkt/`.

Done - proj4 v2.14 is required for `fromEPSGCode` to set an accurate transform.

TomPohys:
Hi, the issue was solved on the MapTiler side. Now the `proj4` and `wkt` correspond with the transformation code.

PR #16325, opened 2024-10-28
Pass the transformed coordinates

tschaub:
When working with projections using `tile-pixels` units, we first apply an affine transform before calling the projection transform function. While it isn't really currently a bug, things only work for simple geometries where the `applyTransform` function passes [the same input and output arrays](https://github.com/openlayers/openlayers/blob/38845b778d2c054bf192f6d6b7bac39918d44b1b/src/ol/geom/SimpleGeometry.js#L211-L212). This change updates the `transform` implementation so it would work if `applyTransform` didn't use the same input and output arrays.

### DIFF
--- a/src/ol/geom/Geometry.js
+++ b/src/ol/geom/Geometry.js
@@ -322,7 +322,7 @@ class Geometry extends BaseObject {
               0,
               0,
             );
-            transform2D(
+            const transformed = transform2D(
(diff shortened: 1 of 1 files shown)


tschaub:
Unrelated fix for the rendering test: #16326

ahocevar (approved):

PR #16327, opened 2024-10-28
Support for UTM transforms

tschaub:
This adds built-in support for transforming coordinates to and from WGS 84 / UTM zones. As with `EPSG:4326` and `EPSG:3857` transforms, with the changes in this branch nothing else is required to support raster reprojection or coordinate transforms using UTM.

The forward and inverse UTM transforms here are approximations that are not as accurate as those provided by PROJ. The tests use an epsilon of 0.1 m when comparing values to those from PROJ. For higher precision, Proj4js can be used.

In contrast with the existing built-in projection support, transforms for all 120 zones are not preemptively registered. Instead, when calling a transform function, if the supplied source or destination projection is UTM, the projection and transform functions are registered lazily.

I also changed the old behavior that used the identity transform when calling `transform` using an unknown source or destination projection. I found this behavior surprising, and instead think it makes more sense to throw an error if the user tries to transform between unsupported projections.

Fixes #16326.
Replaces #16314.

### DIFF
--- a/site/src/doc/tutorials/raster-reprojection.md
+++ b/site/src/doc/tutorials/raster-reprojection.md
@@ -5,13 +5,24 @@ layout: default.hbs
 
 # Raster Reprojection
 
-OpenLayers has an ability to display raster data from WMS, WMTS, static images and many other sources in a different coordinate system than delivered from the server.
+OpenLayers can display raster data from WMS, WMTS, static images, and many other sources in a different coordinate system than delivered from the server. In cases where the source projection differs from the map view projection, source data can be reprojected on the client (in the browser).
+
+OpenLayers comes with built-in support for transforming coordinates (and reprojecting rasters) between a handful of projections or coordinate reference systems.
+
+The built-in reprojection support applies to the following projections:
+
+ * WGS 84 / Geographic (`EPSG:4326`)
+ * WGS 84 / Web or Spherical Mercator (`EPSG:3857`)
+ * WGS 84 / Universal Transverse Mercator (`EPSG:32601` through `EPSG:32660` and `EPSG:32701` through `EPSG:32760`)
+
+For transforms between other, non-built-in projections, the Proj4js library can be used.
+
 Transformation of the map projections of the image happens directly in a web browser.
 The view in any Proj4js supported coordinate reference system is possible and previously incompatible layers can now be combined and overlaid.
 
 # Usage
 
-The API usage is very simple. Just specify proper projection (e.g. using [EPSG](https://epsg.io) code) on `ol/View`:
+The API usage for built-in projection support involves specifying the projection identifier on the source and the view. String [EPSG codes](https://epsg.io) can be used to identify the projections:
 
 ```js
 import Map from 'ol/Map.js';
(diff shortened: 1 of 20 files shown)


mike-000:
It looks like the failing rendering test may be due to transition on the debug tiles.  I see no point in adding another option to that, it might be better to simply pass `transition: 0` to the `super`.

tschaub:
> It looks like the failing rendering test may be due to transition on the debug tiles. I see no point in adding another option to that, it might be better to simply pass `transition: 0` to the `super`.

I added a fix for the (unrelated) failing rendering test to #16326.

ahocevar (commented):
I do like that transforming between unknown projections now throws an error. I'm not sure if it makes sense to add UTM projections to OpenLayers though. I do see the rationale behind it, but I'm wondering if dynamically importing proj4 when `fromEPSGCode()` is called wouldn't be a better option.

site/src/doc/tutorials/raster-reprojection.md
```diff
@@ -5,13 +5,24 @@ layout: default.hbs
...
-OpenLayers has an ability to display raster data from WMS, WMTS, static images and many other sources in a different coordinate system than delivered from the server.
+OpenLayers can display raster data from WMS, WMTS, static images, and many other sources in a different coordinate system than delivered from the server. In cases where the source projection differs from the map view projection, source data can be reprojected on the client (in the browser).
+
+OpenLayers comes with built-in support for transforming coordinates (and reprojecting rasters) between a handful of projections or coordinate reference systems.
+
+The built-in reprojection support applies to the following projections:
+
+ * WGS 84 / Geographic (`EPSG:4326`)
+ * WGS 84 / Web or Spherical Mercator (`EPSG:3857`)
+ * WGS 84 / Universal Transverse Mercator (`EPSG:32601` through `EPSG:32660` and `EPSG:32701` through `EPSG:32760`)
```
ahocevar, line 16: Since it not only affects raster reprojection, this should also be added to the class docs of `ol/proj/Projection`. It should also contain a note about the reduced accuracy, and that proj4js can be used to achieve higher accuracy if required.

mike-000:
It is appropriate to hard code the zone extents?  It could be a limitation where the map overlaps zone boundaries as in https://jsfiddle.net/1bd3xjnu/1/  Also in some places (e.g. Spain) the projection used can depend on provincial boundaries, not the official zone boundaries.

mike-000:
> It is appropriate to hard code the zone extents?

It certainly should not be setting an extent if the projection is already defined using proj4 or has yet to be used.  That was the real reason for the failing rendering test.

tschaub:
> It certainly should not be setting an extent if the projection is already defined using proj4 or has yet to be used

I updated the `register` function so it now makes use of proj4 transforms if `addProjection` has not already been called.

tschaub:
> It is appropriate to hard code the zone extents

The latest commit removes the projection extents (in line with those registered from proj4).

site/src/doc/tutorials/raster-reprojection.md
tschaub: The docs now look like this
image

Previously they looked like this
image

ahocevar (approved):

PR #16332, opened 2024-10-30
Revert "Remove debug tile layer in rendering test"

mike-000:
Reverts openlayers/openlayers#16329

The test was failing in #16327 because it was setting a UTM projection extent which resulted in a different tile grid for projection being used when the debug tiles were reprojected from EPSG:3857 to UTM.  Since there should have been no attempt to make any transforms before the view was set I do not think that code should have been called so soon,  It is setting a projection extent even if the projection is already defined in proj4.

The debug tiles were there to show the COG is corrected positioned when rotated.


tschaub:
> Since there should have been no attempt to make any transforms before the view was set I do not think that code should have been called so soon, It is setting a projection extent even if the projection is already defined in proj4.

I'm not sure I understand this. Here is what I see:

In the GeoTIFF source, we read the metadata and try to create a projection instance.  https://github.com/openlayers/openlayers/blob/4e35e39f60843021b7b13e25a7076c39f9dec3f2/src/ol/source/GeoTIFF.js#L727-L729

This is where the OpenLayers built-in UTM support is used in #16327.

After that, in your rendering test code, you register proj4. https://github.com/openlayers/openlayers/blob/4e35e39f60843021b7b13e25a7076c39f9dec3f2/test/rendering/cases/cog-rotation/main.js#L27-L34

Since you are not using any information from the resolved `viewConfig`, I think it would make sense to move the `proj4.defs()` and `register()` calls to the top.

tschaub:
Actually, it looks like `register` doesn't use the proj4 transforms if there already is one available.
https://github.com/openlayers/openlayers/blob/4e35e39f60843021b7b13e25a7076c39f9dec3f2/src/ol/proj/proj4.js#L52-L66

This was surprising to me. It would be nice if there was a way to say you prefer to use proj4.

tschaub:
> Actually, it looks like register doesn't use the proj4 transforms if there already is one available.

I updated this code in #16327 so that `register` will use the proj4 transforms if a projection has not been added with `addProjection` already.

PR #16333, opened 2024-10-31
Turn off transition in more rendering tests

mike-000:
Following on from #16326 several other tests are potentially affected by the same problem, as many are closely related and based on others, e.g. `layer-tile-reset-projection` was based on `webgl-tile-reset-projection`.  I found several such chains and updated each (unless transition was explicitly set to a non-zero value).  This may not be an exhaustive list.


ahocevar (approved):
Thanks, @mike-000

PR #16345, opened 2024-11-05
Add template to RBush forEach

M393:
This improves the return type for `forEach` and `forEachInExtent`.


tschaub (approved):
Thanks, @M393

PR #16348, opened 2024-11-06
Fix type generics for VectorImageLayer

### LINKED ISSUE #16347: VectorImage type declaration errors
Stack:
- OL v10.2.1 
- Typescript v5.3.3 

After the update from OL 8.2 to 10.2.1, I get the error **"Type 'VectorSource<Feature>' is missing the following properties from type 'Feature': id_, geometryName_, style_, styleFunction_, and 13 more.ts(2344)"** for a VectorImageLayer.

In my opinion I have the same effects like in the issue #15897

ahocevar:
Fixes #16347

### DIFF
--- a/src/ol/layer/BaseVector.js
+++ b/src/ol/layer/BaseVector.js
@@ -12,6 +12,11 @@ import {
   rulesToStyleFunction,
 } from '../render/canvas/style.js';
 
+/***
+ * @template T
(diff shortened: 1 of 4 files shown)


tschaub (approved):

bad-ch:
Hi, I think that will not fix the issue, that last working version for me was 9.2.0, every release afterwards 9.2.1 - 10.2.1 has this problem, which I has described in the issue #16347

bad-ch:
Agree sorry, after installing the dev version it solves the problem, so many thank's to you :o)

PR #16350, opened 2024-11-07
Update deprecated applyStyle syntax in Vector tiles in EPSG:4326 example

mike-000:
The ol-mapbox-style documentation describes the 4th parameter to `applyStyle` as deprecated as the 3rd parameter can be used for options.  https://deploy-preview-16350--ol-site.netlify.app/en/latest/examples/vector-tiles-4326.html


ahocevar (approved):
Thanks, @mike-000

PR #16352, opened 2024-11-08
Add Unsnap Event

### LINKED ISSUE #15404: Snap Interaction: dispatch event when not snapped anymore

RubenKad:
Fixes #15404

PR was based on https://github.com/openlayers/openlayers/pull/15494


RubenKad:
Is there a reason why the review takes a while? Did I miss anything?

RubenKad:
@ahocevar 
I tried to make a fork of https://github.com/openlayers/openlayers/pull/15494, but that was unfortunately not possible. Therefore, I did some small changes and here are some comments on the review comments of the older PR:
- https://github.com/openlayers/openlayers/pull/15494/files/1bab1e0a76c1628960932460b7ced74a9b1a12c3#r1475855149 -> I changed it to the suggested order
- https://github.com/openlayers/openlayers/pull/15494/files/1bab1e0a76c1628960932460b7ced74a9b1a12c3#r1475871267 and https://github.com/openlayers/openlayers/pull/15494/files/1bab1e0a76c1628960932460b7ced74a9b1a12c3#r1475864287 -> I implemented your suggestions
- https://github.com/openlayers/openlayers/pull/15494/files/1bab1e0a76c1628960932460b7ced74a9b1a12c3#r1475865446 -> This was not relevant anymore. I changed it a bit (in the old PR, only 1 snap event was thrown), such that multiple snap events are always thrown when the cursor moves, similarly as it behaved before.

The other changes are exactly the same, so I hope this is enough information for the review!

ahocevar (changes requested):
Thanks for the explanations, @RubenKad, now I was able to remember and properly review. The changes look good! Could you please also add a test, to avoid future regressions on this? Also, please run `npm run lint -- --fix` to apply the Prettier fixes.

examples/snap.js
```diff
@@ -131,4 +131,17 @@ ExampleModify.setActive(false);
...
+  document.getElementById('map').style.cursor = 'grabbing';
+  snappedElement.innerHTML = 'Snapped: true';
+});
+
+snap.on('unsnap', () => {
+  document.getElementById('map').style.cursor = 'default';
+  snappedElement.innerHTML = 'Snapped: false';
+});
+
+map.addInteraction(snap);
```
ahocevar, line 147:
```suggestion
map.addInteraction(snap);

```

ahocevar (approved):
Thanks a lot @RubenKad for picking this up and finishing it!

